PR #3079, opened 2015-04-27
Configuration changes for handling multiple of the same sections

ethomson:
You can have a configuration file with two sections with the same name.  For example:

    [section "foo"]
        name = value
    [section "bar"]
        barname = barvalue
    [section "foo"]
        name2 = value2

Our existing `config_write` makes assumptions that once it has seen a section it will never see that section again, and will simply copy the rest of the file in.  This means that with the above example, we cannot change `section.foo.name2`; we would add the new value to the first section, then copy the rest of the file, which would leave us with the second section.

Instead, we need to read the rest of the file when writing, in case we see another section with the same name.  Doing so means that we need to examine each line of the config file.

I've moved the common configuration parsing code into a function that will take callbacks to handle both the reading and writing scenarios.  I've also added a handful more tests to make sure that we're respectful when rewriting the files (eg, we don't delete comments or mangle whitespace unnecessarily.)

Finally, in `config_write`, take the lock before we read the existing configuration file to eliminate some races.

### DIFF
--- a/tests/config/write.c
+++ b/tests/config/write.c
@@ -230,12 +230,24 @@ void test_config_write__overwrite_value_with_duplicate_header(void)
 	git_config_free(cfg);
 }
 
+static int multivar_cb(const git_config_entry *entry, void *data)
+{
+	int *n = (int *)data;
+
+	cl_assert_equal_s(entry->value, "newurl");
+
+	(*n)++;
+
+	return 0;
+}
+
 void test_config_write__overwrite_multivar_within_duplicate_header(void)
 {
 	const char *file_name  = "config-duplicate-header";
 	const char *entry_name = "remote.origin.url";
 	git_config *cfg;
 	git_config_entry *entry;
+	int n = 0;
 
 	/* This config can occur after removing and re-adding the origin remote */
 	const char *file_content =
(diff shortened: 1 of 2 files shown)


tests/config/write.c
```diff
@@ -395,6 +517,75 @@ void test_config_write__outside_change(void)
...
+	cl_assert_equal_strn(SECTION_BAR, n, strlen(SECTION_BAR));
+	n += strlen(SECTION_BAR);
+
+	cl_assert_equal_s("[newsection]\n\tnewname = new_value\n", n);
+
+	git_buf_free(&newfile);
+	git_config_free(cfg);
+}
+
+void test_config_write__preserves_invalid_line(void)
```
carlosmn: What do you mean by "invalid line"? It looks to mean the line with just "name" in it, but that's a perfectly valid line in boolean context.

tests/config/write.c
```diff
@@ -145,6 +145,128 @@ void test_config_write__delete_value_with_duplicate_header(void)
...
+		"[remote \"origin\"]\n"		\
+		"	url = \"foo\"\n";
+
+	/* Write the test config and make sure the expected entry exists */
+	cl_git_mkfile(file_name, file_content);
+	cl_git_pass(git_config_open_ondisk(&cfg, file_name));
+	cl_git_pass(git_config_get_entry(&entry, cfg, entry_name));
+
+	/* Update that entry */
+	cl_git_pass(git_config_set_multivar(cfg, entry_name, ".*", "newurl"));
```
carlosmn: This should update both entries, should we test that they both got changed?

ethomson: Yes, we should; I updated with a test that iterates the entries and ensures that they were all updated.

tests/config/write.c
ethomson: Yep, that was a mistake.  Fixed.

PR #3081, opened 2015-04-28
Fix some build warnings

leoyanggit:
In checkout.c and filter.c we were casting a sub struct
to a parent struct which breaks the strict aliasing rules
in C. However we can use .parent or .base to access the
parent struct to avoid the build warnings.

In remote.c the local variable error was not initialized
or updated in some cases. For unintialized error a build
warning will be generated. So always keep error variable
up-to-date.

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -1457,7 +1457,7 @@ static int blob_content_to_file(
 	writer.fd = fd;
 	writer.open = 1;
 
-	error = git_filter_list_stream_blob(fl, blob, (git_writestream *)&writer);
+	error = git_filter_list_stream_blob(fl, blob, &writer.base);
 
 	assert(writer.open == 0);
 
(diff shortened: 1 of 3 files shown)


carlosmn:
We cast structs to their parent types all the time, I wonder if these are different because they're on the stack rather than the heap.

Looks good, :+1:

leoyanggit:
I'm not sure if they are different because of location. IMO we should avoid cast if possible. This way even if someone forget to place parent/base at the beginning of the child struct we are also fine. Thanks for your review and please land the patch if everything is OK.

carlosmn:
Not putting the base class at the start is not feasible without involving a `container_of`macro which isn't necessarily better at avoiding casts.

leoyanggit:
For those cases we could leave them AS-IS. With this patch my gcc 4.9.2 is not complaining about breaking strict aliasing rules. So I guess the rest of them is fine.

ethomson:
Nice, thanks!

:sparkles:

Issue #3082, opened 2015-04-28
Android build crashes in regfree() when using system's regex API


yongthecoder:
Currently the Android build uses deps/regex. That works. However Android does have the POSIX regular expression API. The real problem is libgit2 crashes in git_remote_create() when using Android's regex functions. The root reason is multivar_iter_free() calls regfree() even when have_regex=0.

My proposed solution is
1) always check "multivar_iter.have_regex" before calling regfree() or regexec()
2) set and use all_iter.has_regex
3) rename has_regex to have_regex or rename have_regex to has_regex, to make the naming consistent.
4) modify the cmake file so ANDROID doesn't need deps/regex.

yongthecoder:
Maybe this is also the reason why Solaris/Sun OS also needs deps/regex

yongthecoder:
I cannot find an Android 2.2 (froyo) device or emulator. But I think the original problem (https://groups.google.com/forum/#!topic/android-ndk/BepvBlDw4_w) is a packaging bug in Android NDK 4, and now people are using Android NDK 10 (which can still be used to develop froyo native libraries). The libc.so in Android NDK 10 for Android 2.2 (froyo) does contain the regex functions. (Froyo's Android API level is 8). This means the native library should be able to use these functions on Android 2.2 devices. Removing deps/regex from Android build can only break those projects that still use Android NDK 4. Should we worry about that?

android-ndk-r10c/platforms/android-8/arch-arm/usr/lib$ arm-linux-androideabi-objdump -t libc.so | grep reg
00008044 g     F .text	00000014 __atexit_register_cleanup
0000b230 g     F .text	00000014 regcomp
0000b244 g     F .text	00000014 regerror
0000b258 g     F .text	00000014 regexec
0000b26c g     F .text	00000014 regfree
0000b6b8 g     F .text	00000014 setregid

yongthecoder:
Of course the crash should be fixed first

carlosmn:
Fixed via #3086

PR #3083, opened 2015-04-28
Fix crash in regfree() on Android when using system's regex API

yongthecoder:
config.c: should always check have_regex before calling regfree() and regexec

With this, Android build no longer requires deps/regex

Also see https://github.com/libgit2/libgit2/issues/3082

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -343,7 +343,6 @@ typedef struct {
 	git_config_iterator *current;
 	const git_config *cfg;
 	regex_t regex;
-	int has_regex;
 	size_t i;
 } all_iter;
 
@@ -983,7 +982,8 @@ void multivar_iter_free(git_config_iterator *_iter)
 	iter->iter->free(iter->iter);
 
 	git__free(iter->name);
-	regfree(&iter->regex);
+	if (iter->have_regex)
+		regfree(&iter->regex);
 	git__free(iter);
 }
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,8 +34,7 @@ OPTION( PROFILE				"Generate profiling information"		OFF )
 OPTION( ENABLE_TRACE		"Enables tracing support"				OFF )
 OPTION( LIBGIT2_FILENAME	"Name of the produced binary"			OFF )
 
-OPTION( ANDROID				"Build for android NDK"	 				OFF )
-
+OPTION( USE_OPENSSL                     "Link with and use openssl library"             ON )
 OPTION( USE_ICONV			"Link with and use iconv library" 		OFF )
 OPTION( USE_SSH				"Link with libssh to enable SSH support" ON )
 OPTION( USE_GSSAPI			"Link with libgssapi for SPNEGO auth"   OFF )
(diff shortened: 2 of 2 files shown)


ethomson:
It seems like a previous version of the android libc [erroneously excluded the regex functions](https://groups.google.com/forum/#!topic/android-ndk/BepvBlDw4_w), which is why this test is the way it is.

I think that this change means that we have to drop froyo support.  However I am keen to know what release this was solidly fixed in.  Do you have previous OS versions to test against?

src/config.c
```diff
@@ -421,7 +421,7 @@ static int all_iter_glob_next(git_config_entry **entry, git_config_iterator *_it
 	 */
 	while ((error = all_iter_next(entry, _iter)) == 0) {
 		/* skip non-matching keys if regexp was provided */
-		if (regexec(&iter->regex, (*entry)->name, 0, NULL, 0) != 0)
+		if (iter->have_regex && (&iter->regex, (*entry)->name, 0, NULL, 0) != 0)
```
ethomson: I think that you dropped the actual function call here...

yongthecoder: oops...

carlosmn:
The commit message suggests that `regfree()` not supporting a `NULL` argument is the reason why we include regex in the android builds, which is not the case at all. As it appears that some versions of android did mistakenly miss these functions in their libc, the commit should explain that this is the reason.

src/config.c
```diff
@@ -343,7 +343,7 @@ typedef struct {
 	git_config_iterator *current;
 	const git_config *cfg;
 	regex_t regex;
-	int has_regex;
+	int have_regex;
```
carlosmn: Why this rename? It seems completely unnecessary.

yongthecoder:
Thanks. I think it is a bug in previous android NDK and it should work on Android 2.2 devices. But I'll try to find a way to test it on Android 2.2.

carlosmn:
All of these changes are to the glob iterator, which should *always* have a regex (which would be the reason why the field ended up unused), so it seems suspicious that we're checking for the flag to be set when these functions are all the implementations for the iterator which with the regex. When would this flag be false when these functions get called? If that's something that happens, that would indicate the source of the bug.

src/config.c
yongthecoder: because multivar_iter has a flag "have_regex". I thought they should be consistent. But now I realize the difference may make sense and "have" is used for multiple entries. Is that true?

yongthecoder:
The only line that fixes the crash is in multivar_iter_free(). multivar_iter.have_regex is being used. It is just not checked for regfree().

For all_iter, I'm not sure and I haven't seen a crash. So we can instead remove the flag?

carlosmn:
Yes, any checks in the glob iterator are unnecessary. Remove the flag instead. All of these changes obscure what this patch is trying to achieve.

carlosmn:
It turns out that the standard says that passing a value to `regfree` which doesn't come from `regcomp` is undefined behaviour, so it looks like the GNU implementation accepts a zeored value and knows it's unimplemented, but whatever the Android implementation is does not have this check. So in order to adhere to the standard, we need check in the case where we may not have compiled the regex.

yongthecoder:
I found there are still cases that all_iter doesn't have a defined regex.

also, it seems we should not call regfree() when regcomp() fails according the sample code:

http://pubs.opengroup.org/onlinepubs/9699919799/functions/regfree.html

PR #3085, opened 2015-04-29
Do not call regfree() on an empty regex that is not successfully created...

yongthecoder:
... by regcomp

(also removed an unused member "has_regex" from all_iter)

See http://pubs.opengroup.org/onlinepubs/9699919799/functions/regfree.html

When regcomp() fails, the "regex" will be undefined. The sample code doesn't call regfree() on regcomp() failure.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -343,7 +343,6 @@ typedef struct {
 	git_config_iterator *current;
 	const git_config *cfg;
 	regex_t regex;
-	int has_regex;
 	size_t i;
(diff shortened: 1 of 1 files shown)


carlosmn:
Thanks. For future reference, please don't open a new pull request for every small change. Instead push to the branch of your existing pull request and it will update, which makes it easier to follow the progress.

yongthecoder:
Thanks for the tip!

PR #3089, opened 2015-04-30
Added call to git_libgit2_shutdown()

### LINKED ISSUE #3076: Looks like git2.c example is missing git_libgit2_shutdown(); call.
It calls git_libgit2_init(), but not shutdown.

volftomas:
Added forgotten call to git_libgit2_shutdown() to the /examples/network/git2.c.

Should fix #3076

### DIFF
--- a/examples/network/git2.c
+++ b/examples/network/git2.c
@@ -23,10 +23,8 @@ static int run_command(git_cb fn, int argc, char **argv)
 	int error;
 	git_repository *repo;
 
-	git_libgit2_init();
-
-// Before running the actual command, create an instance of the local
-// repository and pass it to the function.
+	// Before running the actual command, create an instance of the local
+	// repository and pass it to the function.
 
 	error = git_repository_open(&repo, ".git");
 	if (error < 0)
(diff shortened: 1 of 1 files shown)


ethomson:
Thank you!

If you wouldn't mind tweaking this a little bit, I think that our recommendation is generally to call `git_libgit2_init` at the start of program execution and call `git_libgit2_shutdown` at the end.  There shouldn't be anything wrong with the pattern in this PR - which calls `init` and `shutdown` repeatedly, but it is less efficient as it would allocate / deallocate data repeatedly.

Since this is an example, we would like to point people towards the most recommended style.

I think that keeping the `init` in its current location, having a `shutdown` at the end of the `main` function would be the most recommended for consumers.  (Looking at the existing logic, of course, we would have to change that a bit so that we no longer return early from `run_command` and make sure to do the cleanup in `main`.)

Do you agree?  If so, would you mind updating this pull request?

Thanks again!

volftomas:
Sounds like a good idea, will update pull request today/tomorrow ^_^

volftomas:
Look it over, should be nicer now.

ethomson:
It is!  This looks great; thanks!

PR #3092, opened 2015-05-01
Don't search iconv in /opt/local

techee:
Since OpenSSL isn't used any more on OS X, there is no dependency
on any MacPorts library under /opt/local and there is no danger of
conflicts between MacPorts and system iconv. For this reason the
system iconv can always be used now.


ethomson:
This is amazing.  :sparkles: 

Thanks for this, and thanks to @carlosmn for the work to help us get here.

learnopengles:
This may be breaking Rust builds that depend on libgit2 when the user is using MacPorts: https://github.com/rust-lang/cargo/issues/3538#issuecomment-272526937

Is there a way to override this behavior somehow?

PR #3097, opened 2015-05-05
Remove run-time configuration settings from submodules

carlosmn:
It's me again, breaking all your code. Hopefully it's for the better.

I was doing the same thing to submodules as to remotes, but found that we have even deeper problems here. We currently have a per-repo cache of submodules, which unfortunately makes code like

```C
puts(git_submodule_url(sm));
```

susceptible to crashing, as we dish out submodules from that cache, but make them mutable so a different thread might have got the same intsance and may be setting the url or simply reloading it.

The first commit gets rid of this cache and returns a new instance per lookup. It leaves some tests broken, but that gets fixed later in "bring back finding by path" and "correct detecting of existing submodules"

The rest of the commits get rid of the load-modify-save pseudo-ORM which was a bad idea to introduce.

- The setters affect the configuration in the repository/`.gitmodules`
- The setters know when to remove the setting from the config (for the `NONE`/`NO` option)
- An instance does not change any configuration options
- In order to run an operation with different settings, you pass the setting you want to use

I finally got rid of `git_remote_save()` but noticed that it has logic for updating the path, bu we have no setter for that, so *shrug*. We can introduce it later.

### DIFF
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -399,7 +399,7 @@ typedef struct {
  * `git_diff_options_init` programmatic initialization.
  */
 #define GIT_DIFF_OPTIONS_INIT \
-	{GIT_DIFF_OPTIONS_VERSION, 0, GIT_SUBMODULE_IGNORE_DEFAULT, {NULL,0}, NULL, NULL, 3}
+	{GIT_DIFF_OPTIONS_VERSION, 0, GIT_SUBMODULE_IGNORE_FALLBACK, {NULL,0}, NULL, NULL, 3}
 
 /**
  * Initializes a `git_diff_options` with default values. Equivalent to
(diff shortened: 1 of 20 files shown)


carlosmn:
This should be ready for review now.

include/git2/diff.h
```diff
@@ -399,7 +399,7 @@ typedef struct {
  * `git_diff_options_init` programmatic initialization.
  */
 #define GIT_DIFF_OPTIONS_INIT \
-	{GIT_DIFF_OPTIONS_VERSION, 0, GIT_SUBMODULE_IGNORE_DEFAULT, {NULL,0}, NULL, NULL, 3}
+	{GIT_DIFF_OPTIONS_VERSION, 0, GIT_SUBMODULE_IGNORE_FALLBACK, {NULL,0}, NULL, NULL, 3}
```
jamill: Had we discussed using `UNSPECIFIED` instead of `FALLBACK` to represent this?

carlosmn: Yeah, you'd mentioned a name for using this in libgit2sharp which I liked, but I had since forgotten what it was. I'll change them to the new name.

arthurschreiber:
You forgot to mention this in the `CHANGELOG.md` file.

PR #3066, opened 2015-04-22
Remove the configuration state we keep in the remote

carlosmn:
There's way to much state which we keep in the remote, which can be changed at run-time and maybe saved awkwardly to the configuration.

As a first step, I've removed `git_remote_{s,g}et_callbacks` and added `git_fetch_options` to keep track of the callbacks, maybe in this PR or later.

My end goal for the remote is to remove `git_remote_save()` and have changes happen on-disc but have the remote's properties be read-only from the point it was looked up and the options for fetch or push be determined via the options you pass. See #2711 for some more on this.

PR #3102, opened 2015-05-06
Show progress during packing for the local transport

carlosmn:
Related to #1902 , start showing git-similar text during the packing process to show the client that we are doing something.


carlosmn:
This touches the callbacks so let's wait for #3066 to get merged.

ethomson:
I don't think that you meant to include your `examples/Makefile` here.

carlosmn:
I did not, it sneaked in during the rebase.

PR #3103, opened 2015-05-07
Use the packbuilder in local push

carlosmn:
This is what #1902 was complaining about all that time ago.

What we currently do is likely one of the worst ways of pushing. By actually using the packbuilder we also get the progress which we lack.

### DIFF
--- a/src/transports/local.c
+++ b/src/transports/local.c
@@ -289,50 +289,6 @@ static int local_negotiate_fetch(
 	return 0;
 }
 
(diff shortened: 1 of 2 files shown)


carlosmn:
The Travis error seems to be saying that at some point we're pushing into our resources dir instead of a scratch repo, which is really not good.

carlosmn:
This touches the callbacks so let's wait for #3066 to get merged.

PR #3104, opened 2015-05-07
Adjusting stream buffer size to 64KB

whoisj:
64K is optimal buffer size per https://technet.microsoft.com/en-us/library/cc938632.aspx. This change should benefit any x86 based system, not just Windows.

### DIFF
--- a/src/filter.c
+++ b/src/filter.c
@@ -875,7 +875,7 @@ void stream_list_free(git_vector *streams)
 	git_vector_free(streams);
 }
 
-#define STREAM_BUFSIZE 10240
+#define STREAM_BUFSIZE 65536
 
 int git_filter_list_stream_file(
 	git_filter_list *filters,


leoyanggit:
Should a system with small stack size be concerned?

whoisj:
I suppose a system with a small stack size should be concerned, but I've not seen an OS in a while that has an actual user-mode stack size limitation. I suppose the more esoteric systems could still have strict limitations.

Would you rather I guard this with `#ifdef GIT_WIN32` or another useful guard word?

leoyanggit:
How about dynamic allocating that buffer?

whoisj:
Dynamic allocations bring in issues with performance and memory compaction. If we allocate it dynamically, reuse it, and never deallocate it - I suppose there's very little difference but that's a large change than I'd anticipated. I'd like @carlosmn and/or @ethomson to comment here.

whoisj:
@leoyanggit any chance you can provide a concrete case where the stack size is a problem?

carlosmn:
Seems legit. You can do stuff like tell pthreads to allocate a teeny stack for each thread, but then you're shooting yourself in the foot with recursive code anyhow. If you have a system where you're extremely memory-constrained, you'd want to toggle a bunch more memory size assumptions, so I don't feel too bad saying you'd want to recompile for that.

whoisj:
Should we make these allocation assumptions in a common global file somewhere to make them easier to adjust as necessary?

carlosmn:
That certainly wouldn't hurt, they're currently spread out all over the place.

Issue #3105, opened 2015-05-07
Secure Transport: SecCopyErrorMessageString() not available on iOS


logancollins:
Recent changes in b7e1c81 to the stransport_stream.c function stransport_error() now use the SecCopyErrorMessageString() function defined in OS X's Security.framework. However, this function is not available on iOS.

PR #3106, opened 2015-05-07
Fix for issue #3105: usage of SecCopyErrorMessageString and libssh2 search path

logancollins:
This pull request includes a fix for issue #3105. Namely, it fixes usage of the SecCopyErrorMessageString() function on iOS (where it's not available) by ifdef'ing out the relevant block in favor of just logging the error code directly (as there's no *good* alternative on iOS to that function).

It also includes a fix to the header search path for libssh2 which allows the library to be found properly when using a self-built copy of libssh2 (which places its headers in "libssh2/".


carlosmn:
This removes any real error reporting as to why there was an issue, which would make it really hard to diagnose cert errors. It doesn't seem right that iOS would not provide any way to retrieve an error string due to a security error.

ethomson:
Can you not just use a different mechanism on iOS to pull the data out of an `OSStatus`?  For example:  `GetMacOSStatusCommentString` or `GetMacOSStatusErrorString`?

logancollins:
@ethomson Neither of those are available on iOS either. As far as I know, there is no public API to turn an OSStatus into a useful string on iOS.

carlosmn:
That's terrible, but ok, if the best we can do there is some number...

As for the libssh2 change, its files are in the include path directly, and as you can see making it look for `libssh2/libssh2.h` breaks the build. If you have not installed libssh2 on the system but are relying on a version you've built, you need to tell the build system, via `CFLAGS` or a CMake option when you run CMake on the project.

carlosmn:
I've cherry-picked the error-code change. The libssh2 includes are still a matter for the build system to tell cmake.

PR #3109, opened 2015-05-11
Use a diff for iteration in index_update_all and index_add_all

### LINKED ISSUE #2687: git_index_add_all() and git_index_update_all() are about 25X slower than git CLT
I have a test repo with 800+ files. Running `git_index_add_all()` or `git_index_update_all()` with only 1 file modified takes 0.5s on my MacBook Pro with SSD. Doing the same with `git -A` takes 0.02s.

Looking at the implementation it appears that both functions re-import all the files as blobs into the database, ignoring which files have actually changed.

carlosmn:
Re-adding every entry means we hash data we already have hashed, which proves quite a bit slower than it has to be. We can know whether we need to re-calculate hashes by iterating over the entries a diff gives us, rather than every entry in the index.

This should resolve #2687

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,9 @@ support for HTTPS connections insead of OpenSSL.
   remote functions now take these options or the callbacks instead of
   setting them beforehand.
 
+* The index now uses diffs for `add_all()` and `update_all()` which
+  gives it a speed boost and closer semantics to git.
+
 
 ### API additions
 
--- a/src/index.c
+++ b/src/index.c
@@ -2668,6 +2668,92 @@ enum {
 	INDEX_ACTION_REMOVE = 2,
 };
 
+struct foreach_diff_data {
+	git_index *index;
+	const git_pathspec *pathspec;
+	git_index_matched_path_cb cb;
+	void *payload;
+};
+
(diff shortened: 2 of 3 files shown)


ethomson:
This is great.  I'll look over this more deeply shortly, and I'm very excited to do so.

Until then, just a random thought that `add_all` is probably also not *correct*.  If you had poorly configured line endings, this would re-filter what's in your working directory and may end up putting stuff into the index.

So this is very nice.  Thanks for the nice monday morning surprise!

carlosmn:
Hm, that's an interesting point about the endings. Maybe we should take the crlf configuration into account in the preparation for the diff?

carlosmn:
Come to think of it, this new method is probably not correct for type changes, since it doesn't pass `INCLUE_TYPECHANGE` to diff.

ethomson:
Sorry, I was a little vague earlier, I meant the *current* `add_all` is probably dreadfully wrong.  I would expect that a diff-based approach would take advantage of the stat cache and not even investigate things that it knows couldn't have changed.  I was responding to your first commit's message that this is not a correctness change; I think that it is both correctness *and* perf.  :)

carlosmn:
It did later think you might have been talking about the current code. I've updated the commit message and added TYPECHANGE to the things for diff to tell us about.

PR #3111, opened 2015-05-11
Centralizing all IO buffer size values

whoisj:
Moving all IO buffer size values into common.h for easy modification for those who actually need to restrict the size of stack allocated buffers.

Defaulting the value of buffer sizes to 64KiB.


carlosmn:
Neat! Some of those buffers were pretty small, so this is likely to give us a speed boost in some areas as well.

PR #3113, opened 2015-05-11
WIP: Added test case for libgit2sharp:846.

jeffhostetler:
WIP.

Here is a test case for the problem described in 
https://github.com/libgit2/libgit2sharp/issues/846

Yet to be answered is how/whether it should be fixed.  That is, whether it just needs a better error message or whether we should try to auto-prune the (now obsolete) remote branch as we receive the new ones. 

As has been mentioned in 846, git.exe also complains about this situation (but with a better error message).

### DIFF
--- a/tests/refs/fetchconflict.c
+++ b/tests/refs/fetchconflict.c
@@ -1,0 +1,69 @@
+#include "clar_libgit2.h"
+
+#include "git2/clone.h"
+#include "clone.h"
+#include "fileops.h"
+#include "remote.h"
+
+/**
+ * Test to confirm problem fetching when the branch namespace
+ * on remote system has changed relative to the set of (now
+ * obsolete) branches listed in the local "remote" which cause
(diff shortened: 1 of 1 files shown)


carlosmn:
We must not delete remote-tracking branches (or really anything) without the user's consent. Pruning is not done by default to avoid data loss, which is quite central to Git (this would allow you to still find commits even if upstream removed a branch).

Pruning is also done after the update, so it would not save us anyway.

jeffhostetler:
Thanks for the clarification.  I'll close this PR WRT making any functional changes.

I'll keep my personal branch around for later if we want to consider changing the error message.

jamill:
Is there something tracking this for inclusion in the warning API work? I think lg2 will currently stop on first error here, which is undesirable.

jeffhostetler:
Do you have a link for the warning API changes?

PR #3116, opened 2015-05-12
Get rid of libssh2 embedding

carlosmn:
It was added as a workaround while the project had code to use WinCNG
but had not made a release with it. There is now a release of libssh2
with WinCNG support, so this option is redundant. Let's get rid of it
before people start liking it too much.

/cc @nulltoken you should be able to put libssh2.dll next to git2-X.dll and have it work, right?


Therzok:
libssh2 is still being distributed as source code, so we'd need some place to build it.

But a questions comes in mind:
How will cmake find out if we're linking against libssh2 or not?

carlosmn:
Right, libssh2 is souce code, just like libgit2, but it's not libgit2's place to build its dependencies. CMake currently finds out by asking pkg-config, as that's not a thing that works on Windows, we might need to add a module to search when on Windows, or you pass a prefix path and we build against it.

ethomson:
YAY.

PR #3117, opened 2015-05-12
index: make the entries have more accurate sizes

carlosmn:
While we are confident about the size of an int in architectures we're
likely to care about, the index format is defined by the exact size of
the fields. Use the definitions which show the exact width of the entry
fields.

As part of that, bring back 32-bit time and size fields, which currently
are 64 bits wide and can bring a false sense of security in how much
data they really store. Document that these fields are not to be taken
as authoritative.

---

Funnily enough, this doesn't throw any warnings during compilation, which means that we're already using these sizes internally.

### DIFF
--- a/include/git2/index.h
+++ b/include/git2/index.h
@@ -24,9 +24,9 @@ GIT_BEGIN_DECL
 
 /** Time structure used in a git index entry */
 typedef struct {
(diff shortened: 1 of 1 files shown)


ethomson:
This seems correct to me.

PR #3118, opened 2015-05-12
odb: make the writestream's size a git_off_t

### LINKED ISSUE #2936: src/blob.c write_file_stream() doesn't handle files larger than SIZE_MAX
While creating a local clone test for https://github.com/libgit2/libgit2/pull/2932
I ran into a problem where write_file_stream() claims to take a "git_off_t" 64 bit file_size,
but casts it to a 32 bit "size_t" or "ssize_t" for processing.

This causes us to return an error because (written != file_size).

```
>	libgit2_clar.exe!write_file_stream(git_oid * id, git_odb * odb, const char * path, __int64 file_size) Line 99	C
 	libgit2_clar.exe!git_blob__create_from_paths(git_oid * id, _stat64 * out_st, git_repository * repo, const char * content_path, const char * hint_path, unsigned short hint_mode, unsigned char try_load_filters) Line 209	C
 	libgit2_clar.exe!index_entry_init(git_index_entry * * entry_out, git_index * index, const char * rel_path) Line 790	C
 	libgit2_clar.exe!git_index_add_bypath(git_index * index, const char * path) Line 1092	C
 	libgit2_clar.exe!_stage_huge_file() Line 51	C
 	libgit2_clar.exe!test_clone_big__initialize() Line 87	C
 	libgit2_clar.exe!clar_run_test(const clar_func * test, const clar_func * initialize, const clar_func * cleanup) Line 196	C
 	libgit2_clar.exe!clar_run_suite(const clar_suite * suite, const char * filter) Line 257	C
 	libgit2_clar.exe!clar_parse_args(int argc, char * * argv) Line 317	C
 	libgit2_clar.exe!clar_test_run(...) Line 383	C
 	libgit2_clar.exe!main(int argc, char * * argv) Line 17	C
 	libgit2_clar.exe!__scrt_main_policy::get_app_type(void)	Unknown
 	libgit2_clar.exe!__ftol2()	Unknown
 	libgit2_clar.exe!__ftol2()	Unknown
 	libgit2_clar.exe!_mainCRTStartup()	Unknown
 	kernel32.dll!@BaseThreadInitThunk@12()	Unknown
 	ntdll.dll!__RtlUserThreadStart()	Unknown
 	ntdll.dll!__RtlUserThreadStart@8()	Unknown
```

carlosmn:
Restricting files to size_t is a silly limitation. The loose backend
writes to a file directly, so there is no issue in using 63 bits for the
size.

We still assume that the header is going to fit in 64 bytes, which does
mean quite a bit smaller files due to the run-length encoding, but it's
still a much larger size than you would want Git to handle.

---

This should be enough to fix #2936 though this return an error if no backends support streaming writes in, as we currently still try to keep everything in memory. We will have to use a spill buffer at some point, but that's longer-term and would only affect users with custom backends which do not support streaming.

### DIFF
--- a/src/blob.c
+++ b/src/blob.c
@@ -76,10 +76,11 @@ static int write_file_stream(
 	int fd, error;
 	char buffer[FILEIO_BUFSIZE];
 	git_odb_stream *stream = NULL;
-	ssize_t read_len = -1, written = 0;
+	ssize_t read_len = -1;
+	git_off_t written = 0;
 
 	if ((error = git_odb_open_wstream(
-			&stream, odb, (size_t)file_size, GIT_OBJ_BLOB)) < 0)
+			&stream, odb, file_size, GIT_OBJ_BLOB)) < 0)
 		return error;
 
 	if ((fd = git_futils_open_ro(path)) < 0) {
--- a/include/git2/odb.h
+++ b/include/git2/odb.h
@@ -247,7 +247,7 @@ GIT_EXTERN(int) git_odb_write(git_oid *out, git_odb *odb, const void *data, size
  * @param type type of the object that will be written
  * @return 0 if the stream was created; error code otherwise
  */
-GIT_EXTERN(int) git_odb_open_wstream(git_odb_stream **out, git_odb *db, size_t size, git_otype type);
+GIT_EXTERN(int) git_odb_open_wstream(git_odb_stream **out, git_odb *db, git_off_t size, git_otype type);
 
 /**
  * Write to an odb stream
--- a/include/git2/odb_backend.h
+++ b/include/git2/odb_backend.h
@@ -86,8 +86,8 @@ struct git_odb_stream {
 	unsigned int mode;
 	void *hash_ctx;
 
-	size_t declared_size;
-	size_t received_bytes;
+	git_off_t declared_size;
+	git_off_t received_bytes;
 
 	/**
 	 * Write at most `len` bytes into `buffer` and advance the stream.
--- a/include/git2/sys/odb_backend.h
+++ b/include/git2/sys/odb_backend.h
@@ -53,7 +53,7 @@ struct git_odb_backend {
 		git_odb_backend *, const git_oid *, const void *, size_t, git_otype);
 
 	int (* writestream)(
-		git_odb_stream **, git_odb_backend *, size_t, git_otype);
+		git_odb_stream **, git_odb_backend *, git_off_t, git_otype);
 
 	int (* readstream)(
 		git_odb_stream **, git_odb_backend *, const git_oid *);
--- a/src/odb.c
+++ b/src/odb.c
@@ -47,7 +47,7 @@ static git_cache *odb_cache(git_odb *odb)
 
 static int load_alternates(git_odb *odb, const char *objects_dir, int alternate_depth);
 
-int git_odb__format_object_header(char *hdr, size_t n, size_t obj_len, git_otype obj_type)
+int git_odb__format_object_header(char *hdr, size_t n, git_off_t obj_len, git_otype obj_type)
 {
 	const char *type_str = git_object_type2string(obj_type);
 	int len = p_snprintf(hdr, n, "%s %"PRIuZ, type_str, obj_len);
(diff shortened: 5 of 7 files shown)


carlosmn:
This lets us write into the odb, but we still can't read an object larger than `SIZE_MAX`, not that we would be able to read a `SIZE_MAX-1`-sized file anyway. This leaves us in the hilarious position which CVS also had, of being able to write data in that we can't read out; but streaming data out of the odb is non-trivial, so that's probably going to have to wait.

whoisj:
Just wondering, is `git_off_t` the best choice here? I agree with 64-bits and the changes look clean and tidy, but `git_off_t` is signed. Logically, length can never be negative so why not an unsigned type? Would a `git_size_t` make more sense?

src/blob.c
```diff
@@ -79,7 +79,7 @@ static int write_file_stream(
 	ssize_t read_len = -1, written = 0;
```
jeffhostetler: Need to make "written" a git_off_t rather than ssize_t so that the while loop will sum correctly.

carlosmn: Indeed. Changed.

carlosmn:
We use a signed `git_off_t` for file sizes because that's what `pread()`, `fseek()` and friends use, which probably betrays our unixy origin (we use `git_off_t` rather than `off_t` directly to make sure it's 64 bits regardless of the LFS support of some program we get compiled into).

It is certainly true that a file size won't ever be negative, but this file type does allow us to specify "no size" or "error" which an unsigned type wouldn't.

ethomson:
Hmm.  Well, to be pedantic, `read(3)` does *take* a `size_t` for its second argument...  Which seems reasonable, since it can't be negative.

Of course, this reasoning gets muddy since it *returns* an `ssize_t` (to account for error handling) which means that for all practical purposes, providing something larger than `SSIZE_MAX` to `nread` is pointless.

I, too, would slightly prefer to use unsigned types where things that are always positive.  It seems reasonable that we could have a `git_size_t` that is always 64 bit unsigned to go along with our `git_off_t`.

I've actually been thinking about this for a long time, but I've been quite lazy about doing anything about it.  We do have a handful of places where we cast `git_off_t`s to `size_t`s or vice-versa (sometimes passing them as args without an explicit cast) and while this is likely to not bite us, it is not technically correct (the best kind of correct).

Anyway.  I am a little on the fence about being too bullish about having an unsigned `git_size_t` because *what the shit, is someone really going to have a file that is bigger than `SSIZE_MAX` and expect us to do anything with it*?  Of course not.  I could also make some argument that it's nice not to have to worry about getting unsigned values from the user when you don't expect it.  Which might be a little bit true, but probably not.  Really it just *feels better* to me to keep unsigned things unsigned, in a way that I can't really articulate.  I fear that this exposes things about my personality that I generally try to keep hidden.

Either way, I very much agree that we should fix this shit.  Whichever way we decide to go with `git_size_t` vs `git_off_t`, we also need to knuckle down on the actually hard part of fixing up the internals of the library to correspond to this all the way through.

carlosmn:
> Hmm. Well, to be pedantic, read(3) does take a size_t for its second argument... Which seems reasonable, since it can't be negative.

To be pedantic, I did mention `pread(2)` rather than `read(2)` because that's the one that takes `off_t`. I did however mention `fseek(2)` when I meant `lseek(2)`. I did these functions because that's where `off_t` comes from. It is the offset inside the file. You still have the `size_t`/`ssize_t` combination for how much data you want to read, which I think reflects the distinction we have/should have in the library. You use `(git_)off_t` to show the size of a file, but you can still only hold up to `size_t` bytes in memory (theoretically anyhow). `off_t` is unsigned because `lseek(2)` needs to be able to say "five bytes from the end" which it encodes as `(off_t) -5`.

I'm not opposed to having a 64-bit unsigned type (though we could just use `uint64_t` for that) but I really don't like the name `git_size_t` because it would give a completely different meaning to `size_t` which we don't do anywhere and it's generally bad form. But I do think that the sign bit is probably the least interesting part of what we need to fix.

We do pass around these variables and assume that it will fit into both, assuming that nobody is going to have a file that won't fit in memory. This is certainly a bug (or many) and something we need to fix. But it's certainly not trivial to find these places or figure out what to do with them.

>is someone really going to have a file that is bigger than SSIZE_MAX and expect us to do anything with it?

DVDs on a 32-bit system spring to mind. While I don't recommend doing this, it's not something we should outright refuse to handle. People do sometimes have large amounts of textual data they'd like to version-control, which would fit more the git use-case (especially with compression and deltification).

ethomson:
Yes.  Sorry, I was awfully vague when I said "a file that is bigger than SSIZE_MAX"...  I was thinking about our always-64-bit size value, so 2^63.  I am definitely interested in being able to cope with files that are larger than 4.3 GB on a 32 bit system.

It's also worth pointing out (playing devil's advocate with myself) that `struct stat` always uses `off_t` as the size of a file.  (Though, frankly, I've always thought that silly for the same can't possibly-be-negative reasons I've been mentioning.)

ethomson:
OK.  So.  I'm convinced.

1. I'd rather have one type than two.
2. We will never, ever, never care about losing that bit and cry about not supporting files bigger than 9 exabytes.
3. It's not like having an unsigned type would save us from having to check for overflow.

So we might need to check some user input for negative values.  Big whoop.  Once this turns green I'm going to merge this.

whoisj:
Reasoning is sound, :+1: I like it.

PR #3119, opened 2015-05-12
Attributes: don't match files for folders

ethomson:
When a rule specifies a folder, eg:

    foo/

Ensure that a file of the same name is not matched.  (In the above example, a file named `foo`).

I added a relatively minor change to attr matching to handle this case, by simply testing whether we're looking at a file with the same name as the directory.  This also involved removing some historic path mangling that is unnecessary after we always use our own `fnmatch` function that supports folders.

However I did a little bit more refactoring in 90997e4 that allows us to do a little less copying and path mangling in general, which seems like a win.

Also, I also added some regression tests in adad518 for ensuring that attrs are actually matched in the directories that specify them.  (This was fixed in c6bf03b but additional tests made seemed to make sense in the context of this fix, especially with case insensitivity.)  And a quick fix in 30e629a where we skipped the error check.

### DIFF
--- a/tests/attr/ignore.c
+++ b/tests/attr/ignore.c
@@ -160,7 +160,7 @@ void test_attr_ignore__subdirectory_gitignore(void)
 
 	assert_is_ignored(true, "file1");
 	assert_is_ignored(true, "dir/file1");
-	assert_is_ignored(true, "dir/file2");  /* in ignored dir */
+	assert_is_ignored(true, "dir/file2/actual_file");  /* in ignored dir */
 	assert_is_ignored(false, "dir/file3");
 }
 
--- a/tests/attr/ignore.c
+++ b/tests/attr/ignore.c
@@ -207,7 +207,6 @@ void test_attr_ignore__gitignore_in_subdir(void)
 	assert_is_ignored(false, "dir1/file");
 	assert_is_ignored(false, "dir1/dir2/file");
 	assert_is_ignored(false, "dir1/dir2/dir3/file");
-	assert_is_ignored(false, "dir1/dir2/dir3/dir1");
 	assert_is_ignored(true,  "dir1/dir2/dir3/dir1/file");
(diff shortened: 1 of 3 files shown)


tests/attr/ignore.c
```diff
@@ -191,3 +191,52 @@ void test_attr_ignore__expand_tilde_to_homedir(void)
...
+	cl_must_pass(p_mkdir("attr/dir1", 0777));
+	cl_must_pass(p_mkdir("attr/dir1/dir2", 0777));
+	cl_must_pass(p_mkdir("attr/dir1/dir2/dir3", 0777));
+
+	cl_git_mkfile("attr/dir1/dir2/dir3/.gitignore", "dir1/\ndir1/subdir/");
+
+	assert_is_ignored(false, "dir1/file");
+	assert_is_ignored(false, "dir1/dir2/file");
+	assert_is_ignored(false, "dir1/dir2/dir3/file");
+	assert_is_ignored(false, "dir1/dir2/dir3/dir1");
```
carlosmn: This assertion seems the wrong way around. Doing this manually with git, it considers `dir1/dir2/dir3/dir1/` to be ignored itself, which is what I was expecting from reading that the first line in the ignore file is `dir1/`.

ethomson: In this case, `dir1/dir2/dir3/dir1` is *not* a directory (we look at the disk to determine if the basename is a directory or not), so a rule for `dir1/` should *not* match this.

Let me juggle some tests a bit to try to better illustrate this behavior.

carlosmn: A right, I seem to have forgotten the point of the PR when looking over that.

ethomson:
OK, I broke out the tests into some things that better illustrate what's going on.

carlosmn:
Thanks for making the tests more explicit. I like that we can remove the memcpy-based path.

PR #3121, opened 2015-05-13
Created clar test to create 4GB+1 byte file.

jeffhostetler:
Here is a test to create, stage, and commit a huge file.


ethomson:
Neat!  Thanks!

carlosmn:
Unfortunately, as this is against the stream-size branch, Travis won't build this, and there's no good way to re-assign a branch. Could you open a PR against mater, since the fix is in master?

whoisj:
Not sure about Linux flavors, but I believe msysgit won't handle repositories with files > 2GB in them.

Does git.git (without a filter like LFS) support files > 2GB? > 4 GB?

carlosmn:
On a 64-bit platform, there's no reason why it shouldn't. Many operations on a 32-bit machine will not succeed as git tends to assume that stuff fits in memory.

carlosmn:
In fact, I recall reports of OOM issues on 32-bit systems with somewhat-reasonable (~350MB) file sizes but unreasonable processor count during a push, where git would spawn 16 or 32 threads and they'd all be reading objects into memory and use up all of its address space and die. IIRC the machine in question had PAE so while the machine wasn't out of usable RAM, git became unable to allocate memory. All this to say that git isn't really the gold standard[0] for what we should be capable of handling.

[0] Alternatively, it *is* the gold standard, a system which nobody sensible has used in decades, fit for a different era

whoisj:
I get a fun error message with msysgit 1.9.5 on Windows 8 x64 when adding a 4.5 GB file:

C:\Src\test>git add test.blob
fatal: Cannot handle files this big

I'll try pushing the content with libgit2, then see what happens to msys after the fact.

linquize:
Can this test be optional? The developer may not have 4GB free space.

ethomson:
It is optional, and defaults to off.  You need to set the environment variable `GITTEST_INVASIVE_FS_SIZE` for it to run.  This is set on our CI builds but is not expected to be set on developer machines except during debugging these sorts of issues.

linquize:
ic, I found the variable `GITTEST_INVASIVE_FS_SIZE`

PR #3122, opened 2015-05-13
Better test support for non-system volumes on Windows

whoisj:
Many non-system volumes do not support 8.3 short names, disabling testing on volumes that do not support the feature.

Many non-systems volumes do not support links (aka reparse points and/or junctions), disabling testing on volumes that do not support the feature.

### DIFF
--- a/tests/checkout/icase.c
+++ b/tests/checkout/icase.c
@@ -100,7 +100,11 @@ static void assert_name_is(const char *expected)
 	char *actual;
 	size_t actual_len, expected_len, start;
 
+	if (!cl_sandbox_supports_8dot3())
+		clar__skip();
+
 	cl_assert(actual = get_filename(expected));
+	cl_assert(actual = test_realpath(expected));
 
 	expected_len = strlen(expected);
 	actual_len = strlen(actual);
--- a/tests/clar_libgit2.c
+++ b/tests/clar_libgit2.c
@@ -558,5 +558,36 @@ bool cl_sandbox_supports_8dot3(void)
 
 	return supported;
 }
+
+bool cl_sandbox_supports_reparse_points(void)
+{
+	unsigned int flags = 0;
+	char path[8];
+	char drive = 'A' + _getdrive() - 1;
+
+	sprintf(path, "%c:\\", drive);
+
+	GetVolumeInformation(path, NULL, 0, NULL, NULL, &flags, NULL, 0);
+
+	return flags & FILE_SUPPORTS_REPARSE_POINTS;
+}
 #endif
 
+bool cl_sandbox_supports_links(void)
+{
+#ifdef GIT_WIN32
+	unsigned int path_flags = 0;
+	unsigned int cl_flags = 0;
+	char path[8];
+	char drive = 'A' + _getdrive() - 1;
+
+	sprintf(path, "%c:\\", drive);
+
+	GetVolumeInformation(_cl_sandbox, NULL, 0, NULL, NULL, &cl_flags, NULL, 0);
+	GetVolumeInformation(path, NULL, 0, NULL, NULL, &path_flags, NULL, 0);
+
+	return cl_flags & path_flags & FILE_SUPPORTS_HARD_LINKS;
+#else
+	return true;
+#endif
+}
--- a/tests/repo/reservedname.c
+++ b/tests/repo/reservedname.c
@@ -13,6 +13,9 @@ void test_repo_reservedname__includes_shortname_on_win32(void)
 	git_buf *reserved;
 	size_t reserved_len;
 
+	if (!cl_sandbox_supports_8dot3())
+		clar__skip();
+
 	repo = cl_git_sandbox_init("nasty");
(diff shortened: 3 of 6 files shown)


tests/checkout/icase.c
```diff
@@ -100,7 +100,11 @@ static void assert_name_is(const char *expected)
 	char *actual;
 	size_t actual_len, expected_len, start;
 
+	if (!cl_sandbox_supports_8dot3())
+		clar__skip();
+
 	cl_assert(actual = get_filename(expected));
+	cl_assert(actual = test_realpath(expected));
```
ethomson: I think this snuck in during a rebase.

tests/checkout/icase.c
```diff
@@ -100,7 +100,11 @@ static void assert_name_is(const char *expected)
 	char *actual;
 	size_t actual_len, expected_len, start;
 
+	if (!cl_sandbox_supports_8dot3())
+		clar__skip();
```
ethomson: So, this is a little weird since we're not actually doing anything with 8.3 paths.  You're really trying to avoid running this test in places where `GetFinalPathNameByHandle` fails...  I think this problem is distinct from a lack of 8.3 paths, though, isn't it?

If so, we should consider refactoring the tests to skip when `GetFinalPathNameByHandle` is not available.

I say this with great reluctance, because I believe that your system is really broken if that's not supported and that this is a terrible bug in your filesystem that warrants replacement.  That's not hyperbole, I would really consider a different ramdisk driver if random function calls fail, because it makes running tests really difficult if you can't trust them.

Maybe instead of trying to decide whether or not your environment is flaky and we should skip the tests we should have an environment variable here.  That way there is no risk of us misinterpreting an actual test failure as one that would have come from an aforementioned flaky environment.

ethomson:
Couple minor notes, but I'm glad to see this.  Thanks.

tests/clar_libgit2.c
```diff
@@ -558,5 +558,36 @@ bool cl_sandbox_supports_8dot3(void)
...
+
+bool cl_sandbox_supports_reparse_points(void)
+{
+	unsigned int flags = 0;
+	char path[8];
+	char drive = 'A' + _getdrive() - 1;
+
+	sprintf(path, "%c:\\", drive);
+
+	GetVolumeInformation(path, NULL, 0, NULL, NULL, &flags, NULL, 0);
```
ethomson: This isn't, I suppose, super critical, but you can mount filesystems within the space of another filesystem.  So strictly speaking, I think that we should query the mount point for the clar sandbox path (`GetVolumePathName`?) and then pass *that* to `GetVolumeInformation`.  Otherwise you could mount a VHD at (say) `C:\Temp\Foo`, and use that as your clar path and we would incorrectly query the volume information for `C:\`.

I would suggest that this is not particularly useful *except* that this PR is about making our ability to run tests in wacky places (ramdisks, in your specific example) more useful.

tests/repo/reservedname.c
```diff
@@ -32,6 +35,9 @@ void test_repo_reservedname__includes_shortname_when_requested(void)
 	git_buf *reserved;
 	size_t reserved_len;
 
+	if (!cl_sandbox_supports_8dot3())
+		clar__skip();
```
ethomson: Actually, this really concerns me.  There's nothing here that should ever fail, ever never.  This test runs *on Unix*.

tests/clar_libgit2.c
whoisj: Agreed. You (@ethomson) and I had discussed something similar to this previously. It completely slipped my mind when I prepared this PR. I'll get it fixed up asap. Thanks.

tests/checkout/icase.c
whoisj: Yes, yes it did. I'll fix that.

tests/clar_libgit2.c
```diff
@@ -558,5 +558,36 @@ bool cl_sandbox_supports_8dot3(void)
...
+#ifdef GIT_WIN32
+	unsigned int path_flags = 0;
+	unsigned int cl_flags = 0;
+	char path[8];
+	char drive = 'A' + _getdrive() - 1;
+
+	sprintf(path, "%c:\\", drive);
+
+	GetVolumeInformation(_cl_sandbox, NULL, 0, NULL, NULL, &cl_flags, NULL, 0);
+	GetVolumeInformation(path, NULL, 0, NULL, NULL, &path_flags, NULL, 0);
```
ethomson: Maybe I misunderstood.  Don't we `chdir` into the `cl_sandbox` path before we start the tests?  If so, why do we need to test both?  Also, it would be nice to make sure that `GetVolumeInformation` succeeds so that we don't start blindly skipping tests.  (Just wrapping these calls in a `cl_win32_pass` should suffice.)

ethomson:
Cleaning up old issues that haven't been updated recently.  This is quite old, so I'm going to close this one; please feel free to reach out if there's more here.

PR #3123, opened 2015-05-13
Examples cleanup

urkud:
Reduce the number of warnings in `examples/`

A commit that touches the code outside `examples/` will be pushed as a separate pull request.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -6,6 +6,17 @@ ELSE()
 	TARGET_LINK_LIBRARIES(cgit2 git2 pthread)
 ENDIF()
 
+# Use -Wno-declaration-after-statement -Wunused-but-set-variable on global.c
+FOREACH(warn declaration-after-statement unused-but-set-variable)
+	SET(no_warn "-Wno-${warn}")
+	STRING(TOUPPER ${warn} _UPCASE)
+	SET(_var "IS_WNO_${_UPCASE}_SUPPORTED")
+	CHECK_C_COMPILER_FLAG(${no_warn} ${_var})
+	IF(${_var})
+		SET_PROPERTY(SOURCE general.c APPEND_STRING PROPERTY COMPILE_FLAGS " ${no_warn}")
+	ENDIF()
+ENDFOREACH()
+
 FILE(GLOB SRC_EXAMPLE_APPS RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} *.c)
 FOREACH(src_app ${SRC_EXAMPLE_APPS})
 	STRING(REPLACE ".c" "" app_name ${src_app})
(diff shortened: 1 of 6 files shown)


examples/CMakeLists.txt
```diff
@@ -6,6 +6,17 @@ ELSE()
...
+FOREACH(warn declaration-after-statement unused-but-set-variable)
+	SET(no_warn "-Wno-${warn}")
+	STRING(TOUPPER ${warn} _UPCASE)
+	SET(_var "IS_WNO_${_UPCASE}_SUPPORTED")
+	CHECK_C_COMPILER_FLAG(${no_warn} ${_var})
+	IF(${_var})
+		SET_PROPERTY(SOURCE general.c APPEND_STRING PROPERTY COMPILE_FLAGS " ${no_warn}")
+	ENDIF()
+ENDFOREACH()
+
```
carlosmn: This feels like *way* too much effort go to in order to handle `general.c`. We should either fix the warnings or give up on the whole file.

examples/log.c
```diff
@@ -340,14 +340,14 @@ static void print_time(const git_time *intime, const char *prefix)
...
-	int i, count;
+	unsigned int i, count;
 	const git_signature *sig;
 	const char *scan, *eol;
 
 	git_oid_tostr(buf, sizeof(buf), git_commit_id(commit));
 	printf("commit %s\n", buf);
 
-	if ((count = (int)git_commit_parentcount(commit)) > 1) {
+	if ((count = git_commit_parentcount(commit)) > 1) {
```
carlosmn: Why change both to `unsigned int` if you then have to cast here as well? Wouldn't it be better if `count` were the right type to start with?

examples/CMakeLists.txt
urkud: What do you mean by "give up"? `general.c` includes a series of loosely coupled code snippets. It makes no sense to move all declarations to the top of `main()`. The other ways to silent these warnings include:
- use IF(CMAKE_COMPILER_IS_GNUCC) instead of checking for compiler flag;
- use `#pragma GCC ignored -Wdeclaration-after-statement` inside `general.c`.

examples/log.c
urkud: After this patch, `count` **does** have the right type `unsigned int` (i.e., the type that coincides with the return type of `git_commit_parentcount`).
What cast are you talking about?

carlosmn: It looks like I misread the diff.

examples/CMakeLists.txt
carlosmn: Declarations don't have to be at the top of a function, they have to be at the top of a block. You can create blocks anywhere. But these are code snippets that aren't meant to be run anyhow, they're just there for a bit of demonstration.

ethomson:
Let's please not go through too much churn to update the CMake definition.  This is - after all - just the examples which should be focused first and foremost on illustrating how to use the library, not necessarily being perfect, warning free code.

The rest does seem reasonable, though.  If you wanted to drop the CMake changes, that would probably be good.

Thanks!

carlosmn:
Closing for inactivity.

Issue #2268, opened 2014-04-14
git_index_add_bypath does not honor core.precomposeunicode


jspahrsummers:
Quoth @arrbee:

> Libgit2 implements unicode composition for the places where it reads file entries (i.e. it performs the `readdir` to get the data from the OS). If you are passing decomposed unicode data into libgit2 APIs (e.g. you are calling `git_index_add_bypath` with a path you got from the filesystem that has decomposed unicode data), it won't compose the data for you - it will just blindly add it to the index.

swisspol:
The way I solved this issue for my app http://gitup.co is simply by replacing all the calls I was doing to `-fileSystemRepresentation` to convert an NSString* to a char* to pass to libgit2 by these wrapper functions:
```objc
// We can't use -[NSString fileSystemRepresentation] as it returns decomposed UTF8 while everything in Git is composed UTF8
// (unless the "core.precomposeUnicode" configuration option is false which shouldn't happen on OS X)
const char* GCGitPathFromFileSystemPath(NSString* string) {
  return string.UTF8String;
}

// We shouldn't use -[NSFileManager stringWithFileSystemRepresentation:length:] for the same reason as above
NSString* GCFileSystemPathFromGitPath(const char* string) {
  return [NSString stringWithUTF8String:string];
}
```

I'm leaving my PR around just in case as I think that might be useful API, but it turns out it wasn't even needed!

swisspol:
I should be clear that the above assumes `core.precomposeunicode` is `true` for the Git repo, but it should really be the case for all repos created on OS X.

PR #3124, opened 2015-05-13
Added git_repository_precompose_path()

swisspol:
This exposes the internal APIs used by libgit2 to handle path precomposition on OS X.

I realized this was needed as I hit the same roadblock as described in #2268.


swisspol:
Feedback welcome on where the API should go and what the name should be!

PR #3125, opened 2015-05-14
Pass current remote to git_push_negotiation callback

swisspol:
This makes it quite easier to implement the Git 'pre-push' hook.


carlosmn:
If we're going to add the sender, we should do it for all of them. But we haven't added this since we prefer the callers use closures to capture whatever information they need for the work of the callbacks.

swisspol:
OK, I'll close this PR then.

PR #3129, opened 2015-05-14
Bring Early EOF message in line with the other two

Therzok:
Early EOF is used 2 more times and uses a capital E in the beginning. Let's have them the same.

### DIFF
--- a/src/transports/smart_protocol.c
+++ b/src/transports/smart_protocol.c
@@ -51,7 +51,7 @@ int git_smart__store_refs(transport_smart *t, int flushes)
 				return recvd;
 
 			if (recvd == 0 && !flush) {
-				giterr_set(GITERR_NET, "Early EOF");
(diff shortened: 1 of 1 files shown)


Therzok:
Sauce: 
https://github.com/libgit2/libgit2/blob/c5c5cdb106d012d132475d9156923857f8d302fc/src/transports/smart_protocol.c#L54
https://github.com/libgit2/libgit2/blob/c5c5cdb106d012d132475d9156923857f8d302fc/src/transports/smart_protocol.c#L772

carlosmn:
We should in fact do the opposite. Error messages should start with a lower-case letter as they are typically prefixed by the application's own "sorry" message. Starting with a lower-case is the project's style, though we're pretty lenient when we write or accept code. But if you're going to bring them all in line, it would be better if the other two were made lower-case.

Therzok:
Sure!

ethomson:
:+1:

PR #3131, opened 2015-05-14
Add `const` qualifier

urkud:
This fixes a warning in `examples/describe.c` without breaking the main build.

OTOH, I'm not sure if this is an API-compatible change.

### DIFF
--- a/include/git2/describe.h
+++ b/include/git2/describe.h
@@ -94,7 +94,7 @@ typedef struct {
 	 * If the workdir is dirty and this is set, this string will
 	 * be appended to the description string.
 	 */
-	char *dirty_suffix;
+	const char *dirty_suffix;
 } git_describe_format_options;
 
 #define GIT_DESCRIBE_FORMAT_OPTIONS_VERSION 1


urkud:
I have no idea what breaks Travis CI. Tests pass on my computer (with clang).

carlosmn:
Why change the data structure rather than the example?

urkud:
Is there any reason for `git_describe_format_options.dirty_suffix` to be mutable? It seems that it is normally used like `fmt_opts.dirty_suffix = "string"`. Doesn't C compiler make a copy of "string" to assign it to a `char *` variable?

ethomson:
Well, no, the compiler doesn't make a copy.  When you're doing something like assigning a constant string literal - for example:

    fmt_opts.dirty_suffix = "string";

then what the compiler does is it compiles that string into your program's data segment, which is read-only  When you pass that string to libgit2, we'll get a pointer into your program's data segment, there won't be a duplicate copy made.

Most compilers will actually take this a step further and find duplicate instances of string literals, and set them all to a single pointer to reduce duplication in the data segment.  So if you did:

    const char *foo = "hello world";
    const char *bar = "hello world";

Then `foo` and `bar` would probably point at the same hunk of memory.  This is common, but I don't think that the language makes a guarantee of this.

Apologies for the long aside.  That might not have even answered the question that you were asking.    In any case, the libgit2 library *does* make a copy of all this data (which may have been the question that you were asking).  We do this, of course, because we don't want to point to memory that the caller owns.

Imagine that the caller was using some hunk of memory that they had allocated, in the heap.  And then wanted to re-use this data.  They would be free to do something like:

     char *foo = malloc(42);
     strcpy(foo, "Hello, world!");

     fmt_opts.dirty_suffix = foo;
     git_describe_commit();

     strcpy(foo, "Different suffix");
     git_describe_commit();

When libgit2 is given the `options`, we will make a copy of them for our use, so that we don't start making assumptions about the memory that the caller allocated, that doesn't belong to us.

Making these `const` is in keeping with what we do elsewhere in the library with options structures, and seems like a nice change so that consumers have confidence that we're not going to munge their data.

urkud:
I understand that a compiler doesn't create a copy when I write `const char *s = "str"`, but what about `char *s = "str"`?
As for your example, it is possible with `char * dirty_suffix` but not with `const char *dirty_suffix`.

P.S.: Am I right that this `const` changes ABI, so the same commit should bump SOVERSION?
P.P.S.: I failed to find where libgit2 creates a copy of `dirty_suffix`.

ethomson:
> I understand that a compiler doesn't create a copy when I write const char *s = "str", but what about char *s = "str"?

No, there is no copy in either case.  You are pointing to the same hunk of memory regardless.  Your compiler may issue you a warning for the second one.  If you try to *modify* that string, something bad might happen.  Your program may crash with an exception.  (Or it may not, this is undefined behavior in the specification.)

You're right, I was mistaken, and we do not make a copy of the string.  If we ever use this structure beyond a single function call, such that a caller could modify the string between two invocations, then we would definitely need to do that.  (Though I don't see any occurrence of that at a glance.)

ethomson:
Regarding you question about `SOVERSION`:  `const`ness is an idea presented by the compiler...  so this actually *doesn't* break ABI compatibility.

Despite that, we do not bump `SOVERSION` for every pull request.  Instead, we increment it at every *release* of the library.  So at our next release (which we should probably do soon!) we will increment that to cover all ABI breaking changes, of which there have been oh so many.

carlosmn:
The constness shouldn't depend on whether one example happens to have a const somewhere. As we only use it on a single function call as options, we do not make any copies of it, and we do not know whether the caller has the string as a const or not. The reasons for making it a const is about what guarantees we want to make to the caller and how to express them. Silencing this one warning can be achieved just as well by explicit casing on the example.

We take the structure as const to assure everyone that we're only going to read from it. The constness isn't transitive IIRC, so the compiler wouldn't warn if we try to overwrite the buffer. This one be one good reason to make it const, and say that the user shouldn't be storing their "real" pointer in this structure if they allocate it dynamically.

Issue #2797, opened 2015-01-04
Local clone fails when cloning between different mount points, same device


russell:
Local clone is linking when it should be copying between different mount points to the same device.

```
$ mkdir ../tmp
$ sudo mount -o bind . ../tmp
$ TMPDIR=../tmp ./libgit2_clar -sclone::nonetwork::local_absolute_path
Loaded 275 suites: 
Started
F

  1) Failure:
clone::nonetwork::local_absolute_path [/home/russell/projects/libgit2/tests/clone/nonetwork.c:91]
  Function call failed: (git_clone(&g_repo, local_src, "./foo", &g_options))
  error -1 - git_path_direach callback returned -1


$ TMPDIR=../tmp strace -e link ./libgit2_clar -sclone::nonetwork::local_absolute_path
Loaded 275 suites: 
Started
link("/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/config.lock", "/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/config") = -1 EEXIST (File exists)
link("/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/config.lock", "/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/config") = -1 EEXIST (File exists)
link("/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/config.lock", "/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/config") = -1 EEXIST (File exists)
link("/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/config.lock", "/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/config") = -1 EEXIST (File exists)
link("/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/HEAD.lock", "/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/HEAD") = 0
link("/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/config.lock", "/home/russell/projects/libgit2/tmp/libgit2_tests_ps5Rdq/foo/.git/config") = -1 EEXIST (File exists)
[...]

PR #3132, opened 2015-05-15
Provide error messages for git_path_direach operations

### LINKED ISSUE #2965: Cryptic error on disk full
When the disk is full, instead of returning an error saying that, we get a rather non-descriptive error message about a particular function saying there was some error. We should probably let the callback set the error or read it from `errno` so we can actually tell the user something useful.

```
Traceback (most recent call last):
  File "test_fmc.py", line 31, in setUp
    bare=True, checkout_branch="foo_7.0_integration")
  File "/Users/carlos/foo/tools/venv/lib/python2.7/site-packages/pygit2-0.22.0-py2.7-macosx-10.9-x86_64.egg/pygit2/__init__.py", line 286, in clone_repository
    check_error(err)
  File "/Users/carlos/foo/tools/venv/lib/python2.7/site-packages/pygit2-0.22.0-py2.7-macosx-10.9-x86_64.egg/pygit2/errors.py", line 64, in check_error
    raise GitError(message)
GitError: git_path_direach callback returned -1
```

carlosmn:
This function currently always overwrites the underlying error message with a rather useless one. Instead let's let the callback set their own error message and only set our own when they fail to do so. 

This sets a more informative error message for #2797 and fixes #2965

### DIFF
--- a/src/fileops.c
+++ b/src/fileops.c
@@ -703,6 +703,9 @@ static int cp_by_fd(int ifd, int ofd, bool close_fd_when_done)
 		error = (int)len;
 	}
 
+	if (error < 0)
+		giterr_set(GITERR_OS, "write error while copying file");
+
 	if (close_fd_when_done) {
 		p_close(ifd);
 		p_close(ofd);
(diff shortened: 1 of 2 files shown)


src/fileops.c
```diff
@@ -703,6 +703,9 @@ static int cp_by_fd(int ifd, int ofd, bool close_fd_when_done)
 		error = (int)len;
 	}
 
+	if (error < 0)
+		giterr_set(GITERR_OS, "write error while coyping file");
```
ethomson: typo here

PR #3137, opened 2015-05-18
Improve server error reporting

carlosmn:
Git's reporting for errors on the server is all over the place. Sometimes you'll get an error via the protocol, sometimes that same process will send the error over stderr. This PR catches those different ways of reporting errors an make it easier for tools to report such errors.

Crucially, if a repository we're trying to access over ssh is not there, the server process will send the error message over stderr rather than send an error pkt (which it does for some other errors, such as three-level repo names on GitHub). If we don't get anything over stdout, we report whatever we get from stderr as the error message.

Adding `GIT_EEOF` comes from the fact that it's a somewhat unique failure mode, and we can expect users of the library to be able to respond to it; e.g. git itself shows the hint about checking the url to be correct and whether the user has access rights. This is different from some internal error or an error which would come from arguments given by the caller.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,11 @@ support for HTTPS connections insead of OpenSSL.
 * `git_stash_pop()` will apply a stashed state (like `git_stash_apply()`)
   but will remove the stashed state after a successful application.
 
+* A new error code `GIT_EEOF` indicates an early EOF from the
+  server. This typically indicates an error with the URL or
+  configuration of the server, and tools can use this to show messages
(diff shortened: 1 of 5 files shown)


Therzok:
:heart:

ethomson:
Sorry; while our network tests were busy timing out against BitBucket, I merged some stuff that caused a merge conflict here.  If you would be so kind as to update this, I'll merge it promptly.

carlosmn:
Rebased on master.

carlosmn:
Merged myself before we change the changelog again

PR #3138, opened 2015-05-18
Maintenance updates for v0.22

carlosmn:
Here's a few things we can backport for v0.22.3. There's a bunch of useful fixes in master which don't quite fit into v0.22 anymore as they rely on newer code, so the list isn't as large as I wanted, but these are still good fixes if anybody can't update to v0.23 immediately.

### DIFF
--- a/include/git2.h
+++ b/include/git2.h
@@ -58,6 +58,7 @@
 #include "git2/submodule.h"
 #include "git2/tag.h"
 #include "git2/transport.h"
+#include "git2/transaction.h"
 #include "git2/tree.h"
 #include "git2/types.h"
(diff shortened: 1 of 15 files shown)


ethomson:
I thought that we had added `maint/*` to the list of branches that Travis and Appveyor would build.  I guess I was mistaken, but perhaps we should do that.

carlosmn:
We still only have mater and development in the list. We should indeed add the maint branches.

tkelman:
Can I nominate https://github.com/libgit2/libgit2/commit/a944c6cc40273b45922fa22949fe6b12ff668889 ?

PR #3139, opened 2015-05-18
Include conflicts when diffing

ethomson:
When diffing a tree with the index, or the index with the working directory, it is useful to know whether you have conflicts.  At present, we only look at zero-stage index entries when iterating the index, which has the effect of pretending as if conflicts don't exist at all.

This has generally been okay (even if bizarre), but by ignoring conflicts, a conflicted file might appear to be ignored (because it does not exist in the index).  This could cause a consumer to fail to add a conflict resolution when [using a mechanism that filters out ignored files during add](https://github.com/libgit2/libgit2sharp/blob/vNext/LibGit2Sharp/Repository.cs#L1640).

Add a new diff delta type, `CONFLICTED`, to return when a conflict exists in the index.  (Do not provide any of the `old_file` or `new_file` data during this case, since determining which ancestor / ours / theirs side to include would be impossible.)

This also updates status to include a new `CONFLICTED` entry.  This does *not* provide any more detailed information about the conflict (eg, "both added" or "both modified"), though I intend to add this later as an addition to the `git_status_entry` structure.  This was an important first-step in this.

I don't think that the lack of that means that we should wait to include this, though; this is a critical bug for consumers (like above) where resolved conflicts may not get added correctly.

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -766,43 +766,45 @@ static int maybe_modified(
 
 	/* if one side is a conflict, mark the whole delta as conflicted */
 	if (git_index_entry_is_conflict(oitem) ||
-		git_index_entry_is_conflict(nitem))
+			git_index_entry_is_conflict(nitem)) {
 		status = GIT_DELTA_CONFLICTED;
 
 	/* support "assume unchanged" (poorly, b/c we still stat everything) */
-	else if ((oitem->flags & GIT_IDXENTRY_VALID) != 0)
+	} else if ((oitem->flags & GIT_IDXENTRY_VALID) != 0) {
 		status = GIT_DELTA_UNMODIFIED;
 
 	/* support "skip worktree" index bit */
-	else if ((oitem->flags_extended & GIT_IDXENTRY_SKIP_WORKTREE) != 0)
+	} else if ((oitem->flags_extended & GIT_IDXENTRY_SKIP_WORKTREE) != 0) {
 		status = GIT_DELTA_UNMODIFIED;
 
 	/* if basic type of file changed, then split into delete and add */
-	else if (GIT_MODE_TYPE(omode) != GIT_MODE_TYPE(nmode)) {
-		if (DIFF_FLAG_IS_SET(diff, GIT_DIFF_INCLUDE_TYPECHANGE))
+	} else if (GIT_MODE_TYPE(omode) != GIT_MODE_TYPE(nmode)) {
+		if (DIFF_FLAG_IS_SET(diff, GIT_DIFF_INCLUDE_TYPECHANGE)) {
 			status = GIT_DELTA_TYPECHANGE;
+		}
+
 		else if (nmode == GIT_FILEMODE_UNREADABLE) {
 			if (!(error = diff_delta__from_one(diff, GIT_DELTA_DELETED, oitem, NULL)))
 				error = diff_delta__from_one(diff, GIT_DELTA_UNREADABLE, NULL, nitem);
 			return error;
 		}
+
 		else {
 			if (!(error = diff_delta__from_one(diff, GIT_DELTA_DELETED, oitem, NULL)))
 				error = diff_delta__from_one(diff, GIT_DELTA_ADDED, NULL, nitem);
 			return error;
 		}
-	}
 
 	/* if oids and modes match (and are valid), then file is unmodified */
-	else if (git_oid_equal(&oitem->id, &nitem->id) &&
+	} else if (git_oid_equal(&oitem->id, &nitem->id) &&
 			 omode == nmode &&
-			 !git_oid_iszero(&oitem->id))
+			 !git_oid_iszero(&oitem->id)) {
 		status = GIT_DELTA_UNMODIFIED;
 
 	/* if we have an unknown OID and a workdir iterator, then check some
 	 * circumstances that can accelerate things or need special handling
 	 */
-	else if (git_oid_iszero(&nitem->id) && new_is_workdir) {
+	} else if (git_oid_iszero(&nitem->id) && new_is_workdir) {
 		bool use_ctime = ((diff->diffcaps & GIT_DIFFCAPS_TRUST_CTIME) != 0);
 		bool use_nanos = ((diff->diffcaps & GIT_DIFFCAPS_TRUST_NANOSECS) != 0);
 
@@ -833,12 +835,12 @@ static int maybe_modified(
 			status = GIT_DELTA_MODIFIED;
 			modified_uncertain = true;
 		}
-	}
 
 	/* if mode is GITLINK and submodules are ignored, then skip */
-	else if (S_ISGITLINK(nmode) &&
-			 DIFF_FLAG_IS_SET(diff, GIT_DIFF_IGNORE_SUBMODULES))
+	} else if (S_ISGITLINK(nmode) &&
+			 DIFF_FLAG_IS_SET(diff, GIT_DIFF_IGNORE_SUBMODULES)) {
 		status = GIT_DELTA_UNMODIFIED;
+	}
 
 	/* if we got here and decided that the files are modified, but we
 	 * haven't calculated the OID of the new item, then calculate it now
@@ -847,6 +849,7 @@ static int maybe_modified(
 		const git_oid *update_check =
 			DIFF_FLAG_IS_SET(diff, GIT_DIFF_UPDATE_INDEX) && omode == nmode ?
 			&oitem->id : NULL;
+
 		if ((error = git_diff__oid_for_entry(
 				&noid, diff, nitem, update_check)) < 0)
 			return error;
--- a/src/diff.c
+++ b/src/diff.c
@@ -127,10 +127,12 @@ static int diff_delta__from_one(
 	if (has_old) {
 		delta->old_file.mode = entry->mode;
 		delta->old_file.size = entry->file_size;
+		delta->old_file.flags |= GIT_DIFF_FLAG_EXISTS;
 		git_oid_cpy(&delta->old_file.id, &entry->id);
 	} else /* ADDED, IGNORED, UNTRACKED */ {
 		delta->new_file.mode = entry->mode;
 		delta->new_file.size = entry->file_size;
+		delta->new_file.flags |= GIT_DIFF_FLAG_EXISTS;
 		git_oid_cpy(&delta->new_file.id, &entry->id);
 	}
 
@@ -184,13 +186,16 @@ static int diff_delta__from_two(
 		delta->old_file.size = old_entry->file_size;
 		delta->old_file.mode = old_mode;
 		git_oid_cpy(&delta->old_file.id, old_id);
-		delta->old_file.flags |= GIT_DIFF_FLAG_VALID_ID;
+		delta->old_file.flags |= GIT_DIFF_FLAG_VALID_ID |
+			GIT_DIFF_FLAG_EXISTS;
 	}
 
 	if (!git_index_entry_is_conflict(new_entry)) {
 		git_oid_cpy(&delta->new_file.id, new_id);
 		delta->new_file.size = new_entry->file_size;
 		delta->new_file.mode = new_mode;
+		delta->old_file.flags |= GIT_DIFF_FLAG_EXISTS;
+		delta->new_file.flags |= GIT_DIFF_FLAG_EXISTS;
 
 		if (!git_oid_iszero(&new_entry->id))
 			delta->new_file.flags |= GIT_DIFF_FLAG_VALID_ID;
--- a/src/diff.c
+++ b/src/diff.c
@@ -77,10 +77,23 @@ static int diff_insert_delta(
 static int diff_delta__from_one(
 	git_diff *diff,
 	git_delta_t status,
-	const git_index_entry *entry)
+	const git_index_entry *oitem,
+	const git_index_entry *nitem)
 {
+	const git_index_entry *entry = nitem;
+	bool has_old = false;
 	git_diff_delta *delta;
 	const char *matched_pathspec;
+
+	assert((oitem != NULL) ^ (nitem != NULL));
+
+	if (oitem) {
+		entry = oitem;
+		has_old = true;
+	}
+
+	if (DIFF_FLAG_IS_SET(diff, GIT_DIFF_REVERSE))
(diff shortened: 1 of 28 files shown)


arrbee:
:heart_eyes:

Therzok:
I love where this is going.

src/index.c
```diff
@@ -2722,7 +2748,7 @@ static int index_apply_to_all(
 
 				error = git_index_remove_bypath(index, path.ptr);
 
-				if (!error) /* back up foreach if we removed this */
+				if (!error) /* back up foconreach if we removed this */
```
carlosmn: Is this a typo? Sounds like a type of bread.

tests/index/conflicts.c
```diff
@@ -96,6 +103,53 @@ void test_index_conflicts__add_fixes_incorrect_stage(void)
...
+
+	staged.path = "test-one.txt";
+	staged.mode = 0100644;
+	git_oid_fromstr(&staged.id, TEST_STAGED_OID);
+	cl_git_pass(git_index_add(repo_index, &staged));
+	cl_assert(git_index_entrycount(repo_index) == 9);
+
+	ancestor_entry.path = "test-one.txt";
+	ancestor_entry.mode = 0100644;
+	ancestor_entry.flags |= (3 << GIT_IDXENTRY_STAGESHIFT);
```
carlosmn: We do have `GIT_INDEX_STAGE_SET()` to make sure it's done quite right. I see that we do it manually in other tests in this area, but maybe we should start switching?

tests/index/conflicts.c
```diff
@@ -96,6 +103,53 @@ void test_index_conflicts__add_fixes_incorrect_stage(void)
...
+	their_entry.path = "test-one.txt";
+	their_entry.mode = 0100644;
+	their_entry.flags |= (2 << GIT_IDXENTRY_STAGESHIFT);
+	git_oid_fromstr(&their_entry.id, TEST_THEIR_OID);
+
+	cl_git_pass(git_index_conflict_add(repo_index, &ancestor_entry, &our_entry, &their_entry));
+
+	cl_assert(git_index_entrycount(repo_index) == 11);
+
+	cl_git_pass(git_index_conflict_get(&conflict_entry[0], &conflict_entry[1], &conflict_entry[2], repo_index, "test-one.txt"));
```
carlosmn: Should we also test explicitly that we can't do a get of a zero-stage entry?

src/index.c
```diff
@@ -1169,6 +1169,9 @@ int git_index_remove_bypath(git_index *index, const char *path)
 		ret != GIT_ENOTFOUND))
 		return ret;
 
+	if (ret == GIT_ENOTFOUND)
+		giterr_clear();
+
```
carlosmn: Do we need the `if` here? We already return if we find a different error code, so ret here could only be 0 or `ENOTFOUND`, and we should clear the error for both, since we're returning 0 in both of these cases.

src/index.c
```diff
@@ -1310,6 +1313,29 @@ int git_index_conflict_add(git_index *index,
 		(ret = index_entry_dup(&entries[2], INDEX_OWNER(index), their_entry)) < 0)
 		goto on_error;
 
+	/* Validate entries */
+	for (i = 0; i < 3; i++) {
+		if (entries[i] && !valid_filemode(entries[i]->mode)) {
+			giterr_set(GITERR_INDEX, "invalid filemode");
+			return -1;
```
carlosmn: It might be nice to include the stage which failed validation.

src/diff.c
```diff
@@ -731,8 +764,12 @@ static int maybe_modified(
 		new_is_workdir)
 		nmode = (nmode & ~MODE_BITS_MASK) | (omode & MODE_BITS_MASK);
 
+	/* if one side is a conflict, mark the whole delta as conflicted */
+	if (git_index_entry_stage(oitem) || git_index_entry_stage(nitem))
+		status = GIT_DELTA_CONFLICTED;
```
carlosmn: Can we get explicit `> 0` comparisons here? Relying on zero as false seems iffy here. It seems to say "if it has a stage" like with pointesr rather than "if the stage is higher than 0" (not so say I wouldn't have done this myself, but when someone else writes it... :)

src/diff.c
```diff
@@ -731,8 +764,12 @@ static int maybe_modified(
...
+
 	/* support "assume unchanged" (poorly, b/c we still stat everything) */
-	if ((oitem->flags & GIT_IDXENTRY_VALID) != 0)
+	else if ((oitem->flags & GIT_IDXENTRY_VALID) != 0)
```
carlosmn: Could we get some braces for the if which extends over multiple lines, including a comment. The compiler might consider this as a single line, but human eyes get confused.

src/diff.c
```diff
@@ -857,6 +895,79 @@ static bool entry_is_prefixed(
...
+
+		cmp = (iterator->flags & GIT_ITERATOR_IGNORE_CASE) ?
+			strcasecmp(prev_entry->path, (*entry)->path) :
+			strcmp(prev_entry->path, (*entry)->path);
+
+		if (cmp)
+			break;
+	}
+
+	return error == GIT_ITEROVER ? (*entry = NULL, 0) : error;
```
carlosmn: While I'm a fan of the comma operator, I'm a bigger fan of the convention we have of doing something like

```C
if (error == GIT_ITEROVER) {
    *entry = NULL;
    error = 0;
}
````

to reset the error state after the iteration is over.

File: src/diff.c [hunk 9]
```diff
@@ -1004,14 +1118,20 @@ static int handle_unmatched_new_item(
...
 	git_diff *diff, diff_in_progress *info)
 {
-	int error = diff_delta__from_one(diff, GIT_DELTA_DELETED, info->oitem);
-	if (error != 0)
+	git_delta_t delta_type = GIT_DELTA_DELETED;
+	int error;
+
+	/* update delta_type if this item is conflicted */
+	if (git_index_entry_stage(info->oitem))
+		delta_type = GIT_DELTA_CONFLICTED;
```
carlosmn: Like above, this reads like "does it have a stage" rather than "is it conflicted". Maybe we could have a bool method which returns whether it represents a conflict.

tests/diff/index.c
```diff
@@ -163,3 +163,76 @@ void test_diff_index__checks_options_version(void)
 	git_tree_free(a);
 }
 
+static void do_conflicted_diff(diff_expects *exp, unsigned long flags)
+{
+	/* grabbed a couple of commit oids from the history of the attr repo */
+	const char *a_commit = "26a125ee1bf"; /* the current HEAD */
```
carlosmn: A couple is traditionally two, no? ;)

tests/diff/index.c
```diff
@@ -163,3 +163,76 @@ void test_diff_index__checks_options_version(void)
...
+	opts.flags |= flags;
+
+	memset(exp, 0, sizeof(diff_expects));
+
+	cl_git_pass(git_repository_index(&index, g_repo));
+
+	ancestor.path = ours.path = theirs.path = "staged_changes";
+	ancestor.mode = ours.mode = theirs.mode = 0100644;
+
+	cl_git_pass(git_index_conflict_add(index, &ancestor, &ours, &theirs));
```
carlosmn: This would make all entries have an of {0}*40, no? I suppose it doesn't quite matter here, but it feels a bit odd.

src/diff.c
```diff
@@ -147,38 +160,42 @@ static int diff_delta__from_two(
...
+		delta->old_file.size = old_entry->file_size;
+		delta->old_file.mode = old_mode;
+		git_oid_cpy(&delta->old_file.id, old_id);
+		delta->old_file.flags |= GIT_DIFF_FLAG_VALID_ID;
+	}
 
-	git_oid_cpy(&delta->new_file.id, &new_entry->id);
-	delta->new_file.size = new_entry->file_size;
-	delta->new_file.mode = new_mode;
+	if (!git_index_entry_stage(new_entry)) {
```
carlosmn: Here too, it reads more like a lack of stage, rather than asking about stage 0.

carlosmn:
I've done a lot of nitpicking on readability of intention, but other than that, it's pretty good.

tests/index/conflicts.c
ethomson: Added.

src/index.c
ethomson: Fixed.

src/index.c
ethomson: It certainly wouldn't be incorrect to `giterr_clear()` in all cases, but this does avoid setting several thread-local variables.

(Yes, I realize this is a *seriously* minor issue, but some vendors have produced code that's inferior around TLS in the past.)

src/index.c
ethomson: Yes it would!  Added.

ethomson:
> Maybe we could have a bool method which returns whether it represents a conflict.

Yep, I like that.  I introduced `git_index_entry_is_conflict`, and moved a bunch of stuff to use it.

I'm not sure I know exactly what you were asking regarding the readability of conditionals in `maybe_modified`.  Let me know if that's not what you were asking for.  (It might be nice to take a bigger readability hammer to that, but I regret that I do not have time at the moment, so I will keep to the little hammer.)

carlosmn:
The reason I was asking for braces is that we have the if, then a blank line, then a comment, and *then* we get to the if. I was looking for something like

```C
	/* if one side is a conflict, mark the whole delta as conflicted */
	if (git_index_entry_is_conflict(oitem) ||
			git_index_entry_is_conflict(nitem)) {
		status = GIT_DELTA_CONFLICTED;
 	/* support "assume unchanged" (poorly, b/c we still stat everything) */
	} else if ((oitem->flags & GIT_IDXENTRY_VALID) != 0) {
 		status = GIT_DELTA_UNMODIFIED;
	}
```

with the closing brace on the same line as its else, otherwise it looks like the if is done at the brace close (or without the brace, after the `status =` statement and the else comes completely as a surprise.

ethomson:
See if that's more readable to you...

Also, please don't merge this.  After merging #3109, I will include some changes to `index_add_all` to handle the case described above.

carlosmn:
Yeah, thanks. I had forgotten that this was what the other ifs were doing it as well, but I'm glad we got it changed, it always irked me when reading it.

ethomson:
Okay!  I think this is all patched up.  As mentioned before, now that our `git_index_add_all` (correctly) uses a diff mechanism to determine what needs to be added, we are faced with the same problem described above.  Thus, I added some changes here to deal with that.

I introduced the a flag on a diff's delta file, `GIT_DIFF_FLAG_EXISTS` to show whether a file exists on a particular side of the diff.  This is because we were previously using a setup like:

    if (GIT_DELTA_DELETED)
        // the file is not in the workdir
    else
        // the file is in the workdir

However now we can have a type (`GIT_DELTA_CONFLICTED`) that doesn't elucidate the status of the working directory file.  Thus we put this information on the delta file directly.

carlosmn:
Great. CHANGELOG is missing the addition of `git_index_entry_is_conflict` but otherwise, looks like we should get it merged.

ethomson:
Added `git_index_entry_is_conflict` to the CHANGELOG, I think this should be ready to go if the CI builds ever complete.

swisspol:
After updating to latest libgit2, I found a few issues related to this PR:

- There's a bug in `handle_unmatched_new_item()` where even if the item is a conflict, it can still be reported as `GIT_DELTA_ADDED` instead of `GIT_DELTA_CONFLICTED` (see https://github.com/libgit2/libgit2/blob/master/src/diff.c#L1089).
- There might other similar issues in `maybe_modified()`, `handle_unmatched_old_item()` and `handle_unmatched_new_item()`.
- If the status is `GIT_DELTA_CONFLICTED`, are `old_file` and `new_file` guaranteed to be defined in `git_diff_delta`? This is unclear looking at the code.

ethomson:
Can you provide repro steps?

If you have `GIT_DELTA_CONFLICTED`, you are only guaranteed to have one of `old_file` or `new_file`.

swisspol:
Oops, sorry. Here you go: I have functional tests that test the 4 possible merge conflict cases on a single file:
- Both modified: status is `GIT_DELTA_CONFLICTED` as expected
- Both added: status is `GIT_DELTA_CONFLICTED` as expected
- Deleted by then: status is `GIT_DELTA_CONFLICTED` as expected
- **Deleted by us: status is `GIT_DELTA_ADDED`**

For the last case, the equivalent CLI status from `git status --ignored --porcelain`:
```
UD hello_world.txt
```

swisspol:
> If you have GIT_DELTA_CONFLICTED, you are only guaranteed to have one of old_file or new_file.

Turns out you said above a different thing:

> Do not provide any of the old_file or new_file data during this case, since determining which ancestor / ours / theirs side to include would be impossible.

So what should the final behavior be here?

After testing it looks like both `old_file` and `new_file` are always defined, which doesn't match either of the above.

ethomson:
What data is populated?  What type of conflict?

swisspol:
I just have a dummy text file containing a single line (with ending newline). It's on 2 branches and I modified it on either branch, then merge, in order to achieve the 4 cases described above. Is that what you mean?

ethomson:
No, I mean what is in `old_file` and `new_file` when you say that they're populated?  You should have the path, since that's obvious.  You should not have the OID since we wouldn't know which OID to give you between the three sides of the conflict.

swisspol:
It depends, this is what I observe for old OID and new OID:
- both modified:
 - diff index / head: OID/OID
 - diff index / workdir: null/null
- deleted by us:
 - diff index / head: (n/a because of bug above since status is `GIT_DELTA_ADDED`)
 - diff index / workdir: null/null
- deleted by them:
 - diff index / head: OID/OID
 - diff index / workdir: null/null
- both added:
 - diff index / head: OID/OID
 - diff index / workdir: null/null

Furthermore when the OID is non-null, it's actually the same value for both old and new.

Not sure if this behavior makes sense...

swisspol:
I edited my comment above as I was printing the old OID twice instead of old & new! The data is all good now.

swisspol:
I've perused the affected areas in the code and I *think* this patch is enough to address the bug: https://github.com/git-up/libgit2/commit/53f05c1c471a30a69a5bf2a6684fe713b2a87051

PR #3142, opened 2015-05-20
Call the proper openssl API to be able to work with SNI servers

### LINKED ISSUE #3091: libgit2 does not call the proper openssl API to implement SNI
[SNI](http://en.wikipedia.org/wiki/Server_Name_Indication) is needed to access https servers that are located on endpoints which host multiple virtual servers.

Because ```SSL_set_tlsext_host_name(st->ssl, st->socket->host)``` is not called I get a:

```
SSL error: error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure (-1)
```

When trying to clone a repository from such a server.

conet:
This is an attempt to fix #3091. If you know a better place for that function call please let me know.

### DIFF
--- a/src/openssl_stream.c
+++ b/src/openssl_stream.c
@@ -243,10 +243,9 @@ int openssl_connect(git_stream *stream)
 		return ssl_set_error(st->ssl, ret);
 	}
 
-#ifdef SSL_CTRL_SET_TLSEXT_HOSTNAME
-	// specify the host
-        SSL_set_tlsext_host_name(st->ssl, st->socket->host);
-#endif
+	/* specify the host in case SNI is needed */
+	SSL_set_tlsext_host_name(st->ssl, st->socket->host);
+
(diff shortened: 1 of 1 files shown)


ethomson:
This seems like an important thing to support.  I don't love silently including or not including support; I think that it would be very confusing to somebody building libgit2 to not have SNI support when they expected it.

Should we drop the ifdef and *always* call the set hostname function?  It seems like OpenSSL has had SNI support for many (8? more?) years...  I am perfectly happy suggesting that you need a newer version of OpenSSL than that to use libgit2.

If somebody does want to use an older OpenSSL that doesn't support SNI, maybe we should make them explicitly set something during the build.

Does that seem reasonable?

carlosmn:
Yeah, I don't feel bad requiring a version of OpenSSL with this function, especially since we don't link against it on OS X.

carlosmn:
Thanks, I squashed and merged this manually.

PR #3144, opened 2015-05-20
ignore: clear the error when matching a pattern negation

### LINKED ISSUE #3143: gitignore not respected in certain situations?
1. `mkdir test; cd test; git init`
2. Use this `.gitignore`:

```
*.com

tags
!tags/

.buildpath
```
3. `touch .buildpath`
4. `git status`. Note that `.buildpath` is properly ignored by git.git.
5. Get status using libgit2. Note that `.buildpath` is not ignored.

carlosmn:
When we discover that we want to keep a negative rule, make sure to
clear the error variable, as it we otherwise return whatever was left by
the previous loop iteration.

This fixes #3143


joshaber:
Awesome turn around time! :metal:

PR #3146, opened 2015-05-20
index_add_all: include untracked files in new subdirs

ethomson:
By default, diff will collapse a new directory that contains only untracked files into a single delta for the whole directory.  (This is akin to `git status` on the command line's behavior.)

Make `git_index_add_all` ask for the individual untracked files, so it can add each of them.


carlosmn:
This looks like the flag I meant to add instead of the ignored dirs one.

ethomson:
Cool; in that case I updated this to drop that flag, since I'm thinking it's not helpful one way or another.

Issue #3136, opened 2015-05-18
Diff progress callback


hackhaslam:
I'm trying to get progress and cancel long diffs. I want it mostly for diffing with the workdir on slow filesystems. I see that there's a notify callback but it's only invoked when a file is added to the diff. Is there already another way to do this? Would a patch to add a progress callback be acceptable?

carlosmn:
Fixed via #3147

PR #3147, opened 2015-05-21
Add diff progress callback.

hackhaslam:
This is meant to address issue #3136. Doesn't add tests or documentation. I wanted to see if there's any interest first.

### DIFF
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -347,6 +347,22 @@ typedef int (*git_diff_notify_cb)(
 	void *payload);
 
 /**
+ * Diff progress callback.
+ *
+ * Called before each file comparison.
+ *
+ * @param diff_so_far The diff being generated.
+ * @param old_path The path to the old file or NULL.
+ * @param new_path The path to the new file or NULL.
+ * @return Non-zero to abort the diff.
+ */
+typedef int (*git_diff_progress_cb)(
+	const git_diff *diff_so_far,
+	const char *old_path,
+	const char *new_path,
+	void *payload);
+
+/**
  * Structure describing options about how the diff should be executed.
  *
  * Setting all values of the structure to zero will yield the default
@@ -366,8 +382,10 @@ typedef int (*git_diff_notify_cb)(
  * - `max_size` is a file size (in bytes) above which a blob will be marked
  *   as binary automatically; pass a negative value to disable.
  * - `notify_cb` is an optional callback function, notifying the consumer of
- *   which files are being examined as the diff is generated
- * - `notify_payload` is the payload data to pass to the `notify_cb` function
+ *   changes to the diff as new deltas are added.
+ * - `progress_cb` is an optional callback function, notifying the consumer of
+ *   which files are being examined as the diff is generated.
+ * - `payload` is the payload to pass to the callback functions.
  * - `ignore_submodules` overrides the submodule ignore setting for all
  *   submodules in the diff.
  */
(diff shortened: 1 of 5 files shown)


ethomson:
I'm not clear on why the existing `notify_cb` isn't adequate for this task?  Is there something we can change that would make it more suitable for the task, instead of adding a second callback that is almost the same?

hackhaslam:
The documentation on notify_cb is somewhat contradictory. It says that it "is an optional callback function, notifying the consumer of which files are being examined as the diff is generated", but in reality it is only called when a new delta is added to the diff. So if I call git_diff_tree_to_workdir and nothing is changed in the wordir, it could take 30 seconds but the callback is never invoked. I want what the documentation claims, notification of which files are being examined and the ability to abort.

carlosmn:
Why not make the existing notification behave as it's documented, instead of creating a new callback and leaving the wrong documentation in place?

hackhaslam:
I meant to update documentation in a follow-up commit. I'm fine with changing the behavior of the existing callback but the interface will be completely different so it will break any existing users.

carlosmn:
If someone is relying on buggy behaviour, then a fix is going to break it, but it'd be good to know which side is the likliest to be wrong. I.e. was the documented behaviour ever accurate? If the code has diverged, then that's the likely candidate for fixing, otherwise the docs are the one which never represented what it did.

ethomson:
The `git_diff_notify_cb` itself is documented as:

> The callback will be called for each file, just before the `git_delta_t` gets inserted into the diff.

So the documentation between the struct member and the `git_diff_notify_cb` definition does not match.  The documentation for the `git_diff_notify_cb` appears to be accurate, so it seems that actually having *both* callbacks does seem quite reasonable.

hackhaslam:
Okay, thanks. I fixed the failing test, made the existing documentation consistent, added some new documentation, and added a trivial test.

include/git2/diff.h
```diff
@@ -347,6 +347,21 @@ typedef int (*git_diff_notify_cb)(
...
 /**
+ * Diff progress callback.
+ *
+ * Called before each file comparison.
+ *
+ * - 'old_path' is the path to the old file or NULL
+ * - 'new_path' is the path to the new file or NULL
+ *
+ * Returning a non-zero value will abort the diff.
+ */
```
carlosmn: The documentation has a specific format so it can be parsed by tools. The parameters have `@param`, return has `@return` etc.

include/git2/diff.h
```diff
@@ -381,6 +399,8 @@ typedef struct {
 	git_strarray       pathspec;     /**< defaults to include all paths */
 	git_diff_notify_cb notify_cb;
 	void              *notify_payload;
+	git_diff_progress_cb progress_cb;
+	void                *progress_payload;
```
carlosmn: I don't think we should have different payloads for each callback. I'd much rather see `notify_payload` be renamed to `payload`.

hackhaslam:
Thanks for the review. I addressed those comments. I also added a diff parameter to the callback.

carlosmn:
The changes in the structures need to have an entry in CHANGELOG. Please also squash together both commits, as they represent a single change.

hackhaslam:
Okay, thanks. I added the CHANGELOG changes, squashed the commits, and rebased on the tip of master.

hackhaslam:
I rebased this and adjusted the CHANGELOG entries for the new release.

ethomson:
Many apologies about letting this linger so long that it became out of date - that led to a merge conflict on the changelog.  And then more letting *that* problem linger.  So sorry.

I manually merged this.  Thanks for the great PR and for your patience.

hackhaslam:
Thanks!

Issue #2702, opened 2014-11-13
Pessimistic crypto locking


carlosmn:
This would depend on #2678 so we have the OpenSSL stuff in a single file that's easier to read.

I forget who mentioned that we might want to simply lock around each OpenSSL operation, and I'm coming to the conclusion that we might have to rely on that. Since we cannot know for sure that OpenSSL or libgcrypt have the right locking set up, we should default to locking around each function call we make into OpenSSL or libssh2 and only trust that they are thread-safe if we are told explicitly so.

I would propose to have a pair of functions, `git_openssl_threadsafe()` and `git_libssh2_threadsafe()` which would set some global variable to tell us not to lock. The `git_openssl_set_locking()` function would then know that it can call `git_openssl_threadsafe()` so we don't attempt to lock, but it wouldn't be able to call `git_libssh2_threadsafe()` unless we can query it at runtime and see that it's also linked to OpenSSL.

Locking around IO functions is terrible, but if you wanted to throw a bunch of threads at fetching data, over ssh, it's going to go wrong, unless the application programmer has taken care to set up the locking, which is not going to be the case in practice.

PR #3150, opened 2015-05-22
Lock around encrypted I/O

carlosmn:
OpenSSL and often whatever libssh2 is using require their own set-up for
concurrent operations to be safe. This means that by default, using both
of these libraries in a threaded context is unsafe.

Lock by default, and allow the user to tell us that they've set up
threading for the underlying libraries.

---

I'm quite conflicted about this approach. It does make use safer by default. E.g. even if you run `git_openssl_set_locking()` on Debian, you're still not safe to perform multiple ssh operations in parallel, but it feels like a huge hack around something we shouldn't be handling at all in the library.

More rationale is at #2702

### DIFF
--- a/src/stream.h
+++ b/src/stream.h
@@ -50,4 +50,14 @@ GIT_INLINE(void) git_stream_free(git_stream *st)
 	st->free(st);
 }
 
+/* Macros for locking around OpenSSL and libssh2; it assumes you have a static 'should_lock' variable */
+#define LOCK do {						   \
+	if (should_lock && git_mutex_lock(&git__io_mutex) < 0) {   \
+		giterr_set(GITERR_NET, "failed to lock IO mutex"); \
+		return -1;					   \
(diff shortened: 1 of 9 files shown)


src/stream.h
```diff
@@ -50,4 +50,14 @@ GIT_INLINE(void) git_stream_free(git_stream *st)
 	st->free(st);
 }
 
+/* Macros for locking around OpenSSL and libssh2; it assumes you have a static 'should_lock' variable */
+#define LOCK do {						   \
```
Therzok: Why not define it as `LOCK(x)` and use the `x` in the check?

carlosmn: Because it's not meant to be a generally-useful macro, but one for this exact use.

carlosmn:
I'm going to close this plus the issue, this is really not something we should be handling in the library.

PR #3151, opened 2015-05-22
Thin packs: don't 

ethomson:
When fixing a thin pack, we should not try to open an index for the packfile that we are currently writing.  (There is guaranteed to be no index.)  The second commit is really not a big win (we will probably have the base cached), but might prevent some unnecessary mucking around in the object cache.

### DIFF
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -325,6 +325,13 @@ on_error:
 	return -1;
 }
 
+GIT_INLINE(bool) has_entry(git_indexer *idx, git_oid *id)
+{
+	khiter_t k;
+	k = kh_get(oid, idx->pack->idx_cache, id);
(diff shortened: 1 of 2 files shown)


carlosmn:
It seems that instead of adding meaning to the index version field, we could return `GIT_EPASSTHROUGH` for ref-delta objects if we did not find the entry in the cooperative cache. This cache is only there to be used by the indexer, and there should be a 1-to-1 correlation between having `GIT_PACK_INDEX_NONE` and the coop cache set.

carlosmn:
As for the second commit, while I'm not sure if it's going to increase the speed a lot, it does look like it will help with correctness. It looks like if two ref-deltas use the same base object, we would try to inject it into the pack twice, which is not something we should ever do.

ethomson:
That seems reasonable; updated.

PR #3155, opened 2015-05-24
cred_helpers: Add 'const' qualifiers to git_cred_userpass_payload

mgorny:
Make both `username` & `password` in `git_cred_userpass_payload` `const`. The values are not altered anywhere, and the extra qualifier allows clients to assign `const` values there.

I've checked most of the bindings listed on homepage and none of them seem to use `git_cred_userpass_payload`. The tests pass. Also confirmed that gitg & subsurface (two random packages using libgit2) don't rely on the lack of `const`. So it should be safe to change.


carlosmn:
Thanks, the constness is more about the promise that we make not to change them rather than something we do because we happen not to change them, though the explanation in the commit is fine.

PR #2989, opened 2015-03-17
Add support to read ssh keys from memory.

calavera:
Libssh2 got the ability to read keys from memory recently.
Although it's only supported when combined with OpenSSL for now. :unamused:

https://github.com/libssh2/libssh2/commit/18cfec8336e88ab7df8a4427b803b9a177053674

This adds a new credentials type to be able to authenticate using keys in memory.
I've used a flag to enable all this, since you'll have to compile libssh2 from master to have access to it.

I'm not really sure how to add tests, but I pretty promise that it works. :grin:

PR #3157, opened 2015-05-24
Support getting SSH keys from memory, pt. 2

mgorny:
This is PR #2989 with small fixes. The API is consistent now, and CMake checks for the function availability.

What remains to do are tests. However, I would like to get some more advice here. Right now, I see that two files are declaring SSH key auth callback:
* online/clone.c
* online/push.c

For clone.c, the callback is hidden a few functions deep. So if I were to introduce an alternative variant using memory credentials, I'd have to pass an additional boolean through all those functions.

For push.c, I don't see the callback being used anywhere…

In either case, we won't be able to test it properly on Travis until libssh2 1.6.0 is released and available there.


carlosmn:
I don't know what you mean by the callback being hidden. The tests that check for this kind of thing, e.g. `test_online_clone__ssh_auth_methods`, only need to set the cred callback function to their own which creates the credentials you needed for the test.

mgorny:
Ah, ok. I was looking for the test that actually used SSH key but I guess I can create my own.

mgorny:
Done and tested locally! It reads the key files supplied in the env and passes them to `git_cred_ssh_key_memory_new()`. The test is skipped when in-memory keys are not supported.

tests/online/clone.c
```diff
@@ -519,6 +519,74 @@ void test_online_clone__ssh_cert(void)
...
+	const char *remote_user = cl_getenv("GITTEST_REMOTE_USER");
+	const char *privkey = cl_getenv("GITTEST_REMOTE_SSH_KEY");
+
+#ifndef GIT_SSH_MEMORY_CREDENTIALS
+	clar__skip();
+#endif
+	if (!remote_url || !remote_user || !privkey || strncmp(remote_url, "ssh://", 5) != 0)
+		clar__skip();
+
+	g_options.remote_callbacks.credentials = ssh_memory_cred_cb;
```
carlosmn: This should be more like `g_options.fetch_options.callbacks.credentials`. It looks like the remote changes got merged after this branched off, so rebasing on top of master would be good to get the tests to use the new API.

mgorny:
Rebased and updated now. Are there any more concerns?

carlosmn:
:+1: from me, @ethomson anything to add?

PR #3160, opened 2015-05-26
Fix ident replacement to match Git behavior

b4n:
Git inserts a space after the SHA1 (as of 2.1.4 at least), so do the same for consistency.

This is particularly useful when diffing a filtered blob against a file checked out by Git, as without this the `$Id$` line will be different.

---
Git behavior:
```console
$ git --version 
git version 2.1.4
$ git init .
Initialized empty Git repository in .../.git/
$ echo '$Id$' > foo
$ echo 'foo ident' > .gitattributes
$ git add foo .gitattributes
$ git commit -m "init"
[master (root-commit) 3684894] init
 2 files changed, 2 insertions(+)
 create mode 100644 .gitattributes
 create mode 100644 foo
$ rm -f foo
$ git checkout .
$ cat foo
$Id: 055c8729cdcc372500a08db659c045e16c4409fb $
```

### DIFF
--- a/src/ident.c
+++ b/src/ident.c
@@ -56,7 +56,7 @@ static int ident_insert_id(
 		return GIT_PASSTHROUGH;
 
 	need_size = (size_t)(id_start - from->ptr) +
-		5 /* "$Id: " */ + GIT_OID_HEXSZ + 1 /* "$" */ +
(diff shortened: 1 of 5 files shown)


Therzok:
Nice catch!

carlosmn:
Thanks

PR #3161, opened 2015-05-28
Changed README to use new OCaml bindings to git

fxfactorial:
As title says.


ethomson:
Cool!  Excited to see a new OCaml binding!

PR #2932, opened 2015-02-27
Fix crash in git_clone on extremely large repos

jeffhostetler:
A crash was observed on Windows when cloning a very large repo. In this instance, the received packfile was nearly 6Gb.

The problem was in sloppy casting of the "size" field from src/indexer.c:append_to_pack() to src/win32/posix_w32.c:p_ftruncate() to _chsize_s().
Where the file position was being truncated and then sign-extended.

Also fixed the return value from _chsize_s() which differs from _chsize().

PR #3162, opened 2015-05-28
Created clar test to create 4GB+1 byte file.

jeffhostetler:
Recreate pull request #3121 using master as a baseline.

This commit adds a test to create, stage, and commit a huge file.

### DIFF
--- a/tests/hugefiles/create.c
+++ b/tests/hugefiles/create.c
@@ -1,0 +1,144 @@
+/**
+ * This test tries to create a huge file and add it.
+ * The purpose is to ensure various APIs accept 64-bit
+ * file sizes/offsets.
+ */
+#include "clar_libgit2.h"
+
+#include "git2/clone.h"
+#include "remote.h"
+#include "fileops.h"
+#include "repository.h"
+
+#define FILE_A      "file_a.zeroes"
+#define REPO_ROOT   "./repo1"
+
+static git_repository *g_repo = NULL;
+static git_oid g_id_initial_commit;
+
+/**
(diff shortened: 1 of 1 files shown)


carlosmn:
Cool. Can we not name functions with leading underscores? We can just have static functions with normal names.

jeffhostetler:
Sorry, old habit.  Fixed.

carlosmn:
As a sanity check, I think we should check that we get back the expected id for the blob, as otherwise we won't be able to detect if we broke reading past `SIZE_MAX`.

jeffhostetler:
Yeah, I was wondering about how far to take this test, given that huge files are a bit of an edge case. When I started it was just additional testing for #2936 and #2932.  I could see adding a bunch of asserts WRT id's and checkouts.  I was also wondering about another test case using data that won't compress since it looks like 4Gb+1 zeroes compresses to about 18Mb. And maybe one using write() rather than ftruncate() so that mingw32 won't have to be skipped.

I'll look into adding something along these lines.
Thanks

carlosmn:
The main reason for asking for an id check is to answer the question "are we actually putting the whole blob into the odb?" We cannot currently *extract* this object where a 4GB+1-sized file would cause an issue so we would expect that to return OOM (though I suppose we might test that, maybe).

As for the mingw32 issue, it occurs to me that we don't need to have the file exist at all. You can use `git_blob_create_fromchunks()` to insert as many gigs into the odb as you want. Though to be honest, it's essentially equivalent to using `write()`, it's just that it's the library doing it for you.

jeffhostetler:
Closing this in favor of #3197.

PR #3163, opened 2015-05-28
Rename GIT_EMERGECONFLICT to GIT_ECONFLICT

ethomson:
We do not error on "merge conflicts"; on the contrary, merge conflicts
are a normal part of merging.  We only error on "checkout conflicts",
where a change exists in the index or the working directory that would
otherwise be overwritten by performing the checkout.

This *may* happen during merge (after the production of the new index
that we're going to checkout) but it could happen during any checkout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -229,6 +229,9 @@ support for HTTPS connections insead of OpenSSL.
   index. Specifically, time and file size are 32 bits intead of 64, as
   these values are truncated.
 
+* `GIT_EMERGECONFLICT` is now `GIT_ECONFLICT`, which more accurately
+  describes the nature of the error.
+
 v0.22
 ------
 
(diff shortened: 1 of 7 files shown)


ethomson:
This has bothered me for a long-ass time, I was always shy about fixing it because I didn't want to break API compatibility.  We've been doing a lot of that lately, though, so what the heck, let's name this correctly lest people assume that our name [indicates something that it doesn't](https://github.com/libgit2/libgit2sharp/pull/1059).

carlosmn:
We do error out if we are asked to perform a checkout when there are conflicts in the index so I think the error doc should still mention them, but I do like the change.

Can we get some CHANGELOG love?

ethomson:
You and your slavish adherence to the CHANGELOG.  Telling users when we've broken them and all that.

jamill:
> Can we get some CHANGELOG love?

:heart_eyes: 

Thank you for keeping this in mind as part of the code review. I actually find the changelog quite useful, so this attention is appreciated.

ethomson:
> Thank you for keeping this in mind as part of the code review. I actually find the changelog quite useful, so this attention is appreciated.

:+1: 

I've been bad about this lately.  (Sorry.)  So thanks be to @carlosmn for keeping me honest.

(I updated this PR by the way.)

carlosmn:
If I don't keep it in mind now, one of us is going to have to put it into the chanelog before the release anyway, so :)

PR #3165, opened 2015-05-29
Downcase

ethomson:
Our `git__strcasecmp` is brain damaged when confronted with things that are outside the realm of strict 7 bit clean ASCII.  When confronted with the leading byte of a multibyte UTF-8 sequence, we'll promptly treat it as a negative value, which will almost certainly screw up any computation of whether another character is bigger or smaller.

While I was here, I moved our usage(s) of `tolower` over to be `git__tolower` because some stupid-ass C libraries (not to name any names) have done terrible things to `tolower` to lock around the locale, lest you change your locale in the middle of...  well, I don't know what stupid test where somebody was changing their locale frequently led to this completely reckless decision, as if `tolower` should be taking - I don't know - single-byte non-ASCII character sets into account anyway.

Since we are explicitly UTF-8, there is no case where `tolower` of something that isn't ASCII `A` - `Z` should ever do anything but return the original byte back, so we can avoid that whole poorly thought out mess and avoid taking a lock that should never be taken in the first place.

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -275,6 +275,15 @@ extern char *git__strsep(char **end, const char *sep);
 extern void git__strntolower(char *str, size_t len);
 extern void git__strtolower(char *str);
 
+#ifdef GIT_WIN32
+GIT_INLINE(int) git__tolower(int c)
+{
+	return (c >= 'A' && c <= 'Z') ? (c + 32) : c;
+}
+#else
+# define git__tolower(a) tolower(a)
+#endif
+
 GIT_INLINE(const char *) git__next_line(const char *s)
 {
 	while (*s && *s != '\n') s++;
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -156,7 +156,7 @@ int git_config_file_normalize_section(char *start, char *end)
 		if (end && scan >= end)
 			break;
 		if (isalnum(*scan))
-			*scan = (char)tolower(*scan);
+			*scan = (char)git__tolower(*scan);
 		else if (*scan != '-' || scan == start)
 			return GIT_EINVALIDSPEC;
 	}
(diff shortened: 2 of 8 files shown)


ethomson:
/cc @jeffhostetler 
who may have some additional information on the mis-efficiency of `tolower` on this particular C library implementation.

src/util.h
```diff
@@ -275,6 +275,15 @@ extern char *git__strsep(char **end, const char *sep);
 extern void git__strntolower(char *str, size_t len);
 extern void git__strtolower(char *str);
 
+#ifdef GIT_WIN32
+GIT_INLINE(int) git__tolower(int c)
+{
+	return (c >= 65 && c <= 90) ? (c + 32) : c;
```
carlosmn: I'd rather see the actual letters here, this test is presumably `(c >= 'A' && c <= 'Z')` but I'd have to consult my ASCII chart.

carlosmn:
>Since we are explicitly UTF-8, there is no case where tolower of something that isn't ASCII A - Z should ever do anything but return the original byte back

What about pre-composed forms of diactrics? I would expect `git__strcasecmp("Í", "í") to return 0. Does our function handle this? I don't see this in the tests.

ethomson:
You can precompose it, but it's still two bytes of UTF-8, no?  Isn't Í `U+00CD` (LATIN CAPITAL LETTER I WITH ACUTE, or UTF-8 `0xc3 0x8d`)?

I would expect `git__strcasecmp` to return non-zero there, and I would expect `strcasecmp` to do the same on all platforms.  I'll add a test for this.

src/util.h
ethomson: I agree, that's much smarter.  I also changed this to compute the diff based on `'a' - 'A'` which I'm not sure that I love.  See what you think.

carlosmn: I'm rather "meh" on it. I'm not sure it conveys more information than the '+ 32'.

ethomson: I am too.  Fixed.

linquize:
Windows 's `tolower()` is **too** smart!

PR #3167, opened 2015-05-30
Introduce blame options FOLLOW_EXACT_RENAMES flag.

Therzok:


Therzok:
Using the example for blame and switching the flags to appropriate values, I get the following results on blaming [this file](https://github.com/mono/monodevelop/blob/master/main/src/addins/VersionControl/MonoDevelop.VersionControl.Git/MonoDevelop.VersionControl.Git/GitRepository.cs):

7.77s with GIT_BLAME_DONT_FOLLOW_RENAMES
7.96s with GIT_BLAME_FOLLOW_EXACT_RENAMES
26.82s with no flag set.

This takes me to the idea that blame is generally really slow, and the heuristics for similarity are even slower themselves. This should work as a nice workaround for slowness, 8s is way more bearable than 27s.

Therzok:
cc @carlosmn @ethomson

carlosmn:
:+1:

ethomson:
I'm curious why we don't just take a `diff_opts` structure here?  It looks like `git blame` itself provides more tunability with `-M` and `-C` flags - it seems like we might want to make this more general, too.

Therzok:
I guess I could do that, if I get a green light, I'll start on it.

ethomson:
What do you mean by "a green light"?

Therzok:
I meant another :+1: on the idea. :D

carlosmn:
Since it all comes down to diff options anyway, it does make sense to add that to the blame signature directly.

Therzok:
So, it will take a pointer to `git_diff_options *` in `git_blame_file` (and anything else that goes through `find_origin`) and set it to `GIT_DIFF_OPTIONS_INIT` if not set?

carlosmn:
Yeah, that sounds good. Ideally this wouldn't introduce redundancies between the blame and diff options, but I'm not familiar enough to know if that's the case off-hand.

Therzok:
Actually, we can't pass a `git_diff_options *`, since those options depend on a `git_diff_find_options *`. So the above would have to be translated to:

`git_diff_find_options *` in `git_blame_file` set to `GIT_DIFF_FIND_OPTIONS_INIT` if not set.

> Ideally this wouldn't introduce redundancies between the blame and diff options, but I'm not familiar enough to know if that's the case off-hand.
Judging on [diff find options](https://github.com/libgit2/libgit2/blob/master/include/git2/diff.h#L658-L684) and [blame options](https://github.com/libgit2/libgit2/blob/master/include/git2/blame.h#L70-L79), I see no collisions.

Therzok:
Fixed these things up. Up for review! :dart:

Therzok:
I seem to have introduced regressions. I'll see wtf I screwed up.

Therzok:
Rebased.

src/diff_tform.c
```diff
@@ -296,6 +296,9 @@ static int normalize_find_opts(
 	if (opts->flags & GIT_DIFF_BREAK_REWRITES)
 		opts->flags |= GIT_DIFF_FIND_REWRITES;
 
+	if (opts->flags & GIT_DIFF_FIND_NO_RENAMES)
+		opts->flags &= ~GIT_DIFF_FIND_ALL;
```
ethomson: This is pretty weird.  You end up turning off your own bit, after doing some unnecessary config lookups.  Why not just exit this function early if `GIT_DIFF_FIND_NO_RENAMES` is set?

Also, `GIT_DIFF_FIND_NO_RENAMES` seems like an odd name (since it also turns off copy detection)?  Does `GIT_DIFF_FIND_NONE` make more sense?

ethomson: Oops, upon looking at this again, I realize that `DIFF_FIND_ALL` doesn't involve you unsetting your own bit.  So disregard that complaint, obviously.

Therzok: I can go with GIT_DIFF_FIND_NONE.

Therzok: Do you still want it to exit early? I just placed it where I thought it was appropriate.

askobara:
Hey guys, do you made a decision on this issue?

Therzok:
Bumpety-bumpy-bump?

Therzok:
Anything else needed on this?

Uncommon:
Overdue bump. I'm seeing the git blame slowdown too. Any improvement would be welcome.

Mange:
Any update on this?

include/git2/diff.h
```diff
@@ -588,6 +588,9 @@ typedef enum {
 	/** Turn on all finding features. */
 	GIT_DIFF_FIND_ALL = (0x0ff),
 
+	/** Does no work on trying to find renames. */
+	GIT_DIFF_FIND_NO_RENAMES = (1u << 8),
```
pks-t: Wait, I'm confused. We already have a flag `GIT_DIFF_FIND_RENAMES`, why do we need another flag `GIT_DIFF_FIND_NO_RENAMES`? Shouldn't it be sufficient to just unset the bit for `GIT_DIFF_FIND_RENAMES`?

Therzok: The logic for this was to toggle `GIT_DIFF_FIND_ALL` when encountering `NO_RENAMES`. At least, that's what I recall from 2 years ago.

pks-t: I guess it would make more sense to toggle that whenever `GIT_DIFF_FIND_RENAMES` is not set, then?

Therzok: You can use other flags without using `GIT_DIFF_FIND_RENAMES`. The option I  chose was to toggle off all find features with this flag, as the defaults/options may change over time. If you have any other preferred way of handling this, do tell. I find no other way to make a catch-all, disable rename heuristics when I first wrote this patch.

PR #3168, opened 2015-06-01
Locking and transactional/atomic updates for config

carlosmn:
You sometimes want to update more than a single entry in the configuration. This is not possible currently, which is why exposing a couple of new functions is necessary.

These allow you to lock a configuration file, locking out concurrent writers, and perform many operations without any readers seeing them (and as an implementation detail, without doing any I/O) until you commit it.

I wanted to do this via `git_transaction` but it turns out the API doesn't really fit what we do here, or we'd have to add more namespacing.

/cc @jamill who mentioned the need for something like this when users prefer clicking to opening up `.git/config` with their editor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,10 @@ v0.23 + 1
 
 ### API additions
 
-* `git_config_lock()` and `git_config_unlock()` have been added, which
-  allow for transactional/atomic complex updates to the configuration,
-  removing the opportunity for concurrent operations and not
-  committing any changes until the unlock.
-
+* `git_config_lock()` has been added, which allow for
+  transactional/atomic complex updates to the configuration, removing
+  the opportunity for concurrent operations and not committing any
+  changes until the unlock.
 
 ### API removals
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ v0.23 + 1
 
 ### API additions
 
+* `git_config_lock()` and `git_config_unlock()` have been added, which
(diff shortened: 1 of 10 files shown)


ethomson:
Generally speaking, I think that this is a great idea.

I do have a minor nit with the semantics:  for a `git_transaction`, we create a transaction and then either `commit` or abort it (implicitly, calling `free`).

It would be nice to avoid `lock` / `unlock(bool commit)` being different.  Could we create a config transaction and then `commit` it or abort it so that these two interfaces feel more similar?

carlosmn:
I'm wondering if you mean to create a transaction type as well? Part of the reason I did it this way is due to the large-ish surface area we have for reading and writing configs.

You are quite right that commit-or-free would be a better public API.

ethomson:
No...  well, I don't think so.  I don't know what I was talking about with `git_transaction`.  :P

carlosmn:
I've made `git_config_lock()` return a `git_transaction` which you can then commit or undo the same way as the refs, which should reduce what you have to think about, and performs automatic unlocking in case of freeing.

ethomson:
This is great and imma merge this - but I would like us to come back to this later and think a little bit about the `git_transaction` API.  Right now it feels very refs-heavy and we've just made transaction a bit more abstract.  We might want to rethink some of the naming around that because it seems like I can `git_transaction_set_symbolic_target` on any `git_transaction` even a config one.

carlosmn:
I had originally envisioned `git_transaction` as being more general, but then had to take some refs-specific steps. And now config has a completely different idea on how to perform transactions... yeah we should do a design check on what we want this API to be.

PR #3170, opened 2015-06-01
git_index_entry__init_from_stat: set nsec fields in entry stats

CmdrMoozy:
This is a potential fix for the bug noted in [this issue](https://github.com/libgit2/libgit2/issues/3169).


src/index.c
```diff
@@ -734,8 +734,11 @@ void git_index_entry__init_from_stat(
 {
 	entry->ctime.seconds = (git_time_t)st->st_ctime;
 	entry->mtime.seconds = (git_time_t)st->st_mtime;
-	/* entry->mtime.nanoseconds = st->st_mtimensec; */
-	/* entry->ctime.nanoseconds = st->st_ctimensec; */
+#if !defined(GIT_WIN32) && !defined(__APPLE__)
+	/* Apple and Windows doesn't provide these struct stat fields. */
```
carlosmn: Do Solaris or the BSDs provide these fields? It feels like what you're really trying to look for is a receng GNU libc on Linux which should be checked for explicitly, not by assuming there are three OSs.

carlosmn: Furthermore, shouldn't we be reading these only if `GIT_USE_NSEC` is defined? I think it would turn out cleaner if we only offer the `USE_NSEC` option when building under Linux and remove the OS checks inside the libgit2 code. If we add support for this in Windows, we'd add it inside the compat layer as well, so we likely wouldn't have to change the code here.

CmdrMoozy: Ah, yep, you're right. The BSD's seem to support these fields (at least, [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=stat&sektion=2) and [OpenBSD](http://www.openssh.com/cgi-bin/man.cgi/OpenBSD-current/man2/stat.2?query=stat&sec=2)), however [Solaris doesn't seem to](http://docs.oracle.com/cd/E19109-01/tsolaris8/816-1056/6m7gi1i51/index.html). Perhaps we should enable the USE_NSEC option based upon some CMake feature check determining that these `struct stat` members are present?

ethomson: Indeed.  Certainly `struct __stat64` on Windows has the `nanoseconds` field.  It would be nice if `USE_NSEC` Just Worked there.

CmdrMoozy:
I think with this new commit we'll behave correctly everywhere but on Windows (`USE_NSEC` is simply unavailable on Windows). For Windows, I believe @carlosmn is correct that we'll need some sort of compat layer which lets us get at Windows' nanosecond-precision m/ctime fields cleanly.

CmdrMoozy:
@ethomson Glancing at the [MSDN docs](https://msdn.microsoft.com/en-us/library/14h5k7ff.aspx), it seems that `struct __stat64` contains a 64-bit `time_t`, but that value is still only in seconds, not nanoseconds. My Google skills are failing me, and I can't seem to find any MSDN documentation on how to get a file's m/ctime in nanoseconds. If you point me in the right direction, I'll happy add a compat layer so this Just Works on Windows as well.

ethomson:
Yes, I must have been making that up.  More importantly, we don't even call the win32 `stat`, we call `GetFileAttributes` and convert all the data ourselves.  I think I commented on that before sufficient coffee.

I had been thinking that this was basically a matter of not `ifdef`ing Windows out.  But obviously I was mistaken.  Given that, I definitely don't have any objections to waiting to deal with Windows until such time as Git for Windows starts doing this also (or until somebody actually cares about libgit2 doing this on Windows).

So anyway, you're welcome to dig in (see `git_win32__file_attribute_to_stat` in `w32_util.h`) if you want, but dealing with `struct stat` on Windows is not an enviable task.

whoisj:
@CmdrMoozy it seems that NTFS doesn't record nano-seconds.

> A file time is a 64-bit value that represents the number of 100-nanosecond intervals that have elapsed since 12:00 A.M. January 1, 1601 Coordinated Universal Time (UTC). The system records file times when applications create, access, and write to files.

Mike Kelly has a rather old blog post about it on MSDN for reference.
http://blogs.msdn.com/b/mikekelly/archive/2009/01/17/unix-time-and-windows-time.aspx

CmdrMoozy:
Thanks, @ethomson and @whoisj!

I'm thinking that the cleanest thing to do to support sub-second time resolution on Windows is to define our own `struct timespec` and a `struct p_stat` structure in win32/posix.h (and typedef it to say `GIT_P_STAT`), which includes the `st_{a,m,c}tim.tv_nsec` fields we expect to find on UNIX systems with a semi-modern glibc. For UNIX, we merely `typedef struct stat GIT_P_STAT`, since that structure already contains the fields we want (or USE_NSEC is unavailable).

This is a bit hacky, but it does keep the hackiness and OS-conditionals confined to posix.h and CMake. Unfortunately, this would mean replacing instances if `struct stat` with `GIT_P_STAT` - a quick grep shows something like 100 +/- 50 instances of this.

Does this sound like a reasonable approach?

ethomson:
I think this is reasonable and this is roughly how I would approach it.

However, I wonder if we might be able to take advantage of the fact that `struct stat` isn't defined by default on Win32 (see https://github.com/libgit2/libgit2/blob/master/src/win32/mingw-compat.h#L14 and https://github.com/libgit2/libgit2/blob/master/src/win32/msvc-compat.h#L13) and just declare our own structure to keep `struct stat` and avoid messing around with typedefs.

kblees:
> just declare our own structure to keep struct stat and avoid messing around with typedefs.

Thats exactly what I did for Git for Windows (see https://github.com/kblees/git/commit/1fd19e8029c1afa350d3c46dc8d78893ee973767). The only problem is that you need to re-implement `stat`, `lstat` and `fstat` from scratch, but you probably have your own implementations for these anyway.

carlosmn:
Yep, we already have our own stat-like implementations for win32, we'd just have to add the nanoseconds to our `struct stat` and fill them in for Windows.

One thing I'd be worried about implementing this on Windows is that Git for Windows does not fill these fields so if *we* do, we might never detect racy entries.

CMakeLists.txt
```diff
@@ -19,9 +19,13 @@ CMAKE_POLICY(SET CMP0015 NEW)
 SET(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules/")
 
 INCLUDE(CheckLibraryExists)
+INCLUDE(CheckStructHasMember)
 INCLUDE(AddCFlagIfSupported)
 INCLUDE(FindPkgConfig)
 
+CHECK_STRUCT_HAS_MEMBER("struct stat" st_mtim.tv_nsec sys/stat.h
+	HAVE_STRUCT_STAT_NSEC LANGUAGE C)
+
```
carlosmn: I'd rather see this check and the addition of the option after the main block, so we keep it uniform and at the beginning.

CmdrMoozy:
@carlosmn If msysgit can fill in the fields, then I think we're fine at detecting racily clean files. Anyway, I don't think it's bad to support it, with the caveat that if you're recompiling to change this option you ought to have some idea of what you're doing.

I think these commits more or less do the job - although I don't have a Windows machine on hand to build, so I'll have to wait for AppVeyor.

CmdrMoozy:
Rebased onto current master, to resolve merge conflicts.

CmdrMoozy:
Is any additional work needed on this fix to get it merged? I believe this pull incorporates everything discussed so far, and is ready to go.

carlosmn:
The index also needs to know whether to consider the nanos in order to consider some entries racily-clean or not. Otherwise you're going to end up with many entries with zero nanos and then use that to compare against file info with filled nanos.

CmdrMoozy:
Sorry it took me so long to get this updated. I switched jobs, so I had to get my new employer's "ok" to continue working on this. I believe this pull is now ready to go, including the changes to account for racily clean file detection in the index code. The Travis failure seems to be due to some sort of networking problem on their OS X build servers. Let me know if more work is needed on this. Thanks!

ethomson:
Sorry for the delay in reviewing this.  Two questions:

1. Why is this a compile-time option and not a run-time option?  (I realize that some systems may lack nsecs and thus we must switch at compile time.)
2. Why is it compiled in by default, if the platform supports it?  Does this match a git setting, or is it optimistic?

These questions may be due to a misunderstanding in how git is computing differences for nsecs.  My git (2.5.3, Windows) is not storing nsecs in the index.  Would this not mean that my libgit2 would always see an index that git had written as dirty (mismatched nsecs)?

kblees:
> Why is it compiled in by default, if the platform supports it?

As I understand these patches, the new USE_NSEC compile time option is disabled by default (as in upstream git).

> My git (2.5.3, Windows) is not storing nsecs in the index.

Git for Windows 2.6.0 *does* store nsecs, but does not use them for comparisons (yet) [1].

> Would this not mean that my libgit2 would always see an index that git had written as dirty (mismatched nsecs)?

Yes. I'm currently working on solutions for the same problem in upstream git [2].

[1] https://github.com/git-for-windows/git/pull/443
[2] http://thread.gmane.org/gmane.comp.version-control.git/278683

CmdrMoozy:
No worries about the delay, thanks for taking the time to look at this. I'm going to push up an extra commit which resolves a merge conflict, and fixes a small issue I noticed when looking at this again myself.

> Why is this a compile-time option and not a run-time option? (I realize that some systems may lack nsecs and thus we must switch at compile time.)

I think the main argument for making it compile-time only is that switching the option back-and-forth at runtime makes it much more likely users will run into the various weird situations @kblees mentioned. Is there a real use-case where we would want to be able to toggle the option at runtime? I think ideally, in the future, all code interacting with Git repositories (Git itself, libgit2, and so on) will all deal with nanoseconds in some common way on all platforms, and we can get rid of this compile-time check.

> Why is it compiled in by default, if the platform supports it? Does this match a git setting, or is it optimistic?

The option is only available at all on platforms which support nanosecond-resolution file times, and on those platforms it is off by default. I think what we want to do in libgit2, if the option is disabled, is:

- Never use nanosecond values for comparisons.
- Never write any nanosecond values to the disk.
- Leave nanosecond values set to 0 for any in-memory data structures (e.g. when we stat a file).

PR #3172, opened 2015-06-02
Change error when running out of ssh agent keys

Therzok:


carlosmn:
I think we should explain in or just before the `if` block what the magical `1` means here. Otherwise, looks sensible.

Therzok:
Done.

PR #3175, opened 2015-06-02
Fixed build warnings on Xcode 6.1

swisspol:


src/odb.c
```diff
@@ -53,7 +53,7 @@ static int load_alternates(git_odb *odb, const char *objects_dir, int alternate_
 int git_odb__format_object_header(char *hdr, size_t n, git_off_t obj_len, git_otype obj_type)
 {
 	const char *type_str = git_object_type2string(obj_type);
-	int len = p_snprintf(hdr, n, "%s %"PRIuZ, type_str, obj_len);
+	int len = p_snprintf(hdr, n, "%s %"PRIuZ, type_str, (size_t)obj_len);
```
carlosmn: This cast is exactly what we don't want to do. This would create an object with the wrong size past `SIZE_MAX`. We might want to use `%lld` for the sizer, but we definitely don't want to truncate the object length.

swisspol:
@carlosmn Agreed. Should be OK now.

swisspol:
Can this be landed?

PR #3177, opened 2015-06-02
Binary diffs: store deltas in the diff structure, include binary data in diff callbacks

ethomson:
First, always load the actual diff file contents when we're being asked to show binary data in the diffs, instead of trying to load it from the ODB to produce it on-demand.  Although it would be nice to simply load it from the ODB, we obviously may not actually be producing a diff from an entry in the ODB.  The diff file contents encompasses all diff data origins, be it from the index/tree (and thus the ODB) or directly from the working directory.

Second, include the binary data in the diff callbacks (instead of pumping it through the file callback).  Although this is a signature change, it enables a much nicer mechanism so that callers like `git_diff_blobs` can load the binary data.

(These changes are important to being able to create binary patches from the index and working directory data.)


ethomson:
Oops, looks like we call `diff` everywhere.  I rebased this on master!

src/diff_patch.c
```diff
@@ -207,6 +220,98 @@ static int diff_patch_invoke_file_callback(
...
+
+	/* The git_delta function accepts unsigned long only */
+	if (!git__is_ulong(a_datalen) || !git__is_ulong(b_datalen))
+		return GIT_EBUFS;
+
+	if ((error = git_zstream_deflatebuf(&deflate, b_data, b_datalen)) < 0)
+		goto done;
+
+	/* The git_delta function accepts unsigned long only */
+	if (!git__is_ulong((git_off_t)deflate.size)) {
```
carlosmn: This seems to be copied from the existing pattern, but this cast... we don't do it in the two calls above with the same type, so it shouldn't even be guarding against a warning.

src/diff_patch.c
```diff
@@ -67,15 +39,38 @@ static void diff_patch_init_common(git_patch *patch)
...
+		memcpy(out, &default_opts, sizeof(git_diff_options));
+	}
+
+	out->old_prefix = opts && opts->old_prefix ?
+		git__strdup(opts->old_prefix) :
+		DIFF_OLD_PREFIX_DEFAULT;
+
+	out->new_prefix = opts && opts->new_prefix ?
+		git__strdup(opts->new_prefix) :
+		DIFF_NEW_PREFIX_DEFAULT;
```
carlosmn: This seems odd and ripe for some errors. Wouldn't it be simpler to always duplicate the string instead of testing for each of them later when we try to free them?

tests/diff/binary.c
```diff
@@ -261,3 +261,232 @@ void test_diff_binary__delta_append(void)
...
+
+	return 0;
+}
+
+static int hunk_cb(
+	const git_diff_delta *delta,
+	const git_diff_hunk *hunk,
+	void *payload)
+{
+	cl_assert("did not expect hunk callback");
```
carlosmn: This (and the next) assert would always succeed, since it's comparing the string against `0`. It looks like this wants to abort the test if these callbacks get called, in which case `cl_fail()` would be more appropriate.

carlosmn:
Neat, though the signature change should have a corresponding entry in the changelog.

src/diff_patch.c
ethomson: WTF this is insane.  Cast a `size_t` to a `git_off_t` to see if it fits in a `ulong`.

ethomson: I dropped the cast, but I'm going to follow up with a PR to tweak `git__is_<inttype>` because this is madness.

src/diff_patch.c
ethomson: Fixed.

tests/diff/binary.c
ethomson: Fixed.

swisspol:
@ethomson You might wanna build your PRs using Clang or enable more compiler warnings - see #3221 :)

Issue #3095, opened 2015-05-02
print_usage functions is defined but not used in describe.c


tfarina:
describe.c:106:13: warning: 'print_usage' defined but not used [-Wunused-function]
 static void print_usage(void)
             ^

carlosmn:
Fixed in #3179

PR #3179, opened 2015-06-03
print_usage functions is defined but not used

arielb2:
Hello.

Use the previously created function to display a message when the arguments are not valid.

https://github.com/libgit2/libgit2/issues/3095

]$ ./describe --abcde
usage: see `git help describe


carlosmn:
Thanks!

PR #3181, opened 2015-06-04
global: Ensure we release OpenSSL-related resources

thentenaar:
This commit ensures that various things, such as the SSL error strings,
are freed at shutdown by adding a uninit_ssl() counterpart to init_ssl().

With a simple program that just inits and shutsdown libgit2, valgrind
reveals:

before: still reachable: 82,271 bytes in 2,560 blocks
after:  still reachable: 88 bytes in 3 blocks

The remaining blocks should be released by SSL_COMP_free_compression_methods
where available. See: openssl/openssl@db7b5e0d768485d9c7f8d3eaa8f28d2f254e20d6

### DIFF
--- a/src/global.c
+++ b/src/global.c
@@ -79,7 +79,8 @@ static void shutdown_ssl_locking(void)
 	int num_locks, i;
 
 	num_locks = CRYPTO_num_locks();
-	CRYPTO_set_locking_callback(NULL);
+	if (CRYPTO_get_locking_callback() == openssl_locking_function)
+		CRYPTO_set_locking_callback(NULL);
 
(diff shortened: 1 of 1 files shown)


carlosmn:
Does OpenSSL reference-count how many users have initialised the library? Otherwise we can't ask it to free the strings which a different part of the program is going to depend on.

thentenaar:
Unfortunately it doesn't, and you're quite right. However, we can still ensure we free the context we allocated, and check the locking callback so we don't step on the toes of other libraries that may also use OpenSSL.

My latest commit should do just that.

thentenaar:
Hell, I'll just make a new PR since it's really a separate change.

PR #3183, opened 2015-06-05
Implement a cURL stream

carlosmn:
cURL has a mode in which it provides TLS and proxy setup but otherwise allows us to use one of its handles very much like our own streams, so it's fairly simple to make the library use libcurl for talking to the server.

~~~Unfortunately I have not found a way to extract the certificate information without implementing it via the multiple crypto engines which curl uses, which is not something we can do.~~~

What we'll do here is use libcurl as the underlying stream and then build our own TLS streams on top of that. This allows us to keep doing what we do with certificates while letting libcurl take care of the proxy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,9 @@ support for HTTPS connections insead of OpenSSL.
 * The race condition mitigations described in `racy-git.txt` have been
   implemented.
 
+* If libcurl is installed, we will use it to connect to HTTP(S)
+  servers.
+
 ### API additions
 
 * The `git_merge_options` gained a `file_flags` member.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,7 @@ OPTION( USE_ICONV			"Link with and use iconv library" 		OFF )
 OPTION( USE_SSH				"Link with libssh to enable SSH support" ON )
 OPTION( USE_GSSAPI			"Link with libgssapi for SPNEGO auth"   OFF )
 OPTION( VALGRIND			"Configure build for valgrind"			OFF )
+OPTION( CURL			"User curl for HTTP if available" ON)
 
 IF(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 	SET( USE_ICONV ON )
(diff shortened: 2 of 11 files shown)


carlosmn:
Extracting certificate information out of the curl handle is rather obtuse. It is only implemented for some crypto libraries, and it's in the form of a list of "key:value" strings so on the libcurl which comes with OSX, we're not able to get any information about the certificate, just whether SecureTransport figured it was valid or not.

This is way below the standard which we've set for ourselves with the other crypto backends, but there's no other way we're going to get proxy support.

carlosmn:
This should be about ready to go in, though the proxy code is still untested. Testing *that* is going to be a barrel of laughs.

There's also the question of how do we test this stream. We shouldn't replace all the builds with a liburl one, so do we add an extra entry to the matrix asking the build system to install libcurl-dev?

carlosmn:
As a way to work around libcurl not quite giving us the information from the certificate, I think it'd be worth considering using libcurl only for the proxying capabilities, and send our TLS stream though it. The SecureTransport stream should be ready to take this on, as it sends data over to the socket stream, but having OpenSSL use a user-defined channel is not something on which I've found much information.

dirkhh:
Have you looked at CURLOPT_HTTPPROXYTUNNEL mode to use libcurl to ONLY provide a tunnel and do all the https negotiation / certificate extraction / etc from the libgit2 side?
For my use case that isn't critical, but since you brought it up... :-)

This is an extremely timely addition as lack of https proxy support was turning into a major headache for me. Thanks.

carlosmn:
Yeah, using libcurl solely as a tunnel is the current plan due to the differences in cert reporting across its crypto backends. Maybe I can get away with grabbing its socket descriptor and feeding it into OpenSSL, otherwise I'll have to figure out how to create a BIO.

carlosmn:
I'm not sure if there's a reliable[0] way to make CI test the proxy bits, but I have just tested locally and it does work, both by setting `http.proxy` and by exporting `http_proxy`, the latter is taken up by libcurl without us having to do anything other than set the option for tunneling. Setting this option unfortunately seems to make socks proxies not work, but this is still a win.

The Mac builds automatically have libcurl available, though I think the Ubuntu VMs don't have libcurl-dev installed, which means we won't exercise that path there, and always installing it would make us not use the plain socket stream. Maybe we can install it on the debug builds?

Regardless, I'd like to see this get merged soon so it gets some testing in the wild before we make the next release with it.

[0] Emphasis on "reliable". We can install tinyproxy and grep its logs, but that's yet another thing that can go wrong.

dirkhh:
I tried it yesterday with my corporsate environment and couldn't get it to work at all.
I tried both setting http.proxy in my application as well as setting http_proxy in my environment.
What I didn't do is "set the option for tunneling" which might explain my issue. I guess I don't know how/where to do that.
It might be nice to have an example or some documentation how to use this somewhere :-)
I'll be happy to try again if you can point me in the right direction.

carlosmn:
You use this the same way you use it with git, you set `http.proxy` or `remote.<name>.proxy` to select a particular proxy, or you specify `http_proxy` to let libcurl automagically set the proxy.

I don't know where you expect to specify the tunneling mode, it is an option we pass to libcurl when we set up the handle.

dirkhh:
The comment to tunneling mode was in response to what you wrote :-)
With just setting http.proxy (or http_proxy) it didn't work for me. And I verified in the debugger that libcurl was indeed used. And the SSL_Connect(st->ssl) in line 329 of openssl_stream.c fails.

carlosmn:
The code for tunneling TLS over it was only working early this morning, so whatever you used last night probably didn't have the tunneling on.

The "us" in the comment is libgit2. Users of the library are explicitly called out as such.

dirkhh:
Duh. The pull request didn't show new commits so I didn't realize you had added more code. Apologies. I just tried with the latest of the cmn/curl-stream branch and it works perfectly.
+1 merging this into master
Thanks for your great work!

PR #3184, opened 2015-06-06
Only write index if updated when passing GIT_DIFF_UPDATE_INDEX

swisspol:
When diffing the index with the workdir and GIT_DIFF_UPDATE_INDEX has been passed,
the previous implementation was always writing the index to disk even if it wasn't
modified.

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -661,6 +661,7 @@ int git_diff__oid_for_entry(
 		if (!(error = git_repository_index__weakptr(&idx, diff->repo))) {
 			git_oid_cpy(&entry.id, out);
 			error = git_index_add(idx, &entry);
+			diff->index_updated = true;
 		}
  	}
 
@@ -1353,7 +1354,7 @@ int git_diff_index_to_workdir(
 			&b, repo, index, NULL, GIT_ITERATOR_DONT_AUTOEXPAND, pfx, pfx)
 	);
 
-	if (!error && DIFF_FLAG_IS_SET(*diff, GIT_DIFF_UPDATE_INDEX))
+	if (!error && DIFF_FLAG_IS_SET(*diff, GIT_DIFF_UPDATE_INDEX) && (*diff)->index_updated)
 		error = git_index_write(index);
 
 	return error;
--- a/src/diff.h
+++ b/src/diff.h
@@ -64,6 +64,7 @@ struct git_diff {
 	git_iterator_type_t new_src;
 	uint32_t diffcaps;
 	git_diff_perfdata perf;
+	bool index_updated;
 
 	int (*strcomp)(const char *, const char *);
 	int (*strncomp)(const char *, const char *, size_t);


swisspol:
Any feedback on this PR? The diff is a bit large but very simple: it's mostly about passing around a `bool` to remember if the index was modified or not.

ethomson:
Sorry, I think that everybody's a bit heads-down on various things right now.  I agree that this seems pretty straightforward (and very useful!) but I'll put a closer set of eyes on this as soon as I'm able.  

Sorry for the delay, and thanks again.

swisspol:
No worries, was just curious if my approach was not good for whatever reason, in which case I'd rather know early so I have time to revise it :)

src/diff.c
```diff
@@ -580,14 +580,15 @@ int git_diff__oid_for_file(
...
+	return git_diff__oid_for_entry(out, diff, &entry, NULL, NULL);
 }
 
 int git_diff__oid_for_entry(
 	git_oid *out,
 	git_diff *diff,
 	const git_index_entry *src,
-	const git_oid *update_match)
+	const git_oid *update_match,
+	bool *index_updated)
```
ethomson: We put `out`-type parameters as the first params - could you change this to `git_diff__oid_for_entry(git_oid *out, bool *index_updated...`

src/diff.c
```diff
@@ -1192,7 +1196,8 @@ int git_diff__from_iterators(
 	git_repository *repo,
 	git_iterator *old_iter,
 	git_iterator *new_iter,
-	const git_diff_options *opts)
+	const git_diff_options *opts,
+	bool *index_updated)
```
ethomson: Similarly, could the `out` type parameter move to the front of the list, after the `git_diff`?

Alternately, it might be nice to move the `index_updated` value into the `git_diff` structure itself.

swisspol:
I rewrote it to simply add a field to `git_diff` which I forgot was an opaque struct! Much better now :)

ethomson:
:+1:

swisspol:
For the record, the way to test this change, put a breakpoint on `git_index_write()` inside `git_diff_index_to_workdir()`, then use libgit2 to stage / unstage files and verify the breakpoint is not hit, then `touch` some files in Terminal and verify the breakpoint is hit.

swisspol:
Can this be landed now?

ethomson:
> For the record, the way to test this change, put a breakpoint on ...

That's really not the way we prefer to test.  We would prefer to see a unit test that ensures that the index is written when there's actually a change, and it's not written when there isn't.

swisspol:
Yeah I know, but that's far from trivial to write such a test (e.g. how
would you verify the index was written or not? Stat the file?). So in the
meantime I wrote this note for the record.

ethomson:
> how would you verify the index was written or not? Stat the file?

Yes, exactly that.

Writing unit tests is indeed nontrivial.  And you're under no obligation to include them, of course, but to set expectations: by not doing so, you're delaying the bug fix until somebody else decides to care enough about this bug to write some.

ethomson:
Closed via #3234

PR #3187, opened 2015-06-06
buffer: Fix growing non-allocated buffers to modulo 8 sizes

b4n:
`git_buf_grow()` ensures the buffer is terminated with a NUL byte, but dares to truncate the buffer if it is too short.  So, allocation has to always be 1 more than the required content size.

This is generally not a problem because for already allocated buffers the grow logic makes it impossible for the content size to be equal to the allocated size -- unless the target size is the same as the allocated size, but this has a special early-out handling.

However, for non-allocated buffers, the allocation is exactly the requested target size if the target size is a multiple of the alignment (8), in which case truncation will occur if the content size is the same as the requested target size (which is especially the case when passing `target_size=0` for automatic growth).

Fix this by allocating one more byte than actually requested when the buffer isn't already allocated.

---

While this fixes the issue, I wonder if it is sane for `git_buf_try_grow()` to even have code for shrinking the buffer.  OK, it tries to always keep a NUL byte at the end, but I'm not sure this should really be a check, but rather an assertion (as AFAIU, there should never have been any code path leading to the shrinking).

BTW, 0.22.2 *can* sometimes return buffers that aren't NUL-terminated, at least through a `git_blob_filtered_content()` when no transformation happens.  I'm not sure if it's considered a bug (the same function in master apparently grows the buffer properly no matter what), but if non-NUL-terminated buffers are not a bug `git_buf_grow()` probably shouldn't enforce a NUL, especially if there isn't room for it.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -48,7 +48,7 @@ int git_buf_try_grow(
 		return 0;
 
 	if (buf->asize == 0) {
-		new_size = target_size;
+		new_size = target_size + 1;
 		new_ptr = NULL;
 	} else {
 		new_size = buf->asize;
--- a/tests/buf/basic.c
+++ b/tests/buf/basic.c
@@ -15,6 +15,45 @@ void test_buf_basic__resize(void)
 	git_buf_free(&buf1);
 }
 
+void test_buf_basic__auto_grow(void)
+{
+	git_buf buf1 = GIT_BUF_INIT_CONST("1234567890", 10);
+	git_buf buf2 = buf1;
+
+	cl_git_pass(git_buf_grow(&buf1, 0));
+	cl_assert(git_buf_oom(&buf1) == 0);
+	cl_assert_equal_i(buf1.size, buf2.size);
+	cl_assert(memcmp(buf1.ptr, buf2.ptr, buf1.size) == 0);
+
+	git_buf_free(&buf1);
+}
+
+void test_buf_basic__auto_grow_8(void)
+{
+	git_buf buf1 = GIT_BUF_INIT_CONST("1234567890", 8);
+	git_buf buf2 = buf1;
+
+	cl_git_pass(git_buf_grow(&buf1, 0));
+	cl_assert(git_buf_oom(&buf1) == 0);
+	cl_assert_equal_i(buf1.size, buf2.size);
+	cl_assert(memcmp(buf1.ptr, buf2.ptr, buf1.size) == 0);
+
+	git_buf_free(&buf1);
+}
+
+void test_buf_basic__grow_8(void)
+{
+	git_buf buf1 = GIT_BUF_INIT_CONST("1234567890", 8);
+	git_buf buf2 = buf1;
+
+	cl_git_pass(git_buf_grow(&buf1, buf1.size));
+	cl_assert(git_buf_oom(&buf1) == 0);
+	cl_assert_equal_i(buf1.size, buf2.size);
+	cl_assert(memcmp(buf1.ptr, buf2.ptr, buf1.size) == 0);
+
+	git_buf_free(&buf1);
+}
+
 void test_buf_basic__resize_incremental(void)
 {
 	git_buf buf1 = GIT_BUF_INIT;


carlosmn:
As I mentioned elsewhere, the fact that we tell the user to do this is a bug in the design. If you do not own the buffer, you should not be doing anything to it and a user of the library should never do anything apart from reading from a buffer it got as output.

We should remove this "feature" from `_grow()` and the recommendation to use it.

b4n:
As also said on IRC, I don't mind much either way.

Though, the current feature permits to lower the unnecessary memory copies, as if the buffer is already allocated, calling `git_buf_grow(buf, 0)` won't do anything, whilst if the caller does it, it has to always copy no matter what (as it can't steal the buffer as it doesn't have `git__free()` to release it).

Also, if this feature is removed I think all API returning a buffer should (obviously I guess) either always return an allocated buffer the user owns and is always responsible to free with `git_buf_free()` (hence rendering `git_buf_grow(buf, 0)` virtually useless as it'd be already grown), or one the user never has to free (memory owned by something else in libgit2, e.g. a blob), but never both as e.g. `git_blob_filtered_content()` does in 0.22 -- as it would mean the user has to always `git_buf_free()` it, but still cannot rely on the buffer actually owning the data.

---

BTW, all this said, I think applying at least the patch fixing the actual issue (the +1) is probably a good idea, even if you want to get rid of this API in the (near) future, as the fix is really trivial, and the issue affects 0.22.

carlosmn:
There doesn't need to be any difference in memory consumption. If you're going to take advantage of the fact that the buffer is a reference to the blob's own buffer, you *already* have to take that into account for figuring out when to free the blob. If you want to do that, you don't call `git_buf_grow(..., 0)` at all.

We can cherry-pick the second commit to the maint branch to fix existing uses, but that should be accompanied by patches removing the recommendation to use this function.

b4n:
> There doesn't need to be any difference in memory consumption. If you're going to take advantage of the fact that the buffer is a reference to the blob's own buffer, you *already* have to take that into account for figuring out when to free the blob. If you want to do that, you don't call `git_buf_grow(..., 0)` at all.

No, my point is when you want to "detach" the buffer from the blob, so you can free the blob but keep the buffer.
Currently, all you have to do is call `git_buf_grow(..., 0)`, and you know you own the buffer.  This call will indeed copy the data if it isn't already allocated, but it will do *nothing* if the buffer is already allocated -- no copy.

OTOH, if you don't have this kind of API and you want to own the buffer, your only option is to duplicate the memory yourself, no matter what (as you can't steal it, you don't have access to `git__free()`) [1].

So currently to "detach" the buffer form the blob, you have to do
```c
...
git_blob_filtered_content(&buf, ...);
git_buf_grow(&buf, 0); // might or might not copy, as needed
// you can git_blob_free() all you want, buffer is still good
```
And without it, you'd have to do something like
```c
...
git_blob_filtered_content(&buf, ...);
ptr = memdup(buf->ptr, buf->size);
size = buf->size;
git_buf_free(&buf, 0);
```

---
[1] or have some highly convoluted logic storing the git buffer itself if it's allocated, and your own data structure otherwise, but that's just becoming crazily complex.

ethomson:
I had to read your paragraph five times to understand what you were trying to get `git_buf_grow` to do.

I understand your motivation.  `git_buf_grow` isn't the answer here.  Create a mechanism that is more obvious about what it's doing.

b4n:
> I had to read your paragraph five times to understand what you were trying to get `git_buf_grow` to do.
>
> I understand your motivation. `git_buf_grow` isn't the answer here.

Heh, I was merely using `git_buf_grow()` the way it's suggested in the [`git_blob_filtered_content()` docs](https://libgit2.github.com/libgit2/#HEAD/group/blob/git_blob_filtered_content) :)

> Create a mechanism that is more obvious about what it's doing.

Well, the only thing I find a bit obscure is the `git_buf_grow()` call itself, so what about simply doing something equivalent to this?
```C
#define git_buf_own(b) (git_buf_grow((b), 0))
```

carlosmn:
Having a function to perform the copy isn't necessarily completely out of the question, if we can find a sensible name and semantics for it. Calling `gi_buf_grow(..., 0)` is not how we'd achieve it, as we won't allow this.

A function such as `git_buf_ensure_data_owned()` which does the job of `git_buf_grow(..., 0)` might be feasible, as it would *only* have the logic to either leave the buffer alone or create a copy of the pointer it's borrowing and `git_buf_grow()` would return an error if it does not own the memory you're asking to grow.

But even then stuff starts to get complicated. A better solution might be never taking a reference to the blob's content but always copy, or provide a way to write the filtered content to a `git_writebuf` so the caller can process stuff as in-line as it gets.

carlosmn:
The recommendation has been removed, and we no longer allow "growing" a borrowed buffer.

PR #3188, opened 2015-06-07
global: Ensure we free our SSL context.

thentenaar:
Additionally, check if the locking callback is already set before
setting it, and if it's our callback before unsetting it. This way we
can play a bit nicer with other libraries that also use OpenSSL that may
also be part of the application using our library.

### DIFF
--- a/src/global.c
+++ b/src/global.c
@@ -116,6 +116,20 @@ static void init_ssl(void)
 #endif
 }
 
+/**
+ * This function aims to clean-up the SSL context which
+ * we allocated.
+ */
+static void uninit_ssl(void)
+{
+#ifdef GIT_OPENSSL
+	if (git__ssl_ctx) {
+		SSL_CTX_free(git__ssl_ctx);
+		git__ssl_ctx = NULL;
+	}
+#endif
+}
+
 int git_openssl_set_locking(void)
 {
 #ifdef GIT_OPENSSL
@@ -333,6 +347,7 @@ int git_libgit2_shutdown(void)
 
 	/* Shut down any subsystems that have global state */
 	git__shutdown();
+	uninit_ssl();
 
 	ptr = pthread_getspecific(_tls_key);
 	pthread_setspecific(_tls_key, NULL);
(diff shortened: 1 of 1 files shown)


carlosmn:
Freeing the SSL context is good, but I don't like the changes to the locking functions. libgit2 isn't the one who is driving the decision to set the locking functions, but the application programmer. If we handles cases of double-assignment, we're just hiding bugs in the caller, and that's not a business we want to be in.

thentenaar:
I've updated my commit to go ahead and set the assignment of the locking callback as it was. However, I think it's worthwhile to check whether or not it's set to libgit2's locking function before setting it to NULL outright. If it's not set to libgit2's locking function, then libcrypto isn't using libgit2's locks anyway, so setting it to NULL to avoid libcrypto trying to use locks which have already been freed is an exercise in futility at that point. Surely it can't hurt to do the check. :P

carlosmn:
If that happens, it's still a bug in the application. The only reason we even allow for these functions to get set is as a last resort in case the language/environment of the application makes it unreasonably hard to do it properly. If you're setting them, then there's absolutely no reason for you to have asked libgit2 to set it up and doing so is a bug. If you're asking two different pieces of code to set up the locks, you're setting yourself up to crashes as well.

thentenaar:
Perhaps then, there should be a way to specify to libgit2 that openssl has already been initialized by the application. I tend to blame openssl for not having a coherent way of doing initialization / uninitialization, but looking again at global.c I see that the function to setup the locks isn't implicitly called, as I'd assumed, so then it's a moot point.

Let's start with freeing the SSL context.

carlosmn:
>Perhaps then, there should be a way to specify to libgit2 that openssl has already been initialized by the application

This is the assumption which libgit2 starts with. libgit2 will just use OpenSSL and assume everything is fine (assuming we're not on Windows or OS X where we don't have to use it) as it's the application's responsibility to initialize the threading.

The function to set up the locking is for situations where libgit2 is the only reason you're even touching OpenSSL, which is primarily the situation for languages with their own crypto stacks. This function is also not sufficient to guarantee safe use of crypto as e.g. libssh2 may depend on libgcrypt, or it may depend on OpenSSL but libgit2 doesn't know about OpenSSL so even calling this function won't do anything.

PR #3189, opened 2015-06-08
Implement the --log-size and -i --regexp-ignore-case option for examples/log.c

arielb2:
Hello.
print the message size in bytes.

./log --log-size

./log --author "search" -i


carlosmn:
There seems to be a merge in your history. For keeping up with the main branch, we much prefer rebases if the code hasn't made it into master yet.

examples/common.h
```diff
@@ -11,7 +11,7 @@
  * with this software. If not, see
  * <http://creativecommons.org/publicdomain/zero/1.0/>.
  */
-
+#define _GNU_SOURCE
```
carlosmn: Why is this necessary?

arielb2: avoid warnings /log.c:195:5: warning: implicit declaration of function ‘strcasestr’ [-Wimplicit-function-declaration]
and for the sake of supporting old versions de glibc.

Info 
http://linux.die.net/man/3/strcasestr
http://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html

carlosmn: Doesn't this restrict this file to GNU systems then? What happens on the BSDs?

carlosmn:
I'm not worried about BSD definitions, I'm worried about BSD (and other) operating systems. Does `strcasestr()` exist on OpenBSD? What about Solaris? These tests are supposed to be able to build on Windows as well, does it exist there?

arielb2:
well, apparently not according to this:
<blockquote>

<p>
https://www.gnu.org/software/gnulib/manual/html_node/strcasestr.html

12.71.8 strcasestr

Gnulib module: strcasestr or strcasestr-simple
<br>
Portability problems fixed by either Gnulib module strcasestr-simple or strcasestr:
<br>
    This function is missing on some platforms: AIX 5.1, HP-UX 11, IRIX 6.5, OSF/1 5.1, Solaris 10, Cygwin 1.5.x, mingw, MSVC 9, BeOS.<br>
    This function can trigger memchr bugs on some platforms: glibc 2.10.<br>
    This function can trigger false positives for long periodic needles on some platforms: glibc 2.12, Cygwin 1.7.7. <br>
<br>
Portability problems fixed by Gnulib module strcasestr:
<br>
    This function has quadratic instead of linear worst-case complexity on some platforms: glibc 2.8, FreeBSD 6.2, NetBSD 5.0, OpenBSD 4.0, Solaris 11 2011-11. <br>
<br>
Portability problems not fixed by Gnulib:
</p>
</blockquote>

<br>

In some versions of * BSD or Solaris exists and not in others.

I will review what the options are for this case and will update the pr.
Thank you for your comments and review.

ethomson:
Indeed we do need to make sure that our examples compile across platforms.  It's much more important to illustrate *driving* libgit2 than necessarily providing useful examples.  For housekeeping purposes, I'm going to close this, since there hasn't been any movement here recently, but please do reopen if you'd like to continue to work on it.

PR #3191, opened 2015-06-08
Add test to investigate case folding on Windows.

jeffhostetler:
Here is a test that explores the quirks of case-folding on Windows.  Before we can choose the right replacement for strcasecmp() and friends, I wanted to know exactly what the filesystem does.  This tries to determine that by brute-force colliding files.  THEN confirming that the our version of strcasecmp() agrees with that.

TODO:
- [] If we like the way this is going, I can write a ntfs/windows version of git__strcasecmp() that behaves like the filesystem using CompareStringOrdinal().
- [] Deal with calls to tolower().
- [x] Maybe add a more detailed Turkish-i case -- It is my belief that NTFS only does type "C" and "S" foldings, not type "T" (nor type "F"). And the non-symmetric Turkish-i cases are type "F", so I think we're OK here.

### DIFF
--- a/tests/fsquirks/windows/case.c
+++ b/tests/fsquirks/windows/case.c
@@ -1,0 +1,569 @@
+/**
+ * This test examines case insensitivity on the Windows filesystem.
+ * Specifically, how non-US-ASCII characters are handled.  It tries
+ * to answer questions about what subset of the full Unicode spec is
+ * ACTUALLY IMPLEMENTED by the filesystem.  We brute-force create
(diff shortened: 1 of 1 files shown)


jeffhostetler:
See also #3165

ethomson:
This PR is now a few years old; closing this — please do open a new one if you're interested in making changes. 🙏

jeffhostetler:
Hey there! Hope you're doing well.

What a blast from the past -- and wow, was that really 8+ years ago?

PR #3197, opened 2015-06-09
Created clone test for 5GB repos.

jeffhostetler:
Here is a test to create a local 5GB repo and locally clone it.  The purpose is to confirm that we send receive packfiles larger than 4GB.  It creates 5 uncompressible 1GB files in a series of commits rather than trying to create a single huge file.

This is to complement #2932.

This test replaces my earlier tests described in #3121 and #3162 (which tried to create a single huge file (and which causes other problems beyond the scope of the original problem)).

### DIFF
--- a/tests/clone/big.c
+++ b/tests/clone/big.c
@@ -1,0 +1,228 @@
+/**
+ * This test creates a big repo (larger than 4GB) and tries to clone it.
+ * The purpose is to confirm that we can build and receive large packfiles.
+ */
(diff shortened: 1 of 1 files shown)


pks-t:
I've adopted this PR, see #4593. I started out from your code and heavily refactored it to match our current code style. As a result, I've taken ownership of the commits, but credit you in the message. I hope you don't mind. Anyway, thanks for this ancient PR ;)

PR #3199, opened 2015-06-09
CRLF

ethomson:
Improvements to handle LF->CRLF filtering into the working directory.  The real change here is to introduce a large corpus of text files that were produced by git core in a variety of settings (as discussed in #2798), and a simple test that will loop through various CRLF and attribute settings, comparing what libgit2 produces to what git core produces.

I did _not_ wire up the build system to do anything special here, but in the future, I would like to have our CI builds run this script so that we have immediate feedback between git and what we perform.  (While I think that we should trust our test resources being checked in, I would love to get yet another data point.  This could show us skew between the test resources we have checked in and what the CI server's git installation produces.)

But after introducing this test and test data, this quickly illustrated the deficiencies that #2798 complained about.  These are:

* Perform CRLF filtering on non-Windows platforms when `core.autocrlf=true`.  This matches core git.  Yes, really.  Why are you laughing?  This is a serious matter.
* Handle `core.autocrlf=input` when the `text` attribute is specified.
* Don't abort when filtering a file with mixed line endings
* Better handle binary data, using statistics from git core.

After doing this, I don't *super love* introducing 1400 new files to hang out in `tests/resources`.  (But I still like this better than having a script that produces 1400 new unit tests.)  If somebody else has a better idea, I'm all ears.


ethomson:
If this is green on the first pass I'm going to go buy a lottery ticket for truly I am fortunate today.

ethomson:
Note that this wasn't really meant to (fully) obviate #2798.  Per the filtering to the worktree, it's very possible that there were other issues that #2798 was fixing that are not covered here.  And certainly there is nothing here to address the filtering into the odb.  I really just wanted to write the test script and then had to fix up the failures to make sure that it was working.  O_o

linquize:
Still have CRLF issue?

carlosmn:
Good times. So if I understand correctly, those new files are all worktrees for the repo, which gets cloned and then different combination of settings are applied, no? That's one way, I guess.

I was recently looking into tackling this, and was thinking of just porting the git tests as an alternative to having the resources handy, which seemed like an easier way to go, as we describe the contents in the tests, rather than via going through a bunch of predefined files, but maybe it's not that easy, since the PRs keep testing against fixed datasets.

ethomson:
Yeah, that's the strategy in a nutshell.  Like I said, I don't love it, but I do like the repeatability of it.

csware:
Lemme have a look, maybe I can adopt your testing strategy for pr #2798

carlosmn:
:+1: having the script is great for repeatability. It looks like the first batch of racy fixes caused conflicts. If you'd rebase, we can hit the button.

PR #3203, opened 2015-06-10
gitattributes: let clients use native line endings

ethomson:
Use "text=auto" to ensure that we get LFs in the repository, but
let clients have their native line endings in their worktree.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,1 @@
-*.c eol=lf
-*.h eol=lf
+* text=auto


whoisj:
do want things like `.gitattributes eof=lf`and `.gitignore eof=lf`? Not sure how Git for Windows handles those files if they have CRLF line-endings.

carlosmn:
There really shouldn't be any issues reading from from files with CRLF.

ethomson:
No, that shouldn't be necessary.  (Indeed, `.gitattributes` can't apply attributes to itself.)

ethomson:
For those following along at home, when I said:

> Indeed, .gitattributes can't apply attributes to itself.

Yeah, that's a lie.

> When the .gitattributes file is missing from the work tree, the path in the index is used as a fall-back

So...  when checking out, if you don't have a `.gitattributes` on disk, it would look at the `.gitattributes` in the index to write the `.gitattributes`.  So...  a `.gitattributes` *can* apply to itself.

PR #3204, opened 2015-06-10
Fixed handling of GIT_DELTA_CONFLICTED  in git_diff_find_similar()

swisspol:
git_diff_find_similar() now ignores git_diff_delta records with a status
of GIT_DELTA_CONFLICTED, which fixes a crash due to assert() being hit.

### DIFF
--- a/src/diff_tform.c
+++ b/src/diff_tform.c
@@ -676,11 +676,13 @@ static bool is_rename_target(
 		return false;
 
 	/* only consider ADDED, RENAMED, COPIED, and split MODIFIED as
-	 * targets; maybe include UNTRACKED and IGNORED if requested.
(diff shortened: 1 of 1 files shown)


swisspol:
To repro the crash, one had to do a rebase with conflicts in http://gitup.co which then shows the conflict resolver UI, which then shows a unified diff HEAD to workdir created by calling:

1. git_diff_tree_to_index()
2. git_diff_index_to_workdir()
3. git_diff_merge()
4. git_diff_find_similar() <------ CRASH in assert() at `assert(delta_is_split(tgt));`

ethomson:
Looks good to me.  AppVeyor failed on one of the remote tests, so I kicked it.

Thanks!

Issue #3206, opened 2015-06-10
Secure Transport path is not tested on OS X


swisspol:
Running all unit tests, it appears that the callbacks set by `SSLSetIOFuncs()` are never called.

swisspol:
This was observed while working on #3195.

carlosmn:
This is simply not true. SecureTransport is the only way the code in master has of talking TLS on OS X. If you don't see them called, you're not running the network tets.

swisspol:
I'm running the clar tests from Xcode using the Xcode project generated with CMake. I pass no arguments and it's running a ton of tests.

Am I supposed to pass an explicit argument to enable network tests?

swisspol:
I just ran again the tests on `master` and the Xcode console says:
> Loaded 291 suites: 
> Started

I have a breakpoint on `git_stransport_stream_new()` and it's never called.

carlosmn:
`-ionline` or `-sonline` will activate the tests which talk to the network.

swisspol:
It works now. Is there any other series of tests that is not enabled by default?

carlosmn:
The expensive tests are disabled, as well as those which require special setup, like the pushing tests, or those which use a huge amount of disk space or try to write outside of the temp dirs. `ctest` should be running the network-facing tests, though.

https://github.com/libgit2/libgit2/blob/master/CMakeLists.txt#L537

PR #3208, opened 2015-06-10
Fixed some Secure Transport issues on OS X

swisspol:
(see updated commit message)

### DIFF
--- a/src/stransport_stream.c
+++ b/src/stransport_stream.c
@@ -110,18 +110,25 @@ int stransport_certificate(git_cert **out, git_stream *stream)
 	return 0;
 }
 
+/*
+ * Contrary to typical network IO callbacks, Secure Transport write callback is
+ * expected to write *all* passed data, not just as much as it can, and any
+ * other case would be considered a failure.
+ *
+ * This behavior is actually not specified in the Apple documentation, but is
+ * required for things to work correctly (and incidentally, that's also how
+ * Apple implements it in its projects at opensource.apple.com).
+ *
+ * Libgit2 streams happen to already have this very behavior so this is just
+ * passthrough.
+ */
 static OSStatus write_cb(SSLConnectionRef conn, const void *data, size_t *len)
 {
 	git_stream *io = (git_stream *) conn;
-	ssize_t ret;
 
-	ret = git_stream_write(io, data, *len, 0);
-	if (ret < 0) {
-		*len = 0;
-		return -1;
+	if (git_stream_write(io, data, *len, 0) < 0) {
+		return -36; /* "ioErr" from MacErrors.h which is not available on iOS */
 	}
-
-	*len = ret;
 
 	return noErr;
 }
@@ -141,29 +148,38 @@ ssize_t stransport_write(git_stream *stream, const char *data, size_t len, int f
 	return processed;
 }
 
+/*
+ * Contrary to typical network IO callbacks, Secure Transport read callback is
+ * expected to read *exactly* the requested number of bytes, not just as much
+ * as it can, and any other case would be considered a failure.
+ *
+ * This behavior is actually not specified in the Apple documentation, but is
+ * required for things to work correctly (and incidentally, that's also how
+ * Apple implements it in its projects at opensource.apple.com).
+ */
 static OSStatus read_cb(SSLConnectionRef conn, void *data, size_t *len)
 {
 	git_stream *io = (git_stream *) conn;
+	OSStatus error = noErr;
+	size_t off = 0;
 	ssize_t ret;
-	size_t left, requested;
 
-	requested = left = *len;
 	do {
-		ret = git_stream_read(io, data + (requested - left), left);
+		ret = git_stream_read(io, data + off, *len - off);
 		if (ret < 0) {
-			*len = 0;
-			return -1;
+			error = -36; /* "ioErr" from MacErrors.h which is not available on iOS */
+			break;
+		}
+		if (ret == 0) {
+			error = errSSLClosedGraceful;
+			break;
 		}
 
-		left -= ret;
-	} while (left);
+		off += ret;
+	} while (off < *len);
 
-	*len = requested;
-
-	if (ret == 0)
-		return errSSLClosedGraceful;
-
-	return noErr;
+	*len = off;
+	return error;
 }
 
 ssize_t stransport_read(git_stream *stream, void *data, size_t len)


swisspol:
As noted in #3206 current libgit2 unit tests don't appear to test the Secure Transport code path. However the functional tests I have on top of libgit2 do test it during HTTPS cloning for instance.

src/stransport_stream.c
```diff
@@ -114,14 +114,20 @@ static OSStatus write_cb(SSLConnectionRef conn, const void *data, size_t *len)
...
+		if (ret <= 0) { /* There was an error or nothing could be written because stream is closed */
+			*len = 0;
+			return errSSLClosedAbort;
+		}
+
+		left -= ret;
+	} while (left);
 
-	*len = ret;
+	*len = requested;
```
carlosmn: This change should not be necessary. The streams return when there is an error or when everything is written.

src/stransport_stream.c
```diff
@@ -150,9 +156,9 @@ static OSStatus read_cb(SSLConnectionRef conn, void *data, size_t *len)
 	requested = left = *len;
 	do {
 		ret = git_stream_read(io, data + (requested - left), left);
-		if (ret < 0) {
+		if (ret <= 0) { /* There was an error or nothing could be read because stream is closed */
 			*len = 0;
-			return -1;
+			return errSSLClosedAbort;
```
carlosmn: If we read 0, it should mean that the other side closed. Why is it an abort?

src/stransport_stream.c
swisspol: Good point, I just copy-pasted the implementation of `read_cb()` where it turns out it's not needed either.

src/stransport_stream.c
swisspol: Originally, I modified this function to behave like "read up to *len bytes from the connection" i.e. get rid of the `do ... while`. It worked the same way than the write function: "write up to *len bytes to the connection". In both cases if fewer bytes were handled, it would still return `noErr`.

But my functional tests failed when cloning a repo over HTTPS. So I figured, even though the docs do not mention that, maybe the behavior should be: "read or write this *exact* number of bytes to the connection or fail entirely if it can't". That's what you had done initially (why BTW?). I restored the `do ... while` in the read function and added the same behavior to the write function.

If you assume that's the desired behavior, and considering 0 can only be returned if fewer than `*len` were read, then the function should return an error. I picked `errSSLClosedAbort` for both read and write functions as it seemed the most sensitive among the possible SSL error codes.

swisspol: BTW if you look at the use of this very API by Apple at (http://www.opensource.apple.com/source/Security/Security-55471/libsecurity_ssl/regressions/ssl-45-tls12.c), you will notice `SocketRead()` is returning an error when reading 0 bytes. And both `SocketRead()~ and `SocketWrite()` use loops to read / write exactly the passed-in number of bytes.

swisspol: The above Apple link is actually from an older Security project version and is not around in the version shipping on 10.10. However, there's another use of the API at http://www.opensource.apple.com/source/Security/Security-55179.13/libsecurity_ssl/sslViewer/ioSock.c which has socket wrapper implementations of `SocketRead()` and `SocketWrite()` to be used with SSLSetIOFuncs().

swisspol: Finally you can see how VLC does it here: https://github.com/videolan/vlc/blob/8a7b1417015346444b6d0a329a95a1dada535429/modules/misc/securetransport.c.

It does return `errSSLClosedGraceful` when reading 0, so a different behavior, and I would assume their code has been tested and deployed extensively.

This whole thing is pretty messy...

swisspol: No matter what, all these uses of the API do loop for the write callback, so that's one thing to fix in libgit2.

swisspol:
*Moving line notes to actual discussion so that they don't disappear when updating the diff*

Originally, I modified this function to behave like "read up to *len bytes from the connection" i.e. get rid of the `do ... while`. It worked the same way than the write function: "write up to *len bytes to the connection". In both cases if fewer bytes were handled, it would still return `noErr`.

But my functional tests failed when cloning a repo over HTTPS. So I figured, even though the docs do not mention that, maybe the behavior should be: "read or write this *exact* number of bytes to the connection or fail entirely if it can't". That's what you had done initially (why BTW?). I restored the `do ... while` in the read function and added the same behavior to the write function.

If you assume that's the desired behavior, and considering 0 can only be returned if fewer than `*len` were read, then the function should return an error. I picked `errSSLClosedAbort` for both read and write functions as it seemed the most sensitive among the possible SSL error codes.

BTW if you look at the use of this very API by Apple at http://www.opensource.apple.com/source/Security/Security-55471/libsecurity_ssl/regressions/ssl-45-tls12.c, you will notice `SocketRead()` is returning an error when reading 0 bytes. And both `SocketRead()` and `SocketWrite()` use loops to read / write exactly the passed-in number of bytes.

The above Apple link is actually from an older Security project version and is not around in the version shipping on 10.10. However, there's another use of the API at http://www.opensource.apple.com/source/Security/Security-55179.13/libsecurity_ssl/sslViewer/ioSock.c which has socket wrapper implementations of `SocketRead()` and `SocketWrite()` to be used with SSLSetIOFuncs().

Finally you can see how VLC does it here: https://github.com/videolan/vlc/blob/8a7b1417015346444b6d0a329a95a1dada535429/modules/misc/securetransport.c.

It does return `errSSLClosedGraceful` when reading 0, so a different behavior, and I would assume their code has been tested and deployed extensively.

This whole thing is pretty messy...

PS: No matter what, all these uses of the API do loop for the write callback, so that's one thing to fix in libgit2.

swisspol:
OK, I've looked at a bunch of implementations of these callbacks (see comments above) and settled for this new version. See commit message for details.

src/stransport_stream.c
```diff
@@ -110,20 +110,38 @@ int stransport_certificate(git_cert **out, git_stream *stream)
...
+			break;
+		}
+		if (ret == 0) {
+			error = errSSLClosedGraceful;
+			break;
+		}
 
-	*len = ret;
+		off += ret;
+	} while (off < *len);
```
carlosmn: Again, this is already the behaviour we expect from our streams, so there is absolutely no need to loop over it. When `git_stream_write()` returns, it will have written all of the data it was given, or it will return an error.

src/stransport_stream.c
```diff
@@ -110,20 +110,38 @@ int stransport_certificate(git_cert **out, git_stream *stream)
...
 
-	ret = git_stream_write(io, data, *len, 0);
-	if (ret < 0) {
-		*len = 0;
-		return -1;
-	}
+	do {
+		ret = git_stream_write(io, data + off, *len - off, 0);
+		if (ret < 0) {
+			error = -36; /* "ioErr" from MacErrors.h */
```
carlosmn: Why use the magic number instead of the name?

src/stransport_stream.c
```diff
@@ -110,20 +110,38 @@ int stransport_certificate(git_cert **out, git_stream *stream)
 	return 0;
 }
 
+/*
+ * Contrary to typical network IO callbacks, Secure Transport write callback is
+ * expected to write *all* passed data, not just as much as it can, and any
+ * other case would be considered a failure.
```
carlosmn: This is not surprising in libgit2. Every write method will write everything or return a failure. It is surprising on the read side, but it's business as usual when writing.

src/stransport_stream.c
swisspol: That's what opensource.apple.com and VLC do. Importing `MacErrors.h` would require `#include <CoreServices/CoreServices.h>` which includes a ton of headers e.g. CarbonCore. Up to you.

swisspol: Actually, here's the main reason: `MacErrors.h` is simply not available on iOS.

swisspol:
All right, hopefully this is good now.

PR #3211, opened 2015-06-11
Added big 1gb crlf test.

jeffhostetler:
This test generates a 1GB text file with CRLF and tries to stage and commit it.  And then tries to clone and checkout.  This currently fails (at least on 32 bit builds) because the crlf filter pipeline tries to allocate multiple 1GB buffers to contain various clean/smudged versions of the file.

Currently, this is by design, so this test is a bit premature.


whoisj:
TDD (test driven design)? How novel :+1:

jeffhostetler:
:+1:   :-)

pks-t:
I've adopted this PR, see #4593. I started out from your code and heavily refactored it to match our current code style. As a result, I've taken ownership of the commits, but credit you in the message. I hope you don't mind. Anyway, thanks for this ancient PR ;)

PR #3216, opened 2015-06-12
Fix visibility of transaction symbol

dprofeta:
Transaction.c did not include the visibility definition of its symbol
(that are in git2/transaction.h) and so was by default hidden.

### DIFF
--- a/src/transaction.c
+++ b/src/transaction.c
@@ -13,6 +13,7 @@
 #include "reflog.h"
 #include "signature.h"
 
+#include "git2/transaction.h"
 #include "git2/signature.h"
 #include "git2/sys/refs.h"
 #include "git2/sys/refdb_backend.h"


carlosmn:
Symbol visibility has a different meaning, relating to whether it is accessible from outside a DLL, which this doesn't change. This is about the *declaration* of the transaction functions.

dprofeta:
It adds the declaration of the functions, but it is in the declaration that the visibility is defined (thanks to the GIT_EXTERN macro). 
Without the patch, I see:

```
nm  -g build/libgit2.so  | grep transaction       
000000000009f5c0 t git_transaction_commit
(...)
000000000009f7d0 t git_transaction_set_target
```
so the symbols are local and cannot be used outside the library and with the patch:
```
nm  -g build/libgit2.so  | grep transaction                                                                                            
0000000000076174 T git_transaction_commit
(...)
0000000000075b06 T git_transaction_set_target
```
The symbol are then exported and can be used outside the library.

ethomson:
Thanks for the additional explanation.  :+1: 

Appreciate the PR!

PR #3219, opened 2015-06-15
Zero out racily-clean entries' file_size

carlosmn:
This is a part of racy-git which we have not implemented, and can lead to diffs between the index and the workdir that do not include files which were modified immediately after the index was written.

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -150,6 +150,15 @@ static int checkout_notify(
 	}
 }
 
+GIT_INLINE(bool) is_workdir_base_or_new(
+	const git_oid *workdir_id,
+	const git_diff_file *baseitem,
+	const git_diff_file *newitem)
+{
+	return (git_oid__cmp(&baseitem->id, workdir_id) == 0 ||
+		git_oid__cmp(&newitem->id, workdir_id) == 0);
+}
+
 static bool checkout_is_workdir_modified(
 	checkout_data *data,
 	const git_diff_file *baseitem,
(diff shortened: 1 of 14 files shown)


carlosmn:
Some of the errors come from assuming that the data in the index is authoritative rather than a cache. Those tests which want to check for the file size either need to allow `0` though as undefined, or set up the timestamps such that it's simulating a much longer-running system.

Similarly for the oid calculation stats, we can test what we're doing (i.e. going really fast) or we need to set up the timestamps so we pretend there was a human driving those commands.

carlosmn:
The remaining test failures seem suspicious, as though the merge was only happening because of the buggy change detection. Or it could just be that we use the index in a way that we shouldn't, maybe. /cc @ethomson

carlosmn:
Looking at the diff we produce for `merge::workdir::dirty::honors_cache` it looks like failing to merge is the right thing to do.

ethomson:
Holy shit, `utimensat` is a godawful function.

ethomson:
The idea behind `merge::workdir::dirty::honors_cache` is that it emulates a badly filtered file.

The actual problem is:  if you were to do a checkout from some client that believes it's doing CRLF filtering, but you do a merge from libgit2 and do not have CRLF filtering configured, we should try very hard to *not* treat a file as dirty if it was checked out properly (filtered and placed in the index) by the upstream client.  So I added a tick there.

This did unearth an actual issue in the checkout code.  One interesting bit of behavior in checkout is that you can put the target contents in the working directory before doing a checkout and checkout will not treat that as a conflict.  We were handling this correctly when we knew the target's OID, but when we calculated the target's OID, we were *not* taking this behavior into account.  The merge test that validated this sort of crazy behavior was passing because we were racy.

I fixed this up and added a Win32 implementation of `utimes` over at https://github.com/libgit2/libgit2/tree/ethomson/racy-diff

ethomson:
Really excited to see this, by the way, thank you for taking care of this.

PR #3222, opened 2015-06-15
Fixed GIT_DELTA_CONFLICTED not returned in some cases

swisspol:
(see commit message)

### DIFF
--- a/tests/diff/index.c
+++ b/tests/diff/index.c
@@ -183,7 +183,7 @@ static void do_conflicted_diff(diff_expects *exp, unsigned long flags)
 	cl_git_pass(git_repository_index(&index, g_repo));
 
 	ancestor.path = ours.path = theirs.path = "staged_changes";
-	ancestor.mode = ours.mode = theirs.mode = 0100644;
+	ancestor.mode = ours.mode = theirs.mode = GIT_FILEMODE_BLOB;
 
 	git_oid_fromstr(&ancestor.id, "deadbeefdeadbeefdeadbeefdeadbeefdeadbeef");
 	git_oid_fromstr(&ours.id, "deadbeefdeadbeefdeadbeefdeadbeefdeadbeef");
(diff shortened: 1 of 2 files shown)


swisspol:
This is the fallout from #3139, see the discussion there. I've reviewed carefully all the diff code paths, and I think that was the only occurence of this bug.

carlosmn:
It would really help to have a test for this so it doesn't go unnoticed if it breaks again.

swisspol:
All right, rebased PR, updated message and added unit test!

tests/diff/index.c
```diff
@@ -239,3 +239,34 @@ void test_diff_index__reports_conflicts_when_reversed(void)
...
+
+	cl_git_pass(git_repository_index(&index, g_repo));
+	cl_git_pass(git_index_read_tree(index, a));
+
+	ancestor.path = ours.path = theirs.path = "file_not_in_head";
+	ancestor.mode = ours.mode = theirs.mode = 0100644;
+	git_oid_fromstr(&ancestor.id, "deadbeefdeadbeefdeadbeefdeadbeefdeadbeef");
+	git_oid_fromstr(&ours.id, "deadbeefdeadbeefdeadbeefdeadbeefdeadbeef");
+	git_oid_fromstr(&theirs.id, "deadbeefdeadbeefdeadbeefdeadbeefdeadbeef");
+	cl_git_pass(git_index_conflict_add(index, &ancestor, &ours, &theirs));
```
carlosmn: If the file does not exist in HEAD, then there won't be an 'ours' entry in the conflict, and possibly not one in 'ancestor' either.

tests/diff/index.c
```diff
@@ -239,3 +239,34 @@ void test_diff_index__reports_conflicts_when_reversed(void)
...
+	git_diff *diff;
+	const git_diff_delta *delta;
+
+	git_tree *a = resolve_commit_oid_to_tree(g_repo, a_commit);
+
+	cl_git_pass(git_repository_index(&index, g_repo));
+	cl_git_pass(git_index_read_tree(index, a));
+
+	ancestor.path = ours.path = theirs.path = "file_not_in_head";
+	ancestor.mode = ours.mode = theirs.mode = 0100644;
```
carlosmn: We have `GIT_FILEMODE_BLOB` so we don't sprinkle the code with numbers.

swisspol:
Rebased and fixed.

ethomson:
Nice.  Thanks for the unit test here, this is a nice fix.

PR #2280, opened 2014-04-18
Patch application

ethomson:
This PR is attempting to take an initial bite out of patch application, a la `git-apply`.  My goal is to parse a git-style patch file created with libgit2 or git.git into a `git_patch` and be able to apply a `git_patch` to a buffer.  This does not yet open up any public API for application, just internal methods; I was hoping to get some eyeballs on this and perhaps even merge it before I did something like applying patches into the working directory.

Things left todo:

- [ ] Binary patches
- [ ] Traditional-format patches (should just need to parse their headers)
- [ ] Improved whitespace tolerance during application
- [ ] Context reduction during application

PR #3018, opened 2015-03-24
Stash apply

ethomson:
A modification on #2705.  I made some notes in that PR, though actually acting on those notes was quite a bit more challenging than I had hoped.  This introduces two lower level ideas:

* `git_merge__iterators`, to allow one to merge a tree and an index and get an index out as a result.  (Interestingly `git_merge` had this broken out as a helper function at one point, and it was brought back in.  So now it's a helper function again.)  This allows us to merge the results of another merge, without having to write it out to a tree.

* `git_index_read_index`, which behaves like `git_index_read_tree` except that it reads an index into the index.  This maintains the stat cache of the existing items.

Sorry about the long delay in finishing this up.  Stash application is completely and totally ridiculous (put these files on disk and these in the index _unless_ this happens or this), as I'm sure @swisspol can attest to.  I think that with the two basic functions mentioned above that the `stash_apply` can be a bit simpler and not necessarily have to write objects for the intermediate steps.

PR #3223, opened 2015-06-15
Reading patch files

ethomson:
Moved #2280 over here and based it on master.  With the refactored binary diff handling, I can proceed on binary application.

This PR is attempting to take an initial bite out of patch application, a la `git-apply`.  My goal is to parse a git-style patch file created with libgit2 or git.git into a `git_patch` and be able to apply a `git_patch` to a buffer.  This does not yet open up any public API for application, just internal methods; I was hoping to get some eyeballs on this and perhaps even merge it before I did something like applying patches into the working directory.

- [x] Binary patches
- [ ] ~~Traditional-format patches (should just need to parse their headers)~~
- [ ] ~~Improved whitespace tolerance during application~~
- [ ] ~~Context reduction during application~~
- [x] Handle abbreviated OIDs better


swisspol:
Not trying to hijack this this discussion or anything, but before going further with support patches, since it's related, could you first fix the regressions in libgit2's stash apply implementation compared to Git CLT as discussed at the end of #3018? :wink:

ethomson:
So I think that we should review this now, and not worry about the three bullet points that I had previously listed, all of which have to do with the ability to apply patches that were created by old (non-git) `diff`.  I think that handling those patches would be very nice indeed, but not critical and we should instead be focused on adopting some APIs that would allow us to apply patches that are created by ourselves or by git.git.

Broadly speaking, these changes lay the groundwork for apply patches that are created by (for example) our `git_diff_*` or the `git-diff` command.  We parse patches for a single file, reading the changes as hunks, and put it into an internal structure that is common for both the patches that we parse and the patches that we create from diffs.  The `git_patch_*` APIs remain unchanged after this refactoring (yay) and can now operate on both types of patches:  those that we computed internally and those that were parsed from the outside world.  This was nice as it allows us to take some corpus of patch files that were created by `git-diff`, parse them, then write them back out, to ensure that we are parsing correctly *and* writing patches correctly.  (In many cases, we were not, so there were a number of bugs fixed around things like exotic filenames and dealing with changes that *only* changed a mode or a filename but no content.)

I'm really happy with the "guts" of this, though diff-based patches and the diff mechanism are still fairly intertwined in a way that I could not easily refactor in the time allotted.  (And I allotted myself *a lot* of time, this has been cooking off and on for a year and a half.  Mostly "off", but still.  Sheesh.)

But the APIs need some work here.  I don't think that the APIs are hard, here, just the terminology.  Some things to point out:

Our `git_patch` object applies to a *single file* - eg the changes to make a preimage `foo.c` into a postimage `foo.c`.  When diffing a collection of files, you would get a collection of patches, one for each modified file.  This makes the naming of functions here a little tricky to talk about, as I think the general use case that we want to support is to take a patch file (which could contain multiple `git_patch`es, in our parlance) and apply all the changes to the working directory.

I think that broadly speaking, we will need two objects - a `git_patch` that contains the changes from one file to another, and a collection of `git_patch`es - let's call it a `git_patchlist` for the sake of argument.  I think that the very compelling APIs will be:

    int git_patchlist_from_file(git_patchlist **out, const char *filename);
    int git_patchlist_from_buffer(git_patchlist **out, const char *buf);
    int git_patchlist_apply_to_workdir(git_repository *repo, git_patchlist *patchlist);

There are interesting other APIs that might surface, like pulling patches out of a patchlist, and perhaps applying a `git_patch` to a buffer.  Another options is to rename `git_patch` (kaboom) into something else that suggests its file scope, though I don't really know what because `git_patchfile` is even worse.  :)

I actually think that the best course of action here is to drop the new `git_patch_from_patchfile` API into an internal and worry about the public API later.  We can merge all the work here which will let us take advantage of the numerous bugfixes in existing APIs and we can worry about the new public APIs for parsing patches and apply patches in a separate PR.

Thoughts?

carlosmn:
I think it makes sense to restrict ourselves to git-produced patches; we shouldn't try to be a general replacement for patch(1).

Regarding the naming for `git_patch` being a particular file, that naming is somewhat unfortunate, but meh. As for what to call a collection of these, when create it, we call it a `git_diff`. Is there a reason we can't use that for the patches we read in?

This is a generous number of commits, so it's bound to take some time to review. Thanks for sticking to it.

ethomson:
>  As for what to call a collection of these, when create it, we call it a git_diff.  Is there a reason we can't use that for the patches we read in?

That's true.  I had initially not wanted to tackle that, because it would mean making `git_diff` abstract enough to deal with both types of data.  Which is quite scary.  But it worked out well enough for `git_patch` that this would perhaps be possible.  Let me ponder this some more later.

ethomson:
A couple of quick additions to squash some warnings and memory leaks.  While doing that, I realized that the single `git_patch` handling was going to be used only as part of a large patchfile handling, and so only created pointers to memory within that larger buffer.  Except that if we expose `git_patch_from_patchfile`, that will be caller memory.  So I added a commit to dup the input.

For now, I'm just dup'ing the whole damned patch file, which is a minor amount of additional memory (just some paths and some prefixes) but reduces memory fragmentation or the need to pool or something else silly.

I don't imagine that it will stick around in its current form for very long, anyway.

src/apply.c
```diff
@@ -0,0 +1,354 @@
...
+	git_diff_line *line;
+	const char *start, *end;
+
+	memset(out, 0x0, sizeof(patch_image));
+
+	git_pool_init(&out->pool, sizeof(git_diff_line));
+
+	for (start = in; start < in + in_len; start = end) {
+		for (end = start; end < in + in_len && *end != '\n'; end++)
+			;
```
carlosmn: Do we maybe wanna use `p_strnlen()` here instead of open-coding it?

carlosmn: Sorry, not `p_strnlen()` but `memchr()`.

src/apply.c
```diff
@@ -0,0 +1,354 @@
...
+
+		if (end < in + in_len)
+			end++;
+
+		line = git_pool_mallocz(&out->pool, 1);
+		GITERR_CHECK_ALLOC(line);
+
+		git_vector_insert(&out->lines, line);
```
carlosmn: This line can also fail on OOM.

src/apply.c
```diff
@@ -0,0 +1,354 @@
...
+		}
+	}
+
+	return match;
+}
+
+static bool find_hunk_linenum(
+	size_t *linenum,
+	patch_image *image,
+	patch_image *preimage)
```
carlosmn: This `in out` param feels odd. Do we do this elsewhere in the library? I'd rather see an input and output param if there's no predecent for this.

src/apply.c
```diff
@@ -0,0 +1,354 @@
...
+	git_patch *patch,
+	git_patch_hunk *hunk)
+{
+	patch_image preimage = PATCH_IMAGE_INIT, postimage = PATCH_IMAGE_INIT;
+	size_t line_num, i;
+	int error = 0;
+
+	for (i = 0; i < hunk->line_count; i++) {
+		git_diff_line *line =
+			git_array_get(patch->lines, hunk->line_start + i);
```
carlosmn: Do we know that this won't go out of bounds and return `NULL`?

src/apply.c
```diff
@@ -0,0 +1,354 @@
...
+
+		if (line->origin == GIT_DIFF_LINE_CONTEXT ||
+			line->origin == GIT_DIFF_LINE_DELETION)
+			git_vector_insert(&preimage.lines, line);
+
+		if (line->origin == GIT_DIFF_LINE_CONTEXT ||
+			line->origin == GIT_DIFF_LINE_ADDITION)
+			git_vector_insert(&postimage.lines, line);
```
carlosmn: Both of these can OOM.

src/apply.c
```diff
@@ -0,0 +1,354 @@
...
+	unsigned int *mode_out,
+	const char *source,
+	size_t source_len,
+	git_patch *patch)
+{
+	char *filename = NULL;
+	unsigned int mode = 0;
+	int error = 0;
+
+	assert(contents_out && filename_out && mode_out);
```
carlosmn: No assert for the `source` or `patch`?

src/vector.h
```diff
@@ -92,6 +92,9 @@ void git_vector_remove_matching(
 	void *payload);
 
 int git_vector_resize_to(git_vector *v, size_t new_length);
+int git_vector_grow_at(git_vector *v, size_t idx, size_t grow_len);
+int git_vector_shrink_at(git_vector *v, size_t idx, size_t shrink_len);
```
carlosmn: While this name has a nice symmetry to `_grow_at()`, isn't this really `git_vector_remove_range()`?

carlosmn: Similarly, I think `_grow_at()` is maybe better called `_create_hole()` since we typically use "grow" to mean allocate but not consider it used (but let me know if I'm misremembering)

include/git2/diff.h
```diff
@@ -479,14 +487,15 @@ typedef int(*git_diff_binary_cb)(
...
-typedef struct {
-	int    old_start;     /**< Starting line number in old_file */
-	int    old_lines;     /**< Number of lines in old_file */
-	int    new_start;     /**< Starting line number in new_file */
-	int    new_lines;     /**< Number of lines in new_file */
-	size_t header_len;    /**< Number of bytes in header text */
-	char   header[128];   /**< Header text, NUL-byte terminated */
-} git_diff_hunk;
+typedef struct git_diff_hunk git_diff_hunk;
+struct git_diff_hunk {
```
carlosmn: Any particular reason to split off the struct and type definitions? IIRC the doc parser does not like this (and also me stylistically)

include/git2/patch.h
```diff
@@ -267,6 +267,32 @@ GIT_EXTERN(int) git_patch_to_buf(
...
+ * @param patchfile The contents of a patch file
+ * @param patchfile_len The length of the patch file
+ * @param opts The git_patch_options
+ * @return 0 on success, <0 on failure.
+ */
+GIT_EXTERN(int) git_patch_from_patchfile(
+	git_patch **out,
+	const char *patchfile,
+	size_t patchfile_len,
+	git_patch_options *opts);
```
carlosmn: Using "file" here makes it sound like we're going to read something from disk. Should we call it something like `git_patch_from_buffer()` or `_new_frombuffer()` to match what we do for other types?

src/patch_parse.c
```diff
@@ -0,0 +1,1025 @@
...
+{
+	ctx->line += char_cnt;
+	ctx->remain -= char_cnt;
+	ctx->line_len -= char_cnt;
+}
+
+static int parse_advance_expected(
+	patch_parse_ctx *ctx,
+	const char *expected,
+	size_t expected_len)
```
carlosmn: It'd be nice to have a version of this function which we can use for static strings so we don't have to count by hand in order to have the value be known at compile time, like we do with `git_buf_PUTS()`.

src/patch_parse.c
```diff
@@ -0,0 +1,1025 @@
...
+}
+
+static int parse_header_traditional(
+	git_patch_parsed *patch, patch_parse_ctx *ctx)
+{
+	GIT_UNUSED(patch);
+	GIT_UNUSED(ctx);
+
+	return 1;
+}
```
carlosmn: What does this do? I only see one use just below, and it doesn't seem to be expecting a counter; or is it a boolean + error?. Either way a comment would be good here as to why there's this function that doesn't do anything. Is it so we can parse other kinds of headers in the future?

src/path.c
```diff
@@ -306,6 +306,23 @@ int git_path_join_unrooted(
...
 }
 
+void git_path_squash_slashes(git_buf *path)
+{
+	char *p, *q;
+
+	if (path->size == 0)
+		return;
+
+	for (p = path->ptr, q = path->ptr; (*p = *q); p++, q++) {
```
carlosmn: I'm not sure how I feel about the `(*p = *q)`. It is concise, but it also looks like you missed an `=`. Since we have the braces and such, would you be against changing the condition to `*q` and putting the `*p = *q` in the body of the loop?

src/util.c
```diff
@@ -66,6 +66,12 @@ int git_strarray_copy(git_strarray *tgt, const git_strarray *src)
 
 int git__strtol64(int64_t *result, const char *nptr, const char **endptr, int base)
 {
+
+	return git__strtonl64(result, nptr, (size_t)-1, endptr, base);
+}
+
+int git__strtonl64(int64_t *result, const char *nptr, size_t nptr_len, const char **endptr, int base)
```
carlosmn: Since the length refers to the input string, would this follow the convention more closely if it were `p_strntol64()`? And the same for the 32-bit version?

src/util.c
```diff
@@ -321,6 +333,31 @@ char *git__strsep(char **end, const char *sep)
...
+
+const char *git__strnchr(const char *str, size_t len, char c)
+{
+	for (; len; len--) {
+		if (str[len] == c)
+			return &str[len];
+	}
+
+	return NULL;
+}
```
carlosmn: How is this different from `memchr()` other than the return type and order of arguments?

src/util.c
```diff
@@ -321,6 +333,31 @@ char *git__strsep(char **end, const char *sep)
...
+
+	while (buffer_len--) {
+		line_len++;
+
+		if (*buffer++ == '\n')
+			break;
+	}
+
+	return line_len;
+}
```
carlosmn: Can we also use `memchr()` here to find the end of the line?

src/apply.c
```diff
@@ -0,0 +1,354 @@
...
+	if (binary_file->datalen == 0)
+		return git_buf_put(out, source, source_len);
+
+	error = git_zstream_inflatebuf(&inflated,
+		binary_file->data, binary_file->datalen);
+
+	if (!error && inflated.size != binary_file->inflatedlen) {
+		giterr_set(GITERR_PATCH, "inflated delta does not match expected length");
+		git_buf_free(out);
+		error = -1;
```
carlosmn: There's that fancy `apply_err` macro-with-comma-operator at the top, why not use it in this function but in the earlier ones? Seems we should be consistent with one or the other method (not that it's that big a deal, but y'know, consistency)

src/patch_parse.c
```diff
@@ -0,0 +1,1025 @@
...
+static int parsed_patch_binary_side(
+	git_diff_binary_file *binary,
+	patch_parse_ctx *ctx)
+{
+	git_diff_binary_t type = GIT_DIFF_BINARY_NONE;
+	git_buf base85 = GIT_BUF_INIT, decoded = GIT_BUF_INIT;
+	git_off_t len;
+	int error = 0;
+
+	if (ctx->line_len >= 8 && memcmp(ctx->line, "literal ", 8) == 0) {
```
carlosmn: In this area of the patches, I'm wondering that, since we've already checked that one side is long enough, why use `memcmp()` here instead of `strcmp()` which would allow us to remove the manual length specification. I admit I don't quite know how well compiler optimise the case for a static string, though, which might be a valid reason to do this.

src/patch_parse.c
```diff
@@ -0,0 +1,1025 @@
...
+	while (ctx->line_len) {
+		char c = ctx->line[0];
+		size_t encoded_len, decoded_len = 0, decoded_orig = decoded.size;
+
+		if (c == '\n')
+			break;
+		else if (c >= 'A' && c <= 'Z')
+			decoded_len = c - 'A' + 1;
+		else if (c >= 'a' && c <= 'z')
+			decoded_len = c - 'a' + 26 + 1;
```
carlosmn: This `26` looks like it's the length of the basic English alphabet, but it still feels rather magic. Maybe we can put it as `('z' - 'a') + 1` or have a comment?

tests/patch/parse.c
```diff
@@ -0,0 +1,33 @@
...
+
+	cl_assert_equal_s(delta->old_file.path, "file.txt");
+	cl_assert(delta->old_file.mode == 0100644);
+	cl_assert_equal_i(7, delta->old_file.id_abbrev);
+	git_oid_nfmt(idstr, delta->old_file.id_abbrev, &delta->old_file.id);
+	cl_assert_equal_s(idstr, "9432026");
+	cl_assert_equal_i(0, delta->old_file.size);
+
+	cl_assert_equal_s(delta->new_file.path, "file.txt");
+	cl_assert(delta->new_file.mode == 0100644);
```
carlosmn: These should use `GIT_FILEMODE_BLOB` no?

src/patch_parse.c
ethomson: Yeah, I was planning on adding support for this but it clearly needs to be removed for now.

src/patch_parse.c
ethomson: We'd have to use `strncmp`, since we're looking for a prefix of `ctx->line`.  My thinking is that `memcmp` is better suited to this than `strncmp` since we know we won't have a NUL in `ctx->line`.

ethomson:
Thanks for the review, @carlosmn .  Very helpful stuff and I have addressed your concerns.

ethomson:
I renamed `patch_diff` to `patch_generated` to be more clear about what it is.  (Also, because I foresee a `diff_generated` coming soon to a pull request near you.)

Anyway, I rebased this branch on master just to be nice and clean, but I think this might be ready to :ship:

ethomson:
Ignore my previous comments.  I want to shave this yak a little bit more before we merge it.

src/patch_parse.c
carlosmn: The main reason was so we don't write the string length by hand in the call, though that'd be more of a job for `git__prefixcmp()`. Nbd though.

ethomson: Aha - sorry to have misunderstood. Yes, I do like that better.  Will fix.

GrahamTheCoder:
@ethomson Looks like you're still doing awesome stuff both fixing bugs and improving code quality on this branch. Any vague ideas on how close it is to done (number of months granularity)?

ethomson:
I changed the name of this PR to "Reading patch files", so that we could consider merging it without anybody getting the wrong idea that the application part is also done.  At this point, this pull request provides the ability to read a patch file that was created by ourselves or by `git` itself, and produces a `git_diff` that represents the contents.  (Reading files created by the Unix `diff` command is not supported.)

Internally, we now have two types of `git_diff`s, those that are created by computing the state of the world, and those that are parsed out of a patch file.  We test a variety of data produced by git.git, and also test that we parse the patch files we create into a compatible format, and vice-versa.

There _are_ bits here around taking the preimage of a file and applying a `git_delta` to it to produce a postimage, but I have not yet wired up the bits around that to do it for an entire `git_diff`, which is why I changed the name of this PR to "reading patch files" from "patch application".  There's not a _ton_ left to do there in the easy case, but there will be a bunch of fun around things like applying patches when there are modified files in the workdir that are going to be terribly tedious.  It would be nice to get this PR 🚢 _finally_ even if we don't support that bit yet.

As a result, this is not yet super useful for end users (they can read a patch file and then...  print it back out, I guess) but it's been hanging around for quite a long time.

ethomson:
I just rebased this on master, which was a bit of a bummer.  @carlosmn if you have a minute, a set of 👀 would be great.

(I think that the only new commits from your last comments are since d5d24dd)

ethomson:
Oops, ignore that last comment.  I meant aa4bfb3 after this latest fixup.

(I accidentally reverted 153fde5 in that rename, and forgot to reapply it manually in my rebase.  Ugh.)

tests/diff/parse.c
```diff
@@ -65,23 +65,30 @@ void test_diff_parse__invalid_patches_fails(void)
...
 static void test_tree_to_tree_computed_to_parsed(
-	const char *sandbox, const char *a_id, const char *b_id)
+	const char *sandbox, const char *a_id, const char *b_id,
+	uint32_t diff_flags, uint32_t find_flags)
 {
 	git_repository *repo;
 	git_diff *computed, *parsed;
 	git_tree *a, *b;
 	git_diff_options opts = GIT_DIFF_OPTIONS_INIT;
+    git_diff_find_options findopts = GIT_DIFF_FIND_OPTIONS_INIT;
```
carlosmn: Odd indentation

carlosmn:
I took a quick look and things seems sensible (other than that one piece of indentation :smiling_imp:) so  I think we should merge it and let is have wider usage.

hackhaslam:
I'd like to see this merged. I've been testing the patch application part. My use case is partial application of patches for staging part of a file. It works well with a small change the addition of a per hunk callback. I'll propose a pull request after this is merged.

PR #3225, opened 2015-06-16
remote: return EINVALIDSPEC when given an empty URL

carlosmn:
This is what we used to return in the settter and there's tests in
bindings which ask for this. There's no particular reason to stop doing
so.

---

This seems like the right place to put it, though I don't know if anybody has a better idea.

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -167,14 +167,18 @@ static int get_check_cert(int *out, git_repository *repo)
 
 static int canonicalize_url(git_buf *out, const char *in)
 {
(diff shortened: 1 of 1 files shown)


ethomson:
I think it's an ok place (though maybe a few lines up for the sake of poor old VS 2012 and its problems with decls in the middle of a block.)

carlosmn:
Do you mean a few lines down then? The char there is a bit annoying as we only use it on Windows.

ethomson:
That whole function is annoying.

PR #3226, opened 2015-06-17
racy-git, the missing link

carlosmn:
It turns out that the description in racy-git is deceptively simple, and the implementation for the smudging of racily-clean entries goes all the way to performing a diff check which we were not doing.

---

I'm not too thrilled with the use of `sleep()` but the alternative is changing the timestamp for both the index and file in unison, which we may actually want to do, but it's good enough for PoC.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,9 @@ support for HTTPS connections insead of OpenSSL.
   and `git_diff_buffers` now accept a new binary callback of type
   `git_diff_binary_cb` that includes the binary diff information.
 
+* The race condition mitigations described in `racy-git.txt` have been
+  implemented.
+
 ### API additions
 
 * The `git_merge_options` gained a `file_flags` member.
--- a/src/diff.c
+++ b/src/diff.c
@@ -816,10 +816,10 @@ static int maybe_modified(
 	} else if (git_oid_iszero(&nitem->id) && new_is_workdir) {
 		bool use_ctime = ((diff->diffcaps & GIT_DIFFCAPS_TRUST_CTIME) != 0);
 		bool use_nanos = ((diff->diffcaps & GIT_DIFFCAPS_TRUST_NANOSECS) != 0);
+		git_index *index;
+		git_iterator_index(&index, info->new_iter);
 
 		status = GIT_DELTA_UNMODIFIED;
-
-		/* TODO: add check against index file st_mtime to avoid racy-git */
 
 		if (S_ISGITLINK(nmode)) {
 			if ((error = maybe_modified_submodule(&status, &noid, diff, info)) < 0)
@@ -839,7 +839,8 @@ static int maybe_modified(
 			 !diff_time_eq(&oitem->ctime, &nitem->ctime, use_nanos)) ||
 			oitem->ino != nitem->ino ||
 			oitem->uid != nitem->uid ||
-			oitem->gid != nitem->gid)
+			oitem->gid != nitem->gid ||
+			(index && nitem->mtime.seconds >= index->stamp.mtime))
 		{
 			status = GIT_DELTA_MODIFIED;
 			modified_uncertain = true;
--- a/src/index.c
+++ b/src/index.c
@@ -688,20 +688,59 @@ int git_index__changed_relative_to(
 	return !!git_oid_cmp(&index->checksum, checksum);
 }
 
+static bool is_racy_timestamp(git_time_t stamp, git_index_entry *entry)
+{
+	/* Git special-cases submodules in the check */
+	if (S_ISGITLINK(entry->mode))
+		return false;
+
+	/* If we never read the index, we can't have this race either */
+	if (stamp == 0)
+		return false;
+
+	/* If the timestamp is the same or newer than the index, it's racy */
+	return ((int32_t) stamp) <= entry->mtime.seconds;
+}
+
 /*
  * Force the next diff to take a look at those entries which have the
  * same timestamp as the current index.
  */
-static void truncate_racily_clean(git_index *index)
+static int truncate_racily_clean(git_index *index)
 {
 	size_t i;
+	int error;
 	git_index_entry *entry;
 	git_time_t ts = index->stamp.mtime;
+	git_diff_options diff_opts = GIT_DIFF_OPTIONS_INIT;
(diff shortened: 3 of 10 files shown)


carlosmn:
This comes with yet another merge failure, @ethomson I'm going to leave this in your capable hands.

arrbee:
I know I've long since given up much right to comment on such things, but pulling the entire typedef of the fs and workdir iterators into `iterator.h` and then accessing the internal structure just to pull out a single field doesn't seem like the right design to me. There is precedent for accessor APIs that only work with a particular type of iterator (such as `git_iterator_current_workdir_path`) that would be less invasive, I think.

carlosmn:
It did seem ugly, but at the time I was trying to get it to work. Since we do have precedent, I'll look into having `git_iterator_index()` which should also make the diff quite a bit more elegant.

carlosmn:
The history needs some love, but this should behave the same way as git.  The check for racily clean entries does a diff for each candidate. It may be faster to collect them all and do a single diff, but having a large amount of racily-clean entries is rather an edge case.

ethomson:
I'll take a look at that merge test shortly.  I'm surprised it's so brittle.

In the meantime, we'll probably need a `p_sleep` type of creation on Windows, since it only offers `Sleep(millis)`.

carlosmn:
I thought we already had it, but its the status examples which has this define. I think I'd rather have us set the timestamps manually long-term, rather than relying on a second of sleep, though.

src/blob.h
```diff
@@ -17,6 +17,10 @@ struct git_blob {
 	git_odb_object *odb_object;
 };
 
+static git_oid empty_blob = {{ 0xe6, 0x9d, 0xe2, 0x9b, 0xb2, 0xd1, 0xd6, 0x43, 0x4b, 0x8b,
```
ethomson: Might be nice to drop this definition in `odb.c` and have it use this one.

carlosmn: I did, it's further down the diff.

ethomson: Hah.  Thanks.

carlosmn:
Tracing a bit what `differently_filtered` is doing, it seems to be running afoul of the fixed racy protections done here. When we write out the index, we check for racy timestamps, which we in fact set the index to have. In this case, we check whether there is a difference between the contents of what we have in the entry and what we have in the workdir. We do have a difference here, so the entry gets smudged in order to avoid the next status/diff to believe that the file is unchanged.

I've changed that test such that we mess with the timestamps so much that we don't detect a change, but I don't believe that this represents a realistic situation anymore. Even if the first change is sensible in order to simulate an implementation which performs a different form of filtering, the second one is stupid, because it's timing-dependent. If we were to see the situation, we *would* detect it as a difference, because the entry/file/index timestamps would all match, and we need to protect ourselves from someone doing

    git add foo
    echo lolchange >foo
    git status || git merge somebranch || ...

which makes us have to look at the change.

Using nanosecs would remediate this a bit on systems where we can trust it, but that's asking too much, I think, as git on Windows does not use this, and we can't use this on OS X at all IIRC.

Which is to say that by hacking the test so much that we pass, it feels like we're mocking the situation so much that we're not really testing anything anymore.

carlosmn:
I've added extra checks so we avoid doing a diff and smudge of files for which the stat data is different, though I wonder if we have something in diff which would tell us already whether we can detect that a file is different just by looking at the stat.

ethomson:
> If we were to see the situation, we would detect it as a difference, 

Well, we *should* detect it as a difference if the file was written at the same time as the index.  And we *shouldn't* detect is as a different if the file was *not* written at the same time.  You can certainly envision a case where checkout takes > 1 second, and some files are not eligible for being rescanned due to raciness, and some (those written in the same second as the index) are.

The goal of this test is really to ensure that merge treats the files as clean/dirty identically to status, to avoid regressions there.

The scenario is that one user has `core.autocrlf=false`, makes some change on Windows, another user with `core.autocrlf=true` checks that file out.  Now - if that file was *not* written at the same time as the index, the racy protections don't kick in and the file is clean.  (Prior to the raciness fixes, this was always clean.)

At this point, our status and `git status` will both show this file as clean, but the merge prior to c0b10c25e02c6922276567600988fb2b85aeede1 would show the file as dirty, as it did not correctly take the index into account when looking at the workdir contents.

This test was really to ensure that we never differed in calculating cleanness in merge vs status.

It's possible that given the additional complexities that raciness throws into the mix that it is no longer viable to maintain this test and - hey - we probably have it right now!  But this is bug was exceptionally costly for us, both in terms of support and in terms of mindshare.  (See, eg http://stackoverflow.com/questions/27458236/visual-studio-2013-does-not-offer-to-do-merge-on-git-pull though there are hundreds of questions addressing this topic.)

So if it *is* viable, I would certainly like to keep this test.

carlosmn:
What the test is proving right now is that if the index' timestamp is ahead of the files, we won't look at the contents before deciding that there's no difference. This should be the same as status does, but we still have the problem that this is timing-dependent. If git happens to update the index in some operation after the checkout, then we'd do this, but even if an operation takes over a second, some of the files will have the same timestamp as the index and if we have a different autocrlf setting from the installation which wrote the files, then we'd still need to figure out if it changed just before or after the index, so we'd have to look at the contents, so we might still fail to merge.

ethomson:
I'm beyond tired of arguing about this.  Just delete the test and I'll put some unit tests in VS instead.

ethomson:
This is looking pretty good to me; how are you feeling about it?

carlosmn:
I think that "index: don't smudge entries with different stat" is completely redundant, as diff does already take this into account, and I'm just re-implementing the logic. That needs to go.

carlosmn:
It turns out that that commit wasn't quite redundant, but it's an optimisation for an edge-case which I believe we can solve much better than by having another copy of the comparison functions. I'll open a PR with a test to show a situation in which we don't need to smudge the entry but currently do (as it would be a failing test and we don't really have a way to express that's expected in clar).

I just remembered another edge case (smuding a file and then quickly making that file empty, I don't think we handle that) which I'll push up after breakfast.

carlosmn:
I'm feeling :shipit: about it atm, that edge case is handled, and there's a test for it.

PR #3227, opened 2015-06-17
Fix memory leak in tests/network/refspecs.c

jeffhostetler:


carlosmn:
It looks like the appveyor errors come from our timing in `tick_index()` being a bit simplistic.

PR #3228, opened 2015-06-17
Explicitly handle GIT_DELTA_CONFLICTED in git_diff_merge()

swisspol:
(see commit message)

### DIFF
--- a/src/diff_tform.c
+++ b/src/diff_tform.c
@@ -65,6 +65,12 @@ static git_diff_delta *diff_delta__merge_like_cgit(
 	 *  f3 = b->new_file
 	 */
 
+	/* If one of the diffs is a conflict, just dup it */
+	if (b->status == GIT_DELTA_CONFLICTED)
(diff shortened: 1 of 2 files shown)


swisspol:
Can someone restart the Travis build? Looking at the logs, seems like a CI problem.

carlosmn:
What *was* happening before? Is there some test which covers this situation?

swisspol:
Before it would create a hybrid `git_diff_delta` between the from and to ones, just because it wasn't special-cased like `GIT_DELTA_UNMODIFIED` or `GIT_DELTA_DELETED`. I'll see if I can put a test case together.

swisspol:
Added unit test and rebased. The issue was actually bigger than I originally thought as a conflicted file in the index, but not existing in the workdir, could be reported as `ADDED` after merging the diffs (HEAD -> index -> workdir) instead of `CONFLICTED`.

PR #3229, opened 2015-06-17
Fixed Xcode 6.1 build warnings

swisspol:


src/unix/posix.h
```diff
@@ -10,6 +10,7 @@
 #include <stdio.h>
 #include <dirent.h>
 #include <sys/param.h>
+#include <sys/time.h>
```
swisspol: This is needed for `utimes()` and `futimes()`

PR #3233, opened 2015-06-19
Don't propagate workdir's mode to the index during diff's update index

ethomson:
When diff updates the index, it updates based on the *workdir* mode.  It should instead preserve the original mode to ensure that symlinks on Windows are not treated as regular files (which is their workdir mode).

### DIFF
--- a/tests/status/worktree.c
+++ b/tests/status/worktree.c
@@ -1096,3 +1096,51 @@ void test_status_worktree__unreadable_as_untracked(void)
 	cl_assert_equal_i(0, counts.wrong_sorted_path);
 }
 
+void test_status_worktree__update_index_with_symlink_doesnt_change_mode(void)
+{
+	git_repository *repo = cl_git_sandbox_init("testrepo");
+	git_reference *head;
+	git_object *head_object;
+	git_index *index;
+	const git_index_entry *idx_entry;
+	git_status_options opts = GIT_STATUS_OPTIONS_INIT;
+	status_entry_counts counts = {0};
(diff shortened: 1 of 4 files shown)


carlosmn:
The status test could do with an explanation that you're expecting `link_to_new.txt` to remain as it is, as there's no mention of it on the test itself.

Are we guaranteed a rewrite of the index? Even if we are right now, there's a PR that would not guarantee this anymore, so maybe we should have a change that would make an index write necessary? AFAICT having a new file would not do that.

ethomson:
Yes you're right.  I updated this to be (I believe) resilient to both the racy changes and the update index changes that are forthcoming.

tests/status/worktree.c
```diff
@@ -1096,3 +1096,51 @@ void test_status_worktree__unreadable_as_untracked(void)
...
+
+	/* now ensure that the status's rewrite of the index did not screw
+	 * up the mode of the symlink `link_to_new.txt`, particularly
+	 * on platforms that don't support symlinks
+	 */
+	cl_git_pass(git_repository_index(&index, repo));
+	cl_git_pass(git_index_read(index, true));
+
+	cl_assert(idx_entry = git_index_get_bypath(index, "link_to_new.txt", 0));
+	cl_assert_equal_i(0120000, idx_entry->mode);
```
carlosmn: Maybe use `cl_assert(S_ISLNK(idx_entry->mode));` here? Or would that miss some failure mode?

ethomson: Fixed!

PR #3234, opened 2015-06-19
Dont update index unnecessarily

ethomson:
Now with unit test goodness.

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -664,8 +664,10 @@ int git_diff__oid_for_entry(
 		updated_entry.mode = mode;
 		git_oid_cpy(&updated_entry.id, out);
 
-		if (!(error = git_repository_index__weakptr(&idx, diff->repo)))
+		if (!(error = git_repository_index__weakptr(&idx, diff->repo))) {
 			error = git_index_add(idx, &updated_entry);
+			diff->index_updated = true;
+		}
  	}
 
 	git_buf_free(&full_path);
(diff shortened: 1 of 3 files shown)


carlosmn:
I was really hoping for fewer sleeps. I'm working on exposing the index' trailer so we should be able to read it and compare an oid instead of relying on timestamps.

ethomson:
Sleeps in tests are horrible.  That would be great.

ethomson:
Well, the problem with using the checksum is that the checksum will be updated when the contents of the file changes.

This change is attempting to ensure that we're not unnecessarily rewriting indexes when they *haven't* changed.  In that case, we're dumping an index that would have the same checksum, so it seems that the only way to identify that is to let the clock tick.

Of course we may be able to `usleep` if we have a more granular `p_stat`.

ethomson:
(Note this requires #3226 since we are currently rewriting the index in the second diff.)

carlosmn:
I think we should still be able to detect changes without using sleep. If we set the index file's timestamp forward five seconds and then instead of `touch_file` we move the file's mtime six seconds forward, we should be able to detect that the index was rewritten because stating it afterward would have changed the timestamp on disk.

Or we can move the index file's timestamp backwards, so we know we won't accidentally take long enough in the test for us to catch up to whatever much we moved it forward.

swisspol:
Thanks for writing the unit tests @ethomson!

ethomson:
Updated to manipulate timestamps and investigate the index's checksum to determine changedness.  So this should perform the same function and avoid `sleep`s.

PR #3239, opened 2015-06-21
Write modified index in git_stash_apply()

swisspol:
Same as with git_stash_save(), there's no reason not to write the index
to disk since it has been modified.


ethomson:
I agree, it was probably a mistake to remove this.  Thanks for fixing!

Issue #762, opened 2012-06-12
Support mergetag header field


arrbee:
When pulling a signed tag, recent versions of git will automatically verify the signature and create a multi-line "mergetag" header field in the merge commit with the signature information. You may want to read @gitster's [blog entry about this](http://git-blame.blogspot.com/2011/11/helping-kernel-workflow-redux.html).  You can see an example of the header [here](http://trac.edgewall.org/ticket/10676). We should make libgit2 support the correct parsing of and access to the multi-line payload.

carlosmn:
I'm going to consider this fixed with #3240. Looking at the contents and verifying them would be a wholly different effort.

PR #3240, opened 2015-06-22
commit: allow retrieving an arbitrary header field

carlosmn:
This allows the user to look up fields which we don't parse in libgit2,
and allows them to access gpgsig or mergetag fields if they wish to
check the signature.

--- 

This should be enough to consider #762 closed. We may perform some checking in the future, but for now it falls to the user.

### DIFF
--- a/src/commit.c
+++ b/src/commit.c
@@ -518,3 +518,58 @@ int git_commit_nth_gen_ancestor(
 	*ancestor = parent;
 	return 0;
 }
+
+int git_commit_header_field(git_buf *out, const git_commit *commit, const char *field)
+{
+	const char *buf = commit->raw_header;
+	const char *h, *eol;
+
+	git_buf_sanitize(out);
+	while ((h = strchr(buf, '\n')) && h[1] != '\0' && h[1] != '\n') {
+		h++;
+		if (git__prefixcmp(h, field)) {
+			buf = h;
+			continue;
+		}
+
+		h += strlen(field);
+		eol = strchr(h, '\n');
+		if (h[0] != ' ') {
+			buf = h;
+			continue;
+		}
+		if (!eol)
+			goto malformed;
+
+		h++; /* skip the SP */
+
+		git_buf_put(out, h, eol - h);
+		if (git_buf_oom(out))
+			goto oom;
+
+		/* If the next line starts with SP, it's multi-line, we must continue */
+		while (eol[1] == ' ') {
+			git_buf_putc(out, '\n');
+			h = eol + 2;
+			eol = strchr(h, '\n');
+			if (!eol)
+				goto malformed;
+
+			git_buf_put(out, h, eol - h);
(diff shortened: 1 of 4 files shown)


src/commit.c
```diff
@@ -518,3 +518,54 @@ int git_commit_nth_gen_ancestor(
...
+		if (h[0] != ' ' || !eol)
+			goto malformed;
+
+		h++; /* skip the SP */
+
+		git_buf_put(out, h, eol - h);
+		if (git_buf_oom(out))
+			goto oom;
+
+		/* If the next line starts with LF, it's multi-line, we must continue */
```
ethomson: Should this comment say "If the next line starts with SP"...?

carlosmn: Yes

src/commit.c
```diff
@@ -518,3 +518,54 @@ int git_commit_nth_gen_ancestor(
...
+	while ((h = strchr(buf, '\n')) && h[1] != '\0' && h[1] != '\n') {
+		h++;
+		if (git__prefixcmp(h, field)) {
+			buf = h;
+			continue;
+		}
+
+		h += strlen(field);
+		eol = strchr(h, '\n');
+		if (h[0] != ' ' || !eol)
```
ethomson: Should this check actually be `if (h[0] != ' ') continue; if (!eol) goto malformed;`?

It looks like if you are asking for the `par` field, then we will match the prefix and then choke because it's not followed by a space...

carlosmn: Good point, I'd forgotten about fields which are prefixes of other fields.

carlosmn: What I was trying to catch here is the case where we have something like "parentXDAEF" alone in a line, which would be a bad line, but it looks like we can't distinguish this from an oddly-named field, so if we have that instead of parent, we'll return not-found, which is probably fine.

ethomson: Yeah.  Regrettably, it seems that returning not found is probably the best we can do in that situation.

tests/commit/parse.c
```diff
@@ -418,3 +418,41 @@ committer Vicent Marti <tanoku@gmail.com> 1273848544 +0200\n\
...
+
+	cl_git_pass(git_commit_header_field(&buf, commit, "parent"));
+	cl_assert_equal_s("34734e478d6cf50c27c9d69026d93974d052c454", buf.ptr);
+	git_buf_clear(&buf);
+
+	cl_git_pass(git_commit_header_field(&buf, commit, "gpgsig"));
+	cl_assert_equal_s(gpgsig, buf.ptr);
+
+	cl_git_fail_with(GIT_ENOTFOUND, git_commit_header_field(&buf, commit, "awesomeness"));
+	cl_git_fail_with(GIT_ENOTFOUND, git_commit_header_field(&buf, commit, "par"));
```
ethomson: :+1:

PR #3062, opened 2015-04-21
Return GIT_EAUTH when authentication callbacks are missing for HTTP or S...

swisspol:
...SH

This allows clients that either do not show an authentication UI, or show one
but let the user cancel and return GIT_PASSTHROUGH, to identify such cases from
the final error code returned by libgit2.

PR #3243, opened 2015-06-22
http: return GIT_EAUTH if auth tried but failed

michaeledgar:
My code currently has to detect HTTP auth failures by checking the error string for "Unexpected HTTP status code: 401" because http (unlike ssh) does not return GIT_EAUTH. This change (should) address that.

Submitted against the master branch because I assumed changing error codes would be a breaking change inappropriate for maint.

### DIFF
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -34,6 +34,7 @@ static const char *post_verb = "POST";
 
 #define PARSE_ERROR_GENERIC	-1
 #define PARSE_ERROR_REPLAY	-2
+#define PARSE_ERROR_AUTHFAIL	-3
 
 #define CHUNK_SIZE	4096
 
@@ -359,6 +360,9 @@ static int on_headers_complete(http_parser *parser)
 					t->parse_error = PARSE_ERROR_REPLAY;
 					return 0;
 				}
+			} else {
+				giterr_set(GITERR_NET, "authentication failed, incorrect credentials");
+				return t->parse_error = PARSE_ERROR_AUTHFAIL;
 			}
 		}
 
@@ -692,6 +696,9 @@ replay:
 
 			goto replay;
 		}
+
+		if (PARSE_ERROR_AUTHFAIL == t->parse_error)
+			return GIT_EAUTH;
 
 		if (t->parse_error < 0)
 			return -1;


src/transports/http.c
```diff
@@ -359,6 +360,9 @@ static int on_headers_complete(http_parser *parser)
 					t->parse_error = PARSE_ERROR_REPLAY;
 					return 0;
 				}
+			} else {
+				giterr_set(GITERR_NET, "authentication failed, incorrect credentials");
```
ethomson: This error message is either imprecise or incorrect.  We don't know if credentials are "correct" or "incorrect", we only know that we got an Unauthorized.  The credentials could be "correct" (authenticate a particular user) but that user doesn't have authorization to the resource.

ethomson:
Should we use `GIT_EAUTH` in other places?  (eg, if you have no credentials?  what if your credentials are of a different type)?  See #3062 which also looked at this problem a little bit.

In any case, adding some new `online` unit tests would be very helpful!

src/transports/http.c
test987987: Typically a 403 is returned for authorization errors, but I can't speak to what every Git server out there does, so I don't mind just "authentication failed".

michaeledgar: Typically a 403 is returned for authorization errors, but I can't speak to what every Git server out there does, so I don't mind just "authentication failed".

carlosmn: The Git server doesn't do much here, this comes typically from the HTTP in front of it. But 403 quite explicitly does not mean that authentication failed. It means it succeeded but authorisation failed.

carlosmn:
In what situation are you seeing this? There should be no need to check the error string; libgit2 will call your code if there is a problem authenticating.

michaeledgar:
I'm using git2go. I currently see this flow:

1. Initial unauthenticated request
2. Server returns 401
3. libgit2 calls my CredentialsCallback
4. CredentialsCallback returns an incorrect plaintext user/pass
5. Server (correctly) rejects that user/pass, returns 401 again
6. libgit2 returns a generic GIT_ERROR with message "Unexpected HTTP status code: 401"

carlosmn:
#3269 has a fix which is more in line with the intent of the transport code. We should keep asking for credentials until either the user or the server gives up.

test987987:
That makes sense, Carlos. I'll look for that in the next libgit2 release. Thanks!

michaeledgar:
That makes sense, Carlos. Thanks for the heads-up - I'll look forward to the next libgit2 release :)

PR #3248, opened 2015-06-23
git_treebuilder* can be const in git_treebuilder_write()

leoyanggit:


ethomson:
I don't think that this is a good idea.  This suggests a promise that we may not be able to keep in the future.  Maybe in the future we'll want to save some cached data in the object, or maybe we'll update some hash in the treebuilder object.  Making this `const` prevents us from doing that (or changes the API later).

(I realize that we've been running roughshod changing the API, but we're doing so to make things more future proof.)

carlosmn:
I agree, it  can be made const right now, but there's nothing to say we won't want to modify it. The lack of const sends the signal that we don't promise to only performs reads on it.

leoyanggit:
Conceptually `git_treebuilder_write()` writes everything in the builder to the database. I don't understand why we would modify the treebuilder during write.

ethomson:
Well, `git_index_write` similarly writes everything in the index to the database.  You can take a look to see the myriad things that are updated in the `git_index` object as a result.

carlosmn:
If we were passing the treebuilder to a different object, we could ask it not to modify our data, but in this case the treebuilder is the subject of the method, and though we currently only read from its list of paths, it may come a time that we want to write a piece of state, such as caching the output oid. Or even more simply, we might convert to using a data structure which sorts in-place when we need to write. We currently read all of the data out of its dictionary in order to sort it, but if we replace that with one which can be asked to do sorting, we'll have to write to it. This modification is an implementation detail that should not escape. The fact that we currently read data out of it to sort it is also an implementation detail. There is nothing in the write operation that says that we do not need to modify the treebuilder or its data structures.

leoyanggit:
OK I see. Thanks @ethomson  and @carlosmn

PR #3249, opened 2015-06-23
Check the repository version

carlosmn:
It hasn't mattered that we don't check for it on open, as it has been 0 since the first release of git, way back when; but times, they're a-changing and we will soon(ish) see version 1in the wild and we must take care that we don't open those and possibly break something (or just produce uncomprehensible results).

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -32,6 +32,8 @@
 # include "win32/w32_util.h"
 #endif
 
+static int check_repositoryformatversion(git_config *config);
+
 #define GIT_FILE_CONTENT_PREFIX "gitdir:"
 
 #define GIT_BRANCH_MASTER "master"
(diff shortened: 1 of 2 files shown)


src/repository.c
```diff
@@ -510,22 +513,36 @@ int git_repository_open_ext(
...
 
-		if ((error = git_repository_config_snapshot(&config, repo)) < 0 ||
-			(error = load_config_data(repo, config)) < 0 ||
-			(error = load_workdir(repo, config, &parent)) < 0)
-			git_repository_free(repo);
-
-		git_config_free(config);
+		if (config &&
+		    ((error = load_config_data(repo, config)) < 0 ||
+		     (error = load_workdir(repo, config, &parent))) < 0)
```
swisspol: Misplaced parenthesis here: see #3251

PR #3250, opened 2015-06-23
Stash workdir correctly when added in the index, modified in the workdir

ethomson:
We do not presently stash files correctly when a file is added in the index and then modified in the working directory.  We are examining only the working directory with regard to the HEAD, and without examining the index as well (with that file staged) we treat that file as untracked and thus do not include it in the workdir tree.

Correct that by using a diff that includes the index similar to `git_diff_tree_to_workdir_with_index` but with different special cases.  (`git_diff_tree_to_workdir_with_index` is aimed at producing something like `git diff` and so collapses some cases - we differ in that we want to ensure that we still pay attention to the workdir side when the index side is deleted.)

### DIFF
--- a/src/diff_tform.c
+++ b/src/diff_tform.c
@@ -109,7 +109,8 @@ static git_diff_delta *diff_delta__merge_like_cgit(
 	return dup;
 }
 
-int git_diff_merge(git_diff *onto, const git_diff *from)
+int git_diff__merge_deltas(
+	git_diff *onto, const git_diff *from, git_diff__merge_cb cb)
 {
 	int error = 0;
 	git_pool onto_pool;
@@ -154,7 +155,7 @@ int git_diff_merge(git_diff *onto, const git_diff *from)
 			const git_diff_delta *left = reversed ? f : o;
 			const git_diff_delta *right = reversed ? o : f;
 
-			delta = diff_delta__merge_like_cgit(left, right, &onto_pool);
+			delta = cb(left, right, &onto_pool);
 			i++;
 			j++;
 		}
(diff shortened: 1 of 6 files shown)


src/diff_tform.c
```diff
@@ -185,15 +146,16 @@ int git_diff_merge(git_diff *onto, const git_diff *from)
...
+			delta = git_diff__delta_dup(f, &onto_pool);
 			j++;
 		} else {
-			delta = reversed ?
-				diff_delta__merge_like_cgit_reversed(o, f, &onto_pool) :
-				diff_delta__merge_like_cgit(o, f, &onto_pool);
+			const git_diff_delta *left = reversed ? f : o;
+			const git_diff_delta *right = reversed ? o : f;
+
+			delta = cb(left, right, &onto_pool);
```
swisspol: So when I working on this code, I wondered myself why `diff_delta__merge_like_cgit_reversed()` even existed vs just swapping the parameters to `diff_delta__merge_like_cgit()`. If you look at the implementation of `diff_delta__merge_like_cgit_reversed()`, it's actually not exactly the "reverse" of `diff_delta__merge_like_cgit` IIRC. Are you sure this is safe?

ethomson: Indeed.  It looks like there are a handful of changes to each since they were written.  I tried to make sure that the unique changes to each were reflected in the new function.  Was there something in particular that you're concerned about?

swisspol: Nothing in particular as I don't remember right now the past edge cases I had to deal when using this merge API. I do remember I spent a lot of time dealing with its peculiarities though. Anyway, looks like you have this covered already!

PR #3251, opened 2015-06-24
Fixes

swisspol:


ethomson:
Good catch, but why change the usage of a weakptr to a refcounted one?

swisspol:
Because `git_repository_config__weakptr()` is not available within test files AFAIK and the compiler complains about missing prototype.

ethomson:
Got it.  For what it's worth, we don't like clar with a build of libgit2, we recompile clar with all the libgit2 sources, so that library-internal bits are available to the tests.  So in this case we could just include `repository.h` for visibility.

Anyway, that's just an aside.  This is a nice fix.  Thanks.

swisspol:
I should have said "declared" not "available" :)

Yes, the build does complete successfully and obviously the symbol is found when running the tests since it's all in a single binary. I was however under the (incorrect?) impression that tests should mostly rely on APIs not SPIs.

ethomson:
Sure - we definitely want to test the public API, but for things like just getting the configuration (and not *testing* the configuration) then just using the internal bits seems reasonable.

PR #3254, opened 2015-06-24
Handle binary DIFFABLEness properly

ethomson:
Set `GIT_DIFF_PATCH_DIFFABLE` correctly for all files that differ, regardless of binary-ness.  It's up to the callbacks what to do about diffable files - either producing an actual binary delta diff, or simply saying "Binary files differ".

Binary files need some deeper inspection into this, since we have avoided loading their data if we knew we wouldn't need it (eg, `SHOW_BINARY` was not provided).  This means we look at the actual file sizes to determine difference, not the file map data size.  Note that we cannot do this for all deltas since some (like submodules) populate the file map data but do not actually have files backing them.

Ugh.

### DIFF
--- a/src/diff_patch.c
+++ b/src/diff_patch.c
@@ -121,6 +121,35 @@ GIT_INLINE(bool) should_skip_binary(git_patch *patch, git_diff_file *file)
 	return (file->flags & GIT_DIFF_FLAG_BINARY) != 0;
 }
 
+static bool diff_patch_diffable(git_patch *patch)
+{
+	size_t olen, nlen;
+
+	if (patch->delta->status == GIT_DELTA_UNMODIFIED)
+		return false;
+
+	/* if we've determined this to be binary (and we are not showing binary
+	 * data) then we have skipped loading the map data.  instead, query the
+	 * file data itself.
+	 */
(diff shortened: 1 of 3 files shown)


carlosmn:
FWIW, applying

```diff
diff --git a/src/diff_patch.c b/src/diff_patch.c
index 60648f8..7375e1b 100644
--- a/src/diff_patch.c
+++ b/src/diff_patch.c
@@ -327,9 +327,6 @@ static int diff_patch_generate(git_patch *patch, git_diff_output *output)
                (error = diff_patch_load(patch, output)) < 0)
                return error;
 
-       if ((patch->flags & GIT_DIFF_PATCH_DIFFABLE) == 0)
-               return 0;
-
        if ((patch->delta->flags & GIT_DIFF_FLAG_BINARY) != 0) {
                if (output->binary_cb)
                        error = diff_binary(output, patch);
```

on top of my patch also makes the test suite pass.

vmg:
Thanks for the 982398th time @ethomson. Gonna pull this into Rugged. :heart_eyes:

arthurschreiber:
:heart:

PR #3255, opened 2015-06-25
Rename FALLBACK to UNSPECIFIED

carlosmn:
Fallback describes the mechanism, while unspecified explains what the
user is thinking.


ethomson:
Dumb question: should the default value be `DEFAULT` and not `UNSPECIFIED`?

arthurschreiber:
:+1: for `DEFAULT`.

carlosmn:
The default would indeed be named `_DEFAULT` but this is not what we're expressing here. What we're expressing is that we do not want to override the value at the level which is specified. It does not mean that we're going to use the default, as there may be a setting lower down.

ethomson:
Got it.  Thanks for the clarification.

Issue #3036, opened 2015-04-08
Attaching extra information to error messages


carlosmn:
Our error reporting mechanism is currently fairly simplistic. We return an error code and store an error message and a class. It is sometimes advantageous to add more detailed information about the error which some consumers can make use of. An example would be the HTTP status which the server returned instead of making the caller code parse it out of the error string. One way to do this would be to add a pointer in the struct we return from `giterr_last()` to read this information and provide a function like

``` C
int giterr_attach(git_error_info *info);
```

by which the caller (in libgit2 or external) can attach extra information to the current/last error message. Alternatively,

``` C
int giterr_set_extra(int error_class git_error_info *info, const char *fmt, ...);
```

to do both at the same would should also work. A `giterr_clear()` would also free this structure. The structure itself would be something like

``` C
typedef struct {
    git_error_info_t type;
    char *custom_type;
    void free(git_error_info *);
} git_error_info;
```

which you can subclass in the usual C way. The string is there in case a custom backend wants to provide more detailed information; it can set type to a value such as `GIT_ERROR_INFO_CUSTOM` and it can provide a string to identify which type of custom error it has stored.

There's at least one place where we want to duplicate or store the error state locally to re-set it again (in clone, so the clean up doesn't mess with the clone error) so we need to make sure we don't double-free it, so we'd either want to refcount it or have it provide a `dup` field as well.

/cc @jamill

ethomson:
This is rather old and we didn't pursue it.  Closing this.

PR #3257, opened 2015-06-25
Errors: drop the error "class", use a string instead of a `git_error` structure

ethomson:
So...  I've never really found the error classes particularly useful.  At some point I was talking to someone (I think @carlosmn ?) and mentioned this and we talked about removing them.  Here's a spike to do so, and now we give callers error messages (in the form of a string) instead of a struct that has an integer that they likely won't do anything with.

Also moved the actual error message to be in a `git_buf` per TLS, which hopefully reduces the number of allocations that we do (we set error messages a lot that never wind up going to the caller.)

Outstanding questions while I'm breaking shit:

* Why `giterr`...  All our other everythings are `git_` prefixed - should this be `git_err` or `git_error`?
* The likely useful class was probably the one that indicated OOM.  Should there be a `GIT_ENOMEM` or similar return code?

### DIFF
--- a/examples/common.c
+++ b/examples/common.c
@@ -16,16 +16,13 @@
 
 void check_lg2(int error, const char *message, const char *extra)
 {
-	const git_error *lg2err;
 	const char *lg2msg = "", *lg2spacer = "";
 
 	if (!error)
 		return;
 
-	if ((lg2err = giterr_last()) != NULL && lg2err->message != NULL) {
-		lg2msg = lg2err->message;
+	if ((lg2msg = giterr_last()) != NULL)
 		lg2spacer = " - ";
-	}
 
 	if (extra)
 		fprintf(stderr, "%s '%s' [%d]%s%s\n",
--- a/examples/general.c
+++ b/examples/general.c
@@ -47,12 +47,12 @@
 // as an example.
 static void check_error(int error_code, const char *action)
 {
-	const git_error *error = giterr_last();
+	const char *error = giterr_last();
 	if (!error_code)
 		return;
 
 	printf("Error %d %s - %s\n", error_code, action,
-		   (error && error->message) ? error->message : "???");
+		error ? error : "???");
 
 	exit(1);
 }
--- a/examples/network/clone.c
+++ b/examples/network/clone.c
@@ -102,8 +102,8 @@ int do_clone(git_repository *repo, int argc, char **argv)
 	error = git_clone(&cloned_repo, url, path, &clone_opts);
 	printf("\n");
 	if (error != 0) {
-		const git_error *err = giterr_last();
-		if (err) printf("ERROR %d: %s\n", err->klass, err->message);
+		const char *err = giterr_last();
+		if (err) printf("ERROR: %s\n", err);
 		else printf("ERROR %d: no detailed info\n", error);
 	}
 	else if (cloned_repo) git_repository_free(cloned_repo);
--- a/examples/network/git2.c
+++ b/examples/network/git2.c
@@ -36,7 +36,7 @@ static int run_command(git_cb fn, int argc, char **argv)
 		if (giterr_last() == NULL)
 			fprintf(stderr, "Error without message");
 		else
-			fprintf(stderr, "Bad news:\n %s\n", giterr_last()->message);
+			fprintf(stderr, "Bad news:\n %s\n", giterr_last());
 	}
 
 	if(repo)
--- a/include/git2/errors.h
+++ b/include/git2/errors.h
@@ -60,43 +60,7 @@ typedef enum {
  */
 typedef struct {
 	char *message;
-	int klass;
 } git_error;
-
-/** Error classes */
-typedef enum {
-	GITERR_NONE = 0,
-	GITERR_NOMEMORY,
-	GITERR_OS,
-	GITERR_INVALID,
-	GITERR_REFERENCE,
-	GITERR_ZLIB,
-	GITERR_REPOSITORY,
-	GITERR_CONFIG,
-	GITERR_REGEX,
-	GITERR_ODB,
-	GITERR_INDEX,
-	GITERR_OBJECT,
-	GITERR_NET,
-	GITERR_TAG,
-	GITERR_TREE,
-	GITERR_INDEXER,
-	GITERR_SSL,
-	GITERR_SUBMODULE,
-	GITERR_THREAD,
-	GITERR_STASH,
-	GITERR_CHECKOUT,
-	GITERR_FETCHHEAD,
-	GITERR_MERGE,
-	GITERR_SSH,
-	GITERR_FILTER,
-	GITERR_REVERT,
-	GITERR_CALLBACK,
-	GITERR_CHERRYPICK,
-	GITERR_DESCRIBE,
-	GITERR_REBASE,
-	GITERR_FILESYSTEM
-} git_error_t;
 
 /**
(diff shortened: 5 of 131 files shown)


mikeando:
I'd much rather have an error code that I'm not going to use 90% of the time, than have to rely on parsing an error string (which might change from release to release!) for the other 10%.

Now I'd be completely OK with `giterr_last_message()` as an alternative to `giterr_last()->message` (though it wouldn't save me any keystrokes.) Hell I'd be happy with `giterr_last()` returning a `char *` so long as there was a way to get that precious error code : `giterr_last_code()` or even `giterr_last_really_i_want_the_whole_struct()->code`. Either would be better than `if(strcmp(giterr_last(),"A problem occurred when fooing the blah")==0)`

end rant.

vmg:
Rugged makes [extensive use](https://github.com/libgit2/rugged/blob/master/ext/rugged/rugged.c#L28-L52) of error classes, so this change will be deeply breaking for any Ruby applications that use Rugged... like, you know, GitHub Dot Com The WebSite.

What's wrong with the exception classes as they are now? I thought they were meant to map to different exception objects in dynamic languages. Does libgit2# always ignore them and raise a generic exception?

rhuijben:
I also use the error classes for wrapping as different error types. If libit2 ever wants to support localization (which it might already do if it forwards OS error messages) parsing strings is never going to work in a stable way.

carlosmn:
There is no proposal, here or anywhere else, for parsing strings to be the way you get information about errors. The error codes aren't going anywhere.

carlosmn:
The change to use a `git_buf` is good independently, we can merge that bit in without issues.

I think we did discuss the relative merits of error classes vs error codes, but it does look like the classes are used more widely than we thought. In rugged each class has an exception associated with it. In pygit2 a few do map to particular exceptions (OOM and OS errors).

So it does seem at least the dynamic languages use the class for determining which exceptions to throw. Then again, one has to wonder how much callers actually care about the type of the exception.

ethomson:
I'm really shocked that these are used.  Obviously nobody is advocating parsing strings.  (Parsing strings would not give you any information that this error class provides anyway.)  A couple of points.

>  I'd much rather have an error code that I'm not going to use 90% of the time, than have to rely on parsing an error string (which might change from release to release!) for the other 10%.

The code is not unique to the error, it's a class of errors about (roughly) where the error occurred.  It gives you little (if any) insight into the error.

If you do a merge and get a ReferenceException that tells you that there was a problem with the reference.  The error class is not actionable at all.

> If libgit2 ever wants to support localization (which it might already do if it forwards OS error messages) parsing strings is never going to work in a stable way.

This proposal has nothing to do with localization.  That would require us to uniquely identity each error since we're not going to localize all reference errors to "a reference error occurred".  This is orthogonal.

>  What's wrong with the exception classes as they are now? I thought they were meant to map to different exception objects in dynamic languages.

There's nothing wrong with them, but my assertion is that they provide very little value.

How does GitHub use the knowledge that an error was a `ConfigurationException` versus a `ReferenceException`?

ethomson:
> The change to use a git_buf is good independently, we can merge that bit in without issues

Yeah, this is useful independent of the error class discussion.  As is some of the dead code cleanup.

One of the other things here than might be nice is the removal of separation between `GITERR_OS` to signal that the setter should append `strerror` versus the error class being an "OS Error".  Thoughts?

If we keep the error class, I wonder about the changes when it's an operating system error.  I feel like having a separate function that appends the error message may be preferable, versus

vmg:
> How does GitHub use the knowledge that an error was a ConfigurationException versus a ReferenceException?

I can think of a couple examples, but they are not very insightful. I'll agree that the current character classes are not particularly useful.

Here's the thing: this is a huge, breaking change that doesn't really improve on the status quo. I'm all for breaking shit when it can be improved, but removing the superfluous error classes changes the API without objectively improving it. I would very much prefer to see a proposal for error handling revamp that strictly improves on our current iteration (ideally before 1.0). *That* would be a solid reason to break the errors API.

What would improve the errors API? Obviously I'm really not sure -- otherwise I would have tried to implement it myself. ;)

On my current mood for C design, I would like to see a more UNIX-oriented approach to error handling. Namely: all functions return 0 or -1; the specific error code is set in a global `libgit2_errno`, together with the detailed error strings, the error classes are removed altogether, and more error codes are added to separate specific error cases without looking at the detailed error string.

Would this be a strict improvement? I don't know. I would pay the price just to be able to remove the thousands of `if ((error = foobar()) < 0) return error;` littered around the codebase. The error passthrough has been living hell for us. I also think the `int` range is big as fuck, and we're at around 30 error codes -- we could definitely fill it up a tad more.

Food for thought. :)

ethomson:
> I can think of a couple examples, but they are not very insightful. I'll agree that the current character classes are not particularly useful.

Sure.  My concern is that I would expect that switching based on class would be very brittle.

> Here's the thing: this is a huge, breaking change that doesn't really improve on the status quo.

That's regrettable.  Obviously the goal here was to improve.  At the moment we have three pieces of error information:

1. Return code
2. Error class
3. Error message

The goal was to distill this into two pieces of information:  one that is actionable programmatically (the return code) and one that is actionable by the end user (the error message).  The error class doesn't really do much except act as a subclass for a generic libgit2 exception in managed languages.  (Though that's not to say that this doesn't make it useful, maybe I underestimate the value of be able to pretty-print "ReferenceException occurred".)

While distilling the information here is an immediate goal, I did have a stretch goal of separating the user-visible error data structure from the internal data.  I anticipate localization will run a freight train through the codebase and the TLS error structure that we use internally is going to get somewhat ugly.  There will be (I suspect) data that the user shouldn't be privy to.

Another nicety of breaking those two is that we may have the ability to store more static data and do less `git_buf_printf`ing at the time we set an error.  It would be nice if (for example) we didn't do a bunch of `printf`s every time we look for a configuration value and don't care whether it's actually found or not.

Obviously neither of these are really dependent on the error class removal.

I didn't know that the error code / return code combination bothered you.  There are a few places where things start getting overloaded (positive is meaningful, negative is an error code) but those are rare enough that this has never bothered me, nor does having to maintain an error code bother me.  (Except when we fail to do so, which may be reason enough to remove it.)

Anyway.  It seems like we all have some complaints about the current error message handling - maybe now is a good time to pile on some grievances in case there are other complaints that others have that nobody else realizes are complaints.

carlosmn:
I certainly agree that we could have a lot more error codes. We've been limiting ourselves to codes where we expect the caller to be able to perform some specific recovery beyond passing the error string on to its caller or the user, rather than as a way to provide details about the error.

Would you do away with the free-form error string and rely on this `git_errno` and `git_strerror()` to provide the error string? Or would you still like to keep it?

Where I see the classes be useful is for complex operation, a `ENOTFOUND` doesn't quite tell you *what* wasn't found, whereas the class can tell you whether it was an index entry or a reference or whatever. I don't know if this is something anybody making use of. I personally don't for the code we have using libgit2.

I'd like to also point to #3036 which is an example of how to solve a somewhat different problem, while still related to how we handle errors. As we allow pluggable bits, there's times where the caller and a backend want to exchange extra information about an error, but we don't currently allow anything other than returning the string and a pre-defined error code.

vmg:
> Would you do away with the free-form error string and rely on this git_errno and git_strerror() to provide the error string? Or would you still like to keep it?

I think the freeform error strings have a lot of value, particularly for displaying details to users.

PR #3259, opened 2015-06-25
Stash apply: stage new files even when not updating the index

ethomson:
Stash application should stage new files, even when we're not updating the index.

````
C:\Temp\stash_apply_sucks>git status
On branch master
Changes to be committed:

        new file:   bar
        modified:   foo


C:\Temp\stash_apply_sucks>git stash
Saved working directory and index state WIP on master: 4a6df6d foo
HEAD is now at 4a6df6d foo

C:\Temp\stash_apply_sucks>git status
On branch master
nothing to commit, working directory clean

C:\Temp\stash_apply_sucks>git stash apply
On branch master
Changes to be committed:

        new file:   bar

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   foo
````

WHAT A SENSIBLE ARCHITECTURE THIS IS.  So we need to do the same thing.  We break the actual iterator walking code out of merge to create a new index that has just the new files, which we will set as the new repo's index (unless the user actually specified that they wanted to restore the index).

### DIFF
--- a/include/git2/errors.h
+++ b/include/git2/errors.h
@@ -47,6 +47,7 @@ typedef enum {
 	GIT_EPEEL           = -19,      /**< The requested peel operation is not possible */
 	GIT_EEOF            = -20,      /**< Unexpected EOF */
 	GIT_EINVALID        = -21,      /**< Invalid operation or input */
+	GIT_EUNCOMMITTED    = -22,	/**< Uncommitted changes in index prevented operation */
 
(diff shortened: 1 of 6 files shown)


ethomson:
Oh, right, it turns out that we should also block stash application if there are any staged changes.

PR #3260, opened 2015-06-25
stash: test we apply using reflog-like indices

ethomson:
Quick test to ensure that we can apply using indexes to get older stashes.  Make sure that the indexes are reflog-style, from the man page:

> stash@{0} is the most recently created stash, stash@{1} is the one before it


ethomson:
Oops.  This test actually depends on #3259

PR #3262, opened 2015-06-26
Desenfuckenate prefix lookups

vmg:
They were previously fucked up since @carlosmn merged https://github.com/libgit2/libgit2/pull/3201

The inconspicuous fix to stop Coverity from complaining broke essentially most prefix lookups in the library where the prefix was more than 20 nibbles long. I'm glad I caught this in github/gitrpc integration testing 'cause this the kinda stuff that keeps me awake at night.

### DIFF
--- a/src/object.c
+++ b/src/object.c
@@ -129,10 +129,10 @@ int git_object_lookup_prefix(
 	if (error < 0)
 		return error;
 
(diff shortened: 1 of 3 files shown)


ethomson:
Actually, *I* merged that.

PR #3264, opened 2015-06-26
Fast builds?

arthurschreiber:


carlosmn:
There's one usage of sudo left, where we install valgrind.

arthurschreiber:
Right. Let me get that fixed as well

arthurschreiber:
FYI. I just pushed the valgrind change and linux builds started immediately, without any wait time.

PR #3271, opened 2015-06-29
memory leak refspec.c

jeffhostetler:
A couple of memory leaks.


jeffhostetler:
I reverted the change in tests/filter/stream.c because it caused a segfault on Linux on Travis.  I'm still seeing it as a leak on Windows though.  So we can address this separately later.

ethomson:
Looks like a few places call `refspec__free` assuming that they can always do so.  Might be a safe thing to zero the refpsec so that subsequent `free` calls succeed.

jeffhostetler:
Yeah, most of the calls to git_refspec__parse() do explicitly free the refspec regardless of the parse result. The leak was because some calls just returned if the parse failed.

PR #3272, opened 2015-06-29
git_cert: child types use proper base type

ethomson:
Add a proper `parent` member to the `git_cert` descendants to prevent type punning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ v0.23 + 1
 
 ### API removals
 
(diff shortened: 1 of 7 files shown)


carlosmn:
The thing about merging this for v0.23 is that is a break in what is supposed to be a freeze period. Otherwise it's fine.

carlosmn:
Could you update where the changelog entry goes? This would make it look like the change was done for v0.23

ethomson:
Fixed!

PR #3275, opened 2015-06-30
http: fixed leak when asking for credentials again

swisspol:
t->cred might have been allocated the previous time and needs to be
freed before asking caller for credentials again.


swisspol:
I had the same fix as 8b3800601866b8a579a1787089673edc6380c1eb on my branch but with this extra code.

ethomson:
Looks good, thanks!

PR #3279, opened 2015-06-30
Fix #3094 - improve use of portable size_t/ssize_t format specifiers.

### LINKED ISSUE #3094: Use PRIuZ in more places at fetch.c?
$ g grep -n zu -- examples/network/fetch.c

examples/network/fetch.c:145:           printf("\rReceived %d/%d objects in %zu bytes (used %d local objects)\n",
examples/network/fetch.c:148:           printf("\rReceived %d/%d objects in %zu bytes\n",

mplough:
The header src/cc-compat.h defines portable format specifiers PRIuZ, PRIdZ, and PRIxZ. The original report highlighted the need to use these specifiers in examples/network/fetch.c.  For this commit, I checked all C source and header files not in deps/ and transitioned to the appropriate format specifier where appropriate.

### DIFF
--- a/examples/general.c
+++ b/examples/general.c
@@ -142,8 +142,8 @@ int main (int argc, char** argv)
   // We provide methods to convert from the object type which is an enum, to
   // a string representation of that value (and vice-versa).
   str_type = git_object_type2string(otype);
-  printf("object length and type: %d, %s\n",
-      (int)git_odb_object_size(obj),
+  printf("object length and type: %"PRIuZ", %s\n",
+      git_odb_object_size(obj),
       str_type);
 
   // For proper memory management, close the object when you are done with
@@ -324,7 +324,7 @@ int main (int argc, char** argv)
   // Getting the count of entries in the tree so you can iterate over them
   // if you want to.
   size_t cnt = git_tree_entrycount(tree); // 3
-  printf("tree entries: %d\n", (int)cnt);
+  printf("tree entries: %"PRIuZ"\n", cnt);
 
   entry = git_tree_entry_byindex(tree, 0);
   printf("Entry name: %s\n", git_tree_entry_name(entry)); // "hello.c"
--- a/examples/network/fetch.c
+++ b/examples/network/fetch.c
@@ -143,10 +143,10 @@ int fetch(git_repository *repo, int argc, char **argv)
 	 * network.
 	 */
 	if (stats->local_objects > 0) {
-		printf("\rReceived %d/%d objects in %zu bytes (used %d local objects)\n",
+		printf("\rReceived %d/%d objects in %" PRIuZ " bytes (used %d local objects)\n",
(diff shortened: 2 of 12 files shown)


ethomson:
The examples do not include the internal headers from libgit2.  (Nor should they, as they are intended to be self-contained examples).  That's where these format specifiers are defined.

I think that it's valuable to have examples that are copy-pasteable (since many people will do exactly that).  Do you think that it's better for people learning about libgit2 to have these format specifiers defined in some header that is common to the examples (at the risk of it not being obvious what these things are) or that it's better to downcast these because everybody knows what `%d` is (despite not being strictly correct)?

mplough:
That's an excellent question, and it definitely doesn't have an easy answer.  I checked into how the sqlite project handles things in their examples, but sqlite3 doesn't use `size_t` much.  The design docs for sqlite4 call for it and other standard types (e.g. `uint32_t`) to be used a lot more, so that'll be interesting to follow.  In the absence of precedent, I'll go over the pros and cons of the two approaches you mentioned.

Taking the relevant pieces from src/cc-compat.h and putting them into an example-specific header file keeps things pedantically correct, which is an important attribute for dealing with something as precious as source code repositories.  However, the additional `#include` directive required in the examples keeps things from being self-contained, and will complicate things, possibly too much, for copy-pasters.  Using a simpler format specifier like a `%d` keeps things self contained, but I'd worry about the consequences of subtly incorrect code operating on repositories.

There's also a third option: It's possible to keep things self-contained but closer to correct than `%d` by sacrificing portability.  The original version of fetch.c uses C99 standard `%zu` format specifiers for printing variables of type `size_t`.  This is correct but won't work with Visual Studio, since according to [Stack Overflow](http://stackoverflow.com/questions/15610053/correct-printf-format-specifier-for-size-t-zu-or-iu) and the [MSVC documentation](https://msdn.microsoft.com/en-us/library/vstudio/tcxf1dw6.aspx), the `z` specifier is not supported.  MSVC checks for unsupported format specifiers, so code using `%zu` will yield a compilation error.  The developer would then be forced to figure out that `%Iu` is the correct specifier to use on MSVC.

Correctness, compactness, portability: choose two.  I don't know which of compactness or portability to choose, but given the application, I'd have to vote for correctness.  What do you think?

carlosmn:
I don't mind the examples using `%d` for `size_t` as our use of that type is due to being somewhat pedantic about the possible ranges. There barely going to be a realistic scenario in which we e.g. have than 2^31 entries in a tree.

`examples/general.c` is a mess anyhow, so I wouldn't worry about that. The fetch example might be able to get a very large packfile, but I don't quite feel like putting the format specifiers just for that one function is worth the bother.

carlosmn:
Mind you, the network example headers already have this, so reverting the `general.c` changes should be enough.

carlosmn:
I reverted those changes and merged manually. Thanks.

PR #3280, opened 2015-07-01
Fix 8.3 filename tests failure when 8.3 is disabled

linquize:
Fix 8.3 filename tests failure when 8.3 is disabled

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -867,7 +867,9 @@ static int reserved_names_add8dot3(git_repository *repo, const char *path)
 {
 	char *name = git_win32_path_8dot3_name(path);
 	const char *def = GIT_DIR_SHORTNAME;
+	const char *def_dot_git = DOT_GIT;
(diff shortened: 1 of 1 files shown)


ethomson:
I don't understand.  Is there just a test failure?  What test?  Why?

We shouldn't include `.git` in the list of shortnames.  That's already a prohibited name.  If this is just working around a test failure then we should fix the test.

linquize:
`test_repo_reservedname__includes_shortname_on_win32` and 
`test_repo_reservedname__includes_shortname_when_requested`

`cl_assert_equal_i(2, reserved_len);`
but `reserved_len` was 3

linquize:
because it added ".git" to the end of the list (repeated), so `reserved_len` was 3

ethomson:
I see.  Out of curiosity, what system / filesystem is that?

linquize:
win 8 x64, NTFS

PR #3285, opened 2015-07-01
libgit2 does not work the same way as vanilla git does for adding files to index

csware:
This is based on issue #2798 and issue #3199.

The code of issue #2798 is rebased onto current master and uses a similar testing approach as issue #3199.
Hopefully there are better chances for getting this merged than issue #2798.

The files in "checkin_input_files" are directly based on the files used for issue #3199.

/cc @ethomson


csware:
This is an open issue for a very long time? Any comments or don't you care for broken CRLF?

tests/generate_crlf_checkin.sh
```diff
@@ -0,0 +1,107 @@
...
+		fi
+	else
+		OBJ=$(ls .git/objects/??/* 2>/dev/null)
+ 		set -e
+		OBJ=${OBJ:13} # 13=".git/objects/"
+		OBJ=${OBJ//\//}
+		echo "$OBJ" > "$destdir/$file.obj"
+	fi
+	rm -f .git/objects/??/*
+	rm -f .git/index
```
carlosmn: Why are we accessing git's private data? If we need the repository, we should store it like we do in every other case (including `tests/generate_crlf.sh` and use `.gitted/` as the git-dir. Looking into this structure is very fragile.

tests/index/crlf.c
```diff
@@ -24,6 +26,210 @@ void test_index_crlf__cleanup(void)
...
+	git_buf_free(&expected_fixture);
+
+	git_buf_free(&attrbuf);
+	git_buf_free(&expected_fixture);
+	git_buf_free(&expected_dirname);
+	git_buf_free(&sandboxname);
+	git_buf_free(&reponame);
+}
+
+static void set_up_workingdir(const char *name)
```
carlosmn: We already have a way to put test resources into the temp dir, why do we need this new one?

carlosmn:
In general, the commits are missing an explanation for why the change is necessary. They don't explain why the old version was wrong or why the new version of the code fixes it. They just claim that they fix something. This makes it much harder to go through the commits and make sense of the changes. It would also help if they mentioned which failing tests they are fixing.

tests/index/crlf.c
csware: This was inspired by https://github.com/libgit2/libgit2/pull/3199/files as we need the subdirectory for the test case results to compare with.

tests/generate_crlf_checkin.sh
csware: In my first attempt all the private data was hardcoded, then it was requested that a script should be used for generating the "known good test results". I'm accessing the private data so that I can be sure which objects (ids) were created on adding a file (after the CRLF handling).

If you have a better idea here, I'm open.

csware:
* LF -> CRLF filter now correctly honors core.safecrlf=true errors
  * LF only files were accepted with core.autocrlf=true on CRLF platforms
  * files containig CRLF in combination with core.autocrlf=input were accepted
  * adding files containing CR and CRLF but not the same number failed

* LF -> CRLF filter now handles files on adding to index the way vanilla git does
  * skip mangling of files marked as "binary" (-crlf)
  * especially files containing single CR chars are added as is now
  * honor "text" attribute for forcing a file being interpreted as text

tests/generate_crlf_checkin.sh
carlosmn: But you don't want the private data, you want to see which objects were created. Ask git for it, don't take it out of `.git/`.

    git ls-files -s | cut -d ' ' -f 2

will give you the object names of whatever is in the index. If you're sure there's only ever one, that will give you the answer already. If you want to clear the index, use

    git read-tree 4b825dc642cb6eb9a060e54bf8d69288fbee4904

where that's the empty tree. You can generate it yourself at the beginning of the script with

    git hash-object -t tree --stdin </dev/null

if you'd rather not hard-code it.

csware: @carlosmn Thanks, updated.

csware:
/cc @ethomson @carlosmn 
Any news on this PR? AutoCrLf i still broken in libgit2... Btw. I'm using this code for several months in TortoiseGit (also, i don't know why the appveyor tests show failures, local tests run w/o problems)

johnhaley81:
:+1: auto-crlf affects us a lot over in GitKraken land as well.

csware:
This PR is now updated to match with the changed CrLf filter logic of Git 2.9.

csware:
I'm sorry that the test is failing as this is dependant on PR #4098. When this is on top of PR #4098 the tests succeed w/o errors (tested locally on WIndows and *nix).

csware:
I'm sorry that the tests are failing right now as this is dependant on PR #4098. When this is on top of PR #4098 all the tests succeed w/o errors:

Rebased commit: https://github.com/libgit2/libgit2/pull/3285/commits/bcf497d9590f9b0fa52e3011235524eb6c414096
AppVeyor: https://ci.appveyor.com/project/libgit2/libgit2/build/3139
Travis CI: https://travis-ci.org/libgit2/libgit2/builds/196256027

src/crlf.c
```diff
@@ -161,9 +161,10 @@ static int crlf_apply_to_odb(
 
 	/* If safecrlf is enabled, sanity-check the result. */
 	if (ca->crlf_action == GIT_CRLF_INPUT ||
-		(ca->auto_crlf == GIT_AUTO_CRLF_INPUT &&
-		(ca->crlf_action == GIT_CRLF_GUESS || ca->crlf_action == GIT_CRLF_AUTO ||
-		(ca->crlf_action == GIT_CRLF_TEXT && ca->eol==GIT_EOL_UNSET)))) {
+		(ca->crlf_action == GIT_CRLF_GUESS || ca->crlf_action == GIT_CRLF_AUTO || ca->crlf_action == GIT_CRLF_TEXT) && ca->eol == GIT_EOL_LF ||
+		(ca->auto_crlf == GIT_AUTO_CRLF_INPUT && (ca->crlf_action == GIT_CRLF_GUESS)) ||
+		(ca->auto_crlf == GIT_AUTO_CRLF_INPUT && (ca->crlf_action == GIT_CRLF_AUTO || ca->crlf_action == GIT_CRLF_TEXT) && ca->eol == GIT_EOL_UNSET) ||
+		(GIT_EOL_NATIVE != GIT_EOL_CRLF && ca->auto_crlf == GIT_AUTO_CRLF_FALSE && (ca->crlf_action == GIT_CRLF_AUTO || ca->crlf_action == GIT_CRLF_TEXT) && ca->eol == GIT_EOL_UNSET)) {
```
novalis: This seems overcomplicated to me.  For one thing, you sometimes have the action test before the auto_crlf test and sometimes after.  This means that you miss some commonalities like the duplicated ca->crlf_action == auto || ca->crlf_action == text.  We could also make it easier to read by breaking out sa->action, sa->eol, and sa->auto_crlf into separate variables.  And we could define a helper GIT_EOF_NATIVE_IS_LF = GIT_EOL_NATIVE != GIT_EOL_CRLF, just to make that bit a bit shorter and clearer.   I made these changes and I was still unable to make heads or tails of it.  So then I thought: what if we make a function?


```
static int is_crlf_reasonable(sa) {

	switch (sa->crlf_action) {
	case GIT_CRLF_INPUT:
		return 1;
		break;
	case GIT_CRLF_GUESS:
		return sa->eol == GIT_EOL_LF || sa->auto_crlf == GIT_AUTO_CRLF_INPUT;
	case GIT_CRLF_AUTO: /* FALL-THROUGH */
	case GIT_CRLF_TEXT:
		switch (sa->eol) {
		case GIT_EOL_LF:
			return 1;
		case GIT_EOL_UNSET:
			return sa->auto_crlf == GIT_AUTO_CRLF_INPUT;
		case GIT_EOL_UNSET:
		return GIT_EOL_NATIVE_IS_LF && sa->auto_crlf == GIT_AUTO_CRLF_FALSE;
		}
	}

	return 0;
}
```

I have not yet actually tested this to ensure that it is identical, but I think the overall approach is much easier to read. Plus it's clear from the function name that success means that the check passed. Perhaps we need a better function name -- I don't really understand what this code is doing, but I can tell that I would be unable to debug it.

tests/index/crlf.c
```diff
@@ -162,7 +168,7 @@ static void test_add_index(const char *safecrlf, const char *autocrlf, const cha
 
 	cl_git_pass(git_index_clear(g_index));
 
-	git_buf_joinpath(&expected_dirname, "crlf_data", "checkin_results");
+	git_buf_joinpath(&expected_dirname, "crlf_data/checkin_results", systype);
```
novalis: If you're doing joinpath anyway, don't use a slash.  add another joinpath.

src/crlf.c
```diff
@@ -146,16 +149,26 @@ static int crlf_apply_to_odb(
...
+		}
+	}
 
-		/* If safecrlf is enabled, sanity-check the result. */
-		if (stats.lf != stats.crlf) {
+	/* If safecrlf is enabled, sanity-check the result. */
+	if (ca->crlf_action == GIT_CRLF_INPUT ||
+		(ca->auto_crlf == GIT_AUTO_CRLF_INPUT &&
+		(ca->crlf_action == GIT_CRLF_GUESS || ca->crlf_action == GIT_CRLF_AUTO ||
+		(ca->crlf_action == GIT_CRLF_TEXT && ca->eol==GIT_EOL_UNSET)))) {
```
novalis: nit: missing spaces around the last ==.

But more importantly, this is another case of "sanity-checking" that doesn't make clear what exactly is being checked and which is an if statement with a thousand clauses. This one is somewhat less heinous than the other one, but it does seem to cover a lot of the same ground, so I wonder if you missed some cases here, or if we could extract the common logic between the two, or something.

csware:
@novalis @pks-t @ethomson I refactored this (again), so hopefully the code is better readable now.

novalis:
Thanks, that refactor looks much cleaner.

ethomson:
Thanks for the work on this @csware - and thanks much for the review @novalis.  I'll take a 👀 as soon as I have a little extra time.  Sorry it's taken so long already, I hope to get to this promptly.

tests/index/crlf.c
```diff
@@ -14,16 +14,221 @@
...
+	int add_bypath_ret;
+	const git_index_entry *entry;
+	git_oid oid;
+	git_error_state error = { 0 };
+
+	basename = git_path_basename(actual_path->ptr);
+	cmp_git = strcmp(basename, ".git");
+	cmp_gitattributes = strcmp(basename, ".gitattributes");
+
+	if (cmp_git == 0 || cmp_gitattributes == 0) {
```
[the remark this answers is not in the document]
csware: I just copied the coe from checkout/cflf.c: compare_file

csware:
@pks-t Updated.

csware:
@pks-t I addressed all points. Any chance to get a new review?

pks-t:
I already had a look a few days ago and didn't find anything noteworthy. I didn't yet state that as I am still thinking about these 4000 new files, which honestly put me a bit off. But I still ain't got a better idea how to properly do this, at least nothing that's as comprehensive as what you propose.

@ethomson, @carlosmn: I'd appreciate if you could have another look at this

ethomson:
I’ve been looking at this as time allows.  I have some changes to propose but in the meantime I appreciate all the hard work here.

csware:
It took long, but finally fixed with issue #4904.

PR #3288, opened 2015-07-02
git__getenv: utf-8 aware env reader

ethomson:
Presently, `cl_getenv` wraps `getenv` on posixy systems and does some UTF16-UTF8 magic on Windows.  The return code must be freed on Windows and must not be on Unix boxes.  Ugh.

`p_getenv` just wraps `getenv` on all systems, which means it's ANSI code page on Windows.  (This hasn't really bitten us, yet, as we don't read potentially Unicode values in the library, only in the tests.)  But still, ugh again.

This introduces `git__getenv` which is a UTF-8 aware `getenv` everywhere.  Plus some helpery clar stuff because we can.

### DIFF
--- a/tests/online/push.c
+++ b/tests/online/push.c
@@ -9,16 +9,16 @@
 
 static git_repository *_repo;
 
-static char *_remote_url;
+static char *_remote_url = NULL;
 
-static char *_remote_ssh_key;
-static char *_remote_ssh_pubkey;
-static char *_remote_ssh_passphrase;
+static char *_remote_user = NULL;
+static char *_remote_pass = NULL;
 
-static char *_remote_user;
-static char *_remote_pass;
+static char *_remote_ssh_key = NULL;
+static char *_remote_ssh_pubkey = NULL;
+static char *_remote_ssh_passphrase = NULL;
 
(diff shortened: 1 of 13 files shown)


ethomson:
Bleh.  `cl_getenv` is now too simplistic as it coalesces not found and empty strings.

ethomson:
Okay, this is ready for review.  This only affects tests so I'm happy to punt on it until 0.23 lands.

tests/online/push.c
```diff
@@ -9,16 +9,16 @@
...
+static char *_remote_pass = NULL;
 
-static char *_remote_user;
-static char *_remote_pass;
+static char *_remote_ssh_key = NULL;
+static char *_remote_ssh_pubkey = NULL;
+static char *_remote_ssh_passphrase = NULL;
 
-static char *_remote_default;
+static char *_remote_default = NULL;
```
carlosmn: All of these `NULL`s are redundant. Static data is initialized to zero. It's not a big deal, but it's a part of the diff that doesn't do anything.

carlosmn:
Yeah, let's hold off on merging it until after the release.

Issue #3266, opened 2015-06-26
path_repository and custom backend


leoyanggit:
Per my understanding `path_repository` is required for the index file and some other stuff like `rebase-merge` dir etc. But I can't find a way to set repository path for a custom backend unless I create a filesystem backend repository with `git_repository_init()` or something like that and then call `git_repository_set_odb()` with my custom backend. The drawback of this way is that it creates a lot of useless files/dirs inside `.git` dir, which is really confusing. I found we have `git_repository_path()` but I can't find `git_repository_set_path()` or something like that. I did find `git_repository_workdir()` and `git_repository_set_workdir()`, which is good. Is there a cleaner way to set the repository path for a custom backend or am I missing something?

carlosmn:
The repository's path belongs solely to the repository. A backend should have no way to tell the repository to change where it thinks it lives. A backend cannot change this idea.

ethomson:
So it sounds to me like you're trying to set custom backends and invoke some functionality that expects to be able to write into the repository directory...  You mention rebase explicitly, so let's use that as an example:  rebase is really aimed at being compatible with core git and writing to the repository directory in a wholly compatible way.  If you don't require that compatibility (and it sounds like you don't, since you're using custom backends) then I think that the best thing to do would be to just use one of the mechanisms that can work in a completely in-memory scenario.

For example: rebase is really just a sequence of cherry-picks, you can simply drive `git_cherrypick_commit`.

If you describe better what you're trying to do, we might be able to provide some better advice.

leoyanggit:
>The repository's path belongs solely to the repository. A backend should have no way to tell the repository to change where it thinks it lives. A backend cannot change this idea.

Agreed.  But I'm not saying that a backend API should take the repository path. There should be a clean way to specify the repository path for a repository with custom backend to make the current implementation of rebase work cleanly. The workaround I mentioned is not a clean way though.

leoyanggit:
> I think that the best thing to do would be to just use one of the mechanisms that can work in a completely in-memory scenario.

I really wish we had in-memory implementation, as in merge, of rebase. The merge API is really cool because I can choose to do merge in an in-memory index.

> For example: rebase is really just a sequence of cherry-picks, you can simply drive git_cherrypick_commit.

I could. It means writing part of the rebase logic (although just walking in the tree) in the user's code. Since we already have rebase, I wonder why rewrite part of its logic in stead of making rebase work with custom backends.

[...]

PR #3291, opened 2015-07-03
Add API that set repository path

leoyanggit:
This commit is trying to address issue #3266

### DIFF
--- a/include/git2/repository.h
+++ b/include/git2/repository.h
@@ -363,6 +363,17 @@ GIT_EXTERN(int) git_repository_is_empty(git_repository *repo);
 GIT_EXTERN(const char *) git_repository_path(git_repository *repo);
 
 /**
+ * Set the path of this repository. Use it with caution. Basically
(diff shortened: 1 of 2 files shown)


carlosmn:
The reason for `git_repository_wrap_odb()` is, as the documentation states, is to use "when all you have is an object database". This is to fix the mismatch between having a `git_odb` and many object lookup functions expecting a `git_repsitory`. This lets you use the repository for these functions which proxies the calls to the odb.

If you have a path to give to the repository, then you don't have the situation which this addresses. Having a path means that you do have a repository, and this function isn't going to help you.

leoyanggit:
> If you have a path to give to the repository, then you don't have the situation which this addresses. Having a path means that you do have a repository, and this function isn't going to help you.

My problem is that I can't find a clean way to create a repository with stuff only needed in a custom backend as I mentioned in the issue #3266. All the ways I can find to create a repository will leave useless files/dirs for a custom backend.

PR #2986, opened 2015-03-16
WinHTTP for MinGW

tkelman:
Rebased #2414 and added in the pieces of the other patches from that thread that were necessary to build locally ~~(have only tested with i686-w64-mingw32 so far)~~, plus a couple of little fixes of my own. Hopefully I didn't screw up the conflict resolution here too badly.

PR #3292, opened 2015-07-05
Increase required version of cmake to 2.8

tkelman:
My bad, this policy line was introduced with #2986

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@
 # > cmake --build . --target install
 
 PROJECT(libgit2 C)
-CMAKE_MINIMUM_REQUIRED(VERSION 2.6)
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
 CMAKE_POLICY(SET CMP0015 NEW)
 
 # Add find modules to the path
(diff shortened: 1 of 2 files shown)


tkelman:
Now that there's a maint/v0.23 branch, this should also be backported to it.

tkelman:
An alternative would be to bump `CMAKE_MINIMUM_REQUIRED` - there's a `cmake28` package for 2.8.11.2 in EPEL5, which is probably the only distribution old enough for this to matter.

carlosmn:
I checked CentOS6 and Debian oldstable and both have 2.8 which is why I didn't merge this before the release. I'd be OK with bumping to 2.8 since it's available where we care. @ethomson ?

tkelman:
I wound up having issues linking libgit2_clar with 2.6 anyway (wasn't listing libcrypto), ref https://github.com/staticfloat/ansible-julia_buildslave/pull/1

I think the anaconda folks also use centos 5 for building binaries for maximum compatibility with as old a glibc as anyone could conceivably be using, though they might be on 6 which we should probably switch to before it gets to be too painful.

ethomson:
>  I'd be OK with bumping to 2.8 since it's available where we care.

Fine with me.

tkelman:
>> I'd be OK with bumping to 2.8 since it's available where we care.
>
> Fine with me.

I know github doesn't send a notification, so just letting you know that I've replaced the PR content to do this instead.

ethomson:
Looks good to me.  Thanks!

PR #3297, opened 2015-07-07
Fix undefined reference with old versions of openssl

tkelman:
Versions prior to 0.9.8f  did not have this function, rhel/centos5 are still on a
heavily backported version of 0.9.8e and theoretically supported until March 2017

Without this ifdef, I get the following link failure:
```
CMakeFiles/libgit2_clar.dir/src/openssl_stream.c.o: In function `openssl_connect':
openssl_stream.c:(.text+0x45a): undefined reference to `SSL_set_tlsext_host_name'
collect2: error: ld returned 1 exit status
make[6]: *** [libgit2_clar] Error 1
```

### DIFF
--- a/src/openssl_stream.c
+++ b/src/openssl_stream.c
@@ -324,7 +324,9 @@ int openssl_connect(git_stream *stream)
 
 	SSL_set_bio(st->ssl, bio, bio);
 	/* specify the host in case SNI is needed */
+#ifdef SSL_CTRL_SET_TLSEXT_HOSTNAME
 	SSL_set_tlsext_host_name(st->ssl, st->host);
+#endif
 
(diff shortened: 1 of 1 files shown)


ethomson:
There's an argument to be made that we should either require SNI or require users to accept that they're okay with not having SNI (via some cmake define, or whatever).

I'm pretty okay with this just because it's remarkably low-impact.  (If you're running a version of OpenSSL that old, you've got other issues.)

/cc @carlosmn (in case he has a strong opinion)

carlosmn:
I don't mind too much. On such an old system you're going to have to consider that you won't have some features.

tkelman:
Thanks!

Issue #3294, opened 2015-07-06
merge: bogus newline appended in the EOF on file conflict


niamster:
If commit with a conflict on a file is merged as `ours` and originally there's no newline in the file, the newline is appended.
In the following snippet file `0` initially didn't have newline, while after the merge it appeared.
```
*~>git lol 0
*   9ba607b Merge 26b6a3e8203aefd5382aa4349f98d577e8f8e53f into refs/heads/master
|\
| * 26b6a3e commit file 0 v.1
* 41c5194 commit file 0 v.0
===================================================================
*~>git show 41c5194
commit 41c5194707433382f32278af6567853d01a6b94a
    commit file 0 v.0
diff --git a/0 b/0
new file mode 100644
index 0000000..23475bd
--- /dev/null
+++ b/0
@@ -0,0 +1 @@
+content of file 0 v.0
\ No newline at end of file
===================================================================
*~>git show 26b6a3e
commit 26b6a3e8203aefd5382aa4349f98d577e8f8e53f
    commit file 0 v.1
diff --git a/0 b/0
new file mode 100644
index 0000000..c9da6f9
--- /dev/null
+++ b/0
@@ -0,0 +1 @@
+content of file 0 v.1
\ No newline at end of file
===================================================================
*~>git show 9ba607b
commit 9ba607b6825a02437c27a7b2d61a4ebf44ae521e
Merge: e147d4d 26b6a3e
    Merge 26b6a3e8203aefd5382aa4349f98d577e8f8e53f into refs/heads/master
diff --cc 0
index 23475bd,c9da6f9..92953f4
--- a/0
+++ b/0
@@@ -1,1 -1,1 +1,1 @@@
- content of file 0 v.0
 -content of file 0 v.1
++content of file 0 v.0
```

ethomson:
So does `git-merge-recursive` when given the `ours` flag.

````
% hexdump -C foo
00000000  62 61 72                                          |bar|
00000003

% git merge -srecursive -Xours foo
Auto-merging foo
Merge made by the 'recursive' strategy.
 foo | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

% hexdump -C foo
00000000  62 61 72 0d 0a                                    |bar..|
00000005
````

If you believe that this is a bug, this should be fixed in git.git and we will update to match.

niamster:
@ethomson  thanks for the reply. In fact it does not (with GIT 2.4.3).
[...]

PR #3299, opened 2015-07-07
Don't add unnecessary trailing newline during file merge

ethomson:
Upgrade xdiff to version used in core git 2.4.5 ([0df0541](https://github.com/git/git/commit/0df0541)).

Corrects an issue where a trailing newline is added at the EOF when all the inputs lacked a trailing newline.


ethomson:
Fixes #3294

PR #3302, opened 2015-07-11
List a submodule only once when the path matches a submodule in the index

### LINKED ISSUE #3293: Submodule whose name doesn't match its path is enumerated over twice
This is a recent regression, similar to #3284.

Given a repository with a `.gitmodules` like:

```
[submodule "TestApp"]
	path = Test_App
	url = ../Test_App
```

`git_submodule_foreach` will yield the submodule twice, once named `TestApp` and once named `Test_App`.

Looking quickly it seems like it's getting one name from the index and the other from the `.gitmodules` while not realizing it's the same submodule.

carlosmn:
This solution is probably fragile, but it'll do until the config-loading gets moved to something less callback-based.

This fixes #3293

### DIFF
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -1647,7 +1647,7 @@ static int submodule_load_from_config(
 	} else {
 		khiter_t pos;
 		git_strmap *map = data->map;
-		pos = git_strmap_lookup_index(map, name.ptr);
+		pos = git_strmap_lookup_index(map, path ? path : name.ptr);
 		if (git_strmap_valid_index(map, pos)) {
 			sm = git_strmap_value_at(map, pos);
 		} else {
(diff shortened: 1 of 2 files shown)


carlosmn:
@joshaber can you take a look at this?

joshaber:
Thanks @carlosmn! This seems to resolve https://github.com/libgit2/objective-git/pull/480, though it does report different results than it did before it regressed. Previously it reported:

```
Archimedes,
Test_App,
Test_App2
```

And now it reports:

```
Archimedes,
TestApp,
Test_App2
```

Which is to say, it now prefers the name over the path. That seems more right to me, but I wanted to make sure you're aware.

carlosmn:
That's probably a side-effect of the slightly different lookup we do bleeding into the name that gets chosen. It does seem more correct, as you do have accessors for each of the properties.

joshaber:
Yeah, seems legit to me :+1:

Issue #2706, opened 2014-11-14
backslashes in .gitmodules path values not handled properly on Windows


jeffhostetler:
Doing a "git submodule add" with a path containing backslashes causes the entry in the .gitmodules file to have backslashes.

The code in libgit2 to populate submodules cache during the first git_submodule_foreach() parses the .gitmodules file and then scans the working directory to build a [name, path] dictionary.

Libgit2 does not normalize the path value (with forward slashes) when reading the .gitmodules file, so the cache/dictionary contains a [name, backslash_path] pair.

When the WD is scanned and the on-disk submodule is observed, a [forward_path, forward_path] pair is *ALSO* added to the cache.

A suggested fix would be in submodule.c in submodule_load_from_config() to normalize the value when the key is "path" before using it to create/update the cache/dictionary.

carlosmn:
I'm going to consider this fixed by #3307; shout if that's not right.

PR #3307, opened 2015-07-13
Normalize submodule urls before looking at them

### LINKED ISSUE #2248: git_submodule_resolve_url fails with backslashes
When calling `git_submodule_resolve_url ` on a repository whose default remote url contains backslashes (local clone on Windows), `git_submodule_resolve_url()` fails. That is because `git_remote_url` returns a path with backslashes, but `git_path_apply_relative` only supports forward slashes. 

Where should this be fixed

* Make `git_path_apply_relative` support backslashes on Windows, 
* Change `git_remote_load` to convert backslashes to slashes or
* Fix it in `git_submodule_resolve_url`?

carlosmn:
Our functions expect slashes in the url, so convert backslashes before looking at it. This is a similar though distinct problem to what we have when fetching. There we act on the url/path so we can ask the filesystem for a realpath. Here we're just acting on the string, so we can't rely on that.

This fixes #2248 and possibly #2706.

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -1671,3 +1671,19 @@ bool git_path_isvalid(
 
 	return verify_component(repo, start, (c - start), flags);
 }
+
+int git_path_normalize_slashes(git_buf *out, const char *path)
+{
+	int error;
+	char *p;
+
+	if ((error = git_buf_puts(out, path)) < 0)
+		return error;
+
+	for (p = out->ptr; *p; p++) {
+		if (*p == '\\')
+			*p = '/';
+	}
+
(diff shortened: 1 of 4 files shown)


carlosmn:
Come to think of it, should this be wrapped in a `#ifdef GIT_WIN32`? Technically a unix path should be capable of having a backslash in it.

carlosmn:
Turns out we shouldn't restrict it to Win32, because someone might have created such a `.gitmodules` file on Windows and we might need to use it on non-Windows.

PR #3309, opened 2015-07-14
Maint/v0.23

Fallso:
This is my first PR, so if I've done something completely wrong, I do apologise! It passes all the unit tests and I've written an extra test for the utf8->utf16 and utf16->utf8 conversions which I needed during testing a client side issue.

This includes a number of fixes/enhancements:

1. Speedy diffs on large trees when using pathspecs
2. Windows hidden file support
3. Pre-vista compatibility
4. Fixed windows path composition issue where directory path could end with "\\" or "//" instead of "\" or "/" which caused issue in path comparison later on
5. Fixed macro redefinition warning


Fallso:
Also - ignore some of the earlier commit messages, I made some changes then promptly reverted them to what was originally in the branch...but I couldn't remove the commits themselves from git!

ethomson:
Welcome!

It's much easier for us to review changes independently of each other.  We like to review a single Pull Request that has a single (logical) change.  We also like to have the rationale explained for each change.  That way if (for example) setting lock files hidden is something very useful, and we would like to have it, but (for example) a somewhat arcane flag to `FindFirstFileEx` is *not* something that we would want, then we can merge each change individually.

Would you mind breaking these changes out into logical pull requests and describing the rationale behind each so that we can review them independently?

Fallso:
Hi! That's fine - I will break them up and submit PRs separately. I've just seen your comment about FindExInfoBasic vs FindExInfoStandard so I will filter that from the list of commits I submit. Do you want me to close this PR off now? Or after I've submitted the other ones?

PR #3310, opened 2015-07-14
Windows hidden file support

Fallso:
Hopefully this one will go a bit smoother! 

This PR implements Windows hidden file support. Before, adding a hidden file to the index caused "access denied" errors, and would also cause git_index_add_all to fall over when providing a pathspec that would include hidden files/directories containing hidden files.

I understand this might be a bit of an edge case, but it caused me a few problems!

### DIFF
--- a/src/filebuf.c
+++ b/src/filebuf.c
@@ -71,6 +71,21 @@ static int lock_file(git_filebuf *file, int flags, mode_t mode)
 		char buffer[FILEIO_BUFSIZE];
 		ssize_t read_bytes;
 
+#ifdef GIT_WIN32
+		bool hidden = false;
+		if (strstr(file->path_original, ".git") == NULL) {
+			if (git_win32__ishidden(file->path_original) == 1) {
+				if (git_win32__setvisible(file->path_original) == -1) {
+					giterr_set(GITERR_OS,
+						"Failed to unhide filebuf '%s'",
+						file->path_original);
+					return -1;
+				}
+
+				hidden = true;
+			}
+		}
+#endif
 		source = p_open(file->path_original, O_RDONLY);
 		if (source < 0) {
 			giterr_set(GITERR_OS,
@@ -86,7 +101,11 @@ static int lock_file(git_filebuf *file, int flags, mode_t mode)
 		}
 
 		p_close(source);
-
+#ifdef GIT_WIN32
+		if (hidden) {
+			git_win32__sethidden(file->path_original);
+		}
+#endif
 		if (read_bytes < 0) {
 			giterr_set(GITERR_OS, "Failed to read file '%s'", file->path_original);
 			return -1;
--- a/src/win32/w32_util.h
+++ b/src/win32/w32_util.h
@@ -48,6 +48,22 @@ bool git_win32__findfirstfile_filter(git_win32_path dest, const char *src);
 int git_win32__sethidden(const char *path);
 
 /**
+* Checks if the given path (file or folder) has the +H (hidden) attribute set.
+*
+* @param path The path which should receive the +H bit.
+* @return True if hidden, false if visible
+*/
+int git_win32__ishidden(const char *path);
+
+/**
+* Ensures the given path (file or folder) does not have the +H (hidden) attribute set.
+*
+* @param path The path which should not have the +H bit.
+* @return 0 on success; -1 on failure
+*/
+int git_win32__setvisible(const char *path);
+
+/**
  * Removes any trailing backslashes from a path, except in the case of a drive
  * letter path (C:\, D:\, etc.). This function cannot fail.
  *
--- a/src/fileops.c
+++ b/src/fileops.c
@@ -223,11 +223,27 @@ int git_futils_writebuffer(
 	const git_buf *buf,	const char *path, int flags, mode_t mode)
 {
 	int fd, error = 0;
+	bool hidden = false;
 
 	if (flags <= 0)
 		flags = O_CREAT | O_TRUNC | O_WRONLY;
 	if (!mode)
 		mode = GIT_FILEMODE_BLOB;
+
+#ifdef GIT_WIN32
+	if (strstr(path, ".git") == NULL) {
+		if (git_win32__ishidden(path) == 1) {
+			if (git_win32__setvisible(path) == -1) {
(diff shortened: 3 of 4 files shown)


carlosmn:
git does add these files, which is not the most obvious thing it would do, but it looks like its environment doesn't respect the hidden flag, so for compatibility, it looks like we should not honour it either (though I wonder whether it's a conscious choice for git).

Unhiding the file, even if for a short time, feels like the wrong approach. We would also have to make sure we set it back to hidden in the error paths, opening yet another can of worms. Is there no other possibility of reading a hidden file?

Fallso:
I agree, but that approach seemed the least intrusive to that part of the codebase. You can read hidden files using the Windows APIs (CreateFile, BackupRead) and we could change the low level functions in posix_w32.c to use that approach but it'd be a big change and would require lots of testing.

carlosmn:
Such a minimal-change approach is fine for a quick fix if you need it in production, but we do need to take the longer approach for merging into mainline. If we can ask Windows to read the hidden files for us, then that's what we should do.

carlosmn:
There's something I can't quite figure out about the effects you're describing the PR description. Both git and libgit2 hide `.git` in order to simulate the effects of a leading dot on unix. This means that every time we run a status, we see a hidden file, and it's been working fine for years. So I wonder if this code is papering over a different issue you're seeing that happens to expose itself via the hidden files.

ethomson:
I'm also a bit curious about the mechanism by which this is failing.  A unit test illustrating this problem (that fails before this change and passes after) would be helpful both in providing some more information on this issue and ensuring that we don't regress.

Fallso:
I'm not sure about *nix, but on Windows the files in the git folder are not
hidden, even though the folder is (due to not having inherited attributes).
It could be that the file open/read mechanisms on *nix allow you to read
hidden files. Windows does odd things sometimes. I'll experiment with
replacing _wopen tomorrow.

Also: good idea about the unit test! Completely forgot about that

src/win32/w32_util.h
```diff
@@ -48,6 +48,22 @@ bool git_win32__findfirstfile_filter(git_win32_path dest, const char *src);
 int git_win32__sethidden(const char *path);
 
 /**
+* Checks if the given path (file or folder) has the +H (hidden) attribute set.
+*
+* @param path The path which should receive the +H bit.
+* @return True if hidden, false if visible
+*/
+int git_win32__ishidden(const char *path);
```
joonhur1234: why we need this line

src/filebuf.c
```diff
@@ -86,7 +101,11 @@ static int lock_file(git_filebuf *file, int flags, mode_t mode)
 		}
 
 		p_close(source);
-
```
joonhur1234: please bring this line back

src/win32/w32_util.h
Fallso: It's a function declaration for a function in w32_util.c

ethomson:
I've added unit tests to try to understand the failure here - that's pull request #3352 - and I do not see that adding hidden files fails.

Does #3352 fail for you?  If so, could you tell me more about your operating system setup?  (Windows version, etc).

If not, can you please provide some unit tests that fail before your change and succeed after your change?

src/filebuf.c
```diff
@@ -71,6 +71,21 @@ static int lock_file(git_filebuf *file, int flags, mode_t mode)
 		char buffer[FILEIO_BUFSIZE];
 		ssize_t read_bytes;
 
+#ifdef GIT_WIN32
+		bool hidden = false;
+		if (strstr(file->path_original, ".git") == NULL) {
```
ethomson: I don't think that you want to use `strstr` here, by the way.  That will find a `.git` anywhere in the string, no?  You actually only care about the `.git` directory, which will take a more nuanced approach to consider.

ethomson:
Is it possible that you're on a network filesystem that is causing some confusion?  Perhaps one that fakes the hidden attribute and enforces it differently than a normal Windows volume would?  (Samba perhaps?)

src/filebuf.c
Fallso: Indeed it will, do you think we could change this to:

strstr(file->path_original, "/.git")

to solve that issue?

Fallso:
I'll try the unit test you've put up when I get a moment, I haven't had a lot of time to look at this unfortunately! I saw the problem just on a standard NTFS Windows drive, so I'll attempt to recreate. 

I suppose it could be the version of visual studio I'm compiling with - when I was fiddling with various wide character conversion functions I noticed they behaved differently when compiled under VS2008 then when compiled under VS2013, so it's possible that this is a similar problem.

whoisj:
I spoke with the NTFS guys (at Microsoft) and there are settings for NTFS/Win32 which can cause failures. Per the NTFS experts:

> Yes, there is a scenario where “access denied” can be returned from a create IRP because of the hidden flag.  If you are trying to overwrite or supersede a file that has the hidden and/or system attributes set and you do not specify those same attribute flags on the create operation “access denied” will be returned.

ethomson:
So you're saying this problem arises trying to write the index when the index is the hidden file?

whoisj:
> So you're saying this problem arises trying to write the index when the index is the hidden file?

No I'm being stupid. I'm looking at several similar issues and replying to the wrong one :disappointed:

Fallso:
I've run your unit test based on the current content of the master branch and indeed it does pass...

I might have to go back to the drawing board with this one and attempt to recreate the issue in v0.22.0 which is where I saw it first. Strangely I swear I saw the issue the first time I migrated to v0.23.0 as well but I'm now unable to reproduce it.

ethomson:
I suspect that @whoisj was correct and that the actual files in your `.git` folder were hidden.  Sorry I'm not in front of a Windows machine to try this at the moment.

ethomson:
We've built test cases that I would have expected to illustrate a failure to handle hidden files, both in #3352 and in #3415.  I cannot find any issues with our existing functionality.

If you can get these to fail on your machine, that would be helpful in understanding this.

Ultimately, though, I agree with @carlosmn that this implementation is not the ideal way to move forward.  Once we understand when and how there are failures with hidden files, we can look at some alternate mechanisms.  In the mean time, I'm going to close this - please do share any further information you have, though.

PR #3312, opened 2015-07-15
Quick diff

Fallso:
Sped up diffing large work trees when using non-fnmatch pathspecs to see if the path we're analysing is part of a provided pathspec before recursion.

Rationale: On my system I had a folder with over 100,000 files in spread over 100s of folders - diffing and index status checks took forever. Now it doesn't!

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -1014,9 +1014,9 @@ static int handle_unmatched_new_item(
 		/* check if user requests recursion into this type of dir */
 		recurse_into_dir = contains_oitem ||
 			(delta_type == GIT_DELTA_UNTRACKED &&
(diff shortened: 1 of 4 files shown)


carlosmn:
There's a few lines of unrelated whitespace change, could you remove those?

Fallso:
I noticed those in the diff but they didn't appear changed in my editor.
I'll check it out tomorrow in another editor

Fallso:
I've removed the whitespace now

ethomson:
Thanks for the pull request.  This is indeed unnecessarily slow in the current code and your change will help speed that up by avoiding unnecessary pathspec matching.

A still more advance option is #3402 - which *also* avoids iterating through entries that are not in the path list, which is a further win for machines with slow filesystems.

I'm closing this in favor of that, but thanks again.

Fallso:
No problem, glad I could help!

PR #3320, opened 2015-07-22
Match git.git default break rewrite threshold

Therzok:


Therzok:
Excerpt from `git help diff`
> When used with -M, a totally-rewritten file is also considered as the source of a rename (usually -M
> only considers a file that disappeared as the source of a rename), and the number n controls this
> aspect of the -B option (defaults to 50%).

Therzok:
And I read the wrong segment.

PR #3322, opened 2015-07-22
Increment `git__n_inits` before doing `init_once`.

### LINKED ISSUE #3318: Errors can happen during init, before `git__n_inits` is initialized, which leads to a failed assertion
1. https://github.com/libgit2/libgit2/pull/3288 introduced `git__getenv`, which introduced a bit more error handling in environment lookup than we had before.
3. In POSIX, `git_libgit2_init` calls `init_once` before incrementing `git__n_inits`. (https://github.com/libgit2/libgit2/blob/42156d561723e92ffe597885719aa63abfe0795c/src/global.c#L347)
4. `init_once` calls various other init functions: https://github.com/libgit2/libgit2/blob/42156d561723e92ffe597885719aa63abfe0795c/src/global.c#L334-338
5. Some of those global init functions (namely `git_sysdir_global_init`) can error (https://github.com/libgit2/libgit2/blob/ea445e06028a1192e7e948cc81c87c4b39bc446e/src/sysdir.c#L34), which means we try to clear the `last_error`.
6. Setting the `last_error` accesses `git__global_state` (https://github.com/libgit2/libgit2/blob/c7f94123569e8fe00ffb3a35e6a12b6ebe9320ec/src/errors.c#L108).
6. `git__global_state` asserts that `git__n_inits > 0` (https://github.com/libgit2/libgit2/blob/42156d561723e92ffe597885719aa63abfe0795c/src/global.c#L383)
7. :boom: We fail the assertion.

joshaber:
Fixes #3318.


joshaber:
>Note that win32 has the same issue.

Yeah I noticed this. At worst, we're fixing #3318 and bring parity to both platforms :stuck_out_tongue_winking_eye:

More seriously, what you raise is certainly an issue but it predates this AFAICT. We don't check `init_error` before shutting down, all we do is check `git__n_inits`. So even if `init_once` failed, shutdown would still proceed as if it succeeded.

ethomson:
I agree.  I think that we should merge this just to get you unwedged, but we need to address the error cases in `init`/`shutdown` separately.

carlosmn:
Sure, the main use-case is to initialize the library at the start of a program, so there shouldn't be many situations where the race condition can happen.

But it looks like we have to split this into two blocks. An early init would set up the TLS data and whatever else we need to report errors. We can then bump up the inits and do the rest of the initialization via a second init (anybody else having flashbacks to OS class?).

joshaber:
@carlosmn Are you saying that should happen in this PR, or can this be merged as-is to solve the immediate problem?

carlosmn:
That's for later, it looks like I didn't write down the timeline. We can merge this as-is but this can't be the state for a release.

joshaber:
:+1:

Issue #3313, opened 2015-07-17
Crash when shutting down with OOM as last error


carlosmn:
During the library shut-down, we free the last error's memory (particularly the error string). The OOM error is however a static buffer (to avoid allocations), so we end up trying to free our own rodata, thus crashing.

It should be a matter of checking whether `last_error` is this buffer or something similar and not doing the free then.

I'm leaving this as up-for-grabs as it's not a crashing bug on most systems, which overcommit memory, but it's still something to fix.

carlosmn:
Fixed with #3351

PR #3323, opened 2015-07-23
Fix attempt to free static OOM error message

mhp:
I noticed a mechanism that can lead to attempts to free the statically allocated OOM error message (whilst trying to reproduce #3313).  The function `giterr_detach` assumes that any error message it detaches is dynamically allocated.  This PR fixes that, so that it becomes safe to detach an OOM error.

I have also added a test that performs the full `giterr_capture`/`giterr_restore` cycle, checking to ensure that the static pointer is not saved in the `git_error_state` structure.

Comments welcome!

### DIFF
--- a/include/git2/errors.h
+++ b/include/git2/errors.h
@@ -114,18 +114,6 @@ GIT_EXTERN(const git_error *) giterr_last(void);
 GIT_EXTERN(void) giterr_clear(void);
 
 /**
- * Get the last error data and clear it.
- *
- * This copies the last error into the given `git_error` struct
- * and returns 0 if the copy was successful, leaving the error
- * cleared as if `giterr_clear` had been called.
- *
- * If there was no existing error in the library, -1 will be returned
- * and the contents of `cpy` will be left unmodified.
- */
-GIT_EXTERN(int) giterr_detach(git_error *cpy);
-
-/**
  * Set the error message string for this thread.
  *
  * This function is public so that custom ODB backends and the like can
--- a/src/errors.c
+++ b/src/errors.c
@@ -116,7 +116,7 @@ void giterr_clear(void)
 #endif
 }
 
-int giterr_detach(git_error *cpy)
+static int giterr_detach(git_error *cpy)
 {
 	git_error *error = GIT_GLOBAL->last_error;
 
--- a/src/errors.c
+++ b/src/errors.c
@@ -125,14 +125,12 @@ static int giterr_detach(git_error *cpy)
 	if (!error)
 		return -1;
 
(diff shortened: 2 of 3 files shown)


ethomson:
Yes, this is definitely problematic.  I noticed similar problems in #3257, though it's not appropriate for inclusion here.  I think there may be some possibility for leaking here also since we don't provide a way to free the detached buffer.

In any case, if you're hitting an OOM then I don't like to assume that a `strdup` will succeed.  Can we return the buffer and test to make sure that we don't try to free the static OOM string?  Or if the `strdup` fails, are we left in an okay place?

mhp:
I don't think I've increased the leak risk - as far as I can see, the detached buffer is expected to be dynamically allocated.  I didn't see a disposal function anywhere (`giterr_free()` perhaps?), so I assume users of this function know to dispose of the buffer themselves.

You are right that `strdup` seems likely to fail under oom conditions.  I believe that this would end up storing a null pointer in the structure.  This isn't ideal, but the `klass` is still valid for diagnostics to some degree.

Returning the static buffer pointer, and checking when freeing would probably work very well for the `giterr_capture`/`giterr_restore` cycle, but I don't think it would work if just the `giterr_detach` function is used, simply because we don't have a free-like function to handle the disposal side.

Have I misinterpreted the semantics of the `giterr_detach` function?  If there is an approved disposal route, we can add the pointer matching there to make it all work.

ethomson:
No, I don't think you've increased leakiness either, I was just pointing out that we're leaky now.

My opinion is that [`giterr_detach`](`https://github.com/libgit2/libgit2/blob/master/include/git2/errors.h#L125) shouldn't be public at all.  We offer callers the error information, I don't see the value in a single call that gets and clears it, with no way for the caller to free that information safely.

To clarify:  you're fixing this because the library is buggy and ended up thrashing your app, correct?  (That is to say, you don't actually want to call `giterr_detach` from your code, and you wouldn't mind seeing it disappear?)

mhp:
That's right - I'm not interested in `giterr_detach`.  I'd like to be able to `giterr_capture`/`giterr_restore` an OOM condition without freeing static data though!

Would you like me to rework this so that `giterr_detach` is no longer public and the static message is handled correctly without attempting to `strdup` it?

mhp:
I've removed `giterr_detach` from the public API, and fixed `giterr_restore` to notice if an OOM condition is being restored.  I think this fixes the problems I saw initially (attempt to free static memory) without resorting to `strdup`!

What do you think, @ethomson?  I could also move the body of `giterr_detach` into `giterr_capture` if you think this is the right direction to move in.

PR #3326, opened 2015-07-25
Fix git_index_add_bypath() crash if git_submodule_lookup() fails

linquize:
Fix `git_index_add_bypath()` crash if `git_submodule_lookup()` fails other than `GIT_ENOTFOUND`

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -1254,6 +1254,8 @@ int git_index_add_bypath(git_index *index, const char *path)
 			return giterr_restore(&err);
 		else
 			git__free(err.error_msg.message);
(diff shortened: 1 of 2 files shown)


carlosmn:
Can we get a test that would cause the lookup function to return something else?

linquize:
test added @ 1eff7c9.

carlosmn:
In the test case you've added, doesn't git add the repository under `not-submodule` as a submodule entry in the index? We should do the same in that situation, rather than error out.

ethomson:
Thanks!  Fixed via #3344

PR #3327, opened 2015-07-25
path: convert diropen's ENOENT to GIT_ENOTFOUND

carlosmn:
If a directory is removed between the listing of its parent and the
attempt to traverse it, we need to indicate that through our error code
so the caller knows to continue.

The Windows code seems to have the same issue, but I don't know if it'd be ok to return `ENOTFOUND` every time we get `INVALID_HANDLE_VALUE` (or maybe we should just return that every time.

Or maybe it'd be a better idea/patch to simply treat all errors from `git_path_diriter_init()` as "oh well, let's try the next one" when we're in the iterator. That would also let us ignore permissions issues, which I think would also cause an error right now.

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -1260,10 +1260,15 @@ int git_path_diriter_init(
 	}
 
 	if ((diriter->dir = opendir(diriter->path.ptr)) == NULL) {
+		int error = -1;
+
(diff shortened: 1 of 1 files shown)


carlosmn:
Closing in favour if #3328

PR #3332, opened 2015-07-28
Resolve documentation warnings

phatblat:
Also enabled `-Wdocumentation` if supported

### DIFF
--- a/src/push.h
+++ b/src/push.h
@@ -83,7 +83,7 @@ int git_push_add_refspec(git_push *push, const char *refspec);
  * Update remote tips after a push
  *
  * @param push The push object
- * @param signature The identity to use when updating reflogs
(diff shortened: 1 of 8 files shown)


src/push.h
```diff
@@ -83,7 +83,7 @@ int git_push_add_refspec(git_push *push, const char *refspec);
  * Update remote tips after a push
  *
  * @param push The push object
- * @param signature The identity to use when updating reflogs
+ * @param callbacks The identity to use when updating reflogs
```
carlosmn: This fixes it syntactically, but this isn't a copy-paste error; it's documentation for an old parameter. The whole line should be replaced by something relevant to the new function signature.

phatblat: Fixed.

PR #3333, opened 2015-07-28
Maintenance updates for v0.23

carlosmn:
Most changes in master are bug fixes, so we can bring them in for v0.23.1 without problems with the backporting.

The xdiff update isn't purely bug fixes, but it does enable a rather important bug.

The one I'm not sure about is the `git_index_add()` behaviour with submodules. It does change the behaviour somewhat, but what we currently do is error out unnecessarily.


tkelman:
37c84dc58f43c609ddadff3f9ff25c1f5b52a746 maybe? Anyone who may have tried with any older cmake version would have hit an error due to the unsupported policy line, so it would be mostly for honesty/OCD's sake.

arthurschreiber:
> The xdiff update isn't purely bug fixes, but it does enable a rather important bug.

Oooh, we're enabling bugs now. :bomb:  :smiley:

carlosmn:
I suppose we could include the CMake one for completeness' sake.

@arthurschreiber you know I'm all about enabling bugs

PR #3335, opened 2015-07-28
Changed proxy lookup to mimic git's behavior

Corillian:
Proxy lookup was only looking for remote.<name>.proxy and http.proxy
within the config whereas git also looks up http.<url>.proxy. This
commit adds support for http.<url>.proxy and fixes an issue with
building on windows caused by two source files named buffer.c


carlosmn:
The `buffer.c` issue is completely unrelated to the purpose of this PR. Please undo it.

ethomson:
Indeed, we'd like to have these two changes separate.  But thank you for doing this - in a non-cmake build system (where the obj-files all get dumped into the same folder) this is definitely a problem.  Thanks!

ethomson:
As for the change itself, is my reading of the configuration correct, that there are actually many (4?) possible `http.*.proxy` lookups per URL?

src/remote.c
```diff
@@ -804,7 +851,8 @@ int git_remote__get_http_proxy(git_remote *remote, bool use_ssl, char **proxy_ur
 	}
 
 	/* HTTP_PROXY / HTTPS_PROXY environment variables */
-	error = git__getenv(&val, use_ssl ? "HTTPS_PROXY" : "HTTP_PROXY");
+	git_buf_init(&buf, 0);
```
carlosmn: The buffer has already been initialised. If what you wanted was to clear its contents, use `git_buf_clear()`.

carlosmn:
The logic to determine which entry to use seems a bit hard to follow but it looks like you're looking for the hostname. We have `git_connection_data` which we can parse a url into, which provides names for all the different information it contains.

Corillian:
I don't believe gitno_connection_data is a good solution as it only works with http and https and it explicitly assigns a default port. I've done some further testing this morning and git actually does 2 checks if the URL contains a port - 1 with the port and if a proxy isn't found it does a second without the port.  I'll update this pull request with additional logic to handle that and I'll rename buffer_win32.c back to buffer.c and submit a separate pull request with it named buffer_win32.c.

Lastly I'll remove the git_buf_init()'s as it appears git_buf_free() calls git_buf_init().

ethomson:
>  I've done some further testing this morning and git actually does 2 checks if the URL contains a port

Again, my reading of `git-config(5)` indicates that it does more than that:

> **http.< url >.***
Any of the http.* options above can be applied selectively to some URLs. For a config key to match a URL, each element of the config key is compared to that of the URL, in the following order:
>
> 1. Scheme (e.g., https in https://example.com/). This field must match exactly between the config key and the URL.
>
> 2. Host/domain name (e.g., example.com in https://example.com/). This field must match exactly between the config key and the URL.
>
> 3. Port number (e.g., 8080 in http://example.com:8080/). This field must match exactly between the config key and the URL. Omitted port numbers are automatically converted to the correct default for the scheme before matching.
>
> 4. Path (e.g., repo.git in https://example.com/repo.git). The path field of the config key must match the path field of the URL either exactly or as a prefix of slash-delimited path elements. This means a config key with path foo/ matches URL path foo/bar. A prefix can only match on a slash (/) boundary. Longer matches take precedence (so a config key with path foo/bar is a better match to URL path foo/bar than a config key with just path foo/).
>
> 5. User name (e.g., user in https://user@example.com/repo.git). If the config key has a user name it must match the user name in the URL exactly. If the config key does not have a user name, that config key will match a URL with any user name (including none), but at a lower precedence than a config key with a user name.
>
> The list above is ordered by decreasing precedence; a URL that matches a config key’s path is preferred to one that matches its user name. For example, if the URL is https://user@example.com/foo/bar a config key match of https://example.com/foo will be preferred over a config key match of https://user@example.com.
>
> All URLs are normalized before attempting any matching (the password part, if embedded in the URL, is always ignored for matching purposes) so that equivalent URLs that are simply spelled differently will match properly. Environment variable settings always override any matches. The URLs that are matched against are those given directly to Git commands. This means any URLs visited as a result of a redirection do not participate in matching.

Though I haven't read this in depth to grok it in fullness.  Can you clarify the above w/r/t your changes?  Are your changes, in fact, complete per the configuration settings?

Corillian:
My changes were based on me noticing that regular git was able to clone just fine within my company's firewall and libgit2 was not. I was unaware that the documentation for git-config contained such detailed information for proxy lookup, thanks for pointing it out, and I will update my implementation accordingly. Full support looks like it's going to be significantly more complicated than my current implementation.

ethomson:
Awesome.  Thank you!  I read that several times and wasn't certain if I was able to really make heads or tails of it, I assumed that I would either need to dive into the `git` source code or blast a bunch of test cases at it to understand it completely.  I'm glad this makes sense to you.  :sparkles: 

Thanks again!

Corillian:
Is there anything else that needs to be done for this pull request? Looking at the CI failure I'm a bit confused as it failed in the vs2012 build on code in gitno_extract_url_data_parts() which is identical to gitno_extract_url_parts() which compiled just fine. I'm running vs2013 and I don't have any problems compiling netops.c. I can cast the pointers to size_t for the arithmetic if you guys want but it seems something is/was messed up with the CI build?

Corillian:
Hey guys I know you're busy but I just wanted to follow up again to see if there's anything else you need me to do for this pull request?

src/netops.c
```diff
@@ -280,3 +280,89 @@ int gitno_extract_url_parts(
...
+{
+	struct http_parser_url u = { 0 };
+	const char *_userinfo;
+
+	if(http_parser_parse_url(url, strlen(url), false, &u)) {
+		giterr_set(GITERR_NET, "Malformed URL '%s'", url);
+		return GIT_EINVALIDSPEC;
+	}
+
+#define GITNO_COPY_URL_PART(dest, field) \
```
carlosmn: The code here needs to be in a `do { } while (0)` block to guard against uses with optional braces.

src/netops.c
```diff
@@ -280,3 +280,89 @@ int gitno_extract_url_parts(
...
+	assert(data && url);
+
+	gitno_url_data_free_ptrs(data);
+
+	return gitno_extract_url_data_parts(&data->scheme, &data->host, &data->port, &data->path, &data->user, &data->pass, &data->query, url);
+}
+
+void gitno_url_data_free_ptrs(gitno_url_data *data)
+{
+#define GITNO_SAFE_FREE(ptr) if(ptr) { git__free(ptr); (ptr) = NULL; }
```
carlosmn: This seems overly elaborate. Freeing plus a `memset()` would be just as effect and more straightforward to read.

src/netops.c
```diff
@@ -280,3 +280,89 @@ int gitno_extract_url_parts(
...
+	return 0;
+}
+
+int gitno_url_data_from_string(
+	gitno_url_data *data,
+	const char *url)
+{
+	assert(data && url);
+
+	gitno_url_data_free_ptrs(data);
```
carlosmn: I don't think we should be freeing anything here. We do not own the memory, and the user is reponsible for freeing the memory. A `memset()` to make sure we only return data we parsed would be more correct here.

carlosmn:
A general note on style: there seem to be a lot of camel-cased variables, which goes against this project's style. That would have to be changed before this goes in.

carlosmn:
This is also a sufficiently complex piece of functionality that tests are rather necessary.

src/netops.c
Corillian: The freeing of the memory mimics the behavior of gitno_connection_data_from_url()

src/netops.c
Corillian: I'm not entirely sure I understand what you're saying here. My assumption is that you're suggesting I wrap the if statement in a do...while(0) to prevent someone from accidentally deleting the braces thus causing it to always return GIT_EINVALIDSPEC. It seems like a far more pragmatic approach would be to require all control flow statements to use braces instead of sprinkling do...while(0)'s around everywhere - this is a well established requirement for preventing the aforementioned bug when writing safety critical software.

ethomson:
There hasn't been any movement on this in over a year, I'm going to close this issue.  Thanks, and please do comment if you'd like to continue this.

PR #3336, opened 2015-07-28
Make libgit2 add old-style submodule without .gitmodules

linquize:
Make libgit2 add old-style submodule (with its own .git dir) without `.gitmodules`.
core-git allows this.


carlosmn:
Returning a value on an error is surprising and not something which we do anywhere else AFAIK, so doing it for this one situation is bound to cause a lot of confusion. It is also achieving the result of adding the submodule to the index in what feels a very sneaky way.

ethomson:
Thanks for this - fixed via #3344

Issue #3338, opened 2015-07-29
cred_aquire_cb never called for https://hub.jazz.net (IBM DevOps Services)


palmin:
I stumbled upon some bad behaviour but I am not sure if the fault is with libgit2 or IBM DevOps Services.

When trying to clone anything from ````https://hub.jazz.net```` the callback to acquire credentials is never called. I am trying with latest commit on master as of this writing (2dfd5ea). The problem can be replicated with the following example:
````
int cred_acquire_cb(git_cred **cred, const char *url, const char *username_from_url,
				    unsigned int allowed_types, void *payload) {
	return git_cred_userpass_plaintext_new(cred, "username", "password");
}

int main() {
	git_libgit2_init();

	git_clone_options clone_options = GIT_CLONE_OPTIONS_INIT;
	clone_options.fetch_opts.callbacks.credentials = cred_acquire_cb;

	git_repository *repo = NULL;
	const char *url = "https://hub.jazz.net/git/palmin/Test";
	const char *path = "Test.git";
	int error = git_clone(&repo, url, path, &clone_options);
	fprintf(stderr, "error = %d, %s\n", error, giterr_last()->message);
	
	return 0;
}
````
This ends up writing:
````
error = -1, Unexpected HTTP status code: 401
````

Digging a little digger it is because the ````WWW-Authenticate```` header is never found in [on_header_ready](https://github.com/libgit2/libgit2/blob/master/src/transports/http.c#L258). It seems to be given in lowercase. Checking with ````strcasecmp```` instead of ````strcmp```` fixes the problem and does not brake any tests. The checks for ````Location````and ````Content-Type````already use  ````strcasecmp````.

Just let me know if you want a pull-request with the given change.

ethomson:
Doh!  Thanks so much for the very detailed investigation and report.  Yes, a pull request would be quite welcome!

palmin:
I will close now that the pull-request has been processed.
Thank you for the speedy response, Edward.

michael-w3:
Any chance this simple fix can be merged in into a new 0.22.x maintenance release?

ethomson:
I don't anticipate a forthcoming 0.22 maintenance release.

PR #3339, opened 2015-07-29
case-insensitive check for WWW-Authenticate header

palmin:
Fixes issue #3338


ethomson:
This is most hilarious because it's sandwiched between two tests that actually do a `strcasecmp`.

PR #3344, opened 2015-08-01
index: stage an unregistered submodule as well

carlosmn:
index: stage an unregistered submodule as well
    
We previously added logic to `_add_bypath()` to update a submodule. Go
further and stage the submodule even if it's not registered to behave
like git.

This is an alternative to #3326 and #3336 which mimics git's behaviour while not changing the behaviour of the submodule API, so we can easily backport it.

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -1228,6 +1228,45 @@ int git_index_add_frombuffer(
 	return 0;
 }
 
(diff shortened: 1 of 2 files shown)


linquize:
LGTM

Issue #3345, opened 2015-08-01
ssh:// url with homedir-relative path (~) not supported.


MrHacky:
An ssh url like 'ssh://user@server/~/path/repo.git' is supported by git and will connect to the repository at '/home/user/path/repo.git' on 'server'.

However libgit2 gives an error for such urls: '/~/path/repo.git' does not appear to be a git repository.

It would be nice to have this working.

(workaround: the scp-style url 'user@server:path/repo.git' does work)

carlosmn:
Shell expansion is not really libgit2's job. git itself does not do it either. It passes the tilde to ssh as well (on my versions). I've just tried and the tilde is passed all the way to the receiving `git-upload-pack` so it may be that your shell has been doing the expansion. If you run the git command with `GIT_TRACE=1` set, does it send the tilde through or is your shell doing the expansion beforehand for you?

MrHacky:
indeed the tilde is passed all the way through to upload-pack:

```
$ GIT_TRACE=1 git clone ssh://user@server/~/path/repo.git
trace: built-in: git 'clone' 'ssh://user@server/~/path/repo.git'
trace: run_command: 'ssh' 'user@server' 'git-upload-pack '\''~/path/repo.git'\'''
```

However note that the path passed to upload-pack starts with the '~' character, compare with this where i replaced the '~' with an 'x':
```
$ GIT_TRACE=1 git clone ssh://user@server/x/path/repo.git
trace: built-in: git 'clone' 'ssh://user@server/x/path/repo.git'
trace: run_command: 'ssh' 'user@server' 'git-upload-pack '\''/x/path/repo.git'\'''
```
Now the path passed starts with the '/' character.

Possibly what git is doing is stripping the leading '/' if the next character is a '~', which lets git-upload-pack correctly interpret the path on the server.

MrHacky:
In fact, that is exacty what git is doing, see:
https://github.com/git/git/blob/master/connect.c#L639
```
if (path[1] == '~')
	path++;
```

So it would make sense for libgit2 to do this as well, right?

carlosmn:
Then we'll have to provide the same workaround.

ethomson:
Indeed, this seems straightforward.  @mrhacky is this something you might be able to send a pull request for?

MrHacky:
@ethomson yeah i'll give it a shot.

ethomson:
Closed via #3348 

Thanks for the helpful report and the pull request!

PR #3348, opened 2015-08-03
Handle ssh:// and git:// urls containing a '~' character.

MrHacky:
For such a path '/~/...' the leading '/' is stripped so the server will
get a path starting with '~' and correctly handle it.

solution for issue #3345


ethomson:
Great catch!  Thanks so much for the pull request!  :sparkles:

PR #3352, opened 2015-08-03
win32: ensure hidden files can be staged

ethomson:


carlosmn:
So I understand from this that we *can* read and stage hidden files regardless of the flag in question (as soon as we ignore these tests on unix).

ethomson:
Yes, that's certainly what I'm seeing on my machine.  (And those are ifdef'd out on unix now.)

ethomson:
Although this doesn't illustrate the problem reported in #3310, or really any problem at all, I think that this is worth merging to prevent truly bizarre regressions in the future.

PR #3353, opened 2015-08-04
index: canonicalize directory case when adding

ethomson:
On case insensitive systems, when given a user-provided path in the
higher-level index addition functions (eg `git_index_add_bypath` /
`git_index_add_frombuffer`), examine the index to try to match the
given path to an existing directory.

Various mechanisms can cause the on-disk representation of a folder
to not match the representation in HEAD or the index - for example,
a case changing rename of some file `a/file.txt` to `A/file.txt`
will update the paths in the index, but not rename the folder on
disk.

If a user subsequently adds `a/other.txt`, then this should be stored
in the index as `A/other.txt`.

For example:

    % git init .
    % mkdir a
    % echo hello > a/file.txt
    % git add a
    % git commit -mfoo
    % git mv -f a/file.txt A/file.txt
    % git status
    On branch master
    Changes to be committed:

            renamed:    a/file.txt -> A/file.txt

    % dir
    ...snip...
    08/04/2015  04:54 PM    <DIR>          a

    % echo hi > a/other.txt
    % git add a/other.txt
    % git status
    On branch master
    Changes to be committed:

            new file:   A/other.txt

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -1102,6 +1102,74 @@ static int check_file_directory_collision(git_index *index,
 	return 0;
 }
 
+static int canonicalize_directory_path(
+	git_index *index, git_index_entry *entry)
+{
+	const git_index_entry *match, *best = NULL;
+	char *search, *sep;
+	size_t pos, search_len, best_len;
+
+	if (!index->ignore_case)
+		return 0;
+
+	/* item already exists in the index, simply re-use the existing case */
+	if ((match = git_index_get_bypath(index, entry->path, 0)) != NULL) {
+		memcpy((char *)entry->path, match->path, strlen(entry->path));
+		return 0;
+	}
+
+	/* nothing to do */
+	if (strchr(entry->path, '/') == NULL)
+		return 0;
+
+	if ((search = git__strdup(entry->path)) == NULL)
+		return -1;
+
+	/* starting at the parent directory and descending to the root, find the
+	 * common parent directory.
+	 */
+	while (!best && (sep = strrchr(search, '/'))) {
+		sep[1] = '\0';
+
+		search_len = strlen(search);
+
+		git_vector_bsearch2(
+			&pos, &index->entries, index->entries_search_path, search);
+
+		while ((match = git_vector_get(&index->entries, pos))) {
+			if (GIT_IDXENTRY_STAGE(match) != 0) {
+				/* conflicts do not contribute to canonical paths */
+			} else if (memcmp(search, match->path, search_len) == 0) {
+				/* prefer an exact match to the input filename */
+				best = match;
+				best_len = search_len;
+				break;
+			} else if (strncasecmp(search, match->path, search_len) == 0) {
+				/* continue walking, there may be a path with an exact
+				 * (case sensitive) match later in the index, but use this
+				 * as the best match until that happens.
+				 */
+				if (!best) {
+					best = match;
+					best_len = search_len;
+				}
+			} else {
+				break;
+			}
+
+			pos++;
+		}
+
+		sep[0] = '\0';
+	}
+
+	if (best)
+		memcpy((char *)entry->path, best->path, best_len);
+
+	git__free(search);
+	return 0;
+}
+
 static int index_no_dups(void **old, void *new)
 {
 	const git_index_entry *entry = new;
@@ -1115,10 +1183,17 @@ static int index_no_dups(void **old, void *new)
  * it, then it will return an error **and also free the entry**.  When
  * it replaces an existing entry, it will update the entry_ptr with the
  * actual entry in the index (and free the passed in one).
+ * trust_path is whether we use the given path, or whether (on case
+ * insensitive systems only) we try to canonicalize the given path to
+ * be within an existing directory.
  * trust_mode is whether we trust the mode in entry_ptr.
  */
 static int index_insert(
-	git_index *index, git_index_entry **entry_ptr, int replace, bool trust_mode)
+	git_index *index,
+	git_index_entry **entry_ptr,
+	int replace,
+	bool trust_path,
+	bool trust_mode)
 {
 	int error = 0;
 	size_t path_length, position;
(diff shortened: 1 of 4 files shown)


ethomson:
Haha, it turns out that I had my last-minute "only run this on `core.ignorecase` systems" check backwards.  Oops!  Corrected, I think that this is ready for review.

tests/index/bypath.c
```diff
@@ -46,3 +46,150 @@ void test_index_bypath__add_submodule_unregistered(void)
...
+	dummy.path = "just_A_dir/fileD.txt";
+	cl_git_pass(git_index_add(g_idx, &dummy));
+
+	dummy.path = "JUST_a_DIR/fileE.txt";
+	cl_git_pass(git_index_add(g_idx, &dummy));
+
+	cl_git_mkfile("submod2/just_a_dir/file1.txt", "This is a file");
+	cl_git_mkfile("submod2/just_a_dir/file2.txt", "This is another file");
+	cl_git_mkfile("submod2/just_a_dir/file3.txt", "This is another file");
+	cl_git_mkfile("submod2/just_a_dir/file4.txt", "And another file");
```
carlosmn: This entry isn't looked up in the assertions below.

tests/index/bypath.c
```diff
@@ -46,3 +46,150 @@ void test_index_bypath__add_submodule_unregistered(void)
...
+	cl_git_pass(git_index_add(g_idx, &dummy));
+
+	dummy.path = "JUST_A_DIR/fileC.txt";
+	cl_git_pass(git_index_add(g_idx, &dummy));
+
+	dummy.path = "just_A_dir/fileD.txt";
+	cl_git_pass(git_index_add(g_idx, &dummy));
+
+	dummy.path = "JUST_a_DIR/fileE.txt";
+	cl_git_pass(git_index_add(g_idx, &dummy));
```
carlosmn: These entries aren't checked anywhere. Is it just to have a bunch of entries before `_add_bypath()`?

src/index.c
```diff
@@ -1065,6 +1065,70 @@ static int check_file_directory_collision(git_index *index,
...
+	const git_index_entry *match, *best = NULL;
+	char *search, *sep;
+	size_t pos, search_len, best_len;
+
+	if (!index->ignore_case)
+		return 0;
+
+	/* item already exists in the index, simply re-use the existing case */
+	if ((match = git_index_get_bypath(index, entry->path, 0)) != NULL) {
+		memcpy((char *)entry->path, match->path, sizeof(entry->path));
```
carlosmn: You probably mean `strlen(entry->path)` here. `sizeof(entry->path)` is `sizeof(char *)`.

src/index.c
```diff
@@ -1065,6 +1065,70 @@ static int check_file_directory_collision(git_index *index,
...
+		while (match = git_vector_get(&index->entries, pos)) {
+			if (GIT_IDXENTRY_STAGE(match) != 0) {
+				/* conflicts do not contribute to canonical paths */
+			} else if (memcmp(search, match->path, search_len) == 0) {
+				/* prefer an exact match to the input filename */
+				best = match;
+				best_len = search_len;
+				break;
+			} else if (strncasecmp(search, match->path, search_len) == 0) {
+				if (!best) {
```
carlosmn: Writing this as `} else if (!best && strcasecmp(...) == 0) {` would provide the same semantics, but would avoid performing the string comparison in cases where we're not going to take it because we already have a best match.

src/index.c
```diff
@@ -1065,6 +1065,70 @@ static int check_file_directory_collision(git_index *index,
...
+	*/
+	while (!best && (sep = strrchr(search, '/'))) {
+		sep[1] = '\0';
+
+		search_len = strlen(search);
+
+		git_vector_bsearch2(
+			&pos, &index->entries, index->entries_search_path, search);
+
+		while (match = git_vector_get(&index->entries, pos)) {
```
carlosmn: Clang throws a warning here. With such a construct, it's also good for humans to wrap it in parens and/or compare against `NULL` explicitly.

tests/index/bypath.c
```diff
@@ -46,3 +46,150 @@ void test_index_bypath__add_submodule_unregistered(void)
...
+	cl_git_rewritefile("submod2/just_a_dir/file3.txt", "Rewritten");
+	cl_git_pass(git_index_add_bypath(g_idx, "Just_A_Dir/file3.txt"));
+
+	cl_assert(entry = git_index_get_bypath(g_idx, "Just_A_Dir/file3.txt", 0));
+	cl_assert_equal_s("just_a_dir/FILE3.txt", entry->path);
+}
+
+void test_index_bypath__add_honors_existing_case_2(void)
+{
+	git_index_entry dummy = {0};
```
carlosmn: These should have double braces; the first element is a struct.

src/index.c
carlosmn: Funnily enough, the tests which trigger this do have their change within the first 8 bytes :)

However, even making it so this `memcpy` fails to update the whole path, the tests do still succeed. So either we have similar logic elsewhere or the tests are deficient.

src/index.c
ethomson: In this case, we want to continue walking as long as `strncasecmp == 0` because an actual (case sensitive) match could be coming later, and we want to use that if it exists.  We merely mark this as the best we've seen yet.  I'll add some commentary here so that it's clearer.

tests/index/bypath.c
ethomson: It is.

ethomson:
So as you mentioned:  there is (was) another mechanism by which we would dup existing entries and keep the original filename as it existed in the index.

I actually think that this is probably not the generally correct behavior:  instead, I think that `git_index_add` should use the given path so that one could perform a case-changing rename.  However, I think that `git_index_add_bypath` and `git_index_add_frombuffer` should stay friendly to the callers and maintain the path as it exists in the index, if it's already there.

I laid down another commit to update this behavior.  Let me know if you don't agree with this plan.

carlosmn:
I do like that approach. When you ask to insert a specific entry we do that, otherwise we take the case-sensitivity of the worktree into account.

(you seem to have some conflicts, presumably due to the changelog stuff)

ethomson:
Rebased!  I had been waiting on this in case the hash affected anything here.  It doesn't, so this should be ready to go.

Issue #2830, opened 2015-01-20
clone from OpenShift accounts fail with "Found invalid hex digit in length" error


palmin:
I have trouble cloning from RedHat OpenShift accounts. They require public-key ssh authentication and if this succeeds the cloning stops with the error message:
 `Found invalid hex digit in length`

I have verified that command-line git will clone without problems. 

The following code will trigger the error for v0.16.0-5730-g2136240
````
int cred_acquire_cb(git_cred **cred, const char *url, const char *username_from_url,
				unsigned int allowed_types, void *payload) {
	return git_cred_ssh_key_new(cred, username_from_url, "id_rsa.pub", "id_rsa", NULL);
}

int main() {
	git_libgit2_init();
	git_clone_options clone_options = GIT_CLONE_OPTIONS_INIT;
	clone_options.remote_callbacks.credentials = cred_acquire_cb;

	git_repository *repo = NULL;
	const char *url = ...;
	int error = git_clone(&repo, url, "test.git", &clone_options);
	fprintf(stderr, "error = %d, %s\n", error, giterr_last()->message);
}
````

Accounts can be created free of charge at [OpenShift](http://openshift.redhat.com).

linquize:
Is it reproducible on libgit2 v0.22?

palmin:
Yes.
Just tried on the v0.22.1 release where it is reproducible.

palmin:
While debugging i found out that the hex-digit that cannot be parsed is in response to something like:
    
    "Invalid repository /~/git/test.git/: not in application root"
    
I did not include the full URL which is unfortunate as this was part of the problem. It seems ~ in the clone-URL is not properly translated into the home-directory. Specifying a clone-URL without ~ works without problems.

I will dig a little deeper and see if I can find out why ~ is not properly translated. I can see that home directories on OpenShift are in a non-standard location:

    /var/lib/openshift/

and perhaps the server-end spills a error-message into the ssh transport.

palmin:
I cannot see that libgit2 does anything wrong.

A bug request has been filed at bugzilla.redhat.com regarding OpenShift.

ethomson:
Thanks for the update.

It might be nice for us to include the four characters that we tried to parse as the hex length, since this is likely to be the usual culprit here - the server sending an error message that we're trying to parse as the length.  So it might be a little bit more obvious as to what's wrong if you saw "Failed to parse hex digits 'Erro' or 'Fail'...

palmin:
[...]

PR #3355, opened 2015-08-05
Include the 4 characters not recognised as hex-number in parse_len

palmin:
Improve error messages as described in issue #2830 by including the 4 characters not recognised as hexadecimal digit.

If there are any control characters among these, we stop at the first one found to avoid garbage in error string.

### DIFF
--- a/src/transports/smart_pkt.c
+++ b/src/transports/smart_pkt.c
@@ -351,7 +351,7 @@ static int unpack_pkt(git_pkt **out, const char *line, size_t len)
 static int32_t parse_len(const char *line)
 {
 	char num[PKT_LEN_SIZE + 1];
-	int i, error;
+	int i, k, error;
 	int32_t len;
 	const char *num_end;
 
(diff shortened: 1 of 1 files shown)


carlosmn:
I'd rather see a dot where we find a control character, like in hex displays. Otherwise we show different lengths for the input even though we always want to look at the same about in bytes. In fact, it's probably better to check the non-hex against something like `isprint()` and replace it with a dot if false. We might otherwise end up printing half of a unicode char.

src/transports/smart_pkt.c
```diff
@@ -360,7 +360,15 @@ static int32_t parse_len(const char *line)
...
-			giterr_set(GITERR_NET, "Found invalid hex digit in length");
+			/* Make sure there are no special characters before passing to error message */
+			for (k = 0; k < PKT_LEN_SIZE; ++k) {
+				if(iscntrl(num[k])) {
+					/* Truncate at first control character */
+					num[k] = '\0';
+				}
+			}
+			
+			giterr_set(GITERR_NET, "Invalid digit in hex length: %s", num);
```
carlosmn: This wording sounds odd. It is the digits which are hex; that adjective should not refer to the length. Since you're changing the string anyway, please make it start with a lowercase letter.

palmin: How does "invalid hex-digit in length: %s" look to you?

ethomson: While we're talking about this, I would like to see single quotes around the output, eg "invalid hex digit in length: '%s'".  (I don't need to see the "found" verb there, but I also admit that I don't have a serious objection.)

carlosmn:
Looks good, could you squash the fixup commits as this is a single logical change?

PR #3357, opened 2015-08-05
Renamed /src/win32/buffer.c to /src/win32/buffer_win32.c

Corillian:
This prevents issues with compiling using MSVC caused by multiple object
files with the same name.


Therzok:
Can you rebase this PR and force push?

ethomson:
Looks like great minds think alike!  #3360 also addressed this issue.  Since that one was cleanly based on master and was not re-adding prior commits, I went ahead and merged that one, but thank you!  Looking forward to future PRs.

Corillian:
No worries!

Issue #3356, opened 2015-08-05
Duplicate base names kills build on older versions of Visual Studio


johnhaley81:
So Visual Studio 2008 and before will ignore the base-names of files. We've seen this before in https://github.com/libgit2/libgit2/issues/330. This is something you guys might not really want to worry about since it is an older version of VS but it does however break toolchains like GYP.

The files in question are:

- https://github.com/libgit2/libgit2/blob/master/src/buffer.c
- https://github.com/libgit2/libgit2/blob/master/src/win32/buffer.c

ethomson:
Yes, this needs to be fixed.  Can you send a PR?

johnhaley81:
:+1: 

I was fixing this on `maint/v0.23`. Do you want it on master?

ethomson:
You're awesome.  Both would be amazing.  :sparkles:

johnhaley81:
Cool! You'll have them later today. :D

johnhaley81:
Thanks @ethomson!

PR #3359, opened 2015-08-05
Fix duplicate basenames to support older VS on maint/v0.23

johnhaley81:
With Visual Studio versions 2008 and older they ignore the full path to files and only check
the basename of the file to find a collision. Additionally, having duplicate basenames can break
other build tools like GYP.

This fixes https://github.com/libgit2/libgit2/issues/3356


ethomson:
Thanks for backporting this!

PR #3360, opened 2015-08-05
Fix duplicate basenames to support older VS on master

johnhaley81:
With Visual Studio versions 2008 and older they ignore the full path to files and only check
the basename of the file to find a collision. Additionally, having duplicate basenames can break
other build tools like GYP.

This fixes https://github.com/libgit2/libgit2/issues/3356

This is the `master` branch fix version of #3359. It needed to update `src/util.c` to point to the new file name.


johnhaley81:
This seems to be a dupe of #3357. Don't know which one you guys would rather take.

ethomson:
Thanks!

Issue #3074, opened 2015-04-25
Use a hash table to store index entries


carlosmn:
We currently have a vector as the data backing for the index' entries. This makes looking up an entry O(n) over the amount of entries, which can grow large. Using a hash table would allow us to do this in amortized O(1) time.

We would still need to have a sorted vector of entries for some operations, but we may be able to get away with generating that one on the fly rather than update it on each operation. There is a PR #1169 which did this for an older version of the structure, so we can't take it as-is.

We have a data structure `git_sortedcache` which is currently used by the refdb code but was made so we could use it for the index as well. We should rewrite that PR with the sorted cache as backing store.

Therzok:
:+1:

PR #3366, opened 2015-08-08
Use a hashmap for path-based lookups in the index

carlosmn:
The whole discussion[0] below is a bit of a mess, so the summary is that we can keep the hash table alongside the vector with a rather small penalty to parsing and filling times as long as we pre-allocate the size of the table (which we do in this series), so here it is.

We hash case-insensitively so we don't have to rehash when the index changes its case-sensitivity. There shouldn't be any extra collisions as long as we don't actually have paths which would collide.

During the switch from case-sensitive to case-insensitive, we do assume that we haven't now created two entries which would match the same (path, stage) key, but the vector sorting seems to assume the same thing.

[0] For lack for a better term, like "Carlos' stream of consciousness while he can't benchmark his way out of a paper bag"

---

~~~This is an attempt at tackling #3074 though the sortedcache, which was made with the intention of eventually making use of it is almost, but not quite, the structure we need. It assumes that the key is the path, which is not true for the index, as we also need to take the stage into account.~~~

I've gone off using sortedcache, as it's too specific to strings to use here, and instead I've added a new hashmap type specific to index entries.

khash doesn't help a lot here, since the hashing and comparison functions are set at compile time, so we need to have a conditional on an insert and lookup.

I suspect the real gains here would come from something like  a PATRICIA tree, which would provide us with ordering and lookup in a single data structure.

~~~Maybe we can do something silly like adding the stage as a part of the key which the sortedcache can see but it's not really part of the path... that would unfortunately mean adding something to a public struct for internal reasons, which is bad. The only other alternative would be to specialise the sortedcache in some way, but that's either a lot of duplication or macro work.~~~

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -17,12 +17,16 @@
 #include "pathspec.h"
 #include "ignore.h"
 #include "blob.h"
+#include "idxmap.h"
 
 #include "git2/odb.h"
 #include "git2/oid.h"
 #include "git2/blob.h"
 #include "git2/config.h"
 #include "git2/sys/index.h"
+
+GIT__USE_IDXMAP
+GIT__USE_IDXMAP_ICASE
 
 static int index_apply_to_wd_diff(git_index *index, int action, const git_strarray *paths,
 				  unsigned int flags,
@@ -425,6 +429,7 @@ int git_index_open(git_index **index_out, const char *index_path)
 	}
 
 	if (git_vector_init(&index->entries, 32, git_index_entry_cmp) < 0 ||
+		git_idxmap_alloc(&index->entries_map) < 0 ||
 		git_vector_init(&index->names, 8, conflict_name_cmp) < 0 ||
 		git_vector_init(&index->reuc, 8, reuc_cmp) < 0 ||
 		git_vector_init(&index->deleted, 8, git_index_entry_cmp) < 0)
@@ -462,6 +467,7 @@ static void index_free(git_index *index)
 	assert(!git_atomic_get(&index->readers));
 
 	git_index_clear(index);
+	git_idxmap_free(index->entries_map);
 	git_vector_free(&index->entries);
 	git_vector_free(&index->names);
 	git_vector_free(&index->reuc);
@@ -508,6 +514,11 @@ static int index_remove_entry(git_index *index, size_t pos)
 	if (entry != NULL)
 		git_tree_cache_invalidate_path(index->tree, entry->path);
 
+	if (index->ignore_case)
+		git_idxmap_icase_delete((khash_t(idxicase) *) index->entries_map, entry);
+	else
+		git_idxmap_delete(index->entries_map, entry);
+
 	error = git_vector_remove(&index->entries, pos);
 
 	if (!error) {
@@ -535,6 +546,7 @@ int git_index_clear(git_index *index)
 		return -1;
 	}
 
+	git_idxmap_clear(index->entries_map);
 	while (!error && index->entries.length > 0)
 		error = index_remove_entry(index, index->entries.length - 1);
 	index_free_deleted(index);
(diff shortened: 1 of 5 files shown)


carlosmn:
Now all we need is some benchmarks to see if this is worth doing (or what we might still have to adjust to get speedups). I'll probably grab the WebKit index file and do a few things like parse it, do a read-tree and look up a few paths to see how those compare.

carlosmn:
And with some basic benchmarks done, this adds a non-trivial amount of time both to parsing an index and read-tree. We end up spending ~29% of the time hashing the entries, and then we still have to put them in the sorted vector.

I do suspect that `git_index_add{,_bypath}` can be made significantly faster, as the checks for collisions would be sped up quite a bit, but if we continue down the route of hashmap+vector, we have to look carefully at which higher-level operations we want to speed up and whether the add and lookup operations are done often enough that we do want to trade that for the increase in each parse and read-tree.

carlosmn:
Here's some benchmarks, first from master

```
opening index took 0.187500s
read-tree took 0.408203s
looking up 100000 times took 0.062500s
adding 10000 times took 0.023438s
```

and with this branch

```
opening index took 0.348145s
read-tree took 0.580078s
looking up 100000 times took 0.029297s
adding 10000 times took 0.024414s
```

So as expected, looking up is much quicker, but actually loading the data is considerably slower. Adding can probably be made faster, as we currently iterate over a lot of the entries to check for collisions. But at the same time, a hash lookup isn't enough to determine whether we're a prefix of an existing entry.

carlosmn:
While this new read-tree is much slower than the old version, it's actually roughly the same as git's read-tree when we take the writing out to disk part, so it looks like that part is mostly fine. Parsing an index is not that much slower (in release mode) but it'd still want to see if that can be made faster.

carlosmn:
Some more ~~~lies~~~ benchmarks with lazy map. On master (the read-tree time has gone up since I added the writing-out portion in order to compare it to git):

```
opening index took 0.188477s
read-tree took 0.638184s
looking up 100000 times took 0.060059s
adding 10000 times took 0.023438s
```

And with this branch

```
opening index took 0.179688s
read-tree took 0.637207s
looking up 100000 times took 0.028809s
adding 10000 times took 0.021484s
```

So as expected, there's change to opening an index or performing a read-tree, and looking up is still cheaper. This time includes how long it took to build the hashmap. Note that the cost of this operation is significantly reduced as we pre-allocate the final size of the hashmap, so the additional cost isn't quite what it used to be.

carlosmn:
OK, nevermind lazy, we don't actually need it. The problem we have is resizing the hashtable, so we just need to insert only when we know the size, which is easy: just fill the vector first.

This adds ~50ms on opening/read-tree for the WebKit index. And that's a 24MB beast, so it's not too big a tax to pay. An index for such a large amount of entries does cost ~4.5MB, which is not awesome (mind you, the index without the map is already ~28MB).

carlosmn:
It's a bit worrying that this is passing as-is since the case-switching behaviour isn't even implemented. Which means we probably don't test it at all.

carlosmn:
Now that the history and diff is somewhat more readable, thoughts @ethomson ?

src/index.c
```diff
@@ -1139,6 +1159,13 @@ static int index_insert(
 		 * check for dups, this is actually cheaper in the long run.)
 		 */
 		error = git_vector_insert_sorted(&index->entries, entry, index_no_dups);
+
+		if (error == 0) {
+			if (index->ignore_case)
+				git_idxmap_icase_insert((khash_t(idxicase) *) index->entries_map, entry, entry, error);			else
+				git_idxmap_insert(index->entries_map, entry, entry, error);
```
ethomson: Bleh, this is hard to parse, can you correct the formatting?

carlosmn: But I was hoping to submit this to the underhanded C contest...

Yeah, I must have hit the wrong button at some point, but it's amazing that it's still perfectly valid C.

ethomson: Yeah after having read the rest of the PR I understand what this was intended to do now.  :P

I wonder if we might be able to have a macro or something to avoid all the `index->ignore_case switches`.  (But then we have two problems.)

carlosmn: A macro for this file would not be bad.

carlosmn: Done, should be more readable now.

barrbrain:
@carlosmn Thanks for reviving and resolving this issue! :+1: :100:

vmg:
Oooh, long time coming. :heart_eyes:

Issue #3365, opened 2015-08-06
git_submodule_path does not return correct path


csware:
Take the repository https://github.com/mpc-hc/mpc-hc and use `git_submodule_foreach` to iterate over the submodules.  And now notice that inside the callback function the function `git_submodule_path` for the submodule with the following definition returns `LAVFilters` instead of `src/thirdparty/LAVFilters/src`.

```
[submodule "LAVFilters"]
    path = src/thirdparty/LAVFilters/src
    url = https://github.com/mpc-hc/LAVFilters.git
```

Fun fact:
It works correctly for the following submodules:
```
[submodule "bs2b"]
    path = src/thirdparty/bs2b/libbs2b
    url = git://github.com/alexmarsev/libbs2b.git
[submodule "lcms2"]
    path = src/thirdparty/lcms2/library
    url = https://github.com/mm2/Little-CMS.git
[submodule "MediaInfoLib"]
    path = src/thirdparty/MediaInfo/library
    url = https://github.com/MediaArea/MediaInfoLib.git
[submodule "soxr"]
    path = src/thirdparty/soxr/libsoxr
    url = git://github.com/alexmarsev/soxr.git
[submodule "ZenLib"]
    path = src/thirdparty/ZenLib/library
    url = https://github.com/MediaArea/ZenLib.git
```

Other failing submodules are:
```
[submodule "RARFileSource"]
    path = src/thirdparty/RARFileSource/library
    url = https://github.com/mpc-hc/rarfilesource.git
[submodule "soundtouch"]
    path = src/thirdparty/SoundTouch/soundtouch
    url = git://github.com/alexmarsev/soundtouch.git
[submodule "sanear"]
    path = src/thirdparty/sanear/sanear
    url = git://github.com/alexmarsev/sanear.git
    branch = core
```

csware:
It seems as if the `map`in `submodule.c` isn't updated with the new key.

carlosmn:
Fixed via #3370

PR #3367, opened 2015-08-08
Correctly handle renames of submodules

csware:
Fixes issue #3365

### DIFF
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -1690,7 +1690,12 @@ static int submodule_load_from_config(
 		} else {
 			if (sm->name != sm->path)
 				replaced = sm->name;
+			git_strmap_delete(data->map, sm->name);
 			alternate = sm->name = git_buf_detach(&name);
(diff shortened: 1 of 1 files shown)


carlosmn:
Is there a test to show that this still lets us look up both by name and url? From this code, it looks like it would stop being available under the name.

csware:
@carlosmn You are right. But if a submodule appears twice in the map it's also free'd twice.

carlosmn:
I'm working on a refactor of the submodule lookup which should also solve this by splitting what each function does, but if you do have a test for this, it'd be great to have.

csware:
See issue #3365 for an example where the current implementation fails

swisspol:
See #3390 for another reproducible bug with submodules having their path different from their name.

swisspol:
This patch fixes #3390 as well.

carlosmn:
Superseded by #3370

PR #3368, opened 2015-08-09
config: Do not write repeated config section

linquize:
1. If the section already exist, new key should write to the end of that section
2. If there are multiple sections of same name, write to the last occurred section

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -76,6 +76,8 @@ typedef struct git_config_file_iter {
 		 (iter) && (((tmp) = CVAR_LIST_NEXT(iter) || 1));\
 		 (iter) = (tmp))
 
+#define READER_POS(r) ((r).read_ptr - git_buf_cstr(&(r).buffer))
+
 struct reader {
 	time_t file_mtime;
 	size_t file_size;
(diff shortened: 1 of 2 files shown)


carlosmn:
The error that fails when applying the test changes to master is the one about preserving whitespace, which is not what this PR is about. This should have its own test, as it's a different kind of test from the ones we have. The existing tests test for functionality.

carlosmn:
Why do you need a stack of positions in order to detect the latest section header? It should only be necessary to keep the last one.

There is a comment about a variable being a multivar which worries me. Whether a variable is a multivar or not depends *entirely* on the interpretation of that variable by the user of the configuration. The configuration parsing code cannot make this decision.

ethomson:
I feel like the less impactful change here would be to simply buffer comments and write them in `write_on_eof` or `write_on_section` after writing the variable...

carlosmn:
Since we operate on a line-by-line basis for writing variables (overwriting one or adding), if we're going to keep track of that stuff, IMO we should be keeping a list of lines, so we can refer to them later by index and we can keep track of two pieces of information, which line we want to affect, and whether we want to replace or append.

PR #3370, opened 2015-08-10
submodule: refactor to be more explicit in the search

carlosmn:
When searching for information about a submdoule, let's be more explicit
in what we expect to find. We currently insert a submodule into the map
and change certain parameters when the config callback gets called.

Switch to asking for the configuration we're interested in, rather than
taking it in an arbitrary order.

---

Not as many minuses as I'd hoped, but a bunch of the code is rather repetitive. This may handle #3365 as a side-effect, but there's currently no test for explicitly such a case.

### DIFF
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -89,9 +89,11 @@ __KHASH_IMPL(
 
 static int submodule_alloc(git_submodule **out, git_repository *repo, const char *name);
 static git_config_backend *open_gitmodules(git_repository *repo, int gitmod);
+static git_config *gitmodules_snapshot(git_repository *repo);
 static int get_url_base(git_buf *url, git_repository *repo);
 static int lookup_head_remote_key(git_buf *remote_key, git_repository *repo);
-static int submodule_load_from_config(const git_config_entry *, void *);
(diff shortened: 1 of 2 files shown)


carlosmn:
This does solve #3365 though I don't know how to reproduce the situation. There does not seem to be anything different about the submodules which don't have the correct path.

csware:
@carlosmn Yeah, issue #3365 is about the order in which the config sections are returned by `git_config_file_foreach`. If `submodule.X.url` is before `submodule.X.path` (or vice versa, don't remember right now) then the issue arises.

carlosmn:
All of those submodule configs you show on that issue have the path before the url, so that doesn't look like it.

I have tested it against the repository in question and it does fix it, but the original cause for the error is still unclear.

csware:
@carlosmn I'm really sure that it is the order, I debugged the old code step by step.

swisspol:
See #3390 for another reproducible bug with submodules having their path different from their name.

csware:
Any ETA for this?

carlosmn:
There's a bug reported against it with no unit test, so it'll take as long as it takes to reproduce, fix and review.

PR #3372, opened 2015-08-11
remote: don't confuse tag auto-follow rules with refspec matching

carlosmn:
We currently skip a ref if it's a tag and the auto-follow says we don't want it, but we still need to check whether the user's refspec says we do want it.


vmg:
Lovely. Thanks @carlosmn! :)))

PR #3375, opened 2015-08-13
Enable ccache for builds on Travis

arthurschreiber:


ethomson:
I don't really have any of the background here - but it looks like ccache is enabled on travis builds?  It's really this simple to enable it?  Are there any hidden downsides?

/cc @carlosmn

carlosmn:
The Mac build is the slowest, as it cannot be containerized the same way as the Ubuntu ones, so this was an attempt to get them faster by re-using object files, which should remain constant most of the time.

There shouldn't be downsides, assuming ccache can correctly detect which objects to re-use (presumably it does remember which compiler flags an object file was built with).

I'm not too fond of this method of using ccache though. It should be possible to set `CC` or pass `CMAKE_C_COMPILER` to `ccache` in the build in order to make CMake use that. It's built so that the build system doesn't need to know that `ccache` isn't a "normal" C compiler.

carlosmn:
Hold on, this is only active on the Ubuntu builds, I thought we were trying to get the OS X builds to go faster.

arthurschreiber:
The idea here was to improve the build speed on both linux as well as os x build machines.

Unfortunately, I couldn't quite get this to work yet, and didn't have the time to take a closer look.

ethomson:
It seems like this didn't work out quite as expected, should we just close this or do we want to spend more time trying to get it to work?

ethomson:
I like this idea, but there hasn't been much movement in the last year, so I'm going to close this.

PR #3377, opened 2015-08-14
Fix bug in git_smart__push: push_transfer_progress cb is only called at end

dleehr:
The conditional checked cbs->transfer_progress then used the value in cbs->push_transfer_progress. In both cases it should be push_transfer_progress


Therzok:
:+1:

ethomson:
Good catch!  Thanks!

PR #3378, opened 2015-08-14
Rename struct entry

phatblat:
I'm experimenting with building libgit2 as a [dynamic framework](https://github.com/phatblat/git2.framework) in Xcode instead of cmake.

One of the issues I've run into is that two structs called `entry` conflict with another `entry` struct in [Darwin.POSIX.search (search.h)](http://www.freebsd.org/cgi/man.cgi?query=hcreate&manpath=SuSE+Linux/i386+11.3) that is included in Apple's SDKs.

```c
typedef struct entry	{
	char *key;
	void *data;
} ENTRY;
```

I'm proposing changing the struct names here in libgit2 since getting Apple to change this in their SDK is next to impossible. This may be a terrible idea, with impacts I'm not aware of. But, it's enough to get Xcode to be happy.

### DIFF
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -24,7 +24,7 @@ extern git_mutex git__mwindow_mutex;
 
 #define UINT31_MAX (0x7FFFFFFF)
 
-struct entry {
+struct idx_entry {
 	git_oid oid;
 	uint32_t crc;
 	uint32_t offset;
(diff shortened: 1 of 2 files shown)


carlosmn:
Why can't you avoid the inclusion of `<search.h>` when building with Xcode? That header exists on OSX when we build with CMake and it exists on many other unices, but it builds fine because neither the indexer or the packbuilder need that header, so there is no redefinition.

phatblat:
Finally figured it out. The conflict was due to the different behavior when modules are enabled. Disabling `CLANG_ENABLE_MODULES` prevents this conflict.

PR #3379, opened 2015-08-14
report libssh2 error if list of authentication methods can't be retrieved

theseion:


carlosmn:
Let's hold off on this for a bit, as it looks like this might be due to an bug in libssh2 not sticking to the blocking-ness of the channel/connection.

theseion:
Opened an issue issue #3382 for this.

ethomson:
It's not clear to me why this is related to #3382...  This looks like a legit change, regardless of why libssh2 might be failing - is that not correct?  Now it looks like we have the ability to return -1 without having set any error, which would be disappointing.

theseion:
It's not directly related. But a bug in libssh2 might be reason to put in additional code at the same place where I made the change.

> Now it looks like we have the ability to return -1 without having set any error, which would be disappointing.

Yes, that was the case before, no error had been set.

carlosmn:
The lone -1 was there because it would mean that the server did not accept the 'none' auth request, but it did not tell us what it *would* accept.

But it looks like we do have a third alternative, which is that we never got to speak to the server because libssh2 never actually went into blocking mode and sent the request.

PR #3381, opened 2015-08-14
New feature: add the ablility to iterate through a directory in index

leoyanggit:
git_index_directory_iterator, git_index_directory_iterator_new,
git_index_directory_iterator_next and git_index_directory_iterator_free
are added.

I know we could achieve this by going through git_tree_entry recursively for stage 0.
But I think my patch here is more efficient compared to that approach. Further this patch
also support going through a directory in other stages.

### DIFF
--- a/include/git2/index.h
+++ b/include/git2/index.h
@@ -643,6 +643,17 @@ GIT_EXTERN(int) git_index_update_all(
  */
 GIT_EXTERN(int) git_index_find(size_t *at_pos, git_index *index, const char *path);
 
+/**
+ * Find the first position of any entries matching a prefix. To find the first position
+ * of a path inside a given folder, suffix the prefix with a '/'.
+ *
+ * @param at_pos the address to which the position of the index entry is written (optional)
+ * @param index an existing index object
+ * @param prefix the prefix to search for
+ * @return 0 with valid value in at_pos; an error code otherwise
+ */
+GIT_EXTERN(int) git_index_find_prefix(size_t *at_pos, git_index *index, const char *prefix);
+
 /**@}*/
 
 /** @name Conflict Index Entry Functions
--- a/src/index.c
+++ b/src/index.c
@@ -1420,6 +1420,30 @@ int git_index_remove_directory(git_index *index, const char *dir, int stage)
 	return error;
 }
 
+int git_index_find_prefix(size_t *at_pos, git_index *index, const char *prefix)
+{
+	int error = 0;
+	size_t pos;
+	const git_index_entry *entry;
+
+	if (git_mutex_lock(&index->lock) < 0) {
+		giterr_set(GITERR_OS, "Failed to lock index");
+		return -1;
+	}
+
+	index_find(&pos, index, prefix, strlen(prefix), GIT_INDEX_STAGE_ANY, false);
+	entry = git_vector_get(&index->entries, pos);
+	if (!entry || git__prefixcmp(entry->path, prefix) != 0)
+		error = GIT_ENOTFOUND;
+
+	if (!error && at_pos)
+		*at_pos = pos;
+
+	git_mutex_unlock(&index->lock);
+
+	return error;
+}
+
 int git_index__find_pos(
 	size_t *out, git_index *index, const char *path, size_t path_len, int stage)
 {
--- a/tests/index/tests.c
+++ b/tests/index/tests.c
@@ -155,6 +155,27 @@ void test_index_tests__find_in_empty(void)
    git_index_free(index);
 }
 
+void test_index_tests__find_prefix(void)
+{
+   git_index *index;
+   const git_index_entry *entry;
+   size_t pos;
+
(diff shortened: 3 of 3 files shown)


ethomson:
So...  this seems like an API with a very narrow focus.  A lot of questions come to mind as to who else this API would be appropriate for.  For example:  why always the same stage?  When would anybody ever want to iterate all stage 1 index entries in a directory?  It seems like we should do something like one of:

1. Always iterate stage 0
2. Always iterate all stages
3. Offer the user to iterate either stages *or* conflicts.

Even with that change, though, I still think is not likely to have broad appeal, and it's something that anybody could already do with the current API, though I admit, not as efficiently.

I wonder if we would benefit *more* by simply having a higher-level function that would provide the first index entry that matches a prefix, so that you can let libgit2 do the binary search into the directory for you, then you can do the iteration on your own, with the existing index iteration functions.

leoyanggit:
Regarding which stage/stages we should support, how about we support a specific stage and `GIT_INDEX_STAGE_ANY` (to support `GIT_INDEX_STAGE_ANY` only a small change needed based on my patch)? This way it is complete and flexible to the user.

> I wonder if we would benefit more by simply having a higher-level function that would provide the first index entry that matches a prefix, so that you can let libgit2 do the binary search into the directory for you, then you can do the iteration on your own, with the existing index iteration functions.

This could be an approach. It implies that we promise that the index is sorted (which is an implementation detail) after the first index entry is provided for a matched prefix. It's true for now. But I'm not sure if it's possible that the fact will change in the future.

carlosmn:
We don't have an index iterator, so introducing one for directories seems strange. Especially because the index does not have directories (though it does know about them).

The index being sorted is not an implementation detail. It's part of the contract. It's a sorted dictionary with the key being (path, stage). That's what makes it a `git_index`. *How* we keep it sorted is an implementation detail.

For now, a much more generally useful convenience to add would be to find the first/last path with a particular prefix. I think that we already have something similar internally which can be wrapped. This removes any decision about staging for the consumer, and they can iterate as they wish and a they have been doing up to now.

leoyanggit:
> We don't have an index iterator, so introducing one for directories seems strange. Especially because the index does not have directories (though it does know about them).

Yes index doesn't have directories but it's an view of directories. We already have `git_index_remove_directory`. Does it apply in your argument?

> The index being sorted is not an implementation detail. It's part of the contract. It's a sorted dictionary with the key being (path, stage). That's what makes it a git_index. How we keep it sorted is an implementation detail.

Good to know through this conversation. 

> For now, a much more generally useful convenience to add would be to find the first/last path with a particular prefix. I think that we already have something similar internally which can be wrapped. This removes any decision about staging for the consumer, and they can iterate as they wish and a they have been doing up to now.

This should work. But I don't see any significant difference between this approach and mine while this approach has less number of APIs but not as convenient as an iteration interface from the user perspective. I can't see how the approach removes any decision about staging for the user. After the user gets the first/last entry matching the prefix the user goes through the `git_index_entry` and then needs to make the decision for every entry. So I think the advantage of this approach is less code change.

leoyanggit:
I just updated my patch with suggested approach.

leoyanggit:
@ethomson @carlosmn What do you think about the new patch?

ethomson:
Sorry about the delay, I was on vacation.

I would like to see this still more general - both because this concept is nice and neat and could be used elsewhere, and because the index is a list of paths and doesn't really have a concept of a directory.  Instead of presenting this new concept (directories in the index), let's keep it more generic and have:

```
/**
 * Find the first position of any entry matching a given prefix.  To find the first path inside a
 * given folder, suffix the prefix with a `/`.
 *
 * @param at_pos the address to which the position of the index entry is written (optional)
 * @param index an existing index object
 * @param prefix the prefix to search for
 * @return 0 with valid value in at_pos; an error code otherwise
 */
GIT_EXTERN(int) git_index_find_prefix(size_t *at_pos, git_index *index, const char *prefix);
```

leoyanggit:
@ethomson That works for me. I updated the patch.

carlosmn:
Do we have to lock the index? This call shouldn't be done in parallel with anything else.

leoyanggit:
To match `git_index_find()` we probably should?

carlosmn:
Could we get a test that uses this function to make sure we don't regress it when we're dealing with the index internal structures?

leoyanggit:
@carlosmn Are there any instructions/wiki pages/examples about writing a test case in libgit2?

carlosmn:
Copy the style of the other tests which test index access. If we don't ship the test suite's README, you can see it at its project page https://github.com/vmg/clar

leoyanggit:
@carlosmn thanks. I updated my patch with a test.

ethomson:
This is a very handy addition.  Thanks, @leoyanggit !

leoyanggit:
Thanks @ethomson and @carlosmn !

PR #3383, opened 2015-08-15
http: propagate the credentials callback's error code

carlosmn:
When we ask for credentials, the user may choose to return EUSER to
indicate that an error has happened on its end and it wants to be given
back control.

We must therefore pass that back to the user instead of mentioning that
it was on_headers_complete() that returned an error code. Since we can,
we return the exact error code from the user (other than PASSTHROUGH)
since it doesn't cost anything, though using other error codes aren't
recommended.

---

We do actually have  a test for this, but it's currently only run for SSH.

### DIFF
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -36,6 +36,8 @@ static const char *post_verb = "POST";
 
 #define PARSE_ERROR_GENERIC	-1
 #define PARSE_ERROR_REPLAY	-2
+/** Look at the user field */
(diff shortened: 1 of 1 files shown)


carlosmn:
@alexcrichton this should solve your problem with the return code. WinHTTP may or may not behave this way, though.

alexcrichton:
Awesome, thanks @carlosmn!

PR #3389, opened 2015-08-19
CMake: fall back to OpenSSL on older OS X

carlosmn:
Starting at OS X 10.8, the Security framework offers some functions
which are unified across OS X and iOS. These are the functions that we
use.

Older versions of OS X do not have these functions and we fail to
compile. In these situations, fall back to using OpenSSL for our TLS
stream instead.

---

@alexcrichton can you give this a try? I tested it locally by reversing the logic, but I'd like to get an actual confirmation.


alexcrichton:
Works for me, thanks @carlosmn!

PR #3394, opened 2015-08-24
Fix 'If we're dealing with a directory' check

vsevolodp:
This change fixes issue [libgit2#3393]

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -97,7 +97,7 @@ static int does_negate_rule(int *out, git_vector *rules, git_attr_fnmatch *match
 		if (rule->containing_dir) {
 			git_buf_puts(&buf, rule->containing_dir);
 		}
-		if (!strchr(rule->pattern, '*'))
+		if (rule->flags & GIT_ATTR_FNMATCH_LEADINGDIR && !(rule->flags & GIT_ATTR_FNMATCH_NEGATIVE))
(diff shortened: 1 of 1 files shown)


carlosmn:
This looks like a correct fix it, but it's not obvious why the check is done this way. This also makes the comment above meaningless since there is no `strchr()` check anymore. Could you update it to say why we check for `LEADINGDIR` and `NEGATIVE`?

It'd also be great to have a unit test, as it's likely we'll regress otherwise. I had this in `tests/status/ignore.c` while testing locally, though I don't know if it quite matches the situation you are facing.

```C
void test_status_ignore__negate_starstar(void)
{
	int ignored;

	g_repo = cl_git_sandbox_init("empty_standard_repo");

	cl_git_mkfile("empty_standard_repo/.gitignore",
		      "code/projects/**/packages/*\n"
		      "!code/projects/**/packages/repositories.config");

	cl_git_pass(git_futils_mkdir_r("code/projects/foo/bar/packages", "empty_standard_repo", 0777));
	cl_git_mkfile("empty_standard_repo/code/projects/foo/bar/packages/repositories.config", "");

	cl_git_pass(git_ignore_path_is_ignored(&ignored, g_repo, "code/projects/foo/bar/packages/repositories.config"));
	cl_assert_equal_i(0, ignored);
}
```

PR #3395, opened 2015-08-24
Don't throw errors in maybe_append_head() if there is no head.

It is vacuously true that if there is no head reference then it does NOT point to anything that has been updated and thus maybe_append_head() has no work to do and shouldn’t throw an error (since it’s called “maybe” not “you_must_append_head_or_die

wilshipley:
This is my first pull request on this project, sorry if I’m doing this wrong.

The documentation for maybe_append_head() says it’ll append to the head if the head is pointing at the thing just modified.

In the case where there is NO head this function was returning an error, which got bubbled up to Objective-Git and was causing what appeared to be failures when there weren’t.

It is vacuously true that if there is no head reference then it does NOT point to anything that has been updated and thus maybe_append_head() has no work to do and shouldn’t throw an error (since it’s called “maybe” not “you_must_append_head_or_die”).


carlosmn:
If `HEAD` does not exist the repository is corrupted and we do want to return an error.

wilshipley:
Ok, I guess I don't understand how to move HEAD? I thought I had to delete it and recreate it to make a detached head. Is that not true? Because just deleting and recreating is impossible right now—in the recreation of HEAD it calls maybe...() and throws the error.

wilshipley:
This is the Swift code I was using to move HEAD. Is there some simpler way I'm missing?

            do {
                let headReference = try repository.lookUpReferenceWithName(PlanModel.headReferenceName)
                try headReference.delete()
            } catch {
            }
            try repository.createReferenceNamed("HEAD", fromOID: commit.OID!, message: "Current state of document")

The call to "repository.createReferenceNamed()" throws an error because of HEAD doesn't exist, even though it's HEAD I'm trying to create.

I know I could use "repository.headReference()" but it was doing something really odd when I first created the repository, where it'd return nil but there really was a HEAD file, so setting the HEAD threw an error. This may all be because I'm using Objective-Git wrong, I'm still figuring this stuff out.

carlosmn:
Never delete `HEAD`. If you delete `HEAD` you no longer have a Git repository. Even it it were a normal reference, you don't delete in order to update.

If you want to detach it, you set it to a commit. If you want it to point to a particular branch, you set it to that reference. We have functions to do this which also deal with the machine-readable reflog message, `git_repository_set_head()` and `git_repository_set_head_detached()`.

wilshipley:
Ok, thank you! I'll check it out. I didn't see these calls in the Objective-Git layer, but I can always call down into git2 itself.

PR #2888, opened 2015-02-06
Added filelist iterators to improve merge/diff performance.

jeffhostetler:
Here is code implementing the filelist iterators that were discussed in:
https://github.com/libgit2/libgit2/issues/2835

This commit contains the "index-filelist" and "workdir-filelist" iterators and unit tests for each.

It also contains changes to "merge" and "diff" to use them when appropriate.

I'll submit a follow-up PR with some large-merge test cases.

PR #3402, opened 2015-08-30
Provide path matching in the iterators (for faster diffs)

ethomson:
I rebased #2888 onto master and was faced with mega conflicts.  In resolving these I:

1. Wanted the filelists to be part of the general iterator mechanism, instead of having a workdir iterator and a workdir+filelist iterator
2. Did not want a third diff path maching mechanism (pathspecs, non-pathspecs and slightly more literal iterator filelists).

So this moves the filelists into the general iterator mechanisms, and it expands the filelist handling to deal with directory prefixes (eg, `foo/`) instead of requiring that patches be listed explicitly.

Note that this should be generally adequate and a huge perf benefit for index and workdir iterators (like the original was), and tree iterators are working, but they expand all trees unnecessarily.  We could perform tree iteration without necessarily expanding all trees (like we do with the filesystem iterator) but this should be correct and still more performant in that case.  We can handle that in the future, though with more aggressive refactoring in that iterator.

### DIFF
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -130,7 +130,9 @@ typedef enum {
 	GIT_DIFF_INCLUDE_CASECHANGE = (1u << 11),
 
 	/** If the pathspec is set in the diff options, this flags means to
-	 *  apply it as an exact match instead of as an fnmatch pattern.
+	 *  use exact prefix matches instead of an fnmatch pattern.  Each
+	 *  path in the list must either be a full filename or a subdirectory
+	 *  prefix.
 	 */
 	GIT_DIFF_DISABLE_PATHSPEC_MATCH = (1u << 12),
 
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -163,6 +163,13 @@ typedef enum {
 	/** Include unreadable files in the diff */
 	GIT_DIFF_INCLUDE_UNREADABLE_AS_UNTRACKED = (1u << 17),
 
+	/** Use literal path matching in the iterators.  This is more broad
+	*  than the DISABLE_PATHSPEC_MATCH flag.  The caller must provide an
+	*  array of paths (no patterns or prefixes).  Only values included in
+	*  that list will be returned.
+	*/
+	GIT_DIFF_ENABLE_FILELIST_MATCH = (1u << 18),
+
 	/*
 	 * Options controlling how output will be generated
 	 */
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -165,13 +165,6 @@ typedef enum {
 	/** Include unreadable files in the diff */
 	GIT_DIFF_INCLUDE_UNREADABLE_AS_UNTRACKED = (1u << 17),
 
-	/** Use literal path matching in the iterators.  This is more broad
-	*  than the DISABLE_PATHSPEC_MATCH flag.  The caller must provide an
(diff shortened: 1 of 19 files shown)


ethomson:
Ideally this obviates both #2888 (while making it a bit less confusing for the users to have to understand the difference between non-pathspec matching and filelist matching) and #3312 (while providing robustness).

ethomson:
Looks like there are still some issues with case insensitivity (amongst others) that I need to investigate, but this should be a decent start, I hope.

include/git2/diff.h
```diff
@@ -130,7 +130,9 @@ typedef enum {
 	GIT_DIFF_INCLUDE_CASECHANGE = (1u << 11),
 
 	/** If the pathspec is set in the diff options, this flags means to
-	 *  apply it as an exact match instead of as an fnmatch pattern.
```
Therzok: This is a rather breaking change. Can we add another switch which changes it to prefix rather together with the disabling version?

ethomson: Huh?  What's breaking here?  Can you give an example of something that matches differently?

The change you're commenting on is to better explain the *current* behavior, not document a change.

The goal in this PR is to give a speed up *without* a change in behavior.

jeffhostetler:
Wow!!   What a nice refactoring.   Sorry to put you through all of that.

ethomson:
Okay, I cleaned this up a bit and fixed the case sensitivity wrongness in the tests and this should be ready for review.

I apologize if I was unclear about the intent of this change.  The goal here is to have *no* behavioral changes in `DISABLE_PATHSPEC_MATCH`, and to give iterators support for the *existing* behavior, which *does* handle paths and directory prefixes.  Which is to say that a list of paths of:

    bar
    foo/

Will match:  a file named `bar` *or* a folder name `bar` and a folder named `foo`.  This match is recursive.  There is not meant to be a change here.  The point of this PR is to bring the previous filelist matching behavior from #2888 (which was filename only, and did not support directories in the list) into the behavior of the `DISABLE_PATHSPEC` matching.

(The change to the documentation is accurate even in the current version, it represents an expansion of the documentation, *not* a change.)

include/git2/diff.h
Therzok: Say:

file1.ext
file1.ext.generated

What will the diff on file1.ext do? Before it would've matched `file1.ext`.

ethomson: There's no change in that.  You get a filename or a directory prefix.

Therzok: Then I read the whole thing wrong. Sorry!

ethomson: Nah, if there's confusion then I failed at writing the docs.  How about:

>  If the pathspec is set in the diff options, this flag indicates that the list will be treated as literal paths instead of fnmatch patterns.  Each path in the list must either be a full path to a file or directory (a trailing slash indicates that the path will *only* match a directory).  If a directory is specified, all children will be included.

Therzok: That's way better.

carlosmn:
Did you want to riff some more on the documentation wording? Otherwise I think we should try to get this merged soon, lest we start messing about with the code and have to rewrite it again.

ethomson:
I updated based on @Therzok 's helpful input.  I think this is ready to :shipit: , with apologies to @jeffhostetler for not having done so earlier.

Therzok:
:+1: I am happy with this.

PR #3406, opened 2015-09-01
(DO NOT MERGE AS IS) Working implementation of Mysql / Mariadb backends + transport

jflesch:
Hello,

I made a working implementation of Mysql / Mariadb backends (ODB + RefDB) + Mysql/Mariadb transport. I assume you could be interested in knowing it exists and that it works quite nicely :). 

As of now, it is a little bit slow (I would say, about 3 to 5 times slower than raw files). I'm still trying to fine-tune my MariaDB configuration. However, I guess it's already good-enough to enjoy Mariadb features like [Galera replication](https://mariadb.com/kb/en/mariadb/what-is-mariadb-galera-cluster/) with Git repositories.

Unfortunately, I'm short on time, so I did all my tests and testing directly from Pygit2 ( https://github.com/jflesch/pygit2/tree/mariadb ). There are no tests in my Libgit2 branch :/ . Also my Git commit history is quite dirty. Initially I did everything in Pygit2 directly until I realized it was a bad idea : I had to make it in Libgit2 directly to have a correct Mysql/Mariadb transport implementation.

So I guess it can't be merged as is, sorry :/


ethomson:
Welcome!

This is very exciting, we're always happy to see new backends.  Inside the libgit2 project itself, however, we want to limit ourselves to *only* the backends that are compatible with git itself (so at the moment, the loose and packed ODB backends and the loose and packed refdb backends).

I'm sure you can understand that adding something like a (for example) sqlite backend into the core code base has the potential to increase the support load on the core library quite a bit.  (For example: what if it doesn't work on the newest version of sqlite?  I don't think any of the core contributors have the expertise to solve that problem.)

However, I think that this could be very useful in the [`libgit2-backends`](https://github.com/libgit2/libgit2-backends) repository.  This is where we keep the non-core backends that aren't for strict git compatibility but are still important and visible.

Would you mind opening a PR with these changes there so that we can review it?

Also, we're always curious to know what people are doing with libgit2 - are you at a stage where you're willing to share what you're building?  (If not, we understand!)

Thanks again!

jflesch:
Sure, I can do a PR for libgit2-backends. I'll do it a little bit later.

However, I just have one question : I did my changes in Libgit2 so I could reuse the code of the transport 'local' and add support for URIs starting with "mariadb://" easily. Is there any plan to make that possible from outside code someday ?

Right now, I would prefer not share the whole project I'm working on. I'll share it, but now is far too soon.

ethomson:
> Is there any plan to make that possible from outside code someday ?

You *should* be able to register a custom URL handler using [`git_transport_register`](https://github.com/libgit2/libgit2/blob/master/include/git2/sys/transport.h#L156).  We do this for HTTP and HTTPS URLs within Visual Studio, for example, so that we can build a single connection to our ALM servers and use that for not just Git operations, but Build, Work Item Tracking, etc, etc.

This would be the architecture that I think we would recommend.  I've only had a quick glance at the transports:  are you reusing part of the local transport code?  If so, we may be able to expose a bit of it to advanced / `sys`-level callers.

PR #3408, opened 2015-09-02
Minor changes to make the creation of new backends and transports easier

jflesch:
Hello,

As suggested on https://github.com/libgit2/libgit2/pull/3406 , I will try to make a standalone version of my backend for MariaDB. However, to make this standalone version as clean as possible, I think I would need those minor changes to be included in Libgit2.

Thanks in advance,

### DIFF
--- a/include/git2/fnmatch.h
+++ b/include/git2/fnmatch.h
@@ -30,20 +30,26 @@
 
 #include "common.h"
 
-#define FNM_NOMATCH		1		/* Match failed. */
-#define FNM_NOSYS		2		/* Function not supported (unused). */
-#define	FNM_NORES		3		/* Out of resources */
+#define GIT_FNM_NOMATCH		1		/* Match failed. */
+#define GIT_FNM_NOSYS		2		/* Function not supported (unused). */
+#define	GIT_FNM_NORES		3		/* Out of resources */
 
-#define FNM_NOESCAPE	0x01		/* Disable backslash escaping. */
-#define FNM_PATHNAME	0x02		/* Slash must be matched by slash. */
-#define FNM_PERIOD		0x04		/* Period must be matched by period. */
-#define FNM_LEADING_DIR 0x08		/* Ignore /<tail> after Imatch. */
-#define FNM_CASEFOLD	0x10		/* Case insensitive search. */
+#define GIT_FNM_NOESCAPE	0x01		/* Disable backslash escaping. */
+#define GIT_FNM_PATHNAME	0x02		/* Slash must be matched by slash. */
+#define GIT_FNM_PERIOD		0x04		/* Period must be matched by period. */
+#define GIT_FNM_LEADING_DIR 0x08		/* Ignore /<tail> after Imatch. */
+#define GIT_FNM_CASEFOLD	0x10		/* Case insensitive search. */
 
-#define FNM_IGNORECASE	FNM_CASEFOLD
-#define FNM_FILE_NAME	FNM_PATHNAME
+#define GIT_FNM_IGNORECASE	GIT_FNM_CASEFOLD
+#define GIT_FNM_FILE_NAME	GIT_FNM_PATHNAME
 
-extern int p_fnmatch(const char *pattern, const char *string, int flags);
+/**
+ * Behave in a way similar to fnmatch() on Linux, but is portable
+ * (see man fnmatch).
+ * For instance, can be useful for people wanting to implement their
+ * own RefDB backend
+ */
+GIT_EXTERN(int) git_fnmatch(const char *pattern, const char *string, int flags);
 
 #endif /* _FNMATCH_H */
 
--- a/include/git2/sys/transport.h
+++ b/include/git2/sys/transport.h
@@ -185,18 +185,41 @@ GIT_EXTERN(int) git_transport_dummy(
 	git_remote *owner,
 	/* NULL */ void *payload);
 
+
+typedef struct git_transport_local_repo_callbacks {
+	/*!
+	 * Callback for the local transport to open a repository
+	 *
+	 * \param[out] out will point to the created repository
+	 * \param[out] user structure to keep associated with the repository
+	 * \param[in] url URL indicating the repository to open
+	 * \return 0 if successful, <0 if error (see git2/errors.h)
+	 */
+	int (*open)(git_repository **out_repo, void **user, const char *url);
+
+	/*!
+	 * Callback for the local transport to close a repository
+	 * opened with .open_repository_cb()
+	 *
+	 * \param repo repository that must be closed (don't forget git_repository_free())
+	 * \param user see the parameter 'user' for .open()
+	 */
+	void (*close)(git_repository *repo, void *user);
+} git_transport_local_repo_callbacks;
+
 /**
  * Create an instance of the local transport.
  *
  * @param out The newly created transport (out)
  * @param owner The git_remote which will own this transport
- * @param payload You must pass NULL for this parameter.
+ * @param payload You can pass NULL, or a pointer to a
+ *   git_transport_local_repo_callbacks structure
  * @return 0 or an error code
  */
 GIT_EXTERN(int) git_transport_local(
 	git_transport **out,
 	git_remote *owner,
-	/* NULL */ void *payload);
+	/* NULL or git_transport_local_repo_callbacks */ void *payload);
 
 /**
  * Create an instance of the smart transport.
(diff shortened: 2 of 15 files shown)


ethomson:
Hmm.  I don't think I understand yet what you're trying to achieve here yet.

When would you override local repository open in this manner?  Are you really trying to map some local repository path (eg `C:\foo`) into your database transparently?  Or are you trying to achieve some other effect?

If you're really looking for some transparent mapping, why can't you handle this in your application?

jflesch:
I'm not sure what you mean by 'mapping'. Here I'm trying to make classic Git repositories able to communicate with Mariadb/Mysql repositories (push, pull, clone, etc).

Here is my reasoning :

Context: we assume here we already have a working ODB + RefDB backend (working on top of Mysql or whatever). So I can instantiate a git_repository using those backends. It just require some extra code (see https://github.com/libgit2/libgit2/pull/3406/files#diff-72a962a4ea8b5f539654a4da46506f88R401 ).

Now if I want to push, pull, clone, etc from this repository, I need access it as a remote one (unless I missed another way of doing it ?). To do so, I need to implement a custom transport.

It seemed to me that the shortest and most reliable way to do that is to reuse the local transport, since it's relying only on a git_repository instance for all its operation.

Thing is, as of now, local transport is always instantiating the git_repository using git_repository_open(), which will always instantiate a classic git_repository (relying on a filesystem path, etc). So I just need a way to tell him how to get a custom git_repository. It seemed to me that the easiest and most flexible way to do that was to pass extra callbacks to the local transport code.

jflesch:
I think I should clarify my end goal here:

I would like to be able to clone a classic repository from a MariaDB one, make some changes in it using the classic git command, and push the changes back into the MariaDB.

In the end, this is useful for me for 2 things:
* Testing: I can basically just push the whole Linux repository to MariaDB and see what happens :-)
* Debuging / cleaning up mess : If a repository content goes wrong for some reason, I can clone it, try to fix the mess, and push it back in.

ethomson:
I think that I see what you're getting at...  but I don't think that the local repository driver is going to be quite right for this.

The local transport is sort of naively opening up another repository instance and writing into it (which happens to work for your case, too).  But I think that we want the local repository driver to really work on *local* (on-disk) repositories, because we might want to tune this in the future, which has thus not been a priority for us.  And so we might do something like dumping a packfile into the objects directory and just moving on.  Or we might want to try to support some non-bare situations like `receive.denyCurrentBranch`.

The latter may affect you less, since you're pretty much guaranteed to be bare.  But I think that you see what I'm getting at.

It feels like you're sort of trying to get around the fact that there's no server implementation built around libgit2.  Out of curiosity, would you be using this mechanism if instead you could just have a server backing your custom refdb/odb backends?  (I'm just curious here.  Regrettably, this is a bunch of work that we *want* to do but haven't yet done, and so we should keep going down this mechanism for now.)

Anyway.  I guess what I'm getting at is that I don't think we really want to modify the *local* transport here.  Instead, we want a more generic base transport that is extensible (and the local transport extends).  I'm thinking of something along the lines of the smart subtransport definition (which implements the smart protocol for HTTP, SSH and Git transports).

I don't think that this should be a super difficult change from where you're going with this.

Maybe some `git_transport_repository` ?  (I don't love this name, just throwing this out there.)  And then the payload for this is a `git_transport_repository_definition` that includes some callbacks?

Then `git_transport_local` could in fact become a subtransport based on this repository transport (and you could as well, similar to how the smart subtransport mechanisms are exposed.)

The advantage here is that this promotes code reuse today but we can do what we want in `git_transport_local` without breaking you in the future.

Having said that, this also seems like a place where we have to be pretty vigilant about not breaking you if we move away from that, so we'll probably want to make some nice test cases here as well.

Now, this is only my opinion.  I'm curious what @carlosmn thinks.

jflesch:
You're right. I didn't anticipate correctly the possible evolutions of the local transport. I'll have a look at how things are done with the smart subtransport, and see what I can/could do.

> Out of curiosity, would you be using this mechanism if instead you could just have a server backing your custom refdb/odb backends?

Actually no, I wouldn't. I had a look at what was required to have a working SSH server with my backends, and it just seemed to be too much work for now. This is why I fell back to this solution.
But I think it is something I'll have to do someday anyway (this is on my todo list for much much later).

carlosmn:
This exposes things which are completely internal to libgit2, some of them accidental, so we wouldn't want to expose any of this. I certainly wouldn't consider these changes minimal, as it makes our external API larger.

As for the local transport using a custom backend... it's implemented via libgit2 currently as that was what the first version used. A more correct implementation of the protocol would arguably call to `git-{upload,receive}-pack` instead of using the library and it was certainly never supposed to do anything more than replicate what git itself would be doing.

Using a custom backend has been meant for use-cases where you control the creation of the repositories rather than for every repository created by libgit2 internally to have them.

This is however where the repository format 1 might help. Its purpose is to allow for repositories to express what extensions they need. git itself will start to support setting the database-backed references via this mechanism, and it would be possible for libgit2 to offer automatic registration of backends based on this configuration.

This is something which we haven't discussed much, so it's still up in the air, but that would be the way how I would prefer for a custom backend to be available for the local transport.

jflesch:
> This is however where the repository format 1 might help

I'm confused. I've never heard anything about this format. Is this a planned evolution for Git and Libgit2 ?

Anyway, if I understand correctly, for now, this idea of reusing the local transport with my backend is a no-go ? (for now, this is fine for me ; I can keep working with my ugly/patched version of libgit2)

Still, would my first commit (the one about exporting p_fnmatch()) makes sense to merge ? I think it would be handy for people developing custom backends. If I'm not mistaken, reusing p_fnmatch() in the refdb backends is the safest way for them to make sure their function git_refdb_backend->iterator() will match the references as expected by Git/Libgit2 on all the platforms, no ?

carlosmn:
The repository format is not about any specific implementation but about Git as a system we all work with. It exposes compulsory extensions which a tool/implementation must support or abide by in order to open the repository.

This is what git is going to use in order to allow for referenced backed by an actual database rather than the filesystem (and stop versions which do not support this from trying to do anything in the repo). And as this is the method which we'll have to use for loading such a backend once it exists, I think it's where we should allow for backends such as this to be created.

The fnmatch is still internal to libgit2. If you need something similar (which I don't think you should for an iterator), your backend would have to use a different implementation.

jflesch:
> If you need something similar (which I don't think you should for an iterator)

The function git_refdb_backend->iterator() takes a parameter "glob". Without a function like fnmatch() / p_fnmatch(), I don't know how the backend can find the references matching the parameter "glob" and those who don't ?

carlosmn:
I must have misremembered. I think one of the frontends supports filtering if the backend doesn't, but it doesn't seem to be this one.

There is the system's `fnmatch(3)` which you can use, or you can copy code from some implementation if you want to have it in your codebase. `p_fnmatch()` is internal to libgit2 and it's not something that we should be exposing for use by anything else.

jflesch:
Ok. For my use, I can use the system's fnmatch().

jflesch:
BTW, thank you both for your feedback and advices :)

ethomson:
I agree that I don't want to expose `fnmatch` - however we should also be a little kinder to refdb backend implementers in this case.  We expect them to handle a glob to the ref iterator and a) don't really define what we mean by that, or even link to [some documentation](http://linux.die.net/man/7/glob) and b) hope that they have an `fnmatch` implementation of their own.

We don't want to put fnmatching at a level above - at least not for everybody - because we want iterators to be able to handle recursion avoidance (ie, avoid walking into directories on slow ass filesystems).  But maybe we should let the globbing iterator be an optional function and let backends provide an iterate-everything function that we would then apply fnmatch to.

I think this has some merit, but maybe there's a broader solution, because I suspect that somebody implementing a refdb backend against SQL would still want the matching as close to the data as possible and really would want to just return the results of a `LIKE`-filtered query.  I don't think there's a good translation between `fnmatch` and anything SQL based...  alas.

carlosmn:
> But maybe we should let the globbing iterator be an optional function and let backends provide an iterate-everything function that we would then apply fnmatch to.

Indeed. I'm pretty sure this is implemented *somewhere*, maybe in the config. It'd be nice if we had something like that for the refs as well. If you're just going to iterate over everything and apply fnmatch to the result, we might as well do it for you, but if you know how to convert that pattern into something better, you would still have the opportunity.

jflesch:
Just so you know, I've just realized there is a problem with custom transports as of now:

To use ```git_transport_register()```, we must provide a callback ```git_transport_cb``` returning a structure ```git_transport```. This structure ```git_transport``` must include a bunch of callbacks, including this one:

```
int (*push)(git_transport *transport, git_push *push, const git_remote_callbacks *callbacks);
```

Problem is, the definition of ```git_push``` is in ```src/push.h``` --> it is unavailable to external libraries / programs.

PR #3409, opened 2015-09-03
Maintenance backports for v23

carlosmn:
Nothing ground-breaking, but a few nice fixes have accumulated over the last month. I did not include the one which changes what we return for a bad pkt length as that's a string change, but it might be convenient enough for finding out what the server sent that I could be convinced to include it.


ethomson:
Awesome!  Thanks for your diligence with maintenance releases.

PR #3410, opened 2015-09-03
Revert "Get rid of libssh2 embedding"

carlosmn:
The embedding was removed as a libssh2 release with Windows crypto
support became available, but dependencies are still annoying so this
ahs been requested again.

This reverts commit 20dcb7315cd4c5760c68402998fd9e5a6bf5505d.

---

We probably want to go over this again, as we might otherwise end up linking to libssh2 in two different ways.


Therzok:
:heart:

ethomson:
Welcome back, libssh2.

PR #3411, opened 2015-09-04
Include custom HTTP headers

spraints:
Libgit2 sets some headers for HTTP requests (User-Agent, Host, Accept, Content-Type, Transfer-Encoding, Content-Length, Accept). In our app, we'd like to set some custom headers (e.g. X-Forwarded-For, X-Request-Id, etc). This branch is a spike to add a pair of get/set accessors for extra HTTP headers for a `git_remote`. Is this general concept OK? Is there a better way to do this?

### DIFF
--- a/src/transports/smart.c
+++ b/src/transports/smart.c
@@ -66,6 +66,17 @@ static int git_smart__set_callbacks(
 	return 0;
 }
 
+static int git_smart__set_custom_headers(
+	git_transport *transport,
+	const git_strarray *custom_headers)
+{
+	transport_smart *t = (transport_smart *)transport;
+
+	t->custom_headers = custom_headers;
+
+	return 0;
+}
+
 int git_smart__update_heads(transport_smart *t, git_vector *symrefs)
 {
 	size_t i;
@@ -399,6 +410,7 @@ int git_transport_smart(git_transport **out, git_remote *owner, void *param)
 
 	t->parent.version = GIT_TRANSPORT_VERSION;
 	t->parent.set_callbacks = git_smart__set_callbacks;
+	t->parent.set_custom_headers = git_smart__set_custom_headers;
 	t->parent.connect = git_smart__connect;
 	t->parent.close = git_smart__close;
 	t->parent.free = git_smart__free;
--- a/src/transports/smart.c
+++ b/src/transports/smart.c
@@ -66,57 +66,55 @@ static int git_smart__set_callbacks(
 	return 0;
 }
 
-static char *forbidden_custom_headers[] = {
-  "User-Agent",
-  "Host",
-  "Accept",
-  "Content-Type",
-  "Transfer-Encoding",
-  "Content-Length",
-};
+int http_header_name_length(const char *http_header)
+{
+	const char *colon = strchr(http_header, ':');
+	if (!colon)
+		return 0;
+	return colon - http_header;
+}
 
-bool is_valid_custom_header(const char *custom_header)
+bool is_malformed_http_header(const char *http_header)
 {
 	const char *c;
 	int name_len;
-	unsigned long i;
 
 	// Disallow \r and \n
-	c = strchr(custom_header, '\r');
-	if (c != NULL)
-		return false;
-	c = strchr(custom_header, '\n');
-	if (c != NULL)
-		return false;
+	c = strchr(http_header, '\r');
+	if (c)
+		return true;
+	c = strchr(http_header, '\n');
+	if (c)
+		return true;
 
 	// Require a header name followed by :
-	c = strchr(custom_header, ':');
-	if (c == NULL)
-		return false;
-	name_len = c - custom_header;
+	name_len = http_header_name_length(http_header);
 	if (name_len < 1)
-		return false;
+		return true;
+
+	return false;
+}
+
+static char *forbidden_custom_headers[] = {
+	"User-Agent",
+	"Host",
+	"Accept",
+	"Content-Type",
+	"Transfer-Encoding",
+	"Content-Length",
+};
+
+bool is_forbidden_custom_header(const char *custom_header)
+{
+	unsigned long i;
+	int name_len = http_header_name_length(custom_header);
 
 	// Disallow headers that we set
 	for (i = 0; i < ARRAY_SIZE(forbidden_custom_headers); i++)
 		if (strncmp(forbidden_custom_headers[i], custom_header, name_len) == 0)
-			return false;
+			return true;
 
-	return true;
-}
-
-const char *find_invalid_custom_header(const git_strarray *custom_headers)
-{
-	size_t i;
-
-	if (custom_headers == NULL || custom_headers->count == 0)
-		return NULL;
-
-	for (i = 0; i < custom_headers->count; i++)
-		if (!is_valid_custom_header(custom_headers->strings[i]))
-			return custom_headers->strings[i];
-
-	return NULL;
+	return false;
 }
 
 static int git_smart__set_custom_headers(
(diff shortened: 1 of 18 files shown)


carlosmn:
I'd rather keep these kind of options away from setters and put the `strarray` in `git_push_options` and `git_fetch_options`.

carlosmn:
We would also want to make sure we don't end up overwriting important header fields, or sending them twice.

spraints:
Thanks for the feedback.

The custom headers are now in `git_push_options`, along with changes to how they're sent down to the actual HTTP request. The headers are also included in the winhttp transport, though I can't compile that locally to try it out. There were a couple other changes that affected the public API.

* Part of connecting push options to the transport went through `git_remote_connect`, so it accepts a `custom_headers` argument.
* `git_transport` (in `include/sys/transport.h`) has a new function, `set_custom_headers`.

@carlosmn @ethomson - Does this look better?

This still needs to ensure that the custom headers are well-formed and don't replace built-in headers.

carlosmn:
This is looking much better. We would also want to have this be available for fetching though, rather than just pushing.

spraints:
> We would also want to have this be available for fetching though, rather than just pushing.

Cool. c49126c adds that. 8c876fa is a first cut at validating headers. I'd like to add unit tests at least for the [header validation](https://github.com/spraints/libgit2/blob/custom-push-headers/src/transports/smart.c#L69-L131). I haven't dug into the tests yet to see where the best place to test this is. Are there http push or fetch tests that I should add to?

spraints:
> I haven't dug into the tests yet to see where the best place to test this is. Are there http push or fetch tests that I should add to?

3245896 adds a test for header validation to the clone tests.

I think this is done. It's ready for review :eyes:. Let me know what needs to be fixed up.

spraints:
I tried d29c541 (via https://github.com/spraints/rugged/commit/fb3bc7eed24108dbbdfb653afe704ae7c06b8b72) in production today, and it's working well.

src/transports/smart.c
```diff
@@ -66,6 +66,76 @@ static int git_smart__set_callbacks(
...
+	if (name_len < 1)
+		return false;
+
+	// Disallow headers that we set
+	forbid_custom_header("User-Agent");
+	forbid_custom_header("Host");
+	forbid_custom_header("Accept");
+	forbid_custom_header("Content-Type");
+	forbid_custom_header("Transfer-Encoding");
+	forbid_custom_header("Content-Length");
```
carlosmn: This construct feels foreign to C. I would rather see a list in a static `git_strarray` which we loop over. If we were to stick to macros, we should pass in the string we're validating instead of relying on a particular one being defined.

The macro should then be all-caps and have the `do { } while (0)` surrounding it, as it consists of more than one line.

src/transports/smart.c
```diff
@@ -66,6 +66,76 @@ static int git_smart__set_callbacks(
...
+
+static int git_smart__set_custom_headers(
+	git_transport *transport,
+	const git_strarray *custom_headers)
+{
+	transport_smart *t = (transport_smart *)transport;
+	const char *invalid_header = find_invalid_custom_header(custom_headers);
+
+	if (invalid_header != NULL) {
+		giterr_set(GITERR_INVALID, "Illegal HTTP header '%s'", invalid_header);
```
carlosmn: This message would suggest that `Host` is an ilegal header, since we consider it invalid. It would be much clearer if we could distinguish between a header field we don't allow to override and one which the user passed in malformed.

The error message should not start with a capital letter.

carlosmn:
As general notes: error messages should not be capitalised, they're often in the middle of sentences. I don't think we protect against a string in a `git_strarray` being `NULL` anywhere else. While it's not wrong, I don't know if we expect to work when the user passes `NULL` in there.

src/transports/smart.c
spraints: :cool: 098f1e6 changes to an array of forbidden custom headers.

src/transports/smart.c
spraints: e60db3c makes two error messages. If the header doesn't look like a header (e.g. if it doesn't have a name followed by ":"), report "custom HTTP header '%s' is malformed". If the header has the same name as a header already set by libgit2 (e.g. "Host"), report "HTTP header '%s' is already set by libgit2".

spraints:
> I don't think we protect against a string in a git_strarray being NULL anywhere else. While it's not wrong, I don't know if we expect to work when the user passes NULL in there.

I saw at least one (in [`git_strarray_copy`](https://github.com/libgit2/libgit2/blob/588e28ebfd16dc030be3226e4bc2107cdbad67b4/src/util.c#L51)). That is a different type of code, though, more general purpose. The rest of the library looks like it doesn't guard against `NULL` in a `git_strarray`, so 63cc572 drops it for the `custom_headers` array, too.

src/transports/smart.c
```diff
@@ -66,6 +66,80 @@ static int git_smart__set_callbacks(
...
+			giterr_set(GITERR_INVALID, "custom HTTP header '%s' is malformed", custom_headers->strings[i]);
+			return -1;
+		}
+		if (is_forbidden_custom_header(custom_headers->strings[i])) {
+			giterr_set(GITERR_INVALID, "custom HTTP header '%s' is already set by libgit2", custom_headers->strings[i]);
+			return -1;
+		}
+	}
+
+	t->custom_headers = custom_headers;
```
spraints: This assumes that the `git_strarray *custom_headers` won't go away before the transport uses it. Is that valid? If you do this:

```c
git_remote *remote;
do_fetch_with_headers(remote);
do_fetch(remote);

void do_fetch_with_headers(git_remote *remote) {
  git_fetch_options opts = /* ... */;
  opts.custom_headers = /* ... */;
  git_remote_fetch(remote, _, &opts, _);
}

void do_fetch(git_remote *remote) {
  git_remote_fetch(remote, _, NULL, _);
}
```

Does the second `git_remote_fetch` reuse any of the transport stuff from the first one? If it does reuse the transport, would it be better to do a `git_strarray_copy` instead of just copying the pointer? I had a hard time figuring this out by reading the code. I can take another look at it, but I'd appreciate another set of :eyes:.

carlosmn: My general advice is not to re-use remote instances, but we should indeed not crash if someone does it. The `git_remote_fetch()` disconnects from the remote, which means that the second `_fetch()` would have to create a new transport, so the options would not be carried over.

But it is generally good form to defensively copy the array, since this is a setter. While keeping the pointer should be OK in the general case (which is why I didn't comment on it), we could have someone use the lower-level functions (the ones which `git_remote_{fetch,push}()` call), which would introduce a chance for the caller to change the array, or garbage-collect it, or...

So overall, copying the strarray would be good. I don't expect memory usage here to be a concern, so copying a few strings will be fine.

spraints: > So overall, copying the strarray would be good. I don't expect memory usage here to be a concern, so copying a few strings will be fine.

:cool: d737566 makes this change.

carlosmn:
:+1: we should eventually have concrete tests for this, but that's going to be a bit of a pain to do across the tests, so we can figure it out later. @ethomson ?

ethomson:
I agree.  More tests here would be nice to have, but also quite painful to write.  I don't think that we should wait to merge this until we have tests here.

My only bit of feedback is to ask at what point `git_remote_connect` gets an options structure?  If we think we're going to continue packing more arguments on in the future, it may be more future-friendly to switch to a struct.

carlosmn:
`_connect()` might yet get an options structure, but I think it'd be better to leave that for later, as it might impact what structs we use in in the fetch and push structures. We should probably make it accept a struct before v1 so we can expand it, but I wouldn't want to add yet another embedded struct in `git_{fetch,push}_options`. At least not for the moment.

I'm merging this, and we can figure out the argument and testing issue later.

Issue #620, opened 2012-04-04
Handle symlinks in .git directory


rrevans:
libgit2 clobbers symlinks in the .git directory instead of updating the linked file.

This behavior is incompatible with tools like git's [contrib/workdir/git-new-workdir](https://github.com/gitster/git/blob/master/contrib/workdir/git-new-workdir).  This tool uses symlinks to create a new working directory for an existing git repository.  Most of the metadata files become symlinks to another .git directory (except HEAD).  This setup works with git but not libgit2 because git's file update implementation honors symbolic links.

git_filebuf_open should resolve symbolic links (including nested links) when opening files for writing.  See git's implementation in [lockfile.c](https://github.com/gitster/git/blob/master/lockfile.c#L80).

carlosmn:
Fixed with #3413

PR #3413, opened 2015-09-05
filebuf: follow symlinks when creating a lock file

carlosmn:
We create a lockfile to update files under GIT_DIR. Sometimes these
files are actually located elsewhere and a symlink takes their place. In
that case we should lock and update the file at its final location
rather than overwrite the symlink.

---

It's quite late, but this should take care of #620.

The code to resolve the symlink is not pretty at all, but it seems we don't have this anywhere else. It might be worth putting in a more general place if there's a need to do this elsewhere, though I don't know if we want to follow symlinks in any other situation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ v0.23 + 1
   example `filter=*`.  Consumers should examine the attributes parameter
   of the `check` function for details.
 
+* Symlinks are now followed when locking a file, which can be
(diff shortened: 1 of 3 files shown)


ethomson:
Yep, let's leave that symlink resolution code there for now.  I'm surprised we haven't needed it yet, but we can punt it somewhere more generic later if we need it.

Thanks!

PR #3415, opened 2015-09-08
filebuf: ensure we can lock a hidden file

ethomson:
Tests the alternate method by which #3310 might be failing.


ethomson:
I added these tests to investigate failures around writing hidden files on Windows - which I can't find.  I think that we should probably merge this to avoid regressions, and in case either somebody sees some test failures, then we can fix them.

In the meantime, I'm going to close #3310.

PR #3418, opened 2015-09-10
pkg-config: fix directory references in libgit2.pc

DimStar77:
Before:
libdir=/usr//usr/lib64
includedir=/usr//usr/include

After:
libdir=/usr/lib64
includedir=/usr/include

(note the duplication of /usr in the before case)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,23 @@ SET(BIN_INSTALL_DIR bin CACHE PATH "Where to install binaries to.")
 SET(LIB_INSTALL_DIR lib CACHE PATH "Where to install libraries to.")
 SET(INCLUDE_INSTALL_DIR include CACHE PATH "Where to install headers to.")
 
+# Set a couple variables to be substituted inside the .pc file.
+# We can't just use LIB_INSTALL_DIR in the .pc file, as passing them as absolue
+# or relative paths is both valid and supported by cmake.
+SET (PKGCONFIG_PREFIX ${CMAKE_INSTALL_PREFIX})
+
+IF(IS_ABSOLUTE ${LIB_INSTALL_DIR})
+  SET (PKGCONFIG_LIBDIR ${LIB_INSTALL_DIR})
+ELSE(IS_ABSOLUTE ${LIB_INSTALL_DIR})
+  SET (PKGCONFIG_LIBDIR "\${prefix}/${LIB_INSTALL_DIR}")
+ENDIF (IS_ABSOLUTE ${LIB_INSTALL_DIR})
+
+IF(IS_ABSOLUTE ${INCLUDE_INSTALL_DIR})
+  SET (PKGCONFIG_INCLUDEDIR ${INCLUDE_INSTALL_DIR})
+ELSE(IS_ABSOLUTE ${INCLUDE_INSTALL_DIR})
+  SET (PKGCONFIG_INCLUDEDIR "\${prefix}/${INCLUDE_INSTALL_DIR}")
+ENDIF(IS_ABSOLUTE ${INCLUDE_INSTALL_DIR})
+
 FUNCTION(TARGET_OS_LIBRARIES target)
 	IF(WIN32)
 		TARGET_LINK_LIBRARIES(${target} ws2_32)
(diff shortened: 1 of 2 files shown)


carlosmn:
What options are you passing to CMake? I don't think I've ever noticed a problem with this before.

DimStar77:
The cmake call used is based on our script/macro used in all packages... for libgit2, the cmake command is:

[   71s] + /usr/bin/cmake .. -DCMAKE_INSTALL_PREFIX:PATH=/usr -DCMAKE_INSTALL_LIBDIR:PATH=/usr/lib64 -DINCLUDE_INSTALL_DIR:PATH=/usr/include -DLIB_INSTALL_DIR:PATH=/usr/lib64 -DSYSCONF_INSTALL_DIR:PATH=/etc -DSHARE_INSTALL_PREFIX:PATH=/usr/share -DCMAKE_BUILD_TYPE=None '-DCMAKE_C_FLAGS=-fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector -funwind-tables -fasynchronous-unwind-tables -g -DNDEBUG' '-DCMAKE_CXX_FLAGS=-fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector -funwind-tables -fasynchronous-unwind-tables -g -DNDEBUG' '-DCMAKE_Fortran_FLAGS=-fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector -funwind-tables -fasynchronous-unwind-tables -g -DNDEBUG' '-DCMAKE_EXE_LINKER_FLAGS=-Wl,--as-needed -Wl,--no-undefined -Wl,-z,now' '-DCMAKE_MODULE_LINKER_FLAGS=-Wl,--as-needed -Wl,--no-undefined -Wl,-z,now' '-DCMAKE_SHARED_LINKER_FLAGS=-Wl,--as-needed -Wl,--no-undefined -Wl,-z,now' -DLIB_SUFFIX=64 -DCMAKE_SKIP_RPATH:BOOL=ON -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DBUILD_SHARED_LIBS:BOOL=ON -DBUILD_STATIC_LIBS:BOOL=OFF -DCMAKE_COLOR_MAKEFILE:BOOL=OFF -DCMAKE_INSTALL_DO_STRIP:BOOL=OFF -DCMAKE_MODULES_INSTALL_DIR=/usr/share/cmake/Modules -DCMAKE_USER_MAKE_RULES_OVERRIDE=/usr/share/cmake/Modules/opensuse_rules.cmake -DTHREADSAFE:BOOL=ON -DUSE_SSH:BOOL=ON

carlosmn:
This is rather different from how CMake sets up these variables itself. When I look at its cache file when running `cmake -DCMAKE_INSTALL_PREFIX=/usr`, it stores `LIB_INSTALL_DIR:PATH=lib`. That is, it expects the path relative to the prefix, not the absolute path, which is how we expect to find it when we write out our `.pc` file.

DimStar77:
http://www.cmake.org/pipermail/cmake/2012-February/049206.html

As long as you use the variables only for installation, it's no big issue... when you want to substitute, we need more logic.

carlosmn:
In that case, this PR isn't enough and we'd need to ask CMake to join the paths and then substitute the result of that.

DimStar77:
Agreed - I will work on something more complete that works in my setup... and hopefully won't break yours.

DimStar77:
A new diff... work is based on how libproxy handles the same problem (see http://github.com/libproxy/libproxy or more specifically https://github.com/libproxy/libproxy/blob/master/libproxy/cmake/devfiles.cmk )

carlosmn:
That seems fine, but this new code doesn't follow the style of the rest of the file. We put the `IF`, `SET`, etc. in capitals.

DimStar77:
PR updated... keyword written in CAPS

carlosmn:
Thanks.

PR #3419, opened 2015-09-10
Fixing dangling pointers in git_mempack_reset

bwoods:
The first time `git_mempack_reset` is called it frees the objects in `db->objects` but the pointers to those objects are left in place. So if an attempt is made to reset the mempack again, a “double free” occurs.

### DIFF
--- a/src/odb_mempack.c
+++ b/src/odb_mempack.c
@@ -161,8 +161,12 @@ void git_mempack_reset(git_odb_backend *_backend)
 
 static void impl__free(git_odb_backend *_backend)
 {
-	git_mempack_reset(_backend);
-	git__free(_backend);
+	struct memory_packer_db *db = (struct memory_packer_db *)_backend;
+
+	git_mempack_reset(db);
+	git_oidmap_free(db->objects);
+
+	git__free(db);
 }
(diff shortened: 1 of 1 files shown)


ethomson:
Good catch!

I think that we're leaking memory in the `free` function, though (both before this change and after it).

`impl__free` will call reset, which will realloc that array right after freeing it.

I wonder if `git_mempack_reset` might want to `git_oidmap_clear()`, and then `impl_free()` might want to do the final `git_oidmap_free()`?

bwoods:
I like the idea of a separate `git_oidmap_clear`, but as it doesn’t seem to exist yet, I’ve added the more obvious fix to the leak to the current pull request.

ethomson:
Hmm.  You need to pass a `git_odb_backend` to reset to avoid a compiler warning.  But let's definitely not `malloc` something just to `free` it immediately.

I exposed the khash clear function as `git_oidmap_clear`.  Can you double-check #3421 ?

bwoods:
Perfect. And the leak is still gone, as expected.

ethomson:
Awesome - thanks again!

PR #3424, opened 2015-09-13
Don't free config in `git_transaction_commit`.

arthurschreiber:
The config is not owned by the transaction, so please don’t free it.

### DIFF
--- a/tests/config/write.c
+++ b/tests/config/write.c
@@ -670,6 +670,16 @@ void test_config_write__locking(void)
 	git_transaction_free(tx);
 
 	/* Now that we've unlocked it, we should see both updates */
+	cl_git_pass(git_config_get_entry(&entry, cfg, "section.name"));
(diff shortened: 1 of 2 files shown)


tests/config/write.c
```diff
@@ -670,7 +670,6 @@ void test_config_write__locking(void)
 	git_transaction_free(tx);
 
 	/* Now that we've unlocked it, we should see both updates */
-	cl_git_pass(git_config_open_ondisk(&cfg, filename));
```
carlosmn: We do want to load the file independently to test that the changes are visible from the outside. I would add a `git_config_free(cfg)` after `git_transaction_free()` instead.

arthurschreiber: I added a separate block beneath. That way, we can check the changes through the `git_config` that was part of the transaction, and then test that the changes were really persisted.

PR #3425, opened 2015-09-13
Handle `git_path_diriter` instances at the drive root on Windows

### LINKED ISSUE #3417: I have find a bug for "git_status_list_entrycount()"
My OS is win7. This is my code
```
...
git_repository_open(&repo, "E:\\cpp_app\\BoostApp"); 
git_status_list *status;
git_status_options statusopt = GIT_STATUS_OPTIONS_INIT;
git_status_list_new(&status, repo, &statusopt);

int maxi = git_status_list_entrycount(status);
...
```
In this case, The value of maxi is right. But when I use "subst" command create the virtual disc, 
```subst i: e:\cpp_app\BoostApp```

And then, change the code at  open function
```
git_repository_open(&repo, "i:"); 
```
this time, the value of maxi is error, it looks like a random number , such as "36","1202", but it is right value is "0".

ethomson:
On Windows, our `git_path_diriter` adds a '/' to combine the path being queried with a filesystem entry.  For example, when iterating a directory `c:/foo` that contains some file `bar.txt`, we create a path `c:/foo/bar.txt`.  (Yay!)

We need to not do this on the root of a volume (eg, `c:/`) since it is already, correctly, suffixed with a `/`.

Fixes #3417

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -1166,7 +1166,11 @@ static int diriter_update_paths(git_path_diriter *diriter)
 	diriter->path[path_len-1] = L'\0';
 
 	git_buf_truncate(&diriter->path_utf8, diriter->parent_utf8_len);
-	git_buf_putc(&diriter->path_utf8, '/');
+
+	if (diriter->parent_utf8_len > 0 &&
+		diriter->path_utf8.ptr[diriter->parent_utf8_len-1] != '/')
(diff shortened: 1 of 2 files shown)


ethomson:
Happens on POSIX, too, of course.

PR #3429, opened 2015-09-16
Checkout: handle mode changes

### LINKED ISSUE #3158: Rebase fails if file in picked commit has both content and permissions change
There seems to be an issue in the rebase functionality somewhere. I have trouble pinpointing where the issue is exactly but I have managed to create a test case to reproduce it.

This issue needs an operating system where the filesystem has permissions. It seems to affect Linux. OSX seems **not** to be affected, I was told. Windows probably isn't relevant for this issue.

## How to reproduce

Get the test case that shows the issue: https://github.com/akikoskinen/libgitrebasebug

Run the `createrepo.sh` script which creates a git repository that is used for the demonstration. This script can be run again and again in order to recreate the test repository - it removes any previous repository first.

In [`CMakeLists.txt` on line 5](https://github.com/akikoskinen/libgitrebasebug/blob/master/CMakeLists.txt#L5) modify the path to libgit2 code.

In [`main.c` line 6](https://github.com/akikoskinen/libgitrebasebug/blob/master/main.c#L6) you can specify whether using libgit2 0.22 branch or master branch. Both branches seem to be affected.

Build the test case and run it (see below). It terminates into an assert but it shouldn't.

    mkdir build && cd build && cmake .. && make && ./libgitrebasebug

## What's going on?

Look at the test repository (right after it's created, before running the test program). There's a `master` branch and a branch called `other`. Master branch is currently checked out.

One file is modified in the commit on the top of the master branch: `problem.txt`. There are two different kind of changes in that file: its contents and its file permissions have been changed.

On the `other` branch the `problem.txt` file is not modified at all.

Rebasing `master` on top of `other` should succeed just fine. Indeed `git rebase other` does that just fine.

The test program tries to do exactly what `git rebase other` does but it fails to do that.

## Other info

The problem seems to occur only when both the contents and the file permissions of the `problem.txt` file are changed. If only one or the other kind of change is present then the rebase succeeds.

ethomson:
As reported in #3158, we do not quite handle mode changes correctly in checkout.  We will always set execute bits properly, but we fail to *unset* execute bits.

This ports the unit tests that were mentioned in #3158.  It does not take the fix that was mentioned in that issue, though, as it was heavy-handed.  Instead, we simply delete the old file and let `p_open` set the mode for us and drop any notion of trying to `chmod`.  (This is handy since we were not providing the umask properly, nor is there a sane way to query the umask in a thread safe manner across architectures.)  

Fixes #3158

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -243,6 +243,12 @@ static int checkout_action_common(
 		if (delta->new_file.mode == GIT_FILEMODE_LINK && wd != NULL)
 			*action |= CHECKOUT_ACTION__REMOVE;
 
+		/* if the file is on disk and doesn't match our mode, force update */
+		if (wd &&
+			GIT_PERMS_IS_EXEC(wd->mode) !=
+			GIT_PERMS_IS_EXEC(delta->new_file.mode))
+				*action |= CHECKOUT_ACTION__REMOVE;
+
 		notify = GIT_CHECKOUT_NOTIFY_UPDATED;
 	}
 
@@ -1499,15 +1505,6 @@ static int blob_content_to_file(
 
 	if (error < 0)
 		return error;
-
-	if (GIT_PERMS_IS_EXEC(mode)) {
-		data->perfdata.chmod_calls++;
-
-		if ((error = p_chmod(path, mode)) < 0) {
-			giterr_set(GITERR_OS, "Failed to set permissions on '%s'", path);
-			return error;
-		}
-	}
 
 	if (st) {
 		data->perfdata.stat_calls++;
--- a/tests/checkout/tree.c
+++ b/tests/checkout/tree.c
@@ -973,6 +973,17 @@ void test_checkout_tree__filemode_preserved_in_index(void)
 	git_commit_free(commit);
 
 
+	/* Finally, check out the text file again and check that the exec bit is cleared */
+	cl_git_pass(git_oid_fromstr(&executable_oid, "cf80f8de9f1185bf3a05f993f6121880dd0cfbc9"));
+	cl_git_pass(git_commit_lookup(&commit, g_repo, &executable_oid));
+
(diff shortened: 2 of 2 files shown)


src/checkout.c
```diff
@@ -243,6 +243,12 @@ static int checkout_action_common(
 		if (delta->new_file.mode == GIT_FILEMODE_LINK && wd != NULL)
 			*action |= CHECKOUT_ACTION__REMOVE;
 
+		/* if the file is on disk and doesn't match our mode, force update */
+		if (wd &&
+			GIT_PERMS_IS_EXEC(wd->mode) ^
+			GIT_PERMS_IS_EXEC(delta->new_file.mode))
```
carlosmn: I'm wondering if this has an adavantage over writing `!=` since we're dealing with a boolean on each side, no?

tests/checkout/tree.c
```diff
@@ -973,9 +973,86 @@ void test_checkout_tree__filemode_preserved_in_index(void)
...
+mode_t read_filemode(const char *path)
+{
+	git_buf fullpath = GIT_BUF_INIT;
+	struct stat st;
+	mode_t result;
+
+	git_buf_joinpath(&fullpath, "testrepo", path);
+	cl_must_pass(p_stat(fullpath.ptr, &st));
+
+	result = GIT_PERMS_IS_EXEC(st.st_mode) ? 0100755 : 0100644;
```
carlosmn: We do have defines for this to avoid magic numbers, even if they're rather recognizable.

src/checkout.c
ethomson: absolutely zero advantage.  yay for 2am pull requests.

ethomson:
Fixed, thanks @carlosmn

ethomson:
I zapped a few other places that had some hardcoded modes and were likely to fail in a `umask` that isn't `022`.

I wonder if we should set random umasks on the build machines to catch brittle tests and / or shitty mode handling.  But that's a different PR.

PR #3432, opened 2015-09-17
`mkdir`: split into `mkdir` and `mkdir_relative`

### LINKED ISSUE #3068: libgit2 tries to remove out-of-tree symlinks
loot/loot#417 details an issue filed by one of my users, who had replaced their AppData folder with a junction link to another location. My program clones a repository to a subfolder within AppData, and they were finding that libgit2 was returning the following error when it tried to do so.

> `Error: -4; Failed to remove symlink 'C:/Users/UsrName/AppData': Access is denied.`

Adding the `GIT_CHECKOUT_DONT_REMOVE_EXISTING` checkout strategy flag to the clone options structure (I was already using `GIT_CHECKOUT_FORCE`) prevented the error from occurring. 

However, my understanding is that libgit2 should be trying to remove files/symlinks *within* the target directory, not removing symlinks at all points in the path to that directory, so this looks like a bug.

ethomson:
When attempting to create a directory recursive (eg, matching `mkdir -p` semantics), we walk *up* from the given `base` to determine what directories need to be created.  For checkout, this is right and good and true as we may stumble across a symbolic link within the working directory, and this needs to be *deleted* and re-created as a proper directory.

However, we should only do this walk up when we have a `base` to build relative paths from (eg, when operating in the repository or working directory).  Otherwise, we'll end up trying to walk all the way up from the directory root to the path that we're trying to create.  This is not correct.  For example: we shouldn't be examining `/home` to create a repository in `/home/user/foo/myrepo.git` - we may get hung up trying to stat `/home`.

This PR splits `git_futils_mkdir` into two functions:  `git_futils_mkdir`, which is a general use `mkdir` type function that creates a directory (perhaps creating intermediate directories ala `mkdir -p`) and `git_futils_mkdir_relative`, which creates a directory relative to a given base directory.

`git_futils_mkdir` - when in recursive creation mode - now examines the requested path and walks toward the root to determine the paths to create, then invokes `git_futils_mkdir_relative` to do the work.

This could have been performed keeping `git_futils_mkdir` a single function and switching on a `NULL` base, but the added complexity was becoming quite horrible to understand.  The `mkdir` function(s) already change behavior significantly based on options like `GIT_MKDIR_SKIP_LAST2` and `GIT_MKDIR_CHMOD_PATH`.  Adding yet another significant behavioral change based on the presence or absence of `base` was complex and an invitation to use the function wrong.

I think that there's opportunity to remove some of these options (and thus clean up the implementation) in the future, but this should be both helpful and correct in the meantime.

Fixes #3068, #3308.

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -1366,7 +1366,7 @@ static int checkout_mkdir(
 	mkdir_opts.dir_map = data->mkdir_map;
 	mkdir_opts.pool = &data->pool;
 
-	error = git_futils_mkdir_ext(
+	error = git_futils_mkdir_relative(
 		path, base, mode, flags, &mkdir_opts);
 
 	data->perfdata.mkdir_calls += mkdir_opts.perfdata.mkdir_calls;
--- a/src/fileops.c
+++ b/src/fileops.c
@@ -331,6 +331,138 @@ GIT_INLINE(int) validate_existing(
 	return 0;
 }
 
+	
+GIT_INLINE(int) mkdir_canonicalize(
+	git_buf *path,
+	uint32_t flags)
+{
+	ssize_t root_len;
+
+	if (path->size == 0) {
+		giterr_set(GITERR_OS, "attempt to create empty path");
+		return -1;
+	}
+
+	/* Trim trailing slashes (except the root) */
+	if ((root_len = git_path_root(path->ptr)) < 0)
+		root_len = 0;
+	else
+		root_len++;
+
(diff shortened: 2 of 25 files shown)


ethomson:
I want to do some additional manual testing before we merge this, since it's quite tough to make unit tests that do crazy things to permissions in a win32 filesystem.

ethomson:
Okay, I manually validated both #3068 and #3308 and they look good to me.

PR #3433, opened 2015-09-18
Keep config comments in the same place as git

carlosmn:
This tries to solve the same issue as #3368 while reducing the amount of changes that are needed by using the buffering technique described by @ethomson in that thread.

We buffer the comments until we are forced to write them by reaching a section end, a variable or the end of the file.


ethomson:
Thanks!  Much cleaner.

PR #3434, opened 2015-09-18
Win32 Reserved names: don't reserve names outside the working directory

ethomson:
Don't try to reserve path names for the name of the repository when the repository is not underneath the working directory.

(Generally the repository is `.git`, immediately beneath the working directory.  In this case we need to figure out its shortname on windows, which is probably something like `GIT~1`, and we need to treat that as a reserved name.)

However, the repository need not be beneath the working directory at all.  In that case, we don't care about trying to figure out its shortname and keeping people from checking it out.  This comes up most often in the case of submodules, where a gitlink points to some repository that is inside the super repository (eg, `super/.git/modules/sub`).  In that case, we should *not* try to reserve the short name of the git repository (in this case, `sub`).  This would prevent somebody from creating the perfectly reasonable filename `sub`.

#2927 tried to solve this, but did so pretty poorly.  There's a race in `git_submodule_update` where the gitlink doesn't yet exist as we're checking out the working directory, so we end up reserving the repository name.  Realistically, though, we should ignore *any* repositories that aren't in our working directory, because we simply don't care about what they're called.


ethomson:
Sigh, apparently I didn't get the submodule fixture magic quite right.

ethomson:
image

PR #3435, opened 2015-09-23
win32: give better error message on long paths

ethomson:
Reset the windows error message when we detect that a path was too long, setting the windows error to `ERROR_FILENAME_EXCED_RANGE`,

(Why the Windows team in their infinite wisdom decided to save that single character to avoid spelling `EXCEED` correctly, I'll be damned if I know.)

We only test clone, where we cannot *create* a file whose name is longer than `MAX_PATH`.  We do not have a test where we try to read a file whose name is longer, because we'd have to first *create* said file and all our internal path munging is already unable to do that.


ethomson:
(Right now the error message is complaining about the buffer length to the unicode conversion function, which is really hard to understand.)

PR #3436, opened 2015-09-23
Worktree implementation

pks-t:
These patches introduce initial support for work trees, which have been introduced with git v.2.5.

Criticism and proposals welcome.

Things that need to be tackled in order to work with $COMMONDIR:
- [x] ref lookup
- [x] loose lookup
- [x] packed lookup
- [x] config
- [x] logs/

The following directories are not really used by us but would require to be made aware of $COMMONDIR:
- [ ] shallow/
- [ ] info/
- [ ] remotes/  (legacy)
- [ ] branches/ directory (legacy)
- [ ] hooks/

Additional things that need to be cared about:
- [x] creating new working trees
- [x] restrictions on ability to check out refs that are being linked to
- [x] restrictions on ability to delete refs that are being linked to (only enforced for `git_branch_delete` function)
- [x] pruning (abandoned) working trees
- [ ] renaming references?


src/refdb_fs.c
```diff
@@ -63,6 +66,18 @@ typedef struct refdb_fs_backend {
 	uint32_t direach_flags;
 } refdb_fs_backend;
 
+/*
+ * List of symbolic reference names that are always stored in the
+ * gitdir. This is required such that we are able to correctly
+ * resolve these references via the gitdir instead of the
+ * commondir when working trees are used.
+ */
+static const char *symrefs[] = {
```
ethomson: I suspect that we should give this a more meaningful name.  (And the `is_symbolic_ref` function, also, below.)  You could have a symbolic ref `refs/heads/foo` -> `refs/heads/bar`, which would be in the commondir not the gitdir.

I don't have any brilliant ideas though.  Special refs?  Local refs?  Workdir refs?

src/refdb_fs.c
```diff
@@ -55,7 +55,10 @@ typedef struct refdb_fs_backend {
 	git_refdb_backend parent;
 
 	git_repository *repo;
-	char *path;
+	/* path to git directory */
+	char *gitpath;
+	/* path to common objects' directory */
+        char *commonpath;
```
ethomson: Spaces here instead of tabs.

src/repository.c
```diff
@@ -152,20 +155,35 @@ void git_repository_free(git_repository *repo)
...
 
-	if (git_path_contains_dir(repository_path, GIT_REFS_DIR)  == false)
+	/* Check files in common dir */
+	if (git_path_contains_dir(common_path, GIT_OBJECTS_DIR) == false)
+		return false;
+	if (git_path_contains_dir(common_path, GIT_REFS_DIR)  == false)
 		return false;
 
-	return true;
+        return true;
```
ethomson: Spaces here instead of tabs.

tests/resources/testrepo/.gitted/worktrees/testrepo-worktree/gitdir
```diff
@@ -0,0 +1 @@
+/home/pks/Development/elego/git/libgit2-worktree/tests/resources/testrepo-worktree/.git
```
ethomson: Can this be a relative path?  If not, we'll have to figure out some mechanism here where we can rewrite this to the sandbox path.  We do something similar with the submodule tests (see [submodule_helpers.c](https://github.com/libgit2/libgit2/blob/master/tests/submodule/submodule_helpers.c)).

src/repository.c
```diff
@@ -152,20 +155,35 @@ void git_repository_free(git_repository *repo)
...
 
 	/* Ensure HEAD file exists */
 	if (git_path_contains_file(repository_path, GIT_HEAD_FILE) == false)
 		return false;
 
-	if (git_path_contains_dir(repository_path, GIT_REFS_DIR)  == false)
+	/* Check files in common dir */
+	if (git_path_contains_dir(common_path, GIT_OBJECTS_DIR) == false)
+		return false;
+	if (git_path_contains_dir(common_path, GIT_REFS_DIR)  == false)
```
ethomson: Extra space here.  (Looks like you inherited it, but let's zap it while we're changing the line.)

ethomson: So I don't love the pattern where the repository path is put into the common path if there wasn't a common path.  I think that this will be confusing later, as I would expect `common_path` to be unset if there's no common path.  I think just a pointer that switches to the repository directory or the common path (if present) would clarify.

ethomson:
Nice!  I'm happy to see that you decided to tackle this, and it looks like this is going to be a nice and clean implementation.

I made a couple of minor comments, but I like where this is going.

pks-t:
Eh. I've already fixed some issues you mention, especially the link-path being relative. Seems I've got an out of date branch on this computer.

src/refdb_fs.c
pks-t: Maybe something like `uncommon_refs`, as those would not be stored in the common directory. Not exactly happy with this name, though, but I'd prefer either this or `workdir_refs`.

tests/resources/testrepo/.gitted/worktrees/testrepo-worktree/gitdir
pks-t: Well, at least my implementation works with relative paths. Thought I did already fix this in my branch, seems as I've been pushing an outdated copy. Guess it's unfortunate that git recognizes those files as binary files, thus the diff didn't show it.

src/repository.c
pks-t: Honestly I have been considering your solution, as well. I settled for the solution of always setting `common_path` as it simplifies logic in some places. Also, consider the case for a repository that's got linked worktrees. If I opened that repository, I'd get `common_path=NULL` in your case, as it is the main repository. But actually `gitdir` is still shared between this repository and its linked working trees and thus I'd consider it a common dir, as well.

src/refdb_fs.c
carlosmn: I think that these are often referred to as pseudorefs in git, or per-worktree refs, as they exist at the top level and exist for each worktree.

src/refdb_fs.c
```diff
@@ -63,6 +66,18 @@ typedef struct refdb_fs_backend {
...
+ * List of symbolic reference names that are always stored in the
+ * gitdir. This is required such that we are able to correctly
+ * resolve these references via the gitdir instead of the
+ * commondir when working trees are used.
+ */
+static const char *symrefs[] = {
+    "HEAD",
+    "ORIG_HEAD",
+    NULL,
+};
```
carlosmn: Is this seems to be missing `MERGE_HEAD` and `FETCH_HEAD`, or are those handled differently? Is there really a list of these or should we perform the `{_,}HEAD` suffix check like we do to determine if we want to allow writing to a particular ref?

ethomson: Well, those aren't really *refs*, per se, they're sort of ref-like but in a non-ref irregular format and should be outside the scope of the refdb.

Though to be completely honest, I would have expected the same was true for `ORIG_HEAD`.  Do we really use the refdb infrastructure for `ORIG_HEAD`?

pks-t:
Rebase onto master and fixed the absolute gitdir path. No changes yet besides that.

src/refdb_fs.c
pks-t: They are pseudorefs, as @carlosmn previously pointed out (see [gmane](http://permalink.gmane.org/gmane.comp.version-control.git/275036)).
The list was not complete yet, wanted to come back to it later. Should've left a TODO in there, I guess. I don't exactly know how pseudorefs are handled yet, I'll have to figure that out.

src/refdb_fs.c
pks-t: I've removed this array altogether as there is currently only one per-worktree ref that needs to be handled this way, which is `HEAD`. I've also renamed `is_symbolic_ref` to `is_per_worktree_ref`.

pks-t:
I've done some bugfixing, cleanup and further implementation now. As far as opening and acting on worktree-repositories goes I'm currently not aware of any gaping holes, at least it seems to work fine now. Object store, packed refs, loose refs, reflog and configs are all accessible and seem to work, even though I did not yet do any major testing (will do that when I consider this feature ready).

In addition to `git_repository_open` support for working trees I've now started a new interface that acts on working trees for a given repository. I want it to make available the same functionality as the upstream `git worktree` command does, that is listing worktrees, creating new worktrees and pruning old ones. Currently I've only implemented listing worktrees and opening them, everything else will follow soon-ish.

One more missing thing is preventing the user from checking out branches that are already checked out by any worktree, as upstream git does. Guess I'll add a ref-field to `struct git_worktree` that will show which reference is currently checked out and use that in the checkout codepath.

I think the current state (excluding items that are still todo and the WIP-commit) is somewhat sane and cleaned up, so I'd appreciate some feedback.

pks-t:
Meh... will fix some time later this week.

pks-t:
*sigh* No luck for me, I guess. Seems as if AppVeyor is having a hiccup right now.

pks-t:
I consider this branch feature complete now. Some points that remain:

* The test suite on AppVeyor fails somehow. I cannot explain why and am not able to reproduce this error with either Visual Studio 2015 or the Cygwin toolchain. I'll try to get my hands on the compilers/runtimes used by AppVeyor and debug it some times soon.
* Restrictions on references are currently only enforced for the branch API, that is only for `git_branch_delete`. This is based on the fact that the current refs API does not do any protection of checked out branches, not even for the current HEAD branch in a normal repository.

@ethomson @carlosmn I'd appreciate if you could review the changes as soon as you find the time to do so.

pks-t:
Finished my final cleanups. All memory leaks indicated by AppVeyor have been closed, additional tests have been implemented and run without failure on all tested systems, TODOs have been fixed. Also cleaned up the commit graph.

pks-t:
Are there any plans on reviewing this? I realize these are quite a lot of changes to be reviewed but I'd be grateful for some feedback on this.

sanemat:
ping @ethomson @carlosmn

carlosmn:
This will get reviewed, but this means that the core team will have to first learn about the worktree details to be able to know about the decisiosn, and we haven't had a lot of time for that kind of stuff. Many of us are in vacation/break right now, but we'll be coming back soon.

pks-t:
@carlosmn Okay, thanks.

There'll have to be some changes to this PR anyway. On the git mailing list there is currently a discussion ongoing about some backwards-incompatible changes regarding how configs should be split between the linked repository and the worktree (see [1]).

In the process they'll probably introduce a worktree version config which is initially set to 0, similar to the repository format version. As of now it is still undecided, though, but I'll monitor the thread and if there if a solution get's merged I'll make the required changes.

[1]: http://thread.gmane.org/gmane.comp.version-control.git/281906

rburgstaler:
@pks-t  Great Work!  When applied to TortoiseGit, this patch gets WorkTrees working in TortoiseGit for the most part.

While testing out your patches I noticed that repo->commondir will inconsistently include a trailing slash (path delimiter).
In the case of regular work tree: No trailing slash
In the case of a submodule: Trailing slash
In the case of a linked work tree: No trailing slash
In the case of a bare repository: No trailing slash

I noticed that repo->path_repository (git_repository_path) always contained a trailing slash (path delimiter) for the cases stated above.

I am wondering if that is by design?

pks-t:
@rburgstaler Well, to be honest I don't recall right now as it's been some time since I wrote these patches. I'll take a look at it soon, though, so thanks for reporting.

On the "for the most part" part: did you spot any obvious errors or things that did not work with worktrees, except for the trailing-slash-inconsistencie?

rburgstaler:
@pks-t `"for the most part" part: did you spot any obvious errors`.... sorry I should have been a little more clear on this.  The only other things that I noticed were not working were worktree related issues that I tracked down into the source code of TortoiseGit (which is not really relevant for this conversation even though TortoiseGit is my means to test your implementation of worktrees in libgit2)

src/worktree.c
```diff
@@ -133,6 +133,41 @@ int git_worktree_lookup(git_worktree **out, git_repository *repo, const char *na
...
+	assert(out && wt);
+
+	*out = NULL;
+	len = strlen(wt->gitlink_path);
+
+	if (len <= 4 &&
+		wt->gitlink_path[len - 4] != '.' &&
+		wt->gitlink_path[len - 4] != 'g' &&
+		wt->gitlink_path[len - 4] != 'i' &&
+		wt->gitlink_path[len - 4] != 't')
```
joshtriplett: This condition looks wrong; did you mean to check successive characters, rather than the character at `len - 4` each time?  You also want `||` rather than `&&`.

src/refdb_fs.c
```diff
@@ -362,6 +364,13 @@ static const char *loose_parse_symbolic(git_buf *file_content)
 	return refname_start;
 }
 
+static bool is_per_worktree_ref(const char *ref_name)
+{
+	return strcmp("HEAD", ref_name) == 0 ||
+	    strcmp("MERGE_HEAD", ref_name) == 0 ||
+	    strcmp("ORIG_HEAD", ref_name) == 0;
```
joshtriplett: Rather than hardcoding the list of refs that belong in the worktree rather than the common dir, could you generalize this to allow access to other refs in the worktree?  This hardcoded list misses FETCH_HEAD, which comes from the worktree.  And other applications may want to have their own special worktree-state refs similar to HEAD.

I'd suggest treating any top-level ref (not starting with `refs/`) as going in the worktree, at least for any function that doesn't today try to DWIM and prepend `refs/heads` or similar itself.

pks-t:
@joshtriplett Thanks for your feedback. I saw you pinged me on IRC. The PR is currently still pending code review.  It is a little bit more involved and requires the reviewers to explore the worktree feature first, so they'll have to find some more spare time to do the reviews. Also, git.git still consideres the feature experimental, so some implementations details are probably going to change anyway.

That being said I cannot estimate any timeline right now. I welcome any code reviews like yours and guess they'll speed up the process of getting this merged.

src/refdb_fs.c
pks-t: For now I'd rather stick with how upstream git.git treats this for simplicity. As soon as this is merged we may still extend the API to also allow custom refs to be per worktree. Added the missing FETCH_HEAD, though.

pks-t:
Rebased the PR on current master, added FETCH_HEAD to per-worktree refs and fixed gitlink path checking. Thanks @joshtriplett for the hints

dakotahawkins:
I'm really looking forward to this!

cz172638:
any update?

pks-t:
This PR is still awaiting an initial review. Considering worktree support in git core is still experimental and that this touches a lot of internals makes the PR not a priority one.

Two important things not yet fleshed out are configuration rules for worktrees and submodules. Until this is decided for git.git, it does not make much sense to merge this, yet. See discussion at https://marc.info/?l=git&m=146896199207391&w=2.

Still, reviews are welcome so that we are ready when worktrees are finally stable. I also plan on improving this PR for proper configuration support some time in the near future (read: next few weeks).

pks-t:
Rebased the branch against latest master. There were multiple conflicts, but tests seem to pass cleanly. So, to anybody interested in this and wishing to help: the most important thing in here is polishing the PR up. So I'd very much welcome code reviews on this.

The most important points missing are correct configuration support (that is merging worktree as well as other configuration files) as well as reference protection (that is refusing to check out a reference which is already checked out in another worktree.

pks-t:
Coming back to this I just realized that I've already implemented branch protection. Furthermore I think worktree-specific configurations have not yet been implemented, so nothing to do there, as well. Well, I at least put in another test.
So I'm back to awaiting reviews.

max630:
Hi.

I'm trying your PR now (I took it before rebasing, so it might be fixed already), and it seems that it does not set workdir correctly. For a worktree `git_repository_workdir()` should return the directory where .git file is located, ignoring "core.worktree" if it is set. It seems like the code to change is in function `load_workdir`

max630:
Testcase which fail looks like this:

```
void test_worktree_worktree__open_as_repo(void)
{
	git_buf repo_path = GIT_BUF_INIT;
	git_repository *repo;

	cl_git_pass(git_repository_discover(&repo_path, git_repository_workdir(fixture.worktree), 1, NULL));
	cl_git_pass(git_repository_open(&repo, repo_path.ptr));
	cl_assert_equal_s(git_repository_workdir(repo), git_repository_workdir(fixture.worktree));

	git_repository_free(repo);
	git_buf_free(&repo_path);
}
```

with failure:

```
  String mismatch: git_repository_workdir(repo) != git_repository_workdir(fixture.worktree)
  '/tmp/libgit2_tests_5uDPl8/testrepo/.git/worktrees/' != '/tmp/libgit2_tests_5uDPl8/testrepo-worktree/' (at byte 34)
```

Should the worktree be openable as repository at all? It would be very nice if it were, because then it would not require change from a client software.

pks-t:
Thanks for the testcase, will investigate

pks-t:
Rebased on a wrong branch. New commit fixing the issue is 214be36

max630:
Thank you.

But I still fail to use it. First, I happen to have a worktree of submodule, and that does not work:

```
  1) Failure:
worktree::worktree::submodule_worktree_open_as_repo [/home/max/src/repos/git/libgit2/tests/worktree/worktree.c:173]
  String mismatch: git_repository_workdir(repo) != "/tmp/pr3436/sub_worktree"
  '/tmp/pr3436/super_main/.git/modules/sub/' != '/tmp/pr3436/sub_worktree' (at byte 14)
```

The testcase code basically the same; I created the data manually with git commands and replaced fixture with hardcoded paths.

Another issue was observed at Windows (which is basically where I am using this). The thing was that path passed to discover was some `"E:/..repo../..subdir.."`, but the git_repository_workdir() returned `"e:/..repo.."`. The case difference caused failure of [this check](https://github.com/libgit2/libgit2sharp/blob/7f4a17f/LibGit2Sharp/RepositoryExtensions.cs#L288). I cannot really say what would be the best way to fix it. It looks like the check should use case-insensitive comparison, but this ruins the idea to use replacement of `libgit2.dll` only.

pks-t:
Well, on Windows (and every other system with case-insensitive filesystems) you always have to do a case-insensitive comparison of paths. This is nothing that can be fixed in the worktree code.

Regarding opening discovered submodules -- I've created a new test which tries to reproduce your failure (see `test_worktree_open__open_discovered_submodule_worktree`). But the test succeeds, so I guess it is not the same as you were testing. Could you maybe take a look and tell me what the difference is between my test and yours?

max630:
Your case does not seem to test the submodule, the path to worktree gitdir is `"/tmp/libgit2_tests_87YKrq/submodules/testrepo/.git/worktrees/submodules-worktree-child/"` while it sould be something like `../.git/modules/.../worktrees/..`. It might take time for me to learn how to write fixtures properly, but if you check my [gist](https://gist.github.com/max630/7bd073658dea357d16fdcf6c1ff6cfca) it should be working. Basically it runs `git worktree add ...` from an initialized submodule.

pks-t:
No, only the _discovered_ path points to `submodules/testrepo/.git/worktrees/submodules-worktree-child`. The opened repository then points to the top-level working directory of the worktree and to the gitdir. So this seems to be testing and doing the right thing.
```
$ print path->ptr
$10 = 0x555555c77820 "/tmp/libgit2_tests_AqLd49/submodules/testrepo/.git/worktrees/submodules-worktree-child/"
$ print repo->workdir
$11 = 0x555555c79670 "/tmp/libgit2_tests_AqLd49/submodules-worktree-child"
$ print repo->gitdir
$12 = 0x555555c778b0 "/tmp/libgit2_tests_AqLd49/submodules/testrepo/.git/worktrees/submodules-worktree-child/"
```

But yeah, I'm able to provoke a test failure when I do something similar to you:
```C
git_worktree_init(&wt, submodule, "submodule-worktree", "path/to/submodule");
git_worktree_open(&wt_repo, wt);
```
I guess that is your scenario, then.

pks-t:
I've added a new test which first creates a worktree for a submodule and then tries to open it. This catched two issues with existing code.

The first issue is the creation of the `gitdir` link in the worktree's gitdir when using `git_worktree_init`. The `gitdir` file should point back to the worktree's working directory's gitlink (confusing, isn't it?). Instead, we pointed directly to the worktree's working directory.

Furthermore, and I guess this might fix your issue, it solves a problem when opening a submodule's worktree. The problem was that submodules actually have `core.worktree` set in their gitdir which (obviously) points to the submodule's worktree. So when opening a work tree worktree, we first saw that `core.worktree` was set and used that as our worktree instead of the real worktree's worktree. The issue was easily fixed by switching the checks if `core.worktree` is set and if we are trying to open a worktree.

max630:
Now it passes. Also, newer git releases consider drive letters as uppercase, so there should not  be the path case issue for Windows. Good. Thanks.

Now, there another issue: the returned workdir does not contain trailing "/", while for usual repositories it does (see assertions for test_repo_open__standard_empty_repo_through_gitdir for example). This might seem a minor issue but libgit2sharp [relies](https://github.com/libgit2/libgit2sharp/blob/7f4a17f/LibGit2Sharp/RepositoryExtensions.cs#L296) on it, so the BuildRelativePathFrom() returns "/foo.c" instead of "foo.c" and later functions which use file path do not work.

pks-t:
Yeah, I stumbled upon this issue, as well. Didn't really care at that moment as both with and without trailing slash is in fact valid. But if existing implementation always returns a trailing '/', I'll do the same.

pks-t:
`load_workdir` now calls `git_path_to_dir` on the working directory for worktrees.

pks-t:
One more issue fixed when opening a worktree via its gitdir where we wouldn't set the gitlink path. Created more tests to open a worktree via its working directory, gitlink and gitdir and then checking all paths set on the repository.

max630:
Thank you, all tests seem passing, will be trying it.
BTW, I'm checking it with this extension so far: https://github.com/laurentkempe/GitDiffMargin

pks-t:
Thanks for your time testing this :+1:

max630:
In function `git_blob_create_fromstream()`, path to object directory is [counted](https://github.com/libgit2/libgit2/blob/debc31e1716c7580566dc63d2af253eebb272198/src/blob.c#L330) with repository path, should be commondir. Probably not the only place. In the git project they made a refactoring replacing all string concatenations with function like path(repo, subpath), which encapsulated all knowledge about what is common and what is private in one place.

Fixing this, and https://github.com/libgit2/libgit2/issues/3994 , and making sure there is no drive letter path issue, I managed to make GitDiffMargin work. Maybe a bit later I could try to check also the official TFS/Git, though I suspect it might be harder.

pks-t:
Yeah, guess it makes sense to have path-knowledge in one place. See commit 77b15eed303554950e263d0d08626f54ba59e69a and the newly introduced `git_repository_item_path`. c69ece21f6a2dc2d8b5238e0e238adc217b25025 converts most callers which were using `git_repository_path` before.

pks-t:
Rebased against latest master to fix conflicts.

ethomson:
Thanks @pks-t - now that we've got the security release behind us, I'm going to try to make some time this weekend to review some outstanding pull requests, especially this one.  I appreciate your patience and dedication.  🙇 

How are you feeling about this?  Is this generally feature complete?  If not, could you give me a quick rundown of what's missing?

pks-t:
Would be great to have this reviewed. I would really like to have this merged for the next release so that it does not continue to grow more and more. It will only get more tedious to review the longer it takes ;)

I think most things just work for now. I am currently not aware of additional features missing, but I guess that is mostly related to the long time it has been since I last touched this PR. So there will likely be some more things that need to be touched, but the most commonly used features _should_ just work.

Anyway. I think if we are going to merge this for the next release we should review and merge it early in this cycle to have increased testing coverage.

pks-t:
Another rebase fixing conflicts

pks-t:
Rebased to fix merge conflicts.

For people waiting for this to get merged: @carlosmn, @ethomson and me will have a meeting at the upcoming Git Merge to discuss and hopefully get this reviewed. So we might actually be able to make progress soon

pks-t:
I've just been reminded of [worktree-level configuration files](http://public-inbox.org/git/20170110112524.12870-1-pclouds@gmail.com/) as I'm working off my mail inbox. It is not yet mainlined, but something that will probably be coming

include/git2/branch.h
```diff
@@ -246,6 +246,18 @@ GIT_EXTERN(int) git_branch_is_head(
...
 /**
+ * Determine if the current branch is checked out in any linked
+ * repository.
+ *
+ * @param branch Reference to the branch.
+ *
+ * @return 1 if branch is checked out, 0 if it isn't,
+ * error code otherwise.
+ */
+GIT_EXTERN(int) git_branch_is_checked_out(
```
ethomson: It'd be nice to coalesce this into `git_branch_is_head`

src/worktree.c
```diff
@@ -0,0 +1,549 @@
...
+	assert(out && wt);
+
+	*out = NULL;
+	len = strlen(wt->gitlink_path);
+
+	if (len <= 4 ||
+		wt->gitlink_path[len - 4] != '.' ||
+		wt->gitlink_path[len - 3] != 'g' ||
+		wt->gitlink_path[len - 2] != 'i' ||
+		wt->gitlink_path[len - 1] != 't')
```
ethomson: I think that you need to check case-insensitively here.  Note that there's a similar check in `verify_component` in `path.c`; it might be possible to combine these, but I don't know how much value that would really provide.

src/worktree.c
```diff
@@ -0,0 +1,549 @@
...
+}
+
+int git_worktree_unlock(git_worktree *wt)
+{
+	git_buf path = GIT_BUF_INIT;
+
+	assert(wt);
+
+	if (!git_worktree_is_locked(NULL, wt))
+		return 1;
```
ethomson: My inclination would be to just `return 0` even if it's not locked - if callers care, they can call `is_locked` ahead of time...  In part because we're thinking about changing places to `if (error)` instead of `if (error < 0)`...

src/worktree.c
```diff
@@ -0,0 +1,549 @@
...
+
+	if ((flags & GIT_WORKTREE_PRUNE_VALID) == 0 &&
+		git_worktree_validate(wt) == 0)
+	{
+		giterr_set(GITERR_WORKTREE, "Not pruning valid working tree");
+		err = -1;
+		goto out;
+	}
+
+	if ((flags & GIT_WORKTREE_PRUNE_FORCE) == 0)
```
ethomson: This is weird.  😀

include/git2/repository.h
```diff
@@ -371,6 +371,42 @@ GIT_EXTERN(int) git_repository_head_unborn(git_repository *repo);
...
+	GIT_REPOSITORY_ITEM_REFS,
+	GIT_REPOSITORY_ITEM_PACKED_REFS,
+	GIT_REPOSITORY_ITEM_REMOTES,
+	GIT_REPOSITORY_ITEM_CONFIG,
+	GIT_REPOSITORY_ITEM_INFO,
+	GIT_REPOSITORY_ITEM_HOOKS,
+	GIT_REPOSITORY_ITEM_LOGS,
+	GIT_REPOSITORY_ITEM_MODULES,
+	GIT_REPOSITORY_ITEM_WORKTREES
+} git_repository_item;
```
ethomson: I think this should be `git_repository_item_t`.

include/git2/worktree.h
```diff
@@ -0,0 +1,180 @@
...
+ * List names of linked working trees
+ *
+ * The returned list should be released with `git_strarray_free`
+ * when no longer needed.
+ *
+ * @param out pointer to the array of working tree names
+ * @param repo the repo to use when listing working trees
+ * @return 0 or an error code
+ */
+GIT_EXTERN(int) git_worktree_list(git_strarray *out, git_repository *repo);
```
ethomson: So @carlosmn are discussing and we think that `git_worktree_list` should also return the primary repository in this list.  I think extending that thinking out, maybe `git_worktree_lookup` should _also_...  (a `NULL` name, perhaps?)

src/repository.c
```diff
@@ -236,17 +284,34 @@ static int load_workdir(git_repository *repo, git_config *config, git_buf *paren
...
+		char *gitlink = git_worktree__read_link(repo->gitdir, GIT_GITDIR_FILE);
+		if (!gitlink) {
+			error = -1;
+			goto cleanup;
+		}
+
+		git_buf_attach(&worktree, gitlink, 0);
+
+		if ((git_path_dirname_r(&worktree, worktree.ptr)) < 0
+			|| git_path_to_dir(&worktree) < 0) {
```
ethomson: One stylistic comment here:  the `||` should be hanging at the end of the line above it.  There are a bunch of occurrences of this that should be changed - I haven't noted each individually, but I can if that would be helpful.

src/repository.c
```diff
@@ -2290,6 +2438,13 @@ int git_repository_set_head(
 	if (error < 0 && error != GIT_ENOTFOUND)
 		goto cleanup;
 
+	if (ref && current->type == GIT_REF_SYMBOLIC && git__strcmp(current->target.symbolic, ref->name)
+		&& git_reference_is_branch(ref) && git_branch_is_checked_out(ref))
+	{
```
ethomson: Stylistic comment:  the `{` should be pulled up to the line above...  There are a few occurrences of this - I did not point each one out, but I can if it would be helpful.

include/git2/worktree.h
ethomson: (That said, I'm certainly happy to hear arguments if you think that this is not appropriate in our API.)

pks-t:
Will respond to your comments later on. Thanks @carlosmn and @ethomson

include/git2/worktree.h
pks-t: So after thinking a bit more about it, I think I'm opposed to this. There in fact is no function for which it would actually make sense to be called on the main repository. You cannot prune the main worktree and cannot lock or unlock it - so why make it accessible as a worktree?

include/git2/branch.h
pks-t: Well, I agree this interface should be modified. Currently, we simply return `false` even if we were not able to open a worktree, which might not be the right thing to do here. So three possibilities here:

1. Add a flag to `git_branch_is_checked_out`, which indicates whether to honor working trees at all and, if they should be honored, what to do if a working tree is not present.
2. Have some kind of return code here that indicates that some working trees could not be inspected.
3. The ideal thing would be to modify `git_branch_is_head` here to accept the new flag, saying how it should actually behave. While at it and breaking the API, I think it is a bit awkward that the function only accepts a branch without any repository - while the repository is retrievable via the branch itself, I find it a bit restricting.

So I'd not really want to coalesce these two functions together without breaking the API here, as it might not always be the case that a user wants to know if the branch is checked out in _any_ linked repository. Opinions?

include/git2/worktree.h
pks-t: This has been agreed upon in an offline discussion with @ethomson and @carlosmn

include/git2/branch.h
ethomson: Oh, I'm sorry; I didn't realize the difference before.  I'm happy with this as you have it written.

ethomson:
I'm 👍 on this.  Thanks again for all the hard work.

I notice that you have `fixup!` commits; did you want to do a rebase?  Do you want to do the honors of merging this?  😀

src/branch.c
```diff
@@ -149,6 +206,12 @@ int git_branch_delete(git_reference *branch)
 		return -1;
 	}
 
+	if (git_reference_is_branch(branch) && git_branch_is_checked_out(branch)) {
+		giterr_set(GITERR_REFERENCE, "Cannot delete branch '%s' as it is "
+			"the current HEAD of a linked repository.", git_reference_name(branch));
```
pks-t: Will do!

pks-t:
Yeah, wanted to rebase it later on after receiving some final feedback. I'll do so in the next few days and then merge this. :smile: 

Thanks for your reviews!

ethomson:
🎉 

Thanks again @pks-t !!!

PR #3442, opened 2015-09-29
Building for solaris fails - futime

arielb2:
When I try to compile on Solaris fails to find futimes. Futimes is not a standard library therefore not exist for solaris. An alternative is futimens located in "sys/stat.h"

......
$ cmake ..

[ 57%] Building C object CMakeFiles/libgit2_clar.dir/tests/core/pool.c.o
[ 57%] Building C object CMakeFiles/libgit2_clar.dir/tests/core/posix.c.o
/desarrollo/libgit2/tests/core/posix.c: In function 'test_core_posix__utimes':
/desarrollo/libgit2/tests/core/posix.c:129:2: warning: implicit declaration of function 'futimes' [-Wimplicit-function-declaration]
  cl_must_pass(p_futimes(fd, times));

.............

Linking C executable libgit2_clar
Undefined                       first referenced
 symbol                             in file
futimes                             CMakeFiles/libgit2_clar.dir/tests/core/posix.c.o
ld: fatal: symbol referencing errors. No output written to libgit2_clar
CMakeFiles/libgit2_clar.dir/build.make:11662: recipe for target 'libgit2_clar' failed
gmake[2]: *** [libgit2_clar] Error 1
CMakeFiles/Makefile2:95: recipe for target 'CMakeFiles/libgit2_clar.dir/all' failed
gmake[1]: *** [CMakeFiles/libgit2_clar.dir/all] Error 2
Makefile:116: recipe for target 'all' failed
gmake: *** [all] Error 2

You can avoid the error by using the flag to not compile test
$ cmake -DBUILD_CLAR=OFF ..


ethomson:
Thanks!  That is an unfortunate oversight.  It looks like you might have based this branch on a previous one?  Would you mind rebasing just `c04564e` on top of master?

arielb2:
sure, no problem.
done.

ethomson:
It looks like this branch still contains a strange merge and reintroduces some changes from a previous pull request.

Also, our `p_*` functions are meant to handle the various cross-platform unpleasantness.  (For example, the lack of a all the POSIX functions on Windows).  So it would make the most sense to not force callers to understand the differences.  It's easy to implement `futimes` in terms of `futimens`.  So an implementation that doesn't require callers to understand which one exists might be:

    #ifdef HAVE_FUTIMENS
    GIT_INLINE(int) p_futimes(int fd, const struct timeval tv[2])
    {
        struct timespec ts[2];
        ts[0].tv_sec = tv[0].tv_sec;
        ts[0].tv_nsec = tv[0].tv_usec * 1000;
        ts[1].tv_sec = tv[1].tv_sec;
        ts[1].tv_nsec = tv[1].tv_usec * 1000;
        return futimens(fd, ts);
    }
    #else
    # define p_futimes futimes
    #endif

carlosmn:
Is Solaris the only OS which is going to have this peculiarity? Otherwise we should check the standard library for it and set he define accordingly, rather than relying on the OS name.

ethomson:
Indeed.  I didn't realize that CMake could handle this for us, though it's quite obvious in retrospect:

    CHECK_FUNCTION_EXISTS(futimens HAVE_FUTIMENS)

should do the trick here.

arielb2:
It may also affect other OS because is not a standard library. Until now, I only tested it on Solaris.
Just to understand, do you think libgit2 should be used along with futimens instead of futimes?

ethomson:
Thanks for the pull request - this conversation led me to try out the `CHECK_FUNCTION_EXISTS` macro in CMake, which looks like it will suit our needs nicely.  I took your usage of `futimens` usage and implemented with that guard in place, and it looks like it's solved the problem nicely.  Thanks!

See #3446

PR #3444, opened 2015-09-29
Preserve modes from a conflict in `git_index_insert`

ethomson:
When we are on a filesystem where we do not trust the mode (ie, Windows), we presently preserve the modes from a given file when adding entries to the index by looking at the existing index entry for that path.

If we are resolving a conflict, we should also look at *those* entries to try to take their mode.  (If you had a file that was executable and are resolving the conflict on Windows, you would not want to lose the executableness.)

### DIFF
--- a/include/git2/index.h
+++ b/include/git2/index.h
@@ -154,13 +154,27 @@ typedef enum {
 	GIT_INDEX_ADD_CHECK_PATHSPEC = (1u << 2),
 } git_index_add_option_t;
 
-/**
- * Match any index stage.
- *
- * Some index APIs take a stage to match; pass this value to match
- * any entry matching the path regardless of stage.
- */
-#define GIT_INDEX_STAGE_ANY -1
+typedef enum {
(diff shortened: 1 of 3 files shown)


include/git2/index.h
```diff
@@ -162,6 +162,15 @@ typedef enum {
...
 
+/** A normal staged file in the index. */
+#define GIT_INDEX_STAGE_NORMAL 0
+/** The ancestor side of a conflict. */
+#define GIT_INDEX_STAGE_ANCESTOR 1
+/** The "ours" side of a conflict. */
+#define GIT_INDEX_STAGE_OURS 2
+/** The "theirs" side of a conflict. */
+#define GIT_INDEX_STAGE_THEIRS 3
+
```
carlosmn: Can we make this an enum so we the doc parser picks it up?

ethomson: Yes!  Thanks!

PR #3448, opened 2015-10-02
Support setting custom user-agent

carlosmn:
It's not just us who wants fame and notoriety, some tools using the library want it to, and want their name in the User-Agent field; let's let them do that.

We do have to keep the "git/1.0" so that web servers still recognise us as wishing to talk to Git rather than a web server.

This is missing the plain/ssh support, since we don't currently send the agent over there. We should also add this capability, which might happen in this PR or not. It would be a new behaviour rather than a change, and would require us to potentially change the string given by the user, as we'd need a slug rather than a generic string.

### DIFF
--- a/tests/core/useragent.c
+++ b/tests/core/useragent.c
@@ -1,0 +1,11 @@
+#include "clar_libgit2.h"
+#include "global.h"
+
+void test_core_useragent__get(void)
+{
+	const char *custom_name = "super duper git";
+
+	cl_assert_equal_p(NULL, git_libgit2__user_agent());
+	cl_git_pass(git_libgit2_opts(GIT_OPT_SET_USER_AGENT, custom_name));
+	cl_assert_equal_s(custom_name, git_libgit2__user_agent());
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ v0.23 + 1
 * Symlinks are now followed when locking a file, which can be
   necessary when multiple worktrees share a base repository.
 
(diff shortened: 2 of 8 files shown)


ethomson:
Let's free the user-agent in `git_libgit2_shutdown` - otherwise, I think this is good to go.

tests/core/useragent.c
```diff
@@ -0,0 +1,11 @@
+#include "clar_libgit2.h"
+#include "global.h"
+
+void test_core_useragent__get(void)
+{
+	const char *custom_name = "super duper git";
+
+	cl_assert_equal_p(NULL, git_libgit2__user_agent());
+	cl_git_pass(git_libgit2_opts(GIT_OPT_SET_USER_AGENT, custom_name));
+	cl_assert_equal_s(custom_name, git_libgit2__user_agent());
```
vmg: This test feels rather obvious. It would be oh-so-much-better if we could test that the generated HTTP requests do have the custom user agent. ;)

carlosmn: Indeed it would be, but actually adding this to the testing is a large amount of work compared to the impact of this (and the impact of it breaking). We do want a way to test header fields in a more general way, and we will work on that to test higher-impact changes, but not for this largely cosmetic change.

ethomson:
:+1:  Let's commit to getting some broader test coverage here for things like custom headers, the user agent, etc.  In the meantime, this is nice and useful and let's merge it and circle back on the tests.

(How many Microsoft-isms did I fit in to that one sentence?  Hehe.)

PR #3450, opened 2015-10-05
Add a cmake option to require openssl

tkelman:
This could be useful in some cases to ensure at build time that the library gets built with https support, on systems that need openssl to do so.

On a somewhat related note, would maintainers here be receptive to a PR that implements optional support for [MbedTLS](https://tls.mbed.org/) as an alternative backend?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,6 +83,7 @@ ENDIF()
 
 IF (NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 	OPTION( USE_OPENSSL                     "Link with and use openssl library"             ON )
+	OPTION( REQUIRE_OPENSSL                 "Fail configuration if openssl not found"       OFF )
 ENDIF()
 
 # This variable will contain the libraries we need to put into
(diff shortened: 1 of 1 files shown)


carlosmn:
This feels like it's trying to solve the problem at the wrong level. Two options for the same feature just doesn't sit right with me. If CMake/we had a yes/no/auto construct, then it would make sense to fail on 'yes' if it's not found, but as it stands, libgit2 doesn't require OpenSSL to exist.

tkelman:
Right, I'd prefer something cleaner here if it exists in cmake. Libgit2 can be built without openssl, but on linux you end up without https support, and we're moving towards doing everything over https. This is the classic autotools complaint about optional dependencies all over again, a partially functional build can be worse than a build failure. At least with a build failure things fail early and loud with an error message that makes sense. A build without https support gets all the way to building successfully then fails our integration tests with an unsupported protocol error.

For now I can apply a patch that just always adds REQUIRED, but would like to get some mechanism in upstream so I won't always have to be carrying and applying patches.

tkelman:
closing in favor of #3462 (which I hope the maintainers here are in favor of)

tkelman:
I'm finding myself wanting to do the same thing for curl now too, fail at configuration time if it's not found rather than building a partially-functional copy of the library. If my project was using cmake itself then I could easily add this at a higher level, but it currently isn't. Any way we could get some nicer mechanism for this upstreamed so I don't always have to carry a patch that adds `REQUIRED` to these 2 lines?

PR #3453, opened 2015-10-06
CMake: be more explicit with python errors

carlosmn:
There's been a few reports of users not understanding what the python
error means, so spell out the options they have.


arthurschreiber:
:+1:

PR #3454, opened 2015-10-07
CMakeLists: Equalize comparisons to CMAKE_SIZEOF_VOID_P

sschuberth:
Compare CMAKE_SIZEOF_VOID_P as a number, not as a string. Also,
consistently use 32-bit as the fallback to work around combinations of
CMake / compiler versions where CMAKE_SIZEOF_VOID_P is unset. This should
be uncritical as there are no relevant architectures (yet) that are
neither 32- nor 64-bit, and even if so, they are still likely to be able
to run 32-bit code, too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,7 +217,7 @@ IF (WIN32 AND WINHTTP)
 		SET(LIBWINHTTP_PATH "${CMAKE_CURRENT_BINARY_DIR}/deps/winhttp")
 		FILE(MAKE_DIRECTORY ${LIBWINHTTP_PATH})
 
-		IF ("${CMAKE_SIZEOF_VOID_P}" EQUAL "8")
+		IF (CMAKE_SIZEOF_VOID_P EQUAL 8)
 			set(WINHTTP_DEF "${CMAKE_CURRENT_SOURCE_DIR}/deps/winhttp/winhttp64.def")
 		ELSE()
 			set(WINHTTP_DEF "${CMAKE_CURRENT_SOURCE_DIR}/deps/winhttp/winhttp.def")
@@ -538,10 +538,8 @@ FILE(GLOB SRC_GIT2 src/*.c src/*.h src/transports/*.c src/transports/*.h src/xdi
 # Determine architecture of the machine
 IF (CMAKE_SIZEOF_VOID_P EQUAL 8)
 	ADD_DEFINITIONS(-DGIT_ARCH_64)
-ELSEIF (CMAKE_SIZEOF_VOID_P EQUAL 4)
-	ADD_DEFINITIONS(-DGIT_ARCH_32)
 ELSE()
-	MESSAGE(FATAL_ERROR "Unsupported architecture")
+	ADD_DEFINITIONS(-DGIT_ARCH_32)
 ENDIF()
 
 # Compile and link libgit2


carlosmn:
The error when we are unable to figure out which architecture we're building on should serve as a way to know that we don't know what's happening. If the pointer size is unknown here, are we going to be able to detect/trust anything in the headers which try to figure out the equivalent bits in the headers? Do we know in that case that we're going to be able to build something which won't crash immediately?

Why assume it's a 32 bit system when we don't know? Why not 64 bits?

sschuberth:
> If the pointer size is unknown here, are we going to be able to detect/trust anything in the headers which try to figure out the equivalent bits in the headers?

I would say yes, because headers are parsed by the compiler, not CMake, so even if CMake is unable to determine the pointer size, the compiler most certainly is.

> Do we know in that case that we're going to be able to build something which won't crash immediately?

That case can and should be covered by tests, no?

> Why assume it's a 32 bit system when we don't know? Why not 64 bits?

As mentioned in the commit message, because 32-bit code runs on 64-bit systems, but not the other way around.

ethomson-test1:
I don't think that we should assume that we're on a 32 bit processor, because we use the setting of `GIT_ARCH_64` / `GIT_ARCH_32` to inform us as to the size of a `size_t`.  This is later used for atomic operations - if we fell back to 32 bit handling, then we would end up using a 32 bit atomic operation on a 64 bit data value, which would be bad.  That's why we're quite explicit about this.

Is this causing you problems in practice?  If so, I realize that I did a different strategy to detect word size (and avoids cmake) in https://github.com/libgit2/libgit2/blob/master/src/integer.h#L58 .  Maybe we should unify the two.

ethomson:
I don't think that we should assume that we're on a 32 bit processor, because we use the setting of GIT_ARCH_64 / GIT_ARCH_32 to inform us as to the size of a size_t. This is later used for atomic operations - if we fell back to 32 bit handling, then we would end up using a 32 bit atomic operation on a 64 bit data value, which would be bad. That's why we're quite explicit about this.

Is this causing you problems in practice? If so, I realize that I did a different strategy to detect word size (and avoids cmake) in https://github.com/libgit2/libgit2/blob/master/src/integer.h#L58 . Maybe we should unify the two.

sschuberth:
> I don't think that we should assume that we're on a 32 bit processor

But that assessment should then apply to [this existing line](https://github.com/libgit2/libgit2/blob/master/CMakeLists.txt#L223), too, or?

> Is this causing you problems in practice?

It was for me, hence the change. I was trying to build [licensee](https://github.com/benbalter/licensee), which depends on [rugged](https://github.com/libgit2/rugged), from my MSYS2 environment (with CMake 3.1.3 coming from the native installer, not from `pacman`) and was seeing the "Unsupported architecture" error.

ethomson:
> But that assessment should then apply to this existing line, too, or? That line was the reason why I thought it's accepted to fall back to 32-bit.

I think that the assumption there was that it was *probably* safe to use the 32 bit WinHTTP definitions on a 64 bit machine.  And this probably predates our vigilance about knowing the word size of the platform so that we could have a 64 bit atomic operation.  More on that in a second.

I suspect that if you wanted to work around this, that we could move this test out of CMake and use a preprocessor macro instead, testing `SIZE_MAX` explicitly.

However, it makes me nervous that we can't figure out your word size at compile time.  For example - I haven't tested the assumption that you could really use the 32 bit WinHTTP safely, or even reviewed the code.  It seems pretty likely that we might cast a `size_t` to some size type.  And if you're using the 32 bit WinHTTP library on a 64 bit machine, perhaps we'll end up overflowing a 32 bit int.

So I'm a little nervous about making this change.

I'd be curious to know _why_ cmake couldn't determine your platform, and if we could do a better check in CMake.  Another alternative is to explicitly specify your word size to cmake, and it will not try to determine that itself.  I'm not sure if that would help here, or not, given that it looks like another build process is driving you here?

sschuberth:
> I'd be curious to know *why* cmake couldn't determine your platform, and if we could do a better check in CMake.

Back when I encountered the issue I did not have the time to investigate further, and now I cannot reproduce it anymore. So I've split the (anyway unrelated) improvement to compare `CMAKE_SIZEOF_VOID_P` as a number into the separate PR #3526, and will close this one.

PR #3455, opened 2015-10-07
Examples: network/fetch.c transfer_progress_cb - should return a value

Eun:
`transfer_progress_cb` should return a value, otherwise the transfer gets aborted sometimes.


carlosmn:
Indeed. This kind of thing should be a compilation error, but oh well.

Issue #3456, opened 2015-10-07
Completion callback in git_remote_callbacks


leoyanggit:
```
/**
 * Argument to the completion callback which tells it which operation
 * finished.
 */
typedef enum git_remote_completion_type {
    GIT_REMOTE_COMPLETION_DOWNLOAD,
    GIT_REMOTE_COMPLETION_INDEXING,
    GIT_REMOTE_COMPLETION_ERROR,
} 
```
The callback takes `git_remote_completion_type` as an argument. Currently the callback is not used. Actually I wanted to use this callback to tell the completed operations. Specifically I wanted to know when a fetch negotiation is finished. To fully use this callback my questions are that

1) Should this callback apply to both fetch and push or fetch only?
2) To tell the completion of negotiation should something like `GIT_REMOTE_COMPLETION_NEGOTIATION` be added?
3) Should `GIT_REMOTE_COMPLETION_INDEXING` mean the completion of writepack?
4) Should `GIT_REMOTE_COMPLETION_ERROR` be thrown when any error occurs at any point?

Thanks!

serg-vinnie:
lol, probably this won't be fixed in nearest time :D

PR #3461, opened 2015-10-08
Add git_remote_download_option_t in git_remote_download

leoyanggit:
Just as `git_push_negotiation`.

### DIFF
--- a/include/git2/remote.h
+++ b/include/git2/remote.h
@@ -511,6 +511,21 @@ typedef enum {
 	GIT_REMOTE_DOWNLOAD_TAGS_ALL,
 } git_remote_autotag_option_t;
 
+typedef enum {
+	/**
+         * Normal download: negotiate and download pack.
+	 */
+	GIT_REMOTE_DOWNLOAD_NORMAL = 0,
+	/**
+	 * Negotiate only.
+	 */
+	GIT_REMOTE_DOWNLOAD_NEGOTIATE_ONLY,
+	/**
+	 * Download pack only assuming already connected and negotiated.
+	 */
+	GIT_REMOTE_DOWNLOAD_DOWNLOAD_PACK_ONLY,
+} git_remote_download_option_t;
+
 /**
  * Fetch options structure.
  *
@@ -546,6 +561,7 @@ typedef struct {
 	 * The default is to auto-follow tags.
 	 */
 	git_remote_autotag_option_t download_tags;
+
 } git_fetch_options;
 
 #define GIT_FETCH_OPTIONS_VERSION 1
@@ -619,7 +635,7 @@ GIT_EXTERN(int) git_push_init_options(
  * @param opts the options to use for this fetch
  * @return 0 or an error code
  */
- GIT_EXTERN(int) git_remote_download(git_remote *remote, const git_strarray *refspecs, const git_fetch_options *opts);
+ GIT_EXTERN(int) git_remote_download(git_remote *remote, const git_strarray *refspecs, const git_fetch_options *opts, git_remote_download_option_t dl_opt);
 
 /**
  * Create a packfile and send it to the server
--- a/src/remote.c
+++ b/src/remote.c
@@ -878,7 +878,7 @@ static int ls_to_vector(git_vector *out, git_remote *remote)
 	return 0;
 }
 
-int git_remote_download(git_remote *remote, const git_strarray *refspecs, const git_fetch_options *opts)
+int git_remote_download(git_remote *remote, const git_strarray *refspecs, const git_fetch_options *opts, git_remote_download_option_t dl_opt)
 {
 	int error = -1;
 	size_t i;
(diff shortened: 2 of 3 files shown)


carlosmn:
What's the use-case for this? The push negotiation callback is there to know what we're going to send to the server, which you can use for something like git-lfs where you fill in the actual contents of files on the server.

When we're fetching, we have no idea what we're going to receive, so it doesn't work like the push side; there's nothing we can do about what we receive until the `update_tips` call, where you know more accurately what's getting updated (the list you're using in this code would be missing any tags which come from auto-following).

leoyanggit:
My use case is to know when a fetch negotiation is finished. Once the negotiation is successful I need to switch transportation from one protocol to another protocol underlying, but from the git point of view the connection is still the same. The reason why I need to switch protocol is that the negotiation part fits more in one of my protocols while the pack file part fits more in another one. In my case I don't need to know the remote heads. I added it just for kind of symmetry purpose. I'm OK to remove remote heads and it's length. 

Or for my case should I implement the remote_completion callback I mentioned in #3456 ?

leoyanggit:
I have another approach in mind. In stead of adding a callback I add an option to `git_remote_download` which is a public API as the last parameter: `GIT_EXTERN(int) git_remote_download(git_remote *remote, const git_strarray *refspecs, const git_fetch_options *opts, git_remote_download_option_t dl_opt);`.
```
typedef enum {
        /**
         * Normal download: negotiate and download pack.
         */
        GIT_REMOTE_DOWNLOAD_NORMAL = 0,
        /**
         * Negotiate only.
         */
        GIT_REMOTE_DOWNLOAD_NEGOTIATE_ONLY,
        /**
         * Download pack only assuming already connected and negotiated.
         */
        GIT_REMOTE_DOWNLOAD_DOWNLOAD_PACK_ONLY,
} git_remote_download_option_t;
```
Inside `git_remote_download()` I do things based on the `dl_option`. This way it is more flexible to the API user who could delay the pack file download to whenever he/she wants. Thoughts?

carlosmn:
If you have your own protocol, there should already be enough information there. The `git_transport` interface has different methods in it for negotiating the objects to download and downloading the packfile so your custom protocol should be using that in order to know when to mode-switch.

leoyanggit:
I don't have my own protocol, instead, I'm using the smart protocol. I do have my own sub-transport of the smart transport. At the sub-transport level I don't know when a negotiation is done unless I look into the data sending to/receiving from the wire which is a layer violation IMHO.

carlosmn:
If you're implementing a smart subtransport, then you're at the level of TCP or HTTP and you don't care about what's happening at the Git layer, only how to move bytes within the particular protocol.

leoyanggit:
Yes I'm at that level. However I have a bunch of protocols that I can use for a git fetch session. For negotiation part the smart protocol is interactive(or I call it chatty), one of my protocols is good to do this kind of byte transfer. For pack file transfer it's actually bulk data transfer I have another protocol to do this better. I need to switch the chatty transport to bulk file transport because the former can't run a relatively long time in all the cases. So for me the chatty part should happen as early and quickly as possible and then the left part should go with another bulk data transportation which can run a longer time in call the circumstances. It's like that I have something like a limited time of TCP transport and a unlimited HTTP transport. My last patch is to simply give more granular control of `git_remote_download_pack`. Someone else with other use cases may take advantage of it.

If we'd like to keep the API signature of `git_remote_download_pack`, I can add new APIs to do negotiation and pack file download separately.

carlosmn:
Having a callback at the `git_remote` level to inform the transport on what to do breaks all the layering and abstraction we're bulding, so it's not something we're going to have in the library.

What you are describing is not he Git Smart Protocol but something else, and you can create your own transport so it has enough information. If it looks a lot like the one which exists, then you can re-use the one we have.

leoyanggit:
I don't want to go with the callback approach as I mentioned before in this thread. How about https://github.com/leoyanggit/libgit2/commit/26a6414bbc86a161e63e62a4c2ecb3c8ad6771c9 which is also mentioned in this thread?

carlosmn:
Making `git_remote_download()` do two different things based off a flag is not an API we're going to support. There used to be a function which would perform just the negotiation which could be brought back, and teach `git_remote_download()` know to skip that part. But you'd also need to figure out how that fits into the push side, which didn't exist when that function was there.

But at the end of the day, you're still coupling what your network transports do to the individual steps done at a higher level, and you're going to be breaking layering in many ways regardless, so I would recommend to accept that and put the necessary code in your version of the library to perform this intra-layer communicatoin. The changes to the network communication as described depend on details which might change at any time by internal refactoring.

leoyanggit:
Yeah I understand your concern here. So what was that function which did the negotiation part? I'd like to have a look.

ethomson:
This is quite outdated, and I'm closing this to tidy up the outstanding pull requests.  Thanks for working on this!

Issue #3458, opened 2015-10-08
git_commit_summary can return newlines which leads to invalid reflog messages


srajko:
Newlines after spaces present a problem to `git_commit_summary`.  For example, `git_commit_summary("Newline after space \nis removed")` returns:
```
"Newline after space  \nis removed"
```

Since `git_reference__update_for_commit` uses `git_commit_summary` to construct the reflog message, the returned newline in the reflog message breaks `git_reflog_read` (it will return an error because it finds the newline, expects a new entry, and then can't parse the oid).

I'd be happy to submit a PR if fixing `git_commit_summary` is indeed the right thing to do here.

ethomson:
Yes!  It certainly looks like `git_commit_summary` is deficient here - a pull request would be great!

carlosmn:
It doesn't seem like `git_commit_summary()` is deficient here, but rather that we're not handling it correctly. The summary is a paragraph and can thus contain many newlines. What does git do in this case? Does it reflow the text to make it a single line or does it cut at the first newline?

srajko:
`git_commit_summary` does process the first paragraph, however it tries to remove newlines and return a single line result.  Here are the original test cases:

```
	assert_commit_summary("One-liner with no trailing newline", "One-liner with no trailing newline");
	assert_commit_summary("One-liner with trailing newline", "One-liner with trailing newline\n");
	assert_commit_summary("Trimmed leading&trailing newlines", "\n\nTrimmed leading&trailing newlines\n\n");
	assert_commit_summary("First paragraph only", "\nFirst paragraph only\n\n(There are more!)");
	assert_commit_summary("First paragraph with  unwrapped trailing\tlines", "\nFirst paragraph\nwith  unwrapped\ntrailing\tlines\n\n(Yes, unwrapped!)");
	assert_commit_summary("\tLeading \ttabs", "\tLeading\n\ttabs\n\nis preserved");
	assert_commit_summary(" Leading  Spaces", " Leading\n Spaces\n\nare preserved");
	assert_commit_summary("Trailing tabs\tare removed", "Trailing tabs\tare removed\t\t");
	assert_commit_summary("Trailing spaces  are removed", "Trailing spaces  are removed  ");
	assert_commit_summary("Trailing tabs", "Trailing tabs\t\n\nare removed");
	assert_commit_summary("Trailing spaces", "Trailing spaces \n\nare removed");
	assert_commit_summary("", "");
	assert_commit_summary("", " ");
[...]

PR #3463, opened 2015-10-09
Fix issues related to newlines in reflog messages

srajko:
This addresses both issues mentioned in #3458.  git_reflog_read should no longer segfault when reading a reflog that has a newline in the middle of a reflog message, and git_commit_summary no longer returns newlines when they occur following whitespace.

Please let me know if you'd like this tweaked in any way.


src/commit.c
```diff
@@ -436,17 +436,23 @@ const char *git_commit_summary(git_commit *commit)
 
 	if (!commit->summary) {
 		for (msg = git_commit_message(commit), space = NULL; *msg; ++msg) {
-			if (msg[0] == '\n' && (!msg[1] || msg[1] == '\n'))
+			char next_character = msg[0];
+			// stop processing at the end of the first paragraph
```
carlosmn: We use C comments

tests/refs/reflog/reflog.c
```diff
@@ -154,6 +154,18 @@ void test_refs_reflog_reflog__reading_the_reflog_from_a_reference_with_no_log_re
...
 
+void test_refs_reflog_reflog__reading_a_reflog_with_invalid_format_returns_error(void)
+{
+	git_reflog *reflog;
+	git_error *error;
+
+	cl_git_fail(git_reflog_read(&reflog, g_repo, "refs/heads/extra-newline"));
+
+	error = giterr_last();
+	cl_assert(error);
```
carlosmn: This isn't testing for error. Error codes are negative numbers.

carlosmn: If you don't clear the error state before calling `git_reflog_read()`, you can't be sure you're looking at what `git_reflog_read()` gave you.

carlosmn:
In #3458 you mention that git cuts off the message at the first line feed, which we still don't do here. Making `git_commit_summary`'s reflow better doesn't make us behave more like git in this situation.

tests/commit/commit.c
```diff
@@ -76,6 +76,8 @@ void test_commit_commit__summary(void)
 	assert_commit_summary("Trailing spaces  are removed", "Trailing spaces  are removed  ");
 	assert_commit_summary("Trailing tabs", "Trailing tabs\t\n\nare removed");
 	assert_commit_summary("Trailing spaces", "Trailing spaces \n\nare removed");
+	assert_commit_summary("Newline is removed", "Newline\nis removed");
+	assert_commit_summary("Newline after space  is removed", "Newline after space \nis removed");
```
carlosmn: If we're going to reflow the text, shouldn't we collapse the whitespace here?

srajko:
Thanks for taking a look, I will incorporate your feedback.

On collapsing the whitespace, I don't really have an opinion so I'm happy to have it do whatever you think is best.  On the paragraph level, `git_commit_summary` preserves leading space and cuts off trailing space - so we could do the same around the single newline for consistency, or it might make more sense to just collapse all whitespace around the newline (and including the newline) into a single space.

srajko:
@carlosmn I started changing `git_commit_summary` so it collapses whitespace around a newline into a single space, and then noticed that the change breaks existing tests:

```
assert_commit_summary("\tLeading \ttabs", "\tLeading\n\ttabs\n\nis preserved");
assert_commit_summary(" Leading  Spaces", " Leading\n Spaces\n\nare preserved"); /* note two spaces between g and S */
```

With the change it would return `"\tLeading tabs"`, and `" Leading Spaces"`

I'm a little weary of breaking existing functionality (with the original change I was just trying to eliminate breaking reflogs with minimal changes in behavior).  Can you please clarify what behavior you would like to see here?

srajko:
FWIW, my inclination is to keep this PR as is, since it fixes a segfault, and avoids a scenario in which reflogs are written incorrectly, with minimal changes in behavior.

We can then consider doing more significant changes to `git_commit_summary`, whether it's just collapsing whitespace or more drastic changes that bring it more in line with git.

ethomson:
IIRC, the existing `git_commit_summary` was meant to emulate the behavior of git when it uses a commit message summary for display purposes (warts, strange behavior and all).

In particular, it was meant to match the results that `git revert` used in its message.  Upon some digging, there are at least three(!!!) different ways that git will summarize a commit message.

With the `"Newline after space \nis removed"` example, `git revert` will produce a commit message of `Revert "Newline after space "`.

However, `git log --oneline` will show different behavior, and yet a different summary is written to the reflog:

```
% git log --oneline
f1ace23 Newline after space is removed
9a294f9 Newline is removed
83b7893         Leading         tabs
72f0988  Leading  Spaces
2f246ab         Leading         tabs
```

```
% git reflog
f1ace23 HEAD@{0}: commit: Newline after space
9a294f9 HEAD@{1}: commit: Newline
83b7893 HEAD@{2}: commit: Leading
72f0988 HEAD@{3}: commit: Leading
2f246ab HEAD@{4}: commit: Leading tabs
```

Bluh.  So...  yeah, I think that we should probably just give up and try to use our own, sane, summary message that *we* like, and use it for summaries across the board.  This means giving up on matching git's behavior and its weird oddities.

In that case, I think that collapsing whitespace would be nice - those previous tests be damned.

One thing that we should do, however is to make sure that we don't *need* to summarize our commits in the reflog identically to git.  I say this because it parses the reflog occasionally to determine the repository's state.  (Bluh again).  But we're *already* broken here, if that's true, so I don't think that it needs to be part of this PR.

srajko:
Sounds good - I added in collapsing whitespace around newlines.

srajko:
Would it be helpful to separate out the segfault fix into a separate PR?  I made one for NodeGit's libgit2 (https://github.com/nodegit/libgit2/pull/2), against `maint/v0.23`.  I can do the same here if you'd like.

srajko:
@carlosmn @ethomson Is there anything I can do to help move this PR along?  I'm not sure if I've addressed all the concerns or if there are still any outstanding issues to tackle.

ethomson:
Sorry for the delay - I was going to merge this manually over the weekend but left my computer at home on a last-minute whim.

If you would be so kind as to revise this into three separate commits, it would make cherry picking this into a maint branch much easier:

One containing the typo fix, one containing the actual buffer overflow fix, and a third with the fixes for computing the new message.

Thanks again for the good work here.

srajko:
Not a problem - will do.  Thanks!

ethomson:
Looks like this has been split up into three separate pull requests, so I'm closing this.

(I apologize if I was unclear - I was thinking just to have three separate commits - pushing those back up into this PR itself would have been fine.)

Either way - thanks for the fine work here.

PR #3465, opened 2015-10-13
stream: allow registering a user-provided TLS constructor

carlosmn:
This allows the application to use their own TLS stream, regardless of
the capabilities of libgit2 itself.

---
The naming feels a bit all over the place, so suggestions there would be great.

### DIFF
--- a/include/git2/sys/stream.h
+++ b/include/git2/sys/stream.h
@@ -39,6 +39,19 @@ typedef struct git_stream {
 	void (*free)(struct git_stream *);
 } git_stream;
 
+typedef int (*git_stream_cb)(git_stream **out, const char *host, const char *port);
+
+/**
+ * Register a TLS stream constructor for the library to use
+ *
+ * If a constructor is already set, it will be overwritten. Pass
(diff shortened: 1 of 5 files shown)


tkelman:
Will this make it into 0.24.0?

include/git2/sys/stream.h
```diff
@@ -39,6 +39,19 @@ typedef struct git_stream {
 	void (*free)(struct git_stream *);
 } git_stream;
 
+typedef int (*git_stream_ctor)(git_stream **out, const char *host, const char *port);
```
ethomson: I think that this might slightly better match some of the existing type names if it were `git_stream_cb`.  (See for example `git_transport_cb` which has a similar purpose.)

ethomson:
Minor nit on the name, otherwise :+1:

include/git2/sys/stream.h
carlosmn: Done.

tkelman:
Lovely, thanks!

PR #3467, opened 2015-10-14
Create LICENSE

dijonkitchen:
For GitHub Licenses API: https://developer.github.com/v3/licenses/

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,960 @@
+ libgit2 is Copyright (C) the libgit2 contributors,
+ unless otherwise stated. See the AUTHORS file for details.
+
+ Note that the only valid version of the GPL as far as this project
+ is concerned is _this_ particular version of the license (ie v2, not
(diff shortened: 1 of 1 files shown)


ethomson:
We already have a licenses file, as `COPYING`.  We do not want to have two different files with the same content, they'll inevitably get out of sync.

Does the licenses API not find files named `COPYING`?

dijonkitchen:
My apologies. It does: https://github.com/benbalter/licensee#what-it-looks-at

vmg:
> Does the licenses API not find files named COPYING?
 
It does. 

@dijonkitchen: `libgit2` is missing license information from our Licenses API because we have a very specific, custom license: GPLv2 + a linking exception. The license detection algorithms we're using cannot currently detect or classify our specific license, that's why the information is missing. Adding a `LICENSE` file will not fix the issue; it will still fail to detect the custom license.

PR #3468, opened 2015-10-14
Fix pathological performance in ODB lookups

vmg:
As discussed in chat -- the commit messages have all the details.

cc @carlosmn @ethomson 
cc @scottjg who spotted this

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -620,22 +620,17 @@ void git_odb_free(git_odb *db)
 	GIT_REFCOUNT_DEC(db, odb_free);
 }
 
-int git_odb_exists(git_odb *db, const git_oid *id)
+static int odb_exists_1(git_odb *db, const git_oid *id, bool only_refreshed)
 {
-	git_odb_object *object;
 	size_t i;
 	bool found = false;
-
-	assert(db && id);
-
-	if ((object = git_cache_get_raw(odb_cache(db), id)) != NULL) {
-		git_odb_object_free(object);
-		return (int)true;
-	}
 
 	for (i = 0; i < db->backends.length && !found; ++i) {
 		backend_internal *internal = git_vector_get(&db->backends, i);
 		git_odb_backend *b = internal->backend;
+
+		if (only_refreshed && !b->refresh)
+			continue;
 
 		if (b->exists != NULL)
 			found = (bool)b->exists(b, id);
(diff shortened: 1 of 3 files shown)


carlosmn:
Cool. It feels cleaner to have the refresh come from outside the backend.

On the priority bits, I wonder if we should prioritize going through all the packed backends before going through the loose ones. They cost going through some in-memory structures rather than asking the OS to go to the filesystem. That is, remove the decision point of something being an alternate, or leave it for when we match priorities, which I think is what @arthurschreiber mentioned.

vmg:
> On the priority bits, I wonder if we should prioritize going through all the packed backends before going through the loose ones.

This makes sense to me. I'll fix up the commits.

vmg:
@carlosmn: Done.

arthurschreiber:
:+1:

mhagger:
You might be interested in Git commit https://github.com/git/git/commit/0eeb077be7d526635995e17808927c94c70be665, which introduces an extra `quick` flag for object lookup to be used when a false negative is acceptable. When this flag is on, there is no attempt to refresh and repeat the lookup. This apparently makes a huge difference for some operations on high-latency filesystems like NFS.

The `quick` feature doesn't replace the need for this PR, but it is in the same performance area so I wanted to point it out.

vmg:
Thanks for the pointer @mhagger. Looks like that only applies to `exist` checks, which we don't use very often in the library. It would be nice to have eventually though.

ethomson:
Thanks for this @vmg - it looks great.

vmg:
:smile:

peff:
FWIW, git does this by starting the pack search at the last pack we found anything in. I think that is a better solution for two reasons:

  - it does not arbitrarily prefer alternates to the local repo (so no matter which has more objects, the distribution will be weighted in the correct direction).

  - it also improves the case of having many packs (since lookups tend to have locality to packs, and also because it weights the distribution towards the big packs).

carlosmn:
We do have (some of) this logic in libgit2, but it's restricted to each backend. In cases where most of the objects are in the alternate, we would choose the right packfile immediately most of the time, but we would go wrong for objects which are only available locally more often, since we'd exhaust the alternate packfiles before looking into the first one.

Storing in which backend we found the last object should be a pretty easy addition which would give us that last bit of overall performance.

But the order should be pretty minor in what this PR achieves. The big win is exhausting our in-memory indices before we ask the operating system for anything.

Issue #3459, opened 2015-10-08
git_repository_init_ext with GIT_REPOSITORY_INIT_EXTERNAL_TEMPLATE does not work if path contains .git


csware:
Set

    [init]
        templatedir = C:/Users/something/.git_template

in `.gitconfig` and call `git_repository_init_ext` with `GIT_REPOSITORY_INIT_EXTERNAL_TEMPLATE` set and notice that no template is applied (fails here: https://github.com/libgit2/libgit2/blob/master/src/fileops.c#L903).

Using `git init` works as expected.

Cross-ref: https://tortoisegit.org/issue/2602

/CC @ethomson

ethomson:
Fixed via #3565

PR #3472, opened 2015-10-16
Fix git_repository_init_ext with template dir which contains a path starting with dot

csware:
This fixes issue #3459.

/CC @ethomson

### DIFF
--- a/tests/repo/init.c
+++ b/tests/repo/init.c
@@ -503,7 +503,7 @@ static void assert_mode_seems_okay(
 
 static const char *template_sandbox(const char *name)
 {
-	git_buf hooks_path = GIT_BUF_INIT, link_path = GIT_BUF_INIT;
+	git_buf hooks_path = GIT_BUF_INIT, link_path = GIT_BUF_INIT, dotfile_path = GIT_BUF_INIT;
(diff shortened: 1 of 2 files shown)


carlosmn:
How does this fix #3459 when the tests in this PR are for something different? In the issue you mention, the problem is libgit2 not copying a template directory when it starts with a dot. The tests here make sure that we copy dotfiles *inside* a template path.

Even if the fix the same, this should test that the issue is fixed, rather than assume that it's also fixed if we include the files under it.

tests/repo/init.c
```diff
@@ -503,7 +503,7 @@ static void assert_mode_seems_okay(
 
 static const char *template_sandbox(const char *name)
 {
-	git_buf hooks_path = GIT_BUF_INIT, link_path = GIT_BUF_INIT;
+	git_buf hooks_path = GIT_BUF_INIT, link_path = GIT_BUF_INIT, dotfile_path = GIT_BUF_INIT;
```
ethomson: `dotfile_path` was never freed.

ethomson:
Thanks for this.  I finally had some time and the inclination to write a unit test for #3459.  I then cherry-picked this change as the fix.  See #3565.

PR #3475, opened 2015-10-20
config: add a ProgramData level

carlosmn:
This is where portable git stores the global configuration which we can
use to adhere to it even though git isn't quite installed on the system.

---

The test isn't amazing, but it gets the job done. Locally I had to add my user as allowed to touch the file, so let's see how this goes on Appveyor.

/cc @ethomson

### DIFF
--- a/include/git2/config.h
+++ b/include/git2/config.h
@@ -29,25 +29,28 @@ GIT_BEGIN_DECL
  * priority levels as well.
  */
 typedef enum {
+	/** System-wide on Windows, for compatibility with portable git */
+	GIT_CONFIG_LEVEL_PROGRAMDATA = 1,
+
 	/** System-wide configuration file; /etc/gitconfig on Linux systems */
-	GIT_CONFIG_LEVEL_SYSTEM = 1,
+	GIT_CONFIG_LEVEL_SYSTEM = 2,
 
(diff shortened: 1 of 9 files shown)


ethomson:
I have good news for you.  It turns out that `%PROGRAMDATA%` is a magical wasteland of terrible.  It's *not* actually world-writable, as was claimed in the first PR, which makes a lot of sense.  Instead, there's some magical redirection that goes on if you don't have permission to write to it, where your file(s) are magically shunted off to a place like `%LOCALAPPDATA%\VirtualStore\ProgramData\...`.

I don't know what subsequent reads of `%PROGRAMDATA%\...` paths do - maybe they look in your per-user `%PROGRAMDATA%` first ...?  

This is really abysmal and I'm sorry that this crime scene has some of my fingerprints on it.

carlosmn:
On my dev box I have `C:\ProgramData\Git\config` which I can see with explorer, and before adding me as having full control over the file, `git_filebuf` would fail to rename the lockfile to the final location with a permission error, so I don't know if G4W does something special or what, but it does seem to exist as a regular file (at least it behaves like it). The changes made by the test do persist as well (I can open it with notepad and see the new variable).

I'm not sure if that means anything, but it looks like the test will have to make sure to create the directory (I was hoping whatever git installation it has would work the same way as on my box, which evidently it does not).

PR #3478, opened 2015-10-21
signature: Strip crud

vmg:
So, Git does this: https://github.com/github/git/blob/github/ident.c#L137-L155

And we don't. So I feel we should be doing this. As to do the same thing that Git does. Even though whatever Git does is weird. But this is causing commits to differ from my experiments. So it would make my life easier if we did the same thing that Git does.

cc @carlosmn @ethomson

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -1193,13 +1193,13 @@ static int index_no_dups(void **old, void *new)
 }
 
 static void index_existing_and_best(
-	const git_index_entry **existing,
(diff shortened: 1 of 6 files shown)


vmg:
Also included two commits to fix warnings. :o

carlosmn:
Makes sense, I guess. If this is considered canonical normalisation, I guess we should do it.

PR #3479, opened 2015-10-21
api: be explicit about our C linkage

ethomson:
C++ users expect to not have to worry too much about linkages.  Let's me explicit about ours.

### DIFF
--- a/include/git2/common.h
+++ b/include/git2/common.h
@@ -10,12 +10,6 @@
 #include <time.h>
 #include <stdlib.h>
 
-#ifdef _MSC_VER
-#	include "inttypes.h"
-#else
-#	include <inttypes.h>
-#endif
-
 #ifdef __cplusplus
 # define GIT_BEGIN_DECL extern "C" {
 # define GIT_END_DECL	}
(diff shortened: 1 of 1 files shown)


carlosmn:
We have this in the start and end macros for each file, so this exterior one shouldn't be necessary (ideally).

ethomson:
Ah, yes, I forgot that's what `GIT_BEGIN_DECL` did.

We didn't do this for `inttypes.h`, since it's not really ours...  I wonder if there are others.

pks-t:
» find include -type f -exec grep -L GIT_BEGIN_DECL '{}' \;
include/git2.h
include/git2/stdint.h
include/git2/inttypes.h
include/git2/version.h

So yes, there are at least some.

ethomson:
`git2.h` doesn't include any declarations directly.  Nor does `version.h`.

`stdint.h` is only included by `inttypes.h`.

So this guard is really only needed around the `inttypes.h` inclusion.

carlosmn:
The system-provided headers do their own extern definition, it's completely outside of libgit2's responsibility to make the system's headers behave nicely with C++ in either case.

PR #3480, opened 2015-10-22
Nanoseconds in the index: ignore for diffing

ethomson:
We should be ignoring nanoseconds in the index for diffing and status computation.  Although we load the nanoseconds out of the index, and preserve literal index entries (yay!), when we read file times from disk via `stat`, we ignore the nanosecond precision and simply truncate it.

If a git client had written an index with nanoseconds, and we were to compare the nanoseconds field from that index to the (empty) nanoseconds field from our `stat`, then we would always think the workdir was dirty and rescan it.

Always ignore the nanos in the index.  A better improvement will come in #3170, but this should stop the bleeding.


carlosmn:
Yep. We really shouldn't be comparing nsecs when we don't take them into account.

Issue #3460, opened 2015-10-08
git__malloc is undefined in mac osx


implausible:
in src/xdiff/xdiff.h libgit2 includes "util.h", but should actually include ../util.h. This causes git__malloc to not be defined and crashes in mac osx and linux during merges.

ethomson:
That would cause a compilation error.

Can you actually paste some error that you're seeing?

ethomson:
Just an update.  This looks like it's something in your custom build process, and nodegit is not using CMake, is that correct?

If so, can you provide a pull request that works for your build process, and for ours?

ethomson:
Fixed by #3481

PR #3481, opened 2015-10-22
xdiff: reference util.h in parent directory

ethomson:
Although CMake will correctly configure include directories for us,
some people may use their own build system, and we should reference
`util.h` based on where it actually lives.


ethomson:
Fixes #3460

PR #3482, opened 2015-10-22
merge: add GIT_MERGE_TREE_FAIL_ON_CONFLICT

ethomson:
Provide a new merge option, GIT_MERGE_TREE_FAIL_ON_CONFLICT, which
will stop on the first conflict and fail the merge operation with
GIT_EMERGECONFLICT.

### DIFF
--- a/include/git2/errors.h
+++ b/include/git2/errors.h
@@ -49,6 +49,7 @@ typedef enum {
 	GIT_EINVALID        = -21,      /**< Invalid operation or input */
 	GIT_EUNCOMMITTED    = -22,	/**< Uncommitted changes in index prevented operation */
 	GIT_EDIRECTORY      = -23,      /**< The operation is not valid for a directory */
+	GIT_EMERGECONFLICT  = -24,	/**< A merge conflict exists and cannot continue */
 
 	GIT_PASSTHROUGH     = -30,	/**< Internal only */
 	GIT_ITEROVER        = -31,	/**< Signals end of iteration with iterator */
--- a/include/git2/merge.h
+++ b/include/git2/merge.h
@@ -72,6 +72,13 @@ typedef enum {
 	 * the ability to merge between a modified and renamed file.
 	 */
 	GIT_MERGE_TREE_FIND_RENAMES = (1 << 0),
+
+	/**
+	 * If a conflict occurs, exit immediately instead of attempting to
+	 * continue resolving conflicts.  The merge operation will fail with
+	 * GIT_EMERGECONFLICT and no index will be returned.
+	 */
+	GIT_MERGE_TREE_FAIL_ON_CONFLICT = (1 << 1),
 } git_merge_tree_flag_t;
 
 /**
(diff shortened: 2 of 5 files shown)


ethomson:
I had initially implemented this as an integer count - if we hit `x` conflicts, then we would fail.  But that seemed really dumb and everybody would either set this to `0` or `1`, so I settled on this instead.  Let me know if you disagree.

/cc @vmg

include/git2/errors.h
```diff
@@ -49,6 +49,7 @@ typedef enum {
 	GIT_EINVALID        = -21,      /**< Invalid operation or input */
 	GIT_EUNCOMMITTED    = -22,	/**< Uncommitted changes in index prevented operation */
 	GIT_EDIRECTORY      = -23,      /**< The operation is not valid for a directory */
+	GIT_EMERGECONFLICT	= -24,	/**< A merge conflict exists and cannot continue */
```
carlosmn: Odd alignment, did you use a tab instead of spaces here?

include/git2/merge.h
```diff
@@ -72,6 +72,13 @@ typedef enum {
 	 * the ability to merge between a modified and renamed file.
 	 */
 	GIT_MERGE_TREE_FIND_RENAMES = (1 << 0),
+
+	/**
+	 * If a conflict occurs, exit immediately instead of attempting to
+	 * continue resolving conflicts.  The merge operation will fail with
+	 * GIT_EMERGECONFLICT.
+	 */
+	GIT_MERGE_TREE_FAIL_ON_CONFLICT = (1 << 1),
```
carlosmn: We should probably specify what happens to the resulting index (though presumably we don't do anything since we return an error, but it's good to be explicit).

tests/merge/merge_helpers.c
```diff
@@ -40,7 +40,7 @@ int merge_trees_from_branches(
 	cl_git_pass(git_commit_tree(&our_tree, our_commit));
 	cl_git_pass(git_commit_tree(&their_tree, their_commit));
 
-	cl_git_pass(git_merge_trees(index, repo, ancestor_tree, our_tree, their_tree, opts));
+	error = git_merge_trees(index, repo, ancestor_tree, our_tree, their_tree, opts);
```
carlosmn: Would this change (and the equivalent below) make the other users of `merge_trees_from_branches()` and `_commits_` to stop making sure that the merge succeeded? Other users should be changed to wrap their call in `cl_git_pass()` since we don't anymore, no?

include/git2/merge.h
ethomson: Updated this to be explicit.

include/git2/errors.h
ethomson: Yes, thanks; fixed.

tests/merge/merge_helpers.c
ethomson: Yes, sorry that's subtle.  Other users had already been asserting success here (wrapping their calls in `cl_git_pass`).

carlosmn: Ok, just wanted to make sure this ramification had been considered.

vmg:
You're an :angel: @ethomson

PR #3484, opened 2015-10-25
Return `git_buf` to an internal type while leaving the public API.

whoisj:
Obscures the `git_buf` data structure.

Add API for reading, copying, etc for dealing with git_buf values.

Has advantages for non-C languages to consume as cdecl functions are easier to call than structs are rebuild and maintain.

Adheres to the existing only exposing opaque types in the API.

** BREAKING CHANGE ***
This change would alter the API in a way that would cause most API consumers to edit code and recompile before they could accept the update.

### DIFF
--- a/include/git2/buffer.h
+++ b/include/git2/buffer.h
@@ -8,6 +8,7 @@
 #define INCLUDE_git_buf_h__
 
 #include "common.h"
+#include "types.h"
 
 /**
  * @file git2/buffer.h
@@ -19,42 +20,6 @@
 GIT_BEGIN_DECL
 
 /**
- * A data buffer for exporting data from libgit2
- *
- * Sometimes libgit2 wants to return an allocated data buffer to the
- * caller and have the caller take responsibility for freeing that memory.
- * This can be awkward if the caller does not have easy access to the same
- * allocation functions that libgit2 is using.  In those cases, libgit2
- * will fill in a `git_buf` and the caller can use `git_buf_free()` to
- * release it when they are done.
- *
- * A `git_buf` may also be used for the caller to pass in a reference to
- * a block of memory they hold.  In this case, libgit2 will not resize or
- * free the memory, but will read from it as needed.
- *
- * A `git_buf` is a public structure with three fields:
- *
- * - `ptr` points to the start of the allocated memory.  If it is NULL,
- *   then the `git_buf` is considered empty and libgit2 will feel free
- *   to overwrite it with new data.
- *
- * - `size` holds the size (in bytes) of the data that is actually used.
- *
- * - `asize` holds the known total amount of allocated memory if the `ptr`
- *    was allocated by libgit2.  It may be larger than `size`.  If `ptr`
- *    was not allocated by libgit2 and should not be resized and/or freed,
- *    then `asize` will be set to zero.
- *
- * Some APIs may occasionally do something slightly unusual with a buffer,
- * such as setting `ptr` to a value that was passed in by the user.  In
- * those cases, the behavior will be clearly documented by the API.
- */
-typedef struct {
-	char   *ptr;
-	size_t asize, size;
-} git_buf;
-
-/**
  * Static initializer for git_buf from static buffer
  */
 #define GIT_BUF_INIT_CONST(STR,LEN) { (char *)(STR), 0, (size_t)(LEN) }
@@ -63,7 +28,7 @@ typedef struct {
  * Free the memory referred to by the git_buf.
  *
  * Note that this does not free the `git_buf` itself, just the memory
- * pointed to by `buffer->ptr`.  This will not free the memory if it looks
+ * referenced by the buffer.  This will not free the memory if it looks
  * like it was not allocated internally, but it will clear the buffer back
  * to the empty state.
  *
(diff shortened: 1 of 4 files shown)


nkeck720:
What advantage does this have if I am using C, then? Would it break my code
for no reason?

whoisj:
> What advantage does this have if I am using C, then? Would it break my code
for no reason?

For the consumer of the API, little. For the developers of the API, much.

By putting the implementation of `git-buf` behind an obscure type reference the developers are able to make improvements to the implementation without breaking changes. Today's `git_buf` is rather trivial, and while it gets the work done it doesn't support reuse, streaming, etc. There may never be a reason to add such support, but if a developer wanted to contribute the functionality they'd be severely hampered by the current "internals exposed" `git_buf` API. By putting the implementation details behind and obsure type, future improvements are much easier -- and since Libgit2 is still pre-v1.0, I would think now is the time to make API breaking changes, not later.

carlosmn:
ObPedant: that'd be an opaque type, rather than internal. Those would not be accessible from the outside.

What we expose to the outside as the `git_buf` type is intentionally very simple. We're not a generic utility library and re-use and whatever other features for a `git_buf` may be available would only be for internal consumption.

We expose `git_buf` solely as a way to expose a buffer we've allocated to the user so they can free it with the right allocator. We're not interested in making this type something that's generally usable as a buffer in user code.

whoisj:
> We expose git_buf solely as a way to expose a buffer we've allocated to the user so they can free it with the right allocator. We're not interested in making this type something that's generally usable as a buffer in user code.

Sure, but you cannot improve upon it if you leave its internals exposed. Not sure what harm it does to make `git_buf` opaque other than to cause a few apps to have to recompile when taking an update (and most will anyways because the API is frequently changing).

ethomson:
So I think that there's two issues here.  From the original description:

> Has advantages for non-C languages to consume as cdecl functions are easier to call than structs are rebuild and maintain.

Obviously this could be a legitimate reason to add some more helper methods.  I'm not really enthused about providing a bunch of APIs for string manipulation, but I can concede that some languages make binding to C tortuous so let's discuss it.

Then there's making `git_buf` opaque.  You mention possible improvements - I'd much rather see a concrete proposal for improvement, because at the moment, this is just making a simple type opaque for some possible futureproofing.  Yet `git_buf` is *not* meant to do anything difficult, and I'm not sure that I see what improvement we might want to make.

Still, making it opaque and adding functions is no guarantee of binary compatibility.  We may need to change those functions - which is why I'd want to talk about future changes in a concrete way and not a theoretical one.

Finally, while this change may be helpful for some managed language users, as it moves the onus of copying data to libgit2 (where it's quite easy to just `memcpy`), it's actually quite unhelpful for C users.  Right now, you can give a buffer that *you* own to libgit2 using the `git_buf` - no copying necessary.  With this change, now I'll call `git_buf_copy` and use twice the memory.

Anyway.  Let's split this discussion up between what is helpful for you to use `git_buf`s, and whether this should be opaque or not.  I'm very keen to tackle the first, but the latter is a different conversation, I think.

whoisj:
Yes, the interop between native and managed (specifically with C#) was my focus, there's also the issue that `git_buf` is intended to be a self-sizing buffer. Today it can be, but it makes a lot of assumptions to achieve that.

First, `git_buf` assumes a single threaded model. There's no way for a `git_buf` to resize beside `realloc` which creates a new contiguous allocation, copies the memory, frees the old memory, and then we updates the `ptr` value in `git_buf` to address the new buffer. Any other thread which was in the middle of utilizing said buffer is now in "undefined" territory. If we ever want `git_buf` to become thread-safe (and maybe one day even make libgit2 thread-safe) we need a way to synchronized operations on the buffer itself.

Second, `git_buf` is always contiguous memory. This can be a challenge as memory usages goes up and/or becomes more fragmented. If we ever decide we want to enable `git_buf` to support multiple, smaller blocks of memory but "read" as a larger block we need to abstract away the simple `char *`.

There's other things - but these are the two big ones.

ethomson:
Do you have use cases for either of these things?

I want to be clear, though - I have *no interest* in putting a lock around a `git_buf`.  I have never thought that sharing a `git_buf` between two threads would be helpful, and throwing a critical section around it seems just painfully wasteful.

whoisj:
> Do you have use cases for either of these things?

We have a use case in Visual Studio where a `git_buf` realloc fails due to insufficient contiguous memory during an extreme usage case. This is considered an OOM and things do not end well. This is very much a corner case, but there are more useful and valid reasons to treat the buffer more like a buffer than an array. Simple buffer pooling / reuse scenarios become available once we can control how the buffers are stitched together and are bound by the implicit contract of memory being contiguous.

As for the critical section, it would be unnecessary if the type were opaque as I'm fairly certain the realloc does the equivollent of an [InterlockedExchange](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683590(v=vs.85).aspx), thus keeping the pointer safe so long as a memory barrier is issued when various threads read the `ptr`. An opaque type could easily enforce the memory barrier issuance, but that would mean hiding the `ptr` field lest developers assume they can treat it like a pointer.

ethomson:
> We have a use case in Visual Studio where a git_buf realloc fails due to insufficient contiguous memory during an extreme usage case.

When?  What are you doing that this happens?

I feel like you're trying to make a `git_buf` a stream.  We have those already - why is a `git_buf` the appropriate data structure to fix this, and not a stream?  What API is the problem here?

whoisj:
> I feel like you're trying to make a git_buf a stream.

In some ways yes, but it is not a stream. The entire buffer resides in memory.

The API issues are as outlined above:

1. Non-extensible.
2. Cannot be thread-safe.
3. Not friendly with C# (and likely other languages as well).
4. Requires contiguous allocations.

ethomson:
All of the things that you're pointing out as negatives are things that I would point out as a positive.

Again, this is a really simple data structure.  We use it *throughout the library* - and we use it exactly because it's a `char *` and a `size_t`.  That's all we want it to be.  We want none of the things that you are asking for.

The things that you are asking for are not unreasonable concerns, but they are not appropriate for `git_buf`.

That's why I'm asking you what the problem that you're actually having is, because we want to fix the problem, but changing `git_buf` around is not going to be the way to do it.

carlosmn:
We do already allow for buffer re-use with the current `git_buf` structure. If you provide a buffer which was filled previously from a different function, we'll re-use that allocation (there may be a few exceptions, but we generally clear the buffer instead of freeing it). Most of the time, we use it to return small strings, so the size of the allocation isn't an issue.

As for times when there is too much data to fit contiguously, it doesn't sound like we'd want a `git_buf` there. If you have a large blob that you want to read from the db, a stream would be a much more appropriate interface, to abstract how many chunks and how we allocate them. We already have the type for the odb, though we never implemented actual streaming, since it's not quite possible with the git format. If we create a more general `git_readstream` analogous to `git_writestream` that would let us have whatever fancy allocations behind the scenes. But that still wouldn't solve the underlying issue with reading large blobs from the db.

The delta format assumes you have it all in memory. You can build in multi-buffers, but unless `git_delta__apply()` learns how to deal with that, it's not going to help much.

PR #3486, opened 2015-10-26
Fix segfault when reading reflog with extra newlines

srajko:
This addresses the issue mentioned in #3458 where `git_reflog_read` segfaults when reading a reflog that has a newline in the middle of a reflog message.

Broken out from #3463 as a separate PR.


carlosmn:
Keeping a reflog for a non-existent branch sounds like a situation we shouldn't be generating as normal. It would indicate a bug in the implementation which removed the branch.

If we want to have a branch with a particular reflog, we should create it during the test in a sandbox, there's no need to put an orphan reflog in the test fixtures.

The cleanup commit should be squashed into the commit creating the test, as it doesn't add value by existing separately.

srajko:
@carlosmn Updated.  I used `git_reference_create` to create an invalid reflog, but I'm not sure that it *should* let me do that so (ab)using it for the test seems a bit sketchy.  Did you have something different in mind?

carlosmn:
Newlines in the message shouldn't be allowed. The documentation says it should be a single line, but evidently we don't actually check for that.

I wouldn't expect this to be solvable via a single call though the API. This is about invalid entries, so it would require manual manipulation of the reflog file.

srajko:
Yeah, I found a number of API calls that will let you corrupt the reflog with a newline.  I was planning to turn my attention to that after this round of changes gets finalized.

I updated the test to manually introduce the newline into the reflog.

tests/refs/reflog/reflog.c
```diff
@@ -154,6 +154,59 @@ void test_refs_reflog_reflog__reading_the_reflog_from_a_reference_with_no_log_re
...
+	cl_git_pass(git_reference_create(&ref, g_repo, refname, &id, 1, refmessage));
+
+	/* corrupt the branch reflog by introducing a newline inside the reflog message (we replace '*' with '\n') */
+	git_buf_join_n(&newline_log_path, '/', 3, git_repository_path(g_repo), GIT_REFLOG_DIR, refname);
+	cl_assert((f = fopen(git_buf_cstr(&newline_log_path), "r+")) != NULL);
+	cl_assert(fseek(f, 0, SEEK_END) != -1);
+	buf_length = ftell(f);
+	cl_assert(fseek(f, 0, SEEK_SET) != -1);
+	cl_assert((buf = malloc(buf_length)) != NULL);
+	cl_assert(fread(buf, 1, buf_length, f) == buf_length);
```
carlosmn: We have functions to do this for us. `git_fileutils_readbuffer()`.

tests/refs/reflog/reflog.c
```diff
@@ -154,6 +154,59 @@ void test_refs_reflog_reflog__reading_the_reflog_from_a_reference_with_no_log_re
...
+	for(i=0; i<buf_length; i++) {
+		if(buf[i] == '*') {
+			buf[i] = '\n';
+		}
+	}
+	cl_assert(fseek(f, 0, SEEK_SET) != -1);
+	cl_assert(fwrite(buf, 1, buf_length, f) == buf_length);
```
carlosmn: We have `cl_git_rewritefile()` to do this for us (and not assume a single write is going to succeed).

tests/refs/reflog/reflog.c
```diff
@@ -154,6 +154,59 @@ void test_refs_reflog_reflog__reading_the_reflog_from_a_reference_with_no_log_re
...
+	cl_assert(fseek(f, 0, SEEK_END) != -1);
+	buf_length = ftell(f);
+	cl_assert(fseek(f, 0, SEEK_SET) != -1);
+	cl_assert((buf = malloc(buf_length)) != NULL);
+	cl_assert(fread(buf, 1, buf_length, f) == buf_length);
+	for(i=0; i<buf_length; i++) {
+		if(buf[i] == '*') {
+			buf[i] = '\n';
+		}
+	}
```
carlosmn: I'd recommend `strchr()` here as a more concise way to express this search. We should also assert that we did perform the replacement. otherwise something's gone quite wrong in the preparation of the test.

srajko:
@carlosmn Thanks for the tips, very helpful!  Switched to `git_fileutils_readbuffer`, `cl_git_rewritefile`, `strchr`, and asserting for successful replacement.

carlosmn:
Thanks

PR #3487, opened 2015-10-26
Fix git_commit_summary to collapse whitespace around newlines

srajko:
This addresses the issue mentioned in #3458 where `git_commit_summary` returns newlines when they occur following whitespace.  `git_commit_summary` now collapses all whitespace around a newline into a single space, as per the discussion in #3458.

Broken out from #3463 as a separate PR.

### DIFF
--- a/src/commit.c
+++ b/src/commit.c
@@ -436,17 +436,23 @@ const char *git_commit_summary(git_commit *commit)
 
 	if (!commit->summary) {
 		for (msg = git_commit_message(commit), space = NULL; *msg; ++msg) {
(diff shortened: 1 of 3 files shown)


carlosmn:
The second commit should be squashed into the first as it's a style fixup.

ethomson:
I was too excited about this change!  So I went ahead and squashed this and merged it manually.

Thanks, again!

srajko:
@ethomson Oops, this totally slipped my mind :-(  My bad - thanks for taking care of it!

PR #3488, opened 2015-10-27
pool: Simplify implementation

vmg:
So, here's the deal: the existing pool implementation is terribly slow. For some merge cases, you can see more than 70% of the runtime dominated by pool allocations!

Why is it so slow? Because it does a lot of very complex calculations & pointer wrangling just for the sake of keeping a "free list" and trying to squeeze some bytes out of each allocation. IMO this basically defeats the point of a memory pool/slab allocator.

I've re-implemented the pool to look more like a slab allocator. There's always an existing slab; when the existing slab has not enough space to fulfill the requested allocation, we allocate a new slab.

This implementation is hundreds of times faster than the old pool for most cases. For instance: a particular merge of the `cdnjs` repository that I discussed with @ethomson used to take 17s on my machine. 10 of those seconds were spent in the pool (lol wtf?).

With the new implementation, the merge finishes in 5.2s.

### Questions

- **How does the new aggressive slab algorithm impact memory usage?** This is hard to guess. For the specific case I'm testing, it doesn't, because all the allocated strings are of similar size. For some pathological cases with very large strings, I could see this wasting a little bit of memory. Frankly, given the terrible performance of the old pool, I'm confident the tradeoff will always be positive.

- **What happens with the APIs that allowed freeing pointers from the pool?** They are gone -- it's not possible to free specific pointers from a traditional slab allocator because there's no free-list. Fortunately, only one place of the existing code was calling the pool `free` API: the iterators code. Removing the call will increase memory usage on the intermediate steps of the iterator, until the pool is fully freed. I think this will be acceptable too.

- **What can possibly go wrong?** I am not sure. It's a pretty large change in a critical code path of the library, but the resulting code is so significantly simpler that I'm pretty confident I haven't introduced any regressions. I'd appreciate @carlosmn and @ethomson's careful eyes on this.

### DIFF
--- a/src/commit_list.c
+++ b/src/commit_list.c
@@ -47,7 +47,7 @@ git_commit_list *git_commit_list_insert_by_date(git_commit_list_node *item, git_
 
 git_commit_list_node *git_commit_list_alloc_node(git_revwalk *walk)
 {
-	return (git_commit_list_node *)git_pool_malloc(&walk->commit_pool, COMMIT_ALLOC);
+	return (git_commit_list_node *)git_pool_mallocz(&walk->commit_pool, 1);
 }
 
 static int commit_error(git_commit_list_node *commit, const char *msg)
--- a/src/merge.c
+++ b/src/merge.c
@@ -1307,7 +1307,6 @@ GIT_INLINE(int) index_entry_dup_pool(
 {
 	if (src != NULL) {
 		memcpy(out, src, sizeof(git_index_entry));
-
 		if ((out->path = git_pool_strdup(pool, src->path)) == NULL)
 			return -1;
 	}
@@ -1442,10 +1441,11 @@ git_merge_diff_list *git_merge_diff_list__alloc(git_repository *repo)
 
 	diff_list->repo = repo;
 
+	git_pool_init(&diff_list->pool, 1);
+
 	if (git_vector_init(&diff_list->staged, 0, NULL) < 0 ||
 		git_vector_init(&diff_list->conflicts, 0, NULL) < 0 ||
-		git_vector_init(&diff_list->resolved, 0, NULL) < 0 ||
-		git_pool_init(&diff_list->pool, 1, 0) < 0) {
+		git_vector_init(&diff_list->resolved, 0, NULL) < 0) {
 		git_merge_diff_list__free(diff_list);
 		return NULL;
 	}
--- a/src/merge.c
+++ b/src/merge.c
@@ -626,7 +626,7 @@ static int merge_conflict_resolve_one_renamed(
 		git_oid__cmp(&conflict->our_entry.id, &conflict->their_entry.id) != 0)
 		return 0;
 
-	if ((merged = git_pool_mallocz(&diff_list->pool, sizeof(git_index_entry))) == NULL)
+	if ((merged = git_pool_malloc(&diff_list->pool, sizeof(git_index_entry))) == NULL)
 		return -1;
 
 	if (ours_changed)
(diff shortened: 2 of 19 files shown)


src/commit_list.c
```diff
@@ -47,7 +47,7 @@ git_commit_list *git_commit_list_insert_by_date(git_commit_list_node *item, git_
 
 git_commit_list_node *git_commit_list_alloc_node(git_revwalk *walk)
 {
-	return (git_commit_list_node *)git_pool_malloc(&walk->commit_pool, COMMIT_ALLOC);
+	return (git_commit_list_node *)git_pool_mallocz(&walk->commit_pool, 1);
```
vmg: Sooo... We were allocating

vmg: Correction: We were not. I introduced the over-allocation as part of the refactoring. :sweat:

tests/core/pool.c
```diff
@@ -32,26 +25,24 @@ void test_core_pool__1(void)
...
+	p.page_size = 4000;
 
 	for (i = 2010; i > 0; i--)
 		cl_assert(git_pool_malloc(&p, i) != NULL);
 
 	/* with fixed page size, allocation must end up with these values */
-	cl_assert_equal_i(1, git_pool__open_pages(&p));
-	cl_assert_equal_i(507, git_pool__full_pages(&p));
-
+	cl_assert_equal_i(590, git_pool__open_pages(&p));
```
carlosmn: This looks like a non-trivial increase in pages. Is this down to the couple of pointer-sizes we take off of each page? If so, why isn't this adjusted for 32-bit systems like the other comparison?

src/pool.h
```diff
@@ -28,19 +28,11 @@ typedef struct git_pool_page git_pool_page;
  * For examples of how to set up a `git_pool` see `git_pool_init`.
  */
 typedef struct {
-	git_pool_page *open; /* pages with space left */
-	git_pool_page *full; /* pages with no space left */
-	void *free_list;     /* optional: list of freed blocks */
+	git_pool_page *pages; /* pages with space left */
```
carlosmn: The comment is now inaccurate since we store a single list, so this is just the count of pages we've allocated.

tests/core/pool.c
vmg: @carlosmn: No, this is because the test is exercising the pathological case for the slab allocator. It creates a 2010 byte object, and then creates another one which is just 2009 bytes, so it falls *just* outside the available space for the current page, and we need to allocate a new one. It ends up with a lot of pages that are roughly 50% full, but without a free-list, the half-full pages cannot be reused like before.

In real world scenarios, we don't allocate objects that are 50% the size of a page, so the waste is nowhere as large.

src/pool.h
vmg: :+1:

tests/core/pool.c
carlosmn: Make sense; thanks.

src/merge.c
```diff
@@ -626,7 +626,7 @@ static int merge_conflict_resolve_one_renamed(
 		git_oid__cmp(&conflict->our_entry.id, &conflict->their_entry.id) != 0)
 		return 0;
 
-	if ((merged = git_pool_malloc(&diff_list->pool, sizeof(git_index_entry))) == NULL)
+	if ((merged = git_pool_mallocz(&diff_list->pool, sizeof(git_index_entry))) == NULL)
```
ethomson: There are a couple of places where we switch to `mallocz` instead of `malloc` but promptly write the whole of the allocated area.  Seems unnecessary.

vmg:
I just ran the full test suite with this branch vs the old pool implementation. `jemalloc` reports 461984 total bytes allocated by the old vs 472696 allocated with the new implementation: that's 10KB difference between the two branches. Considering that the test suite seems to suffer from a lot of repetitive patterns, I think these are pretty good results.

src/merge.c
vmg: Can you help me track down all of them? I've probably been overzealous here.

vmg:
OK, I think my previous `jemalloc` metrics where bogus. I've manually added stat accounting and we're allocating 88mb of pool pages in the old code vs 94mb in the new code.

This makes a lot more sense and feels good to me.  :+1::+1::+1:

src/merge.c
vmg: OK, I think I have all of them.

PR #3489, opened 2015-10-27
Better REUC generation when merging

vmg:
First, we optimize the way REUC insertion works to make it hundreds of times faster (from 17s to 200ms when inserting 39k entries).

Second, we add a merge trees flag that allows us to skip the REUC generation altogether (for users that don't need to display merge results).

Lastly, we plug a small memory leak that we introduced with the `FAIL_ON_CONFLICT` flag.

cc @carlosmn @ethomson


nkeck720:
Nice, clean code here.

carlosmn:
:+1: the new reuc insertion code is easier to read to boot.

ethomson:
:+1:

PR #3490, opened 2015-10-29
branch: make cleanup labels consistent

travisjeffery:
All the other cleanup labels are named "cleanup"  other than this one called "done", so this makes everything nice and consistent.


carlosmn:
Thanks, but we don't generally accept this kind of patch, as it increases code churn (making it harder to figure out where the code came from through the log and blame) without a noticeable impact on the code as this is just internally used labels where the purpose is clear anyhow.

PR #3491, opened 2015-10-29
Use checksums to detect config file changes

carlosmn:
Instead of using the file size and timestamp, which can lead us to ignore changes performed in the same second, use a checksum of the file contents to detect when it has changed.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -77,8 +77,7 @@ typedef struct git_config_file_iter {
 		 (iter) = (tmp))
 
 struct reader {
-	time_t file_mtime;
-	size_t file_size;
+	git_oid checksum;
 	char *file_path;
 	git_buf buffer;
 	char *read_ptr;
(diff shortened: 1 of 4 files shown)


mikeando:
You could keep the time stamp check and only fall back to a checksum if they differ by less than one second (or some other suitable duration...)

carlosmn:
That doesn't buy us anything. We still need to keep track of the checksum in case the next update is in the same second.

mikeando:
Ah yep, you're right Carlos.

Though there is a time based check that might work - rather than caching the modification time, cache the last time you checked the file. Then when you get the files stats and if the modification time is earlier than your last check - it wont have changed and you wont need to update the checksum. If its equal or more recent it has (or may have) changed and you will need to recalculate the checksum.

However, this requires the local clock to match the filesystem, and I think there are probably cases where you can't trust this to be the case - so it may be a non-starter.

PR #3494, opened 2015-10-30
Fix build for unit test

leoyanggit:
If none of GIT_OPENSSL, GIT_WINHTTP or GIT_SECURE_TRANSPORT
is defined we should also be able to build the unit test.

### DIFF
--- a/tests/online/badssl.c
+++ b/tests/online/badssl.c
@@ -5,23 +5,34 @@
 static git_repository *g_repo;
 
 #if defined(GIT_OPENSSL) || defined(GIT_WINHTTP) || defined(GIT_SECURE_TRANSPORT)
+static bool g_has_ssl = true;
+#else
+static bool g_has_ssl = false;
+#endif
 
 void test_online_badssl__expired(void)
 {
+	if (!g_has_ssl)
+		cl_skip();
+
 	cl_git_fail_with(GIT_ECERTIFICATE,
 			 git_clone(&g_repo, "https://expired.badssl.com/fake.git", "./fake", NULL));
 }
 
 void test_online_badssl__wrong_host(void)
 {
+	if (!g_has_ssl)
+		cl_skip();
+
 	cl_git_fail_with(GIT_ECERTIFICATE,
 			 git_clone(&g_repo, "https://wrong.host.badssl.com/fake.git", "./fake", NULL));
 }
 
 void test_online_badssl__self_signed(void)
 {
+	if (!g_has_ssl)
+		cl_skip();
+
 	cl_git_fail_with(GIT_ECERTIFICATE,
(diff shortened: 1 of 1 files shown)


carlosmn:
These should not be successful, as we did not run the tests. They should be skipped.

leoyanggit:
How to skip it?

carlosmn:
With `cl_skip()`.

leoyanggit:
Hmm `cl_skip()` seems to skip test in the runtime. But the problem here is that if none of those macros is defined the code won't compile. Or you mean in addition to my patch we should also use `cl_skip()` in my those 3 empty functions?

ethomson:
Hmm.  I'm not sure I understand why you can't *build* those tests without those SSL?  `git_clone` is public API, as is `GIT_ECERTIFICATE`...  I would expect that you could build them but that they would fail.

If that's not the case, then that's the first thing that we need to fix.

Once that's done, I would expect to see something like:

```
void test_online_badssl__self_signed(void)
{
    if (!has_ssl)
        cl_skip();

    cl_git_fail_with(GIT_ECERTIFICATE,
        git_clone(&g_repo, "https://self-signed.badssl.com/fake.git", "./fake", NULL));
}
```

leoyanggit:
As I mentioned above:  I have none of GIT_OPENSSL, GIT_WINHTTP or GIT_SECURE_TRANSPORT defined. Those 3 test functions are guarded by them. So I can't compile.

ethomson:
But there's nothing about those functions that you can't compile, right?  In that case, you should put a `cl_skip` at the beginning of the test if you don't want to run it.  You can leverage the existing `#ifdefs` to determine whether to skip or not.

leoyanggit:
> But there's nothing about those functions that you can't compile, right?

right.

Seems like I don't get how `cl_skip` works. By at the beginning of the test what do you mean?

ethomson:
Like what I posted above.  I think you want to do something like:

```
#include "git2/clone.h"

static git_repository *g_repo;

#if defined(GIT_OPENSSL) || defined(GIT_WINHTTP) || defined(GIT_SECURE_TRANSPORT)
static bool has_ssl = true;
#else
static bool has_ssl = false;
#endif
 
void test_online_badssl__self_signed(void)
{
    if (!has_ssl)
        cl_skip();

    cl_git_fail_with(GIT_ECERTIFICATE,
        git_clone(&g_repo, "https://self-signed.badssl.com/fake.git", "./fake", NULL));
}

...etc...
```

leoyanggit:
I see. Instead of forking the test functions just adding some flags. I'll update my patch to follow this style.

leoyanggit:
OK, patch updated.

ethomson:
Awesome!  :dancers: 

Thanks for this.

leoyanggit:
Thank you for the review!

PR #3495, opened 2015-10-30
Fix build for custom transport users

leoyanggit:
We should explicitly include the declaration of git_strarray
from "include/git2/sys/transport.h"


ethomson:
Thanks!

PR #3497, opened 2015-10-31
Support union merges via .gitattributes file

### LINKED ISSUE #2180: Custom merge driver support
There are quite a number of use cases where having support for custom merge drivers would be useful.

The simplest example is probably a change log. Two branches are created, starting at the same commit. Both branches insert an entry at the top of the change log file. The first one merges without problems, the second requires a rebase.

It would be nice if both branches could merge without the need for a rebase.

stanhu:
When a repository has a `.gitattributes` file that has a `merge=union` line for a CHANGELOG file, for example, libgit2 ignores this file and reports a conflict when git doesn't have one. This is partially to fix #2180.

I tried adding a test for this, but adding `.gitattributes` file to the `merge-resolve` directory seems to cause additional failures due to the change in entries. I wanted to get feedback on this before spending even more time on that.

### DIFF
--- a/src/merge.c
+++ b/src/merge.c
@@ -1822,6 +1822,28 @@ static git_iterator *iterator_given_or_empty(git_iterator **empty, git_iterator 
 	return *empty;
 }
 
+static int lookup_file_favor(
+	git_merge_file_favor_t *file_favor,
+	git_repository *repo,
+	const char *path)
+{
+	int error = 0;
+	const char *value = NULL;
+
+	if (path) {
+		if ((error = git_attr_get(&value, repo, 0, path, "merge")) < 0)
+			goto done;
+
+		if (*file_favor == GIT_MERGE_FILE_FAVOR_NORMAL &&
+		    value && strcmp(value, "union") == 0) {
+			*file_favor |= GIT_MERGE_FILE_FAVOR_UNION;
+		}
+	}
+
+done:
+	return error;
+}
+
 int git_merge__iterators(
 	git_index **out,
 	git_repository *repo,
@@ -1879,6 +1901,10 @@ int git_merge__iterators(
 
 	git_vector_foreach(&changes, i, conflict) {
 		int resolved = 0;
+
+		/* Check for merge options in .gitattributes */
+		if ((error = lookup_file_favor(&opts.file_favor, repo, conflict->our_entry.path) < 0))
+			goto done;
 
 		if ((error = merge_conflict_resolve(
 			&resolved, diff_list, conflict, &file_opts)) < 0)
@@ -2087,7 +2113,7 @@ static int iterator_for_annotated_commit(
 	opts.flags = GIT_ITERATOR_DONT_IGNORE_CASE;
 
 	if (commit == NULL) {
-		error = git_iterator_for_nothing(out, &opts); 
+		error = git_iterator_for_nothing(out, &opts);
 	} else if (commit->type == GIT_ANNOTATED_COMMIT_VIRTUAL) {
 		error = git_iterator_for_index(out, commit->index, &opts);
 	} else {
@@ -2431,7 +2457,7 @@ static int write_merge_msg(
 	assert(repo && heads);
 
 	entries = git__calloc(heads_len, sizeof(struct merge_msg_entry));
-	GITERR_CHECK_ALLOC(entries); 
+	GITERR_CHECK_ALLOC(entries);
 
 	if (git_vector_init(&matching, heads_len, NULL) < 0) {
 		git__free(entries);
@@ -2485,7 +2511,7 @@ static int write_merge_msg(
 
 	if (matching.length)
 		sep =',';
-	
+
 	if ((error = merge_msg_entries(&matching, entries, heads_len, msg_entry_is_tag)) < 0 ||
 		(error = merge_msg_write_tags(&file, &matching, sep)) < 0)
 		goto cleanup;
--- a/src/merge.c
+++ b/src/merge.c
@@ -1903,7 +1903,7 @@ int git_merge__iterators(
 		int resolved = 0;
 
 		/* Check for merge options in .gitattributes */
-		if ((error = lookup_file_favor(&opts.file_favor, repo, conflict->our_entry.path) < 0))
+		if ((error = lookup_file_favor(&file_opts.favor, repo, conflict->our_entry.path) < 0))
 			goto done;
 
 		if ((error = merge_conflict_resolve(
--- a/tests/merge/workdir/simple.c
+++ b/tests/merge/workdir/simple.c
@@ -330,6 +330,42 @@ void test_merge_workdir_simple__union(void)
 	cl_assert(merge_test_reuc(repo_index, merge_reuc_entries, 4));
 }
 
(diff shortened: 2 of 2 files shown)


src/merge.c
```diff
@@ -1712,6 +1733,10 @@ int git_merge__iterators(
 	git_vector_foreach(&changes, i, conflict) {
 		int resolved = 0;
 
+		/* Check for merge options in .gitattributes */
+		if ((error = git_merge__lookup_file_favor(repo, conflict->our_entry.path, &opts) < 0))
```
DouweM: Since this modifies `opts` in place, won't the newly set `opts->file_favor` bleed into conflict resolution for other files, processed after the file that had the attribute set?

stanhu: Yes, good point. I've updated the code to not do that.

src/merge.c
```diff
@@ -1666,6 +1666,27 @@ static git_iterator *iterator_given_or_empty(git_iterator **empty, git_iterator
...
+	int *file_favor)
+{
+	int error = 0;
+
+	if (path) {
+		const char *value = NULL;
+		if ((error = git_attr_get(&value, repo, 0, path, "merge")) < 0)
+			goto done;
+
+		if (value && strncmp(value, "union", 5) == 0) {
```
ethomson: Why `strncmp`?  What if I had a custom merge driver like `unionjack`?

src/merge.c
```diff
@@ -1711,8 +1732,13 @@ int git_merge__iterators(
 
 	git_vector_foreach(&changes, i, conflict) {
 		int resolved = 0;
+		int file_favor = opts.file_favor;
+
+		/* Check for merge options in .gitattributes */
+		if ((error = git_merge__lookup_file_favor(repo, conflict->our_entry.path, &file_favor) < 0))
```
ethomson: Why ours as opposed to the other paths?  What if there is no ours?

src/merge.c
stanhu: Oh, I didn't think that was a valid option at the moment. git seems to support `ours`, `theirs`, and `union`. This could be updated to make sure the `union` were string-terminated.

ethomson: Hmm...  are you seeing reports that `value` isn't null terminated?

stanhu: No, I mean to ensure that `unionjack` doesn't match, the `strncmp` could be updated to match length 6 instead of 5.

ethomson: Why wouldn't you just do a `strcmp` though?

stanhu: That would work too. I'll update.

src/merge.c
stanhu: Good question. It was an arbitrary decision because for a given conflict, the presumption was that `our_entry` and `their_entry` should have the same path name. I'm not totally sure if that is correct. Reading `ll-merge.c` from git, I'm not quite sure what it is doing yet.

ethomson: Right...  to be completely honest, it's probably easier to simply experiment.  I *suspect* that the ours side is the name to use, but I don't know for sure...  So they can definitely be different, when there's a rename conflict.  Eg, if you have an ancestor of `foo`, and the ours side renames `foo` -> `bar` and the theirs side renames `foo` -> `baz` then it's not clear to me which one you should use.

But also, there's no guarantee that there is an ours side (if it was deleted), so do be sure to test before you dereference it.

stanhu: Right. I believe union merges should only apply when their is conflict between `ours` and theirs`, so I've tightened up the check so that it will only be enabled when both paths exist and match.

ethomson: Well, I'd still be curious what git does.  :)

stanhu: I traced through the git source code. It uses `ours` here:

https://github.com/git/git/blob/master/merge-recursive.c#L875

dblessing:
Is there anything else needed on this PR? Please let us know if we can help.

ethomson:
Sorry about the delay!  A couple of things, most are stylistic and minor:

1. We like to put our `out`-type parameters to the function up front.  Eg, something more like:

        static int git_merge__lookup_file_favor(
            int *file_favor,
            git_repository *repo,
            const char *path)

2. Regarding function naming, those that are fully "namespaced" with double underscores (eg, `git_merge__function`) are meant to be internal to the entire library.  (As opposed to those that are fully namespaced with single underscores, like `git_merge_files`, which are public API.)  For a simple static helper function like this, it should not be namespaced, and something like `lookup_file_favor` would be more appropriate.

3. It looks like the configuration `file_favor` option overrides the option specified by the caller.  This should be the other way around:  the configuration should override the `git_merge_file_favor_t` in the merge options if the options did not specify a side to favor (eg, `file_favor == GIT_MERGE_FILE_FAVOR_NORMAL`).  However, if the options *did* specify a side to favor, then that should override any configuration setting.

4. Fundamentally, it would be nice to support more than just this one option to the `merge` attribute (eg, `binary`, `text`, etc).  I don't know that this *necessarily* needs to be done in this pull request, but I feel like it would actually affect the architecture more fundamentally.  (You would likely find that the `file_favor` is not the appropriate way to report this data, for example.)  So this might be something that you find valuable to explore, I don't know.

stanhu:
Thanks, @ethomson. I've updated the PR based on your comments. I actually think the right way to support different `merge` attributes would be to implement custom merge drivers as mentioned in #2180 so that conflicts never actually surface here, but this seemed like an easier change to support the `CHANGELOG` case in the short term.

I did attempt to write a test for this case, but because I added a `.gitattributes` file to one of the test repositories, that seemed to have break other tests. I'm not sure what the best way to proceed there.

ethomson:
Neat!  I think that the easiest way to test this would be to create a test that writes the gitattributes file before doing the merge.  (As you point out - a lot of the other tests depend on the current state of those test repositories.)

You should be able to do something as simple as:

    cl_git_mkfile("merge-resolve/.gitattributes", "*.cs merge=union\n");

(Replacing the path and contents with something more appropriate.)

stanhu:
Thanks for the tip, @ethomson! I've updated the PR with a test.

stanhu:
@ethomson, would you mind taking a look? Thank you.

stanhu:
Somehow my test case vanished when I rebased. Working on that...

stanhu:
Ok, fixed. Looks like the rebase introduced a bug that prevented this functionality from working that was caught by the test.

DouweM:
@ethomson Friendly reminder to review this pull request :)

ethomson:
Hi - sorry about the delay.  Thanks so much for the fine work here.

As I reviewed this, I must admit that I felt very bad that we were going to open up `merge=union` as an attribute but *only* `merge=union`.  This seemed begging for the custom merge driver support that #2180 had requested and been so patiently waiting for.

As a result, I opened #3564 that includes this as well as `merge=<custom>` support for custom merge drivers.

Thanks again!

stanhu:
@ethomson Thanks for implementing custom merge support--even better! I suppose we can close this PR then?

ethomson:
Sounds good, I rolled both these commits into that PR, so I'll go ahead and close this one.

Thanks again for the fine work.

PR #3498, opened 2015-11-03
Diff: Honor `core.symlinks=false` and fake symlinks

ethomson:
On platforms where `core.symlinks=false`, diff needs to honor the fake symlinks that we (and Git for Windows) lay down, where a regular file is written with the symlink target.

At present, we propagate the mode bits from the index into diff, so that it knows that the file should be treated as a symlink.  Regrettably, diff then tries to *read the symlink*.  On platforms where `core.symlinks=false`, we should instead try to read the regular file, but preserve the mode.

There's a bonus unit test to ensure that `git_index_add_bypath` gets this right also.  (It does!)

### DIFF
--- a/src/diff_file.c
+++ b/src/diff_file.c
@@ -259,10 +259,35 @@ static int diff_file_content_load_blob(
 	return error;
 }
 
+static int diff_file_content_load_workdir_symlink_fake(
+	git_diff_file_content *fc, git_buf *path)
+{
+	git_buf target = GIT_BUF_INIT;
+	int error;
+
+	if ((error = git_futils_readbuffer(&target, path->ptr)) < 0)
+		return error;
+
+	fc->map.len = git_buf_len(&target);
+	fc->map.data = git_buf_detach(&target);
+	fc->flags |= GIT_DIFF_FLAG__FREE_DATA;
+
+	git_buf_free(&target);
+	return error;
+}
+
 static int diff_file_content_load_workdir_symlink(
 	git_diff_file_content *fc, git_buf *path)
 {
 	ssize_t alloc_len, read_len;
+	int symlink_supported, error;
+
+	if ((error = git_repository__cvar(
+		&symlink_supported, fc->repo, GIT_CVAR_SYMLINKS)) < 0)
+		return -1;
+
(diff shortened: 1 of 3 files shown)


ethomson:
Note that this was inspired by our *oldest active pull request*.  I changed the unit test around just a little bit.

Fixes #878

src/diff_file.c
```diff
@@ -259,10 +259,35 @@ static int diff_file_content_load_blob(
...
+}
+
 static int diff_file_content_load_workdir_symlink(
 	git_diff_file_content *fc, git_buf *path)
 {
 	ssize_t alloc_len, read_len;
+	int symlink_supported, error;
+
+	if ((error = git_repository__cvar(
+		&symlink_supported, fc->repo, GIT_CVAR_SYMLINKS)) < 0)
```
vmg: Couldn't we cache this more efficiently than calling `__cvar`? I'm afraid `__cvar` already caches, but it feels like this should be something we should lookup once and only once.

ethomson: Maybe...

This value needs to be scoped to the repository (since you could set `core.symlinks=true` on Windows for a repo - which is something we don't *yet* support but probably want to, since Git for Windows has blazed that trail for us.)  And we probably want to use the whole configuration caching mechanism when we do.

We might have an opportunity to read this value when we start the diff, and persist it throughout the creation of the `git_diff_file_content` objects if they're created by a `git_diff`.  But it seems that those can also be [created without a lot of context](https://github.com/libgit2/libgit2/blob/master/src/diff_file.c#L130) so we can't necessarily rely on that.  In that case, we'd fall back to querying the config cache in the repo...

I suspect that we could refactor ourselves into a place where we're only calling `git_repository__cvar` once in the common case, but I worry that we'll spend more time doing the refactoring than would ever be spent calling that function in actual use.

carlosmn: Once `__cvar()` has cached the value (which would only need to happen once), it's an access in a vector and a check to see that it has been cached. So while it'd be nice to see the value being passed in, I'm not worried about the performance aspect here, especially considering how we're doing IO in this area.

Issue #3504, opened 2015-11-09
EXC_BAD_ACCESS in attr_cache_upsert compare-and-swap when iterating file status


logancollins:
In the current commit on master, I'm seeing a bad memory access when using git_status_file. The crash is in git__compare_and_swap, called by attr_cache_upsert. I'm not at all familiar with how git's attr caching works, so with some limited investigation I'm not sure offhand what's wrong (whether on my end, or the library's).

The bad pointer appears to be entry->file or one of its members, as the compare-and-swap is attempting to set one of the "entry->file" members. I'm attaching screenshots of the stacktrace, the section of code in attrcache.c, and the debugger output of the potentially relevant variables.

If you're not able to reproduce this (I can reproduce 100%), I can try to provide more debugging info. I am experiencing this on iOS 9, arm64, compiled with Clang.

![screen shot 2015-11-09 at 9 01 30 am](https://cloud.githubusercontent.com/assets/714246/11040420/018ce486-86c1-11e5-81d5-0c8c507ee6be.png)
![screen shot 2015-11-09 at 9 02 13 am](https://cloud.githubusercontent.com/assets/714246/11040422/05157ca8-86c1-11e5-83ce-e78b1d48a354.png)
![screen shot 2015-11-09 at 9 03 00 am](https://cloud.githubusercontent.com/assets/714246/11040425/079befc0-86c1-11e5-892e-2497c641388f.png)

carlosmn:
Does this happen against all repos or a specific one? Can you reproduce on a "desktop" arch? Did you call `git_libgit2_init()` at the start of your program?

logancollins:
I am calling `git_libgit2_init()`. I just tried reproducing this on OS X, but it's working fine. It doesn't seem to be repository specific, as I can reproduce it with two repos, one that has working directory changes, and one that's a clean working directory.

Offhand, the first major difference I can think for iOS and OS X is that iOS has a case-sensitive filesystem? Would that impact this at all?

In the meantime, I'm going to experiment a bit more with both OSs and see if I can trace to a common difference.

logancollins:
The previous commit I was using before hitting this issue was `d0976a373e48b7ba3d919432b99e64de1812208f`, so I'm guessing some change in the interim 225 commits (!) is the source. I can also bisect back and see if I can trace it to a specific point.

carlosmn:
[...]

PR #3506, opened 2015-11-12
pool: Never return unaligned buffers

vmg:
Fixes https://github.com/libgit2/libgit2/issues/3504

@logancollins: Could you please verify that this works now in ARM64 & ARM32? I think we're doing alignment properly now for all architectures, but I don't have that hardware to try. :)

### DIFF
--- a/src/pool.c
+++ b/src/pool.c
@@ -8,7 +8,7 @@ struct git_pool_page {
 	git_pool_page *next;
 	uint32_t size;
 	uint32_t avail;
-	char data[GIT_FLEX_ARRAY];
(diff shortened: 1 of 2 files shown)


logancollins:
Yes! This appears to fix the issue.

ethomson:
Thanks @vmg for the fix and @logancollins for the helpful bug report!

carlosmn:
32-bit Windows is finding a mismatch in the number of allocated pages. Presumably other 32-bit arches would do too, but I think Windows is the only one we build.

vmg:
@carlosmn: :kissing: thanks for the fix

PR #3508, opened 2015-11-14
Improvements to tree parsing speed

carlosmn:
Here's a couple of simple changes to the way we parse trees which gives us significant improvements.

The first one is just silly, really. We've already calculated how long the filename is, so we can just skip over it instead of looking for the terminator again. This gives us about half the gains.

Then, we can also very easily avoid constantly asking the system for memory by allocating the entries in a pool owned by the tree. The lifetime of the entries is tied to the tree already, so this is a great place to use a pool. Those entries for which we give ownership to the user don't need to change, as we already perform an extra allocation to give them their own lifetime.

I checked the speedup by parsing the top-level tree for `git.git` for ~41k commits. The timing is a bit rough, but the speedup ends up being a bit under 1/3, which is not bad, considering. The most expensive thing right now is parsing the filemode number; and using libc's (presumably) optimised one doesn't really help.

|             |  master | this |
-----------|-----------|--------
| Debug | 4.7s | 3.8s |
| Release | 2.7s | 2.0 s |

I've also tested by grabbing the parent's tree for each commit we walk and diffing it with a pathspec of "README" (I initially tried with a full diff but that takes over two minutes, which is also pretty bad but a different story). The speedup isn't quite as drastic since we're still doing the diff, which I haven't touched here, but still noticeable, especially in release mode.

|             |  master | this |
-----------|-----------|--------
| Debug | 11.5s | 9.5s |
| Release | 6.0s | 4.3 s |

### DIFF
--- a/src/tree.c
+++ b/src/tree.c
@@ -414,10 +414,8 @@ int git_tree__parse(void *_tree, git_odb_object *odb_obj)
 			entry->attr = attr;
 		}
 
-		while (buffer < buffer_end && *buffer != 0)
-			buffer++;
-
-		buffer++;
+		/* Advance to the ID just after the path */
+		buffer += entry->filename_len + 1;
 
 		git_oid_fromraw(&entry->oid, (const unsigned char *)buffer);
 		buffer += GIT_OID_RAWSZ;
--- a/src/tree.c
+++ b/src/tree.c
@@ -125,7 +125,7 @@ static git_tree_entry *alloc_entry(const char *filename)
 
 struct tree_key_search {
 	const char *filename;
-	size_t filename_len;
+	uint16_t filename_len;
 };
 
 static int homing_search_cmp(const void *key, const void *array_member)
--- a/src/tree.c
+++ b/src/tree.c
@@ -82,24 +82,47 @@ int git_tree_entry_icmp(const git_tree_entry *e1, const git_tree_entry *e2)
 }
 
 /**
- * Allocate a tree entry, borrowing the filename from the tree which
- * owns it. This is useful when reading trees, so we don't allocate a
- * ton of small strings but can use the pool.
+ * Allocate either from the pool or from the system allocator
  */
-static git_tree_entry *alloc_entry_pooled(git_pool *pool, const char *filename, size_t filename_len)
+static git_tree_entry *alloc_entry_base(git_pool *pool, const char *filename, size_t filename_len)
 {
 	git_tree_entry *entry = NULL;
 	size_t tree_len;
 
+	if (filename_len > UINT16_MAX) {
+		giterr_set(GITERR_INVALID, "tree entry is over UINT16_MAX in length");
+		return NULL;
+	}
+
 	if (GIT_ADD_SIZET_OVERFLOW(&tree_len, sizeof(git_tree_entry), filename_len) ||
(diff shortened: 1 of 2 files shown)


carlosmn:
Unfortunately the larger benchmark which also uses diff actually slows down, which I can't explain. This needs some investigation into that before accepting it.

src/tree.h
```diff
@@ -12,17 +12,20 @@
 #include "odb.h"
 #include "vector.h"
 #include "strmap.h"
+#include "pool.h"
 
 struct git_tree_entry {
 	uint16_t attr;
 	git_oid oid;
+	bool pooled;
```
vmg: This mis-aligns the struct packing, which is not neat for pool allocations. I think we could very easily change `filename_len` to `uint16_t` and get a couple extra bytes to play with.

carlosmn: It does indeed bring sadness to alignment, I forgot to double-check after testing it out. It's probably enough to have a `uint16_t` as length for anything we would actually care to support, tbh (and lol if you actually need the 64 bits for your path length).

carlosmn: I changed to 16-bit and avoided any padding (except for the final one due to the zero-length array). This brings the size down from 32 to 26 bytes. It doesn't seem to make a difference in speed one way or another.

carlosmn:
Both the tree-read microbench and the larger diff test are sped up, so I'm happy to merge this as-is.

ethomson:
Out of curiosity, why did the larger benchmark slow down in the intermediate changes?

carlosmn:
I'm not sure; and I seem to have misplaced the specific benchmark which showed a slowdown. I was probably doing something dumb, as it was still just loading a commit's tree and diffing with the first parent's tree, which I just saw (expectedly) speed up.

vmg:
I like the struct re-ordering. :+1:

src/tree.c
```diff
@@ -81,6 +81,30 @@ int git_tree_entry_icmp(const git_tree_entry *e1, const git_tree_entry *e2)
...
+/**
+ * Allocate a tree entry, borrowing the filename from the tree which
+ * owns it. This is useful when reading trees, so we don't allocate a
+ * ton of small strings but can use the pool.
+ */
+static git_tree_entry *alloc_entry_pooled(git_pool *pool, const char *filename, size_t filename_len)
+{
+	git_tree_entry *entry = NULL;
+	size_t tree_len;
+
```
ethomson: Can we add a check here to make sure that `filename_len` fits in a `uint16_t`?  I don't see any obvious vulnerabilities if one were to somehow build a tree that had a longer filename in it, but I'm also not the most creative person in the owrld.

carlosmn: Absolutely, I've pushed up a commit which does this and de-duplicates the size and overflow checks.

pks-t:
I'd be interested as to how PR #3527 affects your benchmarks. These changes introduced quite a big memory leak and we did not call `git__free` for a lot of tree entries, so this _might_ affect the outcome of your timings.

carlosmn:
I'm not sure if we dup tree entries during a diff, but I'll definitely check it out when I'm back on my desktop. When I've fixed other leaks in this area the benchmarks did improve, though.

carlosmn:
Without the leak, Debug goes down to 9.3s; in Release it doesn't seem to make a noticeable difference.

PR #3512, opened 2015-11-17
Protect windows SYSDIR when running tests

ethomson:
Protect the windows SYSDIR when running tests, like we protect the other configuration levels, setting their paths to our sandbox path.


carlosmn:
Yeah... so it turns out that we do write out to the actual `%PROGRAMDATA%` during the test for that level. That test needs to be adjusted so it uses the path we configure in the library instead.

ethomson:
Yeah I just noticed this - fixing it now.

PR #3513, opened 2015-11-18
Recursive Merge

ethomson:
When we built merge, we didn't do recursive base-building.  We thought it would be super critical (after all, `git-merge-recursive` is the default for git itself) but here we are several years later and we're just now getting to it.  We have a customer who does a lot of merging master into their feature branches and then merging back and end up with crazy criss-crosses.  And despite our general disapproval with this workflow, it's something that git does better than libgit2, so here we are.

Conceptually this is quite straightforward:  we still do a three-way merge, but instead of looking for a single common ancestor for two commits, now we look for *all* the merge bases.  If there's more than one, we merge them and use the resultant tree as a merge base.  (Of course, *those* merge bases could have multiple merge bases themselves, hence the name "recursive merge").

But that's not the only way you could end up recursing to build a base - you could also have three merge bases for a pair of commits.  In this case, you need to merge the first two, creating a "virtual merge base", then merge *that* with the third to create a final virtual merge base that you can use as the actual common ancestor.

Because of this, you can't keep only the resultant merge data (the index), you actually need to build a "virtual commit", so that you can keep ancestry (so that you can find the merge base between the new virtual merge base and the third real commit).

My initial thought here was to spin up an in-memory ODB backend (using vmg's mempack backend) and put these virtual commits into that ODB backend.  However, this turns out to be pretty unfortunate when you end up trying to write the index into trees, which is super unnecessary work.

So instead, we make `git_annotated_commit` have a "virtual" mode, that contains only parents and a `git_index` that represents the contents, and we merge those.

Remaining work:
- [x] Recursion limit.  At present, there is no limit on the recursion.  On a very large tree, there should be limits and we should either fail or just stop recursing and just use the first merge base.
- [x] Conflicts merging the bases to create a virtual merge base.  This is just going to be truly godawful.

This remaining work is pretty likely to be truly terrible because they're places where git itself assumes that it's going to crap its bad output right out into the working directory with terrible conflict markers and assume that everything will be okay.


ethomson:
Well, handling conflicts was less bad than I expected.  My bias against it was because I use `diff3` style merges, and conflicts in the recursive base are horrible in that case, but otherwise, it does seem to work out quite neatly.

I added some private flags here.  I considered making this a public setting (the stage-conflicts file favor type).  But then the labels for ancestor / ours / theirs would likely also need to be public (in the merge options).  This has a really nice advantage that we can now get rid of the `git_merge_file_options` type and just use `git_merge_options` everywhere.  But is has the *disadvantage* that the `git_merge` function takes a `git_merge_options` that specifies these labels...  and a `git_checkout_options` that *also* specifies these labels.

And the `git_checkout_options` one is likely the one that will win (since it's actually writing conflict files - the `git_merge_options` would actually only be used when building the recursive merge base!  Very confusing.

So although it seemed like quite the right approach initially, I decided it was not and did this quiet private flags approach that worked out quite cleanly in the end.

include/git2/merge.h
```diff
@@ -152,7 +160,7 @@ typedef enum {
 
 	/** Take extra time to find minimal diff */
 	GIT_MERGE_FILE_DIFF_MINIMAL = (1 << 7),
-} git_merge_file_flags_t;
+} git_merge_file_flag_t;
```
carlosmn: This name change should also be reflected in the CHANGELOG.

src/merge.c
```diff
@@ -1922,6 +1945,202 @@ int git_merge_trees(
...
+	const git_annotated_commit *one,
+	const git_annotated_commit *two,
+	const git_merge_options *opts,
+	size_t recursion_level)
+{
+	git_array_oid_t head_ids = GIT_ARRAY_INIT;
+	git_oidarray bases = {0};
+	git_annotated_commit *base = NULL, *other = NULL, *new_base = NULL;
+	unsigned int recursion_limit = (opts && opts->recursion_limit) ?
+		opts->recursion_limit : DEFAULT_RECURSION_LIMIT;
```
carlosmn: The typical way we have default options is to initialize it to the default in the `_INIT` macro. I don't see that this is different from the other default args.

src/annotated_commit.h
```diff
@@ -7,11 +7,31 @@
...
+ * "real" and backed by an actual commit in the repository, but merge will
+ * internally create "virtual" commits that are in-memory intermediate
+ * commits backed by an index.
+ */
 struct git_annotated_commit {
+	git_annotated_commit_t type;
+
+	/* real commit */
 	git_commit *commit;
+	git_tree *tree;
```
carlosmn: Why is there a new tree pointer for real annotated commits? Do we need to keep it around so the iterator can borrow the value?

src/annotated_commit.h
```diff
@@ -7,11 +7,31 @@
...
 
+#include "oidarray.h"
+
 #include "git2/oid.h"
 
-/** Internal structure for merge inputs */
+typedef enum {
+	GIT_ANNOTATED_COMMIT_REAL = 0,
+	GIT_ANNOTATED_COMMIT_VIRTUAL = 1,
+} git_annotated_commit_t;
```
carlosmn: I'm wondering whether we should have 0 be an invalid value and have 1 and 2 as real and virtual resp. So we can detect in cleanup functions that we failed to set a value.

src/merge.c
```diff
@@ -1922,6 +1945,202 @@ int git_merge_trees(
...
+	result = git__calloc(1, sizeof(git_annotated_commit));
+	GITERR_CHECK_ALLOC(result);
+
+	/* Conflicts in the merge base creation do not propagate to conflicts
+	 * in the result; the conflicted base will act as the common ancestor.
+	 */
+	if (opts)
+		memcpy(&virtual_opts, opts, sizeof(git_merge_options));
+
+	virtual_opts.flags &= ~GIT_MERGE_FAIL_ON_CONFLICT;
```
carlosmn: Even if we don't consider a conflict in the common merge base to be a "real" conflict, is there a chance that having a conflict (and thus conflict markers) in the common base isn't going to result in a conflict for the later merge?

ethomson: Definitely - in the case where both branches changed to be identical.  In which case you would have a virtual merge base with conflict markers, and both ours and theirs are identical, we will *not* produce a conflict and the new version will be accepted as the merge result.

I have thought about this and poked and prodded to try to come up with other cases where we *should* produce conflicts but would not.  I have not yet come up with any.

src/annotated_commit.h
ethomson: Yes, that's correct.

include/git2/merge.h
ethomson: Updated.

src/annotated_commit.h
ethomson: Makes sense.  Updated.

src/merge.c
ethomson: So upon further reflection, I dropped the default recursion limit because it seems a bit unnecessary for most uses.  I also cleaned this up a bit.

include/git2/merge.h
```diff
@@ -181,8 +189,8 @@ typedef struct {
 	/** The file to favor in region conflicts. */
 	git_merge_file_favor_t favor;
 
-	/** see `git_merge_file_flags_t` above */
-	unsigned int flags;
+	/** see `git_merge_file_flag_t` above */
+	git_merge_file_flag_t flags;
```
jacquesg: Multiple options from ```git_merge_file_flag_t``` may be specified, that is the diff style, whitespace options etc.

ethomson: Yes, and this doesn't preclude that.  This is to match the style in the rest of the library where enum types are named in the singular, even when they will be combined.  For example:  `git_repository_open_flag_t`.

PR #3514, opened 2015-11-20
Stat fixes

jacquesg:
This PR fixes some issues related to the detection of ```stat```'s members. There is no real consensus between systems or even different versions of the same system when it comes to ```st_atim```/```st_atimespec``` and friends.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,9 +86,20 @@ IF (NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 	OPTION( USE_OPENSSL                     "Link with and use openssl library"             ON )
 ENDIF()
 
-CHECK_STRUCT_HAS_MEMBER("struct stat" st_mtim.tv_nsec sys/stat.h
-	HAVE_STRUCT_STAT_NSEC LANGUAGE C)
-IF(HAVE_STRUCT_STAT_NSEC OR WIN32)
+CHECK_STRUCT_HAS_MEMBER ("struct stat" st_atim "sys/types.h;sys/stat.h"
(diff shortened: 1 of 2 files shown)


jacquesg:
One report illustrating the problem:

http://www.cpantesters.org/cpan/report/c73e64ea-8b1a-11e5-afef-e7dd1d056eec

CMakeLists.txt
```diff
@@ -86,9 +86,20 @@ IF (NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
...
-IF(HAVE_STRUCT_STAT_NSEC OR WIN32)
+CHECK_STRUCT_HAS_MEMBER ("struct stat" st_atim "sys/types.h;sys/stat.h"
+	HAVE_STRUCT_STAT_ST_ATIM LANGUAGE C)
+CHECK_STRUCT_HAS_MEMBER ("struct stat" st_atimespec "sys/types.h;sys/stat.h"
+	HAVE_STRUCT_STAT_ST_ATIMESPEC LANGUAGE C)
+
+IF (HAVE_STRUCT_STAT_ST_ATIM)
+	CHECK_STRUCT_HAS_MEMBER("struct stat" st_mtim.tv_nsec sys/stat.h
+		HAVE_STRUCT_STAT_NSEC LANGUAGE C)
+ELSE (HAVE_STRUCT_STAT_ST_ATIMESPEC)
```
carlosmn: It's surprising that this works. The 'else' having a different expression from the if is supposed to generate an error.

jacquesg: Yeah, thats not right, fixed.

ethomson:
Much nicer than the `__APPLE__` ifdef.  Thanks!

PR #3515, opened 2015-11-20
Fix <0 unsigned comparison (stat.st_size should be an off_t)

jacquesg:
Turns out that the error code path will never be entered seeing that assigning an ```int``` of value ```-1``` to a ```uint64_t``` will give you 18446744073709551615.


ethomson:
POSIX defines `st_size` as an `off_t`.  We probably should, too.

jacquesg:
That actually seems more reasonable, updated.

ethomson:
Thanks!

PR #3518, opened 2015-11-20
Cleanup of #ifdefs that split up parts of statements and fixing a small leak.

flaviommedeiros:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

- https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
- https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness. At the same time, fixing a small memory leak.

### DIFF
--- a/deps/regex/regcomp.c
+++ b/deps/regex/regcomp.c
@@ -3026,7 +3026,7 @@ parse_bracket_exp (re_string_t *regexp, re_dfa_t *dfa, re_token_t *token,
 #endif /* not RE_ENABLE_I18N */
   int non_match = 0;
   bin_tree_t *work_tree;
-  int token_len;
+  int token_len, error_return;
   int first_round = 1;
 #ifdef _LIBC
   collseqmb = (const unsigned char *)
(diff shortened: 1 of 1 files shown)


ethomson:
Hi!  Thanks for the contribution.  However, the regex implementation is lifted directly from the GNU C Library.

Since it is a direct dependency, and not part of *our* code, we are not accepting changes directly.

If you have suggestions or improvements for the GNU C Library, we'd kindly ask that you make those changes in that project.  We will then include those changes once they have been code reviewed and accepted by those feature experts.

Thanks!

RomeroMalaquias:
Hello @ethomson   i'm from the same research team of flaviommedeiros  and i know that code is part of the GNU C Library, but i have to ask if you prefer this code before or after the commit. What do you think about this kind of change? 
Thanks in advance.

ethomson:
I prefer the change _after_, I definitely find it more readable.

Issue #3440, opened 2015-09-26
The certificate callback should support pass-through


carlosmn:
This callback wants a response from the callback; if we have none to give (we're wrapping the callbacks in some language, so we always end up setting the fields) we should be able to tell libgit2 not to take our answer into account.

Returning `GIT_PASSTHROUGH` should make the callers of the callback behave as though it was not there, like they do with the credentials callback.

SantiagoTorres:
Hi, @carlosmn. I'd like to work on this ticket. What is a good place to start looking into it?
I'm already familiar with some of the codebase, but I don't seem to pinpoint:

* which is the credentials callback? I'm looking at [this](https://libgit2.github.com/libgit2/#HEAD/group/cred/git_cred_userpass). Or is [this](https://github.com/libgit2/libgit2/blob/a8d0b4de92b83745308b2d9af0f62828bfb550a1/src/transports/ssh.c#L418) the right one?
* which is the certificate callback? 

Thanks!

carlosmn:
The credentials callback is a field in the `git_remote_callbacks` structure, passed in their respective options to fetch and push. The transports are then passed in the callbacks via a setter function. For SSH, the call to it is done in [line 426](https://github.com/libgit2/libgit2/blob/a8d0b4de92b83745308b2d9af0f62828bfb550a1/src/transports/ssh.c#L426). The HTTP backends do something similar.

The certificate callback is another one of the remote callback fields, called by the transports (or their underlying IO streams). Each of the connection methods calls this certificate when they receive the server's certificate (or hostkey for SSH). E.g. in (line 558)[https://github.com/libgit2/libgit2/blob/a8d0b4de92b83745308b2d9af0f62828bfb550a1/src/transports/ssh.c#L558]. It currently returns an error for any error value returned by the callback. It should instead check if it's `GIT_EPASSTHROUGH` and then do what it would do if the callback weren't set.

The OpenSSL, SecureTransport and WinHTTP code also calls the certificate callback, so those need to be converted as well.

tiennou:
Closing this, as a fix for this has landed via #4879.

PR #3519, opened 2015-11-22
[examples/log.c]: Adds --log-size function

SantiagoTorres:
Hello, everyone.

I wanted to play around with the library and saw these starter-project suggestions (I'm also interested in getting #3440 but that might take a little longer). 

I tried to keep everything pretty self-contained and straightforward, but I could see why some design decisions (e.g., passing the whole options structure to the print commit function) might not be appealing to everyone. 

Let me know how this looks.

### DIFF
--- a/examples/log.c
+++ b/examples/log.c
@@ -52,6 +52,7 @@ struct log_options {
 	int show_diff;
 	int skip, limit;
 	int min_parents, max_parents;
+	int show_log_size;
 	git_time_t before;
 	git_time_t after;
 	const char *author;
@@ -63,7 +64,7 @@ struct log_options {
 static int parse_options(
 	struct log_state *s, struct log_options *opt, int argc, char **argv);
 static void print_time(const git_time *intime, const char *prefix);
-static void print_commit(git_commit *commit);
+static void print_commit(git_commit *commit, struct log_options options);
 static int match_with_parent(git_commit *commit, int i, git_diff_options *);
 
 /** utility functions for filtering */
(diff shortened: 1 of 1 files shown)


examples/log.c
```diff
@@ -337,17 +338,57 @@ static void print_time(const git_time *intime, const char *prefix)
...
 	git_oid_tostr(buf, sizeof(buf), git_commit_id(commit));
 	printf("commit %s\n", buf);
 
-	if ((count = (int)git_commit_parentcount(commit)) > 1) {
+	sig = git_commit_author(commit);
+	scan = git_commit_message(commit);
+	count = (int)git_commit_parentcount(commit);
+
+	if (options.show_log_size) {
+
```
pks-t: You're using newlines quite excessively. It makes sense to group logical blocks together by whitespaces, but we usually avoid using empty lines directly after a block-opening brace. Furthermore, we normally don't use two successive empty lines

examples/log.c
```diff
@@ -337,17 +338,57 @@ static void print_time(const git_time *intime, const char *prefix)
...
+
+		/* simulate the message-identation */
+		while (scan && *scan) {
+
+			if (*scan == '\n')
+				length += 4;
+			scan++;
+
+		}
```
pks-t: All of the above looks quite a lot like guessworking. Wouldn't it be a lot easier if you just addded the return value of all those `printf`s to `length`? That'd match exactly all the bytes we're outputting, and make the logic much easier to follow.

ethomson:
This PR is now a few years old; closing this — please do open a new one if you're interested in making changes. 🙏

PR #3521, opened 2015-11-24
Line count overflow in git_blame_hunk and git_blame__entry

### LINKED ISSUE #3505: [Bug] Incorrect blame line numbers and hunk sizes for very large files
I'm using Rugged to store large files (over 100.000 lines). When I perform a blame on these files all hunk sizes and line numbers are wrapped (modulo) to 65536 (2^16) as they are stored in an unsigned 16 bit integer in libgit2 (rugged/vendor/libgit2/src/blame.c (search for uint16_t)).

I realize that this use-case is pretty far out there as files over 65536 lines are uncommon but I would really appreciate any help on this. The files I'm storing are professional telecom router configuration backups and they simply are this large: I did not generate them myself.

I also reported this to the Rugged issue tracker at: https://github.com/libgit2/rugged/issues/547

pks-t:
This fixes #3505 and libgit2/rugged#547.

I guess this shouldn't be part of a bugfix-only release as the changes break the ABI.


include/git2/blame.h
```diff
@@ -113,15 +113,15 @@ GIT_EXTERN(int) git_blame_init_options(
...
 
 	git_oid final_commit_id;
-	uint16_t final_start_line_number;
+	uint32_t final_start_line_number;
 	git_signature *final_signature;
 
 	git_oid orig_commit_id;
 	const char *orig_path;
-	uint16_t orig_start_line_number;
+	uint32_t orig_start_line_number;
```
ethomson: Why stop at `uint32_t`?  Why not make this a `size_t`?

src/blame_git.c
```diff
@@ -93,10 +93,10 @@ static bool same_suspect(git_blame__origin *a, git_blame__origin *b)
 }
 
 /* find the line number of the last line the target is suspected for */
-static int find_last_in_target(git_blame *blame, git_blame__origin *target)
+static int64_t find_last_in_target(git_blame *blame, git_blame__origin *target)
```
ethomson: Does this need to be signed or are we overloading the line number with the error code?  Why not:

    static int find_last_in_target(size_t *out, git_blame *blame, git_blame__origin *target)

pks-t:
Changes according to  @ethomson's comments:

* uint32_t -> size_t
* improved signature of `find_last_in_target` so that we do not have to overload the return value for errors

src/blame.c
```diff
@@ -223,8 +223,8 @@ static git_blame_hunk *split_hunk_in_vector(
 	}
 
 	new_line_count = hunk->lines_in_hunk - rel_line;
-	nh = new_hunk((uint16_t)(hunk->final_start_line_number+rel_line), (uint16_t)new_line_count,
-			(uint16_t)(hunk->orig_start_line_number+rel_line), hunk->orig_path);
+	nh = new_hunk((size_t)(hunk->final_start_line_number+rel_line), (size_t)new_line_count,
+			(size_t)(hunk->orig_start_line_number+rel_line), hunk->orig_path);
```
ethomson: I think that these casts are now unnecessary.

src/blame.c
```diff
@@ -233,7 +233,7 @@ static git_blame_hunk *split_hunk_in_vector(
 	git_oid_cpy(&nh->orig_commit_id, &hunk->orig_commit_id);
 
 	/* Adjust hunk that was split */
-	hunk->lines_in_hunk -= (uint16_t)new_line_count;
+	hunk->lines_in_hunk -= (size_t)new_line_count;
```
ethomson: I think that this cast is not necessary any longer.

src/blame.c
```diff
@@ -442,7 +442,7 @@ static int buffer_line_cb(
 		} else {
 			/* Create a new buffer-blame hunk with this line */
 			shift_hunks_by(&blame->hunks, blame->current_diff_line, 1);
-			blame->current_hunk = new_hunk((uint16_t)blame->current_diff_line, 1, 0, blame->path);
+			blame->current_hunk = new_hunk((size_t)blame->current_diff_line, 1, 0, blame->path);
```
ethomson: This cast, too, I think.

pks-t:
Removed those casts. There were some more useless casts I've removed.

ethomson:
:sparkles: Thanks for this!

PR #3522, opened 2015-11-24
diff: include commit message when formatting patch

### LINKED ISSUE #3371: When formatting patch as mbox, commit.message isn't added
When trying to format a commit as an email, libgit2 will not add the commit.message to the output leaving out rather crucial information about what author thought s/he was doing.

The original git format-patch adds the commit message to mbox, so users have come to expect that when generating mbox versions of commits.

I have written a patch (yet to test, only build-tested so far) which would add commit.message to the final mbox, after testing I could send a pull request, assuming the community agrees with the need.

```Diff
    diff: add commit.message to git_diff_commit_as_email()
    
    The original git format-patch command, will also
    add git commit message, in order to be compatible
    with users' expectations, let's also add commit.message
    when generating commit as email.
    
    Signed-off-by: Felipe Balbi <balbif@gmail.com>

diff --git a/include/git2/diff.h b/include/git2/diff.h
index 0abbc7f06f85..cd8389170969 100644
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -1263,6 +1263,9 @@ typedef struct {
        /** Summary of the change */
        const char *summary;
 
+       /** Commit message */
+       const char *message;
+
        /** Author of the change */
        const git_signature *author;
 } git_diff_format_email_options;
diff --git a/src/diff.c b/src/diff.c
index 44f62788086f..759c1a5de626 100644
--- a/src/diff.c
+++ b/src/diff.c
@@ -1743,6 +1743,8 @@ int git_diff_format_email(
        if (error < 0)
                goto on_error;
 
+       error = git_buf_puts(out, opts->message);
+
        format_flags = GIT_DIFF_STATS_FULL | GIT_DIFF_STATS_INCLUDE_SUMMARY;
 
        if ((error = git_buf_puts(out, "---\n")) < 0 ||
@@ -1782,6 +1784,7 @@ int git_diff_commit_as_email(
        opts.id = git_commit_id(commit);
        opts.summary = git_commit_summary(commit);
        opts.author = git_commit_author(commit);
+       opts.message = git_commit_message(commit);
 
        if ((error = git_diff__commit(&diff, repo, commit, diff_opts)) < 0)
                return error;
```

pks-t:
When formatting a patch as email we do not include the commit's
message in the formatted patch output. Implement this and add a
test that verifies behavior.

Fixes #3371.


src/diff.c
```diff
@@ -1863,6 +1872,16 @@ int git_diff_commit_as_email(
 	opts.summary = git_commit_summary(commit);
 	opts.author = git_commit_author(commit);
 
+	/* The first line of the commit message is identical to
+	 * the summary and thus needs to be skipped */
+	if ((ptr = git_commit_message(commit)) != NULL) {
+		/* Find end of first line */
+		ptr = strpbrk(ptr, "\r\n");
```
ethomson: This seems strange to me.  Why are we looking for `\r\n` and not `\n` as a line terminator?

carlosmn: Fun fact: the second argument to `strpbrk()` is a list of chars to find, in string form (because how else would you represent a list of characters ;)

carlosmn: However, note that what we want to find is the end of the first paragraph, not the first line. That is, we're looking for the first `"\n\n"` (or, I suppose maybe also `"\r\n\r\n"`, though I don't know if any editor will actually use CRLF for the commit message) as a separator between the subject and the body.

ethomson: Right you are.  In that case, why are we looking for `\r`?  Does Git actually allow `\r` as a line terminator?  Or is the goal to find the `\n` but back up over the `\r` if the line was terminated with `\r\n`?

pks-t: Well, actually I'm not sure about this. Just searched git-am(1), git-format-patch(1) and git-apply(1), but there was no mention of line endings. There are some platforms that use `\r` as line endings, that's why I used this.

I think it might make sense to expose a new function `git_commit_body`, that will return a pointer to the commit message excluding the summary. This could be used here, then, and might prove useful for others.

pks-t:
As proposed in my comment I've created a new public function `git_commit_body`, which is used by the format-email code.

PR #3537, opened 2015-12-06
tree: mark a tree as already sorted

carlosmn:
The trees are sorted on-disk, so we don't have to go over them
again. This cuts almost a fifth of time spent parsing trees.


linquize:
Together with another earlier commit,  we will have more improvements to boost tree parsing performance on the next release.  Right?

Issue #3541, opened 2015-12-09
Reset fails on file to folder transition


vmg:
Pretty easy to reproduce. Have a repo that looks like this:

```
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391    README
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391    folder
```

Then remove the file called `folder` and replace it with an actual folder.

```
100644 blob e69de29bb2d1d6434b8b29ae775ad8c2e48c5391    README
100644 blob 03ed41df1178c190a425bcbc5ba9363acbd5533b    folder/FILE
```

Calling `git_reset(HARD)` through libgit2 will fail with `folder/FILE appears as both a file and a directory`.

How do we work around this? We're calling `git_read_tree` into the existing index after the checkout, without clearing the existing index. Is this intended? `read_tree` eventually clears the index, but _after_ the entries have been loaded into a temporary vector.

cc @ethomson @carlosmn @ptoomey3

vmg:
Actually, `git_read_tree` is a red herring. The error happens earlier during the checkout:

```
#0  index_insert (index=index@entry=0x555556036ac0, entry_ptr=entry_ptr@entry=0x7fffffffcd70, trust_path=trust_path@entry=true, 
    trust_mode=trust_mode@entry=true, replace=1) at /home/vmg/src/rugged/vendor/libgit2/src/index.c:1316
#1  0x00007ffff6145e36 in git_index_add (index=0x555556036ac0, source_entry=source_entry@entry=0x7fffffffcda0)
    at /home/vmg/src/rugged/vendor/libgit2/src/index.c:1558
#2  0x00007ffff6113474 in checkout_update_index (file=0x55555603e4f0, st=<optimized out>, data=<optimized out>)
    at /home/vmg/src/rugged/vendor/libgit2/src/checkout.c:1576
#3  0x00007ffff6115e90 in checkout_blob (file=<optimized out>, data=<optimized out>)
    at /home/vmg/src/rugged/vendor/libgit2/src/checkout.c:1738
#4  checkout_create_the_new (data=<optimized out>, actions=<optimized out>)
    at /home/vmg/src/rugged/vendor/libgit2/src/checkout.c:1842
#5  git_checkout_iterator (target=0x1, index=0x55555603d0e0, opts=0x7fffffffd338)
    at /home/vmg/src/rugged/vendor/libgit2/src/checkout.c:2566
[...]

PR #3542, opened 2015-12-09
reset: perform the checkout before moving HEAD or the index

carlosmn:
This keeps the state of the workdir the same as one from HEAD, removing
a source of possible confusion when calculating the work that is to be
done.

This should handle  #3541 and bring us close to what we generally should be doing when switching the current commit.

/cc @vmg

### DIFF
--- a/src/reset.c
+++ b/src/reset.c
@@ -145,18 +145,18 @@ static int reset(
 	if ((error = git_buf_printf(&log_message, "reset: moving to %s", to)) < 0)
 		return error;
 
-	/* move HEAD to the new target */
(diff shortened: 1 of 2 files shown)


vmg:
Ugh, just like you told me in chat... I wrote this myself and it turns out my build was broken because of unrelated reasons. :sweat:

The patch looks correct. Thanks Carlos. :o

@ethomson: any last thoughts?

vmg:
Oh shit, I guess @ethomson is out sabbaticating already! @carlosmn: Good to merge when you're ready. :)

PR #3544, opened 2015-12-10
CMake: activate the rpath policy for OSX

carlosmn:
This activates the newer policy of using rpath in the build and shuts
CMake up about it when we're building.

/cc @Therzok I figure you'd at least know whether this impacts people trying to use the library.


Therzok:
Okay so, this _shouldn't_ break anything. It will just generate relative paths for the linker to look in instead of using absolute paths (the default before this).

I can't try this atm, but I'll look tomorrow to see what `otool -L` says about the generated paths before and after this PR.

Looks good to me otherwise.

Therzok:
Okay, so I've just taken a look at this, luckily I stumbled upon it:

```
	@rpath/libgit2.23.dylib (compatibility version 23.0.0, current version 0.23.0)
```

This is what the linker path would look like. So, instead of looking for the identity library - aka libgit2 itself - in an absolute path /path/to/libgit2, and that wouldn't load libgit2 unless it was found in that path.

tl;dr this won't hurt, :+1:

carlosmn:
OK, then we'd just need to make this version-dependent.

ethomson:
Closed via #4513

PR #3547, opened 2015-12-15
refdb: Add an API git_refdb_get_backend

pmq20:


ethomson:
Can you explain why this API would be useful?  If you're setting a custom refdb backend, I would expect you to keep a pointer to that, and not need to query it from libgit2...?

pmq20:
@ethomson Thanks for reviewing. When implementing the `next` function of `git_reference_iterator`, we would love to get the backend-pointer out of `git_reference_iterator`'s `db` field.

carlosmn:
if you're writing your own backend and iterator code, you're in control of which struct the iterator is backed by and can store whatever information you need there. If you need a pointer to the backend, you should store that pointer in your iterator struct, which all the iterator methods receive.

ethomson:
I agree with @carlosmn - we would ask you to maintain this pointer yourself.  You should be able to simply create your `git_reference_iterator` as something like:

    struct my_custom_iterator {
        git_reference_iterator base;
        my_custom_refdb *refdb;
    }

And when you populate your custom iterator, you can put a link to your custom refdb.

PR #3549, opened 2015-12-16
merge: Use `git_index__fill` to populate the index

vmg:
See the commit message for details. This fixes a performance issue when merging _truly_ large trees (roughly 1.5 million entries).

cc @carlosmn @ethomson (HAH how's that for a sabbatical)
cc @piki @simonsj

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -1542,6 +1542,43 @@ int git_index_remove_bypath(git_index *index, const char *path)
 	return 0;
 }
 
+int git_index__fill(git_index *index, const git_vector *source_entries)
+{
+	const git_index_entry *source_entry = NULL;
+	size_t i;
+	int ret = 0;
+
+	assert(index);
+
+	if (git_mutex_lock(&index->lock) < 0) {
+		giterr_set(GITERR_OS, "Unable to acquire index lock");
+		return -1;
+	}
+
+	git_vector_foreach(source_entries, i, source_entry) {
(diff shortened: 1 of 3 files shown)


ethomson:
:+1:

vmg:
Damn, that was some quick code review hmmpf.

vmg:
By hmmmpf I meant that I was not sure at all of the quality of this implementation (am I right on my assumptions about entries not colliding and being valid?).

Guess it's time to blindly trust @ethomson again :D

carlosmn:
This is in a new index, so we won't collide with any old entries. And the staged/successfully-merged entries would only have a single name, so I figured it was fine.

vmg:
Ayup, I think I agree. I have however noticed that we're not adjusting the length mask on the flags of the index...  It doesn't seem like the length mask is used anywhere, but we should probably adjust it nonetheless. And we should also perform filemode normalization, just in case the original filemode from the tree is not valid.

peff:
Nice catch. Seems like a classic "accidentally quadratic" case.

If you were really worried about collisions, you should be able to do an `O(n)` scan of the vector after the sort. I'm not sure how concerned to be, since I'm not sure where the entries are coming from. What if you had a tree with duplicate entries? That's not _supposed_ to happen, and it's going to produce a slightly-off answer in the name of speed (at least, that's Git's attitude). As long as it wouldn't send us into an infinite loop or anything. :)

Issue #2926, opened 2015-02-26
git_merge() / git_index_add() errors with "invalid filemode" on 0664 mode file


jeffhostetler:
git_index_add() errors with "invalid filemode" on files with mode 0664.
(It does accept files with modes 0644 and 0755.)

This issue came up during a call to git_merge() on Windows.

```
libgit2_clar.exe!git_index_add(git_index * index, const git_index_entry * source_entry) Line 1151
libgit2_clar.exe!checkout_conflict_add(checkout_data * data, const git_index_entry * conflict) Line 2051
libgit2_clar.exe!checkout_conflict_update_index(checkout_data * data, checkout_conflictdata * conflict) Line 2061
libgit2_clar.exe!checkout_create_conflicts(checkout_data * data) Line 2138
libgit2_clar.exe!git_checkout_iterator(git_iterator * target, git_index * index, const git_checkout_options * opts) Line 2469
libgit2_clar.exe!git_checkout_index(git_repository * repo, git_index * index, const git_checkout_options * opts) Line 2530
libgit2_clar.exe!git_merge(git_repository * repo, const git_annotated_commit * * their_heads, unsigned int their_heads_len, const git_merge_options * merge_opts, const git_checkout_options * given_checkout_opts) Line 2695
```

In this test:
* git clone https://github.com/git/git.git gitgit
* cd gitgit
* git checkout -b HACK e83c516   *(the initial commit)*
* echo hello > hello.txt
* git commit -a -m "Added hello"
* call git_merge() to effectively do: "git merge master"

On Windows, this exits with the invalid filemode.  Odd that it is in the conflict code at the time.  I haven't tried it on other platforms yet.

ethomson:
> Odd that it is in the conflict code at the time.

IIRC, when building the index for merge, we skip the `git_index_add` (and just build a collection of `git_index_entry`s ourself.)  So if you had a damaged index entry, we would simply maintain it.  This seems like the right thing to do, lest we run into errors like this.  ;)

ethomson:
Without even looking at the code, I think that conflicts should call an internal index add function that eschews sanity checks, while the public `git_index_add` should maintain those checks.

We could also make a public API that does this.  I can imagine that a consumer of the library might want to maintain their junky data in the same way.  (Maybe they're writing their own git-merge-whatever.)  Or we could punt on that for later.

carlosmn:
[...]

PR #3550, opened 2015-12-16
Index fill: Small fixups

vmg:
- Adjust the namemask for the index entry (just in case!)
- Normalize the filemode (also just in case!)
- Pre-allocate the entries array (since we already know the final size!)

cc @carlosmn @ethomson

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -879,6 +879,18 @@ void git_index_entry__init_from_stat(
 	entry->file_size = st->st_size;
 }
 
+static void index_entry_adjust_namemask(
+		git_index_entry *entry,
+		size_t path_length)
+{
+	entry->flags &= ~GIT_IDXENTRY_NAMEMASK;
+
+	if (path_length < GIT_IDXENTRY_NAMEMASK)
+		entry->flags |= path_length & GIT_IDXENTRY_NAMEMASK;
+	else
+		entry->flags |= GIT_IDXENTRY_NAMEMASK;
+}
+
 static int index_entry_create(
 	git_index_entry **out,
 	git_repository *repo,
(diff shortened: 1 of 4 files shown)


carlosmn:
The "normal" and `__fill` versions of adding to the struct and hashmap look rather similar now, don't they? I wonder if makes sense to have a function do that and then use it in each of the functions.

src/index.c
```diff
@@ -1550,21 +1556,30 @@ int git_index__fill(git_index *index, const git_vector *source_entries)
...
+
 	if (git_mutex_lock(&index->lock) < 0) {
 		giterr_set(GITERR_OS, "Unable to acquire index lock");
 		return -1;
 	}
 
+	if (git_vector_size_hint(&index->entries, source_entries->length) < 0) {
+		git_mutex_unlock(&index->lock);
+		return -1;
+	}
```
carlosmn: We should set the size of the hashmap here as well. Resizing that is a lot more expensive than resizing the vector (because of all the hashing in the hashmap).

vmg:
@carlosmn: I think there's quite a bit of "extra" boilerplate in the normal codepath which is not easy to abstract. At least my attempt was not coming out clean at all. If you wanna give it a go...

src/index.c
vmg: Ah very good point. :+1:

ethomson:
Does git correct modes when faced with an odd one?  I thought it would preserve strange modes.

(Certainly we should not be falling like we were before though.)

Related #2926

vmg:
@carlosmn: I've added the hash table resize!
@ethomson: Good question, I'm not sure. Maybe @peff can chime in, but I feel if we're getting a bad file mode, normalizing it (since generally we can always normalize filemodes) for the output index is the right thing to do.

peff:
Git will normalize modes it reads from trees (and elsewhere). The code is in [canon_mode](https://github.com/gitster/git/blob/d81279c8ed76b556017c37b679cf773fbfd2b5c0/cache.h#L299-L308). That gets called at the lowest level of the tree reader (in `decode_tree_entry`), and I think we should do the same when picking up modes from the filesystem.

PR #3552, opened 2015-12-17
Update README URLs based on HTTP redirects

ReadmeCritic:
Created with https://github.com/dkhamsing/frankenstein

### GitHub Corrected URLs 
Was | Now 
--- | --- 
https://github.com/fpco/gitlib | https://github.com/jwiegley/gitlib 
https://github.com/tbranyen/nodegit | https://github.com/nodegit/nodegit 

### Other Corrected URLs 
Was | Now 
--- | --- 
http://libgit2.github.com/libgit2 | http://libgit2.github.com/libgit2/ 
http://www.cmake.org | https://cmake.org/ 
http://www.cmake.org/Wiki/CMake_FAQ | https://cmake.org/Wiki/CMake_FAQ 
https://libgit2.github.com/docs/guides/build-and-link | http://libgit2.github.com/docs/guides/build-and-link/ 
https://live.gnome.org/Libgit2-glib | https://wiki.gnome.org/Projects/Libgit2-glib

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Additionally, the example code has been released to the public domain (see the
 * Website: [libgit2.github.com](http://libgit2.github.com)
 * StackOverflow Tag: [libgit2](http://stackoverflow.com/questions/tagged/libgit2)
 * Issues: [GitHub Issues](https://github.com/libgit2/libgit2/issues) (Right here!)
-* API documentation: <http://libgit2.github.com/libgit2>
+* API documentation: <http://libgit2.github.com/libgit2/>
 * IRC: [#libgit2](irc://irc.freenode.net/libgit2) on irc.freenode.net.
 * Mailing list: The libgit2 mailing list was
     traditionally hosted in Librelist but has been deprecated. We encourage you to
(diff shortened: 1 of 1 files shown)


ethomson:
I'm always bummed out by automated tools opening pull requests without any human intervention.

I guess that I see some value in saving the end user a few redirects - if every `http://www.cmake.org` became `https://cmake.org/` then we could save the entire Internet perhaps tens of dollars per month.

But these sorts of automated tools are stylistically empty and mundane.  Why not just `https://cmake.org`, which was in keeping with the original URI?

But more importantly, just because one URI redirects to another URI - that doesn't mean that the destination is useful at all.  For example, `https://wiki.gnome.org/Libgit2-glib` is a not found page, which is certainly not any more useful than before.

If an actual human had done any of this, they would have obviously given up a precious 2 seconds of thinking and decided that obviously that was idiotic, and that instead the new URI should be `https://wiki.gnome.org/Projects/Libgit2-glib`.

So hopefully, there's actually somebody *listening* to the pull requests that you're opening and responding to them, and this isn't just a spammy robot service.  If you fix up the URIs to actually point *to the right places*, then I'll (begrudgingly) merge this.

ReadmeCritic:
@ethomson updated

ethomson:
Thanks for the update.  I'm glad to see that there's a real human monitoring the issues that are being opened (and a quite responsive human at that.)

I, for one, don't quite welcome my robotic software writing overloads just yet, but I suppose I should be a bit more lax on human / robot hybrids.  ;)

Thanks!

ethomson:
:heart:

ReadmeCritic:
:wink:

PR #3554, opened 2015-12-21
typos in comments

olshevskiy87:


whoisj:
:+1: and wow, none of them are my spelling mistakes :open_mouth:

carlosmn:
The code under `deps/` is not ours so we should not change it at all as it would just make it harder to merge any updates we might do.

whoisj:
> the code under deps/ is not ours 

ahh yeah, glad you're paying attention. I was only looking at the spelling mistakes / fixes -- not the paths. :hushed: Nice catch.

olshevskiy87:
oh.. if it's not yours, then I'm sorry )

carlosmn:
There are three changes or so which do change our code; if you undo the changes to `deps/` and squash, we could merge that.

olshevskiy87:
> if you undo the changes to deps/ and squash

ok. done

PR #3555, opened 2015-12-22
Support for ssh+git and git+ssh protocols

cbargren:
This should add support for the `ssh+git://` and `git+ssh://` protocol schemes. The behavior is to pass directly through to the `ssh://` protocol scheme.

So, as atrocious as this is, this is a protocol that people use. From everything I can tell, they're both functionally identical to `ssh://` and are treated as such in this PR. [Here](http://git.kernel.org/cgit/git/git.git/commit/?id=c05186cc38ca4605bff1f275619d7d0faeaf2fa5) is an example of it being used in ~~the linux kernel~~ git core

### DIFF
--- a/src/transport.c
+++ b/src/transport.c
@@ -35,8 +35,8 @@ static transport_definition transports[] = {
 	{ "file://",  git_transport_local, NULL },
 #ifdef GIT_SSH
 	{ "ssh://",   git_transport_smart, &ssh_subtransport_definition },
-  { "ssh+git://",   git_transport_smart, &ssh_subtransport_definition },
-  { "git+ssh://",   git_transport_smart, &ssh_subtransport_definition },
+	{ "ssh+git://",   git_transport_smart, &ssh_subtransport_definition },
+	{ "git+ssh://",   git_transport_smart, &ssh_subtransport_definition },
 #endif
 	{ NULL, 0, 0 }
 };
--- a/src/transport.c
+++ b/src/transport.c
@@ -35,6 +35,8 @@ static transport_definition transports[] = {
 	{ "file://",  git_transport_local, NULL },
 #ifdef GIT_SSH
 	{ "ssh://",   git_transport_smart, &ssh_subtransport_definition },
+  { "ssh+git://",   git_transport_smart, &ssh_subtransport_definition },
+  { "git+ssh://",   git_transport_smart, &ssh_subtransport_definition },
 #endif
 	{ NULL, 0, 0 }
 };
--- a/src/transports/ssh.c
+++ b/src/transports/ssh.c
@@ -20,8 +20,7 @@
 
 #define OWNING_SUBTRANSPORT(s) ((ssh_subtransport *)(s)->parent.subtransport)
 
-static const char * ssh_prefixes[] = { "ssh://", "ssh+git://", "git+ssh://" };
-#define SSH_PREFIX_COUNT (sizeof(ssh_prefixes) / sizeof(ssh_prefixes[0]))
+static const char *ssh_prefixes[] = { "ssh://", "ssh+git://", "git+ssh://" };
 
 static const char cmd_uploadpack[] = "git-upload-pack";
 static const char cmd_receivepack[] = "git-receive-pack";
@@ -64,9 +63,9 @@ static int gen_proto(git_buf *request, const char *cmd, const char *url)
 {
 	char *repo;
 	int len;
+	size_t i;
 
-	size_t i = 0;
-	for (i = 0; i < SSH_PREFIX_COUNT; ++i) {
+	for (i = 0; i < ARRAY_SIZE(ssh_prefixes); ++i) {
 		const char *p = ssh_prefixes[i];
 
 		if (!git__prefixcmp(url, p)) {
@@ -503,6 +502,7 @@ static int _git_ssh_setup_conn(
 	char *host=NULL, *port=NULL, *path=NULL, *user=NULL, *pass=NULL;
 	const char *default_port="22";
 	int auth_methods, error = 0;
+	size_t i;
 	ssh_stream *s;
 	git_cred *cred = NULL;
 	LIBSSH2_SESSION* session=NULL;
@@ -518,8 +518,7 @@ static int _git_ssh_setup_conn(
 	s->session = NULL;
 	s->channel = NULL;
 
-	size_t i = 0;
-	for (i = 0; i < SSH_PREFIX_COUNT; ++i) {
+	for (i = 0; i < ARRAY_SIZE(ssh_prefixes); ++i) {
 		const char *p = ssh_prefixes[i];
 
 		if (!git__prefixcmp(url, p)) {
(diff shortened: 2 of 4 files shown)


johnhaley81:
I'd like to see this backported to `maint/v0.23` if possible.

carlosmn:
These aren't documented by git to be acceptable formats in its list of URLs in [man git-clone](http://git-scm.com/docs/git-clone).

Additionally, I really don't like having to change external code; this especially true for the HTTP parser since we may not use our copy so this won't be supported if it's installed on the system.

cbargren:
Hm... While they're not documented as acceptable formats, they're both supported in the git CLI. I'm having a really hard time finding out why people use this format in the first place (any amount of searching turns up basically no documentation on the format), but for whatever reason, they do and it works in the CLI so I figured that libgit2 should support it as well.

The HTTP parser is definitely something that I missed as being an external dependancy. I can attempt to PR that change upstream as it is probably something that should be supported by them. It is a part of the [URI Scheme Spec](https://tools.ietf.org/html/rfc3986#section-3.1). Is libgit2's version pulled from https://github.com/nodejs/http-parser?

carlosmn:
People probably use it because Subversion uses `svn+ssh`, which doesn't make it a good idea to support; I don't know of any place which advertises its Git URLs this way. And, again, it's not something that's specified to work.

Anything under `deps/` is not our code, but stuff we need, typically on Windows. The parser is indeed the nodejs one (though I think we grabbed the code from Ryan directly back in the day).

johnhaley81:
Hey @carlosmn. I can't disagree with wrt it being not documented but the issue does remain that there are users that have their urls formatted like this and that git-core does support this. The commit that implements it in git-core does even say that it's stupid syntax.

The bigger issue and sticking point is that there is no current way to work around this issue either. In 0.22 and below you could modify the remote url in memory so a consumer of this library could intercept a url like this and reformat it into a more sane syntax. Now in 0.23+ you'd have to load a remote, mutate its url to something more sane, save it to disk, reload the remote, mutate its url back to the craziness, and then use the remote loaded off of disk to do an operation.

This seems pretty bad to me especially if you are changing a users repo in order to support a feature that they already have working in git-core.

If you aren't open to explicitly supporting this in the library would you be more open to allowing a mutation of remote urls in memory again? Or some other way to work around this issue?

cbargren:
PR for the http-parser: https://github.com/nodejs/http-parser/pull/276

ethomson:
Wow, I didn't know that anybody used `git+ssh` or `ssh+git`, and am dismayed to learn about it.  That said, I think that supporting this the correct thing to do.

Would you mind updating your PR to include support for `-` and `.` as well, per the changes in https://github.com/nodejs/http-parser/pull/276.

Additionally, I made some nitpicky comments about style - if you wouldn't mind addressing those since you're already in there.  Thanks!

src/transport.c
```diff
@@ -35,6 +35,8 @@ static transport_definition transports[] = {
 	{ "file://",  git_transport_local, NULL },
 #ifdef GIT_SSH
 	{ "ssh://",   git_transport_smart, &ssh_subtransport_definition },
+  { "ssh+git://",   git_transport_smart, &ssh_subtransport_definition },
+  { "git+ssh://",   git_transport_smart, &ssh_subtransport_definition },
```
ethomson: Could you make this (and the line above) a tab not spaces?

src/transports/ssh.c
```diff
@@ -509,16 +518,23 @@ static int _git_ssh_setup_conn(
...
 
-	if (!git__prefixcmp(url, prefix_ssh)) {
-		if ((error = gitno_extract_url_parts(&host, &port, &path, &user, &pass, url, default_port)) < 0)
-			goto done;
-	} else {
-		if ((error = git_ssh_extract_url_parts(&host, &user, url)) < 0)
-			goto done;
-		port = git__strdup(default_port);
-		GITERR_CHECK_ALLOC(port);
+	size_t i = 0;
```
ethomson: Please declare `i` at the top of the block to support bad old compilers.

tests/transport/register.c
```diff
@@ -44,6 +44,8 @@ void test_transport_register__custom_transport_ssh(void)
 
 #ifndef GIT_SSH
 	cl_git_fail_with(git_transport_new(&transport, NULL, "ssh://somehost:somepath"), -1);
+	cl_git_fail_with(git_transport_new(&transport, NULL, "ssh+git://somehost:somepath"), -1);
+	cl_git_fail_with(git_transport_new(&transport, NULL, "git+ssh://somehost:somepath"), -1);
```
ethomson: It looks like you're only testing that `ssh+git` fails when not compiling with `GIT_SSH`.  But I would expect that with or without this patch.

Can you test that you can successfully build a transport with `git+ssh` and with `ssh+git` below in the else block?

tests/transport/register.c
```diff
@@ -60,6 +62,8 @@ void test_transport_register__custom_transport_ssh(void)
...
```
ethomson: Similarly, I think you want to test `git+ssh` and `ssh+git` succeeding, below.

src/transports/ssh.c
```diff
@@ -20,7 +20,9 @@
 
 #define OWNING_SUBTRANSPORT(s) ((ssh_subtransport *)(s)->parent.subtransport)
 
-static const char prefix_ssh[] = "ssh://";
+static const char * ssh_prefixes[] = { "ssh://", "ssh+git://", "git+ssh://" };
+#define SSH_PREFIX_COUNT (sizeof(ssh_prefixes) / sizeof(ssh_prefixes[0]))
```
ethomson: We have an `ARRAY_SIZE` macro (defined in `util.h`) that will do this for you.  We would typically drop this `define` entirely and use `ARRAY_SIZE()` in the `for` loop itself.

If you're dropping that, could you rewrite the line above as `char *ssh_prefixes` (without the space between the `*` and `ssh_prefixes`).

cbargren:
@ethomson Wow, great feedback, thanks! I too was disappointed to hear that people are using git+ssh and ssh+git... Seems like a weird schema to use and they add nothing to functionality on top of already existing schemas. :shrug: What can you do... I've addressed all of your comments now. Let me know if there's anything else needed. Thanks again!

cbargren:
Is there anything else you guys need for this before pulling it in?

cbargren:
@ethomson @carlosmn Hey, sorry to be a bother but I haven't heard anything from you guys after your feedback. Is there anything else you needed for this PR?

ethomson:
Hi!  Terribly sorry - a bunch of us are coincidentally on vacation right now.  I'm just getting back to having internet; I'll take a look at this soon!  Thanks for your help and your patience!

cbargren:
Cool, thanks for the reply!

ethomson:
So I think that Carlos wanted to double-check with the Git mailing list and either _document_ this in git.git or _remove_ this behavior from git.git.  But if the latter, and Git removes this behavior (and I don't think that they will) then we should not merge this.  So let's just make sure that this is the right thing to do.

/cc @carlosmn

carlosmn:
The consensus from the list is that these were a bad idea and they never should have done it, but git has to support them for backwards compatibility.

Which means we'll have to support it too; it won't be the first bad idea we do.

johnhaley81:
Is there anything else that needs to be done to finish this up?

ethomson:
Thanks for the contribution!  :)

PR #3557, opened 2015-12-23
ssl: better error reporting if SSL_new fails

nacho:


carlosmn:
Shouldn't this logic be put into `set_ssl_error()`? Why *does* `SSL_new()` return an error?

nacho:
From the docs: https://www.openssl.org/docs/manmaster/ssl/SSL_new.html
    The creation of a new SSL structure failed. Check the error stack to find out the reason.

Basically I have no idea why that is failing for us, that's why I added it to get some more info.

nacho:
So we just got the reason applying my patch downstream:
<alexey> error:140BA0C3:SSL routines:SSL_new:null ssl ctx

Why is the context null? Why don't we get an error when creating the context?

carlosmn:
The only way the context would be `NULL` is either the consumer of libgit2 not initializing the library, or the initialization failing to set the search paths. We don't really have a way to report this kind of error during initialization, so we free the context.

nacho:
How about something like this?
diff --git a/src/global.c b/src/global.c
index 0eab8d5..b02c3b3 100644
--- a/src/global.c
+++ b/src/global.c
@@ -110,10 +110,18 @@ static void init_ssl(void)
         * to speak TLSv1 to perform the encryption itself.
         */
        git__ssl_ctx = SSL_CTX_new(SSLv23_method());
+       if (git__ssl_ctx == NULL) {
+               giterr_set(GITERR_SSL, "failed to create ssl context: %s",
+                          ERR_error_string(ERR_get_error(), NULL));
+               return;
+       }
+
        SSL_CTX_set_options(git__ssl_ctx, ssl_opts);
        SSL_CTX_set_mode(git__ssl_ctx, SSL_MODE_AUTO_RETRY);
        SSL_CTX_set_verify(git__ssl_ctx, SSL_VERIFY_NONE, NULL);
        if (!SSL_CTX_set_default_verify_paths(git__ssl_ctx)) {
+               giterr_set(GITERR_SSL, "could not set default verify paths: %s",
+                          ERR_error_string(ERR_get_error(), NULL));
                SSL_CTX_free(git__ssl_ctx);
                git__ssl_ctx = NULL;
        }
diff --git a/src/openssl_stream.c b/src/openssl_stream.c
index 1e2edee..e4f4c18 100644
--- a/src/openssl_stream.c
+++ b/src/openssl_stream.c
@@ -444,6 +444,10 @@ int git_openssl_stream_new(git_stream **out, const char *host, const char *port)
        if (error < 0)
                return error;
 
+       if (git__ssl_ctx == NULL) {
+               return -1;
+       }
+

No idea if it will work though, or if the error is put up on the stack. Better ideas?

carlosmn:
That error will be overwritten when you open a repository. We *might* be able to get the error out of the OpenSSL stack the first time the stream is created by setting the error in the constructor.

Have you determined that the error does come from failing to set the paths? What is the error that's reported there?

ethomson:
There hasn't been any movement on this issue in a year; I'm closing this.  Please do mention if this is still relevant to your interests.

PR #3558, opened 2015-12-23
index: get rid of the locking

carlosmn:
We don't support using an index object from multiple threads at the same
time, so the locking doesn't have any effect when following the
rules. If not following the rules, things are going to break down
anyway.

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -356,28 +356,6 @@ static unsigned int index_merge_mode(
 	return git_index__create_mode(mode);
 }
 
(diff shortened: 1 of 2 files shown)


ethomson:
I'll +1 the idea here.  We make a pretty clear statement about threading and - presuming one follows that - this locking we do in the index is unnecessary at best.  (At worst, we have not significantly tested this because it is wholly pointless given our stated threading model.)

carlosmn:
I'll just go ahead and merge it then, now that the tests pass.

PR #3559, opened 2015-12-23
Add a sanity check in git_indexer_commit to avoid subtraction overflow.

yongthecoder:
I have seen the subtraction overflow when the pack file contains nothing with custom transport protocol (when the serving side close the connection). We have fixed it in our custom writepack implementation, but I think the right fix should be in git_indexer_commit

### DIFF
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -914,12 +914,17 @@ int git_indexer_commit(git_indexer *idx, git_transfer_progress *stats)
 	git_filebuf index_file = {0};
 	void *packfile_trailer;
 
+	if (!idx->parsed_header) {
+		giterr_set(GITERR_INDEXER, "incomplete pack header");
+		return -1;
+	}
+
 	if (git_hash_ctx_init(&ctx) < 0)
 		return -1;
 
 	/* Test for this before resolve_deltas(), as it plays with idx->off */
-	if (idx->off < idx->pack->mwf.size - 20) {
-		giterr_set(GITERR_INDEXER, "Unexpected data at the end of the pack");
+	if (idx->off + 20 < idx->pack->mwf.size) {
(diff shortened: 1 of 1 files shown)


ethomson:
Agreed, let's have this check before the subtraction.  :+1: 

Would you mind changing the error message string to start with a lower-case letter?  (I realize that we're not terribly consistent here yet, but we're trying to move to that.)

Thanks!

carlosmn:
While this does indeed fix an underflow, it looks like there's a larger issue here; do we call `git_indexer_commit()` even when the connection to the remote was closed unexpectedly? We assume here that you wouldn't call this unless the transport signaled a correct completion, but it certainly doesn't hurt to double-check we're not obviously in an incorrect state.

An empty packfile would be larger than `GIT_OID_RAWSZ` since it also contains the header, so we should be looking for it to be `sizeof(struct git_pack_header) + GIT_OID_RAWSZ`.

yongthecoder:
Hi Carlos,

Yes, there must be an issue on the server side so it sends an empty message to close the connection without giving an error. But the client side should be able to survive even when the server side returns bad data. What we are doing is using a custom transport protocol based on smart protocol, and having a custom writepack that uses an indexer to save the pack file. Currently, I use a byte counter in the writepack implementation to work around the problem. git_pack_header is a private struct. This can also be fixed in no_sideband() in smart_protocol.c. But git_index_xyz is public API, I think it also makes sense that git_index_commit verifies the pack file is good before doing its job.

After reading the code carefully, I found I should check idx->parsed_header to know if the header is parsed, and change if (idx->off < idx->pack->mwf.size - 20) into if (idx->off + 20 < idx->pack->mwf.size).

Thanks

carlosmn:
Thanks. Sorry it took so long.

Issue #1534, opened 2013-05-02
do not write cached refs into packed-refs file


peff:
When we write the packed-refs file (e.g., because of a ref deletion), we use the filebuf code, which takes a lock on packed-refs and writes it out atomically. However, the values we write out come from our in-memory list of refs, which might actually be stale (e.g., if somebody else has modified the packed-refs file since we last read it). This can lead to unrelated refs appearing to rewind (or even disappear) when a ref is deleted and the repo is under load (especially if there are multiple deleters operating at once).

More details on the fix in upstream git can be found at git/git@b3f1280ec740d8012d18e870a50a5ff76c4e3c42.

PR #3561, opened 2015-12-24
Concurrency fixes for the reference db

carlosmn:
There's all sorts of races in there if you run many threads which all want to create and compress references.

We start by removing a useless test which doesn't even follow our own rules for concurrency and fixing a second one to use different objects and actually perform a concurrency test.

A bunch of this is simply to bubble up error codes so we know what we're dealing with, or fixing how we report them.

This also makes the packing logic more robust and safer by ignoring transient errors (and really the non-transient ones but it's more important that we continue working) and only deleting references if they haven't changed since we packed.

We still need to make sure #1534 doesn't happen by locking the `packed-refs` file before reloading, but the test no longer fails every second run.

### DIFF
--- a/src/refdb_fs.c
+++ b/src/refdb_fs.c
@@ -902,7 +902,7 @@ static int packed_remove_loose(refdb_fs_backend *backend)
 {
 	size_t i;
 	git_buf ref_content = GIT_BUF_INIT;
-	int failed = 0, error = 0;
+	int error = 0;
 
 	/* backend->refcache is already locked when this is called */
 
@@ -917,12 +917,12 @@ static int packed_remove_loose(refdb_fs_backend *backend)
 		/* We need to stop anybody from updating the ref while we try to do a safe delete */
 		error = loose_lock(&lock, backend, ref->name);
 		/* If someone else is updating it, let them do it */
-		if (error == GIT_EEXISTS)
+		if (error == GIT_EEXISTS || error == GIT_ENOTFOUND)
 			continue;
 
 		if (error < 0) {
-			failed = 1;
-			continue;
+			giterr_set(GITERR_REFERENCE, "failed to lock loose reference '%s'", ref->name);
+			return error;
 		}
 
 		error = git_futils_readbuffer(&ref_content, lock.path_original);
@@ -940,7 +940,6 @@ static int packed_remove_loose(refdb_fs_backend *backend)
 
 		/* Figure out the current id; if we fail record it but don't fail the whole operation */
 		if ((error = loose_parse_oid(&current_id, lock.path_original, &ref_content)) < 0) {
-			failed = 1;
 			git_filebuf_cleanup(&lock);
 			continue;
 		}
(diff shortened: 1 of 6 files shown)


carlosmn:
This finally works fine on my computers. I don't know why AppVeyor doesn't like it. I've tried it with VS2015 and it works fine for me.

stanhu:
Thanks for working on this fix. I can reproduce #1534 quite easily by running `git gc` continuously on a networked fileystem and checking the SHA of a specific branch with Rugged. At times, it appears the HEAD goes "back in time" momentarily; it's also possible to cause the branch to appear to go missing entirely. More details here: https://gitlab.com/gitlab-org/gitlab-ce/issues/15392#note_14530450

The current workaround seems to be to initialize a new `Rugged::Repository` anytime you need to look up the latest SHA of a commit. This does not seem ideal, so I'm curious if there's anything the community can do to help move this PR forward.

src/refdb_fs.c
```diff
@@ -901,40 +905,60 @@ static int packed_write_ref(struct packref *ref, git_filebuf *file)
...
 			continue;
 
-		if (git_buf_joinpath(&full_path, backend->path, ref->name) < 0)
-			return -1; /* critical; do not try to recover on oom */
+		git_filebuf_cleanup(&lock);
 
-		if (git_path_exists(full_path.ptr) && p_unlink(full_path.ptr) < 0) {
-			if (failed)
-				continue;
+		/* We need to stopy anybody from updating the ref while we try to do a safe delete */
```
stanhu: stopy -> stop :)

tests/threads/refdb.c
```diff
@@ -37,120 +48,99 @@ static void *iterate_refs(void *arg)
...
+			if (error == GIT_ENOTFOUND)
+				error = 0;
 
-		if (!git_reference_lookup(&ref, g_repo, name)) {
-			cl_git_pass(git_reference_delete(ref));
+			cl_git_pass(error);
 			git_reference_free(ref);
 		}
 
 		if (i == 5) {
```
ethomson: This might be nice to have as a constant.

tests/threads/refdb.c
```diff
@@ -37,120 +48,99 @@ static void *iterate_refs(void *arg)
...
-	int *id = arg, i;
+	int i, error;
+	struct th_data *data = (struct th_data *) arg;
 	git_reference *ref;
 	char name[128];
+	git_repository *repo;
+
+	cl_git_pass(git_repository_open(&repo, data->path));
 
 	for (i = 0; i < 10; ++i) {
```
ethomson: This might be nice to have as a constant.

tests/threads/refdb.c
```diff
@@ -37,120 +48,99 @@ static void *iterate_refs(void *arg)
...
 
-	cl_git_pass(git_reference_name_to_id(&head, g_repo, "HEAD"));
+	cl_git_pass(git_repository_open(&repo, data->path));
+
+	do {
+		error = git_reference_name_to_id(&head, repo, "HEAD");
+	} while (error == GIT_ELOCKED);
+	cl_git_pass(error);
 
 	for (i = 0; i < 10; ++i) {
```
ethomson: This might be nice to have as a constant.  (I realize this was already here but it might be a nice cleanup.)

tests/threads/refdb.c
```diff
@@ -37,120 +48,99 @@ static void *iterate_refs(void *arg)
...
-		p_snprintf(name, sizeof(name), "refs/heads/thread-%03d-%02d", *id, i);
-		cl_git_pass(git_reference_create(&ref[i], g_repo, name, &head, 0, NULL));
+		p_snprintf(name, sizeof(name), "refs/heads/thread-%03d-%02d", data->id, i);
+		do {
+			error = git_reference_create(&ref[i], repo, name, &head, 0, NULL);
+		} while (error == GIT_ELOCKED);
+		cl_git_pass(error);
 
 		if (i == 5) {
```
ethomson: This might be nice to have as a constant.

PR #3564, opened 2015-12-26
Custom merge drivers and proper gitattributes `merge` handling

### LINKED ISSUE #2180: Custom merge driver support

ethomson:
While #3497 added support for `merge=union` in `.gitattributes`, I felt that we should have a more complete story for handling the `merge` attribute.  This PR:

1. Adds a framework for adding custom merge drivers.  This is heavily inspired by the custom filter framework, and a consumer can add one or more custom merge drivers by name (for example, you may add a new `custom` driver) or - probably more useful - can add a wildcard merge driver.  This allows a consumer to be given the option to handle all configured merge drivers that are not built-in.

    For example, a consumer could register a custom merge driver for the wildcard (`*`).  If there were a `merge=custom` line in a `.gitattributes` for a file, the configured merge driver would be given the information about the file changes and the `.gitattributes` configuration.  The driver could (for example) look for the corresponding `custom` merge driver and - if found - elect to accept responsibility for this change.  It would then invoke the `custom` merge driver and provide the results back.

    The merge driver may also defer the check, at which point libgit2 will invoke the default built-in merge driver.  This is equivalent to when a user has configured `merge=foobar` in their `.gitattributes` but has not configured the `foobar` merge driver.  The merge driver configuration is ignored, and git will simply behave as if it was not configured at all.

2. Additional handling for the [`merge` attribute](http://git-scm.com/docs/gitattributes).  When set (eg, `merge`) this will indicate that the file is mergeable and the default built-in `text` driver should be used.  When unset (eg, `-merge`) this will indicate that the file is binary and should not be merged (ie, is simply a conflict).  When unspecified (eg, no merge line) this indicates that the file is text but if a `merge.default` configuration setting exists, that will specify the driver to be used (otherwise the built-in `text` driver will be used).  Finally, if it is set to a string (eg, `merge=foobar`) then that indicates the driver to be used.

    Any of the built-in drivers (`text`, `binary` or `union`) may be specified.

3. Proper handling of the `merge.default` configuration setting.  Query the `merge.default` to get the default merge driver name and use it when the `merge` attribute is not specified for a file.

Fixes #3497 
Fixes #2180

### DIFF
--- a/include/git2/sys/merge.h
+++ b/include/git2/sys/merge.h
@@ -56,22 +56,12 @@ GIT_EXTERN(git_merge_file_options *) git_merge_driver_source_file_options(
 	const git_merge_driver_source *src);
 
 
-/*
- * struct git_merge_driver
- *
- * The merge driver lifecycle:
- * - initialize - first use of merge driver
- * - shutdown   - merge driver removed/unregistered from system
- * - check      - considering using merge driver for file
- * - apply      - apply merge driver to the file
- * - cleanup    - done with file
- */
-
 /**
  * Initialize callback on merge driver
  *
  * Specified as `driver.initialize`, this is an optional callback invoked
- * before a merge driver is first used.  It will be called once at most.
+ * before a merge driver is first used.  It will be called once at most
+ * per library lifetime.
  *
  * If non-NULL, the merge driver's `initialize` callback will be invoked
  * right before the first use of the driver, so you can defer expensive
@@ -170,20 +160,33 @@ typedef void (*git_merge_driver_cleanup_fn)(
  * To associate extra data with a driver, allocate extra data and put the
  * `git_merge_driver` struct at the start of your data buffer, then cast
  * the `self` pointer to your larger structure when your callback is invoked.
- *
- * `version` should be set to GIT_MERGE_DRIVER_VERSION
- *
- * The `initialize`, `shutdown`, `check`, `apply`, and `cleanup`
- * callbacks are all documented above with the respective function pointer
- * typedefs.
  */
 struct git_merge_driver {
+	/** The `version` should be set to `GIT_MERGE_DRIVER_VERSION`. */
 	unsigned int                 version;
 
+	/** Called when the merge driver is first used for any file. */
 	git_merge_driver_init_fn     initialize;
+
+	/** Called when the merge driver is unregistered from the system. */
 	git_merge_driver_shutdown_fn shutdown;
+
+	/**
+	 * Called to determine whether the merge driver should be invoked
+	 * for a given file.  If this function returns `GIT_PASSTHROUGH`
+	 * then the `apply` function will not be invoked and the default
+	 * (`text`) merge driver will instead be run.
+	 */
 	git_merge_driver_check_fn    check;
+
+	/**
+	 * Called to merge the contents of a conflict.  If this function
+	 * returns `GIT_PASSTHROUGH` then the default (`text`) merge driver
+	 * will instead be invoked.  If this function returns
+	 * `GIT_EMERGECONFLICT` then the file will remain conflicted.
 	git_merge_driver_apply_fn    apply;
+
+	/** Called when the system is done filtering for a file. */
 	git_merge_driver_cleanup_fn  cleanup;
 };
 
--- a/include/git2/sys/merge.h
+++ b/include/git2/sys/merge.h
@@ -83,41 +83,7 @@ typedef int (*git_merge_driver_init_fn)(git_merge_driver *self);
 typedef void (*git_merge_driver_shutdown_fn)(git_merge_driver *self);
 
 /**
- * Callback to decide if a given conflict can be resolved with this merge
- * driver.
- *
- * Specified as `driver.check`, this is an optional callback that checks
- * if the given conflict can be resolved with this merge driver.
- *
- * It should return 0 if the merge driver should be applied (i.e. success),
- * `GIT_PASSTHROUGH` if the driver is not available, which is the equivalent
- * of an unregistered or nonexistent merge driver.  In this case, the default
- * (`text`) driver will be run.  This is useful if you register a wildcard
- * merge driver but are not interested in handling the requested file (and
- * should just fallback).  The driver can also return `GIT_EMERGECONFLICT`
- * if the driver is not able to produce a merge result, and the file will
- * remain conflicted.  Any other errors will fail and return to the caller.
- *
- * The `name` will be set to the name of the driver as configured in the
- * attributes.
- *
- * The `src` contains the data about the file to be merged.
- *
- * The `payload` will be a pointer to a reference payload for the driver.
- * This will start as NULL, but `check` can assign to this pointer for
- * later use by the `apply` callback.  Note that the value should be heap
- * allocated (not stack), so that it doesn't go away before the `apply`
- * callback can use it.  If a driver allocates and assigns a value to the
- * `payload`, it will need a `cleanup` callback to free the payload.
- */
-typedef int (*git_merge_driver_check_fn)(
-	git_merge_driver *self,
-	void **payload,
-	const char *name,
-	const git_merge_driver_source *src);
-
-/**
- * Callback to actually perform the merge.
+ * Callback to perform the merge.
  *
  * Specified as `driver.apply`, this is the callback that actually does the
  * merge.  If it can successfully perform a merge, it should populate
@@ -129,30 +95,18 @@ typedef int (*git_merge_driver_check_fn)(
  * and the file will remain conflicted.  Any other errors will fail and
  * return to the caller.
  *
- * The `src` contains the data about the file to be merged.
+ * The `filter_name` contains the name of the filter that was invoked, as
+ * specified by the file's attributes.
  *
- * The `payload` value will refer to any payload that was set by the
- * `check` callback.  It may be read from or written to as needed.
+ * The `src` contains the data about the file to be merged.
  */
 typedef int (*git_merge_driver_apply_fn)(
 	git_merge_driver *self,
-	void **payload,
 	const char **path_out,
 	uint32_t *mode_out,
 	git_buf *merged_out,
+	const char *filter_name,
 	const git_merge_driver_source *src);
-
-/**
- * Callback to clean up after merge has been performed.
- *
- * Specified as `driver.cleanup`, this is an optional callback invoked
- * after the driver has been run.  If the `check` or `apply` callbacks
- * allocated a `payload` to keep per-source merge driver state, use this
- * callback to free that payload and release resources as required.
- */
-typedef void (*git_merge_driver_cleanup_fn)(
-	git_merge_driver *self,
-	void *payload);
 
 /**
  * Merge driver structure used to register custom merge drivers.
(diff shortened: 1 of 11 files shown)


include/git2/sys/merge.h
```diff
@@ -0,0 +1,230 @@
...
+/*
+ * struct git_merge_driver
+ *
+ * The merge driver lifecycle:
+ * - initialize - first use of merge driver
+ * - shutdown   - merge driver removed/unregistered from system
+ * - check      - considering using merge driver for file
+ * - apply      - apply merge driver to the file
+ * - cleanup    - done with file
+ */
```
carlosmn: This block won't show up in the doc viewer, so the "as documented above" comment there will seem rather nonsensical. This should probably be part of the documentation for the struct.

include/git2/sys/merge.h
```diff
@@ -0,0 +1,230 @@
...
+ * Currently the merge driver registry is not thread safe, so any
+ * registering or deregistering of merge drivers must be done outside of
+ * any possible usage of the drivers (i.e. during application setup or
+ * shutdown).
+ *
+ * @param name The name of this driver to match an attribute.  Attempting
+ * 			to register with an in-use name will return GIT_EEXISTS.
+ * @param driver The merge driver definition.  This pointer will be stored
+ *			as is by libgit2 so it must be a durable allocation (either
+ *			static or on the heap).
```
carlosmn: I'm wondering about the rationale for this behaviour. Everywhere (or nearly everywhere) else we take a copy of whatever the user gave us, why don't we make a copy here?

ethomson: This is actually how we handle the filter drivers and the backends.  Since the user may extend the strict arbitrarily, we actually don't know how big it is so we couldn't make a copy if we wanted (and get the users data after our struct definition.)

include/git2/sys/merge.h
ethomson: Yeah I'll fix that; I think we're similarly deficient in the filter code, I copied this pattern from that.

include/git2/sys/merge.h
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Callback to actually perform the merge.
+ *
+ * Specified as `driver.apply`, this is the callback that actually does the
+ * merge.  If it can successfully perform a merge, it should populate
+ * `path_out` with a pointer to the filename to accept, `mode_out` with
+ * the resultant mode, and `merged_out` with the buffer of the merged file
+ * and then return 0.  If the driver returns `GIT_PASSTHROUGH`, then the
+ * default merge driver should instead be run.  It can also return
```
carlosmn: Shouldn't the `check` method have already told us whether a particular driver can handle the file? Are we treating that as a hard no vs "probably"? Or what's the use-case you're envisioning here?

include/git2/sys/merge.h
carlosmn: I figured out we're taking the user's interface for an existing object after I asked. I guess then I'm wondering why there is *the* merge driver/filter which we use instead of having a constructor to create an instance we can use per-use, which would remove the need for the payload `out`-arg in `check` as that's state-keeping that needs libgit2's help because we have a singleton.

This also brings me to concurrency. If a caller tries to perform merge in two repos at the same time (each thread with its own objects as we ask for in the docs) and both need filters; we have one filter (per name) per library instance, so who's responsible for concurrency control? We don't seem to think we are, so is the singleton filter responsible for `check`, `apply` and such to use locks/whatever when they look at global state? Should this be documented in THREADING or in here?

This should look like the filters, but I guess I never quite looked at them in detail. At least we should document what may be called concurrently.

include/git2/sys/merge.h
```diff
@@ -0,0 +1,230 @@
...
+ * - check      - considering using merge driver for file
+ * - apply      - apply merge driver to the file
+ * - cleanup    - done with file
+ */
+
+/**
+ * Initialize callback on merge driver
+ *
+ * Specified as `driver.initialize`, this is an optional callback invoked
+ * before a merge driver is first used.  It will be called once at most.
```
carlosmn: This is true as long as the user doesn't register the same driver under multiple names. I'm not sure that we want to bother with that so much, but we should probably add "per library lifetime" or similar here, as shutting down the library and then starting it back up (and registering it again) would cause us to call it again.

src/merge.c
```diff
@@ -810,76 +798,150 @@ static int merge_conflict_resolve_one_renamed(
...
+	} else {
+		/* find the merge driver for this file */
+		if ((error = git_merge_driver_for_source(&driver, &data, &source)) < 0)
+			goto done;
+	}
 
-	if ((index_entry = git_pool_mallocz(&diff_list->pool, sizeof(git_index_entry))) == NULL)
-	GITERR_CHECK_ALLOC(index_entry);
+    error = merge_conflict_invoke_driver(&merge_result,
+        driver, data, diff_list, &source);
```
carlosmn: This block seems mis-indented.

src/merge_driver.c
```diff
@@ -0,0 +1,397 @@
...
+static void merge_driver_registry_shutdown(void)
+{
+	struct merge_driver_registry *reg;
+	git_merge_driver_entry *entry;
+    size_t i;
+
+    if ((reg = git__swap(merge_driver_registry, NULL)) == NULL)
+        return;
+
+    git_vector_foreach(&reg->drivers, i, entry) {
```
carlosmn: The indentation here is also off.

src/merge_driver.c
```diff
@@ -0,0 +1,397 @@
...
+	git_merge_driver *self,
+	void **payload,
+	const char *name,
+	const git_merge_driver_source *src)
+{
+	GIT_UNUSED(self);
+	GIT_UNUSED(name);
+	GIT_UNUSED(src);
+
+	*payload = (void *)GIT_MERGE_FILE_FAVOR_NORMAL;
```
carlosmn: I'm a bit conflicted about this. Sure, pointers are just numbers underneath. But inserting numbers into them is the kind of thing that makes portability terrible. I'd create a dummy static variable and point there. Comparing afterwards would still be a simple number comparison, but we'd at least have an actual pointer.

src/merge_driver.c
```diff
@@ -0,0 +1,397 @@
...
+
+    if ((error = git_vector_init(&reg->drivers, 3, merge_driver_entry_cmp)) < 0)
+		goto done;
+	
+    reg = git__compare_and_swap(&merge_driver_registry, NULL, reg);
+
+    if (reg != NULL)
+        goto done;
+
+    git__on_shutdown(merge_driver_registry_shutdown);
```
carlosmn: This block has odd indentation.

src/merge_driver.c
```diff
@@ -0,0 +1,397 @@
...
+
+	if ((error = git_merge_driver_register(
+			merge_driver_name__text, &git_merge_driver__text)) < 0 ||
+		(error = git_merge_driver_register(
+			merge_driver_name__union, &git_merge_driver__union)) < 0 ||
+		(error = git_merge_driver_register(
+			merge_driver_name__binary, &git_merge_driver__binary)) < 0)
+		goto done;
```
carlosmn: These registrations happen after we've set the global pointer. If we're pre-empted just before this, a second thread could come, see `merge_driver_registry` is set and return immediately. If we don't come back in time, the second thread might try to look up in the registration which we haven't fully initialized.

src/merge_driver.c
```diff
@@ -0,0 +1,397 @@
...
+git_merge_driver *git_merge_driver_lookup(const char *name)
+{
+	git_merge_driver_entry *entry;
+	size_t pos;
+	int error;
+
+	if (name == merge_driver_name__text)
+		return &git_merge_driver__text;
+	else if (name == merge_driver_name__binary)
+		return &git_merge_driver__binary;
```
carlosmn: Why are we registering them during initialisation if we're going to return a hard-coded version here? Why is "union" missing from this check?

src/merge_driver.c
```diff
@@ -0,0 +1,397 @@
...
+	if (merge_driver_registry_initialize() < 0)
+		return NULL;
+
+	error = git_vector_search2(&pos, &merge_driver_registry->drivers,
+		merge_driver_entry_search, name);
+
+	if (error == GIT_ENOTFOUND)
+		return NULL;
+
+    entry = git_vector_get(&merge_driver_registry->drivers, pos);
```
carlosmn: This line is mis-indented.

carlosmn:
The rest of the PR is fine. I'm mostly worried about documenting when we call certain things, and about having a global registration which can be called from multiple threads at the same time.

Envek:
Hello everyone! I have a question: it's a bit unclear for me, but will this PR give an ability to configure and call external merge drivers (external executables for example)? See discussion at [GitLab's issue #5902](https://gitlab.com/gitlab-org/gitlab-ce/issues/5902) for context and possible use cases.

arthurschreiber:
@Envek If I'm not mistaken, this does not "magically" add support external merge drivers.

But you could write a custom libgit2 merge driver that either calls out to your existing merge driver executable or reimplements similar functionality.

carlosmn:
This will give you the ability to intercept those external merge drivers and call them, but libgit2 does not execute programs. It does give you the information you need to execute them as your environment needs.

stanhu:
Would love to see this PR merged soon. Anything I can do to help?

ethomson:
Sorry for the delay, @stanhu - I think that once we get #3597 merged, I'll be able to get this one fixed up!

stanhu:
Thanks, @ethomson! Awesome work.

MiroslavHoudek:
A dumb question (possibly in a inappropriate place): I can't use this custom merge driver as developer linking to libgit2 library, right? I want to do my own merging, so I was hoping I will just simply register my function as a default driver somehow, and then it would get called, it'd get reference to two conflicting files as parameter, it'd produce the merged version and return it (and it'd get written). But looking at the code, it looks buried deep in the library, for some internal purpose, not accessible to me at all ... right?

ethomson:
Hi @MiroslavHoudek - you can wire this up as a developer, however it doesn't sound like this is *quite* what you're looking for.  The purpose of this is to be able to wire up a custom merge driver so that you can specify `merge=foobar` in `.gitattributes` and then be able to have your custom function get called for those files.

But it will only invoke your custom code for those files that have those attributes, not for every file.  If that's an acceptable limitation for you, then this will work fine.  If not, we might need to think a bit more deeply about how to get you what you need.

MiroslavHoudek:
Hi, I think I know where my brain took a wrong turn - I thought that I can add the custom driver in my (e.g.) main.cpp, and there I will write my handler, and then register it in libgit2.dll by calling git_merge_driver_register() ... whereas in reality the intended use seems to be to develop the driver and compile/link it with libgit2.dll, and then specify its invocation in the `.gitattributes`. Thank you @ethomson for your response (as well as for all the great work on libgit2), both much appreciated!

carlosmn:
The point of the custom drivers is that they don't live in this codebase. If you have a custom one, it should definitely be in your own code. This is why we publish registration, otherwise we'd hard-code it.

matto1990:
Will this be looked at now #3597 has been merged? Is there anything I can do to help with this?

carlosmn:
That PR existed _because_ this was looked at. We'll get back to this PR when we have time.

matto1990:
That makes sense. Thanks for the swift reply and for working on libgit2!

MiroslavHoudek:
@carlosmn thank you for your response, it gave me the confidence to find the real problem and get things straight. I have a functional custom driver now, yay!

include/git2/sys/merge.h
ethomson: I will open a subsequent PR to document this in `THREADING`.

include/git2/sys/merge.h
ethomson: Yeah, so, this is a good point.  I've refactored the `check` method out; now `apply` is in charge of deferring.  With `check` gone, you don't need the `payload` data that let you keep some state between the `check` and `apply` invocations, so that goes away too.  So, too, does the `cleanup` that let you clean up that shared state.  Much simpler.  Many thanks.

ethomson:
Thanks for the sage comments as always @carlosmn - I think that I've addressed your concerns.

ethomson:
I rebased this on master because some conflicts had crept in.  @carlosmn - anything keeping us from :ship: this?

metcalf:
Will this eventually change the behavior of merge conflict detection on Github pull requests? For certain generated files, I'm thinking of defaulting to union for rebases and merges since we can rebuild broken files. On the other hand, we don't want people pressing the big green button unless the merge is truly clean.

ethomson:
This will not change the behavior of GitHub PR merges directly, no.  If GitHub wants to change behavior, that's not something that libgit2 would impose upon them.

metcalf:
Makes sense, thanks!

PR #3567, opened 2015-12-31
Few p_getaddrinfo fixes

sba1:
These are some (hopefully correct) fixes for the p_getaddrinfo() function.

If you prefer to use GITERR_CHECK_ALLOC() in commit ea5942b I can also do it of course.

### DIFF
--- a/src/posix.c
+++ b/src/posix.c
@@ -62,7 +62,10 @@ int p_getaddrinfo(
 	ai = ainfo;
 
 	for (p = 1; ainfo->ai_hostent->h_addr_list[p] != NULL; p++) {
-		ai->ai_next = malloc(sizeof(struct addrinfo));
+		if (!(ai->ai_next = malloc(sizeof(struct addrinfo)))) {
+			p_freeaddrinfo(ainfo);
+			return -1;
+		}
 		memcpy(&ai->ai_next, ainfo, sizeof(struct addrinfo));
 		memcpy(&ai->ai_next->ai_addr_in.sin_addr,
 			ainfo->ai_hostent->h_addr_list[p],
--- a/src/posix.c
+++ b/src/posix.c
@@ -66,7 +66,7 @@ int p_getaddrinfo(
 			p_freeaddrinfo(ainfo);
 			return -1;
 		}
-		memcpy(&ai->ai_next, ainfo, sizeof(struct addrinfo));
+		memcpy(ai->ai_next, ainfo, sizeof(struct addrinfo));
 		memcpy(&ai->ai_next->ai_addr_in.sin_addr,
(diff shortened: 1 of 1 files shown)


src/posix.c
```diff
@@ -62,8 +62,11 @@ int p_getaddrinfo(
 	ai = ainfo;
 
 	for (p = 1; ainfo->ai_hostent->h_addr_list[p] != NULL; p++) {
-		ai->ai_next = malloc(sizeof(struct addrinfo));
-		memcpy(&ai->ai_next, ainfo, sizeof(struct addrinfo));
+		if (!(ai->ai_next = malloc(sizeof(struct addrinfo)))) {
```
carlosmn: We should be using `git__malloc()` here, or even `git__calloc()`. And yes, we should just bail out with the macro.

sba1: Am 2015-12-31 16:19, schrieb Carlos Martín Nieto:
> In src/posix.c [1]:
> 
>> @@ -62,8 +62,11 @@ int p_getaddrinfo(
>>      ai = ainfo;
>> 
>>      for (p = 1; ainfo->ai_hostent->h_addr_list[p] != NULL; p++) {
>> -        ai->ai_next = malloc(sizeof(struct addrinfo));
>> -        memcpy(&ai->ai_next, ainfo, sizeof(struct addrinfo));
>> +        if (!(ai->ai_next = malloc(sizeof(struct addrinfo)))) {
> 
> We should be using git__malloc() here, or even git__calloc(). And yes,
> we should just bail out with the macro.

Note however that this is a replacement function for getaddrinfo() that 
also doesn't follow libgit2 semantics. For instance, GITERR_CHECK_ALLOC 
will not clean up properly while the standard getaddrinfo() is supposed 
to cleanup properly in a failure case. So should this really being used? 
In posix.c there is quite a mixture of styles so it is not clear what 
style should be really used. But if you say git__malloc() and 
GITERR_CHECK_ALLOC then I will adjust my changes.

ethomson: It's not 100% expected that the `p_*` functions that we write should set library error codes.  These are often used as fallbacks when an actual posix function is not available.  And - of course - the actual posix functions (like the system's `getaddrinfo`) will never set the library error codes.

Therefore `p_getaddrinfo` doesn't need to set the library error code.  Instead, it needs to set something that `p_gai_strerror` can cope with (which is our reimplementation of `gai_strerror`.  [Indeed that does so - `-1` is treated as an OOM scenario.](https://github.com/libgit2/libgit2/blob/a8d0b4de92b83745308b2d9af0f62828bfb550a1/src/posix.c#L94)

As a result, I think this change is correct.

PR #3568, opened 2015-12-31
Fix access violation in diff/status

jeffhostetler:
This change addresses an access violation seen
during status.  The value of entry->path given
to diff_delta__from_one() is null at line 1174
in handle_unmatched_new_item().

This happens when GIT_DIFF_RECURSE_UNTRACKED_DIRS
and GIT_DIFF_INCLUDE_IGNORED are set *AND* the
working directory contains a new untracked/ignored
folder that sorts after anything in the index.

    git2.dll!strlen() Line 81
    git2.dll!git_pool_strdup() Line 216
    git2.dll!diff_delta__alloc() Line 32
    git2.dll!diff_delta__from_one() Line 142
    git2.dll!handle_unmatched_new_item() Line 1174
    git2.dll!git_diff__from_iterators() Line 1295
    git2.dll!git_diff_index_to_workdir() Line 1421
    git2.dll!git_status_list_new() Line 345
    status.exe!main() Line 113

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -1090,6 +1090,28 @@ static int handle_unmatched_new_item(
 				return error;
 			giterr_clear();
 
+			/* The above call to iterator_advance_into() returns GIT_ENOTFOUND
+			 * both when an existing directory is empty *AND* when the directory
+			 * itself does not exist. The latter can happen if the directory is
+			 * deleted after the fs_iterator__expand_dir() / dirload_with_stat()
+			 * code returns. In this case, the fs_iterator__pop_frame() code
+			 * memset's the "fi->entry" structure during the iteration.
+			 *
+			 * We have an alias to it in our "nitem" variable.
+			 *
+			 * Therefore, if the path is null, we need to skip it as if it never
+			 * existed, rather than trying to mark it ignored, since we no longer
+			 * have the pathname.  Also note that the iterator has already been
+			 * advanced, so we just return 0.
+			 *
+			 * This has only been observed when GIT_DIFF_RECURSE_UNTRACKED_DIRS
+			 * and GIT_DIFF_INCLUDE_IGNORED are set. (The former gets us into
+			 * this block and the latter is filtered in diff_delta__from_one().)
+			 */
+			if (info->new_iter->type == GIT_ITERATOR_TYPE_WORKDIR &&
+				nitem->path == NULL)
+				return 0;
+
 			/* if directory is empty, can't advance into it, so either skip
 			 * it or ignore it
 			 */
--- a/tests/diff/racediffiter.c
+++ b/tests/diff/racediffiter.c
@@ -1,0 +1,126 @@
+/* This test exercises the problem described in
+** https://github.com/libgit2/libgit2/pull/3568
+** where deleting a directory during a diff/status
+** operation can cause an access violation.
+**
+** The "test_diff_racediffiter__basic() test confirms
+** the normal operation of diff on the given repo.
+**
+** The "test_diff_racediffiter__racy_rmdir() test
+** uses the new diff progress callback to delete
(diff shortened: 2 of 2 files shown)


jeffhostetler:
(Once again I buried the lead in my commit message.)
The problem is seen when a new untracked/ignored directory
is deleted from the working directory while status is running.
In particular, after the call to fs_iterator__expand_dir() / dirload_with_stat()
completes (and the directory is in the fs_iterator_frame) but before the
directory is actually examined in handle_unmatched_new_item().

The directory needs to sort last anything in the index.

jeffhostetler:
To test this:
1. Create a repo with a directory "A" and some files and commit that.
2. Create a untracked/ignored directory "Z".
3. Run "example/status.exe --ignored -uall" in the debugger with a breakpoint
set on line 1099 in src/diff.c on the iterator_advance_into() call.
4. In another window, rmdir "Z".
5. Allow status.exe to continue.

jeffhostetler:
@ethomson I tried to isolate this here and it fixes the problem, but it feels a little weird -- like the iterator is unwinding/popping 1 too many frames in this edge case, but it should be safer overall.

whoisj:
Seems simple enough :+1:

jeffhostetler:
The other day I pushed up a clar test to demonstrate this so that we don't have to rely on the interactive test scenario I described earlier.

Just wanted to send a ping on this.

ethomson:
Hi!  Terribly sorry - a bunch of us are coincidentally on vacation right now.  I'm just getting back to having internet; I'll take a look at this tomorrow!  And thanks much for the unit test that's very helpful.

ethomson:
Thanks for reminding me to pay attention to this.  I love that you wrote a unit test for this.

I don't feel awesome about relying on the iterator to memset that to `0`.  That sounds like the sort of thing that will get accidentally refactored away some day.  I wonder if disambiguating the error codes between an empty dir and a suddenly-missing dir would be helpful here?

Unless you think that's dumb I'm going to pull this in to #3691 and try that avenue.  If that's a bust then I'll merge this as-is.

jeffhostetler:
Yeah, there's a lot of weirdness in the iterator code and variable aliasing, so the fix itself is probably an accident-in-waiting as you suggest.  And it sounds like your overhaul of the iterators will hopefully address this.

Please take/merge this work into your WIP as you see fit.  Especially the test case.

I don't know about the error code change. (It's been too long since I was in that deep.)  This may be the tip of a bigger set of race problems -- in that the disk is free to change during the scan -- it maybe simpler to just eat the missing item and go on (since the readdir() is the only thing that knew it was there in the first place).

ethomson:
> it maybe simpler to just eat the missing item and go on (since the readdir() is the only thing that knew it was there in the first place).

So the iterators do eat things that disappear between the `readdir` and the subsequent `stat`s.  At that point, they've populated all the directory entries, and now the caller can `advance` over them.  But if one of those was a directory, then we load that at the point they `advance` (or `advance_into`).  If the directory is gone at that point, that's when we return the not found.  So it's race handling.

There may well be other issues.  :P

ethomson:
A quick followup:  it turns out that this is *exactly* the sort of thing that accidentally got refactored away.  It turns out that I (inadvertently, I suppose) removed the behavior of `advance_into` where it returned `GIT_ENOTFOUND` on an empty directory.

This seems like the _right thing_ to me.  So I added a couple of unit tests to validate that behavior.

Then I cherry-picked your unit test, and I made the logical equivalent of your fix into `diff.c`, with this newfangled behavior in mind.

Thanks for this.  Again, keeping this open until #3691 lands for my own sanity.

carlosmn:
The big iterators PR was merged, so closing.

PR #3569, opened 2016-01-06
Fix some issues with generated pkg-config file

arthurschreiber:
This is related to some issues we're running into when building rugged on windows with SSH support, but is relevant for other platforms as well.

See https://github.com/libgit2/rugged/pull/559 and https://github.com/libgit2/rugged/issues/558 for some further information.


CMakeLists.txt
```diff
@@ -255,9 +255,11 @@ IF (WIN32 AND WINHTTP)
 
 		INCLUDE_DIRECTORIES(deps/winhttp)
 		LINK_DIRECTORIES(${LIBWINHTTP_PATH})
+		LIST(APPEND LIBGIT2_PC_LIBS "-L${LIBWINHTTP_PATH}")
```
arthurschreiber: I'm not 100% sure about this line here. I think it's wrong to specify this here, but maybe someone else knows better. :smile:

CMakeLists.txt
```diff
@@ -255,9 +255,11 @@ IF (WIN32 AND WINHTTP)
...
 	ENDIF ()
 
 	LINK_LIBRARIES(winhttp rpcrt4 crypt32)
+	LIST(APPEND LIBGIT2_PC_LIBS "-lwinhttp" "-lrpcrt4" "-lcrypt32" "-lole32")
```
mscharley: Should `ole32` be added the line above this as well?

CMakeLists.txt
mscharley: Seems reasonable to me since you're updating all the other variables here.

CMakeLists.txt
arthurschreiber: Not sure. @ethomson and/or @carlosmn will know better.

CMakeLists.txt
arthurschreiber: I dropped this line again. the winhttp stub is only generated because mingw is missing an import library for winhttp, and the generated file/path will only be available during the compile phase of libgit2

arthurschreiber:
@jacquesg But the perl bindings link dynamically, not statically.

jacquesg:
It definitely links libssh2 statically into the perl extension (which itself is a DLL/so) and uses the winhttp import library.

Here's a link to the Windows build: https://ci.appveyor.com/project/jacquesg/p5-git-raw

carlosmn:
`CoInitializeEx` and `CoCreateInstance` live in `ole32`. We use them to figure out whether 'default' auth is available. It is however odd that we don't explicitly link to it. Perhaps one of the other libraries does so for us. But we should still link to it, since we use its functions.

CMakeLists.txt
```diff
@@ -339,8 +340,7 @@ IF (LIBSSH2_FOUND)
 	ADD_DEFINITIONS(-DGIT_SSH)
 	INCLUDE_DIRECTORIES(${LIBSSH2_INCLUDE_DIRS})
 	LINK_DIRECTORIES(${LIBSSH2_LIBRARY_DIRS})
-	LIST(APPEND LIBGIT2_PC_LIBS ${LIBSSH2_LDFLAGS})
-	#SET(LIBGIT2_PC_LIBS "${LIBGIT2_PC_LIBS} ${LIBSSH2_LDFLAGS}")
+	SET(LIBGIT2_PC_REQUIRES "${LIBGIT2_PC_REQUIRES} libssh2")
```
carlosmn: This might cause the user to link to a different libssh2 than the we one libgit2 was provided with. I think we have this form as a reaction to an issue which was reported. Why change this to require?

arthurschreiber: This allows pkg-config canto automatically identify the linker flags needed for libssh2 when statically linking against a libgit2 which was built with ssh support.

arthurschreiber:
> But we should still link to it, since we use its functions.

I added another commit to fix this.

CMakeLists.txt
carlosmn: I found the commit explaining why we do this: 790cabf. If we put them in requires, `pkg-config` will put *their* private libs in the linker args as well, which we don't want. We want to give tell whoever is embedding libgit2 which options we would have used, were we to create the shared library.

arthurschreiber: Mh, yeah, that sounds reasonable. Why do we do this for openssl and zlib then?

carlosmn: I think it just comes down to them not having any extra dependencies. zlib doesn't have dependencies, and OpenSSL only has zlib, which we depend on anyweay.

arthurschreiber:
@carlosmn Okay, I changed the issues with this and cleaned up the history.

PR #3573, opened 2016-01-12
Make packfile_unpack_compressed a private API

pmq20:
packfile_unpack_compressed has already been covered in pack.h


carlosmn:
Then it should be removed from `pack.h` since it's not a function that should be used outside of `pack.c`.

pmq20:
@carlosmn Thanks for reviewing. Changed according to your advice.

PR #3574, opened 2016-01-12
Buffer sideband packet data

chescock:
I'm experimenting with using LibGit2Sharp with [AWS CodeCommit](https://aws.amazon.com/codecommit/) over HTTPS, and one of the problems I've hit is that libgit2 is failing to parse the git-receive-pack-result.  libgit2 expects to read complete pkt-lines from each sideband packet, but CodeCommit is splitting the pkt-lines across multiple side-channel packets.  

The HTTP response looks like this (formatted as a C string literal with line breaks for clarity): 

```
"0009\x01000e"
"000f\x01unpack ok\n"
"0009\x010019"
"001a\x01ok refs/heads/master\n"
"0009\x010000"
"0006\x02\n"
"0000"
```

Core git accepts this format, and the git technical documentation describes sidebands as "streams", so I believe it's valid and libgit2 should accept it.

### DIFF
--- a/src/transports/smart_protocol.c
+++ b/src/transports/smart_protocol.c
@@ -719,30 +719,58 @@ static int add_push_report_pkt(git_push *push, git_pkt *pkt)
 	return 0;
 }
 
-static int add_push_report_sideband_pkt(git_push *push, git_buf *data_pkt_buf)
+static int add_push_report_sideband_pkt(git_push *push, git_pkt_data *data_pkt, git_buf *data_pkt_buf)
 {
 	git_pkt *pkt;
-	const char *line_end;
+	const char *line, *line_end;
+	size_t line_len;
 	int error;
+	int reading_from_buf = data_pkt_buf->size > 0;
 
-	while (data_pkt_buf->size > 0) {
-		error = git_pkt_parse_line(&pkt, data_pkt_buf->ptr, &line_end, data_pkt_buf->size);
+	if (reading_from_buf) {
+		/* We had an existing partial packet, so add the new
+		 * packet to the buffer and parse the whole thing */
+		git_buf_put(data_pkt_buf, data_pkt->data, data_pkt->len);
+		line = data_pkt_buf->ptr;
+		line_len = data_pkt_buf->size;
+	}
+	else {
+		line = data_pkt->data;
+		line_len = data_pkt->len;
+	}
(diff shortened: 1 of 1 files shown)


carlosmn:
When is it sending these half-packets? From your example it looks like it's during push? Is that the only time it would do this?

chescock:
Yes, this is during a push (sorry for being unclear).  I'm able to successfully clone, so I think this is only a problem for push.

chescock:
Is there anything else I should be doing to make progress on this?  I think it may have gotten forgotten because I submitted it while everyone was on vacation :).

ethomson:
Hi - very sorry for the delay here.  I agree that we need to support this case, but I do have a few questions, if you don't mind:

1. It looks like this change means that now we support packets where data packets are interleaved - so now the inner packet might be split across multiple sideband packets...  but can a progress packet come in between those sideband packets?  It looks like we are supporting that, but that seems very odd.  I wonder if we shouldn't validate that `data_pkt_buf.size == 0` unless `pkt-type == GIT_PKT_DATA`.

    It may also be the case that this is not important and we should ignore it.

2. It looks like every data packet will now have an associated malloc and two more `memcpy`s, even in the case where they're whole (which is the common case).  I suppose this would only be problematic with a very chatty server.  But I wonder if it's possible to only buffer the packet for the next read if it's a runt...

chescock:
> Can a progress packet come in between those sideband packets? It looks like we are supporting that, but that seems very odd. 

I haven't seen packets get interleaved like that, but I can't say for certain that they won't be.  I'd be inclined to accept them, since the interpretation seems clear and it leads to a simpler implementation.  Raising an error only feels necessary if there is a risk that we're sending incomplete or misleading data back to the caller, such as when the final packet is incomplete.  

It would be interesting to see how core git responds to an interleaved response, but I don't know enough to set up a server that would send one.

> I wonder if it's possible to only buffer the packet for the next read if it's a runt...

Sure, let me try to do that.  I don't normally develop in C, so I had started with the simplest implementation to make it less likely that I screwed something up :).

chescock:
@ethomson I changed the code to no longer do extra `malloc` or `memcpy` calls in the normal case of complete packets.  How does it look now?

carlosmn:
Thanks.

PR #3575, opened 2016-01-13
Remove duplicated calls to git_mwindow_close

pmq20:
git_mwindow_close has already been called git_packfile_unpack_header, needless to be called again from the outside.


carlosmn:
Thanks.

PR #3577, opened 2016-01-14
Add a new build flag to disable the pool allocator

rossdylan:
Calls to git_pool__malloc are proxied directly to git__malloc. Allocations are tracked using a git_vector and freed when git_pool_clear is called. I am currently using ifdef/ifndef to enable and disable the debug code. I also enable/disable a couple of the pool test cases which won't work when debug is turned on.

### DIFF
--- a/src/pool.h
+++ b/src/pool.h
@@ -9,8 +9,13 @@
 
 #include "common.h"
 
+#ifdef GIT_DEBUG_POOL
+#include "vector.h"
+#endif
+
 typedef struct git_pool_page git_pool_page;
 
+#ifndef GIT_DEBUG_POOL
 /**
  * Chunked allocator.
  *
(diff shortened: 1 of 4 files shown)


src/pool.h
```diff
@@ -9,8 +9,13 @@
 
 #include "common.h"
 
+#ifdef GIT_DEBUG_POOL
+#include "vector.h"
+#endif
```
carlosmn: This ifdef isn't necessary.

carlosmn:
Thanks, this looks pretty straightforward. Other than the ifdef'd include, I think we could include this.

rossdylan:
awesome, I just pushed a second commit which removes the ifdef.

src/pool.h
```diff
@@ -98,7 +124,9 @@ extern char *git_pool_strcat(git_pool *pool, const char *a, const char *b);
 /*
  * Misc utilities
  */
+#ifndef _DEBUG_POOL
```
ethomson: Should this be `GIT_DEBUG_POOL`?

rossdylan: yep, looks like I typo'd that.

ethomson:
Thanks for tackling this!

PR #3582, opened 2016-01-22
Validate pointer before access the member.

ivellioscolin:
When Git repository at network locations, sometimes git_iterator_for_tree
fails at iterator__update_ignore_case so it goes to git_iterator_free.
Null pointer will crash the process if not check.

Signed-off-by: Colin Xu <colin.xu@gmail.com>


ivellioscolin:
Odd appveyor failure.

PR #3583, opened 2016-01-23
Don't generate checkout conflicts for submodules.

hackhaslam:
Checkout doesn't ever actually modify submodules in the workdir (does it?), so it shouldn't generate conflicts. Currently checkout generates a conflict if the submodule is dirty or if the submodule commit doesn't match the baseline. Neither of these match the behavior of git.git.

This change removes that logic and assumes that submodules aren't modified by checkout.

### DIFF
--- a/docs/checkout-internals.md
+++ b/docs/checkout-internals.md
@@ -66,6 +66,8 @@ Key
 - Bi       - ignored blob (WD only)
 - T1,T2,T3 - trees with different SHAs,
 - Ti       - ignored tree (WD only)
+- S1,S2    - submodules with different SHAs
+- Sd       - dirty submodule (WD only)
 - x        - nothing
 
 Diff with 2 non-workdir iterators
@@ -162,6 +164,27 @@ Checkout From 3 Iterators (2 not workdir, 1 workdir)
 | 35+ |  T1 |     T2 |      x         | update locally deleted tree (SAFE+MISSING)                         |
 | 36* |  T1 |     T2 |  B1/Bi         | update to tree with typechanged tree->blob conflict (F-1)          |
 | 37  |  T1 |     T2 | T1/T2/T3       | update to existing tree (MAYBE SAFE)                               |
+| 38+ |   x |     S1 |      x         | add submodule (SAFE)                                               |
+| 39  |   x |     S1 |  S1/Sd         | independently added submodule (SUBMODULE)                          |
+| 40* |   x |     S1 |     B1         | add submodule with blob confilct (FORCEABLE)                       |
+| 41* |   x |     S1 |     T1         | add submodule with tree conflict (FORCEABLE)                       |
+| 42  |  S1 |      x |  S1/Sd         | deleted submodule (SUBMODULE)                                      |
+| 43  |  S1 |      x |      x         | independently deleted submodule (SUBMODULE)                        |
+| 44  |  S1 |      x |     B1         | independently deleted submodule with added blob (SAFE+MISSING)     |
+| 45  |  S1 |      x |     T1         | independently deleted submodule with added tree (SAFE+MISSING)     |
+| 46  |  S1 |     S1 |      x         | locally deleted submodule (SUBMODULE)                              |
+| 47+ |  S1 |     S2 |      x         | update locally deleted submodule (SAFE)                            |
+| 48  |  S1 |     S1 |     S2         | locally updated submodule commit (SUBMODULE)                       |
+| 49  |  S1 |     S2 |     S1         | updated submodule commit (SUBMODULE)                               |
+| 50+ |  S1 |     B1 |      x         | add blob with locally deleted submodule (SAFE+MISSING)             |
+| 51* |  S1 |     B1 |     S1         | typechange submodule->blob (SAFE)                                  |
+| 52* |  S1 |     B1 |     Sd         | typechange dirty submodule->blob (SAFE!?!?)                        |
+| 53+ |  S1 |     T1 |      x         | add tree with locally deleted submodule (SAFE+MISSING)             |
+| 54* |  S1 |     T1 |  S1/Sd         | typechange submodule->tree (MAYBE SAFE)                            |
+| 55+ |  B1 |     S1 |      x         | add submodule with locally deleted blob (SAFE+MISSING)             |
+| 56* |  B1 |     S1 |     B1         | typechange blob->submodule (SAFE)                                  |
+| 57+ |  T1 |     S1 |      x         | add submodule with locally deleted tree (SAFE+MISSING)             |
+| 58* |  T1 |     S1 |     T1         | typechange tree->submodule (SAFE)                                  |
 
 
 The number is followed by ' ' if no change is needed or '+' if the case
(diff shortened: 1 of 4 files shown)


srajko:
I used a similar fix (definitely a hack in my case) to get around the problem - limited to an oid mismatch only: https://github.com/srajko/libgit2/commit/e07639fb8c7936feb864efd54ac619ea9790ff22

Going beyond that and ignoring wd changes (like in this PR) is probably the right thing to do for the case when the submodule is there both before and after the checkout.  I'm not sure about other cases - mostly worried about the `GIT_DELTA_TYPECHANGE` case here - https://github.com/libgit2/libgit2/blob/e8feafe32007ebd16a61820c70abd221655d053c/src/checkout.c#L491 and `GIT_DELTA_DELETED` here - https://github.com/libgit2/libgit2/blob/e8feafe32007ebd16a61820c70abd221655d053c/src/checkout.c#L462

Also, will `GIT_CHECKOUT_UPDATE_SUBMODULES` and `GIT_CHECKOUT_UPDATE_SUBMODULES_IF_CHANGED` affect this in the future?

hackhaslam:
Yeah, I need to investigate what git does in those cases. I guess that there are actually a bunch of permutations to consider between baseline, target and working directory.

I saw `GIT_CHECKOUT_UPDATE_SUBMODULES` and `GIT_CHECKOUT_UPDATE_SUBMODULES_IF_CHANGED` and I wondered if the conflict logic was looking forward to those options getting added. If that's the case then maybe the current logic should just be hidden behind those flags instead of removed.

hackhaslam:
I added some notes to the checkout internals doc. The new cases describe how git treats submodules during checkout. I haven't yet tried to determine where the behavior of libgit2 differs. I'll do that next.

srajko:
@hackhaslam thanks for pushing this forward.  If there is anything I can help with to keep this going please let me know.

srajko:
@hackhaslam I found a test that goes through different type change scenarios (https://github.com/libgit2/libgit2/blob/master/tests/checkout/typechange.c).  I'm trying to see if we could expand it to give us some more coverage of this change.

srajko:
I expanded the test a bit (https://github.com/srajko/libgit2/commit/4ae849f0e6a4ab4bbc6abc13f19cea28243cee36) and the test continues to pass after the change from this PR - and I took the expected values before this change.

hackhaslam:
Thanks @srajko. So was the change you made in srajko@c558840 was already working before this PR?

srajko:
Correct, that seems to have been resolved at some point, so I put in the suggested change to get tighter coverage of this PR.  I submitted a PR for your PR in case you want to include that here :-)

The tighter tests give me more confidence that this PR doesn't change libgit2 behavior with checkouts across type changes involving submodules.  @ethomson @carlosmn is there anything else you would like to see to accept this PR?

srajko:
I just did some manual testing, and looks like after the PR change libgit2 is letting me do a safe checkout from a dirty submodule to a tree, losing the uncommitted changes.  So maybe we need to check for a type change, and if the type is changing report a conflict if the submodule is dirty?

hackhaslam:
Okay, I merged your PR into my branch. The type change from submodule to tree with a dirty submodule was number 54 in the notes I made. I marked it as MAYBE SAFE, although I can't remember why now. It seems like it should practically always be a conflict. Also number 52 which core git gets wrong should probably conflict. There are at a couple of others cases (40, 41) that don't involve a dirty submodules that should also conflict.

So yeah, I think those cases need to be addressed (and tested) before this can be merged.

srajko:
I got the test to cover the scenarios where a submodule is dirty - PRed to your PR.

With the expected values taken before the change in this PR, the tests now fail - the actual values of conflicts are exactly one less for each checkout involving a type change from a dirty submodule. Hopefully, updating the conflict logic to take into account type changes should make the test pass again.

@hackhaslam do you have any time in the near future to update the conflict logic? If not, I'd be happy to take a stab at it.

hackhaslam:
I'm on vacation away from my computer for the next couple of days. I can work on it then if you don't get to it first.

srajko:
Pushed a potential fix to https://github.com/stinb/libgit2/pull/2

hackhaslam:
@srajko sorry I'm so slow to respond. I think that something like I pushed in 488ade6 might be better. It keeps the is_workdir_modified logic the same and only checks for a submodule in the modified case. I rebased it all on the tip of master and dropped my initial commit. I'm still testing some of the other cases.

BTW, I remember now why I put MAYBE SAFE on the typechange to tree case. Git only generates a conflict if the trees intersect in some way. The submodule tree basically becomes untracked, then it gives an "untracked file would be overwritten by checkout" conflict if the new tree contains any files with the same name.

srajko:
@hackhaslam sorry now it was me that got swept away in other things :-)

I got the CI passing on a test PR - #3734.  https://github.com/libgit2/libgit2/pull/3734/commits/8da601d383b7cc9a3113e762494bf1ec59ee6896 resolves a build error with the test code I added, and you'll probably want to cherry-pick that into this PR (or I can PR it to your branch).  When I tried that on top of this PR though, the CI still failed on something. I did get it to pass on the other PR, but that one has a slightly different diff in `src/checkout.c` (there, the bypass code you added is only applied when both the `wditem` and the `newitem` are `GIT_FILEMODE_COMMIT`)

srajko:
@hackhaslam Let's finish this up :-)  I think we have all the pieces, we just need to get this PR finalized and passing.  Is there anything I can do to help?

hackhaslam:
@srajko So is there a case that doesn't work after the changes I made? If there is then it doesn't look like it's tested.

johnhaley81:
CI passes! <3 @hackhaslam and @srajko

srajko:
@hackhaslam perhaps the test failure I got was intermittent / unrelated, sorry for the false alarm.

I think this PR is ready.  We may want to squash https://github.com/libgit2/libgit2/pull/3583/commits/e756a84e08d4c5160917e8457ece47c483f07927 into https://github.com/libgit2/libgit2/pull/3583/commits/aee92e73ffa25f4c88d469e45fff618bcd1999cf just to clean up the history.  @carlosmn @ethomson anything else you'd like to see done here?

jdgarcia:
What's the status on this?

ethomson:
@srajko Sorry about the delay.  I really love that this is a one-line code change with so much docs and test around it.  ❤️ 

I will take a look at this later today to make sure that I understand all the documentation.  :)

hackhaslam:
I added new commit to address #3774. Merge has a check for modifications in the workdir that would be overwritten by the merge operation. I think that it's correct (and safe) to ignore submodules here as well.

@ethomson If this looks good I can add a changelog entry as well.

johnhaley81:
I think that failure in AppVeyor was due to networking issues. @ethomson can you rerun AppVeyor please?

johnhaley81:
Now that CI is 💚 is there anything else that this needs?

ethomson:
Sorry about the delay here; I'm manually merging this because it's been sitting for too long and it should be merged.  Apologies.  :/

In the future, we'd ask that you squash this sort of thing up into nice, standalone commits.  But I'm going to go ahead and do this one because we've been so tardy in taking care of it.

Thanks again, this is a big help.  :)

PR #3586, opened 2016-01-25
Fix the build when defining USE_NSEC

Therzok:
This fixes the build on Windows machines, I'm not sure about other kinds of machines.

One other issue I have observed on OS X:
https://github.com/libgit2/libgit2/blob/master/src/index.c#L831-L834

They're typo'd, they should be st_mtime, st_ctime, but on Darwin, they should be st_mtimespec, st_ctimespec.

### DIFF
--- a/src/index.h
+++ b/src/index.h
@@ -92,7 +92,7 @@ GIT_INLINE(bool) git_index_entry_newer_than_index(
 
 	/* If the timestamp is the same or newer than the index, it's racy */
 #if defined(GIT_USE_NSEC)
-	if ((int32_t)index->stamp.tv_sec < entry->mtime.seconds)
+	if ((int32_t)index->stamp.mtime.tv_sec < entry->mtime.seconds)
 		return true;
 	else if ((int32_t)index->stamp.mtime.tv_sec > entry->mtime.seconds)
 		return false;


jacquesg:
#3514 solved this problem in {{CMakeLists.txt}} and {{fileops.c}}

Therzok:
Hmmmm, something's wrong then. :p

ethomson:
What build problem?  How are you building that you're seeing a problem here?  Are you not using cmake?  Are you using MSVC?  If so, what version?  Platform?  Etc etc...

Therzok:
`cmake -G "Visual Studio 14" -D ENABLE_TRACE=ON -D "LIBGIT2_FILENAME=$binaryFilename" -DUSE_NSEC=ON -DSTDCALL=ON -DBUILD_CLAR:BOOL=OFF`

For some reason, the cmake magic isn't happening.

jacquesg:
Well ... CMake will only give you ```GIT_USE_STAT_ATIMESPEC```, I think something similar to what is done in ```fileops.c``` is required: https://github.com/libgit2/libgit2/blob/master/src/fileops.c#L16-L20

It would be nice though to not have to duplicate this...

ethomson:
Yes - thanks.  There's actually two issues here - you are very correct that this change is needed.  I think that @jacquesg noticed a separate issue, which is that #3514 needs to grow in scope to affect some other things.

I took this patch over into #3604 which includes a fix for the latter as well.

Thanks!

Therzok:
Glad to see this unearthed.

PR #3589, opened 2016-01-27
less vague uploadpack and receivepack error messages

yofreke:
Trying to identify an error as an end user is difficult when you get the error message
`Must call UPLOADPACK_LS before UPLOADPACK`
since this error message is the same for a git or ssh error.

Added `(git)` and `(ssh)` to the end of the error messages for easier identification.

### DIFF
--- a/src/transports/git.c
+++ b/src/transports/git.c
@@ -240,7 +240,7 @@ static int _git_uploadpack(
 		return 0;
 	}
 
-	giterr_set(GITERR_NET, "Must call UPLOADPACK_LS before UPLOADPACK");
+	giterr_set(GITERR_NET, "Must call UPLOADPACK_LS before UPLOADPACK (git)");
 	return -1;
 }
 
@@ -296,7 +296,7 @@ static int _git_receivepack(
 		return 0;
 	}
 
-	giterr_set(GITERR_NET, "Must call RECEIVEPACK_LS before RECEIVEPACK");
+	giterr_set(GITERR_NET, "Must call RECEIVEPACK_LS before RECEIVEPACK (git)");
 	return -1;
 }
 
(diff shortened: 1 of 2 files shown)


ethomson:
When you say "end user", do you mean as a developer building a tool using libgit2?  Or do you mean an end user of your tool?

This message looks like something that would only be surfaced as a result of a programmatic error, is that correct?

carlosmn:
The end user seeing this would mean programmer error. How come this does more to identify the protocol used than asking for the URL? The code you're writing would be the same anyhow if you're calling it out of order, so the protocol shouldn't matter.

ethomson:
Indeed - if we're going to change this, I would prefer to just `abort` instead of tweaking error messages that nobody will ever see.  Please do feel free to reopen if we've misunderstood the problem.

yofreke:
>When you say "end user", do you mean as a developer building a tool using libgit2? Or do you mean an end user of your tool?

In this case it was both.  I was using nodegit to clone a repository, and was getting this message bubbled up to my console.  Curious as to the error I started digging around in nodegit, and then here.

>This message looks like something that would only be surfaced as a result of a programmatic error, is that correct?

The error is still not clear, seems like the cause might have been trying to connect a remote twice in nodegit.  Is this an error in nodegit or libgit; not sure (one of them isn't handling trying to connect an already connected remote).

>instead of tweaking error messages that nobody will ever see.

I am a fan of having very clear error messages bubbling up to _any_ user (makes developing and makes supporting easier).  Also not a great justification for closing the issue, since someone did see d: .  There is definitely an issue with having two error messages with the exact same text.

>if we're going to change this, I would prefer to just abort 

I would be happy to add in abort logic here; is there an example somewhere else that I can draw from?

MichaelRo:
Having the same problem

PR #3596, opened 2016-02-06
Bug on WindowsXP

burantino:
WideCharToMultiByte does not support WC_ERR_INVALID_CHARS flag on WindowsXp, so we need to fix w32_buffer.c not to use this flag straightforward but with the help of get_wc_flags() function. This function(get_wc_flags) is in src/win32/utf-conv.c but it is not exposed by header, so I decided to expose it.

### DIFF
--- a/src/win32/utf-conv.c
+++ b/src/win32/utf-conv.c
@@ -8,7 +8,7 @@
 #include "common.h"
 #include "utf-conv.h"
 
-GIT_INLINE(DWORD) get_wc_flags(void)
+DWORD get_wc_flags(void)
 {
 	static char inited = 0;
 	static DWORD flags;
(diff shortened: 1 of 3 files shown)


ethomson:
libgit2 has not supported Windows XP since Microsoft dropped support for it.  There are many other places in the code base that assume Vista or better.

burantino:
Ok (will know it:)) The only thing I want to say - this fix was a must and sufficient to work with local-only repo on winxp, so may be it be useful for someone.

carlosmn:
Should we get rid of `get_wc_flags()` entirely then? Its only purpose seems to be to support XP.

ethomson:
Yes, that function is only for Windows XP support.

vmg:
Let's remove it then!

ethomson:
Sorry, as noted, Windows XP is not supported any longer.

PR #3597, opened 2016-02-07
Filter registration

ethomson:
While reviewing #3564 , @carlosmn noticed that we were a bit racy in registration of merge drivers.  This registration pattern was primarily copied from filter registration, so I wanted to fix up the existing code before banging on that.

Previously we would set the global filter registry structure before adding filters to the structure, without a lock, which is quite racy. Now, register default filters during global registration and use an rwlock to read and write the filter registry (as appropriate)

### DIFF
--- a/src/openssl_stream.c
+++ b/src/openssl_stream.c
@@ -31,6 +31,115 @@
 #include <openssl/x509v3.h>
 #include <openssl/bio.h>
 
+SSL_CTX *git__ssl_ctx;
+
+#ifdef GIT_THREADS
+
+static git_mutex *openssl_locks;
+
+static void openssl_locking_function(
+	int mode, int n, const char *file, int line)
+{
+	int lock;
+
+	GIT_UNUSED(file);
+	GIT_UNUSED(line);
+
+	lock = mode & CRYPTO_LOCK;
+
+	if (lock) {
+		git_mutex_lock(&openssl_locks[n]);
+	} else {
+		git_mutex_unlock(&openssl_locks[n]);
+	}
+}
+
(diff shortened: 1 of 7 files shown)


ethomson:
Okay - I nailed down the problems on Windows; it turns out that my refactoring was a bit overly aggressive when it came to the windows CRT debug functions.  Those *must* be first, before your first allocation.  This is quite sensible, but not something that had occurred to me.

In any case, I think this nicely tightens up the global setup and teardown, and makes our threads safer for the filters.

This should be ready for review.

src/openssl_stream.c
```diff
@@ -31,6 +31,115 @@
...
+		return -1;
+	}
+#endif
+
+	git__on_shutdown(shutdown_ssl);
+
+	return 0;
+}
+
+int git_openssl_set_locking(void)
```
carlosmn: Thing is... this function isn't really about the streams. It's there to stop bindings from crashing when they use multiple threads without having to conditionally link to OpenSSL themselves.

ethomson: Yes but short of deleting it, t this still seems to be the most logical file to move this function to.

stanhu:
Does a similar change have to be made with the diff driver registry?

ethomson:
What diff driver registry?

stanhu:
I may be mistaken, but I thought this was it:

https://github.com/libgit2/libgit2/blob/master/src/diff_driver.h#L13

ethomson:
Ah - no, those are per-repository and so have very different lifecycles.

PR #3598, opened 2016-02-08
Coverity fixes

pks-t:
Two fixes for issues displayed by Coverity.


src/curl_stream.c
```diff
@@ -79,6 +79,7 @@ static int curls_certificate(git_cert **out, git_stream *stream)
 	for (slist = certinfo->certinfo[0]; slist; slist = slist->next) {
 		char *str = git__strdup(slist->data);
 		GITERR_CHECK_ALLOC(str);
+		git_vector_insert(&strings, str);
```
vmg: Hmmpf. This is a good catch.

carlosmn: *sigh* this probably explains why I wasn't seeing information from curl; thanks

PR #3599, opened 2016-02-08
Introduce git_commit_extract_signature

carlosmn:
This returns the GPG signature for a commit and its contents without the
signature block, allowing for the verification of the commit's
signature.

---

This code is already written to expect to be refactored into an internal function which accepts other header fields to use for the signature. I wonder if we should just make the public function accept a string directly so we don't have to worry about timing if/when git introduces new ones. And it would let fancy users have their own crypto in there.

/cc @mastahyeti

### DIFF
--- a/include/git2/commit.h
+++ b/include/git2/commit.h
@@ -264,6 +264,18 @@ GIT_EXTERN(int) git_commit_nth_gen_ancestor(
 GIT_EXTERN(int) git_commit_header_field(git_buf *out, const git_commit *commit, const char *field);
 
 /**
+ * Extract the signature from a commit
+ *
+ * @param signature the signature block
+ * @param signed_data signed data; this is the commit contents minus the signature block
+ * @param repo the repository in which the commit exists
+ * @param commit_id the commit from which to extract the data
+ * @param field the name of the header field containing the signature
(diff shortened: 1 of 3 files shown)


vmg:
This is fantastic @carlosmn :angel: 

>  I wonder if we should just make the public function accept a string directly so we don't have to worry about timing if/when git introduces new ones. And it would let fancy users have their own crypto in there.

Let's do this. We can export a constant for the default GPG header.

carlosmn:
The function now takes the field name. The `#define` feels a bit awkward since we don't seem to have anything quite like it anywhere else, though.

vmg:
Looks good! The define does seem a tad weird. Maybe an `extern const char *` instead?

carlosmn:
We don't really do that either for other defines, even the `GIT_REFS_` ones are all internal. I'm wondering if we should remove it instead.

vmg:
@carlosmn: OK, let's remove it, but can we have it default to a sane behavior (the GPG header) if you pass NULL as the argument?

carlosmn:
Ok, done.

vmg:
:sparkles:

brianmario:
💥

PR #3602, opened 2016-02-09
commit: also match the first header field when searching

carlosmn:
We were searching only past the first header field, which meant we were
unable to find e.g. `tree` which is the first field.

While here, make sure to set an error message in case we cannot find the
field.

---
I'm not sure wtf I was doing when I wrote the first one, but that code was stupid.

### DIFF
--- a/src/commit.c
+++ b/src/commit.c
@@ -564,41 +564,45 @@ int git_commit_nth_gen_ancestor(
 
 int git_commit_header_field(git_buf *out, const git_commit *commit, const char *field)
 {
-	const char *buf = commit->raw_header;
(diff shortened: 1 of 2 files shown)


vmg:
:+1:

mastahyeti:
This broke the parsing of the last header, if that header doesn't span multiple lines. [Here's](https://github.com/libgit2/libgit2/compare/master...mastahyeti:header_example) a failing test case.

carlosmn:
Man, that code was messy... I've pushed up a fix for that, testing on a non-multiline last header field.

mastahyeti:
Thanks

PR #3603, opened 2016-02-09
Coverity fixes

pks-t:
Some more small-ish fixes. I'm still unsure if it's of any benefit to fix memory leaks deep inside the internal APIs, but I guess it can't really hurt to do so.

On another note, I'd like to mark all those errors regarding GITERR_CHECK_ALLOC as false positives or intended in Coverity. Any objections to this?

### DIFF
--- a/src/attr_file.c
+++ b/src/attr_file.c
@@ -123,7 +123,7 @@ int git_attr_file__load(
 		break;
 	}
 	case GIT_ATTR_FILE__FROM_FILE: {
-		int fd;
+		int fd = -1;
 
 		/* For open or read errors, pretend that we got ENOTFOUND. */
 		/* TODO: issue warning when warning API is available */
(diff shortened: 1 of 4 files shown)


carlosmn:
If you can get coverity to consider `GITERR_CHECK_ALLOC` like an abort, that would be quite useful as it's the largest source of false positives IIRC.

pks-t:
Well, usually this can be done with Coverity annotations. I doubt though that they apply to macros. I've just set up Coverity for my repository to test if pks-t/libgit2@99403320c9332e04f31fa9859bfd5b22323c7082 suppresses these issues or not. But I'd bet that it won't fix the issue.

ethomson:
> If you can get coverity to consider GITERR_CHECK_ALLOC like an abort

Now that you mention it, yeah, something like:

```
#ifdef COVERITY
#define GITERR_ALLOC_CHECK(x) if(!x) abort()
#endif
```

might be sufficient.

PR #3604, opened 2016-02-09
Handle `USE_NSECS` 

ethomson:
Include @Therzok 's fix from #3586 which was missed after some index timestamp refactoring.

Also increase the scope of `st_mtim` -> `st_mtimespec` so that it affects all the files.


jacquesg:
:+1:

Therzok:
:+1:

PR #3607, opened 2016-02-10
Coverity improvements for GITERR_CHECK_ALLOC

pks-t:
Two small improvements to Coverity. The first commit tries to mitigate the issue with `GITERR_CHECK_ALLOC` by providing a `user_nodefs.h` file which overrides the macro. The second is a security enhancement stopping us from sending over the Coverity authorization token in plain text.

This is a follow-up for libgit2/libgit2#3603. I was not able to verify that the `user_nodefs.h` thing suppresses the resource leak errors as I cannot get Coverity to provide the analysis results to me. It builds, but after submitting the instrumentalized archive I get no results back.

### DIFF
--- a/script/coverity.sh
+++ b/script/coverity.sh
@@ -50,10 +50,9 @@ COVERITY_UNSUPPORTED=1 \
 tar czf libgit2.tgz cov-int
 SHA=$(git rev-parse --short HEAD)
 curl \
-	--form project=libgit2 \
 	--form token="$COVERITY_TOKEN" \
 	--form email=bs@github.com \
 	--form file=@libgit2.tgz \
 	--form version="$SHA" \
 	--form description="Travis build" \
-	http://scan5.coverity.com/cgi-bin/upload.py
+	https://scan.coverity.com/builds?project=libgit2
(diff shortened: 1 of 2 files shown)


ethomson:
This seems fairly magical.  :)  Should we just merge this so that it will get picked up on the next Coverity run and we'll see if it worked and iterate from there?

/cc @carlosmn

carlosmn:
Let's try it out.

pks-t:
Well, I'd say it did work out quite good :) Down to a defect density of 0.74 from something above 1, if I remember correctly.

2016-02-11-224552_700x306_scrot

carlosmn:
Yep, it reports 47 closed issues, and I don't see those noisy malloc-error issues in the list :dancers: :balloon:

ethomson:
This.  Is.  Amazing.

Thanks so much for this @pks-t , I am thrilled that that noise is gone.

pks-t:
Glad you're so thrilled ;) There are some more things I'd like to fix up with Coverity, so expect some pull requests next week.

Issue #3576, opened 2016-01-14
fs_iterator crashes (still) in the presence of nested empty directories


mhagger:
It appears that fix #3426 for problem #3056 wasn't quite complete. It is still possible to get unnecessary recursion and even stack overflow if the `refs/` directory contains many almost-empty directories (directories that only contain other empty directories). We've hit this problem in the real world, though admittedly in extreme repositories.

I was able to reproduce this error using [this script](https://gist.github.com/mhagger/46fd6d3dd835f089c512). It creates tons of directories in one of three different patterns. To use it, copy it into your libgit2 working directory and use it to gunk up the test repository `tests/resources/testrepo` with directories in the `--two-level` pattern:

```
$ ./gunk-up-test-repo --two-level tests/resources/testrepo/.gitted 100000
Creating approximately 100000 repetitions
Created 201012 directories and 0 files
$ cmake --build build && ./build/libgit2_clar -scheckout::index
[ 21%] Built target git2
[100%] Built target libgit2_clar
Loaded 304 suites: 
Started
......Segmentation fault (core dumped)
```

For example, the problem occurs if you have a lot of directories like

    refs/foo/1/
    refs/foo/1/d/
    refs/foo/2/
    refs/foo/2/d/
    ...

which are otherwise empty.

The other patterns, `--one-level` and `--with-ref`, don't lead to problems even with double the number of repetitions. For example, if the directories are only one-level deep, like

    refs/foo/1/
    refs/foo/2/
    ...

, then there is no problem. (Well, there's an extra recursion for directory `refs/foo`, but since it's only a single recursion there's enough space on the C stack for it.)

When the problem happens, the call stack looks like

```
#79 0x00007ff2ddf7796d in fs_iterator__advance_into (entry=0x0, iter=<optimized out>)
    at .../rugged-0.24.0b11/vendor/libgit2/src/iterator.c:1425
#80 0x00007ff2ddf77b17 in fs_iterator__update_entry (fi=0x7ff2e9202160)
    at .../rugged-0.24.0b11/vendor/libgit2/src/iterator.c:1573
#81 0x00007ff2ddf776f3 in fs_iterator__expand_dir (fi=0x7ff2e9202160)
    at .../rugged-0.24.0b11/vendor/libgit2/src/iterator.c:1386
#82 0x00007ff2ddf7796d in fs_iterator__advance_into (entry=0x0, iter=<optimized out>)
[...]

PR #3609, opened 2016-02-10
[WIP] Status within subdir

joshaber:
This is a failing test where we get the status within a subdir and it reports changes that don’t exist.

### DIFF
--- a/src/iterator.c
+++ b/src/iterator.c
@@ -109,15 +109,26 @@ static iterator_pathlist__match_t iterator_pathlist__match(
 	 * don't know that yet, since we're avoiding a stat unless it's necessary)
 	 * so see if the pathlist contains a file beneath this directory.
 	 */
+	idx = 0;
 	while ((p = git_vector_get(&iter->pathlist, idx)) != NULL) {
-		if (iter->prefixcomp(p, path) != 0)
+		size_t p_len;
+		int reversed, cmp, try_dir;
+
+		p_len = strlen(p);
+		reversed = path_len > p_len && p[p_len-1] == '/' && path[p_len-1] == '/';
+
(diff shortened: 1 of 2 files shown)


carlosmn:
I've narrowed down where this happens, but I'm still confused as to why.

For one we're re-using the index counter in `iterator_pathlist__match()` so we can end up missing the path we're after. But even when fixing that, we failt to match the prefix, even though we should.

For the tree-to-index iterator, we pass the arguments to the prefix comparison the wrong way around, so we check whether "zzz_new_dir/new_file" is a prefix of "zzz_new_dir".

In the case where we don't disable pathspec matching, we skip this function entirely so presumably that's why it works in that case. If you pass in exact paths, we find it in a bsearch and simply return that it matches. Just swapping the arguments in the comparison seems to break the worktree-to-index iterator, since in that case we actually have directories rather than full paths which does match correctly.

carlosmn:
I've made this work for this test (and a few around it, let's see what CI says about the rest) but this is pretty terrible code. We're switching what we're doing depending on which side we detect is longer and is likely a directory...

carlosmn:
Unsuprisingly this breaks other things. We should probably just claim directories aren't supported and figure out if we do want that featur (which we probably do).

ethomson:
Yeah.  Faced with this issue and #3576, it's time for me to spend a little time digging in to iterators and refactoring them.

In the mean time, please avoid `ITERATOR_PATHLIST_NONE` when using tree iterators and directories as it appears that it will most certainly fail.  I'll ship you the [speedup loop](http://thedailywtf.com/articles/The-Speedup-Loop) for this soon.

joshaber:
Should I close this out then, and open an issue?

ethomson:
That would be great.  Please assign it to me, I'm working on refactoring this now.

joshaber:
:boom:

PR #3610, opened 2016-02-10
rebase: introduce bare rebasing

ethomson:
Introduce the ability to rebase in-memory or in a bare repository.

When `rebase_options.bare` is specified, the resultant `git_rebase`
session will not be persisted to disk.  Callers may still analyze
the rebase operations, resolve any conflicts against the in-memory
index and create the commits.  Neither `HEAD` nor the working
directory will be updated during this process.

I played with a few ways of tackling this - right now we have the concept of a `git_rebase_type_t` that indicates the rebase method (`apply`, `merge` or `interactive`).  I originally thought that this should be another method (`inmemory` or `bare`), but this seems to be a different axis.  Since `apply` and `merge` may yield different results, you probably want to be able to have bare forms of each (and may in fact want an interactive in-memory rebase, ick.)

### DIFF
--- a/include/git2/rebase.h
+++ b/include/git2/rebase.h
@@ -58,6 +58,11 @@ typedef struct {
 	const char *rewrite_notes_ref;
 
 	/**
+	 * Options to control how trees are merged during `git_rebase_next`.
+	 */
+	git_merge_options merge_options;
+
+	/**
 	 * Options to control how files are written during `git_rebase_init`,
 	 * `git_checkout_next` and `git_checkout_abort`.  Note that a minimum
 	 * strategy of `GIT_CHECKOUT_SAFE` is defaulted in `init` and `next`,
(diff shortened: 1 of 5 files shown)


vmg:
This is real fucking neato. I think the extra flag makes sense, `bare` is mostly unrelated to the actual rebase method.

@carlosmn any thoughts before merging?

carlosmn:
I'm not a fan of the "bare" name. The bareness of the repository is orthogonal to us wanting to do an in-memory rebase. I'm taking a look at the diff, but the name is something that jumped out at me as I saw it.

vmg:
Yes, maybe call it `in_memory` or the likes.

carlosmn:
It looks good otherwise, I'm glad we can even resolve conflicts with this.

ethomson:
Fixed.  I also added `git_merge_options` to the `git_rebase_options` struct so that we can pass in fancy flags like `GIT_MERGE_FAIL_ON_CONFLICT`.

vmg:
> I also added git_merge_options to the git_rebase_options struct so that we can pass in fancy flags like GIT_MERGE_FAIL_ON_CONFLICT.

That's some prescience right there. :ok_hand:

PR #3611, opened 2016-02-11
Fix some race conditions in concurrent initialization

alexcrichton:
Some of the travis tests I have for the [git2-rs](https://github.com/alexcrichton/git2-rs) Rust library for libgit2 have been failing a lot recently with the error:

```
src/global.c:38: git__on_shutdown: Assertion `count <= 8 && count > 0' failed.
```

Curious as to why this was happening, I scrutinized some of the calls to `git__on_shutdown`, and I think the latter of these two commits may be the cause of this. Unfortunately I wasn't ever able to reproduce locally, so I'm not *sure* that this fixes the bug I'm seeing, but it seems like these may want to be fixed regardless!

### DIFF
--- a/src/filter.c
+++ b/src/filter.c
@@ -89,7 +89,7 @@ static void filter_registry_shutdown(void)
 static int filter_registry_initialize(void)
 {
 	int error = 0;
-	struct filter_registry *reg;
+	struct filter_registry *reg, *found;
 
 	if (git__filter_registry)
 		return 0;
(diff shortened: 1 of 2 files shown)


ethomson:
What do you mean exactly by "concurrent initialization"...  Are you running `git_libgit2_init` from multiple threads?

alexcrichton:
Ah sorry, by concurrent initialization I meant on behalf of the threads calling these functions, for example if two threads call `filter_registry_initialize` or `git_sysdir_get` I think these situations will arise.

carlosmn:
The filter initialisation looks correct, though @ethomson has done a larger refactor around them in #3597.

It does look like we would need better synchronization around `git_sysdir_get()` but this doesn't look like it's enough. This would still call the guess function multiple times. I think what this needs is a mutex which we enter once we notice that we haven't set the sysdir yet. Upon entering the mutex we would check again and set if we are the first ones. This would be safe and still allow the normal path to be fast. We do something similar in the packfille code.

alexcrichton:
Ok, sounds like some more invasive changes are probably needed here, so closing.

PR #3612, opened 2016-02-11
Horrible fix for #3173.

arthurschreiber:
This should fix #3173, https://github.com/libgit2/rugged/issues/566 and https://github.com/libgit2/rugged/pull/42

@ethomson Please take a look. I don't think this really fixes all cases where this might segfault, but it fixes the two cases that were reported by rugged users.

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -2521,7 +2521,8 @@ int git_checkout_iterator(
 
 	if (data.opts.baseline_index) {
 		if ((error = git_iterator_for_index(
-				&baseline, data.opts.baseline_index, &baseline_opts)) < 0)
+				&baseline, git_index_owner(data.opts.baseline_index),
+				data.opts.baseline_index, &baseline_opts)) < 0)
 			goto cleanup;
 	} else {
 		if ((error = git_iterator_for_tree(
@@ -2633,7 +2634,7 @@ int git_checkout_index(
 		return error;
 	GIT_REFCOUNT_INC(index);
 
-	if (!(error = git_iterator_for_index(&index_i, index, NULL)))
+	if (!(error = git_iterator_for_index(&index_i, repo, index, NULL)))
 		error = git_checkout_iterator(index_i, index, opts);
 
 	if (owned)
--- a/src/index.c
+++ b/src/index.c
@@ -2907,8 +2907,8 @@ int git_index_read_index(
 
 	opts.flags = GIT_ITERATOR_DONT_IGNORE_CASE;
 
-	if ((error = git_iterator_for_index(&index_iterator, index, &opts)) < 0 ||
-		(error = git_iterator_for_index(&new_iterator, (git_index *)new_index, &opts)) < 0)
+	if ((error = git_iterator_for_index(&index_iterator, git_index_owner(index), index, &opts)) < 0 ||
+		(error = git_iterator_for_index(&new_iterator, git_index_owner(new_index), (git_index *)new_index, &opts)) < 0)
 		goto done;
 
 	if (((error = git_iterator_current(&old_entry, index_iterator)) < 0 &&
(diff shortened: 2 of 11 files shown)


ethomson:
Haha, this doesn't really look that horrible.  Certainly I don't see a better way to do it.

arthurschreiber:
@ethomson Thanks for taking a look!

src/checkout.c
```diff
@@ -2521,7 +2521,8 @@ int git_checkout_iterator(
 
 	if (data.opts.baseline_index) {
 		if ((error = git_iterator_for_index(
-				&baseline, data.opts.baseline_index, &baseline_opts)) < 0)
+				&baseline, git_index_owner(data.opts.baseline_index),
```
arthurschreiber: I checked all calls to  `git_checkout_iterator`, and `git_index_owner(data.opts.baseline_index)` will be non-null in all cases, but we could also change `git_checkout_iterator` to take an additional `repo` param.

ethomson: Imma hit up some refactoring of iterators, so I didn't want to bikeshed too hard on this yet.  That's a good idea though that I'll keep in mind for working on this.

src/index.c
```diff
@@ -2907,8 +2907,8 @@ int git_index_read_index(
 
 	opts.flags = GIT_ITERATOR_DONT_IGNORE_CASE;
 
-	if ((error = git_iterator_for_index(&index_iterator, index, &opts)) < 0 ||
-		(error = git_iterator_for_index(&new_iterator, (git_index *)new_index, &opts)) < 0)
+	if ((error = git_iterator_for_index(&index_iterator, git_index_owner(index), index, &opts)) < 0 ||
+		(error = git_iterator_for_index(&new_iterator, git_index_owner(new_index), (git_index *)new_index, &opts)) < 0)
```
arthurschreiber: I believe that `git_index_owner(new_index)` can return `NULL` here, but `git_index_owner(index)` will not. Should we use `git_index_owner(index)` in both cases here?

src/merge.c
```diff
@@ -2083,9 +2083,9 @@ static int iterator_for_annotated_commit(
 	opts.flags = GIT_ITERATOR_DONT_IGNORE_CASE;
 
 	if (commit == NULL) {
-		error = git_iterator_for_nothing(out, &opts); 
+		error = git_iterator_for_nothing(out, &opts);
 	} else if (commit->type == GIT_ANNOTATED_COMMIT_VIRTUAL) {
-		error = git_iterator_for_index(out, commit->index, &opts);
+		error = git_iterator_for_index(out, git_index_owner(commit->index), commit->index, &opts);
```
arthurschreiber: I don't think this can be `git_index_owner(commit->index)`, but we could also change `iterator_for_annotated_commit` to take an additional `repo` argument easily.

src/pathspec.c
```diff
@@ -550,7 +550,7 @@ int git_pathspec_match_index(
 
 	iter_opts.flags = pathspec_match_iter_flags(flags);
 
-	if (!(error = git_iterator_for_index(&iter, index, &iter_opts))) {
+	if (!(error = git_iterator_for_index(&iter, git_index_owner(index), index, &iter_opts))) {
```
arthurschreiber: This again is prone to segfaulting if the index passed to `git_pathspec_match_index` is unowned.

src/submodule.c
```diff
@@ -327,7 +327,7 @@ static int submodules_from_index(git_strmap *map, git_index *idx, git_config *cf
        const git_index_entry *entry;
        git_buf name = GIT_BUF_INIT;
 
-       if ((error = git_iterator_for_index(&i, idx, NULL)) < 0)
+       if ((error = git_iterator_for_index(&i, git_index_owner(idx), idx, NULL)) < 0)
```
arthurschreiber: `git_index_owner(idx)` will never return `NULL` here, as the only place `idx` can be coming from is through `git_repository_index(&idx, repo)`.

arthurschreiber:
I went through all the cases where we're still calling `git_index_owner` to get to the owner of the index, and there's only one case where this could return `NULL`.

PR #3613, opened 2016-02-12
Remove most of the silly warnings

ethomson:
Fix up some of the silly warning:  correct casts, constness, etc.  This doesn't get us to zero warnings on all platforms, but it gets us closer.

### DIFF
--- a/src/tree.c
+++ b/src/tree.c
@@ -17,6 +17,9 @@
 #define DEFAULT_TREE_SIZE 16
 #define MAX_FILEMODE_BYTES 6
 
+#define TREE_ENTRY_CHECK_NAMELEN(n) \
+	if (n > UINT16_MAX) { giterr_set(GITERR_INVALID, "tree entry path too long"); }
+
 GIT__USE_STRMAP
 
 static bool valid_filemode(const int filemode)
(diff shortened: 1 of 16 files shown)


src/tree.c
```diff
@@ -135,7 +135,7 @@ static git_tree_entry *alloc_entry(const char *filename)
 
 struct tree_key_search {
 	const char *filename;
-	uint16_t filename_len;
+	size_t filename_len;
```
carlosmn: This makes the `filename_len` field a different size for the field in the `git_tree_entry` which I'd rather avoid.

ethomson: Hm, yes that's true.  A quick perusal of the git source shows that they use an [`unsigned int`](https://github.com/git/git/blob/563e38491eaee6e02643a22c9503d4f774d6c5be/tree-walk.h#L13) here for the tree entry length.

There's an obvious compatibility issue here, then.  And although I'm not surprised that we haven't run into it yet, one can certainly imagine a world where somebody is using git as a general data exchange mechanism (perhaps not directly related to filesystem or filename sizes) and a `uint16_t` is too small to hold the tree entries.  So we should probably increase `git_tree_entry`'s size as well.

carlosmn: We only just moved to 16 bits. This is in part to reduce the size and improve the alignment of the tree entries, and partly to make sure we can allocate the entries in the pool. We figured that 16 bits is more than actual OSs allow you to use for file names, and we're not going to go out of our way to support people who should be using actual databases.

ethomson: Updated to use `uint16_t`.

PR #3614, opened 2016-02-15
Coverity fixes

pks-t:
This PR fixes another 14 issues displayed by Coverity, bringing the total count down from 85 defects to 71.

### DIFF
--- a/script/user_nodefs.h
+++ b/script/user_nodefs.h
@@ -23,3 +23,5 @@
 	if (GIT_MULTIPLY_SIZET_OVERFLOW(out, nelem, elsize)) { __coverity_panic__(); }
 
 #nodef GITERR_CHECK_VERSION(S,V,N) if (giterr__check_version(S,V,N) < 0)  { __coverity_panic__(); }
+
+#nodef LOOKS_LIKE_DRIVE_PREFIX(S) (strlen(S) >= 2 && git__isalpha((S)[0]) && (S)[1] == ':')
--- a/script/user_nodefs.h
+++ b/script/user_nodefs.h
@@ -6,3 +6,20 @@
  */
 
 #nodef GITERR_CHECK_ALLOC(ptr) if (ptr == NULL) { __coverity_panic__(); }
+
+#nodef GITERR_CHECK_ALLOC_ADD(out, one, two) \
+	if (GIT_ADD_SIZET_OVERFLOW(out, one, two)) { __coverity_panic__(); }
+
+#nodef GITERR_CHECK_ALLOC_ADD3(out, one, two, three) \
+	if (GIT_ADD_SIZET_OVERFLOW(out, one, two) || \
+		GIT_ADD_SIZET_OVERFLOW(out, *(out), three)) { __coverity_panic__(); }
+
+#nodef GITERR_CHECK_ALLOC_ADD4(out, one, two, three, four) \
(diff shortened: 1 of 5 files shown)


pks-t:
Eh, fail. Submitted a commit that shouldn't be contained at all. Will fix

script/user_nodefs.h
```diff
@@ -6,3 +6,22 @@
...
+	if (GIT_ADD_SIZET_OVERFLOW(out, one, two) || \
+		GIT_ADD_SIZET_OVERFLOW(out, *(out), three) || \
+		GIT_ADD_SIZET_OVERFLOW(out, *(out), four)) { __coverity_panic__(); }
+
+#nodef GITERR_CHECK_ALLOC_MULTIPLY(out, nelem, elsize) \
+	if (GIT_MULTIPLY_SIZET_OVERFLOW(out, nelem, elsize)) { __coverity_panic__(); }
+
+#nodef GITERR_CHECK_VERSION(S,V,N) if (giterr__check_version(S,V,N) < 0)  { __coverity_panic__(); }
+
+#nodef LOOKS_LIKE_DRIVE_PREFIX(S) (strlen(S) > 2= && git__isalpha((S)[0]) && (S)[1] == ':')
```
carlosmn: Just want to double-check that `trlen(S) > 2=` is the correct coverity syntax, since it looks like a mistyped `trlen(S) >= 2`

src/transports/smart_protocol.c
```diff
@@ -108,6 +108,7 @@ static int append_symref(const char **out, git_vector *symrefs, const char *ptr)
 		if (giterr_last()->klass != GITERR_NOMEMORY)
 			goto on_invalid;
 
+		git__free(mapping);
```
carlosmn: I think this should be part of the `on_invalid` label since that's where the code to free the contents of the refspec struct are.

src/netops.c
```diff
@@ -261,6 +261,8 @@ int gitno_extract_url_parts(
 		*path = git__substrdup(_path, u.field_data[UF_PATH].len);
 		GITERR_CHECK_ALLOC(*path);
 	} else {
+		git__free(*port);
+		*port = NULL;
```
carlosmn: We've also allocated `host` here so we should free it here too.

script/user_nodefs.h
pks-t: I'm wondering how that could even pass the Coverity check without triggering any errors. `#nodef`'s do not have any special syntax, they are essentially the same as `#idef`'s but replace existing macros.

carlosmn: The coverity script doesn't actually run on pull requests, we only run it on main branch pushes which is where we want it to upload to coverity.

pks-t: I've got an own branch that's built by Coverity, though, which I currently use for checking my changes actually fix the issues.

pks-t:
Fixed the issues pointed out by Carlos. I could not yet verify that Coverity is happy with the additional changes as it's having a hiccup right now and will re-schedule the analysis at an unspecified later date.

ethomson:
:+1: If you can ping one of us once you've verified it, that'd be awesome.

pks-t:
@ethomson @carlosmn Analysis finished successfully. Defects down by 1 to 70 outstanding defects.

PR #3617, opened 2016-02-16
commit: expose the different kinds of errors

carlosmn:
We should be checking whether the object we're looking up is a commit,
and we should let the caller know whether the not-found return code
comes from a bad object type or just a missing signature.

---

This is messy because it comes down to error class for the same return code, but `GIT_ENOTFOUND` is consistent in both cases with what we do for each kind of error.


vmg:
:+1:

PR #3623, opened 2016-02-18
rebase: additional setup tests of exotic behavior

ethomson:
Test some additional exotic rebase setup behavior:  that we are
able to set up properly when already in a detached HEAD state,
that the caller specifies all of branch, upstream and onto,
and that the caller specifies branch, upstream and onto by ID.

### DIFF
--- a/tests/rebase/abort.c
+++ b/tests/rebase/abort.c
@@ -86,19 +86,41 @@ void test_rebase_abort__merge(void)
 	git_rebase_free(rebase);
 }
 
+void test_rebase_abort__merge_by_id(void)
+{
+	git_rebase *rebase;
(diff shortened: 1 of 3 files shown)


ethomson:
@carlosmn I was trying to get a repro case for the issue you mentioned to me in chat, where you got into a strange situation using libgit2's rebase.  Unfortunately, I do not remember the details, so I banged out a couple of unit tests around what I *thought* you had said.  These all match up with git's behavior, though.

Could you remind me what you did that was odd?

carlosmn:
The problem shows up when using `git_annotated_commit_lookup()` and then calling `git_rebase_commit()` or `git_rebase_abort()`, which I don't see here.

PR #3628, opened 2016-02-22
Coverity fixes

pks-t:
Another round of Coverity fixes. Mostly fixes for memory leaks on error paths and some smallish error value fixes. Otherwise some changes such that Coverity won't complain anymore on wrong positives, especially for `git_vector_foreach`.

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -705,8 +705,7 @@ int git_path_resolve_relative(git_buf *path, size_t ceiling)
 	char *base, *to, *from, *next;
 	size_t len;
 
-	if (!path || git_buf_oom(path))
-		return -1;
+	GITERR_CHECK_ALLOC_BUF(path);
 
 	if (ceiling > path->size)
 		ceiling = path->size;
--- a/src/rebase.c
+++ b/src/rebase.c
@@ -257,12 +257,12 @@ done:
 	return error;
 }
 
-static git_rebase *rebase_alloc(const git_rebase_options *rebase_opts)
+static int rebase_alloc(git_rebase **out, const git_rebase_options *rebase_opts)
 {
 	git_rebase *rebase = git__calloc(1, sizeof(git_rebase));
+	GITERR_CHECK_ALLOC(rebase);
 
-	if (!rebase)
-		return NULL;
+	*out = NULL;
 
 	if (rebase_opts)
 		memcpy(&rebase->options, rebase_opts, sizeof(git_rebase_options));
@@ -270,14 +270,16 @@ static git_rebase *rebase_alloc(const git_rebase_options *rebase_opts)
 		git_rebase_init_options(&rebase->options, GIT_REBASE_OPTIONS_VERSION);
 
 	if (rebase_opts && rebase_opts->rewrite_notes_ref) {
-		if ((rebase->options.rewrite_notes_ref = git__strdup(rebase_opts->rewrite_notes_ref)) == NULL)
-			return NULL;
+		rebase->options.rewrite_notes_ref = git__strdup(rebase_opts->rewrite_notes_ref);
+		GITERR_CHECK_ALLOC(rebase->options.rewrite_notes_ref);
 	}
 
 	if ((rebase->options.checkout_options.checkout_strategy & (GIT_CHECKOUT_SAFE | GIT_CHECKOUT_FORCE)) == 0)
 		rebase->options.checkout_options.checkout_strategy = GIT_CHECKOUT_SAFE;
 
-	return rebase;
+	*out = rebase;
+
+	return 0;
 }
 
 static int rebase_check_versions(const git_rebase_options *given_opts)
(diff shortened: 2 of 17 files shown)


src/path.c
```diff
@@ -782,7 +782,8 @@ int git_path_resolve_relative(git_buf *path, size_t ceiling)
 
 int git_path_apply_relative(git_buf *target, const char *relpath)
 {
-	git_buf_joinpath(target, git_buf_cstr(target), relpath);
+	if (git_buf_joinpath(target, git_buf_cstr(target), relpath) < 0)
```
ethomson: This should not be necessary.  `git_path_resolve_relative` [checks for an OOM on `target`](https://github.com/libgit2/libgit2/blob/594a5d12d4b9ba3ea542722805ada1028bd1b646/src/path.c#L708) as its first check...

src/xdiff/xmerge.c
```diff
@@ -646,6 +646,8 @@ int xdl_merge(mmfile_t *orig, mmfile_t *mf1, mmfile_t *mf2,
 	if (xdl_change_compact(&xe2.xdf1, &xe2.xdf2, xpp->flags) < 0 ||
 	    xdl_change_compact(&xe2.xdf2, &xe2.xdf1, xpp->flags) < 0 ||
 	    xdl_build_script(&xe2, &xscr2) < 0) {
+		xdl_free_script(xscr1);
+		xdl_free_env(&xe1);
```
ethomson: We should upstream these changes to git itself.

src/rebase.c
```diff
@@ -270,8 +269,11 @@ static git_rebase *rebase_alloc(const git_rebase_options *rebase_opts)
 		git_rebase_init_options(&rebase->options, GIT_REBASE_OPTIONS_VERSION);
 
 	if (rebase_opts && rebase_opts->rewrite_notes_ref) {
-		if ((rebase->options.rewrite_notes_ref = git__strdup(rebase_opts->rewrite_notes_ref)) == NULL)
+		if ((rebase->options.rewrite_notes_ref =
+			git__strdup(rebase_opts->rewrite_notes_ref)) == NULL) {
+			git__free(rebase);
```
ethomson: We're not in the habit of `free`ing things on allocation failure.  I don't like to have a different pattern here.  If the goal is to silence static code analysis warnings, then we could either:

1. Change this to return an `int` error code so that `GITERR_CHECK_ALLOC` works here, or
2. Silence this another way, perhaps with a macro like `GITERR_CHECK_ALLOC` that returns `NULL` instead of `-1`.

Thoughts?

src/refspec.c
```diff
@@ -324,7 +324,7 @@ int git_refspec__dwim_one(git_vector *out, git_refspec *spec, git_vector *refs)
 		for (j = 0; formatters[j]; j++) {
 			git_buf_clear(&buf);
 			if (git_buf_printf(&buf, formatters[j], spec->src) < 0)
-				return -1;
+				goto out_err;
```
ethomson: Similar to the changes in `rebase.c`, we could have a `GITERR_CHECK_ALLOC_BUF` or something like that to check a `git_buf` for OOM status.  Or something.

src/path.c
pks-t: Ah, okay, didn't check the other function indeed.

src/rebase.c
pks-t: Yes, I definitly agree, I've been reasoning about the same two options when I fixed this. I'd prefer to convert the function to return an `int` instead, as this is how nearly all of our functions which may fail operate. As such I deem it more idiomatic for this code base.

src/xdiff/xmerge.c
pks-t: Didn't realise this was imported from git. Will post to the ML tomorrow.

src/refspec.c
pks-t: Sounds reasonable and it would help the analysis as we'd be able to #nodef the macro and treat it as an abort when we run into allocation issues.

ethomson:
Thank you for this!  These sorts of things are super tough and annoying work but are amazingly beneficial.

src/xdiff/xmerge.c
pks-t: See [gmane](http://article.gmane.org/gmane.comp.version-control.git/287034) for the proposed patch.

pks-t:
New version of the PR. Changes:

* introduced `GITERR_CHECK_ALLOC_BUF`
* removed the check in `git_buf_joinpath`
* rewrote `rebase_alloc` to return an error code
* additional fixes for Coverity warnings

I'd like to highlight commit 0f1e2d2, which changes a field in `struct git_index_reuc_entry`. I couldn't think of any other way to reliably fix this issue, so I'm open to different proposals.

pks-t:
By the way, I'd like to wait for Coverity to check this through on my branch. I've currently exceeded the rate limit, will ping you when it's finished.

pks-t:
@ethomson Looks good to me, another three issues fixed and no new defects found.

ethomson:
Looks good to me.  Thanks for all the work here.

Issue #3556, opened 2015-12-22
Access violation in `nth_packed_object_offset`


csware:
Try to read the "/Cross/" tree of commit "a9010e2d805a424e2c9279da22b56b9b1eba9b7c" (7th entry of the root tree) and when executing "git_tree_entry_to_object" on the repository you can get from https://yadi.sk/d/50mlHdydkLmZn and you'll get an access violation.

	libgit2_tgit.dll!nth_packed_object_offset(const git_pack_file * p, unsigned int n) Zeile 1190	C
 	libgit2_tgit.dll!pack_entry_find_offset(__int64 * offset_out, git_oid * found_oid, git_pack_file * p, const git_oid * short_oid, unsigned __int64 len) Zeile 1340	C
 	libgit2_tgit.dll!git_pack_entry_find(git_pack_entry * e, git_pack_file * p, const git_oid * short_oid, unsigned __int64 len) Zeile 1374	C
 	libgit2_tgit.dll!pack_entry_find_inner(git_pack_entry * e, pack_backend * backend, const git_oid * oid, git_pack_file * last_found) Zeile 247	C
 	libgit2_tgit.dll!pack_entry_find(git_pack_entry * e, pack_backend * backend, const git_oid * oid) Zeile 264	C
 	libgit2_tgit.dll!pack_backend__read(void * * buffer_p, unsigned __int64 * len_p, git_otype * type_p, git_odb_backend * backend, const git_oid * oid) Zeile 373	C
 	libgit2_tgit.dll!odb_read_1(git_odb_object * * out, git_odb * db, const git_oid * id, unsigned char only_refreshed) Zeile 846	C
 	libgit2_tgit.dll!git_odb_read(git_odb_object * * out, git_odb * db, const git_oid * id) Zeile 878	C
 	libgit2_tgit.dll!git_object_lookup_prefix(git_object * * object_out, git_repository * repo, const git_oid * id, unsigned __int64 len, git_otype type) Zeile 164	C
 	libgit2_tgit.dll!git_object_lookup(git_object * * object_out, git_repository * repo, const git_oid * id, git_otype type) Zeile 200	C
 	libgit2_tgit.dll!git_tree_entry_to_object(git_object * * object_out, git_repository * repo, const git_tree_entry * entry) Zeile 317	C

Line numbers (Zeile) according to 6aa06b65cfcce164d7ad3ef184f40acb68406078.

`git log` outputs the following error:

[...]

PR #3630, opened 2016-02-23
Extra checks for packfile indices

carlosmn:
There's a few things we can do to stop things like #3556 from happening. Here's one, also proposed for git, which checks that we're not getting sent past the end of the index for extended entries. We already check the size upon loading so the initial lookup is still safe.

We're still not checking the hashes to make sure the index corresponds to the packfile, or that the hash matches the data. We should be able to check the packfile's trailer match without too much effort. We might not want to check the hash every time as it could become a bit expensive (though it's a one-time cost (except when we miss and reload))

### DIFF
--- a/src/pack.c
+++ b/src/pack.c
@@ -365,8 +365,13 @@ static unsigned char *pack_window_open(
 	 * pointless to ask for an offset into the middle of that
 	 * hash, and the pack_window_contains function above wouldn't match
 	 * don't allow an offset too close to the end of the file.
+	 *
+	 * Don't allow a negative offset, as that means we've wrapped
+	 * around.
 	 */
 	if (offset > (p->mwf.size - 20))
+		return NULL;
+	if (offset < 0)
 		return NULL;
 
 	return git_mwindow_open(&p->mwf, w_cursor, offset, 20, left);
--- a/src/pack.c
+++ b/src/pack.c
@@ -1175,6 +1175,7 @@ int git_packfile_alloc(struct git_pack_file **pack_out, const char *path)
 static git_off_t nth_packed_object_offset(const struct git_pack_file *p, uint32_t n)
 {
 	const unsigned char *index = p->index_map.data;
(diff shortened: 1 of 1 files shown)


src/pack.c
```diff
@@ -1185,6 +1186,11 @@ static git_off_t nth_packed_object_offset(const struct git_pack_file *p, uint32_
 		if (!(off & 0x80000000))
 			return off;
 		index += p->num_objects * 4 + (off & 0x7fffffff) * 8;
+
+		/* Make sure we're not being sent out of bounds */
+		if (index + 4 >= end)
```
csware: Is `>=` correct or just `>`?

src/pack.c
```diff
@@ -1175,6 +1175,7 @@ int git_packfile_alloc(struct git_pack_file **pack_out, const char *path)
 static git_off_t nth_packed_object_offset(const struct git_pack_file *p, uint32_t n)
 {
 	const unsigned char *index = p->index_map.data;
+	const unsigned char *end = p->index_map.data + p->index_map.len;
```
csware: Use `index + p->index_map.len` here or cast the first addend to `const unsigned char *`.

src/pack.c
```diff
@@ -1175,6 +1175,7 @@ int git_packfile_alloc(struct git_pack_file **pack_out, const char *path)
...
 	index += 4 * 256;
 	if (p->index_version == 1) {
 		return ntohl(*((uint32_t *)(index + 24 * n)));
```
csware: No test here?

carlosmn: We check that the index is large enough upon load.

src/pack.c
carlosmn: Anyting past `end - 2*GIT_OID_RAWSZ` would be out of bounds, just wouldn't cause a segfault, so this is being generous. But we're reading 4 bytes so this is actually not enough, we need to check if `index + 8 > end`.

PR #3633, opened 2016-02-24
Stricter object dependency checking during creation

ethomson:
At present, we are very loose about the creation of objects that are based on other objects; for example, when calling `git_commit_create`, we do not check that the dependent objects (its tree, its commits) are actually valid objects, or that they are the requisite type.

This PR adds a new option `GIT_OPT_ENABLE_STRICT_OBJECT_CREATION`, that will turn on validation of the referenced objects:  that they exist, and that they are the required type.

This validates:

1. `git_commit_create`:  that the tree is a tree, and exists, and that the parents are commits and exist
2. `git_index_add`:  that the index entry object exists and is of the specified type, and
3. `git_treebuilder_insert`: that the specified entry exists and is of the specified type

These checks are still optional, and default to off, for performance.

/cc @vmg


ethomson:
Questions:

1. Are we cool to do this at the `git_libgit2_opts` level?  Do we ever want to only validate these things *sometimes*?
2. Ref creation *already* validates the targets.  Is there any reason that validation *shouldn't* get folded into this logic?

src/object.h
```diff
@@ -28,4 +30,36 @@ int git_oid__parse(git_oid *oid, const char **buffer_out, const char *buffer_end
...
 void git_oid__writebuf(git_buf *buf, const char *header, const git_oid *oid);
 
+GIT_INLINE(bool) git_object__is_valid(
+	git_repository *repo, const git_oid *id, git_otype type)
+{
+	git_object *obj = NULL;
+	bool valid = true;
+
+	if (git_object__strict_input_validation) {
+		if (git_object_lookup(&obj, repo, id, type) < 0)
```
vmg: Do we want to maybe `git_odb_read_header` instead, to check for the type? Seems much more cheap than a full object lookup.

ethomson: Yes, that's true, but what would be really nice is to have an `object_lookup_type` or something that would use the cached object if it was in the cache and then did a `odb_read_header` if it wasn't.  One sec.

ethomson: Okay, I did indeed add `git_object__lookup_type` which will use the cache when possible and then go to `git_odb_read_header` when the requested object is not in the cache.  Thanks, @vmg !

ethomson:
Okay, after some thought, I think that it *should* be turned on by default.  I think that most consumers are willing to sacrifice a quick lookup for safety.  Those of us that are in an environment where you're confident about the dependencies of the objects that you're writing then you can turn this setting off.  As a result, I pulled reference creation target validation into the same check.

A lot of tests assumed that they could just spam random (or cleverly named) object IDs into the index.  I updated these tests to use legitimate object IDs so that we can turn this off by default.

Any objections to this being on by default?

/cc @vmg @carlosmn

PR #3637, opened 2016-02-25
CONVENTIONS: update to include general public API principles

### LINKED ISSUE #3012: Windows and backslashes in paths
Using libgit2 0.22.1 on Windows, I noticed git_tree_entry_bypath() seems to require foward slashes as path seperator while git_repository_open_ext() accepts a path to the repository with backslashes happily.

I understand paths within a repository are internally handled with forward slashes only. But I'm wondering how to get around the issue.
Should libgit2 normalize the path it receives in git_tree_entry_bypath() or is it the at the caller to pass in relative path with forward slashes already? In case of the latter case, maybe the docs should mention this to make it more clear.

carlosmn:
Fixes #3012.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -3,6 +3,38 @@
 We like to keep the source consistent and readable.  Herein are some
 guidelines that should help with that.
 
+## External API
+
+We have a few rules to avoid surprising ways of calling functions and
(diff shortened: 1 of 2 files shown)


ethomson:
:+1:

Issue #3635, opened 2016-02-24
Error building v0.24.4-rc1on Android


swansontec:
This build failure occurs under the Android NDK for v0.24.4-rc1. This is a regression since version, 0.23.4 which compiled without errors.

```
libgit2-0.24.0-rc1/src/fileops.c: In function 'git_futils_filestamp_check':
libgit2-0.24.0-rc1/src/fileops.c:1037:40: error: 'struct stat' has no member named 'st_mtim'
  const struct timespec *statmtime = &st.st_mtim;
                                        ^
libgit2-0.24.0-rc1/src/fileops.c: In function 'git_futils_filestamp_set_from_stat':
libgit2-0.24.0-rc1/src/fileops.c:1079:40: error: 'struct stat' has no member named 'st_mtim'
  const struct timespec *statmtime = &st->st_mtim;
```

carlosmn:
We switched over to do auto-detection, so we shouldn't need to care about the particular OS. This is presumably a cross-compilation, right? Are we not doing the right thing there?

ethomson:
We're not - in this case we're assuming that a `struct stat` has a `struct timespec`.  Android's does not:

```
    unsigned long       st_atime;
    unsigned long       st_atime_nsec;
    unsigned long       st_mtime;
    unsigned long       st_mtime_nsec;
```

Which is despicable.  But whatever.  Patch forthcoming.

swansontec:
Thanks.

ethomson:
@swansontec No problem, thanks for the helpful report.

carlosmn:
Fixed via #3638

PR #3638, opened 2016-02-25
USE_NSECS fixes

ethomson:
Include some nanosecond handling fixes for Android NDK's `struct stat`, which uses neither `st_mtim` nor `st_mtimespec` to contain a `struct timespec`, and instead simply lays down an `st_mtime_nsec` field next to `st_mtime`.

Move us over to using that by default, and then setting:

    #define st_mtime_nsec mtim.tv_nsec

for compatibility with saner platforms.  This is akin to the backcompat behavior with `st_mtime` and so should seem relatively normal everywhere.

Also, fix a unit test that assumed that it would not be run in a `USE_NSECS=ON` environment:  it validated that when writing a new index entry that the nanoseconds would be truncated and thus `0`.  (It was meant to ensure that we did not propagate *old* nanosecond entries that git.git wrote.  We should, of course, validate that we write *our own* values when we're running with nanoseconds enabled.)


ethomson:
Fixes #3635

ethomson:
(Note that #3635 failed when `USE_NSECS=ON` or not, because we always assumed there was a `struct timespec` in `struct stat`, even when building without `USE_NSECS=ON`).

carlosmn:
This is beautiful in a train-crash kind of way :+1:

ethomson:
> This is beautiful in a train-crash kind of way 

That's not the first time that's been said about me.

Issue #3624, opened 2016-02-18
Crashes when concurrently fetching in two different repositories


srajko:
We have an example using NodeGit in which a fetch is performed on two separate repos at the same time (https://github.com/nodegit/nodegit/issues/907), and it crashes.  We have an experimental thread safety system that can prevent overlapping calls to libgit2 per repository (so libgit2 calls working on the same repository are not allowed to overlap, libgit2 calls working on different repositories are allowed to overlap in different threads), but in this case our thread safety system doesn't help.  Modifying it to completely prevent overlapping calls to libgit2 does help though.

Is there any part of a fetch that uses shared objects, even for separate remotes and separate repositories?  We have seen similar crashes on different systems, and they seemed to implicate transactions, transports, and credentials.

From https://github.com/libgit2/libgit2/blob/master/THREADING.md I see that the crypto library could be at fault as well.  For https://github.com/nodegit/nodegit/issues/907, I was reproducing it with openssl.  I did try calling `git_openssl_set_locking`, but it did not make a difference in this case.

carlosmn:
If you use OpenSSL (which you would if you use a released version of libssh2) you need to set up *its* thread safety. There's nothing libgit2 can do for you there.

carlosmn:
Note that `git_openssl_set_locking()` won't do anything for you on OS X because libgit2 isn't linking against OpenSSL, so it's a no-op.

ethomson:
Interesting.  There are very few pieces of state that are shared between repository instances; though indeed there are some that you might be running into.

Just to clarify one point:  you're fetching to two different on-disk repositories (eg `/foo/bar` and `/baz/quz`)?  Not two different repository instances of the same on-disk repository?

Do you have minidumps or core dumps that you can share?

srajko:
@ethomson Correct, it is two different on-disk repositories, each with a remote pointing to a distinct github repository.  We do have some crash logs, I will paste them below.

@carlosmn Thanks for the pointers - we looked further into OpenSSL and libssh2 thread safety and found some things that we could be doing better (we are now setting up the OpenSSL locking callbacks manually, and calling `libssh2_init` ourselves - we found that if you don't init it explicitly, it will auto-init when needed, but the auto-init is not thread safe).  This much seemed to address the crash reported in nodegit/nodegit#907. I will do a PR on the documentation with some additional references we found helpful.

[...]

PR #3639, opened 2016-02-25
Expand OpenSSL and libssh2 thread safety documentation

srajko:
Adds a few useful bits of information that helped in resolving #3624.

For the `libssh2_init` argument, see https://github.com/libssh2/libssh2/blob/d441da3086dcc49989b7c5b9bf3043c4a6a98211/src/global.c#L74-L78 and https://www.libssh2.org/libssh2_init.html

### DIFF
--- a/THREADING.md
+++ b/THREADING.md
@@ -72,13 +72,19 @@ which locking function it should use. This means that libgit2 cannot
 know what to set as the user of libgit2 may use OpenSSL independently
 and the locking settings must survive libgit2 shutting down.
 
+Even if libgit2 doesn't use OpenSSL directly, OpenSSL can still be used
+by libssh2 depending on the configuration.  If OpenSSL is used both by
+libgit2 and libssh2, you only need to set up threading for OpenSSL once.
+
 libgit2 does provide a last-resort convenience function
 `git_openssl_set_locking()` (available in `sys/openssl.h`) to use the
 platform-native mutex mechanisms to perform the locking, which you may
 rely on if you do not want to use OpenSSL outside of libgit2, or you
 know that libgit2 will outlive the rest of the operations. It is not
 safe to use OpenSSL multi-threaded after libgit2's shutdown function
-has been called.
+has been called.  Note `git_openssl_set_locking()` only works if
+libgit2 uses OpenSSL directly - if OpenSSL is only used as a dependency
+of libssh2 as described above, `git_openssl_set_locking()` is a no-op.
 
 If your programming language offers a package/bindings for OpenSSL,
 you should very strongly prefer to use that in order to set up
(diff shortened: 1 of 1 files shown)


THREADING.md
```diff
@@ -100,6 +104,13 @@ direct connection to libgcrypt and thus has not convenience functions for
 it (but libgcrypt has macros). Read libgcrypt's
 [threading documentation for more information](http://www.gnupg.org/documentation/manuals/gcrypt/Multi_002dThreading.html)
 
+libssh2 provides a `libssh2_init` function, which is called
+automatically by libssh2 code if needed.  However, `libssh2_init`
+is not thread-safe, and the way in which it is automatically called
+is also not thread-safe, so in a multi-threaded environment you may need
+to call libssh2_init directly to prevent the thread-unsafe automatic
+initialization.
```
carlosmn: Shouldn't we call it? Either we rely on the libgit2 consumer to init the dependencies or we don't. Half-way won't work.

THREADING.md
```diff
@@ -78,7 +78,10 @@ platform-native mutex mechanisms to perform the locking, which you may
 rely on if you do not want to use OpenSSL outside of libgit2, or you
 know that libgit2 will outlive the rest of the operations. It is not
 safe to use OpenSSL multi-threaded after libgit2's shutdown function
-has been called.
+has been called.  Note `git_openssl_set_locking()` only works if
+libgit2 uses OpenSSL directly (if GIT_OPENSSL is defined)
+- if OpenSSL is only used as a dependency of libssh2 as described below,
+`git_openssl_set_locking()` is a no-op.
```
carlosmn: We mention libgit2 not linking against OpenSSL later on. We should either put that there, or merge the paragraphs.

THREADING.md
srajko: I guess I was treating this as related to thread safety, and staying consistent with "we don't set up thread safety for dependencies" :-)

But yes, we could call it as part of `libgit2_init` and that would probably be for the best.  libssh2 keeps an internal init count so it shouldn't hurt.  I can do that in a separate PR - I pulled out this docs section.

THREADING.md
srajko: I reorganized it a bit, lemme know if that works better or not.

THREADING.md
carlosmn: This isn't about thread safety, but about initializing our dependencies. Locking is either automatic or a *application-wide* configuration so libgit2 can't assume it knows what to do.

srajko: It's about both, IMO.  Calling `libssh2_init` in a controlled fashion (like from `libgit2_init`) is a lot more thread-safe than letting libssh2 call it automatically when libgit2 uses it for the first time (because the way libssh2 detects whether it's been used for the first time is not thread-safe, and libgit2 could be calling out to it simultaneously from multiple threads - this is exactly what was happening in nodegit/nodegit#907).

carlosmn: That sounds like it's a bug in libssh2, or an implicit assumption they should advertise. The same goes for OpenSSL. We have to init the context even if we never us it because otherwise we crash.

carlosmn:
Thanks. I'm wondering if we should stop advertising `git_openssl_set_locking()` since it's not enough to guarantee thread safety a lot of the time. Debian has libssh2 link to gcrypt but we don't, on OS X it links with OpenSSL but we don't. And the next OpenSSL release is going to remove the user-provided locking (yay!) which means we'd have to start checking at compile time whether it supports them.

Unfortunately the only real way we'd have to work everywhere would be to detect at runtime whether the OpenSSL or gcrypt user-provided locking functions exist in the image and use them in that case... but even then it's the user of libgit2 and whatever else who needs to set them.

PR #3645, opened 2016-02-29
Respect negation rules where pattern length is greater than negated rule

smith-kyle:
In git core, the following .gitignore rules
```
dir
!foo/dir
```
would leave a file `foo/dir/file1` unignored. But in libgit2  `foo/dir/file1` is still ignored.

As far as I can tell this is caused by the validation done before an ignore rule is added. If the negation rule is longer the rule it is checked against the rule is considered not to match so the rule is thrown out.

Not sure if this is the right spot to tweak but `does_negate_pattern` now returns true if the rule and negation pattern's tails match one another.

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -13,32 +13,31 @@
 /**
  * A negative ignore pattern can match a positive one without
  * wildcards if its pattern equals the tail of the positive
- * pattern. Thus
+ * pattern or if the positive pattern matches the tail of the
+ * negative one. Thus
  *
  * foo/bar
  * !bar
  *
- * would result in foo/bar being unignored again.
+ * would result in foo/bar being unignored again. Likewise
+ *
+ * bar
+ * !foo/bar
+ *
+ * would also result in foo/bar being unignored again.
+ *
  */
 static int does_negate_pattern(git_attr_fnmatch *rule, git_attr_fnmatch *neg)
 {
-	char *p;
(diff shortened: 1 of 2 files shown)


carlosmn:
We check for the length because we follow the rule that you cannot unignore something which you have not put into the ignore list previously. It seems like in the past year or so _git_ itself has moved away from this, which unfortunately leaves us with mismatched rules.

So completely getting rid of `does_negate_pattern()` might be a better solution, but unless we do a comprehensive review of the ignore tests, we won't know if we are using the new rules or if what we do follows the old or the new rules, or something in between that's harder to figure out.

smith-kyle:
After looking at it more closely, it seems it's assumed negations are meant to unignore entire patterns rather than files that were ignored by previous patterns. From git docs WRT negation rules:
> any matching file excluded by a previous pattern will become included again

So instead of only allowing negation rules to flag an existing rule as negated (which I believe is happening now) negation rules should also be allowed to negate specific files that were ignored in previous rules. Does that all sound correct?

carlosmn:
It's always patterns. Even if you refer to a single file it's with a specific pattern. I'm not sure what git does if you ignore a particular file and then unignore it, but it doesn't make a lot of semantic sense.

Some of the recent ignore changes in git were rolled back, so I was waiting for 2.8 to be released before taking a look to see what rules did change.

PR #3647, opened 2016-03-02
Coverity fixes round 6

pks-t:
Another week, another round of Coverity fixes. This brings Coverity down by another ten issues. By now fixes are not that obvious anymore and require a bit more knowledge about surrounding code, so the pace is going a bit slower by now.

On top of the bug fixing commits is another fix improving interoperation between Coverity and Travis. Curl does not report any errors when receiving error status codes and thus Travis happily pretends the submission of the tarball was successful when it indeed failed. The most common case of failing (at least for me) is the rate limit hitting.

I've improved this by instructing curl to output the HTTP code and then manually parsing the whole output to not lose any information. Unfortunately curl has no easier way of accomplishing this - we want both the returned body as it contains why the build actually failed and the status code. When using `curl --fail` we get the code but output is suppressed. When using `curl --write-out %{html_code}` we have to manually parse body and status code. Whatever, at least it works now, but I'd be glad if somebody pointed out a prettier solution than this.

### DIFF
--- a/src/diff_tform.c
+++ b/src/diff_tform.c
@@ -261,7 +261,7 @@ static int normalize_find_opts(
 	if (!given ||
 		 (given->flags & GIT_DIFF_FIND_ALL) == GIT_DIFF_FIND_BY_CONFIG)
 	{
-		if (diff->repo) {
+		if (cfg) {
 			char *rule =
 				git_config__get_string_force(cfg, "diff.renames", "true");
 			int boolval;
@@ -318,8 +318,10 @@ static int normalize_find_opts(
 #undef USE_DEFAULT
 
 	if (!opts->rename_limit) {
-		opts->rename_limit = git_config__get_int_force(
-			cfg, "diff.renamelimit", DEFAULT_RENAME_LIMIT);
+		if (cfg) {
+			opts->rename_limit = git_config__get_int_force(
+				cfg, "diff.renamelimit", DEFAULT_RENAME_LIMIT);
+		}
 
 		if (opts->rename_limit <= 0)
 			opts->rename_limit = DEFAULT_RENAME_LIMIT;
--- a/src/object.c
+++ b/src/object.c
@@ -12,6 +12,7 @@
 #include "commit.h"
 #include "tree.h"
 #include "blob.h"
+#include "oid.h"
 #include "tag.h"
 
 bool git_object__strict_input_validation = true;
@@ -166,13 +167,9 @@ int git_object_lookup_prefix(
 			error = git_odb_read(&odb_obj, odb, id);
 		}
 	} else {
-		git_oid short_oid;
+		git_oid short_oid = {{ 0 }};
 
-		/* We copy the first len*4 bits from id and fill the remaining with 0s */
-		memcpy(short_oid.id, id->id, (len + 1) / 2);
-		if (len % 2)
-			short_oid.id[len / 2] &= 0xF0;
-		memset(short_oid.id + (len + 1) / 2, 0, (GIT_OID_HEXSZ - len) / 2);
+		git_oid__cpy_prefix(&short_oid, id, len);
 
 		/* If len < GIT_OID_HEXSZ (a strict short oid was given), we have
 		 * 2 options :
--- a/script/coverity.sh
+++ b/script/coverity.sh
@@ -49,10 +49,24 @@ COVERITY_UNSUPPORTED=1 \
 # Upload results
 tar czf libgit2.tgz cov-int
 SHA=$(git rev-parse --short HEAD)
-curl \
+
+HTML="$(curl \
+	--silent \
+	--write-out "\n%{http_code}" \
 	--form token="$COVERITY_TOKEN" \
 	--form email=bs@github.com \
 	--form file=@libgit2.tgz \
 	--form version="$SHA" \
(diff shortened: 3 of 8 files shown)


pks-t:
Verification through Coverity is still pending but I want to put it up for review. Will notify you on completion.

pks-t:
First round has not been as successful. While 10 issues were fixed, 9 new ones popped up. I'll have to wait until tomorrow 11:00 UTC as the quota is reached for today.

carlosmn:
I'm not sure about the change tot he config's `_force()` functions. Even though it does say "any errors", having a `NULL` config seems like an error that happens before we would call this, so we should instead `assert()` that the config is there. If there were errors retrieving the config object, that should be handled differently by the caller. /cc @ethomson for thoughts

ethomson:
> we should instead assert() that the config is there.

Agreed.

pks-t:
Okay. There are paths where we do not check if we actually have a valid config and where we use the `_force` functions. I'll fix those then.

src/object.c
```diff
@@ -164,13 +164,12 @@ int git_object_lookup_prefix(
 			error = git_odb_read(&odb_obj, odb, id);
 		}
 	} else {
-		git_oid short_oid;
+		git_oid short_oid = { 0 };
```
ethomson: Can you make this initializer `{{0}}` to silence a clang warning about it?

pks-t:
Okay, fixed up the initialization warning pointed out by @ethomson and removed returning the default values when `git_config_*_force` receives a `NULL` configuration. I've fixed up the caller in 1f6361e instead.

pks-t:
Commits have been verified by Coverity by the way, so I consider this good to go.

vmg:
All these changes look good to me. Good job @pks-t!

carlosmn:
The changes to the coverity script should be backed out, no?

tests/object/lookup.c
```diff
@@ -48,6 +48,19 @@ void test_object_lookup__lookup_wrong_type_by_abbreviated_id_returns_enotfound(v
...
+{
+	const char *commit = "e90810b8df3e80c413d903f631643c716887138d";
+	git_oid oid;
+	git_object *object;
+
+	cl_git_pass(git_oid_fromstrn(&oid, commit, strlen(commit)));
+	cl_git_pass(git_object_lookup_prefix(&object, g_repo, &oid, GIT_OID_HEXSZ - 1, GIT_OBJ_COMMIT));
+	git_object_free(object);
+
+	cl_assert_equal_i(0, oid.id[GIT_OID_HEXSZ - 1]);
```
carlosmn: The id is purely an input parameter. Nothing we do inside `git_object_lookup_prefix()` would modify what this value contains. Even if we did, the size is in nibbles so this byte wouldn't be fully zeroed out, just the lower half. It is rather surprising that this succeeds in most builds.

src/object.c
```diff
@@ -166,13 +166,12 @@ int git_object_lookup_prefix(
 			error = git_odb_read(&odb_obj, odb, id);
 		}
 	} else {
-		git_oid short_oid;
+		git_oid short_oid = {{ 0 }};
 
 		/* We copy the first len*4 bits from id and fill the remaining with 0s */
```
carlosmn: This comment isn't quite accurate anymore. Maybe "We only copy over the first len*4 bits from the id"

tests/object/lookup.c
pks-t: Oh, you're right. Hm, will remove the test.

src/object.c
pks-t: I could just include "oid.h" and then use the new `git_oid__cpy_prefix` function introduced by @ethomson

pks-t:
Oh, obviously I've failed to remove my specific Coverity setup params. Sorry for that one, will fix up

src/object.c
carlosmn: Yep, using a named function would be great here.

pks-t: @ethomson: Just took a look at `git_oid__cpy_prefix`. Shouldn't it first memset `out` to all-zero? Currently we just copy the prefix and only set `out->id[len/2] &= 0xF0` without touching the trailing bits. I see other invocations just initialize the oid with `= {{0}}`, but I'd at least like to document the requirement.

src/diff_tform.c
```diff
@@ -318,8 +318,10 @@ static int normalize_find_opts(
 #undef USE_DEFAULT
 
 	if (!opts->rename_limit) {
-		opts->rename_limit = git_config__get_int_force(
-			cfg, "diff.renamelimit", DEFAULT_RENAME_LIMIT);
+		if (diff->repo) {
```
carlosmn: Shouldn't we be hecking for `cfg` here since that's where we're reading from?

pks-t: Well, see a few lines above. If `diff->repo != NULL` we return -1 if `git_repository_config__weakptr(&cfg, diff->repo) < 0`. I mainly used `if (diff->repo)` because we use the same check when retrieving the "diff.renames" config.

carlosmn: I'd also change those above, it's too roundabout when we care about the config existing.

pks-t: Okay, fixed

PR #3649, opened 2016-03-03
index writing without mmap

ehmry:
Index writing does not work with the mmap emulation in posix.c.

See these two lines:
https://github.com/libgit2/libgit2/blob/master/src/indexer.c#L468
https://github.com/libgit2/libgit2/blob/master/src/posix.c#L235

### DIFF
--- a/src/posix.c
+++ b/src/posix.c
@@ -239,9 +239,13 @@ int p_mmap(git_map *out, size_t len, int prot, int flags, int fd, git_off_t offs
 	out->data = NULL;
 	out->len = 0;
 
-	if ((prot & GIT_PROT_WRITE) && ((flags & GIT_MAP_TYPE) == GIT_MAP_SHARED)) {
-		giterr_set(GITERR_OS, "Trying to map shared-writeable");
-		return -1;
+	if (prot & GIT_PROT_WRITE) {
+		if ((p_lseek(fd, offset, SEEK_SET) < 0) || p_write(fd, out->data, len)) {
+			giterr_set(GITERR_OS, "index write failed");
+			return -1;
+		}
+		out->len = 0; /* write not shared */
+		return 0;
 	}
 
 	out->data = malloc(len);


carlosmn:
It does indeed not work, but I don't see why the indexer should know about it. It's a missing feature from the emulation.

ethomson:
Indeed - the point of the emulation layer is that we *don't* need terrible `ifdef`s all over the code base.

It would be nice if you would make `p_mmap` and `p_unmap` do the write.

ehmry:
I can do that, but give me a day or two.

ehmry:
I don't want to be that guy that shows up and litters macros.
Fixed in bc609c9400be00522462ace47b4d20f9a528cfae

ehmry:
Actually, I complied and tested but forgot I was on the wrong OS, I still need to test this.

src/posix.c
```diff
@@ -239,9 +239,13 @@ int p_mmap(git_map *out, size_t len, int prot, int flags, int fd, git_off_t offs
 	out->data = NULL;
 	out->len = 0;
 
-	if ((prot & GIT_PROT_WRITE) && ((flags & GIT_MAP_TYPE) == GIT_MAP_SHARED)) {
-		giterr_set(GITERR_OS, "Trying to map shared-writeable");
-		return -1;
+	if (prot & GIT_PROT_WRITE) {
+		if ((p_lseek(fd, offset, SEEK_SET) < 0) || p_write(fd, out->data, len)) {
+			giterr_set(GITERR_OS, "index write failed");
```
carlosmn: Why does this mention an index? The problem is trying to user mmap emulation for writable pages, which is what we don't support. Whether it's the indexer or anything else is irrelevant and having this error message doesn't explain _why_ it failed.

src/posix.c
```diff
@@ -239,9 +239,13 @@ int p_mmap(git_map *out, size_t len, int prot, int flags, int fd, git_off_t offs
 	out->data = NULL;
 	out->len = 0;
 
-	if ((prot & GIT_PROT_WRITE) && ((flags & GIT_MAP_TYPE) == GIT_MAP_SHARED)) {
-		giterr_set(GITERR_OS, "Trying to map shared-writeable");
-		return -1;
+	if (prot & GIT_PROT_WRITE) {
+		if ((p_lseek(fd, offset, SEEK_SET) < 0) || p_write(fd, out->data, len)) {
```
carlosmn: What is the `p_write()` call supposed to do? `out->data` is `NULL` and `len` is `0` since we just set them. What would this do besides be a no-op?

ehmry:
I'm too lazy to worry about this anymore.

PR #3651, opened 2016-03-03
ssh: initialize libssh2

carlosmn:
We should have been doing this, but it initializes itself upon first
use, which works as long as nobody's doing concurrent network
operations. Initialize it on our init to make sure it's not getting
initialized concurrently.

### DIFF
--- a/src/global.c
+++ b/src/global.c
@@ -12,6 +12,7 @@
 #include "openssl_stream.h"
 #include "thread-utils.h"
 #include "git2/global.h"
+#include "transports/ssh.h"
 
 #if defined(GIT_MSVC_CRTDBG)
 #include "win32/w32_stack.h"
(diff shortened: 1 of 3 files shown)


srajko:
Nice, I started putting this together too based on your suggestion (mine needs to be fixed up though).

Might be good to do a matching call to `libssh2_exit` too - https://github.com/srajko/libgit2/blob/b7187061277e15daf92cff5451f47bcfc019c6e0/src/transports/ssh.c#L840-L850

carlosmn:
We don't free OpenSSL either. It's a bit of an edge-case if libgit2's lifetime doesn't match the applications, so we don't expect many will actually bother calling libgit2's shutdown function either, so I'm OK with not freeing this right now.

ethomson:
> so I'm OK with not freeing this right now.

Me, too.  I would like to beat on init (and perhaps shutdown) a little bit more, but that's something to do after this release.  :)

PR #3652, opened 2016-03-03
commit: split creating the commit and writing it out

carlosmn:
Sometimes you want to create a commit but not write it out to the
objectdb immediately. For these cases, provide a new function to
retrieve the buffer instead of having to go through the db.

---

This is new, so we should wait to merge until after the release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ v0.24 + 1
 
 ### API additions
 
+* `git_commit_create_buffer()` creates a commit and writes it into a
+  user-provided buffer instead of writing it into the object db.
+
 ### API removals
 
 ### Breaking API changes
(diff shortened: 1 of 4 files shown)


ethomson:
Hm.  So, I must admit that it seems odd to me that `git_commit__create_buffer_internal` does more than just stuffing the commit information into a buffer, and that it does the `current_id` validation for `git_commit__create_internal `.

I would personally rather see the `git_commit__create_internal` function do that test and then pass an array of commit ids to the buffer creation function, which seems like it keeps the separation of concerns a bit more separate.

carlosmn:
I wasn't quite sure what to do about that. I went with keeping the solution we have which does not perform extra allocations. But come to think about it, when we split this off, the parent check doesn't mean much, as that's a check that should happen at the writing side.

carlosmn:
I've updated the function to accept just the data, and we perform the validation at a higher level.

PR #3653, opened 2016-03-04
treebuilder: don't try to verify submodules exist in the odb

carlosmn:
Submodules don't exist in the objectdb and the code is making us try to
look for a blob with its commit id, which is obviously not going to
work.

Skip the test if the user wants to insert a submodule.


ethomson:
Obviously it is not.  Thanks for the fix.

PR #3654, opened 2016-03-04
[WIP] config: show we write a spurious duplicated section header

carlosmn:
We should notice that we are in the correct section to add. This is a
cosmetic bug, since replacing any of these settings does work.


ethomson:
Included in #3712

PR #3655, opened 2016-03-04
Enable nanosecond resolution by default

ethomson:
Nanosecond resolution is now the default in git itself.  Enable this
as our default as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,7 +109,7 @@ ELSE ()
 ENDIF()
 
 IF (HAVE_STRUCT_STAT_NSEC OR WIN32)
-	OPTION( USE_NSEC		"Care about sub-second file mtimes and ctimes"	OFF )
+	OPTION( USE_NSEC		"Care about sub-second file mtimes and ctimes"	ON  )
 ENDIF()
 
 # This variable will contain the libraries we need to put into
--- a/tests/index/nsec.c
+++ b/tests/index/nsec.c
@@ -24,6 +24,43 @@ void test_index_nsec__cleanup(void)
 	cl_git_sandbox_cleanup();
 }
 
+static bool try_create_file_with_nsec_timestamp(const char *path)
+{
+	struct stat st;
+	int try;
+
+	/* retry a few times to avoid nanos *actually* equal 0 race condition */
+	for (try = 0; try < 3; try++) {
+		cl_git_mkfile(path, "This is hopefully a file with nanoseconds!");
+
+		cl_must_pass(p_stat(path, &st));
+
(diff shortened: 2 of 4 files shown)


carlosmn:
The ones building "Release" fail... should we be worried this is a timing bug in the test suite or the library?

ethomson:
Indeed it was a timing bug in the test suite.  We were trusting that we could do a bunch of operations really quickly to get ourselves a racy entry in the index - which is complete horseshit buuuuut it also happens to be true on most filesystems because they have shit resolution.

Instead of doing stuff real fast and hoping that it works out, I twiddled the index's timestamp, so we now invent some raciness instead.

hackhaslam:
Changing the default regressed performance of things like `status` on Windows. I haven't delved too deeply, but it looks like maybe it breaks the attribute cache. Do you want me to add a new issue for this?

ethomson:
@hackhaslam Hmm.  We haven't seen anything like this, and everything certainly seems normal to me.  (Faster, in fact, since there's better resolution on some platforms.)

Please do open a new issue if you think that there's  a problem, but please provide plenty of details.  A broken index would be helpful, as would information about the client that created it and the client that is deciding that everything's dirty.

Is the slow client in question actually an _old_ version of libgit2?  Are you running into  https://github.com/libgit2/libgit2/pull/3480 ?

hackhaslam:
No, it's not an old version. I'm using the tip of master. I will look more closely to see if I can provide some more details.

hackhaslam:
Okay, maybe I misunderstood your question. It turns out that the index (or at least the index of some rather large submodules) was written with an old version of git (1.8.4). After rebuilding the indexes of all of the submodules with a new version of git the problem goes away. So presumably, older versions of git wrote an index without nanosecond resolution and now I'm trying to compare it with stats of nanosecond resolution? Is there any way that I can detect this at runtime and only compare nanoseconds if the index has nanosecond resolution?

ethomson:
Aha.  You didn't misunderstand, my first guess was incorrect.  But that makes sense.  We're quite obviously behaving incorrectly here, but at least it makes sense.  :)

This is the opposite problem that we were stuck with when git itself turned on nanoseconds and we were unprepared.  Now we're expecting nanoseconds always, it appears.  What we should be doing here is detecting that the index doesn't have nanoseconds and falling back to non-nanosecond behavior.  Looks like we're screwing that one up.

I'm afraid that I can't fix this today - but feel free to open an issue and I should be able to get to it later this week.  

Thanks for the report!

hackhaslam:
Thanks! I added #3754.

PR #3656, opened 2016-03-07
Introduce `git_odb_expand_ids`

ethomson:
Introduce an API to take many OID prefixes and determine which (if any) are uniquely resolvable.  The full OID is returned for those that exist.

Unlike most of the ODB APIs, this *does not* reload the ODB to try to refresh the backends if a lookup fails.  That would, obviously, be a totally insane number of ODB refreshes.

### DIFF
--- a/include/git2/odb.h
+++ b/include/git2/odb.h
@@ -175,11 +175,12 @@ GIT_EXTERN(int) git_odb_exists_prefix(
 
 /**
  * Determine if one or more objects can be found in the object database
- * by their abbreviated object IDs.  Callers may further restrict the
- * lookup based on type.  This function will write the complete object
- * ID to the `id`s array, and the updated length to the `id_lengths`
- * array.  (If an object is found, it will have its length updated to
- * `GIT_OID_HEXSZ`; if an object is not found, will be be `0`.)
+ * by their abbreviated object ID and type.  The given array will be
+ * updated in place:  for each abbreviated ID that is unique in the
+ * database, and of the given type (if specified), the full object ID,
+ * object ID length (`GIT_OID_HEXSZ`) and type will be written back to
+ * the array.  For IDs that are not found (or are ambiguous), the
+ * array entry will be zeroed.
  *
  * Note that since this function operates on multiple objects, the
  * underlying database will not be asked to be reloaded if an object is
@@ -194,7 +195,7 @@ GIT_EXTERN(int) git_odb_exists_prefix(
  * @param cnt The length of the `ids`, `id_lengths` and `types` arrays
  * @return 0 on success or an error code on failure
  */
-GIT_EXTERN(int) git_odb_exists_many_prefixes(
+GIT_EXTERN(int) git_odb_expand_ids(
 	git_odb *db,
 	git_oid *ids,
 	size_t *id_lengths,
--- a/include/git2/odb.h
+++ b/include/git2/odb.h
@@ -174,6 +174,27 @@ GIT_EXTERN(int) git_odb_exists_prefix(
 	git_oid *out, git_odb *db, const git_oid *short_id, size_t len);
 
 /**
+ * The information about object IDs to query in `git_odb_expand_ids`,
+ * which will be populated upon return.
+ */
+typedef struct git_odb_expand_id {
+	/** The object ID to expand */
+	git_oid id;
+
+	/**
+	 * The length of the object ID (in nibbles, or packets of 4 bits; the
+	 * number of hex characters)
+	 * */
+	unsigned short length;
+
+	/**
+	 * The (optional) type of the object to search for; leave as `0` or set
+	 * to `GIT_OBJ_ANY` to query for any object matching the ID.
+	 */
+	git_otype type;
+} git_odb_expand_id;
+
+/**
  * Determine if one or more objects can be found in the object database
  * by their abbreviated object ID and type.  The given array will be
  * updated in place:  for each abbreviated ID that is unique in the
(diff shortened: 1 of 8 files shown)


ethomson:
So, the tricky part here is not the actual doing of the lookups.  It's putting a reasonable API on top of it.  I fear that I have not done that.

One thing that this does that I feel weird about is that it takes an array that is set to the correct size by the caller, which this API populates.  That's reasonable(ish) because the caller knows how big it is...  and that prevents us from having to do a malloc ourselves and provide a free function.  But this is also the only place, IIRC, that we do this, so it seems a bit odd.

ethomson:
Well, one way to avoid being worried about the caller having to have some special knowledge is to just write all over the memory they handed us in the first place.

ethomson:
So this was @vmg 's suggestion, which was to modify the array in place.  I'm still not suuuuuper in love with the signature, and I still wonder if we shouldn't return a separate list of `id`s that were found.

I'm curious what @carlosmn thinks about this, we've avoided "in/out" type arguments in the past in favor of explicit in vs out arguments.

carlosmn:
We have avoided them, but this is a function we're writing for performance, so I think overwriting memory is fine.

However, if we're going to have this behaviour, I don't like the name. It reads too much like something that should return a list of booleans. Something like `git_odb_expand_existing()` might be more appropriate.

src/pack.c
```diff
@@ -1018,7 +1018,7 @@ static int packfile_open(struct git_pack_file *p)
 	unsigned char *idx_sha1;
 
 	if (p->index_version == -1 && pack_index_open(p) < 0)
-		return git_odb__error_notfound("failed to open packfile", NULL);
+		return git_odb__error_notfound("failed to open packfile", NULL, 0);
```
carlosmn: Won't this kind of thing make us show empty parens after the error message (I suppose this might happen already, in which case meh)

carlosmn:
The only other comment I have is about caches: by putting each data type in a different array, we're going to be jumping back and forth between memory areas, so we'll be throwing away a bunch of pre-fetched memory. If we used a structure with `(id, len, type)`, we would be able to move strictly forward in the user-provided memory.

Of course we do so much I/O in-between that this might not actually matter.

vmg:
> However, if we're going to have this behaviour, I don't like the name. It reads too much like something that should return a list of booleans. Something like git_odb_expand_existing() might be more appropriate.

I agree. `expand_existing` or `expand_oids` sounds better.

> The only other comment I have is about caches: by putting each data type in a different array, we're going to be jumping back and forth between memory areas

That will be profoundly inconsequential in practice. We will never fetch more than 100 objects at the same time.

> If we used a structure with (id, len, type), we would be able to move strictly forward in the user-provided memory.

The only reason to use this structure would be to make the API cleaner, but I'm not sure it's worth it to add another data type to the library.

carlosmn:
>That will be profoundly inconsequential in practice. We will never fetch more than 100 objects at the same time.

Fair

>The only reason to use this structure would be to make the API cleaner, but I'm not sure it's worth it to add another data type to the library.

Nah, it's more annoying to set up. Let's leave it as-is.

vmg:
@ethomson: To make the API more "consistent", it'd be real neat if the `types` array could also be filled up as we expand the SHAs (in the case any of the types is passed as `OBJ_ANY`). We already have the information, so it should be easy to do. :)

src/pack.c
ethomson: `git_odb__error_notfound` will not include the parens when the given `id` is `NULL`.  :)

ethomson:
Okay, we now populate types on the way back out.  I also moved this to take an array of structs instead of three different arrays.  I actually think that this is easier to set up and tear down, but if there is violent disagreement then we can drop the last commit.

vmg:
Love it. @carlosmn any last comments?

include/git2/odb.h
```diff
@@ -171,6 +173,40 @@ GIT_EXTERN(int) git_odb_exists(git_odb *db, const git_oid *id);
...
 
+typedef struct git_odb_expand_id {
+	/** The object ID to expand */
+	git_oid id;
+
+	/** The length of the object ID */
+	unsigned short length;
+
+	/** The type of the object (optionally) */
+	git_otype type;
```
carlosmn: We mark this as optional, which would generally be understood as "leave zeroed and we don't do anything" but leaving this field zeroed means `GIT_OBJ__EXT1`, and won't pass the check against `GIT_OBJECT_ANY` which is what we actually mean by optional here.

It should say to set to `_ANY` if you don't want to enforce the type.

include/git2/odb.h
```diff
@@ -171,6 +173,40 @@ GIT_EXTERN(int) git_odb_exists(git_odb *db, const git_oid *id);
 GIT_EXTERN(int) git_odb_exists_prefix(
 	git_oid *out, git_odb *db, const git_oid *short_id, size_t len);
 
+typedef struct git_odb_expand_id {
```
carlosmn: You'll wanna put some comment here so it does show up in the docs.

include/git2/odb.h
ethomson: This function accepts `0` or `_ANY` for the sanity of callers.

include/git2/odb.h
```diff
@@ -171,6 +173,40 @@ GIT_EXTERN(int) git_odb_exists(git_odb *db, const git_oid *id);
...
+	/** The object ID to expand */
+	git_oid id;
+
+	/** The length of the object ID */
+	unsigned short length;
```
carlosmn: This should specify the units, since it'll be nibbles rather than the more common unit of byte.

include/git2/odb.h
vmg: Nice design. :ok_hand:

carlosmn: Ah yes, I see the code. But I don't think we have tests for the zero, just for `_ANY`.

include/git2/odb.h
vmg: Hah, like that time thousands the OID lookups in production became ambiguous because we were passing bytes instead of nibbles. :))))))

...good times.

carlosmn: Right, that.

include/git2/odb.h
ethomson: Tests added.

include/git2/odb.h
ethomson: Fixed.

include/git2/odb.h
ethomson: Fixed.  :)

carlosmn:
:+1:

PR #3658, opened 2016-03-08
Update link to Pro Git's Git internals chapter

patrickmckenna:
The existing, broken link points to v1 of the book. The new link points to the same chapter in the current version (v2) of the book.

(After finding this, I searched for other references to "Pro Git" or `progit.org`, but only found this one, so I don't think there are other URLs to update.)


ethomson:
Thanks!

PR #3660, opened 2016-03-08
MinGW builds should optionally create DLLs without "lib" prefix

mstrap:
This is required to fix jagged MinGW builds


carlosmn:
From my understanding, MinGW is supposed to look like unix internally. Wouldn't this be jagged's bug if it can't find this variation of the library name?

ethomson:
It is supposed to look like unix internally, but its purpose is to build actual win32 DLLs.  So I would have expected it to emit `foo.dll` (just like MSVC does) not `libfoo.dll`.

That said, it quite surprises me that nobody has ever complained about this.  Certainly there are people using mingw, as much as I wish they wouldn't.  mingw has hacked gcc such that `-lfoo` will look for `libfoo.dll` _and_ `foo.dll`.  So if people are using mingw up and down their toolchain, and never interoperating with anything else Windows then I guess it might work.

cc @tkelman who knows more about mingw than I do.

tkelman:
If you look at large sets of mingw dll packages in msys2 or cross compiled collections, they mostly use the lib prefix (libtool and other pieces of legacy cruft handle linking to dll's better when they have a lib prefix) but not universally. I'm inclined to stick with cmake's defaults.

ethomson:
Yeah, but almost all of the mingw libraries that mingw includes only exist for the sake of mingw, right?  If you're using mingw to build a DLL that you actually want to share with the rest of the world, will other tools understand the `lib` prefix?

Let me provide a concrete example:  if we wanted to build libgit2 with mingw and then use it in LibGit2Sharp...  Does P/Invoke understand that it should look for `git2.dll` *and* `libgit2.dll`?  I would wager that it does not...

I'm not completely opposed to making jagged's JNI loader try to load `git2` and `libgit2`.  (Presumably on Windows only, to avoid looking for `liblibgit2`.  Sigh.)

But if mingw *is* capable of handling both, then I think that having it emit `git2.dll` makes sense for maximum compatibility...

tkelman:
I don't know what you mean by "for the sake of mingw." Can P/Invoke refer to a dll by full path?

I don't personally need to link against libgit2, I only ever load it dynamically, but I like being able to use the same library name on all platforms - the Julia C FFI handles the extension automatically but doesn't mess with the rest of the library name.

For mingw applications that do need to link against libgit2, doing so without the lib prefix may be harder to make work with some build systems. Especially if you aren't creating an import library as `libgit2.dll.a` (the mingw name for the import library version of `git2.lib`, the static library is `libgit2.a`). I don't know if that's being done here yet since I don't personally need the import lib. I've definitely seen this with libtool and dll's where you don't have import libs, if you can skip making an import lib it saves some trouble.

@mingwandroid or @yselkowitz might have more specific reasons why the lib prefix is generally preferred.

mstrap:
I'm running the CMake build from within CLion IDE which requires MinGW instead of MSVC. (When running from command line, CMake is using MSVC.) Now I'm not exactly sure what is CLion-part and what is plain CMake-part, but when not having `-D_DEBUG` present, debug information is properly included and `flags.make` contains:

    C_FLAGS = -D_GNU_SOURCE -Wall -Wextra  -Wno-missing-field-initializers -Wstrict-aliasing=2 -Wstrict-prototypes -Wdeclaration-after-statement -Wno-unused-function -g  

When having `-D_DEBUG` present, debug information is not included (DLL is just about half of the size) and `flags.make` contains:

    C_FLAGS = -D_GNU_SOURCE -Wall -Wextra  -Wno-missing-field-initializers -Wstrict-aliasing=2 -Wstrict-prototypes -Wdeclaration-after-statement -Wno-unused-function -D_DEBUG  

So the `-g` seems to be missing.

mingwandroid:
The lib prefix and dll.a suffix for import libraries and .a suffix for static libraries are the defacto standards across all mingw-* based toolchains and distributions.

Please don't swim upstream by doing otherwise.

I'm not going to reply to the flamebait about toolchains except to say that open source matters.

ethomson:
Indeed open source does matter, and we believe that mingw matters.  We support it as a first-class platform and will continue to do so.  That isn't up for debate.

The goal of this PR is to use mingw to produce libraries that can be used within the larger Windows ecosystem.  And, of course, the larger Windows ecosystem doesn't prefix libraries with a `lib` name.  So a project like LibGit2Sharp, which needs to dynamically load libgit2, tries to (basically) `LoadLibrary("git2.dll")`...  Obviously, all these programs that depend on libgit2 can't be expected to know whether you used MSVC (or whatever) or mingw to build libgit2.

Our build system needs to be able to emit `git2.dll` when being built by mingw.  Swimming upstream would be to refuse to support this scenario and not interoperate with the rest of the Windows world.  The question is how to implement this.

@mstrap Can you change this so that it's optional - say have an option like `LIBGIT2_PREFIX` that - if set - updates the cmake `PREFIX`?  Does cmake treat unset and empty as distinct?  So that not setting this option can keep the per-platform defaults, but a user could override it by setting it to an empty string?

Also, [the docs](https://cmake.org/cmake/help/v3.0/command/set_target_properties.html) suggest that we might be able to use something more general than `CMAKE_SHARED_LIBRARY_PREFIX`, which would affect static library filename creation as well?

yselkowitz:
Code which is dynamically loading libraries needs to take existing naming conventions into account, not attempt to force them to change.  Also, keep in mind that MSVC and MinGW toolchains aren't always 100% ABI compatible, although for pure C libraries this *might* work.

tkelman:
> And, of course, the larger Windows ecosystem doesn't prefix libraries with a lib name.

Depends what you mean by "the larger Windows ecosystem." If the majority of libraries you use are open source and you use mingw to build them, then you will have the lib prefix everywhere. MSVC as a compiler fails to meet a lot of communities' needs, even ignoring that it's not open source. You get a shockingly long way without having to ever use MSVC these days.

> So a project like LibGit2Sharp, which needs to dynamically load libgit2, tries to (basically) LoadLibrary("git2.dll")... Obviously, all these programs that depend on libgit2 can't be expected to know whether you used MSVC (or whatever) or mingw to build libgit2.

Why not? PyGit2 does exactly this. It's usually safer to use MSVC built libraries when building your own application with MSVC, and mingw built libraries when building your application with mingw. So if you know which compiler you're using, you can use that compiler's naming convention. Does LibGit2Sharp work under mono? If so does it have code that looks for libgit2.so on linux and libgit2.dylib on mac? git2.dll on msvc windows and libgit2.dll on mingw windows would be the established convention, and at least in my use case if I had to deal with a git2.dll (instead of choosing to ignore MSVC right now, which serves me fine) I would do so with an `mv` in my build system to always add a lib prefix. MSVC's convention is the one that pointlessly stands out as nonconforming, from my standpoint.

I can name half a dozen different binary distributions of compiled mingw built libgit2.dll libraries today, users of which likely won't want to change here. Are there any such binary distributions of compiled msvc built git2.dll libraries (just the library/headers, not talking about larger applications it may be bundled with)? I'm not aware of any.

@mstrap the issue with DEBUG in the CFLAGS sounds like it's probably a CLion issue.

tkelman:
Making something optional to adjust to prefix at configure time seems fine though if cmake can be made to cooperate. Or LibGit2Sharp could say "make a renamed copy of the file," if that's what it takes. You won't always have control over how the binaries you're trying to use were configured.

ethomson:
> I can name half a dozen different binary distributions of compiled mingw built libgit2.dll libraries today, users of which likely won't want to change here
> I would do so with an `mv` in my build system to always add a lib prefix.

I don't know why we're arguing about this.

ethomson:
> Making something optional to adjust the prefix at configure time seems fine though if cmake can be made to cooperate.

Then it sounds like we're all on the same page, no?

tkelman:
Sure, but you'd be adding work and more configuration options here to avoid fixing the bug in LibGit2Sharp and jagged that they don't handle an established convention. You don't have support for library name aliases, so you could also deal with things like sonames?

carlosmn:
There is no libgit2sharp bug to work around.

ethomson:
> Sure, but you'd be adding work and more configuration options here to avoid fixing the bug in LibGit2Sharp...

This isn't something specific to LibGit2Sharp.  LibGit2Sharp uses P/Invoke, which expects you to give it a single path to the library to load.  And *of course* that path is `git2.dll` on Windows, because that's the normal Windows thing.

This isn't even - as far as I remember - an MSVC thing.  Borland and Watcom behaved the same way, no?  mingw is the odd man out here.  And that's fine.

But that doesn't mean that there's a bug in LibGit2Sharp to fix.  Even if [.NET](https://github.com/Microsoft/dotnet) changed this behavior tomorrow, though, we still need to support old framework revisions, so this would *still* be important for us.

tkelman:
Odd man out w.r.t. other windows compilers, but more like every other platform. Anyway, looked into it and https://github.com/libgit2/libgit2sharp.nativebinaries#specifying-custom-dll-names seems to completely address this for LibGit2Sharp. Other language bindings could implement similar functionality since it seems like a more useful thing beyond just what compiler you prefer.

ethomson:
The instructions that you're pointing to are for building the LibGit2Sharp.NativeBinaries nuget package.  Those are, in fact, able to hardcode the library name that LibGit2Sharp itself should look for.  And this is the primary mechanism that most regular users will end up using to use libgit2 within LibGit2Sharp.

But some people will be compiling their own libgit2 and would not be using this nuget packaged system.

mstrap:
I've now introduced `LIBGIT2_PREFIX` option as suggested by @ethomson.

mstrap:
@tkelman OK, so it's probably a CLion issue, however I didn't succeed to get a valid debug build without modifying CMakeLists.txt. I'm wondering why `AND NOT CYGWIN` is already present there? Could the MSVC-Option `-D_DEBUG` have caused other kind of problems with Cygwin?

Another way to get CLion running would be to explicitly specify `-g` for `${CMAKE_C_FLAGS_DEBUG}`. Unfortunately these flags are currently reset (but again only in case of WIN32), so another solution to my problem would be to only *append* to ${CMAKE_C_FLAGS_DEBUG}:

	IF (WIN32 AND NOT CYGWIN)
		SET(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -D_DEBUG")
	ENDIF ()

Does that sounds reasonable?

CMakeLists.txt
```diff
@@ -608,6 +608,8 @@ IF (SONAME)
 	IF (LIBGIT2_FILENAME)
```
tkelman: could you have already just used this, setting `-DLIBGIT2_FILENAME=git2.dll` ?

appending to debug flags sounds like a good idea anyway. maybe check blame for why that line says cygwin.

mstrap: When using `-DLIBGIT2_FILENAME=git2.dll`, the resulting file will be `libgit2.dll.dll`.

mstrap:
@tkelman The entire block: 

	IF (WIN32 AND NOT CYGWIN)
		SET(CMAKE_C_FLAGS_DEBUG "-D_DEBUG")
	ENDIF ()

has been introduced three years ago as part of one large change "Improve MSVC compiler, linker flags". While I still don't think that the author has intentionally left out `AND NOT MINGW`, I've noticed that `_DEBUG` is actually used in MinGW's header file `dxtmpl.h` (whatever this is good for). So definitely better to not touch this and hence I've create new pull request #3687.

CMakeLists.txt
tkelman: yeah, nevermind, [sets the base name](https://cmake.org/cmake/help/v3.0/prop_tgt/OUTPUT_NAME.html) - fair enough

mstrap:
Author and committer changed to my personal account.

ethomson:
Thanks!

PR #3661, opened 2016-03-08
Ignore .idea (used by CLion IDE)

mstrap:


carlosmn:
Editor-specific rules should be in a per-user global ignore file. Otherwise we'd end up adding rules for every single editor out there.

PR #3669, opened 2016-03-09
Initialize `error` variable

arthurschreiber:
If `count` is 0, this might lead to some random error value be returned from `git_odb_expand_ids`.

(From the most recent coverity report.)


arthurschreiber:
Made irrelevant by #3670. :smile:

PR #3670, opened 2016-03-09
Fixes for `gid_odb_expand_ids`

vmg:
Fixes the issue with leaking the error code for the last OID in the array, and the ODB refreshes that were (still) being performed on missing objects.

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -49,7 +49,36 @@ static git_cache *odb_cache(git_odb *odb)
 	return &odb->own_cache;
 }
 
+static int odb_otype_fast(git_otype *type_p, git_odb *db, const git_oid *id);
 static int load_alternates(git_odb *odb, const char *objects_dir, int alternate_depth);
+
+static git_otype odb_hardcoded_type(const git_oid *id)
+{
+	static git_oid empty_blob = {{ 0xe6, 0x9d, 0xe2, 0x9b, 0xb2, 0xd1, 0xd6, 0x43, 0x4b, 0x8b,
+					   0x29, 0xae, 0x77, 0x5a, 0xd8, 0xc2, 0xe4, 0x8c, 0x53, 0x91 }};
+	static git_oid empty_tree = {{ 0x4b, 0x82, 0x5d, 0xc6, 0x42, 0xcb, 0x6e, 0xb9, 0xa0, 0x60,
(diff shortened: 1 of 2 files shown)


vmg:
cc @ethomson @carlosmn

src/odb.c
```diff
@@ -829,52 +895,32 @@ int git_odb__read_header_or_object(
...
-		raw->data = git__calloc(1, sizeof(uint8_t));
-		return 0;
-	} else if (!git_oid_cmp(id, &empty_tree)) {
-		raw->type = GIT_OBJ_TREE;
-		raw->len = 0;
-		raw->data = git__calloc(1, sizeof(uint8_t));
-		return 0;
-	} else {
-		return GIT_ENOTFOUND;
+	if (error == GIT_PASSTHROUGH) {
```
vmg: Note that we can know distinguish between the backends being unable to read headers, and the backends not finding the header at all.

Also, I've actually implemented ODB refreshing for `git_odb_read_header`. Before that, we wouldn't refresh the ODB on `ENOTFOUND` under no circumstances, we would just fall back to `git_odb_read`, which does refresh.

src/odb.c
```diff
@@ -811,11 +850,38 @@ int git_odb_read_header(size_t *len_p, git_otype *type_p, git_odb *db, const git
...
+
+	for (i = 0; i < db->backends.length && error < 0; ++i) {
+		backend_internal *internal = git_vector_get(&db->backends, i);
+		git_odb_backend *b = internal->backend;
+
+		if (only_refreshed && !b->refresh)
+			continue;
+
+		if (b->read_header != NULL)
+			error = b->read_header(len_p, type_p, b, id);
```
ethomson: We will now stop this loop immediately and return `E_PASSTHROUGH` if any of the `read_header`s returned `E_PASSTHROUGH`.  I feel like the correct thing to do here would be to remember that one of the backends returned `E_PASSTHROUGH`, but continue.  One of the other backends may be the one that actually has this object.

This lets that later-ordered backend service the `read_header` request, on the assumption that `read_header` is cheaper than the full `read`.

ethomson:
@vmg - fixing my bugs since 2003.

Okay, maybe not that long ago, but a more accurate year doesn't rhyme and in any case, it's been a long ass time

vmg:
:joy:

Issue #3662, opened 2016-03-08
Destroy all mingw32?


ethomson:
At present, we have a variety of issues in 32 bit mingw.  I believe that at present you can build libgit2 in mingw (32 bit) but a variety of things don't work and our unit tests do not pass.  In fact, our tests haven't passed in mingw32 since we set up the build environment.

Since we have a finite number of dollars to spend on CI builds, I propose that we get rid of the mingw32 builds.  We have ignored those failures since always, and will continue to unless somebody comes forward and makes mingw32 work in fullness.  Nobody has done so, thus far.

There is a risk that this means that we will bitrot completely on 32 bit mingw.  I am not upset by this.  If anybody is actually concerned about mingw 32, please let me know and also tell me how this is useful to you despite not having passed a test suite, well, ever...

As much as I wish we would kill mingw outright, I do not propose changing our commitment to mingw64.

cc @mstrap @tkelman

tkelman:
As long as the 32 bit mingw-w64 build with `i686-w64-mingw32` still works, no objection from me for killing legacy mingw.org `i686-pc-mingw32`. That's an obsolete zombie project that should have put a big "deprecated" notice on their website years ago.

vmg:
Clang is getting real close to being able to compile any C code in Windows using the native Win APIs sooOooooo... Yeah I strongly support nuking anything Mingw related. :0

arthurschreiber:
:+1: from me as well. The Ruby DevKit has moved from 32-bit mingw to mingw-w64 years ago, so there'll be no impact on Rugged either.

tkelman:
> Clang is getting real close to being able to compile any C code in Windows using the native Win APIs 

Given libgit2 is all C, I'd be amazed if clang didn't work for it already on Windows. C++ exception handling and MSVC-compatible debug info are the last few things missing in clang and those should be close enough for most purposes with 3.8 (just released). Clang isn't really a standalone toolchain on Windows though, you need either MSVC or MinGW for headers and a CRT. So Clang doesn't really eliminate the need for MinGW-w64 to exist or make "Windows == MSVC" any closer to an accurate assumption, though clang is a way for a lot of projects to have a C compiler that's less of a special snowflake than MSVC but more easily compatible with it than GCC.

carlosmn:
Removed in #3671

PR #3671, opened 2016-03-09
Remove CI support for mingw32

carlosmn:
The tests have never run successfully and we do have successful builds
of mingw-w64, so remove these CI builds which do not add value.

---

As mentioned in #3662 we still use the w64 variant.


carlosmn:
This saves about five minutes per full Appveyor build, which is nice, but it still takes a full half hour.

tkelman:
Can also delete
```
matrix:
  allow_failures:
    - GENERATOR: "MSYS Makefiles"
      ARCH: 32
```
from `appveyor.yml`

carlosmn:
Thanks, the allowed-failure change had got lost in the weird way I cherry-picked this.

PR #3673, opened 2016-03-10
commit: add function to attach a signature to a commit

carlosmn:
In combination with the function which creates a commit into a buffer,
this allows us to more easily create signed commits.

---

The tests end up rather messy with all that inline text, but putting the text outside doesn't really look better.

### DIFF
--- a/include/git2/commit.h
+++ b/include/git2/commit.h
@@ -440,6 +440,27 @@ GIT_EXTERN(int) git_commit_create_buffer(
 	size_t parent_count,
 	const git_commit *parents[]);
 
+/**
+ * Create a commit object from the given buffer and signature
+ *
+ * Given the unsigned commit object's contents, its signature and the
+ * header field in which to store the signature, attach the signature
+ * to the commit and write it into the given repository.
+ *
+ * @param out the resulting commit id
+ * @param commit_content the content of the unsigned commit object
+ * @param signature the signature to add to the commit
+ * @param signature_field which header field should contain this
+ * signature. Leave `NULL` for the default of "gpgsig"
+ * @return 0 or an error code
+ */
+GIT_EXTERN(int) git_commit_create_with_signature(
+	git_oid *out,
+	git_repository *repo,
+	const char *commit_content,
+	const char *signature,
+	const char *signature_field);
+
 /** @} */
 GIT_END_DECL
 #endif
--- a/src/commit.c
+++ b/src/commit.c
@@ -820,3 +820,67 @@ int git_commit_create_buffer(git_buf *out,
 	git_array_clear(parents_arr);
 	return error;
 }
+
+/**
+ * Append to 'out' properly marking continuations when there's a newline in 'content'
+ */
+static void format_header_field(git_buf *out, const char *field, const char *content)
+{
+	const char *lf;
+
+	assert(out && field && content);
+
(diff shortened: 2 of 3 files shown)


mikeando:
Probably worth commenting that no validation is done on the signature, and that the user needs to generate the signature themselves. (If you're feeling particularly generous it would be nice to show in the comment the way to generate a signature that matches the default gpgsig key)

carlosmn:
While we can make the working more explicit about the caller providing the signature, I don't see that we hint at the library performing validation anywhere. I'd worry that adding such a comment would imply that validation is the norm.

Figuring out how to shell out to your favourite program is a problem domain which libgit2 stays away from. Even a simplified example would be problematic.

mikeando:
Yep, I agree that shelling out is completely out of line for libgit2.  However I'd assume a user could link with gpg and use its functions to get what they need for a signature. 

Git picks up and uses gpg automatically, which also isn't appropriate here, but is partly why I think its worth making things a little more obvious for the user. 

A change to the wording here (and maybe wherever else we do signing in the same kind of way) would suffice IMO, since just reading the function comments doesn't really give the use much of an idea how to actually hook this all together to produce something correct. And the fun part is its relatively easy to hook it all together in a way that it runs, and has an incorrect signature... (since we don't (can't) validate the provided signature)

(Note - I don't envisage _me_ ever using this function - signing commits is something I've not had to do in the past and surely don't envisage having to hook it into my applications. I'm purely considering other potential users here - and maybe they're all savvy enough that my concerns are moot...)

ethomson:
I agree with @mikeando that it would be nice to have some nice documentation around here.  I think that we're still building the pieces that will let you work with signed commits.  It's going to take a while to get this done and then to get it right.  Until then, though, I don't want to spend a lot of time documenting moving parts.  Instead, let's document it nicely and build an example once we're done.

matthauck:
Really cool to see this functionality being added to libgit2! I think this will open up really cool features for applications being built on libgit2 (like GitKraken) to enable gpg signing/verification. Glad there is progress being made here in tooling!

However, this PR rather disappointing from the consumer side of things that this is all libgit2 appears to be planning for commit signing. I understand you don't want to shell out or bring in a gpg dependency. Totally get that. But I think you could do more here for a fully-fledged git commit signature api.  

Granted, I'm just now looking at libgit2 for the first time, but I have no clue how one should go about constructing the `commit_content` string. I don't see anything that does this for you in commit.h. 

Better ways of doing this that come to mind:
1. Make `git_commit_create_with_signature` take all the same parameters as `git_commit_create` as well as signature_field and a signature_callback function that calculates the signature from the given commit contents string. This may not be a pattern libgit2 uses, and maybe there's some memory ownership issues here since this is a C api and we can't use `std::string`.
2. First make the user call `git_commit_create_buffer` to create the commit in a `git_buf` (or some other intermediate structure?), and then have `git_commit_create_with_signature` take the `git_buf *` instead of an unstructured `const char *`. 

Hope you can see from this side of things this function feels like a (really exciting!) bridge halfway... :cry:

kuwv:
@carlosmn I noticed that I can create signed commits but no references are created. I'm going to dig into the source to verify but I wanted to ask (as it's not covered here). I guess it should be expected though as no ref is passed to either function.

PR #3677, opened 2016-03-11
Coverity fixes round 7

pks-t:
Mostly fixes involving ignored return codes. Down to a defect density of 0.25 and 29 defects outstanding in my fork.

While we're at it: I've got a custom user model that fixes how Coverity models vectors (see [user_model.c](https://github.com/pks-t/libgit2/blob/f943300f1001e146445802df3ee3cbefd6bf352e/script/user_model.c)). It fixes some issues but requires us to manually upload the file into Coverity. Do we want this added?

### DIFF
--- a/src/config_cache.c
+++ b/src/config_cache.c
@@ -86,7 +86,8 @@ int git_config__cvar(int *out, git_config *config, git_cvar_cached cvar)
 	struct map_data *data = &_cvar_maps[(int)cvar];
 	git_config_entry *entry;
 
-	git_config__lookup_entry(&entry, config, data->cvar_name, false);
+	if ((error = git_config__lookup_entry(&entry, config, data->cvar_name, false)) < 0)
+		return error;
 
 	if (!entry)
 		*out = data->default_value;
--- a/src/refdb_fs.c
+++ b/src/refdb_fs.c
@@ -962,6 +962,7 @@ static int packed_write(refdb_fs_backend *backend)
 
 	for (i = 0; i < git_sortedcache_entrycount(refcache); ++i) {
 		struct packref *ref = git_sortedcache_entry(refcache, i);
+		assert(ref);
 
 		if (packed_find_peel(backend, ref) < 0)
 			goto fail;
--- a/src/blame.c
+++ b/src/blame.c
@@ -178,7 +178,7 @@ const git_blame_hunk *git_blame_get_hunk_byline(git_blame *blame, size_t lineno)
 	return NULL;
 }
 
-static void normalize_options(
+static int normalize_options(
 		git_blame_options *out,
 		const git_blame_options *in,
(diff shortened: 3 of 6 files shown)


src/config_cache.c
```diff
@@ -86,7 +86,8 @@ int git_config__cvar(int *out, git_config *config, git_cvar_cached cvar)
 	struct map_data *data = &_cvar_maps[(int)cvar];
 	git_config_entry *entry;
 
-	git_config__lookup_entry(&entry, config, data->cvar_name, false);
+	if ((error = git_config__lookup_entry(&entry, config, data->cvar_name, false)) < 0)
+		return error;
```
carlosmn: This bypasses setting our output variable to the default value. We were doing the equivalent of detecting a not-found entry (or whatever error) by checking `entry` against `NULL` in the check below.

This indicates we don't have nearly enough testing here, since now we're not setting `*out` anymore, but the tests still pass.

pks-t: In fact it does not, see the commit message. `git_config__lookup_entry` with `no_errors == false` calls `get_entry` with `GET_NO_MISSING`. So we shouldn't get an error for missing entries.

src/refdb_fs.c
```diff
@@ -962,6 +962,8 @@ static int packed_write(refdb_fs_backend *backend)
 
 	for (i = 0; i < git_sortedcache_entrycount(refcache); ++i) {
 		struct packref *ref = git_sortedcache_entry(refcache, i);
+		if (!ref)
+			goto fail;
```
carlosmn: Can this happen? We've taken a write lock and are iterating over the amount of entries which the data structure has told us exist. If we're shut up a warning, `assert(ref)` would be more appropriate imo.

pks-t: Ah, okay. I've seen the other loops where we iterate over the cache using the same pattern. Didn't realize we acquire a write lock here, so `assert`ing this should indeed be the way to go.

src/config_cache.c
carlosmn: That's what happens when you look at the diff on its own. Ok, let's leave it as-is.

pks-t:
@carlosmn @ethomson @vmg 
Any comment about the usermodel C file that I've proposed in this PR's description?

ethomson:
@pks-t That looks reasonable to me - is this just a one-time manual upload or is this something automated?

pks-t:
@ethomson In theory there should be a `--user-model-file` switch for `coverity-analyze` which you feed a compiled user model which might be obtained by another Coverity tool. But we actually aren't using `coverity-analyze` and documentation is certainly lacking.

So for now this would be a one-time manual upload and nothing automated. But right now I don't see anything that should be modeled anyway, so I don't expect it to change frequently.

PR #3398, opened 2015-08-25
examples: Add '-i' option for log

JIghtuse:

PR #3679, opened 2016-03-12
examples/log: add a option -i to ignore case

pranitbauva1997:
A commonly used option -i to ignore case when used with --grep
to find patterns in commits.

Signed-off-by: Pranit Bauva pranit.bauva@gmail.com

### DIFF
--- a/examples/log.c
+++ b/examples/log.c
@@ -13,6 +13,7 @@
  */
 
 #include "common.h"
+#include <ctype.h>
 
 /**
  * This example demonstrates the libgit2 rev walker APIs to roughly
@@ -52,6 +53,7 @@ struct log_options {
 	int show_diff;
 	int skip, limit;
 	int min_parents, max_parents;
+	int ignore_case;
 	git_time_t before;
 	git_time_t after;
 	const char *author;
@@ -67,8 +69,13 @@ static void print_commit(git_commit *commit);
 static int match_with_parent(git_commit *commit, int i, git_diff_options *);
 
 /** utility functions for filtering */
-static int signature_matches(const git_signature *sig, const char *filter);
-static int log_message_matches(const git_commit *commit, const char *filter);
+static int signature_matches(const git_signature *sig, const char *filter,
+				const int ignore_case);
+static int log_message_matches(const git_commit *commit, const char *filter,
+				const int ignore_case);
+char *find_str(const char *haystack, const char *needle, const int ignore_case);
+char *str_ignorecase_str(const char *haystack, const char *needle);
+char *tolowerstr(const char *str);
 
 int main(int argc, char *argv[])
 {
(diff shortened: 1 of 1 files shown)


carlosmn:
The same question as in #3189 and #3398. How widespread is `strcasestr()`?

pranitbauva1997:
@carlosmn My bad! strcasestr() is not widespread. It depends on systems and is not present in many systems and has some dependencies. Since this is a "no dependence" software, I have written my own kind of strcasestr() namely str_ignorecase_str() whose efficiency is not as good as its counter part.

examples/log.c
```diff
@@ -477,3 +490,35 @@ static int parse_options(
...
+
+char *tolowerstr(const char *str) {
+	char *final;
+	int len = 0;
+	int i = 0;
+	while(str[i++] != '\0')
+		;
+	len = i;
+
+	final = (char *) malloc(sizeof(char) * len);
```
ethomson: A stylistic tip - please don't cast `malloc`.  http://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc

Also, you do not need `sizeof(char)` here.  That's `1`.

ethomson:
Although I'm not super concerned about the efficiency of our `strcasestr`, it seems like we should be able to get by without `malloc`ing and copying the results?

examples/log.c
pranitbauva1997: Oh! I wasn't aware about the malloc thing and sizeof(char). I was under the impression that the size of char is dependent on the compiler. But I just found that its defined in the standard and its always 1. Thanks!

pranitbauva1997:
@ethomson I tried your suggestion. It just gave a warning when I compiled on my system. I thought its okay. But this gives problem with others. Should I just revert back to the previous version?

ethomson:
Sorry if I was unclear - I wasn't suggesting that you change the case of the string in-place; indeed, changing the case of the commit message in-place is not appropriate.

I meant that you should be able to implement a `strcasestr` that works in place, without changing the case of the strings...  a very inefficient implementation might look something like:

```
for (; *haystack; haystack++) {
    char *n, *h;
    bool match = true;

    for (n = needle, h = haystack; *n && *h; n++, h++) {
        if (*n != *h)
            break;
    }

    if (*n != *h)
        continue;

    return haystack;
}
```

(Note this is not tested, I'm just coding in a web browser.)

ethomson:
Hi @pranitbauva1997 - are you interested in trying a `strcasestr` that will not modify the contents of the buffer passed to it?

pranitbauva1997:
@ethomson I am sorry I almost forgot about this PR. I was busy with git. I will look into this during the weekend. Thanks for reminding!

ethomson:
This is quite outdated, and I'm closing this to tidy up the outstanding pull requests.  Thanks for working on this!

Issue #3682, opened 2016-03-14
Consider limiting the TLS ciphers we use by default


carlosmn:
OpenSSL is not particularly proactive in changing the defaults, and it's probably too much to expect each user of the library to worry about this when it generally starts from a PoC which gains features.

So let's set our own list of safe ciphers to use. Because TLS is terrible in general, we should also provide an escape hatch and provide a `git_libgit2_opts()` way of the user letting us know not to do it.

Unless someone comes up with a better list, we'll probably go with https://github.com/ruby/ruby/blob/trunk/ext/openssl/lib/openssl/ssl.rb#L22 but exclude RC4 (because lol).

 - [ ] Do this for OpenSSL
 - [ ] Do this for SecureTransport
 - [ ] Do this for WinHTTP

tiennou:
I'd like to discuss this, because I wanted to tackle #4418 failing because our cipherlists are inconsistent between backends.

Right now, only OpenSSL handles our `GIT_OPT_SET_SSL_CIPHERS` option. The mbedTLS backend inherited its own separate (handtuned by yours truly) copy of it because the names didn't match, and I'm looking at SecureTransport, which uses [enums](https://developer.apple.com/documentation/security/1550981-ssl_cipher_suite_values?language=objc).

It means the OpenSSL names are somewhat, well, "[not-that-standard](https://testssl.sh/openssl-rfc.mapping.html)". [RFC5246](https://tools.ietf.org/html/rfc5246#appendix-A.5) defines a list, and that's what mbedTLS, SecureTransport (and GNUTLS, just for kicks) use, apart that the latter has no easy string-mapping API. It seems the libcurl ones depends on what backend it uses, and WinHTTP does its thing too, so 😭.

Anyway, I think a shared master list of default ciphers is preferable, so would it be a problem if I broke the API/OpenSSL users by switching names ? At least we could have a "consistent API" in that regard…

I don't think there's any way around that but to have a "table of ciphers + names" somewhere though, so I want to know if I'm overthinking it 😉.

pks-t:
I agree the current API has some problems, especially considering that there is not even documentation regarding the cipher list's format. And it definitly is unfortunate that the user has to be aware of what SSL backend is in use. So I'm definitly in favor of having a mapping table for matching "our" name with "their" name, with "their" being the respective backend that is currently in use.

PR #3683, opened 2016-03-14
Setup better defaults for OpenSSL ciphers

dbussink:
This ensures that when using OpenSSL a safe default set of ciphers
is selected. This is done so that the client communicates securely
and we don't accidentally enable unsafe ciphers like RC4, or even
worse some old export ciphers.

Implements the first part of https://github.com/libgit2/libgit2/issues/3682

cc @carlosmn

### DIFF
--- a/include/git2/common.h
+++ b/include/git2/common.h
@@ -149,6 +149,7 @@ typedef enum {
 	GIT_OPT_SET_SSL_CERT_LOCATIONS,
 	GIT_OPT_SET_USER_AGENT,
 	GIT_OPT_ENABLE_STRICT_OBJECT_CREATION,
+	GIT_OPT_SET_SSL_CIPHERS,
 } git_libgit2_opt_t;
 
 /**
@@ -260,6 +261,11 @@ typedef enum {
  *		> example, when this is enabled, the parent(s) and tree inputs
  *		> will be validated when creating a new commit.  This defaults
  *		> to disabled.
+ *	* opts(GIT_OPT_SET_SSL_CIPHERS, const char *ciphers)
+ *
+ *		> Set the SSL ciphers use for HTTPS connections.
+ *		>
+ *		> - `ciphers` is the list of ciphers that are eanbled.
  *
  * @param option Option key
  * @param ... value to set the option
--- a/src/settings.c
+++ b/src/settings.c
@@ -71,10 +71,16 @@ static int config_level_to_sysdir(int config_level)
 }
 
 extern char *git__user_agent;
+extern char *git__ssl_ciphers;
 
 const char *git_libgit2__user_agent()
 {
 	return git__user_agent;
+}
+
+const char *git_libgit2__ssl_ciphers()
+{
+	return git__ssl_ciphers;
 }
 
 int git_libgit2_opts(int key, ...)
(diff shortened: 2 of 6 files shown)


tests/online/badssl.c
```diff
@@ -36,3 +36,12 @@ void test_online_badssl__self_signed(void)
...
 			 git_clone(&g_repo, "https://self-signed.badssl.com/fake.git", "./fake", NULL));
 }
+
+void test_online_badssl__old_cipher(void)
+{
+	if (!g_has_ssl)
+		cl_skip();
+
+	cl_git_fail_with(GIT_ERROR,
+			 git_clone(&g_repo, "https://rc4.badssl.com/fake.git", "./fake", NULL));
```
dbussink: Not really happy about the error type here, but there doesn't seem to be an easy way to get a more specific error.

carlosmn: I'd go with `cl_git_fail()` here since we expect a generic failure. It is indeed sad that we cannot detect this as its own error.

include/git2/errors.h
```diff
@@ -50,6 +50,7 @@ typedef enum {
 	GIT_EUNCOMMITTED    = -22,	/**< Uncommitted changes in index prevented operation */
 	GIT_EDIRECTORY      = -23,      /**< The operation is not valid for a directory */
 	GIT_EMERGECONFLICT  = -24,	/**< A merge conflict exists and cannot continue */
+  GIT_ETLSCIPHERS     = -25,  /**< No ciphers could be negotiated */
```
carlosmn: Are you using spaces for indentation?

carlosmn: Actually if we can't detect this, we should get rid of this error, maybe we'll bring it back if SecureTransport does want to play nice.

dbussink: This shouldn't have gotten in in the first place yeah, removed it.

src/settings.c
```diff
@@ -187,6 +193,22 @@ int git_libgit2_opts(int key, ...)
...
+		{
+			git__free(git__ssl_ciphers);
+			git__ssl_ciphers = git__strdup(va_arg(ap, const char *));
+			if (!git__ssl_ciphers) {
+				giterr_set_oom();
+				error = -1;
+			}
+		}
+#else
+		giterr_set(GITERR_NET, "Cannot set custom ciphers: OpenSSL is not enabled");
```
carlosmn: The error message should start with a lowercase (I know we're not very good about this, but let's get new strings to have lowercase).

carlosmn:
:+1: other than the nit-picking I did.

dbussink:
@carlosmn Fixed up those nit-pickings.

include/git2/common.h
```diff
@@ -260,6 +261,11 @@ typedef enum {
  *		> example, when this is enabled, the parent(s) and tree inputs
  *		> will be validated when creating a new commit.  This defaults
  *		> to disabled.
+ *	* opts(GIT_OPT_SET_SSL_CIPHERS, const char *ciphers)
+ *
+ *		> Set the SSL ciphers use for HTTPS connections.
+ *		>
+ *		> - `ciphers` is the list of ciphers that are eanbled.
```
Therzok: Typo'd `enabled`

PR #3685, opened 2016-03-14
Test memleaks

pks-t:
Two memory leaks pointed out by AppVeyor. There is still one memory leak missing involving TLS and the init/shutdown logic. It somehow seems like a race condition but I haven't been able to figure it out yet as I do not have a Windows installation at hand currently.


ethomson:
:+1:

PR #3687, opened 2016-03-15
CMake: do not overwrite but only append to CMAKE_C_FLAGS_DEBUG 

mstrap:
This is useful to force "smart" IDEs (like CLIon) to use debug
flag -g even it may have decided that "-D_DEBUG" (which is
already present) is sufficient.


carlosmn:
The title seems misleading. This isn't about non-MSVC, we should append regardless of the compiler. Also, who is the author/commiter of the commit and do you have permission to license these changes under our license?

mstrap:
@carlosmn I've fixed my identity to my personal account. Is there anything else I have to do to proof that I have permission to license these changes under your license?

ethomson:
Thanks!

PR #3688, opened 2016-03-15
Fix the examples/diff.c implementation of the -B

saurabhsuniljain:
Implemented command line options to actually look for the optional configuration values.

I wanted to fix the implementation of examples/diff.c I understand i need to check whether there is anything extra after -B like /70% or 20%, and set the values for appropriate flags. So is it okay for me to write a another if else statement inside that ?

I have made some changes please someone review it before I make a pull request.
Thanks :)

### DIFF
--- a/examples/diff.c
+++ b/examples/diff.c
@@ -47,6 +47,10 @@ enum {
 	CACHE_NONE = 2
 };
 
+#define DEFAULT_THRESHOLD 50
+#define DEFAULT_BREAK_REWRITE_THRESHOLD 60
+#define USE_DEFAULT(X) ((X) == 0 || (X) > 100) 
+
 /** The 'opts' struct captures all the various parsed command line options. */
 struct opts {
 	git_diff_options diffopts;
@@ -66,6 +70,7 @@ static void parse_opts(struct opts *o, int argc, char *argv[]);
 static int color_printer(
 	const git_diff_delta*, const git_diff_hunk*, const git_diff_line*, void*);
 static void diff_print_stats(git_diff *diff, struct opts *o);
+int extract_break_rewrites(int *m, int *n, const char *arg);
 
 int main(int argc, char *argv[])
 {
@@ -215,10 +220,13 @@ static int color_printer(
 static void parse_opts(struct opts *o, int argc, char *argv[])
 {
 	struct args_info args = ARGS_INFO_INIT;
+	const char *break_rewrite_arg;
+	int m,n,err;
 
 
 	for (args.pos = 1; args.pos < argc; ++args.pos) {
 		const char *a = argv[args.pos];
+
 
 		if (a[0] != '-') {
 			if (o->treeish1 == NULL)
@@ -290,9 +298,22 @@ static void parse_opts(struct opts *o, int argc, char *argv[])
 			o->findopts.flags |= GIT_DIFF_FIND_COPIES;
 		else if (!strcmp(a, "--find-copies-harder"))
 			o->findopts.flags |= GIT_DIFF_FIND_COPIES_FROM_UNMODIFIED;
-		else if (is_prefixed(a, "-B") || is_prefixed(a, "--break-rewrites"))
-			/* TODO: parse thresholds */
+		else if (optional_str_arg(
+				&break_rewrite_arg, &args, "-B", NULL) || 
+			optional_str_arg(
+				&break_rewrite_arg, &args, "--break-rewrites", NULL)) {
+
 			o->findopts.flags |= GIT_DIFF_FIND_REWRITES;
+			err = extract_break_rewrites(&m, &n, break_rewrite_arg);
+			if (err==-1) 
+				usage("Invalid input format", a);
+			o->findopts.rename_from_rewrite_threshold = (uint16_t) m;
+			o->findopts.break_rewrite_threshold = (uint16_t) n;
+			if (USE_DEFAULT(o->findopts.rename_from_rewrite_threshold))
+				o->findopts.rename_from_rewrite_threshold = DEFAULT_THRESHOLD;
+			if (USE_DEFAULT(o->findopts.break_rewrite_threshold))
+				o->findopts.break_rewrite_threshold = DEFAULT_BREAK_REWRITE_THRESHOLD;
+		}
 		else if (!match_uint16_arg(
 				&o->diffopts.context_lines, &args, "-U") &&
 			!match_uint16_arg(
@@ -334,4 +355,31 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
 
 	git_buf_free(&b);
 	git_diff_stats_free(stats);
+}
+
+int extract_break_rewrites(int *m, int *n, const char *arg)
+{
+	int err;
+	const char *slash = strchr(arg, '/');
+
+	/* slash either not found or not at the beginning of arg, parse n */
+	if (slash != arg) {
+		err = sscanf(arg, "%d", n);
+		/* Invalid input format, return -1 to indicate error */
+		if (err != 1 || err == EOF)
+			return -1;
+	} else {
+		*n = 0;
+	}
+
+	/* slash found, parse m one character after it */
+	if (slash != NULL) {
+		err = sscanf(slash + 1, "%d", m);
+		/* Invalid input format, return -1 to indicate error */
+		if (err != 1 || err == EOF)
+			return -1;
+	} else {
+		*m = 0;
+	}
+	return 0;
 }


pranitbauva1997:
Firstly, never make any pull request from master branch. Its a very bad practice. Let's say this PR got merged, then it would create an extra 3-way merge commit when you try to pull. If the PR doesn't get merged, then you would need to git reset --hard. And if you pull the commits into your master branch before the PR being accepted then the commits in this PR will get updated and it will have a very cluttered history!

examples/diff.c
```diff
@@ -290,9 +290,31 @@ static void parse_opts(struct opts *o, int argc, char *argv[])
 			o->findopts.flags |= GIT_DIFF_FIND_COPIES;
 		else if (!strcmp(a, "--find-copies-harder"))
 			o->findopts.flags |= GIT_DIFF_FIND_COPIES_FROM_UNMODIFIED;
-		else if (is_prefixed(a, "-B") || is_prefixed(a, "--break-rewrites"))
+		else if (is_prefixed(a, "-B") || is_prefixed(a, "--break-rewrites")){
```
ethomson: Please follow the existing style for indentation, braces, etc.  It's roughly the Linux Kernel style:  https://www.kernel.org/doc/Documentation/CodingStyle will be a good introduction.

examples/diff.c
```diff
@@ -290,9 +290,31 @@ static void parse_opts(struct opts *o, int argc, char *argv[])
...
+				uint16_t m = 0;
+				int i = 3;
+				while(a[i]!='%' || a[i]!=' '){
+					m *= 10;
+					m += a[i]-'0';
+					i++;
+				}
+				if(m<=100) o->findopts.break_rewrite_threshold = m;
+				//else there is error, show message;
+			}else if(a[2]!=' '){
```
ethomson: If `a[2] == '\0'` then you'll continue walking off the end of this string.

ethomson: When would `a[2] == ' '` ever be true?  When would somebody quote a space in a `-B` and have that be valid?

examples/diff.c
```diff
@@ -290,9 +290,31 @@ static void parse_opts(struct opts *o, int argc, char *argv[])
...
+				uint16_t n = 0;
+				int i = 2;
+				while(a[i]!='%' || a[i]!=' '){
+					n *= 10;
+					n += a[i]-'0';
+					i++;
+				}
```
ethomson: This looks like the code block directly above it.  I suggest that you follow the pattern around this and use `match_str_arg`.  You can then create a helper function to parse the resulting string...

examples/diff.c
```diff
@@ -290,9 +290,31 @@ static void parse_opts(struct opts *o, int argc, char *argv[])
...
+				if(n<=100) o->findopts.rename_from_rewrite_threshold = n;
+				//else there is error, show message;
```
ethomson: Does git show an error message here?

examples/diff.c
saurabhsuniljain: Where can I read about this function match_str_arg ?

examples/diff.c
saurabhsuniljain: Right now, no error message is shown, I just added that line to add it in future.

examples/diff.c
ethomson: It's in `common.c`.

examples/diff.c
ethomson: My question is about what *git* does when you give it a number > 100.  We should match its behavior.

examples/diff.c
saurabhsuniljain: a[2] == ' ', would be true when someone use -B but doesn't provide any extra option like [<n>][</m>]

ethomson: Is that true on Windows?  It's certainly not true on Unix.

examples/diff.c
saurabhsuniljain: I currently don't know about what git does, but surely greater than 100 is not valid.

examples/diff.c
saurabhsuniljain: You are talking about a[2] == '\0' or a[2] == ' ' ?

ethomson: On Unix, the shell splits your arguments for you.  The only time you would have a space in `a` is if somebody literally quoted it.  For example:  `git diff "-B 20%"`.  Which nobody would likely do.  Instead, if you ran:  `git diff -B 20%`, then `20%` is the _next argument_ and this current `a` would not have a space in it.  It would simply end.  That's why using the helper functions that already exist to understand this sort of thing is very helpful.

This is illustrated very quickly by building this and running it on a POSIX-type environment - running `-B 20%` crashes immediately.  I would recommend you download some Linux distribution and run it in a VM, as we want to make sure that libgit2 and the examples work everywhere.

saurabhsuniljain: I do use linux (Ubuntu 14.04), will check there.

examples/diff.c
saurabhsuniljain: I couldn't read the common.c file, [common.c](https://libgit2.github.com/libgit2/ex/HEAD/common.html)

examples/diff.c
carlosmn: That's why we need to know how git behaves. Does it ignore it? Adjust it? Raise an error? Print a warning?

examples/diff.c
carlosmn: All the docs page does is show comments side-by-side if they're formatted in a particular way. Read the source directly if this isn't working.

saurabhsuniljain: So, what I should I do is check whether the string has "-B/" or "-B" as prefix and then take remaining string as [n][/m] options, but how can I distinguish between only -B, when no [n/[/m] options are provided ?

saurabhsuniljain: How to build the library on my local repository, I followed the instructions in README.md for "Building libgit2 - Using CMake", it passed on my system but failed when I made pull request.

saurabhsuniljain:
How to build the library on my local repository, I followed the instructions in README.md for "Building libgit2 - Using CMake", it passed on my system but failed when I made pull request.

pranitbauva1997:
@saurabhsuniljain Firstly, do you have all the required packages installed? I guess the developer packages of libgit2 is required. If yes, then you should set up an environment variable "LD_LIBRARY_PATH" to reflect to the directory of install (by default /usr/local/lib")

ethomson:
@saurabhsuniljain I'm not sure what you mean when you say that it failed when you made the pull request?  It looks like this built successfully in the pull request's CI system.  Why do you think that it failed?

saurabhsuniljain:
First I made a pull request which failed, then I squashed all the commits into single commit and then made a pull request.

saurabhsuniljain:
Is there any problem with the code ?
@ethomson @carlosmn

carlosmn:
This is still not in the same style as the rest of the codebase. There are variables declared in the middle of blocks, which is not portable.

saurabhsuniljain:
@carlosmn I have updated the code, please have a look.

examples/diff.c
```diff
@@ -335,3 +350,60 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
 	git_buf_free(&b);
 	git_diff_stats_free(stats);
 }
+
+int string_to_int(const char ** out)
```
carlosmn: This should just be `const char *` and if it's an input parameter, it definitely shouldn't be named `out`.

examples/diff.c
```diff
@@ -335,3 +350,60 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
...
+
+	while(*out[i]!='%') {
+		to_return *= 10;
+		to_return += *out[i] - '0';
+		i++;
+	}
+	return to_return;
+}
+
+struct pair extract_from_string(const char **out)
```
carlosmn: Same here, there's no reason for this to be a pointer-to-pointer.

examples/diff.c
```diff
@@ -335,3 +349,58 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
...
+	int i = 0;
+
+
+	while(cmd_opt[i]!='%') {
+		to_return *= 10;
+		to_return += cmd_opt[i] - '0';
+		i++;
+	}
+	return to_return;
+}
```
carlosmn: This function doesn't make sure that the user did pass in a number. You could pass in "foo" and this would use the difference between `'f'` and `'0'` as the number. It also only terminates on `'%'` which means that if you do not pass in the percent, it'll run off the end of the string, causing a segfault.

I would recommend using a standard function like `strtol()` like we do elsewhere to parse strings into numbers. You can then check that what comes after the number is allowed in this context.

saurabhsuniljain: @carlosmn If string contains a number but doesn't contain a % sign in the end, should I take it as valid input or give a error message ?

carlosmn: Do whatever git does. It may error out, let it through or reset to some default. Look at the code of this command in git to see how it deals with its arguments.

saurabhsuniljain:
@carlosmn Made the changes as suggested, please have a look. I think this should be fine.

examples/diff.c
```diff
@@ -335,3 +357,47 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
...
+
+	if(*cmd_opt=='/') {
+		p.first = DEFAULT_THRESHOLD;
+		cmd_opt++;
+		p.second = strtol(cmd_opt,endptr,0);
+		if(**endptr!='%') giterr_set_str(GITERR_INVALID, "Invalid input format.");
+	} else {
+		p.second = DEFAULT_BREAK_REWRITE_THRESHOLD;
+		p.first = strtol(cmd_opt,endptr,0);
+		if(**endptr!='%') giterr_set_str(GITERR_INVALID, "Invalid input format.");
```
carlosmn: `giterr_set_str()` is for setting the library's error for this thread; a consumer of the library (which all of these examples are) should not be using it to set its error. I also don't see how this function is signaling an error. How does a caller of `extract_from_string()` know that the option was invalid? `extract_m_n()` doesn't seem to care whether there was an error.

The program should either die if there are errors parsing the command or this function has to signal an error to the caller.

This isn't checking whether `strtol` reported an error. Passing just `%` as the option would make this code think that I passed in a valid number 0, which is not true. AFAICT git also restricts this to base-10 numbers, so we shouldn't be letting the user of this diff program choose to use octal or hex for their percentages.

The code is also not formatted according to the rules we follow in the codebase. Don't put the code inside the if on the same line and leave a space between `if` and the parenthesis.

examples/diff.c
```diff
@@ -47,6 +47,15 @@ enum {
...
 };
 
+#define DEFAULT_THRESHOLD 50
+#define DEFAULT_BREAK_REWRITE_THRESHOLD 60
+#define USE_DEFAULT(X) ((X) == 0 || (X) > 100)
+
+struct pair {
+	int first;
+	int second;
+};
```
carlosmn: Our style would be to pass these two as named output parameters from the function and use the return value as an `int` to signal errors.

saurabhsuniljain: How to signal the caller that there has been error in input, by setting the values of first or second to negative ? Also will it be valid for the program to just die without explaining the cause of the error ?

carlosmn: The way we do it in the library is to return 0 for success and a negative number for error, which is common is unix systems. The purpose of a `die()` function is to show the user an error and then exiting the program with an appropriate error code. You should show the error to the user, which AFAICT is more than this does right now.

saurabhsuniljain: Which library to include for die function or we need to define one of our own ?

carlosmn: There is no library. Each application has their own needs for such a function. You can look at git or other projects for some samples, but such a function generally simply prints whatever went wrong to `stderr` and exits the program with `1` or sometimes a more specific error code, depending on what the program does.

saurabhsuniljain: Done, have a look.

examples/diff.c
```diff
@@ -335,3 +357,54 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
...
+	if (*cmd_opt=='/') {
+		p.first = DEFAULT_THRESHOLD;
+		cmd_opt++;
+		p.second = strtol(cmd_opt,endptr,'A');
+		if (**endptr!='%') 
+			die("Invalid input format.");
+	} else {
+		p.second = DEFAULT_BREAK_REWRITE_THRESHOLD;
+		p.first = strtol(cmd_opt,endptr,'A');
+		if (**endptr!='%')
```
carlosmn: The comment about error checking is still valid. This function still has no error handling.

examples/diff.c
```diff
@@ -335,3 +357,54 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
...
+	exit(EXIT_FAILURE);
+}
+
+struct pair extract_from_string(const char *cmd_opt)
+{
+	struct pair p;
+	char **endptr;
+
+
+	if (*cmd_opt=='/') {
```
carlosmn: The slash is not only at the front. The format is `-B[<n>][/<m>]` and the explanation in the manpage mentions that these are two values handled differently. Not having the slash at the front doesn't mean that `m` should be the default. You can write `-B10%/20%` which this does not cover.

examples/diff.c
```diff
@@ -335,3 +357,54 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
...
+		p.first = DEFAULT_THRESHOLD;
+		cmd_opt++;
+		p.second = strtol(cmd_opt,endptr,'A');
+		if (**endptr!='%')
```
carlosmn: This itself should not be an error. It is possible to write `-B10/20` without git complaining about it.

saurabhsuniljain:
@carlosmn Made some changes have a look.

saurabhsuniljain:
@carlosmn @ethomson If you guys are done with GSoC process, can you please review my patch.

examples/diff.c
```diff
@@ -335,3 +357,105 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
...
+			p.second = tmp;
+		}
+	} else if (**endptr!='\0') {
+		die("Invalid input format");
+	}
+
+	return p;
+}
+
+struct pair extract_m_n(struct args_info *args, const char *opt)
```
tiennou: This still returns a struct ? I think `int extract_break_rewrites(int *m, int *n, const char *arg)` would be nicer. Also, if using `sscanf` above makes it shorter, it could be merged in this one instead of being in a separate function.

examples/diff.c
```diff
@@ -335,3 +357,105 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
...
+				} else if (**endptr != '\0') {
+					die("Invalid input format");
+				} else {
+					p.second = tmp;
+				}
+			} else {
+				p.second = tmp;
+			}
+		} else if (**endptr != '\0') {
+			die("Invalid input format");
```
tiennou: I think `die()` should be done inside parse_opt (when this returns != 0). It's IMHO generally better to act (esp. when you intend to exit) as "high" in the abstraction as you can.

Maybe that can use `errx` [here](http://man7.org/linux/man-pages/man3/err.3.html), but I don't know how portable that is. YMMV.

examples/diff.c
```diff
@@ -290,9 +305,16 @@ static void parse_opts(struct opts *o, int argc, char *argv[])
 			o->findopts.flags |= GIT_DIFF_FIND_COPIES;
 		else if (!strcmp(a, "--find-copies-harder"))
 			o->findopts.flags |= GIT_DIFF_FIND_COPIES_FROM_UNMODIFIED;
-		else if (is_prefixed(a, "-B") || is_prefixed(a, "--break-rewrites"))
-			/* TODO: parse thresholds */
+		else if (is_prefixed(a, "-B") || is_prefixed(a, "--break-rewrites")) {
```
saurabhsuniljain: I think it should be char **break_rewrite_arg, right?

carlosmn:
You still haven't reacted to the multiple comments about error-checking the result of `strtol`. I'm having a hard time believing you are testing this out, since you are passing it a base of 65, which should always cause an error, and thus always return `0`.

saurabhsuniljain:
Firstly, I am not passing the base of 65, [here](http://man7.org/linux/man-pages/man3/strtol.3.html) it clearly says "the letter 'A' in either uppercase or lowercase represents 10". 
Secondly I have managed to take both as input at the same time, like `20%/70%`.
I am also accepting it as valid input if there is not any `%` sign.
For only a `%` sign not any number, I agree that function `struct pair extract_from_string(const char *cmd_opt)` sets the values of m and n to their default values and doesn't report an error. I will take care of this in my next commit / where I will be parsing the command using method suggested by @tiennou as above.

Any other error issues to be handled ?
Also, when I call a die function, how I am supposed to return a negative number to suggest that an error has occurred, I think you mean to say that I shouldn't call the die function from the helper functions but from the `parse_opts(struct opts *o, int argc, char *argv[])` function, when these helper functions return an error with negative number.

carlosmn:
That pareagraph is not describing the `base` argument. What it has to say about it is

>The  strtol()  function  converts the initial part of the string in nptr to a long integer value according to the given base, which must be between 2 and 36 inclusive, or be the special value 0.

The next two paragraphs (including what you quote) describe the string to be parsed. Checking `errno` with `'A'` as the base does cause an `EINVAL` error, as can be reproduced by this simple program, which prints out `strtol failed: Invalid argument` when executed

```C
#include <stdio.h>
#include <stdlib.h>
#include <errno.h>

int main(char **argv, int argc)
{
	char *endptr;
	long int res;

	errno = 0;
	res = strtol("15", &endptr, 'A');
	if (errno != 0) {
		perror("strtol failed");
		return EXIT_FAILURE;
	}

	return EXIT_SUCCESS;
}
```

saurabhsuniljain:
Oh, sorry misunderstood, that I thought A is passed as base, but it is the value that has to be interpreted as 10 in the string to be parsed. Will take care of this. Will be putting the pull request soon.

tiennou:
Note that I *also* misread the `strtol` man page, because I was finding using 'A' as a proxy for 10 "obscure", so there's that :wink:.

@saurabhsuniljain Yeah, that's what I meant — the helpers returns a negative number to represent an error, or 0 for success, leaving it to the parse machinery to do whatever it want with the result (exiting the process, retrying, whatever).

examples/diff.c
```diff
@@ -335,3 +357,105 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
...
+
+struct pair extract_from_string(const char *cmd_opt)
+{
+	struct pair p;
+	char **endptr;
+	const char * cmd_opt2;
+	int tmp = strtol(cmd_opt,endptr,'A');
+
+
+	if (tmp==0) {
```
carlosmn: `sscanf` would be another option. But even if it's not that, this code needs to be refactored into some helper function or loop as it has multiple copies of the same parsing code in different branches.

examples/diff.c
carlosmn: `errx` is a BSD extension. The POSIX function is called `perror`. If this were generic code, it shouldn't die, but this is specific to this program so dying next to the error makes sense as you have the information here that you would want to present.

saurabhsuniljain:
@tiennou @ethomson 
I tried to work using `sscanf` but this works fine with a space in between but fails when there is a `/`.
You can see the exemplar code with the defined function [here](http://ideone.com/1x0un0).

examples/diff.c
```diff
@@ -215,11 +227,14 @@ static int color_printer(
 static void parse_opts(struct opts *o, int argc, char *argv[])
 {
 	struct args_info args = ARGS_INFO_INIT;
+	const char **break_rewrite_arg;
```
tiennou: Couldn't this be `const char *` ?

tiennou:
Now that I see the `sscanf` code, it might be cleaner to check for the existence of '/' in the string. Something like that :
```
int err;
char *slash = strchr(arg, '/');
if (slash == NULL) {
    /* Parse m */
    err = sscanf(arg, "%d", m);
    if (err != 1) /* error */
if (slash == NULL) {
    /* error */
} else if (slash != arg) {
    /* n present */
    err = sscanf(arg, "%d", n);
    if (err != 1) /* error */
}

File: examples/diff.c [hunk 18]
```diff
@@ -290,9 +305,22 @@ static void parse_opts(struct opts *o, int argc, char *argv[])
...
-		else if (is_prefixed(a, "-B") || is_prefixed(a, "--break-rewrites"))
-			/* TODO: parse thresholds */
+		else if (optional_str_arg(
+				break_rewrite_arg, &args, "-B", NULL) || 
+			optional_str_arg(
+				break_rewrite_arg, &args, "--break-rewrites", NULL)) {
+
 			o->findopts.flags |= GIT_DIFF_FIND_REWRITES;
+			err = extract_break_rewrites(&m, &n, *break_rewrite_arg);
+			if (err==-1)
```
tiennou: `if (err != 0) usage("Invalid --break-rewrite argument", break_rewrite_arg);` instead of declaring `die()` ?

examples/diff.c
tiennou: That structure can go right ?

examples/diff.c
saurabhsuniljain: No because in the `optional_str_arg` needs first argument to be const char **, look [here](https://github.com/libgit2/libgit2/blob/77394a27af283b366fa8bb444d29670131bfa104/examples/common.h).

examples/diff.c
saurabhsuniljain: This goes according to the standards, but here this is for demonstration purpose (examples/diff.c) and user should know the reason why it failed and program should terminate. We need to just provide a functionality to the user, how to use the library, after that it's his choice how to use.

saurabhsuniljain: Also checking for the existence of `/` is not necessary as the `sscanf` functions returns how many parameters it has filled successfully. So if `/` is not there I will end up getting 0 and it will be handled accordingly.
If there's any other reason or correct testcase which might end up reported as error then there's a need to look into this.
You can do testing [here](http://ideone.com/1x0un0) if required.

saurabhsuniljain: After giving it a second thought I found your approach to be better, because on legitimate correct testcases my function was failing. You can find my new code on the same link as above.
Please review this function here before I made a new pull request [I had made enough of them].

`
int extract_break_rewrites(int *m, int *n, const char *arg)
	{
		char m_str[10];
		char n_str[10];
		int err;
		const char *slash = strchr(arg, '/');
		if (slash != arg) {
			err = sscanf(arg, "%d", n);
			if (err != 1 || err == EOF)
				return -1;
		} else {
			*n = 0;
		}
		if (slash != NULL) {
			err = sscanf(slash + 1, "%d", m);
			if (err != 1 || err == EOF)
				return -1;
		} else {
			*m = 0;
		}
		return 1;
	}
`

saurabhsuniljain: Why haven't this code been reviewed yet, if there aren't any mistakes please merge the pull request.
If there is any other problem please tell.
@ethomson @carlosmn @tiennou

examples/diff.c
tiennou: This is *still* unused.

examples/diff.c
```diff
@@ -66,6 +75,9 @@ static void parse_opts(struct opts *o, int argc, char *argv[]);
 static int color_printer(
 	const git_diff_delta*, const git_diff_hunk*, const git_diff_line*, void*);
 static void diff_print_stats(git_diff *diff, struct opts *o);
+int extract_break_rewrites(int *m, int *n, const char *arg);
+void die(char* message);
```
tiennou: Unused function.

examples/diff.c
tiennou:
```
const char *arg;
optional_str_arg(&arg, ...);
``` ?

File: examples/diff.c (hunk 19 above)
Comment (saurabhsuniljain) [inline examples/diff.c]: Yeah, I will remove them, also need to remove pair struct above, is there is other problem with code ?

File: examples/diff.c [hunk 20]
```diff
@@ -335,3 +363,27 @@ static void diff_print_stats(git_diff *diff, struct opts *o)
 	git_buf_free(&b);
 	git_diff_stats_free(stats);
 }
+
+int extract_break_rewrites(int *m, int *n, const char *arg)
+{
```
tiennou: - Comments would be nice on the 2 main `if`s. I can't tell quickly enough was happens, and *I'm the one that proposed it* (and I remember adding some because "clarity").
- Function returns either `1` or `-1`, both of which are not 0, and thus always "false". In C, `0` means success, everything else is "error", because if someone does the obvious `if(!function) exit(-1);` they'll exit with wonder.
- m_str/n_str are unused.

saurabhsuniljain: - Ok, will add comments on the top of two `if`'s 

- Will change return 1 to return 0 and other code accordingly

- Will delete all unused items.
Is anything left, please suggest now only, because this has been pending for quite some time now.

saurabhsuniljain: @tiennou any updates ? Should I make a final pull request.

ethomson: There are comments here that are still outstanding.  For example, `struct pair` remains unused.

saurabhsuniljain: Made all the changes mentioned in all the comments, I read all the comments once again. I pretty sure there should not be any mistakes now.

ethomson: I'm just going through the things that @tiennou has pointed out one by one.  The next one that jumps out at me is that `m_str` and `n_str` are unused.

saurabhsuniljain: @ethomson @tiennou is there any problem ? Please review this.

tiennou:
Oh well...

```
$ ./diff --color -B 10%/20%
[1]    14874 segmentation fault  ./diff --color -B 10%/20%
```
```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   diff                          	0x000000010efea38c optional_str_arg + 188 (common.c:71)
1   diff                          	0x000000010efe9a12 parse_opts + 1682 (diff.c:301)
2   diff                          	0x000000010efe908f main + 111 (diff.c:87)
3   libdyld.dylib                 	0x00007fff8cffe5ad start + 1
```

If you had fixed the note I made about `const char **` then it would have worked. You'll also have to pass the address around in `extract_break_rewrite`.

saurabhsuniljain:
@tiennou Can you test it now ?

tiennou:
```
./examples/diff --color -B | less
[1]    28642 segmentation fault  ./examples/diff --color -B
```
```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_platform.dylib      	0x00007fff8a59b71c _platform_strchr$VARIANT$Haswell + 28
1   diff                          	0x0000000100992fa4 extract_break_rewrites + 36 (diff.c:363)
2   diff                          	0x0000000100992a6a parse_opts + 1754 (diff.c:307)
3   diff                          	0x000000010099209f main + 111 (diff.c:87)
4   libdyld.dylib                 	0x00007fff8cffe5ad start + 1
```

ethomson:
@saurabhsuniljain We'd like you to create test cases, and run through them yourself.  They could be unit tests (you could, for example, write a shell script that runs the `diff` example) or you could simply run through manual acceptance testing.

There's an expectation of well-tested pull requests - it's not very polite to ask other people to do your testing for you!  :)

Thanks!

tiennou:
I've tested that renaming commit.c to rename.c and smashing some keys in makes it show up in the output but that's pretty much all I did.

I do wonder if we want to error on things like `-B 0x10`, `-B -20` or something ;-).

saurabhsuniljain:
@tiennue @ethomson I haven't created test cases before, but I will try to make some. Could you just provide me with sample test case, like do I have to just pass the command or anything else ?

tiennou:
I don't think that's what @ethomson meant — examples are not part per se of libgit2's test suite. Just make sure that the example is sound, and looks good. Aaaand doesn't segfault :wink:.

ethomson:
Exactly so.  I think that you should have a list of commands that you run and ensure that you get the expected output out of them.  Probably you should create a list of these commands and run through them manually as part of your acceptance testing of the feature.  (I think that @tiennou has gotten you started with some examples.  :)

Certainly you should validate that they produce some useful output and that they do not crash.

saurabhsuniljain:
Should I create a commands for which there should be error?

saurabhsuniljain:
@carlosmn @tiennou Can you give a example of script which does this kind of testing somewhere in this library, so that I am familiar with how to call the program (which arguments and all to pass), what environment to assume and how optimized that script is ?
Also should I deliberately add erroneous test cases, to check actually the program reports error for those ?

tiennou:
As I said, examples are just documentation — they're not under any kind of automated testing. What we where hinting at is that running that example can be considered a good check, so you can be sure the example works as intended (arguably so it demonstrate usage of the library, which is the point of examples).

Now, given that I was able to get it to segfault instantly while testing, I'll assume you never actually built it. I'll refer you to the "Building libgit2 - Using CMake" in README.md, you'll just have to add a `-DBUILD_EXAMPLES=ON` option to the first cmake command, and you should end up with the various examples executables built under `build/examples`, and you'll be able to run that diff executable and do some manual tests.

ethomson:
This looks like it hasn't been updated in a while; I appreciate the effort here, but I'm going to close this.  Please let me know if you have further questions.

PR #3689, opened 2016-03-16
tree_iterator: fix pathlist handling for directories

mstrap:
tree_iterator was only working properly for a pathlist containing
file paths. In case of directory paths, it didn't match children
which contradicts GIT_DIFF_DISABLE_PATHSPEC_MATCH and
is different from index_iterator and fs_iterator.

As a consequence head-to-index status reporting for a specific
directory did not work properly -- all files have been reported
as added.

### DIFF
--- a/src/iterator.c
+++ b/src/iterator.c
@@ -351,6 +351,9 @@ typedef struct {
 	int path_ambiguities;
 	bool path_has_filename;
 	bool entry_is_current;
+
+	/* used internally to track the currently matching root) */
+	char* pathlist_match_root;
 } tree_iterator;
 
 static char *tree_iterator__current_filename(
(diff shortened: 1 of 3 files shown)


ethomson:
Hey, thanks for this.  I've been working on refactoring the iterators because they've become a bit clunky to work with.  I mentioned this in another issue, but that's not super visible.  :/  I regret not opening a PR sooner, but I didn't think that anybody would be ambitious enough to dig in themselves.  Now I feel really bad that we've duplicated some effort here.

I cherry picked your tests into #3691.  They're very helpful.

I'm not done refactoring #3691, but if you wanted to take it for a spin and see if there are additional problems you can find, that would be amazing.  (And if not, the tests you wrote were already a big help.)

I expect to finish up #3691 quickly, but it has already taken quite a while.  So I'm going to leave this open for a little bit in case we want to merge it before the big refactor is ready.

Thanks again!

mstrap:
Obsoleted by #3691.

PR #3690, opened 2016-03-16
win32: choose the page size as our value for the page size

carlosmn:
We've been using the allocation granularity instead for this value,
but his refers to VirtualAlloc() rather than anything we do. Using this
as a value for the pool means every pool would become at least 64kB with
foreseeable results on memory pressure.

Do use the system's page size as our value for the page size.

### DIFF
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -449,7 +449,7 @@ static void hash_partially(git_indexer *idx, const uint8_t *data, size_t size)
 static int write_at(git_indexer *idx, const void *data, git_off_t offset, size_t size)
 {
 	git_file fd = idx->pack->mwf.fd;
-	size_t page_size;
+	size_t mmap_alignment;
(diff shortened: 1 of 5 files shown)


ethomson:
Hmm.  It's okay to change the page size that we use for pools, but mmap emulation is a bit stricter.  We need to observe the allocation granularity for [`MapViewOfFile`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa366761(v=vs.85).aspx):

> _dwFileOffsetLow_ [in]
A low-order **DWORD** of the file offset where the view is to begin. The combination of the high and low offsets must specify an offset within the file mapping. They must also match the memory allocation granularity of the system. That is, the offset must be a multiple of the allocation granularity. To obtain the memory allocation granularity of the system, use the **GetSystemInfo** function, which fills in the members of a **SYSTEM_INFO** structure.

Issue #3675, opened 2016-03-10
git_status_options: support non-recursive directories status for specified pathspecs


mstrap:
`git_status_options` should provide an additional flag to only query the status for directories matching the pathspecs **non-recursively**.

Reason: when monitoring file systems, it's not always possible to exactly tell which files have been added/modified/deleted, but change information is only available on a per-directory level (for instance in case of OSX FSEvents [1]). To refresh an affected directory efficiently, it would be helpful to invoke a non-recursive status query for this directory.

Btw, in case this should already be possible I've posted a related question on stackoverflow.com ([2]).

[1] https://developer.apple.com/library/mac/documentation/Darwin/Reference/FSEvents_Ref/
[2] http://stackoverflow.com/questions/35582759

PR #3691, opened 2016-03-17
Some FANTASTIC iterator refactoring

ethomson:
I've been on the warpath to refactor iterators a bit:  I've been adding debt to them for a while now and they need a bit cleanup to get them back to the nice model of clarity that they started with.  The major goals here are threefold:  remove recursion (per #3576), clean things up a bit (things have gotten a bit hard to maintain) and fix some pathlist bugs (per #3609 and now #3689).

As part of "cleaning up", I'm adding some helpful comments around the tricky bits.  Although a lot of this seems straightforward when you write it, there is some tricky interdependencies here that take time to untangle.  So I've added docs so that - hopefully - the next time you need to deal with `advance_over_with_status` you understand what it does.

Thus far, I've attacked the tree and filesystem (aka workdir) iterators.

Although perf isn't the _first_ goal, it's important to ensure that we aren't slipping.  As much as I distrust microbenchmarks, I'm seeing an approximately 3-5% speedup on the workdir iterator and the tree iterator when ignore casing.  There are likely a couple of reasons for this:  in the workdir iterator, we no longer allocate the entries that we're about to throw away.  (This was a hold over from the abstraction layer:  the filesystem iterator would allocate all the entries, the workdir iterator's callback would then promptly remove the ones it wasn't interested it.)  The case _sensitive_ tree iterator, however, is about 30% faster.  Wheeeeee!

There's still a bunch of things to be done here; I'm opening this for visibility because other people are now seeing some of the issues that this is meant to fix and I want to try to reduce the amount of duplication in efforts.

I've cherry-picked the test cases from #3576, #3609 and #3689.  It's possible that there's overlap in the test cases that can be trimmed; I haven't done that yet.

Todo:

- [x] Don't run that crazy expensive test from #3576 on the CI builds by default, it's super slow on crappy filesystems.
- [x] Refactor the index iterators a bit
- [x] Fix all the extraneous whitespace. Xcode's editor is really terrible when it comes to whitespace.  Just abysmal.

Some additional refactoring would be nice to clean things up:

- [x] ~~Don't let people call `iterator_current` when they mean `iterator_advance`.  If the first call to an iterator is `current`, it acts like `advance`.  It would be nice if this *failed* and `current` returned the already advanced `current` entry.~~
- [x] Creating filesystem vs workdir iterators is still pretty clunky.  Clean this up.
- [x] ~~Require a repository for all iterators?  This would be some more polish on #3612.~~

There have been some changes to the internal iterator API:  this is all cosmetic, and I have pondered spinning out a version that has the same API.  This would let people drop this new `iterator.c` in to whatever version they happen to be running; if @jeffhostetler and friends are interested in taking this for a spin in their current version of the library then I will do that, which would be nice to get us additional testing of the workdir and index iterators.


src/object_api.c
```diff
@@ -42,6 +42,10 @@ git_repository *git_commit_owner(const git_commit *obj)
 	return git_object_owner((const git_object *)obj);
 }
 
+int git_commit_dup(git_commit **out, git_commit *obj)
+{
+	return git_object_dup((git_object **)out, (git_object *)obj);
```
txdv: Maybe you want to check the otype of the obj?

mstrap:
This obsoletes my changes at #3689 which I had started in preparation to address issue #3675 and for which first steps can be found at d59834cd8a83a9fa78bbaac87a8d924fa17ea4e4. Now I'll better postpone my work here and wait for your refactoring to finish. Though, from first debugging, I think your refactoring will make implementing non-recursive status easier (especially regarding the `tree_iterator`) and here are just a few thoughts from this perspective (I understand that this is still work in progress):

The main task for non-recursive status is to hook into the pathlist-matching methods. For the `tree_iterator`, this should work straight forward now in `iterator_pathlist_next_is`. For the `index_iterator` it's `iterator_pathlist_walk__contains` and for the `filesystem_iterator` it's `filesystem_iterator_examine_path` now. When looking at these three methods, they look somewhat similar and I'm wondering whether the core pathlist-matching code can be pulled out into one method?

When debugging the old code, for me `iterator_pathlist_walk__contains` was easier to understand than `iterator_pathlist__match`. So I'm wondering whether `filesystem_iterator` can work in a similar way to the other two `iterators`? So you could get rid of the `iterator_pathlist_search_t` and related logic.  Drawback is that `pathlist_walk__contains` seems to be less efficient than the binary search, but I'm wondering whether there are any real-world use cases where you (1) just request index-workdir status and (2) have a large pathlist (which should be the only scenario where the drawback becomes effective).

jeffhostetler:
@ethomson You might also take a look at #3568 while you're looking at the iterator code.

ethomson:
> Though, from first debugging, I think your refactoring will make implementing non-recursive status easier (especially regarding the tree_iterator) and here are just a few thoughts from this perspective (I understand that this is still work in progress):

Yeah.  I think it will be fairly easy to bolt on a `depth` that you're interested in.

> iterator_pathlist_walk__contains was easier to understand than iterator_pathlist__match.

Definitely easier to understand.  The nice thing about tree and index iterators is that they're already sorted according to git's rules.  Which is to say that some directory named `foo` is sorted as if it was `foo/`.  As a result, we can simply walk two sorted lists of paths side-by-side.  Efficient and easy.

The workdir, on the other hand, is tricky because the big goal of workdir iteration (and the pathlist matching) is to avoid `stat`s at all costs.  So unfortunately you can't know whether some file `foo` is a directory until you stat it.  Which is why we build that indeterminate state when looking at the pathlist matching:  if we have `foo/` in a pathlist (or some entry like `foo/bar`), then we're interested in some on-disk item `foo` if and only if it's a directory.  So we stat it to find out.

This sucks, but it saves us from `stat`ing every file in the folder that we enter so that we can sort them.  That's a real perf killer.  So instead we binary search into the pathlist, which is less nice in every way but saves us `stat`s.

>  I'm wondering whether there are any real-world use cases where you (1) just request index-workdir status and (2) have a large pathlist (which should be the only scenario where the drawback becomes effective).

There are, in fact.  When you're doing a merge, we merge the tree trees that you're interested in to get some merge result.  We then compare the resultant tree with your `HEAD` to see what files need to be updated on disk.  Once we have the list of changed paths *now* we have to examine them on-disk to validate the merge rules:  Git lets you have files changed in the working directory when you perform a merge if and only if none of those files were actually changed in the merge itself.  To compute this we do a index to workdir diff, limiting ourselves with the pathlist of the changed paths.

(All the code to move the pathlist down into the iteration for `stat` avoidance was predicated exactly upon the `stat` avoidance in workdir diffs for this scenario.)

src/object_api.c
ethomson: This is really just :lipstick: so I'm not sure that's necessary.  For `obj` checking to ever find a false positive, then a caller would have done something like:

    git_tree *tree;
    git_commit *dup;
    git_commit_dup(&dup, (git_commit *)tree);

Which probably makes no sense.  Recall that `git_object_dup` and friends are internal-only, so this isn't anything that would help our callers.  But maybe I am missing the use case?

carlosmn: `git_object_dup()` is public, but in either case, if you're passing in the wrong type, you've purposefully done a wild cast in your code, so these would only be for convenience so the compiler sees the types it wants in the function signature.

The actual object type is irrelevant for these functions anyway, as it only touches common areas.

ethomson: That's right, `git_object_dup` _is_ public.  I don't know why I avoided making the `git_commit_dup` and friends public.  They seem useful.

jeffhostetler:
@ethomson While you're looking at the workdir iterator stuff, I thought I'd add #3696 to the list.  Specifically how to handle deeply nested trees on Windows.  Could be as simple as another iterator flag to control diving....

mstrap:
> > Though, from first debugging, I think your refactoring will make implementing non-recursive status
> > easier (especially regarding the tree_iterator) and here are just a few thoughts from this perspective
> >  (I understand that this is still work in progress):

> Yeah. I think it will be fairly easy to bolt on a depth that you're interested in.

@ethomson Is this something I should have a look at, based on the current state of your changes? Maybe it will generate some more feedback from my side. Or do you think that remaining TODOs will affect pathlist-matching code (significantly), so it's too early now?

Btw, would you actually allow to specify an (int) "depth" or just recursive vs. non-recursive? Personally, I don't see a use case for depth=2 (or more). But, um, we had that different viewpoint on use-cases already before :)

ethomson:
I don't particularly care whether it's an arbitrary depth or not.  My initial thought was that depth would just be easier to implement, since you're simply counting frames.  But that's _total_ depth from the workdir root, not depth from the start of some path, which is I think what you want.

mstrap:
> But that's total depth from the workdir root, not depth from the start of some path, which is I think
> what you want.

Yes, it's actually the relative depth I'm looking for, ultimately allowing me to query the status of all files within one single sub-directory and do no other work. Using a *total* depth + a *single*-entry-pathlist would work as well here. So if a maximum *total* depth has other advantages, why not.

ethomson:
Okay, well, I think that I've done about as much as I can do here.  This is only a minor reduction in the number of lines of code, but it's a significant reduction in the complexity.  There is now no recursion, and now the `advance` function of each iterator does all the heavy lifting to determine the state and identify the next entry.  (The `advance_into` and `advance_over` have been moved to supporting roles that twiddle a few bits and then call `advance`.)

Again, the improvements here are maintainability, but there are real perf gains as well.  On my Mac laptop, on an optimized release build, I see minor improvements on the workdir iterator (1% when ignoring case, 4% when not), decent improvement on the tree iterator (no change when ignoring case, 10% when not) and really delightful improvements on the index iterator (46% when ignoring case, 57% when not).

I suspect that we'll see further improvements to the tree iterator once this is rebased on top of #3702 since that will do fewer sorts and fewer allocations.  (Yay!)

ethomson:
(As I've mentioned, I sort of hate microbenchmarks, but I'm publicizing them anyway.  These all came from iterating the workdir, HEAD and index of the linux repository.)

ethomson:
So, it turns out that https://github.com/libgit2/libgit2/pull/3702 made the tree iterator in this PR slower than the *previous* tree iterator (and a *lot* slower than this PR before rebasing on top of #3702.)

However, rebasing *this PR* on top of #3704, which removes the unnecessary vector manipulation, is a *big* improvement.  My bullshit benchmarks show an 11% improvement in case insensitive tree iteration and 23% improvement in the case insensitive.

Exciting stuff!

tests/iterator/workdir.c
```diff
@@ -0,0 +1,1428 @@
...
+	cl_git_pass(git_iterator_advance(&e, i)); /* a */
+	cl_assert_equal_s("a", e->path);
+
+	cl_git_pass(git_iterator_advance(&e, i)); /* c/foo */
+	cl_assert_equal_s("c/foo", e->path);
+
+	cl_git_pass(git_iterator_advance(&e, i)); /* d */
+	cl_assert_equal_s("d", e->path);
+
+	cl_must_pass(p_chmod("empty_standard_repo/r/b", 0777));
```
carlosmn: Should we be testing here that we do still get a `GIT_ITEROVER` to make sure the entry isn't lurking behind where we looked?

tests/iterator/workdir.c
```diff
@@ -0,0 +1,1428 @@
...
+	cl_git_pass(git_vector_insert(&filelist, "k.b"));
+	cl_git_pass(git_vector_insert(&filelist, "k/1"));
+	cl_git_pass(git_vector_insert(&filelist, "k/a"));
+	cl_git_pass(git_vector_insert(&filelist, "kZZZZZZZ"));
+	cl_git_pass(git_vector_insert(&filelist, "L/1"));
+
+	g_repo = cl_git_sandbox_init("icase");
+
+	/* Test iterators with default case sensitivity, without returning
+	 * tree entries (but autoexpanding.
```
carlosmn: Missing closing parens.

carlosmn:
`checkout_target_fullpath` won't work well with concurrency, although we can fix that if and when we do implement that.

f16216 should be amended not to say it's fixing anything since it's just the tests.

carlosmn:
Looks good other than that, let's fix the conflicts and :shipit:

tests/iterator/workdir.c
ethomson: Good call; fixed.

ethomson:
@carlosmn I rebased this and addressed your concerns.

Indeed the checkout target buf is :hankey: for concurrency, but note that this was the behavior before (I just changed the name to more accurately reflect that it's the target path, which is probbaly not _necessarily_ the workdir path.)  When we tackle multithreaded checkout this will indeed be one of the areas that need to be addressed.

PR #3702, opened 2016-03-20
Reuse a tree's buffer and allocate constant-sized entries in an array

carlosmn:
The memory management for the small and sometimes numerous tree-owned entries has non-trivial effects on performance. We previously tried to bundle up allocations with a pool, but this backfires quite significantly on Windows where 32-bit address spaces are common.

These patches still try to bundle up allocations, but let the system allocator do more of the work. As we've already gone though the work of allocating and inflating the tree object's buffer, and we store it in a cache so it lives in memory after we load the tree, make use of its memory to store the id and filename for the entries. The entries now simply point to them.

This makes the entries have a constant size so we can use a `git_array_t(git_tree_entry)` to allocate them in blocks and free them with a single call while keeping the ownership in the tree.

This does mean extra indirection so we will have to make sure that this does help as we expect, particularly with large diffs.

/cc @niik

### DIFF
--- a/src/tree.h
+++ b/src/tree.h
@@ -17,13 +17,14 @@
 struct git_tree_entry {
 	uint16_t attr;
 	uint16_t filename_len;
-	git_oid oid;
+	git_oid *oid;
 	bool pooled;
-	char filename[GIT_FLEX_ARRAY];
+	const char *filename;
 };
 
 struct git_tree {
 	git_object object;
+	git_odb_object *odb_obj;
 	git_vector entries;
 	git_pool pool;
 };
--- a/src/tree.h
+++ b/src/tree.h
@@ -17,16 +17,15 @@
 struct git_tree_entry {
 	uint16_t attr;
 	uint16_t filename_len;
-	git_oid *oid;
-	bool pooled;
+	const git_oid *oid;
 	const char *filename;
 };
 
 struct git_tree {
 	git_object object;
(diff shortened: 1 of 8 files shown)


src/tree.h
```diff
@@ -17,15 +17,15 @@
...
-	char filename[GIT_FLEX_ARRAY];
+	const git_oid *oid;
+	const char *filename;
 };
 
 struct git_tree {
 	git_object object;
+	git_odb_object *odb_obj;
+	git_array_t(git_tree_entry) entries_arr;
 	git_vector entries;
```
ethomson: I'm curious why we're keeping this vector here now that we have the array...  Is it just for keeping the entries sorted (differently than the array)?  And if so, is that necessary?

carlosmn: We're keeping it so our homing search works. It expects to deal with a `git_vector` and I didn't want to deal with reimplementing the binary sorting algo on top of a different data structure.

vmg: A reimplementation could probably be very useful in the long run, tho. I can give it a go after this PR is merged.

ethomson: Cool.  We had talked once about how we do a sort on tree entries instead of trusting that the entries in the tree were written correctly, and how we should ⚡ this behavior.

I'm happy to merge this as-is but I think we should drop the vector when time permits.  (I would also be happy to tackle this later.)

PR #3704, opened 2016-03-22
tree: drop the now-unnecessary entries vector

ethomson:
Remove the now-unnecessary entries vector.  Add `git_array_search`
to binary search through an array to accomplish this.


carlosmn:
Ah, good old MSVC which doesn't like doing arithmetic on void pointers. I suppose we'll need a `unsigned char *` in a few places.

ethomson:
Right you are.  Fixed on up.  Thanks, MSVC!

PR #3705, opened 2016-03-22
win32: open read-only files with read sharing

carlosmn:
By default Windows does not let two processes read the same file
concurrently even if both want read-only access.

Make the fileops function through which reference and config reading go
call CreateFile directly so we can tell Windows it's ok for multiple
processes to read from these files.

---

The flags I took from the Win32 `p_mmap()` so I'd appreciate someone seeing if they make sense.

### DIFF
--- a/src/fileops.c
+++ b/src/fileops.c
@@ -89,10 +89,46 @@ int git_futils_creat_locked_withpath(const char *path, const mode_t dirmode, con
 
 int git_futils_open_ro(const char *path)
 {
+#ifdef GIT_WIN32
+
+	git_win32_path buf;
+	HANDLE h;
+	int fd;
+
+	if (git_win32_path_from_utf8(buf, path) < 0)
+		return -1;
+
+	/*
+	 * We use this function to read references and config, which should be readable by multiple
+	 * proceses concurrently.
+	 */
+	h = CreateFileW(buf,
+		GENERIC_READ,
+		FILE_SHARE_READ | FILE_SHARE_DELETE,
+		NULL,
+		OPEN_EXISTING,
+		FILE_FLAG_BACKUP_SEMANTICS,
+		NULL);
+	
+	if (h == INVALID_HANDLE_VALUE) {
+		giterr_set(GITERR_OS, "failed to open '%s'", path);
+		return -1;
+	}
+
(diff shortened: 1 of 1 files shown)


ethomson:
Not sure about those arguments.  I think that `FILE_SHARE_DELETE` will let somebody delete the file out from under us and (unlike a POSIXy system) our `read`s would just start failing.  I worry that we're not quite set up for that.  I think that `FILE_SHARE_READ` is probably a nice middle ground?

Similarly, I don't think that we want `FILE_FLAG_BACKUP_SEMANTICS`...  I think we just want `FILE_ATTRIBUTE_NORMAL`...

Maybe one of the Microsofties can tell us how `_open` maps to `CreateFileEx`?

cc @jeffhostetler @whoisj

ethomson:
I also wonder if we shouldn't gut `p_open` instead of `git_futils_open_ro`, and provide an internal flag to specify that sharing is okay.  (Perhaps even turning this on by default?  When would we _not_ want this?)  That way, of course, any caller could take advantage of this.

carlosmn:
I did look at reimplementing `p_open` with `CreateFileW` but mapping all the `open` semantics got a bit awkward so I went for this instead, which should cover the uses where we do want to share (and we could probably port the other ones for clarity if nothing else).

I do think that if we open a file read-only we always want sharing. It matches what we'd expect from the posixy systems anyway.

carlosmn:
The error we saw with filebuf makes me think that `_wopen` does actually let multiple processes read from the same file, making this moot.

That would unfortunately make the read-blocked config files down to git.

carlosmn:
By looking at the CRT source, It turns out that [`_wopen`](https://msdn.microsoft.com/en-us/library/z0kc8e3z.aspx) ends up calling (though some amazing feats of indirection [`_wsopen`](https://msdn.microsoft.com/en-us/library/w7sa2b22.aspx) with a `shflag` of `_SH_DENYNO` which "Permits read and write access" so we already had read sharing.

This makes it a bit more mysterious why we sometimes see blocked reads from files.

carlosmn:
It seems this was just a red herring. We do still see locked files, but this does not seem to be the cause.

PR #3706, opened 2016-03-23
[GSoC] Adding a basic example of `git reset`

fanzeyi:
Hello,

I am opening this pull request because I am applying for the git project GSoC 2016.

This pull request implemented an example of `git reset` using libgit2's API. It only provides a basic functionality of `git reset`. (The missing part is files path related.)

Example:

```shell
$ ./reset --hard HEAD
$ ./reset HEAD~1
$ ./reset --soft HEAD~1
```

It only supports three reset mode due to libgit2 currently only support these three.

This is my first time submitting actual code to open source project like `libgit2`. I have tried my best to format my code to fit the coding style and other conventions. Please let me know if there are anything I need to change or add.

Thanks.

---------

Question about GSoC 2016: 

I would like to show my proposal to the community but it seems the mail list of `libgit2` is already closed. Where should I post it?

### DIFF
--- a/examples/reset.c
+++ b/examples/reset.c
@@ -1,0 +1,109 @@
+/*
+ * libgit2 "reset" example - shows how to reset HEAD and working tree.
+ *
+ * Written by the libgit2 contributors
+ *
+ * To the extent possible under law, the author(s) have dedicated all copyright
+ * and related and neighboring rights to this software to the public domain
+ * worldwide. This software is distributed without any warranty.
+ *
+ * You should have received a copy of the CC0 Public Domain Dedication along
+ * with this software. If not, see
+ * <http://creativecommons.org/publicdomain/zero/1.0/>.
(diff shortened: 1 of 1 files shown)


examples/reset.c
```diff
@@ -0,0 +1,109 @@
...
+
+		if (a[0] != '-') {
+			/**
+			 * Options have to be in front of any revision/commit/filepath.
+			 * Only support `git reset [<type>] [<commit>]` for now.
+			 **/
+			opt->rev = a;
+			break;
+		} else if (!strcmp(a, "-q") || !strcmp(a, "--quiet"))
+			opt->quiet = 1;
```
carlosmn: Do we do anything with the quiet option? We shouldn't put it there just because git does if we don't output anything. Alternatively we could output whatever messages git does.

PR #3711, opened 2016-03-27
Add GIT_REPOSITORY_OPEN_FROM_ENV flag to respect $GIT_* environment vars

joshtriplett:
git_repository_open_ext provides parameters for the start path, whether to search across filesystems, and what ceiling directories to stop at.  git commands have standard environment variables and defaults for each of those, as well as various other parameters of the repository. To avoid duplicate environment variable handling in users of libgit2, add a GIT_REPOSITORY_OPEN_FROM_ENV flag, which makes git_repository_open_ext automatically handle the appropriate environment variables. Commands that intend to act just like those built into git itself can use this flag to get the expected default behavior.

git_repository_open_ext with the GIT_REPOSITORY_OPEN_FROM_ENV flag respects $GIT_DIR, $GIT_DISCOVERY_ACROSS_FILESYSTEM, $GIT_CEILING_DIRECTORIES, $GIT_INDEX_FILE, $GIT_NAMESPACE, $GIT_OBJECT_DIRECTORY, and $GIT_ALTERNATE_OBJECT_DIRECTORIES.  In the future, when libgit2 gets worktree support, git_repository_open_env will also respect $GIT_WORK_TREE and $GIT_COMMON_DIR; until then, git_repository_open_ext with this flag will error out if either $GIT_WORK_TREE or $GIT_COMMON_DIR is set.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -359,7 +359,8 @@ static int find_repo(
 	git_buf path = GIT_BUF_INIT;
 	struct stat st;
 	dev_t initial_device = 0;
-	bool try_with_dot_git = ((flags & GIT_REPOSITORY_OPEN_BARE) != 0);
+	int min_iterations;
+	bool in_dot_git;
 	int ceiling_offset;
 
 	git_buf_free(repo_path);
@@ -367,13 +368,27 @@ static int find_repo(
 	if ((error = git_path_prettify(&path, start_path, NULL)) < 0)
 		return error;
 
-	ceiling_offset = find_ceiling_dir_offset(path.ptr, ceiling_dirs);
+	/* in_dot_git toggles each loop:
+	 * /a/b/c/.git, /a/b/c, /a/b/.git, /a/b, /a/.git, /a
+	 * With GIT_REPOSITORY_OPEN_BARE, we assume we started with /a/b/c.git
+	 * and don't append .git the first time through.
+	 * min_iterations indicates the number of iterations left before going
+	 * further counts as a search. */
+	if (flags & GIT_REPOSITORY_OPEN_BARE) {
+		in_dot_git = true;
+		min_iterations = 1;
+	} else {
+		in_dot_git = false;
+		min_iterations = 2;
+	}
 
-	if (!try_with_dot_git &&
-		(error = git_buf_joinpath(&path, path.ptr, DOT_GIT)) < 0)
-		return error;
+	while (!error && (min_iterations || !(path.ptr[ceiling_offset] == 0 ||
+					      (flags & GIT_REPOSITORY_OPEN_NO_SEARCH)))) {
+		if (!in_dot_git)
+			if ((error = git_buf_joinpath(&path, path.ptr, DOT_GIT)) < 0)
+				break;
+		in_dot_git = !in_dot_git;
 
-	while (!error && !git_buf_len(repo_path)) {
 		if (p_stat(path.ptr, &st) == 0) {
 			/* check that we have not crossed device boundaries */
 			if (initial_device == 0)
@@ -414,17 +429,10 @@ static int find_repo(
 			break;
 		}
 
-		if (try_with_dot_git) {
-			/* if we tried original dir with and without .git AND either hit
-			 * directory ceiling or NO_SEARCH was requested, then be done.
-			 */
-			if (path.ptr[ceiling_offset] == '\0' ||
-				(flags & GIT_REPOSITORY_OPEN_NO_SEARCH) != 0)
-				break;
-			/* otherwise look first for .git item */
-			error = git_buf_joinpath(&path, path.ptr, DOT_GIT);
-		}
-		try_with_dot_git = !try_with_dot_git;
+		/* Once we've checked the directory (and .git if applicable),
+		 * find the ceiling for a search. */
+		if (min_iterations && (--min_iterations == 0))
+			ceiling_offset = find_ceiling_dir_offset(path.ptr, ceiling_dirs);
 	}
 
 	if (!error && parent_path && !(flags & GIT_REPOSITORY_OPEN_BARE)) {
--- a/src/repository.c
+++ b/src/repository.c
@@ -370,11 +370,12 @@ static int find_repo(
 
 	/* in_dot_git toggles each loop:
 	 * /a/b/c/.git, /a/b/c, /a/b/.git, /a/b, /a/.git, /a
-	 * With GIT_REPOSITORY_OPEN_BARE, we assume we started with /a/b/c.git
-	 * and don't append .git the first time through.
+	 * With GIT_REPOSITORY_OPEN_BARE or GIT_REPOSITORY_OPEN_NO_DOTGIT, we
+	 * assume we started with /a/b/c.git and don't append .git the first
+	 * time through.
 	 * min_iterations indicates the number of iterations left before going
 	 * further counts as a search. */
-	if (flags & GIT_REPOSITORY_OPEN_BARE) {
+	if (flags & (GIT_REPOSITORY_OPEN_BARE | GIT_REPOSITORY_OPEN_NO_DOTGIT)) {
 		in_dot_git = true;
 		min_iterations = 1;
 	} else {
@@ -384,10 +385,12 @@ static int find_repo(
 
 	while (!error && (min_iterations || !(path.ptr[ceiling_offset] == 0 ||
 					      (flags & GIT_REPOSITORY_OPEN_NO_SEARCH)))) {
-		if (!in_dot_git)
-			if ((error = git_buf_joinpath(&path, path.ptr, DOT_GIT)) < 0)
-				break;
-		in_dot_git = !in_dot_git;
+		if (!(flags & GIT_REPOSITORY_OPEN_NO_DOTGIT)) {
+			if (!in_dot_git)
+				if ((error = git_buf_joinpath(&path, path.ptr, DOT_GIT)) < 0)
+					break;
+			in_dot_git = !in_dot_git;
+		}
 
 		if (p_stat(path.ptr, &st) == 0) {
 			/* check that we have not crossed device boundaries */
--- a/src/repository.c
+++ b/src/repository.c
@@ -491,6 +491,172 @@ int git_repository_open_bare(
 	return 0;
 }
 
+static int _git_repository_open_ext_from_env(
+	git_repository **out,
+	const char *start_path)
+{
+	git_repository *repo = NULL;
+	git_index *index = NULL;
+	git_odb *odb = NULL;
+	git_buf dir_buf = GIT_BUF_INIT;
+	git_buf ceiling_dirs_buf = GIT_BUF_INIT;
+	git_buf across_fs_buf = GIT_BUF_INIT;
+	git_buf index_file_buf = GIT_BUF_INIT;
+	git_buf namespace_buf = GIT_BUF_INIT;
+	git_buf object_dir_buf = GIT_BUF_INIT;
+	git_buf alts_buf = GIT_BUF_INIT;
+	git_buf work_tree_buf = GIT_BUF_INIT;
+	git_buf common_dir_buf = GIT_BUF_INIT;
+	const char *ceiling_dirs = NULL;
+	unsigned flags = 0;
+	int error;
+
+	if (!start_path) {
+		error = git__getenv(&dir_buf, "GIT_DIR");
+		if (error == GIT_ENOTFOUND) {
+			giterr_clear();
+			start_path = ".";
+		} else if (error < 0)
+			goto error;
+		else {
+			start_path = git_buf_cstr(&dir_buf);
+			flags |= GIT_REPOSITORY_OPEN_NO_SEARCH;
+			flags |= GIT_REPOSITORY_OPEN_NO_DOTGIT;
+		}
+	}
+
+	error = git__getenv(&ceiling_dirs_buf, "GIT_CEILING_DIRECTORIES");
+	if (error == GIT_ENOTFOUND)
+		giterr_clear();
+	else if (error < 0)
+		goto error;
+	else
+		ceiling_dirs = git_buf_cstr(&ceiling_dirs_buf);
+
+	error = git__getenv(&across_fs_buf, "GIT_DISCOVERY_ACROSS_FILESYSTEM");
+	if (error == GIT_ENOTFOUND)
+		giterr_clear();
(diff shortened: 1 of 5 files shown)


tests/repo/discover.c
```diff
@@ -32,6 +32,15 @@ static void ensure_repository_discover(const char *start_path,
 	//across_fs is always 0 as we can't automate the filesystem change tests
 	cl_assert_equal_s(found_path.ptr, expected_path->ptr);
 	git_buf_free(&found_path);
+
+	//test again via environment variables
+	cl_setenv("GIT_DIR", start_path);
+	cl_setenv("GIT_CEILING_DIRECTORIES", ceiling_dirs);
+	cl_git_pass(git_repository_discover_default(&found_path));
+	cl_setenv("GIT_DIR", NULL);
+	cl_setenv("GIT_CEILING_DIRECTORIES", NULL);
```
ethomson: We should probably move these into the test cleanup code - that way they will get unset properly even if `git_repository_discover_default` fails.

vmg:
This looks good to me. I like the careful error handling. :)

joshtriplett:
Looks like one of the Travis builds spuriously failed.  Retrying...

joshtriplett:
OK, I updated the test to add a cleanup that unsets the environment variables.

tests/repo/discover.c
joshtriplett: Done.

carlosmn:
This is very different from how the rest of the library operates. This would be the first place where we ourselves act on git environment variables so we need to be careful in how we introduce this kind of thing.

We do act on `HOME` and `XDG_CONFIG_HOME` but those are mandated by the git config rules and we only use them if the application hasn't set their own paths to use. This would be a wholly different way of acting.

This does not act on the `GIT_DIR` rules as I understand them to work. AFAIK when git finds `GIT_DIR`, it's supposed to set the git-dir path, not a base directory from which to search. This function could end up choosing a repository higher up the path while git would error out.

By defaulting to `"."` if there is no `GIT_DIR`, we use the current working directory as the base to search from, which is something we consciously avoid doing anywhere in the library since we can't trust the cwd is relevant for any particular directory.

Why stop at these variables? Why not look at `GIT_WORK_TREE`, `GIT_INDEX_FILE` or `GIT_OBJECT_DIRECTORY` if the aim is to behave as git would? As-is, this feels unfortunately a lot like a function for a specific usage of an application, rather than something that applies generally.

joshtriplett:
@carlosmn Building a command-line application similar to those provided by git itself seems like one of the primary use cases of libgit2.  Such applications should, by default, act on the standard git environment variables the same way git does, so that they fit in.  I absolutely do agree that libgit2 should never force the use of these environment variables, as that would break other use cases; however, it seems appropriate to me to provide a function that handles one of the most common use cases correctly, to avoid having to reimplement that functionality (potentially incorrectly) in each client.  Right now, it's easier to ignore variables like $GIT_CEILING_DIRECTORIES than to implement support for them; I'd like to make it easier to use them than to ignore them, so that command-line tools built on libgit2 will likely respect them.

Anyone who wants to ignore all of these variables can still call `git_repository_open_ext` or `git_repository_discover` directly; those code paths should *never* look at these environment variables.  That's why I added a separate function that explicitly does.

> This does not act on the GIT_DIR rules as I understand them to work. AFAIK when git finds GIT_DIR, it's supposed to set the git-dir path, not a base directory from which to search. This function could end up choosing a repository higher up the path while git would error out.

Good catch; I can easily fix that.  I can just pass GIT_REPOSITORY_OPEN_NO_SEARCH if $GIT_DIR was set.  I'll update the patch and the tests to fix that.

> By defaulting to "." if there is no GIT_DIR, we use the current working directory as the base to search from, which is something we consciously avoid doing anywhere in the library since we can't trust the cwd is relevant for any particular directory.

This function is explicitly supposed to be "open the default repository that a git command-line tool would open".  So, it should match git's behavior of searching from the current directory up to the ceiling dirs.  I agree that other functions in the library should *not* care about the current directory (though it's still possible to pass them relative paths), but in this case...

> Why stop at these variables? Why not look at GIT_WORK_TREE, GIT_INDEX_FILE or GIT_OBJECT_DIRECTORY if the aim is to behave as git would?

I'd like to do so.  $GIT_WORK_TREE and $GIT_INDEX_FILE seem trivial to handle, actually.  I should change this function to `git_repository_open_default` instead of making it a variant of discover.  Then, I can set the worktree and index before returning the repository.  As far as I can tell, libgit2 doesn't provide any support for separated object directories, but if it ever does, $GIT_OBJECT_DIRECTORY should be similarly easy to support.

> As-is, this feels unfortunately a lot like a function for a specific usage of an application, rather than something that applies generally.

This is intended to be the "give me the repository a git command-line tool would open" function; it should be general across all tools that behave like those git itself ships, which seems like a sufficiently common use case for libgit2.  To the extent that this *isn't* sufficiently general, I can and should fix that. :)

vmg:
I agree with @joshtriplett's rationale. As long as libgit2 doesn't take into account the environment variables by default, adding a helper API that mimic's Git's convention seems useful. The alternative is making many users of the library re-write the exact same logic. That's shitty for the users.

@ethomson care to weight in?

ethomson:
I'm mostly in agreement.  I think that this is a useful addition to the library, provided it has git-like semantics (and it seems like we're all in agreement that it should).

> This function is explicitly supposed to be "open the default repository that a git command-line tool would open".

I don't like calling this the "default repository", it makes it sound like there _is_ a default directory, when this is really the repository based on the current working directory.  This may not be a significant distinction if you're writing a CLI, but the concept of a "default repository" doesn't make much sense to anybody who's not.

`git_repository_open_current`?  `git_repository_open_from_cwd`?

> As far as I can tell, libgit2 doesn't provide any support for separated object directories, but if it ever does, $GIT_OBJECT_DIRECTORY should be similarly easy to support.

We're very flexible about object databases and this should be quite supportable with the existing code.

carlosmn:
>This is intended to be the "give me the repository a git command-line tool would open" function

Sure, I do agree that's a useful functionality. But that's not what the function does. It only covers one small aspect. If you want to do that, you'd still need to reimplement the logic even with this function available. Which is why I said that it felt like it came out of a specific use-case.

Name-wise, I'd rather go with something like `git_repository_open_fromenvironment()` or similar, since it's not just the "cwd"; after all, that bit is trivial by passing `"."` as a path to `git_repository_open()`, it's all the other stuff.

I would still prefer it if it accepted the path to take as the basis instead of assuming you would always want the cwd (you can accept `NULL` as defaulting to `getcwd()`). The purpose of the library is not to replicate what git has done, that tool already exists, but to allow you to be in control of what happens in a git repository. Even if a common use-case is to behave like git, some of the environment variables would apply just as well to something which does want to access multiple repositories.

joshtriplett:
I'll update the function to take more of the git environment variables into account, as well as the correct `$GIT_DIR` handling, and rename it to `git_repository_open_env`.

joshtriplett:
Pushing a WIP version for review; I still need to write test cases. I've added support for all the standard git environment variables: `$GIT_DIR`, `$GIT_DISCOVERY_ACROSS_FILESYSTEM`, `$GIT_CEILING_DIRECTORIES`, `$GIT_INDEX_FILE`, `$GIT_NAMESPACE`, `$GIT_OBJECT_DIRECTORY`, and `$GIT_ALTERNATE_OBJECT_DIRECTORIES`.  In the future, when [libgit2 gets worktree support](https://github.com/libgit2/libgit2/pull/3436), `git_repository_open_env` will also respect `$GIT_WORK_TREE` and `$GIT_COMMON_DIR`; until then, `git_repository_open_env` errors out if either of those variables is set, to avoid surprises.  I've also added a `start_path` parameter, with NULL meaning "use `$GIT_DIR` or search from "." if not set.

linquize:
Do you want libgit2 to respect any environment variables?

joshtriplett:
Updated with extensive testcases, covering all the supported environment variables.

In the course of developing these test cases, I found a bug in the existing ceiling_dirs handling in `find_repo`, which led to pull request #3727.  I based this branch on that one, as the tests need the same fix.

joshtriplett:
Any thoughts on this approach, either on the API or on this implementation of it?

src/repository.c
```diff
@@ -580,6 +591,173 @@ int git_repository_discover(
...
+	}
+
+	error = git_repository_open_ext(&repo, start_path, flags, ceiling_dirs);
+	if (error < 0)
+		goto error;
+
+	if (odb)
+		git_repository_set_odb(repo, odb);
+
+	if (git_buf_is_allocated(&alts_buf)) {
```
ethomson: Not sure that I understand this construct.  `git_buf_size(&alts_buf) > 0` might make more sense?  But I think the best option is to move the `git__gitenv` down here to keep the same pattern as the above variables.

tests/clar_libgit2.c
```diff
@@ -158,6 +158,18 @@ int cl_rename(const char *source, const char *dest)
...
+int cl_setenv_printf(const char *name, const char *fmt, ...)
+{
+	va_list args;
+	git_buf buf = GIT_BUF_INIT;
+
+	va_start(args, fmt);
+	cl_git_pass(git_buf_vprintf(&buf, fmt, args));
+	va_end(args);
+
+	return cl_setenv(name, git_buf_cstr(&buf));
```
ethomson: I'm not sure that this helper method is necessary, because it makes things like the fact that it's leaking `buf` a bit more subtle.

ethomson:
I'm not in favor of the name `git_repository_open_env`, it makes it sound like it's opening the environment, rather than opening the repository while _respecting_ the environment.

Something like `git_repository_open_with_env` or `open_respecting_env` would be better, but is sort of awkward still.

Why can't we simply make `GIT_REPOSITORY_OPEN_RESPECT_ENVIRONMENT` (or something perhaps less wordy) an option to `open_ext`?

src/repository.c
joshtriplett: `git_buf_is_allocated` checks whether the buffer has been initialized.

In this function, I kept all the `git__gitenv` calls and associated errors together and above the `git_repository_open_ext`, so that any error obtaining the environment variable would fail early and not have to clean up the open repository.  However, setting up alternatives requires the repository.

I can move the call, though; the order of error handling doesn't really matter.

tests/clar_libgit2.c
joshtriplett: Good catch. I'd argue that's a good reason to *have* the helper method, because we can fix any bugs in one place and avoid duplicating the same code pattern repeatedly.  The new test suite bits would get significantly longer if they have to duplicate this logic many times over.

I'd really like to fix this bug and keep the method.

joshtriplett:
> Why can't we simply make `GIT_REPOSITORY_OPEN_RESPECT_ENVIRONMENT` (or something perhaps less wordy) an option to `open_ext`?

Something like this?

```c
git_repository_open_ext(&repo, NULL, GIT_REPOSITORY_OPEN_RESPECT_ENV, NULL);
```

I could do that, but I think it'd complicate `git_repository_open_ext` with an almost completely independent chunk of logic.  And it doesn't make sense to combine this flag with any of the existing three `GIT_REPOSITORY_OPEN_*` flags.  I think I'd end up implementing this by checking that one flag at the top of `git_repository_open_ext` and calling a separate helper function that looks like the implementation in this commit (including the call to `git_repository_open_ext` without that flag).

The intention was to make this new function the obvious call to make if you want to open the same repository that the `git` command-line tool would open.

I can certainly rename the function; `git_repository_open_with_env` seems reasonable, as does `git_repository_open_from_env`.  I don't really care what color the bikeshed is painted. :)

tests/clar_libgit2.c
ethomson: Yeah, you do use that a lot.  My inclination would have been to throw it in as a static function next to the test itself though instead of as part of the text harness.

src/repository.c
ethomson: Sorry, my point about `is_allocated` is that it's not _quite_ what you're interested in; it measures whether the library was the one to allocate a `git_buf` (as opposed to an external caller).  It doesn't measure whether a `git_buf` has contents in it.

Are you preferring this because you want to support environment variables that are an empty string?  If so this is effective but feels a bit fragile.

tests/clar_libgit2.c
joshtriplett: OK, I'll move it to the test. I'd put it here in case any other test found it useful, but it seems easy enough for a future test to move the function here if they want to use it.

ethomson:
No, I was suggesting adding `GIT_REPOSITORY_OPEN_FROM_ENVIRONMENT` as a [`git_repository_open_flag_t`](https://github.com/libgit2/libgit2/blob/master/include/git2/repository.h#L103), which would invoke this functionality when set (incompatible flags would be ignored).

This gives us flexibility in the future if we (for example) add some `GIT_REPOSITORY_OPEN_...` flag that should be used by both `open_ext` _and_ by this new function.  If we have a `git_repository_open_from_env` then we will have to break its compatibility to add a `flags` argument to cope with that new flag.

We could of course go ahead and take `flags` in your new function, but once we do, they have a pretty similar signature, and we could instead have only one function and have it just switch based on that flag.

Just a thought.

src/repository.c
joshtriplett: Supporting empty strings was part of it; I used `is_allocated` as a "has this been set by `git__getenv`" check.  It returns false for `GIT_BUF_INIT` and true for the value produced by `git__getenv`.  That doesn't seem fragile.

I think it'll be easiest to just move the `git__getenv` call.

joshtriplett:
@ethomson I had that exact set of three flags in mind when I mentioned that none of them made sense together with this new one.  But forward compatibility makes sense, sure.  It does make the common case a bit less clean, but it still seems workable.

I'll add the new flag, and document the (non-)interaction between it and the other flags.

ethomson:
Yeah, it's tough to know whether this is really worthwhile or not (how much are we really going to add to `repository_open`, after all)?

(I think a totally reasonable thing is just to change the existing function and the new function to be internal and make `open_ext` a switch to the two.  Don't know offhand if that really makes sense or not.)

ethomson:
I think this makes sense - @carlosmn do you like the way this turned out?

ethomson:
Okay, one last request - would you mind documenting this in the `CHANGELOG`?

joshtriplett:
@ethomson Will do.

joshtriplett:
@ethomson Done.

PR #3715, opened 2016-03-31
xprepare memleaks

pks-t:
Cherry-picked two upstream fixes for memory leaks in the xdiff code from Ramsay Jones.

### DIFF
--- a/src/xdiff/xprepare.c
+++ b/src/xdiff/xprepare.c
@@ -301,6 +301,7 @@ int xdl_prepare_env(mmfile_t *mf1, mmfile_t *mf2, xpparam_t const *xpp,
 
 		xdl_free_ctx(&xe->xdf2);
 		xdl_free_ctx(&xe->xdf1);
+		xdl_free_classifier(&cf);
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


carlosmn:
Thanks.

ethomson:
FWIW, you translated that patch incorrectly.  The patch to git.git was:

```
diff --git a/xdiff/xprepare.c b/xdiff/xprepare.c
index 63a22c6..5ffcf99 100644
--- a/xdiff/xprepare.c
+++ b/xdiff/xprepare.c
@@ -304,7 +304,7 @@ int xdl_prepare_env(mmfile_t *mf1, mmfile_t *mf2, xpparam_t const *xpp,
        return -1;
    }
 
-   if (!(xpp->flags & XDF_HISTOGRAM_DIFF))
+   if (XDF_DIFF_ALG(xpp->flags) != XDF_HISTOGRAM_DIFF)
        xdl_free_classifier(&cf);
 
    return 0;
```

pks-t:
@ethomson Oh, you are right. That's what one gets from manually applying the patch instead of just using the tools correctly. I've created a fixup PR in #3718. Thanks for noticing

PR #3716, opened 2016-03-31
Add a no-op size_t typedef for the doc parser

carlosmn:
Clang's documentation parser, which we use in our documentation system
does not report any comments for functions which use size_t as a type.

The root cause is buried somewhere in libclang but we can work around it
by defining the type ourselves. This typedef makes sure that libclang
sees it and that we do not change its size.

---

I'm not particularly happy with this, but I've already burnt a lot of time tracking down the root cause and a work-around would still be needed for released versions anyhow.

### DIFF
--- a/include/git2/common.h
+++ b/include/git2/common.h
@@ -29,6 +29,14 @@
 # include <inttypes.h>
 #endif
 
+#ifdef DOCURIUM
+/*
+ * This is so clang's doc parser acknowledges comments on functions
+ * with size_t parameters.
+ */
+typedef size_t size_t;
+#endif
+
 /** Declare a public function exported for application use. */
 #if __GNUC__ >= 4
 # define GIT_EXTERN(type) extern \


vmg:
@carlosmn: This empty typedef ought to break other compilers. Can you please hide it behind a define so that only applies when building documentation?

carlosmn:
That's what I'm trying to figure out with the PR. Note that `size_t` is already defined when we do this so it doesn't become an opaque type.

carlosmn:
This is now only done when `DOCURIUM` is set. It needs a corresponding update to an as-of-yet unreleased version of docurium.

include/git2/common.h
```diff
@@ -29,6 +29,14 @@
 # include <inttypes.h>
 #endif
 
+#ifdef DOCURIUM
+/*
+ * This is so clang's doc parser acknowledges comments on functions
+ * with size_t parameters.
+ */
+typedef size_t size_t;
```
ethomson: scream-sum_2209650k

PR #3717, opened 2016-03-31
Plug some leaks

ethomson:
hb5-1 bmp


ethomson:
Hrmf.  Looks like there are some _new_ leaks.  Let me rebase this on master; please hold.

Issue #3713, opened 2016-03-29
libgit2 workdir to index diff reporting renamed submodules out of nowhere


swisspol:
I have a repo with 6 submodules set up the exact same way. With the latest libgit2, the diff between workdir and index (which is clear) reports that 3 of them have been "renamed" with an extra `/` at the end of their path:

![image](https://cloud.githubusercontent.com/assets/627915/14094603/bcbbd604-f50c-11e5-9c9d-b2a75818cbf4.png)

This is observed on OS X 10.11.4 using my app [GitUp](http://gitup.co) in which I just updated libgit2 from e0ab1ca0288db57fabbd63dfb6a0cd27f3acf8b2 to e89abec69f114a4480431c74160558b86148af41.

@ethomson The major difference I immediately see is the major iterator refactor you did. Could this be related?

`git_diff_options` passed in to `git_diff_index_to_workdir()`:
```
(lldb) p *diffOptions
(git_diff_options) $1 = {
  version = 1
  flags = 33587224
  ignore_submodules = GIT_SUBMODULE_IGNORE_NONE
  pathspec = {
    strings = 0x0000000000000000
    count = 0
  }
  notify_cb = 0x0000000000000000
  progress_cb = 0x0000000000000000
  payload = 0x0000000000000000
  context_lines = 3
  interhunk_lines = 0
  id_abbrev = 0
  max_size = 8388608
  old_prefix = 0x0000000000000000 <no value available>
  new_prefix = 0x0000000000000000 <no value available>
}
```

`git_diff_delta` contents for `External/libPusher`:
```
(lldb) p *delta
(git_diff_delta) $1 = {
  status = GIT_DELTA_DELETED
  flags = 0
  similarity = 0
  nfiles = 1
  old_file = {
    id = (id = unsigned char [20] @ 0x00007fba9ffe5050)
    path = 0x00006210001d3960 "External/libPusher"
    size = 1088
    flags = 12
    mode = 57344
    ctime = 1442841017
    mtime = 1442841017
  }
  new_file = {
    id = (id = unsigned char [20] @ 0x00007fba9ffe5090)
    path = 0x00006210001d3960 "External/libPusher"
    size = 0
    flags = 4
    mode = 0
    ctime = 0
    mtime = 0
  }
}

(lldb) p *delta
(git_diff_delta) $2 = {
  status = GIT_DELTA_UNTRACKED
  flags = 0
  similarity = 0
  nfiles = 1
  old_file = {
[...]

PR #3719, opened 2016-03-31
WD iterator: properly identify submodules

ethomson:
In a quite obvious mistake:  when looking for submodules, we were searching case sensitively ordered vectors with the case insensitive search function and vice versa.


ethomson:
Fixes #3713

PR #3720, opened 2016-04-01
merge_driver: fix missing `goto done;`

pks-t:
The code initializing the merge driver registry accidentally
forgot a `goto done` in case of an error. Because of this the
next line, which registers the global shutdown callback for the
merge drivers, is only called when an error occured.

Fix this by adding the missing `goto done`. This fixes some
memory leaks when the global state is shut down.


ethomson:
<3  Nice catch.

PR #3721, opened 2016-04-01
Fix .gitignore pattern matching with globs and path delimiters

as-cii:
Refs: https://github.com/atom/atom/issues/10689

We are observing a weird behavior in Atom where some folders are being ignored even when `git status` doesn't. It seems like the problem is related to how libgit2 handles `.gitignore` pattern matching, so this PR is an attempt to fix it.

Previously, when encountering a glob pattern, we would skip any character until some of them start to match, in which case we return `0` (i.e. "matched") in `p_fnmatchx`. This seems inconsistent with how `git` handles the same scenario: for example, patterns like `**/_*/` gets interpreted as: "find any folder that starts with `_`"; conversely, libgit2 interprets them as: "find any folder that contains `_`".

This PR changes the way we perform the matching, so that when encountering a glob followed by a path delimiter, we enforce that the string to perform pattern matching against also starts with a path delimiter.

Specs pass locally, but this is my first contribution to libgit2, so I am not totally sure about the consequences of this change. Hopefully, however, this will be a good starting point for discussion and a first step towards the resolution. :sparkles: 

What do you think? Thanks! :bow:

### DIFF
--- a/src/fnmatch.c
+++ b/src/fnmatch.c
@@ -61,6 +61,7 @@ static int rangematch(const char *, char, int, char **);
 static int
 p_fnmatchx(const char *pattern, const char *string, int flags, size_t recurs)
 {
+		bool string_is_after_path_delimiter, match_path_delimiter;
 		const char *stringstart;
 		char *newp;
 		char c, test;
@@ -88,7 +89,7 @@ p_fnmatchx(const char *pattern, const char *string, int flags, size_t recurs)
 						break;
 				case '*':
 						c = *pattern;
-						bool pattern_path_delimiter_start = false;
+						match_path_delimiter = false;
 
 						/* Let '**' override PATHNAME match for this segment.
 						 * It will be restored if/when we recurse below.
@@ -99,7 +100,7 @@ p_fnmatchx(const char *pattern, const char *string, int flags, size_t recurs)
 								c = *++pattern;
 							if (c == '/') {
 								c = *++pattern;
-								pattern_path_delimiter_start = true;
+								match_path_delimiter = true;
 							}
 						}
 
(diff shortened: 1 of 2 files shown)


src/fnmatch.c
```diff
@@ -119,15 +122,18 @@ p_fnmatchx(const char *pattern, const char *string, int flags, size_t recurs)
...
 						/* General case, use recursion. */
 						while ((test = *string) != EOS) {
 								int e;
 
 								e = p_fnmatchx(pattern, string, recurs_flags, recurs);
-								if (e != FNM_NOMATCH)
-										return e;
-								if (test == '/' && (flags & FNM_PATHNAME))
-										break;
+								if (e != FNM_NOMATCH && !(pattern_path_delimiter_start && !string_path_delimiter_start)) {
```
as-cii: Thinking more about this, I guess we might also break early when a match is found but it doesn't satisfy the path delimiter constraint:

```c
if (e != FNM_NOMATCH) {
	/* Break if the string doesn't start with a path delimiter but the pattern does. */
	if (pattern_path_delimiter_start && !string_path_delimiter_start) {
		break;
	} else {
		return e;
	}
}
```

Thoughts? :thought_balloon:

carlosmn:
These new tests pass by adding `break` where we match the slash after multiple stars (though other patterns do fail), so I've a feeling there's a much simpler implementation that can be extracted here.

carlosmn:
FWIW, starting a pattern with `**/` is completely unnecessary, it has the same meaning as not putting anything at all.

But I've been trying to see if we can avoid threading the state so much between the two use-cases for this case statement, which are completely different in what they actually want to accomplish. I'm partial to (this variant)[https://github.com/libgit2/libgit2/compare/cmn/ignore-starstar] but I'm not sure if it really is that much better. @ethomson thoughts?

ethomson:
Nice.  I'm also partial to [that refactoring](https://github.com/libgit2/libgit2/compare/cmn/ignore-starstar).  @as-cii how do you feel about that?

as-cii:
Looks good to me, too. 👍 Just two small refinements:

1. Delete [this extra newline](https://github.com/libgit2/libgit2/compare/cmn/ignore-starstar#diff-810cfaa06186fe52a865b683f2644429R159).
2. Move the comma [in that line](https://github.com/libgit2/libgit2/compare/cmn/ignore-starstar#diff-56d5f0709bd9592e5f07ece82522cbc5R97) after `end`.

Closing this so that it can be superseded by @carlosmn's changes. Thank you both for taking a look and fixing this! 🙏

PR #2101, opened 2014-02-04
Add a warning API and use it for bad signatures

arrbee:
The idea here is to add an API for reporting warnings where data appears to be bad but recoverable. For example, having an invalid timestamp in a commit signature or finding a `.gitattributes` file that can't actually be read. In these cases, we are currently faced with either reporting an error and failing in a situation that we could continue from (and Git often would), or ignoring the problem without giving the caller any notice that some questionable data was encountered (and in some cases, returning partially incorrect data as a result - e.g. we have no way of representing a bad signature timestamp, so we would just use a fallback value).

The driver for this is that we've been encountering some cases where libgit2 is more strict that Git for handling some data and I wanted a better option that just ignoring data problems when they are seen.

**This is just as initial take.**  I expect that the API will need to change quite a bit before we accept it into libgit2. Currently, I wanted the warning callback to get a lot of data about what was happening, so I tried to shoehorn things into a generic API with a lot of parameters.

Thinking it over, I think there are a number of options to proceed:
1. Use something like I have here with a generic warning signature (probably with fewer parameters than I've specified) and make the warning handler look at the message to decide what to do.
2. Create warning specific structures that get filled in depending on the warning, along with adding the specific warning codes to `include/git2/sys/warning.h` so the callback can know what structure it is looking at.
3. Add some sort of warning/error context stack where we could push data frames (e.g. top frame is a repository object, next frame down is a commit object being loaded, next frame down is the signature object being filled in, final frame is the string being parsed), and then the callback gets the stack of data frames with the option to report the error as desired. This could be useful for errors, too. Obviously this would have to be done with thread-local data.

We'll also have to decide where the warning callback should reside and how it should be set. I've opted to make it a single global value right now. I think that's probably the only good option. I've also opted to set it via `git_libgit2_opts` but I could easily add a setter API directly to `include/git2/sys/warning.h` which would restore type-safety (i.e. I should probably just do that).

PR #3722, opened 2016-04-01
Warnings API redux

carlosmn:
This is another attempt at #2101 which incorporates the subclassing described in the comments. An application may choose to print to stdout/stderr or it may look into the subclass to show the warnings in a GUI or to provide its own format.

The commit signature parsing code has changed quite a bit since the original PR, so I've used CRLF as an example of how we'd use it.

### DIFF
--- a/include/git2/warning.h
+++ b/include/git2/warning.h
@@ -1,0 +1,80 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+#ifndef INCLUDE_git_warning_h__
+#define INCLUDE_git_warning_h__
+
+#include "common.h"
+#include "types.h"
+
+GIT_BEGIN_DECL
+
+/**
+ * The kind of warning that was generated.
+ */
+typedef enum {
+	/**
+	 * Sentinel value. Should never be used.
+	 */
+	GIT_GENERIC_NONE = 0,
+
+	/**
+	 * Generic warning. There is no extended information
+	 * available.
+	 */
+	GIT_WARNING_GENERIC,
+} git_warning_t;
+
+/**
+ * Base struct for warnings.
+
+ * These fields are always available for warnings.
+ */
+typedef struct {
+	/**
+	 * The kind of warning.
+	 */
+	git_warning_t type;
+
+	/**
+	 * The text for this warning
+	 */
+	const char *str;
+} git_warning;
+
+/**
+ * User-specified callback for warnings
+ *
+ * The warning object is owned by the library and may be deallocated
+ * on function return. Make a copy if you want to store the data for
+ * later processing. Do not attempt to free it.
+ *
+ * Note that this may be called concurrently from multiple threads.
+ *
+ * @param warning the current warning
+ * @param payload user-provided payload when registering
+ * @return 0 to continue, a negative number to stop processing
+ */
+typedef int (*git_warning_cb)(git_warning *warning, void *payload);
+
+/**
+ * Set the warning callback
+ *
+ * This sets the global warning callback which be called in places
+ * where issues were found which might be of interest to a user but
+ * would not cause an error to be returned.
+ *
+ * This function does not perform locking. Do not call it
+ * concurrently.
+ *
+ * @param callback the function to call; pass `NULL` to unregister
+ * @param payload user-specified data to be passed to the callback
+ */
+GIT_EXTERN(int) git_warning_set_callback(git_warning_cb callback, void *payload);
+
+GIT_END_DECL
+
+#endif
--- a/include/git2/warning.h
+++ b/include/git2/warning.h
@@ -26,6 +26,11 @@ typedef enum {
 	 * available.
 	 */
 	GIT_WARNING_GENERIC,
+
+	/**
+	 * Warning related to line ending conversion.
+	 */
+	GIT_WARNING_CRLF,
 } git_warning_t;
 
 /**
@@ -44,6 +49,14 @@ typedef struct {
 	 */
 	const char *str;
 } git_warning;
+
+typedef struct {
+	/** The base struct */
+	git_warning parent;
+
+	/** The file this warning refers to */
+	const char *path;
+} git_warning_crlf;
 
 /**
  * User-specified callback for warnings
--- a/src/crlf.c
+++ b/src/crlf.c
@@ -16,6 +16,7 @@
 #include "filter.h"
 #include "buf_text.h"
 #include "repository.h"
+#include "warning.h"
 
 struct crlf_attrs {
 	int crlf_action;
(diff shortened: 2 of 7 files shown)


ethomson:
I like this - very simple and elegant.

One concern that I have is that - as a consumer - I may have to carry a lot of state to figure out what I should do in the default cases.  With this implementation, I register a single warning handler that will get called for all types of warnings.  I think there are two types of warnings that we will end up adding:

1. Things that really aren't all that important and git itself generally prints a message to the console and soldiers on.
2. Things that are really more like _errors_ but perhaps we are able to continue on, perhaps in a diminished capacity.

An example of #2 is a malformed commit, which we _could_ fill with crazy default values ("Unknown", perhaps) but we would not want to in the default case.

A user - if they want _just_ CRLF warnings (for example) would be likely to build out a default handler like:

```
static int values_callback(git_warning *warning, void *payload)
{
    if (warning->type == GIT_WARNING_CRLF) {
        /* i'm interested in this warning! */
        printf("%s\n", warning->message);
        return 0;
    }

    /* i'm not interested in this warning, just return the default */
    return 0;
}
```

Which would - inadvertently - continue on in the should-be-error cases (which would be errors if there were no warning configured.)

This is a sort of "you're holding it wrong" situation, but I think we're really _encouraging_ people to hold it wrong.  In this sort of situation, people would really need to know what the default should be for each type of warning.  There are a couple of ways to mitigate this:

1. We could let people register each type of warning.  This was what I was thinking that we would do, but obviously this complicates _our_ implementation and I like the simplicity here.

2. Make a return code of 0 mean "do the default", a negative return code a failure, and a positive return code a success.  This is sort of non-obvious, but has an interesting bit of simplicity.

3. _Give_ them the default (pass / fail).  This is probably the easiest, and most sensible, but still feels a little bit weird, in a way I can't really explain.

I lean a bit towards 1, but I'm pretty flexible.  I'm more interested in making sure that we don't trap the user here.

carlosmn:
>Things that really aren't all that important and git itself generally prints a message to the console and soldiers on.

This is what I'm targeting with this. You may want to print a message, or present a neat table of how badly formatted your files are.

>Things that are really more like errors but perhaps we are able to continue on, perhaps in a diminished capacity.

Recoverable errors feel like they're a different kind of beast, and I think separating the callbacks would help to hold it wrong less often.

>We could let people register each type of warning. This was what I was thinking that we would do, but obviously this complicates our implementation and I like the simplicity here.

Would this be a different registration per warning type? I can see this not being too bad as we know the size of the array at compile time.

>Make a return code of 0 mean "do the default", a negative return code a failure, and a positive return code a success. This is sort of non-obvious, but has an interesting bit of simplicity.

We can have `PASSTHROUGH` behaviour here so the user can explicitly tell us they didn't make any decisions.

>Give them the default (pass / fail). This is probably the easiest, and most sensible, but still feels a little bit weird, in a way I can't really explain.

The previous PR includes this, and I feel the same way :) We do however already provide something similar with the host certificate check, where we say whether we think it's valid, and then let the user decide (or not).

mikeando:
Since an application may have more than one repository open at a given time and the callback is global, it might be sensible to pass the repository that caused the problem as an argument.

Or the callback could be a per-repository setting.

I guess the repository can probably be deduced from the filename - but that seems a little finicky.

carlosmn:
There is no guarantee that we know what the repository is where we'd want to generate a warning, or that there even _is_ a repository so it ends up being the same issue with figuring out where the issue comes from.

If you do use multiple repositories, you already have to handle concurrent callbacks for certain operations, so you can extend whatever mechanism you already have.

ethomson:
One thing that makes this tricky is that generally you provide the `payload` during the method call.  Here, there is no such mechanism.

_Generally_ the warnings will get fired on the same thread that you started from, but that's not always true.  It's hard for me to imagine how you would reconcile which warnings were fired from which operation if each of them fired off multiple threads.  Unless we either arbitrate the warnings through the calling thread (which sounds crazy) or we stamp some sort of ID into the warning.  Perhaps the thread ID of the originating thread.  I dunno.  This is a little icky, but I think that we need to give a little bit more context to the callers.

carlosmn:
The only way to provide a per-call payload would be to carry it around anywhere, so I think we're realistically stuck with making the user provide a global payload.

>Generally the warnings will get fired on the same thread that you started from, but that's not always true. 

Do you have something particular operation in mind where we would raise warnings in a thread _we_ created as opposed to the caller?

>Unless we either arbitrate the warnings through the calling thread (which sounds crazy)

We have something similar for pack-objects progress (and I would expect a multi-threaded indexer or checkout would behave the same way) . The threads update the stats and it's the calling thread (which is otherwise sleeping) the one which calls the callback, since we do have the guarantee that we won't jump around threads (at least implicitly). As the places where we do perform multi-threaded operations on behalf of the user should be rather limited, and we would be performing status/progress updates by proxying through the calling thread anyhow. It doesn't seem like a big deal to also make sure warnings in these cases are raised by the calling thread.

>or we stamp some sort of ID into the warning. Perhaps the thread ID of the originating thread

Originating as the worker thread that generated the warning when proxying? I suppose we could add this (though iirc thread IDs are tricky to get the right size). But the way I did expect the (presumably tiny amount of) callers which actually do multiple operations concurrently is to set either thread-local data to know which thread is dealing with which repository or attach some global hash table associating a task id (for thread-hopping workpool systems) with the repository.

I would like to provide some more context, if we can have it, but if we only have it sometimes, I'm not sure it'd help too much since a caller would have to handle the case without the information already.

include/git2/warning.h
```diff
@@ -0,0 +1,93 @@
...
+GIT_BEGIN_DECL
+
+/**
+ * The kind of warning that was generated.
+ */
+typedef enum {
+	/**
+	 * Sentinel value. Should never be used.
+	 */
+	GIT_GENERIC_NONE = 0,
```
pks-t: Shouldn't this rather be `GIT_WARNING_NONE`?

include/git2/warning.h
```diff
@@ -0,0 +1,93 @@
...
+	 * Generic warning. There is no extended information
+	 * available.
+	 */
+	GIT_WARNING_GENERIC,
+
+	/**
+	 * Warning related to line ending conversion.
+	 */
+	GIT_WARNING_CRLF,
+} git_warning_t;
```
pks-t: `git_warning_type` would be a bit more obvious

src/crlf.c
```diff
@@ -147,11 +150,28 @@ static int crlf_apply_to_odb(
...
+					GITERR_FILTER, lfwarning, git_filter_source_path(src));
 				return -1;
 			case GIT_SAFE_CRLF_WARN:
-				/* TODO: issue warning when warning API is available */;
+			{
+				int ret;
+				git_warning_crlf warning;
+				git_buf buf = GIT_BUF_INIT;
+
+				if (git_buf_printf(&buf, lfwarning, git_filter_source_path(src)) < 0)
```
pks-t: I bet this kind of stuff would come up quite a lot. So a function to build a warning with a type and a format string would be nice, as well as one cleaning up the warning

tests/core/warning.c
```diff
@@ -0,0 +1,49 @@
...
+void test_core_warning__zero_on_unset(void)
+{
+	cl_git_pass(git_warning__raise(&g_warning));
+}
+
+static int values_callback(git_warning *warning, void *payload)
+{
+	cl_assert_equal_p(&g_warning, warning);
+	cl_assert_equal_p(&g_dummy_payload, payload);
+
```
pks-t: Shouldn't you set a static value here to assert in the caller that the callback was in fact invoked?

PR #3727, opened 2016-04-04
Fix repository discovery with ceiling_dirs at current directory

joshtriplett:
git only checks ceiling directories when its search ascends to a parent
directory.  A ceiling directory matching the starting directory will not
prevent git from finding a repository in the starting directory or a
parent directory.  libgit2 handled the former case correctly, but
differed from git in the latter case: given a ceiling directory matching
the starting directory, but no repository at the starting directory,
libgit2 would stop the search at that point rather than finding a
repository in a parent directory.

Test case using git command-line tools:

/tmp$ git init x
Initialized empty Git repository in /tmp/x/.git/
/tmp$ cd x/
/tmp/x$ mkdir subdir
/tmp/x$ cd subdir/
/tmp/x/subdir$ GIT_CEILING_DIRECTORIES=/tmp/x git rev-parse --git-dir
fatal: Not a git repository (or any of the parent directories): .git
/tmp/x/subdir$ GIT_CEILING_DIRECTORIES=/tmp/x/subdir git rev-parse --git-dir
/tmp/x/.git

Fix the testsuite to test this case (in one case fixing a test that
depended on the current behavior), and then fix find_repo to handle this
case correctly.

In the process, simplify and document the logic in find_repo():
- Separate the concepts of "currently checking a .git directory" and
  "number of iterations left before going further counts as a search"
  into two separate variables, in_dot_git and min_iterations.
- Move the logic to handle in_dot_git and append /.git to the top of the
  loop.
- Only search ceiling_dirs and find ceiling_offset after running out of
  min_iterations; since ceiling_offset only tracks the longest matching
  ceiling directory, if ceiling_dirs contained both the current
  directory and a parent directory, this change makes find_repo stop the
  search at the parent directory.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -359,7 +359,8 @@ static int find_repo(
 	git_buf path = GIT_BUF_INIT;
 	struct stat st;
 	dev_t initial_device = 0;
-	bool try_with_dot_git = ((flags & GIT_REPOSITORY_OPEN_BARE) != 0);
+	int min_iterations;
+	bool in_dot_git;
 	int ceiling_offset;
 
 	git_buf_free(repo_path);
@@ -367,13 +368,27 @@ static int find_repo(
 	if ((error = git_path_prettify(&path, start_path, NULL)) < 0)
 		return error;
 
-	ceiling_offset = find_ceiling_dir_offset(path.ptr, ceiling_dirs);
+	/* in_dot_git toggles each loop:
+	 * /a/b/c/.git, /a/b/c, /a/b/.git, /a/b, /a/.git, /a
+	 * With GIT_REPOSITORY_OPEN_BARE, we assume we started with /a/b/c.git
+	 * and don't append .git the first time through.
+	 * min_iterations indicates the number of iterations left before going
+	 * further counts as a search. */
+	if (flags & GIT_REPOSITORY_OPEN_BARE) {
+		in_dot_git = true;
+		min_iterations = 1;
+	} else {
+		in_dot_git = false;
+		min_iterations = 2;
+	}
 
-	if (!try_with_dot_git &&
-		(error = git_buf_joinpath(&path, path.ptr, DOT_GIT)) < 0)
-		return error;
+	while (!error && (min_iterations || !(path.ptr[ceiling_offset] == 0 ||
+					      (flags & GIT_REPOSITORY_OPEN_NO_SEARCH)))) {
+		if (!in_dot_git)
+			if ((error = git_buf_joinpath(&path, path.ptr, DOT_GIT)) < 0)
+				break;
+		in_dot_git = !in_dot_git;
 
-	while (!error && !git_buf_len(repo_path)) {
 		if (p_stat(path.ptr, &st) == 0) {
 			/* check that we have not crossed device boundaries */
 			if (initial_device == 0)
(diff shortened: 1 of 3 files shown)


joshtriplett:
I ran into this issue while writing and testing test cases for #3711.  I wrote some test cases for this corner case in `$GIT_CEILING_DIRECTORIES`, found that they failed, and then found I could reproduce the issue using `git_repository_open_ext` or `git_repository_discover` without using any of the new code.

The new behavior, and the new tests, match the way the git command-line tools handle `$GIT_CEILING_DIRECTORIES`.

joshtriplett:
Any thoughts on this bug? Do the test cases seem reasonable, as expressions of git's own behavior? Does the rewritten logic seem like an improvement in clarity?

tests/repo/discover.c
```diff
@@ -118,12 +118,22 @@ void test_repo_discover__0(void)
 	cl_git_fail(git_repository_discover(&found_path, ALTERNATE_MALFORMED_FOLDER3, 0, ceiling_dirs));
 	cl_assert_equal_i(GIT_ENOTFOUND, git_repository_discover(&found_path, ALTERNATE_NOT_FOUND_FOLDER, 0, ceiling_dirs));
 
+	append_ceiling_dir(&ceiling_dirs_buf, SUB_REPOSITORY_FOLDER_SUB);
+	ceiling_dirs = git_buf_cstr(&ceiling_dirs_buf);
+
+	//this must pass as ceiling_directories cannot prevent the current
+	//working directory to be checked
```
carlosmn: We avoid C++-style comments (it seems this file has at least one instance of this already, but let's not let it get out of control)

src/repository.c
```diff
@@ -359,21 +359,36 @@ static int find_repo(
...
 
 	git_buf_free(repo_path);
 
 	if ((error = git_path_prettify(&path, start_path, NULL)) < 0)
 		return error;
 
-	ceiling_offset = find_ceiling_dir_offset(path.ptr, ceiling_dirs);
+	/* in_dot_git toggles each loop:
+	 * /a/b/c/.git, /a/b/c, /a/b/.git, /a/b, /a/.git, /a
+	 * With GIT_REPOSITORY_OPEN_BARE, we assume we started with /a/b/c.git
```
carlosmn: Do we want to assume this? `_OPEN_BARE` is about avoiding loading things like workdir and config. Anything else should be the same.

tests/repo/discover.c
joshtriplett: No problem, I can change all of the new comments to `/* C style */`.

src/repository.c
joshtriplett: That's the defined existing behavior of `_OPEN_BARE`; it already causes the search to not append `/.git`.  I just documented that behavior.

joshtriplett:
Updated to use exclusively `/* C-style comments */`.

joshtriplett:
Following up on this; any issues with the new version of the pull request?

joshtriplett:
>  I think that this is overall a good change, but I'm finding the find_repo function a bit hard to process. This was true before your change, but since we're refactoring it, I wonder if we can clean it up a bit to reduce some of the cognitive load it puts on us.

:+1: It took me quite a while to work through its logic myself, so that sounds good to me.

> It seems like the logic here is correct, but cleaning this up a bit might make it easier to review, if that's something you're willing to do. Thanks!

Sure, I'd be happy to do so.  To some extent, I'd tried to reduce the amount of refactoring to the minimal amount needed to make the logic clearer, because I didn't want to make this more difficult to review.

If you don't mind, I'd like to make this a two-patch series, with the first patch fixing this logic bug and doing some minimal refactoring, and the second doing some more aggressive refactoring.  Does that sound reasonable?

ethomson:
Awesome, thank you.

> If you don't mind, I'd like to make this a two-patch series, with the first patch fixing this logic bug and doing some minimal refactoring, and the second doing some more aggressive refactoring. Does that sound reasonable?

Fine with me!  Thanks for tackling the refactoring.  😀

ethomson:
Included with #3711

PR #3728, opened 2016-04-05
Remove Makefile.embed

carlosmn:
This has not been a supported build mode for quite some time, and it
correspondingly hasn't worked to build the library for a long time. Get
rid of it, as the only build mode we support is though CMakek.


ethomson:
Whoa, I forgot this existed.

Issue #3334, opened 2015-07-28
An unified approach to cancellation


carlosmn:
Cancellation is important in long-running operations, which include network operations and checkout. These also involve multiple steps (particularly the network parts) so the cancellation needs to be expressed though all the layers. We currently have `git_remote_stop()` which gets looked at whenever the network code is parsing the data, but we don't make sure that we can respond in a reasonable time-frame. This also doesn't include the packing process during push.

Without cancellation, you must either wait for these operations to end or you must kill the thread or process. This can leave locks around and can stop system resources from getting released. So this is something we really should have.

Since this cancellation operation needs to cross some barriers, we would need this to be part of the options given to functions as a heap-allocated object, rather than an embedded struct in the options. Maybe something like

```C
/**
 * Allocate a new cancellation token
 */
int git_cancellation_new(git_cancellation **out);

/*
 * Trigger a cancellation
 */
int git_cancellation_cancel(git_cancellation *token);

/*
 * Returns whether cancellation was requested.
 *
 * This would be a private function.
 */
bool git_cancellation_requested(git_cancellation *token);

/*
 * Free a cancellation token.
 */
int git_cancellation_free(git_cancellation *token);
```

Many of the options structs would get a `git_cancellation *cancel_token` field, which the user allocates and assigns if they want to be able to cancel an operation. If a long-running function takes such a token, it must also pass its cancellation token to other long-running functions it calls.

This is modelled after .NET's `CancellationToken` which seems like a reasonable API. We do have a single object for both the caller and the calle to check rather than the two-object approach from .NET, but that shouldn't represent an issue for our use-case.

/cc @jamill @ethomson

PR #3729, opened 2016-04-05
[WIP] Operation cancellation support

carlosmn:
This is a possible implementation of #3334, but closer to the .NET design with an object that you pass into the operations which cannot be used to request cancellation.

This needs some naming bikeshedding and possibly more exact docs and a PoC implementation in e.g. the network or checkout operations.

It also needs a well-crafted history and commit messages.


include/git2/cancellable.h
```diff
@@ -0,0 +1,90 @@
...
+ */
+typedef int (*git_cancellable_cb)(git_cancellable *cancellable, void *payload);
+
+/**
+ * Create a cancellable source
+ *
+ * Create a new cancellable source. You can pass its token to functions which
+ * accept them and cancel them.
+ *
+ * @param out ponter in which to store the cancellable source
```
Therzok: Nit: typo pointer.

Therzok:
Looks good to me!

src/cancellable.c
```diff
@@ -0,0 +1,106 @@
...
+	/* Should we have a memory barrier here? a full lock? */
+	return c->cancelled;
+}
+
+int git_cancellable_register(git_cancellable *c, git_cancellable_cb cb, void *payload)
+{
+	registration *reg = git_array_alloc(c->registrations);
+	GITERR_CHECK_ALLOC(reg);
+
+	reg->cb = cb;
```
Therzok: This should be executed immediately if it's already been canceled.

Therzok:
Would this proposal include timeout support to `git_cancellable`?

src/cancellable.c
carlosmn: Ah yes, we can just skip it in that case.

carlosmn:
Adding built-in support for timeouts is something I've thought about and can be added later if we want to. But the primary/first user of this would be libgit2sharp, and we can piggy-back off of its `CancellationToken` so a timeout cancellation token there would make us cancel.

So we would only need it for C itself or some other language which doesn't have an equivalent.

carlosmn:
I've updated the name and removed the distinction between token and source.

ethomson:
Closing as stale.

PR #3730, opened 2016-04-06
tests: fix core/stream test when built with openssl off

andhe:
When passing -DUSE_OPENSSL:BOOL=OFF to cmake the testsuite will
fail with the following error:

core::stream::register_tls [/tmp/libgit2/tests/core/stream.c:40]
  Function call failed: (error)
  error -1 - <no message>

Fix test to assume failure for tls when built without openssl.
While at it also fix GIT_WIN32 cpp to check if it's defined
or not.


ethomson:
Looks good, thanks!

PR #3735, opened 2016-04-11
Fix reset tests with strict object creation

### LINKED ISSUE #3726: reset::hard::resetting_reverts_unmerged depends on state from previous tests
Running `./libgit2_clar -v -sreset::hard::resetting_reverts_unmerged` , or the full reset suite with `./libgit2_clar -v -sreset`, produces the following failure:

```
  1) Failure:
reset::hard::resetting_reverts_unmerged [/home/josh/src/libgit2/tests/reset/hard.c:115]
  Function call failed: (git_index_add(index, &entry))
  error -1 - Object not found - cannot read header for (6bb0d9f700543ba3d318ba7075fc3bd696b4287)
```

This failure doesn't occur when running the full test suite.  I think something in this test depends on state set up by a previous test.

carlosmn:
We were leaving strict object creation off after refs tests, so we didn't notice the reset tests used fake ids.

Fixes #3726

### DIFF
--- a/tests/refs/create.c
+++ b/tests/refs/create.c
@@ -13,13 +13,14 @@ static git_repository *g_repo;
 void test_refs_create__initialize(void)
 {
    g_repo = cl_git_sandbox_init("testrepo");
+	cl_git_pass(git_libgit2_opts(GIT_OPT_ENABLE_STRICT_OBJECT_CREATION, 0));
 }
 
 void test_refs_create__cleanup(void)
 {
    cl_git_sandbox_cleanup();
 
-	cl_git_pass(git_libgit2_opts(GIT_OPT_ENABLE_STRICT_OBJECT_CREATION, 0));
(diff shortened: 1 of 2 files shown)


ethomson:
Sigh.  Oops.  This was quite my fault.

It looks like the `refs::create` tests didn't get the memo when we changed the default to enforce strict validation by default.  This was trying to turn it back to the default of _off_.

I think that we should not be running the `refs::create` tests with strict object creation off.  Instead, we should change `refs::create::oid_unknown_succeeds_by_default` to be something like `refs::create::oid_unknown_can_succeed_when_not_strict` and then change `refs::create::oid_unknown_fails_strict_mode` to be something like `refs::create::oid_unknown_fails_by_default`.

This is my mess, I'm happy to fix it up in a moment.

carlosmn:
Superseded by #3737

PR #3736, opened 2016-04-11
refs: provide a more general error message for dwim

carlosmn:
If we cannot dwim the input, set the error message to be explicit about
that. Otherwise we leave the error for the last failed lookup, which
can be rather unexpected as it mentions a remote when the user thought
they were trying to look up a branch.


johnhaley81:
Holy cow you're fast :)

carlosmn:
Sometimes a man sees an error message and is deeply embarrassed by it :)

PR #3742, opened 2016-04-14
[WIP] git_status_options: support non-recursive directories status (issue #3675)

mstrap:


Therzok:
👍 This looks awesome.

mstrap:
@ethomson the code is basically working, but do you think I got the right places to plug-in the non-recursive behavior? Maybe I should also write some more comments as the entire topic is fairly complex?

PR #3745, opened 2016-04-19
Improve star-star matching

carlosmn:
Add some tests and handling for a few situations where our star-star handling in gitignore patterns was behaving incorrectly.

This supersede #3721


ethomson:
Thanks @as-cii and @carlosmn !

as-cii:
✨

PR #3746, opened 2016-04-19
[WIP] Add a script to help CTest run clar suites in parallel

carlosmn:
This script asks the clar runner for its suites and then creates a CTest
file so ctest knows about them. We can use this script together wtih
CTest's '-j' option to run multiple test suites in parallel.

---

This should help especially on Windows. I just run this with `ctest -j8` on my machine and the test runtime went from 70s to 20s.

I think we would want to run this as a CMake rule if we can, so it knows where the python executable is.


ethomson:
I like where this is going!  👍

carlosmn:
Funnily enough, we fail to parse the suites in the script because of the leak detection. It looks like we're leaking the filters global registry, which I think we should indeed be freeing during library cleanup.

xuan-72:
In include/git2/common.h:

> @@ -59,6 +59,12 @@ typedef size_t size_t;
>  #define GIT_WIN32 1
>  #endif
>  
> +#if defined(__has_feature)
> +#if __has_feature(address_sanitizer)
> +#define GIT_ASAN 1

ethomson:
clar is learning how to parallelize in https://github.com/clar-test/clar

PR #3749, opened 2016-04-22
Allow creating copies of `git_reference` objects.

arthurschreiber:
I'd like to make use of this in Rugged. 😄

### DIFF
--- a/tests/refs/dup.c
+++ b/tests/refs/dup.c
@@ -1,0 +1,40 @@
+#include "clar_libgit2.h"
+#include "refs.h"
+
+static git_repository *g_repo;
+
+void test_refs_dup__initialize(void)
+{
+	g_repo = cl_git_sandbox_init("testrepo.git");
+}
+
+void test_refs_dup__cleanup(void)
+{
+	cl_git_sandbox_cleanup();
+}
+
+void test_refs_dup__direct(void)
+{
+	git_reference *a, *b;
+
+	cl_git_pass(git_reference_lookup(&a, g_repo, "refs/heads/master"));
+	cl_git_pass(git_reference_dup(&b, a));
+
+	cl_assert(git_reference_cmp(a, b) == 0);
+
+	git_reference_free(b);
+	git_reference_free(a);
+}
+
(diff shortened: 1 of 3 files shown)


tests/refs/dup.c
```diff
@@ -0,0 +1,40 @@
...
+{
+	g_repo = cl_git_sandbox_init("testrepo.git");
+}
+
+void test_refs_dup__cleanup(void)
+{
+	cl_git_sandbox_cleanup();
+}
+
+void test_refs_dup(void)
```
ethomson: This test needs a proper name - `test_refs_dup` is the prefix for the `refs::dup` tests.  This test needs to be `test_refs_dup__direct` (for example).  (Clar is quite particular about naming.)

(You can run just the `refs::dup` tests to see that it only runs the one.)

tests/refs/dup.c
```diff
@@ -0,0 +1,40 @@
...
+{
+  git_reference *a, *b;
+
+	cl_git_pass(git_reference_lookup(&a, g_repo, "master"));
```
ethomson: This should be `refs/heads/master` instead of `master`.

tests/refs/dup.c
```diff
@@ -0,0 +1,40 @@
...
+
+	cl_assert(git_reference_cmp(a, b) == 0);
+
+	git_reference_free(b);
+	git_reference_free(a);
+}
+
+void test_refs_dup__symbolic(void)
+{
+  git_reference *a, *b;
```
ethomson: Indentation is off here.

tests/refs/dup.c
```diff
@@ -0,0 +1,40 @@
...
+}
+
+void test_refs_dup__cleanup(void)
+{
+	cl_git_sandbox_cleanup();
+}
+
+void test_refs_dup(void)
+{
+  git_reference *a, *b;
```
ethomson: The indentation here is a bit off.

ethomson:
👍

arthurschreiber:
Fixed everything up!

PR #3750, opened 2016-04-24
defines in version.h were not updated

mrh1997:
version 0.24.1 is still marked as 0.24.0.
This cannot be fixed for 0.24.1, but at least the master can be fixed...


carlosmn:
Indeed we forgot to update the patch version in the release, thanks for noticing.

However, the master branch does not get these version updates. The v0.24.X versions are made on a different branch. mater's version number is largely irrelevant, as you should not be installing this on your system.

PR #3757, opened 2016-04-26
Fix `git_commit_create` for an initial commit

johnhaley81:
When calling `git_commit_create` with an empty array of `parents` and `parent_count == 0` 
the call will segfault at https://github.com/libgit2/libgit2/blob/master/src/commit.c#L107 
when it's trying to compare `current_id` to a null parent oid. 

This just puts in a check to stop that segfault.

### DIFF
--- a/tests/commit/commit.c
+++ b/tests/commit/commit.c
@@ -84,13 +84,10 @@ void test_commit_commit__create_initial_commit(void)
 void test_commit_commit__create_initial_commit_parent_not_current(void)
 {
 	git_oid oid;
+	git_oid original_oid;
 	git_tree *tree;
 	git_commit *commit;
 	git_signature *s;
-	git_reference *origRef;
-	git_reference *origRefTarget;
-	git_reference *ref;
-	git_reference *refTarget;
 
 	git_oid_fromstr(&oid, "a65fedf39aefe402d3bb6e24df4d4f5fe4547750");
 	cl_git_pass(git_commit_lookup(&commit, _repo, &oid));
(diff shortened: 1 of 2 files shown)


johnhaley81:
Confirmed this fixes creating an initial commit in NodeGit as well.

ethomson:
If I can be a nag - it would be great to see a unit test here just so that we don't regress this in the future.  :)

johnhaley81:
@ethomson put in tests for successfully creating an initial commit and the failure case (where the segfault was happening). Anything else you'd like?

tests/commit/commit.c
```diff
@@ -51,6 +51,91 @@ void test_commit_commit__create_unexisting_update_ref(void)
...
+
+	git_oid_fromstr(&oid, "a65fedf39aefe402d3bb6e24df4d4f5fe4547750");
+	cl_git_pass(git_commit_lookup(&commit, _repo, &oid));
+
+	git_oid_fromstr(&oid, "944c0f6e4dfa41595e6eb3ceecdb14f50fe18162");
+	cl_git_pass(git_tree_lookup(&tree, _repo, &oid));
+
+	cl_git_pass(git_signature_now(&s, "alice", "alice@example.com"));
+
+	cl_git_pass(git_reference_lookup(&origRef, _repo, "HEAD"));
```
carlosmn: It looks like what you want here and in the other HEAD lookup is `git_reference_name_to_id()` which will reduce the boilerplate.

tests/commit/commit.c
```diff
@@ -51,6 +51,91 @@ void test_commit_commit__create_unexisting_update_ref(void)
...
+			git_reference_symbolic_target(origRef)
+		)
+	);
+	cl_git_pass(
+		git_reference_lookup(
+			&refTarget,
+			_repo,
+			git_reference_symbolic_target(ref)
+		)
+	);
```
carlosmn: `ref` and `origRef` are both the symbolic ref "HEAD", what you're checking here is that the commit creation function didn't switch branches.

carlosmn:
We use `snake_case` in the library.

tests/commit/commit.c
johnhaley81: You're right, I'll change it to grab the `git_oid` before and then compare what `HEAD` resolves to after the `git_commit_create` to ensure that nothing was changed on `HEAD`.

johnhaley81:
@carlosmn fixed up.

PR #3759, opened 2016-04-27
odb: avoid inflating the full delta to read the header

carlosmn:
When we read the header, we want to know the size and type of the
object. We're currently inflating the full delta in order to read the
first few bytes. This can mean hundreds of kB needlessly inflated for
large objects.

Instead use a packfile stream to read just enough so we can read the two
varints in the header and avoid inflating most of the delta.

---

For a particular repository which surfaced the issue, finding out the sizes for ~1300 blobs in a tree used to take ~13s whereas with this it takes under 50ms (and up to 20 of those are just the test suite starting up).

The one thing I'm not sure about is whether we can rely on zlib giving us those 16 bytes or if we should be paranoid and loop that read.

### DIFF
--- a/src/delta-apply.c
+++ b/src/delta-apply.c
@@ -49,6 +49,37 @@ int git__delta_read_header(
 	return 0;
 }
 
+#define DELTA_HEADER_BUFFER_LEN 16
+int git__delta_read_header_fromstream(size_t *base_sz, size_t *res_sz, git_packfile_stream *stream)
+{
+	static const size_t buffer_len = DELTA_HEADER_BUFFER_LEN;
+	unsigned char buffer[DELTA_HEADER_BUFFER_LEN];
+	const unsigned char *delta, *delta_end;
+	size_t len;
+	ssize_t read;
+
+	len = read = 0;
+	while (len < buffer_len) {
+		read = git_packfile_stream_read(stream, &buffer[len], buffer_len - len);
+
+		if (read == 0)
+			break;
+
+		if (read == GIT_EBUFS)
+			continue;
+
+		len += read;
+	}
+
+	delta = buffer;
+	delta_end = delta + len;
+	if ((hdr_sz(base_sz, &delta, delta_end) < 0) ||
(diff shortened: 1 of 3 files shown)


src/delta-apply.c
```diff
@@ -49,6 +49,26 @@ int git__delta_read_header(
 	return 0;
 }
 
+int git__delta_read_header_fromstream(size_t *base_sz, size_t *res_sz, git_packfile_stream *stream)
+{
+	/* The max size for a 32-bit varint is 5, we want two plus round up to power-of-two */
+	ssize_t read;
+	size_t buffer_len = 16;
+	unsigned char buffer[buffer_len];
```
vmg: Make `buffer_len` a static constant, otherwise this will be a VLA.

carlosmn: Done.

vmg:
@carlosmn: Still looking like a VLA for MSVC. Stupid compiler. I think we'll need to use a `define`. :(

carlosmn:
MSVC is hilarious. `#define` it is

ethomson:
> The one thing I'm not sure about is whether we can rely on zlib giving us those 16 bytes or if we should be paranoid and loop that read.

carlosmn:
zlib is always going to get at least 20 bytes for _its_ buffer. I'm not sure if you can assume that's going to be enough to get 10 bytes out of it, but I guess yeah, let's add a loop.

carlosmn:
Loop added. Should be good to go.

src/delta-apply.c
```diff
@@ -49,6 +49,39 @@ int git__delta_read_header(
...
+			continue;
+
+		len += read;
+	}
+
+	delta = buffer;
+	delta_end = delta + len;
+	if ((hdr_sz(base_sz, &delta, delta_end) < 0) ||
+	    (hdr_sz(res_sz, &delta, delta_end) < 0)) {
+		puts(":(");
```
ethomson: Can we make this an emoji instead?  ;)

carlosmn: :computer: :speech_balloon:

PR #3760, opened 2016-04-27
Fix return value of openssl_read (infinite loop)

cschlack:
openssl_read should return -1 in case of error.

SSL_read returns values <= 0 in case of error.

A return value of 0 can lead to an infinite loop, so the return value
of ssl_set_error will be returned if SSL_read is not successful (analog
to openssl_write).

### DIFF
--- a/src/openssl_stream.c
+++ b/src/openssl_stream.c
@@ -522,9 +522,8 @@ ssize_t openssl_read(git_stream *stream, void *data, size_t len)
 	openssl_stream *st = (openssl_stream *) stream;
 	int ret;
 
-	if ((ret = SSL_read(st->ssl, data, len)) <= 0) {
+	if ((ret = SSL_read(st->ssl, data, len)) <= 0)
 		return ssl_set_error(st->ssl, ret);
-	}
 
 	return ret;
 }
--- a/src/openssl_stream.c
+++ b/src/openssl_stream.c
@@ -522,8 +522,9 @@ ssize_t openssl_read(git_stream *stream, void *data, size_t len)
 	openssl_stream *st = (openssl_stream *) stream;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


src/openssl_stream.c
```diff
@@ -522,8 +522,9 @@ ssize_t openssl_read(git_stream *stream, void *data, size_t len)
 	openssl_stream *st = (openssl_stream *) stream;
 	int ret;
 
-	if ((ret = SSL_read(st->ssl, data, len)) <= 0)
-		ssl_set_error(st->ssl, ret);
+	if ((ret = SSL_read(st->ssl, data, len)) <= 0) {
+		return ssl_set_error(st->ssl, ret);
+	}
```
carlosmn: Please stick to the surrounding style: no braces here

carlosmn:
Nice catch, thanks!

src/openssl_stream.c
cschlack: Actually I did copy the style from openssl_write.

carlosmn: That's an odd one out; the style of the library is not to use branches, and _this function_ did not use braces

stmllr:
Looks like the build on Environment: GENERATOR=Visual Studio 11 Win64, ARCH=64 is broken on master: https://ci.appveyor.com/project/libgit2/libgit2/branch/master

ethomson:
> Looks like the build on Environment: GENERATOR=Visual Studio 11 Win64, ARCH=64 is broken on master: https://ci.appveyor.com/project/libgit2/libgit2/branch/master

The network tests will occasionally fail as networks and target systems are imperfect, as you might expect.

carlosmn:
It's not the network but threading tests. They occasionally fail as they don't make use of the promises the library itself makes.

ethomson:
Thanks!

PR #3763, opened 2016-04-28
Introduce `git_signature_from_buffer`

ethomson:
Allow users to construct a signature from the type of signature lines that actually appear in commits.  It may be natural for users to serialize author/committer information into this type of format, a helper method will provide some assistance for that.


spraints:
😍  I've wanted this a few times.

carlosmn:
I was trying to figure out if `_from_buffer` or `_frombuffer` would be more in line with our nomenclature but we don't seem to actually stick to one or the other.

PR #3767, opened 2016-05-02
Misc fixes

pks-t:
Some fixes hinted at by Coverity. I want to wait for one more build from Coverity due to 153fde5, which wasn't fixed in my previous run, but I hit the analysis quota. So for now this is up for discussion only until I hear back from Coverity tomorrow.

All fixes except fcd1b78 are of non-comsetic nature, that is they fix actual bugs. Most are rather obsocure though, to be honest.

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -1360,9 +1360,11 @@ fail:
 
 static bool should_remove_existing(checkout_data *data)
 {
-	int ignorecase = 0;
+	int ignorecase;
 
-	git_repository__cvar(&ignorecase, data->repo, GIT_CVAR_IGNORECASE);
+	if (git_repository__cvar(&ignorecase, data->repo, GIT_CVAR_IGNORECASE) < 0) {
+		ignorecase = 0;
+	}
 
 	return (ignorecase &&
 		(data->strategy & GIT_CHECKOUT_DONT_REMOVE_EXISTING) == 0);
--- a/src/delta-apply.c
+++ b/src/delta-apply.c
@@ -90,13 +90,13 @@ int git__delta_apply(
 			size_t off = 0, len = 0;
 
 			if (cmd & 0x01) off = *delta++;
-			if (cmd & 0x02) off |= *delta++ << 8;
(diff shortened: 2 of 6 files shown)


src/transports/smart_protocol.c
```diff
@@ -751,9 +751,10 @@ static int add_push_report_sideband_pkt(git_push *push, git_pkt_data *data_pkt,
 				git_buf_put(data_pkt_buf, line, line_len);
 			error = 0;
 			goto done;
-		}
-		else if (error < 0)
+		} else if (error) {
```
ethomson: When would we get a positive `error` here?  It's not immediately clear to me that `git_pkt_parse_line` would return a positive value and we would then want to fail because of it.

I worry a bit that static analysis has detected that this is possible, but (at a quick glance) it's not clear to me when this would happen, so it's not clear to me that it should be an error.  :/

pks-t: Oh, never mind this one. I've been confused by the analysis, as well, and first thought it was a false positive. I then went and checket the function and actually saw what I wanted to see in there I guess, so I misread two different locations and thought it obvious that positive values might be returned.

Will drop the patch. Thanks for double-checking.

pks-t:
I've stripped the commit pointed out by @ethomson, but included a new one. fe3057b simplifies some code in diff.c without changing functionality, but I'm actually at odds if the code was right at all as it seems to contradict https://github.com/libgit2/libgit2/blob/master/include/git2/diff.h#L147-L155. That is, we never mark the directory as UNTRACKED when there are no files contained on the other side.
I've tried to "fix" the code such that we set the directory as untracked when `!contains_oitem`, but that lead to a plethora of test failures. Comments?

pks-t:
Coverity verified the branch by the way, so I'd consider it good to go. I'd still be glad if one of the core devs would verify fe3057b, though.

ethomson:
This looks good to me.  I think that your changes to the diff code are correct.  My reading suggests that handling untracked directories (eg, where there's no oitem) is handled before this block.

Thanks for tackling these!

PR #3770, opened 2016-05-04
Add a method specifically for modifying trees

carlosmn:
The way we currently recommend modifying trees is via an index. This method has a couple of advantages: it's what any tool would do on their `git-commit` equivalent, and it's code that we already have and use. It does however mean doing a recursive read of the root tree, and that can get big. Modifying a file becomes more expensive as files are added to the tree regardless of where the updated file lives.

On a workdir, it's not so bad because the index is kept across operations but when doing this without the advantage of a user-provided workdir, the recursive read can become the primary cost of the update operation.

This PR adds a function which, given a baseline tree and a list of updates, reads just enough trees to perform the necessary updates.

---

This needs better history and some tests to make sure that performing modifications via the index and this new function has the same result, but I think the design of the function seems reasonable.

### DIFF
--- a/include/git2/tree.h
+++ b/include/git2/tree.h
@@ -418,6 +418,52 @@ GIT_EXTERN(int) git_tree_walk(
  */
 GIT_EXTERN(int) git_tree_dup(git_tree **out, git_tree *source);
 
+/**
+ * The kind of update to perform
+ */
+typedef enum {
+	/** Update or insert an entry at the specified path */
+	GIT_TREE_UPDATE_UPSERT,
+	/** Remove an entry from the specified path */
+	GIT_TREE_UPDATE_REMOVE,
+} git_tree_update_t;
+
+/**
+ * An action to perform during the update of a tree
+ */
+typedef struct {
+	/** Update action. If it's an removal, only the path is looked at */
+	git_tree_update_t action;
+	/** The entry's id */
+	git_oid id;
+	/** The filemode/kind of object */
+	git_filemode_t filemode;
+	/** The full path from the root tree */
+	const char *path;
+} git_tree_update;
+
+/**
+ * Create a tree based on another one with the specified modifications
+ *
+ * Given the `baseline` perform the changes described in the list of
+ * `updates` and create a new tree.
+ *
+ * This function is optimized for common file/directory addition, removal and
+ * replacement in trees. It is much more efficient than reading the tree into a
+ * `git_index` and modifying that, but in exchange it is not as flexible.
+ *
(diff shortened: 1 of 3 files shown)


include/git2/tree.h
```diff
@@ -418,6 +418,43 @@ GIT_EXTERN(int) git_tree_walk(
  */
 GIT_EXTERN(int) git_tree_dup(git_tree **out, git_tree *source);
 
+/**
+ * The kind of update to perform
+ */
+typedef enum {
+	/** Update or inert an entry at the specified path */
```
arthurschreiber: Typo: `insert`

carlosmn:
This seems ready now, @ethomson; I'd still like to actually benchmark some of those problematic cases, but it should be good for review.

carlosmn:
I used the replace-blob test case to run a benchmark on a repository with a particularly troublesome layout.

```
Started
index took 6.780517 seconds
updater took 0.000084 seconds
.
```

so it's not bad. Avoiding reading most of the data makes it instant instead of timing out on a busy server.

src/tree.c
```diff
@@ -281,18 +281,24 @@ const git_oid *git_tree_entry_id(const git_tree_entry *entry)
 	return entry->oid;
 }
 
-git_otype git_tree_entry_type(const git_tree_entry *entry)
+static git_otype filemode2type(git_filemode_t filemode)
```
ethomson: We have `git_object__type_from_filemode` which we may be able to use.  (I didn't think to spelunk for other users when I added it, oops.)

carlosmn: Fixed.

include/git2/tree.h
```diff
@@ -418,6 +418,49 @@ GIT_EXTERN(int) git_tree_walk(
...
+
+/**
+ * Create a tree based on another one with the specified modifications
+ *
+ * Given the `baseline` perform the changes described in the list of
+ * `updates` and create a new tree.
+ *
+ * This function is optimized for common file/directory addition, removal and
+ * replacement in trees. It is much more efficient than reading the tree into a
+ * `git_index` and modifying htat, but in exchange it is not as flexible.
```
arthurschreiber: Typo: `htat`

include/git2/tree.h
```diff
@@ -418,6 +418,49 @@ GIT_EXTERN(int) git_tree_walk(
...
+ *
+ * @param out id of the new tree
+ * @param repo the repository in which to creat the tree, must be the
```
arthurschreiber: Typo: `creat`

PR #3773, opened 2016-05-05
Fix unused variable 'message' warning

lucasderraugh:


ethomson:
Thanks!  We try very hard to keep variable declarations at the beginning of blocks, because we're hopelessly antiquated - or, more accurately - some of the compilers we have to support are.  Although it's quite reasonable to expect that clang is quite modern (and that you're surely using clang to compile stransport code) we do try to keep these conventions consistent across the library.

Would you mind keeping the declaration at the top of the block, and perhaps denoting that it's unused in the `#else` block?  We have the helpful `GIT_UNUSED` macro for this.

Thanks again!

lucasderraugh:
Sorry about that, should have looked at your conventions. Should be good now.

ethomson:
Delightful!  Thank you for noticing this and fixing it.

PR #3781, opened 2016-05-10
[WIP] Allow building as a framework on OSX

tiennou:
Use `cmake -DBUILD_FRAMEWORK=1 -G Xcode` to generate the project.

This is a first, rough pass at making libgit2 package itself on OSX using only CMake, so I haven't actually tried to use the resulting framework (it looks like there are a few things to fix in OCGit for that to go smoothly).

libgit2/objective-git#542


tiennou:
I think handling the dependencies is going to be **FUN**. If someone whose CMake-fu is up-to-par, I'm looking for a way to grab the path to a linked library, because they need to be copied them inside the framework.

tiennou:
Closing because It's Too Old And Build Systems Are Too Old.

Issue #3744, opened 2016-04-19
Better support for detecting libssh2 


mtangoo:
This might apply for those who love to use OpenSSL on Windows, but I will leave that for later discussion. As Introduction is about a month now I have been fighting with libgit2 on how to link libssh2 so that I have support for SSH. Today I finally solved it and wanted to present what I did and if it can be integrated to libgit2 cmake system.

I found a cmake module [here](https://github.com/datastax/cpp-driver/blob/master/cmake/modules/FindLIBSSH2.cmake)(Thanks to the author). After droping it into libgit2 cmake modules folder and added `SET(LIBSSH2_ROOT  "MY_PATH_TO_SSH_FOLDER")` and changed CMakeList.txt

```
IF (USE_SSH)
	PKG_CHECK_MODULES(LIBSSH2 libssh2)
ENDIF()
```

to

```
IF (USE_SSH)
	FIND_PACKAGE(LIBSSH)
ENDIF()
```
I had SSH2 Support. I wanted to know if the solution is acceptable to libgit2 standard and if I should make a PR to this.

TIA

carlosmn:
On unix, we would much rather depend on `pkg-config` which is the standard way to figure out where libraries are located. This way it's the library authors themselves who say how to link to things.

On Windows this is a different story and AFAIK there is no good way to dealing with this. I suppose we could see about taking in that CMake module for Windows, but it seems to only search for the static library, which is a limitation which I'm not sure we want to have for our build.

mtangoo:
I would like to know what is missing in that file so that I will fix it as I get time and eventually do PR.
As of Linux, I think the best way is to check if `LIBSSH2_ROOT` is set then use it to search the library and if not (the default) search using `package-config`

ethomson:
I think that dynamically loading is the best way to solve this problem; closing in favor of that.

mtangoo:
Can you explain what you mean? Afaik the issue is detection of libssh2 during compile time not runtime

carlosmn:
Detecting at compile-time is one issue, but it's not the big issue. In order for libgit2 to be able to use libssh2 effectively on Windows, it has to be loaded at run-time when the user puts the dll next to us. Anything else makes the distribution of libgit2 for Windows a nightmare.

mtangoo:
You're both right but the ticket was about detecting libssh2 on compile time.

Issue #3762, opened 2016-04-27
libssh2 support as a runtime option instead of compile-time?


tkelman:
I have a use case where I'd like to make ssh support optional, where in a default configuration it is not available but if the libssh2 library becomes available at runtime I'd like libgit2 to be able to use it without having to recompile. Assuming I could ensure the versioning and build configuration of the runtime-available libssh2 matches what gets used at build time for libgit2, would this be at all technically feasible? Would it be possible to change some of the conditional compilation into some form of runtime check and associated API for initialization?

ethomson:
Yes, this is definitely the avenue that we would like to pursue.  At the moment we're stuck in a terrible place where many consumers seem to want or expect SSH support, but building, releasing and maintaining libssh2 is _not_ something that we are prepared to do.  And asking the users of many of the bindings to recompile libgit2 is quite terrible.  (Most of them are either presented with a libgit2 binary, as with LibGit2Sharp, or their libgit2 is built for them, as with Rugged, and the end-user is not given any ability to tailor this behavior to even add the flags that indicate libssh2 support.)

My opinion is that using `dlopen` and `LoadLibraryEx` to resolve libssh2 is the superior solution here.  That way end-users of bindings, at least, can throw their own libssh2 down and we will load it, without having to recompile libgit2.

Now, this doesn't really fix _all_ the problems of distributing this to end-users, but it should (hopefully) help a bit.

tiennou:
I'd be willing to work on that (the *nix part at least :trollface:). I'm wondering how to handle the "function call" part though... Can CMake do weak-linking ?

tiennou:
Also, would it be acceptable to have a hard requirement for building with libssh2 or should the `GIT_SSH` tests be kept ?

tkelman:
> would it be acceptable to have a hard requirement for building with libssh2

I think the way this would work is if libgit2 is present at build time, use it and link against it in the current way. If it isn't present at build time, then build some new runtime plugin stubs that aren't activated by default, or have to try `dlopen`ing a libssh at runtime and checking if it's compatible somehow. Maybe the build-time-only version could be made obsolete eventually, but since it's working now there's no reason to mess with it too much just yet.

tiennou:
@tkelman Please have a look at the PR I opened ;-).

PR #3785, opened 2016-05-13
Weak-link against libssh2 and load it at runtime

tiennou:
I still need to check how weak-linking is done in Visual Studio.
The .dylib makes it OSX-specific at the moment.
And I don't know how you guys feel about dynamic-loading _some_ file called libssh2 out of the blue, but it sure makes me wary :wink:

#3744


src/transports/ssh.c
```diff
@@ -893,9 +913,24 @@ int git_transport_ssh_with_paths(git_transport **out, git_remote *owner, void *p
...
+void git_transport_ssh_global_shutdown(void);
+
 int git_transport_ssh_global_init(void)
 {
 #ifdef GIT_SSH
+#ifdef GIT_WIN32
+	const char *(*libssh2_version)(int) = NULL;
+#else
+	libssh2_handle = dlopen("libssh2.dylib", RTLD_LAZY);
+	const char *(*libssh2_version)(int) = dlsym(libssh2_handle, "libssh2_version");
```
mikeando: Can you get away without a `libssh2_version` function pointer if you use `RTLD_NOW | RTLD_GLOBAL` in the dlopen? (as outlined in the end of the answer http://stackoverflow.com/questions/20658809/dynamic-loading-and-weak-symbol-resolution )

src/transports/ssh.c
```diff
@@ -893,9 +913,24 @@ int git_transport_ssh_with_paths(git_transport **out, git_remote *owner, void *p
...
+#endif
+
+	// Check that libssh2's version is sufficient for us
+	if (!libssh2_version || !libssh2_version(LIBSSH2_VERSION_NUM)) {
+		return 0;
+	}
```
mikeando: Should failure to load the library be an error?
Should a mismatched version be an error?

tiennou: I do agree it's kinda unclear, but if we were to report an error here, that would make the library completely fail to initialize. Right now this stashes the handle away, and the actual error-returning check is done when we attempt to create a transport — mostly because handles are ref-counted, and one handle per library initialization Ought To Be Enough For Anybody.

tiennou: Note though that I'm unclear about which specific `LIBSSH2_VERSION_NUM` define we should use. Should we always expect the version of libssh2 we built against to be available (and use the define), or be conservative and provide a specific version. I'd prefer the latter, but I think that can wait until the actual dynamic linking thing looks good.

tiennou:
Also, #3762

src/transports/ssh.c
mikeando: Delaying the error message until first use seems reasonable. In that case I'd suggest stashing away an appropriate error message about what exactly failed during init. So the user can get something more useful than just ` "Cannot create SSH transport. SSH support unavailable"`. I'm guessing it will then be either something like "Cannot create SSH transport. SSH support unavailable. Failed loading libssh2.dylib" or "Cannot create SSH transport. SSH support unavailable.  libssh2.dylib is version XYZ which is not compatible with version PQR".

I'd also pull the checks out into their own function, so that we can do something nicer if libssh2 is statically linked.
(I assume that is planned to remain an option?). At the moment it looks like if its statically linked, it will still try to `dlopen` then `.dylib` and then report errors about not having ssh.

carlosmn:
Why do we want to weak-link _and_ use `dlsym` to look up the function to give us the version? The advantage of weak-linking is that we get to use our own stubs. But while weak-linking is useful, I don't think weak-linking is what we want to do in most cases. Weakly linking to the libssh2 libraries would allow us to use it when the program using libgit2 links against libssh2 itself, so _it_ has the choice of whether to use libssh2. This is something that happens at load time, so loading the library at run time should have no effect (at least on ELF). This would also be surprising behaviour since we would allow for two unrelated ways of getting to libssh2.

But that's not what we'll see in most uses, which go though a module in some other programming language or OS which does not support weak-linking the same way ELF or Mach-O do. That needs run-time loading of the library if we find it in the load path.

src/transports/ssh.c
```diff
@@ -893,9 +913,24 @@ int git_transport_ssh_with_paths(git_transport **out, git_remote *owner, void *p
...
 
+void git_transport_ssh_global_shutdown(void);
+
 int git_transport_ssh_global_init(void)
 {
 #ifdef GIT_SSH
+#ifdef GIT_WIN32
+	const char *(*libssh2_version)(int) = NULL;
+#else
+	libssh2_handle = dlopen("libssh2.dylib", RTLD_LAZY);
```
carlosmn: We do not want to load the latest version available for developing on this platform, we want to load a specific SOVERSION of this library for which we know what the function pointers look like.

src/transports/ssh.c
carlosmn: This check should not be necessary. As mentioned in the other comment, we want to load a specific SOVERSION and otherwise we'll fail to load the library.

tiennou:
Meh I'm overwhelmed now ;-). I'm happy to drop the weak-linking part — arguably because CMake doesn't really support it, and the build failed for a weird (and silent) reason because it didn't like the flags I forcibly passed.

Are you sure about loading a specific SOVERSION though ? Right now, if we were to `dlopen("libssh2.X.Y", ...)`, it wouldn't work for a more up-to-date version, while the generic "load and test" would still work.

Also, with weak-linking out of the picture, does that mean I'll have to declare a bunch of pointers to the functions we use in libssh2 and fill them in and use those instead so I don't get hard linker errors ?

And finally, @mikeando's point about static linking makes me nervous ;-). I'd say if someone is using a statically-linked libgit2 somewhere, it's up to them to set up their toolchain accordingly... Opinions ?

carlosmn:
>does that mean I'll have to declare a bunch of pointers to the functions we use in libssh2

Yes, which is why we haven't done it yet. It's a lot of work and the core team doesn't need it.

>Are you sure about loading a specific SOVERSION though ? Right now, if we were to dlopen("libssh2.X.Y", ...), it wouldn't work for a more up-to-date version, while the generic "load and test" would still work.

There is zero guarantee that it would work. A changed SOVERSION means anything could have changed. Maybe it's an added symbol, maybe they removed another one. We can try to load a few that we know we do support, since most changes to the SOVERSION won't be a big issue.

But the larger point is that we cannot depend on `.dylib` or `.so` being there, since those exist purely as a way for the linker to figure out which one to use. The resulting programs get linked to `libssh2.so.1` and if someone install the libssh2 package, they won't get the `.so` unless they install the development package, which they shouldn't be made to do. We would also not know which version of the ABI it actually links to, making its use unsafe. Note that this is SOVERSION 1 regardless of whether the library is 1.5 or 1.7 as the ABI is compatible.

tiennou:
>> does that mean I'll have to declare a bunch of pointers to the functions we use in libssh2
> Yes, which is why we haven't done it yet. It's a lot of work and the core team doesn't need it.

Thanks, at least this clears up what must be done.

> But the larger point is that we cannot depend on .dylib or .so being there

Wait, does that mean that there's no guarantee that `libssh2.dylib/so` exists and is a symlink to the real library ? I'm only aware of how OS X works here, so maybe *nix does it differently...

Right now I was under the impression that `dlopen` would either fail (meaning we can't do ssh at all), or succeed, (meaning we would be able to check via `libssh2_version` that it supports what we expect)...

carlosmn:
>Wait, does that mean that there's no guarantee that `libssh2.dylib/so` exists and is a symlink to the real library ? I'm only aware of how OS X works here, so maybe *nix does it differently...

That's right. This is the same for any kind of unix. You're probably used to how *homebrew* works by installing a single package which contains the whole thing. Most other package managers (the ones that call themselves that) tend to split packages between what you need to use a library versus what you need to compile against it. It's more about the userland than the actual OS.

But even if you were guaranteed to have it, there would still be no guarantee that it's the version we want. If you suddently install libssh2 v2.0.0 with a completely different API and SOVERSION 2, you'd likely have the `.dylib`/`.so` symlink point to the latest one. Loading that might make us crash if some function names are the same but the arguments are all different. If the API is completely different many projects would change the basename to indicate how different it is, but the more common case is fixing up types or changing the API slightly, where you'd bump the SOVERSION.

In libgit2 we do this with the SOVERSION so you can keep linking to `libgit2.so.23` even if the latest package is for v0.24.0 and newly-linked programs would link against that.

tiennou:
Oh thanks, now I see why the SOVERSION must be hardcoded 💡. Thanks for the thorough explanation !

tiennou:
So, I decided against flattening the libssh2 macro-functions because that was IMHO too messy and instead just redefined their underlying functions in our namespace. It means those macro-functions are still namespaced in libssh2 which is unfortunate — if that's really a bother, I'll import those #defines instead but this looked cleaner to me.

- It actually builds and passes the test suite locally. Travis just doesn't like me. I think.
- I'm not sure what should I do to grab a correct definition for `GIT_LIBSSH2_SOVERSION`. Should that be filled in at runtime, or is something static (as it is) sufficient ?
- I took a look at WIN32 but failed.

PR #3789, opened 2016-05-17
Support index v4

dturner-tw:
This PR includes code from Junio, which seems to be permitted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ v0.24 + 1
 
 ### Changes or improvements
 
+* Support for reading and writing git index v4 files
+
 ### API additions
 
 * `git_commit_create_buffer()` creates a commit and writes it into a
@@ -12,6 +14,9 @@ v0.24 + 1
   `git_blob_create_fromstream_commit()` allow you to create a blob by
   writing into a stream. Useful when you do not know the final size or
   want to copy the contents from another stream.
+
+* `git_index_version()` and `git_set_index_version()` to get and set
+  the index version
 
 ### API removals
 
--- a/src/index.c
+++ b/src/index.c
@@ -19,6 +19,7 @@
 #include "blob.h"
 #include "idxmap.h"
 #include "diff.h"
+#include "varint.h"
 
 #include "git2/odb.h"
 #include "git2/oid.h"
@@ -65,8 +66,11 @@ static int index_apply_to_wd_diff(git_index *index, int action, const git_strarr
 static const size_t INDEX_FOOTER_SIZE = GIT_OID_RAWSZ;
 static const size_t INDEX_HEADER_SIZE = 12;
 
-static const unsigned int INDEX_VERSION_NUMBER = 2;
+static const unsigned int INDEX_VERSION_NUMBER_DEFAULT = 2;
+static const unsigned int INDEX_VERSION_NUMBER_LB = 2;
 static const unsigned int INDEX_VERSION_NUMBER_EXT = 3;
+static const unsigned int INDEX_VERSION_NUMBER_COMP = 4;
+static const unsigned int INDEX_VERSION_NUMBER_UB = 4;
 
 static const unsigned int INDEX_HEADER_SIG = 0x44495243;
 static const char INDEX_EXT_TREECACHE_SIG[] = {'T', 'R', 'E', 'E'};
@@ -434,6 +438,7 @@ int git_index_open(git_index **index_out, const char *index_path)
 	index->entries_search = git_index_entry_srch;
 	index->entries_search_path = index_entry_srch_path;
 	index->reuc_search = reuc_srch;
+	index->version = INDEX_VERSION_NUMBER_DEFAULT;
 
 	if (index_path != NULL && (error = git_index_read(index, true)) < 0)
 		goto fail;
@@ -743,6 +748,26 @@ static int truncate_racily_clean(git_index *index)
 done:
 	git_diff_free(diff);
 	git_vector_free(&paths);
+	return 0;
+}
+
+unsigned git_index_version(git_index *index)
+{
+	assert(index);
+
+	return index->version;
+}
+
+int git_index_set_version(git_index *index, unsigned int version)
+{
+	assert(index);
+
+	if (version < INDEX_VERSION_NUMBER_LB ||
+		version > INDEX_VERSION_NUMBER_UB)
+		return -1;
+
+	index->version = version;
+
 	return 0;
 }
 
@@ -2261,12 +2286,15 @@ static size_t read_entry(
 	git_index_entry **out,
 	git_index *index,
 	const void *buffer,
-	size_t buffer_size)
+	size_t buffer_size,
+	const char **last)
 {
 	size_t path_length, entry_size;
 	const char *path_ptr;
 	struct entry_short source;
 	git_index_entry entry = {{0}};
+	bool compressed = index->version >= INDEX_VERSION_NUMBER_COMP;
+	char *tmp_path = NULL;
 
 	if (INDEX_FOOTER_SIZE + minimal_entry_size > buffer_size)
 		return 0;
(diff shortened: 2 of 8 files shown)


src/index.c
```diff
@@ -434,6 +437,7 @@ int git_index_open(git_index **index_out, const char *index_path)
 	index->entries_search = git_index_entry_srch;
 	index->entries_search_path = index_entry_srch_path;
 	index->reuc_search = reuc_srch;
+	index->version = 2;
```
mikeando: Should we be using `INDEX_VERSION_NUMBER` instead of magic number 2?

src/index.c
```diff
@@ -2517,15 +2565,32 @@ static bool is_index_extended(git_index *index)
...
 	char *path;
+	const char *path_start = entry->path;
+	size_t same_len = 0;
 
 	path_len = ((struct entry_internal *)entry)->pathlen;
 
+	if (last) {
+		const char *last_c = *last;
+
+		path_start = entry->path;
```
mikeando: This is already done on definition of `path_start`.

mikeando:
Is it worth not adding the hooks in the test repo that should be unused?

mikeando:
I'd also love to see a couple of tests for the varint encoding / decoding.

dturner-tw:
I addressed these.

dturner-tw:
Hm, this AppVeyor failure seems to be bogus -- it's operating on the previous version of the patch. This version uses 251-character filenames (I meant to do 257, but somehow screwed it up).  But AppVeyor is still trying to check out the >260 char names.  I guess I'll add six chars to the names to make them 257 (I want to exceed 256 to ensure that the varint thing is really working right), and hope AppVeyor gets its act together.

ethomson:
Just FYI: it's not a 260 character limit on the filename, it's a 260 character limit on the entire _path_.

dturner-tw:
Hm, it seems to be a limit on the *absolute* path, which is insane.  But I have no idea what the absolute path length is.  Also I broke this one because I forgot to change the test code.  OK, forget long filenames -- we'll just do short ones and assume that the varint stuff works OK.

carlosmn:
Do we need the new repository? Why not add a second index file to an existing with with v4?

ethomson:
@dturner-tw The test seems useful, it's easy enough to `clar_skip` it on Windows and run it only on sane operating systems.

dturner-tw:
@ethomson naturally, skipping it on Windows was my first thought, but no dice -- merely having the file in the libgit2 repo was enough to make AppVeyor unable to check out libgit2 on msys.  (Yeah, we could probably do some hack where the repo is storred bare and then we clone a non-bare copy, but that seems like a giant hassle).

dturner-tw:
@carlosmn I guess we don't need it anymore, now that we've eliminated the long filenames.  Will amend.

dturner-tw:
@ethomson Given the limitations of AppVeyor, can we use merge version?

ethomson:
> (Yeah, we could probably do some hack where the repo is storred bare and then we clone a non-bare copy, but that seems like a giant hassle).

Ah, right you are.  Indeed, this seems unnecessary.

This is really helpful, I appreciate you doing this.  I have a couple of (minor) comments before we merge this...  is this something that you're still able to work on?

dturner-tw:
Until June 3, yes.

dturner-tw:
It occurs to me that perhaps you meant "is this something you're willing to support".  And the answer is yes, I'm willing to support it even after today.  Tag @novalis (my non-twitter github account) with questions.

ethomson:
Yes - I'm sorry if I was unclear, I noticed your message to the git mailing list and wasn't sure if you were dropping the metaphorical mic and walking out and didn't want to be bothered with git-type-things any longer.  :)

If not I was going to make a few tweaks myself.  But since you are that's great, I'll write up some comments / minor requests when I have a moment.  (Sorry about the delay - I missed the notification for your first message, and I'm not going to be in front of a computer much for the next week, but I'll try to get these to you as soon as I can.)

Thanks again.

CHANGELOG.md
```diff
@@ -13,6 +15,9 @@ v0.24 + 1
   writing into a stream. Useful when you do not know the final size or
   want to copy the contents from another stream.
 
+* `git_index_version()` and `git_set_index_version()` to get and set
```
ethomson: Typo here - the function is `git_index_set_version`.

src/index.c
```diff
@@ -746,6 +751,26 @@ static int truncate_racily_clean(git_index *index)
...
+	return index->version;
+}
+
+int git_index_set_version(git_index *index, unsigned int version)
+{
+	assert(index);
+
+	if (version < INDEX_VERSION_NUMBER_LB ||
+		version > INDEX_VERSION_NUMBER_UB)
+		return -1;
```
ethomson: Please `giterr_set` a specific error message here.  Eg, `giterr_set(GITERR_INDEX, "Invalid version number")`.

If this is wildly bogus data that the caller is feeding us, you can also feel free to assert here instead.

src/varint.c
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#include "common.h"
+#include "varint.h"
+
+uintmax_t decode_varint(const unsigned char **bufp)
```
ethomson: Let's make sure we have a `git_` prefix on all the functions that are scoped larger than a compilation unit.  Even when this is not explicitly exported, it could end up that way and we could [colliding with another symbol](https://github.com/libgit2/libgit2/issues/3825).

`git_varint_encode` and `git_varint_decode` would be my recommendations.

tests/varint/encoding.c
```diff
@@ -0,0 +1,34 @@
+#include "clar_libgit2.h"
+#include "varint.h"
+
+void test_varint_encoding__decode(void)
```
ethomson: I would actually move these into the `core` directory, so that they're in `tests/core/varint.c` and they become `tests_core_varint__encode` and `tests_core_varint__decode`.

src/index.c
```diff
@@ -2301,33 +2329,48 @@ static size_t read_entry(
...
 
-	entry.path = (char *)path_ptr;
+		entry.path = (char *)path_ptr;
+	} else {
+		size_t shared = decode_varint((const unsigned char **) &path_ptr);
+		size_t len = strlen(path_ptr);
+		size_t last_len = strlen(*last);
 
-	if (index_entry_dup(out, index, &entry) < 0)
+		tmp_path = git__malloc(shared + len + 1);
```
ethomson: Please `GITERR_CHECK_ALLOC` this allocation

ethomson: Also, `decode_varint` returns a `uintmax_t`, which could be larger than a `size_t`.  Since we are probably using these varints to hold array sizes / indexes, perhaps having encode/decode use `size_t`s makes more sense.

Independent of return types, it looks like this arithmetic could overflow if somebody stuffs a varint value that is (say) `SIZE_MAX - 1`  It looks like we would then write outside of `tmp_path` and corrupt the heap.

src/index.c
```diff
@@ -2301,33 +2329,48 @@ static size_t read_entry(
...
+		memcpy(tmp_path, last, last_len);
+		strcpy(tmp_path + last_len, path_ptr);
```
ethomson: Prefer `memcpy` here since we have the length.

src/index.c
```diff
@@ -2517,15 +2566,31 @@ static bool is_index_extended(git_index *index)
...
 	void *mem = NULL;
 	struct entry_short *ondisk;
 	size_t path_len, disk_size;
 	char *path;
+	const char *path_start = entry->path;
+	size_t same_len = 0;
 
 	path_len = ((struct entry_internal *)entry)->pathlen;
 
+	if (last) {
```
ethomson: All the callers pass this so this check seems unnecessary.

src/index.c
```diff
@@ -2517,15 +2566,31 @@ static bool is_index_extended(git_index *index)
 	return (extended > 0);
 }
 
-static int write_disk_entry(git_filebuf *file, git_index_entry *entry)
+static int write_disk_entry(git_filebuf *file, git_index_entry *entry, const char **last)
```
ethomson: As our convention, we put `out`-params first.

src/varint.c
ethomson: Our convention is generally to not have "in/out" params like this.  Generally we prefer to take the `char *` and return the size used rather than also modifying it.

src/varint.c
```diff
@@ -0,0 +1,37 @@
...
+		if (!val || MSB(val, 7))
+			return 0; /* overflow */
+		c = *buf++;
+		val = (val << 7) + (c & 127);
+	}
+	*bufp = buf;
+	return val;
+}
+
+int encode_varint(uintmax_t value, unsigned char *buf)
```
ethomson: It would be nice to switch the order of these; our convention is to put "out" type params first.

ethomson:
Apologies that this took so long to get back to you.  Day jobs, amirite?

This will be very nice to have - I had some stylistic complaints, mostly just very minor nitpicking.

Thanks again.

src/varint.c
novalis: We need that return value for the decoded value.  Do you mean that there should be a separate out parameter?

ethomson: Yes, sorry, something along the lines of `git_varint_decode(size_t *buflen, const char *buf)` or some such.

src/index.c
novalis: No, it's null if index->version < INDEX_VERSION_NUMBER_COMP

src/index.c
novalis: last is an in/out parameter -- do you want me to change it to two parameters as you suggested in decode_varint?

src/index.c
novalis: I want to keep it general; maybe the code is used somewhere else where overflow is not an issue.

novalis:
moved to https://github.com/libgit2/libgit2/pull/3837 -- please close this one.

PR #3790, opened 2016-05-18
Add retries to win32 p_unlink and p_open.

tylerchurch:
This pull requests follows the pattern used in this previous pull request: https://github.com/libgit2/libgit2/pull/2211 to add retries to p_unlink and p_open.

When doing checkouts I was seeing Dropbox and/or anti-virus locking files while libgit2 was trying to call p_unlink or p_open. These changes seem to make checkout much more reliable for me on Windows.

### DIFF
--- a/src/win32/posix_w32.c
+++ b/src/win32/posix_w32.c
@@ -93,17 +93,31 @@ int p_unlink(const char *path)
 {
 	git_win32_path buf;
 	int error;
+	int unlink_tries;
 
 	if (git_win32_path_from_utf8(buf, path) < 0)
 		return -1;
 
-	error = _wunlink(buf);
-
-	/* If the file could not be deleted because it was
-	 * read-only, clear the bit and try again */
-	if (error == -1 && errno == EACCES) {
-		_wchmod(buf, 0666);
+	/* wait up to 50ms if file is locked by another thread or process */
+	unlink_tries = 0;
+	while (unlink_tries < 10) {
 		error = _wunlink(buf);
+
+		/* If the file could not be deleted because it was
+		 * read-only, clear the bit and try again */
+		if (error == -1 && errno == EACCES) {
+			_wchmod(buf, 0666);
+			error = _wunlink(buf);
+
+			if (error == -1 && errno == EACCES) {
+				Sleep(5);
+				unlink_tries++;
+			} else {
+				break;
+			}
+		} else {
+			break;
+		}
 	}
 
 	return error;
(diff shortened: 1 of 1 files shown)


src/win32/posix_w32.c
```diff
@@ -296,7 +312,23 @@ int p_open(const char *path, int flags, ...)
...
-	return _wopen(buf, flags | STANDARD_OPEN_FLAGS, mode & WIN32_MODE_MASK);
+	/* wait up to 50ms if file is locked by another thread or process */
+	open_tries = 0;
+	while (open_tries < 10) {
+		handle = _wopen(buf, flags | STANDARD_OPEN_FLAGS, mode & WIN32_MODE_MASK);
+		if (handle != -1) {
+			break;
+		}
+
+		if (errno == EACCES) {
```
ethomson: When will this be reached?  MSDN says that `errno` is set to `EACCES` when `_wopen` returns `-1`, but the condition above would have caught that case.

ethomson:
I think use of `goto`s may simplify this code a bit.  However, (and keep in mind that I use this phrase quite sparingly) a macro might be a decent solution here, too, since this seems like a common pattern in the Windows stuff.

I seem to remember that we do this retry dance elsewhere in the win32 compat layer.  It would be nice to unify the style amongst them (give them a consistent timeout, with a `#define`d constant, etc).

Thanks for starting this, I think it will benefit win32 users quite a bit.

src/win32/posix_w32.c
tylerchurch: My interpretation of the docs here: https://msdn.microsoft.com/en-us/library/z0kc8e3z.aspx was that it could be `EACCES`, `EEXIST`, `EINVAL`, `EMFILE`, or `ENOENT`. `EACCES` was the only one I wanted to retry on, but if you'd like I can kill the conditional and we can just retry no matter the errno.

tylerchurch:
Yeah, a macro would be nice, though so far the usages of the different functions seem just slightly different enough that I'm not sure how useful it would actually be (e.g. p_open needs to return a handle, but p_unlink doesn't.) If you've got an idea for a macro you think would work though, I'm happy to integrate it.

I'm with you on adding some global timeout constants. My preference would be for there to be some library wide setting that could be changed by whatever application is loading libgit2, rather than a `#defined` constant. Are there any other similar settings defined anywhere? If you could point me in the right direction, I can take a look at adding it.

ethomson:
I'm not tied to this needing a nice configuration.  I think that it's fine to `#define` these - I don't have any notion that this is something that somebody wants to tune.  But if you would benefit from the ability to change it then 👍 .

Our globals are set in https://github.com/libgit2/libgit2/blob/master/src/global.c - let me know if you have questions about the patterns in there.

PR #3792, opened 2016-05-18
Fix comment for GIT_FILEMODE_LINK

edquist:
0120000 is symbolic link, not commit.

(No functional change.)


ethomson:
Yep!  That's definitely an oversight.  Thanks for the fix!  :)

PR #3794, opened 2016-05-19
Tree updater fixups

carlosmn:
Fix removing a deep blob, as well as plug leaks in this code.


carlosmn:
I'm just gonna go ahead and merge this in since it's fixups validated by tests.

PR #3797, opened 2016-05-24
tree: handle removal of all entries in the updater

carlosmn:
When we remove all entries in a tree, we should remove that tree from
its parent rather than include the empty tree.


carlosmn:
I've made the tests in gitrpc run with this, so it's looking good.

PR #3798, opened 2016-05-24
test: Fix stat() test to mask out unwanted bits

mmuman:
Haiku and Hurd both pass extra bits in struct stat::st_mode.

### DIFF
--- a/tests/checkout/index.c
+++ b/tests/checkout/index.c
@@ -294,11 +294,12 @@ void test_checkout_index__options_dir_modes(void)
 	(void)p_umask(um = p_umask(022));
 
 	cl_git_pass(p_stat("./testrepo/a", &st));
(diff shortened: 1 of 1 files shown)


mmuman:
Of course win32 doesn't know about the most basic POSIX constants:
```
'S_IUMSK' : undeclared identifier
```
Should I patch src/posix.h ?

ethomson:
Neither does Linux or Mac, though.  I don't think that's a POSIX thing, I think it's something Haiku invented.

mmuman:
Oh, indeed, I didn't check other stat headers... Linux' sys/stat.h has an ALLPERMS but ifdeffed, so I'm not sure it's always available. And mingw headers don't seem to have it. I'll just hardcode it but it's ugly.

mmuman:
Let's see this way...

mmuman:
Looks like tests passed :)

ethomson:
Looks good to me, thanks!

PR #3799, opened 2016-05-25
Use AppVeyor's Start-FileDownload cmdlet

sschuberth:
Start-FileDownload maintains current directory context and allows
specifying a request timeout, see [1].

[1] https://www.appveyor.com/docs/how-to/download-file#start-filedownload-cmdlet

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,7 +33,7 @@ build_script:
 test_script:
 - ps: |
     $ErrorActionPreference="Stop"
-    Invoke-WebRequest https://github.com/ethomson/poxyproxy/releases/download/v0.1.0/poxyproxy-0.1.0.jar -OutFile poxyproxy.jar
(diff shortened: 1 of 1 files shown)


ethomson:
Sorry...  I guess I'm not clear on the advantage.  This seems reasonable, but do we need to maintain the directory context, or are we interested in the timeout?

sschuberth:
There's nothing to be sorry about. You're right that in this particular case there's not (yet) any advantage of using `Start-FileDownload` instead of `Invoke-WebRequest`. To be honest, I did this commit as part of a bulk change across different GitHub projects I've forked. For some of the other projects this change makes more sense as they benefit e.g. from specifying a timeout. I still submitted the change for libgit2 as `Start-FileDownload` now feels like the standard way downloads should be done in AppVeyor. But if you don't agree or would like to keep things simpler / more vanilla PowerShell-ish, I'm completely fine with closing this unmerged.

ethomson:
Okay!  Thanks for clarifying, I just didn't see the immediate benefit so I was wondering if there was some detail that I did not understand.

If this is both recommended by AppVeyor and has benefits that we can use in the future, then I'll merge this.  Thanks!

PR #3805, opened 2016-05-30
Don't reuse the old index.

arthurschreiber:
`rebase_next_inmemory` was reusing the same index throughout all rebase operations. This was causing entries to get lost sporadically and would lead to incorrect rebase results.

Instead of reusing the index, always switch to the index created by the latest merge operation.

This fixes the issue demonstrated in https://github.com/libgit2/rugged/pull/609.

### DIFF
--- a/src/rebase.c
+++ b/src/rebase.c
@@ -881,8 +881,11 @@ static int rebase_next_inmemory(
 		rebase->index = index;
 		index = NULL;
 	} else {
-		if ((error = git_index_read_index(rebase->index, index)) < 0)
-			goto done;
+		git_index *tmp = NULL;
+
+		tmp = rebase->index;
+		rebase->index = index;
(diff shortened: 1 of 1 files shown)


savalaraymond:
I agree on term and prvcy..please continue with the date on when it will be
fixed so i can resolve on git hub...please be spacific.expections on
benefits are high....

savalaraymond:
Show me a fix to stay secure

arthurschreiber:
If it's true that `git_index_read_index` is the culprit here, `git_stash_apply` is probably broken in a similar fashion. 😞

ethomson:
I'm closing this; reusing the index should reduce the number of allocations, which should be a perf win.  I think that we nailed down the problems that we were seeing (in `git_index_read_index`) that this was meant to address independently.

PR #3810, opened 2016-06-04
Return `HEAD` itself for `git_repository_head` when unborn

ethomson:
When consulting `git_repository_head` for an unborn branch, set the out pointer to the current `HEAD` as well as returning `GIT_EUNBORNBRANCH`.  This allows callers to determine what the unborn branch _is_.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -1747,10 +1747,16 @@ int git_repository_head(git_reference **head_out, git_repository *repo)
 		return 0;
 	}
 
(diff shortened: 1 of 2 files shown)


ethomson:
Nah, I think this is actually quite stupid.  Nobody expects to have to free a reference when there's an error code returned.

ethomson:
(That said, it might be nice to have a mechanism to get peel HEAD as much as possible, but this probably isn't it.)

PR #3812, opened 2016-06-06
fetch: Fixed spurious update callback for existing tags.

hackhaslam:
When an existing tag is changed on the remote, the default (autotag) behavior is to not overwrite the local tag, but the update callback is still called as if the tag had been updated. This commit fixes that.

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -1423,7 +1423,11 @@ static int update_tips_for_spec(
 		/* In autotag mode, don't overwrite any locally-existing tags */
 		error = git_reference_create(&ref, remote->repo, refname.ptr, &head->oid, !autotag, 
 				log_message);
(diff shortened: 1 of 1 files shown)


ethomson:
This is very minor:  I wonder if it would be clearer to write this as `if (error == GIT_EEXISTS) continue`?

This may help us in the future if we want to add additional logic to do with the remote, as right now the conditional seems to be about whether to call the callbacks, but we really want to stop all processing if the ref already existed.  If there's additional processing in the future then it may make it more clear.

Thanks again for tackling this - and apologies for the delay in reviewing.

hackhaslam:
Yeah, that makes sense. Updated accordingly.

hackhaslam:
And thanks for the review!

PR #3813, opened 2016-06-07
submodule: Try to fetch when update fails to find the target commit.

### LINKED ISSUE #3783: git_submodule_update doesn't fetch
When a repository is fetched without recursively fetching its submodules git_submodule_update can fail with missing target commit in the submodule. I believe that core git has some logic to fetch if the submodule doesn't have the target object.

hackhaslam:
This fixes #3783. When submodule update doesn't find the target commit in the submodule, fetch from the default remote and then try to lookup the commit again.

### DIFF
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -93,6 +93,7 @@ static git_config_backend *open_gitmodules(git_repository *repo, int gitmod);
 static git_config *gitmodules_snapshot(git_repository *repo);
 static int get_url_base(git_buf *url, git_repository *repo);
 static int lookup_head_remote_key(git_buf *remote_key, git_repository *repo);
+static int lookup_default_remote(git_remote **remote, git_repository *repo);
 static int submodule_load_each(const git_config_entry *entry, void *payload);
 static int submodule_read_config(git_submodule *sm, git_config *cfg);
 static int submodule_load_from_wd_lite(git_submodule *);
@@ -1131,7 +1132,7 @@ int git_submodule_update(git_submodule *sm, int init, git_submodule_update_optio
 			if (error != GIT_ENOTFOUND)
 				goto done;
 
-			if (error == GIT_ENOTFOUND && !init) {
+			if (!init) {
 				giterr_set(GITERR_SUBMODULE, "Submodule is not initialized.");
 				error = GIT_ERROR;
 				goto done;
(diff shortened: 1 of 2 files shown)


ethomson:
@jamill Would you have a moment to spare a second set of eyes here?  I realize that you're not spending as much time with submodules these days, but you certainly still know more about them than I do.  :)

src/submodule.c
```diff
@@ -1171,9 +1172,21 @@ int git_submodule_update(git_submodule *sm, int init, git_submodule_update_optio
...
-			(error = git_object_lookup(&target_commit, sub_repo, git_submodule_index_id(sm), GIT_OBJ_COMMIT)) < 0 ||
-			(error = git_checkout_tree(sub_repo, target_commit, &update_options.checkout_opts)) != 0 ||
+		if ((error = git_submodule_open(&sub_repo, sm)) < 0)
+			goto done;
+
+		/**
+		 * Look up the target commit in the submodule.
+		 * If it isn't found then fetch and try again.
+		 */
+		if ((error = git_object_lookup(&target_commit, sub_repo, git_submodule_index_id(sm), GIT_OBJ_COMMIT)) < 0) {
```
jamill: Is there a specific error code that is returned when the target commit could not be found? Could we be more selective and only attempt to fetch for the `not found` error?

jamill:
Looks OK to me - and I think this matches the expected behavior. I might also suggest updating the comments for this function to clarify that this will fetch if there are missing commits. Thanks!

hackhaslam:
Thanks for the review! I restricted the fetch logic to the not found error, and changed the documenation comment to mention the new behavior. I also made a slight drive-by fix for a duplicate check in the clone side.

I'm noticing now that `submodule update` has a `-no-fetch` flag. Maybe there should be a `no_fetch` flag in the submodule update options? Probably not a big deal though.

ethomson:
> I'm noticing now that submodule update has a -no-fetch flag. Maybe there should be a no_fetch flag in the submodule update options? Probably not a big deal though.

I agree, that would be a really nice option.  Is this something you want to tackle in this PR or do you want me to merge this and tackle this later?

hackhaslam:
I guess that I should probably add it in this PR so that I don't forget it. I'll add it soon.

hackhaslam:
I added the flag. I also made another small drive-by fix to the version used in the update options initializer.

ethomson:
> I also made another small drive-by fix to the version used in the update options initializer.

Good catch on that one.  👌 Thanks!

PR #3814, opened 2016-06-07
Fix invalid memory references

pks-t:
Two small fixes for invalid memory references. As always, some details are included in the commit messages.

I've pushed a fix similar to the smart-transports commit once before, but back then the fix wasn't as obvious and as I wasn't able to recap my own thought process when you asked about the non-obvious fix I simply dropped the commit. But I hope the issue should be clear now.

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -505,10 +505,11 @@ static int index_remove_entry(git_index *index, size_t pos)
 	int error = 0;
 	git_index_entry *entry = git_vector_get(&index->entries, pos);
 
-	if (entry != NULL)
+	if (entry != NULL) {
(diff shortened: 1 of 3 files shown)


ethomson:
This looks good to me.  In reference to one of your comments:

> While they could always reset the passed out-pointer to `NULL`
before calling `git_pkt_parse_line` and determine if the pointer
has been set afterwards, it makes more sense to update
`git_pkt_parse_line` to set the out-pointer to `NULL` itself when
it encounters such an empty packet.

Indeed this is the expectation - although this is not codified in a document, the expectation is that functions that take out-params should always set those, even in error cases.  👍

PR #3815, opened 2016-06-07
Coverity user model

pks-t:
So by now more than half of the issues reported by Coverity can be labeled as false positives as the number of issues went down. I've written a user model that I am using to fix most of these false positives in my own fork, which brings down the number of issues from 37 to 14. Most of these issues are related to either git_vector or git_buf, which seem to be too complex to handle for Coverity.

I've split up this PR into two commits. The first commit includes the user model I am actually using right now. Due to an expired license (dunno why, I've contacted Coverity support but got no feedback until now) I am not able to update the user model, though, so I cannot test changes made in the second commit.

I'd be glad if this got accepted. Accepting alone won't help, though, as there seems to be no simple way of using the model when executing Coverity's tooling. Instead, someone needs to update the user model online. To do so, you'll have to go to scan.coverity.com, select the libgit2 project and then select the tab "Analysis Settings". There is a section "Modeling File" at the bottom where the file can be uploaded.

### DIFF
--- a/script/user_model.c
+++ b/script/user_model.c
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+void *realloc(void *ptr, size_t size);
+size_t strlen(const char *s);
+
+typedef struct git_vector {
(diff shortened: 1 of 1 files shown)


ethomson:
Merging this I did upload the modeling file, we can iterate if it's not successful.

pks-t:
While not fixing as much as I'd hoped, at least some bugs went away. Thanks for merging this.

PR #3816, opened 2016-06-07
Memory leak fixes

pks-t:
This fixes all memory leaks displayed by AppVeyor's debug CRT. The winhttp-related plug is the only actual fix for memory leaks inside of libgit2 and not its tests.


ethomson:
Thanks again for tackling these.  ❤️

PR #3818, opened 2016-06-09
Fix truncation of SHA in error message for git_odb_read

### LINKED ISSUE #3753: `git_odb_read` is giving error with off-by-1 sha string
I'm getting an off-by-1 error [here](https://github.com/libgit2/libgit2/blob/master/src/odb.c#L998). Making a `git_graph_descendant_of` that I'm expecting to error gives me the correct error but the sha has the last char removed.

image

Not really game breaking, but I thought I should at least log it.

meatcoder:
This is an attempt to fix #3753. The bug appears to only occur when the `oid_len` parameter passed to `git_odb__error_notfound()` is the maximum length of GIT_OID_HEXSZ. This PR fixes it by passing the actual buffer length to `git_oid_tostr()`.

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -1362,8 +1362,8 @@ int git_odb__error_notfound(
 {
 	if (oid != NULL) {
 		size_t buf_len = oid_len;
-		if (oid_len == GIT_OID_HEXSZ) buf_len++;
 		char oid_str[GIT_OID_HEXSZ + 1];
+		if (oid_len == GIT_OID_HEXSZ) buf_len++;
 		git_oid_tostr(oid_str, buf_len, oid);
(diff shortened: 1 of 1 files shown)


meatcoder:
Push a commit to fix MSVC compile error. Should be good to go.

meatcoder:
Thanks for looking at this and the detailed explanation @ethomson. I've updated the pull request to to also fix the cases where shorter ID strings are passed to `git_odb_read()`. Let me know how it looks. Thanks again.

ethomson:
This looks great, thanks @meatcoder !

ethomson:
I squashed and manually merged this, thanks again!

PR #3822, opened 2016-06-14
documentation: improve docs for `checkout_head`

### LINKED ISSUE #2864: git_checkout_head() doesn't do anything (in practice)
The doc says:

    * Updates files in the index and the working tree to match the content of
    * the commit pointed at by HEAD.

But since `git_checkout_head()` simply calls `git_checkout_tree()` with a `NULL` tree, the checkout code will use `HEAD` as a baseline and attempts to check out `HEAD` against it. In practice this means, no attempt will be made to touch modified files (whatever the mode) or create missing ones (unless you happen to be `GIT_CHECKOUT_SAFE_CREATE` mode).

As an example, if you move the `HEAD` to a different commit then call `git_checkout_head()` expecting the index and workdir to be updated to reflect the new commit (as per the docs), nothing happens (and there are no errors either of course).

Is there a bug in the implementation, is the doc wrong, or is it really the intended behavior of `git_checkout_head()`?

PS: If the latter, what's the actual and most efficient way using libgit2 to update the index and workdir to match the `HEAD`?

ethomson:
`git_checkout_head` is sadly misunderstood as something that can
switch branches.  It cannot.  Update the documentation to reflect this.

Fixes #2864

### DIFF
--- a/include/git2/checkout.h
+++ b/include/git2/checkout.h
@@ -313,6 +313,13 @@ GIT_EXTERN(int) git_checkout_init_options(
  * Updates files in the index and the working tree to match the content of
  * the commit pointed at by HEAD.
  *
+ * Note that this is _not_ the correct mechanism used to switch branches;
+ * do not change your `HEAD` and then call this method, that would leave
+ * you with checkout conflicts since your working directory would then
+ * appear to be dirty.  Instead, checkout the target of the branch and
+ * then update `HEAD` using `git_repository_set_head` to point to the
+ * branch you checked out.
+ *
  * @param repo repository to check out (must be non-bare)
  * @param opts specifies checkout options (may be NULL)
  * @return 0 on success, GIT_EUNBORNBRANCH if HEAD points to a non


include/git2/checkout.h
```diff
@@ -313,6 +313,12 @@ GIT_EXTERN(int) git_checkout_init_options(
  * Updates files in the index and the working tree to match the content of
  * the commit pointed at by HEAD.
  *
+ * Note that this is _not_ the correct mechanism used to switch branches;
+ * do not change your `HEAD` and then call this method, that would leave
+ * you with checkout conflicts since your working directory would then
+ * appear to be dirty.  Instead, checkout the target of the branch and
+ * then update `HEAD` if that was successful.
```
carlosmn: Should we mention `git_repository_set_head()` here, as that's the convenient way we offer to switch the active branch.

ethomson: Updated to mention `git_repository_set_head`.

PR #3823, opened 2016-06-15
checkout: use empty baseline when no index file exists

### LINKED ISSUE #3811: Difference between `git checkout --force HEAD` and `git_checkout_head()` with force
It seems that `git checkout --force HEAD` and `git_checkout_head()` (with force) behave differently in some cases. `git_checkout_head()` appears to do nothing.

To illustrate this I created a repo that can be cloned and tested - [clns/libgit2-checkout-head](https://github.com/clns/libgit2-checkout-head). All the details are included there. The repo actually uses [git2go](https://github.com/libgit2/git2go) instead of [libgit2](https://github.com/libgit2/libgit2), for convenience.

I think this is related to #2864. It happens when the HEAD is updated before the checkout, and in that issue it is explained why it happens. However `git checkout --force HEAD` behaves differently (it does modify the worktree even if the HEAD was updated prior), and my understanding is that the too should work the same.

ethomson:
When no index file exists and a baseline is not explicitly provided, use an empty baseline instead of trying to use `HEAD` as the baseline.

This is for compatibility with git:

```
% git status
On branch master
nothing to commit, working directory clean
% rm .git/index
% git status
On branch master
Changes to be committed:

	deleted:    README
	deleted:    branch_file.txt
	deleted:    link_to_new.txt
	deleted:    new.txt

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	README
	branch_file.txt
	link_to_new.txt
	new.txt
% git checkout HEAD
error: The following untracked working tree files would be overwritten by checkout:
	README
	branch_file.txt
	link_to_new.txt
	new.txt
Please move or remove them before you can switch branches.
Aborting
% git checkout --force HEAD
% git status
On branch master
nothing to commit, working directory clean
```

Our previous behavior was to try to load `HEAD` when the index does not exist, which would simply do a diff between the baseline (`HEAD`) and the target (again, `HEAD`), which would be a noop.

Fixes #3811

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -2430,8 +2430,13 @@ static int checkout_data_init(
 
 	if (!data->opts.baseline && !data->opts.baseline_index) {
 		data->opts_free_baseline = true;
+		error = 0;
 
-		error = checkout_lookup_head_tree(&data->opts.baseline, repo);
+		/* if we don't have an index, this is an initial checkout and
+		 * should be against an empty baseline
+		 */
+		if (data->index->on_disk)
+			error = checkout_lookup_head_tree(&data->opts.baseline, repo);
 
 		if (error == GIT_EUNBORNBRANCH) {
 			error = 0;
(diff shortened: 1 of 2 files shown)


ethomson:
Note that this is still not completely correct:  the "is this dirty" checkout conflict handling is not correctly taking the index into account but that's a different (and bigger) issue and I will follow up with a related pull request.

Issue #964, opened 2012-10-02
support for hooks (pre-commit / post-commit / post-merge)


fbergmann:
It seems that libgit2 does not allow the processing of hooks, that git would perform. Any chance for adding this to the api in a future version?

ethomson:
We've discussed this back and forth a lot.  We will *not* be executing any hooks directly.  We don't want the library to suddenly start forking and executing processes and such.  This is firmly the territory of the application.

However, we do want to have the right callbacks in the right places to make sure that it's easy for people to execute hooks if they want to.

lukeapage:
@ethomson why was this closed? should there not be a new issue for allowing applications to find hooks?

carlosmn:
Finding hooks amounts to getting the contents of the `$gitdir/hooks/` directory and seeing what files are executable (if the system has this). This is not something which you need the library to do. While we could provide an accessor to the hooks directory, there's nothing the library can do better than you using your language's features to look at a directory.

lukeapage:
Makes sense, thanks for answering.

knocte:
This is very sad, as it renders basically all IDEs (at least all of them using libgit2) unable to integrate with commit hooks workflows.

The point of git commit hooks is that they are respected, not that they may ignored depending on the tool you use to interact with git.

ethomson:
> This is very sad, as it renders basically all IDEs (at least all of them using libgit2) unable to integrate with commit hooks workflows.

No, there's nothing stopping your IDE from running hooks.  If your IDE does not support git hooks, then that's a bug / missing feature in the IDE.  As has been mentioned previously in this issue:  libgit2 is not going to run external commands magically.  There are a plurality of reasons for this, but primarily, running programs is a concern of the IDE.  For instance:  Visual Studio has very specific mechanisms for running external commands and if libgit2 simply starts running things, this would be a trainwreck.  No cancellation, no way to get output to the console window, etc.

Thus, this is a concern of the IDE, not a concern of libgit2.

If you are an IDE author and we can help improve this for you, then please let us know how.  If not, please file an issue with your IDE.

linquize:
"Using callback functions as hooks" vs "executing bash shell scripts directly"?

ethomson:
> "Using callback functions as hooks" vs "executing bash shell scripts directly"?

Indeed, we try to provide callbacks at all the states in the workflow so that you can invoke the necessary callbacks at that time, with the necessary information so that you can emulate running the hooks.

[...]

PR #3824, opened 2016-06-16
Hook support

tiennou:
This is a first pass at supporting hooks. It aims to provide the basic blocks (as per PROJECTS.md), so enumeration, loading and saving is supported, and there's a quick skeleton of how I'm planning to handle hook execution.

A few points/questions :
- I was under the impression there had already been some discussion around the design, but I haven't been able to find them.
- The list of hooks is hardcoded. This is completely debatable, but given variable argument count, I thought something that acts as a "we support these hooks" was preferable to YMMV.
- I seem to remember bindings don't like callbacks (and me too, given my current enumeration test is human-based :wink:). Should I provide an alternate way (e.g. `git_hook_iterator_new()`) ?
- The plan for actually executing the hooks is to have something like `git_hook__execute(hook_name, ...)` and sprinkle those everywhere, but it might or might not be doable — I haven't been that far yet.


whoisj:
Some hook support already exists. The trick with hooks is that they're basically shell scripts and require the platform to support executing then.

tiennou:
Yep, I'm aware that direct hook execution does not belong in the library. The current plan is to have clients use `git_hook_register` to say they're interested in hook notifications, and then (maybe ?) provide the full path to the script to execute if applicable, or even no path at all if we want to support that.

Also, what did you meant by "already exists" ? I haven't been able to find anything other than some references in tests, but maybe I'm blind 😉 .

whoisj:
> Also, what did you meant by "already exists" ?

I'm sorry, the hook support is in the C# wrapper library, not in LibGit2 core. My mistake.

ethomson:
It feels to me like you're trying to design the support in libgit2 that you think you want before you've tried to implement any hooks in your application itself.  I feel like trying to build "hooks support" into libgit2 before trying to get your application to do anything is actually the wrong direction to attack this problem.  Instead, I think that you should get your application to exec some hooks so that you know _what_ you can get libgit2 to help you with.

I realize it seems counterintuitive to do it in the application first, since generally you're coding a functionality or a UI on top of libgit2 and in that case the libgit2 part must already exist.  But without having actually succeeded in running a hook from your application, I think that we will merely be making wild guesses as to what we want the hooks support (if any) to look like and will very likely design this wrong.

As a concrete example, `it's unclear to me how `git_hook_load` would be a useful addition to the library.  You want to _run_ a hook, you never want to load a hook into memory.  It's an executable - _probably_ (but not necessarily) a shell script, but it could be an ELF binary (or COFF, or a.out, or whatever).  Heck, it could be a Java class and you're relying on `binfmt_misc` to know how to execute it for you.  The point is, you need to `exec` it, you don't want to have a copy of the image in memory.

Similarly, most hooks don't make sense in the context of libgit2.  Things like the `prepare-commit-msg` and `commit-msg` hooks are things that could never possibly be run, since we don't launch an editor to get the commit message.

We think that we have callbacks in the appropriate places that you can map those to hook execution if that's something that you're interested in (though they are not called "hooks" callbacks, they are generally more generic than that.)  I think the best first place to start is to identify hooks that you want to support that are lacking callbacks (that _can_ be supported, that is, unlike the aforementioned hooks around commit messages).  That's a clear improvement.

libgit2 certainly can provide some support for hooks, like providing the path.  This seems of limited use to me, since it's mostly just string concatenation, (`.git/hooks/commit-msg`)  But I can understand wanting (say) to be able to query for that path to reduce the amount of git repository-type business logic you have in your application.  Similarly, I can imagine asking for a map of the hooks that exist in `.hooks`.  Again, this is a relatively trivial `readdir` but I can imagine wanting an API to provide some abstraction here.

Anyway.  There are probably some things that libgit2 can help you with here, but I would like to be conservative with these APIs.

tiennou:
Thanks for the thorough reply @ethomson. Just so we're clear, the API I wrote is inspired from what's in PROJECTS.md, and `git_hook_load/save` is suggested there. As you point out (and I do had a pause when I sketched the API), it might not make sense every time, and I'm now tempted to drop it completely (maybe keeping just a few helpers for path-wrangling)...

Looking at that as an app developer, what I want is a way to be told when I need to exec a hook, because I want to be compatible with git/not break expectations. For example, I would expect (and I might be wrong 😉) libgit2 to detect that there's a `commit-msg` hook set up, trigger a callback to my hook execution function*, and abort or proceed with committing based on what was reported by the hook. It should not allow the commit to go through without telling me about it. I do see how the `prepare-commit-msg` would be weird to handle though (I'm not even sure if we support `commit.template`).

* This is the wacky part, because we could either pass data that makes sense to us (like in that case the commit message), or helpfully write a file and pass that so the execution function doesn't have to do it (which IMHO reduces duplication).

But then, maybe that whole responsibility could indeed be moved down to the application (provided the same effect can be achieved through the generic callbacks). I might be worrying too much about duplicated code 😉.

ethomson:
> Just so we're clear, the API I wrote is inspired from what's in PROJECTS.md, and git_hook_load/save is suggested there.

Thanks for the clarification - I didn't realize that was there.  :P

> For example, I would expect (and I might be wrong 😉) libgit2 to detect that there's a commit-msg hook set up, trigger a callback to my hook execution function*, and abort or proceed with committing based on what was reported by the hook.

Right - so the thing is that we don't have an API that maps to the `git commit` command.  We have a handful of `git_commit_create` commands, but these correspond more closely to the `git-commit-tree` plumbing command.

This is important because if we were to (say) fire the `commit-msg` command in the `git_commit_create` functions, then we would be firing it every time a commit was created, not every time the user was "performing a commit" (like what `git commit` does).  So, during merges and rebases.  If you had some `commit-msg` hook that actually validated the contents of a message, then we would fire it when we create recursive merge bases (which have totally bogus internal messages, since they're not intended to leak out).  It would be quite disastrous.  :/

This is why we've asked the callers to handle hooks-like things:  you know when you're going through a commit workflow like the `git commit` command line, versus creating commits that are (say) part of a merge or a cherry-pick operation.

Of course, the problem here is that various other libraries (say, LibGit2Sharp) may provide a commit like workflow as API.  So this makes the problem a little bit more difficult.

Anyway.  I'm sorry that I don't have clearer answers here yet.  This is something that's been discussed a few times, and I think that all the vendors who are building UI on top of libgit2 should absolutely be supporting.  As a result, I would love to give as much help as we can to you (and the rest) so that it's less of a slog.

tiennou:
Thanks, that makes it clear why hook support is a hard problem. IIUC, if libgit2 had a `git_perform_commit` (that would be considered "porcelain"/high-level) it would make sense.

I'll go ponder what the bindings are currently up to, to see if there's any common infrastructure we could bring in, but arguably this API is bound to be a simple wrapper around `exec`/`popen`/Windows-y stuff with some sugar on top, like `git_hook_exec("update", "the-ref", "bda34...", "eff453...", my_execution_callback)`.

tiennou:
#964 (I knew there had been some discussion already).

tiennou:
Here's a first pass at hook execution. It allows the client to register a callback for hook execution and a wrapper that we (or, for some hooks, the client) can call to cause that callback to be executed with the expected arguments after the checks are done.

This still needs to tackle the case of the few hooks that do I/O, but I think I'll just pass a `git_buf` in the `git_hook_env` and let the client do its thing.

I'm saddened by the fact that this is so STRINGS EVERYWHERE, with additional NULL SENTINELS 😅 . There might be a way to make it cleaner, but as "conservative" is a keyword, let's keep it simple for now.

It's related to the point above, but there will be times where the client has OIDs and `git_buf`s, and the hook expects SHAs and filenames. I was thinking that there could be some wrapper (for example `git_hook_execute_pre_push("local-ref", "remote-ref", ...)` and have that automatically build the required strings for execution. But again, "conservative".

Right now the code has a list of hardcoded hook names, but there might be a way to make that more generic (just found out about `git_path_diriter`). And the generic execution call could be used by clients who which to call those (with the ~~pain~~caveats of argument-building bestowed upon them).

include/git2/hook.h
```diff
@@ -0,0 +1,54 @@
...
+ */
+GIT_BEGIN_DECL
+
+typedef int (*git_hook_foreach_cb)(const char *hook_name, void *payload);
+
+GIT_EXTERN(int) git_hook_enumerate(
+	git_repository *repo,
+	git_hook_foreach_cb callback,
+	void *payload
+);
```
pks-t: Our coding style has the closing brace on the same line as the last argument. Also applies for the following functions.

include/git2/hook.h
```diff
@@ -0,0 +1,54 @@
...
+ * @brief Git hook management routines
+ * @defgroup git_hook Git hook management routines
+ * @ingroup Git
+ * @{
+ */
+GIT_BEGIN_DECL
+
+typedef int (*git_hook_foreach_cb)(const char *hook_name, void *payload);
+
+GIT_EXTERN(int) git_hook_enumerate(
```
pks-t: I think `git_hook_foreach` would be more fitting

include/git2/hook.h
```diff
@@ -0,0 +1,54 @@
...
+
+GIT_EXTERN(int) git_hook_execute(
+	git_repository *repo,
+	const char *hook_name,
+	...
+);
+
+/* @} */
+GIT_END_DECL
+#endif
```
pks-t: Newline at end is missing here

src/hook.c
```diff
@@ -0,0 +1,152 @@
...
+
+	assert(hook_path);
+
+	hook_dir_len = strlen(repo->path_repository) + strlen(GIT_HOOKS_DIR);
+	*hook_path = git__malloc(hook_dir_len + MAX_HOOK_LEN);
+	GITERR_CHECK_ALLOC(*hook_path);
+
+	strcpy(*hook_path, repo->path_repository);
+	strcat(*hook_path, GIT_HOOKS_DIR);
+	strcat(*hook_path, hook_name);
```
pks-t: You might use our `git_buf` API instead, it would make this function much easier on the reader.

src/hook.c
```diff
@@ -0,0 +1,152 @@
...
+	struct stat hook_stat;
+
+	assert(hook_path);
+
+	/* Skip missing hooks */
+	err = p_stat(hook_path, &hook_stat);
+	if (err) return -1;
+
+	/* Check exec bits */
+	if ((hook_stat.st_mode & (S_IXUSR|S_IXGRP|S_IXOTH)) == 0) return -1;
```
pks-t: Just wondering, do we really care for `S_IXOTH`? Above you've defined `HOOK_FILE_MODE` as being `0776`, that is the file is not executable for others.

src/hook.c
```diff
@@ -0,0 +1,152 @@
...
+}
+
+int git_hook_enumerate(
+	git_repository *repo,
+	git_hook_foreach_cb callback,
+	void *payload)
+{
+	assert(repo && callback);
+
+	for (size_t hook_id = 0; hook_id < sizeof(githooks) / sizeof(char *); hook_id++) {
```
pks-t: We use strict C89, so declarations have to be at the function's head.

pks-t: We have a macro `ARRAY_SIZE` which might be used here instead of manually computing the array's size.

src/hook.c
```diff
@@ -0,0 +1,152 @@
...
+{
+	assert(repo && callback);
+
+	repo->hook_executor = callback;
+	repo->hook_payload = payload;
+
+	return 0;
+}
+
+int git_hook_exec_va(git_repository *repo, const char *name, va_list args)
```
pks-t: I'd either make the function available via the header or make it a static function, dropping the `git_` prefix. Furthermore, I'd remove the naming inconsistency of `git_hook_exec_va` vs `git_hook_execute`, that is either call both `exec` or both `execute`.

src/hook.c
```diff
@@ -0,0 +1,152 @@
...
+	assert(repo && name);
+
+	err = build_hook_path(&env.path, repo, name);
+	if (err != 0) goto cleanup;
+
+	err = check_hook_path(env.path);
+	if (err != 0) goto cleanup;
+
+	while ((arg = va_arg(args, char *)))
+	{
```
pks-t: The opening brace has to go on the same line.

src/hook.c
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	va_end(args);
+
+	env.args.strings = (char **)git_vector_detach(&env.args.count, NULL, &arg_vector);
+
+	err = repo->hook_executor(env, repo->hook_payload);
+
+cleanup:
+	git__free(env.path);
```
pks-t: You are leaking `env.args.strings` here.

src/hook.c
```diff
@@ -0,0 +1,152 @@
...
+{
+	int err = 0;
+	va_list hook_args;
+
+	va_start(hook_args, hook_name);
+	err = git_hook_exec_va(repo, hook_name, hook_args);
+	va_end(hook_args);
+
+	return err;
+}
```
pks-t: Newline is missing

src/repository.h
```diff
@@ -142,6 +143,9 @@ struct git_repository {
 
 	git_atomic attr_session_key;
 
+	git_hook_execution_cb hook_executor;
+	void *hook_payload;
```
pks-t: Hm. There's some memory management issues with the payload here. When the repository is free'd, we'll loose the payload if the caller does not explicitly free it for himself, which is rather cumbersome. We'd need to have either some function `hook_free` which can be registered with the repository or another API approach here. Will discuss in a later comment.

pks-t:
I like the direction this is going. The code is nice to read, barring the few comments I've added, so good work.

The only thing I don't like about it at the moment is that hooks are registered per repository. This introduces some problems with lifecycle management of the repository as the callback payload is attached to the repo and cannot be simply free'd by the repo when calling `git_repository_free`. So we'd at least need to have another function `hook_payload_free` at the repository struct which will be called on the payload as soon as the repo is freed.
Furthermore, developers might wish to have multiple hooks registered for a repository. Currently it is only possible to have a singular hook, which feels a bit cumbersome. It sure would be possible to simply have the callback call out to multiple other functions, but this feels more like a hack than anything else.

To be honest I've got no obvious solution for a better execute-API. One thing I've been thinking about is to have a global list of callbacks instead of a per-repository one, which might make it easier to manage them lifecycle-wise. But in fact this would introduce its own distinct set of problems. Maybe per-repo ones aren't that bad after all. *shrug*

src/hook.c
tiennou: `HOOK_FILE_MODE` is a leftover from the save/load function (for creating with the correct mode). I'll reuse it here.

src/hook.c
tiennou: Yeah, this thing is 🔥 . I'll try to clean that up, but I still hesitate between `git_buf` and good-ol `char *` for FS.

src/hook.c
tiennou: This one was a first step toward the convenient wrappers, I'll change its visibility when I mind is made up 😉 .

src/hook.c
tiennou: Good catch ! Will fix.

tiennou:
The memory management of the execution callback is an afterthought at best. My first attempt was to pass the callback on each call, but then there's a few times where *we* would need to trigger a hook, but would not know the user's. Hence afterthought.

About the multiple hooks per repo, I had the understanding that we would pretty much prefer clients to use function and callbacks provided by the API. In other words, I came to think of the hook execution responsibility of `libgit2` only as a way to ease the execution of the things in `.git/hooks`, not as a general-purpose mechanism. Then, from that point of view, having a global list of callback makes no sense IMHO — the goal is to execute hooks, which are per-repo anyway.

pks-t:
Yeah, as I said, I've still got no idea how I'd actually want it implemented myself. So this was more or less just a dump of my thoughts on this topic.

src/hook.c
tiennou: I've done the reverse and removed `HOOK_FILE_MODE` instead. I'm still not sure about `S_IXOTH` though, given that the client can do whatever it wants in the executor, they can switch UIDs...

src/hook.c
tiennou: Fixed !

tiennou:
I think the API part is complete. Mostly `TODO`:

- [ ] the memory-management of the executor's payload to tackle. I can either ask the client to provide a "destructor", or maybe calling the executor with some kind of `NULL` hook would be sufficient ? The latter feels hacky though.
- [ ] maybe wire up some hooks internally — those that we can handle at least (like `pre-rebase`). I'll try to come up with a list of those.

src/hook.c
ethomson: Probably going to want an operating system check here, Windows has no concept of executable files.

tiennou:
Yeah, I switched to `git_buf` for everything — after I found out about `git_path` and friends — `char *` is only used for hook names. If I missed some, then that's something I'll fix.
Thanks for pointing out the style guide, I'll do a more though pass for these (likely while I write documentation, which strangely doesn't appear on my `TODO` list 😉 ).

tiennou:
Documentation completed. Memory-management of the payload handled via another callback. Style *should* be correct.

@ethomson I'm not sure about enumerating hooks, given that there's nothing you can do with them (since save/load is gone). It also bothers me that it uses an hardcoded list, but as there might be other non-hooks executables in there, it's been hardcoded. And it would also (with a little struct-uring) be used for internal calling of hooks — my pipe dream would have been to add some functions/macros around the easy ones that would handle converting from `git_oids/refs/whatever` to their string representation.

```
git_hook_execute_pre_rebase(repo, upstream_ref, rebased_ref, NULL);
git_hook_execute_post_checkout_branch(repo, old_head, head);
git_hook_execute_post_checkout_file(repo, head, file);
git_hook_execute_post_checkout_clone(repo, head);
```
etc. so the rest of the library can just insert one of those without doing too much string-wrangling.

I think it would be most useful for the push part, where there's lot of things to be given to the hook through stdin.

include/git2/hook.h
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * This will obey core.hooksPath.
+ *
+ * @param out_dir A pointer to a git_buf that will be filled with the absolute path.
+ * @param repo The repository to get the hooks location from.
+ * @return 0 on success, or an error code.
+ */
+GIT_EXTERN(int) git_hook_dir(git_buf *out_dir, git_repository *repo);
+
+typedef int (*git_hook_foreach_cb)(const char *hook_name, void *payload);
```
pks-t: Dumping my brain here as to what might be simpler to use for application developers. When using a callback like this, application developers are required to explicitly know names of possible hooks in order to act on them (assuming he doesn't simply want to know _all_ hooks available). Furthermore, in case the client needs to inspect the hooks, he'll have to assemble each hook's path manually.

I'd propose a revamped API here in that instead of receiving a hook name, we'll introduce an enum `git_hook_type` or similar, which has an enum value for each existing hook. Now the callback will first receive the enum as well as the actual path to the hook so that the client may inspect the file. E.g. something like this:

```C
typedef int (*git_hook_foreach_cb)(git_hook_type type, const char *hook_path, void *payload);
```

Thoughts?

src/hook.c
```diff
@@ -0,0 +1,216 @@
...
+	if (err == GIT_ENOTFOUND) {
+		git_buf_puts(&tmp_path, repo->path_repository);
+		git_buf_joinpath(&tmp_path, tmp_path.ptr, GIT_HOOKS_DIR);
+	} else if (err == GIT_OK) {
+		git_buf_joinpath(&tmp_path, repo->path_repository, cfg_path.ptr);
+		git_path_resolve_relative(&tmp_path, 0);
+		git_buf_free(&cfg_path);
+	} else {
+		/* XXX: error reporting */
+		return NULL;
```
pks-t: I guess this is a forgotten TODO item :P

src/hook.c
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	*out_dir = tmp_path;
```
pks-t: I'd rather make the folowing instead of doing a struct-copy here:

```C
git_buf_attach(out_dir, git_buf_cstr(&tmp_path), git_buf_len(&tmp_path));
```

You could also consider making the out-pointer point to a C string instead of to a buffer.

src/hook.c
```diff
@@ -0,0 +1,216 @@
...
+	int err;
+	git_buf hook_path = GIT_BUF_INIT;
+
+	assert(hook_name);
+
+	err = git_hook_dir(&hook_path, repo);
+	if (err != 0)
+		return -1;
+
+	git_buf_joinpath(&hook_path, hook_path.ptr, hook_name);
```
pks-t: Error handling is missing here.

src/hook.c
```diff
@@ -0,0 +1,216 @@
...
+		callback(hook_name, payload);
+	}
+
+	return 0;
+}
+
+int git_hook_register_callback(git_repository *repo,
+							   git_hook_execution_cb executor,
+							   git_hook_destructor_cb destructor,
+							   void *payload)
```
pks-t: Indentation seems to be a bit off here.

src/hook.c
```diff
@@ -0,0 +1,216 @@
...
+		goto cleanup;
+
+	while ((arg = va_arg(args, char *))) {
+		if (arg == NULL)
+			break;
+
+		git_vector_insert(&arg_vector, arg);
+	}
+
+	va_end(args);
```
pks-t: You do not need `va_end` here as you already call it in `git_hook_execute` and `git_hook_execute_io`. Otherwise you have only one `va_start` and two calls to `va_end`.

src/repository.c
```diff
@@ -143,6 +143,10 @@ void git_repository_free(git_repository *repo)
 	git__free(repo->ident_name);
 	git__free(repo->ident_email);
 
+	if (repo->hook_payload != NULL && repo->hook_payload_free) {
```
pks-t: Style is a bit inconsistent: either check both against `NULL` or none.

tests/hook/enumerate.c
```diff
@@ -0,0 +1,84 @@
...
+	git_buf_puts(hook_list, hook_name);
+	return 0;
+}
+
+void test_hook_enumerate__foreach_hooks(void)
+{
+	git_buf hook_list;
+	git_buf expected_list;
+	git_buf_init(&hook_list, 0);
+	git_buf_init(&expected_list, 0);
```
pks-t: Use `git_buf buf = GIT_BUF_INIT;` instead. You might also consider using `git_vector` instead of `git_buf`, as it is conceptually more correct.

tests/hook/enumerate.c
```diff
@@ -0,0 +1,84 @@
...
+	/* Reset our hook dir for the next hook */
+	git_buf_rtruncate_at_char(&alt_hook, '/');
+	git_buf_joinpath(&alt_hook, alt_hook.ptr, "post-merge");
+	cl_git_mkfile(git_buf_cstr(&alt_hook), NULL);
+	cl_must_pass(p_chmod(git_buf_cstr(&alt_hook), 0776));
+	git_buf_free(&alt_hook);
+
+	/* Check that we get the correct hooks */
+	git_buf_init(&hook_list, 0);
+	git_buf_init(&expected_list, 0);
```
pks-t: These calls are not necessary as you alread initialized the buffers with `GIT_BUF_INIT`. Same comment as above applies, you might consider changing this to use a `git_vector` instead.

tests/hook/execute.c
```diff
@@ -0,0 +1,99 @@
...
+
+void test_hook_execute__hook_called(void)
+{
+	int hook_called = 0;
+	cl_must_pass(git_hook_register_callback(g_repo, hook_exec_1, NULL, &hook_called));
+	cl_must_pass(git_hook_execute(g_repo, "post-merge", "1", NULL));
+	cl_assert_equal_i_(hook_called, 1, "hook wasn't called");
+}
+
+static int hook_exec_2(git_hook_env env, void *payload)
```
pks-t: I expected this hook callback to consume two arguments. Maybe name the callbacks a bit clearer, e.g. `hook_with_argument_cb` and `hook_with_io_cb` (those names suck, I know, but I guess they're still better than enumerating the callbacks).

tests/hook/execute.c
```diff
@@ -0,0 +1,99 @@
...
+static void hook_exec_destruct(void *payload)
+{
+	char **malloc_str = (char **)payload;
+	git__free(*malloc_str);
+	git__free(malloc_str);
+
+	destruct_called = 1;
+}
+
+static int hook_exec_cleanup(git_hook_env env, void *payload)
```
pks-t: Isn't this the executor callback? `cleanup` is a bit misleading here.

tiennou:
Don't worry about the enum-vs-string uncertainty, it's shared 😉 .

My rationale for keeping strings around was that it allows the executor to be hook-agnostic, which decouples the API — eg. let's say `git` deprecates some hook, you're now stuck with your enum value. I think that some of the higher-level things I wanted to handle actually require something like a list of known, "automatic" hooks, but keeping the "lower level" unaware of a list is better.

As you say, this forces the client application to be aware of which hook *they* actually need to call (things like the `*commit` hooks), but some of those calls can be completely "hidden" away.

Also, another thing I thought about but forgot to ask is `SIGPIPE`. I think we should delegate that to the client (because signal handlers are messy), but this isn't clearly stated in the docs yet.

Thanks for the review, I'll try to find some time to handle your comments tonight.

include/git2/hook.h
```diff
@@ -0,0 +1,124 @@
...
+ * A hook environment
+ *
+ * This structure will be provided by the library when a hook needs to
+ * be executed.
+ *
+ * - `path` is the absolute path to the hook executable.
+ * - `args` is the argument list for the hook.
+ * - `io` is either an in/out pointer to either data that must be provided to
+ *        the hook, a pointer where the executor will stash data outputted by
+ *        the hook, or both.
```
tiennou: I'm utterly at a loss about the syntax to use for documentation of typedefs. After a quick look around I can see that most use inline `/**` comments, some use a preceding `/**` comment with `*` or `-`, but it seems the doc parser ignores `*` ([case](https://github.com/libgit2/libgit2/blob/HEAD/include/git2/repository.h#L278-287) in [point](https://libgit2.github.com/libgit2/#HEAD/type/git_repository_init_options)).

In a few words, it's supposed to contain whatever input the client will have to feed the hook when the executor runs, and you can set it to some buffer on return to hand back data to whatever called the hook. It's a little fuzzy because right now the only hooks that (could) make use of this facility are `pre-push`,  `pre-receive`, `update`, `post-receive`, `post-update`, but most of these are only used on the remote side, and I don't think `libgit2` supports being used as a server (yet).

That wasn't a few words 😟.

include/git2/hook.h
```diff
@@ -0,0 +1,124 @@
...
+typedef struct {
+	char *path;
+	git_strarray args;
+	git_buf *io;
+} git_hook_env;
+
+typedef void (*git_hook_destructor_cb)(void *payload);
+
+typedef int (*git_hook_execution_cb)(
+	git_hook_env env,
```
tiennou: Fixed.

src/hook.c
tiennou: IIRC, I wasn't sure under which `GITERR_*` I should report it. I was tempted to create a `GITERR_HOOK`, but since the rest of the API isn't yet complete, I wasn't sure it would be used somewhere else.

src/hook.c
tiennou: Fixed.

tests/hook/execute.c
tiennou: Fixed.

src/repository.h
tiennou: This is fixed. `git_hook_register` now has a destructor callback.

include/git2/hook.h
tiennou: `git_hook_foreach` just allows the client to know the list of available hooks in a given repository, which is why you don't get a path here. I don't really see a use case for that — normally, the client just blindly calls `git_hook_execute()`, which will cause the registered executor to be called with the required parameters if the hook exists and is valid. In other words, the client doesn't need to know whether the hook is available or not.

My thought would be to replace all that with a `git_hook_exists(git_repository *repo, const char *hook_name)`, but I have no idea if it's worth it...

src/hook.c
pks-t: I would classify this as a configuration error, so I'd return `GITERR_CONFIG`. But yeah, it's not that obvious and you can certainly find arguments for having a `GITERR_HOOK` instead.

pks-t:
Hummm... I think we shouldn't actually worry about git.git deprecating certain hooks. For one, I think it rather unlikely that they'd deprecate hooks as there will always be use-cases which are broken by removing a hook. And furthermore, git.git always pays strong attention not to remove existing features.

And even in the case that git.git _does_ deprecate a hook I think there's a big benefit in having an enum here. If the client will just compare strings, he cannot easily know when a hook has been deprecated and may fail silently. But in the case of an enum which subsequently got removed due to deprecation, the client will fail to compile. While I agree it is not a nice thing to break API I deem it better still than causing a client to silently fail.

tests/clar_libgit2.h
```diff
@@ -97,6 +98,23 @@ GIT_INLINE(void) clar__assert_equal_oid(
...
 		"OID mismatch: " #one " != " #two, (one), (two))
 
+GIT_INLINE(void) clar__assert_equal_vector_str(
+	const char *file, int line,
+	const git_vector *a, const git_vector *b)
+{
+	clar__assert_equal(file, line, "Vector length mismatch: %" PRIuZ " != %" PRIuZ, 1, PRIuZ, git_vector_length(a), git_vector_length(b));
+
+	size_t i;
+	char *string;
```
pks-t: This is illegal in C90. Declarations have to come before all statements.

pks-t: And by the way, there's some compilation error on Windows with MSVC here. Seems to be related to `size_t`, but I didn't really take a look into it.

tests/hook/execute.c
```diff
@@ -0,0 +1,102 @@
...
+	destruct_called = 1;
+}
+
+void test_hook_execute__frees_payload_with_repo(void)
+{
+	git_repository *repo;
+	const char * test_str = "a test string";
+	char **malloc_str = git__malloc(sizeof(*malloc_str));
+	*malloc_str = git__calloc(strlen(test_str), sizeof(*malloc_str));
+	strcpy(*malloc_str, test_str);
```
pks-t: You could just have a normal `char *malloc_str` and pass that to the callback. No need to allocate a pointer to a string here. The callback would then directly get the string and cast it to `char *`.

include/git2/hook.h
pks-t: Okay. Wasn't clear to me how this would be used

include/git2/hook.h
pks-t: Yeah. I'm no that familiar with Doxygen, I just searched for examples in our codebase any by chance found mostly `*` as bullet points.

tests/hook/execute.c
tiennou: Yeah, I seem to remember having a hard time *not* freeing a stack-allocated string, hence the mess. I'll try to clean that up.

tiennou: Actually, looking at it, I'm not even sure what's the point of building something to pass to the destructor. The test is only concerned with making sure its called, freeing things is not its responsibility.

tests/hook/enumerate.c
```diff
@@ -0,0 +1,83 @@
...
+	char *alt_hook_str;
+	git_vector hook_list = GIT_VECTOR_INIT;
+	git_vector expected_list = GIT_VECTOR_INIT;
+
+	/* Change the repository hooks  */
+	cl_git_pass(git_repository_config(&cfg, g_repo));
+	cl_git_pass(git_config_set_string(cfg, "core.hooksPath", ALT_HOOK_DIR));
+
+	cl_git_pass(git_hook_dir(&alt_hook_str, g_repo));
+	cl_git_pass(git_buf_sets(&alt_hook, alt_hook_str));
```
pks-t: You're leaking `alt_hook_str`. You probably want `git_buf_attach(&alt_hook, alt_hook_str, strlen(alt_hook_str) + 1)` here, which will also free the string later on.

src/hook.c
tiennou: I've taken a closer look at it, and as you point out it means something is 🔥 in config, so failing in that case is OK. I'm actually not handling a non-existent `.git/hooks` directory, or a non-existent core.hooksPath for that matter, so I'll have to check what git does if that happens.

tiennou:
I've added `sys/hook.h`, with the goal of having the "higher-level" versions of hooks there, as a bunch of `git_hook_call_*` functions that the client can use to trigger hook execution (a missing hook is ignored, but it could be changed to `GIT_ENOTFOUND` if that's better). Now that I think about it, I feel like the current headers should be reversed — `hook.h` contains the nice, supported versions, and `sys/hook.h` has the low-level `git_hook_execute` function and friends...

I hooked up (pun intended) `pre-rebase` and added a quick test to see how it works and check that a hook can prevent a rebase. I'm completely abusing the executor to do that but well...

As for the enum vs strings discussion, providing those higher-level functions make the enum useless from the executor standpoint. As I see it, it's only concerned with executing the given hook, with the given arguments and optional input.

If you (still) like where this is going, I'll add more of the high-level functions. For the record, here is my current FIXME comment 😉 :

```
/**
 * FIXME: Unsupported hooks :
 *
 * - applypatch-msg, pre-applypatch, post-applypatch: no git-am.
 * - pre-commit, prepare-commit-msg, commit-msg, post-commit: our "commit workflow"
 * doesn't match git.
 * - post-checkout: the "normal checkout" looks like a "no matching workflow" too,
 * as our checkout doesn't work on references, but directly on HEAD/index/tree.
 * The "clone checkout" looks ok though.
 * - post-merge: no git-pull.
 * - pre-push: looks ok-ish.
 * - pre-receive, update, post-receive, post-update, push-to-checkout:
 * no git-receive-pack.
 * - pre-auto-gc: no git-gc.
 * - post-rewrite: looks ok-ish.
 */
```

tiennou:
Closing in favor of #4620.

PR #3827, opened 2016-06-19
HTTP authentication scheme name is case insensitive.

dbrnz:
From RFC 2617 (my **_emphasis_**):

> 1.2 Access Authentication Framework
>
> HTTP provides a simple challenge-response authentication mechanism that MAY be used by a server to challenge a client request and by a client to provide authentication information. It uses an extensible, **_case-insensitive_** token to identify the authentication scheme, followed by a comma-separated list of attribute-value pairs which carry the parameters necessary for achieving authentication via that scheme.

### DIFF
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -114,7 +114,7 @@ static bool challenge_match(git_http_auth_scheme *scheme, void *data)
 	size_t scheme_len;
 
 	scheme_len = strlen(scheme_name);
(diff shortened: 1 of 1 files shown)


ethomson:
Hahaha, oooooops.  Thanks @dbrnz for catching and fixing this.

Also, apologies to @jeffhostetler for not reading your RFC very carefully.  :D

PR #3828, opened 2016-06-20
Fix include.path stopping on non-existent file relative to the home directory

### LINKED ISSUE #3681: include.path with non-existing file
I have my `~/.gitconfig` set up as a symlink (shared among my terminals) and with content like

```
[include]
    path                = ~/.gitconfig_pre

..... main config goes here ....

[include]
    path                = ~/.gitconfig_extra
```

This allows me to have local modifications, specific to the terminal.
But on some terminals, I don't have any modifications, so those include paths are missing.
libgit2 seems to crash in that scenario and returns no global configurations.

meatcoder:
This is an attempt to fix #3681. While looking at that issue, I was able to determine that parsing a config file stops, when it attempts to include a missing config file relative to the home directory using the following syntax:

```
[include]
path = ~/.nonexistent
```

I was able to demonstrate this by adding a new test that fails when using the above. I therefore committed and push a fix to allow parsing to continue, even if the file referenced does not exist. 

Note that including a non-existent file relative to the search paths or using absolute paths, currently work without any issues. The fix only addresses the case for missing files relatives to the home directory.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -1250,8 +1250,16 @@ static int strip_comments(char *line, int in_quotes)
 static int included_path(git_buf *out, const char *dir, const char *path)
 {
 	/* From the user's home */
-	if (path[0] == '~' && path[1] == '/')
-		return git_sysdir_find_global_file(out, &path[1]);
+	int result;
+	if (path[0] == '~' && path[1] == '/') {
+		result = git_sysdir_find_global_file(out, &path[1]);
+		if (result == GIT_ENOTFOUND) {
+			git_buf_sets(out, &path[1]);
+			return 0;
+		}
+
+		return result;
+	}
 
 	return git_path_join_unrooted(out, path, dir, NULL);
 }
--- a/tests/config/include.c
+++ b/tests/config/include.c
@@ -108,6 +108,23 @@ void test_config_include__missing(void)
 	git_config_free(cfg);
 }
 
+void test_config_include__missing_homedir(void)
+{
+	git_config *cfg;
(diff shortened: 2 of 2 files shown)


ethomson:
Awesome - thanks for tackling this.  You're cleaning up the error perfectly - I wonder though if we shouldn't move this block out of `included_path` and into the caller (`read_on_variable` in `config_file.c:1613` or so).

Right now we'll actually fall back to looking for a file literally named `~/.nonexistent` - that is to say if there is no `.nonexistent` file in `$HOME` then we will look for a file named `.nonexistent` in a folder named `~` relative to the config file being included.  A quick test shows that git does _not_ do this, it merely translates `~` to `$HOME` and does not try to fall back to treating it like a literal path.

```
% tail -2 .git/config
[include]
	path = ~/whee
% ls -Flas .git/\~/whee  
4 -rw-r--r--  1 ethomson  wheel  17 Jun 20 10:13 .git/~/whee
% cat .git/\~/whee
[foo]
	bar = baz
% git config foo.bar
```

I feel like we should emulate git's behavior here, even though it will likely never cause a problem.  What do you think?

Thanks for the great unit test, also.

meatcoder:
@ethomson Thanks again for looking at this. Since this involves parsing config files, which should be compatible with git anyway, it might be best to mimic its behavior as closely as possible. I'll try to revise this PR to do that.

ethomson:
This ended up being a bit more challenging than just moving some of the error handling logic!  I rebased this on `master` and made a few additional changes in #4179.

Thanks for working on this!

Issue #3825, opened 2016-06-17
Segmentation fault on MSYS2


eidheim:
Compiling the following program causes segmentation fault on MSYS2:
main.cc:
```c++
#include <git2.h>
#include "someclass.h"

std::mutex mutex; //removing this line fixes the segmentation fault

int main() {
  git_libgit2_init();
  
  {
    std::unique_lock<std::mutex> lock(SomeClass::mutex); //removing this line also fixes the segmentation fault
  }
  
  git_repository *repository_ptr;
  git_repository_open_ext(&repository_ptr, ".", 0, NULL);
  
  git_repository_free(repository_ptr);
}
```

someclass.h:
```c++
#ifndef SOMECLASS_H_
#define SOMECLASS_H_
#include <mutex>

class SomeClass {
public:
  static std::mutex mutex;
};
#endif //SOMECLASS_H_
```

someclass.cc:
```c++
#include "someclass.h"

std::mutex SomeClass::mutex;
```

Compile with:
```sh
g++ -std=c++11 main.cc someclass.cc -lgit2
```

Gdb backtrace:
```
(gdb) run
Starting program: C:\msys\msys64\home\ole\libgit2-test\a.exe
[New Thread 1036.0x438]
[New Thread 1036.0x5a8]
[New Thread 1036.0xa6c]

Thread 1 received signal SIGSEGV, Segmentation fault.
0x0000000077668d84 in ntdll!RtlDeNormalizeProcessParams ()
   from C:\Windows\SYSTEM32\ntdll.dll
(gdb) bt
#0  0x0000000077668d84 in ntdll!RtlDeNormalizeProcessParams ()
   from C:\Windows\SYSTEM32\ntdll.dll
#1  0x0000000077668cb4 in ntdll!RtlDeNormalizeProcessParams ()
   from C:\Windows\SYSTEM32\ntdll.dll
#2  0x000000006cb633aa in ?? () from C:\msys\msys64\mingw64\bin\libgit2.dll
#3  0x000000000040164d in __gthread_mutex_lock (
    __mutex=0x408040 <SomeClass::mutex>)
    at C:/msys/msys64/mingw64/include/c++/5.4.0/x86_64-w64-mingw32/bits/gthr-default.h:748
#4  0x00000000004031c8 in std::mutex::lock (this=0x408040 <SomeClass::mutex>)
    at C:/msys/msys64/mingw64/include/c++/5.4.0/mutex:135
[...]

PR #3830, opened 2016-06-20
Thread namespacing

pks-t:
This PR aims to fix the namespacing-issue in issue #3825. The Windows-specific pthreads.c source file tries to re-implement the API exposed by pthreads by implementing many of the pthread-symbols with Windows-specific threading code. This unfortunately results in runtime errors when a user links his application to both libgit2 and winpthreads, causing one of both libraries to override symbols of the other library, causing inconsistencies.

Fix the issue by introducing proper namespacing to our threading API. Instead of trying to re-implement all functions, implement functions inside the `git_` namespace and adjust the `#define`s in Unix-specific code.

This solves the part where linking against both libraries causes runtime errors. What is still missing is to allow Windows-users to link libgit2 not against the internal thread API but against winpthread provided by MinGW/MSYS2. This will follow later either by updating this PR or creating a new one if this one got merged already.

### DIFF
--- a/src/unix/pthread.h
+++ b/src/unix/pthread.h
@@ -24,4 +24,12 @@ typedef struct {
 #define git_mutex_unlock(a)     pthread_mutex_unlock(a)
 #define git_mutex_free(a)	pthread_mutex_destroy(a)
 
+/* Git condition vars */
+#define git_cond pthread_cond_t
+#define git_cond_init(c)	pthread_cond_init(c, NULL)
+#define git_cond_free(c) 	pthread_cond_destroy(c)
+#define git_cond_wait(c, l)	pthread_cond_wait(c, l)
+#define git_cond_signal(c)	pthread_cond_signal(c)
+#define git_cond_broadcast(c)	pthread_cond_broadcast(c)
+
 #endif /* INCLUDE_unix_pthread_h__ */
--- a/src/unix/pthread.h
+++ b/src/unix/pthread.h
@@ -12,6 +12,7 @@ typedef struct {
 	pthread_t thread;
 } git_thread;
 
+#define git_threads_init() (void)0
 #define git_thread_create(git_thread_ptr, start_routine, arg) \
 	pthread_create(&(git_thread_ptr)->thread, NULL, start_routine, arg)
 #define git_thread_join(git_thread_ptr, status) \
--- a/src/unix/pthread.h
+++ b/src/unix/pthread.h
@@ -1,0 +1,20 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#ifndef INCLUDE_unix_pthread_h__
+#define INCLUDE_unix_pthread_h__
+
+typedef struct {
+	pthread_t thread;
+} git_thread;
+
+#define git_thread_create(git_thread_ptr, start_routine, arg) \
+	pthread_create(&(git_thread_ptr)->thread, NULL, start_routine, arg)
+#define git_thread_join(git_thread_ptr, status) \
+	pthread_join((git_thread_ptr)->thread, status)
+
+#endif /* INCLUDE_unix_pthread_h__ */
--- a/src/unix/pthread.h
+++ b/src/unix/pthread.h
@@ -17,4 +17,11 @@ typedef struct {
 #define git_thread_join(git_thread_ptr, status) \
 	pthread_join((git_thread_ptr)->thread, status)
 
+/* Git Mutex */
+#define git_mutex pthread_mutex_t
+#define git_mutex_init(a)	pthread_mutex_init(a, NULL)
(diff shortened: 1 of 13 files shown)


src/win32/thread.h
```diff
@@ -0,0 +1,67 @@
...
+typedef struct { void *Ptr; } GIT_SRWLOCK;
+
+typedef struct {
+	union {
+		GIT_SRWLOCK srwl;
+		CRITICAL_SECTION csec;
+	} native;
+} git_rwlock;
+
+#define PTHREAD_MUTEX_INITIALIZER  {(void*)-1}
```
ethomson: I was going to suggest we rename this, but I think it's wholly unused so perhaps we should just remove it.

src/unix/pthread.h
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#ifndef GIT_PTHREAD_H
+#define GIT_PTHREAD_H
```
ethomson: minor nit: our header guards are of the form `INCLUDE_unix_pthread_h__`.

I don't know the history here but we are reasonably consistent 😀

src/win32/thread.h
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#ifndef GIT_THREAD_H
```
ethomson: Similarly, this guard should probably be `INCLUDE_win32_thread_h__`

src/win32/thread.h
pks-t: Agreed, will remove.

src/unix/pthread.h
pks-t: Heh, I've seen it being used like this in `win32/pthread.h` and just re-used it, assuming it was being done like this for internal headers while externally visible headers used the `INCLUDE_` prefix. Will change.

src/win32/thread.h
```diff
@@ -0,0 +1,67 @@
...
+int git_rwlock_init(git_rwlock *GIT_RESTRICT lock);
+int git_rwlock_rdlock(git_rwlock *);
+int git_rwlock_rdunlock(git_rwlock *);
+int git_rwlock_wrlock(git_rwlock *);
+int git_rwlock_wrunlock(git_rwlock *);
+int git_rwlock_free(git_rwlock *);
+
+int git_num_processors(void);
+
+extern int win32_pthread_initialize(void);
```
ethomson: We should probably rename this.  Something like `git_thread_win32_init`?

src/win32/thread.c
```diff
@@ -240,14 +215,26 @@ int pthread_rwlock_wrunlock(pthread_rwlock_t *lock)
...
-int pthread_rwlock_destroy(pthread_rwlock_t *lock)
+int git_rwlock_free(git_rwlock *lock)
 {
 	if (!win32_srwlock_initialize)
 		DeleteCriticalSection(&lock->native.csec);
 	git__memzero(lock, sizeof(*lock));
 	return 0;
 }
 
+int git_num_processors(void)
```
ethomson: I don't see where this is getting called.  Can we drop it if it's unused?

ethomson:
Thanks for taking care of this.  I made a couple of quite trivial comments, but this is great.  😀

src/win32/thread.h
pks-t: Maybe even go all the way and rename it to `git_thread_init`, having a noop `#define git_thread_init` in `unix/pthread.h`? But maybe I'm just overthinking it, dunno.

src/win32/thread.c
pks-t: Yes, it is currently unused, as are some other functions (e.g. `git_cond_broadcast`, which is even unimplemented for Windows as of now). I'm all in for removing unused cruft

src/win32/thread.h
ethomson: Hah, I almost said the same thing!  I didn't suggest it because I worry that I might think that `git_thread_init` initializes a thread and not the whole threading subsystem.  But I'm perfectly happy with either approach.

src/win32/thread.c
ethomson: Aha, yes, it seems like we now have [`git_online_cpus`](https://github.com/libgit2/libgit2/blob/77394a27af283b366fa8bb444d29670131bfa104/src/thread-utils.c#L32) and so this is probably unused.  Let's definitely get rid of this one.  I'll let you use your discretion on the others.

src/win32/thread.h
pks-t: Yeah, I actually had the same thought on the `git_thread_init` problem. Maybe it would be more obvious to name it `git_threads_init`, even though the difference in plurality may be still be easy to skip.

pks-t:
New version fixes comments by @ethomson

ethomson:
Thanks again, this is great.

PR #3831, opened 2016-06-20
Improve the README

ethomson:
After @cmrberry 's awesome [OSCON talk about supporting Open Source projects](https://speakerdeck.com/cmrberry/from-behind-the-contact-form-tips-from-github-support-on-helping-your-community), I wanted to make our "Getting Help" section more prominent, especially now that we have an awesome slack channel that nobody knows about.

In addition:
* Call out our `easy fix` and `up for grabs` tagged items.
* Drop the outdated / abandoned bindings
* Clarify the linking exception slightly

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,18 +15,30 @@ with any kind of software without having to release its source code.
 Additionally, the example code has been released to the public domain (see the
 [separate license](examples/COPYING) for more information).
 
-* Website: [libgit2.github.com](http://libgit2.github.com)
-* StackOverflow Tag: [libgit2](http://stackoverflow.com/questions/tagged/libgit2)
-* Issues: [GitHub Issues](https://github.com/libgit2/libgit2/issues) (Right here!)
-* API documentation: <http://libgit2.github.com/libgit2/>
-* IRC: [#libgit2](irc://irc.freenode.net/libgit2) on irc.freenode.net.
-* Mailing list: The libgit2 mailing list was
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,18 +15,30 @@ with any kind of software without having to release its source code.
...
-    on GitHub for bug reports.  The mailing list archives are still available at
-    <http://librelist.com/browser/libgit2/>.
+Getting Help
+============
 
+**Join us on Slack**
+
+Visit [slack.libgit2.org](http://slack.libgit2.org/) to sign up, then join
+us in `#libgit2`.  If you prefer IRC, you can also point your client to our
+slack channel once you've registered.
```
pks-t: What about #libgit2 on the Freenode channel? Is it now officially superseded by Slack?

ethomson: I think superseded is a good way to describe it.  There are a couple of really nice things about slack:  first, we can have a room for each of the bindings and have a nice place to chat amongst each other, since they're mostly trying to solve the same problems.  As a result, we can get rid of several overlapping places to chat - getting more people into one place will, I think, be good for easier conversations and cross-pollination.  Second, we can get some easier integrations and has the potential for some chatops workflows which I think could be really neat.

That's my thinking, anyway.  I hope that overall this means that more people have _fewer_ places to communicate with each other (since many contributors already use slack) but I realize that a number of people are not using slack at all and so this could increase the cognitive load for them.  (I apologize to those people.)

PR #3833, opened 2016-06-21
cmake: do not use -fPIC for MSYS2

pks-t:
The MSYS2 build system automatically compiles all code with position-independent
code. When we manually add the -fPIC flag to the compiler flags, MSYS2 will
loudly complain about PIC being the default and thus not required.

Fix the annoyance by stripping -fPIC in MSYS2 enviroments like it is already
done for MinGW.


ethomson:
Neat.  Thanks!

PR #3835, opened 2016-06-21
cmake: use pthreads library on win32 if available

pks-t:
MSYS2 and MinGW both provide pthread libraries. Instead of
re-implementing a pthreads-like interface with Windows APIs, just
use libgit2's POSIX threads abstraction and link against
pthreads.


carlosmn:
I don't think I like this change. We have a pthreads-like interface that we use because the core team is mostly on unices, but part of the way we build libgit2 is to make use of the system's capabilities as much as possible.

This moves us away from that, and using something that pretends to be pthreads but is surely just calling the Win32 APIs anyway.

pks-t:
Well, I'm not too passionate about this PR. So I don't care if it gets pulled or discarded.

I agree that we want to leverage the platforms' capabilities as much as possible. But in my eyes it is debatable what "platform capabilities" in the case of Windows really are. Is it the API the base system provides? Or is it instead the API provided by the environment I'm basing the build on? With this PR, we will still use the native Windows API when compiling with MSVC. But as soon as we have another platform, that is MSYS or MinGW, which provide different APIs, I don't think it is a bad thing to use what they provide instead of working around them and using the Windows API.

In my opinion it makes sense to use the build environment's capabilities instead of the operating system's capabilities, even if these build environments will eventually call out to the operating system. In my eyes, this increases consistency, as other applications using that build environment are likely to use the build environment's capabilities as well.

ethomson:
I have literally no idea what msys/mingw/cygwin users expect here, or if they care at all.  My impression is that this is probably an implementation detail - I don't know which is a better default, if at all.

/cc @tkelman

tkelman:
The Windows pthread implementation generally calls through to the same (maybe not exactly, would have to read the implementation and compare with the win32 thread calls here) underlying OS API's. Not every flavor of how you built mingw can use pthreads, which threading model to use is a configuration option for how you build the toolchain.

pks-t:
Well, yes. The PR should in fact accout for the scenario when mingw is built without pthreads as in this case, the FindThreads.cmake module shouldn't set `CMAKE_USE_PTHREADS_INIT`.

If we are unsure about this we might let the user choose for himself. But on the other hand I don't know if I'd like to introduce more complexity in the build system when we could instead just drop the PR and use our own internal threads interface.

tkelman:
Doesn't setting it to `REQUIRED` mean cmake will error if it isn't found?

pks-t:
No, it doesn't. FindThreads.cmake does not explicitly search for _pthreads_ but instead for multiple threading libraries available to the system, setting a specific variable depending on which threading library has been deemed as the one to be used. Threading libraries searched for are the following:

CMAKE_THREAD_LIBS_INIT     - the thread library
CMAKE_USE_SPROC_INIT       - are we using sproc?
CMAKE_USE_WIN32_THREADS_INIT - using WIN32 threads?
CMAKE_USE_PTHREADS_INIT    - are we using pthreads
CMAKE_HP_PTHREADS_INIT     - are we using hp pthreads

If none of these libraries are found, FindThreads will indeed fail. But as we only execute `FIND_PACKAGE(Threads REQUIRED)` when `THREADSAFE=ON`, this is in fact what we want.

carlosmn:
The question here is I think why would the application care what kind of threading wrapper we end up using? If they ask for a threaded operation they'll get it, otherwise all we're doing is storing thread-local data.

pks-t:
Okay. I'll just close this, as I don't really care anyway.

PR #3837, opened 2016-06-25
Support index v4

novalis:
moved from https://github.com/libgit2/libgit2/pull/3789 because I no longer have access to the twitter github account.

This version leaves one in/out parameter, but I think otherwise addresses the concerns in the prev request.

### DIFF
--- a/include/git2/index.h
+++ b/include/git2/index.h
@@ -252,6 +252,31 @@ GIT_EXTERN(int) git_index_caps(const git_index *index);
 GIT_EXTERN(int) git_index_set_caps(git_index *index, int caps);
 
 /**
+ * Get index on-disk version.
+ *
+ * Valid return values are 2, 3, or 4.  If 3 is returned, an index
+ * with version 2 may be written instead, if the extension data in
+ * version 3 is not necessary.
+ *
+ * @param index An existing index object
+ * @return the index version
+ */
+GIT_EXTERN(unsigned int) git_index_version(git_index *index);
+
+/**
+ * Set index on-disk version.
+ *
+ * Valid values are 2, 3, or 4.  If 2 is given, git_index_write may
+ * write an index with version 3 instead, if necessary to accurately
+ * represent the index.
+ *
+ * @param index An existing index object
+ * @param version The new version number
+ * @return 0 on success, -1 on failure
+ */
+GIT_EXTERN(int) git_index_set_version(git_index *index, unsigned int version);
+
+/**
  * Update the contents of an existing index object in memory by reading
  * from the hard disk.
  *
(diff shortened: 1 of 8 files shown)


src/varint.c
```diff
@@ -0,0 +1,41 @@
...
+			return 0; /* overflow */
+		}
+		c = *buf++;
+		val = (val << 7) + (c & 127);
+	}
+	*varint_len = buf - bufp;
+	return val;
+}
+
+int git_encode_varint(unsigned char *buf, uintmax_t value)
```
pks-t: In my opinion it would be a cleaner interface design if we have the caller pass in the buffer's maximum length so we do not overflow it. Otherwise it's somewhat magical to the caller how big the buffer is required to be.

tests/core/encoding.c
```diff
@@ -0,0 +1,33 @@
...
+	cl_assert(size == 1);
+
+	buf = (unsigned char *)"\xfe\xdc\xbaXY";
+	cl_assert(git_decode_varint(buf, &size) == 267869656);
+	cl_assert(size == 4);
+
+	buf = (unsigned char *)"\xaa\xaa\xfe\xdc\xbaXY";
+	cl_assert(git_decode_varint(buf, &size) == 1489279344088ULL);
+	cl_assert(size == 6);
+
```
pks-t: For completeness, I'd also add a test where parsing actually fails so that `size == 0`.

include/git2/index.h
```diff
@@ -252,6 +252,31 @@ GIT_EXTERN(int) git_index_caps(const git_index *index);
...
+ * @return the index version
+ */
+GIT_EXTERN(unsigned int) git_index_version(git_index *index);
+
+/**
+ * Set index on-disk version.
+ *
+ * Valid values are 2, 3, or 4.  If 2 is given, an index with version
+ * 3 may be written instead, if necessary to accurately represent the
+ * index.
```
carlosmn: We should mention this is the version that a `git_index_write()` will use to write to disk. The part about 2 vs 3 reads as though this would write out the index.

novalis:
Pretty sure these test failures are spurious and that this patch is ready to go.

pks-t:
I've restarted these jobs. But you'll need to rebase the branch upon current master, as your changes now create conflicts.

novalis:
Conflicts fixed, tests pass.

novalis:
@pks-t I keep having to rebase to deal with ChangeLog conflicts.  All of the substantive issues raised in review have been solved. CI passes.  Can you (or someone else) please merge this?

pks-t:
I don't have enough knowledge on the index format code, so I can't really comment on or merge this PR despite stylistic issues.

novalis:
@ethomson can you merge this?

ethomson:
@novalis Sure - thanks for hacking on this!

PR #3838, opened 2016-06-26
revwalk: introduce tests that hide old commits

ethomson:
Introduce some tests that show some commits, while hiding some commits that have a timestamp older than the common ancestors of these two commits.

For example, I've introduced 43e968a and 8e73b76 with timestamps from 2010 (earlier than 5b5b025), and bd75801 with a timestamp from today:

```
* bd75801 new change
| *   8e73b76 Merge branch 'subtrees' into old
| |\  
| |/  
|/|   
* | 763d71a Add some files into subdirectories
* | c47800c branch commit one
* | 5b5b025 another commit
| * 43e968a old
|/  
* 8496071 testing
```

Note that they have an obvious common ancestor of 763d71a, and:

```
% git rev-list 8e73b76..bd75801 
bd758010071961f28336333bc41e9c64c9a64866
```

However, when hiding 8e73b76 and showing bd75801, our revwalk does not hide 8e73b76's ancestors.  Ours returns:

```
bd758010071961f28336333bc41e9c64c9a64866
763d71aadf09a7951596c9746c024e7eece7c7af
c47800c7266a2be04c571c04d5a6614691ea99bd
5b5b025afb0b4c913b4c338a42934a3863bf3644
8496071c1b46c854b31185ea97743be6a8774479
```

As best I can tell, this is because we're using a time-based priority queue in `premark_uninteresting`, which premarks the (timewise) new commit and all its parents, deciding that they're interesting, before dealing with the (timewise) old commit, and the common ancestors that _should_ be uninteresting are not.

I spend some time poking around here but I'm afraid that I need to defer to the mad revwalk skills.

/cc @carlosmn


carlosmn:
The priority queue is only supposed to make the order of the traversal more likely to be efficient and should not dictate _when_ we stop looking for the merge bases. This is indeed troubling.

carlosmn:
One of the issues was a reversed logic. The second one seems to be that we do in fact perform the marking until we run out of interesting commits, but that means we never get to the uninterested commit that got pushed and thus never end up hiding it.

It looks like we might have to make sure to go and mark commits uninteresting, but it's starting to look a lot like special-casing so I'm reluctant to just add another loop to `premark_uninteresting()` (and also my plane landed before I could experiment more with that).

carlosmn:
Both of these revwalk tests are now passing (CI failure comes down to the extra objects this adds). But I would still like to make sure this isn't just because we've ended up special-casing this exact situation. The main premark loop might still need a change.

carlosmn:
Fixed via #3921

PR #3839, opened 2016-06-26
WIP: respect index during checkout

ethomson:
Our current checkout implementation looks to see if the working directory is dirty to determine whether it's safe to checkout a given file.  It _does not_ look at the index, however.

This is fine if we're checking out the repository's index itself, but if we're checking out a `git_tree` then we could get into a place where the repository's index has modifications, but the workdir itself has been reverted.  We would obliterate those staged changes.

This adds checking for changes in the index before checking out (raising conflicts as appropriate otherwise) when a working directory file exists.  This does not yet cover other cases (for example, when an entry exists in the index but not in the working directory.)

I think that we can simply change the `with_wd` case to a `with_wd_or_index` case and use the existing `no_wd` case to be `no_wd_or_index`.


tests/checkout/tree.c
```diff
@@ -1479,3 +1479,75 @@ void test_checkout_tree__baseline_is_empty_when_no_index(void)
...
+	assert_on_branch(g_repo, "master");
+	cl_git_pass(git_repository_index(&index, g_repo));
+	cl_git_pass(git_repository_head(&head, g_repo));
+	cl_git_pass(git_reference_peel(&obj, head, GIT_OBJ_COMMIT));
+
+	cl_git_pass(git_reset(g_repo, obj, GIT_RESET_HARD, NULL));
+
+	cl_git_pass(git_status_list_new(&status, g_repo, NULL));
+	cl_assert_equal_i(0, git_status_list_entrycount(status));
+	git_status_list_free(status);
```
pks-t: It's the third test now where I see this pattern, checking that our repository is actually in a clean state (that is, asserting the branch and then asserting that no modifications are present). Maybe put this into a function/macro? I bet it would be useful to other tests.

cjhoward92:
Are there any plans to continue this work? It seems like libgit2 does not properly manage file mode only changes. If I try to force checkout a file with only file mode changes nothing happens, as I believe libgit2 is only checking content, which has not changed. Can't say for sure, but it looks like you fixed that in this WIP.

ethomson:
Thanks @cjhoward92 for the ping.  I'll bring forward those fixes in a separate PR.

cjhoward92:
You rock!

pks-t:
@ethomson: can this be closed? I feel like you've already committed most changes in here to master

Issue #3793, opened 2016-05-19
Git blame possible double free exception


ameily:
I'm writing a tool that iterates over a repository's commit history and performs a blame on an entire tree for each iteration. I'm seeing a double free issue within libgit2 and am having difficulty tracing the problem down. I've confirmed that the problem exist in both `master` and `v0.24.1`. The line numbers below are from the `v0.24.1` tag.

```
*** Error in `/home/user/repos/git-stock/build/git-stock': double free or corruption (!prev): 0x000000000f3062a0 ***


#0  0x00007ffff6aa7418 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff6aa901a in __GI_abort () at abort.c:89
#2  0x00007ffff6ae972a in __libc_message (do_abort=do_abort@entry=2, 
    fmt=fmt@entry=0x7ffff6c026b0 "*** Error in `%s': %s: 0x%s ***\n")
    at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007ffff6af1f4a in malloc_printerr (ar_ptr=<optimized out>, 
    ptr=<optimized out>, str=0x7ffff6c027e0 "double free or corruption (!prev)", 
    action=3) at malloc.c:5007
#4  _int_free (av=<optimized out>, p=<optimized out>, have_lock=0) at malloc.c:3868
#5  0x00007ffff6af5abc in __GI___libc_free (mem=<optimized out>) at malloc.c:2969
#6  0x00007ffff7a9d3e1 in git__free (ptr=0xf3062a0)
    at /home/user/repos/libgit2/src/util.h:246
#7  0x00007ffff7a9d7d2 in git_odb_object__free (object=0xe6aac90)
    at /home/user/repos/libgit2/src/odb.c:105
#8  0x00007ffff7b2ab1c in git_cached_obj_decref (_obj=0xe6aac90)
    at /home/user/repos/libgit2/src/cache.c:269
#9  0x00007ffff7a9d8a7 in git_odb_object_free (object=0xe6aac90)
    at /home/user/repos/libgit2/src/odb.c:142
#10 0x00007ffff7add89f in git_object_lookup_prefix (object_out=0x7fffffffd678, 
    repo=0x6610a0, id=0x6d6e00, len=40, type=GIT_OBJ_COMMIT)
    at /home/user/repos/libgit2/src/object.c:192
[...]

PR #3842, opened 2016-06-27
blame: increment reference count for origin's commit

pks-t:
blame: do not decrement commit refcount in make_origin

When we create a blame origin, we try to look up the blob that is
to be blamed at a certain revision. When this lookup fails, e.g.
because the file did not exist at that certain revision, we fail
to create the blame origin and return `NULL`. The blame origin
that we have just allocated is thereby free'd with
`origin_decref`.

The `origin_decref` function does not only decrement reference
counts for the blame origin, though, but also for its commit and
blob. When this is done in the error case, we will cause an
uneven reference count for these objects. This may result in
hard-to-debug failures at seemingly unrelated code paths, where
we try to access these objects when they in fact have already
been free'd.

Fix the issue by refactoring `make_origin` such that we only
allocate the object after the only function that may fail so that
we do not have to call `origin_decref` at all. Also fix the
`pass_blame` function, which indirectly calls `make_origin`, to
free the commit when `make_origin` failed.

### DIFF
--- a/src/blame_git.c
+++ b/src/blame_git.c
@@ -37,8 +37,13 @@ static void origin_decref(git_blame__origin *o)
 static int make_origin(git_blame__origin **out, git_commit *commit, const char *path)
 {
 	git_blame__origin *o;
+	git_object *blob;
 	size_t path_len = strlen(path), alloc_len;
 	int error = 0;
+
+	if ((error = git_object_lookup_bypath(&blob, (git_object*)commit,
+			path, GIT_OBJ_BLOB)) < 0)
+		return error;
 
 	GITERR_CHECK_ALLOC_ADD(&alloc_len, sizeof(*o), path_len);
 	GITERR_CHECK_ALLOC_ADD(&alloc_len, alloc_len, 1);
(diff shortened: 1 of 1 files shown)


pks-t:
Ok, I revert my previous comment about no additional memory leaks. Seems like Valgrind just wasn't able to catch them. More fiddeling with refcounts then :/

pks-t:
Phew. I've finally been able to get to a version that has no memory leaks in our test suite while seemingly fixing the issues encountered in #3793. Running the modified example of @ameily does not cause a segfault after a few seconds anymore. I've been running the test on the libgit2 repo now for 20 minutes and until now everything looks good.

These small issues are always a tad depressing when one looks at how long it took to fix related to the actual outcome, which most often looks rather trivial :/ But meh, whatever.

ethomson:
> These small issues are always a tad depressing when one looks at how long it took to fix related to the actual outcome, which most often looks rather trivial :/ But meh, whatever.

Yeah, and blame is a special kind of terrible.  :(

As always, thanks.

Issue #3817, opened 2016-06-08
git_config_parse_int64 (and git__strtol64) fails to parse maximum negative int64


joshtriplett:
`git__strtol64`, used by `git_config_parse_int64`, parses the integer as positive and tracks the sign separately.  As a result, it cannot parse the maximum negative int64, -9223372036854775808.

(This is incredibly minor and low-priority; I ran into it when writing a test suite for a language binding.)

rkrp:
I am a beginner in contributing to opensource projects. I would like to pick this one up.

I am thinking, this can be fixed by replacing 

```c
nn = n*base + v;
```
as

```c
nn = n * base + (neg ? -v : v);
```

If this is illegible, it can be replaced with an if..else block.

Also, the check for overflows must be rewritten accordingly. Otherwise `(nn < n)` would be true for any negative number, resulting in an overflow. 

Please let me know if I am on the right track.

pks-t:
@rkrp I think your best bet to fix the issue is to write some unit tests first. Unit tests help in specifying the actual behavior of functions so that you can then check if your fix actually solves the error mentioned in this ticket as well as not breaking other use-cases.

`git_config_parse_int64` is called by `git_config_get_int64`, which is being tested in `tests/config/read.c`. You could simply add a test case which tries to parse various "extreme" values like INT64_MAX or INT64_MIN and then verify that we actually parse the numbers correctly. As we currently don't parse them correctly, you should see an error in the test suite now.

When you have reproduced the error, try to fix it and re-run the test suite to verify that your tests now actually succeed. Like this you can be sure your fix is correct and we will also have a unit test present that will help spot regressions for this code in the future.

Let me know if you need any help with this.

rkrp:
@pks-t Thanks for the instructions. Now that I have done it, starting out by writing unit tests make more sense. 

- added the borderline cases to the unit tests which obviously caused it to fail. 
- fixed `git__strntol64()` at `src/util.c` to parse `INT64_MIN` without failing
- tests are now running without any fail

Please let me know if I can fork and send a pull request now.

ethomson:
> Please let me know if I can fork and send a pull request now.

That sounds great, please do!  😀

rkrp:
[...]

PR #3846, opened 2016-06-28
Fixed bug while parsing INT64_MIN

rkrp:
Fixed the bug which caused `strtol()` to fail parsing `INT64_MIN`. Addressed in #3817.

Added test cases in `tests/core/strtol.c` to check for the boundary values.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -128,8 +128,8 @@ int git__strntol64(int64_t *result, const char *nptr, size_t nptr_len, const cha
 			v = c - 'A' + 10;
 		if (v >= base)
 			break;
-		nn = n*base + v;
-		if (nn < n)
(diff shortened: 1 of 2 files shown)


pks-t:
Looks good to me, thanks for the fix. The AppVeyor failure is unrelated, so I'd consider this PR good to go (even though it should be squashed when pulled).

rkrp:
@pks-t So, should I squash these two commits and open another pull request, deleting this one?

rkrp:
@pks-t  Squashed the commits.

pks-t:
@rkrp Thanks. Wouldn't have been necessary, though, as GitHub recently introduced a feature to auto-squash commits when a pull request is accepted. But can't hurt if you do it ;)

PR #3850, opened 2016-07-03
Enable https transport for custom TLS streams

wildart:
Unfortunately, custom TLS stream API is useless because `https` transport become disabled when [any of secure connection frameworks aren't used](https://github.com/libgit2/libgit2/blob/813d73f64d779257fd24cd70bbc281352c34812b/src/transport.c#L32).

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -29,9 +29,7 @@ int git_libgit2_features(void)
 #ifdef GIT_THREADS
 		| GIT_FEATURE_THREADS
 #endif
-#if defined(GIT_OPENSSL) || defined(GIT_WINHTTP) || defined(GIT_SECURE_TRANSPORT)
 		| GIT_FEATURE_HTTPS
-#endif
 #if defined(GIT_SSH)
 		| GIT_FEATURE_SSH
 #endif
--- a/src/transport.c
+++ b/src/transport.c
@@ -29,9 +29,7 @@ static transport_definition local_transport_definition = { "file://", git_transp
 static transport_definition transports[] = {
 	{ "git://",   git_transport_smart, &git_subtransport_definition },
 	{ "http://",  git_transport_smart, &http_subtransport_definition },
-#if defined(GIT_OPENSSL) || defined(GIT_WINHTTP) || defined(GIT_SECURE_TRANSPORT)
 	{ "https://", git_transport_smart, &http_subtransport_definition },
-#endif
 	{ "file://",  git_transport_local, NULL },
 #ifdef GIT_SSH
 	{ "ssh://",   git_transport_smart, &ssh_subtransport_definition },
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -620,7 +620,6 @@ static int http_connect(http_subtransport *t)
 
 	error = git_stream_connect(t->io);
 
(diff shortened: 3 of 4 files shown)


carlosmn:
This is something we've gone back and forth on a bit. Whether we should treat HTTPS or SSH any different just because git ships with them. We currently treat them as any other pluggable protocol, but there's been some confusion as it often ends up without HTTPS on the field.

The way to fix this should be to remove the special-casing, rather than adding more. If adding this define makes the code work sensibly, it shows that we don't actually need it and should instead let the TLS stream constructors error out instead.

src/settings.c
```diff
@@ -29,7 +29,7 @@ int git_libgit2_features(void)
 #ifdef GIT_THREADS
 		| GIT_FEATURE_THREADS
 #endif
-#if defined(GIT_OPENSSL) || defined(GIT_WINHTTP) || defined(GIT_SECURE_TRANSPORT)
+#if defined(GIT_OPENSSL) || defined(GIT_WINHTTP) || defined(GIT_SECURE_TRANSPORT) || defined(GIT_CUSTOM_TLS)
```
carlosmn: This defines are there to show how libgit2 was compiled. Adding this would make it meaningless.

src/transport.c
```diff
@@ -29,7 +29,7 @@ static transport_definition local_transport_definition = { "file://", git_transp
 static transport_definition transports[] = {
 	{ "git://",   git_transport_smart, &git_subtransport_definition },
 	{ "http://",  git_transport_smart, &http_subtransport_definition },
-#if defined(GIT_OPENSSL) || defined(GIT_WINHTTP) || defined(GIT_SECURE_TRANSPORT)
+#if defined(GIT_OPENSSL) || defined(GIT_WINHTTP) || defined(GIT_SECURE_TRANSPORT) || defined(GIT_CUSTOM_TLS)
```
carlosmn: This whole line should just go. The ability to put in your own TLS transport exists regardless of whether you compiled libgit2 with this define.

src/transports/http.c
```diff
@@ -620,7 +620,7 @@ static int http_connect(http_subtransport *t)
 
 	error = git_stream_connect(t->io);
 
-#if defined(GIT_OPENSSL) || defined(GIT_SECURE_TRANSPORT) || defined(GIT_CURL)
+#if defined(GIT_OPENSSL) || defined(GIT_SECURE_TRANSPORT) || defined(GIT_CURL) || defined(GIT_CUSTOM_TLS)
```
carlosmn: Since the stream is pluggable, it doesn't make sense to have this line either. Either we have a stream and need to run this code or it'll error out before we get here.

wildart:
Fair enough. Removing this make sense as TLS stream creation [would fail](https://github.com/libgit2/libgit2/blob/813d73f64d779257fd24cd70bbc281352c34812b/src/tls_stream.c#L38) if stream constructor is not provided.

tkelman:
Bump. Thoughts on this without any of the ifdefs?

wildart:
@ethomson Review, plz.

ethomson:
Yep!  This looks good to me - thanks!

PR #3853, opened 2016-07-07
Shallow roots

tiennou:
This adds a helper function to grab the shallow OIDs, as right now there's only a yes/no check available.

I figured a way to get that info (even though our behavior might not be optimal) is a first step toward support.

### DIFF
--- a/include/git2/repository.h
+++ b/include/git2/repository.h
@@ -11,6 +11,7 @@
 #include "types.h"
 #include "oid.h"
 #include "buffer.h"
+#include "oidarray.h"
 
 /**
  * @file git2/repository.h
@@ -744,6 +745,15 @@ GIT_EXTERN(const char *) git_repository_get_namespace(git_repository *repo);
  * @return 1 if shallow, zero if not
  */
 GIT_EXTERN(int) git_repository_is_shallow(git_repository *repo);
+
+/**
+ * Determine the shallow roots of the repository
+ *
+ * @param out An array of shallow oids. Can be NULL.
+ * @param repo The repository
+ * @return 1 if shallow, zero if not
+ */
+GIT_EXTERN(int) git_repository_shallow_roots(git_oidarray *out, git_repository *repo);
 
 /**
  * Retrieve the configured identity to use for reflogs
--- a/include/git2/repository.h
+++ b/include/git2/repository.h
@@ -749,9 +749,9 @@ GIT_EXTERN(int) git_repository_is_shallow(git_repository *repo);
 /**
  * Determine the shallow roots of the repository
  *
- * @param out An array of shallow oids. Can be NULL.
+ * @param out An array of shallow oids.
  * @param repo The repository
- * @return 1 if shallow, zero if not
+ * @return 0 on success, an error otherwise.
  */
 GIT_EXTERN(int) git_repository_shallow_roots(git_oidarray *out, git_repository *repo);
 
--- a/src/repository.c
+++ b/src/repository.c
@@ -2477,26 +2477,62 @@ int git_repository_state_cleanup(git_repository *repo)
 	return git_repository__cleanup_files(repo, state_files, ARRAY_SIZE(state_files));
 }
 
-int git_repository_is_shallow(git_repository *repo)
+int git_repository_shallow_roots(git_oidarray *out, git_repository *repo)
 {
 	git_buf path = GIT_BUF_INIT;
+	git_buf contents = GIT_BUF_INIT;
+	git_array_oid_t array;
 	struct stat st;
 	int error;
+	size_t start, end;
 
 	if ((error = git_buf_joinpath(&path, repo->path_repository, "shallow")) < 0)
 		return error;
 
 	error = git_path_lstat(path.ptr, &st);
-	git_buf_free(&path);
 
 	if (error == GIT_ENOTFOUND) {
+		git_buf_free(&path);
 		giterr_clear();
 		return 0;
 	}
 
+	if (out == NULL) {
+		git_buf_free(&path);
+		return st.st_size == 0 ? 0 : 1;
+	}
+
+	error = git_futils_readbuffer(&contents, git_buf_cstr(&path));
+	git_buf_free(&path);
+
 	if (error < 0)
 		return error;
-	return st.st_size == 0 ? 0 : 1;
+
+	git_array_init(array);
+
+	start = end = 0;
+	while (end < contents.size) {
+		if (contents.ptr[end] == '\n' || end == contents.size) {
+			if (end - start == GIT_OID_HEXSZ) {
+				git_oid *oid = git_array_alloc(array);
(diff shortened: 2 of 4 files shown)


tiennou:
I think, with a helpful nudge, I might be able to have revwalk not error out on missing parents (as per a Slack discussion, and my goal atm), but I'm still not clear on the merge-base case...

tiennou:
Will fix the leaking cache on update first ;-).

include/git2/repository.h
```diff
@@ -746,6 +747,15 @@ GIT_EXTERN(const char *) git_repository_get_namespace(git_repository *repo);
 GIT_EXTERN(int) git_repository_is_shallow(git_repository *repo);
 
 /**
+ * Determine the shallow roots of the repository
+ *
+ * @param out An array of shallow oids. Can be NULL.
```
carlosmn: Why can it be `NULL`? Why would you call this function if you did not want the answer?

include/git2/repository.h
```diff
@@ -746,6 +747,15 @@ GIT_EXTERN(const char *) git_repository_get_namespace(git_repository *repo);
...
+ * @param repo The repository
+ * @return 1 if shallow, zero if not
```
carlosmn: This behaves completely differently from the rest of the functions. This is not `git_reposistory_is_shallow`, this reads the file and should thud return 0 or an error code.

src/repository.c
```diff
@@ -2477,26 +2477,78 @@ int git_repository_state_cleanup(git_repository *repo)
...
+	git_array_clear(repo->shallow_oids);
+
+	start = end = 0;
+	while (end < contents.size) {
+		if (contents.ptr[end] == '\n' || end == contents.size) {
+			if (end - start == GIT_OID_HEXSZ) {
+				git_oid *oid = git_array_alloc(repo->shallow_oids);
+				error = git_oid_fromstrn(oid, contents.ptr + start, GIT_OID_HEXSZ);
+				if (error < 0) {
+					/* TODO: Warn ? */
```
carlosmn: It should error out if the file is corrupt

tests/repo/shallow.c
```diff
@@ -37,3 +39,39 @@ void test_repo_shallow__clears_errors(void)
 	cl_assert_equal_i(0, git_repository_is_shallow(g_repo));
 	cl_assert_equal_p(NULL, giterr_last());
 }
+
+void test_repo_shallow__shallow_oids(void)
+{
+	git_oidarray oids, oids2;
```
carlosmn: These variables need to be freed after we're done with them.

include/git2/repository.h
tiennou: Since I did the extraction in two pass, it behaved like `git_repository_is_shallow()` (which used it internally). But it doesn't make sense anymore. I'll make it mandatory.

include/git2/repository.h
tiennou: Same reason. And it also doesn't make sense 😉 .

src/repository.c
tiennou: Yeah, that feels better actually. Also, I don't like that loop's nesting, I'll try to cleanup. If you have some idea, I'll take it.

tests/repo/shallow.c
tiennou: Grr 😆. Any chance I could graft [`leaks`](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/leaks.1.html) on clar's exit, so I get a nice reminder when that happens ?

ethomson:
> Grr 😆. Any chance I could graft leaks on clar's exit, so I get a nice reminder when that happens ?

We run valgrind on the Linux builds for each PR, and the Windows builds are instrumented with leak detection as well.

tiennou:
@ethomson I was wondering about having a way to run leaks just before `clar` exists, so that I can catch those locally. Right now I just busy-loop in `main`, run `leaks`, and kill clar manually afterward, but it's hacky, and I was wondering if there was a better way...

ethomson:
> @ethomson I was wondering about having a way to run leaks just before clar exists, so that I can catch those locally. Right now I just busy-loop in main, run leaks, and kill clar manually afterward, but it's hacky, and I was wondering if there was a better way...

Ugh.  I don't know why leaks doesn't have the ability to run the process for you and catch its leaks.

Anyway.  I guess you could add some functionality to clar that would exec a process at the end of the run.  Then you could, I guess, do something like `./libgit2_clar -smytest -e"leaks libgit2_clar"`.  I guess you could also hardcode `leaks` detection, but that sort of bums me out.

If it were me, I would just run `valgrind` in a VM.  (Unless it's stable on Mac again, then I wouldn't even bother with the VM.)

src/repository.c
```diff
@@ -2477,6 +2477,65 @@ int git_repository_state_cleanup(git_repository *repo)
...
+			if (end - start == GIT_OID_HEXSZ) {
+				git_oid *oid = git_array_alloc(repo->shallow_oids);
+
+				error = git_oid_fromstrn(oid, contents.ptr + start, GIT_OID_HEXSZ);
+				if (error < 0)
+					return error;
+			}
+			start = end + 1;
+		}
+		end++;
```
ethomson: It seems like there's not a lot of error checking here...  It looks like it would handle a very malformed line if it was 80 characters of things that looked like hex characters...?  Also, walking every character yourself seems a bit unnecessary and I think that you could solve both problems with some minor refactoring.

I would rather see you use some existing standard library function (`git__strtok` may be a good choice) to find the terminating newline, validate that it's `GIT_OID_HEXSZ` away from your start, and then try to parse it...  

Does `git` handle blank linkes?  Comments?  Windows style newlines?  I assume not, but I wanted to ask.

tiennou: The documentation about the shallow file format is pretty terse (what I know of it is what's in `Documentation/technical/shallow.txt`), and I wanted to know if it's acceptable to look at the source (since licensing).

I rebased and pushed a version that uses `git__strsep`, which really feels cleaner. I do have a point to make : our implementation is "non-conforming" (for some definition of it :wink:). Here's the relevant tidbit from my OS X man pages (which I was referring to when I wrote it) :

> The location of the next character after the delimiter character (or NULL, if the end of the string was reached) is stored in *stringp.  The original value of *stringp is returned.

Ours just leave `*stringp` as it is and returns `NULL` in that case, which can complicate parsing files where newlines-at-EOF are not guaranteed.

If that's expected, allow me to either 1) fix it or 2) add a comment on top of that function so you don't blindly trust your manpages (as I did).

For the record, the relevant changes are :

``` C
while ((line = git_strsep(&buffer, "\n")) != NULL) {
    /* stuff with line */
}
if (*buffer) {
    /* missing EOL */
}
```
to
``` C
while ((line = git_strsep(&buffer, "\n")) != NULL) {
    if (buffer == NULL)
        /*missing EOL */;
    /* stuff with line */
}
```

tiennou:
Is the AppVeyor failure relevant or not ?

Issue #2923, opened 2015-02-24
Allow to create a remote without a repository


carlosmn:
Every remote constructor we have takes a repository, but you do not always have a repository available. Commands like

```
git ls-remote https://example.com/repo
```

do not require a repository at all. We should be able to lift the requirement of a repository for connecting to a remote and seeing what references there are available.

Performing a fetch would obviously require a repository.

wonson:
libgit2sharp got it solved, how about the c++ lib?? does it go any further now?

jzahka:
+1

mathieumg:
There is a PR (linked above, https://github.com/libgit2/libgit2/pull/3855) but this is marked as "up for grabs". Is something blocking this? I would be glad to help!

pks-t:
I just grabbed #3855, will push it as it's been open for so long and is obviously requested by more than just one person

carlosmn:
Fixed via #4233

PR #3855, opened 2016-07-08
 Allow creating a remote without a repository

heavenlyhash:
This addresses #2923.

A new function `git_remote_create_anonymous2()` creates a remote, given only a url.  It behaves much like `git_remote_create_anonymous()`, but performs much _less_ work: it does not accept a repo param, it thus does not attempt to load any configuration or apply url rewriting rules.

WIP/RFC.  Dunno if this is a great solution; just took stab and this is the first place my code came to rest and, like, compiled.  Not a C expert, so I put on my party hat and figured I'd PR (read: ask for adult supervision) from here :)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -22,6 +22,7 @@ Dmitry Kakurin
 Dmitry Kovega
 Emeric Fermas
 Emmanuel Rodriguez
+Eric Myhre
 Florian Forster
 Holger Weiss
 Ingmar Vanhassel
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,10 @@ v0.24 + 1
       `git_repository_open_ext` with this flag will error out if either
       `$GIT_WORK_TREE` or `$GIT_COMMON_DIR` is set.
 
+* `git_remote_create_anonymous2()` creates a remote that is not associated
+  to any repository (and does not apply configuration like 'insteadof' rules).
+  This is mostly useful for e.g. emulating `git ls-remote` behavior.
+
 ### API removals
 
 * `git_blob_create_fromchunks()` has been removed in favour of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ v0.24 + 1
       `git_repository_open_ext` with this flag will error out if either
       `$GIT_WORK_TREE` or `$GIT_COMMON_DIR` is set.
 
-* `git_remote_create_anonymous2()` creates a remote that is not associated
+* `git_remote_create_unattached()` creates a remote that is not associated
   to any repository (and does not apply configuration like 'insteadof' rules).
   This is mostly useful for e.g. emulating `git ls-remote` behavior.
 
--- a/include/git2/remote.h
+++ b/include/git2/remote.h
@@ -78,16 +78,20 @@ GIT_EXTERN(int) git_remote_create_anonymous(
 		const char *url);
 
 /**
- * Create an anonymous remote, without a connected local repo
+ * Create a remote without a connected local repo
  *
  * Create a remote with the given url in-memory. You can use this when
  * you have a URL instead of a remote's name.
+ *
+ * Contrasted with git_remote_create_anonymous, an unattached remote
+ * will not consider any repo configuration values (such as insteadof url
+ * substitutions).
  *
  * @param out pointer to the new remote objects
  * @param url the remote repository's URL
  * @return 0 or an error code
  */
-GIT_EXTERN(int) git_remote_create_anonymous2(
+GIT_EXTERN(int) git_remote_create_unattached(
 		git_remote **out,
 		const char *url);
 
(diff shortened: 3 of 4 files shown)


heavenlyhash:
Maintainerfolk: is adding a new function like this a good idea -- Or it be better to re-do and change the behavior of `create_internal()` to work with a null repo?  I'm afraid `create_internal` would become quite a twisty maze of conditionals, but having less API surface area also makes sense.  Would appreciate your feedback :)

ethomson:
Hi @heavenlyhash !  Welcome and thanks for the nice first PR - this is delightful.

I agree that I would like to avoid less API surface and make `repo` optional to the existing `create_anonymous`.  I _suspect_ that you'll be able to be able to get surgical with `create_internal`, perhaps pulling out the repo-specific stuff into its own function?  Let me know if it gets gnarly, but it looks like you've got a good handle on this.  😀 

Would you mind hacking on a unit tests while you're in here?

Do we need to beef up any other parts of the code base that assume that a remote has a `repo`?
/cc @carlosmn

carlosmn:
Right, the idea is to make passing in a repsitory optional, rather than adding new APIs. In the `git_remote` namespace there'll be quite a few things that try to access the repository. We definitely have to make sure that trying to run fetch on such a `git_remote` does not crash but returns an error.

heavenlyhash:
I might need a little help getting the test coverage and additional checks in place.  0% skill navigating a large project in C (much less convincing myself I've done so exhaustively/correctly..)  Will try to get a start though (especially since you have such lovely CI set up).

The first thing that jumps out at me as a liiittle odd is that the `apply_insteadof` call in `create_internal` returns new mem.  Is it reasonable to have an if|else in `create_internal` that does strdup if there's no insteadof configuration to apply?

ethomson:
> I might need a little help getting the test coverage and additional checks in place. 0% skill navigating a large project in C (much less convincing myself I've done so exhaustively/correctly..) Will try to get a start though (especially since you have such lovely CI set up).

Yeah, it's no joke.  We've got a lot of unit tests, and our own test runner framework to boot.  So there's a lot of overhead.  If you want to get a start poking around with it, that would be great, and feel free to ask questions (here or in slack).

> Is it reasonable to have an if|else in create_internal that does strdup if there's no insteadof configuration to apply?

Yeah, that's definitely what I would do, too. 😄

heavenlyhash:
Okay, great, thanks!  Really appreciate the quick guidance on memory stuff

Tests: looks like `tests/{.,network,online}/{remotes,insteadof}.c` is the target area?  The main thing I'm thinking of writing would probably be an offline test that uses a local filesystem repo to as a remote and exercises `git_remote_ls` to make sure the data saved by `create_internal` is operable.

Error on fetch on a repoless remote: EINVALID seems like the closest match?

Combinations of valid parameters: a `name` param doesn't make much sense if there's no repo config to interact with, should providing `name` with a null `repo` be an error of some kind?  Not sure on fetchspec -- `git ls-remote [url] refs/unusual/*` is a thing in exec'ing cgit, but that's not a fetchspec, so maybe fetchspecs don't make sense without a repo either?

heavenlyhash:
Follow up on that last question about parameter validation: yeah, what's the convention to follow around combinatorially valid parameters and null checks?  I fixed the repetitive code; all goes through `create_internal` now.  But of the existing functions, `git_remote_create` and `[...]_with_fetchspec` are both chock full of parameters that don't make any sense if there's no repo (and the only thing those functions do is... check params); only `git_remote_create_anonymous` really makes any sense with a null repo.

- Option 1: allow null repos through any of the funcs; just make sure `create_internal` never crashes on any of them.
- Option 2: Move the assert against null repos that I removed from `create_internal` into the other exported functions clearly need a repo.

I tried Option 1 but ISTM it spirals towards messiness.  For example, if `create_internal` started returning EINVALID to indicate bad usage like providing a fetchspec with a null repo, it would mean `git_remote_create` would *still* never DTRT with a null repo, because it hands down a default fetchspec!  Fixing it creates epicycles; ignoring it means real incorrect usage from a client would be silent, which is not great.

Option 2 seems a little better, but means two of the functions panic on null repos, and the third one is fine.  Maybe that's ok, but that seems like a api style and consistency thing I should check with you folks on?

... or I guess Option 3: make all three existing functions assert against null repos (as they do before this PR) and add a new function that has no repo param.  (At least the implementation would be DRY now :3)  Adds to API surface area; but seems most consistent.

Preferences?

heavenlyhash:
Ping :)

I put this down for a while and my memory's already getting dusty, but I think last time I looked, my favor was still towards adding a new function, so that the others can continue to reject null values as a whole instead of needing a matrix of conditions to be documented.  The current state of the branch is with such a new function, and the implementation cleaned up.

herman-rogers:
So ... is this being worked on or is it dead?

ethomson:
Closed via #4233

ethomson:
Thanks @heavenlyhash !

PR #3858, opened 2016-07-12
consitency between _from_buffer and _from_string

### LINKED ISSUE #3857: git_signature_from_buffer parameters
I was trying to port libgit2sharp to the newest libgit2 and I saw some difference in some signatures. There is [git_signature_from_buffer](https://github.com/libgit2/libgit2/blob/master/include/git2/signature.h#L76) and [git_diff_from_buffer](https://github.com/libgit2/libgit2/blob/master/include/git2/diff.h#L1192).

``` c
GIT_EXTERN(int) git_signature_from_buffer(git_signature **out, const char *buf);
// ...
GIT_EXTERN(int) git_diff_from_buffer(
    git_diff **out,
    const char *content,
    size_t content_len);
```

It feels a bit strange that two signatures with basically the same functionality (do something with a given buffer) have different signatures. I personally would prefer later signature (the one with the content_len specified).

C has null termination for strings (by convention) and of course one can use the good old (pointer, length) pair. However, in C# if I do a conversation with Encoding::GetBytes from a string to a byte[] buffer, I get a byte buffer which is not null terminted. So, if I want to expose a _from_buffer function i'd in csharp with a byte[] signature

``` csharp
static extern int git_signature_from_buffer(ref git_signature *out, byte[] buf)
```

for lets say a function like

``` csharp
public class Signature
{
  public static From(byte[] buffer)
  {
    // need to create another buffer with size + 1 and copy everything
    // call git_signature_from_buffer
  }
  // or we can just create another constructor
  public static Signature(byte[] buffer)
  {
    // need to create another buffer with size + 1 and copy everything
    // call git_signature_from_buffer
  }
}
```

I would have to create another buffer (like the comment already says) with the size of the original buffer and the last one byte empty set to 0 (the null terminator).

If I want to expose a the same functionality with a string as a parameter, like:

``` csharp
public class Signature
{
  public static From(string signature) { }
}
```

The marshaller of course takes care of the null termination:

``` csharp
[DllImport(libgit2)]
internal static extern unsafe int git_signature_from_buffer(
    out git_signature* signature,
    [MarshalAs(UnmanagedType.CustomMarshaler, MarshalCookie = UniqueId.UniqueIdentifier, MarshalTypeRef = typeof(StrictUtf8Marshaler))] string buffer);
```

but it becomes much more difficult to create a function with a byte[] signature.

So my proposition is to use always a pointer, length pair in _from_buffer functions and maybe introduce a _from_string convenience function to each of them which accepts a pointer to a a null terminated string and does a strlen and passes it to the _from_buffer method.

txdv:
fixes #3857

### DIFF
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -1194,6 +1194,10 @@ GIT_EXTERN(int) git_diff_from_buffer(
 	const char *content,
 	size_t content_len);
 
+GIT_EXTERN(int) git_diff_from_string(
+	git_diff **out,
+	const char *content);
+
 /**
  * This is an opaque structure which is allocated by `git_diff_get_stats`.
  * You are responsible for releasing the object memory when done, using the
--- a/include/git2/signature.h
+++ b/include/git2/signature.h
@@ -70,10 +70,29 @@ GIT_EXTERN(int) git_signature_default(git_signature **out, git_repository *repo)
  * of a colon separator).
  *
  * @param out new signature
- * @param buf signature string
+ * @param buf buffer containing the signature
+ * @param buf_len length of the buffer
  * @return 0 on success, or an error code
  */
-GIT_EXTERN(int) git_signature_from_buffer(git_signature **out, const char *buf);
+GIT_EXTERN(int) git_signature_from_buffer(
+	git_signature **out,
+	const char *buf,
+	size_t buf_len);
+
+/**
+ * Create a new signature by parsing the given buffer, which is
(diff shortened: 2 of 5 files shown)


txdv:
I guess consistency between the names of the parameters would be useful too?

ethomson:
I'm :-1: on having overloads just for the type.  If `from_buffer` taking `char *, size_t` is superior, then that should be the solution.

But I don't think that this is what callers want from C, or even from C#.  LibGit2Sharp also produces null-terminated strings in its marshaller, I don't see what's wrong with that.

Issue #3650, opened 2016-03-03
We should refresh objects when they're referenced


carlosmn:
git refreshes the mtime of objects (or the pack) when we add a reference to them, to help prune avoid race conditions with older objects which are de- and then re-referenced.

carlosmn:
Quoting @mhagger from another issue:

---

When is a new reference made to an object?

 - When another new object is created that refers to it (e.g., when a newly-created tree refers to an existing blob).
 - When a reference (or, theoretically, a reflog entry) is updated to point directly at the object. Refreshing is needed in this case to cover the race that could otherwise occur if git prune reads the references before the new reference exists, and therefore thinks that the object is unreachable.

It is only necessary to refresh the timestamp on the object that is directly referred-to, not those of its ancestors.

ethomson:
Fixed with #3861

PR #3861, opened 2016-07-14
odb: freshen existing objects when writing

ethomson:
When writing an object, we calculate its OID and see if it exists in the object database.  If it does, we need to freshen the file that contains it.

This adds a new function to backends, `freshen`, which will freshen the object if it exists.

Previously, we would do a simple `exists` check during `git_odb_write`, after calculating the new object ID.  Now, we will use the `freshen` function if it is implemented on a particular backend (falling back to the `exists` check if it doesn't).

A slight disappointment is that we no longer look in the cache first to see if an object exists.  To do that, we would need to store more information in the cache, namely the backend that the object was found in (or else we would need to go locate the object again to freshen it, and if we do that, there would be no point in looking in the cache at all.)  That's quite a yak to shave, it turns out.  But it's probably not really a big deal, since this is only used in the `write` functions, where it is presumed that the object _doesn't_ exist (and thus wouldn't be in the cache).

(So in the general case, this is as performant as before.)

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -654,7 +654,10 @@ void git_odb_free(git_odb *db)
 	GIT_REFCOUNT_DEC(db, odb_free);
 }
 
-static int odb_exists_1(git_odb *db, const git_oid *id, bool only_refreshed)
+static int odb_exists_1(
+	git_odb *db,
+	const git_oid *id,
+	bool only_refreshed)
 {
 	size_t i;
 	bool found = false;
@@ -671,6 +674,44 @@ static int odb_exists_1(git_odb *db, const git_oid *id, bool only_refreshed)
 	}
 
 	return (int)found;
+}
+
+static int odb_freshen_1(
+	git_odb *db,
+	const git_oid *id,
+	bool only_refreshed)
+{
+	size_t i;
+	bool found = false;
+
+	for (i = 0; i < db->backends.length && !found; ++i) {
+		backend_internal *internal = git_vector_get(&db->backends, i);
+		git_odb_backend *b = internal->backend;
+
+		if (only_refreshed && !b->refresh)
+			continue;
+
+		if (b->freshen != NULL)
+			found = !b->freshen(b, id);
+		else if (b->exists != NULL)
+			found = b->exists(b, id);
+	}
+
+	return (int)found;
+}
+
+static int odb_freshen(git_odb *db, const git_oid *id)
+{
+	assert(db && id);
+
+	if (odb_freshen_1(db, id, false))
+		return 1;
+
+	if (!git_odb_refresh(db))
+		return odb_freshen_1(db, id, true);
+
+	/* Failed to refresh, hence not found */
+	return 0;
 }
 
 int git_odb_exists(git_odb *db, const git_oid *id)
@@ -1131,7 +1172,7 @@ int git_odb_write(
 	assert(oid && db);
 
 	git_odb_hash(oid, data, len, type);
-	if (git_odb_exists(db, oid))
+	if (odb_freshen(db, oid))
 		return 0;
 
 	for (i = 0; i < db->backends.length && error < 0; ++i) {
@@ -1257,7 +1298,7 @@ int git_odb_stream_finalize_write(git_oid *out, git_odb_stream *stream)
 
 	git_hash_final(out, stream->hash_ctx);
 
-	if (git_odb_exists(stream->backend->odb, out))
+	if (odb_freshen(stream->backend->odb, out))
 		return 0;
 
 	return stream->finalize_write(stream, out);
--- a/src/odb_pack.c
+++ b/src/odb_pack.c
@@ -363,6 +363,18 @@ static int pack_backend__read_header(
 	return git_packfile_resolve_header(len_p, type_p, e.p, e.offset);
 }
 
+static int pack_backend__freshen(
+	git_odb_backend *backend, const git_oid *oid)
+{
+	struct git_pack_entry e;
+	int error;
+
+	if ((error = pack_entry_find(&e, (struct pack_backend *)backend, oid)) < 0)
+		return error;
+
+	return git_futils_touch(e.p->pack_name);
+}
+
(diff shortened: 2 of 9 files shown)


carlosmn:
xref #3650

ethomson:
Note that git itself will only freshen a pack file once per execution to avoid thrashing a packfile with touches, which would be quite annoying.  That's probably not appropriate for us but we can keep track of the last time we've freshened a packfile and re-touch it only every `n` seconds.

I'm all ears if you have a good idea for what that `n` should be, otherwise I'll pick something that seems reasonable and we'll hope that we don't regret it.  :P

arthurschreiber:
@ethomson What's the status here? Anything I could help with?

ethomson:
@arthurschreiber Just needs a review and some sanity checking, especially on the 2 second number that I picked out of thin air.

src/odb.c
```diff
@@ -673,6 +676,44 @@ static int odb_exists_1(git_odb *db, const git_oid *id, bool only_refreshed)
...
+	assert(db && id);
+
+	if (odb_freshen_1(db, id, false))
+		return 1;
+
+	if (!git_odb_refresh(db))
+		return odb_freshen_1(db, id, true);
+
+	/* Failed to refresh, hence not found */
+	return 0;
```
pks-t: I think the return code semantics are a tad confusing - I guess this results from `odb_freshen` replacing the `git_odb_exists` calls, where the exists-function is obviously returning a boolean value. In this case, though, I'd rather expect to see our usual error semantics with `-1` as error code, as `odb_freshen` does not hint at a returned boolean value.

tests/odb/freshen.c
```diff
@@ -0,0 +1,80 @@
+#include "clar_libgit2.h"
+#include "odb.h"
+#include "posix.h"
+
+git_repository *repo;
+git_odb *odb;
```
pks-t: These should be file-scope only

tests/odb/freshen.c
```diff
@@ -0,0 +1,80 @@
...
+
+	old_times[0].tv_sec = 1234567890;
+	old_times[0].tv_usec = 0;
+	old_times[1].tv_sec = 1234567890;
+	old_times[1].tv_usec = 0;
+
+	/* set time to way back */
+	cl_must_pass(p_utimes("testrepo.git/objects/pack/" PACKED_FN, old_times));
+
+	cl_must_pass(p_lstat("testrepo.git/objects/pack/" PACKED_FN, &before));
```
pks-t: Minor style nit: additional newline not present in the above test

src/odb_pack.c
```diff
@@ -367,12 +370,28 @@ static int pack_backend__freshen(
...
+	if (e.p->last_freshen > now - FRESHEN_FREQUENCY)
+		return 0;
+
+	times[0].tv_sec = now;
+	times[0].tv_usec = 0;
+	times[1].tv_sec = now;
+	times[1].tv_usec = 0;
+
+	if ((error = p_utimes(e.p->pack_name, times)) < 0)
+		return git_path_set_error(errno, e.p->pack_name, "touch");
```
pks-t: Why is the `git_futils_touch` call expanded? Ah, now I see, probably to avoid the second `time(NULL)` call? Maybe include `time_t` in `git_futils_touch(file)`, instead of duplicating logic? If it is set to `NULL`, `git_futils_touch` may still compute its own `time_t`.

pks-t:
I don't really like the two-second span between refreshes, as the value seems to be rather arbitrary. On the other side we have to choose a value here and I have no arguments for or against any other value besides the trade-off between thrashing and accuracy.

Looks good besides this and the few comments I've left.

ethomson:
> I don't really like the two-second span between refreshes, as the value seems to be rather arbitrary.

My initial thought was to just do _one_ second but I really wanted to avoid dealing with sub-second precision, so to avoid refreshing between a rollover from one second to the next (but within the same one second span) I just chose two.

An argument to going much larger (10 seconds) could certainly be made, but I'm not one to make it.  I wonder if @peff has any thoughts here.

Thanks for the review.

peff:
I considered doing a refresh-after-n-seconds like this in git, but didn't bother since we can usually assume our processes are short-lived. I agree it's a good thing for libgit2 to do.

When picking `n`, I think you have to consider what you're scaling against. And that's basically two things:

  - you're freshening timestamps so that we don't exceed the grace period for `gc.pruneExpire`, which defaults to 2 weeks (though some hosting sites I could mention drop that to 1 hour, and when trying to aggressively drop objects, even to 5 minutes). So if you've freshened within that time period, there's no need (for these purposes) to freshen again. Probably anything up to about 60 seconds would be completely reasonable, even for insanely aggressive attempts to drop recent-but-not-currently-in-use objects.

  - the act of checking the timestamp and pruning the objects is not actually atomic. The worst case of this is probably `git repack --expire-unreachable`, as repacks may take several minutes, during which packs could be freshened, but we would end up deleting them anyway (basically we come up with the list of objects to pack at the start of the program, chug on repacking for a while, then delete all the packs we assme we've obsoleted). Naively I want to say that a shorter freshening frequency would help there, but I actually don't think it would. Once you're inside the `pruneExpire` grace time you're good as long as you don't lose that race, but once you've lost it, it doesn't matter how many times you've freshened.

So...I think you could probably go much higher than you're at. But there's no reason to do so unless you're worried that calling `utime` once every 2 seconds can be considered thrashing. It probably isn't, though (especially given that it's letting you skip an object write entirely).

ethomson:
Thanks, @peff - I agree that every two seconds ought not to be very expensive, even on a truly godawful filesystem.

Thanks @pks-t for the review.

src/odb.c
ethomson: At the moment, I think that the similarity with `refresh` is beneficial, so I'm going to keep it returning a boolean for now, especially since it's just internal.  But I think that `odb.c` has become a bit confusing overall, so I am going to give the whole file a closer look this weekend.

Issue #3856, opened 2016-07-12
Weird ~/.gitconfig causes errors when removing a remote


novalis:
Somehow I ended up with the following in my ~/.gitconfig:
[branch "master"]
    remote = origin
    merge = master
    rebase = true

I assume something went crazy while I was running git tests some time, and that I never noticed until now.

This is not a good configuration, but it shouldn't cause problems.  But it does. Specifically, if you have this in your config, and you perform the following set of actions:

create a remote
fetch from that remote
create a branch off of the remote master branch called "master"
delete the branch
delete the remote

The remote delete fails with the message "Could not find key 'branch.master.rebase' to delete". This is because it's iterating over the config entries (including the ones in the global config) and believes that there is a master branch which must therefore have these config keys.

IMO, failing to delete branch.master.rebase (or branch.master.merge) shouldn't be an error.

This is kind of stupid, sorry.

linquize:
Do you use libgit2 to write config file directly? Or via a git client that uses libgit2?

novalis:
No, I'm not blaming libgit2 for writing the bad config.  I have been a git developer, and while hacking on git, I probably screwed something up at some point that caused that to get added.  I've definitely seen some weird shit.

What I'm complaining about is that libgit2 cannot handle the weird config.

carlosmn:
Fixed via #3862

PR #3862, opened 2016-07-15
remote: Handle missing config values when deleting a remote

novalis:
Somehow I ended up with the following in my ~/.gitconfig:
[branch "master"]
remote = origin
merge = master
rebase = true

I assume something went crazy while I was running the git.git tests
some time ago, and that I never noticed until now.

This is not a good configuration, but it shouldn't cause problems. But
it does. Specifically, if you have this in your config, and you
perform the following set of actions:

create a remote
fetch from that remote
create a branch off of the remote master branch called "master"
delete the branch
delete the remote

The remote delete fails with the message "Could not find key
'branch.master.rebase' to delete". This is because it's iterating over
the config entries (including the ones in the global config) and
believes that there is a master branch which must therefore have these
config keys.

https://github.com/libgit2/libgit2/issues/3856

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -2162,15 +2162,21 @@ static int remove_branch_config_related_entries(
 		if (git_buf_printf(&buf, "branch.%.*s.merge", (int)branch_len, branch) < 0)
 			break;
 
-		if ((error = git_config_delete_entry(config, git_buf_cstr(&buf))) < 0)
-			break;
+		if ((error = git_config_delete_entry(config, git_buf_cstr(&buf))) < 0) {
+			if (error != GIT_ENOTFOUND)
+				break;
+			giterr_clear();
+		}
 
 		git_buf_clear(&buf);
 		if (git_buf_printf(&buf, "branch.%.*s.remote", (int)branch_len, branch) < 0)
 			break;
 
-		if ((error = git_config_delete_entry(config, git_buf_cstr(&buf))) < 0)
-			break;
+		if ((error = git_config_delete_entry(config, git_buf_cstr(&buf))) < 0) {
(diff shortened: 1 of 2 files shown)


src/remote.c
```diff
@@ -2162,15 +2162,19 @@ static int remove_branch_config_related_entries(
 		if (git_buf_printf(&buf, "branch.%.*s.merge", (int)branch_len, branch) < 0)
 			break;
 
-		if ((error = git_config_delete_entry(config, git_buf_cstr(&buf))) < 0)
-			break;
+		if ((error = git_config_delete_entry(config, git_buf_cstr(&buf))) < 0) {
+			if (error != GIT_ENOTFOUND)
+				break;
```
ethomson: Could I trouble you to do a `giterr_clear()` here to remove the error message that `git_config_delete_entry` added?  We would like to avoid leaving that error message hanging around when there isn't actually an error.

src/remote.c
```diff
@@ -2162,15 +2162,19 @@ static int remove_branch_config_related_entries(
...
 
 		git_buf_clear(&buf);
 		if (git_buf_printf(&buf, "branch.%.*s.remote", (int)branch_len, branch) < 0)
 			break;
 
-		if ((error = git_config_delete_entry(config, git_buf_cstr(&buf))) < 0)
-			break;
+		if ((error = git_config_delete_entry(config, git_buf_cstr(&buf))) < 0) {
+			if (error != GIT_ENOTFOUND)
+				break;
```
ethomson: (Here, too.)

ethomson:
👍 Thanks.  :D

novalis:
(done)

ethomson:
Thanks!

PR #3865, opened 2016-07-19
Fix leaks, some warnings and an error

ethomson:
Fix some leaks, and, oh, right, a terrible buffer overflow that I introduced in 1cd65991.


ethomson:
Having Windows 64 be completely broken is bumming me out.  I'm going to merge this later today so that we can get through the backlog of pull requests that have piled up.

PR #3869, opened 2016-07-23
Fix outdated comment

richardipsum:
Hey,

So this is a very trivial pr,
the original comment is from 2012 and seems to no longer apply.
The contribution guidelines suggest // style comments should be avoided,
so I figure I may as well fix that up while I'm here.

Apologies if this is a waste of reviewer time,
but I was working in the perl binding and this comment misled me.

Hope this helps! :)


ethomson:
Thanks for your careful eye!  :D

PR #3877, opened 2016-07-29
sysdir: don't assume an empty dir is uninitialized

### LINKED ISSUE #3871: sysdir handling initialization racy
See https://github.com/libgit2/libgit2/blob/master/src/sysdir.c#L140

This tries to lazily initialize git_sysdir__dirs, but does so in a thread-unsafe manner. In particular if two threads call git_open_repository(), both threads will race for the initialization, causing error messages from ThreadSanitizer.

ethomson:
Don't assume that an empty system directory buffer is uninitialized.
In fact, there simply may be no appropriate path for that value.
(For example, the Windows-specific programdata directory has no value
on non-Windows machines.)

This prevents us from continually trying to re-lookup these values,
which could get racy if two different threads are each calling
`git_sysdir_get` and trying to lookup / clear the value simultaneously.

Fixes #3871

### DIFF
--- a/src/sysdir.c
+++ b/src/sysdir.c
@@ -83,45 +83,43 @@ static int git_sysdir_guess_template_dirs(git_buf *out)
 #endif
 }
 
-typedef int (*git_sysdir_guess_cb)(git_buf *out);
-
-static git_buf git_sysdir__dirs[GIT_SYSDIR__MAX] =
-	{ GIT_BUF_INIT, GIT_BUF_INIT, GIT_BUF_INIT, GIT_BUF_INIT, GIT_BUF_INIT };
-
-static git_sysdir_guess_cb git_sysdir__dir_guess[GIT_SYSDIR__MAX] = {
-	git_sysdir_guess_system_dirs,
-	git_sysdir_guess_global_dirs,
-	git_sysdir_guess_xdg_dirs,
-	git_sysdir_guess_programdata_dirs,
-	git_sysdir_guess_template_dirs,
+struct git_sysdir__dir {
+	git_buf buf;
+	int (*guess)(git_buf *out);
 };
 
-static int git_sysdir__dirs_shutdown_set = 0;
+static struct git_sysdir__dir git_sysdir__dirs[] = {
+	{ GIT_BUF_INIT, git_sysdir_guess_system_dirs },
+	{ GIT_BUF_INIT, git_sysdir_guess_global_dirs },
+	{ GIT_BUF_INIT, git_sysdir_guess_xdg_dirs },
+	{ GIT_BUF_INIT, git_sysdir_guess_programdata_dirs },
+	{ GIT_BUF_INIT, git_sysdir_guess_template_dirs },
+};
+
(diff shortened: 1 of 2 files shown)


src/sysdir.c
```diff
@@ -85,8 +85,12 @@ static int git_sysdir_guess_template_dirs(git_buf *out)
...
 typedef int (*git_sysdir_guess_cb)(git_buf *out);
 
-static git_buf git_sysdir__dirs[GIT_SYSDIR__MAX] =
-	{ GIT_BUF_INIT, GIT_BUF_INIT, GIT_BUF_INIT, GIT_BUF_INIT, GIT_BUF_INIT };
+struct git_sysdir__dir {
+	int initialized;
+	git_buf buf;
+};
+
+static struct git_sysdir__dir git_sysdir__dirs[GIT_SYSDIR__MAX];
```
pks-t: Hum. If you're already creating a new sysdir struct, what do you think about including the guess callbacks into the struct and then drop the `git_sysdir__dir_guess` array? So basically

```
static struct git_sysdir__dir dirs[] = {
    { git_sysdir_guess_system_dirs, GIT_BUF_INIT },
    { git_sysdir_guess_global_dirs, GIT_BUF_INIT },
    ...
};
```

This would make the correlation of these two structures clearer and reduce a bit of its magic.

src/sysdir.c
```diff
@@ -85,8 +85,12 @@ static int git_sysdir_guess_template_dirs(git_buf *out)
 
 typedef int (*git_sysdir_guess_cb)(git_buf *out);
 
-static git_buf git_sysdir__dirs[GIT_SYSDIR__MAX] =
-	{ GIT_BUF_INIT, GIT_BUF_INIT, GIT_BUF_INIT, GIT_BUF_INIT, GIT_BUF_INIT };
+struct git_sysdir__dir {
+	int initialized;
```
hanwen: this looks a littlie bit clumsy, tbh.

Why not simply

  static git_buf *sysdirs 

int init_sysdirs() {
   if (sysdirs != NULL) { return }
   int (**guess)(git_buf *out) = {
      guess_system_dirs, guess_global_dir, ... , NULL };
   sz = 0;
   for (g = guess; *g; g++) { sz ++ }
   sysdirs = malloc(sz*sizeof(git_buf*));

   i = 0;
   for g = guess; *g; g++, i++ { 
     g(&sysdirs[i]);
  }
}

ethomson: The rationale is that we may not need to query system directories, especially for bare repository users.  They _must_ configure their system directories (if any) after calling `git_libgit2_init`, but they would want to avoid hitting the disk for unnecessary lookups that they were going to override.

ethomson: Nope, I'm wrong.  I didn't see the `git_sysdirs_global_init` being called by `git_libgit2_init`.

You're right:  this has become fairly overwrought.

ethomson:
Thanks to both @pks-t and @hanwen, I think that this is a big improvement.  I've incorporated both of your fine ideas.

Issue #3873, opened 2016-07-28
git_remote_fetch() with refspec "<branch>:" fails, Git does not


psprint:
Luckily, refspec "&lt;branch&gt;" does what "&lt;branch&gt;:" should, so there is a workaround, and ability to utilize FETCH_HEAD's non-durability properties, e.g. to just peek at a repo given via URL, doing git log ..FETCH_HEAD, without incorporating anything durable from the URL.

What I observe is:

Progress:  0
Progress:  0.25
Progress:  nan
Progress:  nan
Progress:  1

(the nans are not a problem). After that git_remote_fetch() returns GIT_ERROR (1), and diff -r after-fetch.git no-fetch.git shows: 

Binary files after-fetch.git/.git/index and no-fetch.git/.git/index differ
Only in after-fetch.git/.git/objects/pack: pack-49c8a6117d6a4fbb155e27cd458bf6941f195cef.idx
Only in after-fetch.git/.git/objects/pack: pack-49c8a6117d6a4fbb155e27cd458bf6941f195cef.pack

Wonder what was changed in index and how to obtain fully clean state?

carlosmn:
I don't know what's changed in your index, but it has nothing to do with fetching. You can use tools like `git ls-files -s` to see what the contents are. The index contains some extra information which can change depending on the exact steps you've taken, so a difference is not surprising.

Similarly with the packfiles, their names depend on their exact content, so having different ones in different repositories is perfectly normal.

I don't know what you mean by "fully clean state". That's only a thing that Git defines as it relates to the worktree, but that again is unrelated to performing network operations.

carlosmn:
The colon w/o rhs fetching issue was fixed via #3882

PR #3882, opened 2016-08-04
refspec: do not set empty rhs for fetch refspecs

pks-t:
According to git-fetch(1), "[t]he colon can be omitted when \<dst\>
is empty." So according to git, the refspec "refs/heads/master"
is the same as the refspec "refs/heads/master:" when fetching
changes. When trying to fetch from a remote with a trailing
colon with libgit2, though, the fetch actually fails while it
works when the trailing colon is left out. So obviously, libgit2
does _not_ treat these two refspec formats the same for fetches.

The problem results from parsing refspecs, where the resulting
refspec has its destination set to an empty string in the case of
a trailing colon and to a `NULL` pointer in the case of no
trailing colon. When passing this to our DWIM machinery, the
empty string gets translated to "refs/heads/", which is simply
wrong.

Fix the problem by having the parsing machinery treat both cases
the same for fetch refspecs.

### DIFF
--- a/src/refspec.c
+++ b/src/refspec.c
@@ -53,8 +53,10 @@ int git_refspec__parse(git_refspec *refspec, const char *input, bool is_fetch)
 
 	if (rhs) {
 		size_t rlen = strlen(++rhs);
-		is_glob = (1 <= rlen && strchr(rhs, '*'));
-		refspec->dst = git__strndup(rhs, rlen);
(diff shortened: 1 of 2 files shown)


pks-t:
Refs #3873

pks-t:
@carlosmn @ethomson: any comments on this PR?

PR #3884, opened 2016-08-05
Patch image initialization

pks-t:
Fixes two build warnings, for example https://travis-ci.org/libgit2/libgit2/jobs/149891598


ethomson:
Thanks!

PR #3889, opened 2016-08-07
Don't normalise symbolic ref target

richardipsum:
git-symbolic-ref(1) allows creation of symbolic refs whose target is
not a valid ref name.

So
    $ git symbolic-ref refs/heads/foo bar
	$ cat .git/refs/heads/foo
	ref: bar

where as attempting the same via libgit2 raises an error:
`The given reference name 'bar' is not valid`

### DIFF
--- a/src/refs.c
+++ b/src/refs.c
@@ -402,12 +402,7 @@ static int reference__create(
 
 		ref = git_reference__alloc(normalized, oid, NULL);
 	} else {
-		git_refname_t normalized_target;
-
-		if ((error = reference_normalize_for_repo(normalized_target, repo, symbolic)) < 0)
-			return error;
-
-		ref = git_reference__alloc_symbolic(normalized, normalized_target);
+		ref = git_reference__alloc_symbolic(normalized, symbolic);
 	}
 
 	GITERR_CHECK_ALLOC(ref);


ethomson:
I'm not sure how I feel about this change, but this doesn't match `git`'s behavior.  As you pointed out, it does allow some lax behavior, but it validates the target of `HEAD`:

```
% git symbolic-ref refs/heads/foo bar
% cat .git/refs/heads/foo 
ref: bar
% git symbolic-ref HEAD bar
fatal: Refusing to point HEAD outside of refs/
```

richardipsum:
Okay I wasn't aware of that.

So validation of symbolic refs should have a separate validation process then?

carlosmn:
This looks like a bug in git, or just a happens-to-work situation due the historically lax treatment of inputs.

richardipsum:
If it is a bug then that would be a shame, for me at least. Since I was planning on making use of the current behaviour. Also imo git's current behaviour is not dissimilar to traditional symbolic links, in that they don't really validate the value passed to them.

peff:
You can do:

```
$ git symbolic-ref refs/heads/foo bar
$ git rev-parse HEAD >.git/bar
$ git rev-parse foo
bd37aaf63f14c831860a7d6e5eba4f1ae6880142
```

Which is possibly useful, but kind of gross. But I think the attitude in core git has been that we don't necessarily guarantee the behavior of symbolic-refs besides `HEAD`. Everything else is "happens to work".

There aren't specific plans to change it, but I don't know if I would actively build on top of it.

peff:
BTW, there _is_ different behavior for `HEAD`, mostly because if you screw it up, then git will not consider the directory to be a repository anymore (so you can't `git symbolic-ref` your way out of it; you have to hand-edit the file to repair it).

ethomson:
So I don't like this change (as-is) for a number of reasons.

First, normalization really is _normalization_ not just validation.  For example: fancy Unicode handling on Mac OS.  And you're turning all of that off.  That's probably not good, when all you really want to do is permit targets that aren't proper refs.

Second, I don't think that this should be at all the default.  I think this validation is good.

However, if you want to create symbolic refs that point to non-refs, I'm not really here to stop you.  I suspect you're using Git (and libgit2) for something a bit custom, and that's cool.

I think that if you wanted to turn off target validation optionally (and that option defaulted to _on_, like it is today) then that would be reasonable.  (See for example `git_libgit2_opts(GIT_OPT_ENABLE_STRICT_OBJECT_CREATION, 0)` for an example of target validation when creating objects.)

I don't really know whether you want to turn the rest of normalization off when you do this or not.  I would recommend _not_, since these generally point to other files, not to just an array of bytes.

If this is an avenue you want to pursue then I would encourage you to write unit tests here, too.

richardipsum:
Okay, thanks for your thoughts on this. :)
I will try to rework this so that target validation can be disabled optionally.

richardipsum:
Hey :), I've opened a new pull request (https://github.com/libgit2/libgit2/pull/3912) which is hopefully a better solution than this one, thanks!

Issue #3885, opened 2016-08-05
Segfault in stransport_connect in Darwin


the-kenny:
Hi,

while packaging [Rust](https://rust-lang.org) and [Cargo](https://crates.io) (the package manager) in for Nix (a declarative package manager used in [NixOS](https://nixos.org), we ran into a strange segfault when using `libgit2` on Darwin:

```
#0  0x000000010076aac8 in CFRelease () from /nix/store/7akpgkhmbmdc9nj69wv0wr01pp9z92x6-CF-osx-10.9.5/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation
#1  0x000000010038e31b in stransport_connect ()
#2  0x00000001003979de in http_connect ()
#3  0x0000000100396f46 in http_action ()
#4  0x000000010039a184 in git_smart.connect ()
#5  0x0000000100383833 in git_remote_connect ()
#6  0x0000000100384002 in git_remote_fetch ()
#7  0x0000000100329504 in git2::remote::Remote::fetch::hf22a703e28682659 ()
#8  0x000000010021bda7 in cargo::sources::git::utils::fetch::_$u7b$$u7b$closure$u7d$$u7d$::h4505ef50fa16b942 ()
#9  0x0000000100214bab in cargo::sources::git::utils::fetch::h969cb9309bcaae70 ()
#10 0x0000000100237a2c in cargo::sources::registry::RegistrySource::do_update::hf310bfc89b20f31c ()
#11 0x000000010008ee5b in _$LT$sources..registry..RegistrySource$LT$$u27$cfg$GT$$u20$as$u20$core..source..Source$GT$::update::h2ca6fb87f187829e ()
#12 0x00000001000de130 in cargo::core::registry::PackageRegistry::ensure_loaded::h756a0c74cb30f33c ()
#13 0x00000001000e4d6b in _$LT$core..registry..PackageRegistry$LT$$u27$cfg$GT$$u20$as$u20$core..registry..Registry$GT$::query::h2b3b59a668f8397b ()
#14 0x00000001000cd272 in _$LT$$RF$$u27$a$u20$mut$u20$I$u20$as$u20$std..iter..Iterator$GT$::next::h2ee138e44a29d8c2 ()
#15 0x00000001000c6811 in cargo::core::resolver::Context::build_deps::h5d514997c1c12bc9 ()
#16 0x00000001000bedc4 in cargo::core::resolver::activate::h84824bf29ba071ae ()
[...]

PR #3891, opened 2016-08-08
stransport memory management improvements

pks-t:
Some fixes to memory management I've encountered when reasoning about #3885. While these fixes probably won't fix the problems encountered in this ticket, I think they are a worthwhile improvement never the less.

### DIFF
--- a/src/stransport_stream.c
+++ b/src/stransport_stream.c
@@ -261,6 +261,7 @@ int git_stransport_stream_new(git_stream **out, const char *host, const char *po
 	st->ctx = SSLCreateContext(NULL, kSSLClientSide, kSSLStreamType);
 	if (!st->ctx) {
 		giterr_set(GITERR_NET, "failed to create SSL context");
+		git__free(st);
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


pks-t:
Embarassing typo... should probably wait until I've got a OS X VM at hand to test changes. Meh

src/stransport_stream.c
```diff
@@ -230,7 +230,8 @@ static void stransport_free(git_stream *stream)
 	stransport_stream *st = (stransport_stream *) stream;
 
 	git_stream_free(st->io);
-	CFRelease(st->ctx);
+	if (st->ctx)
+		CFRelease(st->ctx);
```
carlosmn: When would this be `NULL`? If we fail to create the context, we error out on the constructor itself.

pks-t: Agreed. The only reasoning behind this was as a safeguard. Anyway, might as well just drop it if you deem this to be unnecessary. The other fix is still valid, I guess.

carlosmn: It's either this or the other block. With this patch we avoid calling the free function when it's `NULL` so it's either one or the other.

pks-t: Okay, I've dropped this patch. Note though that previously we haven't been calling the free function either, as at the point of time when calling it we didn't yet initialize the `free` function pointer.

PR #3892, opened 2016-08-08
Use a shared buffer in calls of git_treebuilder_write to avoid heap contention

mitesch:
When doing many merges in parallel in a large repository, we saw performance degrade rapidly due to memory allocations in git_treebuilder_write. This change uses a shared buffer created in git_tree__write_index to avoid the small allocation per tree in git_treebuilder_write.

In our tests against that repository we saw improvements of ~50% in merges under heavy load.

### DIFF
--- a/src/tree.c
+++ b/src/tree.c
@@ -515,7 +515,8 @@ static int write_tree(
 	git_repository *repo,
 	git_index *index,
 	const char *dirname,
-	size_t start)
+	size_t start,
+	git_buf *shared_buf)
 {
 	git_treebuilder *bld = NULL;
 	size_t i, entries = git_index_entrycount(index);
@@ -568,7 +569,7 @@ static int write_tree(
 			GITERR_CHECK_ALLOC(subdir);
 
 			/* Write out the subtree */
-			written = write_tree(&sub_oid, repo, index, subdir, i);
+			written = write_tree(&sub_oid, repo, index, subdir, i, shared_buf);
 			if (written < 0) {
 				git__free(subdir);
 				goto on_error;
@@ -600,7 +601,7 @@ static int write_tree(
 		}
 	}
 
-	if (git_treebuilder_write(oid, bld) < 0)
+	if (git_treebuilder_write_with_buffer(oid, bld, shared_buf) < 0)
 		goto on_error;
 
 	git_treebuilder_free(bld);
@@ -616,6 +617,7 @@ int git_tree__write_index(
 {
 	int ret;
 	git_tree *tree;
+	git_buf shared_buf = GIT_BUF_INIT;
 	bool old_ignore_case = false;
 
 	assert(oid && index && repo);
@@ -641,7 +643,8 @@ int git_tree__write_index(
 		git_index__set_ignore_case(index, false);
 	}
 
-	ret = write_tree(oid, repo, index, "", 0);
+	ret = write_tree(oid, repo, index, "", 0, &shared_buf);
+	git_buf_free(&shared_buf);
 
 	if (old_ignore_case)
 		git_index__set_ignore_case(index, true);
@@ -797,18 +800,35 @@ int git_treebuilder_remove(git_treebuilder *bld, const char *filename)
 
 int git_treebuilder_write(git_oid *oid, git_treebuilder *bld)
 {
+	int error;
+	git_buf buffer = GIT_BUF_INIT;
+
+	error = git_treebuilder_write_with_buffer(oid, bld, &buffer);
+
+	git_buf_free(&buffer);
+	return error;
+}
+
+int git_treebuilder_write_with_buffer(git_oid *oid, git_treebuilder *bld, git_buf *tree)
+{
 	int error = 0;
 	size_t i, entrycount;
-	git_buf tree = GIT_BUF_INIT;
 	git_odb *odb;
 	git_tree_entry *entry;
 	git_vector entries;
 
 	assert(bld);
+	assert(tree);
+
+	git_buf_clear(tree);
 
 	entrycount = git_strmap_num_entries(bld->map);
 	if (git_vector_init(&entries, entrycount, entry_sort_cmp) < 0)
 		return -1;
+
+	if (tree->asize == 0 &&
+		(error = git_buf_grow(tree, entrycount * 72)) < 0)
+		return error;
 
 	git_strmap_foreach_value(bld->map, entry, {
 		if (git_vector_insert(&entries, entry) < 0)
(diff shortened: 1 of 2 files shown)


ethomson:
Cool, that sounds like a great improvement!

First, please revert any changes that are only whitespace - it seems like a bunch of those snuck in.  This will help avoid unnecessary churn.

Second, I think that a more straightforward change would be to give `git_treebuilder` its own `git_buf` that it reuses - that will ensure that all users of the treebuilder get this improvement, not just `git_tree__write_index`.  Does that seem reasonable?

src/tree.c
```diff
@@ -797,14 +804,35 @@ int git_treebuilder_remove(git_treebuilder *bld, const char *filename)
...
+	int error;
+	size_t entrycount;
+	git_buf buffer = GIT_BUF_INIT;
+
+	entrycount = git_strmap_num_entries(bld->map);
+
+	/* Grow the buffer beforehand to an estimated size */
+	error = git_buf_grow(&buffer, entrycount * 72);
+	if (error >= 0)
+	{
```
pks-t: We usually avoid braces around single-statement branches

src/tree.c
```diff
@@ -641,7 +643,12 @@ int git_tree__write_index(
 		git_index__set_ignore_case(index, false);
 	}
 
-	ret = write_tree(oid, repo, index, "", 0);
+	/* Grow the buffer beforehand to 1kb */
+	if ((ret = git_buf_grow(&shared_buf, 1024)) >= 0)
+	{
```
pks-t: For single-line conditionals, we keep the opening brace on the same line

pks-t:
Small nit on the commit message itself: please try to keep the initial line short, as otherwise many git commands become hard to read (e.g. `git log --oneline`). I'd propose something along the lines of 

```
treebuilder: use shared buffer for writing trees

The function to write trees allocates a new buffer for each tree.
This causes problems with performance when performing a lot
of actions involving writing trees, e.g. when doing many merges.
Fix the issue by instead handing in a shared buffer, which is then
re-used across the calls without having to re-allocate between
calls.
```

And I agree with @ethomson on that it would be nicer to have a buffer at the treebuilder itself which would be reused for other functions, as well. 

Anyway, this heads in a good direction. Thanks for contributing :)

mitesch:
Placing the buffer on the builder is a bit tricky, I think. write_tree creates a new treebuilder each call and only calls git_treebuilder_write once with it, so I don't think just that change would reduce the allocations. The builder can't be shared across write_tree calls because of the recursion in the loop. 

I could put the buffer on the treebuilder, but I'd have to manage the buffer the same way I do now and just set bld->buf in write_tree. That doesn't seem better though.

ethomson:
Aha - thanks for clarifying, @mitesch - obviously you're quite right and I didn't understand the lifecycle there when I was re-reading it.

In that case, yes, I think this is great.  @pks-t 's correct about our style - we follow the kernel / git styles.  If you wouldn't mind squashing a fixup commit that matches the style, that would be great!

src/tree.c
```diff
@@ -641,7 +643,12 @@ int git_tree__write_index(
...
+		ret = write_tree(oid, repo, index, "", 0, &shared_buf);
+		git_buf_free(&shared_buf);
```
carlosmn: The free doesn't need to be conditional, it'll do the right thing in case of OOM.

src/tree.c
```diff
@@ -797,14 +804,35 @@ int git_treebuilder_remove(git_treebuilder *bld, const char *filename)
...
 {
+	int error;
+	size_t entrycount;
+	git_buf buffer = GIT_BUF_INIT;
+
+	entrycount = git_strmap_num_entries(bld->map);
+
+	/* Grow the buffer beforehand to an estimated size */
+	error = git_buf_grow(&buffer, entrycount * 72);
+	if (error >= 0)
```
carlosmn: This is the opposite of how we do error handling in the library. We use early returns or jumps to cleanup if things go wrong instead of putting the success path inside `if`s.

src/tree.c
```diff
@@ -797,14 +803,33 @@ int git_treebuilder_remove(git_treebuilder *bld, const char *filename)
...
 int git_treebuilder_write(git_oid *oid, git_treebuilder *bld)
 {
+	int error;
+	size_t entrycount;
+	git_buf buffer = GIT_BUF_INIT;
+
+	entrycount = git_strmap_num_entries(bld->map);
+
+	/* Grow the buffer beforehand to an estimated size */
+	error = git_buf_grow(&buffer, entrycount * 72);
```
carlosmn: Why is the caller growing the buffer instead of leaving the decision to the consumer?

src/tree.c
```diff
@@ -797,14 +803,33 @@ int git_treebuilder_remove(git_treebuilder *bld, const char *filename)
...
+
+	git_buf_free(&buffer);
+	return error;
+}
+
+int git_treebuilder_write_buffered(git_oid *oid, git_treebuilder *bld, git_buf *shared_buf)
+{
 	int error = 0;
 	size_t i, entrycount;
-	git_buf tree = GIT_BUF_INIT;
```
carlosmn: The rename from `tree` to `shared_buf` adds noise to the change without aiding to the readability of the code. It's _still_ the buffer where we store the tree, it just happens to be an input.

src/tree.c
```diff
@@ -641,7 +643,11 @@ int git_tree__write_index(
 		git_index__set_ignore_case(index, false);
 	}
 
-	ret = write_tree(oid, repo, index, "", 0);
+	/* Grow the buffer beforehand to 1kb */
+	if ((ret = git_buf_grow(&shared_buf, 1024)) >= 0)
```
carlosmn: Why grow it to 1kB instead of leaving that to the tree-writing function?

src/tree.h
```diff
@@ -42,6 +42,8 @@ extern int git_tree_entry_icmp(const git_tree_entry *e1, const git_tree_entry *e
 void git_tree__free(void *tree);
 int git_tree__parse(void *tree, git_odb_object *obj);
 
+int git_treebuilder_write_buffered(git_oid *oid, git_treebuilder *bld, git_buf *shared_buf);
```
carlosmn: The name `_buffered` sounds like it's related to buffered I/O, or that the buffer is the output. Something like `_with_buffer` would be more in line with our naming.

carlosmn:
Mostly some style/nomenclature comments, but overall it's looking good. It might be worth making this part of the public API so external users of this function can make use of it as well.

The tree update might also benefit from doing something like this, which is another semi-heavy user of this.

src/tree.c
mitesch: In `git_treebuilder_write` I wanted to be a bit more conservative with the buffer allocation since it is not going to be reused. 

In `git_tree__write_index` I was going for more consistent performance by reducing the number of growths required. 

The previous buffer guess was `entrycount * 72`, so 1024 bytes is about 14 entries worth of buffer space. From there it would grow by 1.5x as needed.

I think leaving the initial estimate of `entrycount * 72` in `git_treebuilder_write_with_buffer` will in the worst case result in maybe 7 more resizes per merge, so I'm not too worried though. 

I could go either way, what do you think?

mitesch:
Pushed a fixup with most of the requested changes.
Couple things still open:
- Where the initial buffer growth should be.
- Should I make this part of the public API?
- Build failure. Looks unrelated, but not 100% confident.

mitesch:
@ethomson @carlosmn squashed and should be good to go

src/tree.h
```diff
@@ -42,6 +42,8 @@ extern int git_tree_entry_icmp(const git_tree_entry *e1, const git_tree_entry *e
 void git_tree__free(void *tree);
 int git_tree__parse(void *tree, git_odb_object *obj);
 
+int git_treebuilder_write_with_buffer(git_oid *oid, git_treebuilder *bld, git_buf *tree);
```
ethomson: Sorry - one more minor nit.  For functions that are internal to the library, we use a double-underscore.  For example, this would be `git_treebuilder__write_with_buffer`.

But there's no need to put this in a header at all; what I would prefer to see is to revert the changes to `tree.h` and instead declare `write_with_buffer` as `static` in `tree.c`.  Would you be so kind?  Thanks again!

carlosmn: The other direction to go to would be to make it part of the public API. This declaration would then be moved to `include/git2/tree.h` and would need some documentation. I'd be in favour of making it public as it can be useful for external users of the tree builder as well.

mitesch:
@ethomson @carlosmn Updated, sorry for the delay. I moved the definition to `include/git2/tree.h`. Let me know if the documentation is sufficient.

PR #3894, opened 2016-08-10
Fix - Negation of rules

tgandrews:
Currently, negative rules are thrown away if they do not affect a rule before.
This change adds a test which failed before but works now that we no longer
throw away negative rules.

I have tried to prove this with tests but I think more are required to cover this.

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -170,10 +170,6 @@ static int parse_ignore_file(
 
 			scan = git__next_line(scan);
 
-			/* if a negative match doesn't actually do anything, throw it away */
-			if (match->flags & GIT_ATTR_FNMATCH_NEGATIVE)
-				error = does_negate_rule(&valid_rule, &attrs->rules, match);
-
 			if (!error && valid_rule)
 				error = git_vector_insert(&attrs->rules, match);
 		}
@@ -580,4 +576,3 @@ int git_ignore__check_pathspec_for_exact_ignores(
 
 	return error;
 }
-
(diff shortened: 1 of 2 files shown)


tgandrews:
I think the failing tests are wrong but have not removed them so that we can have a discussion.

pks-t:
No, the tests are actually correct in that they match the behavior of git.git. The following little hacky script tests this against upstream:

```
#!/bin/sh

REPO="$(mktemp -d)"

trap "rm -rf \"$REPO\"" EXIT

cd "$REPO"
git init -q

cat > .gitignore <<EOF
foo.txt
parent/child1
parent/child2
parent/child4
parent/nested/child5
nested/child6
nested/child7
padded_parent/child8
!parent/child1
!parent/child2/bar.txt
!parent/child3
!child4
!nested/child5
!nested/child6
!child7/
!_parent/child8
top
!top/mid/btm
zoo/*
!zoo/bar
!zoo/foo/bar
EOF

for path in "padded_parent/child8/bar.txt" "parent/child2/bar.txt" "parent/child3/foo.txt" "parent/nested/child5/bar.txt" "top/mid/btm/tracked" "top/mid/btm/untracked"
do
    if ! git check-ignore -q "$path"
    then
        echo "$path should have been ignored"
        exit 1
    fi
done

for path in "parent/child1/bar.txt" "parent/child4/bar.txt" "parent/nested/child6/bar.txt" "parent/nested/child7/bar.txt" "foo/bar"
do
    if git check-ignore -q "$path"
    then
        echo "$path should not have been ignored"
        exit 1
    fi
done
```

tgandrews:
OK -  I have clearly misunderstood this.

The problem I am having is the scenario I have created the test for:

A .gitignore containing
```
bar.txt
!parent/child/bar.txt
```

Currently libgit2 says the child bar.txt is ignored but git does not.

pks-t:
Well, you're actually correct with your observation that we fail to treat your case correctly. Created a new PR which should fix your issue.

PR #3897, opened 2016-08-16
Squelch example warnings, enable CI

pks-t:
Recently, I wanted to build our examples and was hugely annoyed by all those warnings that were spilled left and right. So I cleaned up our tests, particularly the `examples/general.c` one, so that no warnings get thrown anymore.

Furthermore, I've changed CI scripts to build all tests to catch errors early on. I know that the examples aren't of huge importance, but considering they might be the first contact point with libgit2 for new developers, I'd say they should be reasonably clean and, as a bare minimum, compile. So that's why I think it is only sensible to break our CI builds when changes inside libgit2 break our examples.

### DIFF
--- a/examples/general.c
+++ b/examples/general.c
@@ -43,6 +43,7 @@
 #include <stdio.h>
 #include <string.h>
 
+static void index_walking(git_repository *repo);
 static void reference_listing(git_repository *repo);
 static void config_files(const char *repo_path);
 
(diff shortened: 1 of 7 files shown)


pks-t:
The test failure is unrelated

examples/general.c
```diff
@@ -12,519 +12,604 @@
...
-// [lg]: http://libgit2.github.com
-// [ap]: http://libgit2.github.com/libgit2
-// [pg]: https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain
-
-// ### Includes
-
-// Including the `git2.h` header will include all the other libgit2 headers
-// that you need.  It should be the only thing you need to include in order
-// to compile properly and get all the libgit2 API.
+/* [**libgit2**][lg] is a portable, pure C implementation of the Git core
```
carlosmn: This would hide this text from the docco output. We need C++ comments or `/**` here for it to show.

pks-t: Now that you mention docco: my patches have another problem in that they change the order of documentation. With these patches, we will not have a logical order of initializing the library, opening repo, etc., but instead first describe all functions and at long last initialize the library.

I guess I'll re-order this such that the single functions will go after the main function.

pks-t:
Okay, fixed up comments such that they appear in the correct order and are evaluated as documentation.

PR #3900, opened 2016-08-17
transports: http: set substream as disconnected after closing

pks-t:
When calling `http_connect` on a subtransport whose stream is already
connected, we first close the stream in case no keep-alive is in use.
When doing so, we do not reset the transport's connection state,
though. Usually, this will do no harm in case the subsequent connect
will succeed. But when the connection fails we are left with a
substransport which is tagged as connected but which has no valid
stream attached.

Fix the issue by resetting the subtransport's connected-state when
closing its stream in `http_connect`.

### DIFF
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -599,6 +599,7 @@ static int http_connect(http_subtransport *t)
 		git_stream_close(t->io);
 		git_stream_free(t->io);
 		t->io = NULL;
+		t->connected = 0;
 	}
 
 	if (t->connection_data.use_ssl) {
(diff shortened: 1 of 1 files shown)


carlosmn:
This comes from investigating #3885 I'm assuming? I suspected something like this, but couldn't quite find where we might be going wrong.

pks-t:
Yeah, I spotted this one when debugging. I don't think this fixes #3885, though, as we actually segfault in `stransport` code. Considering we already set the sub transport to `NULL`, this should have no impact. Still, this is a bug which might surface in different situations.

pks-t:
Ugh. We also didn't reset the `connected` flag in `http_close`.

PR #3903, opened 2016-08-18
Handle cred_acquire_cb returning no credentials

### LINKED ISSUE #3902: cred_acquire_cb not returning credentials causes a segmentation fault
```ruby
require 'rugged'

def credentials
  lambda do |_, _, _|
    return nil
  end
end

Rugged::Repository.clone_at 'https://github.com/example/my-project.git', 'my-project' , { credentials: credentials }
```

Calling `clone_at` with a credentials callback that returns nil against a repository that requires authentication causes a segmentation fault. When `on_headers_complete` tries to dereference `t->cred->credtype` (http.c:379) and `t->cred` is empty the segfault is triggered.

brandonweeks:
Resolves #3902.

### DIFF
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -374,7 +374,10 @@ static int on_headers_complete(http_parser *parser)
 					t->error = error;
 					return t->parse_error = PARSE_ERROR_EXT;
 				} else {
-					assert(t->cred);
+					if (!t->cred) {
+						giterr_set(GITERR_NET, "authentication required but no creds returned");
+						return t->parse_error = PARSE_ERROR_GENERIC;
+					}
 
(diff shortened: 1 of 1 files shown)


ethomson:
Hmm, I don't agree with this change.  libgit2 uses a pattern of `assert`s where there is programmer error.  This is such a case, and it should not return an error.

If the `cred_acquire_cb` returns 0 then that's a guarantee that they have populated `t->cred`.  If they have not, we _should_ assert.

This looks like an issue in rugged; it should return `GIT_PASSTHROUGH` or an error code - it should not be returning `0` and failing to set `t->cred`.

brandonweeks:
Fair enough, I'll take a look at fixing it in rugged instead. Thank you for the tips. 👍

PR #3904, opened 2016-08-19
filesystem_iterator: fixed double free on error

hackhaslam:
`git_iterator_free` tries to free `iter->root` (and `iter->current_path`) via the `filesystem_iterator_free` callback after it has already been freed. I couldn't figure out what caused the `on_error` case to begin with or else I would have tried to make a test case.

### DIFF
--- a/src/iterator.c
+++ b/src/iterator.c
@@ -1834,6 +1834,9 @@ static int iterator_for_filesystem(
 	iter = git__calloc(1, sizeof(filesystem_iterator));
 	GITERR_CHECK_ALLOC(iter);
 
+	iter->base.type = type;
+	iter->base.cb = &callbacks;
+
 	root_len = strlen(root);
 
 	iter->root = git__malloc(root_len+2);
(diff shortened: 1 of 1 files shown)


pks-t:
The fix as-is does not suffice, unfortunately. Problem is that we may jump to the `on_error` label previous to initializing `iterator->base.cb`, where we call `git_iterator_free(&iterator->base)`. But `git_iterator_free` will dereference `iterator->base.cb->free` then, leading to undefined behavior.

So in addition to your improvement, we should also move up initialization of the callbacks.

hackhaslam:
Ugh, sorry. You're right. I didn't pay close enough attention to where the callbacks are set. I moved the initialization of the base members above the initialization of the filesystem members.

pks-t:
No worries, the error existed already before you touching the code. Thanks for the fix. I'll wait another day for comments and then merge tomorrow if none are incoming.

ethomson:
Thanks @hackhaslam !  ❤️

PR #3907, opened 2016-08-23
git_checkout_tree options fix

steffhip:
According to the reference the git_checkout_tree and git_checkout_head
functions should accept NULL in the opts field

This was broken since the opts field was dereferenced and thus lead to a
crash.


Inesteem:
Yeah, this would fix my work, too:D
(Up to now it will only work using version 0.23)

ethomson:
Definitely looks like a good fix.  Could I also trouble you to add a test that calls `git_checkout_tree` with a `NULL` opts? Otherwise I think that I'm likely to break this again.  ;)

Thanks!

pks-t:
Thanks for the fix and test. Considering it is only one test and we already have tests for `git_checkout_tree`, could you maybe move the test into `tests/checkout/tree.c`? Thanks :+1:

pks-t:
Manually squashed and merged via a08e88259fe7ef3d7514a4774acd6eec5a6a2ca7

pks-t:
Thanks for the fix :sparkles:

PR #3908, opened 2016-08-24
Teach `git_patch_from_diff` about parsed diffs

ethomson:
Ensure that `git_patch_from_diff` can return the patch for parsed diffs,
not just generate a patch for a generated diff.

### DIFF
--- a/src/diff.h
+++ b/src/diff.h
@@ -8,6 +8,7 @@
 #define INCLUDE_diff_h__
 
 #include "git2/diff.h"
+#include "git2/patch.h"
 #include "git2/sys/diff.h"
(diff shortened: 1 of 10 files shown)


pks-t:
Strange - why did AppVeyor not trigger for this PR?

ethomson:
I'll rebase on `master` to kick it.

ethomson:
I was a bit worried that valgrind was showing some memory leaks, but it seems that those are already existing in master.  Alas.  I'll try to find some time to look at that soon, but I'm going to ship this in the meantime.

/cc @brianmario

brianmario:
So rad thank you!!

PR #3912, opened 2016-08-27
Make symbolic ref target validation optional

richardipsum:
Introduce GIT_OPT_ENABLE_SYMBOLIC_REF_TARGET_VALIDATION option.
Setting this option to 0 allows
validation of a symbolic ref's target to be bypassed.
This option is enabled by default.

This mechanism is added primarily to address a discrepancy between git
behaviour and libgit2 behaviour, whereby the former allows the symbolic
ref target to carry an arbitrary string and the latter does not, so:

```
$ git symbolic-ref refs/heads/foo bar
$ cat .git/refs/heads/foo
ref: bar
```

where as attempting the same via libgit2 raises an error:

```
The given reference name 'bar' is not valid
```

this mechanism also allows those that might want to make use of
git's more lenient treatment of symbolic ref targets to do so.

### DIFF
--- a/include/git2/common.h
+++ b/include/git2/common.h
@@ -157,6 +157,7 @@ typedef enum {
 	GIT_OPT_SET_SSL_CERT_LOCATIONS,
 	GIT_OPT_SET_USER_AGENT,
 	GIT_OPT_ENABLE_STRICT_OBJECT_CREATION,
+	GIT_OPT_ENABLE_SYMBOLIC_REF_TARGET_VALIDATION,
 	GIT_OPT_SET_SSL_CIPHERS,
 	GIT_OPT_GET_USER_AGENT,
 } git_libgit2_opt_t;
@@ -270,6 +271,18 @@ typedef enum {
  *		> example, when this is enabled, the parent(s) and tree inputs
  *		> will be validated when creating a new commit.  This defaults
  *		> to disabled.
+ *
+ *	* opts(GIT_OPT_ENABLE_SYMBOLIC_REF_TARGET_VALIDATION, int enabled)
+ *
+ *		> Validate the target of a symbolic ref when creating it.
+ *		> For example, 'foobar' is not a valid ref,
+ *		> therefore 'foobar' is not a valid target
+ *		> for a symbolic ref by default,
+ *		> where as 'refs/heads/foobar' is.
+ *		> Disabling this bypasses validation so that an arbitrary
+ *		> strings such as 'foobar' can be used for a symbolic ref target.
+ *		> This defaults to enabled.
+ *
  *	* opts(GIT_OPT_SET_SSL_CIPHERS, const char *ciphers)
  *
  *		> Set the SSL ciphers use for HTTPS connections.
--- a/src/refs.c
+++ b/src/refs.c
@@ -24,6 +24,8 @@
 #include <git2/signature.h>
 #include <git2/commit.h>
 
+bool git_reference__enable_symbolic_ref_target_validation = true;
+
 GIT__USE_STRMAP
 
 #define DEFAULT_NESTING_LEVEL	5
@@ -175,10 +177,11 @@ int git_reference_name_to_id(
 	return 0;
 }
 
-static int reference_normalize_for_repo(
+static int reference__normalize_for_repo(
 	git_refname_t out,
 	git_repository *repo,
-	const char *name)
+	const char *name,
+	bool validate)
 {
 	int precompose;
 	unsigned int flags = GIT_REF_FORMAT_ALLOW_ONELEVEL;
@@ -187,7 +190,27 @@ static int reference_normalize_for_repo(
 		precompose)
 		flags |= GIT_REF_FORMAT__PRECOMPOSE_UNICODE;
 
+	if (!validate) {
+		flags |= GIT_REF_VALIDATION_DISABLE;
+	}
+
 	return git_reference_normalize_name(out, GIT_REFNAME_MAX, name, flags);
+}
+
+static int reference_normalize_for_repo(
+	git_refname_t out,
+	git_repository *repo,
+	const char *name)
+{
+	return reference__normalize_for_repo(out, repo, name, true);
+}
+
+static int reference_normalize_for_repo_without_validation(
+	git_refname_t out,
+	git_repository *repo,
+	const char *name)
+{
+	return reference__normalize_for_repo(out, repo, name, false);
 }
 
 int git_reference_lookup_resolved(
@@ -404,7 +427,13 @@ static int reference__create(
 	} else {
 		git_refname_t normalized_target;
 
-		if ((error = reference_normalize_for_repo(normalized_target, repo, symbolic)) < 0)
+		if (git_reference__enable_symbolic_ref_target_validation) {
+			error = reference_normalize_for_repo(normalized_target, repo, symbolic);
+		} else {
+			error = reference_normalize_for_repo_without_validation(normalized_target, repo, symbolic);
+		}
+
+		if (error < 0)
 			return error;
 
 		ref = git_reference__alloc_symbolic(normalized, normalized_target);
@@ -876,6 +905,7 @@ int git_reference__normalize_name(
 	int segment_len, segments_count = 0, error = GIT_EINVALIDSPEC;
 	unsigned int process_flags;
 	bool normalize = (buf != NULL);
+	bool validate = (flags & GIT_REF_VALIDATION_DISABLE) == 0;
 
 #ifdef GIT_USE_ICONV
 	git_path_iconv_t ic = GIT_PATH_ICONV_INIT;
@@ -886,7 +916,7 @@ int git_reference__normalize_name(
 	process_flags = flags;
 	current = (char *)name;
 
-	if (*current == '/')
+	if (validate && *current == '/')
 		goto cleanup;
 
 	if (normalize)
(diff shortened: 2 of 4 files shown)


richardipsum:
Hi again! If this approach seems vaguely reasonable I'll be quite happy to provide whatever tests may be needed, thanks!

richardipsum:
Hi, I opened this a while ago, it implements the approach suggested to me by this project when I attempted to do something similar with, https://github.com/libgit2/libgit2/pull/3889

I could really use this functionality in my own projects, so any feedback on the current implementation would be appreciated, even if you'd rather I just abandon this change.

Thanks,
Richard

richardipsum:
@ethomson Penny for your thoughts?

ethomson:
Hi @richardipsum - sorry about the delay.  I like this idea a lot, this is definitely worthwhile.  I'll add a few notes to the code itself.

Thanks so much for your patience.

include/git2/common.h
```diff
@@ -157,6 +157,7 @@ typedef enum {
 	GIT_OPT_SET_SSL_CERT_LOCATIONS,
 	GIT_OPT_SET_USER_AGENT,
 	GIT_OPT_ENABLE_STRICT_OBJECT_CREATION,
+	GIT_OPT_ENABLE_SYMBOLIC_REF_TARGET_VALIDATION,
```
ethomson: Just to be consistent with the name above it, could this be `GIT_OPT_ENABLE_STRICT_SYMBOLIC_REF_CREATION`?

include/git2/common.h
```diff
@@ -270,6 +271,18 @@ typedef enum {
...
  *		> will be validated when creating a new commit.  This defaults
  *		> to disabled.
+ *
+ *	* opts(GIT_OPT_ENABLE_SYMBOLIC_REF_TARGET_VALIDATION, int enabled)
+ *
+ *		> Validate the target of a symbolic ref when creating it.
+ *		> For example, 'foobar' is not a valid ref,
+ *		> therefore 'foobar' is not a valid target
+ *		> for a symbolic ref by default,
+ *		> where as 'refs/heads/foobar' is.
```
ethomson: Could you change these to backticks for readability in the documentation?  Also, one nitpick: `whereas` should be one word.

src/refs.c
```diff
@@ -187,9 +190,29 @@ static int reference_normalize_for_repo(
 		precompose)
 		flags |= GIT_REF_FORMAT__PRECOMPOSE_UNICODE;
 
+	if (!validate) {
```
ethomson: Stylistic nitpick: we don't put braces around single lines (unless there was another multiline condition).  Eg, we would prefer this as:

    if (!validate)
        flags |= GIT_REF_VALIDATION_DISABLE;

src/refs.c
```diff
@@ -404,7 +427,13 @@ static int reference__create(
 	} else {
 		git_refname_t normalized_target;
 
-		if ((error = reference_normalize_for_repo(normalized_target, repo, symbolic)) < 0)
+		if (git_reference__enable_symbolic_ref_target_validation) {
+			error = reference_normalize_for_repo(normalized_target, repo, symbolic);
+		} else {
+			error = reference_normalize_for_repo_without_validation(normalized_target, repo, symbolic);
+		}
```
ethomson: This is awkward, and we can drop both of these functions by simply passing the `bool` directly:

    error = reference_normalize_for_repo(normalized_target, repo, symbolic, git_reference__enable_symbolic_ref_target_validation);

src/refs.c
```diff
@@ -175,10 +177,11 @@ int git_reference_name_to_id(
 	return 0;
 }
 
-static int reference_normalize_for_repo(
+static int reference__normalize_for_repo(
```
ethomson: Sylistic nitpick: we use the double-underscores for things that are internal to the library (but have a scope outside the current file).  That lets us disambiguate public functions.  For example:

* `git_reference_some_function` would be exported and public to libgit2 callers.
* `git_reference__different_function` would be internal to the library

We generally don't do double-underscores for `static`s within a file.  I'm sure there are outliers to this rule, which is not a big deal, but I don't want to add new ones.

src/refs.h
```diff
@@ -53,6 +55,7 @@
 #define GIT_REFS_STASH_FILE GIT_REFS_DIR GIT_STASH_FILE
 
 #define GIT_REF_FORMAT__PRECOMPOSE_UNICODE	(1u << 16)
+#define GIT_REF_VALIDATION_DISABLE	(1u << 15)
```
ethomson: For this one - please do use double-underscores, since it's internal to the library, and the flags are all the  `GIT_REF_FORMAT_...` constants.  So I think this would be most clear as `GIT_REF_FORMAT__VALIDATION_DISABLE`.

ethomson:
I think that overall this is great - thanks!  ❤️  There are a few stylistic changes that I would like to see, and a unit test would be really great as well.  I think that the tests around strict object validation, and the existing reference tests should be indicative, but since this is your first time with our unit test framework, please let me know if I can help.

ethomson:
Added these fixups myself and merged manually.

richardipsum:
@ethomson Apologies I wasn't able to complete this change, once I added unit tests for this I found that this patch broke ref handling in several other places, and I never found the time to make the apparently necessary changes. If you have satisfied an equivalent test set then so much the better, if however these changes were merged without any supporting tests I'd honestly consider reverting the merge.

I hope this helps, and sorry I wasn't able to complete this PR in reasonable time.

richardipsum:
@ethomson please see https://github.com/richardipsum/libgit2/commit/bff4855455f4765615b93c5a132fc579b4a06255 which is the start of a unit test I added, we should be able to resolve symbolic refs with arbitrary data in them but we can't:

     1) Failure:
     refs::create::symbolic_with_arbitrary_content [/home/richardipsum/projects/contrib/libgit2/tests/refs/create.c:108]
     Function call failed: (git_reference_resolve(&resolved_ref, looked_up_ref))
     error -12 - The given reference name 'ARBITRARY DATA' is not valid

richardipsum:
@ethomson Hi, I'd really like to help you resolve this but the time I had to work on this change was extremely limited and I just wasn't able to get the tests done in concert with all the other projects I'm working on.

I have a bit more time now, if you can provide me with some direction as to how I can improve this line of changes then I'd really appreciate it.

richardipsum:
Meanwhile, I will look into what is necessary to make the tests I had for this PR pass.

include/git2/common.h
richardipsum: So, I'm afraid that the belated answer here would be no, in my opinion, because we need to be able to resolve refs that have been created without strict validation.

I'm not sure what this means for us presently, given that your fix up renames the constant and has now been merged.

richardipsum:
So the issue seems to be that `git_reference_lookup_resolved` calls `reference_normalize_for_repo` with validation always enabled. This is a problem if the ref we're trying to resolve is a symbolic ref with arbitrary data as a target. The issue with fixing this in the most obvious way is that `git_reference_lookup_resolved` is apart of the API, so I suppose the question is how to fix this without changing the existing API.

richardipsum:
I think the root of the issue might be that `git_reference__normalize_name` conflates unicode normalisation with validation. And we now have a case where we want to perform the unicode normalization but we don't want to perform validation on the ref.

Could we split the validation into its own function?

For the general case, if refs are validated on creation, do they also need to be validated when they're resolved?

ethomson:
I think that it _may_ make sense to split unicode normalization and validation.

However, I do want to point out that `git_reference_resolve` should _not_ be working in the unit test that you posted, if I'm reading it correctly.  `git_reference_resolve` unpeels symbolic references until a final direct reference is reached and returns the OID.

You have a reference `ANOTHER_HEAD_TRACKER` that has contents `ARBITRARY DATA`.  You don't have a direct reference `ARBITRARY DATA`.

You should not be able to `git_reference_resolve` either `ANOTHER_HEAD_TRACKER` or `ARBITRARY DATA`.

I don't think your unit test is failing because of the contents of your symbolic reference _target_.  It's failing because you're (implicitly) trying to use `ARBITRARY DATA` as a ref name, and it's not a valid one.  That seems to be a very different issue than the one that you're trying to solve?

richardipsum:
I may well be confused here, but the test I have added is essentially just a modified version of `test_refs_create__symbolic`, the only real difference as far as I understand this is that `test_refs_create__symbolic` makes a ref that has target `refs/heads/master` where my arbitrary content test replaces `refs/heads/master` with `ARBITRARY DATA`.

It is the case that when you `git_reference_resolve` a symbolic ref, `git_reference_lookup_resolved` is called and the target of the symbolic ref is given as the `name` argument. We do surely need to be able to resolve symbolic refs whether they contain arbitrary content or not?

If my current `ANOTHER_HEAD_TRACKER` points to arbitrary content, I need to be able to resolve `ANOTHER_HEAD_TRACKER` to obtain the arbitrary content.

ethomson:
You don't want to _resolve_ it, you want to know its target.  You should be able to call [`git_reference_symbolic_target`](https://github.com/libgit2/libgit2/blob/master/include/git2/refs.h#L261) on `ANOTHER_HEAD_TRACKER`, and it should return `ARBITRARY DATE`.

But you should _not_ be able to resolve this reference.  Resolving [walks a chain of references until it finds a direct reference](https://github.com/libgit2/libgit2/blob/master/include/git2/refs.h#L284).  That is to say, if you have:

`HEAD` -> `refs/heads/symbolic1` -> `refs/heads/symbolic2` -> `refs/heads/master` -> `38762cf7f55934b34d179ae6a4c80cadccbb7f0a`

then _resolving_ `HEAD` returns you the direct reference `refs/heads/master` so that you can look at its OID.

If you try to resolve a symbolic reference that points to something that isn't a valid reference, then it should fail as you never end up at a direct reference.

But you should be able to read the target.

richardipsum:
Oh I see now, so the test should just be that we can use `git_reference_symbolic_target` to obtain the target. Sorry I hadn't seen that function, I obviously mistakenly thought that `resolve` was the way to do this.

Thanks for explaining, I'll rework the test now.

PR #3918, opened 2016-08-29
ASan Support

tiennou:
Extracted from #3883, with CMake support thanks to arsenm/sanitizers-cmake.

Add `-DASAN=ON -DBUILD_CLAR=ON` to your favorite CMake invocation.

Notes:
- This only enable ASan on `libgit2_clar`, so the main library isn't instrumented.
- It blanket-enables all sanitizers (eg. `address`, `thread`, `memory` and `undefined`). Apple's clang only supports `address` atm, so I'm interested to see what happens on the Linux build :wink:
- If your compiler doesn't support some of the sanitizers, cmake will spew warnings, which may or may not be wanted.


include/git2/common.h
```diff
@@ -59,6 +59,12 @@ typedef size_t size_t;
 #define GIT_WIN32 1
 #endif
 
+#if defined(__has_feature)
+#if __has_feature(address_sanitizer)
+#define GIT_ASAN 1
```
ethomson: I'm not sure I understand what's going on here.  When is `__has_feature(address_sanitizer)` true?  It seems like it's true when `address_sanitizer` is available, not when libgit2 was built with it, no?

If your goal is to provide some functionality to callers about whether libgit2 was built with address_sanitizer, then please have cmake set `-DGIT_ASAN` and expose a function that will check and return that.  (Probably a query key on `git_libgit2_opts` makes sense.)

tiennou: The feature is enabled when AddressSanitizer is available and currently being used for the current compile run.

The CMake modules take care of automatically passing in the correct flags to the compiler when you `-DASAN=ON`, which will in turn cause that define to exist. The ultimate goal is to be able to tell at compile-time whether ASan is enabled or not (because the modules don't set anything since there's a clang feature macro for it), so we can skip the oom test below (e.g. doing weird things that makes ASan unhappy, like allocating the whole memory). That's the only point of it actually, so if having that in `common.h` feels wrong, I can move it somewhere else.

I don't really get the point about providing functionality. Because ASan incurs a [typical 2x slowdown](http://clang.llvm.org/docs/AddressSanitizer.html), it's not something I expect to be queried at runtime. It can be used as a debugging tool (build libgit2 with ASan, link your app with that version, run and wait).

CMakeLists.txt
```diff
@@ -578,6 +579,8 @@ ELSEIF (AMIGA)
 ELSE()
 	IF (VALGRIND)
 		ADD_DEFINITIONS(-DNO_MMAP)
+	ELSEIF (ASAN)
+		FIND_PACKAGE(Sanitizers)
```
ethomson: Can this move into the `IF (ASAN)` block down below?

include/git2/common.h
ethomson: Got it - yeah, I don't think that a public header is the best place for this.  I would suggest `src/common.h` if you do want to put this in a header file.

But my preference would be to put this in the `CMakeLists.txt` itself, which I think makes it a bit more discoverable.  (And there's some prior art there - the similar functionality of `DEBUG_POOL` and `MSVC_CRTDBG` just fiddles compiler flags.)

I would just put:

    ADD_DEFINITIONS(-D GIT_ASAN)

in the `IF (ASAN)` block.

tiennou:
Rebased (ping @lhchavez).

Caveats/notes :
- I've removed `-DASAN` in favor of the (module-provided) individual `SANITIZE_.*` flags, since there are compatibility issues between each.
- It should work on the library as well (though there's no selection possible). YMMV.
- Sanitizers shows up in the CMake feature list.
- It seems there's no way to know if a specific sanitizer could be successfully enabled. It might be possible to unset the "offending" variable, but it's in nested functions so scoping might be an issue. This can make both the feature list wrong, or impact pkgconfig generation.
- It can be submoduled. Dunno if we want that though.

(Speaking about pkgconfig generation, I've written a module to do this for the mbedTLS project)

CMakeLists.txt
```diff
@@ -252,6 +252,15 @@ ELSE()
 	# that uses CMAKE_CONFIGURATION_TYPES and not CMAKE_BUILD_TYPE
 ENDIF()
 
+ADD_FEATURE_INFO(sanitize_address   SANITIZE_ADDRESS   "ASan support")
```
lhchavez: maybe use `USE_SANITIZE_ADDRESS` for consistency with all the other options? and maybe add `ADD_OPTION` for them.

tests/CMakeLists.txt
```diff
@@ -48,6 +48,11 @@ ENDIF()
 TARGET_LINK_LIBRARIES(libgit2_clar ${LIBGIT2_LIBS})
 IDE_SPLIT_SOURCES(libgit2_clar)
 
+
```
lhchavez: extra newline?

CMakeLists.txt
tiennou: I'm piggy-backing on the variable names(s) used by the modules, so I don't think it's worthwhile to carry a local patch…

For the record, I'd think I'd prefer a `-DSANITIZE=address,stuff,stuff` to what's currently in the modules, for various reasons. The main one is that those variables are (obviously) cached, so if you want to change them, you'd have to remember to pass `-DSANITIZE.*=0` along, lest you run into the issue of sanitizer not being compatible with each other.

lhchavez: ah makes sense. that sounds like something that can be done in a follow-up change.

tiennou:
Rebased, I've added the `-DSANITIZERS=address,whatever` flag. Do we want that as part of CI/fuzzing ? I have no idea how fuzzing works though…

tiennou:
> But we (most likely I) can enable that later if you don't want to do it now.

I'll happily defer that to you, then 😉. Feel free to push those tweaks there or ping me to cherrypick something, in case merging takes longer.

pks-t:
The OSS project already enables sanitizers by itself, so there's nothing we need to change there. And I don't think it makes a lot of sense for our standalone fuzzers: they're mainly intended as a way to verify that our fuzzers build, all the real work is deferred to OSS Fuzz.

ethomson:
I'm going through some older PRs — I think that this is quite outdated, apologies. We have a lot of fuzzing in `main`, but I'd love to add more if there's something here that we're missing.

PR #3919, opened 2016-08-30
Delete 这

paddy2k:
Removed this file as it causes issues with a build system.
What is it's purpose?


carlosmn:
It's to test that libgit2 handles non-ASCII characters fine across filesystems. As you can see, removing this file makes all the tests for that fail.

ethomson:
Indeed, this is an important consideration for our tests.  I'm going to close this since there hasn't been additional activity, so I hope that you've found a workaround.  If not, please let us know and we might be able to make a suggestion.

PR #3921, opened 2016-09-01
Improve revision walk preparation logic

### LINKED ISSUE #3916: Rebase via libgit2 take a long time / forever
I've run into a case where a rebase via libgit2 / rugged takes a very, very long time.

Here's a reproducible test case:

```
git clone https://github.com/ollie314/gecko-dev
cd gecko-dev
git fetch mozilla https://github.com/mozilla/gecko-dev
```

then run this rugged script:

```ruby
require "rugged"

repo = Rugged::Repository.new(".")

head = repo.lookup("0dd403224a5acb0702bdbf7ff405067f5d29c239")
base = repo.lookup("b7083959a30f2137d8a6e27a8489f8729873950c")

rebase = Rugged::Rebase.new(repo, head, base, {
  inmemory: true,
  fail_on_conflict: true,
  skip_reuc: true,
  no_recursive: true
})

# Start from the upstream commit
final_commit_oid = base.oid
committer = { name: "Arthur", email: "test@example.com" }

trees = []
pos = 0
while step = rebase.next
  index = rebase.inmemory_index

  if (index.nil? || index.conflicts?)
    rebase.abort
    return nil
  end

  commit_oid = rebase.commit({ committer: committer })

  final_commit_oid = commit_oid if commit_oid
end

rebase.finish(committer)
```

The same rebase via git cli takes around a minute:

```
$ time git rebase 0dd403224a5acb0702bdbf7ff405067f5d29c239 b7083959a30f2137d8a6e27a8489f8729873950c
First, rewinding head to replay your work on top of it...

real	1m13.621s
user	1m4.894s
sys	0m9.977s
```

carlosmn:
This brings us closer to the code that's in git, makes it more efficient and introduces the slop mechanic in order to make it less likely a complex graph will trip us up.

This is solves the failing tests presented in #3838 in a much more elegant manner than the commits I pushed to that branch and resolves #3916. The gecho-dev walk in question now runs in 1.5s instead of longer than we care to measure.

Some of these tests now set the sorting since our unsorted iteration is now much less sorted than it used to be.

---

Chances are we're still doing something silly performance-wise like the way we deal with parents in `mark_uninteresting()` but this produces correct results and solves the immediate performance issue we're facing.


src/revwalk.c
```diff
@@ -398,81 +398,191 @@ static int revwalk_next_reverse(git_commit_list_node **object_out, git_revwalk *
 	return *object_out ? 0 : GIT_ITEROVER;
 }
 
-
-static int interesting(git_pqueue *list)
+static int contains(git_pqueue *list, git_commit_list_node *node)
```
pks-t: This looks like `git_vector_search`, which should in fact be more efficient. So maybe just add `#define git_pqueue_search git_pqueue_search` and remove this function altogether?

src/revwalk.c
```diff
@@ -398,81 +398,191 @@ static int revwalk_next_reverse(git_commit_list_node **object_out, git_revwalk *
...
-	}
+	for (i = 0; i < commit->out_degree; i++)
+		git_commit_list_insert(commit->parents[i], &parents);
 
-	return 0;
+
+	while (parents) {
+		git_commit_list_node *commit = git_commit_list_pop(&parents);
+
+		while (commit) {
```
pks-t: Is this second loop walking up the first-parent chain even necessary? We add the commit's parents to `parents` either way and would arrive at all parents either way. Additionally, I think we iterate over all first-parents twice due to this (even though the second time we bail out immediately due to `commit->uninteresting == 1`).

src/revwalk.c
```diff
@@ -398,81 +398,191 @@ static int revwalk_next_reverse(git_commit_list_node **object_out, git_revwalk *
...
+			git_commit_list_insert_by_date(p, list);
+		}
+
+		if (walk->first_parent)
+			break;
+	}
+	return 0;
+}
 
-			if (contains(&q, parent))
```
pks-t: Wow. Worst diff ever...

src/revwalk.c
```diff
@@ -398,81 +398,191 @@ static int revwalk_next_reverse(git_commit_list_node **object_out, git_revwalk *
...
+static int everybody_uninteresting(git_commit_list *orig)
+{
+	git_commit_list *list = orig;
+
+	while (list) {
+		git_commit_list_node *commit = list->item;
+		list = list->next;
+		if (commit->uninteresting)
+			continue;
```
pks-t: `if (!commit->uninteresting) return 1;` is maybe a tad easier to read here. Might be only me, though.

src/revwalk.c
```diff
@@ -398,81 +398,191 @@ static int revwalk_next_reverse(git_commit_list_node **object_out, git_revwalk *
...
+/* How many unintersting commits we want to look at after we run out of interesting ones */
+#define SLOP 5
+
+static int still_interesting(git_commit_list *list, int64_t time, int slop)
+{
+	/* The empty list is pretty boring */
+	if (!list)
+		return 0;
+
+	/* If destination list has commits with an earlier, deffo not done */
```
pks-t: I don't really get this comment :/

src/revwalk.c
```diff
@@ -398,81 +398,191 @@ static int revwalk_next_reverse(git_commit_list_node **object_out, git_revwalk *
...
+		return SLOP;
+
+	/* Everything's uninteresting, reduce the count */
+	return slop - 1;
+}
+
+static int limit_list(git_commit_list **out, git_revwalk *walk, git_commit_list *commits)
+{
+	int error, slop = SLOP;
+	int64_t time = ~0ll;
```
pks-t: Isn't this easier to understand when saying the equivalent `int64_t time = -1`?

pks-t:
Mostly minor nits, looks very nice otherwise :+1:

ethomson:
I'm seeing some results that differ from git, using the repo in #3916 :

```
% git rev-list 0dd403224a5acb0702bdbf7ff405067f5d29c239 ^b7083959a30f2137d8a6e27a8489f8729873950c --date-order |head -10             
0dd403224a5acb0702bdbf7ff405067f5d29c239
a2812fa126be538f73efed589e78d6973f23df2f
21ac721516934679f9d6528eba41364bbb7f6f5d
a2da90fae1c4b5fd0cd33ff1a509d8589f8ce695
7f0262e9054aac9f44ee307ea2d1b9a2f2993da3
f09e8fef1a803905ee29457e12cd68a04af256c4
44a196676794033e6dc0a66b890cdf55e7a3c999
247986c342e2cab0f95b35c8f841ac609aa0882d
d07b49ff57344a58c389d31d1c0235c469c215b0
f9242cf7754ab3f64b2f6650f40b24c7020ac61c
```

And in libgit2, using this simple program:
```
git_repository_open(&repo, "/Users/ethomson/Temp/gecko-dev");
git_revwalk_new(&revwalk, repo);
git_oid_fromstr(&head_id, "0dd403224a5acb0702bdbf7ff405067f5d29c239");
git_revwalk_push(revwalk, &head_id);
git_oid_fromstr(&base_id, "b7083959a30f2137d8a6e27a8489f8729873950c");
git_revwalk_push(revwalk, &base_id);
git_revwalk_sorting(revwalk, GIT_SORT_TIME);

while (git_revwalk_next(&id, revwalk) == 0) {
    char idstr[GIT_OID_HEXSZ];
    git_oid_fmt(idstr, &id);
    printf("%.*s\n", GIT_OID_HEXSZ, idstr);
}
```

We get:
```
% revwalk | head -10
0dd403224a5acb0702bdbf7ff405067f5d29c239
a2812fa126be538f73efed589e78d6973f23df2f
21ac721516934679f9d6528eba41364bbb7f6f5d
a2da90fae1c4b5fd0cd33ff1a509d8589f8ce695
7f0262e9054aac9f44ee307ea2d1b9a2f2993da3
f09e8fef1a803905ee29457e12cd68a04af256c4
44a196676794033e6dc0a66b890cdf55e7a3c999
247986c342e2cab0f95b35c8f841ac609aa0882d
e1f9b3132b193b95d8d70acd0bdc2edc0ac33046
529df92f7d1e13e0ca613af7548509c23d919644
```

src/revwalk.c
Scottydgregory: It looks great I like it...

src/revwalk.c
carlosmn: This is what the style in the git codebase is, but I agree it's a rather awkward way of expressing it.

src/revwalk.c
carlosmn: it's missing "an earlier timestamp than the source", we use the time to figure out whether we should continue looking through the list.

carlosmn:
I was testing this pair of commits with `rev-list --topo-order` which does produce the same result as `GIT_SORT_TIME | GIT_SORT_TOPOLOGICAL`.

Unfortunately it does seem that these combination isn't quite the right one, as `TIME | TOPO` is what `--date-order` describes, rather than `--topo-order` which is what we're getting.

carlosmn:
As an aside the snipped as-given does not have an equivalent git incantation since `--date-order` does also imply a topological sort, so the options we get with git would be the equivalent to `TOPO` or `TIME | TOPO`.

carlosmn:
I have ported more git code and now we do agree on `--topo-order` with `GIT_SORT_TOPOGRAPHICAL` and `--date-order` with `GIT_SORT_TIME | GIT_SORT_TOPOGRAPHICAL`.

With the exception of a single commit, which git shows but we don't. It's probably some edge condition I'm not taking into account, but we're almost there.

src/revwalk.c
carlosmn: We use the first parent as a guide to what to do next, going "deep" into the history first, but we'll go though the other parents as well as we've added them to the `parents` list.

src/revwalk.c
carlosmn: I've included a commit which adjusts the style to something we're expecting. It should be `return 0;` though, as we want to say that not everything is uninteresting.

src/revwalk.c
carlosmn: I ended up removing the whole block since we're not longer using this code.

carlosmn:
This should be good to go. We're not quite as fast as git, but fairly close. We're not as careful with memory allocations which is likely part of the reason.

But with this port of the code, we produce the same outputs for the `--date-order` and `--topo-order` equivalents.

tests/revwalk/basic.c
```diff
@@ -473,3 +473,51 @@ void test_revwalk_basic__big_timestamp(void)
...
+
+	revwalk_basic_setup_walk("testrepo.git");
+
+	cl_git_pass(git_oid_fromstr(&new_id, "bd758010071961f28336333bc41e9c64c9a64866"));
+	cl_git_pass(git_revwalk_push(_walk, &new_id));
+
+	cl_git_pass(git_oid_fromstr(&old_id, "8e73b769e97678d684b809b163bebdae2911720f"));
+	cl_git_pass(git_revwalk_hide(_walk, &old_id));
+
+   cl_git_pass(git_revwalk_next(&oid, _walk));
```
arthurschreiber: It looks like you used spaces here instead of tabs.

tests/revwalk/basic.c
```diff
@@ -473,3 +473,51 @@ void test_revwalk_basic__big_timestamp(void)
...
+
+	revwalk_basic_setup_walk("testrepo.git");
+
+	cl_git_pass(git_oid_fromstr(&new_id, "bd758010071961f28336333bc41e9c64c9a64866"));
+	cl_git_pass(git_revwalk_push(_walk, &new_id));
+
+	cl_git_pass(git_oid_fromstr(&old_id, "b91e763008b10db366442469339f90a2b8400d0a"));
+	cl_git_pass(git_revwalk_hide(_walk, &old_id));
+
+   cl_git_pass(git_revwalk_next(&oid, _walk));
```
arthurschreiber: Also spaces instead of tabs here.

carlosmn:
I've discovered that just passing in `REVERSE` will in fact not reverse things, but not giving it anything will... so I guess we'll have to fix that before merging.

ethomson:
Hmm.  With the test program above (showing and hiding the same commits) I'm still seeing differences.

`GIT_SORT_TIME | GIT_SORT_TOPOLOGICAL` (versus `--topo-order`) gives me the first several commits as being the same, but using only `GIT_SORT_TIME` (versus `--date-order`) gives me several differences even in the first few commits.

Worse, using either `GIT_SORT_TIME` or `GIT_SORT_TIME|GIT_SORT_TOPOLOGICAL`, we walk 489,028 commits while git walks 8,767.

carlosmn:
The equivalent options are:

* `--topo-order` is `GIT_SORT_TOPOLOGICAL`
* `--date-order` is `GIT_SORT_TOPOLOGICAL | GIT_SORT_TIME`
* `GIT_SORT_TIME` is something libgit2 has without any equivalency to anything anywhere at any time

I think you might have forgotten to change that `_push` int your sample program to a `_hide` since

```
% git rev-list --count 0dd403224a5acb0702bdbf7ff405067f5d29c239 b7083959a30f2137d8a6e27a8489f8729873950c
489028
```

so it definitely looks like you're just listing everything starting from those commits.

ethomson:
> I think you might have forgotten to change that _push int your sample program to a _hide ...

Oh, yes, duh, that's exactly my problem.  I suspected I was doing something dumb but didn't expect it was that dumb.  Alas.

Issue #3920, opened 2016-09-01
Diff via libgit2 on bare repo almost 10 times slower than git


arthurschreiber:
I believe that generating a diff via libgit2 is considerably slower than via git.

I've encountered this while using rugged, but I'm able to reproduce the "bad performance" I've seen in rugged using the libgit2 `diff` example.

Here's a "simple way" to reproduce this:

```
arthurschreiber$ git clone https://github.com/Seeed-Studio/Grove_LED_Matrix_Driver --bare

Cloning into bare repository 'Grove_LED_Matrix_Driver.git'...
remote: Counting objects: 540, done.
remote: Total 540 (delta 0), reused 0 (delta 0), pack-reused 540
Receiving objects: 100% (540/540), 5.37 MiB | 910.00 KiB/s, done.
Resolving deltas: 100% (264/264), done.
Checking connectivity... done.

arthurschreiber$ time libgit2/build/examples/diff --git-dir Grove_LED_Matrix_Driver.git 9bdc746c1d1999c9a073cd2839899b84af2ff0cd fdb7257222e63a9b241b3899ff426bcd387b3821 >/dev/null

real	0m1.551s
user	0m1.520s
sys	0m0.023s

arthurschreiber$ time git --git-dir Grove_LED_Matrix_Driver.git diff 9bdc746c1d1999c9a073cd2839899b84af2ff0cd fdb7257222e63a9b241b3899ff426bcd387b3821 >/dev/null

real	0m0.153s
user	0m0.125s
sys	0m0.018s

```

---

/cc @ethomson @brianmario

arthurschreiber:
One more thing I missed:

```
arthurschreiber$ git --version
git version 2.9.0
```

pks-t:
This is expected in this case, as in contrast to git.git we generate a diff for the binary file `Firmware/grove_led_matrix_mbed/build/MBED_A1.axf` and other binary files, which requires roughly 95% of the time generating the diff. When you enable binary diffs for git.git with `git diff --binary`, you get roughly the same timings.

arthurschreiber:
> This is expected in this case, as in contrast to git.git we generate a diff for the binary file Firmware/grove_led_matrix_mbed/build/MBED_A1.axf and other binary files, which requires roughly 95% of the time generating the diff

[...]

PR #3922, opened 2016-09-01
patch_generate: only calculate binary diffs if requested

pks-t:
When generating diffs for binary files, we load and decompress
the blobs in order to generate the actual diff, which can be very
costly. While we cannot avoid this for the case when we are
called with the `GIT_DIFF_SHOW_BINARY` flag, we do not have to
load the blobs in the case where this flag is not set, as the
caller is expected to have no interest in the actual content of
binary files.

Fix the issue by only generating a binary diff when the caller is
actually interested in the diff. As libgit2 uses heuristics to
determine that a blob contains binary data by inspecting its size
without loading from the ODB, this saves us quite some time when
diffing in a repository with binary files.

### DIFF
--- a/src/patch_generate.c
+++ b/src/patch_generate.c
@@ -349,20 +349,24 @@ static int diff_binary(git_patch_generated_output *output, git_patch_generated *
 		new_len = patch->nfile.map.len;
 	int error;
 
-	/* Create the old->new delta (as the "new" side of the patch),
-	 * and the new->old delta (as the "old" side)
-	 */
-	if ((error = create_binary(&binary.old_file.type,
-			(char **)&binary.old_file.data,
-			&binary.old_file.datalen,
-			&binary.old_file.inflatedlen,
-			new_data, new_len, old_data, old_len)) < 0 ||
-		(error = create_binary(&binary.new_file.type,
-			(char **)&binary.new_file.data,
-			&binary.new_file.datalen,
-			&binary.new_file.inflatedlen,
-			old_data, old_len, new_data, new_len)) < 0)
-		return error;
+	/* Only load contents if the user actually wants to diff
+	 * binary files. */
+	if (patch->base.diff_opts.flags & GIT_DIFF_SHOW_BINARY) {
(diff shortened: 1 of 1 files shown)


pks-t:
Refs #3920 

One thing I'd like to discuss is whether we should pass `NULL` to the `binary_cb` or, as I do in the patch, a `git_diff_binary` struct initialized to all-zero. I only took the second approach as passing `NULL` breaks some tests, so it was more of a lazy shortcut. I'm still open to changing this, though.

carlosmn:
We should probably be passing in the minimal information there. The comment for the type doesn't say what fields you can expect to have filled, so it looks like we should define what gets filled and then fill that. Presumably we'd want to pass in the path, object ids and mode so you can display the message that it changed if that's all you're after.

ethomson:
I think I would prefer to pass a `NULL` here if we didn't load it.  The path and object IDs and such are in the `delta`, and is suitable for a caller that just wanted to emulate `Binary files a/foo.txt and b/foo.txt differ`.

The `git_diff_binary` contains the old side and the new side of the binary data which is only useful for displaying the actual binary contents and emulating `git diff --binary`.  If we don't load that data then I think `NULL` would be very appropriate.

This is a good fix, I'm going to merge it and add `NULL` handling on top.

Thanks @pks-t !

ethomson:
After poking around with this a bit more, I realized that `NULL` probably isn't appropriate but that instead a `git_diff_delta` should indicate whether there's any data in its files...  (It can't look at the file sides since it may have been generated by parsing a patch that said `Binary files differ` or by running `git_diff_blob_to_blob` on two zero byte files, which are harder to discern.)

While looking at this I realized that we didn't have any way to parse a patch that had `Binary files differ` situation, so I'm fixing that up.

ethomson:
The aforementioned `Binary files differ` situation is a little more obnoxious than I had initially thought; there are some tests that assume behavior that I'm not sure makes sense.  I'm going to go ahead and merge this to unblock #3920 since this is correct.

PR #3923, opened 2016-09-02
Read binary patches (with no binary data)

ethomson:
When parsing a patch file, when we encounter a diff that was produced without `--binary`, we should include a binary patch that contains no data.  This type of patch node is special, and _cannot_ be applied, but it allows us to round-trip a patch file accurately.

`git_diff_binary` objects now indicate whether they contain data or not.  They "contain data" if they were produced by us when `SHOW_BINARY` is set or when they are parsed from a patch file produced by git with `--binary` set.  They do _not_ "contain data" when they are parsed from a patch file that simply says `Binary files ... differ`.

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -291,7 +291,15 @@ static int apply_binary(
 	git_patch *patch)
 {
 	git_buf reverse = GIT_BUF_INIT;
-	int error;
+	int error = 0;
+
+	if (!patch->binary.contains_data) {
+		error = apply_err("patch does not contain binary data");
+		goto done;
+	}
+
+	if (!patch->binary.old_file.datalen && !patch->binary.new_file.datalen)
+		goto done;
 
 	/* first, apply the new_file delta to the given source */
 	if ((error = apply_binary_delta(out, source, source_len,
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -74,8 +74,8 @@ static int parse_advance_expected(
 	return 0;
 }
 
-#define parse_advance_expected_s(ctx, str) \
-	parse_advance_expected(ctx, str, sizeof(str) - 1)
+#define parse_advance_expected_str(ctx, str) \
+	parse_advance_expected(ctx, str, strlen(str))
 
 static int parse_advance_ws(git_patch_parse_ctx *ctx)
 {
@@ -220,7 +220,7 @@ static int parse_header_git_index(
 {
 	if (parse_header_oid(&patch->base.delta->old_file.id,
 			&patch->base.delta->old_file.id_abbrev, ctx) < 0 ||
-		parse_advance_expected_s(ctx, "..") < 0 ||
+		parse_advance_expected_str(ctx, "..") < 0 ||
 		parse_header_oid(&patch->base.delta->new_file.id,
 			&patch->base.delta->new_file.id_abbrev, ctx) < 0)
 		return -1;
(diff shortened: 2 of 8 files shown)


src/patch_parse.c
```diff
@@ -74,7 +74,10 @@ static int parse_advance_expected(
 	return 0;
 }
 
-#define parse_advance_expected_s(ctx, str) \
+#define parse_advance_expected_str(ctx, str) \
+	parse_advance_expected(ctx, str, strlen(str))
+
+#define parse_advance_expected_const_str(ctx, str) \
```
pks-t: Is this distinction really required? To the best of my knowledge, most compilers convert `strlen("CONST_STRING")` to a compile-time constant anyway.

src/apply.c
```diff
@@ -291,7 +291,15 @@ static int apply_binary(
...
-	int error;
+	int error = 0;
+
+	if (!patch->binary.old_file.datalen && !patch->binary.new_file.datalen)
+		goto done;
+
+	if (!patch->binary.contains_data) {
+		error = apply_err("patch does not contain binary data");
+		goto done;
+	}
```
pks-t: Can this even be triggered? If `contains_data == 0`, we'll usually have `old_file.datalen == 0 && new_file.datalen == 0` due to the patch not being parsed. So actually, we'd already skip this check due to the previous check. I'd rather expect that this statement and the previous statement needs to be swapped.

src/patch_parse.c
```diff
@@ -804,6 +808,23 @@ static int parse_patch_binary(
...
+	git_patch_parse_ctx *ctx)
+{
+	if (parse_advance_expected_const_str(ctx, "Binary files ") < 0 ||
+		parse_advance_expected_str(ctx, patch->header_old_path) < 0 ||
+		parse_advance_expected_const_str(ctx, " and ") < 0 ||
+		parse_advance_expected_str(ctx, patch->header_new_path) < 0 ||
+		parse_advance_expected_const_str(ctx, " differ") < 0 ||
+		parse_advance_nl(ctx) < 0)
+		return parse_err("corrupt git binary header at line %d", ctx->line_num);
+
```
pks-t: Just wondering: can it happen that we get as input a `patch` with `patch->base.binary.contains_data == 1`? Probably not, but maybe we should still explicitly set `contains_data = 0` to make it more explicit.

ethomson: Only when we're actually given binary data, eg `GIT binary patch...`, which will branch to `parse_patch_binary_data`.  `patch->base.binary` gets the `memset(0)` treatment, so the default is `contains_data = 0`, but I will make this more explicit.

src/patch_parse.c
ethomson: Yeah, I think that's reasonable.  I think that any compiler that anybody is likely to use will convert `strlen("const")` to a constant and for the person using MSVC 6.0, this isn't likely to be the slow part of reading patches.

src/apply.c
ethomson: Yep, you're right.  I did flip these checks, which covered up the fact that I wasn't successfully setting `contains_data = 1` when we were creating a diff from unmodified binary files.  Fixed this up.

ethomson:
Thanks for the careful eyes, @pks-t !

pks-t:
Looks good to me now, thanks :+1:

PR #3929, opened 2016-09-13
time: Export `git_time_monotonic`

vmg:
This will allow libgit2 wrappers like Rugged to implement their own custom timeout logic while reusing the work we've already done in libgit2 to provide cross-platform, accurate monotonic timers.

See: https://github.com/libgit2/rugged/pull/631

cc @brianmario @mclark @ethomson @carlosmn


ethomson:
vmg: "`sys` is like this magical place where we make poor design choices."

😢

PR #3931, opened 2016-09-14
checkout: don't try to calculate oid for directories

### LINKED ISSUE #3911: libgit2_clar test fails on FreeBSD
Using a freshly cloned libgit2 repo at rev c60210d36780a39db669e52d6d592d339c6a2ed3 on FreeBSD 11.0-BETA1 amd64, the libgit2_clar test fails.

Test project /usr/home/somers/src/libgit2/build
    Start 1: libgit2_clar
1/4 Test #1: libgit2_clar .............................***Failed  649.44 sec
    Start 2: libgit2_clar-cred_callback
2/4 Test #2: libgit2_clar-cred_callback ...............   Passed    0.02 sec
    Start 3: libgit2_clar-proxy_credentials_in_url
3/4 Test #3: libgit2_clar-proxy_credentials_in_url ....   Passed    0.02 sec
    Start 4: libgit2_clar-proxy_credentials_request
4/4 Test #4: libgit2_clar-proxy_credentials_request ...   Passed    0.02 sec

75% tests passed, 1 tests failed out of 4

Total Test time (real) = 649.61 sec

The following tests FAILED:
	  1 - libgit2_clar (Failed)
Errors while running CTest

ethomson:
When trying to determine if we can safely overwrite an existing workdir item, we may need to calculate the oid for the workdir item to determine if its identical to the old side (and eligible for removal).

We previously did this regardless of the type of entry in the workdir; if it was a directory, we would open(2) it and then try to read(2).  The read(2) of a directory fails on many platforms, so we would treat it as if it were unmodified and continue to perform the checkout.

On FreeBSD, you _can_ read(2) a directory, so this pattern failed.  We would calculate an oid from the data read and determine that the directory was modified and would therefore generate a checkout conflict.

This reliance on read(2) is silly (and was most likely accidentally giving us the behavior we wanted), we should be explicit about the directory test.

Fixes #3911

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -212,6 +212,10 @@ static bool checkout_is_workdir_modified(
 	if (baseitem->size && wditem->file_size != baseitem->size)
 		return true;
 
+	/* if the workdir item is a directory, it cannot be a modified file */
+	if (S_ISDIR(wditem->mode))
+		return false;
+
 	if (git_diff__oid_for_entry(&oid, data->diff, wditem, wditem->mode, NULL) < 0)
 		return false;
 


src/checkout.c
```diff
@@ -212,6 +212,10 @@ static bool checkout_is_workdir_modified(
 	if (baseitem->size && wditem->file_size != baseitem->size)
 		return true;
 
+	/* if the workdir item is a directory, it cannot be a modified file */
+	if (S_ISDIR(wditem->mode))
+		return false;
```
pks-t: Shouldn't we consider this as modified if `!S_ISDIR(baseitem->mode)`? E.g. something like

```C
if (S_ISDIR(wditem->mode))
    return !S_ISDIR(baseitem->mode);
```

ethomson: That can't happen - we only process files in the checkout loop, not folders.  (We'll make intermediate folders if we need them.)  So we'll never see a `baseitem` that is a folder.

PR #3935, opened 2016-09-25
Support mbedTLS

joshtriplett:
Based on the mbedtls branch from https://github.com/wildart/libgit2 ,
with merge conflicts fixed, updates for subsequent libgit2 API changes,
and cleanups.

Passes all tests, including online tests.

Patches separated for review, but this does not bisect; squash-merge
recommended.


CMakeLists.txt
```diff
@@ -112,6 +112,10 @@ IF (HAVE_STRUCT_STAT_NSEC OR WIN32)
 	OPTION( USE_NSEC		"Care about sub-second file mtimes and ctimes"	ON  )
 ENDIF()
 
+IF (NOT USE_OPENSSL)
+	OPTION( USE_MBEDTLS     "Link with and use mbedTLS library"             ON )
```
tkelman: we should default this to off, otherwise on darwin it'll try to use both the apple crypto and mbedtls if found

CMakeLists.txt
```diff
@@ -283,6 +287,10 @@ ELSE ()
 		FIND_PACKAGE(OpenSSL)
 	ENDIF ()
 
+	IF (NOT AMIGA AND USE_MBEDTLS)
```
tkelman: is amiga relevant? wonder why openssl is gated on `NOT AMIGA` ...

tkelman:
I was about to do this since we're very likely to start carrying and using this patch in Julia. We'll effectively be forking libgit2 to maintain and rebase the patch going forward if it doesn't get upstreamed.

Anyway, maybe best to tab-indent the new code for style consistency with the rest of libgit2?

CMakeLists.txt
joshtriplett: I think it needs to default to on, but it needs to get wrapped in the same "not darwin".  Will add a patch to fix.

CMakeLists.txt
joshtriplett: Not sure.  @sba1 made that change in commit c57c4af327979ad4189b108eaf751c912c8dea01.

joshtriplett:
@tkelman Happy to do so as the last step before merging, but that'll completely break diffs with any of the previous code, so I think I should wait for review on this version first.

CMakeLists.txt
sba1: I don't think that it is relevant for the Amiga port right now.

However, the identification of OpenSSL is disabled on Amiga, because FIND_PACKAGE produced wrong results when cross compiling (it added wrong (i.e., native) include paths), at least, if I remember it correctly. Instead, it is assumed that OpenSSL is available on this platform in the (cross) compiler's standard include path and libgit2 is only statically linked.

In essence, special care for Amiga doesn't need to be done in new stuff. From time to time, I'll take care of it and submit pull requests.

CMakeLists.txt
tkelman: why should it default to on?

tkelman: it might be possible to use openssl or mbedtls on darwin instead of the apple crypto, but that might not be supported here

joshtriplett: It should default to on when not on Darwin and not using OpenSSL, to attempt to autodetect mbedTLS if it doesn't find OpenSSL.  If it defaulted to off, you'd have to *explicitly* enable mbedTLS even if you have mbedTLS installed and OpenSSL not installed.

tkelman: right that makes sense, but that's what should determine the default, not whether or not the option gets defined at all? though since the openssl option isn't defined at all for darwin, at least this would be consistent

joshtriplett: Right; consistency seems preferable here.  It might make sense for a future patch to go through CMakeLists.txt, define all options unconditionally, and change all the conditionals to just set the default values.  But I don't think this patch should make that unrelated change.

ethomson:
Interesting.  I'll take a look at this as soon as I have some free time.  Definitely hoping to unblock debian users.

src/mbedtls_stream.c
```diff
@@ -0,0 +1,474 @@
...
+#include <mbedtls/error.h>
+#include "mbedtls/net.h"
+#include "mbedtls/debug.h"
+#include "mbedtls/entropy.h"
+#include "mbedtls/ctr_drbg.h"
+#include "mbedtls/error.h"
+#include "mbedtls/certs.h"
+
+#ifndef OPENSSLDIR
+# define OPENSSLDIR              "/usr/lib/ssl"
```
tkelman: cross-linking to https://github.com/JuliaLang/julia/issues/18693, this isn't very uniform across distributions and we might want to come up with a better certificate search mechanism?

joshtriplett: How about taking the value for this from a cmake-time variable with `/usr/lib/ssl` as the default?

tkelman: I guess that can already be done in CFLAGS (however cmake spells its equivalent) with a `-DOPENSSLDIR` mostly-equivalently. Maybe this isn't really libgit2's problem to solve.

staticfloat:
The current state of this PR has an important caveat: if SSL certificate files cannot be found, [`git_libgit2_init()` fails](https://github.com/joshtriplett/libgit2/blob/81555f5de02d064921018647c4156257a8df748c/src/mbedtls_stream.c#L146-L155).  This severely limits the usefulness of the ability for users of `libgit2` to use the `git_libgit2_opts()` function to [modify the `SSL_CERT_LOCATIONS` setting](https://github.com/joshtriplett/libgit2/blob/81555f5de02d064921018647c4156257a8df748c/src/settings.c#L179-L205), as we cannot change options before initializing the library.

As it stands, if a user chooses to compile against MbedTLS and wants to create a portable binary (e.g. one that can run on both Ubuntu and CentOS, which store their certificates in very different locations) the software using `l]ibgit2` cannot, using the C API, set the locations for libgit2 to search for certificate files.  The user must set environment variables in order to change the search path.

I see two possible solutions:

* Somehow change the MbedTLS certificate loading to be lazier, allowing the user the chance to change the search path for certificates before they are actually needed.

* Somehow change the options setting so that global options like this can be set before `init()` actually occurs.

Personally, I think the first solution is the vastly superior one.  What do you guys think?

src/mbedtls_stream.c
```diff
@@ -0,0 +1,474 @@
...
+        mbedtls_ssl_config_free(git__ssl_conf);
+        git__free(git__ssl_conf);
+        git__ssl_conf = NULL;
+        return -1;
+    }
+
+    mbedtls_ssl_conf_authmode(git__ssl_conf, MBEDTLS_SSL_VERIFY_REQUIRED);
+    mbedtls_ssl_conf_rng(git__ssl_conf, mbedtls_ctr_drbg_random, ctr_drbg);
+
+    // find locations for which CA certificates
```
carlosmn: libgit2 is very much not in the business of guessing where SSL certificates live on everyone's machine. A distribution-provided mbedtls should know already. And if this is being shipped as part of a bundle, the application is the only one which would know what to use.

carlosmn:
I've gone through a few of the higher-level aspects of adding this TLS stream. I have to say, for a library with copy on their page about OpenSSL's bad API, it's own API is disappointing. Only in very specific situations is the user of the library in a better position than the library itself to know where to get a random number.

tkelman:
Thanks for the review @carlosmn.

@joshtriplett will need to include https://github.com/JuliaLang/julia/blob/e6a5cc38700302fe0986c55db8264f6c3673d1c3/deps/patches/libgit2-mbedtls-writer-fix.patch here

tkelman:
ping @joshtriplett any update here?

CDLuminate:
Any update? Julia 0.6 will be out soon however Debian still ships 0.4 and blocked by this problem.

tkelman:
Someone will have to own responding to the (much-appreciated) review feedback and updating the code to get it upstreamed. I'm fine for now with carrying the existing patch as-is in Julia, we've been using it this way for some time, but any updates can of course be incorporated in what Julia is using.

cmake/Modules/FindmbedTLS.cmake
```diff
@@ -0,0 +1,64 @@
+# - Try to find mbedTLS
+# Once done this will define
```
tiennou: ARMmbed/mbedtls#228, so not yet 😉

tiennou:
I've tried to tackle a few things here : https://github.com/tiennou/libgit2/tree/mbedtls

Mostly, I ended up in agreement of @carlosmn sentiment above. I can understand being "modular", thus forcing the user to provide whatever "entropy pool" it likes, but providing SSL cert bundles is too much. I think the way forward is to remove the "global" CA thing and hope that mbedTLS users will set `GIT_OPT_SET_SSL_CERT_LOCATIONS` correctly.

@joshtriplett Would you kindly pull that here ? I'd like to know what Travis thinks of the CI changes I made.

joshtriplett:
After working on this a while, I really don't have a lot of confidence in mbedTLS itself as an implementation, and I don't think I'm the right person to keep working on this.  @tiennou, please feel free to make a pull request with your version.

tkelman:
Will debian ever build its libgit2 package with https support then?

tiennou:
@joshtriplett 😆  🚌  💥 . FWIW, the sentiment is shared w.r.t. to confidence.

@tkelman Are there any other alternative to mbedTLS that would be Debian-compatible ?

I can't help but wonder how many wrappers were written, collectively 🙄.

tkelman:
What are you not confident about?Julia has switched to mbedtls for our entire network stack and it's working pretty well, certainly better than any other GPL-compatible alternative.

If Debian doesn't plan on enabling https support in their libgit2 packaging any time soon, I'm going to request that julia be removed from Debian rather than kept permanently stuck at an obsolete version.

ethomson:
Same with libgit2.  I don't like fielding requests from people who can't clone a repository because there's no underlying crypto stack.

joshtriplett:
I'd like to see this move forward; I just don't think it makes sense for me to be an intermediary between libgit2 and the people actually doing the work.  I'm not a crypto developer.  If there's someone actively working on this code, I think it makes more sense for them to own the pull request.  (I wish that github had a way to transfer ownership of a PR and point it at someone else's branch, rather than having to close it and open a new one.)

tkelman:
Granted mbedtls isn't the only way to accomplish that, debian could just flip a switch in the way they build libgit2 (afaiu most other distros do link theirs against openssl), but I guess there's a lingering legality question about gpl compatibility of other applications that link to libgit2.

@joshtriplett okay makes sense, thanks for working on it as much as you did. We don't currently have anyone working on this code - the original author vanished. I rebase it occasionally but don't understand the code much beyond that it seems to work and passes our tests.

joshtriplett:
@tkelman @tiennou If you're willing to rebase it and similar, please do consider upstreaming your changes; it doesn't seem like there are *that* many blockers here.

tiennou:
I share @joshtriplett non-crypto background. I was just trying to help get that code in, and I have no interest in that (apart from helping Debian).

@joshtriplett "Allow edits from maintainers" maybe ? I don't know if I'd qualify for that though, but it's worth a shot. If that doesn't work, I'll open a new one.

About confidence : I was looking at @carlosmn comment on the "missing" TLSv1 setup. There's a [function](https://tls.mbed.org/api/ssl_8h.html#a0eade5c83cc08001672061c5925caaaa) for that. Now, what does those constants mean ? I have no idea, and I found comments somewhere where the mapping from those "version numbers" to actual TLS version numbers was apparent. I do realize it's pretty minor, but that's how confidence works 😜.

@tkelman I'll happily try to get it in shape. It's just that I'm developing from a non-Debian OS things like certificate store paths (which macOS has no file-based equivalent of), and have to setup Travis.

tkelman:
we could grant each other collaborator access to our forks and keep it in this pr, or to collaborate on a new one

joshtriplett:
@tiennou 
> Are there any other alternative to mbedTLS that would be Debian-compatible ?

OpenSSL and the various libraries based on it, soon: https://www.openssl.org/blog/blog/2017/03/20/license/

Issue #3784, opened 2016-05-12
Remove unnecessary `clone_checkout_strategy in `git_submodule_update_options`


ethomson:
I think that [`clone_checkout_strategy`](https://github.com/libgit2/libgit2/blob/master/include/git2/submodule.h#L156) is no longer necessary, since we have removed the `GIT_CHECKOUT_SAFE_CREATE` flag and folded its special case behavior into `GIT_CHECKOUT_SAFE`.  (This was done after https://github.com/libgit2/libgit2/pull/2632 was merged, but I missed it when I was fixing things as part of #2908).

It would be nice to get rid of this, and make `clone` use the given checkout options (like checkout itself does).

ethomson:
Fixed via #3938

PR #3938, opened 2016-09-26
Remove `clone_checkout_strategy` in submodule update opts

joshleeb:
Remove `clone_checkout_strategy` in `git_submodule_update_options` as
per issue #3784.


include/git2/submodule.h
```diff
@@ -166,7 +157,7 @@ typedef struct git_submodule_update_options {
 #define GIT_SUBMODULE_UPDATE_OPTIONS_INIT \
 	{ GIT_SUBMODULE_UPDATE_OPTIONS_VERSION, \
 		{ GIT_CHECKOUT_OPTIONS_VERSION, GIT_CHECKOUT_SAFE }, \
-	GIT_FETCH_OPTIONS_INIT, GIT_CHECKOUT_SAFE, 1 }
+	GIT_FETCH_OPTIONS_INIT, GIT_CHECKOUT_SAFE }
```
ethomson: This doesn't seem quite right?  If I'm reading this right, it looks like you've dropped the last initializer (`allow_fetch`) instead of the second to last (`clone_checkout_strategy`).  Instead shouldn't you remove `GIT_CHECKOUT_SAFE` and leave the `1`?

CHANGELOG.md
```diff
@@ -59,6 +59,10 @@ v0.24 + 1
 * `git_blob_create_fromchunks()` has been removed in favour of
   `git_blob_create_fromstream()`.
 
+* `clone_checkout_strategy` has been removed from
+  `git_submodule_update_option` as it is no longer necessary after
+  removing the `GIT_CHECKOUT_SAFE_CREATE` flag in favour of the 'GIT_CHECKOUT_SAFE' flag.
```
ethomson: I think that this could be worded in a way that suggests why it is no longer necessary or what the end-result is.  What do you think about:

>  `clone_checkout_strategy` has been removed from `git_submodule_update_option`.  The checkout strategy used to clone will be the same strategy specified in `checkout_opts`.

joshleeb: :+1:

include/git2/submodule.h
joshleeb: Yep you're right. I'll update the diff to reflect this.

ethomson:
Sorry for taking so long to merge this @joshleeb - thanks for the fix.  I merged this manually so that I could move the changelog changes to the next version.

PR #3939, opened 2016-09-26
MinGW m32 vsnprintf compatibility

shibukawa:
https://ffmpeg.org/pipermail/ffmpeg-cvslog/2012-September/054932.html

MinGW m32 environment has broken vsnprintf. It needs same work around with MSVC.

### DIFF
--- a/src/win32/posix_w32.c
+++ b/src/win32/posix_w32.c
@@ -524,7 +524,7 @@ char *p_realpath(const char *orig_path, char *buffer)
 
 int p_vsnprintf(char *buffer, size_t count, const char *format, va_list argptr)
 {
-#if defined(_MSC_VER)
+#if defined(_MSC_VER) || !_WIN64
 	int len;
 
 	if (count == 0)


shibukawa:
I got this issue when reading .gitconfig. Joining section ``"filter.gitlfs"`` with any variable names(like ``required``, ``smudge``) got error.

shibukawa:
I use MinGW 5.3 bundled with Qt 5.7.

tkelman:
Why isn't qt using the 32 bit version of mingw-w64? ref https://github.com/libgit2/libgit2/issues/3662

shibukawa:
It was discussed here: https://blog.qt.io/blog/2016/03/16/qt-5-6-released/#comment-1197363

ethomson:
Can you narrow this test?  It seems like this will substitute the sort of ugly compatibility layer for any non-64 bit non-MSVC architecture, is that correct?  For example, it seems like this will apply to 32 bit mingw-w64, which I think does have `vsnprintf`?  Can you provide a test that applies to _only_ the deficient architecture?

ethomson:
I'm happy to support MinGW m32, but this test is overly broad.  Please open a new PR if you want to address it.

PR #3944, opened 2016-09-29
Implement a diff indent heuristic

mhagger:
Implement a heuristic from the Git project that, on average, significantly improves diffs. See the commit message for 19f1a8e6f289b07389d525a12a13a4aaeaabe443 for more detail than you probably ever wanted. Briefly, this patch causes `diff` to take into account the presence of blank lines and the relative indentation of other lines in the neighborhood of the diff to decide how to position ambiguous diff hunks. The result is that it very frequently improves diffs that look like

```
--- a/9c572b21dd090a1e5c5bb397053bf8043ffe7fb4:git-send-email.perl
+++ b/6dcfa306f2b67b733a7eb2d7ded1bc9987809edb:git-send-email.perl
@@ -231,6 +231,9 @@ if (!defined $initial_reply_to && $prompting) {
 }

 if (!$smtp_server) {
+       $smtp_server = $repo->config('sendemail.smtpserver');
+}
+if (!$smtp_server) {
        foreach (qw( /usr/sbin/sendmail /usr/lib/sendmail )) {
                if (-x $_) {
                        $smtp_server = $_;
```

into equivalent but more aesthetically pleasing diffs like

```
--- a/9c572b21dd090a1e5c5bb397053bf8043ffe7fb4:git-send-email.perl
+++ b/6dcfa306f2b67b733a7eb2d7ded1bc9987809edb:git-send-email.perl
@@ -230,6 +230,9 @@ if (!defined $initial_reply_to && $prompting) {
        $initial_reply_to =~ s/(^\s+|\s+$)//g;
 }

+if (!$smtp_server) {
+       $smtp_server = $repo->config('sendemail.smtpserver');
+}
 if (!$smtp_server) {
        foreach (qw( /usr/sbin/sendmail /usr/lib/sendmail )) {
                if (-x $_) {
```

The commit message that I referenced above explains how the heuristic was trained and documents the extent to which it improves on the existing diff algorithm.

This change is in Git's master branch. It currently can be enabled in Git by passing the `--indent-heuristic` option to `git diff` and similar commands. I expect that this change will be in the next Git release, though it is not yet clear how it will be enabled; indeed, it might become the default.

This patch series includes the parts of the relevant Git patches that apply to `src/xdiff/xdiff.h` and `src/xdiff/xdiffi.c`, but omitting the changes related to the "compaction heuristic" that was experimented with in the Git project but is likely to be discarded.

I haven't added any tests whatsoever, nor anything more than an xdiff flag that can be used to turn on this mode. Since I don't have a lot more time to work on this, and since I'm not familiar with the internals or test infrastructure of this project, I'm hoping that a more regular libgit2 developer can take over from here :four_leaf_clover:


mhagger:
I just added another commit from Git, which renames a struct in the new code from `group` to `xdlgroup` to avoid conflicts with a `struct group` from POSIX `grp.h`. This commit hasn't been merged to Git's `master` branch yet, but I think it (or something very like it) will be.

mhagger:
I see CI is failing. Are these cases where Visual Studio behavior is different than gcc?

If `inline` is not supported by VS, that's perhaps a problem, because the two functions marked as `inline` are invoked in rather tight loops, and removing the `inline` designation results in a measurable performance hit under gcc. Does VS do equivalent optimization automatically? If not, is there an equivalent that could be used under VS? (I suppose these could be turned into macros if unavoidable.)

tiennou:
There's a `GIT_INLINE` in src/common.h that handles inlining for VS, but since xdiff.h is considered "external" to the project, it's not included in here.

carlosmn:
Indeed. Unfortunately inlining functions is compiler-specific. Each of the codebases we import (zlib, regex, klib) have their own macros to deal with it. I'm surprised this hasn't been reported as an issue upstream yet, but xdiff is going to have to get its own macros for this.

mhagger:
Sorry, I accidentally deleted the source branch, which closed the PR. I just restored it and reopened the PR.

mhagger:
I just took care of the `inline` problem.

BTW, starting with Git 2.14, the indent heuristic is now the default behavior for upstream Git.

carlosmn:
With the merge from master, the tests are now passing.

Since this is already the behaviour in git, I'm going to go ahead and merge it.

@mhagger IIUC this enables us to choose this heuristic for xdiff, we still need to ask for it?

Issue #3864, opened 2016-07-18
Release v0.24.2


carlosmn:
- [x] Backport fixes from master
- [x] Check for leaks
- [x] Update version defines and string
- [x] Create changelog/release notes

tkelman:
I'd like to nominate https://github.com/libgit2/libgit2/pull/3760 for backporting to 0.24.2, we hit that issue in Julia causing a freeze on an https clone.

hanwen:
I'd like to nominate:

* https://github.com/libgit2/libgit2/pull/3877
* https://github.com/libgit2/libgit2/pull/3879

carlosmn:
The changelog would be (removing administrivia commits):

```
adfece0 array: fix search for empty arrays
5a9d850 odb: only provide the empty tree
ed5299a odb: actually insert the empty blob in tests
886bd6a mwindow: init mwindow files in git_libgit2_init
edf420f cmake: add curl library path
e499b13 git_checkout_tree options fix
12b73ff transports: http: reset `connected` flag when re-connecting transport
0dea429 ignore: allow unignoring basenames in subdirectories
8f342c6 stransport: do not use `git_stream_free` on uninitialized stransport
9aee7bc stransport: make internal functions static
b64722f SecureTransport: handle NULL trust on success
1fafead sysdir: use the standard `init` pattern
85adddd refspec: do not set empty rhs for fetch refspecs
d711165 repository: don't cast to `int` for no reason
9894c7d remote: Handle missing config values when deleting a remote
49188d2 blame: do not decrement commit refcount in make_origin
1edbfa1 Fixed bug while parsing INT64_MIN
a200dc9 Fix repository discovery with ceiling_dirs at current directory
c7a0336 cmake: do not use -fPIC for MSYS2
cc43d18 README: update "Getting Help" section
ce124fc README: improve contributing paragraph
90ec160 README: disambiguate what to distribute source of
bb582f0 threads: add platform-independent thread initialization function
5d03db8 win32: rename pthread.{c,h} to thread.{c,h}
961bdbd threads: remove now-useless typedefs
2aa5c6f threads: remove unused function pthread_num_processors_np
68343f2 threads: split up OS-dependent rwlock code
fabd477 threads: split up OS-dependent thread-condition code
1b82531 threads: remove unused function pthread_cond_broadcast
40b243b threads: split up OS-dependent mutex code
fc2b97d threads: split up OS-dependent thread code
[...]

PR #3948, opened 2016-10-01
Backport fixes to v0.24

carlosmn:
These are the fixes that should make up v0.24.2, backported from master. They're a combination of bug fixes as well as memory safety enhancements.

/cc #3864


carlosmn:
Unfortunately I'm not quite sure whether the OSX failures are due to these changes or just the fact that the new one got released so now we're on a old version and homebrew won't link.

ethomson:
There were some changes in master for homebrew that I'll cherry-pick.

Thanks for doing this!

PR #3950, opened 2016-10-02
cmake: use a library target to avoid recompilation

ethomson:
Instead of listing the sources separately for the git2 target and the libgit2_clar target, create an ["object library" target](https://cmake.org/Wiki/CMake/Tutorials/Object_Library).

The object library will then be used to link both the git2 library target and the libgit2_clar target instead of having to compile all the source twice.  The object library is compiled with the same flags as the prior git2 library target was, so it will correctly use the shared library flags (or the static library flags, if we're building a static libgit2) so the end results should be unchanged.


carlosmn:
The `OBJECT` target was added in 2.8.8. Unfortunately in Ubuntu 12.04 the version of CMake is 2.8.7, so it won't kow what to do with it unless we bump the version ourselves. But then we don't support that LTS version out of the box.

tkelman:
It's relatively easy to get standalone binaries of new cmake versions, if this is worth requiring that.

ethomson:
Hmm, interesting.

The approach that I set out to take was simply to make a static library in the same manner and then create a shared library and the tests from that.  This seemed a little less kludgey once I found it.

Maybe I'll give that a whirl and see how effective it actually is.

ethomson:
Closed, via #4513

PR #3953, opened 2016-10-05
Make sure we use the `C` locale for `regcomp` on macOS.

arthurschreiber:
On macOS (and probably on other systems as well), `regcomp` behaves differently based on the current `locale`.

On macOS, this was causing some issues when enabling diff drivers: The `[\xc0-\xff][\x80-\xbf]` part of the regular expression defined at https://github.com/libgit2/libgit2/blob/master/src/userdiff.h#L23 was incorrectly parsed as multibyte if the locale character type was set to `UTF-8` (the default on macOS), and made `regcomp` fail with a `REG_ILLSEQ` error.

Fortunately, macOS ships with `regcomp_l`, which allows specifying a fixed locale instead of using whatever locale was set on the system or via environment variables.

/cc @carlosmn @ethomson


CMakeLists.txt
```diff
@@ -507,6 +507,11 @@ ELSE ()
 	ENDIF ()
 ENDIF()
 
+CHECK_FUNCTION_EXISTS(regcomp_l HAVE_REGCOMP_L)
```
arthurschreiber: Should this use `CHECK_SYMBOL_EXISTS(regcomp_l "xlocale.h" HAVE_REGCOMP_L)` instead?

src/unix/posix.h
```diff
@@ -80,4 +80,14 @@ GIT_INLINE(int) p_futimes(int f, const struct p_timeval t[2])
 # define p_futimes futimes
 #endif
 
+#ifdef HAVE_REGCOMP_L
+#include <xlocale.h>
+GIT_INLINE(int) p_regcomp(regex_t *preg, const char *pattern, int cflags)
+{
+	return regcomp_l(preg, pattern, cflags, (locale_t) 0);
```
carlosmn: If `(locale_t) 0` defined somewhere to be special? I've found `LC_GLOBAL_LOCALE` in the headers which is defined as `(locale_t) -1` but not for `0`.

CMakeLists.txt
carlosmn: Yes, that's what it looks like from its documentation. I wonder if `CHECK_FUNCTION_EXISTS` might work here as well, since it's system-provided.

carlosmn:
It'd be great to have a test here so we don't break this in the future. It should also let us know whether this kind of thing would fail in systems we're not aware of yet.

src/unix/posix.h
arthurschreiber: Unfortunately, I can't find anything either.

Initially, I created a new locale through `newlocale(LC_ALL_MASK, "C", (locale_t) 0);`, but I tested using `(locale_t) 0` directly and that seems to work as well.

We could change it back to using `newlocale` and storing that locale somewhere if you think that'd be safer?

arthurschreiber:
> It'd be great to have a test here so we don't break this in the future. It should also let us know whether this kind of thing would fail in systems we're not aware of yet.

What `git` does for testing the userdiff patterns is just looping over every pattern and trying to `regcomp` each one. Would that be good enough as a test?

carlosmn:
Yes, that should be enough. The main thing would be to test out  few patterns which we know should fail if we don't do this fix, to make sure this does do something.

src/unix/posix.h
arthurschreiber: From Apple's macOS source code release:

```c
int
tre_regncomp_l(regex_t *preg, const char *regex, size_t n, int cflags, locale_t loc)
{
  ...

  NORMALIZE_LOCALE(loc);

  ...
}
```

```c
#define NORMALIZE_LOCALE(x)	if ((x) == NULL) { \
	(x) = _c_locale; \
} else if ((x) == LC_GLOBAL_LOCALE) { \
	(x) = &__global_locale; \
}
```

So we should be fine passing `(locale_t) 0` (or even just `NULL`) here. Do you want me to change anything here?

arthurschreiber:
I added some tests, I hope they cover the basic functionality here.

carlosmn:
Those tests should be enough to notice things going awry.

PR #3954, opened 2016-10-05
[WIP] Travis Mac SSH key issues

carlosmn:
We're failing this check and I'm trying to figure out why. It might just be that the format from `ssh-keygen` changed so we need to account for that in our tests.


carlosmn:
It looks like the newer ssh or ssh-keygen now lists the SHA256, base64-encoded by default. Unfortunately libssh2 only supports MD5 and SHA1 for the fingerprint so that's what we're going to have to ask for on Mac, until the ssh version on Ubuntu caches up.

PR #3955, opened 2016-10-07
Fix the existence check for `regcomp_l`.

arthurschreiber:
`xlocale.h` only defines `regcomp_l` if `regex.h` was included as well.

@carlosmn I have NO idea why the unit tests were passing - looks like they are not really testing this?! 😕


arthurschreiber:
@carlosmn I fixed the tests to actually test with a multibyte locale as well.

carlosmn:
It's really annoying that we need code like this, but yeah let's merge it.

PR #3957, opened 2016-10-07
Perf: Don't perform merge operations for trivial rebases.

arthurschreiber:
When we rebase a commit on top of one of it's own ancestors, we can skip performing actual merges if the head tree and the parent tree are equal.

This can considerably speed up trivial rebases.

---

@ethomson I'm not really sure how to add tests for this. Should I just add a test that performs such a trivial rebase?

### DIFF
--- a/src/merge.c
+++ b/src/merge.c
@@ -2018,6 +2018,26 @@ int git_merge_trees(
 	git_iterator_options iter_opts = GIT_ITERATOR_OPTIONS_INIT;
 	int error;
 
+	assert(out && repo);
+
+	/* if one side is treesame to the ancestor, take the other side */
+	if (ancestor_tree && merge_opts && (merge_opts->flags & GIT_MERGE_SKIP_REUC)) {
+		const git_tree *result = NULL;
+		const git_oid *ancestor_tree_id = git_tree_id(ancestor_tree);
+
+		if (our_tree && !git_oid_cmp(ancestor_tree_id, git_tree_id(our_tree)))
+			result = their_tree;
+		else if (their_tree && !git_oid_cmp(ancestor_tree_id, git_tree_id(their_tree)))
+			result = our_tree;
+
+		if (result) {
+			if ((error = git_index_new(out)) == 0)
+			error = git_index_read_tree(*out, result);
+
+			return error;
+		}
+	}
+
 	iter_opts.flags = GIT_ITERATOR_DONT_IGNORE_CASE;
 
 	if ((error = git_iterator_for_tree(
--- a/src/rebase.c
+++ b/src/rebase.c
@@ -873,16 +873,27 @@ static int rebase_next_inmemory(
 			goto done;
 	}
 
-	if ((error = git_commit_tree(&head_tree, rebase->last_commit)) < 0 ||
-		(error = git_merge_trees(&index, rebase->repo, parent_tree, head_tree, current_tree, &rebase->options.merge_options)) < 0)
+	if ((error = git_commit_tree(&head_tree, rebase->last_commit)) < 0)
 		goto done;
 
-	if (!rebase->index) {
-		rebase->index = index;
-		index = NULL;
-	} else {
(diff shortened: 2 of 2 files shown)


arthurschreiber:
I was thinking - can't we even simply push this optimization into `git_merge_trees` as well?

ethomson:
Yeah - I do actually think we could move this into `git_merge_trees`.  (We could probably even move this into `git_merge_iterators` if we wanted to check the type of an iterator, grab its tree if it's a tree iterator, then when the ancestor and a side are both trees, see if they're the same tree.  But the only consumer of `git_merge_iterators` with two trees is stash, and it [already checks to see if the trees are the same or not](https://github.com/libgit2/libgit2/blob/master/src/stash.c#L878), so it wouldn't benefit from this.)

The rub here is that `merge_trees` can't just call `git_index_read_tree` and be done in the general case, because git produces REUC entries in [trivial three-way merge case 8](https://github.com/git/git/blob/master/Documentation/technical/trivial-merge.txt#L87) and [trivial three-way merge case 10](https://github.com/git/git/blob/master/Documentation/technical/trivial-merge.txt#L89).  That is to say when one side removes a file and the other side makes no changes to it, that should be recorded in the REUC.

I don't know the rationale here, and you could argue that we don't need to be strictly compatible with REUC entry creation, and you're probably right.  But I would prefer to be conservative about changing this, even if it's probably smart and safe.

So we can do this in `git_merge_trees` when `GIT_MERGE_SKIP_REUC` is set, without impacting compatibility.  I think this will be helpful for your uses, since you should never want a REUC.

ethomson:
I'm thinking something like:

```diff
diff --git a/src/merge.c b/src/merge.c
index 6934aa7..fd4510d 100644
--- a/src/merge.c
+++ b/src/merge.c
@@ -2020,6 +2020,27 @@ int git_merge_trees(
 
 	iter_opts.flags = GIT_ITERATOR_DONT_IGNORE_CASE;
 
+	assert(out && repo);
+
+	/* if one side is treesame to the ancestor, take the other side */
+	if (ancestor_tree && merge_opts && (merge_opts->flags & GIT_MERGE_SKIP_REUC)) {
+		const git_tree *result = NULL;
+
+		if (our_tree &&
+			!git_oid_cmp(git_tree_id(ancestor_tree), git_tree_id(our_tree)))
+			result = their_tree;
+		else if (their_tree &&
+			!git_oid_cmp(git_tree_id(ancestor_tree), git_tree_id(their_tree)))
+			result = our_tree;
+
+		if (result) {
+			if ((error = git_index_new(out)) == 0)
+				error = git_index_read_tree(*out, result);
+
+			return error;
+		}
+	}
+
 	if ((error = git_iterator_for_tree(
 			&ancestor_iter, (git_tree *)ancestor_tree, &iter_opts)) < 0 ||
 		(error = git_iterator_for_tree(
```

Also, I forgot to ask:  _does it_ speed things up?  :D

ethomson:
@arthurschreiber This looks pretty good to me, are you happy with it?  Does this provide enough benefits that you want to merge it?  Do you want to keep hacking?  :D

src/merge.c
```diff
@@ -2018,6 +2018,26 @@ int git_merge_trees(
...
+		const git_oid *ancestor_tree_id = git_tree_id(ancestor_tree);
+
+		if (our_tree && !git_oid_cmp(ancestor_tree_id, git_tree_id(our_tree)))
+			result = their_tree;
+		else if (their_tree && !git_oid_cmp(ancestor_tree_id, git_tree_id(their_tree)))
+			result = our_tree;
+
+		if (result) {
+			if ((error = git_index_new(out)) == 0)
+			error = git_index_read_tree(*out, result);
```
ethomson: Indentation is off here.

ethomson:
Am I wrong in thinking that `3e0c422` is superfluous and can be 🔥 d?

ethomson:
Manually merged; thanks!

Issue #3959, opened 2016-10-11
openssl 1.1.0 support


ignatenkobrain:
Unfortunately currently libgit2 is fails to build from source with OpenSSL 1.1.0 (https://bugzilla.redhat.com/show_bug.cgi?id=1383753):

``` c
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c: In function 'bio_create':
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c:159:3: error: dereferencing pointer to incomplete type 'BIO {aka struct bio_st}'
  b->init = 1;
   ^~
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c: At top level:
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c:217:1: error: variable 'git_stream_bio_method' has initializer but incomplete type
 static BIO_METHOD git_stream_bio_method = {
 ^~~~~~
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c:219:2: warning: excess elements in struct initializer
  "git_stream",
  ^~~~~~~~~~~~
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c:219:2: note: (near initialization for 'git_stream_bio_method')
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c:220:2: warning: excess elements in struct initializer
  bio_write,
  ^~~~~~~~~
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c:220:2: note: (near initialization for 'git_stream_bio_method')
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c:221:2: warning: excess elements in struct initializer
  bio_read,
  ^~~~~~~~
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c:221:2: note: (near initialization for 'git_stream_bio_method')
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c:222:2: warning: excess elements in struct initializer
  bio_puts,
  ^~~~~~~~
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c:222:2: note: (near initialization for 'git_stream_bio_method')
/builddir/build/BUILD/libgit2-0.24.2/src/openssl_stream.c:223:2: warning: excess elements in struct initializer
  bio_gets,
  ^~~~~~~~
[...]

PR #3960, opened 2016-10-12
add support for OpenSSL 1.1.0 for BIO filter

ignatenkobrain:
Closes: https://github.com/libgit2/libgit2/issues/3959
Signed-off-by: Igor Gnatenko i.gnatenko.brain@gmail.com

### DIFF
--- a/src/openssl_stream.c
+++ b/src/openssl_stream.c
@@ -357,7 +357,7 @@ static int verify_server_cert(SSL *ssl, const char *host)
 		num = sk_GENERAL_NAME_num(alts);
 		for (i = 0; i < num && matched != 1; i++) {
 			const GENERAL_NAME *gn = sk_GENERAL_NAME_value(alts, i);
-			const char *name = (char *) ASN1_STRING_data(gn->d.ia5);
+			const char *name = (char *) ASN1_STRING_get0_data(gn->d.ia5);
 			size_t namelen = (size_t) ASN1_STRING_length(gn->d.ia5);
 
 			/* Skip any names of a type we're not looking for */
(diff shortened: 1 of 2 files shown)


ignatenkobrain:
P.S. I didn't test it yet

carlosmn:
I would rather have us `#define` these `BIO_` setters and getters if we detect that OpenSSL doesn't provide them so we don't have our code littered with preprocessor statements.

mikeando:
No need to use a `#define`, real functions will work just fine. Just add a compatibility header that only defines the functions if SSH isn't providing them.

~~~
#if !OPENSSL_HAS_BIO_SET_INIT
void BIO_set_init(BIO* b, int v) {
 	b->init = v;
 	b->num = 0;
 	b->ptr = NULL;
 	b->flags = 0;
}
#endif
~~~

etc.

carlosmn:
Declaring functions in someone else's namespace is something you have to be very careful about, or you might end up overwriting the one hat libssl itself provides, depending on exactly how the libraries are loaded, like in the example above.

ignatenkobrain:
so what should I do after all?

carlosmn:
Either would work, but if you prefer functions, they should be using the `GIT_INLINE` macro so they're local and the compiler is more likely to inline these single-use tiny functions.

carlosmn:
My distribution just upgraded to 1.1 in the headers, so I've gone ahead and updated to what the kind of interfaces we program against.

I also moved the creation of the method to the global initialisation function. Creating a new one on each connection is unsafe thread-wise and it's also a memory leak.

carlosmn:
Thanks for doing the 1.1 conversion.

PR #3962, opened 2016-10-13
Introduce a GitHub Issue Template

ethomson:
Encourage people to ask questions on StackOverflow and use Issues for bug reports.

cc @carlosmn

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -1,0 +1,14 @@
+You are opening a _bug report_ against the libgit2 project.  If you have a
+question about an API or usage, please ask on StackOverflow:
+http://stackoverflow.com/questions/tagged/libgit2.  Please fill out the
+reproduction steps (below) and delete this introductory paragraph.  Thanks!
+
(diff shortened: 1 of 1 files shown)


pks-t:
I'd like to have the user specify on which version he tried whatever he did, as not everybody is always using the latest release. I'd also put the reproduction steps first and have the actual/expected results afterwards, as this is a more natural flow.

carlosmn:
+1 on asking the version. We might also want to add something about not opening bugs because some program is using libgit2 and it seems to be missing a feature. Unless you are a developer using libgit2, we can't help you.

ethomson:
I've rearranged the headers, and asked now for the version being used (release or SHA1 built from) and the operating system(s) that show the behavior.

pks-t:
I'd again re-arrange things a bit. Reproduction steps and expected/observed behavior certainly belong together, so I'd rather rearrange version/operating system to either come before or after the other three points. Looks good, other than that :)

ethomson:
Yep, good call @pks-t - I've re-rearranged things.  :D

PR #3964, opened 2016-10-14
Fix UTF-16 to UTF-8 issue in Windows XP

Sanzath:
When on Windows XP, the Windows function WideCharToMultiByte cases an "ERROR_INVALID_FLAGS" error. According to [the documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/dd374130(v=vs.85).aspx), the WC_ERR_INVALID_CHARS flag only works starting from Windows Vista.

Using the VersionHelpers file, we can easily check if we should use the WC_ERR_INVALID_CHARS. I added a git__get_utf16_to_8_flags() function to return the correct flag (0 or WC_ERR_INVALIR_CHARS). 

Since it looks like all conversions to UTF-8 come from strings that were converted from UTF-8 in the first place, another solution would be to remove the version check and always provide 0 as the dwFlags parameter.

### DIFF
--- a/src/win32/utf-conv.c
+++ b/src/win32/utf-conv.c
@@ -25,14 +25,14 @@ GIT_INLINE(void) git__set_errno(void)
 */
 DWORD git__get_utf6_to_8_flags()
 {
-    DWORD flags = 0;
-    
-    /* The WC_ERR_INVALID_CHARS flag is only usable on Windows Vista or later. On Windows XP, using this
-     * flag when calling WideStringToMultiByte causes the function to fail with ERROR_INVALID_FLAGS */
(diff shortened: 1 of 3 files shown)


arthurschreiber:
I don't think we officially support Windows XP anymore, I think the minimum Windows version we expect is Vista.

This change might be simple enough that it will be accepted into libgit2 (but don't keep your hopes up), and even then, there might be other functions that will not work correctly on Windows XP.

ethomson:
Indeed, we stopped supporting XP several years ago.  There will be many other things that fail on XP.

Sanzath:
Alright, I didn't see that XP wasn't supported. Thanks for the heads-up!

Issue #3969, opened 2016-10-20
assert(count <= MAX_SHUTDOWN_CB && count > 0)


Vict0r-Lee:
Hi ,@carlosmn. I always get thsi assertion in function git__on_shutdown() when I use many threads to open the repository and make some git operation.Certainly,I use git_libgit2_init and git_libgit2_shutdown to start up and shutdown. 

I don't know whta that assertion means and how to avoid it.

Vict0r-Lee:
Please forgive my awful English syntax. ^ ^

pks-t:
libgit2 provides a mechanism to set up global state which is shared among all threads. This global state is initialized via the first call to `git_libgit2_init` and de-initialized on the last call to `git_libgit2_shutdown`. While initializing the global state, we also set-up a number of shutdown callbacks for subsystems. The number of callbacks which can be registered is currently fixed to 8, where 7 of these callback slots are used by libgit2 itself. We try to make sure that these callback slots are filled only once by using platform-specific support for atomics.

Regarding your bug: I can certainly envision some scenarios where we might be racy in initializing our global state when many threads are calling init/shutdown at the same time. Especially the pthread code looks racy to me when these calls are intermangled, which would be a bug on our side.

So: what platform are you on? Is your libgit2 compiled with threading support?

Vict0r-Lee:
@pks-t  thanks for your answer. I changed the source code to become a single thread program and this bug has not appeared again.

Another question :how can I use the multi thread with libgit2.

thanks  again.

carlosmn:
You can see the rules to follow for multi-threaded programs in the [THREADING.md](https://github.com/libgit2/libgit2/blob/master/THREADING.md) document.

The initialization functions have to be called in single-threaded context, as this is what is used to set up the multithreading capabilities.

pks-t:
Ah, I've been searching for such a document but didn't spot it. I did expect that these requirements would have been part of `git_libgit2_init`'s documentation. Especially since it states that it "may be called multiple times", without any remarks. I'll fix this.

Vict0r-Lee:
Thanks for everyone's "return".It's my honour to improve libgit2 with you.

ethomson:
[...]

PR #3972, opened 2016-10-27
global: clarify correct usage of `git_libgit2_init` with threads

pks-t:
`git_libgit2_init` needs to be called is a single-threaded
context first, as it also sets up threading-related data
structures. While the fact that it does so is mentioned in the
function's documentation, we do not mention the consequences.

Improve the documentation so that we explicitly note that
`git_libgit2_init` shall be called before making use of threads.

Refs #3969


pks-t:
Closing in favor of #3974

PR #3974, opened 2016-10-27
global: synchronize initialization and shutdown with pthreads

pks-t:
When trying to initialize and tear down global data structures
from different threads at once with `git_libgit2_init` and
`git_libgit2_shutdown`, we race around initializing data. While
we use `pthread_once` to assert that we only initilize data a
single time, we actually reset the `pthread_once_t` on the last
call to `git_libgit2_shutdown`. As resetting this variable is not
synchronized with other threads trying to access it, this is
actually racy when one thread tries to do a complete shutdown of
libgit2 while another thread tries to initialize it.

Fix the issue by creating a mutex which synchronizes `init_once`
and the library shutdown.

### DIFF
--- a/src/global.c
+++ b/src/global.c
@@ -247,6 +247,7 @@ BOOL WINAPI DllMain(HINSTANCE hInstDll, DWORD fdwReason, LPVOID lpvReserved)
 #elif defined(GIT_THREADS) && defined(_POSIX_THREADS)
 
 static pthread_key_t _tls_key;
+static pthread_mutex_t _init_mutex = PTHREAD_MUTEX_INITIALIZER;
 static pthread_once_t _once_init = PTHREAD_ONCE_INIT;
 int init_error = 0;
 
@@ -268,12 +269,19 @@ static void init_once(void)
 
 int git_libgit2_init(void)
 {
-	int ret;
+	int ret, err;
 
 	ret = git_atomic_inc(&git__n_inits);
-	pthread_once(&_once_init, init_once);
 
-	return init_error ? init_error : ret;
+	if ((err = pthread_mutex_lock(&_init_mutex)) != 0)
+		return err;
+	err = pthread_once(&_once_init, init_once);
+	err |= pthread_mutex_unlock(&_init_mutex);
+
+	if (err || init_error)
+		return err | init_error;
+
+	return ret;
 }
 
 int git_libgit2_shutdown(void)
@@ -283,6 +291,9 @@ int git_libgit2_shutdown(void)
 	int ret;
 
 	if ((ret = git_atomic_dec(&git__n_inits)) != 0)
+		return ret;
+
+	if ((ret = pthread_mutex_lock(&_init_mutex)) != 0)
 		return ret;
 
 	/* Shut down any subsystems that have global state */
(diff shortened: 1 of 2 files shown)


pks-t:
I tried to come up with a test, but it wasn't easily possible as our testing frameworks requires global state as well. I used the following test program to verify the issue vanished:

```C
#include <pthread.h>

#include "common.h"

void *run(void *data)
{
	int i;
	for (i = 0; i < 20; i++)
	{
		git_libgit2_init();
		git_libgit2_shutdown();
	}
	return data;
}

{
	pthread_t threads[10];
	unsigned i;

	for (i = 0; i < sizeof(threads) / sizeof(*threads); i++)
	{
	    pthread_create(&threads[i], NULL, run, NULL);
	}

	for (i = 0; i < sizeof(threads) / sizeof(*threads); i++)
	{
	    pthread_join(threads[i], NULL);
	}

	return 0;
}
```

pks-t:
This in fact supersedes #3972 and should fix #3969

ethomson:
So I may not be reasoning about this correctly, but I don't think this is quite sufficient.  `pthread_once` should ensure that the `init_once` function is only called once, and synchronously (with respect to other callers with the same once token).

In your sample program, you're `init`ing and `shutdown`ing.  But if you ever have a `shutdown` that actually tears down the global state, you'll never bring it back.  That `once_init` should really only ever get called _once_.

I suspect that - independent of threading - if you just `git_libgit2_init`, `git_libgit2_shutdown`, `git_libgit2_init` that you are not actually initialized and the global state is a mess...

Or am I not thinking about this correctly?

pks-t:
No, we do actually re-initialize the pthread_once thingy. See `git_libgit2_shutdown`, where we reset `_once_init` with `pthread_once_t new_once = PTHREAD_ONCE_INIT; _once_init = new_once;`.

pks-t:
Just to write down my thoughts and why it is correct to reset `_once_init`: when the last shutdown has been called, we actually _do_ want to free all related data structures such as to avoid memory leaks in libgit2. So we are forced to reset `_once_init`, because when somebody calls `init` again we have to re-initialize all data structures.

Usually I'd advice callers to not initialize/shutdown repeatedly, but I think we should nevertheless support this use case.

ethomson:
> No, we do actually re-initialize the pthread_once thingy. See git_libgit2_shutdown, where we reset _once_init with pthread_once_t new_once = PTHREAD_ONCE_INIT; _once_init = new_once;.

🙇  Thank you, I didn't see that.  😪😫😪

So I think that there's still a situation here.  Since we increment and decrement outside of the mutex, we could have two threads racing:

Thread A enters `git_libgit2_shutdown` and decrements `git__n_inits` from 1 -> 0 and is paused before locking the mutex and performing the shutdown.
Thread B enters `git_libgit2_init` and increments `git__n_inits` from 0 -> 1.  `pthread_once` will not run in this case.
Thread A is woken back up, locks the mutex and finally performs the shutdown.

pks-t:
Oh, yeah, you're right. The mutex also has to synchronize `pthread_once(once_init)`, not only the function called by `pthread_once`. Thanks :)

pks-t:
The issue should be fixed. I synchronize over `pthread_once(&_once_init, init_once)` now instead of synchronizing inside of `init_once`.

pks-t:
The failing test is unrelated

pks-t:
I accidentally included the example demonstrating the fix. Removed the file

pks-t:
Fixed another issue with re-initialization when not using threads which were caught by a set of new tests

PR #3975, opened 2016-10-28
CI Improvements

pks-t:
Just realized our Coverity builds were not running for nearly three months. Seems as if Coverity did change the download URLs for their build tools. I've fixed these and re-added a check so that we only let Coverity analyze our master branch.

I also removed valgrind from the allowed failures. I didn't spot any failures from valgrinds in the last builds, so I'd argue its a good thing to do to notice new failures. I'd also like to disallow failures for Coverity, but unfortunately the build is still too unstable due to quotas and occasional problems with downloading their toolset.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,9 +47,6 @@ matrix:
      os: linux
  allow_failures:
    - env: COVERITY=1
-   - env:
-       - VALGRIND=1
-         OPTIONS="-DBUILD_CLAR=ON -DBUILD_EXAMPLES=OFF -DDEBUG_POOL=ON -DCMAKE_BUILD_TYPE=Debug"
 
 install:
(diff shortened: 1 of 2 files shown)


carlosmn:
The main reason for allowing valgrind runs to fail is so that travis can report back with success before they're done, as they take considerably more to run than the others.

pks-t:
Does this really make sense though? I mean when you take this particular build, the macOS builds are still two minutes slower compared to the valgrind builds (11 minutes for valgrind, 13 minutes for macOS). So there's not much to be gained by this, actually.

pks-t:
Just realized that we're also producing an errored job because of a missing Coverity authentication token if Coverity is actually not going to run. Added another commit fixing this issue.

carlosmn:
The Mac build wasn't always this slow. But I noticed on a different project that Travis won't send the "green" message until the whole build is complete anyway, so it's probably fine.

We're not doing a lot in the valgrind build anyway, since we've no code to fail in case valgrind detects new leaks.

PR #3977, opened 2016-10-28
Fix forced branch creation on HEAD of a bare repo.

jfultz:
The code correctly detects that forced creation of a branch on a
nonbare repo should not be able to overwrite a branch which is
the HEAD reference.  But there's no reason to prevent this on
a bare repo, and in fact, git allows this.  I.e.,

   git branch -f master new_sha

works on a bare repo with HEAD set to master.  This change fixes
that problem, and updates tests so that, for this case, both the
bare and nonbare cases are checked for correct behavior.

### DIFF
--- a/tests/refs/branches/create.c
+++ b/tests/refs/branches/create.c
@@ -100,6 +100,7 @@ void test_refs_branches_create__can_force_create_over_current_branch_in_bare_rep
 	oid = git_commit_id(target);
 
 	cl_git_pass(git_branch_create(&branch, repo, "master", target, 1));
+	git_reference_free(branch);
 	branch = NULL;
 	cl_git_pass(git_branch_lookup(&branch, repo, "master", GIT_BRANCH_LOCAL));
 	cl_assert_equal_s("refs/heads/master", git_reference_name(branch));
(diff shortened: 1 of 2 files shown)


tests/refs/branches/create.c
```diff
@@ -84,6 +88,25 @@ void test_refs_branches_create__cannot_force_create_over_current_branch(void)
...
+	git_reference *branch2;
+	retrieve_known_commit(&target, repo);
+
+	cl_git_pass(git_branch_lookup(&branch2, repo, "master", GIT_BRANCH_LOCAL));
+	cl_assert_equal_s("refs/heads/master", git_reference_name(branch2));
+	cl_assert_equal_i(true, git_branch_is_head(branch2));
+	oid = git_commit_id(target);
+
+	cl_git_pass(git_branch_create(&branch, repo, "master", target, 1));
+	branch = NULL;
```
pks-t: This leaks memory as you do not free `branch` here and overwrite it afterwards.

jfultz: Copied and pasted the code from the legacy test case, but in the legacy test case it should fail, so there's nothing to clean up. Oops. :P

pks-t:
Looks good. Will wait for comments and merge later this week if nothing comes up. Thanks :tada:

pks-t:
Manually squashed in the fixup and merged via 6e2fab9. Thanks for your work

PR #3978, opened 2016-10-31
Small documentation improvements

pks-t:


THREADING.md
```diff
@@ -88,7 +88,7 @@ of libssh2 as described above, `git_openssl_set_locking()` is a no-op.
 
 If your programming language offers a package/bindings for OpenSSL,
 you should very strongly prefer to use that in order to set up
-locking, as they provide a level of coördination which is impossible
+locking, as they provide a level of coordination which is impossible
```
ethomson: But, but, but...  the coordination was so [_metal_](https://en.wikipedia.org/wiki/Metal_umlaut) before...!

carlosmn: This _diaeresis_ is very much on purpose https://en.wikipedia.org/wiki/Diaeresis_(diacritic)#English

PR #3980, opened 2016-11-01
Documentation fixes

tiennou:
Note: 8149b17 removes git_remote_rename_problem_cb. Not sure when it stopped being used though.

### DIFF
--- a/include/git2/describe.h
+++ b/include/git2/describe.h
@@ -44,8 +44,8 @@ typedef enum {
 typedef struct git_describe_options {
 	unsigned int version;
 
-	unsigned int max_candidates_tags; /** default: 10 */
-	unsigned int describe_strategy; /** default: GIT_DESCRIBE_DEFAULT */
+	unsigned int max_candidates_tags; /**< default: 10 */
+	unsigned int describe_strategy; /**< default: GIT_DESCRIBE_DEFAULT */
 	const char *pattern;
 	/**
 	 * When calculating the distance from the matching tag or
@@ -105,6 +105,9 @@ typedef struct {
 
 GIT_EXTERN(int) git_describe_init_format_options(git_describe_format_options *opts, unsigned int version);
 
+/**
+ * A struct that stores the result of a describe operation.
+ */
 typedef struct git_describe_result git_describe_result;
 
 /**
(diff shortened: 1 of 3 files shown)


include/git2/transaction.h
```diff
@@ -8,6 +8,14 @@
 #define INCLUDE_git_transaction_h__
 
 #include "common.h"
+
+/**
+ * @file git2/transaction.h
+ * @brief Git transaction routines.
+ * @defgroup Git transaction routines.
```
pks-t: This should be `@defgroup git_transaction Git transaction routines` instead. Please also remove the trailing dots from `@brief` and `@defgroup`.

pks-t:
Oh, one more thing. Could you maybe clarify the commit messages a bit by prefixing them with the module they're touch primaily, please? E.g. "remote: fix indentation".

Thanks for your work :+1:

include/git2/transaction.h
carlosmn: We don't actually use these blocks for anything with our current documentation generator, so it should just go away.

include/git2/describe.h
```diff
@@ -105,6 +105,9 @@ typedef struct {
 
 GIT_EXTERN(int) git_describe_init_format_options(git_describe_format_options *opts, unsigned int version);
 
+/**
+ * An opaque struct that stores the result of a describe operation.
+ */
 typedef struct git_describe_result git_describe_result;
```
carlosmn: The opaqueness is not something we need to specify in the text.

include/git2/transaction.h
tiennou: I was trying to make docurium document those functions. Wasn't sure if the group being defined was a prerequisite, but I'll take them out then.

include/git2/describe.h
tiennou: Will fix.

tiennou:
Patch series redone from scratch.

`docurium` is giving me a hard time tough, as some things are left out of the parsing (eg. the whole of `transaction.h`, and things like `git_branch_remote_name` don't appear in the docs). Using the debug-heavy version I have here, it seems that clang sometimes misinterprets comments in some cases (eg, I get lots of `:comment_null`).

PR #3983, opened 2016-11-01
transports: smart: abort on early end of stream

pks-t:
I've started playing around with mitmproxy to test our online protocol and was able to force libgit2 into an infinite loop on first try. The issue occurs if we fetch from a remote via the smart protocol if the last flush packet from the remote is modified.

This PR fixes the infinite loop. I'm actually not sure why we only returned an EOF before when `flush == 0`, I couldn't find any reason here. One thing I could imagine is that we actually wanted to return `0` if the last received packet was actually a flush packet _and_ the socket returned EOF so that we wouldn't produce an error if we received less flushes than expected. But we actually only use the function once and only check if the return code was an error, so we wouldn't handle positive values (e.g. there are outstanding flushes).

Yeah, so I'm quite clueless here. Anyway, the issue obviously requires a fix as it's remotely exploitable, so I'm putting this up for discussion here.

### DIFF
--- a/include/git2/proxy.h
+++ b/include/git2/proxy.h
@@ -19,7 +19,7 @@ typedef enum {
 	/**
 	 * Do not attempt to connect through a proxy
 	 *
-	 * If built against lbicurl, it itself may attempt to connect
+	 * If built against libcurl, it itself may attempt to connect
 	 * to a proxy if the environment variables specify it.
 	 */
 	GIT_PROXY_NONE,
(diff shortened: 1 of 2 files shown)


pks-t:
Got another case where we were looping endlessly in search for additional data.

pks-t:
Any comments on this PR? I don't feel like just merging it without anybody else having a look at it, but I feel like this PR should be merged due to its possible security implications.

carlosmn:
This makes sense. I suspect the code was trying to handle the case where it times out before reading, but we don't actually do that currently, so we should indeed consider a `0` to mean the server doesn't want to talk to us anymore.

PR #3985, opened 2016-11-02
Update THREADING for OpenSSL 1.1

carlosmn:
Starting at OpenSSL 1.1.0 it is thead-safe by itself, so mention that for people linking against it, they do not need to take further steps.

Also update the cURL link which has changed its location.

I've tried to reword things a bit to make it clearer (and remove a "may" for our last-resort locking function which might be taken as too much of a promise) but knowing what to do in order to use the network safely is still a minefield where you might need to initialise threading for two different libraries, and you don't know whether you have to until runtime.

### DIFF
--- a/THREADING.md
+++ b/THREADING.md
@@ -63,7 +63,7 @@ General Case
 ------------
 
 By default we use libcurl, which has its own ![recommendations for
-thread safety](http://curl.haxx.se/libcurl/c/libcurl-tutorial.html#Multi-threading).
+thread safety](https://curl.haxx.se/libcurl/c/threadsafe.html).
 
 If libcurl was not found or was disabled, libgit2 uses OpenSSL to be
 able to use HTTPS as a transport. This library is made to be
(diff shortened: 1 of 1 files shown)


ethomson:
👍 from me - I'll leave this open for a hot minute in case somebody else has some input.

THREADING.md
```diff
@@ -62,29 +62,34 @@ general case still affects you if you use ssh.
...
-libgit2 and libssh2, you only need to set up threading for OpenSSL once.
-
-libgit2 does provide a last-resort convenience function
+If it's available, by default we use libcurl to provide HTTP tunneling support,
+which may be linked against a number of cryptographic libraries and has its
+own
+[recommendations for thread safety](https://curl.haxx.se/libcurl/c/threadsafe.html).
+
+If there are no alternative TLS implementations (currently only
+SecureTransport), libgit2 uses OpenSSL in order to use HTTPS as a trasnport.
```
pks-t: There's a typo in "trasnport"

PR #3992, opened 2016-11-10
git_repository_open_ext: fix handling of $GIT_NAMESPACE

joshtriplett:
The existing code would set a namespace of "" (empty string) with
GIT_NAMESPACE unset.  In a repository where refs/heads/namespaces/
exists, that can produce incorrect results.  Detect that case and avoid
setting the namespace at all.

Since that makes the last assignment to error, and the previous
assignment can potentially get GIT_ENOTFOUND, set error to 0 explicitly
to prevent the call from incorrectly failing with GIT_ENOTFOUND.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -613,9 +613,10 @@ static int _git_repository_open_ext_from_env(
 		git_repository_set_odb(repo, odb);
 
 	error = git__getenv(&alts_buf, "GIT_ALTERNATE_OBJECT_DIRECTORIES");
-	if (error == GIT_ENOTFOUND)
+	if (error == GIT_ENOTFOUND) {
 		giterr_clear();
-	else if (error < 0)
+		error = 0;
+	} else if (error < 0)
 		goto error;
         else {
 		const char *end;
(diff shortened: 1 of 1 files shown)


src/repository.c
```diff
@@ -638,9 +638,12 @@ static int _git_repository_open_ext_from_env(
...
 
-	error = git_repository_set_namespace(repo, git_buf_cstr(&namespace_buf));
-	if (error < 0)
-		goto error;
+	if (git_buf_len(&namespace_buf)) {
+		error = git_repository_set_namespace(repo, git_buf_cstr(&namespace_buf));
+		if (error < 0)
+			goto error;
+	}
+	error = 0;
```
pks-t: This looks a bit backwards here, as you are in fact not clearing the error of the leading block but instead the error from the previous `git__getenv`. It would be much clearer if you'd instead clear the error in the previous `if (error == GIT_ENOTFOUND)` condition, where we already call `giterr_clear();`.

pks-t: An additional note, which is not your fault but might be improved when we're already touching this. I think the code would be nicer if you moved the call to `git__getenv` and setting the namespace together. So you could just move down getting the environment variable and then set the namespace based on the function's return value.

joshtriplett:
@pks-t Thanks for the review.

Updated to move the error handling to a clearer location.

pks-t:
Looks good now, thanks for this :+1:

Issue #3994, opened 2016-11-10
git_repository_discover() does not work for files


max630:
After merges d8243465be and ebeb56f0f5, `git_repository_discover()` fails when start_path points to a file in repository with message:

> The `.git` file at '/tmp/libgit2_tests_tgTHjU/discover__file_repo/subdir/file' is malformed

Documentation is not clear if passing file to the function expected or not, but apparently there is already software which uses it - for example, GitDiffMargin -, so it's sort of regression.

pks-t:
This should be fixed by #3998 as soon as it is merged

carlosmn:
Fixed via #3998

PR #3998, opened 2016-11-11
Repository discovery starting from files

pks-t:
This PR is as a response to #3994. I think it does make sense to make a repository discoverable even if initially, a path to a file inside the repository was specified.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -410,7 +410,7 @@ static int find_repo(
 					break;
 				}
 			}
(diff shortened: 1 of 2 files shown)


ethomson:
This looks 👍 to me, this seems quite low risk for 0.25, @carlosmn do you agree?

carlosmn:
Yeah, since this was apparently working before, I suppose not supporting it would technically break the API.

Issue #4000, opened 2016-11-13
Crash listing git files on mounted cifs share


dhaumann:
This is an issue that was reported against the Kate text editor here:
https://bugs.kde.org/show_bug.cgi?id=372196

The bug report says:
> I just tried to open a shell script on a mounted shared. Kate simply crashed.
> [...] The same file opens nicely if copied to local FS.
> The crash can be reproduced every time.

The important parts of the backtrace:
Thread 15 (Thread 0x7fb271e86700 (LWP 31811)):
[KCrash Handler]
#6  0x00007fb2a5bb5956 in strlen () from /lib64/libc.so.6
#7  0x00007fb2a5b7e323 in vfprintf () from /lib64/libc.so.6
#8  0x00007fb2a5c2b366 in __vsnprintf_chk () from /lib64/libc.so.6
#9  0x00007fb2a53dbe60 in vsnprintf (__ap=0x7fb271e85240, __fmt=0x7fb2a547663f "Could not read directory '%s'", __n=<optimized out>, __s=<optimized out>) at /usr/include/bits/stdio2.h:77
#10 git_buf_vprintf (buf=buf@entry=0x7fb26c00d5d8, format=format@entry=0x7fb2a547663f "Could not read directory '%s'", ap=ap@entry=0x7fb271e85290) at /usr/src/debug/libgit2-0.24.1/src/buffer.c:379
#11 0x00007fb2a53fe0b4 in giterr_set (error_class=error_class@entry=2, string=string@entry=0x7fb2a547663f "Could not read directory '%s'") at /usr/src/debug/libgit2-0.24.1/src/errors.c:62
#12 0x00007fb2a543418d in git_path_diriter_next (diriter=diriter@entry=0x7fb271e85420) at /usr/src/debug/libgit2-0.24.1/src/path.c:1316
#13 0x00007fb2a54179f0 in dirload_with_stat (fi=0x7fb26c015a70, contents=0x7fb26c013e48) at /usr/src/debug/libgit2-0.24.1/src/iterator.c:1256
#14 fs_iterator__expand_dir (fi=fi@entry=0x7fb26c015a70) at /usr/src/debug/libgit2-0.24.1/src/iterator.c:1363
#15 0x00007fb2a5417e45 in fs_iterator__advance_into (entry=0x7fb271e85580, iter=0x7fb26c015a70) at /usr/src/debug/libgit2-0.24.1/src/iterator.c:1432
[...]

PR #4002, opened 2016-11-14
giterr format

pks-t:
While inspecting #4000, I've been very surprised to find out that we were passing a `git_buf` pointer to `giterr_set` without any warning of the compiler. Seems we didn't put a printf-attribute hint at the declaration, causing the compiler to accept whatever we put there.

Next to the many cases of `"%d"` instead of `"%"PRIuZ` which can cause us to overflow in the output, there were also three real errors where we passed a structure pointer to "%s" instead of the real value.

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -173,7 +173,7 @@ static int apply_hunk(
 		git_diff_line *line = git_array_get(patch->lines, linenum);
 
 		if (!line) {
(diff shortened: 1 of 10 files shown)


carlosmn:
While this shouldn't have any effect on that issue as the address of a struct is the address of its first field, it's a good idea to tell the compiler about the printf format and fix its warnings.

ethomson:
Nice, thanks!

PR #4003, opened 2016-11-14
Use the sorted input in the tree updater

### LINKED ISSUE #3804: Tree updater does not use the sorted inputs
Oops. I'm leaving this here so I don't forget to fix it when I get back.

carlosmn:
The loop/algo is made to assume that we have sorted inputs. We sort them to make sure but we failed to actually _use_ the sorted list. The tests were inadvertently written with inputs that did not show this bad behaviour.

This fixes #3804


ethomson:
👍

PR #4008, opened 2016-11-15
sortedcache: plug leaked file descriptor

pks-t:


pks-t:
Just plugging another memleak while at it

PR #4009, opened 2016-11-16
use `giterr_set_str()` wherever possible

pranitbauva1997:
`giterr_set()` is used when it is required to format a string, and since
we don't really require it for this case, it is better to stick to
`giterr_set_str()`.

This also suppresses a warning(-Wformat-security) raised by the compiler.

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -552,7 +552,7 @@ int git_index_clear(git_index *index)
 
 static int create_index_error(int error, const char *msg)
 {
-	giterr_set(GITERR_INDEX, msg);
+	giterr_set_str(GITERR_INDEX, msg);
 	return error;
 }
 
--- a/src/zstream.c
+++ b/src/zstream.c
@@ -21,7 +21,7 @@ static int zstream_seterr(git_zstream *zs)
 	if (zs->zerr == Z_MEM_ERROR)
 		giterr_set_oom();
 	else if (zs->z.msg)
-		giterr_set(GITERR_ZLIB, zs->z.msg);
+		giterr_set_str(GITERR_ZLIB, zs->z.msg);
 	else
 		giterr_set(GITERR_ZLIB, "Unknown compression error");
 


ethomson:
I'm not sure that I fully understand your comment - `create_index_error` does pass a `const char *` for example.  That aside, this is a good fix since treating strings that contain an error message as a format string is definitely incorrect.  (If the error message was `%s`, obviously this would be a bummer.)

carlosmn:
What compiler are you using? We do run with `-Wall` by default, and I haven't seen such warnings. As a general rule, C does not complain when you _add_ `const`ness to a variable.

In the places were we don't expect to perform any substitution, we should rather be using `giterr_set_str()` which doesn't involve any parsing machinery.

src/index.c
```diff
@@ -2358,7 +2358,7 @@ static size_t read_entry(
 		entry.path = (char *)path_ptr;
 	} else {
 		size_t varint_len;
-		size_t shared = git_decode_varint((const unsigned char *)path_ptr, 
+		size_t shared = git_decode_varint((const unsigned char *)path_ptr,
```
carlosmn: Let's not do unrelated whitespace changes.

pranitbauva1997: Sure! My editor corrected it by default.

pranitbauva1997:
@ethomson I was unsure about it. But since that warning only occurs when the above-specified thing happens, I _assumed_ it to be some technicality within variable arguments (I am not sure about this). I guess it is better to remove it altogether.

pranitbauva1997:
@carlosmn There seems to be a problem with both of our local builds I guess because after I pushed this out, I was checking out the travis-ci builds for [master](https://travis-ci.org/libgit2/libgit2/jobs/176126555) as well as for this [PR](https://travis-ci.org/libgit2/libgit2/builds/176476155) where you can easily see that it is displaying 3 warnings in which this is one of them.
I did not add an extra "-Wall flag".

Also, I am not getting the other two warnings locally which are clearly shown in the master build on travis-ci.

pranitbauva1997:
@ethomson @carlosmn I think using `giterr_set_str` would be a better way out! Thanks!

pranitbauva1997:
@carlosmn @ethomson I have updated the PR.

carlosmn:
The warning I see on that build is

    /home/travis/build/libgit2/libgit2/src/index.c:555:2: warning: format not a string literal and no format arguments [-Wformat-security]

 which is the issue with formatting that @ethomson mentioned, it's nothing to do with `const`.

Ubuntu used to harden their GCC and the version string on Travis suggests they're still adding patches which is why I don't see this on my Debian GCC.

carlosmn:
The warning is likely a result of #4002 which told the compiler that we accept format strings. @pks-t fixed a lot of warnings, but presumably his GCC doesn't contain `-Wformat-security` in the "all" flags either.

pranitbauva1997:
@carlosmn Thanks for explaining! I was under the impression that it would only occur because of `const` things.

pks-t:
I've got a hardened gcc, as well, but it in fact does not contain "-Wformat-security".

Which makes me think - how about adding a build job that compiles with "-Werror" so that we catch compilation warnings early on? Oh wait, there are these curl warnings which would cause the build to fail... sucks :/

carlosmn:
That curl weirdness is disappointing, but I'd like to get rid of curl regardless. We can certainly add a few choice variables like `-Wformat-security` to the block of add-if-supported like we do for `-Wdocumentation`.

pks-t:
Makes sense to me, yes

PR #4010, opened 2016-11-18
Introduce some clar helpers for child threads

ethomson:
When running in a child thread, we cannot `clar_git_pass` (or anything that will cause clar to `longjmp` to its error handler):

> The effect of a call to `longjmp()` where initialization of the `jmp_buf` structure was not performed in the calling thread is undefined.

Instead, provide some helpers to set up an error context in the child thread that the clar execution thread can then restore and abort when safe.  Use this for the `threads::refdb` test so that it errors gracefully, instead of crashing.

Obviously this does not fix the underlying problem in the`threads::refdb` test, but it does make it easier to reason about.

### DIFF
--- a/tests/clar_libgit2.h
+++ b/tests/clar_libgit2.h
@@ -41,6 +41,47 @@
 	} \
 	} while(0)
 
+/**
+ * Thread safe assertions; you cannot use `cl_git_report_failure` from a
+ * child thread since it will try to `longjmp` to abort and "the effect of
+ * a call to longjmp() where initialization of the jmp_buf structure was
+ * not performed in the calling thread is undefined."
+ *
+ * Instead, callers can provide a clar thread error context to a thread,
+ * which will populate and return it on failure.  Callers can check the
+ * status with `cl_git_thread_check`.
+ */
+typedef struct {
+	int error;
+	const char *file;
+	int line;
+	const char *expr;
+	char error_msg[4096];
+} cl_git_thread_err;
+
+#define cl_git_thread_pass(threaderr, expr) cl_git_thread_pass_(threaderr, (expr), __FILE__, __LINE__)
+
+#define cl_git_thread_pass_(__threaderr, __expr, __file, __line) do { \
+	giterr_clear(); \
+	if ((((cl_git_thread_err *)__threaderr)->error = (__expr)) != 0) { \
+		const git_error *_last = giterr_last(); \
+		((cl_git_thread_err *)__threaderr)->file = __file; \
+		((cl_git_thread_err *)__threaderr)->line = __line; \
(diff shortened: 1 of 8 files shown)


tests/clar_libgit2.h
```diff
@@ -41,6 +41,47 @@
...
+	const char *expr;
+	char error_msg[4096];
+} cl_git_thread_err;
+
+#define cl_git_thread_pass(threaderr, expr) cl_git_thread_pass_(threaderr, (expr), __FILE__, __LINE__)
+
+#define cl_git_thread_pass_(__threaderr, __expr, __file, __line) do { \
+	giterr_clear(); \
+	if ((((cl_git_thread_err *)__threaderr)->error = (__expr)) != 0) { \
+		const git_error *_last = giterr_last(); \
```
carlosmn: Doesn't the `giterr_clear()` just above this get rid of the information we're trying to store?

tests/clar_libgit2.h
```diff
@@ -41,6 +41,47 @@
...
+		((cl_git_thread_err *)__threaderr)->file = __file; \
+		((cl_git_thread_err *)__threaderr)->line = __line; \
+		((cl_git_thread_err *)__threaderr)->expr = "Function call failed: " #__expr; \
+		p_snprintf(((cl_git_thread_err *)__threaderr)->error_msg, 4096, "thread 0x%" PRIxZ " - error %d - %s", \
+			git_thread_currentid(), ((cl_git_thread_err *)__threaderr)->error, \
+			_last ? _last->message : "<no message>"); \
```
carlosmn: `_last->message` would get freed when the thread exits, no? Wouldn't we want to `strdup()` it or `git_buf_detach()`?

tests/clar_libgit2.h
ethomson: This is the same mechanism that `cl_git_pass` uses.  It's aimed to work as `cl_git_pass(git_function...)`, so it clears any prior errors before executing `git_function`.

Getting rid of this would be annoying when we fail for some reason that _doesn't_ set the git error.  For example, when a function returns `GIT_ITEROVER` or something unexpectedly.  It would be weird to report the last error message which probably has nothing to do with the failure.

It breaks down in this case, where we're `cl_git_pass`ing after executing the thing that failed - sort of like we do here.

tests/clar_libgit2.h
ethomson: It will, that's why we're stuffing it into the `error_msg` to preserve that data.

carlosmn: Yes, I completely missed the fact that `expr` is an expression rather than the result; it must be all of those backticks confusing me.

PR #4014, opened 2016-11-20
Properly pass `wchar *` type to giterr_set

bokic:
Ensure correct type is passed, no garbage text is sent, and mute compiler warning.

### DIFF
--- a/src/win32/w32_util.h
+++ b/src/win32/w32_util.h
@@ -174,7 +174,7 @@ GIT_INLINE(int) git_win32__file_attribute_to_stat(
 			/* st_size gets the UTF-8 length of the target name, in bytes,
 			 * not counting the NULL terminator */
 			if ((st->st_size = git__utf16_to_8(NULL, 0, target)) < 0) {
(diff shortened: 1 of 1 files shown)


src/win32/w32_util.h
```diff
@@ -174,7 +174,7 @@ GIT_INLINE(int) git_win32__file_attribute_to_stat(
 			/* st_size gets the UTF-8 length of the target name, in bytes,
 			 * not counting the NULL terminator */
 			if ((st->st_size = git__utf16_to_8(NULL, 0, target)) < 0) {
-				giterr_set(GITERR_OS, "Could not convert reparse point name for '%s'", path);
+				giterr_set(GITERR_OS, "Could not convert reparse point name for '%S'", path);
```
bokic: Just checked with MinGW 5.3.0, and, you are right. `%S` seem to reverse the character width.

So, yes `%ls` format is better for this case.

ethomson:
Thanks!  😀

PR #4015, opened 2016-11-21
Allow Windows with WinHTTP to use external http-parser

staticfloat:
Currently, when building `libgit2` on windows, if building against WinHTTP, there is no option to use an external `http-parser` library, we are [locked into using the builtin `http-parser` library](https://github.com/libgit2/libgit2/blob/ae5838f118a4819e608990a815bf8fc482be5772/CMakeLists.txt#L238-L241).  I'd guess this was done because there were purportedly some differences between the third-party `http-parser` library on windows and the builtin `http-parser` library.  However, I cannot find any special treatment of the windows build of the internal `http-parser`, so it stands to reason that we should be able to use the same external `http-parser` library on windows as we do on Linux or OSX.

This patch allows searching for `http-parser` libraries on any platform.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -235,10 +235,21 @@ IF (WIN32 AND EMBED_SSH_PATH)
 	ADD_DEFINITIONS(-DGIT_SSH)
 ENDIF()
 
-IF (WIN32 AND WINHTTP)
-	ADD_DEFINITIONS(-DGIT_WINHTTP)
+
+FIND_PACKAGE(HTTP_Parser)
+IF (HTTP_PARSER_FOUND AND HTTP_PARSER_VERSION_MAJOR EQUAL 2)
+	INCLUDE_DIRECTORIES(${HTTP_PARSER_INCLUDE_DIRS})
+	LINK_LIBRARIES(${HTTP_PARSER_LIBRARIES})
+	LIST(APPEND LIBGIT2_PC_LIBS "-lhttp_parser")
+ELSE()
+	MESSAGE(STATUS "http-parser was not found or is too old; using bundled 3rd-party sources.")
 	INCLUDE_DIRECTORIES(deps/http-parser)
 	FILE(GLOB SRC_HTTP deps/http-parser/*.c deps/http-parser/*.h)
+ENDIF()
+
+
+IF (WIN32 AND WINHTTP)
+	ADD_DEFINITIONS(-DGIT_WINHTTP)
 
 	# Since MinGW does not come with headers or an import library for winhttp,
 	# we have to include a private header and generate our own import library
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -235,10 +235,21 @@ IF (WIN32 AND EMBED_SSH_PATH)
 	ADD_DEFINITIONS(-DGIT_SSH)
 ENDIF()
 
-IF (WIN32 AND WINHTTP)
-	ADD_DEFINITIONS(-DGIT_WINHTTP)
+
+FIND_PACKAGE(HTTP_Parser)
+IF (HTTP_PARSER_FOUND AND HTTP_PARSER_VERSION_MAJOR EQUAL 2)
```
mikeando: I'd say this check `HTTP_PARSER_VERSION_MAJOR EQUAL 2` doesn't quite match up with the message for when it fails. `http-parser was not found or is too old; using bundled 3rd-party sources.` since the failure could (in the future) be due to a version 3 being released?

But this was the case in the old version too, so maybe it doesn't matter.

staticfloat: I'd be happy to bundle in a change to the error message as well if the maintainers think that's a good idea.

staticfloat:
I'm assuming the win64 failures are unrelated because they both fail during tests and correctly identify that there is no third-party http-parser library available, so their behavior should be unchanged with respect to this change.

ethomson:
What's the goal here?  Do you have a newer http-parser that you want to use?  Independent of this, does it make sense to also upgrade the http-parser that we're bundling?

carlosmn:
Why do we use http-parser on Windows? If it's because of trying to parse out the URL, we should make it ask WinHTTP instead and just not use http-parser. We're certainly not trying to parse any HTTP with it.

This is also not how you link to libraries on Windows, but if you expect to find something with `pkg-config` then I suppose you're not all that in Windows.

staticfloat:
> What's the goal here? Do you have a newer http-parser that you want to use?

We use `libgit2` across a variety of platforms, and we like to make things as homogenous across platforms as possible.  We recently started building `http-parser` as a shared library (for use with other software) and thought it was a shame to have the same code built and linked in twice.

> This is also not how you link to libraries on Windows, but if you expect to find something with `pkg-config` then I suppose you're not all that in Windows.

I'm open to corrections. :)

CMakeLists.txt
ethomson: Yeah, I think that would be delightful.  Something like `http-parser version 2 not found...` would be preferable.

CMakeLists.txt
```diff
@@ -235,10 +235,21 @@ IF (WIN32 AND EMBED_SSH_PATH)
 	ADD_DEFINITIONS(-DGIT_SSH)
 ENDIF()
 
-IF (WIN32 AND WINHTTP)
-	ADD_DEFINITIONS(-DGIT_WINHTTP)
+
+FIND_PACKAGE(HTTP_Parser)
```
ethomson: It probably makes sense to move this block down toward the zlib block; these are quite similar, they both decide whether to load external dependencies, or to use the bundled third party libraries.  Also, this looks like it adds some extra whitespace, which would be nice to ⚡️ in that move.

staticfloat:
Addressed @ethomson's comments.

ethomson:
Thanks @staticfloat , this looks good to me.  Enjoy your external httpparser goodness!

PR #4016, opened 2016-11-24
Submodule optimization

### LINKED ISSUE #3756: git_submodule_lookup is O(N) w.r.t. number of submodules
causing many operations, such as git_rebase_init (which results in calls to git_submodule_lookup several times for each submodule) to be O(N^2).

On a 3.1Ghz MBP with SSD, for example, a single call to git_submodule_lookup in a repository with 4,000 submodules takes over 20ms.

I am very happy to work on this myself and submit a patch, but I wanted to open a discussion and get advice on how to proceed. One suggestion: allow an option to configure a repository to cache this information on first submodule access.

novalis:
The bulk of this work was done by @bpeabody.  I just made some minor adjustments.  This fixes #3756.

### DIFF
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -1544,13 +1544,22 @@ int git_submodule__status(
 		return 0;
 	}
 
-	/* refresh the index OID */
-	if (submodule_update_index(sm) < 0)
-		return -1;
+	/* If the user has requested caching submodule state, performing these
+	 * expensive operations (especially `submodule_update_head`, which is
+	 * bottlenecked on `git_repository_head_tree`) eliminates much of the
+	 * advantage.  We will, therefore, interpret the request for caching to
+	 * apply here to and skip them.
+	 */
 
-	/* refresh the HEAD OID */
-	if (submodule_update_head(sm) < 0)
-		return -1;
+	if (sm->repo->submodule_cache == NULL) {
+		/* refresh the index OID */
+		if (submodule_update_index(sm) < 0)
+			return -1;
+
+		/* refresh the HEAD OID */
+		if (submodule_update_head(sm) < 0)
+			return -1;
+	}
 
 	/* for ignore == dirty, don't scan the working directory */
 	if (ign == GIT_SUBMODULE_IGNORE_DIRTY) {
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -149,40 +149,53 @@ static int find_by_path(const git_config_entry *entry, void *payload)
 }
 
 /**
- * Find out the name of a submodule from its path
+ * Release the name map returned by 'load_submodule_names'.
  */
-static int name_from_path(git_buf *out, git_config *cfg, const char *path)
+static void free_submodule_names(git_strmap *names)
+{
+	git_buf *name = 0;
+	if (names == NULL)
+		return;
+	git_strmap_foreach_value(names, name, {
+		git__free(name);
+	});
+	git_strmap_free(names);
+	return;
+}
+
+/**
+ * Map submodule paths to names.
+ * TODO: for some use-cases, this might need case-folding on a
+ * case-insensitive filesystem
+ */
+static int load_submodule_names(git_strmap *out, git_config *cfg)
 {
 	const char *key = "submodule\\..*\\.path";
 	git_config_iterator *iter;
 	git_config_entry *entry;
-	int error;
+	git_buf buf = GIT_BUF_INIT;
+	int rval;
+	int error = 0;
 
 	if ((error = git_config_iterator_glob_new(&iter, cfg, key)) < 0)
 		return error;
 
-	while ((error = git_config_next(&entry, iter)) == 0) {
+	while (git_config_next(&entry, iter) == 0) {
 		const char *fdot, *ldot;
-		/* TODO: this should maybe be strcasecmp on a case-insensitive fs */
-		if (strcmp(path, entry->value) != 0)
-			continue;
-
 		fdot = strchr(entry->name, '.');
 		ldot = strrchr(entry->name, '.');
 
-		git_buf_clear(out);
-		git_buf_put(out, fdot + 1, ldot - fdot - 1);
-		goto cleanup;
+		git_buf_put(&buf, fdot + 1, ldot - fdot - 1);
+		git_strmap_insert(out, entry->value, git_buf_detach(&buf), rval);
+		if (rval < 0) {
+			giterr_set(GITERR_NOMEMORY, "Error inserting submodule into hash table");
+			free_submodule_names(out);
+			return -1;
+		}
 	}
 
-	if (error == GIT_ITEROVER) {
-		giterr_set(GITERR_SUBMODULE, "could not find a submodule name for '%s'", path);
-		error = GIT_ENOTFOUND;
-	}
-
-cleanup:
 	git_config_iterator_free(iter);
-	return error;
+	return 0;
 }
 
 int git_submodule_lookup(
@@ -195,6 +208,17 @@ int git_submodule_lookup(
 	git_submodule *sm;
 
 	assert(repo && name);
+
+	if (repo->submodule_cache != NULL) {
+		khiter_t pos = git_strmap_lookup_index(repo->submodule_cache, name);
+		if (git_strmap_valid_index(repo->submodule_cache, pos)) {
+			if (out) {
+				*out = git_strmap_value_at(repo->submodule_cache, pos);
+				GIT_REFCOUNT_INC(*out);
+			}
+			return 0;
+		}
+	}
 
 	if ((error = submodule_alloc(&sm, repo, name)) < 0)
 		return error;
@@ -327,19 +351,17 @@ static int submodules_from_index(git_strmap *map, git_index *idx, git_config *cf
 	int error;
 	git_iterator *i;
 	const git_index_entry *entry;
-	git_buf name = GIT_BUF_INIT;
+	git_strmap *names = 0;
+	git_strmap_alloc(&names);
+	if ((error = load_submodule_names(names, cfg)))
+		goto done;
 
 	if ((error = git_iterator_for_index(&i, git_index_owner(idx), idx, NULL)) < 0)
-		return error;
+		goto done;
 
 	while (!(error = git_iterator_advance(&entry, i))) {
 		khiter_t pos = git_strmap_lookup_index(map, entry->path);
 		git_submodule *sm;
-
-		git_buf_clear(&name);
-		if (!name_from_path(&name, cfg, entry->path)) {
-			git_strmap_lookup_index(map, name.ptr);
-		}
 
 		if (git_strmap_valid_index(map, pos)) {
 			sm = git_strmap_value_at(map, pos);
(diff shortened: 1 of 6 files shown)


novalis:
So, I know it's kind of audacious for me to request a merge with both CI systems failing, but honestly I think neither of these failures are my fault.

The same AppVeyor test failures appear on a few other recent pull requests: 
https://ci.appveyor.com/project/libgit2/libgit2/build/2906/job/6da4tyuhgrdkgsb2
https://ci.appveyor.com/project/libgit2/libgit2/build/2903/job/1hfndyy74s5y2hdn

I also think the segfault on Travis is not my fault.  The tests all pass for me locally, and give no warnings under valgrind's memcheck.  There are, however, helgrind errors.

Here are a few other relatively recent cases of segfaults (on AppVeyor rather than Travis, but if the cause is what I think it is, it could appear on any platform).
https://ci.appveyor.com/project/libgit2/libgit2/build/2630/job/l4r12jcb4e42ur89
https://ci.appveyor.com/project/libgit2/libgit2/build/2543/job/vdmkmg8pg02xaei8

I can repro the segfault under a version of libgit2 immediately prior to my changes -- simply running the test suite a few dozen times is sufficient.  The segfault is in attr_cache.free from git_repository.cleanup, which is one of the areas that helgrind flags as hinky.

Sadly, the bug never seems to reproduce itself under valgrind (or even gdb), but since helgrind is complaining, it seems probably threading-related.

Helgrind does not, afaict, complain about our new code.

novalis:
I just did a no-op amend and now Travis passes.  AppVeyor is still bad, but as I noted, it frequently fails this way.

novalis:
Actually, after talking to some colleagues, I decided to rename the functions.  No substantive changes.

novalis:
BTW this fixes https://github.com/libgit2/libgit2/issues/3756

ethomson:
I mentioned in [slack](http://slack.libgit2.org) that this is probably not in scope for this release, which we're just finishing up.

There's not a lot of documentation around how one might want to use this - when would I, as a consumer, need to clear the cache?  Would I then benefit from repopulating it?

I think that @carlosmn asked in #3756 - why can this not be a thing that the library handles for you?  Why would I, as a consumer, have better knowledge of when caching should happen than the library itself?

bpeabody:
@ethomson I can think of a couple of ways the library could handle it for you.

- The library could do caching automatically, invalidating it as necessary, etc.
- Library operations (such as the diff that happens in the beginning of a rebase) that will access all submodules could be changed to load them in a batch operation so that caching isn't necessary.

bpeabody:
Additionally, we could probably separate into a different pull request the performance fixes that don't involve a cache (e.g., s.t. `all_submodules` isn't O(N^2)).

novalis:
Looking at the automatic caching idea:

1. The idea of manual caching is that the user sometimes knows better than libgit2 when it is safe to use the cache.

In the absence of the cache, each `git_submodule_lookup` checks the index, HEAD, and the working directory.  The index check does not usually involve any i/o (because it is cached).  The HEAD and wd checks do.  This is somewhat inconsistent: why do we assume that HEAD could have changed out from under us, but not make that assumption about the index?

User code (e.g our git-meta, via nodegit) might know that it's going to open a repository once, do a thing, and then close it.  So it's OK for that code to turn on caching as soon as it opens the module.

But the Atom editor (for instance) might not be wiling to make that assumption; maybe at some point you go in and checkout a new branch from the command-line.  (I don't actually know if Atom uses this code -- it's just an example).  And at that point, the cache needs invalidation.

2. More generally, `git_submodule_lookup` is a user-visible function. Imagine that it did do automatic caching, with 100% correct invalidation (modulo the index issue mentioned above).  This would mean most every call to `git_submodule_lookup would have to do wix iops (open HEAD, read HEAD, open refs/heads/master, read refs/heads/master, stat .gitmodules (to check whether it's clean), stat the submodule's wd path).  The only thing the cache would be saving would be parsing .gitmodules, which admittedly is somewhat expensive.  Which is more expensive depends on the size of the gitmodules file and the ratio of disk speed to cpu/memory speed.

3. We could also have git_submodule_lookup not expose a user-visible way of manipulating the cache -- instead, ordinary user calls to `git_submodule_lookup` would be uncached.  But internal calls (e.g. from `git_diff` or `git_checkout_index` could first initialize the cache, then do their multiple calls, then tear down the cache.  But this would not help users who do multiple submodule lookups on their own.

novalis:
My conclusion from the above, btw, is that automatic caching doesn't make sense and that we ought to just merge this as-is.

ethomson:
Thanks for the writeup @novalis .  I think that makes sense.  I have some review comments to make but I think the overall strategy makes sense.

src/submodule.c
```diff
@@ -1566,6 +1617,33 @@ int git_submodule_location(unsigned int *location, git_submodule *sm)
 		location, NULL, NULL, NULL, sm, GIT_SUBMODULE_IGNORE_ALL);
 }
 
+int git_repository_submodule_cache_all(git_repository *repo)
```
ethomson: Please don't put `git_repository` functions in a file that is not `repository.c`.  Either:  1) these should either go in `repository.c`, or 2) they should be `git_submodule_cache_all` and `git_submodule_cache_clear`.  The latter is obviously more brief, but I think (1) makes more sense.

include/git2/repository.h
```diff
@@ -770,6 +770,31 @@ GIT_EXTERN(int) git_repository_ident(const char **name, const char **email, cons
...
+ *
+ * Because the `.gitmodules` file is unstructured, loading submodules is an
+ * O(N) operation.  Any operation (such as `git_rebase_init`) that requires
+ * accessing all submodules is O(N^2) in the number of submodules, if it
+ * has to look each one up individually.  This function loads all submodules
+ * and caches them so that subsequent calls to `git_submodule_lookup` are O(1).
+ *
+ * @param repo the repository whose submodules will be cached.
+ */
+GIT_EXTERN(int) git_repository_submodule_cache_all(
```
ethomson: There's no indication of when somebody should call `cache_clear`.  You need to describe what operations would cause a cache coherency problem; it's not at all obvious to the reader and I suspect it's wildly dangerous to call `cache_all` and then (say) check out a new branch that would update a submodule?  I'm just guessing, I don't know, and I'd like this to expand upon that.

I also think that you should move this into `include/git2/sys/repository.h`.  This seems to me to be a cache that you have to manage rather carefully, and the `sys/` directory is for the things that will hurt you badly if you get wrong.

src/submodule.c
```diff
@@ -149,40 +149,53 @@ static int find_by_path(const git_config_entry *entry, void *payload)
...
 
 /**
- * Find out the name of a submodule from its path
+ * Release the name map returned by 'load_submodule_names'.
  */
-static int name_from_path(git_buf *out, git_config *cfg, const char *path)
+static void free_submodule_names(git_strmap *names)
+{
+	git_buf *name = 0;
+	if (0 == names)
```
ethomson: Please don't reverse these tests.  I know that many people have gotten into a habit of this and find it a good practice, but please follow our existing code style.

Please don't use `0` for pointers.

This should be `if (names == NULL)`.

src/submodule.c
```diff
@@ -196,6 +209,17 @@ int git_submodule_lookup(
 
 	assert(repo && name);
 
+	if (0 != repo->submodule_cache) {
```
ethomson: `if (repo->submodule_cache != NULL)`, please.

src/submodule.c
```diff
@@ -1566,6 +1617,33 @@ int git_submodule_location(unsigned int *location, git_submodule *sm)
...
+
+	error = all_submodules(repo, repo->submodule_cache);
+	return error;
+}
+
+int git_repository_submodule_cache_clear(git_repository *repo)
+{
+	git_submodule *sm;
+	assert(repo);
+	if (0 == repo->submodule_cache) {
```
ethomson: `if (repo->submodule_cache == NULL)`...

tests/submodule/lookup.c
```diff
@@ -388,3 +388,28 @@ void test_submodule_lookup__renamed(void)
...
+	git_repository_submodule_cache_all(g_repo);
+	test_submodule_lookup__simple_lookup();
+	git_repository_submodule_cache_clear(g_repo);
+	test_submodule_lookup__simple_lookup();
+
+	/* Check that subsequent calls return different objects when cached. */
+	git_repository_submodule_cache_all(g_repo);
+	cl_assert(0 == git_submodule_lookup(&sm, g_repo, "sm_unchanged"));
+	cl_assert(0 == git_submodule_lookup(&sm2, g_repo, "sm_unchanged"));
+	cl_assert(sm == sm2);
```
ethomson: Using `cl_assert` is not ideal here, since it just will say "an assertion failed".  We have a variety of helper functions that provide more explicit messages, which are helpful in debugging and understanding test failures:

For git functions, please use `cl_git_pass`, which will show the error messages properly when a function fails and returns nonzero.  (eg, `cl_git_pass(git_submodule_lookup...)`)

To compare two pointers, please use `cl_assert_equal_p` which will show more information in a nice format about the actual and expected values.

tests/submodule/lookup.c
```diff
@@ -388,3 +388,28 @@ void test_submodule_lookup__renamed(void)
...
+	git_submodule_free(sm2);
+
+	/* and that we get new objects again after clearing the cache. */
+	git_repository_submodule_cache_clear(g_repo);
+	cl_assert(0 == git_submodule_lookup(&sm2, g_repo, "sm_unchanged"));
```
ethomson: Here, too, this should be `cl_git_pass(...)`.

ethomson:
I have some comments.  Note that I haven't looked at the submodule logic to know if that was just changed whitespace or if there are substantive changes, too.  I'll have to do that before we merge, unless you revert it.  (Mixing whitespace changes with actual code changes is very hard to follow in many diff tools.)

novalis:
I have a separate commit for the whitespace changes.  Are you seeing other whitespace changes outside that commit?

ethomson:
> I have a separate commit for the whitespace changes.

Yep, thanks, that's fine then; I was looking at this in the GitHub web view and didn't realize that.

src/submodule.c
```diff
@@ -1502,13 +1544,22 @@ int git_submodule__status(
...
+	 * expensive operations (especially `submodule_update_head`, which is
+	 * bottlenecked on `git_repository_head_tree`) eliminates much of the
+	 * advantage.  We will, therefore, interpret the request for caching to
+	 * apply here to and skip them.
+	 */
 
-	/* refresh the HEAD OID */
-	if (submodule_update_head(sm) < 0)
-		return -1;
+	if (0 == sm->repo->submodule_cache) {
```
ethomson: Can you also reverse this test?

src/submodule.c
```diff
@@ -416,7 +458,7 @@ typedef struct {
 	git_repository *repo;
 } lfc_data;
 
-static int all_submodules(git_repository *repo, git_strmap *map)
+int all_submodules(git_repository *repo, git_strmap *map)
```
ethomson: Now that this is public (within the library anyway), this should have a name that indicates it's internal to the library.  Something like `git_submodule__all` or `git_submodule__map`, whichever makes more sense to you.

novalis:
OK, I think I've addressed all comments; there are no merge conflicts; the tests are green.

ethomson:
Thanks for doing this @bpeabody and @novalis - and thank you especially for putting up with my laziness and bikeshedding.  😅

bpeabody:
@ethomson: it was no trouble at all; thanks for all your help!

PR #4018, opened 2016-11-25
Various fixes

pks-t:
The title says it all. Just played around with VS2017 RC1 and wanted to see if there's anything interesting in the warnings. Nothing too interesting, I guess. Fixes for multiple uses of uninitialized values is the most interesting outcome.


carlosmn:
The last commit is a duplicate of the one that was merged in #4014, can you remove it? The rest look fine, just need to get the Windows CI to avoid the race.

pks-t:
Rebased on latest master

PR #4020, opened 2016-12-02
git_rebase_init: correctly handle detached HEAD

novalis:
git_rebase_finish relies on head_detached being set, but rebase_init_merge was only setting it when branch->ref_name was unset. But branch->ref_name would be set to "HEAD" in the case of detached
HEAD being either implicitly (NULL) or explicitly passed to git_rebase_init.


ethomson:
Thanks!

PR #4021, opened 2016-12-02
FETCH_HEAD and multiple refspecs

carlosmn:
The functions that write to FETCH_HEAD assume we only handle a single refspec at a time, which means we might fail to write part of the data that should be there, or we'll overwrite one refspec's information with another's.

### DIFF
--- a/src/fetchhead.c
+++ b/src/fetchhead.c
@@ -118,7 +118,7 @@ int git_fetchhead_write(git_repository *repo, git_vector *fetchhead_refs)
 	if (git_buf_joinpath(&path, repo->gitdir, GIT_FETCH_HEAD_FILE) < 0)
 		return -1;
 
-	if (git_filebuf_open(&file, path.ptr, GIT_FILEBUF_FORCE, GIT_REFS_FILE_MODE) < 0) {
+	if (git_filebuf_open(&file, path.ptr, GIT_FILEBUF_APPEND, GIT_REFS_FILE_MODE) < 0) {
 		git_buf_free(&path);
 		return -1;
 	}
--- a/src/remote.c
+++ b/src/remote.c
@@ -1541,6 +1541,20 @@ cleanup:
 	return error;
 }
 
+static int truncate_fetch_head(const char *gitdir)
+{
+	git_buf path = GIT_BUF_INIT;
+	int error;
+
+	if ((error = git_buf_joinpath(&path, gitdir, GIT_FETCH_HEAD_FILE)) < 0)
+		return error;
+
+	error = git_futils_truncate(path.ptr, GIT_REFS_FILE_MODE);
+	git_buf_free(&path);
+
+	return error;
+}
+
 int git_remote_update_tips(
 		git_remote *remote,
 		const git_remote_callbacks *callbacks,
@@ -1570,6 +1584,9 @@ int git_remote_update_tips(
 		tagopt = remote->download_tags;
 	else
 		tagopt = download_tags;
+
+	if ((error = truncate_fetch_head(git_repository_path(remote->repo))) < 0)
+		goto out;
 
 	if (tagopt == GIT_REMOTE_DOWNLOAD_TAGS_ALL) {
 		if ((error = update_tips_for_spec(remote, callbacks, update_fetchhead, tagopt, &tagspec, &refs, reflog_message)) < 0)
--- a/tests/fetchhead/nonetwork.c
+++ b/tests/fetchhead/nonetwork.c
@@ -406,3 +406,62 @@ void test_fetchhead_nonetwork__create_when_refpecs_given(void)
 	git_remote_free(remote);
 	git_buf_free(&path);
 }
+
+static int nrefs_found;
+static bool count_refs_called;
+
+int count_refs(const char *ref_name, const char *remote_url, const git_oid *oid, unsigned int is_merge, void *payload)
+{
+	const char *pfix = (const char *) payload;
+
+	GIT_UNUSED(remote_url);
+	GIT_UNUSED(oid);
+	GIT_UNUSED(is_merge);
+
+	count_refs_called = true;
+	if (!git__prefixcmp(ref_name, pfix))
+		nrefs_found++;
+
+	return 0;
+}
+
+void test_fetchhead_nonetwork__create_with_multiple_refspecs(void)
+{
+	git_remote *remote;
+	git_buf path = GIT_BUF_INIT;
+
+	cl_set_cleanup(&cleanup_repository, "./test1");
+	cl_git_pass(git_repository_init(&g_repo, "./test1", 0));
+
+	cl_git_pass(git_remote_create(&remote, g_repo, "origin", cl_fixture("testrepo.git")));
+	git_remote_free(remote);
+	cl_git_pass(git_remote_add_fetch(g_repo, "origin", "+refs/notes/*:refs/origin/notes/*"));
+	/* Pick up the new refspec */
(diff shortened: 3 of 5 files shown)


carlosmn:
After some detours, this should be ready to go now.

/cc @ethomson @pks-t

tests/fetchhead/nonetwork.c
```diff
@@ -406,3 +406,64 @@ void test_fetchhead_nonetwork__create_when_refpecs_given(void)
 	git_remote_free(remote);
 	git_buf_free(&path);
 }
+
+static int nrefs_found;
+static bool count_refs_called;
```
pks-t: Shouldn't we pass that in via the payload?

tests/fetchhead/nonetwork.c
```diff
@@ -406,3 +406,64 @@ void test_fetchhead_nonetwork__create_when_refpecs_given(void)
...
+	const char *pfix = (const char *) payload;
+
+	GIT_UNUSED(remote_url);
+	GIT_UNUSED(oid);
+	GIT_UNUSED(is_merge);
+
+	count_refs_called = true;
+
+
+	if (!git__prefixcmp(ref_name, pfix))
```
pks-t: Some newlines too much here :)

tests/fetchhead/nonetwork.c
```diff
@@ -406,3 +406,64 @@ void test_fetchhead_nonetwork__create_when_refpecs_given(void)
...
+void test_fetchhead_nonetwork__create_with_multiple_refspecs(void)
+{
+	git_remote *remote;
+	git_buf path = GIT_BUF_INIT;
+
+	cl_set_cleanup(&cleanup_repository, "./test1");
+	cl_git_pass(git_repository_init(&g_repo, "./test1", 0));
+
+	cl_git_pass(git_buf_joinpath(&path, git_repository_path(g_repo), "FETCH_HEAD"));
+	cl_git_pass(git_remote_create(&remote, g_repo, "origin", cl_fixture("testrepo.git")));
```
pks-t: We really should allow to pass a `NULL` remote here

tests/fetchhead/nonetwork.c
```diff
@@ -406,3 +406,64 @@ void test_fetchhead_nonetwork__create_when_refpecs_given(void)
...
+
+void test_fetchhead_nonetwork__create_with_multiple_refspecs(void)
+{
+	git_remote *remote;
+	git_buf path = GIT_BUF_INIT;
+
+	cl_set_cleanup(&cleanup_repository, "./test1");
+	cl_git_pass(git_repository_init(&g_repo, "./test1", 0));
+
+	cl_git_pass(git_buf_joinpath(&path, git_repository_path(g_repo), "FETCH_HEAD"));
```
pks-t: Would be a bit easier if you moved down that line to where it's being used

tests/fetchhead/nonetwork.c
```diff
@@ -406,3 +406,64 @@ void test_fetchhead_nonetwork__create_when_refpecs_given(void)
...
+	/* Pick up the new refspec */
+	cl_git_pass(git_remote_lookup(&remote, g_repo, "origin"));
+
+	cl_assert(!git_path_exists(path.ptr));
+	cl_git_pass(git_remote_fetch(remote, NULL, NULL, NULL));
+	cl_assert(git_path_exists(path.ptr));
+
+	count_refs_called = 0;
+	nrefs_found = 0;
+	cl_git_pass(git_repository_fetchhead_foreach(g_repo, count_refs, "refs/notes/"));
```
pks-t: Instead of calling this function trice, you could just:

```
struct fetchhead_foreach_payload {
    int refs_notes_found;
    int refs_heads_found;
    int refs_tags_found;
}

int count_refs(..., void *payload)
{
    struct fetchhead_foreach_payload *f = (...) payload;

    if (!git__prefixcmp("refs/notes/", pfix))
        f->refs_notes_found++;
   ....
}
```

src/remote.c
```diff
@@ -1541,6 +1541,20 @@ static int opportunistic_updates(const git_remote *remote, const git_remote_call
...
 
+static int truncate_fetch_head(const char *gitdir)
+{
+	git_buf path = GIT_BUF_INIT;
+	int error;
+
+	if ((error = git_buf_joinpath(&path, gitdir, GIT_FETCH_HEAD_FILE)) < 0)
+		return error;
+
+	error = git_futils_truncate(path.ptr, 0666);
```
pks-t: Shouldn't this be `0644`?

src/fetchhead.c
```diff
@@ -118,7 +118,7 @@ int git_fetchhead_write(git_repository *repo, git_vector *fetchhead_refs)
 	if (git_buf_joinpath(&path, repo->gitdir, GIT_FETCH_HEAD_FILE) < 0)
 		return -1;
 
-	if (git_filebuf_open(&file, path.ptr, GIT_FILEBUF_FORCE, GIT_REFS_FILE_MODE) < 0) {
+	if (git_filebuf_open(&file, path.ptr, GIT_FILEBUF_APPEND, GIT_REFS_FILE_MODE) < 0) {
```
pks-t: Without having a look at the surrounding code it feels a bit strange that we reopen the buffer several times to append to it. But this is not a fault of the current PR, so I guess it's fine as-is.

pks-t:
Looks definitly sane from what you're trying to do. Even though the logic that we reopen the FETCHHEAD file several times to append seems strange, that is none of your business.

carlosmn:
>Even though the logic that we reopen the `FETCH_HEAD` file several times to append seems strange

That's what git does and it in fact enables something we would still not have, which is `git remote update` aka `git fetch --all` in which _all_ the fetches from all remotes end up in `FETCH_HEAD`.

src/remote.c
ethomson: No, it should be `0666` and the user's `umask` will apply.  However, we should be using `GIT_REFS_FILE_MODE`.

tests/fetchhead/nonetwork.c
carlosmn: Done.

tests/fetchhead/nonetwork.c
carlosmn: Removed.

src/remote.c
carlosmn: Changed.

tests/fetchhead/nonetwork.c
carlosmn: I didn't want to make this function to tightly coupled with these particular prefixes we want to count right now. I'll see if passing prefix-count tuples in ends up being less verbose.

src/fetchhead.c
carlosmn: It doesn't feel like the cleanest way, but it does let us behave like git (which also reopens the file) without changing a whole lot of functions to propagate a multi-step creation of the file contents.

tests/fetchhead/nonetwork.c
carlosmn: I've just pushed up a new commit which just runs it the one time without coupling it to the specific prefixes we happen to use right now. It doesn't save too much space, but I think it does show the intent more clearly.

ethomson:
Thanks @carlosmn !  ✨

src/fetchhead.c
ethomson: Agreed - this seems a bit odd, but it is intentionally so to be compatible with git's behavior.

PR #4024, opened 2016-12-08
Allow compile with gcc -Werror=format-security

dkelson:
When building a v.25-rc1 RPM on Fedora 25, the default compiler flags include -Werror=format-security and a compile of v.25-rc1 fails in two locations with "format not a string literal and no format arguments".

Please merge this PR.

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -552,7 +552,7 @@ int git_index_clear(git_index *index)
 
 static int create_index_error(int error, const char *msg)
 {
-	giterr_set_str(GITERR_INDEX, msg);
+	giterr_set_str(GITERR_INDEX, "index creation error '%s'", msg);
 	return error;
 }
 
--- a/src/zstream.c
+++ b/src/zstream.c
@@ -21,7 +21,7 @@ static int zstream_seterr(git_zstream *zs)
 	if (zs->zerr == Z_MEM_ERROR)
 		giterr_set_oom();
 	else if (zs->z.msg)
-		giterr_set_str(GITERR_ZLIB, zs->z.msg);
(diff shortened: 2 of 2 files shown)


src/index.c
```diff
@@ -552,7 +552,7 @@ int git_index_clear(git_index *index)
 
 static int create_index_error(int error, const char *msg)
 {
-	giterr_set_str(GITERR_INDEX, msg);
+	giterr_set_str(GITERR_INDEX, "index creation error '%s'", msg);
```
ethomson: You're changing the behavior rather drastically.  The name `create_index_error` indicates that this function _creates_ an _index error_; it doesn't indicate that there was an error _creating the index_.

This will produce errors like:

    index creation error 'Failed to read index: The index is in-memory only'

Which does not make sense.

There are a variety of options for fixing this, simply `giterr_set_str(GITERR_INDEX, "%s", msg);` is certainly the simplest.

carlosmn:
`giterr_set_str` does not indicate to the compiler that it takes format arguments, and neither do any of the functions it calls. Indeed, this is the function you use in order to fix those warnings.

Can you share the warnings your compiler is generating? The format errors should have been fixed in #4009 which IIRC came after the RC.

ethomson:
It feels to me like you were seeing these warnings reported in your build, but [those issues](https://github.com/libgit2/libgit2/commit/65b78ea3013c3e1a5fd70e4dee6e08996cdcaa62) were coincidentally fixed to use `giterr_set_str` before you were able to submit a PR.  Does that sound correct to you?

I'm closing this; please do resubmit a change if you think that this is not yet fixed in master.

Thanks!

Issue #4023, opened 2016-12-07
Assertion failure: e->refcount.val == 0


alexcrichton:
### Reproduction steps

Unfortunately there's not a list of concrete steps for reproduction for this. I'm filing this to track a Cargo bug (https://github.com/rust-lang/cargo/issues/3042) where libgit2 appears to be segfaulting. There's a few layers of indirection here to actually get to libgit2 itself, but I'm hoping that this is perhaps enough information to at least get started debugging! I can always ask for more info from the original source to explore reproduction steps as well if that'd be helpful.

It looks like the bug here though is that when Cargo has a git dependency and attempts to work with it over an NFS mount it appears to cause a segfault. The specific operation here seems to be freeing a repository causes a segfault. I'm unfortunately unsure, though, from the report what actually led up to that point.

cc @asomers, perhaps you can help provide more info here as well?

### Expected behavior

No segfaults

### Actual behavior

Segfaults! A stack trace from a segfault is here: https://github.com/rust-lang/cargo/issues/3042#issuecomment-264871119

### Version of libgit2 (release number or SHA1)

I've asked to clarify what version of Cargo to learn specifically what sha we're using, but if it's a recent-ish Cargo I believe it's very close to c9e967a1b48c7a0b484fd3a4860e9c7a0aa4a319.

### Operating system(s) tested

Looks like the error happens on both FreeBSD and Linux, but it specifically NFS related.

carlosmn:
There's been a few NFS-related issues that have shown up clustered.  As these others are about permissions or read errors, I suspect we're freeing to early or failing to unreference objects when we encounter such a situation.

PR #4027, opened 2016-12-12
pack: dereference cached pack entry on error

pks-t:
When trying to uncompress deltas in a packfile's delta chain, we try to
add object bases to the packfile cache, subsequently decrementing its
reference count if it has been added successfully. This may lead to a
mismatched reference count in the case where we exit the loop early due
to an encountered error.

Fix the issue by decrementing the reference count in error cleanup.

### DIFF
--- a/src/pack.c
+++ b/src/pack.c
@@ -757,8 +757,11 @@ int git_packfile_unpack(
 	}
 
 cleanup:
-	if (error < 0)
+	if (error < 0) {
 		git__free(obj->data);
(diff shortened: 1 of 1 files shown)


pks-t:
Just noticed this possible memory leak/reference count error as I took a quick look at #4023. It _might_ cause the issue in case where we fail to unpack packfiles over NFS, as we would have a mismatched reference count then, triggering the assert. But I think it is rather unlikely.

Issue #3932, opened 2016-09-14
git_remote_fetch downloads all the objects instead of the objects that don't exist


VictorYudin:
Hi there,

I have a bare repository on my local drive. I cloned it with git clone and when I try to update the cloned repository with git_remote_fetch(), it compresses and copies all the objects from the bare repository. Git behaves different, it compresses and copies only objects that don't exist.

You can reproduce it with cgit2:

```
E:\DD\split> cgit2.exe fetch origin
Fetching origin for repo 000000000014CFF0
remote: Counting objects 1926
remote: Compressing objects: 100% (1926/1926), done
[updated] 7be9d1aaa6..3bf8808e34 refs/remotes/origin/master
Received 1926/1926 objects in 169263947bytes
```
But git received only 164 objects, and it took a couple of seconds:
```
E:\DD\split> git fetch origin --verbose
remote: Counting objects: 164, done.
remote: Compressing objects: 100% (140/140), done.
emote: Total 164 (delta 21), reused 161 (delta 21)
Receiving objects: 100% (164/164), 1006.43 KiB | 0 bytes/s, done.
Resolving deltas: 100% (21/21), completed with 1 local objects.
From R:\repotest\split
   7be9d1a..3bf8808  master     -> origin/master
```

Thanks.

xuan-72:
Thanks you!
Well for starters let me say, off to the world? I don't see a problem at
all with creating, If he loses all the faith. After a decline in learning
English not reconciled to

2016-09-15 3:12 GMT+08:00 Victor Yudin <notifications@github.com>:

PR #2972, opened 2015-03-11
[WIP] Smarter pack-building

carlosmn:
This attempts to let the caller describe what objects they'd like to have packed as a walk and make the way we insert objects into the set less wasteful.

The first place to make use of this is local fetching, as it's one of the places where it should have the most effect. The time it takes is still pretty terrible, with 35s vs 41s to perform a clone+checkout of the libgit2 repository. It is however quite a bit less wasteful, at 1_067_758 vs 23_192_189 duplicate insertions into the set and 2s vs 10s of object-insertion time. The rest of those duplicate insertions are likely blobs and thus letting them be inserted multiple times would still faster than diffing (but I still need to test this hypothesis).

PR #4028, opened 2016-12-12
Transfer fewer objects on push and local fetch

chescock:
I have an application that uses LibGit2Sharp, and it's been having performance issues.  One thing I discovered during the investigation is that when doing a fetch using the local transport, it creates a packfile with a copy of the entire repository instead of just the deltas.  I think this is the same issue as #3932.  

The first commit attempts to address that by calling `git_revwalk_hide` on all local references.  

I also tested doing a push in the opposite direction, and saw similar issues where too many objects were sent.  In the second commit, I replaced the complex diffing in the push code with a call to `git_packbuilder_insert_walk`, and it seems to give better results with less code.  Looking through the issue history, I think this may be the change @carlosmn suggested in #2972, "The push code should also be able to take advantage of this instead of the diffing it currently performs".

### DIFF
--- a/src/pack-objects.c
+++ b/src/pack-objects.c
@@ -1641,7 +1641,7 @@ int insert_tree(git_packbuilder *pb, git_tree *tree)
 	if ((error = retrieve_object(&obj, pb, git_tree_id(tree))) < 0)
 		return error;
 
-	if (obj->seen)
+	if (obj->seen || obj->uninteresting)
 		return 0;
 
 	obj->seen = 1;
@@ -1665,6 +1665,10 @@ int insert_tree(git_packbuilder *pb, git_tree *tree)
 
 			break;
 		case GIT_OBJ_BLOB:
+			if ((error = retrieve_object(&obj, pb, git_tree_id(tree))) < 0)
+				return error;
+			if (obj->uninteresting)
+				continue;
 			name = git_tree_entry_name(entry);
 			if ((error = git_packbuilder_insert(pb, entry_id, name)) < 0)
 				return error;
--- a/src/push.c
+++ b/src/push.c
@@ -260,12 +260,11 @@ static int enqueue_tag(git_object **out, git_push *push, git_oid *id)
 	return error;
 }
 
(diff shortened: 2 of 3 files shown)


chescock:
I'm not sure what I need to do to resolve the test failures.  The AppVeyor ones are in `online::badssl::self_signed` and `threads::refdb::edit_while_iterate`, and the Travis one is a SegFault that appears to ocur after `threads::diff..`.  I can't find anything in those tests that calls the code I changed, and I can't reproduce the failures running `libgit2_clar` locally.  What should I do to investigate?  

I notice there are some recent pull requests that mention fixing CI errors.  Should I try rebasing this branch to include those fixes?

ethomson:
>  Should I try rebasing this branch to include those fixes?

If you don't mind, that would be great.  I don't expect that you would have any conflicts, and it will definitely resolve the second AppVeyor error.  (The first AppVeyor error that you mentioned is a transient error, it likely couldn't reach BitBucket.)

chescock:
@ethomson Okay, I rebased.  I got the `online::badssl::self_signed` failure on AppVeyor again, but everything else is green.  It sounds like it shouldn't worry about that failure?

chescock:
Is there anything else I should be doing to make progress on this?

ethomson:
@chescock No - thanks for the fix.  This seems reasonable to me, but I think that this requires expert review from @carlosmn who is master of all things revwalking and packbuilding...

chescock:
@carlosmn Are you going to be able to review this change?  Is there anything I can do to help?

chescock:
@ethomson @carlosmn Is there anything I can do to help make progress on this?  

We've been running with a fork of libgit2 that incorporates this change, and it would be nice to be able to switch back to the regular LibGit2Sharp.NativeBinaries NuGet package.  :)

ethomson:
@carlosmn - as the resident pack expert, it would be great to get your 👀 on this before I :shipit: ?

ethomson:
Thanks @chescock!  😀

PR #4029, opened 2016-12-12
refdb: disable concurrent compress in the threading tests on Windows

carlosmn:
This is far from an ideal situation, but this causes issues on Windows which
make it harder to develop anything, as these tests hit issues which relate
specifically to the Windows filesystem like permission errors for files we
should be able to access. There is an issue likely related to the ordering of
the repack, but there's enough noise that it does not currently help us to run
this aspect of the test in CI.

---

This is basically an approach out of desperation. There is too much that the Windows filesystem is throwing at us for us to resolve this before the release, and these edge cases shouldn't block the release, nor should it block CI for unrelated PRs as much as it does.

### DIFF
--- a/tests/threads/refdb.c
+++ b/tests/threads/refdb.c
@@ -5,6 +5,12 @@
 static git_repository *g_repo;
 static int g_expected = 0;
 
+#ifdef GIT_WIN32
+static bool concurrent_compress = false;
+#else
+static bool concurrent_compress = true;
+#endif
+
 void test_threads_refdb__initialize(void)
 {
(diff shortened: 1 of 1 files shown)


ethomson:
So, this definitely bums me out.  But yes, it's unlikely that we're going to get the underlying issues resolved immediately.  It's not a regression, so it shouldn't hold up the release and it definitely shouldn't be failing our CI builds.

I mention this only because I _also_ want to make sure that we circle back on this after the release on that magical day when we both have lots of spare time, har har.

Issue #3970, opened 2016-10-22
refs::reflog::reflog::append_then_read test fail in TZ GMT-14


russell:
Hey,

`refs::reflog::reflog::append_then_read` fails if in the timezone GMT-14, it seems to be the only timezone that causes the failure.

```
$ TZ=/usr/share/zoneinfo/Etc/GMT-14 ./libgit2_clar -srefs
Loaded 323 suites: 
Started

refs::branches::create.........
refs::branches::delete........
refs::branches::ishead......
refs::branches::iterator.....
refs::branches::lookup....
refs::branches::move...........
refs::branches::name...
refs::branches::remote....
refs::branches::upstream..........
refs::branches::upstreamname..
refs::crashes.
refs::create.........
refs::createwithlog.
refs::delete...
refs::dup..
refs::foreachglob.....
refs::isvalidname..
refs::iterator.......
refs::list..
refs::listall...
refs::lookup.....
refs::normalize......
refs::overwrite....
refs::pack...
refs::peel......
refs::races.....
refs::read...............
refs::reflog::drop.......
refs::reflog::reflogF................
refs::rename............
refs::revparse.................................
refs::settargetwithlog.
refs::setter....
refs::shorthand.
refs::transactions......
refs::unicode.
refs::update.

  1) Failure:
refs::reflog::reflog::append_then_read [/home/russell/projects/libgit2/tests/refs/reflog/reflog.c:21]
  Expression is not true: expected->when.offset == actual->when.offset

```

PR #4033, opened 2016-12-17
Fix off-by-one problems in git_signature__parse

andhe:
Etc/GMT-14 aka UTC+14:00 is a thing....
https://en.wikipedia.org/wiki/UTC%2B14:00

Also allow offsets on the last minute (59).

Addresses: https://bugs.debian.org/841532
Fixes: #3970


carlosmn:
Fair enough.

Issue #3926, opened 2016-09-09
Proper instructions on how to build libgit2 + libssh2 on Windows/via cross-compilation


Happy-Ferret:
Consider this a feature request.

I'd like some solid build instructions (particularly cross compiling libgit2 for Windows on a GNU/Linux machine) dealing with the possible problems arising from attempting to link to ssh2.

As it stands, the whole thing is an infuriating mess that seems to always fail at one point or another (it currently fails attempting to locate libssh2.a for me. Which I've built inside /usr/local/lib).

Windows still being somewhat important to the enterprise, this should have immediate priority.

ethomson:
> this should have immediate priority.

Yes, Windows _is_ important.  Which is why we go out of our way to ensure that Windows - in many configurations - is in our CI matrix.  But cross-compilation is much less so.  I'm sorry that you're having trouble, but this is not in fact an immediate priority for us and I would instead point you to Microsoft's free tools like [Visual Studio 2015 Community Edition](https://www.visualstudio.com/en-us/products/visual-studio-community-vs.aspx).

If you succeed in making this work, I would encourage you to open a pull request documenting it, in case other people are interested in cross-compiling.

Thanks!

tkelman:
We cross compile all the time in Julia. Works fine, you just need to make cmake happy. See libgit2 and libssh2 related makefiles under https://github.com/JuliaLang/julia/blob/master/deps

@ethomson msvc as a compiler doesn't meet a lot of requirements for projects that may want to incorporate libgit2 along with other libraries, and that's unlikely to change. clang fills in most of the things msvc can't do, including cross compiling if you use it against the mingw runtime. But then there's not much difference between using mingw proper and clang-on-mingw.

Happy-Ferret:
@tkelman @ethomson 

Thanks for the heads-up. Followed your advice and incorporated some of the flags you set in your makefiles. Unfortunately, I'm now back at work and forgot copying the resulting libgit2.a from my home computer. 

Following the same steps on Windows ends with an ``No package 'libssh2' found`` both on MinGW and Visual Studio 14 2015.

Would you guys happen to know anything about that? The libssh2 I built lives inside ``C:\Program Files (x86)\libssh2`` (which is strange by itself. Since I built it with TDM-GCC-64, shouldn't it live inside ``Program Files`` proper rather than ``x86``?)

Happy-Ferret:
Never mind. Had to clean up my build environment a little.

[...]

PR #4036, opened 2016-12-18
Load libssh2 at runtime

carlosmn:
This is a continuation of #3785 which takes it further by removing the attempt to detect and link against libssh2 at build time. Once we go run-time we should own that.

There's an aspect which is not yet quite complete, namely which SOVERSIONs we should look up. The Linux version should probably go like the Darwin one and use simply "1" but we'll see what CI says.

I've tested that loading the library works when placed alongside the clar binary, but we still need to figure out what to do for testing on Windows.


src/transports/ssh.c
```diff
@@ -893,17 +978,101 @@ int git_transport_ssh_with_paths(git_transport **out, git_remote *owner, void *p
...
+};
+
+/**
+ * Try to load a supported version of libssh2
+ */
+static void load_libssh2(void)
+{
+#ifdef GIT_WIN32
+	/* Windows doesn't do SOVERSIONs but you ship your own library so we're probably OK */
+	git_libssh2_handle = LoadLibrary("libssh2.dll");
```
tkelman: some environments (msys2 probably, will have to check) approximate sonames via libssh2-#.dll

carlosmn:
This opens the question of what the SSH capability means. Does it mean we'd support ssh if we could find libssh2 or does it mean we've found it and can do ssh?

tiennou:
From the top of my head, I'd prefer that it meant we've found it and can do ssh. IMHO the other case is meaningful only when we allow the application to provide its own libssh2.

Alternatively, split the `GIT_FEATURE_SSH` into `GIT_FEATURE_SSH_ENABLED`/`GIT_FEATURE_SSH_AVAILABLE` ?

carlosmn:
>From the top of my head, I'd prefer that it meant we've found it and can do ssh

Yes, that make sense, but we've been using this as a way to see the compile-time options so it'd be a change in function. It might end up being the better solution, though.

I've tried to get this working on Appveyor but can't reproduce the success from my VM. I've uploaded the same dll I'm using, but neither the 32-bit or 64-bit versions seem to want to load them.

ethomson:
> I've tried to get this working on Appveyor but can't reproduce the success from my VM. I've uploaded the same dll I'm using, but neither the 32-bit or 64-bit versions seem to want to load them.

Naive guess:  is libssh2 built dynamically with a different version of the C runtime?  Can you load it (on your machine) with dependency walker?

Are you building libssh2 with `/MD`?  Can you instead build with `/MT`?

carlosmn:
💃  once I built them statically (so they don't depend at runtime on a newer VS), we're able to find `libssh2.dll`.

The mingw build is still failing because apparently they don't have the unix library loading functionality headers; hopefully we can just use the Win32 ones.

carlosmn:
This is now passing al-round, so "all" that's left is to make a comprehensible commit history and adjust the documentation to mention the DLL loading.

jacquesg:
Does this mean that we can longer link libssh2 statically?

carlosmn:
With the code as-is, it wouldn't support a statically-linked libssh2 but you can load symbols in the same object so it could with small changes.

jacquesg:
That would be great as it is often more convenient to statically link dependencies if their licenses allow for it.

src/transports/ssh.c
```diff
@@ -15,10 +19,75 @@
...
+
+#define LOOKUP_LIBSSH2_SYMBOL(symbol) \
+	git_ ## symbol = (void *) GetProcAddress(git_libssh2_handle, #symbol)
+
+#else
+
+static void *git_libssh2_handle;
+
+#define LOOKUP_LIBSSH2_SYMBOL(symbol) \
+	git_ ## symbol = dlsym(git_libssh2_handle, #symbol)
```
ethomson: Need error checking here.

src/transports/ssh.c
```diff
@@ -15,10 +19,75 @@
...
+
+/* These macros allow us to refer to libssh2 symbols while
+ * supporting dynamic linking.
+ */
+#ifdef GIT_SSH_LOADLIBRARY
+
+static HMODULE git_libssh2_handle;
+
+#define LOOKUP_LIBSSH2_SYMBOL(symbol) \
+	git_ ## symbol = (void *) GetProcAddress(git_libssh2_handle, #symbol)
```
ethomson: Need error checking here.

ethomson:
Yeah, I agree with @jacquesg here - it would be nice if we could allow for static linking and then just `#define git_libssh2_whatever libssh2_whatever`.  I actually think that static linking should be the general case for ssh support and this dynamic loading should be explicitly opt-in.

carlosmn:
I've added the error handling for not finding the required functions. I've also re-enabled finding the library at runtime, though I'm not sure that adds a lot considering `-lcurl` nowadays implies `-lssh2` as well (that was fun to try to piece together).

So we now have `-DUSE_SSH=ON` and `-DUSE_SSH_RUNTIME=ON`, the latter of which will enable runtime detection. In order to keep code similar, we still load them with `dlsym` at runtime, but look at the current image instead of trying to load a library. I haven't tested that one out too much yet, but the manpage leads me to believe that it should work in that situation.

carlosmn:
And of course `RTLD_DEFAULT` is actually `NULL` in many platforms, so we can't distinguish those values.

carlosmn:
I'm hitting a wall trying to get the same codebase to target all three forms of linking. The code here seems to be unable to properly load symbols which are already available and is making unrelated code crash.

I've spent all day today trying to approach this via weak linking, which was my hope for a clean implementation with overridable functions, but glibc's loader is trying to be efficient and won't let libssh2 override the weak symbols in `libgit2.so` or `libgit2_clar`, so that's not going to go anywhere.

seltzy:
Why does this remove the `EMBED_SSH_PATH` option? It seemed like a more convenient option, especially on toolchains/systems that don't include `pkg-config`. In particular, this seems like it would make cross-compilation more difficult. I already have enough trouble using MinGW on Linux to target Windows as also noted in #3926 - in my case it ends up using the `PC` definitions to get `gcrypt` and `libssh2` even with `EMBED_SSH_PATH` which subsequently causes builds to fail at link-time since there's no shared libssh2 built with the same toolchain. Will runtime loading mitigate this, or make things more confusing?

In my case, I have a project where I mostly build my dependencies from source as this yields more consistent building behaviour across target platforms and has helped reduced maintenance for me. Embedding libssh2 is quite nice for this, but perhaps it's not The Right Way™. This concern could be fueled by inexperience or lack of understanding, but I would like to better understand the implications of this for Windows builds and also for cross-compiling.

carlosmn:
After some further discussion elsewhere, I've come to the conclusion that this isn't a worthwhile time investment.

The primary driver for this was to allow libssh2 to be picked up dynamically on Windows while keeping the "normal" dynamic linking in the other platforms, where the normal course of business lets us detect and support libssh2.

But in order for this to work properly, we'd still have to provide scripts or instructions for people who are not used to the C world on how to build it. At that point, it'd be about as much effort on that side, and significantly less on this side, to allow them to build libgit2 with a built-in libssh2 (though redistribution would still be an unsolved question).

PR #4037, opened 2016-12-19
README: be more explicit in the goals and scope

carlosmn:
Make it clearer from the get-go that we do not aim to implement
user-facing commands from the git tool.

---

I'd been meaning to add some wording about this for a while.


pks-t:
Looks good to me :+1:

PR #4038, opened 2016-12-19
Gift deprecated in favor of SwiftGit2

lucasderraugh:
According to the Gift repo it is now deprecated in favor of SwiftGit2.
> This library isn't really maintained anymore. You should use SwiftGit2 instead.

So changing the link for that here.


Uncommon:
Too bad, Gift was a better name ;)

PR #4040, opened 2016-12-20
Merge example

tiennou:
This adds an `merge` example to the party. I don't really like the interface there is between `main/analyze_merge`, but that's what you get when you code on the fly against something you have no idea how its used.

- I might add a `-Sours/theirs` flag, to demonstrate conflict handling and resolution.
- One minor pain point I had is with `git_merge` using `git_annotated_commit`s, then having to go back and grab those underlying `git_commit`s for doing the final merge commit.
- The commit part feels like a hassle to do, and I would love for a `git_merge_finalize()` function that takes care of the good, non-conflicting case.


tiennou:
If someone is in Travis' vicinity, would you be so kind as to nudge it on my behalf ?

This should be green anyway, I just want to
- discuss the last couple points I raised above
- make sure someone double-check that I'm not doing something stupid in a future "example".

But this is not high-priority in any way. Just one less open PR 😜 .

examples/merge.c
```diff
@@ -0,0 +1,355 @@
...
+	checkout_opts.checkout_strategy = GIT_CHECKOUT_FORCE|GIT_CHECKOUT_ALLOW_CONFLICTS;
+
+	err = git_merge_analysis(&analysis, &preference,
+							 repo,
+							 (const git_annotated_commit **)opts->annotated,
+							 opts->annotated_count);
+	check_lg2(err, "merge analysis failed", NULL);
+
+	if (analysis & GIT_MERGE_ANALYSIS_UP_TO_DATE) {
+		printf("Already up-to-date");
```
richardipsum: Missing newline?

examples/network/git2.c
```diff
@@ -57,16 +49,42 @@ int main(int argc, char **argv)
...
+	struct args_info args = ARGS_INFO_INIT;
+	const char *git_dir = NULL;
+
+	for (args.pos = 1; args.pos < args.argc; ++args.pos) {
+		char *a = args.argv[args.pos];
+
+		if (a[0] != '-') {
+			/* non-arg */
+			break;
+		} else if (optional_str_arg(&git_dir, &args, "--git-dir", ".git")) {
```
richardipsum: Is the empty body for this if statement also left intentionally empty?

examples/merge.c
```diff
@@ -0,0 +1,355 @@
...
+		/* Maybe that's a ref, so DWIM it */
+		git_reference_dwim(&merge_ref, repo, opts.heads[0]);
+
+		/* Grab a signature */
+		check_lg2(git_signature_now(&sign, "Me", "me@example.com"), "failed to create signature", NULL);
+
+		/* Prepare a standard merge commit message */
+		if (merge_ref != NULL) {
+			check_lg2(git_branch_name(&merge_branch_name, merge_ref), "failed to get branch name of merged ref", NULL);
+			err = asprintf(&msg, "Merge branch '%s'", merge_branch_name);
```
richardipsum: asprintf is a GNU extension, so probably needs to be dropped for portability

examples/merge.c
```diff
@@ -0,0 +1,355 @@
...
+		} else {
+			err = asprintf(&msg, "Merge commit '%s'", git_oid_tostr_s(git_annotated_commit_id(merge_commit)));
```
richardipsum: same here (asprintf)

examples/merge.c
```diff
@@ -0,0 +1,355 @@
+/*
+ * libgit2 "describe" example - shows how to describe commits
```
richardipsum: this should be updated for "merge" example

examples/merge.c
```diff
@@ -0,0 +1,355 @@
...
+ *
+ * To the extent possible under law, the author(s) have dedicated all copyright
+ * and related and neighboring rights to this software to the public domain
+ * worldwide. This software is distributed without any warranty.
+ *
+ * You should have received a copy of the CC0 Public Domain Dedication along
+ * with this software. If not, see
+ * <http://creativecommons.org/publicdomain/zero/1.0/>.
+ */
+
```
richardipsum: Other examples seem to have a comment at the top of the file explaining what the example does exactly, and how it should be used, I think this would be really useful for this example.

examples/merge.c
```diff
@@ -0,0 +1,355 @@
...
+		print_usage();
+
+	for (args.pos = 1; args.pos < argc; ++args.pos) {
+		const char *curr = argv[args.pos];
+
+		if (curr[0] != '-') {
+			opts_add_refish(opts, curr);
+		} else if (!strcmp(curr, "--no-commit")) {
+			opts->no_commit = 1;
+		} else if (match_str_arg(repo_path, &args, "--git-dir")) {
```
tiennou: Yes. `match_str_arg` will perform the copy of the value after "--git-dir", so there's nothing else to do. The same goes for `optional_str_arg` below, even though I'm now wondering why this one isn't optional...

I think it's a pretty common idiom in the example code, since they share their option/argument wrangling.

examples/merge.c
```diff
@@ -0,0 +1,355 @@
...
+		git_repository_state_cleanup(repo);
+	}
+cleanup:
+	free(opts.heads);
+	free(opts.annotated);
+	git_repository_free(repo);
+	git_libgit2_shutdown();
+
+	return 0;
+}
```
tiennou: Missing newline !

examples/merge.c
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+	/* force-checkout the result. */
+	err = git_checkout_head(repo, &ff_checkout_options);
+	if (err != 0) {
+		fprintf(stderr, "failed to checkout HEAD reference\n");
+		return -1;
+	}
+	return 0;
+}
```
richardipsum: I personally am not certain this is the right way to do a fast forward merge. `git_merge` can be used for this and by the end of the merge the index is set appropriately, all that remains is to update HEAD to the head of the branch you're merging. It seems you can also pass options to `git_merge` to ensure that it is fast forward only.

The other thing about this is that checkout leaves the working directory in the wrong state since it won't modify files by default.

richardipsum: To be honest, for an example I'm not convinced a version that performs a merge commit regardless of fast forward would be such a disaster either. @ethomson do you have any thoughts on this?

richardipsum: @pks-t any thoughts?

pks-t: As long as we don't pretend to allow fast-forward merges (e.g. accepting the "--ff-only" flag, which I bet isn't part of this PR anyway) I don't mind it. The examples are probably not used by anybody anyway and only aim to give examples of how to use our API. Handling all the intricacies of upstream git tools does not work towards that goal and can even overcomplicate examples, working straight against the desire to demonstrate our API.

Uncommon:
Suggestion: Include a workaround for #4203 until it's fixed.

tiennou:
Thanks for the reviews @richardipsum, I've rebased it and fixed most of those, except `asnprintf` (which I prefer to pointer-wrangling, since it's example code), and the `--ff-only` discussion (since I don't feel that's in-scope, and the code should perform a fast-forward anyway if that's possible).

pks-t:
Unfortunately, you'll have to fix up the `asnprintf` thing though, as that is not part of POSIX and not supported on at least Windows platforms.

examples/merge.c
tiennou: Well, that exact bug came up while testing 🤣. Since I'm trying to "stick to the letter", this will attempt a fast-forward if possible, and will end up with a nonsensical index (IOW the fast-forwarded commit has its reverse staged).
I'll happily fix that, I'm just obviously unable to find the correct steps (I was going by gut-feeling here : move the branch, then force-checkout).

pks-t:
Any feedback you require for that or do you simply need some more time to get that done? Just checking whether you expect me to do anything right now :)

examples/merge.c
ethomson: This looks like the correct way to do a fast-forward.  I would correct the comment - we're not _force_ checking out, just a regular checkout.

tiennou:
Fast-forward fixed. I also checked that the unborn case is correct with some manual `rm`-ing.

examples/network/git2.c
```diff
@@ -57,16 +49,42 @@ int main(int argc, char **argv)
 
 	git_libgit2_init();
 
+	struct args_info args = ARGS_INFO_INIT;
+	const char *git_dir = NULL;
```
pks-t: Declarations after statements are disallowed in C90. Granted, the file here isn't conforming to the C90 standard right now, but that should be no reason to make the situation worse :P

examples/network/git2.c
pks-t: Yeah, a `continue` would be nice to have here

examples/network/git2.c
```diff
@@ -57,16 +49,42 @@ int main(int argc, char **argv)
...
+			break;
+		} else if (optional_str_arg(&git_dir, &args, "--git-dir", ".git")) {
+		} else if (!strcmp(a, "--")) {
+			/* arg separator */
+			break;
+		}
+	}
+
+	// Before running the actual command, create an instance of the local
+	// repository and pass it to the function.
```
pks-t: Please use `/* */`-style comments

examples/network/git2.c
```diff
@@ -57,16 +49,42 @@ int main(int argc, char **argv)
...
+		if (!strcmp(args.argv[args.pos], commands[i].name)) {
+			return_code = run_command(commands[i].fn, repo, args);
 			goto shutdown;
 		}
 	}
 
 	fprintf(stderr, "Command not found: %s\n", argv[1]);
 
 shutdown:
+	if(repo)
```
pks-t: This check is not required. `git_repository_free` already returns if `repo == NULL`.

examples/merge.c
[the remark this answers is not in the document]
pks-t: An explicit `continue` might be avoid confusion, though.

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+
+	err = git_oid_fromstr(&oid, refish);
+	if (err == GIT_OK) {
+		err = git_annotated_commit_lookup(commit, repo, &oid);
+	}
+
+	return err;
+}
+
+static int resolve_heads(git_repository *repo, merge_options *opts) {
```
pks-t: Opening function braces belong on the next line

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+	git_annotated_commit **annotated = calloc(opts->heads_count, sizeof(git_annotated_commit *));
+	size_t annotated_count = 0, i;
+	int err = 0;
+
+	for (i = 0; i < opts->heads_count; i++) {
+		err = resolve_refish((git_annotated_commit **)&annotated[annotated_count++], repo, opts->heads[i]);
```
pks-t: Is this cast really necessary? Shouldn't be, I think

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+		if (err != 0) {
+			fprintf(stderr, "failed to resolve refish %s (%d)\n", opts->heads[i], err);
```
pks-t: Maybe print out the error message if available? `giterr_last()->message`

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+			annotated_count--;
+			continue;
+		}
+	}
+
+	if (annotated_count <= 0) {
+		fprintf(stderr, "unable to parse any refish\n");
+		free(annotated);
+		return -1;
```
pks-t: Do I interpret it correctly that you just proceed with less than the specified count of commits? If so, you should instead abort completely if any of the commits would not be parsed.

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+		err = git_reference_lookup(&head_ref, repo, "HEAD");
+		if (err != 0) {
+			fprintf(stderr, "failed to lookup HEAD ref\n");
+			return -1;
+		}
+
+		/* Grab the reference HEAD should be pointing to */
+		symbolic_ref = git_reference_symbolic_target(head_ref);
+
+		/* Force-create that reference on the target OID */
```
pks-t: We're not forcing anything, as the target doesn't yet exist it's just a simple reference creation

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+		/* HEAD exists, just lookup and resolve */
+		err = git_repository_head(&target_ref, repo);
+		if (err != 0) {
+			fprintf(stderr, "failed to get HEAD reference\n");
+			return -1;
+		}
+	}
+
+	/* Lookup the target object */
+	err = git_object_lookup(&target, repo, target_oid, GIT_OBJ_COMMIT);
```
pks-t: Shouldn't this happen before we create the target reference in the case of an unborn branch? Otherwise, we write a reference whose OID may not even exist

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+	if (err != 0) {
+		fprintf(stderr, "failed to lookup OID %s\n", git_oid_tostr_s(target_oid));
+		return -1;
+	}
+
+	/* Force-checkout the result so the workdir is in the expected state */
+	ff_checkout_options.checkout_strategy = GIT_CHECKOUT_FORCE;
```
pks-t: Do we really want a force checkout? In case there's unstaged changes, we certainly wouldn't want to discard those in case of a fast-forward merge. Git definitly makes sure to keep your data in that scenario

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+
+static int analyze_merge(git_repository *repo, merge_options *opts)
+{
+	git_merge_options merge_opts = GIT_MERGE_OPTIONS_INIT;
+	git_checkout_options checkout_opts = GIT_CHECKOUT_OPTIONS_INIT;
+	git_merge_analysis_t analysis;
+	git_merge_preference_t preference;
+	int err = 0;
+
+	merge_opts.flags = 0; // GIT_MERGE_FIND_RENAMES;
```
pks-t: C99-style comments are not allowed

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+		       (analysis & GIT_MERGE_ANALYSIS_FASTFORWARD &&
+		       !(preference & GIT_MERGE_PREFERENCE_NO_FASTFORWARD))) {
+		const git_oid *target_oid;
+		if (analysis & GIT_MERGE_ANALYSIS_UNBORN) {
+			printf("Unborn\n");
+		} else {
+			printf("Fast-forward\n");
+		}
+
+		/* Since this is a fast-forward, there can be only one merge head */
```
pks-t: s/can/should. You should probably make sure that's actually the case.

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+	}
+
+	git_reference_free(target_ref);
+	git_reference_free(new_target_ref);
+	git_object_free(target);
+
+	return 0;
+}
+
+static int analyze_merge(git_repository *repo, merge_options *opts)
```
pks-t: The function is a bit misnamed. You're not only analyzing the merge, but also actually performing it.

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+		git_index_conflict_iterator *conflicts;
+		const git_index_entry *ancestor;
+		const git_index_entry *our;
+		const git_index_entry *their;
+
+		check_lg2(git_index_conflict_iterator_new(&conflicts, index), "failed to create conflict iterator", NULL);
+
+		while ((err = git_index_conflict_next(&ancestor, &our, &their, conflicts)) == 0) {
+			fprintf(stderr, "conflict: a:%s o:%s t:%s\n",
+			        ancestor ? ancestor->path : "", our->path, their->path);
```
pks-t: `our` and `their` can also be `NULL` in case of add-remove, modify-remove or similar merge conflicts.

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+		}
+
+		if (err != GIT_ITEROVER) {
+			fprintf(stderr, "error iterating conflicts\n");
+		}
+
+		git_index_conflict_iterator_free(conflicts);
+	} else if (!opts.no_commit) {
```
pks-t: Maybe move the following block into its own function?

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+		const char *msg_target = NULL;
+		size_t msglen = 0;
+		char *msg;
+		size_t i;
+		int err;
+
+		/* Grab our needed references */
+		check_lg2(git_repository_head(&head_ref, repo), "failed to get repo HEAD", NULL);
+		if (resolve_refish(&merge_commit, repo, opts.heads[0])) {
+			fprintf(stderr, "failed to resolve refish %s", opts.heads[0]);
```
pks-t: Shouldn't you error out in that case?

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+		}
+
+		/* Maybe that's a ref, so DWIM it */
+		git_reference_dwim(&merge_ref, repo, opts.heads[0]);
```
pks-t: You should check for (unexpected) errors here

examples/merge.c
```diff
@@ -0,0 +1,387 @@
...
+		err = git_commit_create(&commit_oid,
+		                        repo, git_reference_name(head_ref),
+		                        sign, sign,
+		                        NULL, msg,
+		                        tree,
+		                        opts.annotated_count + 1, (const git_commit **)parents);
+		check_lg2(err, "failed to create commit", NULL);
+
+		/* We're done merging, cleanup the repository state */
+		git_repository_state_cleanup(repo);
```
pks-t: You're leaking `parents` here

examples/merge.c
tiennou: Weeeell, I'd be surprised. From what I can tell "unborn" mostly means "someone just cloned a repo, and they're now somehow merging their origin/master into (unborn) master" (I had to resort to `rm` to check for unborn-ness).

In the current code, that OID comes from the list of annotated commit, which we (indirectly) looked up at the beginning. I can clarify that though, if you feel that's warranted.

examples/merge.c
tiennou: Technically, the merge analysis would not have said that in that case (that would be an octopus merge). I made it an assert, just in case. *Unless* you can octopus-merge unborn branches, maybe 😉 ?

tiennou:
@pks-t You were right on time 😉. Thanks for the review, as ever ! I took care of most of your comments.

Looks like AppVeyor doesn't like snprintf though…

examples/merge.c
pks-t: Fair enough. Just checked `git_merge_analysis`, and the first thing we do is make sure that `their_heads_len == 1`.

pks-t: Huh. Doesn't this even imply that we are unable to do octopus merges at all? I honestly don't know right now. In that case, the code might even be greatly simplified

examples/merge.c
pks-t: True. It's quite far away that we load and thus verify the target OID, so this wasn't immediately obvious to me.

examples/merge.c
```diff
@@ -0,0 +1,389 @@
...
+	git_annotated_commit **annotated = calloc(opts->heads_count, sizeof(git_annotated_commit *));
+	size_t annotated_count = 0, i;
+	int err = 0;
+
+	for (i = 0; i < opts->heads_count; i++) {
+		err = resolve_refish(&annotated[annotated_count++], repo, opts->heads[i]);
+		if (err != 0) {
+			fprintf(stderr, "failed to resolve refish %s: %s\n", opts->heads[i], giterr_last()->message);
+			annotated_count--;
+			continue;
```
pks-t: I still think we should actually abort here in case anything bad happened. Furthermore, regarding that I think we cannot even do octopus merges: do we even have to resolve multiple heads when we cannot handle them at all?

examples/merge.c
```diff
@@ -0,0 +1,389 @@
...
+
+		checkout_opts.checkout_strategy = GIT_CHECKOUT_FORCE|GIT_CHECKOUT_ALLOW_CONFLICTS;
+
+		if (preference & GIT_MERGE_PREFERENCE_FASTFORWARD_ONLY) {
+			printf("Fast-forward is preferred, but only a merge is possible\n");
+			return -1;
+		}
+
+		err = git_merge(repo,
+						(const git_annotated_commit **)opts.annotated, opts.annotated_count,
```
tiennou: Xcode indent.

examples/merge.c
tiennou: [Yep](https://github.com/libgit2/libgit2/blob/master/src/merge.c#L3264). I'm just aiming for completeness though and mirroring the `git-merge` tool.

examples/merge.c
tiennou: We should be aborting just below (we parsed less than `heads_count` refishs). The goal is to do all parsing upfront so we report all unparseable arguments, and exit otherwise.

pks-t: Ugh. Yeah. :sweat:

tiennou:
Fixed usage of snprintf with a `#define`, and corrected indentation.

tiennou:
Well, that assert should read `== 1` 🙄 .

tiennou:
Fixed and rebased. I've taken care of all 4 points above, so let's wait to see what CI thinks.

pks-t:
Just did some quick testing, this looks good to me. As this is not in any way changing the libgit2 library, I'm merging this even though we're actually preparing a release right now.

pks-t:
Thanks a lot for keeping on top of this!

ethomson:
Thanks @tiennou for the hard work of building this, and thanks @pks-t for the hard work of reviewing it!  🎉

Issue #2716, opened 2014-11-19
Handling of return codes in user callbacks is fragile


jamill:
The library guidelines state:

```
As is typical with most C library functions, a zero value indicates success and a negative value indicates failure.
```

However, the conflicts a bit with another pattern / convention around the handling of return codes from user callbacks where:
1. Many user callbacks interpret any non-zero value to stop further processing (e.g. [checkout_tree](https://github.com/libgit2/libgit2/blob/091165c53b2bcd5d41fb71d43ed5a23a3d96bf5d/include/git2/checkout.h#L320), [git_push_status_foreach](https://github.com/libgit2/libgit2/blob/091165c53b2bcd5d41fb71d43ed5a23a3d96bf5d/include/git2/push.h#L168)
2. Many functions propagate the user provided return value back up through the calling function.

This leads to a state where functions that might call a user callback can return positive return codes, which is a bit ambiguous in meaning. This is also fragile / unexpected as consumers have to be aware of whether a function might return a positive return code and different functions have different semantics depending on whether they have any user callbacks. Even library code has instances that does not properly handle functions that might return positive return codes (e.g. [checkout_tree](https://github.com/libgit2/libgit2/blob/b07db1eb7696c1ef71f122c2103780c7bb55ed47/src/reset.c#L158) )

Potential suggestion:
Functions should always use negative values to indicate failure. If a user callback returns a positive number, convert it to `GIT_EUSER`

PR #4042, opened 2016-12-20
Allow the certificate check to be ignored

### LINKED ISSUE #3440: The certificate callback should support pass-through

tiennou:
Fixes #3440

### DIFF
--- a/src/transports/smart.c
+++ b/src/transports/smart.c
@@ -460,11 +460,28 @@ static int ref_name_cmp(const void *a, const void *b)
 	return strcmp(ref_a->head.name, ref_b->head.name);
 }
 
-int git_transport_smart_certificate_check(git_transport *transport, git_cert *cert, int valid, const char *hostname)
+int git_transport_smart_certificate_check(git_transport *transport, git_cert *cert, int cert_error, const char *hostname)
 {
 	transport_smart *t = (transport_smart *)transport;
+	git_error_state state = {0};
+	int cberror, error;
+	int is_valid = (cert_error == GIT_OK);
 
-	return t->certificate_check_cb(cert, valid, hostname, t->message_cb_payload);
+	if (t->certificate_check_cb == NULL)
+		return cert_error;
+
+	giterr_state_capture(&state, cert_error);
+	cberror = t->certificate_check_cb(cert, is_valid, hostname, t->message_cb_payload);
+	if (cberror == GIT_PASSTHROUGH) {
+		error = giterr_state_restore(&state);
+	} else {
+		error = cberror;
+		if (error < 0 && !giterr_last())
+			giterr_set(GITERR_NET, "user cancelled certificate check");
+	}
+	giterr_state_free(&state);
+
+	return error;
 }
 
 int git_transport_smart_credentials(git_cred **out, git_transport *transport, const char *user, int methods)
--- a/src/transports/winhttp.c
+++ b/src/transports/winhttp.c
@@ -266,7 +266,6 @@ static int certificate_check(winhttp_stream *s, int valid)
 	PCERT_CONTEXT cert_ctx;
 	DWORD cert_ctx_size = sizeof(cert_ctx);
 	git_cert_x509 cert;
-	git_error_state error_state = {0};
 
 	/* If there is no override, we should fail if WinHTTP doesn't think it's fine */
 	if (t->owner->certificate_check_cb == NULL && !valid) {
@@ -284,20 +283,13 @@ static int certificate_check(winhttp_stream *s, int valid)
 		return -1;
 	}
 
-	giterr_state_capture(&error_state, GIT_ERROR);
 	cert.parent.cert_type = GIT_CERT_X509;
 	cert.data = cert_ctx->pbCertEncoded;
 	cert.len = cert_ctx->cbCertEncoded;
-	error = t->owner->certificate_check_cb((git_cert *) &cert, valid, t->connection_data.host, t->owner->message_cb_payload);
+	/* We have to reconstruct the error here */
+	error = (valid ? GIT_OK : GIT_ECERTIFICATE);
+	error = git_transport_smart_certificate_check(&t->owner->parent, (git_cert *) &cert, error, t->connection_data.host);
 	CertFreeCertificateContext(cert_ctx);
-
-	if (error == GIT_PASSTHROUGH) {
-		/* Nothing could have failed previously, return success */
-		error = giterr_state_restore(&error_state);
-	} else if (error < 0 && !giterr_last()) {
-		giterr_set(GITERR_NET, "user cancelled certificate check");
-	}
-	giterr_state_free(&error_state);
 
 	return error;
 }
--- a/src/transports/winhttp.c
+++ b/src/transports/winhttp.c
@@ -266,6 +266,7 @@ static int certificate_check(winhttp_stream *s, int valid)
 	PCERT_CONTEXT cert_ctx;
 	DWORD cert_ctx_size = sizeof(cert_ctx);
 	git_cert_x509 cert;
+	git_error_state error_state = {0};
 
 	/* If there is no override, we should fail if WinHTTP doesn't think it's fine */
 	if (t->owner->certificate_check_cb == NULL && !valid) {
@@ -283,15 +284,20 @@ static int certificate_check(winhttp_stream *s, int valid)
 		return -1;
 	}
 
-	giterr_clear();
+	giterr_state_capture(&error_state, GIT_ERROR);
 	cert.parent.cert_type = GIT_CERT_X509;
 	cert.data = cert_ctx->pbCertEncoded;
 	cert.len = cert_ctx->cbCertEncoded;
 	error = t->owner->certificate_check_cb((git_cert *) &cert, valid, t->connection_data.host, t->owner->message_cb_payload);
 	CertFreeCertificateContext(cert_ctx);
 
-	if (error < 0 && !giterr_last())
+	if (error == GIT_PASSTHROUGH) {
+		/* Nothing could have failed previously, return success */
+		error = giterr_state_restore(&error_state);
+	} else if (error < 0 && !giterr_last()) {
 		giterr_set(GITERR_NET, "user cancelled certificate check");
+	}
+	giterr_state_free(&error_state);
 
 	return error;
 }
@@ -880,11 +886,7 @@ static int send_request(winhttp_stream *s, size_t len, int ignore_length)
 		cert_valid = 0;
 	}
 
-	giterr_clear();
 	if ((error = certificate_check(s, cert_valid)) < 0) {
-		if (!giterr_last())
-			giterr_set(GITERR_OS, "user cancelled certificate check");
-
 		return error;
 	}
 
--- a/tests/online/cert.c
+++ b/tests/online/cert.c
@@ -1,0 +1,141 @@
+#include "clar_libgit2.h"
+
+#include "git2/clone.h"
+
+static git_repository *g_repo;
+
+#ifdef GIT_HTTPS
+static bool g_has_ssl = true;
+#else
+static bool g_has_ssl = false;
+#endif
+
+#if defined(GIT_SSH)
+static bool g_has_ssh = true;
+#else
+static bool g_has_ssh = false;
+#endif
+
+struct cert_options {
+	int expect_valid;
+	int had_callback;
+	int return_code;
+	char *expect_host;
+};
+
+static git_clone_options g_options;
+struct cert_options g_cert_options;
+
+void test_online_cert__initialize(void)
+{
+	git_checkout_options dummy_checkout = GIT_CHECKOUT_OPTIONS_INIT;
+	git_fetch_options dummy_fetch = GIT_FETCH_OPTIONS_INIT;
+
+	g_repo = NULL;
+
+	memset(&g_options, 0, sizeof(git_clone_options));
+	g_options.version = GIT_CLONE_OPTIONS_VERSION;
+	g_options.checkout_opts = dummy_checkout;
+	g_options.checkout_opts.checkout_strategy = GIT_CHECKOUT_SAFE;
+	g_options.fetch_opts = dummy_fetch;
+
+	memset(&g_cert_options, 0, sizeof(struct cert_options));
+	g_options.fetch_opts.callbacks.payload = &g_cert_options;
+}
+
+void test_online_cert__cleanup(void)
+{
+	if (g_repo) {
+		git_repository_free(g_repo);
+		g_repo = NULL;
+	}
+	cl_fixture_cleanup("./fake");
+}
+
+static int certificate_check_cb(git_cert *cert, int valid, const char *host, void *payload)
+{
+	struct cert_options *options = payload;
+
+	GIT_UNUSED(cert);
+
+	if (options) {
+		options->had_callback = 1;
+		cl_assert_equal_i(valid, options->expect_valid);
+		if (options->expect_host) {
+			cl_assert_equal_s(host, options->expect_host);
+		}
+	}
+
+	return (options ? options->return_code : GIT_ERROR);
+}
+
+void test_online_cert__passthrough_https(void)
+{
+	if (!g_has_ssl)
+		cl_skip();
+
+	g_options.fetch_opts.callbacks.certificate_check = certificate_check_cb;
+
+	g_cert_options.expect_valid = 1;
+	g_cert_options.return_code = GIT_PASSTHROUGH;
+
+	cl_git_pass(git_clone(&g_repo, "https://github.com/libgit2/TestGitRepository", "./fake", &g_options));
+}
+
+void test_online_cert__passthrough_https_bad_certificate(void)
+{
+	if (!g_has_ssl)
+		cl_skip();
+
+	g_options.fetch_opts.callbacks.certificate_check = certificate_check_cb;
+
+	g_cert_options.expect_valid = 0;
+	g_cert_options.return_code = GIT_PASSTHROUGH;
+
+	cl_git_fail_with(GIT_ECERTIFICATE,
+					 git_clone(&g_repo, "https://wrong.host.badssl.com/fake.git", "./fake", &g_options));
+}
+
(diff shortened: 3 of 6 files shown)


carlosmn:
It looks like you've misunderstood the purpose of the `GIT_PASSTHROUGH` return value. Its purpose is not to ignore the check, but to ignore that the certificate check _callback_ was set.

Some systems (my primary motivation are binidngs) might want to always set a certificate check callback for simplicity in their setup but if they themselves have users which have not set the certificate check, returning this lets them and the library behave as though it wasn't set.

What this code instead does is ignore the determination from the library about the validity of the certificate which is something we don't want to do in the library. If someone wants to ignore the certificate, we want them to return the all-ok return code.

tiennou:
I was actually worried that it would allow such easy circumventing of certificate checks 😆 .

I understand the motivation behind, but I'm not sure how to model that. IMHO the test I wrote was correct, in that the bindings `certificate_check_cb`, finding that there was no actual "checker function" would return `E_PASSTHROUGH`. If you have the time, could you describe (even in pseudo-code) what a correct test would be ? Then I'll arrange to fix the behavior.

carlosmn:
The simplest way I think would be to store the previous error and then return that if the callback returns `GIT_PASSTHROUGH`, something like

```
cberror = callback(...)
if cberror < 0 && cberror != GIT_PASSTHROUGH
    error = cberror
    handle error message, return error
```

should be almost enough (we still need. The main point is that if we would have returned `GIT_ECERTIFICATE` if the pointer isn't set, then we must still return that. At least for the http code, that should still let if fall out of the `if` with `GIT_ECERTIFICATE` in `error`.

tiennou:
Should we allow the user to return anything he wants ? I think the http & ssh backend differ already, so I'm considering supporting only the following cases :
- user returns 0 : `GIT_OK`
- user returns `E_PASSTHROUGH`: we return the last error if there was one, or we continue
- user returns anything else: we error, returning `GIT_EUSER`

Opinions ?

ethomson:
> user returns anything else: we error, returning GIT_EUSER

For anything non-zero, and not `E_PASSTHROUGH`, we should maintain the return code and bubble it back up to the user.

If I'm a user, and I return `-42` from a callback and suddenly my call exits with `-1` then that's surprising.  I want to be able to propagate return values from my callback to myself.

Related:  https://github.com/libgit2/libgit2/issues/2716

tiennou:
Thanks for the cross-ref, that clears it up. IMHO, I don't think it's sane because I don't think there's a way to discern a `libgit2` error code from one of my (user) error codes, but that's beside the point. What might make it cleaner is stash the user error code inside `git_error` and return `GIT_EUSER`.

ethomson:
> I don't think it's sane because I don't think there's a way to discern a libgit2 error code from one of my (user) error codes

There is, that's what `GIT_EUSER` is for.

My callbacks want to be able to round trip error codes; if my callback runs out of memory, I want to return `GITERR_NOMEMORY`.  That way my error handling code will handle it in a single case, whether libgit2 is the one that had a failed `malloc` or whether it was my code.  I don't really care _where_ I ran out of memory,  just that I _did_ run out of memory.

Otherwise, I would have to set some flag in my callback, `out_of_memory`, then return `GIT_EUSER`.  Then _all_ my code that does some networking that might invoke the callback has to `if (error == GIT_ENOMEMORY || (error == GIT_EUSER && out_of_memory))`.  Which is pretty disappointing.

This is obviously a contrived example, but I think that it illustrates that error handling is simplified if you don't try to coalesce all callback returns to a single error code.

This doesn't mean that anybody _has_ to use these error codes, and generally they shouldn't, but it allows them to use them when it simplifies their lives.

tiennou:
It seems I might not understand the relation between `GIT_EUSER` and the `GITERR_*` classes... I was under the impression that I shouldn't need to touch those `GITERR_*` values (and the docs warn that they might not be preserved, as per `giterr_set_str`). IMHO it also doesn't work in the face of error code conflicts. For example, let's say I'm wrapping another library as a smart transport, I'll have to remap every of its error codes to something that's not already used by us so I don't accidentally trigger error handling within `libgit2` by reusing -16 (`GIT_EAUTH`). API-wise this would prevent us from adding more negative `GIT_E*` values since there might be any number of remaps out there.

My (original) proposal would be to do have internals calls to client code handled like this :
```
int user_error = ...
if (user_error < 0) {
  giterr_set_user(user_error);
  return GIT_EUSER;
}
```
so the client can do
```
int err = git_function();
if (err == GIT_USER) {
  /* grab the underlying error from giterr_last() */
} else if (err  != GIT_OK) {
  /* libgit2 error */
}
```

Another way would be to make a new `GITERR_USER` error class. That would allow the error to be preserved as well as providing a nicer string message.

Sorry, I have a knack for nice error handling 😉.

ethomson:
> It seems I might not understand the relation between GIT_EUSER and the GITERR_* classes...

You probably do, I wrote `GITERR_NOMEMORY` without really thinking about it.  That was my mistake and I apologize for accidentally being misleading.  It was meant to be an example; consider `GIT_EAMBIGUOUS`, as a more accurate one.

> IMHO it also doesn't work in the face of error code conflicts.

There shouldn't be any, because you shouldn't be returning your own error codes.  You need to return libgit2 error codes.  If an odb backend returns `GIT_ELOCKED`, then that should get sent all the way back to the caller.  The bindings will then convert this to a `LockedException` or what not.

Custom backends _do this today_ and these semantics are already well agreed upon.  Your example of converting everything to `GIT_EUSER` breaks these sorts of semantics.

> For example, let's say I'm wrapping another library as a smart transport, I'll have to remap every of its error codes to something that's not already used by us so I don't accidentally trigger error handling within libgit2 by reusing -16 (GIT_EAUTH).

Yes, this is _exactly_ what you need to do.  These error values are completely meaningless to libgit2 and you shouldn't return them to the library.  You need to return a libgit2 error code that the library understand.

If you're writing a custom backend (be it transport, odb, refdb, etc) and you need to communicate with it, then you should set a flag in the backend that your library consumer can interrogate later.  That's what `GIT_EUSER` is for, generally speaking.

This looks like what you're trying to do with `giterr_set_user`, but this sort of custom error storage doesn't belong in the library, it's your business logic.

tiennou:
At least I'm green on the error class vs error code nuance 😉 .

> You need to return libgit2 error codes.

That's new to me, and I don't think it's clearly documented anywhere ? So callbacks are expected to use `GIT_EUSER` on any "unmappable" `git_error_code` ? That clears it up.

> we should maintain the return code and bubble it back up to the user.

> you should set a flag in the backend

> This looks like what you're trying to do with `giterr_set_user`

It is, as a way of preserving user error codes without conflating them with our own. So a user can know whether it's a `libgit2` failure, or something caused by "external" code called by `libgit2`, and still know what was the underlying error without having to jump though hoops (I'm not familiar with the custom backend code).

Using the custom backends as an example, apart from the redis one, none of them use `giterr_set_str`, and none return `GIT_EUSER`. The way I envision it, `GIT_E*` codes would be reserved to the library, and anything failing external functions would either return != `GIT_EUSER` and might be clobbered, or return `GIT_EUSER` (and optionally call `giterr_set_user(code, msg)`). At the outer layer, I would be able to discern a generic custom odb failure/libgit2 (using != `GIT_EUSER`), or something like "the database server is unhappy because whatever" (using `GIT_EUSER` and the optional user error code).

Again, I'm just thinking out loud. If you think it's out of scope, fine (at least the current rules are now clear to me, even though IMHO they're ambiguous).

tiennou:
Rebased, now that the security fixes are in. I had shuffled all cert-related tests around (I should have included that, sorry). I hope I got the tests right, because they pass 😉.

tests/online/cert.c
```diff
@@ -0,0 +1,256 @@
...
+		g_repo = NULL;
+	}
+	cl_fixture_cleanup("./fake");
+}
+
+static int certificate_check_cb(git_cert *cert, int valid, const char *host, void *payload)
+{
+	GIT_UNUSED(cert);
+
+	struct cert_options *options = payload;
```
ethomson: The windows builds fail because this isn't declared at the top of the block.

tiennou: Oops, will fix ASAP !

carlosmn:
It looks like the test doesn't quite work for WinHTTP. There we seem to be going all the way through to requesting the repository and thus getting a 404 instead of our own error.

tiennou:
Should be good now. The failure comes from a HTTP 502 from bitbucket.

tiennou:
Rebased. I don't think the failures are related, though I might be wrong.

pks-t:
Failures are unrelated, yes. They are due to our recently broken master branch when our bundled zlib is used, which I've fixed a few days ago. The other errors are due to the usual segfaults with poxyproxy... *sigh* Time to finally get our test suite polished up again.

tests/online/cert.c
```diff
@@ -0,0 +1,257 @@
...
+	if (!g_has_ssl)
+		cl_skip();
+
+	g_options.fetch_opts.callbacks.certificate_check = certificate_check_cb;
+
+	g_cert_options.expect_valid = 0;
+	g_cert_options.return_code = GIT_EPEEL;
+
+	cl_git_fail_with(GIT_EPEEL,
+					 git_clone(&g_repo, "https://wrong.host.badssl.com/fake.git", "./fake", &g_options));
```
carlosmn: `GIT_EPEEL` is a very confusing error to use  here. `GIT_EUSER` would be much better as it's the one we have set up for people to return and have it passed all the way back.

tests/online/cert.c
```diff
@@ -0,0 +1,257 @@
...
+	if (!g_has_ssl)
+		cl_skip();
+
+	g_options.fetch_opts.callbacks.certificate_check = certificate_check_cb;
+
+	g_cert_options.expect_valid = 0;
+	g_cert_options.return_code = GIT_PASSTHROUGH;
+
+	cl_git_fail_with(GIT_ERROR,
+					 git_clone(&g_repo, "https://wrong.host.badssl.com/fake.git", "./fake", &g_options));
```
carlosmn: This should be returning `GIT_ECERTIFICATE` as the passthrough means we want to use the result that the library was going to use if we didn't have the callback.

I believe this `GIT_ERROR`/`-1` actually comes from there not actually being a git repository at the given location.

tests/online/cert.c
```diff
@@ -0,0 +1,257 @@
...
+	opts.fetch_opts.callbacks.certificate_check = cert_check_assert_invalid;
+
+	if (!g_has_ssl)
+		cl_skip();
+
+	cl_git_fail_with(GIT_ECERTIFICATE,
+			 git_clone(&g_repo, "https://expired.badssl.com/fake.git", "./fake", NULL));
+
+	cl_git_fail_with(GIT_ECERTIFICATE,
+			 git_clone(&g_repo, "https://expired.badssl.com/fake.git", "./fake", &opts));
```
carlosmn: This isn't testing anything other than us bubbling up the return code, does it? We've already tested this elsewhere.

tests/online/cert.c
```diff
@@ -0,0 +1,257 @@
...
+	git_clone_options opts = GIT_CLONE_OPTIONS_INIT;
+	opts.fetch_opts.callbacks.certificate_check = cert_check_assert_invalid;
+
+	if (!g_has_ssl)
+		cl_skip();
+
+	cl_git_fail_with(GIT_ECERTIFICATE,
+			 git_clone(&g_repo, "https://wrong.host.badssl.com/fake.git", "./fake", NULL));
+	cl_git_fail_with(GIT_ECERTIFICATE,
+			 git_clone(&g_repo, "https://wrong.host.badssl.com/fake.git", "./fake", &opts));
```
carlosmn: It's the same here as above. Adding the callback here doesn't add anything to any of these tests which are about making sure we do reject a bunch of common certificate errors.

tests/online/cert.c
```diff
@@ -0,0 +1,257 @@
...
+
+	g_cert_options.expect_valid = 1;
+	g_cert_options.expect_host = "github.com";
+	g_cert_options.return_code = GIT_ECERTIFICATE;
+
+	cl_git_fail_with(GIT_ECERTIFICATE,
+					 git_clone(&g_repo, "https://github.com/libgit2/TestGitRepository", "./fake", &g_options));
+}
+
+void test_online_cert__certificate_invalid_ssh(void)
```
carlosmn: This test name (and `__certificate_invalid` above) is misleading. These certificates are valid, but we're testing here the ability to abort a clone in the certificate callback.

tiennou: Agreed, would `__valid_certificate_user_invalidated` be somewhat clearer that we respect the users' decision ?

tests/online/cert.c
tiennou: I'll have to recheck what's actually going on w.r.t to the first part. The second part is correct, we're actually succeeding the cert check, but the clone machinery obviously fails down the line, returning `GIT_ERROR`.

tests/online/cert.c
tiennou: Ok, will fix. The intent was to make sure we preserved whatever the users' callback replied to us (hence the non-sensical `GIT_EPEEL`).

tests/online/cert.c
tiennou: I'm merging all cert-related test cases in this one, so the "elsewhere" is this one now, thus it's testing that the TLS layer doesn't accept invalid certificates by default.

tests/online/cert.c
tiennou: Adding the callback is there to check we don't have a difference in behavior w.r.t invalid certificates whether there's a callback or not. I can remove if you feel like it's too much though.

tests/online/cert.c
carlosmn: >I'm merging all cert-related test cases in this one,

And I wish you hadn't done that, because I'm now reading the badssl test cases as though they're testing how we handle the feature you're writing instead of what they're for.

tiennou:
Rebased, with a focus on only passthrough (so the test cases aren't merged anymore) and squashed everything into one commit, with some more testing on top (and comments). I'm surprised how confusing this code can become when looking at it too much 😫.

I have been tempted to move that whole certificate check thing into a helper function somewhere, for DRY reasons, but didn't find a satisfying location.

src/transports/winhttp.c
```diff
@@ -283,15 +284,20 @@ static int certificate_check(winhttp_stream *s, int valid)
 		return -1;
 	}
 
-	giterr_clear();
+	giterr_state_capture(&git_error_state, GIT_ERROR);
```
ethomson: Shouldn't this be `giterr_state_capture(&error_state, GIT_ERROR);`?

src/transports/winhttp.c
```diff
@@ -283,15 +284,20 @@ static int certificate_check(winhttp_stream *s, int valid)
...
 	cert.parent.cert_type = GIT_CERT_X509;
 	cert.data = cert_ctx->pbCertEncoded;
 	cert.len = cert_ctx->cbCertEncoded;
 	error = t->owner->certificate_check_cb((git_cert *) &cert, valid, t->connection_data.host, t->owner->message_cb_payload);
 	CertFreeCertificateContext(cert_ctx);
 
-	if (error < 0 && !giterr_last())
+	if (error == GIT_PASSTHROUGH) {
+		/* Nothing could have failed previously, return success */
+		git_error_restore(&error_state);
```
pks-t: This has to be `giterr_state_restore`

src/transports/winhttp.c
```diff
@@ -283,15 +284,20 @@ static int certificate_check(winhttp_stream *s, int valid)
...
+	} else if (error < 0 && !giterr_last()) {
 		giterr_set(GITERR_NET, "user cancelled certificate check");
+	}
+	git_error_free(&error_state);
```
pks-t: This has to be `giterr_state_free`

tiennou:
🙄 Good grief, thanks @pks-t. Travis doesn't care about WinHTTP so that explains it.

src/transports/smart.c
```diff
@@ -460,11 +460,28 @@ static int ref_name_cmp(const void *a, const void *b)
...
+	} else {
+		error = cberror;
+		if (error < 0 && !giterr_last())
+			giterr_set(GITERR_NET, "user cancelled certificate check");
+	}
+	giterr_state_free(&state);
 
-	return t->certificate_check_cb(cert, valid, hostname, t->message_cb_payload);
+	return error;
 }
```
carlosmn: All this code shouldn't go here. This function is a way for code external to the library to execute the certificate check callback, The error you want to insert here may not exist in the code implementing a transport. They may be catching an exception several languages apart, and there isn't going to be any error state to capture.

If we need common code inside the library to do this repeated code, that needs its own function.

tiennou: I really wanted to make that passthrough-dance generic across all transports, and having it here removes the need for an external transport implementation to handle this case on their side. I can easily move it in a private function, but IMHO that kinda defeat the purpose, since it will no longer be "externally usable". I also like the eat-your-own-dogfood aspect of it.

The usage pattern I had in mind was that the external transport would "check the certificate" while connecting (whatever that means), call this with cert_error set to `GIT_OK` for valid, `GIT_ECERTIFICATE` for invalid (that's the "non-existing error case" from your point above, which our SSH transport demonstrates), `anything else` for spurious unrelated error, and this function would automatically preserve the callback error (which might not even be aware that an external transport is in use) or restore the transport's original success/failure state on a passthrough.

I'm not sure I understand your point about exceptions being several languages apart : should it be safe for an higher-level (say Ruby) external transport to raise an exception and catch that around the (say `git_remote_download`) call that started the activity, without going through our code at all 😨 ?

As an API sidenote, I was actually wondering about the usage pattern of those functions : the external code *might* end up calling a NULL function pointer, and I don't think there's a way for external users to know they're about to SEGV (`git_transport_smart_credentials` shares the same problem).

tiennou:
This has been tackled as part of #4879.

PR #4044, opened 2016-12-25
mempack: set the odb backend version

jacquesg:
A ```git_mempack``` backend is currently unusable due to the ```git_odb_backend``` version being set to ```0``` instead of ```GIT_ODB_BACKEND_VERSION```


ethomson:
Thanks for fixing this!

frzland777:
This is Japonese to me

PR #4045, opened 2016-12-26
Fix unpack double free

lhchavez:
If an element has been cached, but then the call to
packfile_unpack_compressed() fails, the very next thing that happens is
that its data is freed and then the element is not removed from the
cache, which frees the data again.

This change sets obj->data to NULL to avoid the double-free. It also
stops trying to resolve deltas after two continuous failed rounds of
resolution, and adds a test for this.

### DIFF
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -844,6 +844,7 @@ static int fix_thin_pack(git_indexer *idx, git_transfer_progress *stats)
 static int resolve_deltas(git_indexer *idx, git_transfer_progress *stats)
 {
 	unsigned int i;
+	int error;
 	struct delta_info *delta;
 	int progressed = 0, non_null = 0, progress_cb_result;
 
@@ -858,8 +859,13 @@ static int resolve_deltas(git_indexer *idx, git_transfer_progress *stats)
 
 			non_null = 1;
 			idx->off = delta->delta_off;
-			if (git_packfile_unpack(&obj, idx->pack, &idx->off) < 0)
-				continue;
+			if ((error = git_packfile_unpack(&obj, idx->pack, &idx->off)) < 0) {
+				if (error == GIT_PASSTHROUGH) {
+					/* We have not seen the base object, we'll try again later. */
+					continue;
+				}
+				return -1;
+			}
 
 			if (hash_and_save(idx, &obj, delta->delta_off) < 0)
 				continue;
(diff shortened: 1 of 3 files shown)


lhchavez:
This should be ready for review. PTAL!

ethomson:
This looks good to me.  @carlosmn - are there any subtleties in fixing thin packs?

carlosmn:
There often are subtleties. I'd like us to do the same kind of check that git does in this case in order to detect when no progress can be made. I'd have to get back to the details, but I'm not sure that we can assert we're stuck just because a single round did no yield any progress.

lhchavez:
IIUC git does this by making sure each unresolved delta is fully resolved before resolving the next one. at its core, there's a check[1] that verifies that progress was made in the last two rounds (similar in spirit to the `stuck` flag I'm proposing). It's probably doable to restructure libgit2's code to make it behave closer to that, but it might take a bit more work.

I did try to find a good scenario in which to test the double free() that didn't involve a malformed packfile to avoid having the fix depend on the stuck logic (otherwise the test never finishes), but I was unfortunately unable to find anything :(

1: https://github.com/git/git/blob/1d1bdafd64266e5ee3bd46c6965228f32e4022ea/builtin/index-pack.c#L1039

DemiMarie:
Is this a security hole?

lhchavez:
@DemiMarie yes, if you feed untrusted/arbitrary packfile data into the `git_indexer_append` function.

@carlosmn sorry for the delay (almost a year!?). i think we can get away without doing significant refactoring by either:

* checking that the inserted object will effectively resolve the delta, or
* by tagging each delta as having been resolved, and only process each delta once in `fix_thin_pack`, which avoids having a coarse two-try attempt.

i already have the latter coded with tests passing, but want to explore the former since it seems more robust and might catch more errors.

lhchavez:
@carlosmn all right, found a much more elegant solution! PTAAL.

lhchavez:
@carlosmn @ethomson @pks-t gentle ping? This is the only thing missing before the OSS-Fuzz support can be committed.

ethomson:
Sorry for the delay - this looks good to me.  😁

ethomson:
Would you mind squashing this?  I'll :shipit:

lhchavez:
I do not mind! Squashed and rebased.

ethomson:
Thanks!  👍 ✨ :shipit:

PR #4048, opened 2016-12-29
rebase: check the result code of rebase_init_merge

jacquesg:


ethomson:
Nice catch, thanks.  😀

PR #4051, opened 2016-12-29
Clang analyzer run

tiennou:
This is a few low-hanging fruits from using Xcode "Analyze" against libgit2.

Current `master` has 85 of those, this one is 81, and I managed to get it down to 64 by having more fine-grained control over `GITERR_CHECK_ALLOC` helps. If you feel that's a worthy goal, I can make another PR with those changes.

There are a few harder things which I can't tell if they're false-positives or not.

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -278,7 +278,7 @@ int git_ignore__for_path(
 	int error = 0;
 	const char *workdir = git_repository_workdir(repo);
 
-	assert(ignores && path);
+	assert(repo && ignores && path);
 
 	memset(ignores, 0, sizeof(*ignores));
 	ignores->repo = repo;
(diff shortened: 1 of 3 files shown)


ethomson:
Thanks, @tiennou !

tiennou:
As a side note, after silencing the warnings about alloc failures and some other things, I still see 47 analyzer reports.

Also, I have a question : the analyzer is "confused" by the differing stance we can have w.r.t to negative/not-zero error checks. Since I already derailed this PR with a discussion about error handling, I'll complete my tour with this one : is there a current, recommended idiom for `int` error returns, or is switching all instances to `!= 0` a goal worthy of being merged ?

ethomson:
> Also, I have a question : the analyzer is "confused" by the differing stance we can have w.r.t to negative/not-zero error checks. Since I already derailed this PR with a discussion about error handling, I'll complete my tour with this one : is there a current, recommended idiom for int error returns, or is switching all instances to != 0 a goal worthy of being merged ?

Yes to both of your questions.  😝 

The current, recommended idiom is:
```
if (error < 0)
```

Which - is pretty deficient, as you point out, and as we've had lingering at #2716.  The _new_ recommended idiom should probably be:
```
if (error)
```

Which is more succinct, readable and handles positive exit codes.

That's my opinion, anyway, but I haven't led the charge to make a fix.

Issue #4046, opened 2016-12-28
Segmentation fault in examples/general


dglsswnsn:
### Reproduction steps
I downloaded libgit2 v0.25.0 (Latest release) onto a virtual machine running Lubuntu 16.10.  After building the shared library (build/libgit2.so.0.25.0) using cmake, I invoked "make" in the examples/ directory.  All the executables built without errors or warnings, but when I invoked "general", it ended with a segmentation fault.

### Expected behavior
I expected "general" to run to completion and terminate normally.

### Actual behavior
Here are the last few messages that "general" writes to the terminal.

refs/tags/test [b25fa35b38051e4ae45d4222e795f9df2e43f1d1]
refs/tags/wrapped_tag [849a5e34a26815e821f865b8479f5815a47af0fe]
refs/heads/packed [41bc8c69075bbdb46c5c6f0566cc8cc5b46e8bd9]

*Config Listing*
Autocorrect: 32697
Segmentation fault (core dumped)
douglas@strangelove:~/libgit2-0.25.0/examples$ 

### Version of libgit2 (release number or SHA1)
v0.25.0

### Operating system(s) tested
Lubuntu 16.10

### Additional
Looking at the config_files() function in general.c, I guessed that the invocation of git_config_get_string() was causing the fault.  I found the following comment before this function's prototype in config.h.

 * This function can only be used on snapshot config objects.

so I guessed that the "cfg" parameter on this invocation was incorrect.  A little shopping around revealed the git_repository_config_snapshot() function, and the following snippet fixed the problem.

	git_config *snap_cfg;
	check_error(git_repository_config_snapshot(&snap_cfg, repo), "config snapshot");
	git_config_get_string(&email, snap_cfg, "user.email");

ethomson:
Thanks for troubleshooting this!  It would be great if you would submit a pull request to fix it!  😀

dglsswnsn:
Thanks for reading my issue!  Sure, I'll be glad to fix the problem and submit a pull request.  

I found a couple of other problems with the config_files() function.  I'll create another issue to document them.

ethomson:
Fixed via #4052

PR #4052, opened 2016-12-29
Fix issue #4046 Seg fault in config_files()

dglsswnsn:
According to the comments in the config.h file, git_config_get_string() needs a config snapshot.  Added call to git_repository_config_snapshot() to get this snapshot.

Added a "repo" parameter to config_files(), which is used by git_repository_config_snapshot().

I also tested (but did not commit) having git_config_get_int32() use the config snapshot.  This worked, and if implemented could eliminate the "repo_path" parameter to config_files().

### DIFF
--- a/examples/general.c
+++ b/examples/general.c
@@ -57,7 +57,7 @@ static void blob_parsing(git_repository *repo);
 static void revwalking(git_repository *repo);
 static void index_walking(git_repository *repo);
 static void reference_listing(git_repository *repo);
-static void config_files(const char *repo_path);
+static void config_files(const char *repo_path, git_repository *repo);
 
 /**
  * Almost all libgit2 functions return 0 on success or negative on error.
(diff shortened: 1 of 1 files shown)


dglsswnsn:
It looks like the Travis CI Build Job #9100.3 failed with a segmentation fault in libgit2_clar.  I enclose the relevant lines from the log.

1: threads::basic....
1: threads::diff..
1/4 Test #1: libgit2_clar .............................***Exception: SegFault 37.42 sec
test 2

My commit changed only one file -- examples/generate.c, and I'm fairly sure that libgit2_clar doesn't run the examples/generate executable.  I also noticed that other Trravis CI Build Jobs for this Pull Request were successful, e.g. #9100.2, where this area of its log looks like

1: threads::basic...S
1: threads::diff..
1: threads::iterator.
1: threads::refdb.
1: trace::traceSSSSSS
1: trace::windows::stacktrace...
1: transport::register....
1: win32::forbidden..........
1: win32::longpath.
1: 
1/4 Test #1: libgit2_clar .............................   Passed   41.60 sec
test 2

I noticed that Pull Request #4051, which was submitted about an hour before mine, has an identical failure for Build Job #9099.8.  

Could this segmentation fault be an intermittent bug?  Is there perhaps a way to bypass this test so Pull Requests can be processes?

ethomson:
Thanks!  😀

PR #4053, opened 2016-12-29
Extend packfile in increments of page_size.

chescock:
I have an application that uses LibGit2Sharp, and it's been having performance issues doing fetches.  Using the profiling technique of running it in the debugger and pausing it, I discovered it was spending a lot of time calling `p_write` from `append_to_pack` to write a single byte.  

I guessed that there was a cost for each call that actually extended the file, and that we could reduce the overall cost by rounding the size up so that we extend the file fewer times in larger chunks.  I'm using `mmap_alignment` for the rounding since that was used in the nearby `write_at` method, but I don't know whether or not that's an appropriate value.  

A local fetch of the libgit2 repository on my Windows 7 machine takes 307s without this change and 43s with it, for a 7x speedup, and the fetch goes from being I/O-bound to CPU-bound.

### DIFF
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -477,13 +477,29 @@ static int write_at(git_indexer *idx, const void *data, git_off_t offset, size_t
 
 static int append_to_pack(git_indexer *idx, const void *data, size_t size)
 {
+	git_off_t new_size;
+	size_t mmap_alignment;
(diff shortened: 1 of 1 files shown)


ethomson:
That's quite an improvement.  Thanks, @chescock !

PR #4054, opened 2016-12-30
Fix handling of GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH flag.

jfultz:
git_checkout_tree() sets up its working directory iterator to respect the
pathlist if GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH is present, which is great.
What's not so great is that this iterator is then used side-by-side with
an iterator created by git_checkout_iterator(), which did not set up its
pathlist appropriately (although the iterator mirrors all other iterator
options).

This could cause git_checkout_tree() to delete working tree files which
were not specified in the pathlist when GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH
was used, as the unsynchronized iterators causes git_checkout_tree() to think
that files have been deleted between the two trees.  Oops.

And added a test which fails without this fix (specifically, the final check
for "testrepo/README" to still be present fails).

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -2553,6 +2553,10 @@ int git_checkout_iterator(
 		GIT_ITERATOR_IGNORE_CASE : GIT_ITERATOR_DONT_IGNORE_CASE;
 	baseline_opts.start = data.pfx;
 	baseline_opts.end = data.pfx;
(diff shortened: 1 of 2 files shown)


ethomson:
Thanks, @jfultz for catching this and the fix.  Thanks especially for including a unit test.  ❤️

PR #4057, opened 2016-12-31
Detail exactly which objects can no longer be used after repository free'd

simonbyrne:
The background is in [this StackOverflow question](http://stackoverflow.com/q/41408798/392585).

Based on my reading of the source, the following ticked exported opaque objects keep pointers to a `git_repository` object (possibly via some intermediate other object):

- [ ] `git_odb`
- [ ] `git_odb_backend`
- [ ] `git_odb_object`
- [ ] `git_odb_stream`
- [ ] `git_odb_writepack`
- [x] `git_refdb`
- [ ] `git_refdb_backend`
- [x] `git_object`
- [x] `git_revwalk`
- [x] `git_tag`
- [x] `git_blob`
- [x] `git_commit`
- [ ] `git_tree_entry`
- [x] `git_tree`
- [x] `git_treebuilder`
- [ ] `git_index`
- [ ] `git_index_conflict_iterator`
- [ ] `git_config`
- [ ] `git_config_backend`
- [ ] `git_reflog_entry`
- [x] `git_reflog`
- [ ] `git_note`
- [x] `git_packbuilder`
- [x] `git_reference`
- [x] `git_reference_iterator`
- [x] `git_transaction`
- [x] `git_annotated_commit`
- [ ] `git_merge_result` (does not appear to be defined/used)
- [x] `git_status_list`
- [x] `git_rebase`
- [ ] `git_refspec`
- [x] `git_remote`
- [ ] `git_transport`
- [x] `git_push`
- [ ] `git_remote_head`
- [ ] `git_remote_callbacks`
- [x] `git_submodule`
- [ ] `git_writestream`

If there are any errors or omissions, please let me know and I can update the PR. I'm not 100% sure about `git_index`, as it's quite a complicated object: could someone confirm this for me?

### DIFF
--- a/include/git2/repository.h
+++ b/include/git2/repository.h
@@ -160,11 +160,13 @@ GIT_EXTERN(int) git_repository_open_bare(git_repository **out, const char *bare_
 /**
  * Free a previously allocated repository
  *
- * Note that after a repository is free'd, all the objects it has spawned
- * will still exist until they are manually closed by the user
- * with `git_object_free`, but accessing any of the attributes of
- * an object without a backing repository will result in undefined
- * behavior
+ * Note that after a repository is free'd, all objects it has spawned will still
+ * exist until they are manually closed by the user (e.g. with
+ * `git_object_free`), but using the object in any other manner without a
+ * backing repository will result in undefined behavior. The exceptions to this
+ * rule are `git_config`, `git_config_entry`, `git_index`, `git_index_entry`,
+ * `git_odb`, `gt_reflog_entry`, `git_tree_entry`, `git_signature` and
+ * `git_note` objects.
  *
  * @param repo repository handle to close. If NULL nothing occurs.
  */


ethomson:
With very few exceptions, users should assume that `git_repository_free` invalidates all the objects that were associated with it.  We reserve the right to cache things, give you pointers to a string that's in a pool, etc.

I'm not sure that I want to enumerate the things that are _not_ safe to deal with after a `git_repository_free`...  this is going to basically be everything, right?

I can't think of any types that we make a guarantee about being usable after a `git_repository_free` off the top of my head.  So the few that there are (_if_ there are) should be whitelisted explicitly.

The list will be shorter and easier to consume, and less likely to be incorrect and / or get out of date.

simonbyrne:
That seems reasonable (I also now realise that above list is incomplete, I had only looked in `types.h`).

Some possible exceptions I have so far found (please confirm if these are not the case):

* `git_config` and `git_config_entry` (from [`git_repository_config`](https://libgit2.github.com/libgit2/#HEAD/group/repository/git_repository_config))
* `git_index` and `git_index_entry` (from [`git_repository_index`](https://libgit2.github.com/libgit2/#HEAD/group/repository/git_repository_index))
* `git_odb` (from [`git_repository_odb`](https://libgit2.github.com/libgit2/#HEAD/group/repository/git_repository_odb))
* `git_reflog_entry` (from [`git_reflog_entry_byindex`](https://libgit2.github.com/libgit2/#HEAD/group/reflog/git_reflog_entry_byindex)
* `git_tree_entry` (from, e.g., [`git_tree_entry_byindex`](https://libgit2.github.com/libgit2/#HEAD/group/tree/git_tree_entry_byindex)
* `git_signature` (from [`git_signature_default`](https://libgit2.github.com/libgit2/#HEAD/group/signature/git_signature_default))
* `git_note` (from [`git_note_read`](https://libgit2.github.com/libgit2/#HEAD/group/note/git_note_read))

simonbyrne:
I've updated the pull request to reflect these changes.

carlosmn:
Using the names of structs isn't nearly enough to know what can be freed, and there doesn't have to be a repository object involved in order to use parts of libgit2.

A `git_tree_entry`, for example, may be owned by the caller or the tree object depending on whether you looked it up by entry name or looked up a deep path, or whether you performed a copy on whatever you got back.

This is why the conventions mention _how_ you receive data from the library, not what their names are. Function calls returning data should also mention whether there are deviations from this (e.g. int he config case where we want to returned borrowed data but also need to be able to return an error).

"Undefined behaviour" is something that happens in a programming language, which we are not.

ethomson:
I agree with @carlosmn - we should have this documented in the conventions.  _I_ think this is roughly clear, but I'm quite biased.  If we're wrong or need to clarify this, then we should, but I don't think that we can document the objects in this way.

Instead, we need to discuss further the patterns and practices that we use...  in the meantime I'm closing this, but if further clarification is needed, a further chat (here or in the slack) on the problems and how to clarify would be the next step forward.

Thanks!

Issue #4047, opened 2016-12-28
examples/general config_files() needs to check return codes from git_config_get_ invocations


dglsswnsn:
### Reproduction steps
I do not have a help.autocorrect configuration parameter.  When I run examples/general multiple times, I see a different value displayed for Autocorrect each time, e.g.

*Config Listing*
Autocorrect: 32697

*Config Listing*
Autocorrect: 32720

### Expected behavior
If a configuration parameter is not set, I would expect config_files() to tell me that the parameter is missing or undefined.

### Actual behavior
The invocation of git_config_get_int32() is failing, but its return code isn't tested before its returned value is used.  Both this invocation and the following invocation of git_config_get_string() should test the return code for three conditions:
- 0 (success)
- GIT_ENOTFOUND (undefined configuration parameter)
- Other (failure)

### Version of libgit2 (release number or SHA1)
v0.25.0

### Operating system(s) tested
Lubuntu 16.10

### Additional
config_files() creates a new git_config object when it calls git_config_open_ondisk().  Before returning config_files() should free this object by calling git_config_free().  I believe that "as is" config_files() is leaking memory.  In this case that's immaterial, but as an example, it should be a good one and not leak memory.

ethomson:
Closed via #4078

PR #4059, opened 2017-01-01
Fix Issue #4047 Check return codes and free objects

dglsswnsn:
Prior to this fix config_files() in examples/general.c produced random / incorrect values for "Autocorrect" in its output.

Prior to this fix config_files() did not free its config objects.

### DIFF
--- a/examples/general.c
+++ b/examples/general.c
@@ -699,6 +699,7 @@ static void config_files(const char *repo_path, git_repository* repo)
 	int32_t j;
 	git_config *cfg;
 	git_config *snap_cfg;
+	int error_code;
 
 	printf("\n*Config Listing*\n");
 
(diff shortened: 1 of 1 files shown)


examples/general.c
```diff
@@ -708,10 +709,32 @@ static void config_files(const char *repo_path, git_repository* repo)
...
+		default:
+			check_error(error_code, "get_int32 failed");
+	}
+	git_config_free(cfg);
 
 	check_error(git_repository_config_snapshot(&snap_cfg, repo), "config snapshot");
-	git_config_get_string(&email, snap_cfg, "user.email");
-	printf("Email: %s\n", email);
+	error_code = git_config_get_string(&email, snap_cfg, "user.email");
+	switch (error_code)
```
ethomson: Instead of copy/pasting this, a macro might clean this up nicely...?

dglsswnsn: I did cut and paste tge fragments starting with the "switch".  A macro would be a nice way to avoid the duplication, but it will have to take a number of parameters:
 1) The displayed name of the variable (Autocorrect or Email).
 2) The variable's type (%d or %s).
 3) The value of the variable.
 4) The error message for check_error().
 5) Maybe "error_code" so the user of the macro could use any name for this variable.

I'll be glad to add a macro if you think that's nicer.

dglsswnsn: I was curious how your proposed macro would look so I coded it.  I enclose below the macro and its invocations.  Is this better than the original?  

```
#define GIT_CFG_ERR_CHK(display_name, format, variable, error_msg, error_code) \
    switch (error_code) \
    { \
        case 0: \
            printf(display_name ": " format "\n", variable); \
            break; \
        case GIT_ENOTFOUND: \
            printf(display_name ": Undefined\n"); \
            break; \
        default: \
            check_error(error_code, error_msg); \
	}

    error_code = git_config_get_int32(&j, cfg, "help.autocorrect");
    GIT_CFG_ERR_CHK("Autocorrect", "%d", j, "get_int32 failed", error_code)

    error_code = git_config_get_string(&email, snap_cfg, "user.email");
    GIT_CFG_ERR_CHK("Email", "%s", email, "get_string failed", error_code)
```

ethomson:
Sorry, I should have merged this long ago.  I don't love the copypasta, but it's superior to the macro for code clarity in an example, I think.

Thanks for the contribution!

ethomson:
Merged via #4370

PR #4060, opened 2017-01-01
Delete temporary packfile in indexer

lhchavez:
This change deletes the temporary packfile that the indexer creates to
avoid littering the pack/ directory with garbage.

### DIFF
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -1091,6 +1091,12 @@ void git_indexer_free(git_indexer *idx)
 	git_vector_free_deep(&idx->deltas);
 
 	/* Try to delete the temporary file in case it was not committed. */
+	git_mwindow_free_all(&idx->pack->mwf);
+	/* We need to close the descriptor here so Windows doesn't choke on unlink */
+	if (idx->pack->mwf.fd != -1) {
+		(void)p_close(idx->pack->mwf.fd);
+		idx->pack->mwf.fd = -1;
+	}
 	(void)p_unlink(idx->pack->pack_name);
 
 	if (!git_mutex_lock(&git__mwindow_mutex)) {
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -151,11 +151,11 @@ cleanup:
 	if (fd != -1)
 		p_close(fd);
 
-	if (git_buf_is_allocated(&tmp_path))
-		(void)p_unlink(git_buf_cstr(&tmp_path));
+	if (git_buf_len(&tmp_path) > 0)
+		p_unlink(git_buf_cstr(&tmp_path));
 
 	if (idx->pack != NULL)
-		(void)p_unlink(idx->pack->pack_name);
+		p_unlink(idx->pack->pack_name);
 
 	git_buf_free(&path);
 	git_buf_free(&tmp_path);
@@ -1094,10 +1094,10 @@ void git_indexer_free(git_indexer *idx)
 	git_mwindow_free_all(&idx->pack->mwf);
 	/* We need to close the descriptor here so Windows doesn't choke on unlink */
 	if (idx->pack->mwf.fd != -1) {
-		(void)p_close(idx->pack->mwf.fd);
+		p_close(idx->pack->mwf.fd);
 		idx->pack->mwf.fd = -1;
 	}
-	(void)p_unlink(idx->pack->pack_name);
+	p_unlink(idx->pack->pack_name);
 
 	if (!git_mutex_lock(&git__mwindow_mutex)) {
 		git_packfile_free(idx->pack);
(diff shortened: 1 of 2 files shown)


carlosmn:
What errors are you seeing that cause this function to fail? It should be exceedingly rare that anything in this function fails and the most likely culprit would be the creation of this file itself.

lhchavez:
in `git_indexer_new`? nothing I can think of, it's just there for correctness, but I can remove it if needed. `git_indexer_free` will always leak the file if it `git_indexer_commit` was not called before (say, because you were writing a validator for a pack that was uploaded, and although the packfile was correct, the commit itself would fail validation).

lhchavez:
Now with a test that actually fails if I remove the check in `git_indexer_commit`. PTAL!

lhchavez:
now the tests actually pass -_-. PTAL.

src/indexer.c
```diff
@@ -151,6 +151,12 @@ int git_indexer_new(
 	if (fd != -1)
 		p_close(fd);
 
+	if (git_buf_is_allocated(&tmp_path))
```
ethomson: `git_buf_len(&tmp_path)` is correct here.

`is_allocated` literally looks at whether we have `malloc`ed the structure, not whether it has contents.  These are orthogonal concepts:  the buffer may point to static memory and thus be nonempty but not allocated, similarly, we could allocate a buffer, set contents into it, then clear the contents, but it would still be allocated.

src/indexer.c
```diff
@@ -151,6 +151,12 @@ int git_indexer_new(
...
+		(void)p_unlink(git_buf_cstr(&tmp_path));
```
ethomson: We do not cast unused return values to `void`.  There are a few other places in here that should be removed also.

tests/pack/indexer.c
```diff
@@ -125,3 +125,51 @@ void test_pack_indexer__fix_thin(void)
...
+{
+	git_indexer *idx = NULL;
+	git_buf path = GIT_BUF_INIT;
+	git_buf first_tmp_file = GIT_BUF_INIT;
+
+	/* Precondition: there are no temporary files. */
+	cl_git_pass(git_buf_sets(&path, clar_sandbox_path()));
+	cl_git_pass(find_tmp_file_recurs(&first_tmp_file, &path));
+	git_buf_free(&path);
+	if (git_buf_is_allocated(&first_tmp_file)) {
```
ethomson: I wouldn't `cl_skip` here.  Do you expect temp files might exist?  Then please don't skip the test, fix it.  If not, assert that they don't so that we fail.

tests/pack/indexer.c
```diff
@@ -125,3 +125,51 @@ void test_pack_indexer__fix_thin(void)
...
+	git_indexer_free(idx);
+
+	cl_git_pass(git_buf_sets(&path, clar_sandbox_path()));
+	cl_git_pass(find_tmp_file_recurs(&first_tmp_file, &path));
+	git_buf_free(&path);
+	if (git_buf_is_allocated(&first_tmp_file)) {
+		cl_warning(git_buf_cstr(&first_tmp_file));
+		git_buf_free(&first_tmp_file);
+		cl_fail("Found a temporary file");
+	}
```
ethomson: There's no need to free things on failures, we don't bother testing for memory leaks if it doesn't even work.  This 5 lines can be more easily rewritten as:

```
cl_assert(git_buf_len(&first_tmp_file) > 0)
git_buf_free(&first_tmp_file);
```

(Which, I think, is probably what the test above should look like.)

tests/pack/indexer.c
lhchavez: i expect they shouldn't exist (and currently don't), but cannot make that guarantee currently (I opened https://github.com/libgit2/libgit2/issues/4066 for that).

If we are cool with this test failing when other tests leak files, i can change the skip to a warning to make it clear that it wasn't this test that caused the failure, but another one.

src/indexer.c
lhchavez: Done

src/indexer.c
lhchavez: Done (saw this was used inconsistently in a few other places).

tests/pack/indexer.c
lhchavez: After a second thought, if this fails, it is obvious from the comment in L154 that it's not this test's fault, so changed it to simple assert with no log after all.

tests/pack/indexer.c
lhchavez: Done

ethomson:
I added a quick check before doing the `unlink` to avoid calling it when we have already moved the file into place.  I merged this manually; thanks again for fixing this!

PR #4061, opened 2017-01-01
merge: set default rename threshold

### LINKED ISSUE #4035: Crash in merge_conflict_can_resolve_contents
I'm trying to write a merge example, and while running it against a trivial test repository, I hit a crash :

### Reproduction steps

- Code [here](https://gist.github.com/tiennou/9c3f022bb59b9648f8f4757a4d807065).
- History :
```
create file1            mod file1
      A ----------------- B            (master)
        \
         \     create file2
          -------- C                   (branch)
```
- Attempt to merge `branch` with `GIT_MERGE_FIND_RENAMES` set.
- Crash [here](https://github.com/libgit2/libgit2/blob/c655def1174f814c2adcafff7cecec4d47dc5830/src/merge.c) with a NULL-ptr dereference on `conflict->ancestor_entry.path`.

SHA: c655def1174f814c2adcafff7cecec4d47dc5830

OS X 10.11.6

ethomson:
When `GIT_MERGE_FIND_RENAMES` is set, provide a default for `rename_threshold` when it is unset.

The merge machinery expects `rename_threshold` to be nonzero if `GIT_MERGE_FIND_RENAMES` is set.  Previously, a consumer could set `GIT_MERGE_FIND_RENAMES` but leave `rename_threshold` unset and merge would get terribly confused.

Fixes #4035

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ v0.25 + 1
 
 ### Changes or improvements
 
+* `GIT_MERGE_OPTIONS_INIT` now includes a setting to perform rename detection.
+  This aligns this structure with the default by `git_merge` and
+  `git_merge_trees` when `NULL` was provided for the options.
+
 ### API additions
 
 ### API removals
--- a/include/git2/merge.h
+++ b/include/git2/merge.h
@@ -290,7 +290,8 @@ typedef struct {
 } git_merge_options;
 
 #define GIT_MERGE_OPTIONS_VERSION 1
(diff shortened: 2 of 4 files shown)


carlosmn:
Did you man to change the library's default to be finding renames? That would certainly need an entry in CHANGELOG and it doesn't sound like what the commit message is saying.

ethomson:
Yeah, that seems reasonable.  This is bringing the defaults in line when you actually set up a structure using `GIT_MERGE_OPTS_INIT` versus simply passing `NULL` for the options.  I added a change log entry to reflect that.

PR #4074, opened 2017-01-09
fix examples/network/clone.c: heap-buffer-overflow

fcharlie:
Format of a length of string to the correct format is:%.*s

How to reproduce? Used when compiling libgit2 AddressSanitizer flag
`-fsanitize=address -fno-omit-frame-pointer`,
 then under the example program gives you.


ethomson:
Oops!  Thanks for fixing this.  😀

PR #4082, opened 2017-01-17
Rebase: Re-use existing commits if nothing changed.

arthurschreiber:
`git rebase` does not create new commit objects in all cases - only if actual attributes of the commit  (like parent ids, tree id, message, etc.) change.

rebase, as exposed by libgit2, does not behave in the same way - new commit objects are always generated and written to the odb.

This is a naïve patch that makes `rebase_commit__create` behave more like `git rebase`.

### DIFF
--- a/src/rebase.c
+++ b/src/rebase.c
@@ -967,6 +967,16 @@ static int rebase_commit__create(
 		goto done;
 	}
 
+	if (git_oid_equal(&tree_id, git_commit_tree_id(current_commit)) &&
+			git_commit_parentcount(current_commit) != 0 &&
+			git_oid_equal(git_commit_id(parent_commit), git_commit_parent_id(current_commit, 0)) &&
+	    (!author || git_signature__equal(author, git_commit_author(current_commit))) &&
+	    (!message || git__strcmp(message, git_commit_message(current_commit)) == 0) &&
+	    (!message_encoding || git__strcmp(message_encoding, git_commit_message_encoding(current_commit)) == 0)) {
+		git_commit_dup(out, current_commit);
+		goto done;
+	}
+
 	if (!author)
 		author = git_commit_author(current_commit);
 
@@ -1022,12 +1032,14 @@ static int rebase_commit_merge(
 			rebase->repo, NULL, "HEAD", git_commit_id(commit), "rebase")) < 0)
 		goto done;
 
-	git_oid_fmt(old_idstr, &operation->id);
-	git_oid_fmt(new_idstr, git_commit_id(commit));
+	if (!git_oid_equal(&operation->id, git_commit_id(commit))) {
+		git_oid_fmt(old_idstr, &operation->id);
+		git_oid_fmt(new_idstr, git_commit_id(commit));
 
-	if ((error = rebase_setupfile(rebase, REWRITTEN_FILE, O_CREAT|O_WRONLY|O_APPEND,
-		"%.*s %.*s\n", GIT_OID_HEXSZ, old_idstr, GIT_OID_HEXSZ, new_idstr)) < 0)
-		goto done;
+		if ((error = rebase_setupfile(rebase, REWRITTEN_FILE, O_CREAT|O_WRONLY|O_APPEND,
+			"%.*s %.*s\n", GIT_OID_HEXSZ, old_idstr, GIT_OID_HEXSZ, new_idstr)) < 0)
+			goto done;
+	}
 
 	git_oid_cpy(commit_id, git_commit_id(commit));
 


ethomson:
I've been beating on this for a while, and I think that this makes sense in generally.  I was initially worried that there might be some cases where git _doesn't_ do this, but it seems fairly straightforward.  `git-rebase--merge.sh` and `git-rebase--am` both seem to do this test.

`git-rebase--am` exits with:
```
No changes -- Patch already applied.
```

and `git-rebase--merge` exits with:
```
Already applied: 0001 23
```

It would be nice if - instead of continuing on like nothing happened and returning an old commit instead of creating a new one - we signaled that this was the case much like we do for the case right above it.  Which makes me wonder if I screwed up that test and it should have been looking at `current_commit` instead of `parent_commit`?

ethomson:
Okay, after further noodling, this is definitely a distinct case from whether the patch is a noop, so ignore my comments about the `parent_commit` test.

We don't want to just return the commit in question from here, though, as I think we need to indicate when a commit was actually rewritten versus simply taken.  (For example, the `rewritten` file should only contain commits that were actually rewritten.)

ethomson:
So as a user of the library, I would like to know when this occurs (so that I could print something like "Already applied").  There are a couple of possibilities here:

1. `git_rebase_commit` could fail completely and you're expected to know what to do.  This seems shitty.
2. `git_rebase_commit` could update `HEAD`, return some distinct code (eg `GIT_EAPPLIED` or `GIT_EAPPLIED2`) and set the out commitid param to the commit id that points to the commit id in question.  Users _may_ print something if they're interested, or simply ignore this code and keep looping (like they do with `GIT_EAPPLIED` already).
3. `git_rebase_commit` could simply return `GIT_OK`, set the out commitid param and assume that nobody cares.  We can set up some callback structure for people to get this information if/when they want it, but that's not something that matters now.

@arthurschreiber what do you think about this from your perspective?

arthurschreiber:
Well, `1` does not sound very convincing.

I prefer `3`, as it sounds more like what we want, and how `git rebase` works.

This still leaves one questions open: As a user, I might want to force the creation of a new commit, because I'm rewriting committer/author information. Is this something we want to support through the `rebase` APIs, or do we tell users to perform similar actions through a different API (git cli has `filter-branch` for this).

We could also break `git_rebase_commit` into 2 functions `git_rebase_commit` which would always create a new commit, and `git_rebase_continue` (or something similar), which would not create a new commit unless `git rebase` would.

arthurschreiber:
I just pushed another change that makes sure these non-rewritten commits don't end up in the `rewritten` file.

src/rebase.c
```diff
@@ -967,6 +967,13 @@ static int rebase_commit__create(
 		goto done;
 	}
 
+	if ((!message || strcmp(message, git_commit_message(current_commit)) == 0) &&
+	    (!message_encoding || strcmp(message_encoding, git_commit_message_encoding(current_commit)) == 0) &&
```
ethomson: Hmm.  I wonder if we shouldn't also consider `author` and `committer` here...?

ethomson:
So I think that this makes sense.  (I did mention author and committer w/r/t determining when to reuse and when not to.)  I'm going to add a couple of tests, but this is a nice improvement.

src/rebase.c
arthurschreiber: Well, committer is a non-optional argument to this function, for one.

Also, what would happen if a user passes author or comitter - would we have to check whether they differ from the previous atuhor/committer? If that makes sense, I can implement the necessary changes.

ethomson: Yes to author, I think...  I expect most people will pass `NULL` but I think that if I wanted to force a new author I would be disappointed here...

No to committer, though.  Users would have to go get the committer information from the old commit to make this work, which seems like a terrible burden.

arthurschreiber:
Okay, I just pushed another change:

* `committer` is only used when we write out new commits.
* If one of `author`, `message` or `message_encoding` is specified *and* different from what it is in the rebased commit, we will write out a new commit.

Do we also need to check the parent commit ids?

ethomson:
> Do we also need to check the parent commit ids?

I don't think so; git doesn't.  If you have some commit that is treesame to another, even if it has different parents, `git-rebase--am` will take the new one.  This was surprising to me, but :man_shrugging: .

> otherwise we might run into some weird issue when rebasing a set of commits onto a commit that introduced no new changes.

Hmm, sorry, what would the weird issue be here?

ethomson:
So I think that we discussed this further in private chat and decided that things were in fact working correctly without this patch - this was not necessary and we had made an error in rugged, perhaps.  So I'm closing this but please correct me if I'm mistaken.

PR #4085, opened 2017-01-21
indexer: introduce `git_packfile_close`

ethomson:
Encapsulation!


ethomson:
When I fixed up #4060, I inadvertently broke things on Windows.  This refactors things a bit to get us a cleaner solution.

PR #4087, opened 2017-01-23
Fix a few recent warnings

tiennou:


include/git2/tree.h
```diff
@@ -381,7 +381,7 @@ GIT_EXTERN(int) git_treebuilder_write(
  *
  * @see git_treebuilder_write
  *
- * @param id Pointer to store the OID of the newly written tree
+ * @param oid Pointer to store the OID of the newly written tree
```
pks-t: This change conflicts with the one from @ethomson regarding the same issue. Only that he changed the argument's name in the function prototype instead of the docs. Hopefully we will not end up with both patches applied, resulting in the same but reversed issue :P

ethomson: Generally, we prefer `id` in things like `commit_id` or `tree_id`.  We're sort of mixed about it when we're just talking about _an_ id.  There are a significant number of `oid`s when we're not explicit about what kind of id.

Anyway, we might want to clarify that, but this needn't be the place.  I'll drop my patches for these two changes.

ethomson:
Thanks, @tiennou !

include/git2/tree.h
carlosmn: Yes, we decided a while back that the style is to use `id` as the name for object names, and "oid" is just a type. We did go through the API at one point to fix it all up, but it's quite probably they've snuck back in.

PR #4091, opened 2017-01-24
tests: checkout: tree: fix memory leak

pks-t:


ethomson:
Great minds! 😂 https://github.com/libgit2/libgit2/pull/4086/commits/1f813cf2b8c9b1976cec15014dec8edd854f8e88

pks-t:
Heh :p Should've realized that this has already been fixed by your PR as I've just looked through it previously :laughing:

PR #4096, opened 2017-01-27
Update docs for git_oid_fromstrn and p

### LINKED ISSUE #4058: Incorrect documentation for git_oid_fromstrn
The docs say:

>If N is odd, N-1 characters will be parsed instead. The remaining space in the git_oid will be set to zero.

This is clearly undesirable behavior when working with prefixes and - more importantly - *wrong*. The function works just fine with odd character counts and parses the nibble.

mplough:
Fix #4058.


ethomson:
Thanks for fixing this!

PR #4097, opened 2017-01-27
Fix proxy auto detect not utilizing callbacks

implausible:
When using an authenticated proxy and setting proxy_options `type` to `GIT_PROXY_AUTO`, the credentials, certificate check, and payload are discarded. The implication is users will not be prompted for credentials when proxy type is set to auto. I've fixed this by transferring the credentials, certificate_check, and payload to the new struct when in apply_proxy_config.


carlosmn:
Looks sensible, thanks!

PR #4098, opened 2017-01-28
Update index_to_workdir CRLF conversion to match Git >= 2.9 logic

csware:
With Git 2.9.0 the CrLf filter logic changed (cf. https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.9.0.txt). This updates libgit2 to match that new logic for checking out files.


csware:
The counterpart is fixed in PR #3285.

pks-t:
Thanks for tackling these issues!

My criticism here is focusses heavily on the commit messages, as these changes are all not obvious at all without the reader digging himself into the source code of upstream git. Thus, having proper commit messages here would help me a lot in making sure that your changes are in fact the right thing to do.

- It would be nice if you'd detail a bit what changes upstream git has done. A pointer to the commits changing behaviour upstream would be greatly appreciated. Furthermore, pointing out what we were missing would be nice, too.
- It would be nice if the test data update commits had some short recipe for how to do these updates. I bet you were just doing some checkouts on Windows/POSIX platforms for these repos -- adding this recipe such that I can verify myself that your changes are correct would help.

While at it, some proposals for the commit subjects:
- tests: crlf: update Windows test data to reflect Git 2.9+ behaviour
- tests: crlf: update POSIX test data to reflect Git 2.9+ behaviour
- crlf: update checkout logic to reflect Git 2.9+ behaviour

Sorry to be nitpicky about commit messages here, but it would really go a long way in explaining what changes you did and why you had to do them.

csware:
@pks-t Done. Also rebased upon current master.

ethomson:
Thanks, @csware.  I started reviewing this PR a while ago, and I was having problems recreating the changes to the test resources that you're adding.  I apologize for not coming back to this until now.  I've re-run the resource creation test on macOS and Windows and agree with all the test resource changes that you've made.  I'm not sure what I did wrong the first time, but 👍 and sorry for the long delay.

ethomson:
To address @pks-t's feedback a little bit better, I added some more information about how the `tests/resources/crlf_data` data is generated.  I added usage information to the script, and added a `README` in the `crlf_data` directory itself to point to it.  I'm manually merging your changes to add that on top.

Thanks again for doing this.

PR #4104, opened 2017-02-01
Fix digest credentials for proxy in windows

cbargren:
Using a proxy with digest credentials was not working in Windows. It was simply always using basic auth whenever a plaintext credential mechanism was specified. Detecting whether the plaintext credential mechanism was digest or basic and applying that mechanism to `WinHttpSetCredentials` fixed the problem.

### DIFF
--- a/src/transports/winhttp.c
+++ b/src/transports/winhttp.c
@@ -1083,7 +1083,7 @@ replay:
 			/* TODO: extract the username from the url, no payload? */
 			if (t->owner->proxy.credentials) {
 				int cred_error = 1;
-				cred_error = t->owner->proxy.credentials(&t->proxy_cred, t->owner->proxy.url, NULL, allowed_types, NULL);
+				cred_error = t->owner->proxy.credentials(&t->proxy_cred, t->owner->proxy.url, NULL, allowed_types, t->owner->proxy.payload);
 
 				if (cred_error < 0)
 					return cred_error;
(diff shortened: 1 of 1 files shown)


src/transports/winhttp.c
```diff
@@ -1072,7 +1083,7 @@ static int winhttp_stream_read(
 			/* TODO: extract the username from the url, no payload? */
 			if (t->owner->proxy.credentials) {
 				int cred_error = 1;
-				cred_error = t->owner->proxy.credentials(&t->proxy_cred, t->owner->proxy.url, NULL, allowed_types, NULL);
+				cred_error = t->owner->proxy.credentials(&t->proxy_cred, t->owner->proxy.url, NULL, allowed_types, t->owner->proxy.payload);
```
ethomson: I don't quite understand where this change is coming from yet.  I don't immediately see anything that's using a payload with regards to the digest changes...  Is this a related change, or should it be split out into its own commit?

cbargren: Without the payload, the `credentials` call was segfaulting in https://github.com/nodegit/nodegit. It isn't directly related to digest credentials, but without it, the credentials callback isn't working. I'll go ahead and split it into its own commit.

ethomson: Got it.  Thanks for clarifying; I do appreciate that you split this out into its own commit.

PR #4109, opened 2017-02-04
Flag checkout_opts in git_reset as optional

rcjsuen:
The documentation for `git_reset` doesn't state that `check-out_opts` can be optional in its documentation. However, if you look at a [test file](https://github.com/libgit2/libgit2/blob/9bc8c80ffa3d20e958406a104c521e2aae0f1255/tests/reset/hard.c#L74) you will see that `NULL` is used everywhere.

The header file should be updated to reflect that `check_opts` is actually an optional argument.

### DIFF
--- a/include/git2/reset.h
+++ b/include/git2/reset.h
@@ -53,7 +53,7 @@ typedef enum {
  *
  * @param reset_type Kind of reset operation to perform.
  *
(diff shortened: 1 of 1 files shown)


ethomson:
Thanks!  I really appreciate you going through the docs here - these are things that really help people get started with using the library.  ❤️

PR #4110, opened 2017-02-07
pkgconfig: do not quote Libs

ignatenkobrain:
It doesn't make sense at all.

### DIFF
--- a/libgit2.pc.in
+++ b/libgit2.pc.in
@@ -6,7 +6,7 @@ Name: libgit2
 Description: The git library, take 2
 Version: @LIBGIT2_VERSION_STRING@
 
-Libs: -L"${libdir}" -lgit2
+Libs: -L${libdir} -lgit2
 Libs.private: @LIBGIT2_PC_LIBS@
 Requires.private: @LIBGIT2_PC_REQUIRES@
 


bapt:
:+1: we have the same issue on FreeBSD

pks-t:
Well, I'm not too sure about this. While I'd also say that the quotes should in fact not be necessary, they have been explicitly added #3569. This is related to Windows build failures with rugged, see libgit2/rugged#559 and 835be6c.

Any comments, @arthurschreiber @mscharlye?

mscharley:
My only comment is that it's not only necessary on Windows, just more likely to be an issue there. GCC and friends (at least under mingw) treat this file as a list of shell arguments, hence the quoting to ensure that paths with spaces or other similar meta characters in them work correctly. Perhaps a solution that might work is moving the open quote to the start of the parameter instead of starting the quote half-way through the option which is something that works in shell, but may not work so well in other systems.

If that still doesn't help, it seems like an incompatibility between different tool chains and I'm not really proficient with C tool chains so I'm not sure how much more help I can be.

pks-t:
A short test shows that we shouldn't actually require quotes at all:

```
cat >/tmp/whitespace.pc <EOF
prefix=/usr
exec_prefix=${prefix}
libdir="${exec_prefix}/white space/lib"

Name: whitespace test
Description: whitespace description
Version: 1
Libs: -L${libdir}
EOF
```

On command line:

```
$ PKG_CONFIG_PATH=/tmp pkg-config --libs whitespace
-L/usr/white\ space/lib
```

So pkg-config should automatically escape whitespace here, if I'm not mistaken.

ethomson:
@pks-t What platform did you just test?  Is that cygwin?

mscharley:
Testing with mingw:

```
libdir="${exec_prefix}/white space/lib"
Libs: -L${libdir}
# Gives...
$ pkg-config --libs whitespace
-L/usr/white\ space/lib
```

```
libdir=${exec_prefix}/white space/lib
Libs: -L${libdir}
# Gives...
C:\Ruby22-x64\DevKit-mingw64-64-4.7.2\tmp $ pkg-config --libs whitespace
-L/usr/white space/lib
```

```
libdir=${exec_prefix}/white space/lib
Libs: -L"${libdir}"
# Gives...
C:\Ruby22-x64\DevKit-mingw64-64-4.7.2\tmp $ pkg-config --libs whitespace
-L/usr/white\ space/lib
```

Looks like the right solution could be to move the quoting to the `libdir` declaration instead of deleting it entirely - seems to work for everyone.

pks-t:
I agree that quoting libdir is the right thing to do here

ethomson:
@ignatenkobrain  Are you able to make this change to quote libdir and test on your platform?

ethomson:
Closing in favor of #4193

Issue #4106, opened 2017-02-03
Infinite loop on windows when libgit2 calls git_repository_discover() for a drive other than C:


eugene2k:
This looks to be a continuation of #387

pks-t:
Can you please confirm that #4111 fixes this?

eugene2k:
The fix works! Sorry for the delay.

ethomson:
Thanks, @eugene2k !

PR #4111, opened 2017-02-07
dirname with DOS prefixes

pks-t:
This is part 1/2 of the fix to #4106. I already debugged into the issue and have a fix in mind, but I'm currently stuck at a copy shop and cannot fire up my Windows VM to actually test it. I'll post the second part tomorrow.

The actual problem fixed here is that e.g. `git_path_dirname_r("C:/")` returns ".", which is simply wrong. It has been fixed to instead return "C:/". The same for network names ("//networkname/").


pks-t:
Okay, wrapped this up now. This fixes #4106

pks-t:
By the way, I'm open to bikeshedding on the `win32_path_prefix` interface. I'm not quite keen on its name and behavior, but got no other ideas here.

ethomson:
Thanks for fixing this.  😀

PR #4112, opened 2017-02-07
Add support for lowercase proxy environment variables

cbargren:
curl supports HTTPS_PROXY in addition to https_proxy (and their http counterparts). This change ensures parity with curl's behavior.

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -774,6 +774,18 @@ int git_remote__get_http_proxy(git_remote *remote, bool use_ssl, char **proxy_ur
 	error = git__getenv(&val, use_ssl ? "HTTPS_PROXY" : "HTTP_PROXY");
 
 	if (error < 0) {
+		if (error != GIT_ENOTFOUND) {
+			return error;
+		}
+
+		giterr_clear();
+		error = 0;
+	}
+
+	/* try lowercase environment variables */
+	error = git__getenv(&val, use_ssl ? "https_proxy" : "http_proxy");
+
+	if (error < 0) {
 		if (error == GIT_ENOTFOUND) {
 			giterr_clear();
 			error = 0;
--- a/src/remote.c
+++ b/src/remote.c
@@ -773,10 +773,9 @@ int git_remote__get_http_proxy(git_remote *remote, bool use_ssl, char **proxy_ur
 	/* http_proxy / https_proxy environment variables */
 	error = git__getenv(&val, use_ssl ? "https_proxy" : "http_proxy");
 
-	if (error == GIT_ENOTFOUND) {
-		/* try uppercase environment variables */
+	/* try uppercase environment variables */
+	if (error == GIT_ENOTFOUND)
 		error = git__getenv(&val, use_ssl ? "HTTPS_PROXY" : "HTTP_PROXY");
-	}
 
 	if (error < 0) {
 		if (error == GIT_ENOTFOUND) {
--- a/src/remote.c
+++ b/src/remote.c
@@ -770,20 +770,13 @@ int git_remote__get_http_proxy(git_remote *remote, bool use_ssl, char **proxy_ur
 		goto found;
 	}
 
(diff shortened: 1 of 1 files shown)


src/remote.c
```diff
@@ -774,6 +774,18 @@ int git_remote__get_http_proxy(git_remote *remote, bool use_ssl, char **proxy_ur
 	error = git__getenv(&val, use_ssl ? "HTTPS_PROXY" : "HTTP_PROXY");
 
 	if (error < 0) {
+		if (error != GIT_ENOTFOUND) {
+			return error;
+		}
+
+		giterr_clear();
+		error = 0;
+	}
```
pks-t: Wouldn't it be easier to just have something like

```
if (error == GIT_ENOTFOUND)
        error = git__getenv(&val, use_ssl ? "https_proxy" : "http_proxy");
```

This would avoid the duplicated logic around `giterr_clear()`.

src/remote.c
```diff
@@ -1374,7 +1386,7 @@ static int update_tips_for_spec(
 			continue;
 
 		/* In autotag mode, don't overwrite any locally-existing tags */
-		error = git_reference_create(&ref, remote->repo, refname.ptr, &head->oid, !autotag, 
+		error = git_reference_create(&ref, remote->repo, refname.ptr, &head->oid, !autotag,
```
pks-t: Given that this is a few hundred lines away from the actual change, I'd avoid unneccessary churn here.

ethomson:
Oh man.  The proxy environment variable mess.

So we are obviously _completely_ wrong in our current implementation and we should be ashamed of ourselves.

Some interesting notes:

> ...and their http counterparts...

It does not.  curl supports _every other_ protocol in both upper and lower case but explicitly _not_ `HTTP_PROXY`.  From `curl(1)`:

> The environment variables can be specified in lower case or upper case.
> The lower case version has precedence. http_proxy is an exception as it
> is only available in lower case.

Meaning, it supports `https_proxy` and `HTTPS_PROXY`, but for http supports only `http_proxy`.  This is for backcompat with lynx and/or libwww (I don't honestly know offhand who loaded the proxy environment variable and I'm certain that I'm not going to go find a copy of the sources :trollface: ).

This is true as of my curl 7.43.0, anyway.

But I'm really saying this only to complain.  Unless somebody can give me a compelling reason why we shouldn't support `HTTP_PROXY` (all caps) then I don't care about strict compatibility here because this whole mess is quite dumb already.

However, I do think that we should honor their preference for `http_proxy` first and `HTTP_PROXY` second, if you don't mind flipping the priority in which you look these things up.

src/remote.c
cbargren: Good call, done.

src/remote.c
cbargren: Right... Atom loves to automatically fix extra spaces for me. :)

cbargren:
I'm not surprised that this stuff is even more complicated that it appeared at first glance. :joy: I'm more than happy to flip the preference here. 👍

src/remote.c
```diff
@@ -770,8 +770,13 @@ int git_remote__get_http_proxy(git_remote *remote, bool use_ssl, char **proxy_ur
...
 
-	/* HTTP_PROXY / HTTPS_PROXY environment variables */
-	error = git__getenv(&val, use_ssl ? "HTTPS_PROXY" : "HTTP_PROXY");
+	/* http_proxy / https_proxy environment variables */
+	error = git__getenv(&val, use_ssl ? "https_proxy" : "http_proxy");
+
+	if (error == GIT_ENOTFOUND) {
+		/* try uppercase environment variables */
+		error = git__getenv(&val, use_ssl ? "HTTPS_PROXY" : "HTTP_PROXY");
+	}
```
pks-t: One minor style nit: we generally avoid braces around one-liners

cbargren: Ah, right, sorry bout that. Force of habit. :)

ethomson: :+1:

PR #4113, opened 2017-02-08
[WIP] Add support for reading attributes out of trees

brianmario:
Finally opening this to get a sense of direction for this functionality, and the changes.

The idea here is to allow reading `.gitattributes` files out of tree objects. For now, these new functions assume that there is no working directory to check. Meaning if a tree oid is specified, a local checkout is never looked at for `.gitattributes` files. Though we may want to allow for both? So, specifying a tree oid would be additive. This was also built assuming the tree oid passed is the root tree, but I don't think it actually matters.

I'm not super happy with adding the new parameter to all of these calls, so we could definitely hang it off of a couple of the existing structs being passed around (and through) the attr APIs. That said, I'm not sure which one(s) make the most sense so I did this for now to get things initially working.

I have a test program locally that links to libgit2, and it appears to work as expected. Though this clearly needs more work.

- [ ] Decide on final API and functionality
- [ ] Make sure all other places that use git attributes internally (diffs?) allow for passing [through] a tree object to read from.
- [ ] Add tests once the API settles down
- [ ] Add/update documentation

cc @mclark @tma @arthurschreiber


brianmario:
Closing this for now as determined by further research that we may not need this.

brianmario:
Turns out we do need this ;)

We'd attempted to use the read tree functionality of the index API, which then let us use the attribute API (which looks in the index). But reading and entire tree structure into the index can be really slow, especially considering we don't care about the _entire_ tree, just the paths for the files we're interested in - and if they have `.gitattributes` files that apply. So, being able to start with a tree (without first reading it into the index) would be a pretty massive optimization.

brianmario:
After thinking about this some more with some folks, we're going to look into taking a more efficient approach which will only look at trees related to the paths we want attributes for. That way we don't have to walk every tree and look at every entry for a file path nested 4 levels deep. With the new approach, we'll only open those 4 tree objects.

src/attr.c
```diff
@@ -83,6 +84,55 @@ int git_attr_get(
 	return error;
 }
 
+int git_attr_get_by_tree(
```
arthurschreiber: While binding this in rugged I noticed that the `repo` parameter here is not really necessary, we can always deduce the repo from the tree object.

ethomson:
We support reading attributes out of trees now!

PR #4114, opened 2017-02-08
Expose the git_repository_set_odb function

hishnash:
This can be usefull when setting a custom odb backend while keeping a fully configured fs refdb.

### DIFF
--- a/include/git2/repository.h
+++ b/include/git2/repository.h
@@ -50,6 +50,17 @@ GIT_EXTERN(int) git_repository_open(git_repository **out, const char *path);
 GIT_EXTERN(int) git_repository_wrap_odb(git_repository **out, git_odb *odb);
 
 /**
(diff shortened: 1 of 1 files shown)


ethomson:
This is already exposed:  https://github.com/libgit2/libgit2/blob/master/include/git2/sys/repository.h#L93

hishnash:
thats embarrassing sorry

ethomson:
😀 

Not for you, it's not.  It's embarrassing for us - obviously, we have not done a good job of pointing out to people that the `include/sys` directory exists and has good stuff in it.

I'll 🤔 about how we might be able to make that better, but if you have ideas, I'd love to hear them.

hishnash:
There is not many docs on custom backends but that fine since its very much a power user use case. When we are done with our one we will open source along with some blog articles to doc the process.

PR #4115, opened 2017-02-08
Changes to provide option to turn off/on ofs_delta

gsaralms:
This change provides an option in git_libgit2_opt_t which can be used in git_libgit2_opts to turn off/on ofs_delta capability in libGit2

This is related to Issue #1414 in libgit2Sharp repository
https://github.com/libgit2/libgit2sharp/issues/1414

### DIFF
--- a/include/git2/common.h
+++ b/include/git2/common.h
@@ -178,6 +178,7 @@ typedef enum {
 	GIT_OPT_ENABLE_STRICT_SYMBOLIC_REF_CREATION,
 	GIT_OPT_SET_SSL_CIPHERS,
 	GIT_OPT_GET_USER_AGENT,
+	GIT_OPT_ENABLE_OFS_DELTA,
 } git_libgit2_opt_t;
 
 /**
@@ -304,6 +305,16 @@ typedef enum {
  *		> Set the SSL ciphers use for HTTPS connections.
  *		>
  *		> - `ciphers` is the list of ciphers that are eanbled.
+ *
+ *	* opts(GIT_OPT_ENABLE_OFS_DELTA, int enabled)
+ *
+ *		> Enable or disable the use of "offset deltas" when creating packfiles,
+ *		> and the negotiation of them when talking to a remote server.
+ *		> Offset deltas store a delta base location as an offset into the
+ *		> packfile from the current location, which provides a shorter encoding
+ *		> and thus smaller resultant packfiles.
+ *		> Packfiles containing offset deltas can still be read.
+ *		> This defaults to enabled.
  *
  * @param option Option key
  * @param ... value to set the option
--- a/src/transports/smart_protocol.c
+++ b/src/transports/smart_protocol.c
@@ -19,6 +19,8 @@
 /* The minimal interval between progress updates (in seconds). */
 #define MIN_PROGRESS_UPDATE_INTERVAL 0.5
 
+bool git_smart__ofs_delta_enabled = true;
+
 int git_smart__store_refs(transport_smart *t, int flushes)
 {
 	gitno_buffer *buf = &t->buffer;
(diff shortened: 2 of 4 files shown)


gsaralms:
@ethomson : please take a look at this PR
this is for https://github.com/libgit2/libgit2sharp/issues/1414

include/git2/common.h
```diff
@@ -305,6 +306,9 @@ typedef enum {
  *		>
  *		> - `ciphers` is the list of ciphers that are eanbled.
  *
+ *	* opts(GIT_OPT_ENABLE_OFS_DELTA, int enabled)
+ *
+ *		> Enable or disable ofs_delta capability.
```
ethomson: I would add some more documentation here.  Here's an example (though I'm not tied to this):

> Enable or disable creating "offset deltas" when creating packfiles.  Offset deltas store a delta base location as an offset into the packfile from the current location, which provides a shorter encoding and thus smaller resultant packfiles.  Packfiles containing offset deltas can still be read.  This defaults to enabled.

gsaralms: 👍  will update the comment

ethomson: Sorry, I now realize that I was rather misunderstanding the goal.  It's not to _create_ them, it's to _request_ them to be created from the server (or not).  In that case, given the broad name given to this option, I would say either:

1. This single option controls whether we _create_ offset deltas _and_ will override whether we _request_ offset deltas.  If we do this, I think that the documentation should read something like:

    > Enable or disable the use of "offset deltas" when creating packfiles, and the negotiation of them when talking to a remote server.  Offset deltas store a delta base location as an offset into the packfile from the current location, which provides a shorter encoding and thus smaller resultant packfiles. Packfiles containing offset deltas can still be read. This defaults to enabled.

2. Separate options for each of these two scenarios.  In that case, I would change the name of this option to something like `GIT_OPT_ENABLE_REMOTE_OFS_DELTA`, and document it something like:

> Enable or disable requesting "offset deltas" when communicating with a remote.  Offset deltas store a delta base location as an offset into the packfile from the current location, which provides a shorter encoding and thus smaller resultant packfiles.  This defaults to enabled.

Note that if we go with (1), there's no more coding work to do...  it appears that we simply don't support creating packfiles with ofs_deltas?  (I didn't realize this until now, and somebody should sanity check me on it.)

src/transports/smart_protocol.c
```diff
@@ -19,6 +19,8 @@
 /* The minimal interval between progress updates (in seconds). */
 #define MIN_PROGRESS_UPDATE_INTERVAL 0.5
 
+bool git_ofs_delta__enabled = true;
```
ethomson: This name doesn't quite make sense in our naming scheme -- we try to scope things to something sort of like an "object" or - when that stops making sense - the file level.  (I concede that given the filename, `smart_protocol.c`, this has stopped making sense here, since this was logical outgrowth from `smart.c` which I guess just got too big).  So `git_smart` is the object, or concept here, and _it's_ the prefix.  Anything that applies to that scope should be double underscored.

Something like:  `git_smart__ofs_delta_enabled`.

(If you squint really hard it seems like we're trying to make C look object oriented, which is either terrible or merely odd depending on your mood.)

include/git2/common.h
ethomson: /cc @carlosmn who may have interesting opinions or insights here.

gsaralms: our goal is actually (1) i.e. to simply not have any "ofs_delta".
does this name "GIT_OPT_ENABLE_OFS_DELTA" convey it properly ?
any better name suggestions?

src/transports/smart_protocol.c
gsaralms: ok, I was not aware of the naming scheme, will update it to `git_smart__ofs_delta_enabled`

include/git2/common.h
ethomson: I think that this name is scoped nicely to (1), _never create packfiles with offset deltas_.  I also think that this is the sensible scoping for this feature.  As it happens, we will never write packfiles with offset deltas (this is still surprising to me!) so that's an easy win for that part of this support.  😀

gsaralms:
AppVeyor build failed, which does not look related to my changes to re-triggering the build by a close/ re-open of PR

gsaralms:
closed and reopened to trigger a build because failure in appveyor build didn't looked to be related to my changes

gsaralms:
@ethomson : gentle ping on this, I have updated the comment and variable name.
Please take a look

ethomson:
Thanks @gsaralms !   Looks good to me.

PR #4117, opened 2017-02-09
Flag options in git_stash_apply and git_stash_pop as being optional

rcjsuen:
The options parameter in both [git_stash_apply](https://github.com/libgit2/libgit2/blob/0175971e91641cd07246f59003930d81019ce40b/tests/stash/apply.c#L105) and [git_stash_pop](https://github.com/libgit2/libgit2/blob/0175971e91641cd07246f59003930d81019ce40b/tests/stash/apply.c#L307) can be `NULL` as can be seen in the linked tests. The documentation in the `stash.h` header file should reflect this.


ethomson:
👍 Thanks!

PR #4119, opened 2017-02-09
rename detection: don't try to detect submodule renames

ethomson:
Consider the case where some branch deletes a submodule, then adds a new submodule with a different path.

Since these two paths have the same mode, they will be eligible for rename detection, and we will try to load the blobs to analyze their contents.  This will, of course, fail since not only are they not blobs but the objects are in a different repository.

Make sure that the things we're looking at are actually blobs when considering them for rename detection (in diff or merge).


src/diff_tform.c
```diff
@@ -554,7 +554,7 @@ static int similarity_measure(
 	*score = -1;
 
 	/* don't try to compare files of different types */
```
carlosmn: This comment is now a bit out of sync, no?

ethomson: Yes.  Willfix.

PR #4126, opened 2017-02-15
negotiate always fails via libcurl

implausible:
Instead we build a strict ordering for which auth methods we want to use by security and walked down available proxy auth methods. Otherwise we fail outright.

We saw repeated failures to negotiate properly vs a Forefront TMG proxy box configured with various settings: basic auth, digest, NTLM where negotiate was always enabled. After digging, so long as negotiate is available, libcurl will pick that as its auth scheme, but it also fails every time using negotiate.

A somewhat hacky solution was to inspect auth types on the proxy response and switch NTLM, Digest, Basic from that. So that's what this is. Discussion is welcome.

### DIFF
--- a/src/curl_stream.c
+++ b/src/curl_stream.c
@@ -52,6 +52,7 @@ static int apply_proxy_creds(curl_stream *s)
 {
 	CURLcode res;
 	git_cred_userpass_plaintext *userpass;
+	long auth_types;
 
 	if (!s->proxy_cred)
 		return GIT_ENOTFOUND;
@@ -61,6 +62,19 @@ static int apply_proxy_creds(curl_stream *s)
 		return seterr_curl(s);
 	if ((res = curl_easy_setopt(s->handle, CURLOPT_PROXYPASSWORD, userpass->password)) != CURLE_OK)
 		return seterr_curl(s);
+
+	curl_easy_getinfo(s->handle, CURLINFO_PROXYAUTH_AVAIL, &auth_types);
+
+	if (auth_types & CURLAUTH_NTLM) {
+		if ((res = curl_easy_setopt(s->handle, CURLOPT_PROXYAUTH, CURLAUTH_NTLM)) != CURLE_OK)
+			return seterr_curl(s);
+	} else if (auth_types & CURLAUTH_DIGEST) {
+		if ((res = curl_easy_setopt(s->handle, CURLOPT_PROXYAUTH, CURLAUTH_DIGEST)) != CURLE_OK)
+			return seterr_curl(s);
+	} else if (auth_types & CURLAUTH_BASIC) {
+		if ((res = curl_easy_setopt(s->handle, CURLOPT_PROXYAUTH, CURLAUTH_BASIC)) != CURLE_OK)
+			return seterr_curl(s);
+	}
 
 	return 0;
 }


ethomson:
Hmm...  this disables negotiate _completely_?  Are you sure that there are no cases where negotiate works, and that it didn't work for you?

_Should_ it have worked for you?  What platform are you trying this on?  Do you have a Kerberos ticket?  If Windows, are you domain joined?

implausible:
It is _probable_ that this is a result of bad Forefront TMG settings or misstep in our configuration; however, I thought it best to make this PR purely to see your thoughts on this issue.

Tested on Linux and MacOS. We have _not_ tested this against Windows.

I believe using libcurl for negotiate protocol _should_ work, but in testing vs our proxy we were unable to get MacOS or Linux to connect. We spent a day tracking down what was going wrong, bouncing back and forth between the proxy server settings itself and LibGit2 built with libcurl. Libcurl in our app was built with support for Basic, Digest, NTLM, Negotiate, and Kerberos authentication protocols. What ends up occurring is that libcurl was always trying negotiate if it was available and consistently failing to perform a proper negotiation, either encountering a SPNEGO error or getting stuck in a credentials loop.

I just found poxyproxy that you are using to test in libgit2. Is this testing libgit2 via negotiate?

ethomson:
Do you have a Kerberos ticket on the Linux and Mac machines?  Can you sniff a browser session?  Are you doing Negotiate successfully with, say, Firefox?  (I forget which browsers will speak SPNEGO; I think Firefox will, but you may have to turn a configuration setting on for it to do so.)

> I just found poxyproxy that you are using to test in libgit2. Is this testing libgit2 via negotiate?

I have used it to test libgit2 with Negotiate but this does not test Negotiate in our CI workflow.  poxyproxy is capable of doing the stateful keepalive necessary to proxy NTLM or SPNEGO authentication through to the destination (like TMG is), but we aren't testing with NTLM _or_ Negotiate.  Setting up a Kerberos controller is a massive PITA, and wasn't really feasible just for tests.

I have never used the Negotiate code in libcurl (via libgit2).  I have used the Negotiate code in _curl_ before with success, and I have used [_our_ Negotiate code](https://github.com/libgit2/libgit2/blob/master/src/transports/auth_negotiate.c) in libgit2 on Linux and Mac with success.

You may want to turn _off_ libcurl support and use our builtin HTTP transport if you have a valid Kerberos ticket and believe that SPNEGO should work to your web browser.  It would be interesting to know if we're not setting up libcurl properly.

It's rather tragic that we have so many different HTTP transports with different sets of features and different levels of testing, but that's a different conversation.

implausible:
>  the stateful keepalive necessary to proxy NTLM or SPNEGO authentication through to the destination (like TMG is)

Just to clarify our TMG is set up as needing its own authentication separate from the destination endpoint.

When I get home tonight I will use try to use the negotiate code in libgit2. Worries are that we lose authenticated proxy support in MacOS and Linux without Libcurl (this is what I think with my current understanding of how the library works). Is the libgit2 auth_negotiate the appropriate method of authenticating via proxy in LibGit2?

ethomson:
Oh, hmm.  Yes, that's a good point.  🤔

We don't do so well with proxies (again with the different sets of features for the different transports).

I don't have a really great idea here offhand.  I would really love to throw some code at the wall and see what sticks, but unfortunately, I don't have a proxy that speaks negotiate these days.

Let me see if I can whip one up.

ethomson:
No more libcurl!  I'm not sure that means that everything magically works now but I hope so!  😁

implausible:
I'm so excited for that! Thank you so much for getting rid of the curl dependency! So many build problems in NodeGit / GitKraken about to go away.

PR #4130, opened 2017-02-17
Improve clar messages

ethomson:
Provide more detailed messages when conditions pass or fail unexpectedly.  In particular, this provides the error messages when a test fails with a different error code than was expected.

`cl_git_pass` failures look unchanged:
```
  1) Failure:
apply::fromfile::test_name [tests/apply/fromfile.c:42]
  Function call failed: (git_patch_from_buffer(&patch, "", 0, ((void *)0)))
  error -3 - no patch found
```

`cl_git_fail` messages now:
```
  1) Failure:
apply::fromfile::test_name [tests/apply/fromfile.c:42]
  Function call succeeded: git_patch_from_buffer(&patch, patchfile, strlen(patchfile), NULL)
  no error, expected non-zero return
```

`cl_git_fail_with` messages now:
```
  1) Failure:
apply::fromfile::test_name [tests/apply/fromfile.c:42]
  Function call failed: (git_patch_from_buffer(&patch, "", 0, ((void *)0)))
  error -3 (expected -99) - no patch found
```

/cc @pks-t

### DIFF
--- a/src/unix/posix.h
+++ b/src/unix/posix.h
@@ -50,7 +50,7 @@ extern char *p_realpath(const char *, char *);
 #define p_strcasecmp(s1, s2) strcasecmp(s1, s2)
 #define p_strncasecmp(s1, s2, c) strncasecmp(s1, s2, c)
 #define p_vsnprintf(b, c, f, a) vsnprintf(b, c, f, a)
-#define p_snprintf(b, c, f, ...) snprintf(b, c, f, __VA_ARGS__)
+#define p_snprintf(b, c, ...) snprintf(b, c, __VA_ARGS__)
 #define p_mkstemp(p) mkstemp(p)
 #define p_chdir(p) chdir(p)
 #define p_chmod(p,m) chmod(p, m)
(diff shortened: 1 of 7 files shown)


ethomson:
The goal here is to see more information about why some tests fail in appveyor.  We expect certificate failures, but are getting a generic `-1` failure.  Shoddy networking, but it would be interesting to see a bit more details.

pks-t:
Very nice.

You forgot to modfiy all existing callsites of `cl_git_report_failure` ;)

What do you think about using `cl_git_expect` instead of `cl_git_exec` as the macro's name?

ethomson:
I did forget win32!  Oops!

I had this as `cl_git_expect` or something similar to begin with, so I was rather on the fence.  It sounds like you prefer that, so I've changed `cl_git_exec` -> `cl_git_expect`.

pks-t:
`cl_git_exec` did not really sound as if it would check anything but only execute a function. So I think `cl_git_expect` makes the function a bit more obvious in its behavior.

Thanks, looks good to me.

PR #4133, opened 2017-02-21
fix regression from #4092

hackhaslam:
This is a crash on 32-bit and I assume that it doesn't do the right thing on
64-bit either. MSVC emits a warning for this, but of course, it's easy to get
lost among all of the similar 'possible loss of data' warnings.

### DIFF
--- a/src/pack.c
+++ b/src/pack.c
@@ -128,13 +128,13 @@ static git_pack_cache_entry *cache_get(git_pack_cache *cache, git_off_t offset)
 /* Run with the cache lock held */
 static void free_lowest_entry(git_pack_cache *cache)
 {
+	git_off_t offset;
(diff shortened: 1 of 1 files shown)


pks-t:
Agreed. I somehow overlooked the fact that the `map_foreach` function does not loop over the indices anymore but instead over values. Interesting that tests didn't catch this issue until now.

Could you please use a more descriptive commit message? It becomes somewhat hard to read when you e.g. do a `git log`, as the user would first need to check #4092 to see what was really changed. Something along the lines of `pack: fix looping over cache entries` would be much easier to read here :)

Thanks for the fix! Will merge after the amended message

hackhaslam:
Okay, I changed the commit message.

pks-t:
Thanks a lot! :+1:

PR #4135, opened 2017-02-22
Provide error on gcc < 4.1 when THREADSAFE

ethomson:
Provide a descriptive error message when compiling `THREADSAFE` on gcc versions < 4.1.  We require the atomic primitives (eg `__sync_synchronize`) that were introduced in that version.

(Note, clang setes `__GNUC__` but appears to set its version > 4.1.)

### DIFF
--- a/src/thread-utils.h
+++ b/src/thread-utils.h
@@ -7,6 +7,12 @@
 #ifndef INCLUDE_thread_utils_h__
 #define INCLUDE_thread_utils_h__
 
+#if defined(__GNUC__) && defined(GIT_THREADS)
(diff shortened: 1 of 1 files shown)


carlosmn:
>(Note, lldb setes __GNUC__ but appears to set its version > 4.1.)

Do you mean llvm/clang? The debugger surely isn't setting the defines.

ethomson:
> Do you mean llvm/clang? The debugger surely isn't setting the defines.

Yes, clearly.  I've updated the commit message to reflect this.

carlosmn:
The Travis build finished after 25m but it looks like it never sent out the notification.

PR #4136, opened 2017-02-24
Introduce (optional) SHA1 collision attack detection

ethomson:
Include SHA1 collision attack detection from https://github.com/cr-marcstevens/sha1collisiondetection when configured with `-DUSE_SHA1DC=1`.

This is the same implementation being proposed for git core; see https://public-inbox.org/git/20170223230536.tdmtsn46e4lnrimx@sigill.intra.peff.net/


pks-t:
The imported code is not ISO C90 compatible, there's a declaration after code in `src/hash/sha1dc/sha1.c:55`.

Otherwise this looks fine, I guess (even though I obviously didn't skim through the SHA1DC implementation).

ethomson:
Dang, thanks @pks-t , I missed that.

ethomson:
AppVeyor's mingw builds are busted.  Sigh.  👀

ethomson:
All fixed up!

tests/core/sha1.c
```diff
@@ -0,0 +1,70 @@
...
+	git_hash_ctx_cleanup(&ctx);
+
+	return ret;
+}
+
+#define expect_id(idstr, oid) { \
+	git_oid __expected; \
+	git_oid_fromstr(&__expected, (idstr)); \
+	cl_assert_equal_oid(&__expected, (oid)); \
+}
```
pks-t: This #define is currently unused?

CMakeLists.txt
```diff
@@ -292,13 +293,16 @@ ELSE ()
...
 # Specify sha1 implementation
-IF (WIN32 AND NOT MINGW AND NOT SHA1_TYPE STREQUAL "builtin")
-	ADD_DEFINITIONS(-DWIN32_SHA1)
+IF (USE_SHA1DC)
+	ADD_DEFINITIONS(-DGIT_SHA1_COLLISIONDETECT)
+	FILE(GLOB SRC_SHA1 src/hash/hash_collisiondetect.c src/hash/sha1dc/*)
+ELSEIF (WIN32 AND NOT MINGW AND NOT SHA1_TYPE STREQUAL "builtin")
+	ADD_DEFINITIONS(-DGIT_SHA1_WIN32)
 	FILE(GLOB SRC_SHA1 src/hash/hash_win32.c)
 ELSEIF (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
```
pks-t: By the way, is it correct that OS X does not check for SHA1_TYPE=="builtin"? Sounds as if it wasn't possible for OS X systems to build with our generic hashing algorithm.

ethomson: Good question...  digging in a bit deeper, it looks like one can't actually _set_ `SHA1_TYPE`.  So I'm removing these tests for `builtin` entirely.

tests/core/sha1.c
ethomson: Yep, good 👁 .  I've dropped this.

CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ OPTION( PROFILE				"Generate profiling information"		OFF )
 OPTION( ENABLE_TRACE		"Enables tracing support"				OFF )
 OPTION( LIBGIT2_FILENAME	"Name of the produced binary"			OFF )
 
+OPTION( USE_SHA1DC			"Use sha1 with collision detection"		OFF )
```
carlosmn: s/sha1/SHA-1/ ?

tests/core/sha1.c
```diff
@@ -0,0 +1,64 @@
...
+
+/* test that sha1 collision detection works when enabled */
+void test_core_sha1__detect_collision_attack(void)
+{
+	git_oid oid, expected;
+
+#ifdef GIT_SHA1_COLLISIONDETECT
+	GIT_UNUSED(expected);
+	cl_git_fail(sha1_file(&oid, FIXTURE_DIR "/shattered-1.pdf"));
```
carlosmn: Do we want to have a specific error code for this as well, so we can do `cl_git_fail_with(GIT_ESHA1ATTACK, ...)`? I'm not sure if this adds that much value, just a thought.

ethomson: 🤔

CMakeLists.txt
ethomson: Done

tests/core/sha1.c
ethomson: So...  my thinking here is that this is only useful if it's something that you want to act on.  I think that if you were a hosting provider that you would actually _not_ want to bother logging these, because once this becomes reasonably cheap and you can create two git objects that have colliding hashes -- or when a research org decides to extend their corpus of files with colliding hashes to git objects and not just PDFs -- then people are going to go crazy uploading them to poke at you.  And if you're a client, then this isn't useful knowledge to have at all.

But it's quite easy to implement, so I'm happy to add it if you're passionate about it.

carlosmn: I just figured a client might want to throw a particular prompt, but it's so far beyond what we have control over that it's not really worth it.

ethomson:
It looks like the git authors and the authors of this SHA-1 collision detection library are interested in optimizing it; we'll revisit this if/when they settle on something more efficient.

PR #4137, opened 2017-02-24
appveyor: don't rewrite the system mingw

ethomson:
Trying to chase down the new appveyor build failures...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,6 +2,7 @@ version: '{build}'
 branches:
   only:
   - master
+  - appveyor
   - /^maint.*/
 environment:
   GITTEST_INVASIVE_FS_STRUCTURE: 1
--- a/script/appveyor-mingw.sh
+++ b/script/appveyor-mingw.sh
@@ -7,15 +7,17 @@ if [ "$ARCH" = "i686" ]; then
     curl -LsSO http://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/4.9.2/threads-win32/sjlj/$f
   fi
   7z x $f > /dev/null
-  mv mingw32 /MinGW
+  export PATH=`pwd`/mingw32/bin:$PATH
 else
   f=x86_64-4.9.2-release-win32-seh-rt_v3-rev1.7z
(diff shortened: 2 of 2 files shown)


ethomson:
ARGH, this didn't even queue an appveyor build?!  😡

tkelman:
what do the failures look like?

ethomson:
@tkelman We used to download newer mingw-w64's and overwrite `/mingw` with them.  I don't know why.  I'm trying to make sure that we download the newer mingw-w64's and _then run them from a neutral place_.

ethomson:
Sorry, I realize I didn't answer your question.  The error was with moving our `mingw` into `/mingw`; it failed with a permissions problem, which seems pretty reasonable, to be honest.

tkelman:
so it does, guess that was good enough at the time and avoided needing to modify path. appveyor may have changed permissions or what's installed by default?

appveyor.yml
```diff
@@ -2,6 +2,7 @@ version: '{build}'
 branches:
   only:
   - master
+  - appveyor
```
tkelman: it should build pr's if the webhook is configured right, there may have just been a delivery failure?

I think without a regex this would need to be an exact match anyway

ethomson: Yes - that's correct; there was just a weird webhook failure here.  But it encouraged me to add the `appveyor` branch so that we could build `appveyor` in isolation and not burn unnecessary travis builds.

ethomson:
> so it does, guess that was good enough at the time and avoided needing to modify path. appveyor may have changed permissions or what's installed by default?

Yes, that's right - AppVeyor updated its mingw installation, and I think changed permissions then, which is why we noticed this.

pks-t:
Is it even reasonable to always re-download MinGW? Shouldn't we just be using whatever AppVeyor is currently providing?

ethomson:
AppVeyor only provides the 32 bit mingw-w64.  There's an issue to provide the x86_64 version as well.  Once they do, I think we should reconsider using the installed versions.

tkelman:
the download should be cached here

pks-t:
Ah okay, that explains it. Thanks!

PR #4138, opened 2017-02-24
tests: Add create__symbolic_with_arbitrary_content

richardipsum:
This test ensures that it's possible to create a symbolic ref that
has arbitrary data as its target. It also ensures it's possible
to obtain the target of that symbolic reference from the git_reference
object.

### DIFF
--- a/tests/refs/create.c
+++ b/tests/refs/create.c
@@ -12,12 +12,12 @@ static git_repository *g_repo;
 
 void test_refs_create__initialize(void)
 {
-   g_repo = cl_git_sandbox_init("testrepo");
+	g_repo = cl_git_sandbox_init("testrepo");
 }
 
 void test_refs_create__cleanup(void)
 {
-   cl_git_sandbox_cleanup();
+	cl_git_sandbox_cleanup();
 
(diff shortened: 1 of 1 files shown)


ethomson:
The AppVeyor mingw builds are busted at the moment.  I'm looking into it; please ignore build failures from those hosts for now.

richardipsum:
I noticed some whitespace/style issues while I was doing that so I fixed them up and tacked them on as an extra commit.

For what it's worth while rerunning the test suite locally, just to be sure, I notice that the clar tests fail if run via the Makefile but not when run directly, this happens on master as well so I'm pretty sure it's unrelated to my changes.

I have some free time today so I might check out why that's happening.

richardipsum:
I think the tests are failing due to networking issues at my end, so nothing to worry about. I needed to do some googling to figure out how to get the tests to print the exact failure to the screen without manually modifying the generated Makefile. I've submitted https://github.com/libgit2/libgit2/pull/4139 to add a note to the README on how to do this.

ethomson:
Cool, thanks!  😀

PR #4141, opened 2017-02-25
Fix inet_pton tests triggering an assert in Haiku

kainjow:
Haiku will assert in a nightly build if the "dst" input to inet_pton() is NULL.


ethomson:
Thanks for catching this; there's definitely some platform specific assumptions in those tests!

Issue #4094, opened 2017-01-26
reflog doesn't include "(merge)" for merge commits


rcjsuen:
When you use the command line, merge commits will have `commit (merge):` recorded in the reflog as a reminder that a merge commit was created. libgit2 doesn't do this though. It only handles the simple `commit:` case and the first `commit (initial):` case.

### Reproduction steps
1. Create a merge commit using libgit2's APIs.
2. Check the output of `git reflog`.
### Expected behavior
`cf97c9d HEAD@{0}: commit (merge): Merge branch 'other'`
### Actual behavior
`d50f21e HEAD@{0}: commit: Merge branch 'other'`
### Version of libgit2 (release number or SHA1)
Looks like the code is still the same on 8d3b39a6ad325fe1d08e17281803e26359756fc6.
https://github.com/libgit2/libgit2/blob/8d3b39a6ad325fe1d08e17281803e26359756fc6/src/refs.c#L1148-L1153
### Operating system(s) tested
Windows 10

ethomson:
Thanks for the report @rcjsuen !

This seems like a straightforward fix, I'm going to add the "Easy Fix" and "Up for Grabs" tags in case someone wants to submit a pull request!

ratlifri:
I was looking here at #4094 this week and was coming toward a similar conclusion but above fix is better than mine.  I am was happy to learn I was in the right spot and looking to change the same field.  I added above fix into v0.25.1 (2fcb870) and all tests passed on Windows 7.

I looked at adding a new test and found that reflog is not tested very much, at least in this fashion, for anything besides "commit (initial)" which failed only four times after breaking on purpose, 3 in rebase::merge and 1 in commit::write.  What I think I learned about a merge commit from testing point of view is it crosses unit test boundaries since merge puts repo in a merge state and until the commit completes.

richardipsum:
Shouldn't this issue have been automatically closed by github once https://github.com/libgit2/libgit2/pull/4143 got merged?

ethomson:
I don't know how the automatic matching works, exactly, but I'll close it manually.

PR #4143, opened 2017-02-26
Fix: make reflog include "(merge)" for merge commits

richardipsum:
This fixes issue #4094

### DIFF
--- a/src/refs.c
+++ b/src/refs.c
@@ -1130,6 +1130,18 @@ int git_reference__update_terminal(
 	return error;
 }
 
+static const char *commit_type(const git_commit *commit)
+{
+	unsigned int count = git_commit_parentcount(commit);
+
+	if (count >= 2)
+		return " (merge)";
+	else if (count == 0)
+		return " (initial)";
+	else
+		return "";
+}
+
 int git_reference__update_for_commit(
 	git_repository *repo,
 	git_reference *ref,
@@ -1146,7 +1158,7 @@ int git_reference__update_for_commit(
 	if ((error = git_commit_lookup(&commit, repo, id)) < 0 ||
 		(error = git_buf_printf(&reflog_msg, "%s%s: %s",
 			operation ? operation : "commit",
-			git_commit_parentcount(commit) == 0 ? " (initial)" : "",
+			commit_type(commit),
 			git_commit_summary(commit))) < 0)
 		goto done;
 
--- a/tests/refs/reflog/reflog.c
+++ b/tests/refs/reflog/reflog.c
@@ -4,7 +4,7 @@
 #include "git2/reflog.h"
 #include "reflog.h"
 
-
+static const char *merge_reflog_message = "commit (merge): Merge commit";
 static const char *new_ref = "refs/heads/test-reflog";
 static const char *current_master_tip = "a65fedf39aefe402d3bb6e24df4d4f5fe4547750";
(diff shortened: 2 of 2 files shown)


src/refs.c
```diff
@@ -1130,6 +1130,18 @@ int git_reference__update_terminal(
 	return error;
 }
 
+const char *git_reference__commit_type(const git_commit *commit)
```
pks-t: This function is used inside of this file only and will probably never be exposed to other callers, as it is very specific. As such, this function should be marked as `static` as well as drop the `git_reference__` prefix.

richardipsum: Thanks for the review, I will make these changes.

richardipsum: Ahh, it will take me a little while longer to add some tests, I'll look into what's needed, thanks!

richardipsum: @pks-t Fixed style issues and added a test. Have verified that test fails on revert of this fix.

tests/commit/merge.c
```diff
@@ -0,0 +1,66 @@
+#include "clar_libgit2.h"
```
pks-t: I hate to be nitpicky, but I would have expected the test to be part of `tests/refs/reflog/reflog.c`. Sorry to not state this earlier

tests/commit/merge.c
```diff
@@ -0,0 +1,66 @@
...
+
+	cl_git_pass(merge_branches(_repo,
+		"refs/heads/branchB-1", "refs/heads/branchB-2",
+		&merge_opts, &checkout_opts));
+
+	cl_git_pass(git_repository_index(&index, _repo));
+	cl_git_pass(git_index_read(index, 1));
+	cl_assert_(!git_index_has_conflicts(index), "Index has conflicts!");
+	cl_git_pass(git_index_write_tree(&tree_oid, index));
+	cl_git_pass(git_tree_lookup(&tree, _repo, &tree_oid));
```
pks-t: What we really want to test is the reflog logic. So you could discard the "real" merge logic and simply create a merge commit re-using the tree of one of both commits. This would make the test a lot simpler and more to the point

tests/commit/merge.c
```diff
@@ -0,0 +1,66 @@
...
+
+	cl_git_pass(git_commit_create(&merge_commit_oid,
+		_repo, "refs/heads/branchB-1", s, s, NULL,
+		"Merge commit", tree,
+		2, (const struct git_commit **) parent_commits));
+
+	cl_git_pass(git_reflog_read(&log, _repo, "refs/heads/branchB-1"));
+	entry = git_reflog_entry_byindex(log, 0);
+	cl_assert_equal_s(merge_reflog_message, git_reflog_entry_message(entry));
```
pks-t: You unfortunately have to clean up all allocated memory after the test ;)

tests/commit/merge.c
richardipsum: No worries, it could be simpler I guess. I'll rework this later.

tests/commit/merge.c
richardipsum: Done

tests/commit/merge.c
richardipsum: Done

pks-t:
Looks good to me now! Thanks again for working on it

ethomson:
Thanks, @richardipsum !  🎉

PR #4150, opened 2017-03-03
git_commit_create: freshen tree objects in commit

ethomson:
Freshen the tree object that a commit points to during commit time.  This can help to freshen trees when a new commit is written that points to an old tree (but that tree id was recorded, instead of being newly computed with `git_odb_write`, which would have done the tree freshening itself.)

/cc @carlosmn

### DIFF
--- a/tests/odb/freshen.c
+++ b/tests/odb/freshen.c
@@ -17,34 +17,99 @@ void test_odb_freshen__cleanup(void)
 	cl_git_sandbox_cleanup();
 }
 
-#define LOOSE_STR "hey\n"
-#define LOOSE_ID  "1385f264afb75a56a5bec74243be9b367ba4ca08"
-#define LOOSE_FN  "13/85f264afb75a56a5bec74243be9b367ba4ca08"
+static void set_time_wayback(struct stat *out, const char *fn)
+{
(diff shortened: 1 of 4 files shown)


tests/odb/freshen.c
```diff
@@ -17,34 +17,99 @@ void test_odb_freshen__cleanup(void)
...
+
+	cl_git_pass(git_oid_fromstr(&tree_id, LOOSE_TREE_ID));
+	cl_git_pass(git_tree_lookup(&tree, repo, &tree_id));
+	set_time_wayback(&before, LOOSE_TREE_FN);
 
+	cl_git_pass(git_oid_fromstr(&parent_id, "a65fedf39aefe402d3bb6e24df4d4f5fe4547750"));
+	cl_git_pass(git_commit_lookup(&parent, repo, &parent_id));
+
+	cl_git_pass(git_signature_new(&signature,
+		"Refresher", "refresher@refresh-me.com", 1488547083, 0));
```
carlosmn: So... this domain belongs to a German web/graphic design company

ethomson: Haha.  Right, yes, I probably shouldn't use stream of consciousness for things that are etched in git's commit history.

PR #4151, opened 2017-03-03
rebase: ignore untracked files in submodules

novalis:
An untracked file in a submodule should not prevent a rebase from
starting.  Even if the submodule's SHA is changed, and that file would
conflict with a new tracked file, it's still OK to start the rebase
and discover the conflict later.

Signed-off-by: David Turner <dturner@twosigma.com>


novalis:
Ping?

ethomson:
Sorry for the delay - this looks good to me.  Thanks!

PR #4154, opened 2017-03-06
Support namespaced references again

ethomson:
Support `git_repository_set_namespace` again (the equivalent of `GIT_NAMESPACE`); this was subtly broken with worktree support.

This adds a simple namespace test (equivalent to rugged's) and simplifies the namespace handling by simply appending the namespace (if it exists) to the gitdir and commondir.

/cc @carlosmn

### DIFF
--- a/src/refdb_fs.c
+++ b/src/refdb_fs.c
@@ -1433,26 +1433,25 @@ static void refdb_fs_backend__free(git_refdb_backend *_backend)
 	git__free(backend);
 }
 
-static int setup_namespace(git_buf *gitpath, git_repository *repo)
+static char *setup_namespace(git_repository *repo, const char *in)
 {
-	char *parts, *start, *end;
+	git_buf path = GIT_BUF_INIT;
+	char *parts, *start, *end, *out = NULL;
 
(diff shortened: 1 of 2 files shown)


ethomson:
So there are two things here that probably bear discussion:

1. The interaction between namespaces and worktrees.  If you set a namespace, this will now look for namespaces _in both_.  It's not exactly clear that this is what git itself would do, because git's namespaces are _very_ narrowly applied to "server" functionality.  I don't know that there are commands that obey namespaces and worktrees both.  So I think our implementation is okay, but if there exists a command in the future (or in the present, and I just don't know it) that obeys both then we may need to rethink this.

2. You can create a refdb that is not backed by a filesystem path.  (See `refs::iterator::empty`, for example.)  We handle this (and have since the inception of the refdb) by simply creating paths like `refs/heads/master` (instead of prefixing them with the gitdir).  This is obviously pretty wrong.  I haven't fixed this here, since it's a) a longstanding bug not a regression and b) I want to unblock https://github.com/libgit2/rugged/pull/687.  But I will open a PR to deal with this when I have a few spare minutes.

Issue #3782, opened 2016-05-11
Please support using notes via a commit rather than a ref


joshtriplett:
The various functions that support git notes, such as `git_note_iterator_new`, `git_note_read`, `git_note_create`, and `git_note_remove` all take a `notes_ref` as a string (defaulting to `refs/notes/commits`).  Please consider adding support to use a commit instead of a ref; the functions to read notes would just need the commit, and the functions to write notes would take a notes commit and return a new notes commit.  This would allow storing the notes commit(s) in another way, rather than as a reference.

The existing functions that use a notes ref would then become simple wrappers around the lower-level API.

richardipsum:
@joshtriplett I'd like to help with this if possible. But I don't see how the `notes_ref' parameter can be avoided in the low level calls that would take a commit instead of a notes_ref, if the higher level functions that take a `notes_ref' are to keep their existing functionality. So long as you provide the means to put notes somewhere other than `refs/notes/commits' you will need the `notes_ref' parameter.

joshtriplett:
@richardipsum The functions that read notes would take a commit instead of a notes_ref string, and the existing functions would read the ref, turn it into a commit, and pass that.  The functions that write notes would take a parent commit and return a new commit, and the existing functions would (with appropriate locking) read the ref, turn it into a commit, pass that, get back a commit, and write that to the ref.

richardipsum:
@joshtriplett Right I see, so literally just swapping the `const char *notes_ref` for `git_commit *notes_commit`? Okay I'll give this a try.

joshtriplett:
@richardipsum That and a `git_commit **` out parameter for calls that write notes to provide the new commit they create.

richardipsum:
Ack :smile_cat:

richardipsum:
Hi! Sorry it's taken me so long to get this done. Pull request #4159 implements this feature request and is ready for review. :)

richardipsum:
@carlosmn Hi, do you have any thoughts on my attempt at implementing this feature request?

richardipsum:
@ethomson I implemented the solution to this issue 3 months ago and I've had no feedback at all, it seems as if there's no real desire for this functionality, can I suggest then that we close both the PR and this issue?

joshtriplett:
@richardipsum I'm still really interested in this, and I'm sorry to see that you haven't gotten much response.

richardipsum:
[...]

PR #4159, opened 2017-03-13
Support using notes via a commit rather than a ref

richardipsum:
Hi!

This is a work in progress implementation to solve issue #3782 
each new function will be added in a separate commit.

Finally existing notes functions will be reimplemented in terms of the new functions.

I'm opening this now in case anyone has any suggestions in the early stages
that will allow me to avoid rework later on.

Thanks,
Richard Ipsum

### DIFF
--- a/include/git2/notes.h
+++ b/include/git2/notes.h
@@ -223,6 +223,32 @@ GIT_EXTERN(int) git_note_remove(
 	const git_oid *oid);
 
 /**
+ * Remove the note for an object
+ *
+ * @param notes_commit_out pointer to store the new notes commit (optional);
+ *					NULL in case of error.
+ *					When removing a note a new tree containing all notes
+ *					sans the note to be removed is created and a new commit
+ *					pointing to that tree is also created.
+ *					In the case where the resulting tree is an empty tree
+ *					a new commit pointing to this empty tree will be returned.
+ * @param repo repository where the note lives
+ * @param notes_commit a pointer to the notes commit object
+ * @param author signature of the notes commit author
+ * @param committer signature of the notes commit committer
+ * @param oid OID of the git object to remove the note from
+ *
+ * @return 0 or an error code
+ */
+GIT_EXTERN(int) git_note_commit_remove(
+		git_oid *notes_commit_out,
+		git_repository *repo,
+		git_commit *notes_commit,
+		const git_signature *author,
+		const git_signature *committer,
+		const git_oid *oid);
+
+/**
  * Free a git_note object
  *
  * @param note git_note object
--- a/include/git2/notes.h
+++ b/include/git2/notes.h
@@ -153,6 +153,36 @@ GIT_EXTERN(int) git_note_create(
 	const char *note,
 	int force);
 
+/**
+ * Add a note for an object from a commit
+ *
+ * This function will create a notes commit for a given object,
+ * the commit is a dangling commit, no reference is created.
+ *
+ * @param notes_commit_out pointer to store the commit (optional);
+ *					NULL in case of error
+ * @param notes_blob_out a point to the id of a note blob (optional)
+ * @param repo repository where the note will live
+ * @param parent Pointer to parent note
+ *					or NULL if this shall start a new notes tree
+ * @param author signature of the notes commit author
+ * @param committer signature of the notes commit committer
+ * @param oid OID of the git object to decorate
+ * @param note Content of the note to add for object oid
+ * @param allow_note_overwrite Overwrite existing note
+ *
+ * @return 0 or an error code
+ */
+GIT_EXTERN(int) git_note_commit_create(
+	git_oid *notes_commit_out,
+	git_oid *notes_blob_out,
+	git_repository *repo,
+	git_commit *parent,
+	const git_signature *author,
+	const git_signature *committer,
+	const git_oid *oid,
+	const char *note,
+	int allow_note_overwrite);
 
 /**
  * Remove the note for an object
--- a/include/git2/notes.h
+++ b/include/git2/notes.h
@@ -52,6 +52,20 @@ GIT_EXTERN(int) git_note_iterator_new(
 	const char *notes_ref);
 
 /**
+ * Creates a new iterator for notes from a commit
+ *
+ * The iterator must be freed manually by the user.
+ *
+ * @param out pointer to the iterator
+ * @param notes_commit a pointer to the notes commit object
+ *
+ * @return 0 or an error code
+ */
+GIT_EXTERN(int) git_note_commit_iterator_new(
+	git_note_iterator **out,
+	git_commit *notes_commit);
+
+/**
  * Frees an git_note_iterator
  *
  * @param it pointer to the iterator
--- a/include/git2/notes.h
+++ b/include/git2/notes.h
@@ -94,6 +94,25 @@ GIT_EXTERN(int) git_note_read(
 	const char *notes_ref,
 	const git_oid *oid);
 
+
+/**
+ * Read the note for an object from a note commit
+ *
+ * The note must be freed manually by the user.
+ *
+ * @param out pointer to the read note; NULL in case of error
+ * @param repo repository where to look up the note
+ * @param notes_commit a pointer to the notes commit object
+ * @param oid OID of the git object to read the note from
+ *
+ * @return 0 or an error code
+ */
+GIT_EXTERN(int) git_note_commit_read(
+	git_note **out,
+	git_repository *repo,
+	git_commit *notes_commit,
+	const git_oid *oid);
+
 /**
  * Get the note author
  *
--- a/src/notes.c
+++ b/src/notes.c
@@ -497,6 +497,37 @@ cleanup:
 	return error;
 }
 
+int git_note_commit_create(
+	git_oid *notes_commit_out,
+	git_oid *notes_blob_out,
+	git_repository *repo,
+	git_commit *parent,
+	const git_signature *author,
+	const git_signature *committer,
+	const git_oid *oid,
+	const char *note,
+	int allow_note_overwrite)
+{
+	int error;
+	git_tree *tree = NULL;
+	char target[GIT_OID_HEXSZ + 1];
+
+	git_oid_tostr(target, sizeof(target), oid);
+
+	if (parent != NULL && (error = git_commit_tree(&tree, parent)) < 0)
+		goto cleanup;
+
+	error = note_write(notes_commit_out, notes_blob_out, repo, author,
+			committer, NULL, note, tree, target, &parent, allow_note_overwrite);
+
+	if (error < 0)
+		goto cleanup;
+
+cleanup:
+	git_tree_free(tree);
+	return error;
+}
+
 int git_note_remove(git_repository *repo, const char *notes_ref_in,
 		const git_signature *author, const git_signature *committer,
 		const git_oid *oid)
--- a/src/notes.c
+++ b/src/notes.c
@@ -463,6 +463,28 @@ int git_note_read(git_note **out, git_repository *repo,
 	return error;
 }
 
+int git_note_commit_read(
+	git_note **out,
+	git_repository *repo,
+	git_commit *notes_commit,
+	const git_oid *oid)
+{
+	int error;
+	git_tree *tree = NULL;
+	char target[GIT_OID_HEXSZ + 1];
+
+	git_oid_tostr(target, sizeof(target), oid);
+
+	if ((error = git_commit_tree(&tree, notes_commit)) < 0)
+		goto cleanup;
+
+	error = note_lookup(out, repo, notes_commit, tree, target);
+
+cleanup:
+	git_tree_free(tree);
+	return error;
+}
+
 int git_note_create(
 	git_oid *out,
 	git_repository *repo,
--- a/src/notes.c
+++ b/src/notes.c
@@ -268,7 +268,9 @@ static int insert_note_in_tree_enotfound_cb(git_tree **out,
 		GIT_FILEMODE_BLOB);
 }
 
-static int note_write(git_oid *out,
+static int note_write(
+	git_oid *notes_commit_out,
+	git_oid *notes_blob_out,
 	git_repository *repo,
 	const git_signature *author,
 	const git_signature *committer,
@@ -294,12 +296,16 @@ static int note_write(git_oid *out,
 		insert_note_in_tree_enotfound_cb)) < 0)
 		goto cleanup;
 
-	if (out)
-		git_oid_cpy(out, &oid);
+	if (notes_blob_out)
+		git_oid_cpy(notes_blob_out, &oid);
+
 
 	error = git_commit_create(&oid, repo, notes_ref, author, committer,
 				  NULL, GIT_NOTES_DEFAULT_MSG_ADD,
 				  tree, *parents == NULL ? 0 : 1, (const git_commit **) parents);
+
+	if (notes_commit_out)
+		git_oid_cpy(notes_commit_out, &oid);
 
 cleanup:
 	git_tree_free(tree);
@@ -480,7 +486,7 @@ int git_note_create(
 	if (error < 0 && error != GIT_ENOTFOUND)
 		goto cleanup;
 
-	error = note_write(out, repo, author, committer, notes_ref,
+	error = note_write(NULL, out, repo, author, committer, notes_ref,
 			note, tree, target, &commit, allow_note_overwrite);
 
 cleanup:
--- a/src/notes.c
+++ b/src/notes.c
@@ -766,6 +766,25 @@ cleanup:
 	return error;
 }
 
+int git_note_commit_iterator_new(
+	git_note_iterator **it,
+	git_commit *notes_commit)
+{
+	int error;
+	git_tree *tree;
+
+	if ((error = git_commit_tree(&tree, notes_commit)) < 0)
+		goto cleanup;
+
+	if ((error = git_iterator_for_tree(it, tree, NULL)) < 0)
+		git_iterator_free(*it);
+
+cleanup:
+	git_tree_free(tree);
+
+	return error;
+}
+
 int git_note_next(
 	git_oid* note_id,
 	git_oid* annotated_id,
--- a/src/notes.c
+++ b/src/notes.c
@@ -424,8 +424,7 @@ static int normalize_namespace(char **out, git_repository *repo, const char *not
 	return note_get_default_ref(out, repo);
 }
 
-static int retrieve_note_tree_and_commit(
-	git_tree **tree_out,
+static int retrieve_note_commit(
 	git_commit **commit_out,
 	char **notes_ref_out,
 	git_repository *repo,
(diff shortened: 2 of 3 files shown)


richardipsum:
@joshtriplett @ethomson I am not certain how to go about the last stage of this feature. It's not clear how to have, for example, `git_note_create` wrap `git_note_create_from_commit` since `git_note_create` returns the oid of the note blob, which is not returned by `git_note_create_from_commit`.

I really have no idea what the best option would be here, other than just leaving the existing functions as they are rather than trying to make them wrap the `from_commit` functions.

Ideas and suggestions would be really helpful at this point!

joshtriplett:
@richardipsum While it seems somewhat unnecessary for most applications, returning the new blob OID as an additional out parameter seems fine.  (A caller could pass NULL if they don't care about the blob OID.)

richardipsum:
@joshtriplett That would work, okay I'll do that then, thanks :)

richardipsum:
@ethomson I'm not sure what the test failure here is really about, but it seems unrelated to my change?

ethomson:
@joshtriplett can you provide some feedback here?  Does this address your problem(s)?  Are you reasonably happy with the API?  Are there things this doesn't do that you wish it did?  (Or things that it does do that you think are unnecessary?)

ethomson:
Also, I'm not super familiar with notes.  It's literally a feature that I've never used... 

Here's my big question that I don't understand yet:  why is our existing API dealing with refs?  It looks like it should be dealing with commits in the first place, doesn't it?  Skimming the `git notes` documentation, it looks like `git` itself deals with commits...?

Is our existing API useful at all?  Ignoring API compatibility for a moment, should we replace what we have with this?

joshtriplett:
@ethomson Yes, this does exactly what I need it to do.

I'd assume the existing API deals with refs because by default, `git notes` uses a ref under refs/notes/commits to accumulate the notes. So, it makes sense to retain APIs that make it easy to say "add a note to the default notes ref". The API I needed was for the case where you want to store and share notes a different way, unlike the `git notes` tool.

pks-t:
@ethomson You got that slightly wrong. Notes do not even work on commits, but instead on arbitrary objects. Will comment on the usefulness of our API in a few minutes, as soon as I've reviewed this PR

include/git2/notes.h
```diff
@@ -153,6 +153,33 @@ GIT_EXTERN(int) git_note_create(
...
+GIT_EXTERN(int) git_note_commit_create(
+	git_commit **notes_commit_out,
+	git_oid *notes_blob_out,
+	git_repository *repo,
+	git_commit *notes_commit,
+	const git_signature *author,
+	const git_signature *committer,
+	const git_oid *oid,
+	const char *note,
+	int allow_note_overwrite);
```
pks-t: I'm currently trying to understand this function. As I understand, a note is created for an object (e.g. either a blob, tree, commit or tag) inside of a specific reference. At least, this is how I understand `git notes add`.

So this function here removes the note reference parameter from `git_note_create` which hints where the note should be put in, and instead uses an existing notes commit as its parent. And subsequently, we do not store the resulting notes commit in any reference, instead it's created as a dangling commit.

So this buys us the ability to easily create a complete notes tree without having to write and re-read the notes reference, correct? And in the end, the developer is free to create the notes reference for himself.

include/git2/notes.h
```diff
@@ -153,6 +153,33 @@ GIT_EXTERN(int) git_note_create(
...
+ *					NULL in case of error
+ * @param notes_blob_out a point to the id of a note blob (optional)
+ * @param repo repository where the note will live
+ * @param notes_commit a pointer to the notes commit object
+ *					or NULL if this is to be the first notes object
+ * @param author signature of the notes commit author
+ * @param committer signature of the notes commit committer
+ * @param oid OID of the git object to decorate
+ * @param note Content of the note to add for object oid
+ * @param force Overwrite existing note
```
pks-t: How can this function possibly overwrite an existing note if no notes reference is created?

include/git2/notes.h
```diff
@@ -153,6 +153,33 @@ GIT_EXTERN(int) git_note_create(
...
 
+/**
+ * Add a note for an object from a commit
+ *
+ * @param notes_commit_out pointer to store the commit (optional);
+ *					NULL in case of error
+ * @param notes_blob_out a point to the id of a note blob (optional)
+ * @param repo repository where the note will live
+ * @param notes_commit a pointer to the notes commit object
+ *					or NULL if this is to be the first notes object
```
pks-t: I think naming this `parent` would be more insightful, wouldn't it?

```
@param parent Pointer to parent note or NULL if this shall start a new notes tree
```

include/git2/notes.h
```diff
@@ -153,6 +153,33 @@ GIT_EXTERN(int) git_note_create(
 	const char *note,
 	int force);
 
+/**
+ * Add a note for an object from a commit
```
pks-t: It would be insightful if this would note that no actual notes reference is written.

pks-t: Or maybe even better: adopt an optional notes reference which is to be written. This would ease finalizing the notes tree.

src/notes.c
```diff
@@ -491,6 +497,50 @@ int git_note_create(
...
+			note, tree, target, &notes_commit, allow_note_overwrite);
+
+	if (error < 0)
+		goto cleanup;
+
+	if (notes_commit_out != NULL) {
+		error = git_commit_lookup(notes_commit_out, repo, &note_commit_oid);
+		if (error < 0)
+			goto cleanup;
+	}
```
pks-t: So we actually do a separate lookup on this commit. I'd instead propose simply giving out the OID of the commit, as with the blob ID, and have the caller handle the lookup in case it is required.

include/git2/notes.h
joshtriplett: @pks-t Yes, that's the exact use case.

include/git2/notes.h
joshtriplett: @pks-t Adding an *optional* notes reference does seem convenient, but on the other hand, then you'd want the ability to use the current value of that reference as the parent, and that starts to make this API look complicated.

src/notes.c
joshtriplett: @pks-t Agreed.

richardipsum:
@pks-t  Thanks for your review, I need to think carefully about what's been said here, I will do so and reply once that's done.

include/git2/notes.h
pks-t: @joshtriplett I think I agree by now that the signature is fine as-is.

include/git2/notes.h
richardipsum: yes

include/git2/notes.h
richardipsum: I believe we can overwrite the notes object itself.

    	if ((error = manipulate_note_in_tree_r(
		&tree, repo, commit_tree, &oid, target, 0,
		allow_note_overwrite ? insert_note_in_tree_enotfound_cb : insert_note_in_tree_eexists_cb,
		insert_note_in_tree_enotfound_cb)) < 0)
		goto cleanup;

include/git2/notes.h
richardipsum: indeed, exactly

src/notes.c
richardipsum: returning the oid would also make this more consistent since `git_commit_create` takes an array of `git_commit` objects and returns an `oid`

richardipsum:
@pks-t I agree with all your points and will make the necessary changes.

richardipsum:
@pks-t 

> I'm a bit on the edge though regarding adding a second set of functions that handle note commits instead of references. It simply feels wrong and is apt to confuse users. The current API is really restricted though in its use and seems a tad mis-fitted for more complex handling of notes. So I tend to think that adjusting the current API to work on commits instead of references would be preferable. Creating the actual notes ref should then be handled via our refs API.

I'm not sure what you're suggesting here, the current PR provides an API that allows more complex handling of notes while still keep backwards compatibility, are you saying you're considering replacing the existing set of functions with these? If so perhaps this is something that could be done in a separate PR, since I imagine it's not possible to merge such an API breaking change without releasing a new major version.

For now I will just make the few small changes you've suggested and not worry about removing the existing notes functions.

richardipsum:
@pks-t I've implemented the changes suggested!

tests/notes/notes.c
```diff
@@ -370,6 +370,46 @@ static char *messages[] = {
 
 #define MESSAGES_COUNT (sizeof(messages)/sizeof(messages[0])) - 1
 
+/* Test that we can read a note */
```
pks-t: No need for the comment here.

tests/notes/notes.c
```diff
@@ -370,6 +370,46 @@ static char *messages[] = {
...
+
+	cl_assert(notes_commit);
+
+	cl_git_pass(git_note_commit_read(&note, _repo, notes_commit, &oid));
+
+	cl_assert_equal_s(git_note_message(note), "I decorate 4a20\n");
+
+	git_commit_free(notes_commit);
+	git_note_free(note);
+}
```
pks-t: I'd recommend to also test some edge-cases here. At least a test for a note that cannot be read would be nice to have.

include/git2/notes.h
```diff
@@ -223,6 +223,27 @@ GIT_EXTERN(int) git_note_remove(
 	const git_oid *oid);
 
 /**
+ * Remove the note for an object
+ *
+ * @param notes_commit_out pointer to store the commit (optional);
+ *					NULL in case of error
```
pks-t: Which commit is this exactly? Is this any different from `notes_commit`? I think this concept should be explained a bit better.

src/notes.c
```diff
@@ -587,6 +595,41 @@ int git_note_remove(git_repository *repo, const char *notes_ref_in,
...
+
+	error = note_remove(&note_oid,
+		repo, author, committer, NULL, tree, target, &notes_commit);
+
+	if (error < 0)
+		goto cleanup;
+
+	if (notes_commit_out != NULL) {
+		git_oid_cpy(notes_commit_out, &note_oid);
+	}
```
pks-t: No braces needed.

src/notes.c
```diff
@@ -587,6 +595,41 @@ int git_note_remove(git_repository *repo, const char *notes_ref_in,
...
+		const git_signature *committer,
+		const git_oid *oid)
+{
+	int error;
+	char *target = NULL;
+	git_tree *tree = NULL;
+	git_oid note_oid;
+
+	target = git_oid_allocfmt(oid);
+	GITERR_CHECK_ALLOC(target);
```
pks-t: No need for dynamic allocation here.

richardipsum:
@pks-t Thanks for the review! I should be able to find time to make the requested changes this weekend, thanks!

include/git2/notes.h
richardipsum: Yes it should be, when a note is removed we make a new tree with all notes sans the note to be removed, so a new commit is created pointing to this new tree. Even if we remove the last note, we still have a pointer (the commit) to an empty tree object.

src/notes.c
```diff
@@ -463,6 +463,30 @@ int git_note_read(git_note **out, git_repository *repo,
...
+	git_repository *repo,
+	git_commit *notes_commit,
+	const git_oid *oid)
+{
+	int error;
+	char *target = NULL;
+	git_tree *tree = NULL;
+
+	target = git_oid_allocfmt(oid);
+	GITERR_CHECK_ALLOC(target);
```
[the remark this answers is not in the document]
richardipsum: I must have missed it, will switch to static buffer, I also notice that in `git_note_read` `target` is allocated then not used, so will fix that also.

src/notes.c
```diff
@@ -491,6 +497,48 @@ int git_note_create(
...
+	const git_oid *oid,
+	const char *note,
+	int allow_note_overwrite)
+{
+	int error;
+	git_tree *tree = NULL;
+	char *target = NULL;
+	git_oid note_commit_oid, note_blob_oid;
+
+	target = git_oid_allocfmt(oid);
```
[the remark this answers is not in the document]
richardipsum: Not sure I follow your first point about accepting an OID. Do you mean why does `note_write` take target as `char *` rather than `git_oid *` ?

For now I'll just switch to static buffer, and `git_oid_tostr`

richardipsum:
@pks-t I've finally completed the requested changes, I'm sorry it took me such a long time to complete them. :disappointed:

pks-t:
No need to be disappointed. To be honest, I'm so swamped in work right now so that I'm happy whenever it's not one of us being responsible for delayed responses :P Even though your comment is already two weeks old as of now...

src/notes.c
```diff
@@ -497,6 +497,44 @@ int git_note_create(
...
+	git_tree *tree = NULL;
+	git_oid note_commit_oid, note_blob_oid;
+	char target[GIT_OID_HEXSZ + 1];
+
+	git_oid_tostr(target, sizeof(target), oid);
+
+	if (parent != NULL && (error = git_commit_tree(&tree, parent)) < 0)
+		goto cleanup;
+
+	error = note_write(&note_commit_oid, &note_blob_oid, repo, author, committer, NULL,
```
pks-t: Wait a second, this is strange. You're passing in `git_oid **`'s here, but `note_write` accepts `git_oid *` only.

pks-t: Ah, never mind. I've been confused by the naming here. You could've istead just passed in the pointers, but this shall not put your PR on hold now.

tests/notes/notes.c
```diff
@@ -73,6 +73,144 @@ static int note_list_cb(
...
+	}
+
+	cl_fail("Did not see expected note");
+	return 0;
+}
+
+static struct note_create_payload can_create_a_note[] = {
+	{ "1c9b1bc36730582a42d56eeee0dc58673d7ae869", "4a202b346bb0fb0db7eff3cffeb3c70babbd2045", 0 },
+	{ NULL, NULL, 0 }
+};
```
pks-t: This should be declared inside of the test function itself

tests/notes/notes.c
```diff
@@ -73,6 +73,144 @@ static int note_list_cb(
...
+
+		if (git_oid_cmp(&expected_note_oid, blob_oid) != 0) {
+			continue;
+		}
+
+		cl_git_pass(git_oid_fromstr(&expected_target_oid, notes[i].object_oid));
+
+		if (git_oid_cmp(&expected_target_oid, annotated_obj_id) != 0) {
+			continue;
+		}
```
pks-t: We usually avoid braces around single-line statements.

tests/notes/notes.c
```diff
@@ -73,6 +73,144 @@ static int note_list_cb(
...
+	}
+
+	cl_assert_equal_i(1, retrieved_notes);
+}
+
+static struct note_create_payload can_create_a_note_from_commit[] = {
+	{ "1c9b1bc36730582a42d56eeee0dc58673d7ae869", "4a202b346bb0fb0db7eff3cffeb3c70babbd2045", 0 },
+	{ NULL, NULL, 0 }
+};
```
pks-t: Same here

tests/notes/notes.c
```diff
@@ -73,6 +73,123 @@ static int note_list_cb(
...
+	cl_git_pass(git_reference_create(&ref, _repo, "refs/notes/i-can-see-dead-notes", &notes_commit_out, 0, NULL));
+
+	cl_git_pass(git_note_foreach(_repo, "refs/notes/i-can-see-dead-notes", note_list_create_cb, &retrieved_notes));
+
+	cl_assert_equal_i(1, retrieved_notes);
+
+	git_reference_free(ref);
+}
+
+/* Test that we can create a note from a commit, given an existing commit */
```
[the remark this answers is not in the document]
pks-t: The comment is still there *heh*

tests/notes/notes.c
```diff
@@ -73,6 +73,144 @@ static int note_list_cb(
...
+	 * so we must manually create the ref, that points to the commit */
+	cl_git_pass(git_reference_create(&ref, _repo, "refs/notes/i-can-see-dead-notes", &notes_commit_out, 0, NULL));
+
+	cl_git_pass(git_note_foreach(_repo, "refs/notes/i-can-see-dead-notes", note_list_create_cb, &can_create_a_note_from_commit_given_an_existing_commit));
+
+	for (i = 0; can_create_a_note_from_commit_given_an_existing_commit[i].note_oid != NULL; i++) {
+		if (can_create_a_note_from_commit_given_an_existing_commit[i].seen) {
+			retrieved_notes++;
+		}
+	}
```
pks-t: Maybe a common function `assert_notes_seens` would be cool here, which simply asserts that all `.seen` fields are set.

src/notes.c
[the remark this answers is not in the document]
pks-t: I guess you missed that one on your revisit :)

src/notes.c
pks-t: Same here

tests/notes/notes.c
```diff
@@ -431,6 +431,31 @@ void test_notes_notes__can_read_a_note_from_a_commit(void)
...
+	git_commit_free(notes_commit);
+
+	git_commit_lookup(&notes_commit, _repo, &notes_commit_oid);
+
+	cl_assert(notes_commit);
+
+	cl_git_fail_with(GIT_ENOTFOUND, git_note_commit_read(&note, _repo, notes_commit, &oid));
+
+	git_commit_free(notes_commit);
+}
```
pks-t: :+1:

richardipsum:
@pks-t Thanks for the review! I'm happy to quickly make these changes right now.

src/notes.c
richardipsum: Ah I see, so I'm duplicating the `git_oid_cpy`, yes that's a shame, I'll change it.

richardipsum:
@pks-t I've made the requested changes :smile:

pks-t:
Thanks @richardipsum for fixing up! I'm fine with this PR now. I'll wait until the end of the week to see if there are additional comments as this is quite big.

/ping @ethomson, @carlosmn

richardipsum:
@ethomson @carlosmn this was approved by @pks-t a while back, do you guys have thoughts on anything I could do to improve it or could it be merged now? thanks!

ethomson:
I'm :+1: on your commit `9019cdb` which drops the `malloc`.  But unless I'm missing something obvious, where are those OID strings actually used?  Can we drop `target` entirely from those two functions?

richardipsum:
@ethomson Yes you're right, target becomes unused when the commit " Rewrite funcs in terms of note_commit funcs" (94e1dd1) is applied, so should be removed by that commit.

I'll make this change.

richardipsum:
@ethomson Done, I also noticed that some of the whitespace is inconsistent, in some cases I did an extra indent to avoid function arguments lining up with the function name. Not sure what the preferred style for this project is. Happy to modify this again for whitespace consistency if someone tells me what is preferred.

ethomson:
Thanks @richardipsum!  😀

joshtriplett:
@richardipsum Thanks for all your work!

pks-t:
Jay, finally! 🍾 😄

PR #4160, opened 2017-03-14
Diff fixes

### LINKED ISSUE #4158: git_diff_from_buffer: there is no initialization of git_diff_options version
### Reproduction steps
```cpp
	git_patch *patch;
	git_diff_options diffOpts;
        git_diff_init_options(&diffOpts, GIT_DIFF_OPTIONS_VERSION);
	diffOpts.flags |= GIT_DIFF_FORCE_BINARY | GIT_DIFF_SHOW_BINARY;
	git_patch_from_buffers(patch, "i am object before", 19, "obj1", "i am object after", 18, "obj1", &diffOpts);
        git_buf *buff;
        git_patch_to_buf(buff, patch);
        git_diff *diff;
        git_diff_from_buffer(diff, buf->ptr, buf->size);
        git_diff_print(diff, GIT_DIFF_FORMAT_PATCH, DiffPrintCB, NULL);
```
### Expected behavior
Print diff
### Actual behavior
error: invalid version 0 on git_diff_options

v0.25 + 1
windows 10

pks-t:
This fixes #4158 along with some other issues

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -120,6 +120,41 @@ int git_diff_get_perfdata(git_diff_perfdata *out, const git_diff *diff)
 	return 0;
 }
 
+int git_diff_foreach(
+	git_diff *diff,
+	git_diff_file_cb file_cb,
+	git_diff_binary_cb binary_cb,
+	git_diff_hunk_cb hunk_cb,
+	git_diff_line_cb data_cb,
+	void *payload)
+{
+	int error = 0;
+	git_diff_delta *delta;
+	size_t idx;
+
+	assert(diff);
+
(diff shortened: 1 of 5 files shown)


ethomson:
So I think that you found (and fixed) the last bit of refactoring that I had forgotten to do when working on the new generated/parsed patch system.  At least, I _hope_ there's no more bits lurking.  😛 

Especially now that this is generalized, I think that it might make sense to hoist the `git_diff_foreach` function out of `patch_generate.c` and into `diff.c` or one of its friends...

pks-t:
I certainly agree with you. I actually already started moving it over, but back then I didn't finish the unification yet, so I ran into some problems. Will add another commit

Issue #4149, opened 2017-03-02
git_submodule_update does not clone and throws assertion


Spotlight0xff:
Hi all,

Libgit2 crashes (throws assertion) during the `git_submodule_update` function.
It may very well be that I don't use it correctly, as the documentation is not entirely exact in how to clone/add a submodule and I haven't found any examples either.

### Reproduction steps

First a script to setup the environment and then the actual code.

#### script to setup the environment
setup script, first the submodule which should be later added and then the parent directory.
The submodule is set up in `/tmp`, because I want absolute paths.
`setup.sh`:
```
#!/bin/bash

# remove previous fixtures
# commented to not delete any stuff if you just copy and pasted it.
#rm -rf /tmp/sm_test
#rm -rf parent
pushd .

# setup submodule
mkdir /tmp/sm_test && cd /tmp/sm_test
git init .
touch sm_file
git add sm_file
git commit -m "initial commit in submodule"
cd ..

popd
pushd .


# setup parent repository
mkdir parent && cd parent
git init .
touch parent_file
git add parent_file
git commit -m "initial commit in parent"

popd
```

#### minimal code for the bug
`bug.c`:
```
#include <git2.h>

#include <assert.h>
#include <stdio.h>


int main(int argc, char *argv[])
{
  git_submodule* sm = NULL;
  git_repository* parent = NULL;
  int result = 0;

  git_libgit2_init();
  result = git_repository_open(&parent, "parent");
  assert(!result);

  result = git_submodule_add_setup(&sm, parent, "file:///tmp/sm_test", "sm_test", 0);
  assert(!result);

  result = git_submodule_update(sm, 1, NULL);
  assert(!result);

  result = git_submodule_add_finalize(sm);
  assert(!result);
  git_submodule_free(sm);
  git_repository_free(parent);
  git_libgit2_shutdown();
  return 0;
}
```

Then just compile and run the whole stuff:
`gcc bug.c -o bug -lgit2 && ./setup.sh && ./bug`

### Expected behavior
The submodule should be cloned correctly and libgit2 shouldn't crash.
`parent/sm_test` should contain the cloned repository.

### Actual behavior
The program crashes with the following 
```
bug: /home/user/build/libgit2/src/object.c:122: git_object_lookup_prefix: Assertion `repo && object_out && id' failed.
```
(id is zero)

### Version of libgit2 (release number or SHA1)
HEAD is at `ba2bc4911d99e20f1060bb6767edab94fd5d97a6`

[...]

PR #4161, opened 2017-03-14
submodule: catch when submodule is not staged on update

pks-t:
When calling `git_submodule_update` on a submodule, we have to retrieve
the ID of the submodule entry in the index. If the function is called on
a submodule which is only partly initialized, the submodule entry may
not be added to the index yet. This leads to an assert when trying to
look up the blob later on.

Fix the issue by checking if the index actually holds the submodule's
ID and erroring out if it does not.

The issue has been uncovered by #4149. Even though it was incorrect usage of the API, this should not result in an assert but in an error, instead.


ethomson:
Thanks!  😄

PR #4166, opened 2017-03-20
Remove `map_free` macros

pks-t:
I just noticed a subtle bug via Coverity which I introduced in my recent refactoring of the map interfaces. The `git_map_free` functions were implemented as macros, but in fact in a subtly wrong way where the statements were not grouped together by a `do { ... } while (0)` block. I think no real issues were introduced by this, but I took it as a hint to remove the macro altogether. One caller needed to be fixed, but this was a double-free anyway and as such wrong.


ethomson:
👍

PR #4167, opened 2017-03-20
Coverity fixes

pks-t:
Some Coverity fixes. The most important one is the `openssl_stream` commit, which fixes an actual use after free/segfault when we shut down OpenSSL's mutexes.

### DIFF
--- a/src/blame_git.c
+++ b/src/blame_git.c
@@ -478,14 +478,15 @@ cleanup:
  * The blobs of origin and porigin exactly match, so everything origin is
  * suspected for can be blamed on the parent.
  */
-static void pass_whole_blame(git_blame *blame,
+static int pass_whole_blame(git_blame *blame,
 		git_blame__origin *origin, git_blame__origin *porigin)
 {
 	git_blame__entry *e;
 
-	if (!porigin->blob)
-		git_object_lookup((git_object**)&porigin->blob, blame->repository,
-				git_blob_id(origin->blob), GIT_OBJ_BLOB);
+	if (!porigin->blob &&
+	    git_object_lookup((git_object**)&porigin->blob, blame->repository,
+				git_blob_id(origin->blob), GIT_OBJ_BLOB) < 0)
+		return -1;
 	for (e=blame->ent; e; e=e->next) {
 		if (!same_suspect(e->suspect, origin))
 			continue;
@@ -493,6 +494,8 @@ static void pass_whole_blame(git_blame *blame,
 		origin_decref(e->suspect);
 		e->suspect = porigin;
 	}
+
+	return 0;
 }
 
 static int pass_blame(git_blame *blame, git_blame__origin *origin, uint32_t opt)
@@ -543,7 +546,8 @@ static int pass_blame(git_blame *blame, git_blame__origin *origin, uint32_t opt)
 		}
 		if (porigin->blob && origin->blob &&
 		    !git_oid_cmp(git_blob_id(porigin->blob), git_blob_id(origin->blob))) {
-			pass_whole_blame(blame, origin, porigin);
+			error = pass_whole_blame(blame, origin, porigin);
+				goto finish;
 			origin_decref(porigin);
 			goto finish;
 		}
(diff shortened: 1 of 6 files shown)


src/config_file.c
```diff
@@ -1041,7 +1041,8 @@ static int parse_section_header_ext(struct reader *reader, const char *line, con
 	GITERR_CHECK_ALLOC_ADD(&alloc_len, base_name_len, quoted_len);
 	GITERR_CHECK_ALLOC_ADD(&alloc_len, alloc_len, 2);
 
-	git_buf_grow(&buf, alloc_len);
+	if (git_buf_grow(&buf, alloc_len) < 0)
+		return -1;
```
ethomson: It looks like we're not doing any error checking on `buf` in the rest of this function.  We should probably check that we didn't OOM while trying to work with this buffer before return.

pks-t: Agreed, thanks!

src/diff_parse.c
```diff
@@ -44,7 +44,11 @@ static git_diff_parsed *diff_parsed_alloc(void)
 	diff->base.patch_fn = git_patch_parsed_from_diff;
 	diff->base.free_fn = diff_parsed_free;
 
-	git_diff_init_options(&diff->base.opts, GIT_DIFF_OPTIONS_VERSION);
+	if (git_diff_init_options(&diff->base.opts, GIT_DIFF_OPTIONS_VERSION) < 0) {
```
ethomson: I was surprised to find that the `git_*_init_options` functions return an `int`, since none of them can fail.  I guess it's nice future proofing.  Anyway, if this shuts coverity up then we should take it.

src/config_file.c
```diff
@@ -1041,8 +1041,9 @@ static int parse_section_header_ext(struct reader *reader, const char *line, con
 	GITERR_CHECK_ALLOC_ADD(&alloc_len, base_name_len, quoted_len);
 	GITERR_CHECK_ALLOC_ADD(&alloc_len, alloc_len, 2);
 
-	git_buf_grow(&buf, alloc_len);
-	git_buf_printf(&buf, "%s.", base_name);
+	if (git_buf_grow(&buf, alloc_len) < 0
+	    || git_buf_printf(&buf, "%s.", base_name) < 0)
+		goto end_parse;
```
ethomson: I was thinking that we didn't even need to test this if we have the `git_buf_oom` test at the bottom.  (Since the buffer will hold on to their OOM failure.)  This lets us fail earlier, which could be nice.  I don't have a strong preference, but I would suggest that you change the formatting so that the `||` is at the end of the first line, not starting the second.

src/blame_git.c
```diff
@@ -478,21 +478,24 @@ static git_blame__origin* find_origin(
...
+static int pass_whole_blame(git_blame *blame,
 		git_blame__origin *origin, git_blame__origin *porigin)
 {
 	git_blame__entry *e;
 
-	if (!porigin->blob)
-		git_object_lookup((git_object**)&porigin->blob, blame->repository,
-				git_blob_id(origin->blob), GIT_OBJ_BLOB);
+	if (!porigin->blob
+	    && git_object_lookup((git_object**)&porigin->blob, blame->repository,
```
ethomson: Can you pull the `&&` up so that it's at the end of the first line, not starting the second?

src/patch_parse.c
```diff
@@ -444,9 +444,9 @@ static int parse_header_git(
 				goto done;
 
 			parse_advance_ws(ctx);
-			parse_advance_expected_str(ctx, "\n");
 
-			if (ctx->line_len > 0) {
+			if (parse_advance_expected_str(ctx, "\n") < 0
+			    || ctx->line_len > 0) {
```
ethomson: Can you pull the `||` up onto the line above?

src/diff_parse.c
pks-t: I was surprised by this, as well. And actually it _can_ fail if somehow compiling against two versions of libgit2. Unlikely, but possible.

src/config_file.c
pks-t: I'm a bit torn, as well. Anyway, I think failing early here is a nice to have and it shuts up Coverity without any real detriment.

Huh. Where did I take up that "||" should start at newline... dunno. Will fix all occasions, thanks.

pks-t:
Thanks for your review, fixed all!

ethomson:
Thanks!

Issue #4168, opened 2017-03-20
git_config_set_string fails for symlink for config file on Windows


csware:
### Reproduction steps

Create a symlink on `%HOME%\.gitconfig` (e.g., using `mklink .gitconfig d:\.gitconfig`).

Try to alter a config value using `git_config_set_string`

### Expected behavior

The config file is updated.

### Actual behavior

`git_config_set_string` fails with: `failed to create locked file 'C:/Users/user/.gitconfigd:/.gitconfig.lock': Die Syntax für den Dateinamen, Verzeichnisnamen oder die Datenträgerbezeichnung ist falsch.`

### Version of libgit2 (release number or SHA1)

7143145f2fcb0f7476f333f9df990b8a06cd52c0

### Operating system(s) tested

Win7

csware:
The error occurrs here: https://github.com/libgit2/libgit2/blob/7143145f2fcb0f7476f333f9df990b8a06cd52c0/src/filebuf.c#L247.

pks-t:
This should be fixed by #4169.

csware:
Yes, the issue is solved now.

PR #4169, opened 2017-03-20
Fix resolving absolute symlinks (on at least Win32)

csware:
The symlink destination is always concatenated to the original path.

See issue #4168.

based on the implementation I assume that this is also broken for *nix symlinks.


pks-t:
I've manually merged this via 31059923bd34b514b046379bf5510d55e7505b5a. I've changed your commit to use `git_buf_sets` instead of `git_buf_clear/git_buf_puts` and added another test which verifies your fix.

Thanks for this fix! :smile:

PR #4170, opened 2017-03-20
[RFC] Overflow macros

pks-t:
So I'm quite regularly confused by the `GIT_CHECK_ALLOC_ADD` kind of macros. While these are used in the context of memory allocations, they are in fact quite generic in their possible use in that these macros simply add/multiplicate integers with each other and check for an overflow.

So I've renamed these macros to remove the `ALLOC` part and switched their error class to a new class `GITERR_OVERFLOW`. I'm not too sure about it, though: while it improves readability for me, these macros are only used in the context of memory allocations. So I'm putting this up as an RFC.

### DIFF
--- a/include/git2/errors.h
+++ b/include/git2/errors.h
@@ -101,7 +101,8 @@ typedef enum {
 	GITERR_FILESYSTEM,
 	GITERR_PATCH,
 	GITERR_WORKTREE,
-	GITERR_SHA1
+	GITERR_SHA1,
+	GITERR_OVERFLOW
 } git_error_t;
 
 /**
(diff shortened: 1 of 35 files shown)


ethomson:
I prefer what we have now, but I also wrote it, so I'm curious about what @carlosmn thinks.

If we tried to allocate -1 bytes, I would expect an OOM error instead of an error saying that there was an integer overflow.  I realize that's actually what we're detecting, but it's in the context of an allocation.

carlosmn:
Do you have some other place in mind where you'd like to use them? These are indeed pretty specific to dealing with memory allocation and without a specific extra user-case for them, I'd rather keep them as-is.

pks-t:
I've got no concrete examples right now. It pops up from time to tie, but it is rather uncommon, I agree. Will close this PR for now

PR #4171, opened 2017-03-20
Fix typo in remote.h API

rcjsuen:
Remotes are being incorrectly referred to as a `remove` instead of a `remote`. This should be fixed.


pks-t:
Oops. Wrong button, sorry. Again: thanks for the fix! :tada:

PR #4173, opened 2017-03-22
mbedTLS support

tiennou:
This is a heavily redacted blame-friendly rejig of #3935. Redacted because commits have been merged/split for blame-friendliness and sometimes tweaked for styling/clarity.

The first part (no `mbedtls:` tag) is preliminary cleanup to various unrelated things.

- [x] Fix indentation
- [x] The CA store problem has been not tackled. I've been thinking about passing a CMake define, but haven't got around to do it.

I've actually softened mbedtls' initialization so it doesn't fail the library completely (hit a crash because mwindow something wasn't initialized, hence ce20d0d1ba6675e56164e5caf74f01b358e547ad).

Cipher list support is new, but had to duplicate the list because OpenSSL names aren't [compatible](https://tls.mbed.org/supported-ssl-ciphersuites). I've preserved the ordering (for now), but do note that mbedTLS doesn't support the `DSS` versions of those ciphers.


tiennou:
@tkelman @joshtriplett

pks-t:
Hi @tiennou. Would you like to split up this PR? I particularly want to merge the first few patches which aren't specific to mbedTLS, as I think they do provide some value. Would also make it easier to review the second part regarding mbedTLS support as its much more local then.

Anyway, it's okay if you're not interested in splitting up and rebasing. Just tell me and I'll do this myself :)

tiennou:
Up to 594e27d ?

pks-t:
Yeah, but excluding 2cdd5dd as I've already implemented the same thing a few weeks ago (even though it's called GIT_HTTPS now.

src/streams/mbedtls.c
```diff
@@ -0,0 +1,538 @@
...
+
+	cacert = git__malloc(sizeof(mbedtls_x509_crt));
+	mbedtls_x509_crt_init(cacert);
+	if (is_dir) {
+		ret = mbedtls_x509_crt_parse_path(cacert, path);
+	} else {
+		ret = mbedtls_x509_crt_parse_file(cacert, path);
+	}
+	// mbedtls_x509_crt_parse_path returns the number of invalid certs on success
+	if (ret <= 0) {
```
omus: `ret == 0` indicates success with no invalid certificates. Should be changed to `ret < 0`.
Docs for [`mbedtls_x509_crt_parse_path`](https://tls.mbed.org/api/group__x509__module.html#ga571fc89b9f3217ab3dd67bd7af905066)

src/settings.c
```diff
@@ -171,14 +177,19 @@ int git_libgit2_opts(int key, ...)
...
 			const char *file = va_arg(ap, const char *);
 			const char *path = va_arg(ap, const char *);
-			if (!SSL_CTX_load_verify_locations(git__ssl_ctx, file, path)) {
-				giterr_set(GITERR_NET, "SSL error: %s",
-					ERR_error_string(ERR_get_error(), NULL));
-				error = -1;
-			}
+			error = git_openssl_set_cert_file(file, path);
+		}
+#elif GIT_MBEDTLS
```
omus: Should be:
```
#elif defined(GIT_MBEDTLS)
```

src/settings.c
```diff
@@ -171,14 +177,19 @@ int git_libgit2_opts(int key, ...)
...
+		{
+			const char *file = va_arg(ap, const char *);
+			const char *path = va_arg(ap, const char *);
+			if (file)
+				error = git_mbedtls_set_cert_file(file, 0);
+			if (error && path)
+				error = git_mbedtls_set_cert_file(path, 0);
```
omus: Error check should check for negatives as `error >= 0` indicates success. Additionally `git_mbedtls_set_cert_file` needs to be passed a `1` to indicate `is_dir`.
```
if (error < 0 && path)
    error = git_mbedtls_set_cert_file(path, 1);
```

src/streams/mbedtls.c
tiennou: Fixed, thanks !

src/settings.c
tiennou: Fixed. Also fixed the cipher setting below.

src/settings.c
tiennou: Fixed passing 1 here. About the error, I'm tempted to keep it the `libgit2` way (success is 0, everything else is failure). @pks-t ?

omus: Thinking about this more we probably don't want to fall back to loading a dir if loading the file failed since the caller chooses if they want to load a file or dir. Maybe this would be better? 

```
if (file) {
	error = git_mbedtls_set_cert_file(file, 0);
} else if (path) {
	error = git_mbedtls_set_cert_file(path, 1);
} else {
	giterr_set(GITERR_NET, "cannot set certificate locations: no file or path given");
	error = -1;
}
```

src/streams/mbedtls.c
```diff
@@ -0,0 +1,538 @@
...
+
+	cert = mbedtls_ssl_get_peer_cert(ssl);
+	if (!cert) {
+		giterr_set(GITERR_SSL, "the server did not provide a certificate");
+		return -1;
+	}
+
+	/* Check the alternative names */
+	alts = &cert->subject_alt_names;
+	while (alts != NULL && matched != 1) {
```
omus: mbedtls can do the alternative names and common name checks for us. This custom code isn't necessary. See https://github.com/JuliaLang/julia/blob/158d65581dda29e6b9d85c6bc9ba612a09e73210/deps/patches/libgit2-mbedtls-verify.patch

src/settings.c
pks-t: I'm a bit undecided on this one. I guess how @omus proposes is the most sensible way, e.g. returning an error if neither file nor path are given. But what is unfortunate about this API is how we handle the case where both parameters are set, as only the file setting would be applied in that case. This does not really follow the way of least astonishment for the developer. On the other hand, we shouldn't throw an error in the case where both have been set, as otherwise developers would need to be aware whether libgit2 is compiled against mbedTLS or OpenSSL due to the latter accepting both at the same time.

I think the most sensible thing to do here is to use both values, file and path, in case both are set. Can this be supported in some way with mbedTLS?

omus: Since we want to support passing in both a file and a path at the same time the following should work:
```
if (file || path) {
	error = -1;
	if (file)
		error = git_mbedtls_set_cert_file(file, 0);
	if (error < 0 && path)
		error = git_mbedtls_set_cert_file(path, 1);
} else {
	giterr_set(GITERR_NET, "cannot set certificate locations: no file or path given");
	error = -1;
}
```
I think this will have to be handled in libgit2 as mbedTLS only has [`mbedtls_x509_crt_parse_file`](https://tls.mbed.org/api/group__x509__module.html#gad4da63133d3590aa311488497d4c38ec) and [`mbedtls_x509_crt_parse_path`](https://tls.mbed.org/api/group__x509__module.html#ga571fc89b9f3217ab3dd67bd7af905066).

tiennou: Just want to point out, this weird call is due to the [insanity](https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_load_verify_locations.html) of OpenSSL. This will cause verification checks to use the first valid certificate from `file` it finds, else it will enumerate cert files under `path`.
I'm afraid mbedTLS has no mechanism to provide a default CA store because it is out-of-scope for them (same issue with 0c7dd1e).

pks-t:
I'm a bit out of context by now -- do you consider this PR ready for review right now or do you intend to revisit it again and do adjustments?

tiennou:
I'll likely revisit it, I originally submitted it so the Julia people would have a mostly up-to-date version with some of their fixes, but I consider it somewhat prototypal, given the differing certificate location behavior between OpenSSL & mbedTLS.

tiennou:
Rebased, still WIP. I can't help but notice that GitHub is completely confused with the commit ordering here 😕...

@tkelman I integrated the few patches I found in the Julia repo (last 2 commits), if you don't mind.

I also have a question related to the default certificate location (since I feel like it's the last contention point). Is Julia actually using the environment variables from 27c3d68 ? I'd be more comfortable dropping those entirely in favor of a CMake setting, since they're here for compatibility with OpenSSL's own err… stuff.

tkelman:
I'll have to check whether we set those, not sure off the top of my head. @omus has been actively revising some of this lately.

At build time we ask openssl where it put its certs https://github.com/JuliaLang/julia/blob/5f582aece525b2ab6cc7a9cfdf40085c75d34bb8/deps/libgit2.mk#L44-L59 since that differs annoyingly between distros.

Some cmake code would make sense for this, though ideally it would be good if there were some way to override the build time search paths (hardcoding absolute paths into the binary is kind of a bummer for relocatability)

tiennou:
Ok, thanks for the input. I've moved the cert location to a `#define` which gets set by CMake, either automatically depending on what OpenSSL says, or manually via `-DCERT_LOCATION`.

tiennou:
With #4282 now settled, #4283 under review, and a working solution to the CA problem, I feel like this is ready to go too !

tiennou:
As a clarification, CMake now asks the local `openssl` installation for its default CA path, unless the "packager" overrides it, and this value will end up as the "compile time" default. On top of that, the "library user" can still override it at runtime via `git_libgit2_opts`.

CMCDragonkai:
@tiennou To clarify how this would work in emscripten, would this mean that you would have to embed CA certificates into emscripten's virtual filesystem, and then override the CA certificate store path at the cmake phase. I wonder if there would be a way to make use of the browser's CA store instead.

joshtriplett:
In an emscripten or wasm environment, you can only make HTTP or HTTPS requests via the browser, so you *have* to go via the browser's certificate store. (Short of tunneling back through another server.)

mashu:
Any ETA on this one ?

tiennou:
Right now it's waiting for #4283 to be merged, and since both those PR need rebasing, I'll try to find some time to do so.

tiennou:
@CMCDragonkai I have absolutely no emscripten experience, but I feel that "browser certificate store" is several layers of abstraction above us. Also, I think if our crypto API provided a callback instead of some file/dir path, you might be able to provide a JS callback. But then it's likely you'll need x509 "things", and our crypto layer would degenerate quickly from there 😉 .

ethomson:
Hi @tiennou - some feedback from somebody who was trying this out is that some error messages from embedTLS itself might be getting cleared or overwritten, or we're not getting the full data out of them.  Does that sound familiar?

Here's the tweet in question:

https://twitter.com/iamed2/status/931244779593256960

omus:
The tweet references comments made by me. Looking at this code again it looks like the mbedTLS error codes are being correctly reported back to the user. At one point this wasn't the case.

omus:
I hope this PR makes it's way into libgit2. Julia has been using a patched version of this PR for sometime and overall this code works very well!

src/streams/mbedtls.c
```diff
@@ -0,0 +1,451 @@
...
+}
+
+static int verify_server_cert(mbedtls_ssl_context *ssl)
+{
+	int ret = -1;
+
+	if ((ret = mbedtls_ssl_get_verify_result(ssl)) != 0) {
+		char vrfy_buf[512];
+		int len = mbedtls_x509_crt_verify_info(vrfy_buf, sizeof(vrfy_buf), "", ret);
+        if (len <= 1) vrfy_buf[len - 1] = '\0'; // Remove trailing \n
```
tiennou: Indentation mismatch. This also seems wrong if `mbedtls_x509_crt_verify_info` failed, though we'll still report a certificate error.

src/streams/mbedtls.c
```diff
@@ -0,0 +1,451 @@
...
+	mbedtls_ssl_conf_ciphersuites(git__ssl_conf, ciphers_list);
+
+	// Seeding the random number generator
+	mbedtls_entropy = git__malloc(sizeof(mbedtls_entropy_context));
+	mbedtls_entropy_init(mbedtls_entropy);
+
+	ctr_drbg = git__malloc(sizeof(mbedtls_ctr_drbg_context));
+	mbedtls_ctr_drbg_init(ctr_drbg);
+	if (mbedtls_ctr_drbg_seed(ctr_drbg,
+							  mbedtls_entropy_func,
```
tiennou: Xcode indent.

src/streams/mbedtls.c
```diff
@@ -0,0 +1,451 @@
...
+	char *cipher_name = NULL;
+	char *cipher_string = NULL;
+	char *cipher_string_tmp = NULL;
+
+	mbedtls_x509_crt *cacert = NULL;
+
+	git__ssl_conf = git__malloc(sizeof(mbedtls_ssl_config));
+	mbedtls_ssl_config_init(git__ssl_conf);
+	if (mbedtls_ssl_config_defaults(git__ssl_conf,
+									MBEDTLS_SSL_IS_CLIENT,
```
tiennou: Xcode indent.

tests/core/stream.c
```diff
@@ -33,9 +33,8 @@ void test_core_stream__register_tls(void)
 	cl_git_pass(git_stream_register_tls(NULL));
 	error = git_tls_stream_new(&stream, "localhost", "443");
 
-	/* We don't have arbitrary TLS stream support on Windows
-	 * or when openssl support is disabled (except on OSX
-	 * with Security framework).
+	/* We don't have TLS support enabled, or we're on Windows,
+     * which has no arbitrary TLS stream support.
```
ethomson: Whitespace problem here, too.

ethomson:
Hey @tiennou I'd like to ship this - can you fix up the whitespace problems that you noted (and there's one more than I made a note of).  :shipit:

tiennou:
Rebased, with a few last minute fixups, mostly stylistic (there were a few `//` comments, and Xcode indent), plus an inverted comparison. 🚀

omus:
Anything left to do here?

tiennou:
Rebased because of recent CMake changes in master. I'll wait for the verdict of CI though.

Also, rereading the diffs, we should maybe define our backend preference when detecting (ie. `cmake -DUSE_HTTPS=ON`) more explicitly. The order when using this PR is SecureTransport (OS X) / WinHTTP (Windows) / mbedTLS / OpenSSL.

This means mbedTLS (if found) will be preferred over OpenSSL, which is "differing behavior" w.r.t to what we did before.

ethomson:
I’m a little nervous about changing the default ordering, TBQH.  I think that it’s probably not _that_ uncommon to have both installed, and this would be a big change if you did.

I sort of wonder if we shouldn’t try to detect OpenSSL explicitly before this new check and fail with a nice error message if you have HTTPS enabled but we couldn’t find any backends.  That would prevent any existing users from having their SSL backend swapped out from under them.

.travis.yml
```diff
@@ -55,6 +55,16 @@ matrix:
...
      dist: trusty
+   - compiler: gcc
+     env:
+         MBEDTLS=1
+         OPTIONS="-DTHREADSAFE=ON -DCMAKE_BUILD_TYPE=Release -DUSE_TLS=mbedTLS -DMBEDTLS_ROOT_DIR=../mbedtls"
+     os: linux
+   - compiler: gcc
+     env:
+         MBEDTLS=1
+         OPTIONS="-DTHREADSAFE=OFF -DBUILD_EXAMPLES=ON -DUSE_TLS=mbedTLS -DMBEDTLS_ROOT_DIR=../mbedtls"
```
ethomson: `-DUSE_TLS` does not seem to do anything - in fact, it looks like the two new travis builds are building for OpenSSL.

src/CMakeLists.txt
```diff
@@ -133,6 +133,7 @@ ELSE ()
 ENDIF()
 
 IF (USE_HTTPS)
```
ethomson: The comment for `USE_HTTPS` says that you can set it to the name of the backend that you want to use.  I don't think that's true, but it would be nice if it _was_.

tiennou:
I just need a more precise definition of the ordering. If you're ok with preferring OpenSSL vs. mbedTLS, I can implement it that way, and make the auto-detection error out if you used `USE_HTTPS=ON`, but we weren't able to load any prerequisites for it.

.travis.yml
tiennou: Good catch, it's a leftover for the old naming.

ethomson:
Yes, please - what you just described sounds perfect.  (Prefer OpenSSL over mbedTLS - now OpenSSL is the second to last choice, mbedTLS is the last choice).  And if `USE_HTTPS=ON` and there is no TLS available then produce an error message.

src/CMakeLists.txt
tiennou: It does though. If you pass `ON`, this will use "autodetection" (a.k.a. use whatever backend we've found and is first in the following tests), Otherwise, this will be passed verbatim as `HTTPS_BACKEND`, which is handled just below (and will error if you passed nonsense).

ethomson: Yep, rereading, I see it now.  Thanks!

tiennou:
Updated with fixes for your last comment round @ethomson.

There's still one thing I'm not happy with though, but it's unrelated to mbedTLS support (and was introduced by the recent SHA1 selection mechanism) : right now, we only `FIND_PACKAGE` inside that `USE_HTTPS` block, so you could do `cmake -DUSE_HTTPS=OFF -DSHA1_BACKEND=OpenSSL` and end up with link errors.

I'd be more than happy to fix that, but I'm not sure it belongs here…

ethomson:
Yes, I agree that that's problematic.  I'd be happy with that here, if you feel like tackling it, or in a separate PR.

tiennou:
I was wrong actually 😆. The SHA1 for OpenSSL checks are actually redone if they've failed/were not done. I'm actually not sure what's the best way to remove that (potential) duplication, especially since our preference w.r.t SHA1 is CollisionDetection, which is a builtin.

ethomson:
> I was wrong actually 😆. The SHA1 for OpenSSL checks are actually redone if they've failed/were not done. I'm actually not sure what's the best way to remove that (potential) duplication, especially since our preference w.r.t SHA1 is CollisionDetection, which is a builtin.

That seems okay with me - it sounds like you're saying that I could use OpenSSL as my SHA1 algorithm but not as my TLS backend...  or vice versa?  In that case, I think that's fine.

tiennou:
The reverse actually : some SHA1 backends would cause a link failure if you didn't use "their" HTTPS backend. Right now, only the OpenSSL SHA1 backend tests are correct, and Windows should be ok too (because it's configured before as a single step). But something like `-DUSE_HTTPS=SecureTransport -DSHA1_BACKEND=mbedTLS` is not, because the mbedTLS/SecureTransport library setup/libgit2.pc file preparation that is done in the `IF(HTTPS_BACKEND=$backend)` would not have been done, leading to a linking failure.

And on top of that, we would actually default to CollisionDetection as our SHA impl.

A workaround would be to move all those "linking preparation" steps outside both our backend selection, and be done after we setup our defines (since at that time we definitely know what we'll end up needing to link with) :
```
IF(GIT_SHA1_OPENSSL OR GIT_HTTPS_OPENSSL)
  # Do the OpenSSL dance
ELSEIF(GIT_SHA1_COMMONCRYPTO OR GIT_HTTPS_SECURETRANSPORT)
  # Do the Security dance
…
```

pks-t:
That should definitly help us deduplicating the logic around that, which would be very welcome I think.

src/CMakeLists.txt
```diff
@@ -133,6 +133,8 @@ ELSE ()
...
+	# We try to find any packages our backends might use
+	FIND_PACKAGE(OpenSSL)
 	FIND_PACKAGE(mbedTLS)
```
pks-t: There's still one thing I really don't like about this: even if I set USE_HTTPS=OpenSSL, it will try to find mbedTLS and vice versa, which is even printed out on command line. I think this is somewhat suprising for the user. The other way round, just making these calls QUIET, is not really a nice way either. If I set USE_HTTPS=OpenSSL, but never see a single line where it actually searches for the OpenSSL library, I'm confused as well. This is why I originally moved the FIND_PACKAGE logic into the OpenSSL branch itself.

Thinking a bit more about it, I feel like we just shouldn't bother too much about the auto detection. Instead of randomly picking whatever is there, we should declare platform-dependent defaults and just don't care whether those are installed or not. In case somebody has mbedTLS installed on Linux only, he will just have to choose that backend manually, as that is _not_ the common case.

So, the end result should just first check the operating system and set HTTPS_BACKEND to the platform default iff USE_HTTPS=ON. Same for the SHA1 backend. After those two very simple blocks which do not do any checking whether those libraries exist, we can use unified logic to do the actual linking based on what backends for HTTPS and SHA1 are set.

pks-t: (And obviously produce an error in case those libraries are not found)

tiennou:
Help needed ! I've been banging my head against the Linux linker for some time, trying to fix the Travis build, and I'm throwing the towel.

Right now it's setup to do a shared library link, but since I'm building against a local install (in `./deps/mbedtls`), `libgit_clar` will fail to run because it can't resolve the mbedTLS libraries. I suspect I need to tweak `LD_LIBRARY_PATH` or something like that, but I'd really prefer not to. I've tried to `make install` so the library can be found "naturally", but that doesn't work without `sudo`.

So I've tried a static build of mbedTLS, and got back this, which I haven't been able to workaround :
```
Linking C shared library ../libgit2.so
/usr/bin/ld: /home/vagrant/libgit2/deps/mbedtls/library/libmbedtls.a(ssl_ciphersuites.c.o): relocation R_X86_64_32S against `.rodata' can not be used when making a shared object; recompile with -fPIC
/home/vagrant/libgit2/deps/mbedtls/library/libmbedtls.a: error adding symbols: Bad value
```

💡I think I got it : CMake only [obeys](https://cmake.org/cmake/help/latest/envvar/CFLAGS.html) `CFLAGS` on the first run, so this should actually work 🤞.

tiennou:
💚

@ethomson @pks-t I realize it's likely to be too late, but should we consider including that in 0.27 ?

pks-t:
Sorry, but this is definitly too late for v.0.27.0. We are only accepting bug fixes right now, as we want to release soonish (for real now :P). So this is only going to be accepted (and reviewed) post-v0.27

ethomson:
@pks-t any further thoughts on this or are we ready to 🚢

ethomson:
:shipit:

ethomson:
Thanks for all your work here @tiennou - and for the Julia team and everyone who's reviewed and tested this.

Issue #3909, opened 2016-08-24
git_repository_set_head() doesn't include tag name in reflog – `git status` shows OID


psprint:
When doing `git checkout <tag>` git lefts a trace in reflog that allows to interpret detached HEAD state as a tag checkout. The message looks like:

```
ec5bd6c HEAD@{0}: checkout: moving from master to testtag
```

I process the reflog too, and show user that he is detached at tag, not at commit. Message that libgit2 produces is:

```
ec5bd6c HEAD@{0}: checkout: moving from master to ec5bd6cce67ff7d961286cb8a55145072c9bccbc
```

The line of code responsible: https://github.com/libgit2/libgit2/blob/c60210d36780a39db669e52d6d592d339c6a2ed3/src/repository.c#L2290 – it doesn't treat tag specially.

carlosmn:
It's not writing anything special into the reflog. It looks like it or you should be calling `git_repository_set_head_detached_from_annotated()` if you want to do what git does when given a tag.

psprint:
IMO this shouldn't be closed. I wasn't fully clear, however @carlosmn rather got at least spark of the picture. The point is, `git_repository_set_head` will set meaningful log message for branch – there is the `log_message` prepared for `git_reference_symbolic_create`. However, it will not do it for tag – the `log_message` variable is not used. This is an omission and fixing it is rather easy.

ethomson:
Hmm...  I suppose that's reasonable, but I agree with @carlosmn that you probably want `git_repository_set_head_detached_from_annotated` to truly capture the range of values that `git checkout` on the command line can take.

For example, you can `git checkout tagname^{commit}` and git will dutifully write `tagname^{commit}` as the target in the reflog.  `git_repository_set_head` can't capture that.

PR #4174, opened 2017-03-22
git_repository_set_head: use tag name in reflog

ethomson:
When `git_repository_set_head` is provided a tag reference, update the reflog with the tag name, like we do with a branch.  This helps consumers match the semantics of `git checkout tag`.

If you think of `git_repository_set_head` as an analog to `git symbolic-ref HEAD ...` then this doesn't make a lot of sense.  But I would argue that it's not a perfect analog already to `git symbolic-ref`, and this is natural for users who want to `git checkout ...` to switch branches.

See #3909

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -2525,7 +2525,7 @@ static int checkout_message(git_buf *out, git_reference *old, const char *new)
 
 	git_buf_puts(out, " to ");
 
-	if (git_reference__is_branch(new))
+	if (git_reference__is_branch(new) || git_reference__is_tag(new))
 		git_buf_puts(out, git_reference__shorthand(new));
 	else
 		git_buf_puts(out, new);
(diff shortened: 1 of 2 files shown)


carlosmn:
If we're going full-on with this, should we short-hand everything? `git checkout origin/master` stores `origin/master` as the destination.

ethomson:
Hrmf.  I dunno...  `origin/master` makes sense to shorthand here...  but I worry that we're running out of the ability to communicate our desires.  You can `git checkout tag^{commit}` and git stores `tag^{commit}` but that gets here.

We could actually be very dwimmy here, but I'm not sure how far we want to go.  This is a very middle of the road approach in this PR, but we could actually try to parse what we're given.

🤷‍♂️

carlosmn:
If you want to include the exact text from the user, that's what `git_repository_set_head_detached_from_annotated()` does, but this is for passing in references. We could swing in the other direction and only accept annotated, but that sucks ergonomically. Using the shorthand for the ref (as long as it's a ref name we can look up) seems to be what you'd expect `git` to do here.

ethomson:
Right - that makes sense.  The acceptance of a ref name string (instead of a ref) was confusing me.  In any case, I updated this to deal with shortnames for remote branches.

PR #4175, opened 2017-03-22
git_futils: don't O_EXCL and O_TRUNC

### LINKED ISSUE #4017: ".git/config permission denied" when cloning on NFS-volume
I'm trying to run [Julia](https://github.com/JuliaLang/julia) on a Linux system with my home directory being mounted through NFS (nfs4) and ran into a problem with the Julia package manager. Julia is using libgit2 to download its packages.

The problem i encountered was that while cloning a repository the file ".git/config" ends up with no permissions at all and thus libgit2 failing to write to that file. I figured out that this only happens when the target directory of the clone is on a NFS mount (everything works fine with local hard disks).
The problem appears to be `git_futils_creat_locked()` in fileops.c and the flags it passes to `p_open()`.

`int fd = p_open(path, O_WRONLY | O_CREAT | O_TRUNC | O_EXCL | O_BINARY | O_CLOEXEC, mode);`

Once I remove `O_TRUNC` and recompiled libgit2 (actually whole Julia) the error was gone.

It seems to me that using `O_CREAT| O_EXCL` together with `O_TRUNC` causes this bug. I'm wondering if it even makes sense to use all those flags at the same time since `O_EXCL` should cause `open()` to fail if a file exists and `O_TRUNC` should only work if a file exists (it says in the manpage of open() that behavior is unspecified when a file does not exist and `O_TRUNC`is used). So it should be safe to remove `O_TRUNC` from the flags in `git_futils_creat_locked()`.

My issue/fix seems to be similar to #2368

## Patch
I used the following patch to fix the problem.
```
--- fileops.c.orig	2016-11-24 18:22:37.274750191 +0100
+++ fileops.c	2016-11-24 18:22:48.346880757 +0100
@@ -68,7 +68,7 @@
 
 int git_futils_creat_locked(const char *path, const mode_t mode)
 {
-	int fd = p_open(path, O_WRONLY | O_CREAT | O_TRUNC |
+	int fd = p_open(path, O_WRONLY | O_CREAT | 
 		O_EXCL | O_BINARY | O_CLOEXEC, mode);
 
 	if (fd < 0) {


```

ethomson:
`O_EXCL` and `O_TRUNC` are mutually exclusive flags to open(2); you can't truncate a file if you're asserting that it can't exist in the first place.  Drop `O_TRUNC`.

Fixes #4017

### DIFF
--- a/src/fileops.c
+++ b/src/fileops.c
@@ -66,8 +66,8 @@ int git_futils_creat_withpath(const char *path, const mode_t dirmode, const mode
 
 int git_futils_creat_locked(const char *path, const mode_t mode)
 {
-	int fd = p_open(path, O_WRONLY | O_CREAT | O_TRUNC |
-		O_EXCL | O_BINARY | O_CLOEXEC, mode);
+	int fd = p_open(path, O_WRONLY | O_CREAT | O_EXCL | O_BINARY | O_CLOEXEC,
+		mode);
 
 	if (fd < 0) {
 		int error = errno;


pks-t:
Looks good! :+1:

Issue #3397, opened 2015-08-25
Suggestion: add support for http.cookieFile


tracervision:
It would be awesome if libgit2 had support for the http.cookieFile configuration. Tools like Xcode Bots use libgit2 and cookie support would be a big help in an environment where they are used for non-interactive repository access.

http://git-scm.com/docs/git-config

carlosmn:
This is an area in git which is trying to expose libcurl functionality around HTTP, which we do not use so we can't just hand off a file.

What is it that you need this file for? Is authentication done via a cookie? How does it fail if you don't provide it? If you need to set a header field with a cookie, that's something I'd been thinking of adding, but wasn't sure how useful it would be.

dborowitz:
Owner of the servers @tracervision is talking to speaking.

> This is an area in git which is trying to expose libcurl functionality around HTTP, which we do not use so we can't just hand off a file.

I understand that supporting this this is going to be more work than just setting a curl option. But I'm not sure I buy that the answer is to provide some alternative libgit2-specific configuration method--this is a documented config option in git-config(1), so if libgit2 wants to be compatible with git, it should use it. Surely there are other config options that libgit2 respects; where do you draw the line?

> Is authentication done via a cookie?

Yes.

> How does it fail if you don't provide it?

The HTTP request fails and you can't access the repository. Let's say it gives HTTP 403.

> If you need to set a header field with a cookie, that's something I'd been thinking of adding, but wasn't sure how useful it would be.

Not useful. We tell our users to set up http.cookiefile in ~/.gitconfig once and expect it to work everywhere. Even if we wanted to write the shim to read from cookies.txt, we don't own Xcode and can't modify its usage of libgit2 to make extra calls to set headers.

ethomson:
Thanks for the additional background.

> so if libgit2 wants to be compatible with git, it should use it.

Well, that's not *quite* so straightforward, as you point out, as there *is* a line.

libgit2 doesn't (strictly speaking) want to be compatible with git.  We want to enable clients to be compatible with git.  There are a lot of configuration options that only really make sense in the context of the client itself and there are a lot of configuration options that are very leaky abstractions and don't really make sense for us.  This is especially true of many of the `http.*` configuration options.

`http.cookieFile` seems like a reasonable addition to me.

[...]

PR #4177, opened 2017-03-23
Support http cookie file

ankurmittal:
This patch supports passing http cookiefile when libcurl is present
#3397

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ v0.26 + 1
 * Improved `p_unlink` in `posix_w32.c` to try and make a file writable
   before sleeping in the retry loop to prevent unnecessary calls to sleep.
 
+* Added support for http.cookieFile on non-windows system.
+
 ### API additions
 
 * `git_remote_create_detached()` creates a remote that is not associated
--- a/src/transports/cookie.c
+++ b/src/transports/cookie.c
@@ -1,0 +1,757 @@
+/***************************************************************************
+ * Copyright (C) 1998 - 2016, Daniel Stenberg, <daniel@haxx.se>, et al.
+ *
+ * This software is licensed as described in the file COPYING, which
+ * you should have received as part of this distribution. The terms
+ * are also available at https://curl.haxx.se/docs/copyright.html.
+ *
+ * You may opt to use, copy, modify, merge, publish, distribute and/or sell
+ * copies of the Software, and permit persons to whom the Software is
+ * furnished to do so, under the terms of the COPYING file.
+ *
+ * This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+ * KIND, either express or implied.
+ *
+ ***************************************************************************/
+
+#ifndef GIT_WINHTTP
+
+#include "cookie.h"
+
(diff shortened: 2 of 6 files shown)


ankurmittal:
+@petrhosek

ankurmittal:
#3397

ethomson:
Thanks for tackling this!

I think that if this is something that we're going to include then this is probably not something that we want to include at the TLS transport level, but when we build requests.  I realize that OpenSSL has the concept of a cookie file but this doesn't match the abstraction that we've built.

Instead, cookies should be considered part of the request itself, not part of the cryptographic or connection layer.  This means that we can probably add cookie code to `transports/http.c` (perhaps in `gen_request`?) and `transports/winhttp.c` (maybe in `do_send_request`?) - which means that we add it in two places instead of however many TLS backends there are.

This means reading the `http.cookieFile` and applying them ourselves, but this seems much more robust than only adding it for a single backend.

ankurmittal:
Don't have a windows machine, so did not add support for it as no way to test

ankurmittal:
Also not sure about license, should I copy license from other files into cookie.cc and cookie.h?

ankurmittal:
Please review

ethomson:
Hi @ankurmittal - I didn't see that you had moved this into `gen_request` - thanks for that, I'll try to take a look at this soon.

ankurmittal:
rebased the changes with latest

ankurmittal:
please review, also build failure is not because of this CL, it is already failing in master

petrhosek:
@ethomson Would it be possible to review and merge this pull request? We need the support for cookies in order to use libgit2 in our setup.

ankurmittal:
Hi, Can someone please review it?

pks-t:
Hi @ankurmittal. Thanks for your PR. Unfortunately, I'm a bit on the edge about it, and we actually had a rather long discussion yesterday on how to handle things like cookie files for our transports.

So the biggest problem I see here is that we actually include a large amount of non-trivial code from a different project into ours, leaving us with the burden to maintain it, fix issues and triage security issues. Unfortunately, this is not something I am personally keen on doing, which is why I am opposed to including the cookie code from libcurl.

So as mentioned, we had a lengthy discussion on how to proceed from here, because such "high-level" features like for example cookie files or SSL certificates, which users of our library assume to just work, are currently not handled by our transports, and we actually want them to stay dumb. But as these kinds of requests pop up quite frequently, we thought about adding an additional transport which "just works" as expected. It would automatically handle encryption, certificates and also cookie files without us getting in the way by simply offloading all this to libcurl. So while this backend would be rather restricted in advanced use cases, it would be simple to use for other scenarios.

So this is not something that exists yet and is currently only in a drafting stage. Furthermore, this transport will probably not be developed as part of core libgit2 but rather in an external repository like "libgit2/libgit2-curl-transport".

garymm:
@pks-t is there an issue tracking the work you described? If not, can you please open one and reference it here?

ethomson:
As mentioned, `http.cookiefile` support is a reasonable request, but this isn't the way we'd like to proceed with this.

PR #4179, opened 2017-03-23
Introduce home directory expansion function for config files, attribute files

### LINKED ISSUE #3681: include.path with non-existing file

ethomson:
When we need to expand a tilde-prefixed path to the home directory (eg `~/file.txt` -> `/home/ethomson/file.txt`) we currently use `git_sysdir_find_global_file`.  Unfortunately, this function will return `GIT_ENOTFOUND` if the file does not exist in the home directory, which can make the error handling odd.

As a result, we currently if a nonexistent config file is included with a `~` path, and if a non-existent attribute file is loaded by a `~`, we will (incorrectly) fall back to looking for a file with a literal `~`.

This introduces `git_sysdir_expand_global_file` so that callers can simply expand the filename and then use it in their normal error handling logic if it is ultimately not found.

This builds on the fixes in #3828, which this obsoletes.

Fixes #3681.

### DIFF
--- a/src/attrcache.c
+++ b/src/attrcache.c
@@ -290,14 +290,16 @@ static int attr_cache__lookup_path(
 		const char *cfgval = entry->value;
 
 		/* expand leading ~/ as needed */
-		if (cfgval && cfgval[0] == '~' && cfgval[1] == '/' &&
-			!git_sysdir_find_global_file(&buf, &cfgval[2]))
-			*out = git_buf_detach(&buf);
-		else if (cfgval)
+		if (cfgval && cfgval[0] == '~' && cfgval[1] == '/') {
+			if (! (error = git_sysdir_expand_global_file(&buf, &cfgval[2])))
+				*out = git_buf_detach(&buf);
+		} else if (cfgval) {
 			*out = git__strdup(cfgval);
+		}
 	}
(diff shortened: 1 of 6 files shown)


carlosmn:
The implementation of this seems backwards, no? Why are we trying to find a file on disk and then ignoring if it doesn't exist, instead of having a function to create the path and then try to use it in `_find_`?

ethomson:
We do have a function to create the path - that's what `git_sysdir_expand_global_file` does - and it does this regardless of whether the file exists.

What we're trying to find on disk is the home directory itself - we do that by calling `git_sysdir_find_global_file` with a `NULL` file portion.  It will populate the out buffer with only the home directory.

The `git_sysdir_expand_global_file` function basically exists to expand `~` (though only `~/`, not `~otheruser`, of course).  The config loading code can then continue with its own error handling (including file not found) that it shares with non-tilde-expanded paths.

PR #4180, opened 2017-03-24
Fix memory leaks

pks-t:
The recent addition of an error code to `pass_whole_blame` in ff8d2eb15
(blame_git: check return value of object lookup, 2017-03-20) introduced
a spurious goto. Remove it.

Just noticed a few thousands of new memory leaks in the CI. The error is trivial and kind of embarassing - I'm surprised the compiler didn't throw any warning for unreachable code.


pks-t:
Also added another commit fixing a new memory leak in `git_worktree_open_from_repository`.

PR #4183, opened 2017-03-28
Coverity

pks-t:
Another set of Coverity fixes. Some of these are mostly cosmetic fixes, some are actual fixes for rather unlikely error cases.

### DIFF
--- a/src/blame_git.c
+++ b/src/blame_git.c
@@ -517,11 +517,12 @@ static int pass_blame(git_blame *blame, git_blame__origin *origin, uint32_t opt)
 	if (!num_parents) {
 		git_oid_cpy(&blame->options.oldest_commit, git_commit_id(commit));
 		goto finish;
-	}
-	else if (num_parents < (int)ARRAY_SIZE(sg_buf))
+	} else if (num_parents < (int)ARRAY_SIZE(sg_buf))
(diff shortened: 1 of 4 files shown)


ethomson:
I have a few concerns about asserting that `mutex_lock` / `mutex_unlock` succeeded...  It seems like it should be something that we should return to the caller instead of `assert`ing, and I don't like that they're only compiled optionally.

I wonder if we could set a thread-local variable when these fail and check it again later?  I don't have a good feeling for how ugly it would be to do that after-the-fact checking, though.  `openssl_stream.c` doesn't seem crazy huge, but it still may be quite invasive.  A quick look around suggests that other users of openssl don't bother checking `pthread_mutex_` calls either.

pks-t:
I definitly agree regarding the locking. What is really to blame here is the OpenSSL API, which does not allow the locking function to return an error. The workaround with a thread-local variable might work, but I do not know off-hand which call paths may invoke the locking function at all. I'm a bit clueless on the design of this part of OpenSSL.

pks-t:
I've removed the OpenSSL-specific code for now. OpenSSL supports user error codes via `ERR_LIB_USER`, but it will take some time as I'm busy fighting the OpenSSL API and extracting any useful information out of it. *sigh*

PR #4184, opened 2017-03-29
patch: add support for partial patch application

hackhaslam:
This adds a callback to allow clients to choose which hunks are applied in `git_apply__patch`. The principal motivation is to enable an interface for staging individual hunks.

I see that a public interface for patch application was never merged, so this adds a possible interface for that as well. I don't know what the public interface should really look like. In particular, this doesn't pass out the filename and mode.

The only non-trivial thing here is that when hunks are skipped we have to keep track of how many lines were skipped to adjust the expected location of subsequent hunks. Other than that I think that it's straightforward.

I expanded a couple of existing tests and added a couple of new tests for skipping hunks.


ethomson:
Hmm.  So I think that a per-hunk callback is a fine idea.  But to be honest, I am not as excited about exposing `git_apply__patch` - and in fact it's not public because I'm not very excited about having an API that applies a `git_patch` to a `git_buf`.  Though we should _definitely_ have an API that applies a `git_diff` to various things.

Out of curiosity - are you running `git_patch_apply` in a loop?  😀 I worry that you're setting up `git_patch_apply` as public API because there's no `git_diff_apply` that takes a `git_diff` and applies it to a tree or a working directory or an index...

If that's the case then maybe we should prioritize `git_diff_apply`.  It's really pretty straightforward, to be honest, except for the tests.  But if it would be valuable then I'd be willing to merge a less rigorously tested version so that you can get moving on the dev against it.

If not, then my thoughts are:

1. I think that `git_apply__patch` should just get renamed `git_patch_apply` and you should handle (or ignore) the filename and mode outparams.

2. What do you think about exposing this in the `include/sys/` folder...?  I'm of two minds about that, but I thought that I would ask.

hackhaslam:
I'm not using it in a loop. My use case is staging individual hunks. When the user clicks on the hunks they want to stage, I create a patch from the status diff and use the old (HEAD) side as the source buffer. Then I call `git_patch_apply` to apply the selected hunks and stage the new buffer with `git_index_add_frombuffer`. It's interactive, so only the file that the user changes needs to be staged, not all files in the diff.

I think that applying a patch to a buffer make sense as a public function, even if there are additional convenience functions that do the higher-level stuff.

1. This sounds good to me.

2. What is the `include/sys` directory for? Is it for stuff that has to be public for one reason or another but isn't really meant to be part of the public API?

ethomson:
Okay, thanks for the information.  Like I said, I wasn't really sure that `git_patch_apply` was really an API that anybody would really want.  But if it's useful to you, then for sure let's make it public.

I don't think there's really a need to put it in `include/sys`.  I would describe it more as "things that are public but that are really edge case concerns"...  This isn't really that, this is a pretty straightforward and reasonable API.

So anyway - my inclination here is to make this `git_patch_apply` (I don't really know what I was thinking with `git_apply__patch`, but I was thinking that we'd figure out a name later.  😛 ) and give it out params.

hackhaslam:
Okay, thanks. I made those changes. I split it into two commits, one to add the hunk callback, and the second to change the name and make it public.

include/git2/patch.h
```diff
@@ -267,6 +267,35 @@ GIT_EXTERN(int) git_patch_to_buf(
...
+ * @param source_len The length of the source buffer
+ * @param hunk_cb A callback function to choose which hunks are applied.
+ *                Return <0 to abort, >0 to skip this hunk or 0 to apply.
+ *                Pass NULL to apply all hunks in the patch.
+ * @param payload The payload to pass to the hunk callback.
+ * @return 0 on success, <0 on failure.
+ */
+GIT_EXTERN(int) git_patch_apply(
+	git_buf *out,
+	char **filename,
```
pks-t: While exposing this function, we should rename `filename` to `path` or similar, as it is not the file name only.

include/git2/patch.h
```diff
@@ -267,6 +267,35 @@ GIT_EXTERN(int) git_patch_to_buf(
...
+/**
+ * Apply a patch to a buffer.
+ *
+ * Attempts to apply the changes represented by a patch to a source buffer
+ * producing a new output buffer. A hunk callback can be given to choose
+ * which hunks are applied.
+ *
+ * @param out The output buffer to be filled
+ * @param filename The output name of the new file
+ * @param mode The output mode of the new file
```
pks-t: Do the out-parameters necessarily have to be non-NULL and if so, does it make sense (hint: they currently have to be non-NULL, but I'm not sure if it should be required)?

tests/apply/partial.c
```diff
@@ -0,0 +1,84 @@
...
+static int skip_addition(
+	const git_diff_delta *delta,
+	const git_diff_hunk *hunk,
+	void *payload)
+{
+	GIT_UNUSED(delta);
+	GIT_UNUSED(payload);
+
+	return (hunk->new_lines > hunk->old_lines) ? 1 : 0;
+}
```
pks-t: Would be nice to also have a test which aborts adding hunks by returning `-1`. Additional kudos if aborting mid-way, that is after adding a few other hunks.

src/apply.c
```diff
@@ -211,19 +210,34 @@ static int apply_hunks(
...
 
 	git_array_foreach(patch->hunks, i, hunk) {
-		if ((error = apply_hunk(&image, patch, hunk)) < 0)
+		if (hunk_cb) {
+			/* abort on <0 */
+			if ((error = hunk_cb(patch->delta, &hunk->hunk, payload)) < 0)
+				goto done;
+
+			/* skip on >0 */
+			if (error > 0) {
```
ethomson: I wonder if `> 0` is a bit broad, is that necessary?  I wonder if there may be other non-error return codes to switch on in the future.  Maybe just look for `1`...?

And if so, it might make sense to define a value here - `GIT_PATCH_SKIP`?

pks-t: I had the same thought here. One potential usecase I thought of was when the callback returns `+n` that we want to skip the current hunk and the next `(n-1)` hunks. I guess though that this is quite uncommon and not really useful after all.

include/git2/patch.h
hackhaslam: Do you mean to make all of the out parameters optional or just the file path and mode? I suppose that if the user only cares about the result code (whether the patch can be applied cleanly) then it would make sense for the output buffer to be optional too.

pks-t: I only think it is worthwhile to have filename and mode optional. We usually do not allow for the "main" out parameter to be optional

s3lvin:
Very eagerly waiting for this PR to get accepted from a long time 👍 Because I'm also in the process of using this feature (staging lines in a hunk) from the library.

ethomson:
It seem like you're trying to use `skipped_lines` to update the location to find the hunk in the postimage.  You're doing that by subtracting the delta of the new line count (lines affected by a hunk in the preimage) from the old line count (lines affected by a hunk in the postimage).

This works only when there are only changes (`new_lines` == `old_lines`) or when there are lines being _added_ by a hunk (`new_lines` > `old_lines`).  This is an underflow when there are lines being removed (`new_lines` > `old_lines`).

I think this is as simple as using a signed type for `skipped_lines`...  but there may be other problems with that...

Seems like we also need some more tests here that skip various types of operations.

hackhaslam:
Ugh, you're right of course. I did all of my testing with `int` but for some reason I decided to change it to `size_t` just before I pushed the PR. I'll fix that, add some more test cases, and try to address the other feedback soon.

hackhaslam:
I changed  `skipped_lines` to `int`. I couldn't actually get it to fail with the unsigned type because of the way it converted from signed and back.

I think that I've addressed all of the other feedback as well. Thanks for the reviews!

pks-t:
As we always use `size_t` for our lines, this could overflow quite fast. I think it should at least be `ssize_t`, if we want to also encode negative values, and in addition we should be checking for an overflow when adding up the values.

hackhaslam:
I went ahead and changed the type of that variable to `ssize_t`, but the other lines involved (in `git_diff_hunk`) have type `int`.

s3lvin:
Guys, Any chance of merging this into master?

pks-t:
Looks good to me :+1: @ethomson?

ethomson:
Right, sorry.  I've been playing with this with more substantial patches and I'm not sure that what I'm seeing is correct.  Let me try to distill it into test cases.

dometto:
What is the status of this? I think various people would still be very interested in this functionality.

s3lvin:
+1 @dometto  I'm one of the various people who would be still VERY interested in this functionality. Any chance of merging this @ethomson ?

ethomson:
> I'm one of the various people who would be still VERY interested in this functionality. Any chance of merging this @ethomson ?

Which?  This patch does two things:

1. It makes patch application (not _diff_ application) public for a given preimage file.
2. It provides a callback to allow for partial patch application.

It does _not_ provide the ability to apply a full `git_diff` to anything.  That is to say that you cannot apply a "patch file" to a worktree.  (Apologies for the nomenclature - a `git_diff` is the equivalent of a "patch file", a `git_patch` is only a single file in the whole diff.)

Like I said, I'm rather meh on this API, I added it only so that we could unit test patch parsing and application.  Anyway.  I'll give this a bit more thought.  Indeed if it's useful, we should merge it.  But I wanted to set expectations properly.

Uncommon:
FWIW For applying hunks I ended up just implementing it myself; the logic is actually pretty simple.

ethomson:
Okay.  Thanks for the info.  Let me give this a little thought and maybe the API a little love so that we can get it merges.

s3lvin:
Awesome, thanks!

pravi:
I'm also interested to see this merged (to have rugged adapter as default in gollum-lib)

ethomson:
Thanks for the feedback.  We need a more comprehensive API here, rather than just bolting on this one callback.  I’m giving this some thought and will have some proposals after 0.27 is released.

Issue #4187, opened 2017-04-02
Documentation refers to nonexistent file


Storyyeller:
libgit2/include/git2/odb.h says "Read <odb_backends.h> for more information." 

However, this file does not exist.

### Version of libgit2 (release number or SHA1)
caf7a7a6b14b3b2f200f88f2f8d7a18ca10fd3ac (master)

ethomson:
Thanks for reporting this - this is now fixed via #4188

PR #4188, opened 2017-04-02
Correct non-existent file references in `odb.h`

rcjsuen:
This fixes issue #4187.


ethomson:
Thanks for fixing this!  I wonder if the confusion here isn't because the file is in the `sys` directory?  Maybe we should say `sys/odb_backend.h`...  what do you think?

rcjsuen:
Good point! Makes sense to me. I've force pushed your suggested change to my branch.

ethomson:
👍 Thanks for taking care of this!

PR #4191, opened 2017-04-05
Branch renames with worktrees

pks-t:
When renaming a branch in a repository, we have to check whether any linked worktree has that branch checked out. If so, we have to update the worktree's HEAD file to point to the renamed branch. This pull requests implements this behavior, which is the same as in upstream git.git.

This PR is currently a WIP. While it works right now, I tend to find it quite ugly. My main grief is reading the worktree HEADs. I couldn't find any way to reliably parse HEAD files at arbitrary paths, especially when it is a symref pointing to a non-existing (read as "just renamed") branch. That's why I implemented `git_reference_read_head` as well as `read_head_file`, but to be honest this feels as if I'm re-inventing the wheel. I'm not sure whether it's my misuse of the API or if we don't have support for reading symref HEAD files at arbitrary locations without validating and resolving them.

### DIFF
--- a/src/branch.c
+++ b/src/branch.c
@@ -127,61 +127,28 @@ int git_branch_create_from_annotated(
 		repository, branch_name, commit->commit, commit->description, force);
 }
 
-int git_branch_is_checked_out(
-	const git_reference *branch)
+static int branch_equals(git_repository *repo, const char *path, void *payload)
 {
-	git_buf path = GIT_BUF_INIT, buf = GIT_BUF_INIT;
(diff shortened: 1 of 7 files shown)


pks-t:
By the way: I'll reword commit messages as well as fix naming of internal functions to have "__" separators, e.g. in "git_repository__foreach_head" and "git_reference__read_head".

pks-t:
I think this should be mostly ready now. At least it's not as ugly anymore as it has previously been.

Issue #4194, opened 2017-04-06
still build issues with openssl-1.1 that does not provide deprecatred features anymore


Polynomial-C:
### Reproduction steps
* Build openssl-1.1 with `--api=1.1.0` or `no-deprecated` switch
* Try to build libgit2

### Expected behavior
libgit2 builds successfully

### Actual behavior
```
/usr/bin/x86_64-pc-linux-gnu-gcc -DGIT_ARCH_64 -DGIT_CURL -DGIT_OPENSSL -DGIT_THREADS -DGIT_USE_NSEC -DGIT_USE_STAT_MTIM -DHAVE_FUTIMENS -DHAVE_QSORT_R -DOPENSSL_SHA1 -D_FILE_OFFSET_BITS=64 -Dgit2_EXPORTS -I/var/tmp/portage/dev-libs/libgit2-0.25.1/work/libgit2-0.25.1/src -I/var/tmp/portage/dev-libs/libgit2-0.25.1/work/libgit2-0.25.1/include   -DNDEBUG -D_GNU_SOURCE -Wall -Wextra -march=native -mtune=native -O2 -pipe -fvisibility=hidden -fPIC -Wno-missing-field-initializers -Wstrict-aliasing=2 -Wstrict-prototypes -Wdeclaration-after-statement -Wno-unused-function -fPIC   -o CMakeFiles/git2.dir/src/openssl_stream.c.o -c /var/tmp/portage/dev-libs/libgit2-0.25.1/work/libgit2-0.25.1/src/openssl_stream.c
/var/tmp/portage/dev-libs/libgit2-0.25.1/work/libgit2-0.25.1/src/openssl_stream.c: In function ‘openssl_locking_function’:
/var/tmp/portage/dev-libs/libgit2-0.25.1/work/libgit2-0.25.1/src/openssl_stream.c:52:16: error: ‘CRYPTO_LOCK’ undeclared (first use in this function)
  lock = mode & CRYPTO_LOCK;
                ^
/var/tmp/portage/dev-libs/libgit2-0.25.1/work/libgit2-0.25.1/src/openssl_stream.c:52:16: note: each undeclared identifier is reported only once for each function it appears in
/var/tmp/portage/dev-libs/libgit2-0.25.1/work/libgit2-0.25.1/src/openssl_stream.c: In function ‘shutdown_ssl_locking’:
[...]

PR #4195, opened 2017-04-07
Fix building against OpenSSL v1.1

pks-t:
This fixes the issues raised by #4194. When bulding against OpenSSL without deprecated features, we currently fail. This can be provoked by defining "-DOPENSSL_API_COMPAT=0x10100000L" when OpenSSL 1.1 is installed.

### DIFF
--- a/src/openssl_stream.c
+++ b/src/openssl_stream.c
@@ -103,8 +103,13 @@ int git_openssl_stream_global_init(void)
 	ssl_opts |= SSL_OP_NO_COMPRESSION;
 #endif
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
 	SSL_load_error_strings();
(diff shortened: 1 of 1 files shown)


src/openssl_stream.c
```diff
@@ -146,7 +151,10 @@ int git_openssl_stream_global_init(void)
 
 int git_openssl_set_locking(void)
 {
-#ifdef GIT_THREADS
+#ifndef GIT_THREADS
+	giterr_set(GITERR_THREAD, "libgit2 was not built with threads");
+	return -1;
```
pks-t: Just realized this does not make any sense when building against 1.1. I should probably just return success if `GIT_THREADS && OPENSSL_VERSION >= 0x10100000L`.

PR #4196, opened 2017-04-07
filter: only close filter if it's been initialized correctly

pks-t:
In the function `git_filter_list_stream_data`, we initialize, write and
subesquently close the stream which should receive content processed by
the filter. While we skip writing to the stream if its initialization
failed, we still try to close it unconditionally -- even if the
initialization failed, where the stream might not be set at all, leading
us to segfault.

Semantics in this code is not really clear. The function handling the
same logic for files instead of data seems to do the right thing here in
only closing the stream when initialization succeeded. When stepping
back a bit, this is only reasonable: if a stream cannot be initialized,
the caller would not expect it to be closed again. So actually, both
callers of `stream_list_init` fail to do so. The data streaming function
will always close the stream and the file streaming function will not
close the stream if writing to it has failed.

The fix is thus two-fold:

- callers of `stream_list_init` now close the stream iff it has been
  initialized
- `stream_list_init` now closes the lastly initialized stream if
  the current stream in the chain failed to initialize

Add a test which segfaulted previous to these changes.

### DIFF
--- a/src/filter.c
+++ b/src/filter.c
@@ -911,14 +911,19 @@ static int stream_list_init(
 				last_stream);
 
 		if (error < 0)
-			return error;
+			goto out;
 
 		git_vector_insert(streams, filter_stream);
 		last_stream = filter_stream;
 	}
 
-	*out = last_stream;
-	return 0;
+out:
+	if (error)
+		last_stream->close(last_stream);
+	else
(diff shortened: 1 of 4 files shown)


pks-t:
This extends the incomplete fix provided in #3659

PR #4197, opened 2017-04-10
Verify object hashes

pks-t:
I've just been triggered by [1] to actuall check how we behave with corrupted on-disk objects. The good news is that we actually detect simple bit-flips in loose objects, but this comes for free through the zlib library, which is unable to restore contents. I've added a test which repeatedly corrupts all the bytes in an object and then tries to read it, and this one succeeds without any change whatsoever.

The bad news is that we actually fail to verify valid objects which are saved under the wrong name. E.g. when a commit "adeadbeef" is not stored in ".git/objects/ad/eabdeef" but in "git/objects/bd/eadbeef", we did not complain. Upstream git.git does actually detect this case and complain.

So I've implemented the upstream behavior, so that we now also compute and verify the object's hash after it's been read from the ODB backend. I've expected this to have an actual performance impact, but at least for our test suite it seems to be drowned out by the actual I/O operations, which weigh far heavier. So I couldn't measure an actual performance hit for our test suite -- this might be different when for example having huge binaries, but I guess people have all kinds of problems then.

If it might prove to be too much of a burden in certain scenarios, we could also introduce a switch later on, which optionally turns off verification of read objects. But as of now, I don't think it worthwhile.

[1]: http://public-inbox.org/git/4D74C1D4-9EA7-4A17-AFC5-0B54B4A6DD0E@gmail.com/

### DIFF
--- a/include/git2/common.h
+++ b/include/git2/common.h
@@ -182,6 +182,7 @@ typedef enum {
 	GIT_OPT_ENABLE_SYNCHRONOUS_OBJECT_CREATION,
 	GIT_OPT_GET_WINDOWS_SHAREMODE,
 	GIT_OPT_SET_WINDOWS_SHAREMODE,
+	GIT_OPT_ENABLE_STRICT_HASH_VERIFICATION,
 } git_libgit2_opt_t;
 
 /**
@@ -336,6 +337,13 @@ typedef enum {
  *		> (or the platform equivalent) to ensure that new object data
  *		> is written to permanent storage, not simply cached.  This
  *		> defaults to disabled.
+ *
+ *	 opts(GIT_OPT_ENABLE_STRICT_HASH_VERIFICATION, int enabled)
+ *
+ *		> Enable strict verification of object hashsums when reading
+ *		> objects from disk. This may impact performance due to an
+ *		> additional checksum calculation on each object. This defaults
+ *		> to enabled.
  *
  * @param option Option key
  * @param ... value to set the option
--- a/include/git2/errors.h
+++ b/include/git2/errors.h
@@ -54,6 +54,7 @@ typedef enum {
 	GIT_PASSTHROUGH     = -30,	/**< Internal only */
 	GIT_ITEROVER        = -31,	/**< Signals end of iteration with iterator */
 	GIT_RETRY           = -32,	/**< Internal only */
+	GIT_EMISMATCH       = -33,	/**< Hashsum mismatch in object */
 } git_error_code;
 
 /**
--- a/src/odb.c
+++ b/src/odb.c
@@ -1146,6 +1146,18 @@ static int read_prefix_1(git_odb_object **out, git_odb *db,
 	if (!found)
 		return GIT_ENOTFOUND;
 
+	if (git_odb__strict_hash_verification) {
+		git_oid hash;
+
+		if ((error = git_odb_hash(&hash, raw.data, raw.len, raw.type)) < 0)
+			goto out;
+
+		if (!git_oid_equal(&found_full_oid, &hash)) {
+			error = git_odb__error_mismatch(&found_full_oid, &hash);
+			goto out;
+		}
+	}
+
 	if ((object = odb_object__alloc(&found_full_oid, &raw)) == NULL)
 		goto out;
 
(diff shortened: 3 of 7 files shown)


ethomson:
Seems right to me.  Should we also do this for the streaming read functions, and raise an error when you `close` the stream if the hashes did not match?

My hunch is that this won't destroy perf, but @carlosmn might be interested in a toggle or something as protection?

pks-t:
We probably should, yes. But it's not that easy to implement here, as `git_odb_open_rstream` does not get to see the content of the streamed object at any point in time. So I guess we'd have to shift this burden on the backend itself -- and as we've got no backend implementing streaming read, I cannot really do anything about it right now.

Or do you have different ideas?

pks-t:
Just a short re to the performance impact: the additional hashing shows a lot when loading big objects, as predicted. Executing our exemplary `cat-file -e <BLOB>` with a 2GB file shows a performance decrease of about 35%, increasing from around 4.8s to 7.2s. So a switch would probably make sense here

carlosmn:
I do worry about the performance impact of this. Our test suite and serving data at scale are pretty different. We rely on keeping interesting bits of repositories in RAM quite a bit, where the I/O cost becomes tiny in comparison with single-use repositories like we use in the test suite.

@peff might have something to chime in here, but I would certainly appreciate a way to disable these checks, as uses of libgit2 are often under much tighter time constraints.

carlosmn:
@pks-t what kind of tests were you doing? The email you refer to is about fetching, where git checks whether we have the object locally and compares the incoming one against the locally-available object. That kind of checking would not happen on each read, but would be a check that exists in the indexer rather than in every read path.

peff:
I don't have much to chime in, except to say that in my informal timings, SHA-1 speed is usually drowned out even in warm-cache situations. Keep in mind that we're zlib inflating the data at the very least, and possibly reconstructing it from deltas as well.

Git always checks the sha1 on trees and commits. I don't think it generally does on blobs (e.g., when doing diffs), and that is the one place where I've noticed that the sha1 speed can have some impact (albeit not that huge).

pks-t:
Yeah, the mail has only been triggering me to take a look if we perform any checks at all when reading the ODB. E.g. with git:

```
$ git init /tmp/foo &&
$ cd /tmp/foo &&
$ touch x &&
$ git add x &&
$ git commit -mx x
$ mv .git/objects/de/adbeefdeadbeef...bee{f,1}
$ git show deadbeefdeadbeef...bee1
error: sha1 mismatch deadbeef...bee1
fatal: bad object deadbeef...bee1
```

With libgit2, we never notice the SHA1 mismatch, and this PR fixes the issue.

So I only did two tests here. The first one, executing the test suite, didn't show any impact. But as you said, it's quite a different load than real-world repositories or even reading some large blobs from ODB. The second test was to run our exemplary `cat-file` implementation on a 2GB blob, which did show a slowdown of around 35%.

So I think we should have some kind of switch to enable/disable this. I'm not too certain how to implement this, though. E.g. we could either have it per command or globally.

ethomson:
> We probably should, yes. But it's not that easy to implement here, as git_odb_open_rstream does not get to see the content of the streamed object at any point in time.

No, but I was thinking that all callers invoked `git_odb_stream_read` and (ultimately) `git_odb_stream_free` and we could do the sha1 calculations there.  But I haven't actually tried this, I was just thinking that we provided the actual functions that callers use and could do the sha1 calculation there.

> So I think we should have some kind of switch to enable/disable this. I'm not too certain how to implement this, though. E.g. we could either have it per command or globally.

I think that a global option is fine.  I think that if various large hosting providers want to turn it off, then they'll want to turn it off wholesale.

pks-t:
I've added a global option to disable hash verification

pks-t:
So this whole git_odb_stream is not really useful at all as it stands currently. We have no real documentation, no implementation and the interface does not even have an obvious way how one can know whether the read stream is at its end or not. So I will not implement verification in this PR but will do it in another PR together with a test implementation and potentially also some interface changes to make it actually usable.

ethomson:
Perhaps we should reconsider its existence at all.

PR #4199, opened 2017-04-10
checkout.h: fix warning when compiling with -Wpedantic

Vaelatern:
This fixes compiler warnings when compiling with `-Wpedantic` with gcc-6.3.0

Test case that demonstrates this concept:
https://gist.github.com/anonymous/7723e6bb90b6721e1b5013c20febf310

I applied this patch against master, but also applied it to version 0.25.1

### DIFF
--- a/include/git2/checkout.h
+++ b/include/git2/checkout.h
@@ -112,7 +112,6 @@ typedef enum {
 	/** Allow all updates to force working directory to look like index */
 	GIT_CHECKOUT_FORCE = (1u << 1),
 
-
 	/** Allow checkout to recreate missing files */
 	GIT_CHECKOUT_RECREATE_MISSING = (1u << 2),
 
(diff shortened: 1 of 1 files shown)


ethomson:
I'm surprised that `-Wpedantic` warns here - there's nothing wrong with skipping field initializers, anything that's not explicitly assigned will implicitly be assigned `0`.

Tagged initializers are new to C99, and we keep compatibility with prior versions for maximum compatibility.

Vaelatern:
There should be nothing wrong here, but then again, it does complain. Maybe I'll go knock on gcc's door.

Vaelatern:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80441

Vaelatern:
GCC closed my issue as not a problem because gcc behaves as intended.
I've decided to add `-Wno-missing-field-initializers` to my Makefile to compensate.
I will close this PR because you have stated the importance of maintaining compatibility where possible.

PR #4210, opened 2017-04-21
Misc fixes

pks-t:
Two small fixes I stumbled upon while re-rolling the object verification PR.


pks-t:
Merged myself as the fixes are trivial.

PR #4211, opened 2017-04-21
travis: upgrade container to Ubuntu 14.04

pks-t:
Ubuntu 12.04 (Precise Pangolin) reaches end of life on April 28th, 2017.
As such, we should update our build infrastructure to use the next
available LTS release, which is Ubuntu 14.04 LTS (Trusty Tahr). Note
that Trusty is still considered beta quality on Travis. But considering
we are able to correctly build and test libgit2, this seems to be a
non-issue for us.

Switch over our default distribution to Trusty. As Precise still has
extended support for paying customers, add an additional job which
compiles libgit2 on the old release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ addons:
     - openssh-server
     - valgrind
 
+dist: trusty
 sudo: false
 
 matrix:
@@ -38,13 +39,19 @@ matrix:
      compiler: gcc
  include:
    - compiler: gcc
+     env: PRECISE=1
+     os: linux
+     dist: precise
+   - compiler: gcc
      env: COVERITY=1
      os: linux
+     dist: trusty
    - compiler: gcc
      env:
        - VALGRIND=1
          OPTIONS="-DBUILD_CLAR=ON -DBUILD_EXAMPLES=OFF -DDEBUG_POOL=ON -DCMAKE_BUILD_TYPE=Debug"
      os: linux
+     dist: trusty
  allow_failures:
    - env: COVERITY=1
 
--- a/script/cibuild.sh
+++ b/script/cibuild.sh
@@ -43,23 +43,39 @@ ctest -V -R libgit2_clar || exit $?
 
 killall git-daemon
 
-if [ "$TRAVIS_OS_NAME" = "osx" ]; then
-    echo 'PasswordAuthentication yes' | sudo tee -a /etc/sshd_config
-fi
+# Set up sshd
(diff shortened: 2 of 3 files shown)


pks-t:
Well. Seems I've been too keen on this one. We certainly cannot "correctly build and test", as there is an issue with the SSH key exchange. Will investigate

pks-t:
So the problem is related to the SSH server not running in the new images. Previously, it has only been working by accident: by default, sshd is already installed but not running in Precise images. But as the version is out-of-date, we upgrade openssh-server and then also start it in the process. For Trusty, the package is already up-to-date and as such will not be reinstalled. As a consequence, no post-upgrade hooks are executed and sshd will not get executed.

Unfortunately, we are unable to start the sshd executable ourselves, as we currently run in sudo-less containers. Three possibilities:

1. query the Travis team
2. disable SSH-related tests if the server is not running
3. switch from sudo-less containers to a fully virtualized virtual machine

The second option is not really an option in the long run, I guess. The third option has downsides regarding performance, as the initial startup of the VM has around ~45 seconds overhead. There'll probably more performance degradations at runtime due to being virtualized. So the first option is obviously the best one, but will take longer than the others.

pks-t:
I take that back. As it seems, sshd _is_ running... back to zero

pks-t:
So the issue is that sshd is missing host keys. We're not able to generate them, as again we're running sudoless. I've created an issue with Travis at travis-ci/travis-ci#7661.

pks-t:
This sucks. We actually have one failing test with the poxyproxy. It causes a segfault in libcurl when it checks whether the established connection is already authenticated via NLTM. While our parameters seem to be set correctly, the connection which is checked against has `NULL` user and passwd, which are then passed to `strcmp`.

So I'm not even sure if it is us or libcurl. I _think_ it's libcurl's fault, but I'm not too sure about this. Dunno what to do about it, though.

pks-t:
Seems they fixed the build environment now. I've done multiple builds, no failures up to now. So I guess this is ready to be merged

ethomson:
Excellent.  Let's merge this right after we ship an 0.26.0.

pks-t:
Agreed. One potential benefit by upgrading is that the libcurl warning is not present anymore, so we are even able to enable `-Werror` for Trusty containers. There are a few other warnings, but they are trivial to fix.

pks-t:
Dropping the `-Werror` patch for now. I'll probably create a new PR later to switch it on.

Issue #4134, opened 2017-02-21
Should git_config_set_* skip readonly backends?


kballard:
I'm reading through the config.c code and it looks like the `git_config_set_*` functions all attempt to set the value on the highest-level backend, and if that fails it returns an error. I'm inclined to say this function should explicitly skip any backends that are marked as `readonly`. This way the app can add an application-level backend that doesn't support writes, mark it as readonly, and the `git_config_set_*` functions will behave as though the app-level backend wasn't added instead of throwing an error.

ethomson:
I agree.  I'm not sure why config backends can be read-only if we're not going to honor that.

ethomson:
Fixed via #4217

PR #4217, opened 2017-04-25
Honor read-only flag when writing to config backends

pks-t:
Configuration backends have a readonly-flag which is currently used to
distinguish configuration snapshots. But somewhat unexpectedly, we do
not use the flag to prevent writing to a readonly backend but happily
proceed to do so.

This commit modifies logic to also honor the readonly flag for
configuration setters. We will now traverse through all backends and
pick the first one which is not marked as read-only whenever we want to
write new configuration.

Implements #4134

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -576,22 +576,50 @@ int git_config_foreach_match(
  * Setters
  **************/
 
-static int config_error_nofiles(const char *name)
+typedef enum {
+	BACKEND_USE_SET,
+	BACKEND_USE_DELETE
+} backend_use;
+
+static const char *uses[] = {
+    "set",
+    "delete"
+};
+
+static int get_backend_for_use(git_config_backend **out,
+	git_config *cfg, const char *name, backend_use use)
 {
+	size_t i;
+	file_internal *f;
+
+	*out = NULL;
+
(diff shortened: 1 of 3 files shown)


src/config.c
```diff
@@ -576,22 +576,37 @@ int git_config_foreach_match(
...
+static int get_writable_backend(git_config_backend **out, git_config *cfg, const char *name)
 {
+	size_t i;
+	file_internal *f;
+
+	*out = NULL;
+
+	if (git_vector_length(&cfg->files) == 0) {
+		giterr_set(GITERR_CONFIG,
+			"cannot set value for '%s' when no config files exist", name);
```
ethomson: If I can bikeshed here just a touch, it might be nice to switch this message on whether we're in a `set` function or in `delete_entry`.

pks-t:
Okay, I've extended the function to accept a use-case. Right now it feels a bit overengineered, so I initially thought about extending the function to also handle the other use cases (lock/unlock and read). It would then be some kind of "get me the right thing for whatever I want to do" function, so it would also honor the readonly flag. But the interface would look horrible. So I refrained from doing so and instead stayed with this version.

ethomson:
Seems fine to me.  Thanks!

PR #4219, opened 2017-04-26
socket_stream: continue to next addrinfo on socket creation failure

pks-t:
When connecting to a remote via socket stream, we first use getaddrinfo
to obtain the possible connection methods followed by creating and
connecting the socket. But when creating the socket, we error out as
soon as we get an invalid socket instead of trying out other address
hints returned by addrinfo.

Fix this by continuing on invalid socket instead of returning an error.
This fixes connection establishment with musl libc.


ethomson:
Neat, thanks!

PR #4224, opened 2017-04-29
Do not free config when creating remote

yuyichao:
The regression was introduced in 22261344de18b3cc60ee6937468d66a6a6a28875


ethomson:
Yep; we're getting a weakptr to the `config`, we don't need to `free` it.  Thanks!

PR #4225, opened 2017-04-29
Allow NULL refspec in git_remote_push

yuyichao:
We've been (incorrectly) using this in julia and see no errors in release builds. Given that `refspec` is only passed to `git_remote_upload` which allows a `NULL` `refspec` I think this assertion can be relaxed.

Ref https://github.com/JuliaLang/julia/issues/21639


ethomson:
This seems reasonable.  I've updated the documentation to reflect that this is allowed, and merged this manually.

PR #4228, opened 2017-05-02
Worktree options

pks-t:
The initial inspiration of this was the recently added `git worktree add --locked`, which I've added in this PR. While at it I also changed signatures of the yet-unreleased "git_worktree_add", "git_worktree_prune" and "git_worktree_is_prunable" functions to accept options structures, so that these functions are potentially extensible without breaking the API in future releases.

### DIFF
--- a/include/git2/worktree.h
+++ b/include/git2/worktree.h
@@ -74,6 +74,25 @@ GIT_EXTERN(void) git_worktree_free(git_worktree *wt);
  */
 GIT_EXTERN(int) git_worktree_validate(const git_worktree *wt);
 
+typedef struct git_worktree_add_options {
+	unsigned int version;
+} git_worktree_add_options;
+
+#define GIT_WORKTREE_ADD_OPTIONS_VERSION 1
+#define GIT_WORKTREE_ADD_OPTIONS_INIT {GIT_WORKTREE_ADD_OPTIONS_VERSION}
+
+/**
+ * Initializes a `git_worktree_add_options` with default vaules.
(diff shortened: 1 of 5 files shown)


include/git2/worktree.h
```diff
@@ -137,23 +161,46 @@ typedef enum {
 	GIT_WORKTREE_PRUNE_WORKING_TREE = 1u << 2,
 } git_worktree_prune_t;
 
+typedef struct git_worktree_prune_options {
+	unsigned int version;
+
+	char prune_valid; /**< prune even if working tree is valid */
+	char prune_locked; /**< prune even if working tree is locked */
+	char delete_worktree; /**< delete complete working directory */
```
ethomson: Do we use a `char` anywhere else to represent a boolean value within a struct?  I feel like the common pattern for our public API would be to have a `flags` value in the options struct...  for example, in [`diff`](https://github.com/libgit2/libgit2/blob/master/include/git2/diff.h#L290).

(Internally we use `int` bitfields in places, but we don't do that in the public API.)

pks-t: Okay, I'll change it back to use flags again, but will keep them inside the struct.

ariksugeng:
Can ask for setting theme and product on my web

Pada tanggal Jum, Mei 5, 2017, 19:33 Patrick Steinhardt <
notifications@github.com> menulis:

ethomson:
I changed the `lock` value from a `char` to an `int` and merged this manually.  Thanks!

PR #4229, opened 2017-05-02
libssh2 shutdown

pks-t:
A few small cleanups regarding libssh2 and git_buf_clear/git_buf_free confusion.

The libssh2 code did not call libssh2_exit on shutdown, which I've fixed. This still leaves us with memory leaks in libssh2/libgcrypt, though, which I cannot track down with the CI only. I'll try to fire up a VM to hunt down the remaining issues, as my host system is incompatible with valgrind due to musl libc. :/

### DIFF
--- a/src/global.c
+++ b/src/global.c
@@ -22,7 +22,7 @@
 
 git_mutex git__mwindow_mutex;
 
-#define MAX_SHUTDOWN_CB 8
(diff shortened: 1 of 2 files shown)


pks-t:
Oh, some are fixed already. Though I've seen those fixes before and have been a bit confused here -- simply forgot to fetch...

PR #4231, opened 2017-05-02
revparse: support open-ended ranges

### LINKED ISSUE #4223: One-sided revision ranges not supported
Command-line git supports leaving one side of a revision range blank, like `origin/master...`. When unspecified, the revision defaults to HEAD. Currently, libgit2 doesn't support these revision formats.

### Reproduction steps

Run `examples/rev-parse` with one-sided revision ranges.

### Expected behavior

The same as command-line git:

```
$ git rev-parse ..HEAD~1
d476d024d0415a566b858daec7faa961587bae14
^f9d3b0d05f19a59e3327deabc2edc93e507c6aa3

$ git rev-parse HEAD~1..
f9d3b0d05f19a59e3327deabc2edc93e507c6aa3
^d476d024d0415a566b858daec7faa961587bae14
```

### Actual behavior

```
$ examples/rev-parse ..HEAD~1
Could not parse '..HEAD~1' [-12] - failed to parse revision specifier - Invalid pattern ''

$ examples/rev-parse HEAD~1..
Could not parse 'HEAD~1..' [-12] - failed to parse revision specifier - Invalid pattern ''
```

v0.16.0-8230-g89141d9

MacOS 10.12.4

wabain:
Support '..' and '...' ranges where one or both sides are not
specified. The unspecified side defaults to HEAD.

Closes #4223

### DIFF
--- a/src/revparse.c
+++ b/src/revparse.c
@@ -892,6 +892,17 @@ int git_revparse(
 		const char *rstr;
 		revspec->flags = GIT_REVPARSE_RANGE;
 
+		/*
+		 * Following git.git, don't allow '..' because it makes command line
+		 * arguments which can be either paths or revisions ambiguous when the
+		 * path is almost certainly intended. The empty range '...' is still
+		 * allowed.
+		 */
+		if (!git__strcmp(spec, "..")) {
+			giterr_set(GITERR_INVALID, "Invalid pattern '..'");
+			return GIT_EINVALIDSPEC;
+		}
+
 		lstr = git__substrdup(spec, dotdot - spec);
 		rstr = dotdot + 2;
 		if (dotdot[2] == '.') {
@@ -899,9 +910,17 @@ int git_revparse(
 			rstr++;
 		}
 
-		error = git_revparse_single(&revspec->from, repo, lstr);
-		if (!error)
-			error = git_revparse_single(&revspec->to, repo, rstr);
+		error = git_revparse_single(
+			&revspec->from,
+			repo,
+			*lstr == '\0' ? "HEAD" : lstr);
+
+		if (!error) {
+			error = git_revparse_single(
+				&revspec->to,
+				repo,
+				*rstr == '\0' ? "HEAD" : rstr);
+		}
 
 		git__free((void*)lstr);
 	} else {
--- a/tests/refs/revparse.c
+++ b/tests/refs/revparse.c
@@ -122,6 +122,14 @@ static void test_id(
 	test_id_inrepo(spec, expected_left, expected_right, expected_flags, g_repo);
 }
 
+static void test_invalid_revspec(const char* invalid_spec)
+{
+	git_revspec revspec;
+
+	cl_assert_equal_i(
+		GIT_EINVALIDSPEC, git_revparse(&revspec, g_repo, invalid_spec));
+}
+
 void test_refs_revparse__initialize(void)
 {
 	cl_git_pass(git_repository_open(&g_repo, cl_fixture("testrepo.git")));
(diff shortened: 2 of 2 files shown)


ethomson:
Look at those beautiful tests!  😍 

Thanks for jumping in to fix this.  This is great, and I appreciate the pull request!

src/revparse.c
```diff
@@ -899,9 +899,17 @@ int git_revparse(
...
+			&revspec->from,
+			repo,
+			*lstr == '\0' ? "HEAD" : lstr);
+
+		if (!error) {
+			error = git_revparse_single(
+				&revspec->to,
+				repo,
+				*rstr == '\0' ? "HEAD" : rstr);
+		}
```
pks-t: Note that this also supports the ranges ".." and "...", that is no limiter at all. While these seem nonsensical, it even seems to be the right thing as git.git also accepts these ranges:

```
$ git rev-parse ..
..
$ git rev-parse ...
a8c90ef62281db933118aa84489eb0e1e9cc347c
a8c90ef62281db933118aa84489eb0e1e9cc347c
^a8c90ef62281db933118aa84489eb0e1e9cc347c
```

While the first result stumps me, there at least is no error here.

Could you maybe add another two tests for these special-cases?

tests/refs/revparse.c
```diff
@@ -749,6 +749,26 @@ void test_refs_revparse__parses_range_operator(void)
...
+
+	test_id("..HEAD~3",
+		"a65fedf39aefe402d3bb6e24df4d4f5fe4547750",
+		"4a202b346bb0fb0db7eff3cffeb3c70babbd2045",
+		GIT_REVPARSE_RANGE);
+
+	test_id("...HEAD~3",
+		"a65fedf39aefe402d3bb6e24df4d4f5fe4547750",
+		"4a202b346bb0fb0db7eff3cffeb3c70babbd2045",
+		GIT_REVPARSE_RANGE | GIT_REVPARSE_MERGE_BASE);
```
pks-t: Could you add two additional tests here to test the two special-cases of ".." and "...", please?

src/revparse.c
wabain: I hadn't thought about it much because `man gitrevisions` seems to explicitly say that the empty ranges are allowed. ("Note that .. would mean HEAD..HEAD which is an empty range that is both reachable and unreachable from HEAD.") However, on inspection, git.git treats `..` as a special case [on purpose](https://github.com/git/git/commit/003c84f6d2b9e9c4d5bbf5262cae994bac7190cb) so that it's always handled as a path for potentially ambiguous commands like `git log ..`. If the argument is disambiguated, `git rev-parse` gives an error:

```
$ git rev-parse .. --
fatal: bad revision '..'
```

I think it makes sense to reject `..`, since trying to parse an argument as a revision and then as a path is a pretty common git CLI idiom. Continuing to allow `...` seems a bit weird, but it should probably be accepted since git.git supports it.

pks-t: Oh, that's what's going on here. The paragraph you probably mean from gitrevisions(7):

```
       In these two shorthand notations, you can omit one end and let it default to HEAD.
       For example, origin.. is a shorthand for origin..HEAD and asks "What did I do
       since I forked from the origin branch?" Similarly, ..origin is a shorthand for
       HEAD..origin and asks "What did the origin do since I forked from them?" Note that
       .. would mean HEAD..HEAD which is an empty range that is both reachable and
       unreachable from HEAD.
```

So it explicitly says that one can omit one end, only. So I think your reasoning about us rejecting when both sides are empty is correct.

wabain:
I've updated to match git.git's behavior for the special cases '..' and '...'.

ariksugeng:
Can ask for setting theme and product on my web

Pada tanggal Jum, Mei 5, 2017, 23:14 William Bain <notifications@github.com>
menulis:

Issue #4186, opened 2017-03-31
Test cases fail on Debian/Ubuntu


bdrung:
I tried to update the Debian package of libgit2, but some test cases fail on Debian testing:

```
  1) Failure:
core::features::0 [/<<PKGBUILDDIR>>/tests/core/features.c:20]
  Expression is not true: (caps & GIT_FEATURE_HTTPS) != 0

  2) Failure:
repo::discover::discovering_repository_with_alternative_gitdir_succeeds [/<<PKGBUILDDIR>>/tests/repo/discover.c:36]
  Function call failed: (git_repository_discover(&found_path, start_path, 0, ceiling_dirs))
  error -3 - could not find repository from 'temprepo/discover.git/alternate_sub_repo/sub'

  3) Failure:
repo::discover::ceiling_should_not_affect_gitdir_redirection [/<<PKGBUILDDIR>>/tests/repo/discover.c:36]
  Function call failed: (git_repository_discover(&found_path, start_path, 0, ceiling_dirs))
  error -3 - could not find repository from 'temprepo/discover.git/alternate_sub_repo/sub'

  4) Failure:
repo::discover::discovery_starting_at_file_succeeds [/<<PKGBUILDDIR>>/tests/repo/discover.c:36]
  Function call failed: (git_repository_discover(&found_path, start_path, 0, ceiling_dirs))
  error -3 - could not find repository from 'temprepo/discover.git/sub_repo/file'

  5) Failure:
repo::open::force_bare [/<<PKGBUILDDIR>>/tests/repo/open.c:401]
  Function call failed: (git_repository_open_ext( &barerepo, "alternate/subdir/sub2", GIT_REPOSITORY_OPEN_BARE, ((void *)0)))
  error -3 - could not find repository from 'alternate/subdir/sub2'
```

Failure 2 to 5 can be only seen when running the tests inside sbuild (which uses schroot and overlay directories in my case).

### Reproduction steps

1. Download the packaging from Debian and extract it into the directory `debian` in your libgit2 directory:
http://http.debian.net/debian/pool/main/libg/libgit2/libgit2_0.24.5-1.debian.tar.xz
2. run `dpkg-buildpackage -b` (or better use sbuild)

### Version of libgit2 (release number or SHA1)

Version 0.25.1 and current head (caf7a7a6b14b3b2f200f88f2f8d7a18ca10fd3ac) are affected.

### Operating system(s) tested

Debian testing and Ubuntu 16.10 (yakkety)

[...]

PR #4232, opened 2017-05-03
tests: repo: fix repo discovery tests on overlayfs

pks-t:
Debian and Ubuntu often use schroot to build their DEB packages in a
controlled environment. Depending on how schroot is configured, our
tests regarding repository discovery break due to not being able to find
the repositories anymore. It turns out that these errors occur when the
schroot is configured to use an overlayfs on the directory structures.

The reason for this failure is that we usually refrain from discovering
repositories across devices. But unfortunately, overlayfs does not have
consistent device identifiers for all its files but will instead use the
device number of the filesystem the file stems from. So whenever we
cross boundaries between the upper and lower layer of the overlay, we
will fail to properly detect the repository and bail out.

This commit fixes the issue by enabling cross-device discovery in our
tests. While it would be preferable to have this turned off, it probably
won't do much harm anyway as we set up our tests in a temporary location
outside of the parent repository.

### DIFF
--- a/tests/repo/discover.c
+++ b/tests/repo/discover.c
@@ -33,7 +33,7 @@ static void ensure_repository_discover(const char *start_path,
 	git_buf_attach(&resolved, p_realpath(expected_path, NULL), 0);
 	cl_assert(resolved.size > 0);
 	cl_git_pass(git_path_to_dir(&resolved));
-	cl_git_pass(git_repository_discover(&found_path, start_path, 0, ceiling_dirs));
+	cl_git_pass(git_repository_discover(&found_path, start_path, 1, ceiling_dirs));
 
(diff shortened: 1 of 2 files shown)


pks-t:
This fixes #4186.

I took a look at also fixing `checkout::tree::extremeley_long_file_name` and `status::worktree::long_filenames` with schroot and aufs, but it requires quite a lot of trimming on the file names (some 20 to 30 bytes less). This is required as aufs supports a shorter maximum file length than most other filesystems. Considering it is possible to build with overlayfs with these fixes here I don't think we should compromise the general test case for a single rather obscure file system, though, so I refrained.

ethomson:
> I don't think we should compromise the general test case for a single rather obscure file system, though, so I refrained.

👍

cschultz702:
Thank you for this

PR #4233, opened 2017-05-05
Detached remotes

pks-t:
Took over from #3855. This includes the original contents from @heavenlyhash squashed together and rebased on master. While at it, I renamed `git_remote_create_unattached` to the more common `git_remote_create_detached`. I added checks to refuse push/download/upload for detached repositories, as it would segfault otherwise. I also added a few tests for `git_remote_ls` with detached remotes and tests checking whether we correctly refuse the other actions.


ethomson:
I updated the changelog to reflect the change in API name, and I manually merged this.

PR #4235, opened 2017-05-10
Out of tree builds

pks-t:
I recently started setting up a Docker-based build infrastructure to test several different Linux distributions (e.g. Debian, Ubuntu, Fedora, etc.). It just become too tedious to always set up a new VM for every issue specific to some kind of distribution. Not sure if it's valuable to upstream the Dockerfiles later on, but I guess it wouldn't hurt after all.

I currently mount the libgit2 sources as a read-only volume into the container. This left me unable to build our tree as our out-of-tree builds are broken. We write both the clar.suite and .clarcache files into the source tree rather than into the binary directory. This PR fixes these issues and brings us back to true out-of-tree builds.


tkelman:
Should one or more of the travis builds be done out of tree so this doesn't regress?

pks-t:
I thought about the same. One solution would be to simply do a "chmod -R a-w /path/to/sources" for all our build jobs and then building out-of-tree.

pks-t:
Something like this maybe. Untested as of now, will probably not build. Opinions?

pks-t:
So my approach to simply do a chmod does not work, as we're just copying all files into our temporary test directory without adjusting permissions. So I'll just leave our builds as-is for now.

pks-t:
Fixed missing sandbox cleanup. Should pass now on AppVeyor.

Issue #4234, opened 2017-05-09
Error reading index v4


jdavid:
### Reproduction steps

Create an index v4 with Git:

    $ git --version
    git version 2.12.2
    $ git init /tmp/test-v4
    $ cd /tmp/test-v4/
    $ git update-index --index-version=4
    $ touch file.txt
    $ git add file.txt 
    $ git commit -m "first commit"

Try to open the index with libgit2:

    #include <git2.h>
    #include <stdio.h>

    int main()
    {
      git_repository *repo;
      git_index *index;
      int err;

      err = git_repository_open(&repo, "/tmp/test-v4");
      if (err < 0) {
        printf("Error opening repo\n");
        return 1;
      }

      err = git_repository_index(&index, repo);
      if (err < 0) {
        printf("%s\n", giterr_last()->message);
        return 1;
      }

      return 0;
    }

### Expected behavior

Run without error.

### Actual behavior

Error reading the index file:

    $ ./a.out
    Invalid data in index - extension is truncated

### Version of libgit2 (release number or SHA1)

0.25.1

### Operating system(s) tested

Linux (Gentoo)

pks-t:
Thanks for the reproduction instructions, it's easily reproducible like this. Seems like we're reading compressed index path names in a wrong way, which causes this error. I'm investigating.

jdavid:
@pks-t it works now, thanks!

ethomson:
Fixed via #4236 - thanks @jdavid for reporting this and @pks-t for the fix!

PR #4236, opened 2017-05-10
Fix path computations for compressed index entries

pks-t:
This fixes the test case from #4234. We got the path calculation for compressed index entries completely wrong, where we ended up with a wrong path and with a wrong index entry size. This actually hints at a bigger problem: we obviously cannot have any tests at all regarding index v4, as that should've immediately shown that we do not really support this.

I'm off for today, but will amend tests later on.


pks-t:
Found and fixed more issues with computing index paths, as we did not update the index entry against which we were compressing paths against. Also added a test to verify we correctly read an index of version 4.

I somehow doubt that `index::version::can_write_v4` really checks what we think it does. It did read an index from disk and then used `git_index_write`, subsequently reading entries by path from the re-opened index. Seeing we weren't able to read these paths at all, I guess we simply wrote uncompressed paths into the index, which by definition is _not_ an index v4. So our write support has to be broken, as well.

pks-t:
Fixed a lot of issues with writing index v4 entries. I think there's still one issue left with us overallocating memory for compressed index entries when writing, causing us to write additional NULs into the index. While this works for us as we're able to read those entries, right-padding with NULs is not allowed in index v4. So there's a bug on both the reading and writing sides of our code.

pks-t:
I fixed two final errors.

First, the `git_varint_encode` function miscalculated required space, causing us to underallocate for big numbers and overallocate for small numbers.

Second, we were not NUL-terminating the path string for compressed index entries, it only worked by accident because the first field of the next entry frequently starts with '\x00'. I've added a few asserts to verify current behavior, and they're not being triggered anymore. I'm not aware of any additional bugs in compressed index entries currently.

pks-t:
The failing tests are due to my shitty test to verify we are actually writing index v4 with path compression. I simply check the size of the resulting written index, which seems to be non-portable. I've got no real other idea how to verify that we're actually writing an index v4 with path compression. though.

src/index.c
```diff
@@ -2344,35 +2357,34 @@ static size_t read_entry(
...
+		size_t suffix_len = strlen(path_ptr + varint_len);
+		size_t path_len;
 
 		if (varint_len == 0)
 			return index_error_invalid("incorrect prefix length");
 
-		GITERR_CHECK_ALLOC_ADD(&tmp_path_len, shared, len + 1);
-		tmp_path = git__malloc(tmp_path_len);
+		GITERR_CHECK_ALLOC_ADD(&path_len, prefix_len, suffix_len);
+		GITERR_CHECK_ALLOC_ADD(&path_len, prefix_len, 1);
```
ethomson: This needs to be `GITERR_CHECK_ALLOC_ADD(&path_len, path_len, 1);`.  As is this is truncating the size of `path_len` to `prefix_len + 1` instead of `prefix_len + suffix_len + 1`.

tests/index/version.c
```diff
@@ -3,39 +3,134 @@
...
+	entry.mode = GIT_FILEMODE_BLOB;
+
+	g_repo = cl_git_sandbox_init("indexv4");
+	cl_git_pass(git_repository_index(&index, g_repo));
+
+	/* write 676 paths of 250 bytes length */
+	for (i = 'a'; i <= 'z'; i++) {
+		for (j = 'a'; j < 'z'; j++) {
+			path[ARRAY_SIZE(path) - 3] = i;
+			path[ARRAY_SIZE(path) - 2] = j;
```
ethomson: `path` is not null-terminated here.  `path[ARRAY_SIZE(path) - 1] = '\0';`  ?

src/index.c
pks-t: Yeah, obviously. Thank?!

pks-t: Yeah, obviously. Thanks!

tests/index/version.c
pks-t: Also correct, will fix.

ethomson: 👍

PR #4238, opened 2017-05-11
Update README: VS -> VSTS

ethomson:
libgit2 is no longer used in Visual Studio (the IDE), it's used in
Visual Studio Team Services.


carlosmn:
Thanks for keeping it up to date.

PR #4240, opened 2017-05-15
Fix GCC warnings

pks-t:
This is split out from #4211. With this, we're free of compiler warnings on Ubuntu Trusty images, which will allow us to selectively enable "-Werror" in the future.

### DIFF
--- a/examples/network/common.c
+++ b/examples/network/common.c
@@ -1,5 +1,7 @@
 #include "common.h"
 #include <stdio.h>
+#include <string.h>
+#include <errno.h>
 
 /* Shamelessly borrowed from http://stackoverflow.com/questions/3417837/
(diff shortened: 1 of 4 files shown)


examples/network/common.c
```diff
@@ -20,15 +22,27 @@ int cred_acquire_cb(git_cred **out,
...
 		void * UNUSED(payload))
 {
-	char username[128] = {0};
-	char password[128] = {0};
+	char *username = NULL, *password = NULL;
+	int error;
 
 	printf("Username: ");
-	scanf("%s", username);
+	if (getline(&username, NULL, stdin) < 0) {
```
carlosmn: According to my manpage, "This buffer should be freed by the user program even if `getline()` failed", not that it's a big deal in this case.

PR #4244, opened 2017-05-24
Conditional includes

pks-t:
A first implementation of conditional includes. I'd mainly like to get some feedback on the implementation, as these conditionals require us to break backwards compatibility for configuration backends. This results from the requirement to have a repository struct available when parsing the config, as otherwise we wouldn't be able to resolve the conditions correctly. Obviously, parsing configurations still works when no repository is available -- in that case, we will simply treat all conditions as non-matching.

### DIFF
--- a/include/git2/config.h
+++ b/include/git2/config.h
@@ -199,6 +199,8 @@ GIT_EXTERN(int) git_config_new(git_config **out);
  * @param path path to the configuration file to add
  * @param level the priority level of the backend
  * @param force replace config file at the given priority level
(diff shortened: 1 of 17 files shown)


pks-t:
Will hold this PR until #4250 is merged.

pks-t:
Rebased on top of #4250 and fixed conflicts. I consider this PR ready now

pks-t:
Oh, I've accidentally closed this PR. Will create a new one.

Issue #4247, opened 2017-05-26
Cannot remove local branch with remote tracking branch if config is included


csware:
### Reproduction steps

1. Set an included config file, e.g. using `git config --local include.path ../something` (a valid config file named something must exist parallel to the .git dir)
2. Create a local branch with a remote tracked branch, e.g. using `git branch --track master2 remotes/origin/master`
3. Delete the local branch using libgit2 with `git_branch_delete` after looking it up using `git_reference_lookup`.

### Expected behavior

The branch is removed, the config file is adjusted and no error.

### Actual behavior

libgit2 fails with: `could not find key 'branch.master2.merge' to delete`

### Version of libgit2 (release number or SHA1)

87f5fbabdcd2318e136a142b9fcddf5614348546

### Operating system(s) tested

Win7 x64

Crossref: https://gitlab.com/tortoisegit/tortoisegit/issues/2976

pks-t:
Thanks for the awesome reproduction steps, I can reproduce the error like this. It is indeed related to the include, erasing it allows me to delete the branch again. I'll take a look at the code.

csware:
Fixed.

PR #4250, opened 2017-05-30
Configuration file fixes with includes

pks-t:
This fixes the issues causing #4247, but it does not yet fix additional issues with includes. One more issue I found was that upon refreshing a config file, we'll duplicate the includes inside the readers struct, if I'm not mistaken. I'll try to tackle this problem tomorrow, alongside with some tests for the broken behaviour.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -1507,10 +1507,10 @@ on_error:
 
 static int config_parse(
 	struct reader *reader,
-	int (*on_section)(struct reader **reader, const char *current_section, const char *line, size_t line_len, void *data),
-	int (*on_variable)(struct reader **reader, const char *current_section, char *var_name, char *var_value, const char *line, size_t line_len, void *data),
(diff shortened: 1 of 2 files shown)


pks-t:
As ever so often, this PR has grown quite a lot again as several other issues have popped up :P So I've started refactoring our configuration includes code to be easier to maintain and fix these bugs. I hope that these refactorings might ease further improvements alongside with include files,for example when modifying variables in the correct config file.

But yeah, I'm done with that code for now, not to keen on revisiting it too soon :slightly_smiling_face:

csware:
Any plans when this gets merged?

pks-t:
While I'd like to get it merged soon-ish, I won't merge without some thorough review as it touches quite some internals of how we handle config files. So doing reviews would definitly accelerate getting this PR ready.

ethomson:
🎉

PR #4251, opened 2017-06-05
fix build with libressl

Keruspe:


pks-t:
The change looks good to me, thank you. I'd welcome an improved commit message, though. E.g. something along the lines of:

```
openssl_stream: fix building with libressl

OpenSSL v1.1 has introduced a new way of initializing the library without having to call various functions of different subsystems. In libgit2, we have been adapting to that change with 88520151f (openssl_stream: use new initialization function on OpenSSL version >=1.1, 2017-04-07), where we added an #ifdef depending on the OpenSSL version. This change broke building with libressl, though, which has not changed its API in the same way.

Fix the issue by expanding the #ifdef condition to use the old way of initializing with libressl.
```

I didn't include any line breaks as it would be tedious to do so inside of the GitHub code editor :P

Keruspe:
Sure, editing the message

pks-t:
I've manually merged your commit via 40139fe6e62e487e6ccae26df6f9b1ddc9b30b36 to also include the 'openssl_stream:' prefix in the commit message. Thanks again for this fix!

PR #4252, opened 2017-06-06
Drop support for Windows Vista

csware:
Support for Windows XP was dropped in libgit2 shortly after Windows XP has reached its EOL. Vista reached its EOL in April this year (cf. https://support.microsoft.com/help/13853/windows-lifecycle-fact-sheet).

This PR removes some historic/unnecessary checks for `> XP` as well as `> Vista` (there is still more cleanup possible in order to drop support for unchunked data transfer in `src/transports/winhttp.c` ).

I also removed `git_has_win32_version` as this is not references any more. I put this removal into a separate commit so that it can be easily skipped if you want to keep this function.


jacquesg:
Is supporting Vista really such an issue at this stage? Is there a specific feature we really need in Windows 7 that is blocking the completion of a feature?

csware:
@jacquesg I opened this PR because support for XP was also dropped right after its EOL whereas XP still is widerly used than Vista.

jacquesg:
The point that I'm making is that dropping Vista for the sake of dropping it doesn't make sense.

pks-t:
I do see reasons to actually drop support for Windows Vista, the main reason being that some code related to supporting Windows Vista is not tested in any of our CI jobs.

That being said, we could decide to stick with Git for Windows' support cycle. Seeing they did also drop support for Windows XP with 2.10.0 and did not yet drop support for Windows Vista, I think we should at least wait until dscho decides to deprecate Windows Vista.

src/transports/winhttp.c
```diff
@@ -922,65 +921,21 @@ static int winhttp_stream_read(
 			s->sent_request = 1;
 		}
 
-		if (s->chunked) {
-			assert(s->verb == post_verb);
+		assert(s->verb == post_verb);
```
pks-t: This line is now causing an assert in our test suite.

csware: I dropped the last commit for now which does http transprot cleanups.

tests/core/posix.c
```diff
@@ -29,12 +29,7 @@ void test_core_posix__initialize(void)
 
 static bool supports_ipv6(void)
 {
-#ifdef GIT_WIN32
-	/* IPv6 is supported on Vista and newer */
-	return git_has_win32_version(6, 0, 0);
-#else
 	return 1;
```
ethomson: This function becomes rather pointless, don't you think?

csware: Oops ;), will fix this...

ethomson:
Yeah, like @jacquesg I'm wondering a bit about the motivation.  We dropped XP support because it was actively difficult to continue to build support for it into libgit2.  Microsoft EOL'ing XP provided us the rationale to be able to drop XP.  There's no such forcing function here, as far as I can tell.  (Though I concede that this code is not tested by CI.)

My inclination would be to drop support for Vista at the moment it becomes painful for us to continue it, not to try to drop it just to match Microsoft's dates.

But to be honest, I really don't care that much one way or another.  I feel like I'm just missing the big picture here.

Is this causing you problems in your app?

csware:
It's not causing issues for me. I just though you want to act consistently to the end of support for XP. When XP support was dropped in libgit2 it were only two lines of code which broke XP support (cf. https://gitlab.com/tortoisegit/tortoisegit/blob/e90660342aaf9e2a2116d58b07dc89ab1c1e3a91/ext/libgit2-0007-Unbreak-Windows-XP.patch).

ethomson:
Yeah but supporting XP had been a vicious pain in our ass.  See #1805 and all the crap we had to do to make it work (while XP was still an officially supported platform from Microsoft).

But I wasn't going to go and start mucking with our pthread emulation layer once we dropped support there.  The overriding goal of dropping XP support was to never be put into that position ever again.  We expected that there would be very similar compatibility problems coming our way in the future.

That turned out to not be the case (as you note).

So if you're saying that we should do the same thing with Vista, then yes, maybe that's true.  But we haven't been burned by Vista.  And it's not the dumpster fire that XP is, it's significantly more modern.

Anyway.  Like I said, I actually don't care that much either way.

Are you supporting XP in your application?  Are you basically re-patching libgit2 to support XP?  Is that what the diff you're linking to indicates?

csware:
TGit dropped support for XP by the beginnig of last year and is going to drop support for Vista with the next version.

ethomson:
Okay, thanks for the background.

I don't want to change support for the 0.26, which will come out very soon.  Will revisit as soon as its out the door and we have updated the bindings as well.

src/hash/hash_win32.c
```diff
@@ -23,10 +23,6 @@ GIT_INLINE(int) hash_cng_prov_init(void)
 	char dll_path[MAX_PATH];
 	DWORD dll_path_len, size_len;
 
-	/* Only use CNG on Windows 2008 / Vista SP1  or better (Windows 6.0 SP1) */
-	if (!git_has_win32_version(6, 0, 1))
-		return -1;
```
ethomson: If you're dropping this test, you can remove the entire indirection between CryptoNG and CryptoAPI and force CryptoNG always.

src/transports/winhttp.c
```diff
@@ -1467,7 +1467,7 @@ static int winhttp_receivepack(
 
 	/* WinHTTP only supports Transfer-Encoding: chunked
 	 * on Windows Vista (NT 6.0) and higher. */
-	s->chunked = git_has_win32_version(6, 0, 0);
+	s->chunked = 1;
 
 	if (s->chunked)
```
ethomson: This test is now always true.

pks-t:
So I feel like our current stance is to not drop Windows Vista support, as the maintenance work required for it is rather low. So I'll close this PR for now -- please feel free to reopen it in case you want to continue the discussion.

csware:
@pks-t I can't reopen a closed pull request...

pks-t:
@csware: this is usually possible, but in fact I cannot reopen this pull request either. Has the corresponding branch been deleted?

PR #4253, opened 2017-06-06
Coverity fixes

pks-t:
Once again, a set of fixes for issues pointed out by Coverity.

### DIFF
--- a/src/refs.c
+++ b/src/refs.c
@@ -622,15 +622,25 @@ typedef struct {
 static int update_wt_heads(git_repository *repo, const char *path, void *payload)
 {
 	rename_cb_data *data = (rename_cb_data *) payload;
-	git_reference *head;
+	git_reference *head = NULL;
 	char *gitdir = NULL;
-	int error = 0;
+	int error;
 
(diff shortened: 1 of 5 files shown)


src/refs.c
```diff
@@ -624,13 +624,23 @@ static int update_wt_heads(git_repository *repo, const char *path, void *payload
 	rename_cb_data *data = (rename_cb_data *) payload;
 	git_reference *head;
 	char *gitdir = NULL;
-	int error = 0;
+	int error;
+
+	if ((error = git_reference__read_head(&head, repo, path)) < 0) {
```
ethomson: If `git_reference__read_head` fails, `head` will never be initialized, and we'll call `git__free` on it.  (`read_head` also does not set it to `NULL`.)  Can we initialize it to `NULL` before this?

ethomson:
:+1: except for that issue in `update_wt_heads`

pks-t:
Thanks for the review, fixed the issue. It kinda sucks -- I've got my kernel compiled with some features causing memory to always be zeroed out. I think I should change that to instead have the kernel fill memory with random garbage to catch these issues earier.

PR #4254, opened 2017-06-07
CHANGELOG: add various changes introduced since v0.25

pks-t:


CHANGELOG.md
```diff
@@ -3,16 +3,83 @@ v0.25 + 1
...
+    * `git_worktree_lock()` and `git_worktree_unlock()` let you lock
+      respectively unlock a worktree.
+    * `git_repository_open_from_worktree()` lets you open a repository via
+    * `git_repository_head_for_worktree()` lets you get the current `HEAD` for a
+      linked worktree.
+    * `git_repository_head_detached_for_worktree()` lets you check whether a
+      linked worktree is in detached HEAD mode.
+
+* `git_repository_item_path()` lets you retrieve paths for various repository
+  files.
```
pks-t: Would `git_repository_file_path()` be a better name?

CHANGELOG.md
```diff
@@ -3,16 +3,83 @@ v0.25 + 1
...
+* `git_repository_submodule_cache_all()` and
+  `git_repository_submodule_cache_clear()` functions allow you to prime or clear
+  the submodule cache of a repository.
+
+* You can disable strict hash verifications via the
+  `GIT_OPT_ENABLE_STRICT_HASH_VERIFICATION` option with `git_libgit2_opts()`.
+
+* You can enable us calling `fsync` for various objects by setting the
+  `GIT_OPT_ENABLE_SYNCHRONIZED_OBJECT_CREATION` option with
+  `git_libgit2_opts()`.
```
pks-t: This option is not only about object creation, but it also modifies how we handle e.g. references. So maybe we should tweak this to `GIT_OPT_ENABLE_SYNCHRONIZED_FILE_CREATION`?

ethomson: Hmm.  I worry that `FILE_CREATION` makes it sound like we'll synchronize everything (including the working directory), which we don't...

`GIT_OPT_ENABLE_SYNCHRONIZED_REPO_ACCESS`...
`GIT_OPT_ENABLE_SYNCHRONIZED_REPOSITORY`...
`GIT_OPT_ENABLE_FSYNC_OBJECTS_AND_REFS`...
`GIT_OPT_ENABLE_FSYNC_REPOSITORY`...?

I dunno.

CHANGELOG.md
ethomson: I think that I prefer `GIT_REPOSITORY_ITEM_*` and `git_repository_item_path` over `file`...  they're both a bit vague, though.  But I can't think of anything superior offhand.

I'm happy with what we have, but if you think it's _too_ vague then we can spend some brain cycles on something superior...?

pks-t: Let's leave it as is, then.

CHANGELOG.md
pks-t: That's true. What about `GIT_OPT_ENABLE_FSYNC_GITDIR`? Granted, we do not currently synchronize e.g. the config files inside of the gitdir, but I think the term repo is not as unambiguous for non-git-devs as one might think.

ethomson: I'm 👍 on that.

ethomson: Do you want to send a PR to fix that up or shall I?

pks-t: I've got some spare time right now, will create a PR

pks-t:
Updated to match new option name `GIT_OPT_ENABLE_FSYNC_GITDIR` introduced with #4259.

PR #4255, opened 2017-06-07
Buffer growing cleanups

pks-t:
Small-ish cleanups for growing buffers. While this is mostly done to silence Coverity, I think it even improves code by a tiny fraction.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -724,9 +724,7 @@ int git_buf_join(
 	GITERR_CHECK_ALLOC_ADD(&alloc_len, strlen_a, strlen_b);
 	GITERR_CHECK_ALLOC_ADD(&alloc_len, alloc_len, need_sep);
 	GITERR_CHECK_ALLOC_ADD(&alloc_len, alloc_len, 1);
-	if (git_buf_grow(buf, alloc_len) < 0)
-		return -1;
-	assert(buf->ptr);
+	ENSURE_SIZE(buf, alloc_len);
 
 	/* fix up internal pointers */
 	if (offset_a >= 0)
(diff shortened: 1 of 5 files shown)


ethomson:
Seems reasonable to me.  One thing I noticed here is that the `ENSURE_SIZE` macro is pretty wacky...!

```
#define ENSURE_SIZE(b, d) \
    if ((d) > buf->asize && git_buf_grow(b, (d)) < 0)\
        return -1;
```

Notice the `buf` there instead of using the `b` argument...  turns out that we're real consistent with naming and didn't notice this!

ethomson:
I'm going to merge this manually, with a fix for `ENSURE_SIZE`.

pks-t:
Oh, good eye! That's just been waiting to explode. Seeing you didn't yet create the commit, I'll fix this up.

ethomson:
Oddly, fixing that makes a unit test fail on Windows...!  I'm digging in now.

ethomson:
But feel free to push a change to this branch - I'm curious if AppVeyor sees the same thing or if it's something bizarre on my machine.

pks-t:
Oh, strange. Didn't get any errors on Linux.

pks-t:
Taking a look at all calls to `ENSURE_SIZE`, it is always called with `buf`, so the generated code should always be the same.

ethomson:
Yeah...  presumably something else entirely, but the `repo::discover` tests are failing to clean up their tempdir on my machine now...  Very odd.

ethomson:
Failing for me in `origin/master` as well, so...  🤷‍♂️   No idea what's going on, but I'm going to assume it's isolated to my machine for now and we can go ahead and merge this as-is.

PR #4257, opened 2017-06-07
Execute stale tests

pks-t:
I've took a quick jab at our tests which are skipped by default and found some things which could be improved. The main outcome of this is that our CI jobs will now execute all invasive tests as well as our performance tests (or rather our performance test, singular). I think it is desirable to always execute as much of our tests as possible on CI jobs. This is especially true if we'd just save a few seconds otherwise, as is the case for our currently skipped tests.


pks-t:
I've included some more patches to clean up how our CI systems execute tests. Previously, it was a jungle of setting environment variables and subsequently running different tests in a fixed order, which was quite cumbersome to maintain across the two CI systems.

pks-t:
Tests fail due to the password variable not being set in AppVeyor. We should change the secure variables in AppVeyor to enable execution

pks-t:
Please hold until #4282 is merged due to rebasing being a major pain

pks-t:
Rebased due to #4282. This can be merged now.

appveyor.yml
```diff
@@ -45,16 +46,10 @@ test_script:
...
     $proxyJob = Start-Job { java -jar $Env:APPVEYOR_BUILD_FOLDER\build\poxyproxy.jar -d --port 8080 --credentials foo:bar }
-    ctest -V -R libgit2_clar
     $env:GITTEST_REMOTE_URL="https://github.com/libgit2/non-existent"
     $env:GITTEST_REMOTE_USER="libgit2test"
-    ctest -V -R libgit2_clar-cred_callback
-    Receive-Job -Job $proxyJob
-    $env:GITTEST_REMOTE_PROXY_URL = "http://foo:bar@localhost:8080"
-    ctest -V -R libgit2_clar-proxy_credentials_in_url
-    $env:GITTEST_REMOTE_PROXY_URL = "http://localhost:8080"
+    $env:GITTEST_REMOTE_PROXY_URL = "localhost:8080"
```
carlosmn: This was testing two different things: giving the proxy creds in the URL and reading them from the user. This eliminates one of those tests.

pks-t: It does not, as I've refactored the test to avoid exactly that. I have split up the test and we now have one where we put the credentials in the URL and one where we have the user provide it via the credentials.

pks-t:
I've stripped some parts of this PR to only include the most beneficial fixups.

ethomson:
> This is especially true if we'd just save a few seconds otherwise, as is the case for our currently skipped tests.

Hmm.  How many seconds is a few?  This adds up _a lot_ on the Windows CI, since this is creating a lot of files on Windows, which is terribly slow on NTFS.  Multiply that out by the number of variations we run for every PR times every PR, and ugh, this gets expensive fast.

I think we should take all the other changes but this one.

Another option is to make this an `int` where tests specify _how slow they are_.  Some are going to be so slow that they time out the CI system.  But I think that this is too complex and we should just say some tests are slow, and not run them as part of the CI.

pks-t:
I agree with you that we should strive to keep it simple, so I think a slowness-factor is kind of overengineering the issue. I'll drop that commit

ethomson:
Thanks as always.

PR #4258, opened 2017-06-07
SHA1DC update

pks-t:
Upstream has fixed issues with both endianess and unaligned access, which we obviously want to include, as well. Furthermore, they have provided a mechanism to use custom includes, which we now use to include our "common.h" instead of the system standard includes. This should also improve our cross-platform compatibility.


ethomson:
Thanks for staying on top of this.

PR #4260, opened 2017-06-08
Update to forced checkout and untracked files

ethomson:
A riff on #4129, with the changes discussed therein.  Opening a new PR to make sure that CI is happy before I merge it.  😄

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -370,10 +370,8 @@ static int checkout_action_wd_only(
 			 */
 			const git_index_entry *e = git_index_get_byindex(data->index, pos);
 
-			if (e != NULL && data->diff->pfxcomp(e->path, wd->path) == 0) {
-				notify = GIT_CHECKOUT_NOTIFY_DIRTY;
-				remove = ((data->strategy & GIT_CHECKOUT_FORCE) != 0);
-			}
+			if (e != NULL && data->diff->pfxcomp(e->path, wd->path) == 0)
+				return git_iterator_advance_into(wditem, workdir);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


ethomson:
Hrmf, looks like I did something stupid here.  👀

pks-t:
I cannot investigate right now, got no Windows VM handy -- I could do so next week, probably, if you've not gotten around to do it by then.

By the way -- feel free to drop the commit containing my previous solution.

ethomson:
Previous to this change, we would try to seize the opportunity to efficiently remove directories when we thought that we wanted to remove all of the files inside them.  But we could not effectively do that, and risked data loss.  This removes that optimization, and makes us process the items inside the directory.  By moving the directory deletion into standard call path, we clean up directories that are made newly empty (as a result of deleting the files inside them).

This change illustrated another failure:  the option `GIT_RMDIR_SKIP_NONEMPTY` allows one to skip `rmdir` failures when the directory is in-use (on Windows).  This option was tested in the optimized delete-the-whole-directory-at-once case but never tested the common remove-the-directory-after-all-its-files-are-gone case.  The latter _always_ enabled this setting, so it would not fail when it could not remove a directory due to `EBUSY`.  Moving the checkout logic to always use the latter showed this failure.

So the last commit teaches the `rmdir` case to fail when `EBUSY` and `GIT_RMDIR_SKIP_NONEMPTY` is omitted.  This brings us in line with the promises made by the documentation and should not be a regression.

Callers on Windows should probably be using `GIT_CHECKOUT_SKIP_LOCKED_DIRECTORIES`.

PR #4263, opened 2017-06-12
Allow creation of a configuration object in an in-memory repository

ethomson:
Regression:  we cannot create a configuration object against an in-memory remote, since the path to the repository configuration cannot be created.

Change `git_repository_item_path` to return `GIT_ENOTFOUND` when the path could not be constructed because the underlying directory does not exist.  (For example, when requesting the `config` file on an in-memory repository.

Add a unit test for a concrete scenario: create an anonymous remote on an in-memory repository (to allow for emulating `git ls-remote`).

### DIFF
--- a/tests/network/remote/local.c
+++ b/tests/network/remote/local.c
@@ -465,3 +465,19 @@ void test_network_remote_local__push_delete(void)
 	cl_fixture_cleanup("target.git");
 	cl_git_sandbox_cleanup();
 }
+
+void test_network_remote_local__anonymous_remote_inmemory_repo(void)
+{
+	git_repository *inmemory;
+	git_remote *remote;
+
+	git_buf_sets(&file_path_buf, cl_git_path_url(cl_fixture("testrepo.git")));
+
+	cl_git_pass(git_repository_new(&inmemory));
+	cl_git_pass(git_remote_create_anonymous(&remote, inmemory, git_buf_cstr(&file_path_buf)));
+	cl_git_pass(git_remote_connect(remote, GIT_DIRECTION_FETCH, NULL, NULL, NULL));
+	cl_assert(git_remote_connected(remote));
+	git_remote_disconnect(remote);
+	git_remote_free(remote);
+	git_repository_free(inmemory);
+}
--- a/include/git2/repository.h
+++ b/include/git2/repository.h
@@ -433,12 +433,12 @@ typedef enum {
  * item. It will thereby honor things like the repository's
  * common directory, gitdir, etc. In case a file path cannot
  * exist for a given item (e.g. the working directory of a bare
- * repository), an error is returned.
+ * repository), GIT_ENOTFOUND is returned.
  *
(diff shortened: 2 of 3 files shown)


ethomson:
Found this when trying to update LibGit2Sharp with the new 0.26 release candidate, so I would like to bring this in for 0.26.

tests/network/remote/local.c
```diff
@@ -465,3 +465,19 @@ void test_network_remote_local__push_delete(void)
 	cl_fixture_cleanup("target.git");
 	cl_git_sandbox_cleanup();
 }
+
+void test_network_remote_local__anonymous_remote_inmemory_repo(void)
+{
+    git_repository *inmemory;
+    git_remote *remote;
```
pks-t: Small nit: tabs vs spaces ;)

ethomson: Thanks!  My editor set up is wacky on this machine.

pks-t:
Sorry for having the change requests tickle in, but I just noticed that the tests are added before the issue is fixed. This would break bisectability, so I'd prefer if the test commit is moved last here.

ethomson:
I can change this.  I've done the test-first pattern for a while, as has nulltoken.  I actually don't know @carlosmn 's opinion here.  I know that he's opened some PRs that have failing tests, but I don't know if he ends up rewriting the history after he's made them 💚 .

I've always felt like keeping things able to be bisected for any particular test is important.  That is to say that we should certainly never commit a build break.  But introducing a failing unit test doesn't bother me.    My assumption here is that most people are running `git bisect` for a single test (or a few tests) not the whole test suite, and this would not get in their way.  But if this is something that you find value in doing then I can change my ways.

pks-t:
I certainly think it provides some value to have a "clean" history here. I'm actually kind of inspired by upstream git, as they also very carefully avoid having commits in between that break the test suite in any way. They do have the benefit of being able to tell a real story, though:

1. introduce a test with `test_expect_failure`, which actually fails in an expected way
2. fix the bug and switch over to `test_expect_success`

What I like about it is that this chain demonstrates both that there actually _was_ a failure before and that it is fixed by the respective commit, so that one can exactly pinpoint both points in time. But we're missing this feature in the clar, and my pull request has been rejected when I tried to introduce it.

But yeah. It's only been the first time that I noticed the inverted pattern here. So while I think it's a nice to have, I'll leave it to you to judge whatever suits us most.

ethomson:
What PR was that?  I must have forgotten about this.

pks-t:
Oh, it wasn't the lbigit2 repo but the clar repo. See https://github.com/vmg/clar/pull/77

ethomson:
Thanks for the background.  I hadn't seen that PR, or had forgotten about it completely.  I'll have a think on this.

I did reorder these commits here but I think we should have a larger thought about how we want to do this or if we care.  I feel like we might be getting past the point of diminishing returns, but that might just be me.

PR #4264, opened 2017-06-12
odb_read_prefix: reset error in backends loop

ethomson:
When looking for an object by prefix, we query all the backends so that we can ensure that there is no ambiguity.  We need to reset the `error` value between backends; otherwise the first backend may find an object by prefix, but subsequent backends may not.  If we do not reset the `error` value then it will remain at `GIT_ENOTFOUND` and `read_prefix_1` will fail, despite having actually found an object.

LibGit2Sharp creates such a scenario in its tests: it sets up its own custom ODB backend, then adds an object and looks that object up again by a prefix.


pks-t:
Looks good as is. I wrote a test which segfaults due to this issue in #4265.

PR #4265, opened 2017-06-12
Read prefix tests

pks-t:
Inspired by #4264. Along other tests, this also adds a test which segfaults due to the error fixed in the other PR.

### DIFF
--- a/tests/odb/backend/backend_helpers.c
+++ b/tests/odb/backend/backend_helpers.c
@@ -1,0 +1,119 @@
+#include "clar_libgit2.h"
+#include "git2/sys/odb_backend.h"
+#include "backend_helpers.h"
+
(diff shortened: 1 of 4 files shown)


pks-t:
So this PR fails in the expected way. Rebasing it on top of the merged #4264

ethomson:
Oh, huh.  I didn't realize that there was a possibility of a segfault there...  I just noticed that it bubbled up the error.

pks-t:
Ah, small correction. I think the segfault is introduced by the tests themselves. It was caused by `read_prefix_1` returning an error and freeing the value, but also setting the out pointer. So when freeing the expected `NULL` value, we actually run into a double free.

ethomson:
Ahhh - thank you, that does make sense.  :D

PR #4266, opened 2017-06-12
travis: install openssl explicitly

ethomson:
I can't entirely understand why this is failing, but it is.  Harumpfh.

### DIFF
--- a/script/install-deps-osx.sh
+++ b/script/install-deps-osx.sh
@@ -5,4 +5,5 @@ set -x
 brew update
 brew install homebrew/dupes/zlib
 brew install curl
+brew install openssl
 brew install libssh2


ethomson:
🤷‍♂️

pks-t:
What exactly does this change fix? I'm a bit confused, as it doesn't really change anything at all in our macOS builds:

```
+brew install openssl
Error: openssl 1.0.2j is already installed
To upgrade to 1.0.2l, run `brew upgrade openssl`
```

ethomson:
Huh.  Prior builds seemed to fail when trying to install libssl - it looked like a warning during the install of the prerequisite OpenSSL was being upgraded to a failure.

So I tried to install OpenSSL manually in the hopes that its warning would stay a warning.

So, I'm not sure.  Possible that my interpretation of the prior build failures is wrong and that the Travis Mac builds were flakey and I waited them out, or it's possible that there's some magic here.

ethomson:
Happy to dig in later when I have some free time.

PR #4267, opened 2017-06-12
adding GIT_FILTER_VERSION to GIT_FILTER_INIT as part of convention

mohseenrm:
Adding GIT_FILTER_VERSION to GIT_FILTER_INIT. 
Stumbled on this while using custom filters via. filter struct.
Also on the recommendation of @implausible.

Similar convention seen in files:
include/git2/sys/blame.h (Line 82)
include/git2/sys/checkout.h (Line 298)
include/git2/sys/cherrypick.h (Line 37)
...

### DIFF
--- a/include/git2/sys/filter.h
+++ b/include/git2/sys/filter.h
@@ -272,6 +272,17 @@ struct git_filter {
 
 #define GIT_FILTER_VERSION 1
 #define GIT_FILTER_INIT {GIT_FILTER_VERSION}
+
+/**
+ * Initializes a `git_filter` with default values. Equivalent to
+ * creating an instance with GIT_FILTER_INIT.
+ *
+ * @param filter the `git_filter` struct to initialize.
+ * @param version Version the struct; pass `GIT_FILTER_VERSION`
+ * @return Zero on success; -1 on failure.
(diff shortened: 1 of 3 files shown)


ethomson:
Yep, this probably makes sense and this changes looks good.  👍 

One thing we also do is provide a function that will `memcpy` the initialization data into a given `struct` - this is nice especially for languages that FFI w/o the C preprocessor.

Would you be so kind as to also add a `git_filter_init` that does this?  

See [`git_odb_init_backend`](https://github.com/libgit2/libgit2/blob/master/include/git2/sys/odb_backend.h#L115) for an example, and its [corresponding implementation](https://github.com/libgit2/libgit2/blob/master/src/odb.c#L1493).  And note the [unit tests](https://github.com/libgit2/libgit2/blob/89c332e41b12a72d89de40d63bc568c56a2c336a/tests/core/structinit.c) to verify them.

Thanks!

mohseenrm:
No problem @ethomson !

Just check if all the changes are consistent with what you had in mind 😊

ethomson:
Awesome, thanks @mohseenrm !

PR #4272, opened 2017-06-16
Patch ID calculation

pks-t:
This is a work in progress. It does produce the same patch IDs for many files, but I'm not quite there yet. I've also discovered an error in parsing patches where the file names have whitespaces inside and added a test, but there is no fix yet.

### DIFF
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -1400,6 +1400,51 @@ GIT_EXTERN(int) git_diff_format_email_init_options(
 	git_diff_format_email_options *opts,
 	unsigned int version);
 
+/**
+ * Patch ID options structure
+ *
+ * Initialize with `GIT_DIFF_PATCHID_OPTIONS_INIT` macro to
+ * correctly set the default values and version.
+ */
+typedef struct git_diff_patchid_options {
+	unsigned int version;
+} git_diff_patchid_options;
+
+#define GIT_DIFF_PATCHID_OPTIONS_VERSION 1
+#define GIT_DIFF_PATCHID_OPTIONS_INIT { GIT_DIFF_PATCHID_OPTIONS_VERSION }
+
+/**
+ * Initialize `git_diff_patchid_options` structure.
+ *
+ * Initializes the structure with default values. Equivalent to
+ * creating an instance with `GIT_DIFF_PATCHID_OPTIONS_INIT`.
+ */
+GIT_EXTERN(int) git_diff_patchid_init_options(
+	git_diff_patchid_options *opts,
+	unsigned int version);
+
+/**
+ * Calculate the patch ID for the given patch.
+ *
+ * Calculate a stable patch ID for the given patch by summing the
+ * hash of the file diffs, ignoring whitespace and line numbers.
+ * This can be used to derive whether two diffs are the same with
+ * a high probability.
+ *
+ * Currently, this function only calculates stable patch IDs, as
+ * defined in git-patch-id(1), and should in fact generate the
+ * same IDs as the upstream git project does.
+ *
+ * @param out Pointer where the calculated patch ID shoul be
+ *  stored
+ * @param diff The diff to calculate the ID for
+ * @param opts Options for how to calculate the patch ID. This is
+ *  intended for future changes, as currently no options are
+ *  available.
+ * @return 0 on success, an error code otherwise.
+ */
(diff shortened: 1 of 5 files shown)


pks-t:
Seems like we also don't parse patches with multiple diffs correctly. At least the following diff only invokes the file callback once:

```
diff --git a/x b/x
index 8a1218a..7059ba5 100644
--- a/x
+++ b/x
@@ -1,5 +1,4 @@
 1
 2
-3
 4
 5
diff --git a/y b/y
index e006065..9405325 100644
--- a/y
+++ b/y
@@ -1,4 +1,5 @@
 a
 b
+c
 d
 e
```

Will investigate later on. But other than multi-file patch IDs and patches of files with white space, this patch seems to correctly generate the stable IDs.

ethomson:
You need to flip that around.  In libgit2 terminology, a patch doesn't have multiple diffs, a diff has multiple patches.

A "patch" is a per-file change.  Multiple patches are aggregated into a diff.  I sort of loathe this terminology, but here we are.  (And please do correct me if I'm wrong, I always get confused by our naming here.)

How we choose to talk about the "patch id" for compatibility with git and our own internal nomenclature woes has yet to be seen.  We could call it a "diff id" instead.  Or not.

pks-t:
Ah, yeah. Have been bitten by that terminology multiple times now, thanks for reminding! Having switched over to the diff interface, all tests pass right now.

So I chose to name the function `git_diff_patchid`, which I think makes most sense. We do not want to introduce new terminology here with e.g. `git_diff_id`, but instead state our relation to the upstream git-patch-id(1).

While at it, I've now fixed parsing patches with unquoted file names containing whitespaces. I've also tried to tighten our diff parsing code to return an error whenever we parse a patch containing multiple files, but quickly noticed that we in fact rely on the ability to ignore trailing garbage in a patch. So this interface is potentially dangerous in that it allows a developer to parse a patch as diff where we simply truncate the diff to the first patch. While it can be argued that the developer is at fault for not reading documentation, I can sympathise due to our confusing terminology. But well, here we are. Meh...

ethomson:
> So this interface is potentially dangerous in that it allows a developer to parse a patch as diff where we simply truncate the diff to the first patch. 

IIRC, this is necessary, in fact, since they can (legally) have trailing data, IIRC.

(sigh.)

src/patch_parse.c
```diff
@@ -109,7 +109,10 @@ static int header_path_len(git_patch_parse_ctx *ctx)
 	size_t len;
 
 	for (len = quoted; len < ctx->line_len; len++) {
-		if (!quoted && git__isspace(ctx->line[len]))
+		if (!quoted &&
+		    (!git__strncmp(ctx->line + len, " b/", strlen(" b/")) ||
+		     !git__strncmp(ctx->line + len, " \"b/", strlen(" \"b/")) ||
```
ethomson: There's no guarantee that the prefixes are `a/` and `b/`.  You can set custom prefixes with `--src-prefix` and `--dst-prefix`, or you can have no prefix at all (with `--no-prefix`)...

This is a bit of a bummer as it means parsing the header line is completely ambiguous.  What does:

```
diff --git a/b c/d/e f/g h/i j/k
```

mean?

I think we'll need to store the whole header buffer then parse it with the knowledge given to us by the `---` and `+++` lines.

:(

ethomson: Except that a patch needn't _have_ `+++` and `---` lines.

```
diff --git foo bar foo bar
old mode 100755
new mode 100644
```

(Note: git cannot apply this patch.)

pks-t: God, that's awful. So at some point, I guess we'd just have to say "This is unsupported." I think it makes most sense to use the "+++" and "---" lines and bail out if they're missing, given that upstream git itself cannot use these kind of patches.

ethomson: Truly hideous, right?  Why these aren't just quoted whenever there's a space in the filename, I don't know.  😢 

Can we split the space in the filename commit out?  I _think_ this removes our ability to deal with custom prefixes.  I doubt anybody is actually using that but it might be thinking about this a bit more before we pull it out.  I think the patchid stuff is 👍 and ready to go.

pks-t: Yeah, wanted to propose the same. Will create a new PR for the other "fix" and keep it on hold for now

ethomson:
🎉

PR #4276, opened 2017-06-21
CMake: INCLUDE_DIRECTORIES -> TARGET_INCLUDE_DIRECTORIES

AndreyG:
It's preferable to use `target_include_directories` instead of `include_directories` in modern CMake. It's also more consistent with `target_link_libraries`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,7 +206,6 @@ FILE(STRINGS "include/git2/version.h" GIT2_HEADER_SOVERSION REGEX "^#define LIBG
 STRING(REGEX REPLACE "^.*LIBGIT2_SOVERSION ([0-9]+)$" "\\1" LIBGIT2_SOVERSION "${GIT2_HEADER_SOVERSION}")
 
 # Find required dependencies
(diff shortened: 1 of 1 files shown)


pks-t:
As of now, we cannot use the `target_include_directories` as it has only been introduced in CMake 2.8.12. While most current distros already have this version (Ubuntu Trusty, CentOS 6), there is at least the usual suspect of Debian, where Debian Wheezy only has v2.8.9. Furthermore, Travis currently still uses Ubuntu Precise, which has an older version, as well.

Anyway, thanks for your proposal!

AndreyG:
Can I use `target_include_directories` under `if(${CMAKE_VERSION} VERSION_GREATER 2.8.12)`? I really need it because I use libgit2 as subproject and this is rather awkward to add explicitly `libgit2/include` to include directories of my target.

pks-t:
Oh, yeah, that'd certainly be possible, sorry for not thinking of this possibility. I can reopen the issue if you decide to put up some code for this.

AndreyG:
I have created another pull request: https://github.com/libgit2/libgit2/pull/4284

PR #4277, opened 2017-06-21
merge: fix potential free of uninitialized memory

pks-t:
The function `merge_diff_mark_similarity_exact` may error our early and,
when it does so, free the `ours_deletes_by_oid` and
`theirs_deletes_by_oid` variables. While the first one can never be
uninitialized due to the first call actually assigning to it, the second
variable can be freed without being initialized.

Fix the issue by initializing both variables to `NULL`.


pks-t:
The build failures are unrelated

ethomson:
Yep, good catch.  Thanks!

PR #4279, opened 2017-06-23
-Werror builds for Travis

pks-t:
As promised in #4211, this PR enables building with "-Werror". I've taken a different approach here than last time by creating new macros `ENABLE_WARNINGS`/`DISABLE_WARNINGS`. When passing "-DENABLE_WERROR" to CMake, the `ENABLE_WARNINGS` macro is switched over to add "-Werror=<warning>" instead of "-W<warning" to our CFLAGS variable.

I've switched over Travis to use this new switch by default. I'll probably have to disable this for the old Precise machines as they still have the issue with `curl_easy_opt` warnings, which are out of our scope to fix.


pks-t:
Please hold until #4282 is merged due to rebasing being a major pain

pks-t:
Rebased for #4282. If CI has no complaints I'd consider this ready to be merged now.

PR #4280, opened 2017-06-23
Convert port with htons() in p_getaddrinfo()

ids1024:
`sin_port` should be in network byte order.

### DIFF
--- a/src/posix.c
+++ b/src/posix.c
@@ -40,7 +40,7 @@ int p_getaddrinfo(
 	if (ainfo->ai_servent)
 		ainfo->ai_port = ainfo->ai_servent->s_port;
 	else
-		ainfo->ai_port = atol(port);
+		ainfo->ai_port = htons(atol(port));
 
 	memcpy(&ainfo->ai_addr_in.sin_addr,
 			ainfo->ai_hostent->h_addr_list[0],


pks-t:
Thanks for this fix, looks good to me! @ethomson?

Just out of curiosity: on which platform did you experience problems with this?

ids1024:
Porting libgit2 to Redox OS (https://redox-os.org/), which at the moment has a very basic Berkeley sockets implementation.

So Redox's implementation could be incorrect, but the Posix standard states clearly that `sin_port` should be in network byte order, and of course `atol` returns in host byte order. It seems odd to me that this would be wrong though, since this presumably would mean that libgit2 is broken on little endian systems without `getaddrinfo`.

ethomson:
Agree that this should be in network byte order.  @ids1024 - I'm not surprised it was broken, this was added explicitly for AmigaOS and I'm pretty sure only ever used by them until now.

ids1024:
Ah, AmigaOS. I thought it was probably something like that, but I wasn't sure what big endian platform the code would have been used with.

pks-t:
So I'm merging this as we all seem to agree this is correct.

Thanks for your fixup! :fireworks:

PR #4282, opened 2017-06-23
Split up CMakeLists.txt build instructions

pks-t:
While discussing build issues with @ethomson, I've been reminded to how unwieldy our current CMakeLists.txt is. Besides being huge, it also has several problems with the build enviroment leaking between targets. One example are defines for our test suite, which are also set when we build our library code, which is unfortunate.

To fix this, I've started taking a look at splitting up the CMakeLists.txt. It's been quite annoying to do and probably has issues left, but I think it's a good think to do in the long run. Furthermore, I hope it will help the issue with precompiled headers from #3950. While at it, this PR also fixes building our library twice by using an object library, as in #3950.


ethomson:
It looks like you've dropped the `IDE_SPLIT_SOURCES` function.  Please don't.  This is the difference between Xcode / Visual Studio being usable on our project and being a disaster.

pks-t:
I know that's currently the state. But in the end, I think it's not required anymore due to us using subdirectories. This will automatically achieve what we've been doing with `IDE_SPLIT_SOURCES`. This'll require me to move the deps inside a subdirectory, as well, though, which I'll be doing next.

ethomson:
Interesting.  It would be nice to drop that hack; I'll take it for a spin.

pks-t:
Note though that this isn't ready yet on Windows, it has some problems with the precompiled headers there. I'm currently investigating.

pks-t:
By the way, I guess was wrong regarding IDE_SPLIT_SOURCES. While it's somewhat better now with different subdirectories it is still far from workable. So I'll re-include our hack.

pks-t:
Finally got it building now. Will have to fix up some locations such that e.g. the library and libgit2_clar are generated in the top-level build directory again. Note that this builds on #4288

tiennou:
I gave it a spin, and it's broken when using the Xcode generator (which is what I'm generally using).

- I've pushed a [fixup](https://github.com/tiennou/libgit2/tree/pks/remove-unused-clar-fixtures) for another issue I encountered.
- `libgit2_clar` [fails to link](https://gist.github.com/tiennou/42a94d0fcb3b286b893104d3408e7daa) because libssh2. I've tried to add the "missing" `link_directories(libgit2_clar, ...)`, but ultimately CMake doesn't want to add the missing -L flag in here. The `git2` target build fine though so 🥂 for that. Strangely, the normal Makefile generator works, but clar segfaults somewhere.
- On the subject of `IDE_SPLIT_SOURCES`, I don't think you've added it back (or at least it doesn't work anymore) ? But 👍 for keeping it.

pks-t:
Thanks for testing! The linking issue is quite easy to fix, I'll just have to add some missing libraries to the LIBGIT2_LIBDIRS variable.

I'm a bit surprised the `#if defined(GIT_USE_ICONV)` is required, I always had the understanding that just defining it implicitly set it to "1". I guess an easier fix would be to "cmakedefine" all variables to "1" to avoid that there are other places where we are missing this.

And yeah, I didn't yet re-add the `IDE_SPLIT_SOURCES` function, I've still been busy to get it working on all platforms. It's quite intricate to split up our build instructions and there were quite a few things I had to pay attention to.

pks-t:
This PR passes CI now. I've also manually tested on Linux, macOS and Windows, all seem to be working as expected. So I think this PR is ready now, except that I'll still have to tweak the commit messages a bit.

Benefits this introduces:

- more modern CMake style due to split build instructions for configuration (top level CMakeLists.txt), sources, dependencies (with one CMakeLists.txt for each of winhttp, zlib, regex and http-parser) and tests
- due to the splitup, we refrain from leaking build infromation between the different stages. E.g. defines, includes and other stuff set for either libgit2.so or libgit2_clar do not leak to the respective other target anymore
- introduction of a git2internal object target on supporting CMake versions (that is, 2.8.7 and higher) to reduce build times. I've measured the benefits on Linux to be around 10% faster build times, on Windows this will be higher
- introduction of a separate precompiled header for libgit2_clar
- feature-related flags are moved to a header file "features.h". While this header is currently private only, we might even consider to install it alongside with our other official headers. This would enable users of our API to conditionally compile code via ifdef's, whether that's a good thing or not.

I've split some preparatory stuff out of this PR in #4288 as it required a whole-tree conversion regarding headers to be prepared for the "features.h" header. You may review it separatly or inside of this PR, as to your liking.

@ethomson @carlosmn Would be nice to get some input, as I'd really like to get this merged early on to catch breakage early in the release cycle and to avoid clashing with all the other topics touching our build instructions.

pks-t:
Also @tiennou, would be nice if you could have another go with this PR.

tiennou:
Sadly, the link failure still happens. I've updated the gist above but it's the exact same.

About the "features.h" header, ISTR that we dislike compile-time checks (ie. the public API must not change as a side-effect of #defines, only behavior). So I don't think having it public is needed.

src/CMakeLists.txt
```diff
@@ -0,0 +1,396 @@
...
+ENDIF()
+
+CONFIGURE_FILE(features.h.in git2/sys/features.h)
+
+SET(GIT2INTERNAL_OBJECTS ${SRC_H} ${SRC_GIT2} ${SRC_OS} ${SRC_SSH} ${SRC_SHA1})
+
+LIST(APPEND LIBGIT2_INCLUDES ${CMAKE_CURRENT_BINARY_DIR} "${CMAKE_SOURCE_DIR}/src" "${CMAKE_SOURCE_DIR}/include")
+
+IF (CMAKE_VERSION VERSION_GREATER 2.8.7)
+	ADD_LIBRARY(git2internal OBJECT ${GIT2INTERNAL_OBJECTS})
```
tiennou: Trial and error™ says you might want to pass "${CMAKE_SOURCE_DIR}/include" here as well so that `IDE_SPLIT_SOURCES` adds the public headers. I have absolutely no idea if this has any other side-effects, but it seems that `SOURCE_GROUP` ignores files if they're not in a target's source list.

src/CMakeLists.txt
```diff
@@ -0,0 +1,396 @@
...
+ENDIF()
+
+# Optional external dependency: libssh2
+IF (USE_SSH)
+	PKG_CHECK_MODULES(LIBSSH2 libssh2)
+ENDIF()
+IF (LIBSSH2_FOUND)
+	SET(GIT_SSH 1)
+	LIST(APPEND LIBGIT2_INCLUDES ${LIBSSH2_INCLUDE_DIRS})
+	LIST(APPEND LIBGIT2_LIBS ${LIBSSH2_LIBRARIES})
```
tiennou: Found the trick : this needs to be changed to `LIST(APPEND LIBGIT2_LIBS "-L${LIBSSH2_LIBRARY_DIRS}" ${LIBSSH2_LIBRARIES})`, and makes the linker line correct. Kudos to `FindIconv.cmake` for the work-around.

pks-t: I don't get it. The next line here we do `LIST(APPEND LIBGIT2_LIBDIRS  {LIBSSH2_LIBRARY_DIRS})`, followed by a `LINK_DIRECTORIES(${LIBGIT2_LIBDIRS})` later on. So your line really shouldn't make any difference at all, as we are already including the link directory. The only thing I can imagine is that our `LINK_DIRECTORIES` instruction may come too late.

Are you using libssh2 from homebrew? I think I didn't have that one installed, will have a go at it.

pks-t:
Yeah, I tend to agree that we should keep the "features.h" header private for now.

src/CMakeLists.txt
pks-t: Fixed the issue. The added `LINK_DIRECTORIES` instructions were called too late indeed, particularly after the call to `ADD_EXECUTABLE(libgit2_clar)`. Interesting this only triggered on Xcode.

src/CMakeLists.txt
pks-t: The issue was not in us missing the includes, we already pass them via "${SRC_H}" inside of "${GIT2INTERNAL_OBJECTS}". But in fact, I forgot to adjust how "${SRC_H}" is constructed and so it was empty.

pks-t:
This now also includes the proper fix for #4291 in 2317fd8.

tiennou:
Xcode reports success !

pks-t:
Nice, thanks for testing!

tiennou:
Can I help in any way in moving that forward ?

pks-t:
@tiennou As always, reviews would be welcome. I'll defintly wait until either @ethomson or @carlosmn have a look at this, but they are unfortunately quite swamped with work.

pks-t:
Rebased due to the merge of #4288. Thanks @ethomson! :+1:

ethomson:
Thanks @pks-t .  So I'm definitely happy with the speed improvements - this shaves about 25% of the original build time off on Windows.  That's an insane improvement!

One thing I _don't_ love though, is where the build results end up.  On Windows, anyway, the test runner ends up in `Debug\libgit2_clar.exe`.  (Substitute `Release` for a release build.)  That's fine, that's exactly where it was before.

But `git2.dll` and friends used to land in the `Debug` directory as well.  Now there's a whole `src` directory in the build output...  And the library itself lands in `src\Debug\git2.dll`.

This seems a bit odd to me, tbqh.  Do you happen to know why this happens?  Is there a way to make the library land in `Debug` again?  Or maybe somewhere a bit more discoverable?

pks-t:
Thanks for giving this a whip, @ethomson.

The changing output location of the library is definitly a bug and can easily be changed. As to why: when you add a subdirectory, all binaries will by default end up in the same subdirectory inside of the build directory. For the clar I already changed the target output directory, but for the library I obviously forgot to do so.

pks-t:
So I stand corrected. We _are_ already setting the LIBRARY_OUTPUT_DIRECTORY to the project's root and on Linux the library ends up inside the correct directory. Will debug.

pks-t:
@ethomson Should be fixed now. Details are in the commit, TLDR: DLLs and import libraries are not treated as libraries by CMake but instead as runtime and archive targets, respectively.

ethomson:
Thanks for this @pks-t !

pks-t:
Yay! :D Thanks @ethomson. I've rebased all my other PRs which also touch the CMakeLists.txt in any way.

ethomson:
So now that #4346 has landed, does this mitigate the concerns of people including libgit2 w/o cmake?  (@jacquesg, @csware?)

csware:
@ethomson Right now it looks fine to me.

Issue #3200, opened 2015-06-10
Have CMake show a summary of the build options/decisions


carlosmn:
It'd be cool to have CMake show a list of the options and decisions it took and which will affect the build. Stuff like which crypto library we're using, what we're using for talking over the network, what we use for HTTP, zlib and regex.

iankronquist:
I'm not terribly familiar with CMake but I may be able to help. Do you have an example of this in other projects? Do you want this to happen for all build targets or just some specific ones?

carlosmn:
I don't know off-hand any projects which do this in CMake, but e.g. mono or smuxi provide a summary at the end. I was thinking that it would look something like

```
Crypto backend: SecureTransport
cURL support: yes
HTTP stack: http-parser
```

It wouldn't be tied to any target, but to the configuring itself. Towards the end of the config script, there'd be a bunch of code to figure out which one is used and what text to write out.

iankronquist:
Is this the sort of thing you would like to see? I threw this together in 5 minutes and want to make sure I'm on the right track.
https://github.com/libgit2/libgit2/compare/master...iankronquist:cmake-lib-summary/3200

```
Ian@kartal:(build)(master) → $ cmake ..
-- Found Security /System/Library/Frameworks/Security.framework
-- Found CoreFoundation /System/Library/Frameworks/CoreFoundation.framework
-- Could NOT find HTTP_Parser (missing:  HTTP_PARSER_INCLUDE_DIR HTTP_PARSER_LIBRARY) 
-- http-parser was not found or is too old; using bundled 3rd-party sources.
-- checking for module 'libssh2'
--   package 'libssh2' not found
-- LIBSSH2 not found. Set CMAKE_PREFIX_PATH if it is installed outside of the default search path.
-- Found Iconv: -L/usr/lib -liconv
-- Performing Test IS_FVISIBILITY=HIDDEN_SUPPORTED
-- Performing Test IS_FVISIBILITY=HIDDEN_SUPPORTED - Success
-- Performing Test IS_WSTRICT-ALIASING=2_SUPPORTED
-- Performing Test IS_WSTRICT-ALIASING=2_SUPPORTED - Success
-- SSL: 
-- SSH: 
-- GSSAPI: 
-- ICONV: -L/usr/lib -liconv
-- HTTP parser: 
-- Configuring done
...
```

carlosmn:
[...]

PR #4283, opened 2017-06-23
CMake: make HTTPS support more generic

tiennou:
Extracted from #4173 and rebased on master, switched to the new HTTPS nomenclature while keeping most of the functional changes.


tiennou:
@pks-t Here it is ! I'll rebase current #4173 on top of that one while I'm at it.

pks-t:
Thanks a lot! Will review in the near future.

tests/main.c
```diff
@@ -11,7 +11,11 @@ int main(int argc, char *argv[])
 
 	clar_test_init(argc, argv);
 
-	git_libgit2_init();
+	res = git_libgit2_init();
+	if (res < 0) {
+		return res;
+	}
+
```
pks-t: An error message would be nice here to aid the user.

src/global.c
```diff
@@ -10,7 +10,7 @@
 #include "sysdir.h"
 #include "filter.h"
 #include "merge_driver.h"
-#include "openssl_stream.h"
+#include "streams/openssl.h"
```
pks-t: Just putting it at an arbitratry position: you should probably adjust the include guards of renamed header files, even though it's mostly a cosmetic change

src/streams/openssl.c
```diff
@@ -627,6 +627,16 @@ int git_openssl_stream_new(git_stream **out, const char *host, const char *port)
 	return error;
 }
 
+int git_openssl_set_cert_file(const char *file, const char *path)
+{
+	if (SSL_CTX_load_verify_locations(git__ssl_ctx, file, path) == 0) {
+		giterr_set(GITERR_SSL, "SSL error: %s",
```
pks-t: Maybe be a bit more specific here? E.g. "could not set default locations for certificates"?

src/streams/openssl.c
```diff
@@ -627,6 +627,16 @@ int git_openssl_stream_new(git_stream **out, const char *host, const char *port)
 	return error;
 }
 
+int git_openssl_set_cert_file(const char *file, const char *path)
```
pks-t: We usually use "__" to denote private API. In this case "git_openssl__set_cert_file".

src/streams/openssl.h
```diff
@@ -13,6 +13,8 @@ extern int git_openssl_stream_global_init(void);
 
 extern int git_openssl_stream_new(git_stream **out, const char *host, const char *port);
 
+extern int git_openssl_set_cert_file(const char *file, const char *path);
```
pks-t: Some documentation would be nice here, even if it's private API. I certainly would be confused what "file" and "path" actually mean in this context

src/streams/openssl.c
```diff
@@ -653,4 +663,13 @@ int git_openssl_stream_new(git_stream **out, const char *host, const char *port)
 	return -1;
 }
 
+int git_openssl_set_ca_location(const char *file, const char *path)
```
pks-t: Don't you mean `git_openssl_set_cert_file` here?

CMakeLists.txt
```diff
@@ -293,6 +275,58 @@ ELSE ()
 	ENDIF()
 ENDIF()
 
+IF (USE_HTTPS STREQUAL "ON")
+	IF (SECURITY_FOUND)
```
pks-t: It's a bit strange that we're actually searching for the security framewok even if `USE_HTTPS=OFF` here. But we can improve that later, no need to do it as part of this PR

CMakeLists.txt
```diff
@@ -293,6 +275,58 @@ ELSE ()
...
+ELSEIF(HTTPS_BACKEND STREQUAL "WinHTTP")
+ENDIF()
+
+IF (USE_HTTPS AND NOT HTTPS_BACKEND)
+	MESSAGE(FATAL_ERROR "Asked for backend " ${HTTPS_BACKEND} " but it wasn't found")
+ENDIF()
+
+IF (HTTPS_BACKEND)
+	ADD_DEFINITIONS(-DGIT_HTTPS)
+ENDIF()
```
pks-t: Seeing all this code we could even think about making this a separate CMake module later on. Not required for this PR, but I'll tackle this at some point, I guess.

CMakeLists.txt
```diff
@@ -303,9 +337,9 @@ IF (USE_SHA1DC)
 ELSEIF (WIN32 AND NOT MINGW)
 	ADD_DEFINITIONS(-DGIT_SHA1_WIN32)
 	FILE(GLOB SRC_SHA1 src/hash/hash_win32.c)
-ELSEIF (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+ELSEIF (TLS_BACKEND MATCHES "SecureTransport")
```
pks-t: This should be HTTPS_BACKEND now, shouldn't it? I guess this stems from the fact that I've been using HTTPS with the `GIT_HTTPS` define. But actually, GIT_TLS and TLS_BACKEND are nicer names here. So we should certainly be sticking with one name -- I don't care if you want to rename existing usage of GIT_HTTPS to GIT_TLS or if you'll just stick with GIT_HTTPS for now.

CMakeLists.txt
```diff
@@ -303,9 +337,9 @@ IF (USE_SHA1DC)
...
 	ADD_DEFINITIONS(-DGIT_SHA1_COMMON_CRYPTO)
-ELSEIF (OPENSSL_FOUND)
+ELSEIF (TLS_BACKEND MATCHES "OpenSSL")
```
pks-t: Same as above.

pks-t:
Ah, one additional comment: there are some commit messages without a scope (e.g. "cmake:") in their header. Could you maybe add one where it's missing? :)

tiennou:
@pks-t Since you've proposed #4282, I'm somewhat wary of touching CMakeLists.txt, so if you don't mind I'll wait until it's merged. I'll go test #4282 though, if that helps.

tests/main.c
tiennou: Fixes

src/global.c
tiennou: Fixed

src/streams/openssl.c
tiennou: Fixed

src/streams/openssl.c
tiennou: Yep, fixed. I also tweaked some error codes where those functions are called.

src/streams/openssl.c
tiennou: Yep, copy-paste failure. Settled on `git_openssl__set_cert_location`.

CMakeLists.txt
tiennou: "Reverted" patchset to `GIT_TLS` then.

CMakeLists.txt
tiennou: Fixed

CMakeLists.txt
tiennou: Yeah, that's what I was thinking too, now that the backend selection is in one place 😉.

CMakeLists.txt
tiennou: Fixed, `USE_TLS=OFF` should disable all crypto-checking we do.

pks-t:
Well, I certainly don't mind at all. It's a bit unfortunate that #4282 is so invasive and clashes with all other PRs touching CMakeLists.txt.

We could move forward your PR and I'll try to fix merge conflicts in #4282. By now I've solved enough merge conflicts in there to know the exact steps required to reconstruct the final move, and the other commits are quite localized and as such easy to rebase. Just state your preference and I'll try to adapt accordingly.

If you want to move this one forward I'll try to focus on getting it ready and merged such that I can include these changes as soon as possible.

tiennou:
As it stands, I feel this one is ready. I rebased it just yesterday.

CMakeLists.txt
```diff
@@ -544,18 +544,20 @@ ELSE()
 ENDIF()
 
 IF (SECURITY_FOUND)
-  ADD_DEFINITIONS(-DGIT_SECURE_TRANSPORT)
```
pks-t: This commit does a bit more than its title states, as it also removes the definitions for platform-specific TLS specs. There are additional uses of these defines which you haven't covered in this commit, though:

- `GIT_SECURE_TRANSPORT` is used in stransport_stream.c and tls_stream.c
- `GIT_OPENSSL` is used in global.h, netops.h, openssl_stream.{c,h}

CMakeLists.txt
```diff
@@ -52,7 +53,9 @@ ENDIF()
 
 IF(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 	SET( USE_ICONV ON )
-	FIND_PACKAGE(Security)
+	IF(USE_TLS STREQUAL "OpenSSL" OR USE_TLS STREQUAL "ON")
+		FIND_PACKAGE(Security)
```
pks-t: Shouldn't this be `IF(USE_TLS STREQUAL "SecureTransport" OR USE_TLS STREQUAL "ON")`?

CMakeLists.txt
```diff
@@ -303,9 +339,9 @@ IF (USE_SHA1DC)
 ELSEIF (WIN32 AND NOT MINGW)
 	ADD_DEFINITIONS(-DGIT_SHA1_WIN32)
 	FILE(GLOB SRC_SHA1 src/hash/hash_win32.c)
-ELSEIF (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+ELSEIF (TLS_BACKEND MATCHES "SecureTransport")
```
pks-t: This looks like it should use `${TLS_BACKEND}`, shouldn't it? Shouldn't this use `STREQUAL`, as all previous code? Same below for OpenSSL.

CMakeLists.txt
tiennou: This is just an indentation fixup. The defines are still here (line +2).

CMakeLists.txt
tiennou: Will fix. I'll also check how this reacts to the CommonCrypto hash backend, because I think that checking for Security encompasses CommonCrypto & SecureTransport, so the hash backend might be broken by this change.

CMakeLists.txt
pks-t: Ah, sorry. I've been confused by the missing `-DGIT_HTTPS`.

CMakeLists.txt
tiennou: IIRC CMake original behavior was to always substitute variables in some cases, `IF` being one of them.

I'll switch to `STREQUAL` though, as `TLS_BACKEND` is supposed to be an internal option.

tiennou: Also, after taking a closer look, this line makes the hash-selection dependent on the backend.  Old behavior was "always use CommonCrypto" on macOS, now it's "choose whatever the selected TLS backend provides, except when using SHA1DC". Is that okay ?

pks-t: Huh, that's right. I think it would make more sense to stick with the DARWIN check here instead. Otherwise, it's becoming hard (impossible?) to use the CommonCrypto SHA1 library when TLS is turned off.

tiennou:
I discovered some issues with it (on top of the `CommonCrypto` thing), so I vote for deferring this one *after* #4282.

CMakeLists.txt
```diff
@@ -293,6 +277,58 @@ ELSE ()
 	ENDIF()
 ENDIF()
 
+IF(USE_TLS STREQUAL "ON")
+	IF (SECURITY_FOUND)
+		# OS X 10.7 and older do not have some functions we use, fall back to OpenSSL there
+		CHECK_LIBRARY_EXISTS("${SECURITY_DIRS}" SSLCreateContext "Security/SecureTransport.h" HAVE_NEWER_SECURITY)
```
tiennou: Aforementioned issue. Forcing the `SecureTransport` backend will skip this test.

pks-t:
Okay, as you wish. Only makes my own life easier :p

Thanks for keeping up the work on this!

tiennou:
Rebased.

I added some commits before tackling the `USE_TLS` option so the backend selection wouldn't be so messy.

- cc50581 moves the "Security too old" check in our module.
- 33aa5c4 adds `FeatureSummary` support (fixes #3200, because I was using that to check what CMake was doing)
- note that we now unconditionally check for *everything* (eg. OpenSSL/Security/whatever), so the backend selection code can decide what's best.

Something I'm still unsold on is the ability to prefer OpenSSL on macOS ; since we are always choosing CommonCrypto for SHA here and nobody has ever complained about not being able to not link against it, I'll keep it that way (but the fix is easy enough if needed).

pks-t:
#4282 has landed now, so we can actually continue working on this PR here.

tiennou:
Rebased and ready to go ! I've reverted the HTTPS->TLS change (again), because I think it predates 0.26 and I didn't want to break things.

src/tls_stream.c
```diff
@@ -9,8 +9,8 @@
```
pks-t: Is the "tls_stream" file special in any way? I'd have expected it to be moved to the "streams" directory, as well.

cmake/Modules/FindCoreFoundation.cmake
```diff
@@ -1,9 +1,30 @@
...
-    SET(COREFOUNDATION_FOUND TRUE)
-  ENDIF ()
+	FIND_PATH(COREFOUNDATION_INCLUDE_DIR NAMES CoreFoundation.h)
+	FIND_LIBRARY(COREFOUNDATION_LIBRARIES NAMES CoreFoundation)
+	IF (COREFOUNDATION_INCLUDE_DIR AND COREFOUNDATION_LIBRARIES)
+		IF (NOT CoreFoundation_FIND_QUIETLY)
+			MESSAGE("-- Found CoreFoundation ${COREFOUNDATION_LIBRARIES}")
+		ENDIF()
+		SET(COREFOUNDATION_FOUND TRUE)
+		SET(COREFOUNDATION_LDFLAGS "-framework CoreFoundation")
```
pks-t: The LDFLAGS variable may not be set in the first case, where you only check whether COREFOUNDATION_INCLUDE_DIR and COREFOUNDATION_LIBRARIES are set. If you touch this place again, I'd love if you could split this up into two patches: the first one fixes indentation only and the second one introduces your changes. This would make it a bit easier to review.

cmake/Modules/FindSecurity.cmake
```diff
@@ -1,9 +1,32 @@
...
-    SET(SECURITY_FOUND TRUE)
-  ENDIF ()
+	FIND_PATH(SECURITY_INCLUDE_DIR NAMES Security/Security.h)
+	FIND_LIBRARY(SECURITY_LIBRARIES NAMES Security)
+	IF (SECURITY_INCLUDE_DIR AND SECURITY_LIBRARIES)
+		IF (NOT Security_FIND_QUIETLY)
+			MESSAGE("-- Found Security ${SECURITY_LIBRARIES}")
+		ENDIF()
+		SET(SECURITY_FOUND TRUE)
+		SET(SECURITY_LDFLAGS "-framework Security")
```
pks-t: Same as for the other module changes. Here, both SECURITY_LDFLAGS and SECURITY_HAS_SSLCREATECONTEXT may be missing in the first case.

src/CMakeLists.txt
```diff
@@ -90,23 +90,17 @@ IF(THREADSAFE)
...
-    MESSAGE("-- Security framework is too old, falling back to OpenSSL")
-    SET(SECURITY_FOUND "NO")
-    SET(USE_OPENSSL "ON")
-  ENDIF()
+	IF (SECURITY_HAS_SSLCREATECONTEXT)
+		LIST(APPEND LIBGIT2_PC_LIBS ${SECURITY_LDFLAGS})
+	ELSE()
+		MESSAGE("-- Security framework is too old, falling back to OpenSSL")
+		SET(USE_OPENSSL "ON")
+	ENDIF()
```
pks-t: Nice to see that the main CMakeLists.txt becomes easier to read :)

CMakeLists.txt
```diff
@@ -51,6 +52,7 @@ OPTION( USE_EXT_HTTP_PARSER		"Use system HTTP_Parser if available" ON)
 OPTION( DEBUG_POOL			"Enable debug pool allocator"			OFF )
 OPTION( ENABLE_WERROR			"Enable compilation with -Werror"		OFF )
 
+
```
pks-t: Random newline here

CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ OPTION( LIBGIT2_FILENAME	"Name of the produced binary"			OFF )
 OPTION( USE_SHA1DC			"Use SHA-1 with collision detection"	OFF )
 OPTION( USE_ICONV			"Link with and use iconv library" 		OFF )
 OPTION( USE_SSH				"Link with libssh to enable SSH support" ON )
+OPTION( USE_HTTPS			"Enable HTTPS support"					ON )
```
pks-t: Maybe mention that this is not a boolean only but instead can also be a specific backend?

src/CMakeLists.txt
```diff
@@ -151,6 +141,64 @@ ELSE ()
...
+ELSEIF(HTTPS_BACKEND STREQUAL "WinHTTP")
+	# WinHTTP setup was handled in the WinHTTP-specific block above
+ELSEIF(NOT(HTTPS_BACKEND))
+	# The auto-selection wasn't able to auto-select, this will error below
+ELSE()
+	MESSAGE(FATAL_ERROR "Unknown backend ${HTTPS_BACKEND}")
+ENDIF()
+
+IF(USE_HTTPS AND NOT HTTPS_BACKEND)
+	MESSAGE(FATAL_ERROR "Asked for backend ${HTTPS_BACKEND} but it wasn't found")
```
pks-t: I think this cannot happen. If USE_HTTPS is set, then we'll always set it to a value with "OpenSSL" being the fallback. So in case of unknown backends we'll always fail with "OpenSSL not found" above, which is unfortunate.

tiennou: Actually, it should trigger when manually asking for a backend we don't support (eg. `-DUSE_HTTPS=gnuTLS`). Otherwise, with `-DUSE_HTTPS=ON` (the default) the automatic detection will auto-select OpenSSL, which will hard-fail if we can't find OpenSSL (which is what we want, since we can't obey USE_HTTPS).

I'm confused now... I'll try to come up with a cleaner set of `IF` blocks then.

src/tls_stream.c
tiennou: Will fix, I remember some conflict when rebasing, which I obviously mishandled.

cmake/Modules/FindCoreFoundation.cmake
tiennou: I'll add a reindent commit before then.

About those changes, what I'm attempting to do is stopping the `cmake` run from echoing

```
-- Found Security /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/System/Library/Frameworks/Security.framework
-- Found CoreFoundation /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/System/Library/Frameworks/CoreFoundation.framework
```

on every run. Since after the first run, everything gets cached, so this just short-circuit the test. I can also make both `FIND_PACKAGE(... QUIET REQUIRED)`, but this silences it completely...

tiennou: I ended up removing my bad attempt at using the cache correctly.

tiennou:
Rebased, and added a commit to not link against CoreFoundation when we don't use the SecureTransport backend.

Also, I feel like I should tell that CMake's OpenSSL module doesn't work against Homebrew - it ends up linking with `/usr/lib/libssl.dylib /usr/lib/libcrypto.dylib`, which are the system-supplied versions (0.9.8z IIRC).

src/CMakeLists.txt
```diff
@@ -142,6 +137,11 @@ ELSE ()
 ENDIF()
 
 IF (USE_HTTPS)
+	IF (CMAKE_SYSTEM_NAME MATCHES "Darwin")
+		FIND_PACKAGE(Security)
+		FIND_PACKAGE(CoreFoundation)
+	ENDIF()
+
```
pks-t: Can't this even be moved into `STREQUAL "SecureTransport"`? Or is it required somewhere else?

cmake/Modules/FindCoreFoundation.cmake
```diff
@@ -1,9 +1,26 @@
...
 ENDIF ()
+
+IF (CoreFoundation_FIND_REQUIRED AND NOT COREFOUNDATION_FOUND)
+	MESSAGE(FATAL "-- CoreFoundation not found")
+ENDIF()
+
+MARK_AS_ADVANCED(
+	COREFOUNDATION_INCLUDE_DIR
+	COREFOUNDATION_LIBRARIES
+)
```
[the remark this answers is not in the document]
tiennou: Yes, I was actually looking at some of CMake's modules for inspiration, and I suspect I misread `LIBRARY`/`LIBRARIES`. Will fix.

src/CMakeLists.txt
tiennou: I need to have checked for those so the auto-detection works (line 147 below). The other backends have their check before (OpenSSL is at line 125, line 103 is where we just trust WinHTTP is available).

tiennou:
On the OpenSSL point, the CMake module itself looks borked: using `-DOPENSSL_ROOT_DIR=/usr/local/opt/openssl@1.1`, it will use headers from that location but link against stuff in `/usr/lib`.

I had been able to make it link to the correct libraries *somehow* with this branch, but I don't think it was possible before those changes (at least not without fixing up what the module fails to do, and obviously if you can fix, then you didn't really have a problem), because I don't think you could prevent the usage of SecureTransport.

pks-t:
Thanks a lot! I'll review that next week.

pks-t:
Sorry for taking that long with this PR. Thanks a lot for your patience!

tiennou:
No sweat ! Imma happy contributor anyways 😉

tiennou:
I found my old cmake incantation for building against OpenSSL on macOS, which I'll post here before it's purged from my shell history :

```
cmake .. -GXcode -DUSE_HTTPS=OpenSSL -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl@1.1 -DOPENSSL_CRYPTO_LIBRARY=/usr/local/opt/openssl@1.1/lib/libcrypto.dylib -DOPENSSL_SSL_LIBRARY=/usr/local/opt/openssl@1.1/lib/libssl.dylib -DBUILD_SHARED_LIBS=ON
```

PR #4285, opened 2017-06-26
patch_parse: fix parsing unquoted filenames with spaces

pks-t:
This is split out from #4272 due to some problems with ambiguous parsing of patches with whitespace and custom prefixes. This should not be merged as-is but is only intended as a reminder.

---

The patch format allows unquoted path names even when the paths contain
whitespaces in between. To disambiguate the path names in thee header,
one can simply use the slash in "a/" respectively "b/", as no file is
allowed to contain a forward slash.

While this format is distinctly parseable, we fail to do so correctly.
To determine the path length, we simply search for the first whitespace
and assume it is the delimiter between "a/foo.file b/foo.file", which
works for the simple cases. Instead, we should be searching for the
first occurrence of either ' b/' (note the leading space) or '\n' to
either delimit the first file path or the second file path. This does
not suffice yet, though, as the format also allows for quoted file
paths. Insofar, we also have to check for ' \"b/' as delimiter for the
current file path.

Fix the issue by improving the logic determining file path length as
outlined in the above paragraph. Include a test to catch future
breakage.

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -53,11 +53,9 @@ static int header_path_len(git_patch_parse_ctx *ctx)
 	return len;
 }
 
-static int parse_header_path_buf(git_buf *path, git_patch_parse_ctx *ctx)
+static int parse_header_path_buf(git_buf *path, git_patch_parse_ctx *ctx, size_t path_len)
 {
-	int path_len, error = 0;
-
-	path_len = header_path_len(ctx);
+	int error;
 
 	if ((error = git_buf_put(path, ctx->parse_ctx.line, path_len)) < 0)
 		goto done;
@@ -81,7 +79,7 @@ done:
 static int parse_header_path(char **out, git_patch_parse_ctx *ctx)
 {
 	git_buf path = GIT_BUF_INIT;
-	int error = parse_header_path_buf(&path, ctx);
+	int error = parse_header_path_buf(&path, ctx, header_path_len(ctx));
 
 	*out = git_buf_detach(&path);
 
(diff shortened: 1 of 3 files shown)


csware:
Well, the first filename could en with a space and be under a path starting with `b`, such as `a/somethi /bla.txt` in this case a patch is not parsed correctly with this patch.

pks-t:
Yeah, correct, that's the actual issue. The issue becomes even more obvious when considering that it is possible to create custom prefixes. So instead of "a/" and "b/", the user can have "foobar/" and "whateverhelikes/" as prefixes, making it totally impossible for us to parse in all cases. One solution pointed out by @ethomson was to instead parse the "--- a/file\n+++ b/file" lines, but unfortunately they are not mandated by the patch format, either.

pks-t:
This PR now depends on #4308.

I've changed the code to follow @ethomson's proposal of just bailing on the "diff --git" line if it seems to be unparseable and instead just use the "---" and "+++" lines. What I'm still undecided about is whether we want to fill in `patch->header_old_part` and `patch->header_new_part`, which is derived from the "diff --git" line, with the contents of  `patch->old_part` and `patch->new_part`, derived from the "---" and "+++" lines. Opinions?

ethomson:
> Opinions?

ethomson:
> Opinions?

Yeah, I don't love my proposal of moving to the `---` and `+++` lines.  It was sort of a half baked idea in the first place...  I wonder if we should just get aggressive about validating the `diff --git` line and - if it's bogus - not fall back at all...?

Would your opinion about this behavior be different if I sent a patch to git to always quote filenames with spaces in them, ending this ambiguity once and for all?

(I should do that either way, of course.)

pks-t:
I definitly think this should be fixed upstream, it's simply awful right now. But that won't help users who are stuck with older versions of Git in the next 10+ years (if you count things like RHEL), so in my opinion this fix here is a nice thing to have. Bailing out whenever we cannot parse `diff --git` simply feels wrong when we can in fact just extract the correct information from the `+++` and `---` lines.

ethomson:
Sure.  Out of curiosity, how does Git itself choose the filenames?  Does it use the `+++` and `---` lines?  I was thinking that it did not, but I've actually forgotten by now, tbqh.  If it does, then it's a no brainer to emulate them.  If they do not, then it would be nice to know what they are doing and how they're working around this issue.

pks-t:
Git uses the `diff --git` line as `defname`, which is a fallback in case `oldname` or `newname` could not be derived from the `+++` and `---` lines. Here's the comment from `git_header_name`, which parses the `diff --git` thing:

```
/*
 * This is to extract the same name that appears on "diff --git"
 * line.  We do not find and return anything if it is a rename
 * patch, and it is OK because we will find the name elsewhere.
 * We need to reliably find name only when it is mode-change only,
 * creation or deletion of an empty file.  In any of these cases,
 * both sides are the same name under a/ and b/ respectively.
 */
static char *git_header_name(struct apply_state *state,
			     const char *line,
			     int llen)
```

Inside of that function:

```
		/* strip the a/b prefix including trailing slash */
		cp = skip_tree_prefix(state, first.buf, first.len);
		if (!cp)
			goto free_and_fail1;
```

So if they're not able to strip the prefixes, they simply bail out and fail. Same for other stuff if they're not able to parse the names. They only abort parsing the patch later on if neither `oldname`, `newname` nor `defname` are set.

So this PR here goes into the same direction, I'd argue.

ethomson:
Thanks for fixing this!  😀

PR #4287, opened 2017-06-30
git_reset_*: pass parameters as const pointers

AndreyG:


pks-t:
Looks good to me. Adding `const` to function parameters does not break ABI/API, so that's a non-issue, as well. 

Test failures stem from the recent breakage of the master branch on systems using bundled zlib.

PR #4288, opened 2017-06-30
Include fixups

pks-t:
Up front: sorry for this big commit, I know it sucks. But as such I'm splitting this out from #4282 as it's too big to make it port of that one.

The first two commits fix up some small stuff regarding missing license headers and to fix some header files from not compiling which weren't included anywhere before. The next commit establishes a common include pattern across all our files in the "src/" folder. Quoting from its commit message:

"""
 Make sure to always include "common.h" first

 Next to including several files, our "common.h" header also declares
 various macros which are then used throughout the project. As such, we
 have to make sure to always include this file first in all
 implementation files. Otherwise, we might encounter problems or even
 silent behavioural differences due to macros or defines not being
 defined as they should be. So in fact, our header and implementation
 files should make sure to always include "common.h" first.

 This commit does so by establishing a common include pattern. Header
 files inside of "src" will now always include "common.h" as its first
 other file, separated by a newline from all the other includes to make
 it stand out as special. There are two cases for the implementation
 files. If they do have a matching header file, they will always include
 this one first, leading to "common.h" being transitively included as
 first file. If they do not have a matching header file, they instead
 include "common.h" as first file themselves.

 This fixes the outlined problems and will become our standard practice
 for header and source files inside of the "src/" from now on.
"""

### DIFF
--- a/deps/zlib/zconf.h
+++ b/deps/zlib/zconf.h
@@ -8,7 +8,8 @@
 #ifndef ZCONF_H
 #define ZCONF_H
 
-#include "../../src/common.h"
+#include "../../include/git2/types.h"
+#include <stdarg.h>
 
 /* Jeez, don't complain about non-prototype
  * forms, we didn't write zlib */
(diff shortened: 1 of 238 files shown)


ethomson:
Seems reasonable to me. Thanks for this!

PR #4290, opened 2017-07-03
Display dependency statuses (issue #3200)

0x41111111:
This PR causes CMake to emit a few lines toward the end of the build script running, telling the user what dependencies will be used during the build process, as well as which variant of a dependency will be used (e.g. SecureTransport vs OpenSSL for HTTPS).

Related to issue #3200.


pks-t:
Hi and thanks for your changes! Just for your information: due to #4282, we're currently keeping most pull requests related to the CMakeLists.txt in hold, as it would result in big merge conflicts otherwise. As soon as #4282 is merged, I will come back to you and review your changes.

tiennou:
For the record, CMake has a [module](https://cmake.org/cmake/help/v3.8/module/FeatureSummary.html) that does something akin to what we want.

pks-t:
@tiennou Oh, that looks interesting. Looks like that should be our way forward instead. Thanks for the hint!

carlosmn:
This CMake module would be available in any installation? I would definitely much prefer an implementation which uses that.

tiennou:
I added that to #4283 actually. Seems the interface varies by CMake version, but is somewhat available.

pks-t:
I'd also say just go with that and simply ignore all older CMake versions which do not support that module.

pks-t:
This has already been implemented via #4283

Issue #4292, opened 2017-07-05
Renaming the current branch doesn't cause a message to be written to the `HEAD` reflog


jgrosso:
In the case that the currently checked-out branch is renamed, Git core has logic to write to `HEAD`'s reflog instead of only the ref-specific reflog ([https://github.com/git/git/blob/e0aaa1b6532cfce93d87af9bc813fb2e7a7ce9d7/refs/files-backend.c#L2156-L2186](https://github.com/git/git/blob/e0aaa1b6532cfce93d87af9bc813fb2e7a7ce9d7/refs/files-backend.c#L2156-L2186)). libgit2, however, only writes to the ref-specific reflog.

### Reproduction steps
1. Rename the currently checked-out branch via `git_reference_rename` (for example, `development` => `develop`).

### Expected behavior
- The reflog for the new name of the branch should be written to (for example, `.git/logs/refs/heads/develop` should contain `branch: renamed...`).
- The reflog for `HEAD` should be written to (for example, `.git/logs/HEAD` should contain `branch: renamed...`.

### Actual behavior
- The reflog for the new name of the branch is written to (for example, `.git/logs/refs/heads/develop` should contain `branch: renamed...`).
- The reflog for `HEAD` is not written to (for example, `.git/logs/HEAD` should contain `branch: renamed...`.

### Version of libgit2 (release number or SHA1)
v0.25.1

### Operating system(s) tested
macOS

tiennou:
Just a heads up, I think you want `git_branch_move` (eg. `git branch -m develop`) instead of `git_reference_rename` (eg. `git update-ref development refs/heads/develop`). I haven't checked what the reflog show in that case though, so YMMV.

jgrosso:
@tiennou Thanks! IIRC, `git_branch_move` calls `git_reference_rename` under the hood; while I'm not 100% sure of that, my testing _did_ show `git_branch_move` to have the same problem with the reflog.

tiennou:
Just checked, and there's still only the reflog line added by `git_repository_set_head` in the `HEAD` reflog. Hopefully, I'll have a preliminary PR soon-ish.

(Sorry for the spam, I rebased the test and pushed the wrong one 😟)

PR #4294, opened 2017-07-07
Fix HEAD reflog on reference renames

tiennou:
While investigating #4292, I found that we have a few tests related to what shows up in the reflog, scattered in various test cases.

This PR consolidates all "simple" ones in the same place, and also standardize some helper methods (I might convert `reflog_check_entry` to something more `clar`-y though, because line numbers are nice).

I've also added a tentative test case for #4292. The symptoms are that `git_reference_rename` reflogs :
```
0000000000000000000000000000000000000000 a65fedf39aefe402d3bb6e24df4d4f5fe4547750 unknown <unknown> 1499373433 +0200	checkout: moving from master to renamed
```
(which comes from the `git_repository_set_head` call it does).

`git` does this :
```
34bccdaf357e32eaa589f8349fb39400d4da192e 0000000000000000000000000000000000000000 me me <me@me.null> 1499372633 +0200	Branch: renamed refs/heads/development to refs/heads/develop
0000000000000000000000000000000000000000 34bccdaf357e32eaa589f8349fb39400d4da192e me <me@me.null> 1499372633 +0200	Branch: renamed refs/heads/development to refs/heads/develop
```

Fixing that requires special-casing `git_repository_set_head` so it doesn't change the reflog, so `git_reference_rename` can append those entries itself. Does that seem correct ?

While I'm at it, this raised a question: why is that reflog line triggered by `git_repository_set_head` ? I'm likely missing something but it might be cleaner to have that handled by `git_checkout_*` ?


pks-t:
Makes sense, thanks! There are a few small issues in the patch, but I currently cannot comment inline. So I'm pointing them out here:

- the new test is missing a `git_buf_free` in the commented out part
- new files are missing our license header
- there's been some trailing (possibly copied) whitespace in in `reflog_check_entry`

And yet, again we cannot really merge this due to a missing macro like `cl_git_breakage` which records known breakages. :/ Are you working on a fix for this issue?

tiennou:
I'll drop the tentative test case from that PR then, and resubmit another one with the testcase + fix.

jgrosso:
@tiennou I actually just opened a PR for #4292: #4297 🙂 . Assuming that PR is correct, am I correct in understanding that this PR will contain tests for it?

tiennou:
Yes, the (disabled part of the) test added in 3679251 checks that the HEAD reflog looks like git does when the checked-out ref is renamed.

Issue #3224, opened 2015-06-16
Pattern-based exceptions in .gitignore not working


swisspol:
Example case:

```
*.data
!dont_ignore/*.data
```

Then add a couple files in the workdir: `foo.data` and `dont_ignore/bar.data`.

`git_diff_index_to_workdir()` will correctly ignore `foo.data` but incorrectly also ignore `dont_ignore/bar.data`.

Note that the `.gitignore` file is working as expected when testing with `git status --ignored`.

swisspol:
Doing a bit more testing, this exception works: `!test.data`.

carlosmn:
It looks like the `*.data` rule satisfies the restriction that negation rules must negate an existing pattern.

```C
void test_status_ignore__deeper(void)
{
   int ignored;

    g_repo = cl_git_sandbox_init("empty_standard_repo");

    cl_git_mkfile("empty_standard_repo/.gitignore",
		  "*.data\n"
		  "!dont_ignore/*.data\n");

    cl_git_pass(p_mkdir("empty_standard_repo/dont_ignore", 0777));
    cl_git_mkfile("empty_standard_repo/foo.data", "");
    cl_git_mkfile("empty_standard_repo/bar.data", "");
    cl_git_mkfile("empty_standard_repo/dont_ignore/foo.data", "");
    cl_git_mkfile("empty_standard_repo/dont_ignore/bar.data", "");

    cl_git_pass(git_ignore_path_is_ignored(&ignored, g_repo, "foo.data"));
    cl_assert_equal_i(1, ignored);
    cl_git_pass(git_ignore_path_is_ignored(&ignored, g_repo, "bar.data"));
    cl_assert_equal_i(1, ignored);

    cl_git_pass(git_ignore_path_is_ignored(&ignored, g_repo, "dont_ignore/foo.data"));
    cl_assert_equal_i(0, ignored);
    cl_git_pass(git_ignore_path_is_ignored(&ignored, g_repo, "dont_ignore/bar.data"));
    cl_assert_equal_i(0, ignored);
}
```

fogaztuc:
Just found this issue as well.

overboen:
I believe that there is an issue in the [Atom tree-view](https://github.com/atom/tree-view) project that is caused by this via [Atom's git-utils'](https://github.com/atom/git-utils) project's dependency on this project. It looks like the project is using v0.24.3 of libgit2.

https://github.com/atom/tree-view/issues/902

[...]

PR #4296, opened 2017-07-07
Fix negative ignore rules with patterns

### LINKED ISSUE #4295: Another .gitignore issue versus standard Git
> https://github.com/libgit2/libgit2/issues/3224#issuecomment-313621374
> 
> @segevfiner I guess your issue is different, as it is caused by issues with case-insensitivity. It would be very nice if you could create a new issue describing the problem.

cc @pks-t 

CPython's has in it's .gitignore the following:
```
# Two-trick pony for OSX and other case insensitive file systems:
# Ignore ./python binary on Unix but still look into ./Python/ directory.
/python
!/Python/
```

The Python directory is not ignored by Git but is ignored by libgit2 (Found this due to Atom).

### Reproduction steps
1. Clone https://github.com/python/cpython
2. `git check-ignore -v Python` &ndash; The file isn't ignored
3. Using Pygit2:
```python
>>> import pygit2
>>> a = pygit2.Repository('.')
>>> a.path_is_ignored('Python')
True
```

### Expected behavior
The Python directory should not be ignored like in standard Git.

### Actual behavior
The Python directory is ignored by libgit2.

`pygit2.LIBGIT2_VERSION` &ndash; v0.26.0
Atom git-utils v5.0.0 uses libgit2 v0.24.3

Windows 10.0.15063 x64

pks-t:
Ignore rules allow for reverting a previously ignored rule by prefixing
it with an exclamation mark. As such, a negative rule can only override
previously ignored files. While computing all ignore patterns, we try to
use this fact to optimize away some negative rules which do not override
any previous patterns, as they won't change the outcome anyway.

In some cases, though, this optimization causes us to get the actual
ignores wrong for some files. This may happen whenever the pattern
contains a wildcard, as we are unable to reason about whether a pattern
overrides a previous pattern in a sane way. This happens for example in
the case where a gitignore file contains "*.c" and "!src/*.c", where we
wouldn't un-ignore files inside of the "src/" subdirectory.

In this case, the first solution coming to mind may be to just strip the
"src/" prefix and simply compare the basenames. While that would work
here, it would stop working as soon as the basename pattern itself is
different, like for example with "*x.c" and "!src/*.c. As such, we
settle for the easier fix of just not optimizing away rules that contain
a wildcard.

---

This should fix #4295. Edit: Oops. I mean #3224

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -125,12 +125,12 @@ static int does_negate_rule(int *out, git_vector *rules, git_attr_fnmatch *match
 				continue;
 		}
 
-	/*
-	 * When dealing with a directory, we add '/<star>' so
-	 * p_fnmatch() honours FNM_PATHNAME. Checking for LEADINGDIR
-	 * alone isn't enough as that's also set for nagations, so we
-	 * need to check that NEGATIVE is off.
-	 */
+		/*
+		 * When dealing with a directory, we add '/<star>' so
+		 * p_fnmatch() honours FNM_PATHNAME. Checking for LEADINGDIR
+		 * alone isn't enough as that's also set for nagations, so we
+		 * need to check that NEGATIVE is off.
+		 */
 		git_buf_clear(&buf);
 		if (rule->containing_dir) {
 			git_buf_puts(&buf, rule->containing_dir);
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -51,35 +51,33 @@ static int does_negate_pattern(git_attr_fnmatch *rule, git_attr_fnmatch *neg)
 	git_attr_fnmatch *longer, *shorter;
 	char *p;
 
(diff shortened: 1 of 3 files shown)


pks-t:
I've added the test from #3224 by @carlosmn

pks-t:
I've added another commit to fix an issue with case-insensitive negative ignores, reported in #4295.

Going forward, I'll most likely remove our optimization which drops negative ignores whenever they do not undo a previous pattern. This optimization has caused too much grief by now and is probably not even worth it. I'll leave this for a future PR, though.

pks-t:
Rebased to re-trigger CI and spark some interest ;)

ethomson:
> This optimization has caused too much grief by now and is probably not even worth it.

Agreed.  Merging this now, but I think this is probably the right long-term mission.

PR #4297, opened 2017-07-07
Write to the HEAD reflog when renaming the currently active branch

### LINKED ISSUE #4292: Renaming the current branch doesn't cause a message to be written to the `HEAD` reflog

jgrosso:
Resolves #4292.

### DIFF
--- a/src/refdb_fs.c
+++ b/src/refdb_fs.c
@@ -1368,7 +1368,7 @@ static int refdb_fs_backend__rename(
 	const char *message)
 {
 	refdb_fs_backend *backend = (refdb_fs_backend *)_backend;
-	git_reference *old, *new;
+	git_reference *old, *old_copy, *new;
 	git_filebuf file = GIT_FILEBUF_INIT;
 	int error;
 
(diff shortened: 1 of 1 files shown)


tiennou:
Here are the lines added to the HEAD reflog after the test executes.

```
0: 000000000000000000000000000000000000000 a65fedf39aefe402d3bb6e24df4d4f5fe454775 somesig checkout: moving from master to renamed
1: 000000000000000000000000000000000000000 a65fedf39aefe402d3bb6e24df4d4f5fe454775 somesig (null)
```

`1` is the message I'm passing when renaming, `0` is the result of the `git_repository_set_head` call.

So "in theory" it works : the message when renaming now appears in the HEAD reflog. But it doesn't match what git does.

ethomson:
Closing as stale — we need to always match what git does

Issue #4068, opened 2017-01-05
git_submodule_foreach segfaults in a bare repo


novalis:
Test case
[submodule_foreach_bare.patch.txt](https://github.com/libgit2/libgit2/files/686364/submodule_foreach_bare.patch.txt)

You can see in the backtrace that, because the wd check here https://github.com/libgit2/libgit2/blob/89c332e41b12a72d89de40d63bc568c56a2c336a/src/submodule.c#L459 fails, mods is still NULL -- but submodules_from_index and submodules_from_head need it to be non-NULL.

We could just set it to an empty config in a bare repo, which would stop the crash, but it wouldn't actually enable data about those modules to be correctly loaded, since e.g. submodules_from_index seems to depend on the .gitmodules in the working directory (that is, it has no provision to load one from the ODB).

The root cause of this is that the submodule API has a few separate concepts mixed together.  Here's a classification:

Operations on in-memory object, or operations that only affect the submodule's dir in .git/modules:
git_submodule_free
git_submodule_owner
git_submodule_repo_init (when used correctly, that is, with use_gitlink=true)
git_submodule_set_fetch_recurse_submodules
git_submodule_set_ignore
git_submodule_set_update
git_submodule_update_init_options
git_submodule_url

git_submodule_branch
git_submodule_fetch_recurse_submodules
git_submodule_ignore
git_submodule_name
git_submodule_path
git_submodule_update_strategy

The second set of these depend on various configuration, some of which will come from the config and some of which originate from a .gitmodules file, probably the one in the wd -- I haven't looked carefully.

Operations that depend on or affect the working directory:
git_submodule_add_setup
git_submodule_init
git_submodule_open
git_submodule_sync
git_submodule_update
git_submodule_wd_id

Operations that depend on or affect the index:
git_submodule_add_finalize
git_submodule_index_id

Operations that depend on or affect the HEAD:
git_submodule_head_id

Operations that depend on or affect the wd and index:
git_submodule_add_to_index

Operations that depend on or affect the wd, index, and HEAD:
git_submodule_foreach
git_submodule_location
git_submodule_lookup
git_submodule_reload
git_submodule_status

Further, there is no way to access information about a submodule's state as-of a prior commit (at least that I can see in the git_submodule_xxx api).  

Perhaps the right way to handle submodules is to model submodules using a few kinds of object: 
- a "submodule configuration set", representing the contents of a .gitmodules file, possibly one in the index or in a commit, and optionally as modified by local config.
[...]

PR #4305, opened 2017-07-10
Submodules with bare repo

pks-t:
Fixes to some small issues with our submodule tests as well as explicit error returns when trying to access submodules in a bare repository. Related to #4068, even though the issue does not seem to occur anymore


pks-t:
Ugh. The added bare repository contained an empty 'refs' directory due to all references being packed info the packed-refs file. As git doesn't store empty directories, it is missing in our repository, and git only regards a directory as a repository if it also contains a 'refs' directory. Should be fixed now.

PR #4306, opened 2017-07-12
signature: don't leave a dangling pointer to the strings on parse failure

carlosmn:
If the signature is invalid but we detect that after allocating the strings, we
free them. We however leave that pointer dangling in the structure the caller
gave us, which can lead to double-free.

Set these pointers to `NULL` after freeing their memory to avoid this.


pks-t:
Looks obviously correct! Test failures were (yet again) unrelated. :+1:

PR #4308, opened 2017-07-14
patch_parse: implement state machine for parsing patch headers

pks-t:
Our code parsing Git patch headers is rather lax in parsing headers of a
Git-style patch. Most notably, we do not care for the exact order in
which header lines appear and as such, we may parse patch files which
are not really valid after all. Furthermore, the state transitions
inside of the parser are not as obvious as they could be, making it
harder than required to follow its logic.

To improve upon this situation, this patch introduces a real state
machine to parse the patches. Instead of simply parsing each line
without caring for previous state and the exact ordering, we define a
set of states with their allowed transitions. This makes the patch
parser more strict in only allowing valid successions of header lines.
As the transition table is defined inside of a single structure with
the expected line, required state as well as the state that we end up
in, all state transitions are immediately obvious from just having a
look at this structure. This improves both maintainability and eases
reasoning about the patch parser.

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -372,44 +372,8 @@ static int parse_header_dissimilarity(
 	return 0;
 }
 
-typedef struct {
-	const char *str;
-	int(*fn)(git_patch_parsed *, git_patch_parse_ctx *);
-} header_git_op;
-
-static const header_git_op header_git_ops[] = {
-	{ "diff --git ", NULL },
-	{ "@@ -", NULL },
-	{ "GIT binary patch", NULL },
-	{ "Binary files ", NULL },
-	{ "--- ", parse_header_git_oldpath },
-	{ "+++ ", parse_header_git_newpath },
-	{ "index ", parse_header_git_index },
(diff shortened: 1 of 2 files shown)


pks-t:
Rebased on current master. This is mainly to have the base same commit in #4310, which builds upon this PR.

src/patch_parse.c
```diff
@@ -439,6 +439,7 @@ static const parse_header_transition transitions[] = {
 	/* Next patch */
 	{ "diff --git "         , STATE_END,        0,                NULL },
 	{ "@@ -"                , STATE_END,        0,                NULL },
+	{ "-- "                 , STATE_END,        0,                NULL },
```
pks-t: Doesn't the fix to #4322 look nice with this new state machine? :)

pks-t:
@ethomson You think you might get some time soonish to review? I'd like to move on with the common parser interface in #4310, which depends on this PR here. More specifically, I wanted to also take a look at moving our attributes parser to use the new parsing interface, as well.

ethomson:
THIS IS AWESOME.  🎉

PR #4310, opened 2017-07-15
Common parser interface

pks-t:
This is a first attempt at pulling out the diff-header parser and unify its interface, such that it can be reused for other code parts. In this PR I did a PoC to replace the config file parser with this common parser code, which turns out to be quite promising in that it removes roughly 150 lines of code which are not required anymore.

There are two issues left with this PR, which is why I labeled it as WIP. First, there are some small-ish issues with proper indentation when writing config files, which I'll handle as soon as the second problem is solved. The second problem is with recursive include files, where the wrong structures are touched due to our faulty include handling. These things should be fixed by #4250.

As there are quite a lot of merge conflicts due to #4250, I'll stop working on this issue right now until the other PR is merged.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -1115,6 +1115,7 @@ static int config_read(
 {
 	struct parse_data parse_data;
 	git_config_parser reader;
+	git_buf contents = GIT_BUF_INIT;
 	int error;
 
 	if (depth >= MAX_INCLUDE_DEPTH) {
@@ -1122,22 +1123,20 @@ static int config_read(
 		return -1;
 	}
 
-	git_buf_init(&reader.buffer, 0);
-
-	if ((error = git_futils_readbuffer(&reader.buffer, file->path)) < 0)
+	if ((error = git_futils_readbuffer(&contents, file->path)) < 0)
 		goto out;
 
-	if ((error = git_hash_buf(&file->checksum, reader.buffer.ptr, reader.buffer.size)) < 0)
+	git_parse_ctx_init(&reader.ctx, contents.ptr, contents.size);
+
+	if ((error = git_hash_buf(&file->checksum, contents.ptr, contents.size)) < 0)
 		goto out;
 
 	/* Initialize the reading position */
 	reader.file = file;
-	reader.line_number = 0;
-	reader.read_ptr = reader.buffer.ptr;
-	reader.eof = 0;
+	git_parse_ctx_init(&reader.ctx, contents.ptr, contents.size);
 
 	/* If the file is empty, there's nothing for us to do */
-	if (*reader.read_ptr == '\0')
+	if (!reader.ctx.content || *reader.ctx.content == '\0')
 		goto out;
 
 	parse_data.repo = repo;
@@ -1149,7 +1148,7 @@ static int config_read(
 	error = git_config_parse(&reader, NULL, read_on_variable, NULL, NULL, &parse_data);
 
 out:
-	git_buf_free(&reader.buffer);
+	git_buf_free(&contents);
 	return error;
 }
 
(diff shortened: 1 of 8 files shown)


pks-t:
I've rebased on master with #4250. As expected, the problem with recursive includes was solved by this. Only remaining problem now is improper indentation, which I'll fix later

pks-t:
Fixed remaining issues. Next steps which I'll not handle inside of this PR:

- rewrite attributes parser to use common interface
- convert existing parsers into state machines

pks-t:
Rebased upon latest master to fix a single conflict due to the new "common.h" headers. I've also added commit messages for every commit.

pks-t:
Gentle ping @ethomson @carlosmn :)

src/patch_parse.c
```diff
@@ -364,39 +284,81 @@ static int parse_header_dissimilarity(
...
+	STATE_MODE,
+	STATE_INDEX,
+	STATE_PATH,
+
+	STATE_SIMILARITY,
+	STATE_RENAME,
+	STATE_COPY,
+
+	STATE_END,
+} parse_header_state;
```
carlosmn: It'd be nice to have some documentation of what each of these mean. I was shortly confused by `STATE_END` being the start start state for `@@` which happens in the middle of a patch.

src/patch_parse.c
```diff
@@ -405,65 +367,58 @@ static int parse_header_git(
...
-		for (i = 0; i < ARRAY_SIZE(header_git_ops); i++) {
-			const header_git_op *op = &header_git_ops[i];
-			size_t op_len = strlen(op->str);
+		for (i = 0; i < ARRAY_SIZE(transitions); i++) {
+			const parse_header_transition *transition = &transitions[i];
+			size_t op_len = strlen(transition->str);
 
-			if (memcmp(ctx->line, op->str, min(op_len, ctx->line_len)) != 0)
+			if (transition->expected_state != state ||
+			    memcmp(ctx->parse_ctx.line, transition->str, min(op_len, ctx->parse_ctx.line_len)) != 0)
```
carlosmn: I think `git__prefixcmp` would be more appropriate here. Otherwise you're letting this match when _either_ string is the prefix of the other.

src/patch_parse.c
```diff
@@ -364,39 +284,81 @@ static int parse_header_dissimilarity(
...
+	STATE_SIMILARITY,
+	STATE_RENAME,
+	STATE_COPY,
+
+	STATE_END,
+} parse_header_state;
+
 typedef struct {
 	const char *str;
+	parse_header_state expected_state;
```
carlosmn: Is this called `expected_state` because it's what you expect to be in in order to match? If we're going to go for a state machine, I think it makes more sense for this to be called `current_state`.

src/parse.c
```diff
@@ -0,0 +1,127 @@
...
+{
+	if (content_len) {
+		if ((ctx->content = git__malloc(content_len)) == NULL) {
+			return -1;
+		}
+
+		memcpy((char *)ctx->content, content, content_len);
+	} else {
+		ctx->content = NULL;
+	}
```
carlosmn: Must we copy the contents? We're already read whatever file or buffer into memory.

src/parse.c
```diff
@@ -0,0 +1,127 @@
...
+int git_parse_advance_nl(git_parse_ctx *ctx)
+{
+	if (ctx->line_len != 1 || ctx->line[0] != '\n')
+		return -1;
+
+	git_parse_advance_line(ctx);
+	return 0;
+}
+
+int git_parse_advance_digit(git_off_t *out, git_parse_ctx *ctx, int base)
```
carlosmn: This doesn't match the type in the header. We should have `int64` here.

src/parse.c
```diff
@@ -0,0 +1,127 @@
...
+		return -1;
+
+	if ((ret = git__strntol64(out, ctx->line, ctx->line_len, &end, base)) < 0)
+		return -1;
+
+	git_parse_advance_chars(ctx, (end - ctx->line));
+	return 0;
+}
+
+int git_parse_peek(char *out, git_parse_ctx *ctx, int flags)
```
carlosmn: A lot of the uses for this do just want to look at the next character. It might not be a huge deal but I worry about the performance hit when we have this loop in the function that we don't usually want.

src/config_file.c
```diff
@@ -1915,36 +1254,30 @@ static int config_write(diskfile_backend *cfg, const char *key, const regex_t *p
...
 	/* Initialise the reading position */
-	if (result == GIT_ENOTFOUND) {
-		reader.read_ptr = NULL;
-		reader.eof = 1;
-		git_buf_clear(&reader.buffer);
-	} else if (result == 0) {
-		reader.read_ptr = reader.buffer.ptr;
-		reader.eof = 0;
+	if (result == 0 || result == GIT_ENOTFOUND) {
+		git_parse_ctx_init(&reader.ctx, contents.ptr, contents.size);
```
carlosmn: Do we guarantee that we won't touch the fields in `contents` if it's not found? It makes sense that we wouldn't touch them, but if that's not part of the contract, this could initialise the context with whatever random pointer.

ethomson: Do you mean the contents of the context?

src/parse.c
pks-t: Well, if the skip-whitespace flag is not set we're not looping at all. Are you worried about bad code generated by the compiler or just about calling that function multiple times? I mean I could also split that up into two different functions `git_parser_peek` and `git_parser_peek_skip_whitespace` (naming should obviously be improved), but I don't think it's required right now.

src/parse.c
pks-t: Fixed

src/parse.c
pks-t: Fixed.

src/patch_parse.c
pks-t: Fixed

src/patch_parse.c
```diff
@@ -364,39 +284,81 @@ static int parse_header_dissimilarity(
...
+	{ "rename from "        , STATE_SIMILARITY, STATE_RENAME,     parse_header_renamefrom },
+	{ "rename old "         , STATE_SIMILARITY, STATE_RENAME,     parse_header_renamefrom },
+	{ "copy from "          , STATE_SIMILARITY, STATE_COPY,       parse_header_copyfrom },
+	{ "rename to "          , STATE_RENAME,     STATE_END,        parse_header_renameto },
+	{ "rename new "         , STATE_RENAME,     STATE_END,        parse_header_renameto },
+	{ "copy to "            , STATE_COPY,       STATE_END,        parse_header_copyto },
+
+	/* Next patch */
+	{ "diff --git "         , STATE_END,        0,                NULL },
+	{ "@@ -"                , STATE_END,        0,                NULL },
```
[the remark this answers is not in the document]
pks-t: Yep. Fixed

src/patch_parse.c
pks-t: Agreed

PR #4311, opened 2017-07-18
win32: provide fast-path for retrying filesystem operations

ethomson:
When using the `do_with_retries` macro for retrying filesystem operations in the posix emulation layer, allow the remediation function to return `GIT_RETRY`, meaning that the error was believed to be remediated, and the operation should be retried immediately, without a sleep.

Use this in the `ensure_writable` remediation function - if the function is able to set the file writable, then retry the operation immediately.  This will prevent a 5ms sleep in the common case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ v0.26 + 1
 
 ### Changes or improvements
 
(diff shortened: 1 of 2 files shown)


ethomson:
Thanks @cjhoward92 !

PR #4312, opened 2017-07-18
fixed win32 p_unlink retry sleep issue

cjhoward92:
Fixed an issue where the retry logic on `p_unlink` in `posix_w32.c` sleeps before it tries setting a file to write mode causing unnecessary slowdown.

This affects windows only.

The issue was introduced [in this commit](https://github.com/nodegit/libgit2/commit/a0f67e4a262cfdc6d430d969a9a257397e4b0406) as the `do_with_retries` macro tries the `unlink_once` routine and sleeps on failure before calling `ensure_writable`.

This is particularly noticeable when staging large groups of files at once as staging creates a temporary git object that is set to read only. When this temporary git object is deleted via `p_unlink` it always ends up hitting the sleep causing a 5ms delay per file.

I spoke with @ethomson on slack about this briefly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ v0.26 + 1
 
 ### Changes or improvements
 
-* Improved `p_unlink` in `posix_w32.c` to try and make a file writable before sleeping in the retry loop to prevent unnecessary sleep calls.
+* Improved `p_unlink` in `posix_w32.c` to try and make a file writable before sleeping in
+  the retry loop to prevent unnecessary calls to sleep.
 
 ### API additions
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ v0.26 + 1
 
 ### Changes or improvements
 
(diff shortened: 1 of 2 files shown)


cjhoward92:
Did not notice [this PR](https://github.com/libgit2/libgit2/pull/4311).

pks-t:
Heh, you've been late by only 5 minutes. Thanks anyway for your PR!

cjhoward92:
@pks-t Yeah! I was  little bummed but that's okay as long as it gets fixed! Thank you for helping to maintain the library! It's great.

ethomson:
Yeah, sorry again for the poor communication on my part @cjhoward92.  I'm manually merging your pull request into my branch, because I want to make sure that your change gets in to the code base.  😀  I've rebased my change on top, since I think that it's a little bit more general-purpose, and I'll merge them both in #4311.

Thanks again for the investigation and the fix!  🎉

cjhoward92:
Wow thanks @ethomson! Appreciate it!

PR #4316, opened 2017-07-21
Fix reference renaming reflog

### LINKED ISSUE #4292: Renaming the current branch doesn't cause a message to be written to the `HEAD` reflog

tiennou:
Note that this is on top of the gather-reflog-message PR, because it's reflog-message-related.

This fixes #4292 (test in 79913e1) and a bug where we would lose reflogs (cdc82238, test in a90cf54).

As this is my first foray into the refdb code, I hope I didn't do toooo badly 😉. The first few commits are prerequisite work on massaging the code to be more amenable to the task at hand. Highlights are :
- eb7349b, where some flags are added so the later code can be more granular on what it needs to be done.
- 68c1d75, where the various `*_delete` calls gain the standard `who/signature` pair, because the reflog needs it.
- cdc8238-64df22c, where the ref-renaming code is gradually updated/bugfixed until it behaves correctly.


tiennou:
Rebased, which makes the commits ID above slightly wrong 🤷‍♂️ .

include/git2/sys/refdb_backend.h
```diff
@@ -107,7 +107,11 @@ struct git_refdb_backend {
 	 * from the refdb.  A refdb implementation must provide this
 	 * function.
 	 */
-	int (*del)(git_refdb_backend *backend, const char *ref_name, const git_oid *old_id, const char *old_target);
+	int (*del)(
+		git_refdb_backend *backend,
+		const char *ref_name,
+		const git_signature *who, const char *message,
+		const git_oid *old_id, const char *old_target);
```
pks-t: This is unfortunate, as the change breaks the API. I'd rather opt for a new method, as libgit2 should finally grow up and be more serious about API stability.

include/git2/transaction.h
```diff
@@ -92,7 +92,7 @@ GIT_EXTERN(int) git_transaction_set_reflog(git_transaction *tx, const char *refn
  * @param refname the reference to remove
  * @return 0, GIT_ENOTFOUND if the reference is not among the locked ones, or an error code
  */
-GIT_EXTERN(int) git_transaction_remove(git_transaction *tx, const char *refname);
+GIT_EXTERN(int) git_transaction_remove(git_transaction *tx, const char *refname, const git_signature *sig, const char *msg);
```
pks-t: This one here is breaking API, as well, and should thus be a new function.

src/refdb_fs.c
```diff
@@ -1064,6 +1064,30 @@ static int packed_write(refdb_fs_backend *backend)
...
 
+static int packed_delete(refdb_fs_backend *backend, const char *ref_name)
+{
+	size_t pack_pos;
+	int error;
+
+	if ((error = packed_reload(backend)) < 0)
+		goto cleanup;
+
+	/* If a packed reference exists, remove it from the packfile and repack */
```
pks-t: This comment belongs to the next block

src/refdb_fs.c
```diff
@@ -1064,6 +1064,30 @@ static int packed_write(refdb_fs_backend *backend)
...
+	if ((error = git_sortedcache_wlock(backend->refcache)) < 0)
+		goto cleanup;
+
+	if (!(error = git_sortedcache_lookup_index(
+			&pack_pos, backend->refcache, ref_name)))
+		error = git_sortedcache_remove(backend->refcache, pack_pos);
+
+	git_sortedcache_wunlock(backend->refcache);
+
+	error = packed_write(backend);
```
pks-t: This line is clobbering the previous error code. Furthermore, I'd expect no `packed_write` to be necessary in case we do not remove the entry. Maybe you can just move the unlock into `cleanup` and set a flag as soon as the lock has been taken. Then you can easily convert the lookup/remove into our usual `if ((error = fn(...)) < 0)) goto cleanup;`.

src/refdb_fs.c
```diff
@@ -1170,20 +1170,18 @@ static int cmp_old_ref(int *cmp, git_refdb_backend *backend, const char *name,
...
-	git_oid old_id;
 	git_reference *tmp = NULL, *head = NULL, *peeled = NULL;
 	const char *name;
 
 	if (ref->type == GIT_REF_SYMBOLIC)
 		return 0;
 
-	/* if we can't resolve, we use {0}*40 as old id */
-	if (git_reference_name_to_id(&old_id, backend->repo, ref->name) < 0)
-		memset(&old_id, 0, sizeof(old_id));
```
pks-t: Previously, we've always passed a valid old ID into `reflog_append`, where we now always pass `NULL` to it (as there seems to be only one caller, which passes `NULL`). This is no problem, though, as `reflog_append` does the exact same dance that is getting deleted here in case `old` is `NULL`.

tests/refs/branches/delete.c
```diff
@@ -45,6 +45,8 @@ void test_refs_branches_delete__can_delete_a_branch_even_if_HEAD_is_missing(void
 	git_reference *head;
 	git_reference *branch;
 
+	cl_skip(); // deleting HEAD feels non-git ?
```
pks-t: So I assume this test doesn't work anymore after your changes? Having no HEAD is definitely weird and shouldn't be supported to do. But being able to still cope with such repos is a nice thing to have -- you never know how a user ended up with this situation, but we support it right now and shouldn't regress that.

src/refdb_fs.c
```diff
@@ -1399,16 +1399,14 @@ static int refdb_fs_backend__delete_tail(
...
-		}
-	}
+    if ((error = should_write_reflog(&should_write, backend, ref)) < 0)
+        goto cleanup;
+
+    if (should_write) {
+        const git_oid *old_id = git_reference_target(ref);
+        if ((error = maybe_append_head(backend, ref, old_id, NULL, who, message)) < 0)
+            goto cleanup;
+    }
```
pks-t: These blocks are using spaces instead of tabs

src/refdb_fs.c
```diff
@@ -1399,16 +1399,14 @@ static int refdb_fs_backend__delete_tail(
 		goto cleanup;
 	}
 
-	if ((flags & GIT_REFDB__SKIP_REFLOG) == 0) {
```
pks-t: It would be nice to have an explanation here why we'd always want to write that log in the commit message. At least I'm a bit clueless why that should be the case, when not writing it has been requested

src/refdb_fs.c
```diff
@@ -1479,12 +1489,25 @@ static int refdb_fs_backend__rename(
...
+	if (is_head) {
+		git_filebuf head_lock = GIT_FILEBUF_INIT;
+		head->target.symbolic = git__strdup(new_name);
+		if ((error = refdb_fs_backend__write_tail(&head_lock, _backend, head,
+				NULL, old_head_target, GIT_REFDB__FORCE_WRITE, who, message)) < 0) {
+			goto cleanup;
+		}
+		if ((error = loose_commit(&head_lock, head)) < 0)
+			goto cleanup;
+	}
```
pks-t: This code feels a bit like a layering violation to me. What happens in case where we rename a reference that is _not_ part of the FS refdb, but part of a different one? Or in case where HEAD itself is not part of the FS refdb? Sure, especially the second scenario is unlikely, but it just feels a bit weird to handle HEAD at this level.

What confuses me is that in a later commit, you delete code that is supposed to update HEAD. Why is it necessary to move the code into refdb_fs and why didn't the other code suffice? Furthermore, shouldn't these two commits belong together then, ideally with your line of thought why the move is necessary?

tiennou:
Sorry for the lack of reasoning. There's additional context in #4294, but the gist of the issue is that 1) we don't reflog things like git does when deleting and renaming branches (we must pass more context when deleting, since we might need to update the HEAD reflog, which happens anytime you can see a `who, msg` signature pair), 2) we will break things when deleting refs if anything goes wrong (usually by eagerly deleting reflogs).

I remember wondering about layering here too: right now the reflog handling seems brittle IMHO, because our refdb_fs will take care of some parts (it `maybe_appends_head` when writing), but it's not sufficient for the harder cases (renames & deletes).

An alternative implementation would be to build on the `git_transaction` API so that we can have `append_reflog`/`rename_reflog` be handled atomically within the actual `git__reference_delete`/`git__reference_rename` functions, and hopefully leave the low-level methods cleaner. That would be a much more invasive change to the refdb API though, because all those `who, msg` could be removed in favor of a `git_transaction_append_reflog/rename_reflog` API.

src/refdb_fs.c
tiennou: I agree on the layering, though I'm not sure I understand your cases : the first one should not happen, because the current refdb is being called on behalf of it being the ref's backend. The second is already true because the refdb already relookups HEAD when writing.

The strange-two step process is there to make the reflog actually correct (before the last step, the `git_repository_set_head` call would insert a third, duplicate message). Looking back, they can be merged, so I don't know why I kept them separate. As to the why, it's because the reference renaming process is much more finicky with locks than what can be done at this level, so it's either "dump everything on the refdb so it can handle it" (this approach) or "make it more transaction-y" so we can work at an higher level.

src/refdb_fs.c
tiennou: IIRC this is so the HEAD reflog gets its `0124adbc 00000000 sig Branch: renamed foo to bar`, which marks the deletion (the first part of the "renaming" pair). But I'm not quite sure anymore 😞.

src/refdb_fs.c
tiennou: Will fix.

src/refdb_fs.c
tiennou: Not sure what to make of that, so thanks, I guess 😉? My rational argument would be that it moves the duties of formatting NULL to `reflog_append`.

src/refdb_fs.c
tiennou: I don't think I can move the unlock to `cleanup` because the `backend->refcache` lock is not reentrant and `packed_write` will (dead-)lock it. I'll fix the clobber though.

pks-t:
Hm. As far as I see, the only API changes are with regards to introducing the author and message arguments so that we are able to write correct reflogs, right? What about just adding two new function calls instead which do take these parameters. In case where a backend has these functions we use them, otherwise we'll fall back to the old ones. In that case, we have to accept the fact that some backends simply can't handle correct reflogs.

As one example, doing the API change would break the redis refdb backend that lives in libgit2/libgit2-backends.

tiennou:
Yeah, looking more closely at it, the API changes is caused because I'm fixing HEAD-reflog-isn't-correct-on-unborning-it (a.k.a the deletion part), submit that as another PR, and then I can consider that "a missing feature" and not get bothered about backward-compatibility 😉.

I've tried to do a clean API break, and it's messy. I think I have to update the version (which we're not enforcing anywhere BTW), else I'd end up checking some trailing memory when using an old-style backend. I also can't add the implementation *next to* `del` without also breaking old backends. So I'm left with a `del_v2` function far from it's deprecated friend, which I'm not fond of, and an update to the version. Which means I could just copy the whole struct as `git_refdb_backend_v2`, fix the del function there, and autoupdate v1 backends to v2 via a wrapper for the `del` style-change.

I'm likely to split out of this PR all that deletion stuff to another one, so at least the reported issue (and bugs) can be fixed. The deletion is just something I found out while experimenting on reflogs.

pks-t:
> I've tried to do a clean API break, and it's messy. I think I have to update the version (which we're not enforcing anywhere BTW), else I'd end up checking some trailing memory when using an old-style backend. I also can't add the implementation next to del without also breaking old backends. So I'm left with a del_v2 function far from it's deprecated friend, which I'm not fond of, and an update to the version. Which means I could just copy the whole struct as git_refdb_backend_v2, fix the del function there, and autoupdate v1 backends to v2 via a wrapper for the del style-change.

Yeah, it'll get messy. I think disambiguating between two different structs based on whether the version field indicates old or new layout would be the easiest way to get it right. It has its own problems, though, as we'd first have to cast the old version struct to the new one, which is not allowed in C.

Easiest might be something like this:

```
struct opts {
    int version;
    char *opt;
};

struct optsv2 {
    struct opts v1;
    char *new_opt;
}

struct optsv3 {
    union {
        struct opts v1;
        struct optsv2 v2;
    }
    char *newest_opt;
}
```

Like this, we can guarantee that the old options struct will never change while being able to add new options. Code might look as following:

```
int fn(struct opts *_o)
{
    struct optsv3 o = INIT_DEFAULTS;

    if (_o && o->version == 1) {
        memcpy(&o.v1, _o, sizeof(o.v1));
    } else if (&_o && o->version == 2) {
        memcpy(&o.v2, _o, sizeof(o.v2));
    } else if (&_o && o->version == 3) {
        memcpy(&o, _o, sizeof(o.v3));
    }

    o.v1.opt;
    o.v2.new_opt;
    o.newest_opt;
}
```

Might be messy, but I think it's the easiest to get right because we don't need to care about keeping the memory layout of old options. Alternatively, one could just make sure to only ever append new fields to the options struct. But it makes it harder than necessary to keep track of how many bytes we're allowed to memcpy.

I don't know. I guess we need to learn how to do this properly now.

ethomson:
Hmm.  This is a little tricky, because there is no `o->version` in your example.  It's only accessible as `o->v1->version` or `o->v2->v1->version`...

I think I see where you're going though, and it's nice in the sense that we don't have to duplicate options in structs (v2 can inherit from v1 and then only worry about the new bits, if I'm thinking about this correctly) but I think it's going to be incredibly confusing for end-users.  In your example, if I have `struct optsv3 o = INIT_DEFAULTS;` then obviously I should set `o.newest_opt`, but what if I don't?  What if I want to keep setting the `v2` option because that's the one that I was using before?  So what if I instead set `o.v2.new_opt`?  Or `o.v1.opt`?  (Or worse yet, `o.v2.v1.opt`?

If I wanted to opt in to a newer set of options, now I have to understand exactly what is deprecated and needs to move to the new-style.  (I think.  Or we have to figure out how to understand this options structure.)

I would propose that we introduce a dummy structure that only has a version component:

```
struct git_opt_t {
    int version;
};
```

Then we can:

```
struct opts {
    int version;
    char *opt;
    char *unchanged;
};

struct optsv2 {
    int version;
    char *new_opt;
    char *unchanged;
}

struct optsv3 {
    int version;
    char *newest_opt;
    char *unchanged;
}
```

Now, as an end-user, I get broken hard and clearly if I try to upgrade to `optsv3` and I know that I have to set `opt` instead of `newest_opt`.  But many options get to stay unchanged and my code stays as-is.

Now in our bits that take opts from a user (eg, `checkout_data_init` and the like) then we would need to do a bit more work, I concede.  Something like:

```
struct optsv3 opts = INIT;

switch (((git_opt_t *)given_opts)->version) {
case 1:
    /* call a bespoke function that knows how to convert v1 to v3 */
    break;
case 2:
    /* call a bespoke function that knows how to convert v2 to v3 */
    break;
case 3:
    memcpy(&opts, given_opts, sizeof(struct optsv3));
    break;    
default:
    system("shutdown -r now"); /* UB */
}
```

src/refdb_fs.c
```diff
@@ -1090,22 +1090,27 @@ static int packed_write(refdb_fs_backend *backend)
...
-	if (!(error = git_sortedcache_lookup_index(
-			&pack_pos, backend->refcache, ref_name)))
+	/* If a packed reference exists, remove it from the packfile and repack if necessary */
+	error = git_sortedcache_lookup_index(&pack_pos, backend->refcache, ref_name);
+	if (error == 0) {
 		error = git_sortedcache_remove(backend->refcache, pack_pos);
+		found = 1;
+	}
+	if (error == GIT_ENOTFOUND)
+		error = 0;
```
pks-t: Wouldn't the general flow be a bit easier if we'd not introduce a `found` but a `locked` variable, instead?

```
static int packed_delete(refdb_fs_backend *backend, const char *ref_name)
{
	int error, locked = 0;
	size_t pack_pos;

	if ((error = packed_reload(backend)) < 0)
		goto cleanup;

	/* If a packed reference exists, remove it from the packfile and repack */
	if ((error = git_sortedcache_wlock(backend->refcache)) < 0)
		goto cleanup;

	locked = 1;

	if ((error = git_sortedcache_lookup_index(&pack_pos, backend->refcache, ref_name)) < 0) {
		if (error == GIT_ENOTFOUND)
			error = 0;
		goto cleanup;
	}

	if ((error = git_sortedcache_remove(backend->refcache, pack_pos)) < 0)
		goto cleanup;

	if ((error = packed_write(backend)) < 0)
		goto cleanup;

cleanup:
	if (locked)
		git_sortedcache_wunlock(backend->refcache);
	return error;
}
```

Might just be me, though

src/refdb_fs.c
```diff
@@ -43,6 +43,14 @@ enum {
 	PEELING_FULL
 };
 
+typedef enum {
+	GIT_REFDB__FORCE_WRITE = (1 << 0),
+	GIT_REFDB__SKIP_REFLOG = (1 << 1),
```
pks-t: I think these two flags should be documented, as well. `SKIP_REFLOG` is obvious, but I'm left to guess with regards to the first one.

src/refdb_fs.c
```diff
@@ -866,12 +875,20 @@ static int refdb_fs_backend__unlock(git_refdb_backend *backend, void *payload, i
 	git_filebuf *lock = (git_filebuf *) payload;
 	int error = 0;
 
-	if (success == 2)
-		error = refdb_fs_backend__delete_tail(backend, lock, ref->name, NULL, NULL);
-	else if (success)
-		error = refdb_fs_backend__write_tail(backend, ref, lock, update_reflog, NULL, NULL, sig, message);
-	else
-		git_filebuf_cleanup(lock);
+	git__refdb_flags flags = (GIT_REFDB__SKIP_LOCK | GIT_REFDB__FORCE_WRITE);
```
pks-t: I think we had this issue once before, where I once encountered a very specific compiler that didn't allow a newline between block of variable declarations and another variable in C90. So the variable declaration should probably be moved to the other declarations.

src/refdb_fs.c
```diff
@@ -866,12 +875,20 @@ static int refdb_fs_backend__unlock(git_refdb_backend *backend, void *payload, i
...
+	if (!update_reflog) {
+		flags |= GIT_REFDB__SKIP_REFLOG;
+	}
```
pks-t: These braces aren't required

src/refdb_fs.c
```diff
@@ -866,12 +875,20 @@ static int refdb_fs_backend__unlock(git_refdb_backend *backend, void *payload, i
...
+	if (success == 2) {
+		error = refdb_fs_backend__delete_tail(lock, backend, ref->name, flags, NULL, NULL);
+	} else if (success) {
+		error = refdb_fs_backend__write_tail(lock, backend, ref, flags, NULL, NULL, sig, message);
+		if (error == 0) {
+			error = loose_commit(lock, ref);
+		}
+	}
```
pks-t: This is awful. What the heck is `success == 2` supposed to mean? Not your fault, though

src/refdb_fs.c
```diff
@@ -866,12 +875,20 @@ static int refdb_fs_backend__unlock(git_refdb_backend *backend, void *payload, i
...
+	git__refdb_flags flags = (GIT_REFDB__SKIP_LOCK | GIT_REFDB__FORCE_WRITE);
+	if (!update_reflog) {
+		flags |= GIT_REFDB__SKIP_REFLOG;
+	}
+	if (success == 2) {
+		error = refdb_fs_backend__delete_tail(lock, backend, ref->name, flags, NULL, NULL);
+	} else if (success) {
+		error = refdb_fs_backend__write_tail(lock, backend, ref, flags, NULL, NULL, sig, message);
+		if (error == 0) {
+			error = loose_commit(lock, ref);
```
pks-t: Why exactly do we need this now and why wasn't it here before? I honestly don't quite know, and `loose_commit` is a rather weird function name that doesn't help

Edit: oh, it makes a bit more sense when one notices that things like `loose_lock` etc. also exist. I still am a bit puzzled, mainly because I don't know this code

src/refdb_fs.c
```diff
@@ -866,12 +875,20 @@ static int refdb_fs_backend__unlock(git_refdb_backend *backend, void *payload, i
...
+		}
+	}
+
+	git_filebuf_cleanup(lock);
```
pks-t: It's weird that `git_filebuf_cleanup` doesn't return anything. After all, it calls `p_close`, which may fail due to buffering semantics of file descriptors. I've certainly already seen it fail

src/refdb_fs.c
```diff
@@ -1269,37 +1286,48 @@ static int refdb_fs_backend__write(
...
-
-	if ((error = reference_path_available(backend, ref->name, NULL, force)) < 0)
-		return error;
-
-	/* We need to perform the reflog append and old value check under the ref's lock */
-	if ((error = loose_lock(&file, backend, ref->name)) < 0)
-		return error;
+	if ((error = refdb_fs_backend__write_tail(&file, _backend, ref, flags, old_id, old_target, who, message)) < 0) {
+		goto cleanup;
+	}
```
pks-t: Haha, you seem to have grown to like braces around single-line-statements :P

src/refdb_fs.c
```diff
@@ -844,19 +852,20 @@ static int refdb_fs_backend__lock(void **out, git_refdb_backend *_backend, const
 }
 
 static int refdb_fs_backend__write_tail(
+	git_filebuf *out_lock,
 	git_refdb_backend *_backend,
 	const git_reference *ref,
-	git_filebuf *file,
-	int update_reflog,
+	git__refdb_flags flags,
```
pks-t: As far as I know it's not allowed to use enums as bitmasks. You'd instead have to use e.g. an `unsigned int` here.

src/refdb_fs.c
pks-t: I'm always in favor of more tests :)

src/refdb_fs.c
```diff
@@ -1430,18 +1442,38 @@ static int refdb_fs_backend__delete(
...
+	const git_oid *old_id, const char *old_target)
+{
+	refdb_fs_backend *backend = (refdb_fs_backend *)_backend;
+	git_signature *who;
+	int error;
+
+	if ((error = git_reference__log_signature(&who, backend->repo)) < 0)
+		return error;
+
+	error = refdb_fs_backend__delete_impl(_backend, ref_name, who, NULL, old_id, old_target);
```
pks-t: Okay, so with this backwards-compatible change we'd be simply passing no message to the refdb deletion. Seems reasonable to me.

src/refdb_fs.c
```diff
@@ -1485,6 +1517,17 @@ static int refdb_fs_backend__delete_tail(
...
+	if ((flags & GIT_REFDB__SKIP_REFLOG) == 0) {
+		if ((error = should_write_reflog(&should_write, backend->repo, ref->name)) < 0)
+			goto cleanup;
+
+		if (should_write) {
+			const git_oid *old_id = git_reference_target(ref);
+			if ((error = maybe_append_head(backend, ref, old_id, NULL, who, message)) < 0)
+				goto cleanup;
+		}
+	}
```
pks-t: This hunk looks unrelated to what this commit is doing?

src/refdb_fs.c
```diff
@@ -1246,9 +1246,17 @@ static int maybe_append_head(
...
+	if ((error = git_reference_lookup(&head, backend->repo, GIT_HEAD_FILE))
+	    && error != GIT_ENOTFOUND)
 		return error;
 
+	/* Our repo is somehow missing its HEAD. Ignore the reflog in that case */
+	if (error == GIT_ENOTFOUND) {
+		/* XXX: warn ? */
+		git_error_clear();
+		return 0;
+	}
```
pks-t:
```
if ((error = git_reference_lookup(&head, backend->repo, GIT_HEAD_FILE)) < 0) {
    /* Our repo is somehow missing its HEAD. Ignore the reflog in that case */
    if (error == GIT_ENOTFOUND)
         error = 0;
    return error;
}
````

Comment (pks-t) [reply src/refdb_fs.c]: I think we also don't need to `git_error_clear()`. We state that errors are only valid immediately after a call to libgit2 that's returned an error.

File: src/refdb_fs.c (hunk 14 above)
Comment (tiennou) [reply src/refdb_fs.c]: Magic values ✨. I noticed, and it's used internally by the transaction code to signal that this is an unlock of a deleted reference (hence the deletion can proceed).

I'm not quite sure how to fix though… Options are :
- make `success` be a enum of what is asked of the backend.
- wrap the lock in a struct so it can keep track of why it was locked in the first place.

My preference goes to 2, as 1 looks like an abstraction leak.

File: src/refdb_fs.c (hunk 17 above)
Comment (tiennou) [reply src/refdb_fs.c]: Syntax is hard 😉. It's usually leftovers because of debugging code being present at some time.

File: src/refdb_fs.c (hunk 20 above)
Comment (tiennou) [reply src/refdb_fs.c]: Well, this is the part that actually writes to the reflog when deleting. That's taken from `write_tail`, minus the "write to our own reflog part" because we already deleted it.

I'll rewrite the commit message though, as it's not clear (must explain why, not what).

File: src/refdb_fs.c (hunk 21 above)
Comment (tiennou) [reply src/refdb_fs.c]: This is a workaround for the test case that explicitely removes `HEAD` and expect the refdb to keep working. This makes a missing `HEAD` be silently ignored (and the warn comment is there to remind me that I need to warn about that when warning support will have landed).

I'll use your version though.

File: src/refdb_fs.c (hunk 14 above)
Comment (pks-t) [reply src/refdb_fs.c]: The second option sounds good to me

File: tests/refs/reflog/messages.c [hunk 22]
```diff
@@ -423,3 +423,37 @@ void test_refs_reflog_messages__detaching_head_default_message(void)
...
+	int nentries, nentries_after;
+	git_buf msg = GIT_BUF_INIT;
+
+	cl_git_pass(git_repository_head(&head_ref, g_repo));
+
+	cl_git_pass(git_reference_resolve(&current_ref, head_ref));
+
+	nentries = reflog_entrycount(g_repo, GIT_HEAD_FILE);
+
+	cl_git_pass(git_branch_move(&renamed_ref, current_ref, "renamed", 1));
```
pks-t: I'd reorder a bit to make the intention clearer:

```
		nentries = reflog_entrycount(g_repo, GIT_HEAD_FILE);

		cl_git_pass(git_repository_head(&head_ref, g_repo));
		cl_git_pass(git_reference_resolve(&current_ref, head_ref));
		cl_git_pass(git_branch_move(&renamed_ref, current_ref, "renamed", 1));

		cl_assert_equal_i(nentries + 2, reflog_entrycount(g_repo, GIT_HEAD_FILE));
```

src/refs.c
```diff
@@ -675,27 +675,13 @@ static int reference__rename(git_reference **out, git_reference *ref, const char
...
 	if ((error = git_refdb_rename(out, ref->db, ref->name, normalized, force, signature, message)) < 0)
 		return error;
 
-	/* Update HEAD if it was pointing to the reference being renamed */
-	if (should_head_be_updated) {
-		error = git_repository_set_head(ref->db->repo, normalized);
-	} else {
-		rename_cb_data payload;
-		payload.old_name = ref->name;
-		memcpy(&payload.new_name, &normalized, sizeof(normalized));
```
pks-t: Just to get it straight, cause I'm not quite sure I understand from the commit message: previously, we have been updating HEAD in the refs code, so it was completely agnostic from the refdb backend. Now, we require every refdb backend to actually implement this by itself. Wouldn't this mean that we effectively break refdb backends that exist outside of our tree?

include/git2/sys/refdb_backend.h
```diff
@@ -251,4 +254,55 @@ GIT_EXTERN(int) git_refdb_set_backend(
 
 GIT_END_DECL
 
+typedef int GIT_CALLBACK(git_refdb_backend_del_v1)(git_refdb_backend *backend,
+					  const char *ref_name,
+					  const git_oid *old_id, const char *old_target);
+
+/* Old v1 struct version */
+struct git_refdb_backend_v1 {
```
pks-t: We still haven't decided how we actually want to proceed with such changes. Thing is, by renaming the old struct we're effectively also breaking old programs, as they cannot be compiled anymore (or worse, the compiler may just print an incompatible function pointer warning that goes unnoticed, and we run into strange behavior). This scenario also isn't helped by GIT_REFDB_BACKEND_VERSION, as a recompile would also get the newer version.

So maybe we need to instead introduce a new `git_refdb_backend_v2`?

PR #4317, opened 2017-07-23
Build with patched libcurl

ethomson:
This builds with a custom version of libcurl, which includes a fix for proxies and NTLM credentials.  It comes as quite a shock to curl 7.35, but you do not need to necessarily use proxy credentials when you are using a proxy with NTLM credentials (on the overall connection) and curl's connection pooling mechanism will segfault if you in fact do not.

This fix skips a `strcmp` against `NULL` proxy username/password:

```
--- curl-7.35.0.orig/lib/url.c
+++ curl-7.35.0/lib/url.c
@@ -3135,6 +3135,11 @@ ConnectionExists(struct SessionHandle *d

         /* Same for Proxy NTLM authentication */
         if(wantProxyNTLMhttp) {
+          /* Both check->http_proxy.user and check->http_proxy.passwd can be
+           * NULL */
+          if(!check->proxyuser || !check->proxypasswd)
+            continue;
+
           if(strcmp(needle->proxyuser, check->proxyuser) ||
              strcmp(needle->proxypasswd, check->proxypasswd))
             continue;
```

Regrettably, we need to move away from the container-based build system so that we can install custom packages that aren't actually part of the ubuntu distribution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@
 # see travis-ci.org for details
 
 language: c
+dist: trusty
 
 os:
   - linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,17 +20,8 @@ env:
   - OPTIONS="-DTHREADSAFE=ON -DCMAKE_BUILD_TYPE=Release"
   - OPTIONS="-DTHREADSAFE=OFF -DBUILD_EXAMPLES=ON"
 
-addons:
-  apt:
-    packages:
-    - cmake
-    - libssh2-1-dev
-    - openssh-client
-    - openssh-server
-    - valgrind
-
 dist: trusty
-sudo: false
+sudo: true
 
 matrix:
  fast_finish: true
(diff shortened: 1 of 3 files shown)


ethomson:
It's not ideal to move back to the VM-based build infra and away from the container-based system, but at least our unit tests should now be consistent.

@pks-t @carlosmn any thoughts?

pks-t:
I'm rather opposed to reverting to VM-based infra again. And luckily, we don't even have to do this, as the apt addon supports adding additional sources. E.g. like follows:

```
addons:
    apt:
        sources:
            - sourceline: 'deb https://repo.example.com/custom_curl'
              key_url: 'https://repo.example.com/gpg.key'
```

It forces us to create an actual repository, though. If you don't want to create the repo itself, I can do so, but I have no means to host these files afterwards.

ethomson:
Cool!  I looked for that but didn't see it.  I have an actual repository created at `http://libgit2deps.edwardthomson.com`.  I'll give this a shot.  Would definitely prefer to go back to the container style builds.

PR #4318, opened 2017-07-24
Add git_status_file_at

Uncommon:
This change takes `git_status_list_new` and splits out a new function called `git_status_list_new_at` (I'm totally open to alternative names) which takes a tree parameter to compare against instead of the HEAD tree.

The purpose of this is to get a preview of an amend commit, where the supplied tree will be from the head commit's first parent.

### DIFF
--- a/include/git2/status.h
+++ b/include/git2/status.h
@@ -173,12 +173,16 @@ typedef enum {
  * The `pathspec` is an array of path patterns to match (using
  * fnmatch-style matching), or just an array of paths to match exactly if
  * `GIT_STATUS_OPT_DISABLE_PATHSPEC_MATCH` is specified in the flags.
+ *
+ * The `baseline` is the tree to be used for comparison to the working directory
+ * and index; defaults to HEAD.
  */
 typedef struct {
 	unsigned int      version;
 	git_status_show_t show;
 	unsigned int      flags;
 	git_strarray      pathspec;
+	git_tree          *baseline;
 } git_status_options;
 
 #define GIT_STATUS_OPTIONS_VERSION 1
--- a/src/status.c
+++ b/src/status.c
@@ -280,12 +280,16 @@ int git_status_list_new(
 	if ((error = git_repository__ensure_not_bare(repo, "status")) < 0 ||
 		(error = git_repository_index(&index, repo)) < 0)
 		return error;
-
-	/* if there is no HEAD, that's okay - we'll make an empty iterator */
-	if ((error = git_repository_head_tree(&head, repo)) < 0) {
-		if (error != GIT_ENOTFOUND && error != GIT_EUNBORNBRANCH)
-			goto done;
-		giterr_clear();
+	
+	if (opts != NULL && opts->baseline != NULL) {
+		head = opts->baseline;
+	} else {
+		/* if there is no HEAD, that's okay - we'll make an empty iterator */
+		if ((error = git_repository_head_tree(&head, repo)) < 0) {
+			if (error != GIT_ENOTFOUND && error != GIT_EUNBORNBRANCH)
+				goto done;
+			giterr_clear();
+		}
 	}
 
 	/* refresh index from disk unless prevented */
@@ -377,7 +381,8 @@ done:
 
 	*out = status;
 
-	git_tree_free(head);
+	if (opts == NULL || opts->baseline != head)
+		git_tree_free(head);
 	git_index_free(index);
 
 	return error;
--- a/tests/status/worktree.c
+++ b/tests/status/worktree.c
@@ -1280,3 +1280,34 @@ void test_status_worktree__with_directory_in_pathlist(void)
 	git_status_list_free(statuslist);
 }
 
+void test_status_worktree__at_head_parent(void)
+{
+	git_repository *repo = cl_git_sandbox_init("empty_standard_repo");
+	git_status_options opts = GIT_STATUS_OPTIONS_INIT;
+	git_status_list *statuslist;
+	git_tree *parent_tree;
+	const git_status_entry *status;
+
+	cl_git_mkfile("empty_standard_repo/file1", "ping");
+	stage_and_commit(repo, "file1");
+
+	cl_git_pass(git_repository_head_tree(&parent_tree, repo));
+
+	cl_git_mkfile("empty_standard_repo/file2", "pong");
+	stage_and_commit(repo, "file2");
+
+	cl_git_rewritefile("empty_standard_repo/file2", "pyng");
+
+	opts.show = GIT_STATUS_SHOW_INDEX_AND_WORKDIR;
+	opts.baseline = parent_tree;
+	cl_git_pass(git_status_list_new(&statuslist, repo, &opts));
+
+	cl_assert_equal_sz(1, git_status_list_entrycount(statuslist));
+	status = git_status_byindex(statuslist, 0);
+	cl_assert(status != NULL);
+	cl_assert_equal_s("file2", status->index_to_workdir->old_file.path);
+	cl_assert_equal_i(GIT_STATUS_WT_MODIFIED | GIT_STATUS_INDEX_NEW, status->status);
+
+	git_tree_free(parent_tree);
+	git_status_list_free(statuslist);
+}


ethomson:
Thanks for the PR.  I think that I need to think a bit more about this, but let me give you my hot take:

I think that this should be an option in `git_status_options` instead of a new API.  Sorry - I know that you asked about this in slack, and I didn't give you an answer.  I wasn't sure at the time and didn't have much time to consider it (I was at a conference last week), but having seen it now, I do think that it would be preferable.

I say that especially since `status` is well-defined as a HEAD<->index<->workdir diff.  It seems very weird to call something `status` that _isn't_ actually looking at the `HEAD`.  It seems a little nicer to make this an option instead of a new API.  🤔

Uncommon:
If I make a change to GIT_STATUS_OPTIONS, do I need to bump the version number?

Uncommon:
Looks like AppVeyor failed because of credentials tests.

Uncommon:
I'd also like to do a similar thing with `git_status_file` which doesn't have a corresponding options parameter, so in that case I'd have to do a separate function (or add a parameter). I think it would be best to use the same approach for both`git_status_file` and `git_status_list_new`.

Uncommon:
Actually on further inspection I see that `git_status_file` uses `git_status_options` internally. So maybe the thing to do is to split `git_status_file` and extend `git_status_options`, leaving the other interfaces as they are.

Uncommon:
As noted above, I'm adding `git_status_file_at` and extending `git_status_options` with a `head_tree` field, and therefore taking `git_status_list_new_at` back out. I'll push it once I've done some more testing.

The question remains: if I'm changing `git_status_options`, does `GIT_STATUS_OPTIONS_VERSION` need to be bumped?

pks-t:
Hi @Uncommon, sorry for the long delay. Right now in the pre-1.0 world, we have no guaranted stable API and so all releases come with a SOVERSION-bump. So all users of our library will have link against the newer version anyway, making the version useless right now. We will start bumping option versions only after the first stable release.

Uncommon:
Ok, thanks!

Uncommon:
For some reason it's not showing up here yet, but I've pushed my new changes as described above: added the options field and `git_status_file_at` instead of `git_status_list_new_at`.

Uncommon:
Looks like I forgot to update the unit tests.

Uncommon:
Looks like AppVeyor failed with a Windows ssl problem.

Uncommon:
One of the Travis builds seems to have stalled out on `brew update`. How do you re-run it?

ethomson:
I'll requeue these builds.

Uncommon:
Thanks. AppVeyor errors look unrelated again.

ethomson:
Sigh.  No idea what's going on with the AppVeyor build.

Uncommon:
Did #4347 fix the build issues seen here?

pks-t:
Yup, exactly. Can you please rebase your changes onto my fix?

Uncommon:
I'm unable to build now. Apparently it's looking for the macOS 10.12 SDK, which I no longer have since upgrading to Xcode 9.

> CMake Warning at /usr/local/Cellar/cmake/3.5.2/share/cmake/Modules/Platform/Darwin-Initialize.cmake:132 (message):
>   Ignoring CMAKE_OSX_SYSROOT value:
> 
>    /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk
> 
>   because the directory does not exist.
> Call Stack (most recent call first):
>   /usr/local/Cellar/cmake/3.5.2/share/cmake/Modules/CMakeSystemSpecificInitialize.cmake:18 (include)
>   CMakeLists.txt:14 (PROJECT)
> 
> 
> CMake Error at /usr/local/Cellar/cmake/3.5.2/share/cmake/Modules/Platform/Darwin.cmake:76 (message):
>   CMAKE_OSX_DEPLOYMENT_TARGET is '10.8' but CMAKE_OSX_SYSROOT:
> 
>    ""
> 
>   is not set to a MacOSX SDK with a recognized version.  Either set
>   CMAKE_OSX_SYSROOT to a valid SDK or set CMAKE_OSX_DEPLOYMENT_TARGET to
>   empty.
> Call Stack (most recent call first):
>   /usr/local/Cellar/cmake/3.5.2/share/cmake/Modules/CMakeSystemSpecificInformation.cmake:36 (include)
>   CMakeLists.txt:14 (PROJECT)
> 
> 
> -- Configuring incomplete, errors occurred!
> See also "/Users/i58922/Developer/Personal/Xit/objective-git/External/libgit2/build/CMakeFiles/CMakeOutput.log".
> See also "/Users/i58922/Developer/Personal/Xit/objective-git/External/libgit2/build/CMakeFiles/CMakeError.log".
> make: *** [cmake_check_build_system] Error 1
>

ethomson:
@Uncommon did you re-run cmake?  It caches locations to things that are probably now missing.  If you have not, delete your `build` directory, create a new one, and run `cmake` again.

Uncommon:
I also had to delete tests/clar.suite. Thanks to @pks-t for that pointer.

Uncommon:
Rebased, and tests are passing for me locally.

Uncommon:
Poke again. Builds are green. I have a feature waiting on this, so I'd like to get it in.

include/git2/status.h
```diff
@@ -294,6 +299,31 @@ GIT_EXTERN(int) git_status_file(
...
+ * for comparing to the indxex.
+ * @return 0 on success, GIT_ENOTFOUND if the file is not found in the
+ *      specified tree, index, and work tree, GIT_EAMBIGUOUS if `path` matches
+ *      multiple files or if it refers to a folder, and -1 on other errors.
+ */
+GIT_EXTERN(int) git_status_file_at(
+	unsigned int *status_flags,
+	git_repository *repo,
+	const char *path,
+	git_tree *head_tree);
```
pks-t: So one problem with this new API is, as @ethomson already mentioned, that we still have to add new calls whenever we want some new option in here. Ideally, we'd have designed `git_status_file` to just take a `git_status_options` structure here, so that we can add more options later on without adding new functions for all of them.

As we do not want to break the API here, we obviously cannot change `git_status_file` right now. What we _can_ do though is to add a `git_status_file_ext(unsigned int *, git_repository *, const char *, const git_status_options *)`, which fixes that shortcoming. See for example `git_status_foreach` and `git_status_foreach_ext`, where we fixed the same shortcoming.

So I'd prefer to introduce that one instead of having to potentially maintain more `git_status_file_foobar` functions in the future.

src/status.c
```diff
@@ -280,14 +280,19 @@ int git_status_list_new(
...
-	if ((error = git_repository_head_tree(&head, repo)) < 0) {
-		if (error != GIT_ENOTFOUND && error != GIT_EUNBORNBRANCH)
-			goto done;
-		giterr_clear();
+	
+	if (opts != NULL && opts->head_tree != NULL) {
+		head = opts->head_tree;
 	}
-
+	else {
```
pks-t: We usually have closing braces and else branches on the same line.

tests/status/worktree.c
```diff
@@ -1280,3 +1280,33 @@ void test_status_worktree__with_directory_in_pathlist(void)
 	git_status_list_free(statuslist);
 }
 
+void test_status_worktree__at_head_parent(void)
+{
+  git_repository *repo = cl_git_sandbox_init("empty_standard_repo");
+  git_status_options opts = GIT_STATUS_OPTIONS_INIT;
+  git_status_list *statuslist;
+  git_tree *parentTree;
```
pks-t: Typically, we do not use camelcase variable names but instead use underscores.

tests/status/worktree.c
```diff
@@ -1280,3 +1280,33 @@ void test_status_worktree__with_directory_in_pathlist(void)
...
+  const git_status_entry *status;
+  
+  cl_git_mkfile("empty_standard_repo/file1", "ping");
+  stage_and_commit(repo, "file1");
+  
+  git_repository_head_tree(&parentTree, repo);
```
pks-t: Please wrap this in an `cl_git_pass`.

tests/status/worktree.c
```diff
@@ -1280,3 +1280,33 @@ void test_status_worktree__with_directory_in_pathlist(void)
...
+  
+  cl_git_mkfile("empty_standard_repo/file2", "pong");
+  stage_and_commit(repo, "file2");
+
+  cl_git_rewritefile("empty_standard_repo/file2", "pyng");
+  
+  opts.show = GIT_STATUS_SHOW_INDEX_AND_WORKDIR;
+  opts.head_tree = parentTree;
+  git_status_list_new(&statuslist, repo, &opts);
```
pks-t: Same, please use `cl_git_pass`

tests/status/worktree.c
```diff
@@ -1280,3 +1280,33 @@ void test_status_worktree__with_directory_in_pathlist(void)
...
+  
+  cl_assert_equal_sz(1, git_status_list_entrycount(statuslist));
+  status = git_status_byindex(statuslist, 0);
```
pks-t: Please wrap this in an `cl_assert`, so that we do not segfault on a `NULL` pointer here.

include/git2/status.h
ethomson: Actually, I was suggesting not adding a new `git_status_file` API _at all_.  I was hoping that the change in the options struct would be enough.

`git_status_file` is a helper method.  It should be for the 99% use cases.  If you want something special (like a different `HEAD` tree, which is _super_ uncommon) then I don't mind saying that you have to do a little more work to get there.

In other words, I am not in favor of adding new `git_status_file`-like functions, and would encourage you to just call `git_status_foreach_ext` (with the new opts structure) for this very special use case.

Uncommon: So you think I should take the login from `git_status_file_at` and move it up into my app? Either way is fine with me.

Uncommon:
Updated with a little bug fix. Coming soon, I'll remove `git_status_file_at`, but I'll still need `head_tree` in the options.

Uncommon:
@pks-t Requested changes to the unit test code have been addressed. I've removed `git_status_file_at`, leaving only the `head` field in the options.

src/status.c
Uncommon: Oops, I missed this in my first round of fixes. Looking at other files, I don't know if "usually" is accurate :) but if it's preferred I can certainly make the change.

Uncommon: I went ahead and did this one too.

Uncommon:
Any further thoughts on this?

pks-t:
Sorry, I'm currently busy with a move. I'll find some time next week to have a further look

Uncommon:
It's been a couple of weeks, so I'm bumping this again.

include/git2/status.h
```diff
@@ -311,6 +311,24 @@ GIT_EXTERN(int) git_status_list_new(
...
+ * @param repo Repository object
+ * @param opts Status options structure
+ * @param head Tree to compare the index to
+ * @return 0 on success or error code
+ */
+GIT_EXTERN(int) git_status_list_new_at(
+  git_status_list **out,
+  git_repository *repo,
+  const git_status_options *opts,
+  git_tree *head);
```
pks-t: Can't the `head` argument also be `const`?

src/status.c
```diff
@@ -261,12 +261,33 @@ int git_status_list_new(
...
+    if (error != GIT_ENOTFOUND && error != GIT_EUNBORNBRANCH)
+      goto done;
+    giterr_clear();
+  }
+  error = git_status_list_new_at(out, repo, opts, head);
+  
+done:
+  git_tree_free(head);
+  return error;
+}
```
pks-t: Same here. Please try to match existing indentation (tabs instead of spaces).

include/git2/status.h
```diff
@@ -311,6 +311,24 @@ GIT_EXTERN(int) git_status_list_new(
...
+/**
+ * Gather file status information compared to a specified tree.
+ *
+ * Similar to `git_status_list_new`, but instead of using HEAD it allows
+ * another target tree to be specified.
+ *
+ * @param out Pointer to store the status results in
+ * @param repo Repository object
+ * @param opts Status options structure
+ * @param head Tree to compare the index to
```
pks-t: Please specify that `head` can actually be `NULL` (behaving as an empty iterator then).

src/status.c
```diff
@@ -261,12 +261,33 @@ int git_status_list_new(
...
+
+int git_status_list_new_at(
+  git_status_list **out,
+  git_repository *repo,
+  const git_status_options *opts,
+  git_tree *head)
```
pks-t: You have a memory management issue here, as you try to `git_tree_free(head)` in the end. You should just make `head` const and have the caller manage memory for you.

tests/status/worktree.c
pks-t: This still applies

tests/status/worktree.c
pks-t: This also still applies :)

tests/status/worktree.c
```diff
@@ -1280,3 +1280,32 @@ void test_status_worktree__with_directory_in_pathlist(void)
...
+  
+  git_repository_head_tree(&parentTree, repo);
+  
+  cl_git_mkfile("empty_standard_repo/file2", "pong");
+  stage_and_commit(repo, "file2");
+
+  cl_git_rewritefile("empty_standard_repo/file2", "pyng");
+  
+  opts.show = GIT_STATUS_SHOW_INDEX_AND_WORKDIR;
+  git_status_list_new_at(&statuslist, repo, &opts, parentTree);
```
pks-t: Please wrap this in an cl_assert, so that we do not segfault on a NULL pointer here.

include/git2/status.h
```diff
@@ -311,6 +311,24 @@ GIT_EXTERN(int) git_status_list_new(
 	git_repository *repo,
 	const git_status_options *opts);
 
+/**
```
pks-t: Please make the commit message conform to our preferred style, e.g.

```
status: add function to create status comparing to a specific tree

In here, you describe why that functionality is needed and not yet
implemented by any of our existing code.
```

include/git2/status.h
```diff
@@ -173,12 +173,17 @@ typedef enum {
  * The `pathspec` is an array of path patterns to match (using
  * fnmatch-style matching), or just an array of paths to match exactly if
  * `GIT_STATUS_OPT_DISABLE_PATHSPEC_MATCH` is specified in the flags.
+ *
+ * The `head_tree` is an optional tree to be used instaed of the head commit's
```
pks-t: `head_tree` is a bit misnamed as you're not trying to diff against a specific `head`. I have no real better idea right now, but at least remove the `head_` prefix. There's also a typo in "instead". Last, there is no need to specify that the tree is optional, as all options in our options arrays are typically optional.

include/git2/status.h
pks-t: Well, if you have `git_status_foreach_ext` handle the new tree option then most of it is already done for you, I guess. But otherwise yes

src/status.c
pks-t: Well, sure. We're being inconsistent with many things, but at least I try to avoid introducing more inconsistencies ;) And in this case we're definitly being consistent most of the time with closing brace and else on the same line.

This still seems to not be fixed here -- did you forget to push your amends?

include/git2/status.h
```diff
@@ -311,6 +311,24 @@ GIT_EXTERN(int) git_status_list_new(
...
+ * another target tree to be specified.
+ *
+ * @param out Pointer to store the status results in
+ * @param repo Repository object
+ * @param opts Status options structure
+ * @param head Tree to compare the index to
+ * @return 0 on success or error code
+ */
+GIT_EXTERN(int) git_status_list_new_at(
+  git_status_list **out,
```
pks-t: I'm a little hesitant regarding this function's name. `at` doesn't really tell the user anything about what it should be at, actually. An option would be `git_status_list_new_with_tree`, for example.

include/git2/status.h
```diff
@@ -173,12 +173,17 @@ typedef enum {
  * The `pathspec` is an array of path patterns to match (using
```
pks-t: Pleas mind the preferred commit message style. First again, the area-prefix ("status" in this case), and then please also mind that we wrap the body at 80 characters. Also, please try to make the message more imperative instead arguing in place of the project, not yourself.

tests/status/worktree.c
```diff
@@ -1299,7 +1299,8 @@ void test_status_worktree__at_head_parent(void)
   cl_git_rewritefile("empty_standard_repo/file2", "pyng");
```
pks-t: Same regaring the commit message. Furthermore, it feels like this should be squashed in to the initial commit where the test has been created with `git_status_list_new` taking the option?

include/git2/status.h
```diff
@@ -298,31 +298,6 @@ GIT_EXTERN(int) git_status_file(
 	git_repository *repo,
 	const char *path);
 
-/**
```
pks-t: Please squash this commit into the initial commit where you created that function

src/status.c
```diff
@@ -283,8 +283,7 @@ int git_status_list_new(
 	
 	if (opts != NULL && opts->head_tree != NULL) {
 		head = opts->head_tree;
-	}
-	else {
+	} else {
```
pks-t: Please squash that commit into the original commit where you created that line

Uncommon:
Updated and squashed. Thanks for the notes.

BTW where is the commit message style documented? I don't see it in CONTRIBUTING or CONVENTIONS.

Uncommon:
Oops, forgot to update the test.

pks-t:
Thanks for yout update!

Good hint regarding the commit messages. It's mostly a thing I try to personally point at where reasonable, the others are more lenient on that side. I'll document that and review your changes coming Friday (hopefully *sigh*)

src/status.c
```diff
@@ -280,14 +280,18 @@ int git_status_list_new(
...
+		/* if there is no HEAD, that's okay - we'll make an empty iterator */
+		if ((error = git_repository_head_tree(&head, repo)) < 0) {
+			if (error != GIT_ENOTFOUND && error != GIT_EUNBORNBRANCH)
+				goto done;
+			giterr_clear();
+		}
 	}
-
+
```
pks-t: There's trailing whitespace here

include/git2/status.h
pks-t: I've now created PR #4422 to document our style. There's still two minor violations:

1. the first letter after the area-prefix is usually lowercased
2. the subject is some characters too long

Proposed: "status: add option to compare to trees other than head". You could add some reasoning in the commit message body, but that's only optional.

tests/status/worktree.c
```diff
@@ -1280,3 +1280,34 @@ void test_status_worktree__with_directory_in_pathlist(void)
...
+  
+  cl_assert_equal_sz(1, git_status_list_entrycount(statuslist));
+  status = git_status_byindex(statuslist, 0);
+  cl_assert(status != NULL);
+  cl_assert_equal_s("file2", status->index_to_workdir->old_file.path);
+  cl_assert_equal_i(GIT_STATUS_WT_MODIFIED | GIT_STATUS_INDEX_NEW, status->status);
+  
+  git_tree_free(parent_tree);
+  git_status_list_free(statuslist);
+}
```
pks-t: You are using spaces instead of tabs here

include/git2/status.h
```diff
@@ -173,12 +173,17 @@ typedef enum {
  * The `pathspec` is an array of path patterns to match (using
  * fnmatch-style matching), or just an array of paths to match exactly if
  * `GIT_STATUS_OPT_DISABLE_PATHSPEC_MATCH` is specified in the flags.
+ *
+ * The `tree` is a tree to be used instead of the head commit's tree for
+ * comparing with the index. For example, this can be useful when preparing
+ * to do an amend commit.
```
ethomson: I don't think that the discussion of an amend commit makes sense in the context of this documentation.  You should also explicitly indicate that it defaults to `HEAD`.   (I know that you're implicitly stating it).  Something like:

> The `tree` is the tree to be used for comparison to the working directory and index; defaults to HEAD.

include/git2/status.h
```diff
@@ -173,12 +173,17 @@ typedef enum {
...
  */
 typedef struct {
 	unsigned int      version;
 	git_status_show_t show;
 	unsigned int      flags;
 	git_strarray      pathspec;
+	git_tree          *tree;
```
ethomson: We use the term `baseline` for the overrideable `HEAD` in `git_checkout`.  I wonder if `baseline` would be more consistent here, too?

ethomson:
Yeah, I think this approach is great.  Thanks for all your work on this @Uncommon.  I might tweak some names for consistency with some other APIs but otherwise this is looking good.

Uncommon:
Fixed and rebased. Thanks for the reviews.

pks-t:
Yeah, I think `baseline` is perfect, thanks for that.

You still forgot to fix up whitespaces for your test. As soon as that is amended I'll be happy to merge.

Uncommon:
Oops, forgot about the tabs. Thanks.

ethomson:
AWESOME.  Thanks for this!  ✨

PR #4319, opened 2017-07-25
 cmake --build .. instead of  cmake --build .

swuecho:
two dot

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ On most systems you can build the library using the following commands
 
 	$ mkdir build && cd build
 	$ cmake ..
-	$ cmake --build .
+	$ cmake --build ..
 
 Alternatively you can point the CMake GUI tool to the CMakeLists.txt file and generate platform specific build project or IDE workspace.
 


ethomson:
I think that the original documentation is correct.  Are you having trouble?

swuecho:
$ mkdir build && cd build
$ cmake ..
$ cmake --build .

after cmake .., still in build dir

tiennou:
Results here :

```
$ mkdir build && cd build
$ cmake ..
[output]
$ cmake --build ..
Error: could not load cache
```

swuecho:
```
➜  build cmake ..
-- Configuring done
-- Generating done
-- Build files have been written to: /home/hwu/Download/libgit2-0.26.0
➜  build cmake --build .
Error: could not load cache
➜  build cmake --build ..
[ 21%] Built target git2
[100%] Built target libgit2_clar
```

probably does not matter. should be easy for anyone to figure out.

tiennou:
You might have run `cmake` from the root of the source tree once, hence there's a `CMakeCache.txt` file & `CMakeFiles` directory and running `cmake --build ..` from `build` finds that ?

ethomson:
Yes, that's a good suggestion - if you ever run `cmake` from the source tree's root, `cmake` will try to re-update those files, instead of in the new place.

Curious if you have the same behavior in a fresh clone?  (Or something where you've `git clean`ed everything.)

PR #4320, opened 2017-07-28
tests: rebase::submodule: verify initialization method calls

pks-t:
Some return codes for functions which may fail are not being checked in
`test_rebase_submodule__initialize`. This may lead us to not notice
errors when initializing the environment and would possibly result in
either memory corruption or segfaults as soon as any of the
initialization steps fails. Fix this by wrapping these function calls
into `cl_git_pass`.

### DIFF
--- a/tests/rebase/submodule.c
+++ b/tests/rebase/submodule.c
@@ -28,25 +28,25 @@ void test_rebase_submodule__initialize(void)
 
 	rewrite_gitmodules(git_repository_workdir(repo));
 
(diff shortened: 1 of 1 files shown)


tiennou:
Oops, my bad. I didn't get you were talking error return 😟 .

ethomson:
Thanks for the careful eyes, @pks-t

pks-t:
@tiennou No problem at all, thanks for your changes!

@ethomson Heh, "careful eyes" being Coverity here.

PR #4321, opened 2017-07-28
travis: add custom apt sources

ethomson:
This basically reverts #4317's move back to Travis's VM infrastructure, and points to our custom repository.


.travis.yml
```diff
@@ -2,7 +2,6 @@
 # see travis-ci.org for details
 
 language: c
-dist: trusty
```
ethomson: Didn't notice that I had duplicated this line in my original PR.

ethomson:
Wheeeee:

```
$ curl -sSL "https://pgp.mit.edu/pks/lookup?op=get&search=0x5656187599131CD5" | sudo -E apt-key add -
OK
$ echo "deb http://libgit2deps.edwardthomson.com trusty libgit2deps" | sudo tee -a /etc/apt/sources.list > /dev/null
```

This is exactly what our prior script was doing.  And it indeed downloads our custom packages:

```
Get:4 http://libgit2deps.edwardthomson.com trusty/libgit2deps amd64 libcurl4-gnutls-dev amd64 7.35.0-1ubuntu2.11 [331 kB]
Get:6 http://libgit2deps.edwardthomson.com trusty/libgit2deps amd64 libcurl3-gnutls amd64 7.35.0-1ubuntu2.11 [256 kB]
Get:7 http://libgit2deps.edwardthomson.com trusty/libgit2deps amd64 curl amd64 7.35.0-1ubuntu2.11 [206 kB]
Get:8 http://libgit2deps.edwardthomson.com trusty/libgit2deps amd64 libcurl3 amd64 7.35.0-1ubuntu2.11 [263 kB]
```

pks-t:
Glad it works and nice to see no more issues with our proxy code, thanks for fixing!

It's quite unfortunate that we're effectively shipping broken code on some platforms and I think we should bug distro maintainers to update the package accordingly. I've just checked with the Ubuntu package, and they are already patching this exact code that is causing our issue as part of CVE-2016-0755. I guess they have simply backported the upstream patch without having the additional patch to fix the check. I'll take care of this issue.

ethomson:
Well.  It works for trusty.  But - unsurprisingly - it makes our precise build fail miserably.

We could just dump our precise build entirely or move back to the VM infra until Ubuntu backports that fix.  (If they backport the fix.)

Dumping precise would be my preference.  Thoughts @pks-t ?

ethomson:
Sorry, there's also a question of if there's a conditional here around apt sources that I'm missing.  But I don't see it.  :/

pks-t:
I've created a bug with Ubuntu at https://bugs.launchpad.net/ubuntu/+source/curl/+bug/1707214.

By the way, are you sure this even works (I cannot get Travis to load right now, cannot check myself)? I've just found `https://github.com/travis-ci/apt-source-whitelist`, which looks like it would filter out your custom source. If so, we should keep the current state and wait until the bug has been resolved by Canonical.

ethomson:
Thanks for opening that ubuntu issue.

Indeed, this does work (though, as mentioned above, for trusty only).  Those lines I pasted above are from the build.

My question was whether it has silently updated us to the VM based infrastructure in order to support this.  (The sources configuration is notably not in the container section of the documentation.)  I just added a `sudo` command to our build, which failed with:

> This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid, and setgid executables.
> If you require sudo, add 'sudo: required' to your .travis.yml

So it does appear that we are both running on the container infrastructure _and_ installing our own custom apt packages.

pks-t:
We might also wait a few days to see whether libcurl3 will get fixed quickly by Canonical (which I somehow doubt). If not, I'd vote for dropping Precise.

ethomson:
👍

ethomson:
We've dropped support for precise, so we can safely move back to the container-based builds on Travis.

PR #4324, opened 2017-07-31
Use find_package instead of pkg-config on Windows for libssh2

jminnie:
Hi,

I hope you will find this change useful.

I was having trouble configuring the build when I enable the "USE_SSH" option. To find the libssh2 installation, pkg-config is used. On Linux this is not uncommon, but on Windows, having pkg-config available is highly unlikely and non-standard.

Since the libssh2 maintainers already provide a CMake config file for their software, it makes the usage of "find_package" a trivial exercise.


CMakeLists.txt
```diff
@@ -358,24 +358,60 @@ ENDIF()
...
+		# pkg-config is not the ideal way to find dependencies on Windows. Use the standard
+		# CMake way of find_package on Windows, or if pkg-config is not available on
+		# non-Windows operating systems.
+		FIND_PACKAGE(libssh2)
+		
+		IF (LIBSSH2_FOUND)
+			ADD_DEFINITIONS(-DGIT_SSH)
+			
+			# TODO: Determine the contents of LIBSSH2_LDFLAGS.
+			#LIST(APPEND LIBGIT2_PC_LIBS ${LIBSSH2_LDFLAGS})
```
jminnie: This is the only change I am unclear about. I am not sure how to obtain the linker flags that are supposed to be in LIBSSH2_LDFLAGS, or if it is even necessary.

pks-t: If I'm not mistaken, the target returned by `find_package` uses absolute paths for the library. If so, there should be no need for setting LDFLAGS, but I wasn't able to find it in CMake's documentation.

CMakeLists.txt
```diff
@@ -358,24 +358,60 @@ ENDIF()
...
-	INCLUDE_DIRECTORIES(${LIBSSH2_INCLUDE_DIRS})
-	LINK_DIRECTORIES(${LIBSSH2_LIBRARY_DIRS})
-	LIST(APPEND LIBGIT2_PC_LIBS ${LIBSSH2_LDFLAGS})
-	#SET(LIBGIT2_PC_LIBS "${LIBGIT2_PC_LIBS} ${LIBSSH2_LDFLAGS}")
-	SET(SSH_LIBRARIES ${LIBSSH2_LIBRARIES})
-
-	CHECK_LIBRARY_EXISTS("${LIBSSH2_LIBRARIES}" libssh2_userauth_publickey_frommemory "${LIBSSH2_LIBRARY_DIRS}" HAVE_LIBSSH2_MEMORY_CREDENTIALS)
-	IF (HAVE_LIBSSH2_MEMORY_CREDENTIALS)
-		ADD_DEFINITIONS(-DGIT_SSH_MEMORY_CREDENTIALS)
+	IF (PKG_CONFIG_EXECUTABLE_FOUND AND NOT WIN32)
```
pks-t: I think it would make even more sense to not have both methods mutually exclusive. If `pkg-config` was found, we could still try `FIND_PACKAGE` if `PKG_CHECK_MODULES` has failed.

jminnie: Sure, that sounds like a good idea. The CMake find_package will be a fall-back method if the pkg-config search failed.

CMakeLists.txt
jminnie: Yeah, based on my experience building with this line commented out, it seems that all necessary information was obtained from CMake and that this line is now redundant.

jminnie:
Hi Patrick, thanks for taking the time to look at my changes. I will make the changes that you suggested. There is no rush in terms of merging. I look forward to hearing from you once #4282 is completed. Thanks!

pks-t:
#4282 has landed now, so you can actually continue working on this. I'll review as soon as you find time to rebase and fix the conflicts :)

jminnie:
Hi Patrick, as you can see I rebased onto latest master and made some minor tweaks to my original implementation. Your automated tests are now all passing, but I now have a slight issue with the include path to libssh2 when I build in Visual Studio. I will have to take another look, and will let you know when this is ready for a closer review. Thanks!

src/CMakeLists.txt
```diff
@@ -217,22 +217,55 @@ ENDIF()
...
+		LIST(APPEND LIBGIT2_PC_LIBS ${LIBSSH2_LDFLAGS})
+		#SET(LIBGIT2_PC_LIBS "${LIBGIT2_PC_LIBS} ${LIBSSH2_LDFLAGS}")
+
+		CHECK_LIBRARY_EXISTS("${LIBSSH2_LIBRARIES}" libssh2_userauth_publickey_frommemory "${LIBSSH2_LIBRARY_DIRS}" HAVE_LIBSSH2_MEMORY_CREDENTIALS)
+		IF (HAVE_LIBSSH2_MEMORY_CREDENTIALS)
+			SET(GIT_SSH_MEMORY_CREDENTIALS 1)
+		ENDIF()
+    ENDIF()
+
+    IF(NOT LIBSSH2_FOUND)
```
pks-t: There's a space missing in front of the first brace.

src/CMakeLists.txt
```diff
@@ -217,22 +217,55 @@ ENDIF()
...
-	LIST(APPEND LIBGIT2_INCLUDES ${LIBSSH2_INCLUDE_DIRS})
-	LIST(APPEND LIBGIT2_LIBS ${LIBSSH2_LIBRARIES})
-	LIST(APPEND LIBGIT2_LIBDIRS ${LIBSSH2_LIBRARY_DIRS})
-	LIST(APPEND LIBGIT2_PC_LIBS ${LIBSSH2_LDFLAGS})
-	#SET(LIBGIT2_PC_LIBS "${LIBGIT2_PC_LIBS} ${LIBSSH2_LDFLAGS}")
-
-	CHECK_LIBRARY_EXISTS("${LIBSSH2_LIBRARIES}" libssh2_userauth_publickey_frommemory "${LIBSSH2_LIBRARY_DIRS}" HAVE_LIBSSH2_MEMORY_CREDENTIALS)
-	IF (HAVE_LIBSSH2_MEMORY_CREDENTIALS)
-		SET(GIT_SSH_MEMORY_CREDENTIALS 1)
+    MESSAGE(STATUS "Trying pkg-config configuration files to locate libssh2.")
```
pks-t: I'd drop those messages. The developer usually does not care whether the package is found via pkg-config or CMake module, but only cares about that we _have_ found the package.

src/CMakeLists.txt
```diff
@@ -217,22 +217,55 @@ ENDIF()
...
+			SET(SSH_LIBRARIES Libssh2::libssh2)
+			
+			# Get the location of the libssh2 headers.
+			GET_PROPERTY(LIBSSH2_INCLUDE_DIRS
+						 TARGET ${SSH_LIBRARIES}
+			             PROPERTY INTERFACE_INCLUDE_DIRECTORIES)
+
+			# Make sure we have access to the symbol.
+			LIST(APPEND CMAKE_REQUIRED_LIBRARIES ${SSH_LIBRARIES})
+			LIST(APPEND CMAKE_REQUIRED_INCLUDES "${LIBSSH2_INCLUDE_DIRS}")
```
pks-t: You should also use the LIBGIT2_* variables here, as for the other case.

jminnie:
Hi, thanks for taking the time to review my changes. I am not sure when I will be able to get back to this. Feel free to close this pull request. I may re-submit in the future when I get some time again. Sorry for the wasted time.

pks-t:
No problem, just take your time. We'll not close the PR but keep it open in the meantime until you've found some time to make the changes. If you feel like you won't update it in the future, though, please just close the PR yourself :)

ethomson:
This was (unintentionally) closed as part of renaming the `master` branch to `main`.

_However_ I regret that it's also very out of date.  I'm not familiar enough to know whether the changes that we've made to cmake for libssh2 render this obsolete, or if it is still useful.  I'm going to leave it closed, but would welcome changes to the current cmake configuration if it is deemed useful.

Issue #3786, opened 2016-05-13
Cannot resolve https addresses (Ubuntu 16.04)


anselg:
When I try to clone from https addresses, libgit2 fails. The message I get when I print out the git_error struct is: 
```
12: Failed to resolve address for https: Name or service not known
```

I'm getting this error when I use the system libgit2 package (Ubuntu 16.04) and when I build libgit2 from source (HEAD of master). I saw https://github.com/libgit2/pygit2/issues/573 and make sure I had libssl-dev installed.

It seems that the library isn't building with openssl properly. When I print out the GIT_FEATURE_THREADS information, I'm only getting the THREADS and SSH features. What's confusing is that when I look through the cmake output, it shows that it found libssl. 
```
-- The C compiler identification is GNU 5.3.1
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.1") 
-- Performing Test HAVE_STRUCT_STAT_ST_MTIM
-- Performing Test HAVE_STRUCT_STAT_ST_MTIM - Success
-- Performing Test HAVE_STRUCT_STAT_ST_MTIMESPEC
-- Performing Test HAVE_STRUCT_STAT_ST_MTIMESPEC - Failed
-- Performing Test HAVE_STRUCT_STAT_MTIME_NSEC
-- Performing Test HAVE_STRUCT_STAT_MTIME_NSEC - Failed
-- Performing Test HAVE_STRUCT_STAT_NSEC
-- Performing Test HAVE_STRUCT_STAT_NSEC - Success
-- Checking for module 'libcurl'
--   Found libcurl, version 7.47.0
-- Found OpenSSL: /usr/lib/i386-linux-gnu/libssl.so;/usr/lib/i386-linux-gnu/libcrypto.so (found version "1.0.2g") 
-- Found HTTP_Parser: /usr/include  
-- Found ZLIB: /usr/lib/i386-linux-gnu/libz.so (found version "1.2.8") 
-- Checking for module 'libssh2'
--   Found libssh2, version 1.5.0
-- Looking for libssh2_userauth_publickey_frommemory in ssh2
-- Looking for libssh2_userauth_publickey_frommemory in ssh2 - not found
[...]

PR #4325, opened 2017-08-01
Add -DUSE_CURL_SSL to allow distros to avoid OpenSSL

infinity0:
This is helpful for distros that don't want to link libgit2 against OpenSSL for licensing reasons, but are able to link against another TLS-enabled version of libcurl. For example see [Debian #798421](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=798421).


infinity0:
Since Ubuntu takes their packages from Debian, this would also prevent future scenarios like #3786 and https://github.com/DronMDF/vanadis/issues/84 and https://github.com/saltstack/salt/issues/35993.

Tested against cargo 0.20.

ethomson:
We explicitly removed curl's TLS support because it does not provide us the ability to get access to the certificates.  We must be able to give users the certificate information to be able to make informed decisions about whether they should continue to negotiate with untrusted certificates (prompt to continue, etc).

See the discussion in #3183, which is the PR that includes the commit that you're proposing be reverted.  @carlosmn should be able to confirm my remembering.

If that's correct, and we cannot get certificate information from libcurl, then this is not likely to be something that we can include.  Perhaps we should instead be working to complete the mbedTLS support, per #4173?

infinity0:
Personally I'd be happy with if cargo or any other program that uses libgit2 simply hard-errors with some appropriately-detailed error message when meeting an untrusted certificate, rather than giving me the option to carry on. What sorts of detailed information about the certificate are you looking for - do you have an example of an existing UI that does this prompting, for example?

mbedTLS support would be very nice indeed. Unfortunately in the meantime this patch is probably the only feasible option forward for us in Debian, the alternative is to maintain two libriaries libgit2-without-ssl and libgit2-with-openssl which is much more effort... I would be verify happy to amend this patch though. Does it currently not hard-error on an untrusted certificate? (I have not yet tested it on different servers.)

ethomson:
> What sorts of detailed information about the certificate are you looking for - do you have an example of an existing UI that does this prompting, for example?

Yes, this is supported functionality that we cannot remove.

infinity0:
OK, but these patches are guarded by a CMake flag `USE_CURL_SSL` which defaults to off. People that need this functionality you're talking about, would still have this functionality even after this patch is accepted. And people using the Debian and Ubuntu packages already don't have this functionality, since OpenSSL support is not enabled in the first place.

infinity0:
I verified that this patch will cause a hard error if curl sees an invalid certificate. So I don't think this opens up any security problems.

In the absence of this patch, distributions like Debian are *already unable* to provide extra details in the UI about untrusted certificates since OpenSSL is not linked into libgit2 in the first place. So this patch does not reduce any existing functionality. Therefore, could you please consider merging it?

onovy:
I agree with infinity0. This bug blocks me for updating python-pygit2 in Debian, because test case needs https support.

Accepting self-signed (or trusting) certificate should be done on system level, not in libgit2 or any other library. libgit2 should not bother about certificates at all. It's TLS library / system problem.

Without this patch we will have libgit+python-pygit2 etc. in Debian/Ubuntu without https support, which is really crazy. And OpenSSL is no-way because of licensing reason.

Thanks again for considering.

ethomson:
> Accepting self-signed (or trusting) certificate should be done on system level, not in libgit2 or any other library. libgit2 should not bother about certificates at all. It's TLS library / system problem.

You're right, obviously it should not be done at the level of the library.  The library simply acts as an intermediary so that an application can decide whether or not to trust a certificate that would otherwise be untrusted.

Requiring users to add all certs to their system certificate store is not acceptable for many applications, especially interactive ones.  And although your application(s) may not require this functionality, it is functionality that many people rely upon.  This is why the SSL backends support it.

Expanding the number of SSL backends we support is not ideal:  this creates a larger support burden for us.  But if we're going to expand the support matrix to include other backends, then they should at least be fully functional, and have somebody who will support it moving forward.

At this moment, libcurl doesn't meet those requirements for us.

infinity0:
I think half-supporting openssl is better than not supporting HTTPS at all, and at the very least you'll stop getting bug reports from Debian and Ubuntu users who get confused by the lack of HTTPS support in the distribution's own libgit2. I imagine there are much fewer users who would send bug reports about "can't force-trust an untrusted certificate". (And this would already happen in the current situation, without this patch applied.)

I can appreciate it's frustrating getting bug reports from users of a version of libgit2 you didn't control the "final edits" to, but this is just a fact of reality of doing open source development and one has to get used to it. I don't like having to spend my time working around these sorts of annoying licensing issues, either.

@onovy In the meantime you can test python-pygit2 Debian with this patch applied, by building them against the libgit2 packages here: https://people.debian.org/~infinity0/apt/pool/main/libg/libgit2/

ethomson:
I'd rather fully support a (very) few crypto systems as well as possible than keep any half supported.  This isn't like the Amiga port that half works.  If none of the maintainers pay attention to Amiga, nobody gets pwned.  If some APIs don't work on the Amiga, nobody's repository gets leaked.

Doing a half-assed job at a crypto implementation is a guarantee that we're going to screw something up that's important.

This isn't a PR that is a complete TLS implementation that's going to get care and feeding from you moving forward.  TLS is simply too important to not have somebody who is continuing to maintain its support.

As it stands, the mbedTLS support in #4173 is a much more attractive solution to enable Debian and Ubuntu support.

And it looks like in the meantime that the Debian package has begun including this patch, is that correct?

infinity0:
Not literally right now - my previous link was a personal unofficial repo - but I plan to include it in the next upload since it's the easiest option for me to re-enable HTTPS support in the Debian package. I agree that it's good to be careful with crypto libraries, that's why I checked that this patch still fails on an untrusted certificate. (The feature you're talking about doesn't reduce security if it's not supported, rather it just reduces convenience for some smaller use-cases.) What I'm trusting is that the curl maintainers are using their crypto library correctly and providing an unsurprising API for libgit2 to use SSL-enabled-curl, and I'll review this patch and the surrounding code again before I upload it to Debian.

I'd be happy to un-patch this in Debian when mbedTLS support is available. I understand that there's no point in applying this patch on your side, if that would be done soon.

ethomson:
mbedTLS has landed and resolves this problem.  Please let me know if our mbedTLS support does not meet your needs.

PR #4326, opened 2017-08-09
sha1_position: convert do-while to while

peff:
If we enter the sha1_position() function with "lo == hi", we have no elements. But the do-while loop means that we'll enter the loop body once anyway, picking "mi" at that same value and comparing nonsense to our desired key. This is unlikely to match in practice, but we still shouldn't be looking at the memory in the first place.

This bug is inherited from git.git; it was fixed there in git/git@e01580cfe01526ec2c4eb4899f776a82ade7e0e1.

### DIFF
--- a/src/sha1_lookup.c
+++ b/src/sha1_lookup.c
@@ -232,7 +232,7 @@ int sha1_position(const void *table,
 {
 	const unsigned char *base = table;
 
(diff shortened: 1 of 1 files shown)


ethomson:
Thanks!

Issue #3058, opened 2015-04-18
Support for shallow repositories


carlosmn:
git can write a file `.git/shallow` to indicate that the history is cut off at particular commits. We currently do not read this at all and thus libgit2 will regularly fail to work on these repositories with error messages about failing to find objects.

The typicall way such a repository is created is by the use of `git clone --depth N`. We do not support this option to clone either, as we do not support the depth negotiation in the protocol.

In order to provide support we would have to make sure to check against the list in `.git/shallow` whether we're at the end of the history we should expect to be at, and then not try to walk further back or ignore errors when looking up parents.

There are a couple of issues which ask questions about this, I will close them and redirect them here to have a cleaner position from which to start.

arthurschreiber:
Maybe this is super naive, but couldn't support for this be added by simply automatically hiding all commits that are in this list when creating a new `revwalk` instance?

carlosmn:
Something like that might indeed be enough for the git-log use-case, but we still need to figure out how to make that work with merge-base.

The network layer also needs to know this because we need to make the server aware that we're on a shallow repo so it doesn't try to re-use objects we don't have, and push needs to make sure we don't attempt to perform an impossible upload.

briansoule:
+1 FWIW

aral:
Just ran into #1430, which led me here. What’s the current status on this? Has anyone managed to get shallow repositories to work with fetches? Thanks!

ethomson:
@aral No, there has been no update here - this is not currently supported, nor is it a high priority feature for us to implement, I'm afraid.

aral:
@ethomson Understood. Thanks.

Qix-:
🤷‍♂️ bump and +1 here.

getify:
My use case: a huge github repo (more huge in terms of history than in terms of file sizes), but I only need to append to it with a single commit. To avoid the big (and growing) working directory, I just do a bare clone, and then use the API (via nodegit, actually) to create and commit the file, and push.

But the repository also has a long history (and growing, by a few thousand commits per day). So over time, all those commits I'm doing just chew up space with all that local git history. Every 20,000 commits or so, I'd like to just clear out the repository and re-clone it shallow (depth=1), because I only need the very last commit to add a new one.

Of course, I can do the clean/re-clone via a cron job, but the problem is that coordinating between a cron job and any running node process (to ensure I don't clobber in progress commit work) is much uglier/hacky than I'd like it to be. :(

[...]

PR #4331, opened 2017-08-22
Shallow commits & grafts support

tiennou:
This PR adds graft support at the ODB layer, with `git_repository` being responsible for telling `git_odb` about known grafts. Shallow object support is then grafted (heh) on top of that.

Supersedes #3853.

Notes :
- I'm not that convinced that having grafts at the ODB layer makes sense. FWIW, the crux of the issue is that object parsing is not done when objects are returned by the ODB, but by the caller (usually the `object` parse machinery, but `revwalk` also perform crude parsing for obvious reasons).
- @carlosmn pointed out the preferred approach (ie. that's what `git` does) is to have grafts be "transparent". IMHO, reporting an error in that case might help in the following cases :
  - every object parsed is checked against the list of known grafts (so perf). The error code would just be a subset of `GIT_ENOTFOUND`.
  - as a library user, I would have to recheck every object to know if I'm looking at a graft (with the caveat that this PR doesn't actually exposes grafts, only shallow commits). Having the error code makes me automatically aware (and I can just treat it as `GIT_EITEROVER` if I don't actually care).


tiennou:
I'll keep working on handling protocol part (eg. `git clone --depth` and `git fetch --unshallow`), which doesn't really matter for correct handing of grafts & shallow commits (at least locally).

tiennou:
Just realized I should ping #3058 so we don't get duplicated work. Also, I have a local branch with the beginning of the protocol support which I'll happily publish if someone plans to work on that.

justinfx:
Is there any progress on this merge? Looking to see a fix for shallow clones!

tiennou:
I'd appreciate a quick technical "looks good" review, since actually shallow support at the protocol level will depend on some of the infrastructure I laid out here — except I'll need an atomic way of updating the shallow file which wasn't needed by the local-only fix.

So, here are the questions :
- ⚠️  GitHub is utterly confused about the commit ordering in this PR.
- commits 9150c16, 2197072, 5f00ac7 are about merging the commit parsing done when revwalking with the "standard" object-lookup one, and could be submitted separately.
- commit d057213 adds grafts registration support at the `git_odb` level, which is used by 8a76d68 aea2e72 c54ea4c and 82c3a5f. I'm thinking this could be moved up to the `git_repository` level, opinions ?
- 974d720 and 7df1483 handle shallow objects on top of the aforementioned graft API.

src/commit.c
```diff
@@ -476,6 +496,17 @@ int git_commit__parse(void *_commit, git_odb_object *odb_obj)
 	return -1;
 }
 
+int git_commit__parse(void *_commit, git_odb_object *odb_obj)
```
pks-t: Okay, so it seems you created the new function to be able to parse commits from raw data. This looks a lot like what I have been doing in PR #4374, but I went even further and actually decoupled the commits completely from the ODB. Seeing that your approach wouldn't solve my particular use case (as I cannot use the ODB to parse the object), would my approach be able to solve your use case?

src/commit.h
```diff
@@ -36,4 +36,17 @@ struct git_commit {
...
+typedef enum {
+	GIT_COMMIT_PARSE_TREE = (1 << 0),
+	GIT_COMMIT_PARSE_PARENTS = (1 << 1),
+	GIT_COMMIT_PARSE_AUTHOR = (1 << 2),
+	GIT_COMMIT_PARSE_COMMITTER = (1 << 3),
+	GIT_COMMIT_PARSE_MESSAGE_ENCODING = (1 << 4),
+	GIT_COMMIT_PARSE_RAW_HEADER = (1 << 5),
+	GIT_COMMIT_PARSE_RAW_MESSAGE = (1 << 6),
+	GIT_COMMIT_PARSE_ALL = 0xFFFF,
+} git_commit__parse_flags;
```
pks-t: What's the purpose of those flags? Is it an optimization or is it that you cannot parse some stuff because of certain conditions? If it's for the purpose of optimization, I'd like to avoid doing that. We currently don't have any notion of "partial" objects, and I bet this could theoretically lead to subtle bugs when passing around those partial objects.

src/commit_list.c
```diff
@@ -107,75 +110,34 @@ git_commit_list_node *git_commit_list_pop(git_commit_list **stack)
...
 
-	commit->parents = alloc_parents(walk, commit, parents);
-	GITERR_CHECK_ALLOC(commit->parents);
+	if ((error = git_object__init_with_odb((git_object **)&commit, obj, walk->repo)) < 0)
+		return error;
 
-	buffer = parents_start;
-	for (i = 0; i < parents; ++i) {
-		git_oid oid;
+	if ((error = git_commit__parse_commit(commit, obj, GIT_COMMIT_PARSE_PARENTS|GIT_COMMIT_PARSE_COMMITTER)) < 0)
```
pks-t: Ok, I guess that answers my question. Nice to see some reduction in duplicated code.

src/commit.c
```diff
@@ -419,6 +418,10 @@ int git_commit__parse_commit(git_commit *commit, git_odb_object *odb_obj, git_co
 		}
 	}
 
+	/* Perform necessary grafts */
+    git_repository_odb__weakptr(&odb, repo);
+    if (odb)
+        git_odb__graft_object(commit, odb);
```
pks-t: Huh? Either it is GitHub which confused the commit ordering or this commit is too early in that it introduces using `git_odb__graft_object` previous to it being implemented.

src/odb.c
```diff
@@ -669,6 +671,97 @@ void git_odb_free(git_odb *db)
...
+	git_oid_cpy(&graft->oid, oid);
+	k = git_oidmap_put(odb->grafts, &graft->oid, &res);
+	if (res < 0) {
+		error = -1;
+		goto cleanup;
+	}
+
+	git_oidmap_set_value_at(odb->grafts, k, graft);
+
+cleanup:
```
pks-t: Shouldn't you free `graft` here in case of an error?

src/odb.c
```diff
@@ -669,6 +671,97 @@ void git_odb_free(git_odb *db)
...
+static int graft_for_oid(git_odb__graft **out, git_odb *odb, const git_oid *oid)
+{
+	size_t pos;
+
+	assert(out && odb && oid);
+
+	pos = git_oidmap_lookup_index(odb->grafts, oid);
+	if (!git_oidmap_valid_index(odb->grafts, pos)) {
+		return GIT_ENOTFOUND;
+	}
```
pks-t: Minor style issue: no need for braces

src/odb.c
```diff
@@ -669,6 +671,97 @@ void git_odb_free(git_odb *db)
...
+	size_t i;
+	int error = 0;
+
+	/* We only handle commit grafts */
+	if (git_object_type((git_object *)commit) != GIT_OBJ_COMMIT)
+		return 0;
+
+	if ((error = graft_for_oid(&graft, odb, git_object_id(&commit->object))) < 0) {
+		return 0;
+	}
```
pks-t: No need for braces

src/odb.c
```diff
@@ -669,6 +671,97 @@ void git_odb_free(git_odb *db)
...
+	if (!git_oidmap_valid_index(odb->grafts, pos)) {
+		return GIT_ENOTFOUND;
+	}
+
+	*out = git_oidmap_value_at(odb->grafts, pos);
+
+	return 0;
+}
+
+int git_odb__graft_object(void *obj, git_odb *odb)
```
pks-t: Why do you use a void pointer here? Shouldn't that be a `git_object *` pointer instead?

src/odb.c
```diff
@@ -669,6 +671,97 @@ void git_odb_free(git_odb *db)
...
+{
+	git_odb__graft *graft = NULL;
+	git_commit *commit = obj;
+	git_oid *oid;
+	size_t i;
+	int error = 0;
+
+	/* We only handle commit grafts */
+	if (git_object_type((git_object *)commit) != GIT_OBJ_COMMIT)
+		return 0;
```
pks-t: Shouldn't we report that error to the user? I'd expect the equivalent of `return GIT_EINVAL;` here

src/odb.c
```diff
@@ -669,6 +671,97 @@ void git_odb_free(git_odb *db)
...
+	}
+
+	git_array_clear(commit->parent_ids);
+	git_array_init_to_size(commit->parent_ids, git_array_size(graft->parents));
+	git_array_foreach(graft->parents, i, oid) {
+		git_oid *id = git_array_alloc(commit->parent_ids);
+		GITERR_CHECK_ALLOC(id);
+
+		git_oid_cpy(id, oid);
+	}
```
pks-t: Okay, so this is the actual grafting mechanism, I guess.

From a cursory read of git-replace(1) this does not look a 100% correct. What you do here is replace the existing commit's parents in-place. From the man page:

>Create a graft commit. A new commit is created with the same content as <commit> except that its parents will be [<parent>...] instead of <commit>'s parents. A replacement ref is then created to replace <commit> with the newly created commit. See contrib/convert-grafts-to-replace-refs.sh for an example script based on this option that can convert grafts to replace refs.

So instead of editing the existing commit, I think you should instead return a new commit which then replaces the grafted commit. You can convince me otherwise though, I'm not familiar with grafting/replace mechanisms.

src/repository.c
```diff
@@ -1066,25 +1066,42 @@ void git_repository_set_config(git_repository *repo, git_config *config)
 	set_config(repo, config);
 }
 
-int git_repository_odb__weakptr(git_odb **out, git_repository *repo)
+static int git_repository__setup_odb(git_odb **out, git_repository *repo)
```
pks-t: We prefer to not have `git_whatever__` prefixes for static functions which are local to the compilation unit only.

src/repository.c
```diff
@@ -1066,25 +1066,42 @@ void git_repository_set_config(git_repository *repo, git_config *config)
...
 {
+	git_buf odb_path = GIT_BUF_INIT;
+	git_odb *odb;
```
pks-t: Not your fault, but god is that diff awful to read.

src/repository.c
```diff
@@ -1066,25 +1066,42 @@ void git_repository_set_config(git_repository *repo, git_config *config)
...
 
-		if ((error = git_repository_item_path(&odb_path, repo,
-				GIT_REPOSITORY_ITEM_OBJECTS)) < 0 ||
-			(error = git_odb_new(&odb)) < 0)
-			return error;
+	if ((error = git_odb__set_caps(odb, GIT_ODB_CAP_FROM_OWNER)) < 0 ||
+		(error = git_odb__add_default_backends(odb, odb_path.ptr, 0, 0)) < 0) {
+		git_buf_free(&odb_path);
+		git_odb_free(odb);
+		return error;
```
pks-t: A goto out with common cleanup would be nice here.

src/repository.c
```diff
@@ -1066,25 +1066,42 @@ void git_repository_set_config(git_repository *repo, git_config *config)
...
+static int git_repository__setup_odb(git_odb **out, git_repository *repo)
 {
+	git_buf odb_path = GIT_BUF_INIT;
+	git_odb *odb;
 	int error = 0;
 
-	assert(repo && out);
+	if ((error = git_repository_item_path(&odb_path, repo,
+			GIT_REPOSITORY_ITEM_OBJECTS)) < 0 ||
+		(error = git_odb_new(&odb)) < 0)
```
pks-t: I know you're just using the preexisting code, but we actually have a memory leak here whenever `git_odb_new` errors. I mean in that case we probably have other problems because that indicates memory allocation errors, but still. Would be cool if you could refactor this a bit to not have multiple statements in a single if-condition and use `goto cleanup`-style to avoid those memory leaks.

src/buffer.h
```diff
@@ -215,4 +215,8 @@ int git_buf_splice(
 	const char *data,
 	size_t nb_to_insert);
 
+#define git_buf_foreach_line(line_start, line_end, line_num, buf) \
+while ((line_start = git__strsep(buf.ptr, "\n")) != NULL && \
+	(line_end = line_start + strlen(line_start)) != NULL && ++line_num)
```
pks-t: Please indent those lines part of the macro

pks-t: Furthermore, please wrap all uses of the macro parameters in braces.

src/repository.c
```diff
@@ -1066,6 +1066,74 @@ void git_repository_set_config(git_repository *repo, git_config *config)
...
+	const char *line_end;
+	int line_num = 0;
+	int error, updated;
+	git_array_oid_t parents = GIT_ARRAY_INIT;
+
+	if ((error = git_repository_item_path(&graft_path, repo, GIT_REPOSITORY_ITEM_INFO)) < 0)
+		return error;
+
+	if (git_buf_joinpath(&graft_path, graft_path.ptr, "grafts")) {
+		git_buf_free(&graft_path);
```
pks-t: Maybe have a common `goto cleanup` again? It's becoming hard to grasp what needs cleaning at which part of the code with those interspersed `free`s and `return`s.

tiennou:
I'll repeat what I said on Slack for the record : since I'm coming at this from the shallow-support angle, which `cgit` uses grafts for, I've tried to keep it logical and do the same thing. This means I'm clueless about the `git-replace` stuff, apart from what I've read from the manual : it "grafts" complete objects, and uses refs to keep track of them, hence those "replacement objects" can be pushed/pulled around, while grafts (and thus shallow objects) can't.

This PR only exists because of shallow-support, hence I've checked how `cgit` does it, and it does not use `git-replace` for that, so I considered `git-replace` support "out of scope".

> So one thing I'm a bit confused about is how grafting and replace work with each other or against each other. As far as I understand it, a graft is a new commit with the exact same content as another commit, but with parents replaced.

"replaces" are real objects and AFAIU work like you expected. "grafts" are just some OIDs loaded from `.git/info/grafts`, which are used to override a commit's parents…

> The thing to note here is that this graft is part of the ODB, that is it is a commit object part of the ODB with its own OID.

…thus, no. "grafts" have no real existence (IOW object form). I located my graft function at the ODB level because 1) I got confused thinking that `.git/info/` was part of the odb, 2) it felt the most obvious place I could tack them on and impact both revwalking and object lookups.

> If I'm not mistaken, you are replacing the old logic used by grafts, which weren't using the new replacement logic. Was this a knowing decision, and if so why?

"replacing" I don't understand. If you meant "implementing", then yes, because that's how `cgit` does it.

We seem to agree on how `git-replace` works. I'll just add that I'm not sure `cgit` "grafts" anything when it's used, since I didn't see anything related to replaces in the graft-handling code (% overlook probability), hence why I consider `git-replace` out of scope.

I'd be happy to discuss what final API we want : for shallow support, I only need "internal" things, so I don't particularly care (I mean, I could *not* provide any public API apart from the list of shallow roots and be done with it 😉). Right now, that's handled by `git_odb__graft_register`, which could be made public.

I'll reinvestigate the layering I chose : because of my 2 points above, I might not need to go near the odb at all, but I have to consider a `.git/shallow`/`.git/info/grafts` file appearing "somehow" (IOW caching, fun things™)

If it's warranted I can split out that PR in 3 pieces (already have them locally, and if it fixes GitHub's confusion that's be even better) : the commit parser merging (because of #4374 and being generally cleanup), the graft support itself (this), and shallow support (which requires this).

src/commit.c
tiennou: As I think you saw below, this is all in the name of merging the parsing done when quickly revwalk with the "standard" way. After a quick look, I don't think your approach allows me to do that (it's just decoupling IIUC).

src/commit.h
tiennou: Idem, revwalk parse merge.

src/commit_list.c
tiennou: 👍

src/odb.c
tiennou: Was expecting a "layering violation", because of `git_odb_object`. Also, I think I saw a precedent, hence thought it would be fine. If it's not, it should be changed to `git_commit *` because you only graft onto commits.

src/odb.c
tiennou: It's the `cgit` graft approach, replacing parents on object parse.

src/object.c
```diff
@@ -51,14 +51,39 @@ static git_object_def git_objects_table[] = {
 	{ "REF_DELTA", 0, NULL, NULL },
 };
 
+int git_object__init_with_odb(git_object **object_out, git_odb_object *odb_obj, git_repository *repo)
```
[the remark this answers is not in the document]
tiennou: I would have loved too 😆. IIRC it's a side-effect of the parse merging. After a more thorough look, it might just be confusion : the revwalk-parsing works on `git_commit`, and the normal parsing did it, so I made them *both* do it. Will investigate if it's really necessary (I remember a crash when freeing though).

src/commit_list.c
```diff
@@ -196,8 +158,7 @@ int git_commit_list_parse(git_revwalk *walk, git_commit_list_node *commit)
 	} else
 		error = commit_quick_parse(
 			walk, commit,
-			(const uint8_t *)git_odb_object_data(obj),
-			git_odb_object_size(obj));
+            obj);
```
tiennou: Indent

src/commit_list.c
```diff
@@ -107,75 +110,34 @@ git_commit_list_node *git_commit_list_pop(git_commit_list **stack)
 
 static int commit_quick_parse(
 	git_revwalk *walk,
-	git_commit_list_node *commit,
-	const uint8_t *buffer,
-	size_t buffer_len)
+	git_commit_list_node *node,
+    git_odb_object *obj)
```
tiennou: Indent

tiennou:
Rebased. I managed to remove most of the "weirdness" by moving the grafts at the repo level.

src/commit_list.c
```diff
@@ -107,75 +110,32 @@ git_commit_list_node *git_commit_list_pop(git_commit_list **stack)
...
-		buffer += parent_len;
-	}
-
-	commit->parents = alloc_parents(walk, commit, parents);
-	GITERR_CHECK_ALLOC(commit->parents);
+	int error;
+	size_t i;
+	git_oid *parent_oid;
+	git_commit *commit = git__calloc(1, git_odb_object_size(obj));
+	commit->object.repo = walk->repo;
```
tiennou: Note : that was what my old `__init_from_odb` function was about. Right now I'm cheating by providing only what's necessary for the parser to work (access to the repo for applying the grafts). I can either bring back the `__init_from_odb` function, or make the parser not crash if there's no repo around. Suggestions welcome !

src/commit.h
```diff
@@ -36,4 +37,20 @@ struct git_commit {
 void git_commit__free(void *commit);
 int git_commit__parse(void *commit, git_odb_object *obj);
 
+typedef enum {
+	GIT_COMMIT_PARSE_QUICK = (1 << 0),
+	GIT_COMMIT_PARSE_ALL   = (1 << 1),
+} git_commit__parse_flags;
+
+int git_commit__parse_commit(git_commit *commit, git_odb_object *obj, git_commit__parse_flags flags);
```
tiennou: I've cut down on the number of flags here because of a bug parsing parents. I think it's better this way, as it makes it less likely you ask for an invalid combination, and we can always split those apart if the need arises.

src/commit.c
pks-t: Yup, you're right. I tend to review by commits, so I didn't see this until later.

src/odb.c
pks-t: The question that occurs to me then is why this function is called `git_odb__graft_object` when you're only supposed to graft commits ;)

src/odb.c
pks-t: Huh. Personally, I'd think this is the wrong approach. If we just replace in place, we have no mechanism to use those grafts for replaces later on, as that requires us to write the grafted object into the ODB to create the references.

pks-t:
Okay, so I think I've been confused a bit by grafts and replaces. What makes me curious is the following part from git-replace(1):

> --graft <commit> [<parent>...]
>    Create a graft commit. A new commit is created with the same content as <commit> except that its
>    parents will be [<parent>...] instead of <commit>'s parents. A replacement ref is then created to
>    replace <commit> with the newly created commit. See contrib/convert-grafts-to-replace-refs.sh for an
>    example script based on this option that can convert grafts to replace refs.

This paragraph implies that the grafted commit is stored inside of the ODB, as otherwise it would be impossible to put the reference into place. From what you say, it looks to me like there's two different things here:

- an in-memory grafted commit which is not written into the ODB, used by the old grafts
- an in-ODB grafted commit which is written into the ODB, used by the new replace mechanism

So I guess both grafting and replaces use the same underlying mechanism of grafting, but differ only in how they are being referenced and persisted. While old-style grafts make use of out-of-band information, the new replace mechanism is simpler by making use of the ODB and refdb. It would be nice though if we could also check how git.git does it, instead of relying on another third-party implementation.

Anyway. Seeing that both replaces and grafts make use of the same mechanisms, it would be nice if they shared some common logic with each other later on.

tiennou:
Your interpretation of `git-replace` looks correct to me. That `--graft` option creates a "new style" replace object that works the same as the "old style" graft mechanism (same name, different implementation).

What I'm sure of is that `cgit` shallow objects use old-style grafts, which don't appear in the ODB at all (or you'd have strange things happen when pushing a shallow clone back to its origin). This means that, even though the old graft mechanism could be considered "deprecated" in the face of "git-replacing", which is superior, it's still used by the shallow code *because* you don't want the user to mess up its shallow clones.

I think that because of that specific point, both "object replacement methods" cannot share the same implementation, but I'll have to investigate how git-replace(1) actually works and what kind of functionality it uses to be sure.

pks-t:
Okay, that's most welcome, thanks a lot.

By the way, I'm all in favor of splitting this up, as you proposed. Especially if you've already got the parts ready it's always easier to get smaller batches reviewed and merged than one big batch.

tiennou:
@pks-t Okay, I've split the 3 logical parts off. Sadly I can't really base them on top of one another because they're in my fork (and opening PRs against that would be silly), so it's not as clean as we might have wanted.

One thing I'd like to discuss still is how we report those shallow objects. Right now, this is completely "transparent" (as per spec, may I say), but you, as a user, would have no way of knowing whether this is a "pristine commit" or a shallow one, without suspiciously checking the OID of every "root" commit against the list of shallow roots. For example, in GitX, I'd be able to open a prompt asking whether the user wants a few more commits loaded when revwalking (or something). Arguably, I might be overthinking it though 😉.

tiennou:
All 3 parts have just been updated, fixing a few leaks and adding some more tests (mostly to the graft API). I've also moved the graft functions in their own file instead of having them in commit.c/h.

This makes the local part of shallow support complete, so libgit2 can work with grafted/shallowed repositories. I'm still undecided on whether `git_revwalk_next` on a grafted commit should return `GIT_EGRAFT` or `GIT_EITEROVER` though (the latter is currently done).

tiennou:
This full PR/patch set has been rebased, and should be ready for review.

I'm drawing a line in the sand as to what this is about: this only makes detection of shallow repositories possible. I'll do another PR about the "protocol" parts of shallow when I feel it's ready, because it's (arguably) harder. Hence, some safety measures are missing: I think a fetch done by us from a shallow repo would unshallow it, since we can't tell the remote about what we're purposefully not interested in (though it could also confuse the server 😬 ?). Same for pushing, though I'm clueless about what's involved there. It might be worthwhile to prevent misbehaving by erroring when those operations are tried on a shallow repository.

My point about checking for a grafted commits when a revwalk ends is also still undecided. I like "erroring differently" better, but I understand that might be considered an API-break. Maybe `git_commit_is_grafted` or something like it…

r-cheologist:
What's the status of this?
Really looking forward to shallow clone support ...

GreatBahram:
Is there any progress on this problem? There are lots of issues that are demand on this features.

Qix-:
Ping @pks-t.

kalkin:
As far as I understand this stops git revwalk from exiting with an error if a local commit is missing.
@carlosmn @ethomson @pks-t What needs to be done to get this PR merged?

include/git2/repository.h
```diff
@@ -851,6 +852,17 @@ GIT_EXTERN(const char *) git_repository_get_namespace(git_repository *repo);
...
+/**
+ * Determine the shallow roots of the repository
+ *
+ * This oidarray is owned by the library. Do not free it.
+ *
+ * @param out An array of shallow oids.
+ * @param repo The repository
+ * @return 0 on success, an error otherwise.
+ */
+GIT_EXTERN(int) git_repository_shallow_roots(git_oidarray *out, git_repository *repo);
```
pks-t: We're probably unable to mark `git_oidarray` as const, right? Maybe we should instead just return a `git_grafts` structure? After all, that's how shallow roots are implemented.

tests/repo/shallow.c
```diff
@@ -90,3 +90,23 @@ void test_repo_shallow__errors_on_borked(void)
...
+
+	cl_git_pass(git_revwalk_next(&oid_1, w)); // a65fedf39aefe402d3bb6e24df4d4f5fe4547750
+	cl_git_pass(git_revwalk_next(&oid_2, w)); // be3563ae3f795b2b4353bcce3a527ad0a4f7f644
+	cl_git_fail_with(GIT_ITEROVER, git_revwalk_next(&oid_3, w));
+
+	cl_assert_equal_s(git_oid_tostr_s(&oid_1), "a65fedf39aefe402d3bb6e24df4d4f5fe4547750");
+	cl_assert_equal_s(git_oid_tostr_s(&oid_2), "be3563ae3f795b2b4353bcce3a527ad0a4f7f644");
+
+	git_revwalk_free(w);
+}
```
pks-t: This test's commit can probably be amended to the other commit that introduced shallow tests.

Qix-:
If OP has gone inactive since submitting this let me know and I'll be happy to take over with a new PR.

pks-t:
@tiennou is one of the most active people nowadays in the project, so I'd be surprised if he was inactive now. But open source being what it is it might still take some time ;)

tiennou:
It's likely I'll only able to take a look at it in a couple of days. No sweat for the "late" review @pks-t, better late than never ! Frankly, if you feel up for it @Qix- I'll gladly defer the PR/polishing part, as I'd personally prefer to keep working on the ssh backend issue — as the review shows it's been a little while.

Just a quick discussion point : as the overarching goal is to be `git`-like, I don't really feel compelled to provide anything more featureful than what's needed to support shallow repositories, with the rationale that git has gained `git-replace` (a.k.a. the refdb-based grafts, which do look nicer). I am thus confused — to me grafts are a low-level detail of a since-replaced-by-a-better-version system, and don't really warrant a more precise cache, or more attention than that. Maybe I'm missing something though, so I'm open to usecases 😉.

pks-t:
> I am thus confused — to me grafts are a low-level detail of a since-replaced-by-a-better-version system, and don't really warrant a more precise cache, or more attention than that. Maybe I'm missing something though, so I'm open to usecases .

I fully agree that we shouldn't blow up the grafts mechanism more than needed. But my intention is to have it as self-contained as possible to not spill grafts-related logic into other places. Thus the standalone `git_grafts` struct, `git_grafts__refresh`, etc. This will make it more maintainable and help us if there is ever a request to expose some of its functions it to users.

I'll create a separate PR that builds on yours to show what I imagine it to look like :)

Qix-:
@pks-t This can be closed, I believe.

PR #4332, opened 2017-08-25
Conditional includes

pks-t:
I've accidentally closed #4244 as I have deleted my branch. Re-opening here.


pks-t:
Rebased to fix conflicts.

pks-t:
I've fixed the issue on Windows now, but tests on macOS currently fail. I have no idea why that is and am unfortunately unable to debug into that issue due to not having a working macOS setup right now. :(

carlosmn:
The error on macOS comes from a peculiarity of how it handles temporary paths. The `clar_sandbox_path()` is (for one example)

    /var/folders/q8/y3fc7wsx7f5cl7pdft67g_f40000gn/T/libgit2_tests_KbEJqZ/

whereas the `git_repoistory_path()` is

    /private/var/folders/q8/y3fc7wsx7f5cl7pdft67g_f40000gn/T/libgit2_tests_KbEJqZ/empty_standard_repo/.git/

The reason we get different paths is that we perform a whole bunch of `realpath(3)` calls when we open the repository which translate the "virtual" path into a more real one.

src/config_file.c
```diff
@@ -1568,11 +1571,148 @@ static int config_parse(
...
+{
+	char *condition;
+	size_t i;
+	int error = 0, matches;
+
+	if (!parse_data->repo)
+		return 0;
+
+	condition = git__substrdup(section + strlen("includeIf."),
+				   strlen(section) - strlen("includeIf.") - strlen(".path"));
```
carlosmn: This might just be me, but I prefer the method used in the normalisation which searches for the first and last `'.'` to figure out how long the middle part is without making us count hard-coded strings.

carlosmn:
This CMake module would be available in any installation? I would definitely much prefer an implementation which uses that.

carlosmn:
I went ahead and did the arg-order change. Yay for features.

pks-t:
The parameter reordering makes sense. Thanks for improving and merging!

csware:
Why pass a repo object into git_config_functions and not just a path? This would allow to construct git_config objects w/o intantiating a repo first.

carlosmn:
Because we don't know what else git is going to come up with and adding a path would mean that we're going to have to constantly change the interface.

PR #4335, opened 2017-09-02
Trying to make the code more readable

flaviommedeiros:
Hi all,

I'm running some C code analysis tool that pointed me to this small code fragment. I'm suggesting this patch to try to make the source code more understandable by removing mixing of ternary if statements with attributions.

Would you guys agree with this type of changes? The tools pointed me to almost 20 cases like this one in the source code of libgit2.

### DIFF
--- a/src/attr_file.c
+++ b/src/attr_file.c
@@ -783,7 +783,8 @@ int git_attr_assignment__parse(
 
 		/* if there is an equals sign, find the value */
 		if (*scan == '=') {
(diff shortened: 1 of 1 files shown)


ethomson:
I don't think this is much more readable.  I definitely prefer the brevity.

However, we could move the increment into the test which would accommodate both.  But I'm not particularly concerned about this - I think it's readable enough, so my inclination is to close this...

PR #4344, opened 2017-09-11
Clear the remote_ref_name buffer in git_push_update_tips()

slavikus:
If `fetch_spec` was a non-pattern, and it is not the first iteration of `push_status` vector, then `git_refspec_transform()` would result in the new value appended via `git_buf_puts` to the previous iteration value.

Forcibly clearing the buffer on each iteration to prevent this behavior.


ethomson:
Thanks for the fix!

PR #4348, opened 2017-09-15
diff: cleanup hash ctx in `git_diff_patchid`

pks-t:
After initializing the hash context in `git_diff_patchid`, we never
proceed to call `git_hash_ctx_cleanup` on it. While this doesn't really
matter on most hash implementations, this causes a memory leak on Win32
due to CNG system requiring a `malloc` call.

Fix the memory leak by always calling `git_hash_ctx_cleanup` before
exiting.


ethomson:
Good catch, thanks @pks-t, as always.

PR #4349, opened 2017-09-15
tests: introduce ability to use config-driven ignore tests

pks-t:
In an ideal world, all projects aiming to implement functional aspects
of Git would share a common test suite. This test suite would be in a
generic format readable by all projects and could then be used to drive
unit tests. Like this, it can be made sure that all these different
projects would at least behave the same regarding to this common base of
tests.

One of the easiest candidates to create such a shared test suite is the
gitignore rules. All that is required is a format that specifies all the
gitignore files that should exist as well as a list of files that should
or should not be ignored, respectively. A natural choice here is to
simply use Git configuration files to convey this information, as all
Git projects should already know how to parse them. Like this, only
minimal additional code is required to have a test driver for those
specifications.

This commit implements a first proof of concept for such a test
specification. Each test specification is contained in a single file.
This file has several keys:

    - "test.name": name for the test, which can be used by the unit test
      implementation to give hints to the user which test is currently
      being executed
    - "assert.ignored": path to a file which should be ignored
    - "assert.not-ignored": path to a file which should not be ignored
    - "gitignore.path": path of a new gitignore file. All the following
      rules will be written into this path.
    - "gitignore.rule": rule that should be written to the current
      "gitignore.path" file. Can only be specified if a "gitignore.path"
      has been configured before.

Due to the dependence between "gitignore.rule" and "gitignore.path", the
configuration file needs to always be parsed from top to bottom.
Adding a second "gitignore.path" key will cause all the following
"gitignore.rule" values to instead be added to the new file path.

This is obviously only a proof of concept. First, libgit2 fails to
provide a way of iterating through all config keys from top to bottom,
but will only allow you to iterate through them in indeterminate order.
Second, we shouldn't be adding the test specifications to our own
projects but instead have them part of a submodule. Its repository
should probably live at an official location, where it is accessible to
other projects like git.git and JGit, as well.

So currently, I'm just seeking feedback in our small group and will
later on publish it for comments on the Git mailing list (obviously,
with code specific to git.git such that they can use the test specs for
their project, as well).

---

So I've been wanting to do something like this for a long time. Actually, the idea came up back in January 2017 at Git Merge, where somebody proposed to have such a generic test suite shared across multiple projects. As I haven't seen any progress for this yet, I now start to work on it.

As said, it doesn't work yet for libgit2. Problem is our configuration interface, which is unable to iterate through config variables from top to bottom. I'll dig into that shortcoming at a later point.


jarrybomb (dismissed):

carlosmn:
>the configuration file needs to always be parsed from top to bottom.

There's a difference between parsing in-order (which we have to do) and making the order of different keys significant. Instead of having `gititnore.rule`, `gitignore.path` that have to be in the right order, I would rather have `gitignore.somename.rule` and` gitignore.somename.path` so they are grouped under a particular name and we can not care about what relative order anything is written in.

We could also have `gitignore.thepath.rule` which tells us in the config key which file they belong to, similar to how we put urls in the middle portion of keys for other things.

tests/resources/attr/ignores/temporary.conf
```diff
@@ -0,0 +1,15 @@
...
+name=Honor temporary rules
+
+[assert]
+not-ignored=File.txt
+ignored=NewFolder
+ignored=NewFolder/NewFolder
+ignored=NewFolder/NewFolder/File.txt
+
+[gitignore]
+file=/.gitignore
```
carlosmn: Does `/` here mean the worktree root? If so why not specify that the path is to be relative to the worktree root and have them bee relative.

tests/attr/ignore.c
```diff
@@ -31,14 +31,75 @@ static void assert_is_ignored_(
...
-void test_attr_ignore__honor_temporary_rules(void)
+static void setup_ignore_config(git_config *cfg)
 {
-	cl_git_rewritefile("attr/.gitignore", "/NewFolder\n/NewFolder/NewFolder");
+	git_buf file = GIT_BUF_INIT, patterns = GIT_BUF_INIT;
+	git_config_iterator *it;
+	git_config_entry *entry;
+	int error;
+
+	cl_git_pass(git_config_iterator_new(&it, cfg));
```
carlosmn: What you want here is not an iterator over all values, but to look up `assert.not-ignored` and `assert.ignored` as multivars.

Same with `gitignore.rule`.

carlosmn:
@pks-t I updated your branch with some code that uses the `gitignore.rule` variables as a mutlivar, which is what you're actually using here (and a rule that was the wrong way around and fixed a free of an unowned resource).

We definitely should have this kind of cross-implementation test resources. Would the idea be that we have some particular reference repository (or just a worktree) and we define all the rules based on that?

I'm of two minds regarding the specification of rules on a config file. The alternative here would be to specify what gitignore file we want to use and leave the ignored/not-ignored in a config file. I'm not sure if it'd really be better, but it does feel more "natural" in a way.

But let's not worry too much about that. With my changes (and I think the check for ignored or not ignored should also use a mutlivar iterator) we're now passing so we should be able to build on that to have something to present to the outside.

pks-t:
I prefer your approach of an explicit gitignore file. Maybe have it like that:

```
[test]
name=Honor temporary rules

[gitignore "/.gitignore"]
file=gitignore-root

[assert]
ignored=File.txt
ignored=NewFolder
ignored=NewFolder/NewFolder
ignored=NewFolder/NewFolder/File.txt
not-ignored=File2.txt
```

gitignore-root:

```
File.txt
NewFolder
NewFolder/NewFolder
NewFolder/NewFolder/File.txt
```

Like that, the gitignore rules are easier to read but we can still dictate where to put the actual gitignore files.

pks-t:
Oh, and regarding your question whether that should be a worktree or a repository. Personally I'd like to have that as a centrally hosted repository, maybe as part of the "git" group in GitHub. Projects making use of that can then easily add it as a submodule (preferred if you ask me, but that's up to the project) or as a subtree.

pks-t:
I've amended the PR like I said previously. I also moved out the tests to a new test suite so that they're separate from our "usual" gitignore test.

PR #4357, opened 2017-09-22
Ask for SSH credentials again when passphrase is wrong

omus:
When trying to decode the private key it looks like LibSSH2 returns a `LIBSSH2_ERROR_PUBLICKEY_UNVERIFIED` when the passphrase is incorrect. This change allows callback to be called again giving the user the chance to correct the passphrase or provide alternative credentials.

Additionally, switched the indentation from spaces to tabs which matches the surrounding code.

### DIFF
--- a/src/transports/ssh.c
+++ b/src/transports/ssh.c
@@ -419,8 +419,10 @@ static int _git_ssh_authenticate_session(
 		}
 	} while (LIBSSH2_ERROR_EAGAIN == rc || LIBSSH2_ERROR_TIMEOUT == rc);
 
-        if (rc == LIBSSH2_ERROR_PASSWORD_EXPIRED || rc == LIBSSH2_ERROR_AUTHENTICATION_FAILED)
-                return GIT_EAUTH;
+	if (rc == LIBSSH2_ERROR_PASSWORD_EXPIRED ||
+		rc == LIBSSH2_ERROR_AUTHENTICATION_FAILED ||
+		rc == LIBSSH2_ERROR_PUBLICKEY_UNVERIFIED)
+			return GIT_EAUTH;
 
 	if (rc != LIBSSH2_ERROR_NONE) {
 		if (!giterr_last())


ethomson:
Interesting.  I'm surprised it returns that, but this seems like a reasonable change when it does.

Would you mind wrapping that line?  eg:

```
	if (rc == LIBSSH2_ERROR_PASSWORD_EXPIRED ||
	    rc == LIBSSH2_ERROR_AUTHENTICATION_FAILED ||
	    rc == LIBSSH2_ERROR_PUBLICKEY_UNVERIFIED)
		return GIT_EAUTH;
```

omus:
Will do. I wasn’t sure what the libgit2 standard format was line wrapping.

omus:
I wrapped the text. Note that since the code is indented with tabs it only looks aligned when using tabs which are 4 spaces.

pks-t:
I'm a bit surprised regarding your explanation. From `libssh2_userauth_publickey_fromfile_ex(3)`:

```
       LIBSSH2_ERROR_PUBLICKEY_UNVERIFIED - The username/public key combination was invalid.
```

From that, it seems like libssh2 has already been talking to the server here. In the end it doesn't matter, we should obviously return  `GIT_EAUTH` for that case, as well.

pks-t:
One minor nit, after that it's good to go for me: it would be nice if the commit message's subject included the component the patch applies to. E.g. "transports: ssh: ask for credentials again when passphrase is wrong".

omus:
I can modify the commit message subject line. I'm guessing you don't care about commit message subject line length? I try to stick keep it to [50 characters](https://chris.beams.io/posts/git-commit/).

pks-t:
We definitly care about the length, but 50 characters is just a little bit too strict in my opinion. I think something like 65 characters is much saner, especially with things like the component prefix.

omus:
Update the commit message subject and additionally rebased the changes.

omus:
Bump

ethomson:
Awesome!  Thanks for fixing this.  ✨

PR #4339, opened 2017-09-05
Static linking for bundled deps

pks-t:
This fixes #4336.

The problem is that we do not link in static dependencies for a static libgit2 library. The mistake stems from expecting `TARGET_LINK_LIBRARIES` to merge static archives, which does not happen.

This PR fixes the issue by using object libraries for all of our dependencies. This is the easiest approach and will keep build instructions separate for all of our dependencies. The main downside is that it is only available for CMake v2.8.8 and higher. Last time I looked there were some current distributions which do not have this version, probably something like RHEL or Debian.

I didn't have time to re-check this, so I simply raised the version to v2.8.8 for now. We can adapt this fix to work with smaller versions, as well, by always setting a variable for dependencies which either contains the object library generator expression or simply the source files, depending on version. I'll check whether that is required soonish and adjust the PR accordingly. For now, it's only intended as a proof of concept.

PR #4359, opened 2017-09-27
Plug some leaks in curl's proxy handling

carlosmn:
I was trying to reproduce the segfault we see in Travis when talking to the proxy, but instead I found a few leaks.

### DIFF
--- a/src/proxy.c
+++ b/src/proxy.c
@@ -32,7 +32,7 @@ int git_proxy_options_dup(git_proxy_options *tgt, const git_proxy_options *src)
 	return 0;
 }
 
-void git_proxy_options_free(git_proxy_options *opts)
+void git_proxy_options_clear(git_proxy_options *opts)
 {
 	git__free((char *) opts->url);
 	opts->url = NULL;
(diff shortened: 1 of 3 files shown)


ethomson:
Neat, thanks for fixing these.  Which segfault?  The one from with NTLM creds from #4317 and #4321?

carlosmn:
I was looking at #4339 where the thread-unsafe suite seems to be failing somewhat reliably, but I've been unable to reproduce it.

ethomson:
Oh, that's interesting, isn't it.  I hadn't seen that.

src/proxy.c
```diff
@@ -31,3 +31,9 @@ int git_proxy_options_dup(git_proxy_options *tgt, const git_proxy_options *src)
 
 	return 0;
 }
+
+void git_proxy_options_free(git_proxy_options *opts)
```
pks-t: We usually call this kind of function `git_proxy_options_clear` instead of `free`. A `free` should also free the structure itself.

carlosmn: Sure, that makes sense. Done.

ethomson:
Thanks @carlosmn !

Issue #3780, opened 2016-05-10
git_graph_descendant_of: Please document that a commit is not a descendant of itself


joshtriplett:
Please consider adding a note to the documentation of `git_graph_descendant_of`, stating that a commit is not considered to be a descendant of itself.

roysc:
This behavior is not consistent with `git merge-base --is-ancestor` - is that intentional? It's worth considering whether this should be changed to match git.

pks-t:
I'm closing this issue as documentation has been added via #4362. For now, I'll treat the behaviour as intentional, until somebody has some good arguments that it is not, and in that case that person should open a new issue :) Thanks for the report!

PR #4362, opened 2017-09-30
Document that a commit is not a descendant of itself

jacwah:
See issue #3780.

### DIFF
--- a/include/git2/graph.h
+++ b/include/git2/graph.h
@@ -40,6 +40,9 @@ GIT_EXTERN(int) git_graph_ahead_behind(size_t *ahead, size_t *behind, git_reposi
 /**
  * Determine if a commit is the descendant of another commit.
  *
+ * Note that a commit is not considered a descendant of itself, in contrast
+ * to `git merge-base --is-ancestor`.
+ *
  * @param commit a previously loaded commit.
  * @param ancestor a potential ancestor commit.
  * @return 1 if the given commit is a descendant of the potential ancestor,


pks-t:
Hi @jacwah, and thanks for your first PR!

I agree that we should at least document that behaviour when it deviates from git.git, so this is definitly a good change. Two comments regarding the change itself:

* the documentation should probably not be part of the return code, but rather of the function documentation's body. See for example how it's done for the preceding function `git_graph_ahead_behind`, where additional information on its behaviour is given after the initial short heading.
* it would be nice if you could prefix the commit message's first line with the scope it applies to. E.g. in this case "graph: document that a commit isn't a descendant of itself".

jacwah:
@pks-t Thanks for the feedback! I updated the PR to your suggestions.

Writing the original commit message, I had a hard time cramming everything into 50 characters (recommended max length). If that limit is softer, your suggestion is obviously superior.

pks-t:
Yeah, the 50 characters limit is simply impossible to fulfill in many cases. We try to be as short as possible here, but I guess it's okay to get to ~75 characters.

pks-t:
I'd still propose that you move the note from the `@return` annotation to between summary and the first `@param`. So

```
/*  Determine if a commit...
 *
 * Your note on descendants here
 *
 * @param...
 * ...
 */
```

This matches existing style more closely and is a bit easier to spot, I guess.

jacwah:
@pks-t Sorry! I didn't commit the changes properly. Should be ok now.

pks-t:
Looks good now, thanks a lot!

PR #4363, opened 2017-10-03
remote: add typedef to normalize push_update_reference callback

cjhoward92:
I added a typedef for `push_update_reference` in `remote.h` to make the callback fit with other well defined callbacks. If this is inappropriate or there is some reason for not having a typedef, let me know.

### DIFF
--- a/include/git2/remote.h
+++ b/include/git2/remote.h
@@ -386,6 +386,20 @@ typedef struct {
 typedef int (*git_push_negotiation)(const git_push_update **updates, size_t len, void *payload);
 
 /**
+ * Callback used to inform of the update status from the remote.
+ *
+ * Called for each updated reference on push. If `status` is
+ * not `NULL`, the update was rejected by the remote server
+ * and `status` contains the reason given.
+ *
+ * @param refname refname specifying to the remote ref
+ * @param status status message sent from the remote
+ * @param data data provided by the caller
+ * @return 0 on success, otherwise an error
+ */
+typedef int (*git_push_update_reference_cb)(const char *refname, const char *status, void *data);
+
+/**
  * The callback settings structure
  *
  * Set the callbacks to be called by the remote when informing the user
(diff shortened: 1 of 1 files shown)


include/git2/remote.h
```diff
@@ -386,6 +386,18 @@ typedef struct {
 typedef int (*git_push_negotiation)(const git_push_update **updates, size_t len, void *payload);
 
 /**
+ * Called for each updated reference on push. If `status` is
+ * not `NULL`, the update was rejected by the remote server
+ * and `status` contains the reason given.
```
pks-t: The style for function documentation is usually as follows:

```
/**
 * Short summary of that thing
 *
 * A longer description which may span multiple lines
 * and describes what that symbol actually is/does.
 *
 * @param frobnicate frobnicate all the stuff
 * @return 0 on foo, otherwise an error
 */
```

So please separate the summary and descriptions here and add an `@return` description, such that developers know what they may return from that callback.

include/git2/remote.h
```diff
@@ -386,6 +386,18 @@ typedef struct {
...
+ *
+ * @param refname being used
+ * @param status message sent from the remote
+ * @param data provided by the caller
+ */
+typedef int (*git_push_update_reference_cb)(const char *refname, const char *status, void *data);
+
+
+/**
```
pks-t: There's one newline too many here

include/git2/remote.h
```diff
@@ -456,7 +468,7 @@ struct git_remote_callbacks {
 	 * not `NULL`, the update was rejected by the remote server
 	 * and `status` contains the reason given.
 	 */
```
pks-t: This comment is duplicated now. You may replace it with something like "See documentation of `git_push_update_reference_cb`".

cjhoward92:
@pks-t Thanks for the feedback! I would very much like to get this merged. Hopefully the changes I made suit the needs of libgit2 better. Let me know if there is anything else I can do for you!

ethomson:
Thanks for fixing this up @cjhoward92.  These changes look 👍 to me - appreciate you making them.

Since this is a blocker for nodegit, I'm going to go ahead and merge this.  If there are additional changes we can iterate from here.

cjhoward92:
Thanks Ed! Right now we should be set. If you apply need someone to go through and add more callback typedefs or clean anything up I'd be more than happy to. One of those simple maintenance things that a new contributor like me can do!

PR #4364, opened 2017-10-04
Use SOCK_CLOEXEC when creating sockets

andoma:
Similar to how O_CLOEXEC is used for `open()` libgit2 should also use SOCK_CLOEXEC (when available) for opening sockets.


pks-t:
Hi @andoma, thanks for your first PR! The change definitly looks sensible to me (even though the interface itself looks horrible, but that's to blame on backwards compatibility here with the `socket` function). Thanks a lot :+1:

PR #4365, opened 2017-10-04
InternalName should not contain the file extension

csware:
InternalName should not contain the file extension as mentioned on <https://msdn.microsoft.com/en-us/library/windows/desktop/aa381058(v=vs.85).aspx>.

### DIFF
--- a/src/win32/git2.rc
+++ b/src/win32/git2.rc
@@ -29,7 +29,7 @@ BEGIN
     BEGIN
       VALUE "FileDescription",	"libgit2 - the Git linkable library\0"
       VALUE "FileVersion",	LIBGIT2_VERSION "\0"
-      VALUE "InternalName",	LIBGIT2_FILENAME ".dll\0"
+      VALUE "InternalName",	LIBGIT2_FILENAME "\0"
       VALUE "LegalCopyright",	"Copyright (C) the libgit2 contributors. All rights reserved.\0"
       VALUE "OriginalFilename",	LIBGIT2_FILENAME ".dll\0"
       VALUE "ProductName",	"libgit2\0"


ethomson:
Interesting...  what _is_ the internal name, and what is it used for?  Anything at all?  It looks like it's (as the name implies) pretty much internal...

Looking around, it seems like this is wholly freeform and can contain any data at all.  So while the text says that:

> this string should be the original filename, without extension...

it says that is true:

> If the file has no internal name...

and it doesn't seem to place restrictions on what the "internal name" is.

I hate changing things that are working, poorly documented and that I don't understand, and this is definitely all three.

Is this causing you a problem?  Will changing this cause other people problems?

csware:
I don't experience problems, I just found out that we did not comply to the rules stated by Microsoft for this entry.

ethomson:
I'm not sure that I agree with your interpretation of the documentation - it makes it sound as if this is guidance if there is no internal name:

>  If the file has no internal name, this string should be the original filename, without extension. 

*However*, it does not say that the internal name cannot be the name of our output file (with extension).  I feel like this documentation is sufficiently vague *and* there is no problem with the current configuration.  I'm concerned that there may be other problems with making a change to the internal name that I cannot anticipate (eg, maybe something, somewhere relies upon this that I'm not aware of).  So my preference is to *not* change this unless the current setup causes some problem or it explicitly violates a specification.  (I feel that this documentation is not clear enough that our current internal name is somehow problematic.)

pks-t:
I'm closing this PR for now given @ethomson's feedback. Please feel free to reopen it in case you see further need to discuss the issue at hand.

Issue #4355, opened 2017-09-19
git_reference_peel() does not return the tag object for packed tags


mark-adams:
### Reproduction steps
1. Create a repository with a packed tag
```
#!/bin/bash
set -e

# Create a repo
rm -rf testrepo
mkdir testrepo
cd testrepo
git init

# Make a commit
echo "Example file" > testfile
git add testfile
git commit -m "Initial commit"

# Add an annotated tag and a lightweight tag
git tag packed-annotate -m "Adding a packed tag"
git pack-refs --all

# Add another tag (which will remain unpacked)
git tag unpacked-annotate -m "Adding an unpacked tag"
```
2. Attempt to peel the packed tag with `git_reference_peel`

```
#include <git2.h>
#include <stdio.h>

void check_tag(git_repository *repo, const char *refpath){
    int err;
    git_reference *ref;
    git_object *obj;
    
    git_reference_lookup(&ref, repo, refpath);
    err = git_reference_peel(&obj, ref, GIT_OBJ_TAG);
    printf("git_reference_peel(%s) returned %d\n", refpath, err);
}

int main(int argc, const char * argv[]) {
    git_libgit2_init();
    
    git_repository *repo;
    git_repository_open(&repo, "/tmp/testrepo");
    
    check_tag(repo, "refs/tags/unpacked-annotate");
    check_tag(repo, "refs/tags/packed-annotate");
    
    git_libgit2_shutdown();
}
```

### Expected behavior
`git_reference_peel(repo, ..., GIT_OBJ_TAG)` should return the tag object for any annotated tag regardless of whether it is packed or unpacked.

Expected test program output:
```
git_reference_peel(refs/tags/unpacked-annotate) returned 0
git_reference_peel(refs/tags/packed-annotate) returned 0
```

### Actual behavior
`git_reference_peel(repo, ..., GIT_OBJ_TAG)` returns an error for annotated tags that have been packed.

```
git_reference_peel(refs/tags/unpacked-annotate) returned 0
git_reference_peel(refs/tags/packed-annotate) returned -12
```

### Version of libgit2 (release number or SHA1)
0.26

### Operating system(s) tested
macos 10.12.6

pks-t:
Oh, how I love bug reports like these. Thanks for that wonderful reproducer, I can reproduce the issue locally. I'm debugging that right now.

pks-t:
Could you please test if #4367 fixes this issue?

ethomson:
[...]

PR #4367, opened 2017-10-06
refs: do not use peeled OID if peeling to a tag

pks-t:
If a reference stored inside of the packed-refs file does not directly
point to a commit object, it will always have a fully-peeled value
pointing to the underlying commit object. If we peel a reference to an
object, we will use that peeled value to speed up resolving the object.
This works as intended for all objects, except for annotated tags.

As a reference for an annotated tag does not directly point to the
commit but instead to the tag object, the packed-refs file will have an
accomodating fully-peeled object ID pointing to the commit of that tag.
When we use that fully-peeled OID pointing to that commit, though, we
obviously cannot peel that to the tag anymore.

Fix this issue by not using the fully-peeled OID whenever we want to
peel to a tag. Note that this does not include the case where we want to
resolve to _any_ object type. Existing code may make use from the fact
that we resolve those to commit objects instead of tag objects, even
though that behaviour is inconsistent between packed and loose
references. Furthermore, some tests of ours make the assumption that we
in fact resolve those references to a commit.

### DIFF
--- a/src/refs.c
+++ b/src/refs.c
@@ -1360,7 +1360,13 @@ int git_reference_peel(
 			return peel_error(error, ref, "Cannot resolve reference");
 	}
 
-	if (!git_oid_iszero(&resolved->peel)) {
+	/*
+	 * If we try to peel an object to a tag, we cannot use
+	 * the fully peeled object, as that will always resolve
+	 * to a commit. So we only want to use the peeled value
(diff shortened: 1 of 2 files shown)


pks-t:
Fixes #4355

ethomson:
Thanks!

PR #4368, opened 2017-10-06
transports: smart: fix memory leak when skipping symbolic refs

pks-t:
When we setup the revision walk for negotiating references with a
remote, we iterate over all references, ignoring tags and symbolic
references. While skipping over symbolic references, we forget to free
the looked up reference, resulting in a memory leak when the next
iteration simply overwrites the variable.

Fix that issue by more clearly scoping the leaking variable to the loop
and freeing it after every iteration.

### DIFF
--- a/src/transports/smart_protocol.c
+++ b/src/transports/smart_protocol.c
@@ -273,7 +273,7 @@ static int fetch_setup_walk(git_revwalk **out, git_repository *repo)
 	git_revwalk *walk = NULL;
 	git_strarray refs;
 	unsigned int i;
-	git_reference *ref;
+	git_reference *ref = NULL;
 	int error;
 
 	if ((error = git_reference_list(&refs, repo)) < 0)
@@ -285,6 +285,9 @@ static int fetch_setup_walk(git_revwalk **out, git_repository *repo)
 	git_revwalk_sorting(walk, GIT_SORT_TIME);
 
 	for (i = 0; i < refs.count; ++i) {
+		git_reference_free(ref);
+		ref = NULL;
+
 		/* No tags */
 		if (!git__prefixcmp(refs.strings[i], GIT_REFS_TAGS_DIR))
 			continue;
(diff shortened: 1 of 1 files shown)


src/transports/smart_protocol.c
```diff
@@ -285,20 +284,26 @@ static int fetch_setup_walk(git_revwalk **out, git_repository *repo)
...
 
 		if (git_reference_type(ref) == GIT_REF_SYMBOLIC)
-			continue;
+			goto on_continue;
 
 		if ((error = git_revwalk_push(walk, git_reference_target(ref))) < 0)
-			goto on_error;
+			goto on_continue;
 
+on_continue:
```
carlosmn: This label name is misleading, as we don't just go here during continue, but for each iteration.

carlosmn:
How about performing a free at the beginning of the loop plus at the end of the function? That should cover both cases without the need for a hand-rolled `ensure` at the end of the loop.

src/transports/smart_protocol.c
pks-t: Yeah, I had some problems here finding a proper name

src/transports/smart_protocol.c
```diff
@@ -285,20 +284,26 @@ static int fetch_setup_walk(git_revwalk **out, git_repository *repo)
 	git_revwalk_sorting(walk, GIT_SORT_TIME);
 
 	for (i = 0; i < refs.count; ++i) {
+		git_reference *ref;
```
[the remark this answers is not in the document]
pks-t: True

pks-t:
If we do free at the beginning of the loop, we'd require three different locations where we free the reference: at the beginning of the loop, after the loop and on success. That's really painful, in my opinion, which is why I went with the thing at the end of the loop.

pks-t:
Ah, we're actually able to collapse the two different out paths. So this version looks good to me.

ethomson:
Thanks, @pks-t, this is much easier to read.

PR #4369, opened 2017-10-06
Checkout typechange-only deltas

ethomson:
When performing a forced checkout, treat files as modified when the workdir or the index is identical *except for the mode*.  This ensures that force checkout will update the mode to the target.

Without this change, we examine the contents of the file on-disk and determine that it's identical to the tree we're checking out, and continue without making any changes.  By including the filemode in the difference check, checkout will now make a file executable (or remove the executable bit) as appropriate to match the checkout target.

We apply this check for regular files only: in the case where one of the items was a file and the other item was *not*, then this would be a typechange (not a file mode change) and would be handled by other codepaths.

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -159,6 +159,11 @@ GIT_INLINE(bool) is_workdir_base_or_new(
 		git_oid__cmp(&newitem->id, workdir_id) == 0);
 }
 
+GIT_INLINE(bool) is_file_mode_changed(git_filemode_t a, git_filemode_t b)
+{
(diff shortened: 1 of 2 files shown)


ethomson:
I cherry-picked these from #3839.  I think that there's more work to do there, but this is nice and isolated and we should include it.

/cc @cjhoward92

pks-t:
Looks good to me, thanks @ethomson. I've verified locally that the test actually catches the chmodded file with `git_status_list_new`.

PR #4372, opened 2017-10-09
cmake: fix linking in Xcode with object libraries only

### LINKED ISSUE #4352: Missing libgit2.dylib
It seems the recent CMake change had a pretty big impact on Xcode after all, there are no libgit2.dylib created after building the git2 target.

Here's the gist of a run, trying to compile the status example : https://gist.github.com/tiennou/b935b3cb00ce318eb11a13f3553a93fd

Strangely, running `cmake --build .` works, but Xcode/`xcodebuild` will fail to link any dependent targets because of the missing library. Also, the standard Darwin symlinks are there, there's just no `libgit2.0.26.0.dylib`.

And clar runs fine, which is why I missed it (shame on me).

### Reproduction steps
```
mkdir build && cd build
cmake -GXcode -DBUILD_EXAMPLES=ON ..
[build the ALL_BUILD target in Xcode]
```

### Expected behavior

```
$ find . -name *.dylib -print
./Debug/libgit2.0.26.0.dylib
./Debug/libgit2.26.dylib
./Debug/libgit2.dylib
```

### Actual behavior

```
$ find . -name *.dylib -print
./Debug/libgit2.26.dylib
./Debug/libgit2.dylib
```

212da30dbf18924cb8878348657a048698856be3

macOS 10.12.6, Xcode 8.2.1

pks-t:
CMake is unable to generate a correct Xcode project when trying to link
libraries with only object libraries as its input. As our new build
infrastructure makes heavy use of object libraries now, this affects our
libgit2 library target, as well, leading to linking errors.

Fix the issue by adding a dummy file to the libgit2 objects. As we
always have the "features.h" header ready which contains defines only,
we can simply link it into the resulting library without any effect
whatsoever. This fixes building with Xcode.

---

Thanks to @tiennou. This should fix #4352.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -2,7 +2,10 @@ IF(DEBUG_POOL)
 	SET(GIT_DEBUG_POOL 1)
 ENDIF()
 
-SET(LIBGIT2_OBJECTS "")
+# Add the features.h file as a dummy. This is required for Xcode
+# to successfully build the libgit2 library when using only
+# object libraries.
+SET(LIBGIT2_OBJECTS "${CMAKE_CURRENT_BINARY_DIR}/git2/sys/features.h")
 
 # This variable will contain the libraries we need to put into
 # libgit2.pc's Requires.private. That is, what we're linking to or


ethomson:
Cool, thanks @tiennou and @pks-t .

PR #4373, opened 2017-10-12
example-log: add support for --log-size

cjhoward92:
Updated the `log` example to use the `--log-size` parameter.

### DIFF
--- a/examples/log.c
+++ b/examples/log.c
@@ -49,7 +49,8 @@ static int add_revision(struct log_state *s, const char *revstr);
 
 /** log_options holds other command line options that affect log output */
 struct log_options {
-	int show_diff, show_log_size;
+	int show_diff;
+	int show_log_size;
 	int skip, limit;
 	int min_parents, max_parents;
 	git_time_t before;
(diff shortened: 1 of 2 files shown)


examples/log.c
```diff
@@ -63,7 +63,7 @@ struct log_options {
 static int parse_options(
 	struct log_state *s, struct log_options *opt, int argc, char **argv);
 static void print_time(const git_time *intime, const char *prefix);
-static void print_commit(git_commit *commit);
+static void print_commit(git_commit *commit, int show_log_size);
```
mikeando: It would feel more logical to me to pass a pointer to the  `log_options` struct here.

examples/log.c
```diff
@@ -49,7 +49,7 @@ static int add_revision(struct log_state *s, const char *revstr);
 
 /** log_options holds other command line options that affect log output */
 struct log_options {
-	int show_diff;
+	int show_diff, show_log_size;
```
mikeando: Maybe a new line for this?

examples/log.c
```diff
@@ -347,6 +347,11 @@ static void print_commit(git_commit *commit)
 	git_oid_tostr(buf, sizeof(buf), git_commit_id(commit));
 	printf("commit %s\n", buf);
 
+	if (show_log_size) {
+		printf("log size %d", (int)strlen(git_commit_message(commit)));
+		printf("\n");
```
mikeando: Why not just one printf call?

~~~
printf("log size %d\n", (int)strlen(git_commit_message(commit)));
~~~

cjhoward92:
@mikeando Thanks for the review and feedback! Hopefully my changes are up to snuff!

carlosmn:
The format matches git and the implementation looks straightforward. Thanks!

PR #4374, opened 2017-10-13
Pack file verification

pks-t:
Warning: this is ugly and, at least to me, in many places it feels like I'm doing hacks over hacks to work around our object system. The main problem here is that we currently cannot parse objects which are not owned by the ODB system, because our objects point into the reference-counted memory of the ODB. So when I try to parse an object which is not owned by the ODB, then this will crash later on after `git_odb_object_deref`, as it tries to free objects which aren't owned by itself or not even part of the heap. So yeah, this doesn't look as nice as it could in an ideal world.

On the other hand, this seems to work just fine right now. After nearly a whole day of cursing and debugging I think I've finally got it right. So what do I do? This is mostly the implementation of `git index-pack --strict`, doing two things:

1. check whether all objects of a pack can be parsed correctly
2. check all references to other objects for commits, tags and trees

What this gets us is that we can verify that a pack is complete (we've got all objects such that we can resolve the complete graph) when we receive the pack.

I've discussed the design a bit in Slack with @carlosmn. My first thought was to just perform an object walk after fetching the complete pack file. But seeing that in most cases we're limited by network bandwidth while fetching the pack file, we agreed that it would be much nicer to just do as much of the heavy lifting as possible during retrieval of the pack. So the algorithm works as follows:

1. for every non-deltified object passed to the indexer
    * strike its object ID from `expected_oids`, which keeps track of OIDs we still expect to find as part of the pack file
    * parse the object
    * add object IDs referenced by the object to the `expected_oids`
2. when resolving delta objects, do the same with regards to the `expected_oids` map
3. afterwards, check that the `expected_oids` map has no entries anymore, as all object references should now be resolved correctly

I'm just putting this up early to get early feedback from the CI and reviewers. I'm not happy with some of the things I had to do here, even though they work.

### DIFF
--- a/examples/network/index-pack.c
+++ b/examples/network/index-pack.c
@@ -46,7 +46,7 @@ int index_pack(git_repository *repo, int argc, char **argv)
 		return EXIT_FAILURE;
 	}
 
-	if (git_indexer_new(&idx, ".", 0, NULL, NULL, NULL) < 0) {
+	if (git_indexer_new(&idx, ".", 0, NULL, NULL) < 0) {
 		puts("bad idx");
 		return -1;
 	}
--- a/include/git2/indexer.h
+++ b/include/git2/indexer.h
@@ -15,6 +15,30 @@ GIT_BEGIN_DECL
 
 typedef struct git_indexer git_indexer;
 
+typedef struct git_indexer_options {
+	unsigned int version;
+
+	/** progress_cb function to call with progress information */
+	git_transfer_progress_cb progress_cb;
+	/** progress_cb_payload payload for the progress callback */
+	void *progress_cb_payload;
+} git_indexer_options;
+
+#define GIT_INDEXER_OPTIONS_VERSION 1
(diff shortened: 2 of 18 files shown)


pks-t:
Oh, and to remind myself: I leak memory in the `expected_oids` right now when it has entries when freeing the indexer.

pks-t:
Yay, this is green now. I've fixed a bug last week where I accessed the object's type through the indexer instead of through the raw object. This has lead to a subtle bug and invalid memory access that could occur seemingly only in very special situations.

pks-t:
Rebased on latest master

pks-t:
Rebased to fix conflicts

ethomson:
I've read through this a few times now and I think it's 👍 .  There's a lot going on but it all looks correct to me and it's definitely valuable.

@carlosmn you're more familiar with this original code than I am.  Any thoughts before we :shipit: ?

ethomson:
Okay, this has been open for a while with no additional comments.  I'm going to :shipit:

PR #4375, opened 2017-10-14
Fix compilation for CMake versions 3.0.x where x >= 1

mhagger:
Apparently policy CMP0015 was added in CMake 3.1. With CMake 3.0.2, the build was failing with

    CMake Error at CMakeLists.txt:18 (CMAKE_POLICY):
      Policy "CMP0051" is not known to this version of CMake.

This patch makes it work.

I know almost nothing about CMake, so this might be totally bogus.


pks-t:
The CI is having hiccups once again. It got creative this time and is erroring out due to being unable to get our APT signing key, but only on a few nodes. But I'm ignoring those issues for now -- this PR here is correct, it captures what I actually wanted to say here. So thanks @mhagger for your fix!

PR #4378, opened 2017-10-16
submodule: check index for path and prefix before adding submodule

cjhoward92:
This is to fix an issue where libgit2 will allow you to add a submodule even if the path you are trying add is found on the index. In the CLI we get a nice error about this, but not in libgit2.

I would obviously like some pointers on how to do this, as I do not think it is 100% correct. For one, when I add the trailing slash to the path, I feel like I could clean that up or have a better implementation. I am also unsure if there is a more streamlined routine to check the index for the path.

Some advice on how I could clean this up would be awesome.

### DIFF
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -153,11 +153,12 @@ static int find_by_path(const git_config_entry *entry, void *payload)
  * Checks to see if the submodule shares its name with a file or directory that
  * already exists on the index. If so, the submodule cannot be added.
  */
-static int can_add_submodule(git_repository *repo, const char *path)
+static int is_path_occupied(bool *occupied, git_repository *repo, const char *path)
 {
 	int error = 0;
 	git_index *index;
 	git_buf dir = GIT_BUF_INIT;
+	*occupied = false;
 
 	if ((error = git_repository_index__weakptr(&index, repo)) < 0)
 		goto out;
@@ -165,10 +166,9 @@ static int can_add_submodule(git_repository *repo, const char *path)
 	if ((error = git_index_find(NULL, index, path)) == 0) {
 		giterr_set(GITERR_SUBMODULE,
 			"File '%s' already exists in the index", path);
-		error = GIT_EEXISTS;
+		*occupied = true;
 		goto out;
 	}
-	error = 0;
 
 	if ((error = git_buf_sets(&dir, path)) < 0)
 		goto out;
@@ -179,8 +179,7 @@ static int can_add_submodule(git_repository *repo, const char *path)
 	if ((error = git_index_find_prefix(NULL, index, dir.ptr)) == 0) {
 		giterr_set(GITERR_SUBMODULE,
 			"Directory '%s' already exists in the index", path);
-		error = GIT_EEXISTS;
-		goto out;
+		*occupied = true;
 	}
 	error = 0;
 
@@ -704,6 +703,7 @@ int git_submodule_add_setup(
 	git_submodule *sm = NULL;
 	git_buf name = GIT_BUF_INIT, real_url = GIT_BUF_INIT;
 	git_repository *subrepo = NULL;
+	bool path_occupied;
 
 	assert(repo && url && path);
 
@@ -728,8 +728,13 @@ int git_submodule_add_setup(
 		goto cleanup;
 	}
 
-	if ((error = can_add_submodule(repo, path)) < 0)
+	if ((error = is_path_occupied(&path_occupied, repo, path)) < 0)
 		goto cleanup;
+
+	if (path_occupied) {
+		error = GIT_EEXISTS;
+		goto cleanup;
+	}
 
 	/* update .gitmodules */
 
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -149,6 +149,41 @@ static int find_by_path(const git_config_entry *entry, void *payload)
 	return 0;
 }
 
+static int can_add_submodule(git_repository *repo, const char *path) {
+	int error;
+	git_index *index;
+	git_buf dir = GIT_BUF_INIT;
+
+	if ((error = git_buf_sets(&dir, path)) < 0)
+		return error;
+
+	if ((error = git_path_to_dir(&dir)) < 0)
+		return error;
+
+	/* get the index for the repo */
+
+	if ((error = git_repository_index__weakptr(&index, repo)) < 0)
+		return error;
+
+	/* see if the submodule name exists as a file on the index */
+
+	if ((error = git_index_find(NULL, index, path)) == 0) {
+		giterr_set(GITERR_SUBMODULE,
+			"'%s' already exists in the index", path);
+		return GIT_EEXISTS;
+	}
+
+	/* see if the submodule name exists as a directory on the index */
+
+	if ((error = git_index_find_prefix(NULL, index, dir.ptr)) == 0) {
+		giterr_set(GITERR_SUBMODULE,
+			"'%s' already exists in the index", path);
+		return GIT_EEXISTS;
+	}
+
+	return 0;
+}
+
 /**
  * Release the name map returned by 'load_submodule_names'.
  */
@@ -660,10 +695,7 @@ int git_submodule_add_setup(
 	int use_gitlink)
 {
 	int error = 0;
-	size_t path_len;
-	const char *dir;
 	git_config_backend *mods = NULL;
-	git_index *index;
 	git_submodule *sm = NULL;
 	git_buf name = GIT_BUF_INIT, real_url = GIT_BUF_INIT;
 	git_repository *subrepo = NULL;
@@ -691,33 +723,8 @@ int git_submodule_add_setup(
 		goto cleanup;
 	}
 
-	/* get the index for the repo */
-
-	if ((error = git_repository_index__weakptr(&index, repo)) < 0)
+	if ((error = can_add_submodule(repo, path)) < 0)
 		goto cleanup;
-
-	/* see if the submodule name exists as a file on the index */
-
-	if ((error = git_index_find(NULL, index, path)) == 0) {
-		giterr_set(GITERR_SUBMODULE,
-			"'%s' already exists in the index", path);
-		return GIT_EEXISTS;
-	}
-
-	/* We need the path to end with '/' so we can check it as a directory prefix */
-
-	path_len = strlen(path);
-	dir = git__malloc(path_len + 1);
-	strcpy(dir, path);
-	git_path_string_to_dir(dir, path_len + 1);
-
-	/* see if the submodule name exists as a directory on the index */
-
-	if ((error = git_index_find_prefix(NULL, index, dir)) == 0) {
-		giterr_set(GITERR_SUBMODULE,
-			"'%s' already exists in the index", path);
-		return GIT_EEXISTS;
-	}
 
 	/* update .gitmodules */
 
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -176,7 +176,10 @@ static int is_path_occupied(bool *occupied, git_repository *repo, const char *pa
 	if ((error = git_path_to_dir(&dir)) < 0)
 		goto out;
 
-	if ((error = git_index_find_prefix(NULL, index, dir.ptr)) == 0) {
+	if ((error = git_index_find_prefix(NULL, index, dir.ptr)) < 0 && error != GIT_ENOTFOUND)
+    goto out;
+
+	if (!error) {
 		giterr_set(GITERR_SUBMODULE,
 			"Directory '%s' already exists in the index", path);
 		*occupied = true;
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -149,39 +149,44 @@ static int find_by_path(const git_config_entry *entry, void *payload)
 	return 0;
 }
 
-static int can_add_submodule(git_repository *repo, const char *path) {
-	int error;
+/*
+ * Checks to see if the submodule shares its name with a file or directory that
+ * already exists on the index. If so, the submodule cannot be added.
+ */
+static int can_add_submodule(git_repository *repo, const char *path)
+{
+	int error = 0;
 	git_index *index;
 	git_buf dir = GIT_BUF_INIT;
 
-	if ((error = git_buf_sets(&dir, path)) < 0)
-		return error;
-
-	if ((error = git_path_to_dir(&dir)) < 0)
-		return error;
-
-	/* get the index for the repo */
-
 	if ((error = git_repository_index__weakptr(&index, repo)) < 0)
-		return error;
-
-	/* see if the submodule name exists as a file on the index */
+		goto out;
 
 	if ((error = git_index_find(NULL, index, path)) == 0) {
 		giterr_set(GITERR_SUBMODULE,
-			"'%s' already exists in the index", path);
-		return GIT_EEXISTS;
+			"File '%s' already exists in the index", path);
+		error = GIT_EEXISTS;
+		goto out;
 	}
+	error = 0;
 
-	/* see if the submodule name exists as a directory on the index */
+	if ((error = git_buf_sets(&dir, path)) < 0)
+		goto out;
+
+	if ((error = git_path_to_dir(&dir)) < 0)
+		goto out;
 
 	if ((error = git_index_find_prefix(NULL, index, dir.ptr)) == 0) {
 		giterr_set(GITERR_SUBMODULE,
-			"'%s' already exists in the index", path);
-		return GIT_EEXISTS;
+			"Directory '%s' already exists in the index", path);
+		error = GIT_EEXISTS;
+		goto out;
 	}
+	error = 0;
 
-	return 0;
+out:
+	git_buf_free(&dir);
+	return error;
 }
 
 /**
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -660,7 +660,10 @@ int git_submodule_add_setup(
 	int use_gitlink)
 {
 	int error = 0;
+	size_t path_len;
+	const char *dir;
 	git_config_backend *mods = NULL;
+	git_index *index;
 	git_submodule *sm = NULL;
 	git_buf name = GIT_BUF_INIT, real_url = GIT_BUF_INIT;
 	git_repository *subrepo = NULL;
@@ -686,6 +689,34 @@ int git_submodule_add_setup(
 		giterr_set(GITERR_SUBMODULE, "submodule path must be a relative path");
 		error = -1;
 		goto cleanup;
+	}
+
+	/* get the index for the repo */
+
+	if ((error = git_repository_index__weakptr(&index, repo)) < 0)
+		goto cleanup;
+
+	/* see if the submodule name exists as a file on the index */
+
+	if ((error = git_index_find(NULL, index, path)) == 0) {
+		giterr_set(GITERR_SUBMODULE,
+			"'%s' already exists in the index", path);
+		return GIT_EEXISTS;
+	}
+
+	/* We need the path to end with '/' so we can check it as a directory prefix */
+
+	path_len = strlen(path);
+	dir = git__malloc(path_len + 1);
+	strcpy(dir, path);
+	git_path_string_to_dir(dir, path_len + 1);
+
+	/* see if the submodule name exists as a directory on the index */
+
+	if ((error = git_index_find_prefix(NULL, index, dir)) == 0) {
+		giterr_set(GITERR_SUBMODULE,
+			"'%s' already exists in the index", path);
+		return GIT_EEXISTS;
 	}
 
 	/* update .gitmodules */
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -163,10 +163,12 @@ static int is_path_occupied(bool *occupied, git_repository *repo, const char *pa
 	if ((error = git_repository_index__weakptr(&index, repo)) < 0)
 		goto out;
 
-	if ((error = git_index_find(NULL, index, path)) == 0) {
-		giterr_set(GITERR_SUBMODULE,
-			"File '%s' already exists in the index", path);
-		*occupied = true;
+	if ((error = git_index_find(NULL, index, path)) != GIT_ENOTFOUND) {
+		if (!error) {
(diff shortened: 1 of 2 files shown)


src/submodule.c
```diff
@@ -688,6 +691,34 @@ int git_submodule_add_setup(
 		goto cleanup;
 	}
 
+	/* get the index for the repo */
```
pks-t: It would be nice to put the following checks into a simple function. Something like `can_add_submodule(repo, path)`.

src/submodule.c
```diff
@@ -688,6 +691,34 @@ int git_submodule_add_setup(
...
+			"'%s' already exists in the index", path);
+		return GIT_EEXISTS;
+	}
+
+	/* We need the path to end with '/' so we can check it as a directory prefix */
+
+	path_len = strlen(path);
+	dir = git__malloc(path_len + 1);
+	strcpy(dir, path);
+	git_path_string_to_dir(dir, path_len + 1);
```
pks-t: You could use a buffer here with `git_path_to_dir`:

```
git_buf path = GIT_BUF_INIT;
git_buf_sets(&path, path);
git_path_to_dir(&path);
```

Error handling is obviously missing.

tests/submodule/add.c
```diff
@@ -128,3 +128,69 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+	git_index *index;
+	git_submodule *sm;
+	git_buf dirname = GIT_BUF_INIT;
+	git_buf filename = GIT_BUF_INIT;
+	FILE *fd;
+
+	/* In this repo, HEAD (master) has no remote tracking branc h*/
+	g_repo = cl_git_sandbox_init("testrepo");
+
+	git_buf_joinpath(&dirname, git_repository_workdir(g_repo), "TestGitRepository");
```
pks-t: Please wrap every call which you expect to just return `0` for success in `cl_git_pass`.

tests/submodule/add.c
```diff
@@ -128,3 +128,69 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+	git_buf_joinpath(&filename, dirname.ptr, "test.txt");
+
+	p_mkdir(dirname.ptr, 0700);
+	fd = fopen(filename.ptr, "w");
+	fclose(fd);
+
+	cl_git_pass(
+		git_repository_index__weakptr(&index, g_repo)
+	);
```
pks-t: There's no need for the newlines here. In most cases, we simply ignore line length limits for our tests and put those things in one line.

tests/submodule/add.c
```diff
@@ -128,3 +128,69 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+
+	/* In this repo, HEAD (master) has no remote tracking branc h*/
+	g_repo = cl_git_sandbox_init("testrepo");
+
+	git_buf_joinpath(&dirname, git_repository_workdir(g_repo), "TestGitRepository");
+	git_buf_joinpath(&filename, dirname.ptr, "test.txt");
+
+	p_mkdir(dirname.ptr, 0700);
+	fd = fopen(filename.ptr, "w");
+	fclose(fd);
```
pks-t: You should also assert that those three functions return expected values.

tests/submodule/add.c
```diff
@@ -128,3 +128,69 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+		git_index_add_bypath(index, "TestGitRepository/test.txt")
+	);
+
+	cl_git_fail_with(
+		git_submodule_add_setup(&sm, g_repo, "./", "TestGitRepository", 1),
+		GIT_EEXISTS
+	);
+
+	git_buf_free(&dirname);
+	git_buf_free(&filename);
```
pks-t: You have to free the submodule, as well.

tests/submodule/add.c
```diff
@@ -128,3 +128,69 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+	FILE *fd;
+
+	/* In this repo, HEAD (master) has no remote tracking branc h*/
+	g_repo = cl_git_sandbox_init("testrepo");
+
+	git_buf_joinpath(&dirname, git_repository_workdir(g_repo), "TestGitRepository");
+	git_buf_joinpath(&filename, dirname.ptr, "test.txt");
+
+	p_mkdir(dirname.ptr, 0700);
+	fd = fopen(filename.ptr, "w");
```
pks-t: We have `cl_git_mkfile` for that

tests/submodule/add.c
```diff
@@ -128,3 +128,69 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+{
+	git_index *index;
+	git_submodule *sm;
+	git_buf name = GIT_BUF_INIT;
+	FILE *fd;
+
+	/* In this repo, HEAD (master) has no remote tracking branc h*/
+	g_repo = cl_git_sandbox_init("testrepo");
+
+	git_buf_joinpath(&name, git_repository_workdir(g_repo), "TestGitRepository");
```
pks-t: Please assert the return value here.

tests/submodule/add.c
```diff
@@ -128,3 +128,69 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+
+	fd = fopen(name.ptr, "w");
+	fclose(fd);
```
pks-t: We have `cl_git_mkfile` for that.

tests/submodule/add.c
```diff
@@ -128,3 +128,69 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+		git_index_add_bypath(index, "TestGitRepository")
+	);
+
+	cl_git_fail_with(
+		git_submodule_add_setup(&sm, g_repo, "./", "TestGitRepository", 1),
+		GIT_EEXISTS
+	);
+
+	git_buf_free(&name);
+}
```
pks-t: Again, you also have to free the submodule here

src/submodule.c
pks-t: The code is obviously bogus, I should've chosen two different variable names. But I guess you get what I mean :P

cjhoward92:
@pks-t Also thank you for reviewing this. I will make these changes as soon as I can! I appreciate you taking the time to look over my work.

cjhoward92:
@pks-t Updated. Let me know what you think!

cjhoward92:
Any chance anyone will take a look at this?

src/submodule.c
```diff
@@ -149,6 +149,41 @@ static int find_by_path(const git_config_entry *entry, void *payload)
 	return 0;
 }
 
+static int can_add_submodule(git_repository *repo, const char *path) {
```
pks-t: Please move the opening curly brace to a new line.

src/submodule.c
```diff
@@ -149,6 +149,41 @@ static int find_by_path(const git_config_entry *entry, void *payload)
...
+	git_index *index;
+	git_buf dir = GIT_BUF_INIT;
+
+	if ((error = git_buf_sets(&dir, path)) < 0)
+		return error;
+
+	if ((error = git_path_to_dir(&dir)) < 0)
+		return error;
+
+	/* get the index for the repo */
```
pks-t: Would be nice if you removed those comments. You shouldn't describe _what_ you are doing with comments, as most often that is obvious from the code itself. Instead, you should try to document _why_ you are doing that. E.g. a useful comment here would be one preceding the function which explains exactly when you can add a submodule and when you can't.

src/submodule.c
```diff
@@ -149,6 +149,41 @@ static int find_by_path(const git_config_entry *entry, void *payload)
...
 
+static int can_add_submodule(git_repository *repo, const char *path) {
+	int error;
+	git_index *index;
+	git_buf dir = GIT_BUF_INIT;
+
+	if ((error = git_buf_sets(&dir, path)) < 0)
+		return error;
+
+	if ((error = git_path_to_dir(&dir)) < 0)
```
pks-t: You start to leak memory as soon as one of those functions returns an error. Instead of "return error;", we usually have e.g. a "goto error;" where all memory is free'd. E.g.:

```
static int can_add_submodule(git_repository *repo, const char *path)
{
...

if ((error = git_path_to_dir(&dir)) < 0)
    goto out;

...

out:
    git_buf_free(&dir);
    return error;
}
```

src/submodule.c
```diff
@@ -149,6 +149,41 @@ static int find_by_path(const git_config_entry *entry, void *payload)
...
+		return error;
+
+	/* get the index for the repo */
+
+	if ((error = git_repository_index__weakptr(&index, repo)) < 0)
+		return error;
+
+	/* see if the submodule name exists as a file on the index */
+
+	if ((error = git_index_find(NULL, index, path)) == 0) {
```
pks-t: I think there's something off here. You first normalize `path` to a directory path by using `git_path_to_dir` and now you try to check if there's a file at that location matching the path. I guess what you wanted to do instead is first using the "normal" `path` value and only normalize it to a directory name after you've used `git_index_find` here to check whether there's a file there.

src/submodule.c
```diff
@@ -149,6 +149,41 @@ static int find_by_path(const git_config_entry *entry, void *payload)
...
+		giterr_set(GITERR_SUBMODULE,
+			"'%s' already exists in the index", path);
+		return GIT_EEXISTS;
+	}
+
+	/* see if the submodule name exists as a directory on the index */
+
+	if ((error = git_index_find_prefix(NULL, index, dir.ptr)) == 0) {
+		giterr_set(GITERR_SUBMODULE,
+			"'%s' already exists in the index", path);
```
pks-t: Would be cool if you disambiguated those two error messages based on their type (that is directory vs file already existing).

src/submodule.c
```diff
@@ -688,6 +723,9 @@ int git_submodule_add_setup(
 		goto cleanup;
 	}
 
+	if ((error = can_add_submodule(repo, path)) < 0)
```
pks-t: There's a space missing in front of the first opening brace.

tests/submodule/add.c
```diff
@@ -128,3 +128,51 @@ void test_submodule_add__url_relative_to_workdir(void)
...
 }
+
+void test_submodule_add__path_exists_in_index(void)
+{
+	git_index *index;
+	git_submodule *sm;
+	git_buf dirname = GIT_BUF_INIT;
+	git_buf filename = GIT_BUF_INIT;
+
+	/* In this repo, HEAD (master) has no remote tracking branc h*/
```
pks-t: There's a typo. Also: why does this matter for our particular unit test here? I don't want to propose it doesn't, but for myself I don't get why this is interesting. Please either extend that comment and explain a bit more why this is important here or just remove it :)

tests/submodule/add.c
```diff
@@ -128,3 +128,51 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+	git_buf_free(&filename);
+}
+
+void test_submodule_add__file_exists_in_index(void)
+{
+	git_index *index;
+	git_submodule *sm;
+	git_buf name = GIT_BUF_INIT;
+
+	/* In this repo, HEAD (master) has no remote tracking branc h*/
```
pks-t: Same as above

tests/submodule/add.c
```diff
@@ -128,3 +128,51 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+{
+	git_index *index;
+	git_submodule *sm;
+	git_buf dirname = GIT_BUF_INIT;
+	git_buf filename = GIT_BUF_INIT;
+
+	/* In this repo, HEAD (master) has no remote tracking branc h*/
+	g_repo = cl_git_sandbox_init("testrepo");
+
+	cl_git_pass(git_buf_joinpath(&dirname, git_repository_workdir(g_repo), "TestGitRepository"));
```
pks-t: I'd rename "TestGitRepository" to e.g. just "subdirectory" or "directory". Otherwise it's a bit confusing, as that path should not represent a real git repository but should only contain a file later on.

tests/submodule/add.c
```diff
@@ -128,3 +128,51 @@ void test_submodule_add__url_relative_to_workdir(void)
 
 	assert_submodule_url("TestGitRepository", git_repository_workdir(g_repo));
 }
+
+void test_submodule_add__path_exists_in_index(void)
```
pks-t: There's one problem with that test. You want to verify that creation of a submodule fails if the path exists in the index, but actually when you create your submodule it exists both inside the index and also inside of the working directory. So us failing here might not be the result from the entry existing in the index, but instead that file existing in the working directory.

You could either just remove the directory and file after you've added them to the index or, which is the preferred method if you ask me, instead add the index entry directly by using `git_index_add`.

tests/submodule/add.c
```diff
@@ -128,3 +128,51 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+	cl_git_mkfile(name.ptr, "Test content");
+
+	cl_git_pass(git_repository_index__weakptr(&index, g_repo));
+	cl_git_pass(git_index_add_bypath(index, "TestGitRepository"));
+
+	cl_git_fail_with(git_submodule_add_setup(&sm, g_repo, "./", "TestGitRepository", 1), GIT_EEXISTS);
+
+	git_submodule_free(sm);
+	git_buf_free(&name);
+}
```
pks-t: Right now you're checking two out of four cases:

1. an index entry exists inside of the directory you want to add
2. a file exists in the working directory

But there are two cases missing:

1. an index entry exists directly where you want to create the submodule
2. a directory exists where you want to create the submodule

Would it make sense to add tests for those two cases?

cjhoward92:
I janked this up bad. I will recover the remote from my other machine :(

tests/submodule/add.c
cjhoward92: I get the problem you are describing, but I am unsure how to use `git_index_add`. How can I create a `git_index_entry` properly? I s there something I am missing? Other than `index_create_entry`, which is not exposed outside of `index.c` I do not know how to create an entry. For now, I am updating the routine to just delete the file and directory after I add to the index.

tests/submodule/add.c
cjhoward92: I am not sure exactly what your test cases mean. I was trying to test two cases:

 1) A file path exists in the index with the same name as the submodule
 2) A directory exists in the index with the same name as the submodule

If you could offer some clarification I would be more than happy to add extra test cases :)

cjhoward92: Do you mean we should test against the workdir as well? I am not sure that my added code does that, it is an explicit check against the index but it makes sense that we should check untracked files in the workdir too. Probably a good check to have in `can_add_submodule`. Thoughts?

src/submodule.c
cjhoward92: I'm not sure what brace you are referring to here.

tests/submodule/add.c
pks-t: Take a look at for example "tests/index/add.c", `test_add_entry()`. You can create an index entry by simply assigning the structure fields.

cjhoward92:
@pks-t I finally got around to making all of the changes you requested, I think. Let me know if there is anything else I need to do. Sorry for the wait, been a busy few months!

cjhoward92:
bump

src/submodule.c
```diff
@@ -149,6 +149,46 @@ static int find_by_path(const git_config_entry *entry, void *payload)
...
+	if ((error = git_repository_index__weakptr(&index, repo)) < 0)
+		goto out;
+
+	if ((error = git_index_find(NULL, index, path)) == 0) {
+		giterr_set(GITERR_SUBMODULE,
+			"File '%s' already exists in the index", path);
+		error = GIT_EEXISTS;
+		goto out;
+	}
+	error = 0;
```
pks-t: This is not needed here, as the next line will already unconditionally set `error`

src/submodule.c
pks-t: Huh. Looks alright to me, too. Dunno either :P

src/submodule.c
```diff
@@ -149,6 +149,46 @@ static int find_by_path(const git_config_entry *entry, void *payload)
 	return 0;
 }
 
+/*
+ * Checks to see if the submodule shares its name with a file or directory that
+ * already exists on the index. If so, the submodule cannot be added.
+ */
+static int can_add_submodule(git_repository *repo, const char *path)
```
pks-t: I think this is a bit misnamed. You do not check if we can add the submodule at a given path, as that would involve other checks as well. Instead, you only check whether the given path is already occupied by something else. How about `is_path_occupied` instead?

One more thing is that functions which should answer a certain question are expected to return a bool, not an int. So I'd propose a signature like `static int is_path_occupied(bool *occupied, git_repository *repo, const char *path)`. Like this, a real unexpected error can be returned by returing a negative value. If no error happens we return `0` and `occupied` will then be set to `true` if the path is occupied and `false` otherwise.

tests/submodule/add.c
```diff
@@ -128,3 +129,57 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+	git_submodule *sm;
+	git_buf filename = GIT_BUF_INIT;
+
+	g_repo = cl_git_sandbox_init("testrepo");
+
+	cl_git_pass(git_buf_joinpath(&filename, "subdirectory", "test.txt"));
+
+	cl_git_pass(git_repository_index(&index, g_repo));
+
+	test_add_entry(index, valid_blob_id, filename.ptr, GIT_FILEMODE_BLOB);
```
pks-t: Perfect, these tests look very nice now! :)

cjhoward92:
@pks-t Should be updated with `is_path_occupied`. Let me know what you think!

src/submodule.c
```diff
@@ -149,6 +149,45 @@ static int find_by_path(const git_config_entry *entry, void *payload)
...
+		goto out;
+	}
+
+	if ((error = git_buf_sets(&dir, path)) < 0)
+		goto out;
+
+	if ((error = git_path_to_dir(&dir)) < 0)
+		goto out;
+
+	if ((error = git_index_find_prefix(NULL, index, dir.ptr)) == 0) {
```
pks-t: You're ignoring any potential error returned by `git_index_find_prefix` other than `GIT_ENOTFOUND`. Granted, right now it will not return any other error, but you never know how it will change in the future. Maybe something like this:

```
if ((error = git_index_find_prefix(NULL, index, dir.ptr)) < 0 && error != GIT_ENOTFOUND)
    goto out;

if (!error) {
    giterr_set(GITERR_SUBMODULE, ...);
    *occupied = true;
    goto out;
}

error = 0;
```

cjhoward92:
This is also updated! Thanks again!

src/submodule.c
```diff
@@ -149,6 +149,48 @@ static int find_by_path(const git_config_entry *entry, void *payload)
...
+{
+	int error = 0;
+	git_index *index;
+	git_buf dir = GIT_BUF_INIT;
+	*occupied = false;
+
+	if ((error = git_repository_index__weakptr(&index, repo)) < 0)
+		goto out;
+
+	if ((error = git_index_find(NULL, index, path)) == 0) {
```
pks-t: I'd say it'd be a bit less fragile if the check was `if ((error = git_index_find(NULL, index, path)) != GIT_ENOTFOUND)`. Otherwise, we might just ignore all errors by `git_index_find` (even though it currently does not return any error different than `GIT_ENOTFOUND`.

src/submodule.c
```diff
@@ -149,6 +149,48 @@ static int find_by_path(const git_config_entry *entry, void *payload)
...
+	}
+
+	if ((error = git_buf_sets(&dir, path)) < 0)
+		goto out;
+
+	if ((error = git_path_to_dir(&dir)) < 0)
+		goto out;
+
+	if ((error = git_index_find_prefix(NULL, index, dir.ptr)) < 0 && error != GIT_ENOTFOUND)
+    goto out;
```
pks-t: This uses spaces instead of tabs.

src/submodule.c
```diff
@@ -149,6 +149,48 @@ static int find_by_path(const git_config_entry *entry, void *payload)
...
+		goto out;
+	}
+
+	if ((error = git_buf_sets(&dir, path)) < 0)
+		goto out;
+
+	if ((error = git_path_to_dir(&dir)) < 0)
+		goto out;
+
+	if ((error = git_index_find_prefix(NULL, index, dir.ptr)) < 0 && error != GIT_ENOTFOUND)
```
pks-t: You could probably do the same as above:

```
if ((error = git_index_find_prefix(NULL, index, dir.ptr)) != GIT_ENOTFOUND) {
    if (!error) {
        giterr_set(GITERR_SUBMODULE, "Already exists");
        *occupied = true;
        error = -1;
    }
    goto out;
}
````

File: tests/submodule/add.c [hunk 29]
```diff
@@ -128,3 +129,57 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+{
+	git_index *index;
+	git_submodule *sm;
+	git_buf filename = GIT_BUF_INIT;
+
+	g_repo = cl_git_sandbox_init("testrepo");
+
+	cl_git_pass(git_buf_joinpath(&filename, "subdirectory", "test.txt"));
+
+	cl_git_pass(git_repository_index(&index, g_repo));
```
pks-t: You should use `git_repository_index__weakptr`, otherwise you'd have to free `index`

tests/submodule/add.c
```diff
@@ -128,3 +129,57 @@ void test_submodule_add__url_relative_to_workdir(void)
...
+
+void test_submodule_add__file_exists_in_index(void)
+{
+	git_index *index;
+	git_submodule *sm;
+	git_buf name = GIT_BUF_INIT;
+
+	g_repo = cl_git_sandbox_init("testrepo");
+
+	cl_git_pass(git_repository_index(&index, g_repo));
```
pks-t: Same here

cjhoward92:
Thanks for all of the hard work. I have no problem making as many iterations as possible to get this to be perfect!

pks-t:
Looks good to me, thanks! Waiting for CI to catch up and will give others a chance to review.

pks-t:
Thanks again for this PR!

PR #4380, opened 2017-10-18
examples: ls-files: add ls-files to list paths in the index

cjhoward92:
Added an example to mimic `git ls-files` using the libgit2 library. It also supports the `--error-unmatch` parameter to determine if the specified paths are in the index. This will hopefully be a useful example for new libgit2 users.

### DIFF
--- a/examples/ls-files.c
+++ b/examples/ls-files.c
@@ -83,7 +83,7 @@ static int print_paths(ls_options *opts, git_index *index)
 {
 	size_t i;
 	const git_index_entry *entry;
-	
+
 	/* if there are not files explicitly listed by the user print all entries in the index */
 	if (opts->file_count == 0) {
 		size_t entry_count = git_index_entrycount(index);
@@ -99,14 +99,13 @@ static int print_paths(ls_options *opts, git_index *index)
 	for (i = 0; i < opts->file_count; ++i) {
 		const char *path = opts->files[i];
 
-		entry = git_index_get_bypath(index, path, GIT_INDEX_STAGE_NORMAL);
-		if (!entry && opts->error_unmatch) {
+		if ((entry = git_index_get_bypath(index, path, GIT_INDEX_STAGE_NORMAL)) != NULL) {
+			printf("%s\n", path);
+		} else if (opts->error_unmatch) {
 			fprintf(stderr, "error: pathspec '%s' did not match any file(s) known to git.\n", path);
 			fprintf(stderr, "Did you forget to 'git add'?\n");
 			return -1;
 		}
-
-		printf("%s\n", path);
 	}
 
 	return 0;
--- a/examples/ls-files.c
+++ b/examples/ls-files.c
@@ -49,7 +49,7 @@ static void usage(const char *message, const char *arg)
 	exit(1);
 }
 
-static void parse_options(ls_options *opts, int argc, char *argv[])
+static int parse_options(ls_options *opts, int argc, char *argv[])
 {
 	int parsing_files = 0;
 	struct args_info args = ARGS_INFO_INIT;
@@ -58,7 +58,7 @@ static void parse_options(ls_options *opts, int argc, char *argv[])
 	memset(opts, 0, sizeof(ls_options));
 
 	if (argc < 2)
-		return;
+		return 0;
 
 	for (args.pos = 1; args.pos < argc; ++args.pos) {
 		char *a = argv[args.pos];
@@ -77,8 +77,11 @@ static void parse_options(ls_options *opts, int argc, char *argv[])
 			opts->error_unmatch = 1;
 		} else {
 			usage("Unsupported argument", a);
+			return -1;
 		}
 	}
+
+	return 0;
 }
 
 static int print_paths(ls_options *opts, git_index *index) 
@@ -113,7 +116,8 @@ int main(int argc, char *argv[])
 	size_t i = 0;
 	int error;
 
-	parse_options(&opts, argc, argv);
+	if ((error = parse_options(&opts, argc, argv)) < 0)
+		return error;
 
 	git_libgit2_init();
 
--- a/examples/ls-files.c
+++ b/examples/ls-files.c
@@ -28,7 +28,7 @@
 
 typedef struct {
 	int error_unmatch;
-	char * files[1024];
+	char *files[1024];
 	size_t file_count;
 } ls_options;
 
@@ -84,13 +84,13 @@ static int print_paths(ls_options *opts, git_index *index)
 	size_t i;
 	const git_index_entry *entry;
 
-	/* if there are not files explicitly listed by the user print all entries in the index */
+	/* if there are no files explicitly listed by the user print all entries in the index */
 	if (opts->file_count == 0) {
 		size_t entry_count = git_index_entrycount(index);
 
 		for (i = 0; i < entry_count; i++) {
 			entry = git_index_get_byindex(index, i);
-			printf("%s\n", entry->path);
+			puts(entry->path);
 		}
 		return 0;
 	}
@@ -100,7 +100,7 @@ static int print_paths(ls_options *opts, git_index *index)
 		const char *path = opts->files[i];
 
 		if ((entry = git_index_get_bypath(index, path, GIT_INDEX_STAGE_NORMAL)) != NULL) {
-			printf("%s\n", path);
+			puts(path);
 		} else if (opts->error_unmatch) {
 			fprintf(stderr, "error: pathspec '%s' did not match any file(s) known to git.\n", path);
 			fprintf(stderr, "Did you forget to 'git add'?\n");
--- a/examples/ls-files.c
+++ b/examples/ls-files.c
@@ -45,6 +45,7 @@ static void usage(const char *message, const char *arg)
 static int parse_options(ls_options *opts, int argc, char *argv[])
 {
 	int parsing_files = 0;
+	int i;
 	char **file;
 
 	memset(opts, 0, sizeof(ls_options));
@@ -53,7 +54,6 @@ static int parse_options(ls_options *opts, int argc, char *argv[])
 	if (argc < 2)
 		return 0;
 
-	int i;
 	for (i = 1; i < argc; ++i) {
 		char *a = argv[i];
 
--- a/examples/ls-files.c
+++ b/examples/ls-files.c
@@ -29,10 +29,9 @@
 typedef struct {
 	int error_unmatch;
 	char * files[1024];
-	int file_count;
+	size_t file_count;
 } ls_options;
 
-/* Print a usage message for the program. */
 static void usage(const char *message, const char *arg)
 {
 	if (message && arg)
@@ -60,9 +59,10 @@ static int parse_options(ls_options *opts, int argc, char *argv[])
 		if (a[0] != '-' || parsing_files) {
 			parsing_files = 1;
 
-			// watch for overflows (just in case)
+			/* watch for overflows (just in case) */
 			if (opts->file_count == 1024) {
-				break;
+				fprintf(stderr, "ls-files can only support 1024 files at this time.\n");
+				return -1;
 			}
 
 			opts->files[opts->file_count++] = a;
@@ -81,8 +81,19 @@ static int parse_options(ls_options *opts, int argc, char *argv[])
 
 static int print_paths(ls_options *opts, git_index *index)
 {
-	int i;
+	size_t i;
 	const git_index_entry *entry;
+	
+	/* if there are not files explicitly listed by the user print all entries in the index */
+	if (opts->file_count == 0) {
+		size_t entry_count = git_index_entrycount(index);
+
+		for (i = 0; i < entry_count; i++) {
+			entry = git_index_get_byindex(index, i);
+			printf("%s\n", entry->path);
+		}
+		return 0;
+	}
 
 	/* loop through the files found in the args and print them if they exist */
 	for (i = 0; i < opts->file_count; ++i) {
@@ -106,9 +117,6 @@ int main(int argc, char *argv[])
 	ls_options opts;
 	git_repository *repo = NULL;
 	git_index *index = NULL;
-	const git_index_entry *entry;
-	size_t entry_count;
-	size_t i = 0;
 	int error;
 
 	if ((error = parse_options(&opts, argc, argv)) < 0)
@@ -122,17 +130,7 @@ int main(int argc, char *argv[])
 	if ((error = git_repository_index(&index, repo)) < 0)
 		goto cleanup;
 
-	/* if there are files explicitly listed by the user, we need to treat this command differently */
-	if (opts.file_count > 0) {
-		error = print_paths(&opts, index);
-	} else {
-		entry_count = git_index_entrycount(index);
-
-		for (i = 0; i < entry_count; i++) {
-			entry = git_index_get_byindex(index, i);
-			printf("%s\n", entry->path);
-		}
-	}
+	error = print_paths(&opts, index);
 
 cleanup:
 	git_index_free(index);
--- a/examples/ls-files.c
+++ b/examples/ls-files.c
@@ -1,37 +1,126 @@
-#include <common.h>
+/*
+ * libgit2 "ls-files" example - shows how to view all files currently in the index
+ *
+ * Written by the libgit2 contributors
+ *
+ * To the extent possible under law, the author(s) have dedicated all copyright
+ * and related and neighboring rights to this software to the public domain
+ * worldwide. This software is distributed without any warranty.
+ *
+ * You should have received a copy of the CC0 Public Domain Dedication along
+ * with this software. If not, see
+ * <http://creativecommons.org/publicdomain/zero/1.0/>.
+ */
 
-typedef struct ls_files_state {
+#include "common.h"
+
+/**
+ * This example demonstrates the libgit2 index APIs to roughly
+ * simulate the output of `git ls-files`.
+ * `git ls-files` has many options and this currently does not show them.
+ * 
+ * `git ls-files` base command shows all paths in the index at that time.
+ * This includes staged and committed files, but unstaged files will not display.
+ */
+
+#define MAX_FILES 64
+
+typedef struct ls_options {
+	int error_unmatch;
+	char *files[MAX_FILES];
+	int file_count;
+} ls_options;
+
+void parse_options(ls_options *opts, int argc, char *argv[]);
+
+int main(int argc, char *argv[]) {
+	ls_options opts;
 	git_repository *repo;
 	git_index *index;
-	char **files;
-	size_t num_entries;
-} ls_files;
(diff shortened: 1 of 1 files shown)


examples/Makefile
```diff
@@ -3,7 +3,7 @@
 CC = gcc
```
pks-t: We have a makefile here? I'll create a PR and delete that.

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+ *  - default ls-files behavior
+ * 
+ * This currently does not support:
+ * 	- anything else
+ * 
+ */
+
+#define MAX_FILES 64
+
+typedef struct ls_options {
```
pks-t: I think we usually avoid giving the struct itself a name if it's already typedef'd.

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+	int error_unmatch;
+	char *files[MAX_FILES];
```
pks-t: You never check whether you overflow that array. Furthermore, I'd propose to just use an array, that should be much simpler. Just have a look at "array.h" :)

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+	int file_count;
+} ls_options;
+
+static void usage(const char *message, const char *arg);
+void parse_options(ls_options *opts, int argc, char *argv[]);
+int print_error_unmatch(ls_options *opts, git_index *index);
```
pks-t: `static` declarations are missing here. You could also consider to reorder functions to avoid these declarations, but I'm rather indifferent to that.

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+
+int main(int argc, char *argv[]) {
```
pks-t: We prefer to have a newline between function head and the opening brace.

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+	exit(1);
+}
+
+void parse_options(ls_options *opts, int argc, char *argv[]) {
+	int parsing_files = 0;
+	struct args_info args = ARGS_INFO_INIT;
+	
+	memset(opts, 0, sizeof(ls_options));
+	opts->error_unmatch = 0;
+	opts->file_count = 0;
```
pks-t: No need to zero those two variables out if you've already used `memset`.

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+	git_repository *repo;
+	git_index *index;
+	const git_index_entry *entry;
+	size_t entry_count;
+	size_t i = 0;
+	int error;
+
+	parse_options(&opts, argc, argv);
+
+	/* we need to initialize libgit2 */
```
pks-t: I think this and the following comments don't really have much value, as they simply state the obvious.

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+
+	/* we need to load the repo's index */
+	if ((error = git_repository_index(&index, repo)) != 0)
+		goto cleanup;
+
+	/* if the error_unmatch flag is set, we need to print it differently */
+	if (opts.error_unmatch) {
+		error = print_error_unmatch(&opts, index);
+		goto cleanup;
+	}
```
pks-t: I think you've got a slight misconception of `error_unmatch`. What you do here is print out different things based on whether `error_unmatch` is set or not. But in fact, in both cases we'd like to either list the complete index if we have no explicit file list, or in the other case just the files. So _what_ is printed should only be decided based on whether the user explicitly passed files into the program or not. `error_unmatch` is then only used in the case where we have paths to check if the path actually exists.

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+	git_repository_free(repo);
+
+	/* we need to shutdown libgit2 */
+	git_libgit2_shutdown();
+
+	return error;
+}
+
+/* Print a usage message for the program. */
+static void usage(const char *message, const char *arg)
```
pks-t: This here is a strange signature, I'd have expected varargs. But as you've simply taken that from other examples, I guess it's fine. Probably it would also be a bit overkill as there is no need for that right now.

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+	for (args.pos = 1; args.pos < argc; ++args.pos) {
+		char *a = argv[args.pos];
+
+		/* if it doesn't start with a '-' or is after the '--' then it is a file */
+		if (a[0] != '-' || !strcmp(a, "--")) {
+			if (parsing_files) {
+				opts->files[opts->file_count++] = a;
+			} else { 
+				parsing_files = 1;
+			}
```
pks-t: No need for braces here.

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+
+	for (args.pos = 1; args.pos < argc; ++args.pos) {
+		char *a = argv[args.pos];
+
+		/* if it doesn't start with a '-' or is after the '--' then it is a file */
+		if (a[0] != '-' || !strcmp(a, "--")) {
+			if (parsing_files) {
+				opts->files[opts->file_count++] = a;
+			} else { 
+				parsing_files = 1;
```
pks-t: To me this looks like you're actually skipping the first file? Or am I misinterpreting that?

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+		} else if (!strcmp(a, "--error-unmatch")) {
+			opts->error_unmatch = 1;
+			parsing_files = 1;
+		} else {
+			usage("Unsupported argument", a);
+		}
+	}
+}
+
+int print_error_unmatch(ls_options *opts, git_index *index) {
```
pks-t: I'd rename that to e.g. `print_paths`. You could then have a second function `print_index` when no paths are given.

examples/ls-files.c
```diff
@@ -0,0 +1,153 @@
...
+
+	/* loop through the files found in the args and print them if they exist */
+	for (i = 0; i < opts->file_count; i++) {
+		const char *path = opts->files[i];
+
+		entry = git_index_get_bypath(index, path, GIT_INDEX_STAGE_NORMAL);
+		if (!entry) {
+			printf("error: pathspec '%s' did not match any file(s) known to git.\n", path);
+			printf("Did you forget to 'git add'?\n");
+			return -1;
```
pks-t: This error should in fact only trigger `if (!entry && opts->error_unmatch)`.

cjhoward92:
@pks-t Thanks a ton for reviewing this! I will be happy to make these changes, hopefully tonight, but maybe tomorrow. I really appreciate you taking the time to look over this and being patient with me! I am pretty new to C, but I would really like to become an active contributor to libgit2, so I will work hard to get you some quality PR's.

Thanks again!

cjhoward92:
@pks-t This should be good to go. Let me know if there is anything else you would have me change! Thanks again for your time!

cjhoward92:
All updated. I figured out how to use `git_array_t` with `char *` as the element. Thanks for the pointers!

cjhoward92:
I was wondering if I should squash this?

cjhoward92:
Any chance anyone will take a look at this?

pks-t:
I'll find some time next week.

examples/ls-files.c
```diff
@@ -0,0 +1,152 @@
...
+
+	if (argc < 2)
+		return 0;
+
+	for (args.pos = 1; args.pos < argc; ++args.pos) {
+		char *a = argv[args.pos];
+
+		/* if it doesn't start with a '-' or is after the '--' then it is a file */
+		if (a[0] != '-') {
+			parsing_files = 1;
```
pks-t: You set this, but you don't use that value to distinguish files/non-files. It should probably be

```
if (a[0] != '-' || parsing_files) {
    ...
}
```

So as soon as we find the first non-argument, we treat all remaining arguments as files.

examples/ls-files.c
```diff
@@ -0,0 +1,152 @@
...
+
+			file = git_array_alloc(opts->files);
+			GITERR_CHECK_ALLOC(file);
+			*file = a;
+		} else if (!strcmp(a, "--")) {
+			parsing_files = 1;
+		} else if (!strcmp(a, "--error-unmatch") && !parsing_files) {
```
pks-t: If you change the first condition as proposed, there's no need for the `!parsing_files` anymore.

examples/ls-files.c
```diff
@@ -0,0 +1,152 @@
...
+	const git_index_entry *entry;
+
+	/* loop through the files found in the args and print them if they exist */
+	for (i = 0; i < git_array_size(opts->files); ++i) {
+		const char *path = *(char **)git_array_get(opts->files, i);
+
+		entry = git_index_get_bypath(index, path, GIT_INDEX_STAGE_NORMAL);
+		if (!entry && opts->error_unmatch) {
+			printf("error: pathspec '%s' did not match any file(s) known to git.\n", path);
+			printf("Did you forget to 'git add'?\n");
```
pks-t: These should probably be `fprintf(stderr, ...)` to print to stderr instead.

examples/ls-files.c
```diff
@@ -0,0 +1,152 @@
...
+		goto cleanup;
+
+	if ((error = git_repository_index(&index, repo)) < 0)
+		goto cleanup;
+
+	/* if there are files explicitly listed by the user, we need to treat this command differently */
+	if (git_array_size(opts.files) > 0) {
+		error = print_paths(&opts, index);
+		goto cleanup;
+	}
```
pks-t: I would simply put the rest into an `else` branch. Like that, you don't need that `goto cleanup` anymore and it becomes a bit more obvious.

examples/ls-files.c
```diff
@@ -0,0 +1,152 @@
...
+	entry_count = git_index_entrycount(index);
+
+	/* loop through the entries by index and display their pathes */
+	for (i = 0; i < entry_count; i++) {
+		entry = git_index_get_byindex(index, i);
+		printf("%s\n", entry->path);
+	}
+
+cleanup:
+	/* free our allocated resources */
```
pks-t: I think there's no need to state the obvious here ;) Same for the two previous comments.

cjhoward92:
@pks-t This is all updated, but I am not too sure what is causing the windows builds to break.

ethomson:
Regarding the Windows builds: [`git_array_alloc`](https://github.com/libgit2/libgit2/pull/4380/files#diff-9839fc44c21b30e5de8f51a65569c036R64) calls `git__reallocarray`.  We're building with the CRT debugging functions on Windows, which ends up calling `git__crtdbg__reallocarray`.

However, I think that we're not linking the examples with the CRT debugging libraries.  Ugh.  Not sure how to fix this offhand.

cjhoward92:
Ok, I will check it out when I get a moment. Thanks for the explanation!

cjhoward92:
Would updating the CMakeLists for the examples to statically link the CRT debugging libraries work? I'm not super familiar with how CMake works to really know. All I found was the flags to link when running the main build process. I don't really understand why this fails on AppVeyor but works on my machine either, unless I installed something that exposes the CRT libraries on my includes path.

ethomson:
AppVeyor is explicitly turning on the CRT memory debugging functionality, with `-DDEBUG_POOL`.  I suspect that this should fail if you enable the same functionality?

In any case, the ideal way to fix this would be to not use `git_array_alloc`.  It's not public API, it's private API, so we should not use it in examples.  Anybody reading (or more likely, copy/pasting) the example code will not be able to build it since we do not expose that macro publicly.

ethomson:
I see why you want to use `git_array_alloc` instead of trying to manage a dynamic array yourself...  If it were me, I'd just create a static array of `char *` and hold a separate counter to the number of arguments that you increment.  You can easily make this `char *files[1024]` and that's practically _way_ more than enough.  It could be 16, tbqh.  This is a sample, not production code, and it should be readable and easy to use not perfect.

I do think that we should make sure that we do not crash if a user enters 17 filenames though.  :grin

cjhoward92:
Looks like there is only an error with some of the credentials stuff now, might be some network error. I think they should be good to go, though.

cjhoward92:
bump

examples/ls-files.c
```diff
@@ -0,0 +1,143 @@
...
+	for (i = 1; i < argc; ++i) {
+		char *a = argv[i];
+
+		/* if it doesn't start with a '-' or is after the '--' then it is a file */
+		if (a[0] != '-' || parsing_files) {
+			parsing_files = 1;
+
+			// watch for overflows (just in case)
+			if (opts->file_count == 1024) {
+				break;
```
pks-t: You probably shouldn't break here but print an error and return an error code.

examples/ls-files.c
```diff
@@ -0,0 +1,143 @@
...
+		return 0;
+
+	for (i = 1; i < argc; ++i) {
+		char *a = argv[i];
+
+		/* if it doesn't start with a '-' or is after the '--' then it is a file */
+		if (a[0] != '-' || parsing_files) {
+			parsing_files = 1;
+
+			// watch for overflows (just in case)
```
pks-t: C++-style comments are not allowed

examples/ls-files.c
```diff
@@ -0,0 +1,143 @@
...
+ * This currently supports the default behavior and the `--error-unmatch` option.
+ */
+
+typedef struct {
+	int error_unmatch;
+	char * files[1024];
+	int file_count;
+} ls_options;
+
+/* Print a usage message for the program. */
```
pks-t: This comment does not provide any useful information

examples/ls-files.c
```diff
@@ -0,0 +1,143 @@
...
+		}
+	}
+
+	return 0;
+}
+
+static int print_paths(ls_options *opts, git_index *index)
+{
+	int i;
+	const git_index_entry *entry;
```
pks-t: This could also be scoped to the loop

examples/ls-files.c
```diff
@@ -0,0 +1,143 @@
...
+	if (opts.file_count > 0) {
+		error = print_paths(&opts, index);
+	} else {
+		entry_count = git_index_entrycount(index);
+
+		for (i = 0; i < entry_count; i++) {
+			entry = git_index_get_byindex(index, i);
+			printf("%s\n", entry->path);
+		}
+	}
```
pks-t: Just a small style issue: I do find it a bit funny that some printing logic is in `print_paths` and some is not. Maybe put the complete priting logic into `print_paths` and destinguish cases by `opts.file_count` inside of that function, then? That would make the `main` function a bit nicer to read.

cjhoward92:
@pks-t Hopefully this is updated in the way you recommended. Let me know what you think! Thanks for all of the feedback!

examples/ls-files.c
```diff
@@ -0,0 +1,141 @@
...
+		char *a = argv[i];
+
+		/* if it doesn't start with a '-' or is after the '--' then it is a file */
+		if (a[0] != '-' || parsing_files) {
+			parsing_files = 1;
+
+			/* watch for overflows (just in case) */
+			if (opts->file_count == 1024) {
+				fprintf(stderr, "ls-files can only support 1024 files at this time.\n");
+				return -1;
```
pks-t: Nice :)

examples/ls-files.c
```diff
@@ -0,0 +1,141 @@
...
+		const char *path = opts->files[i];
+
+		entry = git_index_get_bypath(index, path, GIT_INDEX_STAGE_NORMAL);
+		if (!entry && opts->error_unmatch) {
+			fprintf(stderr, "error: pathspec '%s' did not match any file(s) known to git.\n", path);
+			fprintf(stderr, "Did you forget to 'git add'?\n");
+			return -1;
+		}
+
+		printf("%s\n", path);
```
pks-t: I just tried the code locally as I wanted to merge it, but I found this bug here. You're inconditionally printing `path` whether it is found or not. But in fact, you only want to print it if `entry != NULL`. So something like the following:

```
if ((entry = git_index_get_bypath(...)) != NULL) {
    printf("%s\n", path);
} else if (opts->error_unmatch) {
    fprintf(stderr, ...);
    return -1;
}
```

examples/ls-files.c
```diff
@@ -0,0 +1,141 @@
...
+
+	git_libgit2_init();
+
+	if ((error = git_repository_open_ext(&repo, ".", 0, NULL)) < 0)
+		goto cleanup;
+
+	if ((error = git_repository_index(&index, repo)) < 0)
+		goto cleanup;
+
+	error = print_paths(&opts, index);
```
pks-t: Yup, this looks like a very clean `main` function now :)

cjhoward92:
Wow, sorry this took _so_ long to get done! I have been really busy lately. Thanks a ton for all of the review you have done! I hope I finally have it clean enough to get merged!

cjhoward92:
Any more review on this? It looks like the test failures and build issues were an internet connectivity issue perhaps. I wonder if restarting the build on Travis would fix that. Thanks!

examples/ls-files.c
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * `git ls-files` base command shows all paths in the index at that time.
+ * This includes staged and committed files, but unstaged files will not display.
+ *
+ * This currently supports the default behavior and the `--error-unmatch` option.
+ */
+
+typedef struct {
+	int error_unmatch;
+	char * files[1024];
```
pks-t: `*` goes with the variable name, so this should rather be `char *files[1024];`.

examples/ls-files.c
```diff
@@ -0,0 +1,140 @@
...
+
+	return 0;
+}
+
+static int print_paths(ls_options *opts, git_index *index)
+{
+	size_t i;
+	const git_index_entry *entry;
+
+	/* if there are not files explicitly listed by the user print all entries in the index */
```
pks-t: "not" should be a "no"

examples/ls-files.c
```diff
@@ -0,0 +1,140 @@
...
+	if (opts->file_count == 0) {
+		size_t entry_count = git_index_entrycount(index);
+
+		for (i = 0; i < entry_count; i++) {
+			entry = git_index_get_byindex(index, i);
+			printf("%s\n", entry->path);
```
pks-t: `puts(entry->path)` would be preferred

examples/ls-files.c
```diff
@@ -0,0 +1,140 @@
...
+		}
+		return 0;
+	}
+
+	/* loop through the files found in the args and print them if they exist */
+	for (i = 0; i < opts->file_count; ++i) {
+		const char *path = opts->files[i];
+
+		if ((entry = git_index_get_bypath(index, path, GIT_INDEX_STAGE_NORMAL)) != NULL) {
+			printf("%s\n", path);
```
pks-t: `puts` instead of `printf`

cjhoward92:
I was in here so I fixed those final issues you had anyway! Thanks again!

PR #4386, opened 2017-10-23
ignore spaces in .gitignore files

novalis:
... as git itself does.

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -213,6 +213,16 @@ static int parse_ignore_file(
 			if (ignore_case)
 				match->flags |= GIT_ATTR_FNMATCH_ICASE;
 
+			while (match->length > 0) {
+				if (match->pattern[match->length - 1] == ' ' ||
+				    match->pattern[match->length - 1] == '\t') {
+					match->pattern[match->length - 1] = 0;
+					match->length --;
+				} else {
+					break;
+				}
+			}
+
 			scan = git__next_line(scan);
 
 			/*
--- a/tests/attr/ignore.c
+++ b/tests/attr/ignore.c
@@ -51,6 +51,16 @@ void test_attr_ignore__allow_root(void)
 	assert_is_ignored(false, "NewFolder/NewFolder/File.txt");
 }
 
+void test_attr_ignore__ignore_space(void)
+{
+	cl_git_rewritefile("attr/.gitignore", "/\n\n/NewFolder \n/NewFolder/NewFolder");
+
+	assert_is_ignored(false, "File.txt");
+	assert_is_ignored(true, "NewFolder");
+	assert_is_ignored(true, "NewFolder/NewFolder");
+	assert_is_ignored(true, "NewFolder/NewFolder/File.txt");
+}
+
 void test_attr_ignore__ignore_root(void)
 {
 	cl_git_rewritefile("attr/.gitignore", "/\n\n/NewFolder\n/NewFolder/NewFolder");


ethomson:
Agree that we should ignore trailing spaces, but can you look at the two test failures that this introduces?  

In particular, we want to make sure that we don't regress on carriage return handling.  It's not uncommon to have a carriage return at the end of a filename (Mac OS uses the metadata file `Icon\r`, which is `.gitignore`able with special rules.

novalis:
Ah, the test suite was segfaulting for me before it got to the status tests, so I didn't see those failures.  Let's try it this way...

ethomson:
The test suite was segfaulting?  :/  hopefully in the `threads` tests?

src/ignore.c
```diff
@@ -203,7 +203,7 @@ static int parse_ignore_file(
 			break;
 		}
 
-		match->flags = GIT_ATTR_FNMATCH_ALLOWSPACE | GIT_ATTR_FNMATCH_ALLOWNEG;
+		match->flags = GIT_ATTR_FNMATCH_ALLOWNEG | GIT_ATTR_FNMATCH_ALLOWSPACE;
```
carlosmn: Is there a reason these need to change order?

carlosmn:
The tests from #2688 though they're in status, should be enough to make sure we don't regress the odd macOS naming.

src/ignore.c
novalis: Nope, sorry.

novalis:
Nope, in refs::iterator.  But maybe there's just something funny in my working tree.

ethomson:
There shouldn't be anything so funny that we crash.  😥

Would love any details you can provide.

novalis:
I only get it when running the whole suite, as far as I can tell.  Here's a backtrace:

Thread 1 "libgit2_clar" received signal SIGSEGV, Segmentation fault.
0x000055555578888e in git_oidmap_size (map=0x0)
    at /home/novalis/libgit2/src/oidmap.c:36
36		return kh_size(map);
(gdb) bt
#0  0x000055555578888e in git_oidmap_size (map=0x0)
    at /home/novalis/libgit2/src/oidmap.c:36
#1  0x000055555572b084 in git_cache_size (cache=0x7fffec027d40)
    at /home/novalis/libgit2/src/cache.h:58
#2  0x000055555572b2f0 in clear_cache (cache=0x7fffec027d40)
    at /home/novalis/libgit2/src/cache.c:82
#3  0x000055555572b3d6 in git_cache_clear (cache=0x7fffec027d40)
    at /home/novalis/libgit2/src/cache.c:99
#4  0x00005555557b2e4a in git_repository__cleanup (repo=0x7fffec027d20)
    at /home/novalis/libgit2/src/repository.c:145
#5  0x00005555557b2ec0 in git_repository_free (repo=0x7fffec027d20)
    at /home/novalis/libgit2/src/repository.c:161
#6  0x00005555556c150d in test_refs_iterator__cleanup ()
    at /home/novalis/libgit2/tests/refs/iterator.c:14
#7  0x000055555559bdae in clar_run_test (
    test=0x555555bbe280 <_clar_cb_refs_iterator+96>, 
    initialize=0x555555bc5a88 <_clar_suites+16136>, 
    cleanup=0x555555bc5a98 <_clar_suites+16152>)
    at /home/novalis/libgit2/tests/clar.c:232
#8  0x000055555559bfdf in clar_run_suite (
    suite=0x555555bc5a80 <_clar_suites+16128>, filter=0x0)
    at /home/novalis/libgit2/tests/clar.c:286
---Type <return> to continue, or q <return> to quit---
#9  0x000055555559c4e8 in clar_test_run ()
    at /home/novalis/libgit2/tests/clar.c:426
#10 0x0000555555599e8e in main (argc=1, argv=0x7fffffffdf98)
    at /home/novalis/libgit2/tests/main.c:19

carlosmn:
The existing plus new tests are passing, so I'm going to merge this. Thanks.

The stack trace for your error suggests a double-free, though it's pretty odd since CI isn't having issues. Running the test suite under `valgrind` is often a pretty good way to find such issues.

PR #4389, opened 2017-10-26
implement NTLM authentication

rcarz:
This patch implements NTLMv2 session negotiation over HTTP. Git repositories served by Team Foundation Server are currently only accessible if IIS is configured for SPNEGO authentication, which isn't always the case. I excluded support for older LM/NTLM versions since they are strongly discouraged and unlikely to be used in the wild.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -20,3 +20,4 @@ Edward Thomson <ethomson@github.com> <ethomson@microsoft.com>
 Edward Thomson <ethomson@github.com> <ethomson@edwardthomson.com>
 J. David Ibáñez <jdavid.ibp@gmail.com> <jdavid@itaapy.com>
 Russell Belfer <rb@github.com> <arrbee@arrbee.com>
+Bob Carroll <bob.carroll@alum.rit.edu> <bobc@qti.qualcomm.com>
--- a/AUTHORS
+++ b/AUTHORS
@@ -75,3 +75,4 @@ Tim Harder
 Torsten Bögershausen
 Trent Mick
 Vicent Marti
+Bob Carroll
--- a/cmake/Modules/FindUnistring.cmake
+++ b/cmake/Modules/FindUnistring.cmake
@@ -1,0 +1,24 @@
+# - Find UNISTRING
+# Find the unistring includes and library
+#
+#  UNISTRING_INCLUDE_DIR - where to find unistr.h, etc.
+#  UNISTRING_LIBRARY     - unistring library.
+#  UNISTRING_FOUND       - True if unistring found.
+
+IF (UNISTRING_INCLUDE_DIR)
+    # Already in cache, be silent
+    SET(UNISTRING_FIND_QUIETLY TRUE)
+ENDIF (UNISTRING_INCLUDE_DIR)
+
+FIND_PATH(UNISTRING_INCLUDE_DIR unistr.h)
+
+SET(UNISTRING_NAMES unistring libunistring)
+FIND_LIBRARY(UNISTRING_LIBRARY NAMES ${UNISTRING_NAMES})
+
+# handle the QUIETLY and REQUIRED arguments and set UNISTRING_FOUND to
+# TRUE if all listed variables are TRUE
+INCLUDE(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(UNISTRING DEFAULT_MSG
+    UNISTRING_INCLUDE_DIR UNISTRING_LIBRARY)
+
+MARK_AS_ADVANCED(UNISTRING_LIBRARY UNISTRING_INCLUDE_DIR UNISTRING_LIBRARY)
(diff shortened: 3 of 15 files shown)


mikeando:
It looks like this could do with some logic for ensuring that the `heimntlm.h` header is actually available and only conditionally enabling these changes when it is. Probably want to be able to configure the heimdal search paths too - in case a user has installed it in a non-standard location.

ethomson:
I would be very happy to see NTLM support on Unix hosts.  But yeah, at a bare minimum this needs to determine whether the GSSAPI implementation is Heimdal or not.  However, limiting this to only Heimdal users is a bit disappointing since it's surely in the minority of installations.

Let me have a think on this.  I have an NTLM implementation in Java that I would like to adapt to C...  (Unicode is the rub here, unsurprisingly.)

rcarz:
@ethomson I had originally intended to use GSSAPI rather than the underlying provider directly. I went with Heimdal because MIT GSSAPI doesn't have an NTLM provider baked in. But it was taking too much time and seemed needlessly complex.

Alternatively, it might be fairly painless for me to swap out this implementation for Libntlm so long as there aren't any objections to a new dependency.

ethomson:
Yeah, I think that I would prefer to not assume Heimdal...  Libntlm is definitely an option.  It looks trivial enough to pull it into our tree ourselves.  (I guess that's actually Libntlm, with a capital L.  Ugh.)

Have you used Libntlm before?  Does it work?  Scanning the code, it looks like it doesn't do Unicode at all, which is a disappointment.  (But it's an improvement over where we're at now.)

rcarz:
I used it a long time ago and it worked then, but I might have a better option if you're keen to pull it into the tree. Let me see if I can make it work.

rcarz:
@ethomson I finally found time to work on this. After comparing the published specification from Microsoft, it seems like Libntlm is missing a lot of functionality. I found a more complete implementation with unicode support that was written to be part of a GSS plugin. I pulled the NTLM bits in wholesale however I'm not sure if that's desirable for libgit2. It does emit compilation warnings that might need some attention.

I'm also open to splitting this out into a separate library, but I wanted to get feedback on the general idea before spending any more time on it. I tested the implementation with a TFS-hosted repo and it worked fine for my purposes.

pks-t:
The code you are pulling in from simo5/gss-ntlmssp is licensed under LGPLv3. This is unfortunately incompatible with the GPLv2 we are using, as far as I know even with the linking exception we provide. Thus it is not an option to use it.

ethomson:
I’ll do some investigation around the licensing.  It’s possible that this is dual licensed or there was a prior version that was under a different license.

ethomson:
If that doesn’t pan out, I’ll dust off some NTLMv2 code that I started working on a while back, and finish it up.

rcarz:
I sent an email to author asking if there’s a possibility to dual license. If not, then I’m willing to help complete your implementation @ethomson if you’re interested. I have to maintain a fork until this feature is upstreamed.

rcarz:
I had a nice discussion with the author about potential licensing changes. He was generally amenable to offering a more compatible license, however, there were concerns about the effects of libgit2's linking exception. In particular, the linking exception appears to allow for distributed, non-copyleft modifications so long as libgit2 is statically linked to other components.

> In addition to the permissions in the GNU General Public License,
> the authors give you unlimited permission to link the compiled
> version of this library into combinations with other programs,
> and to **distribute those combinations without any restriction
> coming from the use of this file**. (The General Public License
> restrictions do apply in other respects; for example, they cover
> modification of the file, and distribution when not linked into
> a combined executable.)

I think I understand the intentions of this exception but it does feel somewhat vague and open to multiple interpretations. In any case, the author is concerned about preserving copyleft protections. He did propose potentially licensing under LGPLv2.1. If you think libgit2's license is compatible then I can continue the discussion.

Otherwise it would need to be installed separately as a GSSAPI mechanism with mitkrb5. I was hoping to include built-in support without the overhead of GSSAPI but I'm not sure writing yet another NTLM library is right choice. What do you think?

pks-t:
Thanks for reaching out. The compatibiliy list at gnu.org [1] explicitly lists GPLv2 as being compatible with LGPLv2.1, so this would be okay. We also have the winhttp code embedded, which is licensed under LGPLv2.1+.

The additional linking exception provided by us does not impact bundled code under a different license, so it would not apply to the GSSAPI library. Thus, I'd say LGPLv2.1 is a legit choice here, both for us as well as for the original author.

[1]: https://www.gnu.org/licenses/license-list.html#GPLCompatibleLicenses

ethomson:
I wanted to avoid GSSAPI or external dependencies (as best as possible) so that this could always be on in our code base and not an optional feature.  This is a port of some Microsoft NTLM code that's available under the MIT license: https://github.com/ethomson/ntlmclient

This will use the Common Crypto libraries on macOS, and either OpenSSL or mbedTLS on Linux.  This is the direction that I would like to take to implement NTLM on non-Windows.  I don't have time to work on integrating that library into libgit2 - if somebody else wants to, that would be great, otherwise I'll do that next week.

rcarz:
@ethomson I can spend some time on it this week

ethomson:
Neat!  I pushed up a branch [ethomson/ntlm](https://github.com/libgit2/libgit2/tree/ethomson/ntlm) that adds the ntlm library as a dependency (and coordinates the choice of crypto provider with libgit2) but I did not wire up the authentication.  I'm out of the office for a few days, so if you want to do that, it should be reasonably straightforward, otherwise we can discuss when I return.

ethomson:
Superceded by #5052

PR #4391, opened 2017-10-29
diff: expose the "indent heuristic" in the diff options

carlosmn:
We default to off, but we might want to consider changing `GIT_DIFF_NORMAL` to
include it.

This is a followup to #3944  that lets consumers of the library activate it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ v0.26 + 1
 
 * We now support conditional includes in configuration files.
 
+* There is a new diff option, `GIT_DIFF_INDENT_HEURISTIC` which activates a
+  heuristic which takes into account whitespace and indentation in order to
+  produce better diffs when dealing with ambiguous diff hunks.
+
 ### API additions
 
 * `git_remote_create_detached()` creates a remote that is not associated
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -206,6 +206,12 @@ typedef enum {
 	 *  can apply given diff information to binary files.
 	 */
 	GIT_DIFF_SHOW_BINARY = (1 << 30),
+
+	/** Use a heuristic that takes indentation and whitespace into account
+	 * which generally can produce better diffs when dealing with ambiguous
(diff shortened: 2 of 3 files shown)


ethomson:
What's the Git default now?

carlosmn:
The release notes for 2.14.0 say this is now the default for git (but you can still turn it off).

Of course, docs being docs, they still say that it's experimental and off by default, but that's a bug in their docs.

ethomson:
Cool.  Might be nice to put this through its paces before making it the default for us, too...  are you planning on turning this on for anything?

carlosmn:
I don't have anything to "put it through its paces". The site has been using git to produce diffs for a while now.

ethomson:
Right, in that case, why not let's just make it the new default?

carlosmn:
Would you rather we set `GIT_DIFF_NORMAL = GIT_DIFF_INDENT_HEURISTIC` or to flip the option so it's `GIT_DIFF_DISABLE_INDENT_HEURISTIC`?

carlosmn:
@ethomson this isn't _great_ but it enables the heuristic while keeping the blame output the same, at least in the tested situations.

I don't see a difference in hunks when running `git blame` but we've diverged from that implementation anyway.

ethomson:
Interesting.  I was going to suggest that we use `ENABLE_...` and then update `GIT_DIFF_OPTIONS_INIT` to set this bit in the flags, instead of defaulting it to on.  This would let the (few) users who redefine the constants themselves (like LibGit2Sharp...perhaps _only_ LibGit2Sharp?) opt-in to this instead of having a surprising behavior change.

🤔 I don't really know how compelling that is, and I'm not sure that I actually have any particularly strong feelings one way or another.

ethomson:
Anyway, I'm fine with the current strategy, so any way you want to do the defaults is fine - I think you should feel free to merge this after fixing the changelog merge conflict.  Thanks!

carlosmn:
We should figure out eventually what's up with blame, but we can probably punt on that for quite a while.

Issue #4390, opened 2017-10-27
Configuration functions not case-sensitive


omus:
I've noticed that functions like `git_config_set_string` write variable names in lowercase and `git_config_iterator_glob_new` don't work when you provide non-lowercase input. I would expect that functions which write config variables to retain the provided casing.

An example of a git configuration variable that uses uppercase letters is [useHttpPath](https://git-scm.com/docs/gitcredentials#gitcredentials-useHttpPath)

Git2 version v0.26.0 (15e1193)

carlosmn:
There's a bunch of different things here, but they look to stem out of a misunderstanding of how Git configuration keys behave.

We should be case-preserving. We likely simply write out of the normalised representation instead of the input string, which would indeed be a bug, though a cosmetic one.

Configuration options like `credential.useHttpPath` are case *in*sensitive in Git configuration.

I don't recall how much of the regexp behaviour was done to explicitly copy git's actual behaviour, but the behaviour of `git config --get-regexp` is also case-sensitive. In particular, its documentation specifies that

>Regular expression matching is currently case-sensitive and done against a canonicalized version of the key in which section and variable names are lowercased, but subsection names are not.

which is about the only workable solution, as Git configuration keys are in part case-sensitive and in part case-insensitive. We also match against the normalised version of the key. The correct thing here if you want a glob to match the url-specific option, is to spell both the `credential` and `usehttppath` in lowercase.

pks-t:
With the writing part being fixed via #4392, I'd consider this issue closed as it is based on a misunderstanding. @omus: feel free to expand on your issue if you disagree, otherwise please close this issue.

omus:
> I've noticed that functions like `git_config_set_string` write variable names in lowercase

#4392 does address this part of the issue.

> `git_config_iterator_glob_new` doesn't work when you provide non-lowercase input

This part of the issue should be addressed by #4396. I'll close this issue once that PR is merged.

pks-t:
Seems fair. I've overlooked issue #4396

PR #4392, opened 2017-10-30
Preserve the input casing when writing config files

carlosmn:
Some keys are presented in camelCase in the configuration, so users might want to write them as such in their tool. It can be confusing if we only ever write out the normalised versions of the keys as it differs from what they'd be writing if they had manually opened the file.

This now also matches the behaviour of `git config` when given the same inputs.

```shell
% git config -f foo sOMe.ThInG foo
% git config -f foo SomE.OtheR thing
% cat foo
[sOMe]
        ThInG = foo
        OtheR = thing
```

This covers the confusing aspect in #4390 which don't stem from the Git config spec.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -1955,10 +1955,12 @@ static int config_write(diskfile_backend *cfg, const char *orig_key, const char 
 	ldot = strrchr(key, '.');
 	name = ldot + 1;
 	section = git__strndup(key, ldot - key);
+	GITERR_CHECK_ALLOC(section);
 
 	ldot = strrchr(orig_key, '.');
 	orig_name = ldot + 1;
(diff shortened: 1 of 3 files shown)


src/config_file.c
```diff
@@ -1954,17 +1956,24 @@ static int config_write(diskfile_backend *cfg, const char *key, const regex_t *p
 	name = ldot + 1;
 	section = git__strndup(key, ldot - key);
 
+	ldot = strrchr(orig_key, '.');
+	orig_name = ldot + 1;
+	orig_section = git__strndup(orig_key, ldot - orig_key);
```
ethomson: We should check for success here, and the `strndup` a few lines above.

ethomson:
Other than my OOM nitpicking, this is :+1: from me.

PR #4394, opened 2017-10-31
travis: put clar's sandbox in a ramdisk on macOS

carlosmn:
The macOS tests are by far the slowest right now. This attempts to remedy the
situation somewhat by asking clar to put its test data on a ramdisk.

### DIFF
--- a/script/cibuild.sh
+++ b/script/cibuild.sh
@@ -10,6 +10,15 @@ fi
 
 if [ "$TRAVIS_OS_NAME" = "osx" ]; then
 	export PKG_CONFIG_PATH=$(ls -d /usr/local/Cellar/{curl,zlib}/*/lib/pkgconfig | paste -s -d':' -)
+
+	# Set up a ramdisk for us to put our test data on to speed up tests on macOS
+	export CLAR_TMP="$HOME"/_clar_tmp
+	mkdir -p $CLAR_TMP
+
+	# 2M sectors aka ~1GB of space
+	device=$(hdiutil attach -nomount ram://$((2 * 1024 * 1024)))
+	newfs_hfs $device
+	mount -t hfs $device $CLAR_TMP
 fi
 
 # Should we ask Travis to cache this file?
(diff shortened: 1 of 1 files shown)


carlosmn:
The tests that attempt to deal with a 2GB and a 4GB file error out, since right now that ramdisk is only 1GB, but everything else passes, and the time has gone down from 15-17m to 9-10m.

pks-t:
That's lovely. Doesn't macOS have a standard tmpfs/ramdisk at /tmp like many other Unixes?

carlosmn:
> a standard tmpfs/ramdisk at /tmp like many other Unixes?

Unless they've introduced it recently, I'm not aware of a tmpfs `/tmp/` being the default.

But either way, no. macOS doesn't even have a tmpfs driver proper. You _have_ to do the dance of creating a ramdisk and putting HFS+ there of all things.

Though I suppose this should be the preferred way on macOS regardless, since we definitely want to test how we behave against HFS+.

pks-t:
We could even test on both HFS+ with case sensitivity and without case sensitivity. Though I don't know whether there's any benefit in that, as we already test case insensitively on Windows.

carlosmn:
A case-sensitive HFS+ should just behave like any other unix filesystem, so I don't think there's much point in adding that to the test runs. The difference would be how it handles composed characters but that shouldn't change with the case-sensitivity.

ethomson:
It's possible that we're doing some "is this a Mac?" tests and assume case sensitivity based on that (instead of looking at `core.ignorecase` or deriving the case sensitivity.)  But for sure the vast majority of default HFS+ installations are going to be case insensitive.  (macOS is famously broken if the system or user folders are on a case sensitive filesystem.)

So we could add some case sensitive Mac test runs, but if we're going to do that I'd rather do APFS, since (IIRC) iOS is doing case-_sensitive_ APFS today, so it's an install base that we actually support in practice and not just in theory.

Anyway.  Going to merge this as-is with its case-insensitive goodness since it halves(!) the test run time on macOS, and we can add add case-sensitive test runs later.

ethomson:
✨ !!

PR #4396, opened 2017-11-04
config: specify how we match the regular expressions

carlosmn:
We do it the same as git does: case-sensitively on the normalized form of the
variable name.

While here also specify that we're case-sensitive on the values when handling
the values when setting or deleting multivars.

---

#4390 highlights this deficiency in the documentation.

### DIFF
--- a/include/git2/config.h
+++ b/include/git2/config.h
@@ -398,23 +398,31 @@ GIT_EXTERN(int) git_config_get_string_buf(git_buf *out, const git_config *cfg, c
  *
  * The callback will be called on each variable found
  *
+ * The regular expression is applied case-sensitively on the normalized form of
+ * the variable name: the case-insensitive parts are lower-case.
+ *
  * @param cfg where to look for the variable
  * @param name the variable's name
  * @param regexp regular expression to filter which variables we're
  * interested in. Use NULL to indicate all
  * @param callback the function to be called on each value of the variable
  * @param payload opaque pointer to pass to the callback
+ *
  */
 GIT_EXTERN(int) git_config_get_multivar_foreach(const git_config *cfg, const char *name, const char *regexp, git_config_foreach_cb callback, void *payload);
 
 /**
  * Get each value of a multivar
  *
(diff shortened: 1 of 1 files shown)


include/git2/config.h
```diff
@@ -398,23 +398,31 @@ GIT_EXTERN(int) git_config_get_string_buf(git_buf *out, const git_config *cfg, c
...
+ * The regular expression is applied case-sensitively on the normalized form of
+ * the variable name: the case-insensitive parts are lower-case.
+ *
  * @param cfg where to look for the variable
  * @param name the variable's name
  * @param regexp regular expression to filter which variables we're
  * interested in. Use NULL to indicate all
  * @param callback the function to be called on each value of the variable
  * @param payload opaque pointer to pass to the callback
+ *
```
pks-t: Is there a reason why you've added those empty comment lines?

carlosmn: I originally put the extra comment after the params, but then realised the other explanations were before, and my cut-and-paste was evidently insufficient. I'll go clean these up.

omus:
If this does work the same way as git does then it maybe worth mentioning that the subsection names are [not normalized](https://git-scm.com/docs/git-config#git-config---get-regexp):

> Regular expression matching is currently case-sensitive and done against a canonicalized version of the key in which section and variable names are lowercased, but subsection names are not.

I found that `git_config_iterator_glob_new` works this way.

carlosmn:
There is no normalization possible for subsections.

include/git2/config.h
carlosmn: The extra line is now removed.

ethomson:
Yeah, I think that including something like what @omus has quoted out of `git-config`'s man page may be helpful.  It was not clear to me (and in fact i was surprised to discover) that subsection names are not lowercased.  I think it may be helpful to include something that explains that...?

ethomson:
This is a quick syntax fix @carlosmn, can we sneak this in for 0.27?

carlosmn:
I've added a commit which explicitly calls out the subsection as case-sensitive.

Issue #4385, opened 2017-10-23
Examples fail to build on windows


cjhoward92:
### Reproduction steps
Build `libgit2` with examples on:

```
mkdir build
cmake .. -DBUILD_EXAMPLES=ON
cmake --build .
```

### Expected behavior
For the examples to build

### Actual behavior
I get an error
```
"C:\Dev\libgit2\build\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\Dev\libgit2\build\examples\cgit2.vcxproj" (default target) (11) ->
(Link target) ->
  common.obj : error LNK2019: unresolved external symbol _getline referenced in function _cred_acquire_cb [C:\Dev\libgit 2\build\examples\cgit2.vcxproj]
  C:\Dev\libgit2\build\examples\Debug\cgit2.exe : fatal error LNK1120: 1 unresolved externals [C:\Dev\libgit2\build\exam ples\cgit2.vcxproj]

    22 Warning(s)
    2 Error(s)
```
### Version of libgit2 (release number or SHA1)
[Current master](https://github.com/libgit2/libgit2/commit/8f05d2d81643177ce44d1449a551ef206b18cd5e) at `8f05d2d81643177ce44d1449a551ef206b18cd5e`

### Operating system(s) tested
Windows 10 build 15063 (I tried on another windows 10 machine, too, but I do not know the build. It also failed)

carlosmn:
What compiler/VS version are you using? It works in CI so there's something different in your setup.

cjhoward92:
I will need to verify, but on one machine I have the 2013 C++ redistributables, and on the other I think it is 2015 or 2017. I am not sure if there is a precedence on the newer versions or not. Is there a way I can look at the CI configuration? Will the `appveryor.yaml` have everything I need?

carlosmn:
The appveyor file shows what rules we follow for building, including what we tell CMake to use. In Appveyor itself you can see the details of what was using in the VMs, e.g. at https://ci.appveyor.com/project/libgit2/libgit2/build/3983

cjhoward92:
I will check it out when I get home and see if I can't get it working. Thanks!

pks-t:
I think AppVeyor doesn't build examples at all. I'll create a PR which fixes that, including the getline stuff if it comes up

pks-t:
This should be fixed by #4397 now. Please verify the issue is resolved and close this ticket if so. Thank you :)

cjhoward92:
[...]

PR #4397, opened 2017-11-06
appveyor: build examples

pks-t:
By default, CMake will not build our examples directory. As we do not
instruct either the MinGW or MSVC builds on AppVeyor to enable building
these examples, we cannot verify that those examples at least build on
Windows systems.

Fix that by passing `-DBUILD_EXAMPLES=ON` to AppVeyor's CMake
invocation.


pks-t:
Yep, we actually fail to build examples on Windows due to `getline`, as described in #4385

ethomson:
Nice, thanks!

PR #4398, opened 2017-11-09
cmake: allow explicitly choosing SHA1 backend

pks-t:
Right now, the SHA1 backend is mostly chosen based on which system
libgit2 is being compiled on and which libraries have been found. The
only exception here is SHA1DC, which can be explicitly requested.

To give developers and distributions more choice, enable them to request
specific backends by passing in a `-DSHA1_BACKEND=<BACKEND>` option. By
default, we will automatically select in the following order:

1. OpenSSL, if the library was found
2. SHA1_WIN32, if the system is Win32
3. CommonCrypto, if the system is macOS
4. generic otherwise

This mostly matches our previous behaviour. One exception though is that
by default, we will now only automatically select OpenSSL as SHA1
library if it has already been found previously. This can only be the
case when the library is also using OpenSSL as TLS backend. Previously,
we always tried to use OpenSSL regardless of whether it's already being
used for other subsystems or not.

Also new is that the packager can now also specify one of our backends
explicitly by setting e.g. `-DSHA1_BACKEND=SHA1DC` to get the collision
detecting backend.

All code is inspired by our HTTPS backend selection.


pks-t:
Cc @tiennou, as this is similar to his generic HTTP work

src/CMakeLists.txt
```diff
@@ -204,34 +202,54 @@ ELSE()
 ENDIF()
 
 # Specify sha1 implementation
-IF (USE_SHA1DC)
+IF ( SHA1_BACKEND STREQUAL "AUTO" )
+	IF ( OPENSSL_FOUND )
+		SET(SHA1_BACKEND "OpenSSL")
+	ELSEIF (WIN32 AND NOT MINGW)
+		SET(SHA1_BACKEND "SHA1_WIN32")
+	ELSEIF (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
```
tiennou: Nitpick: ISTR the `${}` syntax is a leftover of an old era where CMake didn't automatically evaluate expressions in `IF` statements. Personally, I find it inconsistent to only use that syntax for `CMAKE_SYSTEM_NAME`, while all other expressions don't.

src/CMakeLists.txt
```diff
@@ -183,6 +179,8 @@ IF (USE_HTTPS)
 		LIST(APPEND LIBGIT2_LIBS ${COREFOUNDATION_LIBRARIES} ${SECURITY_LIBRARIES})
 		LIST(APPEND LIBGIT2_PC_LIBS ${COREFOUNDATION_LDFLAGS} ${SECURITY_LDFLAGS})
 	ELSEIF (HTTPS_BACKEND STREQUAL "OpenSSL")
+		FIND_PACKAGE(OpenSSL)
```
tiennou: My original intention was to have all the various "Is this available" tests outside of the main `USE_HTTPS` block (or at least at the beginning of it). The `mbedTLS` backend might be used as some kind of reference. I don't really care in any way, just stating what intentions where (and at the time, I refrained from hoisting that `FIND_PACKAGE` because of `AMIGA`).

pks-t: Dunno. Personally, I find it preferential to keep packages as close as possible to the location where they're actually used. Furthermore, your original implementation has the problem that you've added a `QUIET` afterwards -- so it can never be seen that we're actually checking for OpenSSL. On the other hand, if we remove that `QUIET`, then it's always being checked regardless of whether it is actually requested.

That's why, at least in my opinion, the change improves the situation.

src/CMakeLists.txt
pks-t: Noted, will fix tomorrow

pks-t:
Thanks for your review, @tiennou. Yeah, I'm also kind of unhappy with our current style-inconsistencies. I found that `IF ( CONDITION )` style somewhere else and just reused it.

So I definitly agree that we should fix that situation. I'm not a fan of rewriting all existing inconsistencies, though, as that may be a lot of code churn without any real advantage. We should cast set of guidelines into stone, though, like we do for our C code. Judging from our current use:

```
$ grep 'IF (' **/CMakeLists.txt | wc -l
106
$ grep 'IF(' **/CMakeLists.txt | wc -l
101
```

Meh, nearly a tie :/ I'd vote for the first style, though, of having conditions with a space in between `IF` and opening brace and without having surrounding spaces inside of the brace. That's the style which also matches our C style, so we'd be consistent across languages. But we should also be consistent with the `ENDIF()` statemente then.

```
$ grep 'ENDIF(' **/CMakeLists.txt
74
$ grep 'ENDIF (' **/CMakeLists.txt
20
```

So to be honest, I'm rather indifferent to that topic, despite the fact that I want consistent guidelines. Any opinions?

tiennou:
I'm inclined to go for `IF()` just for the sake of not having `ENDIF ()` 😉.

pks-t:
I've fixed the stylistic inconsistencies. I've adopted `IF(CONDITION)`-style conditions.

pks-t:
Unrelated BadSSL error. I feel like this is good to go

src/CMakeLists.txt
tiennou: I have no hard opinion about it, since doing the check isn't that expensive, and might actually help pinpointing a non-detection event of OpenSSL (it has a helpful "pass `OPENSSL_ROOT_DIR`, which `QUIET` swallows). In the end, it's up to the `HTTPS` block to actually decide whether to use the found package or not.

ethomson:
git has made the collision detecting SHA1 backend [the default](https://github.com/git/git/commit/e6b07da2780f349c29809bd75d3eca6ad3c35d19), which means that we probably should, too.

I suggest that we drop the `auto` option and let people specify an explicit backend if they don't want sha1dc.

pks-t:
Agreed. Only makes this easier to implement. I've split out changing the default into a separate commit.

ethomson:
Sorry for the delay here - I'd love it if we could standardize a bit on the names of the implementations.  Some are MixedCase, generic is all-lower and `SHA1_WIN32` is all-uppercase and has the `SHA1` prefix.  How about:

- CommonCrypto
- OpenSSL
- Win32
- Generic
- CollisionDetecting

I'm "meh" on the last one, but since it's the default, probably nobody will type it anyway.  :P

pks-t:
I actually couldn't care less about the naming here, so I'll just do as you wish :P

pks-t:
Rebased to fix conflicts and adjusted naming.

ethomson:
Thanks!  ✨

PR #4403, opened 2017-11-11
cmake: Allow user to select bundled zlib

hkleynhans:
Under some circumstances the installed / system version of zlib may not
be desirable due to being too old or buggy.  This patch adds the option
`USE_BUNDLED_ZLIB` that will cause the bundled version of zlib to be
used.

We may also want to add similar functionality to allow the user to
select other bundled 3rd-party dependencies instead of using the system
versions.

/cc @pks-t @ethomson


src/CMakeLists.txt
```diff
@@ -255,19 +255,26 @@ ELSE()
...
+		ENDIF()
+		ADD_FEATURE_INFO(zlib ON "Zlib support")
+	ENDIF()
+ENDIF()
+IF (USE_BUNDLED_ZLIB OR NOT ZLIB_FOUND)
+	IF (USE_BUNDLED_ZLIB)
+	    MESSAGE(STATUS "zlib: user opted to use bundled 3rd-party sources." )
 	ELSE()
-		SET(LIBGIT2_PC_REQUIRES "${LIBGIT2_PC_REQUIRES} zlib")
+	    MESSAGE(STATUS "zlib was not found; using bundled 3rd-party sources." )
```
pks-t: Please use spaces here to indent that message

src/CMakeLists.txt
```diff
@@ -255,19 +255,26 @@ ELSE()
...
 
 # Optional external dependency: zlib
-FIND_PACKAGE(ZLIB)
-IF (ZLIB_FOUND)
-	LIST(APPEND LIBGIT2_INCLUDES ${ZLIB_INCLUDE_DIRS})
-	LIST(APPEND LIBGIT2_LIBS ${ZLIB_LIBRARIES})
-	IF(APPLE OR CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
-		LIST(APPEND LIBGIT2_LIBS "z")
-		LIST(APPEND LIBGIT2_PC_LIBS "-lz")
+IF (NOT USE_BUNDLED_ZLIB)
```
pks-t: Not a fault of yours, as we have just decided to enforce a consistent CMake style: please do not put any spaces between `IF(CONDITION)` instructions

src/CMakeLists.txt
```diff
@@ -255,19 +255,26 @@ ELSE()
...
+	FIND_PACKAGE(ZLIB)
+	IF (NOT USE_BUNDLED_ZLIP AND ZLIB_FOUND)
```
pks-t: `NOT USE_BUNDLED_ZLIB` is not required here, as it's already a prerequisite to getting here.

src/CMakeLists.txt
```diff
@@ -255,19 +255,26 @@ ELSE()
...
+		LIST(APPEND LIBGIT2_INCLUDES ${ZLIB_INCLUDE_DIRS})
+		LIST(APPEND LIBGIT2_LIBS ${ZLIB_LIBRARIES})
+		IF(APPLE OR CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
+			LIST(APPEND LIBGIT2_LIBS "z")
+			LIST(APPEND LIBGIT2_PC_LIBS "-lz")
+		ELSE()
+			SET(LIBGIT2_PC_REQUIRES "${LIBGIT2_PC_REQUIRES} zlib")
+		ENDIF()
+		ADD_FEATURE_INFO(zlib ON "Zlib support")
```
pks-t: Better: `ADD_FEATURE_INFO(zlib ON "using system zlib")`

src/CMakeLists.txt
```diff
@@ -255,19 +255,26 @@ ELSE()
...
+	ENDIF()
+ENDIF()
+IF (USE_BUNDLED_ZLIB OR NOT ZLIB_FOUND)
+	IF (USE_BUNDLED_ZLIB)
+	    MESSAGE(STATUS "zlib: user opted to use bundled 3rd-party sources." )
```
pks-t: I don't think this message is required. The user will probably know when he's opted in to that

src/CMakeLists.txt
```diff
@@ -255,19 +255,26 @@ ELSE()
...
 	ELSE()
-		SET(LIBGIT2_PC_REQUIRES "${LIBGIT2_PC_REQUIRES} zlib")
+	    MESSAGE(STATUS "zlib was not found; using bundled 3rd-party sources." )
 	ENDIF()
-	ADD_FEATURE_INFO(zlib ON "Zlib support")
-ELSE()
```
pks-t: Would be nice to have a `ADD_FEATURE_INFO(zlib ON "using bundled zlib")`

src/CMakeLists.txt
hkleynhans: I think the rest of the file use tabs.  I have replaced by spaces with a tab which should look better.

src/CMakeLists.txt
```diff
@@ -255,23 +255,28 @@ ELSE()
...
+			LIST(APPEND LIBGIT2_PC_LIBS "-lz")
+		ELSE()
+			SET(LIBGIT2_PC_REQUIRES "${LIBGIT2_PC_REQUIRES} zlib")
+		ENDIF()
+		ADD_FEATURE_INFO(zlib ON "using system zlib")
+	ENDIF()
+ENDIF()
+IF(USE_BUNDLED_ZLIB OR NOT ZLIB_FOUND)
+	IF(NOT USE_BUNDLED_ZLIB)
+		MESSAGE(STATUS "zlib was not found; using bundled 3rd-party sources." )
```
pks-t: Sorry to be nitpicky here, didn't realize that in the first review .I'd love if this was actually part of the `IF(ZLIB_FOUND) ... ELSE()` branch, starting at line 270.

hkleynhans: Ah yes, that makes sense! 👍

ethomson:
👍

PR #4405, opened 2017-11-11
add a more verbose mode to the test runner with better progress reporting

gaborbernat:
```bash
Loaded 340 suites, with 2504 tests, starting

attr::ignore [#1 with 18 tests]
  honor_temporary_rules [#1] .
  allow_root [#2] .
  ignore_space [#3] .
  ignore_root [#4] .
  full_paths [#5] F
  more_starstar_cases [#6] .
  leading_stars [#7] .
  globs_and_path_delimiters [#8] .
  skip_gitignore_directory [#9] .
  subdirectory_gitignore [#10] .
  expand_tilde_to_homedir [#11] .
  gitignore_in_subdir [#12] .
  dont_ignore_files_for_folder [#13] .
  symlink_to_outside [#14] S
  test [#15] .
  unignore_dir_succeeds [#16] .
  case_insensitive_unignores_previous_rule [#17] .
  case_sensitive_unignore_does_nothing [#18] .

  1) Failure:
attr::ignore::full_paths [C:\Users\berna\git\libgit2\tests\attr\ignore.c:83]
  expected != is_ignored
  0 != 1


ran 18 [16 passed 1 skipped 1 failed] in 1 suits in 2.9370 seconds
```

### DIFF
--- a/tests/clar.c
+++ b/tests/clar.c
@@ -86,6 +86,7 @@
 	typedef struct stat STAT_T;
 #endif
 
+#include "util.h"
 #include "clar.h"
 
 static void fs_rm(const char *_source);
@@ -124,6 +125,9 @@ static struct {
 	int exit_on_error;
 	int report_suite_names;
 
+	int verbosity;
+	double evaluation_duration;
+
 	struct clar_error *errors;
 	struct clar_error *last_error;
 
@@ -154,10 +158,10 @@ struct clar_suite {
 
 /* From clar_print_*.c */
 static void clar_print_init(int test_count, int suite_count, const char *suite_names);
-static void clar_print_shutdown(int test_count, int suite_count, int error_count);
+static void clar_print_shutdown(const double evaluation_duration, int test_count, int suite_count, int error_count, int skip_count);
 static void clar_print_error(int num, const struct clar_error *error);
 static void clar_print_ontest(const char *test_name, int test_number, enum cl_test_status failed);
-static void clar_print_onsuite(const char *suite_name, int suite_index);
+static void clar_print_onsuite(const char *suite_name, int suite_index, size_t tests_in_suite);
 static void clar_print_onabort(const char *msg, ...);
 
 /* From clar_sandbox.c */
@@ -259,7 +263,7 @@ clar_run_suite(const struct clar_suite *suite, const char *filter)
 		return;
 
 	if (!_clar.report_errors_only)
-		clar_print_onsuite(suite->name, ++_clar.suites_ran);
+		clar_print_onsuite(suite->name, ++_clar.suites_ran, suite->test_count);
 
 	_clar.active_suite = suite->name;
 	_clar.active_test = NULL;
@@ -312,6 +316,8 @@ static void
 clar_parse_args(int argc, char **argv)
 {
 	int i;
+
+	_clar.verbosity = 0;
 
 	/* Verify options before execute */
 	for (i = 1; i < argc; ++i) {
@@ -394,7 +400,7 @@ clar_parse_args(int argc, char **argv)
 		}
 
 		case 'v':
-			_clar.report_suite_names = 1;
+			_clar.verbosity += 1;
 			break;
 
 		default:
@@ -440,21 +446,26 @@ void
 clar_test_shutdown(void)
 {
 	clar_print_shutdown(
+		_clar.evaluation_duration,
 		_clar.tests_ran,
-		(int)_clar_suite_count,
-		_clar.total_errors
+		_clar.suites_ran,
+		_clar.total_errors,
+		_clar.total_skipped
 	);
 
 	clar_unsandbox();
 }
 
 int
-clar_test(int argc, char **argv)
+clar_test(int argc, char *argv[])
 {
+	(int)argc;
+	(char**)argv;
 	int errors;
+	double start = git__timer();
 
-	clar_test_init(argc, argv);
 	errors = clar_test_run();
+	_clar.evaluation_duration = git__timer() - start;
 	clar_test_shutdown();
 
 	return errors;
--- a/tests/clar/print.h
+++ b/tests/clar/print.h
@@ -1,19 +1,35 @@
 
 static void clar_print_init(int test_count, int suite_count, const char *suite_names)
 {
-	(void)test_count;
-	printf("Loaded %d suites: %s\n", (int)suite_count, suite_names);
-	printf("Started (test status codes: OK='.' FAILURE='F' SKIPPED='S')\n");
+	(void)suite_names;
+	printf("Loaded %d suites, with %d tests, starting\n", suite_count, test_count);
+	printf("test status codes: OK='.' FAILURE='F' SKIPPED='S'\n");
 }
 
-static void clar_print_shutdown(int test_count, int suite_count, int error_count)
+static void clar_print_shutdown(const double evaluation_duration,
+    int test_count, int suite_count, int error_count, int skip_count)
 {
-	(void)test_count;
-	(void)suite_count;
-	(void)error_count;
-
+	int success_count;
 	printf("\n\n");
 	clar_report_errors();
+
+	if (_clar.verbosity > 1) {
+		if (test_count == 0)
+			printf("no tests found");
+		else {
+			printf("\nran %d [", test_count);
+			success_count = test_count - skip_count - error_count;
+			if (success_count)
+				printf("%d passed", success_count);
+			if (skip_count)
+				printf(" %d skipped", skip_count);
+			if (error_count)
+				printf(" %d failed", error_count);
+			printf("] in %d suits", suite_count);
+			printf(" in %.4f seconds", evaluation_duration);
+		}
+		printf("\n");
+	}
 }
 
 static void clar_print_error(int num, const struct clar_error *error)
@@ -37,8 +53,8 @@ static void clar_print_error(int num, const struct clar_error *error)
 
 static void clar_print_ontest(const char *test_name, int test_number, enum cl_test_status status)
 {
-	(void)test_name;
-	(void)test_number;
+	if (_clar.verbosity > 1)
+		printf("\n  %s [#%d] ", test_name, test_number);
 
 	switch(status) {
 	case CL_TEST_OK: printf("."); break;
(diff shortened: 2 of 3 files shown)


tests/clar.c
```diff
@@ -124,6 +125,8 @@ static struct {
 	int exit_on_error;
 	int report_suite_names;
 
+	int verbose;
```
pks-t: As this doesn't only specify whether it's verbose or not but also how verbose it is, I think like `verbosity` would be a better name here

tests/clar.c
```diff
@@ -124,6 +124,8 @@ static struct {
 	int exit_on_error;
```
pks-t: Putting this at the topmost: the commit message is a bit off, I think. Better would be something like 

```
clar: add option to increase verbosity even 

Here comes the rationale for why you did that change and
what changes because of that. It looks like you could even
copy your PR description for most of that.
```

tests/clar.c
```diff
@@ -419,37 +423,33 @@ clar_test_run(void)
 {
 	if (_clar.argc > 1)
 		clar_parse_args(_clar.argc, _clar.argv);
-
 	if (!_clar.suites_ran) {
 		size_t i;
 		for (i = 0; i < _clar_suite_count; ++i)
 			clar_run_suite(&_clar_suites[i], NULL);
 	}
-
```
pks-t: Please avoid unnecessary code churn at places that aren't really touched by your patch.

tests/clar.c
```diff
@@ -419,37 +423,33 @@ clar_test_run(void)
...
 	return _clar.total_errors;
 }
 
 void
-clar_test_shutdown(void)
+clar_test_shutdown()
```
pks-t: Huh? This looks strange. Could you please elaborate on that change?

tests/clar.c
```diff
@@ -419,37 +423,33 @@ clar_test_run(void)
...
 {
 	clar_print_shutdown(
 		_clar.tests_ran,
-		(int)_clar_suite_count,
-		_clar.total_errors
+		_clar.suites_ran,
+		_clar.total_errors,
+		_clar.total_skipped
```
pks-t: Might be worth it to just pass in the complete `clar` struct?

tests/clar.c
```diff
@@ -419,37 +423,33 @@ clar_test_run(void)
...
 	clar_unsandbox();
 }
 
 int
-clar_test(int argc, char **argv)
+clar_test(void)
 {
 	int errors;
-
-	clar_test_init(argc, argv);
```
pks-t: Why is this being removed? Looks like it should stay here after all to parse our arguments. At least I couldn't see any other place where it's been added instead

tests/clar.h
```diff
@@ -17,9 +17,9 @@ enum cl_test_status {
 
 void clar_test_init(int argc, char *argv[]);
 int clar_test_run(void);
-void clar_test_shutdown(void);
+void clar_test_shutdown();
```
pks-t: We prefer to keep the `void` in function signatures, as it should be done for C-style method signatures. Otherwise, the function can actually accept arguments

tests/clar/print.h
```diff
@@ -1,19 +1,28 @@
...
-	(void)test_count;
-	(void)suite_count;
-	(void)error_count;
-
+	int success_count;
 	printf("\n\n");
 	clar_report_errors();
+
+	if (_clar.verbose > 1) {
+	    if (test_count == 0) printf("no tests found");
```
pks-t: Please split that up into two lines.

tests/clar/print.h
```diff
@@ -1,19 +1,28 @@
...
+	    else {
+		    printf("\nran %d [", test_count);
+		    success_count = test_count - skip_count  - error_count;
+		    if (success_count) printf("%d passed", success_count);
+		    if (skip_count) printf(" %d skipped", skip_count);
+		    if (error_count) printf(" %d failed", error_count);
```
pks-t: Same as above, please split those up into multiple lines

tests/clar/print.h
```diff
@@ -37,8 +46,7 @@ static void clar_print_error(int num, const struct clar_error *error)
 
 static void clar_print_ontest(const char *test_name, int test_number, enum cl_test_status status)
 {
-	(void)test_name;
-	(void)test_number;
+	if (_clar.verbose > 1)  printf("\n  %s [#%d] ", test_name, test_number);
```
pks-t: Same as above

tests/clar/print.h
```diff
@@ -49,12 +57,12 @@ static void clar_print_ontest(const char *test_name, int test_number, enum cl_te
...
-static void clar_print_onsuite(const char *suite_name, int suite_index)
+static void clar_print_onsuite(const struct clar_suite *suite, int suite_index)
 {
-	if (_clar.report_suite_names)
-		printf("\n%s", suite_name);
-
-	(void)suite_index;
+	if (_clar.verbose > 0 || _clar.report_suite_names)
+		printf("\n%s", suite->name);
+	if(_clar.verbose > 1)
```
pks-t: There's a space missing here after the `if` statement

tests/main.c
```diff
@@ -21,9 +21,7 @@ int main(int argc, char *argv[])
 	cl_sandbox_set_search_path_defaults();
 
 	/* Run the test suite */
-	res = clar_test_run();
-
-	clar_test_shutdown();
+	res = clar_test();
```
pks-t: This seems like a change which isn't really related to your actual topic. If you think this should be done regardless, please put it in a separate commit and explain why we should use that way instead.

tests/clar.c
```diff
@@ -432,9 +433,10 @@ clar_test_run(void)
 }
 
 void
-clar_test_shutdown()
+clar_test_shutdown(const double elapsed)
```
pks-t: To me, it looks like the elapsed time could also be part of the `_clar` structure, no? That'd also avoid piling up all that function arguments in `clar_print_shutdown`, as we can simply pass in `_clar` there.

pks-t:
Thanks for your PR! Code wise it looks mostly good to me, but there's some work to do regarding formatting

tests/clar.c
gaborbernat: I'm generally not too big fan of that, seems just a global variable; but given that that's the main direction we're going into we can do that :+1:

tests/clar.c
gaborbernat: tried to not change how it worked before (additional code churn) but OK, I'll amend it!

tests/clar.c
gaborbernat: because this function was never used (see how things work in the main), mainly because we  added the tracing setup (some code coverage would have caught this). Removing this init here allowed us to use this function in the main.

tests/clar.c
gaborbernat: actually we can't do that, the print.h file where this function is defined does not know about the ``_clar`` struct and its type

gaborbernat:
@pks-t can you take a look? I'll break up this into commits once the content looks good :+1:

gaborbernat:
@pks-t ping?

tests/clar.c
pks-t: Okay, fair enough

.gitignore
```diff
@@ -31,3 +31,5 @@ msvc/Release/
 .*.swp
 tags
 mkmf.log
+.idea
+cmake-*
```
pks-t: This shouldn't be part of your commit ;)

tests/clar.c
```diff
@@ -440,21 +446,24 @@ void
...
+		_clar.total_errors,
+		_clar.total_skipped
 	);
 
 	clar_unsandbox();
 }
 
 int
-clar_test(int argc, char **argv)
+clar_test(void)
```
pks-t: I'm still on the edge regarding this change. Just assuming that clar might be used elsewhere (which I don't really know of) this would break the API for them. So I'm opposed to changing the signature here without us (that is, the libgit2 project) having decided whether to diverge from upstream or not. The git__timer code is obviously already diverging, but that's just a minor thing compared to also breaking the API.

tests/clar/print.h
```diff
@@ -1,19 +1,34 @@
...
-	(void)test_count;
-	printf("Loaded %d suites: %s\n", (int)suite_count, suite_names);
-	printf("Started\n");
+	(void)suite_names;
+	printf("Loaded %d suites, with %d tests, starting\n", suite_count, test_count);
 }
 
-static void clar_print_shutdown(int test_count, int suite_count, int error_count)
+static void clar_print_shutdown(const double evaluation_duration,
+								int test_count, int suite_count, int error_count, int skip_count)
```
pks-t: I think indentation is a bit funny here

tests/clar.c
```diff
@@ -440,21 +446,24 @@ void
...
 {
 	int errors;
+	double start = git__timer();
 
-	clar_test_init(argc, argv);
 	errors = clar_test_run();
+	_clar.evaluation_duration = git__timer() - start;
```
pks-t: As said, I'd like to retain the current signature. If doing so, you could still easily put the timer-code into `clar_test_run`.

tests/clar.c
gaborbernat: done 👍

gaborbernat:
@pks-t take a look again please, thanks!

tests/clar.c
```diff
@@ -440,21 +446,26 @@ void
...
 int
-clar_test(int argc, char **argv)
+clar_test(int argc, char *argv[])
 {
+	(int)argc;
+	(char**)argv;
 	int errors;
+	double start = git__timer();
 
-	clar_test_init(argc, argv);
```
pks-t: Why do you stop calling clar_test_init here?

tests/clar.c
```diff
@@ -440,21 +446,26 @@ void
...
 }
 
 int
-clar_test(int argc, char **argv)
+clar_test(int argc, char *argv[])
 {
+	(int)argc;
+	(char**)argv;
 	int errors;
+	double start = git__timer();
```
pks-t: I still think this should actually be part of clar_test_run, as otherwise the timer will only ever work when clar_test is being called.

tests/clar.c
```diff
@@ -440,21 +446,26 @@ void
...
+		_clar.total_errors,
+		_clar.total_skipped
 	);
 
 	clar_unsandbox();
 }
 
 int
-clar_test(int argc, char **argv)
+clar_test(int argc, char *argv[])
```
pks-t: Why change the signature here?

PR #4406, opened 2017-11-11
Fix clar to verify command line arguments before execute

yoney:
Minor fix from git event at Bloomberg London.

Description from the board:

> When executing for example `libgit2_clar -smerge --help`, it will first execute the merge test suite and afterwards output help. This should be improved by first inspecting all parameters and outputting help if it is being requested, exiting without actually executing the test suites.

### DIFF
--- a/tests/clar.c
+++ b/tests/clar.c
@@ -313,11 +313,18 @@ clar_parse_args(int argc, char **argv)
 {
 	int i;
 
+	/* Verify options before execute */
 	for (i = 1; i < argc; ++i) {
 		char *argument = argv[i];
 
-		if (argument[0] != '-')
+		if (argument[0] != '-' || argument[1] == '\0'
+		    || strchr("sixvqQl", argument[1]) == NULL) {
 			clar_usage(argv[0]);
+		}
+	}
+
+	for (i = 1; i < argc; ++i) {
+		char *argument = argv[i];
 
 		switch (argument[1]) {
 		case 's':
@@ -391,7 +398,7 @@ clar_parse_args(int argc, char **argv)
 			break;
 
 		default:
-			clar_usage(argv[0]);
+			assert(!"Unexpected commandline argument!");
 		}
 	}
 }


tests/clar.c
```diff
@@ -313,11 +313,18 @@ clar_parse_args(int argc, char **argv)
 {
 	int i;
 
+	// Verify options before execute
```
pks-t: We are strictly tied to C90, so you'd have to use `/* */` style comments.

tests/clar.c
```diff
@@ -389,9 +396,6 @@ clar_parse_args(int argc, char **argv)
 		case 'v':
 			_clar.report_suite_names = 1;
 			break;
-
-		default:
-			clar_usage(argv[0]);
```
ethomson: Can you keep a default here?  An `assert` is fine, just something so that in case somebody adds an option to the `strchr` validation (above) during development that we won't simply barge on through without validation.

tests/clar.c
ethomson: This is a nitpick, but would you mind changing this to C-style `/*` comments `*/`?  We don't use `//` within the project.

(I know - I can't imagine that anybody's using a compiler that doesn't accept `//` style comments these days _but_ we try to be generous about that stuff.)

pks-t (changes requested):
Thanks for that PR, it looks perfectly reasonable. One small nit about C++-style comment. It'd also be nice to improve the commit message a bit. E.g.:

```
clar: verify command line arguments before execute

This is a description of the actual behaviour that we don't want. And some reasoning about why your changes improve upon that situation.
```

Please be sure to wrap at 80 characters. I'm too lazy to do so because GitHub ;)

ethomson (commented):
I'm generally 👍 on this.  I added to minor nitpicking requests, if you wouldn't mind addressing those, this is otherwise great.

pks-t:
Thanks a lot!

PR #4407, opened 2017-11-11
merge: reload index before git_merge

### LINKED ISSUE #4203: git_merge needs to reload the index
### Reproduction steps
* Do something that makes libgit2 load the index
* Do something to the index from the command line (like commit) so that the index & worktree are clean
* Call `git_merge()`

### Expected behavior
Merge proceeds, recognizing the clean state

### Actual behavior
Using a stale index, git_merge aborts with `GIT_ECONFLICT`

### Workaround
Manually refreshing the index produces the expected behavior.

0.25.1

macOS 10.12.3

Etiene:
When the index in memory diverges from the index in disk `git_merge`
aborts with `GIT_ECONFLICT`. More details about this are found in the 
issue #4203, which this PR attempts to fix.

A further discussion is needed wether this behaviour is actually desired.
(Suppose someone actually is writing something to the index in memory
and a concurrent task is merging. A possible argument is that the index
should be written to disk then.)

Closes #4203

### DIFF
--- a/src/merge.c
+++ b/src/merge.c
@@ -3239,9 +3239,6 @@ int git_merge(
 
 	assert(repo && their_heads);
 
-	git_repository_index(&index, repo);
-	git_index_read(index, 0);
-
 	if (their_heads_len != 1) {
 		giterr_set(GITERR_MERGE, "can only merge a single branch");
 		return -1;
@@ -3256,6 +3253,10 @@ int git_merge(
 
 	if ((error = git_indexwriter_init_for_operation(&indexwriter, repo,
 		&checkout_strategy)) < 0)
+		goto done;
+
+	if ((error = git_repository_index(&index, repo) < 0) ||
+	    (error = git_index_read(index, 0) < 0))
 		goto done;
 
 	/* Write the merge setup files to the repository. */
--- a/src/merge.c
+++ b/src/merge.c
@@ -3239,6 +3239,9 @@ int git_merge(
 
 	assert(repo && their_heads);
 
+	git_repository_index(&index, repo);
+	git_index_read(index, 0);
+
 	if (their_heads_len != 1) {
 		giterr_set(GITERR_MERGE, "can only merge a single branch");
 		return -1;
--- a/tests/merge/workdir/simple.c
+++ b/tests/merge/workdir/simple.c
@@ -168,6 +168,35 @@ void test_merge_workdir_simple__automerge(void)
 	git_index_free(index);
 }
 
+void test_merge_workdir_simple__index_reload(void)
+{
+	git_repository *tmp_repo;
+	git_annotated_commit *their_heads[1];
+	git_oid their_oid;
+	git_index_entry entry = {{0}};
+	git_index *tmp_index;
+
+	cl_git_pass(git_repository_open(&tmp_repo, git_repository_workdir(repo)));
+	cl_git_pass(git_repository_index(&tmp_index, tmp_repo));
+	cl_git_pass(git_index_read(repo_index, 0));
+
+	entry.mode = GIT_FILEMODE_BLOB;
+	cl_git_pass(git_oid_fromstr(&entry.id, "11deab00b2d3a6f5a3073988ac050c2d7b6655e2"));
+	entry.path = "automergeable.txt";
+	cl_git_pass(git_index_add(repo_index, &entry));
+
+	cl_git_pass(git_index_add_bypath(tmp_index, "automergeable.txt"));
+	cl_git_pass(git_index_write(tmp_index));
+
+	cl_git_pass(git_oid_fromstr(&their_oid, THEIRS_SIMPLE_OID));
+	cl_git_pass(git_annotated_commit_lookup(&their_heads[0], repo, &their_oid));
+	cl_git_pass(git_merge(repo, (const git_annotated_commit **)their_heads, 1, NULL, NULL));
+
+	git_index_free(tmp_index);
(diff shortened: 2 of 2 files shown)


tests/merge/workdir/simple.c
```diff
@@ -168,6 +168,36 @@ void test_merge_workdir_simple__automerge(void)
 	git_index_free(index);
 }
 
+void test_merge_workdir_simple__index_reload(void)
+{
+	git_repository *tmp_repo;
+	git_annotated_commit *their_heads[1];
+	git_oid their_oid;
+
+	git_index_entry entry = {{0}};
```
pks-t: We usually don't have newlines in the list of variable declarations.

ethomson:
One note: @pks-t and I have been discussing correct behavior here and whether we should always reload the index from disk.  (Like checkout does.)

It's worth noting that merge's behavior is to create a _new_ index (that represents the merge results) and check _that_ out, setting it as the new index.

During discussion, we were concerned that somebody may want to `git_index_add` an entry and expect it to be preserved even though they had not done a `git_index_write`.  This change might remove that functionality.

After further thoughts, _it does not_.  AFAICT, there is no case where somebody could `git_index_add` something, and _not_ write the changes to the index, and have those preserved.  That's because the on-disk index is overwritten with the merge results index.  So unsaved index changes are lost (today) and this does nothing to change that behavior one way or another.

src/merge.c
```diff
@@ -3239,6 +3239,10 @@ int git_merge(
 
 	assert(repo && their_heads);
 
+	if((error = git_repository_index(&index, repo) < 0) ||
```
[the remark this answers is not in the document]
ethomson: Actually, I think that the error message set by `git_repository_index` itself would be superior to changing the error message.  It will have the OS error details, like `permission denied`, which we would not want to overwrite with a less descriptive error.

ethomson: This actually needs to move down in the code a bit, write _after_ `git_indexwriter_init_for_operation`.

`git_indexwriter` will take set up the lockfile on the index.  We should re-read the index under that lock - if we do not, there's still a race condition where we read the index, and somebody makes changes before we take the lock.

pks-t:
@ethomson Unfortunately, I don't think that's actually true. First, there's the `GIT_CHECKOUT_DONT_WRITE_INDEX` flag, which gets evaluated by the index writer. Second, it's not really about the writing but about the reading of the index from disk. If we have a modified index, the indexwriter will simply use that, and necessarily so. If there are changes which are not written to disk, then currently the index writer will _not_ re-read that from the index but will simply do a `git_repository_index__weakptr` and get the currently modified one.

So re-reading the index from disk _does_ change behaviour in some cases. Whether those cases are cases are sane or not is an entirely different topic, though.

Etiene:
@pks-t please re-review the code style? :)

@ethomson is this where you wanted it moved? :)

Concerning the desired behaviour, let me know when the council has made a decision ;)

ethomson:
> So re-reading the index from disk does change behaviour in some cases. Whether those cases are cases are sane or not is an entirely different topic, though.

Correct.  I was totally mistaken about that.

I'm going to introduce a test during index refresh paths that ensures that we do not have unwritten index entries (ie someone has called `git_index_add` but _not_ `git_index_write`) to ensure that we do not lose their changes and can refresh safely.

Etiene:
Summary of future improvements based on IRL discussion: 
* erring on dirty index
* still reloading on stale

ethomson:
I've been working on adding a guard to the index so that you can't run a command that reloads the index when the index is "dirty" - which is to say that you've made changes via `git_index_add` and friends but you haven't saved those changes (with `git_index_write`) or aborted them (with `git_index_read`).

Regrettably, we have a _bunch_ of tests that believe that they can work in the old way and not flush the index to disk.  Sensibly, because why should they?  That would just slow our test runs down.

I've mostly got everything fixed here, but there are still a few tests failing.  I'm chipping away at these, and once that's done, I think we'll be good to merge this.

Etiene:
@ethomson wonderful! Do you need help in that?

ethomson:
Now that #4536 has landed, I think we can move forward with this.  I have one code review point.  And we should add some data to the changelog, something like:

> Merge now reloads the index before proceeding, to ensure that it has the latest changes that were made by other clients.  In the unlikely event that you have made changes to the index and not written them before calling `git_merge`, those changes may now be lost.  You should ensure that your changes are saved by calling `git_index_write`.  If you would like an error in the case where your index entries would be overwritten, you can set the `GIT_OPT_ENABLE_UNSAVED_INDEX_SAFETY` option.

src/merge.c
```diff
@@ -3255,6 +3255,10 @@ int git_merge(
 		&checkout_strategy)) < 0)
 		goto done;
 
+	if ((error = git_repository_index(&index, repo) < 0) ||
+	    (error = git_index_read(index, 0) < 0))
```
ethomson: We're leaking this index now.  I would encourage us to use a separate variable here, eg:

```
if ((error = git_repository_index(&repo_index, repo) < 0) ||
    (error = git_index_read(repo_index, 0) < 0))
```

And be sure to `git_index_free` this `repo_index`.

ethomson:
@Etiene do you have interest in finishing this up?  Or would you like me to fix those two last little bits?

ethomson:
I added a commit to fix the memory leak and merged this manually.  Thanks again for doing this, @Etiene!

PR #4408, opened 2017-11-12
signature: distinguish +0000 and -0000 UTC offsets

hkleynhans:
Git considers '-0000' a valid offset for signature lines.  They need to
be treated as _not_ equal to a '+0000' signature offset.  Parsing a
signature line stores the offset in a signed integer which does not
distinguish between `+0` and `-0`.

This patch adds an additional flag `sign` to the `git_time` in the
`signature` object which is populated with the sign of the offset.  In
addition to exposing this information to the user, this information is
also used to compare signatures.

/cc @pks-t @ethomson

### DIFF
--- a/tests/commit/signature.c
+++ b/tests/commit/signature.c
@@ -1,4 +1,5 @@
 #include "clar_libgit2.h"
+#include "signature.h"
 
 static int try_build_signature(const char *name, const char *email, git_time_t time, int offset)
 {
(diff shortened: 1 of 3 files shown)


tests/commit/signature.c
```diff
@@ -99,3 +100,30 @@ void test_commit_signature__from_buf(void)
...
+
+	cl_git_pass(git_signature_from_buffer(&with_neg_zero, "Test User <test@test.tt> 1461698487 -0000"));
+	cl_git_pass(git_signature_from_buffer(&with_pos_zero, "Test User <test@test.tt> 1461698487 +0000"));
+
+	cl_assert(!git_signature__equal(with_neg_zero, with_pos_zero));
+
+	git_signature_free((git_signature *)with_neg_zero);
+	git_signature_free((git_signature *)with_pos_zero);
+}
+
```
pks-t: Really nitpicky: there's a trailing newline here

pks-t:
Thanks a lot for your fixup!

hkleynhans:
👍

PR #4410, opened 2017-11-12
Getting started README improvements

ethomson:
Some improvements for people who are getting started with libgit2, either as contributors or simply as users building the project.  First, we bring a quick start guide to (nearly) the top of the file, outlining prerequisites and the quick build instructions.  Next, change test guidance to `ctest -V`, since it has superior output for people getting started.  Also, drop the paragraph about running `cmake` from the Visual Studio command prompt since it is inaccurate.  Finally, introduce an easily-greppable `TROUBLESHOOTING.md` with resolution steps.

These were all direct feedback or resolutions from problems witnessed at today's hackathon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,16 @@ function (`test_index_racy__diff`)[https://github.com/libgit2/libgit2/blob/maste
 
     $ ./libgit2_clar -sindex::racy::diff
 
+The test suite will print a `.` for every passing test, and an `F` for any
+failing test.  An `S` indicates that a test was skipped because it is not
+applicable to your platform or is particularly expensive.
+
+**Note:** There should be _no_ failing tests when you build an unmodified
+source tree from a [release](https://github.com/libgit2/libgit2/releases),
+or from the [master branch](https://github.com/libgit2/libgit2/tree/master).
+Please contact us or [open an issue](https://github.com/libgit2/libgit2/issues)
+if you see test failures.
+
 Installation
 ------------
 
--- a/README.md
+++ b/README.md
@@ -177,16 +177,6 @@ If you want to build a universal binary for Mac OS X, CMake sets it
 all up for you if you use `-DCMAKE_OSX_ARCHITECTURES="i386;x86_64"`
 when configuring.
 
-Windows
--------
-
-You need to run the CMake commands from the Visual Studio command
-prompt, not the regular or Windows SDK one. Select the right generator
-for your version with the `-G "Visual Studio X" option.
-
-See [the website](http://libgit2.github.com/docs/guides/build-and-link/)
-for more detailed instructions.
-
 Android
 -------
 
--- a/README.md
+++ b/README.md
@@ -142,6 +142,9 @@ and internal API/coding conventions we use.
 Building libgit2 - Using CMake
 ==============================
 
+Building
+--------
+
 `libgit2` builds cleanly on most platforms without any external dependencies.
 Under Unix-like systems, like Linux, \*BSD and Mac OS X, libgit2 expects `pthreads` to be available;
 they should be installed by default on all systems. Under Windows, libgit2 uses the native Windows API
@@ -157,6 +160,9 @@ On most systems you can build the library using the following commands
 
 Alternatively you can point the CMake GUI tool to the CMakeLists.txt file and generate platform specific build project or IDE workspace.
 
+Running Tests
+-------------
+
 Once built, you can run the tests from the `build` directory with the command
 
 	$ ctest -V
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -135,7 +135,7 @@ Alternatively you can point the CMake GUI tool to the CMakeLists.txt file and ge
```
pks-t: You could also argue about why it's better than executing `./libgit2_clar` directly. But yeah, I don't really mind

README.md
```diff
@@ -15,6 +15,27 @@ with any kind of software without having to release its source code.
 Additionally, the example code has been released to the public domain (see the
 [separate license](examples/COPYING) for more information).
 
+Quick Started
```
pks-t: Shouldn't that be "Quick Start"?

README.md
```diff
@@ -15,6 +15,27 @@ with any kind of software without having to release its source code.
...
+===============
+
+**Prerequisites** for building libgit2:
+
+1. [CMake](https://cmake.org/)
+2. [Python](https://www.python.org) is used by our test framework.
```
pks-t: Maybe specify that the user should make sure its in his path? Dunno if that is default for the Windows installer, but I think I remember there being at least an option for that

README.md
```diff
@@ -15,6 +15,27 @@ with any kind of software without having to release its source code.
...
+3. C compiler
+   * Windows: Visual Studio is recommended
+   * Mac: Xcode is recommended
```
pks-t: Shouldn't we also declare that on Linux, it is fine if you use either GCC or Clang? Otherwise they might feel like third-class citizen here ;)

README.md
```diff
@@ -15,6 +15,27 @@ with any kind of software without having to release its source code.
...
+
+**Build**
+
+1. Create a build directory beneath the libgit2 source directory, and change
+   into it: `mkdir build && cd build`
+2. Create the cmake build environment: `cmake ..`
+3. Build libgit2: `cmake --build .`
```
pks-t: These instructions may work just fine for users used to the command line. But maybe we should also add instructions for the CMake GUI?

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,14 @@
...
+
+CMake Failures
+--------------
+
+* **`Asked for OpenSSL TLS backend, but it wasn't found`**
+  CMake cannot find your SSL/TLS libraries.  By default, libgit2 always
+  builds with HTTPS support, and you are encouraged to install the
+  OpenSSL libraries for your system (eg, `apt-get install libssl-dev`).
+
+  For development, if you want to simply want to disable HTTPS support
```
pks-t: s/simply want to//

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,14 @@
...
+  entirely, pass the `-DUSE_HTTPS=OFF` to `cmake` when configuring it.
```
pks-t: Either remove the "the" or add "argument"/"option" here

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,14 @@
...
+
```
pks-t: Being nitpicky again: There's a trailing newline here

ethomson:
Fixed up per your review.

cpoerschke:
LGTM.

When running the `repo` tests yesterday at  the hackathon I was a little puzzled by the `S` in what I now know to be the test status codes output -- opened #4414 to perhaps avoid similar confusion for others in future.

ethomson:
Thanks @cpoerschke, I think that's a really helpful change.  I'll add a note here about the same thing just to expand a bit on it (ie, to explain why we skip tests.)

ethomson:
I added some information about test output (which is probably largely superfluous thanks to @cpoerschke's clar changes) and a wee bit of marketing.

ethomson:
I added some information about test output (which is probably largely superfluous thanks to @cpoerschke's clar changes) and a wee bit of marketing.

README.md
```diff
@@ -6,8 +6,19 @@ libgit2 - the Git linkable library
 [![Coverity Scan Build Status](https://scan.coverity.com/projects/639/badge.svg)](https://scan.coverity.com/projects/639)
 
 `libgit2` is a portable, pure C implementation of the Git core methods
-provided as a re-entrant linkable library with a solid API, allowing you to
-write native speed custom Git applications in any language with bindings.
+provided as a linkable library with a solid API, allowing to build Git
+functionality into your application.  Language bindings like
+[Rugged](https://github.com/libgit2/rugged),
```
pks-t: Why no mention of pygit2? It's rather popular and comprehensive, and Python is being used widely

ethomson: Sure, why not.

PR #4411, opened 2017-11-12
Config parser cleanups

pks-t:
Two small cleanups to make code more readable in our config parser.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -876,10 +876,8 @@ static char *escape_value(const char *ptr)
 		ptr++;
 	}
 
-	if (git_buf_oom(&buf)) {
-		git_buf_dispose(&buf);
+	if (git_buf_oom(&buf))
 		return NULL;
-	}
 
 	return git_buf_detach(&buf);
 }
@@ -1042,10 +1040,8 @@ static int read_on_variable(
 	for (c = var_name; *c; c++)
 		git_buf_putc(&buf, git__tolower(*c));
 
-	if (git_buf_oom(&buf)) {
-		git_buf_free(&buf);
+	if (git_buf_oom(&buf))
 		return -1;
-	}
 
 	entry = git__calloc(1, sizeof(git_config_entry));
 	GITERR_CHECK_ALLOC(entry);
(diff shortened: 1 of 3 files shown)


src/config_file.c
```diff
@@ -1056,36 +1056,38 @@ static int parse_conditional_include(git_config_parser *reader,
...
-	git_buf_printf(&buf, "%s.%s", current_section, var_name);
-	git__free(var_name);
+	name = git__strdup(var_name);
+	GITERR_CHECK_ALLOC(name);
 
-	if (git_buf_oom(&buf)) {
-		git__free(var_value);
+	git__strtolower(name);
+	git_buf_printf(&buf, "%s.%s", current_section, name);
+	git__free(name);
```
ethomson: It seems disappointing to have to do an extra tiny `malloc` here for `name` n every variable that we read and immediately `free` it.  Doing a little bit more manipulation and not doing a `printf` here might help.  eg:

```
git_buf_puts(&buf, current_section);
git_buf_putc(&buf, '.');

for (c = name, *c; c++)
    git_buf_putc(&buf, git__tolower(*c));
```

pks-t: Agreed. Your solution looks a lot cleaner, thanks!

pks-t:
Improved as to @ethomson's proposal

src/config_file.c
```diff
@@ -1056,26 +1056,28 @@ static int parse_conditional_include(git_config_parser *reader,
...
-	git_buf_printf(&buf, "%s.%s", current_section, var_name);
-	git__free(var_name);
+	git_buf_puts(&buf, current_section);
+	git_buf_putc(&buf, '.');
+	for (c = var_name; *c; c++)
+		git_buf_putc(&buf, git__tolower(*c));
 
 	if (git_buf_oom(&buf)) {
-		git__free(var_value);
+		git_buf_free(&buf);
```
carlosmn: There's no need to free an OOM'd buffer. If the realloc fails, we free the buffer we had before setting the buffer to the OOM one.

pks-t: Yeah, obviously. I've created a new commit for that, as there is also a second occasion where we had the same pattern nearby. Thanks for pointing out.

pks-t:
Rebased to fix conflicts

pks-t:
Rebased again to fix conflicts

ethomson:
🎉

PR #4414, opened 2017-11-13
add test status codes legend to the 'Started' test trace

cpoerschke:
motivation: (for someone new to the tests) it's puzzling to find the odd 'S' interspersed in the test output

proposed alternative test output (extract):

$ cmake --build . && ./libgit2_clar -srepo -v
...
Loaded 340 suites:

repo::config...
repo::discover..........
repo::env.
repo::getters...
repo::hashfile..
repo::head......................
repo::headtree....
repo::init.........................S
repo::message..
repo::new..
repo::open.............
repo::pathspec..........
repo::reservedname.....
repo::setters.....
repo::shallow....
repo::state.............

### DIFF
--- a/tests/clar/print.h
+++ b/tests/clar/print.h
@@ -3,7 +3,7 @@ static void clar_print_init(int test_count, int suite_count, const char *suite_n
 {
 	(void)test_count;
 	printf("Loaded %d suites: %s\n", (int)suite_count, suite_names);
-	printf("Started\n");
+	printf("Started (test status codes: OK='.' FAILURE='F' SKIPPED='S')\n");
 }
 
 static void clar_print_shutdown(int test_count, int suite_count, int error_count)


ethomson:
Yep, this is really helpful.

One wrinkle is that the unit test framework ([clar](https://github.com/vmg/vclar)) is actually a third-party tool that we include.

Welllllll that's not exactly true: it's actually something that was written explicitly for libgit2 but pulled out as a general purpose C test framework.

In any case, we try to keep our version of clar up-to-date with vmg's.  So it would be great if you could also open a PR over there with this same change, to make sure that we stay in sync?

cpoerschke:
Thanks for the quick feedback! I've opened https://github.com/vmg/clar/pull/79 as counter-part pull request.

ethomson:
Awesome!  Thanks!

PR #4416, opened 2017-11-16
WIP: travis: fail on memory leaks

ethomson:
Pass `--error-exitcode=125` to valgrind so that it will exit with a non-zero exit code when it detects a memory leak.  Use an exit code of `125` to stay safely above any clar errors and safely below any shell errors.

**Note**: I suspect that there are some tests that we'll need to either exclude from valgrind's examination, or perhaps not run at all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ script:
 
 # Run Tests
 after_success:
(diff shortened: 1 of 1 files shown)


pks-t:
We'll have to migrate the Valgrind execution out of `after_success`. Quoting Travis' documentation:

> The exit code of after_success, after_failure, after_script and subsequent stages do not affect the build result. However, if one of these stages times out, the build is marked as a failure.

pks-t:
I'm closing this in favour of #4558

PR #4417, opened 2017-11-17
[RFC] Khash refactorings (pt. 2)

pks-t:
Another round of khash fixes. While my first round of fixes only touched our own abstractions (git_oidmap, git_strmap, git_idxmap and git_offmap), this PR revamps the complete khash implementation. I've been too fed up with digging into those awful macros, so I decided to take the bait and decided to refactor that part.

So what I did is to make the khash implementation itself non-generic. Instead of having a macro to create per-type map structs, we now have a single map which has a void pointer only. This relieves us of all those macro magic, but possibliy at the cost of performance. I haven't measured that yet, but will do so if desired.

So this is mostly cosmetic only. I plan to do further refactorings in the future, as I think our map interfaces simply suck right now. They are not intuitive and do not follow our coding guidelines. This here is a first step, but seeing that it may be controverse I wanted to put up a RFC first.


pks-t:
Closing in favor of #4901

PR #4418, opened 2017-11-18
Re-enable the RC4 test

carlosmn:
We disabled this test while dealing with a security issue. Enable it again and make the error handling in the OpenSSL stream let us return the actual error due to trying to talk to RC4.

`GIT_ECERTIFICATE` is likely not _really_ the right thing to return here since it's not really about the certificate but about not supporting our cipherlist. We will likely want a specific error code for this (or we can overload this one, but TLS is confusing enough as it is).


pks-t:
The first commit looks obviously good. But it looks like the test still cannot be disabled, as all builds fail now.

PR #4420, opened 2017-11-18
refcount: make refcounting conform to aliasing rules

pks-t:
This should at least fix a huge swath of those new warnings in Travis. For now, I didn't care to fix the rest of them, and for most it is even impossible without rewriting our inheritance-structures. I mean we _could_ probably have e.g.

```
struct git_diff {
    ...
    void *child_data;
};

struct git_diff_parsed {
    struct git_diff base;
   ...
};

struct git_diff *git_diff_parsed_new(void)
{
    struct git_diff_parsed *parsed = malloc(sizeof(*parsed));
    parsed->base.child_data = parsed;
    return &parsed->base;
}
```

But I guess it's not worth it at all.

---

Strict aliasing rules dictate that for most data types, you are not
allowed to cast them to another data type and then access the casted
pointers. While this works just fine for most compilers, technically we
end up in undefined behaviour when we hurt that rule.

Our current refcounting code makes heavy use of casting and thus
violates that rule. While we didn't have any problems with that code,
Travis started spitting out a lot of warnings due to a change in their
toolchain. In the refcounting case, the code is also easy to fix:
as all refcounting-statements are actually macros, we can just access
the `rc` field directly instead of casting.

There are two outliers in our code where that doesn't work. Both the
`git_diff` and `git_patch` structures have specializations for generated
and parsed diffs/patches, which directly inherit from them. Because of
that, the refcounting code is only part of the base structure and not of
the children themselves. We can help that by instead passing their base
into `GIT_REFCOUNT_INC`, though.

### DIFF
--- a/src/diff_generate.c
+++ b/src/diff_generate.c
@@ -411,7 +411,7 @@ static git_diff_generated *diff_generated_alloc(
 	if ((diff = git__calloc(1, sizeof(git_diff_generated))) == NULL)
 		return NULL;
 
-	GIT_REFCOUNT_INC(diff);
+	GIT_REFCOUNT_INC(&diff->base);
 	diff->base.type = GIT_DIFF_TYPE_GENERATED;
 	diff->base.repo = repo;
 	diff->base.old_src = old_iter->type;
--- a/src/diff_parse.c
+++ b/src/diff_parse.c
@@ -36,7 +36,7 @@ static git_diff_parsed *diff_parsed_alloc(void)
 	if ((diff = git__calloc(1, sizeof(git_diff_parsed))) == NULL)
 		return NULL;
 
-	GIT_REFCOUNT_INC(diff);
+	GIT_REFCOUNT_INC(&diff->base);
 	diff->base.type = GIT_DIFF_TYPE_PARSED;
 	diff->base.strcomp = git__strcmp;
 	diff->base.strncomp = git__strncmp;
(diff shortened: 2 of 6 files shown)


carlosmn:
Yes, technically we should be passing pointers to the "parent" or embedded structs, but these casts are generally guaranteed to work regardless since the C standard says that the first field in a struct shares the address with the struct itself.

The reason strict aliasing is important is because it lets the compiler assume that two pointers of different types may not point to the same data, and can then avoid performing safety copies, helping performance. Our use is to cast first and then make use of the "real" type, so there shouldn't be any particular effect here, but it's a good pedantic habit to get into.

pks-t:
The thing here is GCC 6. Quoting from its release notes [1]:

> Type-based alias analysis now disambiguates accesses to different pointers. This improves precision of the alias oracle by about 20-30% on higher-level C++ programs. Programs doing invalid type punning of pointer types may now need -fno-strict-aliasing to work correctly.

So it seems like GCC disagrees with you here. I don't know whether their alias analysis is simply wrong or not, but at least the current state with warnings all over sucks. To fix that we _could_ simply pass in `-fno-strict-aliasing` as pointed out by them.

The thing is: this may even bite us. It's not about the first member being guaranteed to have the same address, but about the compiler who may optimize code in a breaking way. There are several horror stories about that when looking around, and I don't ever want to experience one of these where the compiler simply does the wrong thing (and even rightly so, because he may do whatever he wants with UB) because we break aliasing rules.

[1]: https://gcc.gnu.org/gcc-6/changes.html

PR #4421, opened 2017-11-18
Suggestion for code style fix.

fmmedeiros:
Adding brackets to show explicitly which else belongs to which if statement.

### DIFF
--- a/deps/regex/regexec.c
+++ b/deps/regex/regexec.c
@@ -369,6 +369,7 @@ re_search_2_stub (struct re_pattern_buffer *bufp,
 
   /* Concatenate the strings.  */
   if (length2 > 0)
+    {
     if (length1 > 0)
       {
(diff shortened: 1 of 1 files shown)


ethomson:
Hi @fmmedeiros - thanks for the pull request.  This regex code is not "ours" (as in, it didn't originate with us).  It's an external dependency from glibc.

When we include external dependencies, we try to bring them in unmodified, and we try not to change them once they're checked in.  The only changes that we make are for critical bugs or security vulnerabilites.  This allows us to easily compare our version to the upstream version, and include new versions of the upstream dependency easily.

As a result, I would encourage you to make changes directly to the upstream dependency, so that we can include them here.  As a result, I"m going to close this pull request, but please let us know when your changes land in the upstream project so that we can include them here.

Thanks!

PR #4423, opened 2017-11-25
Include git2/worktree.h in git2.h

apnadkarni:
I'm not sure if worktree.h was intentionally left out of git2.h. Looks like an oversight since it is in fact documented.


pks-t:
Thanks a lot for your fix! This was indeed unintential.

PR #4424, opened 2017-11-26
Use the same cert checking payload in WinHTTP

tiennou:
As recently discussed in Slack, this makes the WinHTTP transport use the same user-specified payload as the other transports.


ethomson:
Thanks for fixing this!  😀

PR #4429, opened 2017-12-04
Do not attempt to check out submodule as blob when merging a submodule modify/deltete conflict

novalis:


ethomson:
Looks great, thanks for the fix!

PR #4430, opened 2017-12-05
Free OpenSSL peer certificate

tiennou:
First commit adds more gotos for "simplicity" (sorry about that :wink:).

I happened to stumble onto the documentation while investigating something else. Per SSL_get_peer_certificate docs:

```
The reference count of the X509 object is incremented by one, so that
it will not be destroyed when the session containing the peer certificate
is freed. The X509 object must be explicitly freed using X509_free().
```

### DIFF
--- a/src/streams/openssl.c
+++ b/src/streams/openssl.c
@@ -341,7 +341,7 @@ static int verify_server_cert(SSL *ssl, const char *host)
 	struct in6_addr addr6;
 	struct in_addr addr4;
 	void *addr;
-	int i = -1,j;
+	int i = -1, j, error = 0;
 
 	if (SSL_get_verify_result(ssl) != X509_V_OK) {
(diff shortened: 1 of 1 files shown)


pks-t:
The code still isn't the prettiest even with your refactoring, but probably better than before. At least we don't have to add the new `X509_free` multiple times, which is good.

The fix itself looks good, but you should make sure that `cert` is always initialized to `NULL` now. While that currently makes no difference as all code previous to the assignment to `cert` does not `goto cleanup`, this might change in the future and result in `free`ing an uninitialized pointer.

As always, thanks for this!

tiennou:
Fixed missing `NULL` initialization, and rebased ! Thanks for the review !

ethomson:
Thanks!  😁

PR #4431, opened 2017-12-06
libFuzzer: Fix a git_packfile_stream leak

lhchavez:
This change ensures that the git_packfile_stream object in
git_indexer_append() does not leak when the stream has errors.

Found using libFuzzer.

### DIFF
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -1119,6 +1119,9 @@ void git_indexer_free(git_indexer *idx)
 	if (idx == NULL)
 		return;
 
+	if (idx->have_stream)
+		git_packfile_stream_free(&idx->stream);
+
 	git_vector_free_deep(&idx->objects);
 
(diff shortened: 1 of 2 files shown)


lhchavez:
PTAL

src/indexer.c
```diff
@@ -1119,6 +1119,9 @@ void git_indexer_free(git_indexer *idx)
 	if (idx == NULL)
 		return;
 
+	if (idx->have_stream)
+		git_packfile_stream_free(&idx->stream);
```
pks-t: Meh, this function doesn't conform to our usual style. 'free' would usually denote the structure itself bein free'd as well, which would have segfaulted here. But it only calls inflateEnd, so this looks good.

ethomson:
Thanks for the great test, too!  😻

pks-t:
Thanks again for your PR!

PR #4432, opened 2017-12-06
libFuzzer: Fix missing trailer crash

lhchavez:
This change fixes an invalid memory access when the trailer is missing /
corrupt.

Found using libFuzzer.

### DIFF
--- a/tests/pack/indexer.c
+++ b/tests/pack/indexer.c
@@ -87,6 +87,23 @@ void test_pack_indexer__leaky(void)
 	git_indexer_free(idx);
 }
 
+void test_pack_indexer__missing_trailer(void)
+{
+	git_indexer *idx = 0;
+	git_transfer_progress stats = { 0 };
+
+	cl_git_pass(git_indexer_new(&idx, ".", 0, NULL, NULL, NULL));
+	/* Truncate a valid packfile */
+	cl_git_pass(git_indexer_append(
+		idx, out_of_order_pack, out_of_order_pack_len - 20, &stats));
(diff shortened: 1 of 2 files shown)


lhchavez:
PTAL

tests/pack/indexer.c
```diff
@@ -60,6 +68,22 @@ void test_pack_indexer__out_of_order(void)
...
 }
 
+void test_pack_indexer__missing_trailer(void)
+{
+	git_indexer *idx = 0;
+	git_transfer_progress stats = { 0 };
+
+	cl_git_pass(git_indexer_new(&idx, ".", 0, NULL, NULL, NULL));
+	cl_git_pass(git_indexer_append(
+		idx, missing_trailer_pack, missing_trailer_pack_len, &stats));
```
pks-t: I'd prefer if you used another pack where we already verify that we're able to parse it correctly and just trim the pack_len by 1. Like that, we make sure that we're really failing because of the damaged trailer and not because the pack itself is broken.

lhchavez: Good idea, done! Had to change it to -20, because for some reason -1 wasn't enough to trigger the leak.

lhchavez:
Re: fuzzing recipes, I was already in the process of cleaning them up to upload them in another PR, expect them soon!

pks-t:
This PR now conflicts with the other indexer test you've added in the now-merged PR #4431. Could you please rebase? It would be nice if you could squash in the fixup commit for the test while at it.

lhchavez:
Rebased+squashed.

pks-t:
Thanks a lot!

PR #4433, opened 2017-12-07
libFuzzer: Add build support and instructions

lhchavez:
This change adds support for building a fuzz target for exercising the
packfile parser, as well as documentation. It also runs the fuzzers in
Travis to avoid regressions.

PR #4434, opened 2017-12-08
libFuzzer: Initialize a hash in the config file initialization code

lhchavez:
Found using libFuzzer.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -291,7 +291,7 @@ static int config_is_modified(int *modified, struct config_file *file)
 {
 	git_config_file *include;
 	git_buf buf = GIT_BUF_INIT;
-	git_oid hash = {};
+	git_oid hash = {0};
 	uint32_t i;
 	int error = 0;
 
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -291,7 +291,7 @@ static int config_is_modified(int *modified, struct config_file *file)
 {
 	git_config_file *include;
 	git_buf buf = GIT_BUF_INIT;
-	git_oid hash;
+	git_oid hash = {};
 	uint32_t i;
 	int error = 0;
 


src/config_file.c
```diff
@@ -291,7 +291,7 @@ static int config_is_modified(int *modified, struct config_file *file)
 {
 	git_config_file *include;
 	git_buf buf = GIT_BUF_INIT;
-	git_oid hash;
+	git_oid hash = {};
```
pks-t: This fix surprises me. `hash` is only ever being used when `git_hash_buf` returned success, so we should be able to assume that `hash` is always initialized when it is being used. The only possibility I can imagine is that one of our hash implementations is not initializing the hash in certain circumstances, for example when `len == 0`. Which hash implementation were you using?

lhchavez: Hmm, not sure. I'd guess whatever's the default.

This was found by running https://github.com/libgit2/libgit2/pull/4433/files#diff-76bfefafc83020b0409259f6bffcc1e9 (which is based off of some of the examples) with the instructions in the accompanying `README.md` and using `memory` as sanitizer. You'll also need a newer version of OpenSSL (1.1k) since they used to have uninitialized memory accesses all over the place in older versions. You might also need to not use the supplied corpus, because for some yet unidentified reason, a call to `opendir` within libFuzzer is triggering an uninitialized memory assertion.

pks-t:
Well, so this is rather a problem in OpenSSL I guess. We do have a lot of places where we don't initialize git_oid, and I don't think we want to adjust every one of those to fix a bug that is actually present in OpenSSL. I'm inclined to say this here is a non-issue for libgit2 as it seems more like a false positive.

lhchavez:
Fair enough. OpenSSL is using an experimental [MSan](https://github.com/google/oss-fuzz/blob/master/projects/openssl/project.yaml), and ultimately I couldn't get MSan to work fully, so I'm willing to let this one go since it's not blocking enabling fuzzing CI / OSS-Fuzz.

PR #4435, opened 2017-12-08
libFuzzer: Prevent a potential shift overflow

lhchavez:
The type of |base_offset| in get_delta_base() is `git_off_t`, which is a
signed `long`. That means that we need to make sure that the 8 most
significant bits are zero (instead of 7) to avoid an overflow when it is
shifted by 7 bits.

Found using libFuzzer.

### DIFF
--- a/src/pack.c
+++ b/src/pack.c
@@ -934,18 +934,20 @@ git_off_t get_delta_base(
 	if (type == GIT_OBJ_OFS_DELTA) {
 		unsigned used = 0;
 		unsigned char c = base_info[used++];
-		base_offset = c & 127;
+		size_t unsigned_base_offset = c & 127;
 		while (c & 128) {
 			if (left <= used)
 				return GIT_EBUFS;
-			base_offset += 1;
-			if (!base_offset || MSB(base_offset, 8))
+			unsigned_base_offset += 1;
+			if (!unsigned_base_offset || MSB(unsigned_base_offset, 7))
 				return 0; /* overflow */
 			c = base_info[used++];
-			base_offset = (base_offset << 7) + (c & 127);
+			unsigned_base_offset = (unsigned_base_offset << 7) + (c & 127);
 		}
-		base_offset = delta_obj_offset - base_offset;
-		if (base_offset <= 0 || base_offset >= delta_obj_offset)
+		if ((size_t)delta_obj_offset <= unsigned_base_offset)
+			return 0; /* out of bound */
+		base_offset = delta_obj_offset - unsigned_base_offset;
+		if (base_offset >= delta_obj_offset)
 			return 0; /* out of bound */
 		*curpos += used;
 	} else if (type == GIT_OBJ_REF_DELTA) {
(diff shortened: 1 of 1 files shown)


lhchavez:
PTAL.

This one I'm not 100% sure this is the right fix. The other option being using an unsigned long to store the intermediate computation.

lhchavez:
ugh, ubsan also complains about the `base_offset + 1`. will wait for comments.

src/pack.c
```diff
@@ -939,7 +939,7 @@ git_off_t get_delta_base(
 			if (left <= used)
 				return GIT_EBUFS;
 			base_offset += 1;
-			if (!base_offset || MSB(base_offset, 7))
```
pks-t: To me, it looks like the author of this code already assumed `base_offset` being unsigned. Otherwise `base_offset += 1; if (!base_offset) /* Overflow */;` wouldn't make any sense at all. So I think using `size_t` instead and then having a check whether its value is greater than the maximum value for `git_off_t` should be good.

lhchavez: Using `size_t` now.

src/pack.c
```diff
@@ -934,18 +934,20 @@ git_off_t get_delta_base(
...
 			c = base_info[used++];
-			base_offset = (base_offset << 7) + (c & 127);
+			unsigned_base_offset = (unsigned_base_offset << 7) + (c & 127);
 		}
-		base_offset = delta_obj_offset - base_offset;
-		if (base_offset <= 0 || base_offset >= delta_obj_offset)
+		if ((size_t)delta_obj_offset <= unsigned_base_offset)
+			return 0; /* out of bound */
+		base_offset = delta_obj_offset - unsigned_base_offset;
+		if (base_offset >= delta_obj_offset)
```
pks-t: This cannot ever be true:

base_offset >= delta_obj_offset
<=> delta_obj_offset - unsigned_base_offset  >= delta_obj_offset

The above could only be true iff unsinged_base_offset is greater than delta_obj_offset or 0, where the first case was filtered out in line 947 and the second case shouldn't ever happen. So you could just reformulate this condition as unsingned_base_offset == 0.

lhchavez: Done

src/pack.c
```diff
@@ -934,19 +934,19 @@ git_off_t get_delta_base(
...
+			unsigned_base_offset += 1;
+			if (!unsigned_base_offset || MSB(unsigned_base_offset, 7))
 				return 0; /* overflow */
 			c = base_info[used++];
-			base_offset = (base_offset << 7) + (c & 127);
+			unsigned_base_offset = (unsigned_base_offset << 7) + (c & 127);
 		}
-		base_offset = delta_obj_offset - base_offset;
-		if (base_offset <= 0 || base_offset >= delta_obj_offset)
+		if (unsigned_base_offset == 0 || (size_t)delta_obj_offset <= unsigned_base_offset)
```
ethomson: Can `delta_obj_offset` ever be negative?

ethomson: A cursory glance through the code suggests that the answer to this question is _no_.  But along those lines, should we really return `0` here or should we raise an error in the out-of-bound case?

ethomson: Oh, `0` is an error, but seems to only be checked occasionally.  😢

ethomson:
Thanks for doing this - this appears to be an improvement.

PR #4437, opened 2017-12-08
hash: openssl: check return values of SHA1_* functions

pks-t:
The OpenSSL functions `SHA1_Init`, `SHA1_Update` and `SHA1_Final` all
return 1 for success and 0 otherwise, but we never check their return
values. Do so.

### DIFF
--- a/src/hash/hash_openssl.h
+++ b/src/hash/hash_openssl.h
@@ -23,21 +23,36 @@ struct git_hash_ctx {
 GIT_INLINE(int) git_hash_init(git_hash_ctx *ctx)
 {
 	assert(ctx);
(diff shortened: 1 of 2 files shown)


ethomson:
Agreed.  We should set an error message though, otherwise it will likely be sent empty down to the caller, which would be unhelpful.

pks-t:
Thanks @ethomson. I am now also setting an error message, even though I'm not using 'ERR_error_message' as I have no idea whether that is being set for SHA1 errors. While digging into its documentation, I also realized we are using 'ERR_error_message' in a thread-unsafe way, as we have it print into a shared static buffer. I fixed that issue as well.

PR #4438, opened 2017-12-08
Multiple hash algorithms

pks-t:
A first step towards an abstraction for the use of multiple of hash algorithms. It lays the groundwork for having `git_hash_ctx` dispatch to different hash algorithms based on which one is being requested. Next steps:

- extend `git_repository` by a `git_hash_algorithm` field
- implement required new algorithms
- extend `git_hash_init` and `git_hash_ctx_init` to accept an algorithm which is then used for hashing by that particular context only
- extend OIDs to be a union of all possible hashes
- scavenge through the tree and adjust all callers. In most cases we'd want to retain SHA1 (e.g. pack trailers, patch IDs, filebuffers), for others we'd simply use whatever the repository format states


pks-t:
Rebased to fix a whole lot of conflicts

pks-t:
Todo: remove this weird separation between `git_hash_init` and `git_hash_ctx_init`. It is only used by Win32 to initialize crypto providers, but this API simply sucks. We should just make sure to always call `git_hash_global_init` and `git_hash_sha1_global_init` and initialize the crypto provider there.

src/hash.h
```diff
@@ -10,32 +10,28 @@
...
-#elif defined(GIT_SHA1_MBEDTLS)
-# include "hash/sha1/mbedtls.h"
-#else
-# include "hash/sha1/generic.h"
-#endif
+typedef struct git_hash_ctx {
+	union {
+		git_hash_sha1_ctx sha1;
+	} ctx;
+	git_hash_algorithm alg;
```
tiennou: Maybe put that first ? I'm thinking we might want a `GIT_HASH_ALGORITHM_CUSTOM`, and having that appear first would allow a user-provided struct to be overlayed transparently ?

ethomson:
> Todo: remove this weird separation between git_hash_init and git_hash_ctx_init.

It may be weird, but the goal is to avoid allocations in `git_hash_init`.  Building a `git_hash_ctx` _may_ perform an allocation of an object that contains the hash state / accumulator registers / etc.  So this cannot be shared across multiple hashes at once (ie, across threads).  On Windows, the `BCryptCreateHash` can't be moved into the global initialization.  (The Windows hashing stuff does load libraries and such in the global startup.)

The current implementation was chosen to be as similar to the existing init / update / finish pattern as possible, just adding a reusable object on top that you need to allocate and free.  (I called that object a _context_, because our constructors are usually called "init" and needed to disambiguate between hash initialization.)

Indeed we could make `git_hash_init` do any necessary allocation work, and introduce a `git_reset` to clean it (ie, call the underlying hash't init method) and introduce a `git_hash_dispose`.  (Now that we've settle on that name, we should fix the `cleanup` name that we're using.)

Certainly moving to having a "hash context" that uses the `init` / `update` / `finish` paradigm was less code churn and the changes were more obvious when we introduced them.  I don't _love_ the current API, but I wanted to at least give you some background on why it is the way it is.

pks-t:
Thanks for the clarafication, @ethomson. I didn't realize that `hash_ctx_cng_init` actually allocates memory, so you've convinced me that the API makes sense. That being said, I think that the names suck and don't quite help to grasp the situation. My proposal would be to have:

- `git_hash_ctx_init` to initialize the context and potentially allocate memory (same)
- `git_hash_ctx_dispose` to free the context's memory (previously `git_hash_ctx_free`)
- `git_hash_update` to add bytes to compute the hash
- `git_hash_final` to compute the final hash
- `git_hash_clear` to reset the hash to its initial state (previously `git_hash_init`)

I think that renaming `git_hash_init` to `git_hash_clear` would make it a lot more obvious what the distinction between `git_hash_init` and `git_hash_ctx_init` are.

pks-t:
Rebased to fix conflicts. I've tried to reduce the scope a bit by only doing the split between generic and SHA1-specific hashing layers. I didn't do any API changes at all

pks-t:
By the way: I noticed that our "Generic" SHA1 implementation doesn't compile right now. This is being fixed in here, too, but it shows that we're not as thorough with testing the different implementations as we maybe should be. Or, alternatively, that we have too many backends and that some aren't used at all.

pks-t:
Rebased to fix conflicts

pks-t:
Rebased to fix conflicts. @ethomson, @tiennou: could we maybe get this merged soonish? This doesn't introduce any behaviour changes yet, but it's required to get started with the actual transition to SHA256. The goal was to get going in small chunks to make later parts easier to review.

src/hash.h
```diff
@@ -4,38 +4,38 @@
...
-# include "hash/hash_generic.h"
-#endif
-
 typedef struct {
 	void *data;
 	size_t len;
 } git_buf_vec;
 
+typedef enum {
+	GIT_HASH_ALGO_UNDEF = 0,
```
tiennou: Nitpick: `UNKNOWN` ? It's never really undefined, as it's either SHA1 or something we don't know. Semantics 😉.

src/hash.h
```diff
@@ -4,38 +4,38 @@
...
+	GIT_HASH_ALGO_SHA1,
+} git_hash_algo;
```
tiennou: (Remark: we've started adding `_t` suffixes to some of these, but I've yet to grasp the rules.)

src/hash/sha1/collisiondetect.h
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#ifndef INCLUDE_hash_sha1_collisiondetect_h__
```
tiennou: Any chance this could be subsumed as "private implementation detail" inside their respective C files (or `sha1.h`) ? I don't think we really care about accessing that hashing context, so apart from a struct-sizing/compilation POV, that would reduce the amount of "small files". Do ignore that if you have a compelling reason though.

src/hash/sha1/mbedtls.h
```diff
@@ -5,20 +5,15 @@
  * a Linking Exception. For full terms see the included COPYING file.
  */
 
-#ifndef INCLUDE_hash_mbedtld_h__
-#define INCLUDE_hash_mbedtld_h__
+#ifndef INCLUDE_hash_sha1_mbedtld_h__
```
tiennou: `mbedtld` 🙄.

src/hash.h
pks-t: Agreed

src/hash/sha1/collisiondetect.h
pks-t: Unfortunately, no. We allocate these structs on the stack, so their size needs to be known at compile time :/ Would've loved that, too

src/hash/sha1/mbedtls.h
pks-t: Oops, yes

pks-t: Heh, in fact I can point at you here :P

src/hash.h
pks-t: No idea. From a quick glance it looks like `_t` prefixes are more common, so I'll just add it

pks-t:
Fixed comments by @tiennou. Thanks for your review!

src/hash/sha1/mbedtls.h
tiennou: (the eyesballs were my own blame-acceptance 😉 )

src/hash/sha1/collisiondetect.h
tiennou: What about moving the required include + struct inside the #ifdef cascade in `sha1.h` ? At least the ones that are 10-20 lines (ie. keep win32.h separate) ? I think I prefer more "complete", straightforward #define blocks vs 3-4 smaller ones.

src/hash/sha1/mbedtls.h
pks-t: I cannot see any smileys, thus I didn't get that :P

pks-t:
Rebased to fix conflicts with #5055

pks-t:
Rebased to fix SHA1DC breakage

src/hash/sha1/collisiondetect.h
pks-t: I'd definitely agree if all of those struct declarations were small. But as you correctly point out, the "win32.h" one is a rather complex beast. I do have the goal of refactoring it in the future to internalize more of its implementation, but I'd like to keep that separate from this PR.

So for now, I think we should keep those headers separate to not have a hybrid of everything in "sha1.h" but not "win32.h".

pks-t:
I'm just going ahead now and merge this. It's been open for a rather long time, and I finally want to continue working on this. There's still the one commen from @tiennou about inlining struct decls into "sha1.h", but we might as well tackle that in a future iteration.

PR #4439, opened 2017-12-10
cmake: create a dummy file for Xcode

tiennou:
Otherwise Xcode will happily not-link our git2 target, resulting in a "missing file" error when building eg. examples

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -3,11 +3,6 @@ IF(DEBUG_POOL)
 ENDIF()
 ADD_FEATURE_INFO(debugpool GIT_DEBUG_POOL "debug pool allocator")
 
-# Add the features.h file as a dummy. This is required for Xcode
-# to successfully build the libgit2 library when using only
-# object libraries.
-SET(LIBGIT2_OBJECTS "${CMAKE_CURRENT_BINARY_DIR}/git2/sys/features.h")
-
 # This variable will contain the libraries we need to put into
 # libgit2.pc's Requires.private. That is, what we're linking to or
 # what someone who's statically linking us needs to link to.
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -404,6 +399,13 @@ SET(LIBGIT2_INCLUDES ${LIBGIT2_INCLUDES} PARENT_SCOPE)
 SET(LIBGIT2_LIBS ${LIBGIT2_LIBS} PARENT_SCOPE)
 SET(LIBGIT2_LIBDIRS ${LIBGIT2_LIBDIRS} PARENT_SCOPE)
 
+IF(XCODE_VERSION)
+	# This is required for Xcode to actually link the libgit2 library
+	# when using only object libraries.
+	FILE(WRITE ${CMAKE_CURRENT_BINARY_DIR}/dummy.c "")
+	LIST(APPEND LIBGIT2_OBJECTS ${CMAKE_CURRENT_BINARY_DIR}/dummy.c)
+ENDIF()
```
pks-t: You're only doing this after already exporting LIBGIT2_OBJECTS into the parent scope. So I guess all callers of this CMakeLists.txt file will not get that dummy file. Is this intended?

tiennou: I'm not sure what is achieved by those `PARENT_SCOPE`, is it so that `clar` can static-link against the whole library code ?

AFAICT, this makes the dummy file appear, it shouldn't impact anything other than the Xcode generator, and if my point above is true then I don't think it matters because `libgit2_clar` is an executable, and it has .c files of its own (which sidesteps the underlying Xcode issue).

If you're concerned, I can move the block just above those `SET` lines.

tiennou: I'm not sure what is achieved by those `PARENT_SCOPE`, is it so that `clar` can static-link against the whole library code ?

AFAICT, this makes the dummy file appear, it shouldn't impact anything other than the Xcode generator, and if my point above is true then I don't think it matters because `libgit2_clar` is an executable, and it has .c files of its own (which sidesteps the underlying Xcode issue).

If you're concerned, I can move the block just above those `SET` lines.

pks-t: Yeah, exactly. We do not want to link against the .so or .a file but instead to the object files directly such that we can use some internal functions.

I think your reasoning makes sense, so it should be a non-issue. Thanks.

pks-t:
Thanks for the workaround!

tiennou:
Sorry for the missed fix : since I was originally prodding Xcode, *and* CMake didn't actually cleanup the expected dylib, it appeared to work when I tested the first time.

PR #4443, opened 2017-12-12
Inflate large loose blobs

ethomson:
As reported at https://github.com/libgit2/libgit2sharp/issues/1515, we are unable to inflate large loose blobs (that are larger than 4GB).  (In that issue, we create a large loose blob but fail to read it immediately after that.)  The culprit, of course, is 32 bitness: in a number of places we either improperly truncate 64 bit values via casting, or fail to read/write in a loop for APIs that take blocks of data, attempting to give them the entire blob at a time.

Fix this by:
* Move the loose object backend to use the `git_zstream` API instead of using zlib directly.  The `git_zstream` API simplifies the common inflate or deflate experience and helps eliminate common problems like neglecting to loop to process the entire file.
* Update the win32 and CommonCrypto hash backends to loop over the entire file

Also add a test for reading/writing loose objects, guarded by an environment variable to avoid running by default.

### DIFF
--- a/src/hash/hash_common_crypto.h
+++ b/src/hash/hash_common_crypto.h
@@ -16,6 +16,8 @@ struct git_hash_ctx {
 	CC_SHA1_CTX c;
 };
 
+#define CC_LONG_MAX ((CC_LONG)-1)
+
 #define git_hash_global_init() 0
 #define git_hash_ctx_init(ctx) git_hash_init(ctx)
 #define git_hash_ctx_cleanup(ctx)
@@ -27,10 +29,21 @@ GIT_INLINE(int) git_hash_init(git_hash_ctx *ctx)
 	return 0;
 }
 
-GIT_INLINE(int) git_hash_update(git_hash_ctx *ctx, const void *data, size_t len)
+GIT_INLINE(int) git_hash_update(git_hash_ctx *ctx, const void *_data, size_t len)
 {
+	const unsigned char *data = _data;
+
 	assert(ctx);
-	CC_SHA1_Update(&ctx->c, data, len);
+
+	while (len > 0) {
+		CC_LONG chunk = (len > CC_LONG_MAX) ? CC_LONG_MAX : (CC_LONG)len;
+
+		CC_SHA1_Update(&ctx->c, data, chunk);
+
+		data += chunk;
+		len -= chunk;
+	}
+
 	return 0;
 }
 
--- a/src/object.c
+++ b/src/object.c
@@ -236,13 +236,22 @@ const char *git_object_type2string(git_otype type)
 
 git_otype git_object_string2type(const char *str)
 {
+	if (!str)
+		return GIT_OBJ_BAD;
+
+	return git_object_stringn2type(str, strlen(str));
+}
+
+git_otype git_object_stringn2type(const char *str, size_t len)
+{
 	size_t i;
 
-	if (!str || !*str)
+	if (!str || !len || !*str)
 		return GIT_OBJ_BAD;
 
 	for (i = 0; i < ARRAY_SIZE(git_objects_table); i++)
-		if (!strcmp(str, git_objects_table[i].str))
+		if (*git_objects_table[i].str &&
+			!git__prefixncmp(str, len, git_objects_table[i].str))
 			return (git_otype)i;
 
 	return GIT_OBJ_BAD;
--- a/src/odb_loose.c
+++ b/src/odb_loose.c
@@ -16,6 +16,7 @@
 #include "delta.h"
 #include "filebuf.h"
 #include "object.h"
+#include "zstream.h"
 
 #include "git2/odb_backend.h"
 #include "git2/types.h"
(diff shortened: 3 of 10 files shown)


ethomson:
lolol the build machines OOM trying to run this test as we inflate a very large blob into memory.  🙄

I'll add a streaming reader test, just to make sure that we can read 5GB of blobs into `/dev/null`, and I'll make sure that these segfaults on Linux are actually due to running out of memory before the OOM killer can get to us (and not us ignoring an error code or something).

src/object.c
```diff
@@ -250,7 +250,7 @@ git_otype git_object_stringn2type(const char *str, size_t len)
 		return GIT_OBJ_BAD;
 
 	for (i = 0; i < ARRAY_SIZE(git_objects_table); i++)
-		if (git_object_typeisloose(i) &&
+		if (*git_objects_table[i].str &&
```
pks-t: Ah, yeah, that line confused me a lot in the last commit. So probably this line change should be moved to the previous commit?

src/odb_loose.c
```diff
@@ -121,8 +121,9 @@ static size_t get_binary_object_header(obj_hdr *hdr, git_buf *obj)
 }
 
 static int parse_header(
-	obj_hdr *out, size_t *out_len, unsigned char *data, size_t data_len)
+	obj_hdr *out, size_t *out_len, unsigned char *_data, size_t data_len)
 {
+	const char *data = (char *)_data;
```
pks-t: Looks like this chunk belongs into "odb: support large loose objects"? You should also make the data parameter const.

src/odb_loose.c
```diff
@@ -152,6 +153,11 @@ static int parse_header(
 		size < 0)
 		goto on_error;
 
+	if ((uint64_t)size > SIZE_MAX) {
+		giterr_set(GITERR_OBJECT, "object is larger than available memory");
```
pks-t: This is technically not correct, as this is not about available memory (which we don't know about right now). Rather, this should be somethink like "object size exceeds address space"?

src/hash/hash_common_crypto.h
```diff
@@ -27,10 +29,21 @@ GIT_INLINE(int) git_hash_init(git_hash_ctx *ctx)
 	return 0;
 }
 
-GIT_INLINE(int) git_hash_update(git_hash_ctx *ctx, const void *data, size_t len)
+GIT_INLINE(int) git_hash_update(git_hash_ctx *ctx, const void *_data, size_t len)
 {
+	const unsigned char *data = _data;
```
pks-t: By the way, why did you add all those casts to the functions? Just curious

src/odb_loose.c
ethomson: Well, I think _that's_ not _technically_ correct either.  `size_t` doesn't reflect the address space, it reflects the size of the largest single object size.  But I think that this is _practically_ correct for the set of computers that are actually in use, and that this is a much more actionable error message.

src/object.c
ethomson: Yes, agreed.  I think I fixup'd the wrong commit here.

src/odb_loose.c
pks-t: Yeah, guess one should be a bit more precise when nagging about technical incorrectness :P

src/hash/hash_common_crypto.h
ethomson: xcode was whinging about a signed/unsigned mismatch on `CC_SHA1_Update`  so I decided to cast them to their proper types as defined by the implementations.

ethomson:
> I'll add a streaming reader test, just to make sure that we can read 5GB of blobs into /dev/null, and I'll make sure that these segfaults on Linux are actually due to running out of memory before the OOM killer can get to us (and not us ignoring an error code or something).

Oh, I remember way back when I wrote this comment, full of naivete.

Yes, I _would_ simply add a streaming reader test...  if we had a streaming reader.  Despite having a streaming reader API in the ODB layer, it turns out that we don't actually have any backends that implement it.

So I've now changed the streaming reader API a bit, to provide the type and length of the object when initializing the stream.  This would be a breaking API change - except, of course, that nobody is actually _using_ the streaming reader API since none of the backends support it.  So I feel pretty good breaking this API that nobody could be using.

Then I added streaming reader support to the loose object ODB backend.  I cleaned up a bit while I was in there to add some more tests for things like `read_header` which had no explicit tests (and in fact failed on some of the loose ODB test corpus), teach `read_header` how to cope with "packlike loose objects" (which were a weird loose object format that was attempted for a bit and will never actually be seen in the wild) and drop the odb_loose internal `zstream` abstraction layer that ultimately became `git_zstream`.

I didn't really expect this PR to become so large.  Alas.

ethomson:
> I didn't really expect this PR to become so large. Alas.

Actually, I'm going to limit this PR to the large loose blob handling.  I will follow up with a PR that includes the refactoring to include `readstream` support for loose odb backends.

src/odb_loose.c
ethomson: Fixed.

src/object.c
ethomson: Fixed.

ethomson:
The new tests were timing out the build during the valgrind run.  So I ve  added still another environment variable so that we can skip them on our CI runs.

Another option is to just stop running the tests marked "invasive" during CI entirely.  But I like running as many tests as is feasible during our CI run, even at the expense of bloating our guard environment variables.

I thought about spinning up a background thread to dump something to the console during the tests, but, you know, ugh.  I'm not sure which is worse: tests failing because they took more than 10 minutes to run, or tests that _just keep running_.

ethomson:
I'd like to unblock users who are facing this.  I'm going to merge this unless there are complaints.

pks-t:
Going down the rabbit hole once again

PR #4444, opened 2017-12-13
stransport: provide error message on trust failures

### LINKED ISSUE #4440: Missing giterr_set before returning an error in stransport.c + git error conventions for wrappers?
Hello! We've recently been seeing some weird error messages on OSX from [Cargo](https://github.com/rust-lang/cargo) along the lines of:

```
$ cargo install --git https://wrong.host.badssl.com
    Updating git repository `https://wrong.host.badssl.com`
error: failed to clone into: /Users/acrichton/.cargo/git/db/_empty-90dfa5825

Caused by:
  [7/-17] config value 'credential.helper' was not found
```

Oddly though on Linux the error is different!

```
$ cargo install --git https://wrong.host.badssl.com
    Updating git repository `https://wrong.host.badssl.com/`
error: failed to fetch into /home/alex/.cargo/git/db/_empty-90dfa5825b276e27

Caused by:
  [16/-17] hostname does not match certificate
```

The Linux behavior is what I'd expect to happen but the OSX behavior is also interesting! Now clearly this error message (and Cargo) is pretty far removed from libgit2, but I think I've managed to trace the bug back to libgit2.

On OSX one piece of the error is -17 which is the code of the error and corresponds to [`GIT_ECERTIFICATE`](https://github.com/libgit2/libgit2/blob/1bf173c33a217a2e4766d8c144b7935010dfa7d6/include/git2/errors.h#L45) which is what I'd expect! Unfortunately though the error message seems to be incorrect (other than the code). I *think* that the error message itself originally came from [git2-rs](https://github.com/alexcrichton/git2-rs), the Rust bindings for libgit2 specifically at one point where [it loads `credential.helper`](https://github.com/alexcrichton/git2-rs/blob/047c984d3b934e536fd4cdd54a691ceb233ce92a/src/cred.rs#L224) and presumably fails (but handles the error locally).

Stepping through in a debugger I found that the `GIT_ECERTIFICATE` error originates from [this location](https://github.com/libgit2/libgit2/blob/1bf173c33a217a2e4766d8c144b7935010dfa7d6/src/streams/stransport.c#L85-L87) where notably there is no invocation of `giterr_set`. When I print out the contents of `giterr_last` at that location it sure enough mentions the credential helper business. Also sure enough [in `openssl.c`](https://github.com/libgit2/libgit2/blob/1bf173c33a217a2e4766d8c144b7935010dfa7d6/src/streams/openssl.c#L457-L458) when `GIT_ECERTIFICATE` is returned a `giterr_set` happens!

So I think most of this just boils down to perhaps a missing call to `giterr_set` in the `stransport.c` bindings? Or maybe there's something else that should be configuring that message?

---

I'd also like to, while I can, ask a more general question as well. I don't think this is the first time we've run across reporting "stale errors" in that the error from one operation persists to accidentally get returned from a different operation. Is this normal and/or expected behavior from libgit2? Should wrappers like git2-rs always clear the error before any operation is invoked in libgit2?

tiennou:
Fixes #4440

### DIFF
--- a/src/streams/stransport.c
+++ b/src/streams/stransport.c
@@ -83,8 +83,10 @@ static int stransport_connect(git_stream *stream)
 	}
 
 	if (sec_res == kSecTrustResultDeny || sec_res == kSecTrustResultRecoverableTrustFailure ||
-	    sec_res == kSecTrustResultFatalTrustFailure)
+	    sec_res == kSecTrustResultFatalTrustFailure) {
+		giterr_set(GITERR_SSL, "untrusted connection error");
 		return GIT_ECERTIFICATE;
+	}
 
 	return 0;
 


src/streams/stransport.c
```diff
@@ -84,6 +84,7 @@ static int stransport_connect(git_stream *stream)
 
 	if (sec_res == kSecTrustResultDeny || sec_res == kSecTrustResultRecoverableTrustFailure ||
 	    sec_res == kSecTrustResultFatalTrustFailure)
```
alexcrichton: I think you may want braces for this if statement now that it's two expressions?

tiennou: Yeah right 😆 . I miss my `goto fail`s so much.

tiennou:
Fixed !

pks-t:
Thanks a lot @tiennou, this look obviously correct.

PR #4445, opened 2017-12-13
DRY commit parsing

tiennou:
Splitted from #4331. See this PR for context.

This adds a "quick" mode to our commit parsing machinery, and makes the revwalk code instead of going with its own. The outcome is a nice, single place where grafts can be handled.

### DIFF
--- a/src/commit.c
+++ b/src/commit.c
@@ -20,6 +20,7 @@
 #include "message.h"
 #include "refs.h"
 #include "object.h"
+#include "array.h"
 #include "oidarray.h"
 
 void git_commit__free(void *_commit)
@@ -383,14 +384,15 @@ int git_commit_amend(
 	return error;
 }
 
-int git_commit__parse_raw(void *_commit, const char *data, size_t size)
+static int commit_parse(git_commit *commit, const char *data, size_t size, unsigned int flags)
 {
-	git_commit *commit = _commit;
 	const char *buffer_start = data, *buffer;
 	const char *buffer_end = buffer_start + size;
 	git_oid parent_id;
 	size_t header_len;
 	git_signature dummy_sig;
+
+	assert(commit && data);
 
 	buffer = buffer_start;
 
@@ -399,8 +401,15 @@ int git_commit__parse_raw(void *_commit, const char *data, size_t size)
 	GIT_ERROR_CHECK_ARRAY(commit->parent_ids);
 
 	/* The tree is always the first field */
-	if (git_oid__parse(&commit->tree_id, &buffer, buffer_end, "tree ") < 0)
-		goto bad_buffer;
+	if (!(flags & GIT_COMMIT_PARSE_QUICK)) {
+	    if (git_oid__parse(&commit->tree_id, &buffer, buffer_end, "tree ") < 0)
+			goto bad_buffer;
+	} else {
+		size_t tree_len = strlen("tree ") + GIT_OID_HEXSZ + 1;
+		if (buffer + tree_len > buffer_end)
+			goto bad_buffer;
+		buffer += tree_len;
+	}
 
 	/*
 	 * TODO: commit grafts!
@@ -413,11 +422,13 @@ int git_commit__parse_raw(void *_commit, const char *data, size_t size)
 		git_oid_cpy(new_id, &parent_id);
 	}
 
-	commit->author = git__malloc(sizeof(git_signature));
-	GIT_ERROR_CHECK_ALLOC(commit->author);
+	if (!(flags & GIT_COMMIT_PARSE_QUICK)) {
+		commit->author = git__malloc(sizeof(git_signature));
+		GIT_ERROR_CHECK_ALLOC(commit->author);
 
-	if (git_signature__parse(commit->author, &buffer, buffer_end, "author ", '\n') < 0)
-		return -1;
+		if (git_signature__parse(commit->author, &buffer, buffer_end, "author ", '\n') < 0)
+			return -1;
+	}
 
 	/* Some tools create multiple author fields, ignore the extra ones */
 	while (!git__prefixncmp(buffer, buffer_end - buffer, "author ")) {
@@ -434,6 +445,9 @@ int git_commit__parse_raw(void *_commit, const char *data, size_t size)
 
 	if (git_signature__parse(commit->committer, &buffer, buffer_end, "committer ", '\n') < 0)
 		return -1;
+
+	if (flags & GIT_COMMIT_PARSE_QUICK)
+		return 0;
 
 	/* Parse add'l header entries */
 	while (buffer < buffer_end) {
@@ -477,11 +491,19 @@ bad_buffer:
 	return -1;
 }
 
+int git_commit__parse_raw(void *commit, const char *data, size_t size)
+{
+	return commit_parse(commit, data, size, 0);
+}
+
+int git_commit__parse_ext(git_commit *commit, git_odb_object *odb_obj, unsigned int flags)
+{
+	return commit_parse(commit, git_odb_object_data(odb_obj), git_odb_object_size(odb_obj), flags);
+}
+
 int git_commit__parse(void *_commit, git_odb_object *odb_obj)
 {
-	return git_commit__parse_raw(_commit,
-		git_odb_object_data(odb_obj),
-		git_odb_object_size(odb_obj));
+	return git_commit__parse_ext(_commit, odb_obj, 0);
 }
 
 #define GIT_COMMIT_GETTER(_rvalue, _name, _return) \
--- a/src/commit.h
+++ b/src/commit.h
@@ -37,4 +37,10 @@ void git_commit__free(void *commit);
 int git_commit__parse(void *commit, git_odb_object *obj);
 int git_commit__parse_raw(void *commit, const char *data, size_t size);
 
+typedef enum {
+	GIT_COMMIT_PARSE_QUICK = (1 << 0), /**< Only parse parents and committer info */
+} git_commit__parse_flags;
+
+int git_commit__parse_ext(git_commit *commit, git_odb_object *odb_obj, unsigned int flags);
+
 #endif
--- a/src/commit_list.c
+++ b/src/commit_list.c
@@ -114,8 +114,14 @@ static int commit_quick_parse(
 		return error;
 	}
 
+	if (!git__is_uint16(git_array_size(commit->parent_ids))) {
+		git__free(commit);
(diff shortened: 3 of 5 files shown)


src/commit_list.c
```diff
@@ -9,7 +9,10 @@
 
 #include "revwalk.h"
 #include "pool.h"
+#include "repository.h"
 #include "odb.h"
+#include "commit.h"
+#include "signature.h"
```
pks-t: I couldn't really see why repository.h and signature.h are required now. At least that's not obvious from the diff.

src/commit.c
```diff
@@ -19,6 +19,7 @@
 #include "message.h"
```
pks-t: I think the commits are out of order. You first use the new generic mechanism and introduce it only in the next commit

src/commit.c
```diff
@@ -382,41 +383,53 @@ int git_commit_amend(
...
-	if (git_oid__parse(&commit->tree_id, &buffer, buffer_end, "tree ") < 0)
-		goto bad_buffer;
+	if (flags & GIT_COMMIT_PARSE_ALL) {
+		if (git_oid__parse(&commit->tree_id, &buffer, buffer_end, "tree ") < 0)
+			goto bad_buffer;
+	} else {
+		buffer += strlen("tree ") + GIT_OID_HEXSZ + 1;
+	}
+
+	if (flags & (GIT_COMMIT_PARSE_ALL|GIT_COMMIT_PARSE_QUICK)) {
```
pks-t: Can't this condition just be removed? You're doing it for all flags anyway, so no real need to check for those flags.

src/commit.c
```diff
@@ -428,11 +441,16 @@ int git_commit__parse(void *_commit, git_odb_object *odb_obj)
 	}
 
 	/* Always parse the committer; we need the commit time */
-	commit->committer = git__malloc(sizeof(git_signature));
-	GITERR_CHECK_ALLOC(commit->committer);
+	if (flags & (GIT_COMMIT_PARSE_ALL|GIT_COMMIT_PARSE_QUICK)) {
```
pks-t: Same here -- no need for the conditional if you do so for all available flags anyway.

src/commit.h
```diff
@@ -36,4 +36,11 @@ struct git_commit {
 void git_commit__free(void *commit);
 int git_commit__parse(void *commit, git_odb_object *obj);
 
+typedef enum {
+	GIT_COMMIT_PARSE_QUICK = (1 << 0),
```
pks-t: You should add some documentation here stating what those two flags mean and what information exactly is not going to get parsed for QUICK.

src/commit.c
tiennou: Yes, I caused some historic wreckage when doing some rebases, so when date-sorting, the commits are reverted. I already tried to fix it but failed, so I think I'll end up extracting patches and reapplying manually.

src/commit.c
tiennou: I did that mostly for completeness — that makes it easier to spot the differences between 
`ALL` and `QUICK`, and also (somewhat) future-proof. I can remove though, if that's preferred.

src/commit.h
tiennou: Right, will do.

src/commit_list.c
tiennou: Leftover from a previous attempt, where I tried to make that parser use already existing functions, before I noticed the "main" parse code. Will remove if that's unwarranted (I don't think it is).

tiennou:
`git rebase master --ignore-date` did the trick w.r.t ordering. I'll investigate the W64 crash, but I'm a slight disadvantage here…

pks-t:
I've debugged and fixed the issue. You were underallocating the commit object due to using `git_odb_object_size` instead of `git_object__size`.

tiennou:
@pks-t Many thanks for the fix ! :shipit: 🌟

src/commit.c
```diff
@@ -382,41 +383,48 @@ int git_commit_amend(
...
+	assert(commit && odb_obj);
 
 	buffer = buffer_start;
 
+	/* The tree is always the first field */
+	if (flags & GIT_COMMIT_PARSE_ALL) {
+		if (git_oid__parse(&commit->tree_id, &buffer, buffer_end, "tree ") < 0)
+			goto bad_buffer;
+	} else {
+		buffer += strlen("tree ") + GIT_OID_HEXSZ + 1;
```
ethomson: Please check the length of `buffer` so that you make sure that you're not pointing off the end of it.  (Pointing out of bounds is UB, even if you don't dereference it.)

src/commit.c
```diff
@@ -382,41 +383,48 @@ int git_commit_amend(
 	return error;
 }
 
-int git_commit__parse(void *_commit, git_odb_object *odb_obj)
+int git_commit__parse_commit(git_commit *commit, git_odb_object *odb_obj, git_commit__parse_flags flags)
```
ethomson: I don't love the name, it's a bit redundant.  How about `git_commit__parse_ext`, which is in keeping with other functions that take "extended" argument lists.

src/commit.c
```diff
@@ -476,6 +490,17 @@ int git_commit__parse(void *_commit, git_odb_object *odb_obj)
...
+int git_commit__parse(void *_commit, git_odb_object *odb_obj)
+{
+	git_commit *commit = _commit;
+
+	int error = git_commit__parse_ext(commit, odb_obj, GIT_COMMIT_PARSE_ALL);
+	if (error < 0)
+		return error;
+
+	return 0;
+}
```
ethomson: Is there a case where `git_commit__parse_ext` will return a positive value that needs to be treated as success?  Reading this function in isolation, I wondered if perhaps `parse_ext` returned the length of the buffer parsed or something.  It appears to not, and this could be simplified as a one-liner `return git_commit__parse_ext...`

Not saying you need to do that, but I'm just trying to clarify my understanding, I guess, by typing into this little box.

tiennou:
There you go !

src/commit.h
```diff
@@ -37,4 +37,11 @@ void git_commit__free(void *commit);
 int git_commit__parse(void *commit, git_odb_object *obj);
 int git_commit__parse_raw(void *commit, const char *data, size_t size);
 
+typedef enum {
+	GIT_COMMIT_PARSE_QUICK = (1 << 0), /**< Only parse parents and committer info */
+	GIT_COMMIT_PARSE_ALL   = (1 << 1), /**< Parse all commit data */
```
pks-t: So... do these two flags really make sense? After all, you're not using `PARSE_QUICK` anywhere in  `git_commit__parse_internal` and in fact `PARSE_QUICK` and `PARSE_ALL` do not make any sense together, right? I guess having only one of them would be more sensible, and I lean towards retaining `PARSE_QUICK`. In that case, default behaviour would be to parse everything and by passing the optional `PARSE_QUICK` flag we would skip certain information

src/commit.c
```diff
@@ -383,24 +384,32 @@ int git_commit_amend(
 	return error;
 }
 
-int git_commit__parse_raw(void *_commit, const char *data, size_t size)
+static int git_commit__parse_internal(git_commit *commit, const char *data, size_t size, git_commit__parse_flags flags)
```
pks-t: Well, I now realize that this really accepts an enum, only. I first thought that this was a bitfield due to its naming (flags instead of flag). So another route to make this a tad easier to grasp might be to just rename it to `git_commit__parse_mode mode`, as it's not really about flags at all

pks-t:
Meh, there's some fallout from recent changes (e.g. changing to Werror on Win32, deprecation of `git_object__size`). I'll follow up shortly to fix these issues.

pks-t:
@ethomson: want to give this another round of reviews or may I just proceed and merge?

src/commit_list.c
```diff
@@ -107,77 +97,38 @@ git_commit_list_node *git_commit_list_pop(git_commit_list **stack)
...
-		commit->parents[i] = git_revwalk__commit_lookup(walk, &oid);
-		if (commit->parents[i] == NULL)
-			return -1;
+	commit = git__calloc(1, sizeof(*commit));
+	GIT_ERROR_CHECK_ALLOC(commit);
+	commit->object.repo = walk->repo;
 
-		buffer += parent_len;
+	if ((error = git_commit__parse_ext(commit, obj, GIT_COMMIT_PARSE_QUICK)) < 0 ||
+	    !git__is_uint16(git_array_size(commit->parent_ids))) {
```
ethomson: We should set an error message if `git__is_uint16` is false.

I’d recommend restructuring this a bit to preserve the parse error in case we want to start using more exotic error codes, but that’s rather optional.

ethomson:
Thanks for the ping.  One minor request but otherwise 👍

src/commit_list.c
```diff
@@ -107,77 +97,38 @@ git_commit_list_node *git_commit_list_pop(git_commit_list **stack)
...
-			return -1;
+	git_oid *parent_oid;
+	git_commit *commit;
+	int error;
+	size_t i;
 
-		commit->parents[i] = git_revwalk__commit_lookup(walk, &oid);
-		if (commit->parents[i] == NULL)
-			return -1;
+	commit = git__calloc(1, sizeof(*commit));
```
tiennou: I have a few "I don't remember" commits related to that, as fixups in the actual shallow implementation, [here](https://github.com/libgit2/libgit2/commit/004b923ab956d21df88e9e6189b39a486ca9dfc6) and [here](https://github.com/libgit2/libgit2/commit/086f72981ac3e7c9ea4cf3e7d9a217adde779502). AFAIR, I was getting asserts/crashes in cache.c, until I made the initialization like this. But it's fuzzy now, and I think I wasn't able to extract a testcase for it anyways. Now that I'm thinking about it, maybe that's because the ODB memory semantics being weird 😆 ?

src/commit_list.c
pks-t: Yup, thanks for catching! I now set an error message and also restructured to retain the error code.

src/commit_list.c
pks-t: They are! That's why we've deprecated the `git_object__size` thingy, because nobody was able to remember what it actually does. And in most cases, it did _not_ do what was intended.

pks-t:
@ethomson: amended to fix your remarks, thanks!

ethomson:
:squirrel:

pks-t:
Thanks a lot, both to you @tiennou for working on this and @ethomson for the review!

tiennou:
Thanks for last-mile touch ups @pks-t ! Glad to see duplicated code go away 😉.

PR #4446, opened 2017-12-13
Graft support

tiennou:
Depends on #4445. Sadly I can't open the PR against libgit2 *and* set that as the base branch, so do note that the first `commit:` commits come from that other PR.

This adds graft support to `git_repository` as a private API. The end goal is to be able to store and obey shallow roots when working with a shallow repository, but this is usable on its own.

As discussed on #4331, this is unrelated to the `git-replace` mechanism.


pks-t:
I don't think it's necessary to have this PR reviewed separately from #4331. #4331 doesn't add much on top of the grafts API, only some small bits of plumbing to wire it all up. I'm thus closing this PR for now in favour of #4331, which I'd like to land directly. @tiennou: if you disagree, please let me know and I'll immediately reopen this.

Issue #4442, opened 2017-12-11
git_patch_to_buf Does not work if underlying blob is freed


brandonio21:
This issue was discovered in pygit2 https://github.com/libgit2/pygit2/issues/754 and there are some pretty detailed reproduction steps there.

# Reproductions steps
Create a patch object from at least one blob. Free the blob. Print the patch to a buffer. The patch will be mangled because the underlying blob was freed.

# Expected behavior
A patch created from a blob should be able to print to a buffer without issue when the blob is freed.

In short, the patch should not need the blob to exist for it to print. The patch should hold a reference to the blob.

ethomson:
I agree - this is unfortunate as it's neither obvious nor documented.  But indeed this is the case at the moment.

We should either clearly document this or find a solution where we refcount the blob or blob data.  I'm in favor of the latter.

pks-t:
All our objects are already refcounted. We simply fail to use the mechanism correctly here. A PR will follow shortly

brandonio21:
Fixed in #4447 :)

PR #4447, opened 2017-12-15
diff_file: properly refcount blobs when initializing file contents

pks-t:
When initializing a `git_diff_file_content` from a source whose data is
derived from a blob, we simply assign the blob's pointer to the
resulting struct without incrementing its refcount. Thus, the structure
can only be used as long as the blob is kept alive by the caller.

Fix the issue by using `git_blob_dup` instead of a direct assignment.
This function will increment the refcount of the blob without allocating
new memory, so it does exactly what we want. As
`git_diff_file_content__unload` already frees the blob when
`GIT_DIFF_FLAG__FREE_BLOB` is set, we don't need to add new code
handling the free but only have to set that flag correctly.

### DIFF
--- a/src/diff_file.c
+++ b/src/diff_file.c
@@ -139,7 +139,6 @@ int git_diff_file_content__init_from_src(
 	memset(fc, 0, sizeof(*fc));
 	fc->repo = repo;
 	fc->file = as_file;
-	fc->blob = src->blob;
 
 	if (!src->blob && !src->buf) {
 		fc->flags |= GIT_DIFF_FLAG__NO_DATA;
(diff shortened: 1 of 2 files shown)


pks-t:
This should probably fix #4442. The test case I added didn't trigger any faults locally at my machine, so I'm not certain that it actually tests what I'd like it to test.

brandonio21:
Thanks for the quick fix! The changes and included test LGTM!

brandonio21:
Just tested it on my machine!

Without the fix:
```
  1) Failure:
diff::blob::patch_with_freed_blobs [/home/bmilton/libgit2-0.26.0/tests/diff/blob.c:101]
  String mismatch: buf.ptr != BLOB_DIFF
  'diff --git a/file b/file
index 45141a7..4d713dc 100644
--- a/file
+++ b/file
@@ -1 +1,6 @@
 p�' != 'diff --git a/file b/file
index 45141a7..4d713dc 100644
--- a/file
+++ b/file
@@ -1 +1,6 @@
 Hello from the root
+
+Some additional lines
+
+Down here below
+
' (at byte 92)
```

With the fix, all tests pass. It looks good to me! Let's merge this in!

ethomson:
Looks good to me.  😁 

Thanks @brandonio21 for debugging and the helpful bug report; thanks @pks-t for the fix.

PR #4448, opened 2017-12-18
Fix failure to fetch annotated tags added to existing commits

### LINKED ISSUE #3870: Fetch doesn't update the same tags as git
Fetch's default automatic tag behavior (GIT_REMOTE_DOWNLOAD_TAGS_AUTO) doesn't match the tags that git updates by default. Git documentation says: "By default, any tag that points into the histories being fetched is also fetched; the effect is to fetch tags that point at branches that you are interested in." In libgit2 we only get the tags that point to commits we're already downloading.

I think that the auto tag behavior should probably be changed to match git. I can propose a patch if that seems like a reasonable change to make.

hackhaslam:
This is an attempt to fix #3870.

### DIFF
--- a/src/fetch.c
+++ b/src/fetch.c
@@ -26,7 +26,7 @@ static int maybe_want(git_remote *remote, git_remote_head *head, git_odb *odb, g
 	if (!git_reference_is_valid_name(head->name))
 		return 0;
 
-	if (tagopt == GIT_REMOTE_DOWNLOAD_TAGS_ALL) {
+	if (tagopt != GIT_REMOTE_DOWNLOAD_TAGS_NONE) {
 		/*
 		 * If tagopt is --tags, always request tags
 		 * in addition to the remote's refspecs


pks-t:
This seems to break our tests in online::fetch. Do you have an interest in debugging that issue?

hackhaslam:
Yes, I will check it out.

jpwhiting:
Hey, I looked into this a bit and it seems the tests are incorrect. Here are two runs with some additional debug printfs added to see which refs are update_tips callback called on:

./libgit2_clar -sonline::fetch
Loaded 343 suites: 
after git create counter is now 0
update_tips called on ref: refs/remotes/test/first-merge
update_tips called on ref: refs/remotes/test/master
update_tips called on ref: refs/remotes/test/no-parent
update_tips called on ref: refs/tags/annotated_tag
update_tips called on ref: refs/tags/blob
update_tips called on ref: refs/tags/commit_tree
update_tips called on ref: refs/tags/nearly-dangling
after remote fetch, counter is now 7
Fafter git create counter is now 0
update_tips called on ref: refs/remotes/test/first-merge
update_tips called on ref: refs/remotes/test/master
update_tips called on ref: refs/remotes/test/no-parent
update_tips called on ref: refs/tags/annotated_tag
update_tips called on ref: refs/tags/blob
update_tips called on ref: refs/tags/commit_tree
update_tips called on ref: refs/tags/nearly-dangling
after remote fetch, counter is now 7
Fafter git create counter is now 0
update_tips called on ref: refs/remotes/test/first-merge
update_tips called on ref: refs/remotes/test/master
update_tips called on ref: refs/remotes/test/no-parent
update_tips called on ref: refs/tags/annotated_tag
update_tips called on ref: refs/tags/blob
update_tips called on ref: refs/tags/commit_tree
update_tips called on ref: refs/tags/nearly-dangling
after remote fetch, counter is now 7
Fafter git create counter is now 0
update_tips called on ref: refs/remotes/test/first-merge
update_tips called on ref: refs/remotes/test/master
update_tips called on ref: refs/remotes/test/no-parent
after remote fetch, counter is now 3
.after git create counter is now 0
update_tips called on ref: refs/remotes/test/first-merge
update_tips called on ref: refs/remotes/test/master
update_tips called on ref: refs/remotes/test/no-parent
after remote fetch, counter is now 3
.......

  1) Failure:
online::fetch::default_git [/Users/jeremy/devel/git/libgit2/tests/online/fetch.c:53]
  counter != n
  7 != 6

  2) Failure:
online::fetch::default_http [/Users/jeremy/devel/git/libgit2/tests/online/fetch.c:53]
  counter != n
  7 != 6

  3) Failure:
online::fetch::default_https [/Users/jeremy/devel/git/libgit2/tests/online/fetch.c:53]
  counter != n
  7 != 6


Now without the referenced patch you see the lack of nearly-dangling tags:

./libgit2_clar -sonline::fetch
Loaded 343 suites: 
after git create counter is now 0
update_tips called on ref: refs/remotes/test/first-merge
update_tips called on ref: refs/remotes/test/master
update_tips called on ref: refs/remotes/test/no-parent
update_tips called on ref: refs/tags/annotated_tag
update_tips called on ref: refs/tags/blob
update_tips called on ref: refs/tags/commit_tree
after remote fetch, counter is now 6
.after git create counter is now 0
update_tips called on ref: refs/remotes/test/first-merge
update_tips called on ref: refs/remotes/test/master
update_tips called on ref: refs/remotes/test/no-parent
update_tips called on ref: refs/tags/annotated_tag
update_tips called on ref: refs/tags/blob
update_tips called on ref: refs/tags/commit_tree
after remote fetch, counter is now 6
.after git create counter is now 0
update_tips called on ref: refs/remotes/test/first-merge
update_tips called on ref: refs/remotes/test/master
update_tips called on ref: refs/remotes/test/no-parent
update_tips called on ref: refs/tags/annotated_tag
update_tips called on ref: refs/tags/blob
update_tips called on ref: refs/tags/commit_tree
after remote fetch, counter is now 6
.after git create counter is now 0
update_tips called on ref: refs/remotes/test/first-merge
update_tips called on ref: refs/remotes/test/master
update_tips called on ref: refs/remotes/test/no-parent
after remote fetch, counter is now 3
.after git create counter is now 0
update_tips called on ref: refs/remotes/test/first-merge
update_tips called on ref: refs/remotes/test/master
update_tips called on ref: refs/remotes/test/no-parent
after remote fetch, counter is now 3
.......

The first run is with this patch, the second run is without this patch. As you can see without this patch the nearly-dangling tags are not fetched at all. While with it they are. I think the 6's in the first 3 tests of online::fetch test on lines 61, 66, and 71 respectively should be 7s to correctly count all the  branches and tags that should be fetched.

pks-t:
Thanks for your analysis, @jpwhiting. "refs/tags/nearly-dangling" is a blob which is not being transitively referenced by any of the other references. Previously, we were only fetching such tags in case where `GIT_REMOTE_DOWNLOADS_TAGS_ALL` is being passed, which is the right thing to do, I'd say. So this PR changes behaviour not only for annotated tags, but will also cause tags to be downloaded which do not point into history. Thus, changed behaviour is not the right thing to do here

jpwhiting:
Thanks, that makes sense to me.

PR #4450, opened 2017-12-19
Streaming read support for the loose ODB backend

ethomson:
Change the streaming reader API a bit, to provide the type and length of the object when initializing the stream. This would be a breaking API change - except, of course, that nobody is actually using the streaming reader API since none of the backends support it. So I feel pretty good breaking this API that nobody could be using.

Add streaming reader support to the loose object ODB backend. I cleaned up a bit while I was in there to add some more tests for things like read_header which had no explicit tests (and in fact failed on some of the loose ODB test corpus), teach read_header how to cope with "packlike loose objects" (which were a weird loose object format that was attempted for a bit and will never actually be seen in the wild) and drop the odb_loose internal zstream abstraction layer that ultimately became git_zstream.

Note that this depends on #4443.

### DIFF
--- a/src/odb_loose.c
+++ b/src/odb_loose.c
@@ -812,7 +812,7 @@ static int loose_backend__foreach(git_odb_backend *_backend, git_odb_foreach_cb 
 	return error;
 }
 
-static int loose_backend__stream_fwrite(git_odb_stream *_stream, const git_oid *oid)
+static int loose_backend__writestream_finalize(git_odb_stream *_stream, const git_oid *oid)
 {
 	loose_writestream *stream = (loose_writestream *)_stream;
 	loose_backend *backend = (loose_backend *)_stream->backend;
@@ -831,13 +831,13 @@ static int loose_backend__stream_fwrite(git_odb_stream *_stream, const git_oid *
 	return error;
 }
 
-static int loose_backend__stream_write(git_odb_stream *_stream, const char *data, size_t len)
+static int loose_backend__writestream_write(git_odb_stream *_stream, const char *data, size_t len)
 {
 	loose_writestream *stream = (loose_writestream *)_stream;
 	return git_filebuf_write(&stream->fbuf, data, len);
 }
 
-static void loose_backend__stream_free(git_odb_stream *_stream)
+static void loose_backend__writestream_free(git_odb_stream *_stream)
 {
 	loose_writestream *stream = (loose_writestream *)_stream;
 
@@ -856,7 +856,7 @@ static int filebuf_flags(loose_backend *backend)
 	return flags;
 }
 
-static int loose_backend__stream(git_odb_stream **stream_out, git_odb_backend *_backend, git_off_t length, git_otype type)
+static int loose_backend__writestream(git_odb_stream **stream_out, git_odb_backend *_backend, git_off_t length, git_otype type)
 {
 	loose_backend *backend;
 	loose_writestream *stream = NULL;
@@ -876,9 +876,9 @@ static int loose_backend__stream(git_odb_stream **stream_out, git_odb_backend *_
 
 	stream->stream.backend = _backend;
 	stream->stream.read = NULL; /* read only */
-	stream->stream.write = &loose_backend__stream_write;
-	stream->stream.finalize_write = &loose_backend__stream_fwrite;
-	stream->stream.free = &loose_backend__stream_free;
+	stream->stream.write = &loose_backend__writestream_write;
+	stream->stream.finalize_write = &loose_backend__writestream_finalize;
+	stream->stream.free = &loose_backend__writestream_free;
 	stream->stream.mode = GIT_STREAM_WRONLY;
 
 	if (git_buf_joinpath(&tmp_path, backend->objects_dir, "tmp_object") < 0 ||
@@ -1002,7 +1002,7 @@ int git_odb_backend_loose(
 	backend->parent.write = &loose_backend__write;
 	backend->parent.read_prefix = &loose_backend__read_prefix;
 	backend->parent.read_header = &loose_backend__read_header;
-	backend->parent.writestream = &loose_backend__stream;
+	backend->parent.writestream = &loose_backend__writestream;
 	backend->parent.exists = &loose_backend__exists;
 	backend->parent.exists_prefix = &loose_backend__exists_prefix;
 	backend->parent.foreach = &loose_backend__foreach;
--- a/src/odb_loose.c
+++ b/src/odb_loose.c
@@ -39,6 +39,7 @@ typedef struct {
 	git_map map;
 	char start[HEADER_LEN];
 	size_t start_len;
+	size_t start_read;
 	git_zstream zstream;
 } loose_readstream;
 
@@ -103,13 +104,13 @@ static int object_mkdir(const git_buf *name, const loose_backend *be)
 		GIT_MKDIR_PATH | GIT_MKDIR_SKIP_LAST | GIT_MKDIR_VERIFY_DIR, NULL);
 }
 
-static size_t get_binary_object_header(obj_hdr *hdr, git_buf *obj)
+static size_t get_binary_object_header(
+	obj_hdr *hdr, const unsigned char *data, size_t len)
 {
 	unsigned long c;
-	unsigned char *data = (unsigned char *)obj->ptr;
 	size_t shift, size, used = 0;
 
-	if (git_buf_len(obj) == 0)
+	if (len == 0)
 		return 0;
 
 	c = data[used++];
@@ -118,7 +119,7 @@ static size_t get_binary_object_header(obj_hdr *hdr, git_buf *obj)
 	size = c & 15;
 	shift = 4;
 	while (c & 0x80) {
-		if (git_buf_len(obj) <= used)
+		if (len <= used)
 			return 0;
 		if (sizeof(size_t) * 8 <= shift)
 			return 0;
@@ -302,7 +303,7 @@ static int inflate_packlike_loose_disk_obj(git_rawobj *out, git_buf *obj)
 	 * read the object header, which is an (uncompressed)
 	 * binary encoding of the object type and size.
 	 */
-	if ((used = get_binary_object_header(&hdr, obj)) == 0 ||
+	if ((used = get_binary_object_header(&hdr, (unsigned char *)obj->ptr, obj->size)) == 0 ||
 		!git_object_typeisloose(hdr.type)) {
 		giterr_set(GITERR_ODB, "failed to inflate loose object");
 		return -1;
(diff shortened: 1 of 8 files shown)


src/odb_loose.c
```diff
@@ -896,11 +907,139 @@ static int loose_backend__writestream(git_odb_stream **stream_out, git_odb_backe
...
+
+	*stream_out = (git_odb_stream *)stream;
+	*len_out = hdr.size;
+	*type_out = hdr.type;
+
+done:
+	if (error < 0) {
+		git_futils_mmap_free(&stream->map);
+		git_zstream_free(&stream->zstream);
+		git_hash_ctx_cleanup(hash_ctx);
```
pks-t: `git_hash_ctx_cleanup` will never free the context itself. (Btw, this is why I intend on distinguishing `free` and `cleanup`-style functions :P)

pks-t: Also: wouldn't it be easier to just have the hash context embedded inside of the stream itself? There's no need for the pointer, I think.

src/odb_loose.c
```diff
@@ -896,11 +907,139 @@ static int loose_backend__writestream(git_odb_stream **stream_out, git_odb_backe
...
+
+	return total;
+}
+
+static void loose_backend__readstream_free(git_odb_stream *_stream)
+{
+	loose_readstream *stream = (loose_readstream *)_stream;
+
+	git_futils_mmap_free(&stream->map);
+	git_zstream_free(&stream->zstream);
```
pks-t: Missing the `git_hash_ctx_cleanup` here. Also missing the `git__free` if you're not changing the hash context to instead be embedded directly.

src/odb_loose.c
```diff
@@ -910,28 +911,29 @@ static int loose_backend__writestream(git_odb_stream **stream_out, git_odb_backe
...
 	/*
 	 * if we read more than just the header in the initial read, play
 	 * that back for the caller.
 	 */
-	if (stream->start_len && len) {
-		size_t chunk = MIN(stream->start_len, len);
-		memcpy(buffer, stream->start, chunk);
+	if (start_remain && buffer_len) {
+		size_t chunk = min(start_remain, buffer_len);
+		memcpy(buffer, stream->start + stream->start_read, chunk);
```
pks-t: Ah, here it's getting fixed. Did you squash that into the wrong commit by accident?

src/odb_loose.c
```diff
@@ -910,28 +911,29 @@ static int loose_backend__writestream(git_odb_stream **stream_out, git_odb_backe
...
-		stream->start_len -= chunk;
-		len -= chunk;
+		total += chunk;
+		buffer_len -= chunk;
 	}
 
-	if (len) {
-		size_t chunk = MIN(len, INT_MAX);
+	if (buffer_len) {
+		size_t chunk = min(buffer_len, INT_MAX);
```
pks-t: I'm surprised we never had a name clash with this `min` macro and any variable.

tests/odb/loose.c
```diff
@@ -55,6 +55,44 @@ static void test_read_object(object_data *data)
...
+	cl_git_pass(git_odb_open(&odb, "test-objects"));
+	cl_git_pass(git_oid_fromstr(&id, data->id));
+	cl_git_pass(git_odb_open_rstream(&stream, &tmp.len, &tmp.type, odb, &id));
+
+	remain = tmp.len;
+
+	while (remain) {
+		cl_assert((ret = git_odb_stream_read(stream, ptr, blocksize)) >= 0);
+		if (ret == 0)
+			break;
```
pks-t: Hum. Shouldn't we return EOF in case the stream is at its end?

src/zstream.c
```diff
@@ -87,9 +87,51 @@ size_t git_zstream_suggest_output_len(git_zstream *zstream)
 		return ZSTREAM_BUFFER_MIN_EXTRA;
```
pks-t: There's a typo in the commit message: `git_output` should be `get_output`.

src/zstream.c
```diff
@@ -87,9 +87,51 @@ size_t git_zstream_suggest_output_len(git_zstream *zstream)
...
+int git_zstream_get_output_chunk(
+	void *out, size_t *out_len, git_zstream *zstream)
+{
+	size_t in_queued, in_used, out_queued;
+
+	/* set up input data */
+	zstream->z.next_in = (Bytef *)zstream->in;
+	zstream->z.avail_in = (uInt)zstream->in_len;
+
+	if ((size_t)zstream->z.avail_in != zstream->in_len) {
```
pks-t: Is this handling the case where `z.avail_in = (uInt) zstream->in_len` overflows? If so, a small comment might be worthwhile.

src/odb_loose.c
```diff
@@ -192,67 +195,6 @@ static int parse_header(
...
-	inflateEnd(s);
-
-	if ((status != Z_STREAM_END) || (s->avail_in != 0)) {
-		giterr_set(GITERR_ZLIB, "failed to finish zlib inflation; stream aborted prematurely");
-		return -1;
-	}
-
-	return 0;
-}
-
```
pks-t: Yay! :)

src/odb_loose.c
```diff
@@ -131,7 +131,10 @@ static int parse_header_packlike(
 	}
 
 	out->size = size;
-	*out_len = used;
+
+	if (out_len)
+		*out_len = used;
+
```
pks-t: This change doesn't look related?

src/odb_loose.c
```diff
@@ -377,33 +413,23 @@ static int read_header_loose(git_rawobj *out, git_buf *loc)
...
-		(error = git_zstream_get_output_chunk(inflated, &inflated_len, &zs)) < 0 ||
-		(error = parse_header(&hdr, &header_len, inflated, inflated_len)) < 0)
-		goto done;
+	if (!is_zlib_compressed_data(obj))
+		error = read_header_loose_packlike(out, obj, (size_t)obj_len);
+	else
+		error = read_header_loose_standard(out, obj, (size_t)obj_len);
 
-	if (!git_object_typeisloose(hdr.type)) {
+	if (!error && !git_object_typeisloose(out->type)) {
```
pks-t: A goto in case of error would be a bit nicer here, I guess. Special-casing error handling just because it is the last statement only asks for small error later on when more code is being inserted after it.

src/odb_loose.c
```diff
@@ -352,7 +355,7 @@ static int read_loose(git_rawobj *out, git_buf *loc)
 	if ((error = git_futils_readbuffer(&obj, loc->ptr)) < 0)
 		goto done;
 
-	if (!is_zlib_compressed_data((unsigned char *)obj.ptr))
+	if (!is_zlib_compressed_data((unsigned char *)obj.ptr, obj.size))
 		error = read_loose_packlike(out, &obj);
 	else
 		error = read_loose_standard(out, &obj);
```
pks-t: Huh. Why should we treat an object as packlike though when `obj.size < 2`? Shouldn't we error out in that case instead? Something like:

```
if (is_zlib_compressed_data(&is_compressed, obj.ptr. obj.size) < 0)
    return -1;

if (!is_compressed)
    packlike();
else
    standard();
```

src/odb_loose.c
ethomson: I had the same thought, but I'm not sure that we can:  the stream is a public struct _but_ our hash context struct should be opaque, since it could actually be implemented by a variety of mechanisms.  I'd like to reevaluate this, though, since it would be nice to not malloc.  But I think that should be a separate PR.

src/odb_loose.c
```diff
@@ -896,11 +907,139 @@ static int loose_backend__writestream(git_odb_stream **stream_out, git_odb_backe
...
+	if (len) {
+		size_t chunk = MIN(len, INT_MAX);
+
+		if ((error = git_zstream_get_output(buffer, &chunk, &stream->zstream)) < 0)
+			return error;
+
+		total += chunk;
+	}
+
+	return total;
```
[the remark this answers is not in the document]
ethomson: We behave like `read(2)` by returning `0` at EOF.

src/odb_loose.c
```diff
@@ -896,11 +907,139 @@ static int loose_backend__writestream(git_odb_stream **stream_out, git_odb_backe
...
+	size_t decompressed, head_len;
+	obj_hdr hdr;
+	int error = 0;
+
+	assert(stream_out && len_out && type_out && _backend && oid);
+
+	backend = (loose_backend *)_backend;
+	*stream_out = NULL;
+	*len_out = 0;
+	*type_out = 0;
```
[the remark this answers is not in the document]
ethomson: Seems reasonable; fixed.

src/odb_loose.c
```diff
@@ -896,11 +907,139 @@ static int loose_backend__writestream(git_odb_stream **stream_out, git_odb_backe
...
+		(error = git_zstream_init(&stream->zstream, GIT_ZSTREAM_INFLATE)) < 0 ||
+		(error = git_zstream_set_input(&stream->zstream, stream->map.data, stream->map.len)) < 0)
+		goto done;
+
+	decompressed = sizeof(head);
+
+	/*
+	* inflate the initial part of the compressed buffer in order to parse the
+	* header; read the largest header possible, then store it in the `start`
+	* field of the stream object.
```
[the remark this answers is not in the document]
ethomson: I fixed this comment.

src/odb_loose.c
```diff
@@ -896,11 +907,139 @@ static int loose_backend__writestream(git_odb_stream **stream_out, git_odb_backe
...
+	*len_out = 0;
+	*type_out = 0;
+
+	stream = git__calloc(1, sizeof(loose_readstream));
+	GITERR_CHECK_ALLOC(stream);
+
+	hash_ctx = git__malloc(sizeof(git_hash_ctx));
+	GITERR_CHECK_ALLOC(hash_ctx);
+
+	if (locate_object(&object_path, backend, oid) < 0) {
```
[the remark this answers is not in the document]
ethomson: Fixed.

src/zstream.c
ethomson: Fixed.

src/zstream.c
ethomson: Updated to add a comment and an improved test.

tests/odb/loose.c
ethomson: `0` indicates EOF (we cannot block).

src/odb_loose.c
ethomson: [The general ODB layer handles this](https://github.com/libgit2/libgit2/blob/master/src/odb.c#L1376).  It cleans up the base object, the implementations do not.

src/odb_loose.c
ethomson: Corrected the missing `git__free(hash_ctx)`

ethomson:
I think that I've taken care of your issues, @pks-t

tests/odb/loose.c
```diff
@@ -103,6 +141,22 @@ void test_odb_loose__simple_reads(void)
 	test_read_object(&some);
 }
 
+void test_odb_loose__streaming_reads(void)
+{
+	size_t blocksizes[] = { 1, 2, 4, 16, 99, 1024, 123456789, 0 };
```
pks-t: Ugh, I was really confused by the `0` blocksize here. Didn't see that it was used as a terminator, so I kept on asking myself for several minutes how we can actually fully read through the stream in that case. Maybe remove that blocksize and iterate by array size instead of by sentinel?

tests/odb/largefiles.c
```diff
@@ -73,6 +77,48 @@ void test_odb_largefiles__streamwrite(void)
 	cl_assert_equal_oid(&expected, &oid);
 }
 
+void test_odb_largefiles__streamread(void)
```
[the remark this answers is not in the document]
ethomson: I don't think we've come to an agreement on this, no.  I'll rearrange these, but we should hash this out at the summit.  I'm definitely not the only person who prefers to add a failing test first, so the question is if we can get a workflow that accommodates all our preferences.

src/odb_loose.c
```diff
@@ -337,10 +348,10 @@ static int inflate_disk_obj(git_rawobj *out, git_buf *obj)
...
 	decompressed = sizeof(head);
 
-	/*
-	* inflate the initial part of the compressed buffer in order to parse the
-	* header; read the largest header possible, then push back the remainder.
-	*/
+	/* inflate the initial part of the compressed buffer in order to
+	 * parse the header; read the largest header possible, then push the
+	 * remainder into the body buffer.
+	 */
```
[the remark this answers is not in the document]
ethomson: lololol, i trust that your :trollface: ing with the "nicely".

tests/odb/loose.c
ethomson: Yeah, I think you're right; I fixed this.

ethomson:
Thanks again, updated with your comments.

tests/odb/largefiles.c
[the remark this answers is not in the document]
pks-t: I definitly agree with you. Adding failing tests first is the best way how we can demonstrate that the next commit fixes the issue. See my pull request #4500, which tackles this exact problem

src/odb_loose.c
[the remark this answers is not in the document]
pks-t: Hehe, I am ;)

pks-t:
Now if only GitHub had an interface to easily view the interdiff between two versions of the PR... don't really feel like going through all of it again right now :(

PR #4451, opened 2017-12-19
Implement message trailer parsing API

charliesome:
This pull request adds a `git_message_trailer` function that parses trailers out of commit messages. This function takes the message as a NUL-terminated C string and a callback. This callback is called once per trailer found. The callback can return non-zero at any time to stop trailer processing and return immediately.

The logic for trailer parsing is mostly taken directly from git core's `trailer.c`, so this will behave the same way as git around all the complexities of trailer parsing such as conflicts, patches, trailer continuations, etc.

I've also added a bunch of tests with all the edge cases I can think of.

cc @brianmario


ethomson:
@pks-t would your parser state machine be helpful here?

src/trailer.c
```diff
@@ -0,0 +1,415 @@
...
+
+static const char *const git_generated_prefixes[] = {
+	"Signed-off-by: ",
+	"(cherry picked from commit ",
+	NULL
+};
+
+static const char *const separators = ":";
+
+static inline int is_blank_line(const char *str)
```
ethomson: `inline` needs to be `GIT_INLINE`

src/trailer.c
```diff
@@ -0,0 +1,415 @@
...
+	i = len - 2;
+
+	for (; i >= 0; i--) {
+		if (buf[i] == '\n')
+			return i + 1;
+	}
+	return 0;
+}
+
+static bool starts_with(const char *str, const char *prefix)
```
ethomson: This looks like we could instead use our existing `git__prefixcmp`.

src/trailer.c
charliesome: I might just drop the `inline` entirely. It's a small static function so it'll probably end up being inlined automatically anyway.

brianmario:
I have the rugged branch going over at https://github.com/libgit2/rugged/compare/brianmario/trailers-api but it's failing because the cibuild script appears to be forcing the `libgit2` submodule at `master`. The tests pass locally, so once this is shipped I'll make sure things are all good to go on the rugged side and get a pull open there as well.

brianmario:
@ethomson Hi and happy new year! 👋  

How are you feeling about this now? Would love to get cranking on the rugged side so we can start using it 😄 

Pinging @carlosmn as well just in case one of you is back online before the other 😉

ethomson:
Hey @brianmario - I hadn't looked at it - I didn't realize y'all were ready for review, since it's not compiling on Windows.  Could you take a look at the errors?

brianmario:
The link to join the slack team isn't working for some reason, but I'm unsure what the error syntax is trying to communicate exactly.

> ..\..\tests\message\trailer.c(45): error C2059: syntax error : '{'

Would lead me to believe line 45 of `trailer.c` is where the issue is, but that and the lines around it are a multi-line code comment. Of which the `{` character is nowhere to be found. I imagine it's something to do with some "newer" (>C89 ?) C syntax that isn't compatible with older versions of VS (2013 in this case). But I'm not seeing it without help from these errors ;)

tiennou:
I think inline struct initialization is at least C90, and it's used in the test file.

pks-t:
@ethomson: I assume you mean the parser API instead of the state machine. And I think this patch would indeed benefit from it, as once again this here duplicates logic which is already provided by `git_parser`. So I would heavily encourage you (@brianmario) to use that one instead. If there are shortcomings in the existing API, we can surely extend it.

src/trailer.c
```diff
@@ -0,0 +1,415 @@
...
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+#include "common.h"
+#include "git2/message.h"
+
+#include <stddef.h>
+#include <string.h>
+#include <ctype.h>
+
+static const char comment_line_char = '#';
```
pks-t: I feel like this should probably be a define instead of a const variable.

src/trailer.c
```diff
@@ -0,0 +1,415 @@
...
+				continue;
+		} else if (isspace(bol[0]))
+			possible_continuation_lines++;
+		else {
+			non_trailer_lines++;
+			non_trailer_lines += possible_continuation_lines;
+			possible_continuation_lines = 0;
+		}
+continue_outer_loop:
+		;
```
pks-t: I'd prefer to have a continue here such that one can easily see that we're iterating the loop.

src/trailer.c
```diff
@@ -0,0 +1,415 @@
...
+			if (recognized_prefix &&
+			    trailer_lines * 3 >= non_trailer_lines)
+				return next_line(bol) - buf;
+			else if (trailer_lines && !non_trailer_lines)
+				return next_line(bol) - buf;
+			return len;
+		}
+		only_spaces = 0;
+
+		for (p = git_generated_prefixes; *p; p++) {
```
pks-t: Why do we treat git generated prefixes differently? Trailer lines can be whatever information the user wants them to be, so in my opinion there is no need to check for git prefixes but instead to check for the expected trailer line format.

src/trailer.c
```diff
@@ -0,0 +1,415 @@
...
+
+static char *find_trailer(const char *message, size_t* len)
+{
+	size_t patch_start = find_patch_start(message);
+	size_t trailer_end = find_trailer_end(message, patch_start);
+	size_t trailer_start = find_trailer_start(message, trailer_end);
+
+	size_t trailer_len = trailer_end - trailer_start;
+
+	char *buffer = git__malloc(trailer_len + 1);
```
pks-t: Not exactly sure, but is this even allowed in C90 wrt variable declaration after statement?

src/trailer.c
```diff
@@ -0,0 +1,403 @@
...
+	S_VALUE = 4,
+	S_VALUE_NL = 5,
+	S_VALUE_END = 6,
+	S_IGNORE = 7,
+};
+
+#define NEXT(st) { state = (st); ptr++; continue; }
+#define GOTO(st) { state = (st); continue; }
+
+int git_message_trailers(const char *message, git_message_trailer_cb cb, void *payload)
```
pks-t: The whole loop could use a state machine as we do for other parts (e.g. patch parsing, see patch_parse.c, parse_header_git())

tests/message/trailer.c
```diff
@@ -0,0 +1,179 @@
...
+		"\n"
+		"Signed-off-by: foo@bar.com\n"
+		"Signed-off-by: someone@else.com\n"
+	,
+		(struct trailer[]){
+			{ "Signed-off-by", "foo@bar.com" },
+			{ "Signed-off-by", "someone@else.com" },
+			{ NULL, NULL },
+		}
+	);
```
pks-t: I think this would be a bit more readable if you didn't declare the arrays as part of the macro but instead as variables first and then pass them into the assert_trailers. This would also get rid of the cast.

include/git2/message.h
```diff
@@ -38,6 +38,29 @@ GIT_BEGIN_DECL
...
+ * @param message The message to be parsed
+ *
+ * @param cb The callback to call for each trailer found in the message. The
+ *     key and value arguments are pointers to NUL-terminated C strings. These
+ *     pointers are only guaranteed to be valid until the callback returns.
+ *     User code must make a copy of this data should it need to be retained
+ *
+ * @param payload Pointer to callback data (optional)
+ *
+ * @return 0 on success, or non-zero callback return value.
```
pks-t: Small nit: we usually don't use empty lines between @-lines.

src/trailer.c
```diff
@@ -0,0 +1,403 @@
...
+
+#include <stddef.h>
+#include <string.h>
+#include <ctype.h>
+
+static const char comment_line_char = '#';
+
+static const char *const git_generated_prefixes[] = {
+	"Signed-off-by: ",
+	"(cherry picked from commit ",
```
pks-t: Is this really a trailer line? I thought the definition of a trailer line is that it is of the form "key: value"

src/trailer.c
```diff
@@ -0,0 +1,403 @@
...
+	NULL
+};
+
+static const char *const separators = ":";
```
pks-t: Same as above

tests/message/trailer.c
pks-t: Btw: this also causes MSVC 2010 builds to fail.

ethomson:
I'm luke warm on shipping it as-is since often we never _actually_ get around to iterating on things.  However I'm sympathetic to your process of including libgit2/rugged updates where you literally ship `master`.  (VSTS ships a fork so that we can merge things before they land here in `master`.)

So if you need this to unblock you on a feature, I'm sympathetic.  And we can take a quick review pass.  **BUT** I would want a firm guarantee that we will continue to refactor this, and promptly.

Does that seem reasonable?

src/trailer.c
brianmario: I think this is one of those git-core-specific intricacies I was talking about. I think technically a trailer can be in just about any configurable format (there are git-config settings we'll need to implement support for here, eventually). The HTTP-style `Key: value` stuff is just the default.

It looks like this stuff was added in git-core in https://github.com/git/git/commit/146245063e286d5a38b146bd5a38da958bd3957b - forgot there was also a heuristic used to determine what a trailer block is...

src/trailer.c
brianmario: My [comment below](https://github.com/libgit2/libgit2/pull/4451#discussion_r159573210) has some context around this but so you don't have to skip around (that comment is shown as outdated now for some reason) - https://github.com/git/git/commit/146245063e286d5a38b146bd5a38da958bd3957b has more context on this. tldr; a trailer block is determined using a heuristic of "regular" trailers and git-created trailers.

charliesome: This logic is taken directly from Git. It seems pretty funky to me, but the goal here is to be compatible with Git so I've brought over this git generated prefix logic as-is.

brianmario:
@ethomson yeah, I think that's fair 😃 

I've made some of the requested changes, and either wasn't sure about the others or am intentionally holding off (like rewriting the parser) for now.

src/trailer.c
charliesome: I believe so - we're still in variable declaration territory at this point since the only expressions preceding this are initializers for other declarations.

tests/message/trailer.c
```diff
@@ -0,0 +1,200 @@
...
+
+	cl_assert_equal_i(0, rc);
+}
+
+void test_message_trailer__simple(void)
+{
+	struct trailer trailers[3];
+
+	trailers[0].key   = "Signed-off-by";
+	trailers[0].value = "foo@bar.com";
```
charliesome: I think we can still use braced array initializers in ANSI C. Something like this should do the trick:

```c
struct trailer trailers[] = {
  { "Signed-off-by", "foo@bar.com" },
  /* ... */
  { NULL, NULL },
};
```

src/trailer.c
charliesome: Just to clarify, replace this lone semicolon with a `continue;`?

tests/message/trailer.c
pks-t: Yep :+1:

src/trailer.c
pks-t: Yeah, I definitly agree that we should strive to follow upstream-logic as close as possible. I really don't like all these workarounds and heuristics being put into so many places just because the initial design wasn't flexible enough. Nothing we can do about that... thanks though for the pointers!

src/trailer.c
pks-t: At least that's what I'd think is more readable. The loop is quite large, so you can easily forget that you're inside of one. Though now that I come to think of it the label already says 'continue_outer_loop', I didn't spot that one before. So this may also stay like it currently is

src/trailer.c
pks-t: I think I saw a compiler once treating intermingled function calls, assignments and newlines inside of the declaration block spitting an error. I bet it was some MSVC-style thing, but I cannot quite recall it.

src/trailer.c
brianmario: We may have not been very clear in the OP or throughout this discussion, but this isn't "workarounds and heuristics" of the initial design - it's [literally the upstream code itself](https://github.com/git/git/blob/master/trailer.c#L861). Which by my understanding is the source of truth for how any of this is supposed to work anyway, right?

The design of the code may not match the standards of libgit2, but if we (libgit2) are to handle trailers exactly the same way git does, we unfortunately need to also take the heuristics in to account whenever we refactor this code anyway.

For now at least, literally using the upstream code seems like the safest (at least in terms of compatibility) way to go.

brianmario:
Is there a way to force a rebuild on appveyor? The current failures were due to an external request failing.

src/trailer.c
pks-t: Sorry for not making myself clear, but with "initial design" I referred to upstream git.git, not your code. So I do not complain about your implementation but all the heuristics employed by git.git.

brianmario:
@ethomson how do you feel about this now? :)

carlosmn:
While the upstream code is a good place to start, our style is to provide iterators in order to allow different languages and environments.

If we merge this, we should be looking at making this into an iterator, optionally having the callback version built on top of that.

src/trailer.c
```diff
@@ -0,0 +1,402 @@
...
+}
+
+/* Return the position of the end of the trailers. */
+static int find_trailer_end(const char *buf, size_t len)
+{
+	return len - ignore_non_trailer(buf, len);
+}
+
+static char *find_trailer(const char *message, size_t* len)
+{
```
carlosmn: The name suggests that we're going to return the offset. Instead we're returning a new buffer and its length. Here we would name that something like `extract_trailer` which would have a `git_buf*` output parameter. `len` here similarly is an output parameter which here we put at the beginning of the parameter list.

tests/message/trailer.c
```diff
@@ -0,0 +1,179 @@
...
+{
+	struct trailer trailers[] = {
+		{"Key", "value"},
+		{NULL, NULL},
+	};
+
+	assert_trailers(
+		"Message\n"
+		"\n"
+		"Key   :   value\n"
```
carlosmn: Why would this be valid?

src/trailer.c
```diff
@@ -0,0 +1,402 @@
...
+ * is no patch in the message.
+ */
+static int find_patch_start(const char *str)
+{
+	const char *s;
+
+	for (s = str; *s; s = next_line(s)) {
+		if (git__prefixcmp(s, "---") == 0)
+			return s - str;
+	}
```
carlosmn: Should we be able to use `strstr(str, "\n---")` to find the end of the message?

carlosmn: I suppose that doesn't work if `---` is at the start of the message because we don't actually have one.

src/trailer.c
```diff
@@ -0,0 +1,402 @@
...
+	return 0;
+}
+
+/*
+ * If the given line is of the form
+ * "<token><optional whitespace><separator>..." or "<separator>...", return the
+ * location of the separator. Otherwise, return -1.  The optional whitespace
+ * is allowed there primarily to allow things like "Bug #43" where <token> is
+ * "Bug" and <separator> is "#".
```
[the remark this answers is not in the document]
brianmario: The the best of my knowledge, we haven't changed any of the logic from the git-core code. The main thing that it does that we don't (yet) is allow for a few git config options to control how trailers are parsed. Like what the trailer key/value separator is.

brianmario:
Would love one final pass on this. After talking to @carlosmn he convinced me of just using an array-based API as we're likely only working with one or maybe two trailers, for the small percentage of messages they even exist in. Plus we're already OK with having the entire commit message in memory anyway.

He and @charliesome already gave the latest change a look and were cool with things structurally. Based on the prior convo it sounds like this was/is ready to merge once CI is green. I'd love to do that ASAP now that we have a public API figured out that follows existing patterns and stuff.

brianmario:
Another spurious failure on AppVeyor and the travis builds are hung not building the mac builds but this is good to go. Going to merge this and we can iterate on further pulls if need-be.

pks-t:
I'm really surprised to see that being merged by you. It's not customary in libgit2 to merge PRs which are non-trivial by the person who is working on that himself, especially when you are kind of new to the project itself (in terms of having contributed to libgit2, not the libgit2 ecosystem in general). So going forward, I'd kindly ask you to not merge such PRs yourself.

ethomson:
Seconding @pks-t.  I slept on this, which mitigated by fury a little bit.  But I'm _incredibly_ tempted to revert this.  I'm scrolling up and see many comments that haven't been addressed here.

Fundamentally:  **this is not a GitHub project**.  This project is maintained by a community of overworked maintainers who largely do this in their spare time.  If your consumption of this library requires that your code have landed in master, then you either need to wait until we review it, **you fix the issues** or do what the rest of us do and ship a fork.

brianmario:
My sincere apologies on the misstep here. We can revert this if you'd like.

I was under the impression that the feedback that was left unaddressed were things we'd address in further refactorings after initially merging this. Once the API was stable and consistent with the standards in libgit2. I've been working with @carlosmn and @charliesome nearly every day on this.

Be that as it may, the code in question (by my understanding) was essentially copied as-is from git-core. Which we already know to be stable and in-use by the entire git-core ecosystem for years. So any changes being requested were for style and wouldn't affect the functionality of the API were they to happen after merging. @ethomson, your [comment here](https://github.com/libgit2/libgit2/pull/4451#issuecomment-355161727) lead me to believe moving forward with the _existing_ code was ok.

It was never my intention to add anything specific to GitHub here. Git trailer parsing is part of git-core (ala `git-interpret-trailers` and I had the impression that in a perfect world libgit2 would support nearly everything git itself does. So this didn't feel like a one-off API, rather the beginning of support for another git-core feature.

I've been a part of this project both in terms of design, its bindings and its use in production against many billions of requests nearly since the first line of code was written 7 or so years ago. I can assure you I mean the project or the community no harm or have any desire to make it specific to my or the company's needs. I misunderstood that we weren't done with this yet and I'm really sorry for that. I feel terrible :\

ethomson:
> I'm really sorry for that. I feel terrible.

Please don’t.  I didn’t mean to say that I thought you didn’t have the best interests of libgit2 at heart.  I know that you do, and it’s hard to balance competing interests like two open source projects and the one that you actually get paid to work on.

This was a miscommunication and I’m as upset with myself and my inattentiveness.  libgit2 is obviously not in a great place when it comes to resources, and trying to keep a high quality bar and do good code reviews is incredibly hard when you don’t have the time.  Sometimes I don’t realize how much time passes.

That’s no excuse though.  And I’m sorry to have taken my frustration out on you.

Thanks for continuing to contribute.  ❤️

PR #4453, opened 2017-12-29
winhttp: properly support ntlm and negotiate

ethomson:
When parsing unauthorized responses, properly parse headers looking for both NTLM and Negotiate challenges.  This was not previously necessary, since we did not previously need to set HTTP credentials.)

Set the HTTP credentials to default credentials (using a `NULL` username and password) with the schemes
supported by ourselves and the server.  It's not clear why this worked previously, but we now must explicitly call this API in order to authenticate with default credentials.

### DIFF
--- a/src/transports/winhttp.c
+++ b/src/transports/winhttp.c
@@ -172,14 +172,23 @@ static int apply_default_credentials(HINTERNET request, int mechanisms)
 	 * is "medium" which applies to the intranet and sounds like it would correspond
 	 * to Internet Explorer security zones, but in fact does not. */
 	DWORD data = WINHTTP_AUTOLOGON_SECURITY_LEVEL_LOW;
+	DWORD native_scheme = 0;
 
-	if ((mechanisms & GIT_WINHTTP_AUTH_NTLM) == 0 &&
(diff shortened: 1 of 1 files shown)


ethomson:
Regrettably, we do not have unit tests that exercise default credentials.  I've thought about this in the past, but it's crazy hard to have Active Directory set up for testing.  Maybe nightlies, but with our CI infrastructure this is basically a non-starter.

pks-t:
Huh. I always thought it would be nice if we had a separate CI system based on containers. Like that, we could easily spin up additional servers like LDAP, OpenSSH, proxies and other stuff. I know GitLab lets you do such stuff with their pipelines. Furthermore, that would also allow developers to easily spin up the same testing environment as it is inside of the CI system. We already had multiple occasions where we couldn't reproduce certain behaviour inside of the CI.

ethomson:
Yeah.  I think that it would be hard to deal with AD/Kerberos in the context of containers, but I would be happy to be proved wrong.

pks-t:
I was hoping that these features are also supported by slapd. Got no real idea though, only had the honor to set up testing LDAP systems twice.

PR #4455, opened 2017-12-30
refs: traverse symlinked directories

ethomson:
When enumerating the references in the loose refs storage area, we should traverse symlinked directories.  eg, if I create `refs/heads/foo` -> `/tmp/refs`, and create a loose ref `/tmp/refs/asdf`, then I should be able to resolve `refs/heads/foo/asdf`.

This is @doanac's change in #4388, with some minor cleanups for error checking and adding a test to easily validate this locally.  I'm opening a PR to put it through the paces on the CI machines.


doanac:
thanks for the Chistmas gift!

pks-t:
Cool 👍 I like the test in particular

PR #4459, opened 2018-01-02
Update callgit_remote_connect arguments

techvoltage:


ethomson:
Hi @techvoltage - I'm not sure of the rationale behind this change.  [`git_remote_connect` _does_ take 5 arguments](https://github.com/libgit2/libgit2/blob/master/include/git2/remote.h#L265)...?

ethomson:
There hasn't been any additional action here, and it appears that the current behavior is correct and that this PR does not build; closing.

Issue #1328, opened 2013-02-09
Textconv support


carlosmn:
Textconv is a way of transforming binary[1] data into something we can feed into our xdiff. Typical use-cases for this are letting `exif` show us an image's metadata, or transforming a PDF or Word file into plain text. Or even transforming an UTF-16/32 file into UTF-8.

We should allow the user to hook into our diffs in order to provide diff-able representations of data. Git does this via file extensions, which seems like a reasonable starting point. We can then go crazy with the rules as useful ones crop up.

The registered functions would get a buffer with the contents and give us the textual representation of the data.

/cc @arrbee @brianmario 

[1] binary means that our xdiff can't show a sensible diff with it, regardless of what the data inside is meant to represent.

softwareplumber:
Thought I might take a look at this but seeing as how this is such an old issue wondering if there is any prior art...

jklaiho:
This would be nice. GitLab, for example, is unable to show UTF-16 diffs, apparently due to this issue. See https://gitlab.com/gitlab-org/gitlab-ce/issues/24916#note_19417632

tiennou:
Ref: #4461

PR #4461, opened 2018-01-03
Textconv support (Issue #1328)

softwareplumber:
Here's an initial stab at providing textconv support.

Caveat: It's about 20 years since I last wrote any C code.

There is a feature switch (a diff option GIT_DIFF_ENABLE_TEXTCONV), and some limited unit tests (tests/textconv). Much code is copied from the existing implementation of filter. Some code common to filter and textconv as been moved from filter.c to a new file file filter_textconv.c to keep the dependency graph relatively sane. 

Next step is to do some testing downstream (I'll be building this into a fork of gitlab). Plan to add more unit tests based on issues identified as I work on that.

Figured the feature switch would make this a relatively benign experimental feature. My life would certainly be easier if this got merged into master early as any fixes to filter would be hard to merge.


ethomson:
Hi @softwareplumber - I'm happy to see a pull request.  I haven't had time to look through this yet, but some stylistic things jump out at me immediately:

- The indentation is not to our existing style.  We follow the [kernel style guide](https://www.kernel.org/doc/html/v4.10/process/coding-style.html) closely.  That means that we indent with tabs (not spaces), among other things.  We have an [editorconfig](http://editorconfig.org/) which should go a long way to describing this.  If you reformat your code with an editorconfig-aware editor, that would go a long way to fixing this on the first pass.  Then you should be able to eyeball your new code against our existing code.  Please take a pass at this, hopefully a mechanical substitution will fix things.  You've changed and added a lot of code and it would take a while for us to note each problem.
- The name `filter_textconv` doesn't really make sense for an object.  I think that this is a base class for both a `filter` and a `textconv`.  We should come up with a better name than `filter` / `textconv`.

I'm looking forward to jumping into this more deeply!  Thanks for the PR!

softwareplumber:
Sorry about the indentation. Newbie mistake. Hopefully fixed the worst formatting problems (also merged in changes from master). I'll see if I can find a linter that will tell me where else I am at variance with the kernel style guide.

How about 'transformer' as a name to replace filter_textconv?

softwareplumber:
Argh the git_diff_driver_lookup function signature has changed in some *very* recent update to master, which breaks my code.

softwareplumber:
@ethomson filter_textconv -> transformer ? waiting on response

include/git2/sys/textconv.h
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Shutdown callback on textconv
+ *
+ * Specified as `textconv.shutdown`, this is an optional callback invoked
+ * when the textconv is unregistered or when libgit2 is shutting down.  It
+ * will be called once at most and should release resources as needed.
+ * This may be called even if the `initialize` callback was not made.
+ *
+ * Typically this function will free the `git_textconv` object itself.
```
tiennou: "Typically" ? I take it means the callback is expected to free that object, right ? It might be clearer to make it a "must".

src/textconv.c
```diff
@@ -0,0 +1,611 @@
...
+    assert(name && textconv);
+    
+    if (git_rwlock_wrlock(&textconv_registry.lock) < 0) {
+        giterr_set(GITERR_OS, "failed to lock textconv registry");
+        return -1;
+    }
+    
+    if (!textconv_registry_find(NULL, name)) {
+        giterr_set(
+                   GITERR_TEXTCONV, "attempt to reregister existing textconv '%s'", name);
```
tiennou: Style: strange line split. Our limit is "soft", so if you're only off by a few characters from 80, don't bother.

src/textconv.c
```diff
@@ -0,0 +1,611 @@
...
+    error = git_attr_get(out, repo, 0, path, GIT_TEXTCONV_ATTR);
+    
+    /* if no values were found there is no textconv, return passthrough */
+    if (error == GIT_ENOTFOUND) {
+        return GIT_PASSTHROUGH;
+    }
+    return error;
+}
+
+int git_textconv_load(git_textconv** out, git_repository* repo, const char* path) {
```
tiennou: Style: please use `type *var`.

src/textconv.c
```diff
@@ -0,0 +1,611 @@
...
+                                        git_repository *repo,
+                                        const char *path)
+{
+    int error;
+    
+    error = git_attr_get(out, repo, 0, path, GIT_TEXTCONV_ATTR);
+    
+    /* if no values were found there is no textconv, return passthrough */
+    if (error == GIT_ENOTFOUND) {
+        return GIT_PASSTHROUGH;
```
tiennou: `GIT_PASSTHROUGH` is supposed to be a special return code from our callbacks only (meaning "there was a callback, but it want to be ignored"), so that would be somewhat unexpected externally. You might want to keep that `GIT_ENOTFOUND` instead.

attributes.c
```diff
@@ -1,8 +0,0 @@
-//
```
tiennou: Created/deleted file. You might want to fixup! that.

src/diff_driver.h
```diff
@@ -8,7 +8,6 @@
 #define INCLUDE_diff_driver_h__
 
 #include "common.h"
-
```
tiennou: Nitpick: stray line deletion ("common.h" is special, IMHO it's best if it stands out).

src/diff_file.c
```diff
@@ -260,6 +260,7 @@ static int diff_file_content_load_blob(
     
     if (!error && (opts->flags & GIT_DIFF_ENABLE_TEXTCONV) != 0) {
         error = git_textconv_load(&tc, fc->driver);
+        if (error == GIT_PASSTHROUGH) error = 0;
```
tiennou: … almost externally visible `GIT_PASSTHROUGH`.

textconv.h
```diff
@@ -1,12 +0,0 @@
-//
```
tiennou: Creation/deletion of file. Fixup ?

include/git2/sys/textconv.h
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Look up a textconv by name
+ *
+ * @param name The name of the textconv
+ * @return Pointer to the textconv object or NULL if not found
+ */
+GIT_EXTERN(git_textconv *) git_textconv_lookup(const char *name);
+
+#define GIT_TEXTCONV_HTTP  "http"
```
tiennou: Maybe I missed it, but I don't think there's actually a default textconv in there ?

tiennou:
About that `filter_textconv` conundrum : I see you started with a slight duplicate of the filter API (which I have no prior experience with). My personal preference would be to keep those in separate "namespaces", like you had at the start, instead of munging them both in `filter_textconv`, which makes the API somewhat weird (having two main parameters that are optional).

Some alternative implementation ideas I can see : 
- shoehorning textconvs as a special kind of `git_filter`, so you can share the `from_data/blob/stream` functions.
- abstract those `from_*` function internally, so those can take either a `git_filter_list` or a `git_textconv` (unless you can both textconv *and* filter at the same time, as I said "no experience").

Disclaimer : those are just suggestions, so please don't take this as a request to change things. I'll (humbly) defer to @ethomson for the final word.

Issue #4460, opened 2018-01-03
High number of `stat` calls for `gitattributes` files


stanhu:
I've noticed that libgit2 makes a high number of `stat` calls to `gitattributes` files, which significantly impacts networked file systems.

### Reproduction steps

With the Linux repo (https://github.com/torvalds/linux.git), I created this Rugged script:

```ruby
require 'rugged'

repo = Rugged::Repository.new('.')

commit_object = repo.rev_parse('master')
original_oid = repo.rev_parse('979e1f08042b83152dfe3d76df10db31eb7edf98')
diff_commits = commit_object.parents[0].diff(original_oid)

puts diff_commits.stat
```

The result is 30,497 files changed with 753,086 additions and 1,345,193 deletions. Now if I run `strace` on this, I see there are 457,782 hits with `gitattributes`:

```sh
$ strace -tt -T -o /tmp/diff-linux.txt -s 1024 ruby diff.rb                                                                                                     
30497
753086
1345193
$ grep gitattributes /tmp/diff-linux.txt | wc -l
457782
```

Looking at it in more detail, I see that we are often duplicating the same lookup:

```sh
07:38:38.120824 stat("/home/stanhu/linux/Documentation/.gitattributes", 0x7ffe261fe130) = -1 ENOENT (No such file or directory) <0.000009>
07:38:38.120852 stat("/home/stanhu/linux/.gitattributes", {st_mode=S_IFREG|0664, st_size=30, ...}) = 0 <0.000010>
07:38:38.120881 access("/etc/gitattributes", F_OK) = -1 ENOENT (No such file or directory) <0.000011>
07:38:38.120949 access("/etc/gitattributes", F_OK) = -1 ENOENT (No such file or directory) <0.000010>
07:38:38.121019 stat("/home/stanhu/linux/.gitattributes", {st_mode=S_IFREG|0664, st_size=30, ...}) = 0 <0.000010>
07:38:38.121270 stat("/home/stanhu/linux/Documentation/ABI/testing/.gitattributes", 0x7ffe261fe130) = -1 ENOENT (No such file or directory) <0.000010>
07:38:38.121301 stat("/home/stanhu/linux/Documentation/ABI/.gitattributes", 0x7ffe261fe130) = -1 ENOENT (No such file or directory) <0.000010>
07:38:38.121329 stat("/home/stanhu/linux/Documentation/.gitattributes", 0x7ffe261fe130) = -1 ENOENT (No such file or directory) <0.000010>
[...]

PR #4462, opened 2018-01-03
diff_generate: avoid excessive stats of .gitattribute files

pks-t:
When generating a diff between two trees, for each file that is to be
diffed we have to determine whether it shall be treated as text or as
binary files. While git has heuristics to determine which kind of diff
to generate, users can also that default behaviour by setting or
unsetting the 'diff' attribute for specific files.

Because of that, we have to query gitattributes in order to determine
how to diff the current files. Instead of hitting the '.gitattributes'
file every time we need to query an attribute, which can get expensive
especially on networked file systems, we try to cache them instead. This
works perfectly fine for every '.gitattributes' file that is found, but
we hit cache invalidation problems when we determine that an attribuse
file is _not_ existing. We do create an entry in the cache for missing
'.gitattributes' files, but as soon as we hit that file again we
invalidate it and stat it again to see if it has now appeared.

In the case of diffing large trees with each other, this behaviour is
very suboptimal. For each pair of files that is to be diffed, we will
repeatedly query every directory component leading towards their
respective location for an attributes file. This leads to thousands or
even hundreds of thousands of wasted syscalls.

The attributes cache already has a mechanism to help in that scenario in
form of the `git_attr_session`. As long as the same attributes session
is still active, we will not try to re-query the gitmodules files at all
but simply retain our currently cached results. To fix our problem, we
can create a session at the top-most level, which is the initialization
of the `git_diff` structure, and use it in order to look up the correct
diff driver. As the `git_diff` structure is used to generate patches for
multiple files at once, this neatly solves our problem by retaining the
session until patches for all files have been generated.

The fix has been tested with linux.git by calling
`git_diff_tree_to_tree` and `git_diff_to_buf` with v4.10^{tree} and
v4.14^{tree}.

                | time    | .gitattributes stats
    without fix | 33.201s | 844614
    with fix    | 30.327s | 4441

While execution only improved by roughly 10%, the stat(3) syscalls for
.gitattributes files decreased by 99.5%. The benchmarks were quite
simple with best-of-three timings on Linux ext4 systems. One can assume
that for network based file systems the performance gain will be a lot
larger due to a much higher latency.

### DIFF
--- a/src/diff.h
+++ b/src/diff.h
@@ -34,6 +34,7 @@ typedef enum {
 struct git_diff {
 	git_refcount     rc;
 	git_repository   *repo;
+	git_attr_session attrsession;
 	git_diff_origin_t type;
 	git_diff_options opts;
 	git_vector       deltas;    /* vector of git_diff_delta */
--- a/src/diff_driver.c
+++ b/src/diff_driver.c
@@ -354,27 +354,30 @@ done:
 }
 
 int git_diff_driver_lookup(
-	git_diff_driver **out, git_repository *repo, const char *path)
+	git_diff_driver **out, git_repository *repo,
+	git_attr_session *attrsession, const char *path)
 {
 	int error = 0;
-	const char *value;
+	const char *values[1], *attrs[] = { "diff" };
 
 	assert(out);
 	*out = NULL;
 
 	if (!repo || !path || !strlen(path))
 		/* just use the auto value */;
-	else if ((error = git_attr_get(&value, repo, 0, path, "diff")) < 0)
+	else if ((error = git_attr_get_many_with_session(values, repo,
+			attrsession, 0, path, 1, attrs)) < 0)
 		/* return error below */;
-	else if (GIT_ATTR_UNSPECIFIED(value))
+
+	else if (GIT_ATTR_UNSPECIFIED(values[0]))
 		/* just use the auto value */;
-	else if (GIT_ATTR_FALSE(value))
+	else if (GIT_ATTR_FALSE(values[0]))
 		*out = &global_drivers[DIFF_DRIVER_BINARY];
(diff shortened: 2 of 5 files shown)


pks-t:
This should fix #4460

stanhu:
@pks-t Thanks so much for the quick turnaround! Will give this a test.

ethomson:
Yep, this is exactly why the `git_attr_session` exists.  Thanks, @pks-t!  😀

stanhu:
@pks-t @ethomson Thanks! What are the plans for releasing a new version of libgit2?

pks-t:
@stanhu We don't have a schedule right now, but we hope to get it out soonish. If you ask me, I'd aim for a release in around three weeks, but please don't take that for granted as we'll have to discuss that first.

softwareplumber:
Any chance of a unit test showing how to initialize a git_attribute_session and use it?

pks-t:
`git_attr_session` is not part of our public API, but it is an internal implementation detail only. So if you intend to use it for your own purpose, then I'm afraid this will not be possible. So I think there's no need for an example of it.

Just in case you're still interested:

```
char *values[1], *attributes[] = { "text" };
char *paths[] = { "path/to/file.txt", "foobar.txt", "bar.txt" };
git_attr_session session;
int i;

git_attr_session__init(&session);
for (i = 0; i < ARRAY_SIZE(paths); i++)
    git_attr_get_many_with_session(values, repo, &session, 0, paths[i], ARRAY_SIZE(attributes), attributes);
git_attr_session__free(&session);
```

softwareplumber:
Thanks! Yep, it works as I thought it did... Was writing a unit test for a PR, had an issue which made me doubt it. But turned out to be Xcode build problem.

PR #4463, opened 2018-01-04
Now vcpkg support build and install libgit2

fcharlie:
https://github.com/Microsoft/vcpkg/tree/master/ports/libgit2

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,14 @@ On most systems you can build the library using the following commands
 
 Alternatively you can point the CMake GUI tool to the CMakeLists.txt file and generate platform specific build project or IDE workspace.
 
+On Windows You can use vcpkg build libgit2:
+
+```
+vcpkg install libgit2:x64-windows libgit2:x86-windows
+```
+
+libgit2 port: [https://github.com/Microsoft/vcpkg/tree/master/ports/libgit2](https://github.com/Microsoft/vcpkg/tree/master/ports/libgit2)
+
 Running Tests
 -------------
 


fcharlie:
@ethomson This PR is update readme about vcpkg install libgit2 (port by me). Please merge, Thanks.

ethomson:
What are the advantages to using `vcpkg`?  I dont like special casing the windows build scenario unless it has obvious advantages.  Can I run the tests after using it?  If not the rest of the guide doesn’t apply.

I’m okay with pointing people to this mechanism but it seems more akin to a package manager and doesn’t fit in this part of the guide.  This seems like “libgit2 acquisition for normal people” not “how to build it and run the tests”...

fcharlie:
@ethomson This update simply tells the user that there is an easy way to use libgit2 on Windows instead of configuring the compilation Libgit2 yourself. That's all. It's actually the same as using apt install libgit2 in Ubuntu.

ethomson:
I’m glad that you added this support - I think it’s useful and I agree that we should document it in the README, but this isn’t the right place for that.  It’s in a section entitlted “Building libgit2 - Using CMake” which is not accurate.

I would instead change the second section, “Quick Start” to include information about how people can install *without* building themselves.  This is probably the most common getting started experience after all.

Something like:

> ## Quick Start
> 
> ### Windows
> 
> Include your instructions here...
> 
> ### Building from source
>
> ...move current section content here...

Better still would be if you were able to include some content about how to get started on Debian based systems (`apt-get install...`), Mac with homebrew, etc...

You may also wish to include a warning that the packages may be outdated.  Some providers are in LTS, etc.

fcharlie:
@ethomson OK, I will close this PR, and I spend some time sorting out how to use libgit2 quickly.

PR #4467, opened 2018-01-07
cmake: move the rule to find static archives close to building clar

carlosmn:
If we're building static libraries, we want to use that for building our clar
binary. This is done in 49551254 (2017-09-22; cmake: use static dependencies
when building static libgit2) but that commit included the rule too early,
making it affect the search for iconv, meaning we did not find it when we were
building a static libgit2.

Move the rule to just before building clar, after we've included the rules for
building the library itself. This lets us find and link to the dynamic libiconv.


carlosmn:
This also fixes the issue I was having with openssl where we wouldn't actually link to libssl even though we're using it.

pks-t:
Doesn't this mean that we will not use static archives when searching for dependencies? When I want to build a static libgit2, I want it to ideally have only static dependencies, as well. With this change, I think we will find shared library dependencies again. This is somewhat contradictory to what I said in 4955125, but well.

carlosmn:
Before this change, we weren't building our archive with static dependencies. We were still relying on whatever linked with `libgit2.a` to link against our dependencies. `libgit2_clar` is linked against iconv, openssl et al. And the build errors in rugged indicate that we were definitely not statically linking the dependencies when we were building the static archive.

Though on a higher-level note, there's two different reasons to build a static libgit2: you might want to include our code into your binary, or you might want to build _everything_ statically. For the latter, it seems like you would be better off passing `-static` to the linker building the final result. Otherwise different dependencies might end up including common deps mutliple times.

Issue #4466, opened 2018-01-07
CMake does not find iconv/put libssl in the pc file when building a static library


carlosmn:
### Reproduction steps

Compare the results of running (from a `build/` subdir)

    $ cmake .. -DBUILD_SHARED_LIBS=ON

 and

    $ cmake .. -DBUILD_SHARED_LIBS=OFF

### Expected behavior

The iconv library should be found both times.

The instructions to link with libssl should be written out to `libgit2.pc`.

### Actual behavior

When disabling shared libs, we do not find the iconv library, with nothing else changing.

The pc file does not say to link to libssl/libcrypto

### Version of libgit2 (release number or SHA1)

It happens in 70db57d4a431dae2edcf19e92a84db8d6cf4f935 but not in c7c5f2c4ecc727ac555f51bd20353f7f3ae9c07a. I have not narrowed it down further yet.

### Operating system(s) tested

macOS Sierra

/cc @pks-t for CMake shennanigans

carlosmn:
Bisecting indicates 49551254c3146274dfb5764318f4661ce47a6f0a is the bad commit. It looks like it went too far and made us look for a static libiconv, which is not what we want to do.

carlosmn:
Just opened #4467 which should hopefully deal with that issue.

carlosmn:
That got fixed, but apparently the SSL issue wasn't actually resolved.

carlosmn:
On the OpenSSL side, it looks like the issue comes from defaulting to the collission-detection code. Now that we don't rely on OpenSSL for SHA-1, we end up not adding it to `libgit2.pc` it even though we depend on it for HTTPS.

We do link clar to it, so it looks like we do know we have to link to it.

carlosmn:
The build in libgit2/rugged#733 is now green when updating to current libgit2 master.

Thanks for tuning in to this week's episode of "Carlos talks to himself for a bit and then fixes a bug".

PR #4468, opened 2018-01-08
Make sure to include 'openssl' as a dep when building statically with SHA1DC

carlosmn:
We need to add `openssl` to the list of private dependencies regardless of whether we want it for SHA-1 or TLS.

During #4398 the rules got changed so we only add it when we want it as the SHA-1 implementation, rather than if we searched for OpenSSL and found it. This means that choosing SHA1DC (or using a newer version where this is the default) would no longer indicate to those including us as a static archive that they need to link against the openssl package.

This takes care of the second part of #4466 and we should again be able to include the library in bindings.


src/CMakeLists.txt
```diff
@@ -439,6 +439,9 @@ IF (SONAME)
 		SET_TARGET_PROPERTIES(git2 PROPERTIES PREFIX "${LIBGIT2_PREFIX}")
 	ENDIF()
 ENDIF()
+
+LIST(REMOVE_DUPLICATES LIBGIT2_PC_REQUIRES)
+STRING(REPLACE ";" " " LIBGIT2_PC_REQUIRES "${LIBGIT2_PC_REQUIRES}")
```
pks-t: Much nicer to treat it as a list, thanks!

src/CMakeLists.txt
```diff
@@ -184,6 +184,7 @@ IF (USE_HTTPS)
 		LIST(APPEND LIBGIT2_INCLUDES ${OPENSSL_INCLUDE_DIR})
 		LIST(APPEND LIBGIT2_LIBS ${OPENSSL_LIBRARIES})
 		LIST(APPEND LIBGIT2_PC_LIBS ${OPENSSL_LDFLAGS})
+		LIST(APPEND LIBGIT2_PC_REQUIRES "openssl")
```
pks-t: Looks obviously good, thanks for this fix!

PR #4470, opened 2018-01-09
Fix error handling of odb_object__alloc() in odb_read_1()

bisho:
If `odb_object__alloc()` returns `NULL` (fails to allocate) it currently goes to `out`. If error is != 0 we will try to free raw.data, when raw couldn't be allocated. If error is 0, then we are silently swallowing the allocation error, and possibly causing access to a wrong `out` pointer. Let's return an error instead.

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -1040,10 +1040,8 @@ static int odb_read_1(git_odb_object **out, git_odb *db, const git_oid *id,
 	}
 
 	giterr_clear();
-	if ((object = odb_object__alloc(id, &raw)) == NULL) {
-		giterr_set(GITERR_ODB, "object could not be allocated");
-		return -1;
-	}
+	object = odb_object__alloc(id, &raw);
+	GITERR_CHECK_ALLOC(object);
 
 	*out = git_cache_store_raw(odb_cache(db), object);
 
--- a/src/odb.c
+++ b/src/odb.c
@@ -1040,8 +1040,10 @@ static int odb_read_1(git_odb_object **out, git_odb *db, const git_oid *id,
 	}
 
 	giterr_clear();
-	if ((object = odb_object__alloc(id, &raw)) == NULL)
-		goto out;
+	if ((object = odb_object__alloc(id, &raw)) == NULL) {
+		giterr_set(GITERR_ODB, "object could not be allocated");
+		return -1;
+	}
 
 	*out = git_cache_store_raw(odb_cache(db), object);
 


carlosmn:
Thanks. We have a macro for this kind of thing, `GITERR_CHECK_ALLOC` which post-dates this code so it's not applied universally, but it would bring more consistency into the code.

We should also avoid having specific messages for everywhere we can OOM. When we fail to allocate, we already set the error to mark the OOM so we can just return with ` -1` (which is what the macro does). This also avoids trying to allocate more memory after the system has told us we're out of memroy.

bisho:
Perfect! I'll update the patch.

ethomson:
I don't agree with your analysis:  if `odb_object__alloc` returns `NULL` then we should still free `raw.data`.  `odb_object__alloc` isn't the function responsible for creating it, it will be allocated by the [`read` function](https://github.com/libgit2/libgit2/pull/4470/files#diff-f6de38fbae0fb818ac2fe81e431f749dL1018).

If `odb_object__alloc` fails then we _should_ be `free`ing `raw.data`.  It could contain a remarkably large blob and we're actually OOM because we have allocated that blob.  

Generally speaking, we do not bother trying to clean up allocation failures and just allow ourselves to leak on the assumption that there's nothing we can do.  The exception is when we've just allocated potentially multi-GB buffers.  

This is one of those cases.

Looking at this function, either `raw.data` will never be populated, because [`read` returned an error] or the object was not found in any of the backends.  If some `read` returned an error, then [we return from the function](https://github.com/libgit2/libgit2/pull/4470/files#diff-f6de38fbae0fb818ac2fe81e431f749dR1018).  If [then we return early](https://github.com/libgit2/libgit2/pull/4470/files#diff-f6de38fbae0fb818ac2fe81e431f749dR1029).

It appears to me that there is no case at which we would reach [`odb_object__alloc`](https://github.com/libgit2/libgit2/pull/4470/files#diff-f6de38fbae0fb818ac2fe81e431f749dL1043) and not have a `raw.data` allocated.

There _is_ a problem with this function, though, which is that failures from `odb_object__alloc` are never detected and returned to the caller.  So I think that the proper solution is:

```
if ((object = odb_object__alloc(id, &raw)) == NULL) {
    error = -1;
    goto out;
}
```

This will ensure that if `odb_object__alloc` fails that we properly free `raw.data` and propagate the error code to the caller.  (Or some variant on that.)

There's also no need for the `giterr_clear` there.

While looking at this, there are some other issues:

* `odb_read_hardcoded` should _also_ do some error checking on allocations.  At the moment, it does none.  Better still, there's no need for it to do any allocation at all, but this may not be worth the refactoring.

bisho:
Oh, I see, I missed that, will prepare a more comprehensive fix.

ethomson:
I opened #4509 with the change to return an error when we cannot allocate an odb object.  Closing this - but if you decide to work on a more comprehensive fix for `odb_read_hardcoded` so that it doesn't do any allocations, that would be 👍

PR #4471, opened 2018-01-10
cmake: use a FEATURE_SUMMARY call compatible with 3.0.2

carlosmn:
When we print features, we make an effort to support all the way back to
pre-3.0. However, in the code for versions from 3 onward we call
`FEATURE_SUMMARY` with multiple kinds of elements to print in the same line.
This is only supported in CMake 3.1 and later, making the rather popular CMake
3.0.2 unable to build the library.

Use a single kind of element per invocation. This means we need to provide a
"description" text, which CMake provides for us if provide multiple kinds of
elements.


pks-t:
Looks obviously good, thanks!

PR #4472, opened 2018-01-10
travis: fetch trusty dependencies from Bintray

ethomson:
I would like to host the trusty dependencies in an Azure Web App - currently I'm hosting them on a linux VM in Azure.  This is a lower-cost option that hurts my pocketbook less.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ addons:
  apt:
   sources:
    - sourceline: 'deb https://dl.bintray.com/libgit2/ci-dependencies trusty libgit2deps'
-     key_url: 'https://www.edwardthomson.com/keys/ethomson@libgit2.org'
+     key_url: 'https://bintray.com/user/downloadSubjectPublicKey?username=bintray'
   packages:
    cmake
    curl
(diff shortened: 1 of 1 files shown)


ethomson:
Ugh.  There's something about Azure Web Apps where they include a `Content-Type` header for files with no extensions, and apt (at least, trusty's version of apt) decides that's not cool.  (Sigh.)  I have yet to find a way to disable the `Content-Type` header.  I'll keep poking at this.

ethomson:
Okay, well, I've given up on Azure.  It seems that GitHub Pages can handle this properly!

ethomson:
Okay, let's try Bintray!  😀

carlosmn:
Your GPG key won't be very useful as I'm having bintray sign it themselves (we could have a specific bintray gpg key, and upload via API so we can provide the passphrase but this seems easier).

ethomson:
Can confirm this setup is 👍 on my trusty box.  Hopefully travis agrees.

carlosmn:
The only missing bits for green CI is the macOS build which isn't affected by this, so I'm just going to merge it.

pks-t:
Adding the backport label as we need this to fix CI builds on maint

PR #4474, opened 2018-01-12
Special-casing null OIDs

pks-t:
I've been triggered by the not-so-recent discussion in upstream git.git around null OIDs (see [1]) to implement some defensive mechanisms about ever reading or writing null OIDs either via the ODB layer or into trees. For the ODB part, I now reject writing and shorcut reading null OIDs while for trees I've only implemented the rejection on inserting entries into the treebuilder. So I've not changed the reading part for our trees.

[1]: https://marc.info/?l=git&m=151120958313567&w=2

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -53,6 +53,7 @@ static git_cache *odb_cache(git_odb *odb)
 
 static int odb_otype_fast(git_otype *type_p, git_odb *db, const git_oid *id);
 static int load_alternates(git_odb *odb, const char *objects_dir, int alternate_depth);
+static int error_null_oid(int error, const char *message);
 
 static git_otype odb_hardcoded_type(const git_oid *id)
 {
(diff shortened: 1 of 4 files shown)


pks-t:
One more note: One thing I've been a bit hesitant is about `git_oid_is_null`. It sounds a bit like checking for `NULL` pointers, which it obviously shouldn't do. While this is how upstream calls those OIDs, we might consider instead calling the function `git_oid_is_zero`.

pks-t:
Ugh. We already have `git_oid_iszero`... will use that later on

src/odb.h
```diff
@@ -112,6 +112,12 @@ int git_odb__error_mismatch(
 int git_odb__error_notfound(
 	const char *message, const git_oid *oid, size_t oid_len);
 
+/*
+ * Generate an error for null OID and return the specified error
+ * code.
+ */
+int git_odb__error_null_oid(int error, const char *message);
```
ethomson: Do we need to have this in the header?  Why not just make `git_odb__error_null_oid` static in `odb.c`?

pks-t: No, definitly not. I think I just blindly followed style of the other error functions without thinking

PR #4475, opened 2018-01-12
v0.26.3 backports

pks-t:
Proposed backports for v0.26.1. This includes all PRs that I've currently marked with the "backport" label:

- #4045
- #4455
- #4447
- #4444
- #4437
- #4435
- #4432
- #4431
- #4427
- #4426
- #4424
- #4423
- #4408
- #4386
- #4369
- #4368
- #4367
- #4364
- #4359
- #4357
- #4344
- #4326
- #4306
- #4296
- #4280
- #4273

I've initially included #4277 as well, but that code was only merged after v0.26.0 in #4202.


pks-t:
Ugh. AppVeyor and Coverity are hung now because I erroneously based this on "master" first.

carlosmn:
We should not include #4408 as it's a breaking change. The size of `git_time` will be different depending on which version of the headers you get.

#4455 and #4357 I'm unsure about as they introduce new behaviour.

tiennou:
Can #4275 be considered ? Also, I don't remember whether 0.26 branched before or after the CMake changes, but maybe #4439 too ?

pks-t:
@tiennou: it branched after the big CMake changes, so #4439 is not needed. #4275 looks good to me, though

@carlosmn: true, didn't think of ABI compatibility here, so #4408 should definitly not be part. I'm kind of indifferent to #4357 and #4455, but I see your point, so I'll remove them as well.

pks-t:
Updated according to my last comment.

pks-t:
Added #4472

pks-t:
Included 54214d618 (cmake: fix linker error with dbghelper library, 2017-09-15) to unbreak AppVeyor builds.

pks-t:
Updated to include #4535, #4497, #4539, #4491

pks-t:
Added #4550, #4545, #4555, #4553

pks-t:
Dropped #4424 as that actually changes the API in a rather subtle way and included the upgrade of zlib to 1.2.11

pks-t:
Rebased on v0.26.1 and made this the v0.26.2 release

pks-t:
Rebased on v0.26.2, bumped to v0.26.3 and added a changelog. Will finally release this either tomorrow or on Monday when the CI is green.

PR #4476, opened 2018-01-12
scripts: add backporting script

pks-t:
This adds a simple script for backporting pull requests to older
branches. It accepts as parameters a list of pull request numbers which
whose commits are to be cherry-picked. The identification of PRs
currently happens by using the commit message of the merge of the PR,
which should conform to the message "Merge pull request #<PR>".

While the heuristic works in practice, we could instead also use the
direct references from GitHub via "pull/#<PR>/head". This requires the
user to have all these references fetched, though, so we can just use
the current heuristic until we experience any issues with that.

---

Dunno if we want this. I've used this script to create #4475, and I think a bit of automation is nice to have.

### DIFF
--- a/script/backport.sh
+++ b/script/backport.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+if test $# -eq 0
+then
+    echo "USAGE: $0 <#PR> [<#PR>...]"
+    exit
+fi
+
+commits=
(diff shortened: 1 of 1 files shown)


ethomson:
If you found this useful, then let's include it.

PR #4478, opened 2018-01-12
refs: include " sorted " in our packed-refs header

carlosmn:
This lets git know that we do in fact have written our packed-refs file
sorted (which is apparently not necessarily the case) and it can then use the
new-ish mmaped access which lets it avoid significant amounts of effort parsing
potentially large files to get to a single piece of data.

### DIFF
--- a/src/refs.h
+++ b/src/refs.h
@@ -30,7 +30,7 @@ extern bool git_reference__enable_symbolic_ref_target_validation;
 
 #define GIT_SYMREF "ref: "
 #define GIT_PACKEDREFS_FILE "packed-refs"
-#define GIT_PACKEDREFS_HEADER "# pack-refs with: peeled fully-peeled "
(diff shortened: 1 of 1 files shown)


mhagger:
Thanks.

Some background: The git command line client now mmaps the `packed-refs` file, doing a binary search through the file to find the reference(s) that it needs, and only parsing those.

If the `packed-refs` file doesn't confirm that it is "sorted" with the new keyword, git reads the whole file and checks whether it is, in fact, sorted. This costs extra I/O and some lightweight parsing. If the entries are sorted, it continues to do the mmap thing. If not, it reads the file into memory and sorts it there, which requires extra RAM.

Though, on Windows, since you can't delete an open file, it *always* reads the file into RAM then closes the `packed-refs` file. So on Windows the "sorted" keyword mostly saves it the trouble of lightly parsing the whole file and checking the ordering.

ethomson:
Thanks for the additional background, @mhagger.

PR #4484, opened 2018-01-19
branch: refuse creating branches named 'HEAD'

pks-t:
Since a625b092c (branch: correctly reject refs/heads/{-dash,HEAD},
2017-11-14), which is included in v2.16.0, upstream git refuses to
create branches which are named HEAD to avoid ambiguity with the
symbolic HEAD reference. Adjust our own code to match that behaviour and
reject creating branches names HEAD.

### DIFF
--- a/src/branch.c
+++ b/src/branch.c
@@ -70,6 +70,12 @@ static int create_branch(
 	assert(branch_name && commit && ref_out);
 	assert(git_object_owner((const git_object *)commit) == repository);
 
+	if (!git__strcmp(branch_name, "HEAD")) {
+		giterr_set(GITERR_REFERENCE, "'HEAD' is not a valid branch name");
+		error = -1;
+		goto cleanup;
+	}
+
 	if (force && !bare && git_branch_lookup(&branch, repository, branch_name, GIT_BRANCH_LOCAL) == 0) {
 		error = git_branch_is_head(branch);
 		git_reference_free(branch);


pks-t:
That same thought occurred to me, and yes, Git is being case sensitive here. See [a625b092](https://github.com/git/git/commit/a625b092cc59940521789fe8a3ff69c8d6b14eb2), or more specifically the following snippet:

```
 int strbuf_check_branch_ref(struct strbuf *sb, const char *name)
 {
 	strbuf_branchname(sb, name, INTERPRET_BRANCH_LOCAL);
-	if (name[0] == '-')
-		return -1;
+
+	/*
+	 * This splice must be done even if we end up rejecting the
+	 * name; builtin/branch.c::copy_or_rename_branch() still wants
+	 * to see what the name expanded to so that "branch -m" can be
+	 * used as a tool to correct earlier mistakes.
+	 */
 	strbuf_splice(sb, 0, 0, "refs/heads/", 11);
+
+	if (*name == '-' ||
+	    !strcmp(sb->buf, "refs/heads/HEAD"))
+		return -1;
+
 	return check_refname_format(sb->buf, 0);
 }
```

ethomson:
Thanks - I assumed so since refs are “case sensitive” where “case sensitive” means “duh of course they’re case sensitive and why are you running on Mac or Windows, you fool?”

carlosmn:
It wouldn't be the first time we're stricter, but yeah let's go with what git has decided.

ethomson:
Yep, let's add something to the CHANGELOG when we finish the release.  Since @pks-t has a goal to polish up the CHANGELOG, I'm going to go ahead and merge this.

PR #4488, opened 2018-01-21
Use longer conflict markers in recursive merge base

ethomson:
When producing a virtual merge base during a recursive merge (because our branches have multiple merge bases from a criss-cross merge), our merge base may contain conflicts.

[Git updated their virtual merge base building to use 9 character conflict markers](https://github.com/git/git/commit/d694a17986a28bbc19e2a6c32404ca24572e400f), which allows users (and programs like `rerere`) to better identify conflicts that occur in the virtual merge base vs conflicts that occur in the actual merge.

### DIFF
--- a/include/git2/merge.h
+++ b/include/git2/merge.h
@@ -162,6 +162,8 @@ typedef enum {
 	GIT_MERGE_FILE_DIFF_MINIMAL = (1 << 7),
 } git_merge_file_flag_t;
 
+#define GIT_MERGE_CONFLICT_MARKER_SIZE	7
+
 /**
  * Options for merging a file
  */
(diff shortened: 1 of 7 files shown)


include/git2/merge.h
```diff
@@ -162,6 +162,8 @@ typedef enum {
 	GIT_MERGE_FILE_DIFF_MINIMAL = (1 << 7),
 } git_merge_file_flag_t;
 
+#define GIT_MERGE_CONFLICT_MARKER_SIZE	7
```
pks-t: Shouldn't you also update `GIT_MERGE_FILE_OPTIONS_INIT`?

ethomson: 🤔

To clarify: the current `INIT` macro initializes it to `0`, which implies this default will be used.  Functionally, this is correct.

We could leave this as-is or we could update the macro to explicitly assign this default.  I feel like in other places (eg, checkout opts) this is the pattern we use, to leave things blank / null / zero and then assign a default after the fact.

I mention this only to be explicit about this.  If you think updating this is better (more visibility by the users?) then I can change it.

pks-t: Okay, I guess that's fine. The only implication of this is that the user cannot ever set the conflict marker size to `0`. I don't know whether there's any strange usecase that would want that, though.

PR #4491, opened 2018-01-21
Recursive merge: reverse the order of merge bases

ethomson:
When git encounters multiple merge bases during a recursive merge of two commits, it reverses the order of the merge bases before starting the recursive merge.  This was introduced in [8918b0c](https://github.com/git/git/commit/8918b0c9c2667c5a69461955135c709b09561f72):

> It seems to be the only sane way to do it: when a two-head merge is
> done, and the merge-base and one of the two branches agree, the
> merge assumes that the other branch has something new.
>
> If we start creating virtual commits from newer merge-bases, and go
> back to older merge-bases, and then merge with newer commits again,
> chances are that a patch is lost, _because_ the merge-base and the
> head agree on it. Unlikely, yes, but it happened to me.

I've updated the tests, with `git` itself, and ensuring that we produce the same output, so many of the unit tests were changed to ensure that we created identical trees to git.  With this change (along with #4488 and #4489), we now produce identical output.

I have included the changes from #4488 and #4489 here to illustrate.  I suggest we review and merge those and then I will rebase this on top of the updated master.


ethomson:
Also, I think that we should bring this base reversal change (but not the CR/LF or conflict marker changes) to backport.

pks-t:
Had the same though, thanks for taking care of it!

PR #4498, opened 2018-01-22
Pass virtual commits as first param

tyrielv:
Reversing the order of the bases after merge_base_many returns is insufficient for the issue I opened.

This PR includes a modified test case that still exposes the issue after your other changes, and proposed fix of reversing the order that commits are passed to merge_base_many.


tyrielv:
I'll try to clarify what the issue in libgit2 is that your reversal didn't fix.


Say we want to merge branch1 and branch2. The output of merge-base for these two is, in order, A, B, C.

git.exe:
1. Find merge bases for branch1 and branch2 -> A,B,C
1. reverses the order to C,B,A. 
1. Recursively merges C+B
    1. Find merge bases for C and B -> Q
    2. merge C and B with base Q -> V
1. Recursively merge V+A
    1. Find merge bases for V and A -> R
    2. merge V and A with base R -> W
1. Merge branch1 and branch2 with base W -> Result


libgit2 after your changes:
1. Find merge bases for branch1 and branch2 -> A,B,C
1. reverses the order to C,B,A. 
1. Recursively merges C+B
    1. Find merge bases for C and B -> Q
    2. merge C and B with base Q -> V
1. Recursively merge V+A
    1. Find merge bases for **C, B** and A -> **Q** *(always the same as the base between C and B)*
    2. merge V and A with base **Q** -> X
1. Merge branch1 and branch2 with base X -> Wrong Result

tyrielv:
with my proposed change:

Find merge bases for branch1 and branch2 -> A,B,C
reverses the order to C,B,A.
Recursively merges C+B 
Find merge bases for **B and C** -> Q
merge C and B with base Q -> V
Recursively merge V+A 
Find merge bases for **A, C, B** -> R
merge V and A with base R -> W
Merge branch1 and branch2 with base R -> Result

tyrielv:
Merge-base treats the first commit differently and the rest equally. So

merge-base(A,B) == merge-base(B,A)

merge-base(A,B,C) == merge-base(A,C,B)

but merge-base(C,B,A) != merge-base(A,B,C)

ethomson:
Ah, yes, of course.  Thanks for the reminder - I had lost track of this detail while debugging this over the weekend while dealing with CR/LF and conflict marker issues.

As you noted, `git merge-base`:

> Given three commits A, B and C, git merge-base A B C will compute the merge base between A and a hypothetical commit M

So, indeed, this is a good fix.  👍  (I would somewhat prefer that we could `merge-base` on our virtual commits.  When we added recursion, I pondered putting virtual commits into an inmemory ODB backend, but that seemed quite unnecessary.)

I'm pulling these changes into my branch manually:

1. The test that you wrote doesn't align with the tree that git produces.  When merging these two branches, git produces:

   ```
   100644 4a06b258fed8a4d15967ec4253ae7366b70f727d 0       targetfile.txt
   100644 b6bd0f9952f396e757d3f91e08c59a7e91707201 1       version.txt
   100644 f0856993e005c0d8ed2dc7cdc222cc1d89fb3c77 2       version.txt
   100644 2cba583804a4a6fad1baf97c959be447238d1489 3       version.txt
   ```

   I updated the test to expect that value of `targetfile.txt`, and the test passes...  🤔 

2. I've reformatted the changes.  We use the linux kernel / git formatting, so I added a newline to separate the comment from the line above, and I've indented the comment:

   ```
   /* comments start at the same indentation as the code they're commenting.
    * and have leading `*`s on continuation lines, that line up with the `*` above.
    */
   ```

3. I updated the commit message with the fix to be more verbose and explain the fix better.

We don't have a particular contribution policy, but there have been some pushes to move to strict bisectability - and not the failing test / fix pattern.  (But I don't feel like we've settled on that, so I've left the commits in the order they were in.  I just thought I'd mention that.)

Thanks again for the fix, sorry I got distracted by the other issues.

pks-t:
@ethomson: I will go ahead and create a pull request to have clar handle known failures. While this wasn't accepted upstream, I still think we should go with that and it would fix the bisectability problem.

PR #4499, opened 2018-01-25
sysdir: do not use environment in setuid case

pks-t:
In order to derive the location of some Git directories, we currently
use the environment variables $HOME and $XDG_CONFIG_HOME. This might
prove to be problematic whenever the binary is run with setuid, that is
when the effective user does not equal the real user. In case the
environment variables do not get sanitized by the caller, we thus might
end up using the real user's configuration when doing stuff as the
effective user.

The fix is to use the passwd entry's directory instead of $HOME in this
situation. As this might break scenarios where the user explicitly sets
$HOME to another path, this fix is only applied in case the effective
user does not equal the real user.

---

I've tested the logic of `get_passwd_home` with the following executable:

```
#include <assert.h>
#include <stdio.h>
#include <unistd.h>
#include <pwd.h>
#include <stdlib.h>
#include <string.h>
#include <errno.h>

static int get_passwd_home(char **out, uid_t uid)
{
	struct passwd pwd, *pwdptr;
	char *buf = NULL;
	long buflen;
	int error;

	assert(out);

	if ((buflen = sysconf(_SC_GETPW_R_SIZE_MAX)) == -1)
		buflen = 1024;

	do {
		buf = realloc(buf, buflen);
		buflen *= 2;
		error = getpwuid_r(uid, &pwd, buf, buflen, &pwdptr);
	} while (error == ERANGE);

	if (error) {
		puts( "failed to get passwd entry");
		goto out;
	}

	if (!pwdptr) {
		puts( "no passwd entry found for user");
		goto out;
	}

    *out = strdup(pwdptr->pw_dir);

out:
	free(buf);
	return error;
}

{
    uid_t euid, uid;
    char *env;

    euid = geteuid();
    uid = getuid();

    if (get_passwd_home(&env, euid) < 0)
        return -1;


    printf("HOME: %s\neuid: %d\nuid: %d\n", env ? env : "none", euid, uid);

    return 0;
}
```

### DIFF
--- a/src/sysdir.c
+++ b/src/sysdir.c
@@ -13,6 +13,9 @@
 #include <ctype.h>
 #if GIT_WIN32
 #include "win32/findfile.h"
+#else
+#include <unistd.h>
+#include <pwd.h>
 #endif
 
 static int git_sysdir_guess_programdata_dirs(git_buf *out)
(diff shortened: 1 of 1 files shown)


fcharlie:
👍

src/sysdir.c
```diff
@@ -34,12 +37,63 @@ static int git_sysdir_guess_system_dirs(git_buf *out)
...
+	assert(out);
+
+	if ((buflen = sysconf(_SC_GETPW_R_SIZE_MAX)) == -1)
+		buflen = 1024;
+
+	do {
+		buf = git__realloc(buf, buflen);
+		buflen *= 2;
+		error = getpwuid_r(uid, &pwd, buf, buflen, &pwdptr);
+	} while (error == ERANGE);
```
ethomson: Let's put a max on this, just in case?  😁

pks-t: Sure, why not. Restricted it to 8kb, which I'd hope should be plenty.

src/sysdir.c
```diff
@@ -34,12 +37,63 @@ static int git_sysdir_guess_system_dirs(git_buf *out)
...
+
+	assert(out);
+
+	if ((buflen = sysconf(_SC_GETPW_R_SIZE_MAX)) == -1)
+		buflen = 1024;
+
+	do {
+		buf = git__realloc(buf, buflen);
+		buflen *= 2;
+		error = getpwuid_r(uid, &pwd, buf, buflen, &pwdptr);
```
ethomson: Don't you want to double `buflen` size _after_ calling `getpwuid_r` so that it will be used to `realloc` in the failure case?

pks-t: Oops. Yes, thanks for spotting this!

PR #4500, opened 2018-01-26
[RFC] clar: known breakage

pks-t:
This PR introduces the concept of known breakages to clar. A known breakage is a bug which exists in our code base which we are currently unable to fix, but which is known to exist. While we have some such tests in our code base (e.g. online::badssl::old_cipher), we just `cl_skip()` those tests. This has two disadvantages:

- a test skipped because of known breakage cannot be distinguished from a test that was skipped e.g. because certain code was not compiled into our library
- in case the bug gets fixed, we will not get any feedback from our tests, as that test is still going to be just skipped over

With known breakages, we can fix this both problems. This PR introduces two new test statuses: "broken" ('B') for a known-broken test and "unbroken" ('U'). Tests which are known broken have a status code of 'B' such that they are easily discoverable, but those tests will _not_ cause an error. In case the breakage goes away, though, the test will return an "unbroken" status _and_ create an error. So first, we know about all broken tests, addressing the first disadvantage, and second we will know when the test gets unbroken, addressing the second disadvantage.

That being said, this was not accepted at clar upstream, based on the reasoning that a breakage should just be fixed (see https://github.com/vmg/clar/pull/77). I don't really think that works for open source projects (and probably also not for software projects in general), as contributors might just not have enough free time to fix the issue in case it is something deeper. But there is still value in documenting the issue in code to have a technically verifiable documentation of it. Anyway, cc'ing @vmg such that he can correct me if I've misrepresented his reasoning.

The above is also the reason why I made this PR an RFC, only. Furthermore, I think the naming is a bit hard to get right here, especially to macros like `clar__break`, `clar_libgit2_broken` etc. I'd welcome better proposals.


pks-t:
Heh. Tests fail on Windows because the test in online::badssl is not actually broken on Windows, but on Linux platforms only. So it already pays off :P

vmg:
I understand the reasoning here. I'm still not fond of the idea of merging this feature for upstream Clar but I think it makes sense for the libgit2 project.

pks-t:
Fair enough, I guess. Thanks for chiming in on that, @vmg!

tiennou:
I'm 👍 with this. I'm not sure I quite like the name too, but my (preferred) solution would be to have macros like `cl_(un)?expected_(success|failure)` (which is far-reaching), so maybe something like `cl_should_(fail|pass)` ?

PR #4501, opened 2018-01-26
CHANGELOG: update for v0.27.0

pks-t:
I feel like this should be the most important changes since v0.26.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,21 +6,75 @@ v0.26 + 1
 * Improved `p_unlink` in `posix_w32.c` to try and make a file writable
   before sleeping in the retry loop to prevent unnecessary calls to sleep.
 
+* The CMake build infrastructure has been improved to speed up building time.
+
+* A new CMake option "-DUSE_HTTPS=<backend>" makes it possible to explicitly
+  choose an HTTP backend.
+
+* A new CMake option "-DSHA1_BACKEND=<backend>" makes it possible to explicitly
+  choose an SHA1 backend. The collision-detecting backend is now the default.
+
+* A new CMake option "-DUSE_BUNDLED_ZLIB" makes it possible to explicitly use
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,26 +6,89 @@ v0.26 + 1
...
+* When using SSH-based transports, we now repeatedly ask for the passphrase to
+  decrypt the private key in case a wrong passphrase is being provided.
+
 ### API additions
 
 * `git_remote_create_detached()` creates a remote that is not associated
   to any repository (and does not apply configuration like 'insteadof' rules).
   This is mostly useful for e.g. emulating `git ls-remote` behavior.
 
+* `git_diff_patchid` lets you generate patch IDs for diffs.
```
carlosmn: I think this should be `git_diff_patchid()` for consistency.

CHANGELOG.md
```diff
@@ -6,26 +6,89 @@ v0.26 + 1
...
 * We now support conditional includes in configuration files.
 
+* Fix for handling re-reading of configuration files with includes.
+
+* Fix for reading patches which contain exact renames only.
+
+* Fix for reading patches with whitespace in the compared files' paths.
+
+* Fix for writing only the last refspec into FETCH_HEAD in case multiple
+  refspecs are being fetched.
```
carlosmn: This makes it sound like the fix we did is that only the last refspec is written, when it's the other way around.  Maybe something like

> We now fill `FETCH_HEAD` from all passed refspecs instead of overwriting with the last one.

pks-t:
Thanks @carlosmn, used both your improvements

pks-t:
Rebased and added another batch of changes which have been merged since the last time. Also, I moved the new conflict marker length option entry to "API additions", as we are not in fact breaking existing APIs but only extending the options structure.

PR #4503, opened 2018-01-29
quick start update

fcharlie:


README.md
```diff
@@ -30,6 +30,33 @@ Additionally, the example code has been released to the public domain (see the
 Quick Start
 ===========
 
+### Windows
+
+You can use vcpkg install libgit2:
```
pks-t: This should probably be "You can use vcpkg _to_ install libgit2:", so there's a "to" missing. Are additional instructions required to include the installed package in your MSVC project then or is this already covered somewhere else?

README.md
```diff
@@ -30,6 +30,33 @@ Additionally, the example code has been released to the public domain (see the
...
+
+```
+vcpkg install libgit2:x64-windows libgit2:x86-windows
+```
+
+### Ubuntu/Debian
+
+You can use apt/apt-get install libgit2
```
pks-t: I'd also add a "to" here.

README.md
```diff
@@ -30,6 +30,33 @@ Additionally, the example code has been released to the public domain (see the
...
+
+```
+sudo apt install libgit2-dev
+```
+Unfortunately, the libgit2 installed in this way is not necessarily the latest version
```
pks-t: Proposal:

"""
The version of libgit2 installed depends on the distribution. As such, it is not guaranteed that you will get the latest version.
"""

README.md
```diff
@@ -30,6 +30,33 @@ Additionally, the example code has been released to the public domain (see the
...
+
+### Mac
+
+You can use brew install libgit2:
```
pks-t: Proposal:

"""
You can use Homebrew to install libgit2:
"""

README.md
pks-t: Also, there's a ":" missing at the end.

README.md
```diff
@@ -51,6 +78,33 @@ Quick Start
 Trouble with these steps?  Read `TROUBLESHOOTING.md`.  More detailed build
 guidance is available below.
 
+**Submodule**
+
+If your project build tool is CMake, you can add libgit2 as a dependency on your project using `Git submodule`
```
pks-t: Proposal:

"If your project is using CMake as its build tool, you can add libgit2 as a dependency by embedding it as a Git 
submodule and then include it as a subdirectory in your build instructions."

pks-t: Furthermore, a ":" is missing at the end

README.md
```diff
@@ -51,6 +78,33 @@ Quick Start
...
+```
+Modify the contents of the CMakeLists.txt file:
+
+```cmake
+# suggest use set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+include_directories(
+    ./include
+    ./vendor/libgit2/include
+)
+add_subdirectory(vendor/libgit2)
```
pks-t: Even though these instructions are targeted at other projects, I'd like to keep our own style here:

```
# Add the libgit2 library
ADD_SUBDIRECTORY(vendor/libgit2)
# Add libgit2 headers to your executable's include directory
TARGET_INCLUDE_DIRECTORIES(${YOUR_TARGET} ${LIBGIT2_INCLUDES}
# Link your executable against libgit2
TARGET_LINK_LIBRARIES(${YOUR_TARGET} git2)
```

PR #4506, opened 2018-01-31
Respect core.filemode when comparing filemodes on checkout

### LINKED ISSUE #4504: checkout fails when checking out branches on an NTFS or Fat32 drive from MacOS or Linux
### Reproduction steps
Create a repo on an NTFS or Fat32 drive and try to checkout different branches from either MacOS or Linux.

### Expected behavior
Checkout works fine

### Actual behavior
Checkout fails due to potential filemode issues

Any version after [this PR was merged](https://github.com/libgit2/libgit2/pull/4369)

MacOSx
Ubuntu 16.04

This works fine on Windows since Windows does not care about filemode, but after said PR was merged there seem to be issues when the target drive does not care about filemodes on an OS that does.

I tracked it down finally and it turns out we get in a weird state where we have an added file in our current branch that is not in the branch I want to check out, and when the checkout code tries to remove this file it finds that the `mtime` and `mode` on the workdir version of the file do not match that of the index. This causes the code to check the `mode` against the `baseitem->mode` which always fails. Since we are doing a safe checkout the checkout says there is a conflict that needs to be resolved and fails.

[Here is where it fails](https://github.com/libgit2/libgit2/pull/4369/files#diff-67bc754cb0e13e1397e0448b7e747311L201)

Not sure what exactly causes the state or how to properly rectify it yet. Attached is a repo that can reproduce the issue. Just unzip it and checkout any of the other branches with libgit2. I tried to diagnose this myself but I ran out of time and haven't had another chance since I originally found the issue. Ed told me to open an issue here so maybe another contributor could have a crack at it if they have time.

[dummyrepo.zip](https://github.com/libgit2/libgit2/files/1674026/dummyrepo.zip)

tiennou:
Fixes #4504

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -70,6 +70,7 @@ typedef struct {
 	git_buf tmp;
 	unsigned int strategy;
 	int can_symlink;
+	int respect_filemode;
 	bool reload_submodules;
 	size_t total_steps;
 	size_t completed_steps;
@@ -159,7 +160,7 @@ GIT_INLINE(bool) is_workdir_base_or_new(
 		git_oid__cmp(&newitem->id, workdir_id) == 0);
 }
 
-GIT_INLINE(bool) is_file_mode_changed(git_filemode_t a, git_filemode_t b)
+GIT_INLINE(bool) is_file_mode_changed(int respect_filemode, git_filemode_t a, git_filemode_t b)
 {
 #ifdef GIT_WIN32
 	/*
@@ -168,7 +169,8 @@ GIT_INLINE(bool) is_file_mode_changed(git_filemode_t a, git_filemode_t b)
 	 */
 	return false;
 #else
-	return (S_ISREG(a) && S_ISREG(b) && a != b);
+	int mask = respect_filemode ? 0 : S_IXUSR|S_IXGRP|S_IXOTH;
+	return (S_ISREG(a) && S_ISREG(b) && (a & mask) != (b & mask));
 #endif
 }
 
@@ -214,7 +216,7 @@ static bool checkout_is_workdir_modified(
 	if ((ie = git_index_get_bypath(data->index, wditem->path, 0)) != NULL) {
 		if (git_index_time_eq(&wditem->mtime, &ie->mtime) &&
 			wditem->file_size == ie->file_size &&
-			!is_file_mode_changed(wditem->mode, ie->mode))
+			!is_file_mode_changed(data->respect_filemode, wditem->mode, ie->mode))
 			return !is_workdir_base_or_new(&ie->id, baseitem, newitem);
 	}
 
@@ -228,7 +230,7 @@ static bool checkout_is_workdir_modified(
 	if (S_ISDIR(wditem->mode))
 		return false;
 
-	if (is_file_mode_changed(baseitem->mode, wditem->mode))
+	if (is_file_mode_changed(data->respect_filemode, baseitem->mode, wditem->mode))
 		return true;
 
 	if (git_diff__oid_for_entry(&oid, data->diff, wditem, wditem->mode, NULL) < 0)
@@ -2446,6 +2448,10 @@ static int checkout_data_init(
 
 	if ((error = git_repository__cvar(
 			 &data->can_symlink, repo, GIT_CVAR_SYMLINKS)) < 0)
+		goto cleanup;
+
+	if ((error = git_repository__cvar(
+			 &data->respect_filemode, repo, GIT_CVAR_FILEMODE)) < 0)
 		goto cleanup;
 
 	if (!data->opts.baseline && !data->opts.baseline_index) {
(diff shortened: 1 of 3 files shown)


tiennou:
No experience with the checkout code (or the Windows permission model for that matter), but I'm suspicious of the `return false`. Shouldn't we at least check for read/write ?

src/checkout.c
```diff
@@ -70,6 +70,7 @@ typedef struct {
 	git_buf tmp;
 	unsigned int strategy;
 	int can_symlink;
+	int respect_filemode;
```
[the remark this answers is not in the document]
tiennou: It's inspired by `can_symlink`, which comes from config too. Is `bool` generally preferred ? I can switch them in that case, or just the new.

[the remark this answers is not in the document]
tiennou: A few remarks on that : it all comes down to `git__parse_bool` taking an `int *` as its argument. I've tried to fix the signature, but causes `bool` to appear in `config.h`, and I can't seem to find any public usage of `bool`/`stdbool.h` (our only reference is in cc-compat). So meh, but if you want a PR for that, I have the code.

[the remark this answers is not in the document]
ethomson: Nah, this is legit.  Sorry for leading you down a rather pointless rabbit hole.

tiennou:
I've extracted the relevant filemode comparison function in `utils.h` and added tests tests for combinations of filemodes. Note that I dumped tests in `futils.c` which is not quite the same namespace (suggestions welcome if a better home is warranted).

I'm WIPing to see how does AppVeyor react to using that code on Windows. If that works, I'll just `#ifdef` out the strict part, because it makes no sense there 🤞.

tiennou:
@ethomson The POSIX part is good to go, I'll defer to you on what to do to make Windows work, because I'm just headlessly hacking around in that last commit 😐.

I suspect the not-added-in-this-PR failing tests are because we're trying to do the full `core.filemode` check but I'm not entirely sure…

tiennou:
I got the filemode test to work everywhere (phew), but I'm now getting checkout failures as well. So either my test is completely wrong (Travis 👀😉) ~~those are running in a repo with an incorrect `core.filemode` value (which should have been handled by `git_repository_reinit_filesystem`)~~ (tests are never strict on Windows so `core.filemode` doesn't matter there) *or* we were wrong on Windows ??

Failing tests are [1](https://github.com/libgit2/libgit2/blob/master/tests/checkout/icase.c?utf8=%E2%9C%93#L269), [2](https://github.com/libgit2/libgit2/blob/master/tests/checkout/tree.c?utf8=%E2%9C%93#L379), [3](https://github.com/libgit2/libgit2/blob/master/tests/checkout/typechange.c?utf8=%E2%9C%93#L155).

src/util.h
```diff
@@ -539,4 +538,36 @@ GIT_INLINE(double) git__timer(void)
 
 extern int git__getenv(git_buf *out, const char *name);
 
+#ifdef WIN32
+#define S_ISREG(m) (((m) & _S_IFMT) == _S_IFREG)
```
tiennou: This is only temporary, but I didn't know where to put it (and VS2010 actually doesn't like it being redefined).

src/util.h
```diff
@@ -539,4 +538,36 @@ GIT_INLINE(double) git__timer(void)
...
+#endif
+
+GIT_INLINE(bool) git__is_filemode_equal(int strict, git_filemode_t a, git_filemode_t b)
+{
+#ifdef WIN32
```
tiennou: Unused `strict` parameter when WIN32-ing

src/util.h
```diff
@@ -539,4 +538,36 @@ GIT_INLINE(double) git__timer(void)
...
+		return a == b;
+
+#ifndef WIN32
+	/* when strict, we care about rw <=> non-rw or x <=> non-x changes */
+	if (strict)
+		return (!!(a & rw) == !!(b & rw)) && (!!(a & x) == !!(b & x));
+#endif
+
+	/* we only care about rw differences */
+	return (!!(a & rw) == !!(b & rw));
```
ethomson: I don't think there's a need to set up this `rw` mask, or test with it at all.  We know they're both files, you can simply `return true` here.

It's reasonable for a file to have _no_ read or write bits set, but this is still a file.  And if you can read the file despite those permissions (eg you're running as root) then this will incorrectly treat those file modes as different.

ethomson: That may actually be why the Windows build is failing.  I don't have a Windows machine in front of me, but it's possible that it's reading the lack of posix permissions as `0100000` for some file(s).

src/util.h
ethomson: Probably safer to:
```
#ifndef S_ISREG
# define S_ISREG...
#endif
```

tiennou: Actually, I wanted to move it in our compat `posix` headers for Windows (with the define guard), but I'm not sure where/which one I should use.

ethomson:
I pulled this down and spent some time to figure out why it was failing on Windows.  I did some refactoring and ended up with #4545.  There were some changes to support Win32 and POSIX both; I would love your feedback.

PR #4507, opened 2018-01-31
Honor 'GIT_USE_NSEC' option in `filesystem_iterator_set_current`

tomas:
This should have been part of PR #3638 (for issue #3635). Without this, we still get nsec-related errors, even when using -DGIT_USE_NSEC:

`error: ‘struct stat’ has no member named ‘st_mtime_nsec’`

### DIFF
--- a/src/iterator.c
+++ b/src/iterator.c
@@ -1458,10 +1458,12 @@ static void filesystem_iterator_set_current(
 	filesystem_iterator_entry *entry)
 {
 	iter->entry.ctime.seconds = entry->st.st_ctime;
-	iter->entry.ctime.nanoseconds = entry->st.st_ctime_nsec;
-
 	iter->entry.mtime.seconds = entry->st.st_mtime;
+
+#if defined(GIT_USE_NSEC)	
+	iter->entry.ctime.nanoseconds = entry->st.st_ctime_nsec;
 	iter->entry.mtime.nanoseconds = entry->st.st_mtime_nsec;
+#endif
 
 	iter->entry.dev = entry->st.st_dev;
 	iter->entry.ino = entry->st.st_ino;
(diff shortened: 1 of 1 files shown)


ethomson:
Thanks for catching this oversight.  Shouldn't we be zero'ing the `nanoseconds` field in this case?  eg:

   #if defined(GIT_USE_NSEC)
       iter->entry.ctime.nanoseconds = entry->st.st_ctime_nsec;
       iter->entry.mtime.nanoseconds = entry->st.st_mtime_nsec;
   #else
       iter->entry.ctime.nanoseconds = 0;
       iter->entry.mtime.nanoseconds = 0;
   #endif

tomas:
@ethomson I guess. I just followed the implementation on the original PR though, so that would also require updating. :)

ethomson:
Sorry, which original change are you looking at?

Generally we zero structs, but I think that in this particular case, we're iterating an index entry from an on-disk index that we've loaded.  That on-disk index _may_ have been written with a nanosecond supporting git client.  If so, the nanoseconds field in the entry would be nonzero.  If that's true, then we need to clear it, and not persist it...

tomas:
[This](https://github.com/libgit2/libgit2/pull/3638/files#diff-a11ba69fe5501a8caf440fca4a8b29baR831) one.

ethomson:
`git_index_entry__from_stat` is called on new, zero'd out index entries that we're creating.  They're generated in `index_entry_create` which [`calloc`s its memory](https://github.com/ethomson/libgit2/blob/3d6a42d1e14281c564d779e7490d761555d014d4/src/index.c#L886).

The index _iterators_ are very different beasts - we don't create new index entries with them, that we know will have zero'd nanosecond entries - we're refreshing index entries that we've loaded from the index on disk.  They may have non-zero nanosecond entries that we need to clear.

tomas:
Alrighty. Just updated the patch.

ethomson:
Thanks!

PR #4508, opened 2018-01-31
http: standardize user-agent addition

ethomson:
The winhttp and posix http each need to add the user-agent to their requests.  Standardize on a single function to include this so that we do not get the version numbers we're sending out of sync.

Currently we send `git/2.0 (libgit2 something)` on non-Windows and `git/1.0 (libgit2 something)` on Windows.  :(

### DIFF
--- a/src/transports/http.h
+++ b/src/transports/http.h
@@ -1,0 +1,23 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#ifndef INCLUDE_transports_http_h__
+#define INCLUDE_transports_http_h__
+
+#include "buffer.h"
+
+GIT_INLINE(int) git_http__user_agent(git_buf *buf)
+{
+	const char *ua = git_libgit2__user_agent();
+
+	if (!ua)
+		ua = "libgit2 " LIBGIT2_VERSION;
+
+	return git_buf_printf(buf, "git/2.0 (%s)", ua);
+}
+
+#endif
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -17,6 +17,7 @@
 #include "remote.h"
 #include "smart.h"
 #include "auth.h"
+#include "http.h"
 #include "auth_negotiate.h"
 #include "streams/tls.h"
 #include "streams/socket.h"
@@ -190,16 +191,6 @@ static int apply_credentials(git_buf *buf, http_subtransport *t)
 	return context->next_token(buf, context, cred);
 }
 
-static const char *user_agent(void)
-{
-	const char *custom = git_libgit2__user_agent();
-
-	if (custom)
-		return custom;
-
-	return "libgit2 " LIBGIT2_VERSION;
-}
-
 static int gen_request(
 	git_buf *buf,
 	http_stream *s,
@@ -211,7 +202,9 @@ static int gen_request(
 
 	git_buf_printf(buf, "%s %s%s HTTP/1.1\r\n", s->verb, path, s->service_url);
 
-	git_buf_printf(buf, "User-Agent: git/2.0 (%s)\r\n", user_agent());
+	git_buf_puts(buf, "User-Agent: ");
+	git_http__user_agent(buf);
+	git_buf_puts(buf, "\r\n");
 	git_buf_printf(buf, "Host: %s\r\n", t->connection_data.host);
 
 	if (s->chunked || content_length > 0) {
--- a/src/transports/winhttp.c
+++ b/src/transports/winhttp.c
@@ -18,6 +18,7 @@
 #include "remote.h"
 #include "repository.h"
 #include "global.h"
+#include "http.h"
 
 #include <wincrypt.h>
 #include <winhttp.h>
@@ -701,21 +702,6 @@ static int winhttp_close_connection(winhttp_subtransport *t)
 	return ret;
 }
 
-static int user_agent(git_buf *ua)
-{
-	const char *custom = git_libgit2__user_agent();
-
-	git_buf_clear(ua);
-	git_buf_PUTS(ua, "git/1.0 (");
-
-	if (custom)
-		git_buf_puts(ua, custom);
-	else
-		git_buf_PUTS(ua, "libgit2 " LIBGIT2_VERSION);
-
-	return git_buf_putc(ua, ')');
-}
-
 static void CALLBACK winhttp_status(
 	HINTERNET connection,
 	DWORD_PTR ctx,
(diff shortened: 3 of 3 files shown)


src/transports/http.h
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#ifndef INCLUDE_http_h__
+#define INCLUDE_http_h__
```
pks-t: The include guard is wrong, this should rather be `INCLUDE_transports_http_h__`

src/transports/http.h
```diff
@@ -0,0 +1,25 @@
...
+
+GIT_INLINE(const char *) git_http__user_agent(void)
+{
+	const char *custom = git_libgit2__user_agent();
+
+	if (custom)
+		return custom;
+
+	return "libgit2 " LIBGIT2_VERSION;
+}
```
pks-t: Why is the definition inline and not part of `http.c`?

src/transports/http.h
```diff
@@ -0,0 +1,25 @@
...
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#ifndef INCLUDE_http_h__
+#define INCLUDE_http_h__
+
+#include "buffer.h"
+
+#define GIT_HTTP_USER_AGENT_FMT "git/2.0 (%s)"
```
pks-t: Considering we never use `git_http__user_agent` without `GIT_HTTP_USER_AGENT_FMT`, why even bother making it a macro instead of just letting `git_http__user_agent` assemble the complete string?

src/transports/http.h
ethomson: The `http.c` and `winhttp.c` files are independent from each other.

src/transports/http.h
ethomson: It would be an awkward flow, since the fundamental strings are constructed differently for the two platforms.  One is complete, the other has `User-Agent: ` appended and newlines suffixed.

src/transports/http.h
ethomson: I guess.  We're not exactly consistent here.  I'll fix it, but let's make a pass to make them consistent if you expect them to be.

ethomson: Actually, I take that back.  This _is_ consistent with everything else in that folder.  Let's address these issues separately.  I'll get a PR together when I have some time.

src/transports/http.h
pks-t: Sure, I got that. The following easily handles that scenario but is still a lot more readable, I'd say:

```
diff --git a/src/transports/http.c b/src/transports/http.c
index 70a299322..6bebca910 100644
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -202,8 +202,7 @@ static int gen_request(
 
 	git_buf_printf(buf, "%s %s%s HTTP/1.1\r\n", s->verb, path, s->service_url);
 
-	git_buf_printf(buf, "User-Agent: " GIT_HTTP_USER_AGENT_FMT "\r\n",
-		git_http__user_agent());
+	git_buf_printf(buf, "User-Agent: %s\r\n", git_http__user_agent());
 	git_buf_printf(buf, "Host: %s\r\n", t->connection_data.host);
 
 	if (s->chunked || content_length > 0) {
diff --git a/src/transports/http.h b/src/transports/http.h
index 0ad1f80bf..f34ef8ee9 100644
--- a/src/transports/http.h
+++ b/src/transports/http.h
@@ -10,8 +10,6 @@
 
 #include "buffer.h"
 
-#define GIT_HTTP_USER_AGENT_FMT "git/2.0 (%s)"
-
 GIT_INLINE(const char *) git_http__user_agent(void)
 {
 	const char *custom = git_libgit2__user_agent();
@@ -19,7 +17,7 @@ GIT_INLINE(const char *) git_http__user_agent(void)
 	if (custom)
 		return custom;
 
-	return "libgit2 " LIBGIT2_VERSION;
+	return "git/2.0 (libgit2 "LIBGIT2_VERSION")";
 }
 
 #endif
diff --git a/src/transports/winhttp.c b/src/transports/winhttp.c
index c004dccc0..5138e3993 100644
--- a/src/transports/winhttp.c
+++ b/src/transports/winhttp.c
@@ -758,7 +758,7 @@ static int winhttp_connect(
 		return -1;
 	}
 
-	if ((error = git_buf_printf(&ua, GIT_HTTP_USER_AGENT_FMT, git_http__user_agent())) < 0) {
+	if ((error = git_buf_puts(&ua, git_http__user_agent())) < 0) {
 		git__free(wide_host);
 		return error;
 	}
```

I think the _real_ issue here is the custom user agent, though. We _could_ do something like the following to also enable the above with custom user agents. But I'm not sure I like that.

```
diff --git a/src/settings.c b/src/settings.c
index 2a52ffbf6..0fa6df13b 100644
--- a/src/settings.c
+++ b/src/settings.c
@@ -181,8 +181,12 @@ int git_libgit2_opts(int key, ...)
 #endif
 		break;
 	case GIT_OPT_SET_USER_AGENT:
+	{
+		git_buf ua = GIT_BUF_INIT;
+		git_buf_printf(&ua, "git/2.0 (%s)", va_arg(ap, const char *));
+
 		git__free(git__user_agent);
-		git__user_agent = git__strdup(va_arg(ap, const char *));
+		git__user_agent = git_buf_detach(&ua);
 		if (!git__user_agent) {
 			giterr_set_oom();
 			error = -1;
@@ -190,6 +194,7 @@ int git_libgit2_opts(int key, ...)
 
 		break;
 
+	}
 	case GIT_OPT_ENABLE_STRICT_OBJECT_CREATION:
 		git_object__strict_input_validation = (va_arg(ap, int) != 0);
 		break;
```

src/transports/http.h
pks-t: Yeah. It feels a bit awkward that "winhttp.c" refers to "http.h", but the headers complement "http.c" does in fact have nothing to do with "winhttp.c". I mean we could rename "http.c" to "unixhttp.c" and have "http.{c,h}" contain common code to both. But I'm definitly not keen on "unixhttp.c", and for the lack of a better name I'd just go with your proposal.

src/transports/http.h
ethomson: Right - your pattern has the same problem that we have now:  we're going to forget to change both places, and we'll end custom user-agents as `git/2.0 (my custom user agent)` and our default user agents as `git/3.0 (libgit2 ...)`.

pks-t: I agree. Seems like the _real_ issue is the API here. If we were able to return an allocated string we could just move assembling the custom user agent string into `git_http__user_agent`, as well, and be done about it.

ethomson: There's no need to introduce another allocation here.

pks-t:
@ethomson: I've amended a proposed commit, just because it's unhandy to make another proposal in a comment. Feel free to discard it if you prefer your approach, in which case I'd just say to go with yours.

ethomson:
@pks-t, I considered your strategy but preferred to be able to use a one-liner.  I'm happy enough with it, though, so I've squashed your commit into mine.

PR #4509, opened 2018-01-31
odb: error when we can't alloc an object

ethomson:
At the moment, we're swallowing the allocation failure.  We need to return the error to the caller.  (See analysis in #4470.)

### DIFF
--- a/src/hash/hash_win32.c
+++ b/src/hash/hash_win32.c
@@ -24,16 +24,20 @@ GIT_INLINE(int) hash_cng_prov_init(void)
 	DWORD dll_path_len, size_len;
 
 	/* Only use CNG on Windows 2008 / Vista SP1  or better (Windows 6.0 SP1) */
-	if (!git_has_win32_version(6, 0, 1))
+	if (!git_has_win32_version(6, 0, 1)) {
+		giterr_set(GITERR_SHA1, "CryptoNG is not supported on this platform");
 		return -1;
+	}
 
 	/* Load bcrypt.dll explicitly from the system directory */
 	if ((dll_path_len = GetSystemDirectory(dll_path, MAX_PATH)) == 0 ||
 		dll_path_len > MAX_PATH ||
 		StringCchCat(dll_path, MAX_PATH, "\\") < 0 ||
 		StringCchCat(dll_path, MAX_PATH, GIT_HASH_CNG_DLL_NAME) < 0 ||
-		(hash_prov.prov.cng.dll = LoadLibrary(dll_path)) == NULL)
+		(hash_prov.prov.cng.dll = LoadLibrary(dll_path)) == NULL) {
+		giterr_set(GITERR_SHA1, "CryptoNG library could not be loaded");
 		return -1;
+	}
 
 	/* Load the function addresses */
 	if ((hash_prov.prov.cng.open_algorithm_provider = (hash_win32_cng_open_algorithm_provider_fn)GetProcAddress(hash_prov.prov.cng.dll, "BCryptOpenAlgorithmProvider")) == NULL ||
@@ -44,12 +48,16 @@ GIT_INLINE(int) hash_cng_prov_init(void)
 		(hash_prov.prov.cng.destroy_hash = (hash_win32_cng_destroy_hash_fn)GetProcAddress(hash_prov.prov.cng.dll, "BCryptDestroyHash")) == NULL ||
 		(hash_prov.prov.cng.close_algorithm_provider = (hash_win32_cng_close_algorithm_provider_fn)GetProcAddress(hash_prov.prov.cng.dll, "BCryptCloseAlgorithmProvider")) == NULL) {
 		FreeLibrary(hash_prov.prov.cng.dll);
+
+		giterr_set(GITERR_OS, "CryptoNG functions could not be loaded");
 		return -1;
 	}
 
 	/* Load the SHA1 algorithm */
 	if (hash_prov.prov.cng.open_algorithm_provider(&hash_prov.prov.cng.handle, GIT_HASH_CNG_HASH_TYPE, NULL, GIT_HASH_CNG_HASH_REUSABLE) < 0) {
 		FreeLibrary(hash_prov.prov.cng.dll);
+
+		giterr_set(GITERR_OS "algorithm provider could not be initialized");
 		return -1;
 	}
 
@@ -57,6 +65,8 @@ GIT_INLINE(int) hash_cng_prov_init(void)
 	if (hash_prov.prov.cng.get_property(hash_prov.prov.cng.handle, GIT_HASH_CNG_HASH_OBJECT_LEN, (PBYTE)&hash_prov.prov.cng.hash_object_size, sizeof(DWORD), &size_len, 0) < 0) {
 		hash_prov.prov.cng.close_algorithm_provider(hash_prov.prov.cng.handle, 0);
 		FreeLibrary(hash_prov.prov.cng.dll);
+
+		giterr_set(GITERR_OS, "algorithm handle could not be found");
 		return -1;
 	}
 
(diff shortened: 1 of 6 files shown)


src/odb.c
```diff
@@ -75,6 +75,10 @@ static int odb_read_hardcoded(git_rawobj *raw, const git_oid *id)
 	raw->type = type;
 	raw->len = 0;
 	raw->data = git__calloc(1, sizeof(uint8_t));
+
+	if (!raw->data)
+		return -1;
```
pks-t: Why don't you use `GITERR_ALLOC_CHECK` here?

ethomson: At the moment, `odb_read_hardcoded` simply returns `-1` to indicate that the file was not found in the hardcoded object list.  I'll change this to actually distinguish between found/notfound and error separately.

ethomson:
I took a pass over several problematic areas and patterns around the ODB.  This should add error messages every place we return an error to the caller, and adds checks for error cases that were getting silently ignored.  Any place where there is a programmatic error, I've changed those from a `return -1` to a stronger `assert`, and there was one place where a function call could reasonably fall and we were `assert`ing.  I've downgraded that to a regular ol' error.

src/odb.c
```diff
@@ -1012,8 +1018,10 @@ static int odb_read_1(git_odb_object **out, git_odb *db, const git_oid *id,
 	bool found = false;
 	int error = 0;
 
-	if (!only_refreshed && odb_read_hardcoded(&raw, id) == 0)
-		found = true;
+	if (!only_refreshed) {
+		if ((error = odb_read_hardcoded(&found, &raw, id)) < 0)
+			return error;
```
pks-t: Shouldn't this rather be a `goto out`? While we obviously have nothing to clean up, it'd be a readability improvement and more consistent to our usual patterns.

src/odb.c
```diff
@@ -352,8 +352,7 @@ static int fake_wstream__write(git_odb_stream *_stream, const char *data, size_t
 {
 	fake_wstream *stream = (fake_wstream *)_stream;
 
-	if (stream->written + len > stream->size)
-		return -1;
+	assert(stream->written + len > stream->size);
```
pks-t: I don't think this makes sense. `assert`s are removed if compiling with `-DNDEBUG` (or something similar, don't know right now), so we would in effect remove this check for certain release builds. Shouldn't we stay on the safe side and always check, regardless of how libgit2 was built?

src/odb.c
```diff
@@ -84,19 +84,34 @@ static int odb_read_hardcoded(bool *found, git_rawobj *raw, const git_oid *id)
...
+	char *hdr,
+	size_t hdr_size,
+	git_off_t obj_len,
+	git_otype obj_type)
 {
 	const char *type_str = git_object_type2string(obj_type);
-	int len = p_snprintf(hdr, n, "%s %lld", type_str, (long long)obj_len);
-	assert(len > 0 && len <= (int)n);
-	return len+1;
+	int hdr_max = (hdr_size > INT_MAX-2) ? (INT_MAX-2) : (int)hdr_size;
```
pks-t: Why don't you use the `MAX` macro here?

pks-t: Especially since this line is not a maximum, but a minimum of `hdr_size` and `INT_MAX-2`. So even though its not wrong, it's a bit surprising to the reader

src/odb.c
```diff
@@ -110,16 +125,16 @@ int git_odb__hashobj(git_oid *id, git_rawobj *obj)
 		return -1;
 	}
 
-	hdrlen = git_odb__format_object_header(header, sizeof(header), obj->len, obj->type);
+	if ((error = git_odb__format_object_header(&hdrlen,
+		header, sizeof(header), obj->len, obj->type)) < 0)
+		return error;
```
pks-t: I think the indentation for all these line-broken if statements is a bit weird. I have no idea if that's because of how GitHub displays them, or if you really use all-tabs for the broken line and the first statement. Thing is, you can't easily discern the broken line from the first statement by indentation, which makes those statements harder to read than necessary.

I think in some places, we just indent the broken line by 4 spaces, instead of by tab.

src/hash/hash_win32.c
```diff
@@ -24,16 +24,20 @@ GIT_INLINE(int) hash_cng_prov_init(void)
...
+	}
 
 	/* Load bcrypt.dll explicitly from the system directory */
 	if ((dll_path_len = GetSystemDirectory(dll_path, MAX_PATH)) == 0 ||
 		dll_path_len > MAX_PATH ||
 		StringCchCat(dll_path, MAX_PATH, "\\") < 0 ||
 		StringCchCat(dll_path, MAX_PATH, GIT_HASH_CNG_DLL_NAME) < 0 ||
-		(hash_prov.prov.cng.dll = LoadLibrary(dll_path)) == NULL)
+		(hash_prov.prov.cng.dll = LoadLibrary(dll_path)) == NULL) {
+		giterr_set(GITERR_SHA1, "CryptoNG is not supported on this platform");
```
pks-t: It'd be nice to have a different message here than the other one, such that we have unique messages for which one can grep in the code. Something like "CryptoNG library not found", maybe?

src/odb.c
ethomson: Because the cast it important in the assignment, and only in the assignment, not in the comparison.

src/odb.c
ethomson: No, I don't think so.  The goal was to move it to an optional `assert`.  It's a rather pointless sanity check that could only happen if we broke our code.  It should be able to be compiled out.

src/odb.c
ethomson: I don't think so; this was intentional.  Doing that relies on knowing that `odb_read_hardcoded` will always exit having set `raw.data` to either a pointer that can be freed, or `NULL`.  This feels brittle, and if we change `odb_read_hardcoded` in the future to have another check and `return -1` before the allocation of `raw.data`, then we would try to dereference that in the `free`.

pks-t: Don't we assume this in most of our code when calling functions which have out-parameters? They either don't touch the out parameter at all in case of an error or sanitize it to always contain a value which can be handled.

src/odb.c
pks-t: Fair enough

src/odb.c
pks-t: :+1:

src/odb.c
ethomson: Yes - and here we're assuming that it doesn't touch it at all.

ethomson: We could have a `goto out` here if we changed the declaration to set the members to `NULL`.  I don't think that's a big win for readability, but if you do, then I can make that change.

src/hash/hash_win32.c
ethomson: Agreed; fixed.

src/odb.c
ethomson: (Or more accurately, the cast is important in assignment, and the lack of a cast is important in the comparison.)

src/odb.c
pks-t: I don't want to be too nitpicky. Just go with your current style ;)

src/hash/hash_win32.c
```diff
@@ -44,19 +48,25 @@ GIT_INLINE(int) hash_cng_prov_init(void)
...
+
+		giterr_set(GITERR_OS, "CryptoNG functions could not be loaded");
 		return -1;
 	}
 
 	/* Load the SHA1 algorithm */
 	if (hash_prov.prov.cng.open_algorithm_provider(&hash_prov.prov.cng.handle, GIT_HASH_CNG_HASH_TYPE, NULL, GIT_HASH_CNG_HASH_REUSABLE) < 0) {
 		FreeLibrary(hash_prov.prov.cng.dll);
+
+		giterr_set(GITERR_OS "algorithm provider could not be initialized");
```
csware: Missing `,`

pks-t: Oh, thanks for noticing! I'm surprised there was no compiler warning here

pks-t: I've just added it to #4527, as that already fixes some other smallish issues that should be merged quite fast

Issue #4480, opened 2018-01-17
Unnecessary syscalls on bare repository


Nukesor:
### Reproduction steps
Diff any two commits on bare repository and observe behaviour with strace.

### Expected behavior
Don't query stat for files on a bare repository.

### Actual behavior
Stat checked for every file in diff, if it's a bare repository.

### Version of libgit2 (release number or SHA1)
Master 782402c271d5467f9890af8e808af938a93bc55f

### Operating system(s) tested
Arch Linux

I just tried diffing with the current `master` libgit2 build. While the other bug with unnecessary syscalls for `.gitattributes` and `attributes` was fixed, there are still some `stat` calls for files, which shouldn't be necessary on a bare repository.

I'm using pygit2 for my interaction with git repositories, but as far as I can see, this is actually a issue with libgit2.

Current  output:
```
[pid 15375] access("/etc/gitattributes", F_OK) = -1 ENOENT (No such file or directory)
[pid 15375] stat("/tmp/gitalizer/phoemur/slackbuilds-1/info/attributes", 0x7ffe5758e5f0) = -1 ENOENT (No such file or directory)
[pid 15375] stat("python/python-webencodings/README", 0x7ffe5758e7e0) = -1 ENOENT (No such file or directory)
[pid 15375] stat("python/python-webencodings/python-webencodings.SlackBuild", 0x7ffe5758e7e0) = -1 ENOENT (No such file or directory)
[pid 15375] stat("python/python-webencodings/python-webencodings.SlackBuild", 0x7ffe5758e7e0) = -1 ENOENT (No such file or directory)
[pid 15375] stat("python/python-webencodings/python-webencodings.info", 0x7ffe5758e7e0) = -1 ENOENT (No such file or directory)
[pid 15375] stat("python/python-webencodings/python-webencodings.info", 0x7ffe5758e7e0) = -1 ENOENT (No such file or directory)
[pid 15375] stat("python/python-webencodings/slack-desc", 0x7ffe5758e7e0) = -1 ENOENT (No such file or directory)
[pid 15375] stat("python/python-webencodings/slack-desc", 0x7ffe5758e7e0) = -1 ENOENT (No such file or directory)
[pid 15375] stat("python/python-webencodings/README", 0x7ffe5758e7e0) = -1 ENOENT (No such file or directory)
[pid 15375] stat("python/python-webencodings/README", 0x7ffe5758e7e0) = -1 ENOENT (No such file or directory)
[...]

PR #4510, opened 2018-02-01
attr: avoid stat'ting files for bare repositories

pks-t:
Depending on whether the path we want to look up an attribute for is a
file or a directory, the fnmatch function will be called with different
flags. Because of this, we have to first stat(3) the path to determine
whether it is a file or directory in `git_attr_path__init`. This is
wasteful though in bare repositories, where we can already be assured
that the path will never exist at all due to there being no worktree. In
this case, we will execute an unnecessary syscall, which might be
noticeable on networked file systems.

What happens right now is that we always pass the `GIT_DIR_FLAG_UNKOWN`
flag to `git_attr_path__init`, which causes it to `stat` the file itself
to determine its type. As it is calling `git_path_isdir` on the path,
which will always return `false` in case the path does not exist, we end
up with the path always being treated as a file in case of a bare
repository. As such, we can just check the bare-repository case in all
callers and then pass in `GIT_DIR_FLAG_FALSE` ourselves, avoiding the
need to `stat`. While this may not always be correct, it at least is no
different from our current behavior.

### DIFF
--- a/src/attr.c
+++ b/src/attr.c
@@ -56,12 +56,16 @@ int git_attr_get(
 	git_attr_file *file;
 	git_attr_name attr;
 	git_attr_rule *rule;
+	git_dir_flag dir_flag = GIT_DIR_FLAG_UNKNOWN;
 
 	assert(value && repo && name);
 
 	*value = NULL;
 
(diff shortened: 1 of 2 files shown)


pks-t:
This might fix #4480

PR #4511, opened 2018-02-01
mmap cleanups

pks-t:
Some smaller mmap/POSIX layer cleanups.

Noteworthy is the last commit, which might be controversial. My goal was to follow what upstream does and use an optimization where we use `read` instead of `mmap` for small read-only files. The claim is an increase in throughput of up to 39% for files smaller than 1MB (see ea68b0ce9 (hash-object: don't use mmap() for small files, 2010-02-21)). I didn't invest that much time into benchmarking, but at least for the benchmarks I did (running assorted tests from our test suite) I couldn't see any immediate improvements. Anyway, I'm still putting this up for discussion.

Disclaimer: please don't merge this as-is. I saw very infrequent crashes, though I have no idea whether the cause is my mmap-stuff or already present in the master branch.


tiennou (dismissed):
I like it, and haven't seen anything worth commenting on, so 👍.

ethomson:
I'm generally +1 on this, but I'm curious why you've put the upstream's `read` fallback into `p_mmap`...  Moving this into `git_futils_mmap_ro` might be somewhat more straightforward.  (I expect the `p_` functions to basically be direct maps to the posix functions on posix, and a reimplementation on win32.)  It would also give win32 the same speed benefit.

pks-t:
Which would make sense if `p_mmap` and `git_futils_mmap_ro_file` had different structures. But as both use `git_map` right now, it would  become kind of awkward due to the `malloced` bit. While we could just have the high-level futils function create optimized mappings, the lower-level mmap would need to ensure the bit is being unset again. Furthermore, the `git_map` structure can in theory be free'd by calling either `git_futils_mmap_free` or `p_unmap`, so the latter needs to know about that bit, too.

Solution would be to either create a new structure specific to the futils_mmap, or to implement the optimization in `git_futils_mmap_ro` and `p_unmap`, resulting in a layering violation. I think the current approach is more reasonable.

ethomson:
But why is `p_mmap` using a structure at all?  The `p_` functions should basically be emulations of the posix functions.  `p_mmap` (on POSIX) looks like a rather unnecessary wrapper around `mmap`, and it feels to me like our high-level `git_futils_mmap` should be adding the high-level wrapper bits.

I propose that either:
* Make POSIX `p_mmap` a direct call to `mmap` (perhaps with the sanity checking that exists now), and implement the high-level logic in the higher level, or
* Remove `p_mmap` entirely, and have two implementations of our higher-level mmap code with this fancy switch.

pks-t:
Yeah, the reason here is Windows. It cannot identify a mapping by its pointer only, but we need to keep track of its handle to properly free the mapping later. That's why we have a platform-dependent structure `git_map`. We simply cannot use a void pointer here as the out parameter of `p_mmap` if we want to use it on multiple platforms.

I'd be hesistant to move all the mmap-logic up into the futils code, as the Windows emulation layer is quite complex, so it needs to be retained. Should we rename it? I don't know -- it is the closest possible variant of `p_mmap` we can support, so I'd say it is fine.

On the other hand, I'd still like to disentangle it with the futils code, they shouldn't use the same structure. So my proposed way here is to

- keep the `git_map` and `p_mmap`/`p_munmap` code as-is
- create a new structure `git_filemap`, which embeds a`p_mmap`
- move up the optimization code into futils
- create another function `git_futils_mmap_rw`, such that the only other user of `p_mmap` can be replaced by futils code and nobody else makes use of that low-level function

pks-t:
This is how that separation would look like. I first tried to use the `git_map` interface for that, but a lot of code uses `git_map` not as an mmapped data structure but simply as a structure keeping a memory pointer. So I've instead created a more generic `git_mem` structure, which can either contain an mmap or a allocated memory. The POSIX mmap stuff is now properly contained in "posix.h" and only used in mem.c.

Let me know what you think. I didn't yet write documentation and proper commit messages, as I don't want to waste effort in case you think this is bad.

pks-t:
Rebased to fix conflicts.

pks-t:
Rebased again to fix another conflict. I've also moved down the optimization of using malloc/read into `git_mem_from_path`, as the mwindow code had some problems when it was in `git_mem_from_fd`. I didn't yet investigate, but at least having it inside of `git_mem_from_path` optimizes lookup of small loose objects as the ODB makes use of that function.

pks-t:
Finally green and ready to go.

pks-t:
@ethomson: you've already voiced your support in the past -- do you intend to to have another look or shall I just go ahead?

PR #4512, opened 2018-02-01
Consistent header guards

ethomson:
Use consistent names for the #include / #define header guard pattern.

### DIFF
--- a/src/cc-compat.h
+++ b/src/cc-compat.h
@@ -4,8 +4,8 @@
  * This file is part of libgit2, distributed under the GNU GPL v2 with
  * a Linking Exception. For full terms see the included COPYING file.
  */
-#ifndef INCLUDE_compat_h__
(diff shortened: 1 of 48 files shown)


ethomson:
Truly this is the bikesheddiest pull request that I have ever opened.

pks-t:
Haha, it probably is. Yesterday I had an itch to rename all include guards such that they do also contain the "libgit2" string, just in case there was anybody else who uses a similar naming scheme who'd collide with our name guards. But as we're already using "_git_" in the public includes, I just thought it to be too unlikely to ever have a collision, as our dependencies are not changing that much.

Anyway, I think it still makes sense to be consistent with our current naming scheme. While it's a lot of churn without any immediate profit, the conversion is trivial enough. Also, it _might_ just save us headache in the future when whe _do_ add another file which ends up colliding by accident.

PR #4513, opened 2018-02-03
CMake: minor fixups

ethomson:
Some minor fixups and refactoring for CMake:

1. Move the `CMP0051` policy check to use the policy check functionality instead of guessing based on version number.
2. Enable the `CMP0042` policy, for macOS; this has no functional change, but it quiets a warning about how it's being applied.
3. Move some things out of the `CMakeList.txt` into modules.


pks-t:
Thanks, these definitly help to make our build instructions easier to read. I think 412e91b is missing a piece, you don't ever include that new module.

ethomson:
> Thanks, these definitly help to make our build instructions easier to read. I think 412e91b is missing a piece, you don't ever include that new module.

It was missing that, thanks @pks-t .

PR #4519, opened 2018-02-07
refspec: add public parsing api

cynecx:
This PR adds public apis for parsing refspec strings. This already exists but they are all internal.
I am not sure if tests are required here because I guess that the internal api must have tests already and this PR just adds trivial wrapping. But I am happy to add tests if it's really requested.

### DIFF
--- a/include/git2/refspec.h
+++ b/include/git2/refspec.h
@@ -22,6 +22,23 @@
 GIT_BEGIN_DECL
 
 /**
+ * Parse a given refspec string
+ *
+ * @param refspec a pointer to hold the refspec handle
+ * @param input the refspec string
+ * @param is_fetch is this a refspec for a fetch
+ * @return 0 if the refspec string could be parsed, -1 otherwise
+ */
+GIT_EXTERN(int) git_refspec_parse(git_refspec **refspec, const char *input, int is_fetch);
+
+/**
+ * Free a refspec object which has been created by git_refspec_parse
+ *
+ * @param refspec the refspec object
+ */
+GIT_EXTERN(void) git_refspec_free(git_refspec *refspec);
+
+/**
  * Get the source specifier
  *
  * @param refspec the refspec
--- a/src/refspec.c
+++ b/src/refspec.c
@@ -160,6 +160,31 @@ void git_refspec__free(git_refspec *refspec)
 	memset(refspec, 0x0, sizeof(git_refspec));
 }
 
+int git_refspec_parse(git_refspec **out_refspec, const char *input, int is_fetch)
+{
+	git_refspec *refspec;
+	assert(out_refspec && input);
+
+	*out_refspec = NULL;
+
+	refspec = git__malloc(sizeof(git_refspec));
+	GITERR_CHECK_ALLOC(refspec);
+
+	if (git_refspec__parse(refspec, input, !!is_fetch) != 0) {
+		git__free(refspec);
+		return -1;
+	}
+
+	*out_refspec = refspec;
+	return 0;
+}
+
+void git_refspec_free(git_refspec *refspec)
+{
+	git_refspec__free(refspec);
+	git__free(refspec);
+}
+
 const char *git_refspec_src(const git_refspec *refspec)
 {
 	return refspec == NULL ? NULL : refspec->src;
--- a/tests/network/refspecs.c
+++ b/tests/network/refspecs.c
@@ -158,3 +158,15 @@ void test_network_refspecs__matching(void)
 
 	git_refspec__free(&spec);
 }
+
+void test_network_refspecs__parse_free(void)
+{
+	git_refspec *spec = NULL;
+
+	cl_git_fail(git_refspec_parse(&spec, "", 0));
+	cl_git_fail(git_refspec_parse(&spec, ":::", 0));
+	cl_git_pass(git_refspec_parse(&spec, "HEAD:", 1));
+
+	cl_assert(spec != NULL);
+	git_refspec_free(spec);
+}


src/refspec.c
```diff
@@ -160,6 +160,35 @@ void git_refspec__free(git_refspec *refspec)
 	memset(refspec, 0x0, sizeof(git_refspec));
 }
 
+int git_refspec_parse(git_refspec **refspec, const char *input, int is_fetch) {
+	git_refspec *new_refspec;
+	if (refspec == NULL || input == NULL) {
```
tiennou: That should be an assert.

src/refspec.c
```diff
@@ -160,6 +160,35 @@ void git_refspec__free(git_refspec *refspec)
...
+		return -1;
+	}
+
+	is_fetch = !!is_fetch;
+	new_refspec = git__malloc(sizeof(git_refspec));
+	if (new_refspec == NULL) {
```
tiennou: `GITERR_CHECK_ALLOC` is what you want here.

src/refspec.c
```diff
@@ -160,6 +160,35 @@ void git_refspec__free(git_refspec *refspec)
...
+		git__free(new_refspec);
+		return -1;
+	}
+
+	*refspec = new_refspec;
+	return 0;
+}
+
+void git_refspec_free(git_refspec *refspec) {
+	if (refspec == NULL) {
```
tiennou: I'd be tempted to say that this test is not needed : `git_refspec__free` already handles it, and `git__free` doesn't care about `NULL`.

src/refspec.c
```diff
@@ -160,6 +160,35 @@ void git_refspec__free(git_refspec *refspec)
 	memset(refspec, 0x0, sizeof(git_refspec));
 }
 
+int git_refspec_parse(git_refspec **refspec, const char *input, int is_fetch) {
+	git_refspec *new_refspec;
+	if (refspec == NULL || input == NULL) {
+		return -1;
+	}
+
+	is_fetch = !!is_fetch;
```
tiennou: I'd move that inside the `git_refspec__parse` call.

tiennou:
It'd be 😍 if you squashed those commits, too !

src/refspec.c
cynecx: done

src/refspec.c
cynecx: done

src/refspec.c
cynecx: done

cynecx:
@tiennou Done!

include/git2/refspec.h
```diff
@@ -21,6 +21,23 @@
  */
 GIT_BEGIN_DECL
 
+/**
+ * Parse a given refspec string
+ *
+ * @param refspec a pointer to hold the refspec handle
+ * @param input the refspec string
+ * @param is_fetch is this a refspec for a fetch
+ * @return 0 if the refspec string could be parsed, 1 otherwise
```
pks-t: `1` otherwise is a bit weird. We always return `-1` for errors.

src/refspec.c
```diff
@@ -160,6 +162,27 @@ void git_refspec__free(git_refspec *refspec)
 	memset(refspec, 0x0, sizeof(git_refspec));
 }
 
+int git_refspec_parse(git_refspec **refspec, const char *input, int is_fetch) {
```
pks-t: Well, this is definitly is a bit more than a "simple wrapper", regarding it has 14 lines of code. So it should be worth a test

src/refspec.c
```diff
@@ -160,6 +162,27 @@ void git_refspec__free(git_refspec *refspec)
...
+	git_refspec *new_refspec;
+	assert(refspec && input);
+
+	new_refspec = git__malloc(sizeof(git_refspec));
+	GITERR_CHECK_ALLOC(new_refspec);
+
+	if (!git_refspec__parse(new_refspec, input, is_fetch)) {
```
pks-t: You shouldn't pass `is_fetch` as, as `git_refspe__parse` accepts a bool. So instead of moving `is_fetch = !!is_fetch` into `git_refspec__parse`, you should have that here.

src/refspec.c
ethomson: While you're fixing up this PR with @pks-t 's changes, please put this `{` on a newline to match our code formatting rules.  (Thanks!)

src/refspec.c
```diff
@@ -160,6 +162,27 @@ void git_refspec__free(git_refspec *refspec)
...
+		git__free(new_refspec);
+		return -1;
+	}
+
+	*refspec = new_refspec;
+	return 0;
+}
+
+void git_refspec_free(git_refspec *refspec) {
```
ethomson: Same with this `{`.

src/refspec.c
pks-t: Ah, thanks! I noticed this, but forgot to write a comment

ethomson:
> I'm a bit on the edge whether that API makes sense to expose. Not saying that it shouldn't, but I still would like to know of a certain use case where it would be handy. I'm especially coming from the angle that we have no functions exposed which make real use of the refspec.

Same.  @cynecx, understanding your motivations here would help us get more excited about this.  As I'm sure you can imagine, every new API we add is something else that we have to maintain, so we're loathe to add new public API unless it's warranted.

We don't want to just expose API for the sake of completeness.  So understanding what you're trying to do (and how this helps you do it) is nice so that we know that we should take on this future maintenance burden in order to help you build your software.  😃

cynecx:
Sorry about the delay.

@pks-t @ethomson Well, I am currently developing a custom git remote handler which accepts a command which accepts a refspec as an argument (and my handler obviously uses libgit2). I was actually aiming for reducing code duplication by reusing parsing code which already exists in libgit2. That's my main motivation here, I am not sure if this will convince you but it would definitely offload some parsing logic, which is always good IMO.

include/git2/refspec.h
cynecx: Sorry the function is actually returning `-1`, but this is obviously a documentation error.

src/refspec.c
cynecx: @tiennou mentioned to move that line away. But I can move it back here.

src/refspec.c
cynecx: Sorry, didn't noticed that. Should be fixed.

src/refspec.c
cynecx: I'll push some tests right away.

cynecx: @pks-t About the tests, would a simple valid/not-valid parsing case be sufficient (testing object creation) and a simple comment that the logic should already be tested in the internal version of the api suffice?

pks-t: @cynex Should suffice in this case, yes.

src/refspec.c
pks-t: Sorry, coding style is always a soft thing and different people have different views :P

pks-t: On another note: you could also just move it into the `git_refspec__parse` line, without first assigning it to itself.

tests/network/refspecs.c
```diff
@@ -158,3 +158,22 @@ void test_network_refspecs__matching(void)
...
 	git_refspec__free(&spec);
 }
+
+void test_network_refspecs__parse_free(void) {
+	git_refspec *spec;
+	int error;
+
+	spec = NULL;
+
+	error = git_refspec_parse(&spec, "", 0);
```
pks-t: You can get rid of the `error` for this and all the following invocations, as we're always using `cl_git_pass` and `cl_git_fail` to check for errors when the invoked function is conforming to libgit2's error style. So:

```
cl_git_fail(git_refspec_parse(&spec, "", 0));
cl_git_fail(git_refspec_parse(&spec, ":::", 0));
cl_git_pass(git_refspec_pares(&spec, "HEAD:", 1));
```

src/refspec.c
```diff
@@ -160,6 +160,30 @@ void git_refspec__free(git_refspec *refspec)
 	memset(refspec, 0x0, sizeof(git_refspec));
 }
 
+int git_refspec_parse(git_refspec **refspec, const char *input, int is_fetch)
+{
+	git_refspec *new_refspec;
+	assert(refspec && input);
+
```
pks-t: I'd like to make sure that we always sanitize the out parameter. So the first thing this function does should be `*refspec = NULL`.

src/refspec.c
```diff
@@ -160,6 +160,30 @@ void git_refspec__free(git_refspec *refspec)
...
+	new_refspec = git__malloc(sizeof(git_refspec));
+	GITERR_CHECK_ALLOC(new_refspec);
+
+	is_fetch = !!is_fetch;
+	if (git_refspec__parse(new_refspec, input, is_fetch) != 0) {
```
pks-t: You could just pull the `!!is_fetch` into this function. So `if (git_refspec__parse(new_refspec, input, !!is_fetch) != 0)`

src/refspec.c
```diff
@@ -160,6 +160,30 @@ void git_refspec__free(git_refspec *refspec)
 	memset(refspec, 0x0, sizeof(git_refspec));
 }
 
+int git_refspec_parse(git_refspec **refspec, const char *input, int is_fetch)
```
pks-t: Minor nit that doesn't need addressing: we usually name out-parameters `out` or similar. In case you rename it to out, you could also just rename `new_refspec` to `refspec`.

cynecx:
I've squashed them and addressed the styling issues :)

cynecx:
ping.

pks-t:
Kicking CI as it had some spurious failures.

pks-t:
Thanks for your contribution!

Issue #3053, opened 2015-04-17
Support `core.longpaths` on Windows


ethomson:
At some point we need to support the `core.longpaths` configuration setting and deal with Windows paths that are longer than `MAX_PATH`.
- [ ] Change all the UTF8 to UTF16 code to not assume `MAX_PATH`
- [ ] Change the checkout path sanitizer to (optionally) allow for long paths
- [ ] Tests tests tests tests tests

Questions:
- [ ] Do long paths work over SMB?

We will need copious tests here:
- [ ] Test that we can create a folder whose absolute path would be longer than 260 characters
- [ ] Test that a file can be created whose absolute path would be longer than 260 characters (inside a directory that is not excessively long)
- [ ] Test a filename (just the name portion) that is itself longer than 260 characters
- [ ] Ensure that we can't suddenly write long references

DavidSimonsAdobe:
This is my #1 request for libgit2. I'd be happy to help test it in our app. We need no interop with other git tools. All access to the private repo is via libgit2 (on mac & win). Possibly on a remote file system.

rhuijben:
Answering the question: **yes** long paths work over SMB if you properly use \\\\?\\UNC\\server\\share\\etc...

csware:
In Git for Windows they have a working workaround see https://github.com/git-for-windows/git/blob/master/compat/mingw.h and https://github.com/git-for-windows/git/blob/master/compat/mingw.c; search for MAX_LONG_PATH.

michelschep:
I'm having the same issue with long file names. 
With git via command line no issue (thanks to "git config --system core.longpaths true"). 
However I also use GitKraken which depends on nodegit  which depends on libgit2. 
Is this issue already solved (and nodegit/GitKraken is not aware of it/not using it) or a PR is required (unfortunately I'm not into C) or already PR is in progress?

ethomson:
No, there is no work in progress yet to deal with this.

gw79:
My company is trying to develop an internal tool to automate cloning many repositories to more easily create and configure a full system developer workspace - we've got some long path problems and having this support would make them go away and save the pain of refactoring a lot of code to reduce file paths.

Unfortunately my knowledge of C is limited to 2 terms of C++ at uni 20 years ago so I doubt my efforts at forking would be very productive ;)

ethomson:
Indeed, this is regrettable, but this remains far from trivial.

[...]

PR #4520, opened 2018-02-08
Support long file and branch names in checkout conflict creation

novalis:
When we have a conflict, we generate conflict files of the form
filename~branchname, or sometimes filename~branchname_[digits], if
there already exists filename~branchname.  When the length of this
would be over NAME_MAX, shorten it by truncating one or more
of the components.


novalis:
(please ignore for a bit, hacking Windows stuff which I can only do through appveyor since I don't run Windows)

novalis:
OK, did what I should have done in the first place.  Please review.

src/checkout.c
```diff
@@ -1966,34 +1966,109 @@ static int conflict_entry_name(
...
+	size_t basename_len,
+	const char *suffix,
+	size_t suffix_len,
+	int sequence) {
+	size_t basename_max = (NAME_MAX * 2) / 3;
+	size_t suffix_max = (NAME_MAX - basename_max) - 1;
+	size_t sequence_len;
+	char sequence_buf[(CHAR_BIT * sizeof(int) / 3) + 3];
+
+	/* Figure out the length of the suffix inclusive of the
```
pks-t: We try to use either 

```
/*
 * Foobar
 */
```
or
```
/* Foobar */
```
style comments

src/checkout.c
```diff
@@ -1966,34 +1966,109 @@ static int conflict_entry_name(
 	return 0;
 }
 
+static int checkout_path_suffixed_0(
+	git_buf *path,
+	size_t basename_offset,
+	size_t basename_len,
+	const char *suffix,
+	size_t suffix_len,
+	int sequence) {
```
pks-t: The function-opening brace belongs on the next line

src/checkout.c
```diff
@@ -1966,34 +1966,109 @@ static int conflict_entry_name(
...
+	size_t basename_max = (NAME_MAX * 2) / 3;
```
pks-t: A comment explaining this would help a lot

src/checkout.c
```diff
@@ -1966,34 +1966,109 @@ static int conflict_entry_name(
...
+	size_t suffix_max = (NAME_MAX - basename_max) - 1;
+	size_t sequence_len;
+	char sequence_buf[(CHAR_BIT * sizeof(int) / 3) + 3];
+
+	/* Figure out the length of the suffix inclusive of the
+	 * sequence number */
+	if (sequence != -1) {
+		/* Would like to use snprintf here, but it's C99 */
```
pks-t: We do have `p_snprintf`

src/checkout.c
```diff
@@ -1966,34 +1966,109 @@ static int conflict_entry_name(
 	return 0;
 }
 
+static int checkout_path_suffixed_0(
```
pks-t: The name is a bit awkward. What exactly does this function do that `checkout_path_suffixed` doesn't?

src/checkout.c
```diff
@@ -1966,34 +1966,109 @@ static int conflict_entry_name(
...
-	int i = 0, error = 0;
+	int i = 0, error = 0, ret = 0;
+	size_t basename_len;
+	size_t suffix_len;
+	size_t basename_offset;
+	git_buf saved_path = GIT_BUF_INIT;
 
-	if ((error = git_buf_putc(path, '~')) < 0 || (error = git_buf_puts(path, suffix)) < 0)
-		return -1;
+	git_buf_put(&saved_path, git_buf_cstr(path), git_buf_len(path));
```
pks-t: You should probably check for an error here

src/checkout.c
```diff
@@ -1966,34 +1966,109 @@ static int conflict_entry_name(
...
+	 * filename) is less than NAME_MAX, which is typically 255.
+	 * We're about to append a suffix (which might include some
+	 * digits)
+	 *
+	 * If our filename is already too long, we need to truncate
+	 * it.  We'll arbitrarily divide the available space as
+	 * follows: filename: 2/3; suffix and digits: 1/3 - 1 (for the
+	 * '~').  But if either filename or suffix doesn't completely
+	 * fill its allotted space, we allow the other to take up the
+	 * slack.
```
pks-t: Aren't these (at least partially) implementation details of `checkout_path_suffixed_0`?

src/checkout.c
```diff
@@ -1966,34 +1966,109 @@ static int conflict_entry_name(
...
-			(error = git_buf_printf(path, "%d", i)) < 0)
-			return error;
+	if ((error = checkout_path_suffixed_0(path, basename_offset, basename_len,
+					      suffix, suffix_len, -1)) < 0) {
+		ret = error;
+		goto out;
+	}
 
-		i++;
+	while (git_path_exists(git_buf_cstr(path)) && i < INT_MAX) {
```
pks-t: `INT_MAX`? I feel like we should restrict this to a sane value, shouldn't we? Checking 2 billion paths hardly feels like the right thing to do

tests/checkout/conflict.c
```diff
@@ -50,13 +50,13 @@ struct checkout_index_entry {
 	uint16_t mode;
 	char oid_str[GIT_OID_HEXSZ+1];
 	int stage;
-	char path[128];
+	char path[256];
```
pks-t: Why don't you just use `PATH_MAX` here and in the following declarations?

novalis: I think these are not actually paths, but filenames.  So NAME_MAX + 1  is probably what's wanted.

src/checkout.c
novalis: I don't know what value would be reasonable.  Whatever limit we put here, someone could create a situation that fails it.  How would you feel if I made it a config value with a default of (say) 10?

novalis:
> I do feel like you should somehow split out that logic to only be used on Windows. On Unix-based systems this is not the right thing to do, I'd argue, but there we should probably just error out. Also, PATH_MAX is usually set to 4096 there, which is much more reasonable than 256.

I don't understand what you're saying here.

This patch is in fact *Unix*-only.  It doesn't work on Windows because Windows doesn't support long file names without funny shenanigans (see #3053, mentioned in the comments here).  We hit this issue on Unix, where we couldn't check out certain files.  I would love to fix it for Windows as well, but when I messed around with it, I discovered that I would have to fix #3053 first, and that's a fair amount of work.

novalis:
Other than the INT_MAX thing, I pushed a change for the rest of the requested changes, I think

pks-t:
Sorry, I was confused a bit. Whenever I hear "long" and "file" I actually think about the Windows thing "core.longpath". So please just ignore that bit

src/checkout.c
pks-t: I agree. I'd aim a bit higher than 10, though. After lots of back and forth I think that 64 is the correct value.

novalis:
The Travis build seems to be messed up -- it can't download from Debian.  Maybe it needs a kick?

pks-t:
Yeah, Travis is having problems again. I already started kicking some other PRs, but now it's the macOS builds which cannot catch up. I'll give jobs another kick as soon as the queue has been processed

novalis:
Is this ready-to-go?

PR #4521, opened 2018-02-08
config: handle CRLF-only lines and BOM

### LINKED ISSUE #4517: libgit2 fails to read config files with empty lines with CRLF ending
### Reproduction steps

1. Have a `.gitconfig` file with CRLF line endings and at least one empty line (e.g. at the end)

### Expected behavior

Config file is parsed correctly (as Git for Windows does or older libgit2 versions such as e0568621535869451c91e915d28a69ac7c84b8b7).

### Actual behavior

libgit2 fails with the error: `failed to parse config file: Invalid configuration key (in C:/Users/Username/.gitconfig:5, column 0)`. (whereas line 5 is an empty line)

f7225946edeae35f48e3d402e2e0c94ea07f9666

Windows 7 x64

### Additional notes ###

This affects all files which are parsed as config files, such as `.gitmodules` (however, in that case `git_submodule_foreach` does not return any error, but fails silently and lists no submodules at all, the callback is not called once; this is maybe another bug in libgit2)...

Maybe this is related to compiler: I'm using MS VisualStudio 2017 15.5.5

Cross-ref: https://gitlab.com/tortoisegit/tortoisegit/issues/3153

pks-t:
Fixes #4517

### DIFF
--- a/src/config_parse.c
+++ b/src/config_parse.c
@@ -217,7 +217,7 @@ static int skip_bom(git_parse_ctx *parser)
 {
 	git_buf buf = GIT_BUF_INIT_CONST(parser->content, parser->content_len);
 	git_bom_t bom;
-	int bom_offset = git_buf_text_detect_bom(&bom, &buf, parser->content_len);
+	int bom_offset = git_buf_text_detect_bom(&bom, &buf, 0);
 
 	if (bom == GIT_BOM_UTF8)
 		git_parse_advance_chars(parser, bom_offset);
--- a/src/config_parse.c
+++ b/src/config_parse.c
@@ -475,6 +475,11 @@ int git_config_parse(
 		size_t line_len = parser->ctx.line_len;
 		char c;
 
+		/*
+		 * Get either first non-whitespace character or, if that does
+		 * not exist, the first whitespace character. This is required
+		 * to preserve whitespaces when writing back the file.
+		 */
 		if (git_parse_peek(&c, ctx, GIT_PARSE_PEEK_SKIP_WHITESPACE) < 0 &&
 		    git_parse_peek(&c, ctx, 0) < 0)
 			continue;
--- a/src/config_parse.c
+++ b/src/config_parse.c
@@ -495,6 +495,7 @@ int git_config_parse(
 			break;
 
 		case '\n': /* comment or whitespace-only */
+		case '\r':
 		case ' ':
 		case '\t':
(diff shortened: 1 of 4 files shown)


csware:
This fixes the issue.

pks-t:
Amended the PR to also fix handling of the BOM.

csware:
Works. Thanks!

src/config_parse.c
```diff
@@ -217,7 +217,7 @@ static int skip_bom(git_parse_ctx *parser)
 {
 	git_buf buf = GIT_BUF_INIT_CONST(parser->content, parser->content_len);
 	git_bom_t bom;
-	int bom_offset = git_buf_text_detect_bom(&bom, &buf, parser->content_len);
+	int bom_offset = git_buf_text_detect_bom(&bom, &buf, 0);
```
ethomson: I understand your confusion.  It appears that we never use `git_buf_text_detect_bom` to detect a BOM with a non-0 offset.

It's not clear to me why this was ever added in the first place - when would we try to detect a BOM in the midst of a buffer?

Whatever the reason, it sounds like we've refactored things to always look at the beginning of the buffer.  Should we drop this confusing third arg?

src/config_parse.c
```diff
@@ -475,6 +475,10 @@ int git_config_parse(
 		size_t line_len = parser->ctx.line_len;
 		char c;
 
+		/*
+		 * Get either first non-whitespace character or, if that does
+		 * not exist, the first whitespace character
```
tiennou: `because we need to preserve whitespace when writing back` ?

Maybe `We skip over whitespace until we find something meaningful, otherwise there's only whitespace and we need to preserve it when writing back`.

pks-t: True, that would certainly go a long way in explaining what happens here. Thanks, will amend!

src/config_parse.c
pks-t: I think so. I even wonder if there is any usecase with a mid-buffer-BOM, considering they are only allowed at the beginning of a file.

pks-t:
Amended. Added @tiennou's proposed improvement for the comment and removed the `offset` parameter of the BOM function as proposed by @ethomson. Thanks to both of you!

ethomson:
Thanks @csware for the report, @tiennou and @pks-t for the collaboration to get it fixed.  👍

Issue #4516, opened 2018-02-07
libgit2 fails to read config files with UTF8-BOM


csware:
### Reproduction steps

1. Create a .gitconfig with a UTF-8 BOM
2. Open an arbitrary repository (using `git_repository_open`)

### Expected behavior

Config file is parsed correctly (as Git for Windows does or older libgit2 versions such as e0568621535869451c91e915d28a69ac7c84b8b7).

### Actual behavior

libgit2 fails with the error: `failed to parse config file: Invalid configuration key (in C:/Users/Username/.gitconfig:1, column 0)`.

In debug mode I get the following assertion:

> ---------------------------
> Microsoft Visual C++ Runtime Library
> ---------------------------
> Debug Assertion Failed!
> 
> Program: D:\TortoiseGit\src\..\bin\Debug64\bin\TortoiseGitProc.exe
> File: minkernel\crts\ucrt\src\appcrt\convert\isctype.cpp
> Line: 42
> 
> Expression: c >= -1 && c <= 255

### Version of libgit2 (release number or SHA1)

f7225946edeae35f48e3d402e2e0c94ea07f9666

### Operating system(s) tested

Windows 7 x64

### Additional notes ###

This affects all files which are parsed as config files, such as `.gitmodules`...

Having UTF-8 encoded umlauts is not a problem.

Maybe this is related to compiler: I'm using MS VisualStudio 2017 15.5.5

#### Stacktrace: ####

```
ucrtbased.dll!_chvalidator_l(__crt_locale_pointers * locale=0x000000000021b150, int c=-17, int mask=263)
ucrtbased.dll!_isalnum_l(int c=-17, __crt_locale_pointers * locale=0x0000000000000000)
ucrtbased.dll!isalnum(int c=-17)
libgit2_tgit.dll!is_namechar(char c='ï')
libgit2_tgit.dll!parse_name(char * * name=0x000000000021b368, const char * * value=0x000000000021b268, git_config_parser * reader=0x000000000021b460, const char * line=0x0000000003e6c820)
libgit2_tgit.dll!parse_variable(git_config_parser * reader=0x000000000021b460, char * * var_name=0x000000000021b368, char * * var_value=0x000000000021b388)
[...]

PR #4522, opened 2018-02-08
Submodules-API should report .gitmodules parse errors instead of ignoring them

csware:
Issue mentioned in issue #4517 and issue #4516.

Git for Windows does report parse errors in `.gitmodules`, so should libgit2.

### DIFF
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -91,7 +91,7 @@ __KHASH_IMPL(
 
 static int submodule_alloc(git_submodule **out, git_repository *repo, const char *name);
 static git_config_backend *open_gitmodules(git_repository *repo, int gitmod);
-static git_config *gitmodules_snapshot(git_repository *repo);
+static int gitmodules_snapshot(git_config **snap, git_repository *repo);
 static int get_url_base(git_buf *url, git_repository *repo);
 static int lookup_head_remote_key(git_buf *remote_key, git_repository *repo);
 static int lookup_default_remote(git_remote **remote, git_repository *repo);
@@ -509,8 +509,11 @@ int git_submodule__map(git_repository *repo, git_strmap *map)
 		data.map = map;
 		data.repo = repo;
 
-		if ((mods = gitmodules_snapshot(repo)) == NULL)
+		if ((error = gitmodules_snapshot(&mods, repo)) < 0) {
+			if (error == GIT_ENOTFOUND)
+				error = 0;
 			goto cleanup;
+		}
 
 		data.mods = mods;
 		if ((error = git_config_foreach(
@@ -1511,7 +1514,8 @@ int git_submodule_reload(git_submodule *sm, int force)
 
 	if (!git_repository_is_bare(sm->repo)) {
 		/* refresh config data */
-		mods = gitmodules_snapshot(sm->repo);
+		if ((error = gitmodules_snapshot(&mods, sm->repo)) < 0 && error != GIT_ENOTFOUND)
+			return error;
 		if (mods != NULL) {
 			error = submodule_read_config(sm, mods);
 			git_config_free(mods);
@@ -1915,32 +1919,37 @@ static int submodule_load_from_wd_lite(git_submodule *sm)
 }
 
 /**
- * Returns a snapshot of $WORK_TREE/.gitmodules.
+ * Requests a snapshot of $WORK_TREE/.gitmodules.
  *
- * We ignore any errors and just pretend the file isn't there.
+ * Returns GIT_ENOTFOUND in case no .gitmodules file exist
  */
-static git_config *gitmodules_snapshot(git_repository *repo)
+static int gitmodules_snapshot(git_config **snap, git_repository *repo)
 {
 	const char *workdir = git_repository_workdir(repo);
-	git_config *mods = NULL, *snap = NULL;
+	git_config *mods = NULL;
 	git_buf path = GIT_BUF_INIT;
+	int error;
 
-	if (workdir != NULL) {
-		if (git_buf_joinpath(&path, workdir, GIT_MODULES_FILE) != 0)
-			return NULL;
+	if (!workdir)
+		return GIT_ENOTFOUND;
 
-		if (git_config_open_ondisk(&mods, path.ptr) < 0)
-			mods = NULL;
-	}
+	if ((error = git_buf_joinpath(&path, workdir, GIT_MODULES_FILE)) < 0)
+		return error;
 
+	if ((error = git_config_open_ondisk(&mods, path.ptr)) < 0)
+		goto cleanup;
+
+	if ((error = git_config_snapshot(snap, mods)) < 0)
+		goto cleanup;
+
+	error = 0;
+
+cleanup:
+	if (mods)
+		git_config_free(mods);
 	git_buf_free(&path);
 
-	if (mods) {
-		git_config_snapshot(&snap, mods);
-		git_config_free(mods);
-	}
-
-	return snap;
+	return error;
 }
 
 static git_config_backend *open_gitmodules(
--- a/tests/submodule/lookup.c
+++ b/tests/submodule/lookup.c
@@ -445,3 +445,19 @@ void test_submodule_lookup__foreach_in_bare_repository_fails(void)
 
 	cl_git_fail(git_submodule_foreach(g_repo, foreach_cb, NULL));
 }
+
+void test_submodule_lookup__fail_invalid_gitmodules(void)
+{
+	git_submodule *sm;
+	sm_lookup_data data;
+	memset(&data, 0, sizeof(data));
+
+	cl_git_rewritefile("submod2/.gitmodules",
+			   "[submodule \"Test_App\"\n"
+			   "    path = Test_App\n"
+			   "    url = ../Test_App\n");
+
+	cl_git_fail(git_submodule_lookup(&sm, g_repo, "Test_App"));
+
+	cl_git_fail(git_submodule_foreach(g_repo, sm_lookup_cb, &data));
+}


src/submodule.c
```diff
@@ -91,7 +91,7 @@ __KHASH_IMPL(
 
 static int submodule_alloc(git_submodule **out, git_repository *repo, const char *name);
 static git_config_backend *open_gitmodules(git_repository *repo, int gitmod);
-static git_config *gitmodules_snapshot(git_repository *repo);
+static int gitmodules_snapshot(git_repository *repo, git_config **snap);
```
pks-t: The `out` parameter is typically first

src/submodule.c
```diff
@@ -509,7 +509,7 @@ int git_submodule__map(git_repository *repo, git_strmap *map)
 		data.map = map;
 		data.repo = repo;
 
-		if ((mods = gitmodules_snapshot(repo)) == NULL)
+		if ((error = gitmodules_snapshot(repo, &mods)) < 0 || mods == NULL)
```
pks-t: I feel like this interface has become a bit more brittle by having to check for `mods == NULL`, as well. What about returning for example `ENOTFOUND` in case there are no submodules?

src/submodule.c
```diff
@@ -1916,31 +1917,33 @@ static int submodule_load_from_wd_lite(git_submodule *sm)
...
+		git_buf path = GIT_BUF_INIT;
+		git_config *mods = NULL;
 		if (git_buf_joinpath(&path, workdir, GIT_MODULES_FILE) != 0)
-			return NULL;
+			return 0;
 
-		if (git_config_open_ondisk(&mods, path.ptr) < 0)
-			mods = NULL;
-	}
+		int error = git_config_open_ondisk(&mods, path.ptr);
```
pks-t: Might be I'm misreading the diff, but isn't this introducing a declaration after instructions?

tests/submodule/lookup.c
```diff
@@ -445,3 +445,20 @@ void test_submodule_lookup__foreach_in_bare_repository_fails(void)
...
+void test_submodule_lookup__fail_invalid_gitignores(void)
+{
+	git_submodule *sm;
+	sm_lookup_data data;
+	memset(&data, 0, sizeof(data));
+
+	cl_git_rewritefile("submod2/.gitmodules",
+			   "[submodule \"Test_App\"\n"
+			   "    path = Test_App\n"
+			   "    url = ../Test_App\n");
```
pks-t: :+1:

tests/submodule/lookup.c
```diff
@@ -445,3 +445,20 @@ void test_submodule_lookup__foreach_in_bare_repository_fails(void)
 
 	cl_git_fail(git_submodule_foreach(g_repo, foreach_cb, NULL));
 }
+
+void test_submodule_lookup__fail_invalid_gitignores(void)
```
pks-t: `gitignores`? Don't you mean `gitmodules`?

csware: Yeah a type, this is still WIP as I can't test locally ATM...

src/submodule.c
csware: I haven't done that in the first place as I then have to reset errors... First lemme see whether this approach will work at all...

csware:
@pks-t All issues should be addressed now.

csware:
@pks-t All change requests should be addressed now.

src/submodule.c
```diff
@@ -509,8 +509,12 @@ int git_submodule__map(git_repository *repo, git_strmap *map)
 		data.map = map;
 		data.repo = repo;
 
-		if ((mods = gitmodules_snapshot(repo)) == NULL)
+		if ((error = gitmodules_snapshot(&mods, repo)) < 0)
+		{
```
pks-t: The opening "{" is on the same line for single-line if-statements.

src/submodule.c
```diff
@@ -509,8 +509,12 @@ int git_submodule__map(git_repository *repo, git_strmap *map)
...
+			if (error == GIT_ENOTFOUND)
+				error = 0;
```
pks-t: In my opinion, it looks much more readable like this

src/submodule.c
```diff
@@ -1916,31 +1921,32 @@ static int submodule_load_from_wd_lite(git_submodule *sm)
 
 /**
  * Returns a snapshot of $WORK_TREE/.gitmodules.
```
pks-t: I'd like to see a short comment about `GIT_ENOTFOUND` here such that people know to expect it in case there are no submodules.

src/submodule.c
```diff
@@ -1916,31 +1921,32 @@ static int submodule_load_from_wd_lite(git_submodule *sm)
...
-	git_config *mods = NULL, *snap = NULL;
-	git_buf path = GIT_BUF_INIT;
 
 	if (workdir != NULL) {
+		int error;
+		git_buf path = GIT_BUF_INIT;
+		git_config *mods = NULL;
 		if (git_buf_joinpath(&path, workdir, GIT_MODULES_FILE) != 0)
-			return NULL;
+			return 0;
```
pks-t: Hum. Shouldn't this bubble up the error? `git_buf_joinpath` failing indicates that there was a memory allocation error, which definitly shouldn't be ignored.

```
if (git_buf_joinpath(&path, workdir, GIT_MODULES_FILE) < 0)
    return -1;
```

src/submodule.c
```diff
@@ -1916,31 +1921,32 @@ static int submodule_load_from_wd_lite(git_submodule *sm)
...
- * We ignore any errors and just pretend the file isn't there.
  */
-static git_config *gitmodules_snapshot(git_repository *repo)
+static int gitmodules_snapshot(git_config **snap, git_repository *repo)
 {
 	const char *workdir = git_repository_workdir(repo);
-	git_config *mods = NULL, *snap = NULL;
-	git_buf path = GIT_BUF_INIT;
 
 	if (workdir != NULL) {
```
pks-t: The whole indentation could be avoided if you return early in case `workdir == NULL`. That's a matter of taste, though -- your call if you want to do that instead.

csware: Yeah I also thought that, however, I did not want to make too big changes...

csware:
@pks-t All change requests should be addressed now.

src/submodule.c
```diff
@@ -1915,32 +1919,34 @@ static int submodule_load_from_wd_lite(git_submodule *sm)
...
+	if (git_buf_joinpath(&path, workdir, GIT_MODULES_FILE) != 0)
+		return -1;
+
+	error = git_config_open_ondisk(&mods, path.ptr);
 
 	git_buf_free(&path);
 
 	if (mods) {
-		git_config_snapshot(&snap, mods);
+		git_config_snapshot(snap, mods);
```
pks-t: Sorry for not catching that earlier, but now that you report errors you should definitly also check the return value of `git_config_snapshot`.

src/submodule.c
```diff
@@ -1915,32 +1919,34 @@ static int submodule_load_from_wd_lite(git_submodule *sm)
...
 		git_config_free(mods);
+		return 0;
```
pks-t: We usually try to avoid returning early for non-trivial code. I'd rather go with something like

```
...
*snap = NULL;
...

if ((error = git_config_open_ondisk(...)) < 0)
    goto out;
if (!mods) {
    error = -1;
    goto out;
}

if ((error = git_config_snapshot(mods, config)) < 0)
    goto out;

out:
    git_config_free(mods);
    if (error)
        git_config_free(*snap);
    return error;
```

csware: `goto` is better? - Shall I change all the code for this?

pks-t: Sorry :sweat_smile: Seeing that your code already improves the function by quite a lot, I'm satisfied if you check the return code of `git_config_snapshot`, without converting the function to `goto` style.

csware:
@pks-t Any reason why lots of PRs are not merged (such as this one)?

ethomson:
@csware because we’re busy.  We appreciate your contributions but we are all volunteers and do not always have as much time to review pull requests as we would like.

We appreciate your patience.

ethomson:
Also, we’ve locked down for 0.27.  This is eligible for merging to master _after_ that release, but this is not a change we’d take before then.

src/submodule.c
```diff
@@ -1915,32 +1919,34 @@ static int submodule_load_from_wd_lite(git_submodule *sm)
...
-		if (git_buf_joinpath(&path, workdir, GIT_MODULES_FILE) != 0)
-			return NULL;
+	if (!workdir)
+		return GIT_ENOTFOUND;
 
-		if (git_config_open_ondisk(&mods, path.ptr) < 0)
-			mods = NULL;
-	}
+	if (git_buf_joinpath(&path, workdir, GIT_MODULES_FILE) != 0)
+		return -1;
```
pks-t: This should probably be `if ((error = git_buf_joinpath(&path, workdir, GIT_MODULES_FILE)) < 0) return error;`

src/submodule.c
```diff
@@ -1915,32 +1919,34 @@ static int submodule_load_from_wd_lite(git_submodule *sm)
...
+
+	error = git_config_open_ondisk(&mods, path.ptr);
```
pks-t: You should probably also check the return value here. While an error will cause `mods` to be `NULL`, which is catched in a few lines, it feels a bit more robust to check directly. You could avoid multiple free's by adding a `goto out`. Additional kudos if you also convert the other error handling branches to `goto out` with a unified section for freeing up memory in case `error != 0`.

tests/submodule/lookup.c
```diff
@@ -445,3 +445,20 @@ void test_submodule_lookup__foreach_in_bare_repository_fails(void)
...
+	cl_git_rewritefile("submod2/.gitmodules",
+			   "[submodule \"Test_App\"\n"
+			   "    path = Test_App\n"
+			   "    url = ../Test_App\n");
+
+	cl_git_fail(git_submodule_lookup(&sm, g_repo, "Test_App"));
+
+	cl_git_fail(git_submodule_foreach(g_repo, sm_lookup_cb, &data));
+}
+
```
pks-t: There's a trailing newline here

ethomson:
Thanks for the contribution, @csware!

PR #4523, opened 2018-02-08
Test library init/shutdown separately

tiennou:
As it says on the tin. The goal is to see how we fare when the library goes through multiple init/shutdown cycles.

### DIFF
--- a/tests/clar.c
+++ b/tests/clar.c
@@ -136,6 +136,8 @@ static struct {
 	cl_trace_cb *pfn_trace_cb;
 	void *trace_payload;
 
+	int (*global_init)(void);
+	int (*global_cleanup)(void);
 } _clar;
 
 struct clar_func {
@@ -261,6 +263,13 @@ clar_run_suite(const struct clar_suite *suite, const char *filter)
 	if (!_clar.report_errors_only)
 		clar_print_onsuite(suite->name, ++_clar.suites_ran);
 
+
+	if (_clar.global_init) {
+		int ret = _clar.global_init();
+		if (ret != 0)
+			exit(-1);
+	}
+
 	_clar.active_suite = suite->name;
 	_clar.active_test = NULL;
 	CL_TRACE(CL_TRACE__SUITE_BEGIN);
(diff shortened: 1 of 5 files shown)


tiennou:
Since clar modifications are supposed to go upstream (when possible), I'll kindly ask for @vmg's review of the clar parts.

pks-t:
Just to let you know: we already have the tests that you propose in "core::init".

tiennou:
Almost ? I'm looking at what happens after "something" really happens (like, we do some cloning, which might involve TLS, which might have some global object). If I can get the suite to got past a double-free in "core::useragent".

pks-t:
Oh, I see. Shouldn't have jumped to conclusions from the PR's description only ;)

So what you are doing here is executing `git_libgit2_init`/`git_libgit2_shutdown` after each test suite. Two questions which were popping into my mind: What is the performance impact of this and why is it being executed after each suite and not after each test? Performance reasons?

tiennou:
Yep, sorry for the almost-scription.

From a performance POV it didn't seem to add that much locally, but I'll try to provide hard data if you want (as soon as I can get a complete test run 😂). Right now local/Travis segfaults in `core::useragent`, and AppVeyor is unhappy for different reasons.

I chose to go the "setup/teardown for each suite" because that was a little more specific, without being too frequent, but it could be changed easily if you want.

ethomson:
Curious about the impetus behind these tests.  I don't think that we've ever said that you could `git_libgit2_init` after you've shutdown.  We're jumping into testing that its successful without thinking about what it means to make this guarantee and whether it's a desirable thing for the library to enable.

tiennou:
Well, I wasn't aware of `init/shutdown` being a process-lived thingy, so I can understand why this is surprising… The rationale was making sure we behaved correctly w.r.t memory-management and the library lifecycle under various workloads (various workloads being simulated individual test runs).

ethomson:
OK, cool.  I'd be curious to see where these test failures are - the user agent thing seems like an obvious double-free that we can and should fix.  So I'm hopeful that all of the issues are like this.  But if there are hard problems that stem from our assumptions about lifecycles, then I think we should consider how to proceed.

PR #4525, opened 2018-02-09
Configuration entry iteration in order

pks-t:
I've finally pulled myself together and tackled my itch of iterating through configuration entries in the order they appear in the file. At first, I started getting annoyed again by the unnecessarily complex config file code, so I preceded the actual logic by quite a lot of refactoring and renaming of structures and functions. The result is actually very pleasant: while maintaining a second data structure alongside the map of entries, this PR still has more deletions than additions. So I definitly think that this PR helps a whole lot regarding readability.

### DIFF
--- a/include/git2/config.h
+++ b/include/git2/config.h
@@ -64,6 +64,7 @@ typedef enum {
 typedef struct git_config_entry {
 	const char *name; /**< Name of the entry (normalised) */
 	const char *value; /**< String value of the entry */
+	unsigned int include_depth; /**< Depth of includes where this variable was found */
 	git_config_level_t level; /**< Which config file this was found in */
 	void (*free)(struct git_config_entry *entry); /**< Free function for this entry */
 	void *payload; /**< Opaque value for the free function. Do not read or write */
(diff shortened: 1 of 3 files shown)


pks-t:
Small note: I highly recommend reviewing the commits one by one. There's a lot of renaming going on step by step, which becomes a whole lot easier to do if not reviewing the complete diff at once.

ethomson:
Are you hoping that we land this for the immediate next release?  Seems like a Big Change and my preference would be to land it after we cut the next release, but I want to make sure we're both on the same page.

pks-t:
We are on the same page ;)

pks-t:
Rebased to fix merge conflicts with #4552

ethomson:
> Small note: I highly recommend reviewing the commits one by one. There's a lot of renaming going on step by step, which becomes a whole lot easier to do if not reviewing the complete diff at once.

This was a good hint.  Reviewing each commit was helpful in seeing the progression.

Thanks for tackling this; there's some nice cleanup in here _and_ improved functionality.  👍

PR #4527, opened 2018-02-09
Plug resource leaks

pks-t:


pks-t:
Cc @ethomson

ethomson:
Thanks!  👍

PR #4529, opened 2018-02-11
git_index_add_frombuffer: only accept files/links

ethomson:
Ensure that the buffer given to `git_index_add_frombuffer` represents a regular blob, an executable blob, or a link.  Explicitly reject commit entries (submodules) - it makes little sense to allow users to add a submodule from a string; there's no possible path to success.


ethomson:
It seems only minimally valuable, but it's also very easy, so I went ahead and added a test.  Thanks, @pks-t.

Issue #4495, opened 2018-01-22
Documentation for `git_mempack_dump` missing


medranocalvo:
Documentation for function `git_mempack_dump` is missing from https://libgit2.github.com/libgit2/#HEAD/group/mempack.

### Version of libgit2 (release number or SHA1)
0.26.0

medranocalvo:
@tiennou Thank you very much!

Issue #4492, opened 2018-01-22
`git2/sys/mempack.h` fails to include `git2/buffer.h`


medranocalvo:
Header file `git2/sys/mempack.h` should include `git2/buffer.h`:
~~~
/usr/local/include/git2/sys/mempack.h:67:22: error: unknown type name 'git_buf'
~~~

### Version of libgit2 (release number or SHA1)
0.26.0

PR #4530, opened 2018-02-12
Fix docurium missing includes

### LINKED ISSUE #4213: does git_stash_save() exists? Or how to perform "git stash" in libgit2
Is there some function like git_stash_save() in libgit2? Or how can I perform "git stash" in libgit2 ? 
I searched in libgit2 API but all I found was functions to pop or drop stashes from stash list, but how do I fill that list?

https://libgit2.github.com/libgit2/#HEAD/group/stash

tiennou:
The first commit makes `git_oid` & `git_buf` "globally" visible (ie. less likely to be forgotten) by including them from `types.h`. If that's not really appropriate, it could be fixed more precisely by going through each error from [this script](https://gist.github.com/tiennou/fcdf9967894c5671c8c5c716405de52d).

Apart from a typechange (see `worktree:` commit), the rest is minor fixups + missing documentation.

Fixes #4213, #4495, maybe #4492.

### DIFF
--- a/include/git2/attr.h
+++ b/include/git2/attr.h
@@ -196,7 +196,7 @@ GIT_EXTERN(int) git_attr_get_many(
  * @see git_attr_foreach.
  *
  * @param name The attribute name.
- * @param value The attribute value. May be NULL is the attribute is explicitly
+ * @param value The attribute value. May be NULL if the attribute is explicitly
  *              set to UNSPECIFIED using the '!' sign.
  * @param payload A user-specified pointer.
  * @return 0 to continue looping, non-zero to stop. This value will be returned
--- a/include/git2/attr.h
+++ b/include/git2/attr.h
@@ -186,6 +186,22 @@ GIT_EXTERN(int) git_attr_get_many(
 	size_t num_attr,
 	const char **names);
 
+/**
+ * The callback used with git_attr_foreach.
+ *
+ * This callback will be invoked only once per attribute name, even if there
+ * are multiple rules for a given file. The highest priority rule will be
+ * used.
+ *
+ * @see git_attr_foreach.
+ *
+ * @param name The attribute name.
+ * @param value The attribute value. May be NULL is the attribute is explicitly
+ *              set to UNSPECIFIED using the '!' sign.
+ * @param payload A user-specified pointer.
+ * @return 0 to continue looping, non-zero to stop. This value will be returned
+ *         from git_attr_foreach.
+ */
 typedef int (*git_attr_foreach_cb)(const char *name, const char *value, void *payload);
 
 /**
@@ -196,13 +212,8 @@ typedef int (*git_attr_foreach_cb)(const char *name, const char *value, void *pa
  * @param path Path inside the repo to check attributes.  This does not have
  *             to exist, but if it does not, then it will be treated as a
  *             plain file (i.e. not a directory).
- * @param callback Function to invoke on each attribute name and value.  The
- *             value may be NULL is the attribute is explicitly set to
- *             UNSPECIFIED using the '!' sign.  Callback will be invoked
- *             only once per attribute name, even if there are multiple
- *             rules for a given file.  The highest priority rule will be
- *             used.  Return a non-zero value from this to stop looping.
- *             The value will be returned from `git_attr_foreach`.
+ * @param callback Function to invoke on each attribute name and value.
+ *                 See git_attr_foreach_cb.
  * @param payload Passed on as extra parameter to callback function.
  * @return 0 on success, non-zero callback return value, or error code
  */
--- a/include/git2/blame.h
+++ b/include/git2/blame.h
@@ -128,7 +128,7 @@ typedef struct git_blame_hunk {
 } git_blame_hunk;
 
 
-/* Opaque structure to hold blame results */
+/** Opaque structure to hold blame results */
 typedef struct git_blame git_blame;
 
 /**
--- a/include/git2/blame.h
+++ b/include/git2/blame.h
@@ -48,9 +48,8 @@ typedef enum {
 /**
  * Blame options structure
  *
- * Use zeros to indicate default settings.  It's easiest to use the
- * `GIT_BLAME_OPTIONS_INIT` macro:
- *     git_blame_options opts = GIT_BLAME_OPTIONS_INIT;
+ * Initialize with `GIT_BLAME_OPTIONS_INIT`. Alternatively, you can
+ * use `git_blame_init_options`.
  *
  */
 typedef struct git_blame_options {
--- a/include/git2/blame.h
+++ b/include/git2/blame.h
@@ -67,15 +67,18 @@ typedef struct git_blame_options {
 	uint16_t min_match_characters;
 	/** The id of the newest commit to consider. The default is HEAD. */
 	git_oid newest_commit;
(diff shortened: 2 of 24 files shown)


tiennou:
I've pushed my work-in-progress to my own gh-pages branch so it's available [here](http://tiennou.github.io/libgit2/#HEAD). Note that it requires tweaks to docurium which aren't upstream (yet, there's some in a PR, but now there's more 😉).

Highlights :
- the `transaction` section makes its first appearance.
- I've disabled a check for debugging reasons which makes every valid identifier at least appear (hence the `imaxdiv` section, which comes from `git2/stdint.h`, and surely a bunch of others). Still not sure whether we want that or not (the rationale for the change is warnings about undocumented things when docurium-ing)
- some types lost their documentation (eg. [git_odb_backend](http://tiennou.github.io/libgit2/#HEAD/type/git_odb_backend)). IIUC, it stems from parsing the "quick" comment from `git2/types.h` before the real struct def in `git2/sys/odb_backend.h`. Still investigating.

include/git2/worktree.h
```diff
@@ -74,6 +74,14 @@ GIT_EXTERN(void) git_worktree_free(git_worktree *wt);
  */
 GIT_EXTERN(int) git_worktree_validate(const git_worktree *wt);
 
+/**
+ * Worktree add options structure
+ *
+ * Zero out for defaults.  Initialize with `GIT_WORKTREE_ADD_OPTIONS_INIT`
```
pks-t: We shouldn't recommend zeroing out the struct, but only using `GIT_WORKTREE_ADD_OPTIONS_INIT` or the corresponding function, I guess.

include/git2/worktree.h
```diff
@@ -161,6 +169,14 @@ typedef enum {
 	GIT_WORKTREE_PRUNE_WORKING_TREE = 1u << 2,
 } git_worktree_prune_t;
 
+/**
+ * Worktree prune options structure
+ *
+ * Zero out for defaults.  Initialize with `GIT_WORKTREE_PRUNE_OPTIONS_INIT`
```
pks-t: Same here wrt zeroing out the struct.

include/git2/worktree.h
```diff
@@ -180,7 +180,7 @@ typedef enum {
 typedef struct git_worktree_prune_options {
 	unsigned int version;
 
-	uint32_t flags;
+	git_worktree_prune_t flags;
```
pks-t: This is not correct, though. An enum should always have a valid enum value, only, and not be used as a field of bits. As we're ORing multiple flags into this field this has to bee an uint32.

include/git2/worktree.h
tiennou: IIRC this is copy-pasted from `git_checkout_options`. So what about something like :
```
/**
 * $STRUCT_DESCRIPTION
 *
 * Initialize with `$STRUCT_INIT_MACRO`. Alternatively, you can
 * use `$STRUCT_INIT_FN`.
 * 
 * @see $STRUCT_USERS
 */
```
and I take a look to make sure all conform to this style ?

include/git2/worktree.h
tiennou: Right, I missed the ORing 🙄. Will drop from patchset. I was trying to get nice a link to show up, so I'll make it a doc-comment instead.

include/git2/attr.h
```diff
@@ -186,6 +186,22 @@ GIT_EXTERN(int) git_attr_get_many(
...
 	const char **names);
 
+/**
+ * The callback used with git_attr_foreach.
+ *
+ * This callback will be invoked only once per attribute name, even if there
+ * are multiple rules for a given file. The highest priority rule will be
+ * used.
+ *
+ * @see git_attr_foreach.
```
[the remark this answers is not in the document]
tiennou: Well, I think it reads better, but I checked and the [usage](https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#see) is clear. I'll change it, though it seems docurium [doesn't](http://tiennou.github.io/libgit2/#HEAD/group/callback/git_attr_foreach_cb) know about `@see` anyways…

include/git2/worktree.h
pks-t: I expected that this was copied from somewhere else, but was too lazy to double-check :P Your proposal would certainly be fine with me, and removing references to zeroing out the structure are definitly welcome.

include/git2/attr.h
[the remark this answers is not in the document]
pks-t: We do have some special syntax for referring to other functions, I just always forget about it... and right now I cannot find any examples for it, either. Maybe prefixing it with an `@`? No idea...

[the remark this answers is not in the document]
tiennou: I haven't been able to pinpoint a specific syntax in the docurium code, or any specific behavior w.r.t javadoc comments for that matter. As far as I can see, it checks function arguments against the list of known types and build links that way, but it only does this on functions. I've tried extending it to struct fields to no avail.

include/git2/transaction.h
```diff
@@ -8,6 +8,7 @@
 #define INCLUDE_git_transaction_h__
 
 #include "common.h"
+#include "types.h"
```
pks-t: Hum. Good enough for now, I'd say. I feel like we should just include "types.h" in "common.h", as nearly every header wants "types.h" to be included. Can be done in a separate PR, though

include/git2/attr.h
```diff
@@ -186,6 +186,22 @@ GIT_EXTERN(int) git_attr_get_many(
...
+ *
+ * @param name The attribute name.
+ * @param value The attribute value. May be NULL is the attribute is explicitly
```
pks-t: Typo: "s/is/if"

include/git2/transaction.h
tiennou: 👍 on that. I fear the next time an header is added, it will be forgotten and the clang-parse phase of docurium will get incomplete results.

include/git2/blame.h
```diff
@@ -52,41 +52,46 @@ typedef enum {
...
+	/** The id of the oldest commit to consider.
+	 * The default is the first commit encountered with a NULL parent.
+	 */
 	git_oid oldest_commit;
+	/** The first line in the file to blame.
+	 * The default is 1 (line numbers start with 1).
+	 */
 	size_t min_line;
+	/** The last line in the file to blame.
+	 * The default is the last line of the file.
```
pks-t: Quick drive-by comment as I accidentally stumbled over these lines: the style should rather be

```
/**
 * Foobar
 */
```

include/git2/worktree.h
ethomson: Agreed with @pks-t, and I like this text you've provided.

include/git2/describe.h
```diff
@@ -70,6 +69,16 @@ typedef struct git_describe_options {
 	GIT_DESCRIBE_DEFAULT_MAX_CANDIDATES_TAGS, \
 }
 
+/**
+ * Initialize git_describe_options structure
```
pks-t: This one and `git_describe_init_format_options` both miss the leading short doc string that you're consistently using for the other init_options functions now

tiennou: Fixed-up !

pks-t:
Thanks @tiennou for handling this!

PR #4531, opened 2018-02-12
checkout: change default strategy to SAFE

### LINKED ISSUE #4200: Change checkout options default strategy to safe
The default checkout strategy of dry run seems confusing. It's a common theme on stack overflow questions. Would it make sense to change the default to safe? It would break people who rely on the default, but does anybody really use that in practice?

tiennou:
Fixes #4200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ v0.27 + 1
 
 ### Breaking API changes
 
+* The default checkout strategy changed from `DRY_RUN` to `SAFE` (#4531).
+
 v0.27
 ---------
 
--- a/include/git2/checkout.h
+++ b/include/git2/checkout.h
@@ -251,7 +251,7 @@ typedef void (*git_checkout_perfdata_cb)(
 typedef struct git_checkout_options {
 	unsigned int version;
 
-	unsigned int checkout_strategy; /**< default will be a dry run */
+	unsigned int checkout_strategy; /**< default will be a safe checkout */
 
 	int disable_filters;    /**< don't apply filters like CRLF conversion */
 	unsigned int dir_mode;  /**< default is 0755 */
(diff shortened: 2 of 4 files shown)


pks-t:
Thanks for this PR, @tiennou. Could you please add an entry to our changelog so that people will (hopefully) notice the change?

ethomson:
There are a couple of places that take a checkout options structure and then upgrade the defaults to `SAFE`.  This was an annoying workaround to avoid having to make consumers of those functions deal with this already esoteric checkout issue.  For example, `rebase` upgrades the checkout options if neither `SAFE` nor `FORCE` is specified.  (Ugh.)

Those can (and should) go away now -- which would allow consumers to actually provide a checkout strategy that does a dry run.  At the moment, there's no way for them to do that.

That doesn't have to be part of this PR, but if not, please open an issue and assign it to me to investigate.

tiennou:
Fixed up the CHANGELOG, the documentation (heh), as well as the uses I found where the fix was clear.

I've left [this](https://github.com/libgit2/libgit2/blob/master/src/merge.c#L2867) (called from [there](https://github.com/libgit2/libgit2/blob/master/src/merge.c#L3285)) because I don't think it's related, but someone with a more complete grasp of the merge code should take a look. My understanding was "we're doing weird stuff with the indexwriter and not-writing-indexes so we're stashing the user-given strategy somewhere, munging the options one, and then use both in different codepaths".

pks-t:
I think the untouched region is fine:

```
	unsigned int checkout_strategy;
	...
	checkout_strategy = given_checkout_opts ?
		given_checkout_opts->checkout_strategy :
GIT_CHECKOUT_SAFE;
	...
	if ((error = merge_normalize_checkout_opts(&checkout_opts, repo,
			given_checkout_opts, checkout_strategy,
			base, our_head, their_heads, their_heads_len)) < 0 ||
		(error = git_checkout_index(repo, index, &checkout_opts)) < 0)
		goto done;
```

So it will use the user-supplied checkout strategy and fall back to `GIT_CHECKOUT_SAFE`, which is exactly what we want to do here.

pks-t:
I'm fine with the PR. Will not approve it though as I'm not as familiar with the code as @ethomson

ethomson:
Awesome, thanks @tiennou!

PR #4535, opened 2018-02-18
checkout: when examining index (instead of workdir), also examine mode

ethomson:
When checking out a file, we determine whether the baseline (what we expect to be in the working directory) actually matches the contents of the working directory.  This is safe behavior to prevent us from overwriting changes in the working directory.

We look at the index to optimize this test: if we know that the index matches the working directory, then we can simply look at the index data compared to the baseline.

We have historically compared the baseline to the index entry by oid.  However, we must also compare the mode of the two items to ensure that they are identical.  Otherwise, we will refuse to update the working directory for a mode change.

Update this behavior and include a test to validate the behavior.

### DIFF
--- a/tests/checkout/head.c
+++ b/tests/checkout/head.c
@@ -156,3 +156,28 @@ void test_checkout_head__typechange_workdir(void)
 
 	git_object_free(target);
 }
+
+void test_checkout_head__typechange_index_and_workdir(void)
+{
+	git_checkout_options opts = GIT_CHECKOUT_OPTIONS_INIT;
+	git_object *target;
+	git_index *index;
+	struct stat st;
+
+	opts.checkout_strategy = GIT_CHECKOUT_FORCE;
+
+	cl_git_pass(git_revparse_single(&target, g_repo, "HEAD"));
+	cl_git_pass(git_reset(g_repo, target, GIT_RESET_HARD, NULL));
+
+	cl_must_pass(p_chmod("testrepo/new.txt", 0755));
+	cl_git_pass(git_repository_index(&index, g_repo));
+	cl_git_pass(git_index_add_bypath(index, "new.txt"));
+	cl_git_pass(git_index_write(index));
+	cl_git_pass(git_checkout_head(g_repo, &opts));
+
+	cl_git_pass(p_stat("testrepo/new.txt", &st));
+	cl_assert(!GIT_PERMS_IS_EXEC(st.st_mode));
+
+	git_object_free(target);
+	git_index_free(index);
+}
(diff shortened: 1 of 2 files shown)


ethomson:
This is a small and hopefully uncontroversial slice of #4536.  I'm opening this separately only because it seems like a simple and obvious fix that we may want to include before the next release.

pks-t:
While this seems to work alright on Travis/AppVeyor, it causes "index::filemodes::frombuffer_requires_files" to fail for me:

```
  1) Failure:
index::filemodes::frombuffer_requires_files [/home/pks/Development/libgit2/tests/index/filemodes.c:276]
  Expression is not true: (ret_entry = git_index_get_bypath(index, "dummy-file.txt", 0))
```

pks-t:
Ugh, no. This is obviously because of #4529.

tests/checkout/head.c
```diff
@@ -136,3 +136,34 @@ void test_checkout_head__do_remove_tracked_subdir(void)
...
+	cl_must_pass(p_chmod("testrepo/new.txt", 0755));
+	cl_git_pass(git_checkout_head(g_repo, &opts));
+
+	cl_git_pass(p_stat("testrepo/new.txt", &st));
+	cl_assert(!GIT_PERMS_IS_EXEC(st.st_mode));
+
+	cl_must_pass(p_chmod("testrepo/new.txt", 0755));
+	cl_git_pass(git_repository_index(&index, g_repo));
+	cl_git_pass(git_index_add_bypath(index, "new.txt"));
+	git_index_write(index);
```
pks-t: This line is missing an `cl_git_pass`.

tests/checkout/head.c
```diff
@@ -136,3 +136,34 @@ void test_checkout_head__do_remove_tracked_subdir(void)
...
+	git_index_free(index);
```
pks-t: I'd prefer to move that line to the end of the test

tests/checkout/head.c
```diff
@@ -136,3 +136,34 @@ void test_checkout_head__do_remove_tracked_subdir(void)
...
+	cl_git_pass(git_checkout_head(g_repo, &opts));
+
+	cl_git_pass(p_stat("testrepo/new.txt", &st));
+	cl_assert(!GIT_PERMS_IS_EXEC(st.st_mode));
```
pks-t: As the second test is concerned about a different thing (index vs workdir), I'd move it into a separate test.

ethomson:
I've fixed up the test per your requests.

tests/checkout/head.c
```diff
@@ -136,3 +136,46 @@ void test_checkout_head__do_remove_tracked_subdir(void)
 	cl_assert(!git_path_isfile("testrepo/subdir/tracked-file"));
 	cl_assert(git_path_isfile("testrepo/subdir/untracked-file"));
 }
+
+void test_checkout_head__typechange_workdir(void)
+{
+	git_checkout_options opts = GIT_CHECKOUT_OPTIONS_INIT;
+	git_object *target;
```
pks-t: This variable is leaking now

ethomson:
Agreed, I added the backport label.

PR #4536, opened 2018-02-19
Add a "dirty" state to the index when it has unsaved changes

ethomson:
Add a new "dirty" state to the index when it has unsaved changes.

At present, you can load the index and make changes to it without saving those changes.  However various functions in the library behave different when the repository's index has unsaved changes.  For example, checkout will reload the index (by default), but merge will not.

This asymmetry is [obviously a problem](https://github.com/libgit2/libgit2/issues/4203).

It was noted during [an attempt to standardize this behavior](https://github.com/libgit2/libgit2/pull/4407) that some people may be relying on the prior behavior that merge does not reload the index.

The discussion from this was that we should add a warning in the unlikely event that someone does rely on this behavior.  This PR adds a "dirty" state to the PR - now if a consumer tries to act on the index in a way that changes would be discarded, they are warned.  As a result, consumers must explicitly either commit their changes (using `git_index_write`) or discard their changes (through any of a variety of mechanisms, including `git_index_read`) before executing a function that will otherwise mutate the index (eg `git_checkout`, `git_merge`, etc).

The first few commits in this PR update tests that are incorrect regardless of the changes to add these restrictions and include a "dirty" state to the index.  The remainder add this change and update several tests to explicitly commit the index before further mutating it.

This should (finally) pave the way for #4407.


ethomson:
Rebased.

ethomson:
Any thoughts on this before I 🚢 or is it self-evident?

src/index.c
```diff
@@ -668,6 +674,9 @@ int git_index_read(git_index *index, int force)
 	if (!error)
 		git_futils_filestamp_set(&index->stamp, &stamp);
 
+	if (!error)
```
tiennou: Maybe merge this with the `if (!error)` above ?

src/index.c
```diff
@@ -527,6 +529,7 @@ int git_index_clear(git_index *index)
 
 	assert(index);
 
+	index->dirty = 1;
```
tiennou: Unsure, but this seems the inverse of the commit's message rationale ?

tests/index/names.c
```diff
@@ -25,6 +25,42 @@ void test_index_names__cleanup(void)
 	cl_git_sandbox_cleanup();
 }
 
+static void index_add_main(void)
```
pks-t: In the commit message you talk of "This test", but this commit doesn't do anything but add a new function which is never being called. The actual caller is only being added in the last commit of this series, which is a confusing when reading commits one by one

tests/index/names.c
```diff
@@ -25,6 +25,42 @@ void test_index_names__cleanup(void)
...
+{
+	git_index_entry entry = {{0}};
+	const char *paths[][3] = {
+		{ "ancestor", "ours", "theirs" },
+		{ "ancestor2", "ours2", "theirs2" },
+		{ "ancestor3", "ours3", "theirs3" } };
+	const char **conflict;
+	size_t i;
+
+	for (i = 0; i < 3; i++) {
```
pks-t: `i < ARRAY_SIZE(paths)`?

tests/index/names.c
pks-t: By the way, one more thing: I don't really get what "main" is referring to in this context.

src/checkout.c
```diff
@@ -2397,20 +2397,20 @@ static int checkout_data_init(
 				GIT_DIR_MODE, GIT_MKDIR_VERIFY_DIR)) < 0)
 		goto cleanup;
 
+	if ((error = git_repository_index(&data->index, data->repo)) < 0)
+		goto cleanup;
```
pks-t: So, previously we were not loading the index into `data->index` when `GIT_CHECKOUT_NO_REFRESH` was set. This leaves me wondering about line 2441,  where we unconditionally dereference `data->index`. Was this code broken previously and would segfault in case `GIT_CHECKOUT_NO_REFRESH` was set or is the index already loaded at this point in time, resulting in a memory leak due to an additional reference to the index?

src/stash.c
```diff
@@ -159,7 +164,37 @@ struct stash_update_rules {
 	bool include_ignored;
 };
 
+/* similar to git_index_add_bypath but able to operate on any
```
pks-t: Multi-line comments should start with an empty `/*` line and upper-case

src/stash.c
```diff
@@ -159,7 +164,37 @@ struct stash_update_rules {
...
+	git_index *index,
+	const char *path)
+{
+	git_index *repo_index;
+	git_index_entry entry = {{0}};
+	struct stat st;
+	int error;
+
+	if (!git_repository_is_bare(repo) &&
+		(error = git_repository_index__weakptr(&repo_index, repo)) < 0)
```
pks-t: I'd like to finally settle on a common style when it comes to multi-line conditionals. We currently have two different styles in our code base. The first one aligned with a tab

```
if (!git_repository_is_bare(repo) &&
	(error = git_repository_index__weakptr(&repo_index, repo)) < 0)
	return error;
```

and the same thing aligned with four spaces instead

```
if (!git_repository_is_bare(repo) &&
    (error = git_repository_index__weakptr(&repo_index, repo)) < 0)
	return error;
```

I'd say the second one is superior, as you can clearly distinguish conditionals from the body. Especially when there are multiple more than two lines in a conditional, this becomes rather valuable and makes code a lot easier to read, in my opinion.

src/stash.c
```diff
@@ -159,7 +164,37 @@ struct stash_update_rules {
...
+		return error;
+
+	if((error = git_blob__create_from_paths(
```
pks-t: Space missing between `if` and opening brace

src/stash.c
```diff
@@ -159,7 +164,37 @@ struct stash_update_rules {
...
+	const char *path)
+{
+	git_index *repo_index;
+	git_index_entry entry = {{0}};
+	struct stat st;
+	int error;
+
+	if (!git_repository_is_bare(repo) &&
+		(error = git_repository_index__weakptr(&repo_index, repo)) < 0)
+		return error;
```
pks-t: Does this even have to be fatal? You're only using this to construct whether we want to trust the filemode or not, so in case we cannot use the repo index we can just ignore it.

```
int trust_filemode = 1;

if (!git_repository_is_bare(repo) && !git_repository_index__weakptr(&repo_index, repo))
     trust_filemode = !repo_index->distrust_filemode;
...
git_index_entry__init_from_stat(&entry, &st, trust_filemode);

File: src/index.c (hunk 2 above)
Comment (pks-t) [reply src/index.c]: Well. Clearing the index here without saving naturally has to leave the index dirty. I agree though that the commit message is confusing

File: src/index.c [hunk 10]
```diff
@@ -3185,6 +3212,7 @@ static int git_index_read_iterator(
 
 	clear_uptodate(index);
 
+	index->dirty = 0;
```
pks-t: Is this correct? I cannot see us saving the index to disk here, so I'd consider it dirty

src/index.c
```diff
@@ -3605,6 +3616,7 @@ int git_indexwriter_commit(git_indexwriter *writer)
 		return -1;
 	}
 
+	writer->index->dirty = 0;
```
pks-t: This looks like it belongs to the previous commit, where you introduce the dirty bit?

tests/checkout/tree.c
```diff
@@ -1551,8 +1566,12 @@ void test_checkout_tree__mode_change_is_force_updated(void)
 	cl_must_pass(p_chmod("testrepo/README", 0755));
 	cl_must_pass(git_index_add_bypath(index, "README"));
 
+	cl_git_pass(git_index_write(index));
 	assert_status_entrycount(g_repo, 1);
-	cl_git_pass(git_checkout_tree(g_repo, obj, &g_opts));
+
+	cl_git_pass(git_reset(g_repo, obj, GIT_RESET_HARD, NULL));
```
pks-t: Doesn't this change what we're actually testing? Previously, we've checked that `git_checkout_tree` resets the mode. Now we check whether `git_reset` resets the mode.

tests/checkout/typechange.c
```diff
@@ -289,6 +290,10 @@ void test_checkout_typechange__checkout_with_conflicts(void)
 		cl_assert_equal_i(cts.updates, 0);
 		cl_assert_equal_i(cts.ignored, 0);
 
+		cl_git_pass(git_repository_index(&index, g_repo));
+		cl_git_pass(git_index_read(index, 1));
```
pks-t: These interspersed changes of `git_index_read` leave me a bit uneasy. Does it mean that users of libgit2 may now face  places where they have to add `git_index_read` or face errors otherwise? If so, we at least have to provide an update guide detailing in which exact situations this is now necessary, and even then people will complain.

tests/cherrypick/workdir.c
```diff
@@ -276,6 +276,7 @@ void test_cherrypick_workdir__conflict_use_ours(void)
 	opts.merge_opts.file_favor = GIT_MERGE_FILE_FAVOR_OURS;
 
 	cl_git_pass(git_reset(repo, (git_object *)head, GIT_RESET_HARD, NULL));
+
```
pks-t: Unrelated whitespace change

src/stash.c
ethomson: > I'd like to finally settle on a common style when it comes to multi-line conditionals.  We currently have two different styles in our code base

I... guess?  I agree that we _shouldn't_ have two different styles, and yes, some things have slipped in here because _it really doesn't matter that much_ (more on that below).  But it feels like this was a solved decision.  We took git's style, which, from a cursory grep, is _typically_ indenting these with tabs.  And we _typically_ indent with tabs.

It seems like anything that's slipped in has done just that - slipped in.

> I'd say the second one is superior, as you can clearly distinguish conditionals from the body. Especially when there are multiple more than two lines in a conditional, this becomes rather valuable and makes code a lot easier to read, in my opinion.

I agree, but that's why I have my tabs set to four spaces.

But for the broader point:  it's _incredibly_ frustrating for contributors when they have to wait several weeks for a code review and then have that code review consist primarily of complaints about whitespace.  (This one was open three months ago!  _Three months_ is a long time, and yes, I realize that the complaints are not _only_ about whitespace.)

Does style matter?  Sure.  But they don't matter _nearly_ enough to stop the velocity of the project, or to turn people off from contributing.

Can you imagine waiting weeks to get a code review then it's picking nits at whitespace changes?   This is not the sign of a healthy community, and that person will probably never contribute again.

We need to fix this when it comes to reviewing pull requests.  This is not a complaint at you, individually; I do mean _we_, the project.

As for the whitespace issues: if you _really_ care about this, then add an automated step to the CI builds and give people something easy that they can run locally.  Otherwise, we need to relax, IMHO.

pks-t: >> I'd like to finally settle on a common style when it comes to multi-line conditionals. We currently have two different styles in our code base
> I... guess? I agree that we shouldn't have two different styles, and yes, some things have slipped in here because it really doesn't matter that much (more on that below). But it feels like this was a solved decision. We took git's style, which, from a cursory grep, is typically indenting these with tabs. And we typically indent with tabs.

It seems like anything that's slipped in has done just that - slipped in.

I might not have brought my intention along crlearly enough, but this was not a request for change but instead a clarification for what our agreed-upon style is. I had asked that specific question about how to indent multiline conditionals in the past and never got an answer (or at least I couldn't recall). So while I feel like the latter option is superior, I don't care which of both styles we use as long as it is consistent.

> I agree, but that's why I have my tabs set to four spaces.

Weird. I always thought the projects recommendation was eight spaces, and that's why I always had a specific setting for libgit2 to display as eight spaces. But in fact, we do recommend to set it to four spaces.

> But for the broader point: it's incredibly frustrating for contributors when they have to wait several weeks for a code review and then have that code review consist primarily of complaints about whitespace. (This one was open three months ago! Three months is a long time, and yes, I realize that the complaints are not only about whitespace.)

I know. And that's why I said in the past that I only add style-related critic in case there are other complaints about a PR.

> Does style matter? Sure. But they don't matter nearly enough to stop the velocity of the project, or to turn people off from contributing.
>
> Can you imagine waiting weeks to get a code review then it's picking nits at whitespace changes? This is not the sign of a healthy community, and that person will probably never contribute again.
> We need to fix this when it comes to reviewing pull requests. This is not a complaint at you, individually; I do mean we, the project.
>
> As for the whitespace issues: if you really care about this, then add an automated step to the CI builds and give people something easy that they can run locally. Otherwise, we need to relax, IMHO.

I'm definitely in favor of having an automated job to do this. While I agree that style shouldn't be the main concern, I think it in fact _is_ important for a project. I've seen enough projects where I wanted to contribute to, but after taking a quick peek into the implementation immediately bailed out as the code was so inconsistently formatted that I had problems to see what was going on. So it in fact is a double-edged sword: you need to care enough about style to not let the code base deteriorate and get hard to read, but you shouldn't be as strict to drive away contributors.

As you mentioned, I think the best way forward would be to have an automated job that does this immediately after a PR was submitted. If properly implemented, this would help both roles: the contributor can quickly fix up stuff while he is still familiar with the topic, and the maintainer can focus on the important parts.

tests/index/names.c
ethomson: Thanks; this got jumbled while rebasing.  I fixed it, and I also changed the name to `add_conflicts`.  Regarding "main", I was referring to the "main index" (as opposed to the other sections like the tree cache) but I updated it to avoid confusion.

src/checkout.c
ethomson: Yes, it would have dereferenced that `NULL` index.  the `NO_REFRESH` bit is really not a great pattern to use, so I'm not surprised that nobody's ever noticed this.  However, I added a test to make sure that we don't regress again.

src/stash.c
ethomson: It doesn't really _need_ to be fatal, per se.  But I expect anybody calling stash to need an index for their repository, so I think failing should not be harmful, and it's easier to reason about what happens if the index cannot be opened for some reason.

src/index.c
ethomson: Yes, I agree.  I've changed it.

tests/checkout/typechange.c
ethomson: This is a good question.  This affects what happens when you are replacing the contents of the index (eg, checkout, merge) and you have unsaved changes, then you will need to save them (`git_index_write`) or revert them (`git_index_read`).  If you do not, this new code path will raise an error.

I don't think that this is a common scenario (except in our tests, since they play fast and loose with things).  But there is a very valid case that we would be breaking:  checking out with the `FORCE` flag or a reset with the `HARD` flag when you have unsaved changes in the index.  At the moment, you can do:

    git_index_add(...);
    /* no save */
    git_reset(..., GIT_RESET_HARD);

And your unsaved changes are ignored - the index is updated to match the checkout / reset target.  With this patch, this will be an error.

Increasingly, I think this is incorrect and that we should allow `git_checkout` / `git_reset` to operate when forcing.  I think that this would fix most of the _intentional_ use cases of this sort of behavior for people who were in a bad place but `reset` their way out of it, knowingly.

I'd also be happy to add a configurable option here for people who need an escape hatch to opt-in to the prior behavior.  It's also possible that this is too heavy handed and we should keep the default as-is.

I'll follow up a commit that updates the `reset` / `checkout` behavior to allow forcing in this instance.

ethomson:
I added a change so that when checking out with `FORCE`, we do not halt on a dirty index.  This seems like a very correct change that is very much in keeping with the way arrbee had originally designed and specified `GIT_CHECKOUT_FORCE`, and how I would have expected things to work.  That also simplifies the tests so that they now don't need to do the `git_index_read` dance to discard their changes.

ethomson:
So this leaves one realistic use case that we're blocking here now, which is `GIT_CHECKOUT_SAFE` with a dirty index.  Previously we would have silently allowed this (obliterating the unsaved changes in the index), now we error on this.  I doubt anybody was doing this before, but I appreciate your concern.

Proposal: add a `git_libgit2_opt` that controls this behavior, default to the current behavior (silently discarding dirty indexes before safe checkout and friends), document that we'll upgrade this to a warning or error in the future?

pks-t:
> Proposal: add a git_libgit2_opt that controls this behavior, default to the current behavior (silently discarding dirty indexes before safe checkout and friends), document that we'll upgrade this to a warning or error in the future?

That sounds like a perfectly reasonable plan. I'm sold

include/git2/errors.h
```diff
@@ -55,6 +55,7 @@ typedef enum {
 	GIT_ITEROVER        = -31,	/**< Signals end of iteration with iterator */
 	GIT_RETRY           = -32,	/**< Internal only */
 	GIT_EMISMATCH       = -33,	/**< Hashsum mismatch in object */
+	GIT_EINDEXDIRTY     = -34,  /**< Unsaved changes in the index would be overwritten */
```
pks-t: This time I'm fast to review, so I'm allowed to nag at the whitespace error here :P

include/git2/common.h
```diff
@@ -363,6 +364,13 @@ typedef enum {
  *		> allocator will then be used to make all memory allocations for
  *		> libgit2 operations.
  *
+ *	 opts(GIT_OPT_ENABLE_UNSAVED_INDEX_SAFETY, int enabled)
+ *
+ *		> Ensure that there are no unsaved changes in the index before
+ *		> beginning any operation that reloads the index from disk (eg,
+ *		> checkout).  If there are unsaved changes, the instruction will
+ *		> fail.
```
pks-t: Should we maybe mention that this does not apply when forcing operations?

docs/changelog.md
```diff
@@ -16,6 +16,15 @@ v0.27 + 1
...
   `GIT_OPT_SET_ALLOCATOR` option with `git_libgit2_opts()`.
 
+* You can now ensure that functions do not discard unwritten changes to the
+  index via the `GIT_OPT_ENABLE_UNSAVED_INDEX_SAFETY` option to
+  `git_libgit2_opts()`.  This will cause functions that implicitly re-read
+  the index (eg, `git_checkout`) to fail if you have staged changes to the
+  index but you have not written the index to disk.
+
+  At present, this defaults to off, but we intend to enable this more
+  broadly in the future, as a warning or error.
```
pks-t: Should we urge consumers to review how they handle dirty indices and then enable the option?

ethomson:
Cool.  I've updated this with your suggestions, @pks-t.  Thanks!

ethomson:
> I've been briefly asking myself whether git_index_read_safely should be exposed to users or whether it should be the default of git_index_read with force set to false.

I don't _think_ so.  But I'll continue to give this some thought and do some testing with this functionality enabled before we make change changes.

PR #4537, opened 2018-02-19
tests: index::filemodes: fix use of uninitialized memory

pks-t:
The new index entry structure was not being initialized to all-zeroes.
As that structure is used to add a new entry to the current index, and
the hashing algorithm of the index making use of the uninitialized flags
to calculate the state, we might miscompute the hash of the entry and
add it at the wrong position. Later lookups would then fail.

Initialize the structure with `memset` to fix the test breaking on some
platforms.


ethomson:
😢 Thanks @pks-t.

Issue #4538, opened 2018-02-19
`git_diff_find_similar` can trigger SIGABRT (repro included)


wchargin:
Invoking `git_diff_find_similar` with specific flags can, on some commits, trigger a SIGABRT. I do not know why it occurs on those commits and not others, but it appears to be fully deterministic.

Ref. nodegit/nodegit#1442.

### Reproduction steps

Copy the following `repro.c` into the root of a fresh clone of the `libgit2` directory:

<details><summary><tt>repro.c</tt></summary>

```c
#include <inttypes.h>
#include <stdio.h>

#include <git2.h>

void try(int, const char *);

int main() {
    git_libgit2_init();

    int error;
    git_repository *repo;

    try(git_repository_open(&repo, "./tensorboard"), "opening repository");

    git_oid oid;
    char hex[] = "0c052866b837e91fdcc7e6c411d4879f062afb2c";
    try(git_oid_fromstr(&oid, hex), "parsing oid");

    git_commit *commit;
    try(git_commit_lookup(&commit, repo, &oid), "loading commit");

    git_commit *parent_commit;
    try(git_commit_parent(&parent_commit, commit, 0), "loading parent commit");

    git_tree *tree;
    try(git_commit_tree(&tree, commit), "loading tree");

    git_tree *parent_tree;
    try(git_commit_tree(&parent_tree, parent_commit), "loading parent tree");

    git_diff *diff;
    try(git_diff_tree_to_tree(&diff, repo, parent_tree, tree, NULL),
            "creating diff");

    git_diff_find_options options;
    try(git_diff_find_init_options(&options, GIT_DIFF_FIND_OPTIONS_VERSION),
            "creating diff find options");
    printf("Old flags: %"PRIu32"\n", options.flags);
    options.flags |= GIT_DIFF_FIND_RENAMES_FROM_REWRITES;
    printf("New flags: %"PRIu32"\n", options.flags);

    try(git_diff_find_similar(diff, &options), "finding similar");

    printf("Done\n");
    return 0;
}

void try(int error, const char *msg) {
    if (!error) {
        printf("[OK] %s\n", msg);
    } else {
        fprintf(stderr, "[%d] %s\n", error, msg);
        exit(1);
    }
}
```
</details>

Use the following script to download the `tensorflow/tensorboard` repository, which contains a commit that triggers the bug, and build and run the program:
<details><summary><tt>repro.sh</tt></summary>

```sh
#!/bin/sh
set -eu

[...]

PR #4539, opened 2018-02-20
diff_tform: fix rename detection with rewrite/delete pair

pks-t:
A rewritten file can either be classified as a modification of its
contents or of a delete of the complete file followed by an addition of
the new content. This distinction becomes important when we want to
detect renames for rewrites. Given a scenario where a file "a" has been
deleted and another file "b" has been renamed to "a", this should be
detected as a deletion of "a" followed by a rename of "a" -> "b". Thus,
splitting of the original rewrite into a delete/add pair is important
here.

This splitting is represented by a flag we can set at the current delta.
While the flag is already being set in case we want to break rewrites,
we do not do so in case where the `GIT_DIFF_FIND_RENAMES_FROM_REWRITES`
flag is set. This can trigger an assert when we try to match the source
and target deltas.

Fix the issue by setting the `GIT_DIFF_FLAG__TO_SPLIT` flag at the delta
when it is a rename target and `GIT_DIFF_FIND_RENAMES_FROM_REWRITES` is
set.

---

As this fixes hitting an `assert`, we should definitly include this in v0.27.0. So I will hold off v0.27.0-rc1 until this is merged.


ethomson:
I'll wait for the CI to finish in case there's some subtlety, but this looks good to me.

pks-t:
Fixes #4538

wchargin:
Thanks for investigating and fixing!

PR #4540, opened 2018-02-21
Fix buffer overrun when hunk header exceeds 80 characters

stanhu:
Closes https://github.com/libgit2/rugged/issues/716


stanhu:
This does not appear to be the right solution, as it breaks tests, and I think there shouldn't be a reason to cap the output at 80 characters?

PR #4542, opened 2018-02-23
Sanitize the hunk header to ensure it contains UTF-8 valid data

stanhu:
The diff driver truncates the hunk header text to 80 bytes, which
can  cut off 4-byte Unicode characters and introduce garbage characters in the
diff output. This change sanitizes the hunk header before it is displayed.

This is consistent with what git does: https://github.com/git/git/blob/e3a80781f5932f5fea12a49eb06f3ade4ed8945c/diff.c#L1927-L1941

Closes https://github.com/libgit2/rugged/issues/716

### DIFF
--- a/src/diff_xdiff.c
+++ b/src/diff_xdiff.c
@@ -6,6 +6,7 @@
  */
 
 #include "diff_xdiff.h"
+#include "util.h"
 
 #include "git2/errors.h"
 #include "diff.h"
@@ -115,6 +116,7 @@ static int git_xdiff_cb(void *priv, mmbuffer_t *bufs, int len)
 	const git_diff_delta *delta = patch->base.delta;
 	git_patch_generated_output *output = &info->xo->output;
 	git_diff_line line;
+	size_t buffer_len;
 
 	if (len == 1) {
 		output->error = git_xdiff_parse_hunk(&info->hunk, bufs[0].ptr);
@@ -124,6 +126,16 @@ static int git_xdiff_cb(void *priv, mmbuffer_t *bufs, int len)
 		info->hunk.header_len = bufs[0].size;
 		if (info->hunk.header_len >= sizeof(info->hunk.header))
 			info->hunk.header_len = sizeof(info->hunk.header) - 1;
+
+		/* Sanitize the hunk header in case there is invalid Unicode */
+		buffer_len = git__utf8_valid_buf_length((const uint8_t *) bufs[0].ptr, info->hunk.header_len);
+		/* Sanitizing the hunk header may delete the newline, so add it back again if there is room */
+		if (buffer_len < info->hunk.header_len) {
+			bufs[0].ptr[buffer_len] = '\n';
+			buffer_len += 1;
+			info->hunk.header_len = buffer_len;
+		}
+
 		memcpy(info->hunk.header, bufs[0].ptr, info->hunk.header_len);
 		info->hunk.header[info->hunk.header_len] = '\0';
 
--- a/src/util.c
+++ b/src/util.c
@@ -806,6 +806,22 @@ double git_time_monotonic(void)
 	return git__timer();
 }
 
+size_t git__utf8_valid_buf_length(const uint8_t *str, size_t str_len)
+{
+	size_t offset = 0;
+
+	while (offset < str_len) {
+		int length = git__utf8_charlen(str + offset, str_len - offset);
+
+		if (length < 0)
+			break;
+
+		offset += length;
+	}
+
+	return offset;
+}
+
 #ifdef GIT_WIN32
 int git__getenv(git_buf *out, const char *name)
 {
--- a/tests/diff/patch.c
+++ b/tests/diff/patch.c
@@ -25,6 +25,12 @@ void test_diff_patch__cleanup(void)
 
 #define EXPECTED_HUNK "@@ -1,2 +0,0 @@\n"
 
+#define UTF8_HUNK_HEADER "\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\n"
+
+#define UTF8_TRUNCATED_A_HUNK_HEADER "\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\xE6\x97\xA5\xE6\x9C\xAC\n"
+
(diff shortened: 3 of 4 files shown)


stanhu:
Does it make sense to pull in https://github.com/stanhu/rugged-encoding-test into `resources` and write a test against it?

src/diff_xdiff.c
```diff
@@ -115,13 +116,17 @@ static int git_xdiff_cb(void *priv, mmbuffer_t *bufs, int len)
...
 
 	if (len == 1) {
 		output->error = git_xdiff_parse_hunk(&info->hunk, bufs[0].ptr);
 		if (output->error < 0)
 			return output->error;
 
-		info->hunk.header_len = bufs[0].size;
+		/* Sanitize the hunk header in case there is invalid Unicode */
+		buffer_len = git__utf8_valid_buf_length((const uint8_t *) bufs[0].ptr, bufs[0].size);
+		info->hunk.header_len = buffer_len;
```
pks-t: So what you're doing here is cutting off at the point where there is the first invalid UTF8 codepoint. I think this solution here is actually not sufficient yet. In case `info->hunk.header_len >= sizeof(info->hunk.header)`, you're actually stripping the last byte again. In case the last byte belongs to a multi-byte  codepoint again, it will become invalid. So you actually have to sanitize after checking the header length.

But stepping back a bit: is it actually guaranteed that the header is in UTF8? Git is agnostic to the encoding in which files are being stored, so I could envision the diff machinery processing, say, UTF16-encoded data.

pks-t:
Oh, and one more thing: I'd definitly welcome to have a test here, as you propose. Especially as the issue at hand is quite easy to reproduce

src/diff_xdiff.c
ethomson: Git doesn't do well with UTF16.  UTF16 files will be treated as binary (since there's a NUL) and will probably not have a header at all.  I'm curious what would happen with non-UTF8 multibyte charsets, but I think it's good enough to emulate Git's behavior, which also seems to punt on those.

pks-t: Yeah, I know about them being treated as binary by default. It's possible to override that via gitattributes, though. But as you say, it's probably okay to just punt on that, then. It's only non-critical metainformation, after all.

stanhu:
Thanks. I addressed the review comments and added a test that passes locally in my environment, but for some reason the CI environment can't find the repository directory that I added.

stanhu:
@pks-t Would you mind doing another round of review? Thanks!

src/diff_xdiff.c
```diff
@@ -115,15 +116,19 @@ static int git_xdiff_cb(void *priv, mmbuffer_t *bufs, int len)
...
 		if (output->error < 0)
 			return output->error;
 
-		info->hunk.header_len = bufs[0].size;
 		if (info->hunk.header_len >= sizeof(info->hunk.header))
 			info->hunk.header_len = sizeof(info->hunk.header) - 1;
+
+		/* Sanitize the hunk header in case there is invalid Unicode */
+		buffer_len = git__utf8_valid_buf_length((const uint8_t *) bufs[0].ptr, bufs[0].size);
+		info->hunk.header_len = buffer_len;
```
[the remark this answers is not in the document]
ethomson: On that note, I don't understand the _test_ then.  Did you produce a file that has invalid utf-8 characters on a line, and are ensuring that they're truncated?  I ask because it appears that the test is doing _something_, since it appears that your `git__utf8_valid_buf_length` function has determined that the last two bytes do not include valid bytes to make up a utf-8 character.

I expected that the test would have done the truncation at a point that produced an some bytes that were invalid UTF-8 and then truncated them.  I think _that_ test would have illustrated this problem, since there would have been no truncation at all.

But I may be misunderstanding the test...?

src/util.c
```diff
@@ -801,6 +801,23 @@ double git_time_monotonic(void)
...
+	while (offset < str_len) {
+		length = git__utf8_charlen(str + offset, str_len - offset);
+
+		if (length < 0)
+			break;
+
+		offset += length;
+	}
+
+	return offset;
```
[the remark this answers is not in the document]
ethomson: I actually think that this function should both _take_ and _return_ `size_t` values.  At the moment, it's taking an `int`...  which should be a small value, due to the truncation, but the caller isn't actually doing the truncation (which highlights the problem).  It should ideally take a `size_t` which is what we use to measure the size of strings.

At that point, we should probably also _return_ a `size_t`.  Neither the inputs nor the outputs need to support negative values.

tests/diff/patch.c
```diff
@@ -610,3 +610,46 @@ void test_diff_patch__line_counts_with_eofnl(void)
...
+		"--- a/Japanese.md\n"
+		"+++ b/Japanese.md\n"
+		"@@ -24,7 +24,7 @@ Redactedでは**レイヤ**における**zIndex**属性を制御可能にする \n"
+		" **zIndexはグローバルではありません!**\n"
+		" \n"
+		"-CSSとは異なり、zIndexはすべてのオブジェクトに対してグローバルではありません。zIndexプロパティは親レイヤに対してローカルです。詳細につきましては、コンテナチュートリアルで説明しています。 [TODO: Link]。\n"
+		"+CSSとは異なり、zIndexはすべてのオブジェクトに対してグローバルではありません。zIndexプロパティは親レイヤに対してローカルです。詳細につきましては、コンテナチュートリアルで説明しています。 [現在作業中です: Link]。\n"
+		" \n"
+		" ### スクリーン上のスプライトの順序付け方法\n"
+		" これまで学んだことを生かして、画面にスプライトを表示して、zIndexの設定をしてみましょう！\n";
```
ethomson: Also, we've generally tried to escape these...  the [source character set in some VS compilers is the currently running codepage](https://msdn.microsoft.com/en-us/library/mt708821.aspx).

src/util.c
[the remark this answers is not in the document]
stanhu: Right, my mistake. I was just using the prototypes from the existing UTF-8 functions.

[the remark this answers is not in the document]
ethomson: No worries.  Thanks for tackling this problem, @stanhu - I've got a busy day, but I'll try to take another pass at reviewing.

src/diff_xdiff.c
```diff
@@ -115,15 +116,22 @@ static int git_xdiff_cb(void *priv, mmbuffer_t *bufs, int len)
...
 			return output->error;
 
 		info->hunk.header_len = bufs[0].size;
+
 		if (info->hunk.header_len >= sizeof(info->hunk.header))
 			info->hunk.header_len = sizeof(info->hunk.header) - 1;
+
+		/* Sanitize the hunk header in case there is invalid Unicode */
+		buffer_len = git__utf8_valid_buf_length((const uint8_t *) bufs[0].ptr, info->hunk.header_len);
+		info->hunk.header_len = buffer_len;
```
pks-t: Small nit only worthwhile to fix up in case you do another round of fixes: you can avoid `buffer_len` by simply directly assigning to `info->hunk.header_len`.

src/util.c
```diff
@@ -801,6 +801,31 @@ double git_time_monotonic(void)
 	return git__timer();
 }
 
+/**
+ * Iterate through an UTF-8 string and stops after finding any invalid UTF-8
```
pks-t: Sorry, I think you misunderstood my initial comment on this documentation. It should definitely be part of the header, but instead of having a `/** */` header, we want it to be a `/* */` header (note the difference of one `*` vs two `**` characters)

src/util.c
```diff
@@ -801,6 +801,31 @@ double git_time_monotonic(void)
...
+ * codepoints.
+ *
+ * @param str string to scan
+ * @param str_len size of the string
+ * @return length in bytes of the string that contains valid data
+ */
+size_t git__utf8_valid_buf_length(const uint8_t *str, size_t str_len)
+{
+	size_t offset = 0;
+	int length = 0;
```
pks-t: This initialization is unnecessary. You could also just directly declare the variable in the loop. It's only a minor nit, though, feel free to ignore this comment

tests/diff/patch.c
```diff
@@ -610,3 +610,46 @@ void test_diff_patch__line_counts_with_eofnl(void)
...
+	const char *another_sha = "cdd42da2d1d53d1db2c98b7d4e972b450601c1ef";
+	git_tree *one, *another;
+	git_diff *diff;
+	git_patch *patch;
+	git_buf buf = GIT_BUF_INIT;
+	const char *expected = "diff --git a/Japanese.md b/Japanese.md\n"
+		"index dd469c9..caf022d 100644\n"
+		"--- a/Japanese.md\n"
+		"+++ b/Japanese.md\n"
+		"@@ -24,7 +24,7 @@ Redactedでは**レイヤ**における**zIndex**属性を制御可能にする \n"
```
pks-t: I'm a bit surprised there's a space right in front of the newline. As the test should actually show that we're correctly truncating UTF8 at the context line's cutoff point, I'd have expected the last character before the NL to be a UTF8 code point.

stanhu: I am confused too. I verified that the buffer truncates the length properly; not sure where this extra space comes from yet. I'll have to dig into it later.

tests/diff/patch.c
stanhu: Ok, if you insist. I have to say https://github.com/libgit2/libgit2/pull/4542/commits/56cccc5e9babec0ab4c155013af164345a564857 looks incredibly ugly. :)

tests/diff/patch.c
stanhu: Ok, I figured it out. Sanitizing the buffer for valid UTF-8 characters will likely remove the newline from the hunk header, so we need to add it back if it was stripped.

stanhu:
@pks-t Ok, finally found some time to figure out what was going on here and respond to your feedback. Please take a look!

tests/diff/patch.c
```diff
@@ -610,3 +610,47 @@ void test_diff_patch__line_counts_with_eofnl(void)
 
 	git_buf_free(&content);
 }
+
+void test_diff_patch__can_strip_bad_utf8(void)
```
pks-t: All that we want to do in this test is to just make sure that the diff of two files produces the correct result. So in fact we don't really need to add the complete test repository here, which would be a bit wasteful. Instead, you could just use `git_diff_buffers`.

stanhu: I could add the two files (https://raw.githubusercontent.com/stanhu/rugged-encoding-test/3c4f7762042ad735d97d4781c4c7ee7d4486646c/Japanese.md) and (https://raw.githubusercontent.com/stanhu/rugged-encoding-test/387914a182ae0868364ec09fadb1beb47690fd4a/Japanese.md) in as character buffers, but wouldn't this pollute the test even more since we have to escape UTF-8 characters)? 

Ideally I'd be able to simplify the test case so that it boils down to a few lines, but when I tried earlier it was challenging to reproduce. I can try again.

stanhu: @pks-t Ok, I think I figured out a better test case. I just mirrored this one: https://github.com/git/git/blob/master/t/t4025-hunk-header.sh

pks-t: Oh, these tests are great! Thanks a lot!

pks-t:
Thanks a lot for this PR!

PR #4543, opened 2018-02-23
Reorder entries to decrease struct sizes

pks-t:
Members of structures may have several requirements regarding alignment,
depending on platform. Because of this, the order of members can
actually impact how big the overall structure is, with a more efficient
ordering saving space. In general, pointers should be first and scalar
types should be ordered by descending width.

We have some internal data structures whose members are not accessible
to the user of libgit2, where we can easily reorder members without
breaking anything. Depending on the platform, this may save a few bytes
for each structure. While this is not a lot, the savings may add up in
certain scenarios.

### DIFF
--- a/src/blame.h
+++ b/src/blame.h
@@ -13,10 +13,10 @@
  * One blob in a commit that is being suspected
  */
 typedef struct git_blame__origin {
-	int refcnt;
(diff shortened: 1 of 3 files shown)


pks-t:
Closing, as it hasn't sparked much interest :)

PR #4547, opened 2018-02-25
[WIP] fix missing TLS 1.3 on Travis

tiennou:
Since the "GitHub TLS deprecation of 2018", the old version of OpenSSL in use on Travis Trusty boxes can't do SSL anymore.

This is a shot in the dark at fixing that.


ethomson:
I don't think that this is an OpenSSL problem.  The failures look to me to be related to libssh.

ethomson:
(I realize that the root cause could still be OpenSSL, but I haven't debugged this far enough to believe that to be true, yet.)

tiennou:
You're right, I saw SS* and mismatched. I'll try to see what happens locally then before involving Travis.

ethomson:
I did, too, at first.  TLS 1.2 is the gift that keeps on giving.

tiennou:
So, I'm not getting those failures locally with a brewed openssl@1.0 and libssh2@1.8.0, so it sounds like a libssh2 problem (trusty has 1.4.3). I'm trying to get us to link against a src build of libssh2@1.4.3 to confirm, but it's not going well.

ethomson:
I have a trusty box, and I can confirm that the libssh2 in the box is too old to talk to github.  Updating to newest is 👍 .  However, there appears to be no libssh2 backports, so we'll have to build our own package.  🙄

PR #4550, opened 2018-02-25
winhttp: enable TLS 1.2

ethomson:
Windows 7 does not enable TLS 1.2 by default, though with Software Updates it does _support_ TLS 1.2 when requested.  Explicitly enable TLS 1.0-1.2 support, in the case that we're on Windows 7.

This lets libgit2 users on Windows 7 talk to GitHub.com (the website), which now requires TLS 1.2 or better.  This assumes that those Windows 7 users have installed the necessary software updates for said TLS 1.2 support.  But those updates do not enable TLS 1.2 by default, we must explicitly enable it.


src/transports/winhttp.c
```diff
@@ -786,6 +790,14 @@ static int winhttp_connect(
 		goto on_error;
 	}
 
+	if (!WinHttpSetOption(t->session,
```
matklad: What happens if we are on the unpatched win7? Will this call return an error because the os does not know about TLS1_2? Should we perhaps ignore the return code then?

ethomson:
So to follow up here after further investigation, here's some more data:

* I believe that this function can fail if and only if the underlying operating system lacks TLS 1.2 support
* When this call does fail because TLS 1.2 support is not available, the connection handle is still valid and the default protocol support is still enabled, so it's safe to continue on failure.
* Windows 7 has support for TLS 1.2 out of the box, without needing service packs or any software updates.  This is not terribly obvious, since there's a KB discussing TLS 1.2 support - that KB is actually for the registry key support for enabling TLS 1.2 as a default protocol.  Even without that update, applications can opt-in to TLS 1.2 support with this patch.
* Windows Vista was the first version of Windows to support TLS 1.2 at all.
* Windows XP does not support TLS 1.2 at all, nor will it, since it's now out of support.  That's true of consumer versions of Windows XP, anyway; note that there is a specialized Windows XP for Point of Sale devices, and it _does_ actually have an update to enable TLS 1.2.

I've updated this patch to do a few things:

1. We now ignore errors when trying to set the protocol versions. (Thanks @matklad for asking about this.)  I feel pretty confident that we should be able to set this on Windows 7 and we could validate the return code, but even though we don't technically support earlier versions of Windows, it seems like we might as well just ignore the errors.  We should only fail when TLS 1.2 is not available _and_ we should still be able to continue to talk to non-TLS 1.2 servers.

2. We hide this behind a version test and only set this version flag on Windows versions prior to version 8.  This ensures that we don't accidentally forget to update this when TLS 1.3 is available.  Instead, we will continue on with the OS defaults on newer platforms.

ethomson:
This was implicit, but to be clear: this patch does now explicitly allow libgit2 to connect to github.com on Windows 7 (out of the box) and better.  Windows XP and Vista can still connect to hosts that accept non-TLS 1.2, including Bitbucket and Visual Studio Team Services.  (Though obviously both those sites will be disabling earlier versions as well and enforcing TLS 1.2.)

ethomson:
Note, though, that non-manifested applications report themselves as Windows 8 (NT 6.2).  This patch only flips the protocols on for versions _prior_ to 6.2.  So Windows 8, 8.1 and 10 are using the system defaults; if they add TLS 1.3 _and_ turn that on as a default, then we should happily work with TLS 1.3.

matklad:
>This patch only flips the protocols on for versions prior to 6.2

Hm, according to the update description, `Windows Server 2012` is affected, and it is 6.2. However, `Windows 8` is not mentioned in the update...

ethomson:
> Hm, according to the update description, Windows Server 2012 is affected, and it is 6.2. However, Windows 8 is not mentioned in the update...

You're looking at the text on [KB 3140245](https://support.microsoft.com/en-gb/help/3140245/update-to-enable-tls-1-1-and-tls-1-2-as-a-default-secure-protocols-in)...

It does indeed say that.  I did not realize that until now, I assumed that since Server 2012 is Windows 8 that they would use the same policy here.  Let me investigate.

Arnavion:
Windows 8 went out of support a few months before that KB, which is probably why it isn't listed.

ethomson:
Unfortunately, this is not related to support: this is simply a misalignment between client and server versions of Windows.  Windows 8 supports TLS 1.2 by default; Windows Server 2012 does not.

I've updated this PR yet again, to _always_ try to enable TLS 1.2.  This means that we'll need to sort out support for TLS 1.3 when it becomes available, but this will suffice in the immediate term.

src/transports/winhttp.c
```diff
@@ -40,6 +40,19 @@
...
+#define WINHTTP_FLAG_SECURE_PROTOCOL_TLS1 0x00000080
+#endif
+
+#ifndef WINHTTP_FLAG_SECURE_PROTOCOL_TLS1_1
+#define WINHTTP_FLAG_SECURE_PROTOCOL_TLS1_1 0x00000200
+#endif
+
+#ifndef WINHTTP_FLAG_SECURE_PROTOCOL_TLS1_2
+#define WINHTTP_FLAG_SECURE_PROTOCOL_TLS1_2 0x00000800
+#endif
```
matklad: Hm, some similar constants are defined [over here](https://github.com/libgit2/libgit2/blob/89c332e41b12a72d89de40d63bc568c56a2c336a/deps/winhttp/winhttp.h#L442). Perhaps it makes sense to move these there as well?

ethomson: Good eye, thanks @matklad, I had forgotten about that file entirely.

deps/winhttp/winhttp.h
```diff
@@ -437,10 +437,12 @@ typedef int INTERNET_SCHEME, *LPINTERNET_SCHEME;
...
-#define WINHTTP_FLAG_SECURE_PROTOCOL_SSL2  0x00000008
-#define WINHTTP_FLAG_SECURE_PROTOCOL_SSL3  0x00000020
-#define WINHTTP_FLAG_SECURE_PROTOCOL_TLS1  0x00000080
-#define WINHTTP_FLAG_SECURE_PROTOCOL_ALL   (WINHTTP_FLAG_SECURE_PROTOCOL_SSL2 | WINHTTP_FLAG_SECURE_PROTOCOL_SSL3 | WINHTTP_FLAG_SECURE_PROTOCOL_TLS1)
+#define WINHTTP_FLAG_SECURE_PROTOCOL_SSL2    0x00000008
+#define WINHTTP_FLAG_SECURE_PROTOCOL_SSL3    0x00000020
+#define WINHTTP_FLAG_SECURE_PROTOCOL_TLS1    0x00000080
+#define WINHTTP_FLAG_SECURE_PROTOCOL_TLS1_1  0x00000200
+#define WINHTTP_FLAG_SECURE_PROTOCOL_TLS1_2  0x00000800
+#define WINHTTP_FLAG_SECURE_PROTOCOL_ALL     (WINHTTP_FLAG_SECURE_PROTOCOL_SSL2 | WINHTTP_FLAG_SECURE_PROTOCOL_SSL3 | WINHTTP_FLAG_SECURE_PROTOCOL_TLS1)
```
ethomson: If you're looking at this and you're thinking "well, that's not all the bits anymore, you just added two!" then you'd be right, but this constant also doesn't actually include _all_ the options!  Ha ha.  So I'm keeping this aligned with the actual value on Windows.

ethomson:
Just to further pile on to this thread a bit - I was really hoping to avoid hardcoding these values on newer systems that already support TLS 1.2 out of the box by default.  My thinking here was that if we did that then this instance of libgit2 would be prevented from automatically including TLS 1.3 support if that were ever added.

I looked at version checks, which fail for the reasons @matklad mentioned - Windows is actively trying to subvert using version checks as a means to determine whether things are supported, and so all of the APIs to get version information basically lie in order to make sure that you don't use them.

I also looked at `WinHttpQueryOption`, which unfortunately does not accept `WINHTTP_OPTION_SECURE_PROTOCOLS`.  That value can [only be set](https://msdn.microsoft.com/en-us/library/windows/desktop/aa384066(v=vs.85).aspx) according to the documentation.  (The usage backs this up, it fails trying to query.)

So, again, this is not ideal, but there's no option to do this, AFAICT.  I've asked around and will revisit this is I've missed something clever here, because allowing the OS to increase the supported protocols would be ideal.

src/transports/winhttp.c
ethomson: 😢 In fact, those are for mingw/mingw32, which needs the whole lib infrastructure that's in `deps/winhttp`.

mingw64 does not use that `deps/winhttp` stuff, it has its own headers for winhttp...  however, they're outdated.  so we actually need to include the constants for mingw64 as well.  (sigh.)

matklad: >so we actually need to include the constants for mingw64 as well. (sigh.)

Hm, I wonder if the headers for msvc on old windows versions (XP and win7) could also be "outdated"? So perhaps the original code with `ifndef/define` was actually correct? :-)

ethomson: Yep.  Bringing those back.

ethomson:
Shipping this and cutting a new release candidate for 0.27.0.  Thanks for your help @matklad!

matklad:
Awesome, thanks a lot, @ethomson !

Issue #4551, opened 2018-02-27
config: invalid read of size 8 reported by valgrind


tiennou:
While looking at #4544, I noticed a bunch of valgrind warnings, so I'm filing issues to give them a little more visibility.

I've tried to fix them [here](https://github.com/libgit2/libgit2/compare/master...tiennou:fix/valgrind-issues), which ended up with valgrind *even less* [happy](https://gist.github.com/50ffceaf7dc3cf8aa02384840dc8c900).

I suspect we're doing something wrong with pointers in `config_iterator_new` because we're working with a snapshot (made in `create_internal`) which has a different layout, hence the b->repo (`size 8`) is actually OOB.

```
==17828== Invalid read of size 8
==17828==    at 0x5C09A5: config_iterator_new (config_file.c:426)
==17828==    by 0x5BD7D5: all_iter_next (config.c:396)
==17828==    by 0x5BD886: all_iter_glob_next (config.c:424)
==17828==    by 0x5BEC29: git_config_next (config.c:1081)
==17828==    by 0x63A9EC: apply_insteadof (remote.c:2506)
==17828==    by 0x634F6C: create_internal (remote.c:216)
==17828==    by 0x63530F: git_remote_create_with_fetchspec (remote.c:321)
==17828==    by 0x63527D: git_remote_create (remote.c:304)
==17828==    by 0x429F53: test_checkout_index__honor_coresymlinks_default (index.c:156)
==17828==    by 0x4124F9: clar_run_test (clar.c:222)
==17828==    by 0x41279C: clar_run_suite (clar.c:286)
==17828==    by 0x412CFE: clar_test_run (clar.c:433)
==17828==  Address 0xb6630f0 is 0 bytes after a block of size 160 alloc'd
==17828==    at 0x4C2CC70: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==17828==    by 0x5BFC81: git__calloc (util.h:99)
==17828==    by 0x5C15F4: git_config_file__snapshot (config_file.c:844)
==17828==    by 0x5C11BA: config_snapshot (config_file.c:684)
==17828==    by 0x5BD1A1: git_config_snapshot (config.c:165)
==17828==    by 0x63D4C6: git_repository_config_snapshot (repository.c:1060)
==17828==    by 0x634EC2: create_internal (remote.c:203)
==17828==    by 0x63530F: git_remote_create_with_fetchspec (remote.c:321)
[...]

PR #4552, opened 2018-02-28
Cast less blindly between configuration objects

carlosmn:
As discussed in #4551 we were cowboying it up with the casts which meant we were passing in nonsense values for levels and reading unallocated memory for the repository pointer.

This fixes that and hardens the code a bit by casting to the common structure rather than the one for the live object.

/cc @tiennou 
//c @ethomson @pks-t this should go in before the release

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -86,13 +86,13 @@ typedef struct {
 	/* mutex to coordinate accessing the values */
 	git_mutex values_mutex;
 	refcounted_strmap *values;
+	const git_repository *repo;
 } diskfile_header;
 
 typedef struct {
 	diskfile_header header;
 
 	git_config_level_t level;
-	const git_repository *repo;
 
 	git_array_t(git_config_parser) readers;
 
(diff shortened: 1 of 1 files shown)


src/config_file.c
```diff
@@ -326,6 +326,9 @@ static int config_refresh(git_config_backend *cfg)
 	int error, modified;
 	uint32_t i;
 
+	if (b->header.parent.readonly)
+		return 0;
```
pks-t: Hm. Not all callers do check that, most notably `config_set` and `config_delete` functions don't. So this does change behavior, no? This change definitly looks a bit scary to me

carlosmn: `config_set` and `config_delete` only are set as the set and delete functions for the "live"/read-write object so. The snapshots (i.e. when `readonly` would be true) sets its own functions which return with an error.

So `config_set` and `config_delete` would only be called when `readonly` is false and thus simply continue.

pks-t: Oh, missed that, thanks for pointing out. The only thought left is that it feels a bit weird to return success even if we bail out early. I'd say returning an error code is a bit cleaner, as that may not leave us wondering at a later point why just some random configs are never being refreshed. As callers don't call `refresh` anyway when the config is read-only, this shouldn't have any negative impact right now but feels safer for the future.

carlosmn: We do have a function that we use in case you accidentally end up calling set and friends on a snapshot. I'll make it return that.

PR #4553, opened 2018-02-28
tree: initialize the id we use for testing submodule insertions

carlosmn:
Instead of laving it uninitialized and relying on luck for it to be non-zero,
let's give it a dummy hash so we make valgrind happy (in this case the hash
comes from `sha1sum </dev/null`.

This is just in the test so it's not _that_ important, but it'd be good to release with a clean valgrind output.

/cc @ethomson @pks-t this should go in before the release


pks-t:
I'm ignoring the unrelated badssl failure. This is obviously correct

PR #4557, opened 2018-02-28
Unescape special characters in SSH repo paths

stevek-axo:
Some repositories (i.e. VSTS repos) may have spaces in the repo path. The Git CLI is successfully able to handle this special case by unescaping special characters when constructing a command, e.g. `git-upload-pack '/Project with Spaces/_ssh/Repo with Spaces'`. The proposed change would ensure the same in libgit2.

### DIFF
--- a/src/netops.c
+++ b/src/netops.c
@@ -207,7 +207,7 @@ void gitno_connection_data_free_ptrs(gitno_connection_data *d)
 }
 
 #define hex2c(c) ((c | 32) % 39 - 9)
-static char* unescape(char *str)
+char* gitno_unescape(char *str)
 {
 	int x, y;
 	int len = (int)strlen(str);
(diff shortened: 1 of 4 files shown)


src/netops.c
```diff
@@ -274,8 +274,8 @@ int gitno_extract_url_parts(
 	if (u.field_set & (1 << UF_USERINFO)) {
 		const char *colon = memchr(_userinfo, ':', u.field_data[UF_USERINFO].len);
 		if (colon) {
-			*username = unescape(git__substrdup(_userinfo, colon - _userinfo));
-			*password = unescape(git__substrdup(colon+1, u.field_data[UF_USERINFO].len - (colon+1-_userinfo)));
+			*username = gitno_unescape(git__substrdup(_userinfo, colon - _userinfo));
+			*password = gitno_unescape(git__substrdup(colon+1, u.field_data[UF_USERINFO].len - (colon+1-_userinfo)));
 			GITERR_CHECK_ALLOC(*password);
 		} else {
 			*username = git__substrdup(_userinfo, u.field_data[UF_USERINFO].len);
```
ethomson: While you're in here, can you `unescape` this as well?  It should have been unescaped the whole time, but was overlooked.

ethomson:
So thanks for raising this issue.  While reviewing this, it feels to me like `gitno_extract_url_parts` is sort of deficient in a few ways.  I've opened #4563 to fix a few of these issues.

I think that we should consider taking this PR for the next release.  We should _perhaps_ think about cherry-picking df96f70 to avoid the unnecessary `malloc`, but I'm happy either way, tbqh.

ethomson:
Thanks for this; I've manually merged it.

PR #4558, opened 2018-02-28
travis: war on leaks

tiennou:
This should make Travis fail the build when leaks are detected.

~There's a prototype implementation on osx, because I just noticed `leaks` gained an `-atExit` argument, which I can't test locally for update-laziness reasons.~ Removed for the time being because it depends on [undocumentation](https://www.reddit.com/r/macprogramming/comments/7iao6d/using_the_leaks_command_on_a_cc_executable/).

Supersedes #4416.


script/citest.sh
```diff
@@ -74,10 +77,14 @@ export GITTEST_REMOTE_SSH_PUBKEY="$HOME/.ssh/id_rsa.pub"
...
-ctest -V -R libgit2_clar-proxy_credentials || exit $?
+if [ -n "$JAVA_FOUND" ]; then
+	# Use the proxy we started at the beginning
+	export GITTEST_REMOTE_PROXY_URL="localhost:8080"
+	export GITTEST_REMOTE_PROXY_USER="foo"
+	export GITTEST_REMOTE_PROXY_PASS="bar"
+	ctest -V -R libgit2_clar-proxy_credentials || exit $?
+else
+	echo "Cannot find java in path, skipping proxy test"
+fi
```
ethomson: Why are you making this change?  Are there instances where travis doesn't have java installed?  Or is this for running locally?  If the latter, then let's make it some explicit opt-out.

Otherwise I fear that travis will change its images without us noticing, we'll stop running these tests silently, and then we'll break something without knowing.

script/cileaks.sh
```diff
@@ -9,5 +9,5 @@ then
 fi
 
 if [ -n "$VALGRIND" -a -e "$(which valgrind)" ]; then
-	valgrind --leak-check=full --show-reachable=yes --error-exitcode=125 --suppressions=./libgit2_clar.supp $@ _build/libgit2_clar -ionline
+	valgrind --leak-check=full --show-reachable=yes --error-exitcode=125 --suppressions=./libgit2_clar.supp $@ _build/libgit2_clar -ionline -xbuf::oom
```
ethomson: Is this test still useful at all?  Maybe we should remove it entirely.

script/citest.sh
```diff
@@ -10,7 +10,7 @@ fi
 
 if [ ! -d _build ]; then
 	echo "no _build dir found; you should run cibuild.sh first"
-	exit 1
+	exit 0
```
ethomson: I need some more explanation here - why does this fail the coverity build?  Again, this seems like something that might could use an explicit flag that the coverity build can pass in so that we don't stop running our build+tests because of some subtle change.

script/citest.sh
tiennou: Yeah, this is slightly conflated with being able to run those scripts in a VM, hence this change (I didn't have `java` at first).

I understand the concern though, but on the other hand, changing that test made the proxy test run on the `osx` machines too. If I find a way to tell if the script is Travis run, I'll be able to make it hard error.

script/cileaks.sh
tiennou: I was originally planning to try and make leaks works too, but I can remove it in the mean time.

script/citest.sh
tiennou: All of the three new "build" steps (cibuild, citest & cileaks) will be run in order now, hence the need to "fail" successfully those 2 new steps so we don't. Alternatively, I could make the coverity.sh explicit instead of calling that from cibuild.sh…

tiennou:
I've removed all changes caused by my trying to reuse the scripts inside VMs. I've also run `travis lint` locally (hence the suppression of `dist`).

I've noticed that Travis has [build stages](https://docs.travis-ci.com/user/build-stages/), which can accept [conditions](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/), so maybe there's a way to make the Coverity/Valgrind scans be less "hacky": right now COVERITY completely shortcuts the build, and VALGRIND will run the test suite twice (once as part of cibuild.sh, and again as part of cileaks.sh).

As this is likely to end up in a mess of pushes, and I can't control AppVeyor builds, I favored being cautious.

script/cileaks.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -x
+
+# If this platform doesn't support test execution, bail out now
+if [ -n "$SKIP_TESTS" ];
+then
+	exit $?;
+fi
+
+if [ -n "$VALGRIND" -a -e "$(which valgrind)" ]; then
```
pks-t: I'd say we should error out if we request a Valgrind build without Valgrind being installed. Otherwise, we'll get a green CI even if the job doesn't do what it's supposed to do

script/cileaks.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -x
+
+# If this platform doesn't support test execution, bail out now
+if [ -n "$SKIP_TESTS" ];
+then
+	exit $?;
```
pks-t: Why do you pass back the value of the previous test? Shouldn't this just exit success?

script/cileaks.sh
```diff
@@ -9,5 +9,5 @@ then
 fi
 
 if [ -n "$VALGRIND" -a -e "$(which valgrind)" ]; then
-	valgrind --leak-check=full --show-reachable=yes --error-exitcode=125 --suppressions=./libgit2_clar.supp _build/libgit2_clar -ionline
+	valgrind --leak-check=full --show-reachable=yes --error-exitcode=125 --suppressions=./libgit2_clar.supp $@ _build/libgit2_clar -ionline
```
pks-t: This change looks unintentional?

libgit2_clar.supp
```diff
@@ -47,3 +47,10 @@
 	...
 	fun:__check_pf
 }
+
+{
+	ignore-curl-global-init
+	Memcheck:Leak
+	...
+	fun:curl_global_init
+}
```
pks-t: Oh, cool that we can do this now with our global curl initialization.

.travis.yml
```diff
@@ -48,13 +48,11 @@ matrix:
    - compiler: gcc
      env: COVERITY=1
      os: linux
-     dist: trusty
```
pks-t: I'd rather prefer if we'd delete the global "dist" key. It's nice to have all that information directly in the build matrix.

script/cileaks.sh
tiennou: This is a vanilla copy-paste of the block just above, which appeared in [8f426d7](https://github.com/libgit2/libgit2/commit/8f426d7dd26856150950d080fb821044bb2ea811#diff-9e7fd40b8d82d14b32f15f7c2b6eb485R16). I honestly have no idea why it's written that way, but it seems to predate AppVeyor, so there's that.

.travis.yml
tiennou: `dist` is a [global](https://docs.travis-ci.com/user/reference/trusty/) key though, so I can't really use the other. Else I'd have kept the other one 😉.

pks-t: I don't think this is true. For some time, our matrix contained jobs for both Trusty and Precise, and it worked perfectly fine

script/cileaks.sh
pks-t: True. Let's leave it as-is, then. But I'd be grateful if you at least remove these useless semicolons ;)

tiennou:
Rebased. I yanked the `dist:` key change out.

script/cileaks.sh
tiennou: It's intentional, the purpose is that you can `VALGRIND=1 ./script/cileaks.sh -stestcase` in a VM and get just that valgrinded (note that I actually changed the location of the `%@`, this one is outdated).

tiennou:
This is ready !

tiennou:
I gave a spin to Travis build stages, and it looks like [this](https://travis-ci.org/tiennou/libgit2/builds/368848563). Note that the Coverity build doesn't show up since it's not coming from libgit2/libgit2.

I think this could be used to do the pipelining @pks-t envisions in libgit2/discussions#1.

src/transports/local.c
```diff
@@ -512,8 +512,10 @@ static int foreach_reference_cb(git_reference *reference, void *payload)
 	git_revwalk *walk = (git_revwalk *)payload;
 	int error;
 
-	if (git_reference_type(reference) != GIT_REF_OID)
+	if (git_reference_type(reference) != GIT_REF_OID) {
+		git_reference_free(reference);
 		return 0;
+	}
```
pks-t: Would you care to put this into an extra PR? It's worthy to be included in v0.27.1, as I've just introduced that leak with v0.27.0

src/worktree.c
```diff
@@ -131,7 +131,7 @@ static int open_worktree_dir(git_worktree **out, const char *parent, const char
 		goto out;
 	}
 
-	if ((wt = git__calloc(1, sizeof(struct git_repository))) == NULL) {
+	if ((wt = git__calloc(1, sizeof(*wt))) == NULL) {
```
pks-t: This one, as well

tests/worktree/worktree.c
```diff
@@ -300,6 +300,7 @@ void test_worktree_worktree__add_with_explicit_branch(void)
 	git_reference_free(branch);
 	git_reference_free(wthead);
 	git_repository_free(wtrepo);
+	git_worktree_free(wt);
```
pks-t: And this

pks-t:
The PR work as advertised ;) Can't merge it due to that, unfortunately, as we got new leaks in our submodule code

pks-t:
#4641 is merged now. Could you please rebase on latest master?

tiennou:
Rebased!

pks-t:
Cool, we can finally merge this! Thanks for your awesome work

PR #4559, opened 2018-03-02
Worktree lock reason should be const

jacquesg:


src/worktree.c
```diff
@@ -383,7 +383,7 @@ int git_worktree_add(git_worktree **out, git_repository *repo,
 	return err;
 }
 
-int git_worktree_lock(git_worktree *wt, char *creason)
+int git_worktree_lock(git_worktree *wt, const char *creason)
```
pks-t: Just as you're already touching this. Would you mind fixing up this variable name to be `reason` instead of `creason`? It hurts my eyes :sweat_smile:

jacquesg: Will do.

jacquesg:
Should be all fixed, Travis seems to be stuck though.

ethomson:
Thanks for noticing this and sending a PR.  🎉

Issue #4528, opened 2018-02-11
git_diff_get_stats fails when LC_ALL=""


rgburke:
Generating diffs for some commits fails with the error `Invalid collation character` when `setlocale(LC_ALL, "");` has been called. A similar issue nodegit/nodegit#1097 suggests that `regcomp` called from `git_diff_driver_builtin` is the source of the error.

### Reproduction steps
Run the following program on the Linux kernel source tree (https://github.com/torvalds/linux):

```c
#include <locale.h>
#include <stdio.h>
#include <git2.h>

int main(int argc, char *argv[]) {
	setlocale(LC_ALL, "");

        git_libgit2_init();

        git_repository *repo;
        git_repository_open(&repo, ".git");

	git_object *obj = NULL;
	git_revparse_single(&obj, repo, "f8de483e7440b0d23ce6372b3ef8358841c8827b");

	git_commit *commit = NULL;
	git_commit_lookup(&commit, repo, git_object_id(obj));

	git_commit *parent = NULL;
	git_commit_parent(&parent, commit, 0);

	git_tree *commit_tree = NULL, *parent_tree = NULL;
	git_commit_tree(&commit_tree, commit);
	git_commit_tree(&parent_tree, parent);

	git_diff *diff = NULL;
	git_diff_tree_to_tree(&diff, repo, parent_tree, commit_tree, NULL);

	git_diff_stats *stats;
	int error = git_diff_get_stats(&stats, diff);

	if (error < 0) {
		fprintf(stderr, "ERROR %d: Could not get diff stats: %s\n", error, giterr_last()->message);
	}

        return 0;
}
```

### Expected behavior
Running this program should produce no output. When `setlocale(LC_ALL, "");` is commented out it produces no output as expected:

```
richard@DevNew:~/code/linux$ ./diff 
richard@DevNew:~/code/linux$ 
```

### Actual behavior
`git_diff_get_stats` fails with the error below:

```
richard@DevNew:~/code/linux$ ./diff 
ERROR -1: Could not get diff stats: Invalid collation character
richard@DevNew:~/code/linux$ 
```

### Version of libgit2 (release number or SHA1)
Using master (408b16c1b06526f509fdb8997818d648d4d1ea1e)

### Operating system(s) tested
Tested on Ubuntu 17.10:
```
[...]

PR #4560, opened 2018-03-02
Regcomp with LC_COLLATE

pks-t:
This is inspired by #4528, where we do fail certain regexes when LC_COLLATE is being set to something other than the default. This is unfortunate, and I actually wasn't able to come up with a satisfying solution. Pulling the following message from my commits. Feedback and ideas would be highly appreciated.

---

Our user diff patterns make heavy use of range notation (e.g. "[a-z]")
which are being passed to `regcomp`. Problem is though that these ranges
heavily depend on locale value of LC_COLLATE, as these ranges might be
different depending on the language. This is why character classes like
[:alnum:], [:digit:] etc. exist for regular expressions, which easily
fix the problem in a locale-independent way.

Convert our user patterns to use these character classes instead.
There's two problems, though:

1. The character classes might include characters which are not allowed
   in an actual function name. E.g. if the current language has any
   weird UTF8 characters which belong to the alphabet of that language,
   we may get invalid matches for these regexes.

2. We have inherited a catch-all word delimiter from git.git, which
   catches multibyte characters not part of ASCII with the high-bit set.
   git.git does this to ensure that unmatched words are not being
   treated as whitespace-only changes and thus not be shown. With
   LC_COLLATE set to a non-C locale, the pattern
   "[\xc0-\xff][\x80-\xbf]+" will actually cause the regex compilation
   to fail. See git.git, commit 664d44ee7 (userdiff: simplify word-diff
   safeguard, 2011-01-11).

   There are multiple possibilities here:

   - Just bail on systems without `regcomp_l`. By default, we use
     `regcomp_l` and pass it no locale, which instructs it to use the
     default locale. This is the best-case scenario. Unfortunately,
     `regcomp_l` is not specified by POSIX and thus is missing on many
     systems.

   - We could obviously open-code it and just write down the complete
     range of characters, but this is _very_ unhandy.

   - We could instruct callers to never set LC_COLLATE to something
     other than the C locale. But this is simply won't work out, and
     expecting callers to always do a `setlocale` dance around calls to
     libgit2 is not going to fly well.

   - Otherwise, there is not much to help us. The only option remaining
     is to remove this range altogether and rely on the calling program
     to correctly set up locales. In this case, we might be lucky and
     all characters with the high-bit set will be matched by the
     respective character classes. But... yeah.

   Nothing is really satisfying at all. This commit uses the last
   option, as it seems to be the least likely to upset everybody.

### DIFF
--- a/src/diff_driver.c
+++ b/src/diff_driver.c
@@ -283,7 +283,9 @@ static int git_diff_driver_load(
 	/* TODO: warn if diff.<name>.command or diff.<name>.textconv are set */
 
 	git_buf_truncate(&name, namelen + strlen("diff.."));
-	git_buf_put(&name, "xfuncname", strlen("xfuncname"));
+	if ((error = git_buf_PUTS(&name, "xfuncname")) < 0)
+		goto done;
+
 	if ((error = git_config_get_multivar_foreach(
 			cfg, name.ptr, NULL, diff_driver_xfuncname, drv)) < 0) {
 		if (error != GIT_ENOTFOUND)
@@ -292,7 +294,9 @@ static int git_diff_driver_load(
 	}
 
 	git_buf_truncate(&name, namelen + strlen("diff.."));
-	git_buf_put(&name, "funcname", strlen("funcname"));
+	if ((error = git_buf_PUTS(&name, "funcname")) < 0)
+		goto done;
+
 	if ((error = git_config_get_multivar_foreach(
 			cfg, name.ptr, NULL, diff_driver_funcname, drv)) < 0) {
 		if (error != GIT_ENOTFOUND)
@@ -307,7 +311,9 @@ static int git_diff_driver_load(
 	}
 
 	git_buf_truncate(&name, namelen + strlen("diff.."));
-	git_buf_put(&name, "wordregex", strlen("wordregex"));
+	if ((error = git_buf_PUTS(&name, "wordregex")) < 0)
+		goto done;
+
 	if ((error = git_config__lookup_entry(&ce, cfg, name.ptr, false)) < 0)
 		goto done;
 	if (!ce || !ce->value)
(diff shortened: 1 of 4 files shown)


ethomson:
> We could obviously open-code it and just write down the complete
range of characters, but this is very unhandy.

Naive question:  it sounds like if we "open-code it and just write down the complete range of characters" that it will resolve this problem in full, regardless of the locale, calls to `setlocale` and existence of `regcomp_l`.  The only drawback is that it is - as you put it - "very unhandy".  (I quite like that description, btw.)

It has an additional drawback that we've now diverged from the git.git regexes, though we may have already done that.

Can we add a function into the mix before compiling the regexes that does the expansion ourselves?  I suspect that there are some crazy rules in regex character classes _but_ since these are all constants in our code base, we probably don't need to support every crazy thing and have a limited subset to work with.

carlosmn:
This does seem like the least terrible option. We would ideally not diverge from git's regexes, if for no other reason that any divergence is going to be counted as a bug against us. So it would be great if we could get git to also use `:digit:` and friends.

pks-t:
It's a bit hard to argue for git to use ":digit:", though. They don't care, as they will never use `setlocale` themselves. And ":digit:" etc. may contain characters that simply shouldn't match (I think? It's weird and hard to come by good descriptions). I wouldn't expect a set of Bengali digits to match a hex number, as example (stupid example, there is xdigit, but you get what I mean).

The translation system would probably be doable, but really inperformant. We'll have to use additional memory allocations to create the actual pattern, and this would not only be used for the userdiff patterns but also for stuff like querying configuration keys.

By the way, I just noticed that my previous statement about glibc providing `regcomp_l` is wrong. They do _not_ provide that, so users will in fact run into that problem as soon as they use `setlocale`.

ethomson:
Circling back - seeing as we already have a regex that we ship as a dependency, could we simply always use our regex engine instead of using it as a fallback when the system doesn't have one?

pks-t:
No, we can't, as deps/regex does not provide `regcomp_l`. :(

pks-t:
Closing in favor of #4935

PR #4561, opened 2018-03-02
[RFC] util: introduce GIT_DOWNCAST macro

pks-t:
In some parts of our code, we make rather heavy use of downcasting
structures to their respective specialized implementation. One example
is the configuration code with the general `git_config_backend` and the
specialized `diskfile_header` structures. At some occasions, it can
get confusing though with regards to the correct inheritance structure,
which led to the recent bug fixed in 2424e64c4 (config: harden our use
of the backend objects a bit, 2018-02-28).

Object-oriented programming in C is hard, but we can at least try to
have some checks when it comes to casting around stuff.
Thus, this commit introduces a `GIT_DOWNCAST` macro, which accepts as
parameters the pointer that is to be casted, the pointer it should be
cast to as well as the member inside of the target structure that is the
parent structure. This macro then tries hard to detect mis-casts:

- It checks whether the source and target pointers are of the same type.
  This requires support by the compiler, as it makes use of the builtin
  `__builtin_types_compatible_p`.

- It checks whether the embedded member of the target structure is the
  first member. This is simply done by verifying that the offset of the
  member in the target structure is 0.

In case either of the above two conditions is not fulfilled, an `assert`
is triggered. Furthermore, these checks do not make use of any
information not known at compile type. As such, the macro should get
optimized away at compile time and have no overhead in case the cast can
be done.

Convert one instance of upcasting to use this new macro as a proof of
concept. This instance is the one that was fixed with 2424e64c4 --
reverting the change and upcasting to the wrong type does indeed trigger
an assert at runtime. This demonstrates that the error would've been
detected earlier if the new macro had existed back then and was used.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -227,9 +227,9 @@ static void backend_free(git_config_backend *_backend)
 
 static int config_iterator_new(
 	git_config_iterator **iter,
-	struct git_config_backend* backend)
+	struct git_config_backend *backend)
 {
-	diskfile_header *bh = (diskfile_header *) backend;
+	diskfile_header *bh = GIT_CONTAINER_OF(backend, diskfile_header, parent);
 	git_config_entries *entries;
 	int error;
 
--- a/src/iterator.c
+++ b/src/iterator.c
@@ -1521,7 +1521,7 @@ static void filesystem_iterator_set_current(
 static int filesystem_iterator_current(
 	const git_index_entry **out, git_iterator *i)
 {
-	filesystem_iterator *iter = (filesystem_iterator *)i;
+	filesystem_iterator *iter = GIT_CONTAINER_OF(i, filesystem_iterator, base);
 
 	if (!iterator__has_been_accessed(i))
 		return iter->base.cb->advance(out, i);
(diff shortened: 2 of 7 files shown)


pks-t:
I've improved upon the implementation. As this PR is already using compiler-builtins in case `__GNUC__` is defined, there really is no reason to not go for the complete thing and provide compile-time errors. The amended PR now makes use of `__builtin_choose_expr`, and either chooses the "correct" cast expression or chooses an expression which results in a compiler error. Like this, we have compile-time sanity checks for downcasting on GCC and Clang.

tiennou:
See 102c52dcba6b95119b0481b2da191623d4562c16, which takes care of downcasts in transports and refdb_fs.

pks-t:
Cool, thanks! I've rebased this PR on master and cherry-picked
your commit.

tiennou:
I just found another potential user, so here's a rebased branch: https://github.com/libgit2/libgit2/compare/master...tiennou:pks/downcasting.

pks-t:
Thanks, @tiennou, I've adopted your branch.

I did make one major change, though: instead of `GIT_DOWNCAST`, I've renamed the macro to `GIT_CONTAINER_OF`. I've realized that the macro does essentially the same thing as `container_of` in linux.git, and I think that `container_of` is a lot easier to grasp on first sight than `downcast`.

pks-t:
@tiennou: want to have another look so that we can get this merged?

tiennou:
I'm fine with the rename, so unless you need a real look, I should still fine with the rest of the changes 😉. That makes me commutative or something, right ? 🤣

pks-t:
You want to approve this PR, then, asOtherwise I'm not allowed to merge? So no, you're not commutative :P

pks-t:
Thanks to both of you!

PR #4562, opened 2018-03-02
Removed unused macros from config_file.c

meatcoder:
This PR does not make any functional changes. It simply removes some macros that don't appear to be used anywhere. I found them while trying to read and understand the code in that file. 

I ran the tests locally and I did not see any failures with these macros removed.


pks-t:
Thanks a lot for pointing these out. Unfortunately, this change is already included as a preceding part of general refactorings in #4525. As that one is a bit more involved and merging this here would create merge conflicts, I'll close this here in favor of #4525.

PR #4563, opened 2018-03-03
Refactor `gitno_extract_url_parts`

ethomson:
`gitno_extract_url_parts` seems a little deficient in a few ways:

* This moves percent decoding into `git_buf`s.  This lets the SSH transport avoid an unnecessary small, temporary `malloc`.
* Now `gitno_extract_url_parts` uses the same buffer-based decoding
* Percent decode usernames even when passwords are not provided
* Percent decode hostnames, per RFC 3986

This builds on #4557.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -212,7 +212,7 @@ int git_buf_put(git_buf *buf, const char *data, size_t len)
 		size_t new_size;
 
 		assert(data);
-		
+
 		GITERR_CHECK_ALLOC_ADD(&new_size, buf->size, len);
 		GITERR_CHECK_ALLOC_ADD(&new_size, new_size, 1);
 		ENSURE_SIZE(buf, new_size);
@@ -453,6 +453,36 @@ on_error:
 
 	giterr_set(GITERR_INVALID, "invalid base85 input");
 	return -1;
+}
+
+#define HEX_DECODE(c) ((c | 32) % 39 - 9)
+
+int git_buf_decode_percent(
+	git_buf *buf,
+	const char *str,
+	size_t str_len)
+{
+	size_t str_pos, new_size;
+
+	GITERR_CHECK_ALLOC_ADD(&new_size, buf->size, str_len);
+	GITERR_CHECK_ALLOC_ADD(&new_size, new_size, 1);
+	ENSURE_SIZE(buf, new_size);
+
+	for (str_pos = 0; str_pos < str_len; buf->size++, str_pos++) {
+		if (str[str_pos] == '%' &&
+			str_len > str_pos + 2 &&
+			isxdigit(str[str_pos + 1]) &&
+			isxdigit(str[str_pos + 2])) {
+			buf->ptr[buf->size] = (HEX_DECODE(str[str_pos + 1]) << 4) +
+				HEX_DECODE(str[str_pos + 2]);
+			str_pos += 2;
+		} else {
+			buf->ptr[buf->size] = str[str_pos];
+		}
+	}
+
+	buf->ptr[buf->size] = '\0';
+	return 0;
 }
 
 int git_buf_vprintf(git_buf *buf, const char *format, va_list ap)
--- a/src/netops.c
+++ b/src/netops.c
@@ -206,24 +206,6 @@ void gitno_connection_data_free_ptrs(gitno_connection_data *d)
 	git__free(d->pass); d->pass = NULL;
 }
 
-#define hex2c(c) ((c | 32) % 39 - 9)
-char* gitno_unescape(char *str)
-{
-	int x, y;
-	int len = (int)strlen(str);
-
-	for (x=y=0; str[y]; ++x, ++y) {
-		if ((str[x] = str[y]) == '%') {
-			if (y < len-2 && isxdigit(str[y+1]) && isxdigit(str[y+2])) {
-				str[x] = (hex2c(str[y+1]) << 4) + hex2c(str[y+2]);
-				y += 2;
-			}
-		}
-	}
-	str[x] = '\0';
-	return str;
-}
-
 int gitno_extract_url_parts(
 	char **host_out,
 	char **port_out,
--- a/src/netops.c
+++ b/src/netops.c
@@ -238,7 +238,7 @@ int gitno_extract_url_parts(
 	if (has_host) {
 		const char *url_host = url + u.field_data[UF_HOST].off;
 		size_t url_host_len = u.field_data[UF_HOST].len;
-		git_buf_put(&host, url_host, url_host_len);
+		git_buf_decode_percent(&host, url_host, url_host_len);
 	}
 
 	if (has_port) {
--- a/src/netops.c
+++ b/src/netops.c
@@ -207,7 +207,7 @@ void gitno_connection_data_free_ptrs(gitno_connection_data *d)
 }
 
 #define hex2c(c) ((c | 32) % 39 - 9)
-static char* unescape(char *str)
+char* gitno_unescape(char *str)
 {
 	int x, y;
 	int len = (int)strlen(str);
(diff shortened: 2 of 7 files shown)


src/buffer.c
```diff
@@ -455,6 +455,37 @@ int git_buf_decode_base85(
...
 
+#define HEX_DECODE(c) ((c | 32) % 39 - 9)
+
+int git_buf_decode_percent(
+	git_buf *buf,
+	const char *str,
+	size_t str_len)
+{
+	size_t str_pos, new_size;
+	int error = 0;
```
pks-t: This variable is unused

src/buffer.c
```diff
@@ -455,6 +455,37 @@ int git_buf_decode_base85(
...
+
+	GITERR_CHECK_ALLOC_ADD(&new_size, buf->size, str_len);
+	GITERR_CHECK_ALLOC_ADD(&new_size, new_size, 1);
+	ENSURE_SIZE(buf, new_size);
+
+	for (str_pos = 0; str_pos < str_len; buf->size++, str_pos++) {
+		if (str[str_pos] == '%' &&
+			str_len > str_pos + 2 &&
+			isxdigit(str[str_pos + 1]) &&
+			isxdigit(str[str_pos + 2])) {
```
pks-t: I'm still in favour of indenting additional lines for if-statements by four spaces. Makes it more obvious what still belongs to the if-statement and what not. Anyway, no need to settle on that issue now.

src/buffer.c
```diff
@@ -455,6 +455,37 @@ int git_buf_decode_base85(
...
+	size_t str_pos, new_size;
+	int error = 0;
+
+	GITERR_CHECK_ALLOC_ADD(&new_size, buf->size, str_len);
+	GITERR_CHECK_ALLOC_ADD(&new_size, new_size, 1);
+	ENSURE_SIZE(buf, new_size);
+
+	for (str_pos = 0; str_pos < str_len; buf->size++, str_pos++) {
+		if (str[str_pos] == '%' &&
+			str_len > str_pos + 2 &&
```
pks-t: Hm. Doesn't this need to be `>=`? I'd expect like this if there is a "%xx" at the end of string it won't be decoded.

tests/buf/percent.c
```diff
@@ -0,0 +1,35 @@
...
+}
+
+void test_buf_percent__decode_succeeds(void)
+{
+	expect_decode_pass("", "");
+	expect_decode_pass("a", "a");
+	expect_decode_pass("github.com", "github.com");
+	expect_decode_pass("github.com", "githu%62.com");
+	expect_decode_pass("github.com", "github%2ecom");
+	expect_decode_pass("foo bar baz", "foo%20bar%20baz");
```
pks-t: You should probably also add the case of trailing "%xx"

src/netops.c
```diff
@@ -206,83 +206,100 @@ void gitno_connection_data_free_ptrs(gitno_connection_data *d)
...
-		GITERR_CHECK_ALLOC(*username);
-
 	}
 
-	return 0;
+	if (git_buf_oom(&host) ||
+		git_buf_oom(&port) ||
+		git_buf_oom(&path) ||
+		git_buf_oom(&username) ||
+		git_buf_oom(&password))
```
pks-t: This leaks memory in case only some of these buffers OOM'd

src/netops.c
```diff
@@ -206,83 +206,100 @@ void gitno_connection_data_free_ptrs(gitno_connection_data *d)
...
-			git__free(*host);
-			*host = NULL;
-			giterr_set(GITERR_NET, "invalid url, missing path");
-			return GIT_EINVALIDSPEC;
-		}
+	has_userinfo = !!(u.field_set & (1 << UF_USERINFO));
+	url_userinfo = url + u.field_data[UF_USERINFO].off;
+	url_userinfo_len = u.field_data[UF_USERINFO].len;
+
+	if (has_host)
```
pks-t: Dunno. Wouldn't this function be easier to read if setting the buffers would be all in one spot? E.g.

```
if (!!(u.field_set & (1 << UF_HOST)))
        git_buf_decode_percent(&host, url + u.field_data[UF_HOST].off, u.field_data[UF_HOST].len);
```

A lot less variables and all handling of that information is at exactly one spot.

src/buffer.c
ethomson: I don't think so: consider some string of length 4 (`str_len` in this case is `4`).  If we were to use `>=` instead - then this comparison becomes `str_len >= str_pos + 2`, and then this would be true for `str_pos = 2`.  Consider that the string of length 4 is the string: `ab%z`.  At the point in the loop when `str_pos = 2`, then we would evaluate true on the first condition (`str[str_pos] == '%'`), and the second (`str_len >= str_pos + 2`), and then would walk off the end of the array in the fourth.  (`str[str_pos + 2]`).

src/netops.c
ethomson: The lines are soooooo long then.  It is much cleaner to when we use some temporary variables.

src/netops.c
ethomson: Yes.

src/buffer.c
ethomson: Right, it's unnecessary.

tests/buf/percent.c
ethomson: Sure.

ethomson:
I added a test for the trailing `%xx` and changed the tests a bit to ensure that the tests only read the `len` bytes (by adding trailing garbage at the end of the strings, past `len` bytes).  This ensures that we actually look at the `len` bytes instead of looking for a NUL terminator.

src/buffer.c
pks-t: Agreed

src/netops.c
pks-t: Okay

src/netops.c
pks-t: Certainly true. On the other hand, you could just have locale variables inside of that block

src/netops.c
pks-t: Okay

tests/buf/percent.c
```diff
@@ -0,0 +1,49 @@
...
+	expect_decode_pass(" ", "%20");
+	expect_decode_pass("a", "a");
+	expect_decode_pass(" a", "%20a");
+	expect_decode_pass("a ", "a%20");
+	expect_decode_pass("github.com", "github.com");
+	expect_decode_pass("github.com", "githu%62.com");
+	expect_decode_pass("github.com", "github%2ecom");
+	expect_decode_pass("foo bar baz", "foo%20bar%20baz");
+	expect_decode_pass("foo bar baz", "foo%20bar%20baz");
+	expect_decode_pass("foo bar ", "foo%20bar%20");
```
pks-t: :+1:

src/buffer.c
pks-t: Convinced, thanks!

pks-t:
Looks good to me. I'm still a bit on the edge with regards to all those variables getting declared at the start of the function. Making them block-local to the conditions would clean that up a bit, I'd guess. Not blocking this PR for it, though. Feel free to merge and do an rc3

ethomson:
> Making them block-local to the conditions would clean that up a bit, I'd guess.

Yeah, that seems like a nice compromise.  I did this, and left the `bool` declarations the way they were.  I think it's an improvement.

ethomson:
What I don't understand is why the tests succeed _sometimes_.  🤔

pks-t:
No idea either. Maybe it depends on the server, where we get redirected to faulty instances by their load balancer in some occasions.

Regarding the compiler I'm completely clueless which one it was. I'm really sure to have experienced it at some point in time, though

ethomson:
> Regarding the compiler I'm completely clueless which one it was. I'm really sure to have experienced it at some point in time, though

We have whitespace between decls all over the codebase.  [Here's just one example](https://github.com/libgit2/libgit2/blob/master/tests/merge/trees/commits.c#L29) that has been in the codebase for years now.

If there's a compiler so broken that it can't parse whitespaces, then thankfully nobody's actually used it to compile us or file an issue about it.  We have a lot of problems that keep us from moving things forward that we should be focusing on instead of that.

PR #4568, opened 2018-03-07
deps: upgrade embedded zlib to version 1.2.11

pks-t:
The current version of zlib bundled with libgit2 is version 1.2.8. This
version has several CVEs assigned:

- CVE-2016-9843
- CVE-2016-9841
- CVE-2016-9842
- CVE-2016-9840

Upgrade the bundled version to the current release 1.2.11, which has
these vulnerabilities fixes.


pks-t:
This is the PR for the 0.26 branch. It also includes the version bump. As this is security critcial, we should probably release this before the original intended v0.26.1

PR #4571, opened 2018-03-08
Integer overflow

jacquesg:
This fixes a problem where we're potentially left shifting a signed number too far. Observed with the SUNPro compiler on Solaris:

```"deps/libgit2/include/git2/diff.h", line 214: warning: integer overflow detected: op "<<" (E_INTEGER_OVERFLOW_DETECTED)```

It also enables the warning to trap it if supported.


pks-t:
Thanks a lot for this PR. Obviously good!

PR #4574, opened 2018-03-10
Allow swapping out the memory allocation functions.

arthurschreiber:
This is mainly useful for bindings like rugged, where we can swap these
functions with the ones provided by Ruby. This helps Ruby to better
track memory usage through libgit2, and allows it to trigger GC runs
more intelligently.

The implementation in this pull request uses the existing `git_libgit2_opts` infrastructure to allow changing the allocator functions.

This is obviously missing documentation, as I want to first get some feedback whether `git_libgit2_opts` is the correct place for this or not. 😅

I tested this from the rugged bindings, and things were working just fine, without any super obvious performance impact. 👍


src/util.h
```diff
@@ -82,28 +82,34 @@
 
 #else
 
+void* (*git__malloc_func)(size_t size) = malloc;
+void* (*git__calloc_func)(size_t nitems, size_t size) = calloc;
```
arthurschreiber: I think we'll want to have `typedef`s for these, so we don't need to repeat them here and over in `src/settings.c`.

ethomson:
Don’t we have a `strndup` also?  I think that we probably want to either wire that up for consistency or drop this `strdup` and always implement our `strdup` in terms of `malloc` and the user need not reimplement it.

src/util.h
```diff
@@ -82,28 +82,34 @@
...
+char* (*git__strdup_func)(const char *str1) = strdup;
+void* (*git__realloc_func)(void *ptr, size_t size) = realloc;
+void* (*git__free_func)(void *ptr) = free;
```
carlosmn: This returns void, not void pointer

src/util.h
```diff
@@ -82,28 +82,34 @@
...
+
```
carlosmn: All of these assignments need to be in a .c file. The header can then have an extern definition

pks-t:
Haha. Should've looked first if there was another PR already. Well, anyway. see #4576 for another implementation.

carlosmn:
I do like the approach of passing in a structure; it should be less error prone than a bunch of untyped varargs.

carlosmn:
#4576 also brings up that we have things like `git__reallocarray` which are implemented on top of the standard allocator with better semantics. I think we should let the user specify those but we should make them optional so we fall back to our own `reallocarray` if the pointer is `NULL`. This would mean another pass through them to make sure that e.g. `git__stdalloc_reallocarray` doesn't call `realloc` but `git__stdalloc_realloc` to use the names from the other PR.

pks-t:
I'm closing this in favor of #4576. Thanks for your work?

PR #4576, opened 2018-03-10
Custom memory allocators

pks-t:
This is an RFC regarding custom memory allocators. Existing standard allocators are being renamed to `git__stdalloc__<fn>`. By default, our usual allocators `git__<fn>` are being defined to `git__stdalloc__<fn>`. For CRTDBG, our allocators `git__<fn>` are being defined to `git__crtdbg__<fn>`.

If the CMake option "PLUGGABLE_ALLOCATORS" is set, we use an indirection. Instead of using typedefs, each of our allocator functions is not a define but instead a function pointer, which is by default set to the respecfive `git__stdalloc__<fn>` function. Our `git_libgit2_opts` function is extended by "GIT_OPT_SET_ALLOCATOR, which receives as argument a `git_allocator` structure. Its members will then be used to change the function pointers.

I don't know whether we really need to distinguish between builds with and without pluggable allocators. The question is whether there's a performance hit when using function pointers. I haven't measured that yet, though. 

The other thing I'm unhappy about is the CRTDBG stuff. It would be rather cool if we could allow users of libgit2 to swap in CRTDBG at runtime, in contrast to how we do it at compile time right now. It would also easily be possible, except for the fact that all CRTDBG functions have two additional parameters to create the correct trace: __LINE__ and __FILE__. We could just extend all our allocators by these parameters. But this could again have a performance penalty, I guess. Again, this needs to be measured.

Anyway. I first want your opinions on these points. The implementation would still need a bit of cleanup. The `git_allocator` stuff should probably be its own module "alloc.{c,h}", but that's minor stuff only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ v0.27 + 1
 
 ### API additions
 
+* You can now swap out memory allocators via the
+  `GIT_OPT_SET_ALLOCATOR` option with `git_libgit2_opts()`.
+
 ### API removals
 
 ### Breaking API changes
--- a/include/git2/common.h
+++ b/include/git2/common.h
@@ -183,6 +183,7 @@ typedef enum {
 	GIT_OPT_GET_WINDOWS_SHAREMODE,
 	GIT_OPT_SET_WINDOWS_SHAREMODE,
 	GIT_OPT_ENABLE_STRICT_HASH_VERIFICATION,
+	GIT_OPT_SET_ALLOCATOR
 } git_libgit2_opt_t;
 
 /**
@@ -344,6 +345,12 @@ typedef enum {
  *		> objects from disk. This may impact performance due to an
  *		> additional checksum calculation on each object. This defaults
  *		> to enabled.
+ *
+ *	 opts(GIT_OPT_SET_ALLOCATOR, git_allocator *allocator)
+ *
+ *		> Set the memory allocator to a different memory allocator. This
+ *		> allocator will then be used to make all memory allocations for
+ *		> libgit2 operations.
  *
  * @param option Option key
  * @param ... value to set the option
--- a/include/git2/sys/alloc.h
+++ b/include/git2/sys/alloc.h
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#ifndef INCLUDE_sys_git_alloc_h__
+#define INCLUDE_sys_git_alloc_h__
+
+#include "git2/common.h"
+
+GIT_BEGIN_DECL
+
+/**
+ * An instance for a custom memory allocator
+ *
+ * Setting the pointers of this structure allows the developer to implement
+ * custom memory allocators. The global memory allocator can be set by using
+ * "GIT_OPT_SET_ALLOCATOR" with the `git_libgit2_opts` function. Keep in mind
+ * that all fields need to be set to a proper function.
+ */
+typedef struct {
+	/* Allocate `n` bytes of memory */
+	void *(*gmalloc)(size_t n, const char *file, int line);
+
+	/*
+	 * Allocate memory for an array of `nelem` elements, where each element
+	 * has a size of `elsize`. Returned memory shall be initialized to
+	 * all-zeroes
+	 */
+	void *(*gcalloc)(size_t nelem, size_t elsize, const char *file, int line);
+
+	/* Allocate memory for the string `str` and duplicate its contents. */
+	char *(*gstrdup)(const char *str, const char *file, int line);
+
+	/*
+	 * Equivalent to the `gstrdup` function, but only duplicating at most
+	 * `n + 1` bytes
+	 */
+	char *(*gstrndup)(const char *str, size_t n, const char *file, int line);
+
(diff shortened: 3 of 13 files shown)


arthurschreiber:
@pks-t Thank you for preparing this! ❤️ 

This definitely looks better than what I prepared over in #4574. 👍

But there's one thing I don't like so much about this: I really just want to replace the low-level `malloc`, `calloc`, `realloc` and `free`. I do not actually care about the other functions like `mallocarray`, as they are implemented on top of the former. I also don't really to have to re-implement the error handling for OOM. As I think `rugged` (and maybe `pygit2`) will be the only consumers of this for the foreseeable future, YAGNI might be a good guiding principle here. 😅 

The other thing is the compile-time flag. Can we benchmark this change, and if it's not an introducing extremely annoying performance issues, just make pluggable allocators always available?

Also, I agree, it would be cool if the CRTDBG stuff could be built on top of this, somehow. 👍

pks-t:
@carlosmn made the proposal of providing default `strdup`, `strndup` etc functions. So while the library may overwrite them, as well, they wouldn't have to.

And I do agree that the compile-time flag is ugly. My micro-benchmarks didn't turn up any notable performance penalties, but I didn't yet do some meaningful benchmarks. Together with those benchmarks we should probably also test if it has any noticable performance hit if all of these functions do have two additional parameters for __LINE__ and __FILE__. Passing these in could even prove to be valuable for developers using libgit2, as they could hook up libgit2 with their own memory tracing implementations

pks-t:
My tests:

```
#include "clar_libgit2.h"

void test_core_alloc__malloc_small(void)
{
	size_t i;
	void *mem;

	for (i = 0; i < 100000000; i++) {
		mem = git__malloc(100);
		git__free(mem);
	}
}

void test_core_alloc__malloc_medium(void)
{
	size_t i;
	void *mem;

	for (i = 0; i < 100000000; i++) {
		mem = git__malloc(100000);
		git__free(mem);
	}
}

void test_core_alloc__malloc_many_huge(void)
{
	size_t i;
	void *mem;

	for (i = 0; i < 1000000; i++) {
		mem = git__malloc(100000000);
		git__free(mem);
	}
}
```

Benchmarks are best of ten, with "build" being the normal build and "build-pluggable" being a build with pluggable allocators:

```
build,malloc_small,4539779566
build,malloc_medium,5912927186
build,malloc_many_huge,5166935308
build-pluggable,malloc_small,4611074505
build-pluggable,malloc_medium,5979185308
build-pluggable,malloc_many_huge,5388776352
```

malloc_small: 1.5% performance degradation
malloc_medium: 1.1% performance degradation
malloc_many_huge: 4.2% performance degradation

arthurschreiber:
> malloc_many_huge: 4.2% performance degradation

Huh. Shouldn't this actually be the benchmark that gets least impacted by the changes? 🤔

pks-t:
And with all functions accepting `__LINE__` and `__FILE__`:

```
build-fileline,malloc_small,4588338192
build-fileline,malloc_medium,6004951910
build-fileline,malloc_many_huge,4942528135
```

Summarized:

```
|------------|-------------------|-------------------|-------------------|
| build/test | small alloc       | medium alloc      | big alloc         |
|------------|-------------------|-------------------|-------------------|
| standard   | 4539779566, +0.0% | 5912927186, +0.0% | 5166935308, +0.0% |
|------------|-------------------|-------------------|-------------------|
| pluggable  | 4611074505, +1.5% | 5979185308, +1.1% | 5388776352, +4.2% |
|------------|-------------------|-------------------|-------------------|
| fileline   | 4588338192, +1.1% | 6004951910, +1.5% | 4942528135, -4.4% |
|------------|-------------------|-------------------|-------------------|
```

You can already see that this is no rocket sience by the fact that the `fileline` allocator actually is faster than the standard non-pluggable allocator...

pks-t:
@arthurschreiber: yeah, I've been very surprised by that. I guess when it comes to bigger allocations there's a lot of variation due to them being passed to the kernel.

ethomson:
It looks like you've dropped the ability to toggle this at build time.  This means that we're now pluggable all the time, with a 1.5% speed hit in the best case (which is also the typical case)?

pks-t:
Well, I wouldn't say in the best case but in the worst case where our complete program consists of nothing else but `malloc`s. I can restore the build toggle again if you feel like it is worth the additional code, but I feel like that's not really necessary

ethomson:
@carlosmn, how do you feel about the perf impact here?

@pks-t Can you add a note to the CHANGELOG?

pks-t:
On Mon, Apr 16, 2018 at 02:41:03PM +0000, Edward Thomson wrote:
> @carlosmn, how do you feel about the perf impact here?
> 
> @pks-t Can you add a note to the CHANGELOG?

I've updated the changelog.

carlosmn:
> how do you feel about the perf impact here?

I don't love it, but for the most part we're IO or CPU-bound, so `malloc` being slightly slower is probably worth the trade-off. If someone really needs to disable it, we can bring back a switch.

ethomson:
> I don't love it, but for the most part we're IO or CPU-bound, so malloc being slightly slower is probably worth the trade-off. If someone really needs to disable it, we can bring back a switch.

OK.  That's roughly how I feel.  :shipit:

pks-t:
So is this ready to be merged then?

pks-t:
So I'm merging this for now. As you said, in case somebody complains about a possible performance impact, then we can still create a build switch for this feature.

Issue #2658, opened 2014-10-26
coalesce() is never called within git_blame__like_git()


swisspol:
Found with Clang warnings:

```
void git_blame__like_git(git_blame *blame, uint32_t opt)
{
    while (true) {
        git_blame__entry *ent;
        git_blame__origin *suspect = NULL;

        /* Find a suspect to break down */
        for (ent = blame->ent; !suspect && ent; ent = ent->next)
            if (!ent->guilty)
                suspect = ent->suspect;
        if (!suspect)
            return; /* all done */

        /* We'll use this suspect later in the loop, so hold on to it for now. */
        origin_incref(suspect);
        pass_blame(blame, suspect, opt);

        /* Take responsibility for the remaining entries */
        for (ent = blame->ent; ent; ent = ent->next) {
            if (same_suspect(ent->suspect, suspect)) {
                ent->guilty = true;
                ent->is_boundary = !git_oid_cmp(
                        git_commit_id(suspect->commit),
                        &blame->options.oldest_commit);
            }
        }
        origin_decref(suspect);
    }

    coalesce(blame);  <----------- NEVER EXECUTED
}
```

ethomson:
Fixed via #4580

PR #4580, opened 2018-03-16
blame_git: fix coalescing step never being executed

pks-t:
Since blame has been imported from git.git and had its first share of
refactorings in b6f60a4d9 (Clean up ported code, 2013-09-21), the code
is actually not doing the coalescing step of the generated blame. While
the code to do the coalescing does exist, it is never being called as
the function `git_blame__like_git` will directly return from its `while
(true)` loop.

The function that was being imported from git.git was the `assign_blame`
function from "builtin/blame.c" from 717d1462b (git-blame --incremental,
2007-01-28), which hasn't really changed much. Upon taking an initial
look, one can seet hat `coalesce` is actually never getting called in
`assign_blame`, as well, so one may assume that not calling `coalesce`
by accident is actually the right thing. But it is not, as `coalesce` is
being called ever since cee7f245d (git-pickaxe: blame rewritten.,
2006-10-19) after the blame has been done in the caller of
`assign_blame`. Thus we can conclude the code of libgit2 is actually
buggy since forever.

To fix the issue, simply break out of the loop instead of doing a direct
return. Note that this does not alter behaviour in any way visible to
our tests, which is unfortunate. But in order to not diverge from what
git.git does, I'd rather adapt to how it is being done upstream in order
to avoid breaking certain edge cases than to just remove that code.

### DIFF
--- a/src/blame_git.c
+++ b/src/blame_git.c
@@ -623,6 +623,8 @@ static void coalesce(git_blame *blame)
 
 int git_blame__like_git(git_blame *blame, uint32_t opt)
 {
+	int error = 0;
+
 	while (true) {
 		git_blame__entry *ent;
 		git_blame__origin *suspect = NULL;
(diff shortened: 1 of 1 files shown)


pks-t:
Fixes #2658. If anybody could come up with a test I'd be happy to include that. Still trying to wrap my head around to our blaming implementation

ethomson:
Thanks for digging in here; I agree that tests would be nice, but this is a strict improvement anyway.

Issue #1146, opened 2012-12-16
There is a flaw in the handling of custom backends


jwiegley:
Higher priority custom backends, if unable to perform a certain action, are supposed to automatically fall back to a lower priority backend.  The logic in `git_odb_write`, for example, indicates this.

However, `git_blob_create_frombuffer` does not call `git_odb_write`, but only `git_odb_open_wstream`, which -- if stream writing is not supported -- creates a fake stream using _the highest priority backend that supports writestream or write_.

The flaw here is that if the write method backing the fake stream fails (expecting an automatic fallback), no fallback occurs because there is no logic to create a new fake stream using the next most applicable backend.

In short, the above makes it impossible to create a tracing backend for writes.  If you try, you end up not being able to write objects.

Since I need this kind of intelligent fallback for regular writes I will fix the Blob behavior and submit a pull request.  I wanted to track the issue here to get comments and to see if anyone has already solved this.

PR #4581, opened 2018-03-16
ODB write stream backend prioritization

pks-t:
When opening a write stream for the ODB, we loop through each backend
and try to either open it as a normal write stream or, in case
`writestream` has not been implented by the backend, a fake write
stream. The fake write stream simply wraps the backend's `write`
callback. This has the problem though that even if there is a backend
with write streams being supported, we will never use them as long as
there is another backend without a `writestream`, but a `write`
function. This is kind of an awkward interface to use, as it effectively
hinders you to use a lower-prioritized `writestream` only backend for
e.g. streaming out large objects as soon as you have another backend
with simple write support.

Rewrite the logic to make two passes over the list of backends. In the
first pass, we will for each of the backends try to use the
`writestream` function, if it exists. Only in the second pass will we
now try to open a fake write stream. This effectively gives backends
implementing `writestream` a higher priority as every other backend only
implementing `write`. As the most important purpose of using streams is
to help in writing huge objects, this is the right thing to do, as we do
not want to silently fall back on `write`.

---

Inspired by #1146. I don't really know if this PR fixes his request, but I feel like prioritizing any backend which implements `writestream` should be the correct thing to do. This is probably up for discussion, though, as I can see arguments for both sides. Anyway, I'd find it rather unexpected in the case where I e.g. have two backends, one high priority one with write and one with writestream, only, to see that the first one would be opened.

So I'm on the edge, please help :(

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -1353,12 +1353,14 @@ int git_odb_open_wstream(
 		}
 	}
 
+	if (!writes) {
+		error = git_odb__error_unsupported_in_backend("write object");
+		goto done;
+	}
+
 	if (error < 0) {
 		if (error == GIT_PASSTHROUGH)
 			error = 0;
-		else if (!writes)
-			error = git_odb__error_unsupported_in_backend("write object");
-
 		goto done;
 	}
 
--- a/src/odb.c
+++ b/src/odb.c
@@ -1340,17 +1340,27 @@ int git_odb_open_wstream(
 		backend_internal *internal = git_vector_get(&db->backends, i);
 		git_odb_backend *b = internal->backend;
 
-		/* we don't write in alternates! */
-		if (internal->is_alternate)
+		if (internal->is_alternate || b->writestream == NULL)
 			continue;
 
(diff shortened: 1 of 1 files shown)


ethomson:
My thoughts are that whether a backend can stream or not is an implementation detail between libgit2 and the backend.  As a user, I shouldn't care.  I want my highest priority backend to get the write, and I don't care how.  It would be very unsurprising to me if I went to the trouble of setting up two backends, one with a higher priority than the other, and the lower priority one was getting the writes.

pks-t:
Yeah, I can definitely see where you're coming from. That's why I initially said that I can see good arguments on both sides.

What is a bit weird though is that we're being asymmetric. We do create a fake stream from the `write` method in case `writestream` is not implemented. But we do not create a fake write from `writestream` in case `write` is not implemented. That feels inconsistent, as we're not really paying attention to the priority in all cases. That being said, the constellation of `writestream` without `write` will probably never happen.

PR #4584, opened 2018-03-19
online::clone: skip creds fallback test

ethomson:
At present, we have three online tests against bitbucket: one which
specifies the credentials in the payload, one which specifies the
correct credentials in the URL and a final one that specifies the
incorrect credentials in the URL.  Bitbucket has begun responding to the
latter test with a 403, which causes us to fail.

Break these three tests into separate tests so that we can skip the
latter until this is resolved on Bitbucket's end or until we can change
the test to a different provider.

### DIFF
--- a/tests/online/clone.c
+++ b/tests/online/clone.c
@@ -357,15 +357,45 @@ void test_online_clone__bitbucket_style(void)
 	cl_git_pass(git_clone(&g_repo, BB_REPO_URL, "./foo", &g_options));
 	git_repository_free(g_repo); g_repo = NULL;
 	cl_fixture_cleanup("./foo");
+}
 
-	/* User and pass from URL */
-	user_pass.password = "wrong";
+void test_online_clone__bitbucket_uses_creds_in_url(void)
+{
+	git_cred_userpass_payload user_pass = {
+		"libgit2", "wrong"
+	};
+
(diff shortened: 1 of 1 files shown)


ethomson:
I don't love resorting to this, but I was hoping that this was just a temporary problem for bitbucket (it seems like it only affects some requests).  I opened #4585 to ensure that we didn't forget about this.

tests/online/clone.c
```diff
@@ -357,15 +357,41 @@ void test_online_clone__bitbucket_style(void)
...
+void test_online_clone__bitbucket_falls_back_to_specified_creds(void)
+{
+	git_cred_userpass_payload user_pass = {
+		"libgit2", "libgit2"
+	};
+
+	g_options.fetch_opts.callbacks.credentials = git_cred_userpass;
+	g_options.fetch_opts.callbacks.payload = &user_pass;
+
+	cl_skip();
```
pks-t: Could you please add a comment stating why this test is skipped? It's easy enough to dig it up via git blame, but I think it's good to have anyway

pks-t:
The fix is not ideal, but I guess we cannot do any better without known breakages from #4500, which I definitely don't want to merge for v0.27.0. So I agree with your solution here

tests/online/clone.c
ethomson: Added a comment.

PR #4586, opened 2018-03-19
Add mailmap support.

### LINKED ISSUE #4565: Mailmap support
Would contributions adding some sort of mailmap-querying API be welcome?

If so, how should that API look like?

Thanks!

---

### Reproduction steps

Try to use libgit2 on a repo with mailmap enabled.

### Expected behavior

There's some sort of API to do a mailmap lookup.

### Actual behavior

There's no such API (I don't see a single reference to `mailmap` on the source either, modulo the mailmap in this repo itself).

2d2a602

N / A

emilio:
Fixes #4565.


emilio:
cc @mystor, who ended up doing most of the work :)

mystor:
I figure I'll add a high level summary of the API @emilio  and I ended up making here:

1. There is a new `git_mailmap` opaque struct which can be created from either a repository, or a byte buffer. It contains mailmap entries.
  a) When creating from a bare repository, the .mailmap file will be looked up in the root tree of HEAD
  b) When creating from a non-bare repository, the working directory will be checked for the .mailmap.
2. This struct provides methods for accessing the individual entries (`git_mailmap_entry_byindex`), and resolving name/email pairs to real names/emails (`git_mailmap_resolve`).
3. The mailmap APIs are also integrated into signatures, commits, and blame
  a) `git_signature_with_mailmap` resolves a signature to the real name/email.
  b) `git_commit_{committer,author}_with_mailmap` gets the committer/author's real name/email by using the given mailmap.
  c) A new flag, `GIT_BLAME_USE_MAILMAP` was added to the blame options. When specified, the mailmap from the repo will be used to ensure that the signatures in the blame chunk output use real names/emails.

This _should_ be a fully non-breaking change, as it exclusively adds APIs, and doesn't change the behaviour of any existing functions. Mailmaps are never used implicitly in the API, and have to be requested.

I also added a bunch of tests into tests/mailmap/**.

include/git2/mailmap.h
```diff
@@ -0,0 +1,127 @@
...
+ * Create a mailmap object by parsing a mailmap file.
+ *
+ * The mailmap must be freed with 'git_mailmap_free'.
+ *
+ * @param out pointer to store the mailmap
+ * @param data raw data buffer to parse
+ * @param size size of the raw data buffer
+ * @return 0 on success
+ */
+GIT_EXTERN(int) git_mailmap_parse(
```
pks-t: To be consistent with `git_mailmap_from_repo`, a function name `git_mailmap_from_buffer` might in fact be favorable. This is only a proposal, the name is fine as is.

include/git2/mailmap.h
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Free a mailmap created by 'git_mailmap_parse' or 'git_mailmap_from_repo'.
+ */
+GIT_EXTERN(void) git_mailmap_free(git_mailmap *mailmap);
+
+/**
+ * Resolve a name and email to the corresponding real name and email.
+ *
+ * @param name_out either 'name', or the real name to use.
+ *             You should NOT free this value.
```
pks-t: So you're simply returning either the original name/email or the mailmap-allocated ones? If so, you should state that lifetime of the out parameters is tied to mailmap and name/email

include/git2/mailmap.h
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Lookup a mailmap entry by name/email pair.
+ *
+ * Do not free the mailmap entry, it is owned by the mailmap.
+ *
+ * @param mailmap the mailmap to perform the lookup in. (may be NULL)
+ * @param name the name to perform the lookup with.
+ * @param email the email to perform the lookup with.
+ * @return the corresponding mailmap entry, or NULL if it cannot be found.
+ */
```
pks-t: Can either `name` or `email` be `NULL`?

include/git2/mailmap.h
```diff
@@ -0,0 +1,127 @@
...
+ * be checked for the '.mailmap' file to be parsed.
+ *
+ * If the repository is bare, the repository's HEAD commit's tree root will be
+ * searched for the '.mailmap' file to be parsed.
+ *
+ * The mailmap must be freed with 'git_mailmap_free'.
+ *
+ * @param out pointer to store the mailmap
+ * @param repo repository to find the .mailmap in
+ * @return 0 on success; GIT_ENOTFOUND if .mailmap could not be found.
```
pks-t: Otherwise an error code?

include/git2/mailmap.h
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Create a mailmap object by parsing a mailmap file.
+ *
+ * The mailmap must be freed with 'git_mailmap_free'.
+ *
+ * @param out pointer to store the mailmap
+ * @param data raw data buffer to parse
+ * @param size size of the raw data buffer
+ * @return 0 on success
```
pks-t: Otherwise an error code?

src/blame.h
```diff
@@ -84,6 +84,8 @@ struct git_blame {
 	int num_lines;
 	const char *final_buf;
 	git_off_t final_buf_size;
+
+	git_mailmap *mailmap;
```
pks-t: You should probably move this just after `repository` to avoid unnecessary padding

src/mailmap.c
```diff
@@ -0,0 +1,381 @@
...
+#include "blob.h"
+#include "commit.h"
+#include "git2/common.h"
+#include "git2/repository.h"
+#include "git2/revparse.h"
+#include "git2/sys/commit.h"
+
+/**
+ * Helper type and methods for the mailmap parser
+ */
```
pks-t: No need for Javadoc-style comments in implementations.

Furthermore, we do have a rather new `git_parse` interface now which should preferably be used to parse data. We want to avoid duplicating common parser functions and structures as much as possible. Have you considered using that interface?

src/mailmap.c
```diff
@@ -0,0 +1,381 @@
...
+	int error = 0;
+
+	if (memchr(data, '\0', size) != NULL)
+		return -1; /* data may not contain '\0's */
+
+	*mailmap = git__calloc(1, sizeof(git_mailmap));
+	if (!*mailmap)
+		return -1;
+
+	/* XXX: Is it worth it to precompute the size? */
```
pks-t: Probably not. We're always expanding vectors by a factor of 1.5 every time, so it's not like every entry will cause a new allocation

src/mailmap.c
```diff
@@ -0,0 +1,381 @@
...
+	char_range file = { data, size };
+	git_mailmap_entry* entry = NULL;
+	int error = 0;
+
+	if (memchr(data, '\0', size) != NULL)
+		return -1; /* data may not contain '\0's */
+
+	*mailmap = git__calloc(1, sizeof(git_mailmap));
+	if (!*mailmap)
+		return -1;
```
pks-t: There's `GITERR_CHECK_ALLOC` for that

src/mailmap.c
```diff
@@ -0,0 +1,381 @@
...
+}
+
+int git_mailmap_parse(
+	git_mailmap **mailmap,
+	const char *data,
+	size_t size)
+{
+	char_range file = { data, size };
+	git_mailmap_entry* entry = NULL;
+	int error = 0;
```
pks-t: No need to initiailze error. You always set it before its being used

src/mailmap.c
```diff
@@ -0,0 +1,381 @@
...
+ * returns a pointer to the copied string, or NULL.
+ */
+static const char *range_copyz(char **buf, size_t *size, char_range src)
+{
+	char *s = NULL;
+	if (src.p == NULL)
+		return NULL;
+
+	if (size)
+		*size += src.len + 1;
```
pks-t: You'd have to check for overflows here. Use e.g. `GITERR_CHECK_ALLOC_ADD`

src/mailmap.c
```diff
@@ -0,0 +1,381 @@
...
+
+	/* XXX: Is it worth it to precompute the size? */
+	error = git_vector_init(&(*mailmap)->entries, 0, NULL);
+	if (error < 0)
+		goto cleanup;
+
+	while (file.len > 0) {
+		bool found = false;
+		char_range real_name, real_email, replace_name, replace_email;
+		size_t size = 0;
```
pks-t: The initialization here is useless

src/mailmap.c
```diff
@@ -0,0 +1,381 @@
...
+}
+
+size_t git_mailmap_entry_count(const git_mailmap *mailmap)
+{
+	if (mailmap)
+		return git_vector_length(&mailmap->entries);
+	return 0;
+}
+
+static int git_mailmap_from_bare_repo(
```
pks-t: You should drop the `git_` prefix from static functions

src/mailmap.c
```diff
@@ -0,0 +1,381 @@
...
+		goto cleanup;
+
+	error = git_reference_peel(&tree, head, GIT_OBJ_TREE);
+	if (error < 0)
+		goto cleanup;
+
+	error = git_object_lookup_bypath(
+		(git_object **) &blob,
+		tree,
+		".mailmap",
```
pks-t: A define in the implementation for this string would be nice

src/mailmap.c
```diff
@@ -0,0 +1,381 @@
...
+
+cleanup:
+	git_reference_free(head);
+	git_object_free(tree);
+	git_blob_free(blob);
+
+	return error;
+}
+
+static int git_mailmap_from_workdir_repo(
```
pks-t: Here as well, dropping the "git_" prefix is the way to go

src/signature.c
```diff
@@ -121,6 +122,44 @@ int git_signature_dup(git_signature **dest, const git_signature *source)
...
+	const git_mailmap *mailmap)
+{
+	git_signature *signature = NULL;
+	const char *name = NULL;
+	const char *email = NULL;
+
+	if (source == NULL)
+		goto on_error;
+
+	git_mailmap_resolve(&name, &email, mailmap, source->name, source->email);
```
pks-t: Shouldn't you check for errors here? While `GIT_ENOTFOUND` is okay, you might for exampleg et allocation errors

src/signature.c
```diff
@@ -121,6 +122,44 @@ int git_signature_dup(git_signature **dest, const git_signature *source)
...
+
+	signature = git__calloc(1, sizeof(git_signature));
+	if (!signature)
+		goto on_error;
```
pks-t: For these three cases, simply use `GITERR_CHECK_ALLOC`. Our current policy is to ignore memory leaks when we experience OOM-conditions

tests/clar_libgit2.c
```diff
@@ -222,7 +226,9 @@ git_repository *cl_git_sandbox_reopen(void)
 		git_repository_free(_cl_repo);
 		_cl_repo = NULL;
 
-		cl_git_pass(git_repository_open(&_cl_repo, _cl_sandbox));
+		cl_git_pass(git_repository_open(
+			&_cl_repo,
+			cl_fixture_basename(_cl_sandbox)));
```
pks-t: Why is this change necessary?

tests/mailmap/basic.c
```diff
@@ -0,0 +1,98 @@
...
+	"<email@foo.com> Other Name <yetanotheremail@foo.com>\n";
+
+void test_mailmap_basic__initialize(void)
+{
+	cl_git_pass(git_mailmap_parse(&mailmap, TEST_MAILMAP, sizeof(TEST_MAILMAP) - 1));
+}
+
+void test_mailmap_basic__cleanup(void)
+{
+	if (mailmap) {
```
pks-t: No need for this condition. `git_mailmap_free` should and does cope with `NULL`s

tests/mailmap/blame.c
```diff
@@ -0,0 +1,69 @@
...
+void test_mailmap_blame__initialize(void)
+{
+	g_repo = NULL;
+	g_blame = NULL;
+}
+
+void test_mailmap_blame__cleanup(void)
+{
+	cl_git_sandbox_cleanup();
+	g_repo = NULL;
```
pks-t: No need for these assignments as you already assign `NULL` in the initializer

tests/mailmap/blame.c
```diff
@@ -0,0 +1,69 @@
...
+
+	git_blame_free(g_blame);
```
pks-t: You should probably free the blame before freeing the repo

tests/mailmap/parsing.c
```diff
@@ -0,0 +1,123 @@
...
+	size_t idx = 0;
+
+	/* Check that the resolver behaves correctly */
+	for (idx = 0; idx < resolved_size; ++idx) {
+		git_mailmap_resolve(
+			&resolved_name,
+			&resolved_email,
+			mailmap,
+			resolved[idx].replace_name,
+			resolved[idx].replace_email);
```
pks-t: I've seen this several times now, with overly many newlines. We usually don't use that many newlines throughout our codebase. The only thing we try to stick to is a maximum line length of 80 characters

tests/mailmap/parsing.c
```diff
@@ -0,0 +1,123 @@
...
+static void check_mailmap_resolve(
+	const git_mailmap *mailmap, const mailmap_entry *resolved, size_t resolved_size)
+{
+	const char *resolved_name = NULL;
+	const char *resolved_email = NULL;
+	size_t idx = 0;
+
+	/* Check that the resolver behaves correctly */
+	for (idx = 0; idx < resolved_size; ++idx) {
+		git_mailmap_resolve(
```
pks-t: You should verify the return code with `cl_git_pass`

tests/mailmap/parsing.c
```diff
@@ -0,0 +1,123 @@
...
+			resolved[idx].replace_name,
+			resolved[idx].replace_email);
+		cl_assert_equal_s(resolved_name, resolved[idx].real_name);
+		cl_assert_equal_s(resolved_email, resolved[idx].real_email);
+	}
+}
+
+void test_mailmap_parsing__string(void)
+{
+	cl_check_pass(git_mailmap_parse(
```
pks-t: You shouldn't use `cl_check_pass` but `cl_git_pass`

tests/resources/mailmap/.gitted/hooks/applypatch-msg.sample
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
pks-t: You should probably remove all these hooks just to keep the churn as small as necessary

include/git2/mailmap.h
mystor: No, although theoretically that could be handled.

src/mailmap.c
mystor: I wasn't aware that the `git_parse` interface existed :-). I grepped around the codebase looking at existing parsers & saw that they seemed to be implemented in an ad-hoc way, so did the same here.

I'll look into using that interface instead.

src/signature.c
mystor: Good to know! I was confused by existing consumers of GITERR_CHECK_ALLOC because they often seemed to leak memory, but that makes a lot of sense.

tests/clar_libgit2.c
mystor: When I was writing the tests for this patch, I wanted to have a single repository which I opened both in a bare configuration, without a working directory, and as a non-bare repository. To do this I tried to use `cl_git_sandbox_init`. Unfortunately, that code doesn't currently handle "nested/paths/to/fixtures". The clar fixture code copies just the leaf node of the path into the sandbox directory (e.g. just the `.gitted` folder), but `cl_git_sandbox_init` changes directory to the sandbox directory, and tries to open the full fixture path as a repository.

This change makes `cl_git_sandbox_init` use only the leaf name of the fixture's path when determining what the path will be relative to the sandbox.

src/signature.c
mystor: git_mailmap_resolve returns `void`.

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+	char c;
+	return git_parse_peek(&c, ctx, GIT_PARSE_PEEK_SKIP_WHITESPACE) < 0 || c == '#';
+}
+
+static int advance_until(
+	const char **start, size_t *len, git_parse_ctx *ctx, char needle)
+{
+	*start = ctx->line;
+	while (ctx->line_len > 0 && *ctx->line != '#' && *ctx->line != needle)
+		git_parse_advance_chars(ctx, 1);
```
pks-t: I think an interface `git_parse_advance_char(&c, ctx, 1);` would be nice to have. Advance one character and put the next one into `c`.

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+	if (ctx->line_len == 0 || *ctx->line == '#')
+		return -1; /* end of line */
+
+	*len = ctx->line - *start;
+	git_parse_advance_chars(ctx, 1); /* advance past needle */
+	return 0;
+}
+
+/* Parse a single entry from a mailmap file.
+ *
```
pks-t: Multi-line comments should start out with an empty `/*`:

```
/*
 * Parse a...
 */
```

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+
+	git_buf_clear(real_name);
+	git_buf_clear(real_email);
+	git_buf_clear(replace_name);
+	git_buf_clear(replace_email);
+
+	/* Parse the real name */
+	git_parse_advance_ws(ctx);
+	if (advance_until(&start, &len, ctx, '<') < 0)
+		return -1;
```
pks-t:
```
if (...) {
    giterr_set(GITERR_MAILMAP, "could not parse real name");
    return -1;
}
```

You'd have to define `GITERR_MAILMAP`, but I think it can be helpful to detail what's wrong here

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+void git_mailmap_free(git_mailmap *mailmap)
+{
+	if (!mailmap)
+		return;
+
+	git_vector_free_deep(&mailmap->entries);
+	git__free(mailmap);
+}
+
+void git_mailmap_resolve(
```
pks-t: You should have a return type here. While the function cannot fail right now, you never know how it'll have to change in the future. So you should already build all callers like the function _could_ fail, just to be safe in the future

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+
+/* Check if we're at the end of line, w/ comments */
+static bool is_eol(git_parse_ctx *ctx)
+{
+	char c;
+	return git_parse_peek(&c, ctx, GIT_PARSE_PEEK_SKIP_WHITESPACE) < 0 || c == '#';
+}
+
+static int advance_until(
+	const char **start, size_t *len, git_parse_ctx *ctx, char needle)
```
pks-t: I feel like this function would be a good addition to the parser interface

src/mailmap.c
pks-t: Would be cool to have a `git_parse_advance_char(&c, ctx, 1)` here to avoid messing with internals.

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+
+	if (ctx->line_len == 0 || *ctx->line == '#')
+		return -1; /* end of line */
+
+	*len = ctx->line - *start;
+	git_parse_advance_chars(ctx, 1); /* advance past needle */
+	return 0;
+}
+
+/* Parse a single entry from a mailmap file.
```
pks-t: Multi-line comments should always start out with an empty `/*`:

```
/*
 * Parse...
 */
```

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+	/* Parse the real name */
+	git_parse_advance_ws(ctx);
+	if (advance_until(&start, &len, ctx, '<') < 0)
+		return -1;
+
+	git_buf_attach_notowned(real_name, start, len);
+	git_buf_rtrim(real_name);
+
+	/* If this is the last email in the line, this is the email to replace,
+	 * otherwise, it's the real email. */
```
pks-t: One more non-conforming multiline comment ;)

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+		goto cleanup;
+
+	/* Run the parser */
+	while (ctx.remain_len > 0) {
+		error = parse_mailmap_entry(
+			&real_name, &real_email, &replace_name, &replace_email, &ctx);
+		if (error < 0) {
+			error = 0; /* Skip lines which don't contain a valid entry */
+			git_parse_advance_line(&ctx);
+			continue;
```
pks-t: This here would in fact be a cool use for the warnings API (which doesn't exist yet, so nothing you need to change)

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+		if (real_email.size > 0) {
+			memcpy(entry_data, real_email.ptr, real_email.size);
+			entry->real_email = entry_data;
+			entry_data += real_email.size + 1;
+		}
+		if (replace_name.size > 0) {
+			memcpy(entry_data, replace_name.ptr, replace_name.size);
+			entry->replace_name = entry_data;
+			entry_data += replace_name.size + 1;
+		}
```
pks-t: Hm. While it's nice to avoid four `malloc`s, I don't know whether that's really needed. It's probably not like parsing the mailmap will be on a hot path, so I rather tend to go for simplicity and simply copy the strings.

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+	size_t i;
+	git_mailmap_entry *entry;
+	assert(name && email);
+
+	if (!mailmap)
+		return NULL;
+
+	git_vector_foreach(&mailmap->entries, i, entry) {
+		if (!git__strcmp(email, entry->replace_email) &&
+		    (!entry->replace_name || !git__strcmp(name, entry->replace_name))) {
```
pks-t: I feel like it'd be a bit easier to read if you just split this up into multiple if statements:

```
if (git__strcmp(email, entry->replace_email))
    continue;
if (entry->replace_name && git__strcmp(name, entry->replace_name))
    continue;
return entry;
```

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+	error = git_repository_head(&head, repo);
+	if (error < 0)
+		goto cleanup;
+
+	error = git_reference_peel(&tree, head, GIT_OBJ_TREE);
+	if (error < 0)
+		goto cleanup;
+
+	error = git_object_lookup_bypath(
+		(git_object **) &blob, tree, MAILMAP_FILE, GIT_OBJ_BLOB);
```
pks-t: You can avoid the cast by simply using `&blob.object`

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+	if (error < 0)
+		goto cleanup;
+
+	error = git_blob_filtered_content(&content, blob, MAILMAP_FILE, false);
```
pks-t: I don't really think it necessary to have that blob filtered. Just using `git_blob__getbuf(&content, blob)` should be fine. Or are there people who define a custom filter that writes the mailmap dynamically... :scream: Users can get creative, but until somebody complains I'd just go for the unfiltered content.

src/mailmap.c
```diff
@@ -0,0 +1,335 @@
...
+	int error;
+
+	assert(git_repository_is_bare(repo));
+
+	/* In bare repositories, fall back to reading from HEAD's tree */
+	error = git_repository_head(&head, repo);
+	if (error < 0)
+		goto cleanup;
+
+	error = git_reference_peel(&tree, head, GIT_OBJ_TREE);
```
pks-t: No need for this. `git_object_lookup_bypath` will peel the treeish itself

tests/clar_libgit2.c
```diff
@@ -171,13 +171,17 @@ static git_repository *_cl_repo = NULL;
 
 git_repository *cl_git_sandbox_init(const char *sandbox)
 {
+	/* Get the name of the sandbox folder which will be created
+	 */
```
pks-t: Another ugly comment :P

tests/mailmap/basic.c
```diff
@@ -0,0 +1,88 @@
...
+	"Foo bar <foo@bar.com> <foo@baz.com>  \n"
+	"Blatantly invalid line\n"
+	"Foo bar <foo@bar.com> <foo@bal.com>\n"
+	"<email@foo.com> <otheremail@foo.com>\n"
+	"<email@foo.com> Other Name <yetanotheremail@foo.com>\n";
+
+void test_mailmap_basic__initialize(void)
+{
+	git_buf buf = GIT_BUF_INIT;
+	git_buf_attach_notowned(&buf, TEST_MAILMAP, sizeof(TEST_MAILMAP) - 1);
```
pks-t: You can just use `strlen(TEST_MAILMAP)` here, which is a bit more obvious. The compiler should do it's thing and optimize that out

tests/mailmap/basic.c
```diff
@@ -0,0 +1,88 @@
...
+#include "git2/mailmap.h"
+
+static git_mailmap *mailmap = NULL;
+
+const char TEST_MAILMAP[] =
+	"Foo bar <foo@bar.com> <foo@baz.com>  \n"
+	"Blatantly invalid line\n"
+	"Foo bar <foo@bar.com> <foo@bal.com>\n"
+	"<email@foo.com> <otheremail@foo.com>\n"
+	"<email@foo.com> Other Name <yetanotheremail@foo.com>\n";
```
pks-t: I don't think this one is valid. At least git-check-mailmap does not mention this syntax

tests/mailmap/basic.c
```diff
@@ -0,0 +1,88 @@
...
+
+	cl_assert(git_mailmap_entry_count(mailmap) == 4);
+
+	entry = git_mailmap_entry_byindex(mailmap, 0);
+	cl_assert(entry);
+	cl_assert(!entry->replace_name);
+	cl_assert(!git__strcmp(entry->replace_email, "foo@baz.com"));
+
+	entry = git_mailmap_entry_byindex(mailmap, 10000);
+	cl_assert(!entry);
```
pks-t: I'd put this one into a separate test. A test should only check one thing

tests/mailmap/basic.c
```diff
@@ -0,0 +1,88 @@
...
+void test_mailmap_basic__entry(void)
+{
+	const git_mailmap_entry *entry;
+
+	cl_assert(git_mailmap_entry_count(mailmap) == 4);
+
+	entry = git_mailmap_entry_byindex(mailmap, 0);
+	cl_assert(entry);
+	cl_assert(!entry->replace_name);
+	cl_assert(!git__strcmp(entry->replace_email, "foo@baz.com"));
```
pks-t: I'd love if you'd validate all four entries.

```
struct {
    char *real_name;
    char *real_mail;
    char *replace_name;
    char *replace_mail;
} expected[] = {
    { "Foo bar", "foo@bar.com", "foo@baz.com" },
   ...
};

for (i = 0; i < git_mailmap_entry_count(mailmap); i++) {
    entry_byindex(mailmap, i);
   assert_all_things();
}
```

tests/mailmap/basic.c
```diff
@@ -0,0 +1,88 @@
...
+		"doesnotexist@fo.com");
+	cl_assert(!entry);
+}
+
+void test_mailmap_basic__lookup(void)
+{
+	const git_mailmap_entry *entry = git_mailmap_entry_lookup(
+		mailmap, "Typoed the name once", "foo@baz.com");
+	cl_assert(entry);
+	cl_assert(!git__strcmp(entry->real_name, "Foo bar"));
```
pks-t: There's `cl_assert_equal_s`, for this one and the following tests

src/mailmap.c
ethomson: Actually, if you could add a `TODO: warn` then we will pick it up when we go through with the warning API (someday).

src/mailmap.c
ethomson: This raises a question: can we properly parse mailmap files with Windows-style line endings?  If not, then we should absolutely _not_ be using the filtered content.

I think that the goal is to use the mailmap as it would exist in the working directory, which makes sense.  But can git cope with mailmaps with Windows-style line endings?  Does it use the filtered content or the content directly from the repository?

pks-t:
```
$ echo -e 'test <foo@bar.com>\r\nother <other@bar.com>\r\n' | git check-mailmap --stdin
test <foo@bar.com>
other <other@bar.com>
fatal: unable to parse contact:
```

So yes, it does cope with Windows-style line endings. The last line is simply that it cannot parse an empty line.

pks-t: That being said, I think _we_ as in the libgit2 project cannot currently parse them with Windows-style endings.

ethomson: Let's make sure that we can, please.  👍

src/mailmap.c
mystor: We always ignore these parser errors & just skip the line (hence the potential desire for warnings). Should we still set giterr if we won't produce a visible error code?

src/mailmap.c
mystor: I believe that `head` is a `git_reference`, which I don't _think_ is a `git_object` which I can pass in and have peeled, though I may be wrong.

```
struct git_reference {
	git_refdb *db;
	git_ref_t type;

	union {
		git_oid oid;
		char *symbolic;
	} target;

	git_oid peel;
	char name[GIT_FLEX_ARRAY];
};
```
vs.
```
struct git_tree {
	git_object object;
	git_odb_object *odb_obj;
	git_array_t(git_tree_entry) entries;
};
```

tests/mailmap/basic.c
mystor: It's not mentioned, but I did test with git & this format is handled by git's internal parser.

src/mailmap.c
mystor: `blob` is a NULL pointer, not a `git_blob`, so I don't think I can do that.

tests/mailmap/basic.c
mystor: git's documentation doesn't mention this format, but when I was testing mailmap, IIRC this format is supported.

src/mailmap.c
pks-t: I realized that a bit later and thought to have deleted that comment. Existing code is fine as-is.

src/mailmap.c
pks-t: True

src/mailmap.c
pks-t: True

tests/mailmap/basic.c
pks-t: Okay, fair enough

mystor:
@pks-t So, I've ended up making a few more changes to the patch than I expected to. While I was looking at the documentation for git's mailmap support more closely I realized that I was looking up mailmap entries for a repository incorrectly. git has a specific way of doing it, as well as a set of configs (mailmap.blob and mailmap.file) which override each other in a particular order. Unfortunately, this behaviour was incompatible with the API we were exposing from the mailmap module.

I've rewritten the API to support the correct overriding behaviour, but it has caused a fairly large surface change. As an example, the individual entry objects are no longer part of the public API, and you can't index into the mailmap, as that doesn't make as much sense with the correct overriding behaviour. I'm mostly glad that I caught this before we merged it & had to make a breaking change >.<.

I decided to continue to implement the backing store as a sorted git_vector, rather than using a git_strmap or similar, because it was simpler to implement, and required less moving parts.

pks-t:
Push for myself such that this thread is further up in my mailbox and I remember to review this soon

tests/mailmap/blame.c
```diff
@@ -0,0 +1,66 @@
...
+	const git_blame_hunk *hunk = NULL;
+	git_blame_options opts = GIT_BLAME_OPTIONS_INIT;
+
+	g_repo = cl_git_sandbox_init("mailmap");
+
+	opts.flags |= GIT_BLAME_USE_MAILMAP;
+
+	cl_git_pass(git_blame_file(&g_blame, g_repo, "file.txt", &opts));
+	if (!g_blame)
+		return;
```
pks-t: This is unexpected. We just silently abort the test, but in fact we should always exactly know what to expect (and error if `g_blame` is not set)

tests/mailmap/blame.c
```diff
@@ -0,0 +1,66 @@
...
+{
+	size_t idx = 0;
+	const git_blame_hunk *hunk = NULL;
+	git_blame_options opts = GIT_BLAME_OPTIONS_INIT;
+
+	g_repo = cl_git_sandbox_init("mailmap");
+
+	cl_git_pass(git_blame_file(&g_blame, g_repo, "file.txt", &opts));
+	if (!g_blame)
+		return;
```
pks-t: Same here

tests/mailmap/mailmap_helpers.h
```diff
@@ -0,0 +1,44 @@
+#include "mailmap.h"
+
+typedef struct mailmap_entry {
+	const char *real_name;
+	const char *real_email;
+	const char *replace_name;
+	const char *replace_email;
+} mailmap_entry;
```
pks-t: "mailmap_helpers.h" is a bit weird, as there are no helpers in here. It's more like data, so "mailmap_testdata.h" would fit better

pks-t:
Thanks for the reminder, @mystor. I'll review this PR no later than this Friday

pks-t:
I'm happy with this PR. There's a memory leak introduced, though, as you don't ever free the `mm->entries` vector.

mystor:
@pks-t I've fixed the memory leak you pointed out. Sorry for the delay. Is there anything else I should be doing for this PR?

pks-t:
Great, CI doesn't mention any memory leaks anymore. @ethomson, got any more comments on this PR?

include/git2/mailmap.h
```diff
@@ -0,0 +1,131 @@
...
+ * it will be replaced with the new entry.
+ *
+ * @param mm mailmap to add the entry to
+ * @param real_name the real name to use, or NULL
+ * @param real_email the real email to use, or NULL
+ * @param replace_name the name to replace, or NULL
+ * @param replace_email the email to replace
+ * @return 0 if it was added, EEXISTS if it replaced an entry, or an error code
+ */
+GIT_EXTERN(int) git_mailmap_add_entry(
```
ethomson: I'm sorry for the last-minute review, but I don't love the `EEXISTS` return here.  That really makes it sound like an error case, not a success.  Do you think that "replaced an entry" is really such a valuable state to know?  If so, what do you think about an out-param that indicates that instead of overloading the return value?  Curious what @pks-t thinks.

mystor: The reasoning behind this decision was to match the git_vector APIs here. Most consumers of the mailmap API will probably just get it from a git_repository which just ignores errors like this.

I'd probably be fine with just not communicating whether or not an entry was replaced entirely.

tiennou: Maybe an `int force` parameter (instead of a int `*replaced`) ?

ethomson: > I'd probably be fine with just not communicating whether or not an entry was replaced entirely.

I'm +1 on this.

mystor:
@ethomson @pks-t I just hid the EEXISTS error from the public API - anything else I should change or is this patch ready to go now?

pks-t:
I'm still fine with this PR as is. :)

include/git2/mailmap.h
```diff
@@ -0,0 +1,131 @@
...
+ * Create a new mailmap instance containing a single mailmap file
+ *
+ * This method is a simple utility wrapper for the following sequence
+ * of calls:
+ *  - git_mailmap_new
+ *  - git_mailmap_add_buffer
+ *
+ * @param out pointer to store the new mailmap
+ * @param buf buffer to parse the mailmap from
+ * @param len the length of the input buffer [optional: 0 defaults to 'strlen']
```
ethomson: This is not a pattern that we have been using - our APIs either take a NUL-terminated string or a buffer with length.  (And if it's the latter, then `0` really means `0`.)  I think it would be confusing to have an API that differs from this pattern.

I don't have a strong opinion on which way this should go - it seems like we should just take a `buf`?  I'm happy either way but the `0` means we do the `strlen` doesn't align with the rest of our APIs.

include/git2/mailmap.h
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Parse mailmap entries from a buffer.
+ *
+ * @param mm mailmap to add the entries to
+ * @param buf the buffer to read the mailmap file from
+ * @param len the length of the input buffer [optional: 0 defaults to 'strlen']
+ * @return 0 on success, or an error code
+ */
+GIT_EXTERN(int) git_mailmap_add_buffer(
```
ethomson: I wonder how much utility there is in making this a public method?  It seems like most uses would prefer the `git_mailmap_from_buffer`.  I'd be perfectly happy seeing this be an internal API.

If it does stay public (which is fine, if you have a use for it or think others will) then I do think that the name does not capture exactly what it does (at least when viewed through the lens of the names of our current APIs - `add_buffer` sounds like it will add a buffer, not add _from_ a buffer).

I would propose `git_mailmap_parse`?  Also, the same notes about the `0` len apply as with `git_mailmap_from_buffer`.

include/git2/mailmap.h
mystor: It used to take a buf, and I changed it in an earlier patch. I'm fine with switching back to that if it's what we want to do.

include/git2/mailmap.h
mystor: Sure, I'm fine with making this internal. Not super attached to any of the public APIs other than the from_repo one.

mystor:
@ethomson Made the changes you requested :-)

ethomson:
Thanks again @emilio and @mystor.  Sorry it took so long to get this reviewed and merged, but I'm really happy that you contributed this.  I hope that you'll continue to contribute more in the future.  😃

emilio:
@mystor did most of the work actually, so thank her! :). Thanks for all the reviews and for getting this merged :)

PR #4587, opened 2018-03-20
Flag options in describe.h as being optional

rcjsuen:
Both `git_describe_options` and `git_describe_formatting_options` are sanitized before they are actually used. They should be noted as being optional in the documentation.

https://github.com/libgit2/libgit2/blob/5585e358b2a240ca8ed65a00008dbc865a1381c1/src/describe.c#L641-L654

https://github.com/libgit2/libgit2/blob/5585e358b2a240ca8ed65a00008dbc865a1381c1/src/describe.c#L767-L778

### DIFF
--- a/include/git2/describe.h
+++ b/include/git2/describe.h
@@ -118,7 +118,7 @@ typedef struct git_describe_result git_describe_result;
  * @param result pointer to store the result. You must free this once
  * you're done with it.
  * @param committish a committish to describe
- * @param opts the lookup options
+ * @param opts the lookup options (or NULL for defaults)
  */
 GIT_EXTERN(int) git_describe_commit(
 	git_describe_result **result,
@@ -135,7 +135,7 @@ GIT_EXTERN(int) git_describe_commit(
  * @param out pointer to store the result. You must free this once
  * you're done with it.
  * @param repo the repository in which to perform the describe
- * @param opts the lookup options
+ * @param opts the lookup options (or NULL for defaults)
  */
 GIT_EXTERN(int) git_describe_workdir(
 	git_describe_result **out,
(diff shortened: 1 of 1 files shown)


tiennou:
I think our "canonical" phrasing is "Can be NULL". Maybe something more to the point, like "Can be NULL, see git_(describe_)?_init_options for the default."

rcjsuen:
@tiennou Thank you for the review. Which would you prefer?

1. the lookup options, can be NULL

2. the lookup options, can be NULL, see git_describe_init_options for the default options

3. the lookup options, optional and can be NULL

4. the lookup options, optional and can be NULL, see git_describe_init_options for the default options

tiennou:
Oooh a vote 😉. I say 2 !

ethomson:
We should say that it _may_ be `NULL` (ie, that we permit it), not that it _can_ be `NULL`.

We probably shouldn't point people to `git_describe_init_options`, since it doesn't have additional documentation about what it's doing.  It feels like our documentation should be self-contained and shouldn't point people to reading the source...  I think that if we think that we want to provide more visibility to what the defaults are then we should figure out a way to surface that information in the documentation itself.

Personally, I think keeping it succinct - like "the lookup options (or null for defaults)" is adequate.  But I also realize that we probably don't have any strict commonality amongst the various things that take options, so ...  🤷‍♂️

rcjsuen:
In the past, I've just appended the optional word (3a133652691c07abb1a898a3704e693cb9c40901 and 93e2c74493cebb4503b8af793d4150b528efbdf5). I've also done what @ethomson suggested (21d4a3783e657c2426511760579e8883b076068f) before. 🤷‍♂️

pks-t:
The most common style we have in our code base is either "Structure with options to influence <whatever> or NULL for defaults." or "Options for <whatever>, or NULL for {defaults,default options}". I'd say "Options for <whatever>, or NULL for defaults" is the best way to go. It's short, concise, already used and one knows what to expect

rcjsuen:
Is the new version (db90e951e365b02d101429ac90bffa6132e0899b) I made okay?

pks-t:
It is. Thanks for this PR!

PR #4592, opened 2018-03-22
Documentation improvements

pks-t:
Some assorted improvements to our documentation. Despite some minor fixups, the most important thing here is probably the table of contents for our README, which should hopefully help new people. Next, I've decided to move nearly all of our documentation into "docs/". The number of doc files has grown and is likely to continue growing, so moving it into "docs/" seems like a sensible thing to do to not clutter our root tree. Furthermore, people expect to find documentation in a "docs/" folder, so it might also help discoverability.


pks-t:
Rebased to fix conflicts

ethomson:
Awesome, thanks for doing this.  GitHub seems to think that there's a conflict here (which is worrisome) so I'll merge this manually.

PR #4593, opened 2018-03-22
tests: create clone test for big repositories

pks-t:
We currently have no tests which exercise our cloning procedure with big
repositories. Create some tests that generate and clone a repository
with multiple files, each filled with 1GB of random data.

The test has originially been written by Jeff Hostetler. Due to it being
rather old, it was heavily refactored to meet our current style of code.

---

Replaces #3211, #3197.

### DIFF
--- a/tests/clone/big.c
+++ b/tests/clone/big.c
@@ -1,0 +1,168 @@
+#include "clar_libgit2.h"
+
+#include "git2/clone.h"
+#include "remote.h"
+#include "fileops.h"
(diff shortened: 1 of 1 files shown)


pks-t:
Let's see if VSTS can handle this :P

pks-t:
Indeed. Travis had these variables set, if I'm not mistaken, which is why tests have failed there. We should probably enable them for VSTS, too

PR #4594, opened 2018-03-23
odb: fix writing to fake write streams

pks-t:
In commit 7ec7aa4 (odb: assert on logic errors when writing objects,
2018-02-01), the check for whether we are trying to overflowing the fake
stream buffer was changed from returning an error to raising an assert.
The conversion forgot though that the logic around `assert`s are
basically inverted. Previously, if the statement

    stream->written + len > steram->size

evaluated to true, we would return a `-1`. Now we are asserting that
this statement is true, and in case it is not we will raise an error. So
the conversion to the `assert` in fact changed the behaviour to the
complete opposite intention.

Fix the assert by inverting its condition again and add a regression
test.

---

/cc @ethomson

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -369,7 +369,7 @@ static int fake_wstream__write(git_odb_stream *_stream, const char *data, size_t
 {
 	fake_wstream *stream = (fake_wstream *)_stream;
 
-	assert(stream->written + len > stream->size);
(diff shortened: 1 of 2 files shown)


pks-t:
This is a regression, so it needs to be part of v0.27.0

PR #4597, opened 2018-03-27
remote/proxy: fix git_transport_certificate_check_db description

cjhoward92:


pks-t:
I've checked winhttp, SSH, and HTTP transports, they all behave like you said. Usually we bubble up positive return values, but that does not happen in any of these transports. Thanks for the documentation fix!

PR #4598, opened 2018-03-27
types: remove unused git_merge_result

cjhoward92:
`git_merge_result` is currently unused in the codebase and generates a blank page in the [documentation](https://libgit2.github.com/libgit2/#HEAD/type/git_merge_result).

Not sure if this type had a use at one point, but it doesn't seem to now so I figured I would remove it.


pks-t:
Thanks for noticing! I've cross-checked just to be sure, and it's never being used indeed

PR #4599, opened 2018-03-28
[WIP] GitLab CI

pks-t:
This is a testing and WIP pull request.


pks-t:
First "succesfull" pipeline run. Online tests fail due to missing secrets

pks-t:
Copying over from #4608. Discussion on the CI purpose should probably be done here, not in the other issue.

Yeah. I feel like GitLab CI would be a lot faster than Travis -- all builds I had were run completely in parallel, so additional jobs didn't really impact us. It does have a shortcoming, though: there are no "real" 32 bit containers available, as nearly all Docker images are 64 bit only. One thought that occurred to me was to thus use both CI systems:

    Travis will build every PR/merge on our highest priority Unix platforms. These are Ubuntu 14.04 32/64 bit and mac OS.
    GitLab will do "variation" builds of mid-priority platforms, probably nightly. These include builds with OpenSSL 1.1 (Fedora?), libressl (Alpine?), a "cutting edge" distro (Arch?) to check with other recent library versions. Furthermore, GitLab should include building high-priority bindings.

Edit: GitLab should also include a recent Ubuntu, like 17.10, as Travis does not support anything different than Ubuntu 14.04

pks-t:
I think currently our best way forward is VSTS, as it supports all three platforms we're interested in. I'm thus closing this PR

PR #4602, opened 2018-03-29
odb: mempack: fix leaking objects when freeing mempacks

pks-t:
When a ODB mempack gets free'd, we take no measures at all to free its
contents, most notably the objects added to the database, resulting in a
memory leak. Call `git_mempack_reset` previous to freeing the ODB
structures themselves, which takes care of releasing all associated
data structures.

### DIFF
--- a/src/odb_mempack.c
+++ b/src/odb_mempack.c
@@ -157,6 +157,7 @@ static void impl__free(git_odb_backend *_backend)
 {
 	struct memory_packer_db *db = (struct memory_packer_db *)_backend;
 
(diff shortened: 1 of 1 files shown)


ethomson:
Seems like valgrind and coverity should have caught this.  `mempack` was basically an experiment that got shelved, but if people are using it then we should at least have some minimal unit tests here to catch these sorts of things?

pks-t:
We do have these tests since last week. That's why they _were_ catched by Valgrind, that's how I noticed

PR #4603, opened 2018-03-29
appveyor: workaround for intermittent test failures

pks-t:
Our CI builds have intermittent failures in our online tests, e.g. with
the message "A provided buffer was too small". This is not a programming
error in libgit2 but rather an error in the SChannel component of
Windows. Under certain circumstances involving Diffie-Hellman key
exchange, SChannel is unable to correctly handle input from the server.
This bug has already been fixed in recent patches for Windows 10 and
Windows Server 2016, but they are not yet available for AppVeyor.

Manually pamper over that issue by disabling all ciphersuites using DHE
via the registry. While this disables more ciphers than necessary, we
really don't care for that at all but just want to avoid build failures
due to that bug.

See [1], [2] or [3] for additional information.

1: aws/aws-sdk-cpp#671
2: dotnet/corefx#7812
3: https://support.microsoft.com/en-us/help/2992611/ms14-066-vulnerability-in-schannel-could-allow-remote-code-execution-n

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,6 +43,11 @@ build_script:
     if "%GENERATOR%"=="MSYS Makefiles" (C:\MinGW\msys\1.0\bin\sh --login /c/projects/libgit2/script/appveyor-mingw.sh)
 test_script:
 - ps: |
+    # Disable DHE key exchange to fix intermittent build failures ("A buffer
+    # provided was too small") due to SChannel bug. See e.g.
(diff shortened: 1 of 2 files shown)


pks-t:
Restarting AppVeyor to gain more confidence that this actually works as advertised

pks-t:
We've had three test runs on AppVeyor now, all without any failure. This doesn't yet prove that this PR works as intended, but it does create some confidence. I'm merging this for now -- we can still revert/amend it if the problem persists

PR #4606, opened 2018-04-01
revwalk: avoid walking the entire history when output is unsorted

carlosmn:
As part of reducing our divergence from git, its code for revwalk was ported
into our codebase. A detail about when to limit the list was lost and we ended
up always calling that code.

Limiting the list means performing the walk and creating the final list of
commits to be output during the preparation stage. This is unavoidable when
sorting and when there are negative refs.

We did this even when asked for unsorted output with no negative refs, which you
might do to retrieve something like the "last 10 commits on HEAD" for a
nominally unsorted meaning of "last".

This commit adds and sets a flag indicating when we do need to limit the list,
letting us avoid doing so when we can. The previously mentioned query thus no
longer loads the entire history of the project during the prepare stage, but
loads it iteratively during the walk.

### DIFF
--- a/src/revwalk.c
+++ b/src/revwalk.c
@@ -453,43 +453,28 @@ static int limit_list(git_commit_list **out, git_revwalk *walk, git_commit_list 
 	return 0;
 }
 
-static int get_one_revision(git_commit_list_node **out, git_revwalk *walk, git_commit_list **list)
-{
-	int error;
-	git_commit_list_node *commit;
-
-	while(true) {
-		commit = git_commit_list_pop(list);
-		if (!commit) {
-			giterr_clear();
-			return GIT_ITEROVER;
-		}
-
-		/*
-		 * If we did not run limit_list and we must add parents to the
-		 * list ourselves.
-		 */
-		if (!walk->limited) {
-			if ((error = add_parents_to_list(walk, commit, list)) < 0)
-				return error;
-		}
-
-		*out = commit;
-		return 0;
-	}
-}
-
 static int get_revision(git_commit_list_node **out, git_revwalk *walk, git_commit_list **list)
 {
 	int error;
 	git_commit_list_node *commit;
 
-	if ((error = get_one_revision(&commit, walk, list)) < 0)
-		return error;
+  commit = git_commit_list_pop(list);
+  if (!commit) {
+    giterr_clear();
+    return GIT_ITEROVER;
+  }
 
-	/* Here is where we would handle boundary commits if we implement that */
-	*out = commit;
-	return 0;
+  /*
+   * If we did not run limit_list and we must add parents to the
+   * list ourselves.
+   */
+  if (!walk->limited) {
+    if ((error = add_parents_to_list(walk, commit, list)) < 0)
+      return error;
+  }
+
+  *out = commit;
+  return 0;
 }
 
 static int sort_in_topological_order(git_commit_list **out, git_revwalk *walk, git_commit_list *list)
--- a/src/revwalk.c
+++ b/src/revwalk.c
@@ -15,6 +15,8 @@
 #include "merge.h"
 #include "vector.h"
 
+static int get_revision(git_commit_list_node **out, git_revwalk *walk, git_commit_list **list);
+
 git_commit_list_node *git_revwalk__commit_lookup(
 	git_revwalk *walk, const git_oid *oid)
 {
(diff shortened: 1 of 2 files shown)


src/revwalk.c
```diff
@@ -449,6 +453,45 @@ static int limit_list(git_commit_list **out, git_revwalk *walk, git_commit_list
...
+			if ((error = add_parents_to_list(walk, commit, list)) < 0)
+				return error;
+		}
+
+		*out = commit;
+		return 0;
+	}
+}
+
+static int get_revision(git_commit_list_node **out, git_revwalk *walk, git_commit_list **list)
```
pks-t: There's not really a difference to `get_one_revision` right now, is there?

src/revwalk.c
```diff
@@ -449,6 +453,45 @@ static int limit_list(git_commit_list **out, git_revwalk *walk, git_commit_list
 	return 0;
 }
 
+static int get_one_revision(git_commit_list_node **out, git_revwalk *walk, git_commit_list **list)
+{
+	int error;
+	git_commit_list_node *commit;
+
+	while(true) {
```
pks-t: This loop is useless right now -- we always return during the first loop

src/revwalk.c
```diff
@@ -664,6 +707,9 @@ void git_revwalk_sorting(git_revwalk *walk, unsigned int sort_mode)
 		walk->get_next = &revwalk_next_unsorted;
 		walk->enqueue = &revwalk_enqueue_unsorted;
 	}
+
+	if (sort_mode != GIT_SORT_NONE)
+		walk->limited = 1;
```
pks-t: This has the limitation that we will stay limited when somebody is switching sorting to `GIT_SORT_NONE` afterwards, again. But handling that case correctly would need another flag, so I guess that's okay

src/revwalk.c
```diff
@@ -245,34 +249,34 @@ static int revwalk_next_timesort(git_commit_list_node **object_out, git_revwalk
...
 			return 0;
 		}
 	}
 
-	giterr_clear();
-	return GIT_ITEROVER;
+	return error;
 }
 
 static int revwalk_next_reverse(git_commit_list_node **object_out, git_revwalk *walk)
```
pks-t: Wouldn't `revwalk_next_reverse` need the `get_revision` treatment, as well?

carlosmn: The reverse iterator does the full walk itself in `prepare_walk` by iterating via whatever other iterator we have set up and inserts them in this iterator.

We could potentially convert it into the same kind of loop but like with the pure time sort, we already have everything in the list.

src/revwalk.c
carlosmn: Yeah, part of the porting. Here is where simplification handling happens in git. It's probably worth removing it to reduce confusion.

src/revwalk.c
carlosmn: Right now it just calls the lower-level function but extra features would get built here.

src/revwalk.c
carlosmn: Handling that would need something like `else if (!walk->did_hide)` which is not too bad, but would be quite an edge case where you're double-guessing what sorting you want for a single walk.

carlosmn:
>have you done any benchmarks by how much this actually speeds up listings?

That depends on how big your history is. With this an "unsorted" walk without negative refs will only read however many commits you walk down (plus some overhead for the parents that end up in there) instead of reading the whole list.

My benchmark was to load up the github repository and call `git_revwalk_next` 11 times. Without the patch it takes about 7s and with it it's about 20ms or so.

src/revwalk.c
pks-t: I'd say we should avoid having two equivalent function calls for now, though. It can still easily be extended later when implementing said extra features

carlosmn:
I just pushed up the simplification removing the extra layer of function call that doesn't do anything. I think it'll be fine to skip the trying to guess whether we can disable limiting when changing the sorting method back to `NONE`.

src/revwalk.c
```diff
@@ -449,6 +453,30 @@ static int limit_list(git_commit_list **out, git_revwalk *walk, git_commit_list
 	return 0;
 }
 
+static int get_revision(git_commit_list_node **out, git_revwalk *walk, git_commit_list **list)
+{
+	int error;
+	git_commit_list_node *commit;
+
+  commit = git_commit_list_pop(list);
```
ethomson: Could you clean up the formatting in this function?

ethomson:
Since @carlosmn updated this to address @pks-t's concerns, I went ahead and fixed up the formatting problems myself.

nebosite:
I have a really big repository (many GB) and repository.Commits.Take(100) takes around **200 seconds** to run.   Compare to git log -n 100 which takes about 1 second.     The culprit is this line in libgit2sharp:

            int res = NativeMethods.git_revwalk_next(out ret, walker);

nebosite:
I was using a same-day clone of master.    Not sure I can reproduce this without a really large repo.  I'm guessing the behavior might be exponential.

PR #4607, opened 2018-04-02
Fix build with LibreSSL 2.7

Sp1l:
LibreSSL 2.7 adds OpenSSL 1.1 API

Signed-off-by: Bernard Spil <brnrd@FreeBSD.org>

PR #4608, opened 2018-04-03
OpenSSL legacy API cleanups

pks-t:
Some follow-ups for #4607.


pks-t:
Which brings me to something different: it would be cool to have different CI jobs for OpenSSL 1.0, OpenSSL 1.1 and libressl 2.6 or 2.7.

ethomson:
> Which brings me to something different: it would be cool to have different CI jobs for OpenSSL 1.0, OpenSSL 1.1 and libressl 2.6 or 2.7.

Maybe, but our CI is already slow.  I would also like to see us have an improved test matrix, but we need to think seriously about both the continuing degradation in speed for pull requests.

I suggest we start thinking about splitting out the platform matrix into the highest priority platforms that will give us a high confidence in the PR (probably a single Linux, Windows and macOS installation) and then do nightlies for the entire set of systems that we want to test.

pks-t:
Yeah. I feel like GitLab CI would be a lot faster than Travis -- all builds I had were run completely in parallel, so additional jobs didn't really impact us. It does have a shortcoming, though: there are no "real" 32 bit containers available, as nearly all Docker images are 64 bit only. One thought that occurred to me was to thus use both CI systems:

- Travis will build every PR/merge on our highest priority Unix platforms. These are Ubuntu 14.04 32/64 bit and mac OS.
- GitLab will do "variation" builds of mid-priority platforms, probably nightly. These include builds with OpenSSL 1.1 (Fedora?), libressl (Alpine?), a "cutting edge" distro (Arch?) to check with other recent library versions. Furthermore, GitLab should include building high-priority bindings.

tiennou:
I like the cleanup !

I have one related dangling [commit](https://github.com/tiennou/libgit2/commit/f89ecd53db306e2d21efdf238608fd8cd048facb), which might be integrated in this (already on top of this branch 😉).

pks-t:
Thanks @tiennou, amended your commit to this PR

PR #4611, opened 2018-04-05
diff: Add missing GIT_DELTA_TYPECHANGE -> 'T' mapping.

erikvanzijst:
This adds the 'T' status character to git_diff_status_char() for diff entries that change type.

@ethomson This follows from our Slack chat earlier today.


ethomson:
Can you move the declaration to the top of the function so that braindead compilers can cope?

pks-t:
Thanks for your PR!

Issue #4610, opened 2018-04-03
libgit2 doesn't handle .gitignore/exclude patterns with a trailing slash and white space


segevfiner:
How "fun"... 😔 

### Reproduction steps
1. Execute:
    ```sh
    git init test-repo && cd test-repo
    mkdir spam
    touch spam/ham.c
    echo "spam/ " > .gitignore
    ```
2. `git status --ignored` shows that "spam/" is ignored.
3. With `pygit2`:
    ```py
    >>> from pygit2 import Repository
    >>> r = Repository('.')
    >>> r.path_is_ignored('spam')
    False
    >>> r.path_is_ignored('spam/ham.c')
    False
    ```

### Expected behavior
The spam directory should be ignored.

### Actual behavior
It isn't.

### Version of libgit2 (release number or SHA1)
Tested both v0.27.0 and master at d9007dc8520942763a21e4058dfa8f159f3e3c56.

### Operating system(s) tested
Ubuntu 16.04 amd64
Windows 10.0.16299.334 x64

### Related
https://github.com/libgit2/libgit2/issues/4309
https://github.com/libgit2/libgit2/commit/5cb6a2c946a071d43be3cdabf2f6d4417a4be694

cc @pks-t @novalis

pks-t:
Thanks for the report. I can reproduce the error and am working on a fix

pks-t:
Please verify that #4614 fixes your issue

ethomson:
Thanks for reporting this.  Fixed via #4614.

PR #4614, opened 2018-04-06
attr_file: fix handling of directory patterns with trailing spaces

pks-t:
When comparing whether a path matches a directory rule, we pass the
both the path and directory name to `fnmatch` with
`GIT_ATTR_FNMATCH_DIRECTORY` being set. `fnmatch` expects the pattern to
contain no trailing directory '/', which is why we try to always strip
patterns of trailing slashes. We do not handle that case correctly
though when the pattern itself has trailing spaces, causing the match to
fail.

Fix the issue by stripping trailing spaces and tabs for a rule previous
to checking whether the pattern is a directory pattern with a trailing
'/'. Add a test to catch future breakage.

### DIFF
--- a/src/attr_file.c
+++ b/src/attr_file.c
@@ -633,6 +633,11 @@ int git_attr_fnmatch__parse(
 		if (--spec->length == 0)
 			return GIT_ENOTFOUND;
 
+	/* Remove trailing spaces. */
+	while (pattern[spec->length - 1] == ' ' || pattern[spec->length - 1] == '\t')
+		if (--spec->length == 0)
+			return GIT_ENOTFOUND;
+
 	if (pattern[spec->length - 1] == '/') {
 		spec->length--;
 		spec->flags = spec->flags | GIT_ATTR_FNMATCH_DIRECTORY;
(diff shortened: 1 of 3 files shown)


src/attr_file.c
```diff
@@ -633,6 +633,11 @@ int git_attr_fnmatch__parse(
 		if (--spec->length == 0)
 			return GIT_ENOTFOUND;
 
+	/* Remove trailing spaces. */
```
segevfiner: Doesn't this make [5cb6a2c#L216-225](https://github.com/libgit2/libgit2/commit/5cb6a2c946a071d43be3cdabf2f6d4417a4be694#diff-2aa744cf46a2b9ef04c23ad8fc42b3ceR216) redundant?

segevfiner:
> Please verify that #4614 fixes your issue
> &ndash; https://github.com/libgit2/libgit2/issues/4610#issuecomment-379226963

It does. 👍

pks-t:
@segevfiner: you're right, it makes the code in the ignore parser redundant. Removed that code

ethomson:
Thanks @segevfiner for reporting the issue and verifying the fix.  Thanks @pks-t for fixing!

PR #4618, opened 2018-04-10
refs: preserve the owning refdb when duping reference

tiennou:
This fixes a segfault in `git_reference_owner` on references returned from `git_reference__read_head` and `git_reference_dup` ones.

### DIFF
--- a/src/refs.c
+++ b/src/refs.c
@@ -115,6 +115,9 @@ int git_reference_dup(git_reference **dest, git_reference *source)
 
 	GITERR_CHECK_ALLOC(*dest);
 
+	(*dest)->db = source->db;
+	GIT_REFCOUNT_INC((*dest)->db);
+
 	return 0;
 }
 
--- a/src/refs.h
+++ b/src/refs.h
@@ -116,6 +116,10 @@ int git_reference_lookup_resolved(
  * with the given name pointing to the reference pointed to by
  * the file. If it is not a symbolic reference, it will return
  * the resolved reference.
+ *
+ * Note that because the refdb is not involved for symbolic references, they
+ * won't be owned, hence you should either not make the returned reference
+ * 'externally visible', or perform the lookup before returning it to the user.
  */
 int git_reference__read_head(
 	git_reference **out,
--- a/tests/refs/dup.c
+++ b/tests/refs/dup.c
@@ -21,6 +21,7 @@ void test_refs_dup__direct(void)
 	cl_git_pass(git_reference_dup(&b, a));
 
 	cl_assert(git_reference_cmp(a, b) == 0);
+	cl_assert(git_reference_owner(b) == g_repo);
 
 	git_reference_free(b);
 	git_reference_free(a);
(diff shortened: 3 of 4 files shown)


tiennou:
Note that I don't feel confident enough to change the behavior of `__read_head`, but maybe it's the right thing to do ?

tiennou:
This was introduced in 5b65ac2, and the fix looks easy : grab the refdb from the repo and put it in. That'll allow the comment to be dropped and preserve the "owner" semantics.

@pks-t maybe you can say if not having an owner is an oversight or a performance improvement of sorts (the commit message implies the latter) ? Mostly because right now I'm looking at the refdb ownership mechanism *and* worktrees, and I feel overwhelmed 😉.

pks-t:
Hm. First thing I notice is that `git_reference__read_head` is poorly named. It really should've been something like `git_reference__from_file`, as it doesn't necessarily have anything to do with HEADs at all. Furthermore, I think that your easy fix would be the right thing to do. We already associate it with the passed in repo in case it is not a symbolic reference, so we should do the same when it is a symbolic reference. It shouldn't impact performance at all to just have an additional refcount increment on the refdb.

pks-t:
Umm... are they? The function should be totally safe to call on any reference, disregarding if it is a symbolic one or not. I probably shouldn't argue, because I'm the one who named that function like that :P

tiennou:
Just for clarification, it's only used to [resolve worktree HEADs](https://github.com/libgit2/libgit2/blob/master/src/repository.c#L2147) at the moment, and it seems to be perfectly safe to use as long as you don't really need the reference.

The only issue it has is that it doesn't setup the owning refdb pointer correctly when resolving a symbolic reference (as it just parses the file's content), so you need to make sure to resolve a second time (which `git_repository_head_for_worktree` does), hence the "performance improvement" argument I made.

I just hit that missing refdb case while moving unborn-ness checking in `refs.c` (see 37b69a581667f01d84168cae6b30f76ef17ab17d, part of [this](https://github.com/libgit2/libgit2/compare/master...tiennou:feature/merge-analysis-any-branch).

In all, as long as that weirdness is documented, I think it's fine to keep it that way, I just have an extra `repo` parameter in that `resolve_unborn` function because I can't depend on the ref's ownership 🤷‍♂️.

tiennou:
(Also also, that worktree head resolution function return NOTFOUND instead of UNBORN)

ethomson:
Thanks, @tiennou!

tiennou:
I'd like to know if that NOTFOUND/UNBORN nuance should be considered a bug/missing feature from the worktree support (note that I have absolutely no idea how worktrees work so this may not even be considered valid) ?

From the cursory look I gave the code, `git_repository_head` creates that nuance (and APIs that are downstream from it), to differentiate cases where the HEAD file is missing (NOTFOUND) from the case where the pointed-to ref is missing (UNBORN). AFAICT the refdb will NOTFOUND on resolving a missing pointed-to symbolic ref.

PR #4620, opened 2018-04-11
[WIP] Hook support

tiennou:
This is a heavily redacted version of #3824, since it's been 2 years in the making and much has changed (like nice commit headers). Most of the design comes from the discussion there, so you might want to take a look.

The plan is to have a low-level API for hook execution (because as a library, we don't want to go anywhere near `fork`/`exec`/the Windows way), so we can instead ping "userspace" that we expect them to execute something with carefully prepared arguments. On top of that comes a somewhat higher-level API so that users should just need to call a single function with some of their current working objects and we'll translate those to one of those low-level "pings".

Note that's there a [feature/hooks-wip branch](https://github.com/tiennou/libgit2/compare/feature/hooks...tiennou:feature/hooks-wip), which is even more WIP than this one, where I'm adding easy-to-use helpers for the various hooks.


tiennou:
Editor's note : I should really have those .h files reversed.

src/repository.c
```diff
@@ -177,6 +177,10 @@ void git_repository_free(git_repository *repo)
 	git__free(repo->ident_name);
 	git__free(repo->ident_email);
 
+	if (repo->hook_payload != NULL && repo->hook_payload_free != NULL) {
```
pks-t: This is a bit weird to use. I'd understand not calling `hook_payload_free` if its type wasn't called `git_hook_destructor`. With that naming, though, a user might expect it to also be called in case there is no payload at all

tests/hook/execute.c
```diff
@@ -0,0 +1,96 @@
...
+}
+
+static int hook_exec__frees_payload_with_repo(git_hook_env *env, void *payload)
+{
+	GIT_UNUSED(env);
+	GIT_UNUSED(payload);
+	return 0;
+}
+
+static int destruct_called = 0;
```
pks-t: You should not use a static variable but instead pass it as payload. Otherwise you don't test whether the payload to destructors is being passed properly

src/repository.c
```diff
@@ -177,7 +177,7 @@ void git_repository_free(git_repository *repo)
 	git__free(repo->ident_name);
 	git__free(repo->ident_email);
 
-	if (repo->hook_payload != NULL && repo->hook_payload_free != NULL) {
+	if (repo->hook_payload_free != NULL) {
```
pks-t: Okay, so you already do what I wanted in a later commit. Never mind, then :)

src/hook.c
```diff
@@ -204,7 +204,11 @@ static int hook_execute_va(git_buf *io, git_repository *repo, const char *name,
 
 	env.args.strings = (char **)git_vector_detach(&env.args.count, NULL, &arg_vector);
 
+	giterr_clear();
```
pks-t: We should probably check whether `git_vector_insert` succeeds or not. By ignoring its errors and clearing internal errors we may pass an invalid environment to the hook, which we should avoid.

src/hook.c
```diff
@@ -204,7 +204,11 @@ static int hook_execute_va(git_buf *io, git_repository *repo, const char *name,
...
 	err = repo->hook_executor(&env, repo->hook_payload);
+	if (err < 0 && !giterr_last()) {
+		giterr_set(GITERR_HOOK, "hook \"%s\" reported failure", name);
```
pks-t: `goto cleanup` should probably be added here. Just to avoid the common pitfall where we have to add more code after this statement, but then forget to add the missing goto.

include/git2/hook.h
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * A helper to get the path of the repository's hooks.
+ *
+ * This will obey core.hooksPath.
+ *
+ * @param out_dir The absolute path to the hooks location.
+ * @param repo The repository to get the hooks location from.
+ * @return 0 on success, or an error code.
+ */
+GIT_EXTERN(int) git_hook_dir(char **out_dir, git_repository *repo);
```
pks-t: You should make clear that `out_dir` is caller-owned (at least I guess it is due to it not being `const`). Seeing that we might have custom allocators, we probably shouldn't even pass back memory ownership, as we do not expose `git__free`. Maybe use a `git_buf` instead, as there we have `git_buf_free`?

src/hook.c
```diff
@@ -0,0 +1,153 @@
...
+	/* We need to check for an override in the repo config */
+	err = git_repository_config__weakptr(&cfg, repo);
+	if (err != 0)
+		return err;
+
+	err = git_config_get_path(&cfg_path, cfg, "core.hooksPath");
+	if (err == GIT_ENOTFOUND) {
+		giterr_clear();
+		git_buf_puts(&tmp_path, git_repository_commondir(repo));
+		git_buf_joinpath(&tmp_path, tmp_path.ptr, GIT_HOOKS_DIR);
```
pks-t: Instead of doing `git_buf_puts`, `git_buf_joinpath`, you can just as well directly `git_buf_joinpath(&tmp_path, git_repository_commondir(repo), GIT_HOOKS_DIR)`.

src/hook.c
```diff
@@ -0,0 +1,153 @@
...
+	} else if (err == GIT_OK) {
+		git_buf_joinpath(&tmp_path, git_repository_commondir(repo), cfg_path.ptr);
+		git_path_resolve_relative(&tmp_path, 0);
+		git_buf_free(&cfg_path);
+	} else {
+		return -1;
+	}
+
+	*out_dir = tmp_path;
```
pks-t: Can't you just avoid `tmp_path` altogether and directly print into `out_dir`? Especially as you leak memory in case `out_dir` was already allocated

src/hook.c
```diff
@@ -0,0 +1,153 @@
...
+	return 0;
+}
+
+int git_hook_dir(char **out_dir, git_repository *repo) {
```
pks-t: The opening brace belongs on the next line ;)

src/hook.c
```diff
@@ -0,0 +1,153 @@
...
+static int check_hook_path(const char *hook_path)
+{
+	int err;
+	struct stat hook_stat;
+
+	assert(hook_path);
+
+	/* Skip missing hooks */
+	err = p_stat(hook_path, &hook_stat);
+	if (err)
```
pks-t: I'm a bit undecided here. I feel like we should check specifically for `ENOTFOUND` here and handle other errors as "real" errors. What do you think?

src/hook.c
```diff
@@ -0,0 +1,153 @@
...
+		if (err != 0)
+			continue;
+
+		err = check_hook_path(hook_path);
+		git__free(hook_path);
+
+		if (err != 0)
+			continue;
+
+		callback(hook_name, payload);
```
pks-t: We usually allow callbacks to abort iteration by returning non-zero. Do we want to do the same here?

tests/hook/enumerate.c
```diff
@@ -0,0 +1,86 @@
...
+	cl_git_pass(git_repository_config(&cfg, g_repo));
+	cl_git_pass(git_config_set_string(cfg, "core.hooksPath", ALT_HOOK_DIR));
+
+	cl_git_pass(git_hook_dir(&alt_hook_str, g_repo));
+	git_buf_attach(&alt_hook, alt_hook_str, strlen(alt_hook_str));
+
+	/* Setup an alternate hook directory */
+	cl_must_pass(p_mkdir(git_buf_cstr(&alt_hook), 0777));
+
+	git_buf_joinpath(&alt_hook, alt_hook.ptr, "commit-msg");
```
pks-t: There are some `cl_git_pass` missing here and in the following lines

tests/hook/enumerate.c
```diff
@@ -0,0 +1,86 @@
...
+	git_buf_puts(&expected_list, "post-merge");
+
+	cl_git_pass(git_hook_foreach(g_repo, test_git_hook_foreach_cb, &hook_list));
+
+	cl_assert_equal_s(git_buf_cstr(&hook_list), git_buf_cstr(&expected_list));
+	git_buf_free(&hook_list);
+	git_buf_free(&expected_list);
+}
+
+#undef ALT_HOOK_DIR
```
pks-t: I wouldn't bother about this undef

tests/hook/call.c
```diff
@@ -0,0 +1,71 @@
+#include "clar_libgit2.h"
+
+static git_repository *repo = NULL;
```
pks-t: No need for this static initialization

src/rebase.c
```diff
@@ -718,6 +719,11 @@ int git_rebase_init(
 		branch = head_branch;
 	}
 
+	if (!inmemory && upstream) {
+		if ((error = git_hook_call_pre_rebase(repo, upstream, (head_ref == NULL ? branch : NULL))) < 0)
```
pks-t: Do we really want to check for `< 0` here? Usually, return codes from exec'd programs are always positive in the range from 0 to 255, where non-zero positive values denote an error. So in case the callback simply exec's the hook and returns its value, we will not detect that error.

tiennou:
There's context about what can be matched to our API/workflow/implementation-level [here](https://github.com/libgit2/libgit2/pull/3824#issuecomment-258997589), which I've moved in this PR's summary.

Also, someone recently asked about merge drivers in OCGit libgit2/objective-git#656, and this made me wonder about that exec support…

src/hook.c
tiennou: Agreed. I'll do a mapping of `ENOENT` to `GIT_ENOTFOUND` and the rest to a generic `GITERR_OS` class.

src/hook.c
tiennou: This ended up somehow in the -wip branch. As I can't see any valid reason to clear the error here, I'll 🔥 it.

src/repository.c
tiennou: Yep, will fixup though, it's confusing.

tests/hook/call.c
tiennou: That's the "standard" test case static repo (albeit not named g_repo) 😕. I'll just rename it then.

src/hook.c
tiennou: Fixed.

src/hook.c
tiennou: Agreed. But I can even use `git_repository_item_path` 😉.

tests/hook/enumerate.c
tiennou: Fixed.

src/rebase.c
tiennou: I'm defaulting to the libgit2 convention. Right now the expectation is that the executor will bubble a `GIT_*` error, not the underlying exec return value.

tiennou:
Review comments handled. Note that I've moved the WIP commit hook support in, but it's really WIP. If someone has a better idea for wrapping prepare-commit-msg (and it's variable 5 arguments + command line flags + some repo state) that does not entails having 5 different functions, I'm 👂…

About [this](https://github.com/libgit2/libgit2/pull/4620#discussion_r189258922) :
> Do we really want to check for < 0 here? Usually, return codes from exec'd programs are always positive in the range from 0 to 255, where non-zero positive values denote an error. So in case the callback simply exec's the hook and returns its value, we will not detect that error.

It all depends on how I specify the valid return codes from the executor. My preference goes to keeping the libgit2 convention, which would make the executor responsible of not breaking us by returning positive error codes.

tiennou:
Note that I've swapped functions around, so the low-level execution is in `git2/sys/hook.h` and the easy-to-use stuff in `git2/hook.h`.

Editor's note: document that `git_hook_execute`'s arguments *must* be NULL-terminated 😜.

tiennou:
Rebased, and updated. I've closed all review comments that seemed to have been resolved, and thus I'm dropping the WIP from it, as I feel it's a sufficient first step to merit inclusion, on the off-chance that it's acceptable to not have all current hook "helpers" wired yet.

ruant:
Can we have this now?

GXGOW:
Is there any status update on this pull request?

pks-t:
On Wed, Feb 05, 2020 at 02:43:24AM -0800, Nicolas Loots wrote:
> Is there any status update on this pull request?

We are currently in code freeze until v1.0 is released, probably around
early March. Until then we're holding off on all new features, but will
consider merging them as soon as v1.0 is out.

extrawurst:
Now that 1.0 shipped, can we consider this PR? That would be amazing 🤞

extrawurst:
Also wondering what keeps us from considering this now post-1.0

ArekPiekarz:
I am also very interested in being able to support precommit hooks in my application using libgit2 (Rust version to be specific). What is the state of work on them?

Whirlwind:
From 2018 to 2021, I can't wait. ...

Pyenb:
2024 and we're still here

ethomson:
@Pyenb What do you _want_ hook support to look like?

Issue #4333, opened 2017-08-30
git_revwalk outputs unwanted commits


ltrzesniewski:
Under some circumstances, `git_revwalk` will output too many commits. The behavior depends on the order of `git_revwalk_hide` calls.

I originally spotted the problem with a program which does many `git_revwalk_push` followed by many `git_revwalk_hide` in a single revwalk (in order to find out which commits are new after a fetch on every branch), but I managed to extract a test case which shows the issue. I'm not sure what _exactly_ is the root cause though.

Here's what the [test repo](https://github.com/libgit2/libgit2/files/1263157/test.git.zip) looks like:

![image](https://user-images.githubusercontent.com/7913492/29867585-56b9705a-8d7c-11e7-95ae-7980db584d44.png)

The test program does the equivalent of the following commands:

```
$ git rev-list D ^A ^B ^E
790ba0facf6fd103699a5c40cd19dad277ff49cd
b82cee5004151ae0c4f82b69fb71b87477664b6f

$ git rev-list D ^B ^A ^E
790ba0facf6fd103699a5c40cd19dad277ff49cd
b82cee5004151ae0c4f82b69fb71b87477664b6f
```

Which produce the same correct result with git, but libgit2 outputs more commits in the second case.

### Reproduction steps

Add the following file to the examples directory (for instance as `bug-repro.c`):

```C
#include "common.h"

static void dump_walk(git_revwalk *walk)
{
	git_oid oid;
	char buf[GIT_OID_HEXSZ+1];

	while (!git_revwalk_next(&oid, walk)) {
		git_oid_fmt(buf, &oid);
		buf[GIT_OID_HEXSZ] = '\0';
		printf("%s\n", buf);
	}
}

int main(void)
{
	git_repository *repo;
	git_revwalk *walk;

	git_libgit2_init();

	check_lg2(git_repository_open_ext(&repo, "test.git", 0, NULL), "opening repository", NULL);
	check_lg2(git_revwalk_new(&walk, repo), "allocating revwalk", NULL);

	puts("Normal behavior:");

	git_revwalk_push_ref(walk, "refs/heads/D");
	git_revwalk_hide_ref(walk, "refs/heads/A");
	git_revwalk_hide_ref(walk, "refs/heads/B");
	git_revwalk_hide_ref(walk, "refs/heads/E");

	dump_walk(walk);

	git_revwalk_free(walk);
	check_lg2(git_revwalk_new(&walk, repo), "reallocating revwalk", NULL);

[...]

PR #4622, opened 2018-04-12
revwalk: fix uninteresting revs sometimes not limiting graphwalk

pks-t:
When we want to limit our graphwalk, we use the heuristic of checking
whether the newest limiting (uninteresting) revision is newer than the
oldest interesting revision. We do so by inspecting whether the first
item's commit time of the user-supplied list of revisions is newer than
the last added interesting revision. This is wrong though, as the user
supplied list is in no way guaranteed to be sorted by increasing commit
dates. This could lead us to abort the revwalk early before applying all
relevant limiting revisions, outputting revisions which should in fact
have been hidden.

Fix the heuristic by instead checking whether _any_ of the limiting
commits was made earlier than the last interesting commit. Add a test.


pks-t:
This should fix #4333

Issue #4619, opened 2018-04-11
Fast iteration over references in a scope/directory (rather than globbing)


neithernut:
In a project of ours, we potentially end up with lots of references ordered in a hierarchical structure (note: we are aware that git is generally not designed for huge amounts of references). For example, we may have a scope/directory `refs/dit/some_hash/` containing references like
```
refs/dit/some_hash/head
refs/dit/some_hash/leaves/123
refs/dit/some_hash/leaves/456
refs/dit/some_hash/leaves/789
...
```
(`refs/dit/some_hash/leaves/` would be a scope of its own).

For various reasons, we may have to iterate over/collect all references within a scope. For this we effectively use an iterator created via `git_reference_iterator_glob_new()`. Globbing turned out to be quite slow (supprise!... not), which manifested in very low performance for some operation (see neithernut/git-dit#174).

In our case, we don't actually require globbing. As described above, we are only interested in all refs within a scope. Retrieving them should be possible as nearly as fast as listing the directory. In fact, we could just get a directory-listing from the respective `refs/` sub-directory and the resulting paths to `git_reference_lookup()`. However, this will not work with ODB-based repos. Also, the functionality may be desired by other users, too.

Hence, I think an API for listing references and sub-scopes within a scope would be beneficial. Well, at least for us. I'm also willing to contribute such a functionality.

Ideas? Opinions?

neithernut:
An alternative which I'm currently investigating is optimizing the globbing.

There appears to be much room for improvement. If I'm reading the source correctly (e.g. `iter_load_loose_paths()` in `refdb_fs.c`), every single reference in the repository is visited during the iteration and the name matched against the `glob`. Clearly, it should be possible to at least skip directories if they are not matched by the respective part of the glob.

Maybe I'll find some time hacking an optimization during the weekend.

PR #4629, opened 2018-04-18
refdb_fs: enhance performance of globbing

neithernut:
This patch-set addresses the performance of reference iteration with a glob specified. The enhancement is specific to FS-based repositories.

libgit2 provides facilities for iterating over references of a repository. The references may be filtered using a glob, e.g. using an iterator created via `git_reference_iterator_glob_new()`. At least for FS-based repositories, all references are visited during the iteration and each of the references encountered is matches against the glob. For the general case, this is the only option.

However, a glob may also start with a literal path. In this case, we may scan only the corresponding subdirectory of `refs` rather than the entire reference space. The cost of this optimization is a single scan of the glob.

Targets #4619.

TODO:
 * [x] Fix errors shown by the test-suite. I do something wrong, probably during path assembly.
 * [x] Measure the performance enhancement/impact for both the "usual" case and repositories with _lots_ of refs.

### DIFF
--- a/src/refdb_fs.c
+++ b/src/refdb_fs.c
@@ -505,26 +505,29 @@ static int iter_load_loose_paths(refdb_fs_backend *backend, refdb_fs_iter *iter)
 	git_iterator *fsit = NULL;
 	git_iterator_options fsit_opts = GIT_ITERATOR_OPTIONS_INIT;
 	const git_index_entry *entry = NULL;
+	const char *ref_prefix = GIT_REFS_DIR;
+	size_t ref_prefix_len = strlen(ref_prefix);
 
 	if (!backend->commonpath) /* do nothing if no commonpath for loose refs */
 		return 0;
 
 	fsit_opts.flags = backend->iterator_flags;
 
-	if ((error = git_buf_printf(&path, "%s/refs", backend->commonpath)) < 0 ||
+	if ((error = git_buf_printf(&path, "%s/", backend->commonpath)) < 0 ||
+		(error = git_buf_put(&path, ref_prefix, ref_prefix_len)) < 0 ||
 		(error = git_iterator_for_filesystem(&fsit, path.ptr, &fsit_opts)) < 0) {
 		git_buf_free(&path);
 		return error;
 	}
 
-	error = git_buf_sets(&path, GIT_REFS_DIR);
+	error = git_buf_sets(&path, ref_prefix);
 
 	while (!error && !git_iterator_advance(&entry, fsit)) {
 		const char *ref_name;
 		struct packref *ref;
 		char *ref_dup;
 
-		git_buf_truncate(&path, strlen(GIT_REFS_DIR));
+		git_buf_truncate(&path, ref_prefix_len);
 		git_buf_puts(&path, entry->path);
 		ref_name = git_buf_cstr(&path);
 
--- a/src/refdb_fs.c
+++ b/src/refdb_fs.c
@@ -513,6 +513,30 @@ static int iter_load_loose_paths(refdb_fs_backend *backend, refdb_fs_iter *iter)
 
 	fsit_opts.flags = backend->iterator_flags;
 
+	if (iter->glob) {
+		const char *last_sep = NULL;
+		const char *pos;
+		for (pos = iter->glob; *pos; ++pos) {
+			switch (*pos) {
+			case '?':
+			case '*':
+			case '[':
+			case '\\':
+				break;
+			case '/':
+				last_sep = pos;
+				/* FALLTHROUGH */
+			default:
+				continue;
+			}
+			break;
+		}
+		if (last_sep) {
+			ref_prefix = iter->glob;
+			ref_prefix_len = (last_sep - ref_prefix) + 1;
+		}
+	}
+
 	if ((error = git_buf_printf(&path, "%s/", backend->commonpath)) < 0 ||
 		(error = git_buf_put(&path, ref_prefix, ref_prefix_len)) < 0 ||
 		(error = git_iterator_for_filesystem(&fsit, path.ptr, &fsit_opts)) < 0) {


neithernut:
This is now be ready for review.

----

I did some preliminary measurements of cases with few references in the repository. The performance was tested by timing 10 consecutive runs of the test-suite:
```
time for i in {1..10}; do ./libgit2_clar; done
```
Obviously, the methodology is not quite what you'd want for a real measurement.

The results with the changes:
```
real	2m44.829s
user	1m4.892s
sys	1m40.423s
```

The results for the commit this PR is based on (d906a8796cec1fc4d14144f53e505ab215bdab0c):
```
real	2m41.803s
user	1m3.824s
sys	1m38.538s
```

As you can see, the tests took a bit more time with the changes. This was to be somewhat expected, since in the tests, the globs usually select a significant portion of the references. Hence, the additional, albeit superficial, parse of the glob gains more weight.

However, I also saw spurious segfaults for both the master and the feature-branch. Those hit once while running the test for the master and, I think, not for any of the runs for the feature-branch. So maybe the difference is just one run ending earlier.

Anyways, I will repeat those tests again with a script which gives me more informative statistics. I will also do some tests with one of those repos I have with lots of refs (the target scenario for the optimization).

Btw, you could have a look at the durations reported by travis. But I'm too lazy to compare them by hand.

neithernut:
I did some measurements using a little test program. It queries the number of references matching a glob and prints some counters ("real" time, user and kernel ticks):
```
#include <stdio.h>
#include <sys/times.h>

#include <git2.h>

static const unsigned int runs = 1000;

int countref(const char* name, void* dummy) {
    ++*((unsigned int*) dummy);
    return 0;
}

int main(int argc, char* argv[]) {
    if (argc < 3)
        return 1;

    struct tms res;
    clock_t t1 = times(&res);

    // Run test
    int err;
    git_libgit2_init();

    git_repository* repo;
    err = git_repository_open(&repo, argv[1]);
    if (err != 0)
        return err;

    unsigned int refs = 0;
    for (unsigned int i = 0; i < runs; ++i) {
        err = git_reference_foreach_glob(repo, argv[2], countref, &refs);
        if (err != 0)
            return err;
    }

    git_repository_free(repo);
    git_libgit2_shutdown();

    // Timing
    clock_t t2 = times(&res);
    printf("%ld\t%ld\t%ld\t%d\n", t2 - t1, res.tms_utime, res.tms_stime, refs/runs);
    return 0;
}
```

I ran the program with three sets of repository and glob:
 * query `refs/heads/*` for the libgit2 repository ("libgit2")
 * query `refs/remotes/origin/dit/1f280d33e42df23110c74efcf63d2989d653b3fe/*` for the git-dit repo ("gitdit1")
 * query `refs/remotes/origin/dit/022fb5e39d1b6a292ee3e7ca375fb6e4f0997382/*` for the git-dit repo ("gitdit2")

Sidenote: [git-dit](https://github.com/neithernut/git-dit) is a distributed issue manager of which I am a co-author. We store messages as commits and use references simply for keeping them alive. For each "issue", we have a sub-directory containing those references. The two scenarios involving git-dit fetch those refs for two separate issues.

Each scenario was run 20 times with both the program linked against the feature-branch version of libgit2 ("enhanced") and the merge base ("master", d906a8796cec1fc4d14144f53e505ab215bdab0c).

I found a slight improved performance even for the "libgit2" scenario, although the number of references omitted should be quite low (there are not tags and only a few remote references in addition to the two references I had in the repo). For my target scenarios ("gitdit1" and "gitdit2"), I found an improvement of factors >10.

TL;DR: the patch-set makes globbing refs faster.

----

For those interested, here are the numbers I got:

"master-libgit2":
```
30	25	5	2
28	22	4	2
27	22	4	2
28	22	5	2
28	23	4	2
28	23	5	2
28	22	4	2
28	22	5	2
28	22	5	2
28	21	5	2
28	21	5	2
28	22	5	2
28	23	4	2
28	22	5	2
28	21	5	2
27	23	4	2
27	22	5	2
28	22	4	2
28	23	4	2
29	23	5	2
```

"enhanced-libgit2":
```
25	23	1	2
22	20	1	2
22	20	1	2
22	20	1	2
23	21	1	2
22	20	1	2
22	20	1	2
22	20	1	2
22	20	1	2
22	20	1	2
22	20	1	2
22	21	0	2
23	21	1	2
22	20	1	2
22	19	3	2
23	21	1	2
22	20	1	2
22	20	1	2
22	20	1	2
23	21	1	2
```

"master-gitdit1"
```
110	36	72	6
111	41	69	6
110	39	70	6
110	37	72	6
110	39	69	6
111	41	69	6
111	39	71	6
111	39	72	6
111	41	70	6
115	41	73	6
111	39	71	6
111	38	71	6
110	40	69	6
111	44	66	6
109	36	73	6
110	38	70	6
110	36	73	6
110	37	71	6
109	35	74	6
110	38	71	6
```

enhanced-gitdit1:
```
7	4	3	6
8	2	4	6
7	2	4	6
7	2	4	6
7	2	4	6
8	2	5	6
7	1	5	6
7	2	4	6
7	2	4	6
7	2	5	6
7	3	4	6
7	2	4	6
7	2	4	6
7	3	3	6
8	2	4	6
7	2	4	6
7	2	5	6
8	2	4	6
7	2	5	6
7	2	4	6
```

master-gitdit2:
```
109	44	64	1
107	35	71	1
108	40	67	1
107	35	71	1
106	43	62	1
110	41	68	1
108	37	70	1
107	34	71	1
107	38	67	1
107	37	69	1
108	41	65	1
109	41	67	1
107	37	69	1
107	38	67	1
107	38	68	1
107	36	70	1
106	39	65	1
108	36	71	1
107	38	68	1
108	36	71	1
```

enhanced-gitdit2:
```
2	0	1	1
3	1	1	1
2	1	0	1
2	1	0	1
2	1	1	1
2	1	1	1
2	0	1	1
2	1	1	1
2	1	1	1
2	1	0	1
2	1	0	1
2	1	1	1
3	0	1	1
2	0	1	1
2	0	1	1
2	1	0	1
3	1	0	1
2	0	1	1
2	1	1	1
2	0	1	1
```

src/refdb_fs.c
```diff
@@ -505,26 +505,53 @@ static int iter_load_loose_paths(refdb_fs_backend *backend, refdb_fs_iter *iter)
...
+			break;
+		}
+		if (last_sep) {
+			ref_prefix = iter->glob;
+			ref_prefix_len = (last_sep - ref_prefix) + 1;
+		}
+	}
+
+	if ((error = git_buf_printf(&path, "%s/", backend->commonpath)) < 0 ||
+		(error = git_buf_put(&path, ref_prefix, ref_prefix_len)) ||
```
pks-t: The `< 0` comparison is missing

src/refdb_fs.c
```diff
@@ -505,26 +505,53 @@ static int iter_load_loose_paths(refdb_fs_backend *backend, refdb_fs_iter *iter)
 	git_iterator *fsit = NULL;
 	git_iterator_options fsit_opts = GIT_ITERATOR_OPTIONS_INIT;
 	const git_index_entry *entry = NULL;
+	const char *ref_prefix = GIT_REFS_DIR;
+	int ref_prefix_len = strlen(ref_prefix);
```
pks-t: `strlen` returns `size_t`

src/refdb_fs.c
```diff
@@ -505,26 +505,53 @@ static int iter_load_loose_paths(refdb_fs_backend *backend, refdb_fs_iter *iter)
...
+		for (pos = iter->glob; *pos; ++pos) {
+			switch (*pos) {
+			case '?':
+			case '*':
+			case '[':
+			case '\\':
+				break;
+			case '/':
+				last_sep = pos;
+				/* FALLTHROUGH */
```
pks-t: So we're looking for the first non-literal character here, using the previous literal directories as our base for looking for references. Makes a lot of sense

neithernut:
Those two things should have been obvious to me. Guess I'm too used to `auto` and paranoid `-W`-flags by now...

I'm ready for squashing as soon as I get an approval.

neithernut:
I decided to squash without an additional review, in order to save a round-trip.

pks-t:
Thanks again for your contribution!

Issue #4625, opened 2018-04-13
git_worktree_lookup fails with "out of memory" on a bare repository


mjnz:
### Reproduction steps
1. Open bare repository that has at least one work tree added to it.  git_repository_open(repo, "repo.git")
2. Get the working tree names (which works fine). git_worktree_list(items, repo)
3. Try to get a git_worktree structure (which fails). git_worktree_lookup(wt, repo, items.strings[0])

### Expected behavior
Should succeed (return 0) and allocate git_worktree structure

### Actual behavior
Fails (return -1), giterr_last returns "Out of memory"

### Version of libgit2 (release number or SHA1)
0.26.3

### Operating system(s) tested
Windows

### Analysis
Based on what I can figure it looks like the problem is because a bare repository returns NULL for git_repository_workdir and that causes the error inside open_worktree_dir

#### git_worktree_lookup
https://github.com/libgit2/libgit2/blob/b55bb43d0a733e6100d4eb52719b619281550a00/src/worktree.c#L177
The above call will result in the second parameter (parent) been passed in as NULL

#### open_worktree_dir
https://github.com/libgit2/libgit2/blob/b55bb43d0a733e6100d4eb52719b619281550a00/src/worktree.c#L143

git__strdup will return NULL (because input parameter parent is NULL) which causes it to error out

tiennou:
This has been fixed by #4630. Thanks for the report @mjnz, this was helpful in pinpointing it!

PR #4628, opened 2018-04-18
Worktree of bare repository is not a bare repository

csware:
Fix issue #4626.

PR #4630, opened 2018-04-18
Worktrees can be made from bare repositories

### LINKED ISSUE #4626: Adding files to worktree of bare repository fails
### Reproduction steps

```shell
mkdir tmp
cd tmp
git init --bare base.git

# This is just to create an initial commit, work1 is not related to the bug at all
git clone base.git/ work1
cd work1
git commit --allow-empty -m .
git push
cd ..

# Create a worktree on work2 based on base.git
cd base.git
git worktree add ../work2 master
```

Now use a libgit2 tool to add a file to the index in `work2` using `git_index_add_bypath`.

### Expected behavior

`git_index_add_bypath` succeeds.

### Actual behavior

Error `cannot create blob from file. This operation is not allowed against bare repositories.`

CLI git works.

2a11eaf3de0fd3fd4717ce72d489954a677f4ede and also 937e7e26407b920dde069a3398e33ccfaab61890.

Windows 7 and Windows 10

Cross-ref: https://tortoisegit.org/issue/3200

tiennou:
Supersedes #4628, fixes #4626.

It fixes the issue by globally disabling the check in `ensure_not_bare` when working against a "worktree repo". I haven't given any thought to what this actually can provoke.

My gut feeling says it's not clean enough, as I don't like :
- all those remaining `repo->is_bare` tests which might now not work against a "worktree repo".
- you can have bare repo (`git_repository_is_bare = true`) with a workdir, since the bare-ness of the parent repo percolates down to the worktree repoes.

Maybe renaming `ensure_not_bare` to `ensure_has_workdir` and using that everywhere ? I'm unsure about the 2nd point though. Opinions, as always, welcome.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -268,11 +268,15 @@ static int load_config_data(git_repository *repo, const git_config *config)
 {
 	int is_bare;
 
+	int err = git_config_get_bool(&is_bare, config, "core.bare");
+	if (err < 0 && err != GIT_ENOTFOUND)
+		return err;
+
 	/* Try to figure out if it's bare, default to non-bare if it's not set */
-	if (git_config_get_bool(&is_bare, config, "core.bare") < 0)
-		repo->is_bare = 0;
+	if (err != GIT_ENOTFOUND)
+		repo->is_bare = is_bare && !repo->is_worktree;
 	else
-		repo->is_bare = is_bare;
+		repo->is_bare = 0;
 
 	return 0;
 }
--- a/src/worktree.c
+++ b/src/worktree.c
@@ -139,7 +139,7 @@ static int open_worktree_dir(git_worktree **out, const char *parent, const char 
 	if ((wt->name = git__strdup(name)) == NULL
 	    || (wt->commondir_path = git_worktree__read_link(dir, "commondir")) == NULL
 	    || (wt->gitlink_path = git_worktree__read_link(dir, "gitdir")) == NULL
-	    || (wt->parent_path = git__strdup(parent)) == NULL
+	    || (parent && (wt->parent_path = git__strdup(parent)) == NULL)
 	    || (wt->worktree_path = git_path_dirname(wt->gitlink_path)) == NULL) {
 		error = -1;
 		goto out;
--- a/tests/worktree/worktree.c
+++ b/tests/worktree/worktree.c
@@ -228,6 +228,26 @@ void test_worktree_worktree__init(void)
 	git_repository_free(repo);
 }
 
+void test_worktree_worktree__add_from_bare(void)
+{
+	git_worktree *wt;
+	git_repository *repo, *wtrepo;
+
+	repo = cl_git_sandbox_init("short_tag.git");
+
+	cl_assert_equal_i(1, git_repository_is_bare(repo));
+	cl_assert_equal_i(0, git_repository_is_worktree(repo));
+
+	cl_git_pass(git_worktree_add(&wt, repo, "worktree-frombare", "worktree-frombare", NULL));
+	cl_git_pass(git_repository_open(&wtrepo, "worktree-frombare"));
+	cl_assert_equal_i(0, git_repository_is_bare(wtrepo));
+	cl_assert_equal_i(1, git_repository_is_worktree(wtrepo));
+
+	git_worktree_free(wt);
+	git_repository_free(repo);
+	git_repository_free(wtrepo);
+}
+
 void test_worktree_worktree__add_locked(void)
 {
 	git_worktree *wt;


tests/worktree/worktree.c
```diff
@@ -228,6 +228,30 @@ void test_worktree_worktree__init(void)
...
+void test_worktree_worktree__add_from_bare(void)
+{
+	git_worktree *wt;
+	git_repository *repo, *wtrepo;
+	git_buf path = GIT_BUF_INIT;
+
+	repo = cl_git_sandbox_init("testrepo.git");
+
+	cl_assert_equal_i(1, git_repository_is_bare(repo));
+	cl_assert_equal_i(0, git_repository_is_worktree(repo));
```
csware: As you also modified `git_repository__ensure_not_bare` you should also test that function!

tests/worktree/worktree.c
```diff
@@ -228,6 +228,30 @@ void test_worktree_worktree__init(void)
...
+
+	cl_git_pass(git_buf_joinpath(&path, repo->gitdir, "../worktree-frombare"));
+	cl_git_pass(git_worktree_add(&wt, repo, "worktree-frombare", path.ptr, NULL));
+
+	cl_git_pass(git_repository_open(&wtrepo, path.ptr));
+	cl_assert_equal_i(1, git_repository_is_bare(wtrepo));
+	cl_assert_equal_i(1, git_repository_is_worktree(wtrepo));
+
```
csware: `git_repository__ensure_not_bare` should be tested again on `repo` to make sure that adding a worktree to a bare repository does not change the return value on the bare repository.

ethomson:
> you can have bare repo (git_repository_is_bare = true) with a workdir, since the bare-ness of the parent repo percolates down to the worktree repoes.

To clarify: you're saying that if I have a bare parent repository, the subsequent worktree repository also identifies itself as bare?  I don't think that's what I would have expected...  if you add a working directory to a bare repository, it reports itself (correctly) as non-bare.  I would expect worktree repositories created from a bare parent to do the same.

tiennou:
The unexpectedness is shared then. Note that `cd $worktree; git config --get core.bare` returns "true", so `git` seems to consider the repository still bare. I expect our config code to do the same, but that `is_bare` function clearly does not.

I found one broken thing in `git`, adding a relative submodule to the worktree of a bare repository always "can't find directory $relative_path".

ethomson:
> The unexpectedness is shared then. Note that cd $worktree; git config --get core.bare returns "true", so git seems to consider the repository still bare. I expect our config code to do the same, but that is_bare function clearly does not.

`git config --get core.bare` is really only showing you the configuration option, which remains unchanged.  If you query the repository configuration through libgit2, we should also report `core.bare = true` even when you've configured a working folder.

I think that `git_repository_is_bare` should report the conceptual / perceived bareness of a repository, not just be a direct proxy to the configuration.  I don't think there's a parallel in git.git for this concept.

tests/worktree/worktree.c
csware: But when you test that function any semantic change will be visible in the future.

src/worktree.c
```diff
@@ -139,7 +139,7 @@ static int open_worktree_dir(git_worktree **out, const char *parent, const char
 	if ((wt->name = git__strdup(name)) == NULL
 	    || (wt->commondir_path = git_worktree__read_link(dir, "commondir")) == NULL
 	    || (wt->gitlink_path = git_worktree__read_link(dir, "gitdir")) == NULL
-	    || (wt->parent_path = git__strdup(parent)) == NULL) {
+	    || (parent && (wt->parent_path = git__strdup(parent)) == NULL)) {
```
csware: This also seems to fix https://github.com/libgit2/libgit2/issues/4625

tests/worktree/worktree.c
csware: Also `git_repository__ensure_not_bare` is used by other functions, e.g. adding files to index, which was the initial reason for this fix.  So, therefore this should be tested.

src/repository.c
```diff
@@ -268,11 +268,15 @@ static int load_config_data(git_repository *repo, const git_config *config)
...
 
+	int err = git_config_get_bool(&is_bare, config, "core.bare");
+	if (err < 0 && err != GIT_ENOTFOUND)
+		return err;
+
 	/* Try to figure out if it's bare, default to non-bare if it's not set */
-	if (git_config_get_bool(&is_bare, config, "core.bare") < 0)
-		repo->is_bare = 0;
+	if (err != GIT_ENOTFOUND)
+		repo->is_bare = is_bare && !repo->is_worktree;
```
pks-t: This definitely makes sense.

tests/worktree/worktree.c
pks-t: Agreed

tests/worktree/worktree.c
```diff
@@ -228,6 +228,30 @@ void test_worktree_worktree__init(void)
...
+	git_worktree *wt;
+	git_repository *repo, *wtrepo;
+	git_buf path = GIT_BUF_INIT;
+
+	repo = cl_git_sandbox_init("testrepo.git");
+
+	cl_assert_equal_i(1, git_repository_is_bare(repo));
+	cl_assert_equal_i(0, git_repository_is_worktree(repo));
+
+	cl_git_pass(git_buf_joinpath(&path, repo->gitdir, "../worktree-frombare"));
```
pks-t: You should be able to just use "./worktree-frombare" instead of assembling the path, as we always chdir into the test data directory.

tests/worktree/worktree.c
```diff
@@ -228,6 +228,30 @@ void test_worktree_worktree__init(void)
 	git_repository_free(repo);
 }
 
+void test_worktree_worktree__add_from_bare(void)
+{
+	git_worktree *wt;
+	git_repository *repo, *wtrepo;
+	git_buf path = GIT_BUF_INIT;
+
+	repo = cl_git_sandbox_init("testrepo.git");
```
csware: Use another bare repository here (e.g., `short_tag.git`) fixes the test issue on win32. I suppose the reason is that testrepo is already set up by `setup_fixture_worktree`.

src/repository.c
tiennou: (Note to myself) the commit message should be `We were previously conflating any error into GIT_ENOTFOUND, which might or might not be correct. This fixes the code so a config error is bubbled up, as well as preserving the semantics in the face of worktree-repositories`.

tests/worktree/worktree.c
tiennou: Note that because of the semantic change in `git_repository_is_bare`, `git_repository__ensure_not_bare` now has the same semantics by default. So apart from the special error handling, it's equivalent (and I think it's better to test the public low-level API part vs. a slightly higher-up private one).

pks-t:
I've rebased this branch and just wanted to push it when I ran the tests for a second time, but this time they failed. The problem is that you're actually adding the worktree-frombare repo outside of our sandbox -- the path of `git_worktree_add` is relative to cwd, not relative to the repository we're creating the worktree from.

I've fixed the test issue and conflict and pushed both to your branch.

pks-t:
Thanks @tiennou!

tiennou:
I've just hit that case locally, and was mightily confused 😉. So I looked this PR up and got even more confused by the backport-0.27.2 addition 👆(with the deletion only registering after I banged my 
 head around longer). Was that the de-backporting intentional ? It's just that OCGit is currently tracking 0.27 and this one's now MIA 😕.

pks-t:
I honestly can't remember anymore. I'd be happy to include it in another bugfix release, though, so I'm tagging it for backport

PR #4577, opened 2018-03-11
worktree: Read worktree specific reflog for HEAD

csware:
Wrong reflog for `HEAD` is read for worktree.

Fixes issue #4566.

PR #4632, opened 2018-04-20
Bugfix release v0.27.2

pks-t:
It's been a month since v0.27.0 was released and we have accumulated some fixes on top of master again. To avoid the mess that was v0.26.4, I'm inclined to release v0.27.1 much faster than the previous bugfix releases. Not necessarily this month, but I'd vote to get it out no later than May 23rd, which would be exactly two months after the previous release. I'd be fine with releasing this earlier, though.

### Changes or improvements

* Fix builds with LibreSSL 2.7.

* Fix for `git_diff_status_char()` not returning the correct mapping for
  `GIT_DELTA_TYPECHANGE`.

* Fix for the submodules API not reporting errors when parsing the ".gitmodules"
  file.

* Fix for hiding references in a graph walk not always limiting the graph
  correctly.

* Fix for directory patterns with trailing spaces in attribute files not being
  handled correctly.

* Fix SSH transports not properly disconnecting from the server.

* Update our copy of SHA1DC to fix errors with endianess on some platforms.


csware:
What about #4633 and #4577?

pks-t:
They both look like they should be included. Up to now, I only added already merged PRs. Yours will probably be merged soonish, I'll then append them here. (I don't have enough time to review them right now, but will try to do so soon)

pks-t:
I'm still holding this off for some fixes, most importantly the build fixes around pkg-config.

tiennou:
Also, #4656 should be in there as well ? I'll try to redo a run of the (other) pkg-config/iOS changes as well, hopefully tonight…

pks-t:
No, #4656 shouldn't be part of v0.27.2 as maint/v0.27 does not have any mbedTLS support at all right now.

pks-t:
Now that the security release is out we can get this back on track. This is now going to become v0.27.2

pks-t:
I wanted to get some agreement on this release during our status
call today. But as it seems like we'll have to delay it once
more: are you two good with releasing this as-is or is there
anything that needs to be changed?

ethomson:
:shipit:

pks-t:
Thanks @ethomson. I've just released v0.27.2.

PR #4633, opened 2018-04-20
Fix deletion of unrelated branch on worktree

csware:
libgit2 fails to delete an unrelated branch on the worktree.

Cross-ref: https://gitlab.com/tortoisegit/tortoisegit/issues/3202


pks-t:
Thanks a lot for this fix!

PR #4635, opened 2018-04-20
Leak fixes for v0.27.1

tiennou:
This fixes the 2 Valgrind-reported leaks that can be backported to v0.27.1


pks-t:
Thanks as always!

PR #4637, opened 2018-04-20
Provide a wrapper for simple submodule clone steps

tiennou:
While working on the submodule-in-worktree issue, I had to finagle with #4149. So let's add that to the API.

Arguably, it could be named `git_submodule_add` and perform all the steps by itself.

### DIFF
--- a/include/git2/submodule.h
+++ b/include/git2/submodule.h
@@ -263,7 +263,8 @@ GIT_EXTERN(int) git_submodule_foreach(
  * from the working directory to the new repo.
  *
  * To fully emulate "git submodule add" call this function, then open the
- * submodule repo and perform the clone step as needed.  Lastly, call
+ * submodule repo and perform the clone step as needed (if you don't need
+ * anything custom see `git_submodule_add_clone()`). Lastly, call
  * `git_submodule_add_finalize()` to wrap up adding the new submodule and
  * .gitmodules to the index to be ready to commit.
  *
@@ -284,6 +285,22 @@ GIT_EXTERN(int) git_submodule_add_setup(
 	const char *url,
 	const char *path,
 	int use_gitlink);
+
+/**
+ * Perform the clone step for a newly created submodule.
+ *
+ * This performs the necessary `git_clone` to setup a newly-created submodule.
+ *
+ * @param out The newly created repository object. Optional.
+ * @param submodule The submodule currently waiting for its clone.
+ * @param opts The options to use.
+ *
+ * @return 0 on success, -1 on other errors (see git_clone).
+ */
+GIT_EXTERN(int) git_submodule_clone(
+	git_repository **out,
+	git_submodule *submodule,
+	const git_submodule_update_options *opts);
 
 /**
  * Resolve the setup of a new git submodule.
--- a/src/clone.c
+++ b/src/clone.c
@@ -382,11 +382,12 @@ done:
 	return is_local;
 }
 
-int git_clone(
+static int git__clone(
 	git_repository **out,
 	const char *url,
 	const char *local_path,
-	const git_clone_options *_options)
+	const git_clone_options *_options,
+	int use_existing)
 {
 	int error = 0;
 	git_repository *repo = NULL;
@@ -403,7 +404,7 @@ int git_clone(
 	GIT_ERROR_CHECK_VERSION(&options, GIT_CLONE_OPTIONS_VERSION, "git_clone_options");
 
 	/* Only clone to a new directory or an empty directory */
-	if (git_path_exists(local_path) && !git_path_is_empty_dir(local_path)) {
+	if (git_path_exists(local_path) && !use_existing && !git_path_is_empty_dir(local_path)) {
 		git_error_set(GIT_ERROR_INVALID,
 			"'%s' exists and is not an empty directory", local_path);
 		return GIT_EEXISTS;
@@ -453,6 +454,24 @@ int git_clone(
 
 	*out = repo;
 	return error;
+}
+
+int git_clone(
+	git_repository **out,
+	const char *url,
+	const char *local_path,
+	const git_clone_options *_options)
+{
+	return git__clone(out, url, local_path, _options, 0);
+}
+
+int git_clone__submodule(
+	git_repository **out,
+	const char *url,
+	const char *local_path,
+	const git_clone_options *_options)
+{
+	return git__clone(out, url, local_path, _options, 1);
 }
 
 int git_clone_options_init(git_clone_options *opts, unsigned int version)
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -23,6 +23,7 @@
 #include "path.h"
 #include "index.h"
 #include "worktree.h"
+#include "clone.h"
 
 #define GIT_MODULES_FILE ".gitmodules"
 
@@ -812,6 +813,64 @@ int git_submodule_repo_init(
 done:
 	git_config_free(cfg);
 	git_buf_dispose(&buf);
+	return error;
+}
+
+static int clone_return_origin(git_remote **out, git_repository *repo, const char *name, const char *url, void *payload)
+{
+	GIT_UNUSED(url);
+	GIT_UNUSED(payload);
+	return git_remote_lookup(out, repo, name);
+}
+
+static int clone_return_repo(git_repository **out, const char *path, int bare, void *payload)
+{
+	git_submodule *sm = payload;
+
+	GIT_UNUSED(path);
+	GIT_UNUSED(bare);
+	return git_submodule_open(out, sm);
+}
+
+int git_submodule_clone(git_repository **out, git_submodule *submodule, const git_submodule_update_options *given_opts)
+{
+	int error;
+	git_repository *clone;
+	git_buf rel_path = GIT_BUF_INIT;
+	git_submodule_update_options sub_opts = GIT_SUBMODULE_UPDATE_OPTIONS_INIT;
+	git_clone_options opts = GIT_CLONE_OPTIONS_INIT;
+
+	assert(submodule);
+
+	if (given_opts)
+		memcpy(&sub_opts, given_opts, sizeof(sub_opts));
+
+	GIT_ERROR_CHECK_VERSION(&sub_opts, GIT_SUBMODULE_UPDATE_OPTIONS_VERSION, "git_submodule_update_options");
+
+	memcpy(&opts.checkout_opts, &sub_opts.checkout_opts, sizeof(sub_opts.checkout_opts));
+	memcpy(&opts.fetch_opts, &sub_opts.fetch_opts, sizeof(sub_opts.fetch_opts));
+	opts.repository_cb = clone_return_repo;
+	opts.repository_cb_payload = submodule;
+	opts.remote_cb = clone_return_origin;
+	opts.remote_cb_payload = submodule;
+
+	git_buf_puts(&rel_path, git_repository_workdir(git_submodule_owner(submodule)));
+	git_buf_joinpath(&rel_path, git_buf_cstr(&rel_path), git_submodule_path(submodule));
+
+	GIT_ERROR_CHECK_ALLOC_BUF(&rel_path);
+
+	error = git_clone__submodule(&clone, git_submodule_url(submodule), git_buf_cstr(&rel_path), &opts);
+	if (error < 0)
+		goto cleanup;
+
+	if (!out)
+		git_repository_free(clone);
+	else
+		*out = clone;
+
+cleanup:
+	git_buf_dispose(&rel_path);
+
 	return error;
 }
 
--- a/tests/clone/nonetwork.c
+++ b/tests/clone/nonetwork.c
@@ -1,7 +1,6 @@
 #include "clar_libgit2.h"
 
 #include "git2/clone.h"
-#include "git2/sys/commit.h"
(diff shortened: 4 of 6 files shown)


src/submodule.c
```diff
@@ -832,6 +832,47 @@ int git_submodule_repo_init(
...
+	git_clone_options opts = GIT_CLONE_OPTIONS_INIT;
+
+	assert(submodule);
+
+	opts.repository_cb = clone_return_repo;
+	opts.repository_cb_payload = submodule;
+	opts.remote_cb = clone_return_origin;
+	opts.remote_cb_payload = submodule;
+
+	error = git_clone(&clone, git_submodule_url(submodule), git_submodule_path(submodule), &opts);
```
pks-t: I'm surprised this actually works in the test suite. The problem is that when I've just initialized a submodule via `git_submodule_add_setup`, it will already contain the ".git" gitlink to the actual gitdir of the submodule. Because of that, `git_clone` will refuse to clone into an non-empty directory. And in fact it does fail when changing my example code to use this function instead of implementing the clone step myself. Puzzled

pks-t: Never mind, now it works. I think the examples just didn't build. And in fact the `clone_return_repo` callback should be the reason why it succeeds

pks-t: Argh. I've managed to hard-reset my adjustment to use your new API, so never mind my never mind comment. It does fail in `git_clone`, line 406, where it checks whether the local path is an empty directory or not.

pks-t: Ah, I got it. `git_submodule_path` returns a relative path, which is why `git_clone` thinks that the target path doesn't yet exist. You have to create an absolute path for submodule relative to the parent repo's working directory, which will subsequently break your test as `git_clone` will error out due to files existing in the target directory.

Sorry for my confused comments

src/submodule.c
```diff
@@ -832,6 +832,47 @@ int git_submodule_repo_init(
...
+static int clone_return_repo(git_repository **out, const char *path, int bare, void *payload)
+{
+	git_submodule *sm = payload;
+
+	GIT_UNUSED(path);
+	GIT_UNUSED(bare);
+	return git_submodule_open(out, sm);
+}
+
+int git_submodule_add_clone(git_repository **out, git_submodule *submodule)
```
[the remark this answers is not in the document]
tiennou: I'd be tempted to say that the "old" way already provides this, and shrug it off. I'm adding this API for the 95% of users that just want to `git submodule add` something and don't care for wacky remote configurations.

tiennou:
Now I don't know what changes you're requesting anymore 😉.

pks-t:
Sorry for my comments being confusing :P

The thing is that your function unfortunately doesn't work. The test you've written works by accident due to a bug in `git_submodule_add_clone`, as you clone into `git_submodule_path(submodule)`, which returns a path relative to the parent repo. So the `git_clone` will in fact clone into the current working directory and not into the parent repo. What you have to do is to make `git_submodule_path(submodule)` an absolute path relative to the parent repo and then do the clone. This is where the next issue will surface: as the submodule directory already exists and due to a ".git" gitlink file existing inside of it, the `git_clone` step will error out due to the target directory not being empty.

tiennou:
Okay, I fixed the function, so yeah the test wasn't working as expected because of pathing issues (which I wasn't expecting so, good catch 😉).

There's two ways to fix : either the clone existence check is wrong, and this should be changed, or it's not, and the clone must be performed manually.

Now for the discussion part, is clone wrong ? We're using the "approved" `clone_opts->repository_cb` function for grabbing a user-provided repository, but the emptiness check happens first. I can think of a few workarounds : 
- trust that the user gave us something sane via the `repository_cb`, in that it would have called `git_repository_open` to be able to do that. The definition of sane in that case is fuzzy though.
- make the empty check more precise by saying that empty allows a valid gitlink at the root of the folder, but that looks to be a deviation from git (I've checked against a manually crafted gitlink). So I'd make it a 👎.

pks-t:
I don't like the second option, it's much too specific. While I think the first option is quite nice, I don't know if we really want to relax the check like that. So, option number three:

- create a permit-clone-into-a-populated-directory flag (defaulting to `false`) in `git_clone_options`, and if it is set  we skip the check for emptiness

tiennou:
I've gone a 4th way: add a private variant of `git_clone` with a relaxed check and use that. That restricts it to this specific function call, which is sufficient, and keep it from being exposed in the API (less API to hang you with).

src/clone.c
```diff
@@ -382,11 +382,12 @@ int git_clone__should_clone_local(const char *url_or_path, git_clone_local_t loc
 	return is_local;
 }
 
-int git_clone(
+static int git__clone(
```
pks-t: `clone` would be sufficient, I guess :)

src/submodule.c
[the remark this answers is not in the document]
pks-t: This function only clones and doesn't do the add, right? Wouldn't `git_submodule_clone` be easier to understand in that case?

src/submodule.c
```diff
@@ -815,6 +816,54 @@ int git_submodule_repo_init(
...
+
+	assert(submodule);
+
+	opts.repository_cb = clone_return_repo;
+	opts.repository_cb_payload = submodule;
+	opts.remote_cb = clone_return_origin;
+	opts.remote_cb_payload = submodule;
+
+	git_buf_puts(&rel_path, git_repository_workdir(git_submodule_owner(submodule)));
+	git_buf_joinpath(&rel_path, git_buf_cstr(&rel_path), git_submodule_path(submodule));
```
pks-t: These two need error checking.

tests/clone/nonetwork.c
```diff
@@ -353,44 +353,19 @@ void test_clone_nonetwork__clone_from_empty_sets_upstream(void)
...
-
 	/* Create the submodule structure, clone into it and finalize */
 	cl_git_pass(git_submodule_add_setup(&sm, g_repo, cl_fixture("testrepo.git"), "testrepo", true));
-
-	clone_opts.repository_cb = just_return_repo;
-	clone_opts.repository_cb_payload = sm;
-	clone_opts.remote_cb = just_return_origin;
-	clone_opts.remote_cb_payload = sm;
-	cl_git_pass(git_clone(&sm_repo, cl_fixture("testrepo.git"), "testrepo", &clone_opts));
+	cl_git_pass(git_submodule_add_clone(NULL, sm));
```
pks-t: I think we should have a new test that uses `git_submodule_add_clone`, as we should retain the test that verifies the old way of doing that.

src/submodule.c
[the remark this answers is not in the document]
tiennou: I… hadn't anticipated it would make sense 😆. My thinking was that, being the step between `add_setup` and `add_finalize`, it made sense to have it under that namespace. If it so happens that it can be used to clone actual uncloned submodules, then why not, but that wasn't the plan (and I'm eyeing towards the unimplemented checkout actions 😉).

src/clone.c
tiennou: I think you got me there, I just had the Linux buildbots remind me that it's some syscall 🤣.

tiennou:
Apart from the naming question, this is good to go (my reply somehow doesn't show up from the conversation view though).

src/submodule.c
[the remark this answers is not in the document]
pks-t: So, uh. Can it be used to clone uncloned submodules right now? If so, I'd rename it. Otherwise it's fine as is, I guess.

tiennou:
Rebased, I've renamed the function to `git_submodule_clone`, and I've moved its tests to submodule/add.c (it seemed more relevant). I hope that's fine.

include/git2/submodule.h
```diff
@@ -285,6 +286,20 @@ GIT_EXTERN(int) git_submodule_add_setup(
...
+ * This performs the necessary `git_clone` to setup a newly-created submodule.
+ *
+ * @param out The newly created repository object. Optional.
+ * @param submodule The submodule currently waiting for its clone.
+ *
+ * @return 0 on success, -1 on other errors (see git_clone).
+ */
+GIT_EXTERN(int) git_submodule_clone(
+	git_repository **out,
+	git_submodule *submodule);
```
pks-t: To not repeat past mistakes, do we maybe want to introduce an options struct here from the get-go? Doesn't need to contain anything options for now

pks-t:
Thanks, @tiennou! Looks good to me except for the one comment I have

include/git2/submodule.h
tiennou: I usually like having a default "do the default thing" and provide an `_ext` on top of that, if/when the need arises. But it feels a little premature (though only because of a lack of use cases, so if you have a definite not too difficult "this is needed" example, I'll reconsider).

ethomson: Hmm.  I would very quickly want progress reporting if I were to use this in anger.  I concede that this is a bit tricky.  I would definitely want, eg, progress reporting but many fields should not be eligible to be set by an end-user, so taking a whole `git_clone_options` may not be appropriate.  Additionally, we may need to do some proxy'ing with some of the callbacks to ensure that payloads are delivered correctly.

Having said that, progress reporting and cancellation still feels like table-stakes for any function that does network I/O.

pks-t: > I usually like having a default "do the default thing" and provide an _ext on top of that, if/when the need arises.

Well, defaults are easily attainable by just passing `NULL`. Personally, I'm not too happy with the `_ext` suffix and I see it rather as a "Okay we f'd up the first time, so let's just tack on a suffix and fix up the function".

pks-t: And yeah, fully agree with the usecases @ethomson proposes.

tiennou: I've looked at it a little, and effectively, just accepting a `git_clone_options` seems a bit too much, and since you could just do the clone yourself, a more limited but easier to use API seems nicer. I'm just not sure how to balance what to expose, especially if it requires another struct 😟.

So, from the clone struct, we're not using :
- `git_checkout_options`: loads of things, but apart from the notify/progress/prefdata, I don't think anything else makes sense.
- `git_remote_options`: `callbacks` looks like a given, not sure about the rest, but I suspect it would be better if this one was accessible.
- `bare`, `local` : I don't think those need to be exposed (though maybe `local` ?)
- `checkout_branch`: sounds related, but I think that's not up to the user at this point ? ie, the submodule should have been setup correctly in the first place.

```c
typedef struct {
/*…*/
 git_checkout_(notify|progress|pref)_cb $step_checkout_cb;
 void *checkout_payload;
 git_remote_options remote_callbacks;
} git_submodule_clone_options;
```

All those callbacks make me want to audit them and see if they couldn't be make more generic 😭.

ethomson: Hmm, yeah that's a tough decision.  I can see pros and cons to simply taking a `git_clone_options` struct and just ignoring and overwriting parts of it (documenting this of course).  Or a simpler struct and proxying that data into a `git_clone_options`.

The former seems like the simpler strategy for you (the implementer).  What I'm not sure of is whether it's easier for users of the library, which is of course where I think the important bits are.

pks-t: Simply taking `git_clone_options` is not going to fly in case there's going to be submodule-cloning specific options at some point in time, which I'd say is not all that unlikely. So I think it should definitely be a new `git_submodule_clone_options` struct, but as you said it may just embed and forward (parts of) the `git_clone_options` struct.

tiennou: I've noticed that `git_submodule_update_options` seemed to have mostly what we needed, so I've added this as our option struct. Sorry, I forgot to rebase-ping, but IIRC it was ready to 🚢 as well.

pks-t: Oh, didn't see that, sorry.

`git_submodule_update_options` does look similar to what we need, that's true. I'm still a bit hesitant, though. It includes options that cannot fit the bill right now and we may need to extend the struct for either `git_submodule_update` or `git_submodule_clone` with options that are incompatible with the respective other function. If you ask me, that's trading ease of implementation (no need for an additional struct) for a confusing API. I'm more in the camp of using one struct per function whenever it's not obvious that they're always going to take the same set of opts.

Just my 2 cents -- I'm happy to be overruled here due to my recent lack of time for libgit2.

tiennou: (Struct for reference)
```c
typedef struct git_submodule_update_options {
	unsigned int version;
	git_checkout_options checkout_opts;
	git_fetch_options fetch_opts;
	int allow_fetch;
} git_submodule_update_options;
```

Of theses, `allow_fetch` feels the most dubious, since we don't have any repository before the call. `checkout` seems [fine](https://libgit2.org/libgit2/#HEAD/type/git_checkout_options) (though I think there's a few rough edges), and `fetch` looks [okay](https://libgit2.org/libgit2/#HEAD/type/git_fetch_options).

Now, I don't have that much experience with submodules, but a shared "those are the settings to obey if an update needs to happen", especially given the `git submodule update --init --recursive` thingie seems to be a nice thing. Writing in the docblock that `allow_fetch` only makes sense if the submodule is already initialized, and ignoring it if it's set seems sufficient.

pks-t:
Rebased your branch to fix conflicts.

pks-t:
I've removed the homegrown clone test, which didn't do what one expected (the resulting submodule was cloned outside of the parent repo) and added two new tests

tiennou:
Thanks for keeping up with this ! ✨

PR #4639, opened 2018-04-24
worktree: Add new git_worktree_name and git_worktree_path functions

mkeeler:
I was starting to write some code using the rust git2-rs library utilizing worktrees and found that there was no way given an already open worktree to look at what the name of the worktree is or to figure out where on the filesystem it is expected to live. The main goal behind this was to be able to get at the information that would be available in `git worktree list` command output without having to iterate through the worktree strarray and open a git_repository for each worktree and then query things. This PR adds two new worktree related functions. git_worktree_name just copies the name field of the internal git_worktree structure into the provided git_buf. git_worktree_path puts the dirname of the gitlink_path into the provided git_buf. 

The tests for git_worktree_path do string comparisons on two paths and on macOS the default temp/var dirs are usually symlinked. This made direct string comparisons impossible as the path as stored in the worktree structure had links already resolved. So the sandbox.h changes attempt to get the realpath for the tmp directory on macOS.

One thing I had questions about regarding the PR was whether I should include the CHANGELOG.md updates in my PR or whether that was handled by a maintainer.

### DIFF
--- a/include/git2/worktree.h
+++ b/include/git2/worktree.h
@@ -153,19 +153,19 @@ GIT_EXTERN(int) git_worktree_is_locked(git_buf *reason, const git_worktree *wt);
  * Retrieve the name of the worktree
  *
  * @param wt Worktree to get the name for
- * @param name Buffer to store the name in. If NULL an error will be returned
- * @return 0 when the name was stored in the buffer, error-code otherwise
+ * @return The worktrees name or NULL for an error. The pointer returned
+ *  is valid for the lifetime of the git_worktree
  */
-GIT_EXTERN(int) git_worktree_name(const git_worktree *wt, git_buf *name);
+GIT_EXTERN(const char *) git_worktree_name(const git_worktree *wt);
 
 /**
  * Retrieve the filesystem path for the worktree
  *
  * @param wt Worktree to get the path for
- * @param path Buffer to store the name in. If NULL an error will be returned
- * @return 0 when the path was stored in the buffer, error-code otherwise
+ * @return The worktrees filesystem path or NULL for an error. The pointer
+ *  returned is valid for the lifetime of the git_worktree.
  */
-GIT_EXTERN(int) git_worktree_path(const git_worktree *wt, git_buf *path);
+GIT_EXTERN(const char *) git_worktree_path(const git_worktree *wt);
  
 /**
  * Flags which can be passed to git_worktree_prune to alter its
--- a/include/git2/worktree.h
+++ b/include/git2/worktree.h
@@ -148,6 +148,25 @@ GIT_EXTERN(int) git_worktree_unlock(git_worktree *wt);
  */
 GIT_EXTERN(int) git_worktree_is_locked(git_buf *reason, const git_worktree *wt);
 
+
+/**
+ * Retrieve the name of the worktree
+ *
+ * @param wt Worktree to get the name for
+ * @param name Buffer to store the name in. If NULL an error will be returned
+ * @return 0 when the name was stored in the buffer, error-code otherwise
+ */
+GIT_EXTERN(int) git_worktree_name(const git_worktree *wt, git_buf *name);
+
+/**
+ * Retrieve the filesystem path for the worktree
+ *
+ * @param wt Worktree to get the path for
+ * @param path Buffer to store the name in. If NULL an error will be returned
+ * @return 0 when the path was stored in the buffer, error-code otherwise
+ */
+GIT_EXTERN(int) git_worktree_path(const git_worktree *wt, git_buf *path);
+ 
 /**
  * Flags which can be passed to git_worktree_prune to alter its
  * behavior.
--- a/src/worktree.c
+++ b/src/worktree.c
@@ -144,7 +144,7 @@ static int open_worktree_dir(git_worktree **out, const char *parent, const char 
 		goto out;
 	}
 	
-	if ((error == git_path_dirname_r(&buf, wt->gitlink_path)) < 0)
+	if ((error = git_path_dirname_r(&buf, wt->gitlink_path)) < 0)
 		goto out;
 	wt->worktree_path = git_buf_detach(&buf);
 	
--- a/src/worktree.c
+++ b/src/worktree.c
@@ -455,6 +455,26 @@ out:
 	return ret;
 }
 
+int git_worktree_name(const git_worktree *wt, git_buf *name)
+{
+	if (!name || !wt)
+		return -1;
+	
+	git_buf_clear(name);
+	return git_buf_puts(name, wt->name);	
+}
+
+int git_worktree_path(const git_worktree *wt, git_buf *path)
+{
+	if (!path || !wt)
+		return -1;
+	
+	git_buf_clear(path);
+	if ( git_path_dirname_r(path, wt->gitlink_path) > 0 )
+		return 0;
+	return -1;
+}
+
 int git_worktree_prune_init_options(
 	git_worktree_prune_options *opts,
 	unsigned int version)
--- a/src/worktree.c
+++ b/src/worktree.c
@@ -122,7 +122,7 @@ out:
 
 static int open_worktree_dir(git_worktree **out, const char *parent, const char *dir, const char *name)
 {
-	git_buf gitdir = GIT_BUF_INIT;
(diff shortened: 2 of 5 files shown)


ethomson:
Hi @mkeeler, thanks for opening a pull request!

Is there a reason that you prefer a `git_buf` here?  Generally we try to avoid using that type - there are a _couple_ places that it leaks out for various interoperability reasons.  But for `git_worktree_name`, at least, it seems like a more direct implementation would be to return the `name` directly.  eg;

```
const char *git_worktree_name(const git_worktree *wt)
{
    return wt->name;
}
```

This is generally quite similar to other "accessors" on "objects".  (That's how we think of them, even if there isn't proper language support for these concepts in C, so I hope you're not rolling your eyes _too_ much at my usage of these words.  😁 )

Anyway.  The nice thing about this is that it avoids a malloc for people who don't need a copy of the name.  (Maybe you're just printing it out to the console, for example.)  Some languages FFIs will need to copy a string into their own format (Java, C#, for example) and so they would avoid an unnecessary malloc and copy there, too.

If I had to guess, I'm thinking that you might have done this for parity with `git_worktree_path` which doesn't return a member of the `worktree` struct directly, and instead mutates a copy of it.  I'm 👍 for parity; but I wonder if we could arrive at two functions that have a similar looking API.  Maybe we create the path, stuff it in the worktree on-demand and return a `const char *` to it.  Its lifecycle would mirror that of the `worktree` struct itself.  We could even move that path creation logic (the `basename`) call to when the `worktree` struct is created, I don't know what makes more sense, I'm just thinking out loud (or in a textbox) right now.
 
I have some stylistic nits ("out" params - if you will - come at the beginning of function declarations, so the prototype would be `int git_worktree_name(git_buf *buf, const git_worktree *wt)`.  But if we change the prototype too much then they become moot so let's hold off on those.

Anyway.  I'd love to hear your thoughts.  Thanks for a great first PR, I think this is useful functionality to have.

src/worktree.c
```diff
@@ -455,6 +455,26 @@ int git_worktree_is_locked(git_buf *reason, const git_worktree *wt)
...
+	return git_buf_puts(name, wt->name);	
+}
+
+int git_worktree_path(const git_worktree *wt, git_buf *path)
+{
+	if (!path || !wt)
+		return -1;
+	
+	git_buf_clear(path);
+	if ( git_path_dirname_r(path, wt->gitlink_path) > 0 )
```
csware: Please stick to existing coding style.

mkeeler:
@ethomson I was looking at some of the other worktree functions like `git_worktree_is_locked` and saw it was using a git_buf for output of the reason so I figured that was the standard way of doing it.
I agree that returning a const char * with the expected lifetime of the `git_worktree` would be a much nicer API to use.

As for `git_worktree_path` adding a worktree_path member to the worktree structure and just returning a const char * there would also be great. I would actually prefer this. I will make the necessary updates and post them soon.

Also I may have found a bug regarding worktree validation. If you were to use `git_worktree_lookup` on a bare repository then later call `git_worktree_validate` on it I believe this will result in either 1) assertion failure in `git_path_exists` for a NULL path, 2) potentially a segfault/invalid read in `git_path_exists` when assertions are disabled if the underlying libc doesn't null check it or 3) `git_worktree_validate` could return that a worktree is invalid even when it should be valid. This could be fixed by null checking the parent_path and ignoring that check in `git_worktree_validate` if NULL.

For my personal usage I usually bare clone repos and then create worktrees from them and the end goal for the project I am working on is to have a cli app to manage this workflow where a bare clone is done and then multiple worktrees get created from it. This works correctly with normal

src/worktree.c
mkeeler: @csware I would definitely like to. Could you tell me what specifically doesn't fit the coding style? Is it just the spaces after/before the parens in the if or something else too?

tiennou:
As you have noticed with the validation, we're investigating how worktrees should work against bare repoes in #4630. This should be fixed soon.

mkeeler:
@tiennou That sounds great. That other PR doesn't appear to address the bug in `git_worktree_validate`. You prevent the strdup of the parent path when NULL but we need to also ignore the parent path check in the validation function. Are you still working on changes to that PR or should I make the updates to mine?

src/worktree.c
```diff
@@ -455,24 +460,16 @@ int git_worktree_is_locked(git_buf *reason, const git_worktree *wt)
 	return ret;
 }
 
-int git_worktree_name(const git_worktree *wt, git_buf *name)
+const char * git_worktree_name(const git_worktree *wt)
```
tiennou: Nitpick: I think the style guide would say "left padded * operators", so `type *func(...)`.

include/git2/worktree.h
```diff
@@ -148,6 +148,25 @@ GIT_EXTERN(int) git_worktree_unlock(git_worktree *wt);
  */
 GIT_EXTERN(int) git_worktree_is_locked(git_buf *reason, const git_worktree *wt);
 
+
+/**
+ * Retrieve the name of the worktree
+ *
+ * @param wt Worktree to get the name for
+ * @return The worktrees name or NULL for an error. The pointer returned
+ *  is valid for the lifetime of the git_worktree
```
tiennou: Nitpick: no leading " ".

About the message itself, I'd be tempted to not say anything about NULL being an error. AFAICT, having no idea what's the underlying implementation, a worktree with no name would be a pretty crazy thing to happen, so it should have been caught way earlier.

include/git2/worktree.h
```diff
@@ -148,6 +148,25 @@ GIT_EXTERN(int) git_worktree_unlock(git_worktree *wt);
...
+ */
+GIT_EXTERN(const char *) git_worktree_name(const git_worktree *wt);
+
+/**
+ * Retrieve the filesystem path for the worktree
+ *
+ * @param wt Worktree to get the path for
+ * @return The worktrees filesystem path or NULL for an error. The pointer
+ *  returned is valid for the lifetime of the git_worktree.
```
tiennou: Ditto.

mkeeler:
@tiennou  If you like I can rework the PR into just those two commits. Its no problem at all. I totally understand wanting to keep the git history clean and organized. Additionally with needing another commit to finish off fixing your last two points I think it would be best. 

As for the /private/tmp thing, I dont think its generally an issue and I don't believe other tests are doing string comparison on paths containing the clar sandbox path.

So I guess we can close this PR and I will open another one with the reworked branch and two commits.

src/worktree.c
```diff
@@ -143,10 +143,14 @@ static int open_worktree_dir(git_worktree **out, const char *parent, const char
 		error = -1;
 		goto out;
 	}
-
-	if ((error = git_path_prettify_dir(&gitdir, dir, NULL)) < 0)
+
```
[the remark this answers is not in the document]
mkeeler: `git_path_dirname_r` is only necessary because the worktrees fs path should be the parent directory of the .git file within a worktree. It is unrelated to the /private/tmp thing. In the new branch I am creating I see that I could do it like this instead

```
	if ((wt->name = git__strdup(name)) == NULL
	    || (wt->commondir_path = git_worktree__read_link(dir, "commondir")) == NULL
	    || (wt->gitlink_path = git_worktree__read_link(dir, "gitdir")) == NULL
	    || (wt->parent_path = git__strdup(parent)) == NULL
	    || (wt->worktree_path = git_path_dirname(wt->gitlink_path)) == NULL) {
		error = -1;
		goto out;
	}
```

This seems a little cleaner to me and is clearer that it is unrelated to the other gitdir.
```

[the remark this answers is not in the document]
tiennou: Agreed.

tiennou:
Reworking would be great, thanks for that.

To be frank, I'm one of the heavy-rebasers around here (because I prefer having the ability to checkout anywhere without ending up with something "broken"), and I remember that stance was not *really* shared (for example, things like bugfixes on top of ongoing work is usually ok). This means I usually reuse my PRs, which keeps context available.

About /private/tmp, I think it's because the library rarely returns absolute paths (they're usually repo-root-relative), so your belief is correct 😉. I was just surprised it never came up.

mkeeler:
Closing this as I opened a new PR after consolidating some git history here: https://github.com/libgit2/libgit2/pull/4640

PR #4641, opened 2018-04-26
Detect duplicated submodules for the same path

pks-t:
We currently have a memory leak in AppVeyor wrt handling submodules. The issue is that we're accepting duplicated submodules for the same path, silently updating an already existing key in the submodule names hash map and leaking its previous value. Git does not allow for duplicated submodule paths, so we shouldn't either.

### DIFF
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -572,7 +572,6 @@ int git_submodule__map(git_repository *repo, git_strmap *map)
 	git_buf path = GIT_BUF_INIT;
 	git_submodule *sm;
 	git_config *mods = NULL;
-	uint32_t mask;
 
 	assert(repo && map);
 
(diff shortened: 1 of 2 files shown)


carlosmn:
Do we want some tests here to make sure we reject the right things?

pks-t:
Probably, but I don't have enough time right now. I was very surprised to see that my changes didn't break anything at all. The "submodules" test repository has a gitmodules file with two modules pointing to the same path, so I'd have expected us erroring out to have at least _some_ kind of visible change in behaviour.

carlosmn:
Then we should be expecting this fail, and since it's not that seems like a failing test on its own.

pks-t:
Rebased to fix conflicts with the submodule name validation fix. This now got a test which detects the issue and works on Windows. This plugs all currently known memory leaks.

pks-t:
@carlosmn: you got any more feedback on this PR? I'd like to get this merged soon, as I think it should be part of v0.27.2

PR #4642, opened 2018-04-27
cmake: resolve libraries found by pkg-config 

### LINKED ISSUE #4120: libssh2_userauth_publickey_frommemory not found in static libssh2
When I compile the library using a static libssh2 I get the following message:

```
-- Looking for libssh2_userauth_publickey_frommemory in ssh2
-- Looking for libssh2_userauth_publickey_frommemory in ssh2 - not found
```

This happens because CHECK_LIBRARY_EXISTS macro in CMakeLists.txt is not working properly with my libssh2.a .

```
CHECK_LIBRARY_EXISTS("${LIBSSH2_LIBRARIES}" libssh2_userauth_publickey_frommemory "${LIBSSH2_LIBRARY_DIRS}" HAVE_LIBSSH2_MEMORY_CREDENTIALS)
```

The LIBSSH2_* constants are ok and the object file contains the right symbols:
```
$ nm -g libssh2/install/lib64/libssh2.a | grep libssh2_userauth_publickey_frommemory
000000000000338a T libssh2_userauth_publickey_frommemory
```

Any idea?

### Expected behavior
```
-- Looking for libssh2_userauth_publickey_frommemory in ssh2
-- Looking for libssh2_userauth_publickey_frommemory in ssh2 - found
```

### Actual behavior
```
-- Looking for libssh2_userauth_publickey_frommemory in ssh2
-- Looking for libssh2_userauth_publickey_frommemory in ssh2 - not found
```

104a1b0bd997433a1838b4583c6604f32953344c

Arch Linux 4.9.6-1-ARCH x86_64

pks-t:
Libraries found by CMake modules are usually handled with their full
path. This makes linking against those libraries a lot more robust when
it comes to libraries in non-standard locations, as otherwise we might
mix up libraries from different locations when link directories are
given.

One excemption are libraries found by PKG_CHECK_MODULES. Instead of
returning libraries with their complete path, it will return the
variable names as well as a set of link directories. In case where
multiple sets of the same library are installed in different locations,
this can lead the compiler to link against the wrong libraries in the
end, when link directories of other dependencies are added.

To fix this shortcoming, we need to manually resolve library paths
returned by CMake against their respective library directories. This is
an easy task to do with `FIND_LIBRARY`.

---

I'm not too sure about the second patch yet. I think we don't need `LIBGIT2_LIBDIRS` anymore due to the change, and getting rid of it is definitely a nice thing. Let's see what CI has to say about it.

This fixes #4120 and partially solves the issue in #4604


tiennou:
@pks-t Can this be scheduled for backport-0.27.2 ?

pks-t:
Sure, I actually intended to backport it to the bugfix release.
Will do so towards the end of this week and get v0.27.2 back on
track now that the security release is finally done.

Keno:
Hi there,

this change seems to have broken cross compile from linux to Windows via mingw.
In my particular case, libssh2 depends on bcrypt via import library (mingw's libbcrypt.a),
which is listed in the pkg-config file. However, using this approach to find the library fails
to find that import library leading to libgit2 refusing to build.

pks-t:
So the initial configuration of the build fails because CMake is now unable to find the transitive dependency libbcrypt.a, which is specified in libssh2's pkg-config file? To me this sounds like the pkg-config file is erroneous, then, as it would have to specify both the linking directory of libssh2 itself as well as the linking directory of libbcrypt.a. Otherwise it is expected that we are not able to resolve libbcrypt.a, and bailing out is the correct thing to do. Do both of these libraries live in different directories?

Please correct me if I misunderstood your issue.

Issue #3676, opened 2016-03-11
Support .gitattributes for bare repositories


matto1990:
I have been looking into why Gitlab (which uses Rugged, which wraps libgit2) does not respect marking files as binary in the diffs on merge requests.

After tracing it down to libgit2 it seems that the library does correctly support marking files as binary through `.gitattributes`, however, Stan Hu from Gitlab found that it does not support it for bare repositories.

You can read more about my research here: https://gitlab.com/gitlab-org/gitlab-ce/issues/2315#note_4045477

I'm new to all these projects codebases, so I'm unsure how big a task it would be to support this behaviour. Could you provide any guidance on where to start looking or any other hints?

stanhu:
Thanks for raising this issue, @matto1990. It looks like if you copy `.giattributes` into the bare repo as `info/attributes` then things behave as they should. There was a post years ago about adding `.gitattributes` support for bare repositories in git:

http://git.661346.n2.nabble.com/RFC-gitattributes-on-a-bare-repo-td666815.html

ethomson:
> it seems that the library does correctly support marking files as binary through .gitattributes

I disagree; we correctly *ignore* `.gitattributes` in a bare repository.  This is the behavior of git.git.

Let's take a simple example, and add a `.gitattributes` to a repository.

```
zoidberg:Temp/testrepo% git diff HEAD~1..HEAD
diff --git a/foo.txt b/foo.txt
index 8c2d357..f82d87d 100644
--- a/foo.txt
+++ b/foo.txt
@@ -1,4 +1,4 @@
-foo
+foo!
 foo
 foo
 bar
zoidberg:Temp/testrepo% echo '*.txt binary' > .gitattributes
zoidberg:Temp/testrepo% git diff HEAD~1..HEAD               
diff --git a/foo.txt b/foo.txt
index 8c2d357..f82d87d 100644
Binary files a/foo.txt and b/foo.txt differ
```

That all seems correct.  Now if we check it in:

```
zoidberg:Temp/testrepo% git add .gitattributes
zoidberg:Temp/testrepo% git commit -maddgitattributes
[master 8653486] addgitattributes
 1 file changed, 1 insertion(+)
 create mode 100644 .gitattributes
```

And clone the repository to a bare one:

```
zoidberg:Temp% git clone --bare testrepo bare-test.git
Cloning into bare repository 'bare-test.git'...
done.
zoidberg:~/Temp% cd bare-test.git 
[...]

PR #4650, opened 2018-05-21
Allow merge_driver_name_for_path to continue when merge attribute is not found.

Zoxive:
I'm not familar with the C library of libgit2 I've only done work with libgit2sharp.

That said i think I found an issue that prevents me from using in-memory repositories and libgit2sharp's new MergeCommitsIntoIndex (Transient Index https://github.com/libgit2/libgit2sharp/pull/1534) feature.

I believe ignoring the GIT_ENOTFOUND error is ok because it will fallback to the attr is unspecified. (However this opens up the question as to how to handle setting these kind of options when you have no file system.. )


.gitignore
```diff
@@ -31,3 +31,4 @@ msvc/Release/
 .*.swp
 tags
 mkmf.log
+.vs
```
neithernut: Not a project maintainer, but this addition to the `.gitignore` should definitely not be part of the same commit as the changes in `src/merge_driver.c`.

ethomson:
Can you better describe the problem that you're seeing and provide some more information?

I don't yet agree with your analysis of the issue - the merge attribute will _typically_ not be found; if it's not, it's unset and should fall into the normal `if (GIT_ATTR_FALSE(value))` test.

You mention in-memory workflows.  Do you maybe not have a configuration at all, and _that's_ what's not found here?

Zoxive:
My problem Its pretty specific to libgit2sharp.. that then cross over to libgit2.

I do not have a configuration my whole repo is in-memory.

I'm using libgit2sharp with a Custom ObjectDatabase backend, along with a custom ref backend. I want to see merge conflicts between two branches. (Previously I would temporarly copy the whole packfile backup of my objects to a file system and call .Merge but this takes 30+ seconds depending on the size) But now theres MergeCommitsIntoIndex with a Transient Index that is way faster and accomplishes my use case with less moving parts. However because I have no real filesystem libgit2 throws at https://github.com/libgit2/libgit2/blob/17fcbdcef0c88006669df925bdad9e567fed4497/src/repository.c#L2283 when trying to do the merge.  I traced it backwards to find it was just trying to look up the merge driver.

Its very likely im doing something wrong, and im also pushing libgit2sharp it its limits trying to do everything in-memory..

Zoxive:
According to https://github.com/libgit2/libgit2/commit/2d486781df0828eae716937b23d2df0a9c1817f9 My configuration doenst have to be on the file system? Is this where i can set all my attribute options to exist for the merge to work?

ethomson:
I see, thanks for that additional debugging information.  It's not trying to read the configuration, it's trying to open the `.gitattributes` file and failing because you don't have a working directory to read it from.

> According to 2d48678 My configuration doenst have to be on the file system? Is this where i can set all my attribute options to exist for the merge to work?

Your configuration does not need to be on the filesystem; configuration is pluggable just like object database and reference database.

But, this is definitely a bug of some sort.  I'm not actually sure the best way to sort it offhand.  I'm not quite sure that this is the right place to ignore this, since it could represent a real problem if we were suddenly unable to find the working directory.

Instead, I think we want to check if we're in a bare repository or not.  Something like:

    if (git_repository_is_bare(...)) {
 	*out = merge_driver_name__text;
        return 0;
    }

Zoxive:
So if bare_repositories dont support git_attributes at all, would it make sense to move it into the git_attr_get function?

But looking into git_attr_get, it was just changed recently revolving around bare repositories (https://github.com/libgit2/libgit2/commit/e28e17e60e7a73640a493918d610b9fe56e7f727). So I guess i'm unsure if .gitattributes are supported in bare repos? According to this older issue(https://github.com/libgit2/libgit2/issues/3676) they are not.

I'm just guessing going down this path of trying to get libgit2sharp to work better with in-memory type objects I'm going to hit a lot of edge cases like this.

ethomson:
> So if bare_repositories dont support git_attributes at all, would it make sense to move it into the git_attr_get function?

Perhaps; I hadn't done a particularly deep analysis of the problem, or an actual solution, I was just trying to provide more information.

> So I guess i'm unsure if .gitattributes are supported in bare repos? According to this older issue(#3676) they are not.

`.gitattributes` files may be read out of the _index_ as well as the filesystem.  A repository that lacks a working directory _may_ have an index.

> I'm just guessing going down this path of trying to get libgit2sharp to work better with in-memory type objects I'm going to hit a lot of edge cases like this.

I don't anticipate that; Visual Studio Team Services uses LibGit2Sharp with in-memory repositories and custom backends.

Zoxive:
Thanks for explaining. I'll keep playing around a bit more. I do like the if (git_repository_is_bare(...)) option a bit more than my original.

ethomson:
Tidying up some out of date PRs. Thanks for the contribution. 🙏

PR #4655, opened 2018-05-24
index: Fix alignment issues in write_disk_entry()

glaubitz:
In order to avoid alignment issues on certain target architectures, it is necessary to use memcpy() when modifying elements of a struct inside a buffer returned by git_filebuf_reserve().

This fixes the testsuite on Debian mipsel and sparc64 where it crashes with "Bus error".
See: https://buildd.debian.org/status/package.php?p=libgit2&suite=sid

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -2629,7 +2629,7 @@ static bool is_index_extended(git_index *index)
 static int write_disk_entry(git_filebuf *file, git_index_entry *entry, const char *last)
 {
 	void *mem = NULL;
-	struct entry_short *ondisk;
+	struct entry_short ondisk;
 	size_t path_len, disk_size;
 	int varint_len = 0;
 	char *path;
@@ -2657,9 +2657,7 @@ static int write_disk_entry(git_filebuf *file, git_index_entry *entry, const cha
 	if (git_filebuf_reserve(file, &mem, disk_size) < 0)
 		return -1;
 
-	ondisk = (struct entry_short *)mem;
-
-	memset(ondisk, 0x0, disk_size);
+	memset(mem, 0x0, disk_size);
 
 	/**
 	 * Yes, we have to truncate.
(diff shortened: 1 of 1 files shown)


src/index.c
```diff
@@ -2633,9 +2633,8 @@ static int write_disk_entry(git_filebuf *file, git_index_entry *entry, const cha
 	if (git_filebuf_reserve(file, &mem, disk_size) < 0)
 		return -1;
 
-	ondisk = (struct entry_short *)mem;
-
-	memset(ondisk, 0x0, disk_size);
+	memset(mem, 0x0, disk_size);
+	memcpy(&ondisk, mem, sizeof(struct entry_short));
```
pks-t: So you first zero out `mem` and then afterwards copy the zeroed out memory into `ondisk`. So simply using `memset(&ondisk, 0, sizeof(ondisk))` would result in more efficient code here, wouldn't it?

src/index.c
```diff
@@ -2647,30 +2646,33 @@ static int write_disk_entry(git_filebuf *file, git_index_entry *entry, const cha
...
 
-	ondisk->flags = htons(entry->flags);
+	memcpy(mem, &ondisk, sizeof(struct entry_short));
 
 	if (entry->flags & GIT_IDXENTRY_EXTENDED) {
-		struct entry_long *ondisk_ext;
-		ondisk_ext = (struct entry_long *)ondisk;
-		ondisk_ext->flags_extended = htons(entry->flags_extended &
+		struct entry_long ondisk_ext;
+		memcpy(&ondisk_ext, mem, sizeof(struct entry_long));
```
pks-t: Huh. So this doesn't make a lot of sense to me. `mem` was mmapped previously, which is why we assigned that and then just continued writing via the `ondisk_ext` pointer. But what you're doing now is first write `ondisk` to disk via the mmapped file buffer, then copy `ondisk` into `ondisk_ext` and overwrite what you've previously written to the mmapped buffer with `ondisk_ext`. It would make sense to avoid that overhead and only write either `ondisk` or `ondisk_ext` to disk, but not both.

So essentially, I guess what I'm asking for is to move the above `memcpy` into these conditionals here:

1. prepare `struct entry_short`
2. is it extended? copy `entry_short` into `entry_long` and `memcpy` that to `mem`. Otherwise, just `memcpy` `entry_short` into `mem`

Does that make sense to you?

src/index.c
glaubitz: Yeah, you're right. I was more focusing on the memory-mapped region pointed to by ```mem``` and wanted to make sure it's all cleared out. It was a bit late yesterday when I wrote this ;).

glaubitz:
Thanks for the quick review. I'll push an updated version later today.

src/index.c
glaubitz: On second thought: ```ondisk``` and ```mem``` can have different sizes here: ```sizeof(struct entry_short)``` vs ```disk_size```. So, if I made the change as suggested, the memory region that would get zeroed would be different.

glaubitz:
Just pushed an updated version which should address your issues.

glaubitz:
@pks-t Have you had a chance yet to review my updated pull request?

src/index.c
```diff
@@ -2647,30 +2645,32 @@ static int write_disk_entry(git_filebuf *file, git_index_entry *entry, const cha
...
 	if (entry->flags & GIT_IDXENTRY_EXTENDED) {
-		struct entry_long *ondisk_ext;
-		ondisk_ext = (struct entry_long *)ondisk;
-		ondisk_ext->flags_extended = htons(entry->flags_extended &
+		struct entry_long ondisk_ext;
+		memcpy(&ondisk_ext, &ondisk, sizeof(struct entry_short));
+		ondisk_ext.flags_extended = htons(entry->flags_extended &
 			GIT_IDXENTRY_EXTENDED_FLAGS);
-		path = ondisk_ext->path;
+		memcpy(mem, &ondisk_ext, sizeof(struct entry_long));
```
pks-t: This is in fact writing `ondisk_ext.path` into `mem`, which is not initialized yet. While I think we're fine here as `path` gets `memcpy`d to anyway, you could also use `offsetof(struct entry_long, path)` instead.

src/index.c
```diff
@@ -2647,30 +2645,32 @@ static int write_disk_entry(git_filebuf *file, git_index_entry *entry, const cha
...
+		path = ((struct entry_long*)mem)->path;
 		disk_size -= offsetof(struct entry_long, path);
 	} else {
-		path = ondisk->path;
+		memcpy(mem, &ondisk, sizeof(struct entry_short));
```
pks-t: Same here, but obviously with `offsetof(struct entry_short, path)`

pks-t:
Sorry, I didn't find the time yet. There's one small nit with regards to uninitialized `path` being copied to the mmap, but otherwise I'm happy

src/index.c
glaubitz: I'm not 100% sure what you mean. Use the offset of ```path``` for what? Why shouldn't ```mem``` not be initialized? It's zero'd in line 2636, isn't it?

pks-t: `mem` is initialized, yes. But `ondisk_ext.path` is not initialized, as you only copy `sizeof(struct entry_short)` bytes into `ondisk_ext`, leaving additional struct entries (which happen to be only `path` and the extended flags) uninitialized.

pks-t: And sorry for being unclear: use `offsetof(struct entry_long, path)` as the third argument of `memcpy`

glaubitz: Ah,  now I understand what you mean. Thanks for the additional details. Yes, you're right, ```path``` not initialized yet. I will use your suggested change.

glaubitz: Ah, and shouldn't it be ```offsetof(struct entry_short, path)``` for both cases? We're only filling up the extended flags after the ```memcpy()``` aren't we?

glaubitz: Wait, no. I looked at the wrong line. Never mind ;).

glaubitz:
Pushed updated and rebased version.

glaubitz:
Ok, let me know in case I need to rebase.

pks-t:
Thanks again for your fix!

Issue #4651, opened 2018-05-21
libgit2.pc still requires openssl even after building with mbedtls


pravi:
### Reproduction steps

Build libgit2 with mbedtls using the patch from #4173 
Used this patch at https://anonscm.debian.org/git/collab-maint/libgit2.git/tree/debian/patches/use-mbedtls.patch

and build it in a clean chroot without openssl.

### Expected behavior
It should not add openssl as a dependency in .pc.

### Actual behavior

The generated ibgit2.pc has openssl

```
prefix="/usr"
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: libgit2
Description: The git library, take 2
Version: 0.27.0

Libs: -L${libdir} -lgit2
Libs.private: -lrt -lpthread -lhttp_parser -lssh2
Requires.private: openssl zlib

Cflags: -I${includedir}
```
and makes build of other software using libgit2 to fail. See 
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=899036

### Version of libgit2 (release number or SHA1)

0.27.0

### Operating system(s) tested

Debian unstable/experimental.

pravi:
```
-- Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY OPENSSL_INCLUDE_DIR) 
-- Found mbedTLS:
--   version 2.8.0
--   TLS: /usr/lib/s390x-linux-gnu/libmbedtls.so
--   X509: /usr/lib/s390x-linux-gnu/libmbedx509.so
--   Crypto: /usr/lib/s390x-linux-gnu/libmbedcrypto.so
Auto-detecting default certificates location
```
Full build log here https://buildd.debian.org/status/fetch.php?pkg=libgit2&arch=s390x&ver=0.27.0%2Bdfsg.1-0.1&stamp=1526122841&raw=0

tiennou:
Looking at the build log, I have the following questions:

- you're not using the correct CMake variable for backend selection (IOW `-DUSE_OPENSSL=NO` does nothing, and you're relying on the backend autodetection (hence the lines you're getting). Use `-DUSE_HTTPS=mbedTLS` instead so the autodetection doesn't kick in.
[...]

PR #4656, opened 2018-05-25
mbedtls: don't require mbedtls from our pkgconfig file

tiennou:
mbedTLS has no pkgconfig file, hence we can't require it. For now, pass its link flags as our own.

Ref: #4651


src/CMakeLists.txt
```diff
@@ -232,8 +232,11 @@ IF (USE_HTTPS)
 		SET(GIT_MBEDTLS 1)
 		LIST(APPEND LIBGIT2_INCLUDES ${MBEDTLS_INCLUDE_DIR})
 		LIST(APPEND LIBGIT2_LIBS ${MBEDTLS_LIBRARIES})
-		LIST(APPEND LIBGIT2_PC_LIBS ${MBEDTLS_LDFLAGS})
-		LIST(APPEND LIBGIT2_PC_REQUIRES "mbedtls")
+		# mbedTLS has no pkgconfig file, hence we can't require it
+		# https://github.com/ARMmbed/mbedtls/issues/228
+		# For now, pass its link flags as our own
+		LIST(APPEND LIBGIT2_PC_LIBS ${MBEDTLS_LIBRARIES})
```
pks-t: Just to be sure here: does `MBEDTLS_LIBRARIES` include `MBEDTLS_LDFLAGS`?

src/CMakeLists.txt
```diff
@@ -232,8 +232,11 @@ IF (USE_HTTPS)
...
+		# LIST(APPEND LIBGIT2_PC_REQUIRES "mbedtls")
```
pks-t: I'd say to just remove this line. Having commented-out sections can be confusing, and you're able to easily dig up this code by just git-blaming the above comment

src/CMakeLists.txt
tiennou: It's a little bit weirder : the FindmbedTLS module doesn't actually set `MBEDTLS_LDFLAGS` [at all](https://github.com/libgit2/libgit2/blob/master/cmake/Modules/FindmbedTLS.cmake#L11), so I hope this is sufficient. AFAIU, `Libs.private` only impacts static linking (ie. pkg-config libgit2 --static --libs`), and the linker should already know what to do when dyn-linking.

The resulting pc file looks like this :
```
prefix="/usr/local"
libdir=${prefix}/lib
includedir=${prefix}/include

Name: libgit2
Description: The git library, take 2
Version: 0.27.0

Libs: -L${libdir} -lgit2
Libs.private: -lcurl -L/usr/local/lib -lmbedtls -lmbedx509 -lmbedcrypto -lz -L/usr/local/Cellar/libssh2/1.8.0/lib -lssh2 -L/usr/lib -liconv
Requires.private: 

Cflags: -I${includedir}
```

src/CMakeLists.txt
tiennou: Yep, removed.

PR #4661, opened 2018-05-30
streams: openssl: add missing check on OPENSSL_LEGACY_API

laomaiweng:
The `CRYPTO_THREADID` type is no longer available in OpenSSL ≥ 1.1.0 with deprecated features disabled, and causes build failures. Since the `threadid_cb()` function is only ever called by `git_openssl_set_locking()` when `defined(OPENSSL_LEGACY_API)`, only define it then.

### DIFF
--- a/src/streams/openssl.c
+++ b/src/streams/openssl.c
@@ -249,7 +249,7 @@ int git_openssl_stream_global_init(void)
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


pks-t:
Thanks for your contribution, this is obviously correct!

PR #4662, opened 2018-05-30
path: unify `git_path_is_*` APIs

pks-t:
Right now, there's quite a lot of different function calls to determine
whether a path component matches a specific name after normalization
from the filesystem. We have a function for each of {gitattributes,
gitmodules, gitignore} multiplicated with {generic, NTFS, HFS} checks.
In the long time, this is unmaintainable in case there are e.g. new
filesystems with specific semantics, blowing up the number of functions
we need to implement.

Replace all functions with a simple `git_path_is_gitfile` function,
which accepts an enum pointing out the filename that is to be checked
against as well as the filesystem normalizations to check for. This
greatly simplifies implementation at the expense of the caller having to
invoke a somewhat longer function call.

---

I'm rather indifferent with regards to function and enum names, I'd definitely be open for proposals here. But I think it's undisputed that the explosion of different path verification functions is not something we want, especially if we're going to expose these later on.

/cc @carlosmn

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -1766,14 +1766,14 @@ static bool verify_component(
 	if (flags & GIT_PATH_REJECT_DOT_GIT_HFS) {
 		if (!verify_dotgit_hfs(component, len))
 			return false;
-		if (S_ISLNK(mode) && git_path_is_hfs_dotgit_modules(component, len))
+		if (S_ISLNK(mode) && git_path_is_gitfile(component, len, GIT_PATH_GITFILE_GITMODULES, GIT_PATH_FS_HFS))
 			return false;
 	}
 
 	if (flags & GIT_PATH_REJECT_DOT_GIT_NTFS) {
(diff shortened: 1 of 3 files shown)


ethomson:
Yeah, this looks like a nice improvement to me.

I might keep the `len` argument just to avoid a `strlen`, since we're often calling this in a loop...  but at the same time I hope every compiler would recognize that as a constant strlen and optimize it out.  🤷‍♂️

pks-t:
I had the same train of thought with regards to the `strlen`. I mean I could easily put it into the struct, as well, but I don't know if it's worthwhile. On the other hand it is not complex to do, either

pks-t:
MSVC 2010 doesn't support explicit indices for array initializers :( To be honest, I didn't even know whether that is allowed in C90

ethomson:
:+1

PR #4665, opened 2018-06-01
refdb_fs: fix regression: failure when globbing for non-existant references

### LINKED ISSUE #4664: Regression: Refdb glob
### Reproduction steps
Call ```git_revwalk_hide_glob``` with a ref that doesn't exist, something like ```refs/heads/notmaster```

### Expected behavior
No error.

### Actual behavior
Errors out with an error similar to the following:

`failed to open directory '/Users/jacquesg/github/p5-Git-Raw/.build/FmeM1xl2sg/t/test_repo/.git//refs/heads/notmaster/': No such file or directory`

Introduced by 20a2b02d9a1bcb4825ec49605146223c565dcacf

Linux

neithernut:
Fixes #4664.

### DIFF
--- a/src/refdb_fs.c
+++ b/src/refdb_fs.c
@@ -538,10 +538,14 @@ static int iter_load_loose_paths(refdb_fs_backend *backend, refdb_fs_iter *iter)
 	}
 
 	if ((error = git_buf_printf(&path, "%s/", backend->commonpath)) < 0 ||
-		(error = git_buf_put(&path, ref_prefix, ref_prefix_len)) < 0 ||
-		(error = git_iterator_for_filesystem(&fsit, path.ptr, &fsit_opts)) < 0) {
+		(error = git_buf_put(&path, ref_prefix, ref_prefix_len)) < 0) {
 		git_buf_free(&path);
 		return error;
+	}
+
+	if ((error = git_iterator_for_filesystem(&fsit, path.ptr, &fsit_opts)) < 0) {
+		git_buf_free(&path);
+		return (iter->glob && error == GIT_ENOTFOUND)? 0 : error;
 	}
 
 	error = git_buf_sets(&path, ref_prefix);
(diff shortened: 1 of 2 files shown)


neithernut:
@jacquesg could you confirm whether this fixes the issue you reported?

jacquesg:
@neithernut Confirmed, I no longer see the previous failures. Thanks!

src/refdb_fs.c
```diff
@@ -538,12 +538,16 @@ static int iter_load_loose_paths(refdb_fs_backend *backend, refdb_fs_iter *iter)
...
-		(error = git_buf_put(&path, ref_prefix, ref_prefix_len)) < 0 ||
-		(error = git_iterator_for_filesystem(&fsit, path.ptr, &fsit_opts)) < 0) {
+		(error = git_buf_put(&path, ref_prefix, ref_prefix_len)) < 0) {
 		git_buf_free(&path);
 		return error;
 	}
 
+	if ((error = git_iterator_for_filesystem(&fsit, path.ptr, &fsit_opts)) < 0) {
+		git_buf_free(&path);
+		return (iter->glob && error == GIT_ENOTFOUND)? 0 : error;
```
pks-t: Is it necessary to check whether we have a globbing iterator or not? In the case where the iterator is not using globs, we're always using "refs/" as our entrypoint, and absence of that would certainly be unexpected. What is our current failure mode in that case? Do we error out or do we just report 0 references?

pks-t:
The fixup makes sense to me, thanks. Just waiting for some clarification on my comment and then I'd be good to merge

src/refdb_fs.c
neithernut: The error gets propagated to the caller. I felt that it would be wrong to hide the error in this case. And the additional check hints at why `GIT_ENOTFOUND` can be ignored.

neithernut:
Squashed in order to save a round-trip.

pks-t:
Waiting a few days whether there's feedback by others

pks-t:
Thanks again for your fix!

Issue #4658, opened 2018-05-28
git_remote_create without insteadof rewrite rules


vors:
> You are opening a _bug report_ against the libgit2 project.  If you have a
> question about an API or usage, please ask on StackOverflow:
> http://stackoverflow.com/questions/tagged/libgit2.  Please fill out the
> reproduction steps (below) and delete this introductory paragraph.  Thanks!

### Prelude
Apoligies in advance for openning an issue that could be viewed more like an API question and not a bug report. I [asked on stackoverflow](https://stackoverflow.com/questions/50548680/libgit2-git-remote-create-without-apply-insteadof) as suggested, but didn't get any replies.
It's unclear to me that this is even possible to achive with the current public APIs at all. Hence, I think it's appropriate to open an issue in this repo.

### Question

This is somewhat similar to #2923 .

I'd like to use an equivalent of [`git_remote_create_anonymous`](https://github.com/libgit2/libgit2/blob/0eca42304a10c9ad6170a38a440dfab8e354d38d/include/git2/remote.h#L73) function, but without applying `insteadof` url substitutions. This is similar to [`git_remote_create_detached`](https://github.com/libgit2/libgit2/blob/0eca42304a10c9ad6170a38a440dfab8e354d38d/include/git2/remote.h#L92) which explicitly says that it doesn't apply `insteadof` substitues.

I could not find an equivalent API or combination of APIs that would get me a fetchable repo without the global config substitutes applied.

Am I missing something? If not, would it be reasonable to add such API?

### Motivation
The use-case of it is to bypass any users global `~/.gitconfig` settings.
For example, it's common to have

```
[url "ssh://git@github.com/"]
    insteadOf = https://github.com/
```

[CircleCI uses](https://github.com/rust-lang/cargo/issues/2429#issuecomment-191370340) this approach.

That causes somewhat unpredictable behavior for tools that use github repos for their online indexes, like [cargo](https://github.com/rust-lang/cargo). Users find workaround for it, but it would be really nice to put more control in the hands of tools authors, so they can bypass the rewrite rules. libgit2 already does it for `git ls-remote` functionality with detached repos.

tiennou:
I'm not really against adding API for that, I'm just wondering about this (after having written the code 🤷‍♂️) :

[...]

PR #4667, opened 2018-06-01
Remote creation API

tiennou:
This is the PR for the code I've written for #4658.

It adds a `git_remote_create_ext` with an option structure for handling the 3 other `_create` calls (ie. `_with_fetchspec`, `_anonymous`, `_detached`).

I've not pondered too much about the difference between anonymous & detached remotes. AFAICT, both are in-memory, and anonymous ones can make use of an underlying repository, so config will be respected.


vors:
😍 thank you for taking a stub at it!

src/remote.c
```diff
@@ -189,6 +189,28 @@ static int canonicalize_url(git_buf *out, const char *in)
...
+		return 0;
+
+	if (error < 0)
+		return error;
+
+	git_remote_free(remote);
+
+	giterr_set(
+			   GITERR_CONFIG,
+			   "remote '%s' already exists", name);
```
pks-t: Indentation here is a bit funny.

include/git2/remote.h
```diff
@@ -41,6 +41,52 @@ GIT_EXTERN(int) git_remote_create(
...
+	const char *name;
+
+	/** The fetchspec the remote should use. */
+	const char *fetchspec;
+} git_remote_create_options;
+
+#define GIT_REMOTE_CREATE_OPTIONS_VERSION 1
+#define GIT_REMOTE_CREATE_OPTIONS_INIT {GIT_REMOTE_CREATE_OPTIONS_VERSION}
+
+GIT_EXTERN(int) git_remote_create_init_options(
```
pks-t: This function is missing its documentation.

include/git2/remote.h
```diff
@@ -41,6 +41,52 @@ GIT_EXTERN(int) git_remote_create(
...
+ * Remote creation options structure
+ *
+ * Initialize with `GIT_REMOTE_CREATE_OPTIONS_INIT`. Alternatively, you can
+ * use `git_remote_create_init_options`.
+ *
+ */
+typedef struct git_remote_create_options {
+	unsigned int version;
+
+	/** The repository that should own the remote.
```
pks-t: Our struct member doc comment style is to either have a single `/** foobar */` line or start with an empty `/**` line.

include/git2/remote.h
```diff
@@ -41,6 +41,52 @@ GIT_EXTERN(int) git_remote_create(
...
+ * Create a remote, with options.
+ *
+ * This function allows more fine-grained control over the remote creation.
+ *
+ * @param out the resulting remote
+ * @param url the remote's url
+ * @param opts the remote creation options
+ * @return 0 on success, WIP
+ */
+GIT_EXTERN(int) git_remote_create_ext(
```
pks-t: Bikeshedding: `create_ext` doesn't make it clear what the "extended" nature of this function is. Instead, I'd vote for `git_remote_create_with_opts`, which clearly states that it takes an `opts` structure.

include/git2/remote.h
```diff
@@ -41,6 +41,52 @@ GIT_EXTERN(int) git_remote_create(
...
+		unsigned int version);
+
+/**
+ * Create a remote, with options.
+ *
+ * This function allows more fine-grained control over the remote creation.
+ *
+ * @param out the resulting remote
+ * @param url the remote's url
+ * @param opts the remote creation options
```
pks-t: You should mention that `opts` is optional and what will happen in that case. (It is optional, isn't it? If not, I'd propose to just create an anonymous remote in that case)

src/remote.c
```diff
@@ -211,58 +211,79 @@ static int ensure_remote_doesnot_exist(git_repository *repo, const char *name)
...
 			goto on_error;
 
 		/* only write for non-anonymous remotes */
-		if (repo && name && (error = write_add_refspec(repo, name, fetch, true)) < 0)
-			goto on_error;
-
-		if (repo && (error = lookup_remote_prune_config(remote, config_ro, name)) < 0)
+		if (!is_anonymous &&
+			((error = write_add_refspec(opts->repository, opts->name, opts->fetchspec, true)) < 0 ||
+			(error = lookup_remote_prune_config(remote, config_ro, opts->name)) < 0))
```
pks-t: Wait, this is different from what we did before. Now we only call `lookup_remote_prune_config` if both `opts->repository` and `opts->name` are set. Previously, we were also calling it in case `opts->name` is unset. Is this intentional and, if so, why can we avoid calling `lookup_remote_prune_config` now?

src/remote.c
```diff
@@ -271,7 +292,7 @@ static int create_internal(git_remote **out, git_repository *repo, const char *n
 	}
 
 	/* A remote without a name doesn't download tags */
-	if (!name)
+	if (is_anonymous)
```
pks-t: This is also different from what we did before, as now an unnamed remote with an associated repository will not fetch tags anymore. Is this intended?

src/remote.c
```diff
@@ -306,37 +327,41 @@ int git_remote_create(git_remote **out, git_repository *repo, const char *name,
...
 
-	if (create_internal(&remote, repo, name, url, fetch) < 0)
-		goto on_error;
-
-	*out = remote;
-
-	return 0;
+	opts.repository = repo;
+	opts.name = name;
+	opts.fetchspec = fetch;
```
pks-t: This conversion is great

src/remote.c
```diff
@@ -306,37 +327,41 @@ int git_remote_create(git_remote **out, git_repository *repo, const char *name,
...
 {
-	git_remote *remote = NULL;
 	int error;
+	git_remote_create_options opts = GIT_REMOTE_CREATE_OPTIONS_INIT;
 
 	if ((error = ensure_remote_name_is_valid(name)) < 0)
 		return error;
 
 	if ((error = ensure_remote_doesnot_exist(repo, name)) < 0)
 		return error;
```
pks-t: These two calls are already done in `create_remote_internal` now. I'd like to avoid this duplication

include/git2/remote.h
```diff
@@ -63,6 +73,9 @@ typedef struct git_remote_create_options {
 
 	/** The fetchspec the remote should use. */
 	const char *fetchspec;
+
+	/** Additional flags for the remote. See git_remote_create_flags. */
+	int flags;
```
pks-t: This should be unsigned, shouldn't it?

src/remote.c
tiennou: Yeah, the tests around detached/anonymous/realized remotes are confusing here. My `is_anonymous` variable should be named `is_not_realized` maybe (bikeshedding welcome on the name).

The original code would attempt to load a `remote..prune` config option, which IMHO doesn't make sense (and I remember a test broke before I changed it to this).

src/remote.c
tiennou: You're right, I went too far trying to make sense of the 3 types of remotes. Reverting.

src/remote.c
```diff
@@ -211,58 +211,79 @@ static int ensure_remote_doesnot_exist(git_repository *repo, const char *name)
...
+		(error = git_repository_config_snapshot(&config_ro, opts->repository)) < 0)
 		return error;
 
 	remote = git__calloc(1, sizeof(git_remote));
 	GITERR_CHECK_ALLOC(remote);
 
-	remote->repo = repo;
+	remote->repo = opts->repository;
 
 	if ((error = git_vector_init(&remote->refs, 32, NULL)) < 0 ||
```
tiennou: Just a heads up, I was pretty surprised about that 32 here. Should I lower that to something like 4 ?

src/remote.c
tiennou: Confirmed. Should I keep the test as is ?
```
  1) Failure:
network::remote::remotes::cannot_add_a_nameless_remote [/Users/tiennou/Projects/libgit2/tests/network/remote/remotes.c:321]
  GIT_EINVALIDSPEC != git_remote_create(&remote, _repo, NULL, "git://github.com/libgit2/libgit2")
  -12 != 0
```

tiennou:
Rebased, with all review comments taken care of apart from the change of behavior [here](https://github.com/libgit2/libgit2/pull/4667/files#diff-2f07edcf8df8f205905e1e7d24631753R284).

src/remote.c
pks-t: You're right, the original code doesn't make a lot of sense. So this change looks correct to me

tests/network/remote/remotes.c
```diff
@@ -318,7 +318,7 @@ void test_network_remote_remotes__cannot_add_a_nameless_remote(void)
 
 	cl_assert_equal_i(
 		GIT_EINVALIDSPEC,
-		git_remote_create(&remote, _repo, NULL, "git://github.com/libgit2/libgit2"));
+		git_remote_create(&remote, _repo, "", "git://github.com/libgit2/libgit2"));
```
pks-t: This is the test you've been talking about, right? I'd say we should just keep this test as it was originally, just to be sure we're not altering behaviour.

include/git2/remote.h
```diff
@@ -41,6 +41,16 @@ GIT_EXTERN(int) git_remote_create(
 		const char *name,
 		const char *url);
 
+/**
+ * Remote creation options flags
+ */
+typedef enum {
+	GIT_REMOTE_CREATE_DEFAULT = 0,
```
pks-t: Why is this flag necessary? We don't have default flags for all the other option enums, do we?

src/remote.c
tiennou: Note that most of what I said above about the broken test is wrong. The test failure was caused when I removed the name & repo existence checks from `git_remote_create` (since it allows `NULL`).

AFAICT, the original code is extra-confusing about what actually needs a repo/name because it indirectly uses the config, so I hope this is better. I'm still dubious of the whole config_ro/config_rw dance that's happening, because we start by doing a snapshot, then we re-grab a "live" config. I might be just paranoid/unaware of the semantics of `config__weakptr`, but I'm tempted to change it to : grab a "live" version IF we have a repo, snapshot it so it doesn't change when applying `url.insteadof`, and set the various other settings on the first live one.

tiennou:
So, major rework of this. I've taken the time to strengthen the test suite because I kept discovering "edge-cases" (like with_fetchspec not building a default fetchspec), so all 3 basic calls are now more cautiously tested (for invalid parameters, config changes, and whatever differing behavior I stumbled upon).

New from last time :
- I've added a flag to prevent building a default fetchspec. I've also re-duplicated code either for backward-compatibility reasons (`git_remote_create` actually shadows its error codes) or simplicity in `create_internal`.
- I've switched the config-management code to use transactions (so the setup/teardown cycle is simpler).

Left to do/open questions :
- check that we set error messages correctly. I noticed a few `<no message>` failures.
- maybe have the `const char *fetchspec` option be a `size_t count, const char fetchspecs[]` ? I didn't notice at first, but using `git_remote_add_fetch` will need a relookup so the refspecs are "in the correct place" (see `dwim_refspecs` at the end of `create_internal`).

tests/remote/create.c
```diff
@@ -27,11 +37,68 @@ void test_remote_create__manual(void)
...
+	cl_git_pass(git_config_get_string(&cfg_val, cfg, "remote.valid-name.url"));
+	cl_assert_equal_s(cfg_val, TEST_URL);
+
+	git_config_free(cfg);
+	git_remote_free(remote);
+}
+
+void test_remote_create__named_fail_on_invalid_name(void)
+{
+	cl_git_assert_cannot_create_remote(GIT_EINVALIDSPEC, git_remote_create(&r, _repo, NULL, TEST_URL));
```
pks-t: I'd actually prefer the same as in the other PR. Instead of using a macro and repeating that multiple times, you should just use an array of strings here and iterate through them. That'd be a lot simpler to read and produce less code.

src/remote.c
```diff
@@ -232,7 +232,7 @@ static int create_internal(git_remote **out, git_repository *repo, const char *n
 
 	remote->repo = repo;
 
-	if ((error = git_vector_init(&remote->refs, 32, NULL)) < 0 ||
+	if ((error = git_vector_init(&remote->refs, 8, NULL)) < 0 ||
```
pks-t: I'd really like to see some reasoning in the commit message as to why this is a good thing.

src/remote.c
```diff
@@ -278,19 +281,19 @@ static int create_internal(git_remote **out, git_repository *repo, const char *n
...
+		git_config_unlock(config, 1);
 	*out = remote;
 	error = 0;
 
 on_error:
 	if (error)
 		git_remote_free(remote);
 
-	git_config_free(config_ro);
+	git_config_free(config);
```
pks-t: Don't we have to unlock the config on error, too?

src/remote.c
tiennou: Memory bloat ? That's enough memory for 32 different refspecs. I expect this is far larger than any "normal" usage of a remote. 8 is selected because it's the minimum allocation size of `git_vector`.

src/remote.c
tiennou: Will fix, I was misusing the transaction config API (hence the CI failure). The transaction handles unlocking when it's freed.

tests/remote/create.c
tiennou: Fixed.

include/git2/remote.h
tiennou: Fixed.

src/remote.c
pks-t: The minimum allocation size is the interesting information I've been looking for :) I wasn't really wondering why the count was removed, but rather why 8 was picked. Thanks for clarifying

src/remote.c
```diff
@@ -299,47 +321,60 @@ int git_remote_create(git_remote **out, git_repository *repo, const char *name,
...
+	git_remote_create_options opts = GIT_REMOTE_CREATE_OPTIONS_INIT;
 
-	if (default_fetchspec_for_name(&buf, name) < 0)
-		return -1;
+	/* Those 2 tests are duplicated here because of backward-compatibility */
+	if ((error = ensure_remote_name_is_valid(name)) < 0)
+		return error;
+
+	if (canonicalize_url(&buf, url) < 0)
+		return GIT_ERROR;
```
pks-t: What kind of backwards compatibility do you refer to? Do callers expect a specific error code here?

src/remote.c
```diff
@@ -263,7 +263,7 @@ static int create_internal(git_remote **out, const char *url, const git_remote_c
 		(error = canonicalize_url(&canonical_url, url)) < 0)
 		goto on_error;
 
-	if (config) {
+	if (config && (opts->flags & GIT_REMOTE_CREATE_SKIP_INSTEADOF) != GIT_REMOTE_CREATE_SKIP_INSTEADOF) {
```
pks-t: `if (config && !(opts->flags & GIT_REMOTE_CREATE_SKIP_INSTEADOF))`

src/remote.c
```diff
@@ -280,13 +281,24 @@ static int create_internal(git_remote **out, const char *url, const git_remote_c
 			goto on_error;
 	}
 
-	if (opts->fetchspec != NULL) {
-		if ((error = add_refspec(remote, opts->fetchspec, true)) < 0)
+	if (opts->fetchspec != NULL ||
+	    (opts->name && (opts->flags & GIT_REMOTE_CREATE_SKIP_DEFAULT_FETCHSPEC) != GIT_REMOTE_CREATE_SKIP_DEFAULT_FETCHSPEC)) {
```
pks-t: `(opts->name && !(opts->flags & GIT_REMOTE_CREATE_SKIP_DEFAULT_FETCHSPEC))`

src/remote.c
tiennou: I found out while writing [this test](https://github.com/libgit2/libgit2/pull/4667/commits/eb725118f18a92cfcf05de58f4ff90139af9c82f#diff-0f6c0f30ddbd94614e1f4f4887955556R93) that [this](https://github.com/libgit2/libgit2/blob/master/src/remote.c#L322) was "misbehaving", in that it never returns what `create_internal` says when it fails. It depends on what kind of backward compatibility we want (I realize I may be too zealous, but maybe this should be discussed separately)…

The `ensure_remote_name_is_valid` is duplicated in case name is NULL (a NULL name in the `with_opts` version would cause an anonymous remote to be created instead, ie. it's not a failure anymore). An alternative would be to assert on the mandatory parameters to be correct…

tiennou:
Should be good now. I've lowered the vector size again, because (if my sizeof math is correct), we have 32 * sizeof(void *) => 256bytes on 64bits, which makes 1Mb lost to mostly empty vectors for every 4 remote you create.

Thanks @pks-t for the reviews and ever helpful comments.

src/remote.c
```diff
@@ -278,19 +280,19 @@ static int create_internal(git_remote **out, git_repository *repo, const char *n
 	else
 		remote->download_tags = GIT_REMOTE_DOWNLOAD_TAGS_AUTO;
 
-
-	git_buf_dispose(&var);
-
+	if (txn)
+		git_transaction_commit(txn);
```
pks-t: Sorry to not have catched this earlier, but you need to do error checking here. Otherwise we might end up silently not writing the config changes.

src/remote.c
pks-t: Yeah. I guess we should fix this, but that can be done in a later PR if you wish.

tiennou:
Just a thought : per https://stackoverflow.com/a/51570151/444323, if I kept the new function's prototype compatible with clone's `git_remote_create_cb`, that would make a nice one-shot customization function, since `git_remote_create_opts` could be passed as `payload`.

tiennou:
@pks-t If you don't mind (and your workqueue is not saturated), I'd like to have this one reviewed 😉.

src/remote.c
```diff
@@ -189,58 +185,121 @@ static int canonicalize_url(git_buf *out, const char *in)
...
+	    (opts->name && !(opts->flags & GIT_REMOTE_CREATE_SKIP_DEFAULT_FETCHSPEC))) {
+		const char *fetch = NULL;
+		if (opts->fetchspec) {
+			fetch = opts->fetchspec;
+		} else {
+			if ((error = default_fetchspec_for_name(&specbuf, opts->name)) < 0)
+				goto on_error;
+
+			fetch = git_buf_cstr(&specbuf);
+		}
```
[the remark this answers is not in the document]
tiennou: I don't expect it does : the `on_error` "handler" unconditionally frees the buffer, so it doesn't matter if it's a self-built fetchspec or not. Arguably, I could rename that to `cleanup`…

tiennou:
Okay, something reaally weird is going on after rebasing to `master`… Just to be sure, I'll submit the rebased version as another "dummy" PR, and I'm restoring this one to the pre-rebase version (though the review comments have been taken care of) because I'd like to make sure it's still green (IIRC it was, but 🤷‍♂️ ). Please don't merge this while investigations are in progress…

tiennou:
Heh, since VSTS does a merge to master beforehand (obviously), it doesn't make sense to dummy-PR anymore (the branch is still up [here](https://github.com/libgit2/libgit2/compare/master...tiennou:feature/remote-create-api-master). I've managed to confused the build view so here's a [link](https://dev.azure.com/libgit2/libgit2/_build/results?buildId=741&view=logs) to its last results.

My initial reaction was to look at the leaks (since @pks-t was actually worried about those fetchspecs leaking and was thus quite surprised), but it finally dawned on me that the issue was a subsequent assertion failing causing the remote to be leaked. Looking at the config file on disk, the remote section is actually written correctly, but *somehow* the repo's view of it is not updated when the transaction completes, which causes the section count tests (in `with_fetchspec` and `with_empty_fetchspec`) to fail.

What's weird is that 1) this issue seems to be new on master, as the "old" version completes with 0 failures (at least locally on macOS) 2) this one's macOS build is green (Windows too), while my local build is not, with the same failing tests.

src/remote.c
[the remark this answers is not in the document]
pks-t: Oh, yeah. That was my mistake, I somehow expected a `git_buf_detach` here. `git_buf_cstr` doesn't transfer memory ownership, so this is fine

pks-t:
So yeah, the memory leak obviously isn't one, that was a thinko on my side. The only other two comments were only nice to have, so I'm approving. So as soon as CI's green and the conflict fixed I'd be happy to merge

tiennou:
Rebased, and ready for review. I tore apart the transaction code I had introduced because somehow it's not a good idea to use those on our `config__weakptr` version. If you feel that's an interesting bug, I can whip up a test case for it.

src/remote.c
```diff
@@ -189,58 +189,120 @@ static int canonicalize_url(git_buf *out, const char *in)
...
+{
+	GIT_INIT_STRUCTURE_FROM_TEMPLATE(
+		opts, version, git_remote_create_options, GIT_REMOTE_CREATE_OPTIONS_INIT);
+	return 0;
+}
+
+int git_remote_create_with_opts(git_remote **out, const char *url, const git_remote_create_options *opts)
+{
+	git_remote *remote = NULL;
+	git_transaction *txn = NULL;
```
pks-t: This variable is now unused, which causes the CI to fail

tiennou: Heh. Should be fixed now.

pks-t:
Thanks a lot @tiennou for this awesome work!

PR #4668, opened 2018-06-04
Fix stash save bug with fast path index check

### LINKED ISSUE #4101: Stash save does not behave the same as command line
### Reproduction steps
1. Create a stash using git_stash_save with default options.
2. use `git show stash@{0}` in command line
3. There are no diffs for the stash

### Expected behavior
The expected behavior is that the stashed diffs are visible in command line.

### Actual behavior
The diffs are not visible via command line.

37dba1a739b5ee6c45dc9f3c0bd1f7f7a18f13f7
Linux, Mac OSX

novalis:
If the index contains stat data for a modified file, and the file is
not racily dirty, and there exists an untracked working tree directory
alphabetically after that file, and there are no other changes to the
repo, then git_stash_save would fail. It would confuse the untracked
working tree directory for the modified file, because they have the
same sha: zero.  The wt directory has a sha of zero because it's a
directory, and the file would have a zero sha because we wouldn't read
the file -- we would just know that it doesn't match the index.  To
fix this confusion, we simply check mode as well as SHA.

Fixes #4101

### DIFF
--- a/src/diff_generate.c
+++ b/src/diff_generate.c
@@ -273,7 +273,8 @@ static git_diff_delta *diff_delta__last_for_item(
 		break;
 	case GIT_DELTA_MODIFIED:
 		if (git_oid__cmp(&delta->old_file.id, &item->id) == 0 ||
-			git_oid__cmp(&delta->new_file.id, &item->id) == 0)
+		    (delta->new_file.mode == item->mode &&
+			git_oid__cmp(&delta->new_file.id, &item->id) == 0))
 			return delta;
 		break;
 	default:
--- a/tests/stash/save.c
+++ b/tests/stash/save.c
@@ -188,6 +188,46 @@ void test_stash_save__can_include_untracked_and_ignored_files(void)
 	cl_assert(!git_path_exists("stash/just.ignore"));
 }
 
+/*
+ * Note: this test was flaky prior to fixing #4101 -- run it several
+ * times to get a failure.  The issues is that whether the fast
+ * (stat-only) codepath is used inside stash's diff operation depends
+ * on whether files are "racily clean", and there doesn't seem to be
+ * an easy way to force the exact required state.
+ */
+void test_stash_save__untracked_regression(void)
+{
+	git_checkout_options opts = GIT_CHECKOUT_OPTIONS_INIT;
+	const char *paths[] = {"what", "where", "how", "why"};
+	git_reference *head;
+	git_commit *head_commit;
+	git_buf untracked_dir;
+
+	const char* workdir = git_repository_workdir(repo);
+
+	git_buf_init(&untracked_dir, 0);
+	git_buf_printf(&untracked_dir, "%sz", workdir);
+
+	cl_assert(!p_mkdir(untracked_dir.ptr, 0777));
+
+	cl_git_pass(git_repository_head(&head, repo));
+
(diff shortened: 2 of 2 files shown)


novalis:
Of course, this fix is just a point fix, and it's possible that there's a cleaner way to solve this. But it does work.

tests/stash/save.c
```diff
@@ -188,6 +188,46 @@ void test_stash_save__can_include_untracked_and_ignored_files(void)
...
+	const char *paths[] = {"what", "where", "how", "why"};
+	git_reference *head;
+	git_commit *head_commit;
+
+	const char* workdir = git_repository_workdir(repo);
+	size_t len = strlen(workdir);
+	char* untracked_dir = malloc(len + 2);
+
+	strcpy(untracked_dir, workdir);
+	strcpy(untracked_dir + len, "z");
```
pks-t: I'd prefer to just use `git_buf` instead of pointer arithmetic. I mean this here is simple enough and unlikely to break, but `git_buf_printf(&untracked_dir, "%sz", workdir)`s is more readable.

tests/stash/save.c
```diff
@@ -188,6 +188,46 @@ void test_stash_save__can_include_untracked_and_ignored_files(void)
...
+void test_stash_save__untracked_regression(void)
+{
+	git_checkout_options opts = GIT_CHECKOUT_OPTIONS_INIT;
+	const char *paths[] = {"what", "where", "how", "why"};
+	git_reference *head;
+	git_commit *head_commit;
+
+	const char* workdir = git_repository_workdir(repo);
+	size_t len = strlen(workdir);
+	char* untracked_dir = malloc(len + 2);
```
pks-t: Pointers belong with the variable :)

src/diff_generate.c
```diff
@@ -273,7 +273,8 @@ static git_diff_delta *diff_delta__last_for_item(
 		break;
 	case GIT_DELTA_MODIFIED:
 		if (git_oid__cmp(&delta->old_file.id, &item->id) == 0 ||
-			git_oid__cmp(&delta->new_file.id, &item->id) == 0)
+		    (delta->new_file.mode == item->mode &&
```
pks-t: Do modes have to be completely the same? I get that it doesn't make sense in case one is a directory and the other is not. But what about when e.g. 644 changes to 755?

novalis: I have to admit that I don't really understand this code.  It seems completely weird to me that something gets added to the diff list and then later removed.  But gdb shows the "mode" for a directory as 16384 -- that is, these are git modes, not filesystem modes.  And looking at all three of the callers, they only happen in the case where one side of the comparison is a directory. So I don't think mode changes are relevant here.  What do you think?

ethomson: Goodness, I don't know what this code is doing, either.

novalis: If the question is "should X get removed from the list of diffs because it's actually identical to this other thing", I think a mode change should give the answer "no", because there is still a diff. 

(But, again, there's no case where this is called on two files; one arg is always a directory)

ethomson:
Ok, I stepped through here a few times to convince myself that I understood what was going on again.  Thanks for digging into this; our diff code is not trivial!  :D

Issue #4624, opened 2018-04-12
libgit2 status fails to ignore untracked file


rgburke:
Given the following directory structure where `bk.d` and `foto` are directories and `file1` and `file2` are files:
```
bk.d
├── file2
└── foto
    └── file1
```
The `.gitignore` file in this repository contains the following entries:
```
bk.d/*
!bk.d/*/
```

The output of `git status` is:
```
richard@DevNew:~/code/goworkspace/src/github.com/rgburke/grv$ git status
On branch master
Your branch is ahead of 'origin/master' by 1 commit.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean
```
However the file `bk.d/file2` is returned with status `GIT_STATUS_WT_NEW` by libgit2. Therefore `git status` ignores the file `bk.d/file2` but libgit2 doesn't.

Tested with version v0.27.0

pks-t:
Thanks for your report. Please check if #4670 fixes your issue.

rgburke:
@pks-t Yes, this fixes the issue. Thanks!

pks-t:
Thanks for checking and your report. Closing this via #4670

Issue #4623, opened 2018-04-12
.gitignore single wildcard negation rule negates subdirs too


hackhaslam:
With a .gitignore file like this:

```
*.test
!dir/*
```

and repository layout like this:

```
dir
  a.test
  subdir
    b.test
```

libgit2 correctly negates the ignore of dir/a.test, but also incorrectly negates the ignore of dir/subdir/b.test. The is in contrast to core git where only dir/a.test is negated and dir/subdir/b.test stays ignored. In core git the negation pattern has to be:

```
!dir/**
```

to negate entries in subdirs.

pks-t:
Thanks for your report. Please check if #4670 fixes your issue.

hackhaslam:
@pks-t That fixes it for me. Thanks for working on this!

ethomson:
Thanks for the report @hackhaslam !

PR #4670, opened 2018-06-05
Fix negative gitignore rules with leading directories 

pks-t:
When computing whether a file is ignored, we simply search for the first
matching rule and return whether it is a positive ignore rule (the file
is really ignored) or whether it is a negative ignore rule (the file is
being unignored). Each rule has a set of flags which are being passed to
`fnmatch`, depending on what kind of rule it is. E.g. in case it is a
negative ignore we add a flag `GIT_ATTR_FNMATCH_NEGATIVE`, in case it
contains a glob we set the `GIT_ATTR_FNMATCH_HASGLOB` flag.

One of these flags is the `GIT_ATTR_FNMATCH_LEADINGDIR` flag, which is
always set in case the pattern has a trailing "/*" or in case the
pattern is negative. The flag causes the `fnmatch` function to return a
match in case a string is a leading directory of another, e.g. "dir/"
matches "dir/foo/bar.c". In case of negative patterns, this is wrong in
certain cases.

Take the following simple example of a gitignore:

    dir/
    !dir/

The `LEADINGDIR` flag causes "!dir/" to match "dir/foo/bar.c", and we
correctly unignore the directory. But take this example:

    *.test
    !dir/*

We expect everything in "dir/" to be unignored, but e.g. a file in a
subdirectory of dir should be ignored, as the "*" does not cross
directory hierarchies. With `LEADINGDIR`, though, we would just see that
"dir/" matches and return that the file is unignored, even if it is
contained in a subdirectory. Instead, we want to ignore leading
directories here and check "*.test". Afterwards, we have to iterate up
to the parent directory and do the same checks.

To fix the issue, disallow matching against leading directories in
gitignore files. This can be trivially done by just adding the
`GIT_ATTR_FNMATCH_NOLEADINGDIR` to the spec passed to
`git_attr_fnmatch__parse`. Due to a bug in that function, though, this
flag is being ignored for negative patterns, which is fixed in this
commit, as well. As a last fix, we need to ignore rules that are
supposed to match a directory when our path itself is a file.

All together, these changes fix the described error case.

---

After two days of despair and constantly pulling my hair out I finally got a fix for both #4623 and #4624.

### DIFF
--- a/src/attr_file.c
+++ b/src/attr_file.c
@@ -594,8 +594,9 @@ int git_attr_fnmatch__parse(
 	}
 
 	if (*pattern == '!' && (spec->flags & GIT_ATTR_FNMATCH_ALLOWNEG) != 0) {
-		spec->flags = spec->flags |
-			GIT_ATTR_FNMATCH_NEGATIVE | GIT_ATTR_FNMATCH_LEADINGDIR;
+		spec->flags = spec->flags | GIT_ATTR_FNMATCH_NEGATIVE;
+		if ((spec->flags & GIT_ATTR_FNMATCH_NOLEADINGDIR) == 0)
+			spec->flags |= GIT_ATTR_FNMATCH_LEADINGDIR;
 		pattern++;
 	}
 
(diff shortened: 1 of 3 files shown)


pks-t:
As usual, I recommend reading this PR in commit-order together with their respective explanations in the commit messages.

ethomson:
I'm glad you tackled this problem, for I could not have solved it as I no longer have hair to pull out in frustration.  👍

PR #4676, opened 2018-06-09
editorconfig: allow trailing whitespace in markdown

ethomson:
Markdown uses trailing whitespace to indicate context; eg, a blank line
with four spaces indicates that there is a hard break between the
(indented) lines surrounding that line.  Ensure that editors do not
remove this blank.


pks-t:
I didn't even know we had such a file. Thanks!

PR #4677, opened 2018-06-10
Stop leaking the memory

ethomson:
We had some memory leaks sneak in yesterday.  This fixes one of them; I haven't tracked down the other.


ethomson:
Duh, it was right in front of me.  (It was a bit tricky at first since valgrind was showing that it was leaking from an allocation in the test setup phase - the original allocation of the object and insertion into the cache.)

PR #4681, opened 2018-06-12
docs: fix statement about tab width

pks-t:
The libgit2 project mostly follows the coding style of git and thus
the linux project. While those two projects use a recommended tab width
of eight spaces, we instruct users to set their editor's tab width to
four spaces. Fix this to say eight instead.


ethomson:
Blaming this change is funny - you're making it 8.  It was 4.  Before that, it was 3.

PR #4682, opened 2018-06-14
Automated indent using uncrustify

tiennou:
Here's the result of a week of munging parameters to whatstyle and/or unformat to make them output configurations for uncrustify/clang-format so code formatting can be automated.

This is the one that has the least amount of churn. I've not looked too precisely at it (I've been using a few select files for checks), so something could be awful (eg. the base64 encode table tends to not survive).

Note that there's a handful of branch in my fork under `indent/` which represents various attempts (some are mishaps from me). Some I find relevant are :
[1)](https://github.com/libgit2/libgit2/compare/master...tiennou:indent/whatstyle-1) whatstyle with clang-format.
[2)](https://github.com/libgit2/libgit2/compare/master...tiennou:indent/uncrustify-2) a mostly-manually done config, using UniversalIndentGUI (1st commit) + some suggestions from `whatstyle --variants`.
[3)](https://github.com/libgit2/libgit2/compare/master...tiennou:indent/unformat-minimal-1-tweak) unformat, after I understood how it worked (1st commit) + some manual tweaks because it missed indent settings.

I've included 2 scripts, `./scripts/format.sh` which uses whatever tool has a config available to format everything under `src` & `include` (apart from the dependencies) and `./script/format-machine.sh`, which calls either `unformat` or `whatstyle` (note, I've not retested it after I added the argument switch mechanism so YMMV) on a few select files (which may or may not have been good examples).

I was disappointed to discover that `clang-format` doesn't cope at all with our public function signatures. It seems to always want to have a newline after the `GIT_EXTERN(type)` part, hence the switch to `uncrustify`. AFAICS `uncrustify` has more knobs, but cannot perform some niceties (eg. aligning assign operators).

Another thing is that none of the tool decided that 8 was the correct indent to use, and none of them seem to support our mixed tab/space indentation, so everything is now tabs everywhere.


tiennou:
Note that it's for discussion purposes only, I don't propose we merge this as is 😉.

src/blame_git.c
```diff
@@ -480,15 +487,15 @@ static git_blame__origin* find_origin(
...
 static int pass_whole_blame(git_blame *blame,
-		git_blame__origin *origin, git_blame__origin *porigin)
+	git_blame__origin *origin, git_blame__origin *porigin)
 {
 	git_blame__entry *e;
 
 	if (!porigin->blob &&
-	    git_object_lookup((git_object**)&porigin->blob, blame->repository,
-				git_blob_id(origin->blob), GIT_OBJ_BLOB) < 0)
+	    git_object_lookup((git_object * *)&porigin->blob, blame->repository,
```
pks-t: Haha, this is so awful

src/clone.c
```diff
@@ -202,26 +202,26 @@ static int update_head_to_remote(
...
-		const char *reflog_message)
+	git_repository *repo,
+	const char *remote_name,
+	const char *branch,
+	const char *reflog_message)
 {
 	int retcode;
 	git_buf remote_branch_name = GIT_BUF_INIT;
-	git_reference* remote_ref = NULL;
+	git_reference*remote_ref = NULL;
```
pks-t: :confused:

ethomson:
I'm sorry that we left this linger for so long — I'm happy for automated formatting tools but we have megaconflicts here. Let's try to get the clang formatter nailed down.

Issue #4684, opened 2018-06-17
git_repository_path is incorrect for worktree from bare repo


GeorgeLyon:
### Reproduction steps

I reproduced this using Rust's `cargo` tool. If you try to run `cargo build` inside of a worktree of a bare git repo, the command fails because git_repository_path returns null.

1. Clone any Rust project with `git clone --bare`
2. In the bare repo, use `git worktree add` to create a worktree
3. try `cargo build` in the newly created worktree 
4. `cargo build` fails because git_repository_path returns NULL

### Expected behavior

git_repository_path should return the path of the worktree.

### Actual behavior

git_repository_path returns NULL

ethomson:
Have you opened an issue with cargo about this issue?

GeorgeLyon:
I have not yet, because it seems that this is specific to libgit2 (and when it is fixed, cargo should work correctly)

ethomson:
It’s not.  Please report it to them so that they can make their tool worktree aware.  Thanks!

ethomson:
Thanks @tiennou for chasing this down.  I was obviously wrong and this is indeed our bug.  Oops!  Thanks for reporting it @GeorgeLyon!

tiennou:
Sorry, misunderstanding. I investigated this in #4686, and I'm [confident](https://github.com/libgit2/libgit2/pull/4686/commits/853657f9f675c0d29aa4752e06f4118fb1566635) this is fixed on our side, so I suspect cargo hasn't updated to 0.27 (which IIRC has #4630, which fixes the underlying issue).

@GeorgeLyon You might want to file an issue with the cargo developers so they can investigate their code, just to make sure they're aware of the issue and can respond accordingly.

pks-t:
Yeah, the test explicitly calls `git_repository_path` on that worktree. Closing this issue again as it should be fixed. @GeorgeLyon, in case I'm wrong, please feel free to reopen.

pks-t:
Sigh. I reviewed @tiennou's PR already and thus thought it was merged already, that's why I already closed this issue. Anyway, as we're already tracking the issue in there I'll just keep this issue closed for now.

PR #4686, opened 2018-06-18
Fix git_worktree_validate failing on bare repositories

### LINKED ISSUE #4669: git_worktree_validate() fails to validate worktree associated with bare repository
### Reproduction steps

1. Open bare repository that has at least one work tree added to it. git_repository_open(repo, "repo.git")

2. Get the working tree names (which works fine). git_worktree_list(items, repo)

3. Try to get a git_worktree structure with git_worktree_lookup(wt, repo, items.strings[0])

4. Validate tree structure with git_worktree_validate(tree) 

### Expected behavior
expected to return 0 (no error)

### Actual behavior
git_worktree_validate() returns error code -2 because there is no parent_path

Why is a tree considered invalid if there's no workdir for the bare repository? Is this related to issue 4625?

bae6ed62010508c2e2b0559f9eee41e62e52eb7a

macOS 10.13.4

tiennou:
This fixes #4669, as well as checking that #4684 really isn't our fault (I wasn't sure, since #4630 was about that problem).

Sidenotes :

I was expecting `git_worktree_validate` to return a "logical C bool", but it instead return negative error codes that don't map to "the standard" ones. That was surprising 😉.

Looking at the diff, I'm also unsure what the `list_bare` test I moved was actually testing, as it's not run against the testsuite's repository.


ethomson:
> I was expecting git_worktree_validate to return a "logical C bool", but it instead return negative error codes that don't map to "the standard" ones. That was surprising 😉.

I wonder if this was an oversight from debugging?  The documentation mentions that these will return 0 "or an error code".  I would _much_ rather see this returning something in the scope of `git_error_code`.

tiennou:
Well, I'm not against changing it, but that's against the no-API-change rule, so we'd have to put up a replacement.

```c
typedef enum {
  ok = 0,
  bad = -1,
  …
} git_worktree_status;
int git_worktree_is_valid(git_worktree_status *status, git_worktree *wt);
```

That would return a 1 on success, 0 on failure, and set *status to the error, which makes it usable like this :

```c
int state;
if (!git_worktree_is_valid(&state, wt) && state == bad1) {
}
```
Not happy with the enum name though…

ethomson:
We're not changing the API.  We're fixing the function to behave as it's documented.

tiennou:
I might be missing something, but I see only three ways to fix that function :

- do nothing. The documentation is right, because we either return a 0, or a negative error code. That negative error code is completely out of touch with the API expectations, but can't be changed because API is frozen.
- do nothing, but clarify the documentation state that it will return `GIT_ERROR`/`-1` if the gitdir is invalid, `-2` (which doesn't actually map to anything) on a missing parent workdir, `GIT_ENOTFOUND`/ `-3` if the commondir cannot be found. Still unexpected, but at least the documentation is correct.
- provide a replacement that doesn't use "untyped" negative error codes and deprecate the old one.

ethomson:
>  The documentation is right, because we either return a 0, or a negative error code. 

The documentation _is_ right.  The code is wrong.  We are not returning error codes, we are returning random numbers.  Nobody is relying on that code returning `-2` or `-3` - because it isn't documented to do that, so there's no backward compatibility to keep.

No need to overthink this, we should just fix the code.

tiennou:
Okay, thanks for the clarification. I'll remap everything to `GIT_ERROR` then.

ethomson:
Awesome, thanks!

tests/worktree/bare.c
```diff
@@ -50,3 +50,13 @@ void test_worktree_bare__add(void)
 	git_worktree_free(wt);
 	git_repository_free(wtrepo);
 }
+
+void test_worktree_bare__added_worktree_validates(void)
+{
+	git_worktree *wt;
+
+	cl_git_pass(git_worktree_add(&wt, g_repo, "name", WORKTREE_REPO, NULL));
+	cl_assert_equal_i(0, git_worktree_validate(wt));
```
pks-t: I would've just added this additional call to the previous test, but yeah, doesn't really matter too much.

tests/clar_libgit2.c
```diff
@@ -319,6 +319,35 @@ const char* cl_git_path_url(const char *path)
 	return url;
 }
 
+const char *cl_git_sandbox_path(int is_dir, const char *path, ...)
+{
+	static char _temp[4096];
```
pks-t: `GIT_PATH_MAX`?

tests/clar_libgit2.c
```diff
@@ -319,6 +319,35 @@ const char* cl_git_path_url(const char *path)
...
+			cl_git_pass(git_buf_joinpath(&buf, buf.ptr, path));
+		}
+		va_end(arg);
+	}
+
+	cl_git_pass(git_path_prettify(&buf, buf.ptr, NULL));
+	if (is_dir)
+		git_path_to_dir(&buf);
+
+	git_buf_copy_cstr(_temp, 4096, &buf);
```
pks-t: You should check whether this truncates `buf`, first. Agreed, it is very unlikely, but I don't want to clean up (rm -rf) truncated paths at any point in time.

tests/worktree/open.c
```diff
@@ -11,28 +11,11 @@ static worktree_fixture fixture =
...
-	cl_git_pass(git_buf_joinpath(&path, path.ptr, "worktrees"));
-	cl_git_pass(git_buf_joinpath(&path, path.ptr, wtdir));
-	cl_git_pass(git_path_prettify(&path, path.ptr, NULL));
-	cl_git_pass(git_path_to_dir(&path));
-	cl_assert_equal_s(wt->gitdir, path.ptr);
-
-	git_buf_dispose(&path);
+	cl_assert_equal_s(wt->workdir, cl_git_sandbox_path(1, wtdir, NULL));
+	cl_assert_equal_s(wt->gitlink, cl_git_sandbox_path(0, wtdir, ".git", NULL));
+	cl_assert_equal_s(wt->gitdir, cl_git_sandbox_path(1, parentdir, ".git", "worktrees", wtdir, NULL));
```
pks-t: Very nice cleanup

tests/clar_libgit2.c
```diff
@@ -319,6 +319,35 @@ const char* cl_git_path_url(const char *path)
...
 }
 
+const char *cl_git_sandbox_path(int is_dir, const char *path, ...)
+{
+	static char _temp[4096];
+	git_buf buf = GIT_BUF_INIT;
+	va_list arg;
+
+	cl_git_pass(git_buf_sets(&buf, clar_sandbox_path()));
+	if (path) {
```
pks-t: This check is kind of superfluous. The signature should rather be `const char *cl_git_sandbox_path(int is_dir, ...)`. Then you could avoid this additional check and just enter right into the `while (va_arg(...) != NULL)` loop.

tests/worktree/bare.c
ethomson: Also it should probably be `cl_git_pass` instead of `cl_assert_equal_i(0, ...`

tests/clar_libgit2.c
tiennou: True. `is_dir` was added afterward, hence the path test. Fixed.

tests/clar_libgit2.c
tiennou: Fixed by checking the length of buf before copying.

tests/worktree/bare.c
tiennou: Fixed

tests/clar_libgit2.c
tiennou: True (inspiration was `clar_sandbox_path()`, which has no such define). Fixed.

tests/worktree/open.c
tiennou: Thanks 😉.

tiennou:
Ok, this is now green and ready to go!

tests/clar_libgit2.c
```diff
@@ -319,6 +319,35 @@ const char* cl_git_path_url(const char *path)
...
+	}
+	va_end(arg);
+
+	cl_git_pass(git_path_prettify(&buf, buf.ptr, NULL));
+	if (is_dir)
+		git_path_to_dir(&buf);
+
+	/* make sure we won't truncate */
+	cl_assert(git_buf_len(&buf) < GIT_PATH_MAX);
+	git_buf_copy_cstr(_temp, 4096, &buf);
```
pks-t: Sorry to be nitpicky, but this should probably use the same constant `GIT_PATH_MAX` or `sizeof(_temp)`, shouldn't it?

tiennou: Good catch. Fixed!

PR #4687, opened 2018-06-18
patch_parse: populate line numbers while parsing diffs

### LINKED ISSUE #4672: git_diff_line.{new,old}_lineno is incorrect
I believe that I have found a bug with git_diff_line, where new_lineno and old_lineno are incorrect. I first found the bug in pygit2 and filed an issue there https://github.com/libgit2/pygit2/issues/797 . I will copy information from that issue to this.

### Reproduction steps
```
#include <stdio.h>
#include <string.h>
#include <git2.h>

const char * text = "diff --git a/a b/a\n"
                    "index 7f129fd..af431f2 100644\n"
                    "--- a/a\n"
                    "+++ b/a\n"
                    "@@ -3 +3 @@\n"
                    "-a contents 2\n"
                    "+a contents\n";

void check(int err, const char *msg)
{
    if (err < 0) { printf("Error: %s\n", msg); exit(1); }
}

void main()
{
    git_diff *diff;
    git_patch *patch;
    const git_diff_hunk *hunk; size_t n;
    const git_diff_line *line;

    printf("%s", text);
    printf("\n\n");

    check(git_diff_from_buffer(&diff, text, strlen(text)), "git_diff_from_buffer");
    check(git_patch_from_diff(&patch, diff, 0), "git_patch_from_diff");
    check(git_patch_get_hunk(&hunk, &n, patch, 0), "git_patch_get_hunk");
    for (int i=0; i<n; i++)
    {
      check(git_patch_get_line_in_hunk(&line, patch, 0, i), "git_patch_get_line_in_hunk");
      printf("%d: %d -> %d\n", i, line->old_lineno, line->new_lineno);
    }
}
```
### Expected behavior
Something like the following is printed:
```
0: 3 -> 3
```

### Actual behavior
Prints:
```
0: 0 -> 0
```

0.27.0

At least CentOS

tiennou:
This should fix #4672. Looks like providing the "parsed" line numbers in the line struct was never actually wired (!).

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -563,6 +563,8 @@ static int parse_hunk_body(
 		char c;
 		int origin;
 		int prefix = 1;
+		int old_lineno = hunk->hunk.old_start + (hunk->hunk.old_lines - oldlines);
+		int new_lineno = hunk->hunk.new_start + (hunk->hunk.new_lines - newlines);
 
 		if (ctx->parse_ctx.line_len == 0 || ctx->parse_ctx.line[ctx->parse_ctx.line_len - 1] != '\n') {
 			error = git_parse_err("invalid patch instruction at line %"PRIuZ,
@@ -586,11 +588,13 @@ static int parse_hunk_body(
 		case '-':
 			origin = GIT_DIFF_LINE_DELETION;
 			oldlines--;
+			new_lineno = -1;
 			break;
 
 		case '+':
 			origin = GIT_DIFF_LINE_ADDITION;
 			newlines--;
+			old_lineno = -1;
 			break;
 
 		default:
@@ -607,6 +611,9 @@ static int parse_hunk_body(
 		line->content_len = ctx->parse_ctx.line_len - prefix;
 		line->content_offset = ctx->parse_ctx.content_len - ctx->parse_ctx.remain_len;
 		line->origin = origin;
+		line->num_lines = 1;
+		line->old_lineno = old_lineno;
+		line->new_lineno = new_lineno;
 
 		hunk->line_count++;
 	}
--- a/tests/diff/parse.c
+++ b/tests/diff/parse.c
@@ -288,3 +288,74 @@ void test_diff_parse__patch_roundtrip_succeeds(void)
 	git_buf_dispose(&patchbuf);
 	git_buf_dispose(&diffbuf);
 }
+
+#define cl_assert_equal_i_src(i1,i2,file,line) clar__assert_equal(file,line,#i1 " != " #i2, 1, "%d", (int)(i1), (int)(i2))
+
+static void cl_git_assert_lineinfo_(int old_lineno, int new_lineno, int num_lines, git_patch *patch, size_t hunk_idx, size_t line_idx, const char *file, int lineno)
+{
+	const git_diff_line *line;
+
+	cl_git_expect(git_patch_get_line_in_hunk(&line, patch, hunk_idx, line_idx), 0, file, lineno);
+	cl_assert_equal_i_src(old_lineno, line->old_lineno, file, lineno);
+	cl_assert_equal_i_src(new_lineno, line->new_lineno, file, lineno);
+	cl_assert_equal_i_src(num_lines, line->num_lines, file, lineno);
+}
+
+#define cl_git_assert_lineinfo(old, new, num, p, h, l) \
+	cl_git_assert_lineinfo_(old,new,num,p,h,l,__FILE__,__LINE__)
+
+
+void test_diff_parse__issue4672(void)
+{
+	const char *text = "diff --git a/a b/a\n"
+	"index 7f129fd..af431f2 100644\n"
+	"--- a/a\n"
+	"+++ b/a\n"
+	"@@ -3 +3 @@\n"
+	"-a contents 2\n"
+	"+a contents\n";
+
+	git_diff *diff;
+	git_patch *patch;
+	const git_diff_hunk *hunk;
+	size_t n, l = 0;
+
+	cl_git_pass(git_diff_from_buffer(&diff, text, strlen(text)));
+	cl_git_pass(git_patch_from_diff(&patch, diff, 0));
+	cl_git_pass(git_patch_get_hunk(&hunk, &n, patch, 0));
+
+	cl_git_assert_lineinfo(3, -1, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(-1, 3, 1, patch, 0, l++);
+
+	cl_assert_equal_i(n, l);
+
+	git_patch_free(patch);
+	git_diff_free(diff);
+}
+
+void test_diff_parse__lineinfo(void)
+{
+	const char *text = PATCH_ORIGINAL_TO_CHANGE_MIDDLE;
+	git_diff *diff;
+	git_patch *patch;
+	const git_diff_hunk *hunk;
+	size_t n, l = 0;
+
+	cl_git_pass(git_diff_from_buffer(&diff, text, strlen(text)));
+	cl_git_pass(git_patch_from_diff(&patch, diff, 0));
+	cl_git_pass(git_patch_get_hunk(&hunk, &n, patch, 0));
+
+	cl_git_assert_lineinfo(3, 3, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(4, 4, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(5, 5, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(6, -1, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(-1, 6, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(7, 7, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(8, 8, 1, patch, 0, l++);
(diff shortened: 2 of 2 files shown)


tests/diff/parse.c
```diff
@@ -288,3 +288,73 @@ void test_diff_parse__patch_roundtrip_succeeds(void)
...
+
+	cl_assert_equal_i(n - 1, l);
+
+	git_patch_free(patch);
+	git_diff_free(diff);
+}
+
+void test_diff_parse__lineinfo(void)
+{
+	const char *text = PATCH_ORIGINAL_TO_CHANGE_MIDDLE;
```
tiennou:
```
#define PATCH_ORIGINAL_TO_CHANGE_MIDDLE \
	"diff --git a/file.txt b/file.txt\n" \
	"index 9432026..cd8fd12 100644\n" \
	"--- a/file.txt\n" \
	"+++ b/file.txt\n" \
	"@@ -3,7 +3,7 @@ this is some context!\n" \
	" around some lines\n" \
	" that will change\n" \
	" yes it is!\n" \
	"-(this line is changed)\n" \
	"+(THIS line is changed!)\n" \
	" and this\n" \
	" is additional context\n" \
	" below it!\n"
```
✨

src/patch_parse.c
```diff
@@ -607,6 +611,9 @@ static int parse_hunk_body(
 		line->content_len = ctx->parse_ctx.line_len - prefix;
 		line->content_offset = ctx->parse_ctx.content_len - ctx->parse_ctx.remain_len;
 		line->origin = origin;
+		line->num_lines = 1;
```
tiennou: I'm suspicious of this : is there actually a way of building a text-based diff with embedded newlines 😧 ?

tests/diff/parse.c
tiennou: Also, I don't know what to make of the "this is some context!\n" line. Right now it's completely skipped.

ethomson:
Yeah, this is an omission.  But note that in #4672, the problem is not with a parsed patch (since there's not a lot of API to produce those yet) but with a generated patch.  So I probably broke what was there (or it never worked right, I'm not sure which).

So we should add some tests for both (or general tests).

tests/diff/parse.c
```diff
@@ -288,3 +288,74 @@ void test_diff_parse__patch_roundtrip_succeeds(void)
...
+	cl_git_pass(git_patch_get_hunk(&hunk, &n, patch, 0));
+
+	cl_git_assert_lineinfo(3, 3, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(4, 4, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(5, 5, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(6, -1, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(-1, 6, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(7, 7, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(8, 8, 1, patch, 0, l++);
+	cl_git_assert_lineinfo(9, 9, 1, patch, 0, l++);
```
ethomson: I don't love this.  I realize that you're aiming for brevity here but it's _really_ hard for me to read and understand what this at a glance.  I would much rather you unrolled these into simple tests.

src/patch_parse.c
ethomson: I don't think that's a thing.  I don't know how this could ever be anything but 1.

tests/diff/parse.c
ethomson: That sounds correct.  We should be surfacing that in the hunk header but not as a line in the diff.

tests/diff/parse.c
tiennou: And it's not even complete… Here's an [example](https://github.com/libgit2/libgit2/blob/master/tests/diff/patch.c?utf8=%E2%9C%93#L313) of the expanded version. My one isn't as exhaustive (origin and offset are missing), but it tests each line, so it's about 5 lines of asserts * 9 lines of patch => 45 lines of test.

Would something like this be more acceptable ?
```
cl_git_pass(git_line_from_hunk_patch_diff(&line, patch, 0, l++));
cl_git_assert_lineinfo(line, origin, old, new, newlines, offset, "stuff\n");
```

tiennou:
> But note that in #4672, the problem is not with a parsed patch (since there's not a lot of API to produce those yet) but with a generated patch.

Sorry, semantics. "parsed" means "from some file", and generated means "comes from xdiff", mostly ?(as in, at some point we had access to both A & B).

I'm pretty sure the OP is working with the first kind of patch, because the 2nd kind is tested ([tests/diff/patch.c:260](https://github.com/libgit2/libgit2/blob/master/tests/diff/patch.c?utf8=%E2%9C%93#L260)).

> So I probably broke what was there (or it never worked right, I'm not sure which).

[Never hooked up](d34f68261ef95b517944d4fa89ee13b4a68d3cb4).

src/patch_parse.c
tiennou: Looks like the xdiff code also doesn't set it. The only setters are in diff_print.c.

tests/diff/parse.c
pks-t: Don't use an expanded version, but simply use an array :)

```
struct {
    size_t start;
    size_t end;
    size_t length; /* No idea what those fields really are, but you get what I mean */
} expected_lineinfos[] = {
    { 3, 3, 1 },
    { 4, 4, 1 },
    ...
};

for (i = 0; i < ARRAY_SIZE(expected_lineinfos); i++) {
    git_line_from_hunk_patch_diff(&line, patch, 0, i);
    cl_assert_equal_i(old_lineno, line->old_lineno);
    cl_assert_equal_i(new_lineno, line->new_lineno);
    cl_assert_equal_i(num_lines, line->num_lines);
}
```

tiennou: ⚠️ Incoming bikeshedding.

I'm not against going the array route (I contemplated doing that). What makes me prefer the macro route is that I can preserve "intent" (eg. `cl_git_assert_lineinfo` vs a bunch of `equal_i`), as well as `__LINE__` numbers, the latter helps to make the failure context relevant straight from CI (which is something I'm personally striving for when writing tests (hence my propension to add macros on top of helper functions). There are quite many tests that check line info around, and that, as pointed out above, more would be quite welcome (eg. I haven't found anything that test hunk headers context lines behavior), and thus my preference goes to a tight one-liner.

The array route doesn't allow for that preservation, and even though it looks cleaner, it's messier to preserve the line info with them. I have a feeling I have already wrote one of these array-ed+macro tests somewhere, which, given my constraints, was 😒.

ethomson: I'm torn.  I appreciate that you're trying to report line numbers, though it's not my highest priority.  I'd rather have readable tests that convey intent than to know the line number.  eg:

```
cl_assert_equal_i(old_lineno, line->old_lineno);
cl_assert_equal_i(new_lineno, line->new_lineno);
cl_assert_equal_i(num_lines, line->num_lines);
```

makes it _incredibly_ clear what we're doing when I look at this.  Now, I concede that regressions would be a bit challenging -- we would have an accurate line number, but that wouldn't necessarily correspond to the failing test data.  We would know, for example, that the `new_lineno` test failed, but not what the data was that caused it to fail, since it could have been any data from the array.

This is a pattern we have in a lot of places and I don't think that we see a lot of regressions around them.  So I'm inclined to just shrug and say it's okay.  I definitely _have_ done big refactorings where this was a bit annoying - for example, when refactoring the iterators - but it didn't scar me so much that I want to avoid it.

Is your motivation borne out of past experiences with this pattern?  Or are you trying to future proof things?  Again, my suggestion is that we go for simplicity in tests over strict failure-reporting accuracy here, but we can probably spend some time working to find a more readable compromise if you disagree.

pks-t: Well, in my opinion it doesn't help much to have a one-shot test helper that's only ever going to be used once. In that case, you can just as well inline it, as you'd be using line info anyway. I wanted to propose to just use `cl_assert_equal_i_fmt`, but it turns out `fmt` is not really a format but simply a string where the two ints should be printed into. Would've been nice to have a macro `cl_assert_equal_i_fmt(tests[i].old, line->old_lineno, "old line numbers not matching: expected %i, got %i in line '%s'", tests[i].old, line->old_lineno, line->str)`. But yeah.

Dunno. I'll retreat from this discussion as I do not think it's important enough. Both styles have their advantages and disadvantages.

tiennou: > a one-shot test helper that's only ever going to be used once.

Actually, the plan was to have one helper that checks `line_info` and use that everywhere. That would reinforce the current "generated patchs" test as well (that's what I was hinting at above, since some checks are missing).

As an aside, sometimes I feel limited by clar's architecture, and long for something like `rspec` behaviors… That would allow reusing complete test suites against different "initial conditions" (think bare repo vs non-bare repo worktrees).

pks-t: Yeah, I had the same feeling sometimes about clar. It' works reasonably well and is a breeze to use in most cases, but some kind of data/configuration/permutation driven interface would in some cases be really helpful. I mean you can easily set that up already, but you do loose information then like we see here in this example.

ethomson:
It's fine.  I'd rather have the functionality than a test that I can reason about.  If we ever have to revisit it, I'll either suffer or pull it apart.

PR #4689, opened 2018-06-19
Address ThreadSanitizer errors

zwaldowski:
* Cleans up and modernizes `thread-utils.h` to use C11 atomics when available, gaining a properly-instrumented use of `atomic_load` in `git_atomic_get`. Replaces the sole use of `.val` in the codebase.
* Disables the caching in `git_pool__system_page_size` under threading. Not caching does not seem to introduce any performance regression and seems similar to other codebases I can find doing the same thing.

### DIFF
--- a/src/pool.c
+++ b/src/pool.c
@@ -23,15 +23,24 @@ static void *pool_alloc_page(git_pool *pool, uint32_t size);
 
 uint32_t git_pool__system_page_size(void)
 {
+#ifndef GIT_THREADS
 	static uint32_t size = 0;
 
 	if (!size) {
+#else
+    uint32_t size = 0;
+    do {
+#endif
 		size_t page_size;
 		if (git__page_size(&page_size) < 0)
 			page_size = 4096;
 		/* allow space for malloc overhead */
 		size = page_size - (2 * sizeof(void *)) - sizeof(git_pool_page);
+#ifndef GIT_THREADS
 	}
+#else
+    } while (0);
+#endif
 
 	return size;
 }
--- a/src/thread-utils.h
+++ b/src/thread-utils.h
@@ -7,17 +7,25 @@
 #ifndef INCLUDE_thread_utils_h__
 #define INCLUDE_thread_utils_h__
 
-#if defined(__GNUC__) && defined(GIT_THREADS)
-# if (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 1))
-#  error Atomic primitives do not exist on this version of gcc; configure libgit2 with -DTHREADSAFE=OFF
+#if defined(GIT_THREADS)
+# if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L && !defined(__STDC_NO_ATOMICS__)
+#  define __GIT_THREADS_USE_C11 1
+#  include <stdatomic.h>
+# elif defined(__GNUC__) && __GNUC__ >= 4
+#  define __GIT_THREADS_USE_GNU 1
+# elif defined(GIT_WIN32)
+#  define __GIT_THREADS_USE_WIN32 1
+# else
+#  error Atomic primitives do not exist on this compiler; configure libgit2 with -DGIT_THREADS=0
 # endif
 #endif
 
-/* Common operations even if threading has been disabled */
 typedef struct {
-#if defined(GIT_WIN32)
+#if __GIT_THREADS_USE_C11
+    atomic_int val;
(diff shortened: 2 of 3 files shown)


src/pool.c
```diff
@@ -23,15 +23,24 @@ static void *pool_alloc_page(git_pool *pool, uint32_t size);
...
 		size_t page_size;
 		if (git__page_size(&page_size) < 0)
 			page_size = 4096;
 		/* allow space for malloc overhead */
 		size = page_size - (2 * sizeof(void *)) - sizeof(git_pool_page);
+#ifndef GIT_THREADS
 	}
+#else
+    } while (0);
+#endif
```
pks-t: Well, this effectively removes the cache altogether on most platforms. On Linux-based systems this is basically not an issue, as `sysconf` will usually just return a static value without any syscalls at all. But on Windows this might be a different story, as we call `GetSystemInfo` there, which might be a bit more involved. And as we're using pools especially in performance-critical paths, this is a no-go.

In the end, the reported threading issue is really a non issue. There's only two states: the value either is uninitialized or it is initialized to a value that is never going to change during execution. So what happens in case two threads race? The worst thing that can happen is that the value gets initialized to the same value twice, which really is not a problem at all. And as stores are guaranteed to be atomic, another thread will only ever see either the correct value or an uninitialized value.

I agree that fixing issues reported by TSAN is nice to make TSAN itself more useful. But I'd argue that pessimizing a potentially performance critical path in order to fix a reported issue that cannot even have any negative consequences is the wrong thing to do.

src/thread-utils.h
```diff
@@ -7,28 +7,38 @@
 #ifndef INCLUDE_thread_utils_h__
 #define INCLUDE_thread_utils_h__
 
-#if defined(__GNUC__) && defined(GIT_THREADS)
-# if (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 1))
-#  error Atomic primitives do not exist on this version of gcc; configure libgit2 with -DTHREADSAFE=OFF
+#if defined(GIT_THREADS)
+# if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L && !defined(__STDC_NO_ATOMICS__)
+#  define __GIT_THREADS_USE_C11 1
```
pks-t: I'm afraid that using C11 for atomics won't work for us. Our project is trying to conform to C90, so we simply cannot use C11 features at all. I agree that it kind of sucks that we need to rely on GCC builtins for our atomic support instead, but for C90 there is no standardized way to do that.

src/pool.c
zwaldowski: There's no such thing as a benign data race. I don't know where you're getting the idea that stores are guaranteed to be atomic. If one stops assuming the happy path of a bog-standard x86 running some archaic Linux, then it becomes obvious that a problem like this must be taken seriously. The read of `size` in `if (!size)` can happen at any point during someone else's write to it.

Is there instrumentation data indicating that a cache is needed? Your uncertainty implies that there isn't. Every use I can find of a modern system that uses the page size in a calculation, including many platform allocators, use the `sysconf` on demand.

src/thread-utils.h
zwaldowski: > Our project is trying to conform to C90

I see it now in the readme, but this requirement is ill-documented. CMake makes no attempt to configure the compiler that way, so certainly most of distributions you package for have a separate default. I see a fair amount of code `#if` in the repo around C99 compilers, so I don't see why properly-checked progressive enhancement is a problem.

> … but for C90 there is no standardized way to do that.

It's kind of absurd to say this when 2 lines below there's a bespoke variant for Windows. The GCC builtins are known to be deficient; that's why they went through standardization. If I use `#if defined(__clang__)` and its builtins, is that somehow better? C11 atomics are also the only guaranteed-correct way for platforms now; if I write `#if defined(__APPLE__) || defined(__ANDROID__) || defined(__Fuchsia__)`, is that somehow better? If so, why?

ethomson:
> Is there instrumentation data indicating that a cache is needed?

You’re proposing to remove it and you didn’t measure the before and after?

src/pool.c
ethomson: >  The read of size in if (!size) can happen at any point during someone else's write to it.

I agree with you that if we’re running on a machine that can’t atomically store a 32 bit number then this would be problematic.  Can you give me a concrete example of such a machine?

ethomson:
It seems like discussion here has waned, so I'm closing this.  If you have an interest in picking this back up, we can reconsider.

PR #2824, opened 2015-01-15
add example "do_checkout_ref"

antis81:
@ethomson Here it is! My first example contribution. Not yet compiling! Needs some discussion!

PR #4692, opened 2018-06-23
Add a checkout example

tiennou:
This adds an example on how to perform checkouts with libgit2.

Note that it only does the "branch moving" part of checkout, so no `git checkout src/file.c`.

Ref #2824.


tiennou:
Note that while working with refishs, I've grown to like `git_annotated_commit` for its ability to abstract away the details (like the automated reflog messages).

And looking around its API, it might be worthwhile to add a `git_repository_set_head_from_annotated`, that would underlie all others `set_head`s we have. Also, I was a little confused that all the "virtual commits" creation lies in "merge.c", so maybe moving that back to "annotated.c" so the separation is cleaner ?

examples/checkout.c
```diff
@@ -0,0 +1,184 @@
...
+		} else {
+			break;
+		}
+	}
+}
+
+static void print_checkout_progress(const char *path, size_t completed_steps, size_t total_steps, void *payload)
+{
+	(void)payload;
+	printf("checkout: %s %" PRIuZ "/%" PRIuZ "\n", path, completed_steps, total_steps);
```
pks-t: Usually you'd have to first erase the previous line and write on that same line again. But for a simple example this is sufficient, I guess

examples/checkout.c
```diff
@@ -0,0 +1,184 @@
...
+
+	err = resolve_refish(&checkout_target, repo, args.argv[args.pos]);
+	if (err == 0) {
+		err = perform_checkout_ref(repo, checkout_target, &opts);
+	} else {
+		printf("failed to resolve %s: %s\n", args.argv[args.pos], giterr_last()->message);
+	}
+
+
+
```
pks-t: A few blank lines too much

examples/checkout.c
```diff
@@ -0,0 +1,184 @@
...
+	checkout_opts.checkout_strategy = GIT_CHECKOUT_SAFE;
+	if (opts->force) {
+		checkout_opts.checkout_strategy = GIT_CHECKOUT_FORCE;
+	}
+	if (opts->progress) {
+		checkout_opts.progress_cb = print_checkout_progress;
+	}
+	if (opts->perf) {
+		checkout_opts.perfdata_cb = print_perf_data;
+	}
```
pks-t: There's no need for all those braces around single-line statements

examples/checkout.c
```diff
@@ -0,0 +1,184 @@
...
+	(void)payload;
+	printf("perf: stat: %" PRIuZ " mkdir: %" PRIuZ " chmod: %" PRIuZ "\n",
+	       perfdata->stat_calls, perfdata->mkdir_calls, perfdata->chmod_calls);
+}
+
+static int perform_checkout_ref(git_repository *repo, git_annotated_commit *target, checkout_options *opts)
+{
+	git_checkout_options checkout_opts = GIT_CHECKOUT_OPTIONS_INIT;
+	git_commit *target_commit = NULL;
+	int err = 0;
```
pks-t: As the first statement that can fail always sets `err` there is no need to initialize it here

examples/checkout.c
```diff
@@ -0,0 +1,184 @@
...
+	}
+	if (opts->perf) {
+		checkout_opts.perfdata_cb = print_perf_data;
+	}
+
+	err = git_commit_lookup(&target_commit, repo, git_annotated_commit_id(target));
+	if (err != 0) {
+		printf("failed to lookup commit: %s\n", giterr_last()->message);
+		goto cleanup;
+	}
```
pks-t: We usually use the `check_lg2` function to do error handling in our examples, which also handles things like `giterr_last()`. I'm fine with not using it, but in that case you should print errors to `stderr`

examples/checkout.c
```diff
@@ -0,0 +1,184 @@
...
+{
+	git_repository *repo = NULL;
+	struct args_info args = ARGS_INFO_INIT;
+	checkout_options opts;
+	git_repository_state_t state;
+	git_annotated_commit *checkout_target = NULL;
+	int err = 0;
+	const char *path = ".";
+
+	opts.progress = 1;
```
pks-t: This line looks a bit lonely. Move it to `parse_options`, maybe, so that it does all the options handling?

examples/checkout.c
```diff
@@ -0,0 +1,184 @@
...
+
+	parse_options(&path, &opts, &args);
+
+	git_libgit2_init();
```
pks-t: You need to check the return value of that function.

examples/checkout.c
```diff
@@ -0,0 +1,184 @@
...
+	if (state != GIT_REPOSITORY_STATE_NONE) {
+		fprintf(stderr, "repository is in unexpected state %d\n", state);
+		goto cleanup;
+	}
+
+	if (args.pos >= args.argc) {
+		fprintf(stderr, "missing argument\n");
+		err = -1;
+		goto cleanup;
+	}
```
pks-t: This check looks like it belongs to `parse_opts`, as well

examples/checkout.c
```diff
@@ -0,0 +1,184 @@
...
+
+	err = resolve_refish(&checkout_target, repo, args.argv[args.pos]);
+	if (err == 0) {
+		err = perform_checkout_ref(repo, checkout_target, &opts);
+	} else {
+		printf("failed to resolve %s: %s\n", args.argv[args.pos], giterr_last()->message);
+	}
```
pks-t: I'd use our usual error handling style instead.

```
if (err < 0) {
    printf("failed...");
    goto cleanup;
}

err = perform_checkout_ref(repo, checkout_target, &opts);

File: examples/common.c [hunk 10]
```diff
@@ -182,6 +184,33 @@ static int match_int_internal(
...
+	int res = 0;
+
+	if (strcmp(found, opt) == 0) {
+		*out = 1;
+		return 1;
+	}
+
+	neg_opt = malloc(strlen(opt) + 3);
+	strcpy(neg_opt, "--no-");
+	strcat(neg_opt, opt + 2);
```
pks-t: This seems a bit heavy-handed ;)

```
if (!strncmp(found, "--no-", strlen("--no-")) &&
    !strcmp(found + strlen("--no-"), opt + 2)) {
        *out = 0;
        res = 1;
}
```

This is untested, but something like that would work.

examples/checkout.c
```diff
@@ -0,0 +1,184 @@
...
+ * with this software. If not, see
+ * <http://creativecommons.org/publicdomain/zero/1.0/>.
+ */
+
+#include "common.h"
+#include <assert.h>
+
+#ifdef _MSC_VER
+#define snprintf sprintf_s
+#endif
```
pks-t: This is unused

examples/common.c
tiennou: Nice, that's soo much better !

tiennou:
I've fixed/clarified most of the review comments. Clarifications because sometime™ we might want to demonstrate path-based checkouts, so the code in main was convoluted because of that.

The code is also commented now.

examples/common.c
```diff
@@ -184,6 +184,25 @@ static int match_int_internal(
...
+		return 1;
+	}
+
+	if (!strncmp(found, "--no-", strlen("--no-")) &&
+	    !strcmp(found + strlen("--no-"), opt + 2)) {
+		*out = 0;
+		res = 1;
+	}
+
+	return res;
```
pks-t: You could avoid the `res` variable by just returning `1` in the previous branch and `0` at the end of this function. That would make the asymmetry between the first and second conditional go away. Not urgent though, you should only address this if you're rerolling anyway

examples/common.h
```diff
@@ -90,6 +90,15 @@ extern int match_uint32_arg(
 extern int match_int_arg(
 	int *out, struct args_info *args, const char *opt, int allow_negative);
 
+/**
+ * Check current `args` entry against a "bool" `opt` (ie. --[no-]progress).
+ * If `opt` matches positively, out will be set to 1, or if `opt` matches
+ * negatively, out will be set to 0, and 1 will be returned.
+ * If neither the positive or the negative form of opt matched, out will be -1,
```
pks-t: The documentation is wrong. You either set `out` to 0 or 1, but never -1.

examples/checkout.c
```diff
@@ -0,0 +1,229 @@
...
+		print_usage();
+
+	/* Default values */
+	opts->progress = 1;
+
+	for (args->pos = 1; args->pos < args->argc; ++args->pos) {
+		const char *curr = args->argv[args->pos];
+		int bool_arg;
+
+		if (strcmp(curr, "--") == 0) {
```
pks-t: You should probably use `!strcmp`, for consistency with the next conditional. Might confuse readers otherwise

examples/checkout.c
```diff
@@ -0,0 +1,229 @@
...
+	/**
+	 * Now that the checkout has completed, we have to update HEAD.
+	 *
+	 * Depending on the "origin" of target (ie. it's an OID or a branch name),
+	 * we might need to detach HEAD.
+	 */
+	if (git_annotated_commit_ref(target)) {
+		err = git_repository_set_head(repo, git_annotated_commit_ref(target));
+	} else {
+		err = git_repository_set_head_detached(repo, git_annotated_commit_id(target));
```
pks-t: There's `git_repository_set_head_detached_from_annotated`

examples/checkout.c
```diff
@@ -0,0 +1,229 @@
...
+	exit(1);
+}
+
+static void parse_options(const char **repo_path, checkout_options *opts, struct args_info *args)
+{
+	if (args->argc <= 1)
+		print_usage();
+
+	/* Default values */
+	opts->progress = 1;
```
pks-t: You should `memset(opts, 0, sizeof(opts))`, as you never initialize the struct to all-zeroes

examples/checkout.c
```diff
@@ -0,0 +1,229 @@
...
+
+	/** Make sure we're not about to checkout while something else is going on */
+	state = git_repository_state(repo);
+	if (state != GIT_REPOSITORY_STATE_NONE) {
+		fprintf(stderr, "repository is in unexpected state %d\n", state);
+		goto cleanup;
+	}
+
+	if (args.pos >= args.argc) {
+		fprintf(stderr, "unhandled\n");
```
pks-t: Shouldn't this check be part of `parse_options`, together with usage instructions? I'd expect that function to only ever return where the options are correct

examples/checkout.c
```diff
@@ -0,0 +1,229 @@
...
+		err = -1;
+		goto cleanup;
+	} else if (strcmp("--", args.argv[args.pos])) {
+		/**
+		 * Try to checkout the given path
+		 */
+
+		fprintf(stderr, "unhandled path-based checkout\n");
+		err = 1;
+		goto cleanup;
```
pks-t: I'd also just remove this and error out in `parse_options` in case where paths are given. It makes the remaining logic a bit easier to read, I guess

examples/checkout.c
```diff
@@ -0,0 +1,229 @@
...
+	} else {
+		/**
+		 * Try to resolve a "refish" argument to a target libgit2 can use
+		 */
+		err = resolve_refish(&checkout_target, repo, args.argv[args.pos]);
```
pks-t: Resolving the ref here makes sense, but I'd parse the reference itself into the  options structure in `parse_options`

tiennou:
I've fixed some of the comments.

It seems we have a slight difference in how we expect option parsing to happen though, so here's my "view" : I can't access libgit2 while parsing options, because there might be a `--git-dir` option *after* the current argument (option which I'm using to test), so I can't resolve refishs, or paths, because the "current working directory" might be updated later. Hence, I really much prefer option parsing to be kept dumb, and I think it actually makes more sense as example code (if you want to perform a checkout, you can basically look at what `main()` is doing and ignore the option parsing).

Also, I've attempted to handle path-based checkouts, but just ended up craving for libgit2's already written support for those (it only took one "any directory relative stat call" to throw my hands up at the incoming strcat mess 😉).

examples/checkout.c
```diff
@@ -0,0 +1,231 @@
...
+			continue;
+		} else {
+			break;
+		}
+	}
+}
+
+/**
+ * This function is called to report progression, ie. it's called once with
+ * a NULL path and the number of total steps, then for each subsequent path,
```
pks-t: If this function is indeed called with a `NULL` path, then the `printf` call is undefined behaviour as "%s" expects a valid pointer

tiennou:
Fixed the UB. As always, thanks for the review @pks-t !

PR #4695, opened 2018-06-24
Fix type confusion in git_smart__connect

nelhage:
Nothing verifies that t->refs[0] is a GIT_PKT_REF. A remote can send
another packet type, ultimately resulting in a type confusion in
`git_smart__detect_caps`.

This was found by an oss-fuzz integration I'm working on.

### DIFF
--- a/src/transports/smart.c
+++ b/src/transports/smart.c
@@ -266,7 +266,12 @@ static int git_smart__connect(
 	/* We now have loaded the refs. */
 	t->have_refs = 1;
 
-	first = (git_pkt_ref *)git_vector_get(&t->refs, 0);
+	pkt = (git_pkt *)git_vector_get(&t->refs, 0);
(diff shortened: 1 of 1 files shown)


src/transports/smart.c
```diff
@@ -266,7 +266,12 @@ static int git_smart__connect(
 	/* We now have loaded the refs. */
 	t->have_refs = 1;
 
-	first = (git_pkt_ref *)git_vector_get(&t->refs, 0);
+	pkt = (git_pkt *)git_vector_get(&t->refs, 0);
+	if (pkt && GIT_PKT_REF != pkt->type) {
```
pks-t: We usually check the other way round, with the constant on the right-hand side. But here you seem to simply be adhering to style of surrounding code, so this is okay

pks-t:
Thanks for this fix!

PR #4696, opened 2018-06-24
Verify ref_pkt's are long enough

nelhage:
If the remote sends a too-short packet, we'll allow `len` to go
negative and eventually issue a malloc for <= 0 bytes on

```
pkt->head.name = git__malloc(alloclen);
```

This was found by an oss-fuzz integration I'm working on.


pks-t:
Thanks for the fix!

PR #4698, opened 2018-06-25
Fix assorted leaks found via fuzzing

nelhage:
These were all found by my work-in-progress oss-fuzz fuzzer.

### DIFF
--- a/src/transports/smart_protocol.c
+++ b/src/transports/smart_protocol.c
@@ -321,9 +321,7 @@ static int wait_while_ack(gitno_buffer *buf)
 	git_pkt_ack *ack = NULL;
 
 	while (1) {
-		if (pkt) {
-			git_pkt_free(pkt);
-		}
+		git_pkt_free(pkt);
 
 		if ((error = recv_pkt(&pkt, NULL, buf)) < 0)
 			return error;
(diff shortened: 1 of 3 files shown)


src/transports/smart_protocol.c
```diff
@@ -317,27 +317,34 @@ static int fetch_setup_walk(git_revwalk **out, git_repository *repo)
 static int wait_while_ack(gitno_buffer *buf)
 {
 	int error;
-	git_pkt_ack *pkt = NULL;
+	git_pkt *pkt = NULL;
+	git_pkt_ack *ack = NULL;
 
 	while (1) {
-		git__free(pkt);
+		if (pkt) {
```
pks-t: We usually do not check whether a pointer is set before freeing it. But in this case it is necessary, as `git_pkt_free` doesn't check for `NULL` pointers itself

src/transports/smart_protocol.c
```diff
@@ -622,7 +622,9 @@ int git_smart__download_pack(
 			}
 		}
 
-		git__free(pkt);
+		if (pkt) {
+			git_pkt_free(pkt);
```
pks-t: Huh, one more time. I know it's not your fault, but instead the existing code is to blame. But I'd highly appreciate if you did a preliminary commit which fixes `git_pkt_free` to return early in case a `NULL` pointer is being passed to it.

nelhage:
I hadn't! Thanks for the pointer, I'll look into that past attempt.

nelhage:
Fixed `git_pkt_free` to allow `NULL`.

ethomson:
Thanks for this.  🎉

PR #4699, opened 2018-06-25
git_refspec_transform: Handle NULL dst

nelhage:
per https://github.com/libgit2/libgit2/blob/b121b7ac972156cdc67b25be075fd62450b57f29/src/refspec.c#L85-L90 this is considered valid, so we should handle it downstream.

Found via triggering the assertion failure in  `git_buf_puts` in my work-in-progress fuzzer.

### DIFF
--- a/src/refspec.c
+++ b/src/refspec.c
@@ -252,7 +252,7 @@ int git_refspec_transform(git_buf *out, const git_refspec *spec, const char *nam
 	}
 
 	if (!spec->pattern)
-		return git_buf_puts(out, spec->dst);
+		return git_buf_puts(out, spec->dst ? spec->dst : "");
 
 	return refspec_transform(out, spec->src, spec->dst, name);
 }
--- a/tests/network/refspecs.c
+++ b/tests/network/refspecs.c
@@ -111,7 +111,8 @@ void test_network_refspecs__transform_mid_star(void)
 	assert_valid_transform("refs/*:refs/*", "refs/heads/master", "refs/heads/master");
 }
 
-void test_network_refspecs__no_dst(void) {
+void test_network_refspecs__no_dst(void)
+{
 	assert_valid_transform("refs/heads/master:", "refs/heads/master", "");
 }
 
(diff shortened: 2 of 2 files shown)


src/refspec.c
```diff
@@ -252,7 +252,7 @@ int git_refspec_transform(git_buf *out, const git_refspec *spec, const char *nam
 	}
 
 	if (!spec->pattern)
-		return git_buf_puts(out, spec->dst);
+		return git_buf_puts(out, spec->dst ? spec->dst : "");
```
pks-t: Do you want to write a test for it? I think it should be rather easy to do as there is already precedence of testing this function in "tests/network/refspecs.c".

nelhage: Done!

pks-t:
Thanks for the test!

I think I need a second opinion on the new behaviour. Thing is, when you parse a refspec "refs/heads/master:" with `git_refspec_parse`, you get a `src` of "refs/heads/master" and `dst` is `NULL`. But when you now transform such a refspec, you do not get `NULL` but instead you'd get an empty string (or rather a buffer containing an emtpy string). I can't help but feel that this is being inconsistent. The problem is obviously that you cannot return a `NULL` pointer via the `git_buf` out parameter.

Going a bit deeper: what would a caller of `git_refspec_transform` actually expect? He wants to transfor a source ref (potentially with globs) to a destination ref. But in case of "refs/heads/master:" there is no destination ref. Should we, instead of passing back an empty string, raise an error?

@ethomson, do you have some thoughts?

tests/network/refspecs.c
```diff
@@ -111,6 +111,10 @@ void test_network_refspecs__transform_mid_star(void)
 	assert_valid_transform("refs/*:refs/*", "refs/heads/master", "refs/heads/master");
 }
 
+void test_network_refspecs__no_dst(void) {
```
pks-t: The opening brace of functions should actually start on their own line :)

pks-t:
Thanks for your reasoning and this change!

PR #4700, opened 2018-06-25
C90 standard compliance

pks-t:
As noted in #4689, we currently do not enforce compliance with ISO C90. This PR fixes that for Unix platforms by first fixing all non-compliant code and then adding `-std=c90` to the compiler flags.


pks-t:
This PR is finally ready to be merged, opinions would be appreciated :)

PR #4701, opened 2018-06-25
streams: openssl: Handle error in SSL_CTX_new

nikital:
SIGSEGV otherwise...

I saw the segfault on Fedora when using libgit2 with nodegit. The segfault was in in SSL_CTX_set_verify because it was the first function that tried to actually write to the NULL pointer.


pks-t:
Once again: thanks for your fix! :)

PR #4702, opened 2018-06-25
Assorted Coverity fixes

tiennou:
This fixes May/June Coverity reports.

### DIFF
--- a/src/blame.c
+++ b/src/blame.c
@@ -133,8 +133,11 @@ git_blame* git_blame__alloc(
 		return NULL;
 	}
 
-	if (opts.flags & GIT_BLAME_USE_MAILMAP)
-		git_mailmap_from_repository(&gbr->mailmap, repo);
+	if (opts.flags & GIT_BLAME_USE_MAILMAP &&
+	    git_mailmap_from_repository(&gbr->mailmap, repo) < 0) {
+		git_blame_free(gbr);
+		return NULL;
+	}
 
 	return gbr;
 }
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -260,7 +260,8 @@ static int load_submodule_names(git_strmap **out, git_repository *repo, git_conf
 		git_strmap_insert(names, entry->value, git_buf_detach(&buf), &rval);
 		if (rval < 0) {
 			giterr_set(GITERR_NOMEMORY, "error inserting submodule into hash table");
-			return -1;
+			error = -1;
+			goto out;
 		}
 	}
 	if (error == GIT_ITEROVER)
--- a/src/transports/smart.c
+++ b/src/transports/smart.c
@@ -280,19 +280,22 @@ static int git_smart__connect(
 
 	/* Detect capabilities */
 	if ((error = git_smart__detect_caps(first, &t->caps, &symrefs)) == 0) {
+		/* If the only ref in the list is capabilities^{} with OID_ZERO, remove it */
+		if (1 == t->refs.length && !strcmp(first->head.name, "capabilities^{}") &&
+			git_oid_iszero(&first->head.oid)) {
+			git_vector_clear(&t->refs);
+			git_pkt_free((git_pkt *)first);
+		}
+
+		/* Keep a list of heads for _ls */
+		git_smart__update_heads(t, &symrefs);
+	} else if (error == GIT_ENOTFOUND) {
+		/* There was no ref packet received, or the cap list was empty */
+		error = 0;
+	} else {
+		giterr_set(GITERR_NET, "invalid response");
 		goto cleanup;
 	}
-
-	/* If the only ref in the list is capabilities^{} with OID_ZERO, remove it */
-	if (1 == t->refs.length && !strcmp(first->head.name, "capabilities^{}") &&
-	    git_oid_iszero(&first->head.oid)) {
-		git_vector_clear(&t->refs);
-		git_pkt_free((git_pkt *)first);
-	}
-
-	/* Keep a list of heads for _ls */
-	git_smart__update_heads(t, &symrefs);
-
 
 	if (t->rpc && (error = git_smart__reset_stream(t, false)) < 0)
 		goto cleanup;
--- a/src/transports/smart.c
+++ b/src/transports/smart.c
@@ -279,14 +279,13 @@ static int git_smart__connect(
 		return error;
 
 	/* Detect capabilities */
-	if (git_smart__detect_caps(first, &t->caps, &symrefs) < 0) {
-		free_symrefs(&symrefs);
-		return -1;
+	if ((error = git_smart__detect_caps(first, &t->caps, &symrefs)) == 0) {
+		goto cleanup;
 	}
 
 	/* If the only ref in the list is capabilities^{} with OID_ZERO, remove it */
 	if (1 == t->refs.length && !strcmp(first->head.name, "capabilities^{}") &&
-		git_oid_iszero(&first->head.oid)) {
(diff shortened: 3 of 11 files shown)


ethomson:
It looks like the smart pkt changes are causing test failures - could you take a look?

tiennou:
Rebased. Note that I had to partially revert a recent change because of the allocation semantics of `symrefs` (it's allocated/populated in smart_protocol.c/detect_caps, but its destructor is in smart.c).
So please don't merge, event though tests should be ok now.

tiennou:
That should be ok now. Feel free to :shipit: after reviewing.

src/blame.c
```diff
@@ -133,8 +133,11 @@ git_blame* git_blame__alloc(
 		return NULL;
 	}
 
-	if (opts.flags & GIT_BLAME_USE_MAILMAP)
-		git_mailmap_from_repository(&gbr->mailmap, repo);
+	if (opts.flags & GIT_BLAME_USE_MAILMAP &&
+		git_mailmap_from_repository(&gbr->mailmap, repo) < 0) {
```
pks-t: No need to change that now, but our agreed-upon code style is to indent continuation lines in conditionals with four spaces:

```
if (opts.flags & GIT_BLAME_USE_MAILMAP &&
    git_mailmap_from_repository(&gbr->mailmap, repo) < 0) {
        git_blame_free(gbr);
        return NULL;
}
```

src/blame.c
```diff
@@ -133,8 +133,11 @@ git_blame* git_blame__alloc(
 		return NULL;
 	}
 
-	if (opts.flags & GIT_BLAME_USE_MAILMAP)
-		git_mailmap_from_repository(&gbr->mailmap, repo);
```
pks-t: I think ignoring the error was actually intended. In case where we're unable to load the mail map, we simply ignore it and output commits with their original author/committer. You could add a comment and cast to `void`, though

src/submodule.c
```diff
@@ -260,7 +260,8 @@ static int load_submodule_names(git_strmap **out, git_repository *repo, git_conf
 		git_strmap_insert(names, entry->value, git_buf_detach(&buf), &rval);
 		if (rval < 0) {
 			giterr_set(GITERR_NOMEMORY, "error inserting submodule into hash table");
-			return -1;
+			error = -1;
+			goto out;
```
pks-t: We usually avoid freeing memory when we run in OOM situations. But as you cannot use `GITERR_CHECK_ALLOC` here I'm fine with this fix

src/transports/smart.c
```diff
@@ -279,30 +279,34 @@ static int git_smart__connect(
...
-	/* If the only ref in the list is capabilities^{} with OID_ZERO, remove it */
-	if (1 == t->refs.length && !strcmp(first->head.name, "capabilities^{}") &&
-		git_oid_iszero(&first->head.oid)) {
-		git_vector_clear(&t->refs);
-		git_pkt_free((git_pkt *)first);
+		/* Keep a list of heads for _ls */
+		git_smart__update_heads(t, &symrefs);
+	} else if (error == GIT_ENOTFOUND) {
+		/* There was no ref packet received, or the cap list was empty */
+        error = 0;
```
pks-t: Indentation is off here

src/transports/smart.c
```diff
@@ -279,30 +279,34 @@ static int git_smart__connect(
...
+	} else {
+        giterr_set(GITERR_NET, "invalid response");
+        goto cleanup;
```
pks-t: And here as well

src/transports/smart.c
```diff
@@ -279,30 +279,34 @@ static int git_smart__connect(
 		return error;
 
 	/* Detect capabilities */
-	if (git_smart__detect_caps(first, &t->caps, &symrefs) < 0) {
-		free_symrefs(&symrefs);
-		return -1;
-	}
```
pks-t: I actually find this commit a bit hard to digest, as it simultaneously moves around code and changes it at the same time. I'd welcome it if you split it up into multiple commits to make it easier to understand what is going on here

tests/describe/t6120.c
```diff
@@ -104,7 +104,7 @@ static void commit_and_tag(
 
 	cl_git_append2file("describe/file", "\n");
 	
-	cl_git_pass(git_index_add_bypath(index, "describe/file"));
+	cl_git_pass(git_index_add_bypath(index, "file"));
```
pks-t: I'd love to hear some details why this is necessary. Tests passed before, so why is it needed now?

tiennou: The path given to `git_index_add_bypath` is relative to the root of the repository. That `describe/file` path is relative to the root of the sandbox directory, hence if I add the missing `cl_git_pass` I rightfully get an error that `$SANDBOX/describe/describe/file doesn't exist`. The path is thus changed to be made relative to the repository, which makes the failure go away and "restore" the test.

Hence the test doesn't work as expected, because it does not care about `file` being added or not to the index, which makes dubious in my eyes.

src/transports/smart.c
tiennou: Sorry, this seemed a simple fix, then memory-management happened. I'll try to split (the crux of the change is that `git_smart__detect_caps` returns `GIT_ENOTFOUND` if it stumble on a weird pkt, which allows its caller to know if the `symrefs` it got are valid or not.

src/blame.c
tiennou: Ignore the error even though I (the user) has asked for it ? I do not know if that's explicit, but I find this strange… How can I tell if the mailmap was applied when that happens ?

src/transports/smart.c
pks-t: Thanks :)

tests/describe/t6120.c
tiennou: I'll put it in the commit message then 😉. I still feel like there's something dubious going on though, but this is about coverity.

src/transports/smart.c
tiennou: Hopefully the commits are more separated now, which clarifies what is happening. Note that I erroneously said `symref` above, but my understanding of the issue is that it might (maliciously) happen that the remote sent back one ref (refs is a vector of git_pkt in this code 😮) but doesn't follow through, so this `strcmp` will segfault on a NULL.

Thinking back a bit more, I don't know if that's plausible protocol-wise, but at least coverity should be pleased…

src/transports/smart.c
```diff
@@ -279,30 +279,31 @@ static int git_smart__connect(
 		return error;
 
 	/* Detect capabilities */
-	if (git_smart__detect_caps(first, &t->caps, &symrefs) < 0) {
-		free_symrefs(&symrefs);
-		return -1;
+	if ((error = git_smart__detect_caps(first, &t->caps, &symrefs)) == 0) {
```
pks-t: This change is wrong. We're new erroring out when the function ran successfully. It's getting fixed by the next commit, though

src/transports/smart_protocol.c
```diff
@@ -142,7 +142,7 @@ int git_smart__detect_caps(git_pkt_ref *pkt, transport_smart_caps *caps, git_vec
 
 	/* No refs or capabilites, odd but not a problem */
 	if (pkt == NULL || pkt->capabilities == NULL)
-		return 0;
+		return GIT_ENOTFOUND;
```
pks-t: This is fine, as there is only one caller of `git_smart__detect_caps`, which is the one you're adjusting in the same commit

pks-t:
Thank you for your fixes, @tiennou!

PR #4703, opened 2018-06-25
tree: reset spurious errors while walking entries

### LINKED ISSUE #4693: Misleading `git_tree_walk` errors
When using `git_tree_walk` with a callback which stops the walk it can appear that a previously generated libgit2 error was created in the callback. To reproduce:

1. Use `giterr_set_str` to create an error which does not use the callback error class. This emulates an error created earlier. 
2. Execute [`git_tree_walk`](https://libgit2.org/#HEAD/group/tree/git_tree_walk) where the provided callback stops the walk (returns a negative value)
3. The error returned will be the error set in step 1 and not from [`giterr_set_after_callback_function`](
https://github.com/libgit2/libgit2/blob/26f5d36d2f14dc1d711ed0a2c844ef4d7887a9b3/src/common.h#L115-L133)

tiennou:
Fixes #4693.

Note that I choose to be paranoid here, in case the callback sets an error but doesn't report it. Maybe it's too paranoid 😉.

### DIFF
--- a/src/tree.c
+++ b/src/tree.c
@@ -947,6 +947,7 @@ static int tree_walk(
 
 	git_array_foreach(tree->entries, i, entry) {
 		if (preorder) {
+			giterr_clear();
 			error = callback(path->ptr, entry, payload);
 			if (error < 0) { /* negative value stops iteration */
 				giterr_set_after_callback_function(error, "git_tree_walk");
@@ -983,6 +984,7 @@ static int tree_walk(
 		}
 
 		if (!preorder) {
+			giterr_clear();
 			error = callback(path->ptr, entry, payload);
 			if (error < 0) { /* negative value stops iteration */
 				giterr_set_after_callback_function(error, "git_tree_walk");


ethomson:
Ehhhhhhhhhh...  I guess?  This feels like a bandaid around poor usage.  🤔 

If we're going to do this (and I'm not sure that we should) then we probably need to do it across all the callbacks.  But this feels like a usage error to me, and we should probably document this better because even if we do take a pass across all the callbacks, we'll probably miss something or forget in the future.

ethomson:
To note - I know I sound unexcited about this (and I am not really very excited) but I don't think that there's probably any _harm_ in doing this, either.

tiennou:
AFAIU, the issue is that the error state from a previous can leak through if you happen to return an error from the cb without calling `giterr_set_str`. Something like "lookup a missing object", followed by "walk a tree, return error" would cause it to return the error state from the missing object.

The less paranoid way would be to only do it when entering the function, not on each loop. And I've seen some callback functions that clear on entry, but yeah, that's something that should be checked globally.

ethomson:
> AFAIU, the issue is that the error state from a previous can leak through if you happen to return an error from the cb without calling giterr_set_str.

Yes, that's exactly right.  My point was that if you return an error and _don't_ call `giterr_set_str`, but then you go _asking_ for the error string...  then why would you be surprised by what you find?  😁 

My thinking is: if I wanted to return an error from a callback, and I wanted to do something special with it, then I would return `-42` (or whatever, some sentinel value that libgit2 itself would never return) and if that got returned, then I would simply not query the error string.  And if libgit2 _itself_ failed, then I _would_ query the error string.

That's how we've designed the callback API - we propagate your error return code so that _you_ can do something with it.  And if you just want to return `-1`, that's fine, too (but you should probably clear the error string yourself).

Returning `-1` and then treating that like you weren't the one who returned it?  Seems odd.  Now maybe there's some FFI interface that Makes This Hard but I'm not familiar with that, and it seems like if you can set up function pointers for callbacks then you should be able to return the full range of `int`s.

I realize that `giterr_clear()` is pretty trivial (it will hopefully only set the length to zero - I don't have it in front of me but maybe it sets string[0] = '\0' as well) but I hate adding any more code to something like tree walk that will get called in a loop like this when this feels like a caller error.

omus:
> Yes, that's exactly right. My point was that if you return an error and _don't_ call `giterr_set_str`, but then you go _asking_ for the error string... then why would you be surprised by what you find?

The problem I'm having is I am not returning an error. As per the [`git_tree_walk`](https://libgit2.org/#HEAD/group/tree/git_tree_walk) documentation if a callback returns a negative value this triggers the walk to stop. In this case I don't have an error to report so I'm not setting one. Looking at the code it seems to me this is a supported usage and if no error is set in the callback the  [`giterr_set_after_callback_function`](https://github.com/libgit2/libgit2/blob/26f5d36d2f14dc1d711ed0a2c844ef4d7887a9b3/src/common.h#L123-L133) is [called](https://github.com/libgit2/libgit2/blob/967da2c71c5eb3ff1292ac3f479757ad51a21298/src/tree.c#L952) to set a default `GITERR_CALLBACK` class error.

This usage scenario breaks however if a non-`GITERR_CALLBACK` class error occurs before. The simple two solutions I can see to this problem is to clear any previous error message (this PR) or document that the callbacks are required to use `giterr_set_str` if a callback is returning a negative value.

ethomson:
> The problem I'm having is I am not returning an error. As per the git_tree_walk documentation if a callback returns a negative value this triggers the walk to stop. In this case I don't have an error to report so I'm not setting one. Looking at the code it seems to me this is a supported usage ...

Yes, this is all correct - and it's definitely a supported usage.  My question is:  why are you looking at the error message at all if you know that you're intentionally stopping the callback and you don't have an error?

omus:
We're looking at the error message since `git_tree_walk` is returning a negative value which is being interpreted as an error code. Maybe there is a better way to handle this in generic code? See this Julia code for the actual usage:

https://github.com/JuliaLang/julia/blob/b610f42249ba6435072919f03b4df343acc9bfed/stdlib/LibGit2/src/tree.jl#L31-L40

ethomson:
What we would encourage you to do is to return something totally arbitrary that is not a libgit2 error code, like `INT_MIN`.  And then do:

    if (err == INT_MIN)
        # i know this isn't actually an error
    elseif (err < 0)
        # handle the error

Many of our other callback-taking APIs actually allow the callback to return any non-zero code that will stop the execution flow and get propagated back to the caller.  This makes it a little more obvious that it's not actually an error at all, but a sentinel non-error value that you, the caller, have configured.  It also future-proofs your usage, since libgit2 returns negative values on error, if you used a negative value as your sentinel then it might someday be a libgit2 error value.

We started doing this a while back, but `git_tree_walk` appears to not have been updated.  We should do that, IMHO.

As for libgit2, my preference would be to:
1. Update `git_tree_walk` to allow callbacks to return values > 0 and return them to the caller.
2. Update the documentation to reflect that callback users should set an error message (perhaps to an empty string).

Do you think that this is reasonable?  We could still include this change as a third step - I realize that `giterr_clear` isn't doing much, but adding more work in a tight loop like `tree_walk` is something I'd rather avoid.

neithernut:
> What we would encourage you to do is to return something totally arbitrary that is not a libgit2 error code, like INT_MIN.

There's already a value for that: `GIT_EUSER`
https://github.com/libgit2/libgit2/blob/967da2c71c5eb3ff1292ac3f479757ad51a21298/include/git2/errors.h#L34

ethomson:
lol, thanks @neithernut, I had completely forgotten about that.  😁

omus:
I think specifying that `git_tree_walk` should return `GIT_EUSER` is reasonable. The [documentation](https://libgit2.org/#HEAD/group/tree/git_tree_walk) should then read something like

> If the callback returns a positive value, the passed entry will be skipped on the traversal (in pre mode). A negative value will stop the walk and typical indicates an error. Use the negative value `GIT_EUSER` indicate traversal has stopped without issue.

ethomson:
I opened #4783 so that we can better document this - in all our callbacks, not just the tree walk callback.

PR #4704, opened 2018-06-26
Remove GIT_PKT_PACK entirely

nelhage:
This is a heavier-handed alternative to https://github.com/libgit2/libgit2/pull/4694

`git_pkt_pack`/`GIT_PKT_PACK` are, as best I can tell, never used by any client of the pack APIs. Futhermore, they are incredibly dangerous; Because https://github.com/libgit2/libgit2/blob/master/src/transports/smart_pkt.c#L419-L423 does not advance `out`, any client is *required* to check for `GIT_PKT_PACK` and bail out, or else go into an infinite loop calling `git_pkt_parse_line` to get a new packet, which will from that point forward always return a `GIT_PKT_PACK`.

My fuzzer has found at least two callers that can be forced into an infinite loop via a misplaced `PACK`, and I suspect there are more.


ethomson:
This seems reasonable to me, but getting 👀 on it by @carlosmn would be nice.

src/transports/smart_pkt.c
```diff
@@ -407,17 +394,7 @@ int git_pkt_parse_line(
...
-		 * server is trying to send us the packfile already.
-		 */
-		if (bufflen >= 4 && !git__prefixcmp(line, "PACK")) {
-			giterr_clear();
-			*out = line;
-			return pack_pkt(head);
-		}
-
-		return (int)len;
+		return GIT_ERROR;
```
ethomson: We should set an error message here, perhaps:

    giterr_set(GITERR_NET, "unexpected pack file");
    return -1;

(We usually do `return -1` instead of `return GIT_ERROR` out of convention.  I'd like to keep that moving forward.)

pks-t:
Looks good to me. @ethomson, any further comments on this?

src/transports/smart.h
```diff
@@ -40,7 +40,7 @@ typedef enum {
 	GIT_PKT_HAVE,
 	GIT_PKT_ACK,
 	GIT_PKT_NAK,
-	GIT_PKT_PACK,
+	GIT_PKT_PACK__UNUSED,
```
carlosmn: This is a private struct, we don't need place-holders for deleted values.

nelhage:
@carlosmn Thanks, removed the placeholder.

pks-t:
Thanks @nelhage!

Issue #1980, opened 2013-12-03
Add support for apply-patch


ghost:
If libgit2 adds support for git's builtin/apply.c, we would be able to use libgit2 (actually I'm more interested in rugged) to apply patches received by email to a git repository.

GrahamTheCoder:
+1

epixa:
+1

For others looking, it looks like https://github.com/libgit2/libgit2/pull/3223 is the latest attempt to get this started.

brandonio21:
Looks like the corresponding PR was closed and this feature is now in master. To close out this task, perhaps add it to the documentation? (I cannot find `git_patch_apply` using the online documentation search function)

ethomson:
#3223 only adds support for reading a patch files into the `git_diff` representation.   There's not yet an API to take a workdir/index/etc and modify it based on that in-memory representation.

tiennou:
Support for this has landed as part of #4705, so this is fixed. Feel free to open issues if you find missing functionality though.

PR #4705, opened 2018-06-26
Patch (diff) application

ethomson:
This pull request takes a `git_diff` that was either "generated" (ie, we calculated the difference between two commitish things) or "parsed" (we opened the output of `git diff` or a file that we created ourselves) and apply it in the same manner that `git apply` does.

In addition, it introduces `git_apply_tree` which will perform a fully in-memory patch application, applying a `git_diff` to a `git_tree` and returning a `git_index`.

My intent is to get this to a minimal viable product that we can merge and then continue to iterate on.  I think that this is a useful API even if we don't support all the options that `git apply` itself does (eg, writing rejection hunks, etc).  However, there are still a few things outstanding before I consider this mergeable:

- [x] Tests for `GIT_APPLY_LOCATION_INDEX`
- [x] Tests for `GIT_APPLY_LOCATION_BOTH`
- [x] Whole-patch callbacks?
- [x] Hunk callbacks (eg, "partial patch application") a la #4184


ethomson:
So as of https://github.com/libgit2/libgit2/pull/4705/commits/20acd2a7cd361a75e7bd6cd3ffea0687391697d3, I'm pretty happy with where this is.  It's not complete yet, or ready for merge -- I do want to add callbacks, at a minimum.  But I do think that this is ready for a first round of review - the patch application is in good shape and the rest of the changes are additive.

If you're the type of person that likes to read commits in order then I've tried to make this a reasonable read.

src/reader.h
```diff
@@ -0,0 +1,42 @@
...
+#include "common.h"
+
+typedef struct git_reader git_reader;
+
+struct git_reader {
+	int (*read)(git_buf *out, git_reader *reader, const char *filename);
+	void (*free)(git_reader *reader);
+};
+
+extern int git_reader_for_nothing(
```
tiennou: Curious: any reason you're using `extern` directly ? I thought `GIT_EXTERN` was mandatory because of MSVC compatibility…

src/apply.c
```diff
@@ -480,3 +483,110 @@ int git_apply_tree(
...
+	size_t i;
+	int error;
+
+	assert(repo && diff);
+
+	GITERR_CHECK_VERSION(
+		given_opts, GIT_APPLY_OPTIONS_VERSION, "git_apply_options");
+
+	if (given_opts)
+		memcpy(&opts, given_opts, sizeof(git_apply_options));
```
tiennou: (FWIW I remember doing it the other way around : `if (given_opts == NULL) *given_opts = opts;`. Saves a memcpy 🙄)

src/apply.c
```diff
@@ -480,3 +483,110 @@ int git_apply_tree(
...
+	for (i = 0; i < git_diff_num_deltas(diff); i++) {
+		delta = git_diff_get_delta(diff, i);
+
+		if ((error = apply_one(repo, pre_reader, contents, diff, i)) < 0)
+			goto done;
+
+		if (do_checkout) {
+			git_vector_insert(&paths, (void *)delta->old_file.path);
+
+			if (strcmp(delta->old_file.path, delta->new_file.path))
```
tiennou: I'm slightly concerned that those `{old,new}_file` members could be `NULL` ?

tests/apply/workdir.c
```diff
@@ -0,0 +1,124 @@
...
+
+	cl_git_pass(git_commit_lookup(&a_commit, repo, &a_oid));
+	cl_git_pass(git_commit_lookup(&b_commit, repo, &b_oid));
+
+	cl_git_pass(git_commit_tree(&a_tree, a_commit));
+	cl_git_pass(git_commit_tree(&b_tree, b_commit));
+
+	cl_git_pass(git_diff_tree_to_tree(&diff, repo, a_tree, b_tree, &opts));
+
+    cl_git_pass(git_reset(repo, (git_object *)a_commit, GIT_RESET_HARD, NULL));
```
tiennou: Nitpick: indent.

tests/apply/workdir.c
```diff
@@ -130,7 +183,6 @@ void test_apply_workdir__generated_diff(void)
 
 	cl_git_pass(git_diff_tree_to_tree(&diff, repo, a_tree, b_tree, &opts));
 
-    cl_git_pass(git_reset(repo, (git_object *)a_commit, GIT_RESET_HARD, NULL));
```
tiennou: Nevermind then 😉.

src/apply.c
```diff
@@ -24,7 +24,7 @@
 #include "reader.h"
 
 #define apply_err(...) \
-	( giterr_set(GITERR_PATCH, __VA_ARGS__), -1 )
+	( giterr_set(GITERR_PATCH, __VA_ARGS__), GIT_EAPPLYFAIL )
```
tiennou: That's debatable, but I'd rather add a `GITERR_APPLY` error class (so you can mix-and-match with error codes, eg `GIT_EINVALID` in this case) instead of adding a new error code ?

src/reader.c
```diff
@@ -0,0 +1,248 @@
...
+		if ((error = git_odb_hash(&id, out->ptr, out->size, GIT_OBJ_BLOB)) < 0)
+			goto done;
+	}
+
+	if (reader->index) {
+		if (!(idx_entry = git_index_get_bypath(reader->index, filename, 0)) ||
+		    !git_oid_equal(&id, &idx_entry->id)) {
+			error = GIT_READER_MISMATCH;
+			goto done;
+		}
```
pks-t: Ah, you need the cleaned file to verify it's got the same content as the index. I still wonder about whether it's the right thing to do in the context of git-apply(1)

src/reader.h
```diff
@@ -0,0 +1,47 @@
...
+
+typedef struct git_reader git_reader;
+
+struct git_reader {
+	int (*read)(git_buf *out, git_oid *out_oid, git_reader *reader, const char *filename);
+	void (*free)(git_reader *reader);
+};
+
+extern int git_reader_for_nothing(
+	git_reader **out);
```
pks-t: This does not have an implementation. Probably a leftover?

src/reader.h
```diff
@@ -0,0 +1,47 @@
...
+extern int git_reader_for_workdir(
+	git_reader **out,
+	git_repository *repo,
+	bool validate_index);
+
+extern int git_reader_read(
+	git_buf *out,
+	git_oid *out_id,
+	git_reader *reader,
+	const char *filename);
```
pks-t: Some documentation around optional parameters and `validate_index` etc. would be nice to have ;)

src/apply.c
```diff
@@ -376,3 +383,316 @@ int git_apply__patch(
...
+
+		if (preimage) {
+			memset(&pre_entry, 0, sizeof(git_index_entry));
+			pre_entry.path = delta->old_file.path;
+			pre_entry.mode = delta->old_file.mode;
+			git_oid_cpy(&pre_entry.id, &pre_id);
+
+			if ((error = git_index_add(preimage, &pre_entry)) < 0)
+				goto done;
+		}
```
pks-t: Just to make sure I understand: do you try to generate the preimage here, assuming that the passed `preimage` parameter is empty? I'm not a hundred percent sure I totally get that code

src/apply.c
```diff
@@ -376,3 +383,316 @@ int git_apply__patch(
...
+	/*
+	 * by default, we apply a patch directly to the working directory;
+	 * in `--cached` or `--index` mode, we apply to the contents already
+	 * in the index.
+	 */
+	if (opts.location == GIT_APPLY_LOCATION_BOTH)
+		error = git_reader_for_workdir(&pre_reader, repo, true);
+	else if (opts.location == GIT_APPLY_LOCATION_INDEX)
+		error = git_reader_for_index(&pre_reader, repo, NULL);
+	else
```
pks-t: I'd prefer if we used a switch here. Like that, the compiler would throw a warning in case we ever add another enum

src/apply.c
```diff
@@ -376,3 +383,316 @@ int git_apply__patch(
...
+
+	for (i = 0; i < git_diff_num_deltas(diff); i++) {
+		if ((error = apply_one(repo, pre_reader, preimage, postimage, diff, i)) < 0)
+			goto done;
+	}
+
+	if (opts.location == GIT_APPLY_LOCATION_INDEX)
+		error = git_apply__to_index(repo, diff, preimage, postimage, &opts);
+	else
+		error = git_apply__to_workdir(repo, diff, preimage, postimage, &opts);
```
pks-t: No need to handle `BOTH` here, I guess, because of the way we initialized the reader? I'd personally vote for a switch, as well

dometto:
@ethomson could you give an approximate timeline for this? Many thanks!

ethomson:
Further testing has revealed problems in the patch application, I don’t have an ETA.

src/reader.h
ethomson: `GIT_EXTERN` is what we use for things in the public api.  It will do all the decoration for DLL exports, as you mentioned.  For private functions within the library, we should simply use `extern`.

src/reader.c
```diff
@@ -0,0 +1,203 @@
...
+
+static void workdir_reader_free(git_reader *_reader)
+{
+	GIT_UNUSED(_reader);
+}
+
+int git_reader_for_workdir(git_reader **out, git_repository *repo)
+{
+	workdir_reader *reader;
+
```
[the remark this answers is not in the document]
ethomson: Fixed.

src/reader.c
```diff
@@ -0,0 +1,203 @@
...
+}
+
+int git_reader_for_index(
+	git_reader **out,
+	git_repository *repo,
+	git_index *index)
+{
+	index_reader *reader;
+	int error;
+
```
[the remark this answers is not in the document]
ethomson: Fixed.

src/reader.c
```diff
@@ -0,0 +1,248 @@
...
+	if ((error = git_tree_entry_bypath(&tree_entry,
+		reader->tree, filename)) < 0 ||
+		(error = git_blob_lookup(&blob,
+		git_tree_owner(reader->tree), git_tree_entry_id(tree_entry))) < 0)
+		goto done;
+
+	git_buf_clear(out);
+	git_buf_put(out, git_blob_rawcontent(blob), git_blob_rawsize(blob));
```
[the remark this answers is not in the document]
ethomson: Fixed.

src/reader.c
```diff
@@ -0,0 +1,248 @@
...
+{
+	tree_reader *reader = (tree_reader *)_reader;
+	git_tree_entry *tree_entry = NULL;
+	git_blob *blob = NULL;
+	int error;
+
+	if ((error = git_tree_entry_bypath(&tree_entry,
+		reader->tree, filename)) < 0 ||
+		(error = git_blob_lookup(&blob,
+		git_tree_owner(reader->tree), git_tree_entry_id(tree_entry))) < 0)
```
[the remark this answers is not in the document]
ethomson: Fixed.

ethomson:
I've added per-delta and per-hunk callbacks.  I'll go through the notes from the prior review (up to 
20acd2a) tomorrow

src/apply.c
ethomson: The nice thing about creating a copy is that you can [set or override options](https://github.com/libgit2/libgit2/blob/master/src/merge.c#L2066) [that the caller didn't provide](https://github.com/libgit2/libgit2/blob/e9b2e4343f636d0cdb030ffe6ec1edc76d344253/src/diff_tform.c#L246).  We don't do that here yet, but I suspect that we will want to over time.

src/apply.c
ethomson: I'm happy to change the error class here if you think that it would be useful, but I think that having an actual `GIT_EAPPLYFAIL` error code is useful.  I don't think that there's any other error code that suggests that patch application was impossible.  `GIT_ECONFLICT` is probably the _closest_ but it's not right.

src/reader.c
```diff
@@ -0,0 +1,248 @@
...
+int git_reader_for_tree(git_reader **out, git_tree *tree)
+{
+	tree_reader *reader;
+
+	reader = git__calloc(1, sizeof(tree_reader));
+	GITERR_CHECK_ALLOC(reader);
+
+	reader->reader.read = tree_reader_read;
+	reader->reader.free = tree_reader_free;
+	reader->tree = tree;
```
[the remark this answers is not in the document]
ethomson: I don't think that's necessary.  These readers really just provide a handy abstraction over querying them, and their lifecycle is never longer than the apply functions -- meaning that a caller couldn't try to free a tree or index in the middle of application (at least not on a single thread).

src/apply.c
ethomson: I'll add a comment, but what we're doing here is populating a data structure that contains the actual on-disk contents as we read them out of the preimage - the actual on-disk contents of the file that we want to apply a patch to.

This is because the patch may be applied to a dirty working directory copy, so we need to know the expected baseline when we do the checkout, so that it can validate that the file hasn't changed.

eg, if file A is at abc1 in `HEAD`, and then modified in the working directory, we're still allowed to try to apply a patch.  So we need to get the hash of the contents of the file that we actually read and applied the patch to.  We'll provide that to checkout as the baseline, so it will verify that although the file is different than `HEAD`, it's what we actually expect it to be.  Without this, we'll have a checkout conflict.

src/reader.c
```diff
@@ -0,0 +1,248 @@
...
+done:
+	git_blob_free(blob);
+	git_tree_entry_free(tree_entry);
+	return error;
+}
+
+static void tree_reader_free(git_reader *_reader)
+{
+	GIT_UNUSED(_reader);
+}
```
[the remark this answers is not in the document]
ethomson: It turns out that none of the readers provide a `free` implementation, so I'll remove it entirely.

src/reader.c
ethomson: I'm copying git's behavior here.  It runs the workdir -> odb filter as well.  This is tough to see with (for example) line ending changes, since line endings are already a mess and patch application tries to deal with them.

However, it's readily evident if you put a text file into LFS.  Here I'll add a simple text file to the repository and apply a patch:

```
> git add bar.txt

>git commit -m"bar"
[master (root-commit) ebc14bf] bar
 1 file changed, 16 insertions(+)
 create mode 100644 bar.txt

>type bar.txt
0. Hello, this file has CR/LF line endings.
1. Hello, this file has CR/LF line endings.
2. Hello, this file has CR/LF line endings.
3. Hello, this file has CR/LF line endings.
4. Hello, this file has CR/LF line endings.
5. Hello, this file has CR/LF line endings.
6. Hello, this file has CR/LF line endings.
7. Hello, this file has CR/LF line endings.
8. Hello, this file has CR/LF line endings.
9. Hello, this file has CR/LF line endings.
a. Hello, this file has CR/LF line endings.
b. Hello, this file has CR/LF line endings.
c. Hello, this file has CR/LF line endings.
d. Hello, this file has CR/LF line endings.
e. Hello, this file has CR/LF line endings.
f. Hello, this file has CR/LF line endings.

> git apply < patch

> type bar.txt
0. Hello, this file has CR/LF line endings.
1. Hello, this file has CR/LF line endings.
2. Hello, this file has CR/LF line endings.
3. Hello, this file has CR/LF line endings.
4. Here's a change on the fifth line.
5. Hello, this file has CR/LF line endings.
6. Hello, this file has CR/LF line endings.
7. Hello, this file has CR/LF line endings.
8. Hello, this file has CR/LF line endings.
9. Hello, this file has CR/LF line endings.
a. Hello, this file has CR/LF line endings.
b. Here's a change on another line.
c. Hello, this file has CR/LF line endings.
d. Hello, this file has CR/LF line endings.
e. Hello, this file has CR/LF line endings.
f. Hello, this file has CR/LF line endings.
```

Now let's add the file as an LFS resource and try to do the same thing:

```
> git reset --hard HEAD
HEAD is now at ebc14bf bar

> git lfs install
Updated git hooks.
Git LFS initialized.

> git lfs track bar.txt
Tracking "bar.txt"

> git add bar.txt

> git status
On branch master
Changes to be committed:

        modified:   bar.txt

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        .gitattributes
        patch

>git commit -mbar
[master 756794c] bar
 1 file changed, 3 insertions(+), 16 deletions(-)
 rewrite bar.txt (100%)

> git apply < patch
error: patch failed: bar.txt:2
error: bar.txt: patch does not apply
```

Which is, of course, because it's applying the filter.  Which suggests that we could actually patch the LFS pointer file.  Which, indeed, we can:

```
>git ls-files --stage
100644 976ef879085369af40ec959843624d6c5b0f5024 0       bar.txt

>git cat-file blob 976ef
version https://git-lfs.github.com/spec/v1
oid sha256:cad943fa9780d676a5236c9fe503d8f63b2115cf3786d0884bc8c07a84aa9d92
size 720

> type patch
diff --git a/bar.txt b/bar.txt
index 976ef87..d2c468a 100644
--- a/bar.txt
+++ b/bar.txt
@@ -1,3 +1,4 @@
 version https://git-lfs.github.com/spec/v1
 oid sha256:cad943fa9780d676a5236c9fe503d8f63b2115cf3786d0884bc8c07a84aa9d92
 size 720
+Yes, of course it's patching the LFS pointer file.

> git apply < patch
Encountered 1 file(s) that should have been pointers, but weren't:
        bar.txt

> type bar.txt
version https://git-lfs.github.com/spec/v1
oid sha256:cad943fa9780d676a5236c9fe503d8f63b2115cf3786d0884bc8c07a84aa9d92
size 720
Yes, of course it's patching the LFS pointer file.
```

Anyway.  I don't really _like_ this design decision.  It seems to me that you would want to patch the filtered content, not the ODB content.  I think that it was probably designed to deal with line ending, like oh so much else.  But whether I like this decision or not, like oh so many things, it's not really ours to make.

src/apply.c
ethomson: They cannot - generated diffs [always have `old_file.path`](https://github.com/libgit2/libgit2/blob/master/src/diff_generate.c#L47) and [`new file.path`](https://github.com/libgit2/libgit2/blob/master/src/diff_generate.c#L53) set.  And parsed patches [ensure that they're both set](https://github.com/libgit2/libgit2/blob/master/src/patch_parse.c#L972).

src/reader.c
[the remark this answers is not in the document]
ethomson: Removed.

src/reader.c
ethomson: Added a bit of documentation here, as well.

src/reader.h
ethomson: Removed.

src/reader.h
ethomson: Added some docs.

src/apply.c
ethomson: Added a comment to explain this.

src/apply.c
ethomson: Changed this to a switch.

ethomson:
@pks-t, @tiennou are you happy with these changes or did you want another 👀 ?

pks-t:
On Wed, Oct 17, 2018 at 06:33:19AM -0700, Edward Thomson wrote:
> @pks-t, @tiennou are you happy with these changes or did you want another 👀 ?

I didn't yet care to do another review as this is still labelled
as WIP ;) I can do that tomorrow, though

tests/apply/both.c
```diff
@@ -0,0 +1,384 @@
...
+void test_apply_both__index_must_match_workdir(void)
+{
+	git_diff *diff;
+	git_index *index;
+	git_index_entry idx_entry;
+
+	const char *diff_file = DIFF_MODIFY_TWO_FILES;
+
+	/*
+	 * Append a line to the end of the file in each the index and the
```
tiennou: `s/each/both/` ?

ethomson:
I'm trying to accept your changes, and manually updating the commit message to be something that `git rebase -i` will parse as a fixup.  (eg, the moral equivalent of `git commit --fixup <id>`.  We'll see how this works...  😀

src/apply.c
```diff
@@ -376,3 +417,364 @@ int git_apply__patch(
...
+	git_vector_free(&paths);
+	return error;
+}
+
+static int git_apply__to_index(
+    git_repository *repo,
+    git_diff *diff,
+    git_index *preimage,
+    git_index *postimage,
+    git_apply_options *opts)
```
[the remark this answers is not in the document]
ethomson: Not seeing the indentation problems here, this looks like it's using spaces everywhere, per the new (or newly reconfirmed) standards.

src/apply.c
```diff
@@ -376,3 +417,364 @@ int git_apply__patch(
...
+
+	return error;
+}
+
+static int git_apply__to_workdir(
+    git_repository *repo,
+    git_diff *diff,
+    git_index *preimage,
+    git_index *postimage,
+	git_apply_location_t location,
```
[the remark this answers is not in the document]
ethomson: Fixed

src/apply.c
```diff
@@ -376,3 +417,364 @@ int git_apply__patch(
...
+ * entirely.
+ *
+ * GIT_APPLY_LOCATION_INDEX: emulates `git apply --cached`.
+ * Applies the diff only to the index items and ignores the workdir
+ * completely.
+ *
+ * GIT_APPLY_LOCATION_BOTH: emulates `git apply --index`.
+ * Applies the diff to both the index items and the working directory
+ * items.
+ */
```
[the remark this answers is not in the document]
ethomson: I found the differences between `git apply --index` and `git apply --cached` (incredibly) confusing, so I found this reminder in the file helpful.  Unless you object strongly, I'd like to keep it to keep my small brain focused the next time I'm in this file.  😉

tests/apply/both.c
ethomson: Fixed.

src/reader.c
```diff
@@ -0,0 +1,239 @@
...
+	git_reader *_reader,
+	const char *filename)
+{
+	tree_reader *reader = (tree_reader *)_reader;
+	git_tree_entry *tree_entry = NULL;
+	git_blob *blob = NULL;
+	int error;
+
+	if ((error = git_tree_entry_bypath(&tree_entry,
+	    reader->tree, filename)) < 0 ||
```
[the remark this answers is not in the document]
ethomson: Fixed.

src/apply.c
```diff
@@ -160,15 +167,36 @@ static int update_hunk(
...
 {
 	patch_image preimage = PATCH_IMAGE_INIT, postimage = PATCH_IMAGE_INIT;
 	size_t line_num, i;
 	int error = 0;
 
+	if (ctx->opts.hunk_cb) {
+		error = ctx->opts.hunk_cb(&hunk->hunk, ctx->opts.payload);
+
+		if (error) {
+			if (error > 0) {
```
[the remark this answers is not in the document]
ethomson: Why does `GIT_EUSER` come to mind?

src/apply.c
```diff
@@ -376,3 +417,364 @@ int git_apply__patch(
...
+	if ((error = git_patch_from_diff(&patch, diff, i)) < 0)
+		goto done;
+
+	delta = git_diff_get_delta(diff, i);
+
+	if (opts->delta_cb) {
+		error = opts->delta_cb(delta, opts->payload);
+
+		if (error) {
+			if (error > 0)
```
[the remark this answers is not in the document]
ethomson: I wrote up your suggestion, but I found it less readable than what's here now, tbqh.  @pks-t can you break a tie?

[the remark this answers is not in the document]
ethomson: A:
```
if (error > 0) {
    error = 0;
    goto done;
} else if (error < 0) {
    goto done;
}
```

B:
```
if (error) {
    if (error > 0)
        error = 0;

    goto done;
}
```

src/reader.c
```diff
@@ -0,0 +1,239 @@
...
+	assert(out && repo);
+
+	reader = git__calloc(1, sizeof(index_reader));
+	GITERR_CHECK_ALLOC(reader);
+
+	reader->reader.read = index_reader_read;
+	reader->repo = repo;
+
+	if (index) {
+		reader->index = index;
```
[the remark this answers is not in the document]
ethomson: I'm not sure if that's really necessary or not - that's not something we do in other places that take a repo and an optional index.  But you're right that we don't take an explicit index anywhere, so I've just gone ahead and make the `reader_for_index` API always take the repo and get the index from that.  (I did it that way to ensure that it's not a bare repository.)

src/apply.c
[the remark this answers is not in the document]
tiennou: 👍 then.

ethomson:
Fixes #1980

src/apply.c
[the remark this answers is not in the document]
tiennou: I might have mistook that for `GIT_PASSTHROUGH`, sorry.

src/apply.c
```diff
@@ -376,3 +417,364 @@ int git_apply__patch(
...
+		error = git_reader_for_workdir(&pre_reader, repo, true);
+		break;
+	case GIT_APPLY_LOCATION_INDEX:
+		error = git_reader_for_index(&pre_reader, repo);
+		break;
+	case GIT_APPLY_LOCATION_WORKDIR:
+		error = git_reader_for_workdir(&pre_reader, repo, false);
+		break;
+	default:
+		abort();
```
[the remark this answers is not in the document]
ethomson: Hmm.  I meant to write an `assert(0)` there.

src/apply.c
```diff
@@ -376,3 +417,364 @@ int git_apply__patch(
...
+	if (given_opts)
+		memcpy(&opts, given_opts, sizeof(git_apply_options));
+
+	/*
+	 * by default, we apply a patch directly to the working directory;
+	 * in `--cached` or `--index` mode, we apply to the contents already
+	 * in the index.
+	 */
+	switch (location) {
+	case GIT_APPLY_LOCATION_BOTH:
```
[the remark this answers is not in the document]
ethomson: No, I don’t think that there is.  I don’t think that there will ever be additional modes without needing to create a `v2` of this function.  That said, I supposed we could make workdir = 1, index = 2 so that both is the logical addition of them.  (But I don’t think that it’s necessary to do this either.)

src/apply.c
```diff
@@ -376,3 +417,364 @@ int git_apply__patch(
...
+		}
+	}
+
+	if (delta->status != GIT_DELTA_ADDED) {
+		error = git_reader_read(&pre_contents, &pre_id,
+			preimage_reader, delta->old_file.path);
+
+		/* ENOTFOUND means the preimage was not found; apply failed. */
+		if (error == GIT_ENOTFOUND)
+			error = GIT_EAPPLYFAIL;
```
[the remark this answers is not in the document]
ethomson: I do overwrite `GIT_ENOTFOUND`, but I don't overwrite the error message itself, which would have been set by the reader.  This means that it's actually more explicit, since the existing error message distinguishes between the index and the working directory.  eg, if it's not found in the index, it will say [`index does not contain 'foo'`](https://github.com/libgit2/libgit2/blob/master/src/index.c#L867).

include/git2/apply.h
```diff
@@ -0,0 +1,129 @@
...
+ * Apply a `git_diff` to a `git_tree`, and return the resulting image
+ * as an index.
+ *
+ * @param out the postimage of the application
+ * @param repo the repository to apply
+ * @param preimage the tree to apply the diff to
+ * @param diff the diff to apply
+ * @param options the options for the apply (or null for defaults)
+ */
+GIT_EXTERN(int) git_apply_tree(
```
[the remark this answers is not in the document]
ethomson: Yes, I agree.  Fixed.

src/apply.c
```diff
@@ -376,3 +417,364 @@ int git_apply__patch(
...
+			pre_entry.path = delta->old_file.path;
+			pre_entry.mode = delta->old_file.mode;
+			git_oid_cpy(&pre_entry.id, &pre_id);
+
+			if ((error = git_index_add(preimage, &pre_entry)) < 0)
+				goto done;
+		}
+	}
+
+	if (delta->status == GIT_DELTA_DELETED)
```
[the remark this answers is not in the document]
ethomson: Indeed, I do like that better, it matches nicely the condition and formatting above it.

tests/apply/both.c
```diff
@@ -0,0 +1,384 @@
...
+	git_commit_free(commit);
+
+	cl_git_pass(git_diff_from_buffer(&diff, diff_file, strlen(diff_file)));
+	cl_git_pass(git_apply(repo, diff, GIT_APPLY_LOCATION_BOTH, NULL));
+
+	validate_apply_index(repo, index_expected, index_expected_cnt);
+	validate_apply_workdir(repo, workdir_expected, workdir_expected_cnt);
+
+	git_diff_free(diff);
+}
```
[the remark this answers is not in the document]
ethomson: I'll add some additional tests.  When you say two-to-one, you mean a mangled patch that has two sources being renamed to the same target?  And by one-to-two, another mangled patch, this one with a single source being renamed to two different targets?  These seem like reasonable tests.

src/apply.c
```diff
@@ -376,3 +417,364 @@ int git_apply__patch(
...
+		delta = git_diff_get_delta(diff, i);
+
+		if ((error = git_index_remove(postimage,
+				delta->old_file.path, 0)) < 0)
+			goto done;
+	}
+
+	for (i = 0; i < git_diff_num_deltas(diff); i++) {
+		if ((error = apply_one(repo, pre_reader, NULL, postimage, diff, i, &opts)) < 0)
+			goto done;
```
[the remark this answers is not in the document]
ethomson: Ick.  But ok.

ethomson:
Thanks for the feedback.  I've rebased to squash my fixup commits.  I'll address @pks-t's requests for some more exotic tests (and any codechanges that those may require) today.

ethomson:
> BTW, I'd be interested in some kind of "report" on the usage of suggestions. They're quite easy to make, but I'd like to be sure that they're not a pain to re-fixup afterward (it sounds like it it).

@tiennou, I thought it was reasonably useful and easy enough to `rebase -i --autosquash` with them (though I had to get the commit message correct in order to support that workflow).  It would be nice if GitHub had an option to do that for me.

My brief tests with suggestions are positive where I'm only changing one-line.  I wish I could produce a more exhaustive suggestion.  Alas.

ethomson:
Okay, I've also added support for applying multiple different deltas (like your example, @pks-t).  The only "tricky" bit here is if you have one delta that renames a file from A -> B and then another delta that applies to B.  In order to make multiple rename scenarios work (eg, rename A -> B and B -> C or rename A -> B and B -> A), git requires that rename deltas renaming a file come first.  eg, this file is applyable:

```patch
diff --git a/foo b/bar
similarity index 81%
rename from foo
rename to bar
index e53eaa1..9b1e563 100644
--- a/foo
+++ b/bar
@@ -1,4 +1,4 @@
-1
+one
 2
 3
 4
diff --git a/bar b/bar
index e53eaa1..32813be 100644
--- a/bar
+++ b/bar
@@ -7,4 +7,4 @@
 7
 8
 9
-0
+zero
```

But reversing this, such that the modification to `bar` is first, is not applyable.  git fails with:

```
error: bar: does not exist in index
```

Modifying the patch so that the modification occurs to `foo` before the rename also fails, interestingly with the same error message, which is a bit confusing given the patch in question:

```
error: bar: does not exist in index
```

And trying to perform a modification to `foo` after it has been renamed fails with:

```
error: path foo has been renamed/deleted
```

ethomson:
I went through some more tests - in particular, the idea that you couldn't modify a file in a delta and then rename it in a second delta bothered me.  I realized that my test was malformed (I was trying to rename a file to a different, already existing file, hence the error message).  Recreating the test validates that git does indeed let you update some file named `foo` in one delta and then rename it in another.

I also added validation that we will not apply modification deltas to files after deleting them, and added several tests around strange behavior with multiple deltas.

ethomson:
Thanks everybody for the reviews!  🚢

pks-t:
Exciting times! Thanks a lot, @ethomson!

PR #4710, opened 2018-06-29
streams: report OpenSSL errors if global init fails

pks-t:
In case when the global initialization of the OpenSSL stream fails, the
user is left without any hint as to what went wrong as we do not provide
any error message at all. This commit refactors the init function to
have a common error path, which now also sets an error message including
the error string provided by OpenSSL.


pks-t:
Oh. Embarassing syntax error. I shouldn't push an untested state :sweat_smile:

ethomson:
Thanks!  👍

PR #4712, opened 2018-07-03
Submodule example and initialization status

pks-t:
I've recently been asked for an example for submodules, which is why I decided to just write one. During the process, I noticed that either our API lacks quite a lot of things or I was just too stupid to figure them out. Either way, in both cases the API would be bad.

- The most important pain point is probably adding a submodule with a checkout. Just take a look at `submodule_add` in the example, it's gross.
- The second thing that got me was that there is simply no way to determine whether a submodule is initialized or not (in the `git submodule init` sense). I've addressed the second point, because there simply has to be a way. I'm not too sure about the interface (see the last paragraph in the commit's message), a new function `git_submodule_is_initialized` may be cleaner with less potential problems. But I wanted to stick with the existing `git_submodule_status` interface for now. Your opinion would be appreciated.


pks-t:
Oh, yeah. We also have no way to deinitialize submodules.

tiennou:
(Related: #4637, which makes that test code into a public function)

pks-t:
Oops, sorry @tiennou, I didn't remember that PR. I'll have a look at the next occasion

tiennou:
I'm just pointing out that I had the same "reasoning" (there's a *lot* of things to get right between `_add` and `_finalize` with no obvious simple way) while working on the bare submodule bug. No swiftness required 😉.

pks-t:
This PR should wait until #4637 is merged to remove this atrocious thing that the initial submodule cloning is. Even though #4637 is running into the exact same problem now, which just didn't surface because of another error in that code :P

pks-t:
Rebased to adapt to our new examples infra. Still waiting on #4637 to land to fix the mess that `submodule_add` is right now

pks-t:
Rebased to make use of the now-merged #4637.

personalizedrefrigerator:
It would be nice to have this!

Issue #4107, opened 2017-02-03
Cloning a repo on Windows breaks symlinks


bquistorff:
I can't clone faithfully repos with symlinks on Windows. I have been able to do this using Git for Windows ([info](https://github.com/git-for-windows/git/wiki/Symbolic-Links)) when in Administrator mode and setting `core.symlinks=true`.

### Reproduction steps
I am using `libgit2` through R with the `git2r` package.
Start R in administrator mode
```
> git2r::config()
programdata:
        color.branch=auto
        color.diff=auto
        color.interactive=true
        color.status=auto
        core.autocrlf=true
        core.fscache=true
        core.symlinks=true
        diff.astextplain.textconv=astextplain
        help.format=html
        http.sslcainfo=C:/Program Files/Git/mingw64/ssl/certs/ca-bundle.crt
        rebase.autosquash=true
global:
        core.symlinks=true
        filter.lfs.clean=git-lfs clean -- %f
        filter.lfs.process=git-lfs filter-process
        filter.lfs.required=true
        filter.lfs.smudge=git-lfs smudge -- %f
> git2r::clone("git://github.com/imbs-hl/ranger.git", "C:/Users/USER/Documents/GitHub/ranger2", progress = FALSE)
Local:    master C:/Users/USER/Documents/GitHub/ranger2/
Remote:   master @ origin (git://github.com/imbs-hl/ranger.git)
Head:     [47eb75a] 2017-01-20: Merge pull request #157 from imbs-hl/fix_extratrees
```

### Expected behavior
Expect many symlink files in `ranger-r-package/ranger/src` folder.

### Actual behavior
Many text files in `ranger-r-package/ranger/src` folder where file contents are the original symlink target paths.

### Version of libgit2 (release number or SHA1)
git2r version 0.18.0.9000 (latest release) which say that it has libgit2 commit 6b0510e from 20 December 2016

### Operating system(s) tested
Windows 10 Enterprise

ethomson:
I truly dislike the `core.symlinks` option, but I suppose it's something that we should indeed support.  (To be honest, though, this isn't really on my radar at the moment.)

csware:
I just want to add that symlinks on Windows require administrative privileges.

jabbera:
@csware  the admin requirement is no longer needed for windows 10 creators.

cyberbit:
@jabbera But only if the user has Developer Mode enabled.

rogersachan:
[...]

PR #4713, opened 2018-07-03
Support symlinks on Windows when core.symlinks=true

ethomson:
Windows 10 enables non-Administrator users to create symbolic links when Developer Mode is enabled.  Git for Windows supports this scenario when `core.symlinks=true`.

Similarly, we should support creation of proper symbolic links on Windows when `core.symlinks=true`.  This is as simple as implementing `p_symlink` properly, as checkout already behaves properly, examining `core.symlinks` and creating a fake symlink when `core.symlinks=false` (and creating a true symbolic link otherwise).

This is not ready for merge:  at present, Git for Windows requires a configuration setting `core.symlinks` in a global configuration area to enable automatic detection of symbolic links during repository initialization.

This probably makes sense: if a repository is created by Administrator - who has symbolic link creation privileges - you may still want to set `core.symlinks=false` on Windows since there is no suggestion that non-Administrator users will be able to use symbolic links.

However, there is some thought of changing this in Git for Windows (https://github.com/git-for-windows/git/pull/1744).  So let's wait for that to be decided - if they change their behavior, then this will be ready for merge.  If not, we will need to modify this PR to adopt their existing behavior.


tests/checkout/index.c
```diff
@@ -175,35 +193,45 @@ void test_checkout_index__honor_coresymlinks_default(void)
...
-		char link_data[1024];
-		size_t link_size = 1024;
-
-		link_size = p_readlink("./testrepo/link_to_new.txt", link_data, link_size);
-		link_data[link_size] = '\0';
-		cl_assert_equal_i(link_size, strlen("new.txt"));
-		cl_assert_equal_s(link_data, "new.txt");
-		check_file_contents("./testrepo/link_to_new.txt", "my new file\n");
-	}
-#endif
```
pks-t: It's nice to get rid of all those ifdefs

tests/checkout/index.c
```diff
@@ -175,35 +193,45 @@ void test_checkout_index__honor_coresymlinks_default(void)
...
+
+	opts.checkout_strategy = GIT_CHECKOUT_SAFE | GIT_CHECKOUT_RECREATE_MISSING;
+	cl_git_fail(git_checkout_index(g_repo, NULL, &opts));
+}
+
 void test_checkout_index__honor_coresymlinks_setting_set_to_true(void)
 {
 	git_checkout_options opts = GIT_CHECKOUT_OPTIONS_INIT;
+	char link_data[1024];
+	size_t link_size = 1024;
```
pks-t: We could use `GIT_PATH_MAX` instead here to make this a bit less magical. But it's not like it'd matter much anyway, so feel free to ignore this remark

src/win32/posix_w32.c
```diff
@@ -390,12 +394,20 @@ int p_readlink(const char *path, char *buf, size_t bufsiz)
 	return (int)bufsiz;
 }
 
-int p_symlink(const char *old, const char *new)
+int p_symlink(const char *target, const char *path)
 {
-	/* Real symlinks on NTFS require admin privileges. Until this changes,
-	 * libgit2 just creates a text file with the link target in the contents.
-	 */
-	return git_futils_fake_symlink(old, new);
```
pks-t: Do you know off-hand how we treated these fake symlink in our diff machinery? Were those files shown as modified or was there some "magic" around fake symlinks?

src/win32/posix_w32.c
```diff
@@ -390,12 +394,20 @@ int p_readlink(const char *path, char *buf, size_t bufsiz)
...
+	git_win32_path target_w, path_w;
+	wchar_t *target_p;
+
+	if (git_win32_path_from_utf8(path_w, path) < 0 ||
+		git__utf8_to_16(target_w, MAX_PATH, target) < 0)
+		return -1;
+
+	if (!CreateSymbolicLinkW(path_w, target_w,
+	    SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE))
+		return -1;
```
pks-t: Don't we have to handle the error case a bit more verbosely? I'm not really familiar with the win32 POSIX layer, but we usually invoke `set_errno()` in case where the call fails, don't we?

tests/repo/init.c
```diff
@@ -247,6 +247,19 @@ void test_repo_init__detect_ignorecase(void)
...
 }
 
+void test_repo_init__detect_symlinks(void)
+{
+	struct stat st;
+	bool no_symlinks;
+
+	no_symlinks = (p_symlink("target", "link") < 0 ||
+	    p_lstat("link", &st) < 0 ||
+	    ! (S_ISLNK(st.st_mode)));
```
pks-t: It might make sense to have a common helper function to detect symlink support which unifies these three lines here and in "checkout::index".

tests/checkout/icase.c
```diff
@@ -44,29 +44,6 @@ void test_checkout_icase__cleanup(void)
...
-		NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, NULL));
-
-	cl_win32_pass(getfinalpathfn(fh, filename, MAX_PATH, VOLUME_NAME_DOS));
-
-	CloseHandle(fh);
-
-	git_path_mkposix(filename);
-
-	return filename;
-#else
```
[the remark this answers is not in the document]
ethomson: There might be a bit of a perf hit here.  This was, in fact, written first, IIRC.  And then we had to emulate the `GetFinalPathNameByHandle` call on POSIX.  I expect that enumerating the directory will be a little bit slower, but:

1. the old behavior wasn't correct with symlinks brought into the picture
2. removal of code, yay
3. it can't possibly be _that_ much slower, and it's test code not production.  😀

src/win32/posix_w32.c
ethomson: They weren't shown as different.  The [diff mechanism](https://github.com/libgit2/libgit2/blob/e6cdd17c846648aa8e1e025fa1988475886a551e/src/diff_file.c#L297) will detect if symlinks are supported and either stat the link to determine its destination, or read the file if symlinks are not supported.

src/win32/posix_w32.c
ethomson: We should be good here.  For the posix emulation functions, they can _generally_ fail with an error code without doing anything else.  Callers should then `giterr_set(GITERR_OS...` and `giterr_set` is smart enough to read the win32 system error message.

We only need to `set_errno` when callers actually read the `errno` back out and make a decision about how to handle the error.  (eg, looking for `EAGAIN`).  We try to carefully map the Windows failure into something that callers can use.  And we do it carefully to ensure that callers can understand that `errno` in the way they want to proceed for both Win32 and the actual POSIX.  Trying to make it general purpose instead would be 😓 .

Since we're not doing that here, we're okay to skip that step.

tests/repo/init.c
ethomson: Agreed and done.

tests/checkout/index.c
ethomson: Agreed and done.

ethomson:
I've updated this to use the Git for Windows behavior.  If `core.symlinks=true` is set in a global/xdg/system configuration then symlink support will be _detected_ when creating a new repository.  (If the underlying filesystem does support symbolic links then there will be nothing set in the repository configuration.  If the underlying filesystem _does not_ support symbolic links then `core.symlinks` will be set to `false`.)

On non-Windows systems, the filesystem will be examined like before.

ethomson:
Fixes #4107

pks-t:
On Fri, Oct 26, 2018 at 02:27:08PM +0000, Edward Thomson wrote:
> @pks-t I've added the 3 newest commits to reflect the Git for Windows behavior - a quick sanity check on those would be 👍 when you have some time.

Sorry for not having reviewed this -- have quite a lot of
travelling to do currently, so I'm a bit short on time right now.

danielgindi:
What about SYMBOLIC_LINK_FLAG_DIRECTORY?

ethomson:
> What about SYMBOLIC_LINK_FLAG_DIRECTORY?

What about it?  Is there something deficient in this PR?  Can you provide a simple repro case?

danielgindi:
The problem is with symlinks to directories.
When creating a symlink you should test the target, see if it's a directory or a file, and add the flag accordingly.
This is because in Windows the symlink is "typed", and right now git creates a symlink to a file, which does not work.


Something like:
```c++
  DWORD dwFlags = SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE;

  if (std::filesystem::is_directory(target_w))
    dwFlags |= SYMBOLIC_LINK_FLAG_DIRECTORY;

  if (!CreateSymbolicLinkW(path_w, target_w, dwFlags))
    ...
```

Or if c++17 is not available:
```c++
  DWORD dwFlags = SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE;

  if (GetFileAttributesW(target_w) & FILE_ATTRIBUTE_DIRECTORY)
    dwFlags |= SYMBOLIC_LINK_FLAG_DIRECTORY;

  if (!CreateSymbolicLinkW(path_w, target_w, dwFlags))
    ...
```

ethomson:
My question wasn't _how_ to do it, but _if_ we should do it.  I ask explicitly because omitting that flag will still give you a symbolic link, one that behaves more like a Unix symbolic link:

```
C:\Temp\TestRepos\004>mklink link_to_folder folder
symbolic link created for link_to_folder <<===>> folder

C:\Temp\TestRepos\004>dir
 Volume in drive C has no label.
 Volume Serial Number is DA35-0653

 Directory of C:\Temp\TestRepos\004

03/05/2019  03:15 PM    <DIR>          .
03/05/2019  03:15 PM    <DIR>          ..
03/05/2019  03:14 PM    <DIR>          folder
03/05/2019  03:15 PM    <SYMLINK>      link_to_folder [folder]
               1 File(s)              0 bytes
               3 Dir(s)  39,293,509,632 bytes free
```

(Note the omission of the `/d` flag, which propagates the `SYMBOLIC_LINK_FLAG_DIRECTORY` option.)

But the question is: what does _git_ do.  And indeed, checking this folder out with git does create a directory-style link:

```
C:\Temp\TestRepos\004>dir
 Volume in drive C has no label.
 Volume Serial Number is DA35-0653

 Directory of C:\Temp\TestRepos\004

03/05/2019  03:14 PM    <DIR>          .
03/05/2019  03:14 PM    <DIR>          ..
03/05/2019  03:14 PM    <DIR>          folder
03/05/2019  03:14 PM    <SYMLINKD>     link_to_folder [folder]
               0 File(s)              0 bytes
               4 Dir(s)  39,293,509,632 bytes free
```

I thought when created this PR that I validated that git itself did _not_ create `SYMBOLIC_LINK_FLAG_DIRECTORY` links; it's possible either that they changed their behavior, but probably more likely that I got this wrong.

I agree with your analysis (and indeed, we don't do C++).  Would you be interested in submitting a pull request?

danielgindi:
@ethomson I just tried to answer both questions at the same time (although the latter was not really asked :-)

I'd love to send a PR, but as you already covered this area I think I'll leave it for you to do.
You'll be able to identify if we're messing something up you know.

PR #4714, opened 2018-07-06
travis: use new libgit2-owned signing key for Bintray

pks-t:
By default, all packages uploaded to our Bintray repositories were
signed by the Bintray-owned private key. As Travis is currently unable
to authenticate the official Bintray-signed packages, it is the perfect
occasion to switch the keypair against our own keypair to try to fix
these problems.


pks-t:
The signing issues were fixed by regenerating the Release file. Closing

Issue #4720, opened 2018-07-10
git_ignore_path_is_ignored documentation misleading?


boogisha:
Due to an issue experienced elsewhere (https://github.com/git-tfs/git-tfs/issues/1208), I've stumbled across `git_ignore_path_is_ignored()` function from "libgit2", where it returned TRUE for a path already committed to the repository.

I instantly thought this was a bug, as something already tracked shouldn't be subject to ".gitignore" exclude rules (think [`git check-ignore`][1]), but eventually reading `git_ignore_path_is_ignored()` description, this seems to be by design...?

https://github.com/libgit2/libgit2/blob/89c332e41b12a72d89de40d63bc568c56a2c336a/include/git2/ignore.h#L56-L60

I guess there is a good reason for this (like making ignore rules parsing standalone, where one can check for index existence elsewhere... or something), but additional comment that follows seems misleading:

https://github.com/libgit2/libgit2/blob/89c332e41b12a72d89de40d63bc568c56a2c336a/include/git2/ignore.h#L62-L63

This is not correct in the above mentioned case of an already tracked file, as even if it matches a pattern inside ".gitignore", `git add .` will add modifications of that file to index, where `git_ignore_path_is_ignored()` will still return TRUE as if the file is to be ignored (which is not true).

Of course, in case file is untracked, `git add .` will ignore it as expected (not adding it to index), and `git_ignore_path_is_ignored()` returning TRUE would make sense... just that it doesn't align with `git add .` in general, but in some case(s) only.

Should this `git add .` comparison be dropped (or somehow reworded) to avoid confusion, or am I missing something?

Thanks, Buga

[1]: https://git-scm.com/docs/git-check-ignore

ethomson:
> Should this git add . comparison be dropped (or somehow reworded) to avoid confusion, or am I missing something?

Yes, let's reword it to avoid confusion.  The implication is whether `git add .` would add an untracked file, but I appreciate that this is confusing and we should fix it.

boogisha:
Thanks, and yeah, I thought of maybe specifying "untracked" explicitly in there, like:
> ... on the directory containing the **untracked** file, ...

[...]

PR #4722, opened 2018-07-13
 ignore: improve `git_ignore_path_is_ignored` description Git analogy

boogisha:
In attempt to provide adequate Git command analogy in regards to
ignored files handling, `git_ignore_path_is_ignored` description
mentions doing `git add .` on directory containing the file, and
whether the file in question would be added or not - but behavior of
the two matches for untracked files only, making the comparison
misleading in general sense.

For tracked files, Git doesn't subject them to ignore rules, so even
if a rule applies, `git add .` would actually add the tracked file
changes to index, while `git_ignore_path_is_ignored` would still
consider the file being ignored (as it doesn't check the index, as
documented).

Let's provide `git check-ignore --no-index` as analogous Git command
example instead, being more aligned with what `git_ignore_path_is_ignored`
is about, no matter if the file in question is already tracked or not.

See issue #4720 (git_ignore_path_is_ignored documentation
misleading?, 2018-07-10) for additional information.

### DIFF
--- a/include/git2/ignore.h
+++ b/include/git2/ignore.h
@@ -59,8 +59,8 @@ GIT_EXTERN(int) git_ignore_clear_internal_rules(
  * given file.  This indicates if the file would be ignored regardless of
  * whether the file is already in the index or committed to the repository.
  *
- * One way to think of this is if you were to do "git add ." on the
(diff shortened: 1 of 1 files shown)


ethomson:
Thanks!

PR #4723, opened 2018-07-14
CI: Refactor and introduce VSTS builds

ethomson:
I've refactored the build and test scripts quite a bit in order to simplify and reduce complexity.  Now all builds are three simple steps:

1. Install prerequisites
    For example: using Homebrew to install libraries, or downloading and installing mingw.  This step will only be run on hosted build agents.  Docker images and private build agents (that we provision and provide ourselves) should simply have the dependencies installed correctly.
2. Build
3. Run tests

I've also refactored the build and tests steps so that they can run isolated, without affecting other parts of the system.  For example: the ssh server is now started and configured in a way that doesn't touch `$HOME/.ssh`.  This allows developers to run these scripts without it removing or updating their actual ssh settings.  😁 

Finally, I've introduced CI builds with [Visual Studio Team Services](https://libgit2.visualstudio.com/libgit2/_build/index?context=mine&path=%5C&definitionId=7&_a=completed).  Immediately, this includes seven build targets:

1. Linux (Ubuntu Trusty, with OpenSSL), provided as a Docker image
2. Linux (Ubuntu Trusty, with mbedTLS), provided as a Docker image
3. macOS
4. Windows (MSVC, x86_64)
5. Windows (MSVC, x86)
6. Windows (mingw, x86_64)
7. Windows (mingw, x86)

Per https://github.com/libgit2/discussions/issues/5, I think that we might want to have a separate set of additional builds that run as nightly builds.  For example, [Linux ARM (Raspberry Pi)](https://libgit2.visualstudio.com/libgit2/_build/index?context=mine&path=%5C&definitionId=8&_a=completed).

I think that this is worth merging now so that we can evaluate our PRs with it and get a feel for it to ensure that it's faster than our existing CI systems.  However it's not ready to completely replace Travis CI or AppVeyor, as we're not doing leak detection here yet.


ethomson:
OK, I've iterated on this a bit.  We now run all the platforms that AppVeyor and Travis are currently executing, and we do leak detection everywhere except mingw.  (Linux uses valgrind, macOS uses leaks and Windows uses the existing MSVC leak framework.)

I've removed the Visual Studio 2010 builds, and I don't think it makes sense to reintroduce them.  They existed to ensure that C90 builds worked (ie, declarations happened at the top of blocks), but since #4700, I think that they're quite unnecessary since we're catching those problems in the Linux builds.

So I think this is ready to merge.

tiennou:
I've looked at it going a few days back, and it was indeed pretty quick to complete. Just a thought, that doesn't support Coverity, right ?

ethomson:
> I've looked at it going a few days back, and it was indeed pretty quick to complete. Just a thought, that doesn't support Coverity, right ?

Right - I think that in the long term that we should kick Coverity out into nightly runs of master.  They don't contribute to the PR process since they take too long to round-trip.  So let's not waste cycles to build and upload the results when they won't be processed immediately anyway.

What I think that we should do here is add the VSTS builds (alongside AppVeyor and Travis) as gates in our PR process and evaluate it to decide if we want to switch over completely.  I'll add nightlies independently, for things that take a long time to complete or that we have limited resources for and would significantly delay the PR process timeline with little gain (Raspberry Pi, Coverity, etc).

tiennou:
Coverity on nightlies makes perfect sense. VSTS hype 😉 !!

ci/test.ps1
```diff
@@ -0,0 +1,45 @@
+Set-StrictMode -Version Latest
```
pks-t: I think we should start adopting common prefixes based on the CI system. E.g. "travis-build.sh", "travis-test.sh", "appveyor-build.sh" etc. Right now our "ci/" dir is a mess. I don't see a need to adjust existing scripts in this PR, but it might make sense if newly added files started to already use such names.

ci/test.ps1
```diff
@@ -0,0 +1,45 @@
...
+if ($LastExitCode -ne 0) { [Environment]::Exit($LastExitCode) }
+
+Write-Host ""
+Write-Host "Running proxy tests"
+Write-Host ""
+
+$Env:GITTEST_REMOTE_PROXY_URL="localhost:8080"
+$Env:GITTEST_REMOTE_PROXY_USER="foo"
+$Env:GITTEST_REMOTE_PROXY_PASS="bar"
+ctest -V -R proxy
```
pks-t: Well, this is not a strict move, but you do add additional things. Not that that would be bad, just a bit surprising given the commit message.

ci/build.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+#
+# Environment variables:
+#
+# SOURCE_DIR: Set to the directory of the libgit2 source (optional)
+#     If not set, it will be derived relative to this script.
```
pks-t: What about the old CI scripts, shouldn't they just get removed if this is a mere refactoring? Also, what's the difference introduced by your refactoring? Sorry to be so pushy, but it's kind of hard to read as the only thing one can see in this refactoring is a new file being added.

ci/test.sh
```diff
@@ -1,101 +1,159 @@
...
-echo "################################################################################"
-echo "## Running (online) tests                                                     ##"
-echo "################################################################################"
+	echo ""
+	echo "##############################################################################"
+	echo "## Running (online) tests"
+	echo "##############################################################################"
 
-#ctest -V -R online
+	ctest -V -R online || die $?
```
pks-t: Do we really want to die if any of those tests fails? I think it would be more sensible to just record that a test has failed and die after _all_ tests ran. If you die eagerly, you'll potentially have to re-run CI every time you have fixed a test just to stumble into the next one that was previously not executed.

ci/test.sh
```diff
@@ -1,101 +1,159 @@
...
+	echo "Running proxy tests"
+	echo ""
+
+	export GITTEST_REMOTE_PROXY_URL="localhost:8080"
+	export GITTEST_REMOTE_PROXY_USER="foo"
+	export GITTEST_REMOTE_PROXY_PASS="bar"
+	ctest -V -R proxy || die $?
+	unset GITTEST_REMOTE_PROXY_URL
+	unset GITTEST_REMOTE_PROXY_USER
+	unset GITTEST_REMOTE_PROXY_PASS
```
pks-t: Instead of using `unset` you should probably just use a subshell

ci/setup-linux.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -x
+
+apt-get update
+apt-get -y install build-essential pkg-config cmake openssl libssl-dev libssh2-1-dev libcurl4-gnutls-dev openssh-server
+
+if [ ! -d /var/run/sshd ]; then
+	sudo mkdir /var/run/sshd
```
pks-t: `sudo` is not necessary here, as we're already running with elevated privileges. You can also avoid the conditional by just using `mkdir -p`, which will not complain if the target directory already exists.

ci/setup-macos.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -x
+
+brew update
+brew install pkgconfig
+brew install zlib
+brew install curl
+brew install openssl
+brew install libssh2
```
pks-t: There's no need to list them all separately, you can just pass multiple formulas to brew (even though it is not documented).

ci/build.sh
```diff
@@ -27,6 +27,7 @@ echo "##########################################################################
 echo "## Configuring build environment"
 echo "##############################################################################"
 
+echo cmake ${SOURCE_DIR} -DBUILD_EXAMPLES=ON ${CMAKE_OPTIONS}
```
pks-t: Leftover debug output? :)

.vsts-ci.yml
```diff
@@ -0,0 +1,134 @@
...
+- phase: linux_trusty_openssl
+  displayName: 'Linux (Trusty OpenSSL)'
+  queue:
+    name: 'Hosted Linux Preview'
+  steps:
+  - task: Docker@0
+    displayName: Build
+    inputs:
+      action: 'Run an image'
+      imageName: 'ethomson/libgit2-trusty-openssl:latest'
```
pks-t: Oh, yeah. Forgot to grant you permissions on DockerHub

.vsts-ci.yml
```diff
@@ -0,0 +1,134 @@
...
+      volumes: |
+       $(Build.SourcesDirectory):/src
+       $(Build.BinariesDirectory):/build
+      workDir: '/build'
+      containerCommand: '/src/ci/build.sh'
+      detached: false
+  - task: Docker@0
```
pks-t: Is it intended that these tasks both have the same identifier? I don't really know the semantics, just left me wondering

ci/setup-mingw-x86.ps1
```diff
@@ -0,0 +1,20 @@
...
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
+[Reflection.Assembly]::LoadWithPartialName("System.IO.Compression.FileSystem");
+
+Write-Host "##############################################################################"
+Write-Host "## Downloading mingw"
+Write-Host "##############################################################################"
+
+$mingw_uri = "https://bintray.com/libgit2/build-dependencies/download_file?file_path=mingw-w64-i686-8.1.0-release-win32-sjlj-rt_v6-rev0.zip"
+$platform = "x86"
```
pks-t: Both scripts are essentially the same, except for "s/x86/x86_64/" and "s/i686/x86_64", right? If so we could just have it accept a parameter "arch" instead of duplicating it

ethomson:
> I don't really like that we have another set of CI scripts now. Is it possible to simply have a set of scripts that can be used from both VSTS and Travis/AppVeyor?

It's possible, yes.  But I didn't really want to put a lot of effort into making Travis and AppVeyor use the new scripts if we're just going to stop using those build hosts altogether.  That would just be unnecessary work.

What I would like to do is merge this to evaluate VSTS.  If we want to move to VSTS only, then just delete the old CI scripts in the `scripts` directory.  If we do not, for some reason, then update AppVeyor and Travis to use the new scripts.

> I feel like Windows is a bit overrepresented right now, with 4 Windows builds and only there builds for the other platforms. Noticeably missing are 32 bit builds for macOS and Linux, which did previously exist. So personally, I'd either want to include 32 bit builds for those platforms or remove 32 bit builds for Windows. I don't want the project to look like it favours any platform, and having so many Windows builds versus so few other builds might just create that look.

We don't do any 32 bit builds on macOS or Linux now.  Nor do I think that we should, at least on pull requests, since I don't think that anybody uses those platforms.

Our current matrix is:
```
Linux gcc -DTHREADSAFE=ON -DENABLE_TRACE=ON -DCMAKE_BUILD_TYPE=Release
Linux clang -DTHREADSAFE=ON -DENABLE_TRACE=ON -DCMAKE_BUILD_TYPE=Release
Linux gcc -DTHREADSAFE=OFF -DBUILD_EXAMPLES=ON
Linux clang -DTHREADSAFE=OFF -DBUILD_EXAMPLES=ON
Linux clang valgrind -DBUILD_CLAR=ON -DBUILD_EXAMPLES=OFF -DDEBUG_POOL=ON -DCMAKE_BUILD_TYPE=Debug
Linux gcc mbedTLS -DTHREADSAFE=ON -DCMAKE_BUILD_TYPE=Release
Linux gcc mbedTLS -DTHREADSAFE=OFF
macOS -DTHREADSAFE=ON
macOS -DTHREADSAFE=OFF
```

So this is amd64 across the board.  What I've changed is removing the gcc vs clang variant, threadsafe vs non-threadsafe variant, and release vs debug variant.    And it adds a whole additional build task for valgrind, when we could simply bake that into the regular runs.

Overall, the reduction of build tasks is desirable and valuable.  I think that I've done a good job of distilling this into the core value of our builds and removing unnecessary variants.  I kept the x86 builds on Windows because my experience is that some people _do_ actually build and run x86 images on Windows - in fact, it's still the default new project experience in Visual Studio.

If you think that we need more variants, then let's add them, but let's do so because we think that adding them is actually worthy of that increase of time-to-validation.

ci/setup-mingw-x86.ps1
ethomson: And the filename, which I'd rather embed in the script than in the yaml.

.vsts-ci.yml
ethomson: It's not an identifier, it's the type of task to run - this is a Docker task, version 0.

ethomson:
> What I would like to do is merge this to evaluate VSTS. If we want to move to VSTS only, then just delete the old CI scripts in the scripts directory. If we do not, for some reason, then update AppVeyor and Travis to use the new scripts.

This is mostly just me being lazy, btw.  I had the same thought of doing this myself, but let me spike it and see how bad it would be.

ci/test.sh
ethomson: This is what we're doing today.  🤷‍♂️ I would rather have us continue to fail fast, on the hopes that we get out of the CI queue as quickly as possible.  I guess that could lead to someone spamming the queue with brokenness, but we haven't seen that so far.

pks-t:
Oh, yeah. I assumed we had 32bit builds on Linux due to there being more jobs. But I completely missed Clang, which I definitely would like to include on Linux. Quite often, Clang spews out different errors and/or warnings than GCC, which is really helpful. Just today I have seen differences in a CI job based on GCC vs. Clang.

I agree about always executing leak checking instead of having a separate job for it. It shouldn't impact runtime behaviour  at all (except for time-based differences), so it's a good thing to have. I also think that having non-threaded builds run on every PR isn't required -- I haven't ever seen a build with non-threading that had errors not existing on threaded builds, only the other way round. So running it as part of the nightly builds is definitely sufficient.

ethomson:
Yes, I agree with the gcc vs clang distinction.  I was wondering if we should build with clang _instead_ of gcc, but I'll include both for now.

ethomson:
> I agree about always executing leak checking instead of having a separate job for it. It shouldn't impact runtime behaviour at all (except for time-based differences), so it's a good thing to have.

Yeah, even if it's a little slower than a non-valgrind run, I think that the overall CI build time drops significantly because we don't have to queue yet another build worker, do the build, etc.

pks-t:
We definitely should have a GCC build on Linux. Nearly all distributions still have GCC as the default compiler, and that's not going to change soon. On macOS it's a different thing, obviously. I even think that gcc is a symlink to clang there.

ethomson:
> On macOS it's a different thing, obviously. I even think that gcc is a symlink to clang there.

Indeed:

```
% gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk/usr/include/c++/4.2.1
Apple LLVM version 9.1.0 (clang-902.0.39.1)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

tests/CMakeLists.txt
```diff
@@ -54,12 +54,8 @@ IF (MSVC_IDE)
...
-ENDIF ()
-
-# Add additional test targets that require special setup
-ADD_TEST(libgit2_clar-proxy_credentials "${libgit2_BINARY_DIR}/libgit2_clar" -v -sonline::clone::proxy_credentials_in_url -sonline::clone::proxy_credentials_request)
-ADD_TEST(libgit2_clar-ssh "${libgit2_BINARY_DIR}/libgit2_clar" -v -sonline::push -sonline::clone::ssh_cert -sonline::clone::ssh_with_paths)
+ADD_TEST(offline   "${libgit2_BINARY_DIR}/libgit2_clar" -v -xonline)
+ADD_TEST(online    "${libgit2_BINARY_DIR}/libgit2_clar" -v -sonline)
+ADD_TEST(gitdaemon "${libgit2_BINARY_DIR}/libgit2_clar" -v -sonline::push)
+ADD_TEST(ssh       "${libgit2_BINARY_DIR}/libgit2_clar" -v -sonline::push -sonline::clone::ssh_cert -sonline::clone::ssh_with_paths)
+ADD_TEST(proxy     "${libgit2_BINARY_DIR}/libgit2_clar" -v -sonline::clone::proxy_credentials_in_url -sonline::clone::proxy_credentials_request)
```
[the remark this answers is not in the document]
pks-t: Good question. I don't think so, at least `ADD_TEST` doesn't mention it. How about just hiding it behind a build option?

ci/test.ps1
pks-t: I totally agree. And in fact, I didn't mean to say that "ci/" is a mess, but "scripts/". I didn't realize at first that you've created the "ci/" directory in this PR, it already makes things a lot cleaner. I'm still not particular happy about just distinguishing scripts by their file ending, only, but if we're able to merge the new Unix CI-scripts and the old Travis CI-scripts in the short run I'm rather indifferent.

ci/setup-mingw-x86.ps1
pks-t: Which can still easily be done, right?

```
case "$ARCH in
    x86)
        mingw_uri="https://..."
        platform=x86
        ;;
    amd64)
        ...
esac
```

(Sorry for using Bash, I'm not familiar with PS)

ethomson: Yes, I agree - this is a good suggestion.

tests/CMakeLists.txt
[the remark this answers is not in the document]
tiennou: Just my 2 unrelated cents, but CMake support reporting a test as skipped, though it's not quite compatible with our "shove every test in libgit2_clar".

Maybe a "standard" output format could be added to clar so we can enable this ?

https://docs.microsoft.com/en-us/vsts/pipelines/tasks/test/publish-test-results?view=vsts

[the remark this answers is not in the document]
ethomson: I'd very much like to make clar emit a standard output format.

ethomson:
OK.  I think that I've addressed your comments, @pks-t.

1. I've migrated Travis and AppVeyor both over to the new scripts and removed the old ones.  I think that they're an improvement over what was there, but I still think reducing the number of CI systems we have should be our goal.  I did a very literal translation, we're doing all the same builds on each of those systems as what we did before.

2. I added clang to the platform matrix on VSTS, which gives us:  
  
    Linux (Trusty; GCC; OpenSSL)
    Linux (Trusty; GCC; mbedTLS)
    Linux (Trusty; Clang; OpenSSL)
    Linux (Trusty; Clang; mbedTLS)
    macOS
    Windows (Visual Studio; amd64)
    Windows (Visual Studio; x86)
    Windows (MinGW; amd64)
    Windows (MinGW; x86)

ethomson:
OK, seeing as there's no other feedback, I'm going to merge this later today to start getting some idea of how it's performing against PRs.

pks-t:
On Thu, Aug 02, 2018 at 02:11:52AM -0700, Edward Thomson wrote:
> OK, seeing as there's no other feedback, I'm going to merge this later today to start getting some idea of how it's performing against PRs.

I was planning to do another review round today. That being said,
as long as Travis/AppVeyor aren't yet being phased out I'm happy
to merge this early-ish to see how it performs.

pks-t:
I ran out of time today, so I only had a cursory read. From what I saw it looked good to me. I'll have more time to review tomorrow, but feel free to go ahead and merge it. We can still improve things after that.

PR #4727, opened 2018-07-18
tree: accept null ids in existing trees when updating

carlosmn:
When we add entries to a treebuilder we validate them. But we validate even
those that we're adding because they exist in the base tree. This disables
using the normal mechanisms on these trees, even to fix them.

Keep track of whether the entry we're appending comes from an existing tree and
bypass the name and id validation if it's from existing data.


carlosmn:
This would be one for backporting into a maintenance release.

tests/object/tree/update.c
```diff
@@ -284,3 +284,18 @@ void test_object_tree_update__add_conflict2(void)
 
 	cl_git_fail(git_tree_create_updated(&tree_updater_id, g_repo, NULL, 2, updates));
 }
+
+void test_object_tree_update__remove_invalid_submodule(void)
```
pks-t: Could you please add a small comment detailing what is so special about the tree 396c7f1? It's rather hard to reason about this test otherwise :)

carlosmn:
Updated from @pks-t 's feedback

PR #4728, opened 2018-07-19
Fuzzers

pks-t:
This includes the fuzzers from @lhchavez and @nelhage, superseding #4433 and superseding the current fuzzers stored in google/oss-fuzz.

I've made some adjustments to the build instructions. Most notably, users don't have to manually add "-DCMAKE_EXE_LINKER_FLAGS=-fsanitize=fuzzer" anymore, and the build errors out during the configuration stage if any of the required sanitizer flags is not supported.

@nelhage, I'd please ask you to give me permissions to add a GPL-2.0 license header to your original fuzzer. It was missing one, and I want to be on the safe side ;) I've also mentioned that the fuzzer is originally yours, but if you wish I can change author information on that commit to pretend it was you who actually committed the C++ one to libgit2. Please tell me your preference.

Please do not merge until @nelhage confirms we're allowed to distribute his code under GPLv2.


pks-t:
Sorry for duplicating the work. I'll see if I can scavenge some bits from your PR.

One question I currently have is where we want to store corporas. We can either store them in the oss-fuzz repo (like it is currently done for the download fuzzer) or in our own repo (like it is done for the pack fuzzer). While the files for the download fuzzers are around 8kB only, the other one is more than a whole megabyte. So it's reasonably sized, but it still got me thinking.

pks-t:
I've decided to remove the USE_SANITIZER and USE_COVERAGE options and adjusted the documentation to recommend setting those CFLAGS via environment variables. I think providing options for all the different variations a user might want to build libgit2 is not really reasonable. In fact we cannot even use that option with oss-fuzz, as they want to pass in the used sanitizers via CFLAGS themselves.

pks-t:
Oops, hit the button too early.

Anyway, I think this PR is ready now. I have a commit waiting at pks-t/oss-fuzz which switches oss-fuzz to use our internal fuzzers and corpora instead of using the one currently hosted at google/oss-fuzz. I'll create a PR as soon as this one here is merged.

fuzzers/packfile_raw_fuzzer.c
```diff
@@ -0,0 +1,118 @@
+/*
+ * libgit2 raw packfile fuzz target.
```
lhchavez: oops, i just realized that i named the file wrong. it should have been `raw_packfile` in the filename for consistency with this comment.

the reason is that originally i wanted to split this into raw and cooked (with a valid hash at the end) packfiles, and it made more sense to name them `packfile_{raw,cooked}`. but then I figured out it was better to use the MSB of the first byte to change behavior, so I ended not using the `cooked` version.

Do you mind renaming this file to `raw_packfile_fuzzer`? or just `packfile_fuzzer` to avoid adding an unneeded qualifier to the name.

pks-t: Sure. I also prefer "packfile_fuzzer", so I've renamed it

nelhage:
I think you'll want to also limit the pack object count for `download_refs_fuzzer`; We've seen oss-fuzz find that codepath there, too.

pks-t:
Right. I've also lowered the limit for the other fuzzer

pks-t:
I've created google/oss-fuzz#1684 to use our own, upstreamed fuzzers in oss-fuzz

PR #4729, opened 2018-07-19
Import download_refs_fuzzer and port to C

nelhage:
I was working on this concurrently with @pks-t writing https://github.com/libgit2/libgit2/pull/4728. That PR should likely supersede this one, but since I believe I have a working port to C I wanted to put it up so we can avoid duplicating that work.


nelhage:
Ah, reading #4728 more closely I see you'd already done the conversion as well!

PR #4732, opened 2018-07-20
Squash some leaks

ethomson:
When running valgrind on the `online::push` tests, some leaks became apparent.  Two in the tests themselves, one in the product code that occurs when restarting a push on an existing remote.

### DIFF
--- a/src/streams/mbedtls.c
+++ b/src/streams/mbedtls.c
@@ -89,6 +89,8 @@ int git_mbedtls_stream_global_init(void)
 	mbedtls_x509_crt *cacert = NULL;
 
 	git__ssl_conf = git__malloc(sizeof(mbedtls_ssl_config));
+	GITERR_CHECK_ALLOC(git__ssl_conf);
+
 	mbedtls_ssl_config_init(git__ssl_conf);
 	if (mbedtls_ssl_config_defaults(git__ssl_conf,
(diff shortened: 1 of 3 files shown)


ethomson:
> I'm slightly ashamed of being responsible for leaks. I forgot that the push suite is disabled on default runs…

No problem - I'm proposing that we enable leak detection on all test runs on all builds, which is where I found it.

pks-t:
This has already been backported partially, excluding the mbedtls changes as this backend doesn't yet exist in v0.27. Removing the backport label

grahamc:
I'm not sure this should have been backported, it seems this has caused some API behavior breakage. In particular, this breaks Cargo's ability to fetch repositories, specifically https://github.com/alexcrichton/git2-rs/blob/master/src/transport.rs#L223 which is fixed after reverting 49ee0ae6ff1621d3093947c08868f4a4aeb56a91. See also https://github.com/rust-lang/cargo/issues/6365

carlosmn:
The built-in transports store a partsed version of the url themselves during initialisation and ignore the URL passed in in `action`.

This does look like a bug in the aforementioned commit. We run `git_smart__reset_stream` all the time in RPC (i.e. HTTP) mode and while it might make sense to clear the current stream, we definitely want to keep the url around.

PR #4733, opened 2018-07-21
utils: make sure memchr() is not invoked with NULL

### LINKED ISSUE #4726: git__linenlen calls memchar with null pointer
### Reproduction steps
`git__linenlen` in `util.c` calls memchar with null pointer passed as argument 1.  Can be reproduced with the following code:

```
#include <git2.h>

int main()
{
    git_repository *repo = NULL;

    git_libgit2_init();
    git_repository_init(&repo, "testrepo", 0);
    git_repository_free(repo);

    return 0;
}
```

### Expected behavior

Avoid call to memchr with null pointer.

### Actual behavior

When I test memory access errors in the R bindings [git2r](https://github.com/ropensci/git2r) using Undefined Behavior Sanitizer, I see `libgit2/src/util.c:354:8: runtime error: null pointer passed as argument 1, which is declared to never be null`

99bb98c0e8abb97b1daa3aff0307b93ed9291c08

Fedora 28

neithernut:
Passing a null pointer as first argument to `memchr()` results in
undefined behavior.

In practice, this should not be a problem since, within libgit2,
`git__linenlen()` will be either called with a non-null `buffer` or with
both `buffer` and `buffer_len` being null. Any sane implementation of
`memchr()` should return `NULL` in this case without causing any
malicious or unintended effects.

However, it is still an UB. It will be reported by both dynamic and
static UB checkers. Hence, we add this additional check.

Resolves #4726.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -351,7 +351,13 @@ char *git__strsep(char **end, const char *sep)
 
 size_t git__linenlen(const char *buffer, size_t buffer_len)
 {
-	char *nl = memchr(buffer, '\n', buffer_len);
+	char *nl;
+
+	assert(buffer || buffer_len == 0);
+	if (!buffer)
+		return 0;
+
+	nl = memchr(buffer, '\n', buffer_len);
 	return nl ? (size_t)(nl - buffer) + 1 : buffer_len;
 }
 


neithernut:
As determined in the referenced issue, there are two callers, both in `parse.c`, both carrying out operations on a `git_parse_ctx`. In both cases, `git__linenlen()` is used for determining and setting the length (field `line_len`) of the next line (pointed to by field `line` of a `git_parse_ctx`). `line` may be `NULL`, e.g. if:
 * the parse context is be initialized with `NULL` for the buffer,
 * `0` is passed for the buffer-length to the initialization function or
 * the context is reset using `git_parse_ctx_clear()`.

Based on this, I considered three options:
 1) Enforcing field `line` inside `git_parse_ctx` to be non-null: having `NULL` mean "empty buffer" is quite common, but I suspect that the context is actually always initialized with some non-null buffer which would, in turn, result in a non-null line. However, I did not check. Also, changing the semantics so that `line` is set to a pointer to an empty string would mean introducing multiple checks, extra code for `git_parse_ctx_clear()` and, most notably, would introduce an invariant rather uncommon (in C) for this type of object.
 2) Adding additional checks to the call-site: every call-site is affected, so every (existing) call-site would need some additional check. In the case of `git_parse_advance_line()`, this could be an early return, which would provide the additional benefit of the `line_num` field not being incremented. However, I assume that this does not really matter, since the line-number may be considered "undefined" anyway for a null-buffer (it's not nice, but I doubt anybody cares).
 3) Adding an additional check to `git__linenlen()` (as proposed in this PR): this required adding only a single check, rather than a check for every single call-site. This makes the function "safe" for the situations of interest, both for the existing and possible future uses of that function.

If desired, I can either change this PR or file an alternative PR implementing the 2nd of those approaches rather then the 3rd.

src/util.c
```diff
@@ -351,7 +351,12 @@ char *git__strsep(char **end, const char *sep)
 
 size_t git__linenlen(const char *buffer, size_t buffer_len)
 {
-	char *nl = memchr(buffer, '\n', buffer_len);
+	char *nl;
+
+	if (!buffer)
```
pks-t: I think this change is sensible. I'd love to have an `assert(buffer || !buffer_len);` here, though, such that we can catch programming errors in case anybody calls `git__linenlen(NULL, n)` with `n > 0`.

neithernut: I added an assertion.

neithernut:
One of the AppVeyor jobs failed because of network problems. It should be sufficient to simply restart the job.

ethomson:
Thanks for the analysis, @neithernut.  The reason I asked is because few (none?) of our other string helper functions do a null check like this.  (I didn't do an exhaustive analysis, I just quickly skimmed `util.c`.)  I'd like to take a moment to do a little more digging to ensure that we don't change a function to be very dissimilar than the rest of the internal library when it comes to its preconditions.

pks-t:
While most (all?) other string functions don't do `NULL` checks, I find the semantics in this case to be rather self-evident: if you've got no string, the line length is clearly "0". On the other hand strlen(3P) exhibits undefined behaviour in case its parameter is `NULL`, which would be a case against including the `NULL` check in `git__linelen`. Also: even though we'd be safe wrt `git_linelen`, it might still be that we pass the `NULL` string to other functions afterwards, as @ethomson said.

So this all speaks against my previous thinking. So maybe we should make sure that a parser context is never being initialized with a `NULL` pointer?

ethomson:
> So this all speaks against my previous thinking. So maybe we should make sure that a parser context is never being initialized with a NULL pointer?

That makes sense to me.

neithernut:
In this case, I'll prepare patches implementing approach 2 (changing `git_parse_*`). Probably over the weekend.

PR #4734, opened 2018-07-21
Warning API (again)

tiennou:
I was looking at PR sorted by "age", hence discovered that proposing warning APIs was a sought-after pastime. So, here's my take on that 😝.

For reference, #2101 is @arrbee's proposal, and #3722 is @carlosmn's.

Note that this is completely not flushed out at all and/or the implementation is hopelessly wrong. I'm just proposing it to get the ball rolling again.

Main differences with the previous proposals:
- this should allow filtering for warning types at registration, using type/code/klass mechanism (akin to giterr), as well as unregistration.
  The rationale is that you might want to (temporarily) be informed of parsing issues near the code that triggered it, while still being specific about them.

- there's no subclassing involved. Specific warning classes get passed specific context structs. I didn't want to see an explosion of `git_warn` functions.

- the "raise" part is separated into 2 functions, "raise" and "ask". The former is "FYI-warnings", the latter is for asking the user what should happen (that's the only documented part).

Example library usage :
```c
int git_object_parse(git_repository *repo, git_oid obj_oid, …)
{
    git_warning_objparse_context parse_ctx;

    /* plain warning, like config not readable ? */;
    res = read_config(&config_file);
    if (res != 0) {
        git_warning_config_context config_ctx;
        config_ctx.file = config_file;
        git_warn_raise(GIT_WARNING_CLASS(GIT_WARNING_CONFIG, GIT_WARNING_CONFIG__UNREADABLE, repo, &config_ctx, "failed to read config file: %s", config_file);
        goto cleanup;
    }

    parse_ctx.oid = obj_oid;
    parse_ctx.data = buffer.ptr + pos;
    res = /* parse buffers */;
    if (res != 0) {
        int reply = git_warn_ask(GIT_WARNING_CLASS(GIT_WARNING_OBJPARSE, GIT_WARNING_OBJPARSE__INVALID_RETICULATION, repo, &parse_ctx, "the reticulation parsing failed");
        if (reply == 1) {
            /* ignore missing reticulation */
        } else if (reply == 2) {
            /* replace reticulation with default value */
        } else if (reply == 0 || reply == GIT_PASSTHROUGH) {
            /* reticulation is allowed to fail */
        } else {
            /* we upset the user */
            goto cleanup;
    }

    /* ... */
}
```

Example user usage :
```c
int lax_parse_function(…)
{
    return 0;
}

int revwalk_the_things(void)
{
    git_warning_registration token;
    git_warning_register(&token, GIT_WARNING_CLASS(GIT_WARNING_OBJPARSE, GIT_WARNING_OBJPARSE_ANY), lax_parse_function, NULL);

    /* git_revwalk */

    git_warning_unregister(token);
}
```

After having written that, I'm not sure I like the `GIT_WARNING_` long names, especially since you have to AND them…


ethomson:
Neat - thanks for picking this up.  I'll need to give this a read through, but my immediate thought is:

> the "raise" part is separated into 2 functions, "raise" and "ask". The former is "FYI-warnings", the latter is for asking the user what should happen (that's the only documented part).

What use case are you imagining for an FYI-warning?  I think that you should just roll that into the regular warning mechanism for simplicity.  In your example, it seems pretty reasonable that callers have a choice of what to do there.  They could upgrade that to an error, or let it continue the program...

ethomson:
I really like where this is going, BTW.  I love that you're striving for simplicity here.

tiennou:
> What use case are you imagining for an FYI-warning?

Cases where we can detect a problem, but there's not really anything to be done, like [this](https://github.com/libgit2/libgit2/pull/2101#issuecomment-34664645). I just made two different functions for semantic reasons.

IMHO ideally those would be the only kind of warnings we have, because I'm kinda suspicious of the other/`ask` kind (like lax-parsing). How you can actually respond to them (since they go through a globally-defined callback) seems *really* unwieldy for users (that's was triggered the "many listener" approach I'm taking).

Do we have an idea of how the various (err…) components we could use warnings for would be `raise` vs. `ask` ? I'm tempted to say that maybe parsing is the odd one out, and it might not be worth the additional complexity just to handle that (ie. a "parse-error" cb on the odb might be sufficient).

tiennou:
> you should just roll that into the regular warning

Yeah, the "regular" word here makes me think we have our warning meanings reversed 😉. To me a "FYI-callback" is a warning (like in compiler terms). The other kind is more like some kind of prompt, with a considerably harder implementation (hence my question on "how exceptional/common" it's expected to be).

ethomson:
I said “regular” because it seems odd to me that there are classes of warnings that a user would never want to stop processing on.  They may not want to prompt but I can imagine that users may want to short circuit any warning.

Even in the “compiler warning” example, the user may have an error in their warning callback and they want to stop.

tiennou:
Okay, I refined the implementation. The "FYI-warn" function is gone (though it's not hard to reimplement). One thing I think it would allow is sidestepping Coverity's "unchecked return" errors 😉.

My comparison function is still buggy, so it doesn't quite work the way I'd like it to, but the test is correct. The notification order should thus be :
- all code-specific callbacks for a given type are called
- all any-code callbacks for a given type are called
- all any-type callbacks are called
- a non-default reply from a warning callback will short-circuit and use that response immediately.

I feel like the short-circuiting is the least ugly, as we only get one reply (vs. having a bunch of potentially conflicting ones). This means that not all "observers" will be informed (a workaround for "userland" is to call its relevant callbacks itself). An alternative design would be to keep notifying, and only act on the first reply, but that's ugly.

PR #4741, opened 2018-07-26
Clar JUnit XML Support

tiennou:
As per https://github.com/libgit2/libgit2/pull/4723#discussion_r205529535, here's a (shoddy) JUnit XML writer for Clar.

It's completely untested, I've only checked the XML against this [XSD](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd) but that's all. Reporting is also quite spotty (a few accounting knobs are missing from clar, but they should be simple). I'll keep my fingers crossed that nothing will put unescaped names because there's no escaping done.

You should be able to pass `-r` and get a summary.xml file in the working directory. Unless clar crashes.

### DIFF
--- a/tests/clar.c
+++ b/tests/clar.c
@@ -132,6 +132,7 @@ static struct {
 	int report_errors_only;
 	int exit_on_error;
 	int report_suite_names;
+	int write_summary;
 
 	struct clar_report *reports;
 	struct clar_report *last_report;
@@ -344,6 +345,7 @@ clar_usage(const char *arg)
 	printf("  -q    \tOnly report tests that had an error\n");
 	printf("  -Q    \tQuit as soon as a test fails\n");
 	printf("  -l    \tPrint suite names\n");
+	printf("  -r    \tWrite summary file\n");
 	exit(-1);
 }
 
@@ -357,7 +359,7 @@ clar_parse_args(int argc, char **argv)
 		char *argument = argv[i];
 
 		if (argument[0] != '-' || argument[1] == '\0'
-		    || strchr("sixvqQl", argument[1]) == NULL) {
+		    || strchr("sixvqQlr", argument[1]) == NULL) {
 			clar_usage(argv[0]);
 		}
 	}
@@ -436,6 +438,10 @@ clar_parse_args(int argc, char **argv)
 			_clar.report_suite_names = 1;
 			break;
 
+		case 'r':
+			_clar.write_summary = 1;
+			break;
+
 		default:
 			assert(!"Unexpected commandline argument!");
 		}
@@ -475,6 +481,8 @@ clar_test_run(void)
 	return _clar.total_errors;
 }
 
+static void clar_summary_write(void);
+
 void
 clar_test_shutdown(void)
 {
@@ -485,6 +493,9 @@ clar_test_shutdown(void)
 	);
 
 	clar_unsandbox();
+
+	if (_clar.write_summary)
+		clar_summary_write();
 }
 
 int
@@ -689,3 +700,4 @@ void cl_set_cleanup(void (*cleanup)(void *), void *opaque)
 #include "clar/fixtures.h"
 #include "clar/fs.h"
 #include "clar/print.h"
+#include "clar/summary.h"
--- a/tests/clar.c
+++ b/tests/clar.c
@@ -95,9 +95,6 @@ static const char *
 fixture_path(const char *base, const char *fixture_name);
 
 struct clar_error {
-	const char *test;
-	int test_number;
-	const char *suite;
 	const char *file;
 	int line_number;
 	const char *error_msg;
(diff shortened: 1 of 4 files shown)


tiennou:
Yeah, I would have *loved* to not have be anywhere near XML 😉. Note that my summary is actually invalid w.r.t to the XML schema.

My main gripe is that AFAIK we're "misusing" ctest when we do a monster-test-executable, as its expectation is that it is the [test runner](https://cmake.org/cmake/help/v3.2/manual/ctest.1.html). IOW, it expects one `ADD_TEST` stanza per testcase. But this would enable its full range of functionality, like parallel (`-j <jobs>`), randomized (`--schedule-random`) runs. Test selection is available through `-L/-R/-E <regex>` so this wouldn't be missed, and there's a label system for things like `online::push`.

So (and this is pretty fuzzy because I'm randomly typing keywords), it seems ctest ought to be able to output NUnit XML, because that's one of the [formats accepted by CDash](https://public.kitware.com/Wiki/CDash:XML) (the CMake "dashboard").

And secondly the VSTS task only has support for XML formats, and most of them look somewhat VisualStudio-oriented (what are assemblies ?). Even though it's [open-source](https://github.com/Microsoft/vsts-tasks/tree/master/Tasks/PublishTestResultsV2), the cursory look I gave it made me think that it's only performing a "transfer" to some other part of the VSTS infrastructure, and I haven't been able to see if TAP support could be added to that.

The main question thus become : do we want to embrace ctest and change clar to be a dumber "test harness", only responsible of one specific test suite, maybe even a single test case ? I'd be willing to make a pass at this ; after more keyword-digging though, I've yet to find a definitive "you can report your tests using those formats" in the ctest documentation…

pks-t:
On Fri, Jul 27, 2018 at 04:56:52AM -0700, Etienne Samson wrote:
> Yeah, I would have *loved* to not have be anywhere near XML 😉. Note that my summary is actually invalid w.r.t to the XML schema.
> 
> My main gripe is that AFAIK we're "misusing" ctest when we do a monster-test-executable, as its expectation is that it is the [test runner](https://cmake.org/cmake/help/v3.2/manual/ctest.1.html). IOW, it expects one `ADD_TEST` stanza per testcase. But this would enable its full range of functionality, like parallel (`-j <jobs>`), randomized (`--schedule-random`) runs. Test selection is available through `-L/-R/-E <regex>` so this wouldn't be missed, and there's a label system for things like `online::push`.
> 
> So (and this is pretty fuzzy because I'm randomly typing keywords), it seems ctest ought to be able to output NUnit XML, because that's one of the [formats accepted by CDash](https://public.kitware.com/Wiki/CDash:XML) (the CMake "dashboard").
> 
> And secondly the VSTS task only has support for XML formats, and most of them look somewhat VisualStudio-oriented (what are assemblies ?). Even though it's [open-source](https://github.com/Microsoft/vsts-tasks/tree/master/Tasks/PublishTestResultsV2), the cursory look I gave it made me think that it's only performing a "transfer" to some other part of the VSTS infrastructure, and I haven't been able to see if TAP support could be added to that.

Fair enough, so no way around XML if we want proper support.

> The main question thus become : do we want to embrace ctest and change clar to be a dumber "test harness", only responsible of one specific test suite, maybe even a single test case ? I'd be willing to make a pass at this ; after more keyword-digging though, I've yet to find a definitive "you can report your tests using those formats" in the ctest documentation…

I think a build system should be just that: a system building our
stuff. I don't want to lock us in to CMake even further, as that
might impact our ability to change the build system if the need
arises at some point in the future. So from my point of view: no,
we don't want to embrace CMake. But I'm open to hearing other
positions.

tiennou:
Note that I'm somewhat torn on the `ctest` issue, since it has obvious benefits that are missing from clar (not to be mean to `clar`, apart from the one-executable-issue, I'm actually impressed at how clean it looks and how well it works).

I should have fixed *most* of the issues with the XML, but I'm not sure : all Schema validators I could find seem to be confused by it (not expecting `<testcase>` in `<testsuite>` (!), and the only thing that didn't was xsdvi, which transforms schemas to SVG. At least my eyes and my brain-parser seem to think this should be at least be consumed, even if the reporting might not actually be helpful.

@ethomson If you don't mind and can afford it, I'd love to have the opinion of VSTS on that XML.

ethomson:
> @ethomson If you don't mind and can afford it, I'd love to have the opinion of VSTS on that XML.

Sorry, I missed this notification - can you just push a new commit to this branch (or rebase, or whatever)?  The VSTS CI should pick it up and run with it.

tiennou:
Rebased. Though it seems VSTS wasn't triggered ?

tiennou:
@ethomson No luck there. Is it because it predates the VSTS setup ?

ethomson:
I guess not - maybe the github web hook is configured at PR creation time and since this predates the VSTS CI setup, it will never run?  I opened #4778, let's party in that branch instead.

PR #4742, opened 2018-07-29
revwalk: remove tautologic condition for hiding a commit

neithernut:
The contition cannot be reached with `commit->uninteresting` being true:
either a `break` or a `continue` statement will be hit in this case.


ethomson:
This seems obviously correct, thanks @neithernut.

PR #4743, opened 2018-07-30
Windows: default credentials / fallback credential handling

Agent00Log:
While trying to use pygit2 with my TFS Git I stumbled accross two problems.

Whenever the server reports that it supports the NTLM and Negotiate authentication schemes both of the enum values are combined with the bitwise or assignment and the subsequent call to WinHttpSetCredentials fails.

The second problem was due to the fact that I had the python code running in PythonWin and it seems that PythonWin is setting the apartment state of the thread to single threaded. In this case the call to CoInitializeEx in the fallback credential function fails with the RPC_E_CHANGED_MODE and the fallback default credentials are never set.

### DIFF
--- a/src/transports/winhttp.c
+++ b/src/transports/winhttp.c
@@ -219,6 +219,7 @@ static int fallback_cred_acquire_cb(
 	 * as an authentication mechanism */
 	if (GIT_CREDTYPE_DEFAULT & allowed_types) {
 		wchar_t *wide_url;
+		HRESULT hCoInitResult;
 
 		/* Convert URL to wide characters */
 		if (git__utf8_to_16_alloc(&wide_url, url) < 0) {
@@ -226,7 +227,9 @@ static int fallback_cred_acquire_cb(
 			return -1;
 		}
 
-		if (SUCCEEDED(CoInitializeEx(NULL, COINIT_MULTITHREADED))) {
+		hCoInitResult = CoInitializeEx(NULL, COINIT_MULTITHREADED);
+			
+		if (SUCCEEDED(hCoInitResult) || hCoInitResult == RPC_E_CHANGED_MODE) {
 			IInternetSecurityManager* pISM;
 
 			/* And if the target URI is in the My Computer, Intranet, or Trusted zones */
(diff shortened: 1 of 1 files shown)


Agent00Log:
I don't quite get the problem with the COM initialization. There is no call to `CoInitializeSecurity`.
Once the concurrency model is set it cannot be changed unless fully uninitialized.

As far as I can see there are four different states:
- Not initialized
  - The call to `CoInitializeEx` returns `S_OK` and the concurrency model is set to MTA
  - After we're finished we call `CoUnitialize` and everything is back to normal.
- Already initialized (MTA)
  - The call to `CoInitializeEx` returns `S_FALSE` and the init count went one up.
  - After we're finished we call `CoUnitialize` and the init count is back to the previous value but the COM library is not closed.
- Already initialized (STA)
  - The call to `CoInitializeEx` returns `RPC_E_CHANGED_MODE` which indicates that the thread was previously initialized with a different concurrency model. As far as I can tell we don't require a specific model for our calls to the `IInternetSecurityManager`.
  - After we're finished we **don't** call `CoUninitialize` since the call to `CoInitializeEx` was not successful.
- Any other error:
  - Nothing happens, error code 1 is returned.

The change is to accept that some other code already set the concurrency model to something else and just go with it.

Agent00Log:
😊 I know this one gave me a lot of trouble as well.
I think the description for the return value in the MSDN docs is a little bit off.

In the remarks section of the same [entry](https://docs.microsoft.com/de-de/windows/desktop/api/combaseapi/nf-combaseapi-coinitializeex#remarks) it says:
> When an object that is configured to run in the neutral threaded apartment (NTA) is called by a thread that is in either an STA or the MTA, that thread transfers to the NTA. If this thread subsequently calls CoInitializeEx, the call fails and returns RPC_E_CHANGED_MODE.

This sounds a lot more plausible. Otherwise it would be possible to change the model when calling `CoInitializeEx` inside a neutral object but at the same time get an error that the model could not be changed as the return value.

ethomson:
Cool, thanks @Agent00Log!

PR #4749, opened 2018-08-04
parse: Do not initialize the content in context to NULL

### LINKED ISSUE #4726: git__linenlen calls memchar with null pointer

neithernut:
String operations in libgit2 are supposed to never receive `NULL`, e.g. they are not `NULL`-save. In the case of `git__linenlen()`, invocation with `NULL` leads to undefined behavior.

In a `git_parse_ctx` however, the `content` field used in these operations was initialized to `NULL` if the `git_parse_ctx_init()` was called with `NULL` for `content` or `0` for `content_len`. For the latter case, the initialization function even contained some logic for initializing `content` with `NULL`.

This commit mitigates triggering undefined behavior by rewriting the logic. Now `content` is always initialized to a non-null buffer. Instead of a null buffer, an empty string is used for denoting an empty buffer.

Supersedes #4733. Resolves #4726.

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -8,12 +8,14 @@
 
 int git_parse_ctx_init(git_parse_ctx *ctx, const char *content, size_t content_len)
 {
-	if (content_len)
+	if (content && content_len) {
 		ctx->content = content;
-	else
-		ctx->content = NULL;
+		ctx->content_len = content_len;
+	} else {
+		ctx->content = "";
+		ctx->content_len = 0;
+	}
 
-	ctx->content_len = content_len;
 	ctx->remain = ctx->content;
 	ctx->remain_len = ctx->content_len;
 	ctx->line = ctx->remain;
@@ -26,6 +28,7 @@ int git_parse_ctx_init(git_parse_ctx *ctx, const char *content, size_t content_l
 void git_parse_ctx_clear(git_parse_ctx *ctx)
 {
 	memset(ctx, 0, sizeof(*ctx));
+	ctx->content = "";
 }
 
 void git_parse_advance_line(git_parse_ctx *ctx)


src/parse.c
```diff
@@ -8,12 +8,14 @@
...
-	if (content_len)
+	if (content && content_len) {
 		ctx->content = content;
-	else
-		ctx->content = NULL;
+		ctx->content_len = content_len;
+	} else {
+		ctx->content = "";
+		ctx->content_len = 0;
+	}
```
pks-t: Yeah, this fixes the issue. I'm wondering though if we should even make it a bug to pass in `NULL` to `git_parse_ctx_init` by adding an assert. That'd require us to review all callers, though

neithernut: That was my initial idea. In a (non-published) earlier version, I added an assertion for catching calls to `git_parse_ctx_init` with `NULL` for `context`. However, I quickly ran into places where this fails.

For example, functions for parsing stuff built upon `git_parse_ctx` would also happily pass `NULL` for `context`. I even run into a test case explicitly checking for the ability to parse a patch from an empty _and_ a `NULL` buffer:

https://github.com/libgit2/libgit2/blob/3be7301151e44537d2423386dc71f1d3e496c664/tests/apply/fromfile.c#L448

pks-t: Huh, so be it then. Thanks for clarifying!

PR #4750, opened 2018-08-05
config_file: Don't crash on options without a section

nelhage:
Found via fuzzing.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -1035,6 +1035,10 @@ static int read_on_variable(
 	GIT_UNUSED(line);
 	GIT_UNUSED(line_len);
 
+	if (!current_section) {
+		giterr_set(GITERR_CONFIG, "no section for key: %s", var_name);
+		return -1;
+	}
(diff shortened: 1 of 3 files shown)


ethomson:
Nice find - but git seems to parse configuration values without a section just fine.  I think that actually we should:

    if (current_section) {
        git_buf_puts(&buf, current_section);
        git_buf_putc(&buf, '.');
    }

    for (c = var_name; *c; c++)
        git_buf_putc(&buf, git__tolower(*c));

ethomson:
I suspected it might be a warning with some commands.  (`git config —list` did not warn, but that was the extent of what I ran.)

Would you mind throwing a comment in that suggests that we should warn too?  Once warnings land we can wire it up.

Thanks!

pks-t:
I'm not sure at all what happens in case we parse a config key without section. Could you please add a test that demonstrates that we do not fail at another place and that demonstrates what's expected to happen?

pks-t:
Thanks for your fix and the test!

PR #4751, opened 2018-08-05
Fix a double-free in config parsing

nelhage:


pks-t:
Thanks, this looks correct to me! By the way, I'd kindly ask you to make your commit messages conform to our accepted style of "section: what has happened". E.g. in this case something like "config: fix double-free when parsing multiline comments". But anyway, keep up your good and important work! :)

PR #4752, opened 2018-08-05
Add a fuzzer for config files

nelhage:


fuzzers/config_file_fuzzer.c
```diff
@@ -0,0 +1,41 @@
...
+    }
+    if (lseek(fd, 0, SEEK_SET) != 0) {
+        abort();
+    }
+    if (write(fd, data, size) != size) {
+        abort();
+    }
+
+    git_config *cfg;
+    int err = git_config_open_ondisk(&cfg, path);
```
pks-t: Our project uses the C89 standard, where declarations after statements aren't allowed. You'll have to move all variable declarations to the head of this function

pks-t: This is not your fault, but it really sucks that libgit2 isn't yet able to open configs from memory. I wanted to refactor that part for a long time already to make it possible, but until now I didn't yet have enough time to do this. I bet that writing a file for each iteration will slow the fuzzer down by quite a bit.

fuzzers/config_file_fuzzer.c
```diff
@@ -0,0 +1,41 @@
+#include <git2.h>
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <limits.h>
+
+int foreach_cb(const git_config_entry *entry, void *payload) {
+    return 0;
```
pks-t: Ah, another thing. You'll have to mark these two function parameters as unused, otherwise the compiler will issue a warning. We usually do this by using `GIT__UNUSED(x)`, but we don't have it available here. So you could probably just declare a macro `#define UNUSED(x) (void)(x)` after the includes and then use it here.

fuzzers/config_file_fuzzer.c
```diff
@@ -1,41 +1,67 @@
...
+
+	if (ftruncate(fd, 0) !=0 ) {
+		abort();
+	}
+	if (lseek(fd, 0, SEEK_SET) != 0) {
+		abort();
+	}
+	if ((size_t)write(fd, data, size) != size) {
+		abort();
+	}
```
pks-t: One more thing. It could very well be that the fuzzer throws huge files at us, in which case it is likely that we abort here due to the write being interrupted. So we'd need a full-blown write loop, like:

```
size_t total = 0;
while (total < size) {
    ssize_t written = write(fd, data, size);
    if (written < 0 && errno != EINTR)
        abort();
    if (written < 0)
        continue;
    total += written;
}
```

nelhage:
Awesome, I've subscribed to that PR and once it lands I'll send in a refactor. I'll definitely be glad to ditch the tempfile.

Issue #4753, opened 2018-08-05
`./libgit2_clar -sthreads::diff` sometimes segfaults on all architectures


infinity0:
This supersedes #4737 which was misdiagnosed to be architecture-specific; in fact the segfault happens on all architectures.

~~~~
$ mkdir build && cd build
$ cmake
$ make -j$(nproc)
$ while ./libgit2_clar -sthreads::diff >/dev/null; do true; done
Segmentation fault
~~~~

Backtrace:
~~~~
$ ulimit -c unlimited
$ while ./libgit2_clar -sthreads::diff >/dev/null; do true; done
Segmentation fault (core dumped)
$ cat /proc/sys/kernel/core_pattern
core
$ find /tmp -name core
[..redacted..]
/tmp/libgit2_tests_nQCj2Y/core
1
$ gdb ./libgit2_clar /tmp/libgit2_tests_nQCj2Y/core
GNU gdb (Debian 8.1-4) 8.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./libgit2_clar...done.
[New LWP 6422]
[New LWP 6423]
[New LWP 6421]
[New LWP 6419]
[New LWP 6417]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `./libgit2_clar -sthreads::diff'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  prefixcmp (str=0x7f28a000cf82 "\374\374\374\374\374\374u", str_n=18446744073709551614, prefix=0x326a43516e5f7375 <error: Cannot access memory at address 0x326a43516e5f7375>, icase=false)
    at $HOME/libgit2/src/util.c:263
263			p = (unsigned char)*prefix++;
[...]

PR #4754, opened 2018-08-05
threads::diff: use separate git_repository objects

ethomson:
Our thread policies state that we cannot re-use the `git_repository`
across threads.  Our tests cannot deviate from that.

### DIFF
--- a/tests/threads/diff.c
+++ b/tests/threads/diff.c
@@ -76,26 +76,29 @@ static void free_trees(void)
 static void *run_index_diffs(void *arg)
 {
 	int thread = *(int *)arg;
+	git_repository *repo;
 	git_diff_options opts = GIT_DIFF_OPTIONS_INIT;
 	git_diff *diff = NULL;
 	size_t i;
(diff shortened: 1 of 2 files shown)


ethomson:
Fixes #4753

ethomson:
I considered simply deleting these tests - it feels rather obvious that this should always succeed now that we're using multiple repository objects.  But after more consideration, this does exercise the object cache, etc, and it seems reasonable to keep it.

pks-t:
Does it really? I thought the object cache was per repository.

infinity0:
This improves the situation but doesn't get rid of the segfault completely, I posted another stack trace in #4753.

carlosmn:
The object caches are per-repository but the packfiles and their caches are shared. I don't see how we would fail on ignore rules unless we're grabbing from a global cache of built-in rules or mismanaging the lifetime of these objects.

ethomson:
Thanks, @infinity0, I've added your change from #4753.

PR #4755, opened 2018-08-06
travis: do not execute Coverity analysis for all cron jobs

pks-t:
The new Travis cron job gets executed daily, but our current
configuration will cause each job to execute our Coverity script instead
of the default build and testing scripts. This cannot work, as Coverity
is heavily rate-limiting its API, so our cron builds are doomed to
always fail. What we want to do instead is execute our normal builds,
but add an additional Coverity jobs.

This can easily be done by adding another Coverity-specific job with a
conditional "type = cron", which sets the "COVERITY" environment
variable. Instead of checking the build type, we then simply check
whether "COVERITY" is set or not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,8 +65,12 @@ matrix:
        - MBEDTLS=1
          CMAKE_OPTIONS="-DTHREADSAFE=OFF -DBUILD_EXAMPLES=ON -DUSE_HTTPS=mbedTLS -DSHA1_BACKEND=mbedTLS -DMBEDTLS_ROOT_DIR=/tmp/mbedtls"
      os: linux
- allow_failures:
-   - env: COVERITY=1
+   - compiler: gcc
+     if: type = cron
+     env:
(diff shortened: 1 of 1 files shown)


ethomson:
Ugh.  Sorry I got this wrong - I thought that we would ignore the rate limited response.  Since we're running this nightly on both travis and vsts, we may still fail (at least) one of them.  But let's try this and see.

I wish we could execute completely separate scripts between the nightly cron and the PR builds.

pks-t:
On Mon, Aug 06, 2018 at 05:14:03AM -0700, Edward Thomson wrote:
> Ugh.  Sorry I got this wrong - I thought that we would ignore
> the rate limited response.  Since we're running this nightly on
> both travis and vsts, we may still fail (at least) one of them.
> But let's try this and see.
> 
> I wish we could execute completely separate scripts between the
> nightly cron and the PR builds.

I do think that two jobs should be fine. If I recall correctly,
the daily rate limit throttled at around 4-5 runs.

PR #4757, opened 2018-08-06
Release v0.26.6

pks-t:
This is a security release fixing out-of-bounds reads when
processing smart-protocol "ng" packets.

When parsing an "ng" packet, we keep track of both the current position
as well as the remaining length of the packet itself. But instead of
taking care not to exceed the length, we pass the current pointer's
position to `strchr`, which will search for a certain character until
hitting NUL. It is thus possible to create a crafted packet which
doesn't contain a NUL byte to trigger an out-of-bounds read.

The issue was discovered by the oss-fuzz project, issue 9406.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ addons:
     - valgrind
 
 sudo: false
+osx_image: xcode8.3
 
 matrix:
  fast_finish: true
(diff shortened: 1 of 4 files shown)


pks-t:
@ethomson: this is the reason I first publish these PRs over here ;) Both PRs are failing on macOS due to changes in the build environment (I think?).

pks-t:
Yup: https://blog.travis-ci.com/2018-07-19-xcode9-4-default-announce. I'm going to force the old Xcode 8.3 images for these releases.

PR #4763, opened 2018-08-11
Fix 'invalid packet line' for ng packets containing errors

### LINKED ISSUE #4762: v0.27.4: invalid packet line
### Reproduction steps

```
error = git_remote_push(remote, &refs, &push_options);
```

Server response (github.com):

```
0013000eunpack ok
00380033ng refs/pull/1/head deny updating a hidden ref
00380033ng refs/pull/4/head deny updating a hidden ref
001e0019ok refs/heads/master
000900000000
```

### Expected behavior

`error == 0`
### Actual behavior

```c
if (error == GITERR_NET) {
  git_error *g_error = giterr_last();
  printf("NetworkError: %s", g_error->message);
}
```

=> `NetworkError: invalid packet line`

v0.27.4 _(v0.27.3 works)_

linux, MacOS

cschlack:
fixes #4762

### DIFF
--- a/src/transports/smart_pkt.c
+++ b/src/transports/smart_pkt.c
@@ -282,7 +282,7 @@ static int ok_pkt(git_pkt **out, const char *line, size_t len)
 static int ng_pkt(git_pkt **out, const char *line, size_t len)
 {
 	git_pkt_ng *pkt;
-	const char *ptr;
+	const char *ptr, *eol;
 	size_t alloclen;
 
 	pkt = git__malloc(sizeof(*pkt));
@@ -291,11 +291,13 @@ static int ng_pkt(git_pkt **out, const char *line, size_t len)
 	pkt->ref = NULL;
 	pkt->type = GIT_PKT_NG;
 
+	eol = line + len;
+
 	if (len < 3)
 		goto out_err;
 	line += 3; /* skip "ng " */
-	len -= 3;
-	if (!(ptr = memchr(line, ' ', len)))
+
+	if (!(ptr = memchr(line, ' ', eol - line)))
 		goto out_err;
 	len = ptr - line;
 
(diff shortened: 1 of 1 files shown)


src/transports/smart_pkt.c
```diff
@@ -291,11 +291,13 @@ static int ng_pkt(git_pkt **out, const char *line, size_t len)
...
+
 	if (len < 3)
 		goto out_err;
 	line += 3; /* skip "ng " */
-	len -= 3;
-	if (!(ptr = memchr(line, ' ', len)))
+
+	if (!(ptr = memchr(line, ' ', eol - line)))
 		goto out_err;
 	len = ptr - line;
```
pks-t: This line is the problem. We overwrite `len` with the length of the reference and thus loose track of the overall line length. Thus the following computations on `len` are invalid when we try to see how much characters are left in the pkt-line.

src/transports/smart_pkt.c
```diff
@@ -309,8 +311,8 @@ static int ng_pkt(git_pkt **out, const char *line, size_t len)
 	if (len < 1)
 		goto out_err;
 	line = ptr + 1;
-	len -= 1;
-	if (!(ptr = memchr(line, '\n', len)))
+
+	if (!(ptr = memchr(line, '\n', eol - line)))
```
pks-t: I think I prefer your `eol`-pointer thing. It reliefs us from always updating the `len` field, and thus is less error-prone.

src/transports/smart_pkt.c
```diff
@@ -309,8 +311,8 @@ static int ng_pkt(git_pkt **out, const char *line, size_t len)
 	if (len < 1)
```
pks-t: This line doesn't make a lot of sense in this context then, either, does it?

```
line = ptr + 1;
if (line >= eol)
    goto out_err;
```

Would make a lot more sense.

cschlack: Thank you @pks-t , that was missing! I added it to my previous commit.

cschlack:
@pks-t I believe I fixed your requested change. Is there anything else you'd like me to change?

pks-t:
No, looks good to me. Thanks for your fix!

PR #4766, opened 2018-08-16
travis: remove Coverity cron job

pks-t:
With the recent addition of VSTS to our CI infrastructure, we now have
two cron jobs running regular Coverity analysis. It doesn't really make
a lot of sense to upload two different analysis on our sources to
Corverity, though:

- in the worst case, Coverity will be repeatedly confused when different
  sets of sources get analyzed and uploaded
- in the best case, nothing is gained because the sources have already
  been analyzed via the other job

Let's just use a single cron job for Coverity. Considering that VSTS
seems to be the more beefy and flexible platform, it is more likely to
be our future target CI platform. Thus, we retain its support for
Coverity and instead remove it from Travis.


ethomson:
Nice one, thanks.

PR #4767, opened 2018-08-16
In-memory configuration

pks-t:
So finally I hold true to my promise to refactor our config code (once again) and create an in-memory configuration backend. I think the result is quite pleasant due to the refactorings, and in fact the in-memory backend implementation is only a bit more than a hundred lines of code (excluding the stubs).

This series has three phases:

1. Refactorings to make our config file more generic with regards to names. Most importantly, many things are renamed from "git_config_file_*" to "git_config_backend_*" to express that they are actually backend-agnostic.
2. I pull out the "config_entries" store from the "git_config_file" backend that abstracts iteration order and memory management for config entries.
3. I create the "git_config_mem" backend based on that store, which is rather trivial to do as everything is in place already.

There are a few more things I'd like to do after this PR is merged. First, I think that we can get rid of the file read-only snapshot implementation and simply use an in-memory backend instead. Second, I'll implement writing to and snapshotting of the in-memory config backend. Both things should be easy to implement, but the series is already long enough and thus I refrained from doing so now.

As always, I tried to make sure that the series "develops" to make a pleasant read. So I highly recommend to read it commit-wise instead of all at once.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -31,15 +31,15 @@ void git_config_entry_free(git_config_entry *entry)
 typedef struct {
 	git_refcount rc;
 
-	git_config_backend *file;
+	git_config_backend *backend;
 	git_config_level_t level;
-} file_internal;
+} backend_internal;
 
-static void file_internal_free(file_internal *internal)
+static void backend_internal_free(backend_internal *internal)
 {
 	git_config_backend *file;
 
-	file = internal->file;
+	file = internal->backend;
 	file->free(file);
 	git__free(internal);
 }
@@ -47,11 +47,11 @@ static void file_internal_free(file_internal *internal)
 static void config_free(git_config *cfg)
 {
 	size_t i;
-	file_internal *internal;
+	backend_internal *internal;
 
 	for (i = 0; i < cfg->backends.length; ++i) {
 		internal = git_vector_get(&cfg->backends, i);
-		GIT_REFCOUNT_DEC(internal, file_internal_free);
+		GIT_REFCOUNT_DEC(internal, backend_internal_free);
 	}
 
 	git_vector_free(&cfg->backends);
@@ -70,8 +70,8 @@ void git_config_free(git_config *cfg)
 
 static int config_backend_cmp(const void *a, const void *b)
 {
-	const file_internal *bk_a = (const file_internal *)(a);
-	const file_internal *bk_b = (const file_internal *)(b);
+	const backend_internal *bk_a = (const backend_internal *)(a);
+	const backend_internal *bk_b = (const backend_internal *)(b);
 
 	return bk_b->level - bk_a->level;
 }
@@ -151,7 +151,7 @@ int git_config_snapshot(git_config **out, git_config *in)
 {
 	int error = 0;
 	size_t i;
-	file_internal *internal;
+	backend_internal *internal;
 	git_config *config;
 
 	*out = NULL;
@@ -162,7 +162,7 @@ int git_config_snapshot(git_config **out, git_config *in)
 	git_vector_foreach(&in->backends, i, internal) {
 		git_config_backend *b;
 
-		if ((error = internal->file->snapshot(&b, internal->file)) < 0)
+		if ((error = internal->backend->snapshot(&b, internal->backend)) < 0)
 			break;
 
 		if ((error = git_config_add_backend(config, b, internal->level, NULL, 0)) < 0) {
@@ -180,12 +180,12 @@ int git_config_snapshot(git_config **out, git_config *in)
 }
 
 static int find_internal_file_by_level(
-	file_internal **internal_out,
+	backend_internal **internal_out,
 	const git_config *cfg,
 	git_config_level_t level)
 {
 	int pos = -1;
-	file_internal *internal;
+	backend_internal *internal;
 	size_t i;
 
 	/* when passing GIT_CONFIG_HIGHEST_LEVEL, the idea is to get the config backend
@@ -215,7 +215,7 @@ static int find_internal_file_by_level(
 
 static int duplicate_level(void **old_raw, void *new_raw)
 {
-	file_internal **old = (file_internal **)old_raw;
+	backend_internal **old = (backend_internal **)old_raw;
 
 	GIT_UNUSED(new_raw);
 
@@ -228,7 +228,7 @@ static void try_remove_existing_file_internal(
 	git_config_level_t level)
 {
 	int pos = -1;
-	file_internal *internal;
+	backend_internal *internal;
 	size_t i;
 
 	git_vector_foreach(&cfg->backends, i, internal) {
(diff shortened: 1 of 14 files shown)


src/config_entries.c
```diff
@@ -7,11 +7,22 @@
...
+	struct config_entry_list *next;
+	git_config_entry *entry;
+} config_entry_list;
+
 typedef struct config_entries_iterator {
 	git_config_iterator parent;
 	config_entry_list *head;
 } config_entries_iterator;
 
+struct git_config_entries{
```
tiennou: nitpick: missing space before {

src/config_mem.c
```diff
@@ -0,0 +1,223 @@
...
+	GIT_UNUSED(backend);
+	GIT_UNUSED(name);
+	GIT_UNUSED(regexp);
+	return config_error_readonly();
+}
+
+static int config_memory_lock(git_config_backend *backend)
+{
+	GIT_UNUSED(backend);
+	return config_error_readonly();
```
tiennou: I have a fuzzy recollection that `lock`/`unlock` can be used by transactions, and that `snapshot` can also be about having a consistent view of the configuration at some point in time (even when read-only).

I just want to point that out, though AFAICT this is mostly intended for testing purposes 😉. Support for those can be added later.

src/config_mem.c
```diff
@@ -0,0 +1,223 @@
...
+
+typedef struct {
+	git_config_entries *entries;
+	git_config_level_t level;
+} config_memory_parse_data;
+
+static int config_error_readonly(void)
+{
+	giterr_set(GITERR_CONFIG, "this backend is read-only");
+	return -1;
```
tiennou: Still longing for a `GIT_ENOTSUP`…

src/config_file.c
```diff
@@ -229,17 +229,20 @@ static int config_iterator_new(
...
-	git_config_backend *snapshot;
 	diskfile_header *bh = (diskfile_header *) backend;
+	git_config_entries *entries;
 	int error;
 
-	if ((error = config_snapshot(&snapshot, backend)) < 0)
+	if ((error = git_config_entries_dup(&entries, bh->entries)) < 0)
 		return error;
 
-	if ((error = snapshot->open(snapshot, bh->level, bh->repo)) < 0)
```
tiennou: 🎉

src/config_mem.c
ethomson: Hmm, interesting.  I'm game for an `ENOTSUPPORTED` or whatever, but I worry that it would get really weird really fast, at least here for config.  Since configuration is so low-level, I worry that it would leak out very strangely.  For example, if I called `git_merge` and got an `GIT_ENOTSUPPORTED` back then I would very much not understand what it was that wasn't supported.

I think?  Is this actually the case that you had in mind that you do in fact want?

tiennou: Well it could be mostly given as higher-level layers that the current thing they're trying to do can't work, that would allow that higher-level to preserve the semantics of what happened.

For example, we don't do octopus merges, that can be used to signal that to the user (instead of the generic error we're currently returning).
In the config case, the higher level could try other backends and either swallow the error if another one was able to handle the request or error (as currently done), or return `ENOTSUPPORTED` to be more specific about the failure (in case there's only one in-memory backend loaded, which the user is ultimately responsible for). Which makes it pretty close to the `ENOTFOUND` semantics some codepaths have.

This is pretty minor though, it's just that there are some places where it would maybe be useful to report to the user that they're facing a "limitation" of the API, not a hard error.

src/config_mem.c
pks-t: I wanted to keep this PR as short as possible while still providing basic functionality. I plan on implementing write support and snapshotting as soon as this is merged, especially so as both things should be rather trivial to implement. I didn't yet take a look at lock/unlock and transactions, but I'll do that at the same point in time, I guess.

src/config_mem.c
pks-t: I agree that `ENOTSUPPORTED` would be a nice to have in some locations, including this one. The PR is already big enough as is right now, if you ask me, so I'd love to postpone this to another PR. Especially so as it might require reviewing of callers, as noted by @ethomson

pks-t:
Thanks for your feedback. I've fixed the stylistic issue spotted by @tiennou

src/config_mem.c
tiennou: It was just wishful thinking 😉. Since it seems we find that a worthwhile error code, I'll try to create a PR for it.

pks-t:
Rebased to fix conflicts with #4799

src/config_mem.c
carlosmn: I'm skeptical of run-time feature detection like this. Do we have something that isn't detectable at compile time and isn't just a missing feature or a bug?

src/config_mem.c
```diff
@@ -0,0 +1,223 @@
...
+	GIT_UNUSED(backend);
+	GIT_UNUSED(success);
+	return config_error_readonly();
+}
+
+static int config_memory_snapshot(git_config_backend **out, git_config_backend *backend)
+{
+	GIT_UNUSED(out);
+	GIT_UNUSED(backend);
+	return config_error_readonly();
```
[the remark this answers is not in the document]
pks-t: True. I've been too eager to just reuse that nifty function.

Also referring to the previous comment of yours, I intend to create a follow-up PR as soon as this is created. Snapshotting is trivial to implement, as you said, but I first wanted to get going the refactoring as well as basic functionality. I definitely intend to implement snapshotting and potentially even write support for the next release

src/config.c
```diff
@@ -179,92 +179,92 @@ int git_config_snapshot(git_config **out, git_config *in)
...
 
 static int duplicate_level(void **old_raw, void *new_raw)
 {
-	file_internal **old = (file_internal **)old_raw;
+	backend_internal **old = (backend_internal **)old_raw;
 
 	GIT_UNUSED(new_raw);
 
-	giterr_set(GITERR_CONFIG, "a file with the same level (%i) has already been added to the config", (int)(*old)->level);
+	giterr_set(GITERR_CONFIG, "a backend with the same level (%i) has already been added to the config", (int)(*old)->level);
```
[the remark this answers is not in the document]
pks-t: What about just using "configuration" instead of backend?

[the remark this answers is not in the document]
pks-t: No, that sounds weird when reading the complete sentence. "a configuration with the same level (%i) already exists"?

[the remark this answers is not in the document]
carlosmn: We could have "configuration at this level (%i) is already set" or "there is already configuration at this level (%i)" which parse better.

pks-t:
Amended once again to improve the error message. Seeing that the criticism only revolved around such minor things, can I assume that this is ready to be merged?

PR #4768, opened 2018-08-16
Fix leak in index.c

abyss7:


ethomson:
LGTM.  Thanks, @abyss7.

PR #4769, opened 2018-08-16
worktree: unlock should return 1 when the worktree isn't locked

tiennou:
The documentation states that git_worktree_unlock returns 0 on success,
and 1 on success if the worktree wasn't locked. Turns out we were
returning 0 in any of those cases.

Found it while writing test cases for Objective-Git.


pks-t:
While I would've designed the interface differently by now (returning `GIT_ELOCKED`, which we already have), this at least fixes code to match our documentation. So thanks for discovering and fixing it!

PR #4770, opened 2018-08-17
Allow merge analysis against any reference

tiennou:
This adds a new `git_merge_analysis_any` function that can be used to check a merge against any given reference, as right now we only support checking `HEAD`.

Inspired by https://github.com/libgit2/objective-git/pull/629.

### DIFF
--- a/include/git2/merge.h
+++ b/include/git2/merge.h
@@ -389,6 +389,25 @@ GIT_EXTERN(int) git_merge_analysis(
 	size_t their_heads_len);
 
 /**
+ * Analyzes the given branch(es) and determines the opportunities for
+ * merging them into a reference.
+ *
+ * @param analysis_out analysis enumeration that the result is written into
+ * @param repo the repository to merge
+ * @param our_ref the reference to perform the analysis from
+ * @param their_heads the heads to merge into
+ * @param their_heads_len the number of heads to merge
+ * @return 0 on success or error code
+ */
+GIT_EXTERN(int) git_merge_analysis_for_ref(
+	git_merge_analysis_t *analysis_out,
+	git_merge_preference_t *preference_out,
+	git_repository *repo,
+	git_reference *our_ref,
+	const git_annotated_commit **their_heads,
+	size_t their_heads_len);
+
+/**
  * Find a merge base between two commits
  *
  * @param out the OID of a merge base between 'one' and 'two'
--- a/src/refs.c
+++ b/src/refs.c
@@ -1430,3 +1430,27 @@ const char *git_reference_shorthand(const git_reference *ref)
 {
 	return git_reference__shorthand(ref->name);
 }
+
+int git_reference__is_unborn_head(bool *unborn, const git_reference *ref, git_repository *repo)
+{
+	int error;
+	git_reference *tmp_ref;
+	assert(unborn && ref && repo);
+
+	if (ref->type == GIT_REF_OID) {
+		*unborn = 0;
+		return 0;
+	}
+
+	error = git_reference_lookup_resolved(&tmp_ref, repo, ref->name, -1);
+	git_reference_free(tmp_ref);
+
+	if (error != 0 && error != GIT_ENOTFOUND)
+		return error;
+	else if (error == GIT_ENOTFOUND && git__strcmp(ref->name, GIT_HEAD_FILE) == 0)
+		*unborn = true;
+	else
+		*unborn = false;
+
+	return 0;
+}
--- a/src/refs.h
+++ b/src/refs.h
@@ -136,4 +136,6 @@ int git_reference__update_for_commit(
 	const git_oid *id,
 	const char *operation);
 
+int git_reference__is_unborn_head(bool *unborn, const git_reference *ref, git_repository *repo);
+
 #endif
--- a/tests/merge/workdir/analysis.c
+++ b/tests/merge/workdir/analysis.c
@@ -40,21 +40,33 @@ void test_merge_workdir_analysis__cleanup(void)
 static void analysis_from_branch(
 	git_merge_analysis_t *merge_analysis,
 	git_merge_preference_t *merge_pref,
-	const char *branchname)
+	const char *our_branchname,
+	const char *their_branchname)
 {
-	git_buf refname = GIT_BUF_INIT;
+	git_buf our_refname = GIT_BUF_INIT;
+	git_buf their_refname = GIT_BUF_INIT;
+	git_reference *our_ref;
 	git_reference *their_ref;
 	git_annotated_commit *their_head;
 
-	git_buf_printf(&refname, "%s%s", GIT_REFS_HEADS_DIR, branchname);
+	if (our_branchname != NULL) {
+		cl_git_pass(git_buf_printf(&our_refname, "%s%s", GIT_REFS_HEADS_DIR, our_branchname));
+		cl_git_pass(git_reference_lookup(&our_ref, repo, git_buf_cstr(&our_refname)));
+	} else {
+		cl_git_pass(git_reference_lookup(&our_ref, repo, GIT_HEAD_FILE));
+	}
 
-	cl_git_pass(git_reference_lookup(&their_ref, repo, git_buf_cstr(&refname)));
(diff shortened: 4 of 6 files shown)


src/refs.h
```diff
@@ -136,4 +136,16 @@ int git_reference__update_for_commit(
...
+ * Lookup a reference, with unborn error semantics.
+ *
+ * @param out_ref Pointer to the resolved reference. Can be null.
+ * @param repo The repository to use for resolution.
+ * @param ref The reference to resolve.
+ * @return 0 on success and out_ref (if provided) will be populated.
+ *         Otherwise, returns GIT_EINVALID if the reference is not symbolic,
+ *         GIT_EUNBORNBRANCH if the reference is unborn, or another error code.
+ */
+int git_reference__resolve_unborn(git_reference **out_ref, git_repository *repo, const git_reference *ref);
```
pks-t: What about `git_reference__resolve_symbolic`? `resolve_unborn` sounds a lot like it is only able to resolve unborn branches, which is not the case. It is rather a function which will resolve symbolic refs only and which doe not care whether the target is unborn or not.

include/git2/merge.h
```diff
@@ -388,6 +388,25 @@ GIT_EXTERN(int) git_merge_analysis(
...
+ * merging them into a reference.
+ *
+ * @param analysis_out analysis enumeration that the result is written into
+ * @param repo the repository to merge
+ * @param their_heads the heads to merge into
+ * @param their_heads_len the number of heads to merge
+ * @param our_ref the reference to perform the analysis from
+ * @return 0 on success or error code
+ */
+GIT_EXTERN(int) git_merge_analysis_any(
```
pks-t: `git_merge_analysis_with_ref`?  `any` is much too generic in my opinion.

I also wonder about whether one should pass in `our_ref` as an annotated commit, as well. All the others are already annotated commits, and the first thing we do is to look up the annotated commit from the ref.

src/refs.c
```diff
@@ -1455,3 +1455,18 @@ int git_reference__resolve_unborn(git_reference **out_ref, git_repository *repo,
...
+	assert(ref && repo);
+
+	if (ref->type == GIT_REF_OID)
+		return 0;
+
+	error = git_reference__resolve_unborn(NULL, repo, ref);
+	if (error != 0 && error != GIT_EUNBORNBRANCH)
+		return error;
+
+	return error == GIT_EUNBORNBRANCH ? true : false;
```
pks-t: I think we usually avoid reusing the return code as value and error code. So maybe use an out parameter instead to return the boolean?

pks-t:
This is definitely a good thing to have, so I'm sold!

include/git2/merge.h
tiennou: (Argument documentation out-of-order. Also, a vararg version might be nice-to-have)

I can change the name, yes, but maybe if we make the function completely generic it's less of a problem… I'm on really shaky ground w.r.t what's considered valid mergeable objects, so I don't know if going full-on with annotated commits makes sense (like `annotated_from_fetchhead`).

I'd prefer to just switch the type of `our_ref` and keep mandatory arguments explicit.

Note that, since I've been doing some "customer support", AFAICT we're lacking a generic "annotated" `git_merge` function anyways, so the benefits to a fully generic-annotated version are minor.

tiennou:
Rebased. I'll try to see if I can make merge handle any base branch without much work, so please don't merge as is.

pks-t:
Okay, I'll hold off my review until then

tiennou:
Rebased, and ready for review. I won't do `git_merge_for_ref`, because I feel it requires too much fiddling around (eg. `MERGE_HEAD` and things), and doesn't look as useful as the use case for analyzing merges.

tests/merge/workdir/analysis.c
```diff
@@ -40,21 +40,33 @@ void test_merge_workdir_analysis__cleanup(void)
...
-	git_buf refname = GIT_BUF_INIT;
+	git_buf our_refname = GIT_BUF_INIT;
+	git_buf their_refname = GIT_BUF_INIT;
+	git_reference *our_ref;
 	git_reference *their_ref;
 	git_annotated_commit *their_head;
 
-	git_buf_printf(&refname, "%s%s", GIT_REFS_HEADS_DIR, branchname);
+	if (our_branchname != NULL) {
+		git_buf_printf(&our_refname, "%s%s", GIT_REFS_HEADS_DIR, our_branchname);
```
pks-t: This is missing a `cl_git_pass`

tests/merge/workdir/analysis.c
```diff
@@ -40,21 +40,33 @@ void test_merge_workdir_analysis__cleanup(void)
...
+		cl_git_pass(git_reference_lookup(&our_ref, repo, git_buf_cstr(&our_refname)));
+	} else {
+		cl_git_pass(git_reference_lookup(&our_ref, repo, GIT_HEAD_FILE));
+	}
 
-	cl_git_pass(git_reference_lookup(&their_ref, repo, git_buf_cstr(&refname)));
+	git_buf_printf(&their_refname, "%s%s", GIT_REFS_HEADS_DIR, their_branchname);
```
pks-t: This one, too

src/refs.c
```diff
@@ -1430,3 +1430,25 @@ const char *git_reference_shorthand(const git_reference *ref)
...
+		return 0;
+	}
+
+	error = git_reference_lookup_resolved(&tmp_ref, repo, ref->name, -1);
+	git_reference_free(tmp_ref);
+
+	if (error != 0 && error != GIT_ENOTFOUND)
+		return error;
+
+	*unborn = (git__strcmp(ref->name, GIT_HEAD_FILE) == 0 && error == GIT_ENOTFOUND ? true : false);
```
pks-t:
```
if (error != 0 && error != GIT_ENOTFOUND)
    return error;
else if (error == GIT_ENOTFOUND && git__strcmp(ref->name, GIT_HEAD_FILE) == 0)
    *unborn = true;
else
    *unborn = false;
```

I think this would be easier to digest

tiennou:
Fixed and rebased. Thanks for the review @pks-t !

src/refs.c
```diff
@@ -1430,3 +1430,27 @@ const char *git_reference_shorthand(const git_reference *ref)
...
+
+	error = git_reference_lookup_resolved(&tmp_ref, repo, ref->name, -1);
+	git_reference_free(tmp_ref);
+
+	if (error != 0 && error != GIT_ENOTFOUND)
+		return error;
+	else if (error == GIT_ENOTFOUND && git__strcmp(ref->name, GIT_HEAD_FILE) == 0)
+		*unborn = true;
+	else
+		*unborn = false;
```
ethomson: Just making sure that I understand the decisions from #4780.  I'm thinking that at present that if we encounter a broken symref that's _not_ `HEAD` that we will return that unborn is false?    Should this block look like:

```
if (error == GIT_ENOTFOUND && git__strcmp(ref->name, GIT_HEAD_FILE) == 0)
    *unborn = true;
else if (!error)
    *unborn = false;

return error;
```

(Ignoring the style around ifs and such, just trying to make sure that I understand the logic.)

tiennou: Yes, that's true, a broken symref whose name is not `HEAD` will not be considered unborn, and will not error (at least here). This is the check we're doing `git_repository_head` BTW, with the `HEAD` check made more explicit.

I find your proposed block harder to reason about w.r.t to `error`, but I think it still returns errors for cases that would be fine (the 1st condition would still be `GIT_ENOTFOUND` at least). Something like this, if the inline return is not okay :

```c
if (error == 0)
    *unborn = false;
else if (error == GIT_ENOTFOUND && git__strcmp(ref->name, GIT_HEAD_FILE) == 0) {
    *unborn = true;
    error = 0;
} else
    /* error will fallthrough */
```

Your version is a step after this one, so now I get it 😉. Yours is only missing the `error = 0` in the "unborn" handler.

ethomson: Yes, you're very much correct, I was missing that error reset.

pks-t:
I'm happy with this PR now, thank you @tiennou! @ethomson, are you good, too?

pks-t:
Thanks again, @tiennou!

PR #4771, opened 2018-08-17
cmake: allow force-erroring specific warning types

tiennou:
This add an optional parameter to ENABLE_WARNINGS so some of our default warnings can be always considered as errors, even without running with ENABLE_WERROR=ON set.

Use it on declaration-after-statement, because it's the one that I can't stop stumbling upon.


pks-t:
I'm always a bit hesitant to enable Werror-like options for the compiler by default, as they can prove to be a major burden for downstream distributions. E.g. when they switch/upgrade compilers, code that was previously warning free may now produce warnings as heuristics have changed, which breaks the build for them if they're specified via Werror. That being said, this here should be kind of safe to enable by default.

What I'm wondering about though is why you don't want to use "ENABLE_WERROR=ON". CI will build with this switch turned on anyway, so code that doesn't compile with it turned on won't get accepted anyway. Are there warnings in your build that keep you from doing so?

tiennou:
I can understand the rationale, but that pain point is really specific IMHO. What's happening is that I got really mad at Xcode for not even *warning* me about declaration-after-statement, even though it's a staple of our C90 support, and thus only discovered that error when CI did, even though it's now "too late" and need a re-run. This is just this warning, to catch this type of "unacceptable code" early (ie. on the contributor's machine).

Arguably, maybe I should be writing code with ENABLE_WERROR=ON anyways, but I fear it's gonna be a bother me in some other ways, in case I'm haphazardly casting pointers around… And it still doesn't work for contributors.

And yes, I have some warnings that show up locally 😨, one on macOS (cmp-always-true), one on iOS (invalid-fmt-string).

pks-t:
So you say you don't even get a warning for declaration-after-statement "-Wdeclaration-after-statement", but there is an error when you use "-Werror=declaration-after-statement"? That's weird.

I sometimes thought about adding a developer mode, similar to what libgit2 has. It would enable certain things like Werror by default, and also for example debug symbols and turn off optimization (or use "-Og"). That would enable us to simply point new contributors to that single switch. I'm not sure if it's worth it, though.

So I'm not opposed to turning this specific warning into an error, as I think it's reasonable and cannot really trigger errors in downstream users (except for when they patch libgit2 with non-conforming code). But I'll let somebody else decide, as personally I'm not too eager to include it. @ethomson?

neithernut:
> I sometimes thought about adding a developer mode, similar to what libgit2 has. It would enable certain things like Werror by default, and also for example debug symbols and turn off optimization (or use "-Og"). That would enable us to simply point new contributors to that single switch. I'm not sure if it's worth it, though.

Isn't this what `CMAKE_BUILD_TYPE=Debug` is for? It should be pretty straight forward to add some compiler flags for debug builds. And flags like `-g` are enabled in debug builds by cmake anyway.

pks-t:
> Isn't this what CMAKE_BUILD_TYPE=Debug is for?

I'd say no. In my opinion, build type and set of flags used by developers are two orthogonal concepts. For example, I might want to do a release build while still keeping intact additional warnings for developers. I already had it happen multiple times that compilers output different messages depending on optimization, so conflating these two things might've swallowed these (mostly valid) additional repots in release builds.

neithernut:
Isn't that intended to provide warnings from the `CMakeLists.txt` rather than from the actual build?

Issue #4772, opened 2018-08-18
We should document that `giterr_last()` should not be called except after an error


KizzyCode:
### Reproduction steps
 1. Create a new repository
 2. Create an initial commit
 3. Check `giterr_last()` which returns an error even if everything was successful

```c
// Init libgit
assert(git_libgit2_init() == 1);

// Create a new repo
git_repository* repo;
assert(git_repository_init(&repo, "empty", 0) == 0);

// Get index
git_index* index;
assert(git_repository_index(&index, repo) == 0);

// Get tree-ID and tree
git_oid tree_id;
git_index_write_tree(&tree_id, index);

git_tree* tree;
assert(git_tree_lookup(&tree, repo, &tree_id) == 0);

// Create signature
git_signature* signature;
assert(git_signature_now(&signature, "Max Musterfrau", "max@musterfrau.invalid") == 0);

// Check for error
assert(giterr_last() == NULL);

// Create commit
git_oid commit_id;
assert(git_commit_create(
    &commit_id, repo, "HEAD",
    signature, signature, NULL, "Initial commit",
    tree, 0, NULL
) == 0); // <- This works

// Check for error
git_error const* error = giterr_last();
if (error != NULL) printf("Git-error: %s\n", error->message); // <- This prints "Git-error: reference 'refs/heads/master' not found" even if everything was successful
```

### Expected behavior
`giterr_last()` should return `NULL`

### Actual behavior
`giterr_last()` returns an error even if everything was successful

### Version of libgit2 (release number or SHA1)
0.27.4 and c1ef58a

### Operating system(s) tested
macOS High Sierra (17G65)

ethomson:
It shouldn’t, in fact.  If you didn’t get an error return code, you should not check the error string.

There are various errors that occur at low layers of the application (eg file not found) that we ignore and do not bother to clear.

I don’t have a strong interest in changing this behavior, as there’s no benefit, and we will _certainly_ miss cases.

KizzyCode:
Ok, I see your point. Maybe it's a good idea to add this to the documentation?

ethomson:
> Ok, I see your point. Maybe it's a good idea to add this info to the documentation?

Yes you are 100% correct.  Going to reopen this so we can ensure that we do that.

ethomson:
Fixed via #4773.  Thanks for reporting this, and thanks for fixing it @RandomSort.

PR #4773, opened 2018-08-20
Document giterr_last() use only after error. #4772

RandomSort:
Hi, thank you for your awesome work.
This is a trivial documentation addition. 
I'll be happy to rework it if it does not fit your wishes.


ethomson:
I'd like to tweak the wording here just a touch:

> Do not use this method unless you received an error, it will lead to undefined behaviour.

"Undefined behavior" is, I think, a bit strong and suggests to the reader that it's C-like UB where you might crash or demons might fly out of your nose.  Also, I'd say "function" or "API" instead of "method" because we're kicking it old-school C.  Perhaps something more like:

> Do not use this function unless the prior call to a libgit2 API returned an error, as it can give misleading results.  libgit2's error strings are not cleared aggressively, and this function may return an error string that reflects a prior error, possibly even reflecting internal state.

We should also update [the API documentation](https://github.com/libgit2/libgit2/blob/master/include/git2/errors.h#L112) as well, it indicates that `NULL` is always returned when there was no recent error.

RandomSort:
That is a lot better. I will push update.

RandomSort:
I have added API documentation, and reworded as requested. Feel free to prioritize other stuff over this. It is minor change from an outsider :)

docs/error-handling.md
```diff
@@ -117,6 +117,11 @@ The public error API
   the last error.  This may return NULL if no error has occurred.
   Otherwise this should return a `git_error` object indicating the class
   of error and the error message that was generated by the library.
+  Do not use this function unless the prior call to a libgit2 API
+  returned an error, as it can give misleading results.
```
pks-t: "as it can otherwise give misleading results." Just to make sure that it may only be misleading in this exact case ;)

include/git2/errors.h
```diff
@@ -109,7 +109,12 @@ typedef enum {
 
 /**
  * Return the last `git_error` object that was generated for the
- * current thread or NULL if no error has occurred.
+ * current thread.
+ *
+ * The default behaviour of this function is to return NULL if no previous error has occurred.
+ * However, libgit2's error strings are not cleared aggressively, so a prior
+ * (unrelated) error may be returned. This can avoided by only calling
```
pks-t: "This can be avoided". Missing "be"

PR #4776, opened 2018-08-24
tests: verify adding index conflicts with invalid filemodes fails

pks-t:
Commit 581d5492f (Fix leak in index.c, 2018-08-16) was fixing a memory
leak in our code adding conflicts to the index when the added index
entries have an invalid file mode. The memory leak was previously
undiscovered as there are no tests covering this scenario, which is now
being added by this commit.


ethomson:
Thanks!  👍

PR #4778, opened 2018-08-24
Clar XML output redux

ethomson:
Reopening a new PR with @tiennou's work in #4741 so that the CI will pick it up.


ethomson:
I didn't really like that it was writing the XML output to the temporary directory (since it opened the file after being sandboxed).  This is a result of the aggressively early sandbox, so that it can handle `-s` test arguments.

So I changed a bit about the way clar handles `-s` arguments.  Previously, in the arg parser, it would just run the tests when it saw a `-s` and then skip the usual test running that it does.

I made it so that `-s` actually enables the tests that were specified (disabling everything else).  This means that we don't have to set up the sandbox before we parse arguments.  So now if we have a `-r`, we can open the file in the current working directory, then do the sandbox.

I also made `-r` take an optional argument with the filename.  This is because we have five separate `libgit2_clar` invocations in the `CMakeLists.txt` and they should have unique filenames.  We also need to be able to write them into the build output directory since on the Linux builds (which are run in containers) we need to write to a directory that's mapped back to the host.

So there's some success now.  You can see that the although there are linux test failures (because of leaks), they're actually publishing the test results.  So that's neat!

But indeed there are still a few problems: this has introduced some leaks, the macOS tests aren't getting picked up, the Windows XML isn't getting written, and there's a failure in one of the Windows builds.  But it's getting close!

ethomson:
It turns out that the linux failures were because my original take at refactoring `-s` support was far too liberal in enabling tests.  It enabled whole suites, so the ssh tests that turn on a single test would actually run the entire `online::clone` suite.  Oops.

Originally, I had hoped to avoid any allocations, but that was a silly and unnecessary false optimization.  As a result, I refactored the `-s` parsing so that we now store the values in the argument parsing loop and run them later.

ethomson:
I introduced a few memory leaks into clar, which are being detected by valgrind (yay!) but are not failing the build (boo).  I'll have to take a look at this before this is ready to :shipit: .

ethomson:
Okay, I just pushed up a quick test to validate that valgrind was still failing on leaks in the actual product code - it is definitely failing when we "definitely lose" memory.  (But not when allocations are "still reachable" at termination.)

This seems reasonable to me _but_ it appears that the macOS and Windows builds are not failing when they detect memory leaks.  Since this didn't regress, I want to limit the scope of this PR, so I'll try to address this in a separate one soon.

.vsts-ci.yml
```diff
@@ -8,281 +8,108 @@ trigger:
...
-  - powershell: . '$(Build.SourcesDirectory)\ci\test.ps1'
-    displayName: Test
-    workingDirectory: '$(Build.BinariesDirectory)'
-  - task: PublishTestResults@2
-    displayName: Publish Test Results
-    condition: succeededOrFailed()
-    inputs:
-      testResultsFiles: 'results_*.xml'
-      searchFolder: '$(Build.BinariesDirectory)'
-      mergeTestResults: true
```
pks-t: Lovely, all these repetitive job descriptions were bothering me since the inception. Really sice that you can simply use templates instead

tests/clar/summary.h
```diff
@@ -0,0 +1,98 @@
...
+	while (report != NULL) {
+		struct clar_error *error = report->errors;
+
+		if (last_suite == NULL || strcmp(last_suite, report->suite) != 0) {
+			clar_summary_testsuite(0, report->suite, "", time(NULL), 0, _clar.tests_ran, _clar.total_errors, 0);
+		}
+
+		last_suite = report->suite;
+
+		clar_summary_testcase(report->test, "what", 0);
```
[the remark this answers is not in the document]
ethomson: Yes, I agree with that.

tests/clar.c
```diff
@@ -193,21 +204,33 @@ void cl_trace_register(cl_trace_cb *cb, void *payload)
 
 /* Core test functions */
 static void
-clar_report_errors(void)
+clar_report(int *i, struct clar_error *error)
```
pks-t: By the way, this interface is a bit funny, as it's kind of emulating an iterator via this integer. But in my opinion, using this integer makes code less understandable compared to simply iterating over the linked list of errors ourselves.

Feel free to ignore this comment -- it's not that important after all and can easily be cleaned up later

pks-t:
Despite the minor comment I have about the weird iterating interface, the PR looks good to me.

By the way -- do we really have 16000 tests? I cannot quite believe that it's _that_ many tests and tend to think the XML output is to blame.

tests/clar.c
ethomson: Huh.  Thanks for the detailed review, I had missed that.  I agree and will fix it up before I merge it.

ethomson: Fixed!

ethomson: And, actually, I decided that I didn't like the way that I fixed it.  One of the report functions was still assuming knowledge of the clar structure that a caller also had knowledge of, so I simply passed that in as an arg and I think it's quite a bit nicer.  Thanks again for catching this.

ethomson:
I was playing around with some other things but I'll open a new PR if I decide to pursue them.  Shipping this one.

PR #4780, opened 2018-08-25
[WIP] Unborn reference semantics

tiennou:
I've extracted this from #4770 because I don't think it's ready ; some of the points made by @pks-t made me look more closely, and I'm now confused (which is a bad sign). I'm also unclear on what  unbornness means from a technical standpoint, so there's that 😉.

So, this is the initial API I was envisioning in #4770 (DRY-ing that distinction into a shared helper function because I needed it), plus my proposal of folding that special handling in our private lookup function for consistency.

Right now that unborn error is very much HEAD-specific (I came to believe that symbolic references are the lonely kind), as the only code paths that return it are `git_repository_head` and `git_repository_head_for_worktree` ; all the others will return `GIT_ENOTFOUND`. It's not even consistent, because you'll get `GIT_EUNBORNBRANCH` with `git_repository_head` vs. `GIT_ENOTFOUND` with `git_reference_lookup("HEAD")`.

The problem I see is that this error code feels like a hack, because we're overriding `GIT_ENOTFOUND` unconditionally. Looking at my latter `git_reference_lookup_resolve`, it doesn't make any difference if the lookup failed to resolve its initial reference, if one of our symbolic references failed to resolve, or if the final OID wasn't found. To me only the 2nd case is unborn, the others are `GIT_ENOTFOUND` (I'm not sure what a broken direct reference is, so 3 could also be considered unborn, but more broken 😛).

I'd really prefer the API to be consistent, so if the definition above is fine (or if there's a better one), I'd like to fix that consistency issue.

### DIFF
--- a/src/commit.c
+++ b/src/commit.c
@@ -139,7 +139,7 @@ static int git_commit__create_internal(
 	git_array_oid_t parents = GIT_ARRAY_INIT;
 
 	if (update_ref) {
-		error = git_reference_lookup_resolved(&ref, repo, update_ref, 10);
+		error = git_reference_lookup_resolved(&ref, repo, update_ref, 10, 0);
 		if (error < 0 && error != GIT_ENOTFOUND)
 			return error;
 	}
@@ -360,7 +360,7 @@ int git_commit_amend(
 	}
 
 	if (update_ref) {
-		if ((error = git_reference_lookup_resolved(&ref, repo, update_ref, 5)) < 0)
+		if ((error = git_reference_lookup_resolved(&ref, repo, update_ref, 5, 0)) < 0)
 			return error;
 
 		if (git_oid_cmp(git_commit_id(commit_to_amend), git_reference_target(ref))) {
--- a/src/describe.c
+++ b/src/describe.c
@@ -148,7 +148,7 @@ static int retrieve_peeled_tag_or_object_oid(
 	git_object *peeled = NULL;
 	int error;
 
-	if ((error = git_reference_lookup_resolved(&ref, repo, refname, -1)) < 0)
+	if ((error = git_reference_lookup_resolved(&ref, repo, refname, -1, 0)) < 0)
 		return error;
 
 	if ((error = git_reference_peel(&peeled, ref, GIT_OBJ_ANY)) < 0)
--- a/src/refs.c
+++ b/src/refs.c
@@ -162,7 +162,7 @@ int git_reference_remove(git_repository *repo, const char *name)
 int git_reference_lookup(git_reference **ref_out,
 	git_repository *repo, const char *name)
 {
-	return git_reference_lookup_resolved(ref_out, repo, name, 0);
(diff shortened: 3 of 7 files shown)


ethomson:
> Right now that unborn error is very much HEAD-specific

I think that's okay.  Realistically, there's only one symbolic reference, and that's HEAD.  I realize that's not _theoretically_ true, but it is practically.  I think that "unbornness" implies that you ran `git checkout --orphan foo` or its moral equivalent.  AFAIK, there is no way to create an unborn branch where a different symbolic reference is in the mix.  So I think that when:

* HEAD points a reference that doesn't exist, that reference is yet to be born.
* any other symbolic reference points to a reference that doesn't exist, that is an error

But that's just my $0.02.  I don't think that this is a common enough problem to really matter either way so we should do the simplest support for dangling non-HEAD references.

pks-t:
> I think that's okay. Realistically, there's only one symbolic reference, and that's HEAD. I realize that's not theoretically true, but it is practically.

It's also not pratically true. I've already seen dangling symbolic non-HEAD-references in the context of libgit2 somewhen in the past. Not to say the use case is/was valid, but such things do exist in the wild. People are creative.

I'm fine with it being HEAD-specific, though, until somebody complains.

carlosmn:
There's two different things being mixed here.

An unborn branch is one which `HEAD` points to but does not exist. That is the _only_ place where you can talk about unborn branches.

Git happens to support symrefs in places other than `HEAD` but that's more-or-less by accident and even git doesn't bother supporting them properly. If a non-`HEAD` symref refers to a nonexitent ref, that ref is just broken, there is no "we'll create this later".

We should not be making any special effort in supporting these as a blessed form of references. Even if people get creative, we don't have to support every way that people wish Git worked.

tiennou:
Okay, I can see that calling various broken symref "unborn" is 👎. What about adding a `GIT_EBROKEN` error to make those broken symrefs stand out (vs. `GIT_ENOTFOUND` for "broken" direct references). Then the unborn check can be moved inside the refdb lookup, and we cover all that can be wrong with references ? FWIW I'd prefer not having a `GIT_EUNBORNBRANCH` which only means that `master` is "broken", and leave all other kinds of broken references under `GIT_ENOTFOUND`.

Just to clarify, my original use case was [here](https://github.com/libgit2/libgit2/pull/4770/files#diff-52351f78c7aa08c781207c4c667238e4R3207), identifying the cases where a merge can be an unborn-fastforward. So I take it means that merging a branch in a broken branch is only allowed if the broken branch is `HEAD` ?

tiennou:
In the context of #4770, it means something that doesn't exist in git : analyzing/doing the merge of branch in another branch without touching `HEAD`, so I guess in that case merging a branch to a possibly broken symref/"unborn" branch (except for `HEAD`) is an error.

I'll close this and tiptoe carefully around possible broken references in #4770.

PR #4781, opened 2018-08-25
config: convert unbounded recursion into a loop

nelhage:
This can cause a stack overflow from a crafted config file.


pks-t:
Looks good to me, thanks for your fix!

pks-t:
This was part of v0.27.5 already, thus removing the backports label

PR #4784, opened 2018-08-26
Some warnings

tiennou:
That's the 2 warnings I know of I've hinted at recently.

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -1887,7 +1887,7 @@ extern int git_path_is_gitfile(const char *path, size_t pathlen, git_path_gitfil
 	const char *file, *hash;
 	size_t filelen;
 
-	if (gitfile < 0 && gitfile >= ARRAY_SIZE(gitfiles)) {
+	if (!(gitfile >= GIT_PATH_GITFILE_GITIGNORE && gitfile < ARRAY_SIZE(gitfiles))) {
 		giterr_set(GITERR_OS, "invalid gitfile for path validation");
 		return -1;
 	}
--- a/src/streams/stransport.c
+++ b/src/streams/stransport.c
@@ -62,7 +62,7 @@ static int stransport_connect(git_stream *stream)
 
 	ret = SSLHandshake(st->ctx);
 	if (ret != errSSLServerAuthCompleted) {
-		giterr_set(GITERR_SSL, "unexpected return value from ssl handshake %d", ret);
+		giterr_set(GITERR_SSL, "unexpected return value from ssl handshake %d", (int)ret);
 		return -1;
 	}
 
--- a/tests/core/path.c
+++ b/tests/core/path.c
@@ -676,3 +676,12 @@ void test_core_path__16_resolve_relative(void)
 	assert_common_dirlen(6, "a/b/c/foo.txt", "a/b/c/d/e/bar.txt");
 	assert_common_dirlen(7, "/a/b/c/foo.txt", "/a/b/c/d/e/bar.txt");
 }
+
+void test_core_path__git_path_is_file(void)
+{
+	cl_git_fail(git_path_is_gitfile("blob", 4, -1, GIT_PATH_FS_HFS));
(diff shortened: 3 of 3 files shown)


src/path.c
```diff
@@ -1887,7 +1887,7 @@ extern int git_path_is_gitfile(const char *path, size_t pathlen, git_path_gitfil
 	const char *file, *hash;
 	size_t filelen;
 
-	if (gitfile < 0 && gitfile >= ARRAY_SIZE(gitfiles)) {
+	if (gitfile >= ARRAY_SIZE(gitfiles)) {
```
tiennou: I'm doubtful of this fix, but as far as I can see, checking enums for validity has always been a perilous business. The alternative is to cast before checking, but it gets messy because the signedness of an enum is compiler-dependent. Since I believe this argument is never "user-controlled", I'm tempted to remove the "useless" test.

carlosmn: What is the specific warning here? This looks like the error is that this should be `||` instead of `&&` as a `a < 0 ∧a > b where b > 0` is never going to be true.

carlosmn: Hm right, this only seems to be an older clang thing. I don't know what its version 900 corresponds to in "real" clang versions, but on Debian I see this with clang 5 but not with 6. I don't know if they've decided to change the signedness of enums, but this is up to the particular implementation so removing this check would break it for some (apparently including later clang versions).

pks-t: I'd be fine with just removing the check altogether. As @tiennou said, this function is not exposed, any all of our internal callers make sure to just pass proper enum values.

tiennou:
```
Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin17.7.0
```
included in Xcode 9.4.1 (which doesn't feel that old). I'm 👍 with removing the check, because I don't think there's a clean way to do it otherwise.

carlosmn: It's internal now, but it is useful outside of the library and when we expose it we'd have to content with random inputs (I've been meaning to expose this for a while as the fork I send to production has this as public).

carlosmn:
```
	if (!(gitfile > 0 && gitfile <= ARRAY_SIZE(gitfiles))) {
```
does not produce the warning for me in clang 5 and it also fixes the logic.

tiennou: Ok, I've used this, fixed up, and rebased.

src/path.c
```diff
@@ -1887,7 +1887,7 @@ extern int git_path_is_gitfile(const char *path, size_t pathlen, git_path_gitfil
 	const char *file, *hash;
 	size_t filelen;
 
-	if (gitfile < 0 && gitfile >= ARRAY_SIZE(gitfiles)) {
+	if (!(gitfile > 0 && gitfile <= ARRAY_SIZE(gitfiles))) {
```
pks-t: But wait, this is wrong. It's now equivalent to `gitfile <= 0 || gitfile > ARRAY_SIZE(gitfiles)`. So if `gitfile == 0`, we fail even though it's a valid enum value

tiennou:
Should be good. I ended up using the enum value for 0 explicitly, which silences the warning 🤷‍♂️.

src/path.c
```diff
@@ -1887,7 +1887,7 @@ extern int git_path_is_gitfile(const char *path, size_t pathlen, git_path_gitfil
 	const char *file, *hash;
 	size_t filelen;
 
-	if (gitfile < 0 && gitfile >= ARRAY_SIZE(gitfiles)) {
+	if (!(gitfile >= GIT_PATH_GITFILE_GITIGNORE && gitfile < ARRAY_SIZE(gitfiles))) {
```
pks-t: I normally wouldn't like relying on the fact that this enum is actually the first one. But there already is a comment in "src/path.h", stating that the enum is sensitive to ordering. So this is fine

pks-t:
Thanks again, @tiennou!

src/path.c
tiennou: I'm also quite sure there's no simple way to check for an enum signedness, without resorting to build-system shenanigans or explicitly making the enum signed. Hence the test case I've added, which should hopefully catch a change in definition.

PR #4785, opened 2018-08-26
remote: store the connection data in a private struct

tiennou:
This makes it easier to pass connection-related options around (proxy & custom headers for now).

This fixes a bug in `git_push_finish`, which didn't reuse the provided proxy if the connection closed between the call to `git_remote_push` and the finish step.

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -657,7 +657,7 @@ static int set_transport_custom_headers(git_transport *t, const git_strarray *cu
 	return t->set_custom_headers(t, custom_headers);
 }
 
-int git_remote_connect(git_remote *remote, git_direction direction, const git_remote_callbacks *callbacks, const git_proxy_options *proxy, const git_strarray *custom_headers)
+int git_remote__connect(git_remote *remote, git_direction direction, const git_remote_callbacks *callbacks, const git_remote_connection_opts *conn)
 {
 	git_transport *t;
 	const char *url;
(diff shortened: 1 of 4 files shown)


src/remote.c
```diff
@@ -951,15 +961,16 @@ int git_remote_fetch(
 	const git_remote_callbacks *cbs = NULL;
 	const git_strarray *custom_headers = NULL;
 	const git_proxy_options *proxy = NULL;
+	struct git_remote_connection conn;
```
pks-t: I cannot see this value ever being used. You probably meant to remove `custom_headers` and `proxy` and then call `git_remote__connect` with `conn`, didn't you?

Edit: and you also need to initialize that field, as you only conditionally assign its fields

src/remote.c
```diff
@@ -2373,8 +2384,7 @@ int git_remote_upload(git_remote *remote, const git_strarray *refspecs, const gi
 	git_push *push;
 	git_refspec *spec;
 	const git_remote_callbacks *cbs = NULL;
-	const git_strarray *custom_headers = NULL;
-	const git_proxy_options *proxy = NULL;
+	git_remote_connection conn;
```
pks-t: You need to initialize `conn`, as you only conditionally set its fields

src/remote.h
```diff
@@ -36,6 +36,15 @@ struct git_remote {
 	int passed_refspecs;
 };
 
+typedef struct git_remote_connection {
+	const git_strarray *custom_headers;
+	const git_proxy_options *proxy;
+} git_remote_connection;
```
pks-t: Sorry to be nitpicky here, but I've just realized that this structure is not a connection, but instead rather an options structure. So `git_remote_connection_opts` might be more fitting? Especially as you also embed this struct below the `/* options */` comment in the `git_push` structure.

pks-t:
I'm good with this PR despite the one naming issue I have right now.

tiennou:
I've updated the naming (good catch) and rebased on top of master. Thanks for the review @pks-t !

pks-t:
Thank you, @tiennou!

PR #4786, opened 2018-08-27
[WIP] Normalize ciphersuites

tiennou:
This is an attempt at having a consistent interface for the ciphersuite management.

Unless I severely misunderstood something, this should be sufficient for our backends, so I think the `tls.h` API is complete. It's based on the NIST names instead of the OpenSSL ones, so it's a backward-compatibility break, but it would not be hard to handle both (though I'd be 👎, because the script should be able to convert to NIST naming).

Backends ported :
- [x] SecureTransport
- [ ] OpenSSL
- [ ] mbedTLS

Ref #3682, #4418.


tiennou:
Most backends ported, only WinHTTP left TODO. Memcheck seems to dislike something I wrote but I haven't investigated yet.

It turns out I could add a `GIT_EHANDSHAKE`/`GIT_ECIPHER` error, but sadly mbedTLS doesn't give access to the actual TLS Alert (it reports `MBEDTLS_ERR_SSL_PEER_CLOSE_NOTIFY`, even though Wireshark was able to see that.

src/streams/openssl.c
```diff
@@ -602,6 +600,8 @@ int openssl_connect(git_stream *stream)
 	if ((ret = SSL_connect(st->ssl)) <= 0)
 		return ssl_set_error(st->ssl, ret);
 
+	st->connected = true;
```
pks-t: The same pattern exists in mbedtls code, so we might want to do the same check over there, too.

script/resolve-ciphersuite.rb
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env ruby -W
```
pks-t: Can we please avoid adding another scripting language dependency to libgit2? We already have a dependency on Python due to clar, so that's the one we should aim for. Just let me know if you feel uncomfortable with rewriting this script in Python and I'll do it for you

src/streams/tls.h
```diff
@@ -20,4 +20,412 @@
...
+	{0xffe1, NULL,                                "SSL_RSA_FIPS_WITH_DES_CBC_SHA"},
+	{0x10080, "RC4-MD5",                           "SSL_CK_RC4_128_WITH_MD5"},
+	{0x20080, "EXP-RC4-MD5",                       "SSL_CK_RC4_128_EXPORT40_WITH_MD5"},
+	{0x30080, "RC2-CBC-MD5",                       "SSL_CK_RC2_128_CBC_WITH_MD5"},
+	{0x40080, "EXP-RC2-CBC-MD5",                   "SSL_CK_RC2_128_CBC_EXPORT40_WITH_MD5"},
+	{0x50080, "IDEA-CBC-MD5",                      "SSL_CK_IDEA_128_CBC_WITH_MD5"},
+	{0x60040, "DES-CBC-MD5",                       "SSL_CK_DES_64_CBC_WITH_MD5"},
+	{0x700c0, "DES-CBC3-MD5",                      "SSL_CK_DES_192_EDE3_CBC_WITH_MD5"},
+	{0x80080, "RC4-64-MD5",                        "SSL_CK_RC4_64_WITH_MD5"},
+};
```
pks-t: This table is huge. It grows the data segment of the resulting shared library by roughly 35%, and overall library size increases by 3.3%. This might be acceptable, but I wonder whether we'll have to expand this table by more mappings for other TLS libraries, causing it to grow even bigger in size.

src/streams/tls.h
```diff
@@ -20,4 +20,412 @@
...
+ * GIT_OPT_SET_SSL_CIPHERS.
+ *
+ * As this function must work with read-only data, the name is never NULL-terminated.
+ *
+ * @param out_name the cipher name
+ * @param out_len the cipher name length
+ * @param cipher_list the current position in the cipher list
+ * @return 0 on success, GIT_ITEROVER if the list is empty
+ */
+GIT_EXTERN(int) git_tls_enabled_ciphers(const char **out_name, size_t *out_len, const char **cipher_list);
```
pks-t: This interface is a bit weird. It does not have anything to do with enabled ciphers, but only acts as some kind of special-cased strtok, right?

src/streams/tls.h
```diff
@@ -20,4 +20,412 @@
...
+} git_tls_cipher;
+
+/**
+ * Parse a list of cipher names from cipher_list.
+ *
+ * The intended usage is to pass a pointer to a list of cipher; either the
+ * default list of ciphers (GIT_TLS_DEFAULT_CIPHERS) or the list configured by
+ * GIT_OPT_SET_SSL_CIPHERS.
+ *
+ * As this function must work with read-only data, the name is never NULL-terminated.
```
pks-t: You probably mean `NUL`, not `NULL`.

src/streams/stransport.c
```diff
@@ -49,6 +51,67 @@ typedef struct {
...
+		error = -1;
+		goto cleanup;
+	}
+
+	while (git_tls_enabled_ciphers(&name, &len, &cipher_list) == 0) {
+		const git_tls_cipher *cipher;
+		size_t idx;
+
+		if (git_tls_cipher_lookup(&cipher, name, len) < 0) {
+			/* TODO: cipher not found, warn ? */
```
pks-t: Warning wouldn't make a lot of sense in case where this is going to be our generic interface for all implementations. Our set of enabled-by-default ciphers shouldn't be what is implemented by all the TLS implementations we support, but instead by what's secure and available in at least a subset of them. Thus warnings wouldn't make sense in that case, as it is expected that some of our preferred ciphers are not available in some of our TLS backends.

src/streams/openssl.c
```diff
@@ -200,10 +199,43 @@ static void shutdown_ssl(void)
...
+	git_buf cipher_buf = GIT_BUF_INIT;
+	const char *name;
+	size_t len;
+	int error = 0;
+
+	while (git_tls_enabled_ciphers(&name, &len, &cipher_list) == 0) {
+		const git_tls_cipher *cipher;
+
+		if (git_tls_cipher_lookup(&cipher, name, len) < 0) {
+			/* TODO: cipher not found, warn ? */
```
pks-t: Same remark as for stransport applies here

src/streams/openssl.c
```diff
@@ -200,10 +199,43 @@ static void shutdown_ssl(void)
...
+			continue;
+		}
+
+		git_buf_puts(&cipher_buf, cipher->openssl_name);
+		git_buf_putc(&cipher_buf, ':');
+	}
```
pks-t: You should check for OOM after the loop

tests/online/badssl.c
```diff
@@ -67,14 +67,9 @@ void test_online_badssl__old_cipher(void)
...
-
 	if (!g_has_ssl)
 		cl_skip();
 
-	cl_git_fail_with(GIT_ECERTIFICATE,
-			 git_clone(&g_repo, "https://rc4.badssl.com/fake.git", "./fake", NULL));
-	cl_git_fail_with(GIT_ECERTIFICATE,
-			 git_clone(&g_repo, "https://rc4.badssl.com/fake.git", "./fake", &opts));
+	cl_git_fail(git_clone(&g_repo, "https://rc4.badssl.com/fake.git", "./fake", NULL));
+	cl_git_fail(git_clone(&g_repo, "https://rc4.badssl.com/fake.git", "./fake", &opts));
```
pks-t: Cool :)

src/streams/mbedtls.c
```diff
@@ -70,6 +69,7 @@ static void shutdown_ssl(void)
 		git__free(mbedtls_entropy);
 		mbedtls_entropy = NULL;
 	}
+	git__free(git__cipher_list);
```
pks-t: You should re-initialize the value, just to make sure we never run into a double free

src/streams/mbedtls.c
```diff
@@ -43,12 +45,9 @@
...
-#define GIT_SSL_DEFAULT_CIPHERS_COUNT 30
-
 mbedtls_ssl_config *git__ssl_conf;
-static int ciphers_list[GIT_SSL_DEFAULT_CIPHERS_COUNT];
 mbedtls_entropy_context *mbedtls_entropy;
+int *git__cipher_list = NULL;
```
pks-t: The initialization is done automatically, so no `NULL` required.

By the way, is there any reason why none of these variables are static?

script/resolve-ciphersuite.rb
tiennou: Well, it's just an one-off script : you only need to use it in case we want to regenerate the mapping list & default ciphers, so the dependency is more transient than clar's Python.

If that's really a problem I can rewrite it, but as I don't expect it to run as part of the build system I didn't think it would matter.

src/streams/tls.h
tiennou: I… didn't think about that 🤔. 35% is pretty big so I'm kinda meh on my own solution here.

As far as I can see, I don't think other mappings will be necessary (famous last words), as there are only 2 solutions in use :
- The OpenSSL/gnuTLS way, which goes through non-canonical names as a :-separated string, with various generic suite names (ie. "RC4"/"!RC4").
- The "enum"/"#define" with standard values (SecureTransport, mbedTLS). The API just accepts a `int *` to protocol IDs.

It might be possible to only build the needed mapping for each backend, which would somewhat lower the global size : mbedTLS/SecureTransport would be a simple `char *` to `int`, OpenSSL would still have a `char *` to `char *`.

src/streams/tls.h
tiennou: True. I'll rename it to `git_tls_cipher_foreach` or something.

src/streams/tls.h
tiennou: Shenanigans 😜. Will fix though.

src/streams/mbedtls.c
tiennou: I wondered the same. I'll change them if that's doable.

src/streams/openssl.c
tiennou: This has been rebased out.

tiennou:
Rebased, I've taught the script which backend liked which kind of identifier, so the cipherlist is now "minimal" (though I expect the OpenSSL one is still massive).

The build is red because the struct is missing for WinHTTP. I suspect there's a call that has the same function in Windows-land, @ethomson ?

Issue #4204, opened 2017-04-13
Client cert support


jjpe:
On [this libgit2sharp issue](https://github.com/libgit2/libgit2sharp/issues/1152#issuecomment-279966002) thread it is mentioned that this project has no support for client certificates at this time. 
I am interested in adding support for this, but am not sure what it needed to achieve it. By submitting this as an issue here I'm hoping to discuss the issue and avoid any potential duplicated effort.

### Reproduction steps
Try to clone any project with some application that uses libgit2 to achieve it, from a server that asks for a client certificate. Some concrete examples of this are [neon](https://github.com/neon-bindings/neon) and [npm](https://github.com/npm/npm).
I'm assuming at this point that the same behavior holds for e.g. a `pull` as well as any other functionality that require interaction with a `remote`.

### Expected behavior
I'd like the clone, pull etc to succeed.

### Actual behavior
At this moment the exact error varies a bit depending on the actual application but it comes down to the server not authenticating the client application due to a missing certificate.

### Version of libgit2 (release number or SHA1)
Any, as I'm discussing new functionality.

### Operating system(s) tested
Windows, OS X, Ubuntu Linux

LordNali:
We'd like to participate or at least support this discussion. Is this feature (request) so unique - many companies I know are you using git repositories with client certificates.

pks-t:
The problem is probably not about unique-ness, but simply that nobody had enough interest and/or time yet to implement this. We'd certainly welcome pull requests implementing this feature.

jjpe:
I have no patches atm, mainly because I'm not sure what's needed conceptually to make this work.

To avoid the X/Y problem, let me state that my real goal is this: I need to be able to use tools like [cargo](https://github.com/rust-lang/cargo)* and [neon](https://github.com/neon-bindings/neon)** in such a way that dependencies located on a GitHub Enterprise server*** resolve as neatly as on github.com or the respective repositories (e.g. [crates.io](https://crates.io) for Cargo). 

* Cargo uses libgit2
** Neon uses Cargo internally
***This GitHub Enterprise server has its own certificates

ethomson:
@jjpe To clarify:  your GitHub Enterprise server "has its own certificates" (ie, it's using a self-signed certificate or a certificate that's signed by your company's CA which isn't trusted)?

[...]

PR #4787, opened 2018-08-29
Disable SSL/TLS client certificate handling in WinHTTP.

achidlow:
I can see there is currently a feature request #4204 for actually being able to supply a client certificate, however this change is still an incremental improvement over the current situation where there is no possible way to connect to a server that asks for an optional client certificate.

Code is basically straight from Microsoft docs: https://docs.microsoft.com/en-us/windows/desktop/winhttp/ssl-in-winhttp#optional-client-ssl-certificates

### DIFF
--- a/src/transports/winhttp.c
+++ b/src/transports/winhttp.c
@@ -404,6 +404,14 @@ static int winhttp_stream_connect(winhttp_stream *s)
 		goto on_error;
 	}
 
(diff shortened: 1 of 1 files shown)


ethomson:
Although I think that this change (as described) might make sense, it appears that this code is not quite right.  Regrettably, I'm closing this PR for now since there's been no activitiy.

PR #4790, opened 2018-08-30
ci: remove travis

ethomson:
Travis has started erroring:

```
WARNING: The following packages cannot be authenticated!
  libssh2-1 libcurl4-gnutls-dev libcurl3-gnutls curl libcurl3 libssh2-1-dev
E: There were unauthenticated packages and -y was used without --allow-unauthenticated
```

I...  don't have any idea why.  And, to be completely honest, I don't have a strong interesting in investigating.  At this point, VSTS supports Linux, WIndows and macOS builds in a single place and at roughly the same speed as Travis.  Travis isn't giving us much additional benefit and I don't want to spend a lot of time to try to figure this out.

Objections?


pks-t:
On Thu, Aug 30, 2018 at 01:57:21PM -0700, Edward Thomson wrote:
> Travis has started erroring:
> 
> ```
> WARNING: The following packages cannot be authenticated!
>   libssh2-1 libcurl4-gnutls-dev libcurl3-gnutls curl libcurl3 libssh2-1-dev
> E: There were unauthenticated packages and -y was used without --allow-unauthenticated
> ```
> 
> I...  don't have any idea why.  And, to be completely honest, I
> don't have a strong interesting in investigating.  At this
> point, VSTS supports Linux, WIndows and macOS builds in a
> single place and at roughly the same speed as Travis.  Travis
> isn't giving us much additional benefit and I don't want to
> spend a lot of time to try to figure this out.

I bet this is because of a MITM RCE report reported to
oss-security which catched my attention a few days ago. The issue
is that Travis by default used "--force-yes" to install packages
via the apt plugin, which allowed to install packages where the
signing key was either missing, out of date or invalid. This may
allow an attacker to inject arbitrary packages, which is bad in
case where you generate artifacts or deploy code from Travis.

To fix the vulnerability, "--force-yes" was replaced on
2018-07-05 by "--allow-downgrades", "--allow-remove-essential"
and "--allow-change-held-packages", but the fix was partially
reverted again due to issues. Maybe they reverted the revert so
that now we are being bitten by this?

> Objections?

The only issue I have right now with VSTS is that I don't have
any access to trigger builds or otherwise do anything to the
environment. Does it require a Microsoft account to get access?

ethomson:
> The only issue I have right now with VSTS is that I don't have any access to trigger builds or otherwise do anything to the environment. Does it require a Microsoft account to get access?

It does.  Do you have one?

It looks like the identity team is starting to [integrate other ways to sign in to Microsoft services](https://cloudblogs.microsoft.com/enterprisemobility/2018/08/28/azure-ad-b2b-collaboration-support-for-google-ids-is-now-in-public-preview/) but it sounds like it's Not Quite Ready Yet.

One thing that I was thinking was to hook [Probot](https://probot.github.io/) up so that you could do a `/rebuild` in a PR comment or over email and have it do some magic.  This might be useful if you were not interested in creating a Microsoft account.

tiennou:
Since we're discussing access, I created an account with my GitHub's profile email a few weeks ago.

Though I wouldn't mind using Probot for that 😉.

ethomson:
I decided to [whip up a Probot script](https://github.com/ethomson/probot-vsts-build) after all.  I need to figure out how/where to deploy it but then I'll continue to move forward with this.

pks-t:
On Fri, Aug 31, 2018 at 03:05:51AM -0700, Edward Thomson wrote:
> > The only issue I have right now with VSTS is that I don't
> > have any access to trigger builds or otherwise do anything to
> > the environment. Does it require a Microsoft account to get
> > access?
> 
> It does.  Do you have one?

No, I don't.

> One thing that I was thinking was to hook
> [Probot](https://probot.github.io/) up so that you could do a
> `/rebuild` in a PR comment or over email and have it do some
> magic.  This might be useful if you were not interested in
> creating a Microsoft account.

That would be very welcome!

ethomson:
> Okay, @ethomson, I started to rebuild this pull request.

😅

ethomson:
So I did create a GitHub App, based on [Probot](https://probot.github.io/).

If you're curious, it's https://github.com/ethomson/probot-vsts-build, and I intend to write up something more proper.  But in a nutshell, it will listen for issue comments that are `/rebuild` (and only that, although whitespace is trimmed).  It will then ensure that the person who issued the command is a collaborator on the project (per the GitHub API notion, which basically means that they've been added to a team).  Then it looks up the corresponding projects in VSTS, and looks for build definitions in those projects that are set to build pull requests in this repository.  (Since a GitHub App is per-account, this lets us use it on other repositories in the libgit2 project).  Finally, it will look at builds that were queued for this particular pull request within each of those build definitions, and re-queues them.

(This is true whether the builds passed or failed.)

pks-t:
That's great! It'd be even greater if the bot comment also linked to the new build that is being performed. No idea if it has all required information for that readily available, though

ethomson:
That's a good idea - let me see what I can do.

pks-t:
Thanks a lot for working on this! It really improves our infrastructure by a whole lot

PR #4791, opened 2018-08-31
Update VSTS YAML files with the latest syntax

DavidStaheli:
The VSTS YAML syntax has evolved to use "pool" instead of "queue" and "job" instead of "phase".  Additionally, this PR adds use of the "Ubuntu 16.04" pool which has performance benefits over "Hosted Linux Preview."

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -5,11 +5,11 @@ trigger:
 - master
 - maint/*
 
-phases:
-- phase: linux_trusty_gcc_openssl
+jobs:
+- job: linux_trusty_gcc_openssl
   displayName: 'Linux (Trusty; GCC; OpenSSL)'
-  queue:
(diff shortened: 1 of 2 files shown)


ethomson:
Neat, thanks @DavidStaheli.  Love that we'll have faster builds!

dheeru007:
[24/08 02:42] First Pre-Launch:

I'm just a little curious, but...

...do you even read my details?

If not, I suggest you read this one,
otherwise you'll be passing up on something
HUGE that you have to Free register, by clicking following link.*

https://www.lockinyourprofits.com/dgbaba

https://fln2.lockinyourprofits.com/dgbaba

https://fln3.lockinyourprofits.com/dgbaba
Do this ASAP, because your time is running out!

-DG007

P.S. You really will kick yourself later if you
miss out on this. Click above links and prevent that.

PR #4792, opened 2018-09-01
config: Fix a leak parsing multi-line config entries

nelhage:


pks-t:
I've amended this PR to improve the error handling of this function in general. It's much too frickle in my opinion. Please let me know if you're fine with those changes

pks-t:
Thanks @nelhage and @tiennou, both for fixing the leak as well as the reviews!

PR #4794, opened 2018-09-02
git_remote_prune to be O(n  * logn)

marcin-krystianc:
This is to improve performance of `git_remote_prune`. Before this change it got **O(n * n)**  complexity, now it is **O(n * logn)**.

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -1234,7 +1234,7 @@ int git_remote_prune(git_remote *remote, const git_remote_callbacks *callbacks)
 				goto cleanup;
 
 			key.name = (char *) git_buf_cstr(&buf);
-			error = git_vector_search(&pos, &remote_refs, &key);
+			error = git_vector_bsearch(&pos, &remote_refs, &key);
 			git_buf_dispose(&buf);
 
 			if (error < 0 && error != GIT_ENOTFOUND)


src/remote.c
```diff
@@ -1234,7 +1234,7 @@ int git_remote_prune(git_remote *remote, const git_remote_callbacks *callbacks)
 				goto cleanup;
 
 			key.name = (char *) git_buf_cstr(&buf);
-			error = git_vector_search(&pos, &remote_refs, &key);
+			error = git_vector_bsearch(&pos, &remote_refs, &key);
```
pks-t: So before we had a complexity of O(j*k*n), where `j=|refnames|`, `k=|active-refspecs|` and `n=|remote_refs|`. We didn't do any sorting before, so no overhead existed because of that sorting.

With your change we now have to sort it, which is automatically performed by `git_vector_bsearch`. It's only done on the first iteration, though, as we set a flag in the vector that avoids us trying to sort an already sorted vector. So that now leaves us with a complexity of `O(n*logn + j*k*logn)` (`O(nlogn)` for sorting, `O(jklogn)` for the binary search inside of the nested loop). So big-O complexity is definitely smaller.

I guess that this matters especially for remotes with a huge number of references. Google's Gerrit instance for Android comes to mind, which has hundreds of thousands of references. Reducing the complexity for searching these refs is definitely a good thing. I bet that we perform a bit slower with these changes for smallish lists of remote references, due to hitting additional code paths now. But the cost is going to be negligible, I'd say.

So: looks good to me. Waiting a few days before I merge this to enable feedback from others.

tiennou: I find this codepath a little strange (even without the OP changes), so here are my thoughts :
- the original code path wasn't sorting anyway, by virtue of using `git_vector_insert` vs `git_vector_insert_sorted` [here](https://github.com/libgit2/libgit2/blob/master/src/remote.c#L866), even though the comparison function is set.
- AFAIU the cmp functions themselves are quite similar ([ls_to_vector](https://github.com/libgit2/libgit2/blob/master/src/remote.c#L851), [prune](https://github.com/libgit2/libgit2/blob/master/src/remote.c#L1204) which will end up calling `strcmp` [anyway](https://github.com/libgit2/libgit2/blob/master/src/util.c#L600), so the whole set-sort-function dance in git_remote_prune seems unnecessary.

So my suggestion, if that's appropriate (and functionally equivalent), would be to drop the `git_vector_set_cmp` plus static function in `git_remote_prune` and maybe switch to a sorted insert in `ls_to_vector` so the sorting happens inline ?

marcin-krystianc: I think that usage of `git_vector_insert_sorted` will be less efficient than simple `git_vector_insert` followed finally by sort operation. When sorted insert is performed, for each element that need to be inserted in the middle of the vector, it will trigger expensive `memmove` operation.  The more elements vector already has the more expensive `memmove` becomes. 
So I don't think there is any benefit in using sorted insert here.

Regarding the `remote_head_cmp` function, I think you didn't notice that it is because, it compares `git_remote_head` objects by `name` field. If we want to keep sorting by `name` we need to keep that function.

pks-t: Yup, I'd also vote for doing the sorting once at the end, only. As @marcin-krystianc said, it'll be less efficient to perform it inline. So I guess we're all on the same page and this can be merged as-is.

pks-t:
Thanks for this PR, @marcin-krystianc!

PR #4801, opened 2018-09-09
CI: Support multiple executables on a single cmake test

ethomson:
CMake allows us to add multiple executable runs on a single ctest via multiple `ADD_TEST` invocations.  Currently, we call ctest to get the executable information, so that we can (optionally) wrap it in a validator like valgrind.  However, we expect a single executable per ctest invocation.

Expand this support to run multiple executable invocations per ctest test - parse all the ctest output and run each command.  Though we do not make use of this in the master branch presently, we probably want to make use of this in maintenance branches (which do have multiple clar test invocations for a single ctest) and we may make use of this in the future.

Since we've added JUnit XML outputs in #4778, this complicates our reporting a bit.  Add a new mode to clar to append summary output files.


ethomson:
To provide some additional rationale here:  I want to use our current CI build + test scripts on the 0.27 and 0.26 maintenance branches without changing their existing ctest structure.  The master branch has refactored the ctest structure a bit, so that there are discrete tests (offline, online, ssh, etc) that correspond to a single clar invocation.  The 0.27 and 0.26 branches are not set up this way - a single ctest invocation maps to multiple clar runs.

I don't want to change the 0.27 and 0.26 branches in case somebody downstream is relying on their tests working a certain way via ctest.  However, the new CI test scripts invoke ctest to learn what it _would_ invoke if it were run (ie, gets the clar command line) and then invokes clar directly or within the framework of valgrind or another leak checker.  This lets people use ctest if they want (and a lot of people do seem to want to.)  But it also ensures that for our automated test runs,  ctest is not part of the mix, we don't want to run valgrind over ctest, after all.  (Nor do we want ctest invoking the macOS leak checker invoking clar, the way leaks deals with stdout/stderr causes absolute madness.)

So, the CI scripts need to learn how to deal with multiple clar invocations per ctest run.  And, if they're doing that, then they also need to learn how not to stomp all over each others test XML output.  So now we have a mode to append more test XML output to an existing XML file, which is... special.

ethomson:
Right, so, thankfully, I'm not real bright.  I was misinterpreting the output of our old ctest configuration (from the 0.27 maintenance branch) and thinking that it was running clar multiple times for a given ctest configuration.  It's _not_.  We're running `ctest -R` which matches a _regex_.  And since all the tests have `libgit2_clar` in their name, trying to run the first test (which is just "libgit2_clar") is a regex match for all the tests.  🙄

I seem to remember learning this once and then forgetting it.  So, thank god, this is totally pointless.

PR #4803, opened 2018-09-11
index: release the snapshot instead of freeing the index

### LINKED ISSUE #4802: assert on aix
### Reproduction steps
Clone a repository on AIX. The following assert occurs:
**Assertion failed: __EX, file .../libgit2-0.27.3/src/index.c, line 460**

Call stack:
...
-> iterator_for_filesystem(...)
  -> git_index_snapshot_new(...)
    -> git_index_free(...)

*git_vector_dup()* fails in *git_index_snapshot_new()*, the index is released but the *readers* variable was increased before which leads to the assert.

*git_vector_dup()* fails because a *malloc(0)* on AIX returns a NULL pointer:

v->contents = git__malloc(bytes);
GITERR_CHECK_ALLOC(v->contents);

### Expected behavior
If *malloc()* is called with 0 bytes the return value of NULL is valid (at least on AIX) and should not be treated as an error.

0.27.3

AIX 7.1
gcc 4.8.2

tiennou:
Previously we would assert in index_free because the reader incrementation would not be balanced. Release the snapshot normally, so the variable gets decremented before the index is freed.

Fixes #4802.

### DIFF
--- a/src/vector.c
+++ b/src/vector.c
@@ -50,22 +50,24 @@ int git_vector_size_hint(git_vector *v, size_t size_hint)
 
 int git_vector_dup(git_vector *v, const git_vector *src, git_vector_cmp cmp)
 {
-	size_t bytes;
-
 	assert(v && src);
 
-	GITERR_CHECK_ALLOC_MULTIPLY(&bytes, src->length, sizeof(void *));
-
-	v->_alloc_size = src->length;
+	v->_alloc_size = 0;
+	v->contents = NULL;
 	v->_cmp = cmp ? cmp : src->_cmp;
 	v->length = src->length;
 	v->flags  = src->flags;
 	if (cmp != src->_cmp)
 		git_vector_set_sorted(v, 0);
-	v->contents = git__malloc(bytes);
-	GITERR_CHECK_ALLOC(v->contents);
 
-	memcpy(v->contents, src->contents, bytes);
+	if (src->length) {
+		size_t bytes;
+		GITERR_CHECK_ALLOC_MULTIPLY(&bytes, src->length, sizeof(void *));
+		v->contents = git__malloc(bytes);
+		GITERR_CHECK_ALLOC(v->contents);
+		v->_alloc_size = src->length;
+		memcpy(v->contents, src->contents, bytes);
+	}
 
 	return 0;
 }
--- a/src/vector.c
+++ b/src/vector.c
@@ -32,6 +32,9 @@ GIT_INLINE(int) resize_vector(git_vector *v, size_t new_size)
 {
 	void *new_contents;
 
+	if (new_size == 0)
+		return 0;
+
 	new_contents = git__reallocarray(v->contents, new_size, sizeof(void *));
 	GITERR_CHECK_ALLOC(new_contents);
 
--- a/src/index.c
+++ b/src/index.c
@@ -3550,7 +3550,7 @@ int git_index_snapshot_new(git_vector *snap, git_index *index)
 	error = git_vector_dup(snap, &index->entries, index->entries._cmp);
 
 	if (error < 0)
-		git_index_free(index);
+		git_index_snapshot_release(snap, index);
 
 	return error;
 }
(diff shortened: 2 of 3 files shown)


ethomson:
Fixes half of it, anyway - we need to not `malloc(0)` as well.  😀

ethomson:
(That wasn't meant as snark, BTW, I just meant that we need to not close the issue after merging this.)

tiennou:
I was under the impression that `NULL` was a valid return value from `malloc(0)`, but it happens that the other value allowed by the standard is a non-NULL, non-dereferencable pointer (thanks, but no thanks 😁).

Would it be acceptable to just `MIN_ALLOCSIZE` a 0-sized vector on dup then ? Or should I just skip the allocation altogether ?

tiennou:
Fixed the `malloc(0)` issue when duping empty/unallocated vectors. I'm still not sure the vector code is completely safe in that regard though, as `resize_vector` calls `realloc`, which has the same caveat.

src/vector.c
```diff
@@ -54,18 +54,21 @@ int git_vector_dup(git_vector *v, const git_vector *src, git_vector_cmp cmp)
...
 		git_vector_set_sorted(v, 0);
-	v->contents = git__malloc(bytes);
-	GITERR_CHECK_ALLOC(v->contents);
 
-	memcpy(v->contents, src->contents, bytes);
+	GITERR_CHECK_ALLOC_MULTIPLY(&bytes, src->length, sizeof(void *));
+	if (bytes != 0) {
+		v->contents = git__malloc(bytes);
+		GITERR_CHECK_ALLOC(v->contents);
+		v->_alloc_size = bytes;
```
ethomson: Prior to this change, `v->_alloc_size` was `src->length`, not the product of `src->length` and `sizeof(void *)`.

src/vector.c
```diff
@@ -54,18 +54,21 @@ int git_vector_dup(git_vector *v, const git_vector *src, git_vector_cmp cmp)
...
 	v->length = src->length;
 	v->flags  = src->flags;
 	if (cmp != src->_cmp)
 		git_vector_set_sorted(v, 0);
-	v->contents = git__malloc(bytes);
-	GITERR_CHECK_ALLOC(v->contents);
 
-	memcpy(v->contents, src->contents, bytes);
+	GITERR_CHECK_ALLOC_MULTIPLY(&bytes, src->length, sizeof(void *));
+	if (bytes != 0) {
```
pks-t:
```
if (src->length) {
    size_t bytes;
    GITERR_CHECKALLOC_MULTIPLY(&bytes, src->length, sizeof(void *));
    ...
}
```

No need to do unnecessary work in case where `src->length` is already `0`. I think like this intentions also become a bit clearer.

src/vector.c
tiennou: Good catch. I misunderstood what `_alloc` means here.

src/vector.c
```diff
@@ -32,6 +32,9 @@ GIT_INLINE(int) resize_vector(git_vector *v, size_t new_size)
 {
 	void *new_contents;
 
+	if (new_size == 0)
+		return 0;
```
pks-t: One might tend to think that resizing to `0` might mean that we should deallocate the vector. But `resize_vector` is only called from `git_vector_init`, `git_vector_insert`, `git_vector_insert_sorted`, `git_vector_insert_null`, `git_vector_size_hint` and `git_vector_resize_to`. Only for the last two it _might_ make sense to do the deallocation dance, though.

- `git_vector_resize_to` is only used in `git_vector_set`, so the deallocation path could never be hit here.
- `git_vector_size_hint` will only do any work if the new size hint is bigger than `alloc_size`. So it also cannot ever hit the deallocation path.

So the dealloc would in fact be useless.

I think the only code path that can actually invoke `resize_vector(v, 0)` is `git_vector_size_hint`. And I can imagine a caller trivially invoking it e.g. if he wants to allocate a vector for a set of `n` items, where `n` might also be `0`.

Anyway, long story short: makes sense to me.

pks-t:
Thanks @tiennou!

PR #4804, opened 2018-09-11
ci: arm docker builds

ethomson:
Docker can leverage QEMU to run images on other platforms!  Introduce an arm32 based build.


ethomson:
```
2018-09-12T10:22:44.6331866Z Operating system version:
2018-09-12T10:22:44.6649758Z     Linux 2c96ecf832f7 4.15.0-1022-azure #22~16.04.1-Ubuntu SMP Thu Aug 16 10:31:05 UTC 2018 aarch64 aarch64 aarch64 GNU/Linux
```

neithernut:
It says it could not bind to `::1`. Maybe they (CI provider) do't provide IPv6 support on arm for some reason?

ethomson:
> It says it could not bind to ::1. Maybe they (CI provider) do't provide IPv6 support on arm for some reason?

Not sure what's going on.  This is running in an emulator in docker, so there's a lot of strange things that could be going on.  I pushed up some changes to the proxy so that it will try to bind only on 127.0.0.1 in case it is the IPv6 problem.

neithernut:
Interestingly, something still tries to bind to `::1`. And all Linux-based tests are now failing.

neithernut:
At least in the [latest test](https://libgit2.visualstudio.com/libgit2/_build/results?buildId=571&view=logs) the proxy server _does_ bind to 127.0.0.1:

    [2018-09-12 17:18:08] com.microsoft.tfs.tools.poxy.Poxy: Starting server on 127.0.0.1:8118

Still, something tries to bind to `::1`, so it's probably another process. Maybe it's just a deceptive error message and it's actually the process trying to connect to the proxy which is printing the error? This would at least explain why the other tests fail all of a sudden.

ethomson:
> Still, something tries to bind to ::1, so it's probably another process.

Yes, I agree.  It's a red herring in this, it appears.  I can repro this locally - I thought I had all the tests passing last night.  So either I misunderstood what I saw last night (likely) or I broke something between now and then (also possible).

ethomson:
Aha.  Running java directly yields a big fat clue:

```
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000004000e18dd4, pid=17, tid=0x000000404d339200
#
# JRE version: OpenJDK Runtime Environment (8.0_181-b13) (build 1.8.0_181-8u181-b13-0ubuntu0.16.04.1-b13)
# Java VM: OpenJDK 64-Bit Server VM (25.181-b13 mixed mode linux-aarch64 compressed oops)
# Problematic frame:
# V  [libjvm.so+0x3e3dd4]
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /build/hs_err_pid17.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
#
```

This appears to happen on any jar.  😞

.vsts-ci.yml
```diff
@@ -56,6 +56,30 @@ jobs:
...
+  displayName: 'Linux (Xenial; arm32)'
+  pool:
+    vmImage: 'Ubuntu 16.04'
+  steps:
+  - template: ci/vsts-docker.yml
+    parameters:
+      qemu: 'true'
+      imageName: 'libgit2/xenial-arm32:test'
+      environmentVariables:
+			  SKIP_PROXY_TESTS: true
```
neithernut: Looks like the wrong kind of whitespace. CI complains about this line, too. *pointing out the obvious*

.vsts-ci.yml
```diff
@@ -56,6 +56,30 @@ jobs:
...
+  displayName: 'Linux (Xenial; arm64)'
+  pool:
+    vmImage: 'Ubuntu 16.04'
+  steps:
+  - template: ci/vsts-docker.yml
+    parameters:
+      qemu: 'true'
+      imageName: 'libgit2/xenial-arm64:test'
+      environmentVariables:
+			  SKIP_PROXY_TESTS: true
```
neithernut: Same here.

ethomson:
Thanks, @neithernut.  In my haste I forgot to clean up tabs vs spaces.

.vsts-ci.yml
```diff
@@ -56,6 +56,32 @@ jobs:
        CMAKE_OPTIONS=-DUSE_HTTPS=mbedTLS -DSHA1_BACKEND=mbedTLS
        LEAK_CHECK=valgrind
 
+- job: linux_xenial_x86
```
neithernut: Maybe you should include "gcc" in the name

.vsts-ci.yml
```diff
@@ -56,6 +56,32 @@ jobs:
...
+  steps:
+  - template: ci/vsts-docker.yml
+    parameters:
+      qemu: 'true'
+      imageName: 'libgit2/xenial-x86:test'
+      environmentVariables: |
+       CC=gcc
+       LEAK_CHECK=valgrind
+
+- job: linux_xenial_x86
```
neithernut: ... and "clang" here.

.vsts-ci.yml
```diff
@@ -56,6 +56,32 @@ jobs:
...
+  displayName: 'Linux (Xenial; x86; Clang)'
+  pool:
+    vmImage: 'Ubuntu 16.04'
+  steps:
+  - template: ci/vsts-docker.yml
+    parameters:
+      qemu: 'true'
+      imageName: 'libgit2/xenial-x86:test'
+      environmentVariables: |
+       CC=gcc
```
neithernut: I suspect you want to have `CC=clang` instead.

ethomson:
Thanks @neithernut!

neithernut:
The current failures look like proper reports. Except for the MacOS build+test, which is canceled after an hour.

ethomson:
> The current failures look like proper reports. Except for the MacOS build+test, which is canceled after an hour.

Yes, the version of libssh2 that's in Xenial doesn't support newer ciphers that GitHub requires.  🙄  So I either need to move this up to Ubuntu Bionic, or build a version of our custom libssh2 for these other platforms, or skip the ssh tests altogether on these platforms.

.vsts-ci.yml
```diff
@@ -56,6 +56,30 @@ jobs:
...
+  displayName: 'Linux (Xenial; arm64)'
+  pool:
+    vmImage: 'Ubuntu 16.04'
+  steps:
+  - template: ci/vsts-docker.yml
+    parameters:
+      qemu: 'true'
+      imageName: 'libgit2/xenial-arm64:test'
+      environmentVariables: |
+       SKIP_PROXY_TESTS=true
```
pks-t: Is this all supposed to be part of our "main" CI infrastructure? If so, by how much does it increase our CI build times? If the delay is quite noticable, I'd vote to have this as a nightly job instead, as we have said in past discussions.

.vsts-ci.yml
```diff
@@ -57,7 +57,7 @@ jobs:
        LEAK_CHECK=valgrind
 
 - job: linux_xenial_arm32
-  displayName: 'Linux (Xenial; arm32)'
+  displayName: 'Linux (Xenial; arm32; GCC)'
```
pks-t: Shouldn't these jobs then also say whether they use OpenSSL or mbedTLS, if we aim for consistency?

tests/buf/oom.c
```diff
@@ -40,6 +41,9 @@ void test_buf_oom__grow(void)
 	cl_assert(git_buf_oom(&buf));
 
 	git_buf_dispose(&buf);
+#else
+    cl_skip();
+#endif
```
pks-t: You know, we should just use our custom allocators here. Create a custom test allocater that only has $n$ bytes available, plug it in and then run this test. This should result in a "real" OOM without us having to trick the host. I also think it would be quite helpful in other tests to verify exceptional behaviour.

Obviously, that doesn't have to be part of this PR.

libgit2_clar.supp
```diff
@@ -56,13 +56,53 @@
...
+	obj:*libssh2.so*
+}
+
+{
+	ignore-libssh2-gcrypt-pksign-leak
+	Memcheck:Leak
+	...
+	fun:gcry_pk_sign
+	obj:*libssh2.so*
+}
```
pks-t: I really cannot imagine that libbsh2 or libgcrypt is to blame. I mean, I've tried multiple times to debug this issue and see whether we fail to use the APIs correctly, without any finds. But I just cannot imagine them being so lenient with regards to memory leaks.

tests/buf/oom.c
ethomson: Ooooooh, that's a good idea.

.vsts-ci.yml
ethomson: I 100% agree.  I was just hoping to iterate on these by putting them in the PR validation configuration so that they would get built.  (There are a handful of other ways to do this, I could have set up a new pipelines configuration, but I'm lazy.)

ethomson:
I've also [created a repository](https://github.com/libgit2/docker-build) that containers the Dockerfiles that we use to create these images.

tests/buf/oom.c
ethomson: Addressing this in a separate pullrequest - forthcoming.

pks-t:
Cool, this is great! Is there any particular reason why you feel like the Dockerfiles shouldn't be part of libgit2 directly? We already have a "ci" directory, so I'd feel like it's the perfect place for them to live in.

ethomson:
> Cool, this is great! Is there any particular reason why you feel like the Dockerfiles shouldn't be part of libgit2 directly? We already have a "ci" directory, so I'd feel like it's the perfect place for them to live in.

🤔  No.  I have no preference either way.  I'd be happy to move them into `ci` if that's what makes the most sense.

pks-t:
I think it makes sense to have them as part of libgit2. It's not like they make any sense as standalone files outside of the scope of libgit2, and they are part of our CI setup. Furthermore, I don't think they're much of a maintenance burden and don't add much to libgit2's size. So only pros, no cons, I think.

PR #4807, opened 2018-09-12
Index API updates for consistency

ethomson:
I've been angry at `index.h` for a little while now and would like to make some (non-breaking) fixes for better consistency with the project.  I'll pile them in here.

### DIFF
--- a/include/git2/index.h
+++ b/include/git2/index.h
@@ -37,11 +37,11 @@ typedef struct {
  * "Documentation/technical/index-format.txt").
  *
  * The `flags` field consists of a number of bit fields which can be
- * accessed via the first set of `GIT_IDXENTRY_...` bitmasks below.  These
- * flags are all read from and persisted to disk.
+ * accessed via the first set of `GIT_INDEX_ENTRY_...` bitmasks below.
+ * These flags are all read from and persisted to disk.
  *
  * The `flags_extended` field also has a number of bit fields which can be
- * accessed via the later `GIT_IDXENTRY_...` bitmasks below.  Some of
+ * accessed via the later `GIT_INDEX_ENTRY_...` bitmasks below.  Some of
  * these flags are read from and written to disk, but some are set aside
  * for in-memory only reference.
  *
@@ -76,24 +76,25 @@ typedef struct git_index_entry {
  * value both in memory and on disk.  You can use them to interpret the
  * data in the `flags`.
  */
-#define GIT_IDXENTRY_NAMEMASK  (0x0fff)
-#define GIT_IDXENTRY_STAGEMASK (0x3000)
-#define GIT_IDXENTRY_STAGESHIFT 12
+
+#define GIT_INDEX_ENTRY_NAMEMASK  (0x0fff)
+#define GIT_INDEX_ENTRY_STAGEMASK (0x3000)
+#define GIT_INDEX_ENTRY_STAGESHIFT 12
 
 /**
  * Flags for index entries
  */
 typedef enum {
-	GIT_IDXENTRY_EXTENDED  = (0x4000),
-	GIT_IDXENTRY_VALID     = (0x8000),
-} git_indxentry_flag_t;
+	GIT_INDEX_ENTRY_EXTENDED  = (0x4000),
+	GIT_INDEX_ENTRY_VALID     = (0x8000),
+} git_index_entry_flag_t;
 
-#define GIT_IDXENTRY_STAGE(E) \
-	(((E)->flags & GIT_IDXENTRY_STAGEMASK) >> GIT_IDXENTRY_STAGESHIFT)
+#define GIT_INDEX_ENTRY_STAGE(E) \
+	(((E)->flags & GIT_INDEX_ENTRY_STAGEMASK) >> GIT_INDEX_ENTRY_STAGESHIFT)
 
-#define GIT_IDXENTRY_STAGE_SET(E,S) do { \
-	(E)->flags = ((E)->flags & ~GIT_IDXENTRY_STAGEMASK) | \
-		(((S) & 0x03) << GIT_IDXENTRY_STAGESHIFT); } while (0)
+#define GIT_INDEX_ENTRY_STAGE_SET(E,S) do { \
+	(E)->flags = ((E)->flags & ~GIT_INDEX_ENTRY_STAGEMASK) | \
+		(((S) & 0x03) << GIT_INDEX_ENTRY_STAGESHIFT); } while (0)
 
 /**
  * Bitmasks for on-disk fields of `git_index_entry`'s `flags_extended`
@@ -101,7 +102,7 @@ typedef enum {
  * In memory, the `flags_extended` fields are divided into two parts: the
  * fields that are read from and written to disk, and other fields that
  * in-memory only and used by libgit2.  Only the flags in
- * `GIT_IDXENTRY_EXTENDED_FLAGS` will get saved on-disk.
+ * `GIT_INDEX_ENTRY_EXTENDED_FLAGS` will get saved on-disk.
  *
  * Thee first three bitmasks match the three fields in the
  * `git_index_entry` `flags_extended` value that belong on disk.  You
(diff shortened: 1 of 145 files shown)


include/git2/index.h
```diff
@@ -86,7 +86,7 @@ typedef struct git_index_entry {
 typedef enum {
 	GIT_IDXENTRY_EXTENDED  = (0x4000),
 	GIT_IDXENTRY_VALID     = (0x8000),
-} git_indxentry_flag_t;
+} git_idxentry_flag_t;
```
pks-t: Huh. This very much does look breaking to me?

pks-t: Sorry, a bit more constructive: what about also providing a second typedef? `typedef git_indxentry_flag_t git_idxentry_flag_t` so that nobody needs to change his code?

tiennou: Maybe, but wouldn't that introduce warnings in user code anyway because of the mismatch ? Thinking about it, I don't think it will, because the usage seems to be limited to bitwise operations. I'm just concerned that a `-Werror` downstream will have to change the code anyways.

Personally, I'd go for `git_index_entry_flag_t` for consistency, as that's the name of the struct, and it's part of the `index` system. I'm fine with just dropping the "n" though, since that's worst 😉.

ethomson: Sorry, yes, this was very much a WIP that is not at all ready for review.  Changing the title to reflect that.

ethomson: As far as the naming goes, I too would prefer `git_index_entry_...` except that the rest of the things in this file are named `git_idxentry`.  Possible that we can get away without breakage here, possibly not.

Appreciate the thoughts about `-Werror`.  I'm not sure that we've ever said that we guarantee that people can avoid warnings using libgit2 (and, indeed, you can see that the 32 bit platforms have loads of them that we should deal with but haven't yet).  But as that's something to strive for it's something interesting to keep in mind.

ethomson:
Okay, I've updated this to update the names for this prefixed with `git_idx_...` or `git_indx...`, moving them to `git_index_...`.  I've also updated the name of `git_otype` to be `git_object_t`.

Background:  I'm playing around with some automated tools to extract our APIs for a binding and having improved consistency in our naming would be 👍 .

I've added `#define old_name new_name` to ensure that we don't break consumers.  This allows people to continue using their existing code referencing the old names, and to do so without warnings.  That is to say, our _new_ APIs will now accept a `git_object_t`.  If we simply keep the old `enum`s around, then compilers will see that old code is sending a `git_otype` to a `git_object_t` and warn.  Although I don't think we've ever defined "warnings" as a a breaking API change, we can certainly avoid them by simply redefining the old enums to the new ones.  eg:

    #define GIT_OBJ_COMMIT GIT_OBJECT_COMMIT

so that compilers will see users sending values in the `git_object_t` enum even though they're using the old names.

While we're here, we can also clean up some unused values that users should not be relying on and in fact represent internal structures that we likely do not want to publish, even if we _do_ start using them internally.

For example, `GIT_OBJ__EXT1` should not really be part of the public API.  This PR keeps `GIT_OBJ__EXT1` for backward compatibility with anybody who is using it, despite its uselessness.  (One could imagine another automated tool that has taken it from our API.)  However, we do not have a corresponding `GIT_OBJECT__EXT1` for the new API.

ethomson:
These first commits change the API definition only.  All our own callers are using the old names; I wanted to open this with only those changes to ensure a CI pass with no warnings to address @tiennou's (very reasonable) concerns.

ethomson:
Anyway, I'd love some feedback @tiennou and @pks-t.  One question: are you happy with these in a new section in the existing files?  Or should they be moved to `deprecated.h`?  I didn't do that but I think I might prefer it.

include/git2/index.h
```diff
@@ -37,11 +37,11 @@ typedef struct {
  * "Documentation/technical/index-format.txt").
  *
  * The `flags` field consists of a number of bit fields which can be
- * accessed via the first set of `GIT_IDXENTRY_...` bitmasks below.  These
- * flags are all read from and persisted to disk.
+ * accessed via the first set of `GIT_INDEX_ENTRY_...` bitmasks below.
+ * These * flags are all read from and persisted to disk.
```
pks-t: Small nit: "*" inside of the sentence

include/git2/index.h
```diff
@@ -807,6 +808,49 @@ GIT_EXTERN(void) git_index_conflict_iterator_free(
...
+#define GIT_IDXENTRY_VALID             GIT_INDEX_ENTRY_VALID
+
+#define GIT_IDXENTRY_STAGE(E)          GIT_INDEX_ENTRY_STAGE(E)
+#define GIT_IDXENTRY_STAGE_SET(E,S)    GIT_INDEX_ENTRY_STAGE_SET(E,S)
+
+/* The git_idxentry_extended_flag_t enum */
+#define GIT_IDXENTRY_INTENT_TO_ADD     GIT_INDEX_ENTRY_INTENT_TO_ADD
+#define GIT_IDXENTRY_SKIP_WORKTREE     GIT_INDEX_ENTRY_SKIP_WORKTREE
+#define GIT_IDXENTRY_EXTENDED2         GIT_INDEX_ENTRY_EXTENDED2
+#define GIT_IDXENTRY_EXTENDED_FLAGS   (GIT_INDEX_ENTRY_INTENT_TO_ADD | GIT_INDEX_ENTRY_SKIP_WORKTREE)
```
pks-t: Why not just define it to `GIT_INDEX_ENTRY_EXTENDED_FLAGS`?

include/git2/index.h
```diff
@@ -116,22 +116,10 @@ typedef enum {
...
-	GIT_INDEX_ENTRY_UNHASHED          =  (1 << 5),
-	GIT_INDEX_ENTRY_WT_REMOVE         =  (1 << 6), /**< remove in work directory */
-	GIT_INDEX_ENTRY_CONFLICTED        =  (1 << 7),
-
-	GIT_INDEX_ENTRY_UNPACKED          =  (1 << 8),
-	GIT_INDEX_ENTRY_NEW_SKIP_WORKTREE =  (1 << 9),
+
+	GIT_INDEX_ENTRY_EXTENDED_FLAGS =  (GIT_INDEX_ENTRY_INTENT_TO_ADD | GIT_INDEX_ENTRY_SKIP_WORKTREE),
+
+    GIT_INDEX_ENTRY_UPTODATE       =  (1 << 2),
```
pks-t: Indentation is off here

include/git2/types.h
```diff
@@ -68,17 +68,17 @@ typedef int64_t git_time_t;
...
+	GIT_OBJECT_BAD =      -1, /**< Object is invalid. */
+	GIT_OBJECT__EXT1 =     0, /**< Reserved. */
+	GIT_OBJECT_COMMIT =    1, /**< A commit object. */
+	GIT_OBJECT_TREE =      2, /**< A tree (directory listing) object. */
+	GIT_OBJECT_BLOB =      3, /**< A file revision object. */
+	GIT_OBJECT_TAG =       4, /**< An annotated tag object. */
+	GIT_OBJECT__EXT2 =     5, /**< Reserved. */
+	GIT_OBJECT_OFS_DELTA = 6, /**< A delta, base is given by an offset. */
+	GIT_OBJECT_REF_DELTA = 7, /**< A delta, base is given by object id. */
+} git_object_t;
```
pks-t: I'm not sure I like `git_object_t`. For me, `_t` usually just says "This is a typedef" instead of "This is a type". So I'd rather vote for either `git_object_type` or `git_object_type_t`.

ethomson: Yeah, I struggled with this one a bit.  I agree with you but we're reasonably consistent with `_t` meaning "this is a type`".  A few examples:

* [`git_ref_t`](https://github.com/libgit2/libgit2/blob/master/include/git2/types.h#L202)
* [`git_branch_t`](https://github.com/libgit2/libgit2/blob/master/include/git2/types.h#L209)
* [`git_filemode_t`](https://github.com/libgit2/libgit2/blob/master/include/git2/types.h#L219)
* [`git_cert_t`](https://github.com/libgit2/libgit2/blob/master/include/git2/types.h#L316)
* [`git_submodule_update_t `](https://github.com/libgit2/libgit2/blob/master/include/git2/types.h#L377)
...

(It's not perfect, eg `git_credtype_t`, which does have values of `GIT_CREDTYPE_...` so that's nice, but they might have been more consistent as `GIT_CREDENTIAL_USERNAME`...)  But anyway, this feels consistent even if it wasn't my preference to begin with.

ethomson:
Thanks - agree with `GIT_IDXENTRY_EXTENDED_FLAGS` -> `GIT_INDEX_ENTRY_EXTENDED_FLAGS` and appreciate the formatting issues.

ethomson:
OK, this seems like a change that isn't contentious.  I'll update internal callers to use the new names, possible tomorrow (Thanksgiving) but more likely over the weekend.

include/git2/types.h
pks-t: Fair enough :+1:

ethomson:
> Looks good to me. In the ideal case, we'd add deprecation notices to all those old typedefs. But I'm not sure I'd want to do this immediately, as downstream users are likely to rage in that case as their code will be full of warnings. And I don't even know if it's possible to deprecate macros 🤔

I'm in agreement with you.  Let's hold off and do it when we remember which will probably be far enough from now that it seems like a reasonable time to do it.  :P

ethomson:
OK, I rebased this.  Going to :shipit:

Issue #4428, opened 2017-12-01
git_revwalk_next is slow on first iteration since v0.25


rgburke:
git_revwalk_next is slow on its first iteration for versions 0.25, 0.26 and master when compared to v0.24. This is especially noticeable on large repositories.

### Reproduction steps
Run the following program on the Linux kernel source tree (https://github.com/torvalds/linux):

```c
#include <stdio.h>
#include <git2.h>

int main(int argc, char *argv[]) {
        git_libgit2_init();

        git_repository *repo;
        if (git_repository_open(&repo, ".git")) {
                fprintf(stderr, "Failed to open repository\n");
                return 1;
        }

        git_oid oid;
        git_oid_fromstr(&oid, "a0651c7fa2c088a605f63792279859608ed7f2c8");

        git_revwalk *revwalk;
        git_revwalk_new(&revwalk, repo);
        git_revwalk_sorting(revwalk, GIT_SORT_TIME);
        git_revwalk_push(revwalk, &oid);

        if (!git_revwalk_next(&oid, revwalk)) {
                git_commit *commit;

                if (git_commit_lookup(&commit, repo, &oid)) {
                        fprintf(stderr, "Failed to lookup commit\n");
                        return 1;
                }

                printf("%s\n", git_commit_message(commit));
                git_commit_free(commit);
        }

        git_revwalk_free(revwalk);
        git_repository_free(repo);

        return 0;
}
```

### Expected behavior
When using v0.24 this program runs in a couple of milliseconds:
```
richard@DevNew:~/code/linux$ time ./revwalktest >/dev/null

real	0m0.006s
user	0m0.000s
sys	0m0.006s
```

### Actual behavior
When using v0.25, v0.26 or master this program has not completed after 1 minute:
```
richard@DevNew:~/code/linux$ timeout 60s ./revwalktest
richard@DevNew:~/code/linux$ echo $?
124
```

### Version of libgit2 (release number or SHA1)
v0.25, v0.26 & master

### Operating system(s) tested
```
Linux DevNew 4.13.0-17-generic #20-Ubuntu SMP Mon Nov 6 10:04:08 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

tiennou:
I've tried to reproduce this by borrowing and hacking our `rev-list` example, but I don't quite get the same results :

[...]

PR #4809, opened 2018-09-17
Fix revwalk limiting regression

### LINKED ISSUE #4740: Performance regression in `revwalk` API
### Explanation

Atom uses libgit2's `revwalk` API to compute the number of commits by which the user's current branch is 'ahead' and 'behind' its corresponding remote branch. After upgrading from libgit2 0.24 to 0.26, we're seeing this code take a long time to execute, even when the branch is up-to-date with its remote (no commits ahead or behind).

### Reproduction steps

1. Clone the [Atom](https://github.com/atom/atom) repository

1. Create a C file called `test.c` with this content:

```c
#include "git2.h"
#include <stdio.h>
#include <assert.h>

unsigned count_commits_between_refs(
  git_repository *repo,
  const char *ref1,
  const char *ref2
) {
  git_oid oid1, oid2;
  assert(git_reference_name_to_id(&oid1, repo, ref1) == GIT_OK);
  assert(git_reference_name_to_id(&oid2, repo, ref2) == GIT_OK);

  git_oid merge_base;
  assert(git_merge_base(&merge_base, repo, &oid1, &oid2) == GIT_OK);

  git_revwalk *revwalk;
  assert(git_revwalk_new(&revwalk, repo) == GIT_OK);

  git_revwalk_push(revwalk, &oid1);
  git_revwalk_hide(revwalk, &oid2);

  unsigned result = 0;
  git_oid current_commit;
  while (git_revwalk_next(&current_commit, revwalk) == GIT_OK) result++;
  git_revwalk_free(revwalk);

  return result;
}

int main() {
  git_libgit2_init();

  git_repository* repo;
  assert(git_repository_open_ext(&repo, getenv("PWD"), 0, NULL) == GIT_OK);

  for (unsigned i = 0; i < 50; i++) {
    const char *ref1 = "refs/heads/master";
    const char *ref2 = "refs/remotes/origin/master";

    unsigned ahead_count = count_commits_between_refs(repo, ref1, ref2);
    unsigned behind_count = count_commits_between_refs(repo, ref2, ref1);
    printf("%u ahead, %u behind\n", ahead_count, behind_count);
  }

  return 0;
}
```

3. Compile this program against the branch of libgit2. Side note: when trying to build libgit2 on macOS, I had trouble linking against openssl. So I build the static libraries using our node.js wrapper module: [git-utils](https://github.com/atom/git-utils) and compiled against them like this:

```sh
clang                  \
  -I libgit2/include   \
  zlib.a               \
  http_parser.a        \
  git2.a               \
  test.c
```

4. Run this program

### Expected behavior

The program should exit immediately, outputting "0 ahead, 0 behind" repeated 50 times.

### Actual behavior

The program takes 4 seconds to run.

The head of master as of today: 42f8384045ca325c6885be21a89c3d8027995e03. I was also able to reproduce this against libgit2 0.26.0.

The problem does *not* happen with libgit2 0.24.0.

Mac OS 10.13.16

/cc @carlosmn

carlosmn:
When porting, we overlooked that the difference between git's and our's time
representation and copied their way of getting the max value.
    
Unfortunately git was using unsigned integers, so `~0ll` does correspond to
their max value, whereas for us it corresponds to `-1`. This means that we
always consider the last date to be smaller than the current commit's and always
think commits are interesting.
    
Change the initial value to the macro that gives us the maximum value on each
platform so we can accurately consider commits interesting or not.

The second commit is mostly just to reduce the drift from git, the actual perf difference should be negligible.

This fixes #4740 and we should backport it as it fixes a regression.

### DIFF
--- a/src/revwalk.c
+++ b/src/revwalk.c
@@ -388,10 +388,16 @@ static int still_interesting(git_commit_list *list, int64_t time, int slop)
 	if (!list)
 		return 0;
 
+	/*
+	 * If the destination list has commits with an earlier date than our
+	 * source, we want to reset the slop counter as we're not done.
+	 */
+	if (time <= list->item->time)
+		return SLOP;
+
 	for (; list; list = list->next) {
 		/*
-		 * If the destination list has commits with an earlier date than
-		 * our source or if it still contains interesting commits we
+		 * If the destination list still contains interesting commits we
 		 * want to continue looking.
 		 */
 		if (!list->item->uninteresting || list->item->time > time)
--- a/src/revwalk.c
+++ b/src/revwalk.c
@@ -405,7 +405,7 @@ static int still_interesting(git_commit_list *list, int64_t time, int slop)
 static int limit_list(git_commit_list **out, git_revwalk *walk, git_commit_list *commits)
 {
 	int error, slop = SLOP;
-	int64_t time = ~0ll;
+	int64_t time = INT64_MAX;
 	git_commit_list *list = commits;
 	git_commit_list *newlist = NULL;
 	git_commit_list **p = &newlist;


neithernut:
I've just ran and timed the reproducers from #4428 and a trimmed version of the reproducer in #4740 (single iteration) on the linux kernel repo.

----

The times reported for libgit2 built from master (bc34cb63ff5b2bc51bcdec7dfc41202dfc34e97f):

#4428:

    real	0m18.426s
    user	0m18.187s
    sys	0m0.227s

#4740:

    real	0m36.638s
    user	0m36.372s
    sys	0m0.236s

----

For the current tip (12a1790d8e71087056d2b2de936ddae439e1d94c)

#4428:

    real	0m18.413s
    user	0m18.156s
    sys	0m0.246s

#4740:

    real	0m0.007s
    user	0m0.005s
    sys	0m0.002s

----

Honestly, I thought I did something wrong when I performed the second measurement for the #4428 reproducer, but apparently this particular regression is not resolved. However, the performance is much better for #4740.

carlosmn:
#4428 doesn't actually seem like a regression, but a fix and a misunderstanding of what different flags ought to do.

ethomson:
Ouch!  Thanks for the fix, @carlosmn.

src/revwalk.c
```diff
@@ -405,7 +405,7 @@ static int still_interesting(git_commit_list *list, int64_t time, int slop)
 static int limit_list(git_commit_list **out, git_revwalk *walk, git_commit_list *commits)
 {
 	int error, slop = SLOP;
-	int64_t time = ~0ll;
+	int64_t time = INT64_MAX;
```
pks-t: I'm surprised none of our tests catched this :(

carlosmn: It's just a performance regression, as all this does is make us think that we never went back far enough. If we had perf tests, maybe we would have caugh tit.

pks-t:
On Sat, Sep 22, 2018 at 12:32:52PM -0700, Carlos Martín Nieto wrote:
> carlosmn commented on this pull request.
> > @@ -405,7 +405,7 @@ static int still_interesting(git_commit_list *list, int64_t time, int slop)
>  static int limit_list(git_commit_list **out, git_revwalk *walk, git_commit_list *commits)
>  {
>  	int error, slop = SLOP;
> -	int64_t time = ~0ll;
> +	int64_t time = INT64_MAX;
> 
> It's just a performance regression, as all this does is make us think that we never went back far enough. If we had perf tests, maybe we would have caugh tit.

Yeah, I realized that at a later point, too. But thanks for
clarifying!

PR #4813, opened 2018-09-18
Rename "VSTS" to "Azure DevOps" and "Azure Pipelines"

ethomson:
Visual Studio Team Services is now a family of tools - the build and release pipelines are now named "Azure Pipelines", and the overall family is named "Azure DevOps".  Update the CI and README to reflect this.


carlosmn:
All this rebranding is a neat way to drive up Microsoft's stats on FLOSS contributions ;)

carlosmn:
It doesn't look like CI triggered on this one, do we need to change something on that side?

ethomson:
Yes - I’ll need to change the CI configuration at the time that we merge it so that it knows to read the new yaml file.

Issue #4077, opened 2017-01-12
Inserting a file into an index which overwrites a directory silently discards the file when a tree is produced.


Parakleta:
### Reproduction steps

```python
from tempfile import mkdtemp
from pygit2 import init_repository, Index, IndexEntry, GIT_FILEMODE_BLOB

repo = init_repository(mkdtemp(),bare=True)
blob1 = repo.create_blob("data1")
blob2 = repo.create_blob("data2")
blob3 = repo.create_blob("data3")

idx = Index()
idx.add(IndexEntry('afile',blob1,GIT_FILEMODE_BLOB))
idx.add(IndexEntry('pathto/afile',blob2,GIT_FILEMODE_BLOB))
idx.add(IndexEntry('pathto',blob3,GIT_FILEMODE_BLOB))

for entry in idx:
    print(entry.path,entry.id)
```

### Expected behavior

```
('afile', 0abc8f194801d3d07af700bae67026ed2695ec59)
('pathto', cdca2c1e94f3d30ed4e7822b2da91c59792899b4)
```

### Actual behavior

```
('afile', 0abc8f194801d3d07af700bae67026ed2695ec59)
('pathto', cdca2c1e94f3d30ed4e7822b2da91c59792899b4)
('pathto/afile', fee93d14fc11d09aeffc1494c3d8cb5b3635b9f0)
```

### Version of libgit2 (release number or SHA1)

Version 0.24.2 (through pygit2 of the same version)

### Operating system(s) tested

FreeBSD

### Other Information

The bug was introduced in commit 21515f228b739a3ecd2237bafbba50e8d219d8dd.  The mechanism is that previously when `index_insert` was run it would return the insertion position of the new entry from the call to `index_find` even if the entry wasn't found.  This position information is required for `check_file_directory_collision` and more specifically as the starting position for the search in `has_file_name`.  After the changes introduced in that commit the call to `index_existing_and_best` which now wraps the call to `index_find` discards the position information if the exact entry wasn't found.

I also believe that there is a mistake in the behaviour of `has_file_name` and `has_dir_name` in that they both return an error which causes the `index_insert` to fail even if its variable `replace` (renamed `ok_to_replace` in the `has_*_name` functions) is set.  In this situation they both remove the replaced entries but then due to the error the replacement is never inserted.  I think their `ok_to_replace` code should change to something like:

```c
if (!ok_to_replace || index_remove_entry(index, pos) < 0) {
    retval = -1;
    break;
}
```

[...]

PR #4814, opened 2018-09-18
index: failing more thorough test for adding colliding path

carlosmn:
We've had tests since v0.21 that we error out when trying to add a colliding
path to the index, but it turns out that right now we remove the
already-existing path from the index instead of leaving it alone.

I was looking at #4077 again and it looks like all we do right now is test that we return an error, but (at least) the tree we write out does not include the original path we inserted. I realise now that it would probably be a better test to look at the index instead of the tree it creates, but if someone wants to tackle this, they can do that I guess.


ethomson:
Fixed via #4818

PR #4816, opened 2018-09-20
online::clone: free url and username before resetting

ethomson:
Before resetting the url and username, ensure that we free them in case they were set by environment variables.  (This is definitely a bit ugly, but since the callbacks make use of the globals, this was the most straightforward way to keep that consistent without leaking.)

### DIFF
--- a/tests/online/clone.c
+++ b/tests/online/clone.c
@@ -276,6 +276,9 @@ static int cred_failure_cb(
 
 void test_online_clone__cred_callback_failure_return_code_is_tunnelled(void)
 {
+	git__free(_remote_url);
+	git__free(_remote_user);
+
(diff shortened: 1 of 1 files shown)


tests/online/clone.c
```diff
@@ -276,6 +276,9 @@ static int cred_failure_cb(
 
 void test_online_clone__cred_callback_failure_return_code_is_tunnelled(void)
 {
+	git__free(_remote_url); _remote_url = NULL;
+	git__free(_remote_user); _remote_user = NULL;
```
pks-t: No need to `NULL` those variables, as you unconditionally set them to the result of `git__strdup` in the next statements. Same applies to the other test

ethomson: Right you are.  Fixed.

PR #4798, opened 2018-09-05
Introduce VSTS CI for the v0.27 branch

ethomson:
Update CI to use VSTS for the v0.27 branch instead of the Travis + AppVeyor combination.

PR #4817, opened 2018-09-20
Introduce Azure Pipelines CI for the v0.26 branch

ethomson:
Like #4798, this introduces Azure Pipelines for the v0.26 maintenance branch.  This is mostly cherry-picks of the CI infrastructure, omitting a few changes (like the names of the ctest configurations to avoid breaking anybody who relies on them) and a few changes like plugging memory leaks that are now noticed.


pks-t:
This is included in v0.26.8

PR #4818, opened 2018-09-21
Index collision fixes

### LINKED ISSUE #4077: Inserting a file into an index which overwrites a directory silently discards the file when a tree is produced.

pks-t:
This fixes our code in the following two scenarios:

- adding an index entry "a" when "a/b" exists
- adding an index entry "a/b" when "a" exists

In both cases, git.git simply removes the obstructing index entry and replaces it with the new one. But we currently fail to do so correctly -- instead, we just remove the old one, error out and then leave the caller with an index that lost data. Not good. So this PR fixes our code to also correctly add the new index entries as expected.

As always, please have a look at the commit messages to get some more rationale about the changes.

Supersedes #4814, fixes #4077


src/index.c
```diff
@@ -1337,57 +1331,57 @@ static int index_insert(
...
-	/* ensure that the given id exists (unless it's a submodule) */
-	if (!error && !trust_id && INDEX_OWNER(index) &&
-		(entry->mode & GIT_FILEMODE_COMMIT) != GIT_FILEMODE_COMMIT) {
+	/* Ensure that the given id exists (unless it's a submodule) */
+	if (!trust_id && INDEX_OWNER(index) &&
+	    (entry->mode & GIT_FILEMODE_COMMIT) != GIT_FILEMODE_COMMIT) {
 
 		if (!git_object__is_valid(INDEX_OWNER(index), &entry->id,
-			git_object__type_from_filemode(entry->mode)))
+					  git_object__type_from_filemode(entry->mode))) {
```
carlosmn: This should be aligned with the `if (`, no?

src/index.c
```diff
@@ -1337,57 +1331,57 @@ static int index_insert(
...
 
 	assert(index && entry_ptr);
 
 	entry = *entry_ptr;
 
-	/* make sure that the path length flag is correct */
-	path_length = ((struct entry_internal *)entry)->pathlen;
-	index_entry_adjust_namemask(entry, path_length);
+	/* Make sure that the path length flag is correct */
+	index_entry_adjust_namemask(entry, ((struct entry_internal *)entry)->pathlen);
```
carlosmn: This is personal preference, but the old version looks clearer to me as we're not doing the casting and dereferencing in the function call.

carlosmn:
Looks reasonable; the behaviour of the "check" functions is indeed confusing.

src/index.c
pks-t: Okay, I've dropped this commit. I prefer not having the variable, but taste is different and I don't care that much

PR #4819, opened 2018-09-28
Configuration variables can appear on the same line as the section header

carlosmn:
This is an quirk of the config format, but you are allowed to do

    [some] var = true

and have `some.var` exist. Our parser currently assumes everything is line-based and completely ignores anything past the closing bracket.

So this changes the section parser functions to return how far we should advance and then we restart. For the "normal" case, we'll detect the rest is an empty line and immediately jump to the next line and if we do find this situation we'll run the loop as though this were a brand new line.

### DIFF
--- a/src/config_parse.c
+++ b/src/config_parse.c
@@ -63,6 +63,7 @@ static int parse_section_header_ext(git_config_parser *reader, const char *line,
 {
 	int c, rpos;
 	char *first_quote, *last_quote;
+	const char *line_start = line;
 	git_buf buf = GIT_BUF_INIT;
 	size_t quoted_len, alloc_len, base_name_len = strlen(base_name);
 
@@ -139,7 +140,7 @@ end_parse:
 	}
 
 	*section_name = git_buf_detach(&buf);
-	return 0;
+	return &line[rpos + 2] - line_start; /* rpos is at the closing quote */
 
 end_error:
 	git_buf_dispose(&buf);
@@ -209,7 +210,7 @@ static int parse_section_header(git_config_parser *reader, char **section_out)
 	name[name_length] = 0;
 	*section_out = name;
 
-	return 0;
+	return pos;
 
 fail_parse:
 	git__free(line);
@@ -481,9 +482,13 @@ int git_config_parse(
 	skip_bom(ctx);
 
 	for (; ctx->remain_len > 0; git_parse_advance_line(ctx)) {
-		const char *line_start = parser->ctx.line;
-		size_t line_len = parser->ctx.line_len;
+		const char *line_start;
+		size_t line_len;
 		char c;
+
+	restart:
+		line_start = ctx->line;
+		line_len = ctx->line_len;
 
 		/*
 		 * Get either first non-whitespace character or, if that does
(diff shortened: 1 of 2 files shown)


src/config_parse.c
```diff
@@ -135,7 +135,7 @@ static int parse_section_header_ext(git_config_parser *reader, const char *line,
 	}
 
 	*section_name = git_buf_detach(&buf);
-	return 0;
+	return base_name_len + 1 /* SP */ + 1 /* " */ + rpos + 1 /* " */ + 1 /* ] */;
```
pks-t: Can we really assume to have "static" length here? What about e.g. `[foo          "bar"]` or `[foo "bar"    ]`? Isn't that accepted by the parser? I don't think it currently is, but that leaves me wondering if we're stricter than git.git itself. A quick check shows that `[foo    "bar"]` is accepted, while `[foo "bar" ]` is not. I don't think we accept the former right now.

If we were to update our logic, would the above calculation still be correct? If possible, I'd instead prefer to do some pointer arithmetic between the start of the line and our end position.

carlosmn: It is invalid to have a space between the closing quotes and the closing bracket. Both git and libgit2 refuse to parse such a file, so we should be fine.

I do agree that a more direct calculation would be nicer to have. The different sub-strings we use here made it a bit awkward so I did this to get the number. But again I do think the math will end up working out.

tests/config/read.c
```diff
@@ -759,6 +759,25 @@ void test_config_read__bom(void)
 	git_buf_dispose(&buf);
 }
 
+void test_config_read__single_line(void)
+{
+	git_buf buf = GIT_BUF_INIT;
+	git_config *cfg;
+
+	cl_set_cleanup(&clean_test_config, NULL);
+	cl_git_mkfile("./testconfig", "[some] var = value\n[some \"OtheR\"] var = value");
```
peff: You might want to have one of these drop the space between the header and the value (e.g., `[some]var = value`) for variety. From my reading of the code, that should be handled just fine, but it's yet another funny corner case.

carlosmn:
Rebased and fixed up to add a no-whitespace test as well as a hopefully more straightforward way of figuring out the section header length.

tests/config/read.c
```diff
@@ -759,6 +759,36 @@ void test_config_read__bom(void)
...
+	cl_git_pass(git_config_open_ondisk(&cfg, "./testconfig"));
+	cl_git_pass(git_config_get_string_buf(&buf, cfg, "some.var"));
+	cl_assert_equal_s(buf.ptr, "value");
+
+	git_buf_clear(&buf);
+	cl_git_pass(git_config_get_string_buf(&buf, cfg, "some.OtheR.var"));
+	cl_assert_equal_s(buf.ptr, "value");
+
+	git_config_free(cfg);
+	cl_git_mkfile("./testconfig", "[some] var = value\n[some \"OtheR\"]var = value");
```
pks-t: Remark: now that we have the memory config backend, we could just use that one to do the tests instead of writing files. This PR predates that, though

PR #4824, opened 2018-10-01
fix check if blob is uninteresting when inserting tree to packbuilder

palmin:
Blobs that have been marked as uninteresting should not be inserted into packbuilder
when inserting a tree. The check as to whether a blob was uninteresting looked at
the status for the tree itself instead of the blob.

This could cause significantly larger packfiles.

### DIFF
--- a/src/pack-objects.c
+++ b/src/pack-objects.c
@@ -1676,7 +1676,7 @@ int insert_tree(git_packbuilder *pb, git_tree *tree)
 
 			break;
 		case GIT_OBJ_BLOB:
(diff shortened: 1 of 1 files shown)


pks-t:
Yeah, this was most definitely an error. We always retrieve the object for the tree we're currently inserting. As that tree cannot ever be uninteresting (as otherwise `insert_tree` wouldn't try to insert the blob in the first place), we always end up inserting the blob into the pack.

pks-t:
Thanks a lot!

Issue #4821, opened 2018-09-29
follow_and_lstat_link does not correctly follow symbolic links for UNC shares


GabeDeBacker:
### Reproduction steps
This was discovered using LibGit2Sharp, but the issue is actually in the native binaries.

Easiest way to set up is through the command prompt.
Copy the %userprofile%\.gitconfig to a UNC share (i.e. \\server\share\myGitConfig\.gitconfig)
Delete the %userprofile%\.gitconfig
mklink %USERPROFILE%\.gitconfig \\server\share\myGitConfig\.gitconfig

Executing the following code from LibGit2Sharp:
                var repo = new Repository(repoRoot);

                var matchesRequiredConfig= repo.Config.Any((config) =>
                {
                });

### Expected behavior
No exception.

### Actual behavior
Exception thrown: 'LibGit2Sharp.NotFoundException' in LibGit2Sharp.dll
could not find 'C:/Users/gabrield/.gitconfig' to stat: The system cannot find the path specified.

### Version of libgit2 (release number or SHA1)
v0.16.0-8972-g6311e886d

### Operating system(s) tested
x64 Windows 

### Additional information
After debugging the issue quite a bit, the issue stems from this stack:

size_t git_win32__canonicalize_path(wchar_t *str, size_t len) -> libgit2.git\src\win32\w32_util.c
static int getfinalpath_w(git_win32_path dest, const wchar_t *path) -> libgit2.git\src\win32\posix_w32.c
static int follow_and_lstat_link(git_win32_path path, struct stat* buf) -> libgit2.git\src\win32\posix_w32.c
int p_stat(const char* path, struct stat* buf) -> libgit2.git\src\win32\posix_w32.c

getfinalpath_w correctly gets the UNC path to the server (using the example above, it results in:
"\\?\UNC\server\share\myGitConfig\.gitconfig"

Then it calls win32__canonicalize_path, which results in stripping the "\\\\?\\UNC\\" from the path, resulting in "server\\share\\myGitConfig\\.gitconfig"

follow_and_lstat_link then calls GetFileAttributesExW with an invalid path of "server\share\myGitConfig\.gitconfig" which should be "\\\\server\share\myGitConfig\.gitconfig"

GabeDeBacker:
Wanted to clarify, the bat to GetFileAttributesExW should be "\\server\share\myGitConfig.gitconfig".

anon99:
I use Windows git with a windows share, containing git worktrees.
[...]

PR #4825, opened 2018-10-01
Fix issue #4821 with path canonicalization for Win32 paths

GabeDeBacker:
This pull request fixes issue #4821.

This issue occurs when a user has a reparse point (symbol link) for their global %userprofile%\.gitconfig file the points to a Windows UNC share.

The p_stat function correctly determines that the path is a symbolic link and uses the getfinalpath_w function to Windows file path that can be used with GetFileAttributesEx. The getfinalypath_w ultimately uses git_win32__canonicalize_path to strip any prefixes that may have been returned by GetFinalPathNameByHandleW (for example "\\\\?\\.\\UNC\\). However, when a UNC path is converted, the beginning "\\\\" characters where incorrectly stripped from the path, leaving an incorrect file name.

This libgit to be unable to locate and read the .gitconfig file.

### DIFF
--- a/src/win32/w32_util.c
+++ b/src/win32/w32_util.c
@@ -132,6 +132,8 @@ size_t git_win32__canonicalize_path(wchar_t *str, size_t len)
 	static const wchar_t dosdevices_prefix[] = L"\\\?\?\\";
 	static const wchar_t nt_prefix[] = L"\\\\?\\";
 	static const wchar_t unc_prefix[] = L"UNC\\";
+	static const wchar_t unc_canonicalized_prefix[] = L"\\\\";
+
 	size_t to_advance = 0;
 
 	/* "\??\" -- DOS Devices prefix */
@@ -150,8 +152,18 @@ size_t git_win32__canonicalize_path(wchar_t *str, size_t len)
 	/* "\??\UNC\", "\\?\UNC\" -- UNC prefix */
 	if (to_advance && len >= CONST_STRLEN(unc_prefix) &&
 		!wcsncmp(str + to_advance, unc_prefix, CONST_STRLEN(unc_prefix))) {
-		to_advance += CONST_STRLEN(unc_prefix);
-		len -= CONST_STRLEN(unc_prefix);
+		/** 
+		 * The proper Win32 path for a UNC share has "\\" at beginning of it
+		 * and looks like "\\server\share\<folderStructure>".
+		 * So, remove th UNC prefix, but leave room for a "\\"
+		 */ 
+		to_advance += (CONST_STRLEN(unc_prefix) - CONST_STRLEN(unc_canonicalized_prefix));
+		len -= (CONST_STRLEN(unc_prefix) - CONST_STRLEN(unc_canonicalized_prefix));
+		
+		/**
+		 * Place a "\\" in the string so the result is "\\server\\share\<folderStructure>"
+		 */
+		memmove(str + to_advance, unc_canonicalized_prefix, CONST_STRLEN(unc_canonicalized_prefix) * sizeof(wchar_t));
 	}
 
 	if (to_advance) {
--- a/tests/path/win32.c
+++ b/tests/path/win32.c
@@ -145,6 +145,22 @@ void test_canonicalize(const wchar_t *in, const wchar_t *expected)
 #endif
 }
 
+void test_path_git_win32__canonicalize_path(const wchar_t *in, const wchar_t *expected)
+{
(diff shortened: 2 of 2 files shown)


tests/path/win32.c
```diff
@@ -145,6 +145,22 @@ void test_canonicalize(const wchar_t *in, const wchar_t *expected)
 #endif
 }
 
+void test_path_git_win32__canonicalize_path(const wchar_t *in, const wchar_t *expected)
```
pks-t: This test's name looks like a unit test, as these need to conform to the pattern `test_testsuite__testname`. It is thus a bit confusing, as this is in fact a helper function. I'd propose to 1. make the function static, as it is only used in this file, and 2. to rename it to e.g. `test_canonicalize_path`.

tests/path/win32.c
```diff
@@ -186,6 +202,8 @@ void test_path_win32__canonicalize(void)
 	test_canonicalize(L"\\\\server\\\\share\\\\foo\\\\bar", L"\\\\server\\share\\foo\\bar");
 	test_canonicalize(L"\\\\server\\share\\..\\foo", L"\\\\server\\foo");
 	test_canonicalize(L"\\\\server\\..\\..\\share\\.\\foo", L"\\\\server\\share\\foo");
+
+	test_path_git_win32__canonicalize_path(L"\\\\?\\UNC\\server\\C$\\folder", L"\\\\server\\C$\\folder");
```
pks-t: All the above tests check whether the function `git_win32_path_canonicalize` works, whereas the new call checks whether `git_win32__canonicalize_path` functions as expected. I'd thus propose to move it into a new unit test (e.g. a function `void test_path_win32__canonicalize_path(void)`).

Our existing API really is a bit confusing. `git_win32_path_canonicalize` and `git_win32__canonicalize_path` sound like they should do the same thing. But the former mostly sanitizes slashes, while the latter will strip leading namespace prefixes. We should probably rename the second one, but that needn't be part of this pull request.

src/win32/w32_util.c
```diff
@@ -150,8 +152,18 @@ size_t git_win32__canonicalize_path(wchar_t *str, size_t len)
 	/* "\??\UNC\", "\\?\UNC\" -- UNC prefix */
 	if (to_advance && len >= CONST_STRLEN(unc_prefix) &&
 		!wcsncmp(str + to_advance, unc_prefix, CONST_STRLEN(unc_prefix))) {
-		to_advance += CONST_STRLEN(unc_prefix);
-		len -= CONST_STRLEN(unc_prefix);
+		/**
```
pks-t: `/** */` style comments are only used for strings that should show up in our documentation. But these comments here are only really relevant to developers of libgit2, not users of our API. So these should instead be `/* */` comments.

GabeDeBacker: Understood.

tests/path/win32.c
GabeDeBacker: Yep. Makes sense. Done.

tests/path/win32.c
GabeDeBacker: Agreed. Done.

tests/path/win32.c
```diff
@@ -145,6 +145,27 @@ void test_canonicalize(const wchar_t *in, const wchar_t *expected)
 #endif
 }
 
+void static test_canonicalize_path(const wchar_t *in, const wchar_t *expected)
```
pks-t: These here are switched ;)

```suggestion
static void test_canonicalize_path(const wchar_t *in, const wchar_t *expected)
```

tests/path/win32.c
```diff
@@ -129,7 +129,7 @@ void test_path_win32__absolute_from_relative(void)
 #endif
 }
 
-void test_canonicalize(const wchar_t *in, const wchar_t *expected)
+void static test_canonicalize(const wchar_t *in, const wchar_t *expected)
```
pks-t: Same over here:

```suggestion
static void test_canonicalize(const wchar_t *in, const wchar_t *expected)
```

pks-t:
Oh, just saw that Ed took this over. So you probably don't have to do anything, he's fixed your commits and they'll be merged as part of #4825 :)

ethomson:
Thanks again @GabeDeBacker for the fix and especially for the tests!  😀

Hope you don't mind that I wanted to fix up some other issues in our code base that I noticed while reviewing this.  (They were obviously our problems, not problems in this PR, but they confused me during the review so I wanted to get them cleaned up.)

Thanks again!

PR #4827, opened 2018-10-01
Documentation fixups

tiennou:
While looking at the worktree documentation, I managed to listen to a few of the "undocumented" warnings my tweaked docurium raises. So here's a few smallish additions/fixes to documentation comments.

### DIFF
--- a/include/git2/checkout.h
+++ b/include/git2/checkout.h
@@ -67,7 +67,7 @@ GIT_BEGIN_DECL
  * To emulate `git checkout -f`, use `GIT_CHECKOUT_FORCE`.
  *
  *
- * There are some additional flags to modified the behavior of checkout:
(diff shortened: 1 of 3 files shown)


tiennou:
Note that AFAICS the documentation is still built using "vanilla" docurium, which will likely result in no actual change to what's [shown](https://libgit2.org/libgit2/#HEAD/type/git_worktree_add_options) (vs the [tweaked version](https://tiennou.github.io/libgit2/#HEAD/type/git_worktree_add_options), which now autolinks known types on documented struct members).

pks-t:
These all look good to me, thanks for working on them. I cannot really help you with your docurium quest, though, as I've neither used it myself nor have I ever done anything with Ruby. :/

PR #4832, opened 2018-10-05
config_file: properly ignore includes without "path" value

pks-t:
config_file: properly ignore includes without "path" value

In case a configuration includes a key "include.path=" without any
value, the generated configuration entry will have its value set to
`NULL`. This is unexpected by the logic handling includes, and as soon
as we try to calculate the included path we will unconditionally
dereference that `NULL` pointer and thus segfault.

Fix the issue by returning early in both `parse_include` and
`parse_conditional_include` in case where the `file` argument is `NULL`.
Add a test to avoid future regression.

The issue has been found by the oss-fuzz project, issue 10810.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -664,6 +664,9 @@ static int parse_include(git_config_parser *reader,
 	char *dir;
 	int result;
 
+	if (!file)
+		return 0;
+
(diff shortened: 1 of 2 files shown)


pks-t:
This has been released as part of v0.26.7 and v0.27.5

PR #4836, opened 2018-10-05
Smart packet security fixes

pks-t:
These fixes were part of v0.26.7 and v0.27.5


pks-t:
This has already been discussed in private and released as part of v0.26.7 and v0.27.5

PR #4837, opened 2018-10-05
submodule: ignore path and url attributes if they look like options

pks-t:
These can be used to inject options in an implementation which performs a
recursive clone by executing an external command via crafted url and path
attributes such that it triggers a local executable to be run.

The library is not vulnerable as we do not rely on external executables but a
user of the library might be relying on that so we add this protection.

This matches this aspect of git's fix for CVE-2018-17456.

---

This has been released as part of v0.26.7 and v0.27.5


pks-t:
This has been discussed in private already

PR #4839, opened 2018-10-06
ignore unsupported http authentication contexts

palmin:
``auth_context_match`` returns -1 when the authentication context is not supported,
but this made ``parse_authenticate_response`` fail in situations where some authentication
contexts are supported and others are not.

``parse_authenticate_response`` will instead ignore authentication contexts where
``auth_context_match`` returns -1

This specifically handles recent changes to visualstudio.com that responds with:

    HTTP/1.1 401 Unauthorized
    WWW-Authenticate: Bearer authorization_uri=https://login.microsoftonline.com/a4c0aef7-6fdd-4c0e-b678-0dd23c73d08a
    WWW-Authenticate: Basic realm=\"https://tfsprodweu2.app.visualstudio.com/\"
    WWW-Authenticate: TFS-Federated

when asking for authentication and the http transport didn't know what to do with ``WWW-Authenticate: Bearer`` returning ``error=-1`` without any further explanation.

I presume the intersection of libgit2 and visualstudio.com users are mostly using the winhttp transport but this change improves behaviour when using the plain http transport.

### DIFF
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -142,7 +142,7 @@ static int auth_context_match(
 	}
 
 	if (!scheme)
-		return -1;
+		return 0;
 
 	/* See if authentication has already started for this scheme */
 	git_vector_foreach(&t->auth_contexts, i, c) {
@@ -187,6 +187,9 @@ static int apply_credentials(git_buf *buf, http_subtransport *t)
 	/* Get or create a context for the best scheme for this cred type */
 	if (auth_context_match(&context, t, credtype_match, &cred->credtype) < 0)
 		return -1;
+
+	if (!context)
+		return 0;
 
 	return context->next_token(buf, context, cred);
 }


src/transports/http.c
```diff
@@ -246,7 +246,7 @@ static int parse_authenticate_response(
 
 	git_vector_foreach(www_authenticate, i, challenge) {
 		if (auth_context_match(&context, t, challenge_match, challenge) < 0)
-			return -1;
+			continue;
```
ethomson: 🤔

One problem here is that it seems that we have overloaded the return value of `auth_context_match`, and it's returning `-1` when there was no scheme match.  I think that might be the root cause of the problems here.

I think that we should change `auth_context_match` to return `0` (and leave `context` unset) when there's no match, instead of [returning `-1`](https://github.com/libgit2/libgit2/blob/453ed9e41eaa321321701409dd908737ff0b28bf/src/transports/http.c#L145).  I think that this bit can then remain unchanged - on `auth_context_match` returning `-1` on error, it will propagate this error, and returning `0` without setting `context`, it will continue.

palmin: Yes, this might be a better idea.

The only other usage is by ``apply_credentials`` and I'm not sure if this needs to do something more to handle the situation where return value is 0 but the output context is NULL?

 https://github.com/libgit2/libgit2/blob/a8d447f68076d1520f69649bb52629941be7031f/src/transports/http.c#L169-L192

palmin:
Pushed a new commit such that ``auth_context_match``returns 0 for unknown schemes and such that ``apply_credentials`` is prepared for that.

palmin:
Will clean this up by squashing the commits into one and making sure whitespace matches the rest of the file.

ethomson:
Thanks for this!

Issue #4838, opened 2018-10-05
ci: use Ninja instead of Makefiles


pks-t:
We should probably switch our Linux and macOS builds over to use Ninja instead of Makefiles to save a few dozen seconds on build time. I'm not too sure about the status on Trusty, but at least CMake 2.8.12 already supports an experimental version of Ninja.

Just putting this up here to not forget about it -- if you're interested, please feel free to pick it up.

PR #4841, opened 2018-10-08
Ninja build

tiennou:
A quick attempt at #4838. This also explicitly selects OpenSSL as a backend, just to make sure it doesn't auto-disable it.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -16,6 +16,7 @@ jobs:
       imageName: 'libgit2/trusty-openssl:latest'
       environmentVariables: |
        CC=gcc
+       CMAKE_OPTIONS=-DUSE_HTTPS=OpenSSL
        LEAK_CHECK=valgrind
 
 - job: linux_trusty_gcc_mbedtls
@@ -41,6 +42,7 @@ jobs:
       imageName: 'libgit2/trusty-openssl:latest'
       environmentVariables: |
        CC=clang
+       CMAKE_OPTIONS=-DUSE_HTTPS=OpenSSL
        LEAK_CHECK=valgrind
 
 - job: linux_trusty_clang_mbedtls
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -69,6 +69,7 @@ jobs:
         TMPDIR: $(Agent.TempDirectory)
         PKG_CONFIG_PATH: /usr/local/opt/openssl/lib/pkgconfig
         LEAK_CHECK: leaks
+        CMAKE_OPTIONS: -G Ninja
 
 - job: windows_vs_amd64
   displayName: 'Windows (Visual Studio; amd64)'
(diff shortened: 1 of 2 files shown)


tiennou:
As suspected, the CMake version on 16.04 doesn't support Ninja, as per its [confused error](https://dev.azure.com/libgit2/libgit2/_build/results?buildId=735&view=logs). Note that the only green Ubuntu build is because I missed switching it to Ninja in the first place.

It looks like it shaves almost a minute from the macOS build step, which is impressive, but ends up being slower overall (which looks to be caused by "generic network latency difference" between compared builds because of "differing datacenter usage" or something). It looks like grabbing Ninja from Homebrew also adds some delay (±15s when comparing "now" baselines).

Example list of builds :
- failed Ninja : https://dev.azure.com/libgit2/libgit2/_build/results?buildId=735&view=logs
- current Ninja : https://dev.azure.com/libgit2/libgit2/_build/results?buildId=736&view=logs
- current baseline (this morning) : https://dev.azure.com/libgit2/libgit2/_build/results?buildId=734&view=logs
- current baseline (now, failed) : https://dev.azure.com/libgit2/libgit2/_build/results?buildId=737&view=logs

(Sorry for the notification noise, I was looking for some candidates PR for baselining)

pks-t:
Cool, thanks for giving this a go! So from the results it looks like we might want to enable it for macOS, right? If it saves about a minute for build but requires 15 seconds more for homebrew that should be a net win of about 45 seconds

azure-pipelines.yml
```diff
@@ -16,6 +16,7 @@ jobs:
       imageName: 'libgit2/trusty-openssl:latest'
       environmentVariables: |
        CC=gcc
+       CMAKE_OPTIONS=-DUSE_HTTPS=OpenSSL -DSHA1_BACKEND=OpenSSL
```
ethomson: Please don't set `SHA1_BACKEND`.  We should encourage people to always use SHA1DC.

In fact, recent conversations on the git mailing list suggest that we should be talking about Git as using SHA1DC instead of using SHA1.  I haven't disabled the ability to choose a SHA1 backend anyway, because some consumers have various requirements about the cryptosystems that they use and I don't want to break them without better knowing their needs here.

tiennou: True, I didn't take this into account. I was just trying to make the testsuite fully specified, and SHA1DC is our default. Or should I drop that commit from the patch set ?

ethomson:
I'm fine with the `-DUSE_HTTPS=OpenSSL` part, I think it makes sense to be explicit there.

azure-pipelines.yml
ethomson: I'm fine with the -DUSE_HTTPS=OpenSSL part, I think it makes sense to be explicit there.

tiennou:
Fixed and rebased !

PR #4842, opened 2018-10-09
config: Port config_file_fuzzer to the new in-memory backend.

nelhage:
In quick testing, I see this hitting up to about 10k exec/s starting with an empty corpus on one core of my laptop, versus about 1500 for the old one.

### DIFF
--- a/fuzzers/config_file_fuzzer.c
+++ b/fuzzers/config_file_fuzzer.c
@@ -8,6 +8,7 @@
  */
 
 #include <git2.h>
+#include "config_backend.h"
 
 #include <stdlib.h>
 #include <stdio.h>
@@ -25,9 +26,6 @@ int foreach_cb(const git_config_entry *entry, void *payload)
 	return 0;
 }
 
-static char path[] = "/tmp/git.XXXXXX";
-static int fd = -1;
-
 int LLVMFuzzerInitialize(int *argc, char ***argv)
 {
 	UNUSED(argc);
@@ -35,10 +33,6 @@ int LLVMFuzzerInitialize(int *argc, char ***argv)
 
 	if (git_libgit2_init() < 0)
 		abort();
-	fd = mkstemp(path);
-	if (fd < 0) {
-		abort();
-	}
 
 	return 0;
 }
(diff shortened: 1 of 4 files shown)


fuzzers/config_file_fuzzer.c
```diff
@@ -25,51 +26,43 @@ int foreach_cb(const git_config_entry *entry, void *payload)
...
 	}
 
-	while (total < size) {
-		ssize_t written = write(fd, data, size);
-		if (written < 0 && errno != EINTR)
-			abort();
-		if (written < 0)
-			continue;
-		total += written;
+	err = git_config_backend_from_string(&backend, (const char*)data, size);
```
pks-t: Nit: there's a space missing between `const char` and `*)` and same style comment as above

fuzzers/config_file_fuzzer.c
```diff
@@ -25,51 +26,43 @@ int foreach_cb(const git_config_entry *entry, void *payload)
...
-	size_t total = 0;
 
-	if (ftruncate(fd, 0) !=0 ) {
-		abort();
-	}
-	if (lseek(fd, 0, SEEK_SET) != 0) {
-		abort();
+	err = git_config_new(&cfg);
+	if (err != 0) {
+		goto out;
```
pks-t: Style nit: `if (err = git_config_new(&cfg)) < 0)`

fuzzers/config_file_fuzzer.c
```diff
@@ -25,51 +26,43 @@ int foreach_cb(const git_config_entry *entry, void *payload)
...
-
-	err = git_config_open_ondisk(&cfg, path);
-	if (err == 0) {
-		git_config_foreach(cfg, foreach_cb, NULL);
-		git_config_free(cfg);
+	err = git_config_add_backend(cfg, backend, 0, NULL, 0);
+	if (err != 0) {
+		goto out;
 	}
+	// Now owned by the config
```
pks-t: Oops. Seems like we aren't yet enforcing C90 for our fuzzing targets. I'm creating a PR for that. This should then be `/* Now owned by the config */`.

src/config_mem.c
```diff
@@ -198,7 +198,7 @@ int git_config_backend_from_string(git_config_backend **out, const char *cfg)
 		return -1;
 	}
 
-	if (git_buf_sets(&backend->cfg, cfg) < 0) {
+	if (git_buf_put(&backend->cfg, cfg, len) < 0) {
```
pks-t: This shouldn't be a put but a `git_buf_set` instead. Put would append the string, while set will replace the current string.

src/config_backend.h
```diff
@@ -30,8 +30,9 @@ extern int git_config_backend_from_file(git_config_backend **out, const char *pa
  *
  * @param out the new backend
  * @param cfg the configuration that is to be parsed
+ * @param len the length of the string pointed to by `cfg`
  */
-extern int git_config_backend_from_string(git_config_backend **out, const char *cfg);
+extern int git_config_backend_from_string(git_config_backend **out, const char *cfg, size_t len);
```
pks-t: Makes sense, I guess

fuzzers/config_file_fuzzer.c
pks-t: Weird, we are enforcing C90. Would've expected it to at least throw a warning...

pks-t:
Thank you!

PR #4843, opened 2018-10-09
Documentation update

tiennou:
This is a rebuild of the project's documentation using my tweaked docurium version. It adds the "missing" v0.26.4..7 and v0.27.1..5. Note that a compilation failure caused v0.27.0 to drop from the list (which I just noticed while checking the list of added versions).

I don't know if that's helpful, but if it's fine with you, I'll happily take care of keeping it up to date from now on !


pks-t:
Ah, so that's how people drive up their lines of code submitted to the project!

But seriously, I highly appreciate any effort that's put into documentation. Keeping docs consistent and up to date is probably one of the most hated tasks, so I'm more than happy if you wish to take care of it from now on. Thanks a lot!

Ccing @carlosmn, as he's more knowledgeable with regards to GH pages and our documentation.

ethomson:
I'm 👍 on this, but I'd also like to have @carlosmn's input.

ethomson:
Instead of maintaining this manually, I'm curious if you think that we could just automatically produce docs as part of a release pipeline?

ethomson:
Any chance that you can base this off the most recent commit in libgit2's `gh-pages` branch so that it's an easy merge?  Or was the rewriting intentional?

tiennou:
It might be possible to handle it as part of a release, tough right now it only does full rebuilds (and it still choked on 0.27).

I've rebuilt against the (moved) gh-pages.

PR #4846, opened 2018-10-12
Bugix release v0.27.6

pks-t:
This cherry-picks all pull requests currently labelled "backport-0.27.6". Most notable is probably the change to our CI infrastructure. Please have a look and let me know if any commits should be dropped or if I missed some importants PRs that should be part of this.


ethomson:
I very intentionally _didn't_ include the XML changes to clar.  Changing the way our test runner works in a rather fundamental way like that seemed very risky for a point release, that's why I had done all this work manually.

pks-t:
I thought it might be something like that, but to be honest I don't quite get the risk,. First, the new XML stuff is optional and only in place with "-r". Sure, it might be the new code actually broke something even without that switch. Second, it's only our tests and not some kind of API. But even if people have scripted against it e.g. in packaging instructions, it would still work just fine as the new XML mode is optional after all. And without the XML stuff, the CI is only half-functioning (well, at least not optimally functioning) as it cannot deduce any information about the tests.

ethomson:
Sure - I don't think it's a big risk, but we changed the way we buffer (or don't) test results, even when you're not using the XML output.  But at this point we've been using it on the master branch for a while now, so it's probably a reasonable enough risk.

pks-t:
Fair enough. I'm fine with taking this risk.

That being said, I'd also be fine with reverting to the old state of this PR using #4798 if you wish, especially as you're the one who has done most of the migration to VSTS.

ethomson:
🤷‍♂️

ethomson:
I think we should probably just merge this.  The important thing for me is getting the new CI scripts merged so that we can set up all the platforms on our point releases.  👍

PR #4848, opened 2018-10-13
Follow 308 redirects (as used by GitLab) 

### LINKED ISSUE #4675: 308 Redirects are not followed
### Reproduction steps

Try to clone a GitLab repository like https://gitlab.com/esodan/gresg/ (including the final / and without .git) - I used gnome-builder to do this

### Expected behaviour

The repository is cloned after any necessary redirects

### Actual behaviour

Clone fails with error quoting status 308

That used in `3.29.2-18-gb86e3ec flatpak-nightly` (not sure what that links to) but should affect all versions

Fedora 28/flatpak

A 308 is defined as:
> **308 Permanent Redirect (RFC 7538)** The request and all future requests should be repeated using another URI. 307 and 308 parallel the behaviors of 302 and 301, but do not allow the HTTP method to change. So, for example, submitting a form to a permanently redirected resource may continue smoothly.

So basically it should be followed in the same way as 301, 302, 303 & 307

https://github.com/libgit2/libgit2/blob/408b16c1b06526f509fdb8997818d648d4d1ea1e/src/transports/http.c#L395-L398

[Originally reported @ GNOME Builder](https://gitlab.gnome.org/GNOME/gnome-builder/issues/521)

ZanderBrown:
Fixes #4675


carlosmn:
Thanks, I've merged the commit manually as you opened the PR from your master branch so your other changes have also made it into here.

ZanderBrown:
"oops"

ethomson:
Thanks @ZanderBrown for diving in!  Very happy to see you tackling this.  😀

PR #4851, opened 2018-10-18
strtol removal

pks-t:
This PR removes all usages of `git__strtol32` and `git__strtol64`. While these functions are fine to use when the string is known to be sanitized, many functions were using them with untrusted input that may not have been NUL-terminated. To avoid future misuse, let's get rid of them altogether in favor of `git__strntol32` and `git__strntol64`.

### DIFF
--- a/src/commit_list.c
+++ b/src/commit_list.c
@@ -171,7 +171,9 @@ static int commit_quick_parse(
 			buffer--;
 	}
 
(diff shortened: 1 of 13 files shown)


pks-t:
I've added another commit to fix undefined behavior due to signed
integer overflow in `git__strntol64`.

ethomson:
This looks good to me.  👍

PR #4854, opened 2018-10-21
 buf::oom tests: use custom allocator for oom failures

ethomson:
Create a custom allocator for the `buf::oom` tests that will fail with out-of-memory errors in predictable ways.  We were previously trying to guess the way that various allocators on various platforms would fail in a way such that `malloc`/`realloc` would return `NULL` (instead of aborting the application, or appearing suspicious to various instrumentation or static code analysis tools like valgrind.)

Introduce a fake `malloc` and `realloc` that will return `NULL` on allocations requesting more than 100 bytes.  Otherwise, we proxy to the default allocator.  (It's important to use the _default_ allocator, not just call `malloc`, since the default allocator on Windows CI builds may be the debugging C runtime allocators which would not be compatible with a standard `malloc`.)

### DIFF
--- a/include/git2/common.h
+++ b/include/git2/common.h
@@ -364,7 +364,8 @@ typedef enum {
  *
  *		> Set the memory allocator to a different memory allocator. This
  *		> allocator will then be used to make all memory allocations for
- *		> libgit2 operations.
+ *		> libgit2 operations.  If the given `allocator` is NULL, then the
+ *		> system default will be restored.
  *
  *	 opts(GIT_OPT_ENABLE_UNSAVED_INDEX_SAFETY, int enabled)
  *
--- a/tests/buf/oom.c
+++ b/tests/buf/oom.c
@@ -1,57 +1,59 @@
 #include "clar_libgit2.h"
 #include "buffer.h"
 
-/*
- * We want to use some ridiculous size that `malloc` will fail with
- * but that does not otherwise interfere with testing.  On Linux, choose
- * a number that is large enough to fail immediately but small enough
- * that valgrind doesn't believe it to erroneously be a negative number.
- * On macOS, choose a number that is large enough to fail immediately
- * without having libc print warnings to stderr.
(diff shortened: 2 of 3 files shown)


include/git2/common.h
```diff
@@ -364,7 +364,8 @@ typedef enum {
  *
  *		> Set the memory allocator to a different memory allocator. This
  *		> allocator will then be used to make all memory allocations for
- *		> libgit2 operations.
+ *		> libgit2 operations.  If the given `allocator` is NULL, then the
+ *		> system default will be restored.
```
pks-t: Makes sense to me

tests/buf/oom.c
```diff
@@ -1,57 +1,59 @@
...
+{
+	/* Reject any allocation of more than 100 bytes */
+	return (n > 100) ? NULL : std_alloc.gmalloc(n, file, line);
+}
+
+static void *oom_realloc(void *p, size_t n, const char *file, int line)
+{
+	/* Reject any allocation of more than 100 bytes */
+	return (n > 100) ? NULL : std_alloc.grealloc(p, n, file, line);
+}
```
pks-t: This is really cool. We might think about extending this OOM allocator to a more general helper that fails on either allocating more than `n` bytes in one allocation or if all allocations in total surpass `m` bytes. So something like `git_oomalloc_init_allocator(git_allocator *a, size_t max_allocation, size_t max_total)`. With such a utility we could easily test our code in various OOM scenarios.

One issue though with the current API is that custom allocators cannot store private data anywhere. Neither is there a field in the `git_allocator` structure that allows one to store a payload nor can users declare a "parent" structure that contains the allocator as its first field, as we do not store the global allocator as a pointer.

From these options, I'd definitely prefer having a payload pointer in the allocator structure. Making `git__allocator` a pointer creates another indirection to the function pointers that may decrease performance in allocation-heavy code

Issue #4815, opened 2018-09-19
The GIT_CONFIG environment variable is not honored


ararslan:
The [`GIT_CONFIG`](https://git-scm.com/docs/git-config#_environment) environment variable tells Git where to look for a configuration file to use. If left unspecified, the default is used. It seems this variable is ignored, be it intentionally or otherwise, by libgit2.

### Reproduction steps

Create a `.gitconfig` file in some directory, and specify user information. For example,
```ini
[user]
    name = Definitely Not Alex
    email = something@else.gov
```
Now set an environment variable `GIT_CONFIG` containing a path to this file. Compile and run this C code, making sure you're inside a Git repo:
```c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <git2.h>

int main(int argc, char** argv) {
    printf("GIT_CONFIG: %s\n", getenv("GIT_CONFIG"));

    git_libgit2_init();

    git_repository* repo = NULL;
    char wd[1024];
    getcwd(wd, sizeof(wd));
    int error = git_repository_open(&repo, wd);

    git_config* cfg = NULL;
    error = git_repository_config(&cfg, repo);

    const git_config_entry* entry = NULL;
    error = git_config_get_entry(&entry, cfg, "user.name");
    printf("libgit2 config username: %s\n", entry->value);

    git_config_entry_free(entry);
    git_config_free(cfg);
    git_repository_free(repo);
    git_libgit2_shutdown();
    return 0;
}
```

### Expected behavior

The configuration as specified in the given file should be used.
```
$ GIT_CONFIG=/Users/alex/testing/.gitconfig ./test
GIT_CONFIG: /Users/alex/testing/.gitconfig
libgit2 config username: Definitely Not Alex
```

### Actual behavior

The default global configuration is used.
```
$ GIT_CONFIG=/Users/alex/testing/.gitconfig ./test
GIT_CONFIG: /Users/alex/testing/.gitconfig
libgit2 config username: Alex Arslan
```

### Version of libgit2 (release number or SHA1)

0.27.4

### Operating system(s) tested

macOS High Sierra

neithernut:
See https://github.com/libgit2/libgit2/issues/2910#issuecomment-385339402: libgit2 (and any other sane library) will not change its behavior based on environment variables (except e.g. `LANG`). It is the responsibility of the application's author to perform/choose a configuration sane for the use-case.

[...]

PR #4856, opened 2018-10-21
WIP: ENV-friendly libgit2

tiennou:
This is a work-in-progress implementation of the design I proposed in #4815. For consistency, here's a summary :

The current `git_repository_open_ext` lacks the flexibility we need to protect users from changes to the API w.r.t environment variable we might start supporting. So, this adds `git_repository_open_with_opts`, which supersedes it, and adds a user-configurable whitelist of what we're allowed to customize using the environment. This also modifies how we load values from the environment, so we can apply any of those allowed envvars while the repository is opened (mostly to support `GIT_CONFIG`, as the previous implementation would have already loaded and used the config before we could override it). Opening bare repositories while using the environment is now available.

I'm labeling it as WIP because I didn't write tests for the new behavior yet, as I have a few points worth discussing:

- as per the last commit in the PR, I propose removing the `git_strarray ceiling_dirs` in favor of a `git_strarray`. Since its value depends on the platform-specific path separator, I feel like we're doing a disservice to users by not handling that more transparently. Since `git_repository_open_with_opts` is brand new, now's the time to do it.

- I've discovered that `GIT_REPOSITORY_OPEN_FROM_ENV`'s documentation said this :

> In the future, this flag will also cause `git_repository_open_ext` to
respect $GIT_WORK_TREE and $GIT_COMMON_DIR; currently,
`git_repository_open_ext` with this flag will error out if either
$GIT_WORK_TREE or $GIT_COMMON_DIR is set.

This is now supported, which means I've changed how `git_repository_from_ext` behaves for unsuspecting users, which was the whole point of the implementation (😁). I have no strong opinion on the matter, we just need to choose whether those envvars are used or not w.r.t to backward-compat.

- I tried to minimize changes to the "find_repo" function, but I haven't checked for work that might be done for nothing, in case the environment specifies everything. IMHO it would not be worthwhile.

- Documentation-checking. From the top of my head, the semantics of what should happen when `ceiling_dirs` is specified as both a parameter and its envvar form. I'm leaning toward using the more explicit value (ie. listening to the code vs. the environment) in that case.


src/repository.c
```diff
@@ -429,6 +429,73 @@ static int read_gitfile(git_buf *path_out, const char *file_path)
...
+		goto cleanup;
+	else {
+		const char *end;
+		char *alt, *sep;
+		error = git_repository_odb(&odb, repo);
+		if (error < 0)
+			goto cleanup;
+
+		end = git_buf_cstr(&alts_buf) + git_buf_len(&alts_buf);
+		for (sep = alt = alts_buf.ptr; sep != end; alt = sep+1) {
```
pks-t: Spaces missing around `+`

src/repository.c
```diff
@@ -598,94 +665,90 @@ int git_repository_open_bare(
 }
 
 static int _git_repository_open_ext_from_env(
-	git_repository **out,
+	struct repo_env *env,
 	const char *start_path,
-	const git_repository_open_options *opts)
+	git_repository_open_options *opts)
```
pks-t: So instead of opening the repository, we now just try to find all necessary paths and return that instead? Shouldn't the function be renamed in that case?

src/repository.c
```diff
@@ -598,94 +665,90 @@ int git_repository_open_bare(
...
+			error = 0;
+			git_buf_sets(&env->gitdir_path, ".");
 		} else if (error < 0)
 			goto error;
 		else {
-			start_path = git_buf_cstr(&dir_buf);
-			flags |= GIT_REPOSITORY_OPEN_NO_SEARCH;
-			flags |= GIT_REPOSITORY_OPEN_NO_DOTGIT;
+			opts->flags |= GIT_REPOSITORY_OPEN_NO_SEARCH;
+			opts->flags |= GIT_REPOSITORY_OPEN_NO_DOTGIT;
```
pks-t: Honestly, it's a bit weird to modify the flags passed in by callers

src/repository.c
```diff
@@ -867,11 +857,19 @@ int git_repository_open_with_opts(
...
+		repo->commondir = git_buf_detach(&env.commondir_path);
+
+	} else if (commondir.size) {
 		repo->commondir = git_buf_detach(&commondir);
 		GITERR_CHECK_ALLOC(repo->commondir);
 	}
 
+	if (use_env && git_buf_len(&env.workdir_path)) {
+		git_buf_swap(&workdir, &env.workdir_path);
+	}
```
pks-t: Useless braces

include/git2/repository.h
```diff
@@ -193,6 +193,9 @@ typedef struct {
 	/**
 	 * A GIT_PATH_LIST_SEPARATOR delimited list of path prefixes at which
 	 * the search for a containing repository should terminate.
+	 *
+	 * FIXME: this depends on the value of GIT_PATH_LIST_SEPARATOR, which is
+	 * platform-specific. Make that a git_strarray ?
```
pks-t: I think it's perfectly fine to fix this API while we're at it. The old function could simply split the string it got into the string array

src/repository.c
```diff
@@ -618,7 +619,7 @@ static int _git_repository_open_ext_from_env(
 	int error;
 
 	if (!start_path) {
-		error = git__getenv(&dir_buf, "GIT_DIR");
+		error = git__getenv_wl(&dir_buf, "GIT_DIR", &opts->allowed_env);
```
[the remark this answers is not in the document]
tiennou: `allowed_env` is initialized by either the backward-compatibility code in `git_repository_open_ext`, or explicitly by the user via the new options struct, which defaults to `NULL`. So unless I'm missing something this *should* behave as expected.

src/repository.c
tiennou: Fixed, I've renamed two offending functions to clarify what's being setup from what (the new `struct repo_env`, and this one).

include/git2/repository.h
```diff
@@ -193,6 +199,14 @@ typedef struct {
 	 * the search for a containing repository should terminate.
 	 */
 	const char *ceiling_dirs;
+
+	/**
+	 * When GIT_REPOSITORY_OPEN_FROM_ENV is set, this can be used to
+	 * whitelist the environment variables we're allowed to access.
+	 *
+	 * An empty list means that we can use any of them.
```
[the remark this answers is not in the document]
tiennou: Sounds like I've confused myself too : a `NULL` whitelist means "any", an empty one means "none". I fail to see the use case for "empty" so I'm not sure if it makes sense. I'll fix & rephrase though.

src/repository.c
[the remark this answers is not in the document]
tiennou: Disregard that, I mistook pointers for non-pointers 🙄.

src/repository.c
```diff
@@ -485,19 +535,22 @@ static int find_repo(
...
 			if (S_ISDIR(st.st_mode)) {
 				if (valid_repository_path(&path, &common_link)) {
 					git_path_to_dir(&path);
-					git_buf_set(gitdir_path, path.ptr, path.size);
+					GIT_ERROR_CHECK_ALLOC_BUF(&path);
+
+					git_buf_set(&env->gitdir_path, path.ptr, path.size);
+					GIT_ERROR_CHECK_ALLOC_BUF(&env->gitdir_path);
+
+					git_buf_attach(&env->gitlink_path,
```
tiennou: This reads a (potentially non-existing) `$GIT_DIR/gitdir` file in `gitlink`.

src/repository.c
```diff
@@ -529,31 +582,37 @@ static int find_repo(
...
+	if (!(env->flags & GIT_REPOSITORY_OPEN_BARE)) {
+		if (!git_buf_len(&env->gitdir_path))
+			git_buf_clear(&env->workdir_path);
 		else {
-			git_path_dirname_r(workdir_path, path.ptr);
-			git_path_to_dir(workdir_path);
+			git_path_dirname_r(&env->workdir_path, path.ptr);
+			git_path_to_dir(&env->workdir_path);
+		}
+		if (git_buf_oom(&env->workdir_path)) {
```
tiennou: Here, we set `workdir_path` as whatever `dirname(path)` is.

src/repository.c
```diff
@@ -790,48 +843,87 @@ static int repo_is_worktree(unsigned *out, const git_repository *repo)
...
 	}
-	if (commondir.size) {
-		repo->commondir = git_buf_detach(&commondir);
+
+	if (git_buf_len(&env.commondir_path) > 0) {
+		repo->commondir = git_buf_detach(&env.commondir_path);
 		GIT_ERROR_CHECK_ALLOC(repo->commondir);
 	}
 
+//	if (git_buf_len(&env.workdir_path) > 0) {
```
tiennou: I've added this, since it seemed to be provided/detected. This causes a crash further down, hence the commenting.

src/repository.c
```diff
@@ -790,48 +843,87 @@ static int repo_is_worktree(unsigned *out, const git_repository *repo)
...
+//		repo->workdir = git_buf_detach(&env.workdir_path);
+//		GIT_ERROR_CHECK_ALLOC(repo->workdir)
+//	}
+
+	if (git_buf_len(&env.configfile_path) > 0) {
+		repo->configfile = git_buf_detach(&env.configfile_path);
+		GIT_ERROR_CHECK_ALLOC(repo->configfile);
+	}
+
 	if ((error = repo_is_worktree(&is_worktree, repo)) < 0)
```
tiennou: [Function](https://github.com/libgit2/libgit2/blob/master/src/repository.c#L770).

This could have easily been part of `repo_env_find`, since it only look at the filesystem.

src/repository.c
```diff
@@ -846,32 +938,66 @@ int git_repository_open_ext(
...
 
-	if ((flags & GIT_REPOSITORY_OPEN_BARE) != 0)
+	if ((opts.flags & GIT_REPOSITORY_OPEN_BARE) != 0)
 		repo->is_bare = 1;
 	else {
 
 		if (config &&
 		    ((error = load_config_data(repo, config)) < 0 ||
-		     (error = load_workdir(repo, config, &workdir)) < 0))
+		     (error = load_workdir(repo, config, &env.workdir_path)) < 0))
```
tiennou: We move on to actually setup the working directory, at https://github.com/libgit2/libgit2/pull/4856/files#diff-7cbc97c802e21b199f176b493cec8e9cR300, and we pass the workdir identified by the "find" step. This is the part where the crash I'm getting comes from, and I'm starting to get completely lost (and I can't inline-review that code anyway, so there goes) :

- `core.worktree` is loaded from the config, as it can now be made use of.
- if `repo->is_worktree` is true (it only performed filesystem-checks), we read `repo->gitdir` + "gitdir" (already loaded while finding in `env->gitlink`), `dirname()` that, and use the result as our workdir.
- if `core.worktree` is set, we check its value and use that as our workdir.
- if the `parent_path` is passed, we check that it's a directory and use that.
- finally, we `dirname(repo->gitdir)` and use that.

It seems that `core.worktree` gets a say iif the repository structure is slightly "wrong" — a `$GIT_DIR/gitdir` file that doesn't point to a valid repo. The fact that it is passing as a parameter the very thing this code should be setting up doesn't help 😉.

tiennou (commented):
Rebased, as I've wanted to try to polish it a little, and hit a snag. This is still WIP, but the current code really needs a refactor, and I don't think I can handle it without re-reading a whole lot of git manpages, so…

This should be green, but as I had to comment "innocuous looking" code to fix it, I had to look more closely at what was going on. I've left review comments about those issues, in case someone else has any insight. Note that there are a bunch of missing oom checks in there, which I didn't attempt to fix yet.

ethomson:
Tidying up some old PRs. We support these environment variables now.

Issue #4857, opened 2018-10-22
Segmentation fault or other errors when reading splitIndex


glasserc:
Does libgit2 support repositories with the `core.splitIndex` feature turned on?

### Reproduction steps

These steps are based on #4234.

```
$ git --version
git version 2.19.1
$ git init /tmp/test-splitindex
$ cd /tmp/test-splitindex
$ git config --add core.splitIndex true
$ git update-index --index-version=4 --split-index
$ touch file.txt
$ git add file.txt
$ git commit -m "first commit"
# Verify that .git/ has sharedindex.* files.
$ ls .git/sharedindex.*
sharedindex.4dc495dafdb005ef4f40b20e171474e626bca290
sharedindex.508851a7f0dfa8691e9f69c7f055865389012491
```

Try to open the index with libgit2:

```C
#include <git2.h>
#include <stdio.h>

int main()
{
  git_repository *repo;
  git_index *index;
  int err;

  err = git_repository_open(&repo, "/tmp/test-splitindex");
  if (err < 0) {
    printf("Error opening repo\n");
    return 1;
  }

  err = git_repository_index(&index, repo);
  if (err < 0) {
    printf("%s\n", giterr_last()->message);
    return 1;
  }

  return 0;
}
```

```
$ gcc main.c -lgit2
$ ./a.out
```

### Expected behavior

No output (i.e. success)

### Actual behavior

`Segmentation fault`

### Version of libgit2 (release number or SHA1)

0.26.6

### Operating system(s) tested

NixOS unstable

### Additional notes

I'm actually trying to debug an issue where [cargo](https://github.com/rust-lang/cargo) refuses to build code in a large repo (where I have turned on `splitIndex`). There, I get cryptic errors like "error: invalid data in index - extension is truncated; class=Index (10)". It may be that the two issues are related, in which case you may not get a seg fault but some other error.

tiennou:
A cursory look at the index code seems to confirm that split indexes are indeed not supported ([here](https://github.com/libgit2/libgit2/blob/master/src/index.c#L2612)'s where the error comes from, and [here](https://github.com/libgit2/libgit2/blob/master/src/index.c#L2521)'s the index extension handling).

As to the segfault, I can't reproduce — I've checked against an empty repo with split index enabled, and I'm getting the aforementioned error message.

glasserc:
OK, that's still pretty helpful, thanks!

grantfar:
I will take this

[...]

PR #4858, opened 2018-10-22
index: preserve extension parsing errors

tiennou:
Previously, we would clobber any extension-specific error message with an "extension is truncated" message. This makes `read_extension` correctly preserve those errors, takes responsibility for truncation errors, and adds a new message with the actual extension signature for unsupported mandatory extensions.

Inspired by #4857.


ethomson:
The index changes look good to me - (the tests are failing on Linux though).

tiennou:
I'm not sure why though. It seems the "splitindex" test repo I added isn't there ? If that makes any sense ?

Wait, the macOS build should be red as well, the test is failing there (at 16957).

ethomson:
> Wait, the macOS build should be red as well, the test is failing there (at 16957).

Ugh, it certainly is.  I think the hoops that I had to jump through to get leaks to play nicely (it wants to background itself but keep some file descriptors open 😡) must not have worked right.  Sigh.

It's already very brittle (if you can't convince it to close that last file descriptor then it will just never exit and hang the builds), so I think I may have to disable leaks until I can make it more stable.

tiennou:
I'm really confused by CI here, I can run that new testcase using the new test resource locally and I get a successful run, but CI fails because the resource is missing. Also, I suspect the Windows also has a problem with considering the build green while it failed…

tiennou:
Rebased, was hoping to get CI to play nice but it still doesn't want to play along. I feel like I'm missing something, as I can't see anything wrong with the testcase setup…

ethomson:
> Rebased, was hoping to get CI to play nice but it still doesn't want to play along. I feel like I'm missing something, as I can't see anything wrong with the testcase setup…

tiennou:
OMG I've been wrangling with the repository open code and it didn't register that there were missing directories… Thanks for the pointer @ethomson, I've now rebased it with some `.gitkeep` files, and will take a note that a nicer validity message would be ❤️.

ethomson:
> I don't think I have the permissions to do that (and I'm kinda evading doing that, because I'm a rebase-junkie and I like to keep it separated). I also was under the impression that it's only dependent on whether "Allow edit from maintainers" being checked ?

My workflow in `ethomson/foo` branches is to rebase also, so I wouldn't feel bad about that.  I'm not in front of my normal computer right now but we should get you these permissions.  Whether you choose to use them will remain up to you.  (It would make my life the tiniest bit easier, since I wouldn't have to set up your fork as a remote to fetch your PRs and build them locally, but that's not yet so common that I'm going to encourage you strongly to change your behavior.)

src/index.c
```diff
@@ -2645,11 +2650,8 @@ static int parse_index(git_index *index, const char *buffer, size_t buffer_size)
 	while (buffer_size > INDEX_FOOTER_SIZE) {
 		size_t extension_size;
 
-		extension_size = read_extension(index, buffer, buffer_size);
-
-		/* see if we have read any bytes from the extension */
-		if (extension_size == 0) {
-			error = index_error_invalid("extension is truncated");
+		error = read_extension(&extension_size, index, buffer, buffer_size);
+		if (error < 0) {
```
pks-t: Minor nit: I'd reformat this to our usual style of

```
if ((error = read_extension(&extension_size, index, buffer, buffer_size)) < 0)
    goto done;
```

Please only consider this if you are already rerolling :)

ethomson:
I'm 👍 on this.  @tiennou let me know if you want to make that change that @pks-t pointed out.  If not, we can go ahead and merge this.  If so, I'll hold off.

tiennou:
Rebased, with style-fix applied. Thanks for the review!

PR #4859, opened 2018-10-23
ci builds: drop macOS leak detection

ethomson:
Memory leak detection using the leaks application on macOS is brittle -
IIRC, it tries to background itself, but keeps a file descriptor open.
I previously got this to the point where it would exit cleanly (instead
of refusing to exit, and causing a build timeout after 60 minutes).  But
apparently these mechanations obscure clar's error code.

Disable leak detection on macOS until we can resolve this problem.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -70,7 +70,6 @@ jobs:
       environmentVariables:
         TMPDIR: $(Agent.TempDirectory)
         PKG_CONFIG_PATH: /usr/local/opt/openssl/lib/pkgconfig
-        LEAK_CHECK: leaks
         CMAKE_OPTIONS: -G Ninja
 
 - job: windows_vs_amd64


ethomson:
lol, for once, the macOS build completes first!

Obviously, correctness and reproducibility are the most important things here.  But I do want to make sure that we're doing leak detection on all our builds, so I'll revisit this and get it enabled again.  

@tiennou, do you happen to know if `leaks` is really the best tool here or if there are other options that we can explore?

tiennou:
Yeah, leaks is slightly painful, but as far as I know the only tool that does this, and last I checked `valgrind` doesn't work on macOS (and is painful to setup), and Instruments is GUI, while this is OS-level stuff that's described in `malloc(3)`.

A few notes :
- We need to unset `MallocStackLogging` and `MallocStackLoggingNoCompact` from the environment so processes we spawn (lots of `rm` and `cp` because of the sandbox) don't also start logging their leaks. At least that would make what clar is outputting usable again.

- I don't have much experience with the `-atExit` option, but AFAICT there is no leak report anywere, so I don't think the leaks detection actually does anything (like failing the build).

My current (out of habit) usage is to set a breakpoint on clar's last return statement, and run `leaks $clarPID`, so I can get to the leak traces, but I don't know how easily that can be scripted.

ethomson:
#4860 is an obvious improvement.

PR #4862, opened 2018-10-25
Windows CI: fail build on test failure

ethomson:
PowerShell can _read_ top-level variables in functions, but cannot _update_ top-level variables in functions unless they're explicitly prefixed with `$global`.


ethomson:
As noticed in #4858, the Windows builds are not failing when the tests fail.  😰

PR #4869, opened 2018-10-28
Nightlies: use `latest` docker images

ethomson:
I was originally trying to get the ppc64 builds working, but the emulator seems to not love emulating that platform.  I'm abandoning that but taking the update to use the `latest` images.

### DIFF
--- a/azure-pipelines/nightly.yml
+++ b/azure-pipelines/nightly.yml
@@ -135,7 +135,7 @@ jobs:
   - template: docker.yml
     parameters:
       qemu: 'true'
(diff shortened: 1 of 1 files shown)


ethomson:
I was seeing this before on the ppc64el builds where the gitdaemon tests appear to never finish.  I'll have to investigate this a bit to better understand what's happening.  😰

pks-t:
Huh. Seems like Azure is confused :/

pks-t:
Cool, thanks @ethomson :)

Issue #4207, opened 2017-04-17
We should get rid of CURL and use our onw HTTP CONNECT proxy implementation


carlosmn:
The only use of libcurl that we have is for it to connect to proxies via the `CONNECT` method, which then simply proxies our data stream unaltered to the final location.

It however adds complexity in the build process and a dependency we're barely using. It introduces a different way of talking to remote machines and its own way of representing X.509 certificates which mean we can't convert them into native objects in higher-level languages.

Since we need to have our own HTTP stack anyway, it sounds like a worthwhile trade-off to implement our own `CONNECT` client so we don't depend on curl being available both on the build and target machines.

mtangoo:
I agree. It took me days only to compile curl with libgit2. Today I don't even remember how I did it.  Curl is great but its just a nightmare to compile with libgit2 esp for app that should work in all major platforms

ethomson:
Fixed via #4870

mtangoo:
@ethomson does it mean libgit2 is no longer depending on libcurl?

ethomson:
@mtangoo Yes.

mtangoo:
That's great! One more dependency off..!
Next is OpenSSL (just kidding)

PR #4870, opened 2018-10-29
Add builtin proxy support for the http transport

### LINKED ISSUE #4515: HTTPS proxy support with curl
### Reproduction steps
Configured a local squid instance to run https with a self signed cert.
Passed this https url for the proxy settings and attempting a clone a github repo 
### Expected behavior
Connection should establish through proxy and allow cloning.
### Actual behavior
code is quitting with the exception:
"curl socket is no longer valid"
libgit2 version 0.25.1
centos-release-6-8.el6.centos.12.3.x86_64

I couldn't find anywhere which stated whether https proxies are supported at this point in time or not. Is this simple a limitation of libgit right now?

ethomson:
Provide proxy support for the HTTP transport.

* When we're using a proxy, we now connect directly to the proxy, instead of trying to connect to the remote (git server).  
* For connections to an HTTP remote, we can simply change the request URL that we send to the proxy to be a fully-qualified URL (eg `GET https://github.com/libgit2/libgit2/...`)
* For connections to an HTTPS remote, we need to request that the proxy build a tunnel to the remote, using the `CONNECT` verb.
  * At that point, we can begin speaking HTTPS to the remote over the socket we've built to the proxy.
  * I've introduce `git_tls_stream_wrap` which will create a TLS stream on top of an existing stream (instead of trying to connect a socket).

I've also updated our test proxy to support speaking HTTPS, so I've updated the tests to support that as well and introduced a second proxy test pass: one that speaks HTTP to the proxy and one that speaks HTTPS.

With our own proxy support, this removes the need for us to use libcurl.  It's been removed from the codebase.

Fixed #4515 #4207


ethomson:
Note: I'm seeing periodic 400's on proxied connections.  As best I can tell, this might actually be a problem with the proxy itself (or perhaps this new code's interaction with our particular test proxy), since I only see them when connecting through our test proxy, not with a debugging proxy like Charles.  

I'll have to dig in further but since I think the problem lies with the proxy, I'm going to go ahead and open this for review.

src/transports/http.c
```diff
@@ -418,15 +424,29 @@ static int on_headers_complete(http_parser *parser)
...
+	if (parse_authenticate_response(&t->proxy_authenticate, t,
+	        &allowed_proxy_auth_types) < 0 ||
+	    parse_authenticate_response(&t->www_authenticate, t,
 	        &allowed_www_auth_types) < 0)
 		return t->parse_error = PARSE_ERROR_GENERIC;
 
+	/* Check for a proxy authentication failure. */
+	if (parser->status_code == 407 && get_verb == s->verb)
+		return on_auth_required(&t->proxy_cred,
+		    parser,
```
tiennou: Indent.

src/transports/http.c
```diff
@@ -418,15 +424,29 @@ static int on_headers_complete(http_parser *parser)
...
+		    t->proxy.url,
+		    "proxy",
```
tiennou: A define, maybe ?

src/transports/http.c
```diff
@@ -349,10 +350,19 @@ static int on_header_value(http_parser *parser, const char *str, size_t len)
 	return 0;
 }
 
+GIT_INLINE(void) free_cred(git_cred **cred)
```
tiennou: I'd be tempted to move that as a function in the cred system. Also, I remember wondering as to why this specific struct happens to be open, but it's unrelated.

src/transports/http.c
```diff
@@ -66,17 +66,28 @@ typedef struct {
...
+typedef struct {
+	gitno_connection_data url;
+	git_stream *stream;
+
+	git_cred *cred;
+	git_cred *url_cred;
+
+	git_vector auth_challenges;
+	git_vector auth_contexts;
+} http_server;
```
tiennou: ❤️

src/streams/mbedtls.c
```diff
@@ -405,12 +402,44 @@ int git_mbedtls_stream_new(git_stream **out, const char *host, const char *port)
...
+
+int git_mbedtls_stream_new(
+	git_stream **out,
+	const char *host,
+	const char *port)
+{
+	git_stream *stream;
+	int error;
+
+	assert(out && host);
```
tiennou:
```suggestion
	assert(out && host && port);
```

src/streams/openssl.c
```diff
@@ -737,10 +734,29 @@ int git_openssl_stream_new(git_stream **out, const char *host, const char *port)
...
+{
+	return openssl_stream_wrap(out, in, host, 0);
+}
+
+int git_openssl_stream_new(git_stream **out, const char *host, const char *port)
+{
+	git_stream *stream = NULL;
+	int error;
+
+	assert(out && host);
```
tiennou:
```suggestion
	assert(out && host && port);
```

src/streams/stransport.c
```diff
@@ -226,41 +226,38 @@ static int stransport_close(git_stream *stream)
...
 static void stransport_free(git_stream *stream)
 {
 	stransport_stream *st = (stransport_stream *) stream;
 
-	git_stream_free(st->io);
-	CFRelease(st->ctx);
+    if (st->owned)
+        git_stream_free(st->io);
+
+    CFRelease(st->ctx);
```
tiennou: Nitpick: indent.

src/transports/http.c
```diff
@@ -555,99 +640,335 @@ static int write_chunk(git_stream *io, const char *buffer, size_t len)
...
+		t->proxy_opts.certificate_check = t->owner->proxy.certificate_check;
+		t->proxy_opts.payload = t->owner->proxy.payload;
+		break;
+
+	case GIT_PROXY_SPECIFIED:
+		memcpy(&t->proxy_opts, &t->owner->proxy, sizeof(git_proxy_options));
+		break;
+
+	default:
+		abort();
```
tiennou: Heavy-handed 😉.

src/transports/http.c
```diff
@@ -555,99 +640,335 @@ static int write_chunk(git_stream *io, const char *buffer, size_t len)
...
+		return -1;
+
+	git_buf_puts(buf, "\r\n");
+
+	return git_buf_oom(buf) ? -1 : 0;
+}
+
+static int proxy_headers_complete(http_parser *parser)
+{
+    parser_context *ctx = (parser_context *) parser->data;
```
tiennou: Incorrect indentation in function body.

include/git2/sys/stream.h
```diff
@@ -40,18 +40,47 @@ typedef struct git_stream {
...
  * If a constructor is already set, it will be overwritten. Pass
  * `NULL` in order to deregister the current constructor.
  *
- * @param ctor the constructor to use
+ * @param registration the registration data
  * @return 0 or an error code
  */
-GIT_EXTERN(int) git_stream_register_tls(git_stream_cb ctor);
+GIT_EXTERN(int) git_stream_register_tls(
+    git_stream_registration *registration);
```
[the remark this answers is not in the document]
ethomson: It does.  My thought has been that things in `sys` are more experimental and fragile and have basically zero users.  We discussed this at some point several years ago but I'm not sure that this is really documented or that we even would still agree on that distinction today.

It's not clear to me what we would do here for compat.  We could keep `git_stream_register_tls` as-is and use the `git_stream_cb` as the `init` function, and hard error on using proxies.  I think as a downstream user, I'd rather be broken than have that happen, but that's pretty debatable.

src/streams/openssl.c
ethomson: That's more restrictive than non-TLS sockets, which allow a `NULL` port, since `getaddrinfo` allows a `NULL` service as well.  I think that asserting that `port` be non-`NULL` makes sense, but we should think through the differences and make sure that we understand how it works (or - more likely - doesn't work) with a `NULL` port  today.

src/streams/tls.h
```diff
@@ -20,4 +23,14 @@
  */
 extern int git_tls_stream_new(git_stream **out, const char *host, const char *port);
 
+/**
+ * Create a TLS stream on top of an existing insecure stram, using
+ * the most appropriate backend available for the current platform.
+ *
+ * This allows us to create a CONNECT stream on top of a proxy;
+ * using SecureTransport or OpenSSL stream according to being on
+ * general Unix vs OS X.
```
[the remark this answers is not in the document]
ethomson: Yeah, that's copy-pasta from the `git_tls_stream_new` docs.  I'll clean them both up to make it sound like they're examples (not a list of supported backends), but it's worth noting that this is internal only.  😀

ethomson:
Thanks @tiennou - I'll give a pass on indentation.  I've been using about 4 different editors on about 4 different systems so something put in soft tabs.  😢

tests/core/stream.c
```diff
@@ -17,13 +17,29 @@ static int test_ctor(git_stream **out, const char *host, const char *port)
...
+	return 0;
+}
+
 void test_core_stream__register_tls(void)
 {
 	git_stream *stream;
+	git_stream_registration registration = {0};
 	int error;
 
+	registration.version = 42;
```
pks-t: Honestly, stream registration should fail if we see an unsupported version as it indicates that the application invoking the registration may have an incompatible structure definition

src/transports/http.c
```diff
@@ -302,6 +306,12 @@ static int on_header_ready(http_subtransport *t)
 			GITERR_CHECK_ALLOC(t->content_type);
 		}
 	}
+	if (!strcasecmp("Content-Length", git_buf_cstr(name))) {
```
pks-t: While not strictly necessary, this should probably be an `else if`

src/transports/http.c
```diff
@@ -302,6 +306,12 @@ static int on_header_ready(http_subtransport *t)
...
+		if (!t->content_length) {
+			t->content_length = git__strdup(git_buf_cstr(value));
+			GITERR_CHECK_ALLOC(t->content_length);
+		}
```
pks-t: So in case we already got a content length we just ignore it? Shouldn't it instead create an error?

src/transports/http.c
```diff
@@ -718,10 +735,177 @@ static int stream_connect(
...
+}
+
+static int proxy_headers_complete(http_parser *parser)
+{
+    parser_context *ctx = (parser_context *) parser->data;
+    http_subtransport *t = ctx->t;
+    int proxy_auth_types = 0;
+
+    /* Enforce a reasonable cap on the number of replays */
+    if (++t->replay_count >= GIT_HTTP_REPLAY_MAX) {
```
pks-t: Again, this caps us to 6 replays due to the pre-increment

pks-t:
Thanks for working on this! I really like the direction -- removing dependencies is a nice thing, and I've really been bothered by libcurl in the past due to Ubuntu not backporting fixes to security-related issues. There's also quite a bunch of nice cleanups in here.

src/streams/tls.c
```diff
@@ -5,41 +5,124 @@
...
+	return 0;
+}
 
-int git_stream_register_tls(git_stream_cb ctor)
+int git_stream_register_tls(git_stream_registration *registration)
 {
-	tls_ctor = ctor;
+	assert(!registration || registration->init);
 
+	if (git_rwlock_wrlock(&stream_registration.lock) < 0) {
```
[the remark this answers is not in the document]
ethomson: This is something we support today and its relied on in LibGit2Sharp.  Since LibGit2Sharp itself supports pluggable streams, we have tests that create a stream and invoke it.  Those tests might get invoked at any time, not necessarily first-thing (nor is their lifecycle until the end of the test runs).

include/git2/sys/stream.h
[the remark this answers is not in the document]
ethomson: I've restored the old API for backcompat.  I've actually made the non-TLS stream pluggable as well; this is something that we could benefit from over in LibGit2Sharp land, as we're desperately trying to use the .NET networking functionality if possible.

src/streams/tls.c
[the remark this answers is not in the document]
ethomson: It was; fixed.

tests/core/stream.c
ethomson: Agreed.

src/transports/http.c
```diff
@@ -1144,38 +1129,46 @@ static int http_action(
...
-	http_subtransport *t = (http_subtransport *) subtransport;
 	git_http_auth_context *context;
 	size_t i;
 
+	git_vector_foreach(contexts, i, context) {
+		if (context->free)
+			context->free(context);
+	}
+
+	git_vector_clear(contexts);
```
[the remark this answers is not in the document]
ethomson: Yes, it's possible that we'll repopulate them and I wanted to avoid the additional free/allocate cycles in that case.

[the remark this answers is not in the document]
ethomson: (Particularly with negotiate-type mechanisms, which have multiple round-trips to get authenticated.)

src/transports/http.c
ethomson: Indeed - this was to emulate https://github.com/libgit2/libgit2/blob/master/src/transports/winhttp.c#L932 but I agree that it's annoying.  I'll fix.

src/transports/http.c
```diff
@@ -1012,11 +1041,24 @@ static int http_action(
...
-		return -1;
+	assert(stream);
+
+	/*
+	 * Callers should provide a non-NULL url for the first connection
+	 * to a server.  Subsequent calls should pass a NULL url here to
+	 * reuse the existing one.
+	 */
+	if (url) {
+		assert(!t->gitserver_data.host && !t->gitserver_data.port && !t->gitserver_data.path);
```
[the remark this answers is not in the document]
ethomson: This is now outdated per https://github.com/libgit2/libgit2/pull/4880.

src/transports/http.c
ethomson: Yeah, I thought about that - there's a `cred->free` function, but this is really concerned with the freeing of the `git_cred *` itself which is rather unique to the transport itself.

tiennou: Actually, there's already a `git_cred_free` function there, but I'm not sure if it matches your semantics (or even if you've switched to it now, as the diff is out of date).

ethomson:
Thanks @pks-t and @tiennou for the feedback.  All good comments and I've updated this accordingly.

I tracked down the 400 errors - in fact it was _not_ an error with the proxy, we were removing our internal proxy configuration state by reloading the user's proxy configuration, causing us to send proxy-looking requests (eg `GET https://host/path`) through the CONNECT tunnel instead of regular requests (eg `GET /path`).  Fixed in `66c78cb`.

I've also removed the ability to connect to a proxy over SSL (we still allow users to connect to an SSL _remote_ through a proxy, using CONNECT, but we connect to the proxy itself over HTTP only).  The SSL tunnel over SSL connection is showing warnings in valgrind in Ubuntu 14.04.  It's unclear whether we're calling OpenSSL incorrectly or if it's an OpenSSL error.  But we need not resolve that immediately, we can instead forego support for connecting to proxies over SSL.

This is not a regression - our curl proxy support also had this limitation.  I think this is a _bad_ limitation, obviously, but I'd like to ship this to unblock some LibGit2Sharp scenarios by removing the libcurl dependency.

So I think that this PR is a strict improvement:  it does not reduce any functionality, and it removes the libcurl dependency.  Therefore, I'd like to merge this and come back for the SSL proxy connection support in a separate PR.

ethomson:
Fixes #4126

src/transports/http.c
```diff
@@ -255,16 +259,16 @@ static int gen_request(
...
 {
 	git_http_auth_context *context;
 	char *challenge;
 	size_t i;
 
-	git_vector_foreach(www_authenticate, i, challenge) {
-		if (auth_context_match(&context, t, challenge_match, challenge) < 0)
+	git_vector_foreach(&server->auth_challenges, i, challenge) {
+		if (auth_context_match(&context, server,
+		    challenge_match, challenge) < 0)
```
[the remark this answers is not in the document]
pks-t: Digging through some files, git does exactly what I said: they align function parameters with the opening brace. E.g. "builtin/clone.c":

```
	for (r = local_refs; r; r = r->next) {
		if (!r->peer_ref)
			continue;
		if (ref_transaction_create(t, r->peer_ref->name, &r->old_oid,
					   0, NULL, &err))
			die("%s", err.buf);
	}
...
		update_ref(msg, ref->name, &ref->old_oid, NULL, 0,
			   UPDATE_REFS_DIE_ON_ERR);

```

"builtin/apply.c":

```
	argc = apply_parse_options(argc, argv,
				   &state, &force_apply, &options,
				   apply_usage);
```

I only found instances where they do align with the opening brace, didn't find any contrary ones.

But I very much agree with your argument against that. I also prefer meaningful names, but still prefer to wrap after a reasonable amount of characters. I don't think we've been following git's style too frequently. I might be the only one who actually (sometimes?) does it. So I'd also be fine with just indenting with a tab.

include/git2/sys/stream.h
```diff
@@ -40,19 +40,78 @@ typedef struct git_stream {
...
- * @param ctor the constructor to use
+ * @param registration the registration data
+ * @param tls 1 if the registration is for TLS streams, 0 for regular
+ *        (insecure) sockets
  * @return 0 or an error code
  */
+GIT_EXTERN(int) git_stream_register(
+	git_stream_registration *registration, int tls);
+
+/** @name Deprecated TLS Stream Registration Funtions
```
ethomson: typo

src/transports/http.c
ethomson: Aha - yes - I see what you're saying; that seems reasonable.

src/streams/stransport.h
```diff
@@ -12,5 +12,6 @@
 #include "git2/sys/stream.h"
 
 extern int git_stransport_stream_new(git_stream **out, const char *host, const char *port);
+extern int git_stransport_stream_wrap(git_stream **out, git_stream *in, const char *host);
```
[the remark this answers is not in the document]
ethomson: 👍

src/streams/socket.c
```diff
@@ -208,3 +212,26 @@ int git_socket_stream_new(git_stream **out, const char *host, const char *port)
...
+	git_stream **out,
+	const char *host,
+	const char *port)
+{
+	int (*init)(git_stream **, const char *, const char *) = NULL;
+	git_stream_registration custom = {0};
+
+	assert(out && host && port);
+
+	if (git_stream_registry_lookup(&custom, 0) == 0)
```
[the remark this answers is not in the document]
ethomson: Makes sense, good catch.

src/streams/registry.c
```diff
@@ -0,0 +1,102 @@
...
+	if (target->init) {
+		memcpy(out, target, sizeof(git_stream_registration));
+		error = 0;
+	}
+
+	git_rwlock_rdunlock(&stream_registry.lock);
+	return error;
+}
+
+int git_stream_register(git_stream_registration *registration, int tls)
```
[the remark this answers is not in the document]
ethomson: A lot of the other functions - like `git_filter_register` and `git_merge_driver_register` take the attribute that they're registering (like the filter name or merge driver name) as a param and then the data as a struct for another parameter.  In this particular use case for streams, this lets you re-use the same struct for HTTP and HTTPS streams, which seems like a reasonable practice.  (Obviously you could just change the single `tls` value in the struct without too much effort.)

However, these other examples put the variant first.  eg:

```
GIT_EXTERN(int) git_merge_driver_register(
    const char *name, git_merge_driver *driver);
```

and

```
GIT_EXTERN(int) git_filter_register(
    const char *name, git_filter *filter, int priority);
```

So I have moved the `tls` to the first argument.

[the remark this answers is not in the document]
ethomson: Actually, thinking more about this, being able to register the same registration structure with a single call would be even nicer, so I changed this to an enum.

src/transports/http.c
```diff
@@ -478,8 +479,6 @@ static int on_headers_complete(http_parser *parser)
 		t->location = NULL;
 
 		t->connected = 0;
-		s->redirect_count++;
```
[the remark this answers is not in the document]
ethomson: Yes, I went through and I agree with your analysis.  I think it's unlikely to occur, but definitely possible.  I've added one change to reset the replay count upon connection, which would be useful if - for example - you're able to send a successful anonymous request (and are thus "connected") but a subsequent request requires authentication.  Especially one of the chatty authentication mechanisms like SPNEGO.

pks-t:
Seems sensible in theory. It only gets hard to phrase that as a guide. But considering we ain't got one anyway I think this point is moot

include/git2/sys/stream.h
```diff
@@ -40,19 +40,90 @@ typedef struct git_stream {
...
- * Register a TLS stream constructor for the library to use
+ * The type of stream to register.
+ */
+typedef enum {
+	/** A standard (non-TLS) socket. */
+	GIT_STREAM_STANDARD = 1,
+
+	/** A TLS-encrypted socket. */
+	GIT_STREAM_TLS = 2,
+} git_stream_t;
```
pks-t: Yeah, this makes a lot of sense to me. Allows us to also have additional custom stream types if the need ever arises

src/streams/registry.c
```diff
@@ -0,0 +1,116 @@
...
+	if (git_rwlock_wrlock(&stream_registry.lock) < 0) {
+		giterr_set(GITERR_OS, "failed to lock stream registry");
+		return -1;
+	}
+
+	if ((type & GIT_STREAM_STANDARD) == GIT_STREAM_STANDARD)
+		stream_registration_cpy(&stream_registry.callbacks, registration);
+
+	if ((type & GIT_STREAM_TLS) == GIT_STREAM_TLS)
+		stream_registration_cpy(&stream_registry.tls_callbacks, registration);
```
pks-t: So now you can ever register a stream that is able to handle both TLS and non-TLS connections.

src/streams/registry.h
```diff
@@ -4,14 +4,16 @@
...
-
 #include "git2/sys/stream.h"
 
-extern int git_curl_stream_global_init(void);
-extern int git_curl_stream_new(git_stream **out, const char *host, const char *port);
+/** Configure stream registry. */
+int git_stream_registry_global_init(void);
+
+/** Lookup a stream registration. */
+extern int git_stream_registry_lookup(git_stream_registration *out, int tls);
```
pks-t: There's a type mismatch here. `tls` is no integer but an enum now

ethomson: Oops.  Thanks, fixed.

ethomson:
Rebased this to fix up conflicts after the certificate `GIT_PASSTHROUGH` changes.

Issue #4615, opened 2018-04-10
"error inflating zlib stream" when reading blobs > 4GB


mark-adams:
### Reproduction steps
1. Initialize a new repository
2. Commit a file larger than 4,294,967,295 bytes to a repository
3. Try to read the resulting blob using `git_tree_entry_to_object` or `git_object_lookup_prefix`

I've included some test scripts below that should make it simpler to duplicate.

```
$ cc -lgit2 test.c

$ ./create_repo.sh 4294967295 && ./a.out /tmp/test-repo testfile
Running with libgit2 0.26.0
Everything is broken! =(
Error: error inflating zlib stream

$ ./create_repo.sh 4294967294 && ./a.out /tmp/test-repo testfile
Running with libgit2 0.26.0
Everything is OK! =)
```

### Expected behavior
The expected object should be returned without any errors (`git cat-file` doesn't experience the same problem)

### Actual behavior
A `git_error` of "error inflating zlib stream" is generated

### Version of libgit2 (release number or SHA1)
 0.26.0 / 0.27.0

### Operating system(s) tested
Linux 4.13.0-38

#### create_repo.sh:
```
#!/bin/bash

rm -rf /tmp/test-repo
mkdir -p /tmp/test-repo

cd /tmp/test-repo
git init 2>&1 > /dev/null
fallocate -l $1 /tmp/test-repo/testfile
git add testfile
git commit -m "Commit a test file" 2>&1 > /dev/null
```

#### test.c:
```
#include <git2.h>
#include <stdio.h>


int main(int argc, const char * argv[]) {
    const char* repopath = argv[1];
    const char* filename = argv[2];

    git_libgit2_init();

    int major, minor, rev;
    git_libgit2_version(&major, &minor, &rev);
    printf("Running with libgit2 %d.%d.%d\n", major, minor, rev);

    git_repository *repo;
    git_repository_open(&repo, repopath);

    git_reference *ref;
    git_reference_lookup(&ref, repo, "refs/heads/master");
    const git_oid *commit_id = git_reference_target(ref);

    git_commit *commit;
    git_commit_lookup(&commit, repo, commit_id);

    git_tree *tree;
    git_commit_tree(&tree, commit);
    const git_tree_entry *entry = git_tree_entry_byname(tree, filename);

    git_object *object;
    int err = git_tree_entry_to_object(&object, repo, entry);

    if (err != 0){
        const git_error *giterr;
        giterr = giterr_last();

[...]

PR #4874, opened 2018-11-01
Test that largefiles can be read through the tree API

tiennou:
This is the test case for #4615, ported to clar.

Now I'm hoping `fcntl` is something portable 🤞.

### DIFF
--- a/src/win32/posix_w32.c
+++ b/src/win32/posix_w32.c
@@ -518,8 +518,54 @@ int p_creat(const char *path, mode_t mode)
 
 int p_fallocate(int fd, off_t offset, off_t len)
 {
-	error = ENOSYS;
-	return -1;
+	HANDLE fh = (HANDLE)_get_osfhandle(fd);
+	LARGE_INTEGER zero, position, oldsize, newsize;
+	size_t size;
+
+	if (fh == INVALID_HANDLE_VALUE) {
+		errno = EBADF;
+		return -1;
+	}
+
+	if (offset < 0 || len <= 0) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	if (git__add_sizet_overflow(&size, offset, len)) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	zero.u.LowPart = 0;
+	zero.u.HighPart = 0;
+
+	newsize.u.LowPart = (size & 0xffffffff);
+
+#if (SIZE_MAX > UINT32_MAX)
+	newsize.u.HighPart = size >> 32;
+#else
+	newsize.u.HighPart = 0;
+#endif
+
+	if (!GetFileSizeEx(fh, &oldsize)) {
+		set_errno();
+		return -1;
+	}
+
+	/* POSIX emulation: attempting to shrink the file is ignored */
+	if (oldsize.QuadPart >= newsize.QuadPart)
+		return 0;
+
+	if (!SetFilePointerEx(fh, zero, &position, FILE_CURRENT) ||
+	    !SetFilePointerEx(fh, newsize, NULL, FILE_BEGIN) ||
+	    !SetEndOfFile(fh) ||
+	    !SetFilePointerEx(fh, position, 0, FILE_BEGIN)) {
+		set_errno();
+		return -1;
+	}
+
+	return 0;
 }
 
 int p_utimes(const char *path, const struct p_timeval times[2])
--- a/src/win32/posix_w32.c
+++ b/src/win32/posix_w32.c
@@ -516,6 +516,12 @@ int p_creat(const char *path, mode_t mode)
 	return p_open(path, O_WRONLY | O_CREAT | O_TRUNC, mode);
 }
 
+int p_fallocate(int fd, off_t offset, off_t len)
+{
+	error = ENOSYS;
+	return -1;
+}
+
 int p_utimes(const char *path, const struct p_timeval times[2])
 {
 	git_win32_path wpath;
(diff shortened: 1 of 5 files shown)


tiennou:
Note that it's a green test locally, so it look like it's fixed (on the off-chance that the build is red, or that invasive tests are not enabled — I don't recall).

tests/object/tree/read.c
```diff
@@ -73,3 +73,59 @@ void test_object_tree_read__two(void)
...
+
+	if (!cl_is_env_set("GITTEST_INVASIVE_FS_SIZE"))
+		cl_skip();
+
+	cl_git_pass(git_reference_lookup(&ref, g_repo, "refs/heads/master"));
+	cl_git_pass(git_repository_index(&idx, g_repo));
+
+	git_buf_puts(&file, git_repository_workdir(g_repo));
+	git_buf_joinpath(&file, file.ptr, BIGFILE);
+	cl_git_pass(git_buf_oom(&file));
```
pks-t: I'd rather avoid this call and wrap the previous to function calls in `cl_git_pass`

tests/object/tree/read.c
```diff
@@ -73,3 +73,59 @@ void test_object_tree_read__two(void)
...
+
+	fd = p_open(git_buf_cstr(&file), O_CREAT|O_RDWR, 0644);
+	cl_assert_(fd >= 0, "invalid file descriptor");
+
+	memset(&prealloc, 0, sizeof(prealloc));
+	prealloc.fst_flags  = F_ALLOCATEALL;
+	prealloc.fst_posmode = F_PEOFPOSMODE;
+	prealloc.fst_offset = 0;
+	prealloc.fst_length = (off_t)4 * 1024 * 1024 * 1024; /* 4 GiB */
```
pks-t: Unfortunately, this is not portable at all. But there's `posix_fallocate` -- it requires _POSIX_C_SOURCE>=200112L, which is currently only set on SunOS and Solaris. No idea why.

tests/object/tree/read.c
```diff
@@ -73,3 +73,59 @@ void test_object_tree_read__two(void)
...
+	git_tree *tree;
+	git_oid oid;
+	const git_tree_entry *entry;
+	git_object *object;
+	git_buf file = GIT_BUF_INIT;
+	int fd;
+	fstore_t prealloc;
+	git_index *idx;
+
+	if (!cl_is_env_set("GITTEST_INVASIVE_FS_SIZE"))
```
pks-t: This reminds me that we still ain't got this environment variable set on Azure

tests/object/tree/read.c
tiennou: As macOS doesn't have that, I'm not sure how to proceed (or if it's worth implementing a shim for a "successful" test case).

tests/object/tree/read.c
tiennou: Filed #4933 for that.

tests/object/tree/read.c
tiennou: I've used preprocessor tricks to make it work on Linux/macOS.

tests/object/tree/read.c
```diff
@@ -73,3 +73,71 @@ void test_object_tree_read__two(void)
...
+	prealloc.fst_posmode = F_PEOFPOSMODE;
+	prealloc.fst_offset = 0;
+	prealloc.fst_length = BIGFILE_SIZE;
+
+	cl_must_pass(fcntl(fd, F_PREALLOCATE, &prealloc));
+	cl_must_pass(p_close(fd));
+#elif defined(GIT_WIN32)
+	/* Not implemented */
+#else
+	cl_must_pass(posix_fallocate(fd, 0, BIGFILE_SIZE));
```
ethomson: I think you're leaking this file descriptor here.

tests/object/tree/read.c
```diff
@@ -73,3 +73,71 @@ void test_object_tree_read__two(void)
...
+	git_object *object;
+	git_buf file = GIT_BUF_INIT;
+	int fd;
+#ifdef __APPLE__
+	fstore_t prealloc;
+#endif
+	git_index *idx;
+
+#ifdef GIT_WIN32
+	cl_skip();
```
ethomson: This concept [does exist in win32](https://blogs.msdn.microsoft.com/oldnewthing/20150710-00/?p=45171).  It might be nice to have a `p_fallocate` that abstracts the macOS / win32 / POSIX functionality.

ethomson: (If you want me to do the win32 bits, I'm happy to.)

tiennou: Yeah, it's just that I wasn't sure adding something to our POSIX layer only for a testcase was a nice use of library size (though maybe it can be deadcode-ed). If you're fine with this, I'd definitely prefer going that route.

tiennou: I ended up doing just that, and there's a WIN32 stub (for now 😉).

ethomson:
Attempting to add a Win32 version of `p_fallocate`...

src/win32/posix_w32.c
```diff
@@ -516,6 +516,48 @@ int p_creat(const char *path, mode_t mode)
 	return p_open(path, O_WRONLY | O_CREAT | O_TRUNC, mode);
 }
 
+int p_fallocate(int fd, off_t offset, off_t len)
```
pks-t: Might make sense to have some tests in tests/core/posix.c, right?

ethomson: I added a test - I'm always a little nervous testing the POSIX stuff because it's easy to get caught up testing everybody else's implementation.  😀  But let's give it a shot.

ethomson:
Cool cool.  Thanks, @tiennou!

Issue #4644, opened 2018-05-04
Error: SSL error: error:140E0197:SSL routines:SSL_shutdown:shutdown while in init


b-3-n:
### Reproduction steps
Unknown. We use nodegit and possibly several threads are cloning and pulling on different repositories at the same time. To avoid any confusion. No two threads ever clone or pull on the same repository. 

### Expected behavior
Can clone or pull repository.

### Actual behavior
Clone and pull fail with 
`Error: SSL error: error:140E0197:SSL routines:SSL_shutdown:shutdown while in init`

### Version of libgit2 (release number or SHA1)
We're using nodegit 0.20.3 and thus libgit v0.26.0.

### Operating system(s) tested
Ubuntu 18.04

Our application that uses nodegit/libgit runs for hours and clones/pulls hundreds of repositories successfully before - for no clear reason - the mentioned error occurs. After the error starts occurring nodegit/libgit is not recovering from it

ethomson:
How does your application (or, perhaps in this case, nodegit) handle its configuration of OpenSSL's locking?  Does it call [`git_openssl_set_locking`](https://github.com/libgit2/libgit2/blob/89c332e41b12a72d89de40d63bc568c56a2c336a/include/git2/sys/openssl.h#L34) or does it configure OpenSSL's locking itself?

pks-t:
We only ever shut down OpenSSL when the last user calls `git_libgit2_shutdown()` (that is, you need to call `git_libgit2_shutdown()` exactly as many times as you've called `git_libgit2_init()`.) What people frequently do is to call these two functions in every thread that uses libgit2, which may make code simpler in some ways, but some subtle races might be encounterd in these scenarios. Most importantly, it could cause us to repeatedly reinitialize libgit2's state, which is kind of unnecessary.

So what might happen here is that you by some mean get an imbalance of these two calls, causing us to deinitialize libgit2 while it is still in use. This is only speculative, but if that is the case that is nothing we can fix on our side. What I'd recommend in that scenario is that you should maybe only globally initialize libgit2 once at the beginning, before spawning threads, and than once at the end, when you don't want to use libgit2 anymore.

Some other scenario might be that you're using OpenSSL functions somewhere outside of libgit2. So if you call `git_libgit2_shutdown()`, we tear down OpenSSL for you, but you might still want to use it somewhere else. Again, the solution would be to only shutdown libgit2 at the very end.

[...]

PR #4875, opened 2018-11-01
Some OpenSSL issues

tiennou:
Reading #4644 was eerily familiar, so it made me start looking for "that patch I had written but now I can't find it" 😉.

This is extracted from #4786, I'm not sure if it is the complete story, but at least it would help downstream (and us) to pinpoint the reason for those failures.

### DIFF
--- a/src/streams/openssl.c
+++ b/src/streams/openssl.c
@@ -373,10 +373,10 @@ static int ssl_set_error(SSL *ssl, int error)
 	switch (err) {
 	case SSL_ERROR_WANT_CONNECT:
 	case SSL_ERROR_WANT_ACCEPT:
-		giterr_set(GITERR_NET, "SSL error: connection failure");
+		giterr_set(GITERR_SSL, "SSL error: connection failure");
 		break;
 	case SSL_ERROR_WANT_X509_LOOKUP:
-		giterr_set(GITERR_NET, "SSL error: x509 error");
+		giterr_set(GITERR_SSL, "SSL error: x509 error");
 		break;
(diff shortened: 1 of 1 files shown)


src/streams/openssl.c
```diff
@@ -373,10 +373,10 @@ static int ssl_set_error(SSL *ssl, int error)
 	switch (err) {
 	case SSL_ERROR_WANT_CONNECT:
 	case SSL_ERROR_WANT_ACCEPT:
-		giterr_set(GITERR_NET, "SSL error: connection failure");
+		giterr_set(GITERR_SSL, "SSL error: connection failure");
```
pks-t: I think these changes make sense. The other secure streams also use `GITERR_SSL` for various SSL-related failures

src/streams/openssl.c
```diff
@@ -602,6 +600,8 @@ int openssl_connect(git_stream *stream)
 	if ((ret = SSL_connect(st->ssl)) <= 0)
 		return ssl_set_error(st->ssl, ret);
 
+	st->connected = true;
```
pks-t: Makes sense, too. One might wonder about whether the BIO will still get free'd if we never call `SSL_shutdown`. But documentation of `SSL_set_bio` states that BIO's ownership is transferred to the SSL handle and will get free'd with `BIO_free_all` as soon the SSL handle is free'd.

ethomson:
Thanks!

tiennou:
@ethomson Though it's minor, also next-release (unless it was backported).

```
* Fix some issues with the error-reporting in the OpenSSL backend
```

PR #4877, opened 2018-11-01
Make cred structure more private

tiennou:
I don't think we want those structs to be "open", as they're only useful for code that has to perform authentication (ie. custom transport). Hence, let's move them in a `sys` header so they're less likely to be depended upon. As I don't expect anyone to have used those structs internal details, I think it's okay from a BC standpoint.

Additionally, rename the `GIT_CREDTYPE_DEFAULT` to `GIT_CREDTYPE_NEGOTIATE`, so it's less confusing. This one's breaking, but I'm really not fond of having "default" mean something specific, since at a glance it looks like the default authentication mechanism to use (which it is not).

### DIFF
--- a/include/git2/sys/cred.h
+++ b/include/git2/sys/cred.h
@@ -30,7 +30,7 @@ typedef struct git_cred_username {
 } git_cred_username;
 
 /** A key for NTLM/Kerberos "default" credentials */
-typedef struct git_cred git_cred_default;
+typedef struct git_cred git_cred_negotiate;
 
 /**
  * A ssh key from disk
--- a/include/git2/sys/cred.h
+++ b/include/git2/sys/cred.h
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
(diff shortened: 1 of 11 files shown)


pks-t:
While I'd be all for it if this were done from the beginning as it is the right thing, I'm not sure I want to merge this right now. This is potentially a thing for which packagers will hate us once again, as we break downstream users of libgit2 without any real benefit. Granted, I found only one user (guile-git) that will definitely break because of these changes. But our picture is at most an incomplete one, and thus I'm very hesitant to make such changes right now.

ethomson:
I’m 👎 on this.  “Default credentials” is what one generally calls [logged in user credentials](https://docs.microsoft.com/en-us/dotnet/api/system.net.credentialcache.defaultcredentials?view=netframework-4.7.2).  Even if it is imperfect, I think it’s more clear than “negotiate” (which is an authentication mechanism not a credential type; other mechanisms support default credentials) and there’s no reason to break this.

tiennou:
Good points in both case. The rename was not the most important, though I appreciate the rationale given. I'm closing, since it doesn't look like this can go in.

PR #4879, opened 2018-11-06
Allow certificate and credential callbacks to decline to act

ethomson:
Custom transports may want to ask libgit2 to invoke a configured credential or certificate callback; however they likely do not know if a callback was actually configured.  Return a sentinal value (greater than zero) if there is no callback configured instead of crashing.

With this, we want to define what `> 0` means for these callbacks, because they themselves might return it.  Credential callbacks are already defined to return `> 0` to defer these checks, but they were only inconsistently implemented (winhttp supported this, http and ssh transports did not).  Add this support consistently to all transports for both credential callbacks and certificate checking.

### DIFF
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -367,6 +367,7 @@ static int on_headers_complete(http_parser *parser)
 								  allowed_auth_types,
 								  t->owner->cred_acquire_payload);
 
+				/* treat GIT_PASSTHROUGH as if callback isn't set */
 				if (error == GIT_PASSTHROUGH) {
 					no_callback = 1;
 				} else if (error < 0) {
@@ -634,6 +635,9 @@ static int http_connect(http_subtransport *t)
 
 		giterr_clear();
 		error = t->owner->certificate_check_cb(cert, is_valid, t->connection_data.host, t->owner->message_cb_payload);
+
+		if (error == GIT_PASSTHROUGH)
+			error = is_valid ? 0 : GIT_ECERTIFICATE;
 
 		if (error < 0) {
 			if (!giterr_last())
--- a/src/transports/ssh.c
+++ b/src/transports/ssh.c
@@ -447,11 +447,11 @@ static int request_creds(git_cred **out, ssh_subtransport *t, const char *user, 
 		error = t->owner->cred_acquire_cb(&cred, t->owner->url, user, auth_methods,
 						  t->owner->cred_acquire_payload);
 
-		if (error == GIT_PASSTHROUGH)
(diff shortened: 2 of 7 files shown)


tiennou:
Did a quick review, so LGTM.

This looks like it's related to #3440 (there's also a test case in #4042 you might be able to reuse).

src/transports/http.c
```diff
@@ -367,11 +367,15 @@ static int on_headers_complete(http_parser *parser)
 								  allowed_auth_types,
 								  t->owner->cred_acquire_payload);
 
-				if (error == GIT_PASSTHROUGH) {
-					no_callback = 1;
-				} else if (error < 0) {
+				/* treat GIT_PASSTHROUGH as if callback isn't set */
+				if (error == GIT_PASSTHROUGH)
+					error = 1;
```
pks-t: This makes me wonder whether we should just document that `GIT_PASSTHROUGH` is treated in a special way instead of accepting positive return values. Especially with regards to consistency with other user-provided callback functions, where we say that we try to pass back positive error values unchanged.

ethomson:
@pks-t I agree that `GIT_PASSTHROUGH` is the more sensible value.  I've fixed this up to use that instead.

@tiennou Sorry, it looks like I'm stepping on your toes a bit with #4042.  😰  I'll take a look and see how these two might interact.

tiennou:
I don't really mind. I hope you're going to fix it right here, but in any case I think it's easier to do so after your cleanup. #4042 is about ignoring a callback's presence in case `GIT_EPASSTHROUGH` is returned.

src/transports/ssh.c
```diff
@@ -584,6 +584,11 @@ static int _git_ssh_setup_conn(
 		cert_ptr = &cert;
 
 		error = t->owner->certificate_check_cb((git_cert *) cert_ptr, 0, host, t->owner->message_cb_payload);
+
+		/* If the callback chose not to act, fail this hostkey */
+		if (error == GIT_PASSTHROUGH)
+			error = GIT_ECERTIFICATE;
```
pks-t: This is the only location that actually left me wondering. From the documentation, you state that we should behave "as if there was no callback set". But that's not was is happening here, because you do actually abort the connection with a certificate error -- instead, just judging from documentation, we should continue after this `if(certificate_check_cb != NULL)` block

ethomson: Agree; not sure what I was thinking.  Fixed.

pks-t:
Thanks Ed!

Issue #4872, opened 2018-10-30
Support for SSH ProxyCommand


logancollins:
The built-in SSH transport currently only supports socket connections directly to a remote host. I'm attempting to write support for Krypton (https://krypt.co) to work with libgit2, but the lack of ProxyCommand (which Krypton uses to perform its operations), means this is a bit of a non-starter unless:

1. I write my own custom transport, reimplementing all features of it on top of libssh2 again
2. The built-in transport is amended to support delegating to the host app for ProxyCommand-like behaviors

I'd prefer not to do the former, as that means not using the already great built-in SSH transport. With that in mind, here are some thoughts for how hooks to support ProxyCommand could be implemented in the library.

From my (limited) knowledge, OpenSSH-style ProxyCommand works, on Unix at least, by starting a subprocess by shell command (either via fork-exec or another method), setting up a socket between the host and subprocess via stdio, and passing all traffic into and out of that subprocess's socket as if it were the remote socket created for normal SSH operations. Krypton also works this way, using its `krssh` executable, which is a wrapper around `ssh`.

This is obviously all very specific to Posix, and implemented differently on Windows, and not supported at all on iOS due to the lack of multiprocess support. Also, most of ProxyCommand's details are stored in a user's SSH config file, and as such, it's unlikely that the `libgit2` library wants to handle anything that specific: SSH config files, exec'ing processes, etc.

With that in mind, the SSH transport currently creates its socket using `git_socket_stream_new`, passing it the expected host and port. A very simple option is to create a callback that asks the host application to create the socket for `libgit2` before this, taking the host and port as arguments, and returning the file descriptor to be used by the libssh2 implementation. This would allow the host app to perform the actual socket creation, or pass through to fall back on the default behavior. The library would not need to handle ProxyCommand-related cross-platform issues. Since the socket is just a file descriptor either way, libssh2 likely doesn't care?

This does, however, introduce a remote callback that is very specific to the SSH transport. I've been thinking about better ways to abstract this operation out (similar to how credential callbacks work, as those have SSH-specific functionality), but have not come up with anything concrete, as this is a very specific case related mostly to just SSH over the other transport types.

I'd love to hear the project contributors' thoughts on how best something like this should be implemented. I'm happy to work on it, but definitely wanted to propose everything and get some feedback before I got started making API-related changes to a fork of the project.

Thanks!

### Version of libgit2 (release number or SHA1)

Currently I am at tag 0.27.0.

### Operating system(s) tested

[...]

PR #4881, opened 2018-11-07
SSH Custom Callbacks to allow support for ProxyCommand and SSH configs

logancollins:
This pull request, in response to my proposals in #4872, contains modifications to the built-in SSH transport to expose custom callback functionality, allowing a user application to more carefully customize the libssh2 SSH transport and how it behaves, without resorting to reimplementing a separate smart subtransport and losing all of the built-in functionality of it.

There were two main goals with this:
1. I need to be able to pass `libgit2` a custom socket.
2. I need to be able to set options on the `LIBSSH2_SESSION` object before its handshake takes place.

… and I really wanted to do it without needing to implement my own transport. 😄

## Changes

The API changes are limited to changes in `sys/transport.h`, while implementation changes are in `socket.c/.h` and `ssh.c`. An overview of the changes follows.

### API Changes

- A new SSH transport creation method, `git_transport_ssh_with_options`, that may be used in the `git_transport_cb` remote callback to create an SSH transport with custom options
- A new `ssh_transport_options` struct that is used with the above method, which contains two new callbacks:
    - A callback to allow the host application to return a custom socket descriptor that is used instead of an INET socket created internally. This is to allow functionality like ProxyCommand, which uses a proxy socket to a child process.
    - A callback to allow the host application to customize the `LIBSSH2_SESSION *` object created by the SSH transport, before its handshake takes place.  This is the place where host apps would generally apply things from the user's `~/.ssh/config`, such as restricting available ciphers and key-exchanges.

### Implementation Changes

- Added a new field to the `ssh_subtransport` struct to hold the custom options, which remains `NULL` when using all previous SSH transport API.
- Added new private API in `socket.h` to create a `git_socket_stream` object with a `GIT_SOCKET` that already exists (otherwise falling back to a value of `INVALID_SOCKET`). `socket_connect` now checks for the existence of this, and skips the `socket()` and `connect()` steps.
- Modified a few internal functions in `ssh.c` to pass along the remote's URL, for use in callbacks invoked from those methods. This change is mainly centered around `_git_ssh_session_create`.

### Sockets

I could not figure out a good way to create a callback for the options struct which returns a socket descriptor and is also cross-platform. Sockets on Posix are just file descriptors / ints. Sockets on Win32 are of type `SOCKET`, which is a potentially varying type and is generally included from `<winsock2.h>`.

libgit2 handles this difference internally with its `GIT_SOCKET` definition. However, without pulling that type into public headers (which I really didn't even want to attempt), I decided just to add an include of `<winsock2.h`> to the `sys/transport.h` header if `GIT_WIN32` is true. This required an `ifdef` branch in the header, which is… not great.

I completely understand if this is not really acceptable (since libgit2's public headers have no includes that reference outside libraries. I'm open to suggestions of how to accomplish this (returning a socket from user code) without such preprocessor branching. Perhaps it would be better to add a CMake flag that even turns on such functionality, to prevent breaks in backwards-compatibility on Windows?

### Unit Tests

I have not yet added any unit tests for this functionality. I mainly want to get these API changes reviewed somewhat to see if this is even a good plan, but I'll also look into adding some tests for this. Even just something that ensures the custom SSH transport callbacks function correctly.

## Conclusion

The good news is, these changes makes it possible for our code to use Krypton and other ProxyCommand functionality, as well apply options from the user's SSH config! Which is great!


pks-t:
Cool, thanks for your contribution! Unfortunately, I ain't got enough time to review this today. I hope to be able to do so on Friday, but please feel free to ping me if I forget to do so.

logancollins:
No worries at all! This isn't hugely pressing, so review it at your convenience.

logancollins:
I got busy and lost track of this. Could I bump it to hopefully get someone to take a look? 🙏

pks-t:
On Wed, Mar 06, 2019 at 01:51:51PM -0800, Logan Collins wrote:
> I got busy and lost track of this. Could I bump it to hopefully get someone to take a look? 🙏

Sure, please go ahead. I'll try to have a look at it

dsully:
Any chance this could be merged?

ethomson:
Two things here:

* libssh2 is... not a great dependency for us. With that in mind
* This level of customization is why we allow custom ssh backends, IMO.

logancollins:
> Two things here:
> 
> * libssh2 is... not a great dependency for us. With that in mind
> * This level of customization is why we allow custom ssh backends, IMO.

Just for my own clarity, is the project's intent to move away from directly supporting libssh2 in favor of doing all SSH via custom backends, or does that _just_ cover things like these custom callbacks? (Whatever you all intend is fine, I'm just planning for the potential of the project dropping direct support for libssh2 if we need to be aware of it.)

ethomson:
I was closing PRs as a sort of a scream test. Consider your scream heard. 😁 

I don't have any interest in dropping support for libssh2. I don't _love_ libssh2, but we've also gotten a lot of utility out of it.

If this is still of use to you, let's revive this and figure out how to move forward? I assume you've been keeping this patch going forward for your own use. Let's figure out how to get you on a version of libgit2 out-of-the-box without patches.

logancollins:
Yeah, we have. It's not a _lot_ of code to reapply with each release so it's been easy enough for us to do.

I think the most important factor that has driven us to maintain it is ProxyCommand support (which a surprising number of our users use), and essentially we start things with a socket or file descriptor that is used to build a libssh2 session, as opposed to libgit2 building the session itself.

So either via something like this PR does, or via a mechanism you all feel is better to maintain going forward (even if the answer is a custom backend or more refined callbacks or delegation), it's probably a fine solution.

ethomson:
Thanks for the details; we should figure this out then.

PR #4883, opened 2018-11-09
signature: fix out-of-bounds read when parsing timezone offset

pks-t:
When parsing a signature's timezone offset, we first check whether there
is a timezone at all by verifying that there are still bytes left to
read following the time itself. The check thus looks like `time_end + 1
< buffer_end`, which is actually correct in this case. After setting the
timezone's start pointer to that location, we compute the remaining
bytes by using the formula `buffer_end - tz_start + 1`, re-using the
previous `time_end + 1`. But this is in fact missing the braces around
`(tz_start + 1)`, thus leading to an overestimation of the remaining
bytes by a length of two. In case of a non-NUL terminated buffer, this
will result in an overflow.

The function `git_signature__parse` is only used in two locations. First
is `git_signature_from_buffer`, which only accepts a string without a
length. The string thus necessarily has to be NUL terminated and cannot
trigger the issue.

The other function is `git_commit__parse_raw`, which can in fact trigger
the error as it may receive non-NUL terminated commit data. But as
objects read from the ODB are always NUL-terminated by us as a
cautionary measure, it cannot trigger the issue either.

In other words, this error does not have any impact on security.

### DIFF
--- a/src/signature.c
+++ b/src/signature.c
@@ -248,7 +248,7 @@ int git_signature__parse(git_signature *sig, const char **buffer_out,
 
 			if ((tz_start[0] != '-' && tz_start[0] != '+') ||
 			    git__strntol32(&offset, tz_start + 1,
-					   buffer_end - tz_start + 1, &tz_end, 10) < 0) {
+					   buffer_end - tz_start - 1, &tz_end, 10) < 0) {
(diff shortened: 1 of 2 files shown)


pks-t:
online::badssl is hitting again :/

PR #4884, opened 2018-11-12
index: introduce git_index_iterator

ethomson:
Provide a public git_index_iterator API that is backed by an index snapshot.  This allows consumers to provide a stable iteration even while manipulating the index during iteration.


ethomson:
Oy, thanks for pointing out my arithmetic error.  🙄  Fixed up with your suggestions.

pks-t:
Looks good to me, thanks!

PR #4885, opened 2018-11-13
apply: small fixups in the test suite

pks-t:
Some minor fixups in the "apply" test suite that have been noticed by Coverity.


ethomson:
🙏  Thanks!

Issue #4887, opened 2018-11-14
revwalk hide callback not changeable?


TheBB:
The documentation for *git_revwalk_new* says

> This revision walker uses a custom memory pool and an internal commit cache, so it is relatively expensive to allocate.
> 
> For maximum performance, this revision walker should be reused for different walks.

This is facilitated by *git_revwalk_reset* which removes all state on the revwalk object *except* (it seems) the hide callback function. This would be fine if I could just unset the callback myself, but *git_revwalk_add_hide_cb* errors out if a callback is already set, even if all I'm trying to do is to set it to `NULL`.

So it seems in the current state that the revwalk object can be reused but only if you want to use the same callback function. This seems a little restrictive, and I don't see any reason why we can't change the callback, either in *git_revwalk_add_hide_cb* or *git_revwalk_reset*, or both.

I'd be happy to submit a fix so long as someone indicates which solution is preferable.

ethomson:
> This is facilitated by git_revwalk_reset which removes all state on the revwalk object except (it seems) the hide callback function.

This _feels_ right to me, without having thought about it in depth.  When I think about resetting this, it simply starts over at the beginning, it doesn't remove the configuration.

> This would be fine if I could just unset the callback myself, but git_revwalk_add_hide_cb errors out if a callback is already set, even if all I'm trying to do is to set it to NULL.

Indeed, this is regrettable.  I think that it's absolutely reasonable to be able to clear a `hide_cb` that you've already set.

> I'd be happy to submit a fix so long as someone indicates which solution is preferable.

Awesome!  🎉

PR #4888, opened 2018-11-14
revwalk: Allow changing hide_cb

TheBB:
Remove the check for existing `hide_cb` and conditionally set the `limited` flag.

Fixes https://github.com/libgit2/libgit2/issues/4887

### DIFF
--- a/src/revwalk.c
+++ b/src/revwalk.c
@@ -756,15 +756,11 @@ int git_revwalk_add_hide_cb(
 	if (walk->walking)
 		git_revwalk_reset(walk);
 
-	if (walk->hide_cb) {
-		/* There is already a callback added */
-		giterr_set(GITERR_INVALID, "there is already a callback added to hide commits in revwalk");
-		return -1;
-	}
-
 	walk->hide_cb = hide_cb;
 	walk->hide_cb_payload = payload;
-	walk->limited = 1;
+
+	if (hide_cb)
+		walk->limited = 1;
 
 	return 0;
 }
--- a/tests/revwalk/hidecb.c
+++ b/tests/revwalk/hidecb.c
@@ -117,13 +117,40 @@ void test_revwalk_hidecb__hide_none_cb(void)
 	git_revwalk_free(walk);
 }
 
-void test_revwalk_hidecb__add_hide_cb_multiple_times(void)
+void test_revwalk_hidecb__unset_cb_before_walk(void)
 {
 	git_revwalk *walk;
+	git_oid id;
+	int i, error;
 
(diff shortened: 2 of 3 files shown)


neithernut:
You may also want to test for `git_revwalk_add_hide_cb(walk, NULL, NULL)`.

TheBB:
Sure, added a test for it.

tests/revwalk/hidecb.c
```diff
@@ -123,7 +123,24 @@ void test_revwalk_hidecb__add_hide_cb_multiple_times(void)
 
 	cl_git_pass(git_revwalk_new(&walk, _repo));
 	cl_git_pass(git_revwalk_add_hide_cb(walk, hide_every_commit_cb, NULL));
-	cl_git_fail(git_revwalk_add_hide_cb(walk, hide_every_commit_cb, NULL));
+	cl_git_pass(git_revwalk_add_hide_cb(walk, hide_every_commit_cb, NULL));
+	cl_git_pass(git_revwalk_add_hide_cb(walk, NULL, NULL));
```
neithernut: Good. But you don't verify that the revwalk actually works (e.g. by calling `git_revwalk_next()`).

TheBB: Ah, gotcha. Updated again.

src/revwalk.c
```diff
@@ -756,15 +756,11 @@ int git_revwalk_add_hide_cb(
...
-		giterr_set(GITERR_INVALID, "there is already a callback added to hide commits in revwalk");
-		return -1;
-	}
-
 	walk->hide_cb = hide_cb;
 	walk->hide_cb_payload = payload;
-	walk->limited = 1;
+
+	if (hide_cb)
+		walk->limited = 1;
```
pks-t: Mh. We don't zero out the struct members, so you'd probably also have to unlimit the revwalk in case there is no `hide_cb`. So:

```
walk_limited = !!hide_cb;
```

tests/revwalk/hidecb.c
```diff
@@ -123,7 +123,46 @@ void test_revwalk_hidecb__add_hide_cb_multiple_times(void)
...
+
+	git_revwalk_free(walk);
+}
+
+void test_revwalk_hidecb__unset_cb_before_walk(void)
```
pks-t: This test and the previous one are actually exactly the same, except for that this new one verifies that we _did_ actually unset the hide callback. So you should probably just delete the old test

TheBB: Yep, updated.

pks-t:
Thanks a lot for this change, @TheBB!

PR #4891, opened 2018-11-17
refs: add support for core.logAllRefUpdates=always

### LINKED ISSUE #4889: Clone fails with core.logAllRefUpdates=always
### Reproduction steps

1. `git config --global core.logAllRefUpdates always`
1. Attempt to clone a repository using libgit2 or a language wrapper
1. Receive an error: "failed to parse 'always' as a boolean value"

### Expected behavior

The git configuration is honored and the clone happens successfully.

### Actual behavior

An error occurs.  The clone fails.

git2go v27 (libgit2 838a2f2918b6d9fad8768d2498575ff5d75c35f0)

I've confirmed that this should be broken in master as well, as on master the option is read as a boolean value, and "always" is not a boolean value.

macOS

tiennou:
Since we were not expecting this config entry to contain a string, we would fail as soon as its (cached) value would be accessed. Hence, provide some constants for the 4 states we use, and account for "always" when we decide to reflog changes.

Fixes #4889.


ethomson:
Thanks!  I merged this manually - our style is:

```
switch (...) {
case ...:
    break;
case ...:
   break;
}
```

Instead of:

```
switch (...) {
    case:
        break;
    case:
        break;
}
```

(Note the lack of an additional indentation.)

tiennou:
@ethomson This is also in next-release.

Thanks for the manual merge, I was a little under the weather back in November. And I blame Xcode for the indents 😉.

PR #4892, opened 2018-11-18
worktree: Expose git_worktree_add_init_options

osener:
This change adds missing `GIT_EXTERN` declaration to `git_worktree_add_init_options` function.


ethomson:
Thanks for noticing this, and thanks for the fix!  👍

PR #4895, opened 2018-11-21
Unused function warnings

pks-t:
Commit 4b84db6a5 (patch_parse: remove unused function `parse_number`, 2018-11-14) led me to wonder why the compiler didn't emit any warnings at all for this unused function. Turns out we were disabling unused function warnings altogether back in 2014, which I think is unfortunate. Removing code is the best thing one can do if it doesn't break any features, so we should be notified by the compiler as soon as there's anything that's unneeded.

This PR re-enables the warnings. As it uses `ENABLE_WARNING`, unused functions will create compilation errors with -DENABLE_WERROR. Let's see how many builds break by this.


ethomson:
Neat.  I'm 👍 on more better warnings as long as we don't have to jump through too many hoops (or `#ifdef` magic).  Certainly looks like that's the case here.  While we're doing this, curious what the impact of `ENABLE_WARNINGS(unused-const-variable)` is?

pks-t:
Yeah, it seems like we're mostly good and don't need any additional ifdef magic. The only ifdef change was in the win32 test stuff, but that one already used ifdefs, so we're no worse off than before.

I've noticed that there's also another unused function `tree_iterator_entry_cmp`, which you've added in be30387e8 (iterators: refactored tree iterator, 2016-02-25), but in fact it was dead from the beginning. Is this a bug or just a leftover?

And I didn't yet dare to touch unused-const-variable. I'd bet there are quite some headers that contain constant variables like that that aren't touched at all. But I'll check.

pks-t:
Okay, I think I know now what the issue was with those unused functions. It's mostly about khash, which we define as static inline functions in a header file. If that header is included but not all of its functions are used, then Clang will complain. I'll see if I can do something about it.

neithernut:
It appears that clang's [`-Wunneeded-internal-declaration`](http://clang.llvm.org/docs/DiagnosticsReference.html#wunneeded-internal-declaration) is the closest to gcc's `-Wunused-function` regarding the handling of `static inline` functions (as found [here](https://gcc.gnu.org/onlinedocs/gcc-8.2.0/gcc/Warning-Options.html#Warning-Options)). But it may not be exactly equivalent. For example, it also emits warnings concerning _variables_. (clang also features a [`-Wunused-function`](http://clang.llvm.org/docs/DiagnosticsReference.html#wunused-function) option, which apparently doesn't take into account whether a function is inline or not).

Unfortunately, clang's documentation is rather lacking in a lot of places, including its command line options. In comparison, the the gcc man page is quite explicit.

pks-t:
@ethomson: got any idea why I can't see the build status here? Is it because I've force-pushed?

ethomson:
I'm not sure why it's not showing the "checks" section, but it is showing the red X next to commit `4a26f14`.   I'll see if I can understand what's going on here.

pks-t:
Thanks for your remarks by the way, @neithernut. I hope to be able to also get Clang to not complain about our inlines by limiting the khash scope, but your proposed flag would definitely be the fallback option in case I cannot get it to build.

ethomson:
Neat - looks like clang is happy indeed.  Are you happy with this now @pks-t or were you thinking about changing something else?

pks-t:
No, I'm very happy with this PR. Finally we've contained all this khash stuff into where they are used to implement our map interfaces, which is a very good thing. Next, we should probably try to fix how the map interfaces look like -- they are still very macro-y, even though we are able to use function semantics just fine now.

PR #4896, opened 2018-11-25
Fix warning C4133 incompatible types in MSVC

csware:
Introduced in commit b433a22a979ae78c28c8b16f8c3487e2787cb73e.


ethomson:
Thanks!

PR #4898, opened 2018-11-28
config: fix adding files if their parent directory is a file

### LINKED ISSUE #4897: Diff errors out failing to find the config if HOME=/dev/null
### Reproduction steps

1. Set `HOME=/dev/null`
1. Try to open the global configuration (or have `git_diff_tree_to_tree` do it as part of operation

### Expected behavior

Ignore being unable to find the global configuration

### Actual behavior

It returns an error. The stack trace I have is in ruby because we're hipsters but it boils down to calling diff in libgit2 (slightly redacted). This seems to be restricted to when `HOME` is a directory. Setting it to nonexitent paths does work.

```
[...]: failed to stat '/dev/null/.gitconfig'
[...]/gems/rugged-0.27.0.22.g238bca5/lib/rugged/tree.rb:149:in `diff_tree_to_tree'
[...]/gems/rugged-0.27.0.22.g238bca5/lib/rugged/tree.rb:149:in `diff'
[...]/gems/github-linguist-7.1.2/lib/linguist/repository.rb:139:in `compute_stats'
```

v0.27.5

Debian

pks-t:
When we try to add a configuration file with `git_config_add_file_ondisk`, we
treat nonexisting files as empty. We do this by performing a stat call, ignoring
ENOENT errors. This works just fine in case the file or any of its parents
simply does not exist, but there is also the case where any of the parent
directories is not a directory, but a file. So e.g. trying to add a
configuration file "/dev/null/.gitconfig" will fail, as `errno` will be ENOTDIR
instead of ENOENT.

Catch ENOTDIR in addition to ENOENT to fix the issue. Add a test that verifies
we are able to add configuration files with such an invalid path file just fine.

This should probably fix #4897.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -109,7 +109,7 @@ int git_config_add_file_ondisk(
 	assert(cfg && path);
 
 	res = p_stat(path, &st);
-	if (res < 0 && errno != ENOENT) {
+	if (res < 0 && errno != ENOENT && errno != ENOTDIR) {
 		giterr_set(GITERR_CONFIG, "failed to stat '%s'", path);
 		return -1;
 	}
--- a/tests/config/read.c
+++ b/tests/config/read.c
@@ -534,6 +534,26 @@ void test_config_read__fallback_from_local_to_global_and_from_global_to_system(v
 	git_config_free(cfg);
 }
 
+void test_config_read__parent_dir_is_file(void)
+{
+	git_config *cfg;
+	int count;
+
+	cl_git_pass(git_config_new(&cfg));
+	/*
+	 * Verify we can add non-existing files when the parent directory is not
+	 * a directory.
+	 */
+	cl_git_pass(git_config_add_file_ondisk(cfg, "/dev/null/.gitconfig",
+		GIT_CONFIG_LEVEL_SYSTEM, NULL, 0));
+
+	count = 0;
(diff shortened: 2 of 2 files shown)


ethomson:
Seems reasonable to me.  Thanks!

carlosmn:
Thanks. This works great for reading, but I wonder if we might leave users with a confusing message if they try to write. Part of the reason we treat nonexistent paths as empty config files is to allow you to e.g. grab the default set of configuration files and write into the appropriate one which might mean writing to the global one (tbh I think that's the only likely scenario where this would happen).

It's probably fine, but I wanted to point this possible unexpected case.

pks-t:
On Thu, Nov 29, 2018 at 02:39:07AM -0800, Carlos Martín Nieto wrote:
> Thanks. This works great for reading, but I wonder if we might
> leave users with a confusing message if they try to write. Part
> of the reason we treat nonexistent paths as empty config files
> is to allow you to e.g. grab the default set of configuration
> files and write into the appropriate one which might mean
> writing to the global one (tbh I think that's the only likely
> scenario where this would happen).
> 
> It's probably fine, but I wanted to point this possible
> unexpected case.

Yeah, I had this in mind already. Thing is that this is really no
different to what we currently have. E.g. HOME=/home/pks and
/home/pks does not exist. We will add the empty configuration
file /home/pks/.gitconfig just fine, but writing to it will fail
as we are not going to create the parent missing directory.

PR #4900, opened 2018-11-29
checkout regular files before symlinks

palmin:
`checkout_create_the_new` fails for symlinks when target does not exist,
making it important which order things are checked out in as a symlink
might target a regular file that is checked out later.

This is fixed by having `checkout_create_the_new` run a regular file pass
followed by a symlink pass.

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -1877,9 +1877,11 @@ static int checkout_deferred_remove(git_repository *repo, const char *path)
 
 static int checkout_create_the_new(
 	unsigned int *actions,
-	checkout_data *data)
+	checkout_data *data,
+	int symlink_pass)
 {
 	int error = 0;
+    int symlink_mode;
 	git_diff_delta *delta;
 	size_t i;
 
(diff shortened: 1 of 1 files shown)


ethomson:
That's surprising - I don't understand why there's any limitation on the target existing?  git can check out dangling symlinks without any trouble.  Can we not?

```
% git show e0d869b6d93444c02c8519a876e5f6fe1e8d4d4b
commit e0d869b6d93444c02c8519a876e5f6fe1e8d4d4b (HEAD -> master)
Date:   Thu Nov 29 22:07:16 2018 +0000

    whoo

diff --git a/bar.txt b/bar.txt
new file mode 120000
index 0000000..e07da69
--- /dev/null
+++ b/bar.txt
@@ -0,0 +1 @@
+nonexistent
\ No newline at end of file
% git reset --hard e0d869b6d93444c02c8519a876e5f6fe1e8d4d4b
HEAD is now at e0d869b y
% ls -Flas
total 0
0 drwxr-xr-x   4 ethomson  staff  128 Nov 29 22:09 ./
0 drwxr-xr-x  24 ethomson  staff  768 Nov 29 22:06 ../
0 drwxr-xr-x  14 ethomson  staff  448 Nov 29 22:08 .git/
0 lrwxr-xr-x   1 ethomson  staff   11 Nov 29 22:08 bar.txt@ -> nonexistent
```

palmin:
@ethomson I had `p_symlink` == `symlink` fail but I might have misunderstood why it failed.

Let me investigate further.

palmin:
@ethomson `symlink` failed because the symlink entry already existed as a regular file at this time. A path changed type from one commit to another.

Having this first pass somehow made the regular file disappear such that `symlink` could complete.

I will close this.

PR #4901, opened 2018-12-01
High-level map APIs

pks-t:
This is a complete rewrite of our map APIs. Previosuly, most places were using operations based on indices of the map entries, which is error prone, too tightly coupled with implementation details of the map and hard to read. This PR rewrites the API of all maps (str, oid, off, idx/idxicase) and introduces a set of high-level functions that act on those maps:

- new/free/realloc/clear to handle memory and lifecycle
- get/set to retrieve entries by key or to insert/update the value associated with a key
- delete to remove a key/value pair by key
- exists to check whether a key exists
- iterate to iterate over all key/value pairs

The endgame of this PR goes on to delete the old API based on indices completely.

Note: please do not merge yet. While it builds and passes all tests, I still need to go through the commits and fix up the commit messages.

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -65,12 +65,15 @@ void git_cache_dump_stats(git_cache *cache)
 int git_cache_init(git_cache *cache)
 {
 	memset(cache, 0, sizeof(*cache));
-	cache->map = git_oidmap_alloc();
-	GIT_ERROR_CHECK_ALLOC(cache->map);
+
+	if ((git_oidmap_new(&cache->map)) < 0)
+		return -1;
+
 	if (git_rwlock_init(&cache->lock)) {
 		git_error_set(GIT_ERROR_OS, "failed to initialize cache rwlock");
 		return -1;
 	}
+
 	return 0;
 }
 
--- a/src/cache.c
+++ b/src/cache.c
@@ -151,16 +151,12 @@ static bool cache_should_store(git_object_t object_type, size_t object_size)
 
 static void *cache_get(git_cache *cache, const git_oid *oid, unsigned int flags)
 {
-	size_t pos;
-	git_cached_obj *entry = NULL;
+	git_cached_obj *entry;
 
 	if (!git_cache__enabled || git_rwlock_rdlock(&cache->lock) < 0)
 		return NULL;
 
-	pos = git_oidmap_lookup_index(cache->map, oid);
-	if (git_oidmap_valid_index(cache->map, pos)) {
-		entry = git_oidmap_value_at(cache->map, pos);
-
+	if ((entry = git_oidmap_get(cache->map, oid)) != NULL) {
 		if (flags && entry->flags != flags) {
 			entry = NULL;
 		} else {
@@ -175,7 +171,7 @@ static void *cache_get(git_cache *cache, const git_oid *oid, unsigned int flags)
 
 static void *cache_store(git_cache *cache, git_cached_obj *entry)
 {
-	size_t pos;
+	git_cached_obj *stored_entry;
 
 	git_cached_obj_incref(entry);
 
@@ -194,10 +190,8 @@ static void *cache_store(git_cache *cache, git_cached_obj *entry)
 	if (git_cache__current_storage.val > git_cache__max_storage)
 		cache_evict_entries(cache);
 
-	pos = git_oidmap_lookup_index(cache->map, &entry->oid);
-
 	/* not found */
-	if (!git_oidmap_valid_index(cache->map, pos)) {
+	if ((stored_entry = git_oidmap_get(cache->map, &entry->oid)) == NULL) {
 		int rval;
 
 		git_oidmap_insert(cache->map, &entry->oid, entry, &rval);
@@ -209,19 +203,18 @@ static void *cache_store(git_cache *cache, git_cached_obj *entry)
 	}
 	/* found */
 	else {
-		git_cached_obj *stored_entry = git_oidmap_value_at(cache->map, pos);
-
 		if (stored_entry->flags == entry->flags) {
 			git_cached_obj_decref(entry);
 			git_cached_obj_incref(stored_entry);
 			entry = stored_entry;
 		} else if (stored_entry->flags == GIT_CACHE_STORE_RAW &&
-			entry->flags == GIT_CACHE_STORE_PARSED) {
+			   entry->flags == GIT_CACHE_STORE_PARSED) {
+			int rval;
+
 			git_cached_obj_decref(stored_entry);
 			git_cached_obj_incref(entry);
 
-			git_oidmap_set_key_at(cache->map, pos, &entry->oid);
-			git_oidmap_set_value_at(cache->map, pos, entry);
+			git_oidmap_insert(cache->map, &entry->oid, entry, &rval);
 		} else {
 			/* NO OP */
 		}
--- a/src/cache.c
+++ b/src/cache.c
@@ -192,10 +192,7 @@ static void *cache_store(git_cache *cache, git_cached_obj *entry)
 
 	/* not found */
 	if ((stored_entry = git_oidmap_get(cache->map, &entry->oid)) == NULL) {
-		int rval;
-
-		git_oidmap_insert(cache->map, &entry->oid, entry, &rval);
-		if (rval >= 0) {
+		if (git_oidmap_set(cache->map, &entry->oid, entry) == 0) {
 			git_cached_obj_incref(entry);
 			cache->used_memory += entry->size;
 			git_atomic_ssize_add(&git_cache__current_storage, (ssize_t)entry->size);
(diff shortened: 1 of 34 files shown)


ethomson:
Oooooooh, I'm excited to review this.  😀

pks-t:
The commits have been rewritten to have proper commit messages now. CI passes now, too, so this is ready for review

pks-t:
By the way, short hint: please have a special look at the commit "cache: use iteration interface for cache eviction". It changes the way we behave with regards to cache eviction, where we would previously evict starting at a random offset and now we evict starting at the first entry.

src/oidmap.h
```diff
@@ -116,10 +137,8 @@ void git_oidmap_insert(git_oidmap *map, const git_oid *key, void *value, int *rv
 size_t git_oidmap_begin(git_oidmap *map);
 size_t git_oidmap_end(git_oidmap *map);
 
-#define git_oidmap_foreach_value(h, vvar, code) { size_t __i;			\
-	for (__i = git_oidmap_begin(h); __i != git_oidmap_end(h); ++__i) {	\
-		if (!git_oidmap_has_data(h,__i)) continue;			\
-		(vvar) = git_oidmap_value_at(h,__i);				\
+#define git_oidmap_foreach_value(h, vvar, code) { size_t __i = 0;		\
```
tiennou: That's unrelated, but I dislike `code` being an argument… (à la `git_array_foreach`).

tests/core/strmap.c
```diff
@@ -35,17 +34,16 @@ static void insert_strings(git_strmap *table, size_t count)
...
-		git_strmap_insert(table, str, str, &err);
-		cl_assert(err >= 0);
+		cl_git_pass(git_strmap_set(table, str, str));
 	}
 
 	cl_assert_equal_i(git_strmap_size(table), count);
 }
 
-void test_core_strmap__1(void)
+void test_core_strmap__inserted_strings_can_be_retrieved(void)
```
tiennou: 👏

src/cache.c
```diff
@@ -115,8 +115,7 @@ void git_cache_free(git_cache *cache)
 /* Called with lock */
 static void cache_evict_entries(git_cache *cache)
 {
-	uint32_t seed = rand();
-	size_t evict_count = 8;
+	size_t evict_count = 8, i;
```
tiennou: commitmsg: `s/completley/completely/`.

src/cache.c
```diff
@@ -125,18 +124,19 @@ static void cache_evict_entries(git_cache *cache)
 		return;
 	}
 
+	i = 0;
```
tiennou: I might be completely wrong, but what about initializing this to `random(0, git_oidmap_size(cache->map) - evict_count)` ?

src/idxmap.c
```diff
@@ -72,6 +72,24 @@ void git_idxmap_icase_clear(git_idxmap_icase *map)
 	kh_clear(idxicase, map);
 }
 
+int git_idxmap_resize(git_idxmap *map, size_t size)
```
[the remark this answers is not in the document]
pks-t: In this example I disagree. Interestingly enough, `kh_resize` lets you resize maps to a _smaller_ size. Nobody should be doing that, but this is why I don't want to name it grow ;)

src/offmap.h
```diff
@@ -99,6 +99,27 @@ int git_offmap_delete(git_offmap *map, const git_off_t key);
...
+ *             with zero previous to the first call.
+ * @param key pointer to the variable where to store the current
+ *            key. May be NULL.
+ * @param value pointer to the variable where to store the current
+ *            value. May be NULL.
+ * @param `0` if the next entry was correctly retrieved.
+ *        GIT_ITEROVER if no entries are left. A negative error
+ *        code otherwise.
+ */
+int git_offmap_iterate(git_offmap *map, size_t *iter, git_off_t *key, void **value);
```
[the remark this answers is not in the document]
pks-t: Agreed

src/offmap.h
```diff
@@ -44,7 +44,13 @@ void git_offmap_free(git_offmap *map);
...
 void git_offmap_clear(git_offmap *map);
 
-size_t git_offmap_num_entries(git_offmap *map);
+/**
+ * Return the number of elements in the map.
+ *
+ * @parameter map map containing the elements
+ * @return number of elements in the map
+ */
+size_t git_offmap_size(git_offmap *map);
```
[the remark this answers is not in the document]
ethomson: And buffer has both `size` and `asize` but we ask for its `len`.   😢

I feel like we most commonly use `size`.  I don't _love_ that - it always feels to me like the allocated size, not the filled / used size of a struct - but I think that consistency is good here.

src/offmap.h
```diff
@@ -99,6 +99,27 @@ int git_offmap_delete(git_offmap *map, const git_off_t key);
  */
 int git_offmap_exists(git_offmap *map, const git_off_t key);
 
+/**
+ * Iterate over entries of the map.
+ *
+ * This functions allows to iterate over all key-value entries of
+ * the map. The current item is stored in the `iter` variable and
```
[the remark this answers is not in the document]
pks-t: Agreed

src/oidmap.h
pks-t: I totally agree. We could just transform the while-loop into a for loop, initializing `__i` in the loop initializer. But let's do this after this PR got merged, it already has a lot of code churn.

src/cache.c
pks-t: Fixed

src/cache.c
pks-t: Yeah, we sure could do this, but only because we do know the map's implementation details. We'd also need to wrap-around when reaching `git_oidmap_size`. But it sure keeps the original intention a lot better than my refactored code.

It just feels like the wrong data structure for the task at hand, at least in this specific cache eviction scenario. I'm not familiar with the code though, so I cannot say anything more to it. @ethomson, you've got any comment?

pks-t:
Rebased on latest master to fix conflicts and address most of @tiennou's comments. Thanks for reviewing!

pks-t:
Rebased to fix havoc caused by #4917 :)

src/offmap.h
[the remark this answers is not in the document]
pks-t: Yeah, had the same feeling. Thanks for chiming in!

ethomson:
One place that we get a boatloat of warnings on Win32 is in the khash users.  Right now we very often cast between `size_t` and `khiter_t`.  But `khiter_t` is actually defined as 32 bits.

I'm a bit surprised that there's no warnings on gcc/clang with the implicit cast, but alas.  I noticed that we're still using `size_t` for indices here; it would be nice if we started using `khiter_t` consistently to avoid that.

pks-t:
On Sat, Jan 26, 2019 at 01:18:43PM -0800, Edward Thomson wrote:
> One place that we get a boatloat of warnings on Win32 is in the khash users.  Right now we very often cast between `size_t` and `khiter_t`.  But `khiter_t` is actually defined as 32 bits.
> 
> I'm a bit surprised that there's no warnings on gcc/clang with the implicit cast, but alas.  I noticed that we're still using `size_t` for indices here; it would be nice if we started using `khiter_t` consistently to avoid that.

I explicitly chose not to expose any implementation details of
the khash stuff so that we can completley avoid including
"khash.h" in our own header files. I'll try to come up with an
alternative solution, though.

ethomson:
True.  This is definitely not worse than what we have now in the khiter_t/size_t differences, and more encapsulated.  It’s not necessarily something we need to deal with before merging.

I’m crazy busy this week but I’ll try to find some time to take a look at this more closely.

pks-t:
On Mon, Jan 28, 2019 at 12:51:37AM -0800, Edward Thomson wrote:
> I’m crazy busy this week but I’ll try to find some time to take a look at this more closely.

No worries. I'll also be busy due to the upcoming Git Merge, so
there's no need to hurry :)

pks-t:
Rebased to verify that no new map users of the old API have been introduced in the meantime.

pks-t:
@ethomson: you got some time soonish to get this reviewed? Compiling with MSVC again reminded me of all the warnings that our current use of khash generates :/

src/cache.c
ethomson: Hrmf.  No?  If I'm reading this correctly, we walk through the cache evicting items in-order, the difference is the starting point.  It feels like if we're going to go to any lengths to change our eviction strategy that we should actually have a strategy based on age or something and not the ordering of the oid.  Starting at a random place in the list seems like it might be slightly better than always evicting starting at the beginning of the map (I can imagine the latter enabling some pathological cases) but they both sort of suck anyway.

(This might be what you were referring to @pks-t with the "wrong data structure" comment.)

pks-t: > It feels like if we're going to go to any lengths to change our eviction strategy that we should actually have a strategy based on age or something and not the ordering of the oid.

That's exactly my thought here, too. An LRU cache might make a lot more sense.It would require us to keep track of the cached objects twice, though: once in the LRU queue for cache eviction and once in the map for quick lookup via OID.

So the question is how to proceed right now. I don't feel like introducing the `random(0, git_oidmap_size(cache->map) - evict_count)` thing, as it again ties us to the map implementation details. I know it's unlikely that it'll change anyway soon, but it just doesn't feel right. If you say that'd be required to get this merged, then fine, so be it and I'll just use @tiennou's proposal. But in that case, we should definitely revisit and fix it soon.

ethomson: No, I don't think that this is significantly worse (in terms of cache eviction) than what's being replaced.  0, random, it's pretty mediocre either way.  I'll file an issue but let's merge this now.

ethomson:
Huge improvement.  Thanks!  🎉

PR #4904, opened 2018-12-03
Update CRLF filtering to match modern git

ethomson:
#3285 indicated that we had issues with compatibility when CRLF-filtering files when adding them to the index.  This takes that pull request with a few modifications:

1. Move the existing `crlf_data` folders from `posix` and `windows` to have a suffix indicating that they're the "to workdir" direction.  This will let us add the "to odb" folders in the same location.
2. Instead of creating a second script to generate a corpus of CRLF test data, this modifies the existing test script.  This means that instead of four invocations (the to-workdir and to-odb scripts on both POSIX and Windows) we now only have two invocations (the CRLF script on both POSIX and Windows).
3. I've tried to simplify the tests that use this corpus a bit - and especially important for debugging, adding a message that indicates which combination of `autocrlf` / `safecrlf` / attributes settings is failing.
4. Instead of trying to reverse engineer git's settings, I've just taken the changes that they've made.  There were a number of changes made by Torsten Bögershausen, and since he [gave permission for us to use his changes](https://github.com/libgit2/libgit2/blob/master/git.git-authors#L77) we can simple examine what he did to git and replicate it ourselves.  I've done that here.
5. Cherry-pick the remaining test cases that #3285 added.

Big thanks to @csware for all the work on this.  Having the test corpus from git is a _huge_ help to ensure that we are getting this right.


ethomson:
There were some questions in #3285 about the value of checking in all this test data.  Yes, it is obnoxious but very useful.  It's hard to generate this on-the-fly since:

1. We would require a dependency on having git on the remote system.  And although it's a pretty safe bet that git is installed on anybody's machine who's developing this (or any build system that's running a CI build), it's not clear what version of git is installed.  This system lets us decide what version of git we want to emulate.
2. More importantly, creating all these files ends up with some disk I/O and a _lot_ of `exec`s of various git commands.  In other words, it's _slow_ on Windows.  _Really_ slow.  Checking in the data is definitely preferred.

We _could_ do something like check in a zip that we extract, which would have the benefit of only having a single file checked in to the repository, but that feels like the _only_ benefit and otherwise feels icky to me.

tests/checkout/crlf.c
```diff
@@ -107,6 +108,9 @@ static void test_checkout(const char *autocrlf, const char *attrs)
 
 	git_buf_puts(&reponame, "crlf");
 
+	git_buf_puts(&systype_and_direction, systype);
+	git_buf_puts(&systype_and_direction, "_to_workdir");
```
pks-t: We definitely need to check return values for these here. Otherwise, if we run into an OOM, then we'll continue just fine and do the wrong thing later on. Same for all the other, pre-existing buffer operations

tests/checkout/crlf.c
```diff
@@ -13,14 +13,38 @@ static git_repository *g_repo;
...
+	git_buf reponame = GIT_BUF_INIT;
+
 	g_repo = cl_git_sandbox_init("crlf");
 
+	/*
+	 * remove the contents of the working directory so that we can
+	 * check out over it.
+	 */
+	git_buf_puts(&reponame, "crlf");
+	cl_git_pass(git_path_direach(&reponame, 0, unlink_file, NULL));
```
pks-t: Can't we use your nifty new workdir iterator for that?

tests/index/crlf.c
```diff
@@ -14,16 +14,221 @@
...
+	struct compare_data *cd = payload;
+	bool failed = true;
+	int cmp_git, cmp_gitattributes;
+	char *basename;
+	int add_bypath_ret;
+	const git_index_entry *entry;
+	git_oid oid;
+	git_error_state error = { 0 };
+
+	basename = git_path_basename(actual_path->ptr);
```
pks-t: `cl_assert` missing around this here, as it may actually fail in OOM situations

tests/index/crlf.c
```diff
@@ -14,16 +14,221 @@
...
+	giterr_state_capture(&error, add_bypath_ret);
+
+	entry = git_index_get_bypath(g_index, basename, 0);
+	if (!add_bypath_ret && !entry)
+		goto done;
+
+	cl_git_pass(git_buf_joinpath(&expected_path_oid, cd->dirname, basename));
+	cl_git_pass(git_buf_joinpath(&expected_path_fail, cd->dirname, basename));
+	git_buf_puts(&expected_path_oid, ".obj");
+	git_buf_puts(&expected_path_fail, ".fail");
```
pks-t: `cl_git_pass`

tests/index/crlf.c
```diff
@@ -14,16 +14,221 @@
...
+	struct compare_data compare_data = { NULL, safecrlf, autocrlf, attrs };
+	const char *c;
+
+	git_buf_puts(&reponame, "crlf");
+
+	git_buf_puts(&sandboxname, "safecrlf_");
+	git_buf_puts(&sandboxname, safecrlf);
+
+	git_buf_puts(&sandboxname, ",autocrlf_");
+	git_buf_puts(&sandboxname, autocrlf);
```
pks-t: `cl_git_pass` missing for all of these

tests/index/crlf.c
```diff
@@ -14,16 +14,221 @@
...
+		cl_git_mkfile("crlf/.gitattributes", attrbuf.ptr);
+	}
+
+	cl_repo_set_string(g_repo, "core.safecrlf", safecrlf);
+	cl_repo_set_string(g_repo, "core.autocrlf", autocrlf);
+
+	cl_git_pass(git_index_clear(g_index));
+
+	git_buf_joinpath(&expected_dirname, "crlf_data", "checkin_results");
+	git_buf_joinpath(&expected_fixture, expected_dirname.ptr, sandboxname.ptr);
```
pks-t: Here, too

tests/index/crlf.c
```diff
@@ -14,16 +14,221 @@
...
+	git_buf_free(&reponame);
+}
+
+static void set_up_workingdir(const char *name)
+{
+	git_vector contents = GIT_VECTOR_INIT;
+	size_t i;
+	const char *fn;
+
+	git_path_dirload(&contents, name, 0, 0);
```
pks-t: `cl_git_pass`

tests/index/crlf.c
```diff
@@ -14,16 +14,221 @@
...
+	git_vector_foreach(&contents, i, fn) {
+		char *basename = git_path_basename(fn);
```
pks-t: `cl_assert`

tests/index/crlf.c
```diff
@@ -14,16 +14,221 @@
...
+		bool skip = strncasecmp(basename, ".git", 4) == 0 && strlen(basename) == 4;
+
+		git__free(basename);
+
+		if (skip)
+			continue;
+		p_unlink(fn);
```
pks-t: `cl_git_pass`

tests/index/crlf.c
```diff
@@ -14,16 +14,221 @@
...
+	}
+	git_vector_free_deep(&contents);
+
+	/* copy input files */
+	git_path_dirload(&contents, cl_fixture("crlf_data/checkin_input_files"), 0, 0);
```
pks-t: And here

tests/index/crlf.c
```diff
@@ -43,84 +43,61 @@ struct compare_data
...
+		cl_git_pass(git_futils_readbuffer(&expected_contents, expected_path_fail.ptr));
 		git_buf_rtrim(&expected_contents);
 
-		if (error.error_msg.klass != GITERR_FILTER || strstr(error.error_msg.message, expected_contents.ptr) == NULL)
-			goto done;
+		cl_git_fail(git_index_add_bypath(g_index, basename));
+		cl_assert_equal_i(GITERR_FILTER, giterr_last()->klass);
+		cl_assert_equal_s(expected_contents.ptr, giterr_last()->message);
+	} else {
+		cl_fail("unexpected index failure");
```
pks-t: Ugh. Okay, so you do already fix up a lot of these. Maybe want to squash those two commits together?

tests/index/crlf.c
```diff
@@ -199,13 +176,18 @@ static void set_up_workingdir(const char *name)
...
-	git_path_dirload(&contents, cl_fixture("crlf_data/checkin_input_files"), 0, 0);
+	git_path_dirload(&contents, cl_fixture("crlf"), 0, 0);
 	git_vector_foreach(&contents, i, fn) {
 		char *basename = git_path_basename(fn);
 		git_buf dest_filename = GIT_BUF_INIT;
-		git_buf_joinpath(&dest_filename, name, basename);
+
+		if (strcmp(basename, ".gitted") &&
+			strcmp(basename, ".gitattributes")) {
+			git_buf_joinpath(&dest_filename, name, basename);
```
pks-t: `cl_git_pass`

tests/index/crlf.c
```diff
@@ -339,13 +339,55 @@ void test_index_crlf__autocrlf_input_text_auto_attr(void)
...
+
+	cl_git_mkfile("./crlf/.gitattributes", "* text=auto\n");
+
+	cl_repo_set_string(g_repo, "core.autocrlf", "input");
+	cl_repo_set_bool(g_repo, "core.safecrlf", true);
+
+	cl_git_mkfile("./crlf/newfile.txt", FILE_CONTENTS_LF);
+
+	cl_git_pass(git_index_add_bypath(g_index, "newfile.txt"));
+	entry = git_index_get_bypath(g_index, "newfile.txt", 0);
```
pks-t: `cl_assert`

tests/index/crlf.c
```diff
@@ -339,13 +339,55 @@ void test_index_crlf__autocrlf_input_text_auto_attr(void)
...
+	const git_index_entry *entry;
+	git_oid oid;
+
+	cl_repo_set_string(g_repo, "core.autocrlf", "input");
+	cl_repo_set_bool(g_repo, "core.safecrlf", true);
+
+	cl_git_mkfile("./crlf/newfile.txt", FILE_CONTENTS_LF);
+
+	cl_git_pass(git_index_add_bypath(g_index, "newfile.txt"));
+	entry = git_index_get_bypath(g_index, "newfile.txt", 0);
```
pks-t: `cl_assert`

src/crlf.c
```diff
@@ -122,147 +112,168 @@ static int has_cr_in_index(const git_filter_source *src)
...
+			if (ca->safe_crlf == GIT_SAFE_CRLF_WARN) {
+				/* TODO: issue a warning when available */
+			} else {
+				if (filename && *filename)
+					giterr_set(
+						GITERR_FILTER, "CRLF would be replaced by LF in '%s'",
+						filename);
+				else
+					giterr_set(
+						GITERR_FILTER, "CRLF would be replaced by LF");
```
pks-t: Oh, how I wish that git'd just track the files you add, without ever caring about their contents at all and leaving that problem to the user instead.

csware: With SafeCRLF enabled Git must care for the content.

pks-t: I know, that's my point ;)

tests/index/crlf.c
ethomson: Yeah, I did that so that @csware's commits kept him as the author to make sure that he got credit in the log.

csware: Thanks ;)

tests/checkout/crlf.c
ethomson: Fixed.

ethomson:
Thanks, @pks-t, I'm going through and resolving your comments as I fix them locally.  I'm stepping away now but I'll push up fixes shortly.

ethomson:
OK, I added some missing `cl_git_pass` and `cl_assert`s.  I did it in a new commit instead of trying to fixup the prior commits - again, to keep authorship straight (and because it's a bit easier).  😀

PR #4905, opened 2018-12-04
make proxy_stream_close close target stream even on errors

palmin:
When a `git_filter_apply_fn` callback returns a error while smudging `proxy_stream_close`
ends up [returning](https://github.com/libgit2/libgit2/blob/master/src/filter.c#L829) without closing the stream. This is turn makes `blob_content_to_file`
crash as it [asserts](https://github.com/libgit2/libgit2/blob/master/src/checkout.c#L1555) the stream being closed whether there are errors or not.

Closing the target stream on error fixes this problem, with the downside being that
error strings set by `proxy_stream->target->close()` will override ones set by `git_filter_apply_fn`
callback.

### DIFF
--- a/src/filter.c
+++ b/src/filter.c
@@ -809,6 +809,7 @@ static int proxy_stream_close(git_writestream *s)
 {
 	struct proxy_stream *proxy_stream = (struct proxy_stream *)s;
 	git_buf *writebuf;
+	git_error_state error_state = {0};
 	int error;
 
 	assert(proxy_stream);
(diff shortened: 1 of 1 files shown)


ethomson:
> Closing the target stream on error fixes this problem, with the downside being that
error strings set by proxy_stream->target->close() will override ones set by git_filter_apply_fn
callback.

Yeah, I think that we should keep the original error message(s).  You can `giterr_save` and `giterr_restore` to avoid this problem.

palmin:
It's ready for another look.

ethomson:
Thanks!

PR #4910, opened 2018-12-14
Support creating annotated commits from annotated tags

### LINKED ISSUE #4909: git_annotated_commit_from_ref() fails if passed in reference is a tag
I don't know if this is on purpose, but if a (non-lightweight) tag is passed as a reference to git_annotated_commit_from_ref(), it fails. That's because target oid of resolved reference passed to annotated_commit_init_from_id() belongs to the tag, not to a commit. For this to work, the (non-lightweight) tag has to be peeled to a commit, and then pass that commit's oid to annotated_commit_init_from_id().

### Operating system(s) tested: macOS 10.14.2

carlosmn:
This got a bit messy with the `const`ness of `git_reference_peel` and I'm happy to move that to its own PR if y'all want.

The crux of this is to use `git_reference_peel` instead of `git_reference_resolve` so we get the id of the commit regardless of how much wrapping there is around it.

This fixes #4909

### DIFF
--- a/include/git2/refs.h
+++ b/include/git2/refs.h
@@ -710,7 +710,7 @@ GIT_EXTERN(int) git_reference_normalize_name(
  */
 GIT_EXTERN(int) git_reference_peel(
 	git_object **out,
-	git_reference *ref,
+	const git_reference *ref,
 	git_object_t type);
 
 /**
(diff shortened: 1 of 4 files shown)


pks-t:
No need to split this into multiple PRs, I'd say. Looks good to me, thanks!

PR #3075, opened 2015-04-26
Split the concerns of the commit creation functions

carlosmn:
As discussed on #2681 this PR splits up the commit creation functions into creating the commit and those which update a reference as well.

`_fromstate()` brings us closer to `git-commit` with its defaults for the parents and tree.

I don't love `git_commit_create_on_head()` since you can get the same by passing `"HEAD"` to `git_commit_create_on()` and we already have enough confusion about what we mean by `head` in function names.

The names are whatever I came up with, better names may come up.

PR #4913, opened 2018-12-18
Add sign capability to git_rebase_commit

implausible:
This PR aims to add the ability to sign a rebase commit using the git_rebase_commit API. It seemed most convenient to extend the method to take a signature callback and signature_field callback, which help provide the signature data for `git_commit_create_with_signature` API.

The general flow is that if you have options and you have a `signature_cb`, then `git_rebase_commit` will call that `signature_cb` with the commit content buffer retrieved from `git_commit_create_buffer`, providing an opportunity to sign the commit content. The owner of the `signature_cb` is expected to create a string, though if GIT_PASSTHROUGH is specified, the signing process will fallback to creating an unsigned commit. If a signature is provided from the `signature_cb`, the `signature_field_cb` will be called if it is provided. We will then create the commit using `git_commit_create_from_signature`.

If `NULL` options are passed, the commit is created using the previously implemented flow.

I have added additional tests to confirm that signing works in git_rebase_commit and fixed any tests that would've broken from a change in API.

### DIFF
--- a/include/git2/commit.h
+++ b/include/git2/commit.h
@@ -480,7 +480,8 @@ GIT_EXTERN(int) git_commit_create_buffer(
  *
  * @param out the resulting commit id
  * @param commit_content the content of the unsigned commit object
- * @param signature the signature to add to the commit
+ * @param signature the signature to add to the commit. Leave `NULL`
+ * to create a commit without adding a signature field.
  * @param signature_field which header field should contain this
  * signature. Leave `NULL` for the default of "gpgsig"
  * @return 0 or an error code
--- a/include/git2/commit.h
+++ b/include/git2/commit.h
@@ -501,6 +501,21 @@ GIT_EXTERN(int) git_commit_create_with_signature(
  */
 GIT_EXTERN(int) git_commit_dup(git_commit **out, git_commit *source);
 
+/**
+ * Commit signing callback.
+ *
+ * The callback will be called with the commit content, giving a user an
+ * opportunity to sign the commit content. The signature_field
+ * buf may be left empty to specify the default field.
+ *
+ * When the callback:
+ * - returns GIT_PASSTHROUGH, no signature will be added to the commit.
+ * - returns < 0, commit creation will be aborted.
+ * - returns GIT_OK, the signature parameter is expected to be filled.
+ */
+typedef int (*git_commit_signing_cb)(
+	git_buf *signature, git_buf *signature_field, const char *commit_content, void *payload);
+
 /** @} */
 GIT_END_DECL
 #endif
--- a/include/git2/commit.h
+++ b/include/git2/commit.h
@@ -507,7 +507,13 @@ GIT_EXTERN(int) git_commit_dup(git_commit **out, git_commit *source);
  *
  * The callback will be called with the commit content, giving a user an
  * opportunity to sign the commit content. The signature_field
- * buf may be left empty to specify the default field.
+ * buf may be left empty to specify the default field "gpgsig".
+ *
+ * Signatures can take the form of any string, and can be created on an arbitrary
+ * header field. Signatures are most commonly used for verifying authorship of a
+ * commit using GPG or a similar cryptographically secure signing algorithm.
+ * See https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work for more
+ * details.
  *
  * When the callback:
  * - returns GIT_PASSTHROUGH, no signature will be added to the commit.
--- a/include/git2/rebase.h
+++ b/include/git2/rebase.h
@@ -134,7 +134,7 @@ typedef enum {
 #define GIT_REBASE_OPTIONS_VERSION 1
 #define GIT_REBASE_OPTIONS_INIT \
 	{ GIT_REBASE_OPTIONS_VERSION, 0, 0, NULL, GIT_MERGE_OPTIONS_INIT, \
-	  GIT_CHECKOUT_OPTIONS_INIT, 0, NULL }
+	  GIT_CHECKOUT_OPTIONS_INIT, NULL, NULL }
 
 /** Indicates that a rebase operation is not (yet) in progress. */
 #define GIT_REBASE_NO_OPERATION SIZE_MAX
--- a/include/git2/rebase.h
+++ b/include/git2/rebase.h
@@ -24,21 +24,6 @@
 GIT_BEGIN_DECL
 
 /**
- * Rebase commit signature callback.
- *
- * The callback will be called with the commit content, giving a user an
- * opportunity to sign the commit content in a rebase. The signature_field
- * buf may be left empty to specify the default field.
- *
- * When the callback:
- * - returns GIT_PASSTHROUGH, no signature will be added to the commit.
- * - returns < 0, git_rebase_commit will be aborted.
- * - returns GIT_OK, the signature parameter is expected to be filled.
- */
-typedef int (*git_rebase_commit_signature_cb)(
-	git_buf *signature, git_buf *signature_field, const char *commit_content, void *payload);
-
-/**
  * Rebase options
  *
  * Use to tell the rebase machinery how to operate.
@@ -95,7 +80,7 @@ typedef struct {
 	 * without a signature.
 	 * This field is only used when performing git_rebase_commit.
 	 */
-	git_rebase_commit_signature_cb signature_cb;
+	git_commit_signing_cb signing_cb;
 
 	/**
 	 * This will be passed to each of the callbacks in this struct
--- a/include/git2/rebase.h
+++ b/include/git2/rebase.h
@@ -24,6 +24,36 @@
 GIT_BEGIN_DECL
 
 /**
+ * Rebase commit signature callback.
+ *
+ * The callback will be called with the commit content, giving a user an
+ * opportunity to sign the commit content in a rebase.
+ * The signature parameter will be owned by LibGit2 after this callback returns.
+ *
+ * When the callback:
+ * - returns GIT_PASSTHROUGH, no signature will be added to the commit.
+ * - returns < 0, git_rebase_commit will be aborted.
+ * - returns GIT_OK, the signature parameter is expected to be filled.
+ */
+typedef int (*git_rebase_commit_signature_cb)(
+	char **signature, const char *commit_content, void *payload);
+
+/**
+ * Rebase commit signature field callback.
+ *
+ * The callback will be called if a signature_cb was called and successful.
+ * This callback will provide the field that a user is signing in a git_rebase_commit.
+ * The signature_field parameter will be owned by LibGit2 after this callback returns.
+ *
+ * When the callback:
+ * - returns GIT_PASSTHROUGH, signature_field is expected to remain null.
+ * - returns < 0, git_rebase_commit will be aborted.
+ * - returns GIT_OK, the signature_field parameter is expected to be filled.
+ */
+typedef int (*git_rebase_commit_signature_field_cb)(
+	char **signature_field, void *payload);
+
+/**
  * Rebase options
  *
  * Use to tell the rebase machinery how to operate.
@@ -72,6 +102,28 @@ typedef struct {
 	 * `abort` to match git semantics.
 	 */
 	git_checkout_options checkout_options;
+
+	/**
+	 * If provided, this will be called with the commit content, allowing
+	 * a signature to be added to the rebase commit. Can be skipped with
+	 * GIT_PASSTHROUGH. If GIT_PASSTHROUGH is returned, a commit will be made
+	 * without a signature.
+	 * This field is only used when performing git_rebase_commit.
+	 */
+	git_rebase_commit_signature_cb signature_cb;
+
+	/**
+	 * If provided and the signature_cb is provided, this will be called asking
+	 * for the field to write the signature to. Can be skipped with GIT_PASSTHROUGH.
+	 * This field is only used when performing git_rebase_commit.
+	 */
+	git_rebase_commit_signature_field_cb signature_field_cb;
+
+	/**
+	 * This will be passed to each of the callbacks in this struct
+	 * as the last parameter.
+	 */
+	void *payload;
 } git_rebase_options;
 
 /**
@@ -118,7 +170,7 @@ typedef enum {
 #define GIT_REBASE_OPTIONS_VERSION 1
 #define GIT_REBASE_OPTIONS_INIT \
 	{ GIT_REBASE_OPTIONS_VERSION, 0, 0, NULL, GIT_MERGE_OPTIONS_INIT, \
-	  GIT_CHECKOUT_OPTIONS_INIT}
+	  GIT_CHECKOUT_OPTIONS_INIT, 0, 0, NULL }
 
 /** Indicates that a rebase operation is not (yet) in progress. */
 #define GIT_REBASE_NO_OPERATION SIZE_MAX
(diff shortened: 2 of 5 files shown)


ethomson:
I understand the need here and this makes sense.  But it's disappointing to break the API.

If we move this callback into the `git_rebase_options` struct that gets passed in to the init function, can we avoid changing the signature of the commit API?

implausible:
I think that seems reasonable. I'll see what I can do about your request. Thank you for taking a look at this!

implausible:
Hopefully that looks a little better, @ethomson.

src/rebase.c
```diff
@@ -987,6 +1019,11 @@ static int rebase_commit__create(
 	if (error < 0)
 		git_commit_free(commit);
 
+	if (signature)
+		free(signature);
+	if (signature_field)
+		free(signature_field);
```
ethomson: We can't free memory that the user allocated because we don't know how it was allocated.  Win32, for example, has several different heap allocators.  Concrete example: if libgit2 is compiled as a release library and the application calling it is compiled as a debug binary, then the `malloc` from the application will be incompatible with the `free` in the library.

Instead, pass a `git_buf` to the callback and let them [`git_buf_set`](https://github.com/libgit2/libgit2/blob/master/include/git2/buffer.h#L117) it.

tests/rebase/sign.c
```diff
@@ -0,0 +1,252 @@
+#include "clar_libgit2.h"
+#include "git2/rebase.h"
+
+static git_repository *repo;
+static git_signature *signature;
+
+/* Fixture setup and teardown */
+void test_rebase_sign__initialize(void)
+{
+  repo = cl_git_sandbox_init("rebase");
```
ethomson: Please reformat this with tab indentation instead of two-spaces.

src/rebase.c
```diff
@@ -975,10 +977,40 @@ static int rebase_commit__create(
...
+				message_encoding, message, tree, 1, (const git_commit **)&parent_commit)) < 0)
+			goto done;
+
+		if ((error = rebase->options.signature_cb(&signature, git_buf_cstr(&commit_content),
+				rebase->options.payload)) < 0 && error != GIT_PASSTHROUGH)
+			goto done;
+
+		if (error != GIT_PASSTHROUGH) {
+			if (rebase->options.signature_field_cb &&
+				(error = rebase->options.signature_field_cb(&signature_field, rebase->options.payload)) < 0) {
```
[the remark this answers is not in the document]
tiennou: I'm also wondering about the genericity of this : given that this is mostly about customizing the commit creation when rebasing, I'd wager we'd hit an explosion of callbacks if someone encounters #3075 (like #4667 but for commits).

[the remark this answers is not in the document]
implausible: @tiennou Are you suggesting something along the lines of `git_rebase_commit_ext` or `git_rebase_commit_with_options`?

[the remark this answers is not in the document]
tiennou: Yeah, or a lower-level struct that would wrap all general commit-related information in it, instead of parameter-combinations to a couple of functions plus options for different behaviors.

```c
typedef struct {
    int version;
    git_tree *tree;
    git_buf *msg;
    git_oidarray parents;
    git_signature *author;
    git_signature *committer;
} git_commit_data;
```

This would make it possible to add custom commit headers (`encoding` being the only one we provide access to at the moment), or the signing callback you're after.

Maybe a more `git_treebuilder`-like API ?

I'd be glad to help with the API design, but I agree we should discuss it first 😉.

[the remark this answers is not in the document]
ethomson: Hmm.  One option is to have a commit creation callback - if provided, we just hand you the tree, parents, etc, and you hand us back a commit ID when you're done.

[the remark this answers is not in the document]
tiennou: Rereading, just to clarify : my initial suggestion was to add a lower-level `git_commit_create` variant that takes a commit-description structure (and maybe some behavioral options but that's part of what #3075 is about). Then that struct would be provided to the callback (the option @ethomson proposes), also insulating that callback's signature from unforeseen parameter changes.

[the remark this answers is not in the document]
implausible: Is this a direction that I should be moving this PR in or will the additional callback added to `git_rebase_options` suffice for now?

tests/rebase/sign.c
```diff
@@ -0,0 +1,249 @@
...
+
+
+int signature_cb_magic_field(
+	git_buf *signature,
+	git_buf *signature_field,
+	const char *commit_content,
+	void *payload)
+{
+	const char *signature_content = "magic word: pretty please";
+	const char * signature_field_content = "magicsig";
```
tiennou: Style: no space after `*`.

include/git2/rebase.h
```diff
@@ -72,6 +87,21 @@ typedef struct {
...
 	git_checkout_options checkout_options;
+
+	/**
+	 * If provided, this will be called with the commit content, allowing
+	 * a signature to be added to the rebase commit. Can be skipped with
+	 * GIT_PASSTHROUGH. If GIT_PASSTHROUGH is returned, a commit will be made
+	 * without a signature.
+	 * This field is only used when performing git_rebase_commit.
+	 */
+	git_rebase_commit_signature_cb signature_cb;
```
tiennou: Same here: `signature_cb` => `signing_cb`.

include/git2/rebase.h
```diff
@@ -23,6 +23,21 @@
...
+ * The callback will be called with the commit content, giving a user an
+ * opportunity to sign the commit content in a rebase. The signature_field
+ * buf may be left empty to specify the default field.
+ *
+ * When the callback:
+ * - returns GIT_PASSTHROUGH, no signature will be added to the commit.
+ * - returns < 0, git_rebase_commit will be aborted.
+ * - returns GIT_OK, the signature parameter is expected to be filled.
+ */
+typedef int (*git_rebase_commit_signature_cb)(
```
tiennou: I'm not sure it makes sense for this to be in the rebase namespace, though I can live with it. Are there other places where this kind of signing-trampoline would be useful (ie. cherry-pick) ?

But `signature` makes it sound like `git_signature` are involved, so maybe `git_(rebase_)?commit_signing_cb` ?

implausible: Currently, I don't believe cherrypick, merge, or revert actually create commits. If we were intending on extending those to create commits rather than an index, I would say it could very well be useful.

implausible:
@tiennou The tests failed due to a connection error. Is there any chance you can restart the offending suite?

tiennou:
I requeued the build. Note that you should be able to do so on your own PRs by commenting `/rebuild`. That will get picked up by a helpful bot 😉.

implausible:
Aha, I will use that next time :).

src/rebase.c
```diff
@@ -988,8 +988,10 @@ static int rebase_commit__create(
 
 		if ((error = rebase->options.signing_cb(&commit_signature, &signature_field,
 				git_buf_cstr(&commit_content), rebase->options.payload)) < 0 &&
-				error != GIT_PASSTHROUGH)
+				error != GIT_PASSTHROUGH) {
+			git_error_set(error, "signing_cb failed");
```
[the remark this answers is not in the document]
tiennou: The guidelines say that we should provide something in the event the callback didn't, but we should preserve something it might have already set. IIRC there's a helper method for those callback functions that takes care of this dance.

So, 
```c
if (git_error_last() == NULL)
    git_error_set(GIT_ERROR_COMMIT, "commit signing_cb failed");
```

Also, you're conflating error codes (ie. return values) with error classes (ie the `GIT_ERROR_*` values), taken as the 1st parameter of `git_error_set`.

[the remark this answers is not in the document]
implausible: Gotcha! Thanks!

src/rebase.c
```diff
@@ -986,10 +986,12 @@ static int rebase_commit__create(
 				message_encoding, message, tree, 1, (const git_commit **)&parent_commit)) < 0)
 			goto done;
 
+		git_error_clear();
 		if ((error = rebase->options.signing_cb(&commit_signature, &signature_field,
 				git_buf_cstr(&commit_content), rebase->options.payload)) < 0 &&
 				error != GIT_PASSTHROUGH) {
-			git_error_set(error, "signing_cb failed");
+			if (git_error_last() == NULL)
```
tiennou: `git_error_set_after_callback_function` 😉.

include/git2/rebase.h
```diff
@@ -118,7 +133,7 @@ typedef enum {
 #define GIT_REBASE_OPTIONS_VERSION 1
 #define GIT_REBASE_OPTIONS_INIT \
 	{ GIT_REBASE_OPTIONS_VERSION, 0, 0, NULL, GIT_MERGE_OPTIONS_INIT, \
-	  GIT_CHECKOUT_OPTIONS_INIT}
+	  GIT_CHECKOUT_OPTIONS_INIT, 0, NULL }
```
pks-t: These both need to be `NULL` pointers, right?

include/git2/commit.h
```diff
@@ -501,6 +501,21 @@ GIT_EXTERN(int) git_commit_create_with_signature(
  */
 GIT_EXTERN(int) git_commit_dup(git_commit **out, git_commit *source);
 
+/**
+ * Commit signing callback.
+ *
+ * The callback will be called with the commit content, giving a user an
+ * opportunity to sign the commit content. The signature_field
+ * buf may be left empty to specify the default field.
```
pks-t: I think a short comment specifying how such a signature needs to look like would be reasonable

src/rebase.c
```diff
@@ -975,10 +978,45 @@ static int rebase_commit__create(
 		message = git_commit_message(current_commit);
 	}
 
-	if ((error = git_commit_create(&commit_id, rebase->repo, NULL, author,
-		committer, message_encoding, message, tree, 1,
-		(const git_commit **)&parent_commit)) < 0 ||
-		(error = git_commit_lookup(&commit, rebase->repo, &commit_id)) < 0)
+	/* this error will be cleared by the signing process, but should be set
+	 * to signal the unsigned commit create process if we are not going to sign */
```
pks-t: Nit: multiline-comments should start with empty `/*` and `*/` lines. No need to fix this, though

src/rebase.c
```diff
@@ -975,10 +978,45 @@ static int rebase_commit__create(
...
+					signature_field_as_string)))
+				goto done;
+		}
+	}
+
+	/* if we skipped signing, create the commit normally */
+	if (error == GIT_PASSTHROUGH &&
+		(error = git_commit_create(&commit_id, rebase->repo, NULL, author, committer,
+			message_encoding, message, tree, 1, (const git_commit **)&parent_commit)) < 0)
+		goto done;
```
pks-t: The flow would be simplified by quite a bit if you simply always called `git_commit_create_with_signature`, leaving `commit_content` and `commit_signature` fields as `NULL` in case there was no signing callback provided or if it returned `GIT_PASSTHROUGH`.

tests/rebase/sign.c
```diff
@@ -0,0 +1,249 @@
...
+	cl_git_pass(git_rebase_next(&rebase_operation, rebase));
+	cl_git_pass(git_rebase_commit(&commit_id, rebase, NULL, signature, NULL, NULL));
+
+	git_oid_fromstr(&expected_id, "bf78348e45c8286f52b760f1db15cb6da030f2ef");
+	cl_assert_equal_oid(&expected_id, &commit_id);
+
+	cl_git_pass(git_commit_lookup(&commit, repo, &commit_id));
+	cl_assert_equal_s(expected_commit_raw_header, git_commit_raw_header(commit));
+
+	cl_git_fail(error = git_rebase_next(&rebase_operation, rebase));
```
pks-t: `cl_git_fail_with(GIT_ITEROVER, git_rebase_next(&rebase_operation, rebase));`

tests/rebase/sign.c
```diff
@@ -0,0 +1,249 @@
...
+	cl_git_pass(git_rebase_commit(&commit_id, rebase, NULL, signature, NULL, NULL));
+
+	git_oid_fromstr(&expected_id, "129183968a65abd6c52da35bff43325001bfc630");
+	cl_assert_equal_oid(&expected_id, &commit_id);
+
+	cl_git_pass(git_commit_lookup(&commit, repo, &commit_id));
+	cl_assert_equal_s(expected_commit_raw_header, git_commit_raw_header(commit));
+
+	cl_git_fail(error = git_rebase_next(&rebase_operation, rebase));
+	cl_assert_equal_i(GIT_ITEROVER, error);
```
pks-t: `cl_git_fail_with(GIT_ITEROVER, git_rebase_next(&rebase_operation, rebase));`

tests/rebase/sign.c
```diff
@@ -0,0 +1,249 @@
...
+	cl_git_pass(git_rebase_commit(&commit_id, rebase, NULL, signature, NULL, NULL));
+
+	git_oid_fromstr(&expected_id, "f46a4a8d26ae411b02aa61b7d69576627f4a1e1c");
+	cl_assert_equal_oid(&expected_id, &commit_id);
+
+	cl_git_pass(git_commit_lookup(&commit, repo, &commit_id));
+	cl_assert_equal_s(expected_commit_raw_header, git_commit_raw_header(commit));
+
+	cl_git_fail(error = git_rebase_next(&rebase_operation, rebase));
+	cl_assert_equal_i(GIT_ITEROVER, error);
```
pks-t: Same

pks-t:
Just to clarify: you didn't yet address the comments from both @tiennou and my reviews, did you? That would be totally fine, I just want to make sure that we do not have mismatching expectations on the current state of this PR.

src/rebase.c
implausible: This ended up being much simpler, though I needed to tweak git_commit_create_with_signature to support NULL signatures.

src/rebase.c
implausible: When I updated the flow to always use `git_commit_create_with_signature`, this comment no longer mattered.

implausible:
@pks-t @tiennou I've worked through your reviews.

tiennou:
This looks fine to me, but I'm really not fond of the layering, so I'm not "in love" API-wise… You can see the vision I had on #5116, but I haven't had the time to finish it (and I think it'd work better with  some kind of `git_commitbuilder` API instead of the struct I was playing with).

Thus, opinions on the low-level part of #5116 would be ❤️, and then those typedefs can move out to somewhere more generally useful ?

ethomson:
I'm +1 on merging this as-is, the changes to the public API seem reasonable to me.

It feels like this is not mutually exclusive to further refactoring the commit creation process and merging this wouldn't

ethomson:
Gah, sorry, tapped the close button accidentally.  🙄

src/commit.c
```diff
@@ -80,8 +80,8 @@ static int git_commit__create_buffer_internal(
 }
 
 static int validate_tree_and_parents(git_array_oid_t *parents, git_repository *repo, const git_oid *tree,
-				     git_commit_parent_callback parent_cb, void *parent_payload,
-				     const git_oid *current_id, bool validate)
+						 git_commit_parent_callback parent_cb, void *parent_payload,
```
tiennou: (There's something going on indent-wise in those 3 hunks).

implausible: No kidding! I wonder what happened there.

ethomson:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

implausible:
Thanks!!

PR #4915, opened 2018-12-19
refdb_fs: refactor error handling in `refdb_reflog_fs__delete`

pks-t:
The function `refdb_reflog_fs__delete` uses the `if (!error && foobar())`
pattern of checking, where error conditions are being checked by following calls
to different code. This does not match our current style, where the call-site of
a function is usually directly responsible for checking the return value.

Convert the function to use `if ((error = foobar()) < 0) goto out;` style. Note
that this changes the code flow a bit: previously, we were always trying to
delete empty reference hierarchies even if deleting the reflog entry has failed.
This wasn't much of a problem -- if deletion failed, the hierarchy will still
contain at least one file and thus the function call was an expensive no-op.
Now, we will only perform this deletion if we have successfully removed the
reflog.

### DIFF
--- a/src/refdb_fs.c
+++ b/src/refdb_fs.c
@@ -2046,28 +2046,27 @@ cleanup:
 
 static int refdb_reflog_fs__delete(git_refdb_backend *_backend, const char *name)
 {
-	int error;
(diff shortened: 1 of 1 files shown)


ethomson:
👍

PR #4917, opened 2018-12-27
Move `giterr` to `git_error`

ethomson:
Much like #4807, refactor `giterr` functions and enums to match more closely with our naming conventions.  This moves the `giterr_` functions to `git_error_`, while providing proxies with the old, now-deprecated names, for backward compatibility.  Similarly, this moves the `GITERR_` enums to `GIT_ERROR_` (while again providing backcompat).


ethomson:
(Apologies for the spam, I'm debugging the rebuilder)

ethomson:
Updating to fix some conflicts; I'll likely merge this soonish to unblock us for a release since I'm not hearing any yelling.

include/git2/errors.h
```diff
@@ -143,19 +149,90 @@ GIT_EXTERN(void) giterr_clear(void);
...
+
+/** @name Deprecated Error Functions
+ *
+ * These functions and enumeration values are retained for backward
+ * compatibility.  The newer versions of these functions should be
+ * preferred in all new code.
+ */
+/**@{*/
+
+#define GITERR_NONE GIT_ERROR_NONE
```
tiennou: An alternative would be to use `const int`, which would allow to mark those as GIT_DEPRECATED.

Last time I checked, compiler support for deprecating macros wasn't standardized.

pks-t:
That's a lot of changes :) But I'm all for it, so thanks for doing it. `giterr` has always annoyed me in the past.

PR #4919, opened 2019-01-02
Shutdown callback count

pks-t:
It randomly occurred to me that we can avoid manually updating the `MAX_SHUTDOWN_CB` macro whenever we add a new global initializer by just declaring the initializers as an array and then using the initializer array's size to declare the shutdown array. This PR does exactly that as a small cleanup.


ethomson:
👍

PR #4924, opened 2019-01-06
Make ENABLE_WERROR actually work

lhchavez:
This change explicitly adds -Werror to the CFLAGS.

Due to the way that the ADD_C_FLAG_IF_SUPPORTED() macro was mangling the
flag name to convert it into a define name, any warning that had a dash
in its name was not being correctly enabled. Additionally, any flag that
is enabled implicitly by the compiler (like -Wunused-result and
-Wdeprecated-declarations) would not cause an error unless they were
explicitly enabled with the ENABLE_WARNINGS() macro.


ethomson:
Hmm, it looks like this only works for GCC-like compilers.  I wonder if we might also enable warnings on errors in MSVC?  (`/WX`, if I recall correctly).

ethomson:
I'd also like @pks-t's thoughts here.  `ENABLE_WERROR` currently _upgrades_ the _explicit_ warnings that we've defined to be errors, it isn't a blanked `-Werror`.  The former option sounds like it remains useful, and that we might want to consider both.

Perhaps a `-DENABLE_WERROR=all` for the latter?

lhchavez:
hmm i thought that was the intention, given that `-Werror=all` and `-Werror=extra` are enabled here: https://github.com/libgit2/libgit2/blob/274a37fc78c4387e5902399600f3df85dbccbc25/CMakeLists.txt#L201-L202

enabling warnings on MSVC is going to take more effort, since I don't have a Windows machine and debugging on Travis is hard ^^;;

ethomson:
FYI, we don't use Travis.  :)

ethomson:
> hmm i thought that was the intention, given that -Werror=all and -Werror=extra are enabled here:

Yeah, I can't speak to that, as I don't think that I understand the distinction between `-Werror=all` and `-Wall -Werror`.  It seems like [it was very explicit](https://github.com/libgit2/libgit2/commit/175ab8e7653778c3ef7c805696c8a63d0023c727) to do `-Werror=all` instead of `-Wall -Werror`.

ethomson:
In any case, don't worry about MSVC, I can tackle that once we've decided the direction that we want to go in.

neithernut:
> Yeah, I can't speak to that, as I don't think that I understand the distinction between `-Werror=all` and `-Wall -Werror`. It seems like [it was very explicit](https://github.com/libgit2/libgit2/commit/175ab8e7653778c3ef7c805696c8a63d0023c727) to do `-Werror=all` instead of `-Wall -Werror`.

If I understand correctly, `-Wall -Werror` causes _all_ warnings to be considered errors, including warnings enabled by other switches (e.g. `-Wextra`). `Werror=all` causes _only_ those warnings enabled by `-Wall` to be considered errors.

ethomson:
> If I understand correctly, -Wall -Werror causes all warnings to be considered errors, including warnings enabled by other switches (e.g. -Wextra). Werror=all causes only those warnings enabled by -Wall to be considered errors.

That's my understanding, too.  Sorry, for additional context, what we do when `ENABLE_WERROR` is turned on is basically [go through the warnings that we enable](https://github.com/libgit2/libgit2/blob/6d6169314a1e49e6f052779d3e52da332ef23ee9/CMakeLists.txt#L225-L236) and [append `=error`](https://github.com/libgit2/libgit2/blob/e6cdd17c846648aa8e1e025fa1988475886a551e/cmake/Modules/EnableWarnings.cmake) to them.

In other words, instead of `-Wall -Wextra -Wdocumentation` we'll have `-Wall=error -Wextra=error -Wdocumentation=error`.

This feels like it should be the equivalent of `-Wall -Wextra -Wdocumentation -Werror` but it appears that it is not.  It feels like there's some warnings implicitly enabled (and then promoted to errors) by `-Werror`?

I'm definitely not understand the distinction yet.  😀

lhchavez:
yes, the two flags that i had trouble with ( `-Wmaybe-uninitialized -Wdeprecated-declarations`) are implicitly enabled _and_ not covered by `-Wall` or `-Werror`. an equivalent change would be to explicitly include all implicit warnings, but that seems subpar.

lhchavez:
All right, after a lot of debugging, I finally traced the root of the issue to the way the `ADD_C_FLAG_IF_SUPPORTED()` macro was being expanded.

Warning-free builds FTW!

cmake/Modules/AddCFlagIfSupported.cmake
```diff
@@ -7,7 +7,7 @@ INCLUDE(CheckCCompilerFlag)
 
 MACRO(ADD_C_FLAG _FLAG)
 	STRING(TOUPPER ${_FLAG} UPCASE)
-	STRING(REGEX REPLACE "^-" "" UPCASE_PRETTY ${UPCASE})
+	STRING(REGEX REPLACE "-" "_" UPCASE_PRETTY ${UPCASE})
```
tiennou: I've given this a local spin, and I'd suggest to make it `REPLACE "[-=]"`.

The rationale is that, in some cases, the cached variable ends up as 
```
//Test IS__FVISIBILITY=HIDDEN_SUPPORTED
IS__FVISIBILITY:UNINITIALIZED=HIDDEN_SUPPORTED:INTERNAL=1

//Test IS__WIMPLICIT_FALLTHROUGH=1_SUPPORTED
IS__WIMPLICIT_FALLTHROUGH:UNINITIALIZED=1_SUPPORTED:INTERNAL=
```
which seems to completely confuse CMake on regeneration, so it always does the compile check. In the case of the Xcode generator, that's around 1-2s per test.

tiennou: Also, since AFAIU the point of that replace was to drop the initial `-` from the argument so it can be used as a variable, all `IS_.*_SUPPORTED` now get an extraneous `_` (as my exerpt ⬆️ shows). That's minor though, but you might want to drop the (now duplicated) `_` from the `IS_` strings.

tiennou:
LGTM, thanks for tackling this @lhchavez !

ethomson:
Cool, thanks!

pks-t:
For some background: `-Wall -Werror` does not turn all warnings into errors. I had this at some point in the past, and it doesn't turn "additional" warnings into errors. So e.g. if I had `-Wformat`, then those warnings wouldn't have been converted to an error. This is why I had to be explicit and always say `-Werror=<option>`.

pks-t:
So yeah, I tried my luck to reproduce the old behaviour I was seeing, but right now I can't. I'm not quite happy that we now pass `-Werror` for every single warning (even the disabled ones), buy anyway. Let's see whether I'll accidentally reproduce the remembered behaviour somewhen.

lhchavez:
Hmm that is very odd. I know that most `-Wxx`/`-Wno-xxx` flags depend on the order in which they are added to the commandline, but `-Werror` is a big exception to that so no matter where you put it in the commandline, _any_ not-explicitly-disabled warning is upgraded to an error (unless you also add `-Wno-error=xxx`, which allows a warning to not be upgraded to an error).

maybe there was an ordering issue with `-Wformat`?

PR #4926, opened 2019-01-07
Fix warning 'function': incompatible types - from 'git_cvar_value *' to 'int *' (C4133) on VS

csware:
All other places don't use git_cvar_value as type.

### DIFF
--- a/src/crlf.c
+++ b/src/crlf.c
@@ -35,7 +35,7 @@ struct crlf_attrs {
 
 	int auto_crlf;
 	int safe_crlf;
-	git_cvar_value core_eol;
+	int core_eol;
 };
 
 struct crlf_filter {


carlosmn:
This error message sounds like what MSVC prints when it doesn't know about a type and thus it defaults to `int` (because of some ancient C rule).

ethomson:
Since this is a config cache, sometimes the values that we're caching _are_ actually [`int`](https://github.com/libgit2/libgit2/blob/924f61bfb5a3af13ade5e9963c73af18144ad19e/src/diff_print.c#L51)s and not things that map to enums, so actually using `git_cvar_value` here is not actually appropriate.  Merging this, and I'm leaving the rest as-is.

PR #4930, opened 2019-01-09
Always build a cdecl library

ethomson:
On Windows, we currently offer end-users the ability to compile the library with cdecl calling conventions (the default) or stdcall calling conventions (an option enabled by `-DSTDCALL=ON` to cmake).

The recommendation from engineers within Microsoft is that libraries should have a calling convention specified in the public API, and that calling convention should be cdecl unless there are strong reasons to use a different calling convention.

We offered stdcall for the LibGit2Sharp project, since by default, PInvoke (the .NET FFI) uses stdcall.  This is because PInvoke is usually used to call standard library functions, and Win32 is stdcall for historic reasons.  But the defaults for all new projects is to use cdecl, and most other modern libraries are using it.

(In either case, we _should_ pick one and declare it in our API.  Without having done so, the end user must match our calling conventions in order to even operate.  This is onerous, and we should allow interop to programs from any calling convention.)

This PR removes the `STDCALL` option from cmake, and adds the necessary `cdecl` attributes to `GIT_EXTERN`.  In addition, we need to identify the calling conventions for our callbacks, so I've introduced the `GIT_CALLBACK` macro to do so and updated our public callbacks to have that attribute.

Note that this _is_ a breaking change for anybody who was building with `STDCALL=ON`.  _However_ I suspect that LibGit2Sharp is the only consumer who's done that, since it made their bindings somewhat easier.  But it's not impossible that there's somebody building with stdcall who will need to update their callbacks to be cdecl (one simply needs to add the attribute).

So this is a breaking API change, but I suspect it will not affect too many end users.  But @csware and @implausible are probably the most likely to be affected and I'd like their input.

I have a PR for LibGit2Sharp ready to go to move over to cdecl, to ensure that it will work (it does).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,14 +74,6 @@ IF (APPLE)
 ENDIF()
 
 IF(MSVC)
-	# This option is only available when building with MSVC. By default, libgit2
-	# is build using the cdecl calling convention, which is useful if you're
-	# writing C. However, the CLR and Win32 API both expect stdcall.
-	#
-	# If you are writing a CLR program and want to link to libgit2, you'll want
-	# to turn this on by invoking CMake with the "-DSTDCALL=ON" argument.
-	OPTION(STDCALL			"Build libgit2 with the __stdcall convention"		OFF)
-
 	# This option must match the settings used in your program, in particular if you
 	# are linking statically
 	OPTION(STATIC_CRT		"Link the static CRT libraries"		 		 ON)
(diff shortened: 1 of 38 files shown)


csware:
TortoiseGit always have been using `cdecl`.

ethomson:
> TortoiseGit always have been using cdecl.

Awesome, thanks for the update.  That was my assumption but I appreciate the confirmation.

implausible:
I know that nodegit itself does not specify stdcall; however, it is /possible/ that the node build system (node-gyp) could be defining stdcall. I've been in meetings for 2 days, but I will get back to you after I have checked.

implausible:
We're using `cdecl`.

ethomson:
OK, I've let a few people who are doing FFI and interop on Windows (LibGit2Sharp, etc) and so I think that most people that we know about  haven't had any complaints.  Holding on to this just so that somebody can have an opportunity to argue with me about the `GIT_CALLBACK` macro.

PR #4933, opened 2019-01-11
Some CI enhancements

tiennou:
Since Coverity is down at the moment with no timeframe, isolate it from the nightly run. I think it also makes sense from a concern-separation POV, as it insulates the Azure "hosted" pipelines from flakyness.

This also attempts to enable the few `GITTEST_INVASIVE_FS_SIZE` (though I'm not sure how Azure will react, or if my `run_test` invocation is valid 🤞).

### DIFF
--- a/azure-pipelines/coverity.yml
+++ b/azure-pipelines/coverity.yml
@@ -1,0 +1,21 @@
+resources:
+- repo: self
+
+- job: coverity
+  displayName: 'Coverity'
+  pool:
+    vmImage: 'Ubuntu 16.04'
+  steps:
+  - task: Docker@0
+    displayName: Build
+    inputs:
+      action: 'Run an image'
+      imageName: 'libgit2/trusty-openssl:latest'
+      volumes: |
+       $(Build.SourcesDirectory):/src
+       $(Build.BinariesDirectory):/build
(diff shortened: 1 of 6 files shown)


ethomson:
Hmm, looks like this slows down the CI builds a fair bit (the test phase for one of the amd64 tests went from 6.5 min to about 10 mins).  Maybe we should run the invasive tests in the nightlies and skip them in PR validation and CI builds?

tiennou:
I wasn't really expecting those to be in the main run, but it sure makes testing easier. I'll move them to nightly now that I know they're green.

Note that I didn't enable `GITTEST_INVASIVE_FS_STRUCTURE`, not being sure if I'd wreck something on the builders (I know macOS would not want anything at `/` without `sudo`). Should I try to enable it as well ?

ethomson:
> Note that I didn't enable GITTEST_INVASIVE_FS_STRUCTURE, not being sure if I'd wreck something on the builders (I know macOS would not want anything at / without sudo). Should I try to enable it as well ?


👍  Let's do it!

ethomson:
Thanks @tiennou!  I set this up so that the invasive tests only run nightly.

PR #4934, opened 2019-01-11
Fix crash on remote connection when GIT_PROXY_AUTO is set but no proxy is detected

hackhaslam:
This broke recently. If no proxy is detected, the URL remains unset but it still gets into `gitno_connection_data_from_url` which has `assert(url)`.

### DIFF
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -679,6 +679,9 @@ static int load_proxy_config(http_subtransport *t)
 		    !!t->server.url.use_ssl, &t->proxy_url)) < 0)
 			return error;
 
+		if (!t->proxy_url)
+			return 0;
+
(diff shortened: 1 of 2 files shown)


ethomson:
If there's no proxy detected, we should be returning `GIT_ENOTFOUND`...  🤔

hackhaslam:
Then `http_connect` would have to be taught to ignore `GIT_ENOTFOUND` and continue connecting anyway. It's not obvious to me that `GIT_PROXY_AUTO` not detecting a proxy should result in `GIT_ENOTFOUND`, but I can change it along those lines if you think it makes more sense.

hackhaslam:
Okay, I made that change. I also added the failing test case.

ethomson:
> I also added the failing test case.

💥  Thanks for doing this.

PR #4935, opened 2019-01-13
Use PCRE for our fallback regex engine when regcomp_l is unavailable

ethomson:
This removes the regex implementation that we took from glibc and replaces it with [PCRE](https://www.pcre.org).  PCRE contains a POSIX regex compatibility layer and ignores the currently running locale settings by default, which ensures that `LC_COLLATE` will not alter our functionality.

Now we will always use `regcomp_l` on systems where it is available, and when it is not, we will fallback to our builtin PCRE implementation.  This means that we will never call the system `regcomp` which could be influenced by environment variables.

This adds the older PCRE 8.42 library, instead of the newer PCRE2 library.  The original PCRE has a smaller footprint, which is nice for us, and is still supported and receiving bugfixes.  Minor modifications have been made to remove the CLI components included with PCRE and to prefix the names of the POSIX compatibility functions and structures with `pcre_` so that they do not collide with the system's library (if it has one).

I've also cherry-picked the tests from #4560.


ethomson:
Aaaaaand we've reached that point in a C project where our circular `#include`s have all collapsed in on their weight.

ethomson:
I broke some things out into their own header files.  I'm not exactly thrilled with this, but 🤷‍♂️

csware:
Please consider pcre2 instead of pcre.

ethomson:
> Please consider pcre2 instead of pcre.

I did consider it, as I mentioned in this PR description and commit messages.  What's the benefit to pcre2?

ethomson:
@pks-t since you went deep on understanding the locale problems with regular expressions in #4560, I'd like your input here when you have a moment.

src/posix_regex.h
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+#ifndef INCLUDE_posix_regex_h__
+#define INCLUDE_posix_regex_h__
```
pks-t: Why do we use a separate header for this? Couldn't it simply be part of "posix.h"? I guess it'll get clearer in the following commits

src/posix_regex.h
```diff
@@ -0,0 +1,40 @@
...
+#define P_REG_NOMATCH REG_NOMATCH
+
+#define p_regex_t regex_t
+#define p_regmatch_t regmatch_t
+
+#define p_regerror regerror
+#define p_regexec regexec
+#define p_regfree regfree
+
+#ifdef GIT_USE_REGCOMP_L
```
pks-t: This is broken, right? If you directly included "posix_regex.h", this define might never be set due to the header missing an include of "common.h"->"features.h"

tests/core/posix.c
```diff
@@ -164,8 +164,8 @@ void test_core_posix__p_regcomp_ignores_global_locale_ctype(void)
 		cl_fail("Expected locale to be switched to multibyte");
 	}
 
-	p_regcomp(&preg, "[\xc0-\xff][\x80-\xbf]", REG_EXTENDED);
-	regfree(&preg);
+	p_regcomp(&preg, "[\xc0-\xff][\x80-\xbf]", P_REG_EXTENDED);
```
pks-t: This really should be wrapped in `cl_must_pass`. Not the problem of this patchset, though

deps/pcre/CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
+INCLUDE(CheckIncludeFile)
```
pks-t: Honestly, I'm not much of a fan of including another dependency into our code which we have to keep up-to-date. I'd rather make this a compile-time option, in the following order:

- libc's `regcomp_l`
- system libpcre
- system libpcre2
- libc's `regcomp`, with the known shortcomings
- built-in `regcomp`, with the known shortcomings

src/posix_regex.h
```diff
@@ -7,34 +7,45 @@
...
 	return regcomp_l(preg, pattern, cflags, (locale_t) 0);
 }
+
+#define p_regerror regerror
+#define p_regexec regexec
+#define p_regfree regfree
+
 #else
-# define p_regcomp regcomp
+# include "pcreposix.h"
```
pks-t: I think it might make sense to restructure our compatibility files a bit. Maybe have a "compat" or "posix" directory that contains all platform-specific abstractions, including our "unix" and "win32" directories as well as these new ones.

Edit: oh, this comes directly from the PCRE library. Didn't get that at first

tests/core/posix.c
```diff
@@ -164,7 +164,7 @@ void test_core_posix__p_regcomp_ignores_global_locale_ctype(void)
 		cl_fail("Expected locale to be switched to multibyte");
 	}
 
-	p_regcomp(&preg, "[\xc0-\xff][\x80-\xbf]", P_REG_EXTENDED);
+	error = p_regcomp(&preg, "[\xc0-\xff][\x80-\xbf]", P_REG_EXTENDED);
```
pks-t: Ah, perfect, already got fixed by myself :P

tests/core/posix.c
```diff
@@ -27,6 +31,17 @@ void test_core_posix__initialize(void)
 #endif
 }
 
+void test_core_posix__cleanup(void)
+{
+	int i;
+
+	for (i = 0; i < LC_ALL; i++) {
+		if (old_locales[i])
+		    setlocale(LC_COLLATE, old_locales[i]);
```
pks-t: Wait, this is wrong, isn't it? This should rather be `setlocal(i, old_locales[i])`. Otherwise, we repeatedly set `LC_COLLATE`

tests/core/posix.c
```diff
@@ -172,6 +172,31 @@ void test_core_posix__p_regcomp_ignores_global_locale_ctype(void)
...
+void test_core_posix__p_regcomp_ignores_global_locale_collate(void)
+{
+	regex_t preg;
+	int error = 0;
+
+	const char* oldlocale = setlocale(LC_COLLATE, NULL);
+
+	if (!setlocale(LC_COLLATE, "UTF-8") &&
+	    !setlocale(LC_COLLATE, "c.utf8") &&
+			!setlocale(LC_COLLATE, "en_US.UTF-8"))
```
pks-t: Indentation

pks-t:
> > The last two obviously still have the notable shortcomings, but developers are actually able to work around them by calling setlocale at the start of their program.
> 
> I don't feel like it's reasonable to expect developers to have to change the locale settings in order for compiled-in parts of libgit2 to work, though. I thought that was the whole point of making these changes.

Well, the point is giving developers at least _some_ way of working around the issues, agreed. In some cases, they might just change locales. If that's not possible, they may decide to use this new route of using PCRE as a dependency instead by packaging it along with their application.

> > Otherwise, I'm not a big fan of including the dependency directly.
> 
> We have to include _some_ regex dependency. Today it's code from glibc that always misbehaves in some situations. If we change that dependency to be PCRE instead, we can potentially fix that issue.

That we need to use some regex dependency is obvious. My point is only whether we want to bundle the dependency into our own codebase or make use of a system-provided PCRE dependency. I'm more inclined to use a system-provided one instead of bundling it with us, as otherwise we are responsible to notice and update whenever important bugfixes happen in bundled dependencies.

> Note that PCRE (v1, not PCRE2) is also [git's fallback](https://www.spinics.net/lists/git/msg300072.html). So even if there are slight incompatibilities between GNU and PCRE, they're the same incompatibilities that git has, meaning paradoxically, that PCRE's incompatibilities keep _us_ more compatible with git.

Ah, good to know. Indeed, git recently gained the ability to also use PCRE2 instead of PCRE1.

ethomson:
> My point is only whether we want to bundle the dependency into our own codebase or make use of a system-provided PCRE dependency.

Hmm, okay.  So what you're saying is ship the current `regcomp` dependency in our codebase, not ship PCRE as a dependency in our codebase _instead_.

You gave this list of priorities:

> libc's regcomp_l
> system libpcre
> system libpcre2
> libc's regcomp, with the known shortcomings
> built-in regcomp, with the known shortcomings

It sounds like you're saying that you want to keep shipping the glibc regcomp.  I guess I don't understand why shipping one dependency (the current glibc `regcomp`) is superior to shipping a different dependency (PCRE).  We have the same problem where we are responsible for noticing and updating when important bugfixes happen, but there's no opportunity for us to fix #4528 on systems where we use our built-in regcomp (eg, Windows).

pks-t:
On Fri, Jan 25, 2019 at 09:19:05AM -0800, Edward Thomson wrote:
> > libc's regcomp_l
> > system libpcre
> > system libpcre2
> > libc's regcomp, with the known shortcomings
> > built-in regcomp, with the known shortcomings
> 
> It sounds like you're saying that you want to keep shipping the
> glibc regcomp.  I guess I don't understand why shipping one
> dependency (the current glibc `regcomp`) is superior to
> shipping a different dependency (PCRE).  We have the same
> problem where we are responsible for noticing and updating when
> important bugfixes happen, but there's no opportunity for us to
> fix #4528 on systems where we use our built-in regcomp (eg,
> Windows).

This is indeed a very good point. I mean we could say "Let's drop
regex and don't include PCRE" to make them kind of equal, but
this would make the setup on Windows a lot harder.

Still, my point kind of still applies: it sucks to have bundled
dependencies. But I'd be happy if we simply allowed users of
libgit2 to choose between system system `regcomp_l`, system
`regcomp`, system PCRE or builtin PCRE. So two things I'd like to
have:

- We should have a compile-time switch that allows users to
  explicitly choose which backend to use. If no preference is
  given, then we should pick: system `regcomp_l`, system PCRE,
  builtin PCRE, in this order. Plain `regcomp` should not
  automatically be chosen due to the reported problems, but it
  might be useful to allow users to explicitly request it.

- The ability to choose between either system-provided libpcre or
  builtin libpcre. On Linux-based distros, I expect that
  distributors would rather like to link against libpcre than
  using the bundled one.

ethomson:
> - We should have a compile-time switch that allows users to
  explicitly choose which backend to use. If no preference is
  given, then we should pick: system `regcomp_l`, system PCRE,
  builtin PCRE, in this order. Plain `regcomp` should not
  automatically be chosen due to the reported problems, but it
  might be useful to allow users to explicitly request it.
>
> - The ability to choose between either system-provided libpcre or
  builtin libpcre. On Linux-based distros, I expect that
  distributors would rather like to link against libpcre than
  using the bundled one.

This seems like a fine plan to me.

pks-t:
You need any help with regards to my proposal? Happy to help, as it's essentially my request :)

ethomson:
OK, I updated this to take a `-DREGEX=...` option during the cmake.  Options are `regcomp_l` (for the system `regcomp_l`), `pcre` (for a system PCRE)`, `regcomp` (for the system `regcomp`) or `builtin` (for the bundled PCRE).

This will also prefer system `regcomp_l`, then system PCRE, and finally the builtin PCRE.  If users want `regcomp` they will need to enable it themselves.

tests/core/posix.c
```diff
@@ -15,11 +15,15 @@
 #include "posix.h"
 #include "userdiff.h"
 
+#if LC_ALL > 0
```
pks-t: This is a bit weird. Does Windows define `LC_ALL` to `0` or doesn't it define it at all? If the latter's the case, then we should definitely use `#ifndef` instead. I'll also upload a PR that enables -Wundef in the short-term future to generate a warning for such uses

src/CMakeLists.txt
```diff
@@ -294,15 +289,39 @@ ELSE()
...
+	CHECK_SYMBOL_EXISTS(regcomp_l "regex.h;xlocale.h" HAVE_REGCOMP_L)
+	CHECK_SYMBOL_EXISTS(pcre_regcomp "pcreposix.h" HAVE_PCRE)
+
+	IF (HAVE_REGCOMP_L)
+		SET(REGEX "regcomp_l")
+	ELSEIF (HAVE_PCRE)
+		SET(REGEX "pcre")
+	ELSE ()
+		SET(REGEX "builtin")
+	ENDIF ()
```
pks-t: We have said in the past that there shall be no space between keyword and opening brace in conditionals

src/CMakeLists.txt
```diff
@@ -294,15 +289,39 @@ ELSE()
...
-ELSE()
+	SET(GIT_REGEX_REGCOMP_L 1)
+ELSEIF(REGEX STREQUAL "pcre")
+	ADD_FEATURE_INFO(regex ON "using system PCRE")
+	SET(GIT_REGEX_PCRE 1)
+ELSEIF(REGEX STREQUAL "regcomp")
+	ADD_FEATURE_INFO(regex ON "using system regcomp")
+	SET(GIT_REGEX_REGCOMP 1)
+ELSEIF(REGEX STREQUAL "builtin")
+	ADD_FEATURE_INFO(regex ON "using bundled PCRE")
```
pks-t: This invocation could be repeated if we instead set a variable. A matter of taste though, please do as you like

src/CMakeLists.txt
```diff
@@ -294,15 +289,39 @@ ELSE()
...
+	ENDIF ()
+ENDIF()
+
+IF(REGEX STREQUAL "regcomp_l")
 	ADD_FEATURE_INFO(regex ON "using system regcomp_l")
-ELSE()
+	SET(GIT_REGEX_REGCOMP_L 1)
+ELSEIF(REGEX STREQUAL "pcre")
+	ADD_FEATURE_INFO(regex ON "using system PCRE")
+	SET(GIT_REGEX_PCRE 1)
```
pks-t: Shouldn't we detect here whether system-pcre exists and set up includes/libraries? Or did I miss a past commit that already sets this up for us?

ethomson: I added this - we will now look for PCRE and, if it's installed, prefer it to our own PCRE routines.  

A user can also opt-in to using PCRE2 if it is installed, but we do _not_ autodetect PCRE2.  The POSIX compatibility layer for PCRE2 is broken on Debian-based installs (the library symbols are prefixed with `PCRE2`, eg, `PCRE2regcomp`, but the headers make no mention of this.  They were supposed to set up a layer of indirection, eg `#define regcomp PCRE2regcomp` but this was neglected.  This is being fixed but we cannot rely on it yet.)

tests/core/posix.c
ethomson: Windows defines `LC_ALL` as `0`.

tests/core/posix.c
ethomson: Fixed.

ethomson:
OK, updated this and rebased it.

src/posix_regex.h
pks-t: This has been fixed :+1:

CMakeLists.txt
```diff
@@ -65,6 +65,7 @@ OPTION(DEBUG_POOL			"Enable debug pool allocator"				OFF)
 OPTION(ENABLE_WERROR			"Enable compilation with -Werror"			OFF)
 OPTION(USE_BUNDLED_ZLIB    		"Use the bundled version of zlib"			OFF)
 OPTION(DEPRECATE_HARD			"Do not include deprecated functions in the library"	OFF)
+   SET(REGEX				"" CACHE STRING "Regular expression implementation. One of regcomp_l, pcre, regcomp, or builtin.")
```
pks-t: This variable name may be a bit problematic, as it is a keyword in CMake. We might want to rename it. "REGEX_BACKEND" would come to my mind, as there's precedent for this already, even though I seem to recall that you don't like the name "backend"

src/posix_regex.h
```diff
@@ -15,7 +15,7 @@
  * compatible implementation.
  */
 
-#ifdef GIT_REGEX_PCRE
+#ifdef GIT_REGEX_BUILTIN
```
pks-t: Huh? Where do we define `p_regex_` stuff for `GIT_REGEX_PCRE` now?

Edit: never mind, the following commit clears that up

src/CMakeLists.txt
```diff
@@ -307,6 +307,18 @@ ENDIF()
 IF(REGEX STREQUAL "regcomp_l")
 	ADD_FEATURE_INFO(regex ON "using system regcomp_l")
 	SET(GIT_REGEX_REGCOMP_L 1)
+ELSEIF(REGEX STREQUAL "pcre2")
```
pks-t: Any reason why we shouldn't auto-detect the PCRE2 library `IF(REGEX STREQUAL "")`?

CMakeLists.txt
ethomson: I'm fine with regex backend (my avoidance of "backend" comes from using it without qualification.  That is I'm fine with having a _regex backend_ but to talk about a "backend" as something concrete sounds odd to me.)

src/CMakeLists.txt
ethomson: Hmm, I tried to mention this in a comment, but it looks like it got buried.  (It is mentioned in the commit that introduces PCRE2.)

Basically, the POSIX compatibility layer for PCRE2 is completely broken on Ubuntu and Debian.  If you look at PCRE, you'll see that it has symbols of pcre_regcomp, and then `pcreposix.h` sets up that indirection as `#define regcomp pcre_regcomp`.  This avoids conflicting with the regex functions that are in libc.

The PCRE2 packages only do half of this, they have symbols for `PCRE2regcomp` but they forgot to set up the indirection in the header files.  (They just declare `regcomp`.)  So you end up using the regular libc regex bits.  Worse, you use the pcre2 header files and their `#define`s for constants, so you end up passing the wrong constants to libc, and things go very wrong.

So we allow people to set PCRE2 manually, in case they have a working installation, but we don't autodetect it to avoid breaking Debian/Ubuntu users.

ethomson:
I'll update `REGEX` in the cmake to `REGEX_BACKEND`.  Are you happy with me 🚢 this once I do?

src/CMakeLists.txt
pks-t: Hm. Couldn't we just use CMake to detect whether the system PCRE2 is broken or not with `CHECK_FUNCTION_EXISTS`?

pks-t:
I'm still a bit hesitant with regards to the PCRE2 bits, but this is really nothing that should hold back this PR as we do not loose any functionality at all. So: go for it!

src/CMakeLists.txt
ethomson: Hmm, maybe yes, but since they haven't fixed it yet, I don't even have a test case that I can verify that we're getting it correct (yet).

My preference would be to hold off until a (shipping) version of Debian/Ubuntu has this fixed and we can test it...

ethomson:
Cool.  I want to get this merged so that we can stomp on the warnings in Windows (a lot of them come from the existing regex implementation).  We can iterate on the PCRE2 support.

src/CMakeLists.txt
pks-t: Fair enough. By holding off you probably just refer to the proper detection of non-broken PCRE2, right?

ethomson: Yeah, keep it as-is where people can explicitly opt-in to PCRE2 support but not enable PCRE2 by default (yet).

ethomson:
Updated to use `REGEX_BACKEND` as the option.

pks-t:
By the way, just came to my mind: should we update Azure for this? Like, having at least one build explicitly using bundled PCRE and `regcomp_l`, respectively?

tiennou:
(relatedly, macOS also has both the POSIX2008.smth `regcomp_l` and pcre1/2, so it might be somewhat easier to test the build system)

ethomson:
OK, I've added PCRE to the trusty docker image, and set some explicit regex configuration during builds.  (One Linux build explicitly sets PCRE, one explicitly sets builtin, and now the macOS built explicitly sets `regcomp_l`.)

pks-t:
> OK, I've added PCRE to the trusty docker image, and set some explicit regex configuration during builds. (One Linux build explicitly sets PCRE, one explicitly sets builtin, and now the macOS built explicitly sets regcomp_l.)

Perfect, thanks! No more remarks from my side now, I promise.

ethomson:
> Perfect, thanks! No more remarks from my side now, I promise.

In that case, 🚢 .  Thanks for the review everyone!

src/CMakeLists.txt
orivej: For the record, pcre2 has added compatibility aliases (`#define regexec  pcre2_regexec`) in 10.33 (in [r1064](https://vcs.pcre.org/pcre2?view=revision&revision=1064)).

PR #4936, opened 2019-01-14
Get rid of some test files that were accidentally committed

implausible:
It seems these files were committed by accident.


ethomson:
Sigh, thanks.  This was my bad.  😢

implausible:
Np. This will just make the rest of my day a little easier. Those files are killing me :laughing:

ethomson:
For extra context, these were accidentally checked in, they're the [temp directories](https://github.com/libgit2/libgit2/blob/master/tests/resources/generate_crlf.sh#L86-L90) from the creation script.  😢 

Our guidance is to actually stuff things into an _actual_ temp directory to avoid this, but apparently I erred.  In any case, they're now removed.  Thanks again @implausible

PR #4937, opened 2019-01-15
Convert tests/resources/push.sh to LF endings

lhchavez:
This changes that file to use UNIX line-endings, which makes sense since
this is a UNIXy file.


ethomson:
lol why does this even have crlf?

Issue #4946, opened 2019-01-20
git_odb_foreach continues looping after callback returns > 0


zlikavac32:
Hi, 

in [include/git2/odb.h:246](https://github.com/libgit2/libgit2/blob/22d2062d954dcb88fa3dc65281d3f3d88ae87d68/include/git2/odb.h#L246) it states `Return a non-zero value from the callback to stop looping.`

Implementation in [odb.c:1263](https://github.com/libgit2/libgit2/blob/22d2062d954dcb88fa3dc65281d3f3d88ae87d68/src/odb.c#L1263) considers only values less than 0 as a condition to finish looping.

So I'm not sure whether this is a bug or not. It's true that public API allows any non zero value, but perhaps implementation makes more sense.

### Reproduction steps

Depends. If this is considered a bug, than any `return > 0` from callback.

### Expected behavior

Depends. If this is considered a bug, then looping should exit for `return > 0`.

### Actual behavior

When callback returns value greater than 0, looping still continues.

### Version of libgit2 (release number or SHA1)

Current `master` (- `22d2062d954dcb88fa3dc65281d3f3d88ae87d68`) and from when, don't know :).

### Operating system(s) tested

Arch, but it doesn't matter.

Thank you for this cool library.

ethomson:
Thanks for reporting this - indeed we should be propagating positive return codes from the callback.

PR #4949, opened 2019-01-20
odb: Fix odb foreach to also close on positive error code

zlikavac32:
Hi,

This should be fix for #4946 .

I'm not familiar with internals, but all tests pass and I did add one new for this case. I don't like that copy/paste, so I was thinking about using struct (or 2 int array) so that only one callback exists in test. What do you think about that?

Also, should anything else be changed?

Thank you.

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -1260,7 +1260,7 @@ int git_odb_foreach(git_odb *db, git_odb_foreach_cb cb, void *payload)
 	git_vector_foreach(&db->backends, i, internal) {
 		git_odb_backend *b = internal->backend;
 		int error = b->foreach(b, cb, payload);
-		if (error < 0)
+		if (error != 0)
 			return error;
 	}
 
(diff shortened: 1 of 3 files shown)


ethomson:
Hiya - thanks for the PR.  This looks good to me, I'm fine with a little copypasta in our tests, I don't think there's a big need to use a `struct` instead.

You put `WIP` in the title - I'm happy to merge this, but wanted to make sure you were :+1: before I did.

zlikavac32:
@ethomson I'm ok, If you're ok :). WIP was more because this is my first PR here, so perhaps I'm missing something out. I'll remove it.

Thanks

ethomson:
Also:

> this is my first PR here

Welcome to the libgit2 project.  😀

zlikavac32:
@ethomson No problem, thank you for your quick response. Now I can continue with your master :D

> Welcome to the libgit2 project.

Thank you, I like this lib. No doubt I'll contribute again when I encounter something :)

PR #4950, opened 2019-01-20
Clean up some warnings

ethomson:
This cleans up some warnings encountered building on Win32 with MSVC.  This PR attacks all the x86 warnings.  (Some remain on amd64.)

### DIFF
--- a/src/blob.c
+++ b/src/blob.c
@@ -36,10 +36,10 @@ git_off_t git_blob_rawsize(const git_blob *blob)
 
 int git_blob__getbuf(git_buf *buffer, git_blob *blob)
 {
-	return git_buf_set(
-		buffer,
-		git_blob_rawcontent(blob),
-		git_blob_rawsize(blob));
+	git_off_t size = git_blob_rawsize(blob);
+
+	GIT_ERROR_CHECK_BLOBSIZE(size);
+	return git_buf_set(buffer, git_blob_rawcontent(blob), (size_t)size);
 }
 
 void git_blob__free(void *_blob)
@@ -389,12 +389,14 @@ cleanup:
 int git_blob_is_binary(const git_blob *blob)
 {
 	git_buf content = GIT_BUF_INIT;
+	git_off_t size;
 
 	assert(blob);
 
+	size = git_blob_rawsize(blob);
+
 	git_buf_attach_notowned(&content, git_blob_rawcontent(blob),
-		min(git_blob_rawsize(blob),
-		GIT_FILTER_BYTES_TO_CHECK_NUL));
+		(size_t)min(size, GIT_FILTER_BYTES_TO_CHECK_NUL));
 	return git_buf_text_is_binary(&content);
 }
 
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -440,7 +440,7 @@ int git_buf_decode_base85(
 
 		acc += de;
 
-		cnt = (output_len < 4) ? output_len : 4;
+		cnt = (output_len < 4) ? (int)output_len : 4;
 		output_len -= cnt;
 		do {
 			acc = (acc << 8) | (acc >> 24);
--- a/src/delta.c
+++ b/src/delta.c
@@ -138,7 +138,7 @@ static int lookup_index_alloc(
 	*out = git__malloc(index_len);
 	GIT_ERROR_CHECK_ALLOC(*out);
 
-	*out_len = index_len;
+	*out_len = (unsigned long)index_len;
 	return 0;
 }
 
@@ -286,6 +286,13 @@ int git_delta_create_from_index(
 	if (!trg_buf || !trg_size)
 		return 0;
 
+	if (index->src_size > UINT_MAX ||
+	    trg_size > UINT_MAX ||
+	    max_size > (UINT_MAX - MAX_OP_SIZE - 1)) {
+		git_error_set(GIT_ERROR_INVALID, "buffer sizes too large for delta processing");
+		return -1;
+	}
+
 	bufpos = 0;
 	bufsize = 8192;
 	if (max_size && bufsize >= max_size)
@@ -294,7 +301,7 @@ int git_delta_create_from_index(
 	GIT_ERROR_CHECK_ALLOC(buf);
 
 	/* store reference buffer size */
-	i = index->src_size;
+	i = (unsigned int)index->src_size;
 	while (i >= 0x80) {
 		buf[bufpos++] = i | 0x80;
 		i >>= 7;
@@ -302,7 +309,7 @@ int git_delta_create_from_index(
 	buf[bufpos++] = i;
 
 	/* store target buffer size */
-	i = trg_size;
+	i = (unsigned int)trg_size;
 	while (i >= 0x80) {
 		buf[bufpos++] = i | 0x80;
 		i >>= 7;
@@ -423,7 +430,7 @@ int git_delta_create_from_index(
 			void *tmp = buf;
 			bufsize = bufsize * 3 / 2;
 			if (max_size && bufsize >= max_size)
-				bufsize = max_size + MAX_OP_SIZE + 1;
+				bufsize = (unsigned int)(max_size + MAX_OP_SIZE + 1);
 			if (max_size && bufpos > max_size)
 				break;
 			buf = git__realloc(buf, bufsize);
--- a/src/describe.c
+++ b/src/describe.c
@@ -366,7 +366,7 @@ static int find_unique_abbrev_size(
 	int *out,
 	git_repository *repo,
 	const git_oid *oid_in,
-	int abbreviated_size)
+	unsigned int abbreviated_size)
 {
 	size_t size = abbreviated_size;
 	git_odb *odb;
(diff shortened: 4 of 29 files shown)


src/notes.c
```diff
@@ -329,7 +331,10 @@ static int note_new(
 		git_signature_dup(&note->committer, git_commit_committer(commit)) < 0)
 		return -1;
 
-	note->message = git__strndup(git_blob_rawcontent(blob), git_blob_rawsize(blob));
+	blobsize = git_blob_rawsize(blob);
+	GITERR_CHECK_BLOBSIZE(blobsize);
+
+	note->message = git__strndup(git_blob_rawcontent(blob), (size_t)blobsize);
```
pks-t: By the way, why do we even have this split between `git_off_t` and `size_t` when representing blobs? Shouldn't we uniformly use the same type everywhere in the first place? `git_blob` is an internal structure, too, so we are free to change it to use `size_t`.

tests/core/zstream.c
```diff
@@ -68,7 +68,7 @@ void test_core_zstream__fails_on_trailing_garbage(void)
 
 	/* append some garbage */
 	for (i = 0; i < 10; i++) {
-		git_buf_putc(&deflated, i);
+		git_buf_putc(&deflated, (char)i);
```
pks-t: Nit: you could also just make `i` a `char`. Not that it'd matter much

src/buffer.c
```diff
@@ -440,7 +440,7 @@ int git_buf_decode_base85(
 
 		acc += de;
 
-		cnt = (output_len < 4) ? output_len : 4;
+		cnt = (output_len < 4) ? (int)output_len : 4;
```
pks-t: Have we really tested that `output_len` fits into an `int`, though? For all I can see, we have

- verified that `output_len` is greater than `base85_len * 4/5`, which doesn't really limit the maximum value of `output_len`
- verified that `output_len + buf->size` does not overflow `size_t`

So `output_len` _could_ in theory still be bigger than `int`, right?

src/streams/mbedtls.c
```diff
@@ -303,22 +303,22 @@ static int mbedtls_set_proxy(git_stream *stream, const git_proxy_options *proxy_
...
 		if (error <= 0) {
 			return ssl_set_error(st->ssl, error);
 		}
-		read += error;
-	} while (read < len);
+		written += error;
+	} while (written < len);
 
-	return read;
+	return written;
```
pks-t: So instead of writing all bytes, we just write the first `SSIZET_MAX` bytes. This unfortunately breaks callers writing more than `SSIZET_MAX` bytes, as they never check for partially written buffers but for errors, only. E.g.:

```
static int write_chunk(git_stream *io, const char *buffer, size_t len)
{
	git_buf buf = GIT_BUF_INIT;

	/* Chunk header */
	git_buf_printf(&buf, "%" PRIxZ "\r\n", len);

	if (git_buf_oom(&buf))
		return -1;

	if (git_stream_write(io, buf.ptr, buf.size, 0) < 0) {
		git_buf_dispose(&buf);
		return -1;
	}

	git_buf_dispose(&buf);

	/* Chunk body */
	if (len > 0 && git_stream_write(io, buffer, len, 0) < 0)
		return -1;

	/* Chunk footer */
	if (git_stream_write(io, "\r\n", 2, 0) < 0)
		return -1;

	return 0;
}
```

src/transports/git.c
```diff
@@ -75,14 +75,17 @@ static int gen_proto(git_buf *request, const char *cmd, const char *url)
...
+	size_t write_size;
+	int error;
 
 	error = gen_proto(&request, s->cmd, s->url);
 	if (error < 0)
 		goto cleanup;
 
-	error = git_stream_write(s->io, request.ptr, request.size, 0);
+	write_size = min(request.size, INT_MAX);
+	error = (int)git_stream_write(s->io, request.ptr, write_size, 0);
```
pks-t: Same here, right? We silently drop everything that's after the first `INT_MAX` bytes.

src/delta.c
```diff
@@ -286,7 +286,9 @@ int git_delta_create_from_index(
 	if (!trg_buf || !trg_size)
 		return 0;
 
-	if (index->src_size > UINT_MAX || trg_size > UINT_MAX) {
+	if (index->src_size > UINT_MAX ||
+	    trg_size > UINT_MAX ||
+	    max_size > (UINT_MAX - MAX_OP_SIZE - 1)) {
```
pks-t: This hunk belongs to the previous commit, right?

src/describe.c
```diff
@@ -366,7 +366,7 @@ static int find_unique_abbrev_size(
 	int *out,
 	git_repository *repo,
 	const git_oid *oid_in,
-	int abbreviated_size)
+	unsigned int abbreviated_size)
 {
 	size_t size = abbreviated_size;
```
pks-t: One might wonder whether this is the right thing to do. Previously, passing `-1` was underflowing to a huge `size`, essentially saying "Do not abbreviate but print all digits". But the only callsite `show_suffix` gets called with `git_describe_format_options.abbreviated_size`, which is an `unsigned int` itself.

src/diff_generate.c
```diff
@@ -564,14 +564,14 @@ int git_diff__oid_for_file(
 {
 	git_index_entry entry;
 
-	if (!git__is_sizet(size)) {
+	if (size < 0 || size > UINT32_MAX) {
```
pks-t: This can be squashed into the previous commit where you introduce the `git__is_sizet` check.

src/filter.c
```diff
@@ -974,7 +974,7 @@ int git_filter_list_stream_file(
 	}
 
 	if (readlen < 0)
-		error = readlen;
+		error = (readlen < INT_MIN) ? -1 : (int)readlen;
```
pks-t: POSIX specifies that `read` can only ever return `-1` or a value `>=0`. So you could just say:

```
if (readlen < 0)
    error = -1;
```

ethomson: Good call.  I had done this before I re-read the spec and been reminded that `ssize_t` is actually only guaranteed to be in the range of `(-1, SSIZE_MAX)`.  🤯

src/streams/mbedtls.c
ethomson: Yup.  I have a TODO item to fix these up also.  I could raise this as an issue or something, but it needs a serious audit.  It looks like we have several places where we assume that `write` (maybe `read`, too) will always write everything we ask it to.

I was thinking about swapping in `p_read` with a function that works byte-at-a-time in our tests to flush out all offenders.

But back to this:  if we ever did write all bytes (where the buffer size was greater than `SSIZE_MAX`) then we'd wrap and end up returning an error.  Probably the current behavior is slightly better (thinking we failed when we didn't, rather than writing half the bytes and thinking we succeeded)...  but we're rather hosed either way.

In any case, I'd be willing to go through and make sure that we always read/write in a loop before merging this or after.  The risk of saying "oh we'll do it later" is, of course, that we never actually do it.

src/transports/git.c
ethomson: Yep.

tests/core/zstream.c
ethomson: Yeah, idk why I didn't do that.  This looks dumb.

src/blob.c
```diff
@@ -389,12 +389,14 @@ int git_blob_create_fromstream_commit(git_oid *out, git_writestream *_stream)
 int git_blob_is_binary(const git_blob *blob)
 {
 	git_buf content = GIT_BUF_INIT;
+	git_off_t size;
 
 	assert(blob);
 
+	size = git_blob_rawsize(blob);
```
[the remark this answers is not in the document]
pks-t: Oh, sure, that's why!

src/notes.c
pks-t: I'm always in favor of being explicit, so I'd be perfectly happy with e.g. `uint64_t`.

src/streams/mbedtls.c
pks-t: Just let me know if you've got the time. Otherwise I can also rip this out and handle it myself, as I should have some time to do so this week

ethomson: I likely won't be able to get to this until the weekend so you're welcome to if you'd like.

pks-t: Both `p_read` and `p_write` are fine, as we already loop around read(3) and write(3)

pks-t: See #4953

src/delta.c
ethomson: Fixed!

src/diff_generate.c
ethomson: Sure.

src/filter.c
ethomson: Fixed.

tests/core/zstream.c
ethomson: Fixed.

src/notes.c
ethomson: I agree, this is noted in #4960

ethomson:
On second thought, it's only the `SSIZE_MAX`, so that's a pretty trivial conflict.  I want to look at the amd64 warnings, so 🚢

PR #4952, opened 2019-01-23
Deprecate functions and constants more gently

ethomson:
Provide a new kindler, gentler deprecation policy for the upcoming release.

1. Move deprecated bits into their own header file `deprecated.h`.
2. Remove the `GIT_DEPRECATED` attribute from the deprecated bits.  This will allow users to more gently move into the new functions at their leisure.  We have no hurry in removing the old functions or values, so let's not rush things.
3. Allow users to opt-in to a harder deprecation, by defining `GIT_DEPRECATE_HARD` to completely remove the deprecated functions.
4. Enable `GIT_DEPRECATE_HARD` ourselves to ensure that we don't use them in the library code, tests, or examples.

I've kept the `GIT_DEPRECATED` macro so that we can mark something as properly deprecated, using the compiler attributes, in the future if we desire.


ethomson:
OK, here's a proposal for a "softer" deprecation; instead of marking things as deprecated with an attribute, we simply stuff them into their own file.  The exception being `sys/streams.h`; since `sys` header files aren't included by default (from `git2.h`), we want to keep the deprecated functions there.

This introduces `GIT_DEPRECATE_HARD`, which you can define if you want to avoid deprecated bits entirely.

src/common.h
```diff
@@ -76,6 +76,7 @@
 
 #include "git2/types.h"
 #include "git2/errors.h"
+#include "git2/deprecated.h"
```
pks-t: This include shouldn't be necessary in the end, right? None of our internal code is supposed to use deprecated functions or constants. With the exception of test code, I guess

ethomson: Yeah, good call.

PR #4953, opened 2019-01-23
stream: fix potentially truncated writes

pks-t:
This builds on parts of #4950 and extends it to fix the potential for partial writes in our `stream` implementation.


src/stream.h
```diff
@@ -55,6 +55,21 @@ GIT_INLINE(ssize_t) git_stream_write(git_stream *st, const char *data, size_t le
 	return st->write(st, data, len, flags);
 }
 
+GIT_INLINE(int) git_stream__write_full(git_stream *st, const char *data, size_t len, int flags)
```
ethomson: Two questions:

1. Should we not just make `git_stream_write` always loop and write everything it can?  I'm not sure I see the value of having both functions.
2. I worry a bit that we should guard against the underlying write returning `0`.  I trust something like `write(2)` to not return `0` when in blocking mode, but we've got a few different implementations and things like the stransport implementation are apparently already expecting us to work in a way that they don't document (from reading that comment in its write function).  I wouldn't want to get into a place where we get in an infinite loop when write returns `0` for some reason we didn't expect.  Maybe convert it to an internal error or assert?

src/streams/socket.c
```diff
@@ -130,10 +130,9 @@ int socket_connect(git_stream *stream)
 	return 0;
 }
 
-ssize_t socket_write(git_stream *stream, const char *data, size_t len, int flags)
+ssize_t socket_write(git_stream *stream, const char *data, size_t data_len, int flags)
 {
-	ssize_t ret;
-	size_t off = 0;
+	ssize_t ret, off = 0, len = min(data_len, SSIZE_MAX);
```
neithernut: Apparently `SSIZE_MAX` is not defined on Windows. At least that's what the CI says.

pks-t: True, it's only getting introduce by @ethomson in #4950

src/stream.h
pks-t: 1. Thing is, `git_stream_write` is used as a primitive in e.g. our mbedtls and openssl streams, where it implements the BIO callbacks. Looping is handled by mbedtls and openssl, so I wasn't quite sure whether we really want to loop ourselves again. Also, we cannot really change `git_stream_write` to loop itself, as it gets passed an `size_t` length but returns `ssize_t`. Thus, we either loop and return a wrong value, have to raise an error if `len > SSIZE_MAX` or we need to break the API.

2. Yeah, we probably should. I've made it return an error now

src/streams/socket.c
pks-t: I've cherry-picked that particular commit

src/stream.h
ethomson: Ah, yes.  I remember thinking about this however briefly before throwing up my hands and realizing that I wasn't going to fix anything quickly, so I just went to bed.  I agree with this solution for #1.

src/streams/mbedtls.c
```diff
@@ -303,25 +301,27 @@ static int mbedtls_set_proxy(git_stream *stream, const git_proxy_options *proxy_
...
-	} while (read < len);
+	/*
+	 * `mbedtls_ssl_write` can only represent INT_MAX bytes
+	 * written via its return value. We thus need to clamp
+	 * the maximum number of bytes written.
+	 */
+	len = min(len, INT_MAX);
+
+	if ((written = mbedtls_ssl_write(st->ssl, (const unsigned char *)data, len)) <= 0)
+		return ssl_set_error(st->ssl, error);
```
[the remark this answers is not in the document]
pks-t: Uhm. Is your point the `<=` comparison or the confusion between `error` and `written`? The second one is obviously wrong, I need to pass in `written`. I'm not quite sure about the comparison, though, and so I'd tend to stay with the existing `<=` comparison

ethomson:
I went ahead and merged #4950.  I wasn't expecting that to cause so many conflicts; apologies.

ethomson:
I fixed the mbedTLS issue, rebased this to fix the conflicts and merged this manually.

ethomson:
Thanks for tackling this!  🎉

pks-t:
Thanks for merging, @ethomson!

Issue #4783, opened 2018-08-26
Better document `GIT_EUSER` usage in callbacks


ethomson:
Per #4703, we do not make it entirely clear how we intend people to stop a callback.  Our current callback policy is that when a callback returns a negative value, the callback will stop and return the given value back to the caller.  We have reserved the `GIT_EUSER` code so that callbacks can indicate (to themselves) that they have stopped the callback, and that a libgit2 error did not occur.

We should document each callback, indicating to users that this special value exists for them.

lrm29:
I just had to look this up just now, good to see an issue on it! Even though it's mentioned in error_handling.md, that's not immediately visible when you're looking through the really excellent API reference on the libgit2 website.

git_reference_foreach_glob is the only callback that documents GIT_EUSER:
 * @return 0 on success, GIT_EUSER to cancel the operation, or error code.

With doxygen, it would be possible to specify an alias, so that you'd be able to to just type @callbackreturn in the documentation of callbacks:

ALIASES += callbackreturn{1}="0 on success, GIT_EUSER to cancel the operation, or error code.  \1" 

But I see the documentation is generated using decorium, not sure if that can handle aliases like that? I feel that might be a better solution than copying and pasting text into all the callback comments.

pks-t:
> But I see the documentation is generated using decorium, not sure if that can handle aliases like that? I feel that might be a better solution than copying and pasting text into all the callback comments.

@tiennou, you've done some work on docurium. Do you know if that's possible? I'd assume it's not.

tiennou:
It's not, but it might be easy to add. A new `aliases` section in the config, a pre-process step before markdown-ification. I'd personally have gone for taking a `files` section (set to, say, "docs/"), build an index of "header slugs" from Markdown files in `files`, and use that index to autolink `@see` remarks when generating the documentation. Sounds more generally useful because you get `@see docs/THREADING.md`, or `@see docs/USAGE.md#callbacks`, but it's slightly harder to implement 😉.

tiennou:
There's something else I'd like to have in that vein: load files under docs and provide `@see` anchors for them, so you could `@see docs/guidelines.md#callback-usage`, which would be usable in comment form as well…

navitux:
Hi all,Is this issue still open ?
I'd like to contribute

Issue #4711, opened 2018-06-30
add.c example doesn't display


tiennou:
While working on #4692, the first example I looked up was [add](https://libgit2.org/ex/HEAD/add.html).

This doesn't display at all in the docurium GUI. I suspect it's because there aren't any comments in it but that needs more investigation.

PR #4954, opened 2019-01-23
Documentation fixes

### LINKED ISSUE #4908: document regarding git_time_t should be clear about its unit
### Reproduction steps

This is a documentation bug. Wherever the type git_time_t is mentioned in the document or in the example, there is no mentioning of its unit. 

Is this the milliseconds since the Epoch or the seconds since the Epoch?

https://github.com/libgit2/libgit2/commit/b5c00c6d41ff4eb809f7deda38fe61ccaacb7cbf

https://libgit2.org/libgit2/#HEAD/group/signature/git_signature_new

Similarly in examples, large, incomprehensible integers are used for git_time_t, hard to understand what they mean.

https://libgit2.org/libgit2/ex/HEAD/general.html#git_signature_new-80

  git_signature_new(&author,
      "Scott Chacon", "schacon@gmail.com", 123456789, 60);
  git_signature_new(&committer,
      "Scott A Chacon", "scott@github.com", 987654321, 90);

### Expected behavior

The document should be clear about the type's meaning. the current description "time when the action happened" isn't a very useful guide on using this api.

b5c00c6d41ff4eb809f7deda38fe61ccaacb7cbf
n/a

tiennou:
This should close #4908, attempts at #4783 and #4711, and some API comment touch-ups.


include/git2/types.h
```diff
@@ -59,7 +59,7 @@ typedef __haiku_std_int64 git_time_t;
  * app, even though /we/ define _FILE_OFFSET_BITS=64.
  */
 typedef int64_t git_off_t;
-typedef int64_t git_time_t;
+typedef int64_t git_time_t; /**< time in seconds from epoch */
```
pks-t: Didn't we have some problems correctly rendering those inline comments? Not quite sure whether I remember correctly

docs/error-handling.md
```diff
@@ -110,6 +110,15 @@ int git_repository_open(git_repository **repository, const char *path)
...
 ~~~
 
+Note that some error codes have been defined with a specific meaning in the
+context of callbacks:
+  - `GIT_EUSER` provides a way to bubble up a non libgit2-related failure, which
+  allows it to be preserved all the way up to the initial function call (a `git_cred`
+  setup trying to access an unavailable LDAP server for instance).
+  - `GIT_EPASSTHROUGH` provides a way to tell libgit2 that it should behave as if
+  no callback was provided. This is of special interest to bindings, which would
+  always provide a C function as a "trampoline", and decide at runtime what to do.
```
pks-t: I guess the paragraphs should be aligned with the first character following the bullet point

include/git2/sys/transport.h
```diff
@@ -58,50 +61,65 @@ struct git_transport {
...
 		size_t count);
 
-	/* This function may be called after a successful call to negotiate_fetch(),
-	 * when the direction is FETCH. This function retrieves the pack file for
-	 * the fetch from the remote end. */
+	/**
+	 * Start downloading the packfile from the remote repository.
+	 *
+	 * This function may be called after a successful call to
+	 * negotiate_fetch(), when the direction is FETCH.
```
pks-t: A bit inconsistent, as you spell out `GIT_DIRECTION_FETCH` above

include/git2/sys/transport.h
```diff
@@ -286,58 +294,81 @@ typedef enum {
...
-/* A stream used by the smart transport to read and write data
+/**
+ * A stream used by the smart transport to read and write data
  * from a subtransport */
 struct git_smart_subtransport_stream {
-	/* The owning subtransport */
+	/** The owning subtransport */
 	git_smart_subtransport *subtransport;
 
+	/** Read available data from the stream */
```
pks-t: For those functions we should definitely document their semantics, which aren't immediately obvious (see e.g. #4953). Most importantly, we should specify whether one needs to loop around read/write

include/git2/sys/transport.h
```diff
@@ -349,25 +380,25 @@ typedef struct git_smart_subtransport_definition {
 	 */
 	unsigned rpc;
 
-	/** Param of the callback
-	 */
```
pks-t: Jeez, who the hell thought "Yeah, this formatting looks perfectly fine"?

examples/add.c
```diff
@@ -15,6 +15,18 @@
...
+ * The following example demonstrates how to add files with libgit2.
+ *
+ * It will use the repository in the current working directory, and act
+ * on files passed as its parameters.
+ *
+ * Recognized options are:
+ *   -v/--verbose: show the file's status after acting on it.
+ *   -n/--dry-run: do not actually change the index.
+ *   -u/--update: update the index instead of adding to it.
+ */
```
pks-t: I'd kindly ask you to drop these from this PR until #4956 is merged. Otherwise I'm afraid that it will result in some non-trivial merge conflicts. I can cherry-pick these into #4956 if you want, or you can propose them at a later point, whatever works for you.

tiennou: I'll do that. I have at least 3 example-touching PR already, and I'd feel better handling them as part of #4956 if that's ok with you.

docs/error-handling.md
tiennou: It doesn't matter to the online GitHub display. I'll remove the indents before the `-` though, for consistency.

ethomson:
Two questions about our documentation that ends up rendered on our website:

1. Do we have any instructions for how to view it locally?  IOW, how can I preview the changes that you're making easily?
2. How do things get updated when we do a release?  I rebuilt our CI system and am just now realizing that it doesn't seem to build docs _at all_...?  Am I crazy or is this a manual process right now?

examples/add.c
pks-t: Fair enough. I've cherry-picked your commit

tiennou:
AFAIU, https://libgit2.org/libgit2/ comes from docurium (aka. this repo's gh-pages), which I run sometimes and was planning to do soon™. The rest of https://libgit2.org comes from https://libgit2.github.com/libgit2.github.com (which has a few PR).

Right now it's completely manual, and though I'm making steps teaching docurium NOT to regenerate every version for every run, I don't think it's easily automatable (at least for the time being). I do have a local Rakefile to quickly spin up updated docs (ie. `rake update && rake serve`, browse local docs), which I can share.

examples/add.c
tiennou: Thanks !

include/git2/sys/transport.h
tiennou: I'll keep that open while it's being specified then 😉.

pks-t: I've just checked again to make sure I get it right. This is the state with #4953:

- Both `git_proto_stream_write` and `http_stream_write_single` use `git_stream__write_full` and thus no looping is required. We should specify that a substream implementation shall either write all bytes or return an error code.
- `git_proto_stream_read` tries to read `buf_size` many bytes, but may return less bytes
- `http_stream_read` loops (I think) until a complete HTTP response has been parsed. 

So I guess for the reading part we need to specify only that the implementation may read less bytes than requested and thus may require looping. But one can already see that different implementations behave different, too

include/git2/types.h
tiennou: I think I had this fixed on my tweaked docurium, but it's currently complaing (loudly) about `/var/folders/f4/b3wl780s12gfc25kpvl1rhk40000gn/T/HEAD-20190128-84175-yftmsd/git2/common.h:10:10: fatal error: 'time.h' file not found`, which I'll have to investigate.

include/git2/sys/transport.h
tiennou: Would you be fine with merging this as it is, and revisit the doc after #4953 has landed and semantics are clear ?

ethomson:
Woo.  You can now download a zip of the docs from the PR validation build to check it out locally.  👀😀

include/git2/types.h
tiennou: So, I can confirm that vanilla & tweaked docurium completely ignores "primitive" typedefs. I can tweak it so it doesn't (modulo more fixes to grab the correct comment), but it means I'll pickup all "primitive" typedefs (like our `size_t` `DOCURIUM` override, or the typedef goop inside our Win32 `inttypes.h`)

include/git2/signature.h
```diff
@@ -30,8 +30,8 @@ GIT_BEGIN_DECL
  * @param out new signature, in case of error NULL
  * @param name name of the person
  * @param email email of the person
- * @param time time when the action happened
- * @param offset timezone offset in minutes for the time
+ * @param time time (in seconds from epoch) when the action happened
+ * @param offset timezone offset (in minutes) for the time
```
ethomson: Hmm, these aren't _minutes_ exactly, this is minutes and hours in ISO 8601 format.  (An hour is _100_ not _60_.)  Would you mind fixing that up while you're here?

include/git2/sys/transport.h
```diff
@@ -286,58 +294,81 @@ typedef enum {
...
  *
- * This is used to let the smart protocol code know about the protocol
- * which you are implementing.
+ * The smart transport knows how to speak the git protocol, but it has no
+ * knowledge of how to establish a connection between it and another endpoint,
+ * or how to move data back and forth. For this, a subtransport interface is
+ * declared, and the smart transport delegates this work to the subtransports.
+ *
+ * Three subtransports are provided by libgit2: git, http, and winhttp.
+ * The http and winhttp transports each implement both http and https.
```
ethomson: I would remove the mention of "winhttp" since people can't really get any level of inspection about this.  I'd talk about what is actually user-facing, and that's the protocol.  I would say that three _types_ of subtransport are provided by libgit2: git, ssh and http.

include/git2/sys/transport.h
```diff
@@ -286,58 +294,81 @@ typedef enum {
...
+ *
+ * Subtransports can either be RPC = 0 (persistent connection) or RPC = 1
+ * (request/response). The smart transport handles the differences in its own
+ * logic. The git subtransport is RPC = 0, while http and winhttp are both
+ * RPC = 1.
```
ethomson: I'd drop mention of winhttp here, too.

include/git2/sys/transport.h
```diff
@@ -349,25 +380,25 @@ typedef struct git_smart_subtransport_definition {
...
 /* Smart transport subtransports that come with libgit2 */
 
 /**
- * Create an instance of the http subtransport. This subtransport
- * also supports https. On Win32, this subtransport may be implemented
- * using the WinHTTP library.
+ * Create an instance of the http subtransport.
+ *
+ * This subtransport also supports https. On Win32, this subtransport may be
+ * implemented using the WinHTTP library.
```
ethomson: This, OTOH, seems like a reasonable place to talk about winhttp.  👍

ethomson:
I had a few comments, otherwise 👍

include/git2/signature.h
tiennou: Sorry, invoking ISO8601 just made me 😨. You mean that it's represented as a 4-digit `hhmm` number ?

tiennou: As far as I can see, [parsing](https://github.com/libgit2/libgit2/blob/master/src/signature.c#L256) does what you expect, but we wrangle this as a [number](https://github.com/libgit2/libgit2/blob/master/src/signature.c#L314) of [minutes](https://github.com/libgit2/libgit2/blob/master/src/signature.c#L171) internally.

ethomson: Yeah, sorry, `hhmm`.

tiennou: I might be misunderstanding, but it looks like the `hhmm` decimal format only relates to parsing signatures, not the public API. My "minutes" link above uses `difftime` (hence seconds) to calculate the offset in minutes, and there's no `100` anywhere.

ethomson: Ha, yes, of course you're right.

ethomson:
Thanks @tiennou!

PR #4955, opened 2019-01-24
Fix VS warning C4098: 'giterr_set_str' : void function returning a value

csware:
Regression of PR #4917.


pks-t:
Yup, obviously correct. Thanks for your fix!

PR #4956, opened 2019-01-24
examples: produce single cgit2 binary

pks-t:
Once in a while, I accidentally take a peek into our examples code. Usually, I immediately regret it due to the sorry state our examples are in, which is really unfortunate. I can imagine that these examples are kind of a starting point for a lot of people new to libgit2, and thus they should be as polished as they can be.

Anyway, following boy scouts rule, I always try to at least clean it up by a tiny fraction. This time, I've decided to merge the network examples and the other examples into a single directory. The main motivation is to increase discoverability of the networking examples. While at it, I realized that we might just adopt the git2 executable approach used in the networking examples for all our code, so that now our examples produce a single executable called cgit2 that arbitrates between subcommands like the venerable git(1) does.

If this gets merged, I'd probably like to do some follow-up work. Polishing code, make it conform to our coding guidelines, etc. But better be quick in reviewing, otherwise my motivation might deteriorate rather quick :P


examples/common.h
```diff
@@ -17,6 +17,17 @@
 #include <stdlib.h>
 #include <git2.h>
 
+#ifndef PRIuZ
```
tiennou: Arguably, if there was a way to grab our platform code from the main `common.h` file that would be really appreciable when writing examples.

examples/cgit2.c
```diff
@@ -0,0 +1,109 @@
...
+
+	git_libgit2_init();
+
+	for (args.pos = 1; args.pos < args.argc; ++args.pos) {
+		char *a = args.argv[args.pos];
+
+		if (a[0] != '-') {
+			/* non-arg */
+			break;
+		} else if (optional_str_arg(&git_dir, &args, "--git-dir", ".git")) {
```
tiennou: 🎆

examples/common.h
pks-t: Yeah, I had the same thought. Didn't yet want to go down that route, but by using it we could definitely remove some boilerplate code.

pks-t:
I've rebased it on latest master. I've also renamed the executable from "cgit2" to "lg2", as I realized that cgit2 is easily confused with the cgit web frontend.

pks-t:
I guess this isn't of much importance and thus don't expect a lot of discussion. So I'm just going ahead and press "Merge"

PR #4957, opened 2019-01-24
Don't use deprecated constants

csware:
Follow up for PR #4917.


ethomson:
Thanks!  This was also in #4952 but I think it makes sense to go ahead and take it while I adjust that PR for any feedback.

PR #4922, opened 2019-01-04
Coverity fixes

tiennou:
Again, this fixes some of the latest Coverity reports.

Do note that the last commit attempts to make Coverity realize that clar's assertions are not dependent on `DEBUG`/`NDEBUG`, but it's (almost) a complete shot in the dark.

Ref: https://community.synopsys.com/s/question/0D534000046Yuzb/suppressing-assertsideeffect-for-functions-that-allow-for-sideeffects

PR #4964, opened 2019-01-28
ci: add an individual coverity pipeline

ethomson:
Split the Coverity runs into its own pipeline.  Give this new pipeline two steps: the build step, which may not fail (otherwise the overall pipeline will fail), and the publish step (which may fail, which would cause the pipeline to "partially succeed").


ethomson:
@tiennou, you had the right idea all along.  I've gone ahead and set up a new nightly build in Azure Pipelines for this particular build definition.

tiennou:
Isolation FTW 😉. Let's hope write-access is restored soon, as I'm still waiting to see the effects of modeling from #4922.

PR #4965, opened 2019-01-29
Allow bypassing check for '.keep' file

hackworks:
Retrieving commits involving looking for pack files. Any file operation for git repositories accessed over NAS has a huge performance impact.
For repositories which do not require retaining pack files, we can avoid a call to check for existence of '.keep' file and reduce the overall latency.

### DIFF
--- a/include/git2/common.h
+++ b/include/git2/common.h
@@ -205,7 +205,8 @@ typedef enum {
 	GIT_OPT_SET_ALLOCATOR,
 	GIT_OPT_ENABLE_UNSAVED_INDEX_SAFETY,
 	GIT_OPT_GET_PACK_MAX_OBJECTS,
-	GIT_OPT_SET_PACK_MAX_OBJECTS
+	GIT_OPT_SET_PACK_MAX_OBJECTS,
+	GIT_OPT_DISABLE_PACK_KEEP_FILE_CHECKS
 } git_libgit2_opt_t;
 
 /**
@@ -394,6 +395,10 @@ typedef enum {
  *
  *		> Set the maximum number of objects libgit2 will allow in a pack
  *		> file when downloading a pack file from a remote.
+ *
+ *	 opts(GIT_OPT_DISABLE_PACK_KEEP_FILE_CHECKS, int enabled)
+ *		> This will cause .keep file existence checks to be skipped when
+ *		> accessing packfiles, which can help performance with remote filesystems.
  *
  * @param option Option key
  * @param ... value to set the option
--- a/src/pack.c
+++ b/src/pack.c
@@ -16,6 +16,9 @@
 
 #include <zlib.h>
 
+/* Option to bypass checking existence of '.keep' files */
+bool git_disable_pack_keep_file_checks = false;
+
 static int packfile_open(struct git_pack_file *p);
 static git_off_t nth_packed_object_offset(const struct git_pack_file *p, uint32_t n);
 static int packfile_unpack_compressed(
@@ -1141,9 +1144,11 @@ int git_packfile_alloc(struct git_pack_file **pack_out, const char *path)
 	if (git__suffixcmp(path, ".idx") == 0) {
 		size_t root_len = path_len - strlen(".idx");
 
-		memcpy(p->pack_name + root_len, ".keep", sizeof(".keep"));
-		if (git_path_exists(p->pack_name) == true)
-			p->pack_keep = 1;
+		if (!git_disable_pack_keep_file_checks) {
+			memcpy(p->pack_name + root_len, ".keep", sizeof(".keep"));
+			if (git_path_exists(p->pack_name) == true)
+				p->pack_keep = 1;
+		}
 
 		memcpy(p->pack_name + root_len, ".pack", sizeof(".pack"));
 	}
--- a/tests/pack/packbuilder.c
+++ b/tests/pack/packbuilder.c
@@ -14,6 +14,8 @@ static git_vector _commits;
 static int _commits_is_initialized;
 static git_transfer_progress _stats;
 
+extern bool git_disable_pack_keep_file_checks;
+
 void test_pack_packbuilder__initialize(void)
 {
 	_repo = cl_git_sandbox_init("testrepo.git");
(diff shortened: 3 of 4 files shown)


src/pack.c
```diff
@@ -46,6 +48,12 @@ static int packfile_error(const char *message)
 	return -1;
 }
 
+int git_pack_global_init(void)
+{
+	global_ignore_keep_pack_file = !!getenv("GIT_IGNORE_PACK_KEEP");
```
ethomson: Generally we try to avoid adding new global initializers.  For something where we're just reading an environment variable, and there aren't really any race conditions or otherwise need to be initialized strictly at the start of library execution, we can probably get by with simpler constructs.

A simple `get`ter type function that does the lookup on-demand would probably suffice?

ethomson:
Welcome to the libgit2 project, @hackworks!  I think that this seems like a reasonable change, but I'd like to iterate on it just a bit.  Thanks for opening the PR!

hackworks:
> Welcome to the libgit2 project, @hackworks! I think that this seems like a reasonable change, but I'd like to iterate on it just a bit. Thanks for opening the PR!

Thank you for encouraging and welcoming. I would love to contribute more. I come from a storage background and have spent time optimizing NFS stack for build and database workloads.

src/pack.c
hackworks: The main reason for not calling 'getenv()' each time is to avoid the performance overhead of such calls when the library is used in a server environment. Fetching the value and storing it in a local variable will make it better.

Had we fetched the settings from the environment during the startup, it would have ensured
thread safety - calls to getenv() racing with setenv() [if it exists in main application].

Ideally, I would have preferred to do it using pthread_once. As you rightly pointed, it could be an overkill for this use case.

On a side topic: If we expect customization of the behavior of libgit2, it might be nice to have a config file that we can load at startup and use it across the code/runtime.

tiennou: Hello, and welcome! I'd suggest adding that as a flag to `git_libgit2_opts`, as it would move the responsibility of initialization (races & all) to the user's code. I can see that there's some prior art in the `git_repository__fsync_gitdir` flag.

hackworks: > Hello, and welcome! I'd suggest adding that as a flag to `git_libgit2_opts`, as it would move the responsibility of initialization (races & all) to the user's code. I can see that there's some prior art in the `git_repository__fsync_gitdir` flag.

This is a good approach. Shifting the responsibility to the main application is best for a library.

ethomson: I agree with @tiennou's suggestion to use a settable option here - we try to avoid environment variables when possible, preferring to let the hosting application handle set up.

tests/pack/packbuilder.c
```diff
@@ -260,3 +262,10 @@ void test_pack_packbuilder__foreach_with_cancel(void)
 		git_packbuilder_foreach(_packbuilder, foreach_cancel_cb, idx), -1111);
 	git_indexer_free(idx);
 }
+
+void test_ignore_pack_keep_file_check(void)
```
ethomson: Apologies for this, but the test framework is a little tricky.  Since C lacks things that other unit testing frameworks use for identifying unit tests (like attributes), we actually use naming conventions to separate a test function from a regular static helper function.

All tests start with `test_` and are followed by the section, which also corresponds to the filename that they're in.  Here, these are the `pack::packbuilder` tests, they're in the `pack/packbuilder.c` file.  (You can probably see how the hierarchy between test names and filenames match up).  A double-underscore in the function name indicates an actual test.  So the function above this one, `test_pack_packbuilder__foreach_with_cancel` is a test named `pack::packbuilder::foreach_with_cancel`.

(Special are the `initialize` and `shutdown` functions, which are not tests but will be invoked at the beginning and end of each test invocation.)

Anything that doesn't follow this naming convention is assumed to be a helper method and ignored.  So changing this to be named `test_pack_packbuilder__keep_file_check` or something similar would ensure that it gets invoked as a test.

include/git2/common.h
```diff
@@ -205,7 +205,8 @@ typedef enum {
 	GIT_OPT_SET_ALLOCATOR,
 	GIT_OPT_ENABLE_UNSAVED_INDEX_SAFETY,
 	GIT_OPT_GET_PACK_MAX_OBJECTS,
-	GIT_OPT_SET_PACK_MAX_OBJECTS
+	GIT_OPT_SET_PACK_MAX_OBJECTS,
+	GIT_OPT_IGNORE_PACK_KEEP_FILE_CHECK
```
tiennou: `GIT_OPT_DISABLE_PACKFILE_KEEP_CHECK/S` maybe ? I'd like to be consistent in what verbiage we use to enable/check/set/ignore…

Also, adding some documentation to the monster just below would be ❤️.

> * opts($name, …)
> This will cause `.keep` file existence checks to be skipped when accessing packfiles,
> which can help performance with remote filesystems.

src/pack.c
```diff
@@ -1141,9 +1144,11 @@ int git_packfile_alloc(struct git_pack_file **pack_out, const char *path)
 	if (git__suffixcmp(path, ".idx") == 0) {
 		size_t root_len = path_len - strlen(".idx");
 
-		memcpy(p->pack_name + root_len, ".keep", sizeof(".keep"));
-		if (git_path_exists(p->pack_name) == true)
-			p->pack_keep = 1;
+		if (!git_ignore_pack_keep_file_check) {
+			memcpy(p->pack_name + root_len, ".keep", sizeof(".keep"));
+			if (git_path_exists(p->pack_name) == true)
+			    p->pack_keep = 1;
```
tiennou: I just noticed that this is "dead code": we don't do anything with that state after setting it, as AFAIK we don't have any API to delete things yet.

tests/pack/packbuilder.c
```diff
@@ -260,3 +262,10 @@ void test_pack_packbuilder__foreach_with_cancel(void)
 		git_packbuilder_foreach(_packbuilder, foreach_cancel_cb, idx), -1111);
 	git_indexer_free(idx);
 }
+
+void test_pack_packbuilder__keep_file_check(void)
+{
+	assert(!git_ignore_pack_keep_file_check);
+	cl_git_pass(git_libgit2_opts(GIT_OPT_IGNORE_PACK_KEEP_FILE_CHECK, true));
```
tiennou: This should be reset in `__cleanup`.

src/pack.c
hackworks: I retained the code just in case you are writing a pack file, this setting is part of the pack file metadata and would get written to filesystem.
The default git client might/could look into this instead of looking for '.keep' file. This makes it more self contained too.

ethomson:
I've no complaints with this.  I'll be happy to merge it as soon as v0.28 is out the door.  Right now I want to keep the master branch frozen to make sure that there are no complaints about v0.28.0-rc1 that need to be addressed.

I expect that we should thaw out the master branch directly, it's been nearly a week with no reports of issues.

ethomson:
Thanks, @hackworks.  Now that 0.28 is out, this is ready to go.

Issue #4671, opened 2018-06-05
git_diff_foreach on bare repository fails with 'path cannot exist in repository'


kcsaul:
Hi,

When attempting to upgrade to a more recent version from v25.1, I encounter the 'path cannot exist in repository' error, which I believe was introduced as part of the worktree changes (I had attempted to upgrade previously around that time when I first seen this error, after which I'd put this off until I could investigate further to try and work it out).

#### Background Info
The purpose of [my project](https://gitlab.com/jadelab/jadegit) is to implement source control for an OO database platform, where all source code is stored in the database.  Hence, there is no working directory as such, instead the goal is to do everything in-memory with a bare repository in combination with custom odb, refdb and config backends which store everything in the OO database.

### Reproduction steps
The problem occurs in a context where a bare repository is being used which doesn't have the commondir path set (presumably because I've used git_repository_new with a custom odb backend?).

When attempting to use git_diff_foreach, it ultimately calls git_repository_item_path to determine the patch for GIT_REPOSITORY_ITEM_INFO, for which it tries to use commondir which is empty.

I started looking at how I might try to recreate this as a unit test, but wasn't sure where such a scenario should be added in and looks like it'd take me a while to figure out (unfamiliar territory) - So thought it best I raise the issue for discussion first, in case it's been seen before and it's a simple fix, or if there's something I've missed that I need to check/adjust within my project.

### Expected behavior
Iterate diff as normal like v25.1, with the default driver being selected as it was previously.

### Actual behavior
git_diff_foreach fails with 'path cannot exist in repository', while trying to determine what driver to use (call stack below for reference).

### Version of libgit2 (release number or SHA1)
v27.1

### Operating system(s) tested
Windows 10

```
git2.dll!git_repository_item_path(git_buf * out, const git_repository * repo, git_repository_item_t item) Line 2271
git2.dll!attr_setup(git_repository * repo, git_attr_session * attr_session) Line 340
git2.dll!collect_attr_files(git_repository * repo, git_attr_session * attr_session, unsigned int flags, const char * path, git_vector * files) Line 502
git2.dll!git_attr_get_many_with_session(const char * * values, git_repository * repo, git_attr_session * attr_session, unsigned int flags, const char * pathname, unsigned __int64 num_attr, const char * * names) Line 134
git2.dll!git_diff_driver_lookup(git_diff_driver * * out, git_repository * repo, git_attr_session * attrsession, const char * path) Line 369
[...]

PR #4967, opened 2019-01-30
Incomplete commondir support

tiennou:
This comes from #4671, but I'd like to discuss the problem because I think it's a shortcoming of our "item locator" (and/or repository initialization), thus my initial fix is only papering over the underlying issue.

What happens is that you can (easily) get yourself a repository (using `git_repository_new`) that has an unset `commondir` member, which would cause failures in the locator, as it has no fallback mechanism if this member is not set (it reports `GIT_ENOTFOUND`).

AFAICS there's a bunch of things we can do:
- [ ] add a fallback system to the locator.
As per [gitrepository-layout], `$GIT_DIR/info` is shadowed by `$GIT_COMMON_DIR/info` if it's set (set meaning envvar, or a `$GIT_DIR/commondir` "gitlink".

- [ ] add some API to change the `commondir`.
Right now there's no way to do this, it's initialized on `new` or `open`, and set in stone at that time. I'm not sure if that's really needed, but to me it seems like some kind of shortcoming (since we allow swapping out ODBs and things, even if it's kinda crazy).

- [ ] make the relevant users of the item locator ignore issues.
This is the approach I started here, before I took a look at the layout document and noticed the discrepancy. The layout doc seems to imply that missing those directories is fine, thus we're just being overzealous.

- [ ] introduce some layout-validity tests

@pks-t I'd like your opinion on these points before tackling them, as my gut feeling is that it's some kind of regression introduced by the worktree support (at least users that use `git_repository_new` now have an unfixable half-working repository, as per #4671).

[gitrepository-layout]: https://git-scm.com/docs/gitrepository-layout

Helpful audit regex:
```
/GIT_REPOSITORY_ITEM_(OBJECTS|(PACKED_)REFS|REMOTES|CONFIG|INFO|HOOKS|LOGS|WORKTREE)/
```


pks-t:
Oops, should've checked whether the #4671 is being referenced already by a PR. I've started running down the rabbit hole with our attrs code to fix this exact issue. Anyway, I'll upload it later and then we can see how to proceed -- at least we ran into the different directions after doing the obvious immediate fix to the attrs code :)

> As per gitrepository-layout, $GIT_DIR/info is shadowed by $GIT_COMMON_DIR/info if it's set (set meaning envvar, or a $GIT_DIR/commondir "gitlink".

Well, right now our support for `GIT_COMMON_DIR` environment variable doesn't exist at all:

```
	error = git__getenv(&work_tree_buf, "GIT_COMMON_DIR");
	if (error == GIT_ENOTFOUND)
		git_error_clear();
	else if (error < 0)
		goto error;
	else {
		git_error_set(GIT_ERROR_INVALID, "GIT_COMMON_DIR unimplemented");
		error = GIT_ERROR;
		goto error;
	}
```

So first step here would be to actually implement it at all, and then afterwards we may implement the fallback mechanism. I don't remember right now whether this was part of #4856 or not.

> add some API to change the commondir.
> Right now there's no way to do this, it's initialized on new or open, and set in stone at that time. I'm not sure if that's really needed, but to me it seems like some kind of shortcoming (since we allow swapping out ODBs and things, even if it's kinda crazy).

Personally, I'd wait until we see somebody requesting it. Until now, nobody has complained about our lacking API here. I just fear that as soon as we do that, people try to change those variables half-way through the repo's lifetime, which may lead to inconsistencies.

That being said, I could easily see us adding an extended API to create repos. As you say, `git_repository_new` doesn't help too much. It might make sense to create a new API here like e.g. `git_repository_new_with_opts` (name is subject to change), where you can pass additional information. This would be stuff like commondir, gitdir, ODB, refdb, etc.

> make the relevant users of the item locator ignore issues.
> This is the approach I started here, before I took a look at the layout document and noticed the discrepancy. The layout doc seems to imply that missing those directories is fine, thus we're just being overzealous.

Yup, definitely agreed. I've also fixed this in a local branch already.

> introduce some layout-validity tests

That'd be rad. Git has grown increasingly complex over time, making the precedence and lookup rules non-trivial. So I'd be very eager to see some tests, even though I can imagine it being quite hard for some parts of libgit2.

ethomson:
> Personally, I'd wait until we see somebody requesting it.

Strongly agree here, I think that there's vast opportunity to increase our complexity and potentially do this "wrong" from the standpoint of a consumer if we eagerly try to implement it without a use case.

As a concrete example of the existing complexity:  I was just creating an inmemory repository over the weekend and got caught up in a few issues that are not obvious and make assumptions about gitdir/commondir existing.  (I'll open up a PR when I have a moment).

tiennou:
> or a $GIT_DIR/commondir "gitlink".

Note that I was referring to this kinda-gitlink way, which we're already supporting : https://github.com/libgit2/libgit2/blob/master/src/repository.c#L193

I'm completely okay with not adding custom commondir support, because it might not make any sense. @pks-t suggestion of finding an API to provide those attributes is arguably more in-memory-ish. I was going the way that creating the file would make it pick it up, but it would act sanely if it's missing.

pks-t:
> I was going the way that creating the file would make it pick it up, but it would act sanely if it's missing.

That's definitely the right thing to do and is completely independent of improved commondir or in-memory attributes support. It's also the same thing that I've been doing in #4995 (which crossed work, as I initially didn't spot this PR as fixing the same issue due to the different title). I'm mighty annoyed though that it's impossible to write a test without either in-memory gitattributes or commondir mangling. But yeah, that's the way it is.

tiennou:
I've pushed my current WIPs on top of that branch, which should have a working `commondir` => `gitdir` override if there's no `commondir`. I'm not sure I understand what is wrong inside the attribute loading, so I'd like to defer more in-depth changes to your other PR and keep the `commondir`/`git_repository_new` fixes (for limited functionality).

include/git2/sys/repository.h
```diff
@@ -25,11 +25,25 @@ GIT_BEGIN_DECL
...
 
+/**
+ * Set or unset a repository's path
+ *
+ * @param repo A new repo to set the path of.
+ * @param path The path to use.
+ *
+ * @return 0 on success, or an error code
+ */
+GIT_EXTERN(int) git_repository_set_path(git_repository *repo, const char *path);
```
tiennou: That allows giving a repository a filesystem location. It's not quite clear what will happen to it, but at least that should fix the "cannot exist in repository", if a user choses to do so.

src/odb.c
```diff
@@ -634,6 +638,42 @@ int git_odb_add_disk_alternate(git_odb *odb, const char *path)
...
+{
+	git_odb *db;
+	git_buf objects_dir = GIT_BUF_INIT;
+	int error;
+
+	assert(out && repo);
+
+	*out = NULL;
+
+	if ((error = git_repository_item_path(&objects_dir, repo,
```
tiennou: This function got extracted because it's the interface point between them, and *normally* the place where the odb would initialize its backends' filesystem locations (hence it was more logical in `odb`). That's still in flux though.

kcsaul:
@tiennou I'm getting some build errors in regards the new constant, do I need to change a compiler setting?  (using MSVC).

C2099	initializer is not a constant	\libgit2\src\repository.c	Line 50	
C2440	'initializing': cannot convert from 'void *' to 'git_repository_item_t' \libgit2\src\repository.c	Line 50

tiennou:
Hmm, I'm confused as well, and can't take a look right now. Maybe it doesn't see a definition for `git_repository_item_t`, uses `* whatever`, and then complains about it not being const-compatible ? I think that approach was actually reverted because of MSVC…

kcsaul:
@tiennou For now, I've been able to fix the build error by moving the 'none' constant to the enumeration in the header.
```
typedef enum {
	GIT_REPOSITORY_ITEM__NONE = -1,
	GIT_REPOSITORY_ITEM_GITDIR,
	GIT_REPOSITORY_ITEM_WORKDIR,
	GIT_REPOSITORY_ITEM_COMMONDIR,
	...
} git_repository_item_t;
```

Happy to confirm I'm no longer getting the 'path cannot exist in repository' error (#4671) and things seem to be working as they were previously now 😄

tiennou:
Rebased, I've made the fallback value public since it can't be "hidden".

Note that I've seen to have messed up my brain when writing the PR, so let's recap the issues:
- As there can be accesses to things inside the gitdir itself, for which we have no "repository backend" (lots of direct file accesses, actually), users going through `git_repository_new` would have no way to set a location. With no gitdir, some systems would fail to work for those users. Those are the first few fixes from the PR.
- We were missing the fallback via the commondir mechanism, used for some items. This has been implemented.
- Since being able to set the location of a repository seemed generally useful (to cater for the users of 1) that do want to make use of those systems "correctly"), so `git_repository_set_path` was added.

I think not much is missing from it, it just needs some thought about whether we should do 3), or if the first two are enough, and 3) can wait for a "repository backend" interface. In other words, I'm happy to drop the test and the setter for now (or make it private and keep them).

src/ignore.c
```diff
@@ -336,14 +336,13 @@ int git_ignore__for_path(
...
+		repo, GIT_REPOSITORY_ITEM_INFO)) < 0 && error != GIT_ENOTFOUND)
 		goto cleanup;
 
 	/* load .git/info/exclude */
-	error = push_ignore_file(
+	if (error != GIT_ENOTFOUND && (error = push_ignore_file(
 		ignores, &ignores->ign_global,
-		infopath.ptr, GIT_IGNORE_FILE_INREPO);
-	if (error < 0)
+		infopath.ptr, GIT_IGNORE_FILE_INREPO)) < 0)
```
pks-t: I think this'd be easier to understand if you refactored it like you did refactor the attr code:

```
if ((error = git_repository_item_path(&infopath, ...)) < 0) ||
    (error = push_ignore_file(ignores, ...)) < 0) {
        if (error != GIT_ENOTFOUND)
            goto cleanup;
        error = 0;
}

File: src/odb.h [hunk 4]
```diff
@@ -54,6 +57,10 @@ typedef enum {
  */
 int git_odb__set_caps(git_odb *odb, int caps);
 
+int git_odb__open_own(git_odb **out, git_repository *repo);
+
+int git_odb__is_default(git_odb *odb);
```
pks-t: This isn't yet used, right? What's the usecase of it?

src/repository.c
```diff
@@ -2323,9 +2314,17 @@ int git_repository_item_path(git_buf *out, const git_repository *repo, git_repos
 			break;
 		default:
 			git_error_set(GIT_ERROR_INVALID, "invalid item directory");
-			return -1;
+			return NULL;
 	}
+	if (!parent && fallback != GIT_REPOSITORY_ITEM__LAST) {
+		return resolved_parent_path(repo, fallback, GIT_REPOSITORY_ITEM__LAST);
+	}
```
pks-t: Style nit: braces aren't required here

include/git2/sys/odb_backend.h
```diff
@@ -100,6 +100,19 @@ struct git_odb_backend {
...
+	 *
+	 * "object" is the git standard, but can be NULL to signal that the backend
+	 * does not have a filesystem location (a.k.a is in-memory).
+	 *
+	 * Note that at the moment, adding a non-default backend to the odb will
+	 * result in a repository where GIT_REPOSITORY_ITEM_OBJECTS cannot make
+	 * sense, and will thus be disabled. Passing "objects" here is thus
+	 * reserved to backends that can make sense of its contents.
+	 */
+	const char *location;
```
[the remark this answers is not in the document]
tiennou: I was attempting *something* with that, but I'm not quite sure now. I felt that knowing that we're using our own "standard" backends was something generally useful. And I felt that knowing if a backend has or has no filesystem location was as well, and that this aspect could be used as part of the definition of what is an in-memory repository. You could well imagine a backend that uses the filesystem, but it's not ".git/objects"-compatible: how can you signal that to the repository, so that it knows to report that its `GIT_REPOSITORY_ITEM_OBJECTS` contents is non-standard ? Not sure where I was going with this though…

include/git2/sys/repository.h
tiennou: 👍 on the naming issue, and I can take care of the deprecation here if we decide for it.

src/odb.h
tiennou: To know if we're running against the "default" backend stack, when only loose/packed backends are present. Not sure if it's useful as well…

src/attr.c
```diff
@@ -306,53 +306,49 @@ static int system_attr_file(
...
-
-	if ((error = preload_attr_file(
-			repo, attr_session, GIT_ATTR_FILE__FROM_FILE,
-			NULL, git_repository_attr_cache(repo)->cfg_attr_file)) < 0)
-		goto out;
+	if ((error = system_attr_file(&path, attr_session)) < 0 ||
+	    (error = preload_attr_file(repo, attr_session, GIT_ATTR_FILE__FROM_FILE,
+				       NULL, path.ptr)) < 0) {
+		if (error != GIT_ENOTFOUND)
+			goto out;
```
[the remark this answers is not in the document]
tiennou: I think that's on you 😆.

src/attr.c
```diff
@@ -306,53 +306,49 @@ static int system_attr_file(
...
+	}
 
-	if ((error = git_repository_item_path(&path,
-			repo, GIT_REPOSITORY_ITEM_INFO)) < 0)
+	if ((error = preload_attr_file(repo, attr_session, GIT_ATTR_FILE__FROM_FILE,
+				       NULL, git_repository_attr_cache(repo)->cfg_attr_file)) < 0)
```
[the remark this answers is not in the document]
tiennou: That's true. I have not really read that code, but I'll check what it expects.

include/git2/sys/odb_backend.h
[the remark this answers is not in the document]
pks-t: I agree that this would be useful. But that'd require a new callback in the backends, as we generally shouldn't assume anything of backends. Might very well be (well, probably not) that somebody has implemented his own implementation of a ".git/objects" backend that is uber fast due to it using platform-specific things.

include/git2/sys/repository.h
pks-t: Mh. Dunno -- I think it would be easier to do in another PR to keep the code churn low and keep this reviewable. The deprecation things always require to make a sweep through all of our code, and I bet that `git_repository_path` is used in quite a lot of places.

src/attr.c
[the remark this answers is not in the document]
pks-t: Haha. At least I seem to consistently like my own style :P

src/odb.h
pks-t: Then let's keep it out for now, ok? Not opposed to adding this at all, but I really would like to merge this PR and discussion of those things might put it on hold longer than needed.

src/attr.c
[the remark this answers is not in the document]
tiennou: `preload_attr_file` calls `git_attr_cache__get`, which swallows `GIT_ENOTFOUND` already, so this looks fine:

https://github.com/libgit2/libgit2/blob/c544850127abb96b472485768dee3ede68e1622e/src/attrcache.c#L236-L250

tiennou:
Rebased, I've separated out all my experiments to `feature/odb-fs-location`, so this fixes the following issues only :
- as custom backends have no way to store files on-disk (they can, but cannot inform us that they support those accesses, and some of those files aren't actually under the ODB's purview), any higher-level system that tries to access a "standard" git file will get back `GIT_ENOTFOUND`. Some systems didn't respond well to that (diff in #4671, attr in #4995). IIRC I've audited all calls to `git_repository_item_path` at the time, so there should be no other fixes.
- additionally, we were missing some of the [fallbacks](https://git-scm.com/docs/gitrepository-layout#Documentation/gitrepository-layout.txt-info) done by `git`. I don't think there's an issue for that though, as it's worktree-related; it's not clear to me what kind of weirdness would have resulted.

I'm kinda bummed that we don't have that many tests for those kind of ODB backend issues… I've looked at our libgit2-backend repository but it's not in a good shape. Would integrating at least one of the simpler ones as part of our CI "examples" runs or something would be good ? SQLite backend maybe, as it still needs a "locally discoverable" file — say `.git/sqlite` — and could thus in theory still support the `item_path` API, but cannot because of the aforementioned issue. Ideally that'd provide some guidance as to what we functionally expect from an ODB backend, be sufficiently un-git-like that it would exercise the API in novel ways, helping to find and correct those "missing parts" ?

include/git2/sys/repository.h
```diff
@@ -25,11 +25,25 @@ GIT_BEGIN_DECL
  * Note that this is only useful if you wish to associate the repository
  * with a non-filesystem-backed object database and config store.
  *
+ * Caveats: since this repository has no physical location, some systems
+ * can fail to function properly: locations under $GIT_DIR, $GIT_COMMON_DIR,
+ * or $GIT_INFO_DIR are impacted.
```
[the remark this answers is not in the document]
pks-t: This comment of mine is now obsolete, as we do not yet have the new APIs

src/attr.c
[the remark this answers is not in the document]
pks-t: Huh, confusing. But well, fair enough, thanks for digging in

pks-t:
Thanks @tiennou!

csware:
This PR breaks various tests in TortoiseGit and changes the output of libgit2.

csware:
This PR breaks various tests in TortoiseGit and changes the output of libgit2 (cf. https://ci.appveyor.com/project/TortoiseGit/tortoisegit-06nst/builds/25633104?fullLog=true).

I bisected the libgit2 commit history and the first faulty commit it 5452e49fce21f726bec19519da7f012e3f19e736.

One case is:
Diffing two trees a call to `git_diff_get_stats` fail with `-1` and `git_error_last` returns `cannot grow a borrowed buffer` (comparing `b02add66f48814a73aa2f0876d6bbc8662d6a9a8` with `b9ef30183497cdad5c30b88d32dc1bed7951dfeb` in the repository https://gitlab.com/tortoisegit/tortoisegit/tree/master/test/UnitTests/resources/git-repo1).

Fix is in PR #5152.

PR #4973, opened 2019-02-07
Update threading.md

tiennou:
As pointed out in our Slack channel, our threading explanation is confusing, and also contains some outdated references (eg. `cURL`, `Mac OS X`).

This fixes the references, and tries to clarify that you should really be careful passing objects between threads (though that explanation is also lacking somewhat, but at least you're tempted to keep reading).

### DIFF
--- a/docs/threading.md
+++ b/docs/threading.md
@@ -1,8 +1,10 @@
-Threads in libgit2
+Threading in libgit2
 ==================
 
-You may safely use any libgit2 object from any thread, though there
-may be issues depending on the cryptographic libraries libgit2 or its
+Unless otherwise specified, libgit2 objects cannot be safely accessed by
+multiple threads.
+
+There are also caveats on the cryptographic libraries libgit2 or its
 dependencies link to (more on this later). For libgit2 itself,
 provided you take the following into consideration you won't run into
 issues:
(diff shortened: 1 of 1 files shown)


docs/threading.md
```diff
@@ -44,31 +46,20 @@ steps necessary. If you are using a MinGW or similar environment where
...
 
-If it's available, by default we use libcurl to provide HTTP tunneling support,
-which may be linked against a number of cryptographic libraries and has its
-own
-[recommendations for thread safety](https://curl.haxx.se/libcurl/c/threadsafe.html).
-
-If there are no alternative TLS implementations (currently only
-SecureTransport), libgit2 uses OpenSSL in order to use HTTPS as a transport.
+If there are no applicable TLS implementations (currently only
+SecureTransport and mbedTLS), libgit2 defaults to OpenSSL in order to use HTTPS as a transport.
```
ethomson: "applicable" feels like the wrong word here (OpenSSL would also be "applicable"), which I think is why it says "alternative" (these are the "alternatives" to the default of OpenSSL).

I think we should just be more clear about the defaults, something like:

> By default, libgit2 uses the Secure Transport on macOS, WinHTTP and CryptoNG on Windows and OpenSSL on all other systems.  mbedTLS may be optionally enabled as the security provider.

docs/threading.md
```diff
@@ -1,8 +1,10 @@
-Threads in libgit2
+Threading in libgit2
 ==================
 
-You may safely use any libgit2 object from any thread, though there
-may be issues depending on the cryptographic libraries libgit2 or its
+Unless otherwise specified, libgit2 objects cannot be safely accessed by
+multiple threads.
```
ethomson: Let's clarify further what this means.  eg:

> Unless otherwise specified, libgit2 objects cannot be safely accessed by multiple threads simultaneously.

ethomson:
Manually merged.  Thanks @tiennou!

PR #4975, opened 2019-02-12
ci: add ssl proxy tests

ethomson:
In theory, we support proxies over HTTPS.  Test that we can support this _in fact_.


src/transports/http.c
```diff
@@ -698,15 +698,7 @@ static int load_proxy_config(http_subtransport *t)
...
-	if ((error = gitno_connection_data_from_url(&t->proxy.url, t->proxy_opts.url, NULL)) < 0)
-		return error;
-
-	if (t->proxy.url.use_ssl) {
-		git_error_set(GIT_ERROR_NET, "SSL connections to proxy are not supported");
-		return -1;
-	}
-
-	return error;
+	return gitno_connection_data_from_url(&t->proxy.url, t->proxy_opts.url, NULL);
```
pks-t: 52478d7 says something about Valgrind errors. Have these actually been fixed or do we know the root cause of them by now?

ci/test.sh
```diff
@@ -182,9 +182,20 @@ if [ -z "$SKIP_PROXY_TESTS" ]; then
...
+    export GITTEST_REMOTE_PROXY_SCHEME="https"
+    export GITTEST_REMOTE_PROXY_HOST="localhost:8081"
+    export GITTEST_REMOTE_PROXY_SELFSIGNED=1
+    run_test proxy
+
+    unset GITTEST_REMOTE_PROXY_SCHEME
 	unset GITTEST_REMOTE_PROXY_HOST
 	unset GITTEST_REMOTE_PROXY_USER
 	unset GITTEST_REMOTE_PROXY_PASS
+    unset GITTEST_REMOTE_PROXY_SELFSIGNED
```
pks-t: The indentation here looks a bit off, at least on GitHub

src/transports/http.c
ethomson: Don't yet understand them.  Could be that we're calling something wrong or setting something up wrong if we want to use two SSL contexts?  But I fear that the old version of OpenSSL that is on our CI images might have some reentrancy problems.

In any case, I wanted to open this up since it was requested in #4515 and @josebalius suggested that he might want to investigate.

PR #4982, opened 2019-02-14
Enable creation of worktree from bare repo's default branch

### LINKED ISSUE #4977: git_worktree_add fails on master branch with already checked out
Other branches work all OK just seems to be the master that fails.  Note that the repo is bare just in case that has any effect.
 
### Reproduction steps
```
git clone --bare url
```
```
    git_reference_lookup(ref, Handle, 'refs/heads/master')
    git_worktree_add_init_options(@opts, GIT_WORKTREE_ADD_OPTIONS_VERSION);
    opts.ref := ref;
    r := git_worktree_add(wt, Handle, 'worktreename', '/some/path', @opts)
```
### Expected behavior
r = 0 (Success)

### Actual behavior
r = -4 (Fail)
Error message = "reference is already checked out"

0.28

Windows

### Additional Information
After the failure partial work tree structure is left behind which maybe should have been cleaned up or maybe the validity checks should happen before any structure gets created - which may or may not be possible.  

Anyway if I clear out the invalid worktree structure that was left behind and run
```
git worktree add --checkout /some/path master
```
it works just fine so we know that the master was not checked out anywhere else

pks-t:
Previously, we have determined whether a branch is checked out or not by simply checking the repo and all its worktree HEADs. This is the wrong thing to do in case where we have a bare repo, as a bare repo's HEAD will not be checked out but only indicate the "default" branch. This is getting fixed by this PR.

Furthermore, we'll now verify a given ref when adding new worktrees before creating the on-disk data structures. In case where the ref is invalid, we'll now stop leaving behind those files.

This fixes #4977

### DIFF
--- a/src/branch.c
+++ b/src/branch.c
@@ -153,7 +153,10 @@ done:
 
 int git_branch_is_checked_out(const git_reference *branch)
 {
-	assert(branch && git_reference_is_branch(branch));
+	assert(branch);
+
+	if (!git_reference_is_branch(branch))
+		return 0;
 
 	return git_repository_foreach_head(git_reference_owner(branch),
 		branch_equals, (void *) branch) == 1;
--- a/src/branch.c
+++ b/src/branch.c
@@ -153,13 +153,20 @@ done:
 
 int git_branch_is_checked_out(const git_reference *branch)
 {
+	git_repository *repo;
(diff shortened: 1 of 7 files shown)


ethomson:
Nice, thanks @pks-t.

PR #4984, opened 2019-02-15
refdb_fs: fix loose/packed refs lookup racing with repacks

pks-t:
This is a different approach to #4981 that should correctly fix the issues described by @peff. 

There had been an ordering dependency between loading loose and packed refs in that loading loose refs set the PACKREF_SHADOWED flag in already loaded packed refs. Upon reloading packed refs, this flag gets reset. Due to that, we always had to load loose and packed refs in the wrong order. This PR fixes this dependency and thus allows us to first load the loose refs and only afterwards load the packed refs, effectively mitigating the race.

It also fixes the same race condition in the `exists` callback that has been pointed out by @peff, too.

### DIFF
--- a/src/refdb_fs.c
+++ b/src/refdb_fs.c
@@ -681,40 +681,41 @@ static int refdb_fs_backend__iterator_next_name(
 static int refdb_fs_backend__iterator(
 	git_reference_iterator **out, git_refdb_backend *_backend, const char *glob)
 {
-	int error;
-	refdb_fs_iter *iter;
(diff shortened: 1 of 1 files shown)


pks-t:
Force-pushed to provide proper commit messages and remove unrelated commits.

ethomson:
Cool, thanks for doing this, and thanks for the report @peff.

ethomson:
Note: I logged #5003 around the `delete_tail` issue that @peff noted in #4981, so that we would not forget.

pks-t:
Thanks, @ethomson! And thanks again @peff.

PR #4986, opened 2019-02-16
Make stdalloc__reallocarray call stdalloc__realloc

lhchavez:
This change avoids calling realloc(3) in more than one place.


ethomson:
> This change avoids calling realloc(3) in more than one place.

What's the advantage of avoiding that?

lhchavez:
consistency, i guess. all the other functions try to delegate the calls to the underlying libc function as much as possible.

PR #4987, opened 2019-02-17
Fix a memory leak in odb_otype_fast()

lhchavez:
This change frees a copy of a cached object in odb_otype_fast().

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -1128,6 +1128,7 @@ static int odb_otype_fast(git_object_t *type_p, git_odb *db, const git_oid *id)
 
 	if ((object = git_cache_get_raw(odb_cache(db), id)) != NULL) {
 		*type_p = object->cached.type;
+		git_odb_object_free(object);
 		return 0;
 	}
 
--- a/tests/odb/mixed.c
+++ b/tests/odb/mixed.c
@@ -263,3 +263,24 @@ void test_odb_mixed__expand_ids(void)
 	git__free(ids);
 }
 
+void test_odb_mixed__expand_ids_cached(void)
+{
+	git_odb_expand_id *ids;
+	size_t i, num;
+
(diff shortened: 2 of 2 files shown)


lhchavez:
Found an uncovered codepath while hunting for an unrelated memory leak.

Unfortunately the test only makes sense if https://github.com/libgit2/libgit2/pull/3918 is merged and ASan is enabled (meaning that the fix can be missing and it would still pass!).

tests/odb/mixed.c
```diff
@@ -261,5 +261,20 @@ void test_odb_mixed__expand_ids(void)
 	cl_git_pass(git_odb_expand_ids(_odb, ids, num));
 	assert_notfound_objects(ids);
 	git__free(ids);
-}
 
+	/* test looking for the actual (correct) types after accessing the object */
```
pks-t: I'd argue that all the previous tests should be split into their own functions. While it's fine to leave existing code as-is, I'd prefer to at least do the right thing with this new test.

src/odb.c
```diff
@@ -1124,6 +1124,7 @@ static int odb_otype_fast(git_object_t *type_p, git_odb *db, const git_oid *id)
 
 	if ((object = git_cache_get_raw(odb_cache(db), id)) != NULL) {
 		*type_p = object->cached.type;
+		git_odb_object_free(object);
```
pks-t: Yup. `git_cache_get_raw` calls `cache_get`, which in turn calls `git_cached_obj_incref` on the returned entry. So it's the right thing to free the object here.

tests/odb/mixed.c
lhchavez: Done

pks-t:
Thanks a lot, @lhchavez!

PR #4990, opened 2019-02-17
Remove `git_time_monotonic`

ethomson:
`git_time_monotonic` was added so that non-native bindings like rugged
could get high-resolution timing for benchmarking.  However, this is
outside the scope of libgit2 *and* rugged decided not to use this
function in the first place.

Google suggests that absolutely _nobody_ is using this function and we
don't want to be in the benchmarking business.  Remove the function.


pks-t:
More code removal 🎉

PR #4991, opened 2019-02-18
Remove public 'inttypes.h' header

ethomson:
Remove an `inttypes.h` header that is too large in scope, and far too
public.

For Visual Studio 2012 and earlier (ie, `_MSC_VER < 1800`), we do need
to include `stdint.h` in our public headers, for types like `uint32_t`.

Internally, we also need to define `PRId64` as a printf formatting
string.


ethomson:
Tested on Visual Studio 2012 (aka `_MSC_VER = 1700`).

src/cc-compat.h
```diff
@@ -48,6 +48,11 @@
 /* Define the printf format specifer to use for size_t output */
 #if defined(_MSC_VER) || defined(__MINGW32__)
 
+/* Visual Studio 2012 and prior lack PRId64 entirely */
+#	if (_MSC_VER < 1800)
```
pks-t: Wouldn't it be easier to just check if `PRId64` is defined instead of relying on the `_MSC_VER`? No idea though if MSVC is using e.g. a typedef, but I guess you'd know :)

neithernut: How would you implement `PRId64` using a _typedef_ (in C)? AFAIK the only way to implement it is by using preprocessor-macros.

neithernut: How would you implement `PRId64` using a _typedef_ (in C)? AFAIK the only way to implement it is by using preprocessor-macros.

pks-t:
On Wed, Feb 20, 2019 at 02:27:34AM -0800, Julian Ganz wrote:
> neithernut commented on this pull request.
> > @@ -48,6 +48,11 @@
>  /* Define the printf format specifer to use for size_t output */
>  #if defined(_MSC_VER) || defined(__MINGW32__)
>  
> +/* Visual Studio 2012 and prior lack PRId64 entirely */
> +#	if (_MSC_VER < 1800)
> 
> How would you implement `PRId64` using a _typedef_ (in C)?
> AFAIK the only way to implement it is by using
> preprocessor-macros.

Oops. Due to reading through the "inttypes.h" header I've somehow
thought of `int64_t` etc.

So given that this is a clear "No, this is not implemented via a
typedef": wouldn't it make sense to just use an `#ifndef PRId64`
instead of intimately checking the MSVC version?

ethomson:
> So given that this is a clear "No, this is not implemented via a typedef": wouldn't it make sense to just use an `#ifndef PRId64` instead of intimately checking the MSVC version?

Agreed.  Fixed!

pks-t:
Cool, thanks for this cleanup!

PR #4992, opened 2019-02-20
cache: fix misnaming of `git_cache_free`

pks-t:
Functions that free a structure's contents but not the structure
itself shall be named `dispose` in the libgit2 project, but the
function `git_cache_free` does not follow this naming pattern.

Fix this by renaming it to `git_cache_dispose` and adjusting all
callers to make use of the new name.


pks-t:
Rebased to fix semantic conflict with dd45539d5470bd9a5747118067d9da0403922252

Issue #4654, opened 2018-05-24
Rename git_push_transfer_progress to git_push_transfer_progress_cb


rcjsuen:
Since `git_push_transfer_progress` is a callback, could it be renamed to `git_push_transfer_progress_cb` instead? NodeGit's code generator currently does not flag it as a callback because of this (see nodegit/nodegit#1500) The other members in the `git_remote_callbacks` struct follow the same convention like `git_cred_acquire_cb`, `git_transfer_progress_cb`, `git_push_update_reference_cb`, and so on.

But I guess this ship has long sailed already?

### Version of libgit2 (release number or SHA1)
Valid in `HEAD` https://github.com/libgit2/libgit2/blob/f9cf9a04ba61cba7b1112764dd25c0e4638bed75/include/git2/remote.h#L460-L466

https://github.com/libgit2/libgit2/blob/f9cf9a04ba61cba7b1112764dd25c0e4638bed75/include/git2/remote.h#L350-L355

pks-t:
Yeah, this is misnamed. But due to API stability, we unfortunately cannot fix it anymore :/ Thanks for your report, but I'm afraid we have no way to fix this anymore. Sorry!

Uncommon:
Perhaps there could be an additional typedef so that both names are available?

rcjsuen:
Given #4997, would it be possible to add another typedef per @Uncommon's suggestion?

pks-t:
Given our new deprecation strategy I'm definitely not opposed to this

PR #4997, opened 2019-02-21
Rename git_transfer_progress to git_indexer_progress

ethomson:
While going through the API, I found it strange that object database functions took a "transfer progress" callback.  In fact, this is a callback for _indexer_ progress that happened to start life as reporting for transfers.

Rename it as such, updating the internal usages and creating a typedef for the old names for backward compatibility.

### DIFF
--- a/examples/clone.c
+++ b/examples/clone.c
@@ -1,7 +1,7 @@
 #include "common.h"
 
 typedef struct progress_data {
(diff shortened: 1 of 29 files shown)


ethomson:
Sure, I added a similar change for #4654.

ethomson:
> Sorry for breaking your PR 🙄

No worries; easy fix.

PR #4998, opened 2019-02-21
Allocator restructuring

pks-t:
I've been annoyed by our two allocators being located at seemingly arbitrary locations. Instead, I propose to move them into a common "allocators/" subdirectory, similar to our transports or streams.


pks-t:
Fixed compilation with MSVC

PR #4999, opened 2019-02-21
cmake: enable -Wundef warning on GCC/Clang

pks-t:
Previously, we have been mostly using `#ifdef` to check whether a
macro is defined or not. Most importantly, this also included
checks for whether a certain feature is enabled via our
"features.h" header. The problem here is that in case where the
checked macro does not exist, we cannot really tell why:

- The feature might simply be unset and thus not available. This
  is what we actually want to check.

- There might be a typo in the check, which causes it to fail. We
  have no way of telling this if we use `#ifdef`.

- The feature define might have been removed. Same as previous
  case, we have no way of telling this as there will be no
  warning.

So this sucks, but surely we can help this situation by using
`#if` instead. Problem though is that in case where a macro is
undefined, compilers will default to "0". So again, we cannot
tell.

To remedy that situation, both GCC and Clang support the
"-Wundef" switch, which prints a warning as soon as we `#if` an
undefined preprocessor variable. As we have converted all macros
to be always set to either "0" or "1", this will now alert us of
typos and removed preprocessor variables.

---

This obviously still needs some cleanups


pks-t:
Not much discussion here, so I'm dropping this for now :)

Issue #4938, opened 2019-01-15
git_branch_lookup should accept GIT_BRANCH_ALL


eaigner:
Unfortunately `git_branch_lookup` does not support `GIT_BRANCH_ALL` as branch type. It seems that there should be no reason for this, since it just restricts the input options for the branch type and basically makes it a programming error if `GIT_BRANCH_ALL` is supplied.

Wouldn't it be more useful and less error prone to support `GIT_BRANCH_ALL` and just try to look up local branches first and if nothing is found, search for a remote branch?

ethomson:
🤷‍♂️  I don't see any reason why not.

ethomson:
Fixed via #5000

PR #5000, opened 2019-02-21
Have git_branch_lookup accept GIT_BRANCH_ALL

augfab:
This PR is a proposed fix for issue #4938.

### DIFF
--- a/src/branch.c
+++ b/src/branch.c
@@ -22,7 +22,7 @@ static int retrieve_branch_reference(
 	git_reference **branch_reference_out,
 	git_repository *repo,
 	const char *branch_name,
-	int is_remote)
+	bool is_remote)
 {
 	git_reference *branch = NULL;
 	int error = 0;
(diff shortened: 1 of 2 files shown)


src/branch.c
```diff
@@ -334,9 +334,23 @@ int git_branch_lookup(
...
 
-	return retrieve_branch_reference(ref_out, repo, branch_name, branch_type == GIT_BRANCH_REMOTE);
+	switch (branch_type) {
+	case GIT_BRANCH_LOCAL:
+	case GIT_BRANCH_REMOTE:
+		error = retrieve_branch_reference(ref_out, repo, branch_name, branch_type == GIT_BRANCH_REMOTE);
+		break;
+	case GIT_BRANCH_ALL:
+		error = retrieve_branch_reference(ref_out, repo, branch_name, false);
+		if (error < 0)
```
ethomson: We should probably only retry when `error == GIT_ENOTFOUND`, otherwise we might hide potentially fatal errors by ignoring them and doing something else.

src/branch.c
```diff
@@ -334,9 +334,23 @@ int git_branch_lookup(
...
+			error = retrieve_branch_reference(ref_out, repo, branch_name, true);
+		break;
+	default:
+		abort();
```
ethomson: `abort` is probably a bit harsh here.  I think `assert(0)` is a reasonable compromise that we've used in the past.

tests/refs/branches/lookup.c
```diff
@@ -23,17 +23,22 @@ void test_refs_branches_lookup__cleanup(void)
 void test_refs_branches_lookup__can_retrieve_a_local_branch(void)
 {
 	cl_git_pass(git_branch_lookup(&branch, repo, "br2", GIT_BRANCH_LOCAL));
+	cl_git_pass(git_branch_lookup(&branch, repo, "br2", GIT_BRANCH_ALL));
+	cl_git_fail(git_branch_lookup(&branch, repo, "br2", GIT_BRANCH_REMOTE));
 }
```
ethomson: You're leaking memory here, namely the results of the prior `git_branch_lookup` that are hidden with the new assignment to `branch`.  Probably best to just make each one of these calls their own test function

augfab:
Thanks a lot @ethomson for the review!
I'll submit a new branch soon.

tests/refs/branches/lookup.c
augfab: I was scratching my head about that, thanks for the explanation :slightly_smiling_face:.

ethomson:
Awesome, thanks @augfab for the PR!

PR #5010, opened 2019-02-28
ci: use sh not bash

ethomson:
There's nothing bash-y about our build and test scripts.  Use sh instead
of bash for bash-deprived systems like FreeBSD.


ethomson:
Gah, forget it, I'll just install bash on FreeBSD.

PR #5012, opened 2019-03-05
[Doc] Update URL to git2-rs

dritter:
The old URL points to https://github.com/alexcrichton/git2-rs, which is redirected to https://github.com/rust-lang/git2-rs. This fix avoids the recirect.


ethomson:
Nice, thanks!

Issue #4230, opened 2017-05-02
git_status_list is slower than `git status`


powerman:
### Reproduction steps

Build provided `libgit2status` tool (I'll post it source in the next comment to keep initial report easier to read) and run it in the root of large enough repo. Compare it speed with `git status`.

```
% go version
go version go1.8.1 linux/amd64
% git --version
git version 2.10.2
% git clone https://github.com/gentoo/gentoo.git
% cd gentoo
% find -type f | wc -l
99540
% find -type d | wc -l
27306
% git config core.untrackedCache false
% time git status --short

0.257 real  0.128 user  0.206 sys  32MB RAM
% git config core.untrackedCache true
% time git status --short

0.097 real  0.067 user  0.110 sys  38MB RAM
% time libgit2status
{AddedFiles:0 ModifiedFiles:0 DeletedFiles:0 RenamedFiles:0 UnmergedFiles:0 HasUntrackedFiles:false}

0.626 real  0.340 user  0.282 sys  154MB RAM
% touch profiles/base/testfile.added
% touch profiles/base/testfile.new
% git add profiles/base/testfile.added
% time git status --short
A  profiles/base/testfile.added
?? profiles/base/testfile.new

0.115 real  0.082 user  0.108 sys  38MB RAM
% git config core.untrackedCache false
% time git status --short
A  profiles/base/testfile.added
?? profiles/base/testfile.new

0.255 real  0.134 user  0.193 sys  32MB RAM
% time libgit2status
{AddedFiles:1 ModifiedFiles:0 DeletedFiles:0 RenamedFiles:0 UnmergedFiles:0 HasUntrackedFiles:true}

0.611 real  0.358 user  0.250 sys  154MB RAM
```

### Expected behavior

Well… in my dream git2go should works at least as fast as `git status`… actually faster if we just needs to know is workdir is dirty or is it has any untracked files - without needs to know full list of all added/modified/untracked/etc. files output by `git status`.

### Actual behavior

It's nearly 2.5 times slower than `git status` on my system (HDD, i7-2600K overcloked @ 4.5GHz) without untrackedCache and 5-6 times slower with untrackedCache.

While strace-ing `git status` I've noticed it calls `lstat` from 2 or 3 threads - can this be the reason?

For all `time`d runs above I take average result from several runs and my system has enough RAM to keep dircache for the full repo, so there was no (noticeable, at least) disk accesses, and thus HDD/SSD shouldn't make any difference.

[...]

PR #5018, opened 2019-03-14
Optimize string comparisons

romkatv:
See https://github.com/libgit2/libgit2/issues/4230#issuecomment-471710359 for context. This is diff optimization PR 1/3. The smallest of the three both in terms of code delta and impact.

Three hand-rolled string comparison functions -- git__prefixcmp, git__strcmp and git__strncmp -- are showing up on the CPU profile when scanning working directory of chromium repository on Linux. This PR optimizes git__prefixcmp and replaces the other two functions with standard equivalents (unless `GIT_WIN32` is defined).

Test:

```zsh
cd ~/libgit2
mkdir build
cd build
cmake \
  -DUSE_ICONV=OFF           \
  -DBUILD_CLAR=ON           \
  -DUSE_SSH=OFF             \
  -DUSE_HTTPS=OFF           \
  -DTHREADSAFE=OFF          \
  -DBUILD_SHARED_LIBS=OFF   \
  -DUSE_EXT_HTTP_PARSER=OFF \
  -DUSE_BUNDLED_ZLIB=ON     \
  -DBUILD_EXAMPLES=ON       \
  -DCMAKE_BUILD_TYPE=Release ..
make -j 20
./libgit2_clar
```

Benchmark:

```zsh
git clone https://github.com/chromium/chromium.git ~/chromium
cd ~/chromium
~/libgit2/build/examples/lg2 status >/dev/null &&
  time (repeat 100; do ~/libgit2/build/examples/lg2 status >/dev/null; done)
```

  * Baseline: 112.11s user 62.75s system 99% cpu 2:55.15 total
  * With this PR: 104.80s user 62.93s system 99% cpu 2:47.96 total

A modest 4% improvement. Working directory traversal increases more than that. The benchmark spends a lot of time doing other things, so the apparent impact of the PR is diluted.

Note:

  * Built and tested only on Ubuntu 18.04 with gcc 8.2. I didn't even attempt to compile on Windows.
  * Benchmarked only on chromium repository on a single machine running Ubuntu 18.04.

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -150,12 +150,13 @@ extern int git__bsearch_r(
 	void *payload,
 	size_t *position);
 
+#define git__strcmp strcmp
+#define git__strncmp strncmp
+
 extern int git__strcmp_cb(const void *a, const void *b);
 extern int git__strcasecmp_cb(const void *a, const void *b);
 
-extern int git__strcmp(const char *a, const char *b);
 extern int git__strcasecmp(const char *a, const char *b);
-extern int git__strncmp(const char *a, const char *b, size_t sz);
 extern int git__strncasecmp(const char *a, const char *b, size_t sz);
 
 extern int git__strcasesort_cmp(const char *a, const char *b);
--- a/src/util.c
+++ b/src/util.c
@@ -204,13 +204,6 @@ int git__strntol32(int32_t *result, const char *nptr, size_t nptr_len, const cha
 	return error;
 }
 
-int git__strcmp(const char *a, const char *b)
-{
-	while (*a && *b && *a == *b)
-		++a, ++b;
-	return (int)(*(const unsigned char *)a) - (int)(*(const unsigned char *)b);
-}
-
 int git__strcasecmp(const char *a, const char *b)
 {
 	while (*a && *b && git__tolower(*a) == git__tolower(*b))
@@ -238,15 +231,6 @@ int git__strcasesort_cmp(const char *a, const char *b)
 		return (unsigned char)git__tolower(*a) - (unsigned char)git__tolower(*b);
 
 	return cmp;
-}
-
-int git__strncmp(const char *a, const char *b, size_t sz)
-{
-	while (sz && *a && *b && *a == *b)
-		--sz, ++a, ++b;
-	if (!sz)
-		return 0;
-	return (int)(*(const unsigned char *)a) - (int)(*(const unsigned char *)b);
 }
 
 int git__strncasecmp(const char *a, const char *b, size_t sz)
@@ -301,7 +285,18 @@ GIT_INLINE(int) prefixcmp(const char *str, size_t str_n, const char *prefix, boo
 
 int git__prefixcmp(const char *str, const char *prefix)
 {
-	return prefixcmp(str, SIZE_MAX, prefix, false);
+	unsigned char s, p;
+
+	while (1) {
+		p = *prefix++;
+		s = *str++;
+
+		if (!p)
+			return 0;
+
+		if (s != p)
+			return s - p;
+	}
 }
 
 int git__prefixncmp(const char *str, size_t str_n, const char *prefix)


src/util.h
```diff
@@ -150,12 +150,34 @@ extern int git__bsearch_r(
...
+#ifdef GIT_WIN32
+	while (sz && *a && *b && *a == *b)
+		--sz, ++a, ++b;
+	if (!sz)
+		return 0;
+	return (int)(*(const unsigned char *)a) - (int)(*(const unsigned char *)b);
+#else
+	return strncmp(a, b, sz);
+#endif
+}
```
pks-t: I'd be surprised if the libc `git_strcmp` would actually be any faster. Also, our `git__strncmp` implementation is nearly the same as that of musl libc, so there's no performance gain to be expected compared to that. That being said, glibc uses an unrolled loop for `strncmp` while `n > 4` to enable the compiler to vectorize code, which is probably why you do see performance differences.

So because of this I'm fine with your proposal in general. Personally, I'd vote for having defines though instead of inlined code. E.g.

```
#ifdef GIT_WIN32
GIT_INLINE(int) git__strcmp(const char *a, const char *b)
{
    ....
}
#else
# define git__strcmp strcmp
#endif
```

By the way, why can't we use `strcmp` and `strncmp` on Win32 platforms?

neithernut: According to @ethomson, `strcmp` does locale-sensitive comparision on windows: https://github.com/libgit2/libgit2/issues/4230#issuecomment-471724613

ethomson: I need to validate that this is still happening...  this started in a prerelease of the 2015 (IIRC) C runtime.  It's very possible that the change got reverted before it shipped and we've been unnecessarily shipping our own `strcmp` for years.

I haven't looked at the assembler output but yeah, I assume that there's vectorization opportunities that we've not taken.  In any case, let me look into this in case we can ⚡️ it on win32 also.  I'll 👀 this weekend (sorry, busy day today).

romkatv: > I'd be surprised if the libc `git_strcmp` would actually be any faster.

It speeds up my benchmark by 0.85%.

> So because of this I'm fine with your proposal in general. Personally, I'd vote for having defines though instead of inlined code.

I usually avoid having the same symbol being a function in one configuration and a macro in another. It can cause trouble when signatures diverge, or when there is hiding involved. And in general fewer macros is better.

However, I'll do anything you ask me to. Shall I change the code the way you suggested?

>  I'll eyes this weekend (sorry, busy day today).

@ethomson I'm waiting for you to come back. It's not urgent. Just so you know that the ball is in your court.

pks-t: >> I'd be surprised if the libc git_strcmp would actually be any faster.
> It speeds up my benchmark by 0.85%.

Huh, I wonder why that is. I guess you didn't take a lot at the generated assembly, right? No need to do so if you didn't, I'm simply curious.

> However, I'll do anything you ask me to. Shall I change the code the way you suggested?

I'd say we should first wait for @ethomson. No need to make you perform additional work when the requirements aren't quite clear yet :)

romkatv:
> I guess you didn't take a lot at the generated assembly, right?

Right, I haven't look at the assembly. This PR was produced with the following methodology.

1. Run `time lg2 status` on chromium repo.
2. Generate CPU profile for `lg2 status` in chromium repo. I used perf + perf_data_converter + pprof for this.
3. Take one of the hand-rolled string comparison functions from the top CPU consumers and replace it with a standard function.
4. Run `time lg2 status` on chromium repo. If it's faster than before, keep the change. Otherwise revert it.
5. Repeat.

src/util.h
ethomson: Quick & dirty test of the existing `git__strcmp` vs `strcmp` on win32 suggests that win32's `strcmp` is nice and fast again.  😌

Methodology: generate 10000 strings of random characters (a-zA-Z0-9) of a fixed length (32 characters each).

```
for (i = 0; i < 10000; i++)
    for (j = i; j < 10000; j++)
        strcmp(strings[i], strings[j]);
```

Do this string generation / string comparison 1000 times in a loop.

The `strcmp` takes 66% as long as the `git__strcmp` execution time.

romkatv: Should I remove git__strcmp and git__strncmp and replace all calls to them with the standard equivalents?

pks-t: Mhh. I don't think the code churn would be worth it. Instead, I'd just remove the self-rolled implementations and unconditionally `#define git__strcmp strcmp`

romkatv: Done. And after much trial and error I managed to merge upstream changes and squash my commits into one.

pks-t: Thanks! @ethomson, any remarks?

ethomson: I think the `#define`s are okay.  IIRC, in the past we could not simply do this since we take function pointers to `git__strcmp` (in `status.c` for example) so that we can switch easily between `git__strcmp` and `git__strcasecmp`.  The problem is that we run into calling convention problems, depending on how we're built vs how the MSVCRT runtime is built.

It looks like since we removed the ability to build as `stdcall` that we're :+1: here now.

ethomson:
Thanks for the fix @romkatv!

Issue #4963, opened 2019-01-28
Negate patterns in .gitignore


buddyspike:
I have a repository with following structure

```
/
  .gitignore
  subdir/
    .gitignore
```

In the root `.gitignore` file there's a negate rule `!*[Cc]ache/`.
`.gitignore` file in `subdir` has the opposite rule `.cache/`

With that setup, if I create a file `subdir/.cache/foo`, `git status` does not report it. However, it is reported when invoking `git_status_list_new` with `GIT_DIFF_INCLUDE_UNTRACKED` option. 

Is that behaviour normal?

BTW: This happens in `v27.x`, `v26` is fine.

pks-t:
Which version of libgit2 are you using? The reason why I'm asking
is because we fixed an issue that sounds exactly like yours in
#4670. This PR is not yet part of any release, but it will be
included in v0.28.0 and the upcoming v0.27.8.

buddyspike:
I'm on `0.27.7`. I will wait until `0.27.8` is out. 
Thanks for the awesome work! 😸

ethomson:
Hey @buddyspike, I just dropped [v0.27.8](https://github.com/libgit2/libgit2/releases/tag/v0.27.8).  Can you see if this fixes your issue?

buddyspike:
Unfortunately, my test is still failing.
Interestingly, the rule only breaks when I introduced `[Cc]` or `*` in the parent `.gitignore`. 
If it has just `!.cache/`, it works (even in `0.27.7`).

buddyspike:
Following the changes in the commit, I wrote this test:

```c

void test_status_ignore__override_negate_rule(void)
{
	static const char *test_files[] = {
		"empty_standard_repo/.cache/a.test",
		"empty_standard_repo/subdir/.cache/b.test",
		NULL
	};

	make_test_data("empty_standard_repo", test_files);
	cl_git_mkfile(
		"empty_standard_repo/.gitignore",
		"!*.[Cc]ache/\n");

	cl_git_mkfile(
		"empty_standard_repo/subdir/.gitignore",
		".cache/\n");

	assert_is_ignored("subdir/.cache/b.test");
	refute_is_ignored(".cache/a.test");
}
```

It passes. I wonder if the problem is somewhere else 😉.

pks-t:
On Tue, Jan 29, 2019 at 09:49:47AM +0000, Buddhike de Silva wrote:
> Following the changes in the commit, I wrote this test:
> 
[...]

PR #5019, opened 2019-03-14
Direct path negation of files under ignored directories are ignored

implausible:
.gitignore seems to break in odd ways with negation patterns. Currently, I know for sure that this pattern is good:
```
dir/
!dir/subdir/*
```
Where `dir/*` is ignored, but `dir/subdir/specific-file.test` is not ignored.

However this pattern is bad:
```
dir/
!dir/subdir/specific-file.test
```
`dir/subdir/specific-file.test` is ignored, even though it's very explicitly not ignored.

This may be related to https://github.com/libgit2/libgit2/issues/4963.


implausible:
This is expected behavior, woops!

Issue #5017, opened 2019-03-13
support merge analysis for bare repos


rcoup:
### Reproduction steps

Call `git_merge_analysis()` or `git_merge_analysis_for_ref()` on a bare repository.

### Expected behaviour

Since we can do `git_merge_trees()` and other things on bare repositories, makes sense analysis should work too since it's non-destructive. From Slack:

> @ethomson
> it doesn't seem like that's something that should need a workdir for.
> i suspect that some refactoring just ended up shuffling the not-bare check there.

### Actual behaviour

It errors with `cannot merge. This operation is not allowed against bare repositories`. The bare repository check appears in `merge_heads()`:

https://github.com/libgit2/libgit2/blob/2e0a304839764236654e73d38fa380b317a3fac1/src/merge.c#L3115-L3116

### Version of libgit2 (release number or SHA1)

0.28.1 @ 2e0a304839764236654e73d38fa380b317a3fac1

### Operating system(s) tested

OSX

rcoup:
Looks like [`git_merge()` already has a not-bare-repo check](https://github.com/libgit2/libgit2/blob/2e0a304839764236654e73d38fa380b317a3fac1/src/merge.c#L3275), are there any other merging functions that actually need the guard, or could it simply be removed from `merge_heads()`?

ethomson:
> Looks like git_merge() already has a not-bare-repo check, are there any other merging functions that actually need the guard, or could it simply be removed from merge_heads()?

Looking at the history, this came in with #2183, so I think it was an effort to make `git_merge_status` fail in bare repositories.

Now, this raises another question: why do we really need _that_ to be non-bare?  I don't think it does.  It's possible that at some point in time it read `MERGE_HEAD` or something but I think it's okay to remove the bare check altogether.

Would you be interested in opening a PR?

ethomson:
Fixed via #5022 - thanks for the report and the PR!

PR #5022, opened 2019-03-19
Merge analysis support for bare repos

rcoup:
Fix for https://github.com/libgit2/libgit2/issues/5017

~For testing, I copied `tests/merge/workdir/analysis.c` into `tests/merge/trees/`, and (manually) converted `tests/resources/merge-resolve/` repo into a bare `tests/resources/merge-resolve.git/` repo.~

For testing, I refactored `tests/merge/workdir/analysis.c` into two matching test suites with a common implementation, one for working dirs and one for bare trees.

- 8/9 new tests fail without the change. The remaining one checks for unborn, which happens before the bare-repo check.
- 9/9 new tests pass with the change
- no change to existing merge tests


rcoup:
> I'm not fond of the duplicate tests, hence some remarks :

Yeah, I wasn't too excited about it either.

> A "this has been duplicated as" comment at the top of both files would be a 👍, just so changes can be propagated.

Can do this easily.

> It might be possible to bare-clone from the other repo when initializing the testcase, but I'm not sure about runtime speed.

Any examples in the test suite for doing this? I guess it's just `git_clone(..., {bare=1})` but I'll need some pointers to figure out the path/tempdir handling in the tests.

> Test case duplication ... I don't think it can be done cleanly without changes to clar 

_backs away slowly_ 🐉🐍 

Alternatively as a simple starting point before clar supports parameterisation I could probably move all the analysis test bodies into `merge/analysis.c` and call them from the `workdir/analysis.c` & `trees/analysis.c` versions with the right setup?

tiennou:
> Any examples in the test suite for doing this? I guess it's just git_clone(..., {bare=1}) but I'll need some pointers to figure out the path/tempdir handling in the tests.

- initialize

```c
#define ANALYZED_REPO "merge-analysis-bare.git"

/* cl_sandbox setup, pretty standard, using the non-bare version */

git_clone_options opts = GIT_CLONE_OPTIONS_INIT;
opts.bare = 1;

git_clone(ANALYZED_REPO, &opts);
```

- cleanup
```c
git_futils_rmdir_r(ANALYZED_REPO);
```

Code mildly brain-compiled. Helpful "low-level" utilities can be found in futils.c/h, path.c/h, good-ol' buffer.c/h, as well as the clar headers.

> Alternatively as a simple starting point before clar supports parameterisation […]

Yeah, I'd much prefer having a clean implementation if that's something we ever want. For now, a documentation comment is sufficient in my eyes. Again, thanks for doing this !

pks-t:
You could just use `git_repository_open_ext(&repo, "/path/to/sandbox/.git", GIT_REPOSITORY_OPEN_BARE, NULL);` instead of doing a copy.

rcoup:
@tiennou sorry, been distracted by some other work. I've refactored the analysis test implementation into one file and removed the duplicate test data (I went with @pks-t's suggestion).

tests/merge/analysis.c
```diff
@@ -0,0 +1,170 @@
+// NOTE: this is the implementation for both merge/trees/analysis.c and merge/workdir/analysis.c
+// You probably want to make changes to both files.
```
pks-t: We do not allow C++ style comments in our codebase due to C89.

tests/merge/analysis.c
```diff
@@ -0,0 +1,170 @@
...
+
+#include "clar_libgit2.h"
+#include "git2/repository.h"
+#include "git2/merge.h"
+#include "git2/annotated_commit.h"
+#include "git2/sys/index.h"
+#include "merge.h"
+#include "./merge_helpers.h"
```
pks-t: There's no need to have the "./" prefix here

tests/merge/workdir/analysis.c
```diff
@@ -1,170 +1,75 @@
...
-#define NOFASTFORWARD_ID		"7cb63eed597130ba4abb87b3e544b85021905520"
-
 
 /* Fixture setup and teardown */
 void test_merge_workdir_analysis__initialize(void)
 {
 	repo = cl_git_sandbox_init(TEST_REPO_PATH);
 	git_repository_index(&repo_index, repo);
+
+    testimpl_merge_analysis__initialize(repo, repo_index);
```
pks-t: Please use tabs instead of spaces to indent code

tests/merge/analysis.h
```diff
@@ -0,0 +1,16 @@
...
+void testimpl_merge_analysis__cleanup(void);
+void testimpl_merge_analysis__fastforward(void);
+void testimpl_merge_analysis__no_fastforward(void);
+void testimpl_merge_analysis__uptodate(void);
+void testimpl_merge_analysis__uptodate_merging_prev_commit(void);
+void testimpl_merge_analysis__unborn(void);
+void testimpl_merge_analysis__fastforward_with_config_noff(void);
+void testimpl_merge_analysis__no_fastforward_with_config_ffonly(void);
+void testimpl_merge_analysis__between_uptodate_refs(void);
+void testimpl_merge_analysis__between_noff_refs(void);
```
pks-t: I'm not really all that happy with this approach, but I guess it's the best we can do right now with what clar gives us. We probably should extend clar with some kind of logic that allows us to do data-driven tests like @tiennou proposes, ideally with not much change at all to how the rest of our tests  work. Meanwhile, this is fine with me.

rcoup:
I addressed all those.

I can't quite spot my error which is triggering valgrind though. Any obvious insight?

tests/merge/analysis.c
```diff
@@ -0,0 +1,172 @@
...
+	git_buf_dispose(&master);
+}
+
+void testimpl_merge_analysis__fastforward_with_config_noff(void)
+{
+	git_config *config;
+	git_merge_analysis_t merge_analysis;
+	git_merge_preference_t merge_pref;
+
+	git_repository_config(&config, repo);
```
ethomson: You're not `free`ing this...

ethomson:
> I can't quite spot my error which is triggering valgrind though. Any obvious insight?

Not sure offhand.  But something is probably incrementing the refcount on the repository, so your call to `repository_free` isn't actually disposing it (just reducing the refcount).

My guess is that it's that `git_repository_config` but I'm not 100%.

tests/merge/analysis.c
```diff
@@ -0,0 +1,172 @@
...
+	git_config_set_string(config, "merge.ff", "false");
```
pks-t: You'd also have to wrap all these calls in `cl_git_pass` in case where they return an error.

tests/merge/analysis.c
```diff
@@ -0,0 +1,172 @@
...
+	cl_assert_equal_i(GIT_MERGE_PREFERENCE_NO_FASTFORWARD, (merge_pref & GIT_MERGE_PREFERENCE_NO_FASTFORWARD));
+}
+
+void testimpl_merge_analysis__no_fastforward_with_config_ffonly(void)
+{
+	git_config *config;
+	git_merge_analysis_t merge_analysis;
+	git_merge_preference_t merge_pref;
+
+	git_repository_config(&config, repo);
```
pks-t: You're not freeing this

rcoup:
Aha, it had passed CI before my test refactoring which had confused me, I assumed it was related to the init/reorg. _\<looking in the wrong place\>_

OT: Valgrind on OSX (3.14.0/Homebrew) reports a pile of errors, even when using `tests/valgrind-supp-mac.txt` & `libgit2_clar.supp`, looks like mostly from dyld/ImageLoader. And doesn't report _these_ test leaks, with or without the suppressions. Built with `-DVALGRIND=ON`. There some other spells I need to cast to make it work?

rcoup:
Rebased onto master

ethomson:
> OT: Valgrind on OSX (3.14.0/Homebrew) reports a pile of leaks, even when using tests/valgrind-supp-mac.txt & libgit2_clar.supp, looks like mostly from dyld/ImageLoader. And doesn't report these test leaks, with or without the suppressions. Built with -DVALGRIND=ON. There some other spells I need to cast to make it work?

Although we checked in `valgrind-supp-mac.txt`, I haven't made valgrind work correctly on macOS in years.  (valgrind itself says that the latest supported version is macOS 10.12.)

We use `leaks` on macOS if you have Xcode installed (which I presume that you do).  You can take a look at the [CI test script](https://github.com/libgit2/libgit2/blob/master/ci/test.sh) to see how we invoke it:

```
MallocStackLogging=1 MallocScribble=1 MallocLogFile=/dev/null CLAR_AT_EXIT=\"leaks -quiet \$PPID\" libgit2_clar
```

rcoup:
> We use leaks on macOS if you have Xcode installed (which I presume that you do). You can take a look at the CI test script to see how we invoke it:

@ethomson aha, victory! that works (I've never used `leaks` before, only valgrind). Is it worth me adding that to docs somewhere? `tests/README`?

ethomson:
> Is it worth me adding that to docs somewhere? tests/README?

Whoa, that would be great.  🙇

rcoup:
@pks-t you happy for this to be merged now?

pks-t:
Looks good to me now, thanks for your work @rcoup! I'll do a follow-up to make use of our new test variants so that we can avoid the duplication, but seeing as that only got merged a few days ago I don't want to shift that burden on you.

PR #5023, opened 2019-03-20
Correctly write to missing locked global config

ianhattendorf:
Opening a default config when ~/.gitconfig doesn't exist, locking it,
and attempting to write to it causes an assertion failure.

Treat non-existent global config file content as an empty string.


ianhattendorf:
Good to hear, just pushed up the rebased changes.

pks-t:
Thanks for resolving the conflict and thanks for your fix in general!

PR #5025, opened 2019-03-21
Include implicitly used header

yursha:
When compiling under `clang -std=c11` the following warning is triggered:

```
blame.c:170:13: warning: implicit declaration of function 'strcasecmp' is invalid in C99 [-Wimplicit-function-declaration]
                else if (!strcasecmp(a, "-M"))
                          ^
1 warning generated.
```

Including the header solves the problem.

### DIFF
--- a/examples/blame.c
+++ b/examples/blame.c
@@ -12,6 +12,7 @@
  * <http://creativecommons.org/publicdomain/zero/1.0/>.
  */
 
+#include <strings.h>
 #include "common.h"
 
 #ifdef _MSC_VER


ethomson:
```
D:\a\1\s\examples\blame.c(15): fatal error C1083: Cannot open include file: 'strings.h': No such file or directory [D:\a\1\b\examples\lg2.vcxproj]
```

I think that you'll either need `ifndef MSVC` or maybe use a `CHECK_INCLUDE_FILE` in the examples `CMakeLists.txt`...

PR #5026, opened 2019-03-23
Each hash implementation should define `git_hash_global_init`

tenderlove:
This means the forward declaration isn't necessary.  The forward
declaration can cause compilation errors as it conflicts with the
`GIT_INLINE` declaration (the signatures are different).

/cc @rafeca

### DIFF
--- a/src/hash.h
+++ b/src/hash.h
@@ -31,8 +31,6 @@ void git_hash_ctx_cleanup(git_hash_ctx *ctx);
 # include "hash/hash_generic.h"
 #endif
 
-int git_hash_global_init(void);
-
 typedef struct {
 	void *data;
 	size_t len;
--- a/src/hash/hash_win32.c
+++ b/src/hash/hash_win32.c
@@ -109,21 +109,6 @@ static void git_hash_global_shutdown(void)
 		hash_cryptoapi_prov_shutdown();
 }
 
-int git_hash_global_init(void)
-{
-	int error = 0;
-
-	if (hash_prov.type != INVALID)
-		return 0;
-
-	if ((error = hash_cng_prov_init()) < 0)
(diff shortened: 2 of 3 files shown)


ethomson:
This seems reasonable.  But if you're using anything but SHA1DC then that's a Big Red Flag.

ethomson:
> This seems reasonable. But if you're using anything but SHA1DC then that's a Big Red Flag.

Oh, I see.  The only implementation that _wouldn't_ throw a warning is the win32 implementation since it's the only one that's _not_ `GIT_INLINE`'d?

OK.  Seems reasonable enough.

tenderlove:
> Oh, I see. The only implementation that _wouldn't_ throw a warning is the win32 implementation since it's the only one that's _not_ `GIT_INLINE`'d?

Yep, that's correct.

Looks like the x86 mingw build is failing due to what looks like a network error.  I don't think the failure is related to this change.

ethomson:
Cool.  Thanks!

pks-t:
Yeah, this has also been part of #4438 :+1: So thanks for this, less stuff to review in the other PR ;)

csware:
This commit makes comnpiling libgit2 with VS2017 15.9.10 fail:

`libgit2\src\hash\hash_win32.h(145): error C2065: "hash_prov": undeclared identifier`

ethomson:
@csware I can't reproduce this, I'm building with VS 2017 (15.9.5) with no problem.

ethomson:
@csware Oh, I see.  You must be building with the Win32 SHA1 provider.

First, I would encourage you not to do that.  As Linus mentioned, Git doesn't use SHA1, it uses SHA1DC.  We haven't removed the other options because it doesn't make sense given the impending move to SHA256.

Regardless, I'll push up a fix.

PR #5027, opened 2019-03-23
patch_parse.c: Handle CRLF in parse_header_start

ddevault:
This becomes a segfault later (in `check_prefix`) when parsing patches where a file was added. CRLF seems to be handled fine otherwise. Added test which demonstrates the segfault.

Supporting CRLF is important because patches are often emailed - and emails use CRLF.

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -328,7 +328,8 @@ static int parse_header_start(git_patch_parsed *patch, git_patch_parse_ctx *ctx)
 	 * proceeed here. We then hope for the "---" and "+++" lines to fix that
 	 * for us.
 	 */
-	if (!git_parse_ctx_contains(&ctx->parse_ctx, "\n", 1)) {
+	if (!git_parse_ctx_contains(&ctx->parse_ctx, "\n", 1) &&
+	    !git_parse_ctx_contains(&ctx->parse_ctx, "\r\n", 2)) {
 		git_parse_advance_chars(&ctx->parse_ctx, ctx->parse_ctx.line_len - 1);
 
 		git__free(patch->header_old_path);
--- a/tests/diff/parse.c
+++ b/tests/diff/parse.c
@@ -360,6 +360,7 @@ void test_diff_parse__lineinfo(void)
 	git_diff_free(diff);
 }
 
+
 void test_diff_parse__new_file_with_space(void)
 {
 	const char *content = PATCH_ORIGINAL_NEW_FILE_WITH_SPACE;
@@ -373,6 +374,24 @@ void test_diff_parse__new_file_with_space(void)
 	cl_assert_equal_p(patch->delta->old_file.path, NULL);
 	cl_assert_equal_s(patch->diff_opts.new_prefix, "b/");
 	cl_assert_equal_s(patch->delta->new_file.path, "sp ace.txt");
+
+	git_patch_free(patch);
+	git_diff_free(diff);
+}
+
+void test_diff_parse__crlf(void)
+{
+	const char *text = PATCH_CRLF;
+	git_diff *diff;
+	git_patch *patch;
+	const git_diff_delta *delta;
+
+	cl_git_pass(git_diff_from_buffer(&diff, text, strlen(text)));
+	cl_git_pass(git_patch_from_diff(&patch, diff, 0));
+	delta = git_patch_get_delta(patch);
+
+	cl_assert_equal_s(delta->old_file.path, "test-file");
+	cl_assert_equal_s(delta->new_file.path, "test-file");
 
 	git_patch_free(patch);
 	git_diff_free(diff);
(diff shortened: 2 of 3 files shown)


ethomson:
Why does this segfault later?  It's not obvious to me how this would prevent that, so I'm concerned that there's two issues here (the CRLF issue something causing _else_ causing a segfault).  Can you provide reproduction steps?

ddevault:
I added a test which reproduces the segfault - roll back the change to `patch_parse.c` and run the test suite to reproduce. The issue is that if this branch is taken, it sets the `header_old_path` to `NULL`, assuming that it just parsed some bunk data and hoping the `+++` line will clear things up (per the comments nearby) - in this case, it doesn't, because the `+++` is `/dev/null` - no prefix. The resulting segfault happens here:

https://github.com/libgit2/libgit2/blob/5e88f13eaa8ca30439b496066f6c57e608e18c4e/src/patch_parse.c#L931-L940

Because the file is added (and the left side would be `/dev/null` and not have a prefix), this chooses to use `header_old_path` - which was erroneously set to `NULL` earlier.

eaigner:
@ethomson @ddevault it seems the underlying issue is something else.

I fixed it here https://github.com/libgit2/libgit2/pull/5030

ddevault:
Hm, that doesn't seem related. There are no spaces in the paths in the test case I added, and when I add my test case to your patch it still segfaults.

eaigner:
Yeah, I added the `\r` at the end to test and it segfaults

pks-t:
> Yeah, I added the \r at the end to test and it segfaults

You mean when testing against your #5030? For what it's worth, it does work just fine with #5035 (at least there's no segfault there, the test still fails due to mis-parsing).

eaigner:
It would segfault before and after my patch

src/patch_parse.c
```diff
@@ -328,7 +328,8 @@ static int parse_header_start(git_patch_parsed *patch, git_patch_parse_ctx *ctx)
 	 * proceeed here. We then hope for the "---" and "+++" lines to fix that
 	 * for us.
 	 */
-	if (!git_parse_ctx_contains(&ctx->parse_ctx, "\n", 1)) {
+	if (!git_parse_ctx_contains(&ctx->parse_ctx, "\n", 1)
+			&& !git_parse_ctx_contains(&ctx->parse_ctx, "\r\n", 2)) {
```
pks-t: Style nit:

```
if (!git_parse_ctx_contains(&ctx->parse_ctx, "\n", 1) &&
    !git_parse_ctx_contains(&ctx->parse_ctx, "\r\n", 2))
```

Continuation lines are aligned with the opening brace and the boolean operators belong to the previous line.

tests/diff/parse.c
```diff
@@ -359,3 +359,28 @@ void test_diff_parse__lineinfo(void)
...
+
+void test_diff_parse__crlf(void)
+{
+	const char *text = "diff --git a/test-file b/test-file\r\n"
+	"new file mode 100644\r\n"
+	"index 0000000..af431f2 100644\r\n"
+	"--- /dev/null\r\n"
+	"+++ b/test-file\r\n"
+	"@@ -0,0 +1 @@\r\n"
+	"+a contents\r\n";
```
pks-t: We usually put those into "tests/patch/patch_common.h", but I'm happy either way.

src/patch_parse.c
ddevault: Should I align with a mix of tabs and spaces?

ddevault: And how wide are the tabs? 8 cols?

pks-t:
On Fri, Apr 05, 2019 at 03:18:47PM +0000, Drew DeVault wrote:
> ddevault commented on this pull request.
> 
> 
> 
> > @@ -328,7 +328,8 @@ static int parse_header_start(git_patch_parsed *patch, git_patch_parse_ctx *ctx)
>  	 * proceeed here. We then hope for the "---" and "+++" lines to fix that
>  	 * for us.
>  	 */
> -	if (!git_parse_ctx_contains(&ctx->parse_ctx, "\n", 1)) {
> +	if (!git_parse_ctx_contains(&ctx->parse_ctx, "\n", 1)
> +			&& !git_parse_ctx_contains(&ctx->parse_ctx, "\r\n", 2)) {
> 
> Should I align with a mix of tabs and spaces?

Yes, please. We basically use git.git's style, which is in turn
built on linux.git's style. Tabs are expected to be 8 spaces, so
the above would be indented by 1 tab and 4 spaces:

```
	if (!git_parse_ctx_contains(&ctx->parse_ctx, "\n", 1) &&
	    !git_parse_ctx_contains(&ctx->parse_ctx, "\r\n", 2)) {
```

ddevault:
Conflicts & style issues addressed.

pks-t:
Thanks a lot!

Issue #4338, opened 2017-09-05
Switching from libssh2 to libssh.


mothacehe:
libssh2 does not provide the ability to parse ssh configuration files. As this is a really important feature when dealing with ssh connections, would it be an option to switch to libssh ?

libssh allows configuration file parsing and seem to have a more complete API.

ethomson:
I've used libssh in the past and been reasonably pleased with it.  I think that adding a libssh backend would be useful.

We'd welcome a pull request, but this is not something that I have on my radar.  Is this something that you would like to contribute?

mothacehe:
Sure, I'll start working on a libssh transport, and propose a pull request if everything goes fine !

carlosmn:
One thing libssh does not do is work with Windows' cryptography. You need to build openssl or libgcrypt on there, which reduces its usefulness as something to be shipped in a library that aims for for its code to work cross-platform.

I would recommend implementing the transport on its own, at least at first. There is nothing special about the transports in this code base, and it'll likely be less of a hassle if you don't have to get libgit2's build system to search for two kinds of ssh libraries.

mothacehe:
@carlosmn, what do you mean by "implement the transport on it's own" ? I have a nearly functional transport/ssh-libssh.c based on libssh who shares 90% of it's code with the existing ssh.c based on libssh2.

Shall I let both ssh transports co-exist and add a compile-time option to select one of them ? Or maybe a run-time option ?

An other option would be to select the libssh2 based transport on windows, the libssh based transport on linux and to  factorize the code between those two transports in a ssh_common.c ?

jacquesg:
I would like to see this as a compile-time option. In many cases people are quite happy with using libssh2 and as such switching to libssh is not needed or even undesired especially when dealing with cross-platform systems: If you have to use libssh2 on Windows for example, it makes sense to also use libssh2 on Linux.

ethomson:
> Shall I let both ssh transports co-exist and add a compile-time option to select one of them ?

_Removing_ libssh2 and replacing it with libssh support is probably not something that we want to do.

> Or maybe a run-time option ?

Having to compile against both is almost certainly not something that anybody wants to do.

If we're going to support libssh, then doing so at compile time seems like the only reasonable choice.

carlosmn:
>what do you mean by "implement the transport on it's own" ? 

[...]

PR #5028, opened 2019-03-23
remote: add callback to resolve URLs before connecting

eaigner:
Since libssh2 doesn't read host configuration from the config file,
this callback can be used to hand over URL resolving to the client
without touching the SSH implementation itself.

Example:

Remote `origin` has URL

    ssh://myhost/~/repo.git

This would obviously fail with the current libgit2 implementation,
given the following SSH config:

    Host myhost
        User gituser
        HostName my.very.long.host.with.subdomain
        Port 1234

The URL callback allows this to be rewritten to e.g.

    gituser@my.very.long.host.with.subdomain:1234/~/repo.git

by the client without much hassle.

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -328,7 +328,8 @@ static int parse_header_start(git_patch_parsed *patch, git_patch_parse_ctx *ctx)
 	 * proceeed here. We then hope for the "---" and "+++" lines to fix that
 	 * for us.
 	 */
-	if (!git_parse_ctx_contains(&ctx->parse_ctx, "\n", 1)) {
+	if (!git_parse_ctx_contains(&ctx->parse_ctx, "\n", 1) &&
+	    !git_parse_ctx_contains(&ctx->parse_ctx, "\r\n", 2)) {
 		git_parse_advance_chars(&ctx->parse_ctx, ctx->parse_ctx.line_len - 1);
 
 		git__free(patch->header_old_path);
--- a/tests/diff/parse.c
+++ b/tests/diff/parse.c
@@ -360,6 +360,7 @@ void test_diff_parse__lineinfo(void)
 	git_diff_free(diff);
 }
 
+
 void test_diff_parse__new_file_with_space(void)
 {
 	const char *content = PATCH_ORIGINAL_NEW_FILE_WITH_SPACE;
@@ -373,6 +374,24 @@ void test_diff_parse__new_file_with_space(void)
 	cl_assert_equal_p(patch->delta->old_file.path, NULL);
 	cl_assert_equal_s(patch->diff_opts.new_prefix, "b/");
 	cl_assert_equal_s(patch->delta->new_file.path, "sp ace.txt");
+
+	git_patch_free(patch);
+	git_diff_free(diff);
+}
+
+void test_diff_parse__crlf(void)
+{
+	const char *text = PATCH_CRLF;
+	git_diff *diff;
+	git_patch *patch;
+	const git_diff_delta *delta;
+
+	cl_git_pass(git_diff_from_buffer(&diff, text, strlen(text)));
+	cl_git_pass(git_patch_from_diff(&patch, diff, 0));
+	delta = git_patch_get_delta(patch);
+
+	cl_assert_equal_s(delta->old_file.path, "test-file");
+	cl_assert_equal_s(delta->new_file.path, "test-file");
 
 	git_patch_free(patch);
 	git_diff_free(diff);
--- a/tests/patch/patch_common.h
+++ b/tests/patch/patch_common.h
@@ -850,3 +850,12 @@
 	"+++ b/sp ace.txt\n" \
 	"@@ -0,0 +1 @@\n" \
 	"+a\n"
+
+#define PATCH_CRLF \
+	"diff --git a/test-file b/test-file\r\n" \
+	"new file mode 100644\r\n" \
+	"index 0000000..af431f2 100644\r\n" \
+	"--- /dev/null\r\n" \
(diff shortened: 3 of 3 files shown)


eaigner:
Possibly also resolves #4338

src/remote.c
```diff
@@ -670,18 +670,26 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
...
-const char* git_remote__urlfordirection(git_remote *remote, int direction)
+static const char *git_remote__urlresolve(const char *url, int direction, const git_remote_callbacks *callbacks)
+{
+    if (callbacks != NULL && callbacks->resolve_url != NULL) {
+        return callbacks->resolve_url(url, direction, callbacks->payload);
+    }
+    return url;
+}
+
+const char* git_remote__urlfordirection(git_remote *remote, int direction, const git_remote_callbacks *callbacks)
```
ethomson: minor nit: we put the `*` on the name not on the type.  eg `const char *git_remote__`...

ethomson:
I don't have a bunch of familiarity with the SSH code, so I'd like @carlosmn and / or @pks-t to take a look as well.

One thing that I'm not clear about though is why this is useful.  You mention:

> Remote origin has URL
>
> `ssh://myhost/~/repo.git`
>
> This would obviously fail with the current libgit2 implementation.

But that's not obvious to me at all.  I'm guessing that there's some preconditions that are thus-far implicit.  Like an ssh config that contains (to continue with your example):

```
Host myhost
    User gituser
    HostName my.very.long.host.with.subdomain
    Port 1234
```

And your goal is to read the `ssh_config` yourself to apply the transformations.  Am I thinking about this correctly?

src/remote.c
eaigner: yes, i noticed, but it already was that way (not my code) so i left it on the type.

eaigner:
Yes, that was my goal. To be able to substitute values from the `.ssh/config` file, because `libssh2` doesn't do that at all.

eaigner:
@pks-t Any thoughts on this? Would really need a way to properly handle hosts configured in SSH confs, even though the SSH lib doesn't support it.

eaigner:
I'd rather not maintain a separate libgit2 fork to enable SSH hosts 😕

eaigner:
I'd appreciate just a brief `yes / no /maybe` if this ever has the chance to be merged. I rely on being able to properly resolve SSH hosts. Thanks! /cc @pks-t @carlosmn @ethomson

ethomson:
> I'd appreciate just a brief yes / no / maybe if this ever has the chance to be merged or if there are other plans for properly supporting SSH configs. I rely on being able to properly resolve SSH hosts.

I think those are two separate questions:  yes, I think that this is reasonable and worth merging.  I'll give other reviewers a little more time to 👀 this.

As for "properly supporting SSH configs", this feels like an implementation detail around libssh2.  I'm happy to add a hook that lets you cope with libssh2's limitations, but I don't feel like this moves us into the realm of "properly supporting configs".  So I think that remains an open question.

eaigner:
I was just wondering if there is maybe another roadmap or planned feature that would make this "stopgap" solution obsolete, like a switch to `libssh`.

I would be perfectly fine with this simple callback, since it can handle most of the non-exotic SSH config values like `Port`, `HostName` and `User`

ethomson:
I think that adding support for libssh makes sense, but that isn't a high priority for me, tbqh.  If somebody else wanted to tackle the change then I'd be happy to review it though.

include/git2/remote.h
```diff
@@ -474,6 +474,16 @@ typedef int GIT_CALLBACK(git_push_negotiation)(const git_push_update **updates,
...
 
+/**
+ * Callback to resolve URLs before connecting to remote
+ *
+ * @param url The URL to resolve
+ * @param direction GIT_DIRECTION_FETCH or GIT_DIRECTION_PUSH
+ * @param payload The callback payload
+ * @return The resolved URL
+ */
+typedef const char *GIT_CALLBACK(git_resolve_url_cb)(const char *url, int direction, void *payload);
```
pks-t: This should return an `int` like other callbacks and instead use an out-parameter for the resolved URL. Furthermore, the out-parameter should be non-const. After having it resolved, the caller of `git_resolve_url_cb` should be responsible for freeing it again.

src/remote.c
```diff
@@ -670,18 +670,26 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
 	return set_url(repo, remote, CONFIG_PUSHURL_FMT, url);
 }
 
-const char* git_remote__urlfordirection(git_remote *remote, int direction)
+static const char *git_remote__urlresolve(const char *url, int direction, const git_remote_callbacks *callbacks)
+{
+    if (callbacks != NULL && callbacks->resolve_url != NULL) {
```
pks-t: `if (callbacks && callbacks->resolve_url)`

Same as above, this should return an integer error code so that we may distinguish fatal errors like e.g. OOM situations

src/remote.c
```diff
@@ -670,18 +670,26 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
...
+        return callbacks->resolve_url(url, direction, callbacks->payload);
+    }
+    return url;
+}
+
+const char *git_remote__urlfordirection(git_remote *remote, int direction, const git_remote_callbacks *callbacks)
```
pks-t: This function can now fail, so we need an error code

eaigner:
@pks-t all done

eaigner:
Unfortunately Github auto-closed this PR when I tried to move the PR to another branch on my fork. The new PR is #5056.

PR #5030, opened 2019-03-24
diff: allow headers with spaces in filename

### LINKED ISSUE #5029: git_diff_from_buffer crashes if filenames have spaces in header
Parsing a diff using `git_diff_from_buffer` that has filenames with spaces in its `diff --git` header crashes libgit

Example:

```
#define PATCH_ORIGINAL_NEW_FILE_WITH_SPACE \
	"diff --git a/sp ace.txt b/sp ace.txt\n" \
	"new file mode 100644\n" \
	"index 000000000..789819226\n" \
	"--- /dev/null\n" \
	"+++ b/sp ace.txt\n" \
	"@@ -0,0 +1 @@\n" \
	"+a\n"

void test_diff_parse__new_file_with_space(void)
{
	const char *content = PATCH_ORIGINAL_NEW_FILE_WITH_SPACE;
	git_diff *diff;

	cl_git_pass(git_diff_from_buffer(&diff, content, strlen(content)));
	git_diff_free(diff);
}
```

crashes in `check_prefix:888`.

This diff was generated by adding a new file named `sp ace.txt` and running `git --cached`.

`master`

macOS 10.14.3 (18D109)

eaigner:
Parsing diff headers with spaces in filenames caused libgit2
to crash because it assumed the new path always starts at
a SPACE character.

    diff --git a/sp ace.txt b/sp ace.txt

header_path_len now looks ahead for the sequence "?b/ to
correctly determine the end of the file path.

Closes #5029

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -41,7 +41,19 @@ static int header_path_len(git_patch_parse_ctx *ctx)
 
 	for (len = quoted; len < ctx->parse_ctx.line_len; len++) {
 		if (!quoted && git__isspace(ctx->parse_ctx.line[len]))
-			break;
+			/* lookahead for the sequence "?b/ to allow for spaces in filenames */
+			if (len < ctx->parse_ctx.line_len - 3) {
+				int offset = 0;
+				if (ctx->parse_ctx.line[len+1] == '"') {
+					offset += 1;
+				}
+				if (ctx->parse_ctx.line[len+offset+1] == 'b'
+					&& ctx->parse_ctx.line[len+offset+2] == '/') {
+					break;
+				}
+			} else {
+				break;
+			}
 		else if (quoted && !inquote && ctx->parse_ctx.line[len] == '"') {
 			len++;
 			break;
--- a/tests/diff/parse.c
+++ b/tests/diff/parse.c
@@ -86,6 +86,50 @@ void test_diff_parse__invalid_patches_fails(void)
 	test_parse_invalid_diff(PATCH_CORRUPT_MISSING_HUNK_HEADER);
 }
 
+void test_diff_parse__new_file_with_space(void)
+{
+	const char *text = "diff --git a/sp ace.txt b/sp ace.txt\n"
+		"new file mode 100644\n"
+		"index 000000000..789819226\n"
+		"--- /dev/null\n"
+		"+++ b/sp ace.txt\n"
+		"@@ -0,0 +1 @@\n"
+		"+a\n";
+
+	git_diff *diff;
+	git_patch *patch;
+	const git_diff_delta *delta;
+
+	cl_git_pass(git_diff_from_buffer(&diff, text, strlen(text)));
+	cl_git_pass(git_patch_from_diff(&patch, diff, 0));
+	delta = git_patch_get_delta(patch);
+
+	cl_assert_equal_s(delta->old_file.path, "sp ace.txt");
+	cl_assert_equal_s(delta->new_file.path, "sp ace.txt");
+
+	git_patch_free(patch);
+	git_diff_free(diff);
+}
+
+void test_diff_parse__new_file_with_space_ab_differs(void)
+{
+	const char *text = "diff --git yours/sp ace.txt mine/sp ace.txt\n"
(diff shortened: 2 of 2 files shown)


src/patch_parse.c
```diff
@@ -41,7 +41,19 @@ static int header_path_len(git_patch_parse_ctx *ctx)
...
 	for (len = quoted; len < ctx->parse_ctx.line_len; len++) {
 		if (!quoted && git__isspace(ctx->parse_ctx.line[len]))
-			break;
+			/* lookahead for the sequence "?b/ to allow for spaces in filenames */
+			if (len < ctx->parse_ctx.line_len - 3) {
+				int offset = 0;
+				if (ctx->parse_ctx.line[len+1] == '"') {
+					offset += 1;
+				}
+				if (ctx->parse_ctx.line[len+offset+1] == 'b'
```
tiennou: I'm not that intimate with the diff code, but `b` is not guaranteed to be called `b` (I think there was much sadness about those lines syntax). That's obviously better, but you might want to also not crash on a `diff --git yours/sp ace.txt mine/sp ace.txt` by erroring now (or at end the because there was no `diff --git` line) if we get to `\n` before.

eaigner: Thought as much (I saw some code somewhere where you could name your `a`/`b` anything). Will give it another try.

eaigner: Actually it errors out already if I try `diff --git yours/sp ace.txt mine/sp ace.txt`

```
.F

  1) Failure:
diff::parse::new_file_with_space_oldnew [/Users/era/git/libgit2/tests/diff/parse.c:128]
  Function call failed: (git_diff_from_buffer(&diff, text, strlen(text)))
  error -1 - corrupt new path in git diff header at line 1
```

eaigner: Are you suggesting that it should error, or be a valid diff header (with `yours/.. theirs/..`)?

tiennou: If it errors cleanly I'm 👍 then.

tiennou:
On the topic of actually parsing those patches, I'm not sure if that's in scope…

@jrn, since `git log` shows you're responsible for the parsing support, but marked as `ask` in our `git.git-authors`, would you be okay if inspiration was taken from your code ?

src/patch_parse.c
ethomson: We should be able to parse patches that have custom src and dst path names.  This patch should be valid and parseable:

```
diff --git foo/asdf/new file.txt bar/asdf/new file.txt
new file mode 100644
index 0000000..8bd6648
--- /dev/null
+++ bar/asdf/new file.txt
@@ -0,0 +1 @@
+asdf
```

ethomson: The `diff` header line is a garbage-in-garbage-out horror.  You can set `src-prefix` to `a/b b/` and `dst-prefix` to something like `b/a a b/a/b /b` and it's basically impossible to reason about!

But it appears that git apply:

1. Treats anything up to the first slash as the prefix, and anything after the first slash as the filename.  That is to say that given:

```
diff --git foo/asdf/new file.txt bar/asdf/new file.txt
new file mode 100644
index 0000000..8bd6648
--- /dev/null
+++ bar/asdf/new file.txt
```

it's impossible to tell whether this is a patch of `new file.txt` that was produced with `src-prefix=foo/asdf dst-prefix=bar/asdf`, or a patch of `asdf/new file.txt` that was produced with `src-prefix=foo dst-prefix=bar`.

git is unconcerned with such subtlety and always this as a patch of `asdf/new file.txt` so prefixes must stop at the first slash.

2. Copes fine with prefixes that have a space in them.  eg, the following is legal:

```
diff --git my prefix/new file.txt my prefix/new file.txt
new file mode 100644
index 0000000..8bd6648
--- /dev/null
+++ my prefix/new file.txt
```

tiennou: https://marc.info/?l=git&m=112885453302488&w=2 also describe some kind of quoted C-escape system, which would look like this (mostly, this is an earlier name) : https://marc.info/?l=git&m=112905394808887&w=2

pks-t:
See #5035

PR #5032, opened 2019-03-25
docs: clarify relation of safe and forced checkout strategy

### LINKED ISSUE #5031: GIT_CHECKOUT_FORCE/SAFE should not be mutually exclusive
`GIT_CHECKOUT_FORCE` documentation reads

    Allow all updates to force working directory to look like index

and `GIT_CHECKOUT_SAFE`

    Allow safe updates that cannot overwrite uncommitted data

To me that doesn't necessarily mean that they are mutually exclusive, but when I try to check out a tree using `git_checkout_tree` with both flags, `GIT_CHECKOUT_SAFE` seems to be completely ignored and all current changes are lost.

To me the meaning of `GIT_CHECKOUT_SAFE` + `GIT_CHECKOUT_FORCE` would be "force the working directory to look like the index and proceed only if we have no uncommitted data".

Maybe its just a docfix, I don't know what is actually expected to happen in this case.

### Expected behavior

The checkout is aborted with uncommitted changes

### Actual behavior

The checkout proceeds deleting all uncommitted changes

`master`

macOS 10.14.3 (18D109)

eaigner:
Closes #5031

### DIFF
--- a/include/git2/checkout.h
+++ b/include/git2/checkout.h
@@ -106,10 +106,22 @@ GIT_BEGIN_DECL
 typedef enum {
 	GIT_CHECKOUT_NONE = 0, /**< default is a dry run, no actual updates */
 
-	/** Allow safe updates that cannot overwrite uncommitted data */
+	/**
+	 * Allow safe updates that cannot overwrite uncommitted data.
+	 * If the uncommitted changes don't conflict with the checked out files,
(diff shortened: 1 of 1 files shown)


ethomson:
Thanks!  This is a helpful improvement.  👍

Issue #5033, opened 2019-03-28
Cannot parse gitconfig file if it is a symlink


musm:
I'm having issues where  `git_repository_open` will fail if my  user `~/.gitconfig` file is symlinked. 
The error given is
`:Config, failed to parse config file: Invalid configuration key`

ethomson:
What version of libgit2 is exhibiting this behavior?

musm:
Version v0.27.7

pks-t:
I'm unfortunately unable to reproduce your issue, neither on current master nor on v0.27.7. Please have a look whether the tests in #5034 match what you describe

musm:
Thanks. Strange, this might then be a downstream issue with Julia. I thought it was an upstream issue since we just make  a ccall to the shared library function  `git_repository_open`.  I'll try to investigate.

musm:
I should have noted that this occurred in a `Cygwin` environment.

musm:
Indeed, I can confirm symlinks work fine with windows. Just not in a Cygwin environment.

musm:
I'm closing this since I'm not reproducing the bug.  Yesterday, I kept hitting this bug, and today it works after cleaning my environment and building julia again 😕 . Heisenbug....    I'll let you know if I see this cropping up again.

pks-t:
Sure, thanks for keeping us up to date!

PR #5034, opened 2019-03-29
Tests for symlinked user config

pks-t:
My failed tries to reproduce #5033. I guess it's worthwhile to include them anyway, even though they didn't uncover any errors.


ethomson:
Cool, thanks!

PR #5039, opened 2019-04-04
sha1: don't inline `git_hash_global_init` for win32

ethomson:
Users of the Win32 hash cannot be inlined, as it uses a static struct.
Don't inline it, but continue to declare the function in the header.

/cc @csware

### DIFF
--- a/src/hash/hash_win32.c
+++ b/src/hash/hash_win32.c
@@ -109,6 +109,21 @@ static void git_hash_global_shutdown(void)
 		hash_cryptoapi_prov_shutdown();
 }
 
+int git_hash_global_init(void)
+{
+	int error = 0;
+
+	if (hash_prov.type != INVALID)
+		return 0;
+
+	if ((error = hash_cng_prov_init()) < 0)
+		error = hash_cryptoapi_prov_init();
+
+	git__on_shutdown(git_hash_global_shutdown);
+
+	return error;
+}
+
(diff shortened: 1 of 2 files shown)


pks-t:
Waiting for @csware to confirm it fixes his problem.

pks-t:
By the way, we could also just go all the way and merge #4438, as it also moves all hash implementations into their respective C files.

csware:
Works.

ethomson:
> By the way, we could also just go all the way and merge #4438, as it also moves all hash implementations into their respective C files.

Aha, sorry.  I thought that #4438 was sort of holding until we were going to add additional hashes.  I'm going to go ahead and merge this to unblock @csware but I agree that we should give that some more consideration in the near term.

pks-t:
On Sun, Apr 07, 2019 at 03:10:17AM -0700, Edward Thomson wrote:
> > By the way, we could also just go all the way and merge
> > #4438, as it also moves all hash implementations into their
> > respective C files.
> 
> Aha, sorry.  I thought that #4438 was sort of holding until we
> were going to add additional hashes.  I'm going to go ahead and
> merge this to unblock @csware but I agree that we should give
> that some more consideration in the near term.

Sorry for not making this clearer. I do intend to get this merged
as soon as possible to allow adding additional hashes and get
started the work towards SHA256. My plan was to create smallish
chunks that are rather uncontroversial and not too burdensome to
review.

PR #5040, opened 2019-04-05
ignore: treat paths with trailing "/" as directories

### LINKED ISSUE #5037: .gitignore syntax with cascaded rules fail
I was tracking down a bug in Atom see https://github.com/atom/tree-view/issues/196#issuecomment-478982049

### Reproduction steps

`.gitignore`

```ini
*           # Ignore all files
!/**/       # Allow directories
!*.keep     # Allow .keep files
!.gitignore # Allow .gitignore
```

files
```
.
├── .gitignore
└── src
    ├── A
    └── A.keep
```

custom test case
```c
void test_attr_ignore__mytest(void)
{
	cl_git_rewritefile(
		"*\n",
		"!/**/\n"
		"!*.keep\n"
		"!.gitignore\n"
	);

	assert_is_ignored(false, "src/"); // PASS
	assert_is_ignored(false, "src/A.keep"); // PASS
	assert_is_ignored(false, ".gitignore"); // PASS
	assert_is_ignored(true, "src/A"); // FAIL
}
```

### Expected behavior

`src/A` should be ignored

Proof:
screenshot 23

### Actual behavior

`src/A` is not ignored

9aa049d4c1578cfd73ce66801fe2575632296060 (includes #5020)

OSX 10.14.4 (18E226)

pks-t:
The function `git_ignore_path_is_ignored` is there to test the
ignore status of paths that need not necessarily exist inside of
a repository. This has the implication that for a given path, we
cannot always decide whether it references a directory or a file,
and we need to distinguish those cases because ignore rules may
treat those differently. E.g. given the following gitignore file:

    *
    !/**/

we'd only want to unignore directories, while keeping files
ignored. But still, calling `git_ignore_path_is_ignored("dir/")`
will say that this directory is ignored because it treats "dir/"
as a file path.

As said, the `is_ignored` function cannot always decide whether
the given path is a file or directory, and thus it may produce
wrong results in some cases. While this is unfixable in the
general case, we can do better when we are being passed a path
name with a trailing path separator (e.g. "dir/") and always
treat them as directories.

---

Fixes #5037

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -534,7 +534,9 @@ int git_ignore_path_is_ignored(
 	memset(&path, 0, sizeof(path));
 	memset(&ignores, 0, sizeof(ignores));
 
-	if (git_repository_is_bare(repo))
+	if (!git__suffixcmp(pathname, "/"))
+		dir_flag = GIT_DIR_FLAG_TRUE;
+	else if (git_repository_is_bare(repo))
 		dir_flag = GIT_DIR_FLAG_FALSE;
 
 	if ((error = git_attr_path__init(&path, pathname, workdir, dir_flag)) < 0 ||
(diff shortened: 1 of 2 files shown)


ethomson:
Nice work, @pks-t.

Issue #5038, opened 2019-04-04
Potential performance improvements


romkatv:
This is not really an issue. I would've sent an email if I could find a mailing list address. Feel free to close with no action.

I've been using libgit2 in my own small project -- [gitstatus](https://github.com/romkatv/gitstatus) -- that aims to provide low latency git repository information for the purpose of displaying it in shell prompt. I enjoyed reading through the code of libgit2 a have learned a lot about good C style.

My project computes the equivalent of `git status` plus `git describe --tags --exact-match` on every prompt. It is bad user experience to have a noticeable delay after every shell command, so I've been trying to make my code faster.

I've implemented several performance optimizations in my fork of libgit2 that reduced CPU time of `lg2 status` by slightly over 2x. I've also extended the API to enable repository sharding so that it can be diffed in parallel from multiple threads. I've fixed a few bugs, mostly in synchronization, that got exposed due to heavily multi-threaded access (missing memory fences for the most part). Plus some functional bugs -- there was a nasty one in `filesystem_iterator_examine_path`. All optimizations combined made my version of `lg2 status` 18 times faster on chromium repository. This is 4 times faster than `git status`. I then reimplemented the most CPU-intensive part from scratch, outside of libgit2, which brought the overall speedup to 42x over libgit2 (10x over git).

I've documented some of this on [gitstatus](https://github.com/romkatv/gitstatus) project page albeit rather sparsely.

I've attempted to upstream some of my changes to libgit2 (see [here](https://github.com/libgit2/libgit2/issues/4230#issuecomment-471710359) and [here](https://github.com/libgit2/libgit2/pull/5018)) but things haven't progressed much. I'm now moving on to other projects and won't be able to dedicate time to libgit2.

I realize that my use case is unusual and there are very few users who care about performance of `status` and `describe` commands. I can understand if none of my optimizations or bug fixes are worth the trouble of sifting through them and applying to libgit2. However, on the off-chance you'll find something valuable, check out the docs of [gitstatus](https://github.com/romkatv/gitstatus) (especially [Why fast](https://github.com/romkatv/gitstatus#why-fast) section) and my [fork of libgit2](https://github.com/romkatv/libgit2). I'm sorry for some of the heavy-handed changes in the fork where I expose internal APIs (most of them aren't instrumental to the performance improvements and bug fixes), and for the mess with tabs-vs-spaces.

pks-t:
[...]

PR #5042, opened 2019-04-05
diff: don't check ignores if it wouldn't matter

tiennou:
#4230 uncovered a nice performance improvement, in that we unnecessarily check ignore rules (that ought to only matter when looking at the workdir), even if it would not matter. Additionally, even when looking at a working directory, some diff flags combinations could benefit as well.

This renames the (leftover?) `GIT_ITERATOR_HONOR_IGNORES` to mean the inverse, as it made more sense for backward-compatibility, and changes diff to pass it if it think it'd do a difference.


src/diff_generate.c
```diff
@@ -989,8 +989,13 @@ static int handle_unmatched_new_item(
 	if (git_index_entry_is_conflict(nitem))
 		delta_type = GIT_DELTA_CONFLICTED;
 
-	/* update delta_type if this item is ignored */
-	else if (git_iterator_current_is_ignored(info->new_iter))
+	/* update delta_type if this item is ignored but
+	 * only if it makes a difference -- ignore checks are expensive
+	 */
+	else if ((!contains_oitem ||
+		  DIFF_FLAG_IS_SET(diff, GIT_DIFF_INCLUDE_IGNORED | GIT_DIFF_RECURSE_IGNORED_DIRS |
```
romkatv: Please double-check this set of flags. I tried to be conservative here and have likely enabled the fast path in fewer cases than I could have.

src/diff_generate.c
```diff
@@ -1383,12 +1389,21 @@ int git_diff_index_to_workdir(
 	if (!index && (error = diff_load_index(&index, repo)) < 0)
 		return error;
 
+	/* checking ignore rules is expensive, so we turn it on only if it
+	 * can make a difference */
+	if (opts && (opts->flags & (GIT_DIFF_INCLUDE_IGNORED |
+				    GIT_DIFF_INCLUDE_UNTRACKED |
```
romkatv: Ditto here. Is the set of flags correct?

tiennou:
Rebased, I've limited the skip check to the workdir iterator, because it'd attempt to run on the others. One of the buildbots seems to have locked up, but it should be green now.

src/iterator.c
```diff
@@ -1716,6 +1724,8 @@ static void filesystem_iterator_update_ignored(filesystem_iterator *iter)
 GIT_INLINE(bool) filesystem_iterator_current_is_ignored(
 	filesystem_iterator *iter)
 {
+	if (iterator__skip_ignores(iter)) return false;
```
romkatv: I think libgit2 style is to put `return false` on a separate line? Same thing in a few other places.

src/diff_generate.c
pks-t: Looks good to me. Missing is `GIT_DIFF_SHOW_UNTRACKED_CONTENT`, but the documentation says that it always enables `GIT_DIFF_INCLUDE_UNTRACKED` which is already included here

src/iterator.c
```diff
@@ -1851,10 +1850,12 @@ static int filesystem_iterator_init(filesystem_iterator *iter)
 {
 	int error;
 
-	if (iterator__honor_ignores(&iter->base) &&
-		(error = git_ignore__for_path(iter->base.repo,
+	if (iter->base.type == GIT_ITERATOR_TYPE_WORKDIR
+	    && !iterator__skip_ignores(&iter->base)) {
```
pks-t: The `&&` belongs to the previous line

src/iterator.c
```diff
@@ -1851,10 +1850,12 @@ static int filesystem_iterator_init(filesystem_iterator *iter)
...
+		if ((error = git_ignore__for_path(iter->base.repo,
 			".gitignore", &iter->ignores)) < 0)
```
pks-t: I think you could just retain this line as part of the surrounding condition

src/diff_generate.c
```diff
@@ -989,8 +989,13 @@ static int handle_unmatched_new_item(
 	if (git_index_entry_is_conflict(nitem))
 		delta_type = GIT_DELTA_CONFLICTED;
 
-	/* update delta_type if this item is ignored */
-	else if (git_iterator_current_is_ignored(info->new_iter))
+	/* update delta_type if this item is ignored but
```
pks-t: Mutliline comment blocks start with an empty `/*` line

src/diff_generate.c
```diff
@@ -989,8 +989,13 @@ static int handle_unmatched_new_item(
...
+	 * only if it makes a difference -- ignore checks are expensive
+	 */
```
pks-t: Hum. I'm honestly not a lot smarter after reading this comment. Could you please explain why it cannot make a difference if the below conditions are not met?

src/iterator.c
```diff
@@ -1338,6 +1335,8 @@ static int filesystem_iterator_frame_push(
 	size_t path_len;
 	int error;
 
+	assert(!frame_entry == !iter->frames.size);
```
pks-t: This is non-trivial to me, a comment would be welcome.

src/diff_generate.c
tiennou:
```
	/*
	 * update delta_type if this item is ignored
	 *
	 * If the item is missing from the other iterator, or if we're not
	 * interested in ignored things, skip the check for performance
	 */
```

@romkatv I'm missing the rationale on the `!contain_oitem` check: why does it need to be added? And also, the actual skipping is done at the iterator layer (previous commit), so I'm also unsure what we're gaining by skipping again here — is it's just about sidestepping some other code in the "unmatched new item" case ?

romkatv: > I'm missing the rationale on the !contain_oitem check: why does it need to be added?

It's to avoid checking every directory in a clean repository against ignore rules. For example, when diffing index vs workdir, we might encounter these entries:

* index: foo/bar.h
* workdir: foo/

The code is trying to avoid checking "foo/" against ignore rules if it's safe to do.

It might be helpful to add a `fprintf(stderr, ...)` in `git_ignore__lookup` to see which files get checked and then run `lg2 status` in a clean (or dirty) repository. By changing the PR and seeing how it affects the results it'll be easy to see what each modification does.

> And also, the actual skipping is done at the iterator layer (previous commit)

The previous commit doesn't set `GIT_ITERATOR_SKIP_IGNORES` if `GIT_DIFF_INCLUDE_UNTRACKED` is set.

By the way, I'm positive that each successive commit has an effect because I only changed libgit2 code when it made a measurable difference on performance. However, it's possible that future commits obviated (perhaps partially) previous ones. I never went back to verify that _all_ changes are necessary to achieve the final performance.

PR #5043, opened 2019-04-05
[WIP] diff: allow sharding the path space

tiennou:
This is all parts relative to the sharding mechanism proposed by @romkatv.

The example use case is at https://github.com/romkatv/gitstatus/blob/master/src/git.cc.


tiennou:
My opinion on that one is that it is definitely sound, but sure needs more polishing. To be clear, in my eyes you're doing atomics black magic, and I'm not (yet) in a position to know if they make any sense 😉.

Also, the end of the patchset (starting with `don't copy index to perform a diff; scan it directly`) steers heavily into "I'm scratching my own itch at the expense of the API", so I've included them for consistency but they really are more of a use-case study: how fast we can make the status information available.

src/diff_generate.c
```diff
@@ -1274,6 +1274,12 @@ int git_diff__from_iterators(
 	b_opts.flags = FLAGS_SECOND; \
 	b_opts.start = pfx; \
 	b_opts.end = pfx; \
+	if (opts && opts->range_start && *opts->range_start && \
+			(!pfx || !*pfx || strcmp(opts->range_start, pfx) > 0)) \
```
romkatv: The unconditional use of `strcmp` is incorrect. If the iterator is case-insensitive this should be `strcasecmp`. I cheated here because this makes no difference for gitstatusd.

src/ignore.c
```diff
@@ -267,19 +267,23 @@ static int push_one_ignore(void *payload, const char *path)
...
 	if ((error = git_attr_cache__init(repo)) < 0)
 		return error;
 
-	error = git_attr_cache__get(
-		out, repo, NULL, GIT_ATTR_FILE__IN_MEMORY, NULL, GIT_IGNORE_INTERNAL, NULL);
-
 	/* if internal rules list is empty, insert default rules */
-	if (!error && !(*out)->rules.length)
-		error = parse_ignore_file(repo, *out, GIT_IGNORE_DEFAULT_RULES);
+	error = git_attr_cache__get(
```
romkatv: This is a fix for a synchronization bug: `parse_ignore_file` must be called under a lock.

src/index.c
```diff
@@ -899,30 +905,17 @@ static void index_entry_adjust_namemask(
...
 	git_index_entry **out,
 	git_repository *repo,
 	const char *path,
 	struct stat *st,
-	bool from_workdir)
+	unsigned int path_valid_flags)
 {
 	size_t pathlen = strlen(path), alloclen;
 	struct entry_internal *entry;
-	unsigned int path_valid_flags = GIT_PATH_REJECT_INDEX_DEFAULTS;
```
romkatv: There is a lot of churn here due to my first attempt to implement `GIT_OPT_DISABLE_INDEX_FILEPATH_VALIDATION`. My second attempt has introduced `GIT_PATH_REJECT_NOTHING` and made this churn unnecessary, but I never cleaned it up.

Note that this is unrelated to sharding.

src/repository.h
```diff
@@ -160,7 +160,9 @@ struct git_repository {
 
 GIT_INLINE(git_attr_cache *) git_repository_attr_cache(git_repository *repo)
 {
-	return repo->attrcache;
+	git_attr_cache *cache;
+	__atomic_load(&repo->attrcache, &cache, __ATOMIC_ACQUIRE);
```
romkatv: By the way, there are synchronization bugs when attrcache is lazily initialized. I added a [workaround](https://github.com/romkatv/gitstatus/blob/d3ce9ef43951fcf6eb04cbb005a06f359fb78610/src/repo.cc#L109) in gitstatusd instead of fixing it in libgit2.

tests/iterator/workdir.c
```diff
@@ -1170,7 +1170,7 @@ void test_iterator_workdir__pathlist_for_deeply_nested_item(void)
 
 		cl_git_pass(git_iterator_for_workdir(&i, g_repo, NULL, NULL, &i_opts));
 		expect_iterator_items(i, expected_len, expected, expected_len, expected);
-		cl_assert_equal_i(14, i->stat_calls);
+		cl_assert_in_range(7, i->stat_calls, 14);
```
romkatv: I think changes that have reduced the number of stat calls aren't in this PR. However, since I merely relaxed the assertions, tests will pass either way.

romkatv:
> My opinion on that one is that it is definitely sound, but sure needs more polishing. To be clear, in my eyes you're doing atomics black magic, and I'm not (yet) in a position to know if they make any sense .

Note that my atomic changes cannot break anything. If you replace a plain read/write with atomic/read write, you cannot introduce a bug. Likewise with adding memory fences. In the worst case you'll make code slower but at least on x64 this won't happen.

I discovered most of these bugs with [ThreadSanitizer](https://github.com/google/sanitizers/wiki/ThreadSanitizerCppManual). I'm 99.9% sure that every one of them is a real bug. As I mentioned above, I'm sure I didn't introduce new bugs there. However, I'm not so sure that my fixes were enough to fix the bugs. And it's very unlikely that I've found all synchronization bugs. libgit2 plays rather loose with atomics and synchronization. The code is written in such a way as if reading/writing volatile was both atomic and involved a full memory fence. Neither of these is true, hence bugs.

romkatv:
> Also, the end of the patchset (starting with `don't copy index to perform a diff; scan it directly`) steers heavily into "I'm scratching my own itch at the expense of the API" [...]

That one is pretty big but I my current solution is pretty bad. In fact, it only helps on case-sensitive filesystems.

Consider what happens on macOS with its case-insensitive filesystem when we want to get a list of changes staged for commit. We shard the index into 40 equal subranges (my machine has 20 cores, so 2 ranges per core) and call `git_diff_tree_to_index` on each of them. Each `git_diff_tree_to_index` will create a new index iterator, which will **clone and sort the whole index**. Sorting the whole index is very expensive because it's made of all those long strings with many strings having long shared prefixes.

Note that index iterator never changes the index. It just wants it sorted with a particular comparator. When comparing index to HEAD, index iterator wants the index sorted case-sensitively. When comparing index to workdir, case-insensitive sorting is required.

This problem could be solved by storing two arrays in git_index: one sorted case-sensitively and one case-insensitively. When a new index iterator is being created, it needs to say which sorting it wants. I would've done it this way if I had this idea earlier. But I was focused on improving performance on Linux, so I overlooked the fact that my solution doesn't help macOS.

PR #5044, opened 2019-04-05
[WIP] Performance improvements

tiennou:
This PR is more like a showdown of little things (and might warrant being split again).


src/iterator.c
```diff
@@ -1069,15 +1075,15 @@ static int filesystem_iterator_entry_cmp(const void *_a, const void *_b)
 	const filesystem_iterator_entry *a = (const filesystem_iterator_entry *)_a;
 	const filesystem_iterator_entry *b = (const filesystem_iterator_entry *)_b;
 
-	return git__strcmp(a->path, b->path);
+	return git__strcmp(a->basename, b->basename);
```
romkatv: This is a significant optimization. The current code sorts entries in a directory by their full paths. Full paths share long prefixes, which makes comparator slow, which makes sorting slow. TimSort, that libgit2 uses for sorting, performs insertion sort when there are fewer than 65 elements. Most directories have fewer than 65 entries. Insertion sort performs _a lot_ of comparisons, so it's very important to make comparator cheap. In gitstatusd I use an even cheaper (albeit more complex) comparator but only on Linux. This allows me to sort directory entries 7 times faster than in libgit2.

It's a good idea to add a comment here saying that `a->path` and `b->path` are guaranteed to be in the same directory, and maybe add an assert.

src/iterator.c
```diff
@@ -1181,27 +1189,46 @@ GIT_INLINE(bool) filesystem_iterator_examine_path(
...
 		/* we haven't stat'ed `path` yet, so we don't yet know if it's a
 		 * directory or not.  special case if the current path may be a
 		 * directory that matches the start prefix.
 		 */
-		if (cmp == 0) {
-			if (iter->base.start[path_len] == '/')
+		if (cmp == 0 && iter->base.start[path_len] == '/') {
+			*expected_dir = true;
+		} else if (cmp == 0) {
+			/* Suppose iter->base.start is "b.c" and path is "b", which is a directory.
```
romkatv: Took me quite some time to track down this bug.

The implementation of diffing in gitstatusd sidesteps this whole business with special handling of slashes. Instead of organizing index as a list of file paths, like it's done in libgit2:

```
  foo/bar/baz.c
  foo/bar/baz.h
```

gitstatusd splits index into directories:

```
foo/bar:
  baz.c
  baz.h
```

To diff a directory in the index against a directory in workdir, we need to `open()` the directory, `fstatat()` every file in it that is listed in the index (this will give us all modified, deleted and type-changed deltas), list all entries under the directory and compare them with the list of entries in the index (this will give us untracked files). Note that nowhere we need to distinguish between file entries and subdirectory entries. We also never need to compare file paths with slashes -- we only compare basenames. This results in simpler and much faster code. Diffing no longer requires comparing strings with long prefixes like it's currently done in `git_diff__from_iterators`.

src/iterator.c
```diff
@@ -1393,61 +1424,68 @@ static int filesystem_iterator_frame_push(
...
-			/* treat the file as unreadable */
-			memset(&statbuf, 0, sizeof(statbuf));
-			statbuf.st_mode = GIT_FILEMODE_UNREADABLE;
-
-			error = 0;
-		}
-
-		iter->base.stat_calls++;
+		if (statbuf.st_mode == 0 &&
+		    diriter.d_type == DT_DIR &&
```
romkatv: This optimization avoids calling `stat()` on directories. About 9% fewer `stat()` calls in chromium.

src/iterator.c
```diff
@@ -2087,9 +2127,33 @@ static int index_iterator_advance(
 	const git_index_entry *entry = NULL;
 	bool is_submodule;
 	int error = 0;
+	git_index_entry key;
+	git_buf buf = GIT_BUF_INIT;
 
 	iter->base.flags |= GIT_ITERATOR_FIRST_ACCESS;
 
+	if (iter->end_idx == (size_t)-1) {
```
romkatv: The current code performs a linear scan over the index until it finds an entry with path >= `iter->base.start`. This uses up a lot of CPU when checking the status of a single file or when using sharding. This linear scan negatively affects scaling afforded by sharding because each shard has to scan O(N) entries in the index where N is the total size of the index. I changed it so that the first entry is found with binary search.

If pathspec is set, the code will still do linear search between elements of the pathspec even though this could also be done with binary search. So if you want to diff just two files -- `"a"` and `"z"` -- the code will do a binary search to find `"a"` and then scan index linearly until it finds `"z"`.

I didn't optimize this part because I have a workaround in my code. I just issue two diff calls -- one for `"a"` and another for `"z"`.

src/path.c
```diff
@@ -1429,9 +1434,14 @@ int git_path_diriter_fullpath(
...
 {
+	const char *fname;
+
 	assert(out && diriter);
 
-	return git_path_lstat(diriter->path.ptr, out);
+	fname = diriter->path.ptr + diriter->parent_len;
+	if (*fname == '/') ++fname;
+
+	return fstatat(dirfd(diriter->dir), fname, out, AT_SYMLINK_NOFOLLOW);
```
romkatv: This is a pretty big optimization. I mentioned it in https://github.com/romkatv/gitstatus/blob/master/README.md#why-fast.

src/sortedcache.c
```diff
@@ -207,6 +207,9 @@ int git_sortedcache_lockandload(git_sortedcache *sc, git_buf *buf)
 	int error, fd;
 	struct stat st;
 
+	if ((error = git_futils_filestamp_check_readonly(&sc->stamp, sc->path)) <= 0)
```
romkatv: This is very big optimization when resolving references in parallel from multiple threads. Without it there is no gain from multiple threads at all.

src/tree.c
```diff
@@ -361,6 +361,26 @@ static int parse_mode(uint16_t *mode_out, const char *buffer, size_t buffer_len,
 	int32_t mode;
 	int error;
 
+	if (buffer_len >= 7) {
```
romkatv: I think this is to speed up index reading. 6% performance increase if I recall correctly.

Basically, this function has an expensive implementation but it virtually always gets called with just 3 different arguments, so I special cased them.

src/util.c
```diff
@@ -74,6 +74,15 @@ int git__strntol64(int64_t *result, const char *nptr, size_t nptr_len, const cha
 	int64_t n, nn;
 	int c, ovfl, v, neg, ndig;
 
+	if (base == 10 &&
```
romkatv: This is similar to the above: handling the common case specially.

PR #5050, opened 2019-04-12
git_repository_init: stop traversing at windows root

### LINKED ISSUE #5013: git_repository_init_ext doesn't return when non-existent path contains subdirectory
Originally reported in https://github.com/libgit2/libgit2sharp/issues/1664, it appears that `git_repository_init_ext` never returns if the path passed in via `repo_path` contain a subdirectory.

For example, passing in `c:\foo` when it doesn't exist works fine. Passing in `c:\foo\foo` when `c:\foo` doesn't exist causes it to never return.

For drives that don't exist, `x:\foo` returns the appropriate error. `x:\foo\foo` never returns.

### Reproduction steps
Here's how it's being called from LibGit2Sharp: 
https://github.com/libgit2/libgit2sharp/blob/72a5be4860779e5d2fd6d2900d3dcc119588fbc9/LibGit2Sharp/Core/Proxy.cs#L2471-L2484
### Expected behavior
`git_repository_init_ext` should return with appropriate error code.
### Actual behavior
`git_repository_init_ext` never returns

572e4d8c1f1d42feac1c770f0cddf6fda6c4eca0

Win10

ethomson:
Stop traversing the filesystem at the Windows directory root.  We were calculating the filesystem root for the given directory to create, and walking up the filesystem hierarchy.  We intended to stop when the traversal path length is equal to the root path length (ie, stopping at the root, since no path may be shorter than the root path).

However, on Windows, the root path may be specified in two different ways, as either `Z:` or `Z:\`, where `Z:` is the current drive letter. `git_path_dirname_r` returns the path _without_ a trailing slash, even for the Windows root.  As a result, during traversal, we need to test that the traversal path is _less than or equal to_ the root path length to determine if we've hit the root to ensure that we stop when our traversal path is `Z:` and our calculated root path was `Z:\`.

Without this change, when we're traversing through a nonexistent directory (eg, `Z:\nonexistent`), we will never detect the filesystem root (`Z:` != `Z:\`) and loop infinitely.

Fixes #5013.

### DIFF
--- a/src/fileops.c
+++ b/src/fileops.c
@@ -489,10 +489,13 @@ int git_futils_mkdir(
 
 		assert(len);
 
-		/* we've walked all the given path's parents and it's either relative
-		 * or rooted.  either way, give up and make the entire path.
+		/*
+		 * We've walked all the given path's parents and it's either relative
+		 * (the parent is simply '.') or rooted (the length is less than or
+		 * equal to length of the root path).  The path may be less than the
+		 * root path length on Windows, where `C:` == `C:/`.
 		 */
-		if ((len == 1 && parent_path.ptr[0] == '.') || len == root_len+1) {
+		if ((len == 1 && parent_path.ptr[0] == '.') || len <= root_len) {
 			relative = make_path.ptr;
 			break;
 		}
--- a/tests/repo/init.c
+++ b/tests/repo/init.c
@@ -877,3 +877,15 @@ void test_repo_init__at_filesystem_root(void)
 	git_buf_dispose(&root);
 	git_repository_free(repo);
 }
+
+void test_repo_init__nonexistent_paths(void)
+{
+	git_repository *repo;
+
+#ifdef GIT_WIN32
+	cl_git_fail(git_repository_init(&repo, "Q:/non/existent/path", 0));
+	cl_git_fail(git_repository_init(&repo, "Q:\\non\\existent\\path", 0));
+#else
+	cl_git_fail(git_repository_init(&repo, "/non/existent/path", 0));
+#endif
+}


ethomson:
Note that there's at least one other obvious way to fix this, and that's to make `dirname_r` return a slash-terminated path for the root.  That is to say that presently, when called with `Z:\dir`, the results of `dirname_r` will be `Z:` instead of `Z:\`.

This change certainly _seems_ correct, but also risky.  It's possibly that other parts of the system may be relying on that behavior.

More importantly, though, there's now a single definition of a root path from `dirname_r`.  If we were to keep the trailing slash when given `Z:\dir`, then we would have a potential difference in the root path returns for some inputs - namely when we give `dirname_r` a root path without a trailing slash.  Namely, if we try to return `Z:\` for the `dirname_r` of `Z:\foo`, then we would also return `Z:\` for the `dirname_r` of `Z:\` itself, which seems like no problem.  However, if we call `dirname_r` on `Z:` then we cannot _append_ a trailing slash.  Thus `dirname_r` of `Z:\` and `Z:` will be different, which seems very wrong.

As a result, I'm not touching `dirname_r` and instead changing the way we calculate if we've hit the root path.

src/fileops.c
```diff
@@ -489,10 +489,13 @@ int git_futils_mkdir(
...
-		/* we've walked all the given path's parents and it's either relative
-		 * or rooted.  either way, give up and make the entire path.
+		/*
+		 * We've walked all the given path's parents and it's either relative
+		 * (the parent is simply '.') or rooted (the length is less than or
+		 * equal to length of the root path).  The path may be less than the
+		 * root path length on Windows, where `C:` == `C:/`.
 		 */
-		if ((len == 1 && parent_path.ptr[0] == '.') || len == root_len+1) {
+		if ((len == 1 && parent_path.ptr[0] == '.') || len < root_len + 1) {
```
pks-t: `len <= root_len`? A bit easier to read in my opinion. Other than that this fix looks good to me

PR #5051, opened 2019-04-16
examples: implement SSH authentication

pks-t:
I've been asked how to implement SSH authentication with libgit2. I've thus decided to implement that support for our examples code, so that one may now clone SSH repos via the lg2 executable.

### DIFF
--- a/examples/blame.c
+++ b/examples/blame.c
@@ -14,11 +14,6 @@
 
 #include "common.h"
 
-#ifdef _MSC_VER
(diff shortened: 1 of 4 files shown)


pks-t:
Thanks for your review, @tiennou!

pks-t:
> I'm frustrated that docurium eats that file, though…

Do you know why that is? Is it just not getting processed at all or is there any magic construct inside that docurium fails to parse?

PR #5052, opened 2019-04-16
Add NTLM support for HTTP(s) servers and proxies

ethomson:
When we removed libcurl, we inadvertantly removed support for NTLM connections to proxies; it was supported automatically, which I don't think any of the contributors realized.

This adds NTLM support by including https://github.com/ethomson/ntlmclient, which is a pure C NTLM client implementation that was built to live nicely within the libgit2 codebase.

In addition, it does some refactoring to the HTTP codebase to support NTLM connections to both endpoint servers and proxies.


ethomson:
Rebased on master.

ethomson:
What would help review this?  Should I break out the necessary http client refactoring so that one can review the necessary bits and then follow up with an addition of the NTLM library?

src/transports/http.c
```diff
@@ -483,22 +537,22 @@ static int on_headers_complete(http_parser *parser)
 	if (parser->status_code == 407 && get_verb == s->verb)
 		return on_auth_required(&t->proxy.cred,
 		    parser,
+			&t->proxy,
```
tiennou: Nitpick: strange indent? Not sure who's wrong though 😅.

src/transports/http.c
```diff
@@ -483,22 +537,22 @@ static int on_headers_complete(http_parser *parser)
...
 		    t->proxy_opts.credentials,
 		    t->proxy_opts.payload,
-		    t->proxy.url.username,
 			proxy_auth_types);
 
 	/* Check for an authentication failure. */
 	if (parser->status_code == 401 && get_verb == s->verb)
 		return on_auth_required(&t->server.cred,
 		    parser,
+			&t->server,
```
tiennou: Ditto.

src/transports/http.c
```diff
@@ -701,15 +701,8 @@ static int load_proxy_config(http_subtransport *t)
...
-	if ((error = git_net_url_parse(&t->proxy.url, t->proxy_opts.url)) < 0)
-		return error;
 
-	if (!strcmp(t->proxy.url.scheme, "https")) {
-		git_error_set(GIT_ERROR_NET, "SSL connections to proxy are not supported");
-		return -1;
-	}
-
-	return error;
+	return git_net_url_parse(&t->proxy.url, t->proxy_opts.url);
```
tiennou: 🔥👏

src/transports/http.c
```diff
@@ -445,18 +468,37 @@ static int on_auth_required(
...
+
+	/* If we're in the middle of challenge/response auth, continue */
+	if (parser->status_code == 407 || parser->status_code == 401) {
+		if (server->auth_context && !auth_context_complete(server)) {
+			t->parse_error = PARSE_ERROR_REPLAY;
+			return 0;
+		}
+	}
+
+    /* Enforce a reasonable cap on the number of replays */
```
tiennou: Nitpick: indent.

src/transports/auth_ntlm.c
```diff
@@ -0,0 +1,223 @@
...
+	http_auth_ntlm_context *ctx = (http_auth_ntlm_context *)c;
+
+	ntlm_client_free(ctx->ntlm);
+	git__free(ctx->challenge);
+	git__free(ctx);
+}
+
+static int ntlm_init_context(
+	http_auth_ntlm_context *ctx,
+	const git_net_url *url)
```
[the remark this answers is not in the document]
ethomson: I think the GitHub UI may be showing commits out of order, 8497027 (the rename to `git_net_url`) is the first commit in this series.

src/transports/http.c
```diff
@@ -1017,27 +1017,21 @@ static int http_stream_read(
...
-		if (gen_request(&request, s, 0) < 0)
-			return -1;
-
-		if (git_stream__write_full(t->server.stream, request.ptr,
-					   request.size, 0) < 0) {
-			git_buf_dispose(&request);
-			return -1;
-		}
-
-		git_buf_dispose(&request);
```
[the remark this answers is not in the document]
ethomson: Sorry, I could have been more clear.  We're not explicitly calling `realloc`, but we are doing that implicitly, since we were defining the `request` buffer inside the `replay` block.  So we would create a new buffer for every replay, which is not necessary.  Instead we can clear the existing buffer and just put the new URI inside of it.

deps/ntlmclient/compat.h
```diff
@@ -0,0 +1,33 @@
+/*
```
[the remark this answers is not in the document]
ethomson: I'm pretty comfortable with it.  The functional portions are a reasonably direct port of Microsoft's code and from a technical perspective, I've run this through fuzzers and memory checkers galore.  So although there may be (and I'm sure there still will be) bugs, I think we'll need to ship this to find them.

PR #5053, opened 2019-04-16
config_file: check result of git_array_alloc

tniessen:
`git_array_alloc` can return `NULL` if no memory is available, causing a segmentation fault in memset. This adds `GIT_ERROR_CHECK_ALLOC` similar to how other parts of the code base deal with the return value of `git_array_alloc`.

This is my first contribution to libgit2, please let me know if I made a mistake.


ethomson:
This is a good catch.  Thanks, @tniessen!

PR #5054, opened 2019-04-16
util: use 64 bit timer on Windows

tniessen:
`git__timer` was originally implemented using a 32 bit timer since Windows XP did not support `GetTickCount64`. Windows XP was discontinued five years ago, so it should be safe to use the new API.

As a benefit, we do not need to care about overflows for the next 585 million years.


tniessen:
I am not sure why *one* of the MinGW builds fails. This API was introduced more than 10 years ago and only requires `_WIN32_WINNT >= 0x0600`, so MinGW should support it.

ethomson:
I'm not either, that's pretty surprising.  But IIRC mingw often makes use of its own headers, so it may simply be something that we need to declare explicitly on that platform.

tniessen:
I am not sure what the default `_WIN32_WINNT` value is in current MinGW releases, but you explicitely changed it to `0x0600` in b8bdffb55b4b249ab445d420038a1a97a8dcab5e last year, so it should work. I don't know why this specific platform does not seem to pick it up.

tniessen:
@ethomson Any ideas or updates on this?

ethomson:
No, sorry, I'm not very familiar with mingw.

tniessen:
Took me a while to set up MinGW locally and figure this out. The problem is that `git__timer` is an inline function which is also compiled in `test/` and `fuzzers/`, which don't explicitly set `_WIN32_WINNT`.

This should be good to go now.

fuzzers/CMakeLists.txt
```diff
@@ -1,3 +1,8 @@
+# Ensure that MinGW provides the correct header files.
+IF (WIN32 AND NOT CYGWIN)
+	ADD_DEFINITIONS(-DWIN32 -D_WIN32_WINNT=0x0600)
```
ethomson: Can we just move this into the top-level `CMakeLists.txt`?

tniessen: Good idea, done!

ethomson: One minor difference is that this is now tucked inside an [ELSE](https://github.com/libgit2/libgit2/pull/5054/files#diff-af3b638bc2a3e6c650974192a53c7291R202) block so it will only execute if we're not building with [MSVC](https://github.com/libgit2/libgit2/pull/5054/files#diff-af3b638bc2a3e6c650974192a53c7291R122).  

I _think_ that we want to continue setting this value on all platforms, when we're using MSVC, MinGW or Cygwin...  Does that sound correct?

tniessen: I moved it out of the `ELSE` block for MSVC (not sure if it makes a difference on MSVC). As before, `_WIN32_WINNT` is not set on Cygwin, I believe Cygwin doesn't provide Windows header files anyway.

PR #5055, opened 2019-04-19
Modularize our TLS & hash detection

tiennou:
It's been bothering me for a while, and I had a leftover attempt already, so here it is.

This (somewhat hackily) moves all the backend detection code in one place, and rewire the SHA1 checks to be more logical w.r.t backend availability.

So now `USE_SHA1=ON/CollisionDetection/Backend/Generic`, with a potential `OFF` for when @pks-t hashing works comes to fruition 😉.


ethomson:
The term "backend" is not obvious in its verbiage to me.  Our goal is to use the same mechanism for SHA1 as we use for TLS.  But I don't think the word "backend" conveys that intent.

tiennou:
Backend is a libssh2-ism. Alternatives are "HTTPS", "TLS" (which should be what "HTTPS" is), "Crypto". I hesitate between TLS and Crypto (since I believe ultimately we're going to use both TLS and hashing).

CMakeLists.txt
```diff
@@ -52,11 +52,9 @@ OPTION(TAGS				"Generate tags"						OFF)
 OPTION(PROFILE				"Generate profiling information"			OFF)
 OPTION(ENABLE_TRACE			"Enables tracing support"				OFF)
 OPTION(LIBGIT2_FILENAME			"Name of the produced binary"				OFF)
-
-   SET(SHA1_BACKEND 			"CollisionDetection"			       CACHE STRING
-       "Backend to use for SHA1. One of Generic, OpenSSL, Win32, CommonCrypto, mbedTLS, CollisionDetection.")
 OPTION(USE_SSH				"Link with libssh2 to enable SSH support"		 ON)
 OPTION(USE_HTTPS			"Enable HTTPS support. Can be set to a specific backend" ON)
+OPTION(USE_SHA1				"Enable SHA1. Can be set to CollisionDetection(ON)/Backend/Generic" ON)
```
pks-t: Doesn't this mean that a user now cannot choose e.g. OpenSSL for HTTPS and mbedTLS for SHA1? No idea why one would want that, but I've learned that there's always a usecase for the weirdest things. Might be fine to drop that support, though

pks-t:
> Backend is a libssh2-ism. Alternatives are "HTTPS", "TLS" (which should be what "HTTPS" is), "Crypto". I hesitate between TLS and Crypto (since I believe ultimately we're going to use both TLS and hashing).

"HTTPS", because if we name it "Crypto" or "TLS" then there's no connection to the "-DUSE_HTTPS" parameter. Even better would be "HTTPS_BACKEND", if you ask me.

CMakeLists.txt
tiennou: True, I forgot to point that out. The rationale is that backend "cross-breeding" was only supported as a result of sheer luck, since the detection is only made as part of the backend selection. Since are orthogonal, the only "sane" combinations would be the ones listed.

cmake/Modules/SelectHashes.cmake
```diff
@@ -0,0 +1,65 @@
+# Select a hash backend
```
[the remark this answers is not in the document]
tiennou: I was planning to just stash whatever build support we need for hashes in that file as needed, hence the generic name. It could be easily morphed into a function/loop over hashes or something else.

tiennou:
I don't have any hard opinion on the naming (this is CMake after all), so I'm okay with anything we like (though my preference would be to switch to `USE_TLS` and back-compat `USE_HTTPS` to that).

So `-DUSE_HTTPS=ON/OpenSSL/mbedTLS/SecureTransport/OFF`, `-DUSE_SHA1=SHA1DC/HTTPS/Generic` ?

Huh. I can't express `-DUSE_HTTPS=NO`, `-DUSE_SHA1=$backend` 🤔. Is that something we want ?

ethomson:
> Backend is a libssh2-ism. 

We _also_ use the term "backend" but we use it so frequently (our own ODB / refdb backends, etc) that it needs disambiguation, IMO.

> Even better would be "HTTPS_BACKEND", if you ask me.

This seems reasonable to me, too.

> "TLS" (which should be what "HTTPS" is)

Why do you say that?  TLS is a cryptosystem for transport layers.  HTTPS is a usage of TLS for one _particular_ transport layer.  At the moment, the only TLS-enabled protocol that we support is HTTPS, so it doesn't seem like much is gained by changing this, at the moment.

> Huh. I can't express -DUSE_HTTPS=NO, -DUSE_SHA1=$backend 🤔. Is that something we want ?

Hmm.  I definitely want to be able to compile w/o a specific HTTPS backend.  (`USE_HTTPS=OFF` or whatever).  But my only motivation would ever be to bring my own stack, so I would never want to disable HTTPS but also use OpenSSL for SHA1.

(Using anything but SHA1DC is a big red flag _anyway_ but I'm trying to remember that we should think about hash backends, since we will likely want to use these for SHA256.)

pks-t:
> Hmm. I definitely want to be able to compile w/o a specific HTTPS backend. (USE_HTTPS=OFF or whatever). But my only motivation would ever be to bring my own stack, so I would never want to disable HTTPS but also use OpenSSL for SHA1.

So... maybe we ought to keep the existing manual backend selection for SHA1 and make the detection work independent of the HTTPS backend? If we do that, then I'd still like to see the `HTTPS_BACKEND` flag added, though, as I think it's quite useful.

tiennou:
Let me clarify: we currently have `USE_HTTPS` and `HTTPS_BACKEND`, the former as a public option, which is used to either autoselect or defer to the latter, which is where the detection is done. Our SHA1 detection doesn't detect anything: it piggybacks on whatever the previous part did, unless it's an "internal" backend (and it usually is because we default to SHA1DC).

So, there are *some* gaps, but 
> I definitely want to be able to compile w/o a specific HTTPS backend.

That looks fine, as long as you stick with generic or SHA1DC.
>  If we do that, then I'd still like to see the HTTPS_BACKEND flag added, though, as I think it's quite useful.

That's mostly-handled, by virtue of either `USE_HTTPS=$backend`. Direct usage of `HTTPS_BACKEND` might be hampered by the autoselection overriding it.

pks-t:
> Our SHA1 detection doesn't detect anything: it piggybacks on whatever the previous part did, unless it's an "internal" backend (and it usually is because we default to SHA1DC).

Yeah, unfortunately. For my part, I consider this a weakness of our current setup and think that they should be untied from each other. As you said, if you were to build with USE_HTTPS=NO, then we do not do any autodetection at all. That's fine as long as we use the SHA1DC or Generic backends, but in case we use external ones then we also didn't set up linking against their libraries and fail.

I don't know. I bet it's mostly a theoretical scenario, and it didn't work without this PR anyway. Maybe it even improves that situation, as now we at least don't give users the chance to choose a SHA1 backend that we then fail to link against in case of USE_HTTPS=OFF.

tiennou:
> That's fine as long as we use the SHA1DC or Generic backends, but in case we use external ones then we also didn't set up linking against their libraries and fail.
> I bet it's mostly a theoretical scenario, and it didn't work without this PR anyway.

It would have failed at link-time, depending on the "staticity" of the built product, only if you had used an explicit `SHA1_BACKEND=$backend`. Else we'd default to SHA1DC anyways.

I think what this PR is after is mostly making that limitation explicit: we only support using the current HTTPS backend as our SHA backend if it's enabled and available, or an internal one (plus bonus modularizing).

I think apart from a potential naming fixup we're ok with such a change ?

tiennou:
Rebased, I've gone through all "unlabelled" Backends I've added and switched to `HTTPS`. `USE_SHA1` now takes `ON/CollisionDetection/HTTPS/Generic`.

pks-t:
As this is a backwards-incompatible change to our build options, could you please add a note to our CHANGELOG.md, both mentioning the changed option name as well as what actually changed? Otherwise I'm fine with this, thanks!

tiennou:
Err, docs/changelog.md seems completely unaware of the impending hem, "changes", so as it stands, it looks more like brewing a merge conflict than anything 😉. I've thus wrote a complete commit message, which should help explain what's happening, but if you feel like I should setup a v0.x changelog in here, that's fine as well.

pks-t:
On Thu, Jun 13, 2019 at 07:51:04AM -0700, Etienne Samson wrote:
> Err, docs/changelog.md seems completely unaware of the
> impending hem, "changes", so as it stands, it looks more like
> brewing a merge conflict than anything 😉. I've thus wrote a
> complete commit message, which should help explain what's
> happening, but if you feel like I should setup a v0.x changelog
> in here, that's fine as well.

Yeah. I know that @ethomson has created the upcoming v0.28.0+1
section in some other PR, detailing changes to the HTTP parser
dependency in CMake. I don't quite care whether there's a merge
conflict, in the end it should be trivial to resolve (and I don't
have to do that *heh*).

If you ask me, this change here should definitely be mentioned in
the changelog, as we copy that into the final release notes when
tagging. Thing is that packagers as well as developers should see
breaking changes immediately. So this does not only include
interface changes, but also build instruction changes.

ethomson:
> If you ask me, this change here should definitely be mentioned in the changelog, as we copy that into the final release notes when tagging.

I think this makes sense, too.  Normally I'd be happy to deal with the conflicts, but let me just fix up #5108 and merge it so that you can add a description to the `v0.28 + 1` section that I'm adding.  One sec...

pks-t:
@tiennou: #5108 is merged, so could you please rebase on top of that and mention the change in our new CMake-breaking-changes section? I want to finally ship this :)

tiennou:
:shipit:

pks-t:
Thanks a lot, @tiennou!

PR #5056, opened 2019-04-21
remote: add callback to resolve URLs before connecting

eaigner:
I tried to move the original PR to another branch on my fork, and Github auto-closed my PR.
This is a new one replacing #5028.
Sorry for the dupe.

----------

Since libssh2 doesn't read host configuration from the config file,
this callback can be used to hand over URL resolving to the client
without touching the SSH implementation itself.

Example:

Remote `origin` has URL

    ssh://myhost/~/repo.git

This would obviously fail with the current libgit2 implementation,
given the following SSH config:

    Host myhost
        User gituser
        HostName my.very.long.host.with.subdomain
        Port 1234

The URL callback allows this to be rewritten to e.g.

    gituser@my.very.long.host.with.subdomain:1234/~/repo.git

by the client without much hassle.

### DIFF
--- a/include/git2/remote.h
+++ b/include/git2/remote.h
@@ -478,7 +478,7 @@ typedef int GIT_CALLBACK(git_push_update_reference_cb)(const char *refname, cons
  * Callback to resolve URLs before connecting to remote
  *
  * @param url_resolved The buffer to write the resolved URL to.
- *                     If GIT_PASSTHROUGH is returned, it is not required to write to the buffer.
+ *                     If GIT_PASSTHROUGH is returned, writing to the buffer it is not required.
  * @param url The URL to resolve
  * @param direction GIT_DIRECTION_FETCH or GIT_DIRECTION_PUSH
  * @param payload The callback payload
--- a/include/git2/remote.h
+++ b/include/git2/remote.h
@@ -477,14 +477,14 @@ typedef int GIT_CALLBACK(git_push_update_reference_cb)(const char *refname, cons
 /**
  * Callback to resolve URLs before connecting to remote
  *
- * @param url_resolved The resolved URL. The URL is freed by the caller.
- *                     If GIT_PASSTHROUGH is returned `url_resolved` should not be allocated.
+ * @param url_resolved The buffer to write the resolved URL to.
+ *                     If GIT_PASSTHROUGH is returned, it is not required to write to the buffer.
  * @param url The URL to resolve
  * @param direction GIT_DIRECTION_FETCH or GIT_DIRECTION_PUSH
  * @param payload The callback payload
  * @return 0 on success, GIT_PASSTHROUGH or an error
  */
-typedef int GIT_CALLBACK(git_resolve_url_cb)(char **url_resolved, const char *url, int direction, void *payload);
+typedef int GIT_CALLBACK(git_resolve_url_cb)(git_buf *url_resolved, const char *url, int direction, void *payload);
 
 /**
  * The callback settings structure
--- a/include/git2/remote.h
+++ b/include/git2/remote.h
@@ -475,6 +475,17 @@ typedef int GIT_CALLBACK(git_push_negotiation)(const git_push_update **updates, 
 typedef int GIT_CALLBACK(git_push_update_reference_cb)(const char *refname, const char *status, void *data);
 
 /**
+ * Callback to resolve URLs before connecting to remote
+ *
+ * @param url The URL to resolve
+ * @param url_resolved The resolved URL
+ * @param direction GIT_DIRECTION_FETCH or GIT_DIRECTION_PUSH
+ * @param payload The callback payload
+ * @return 0 on success, otherwise an error
+ */
+typedef int GIT_CALLBACK(git_resolve_url_cb)(const char *url, char **url_resolved, int direction, void *payload);
+
+/**
  * The callback settings structure
  *
  * Set the callbacks to be called by the remote when informing the user
@@ -494,6 +505,12 @@ struct git_remote_callbacks {
 	 * process are done (currently unused).
 	 */
 	int GIT_CALLBACK(completion)(git_remote_completion_t type, void *data);
+
+	/**
+	 * Resolve URL before connecting to remote.
+	 * The returned URL will be used to connect to the remote instead.
+	 */
+	git_resolve_url_cb resolve_url;
 
 	/**
 	 * This will be called if the remote host requires
--- a/include/git2/remote.h
+++ b/include/git2/remote.h
@@ -477,13 +477,14 @@ typedef int GIT_CALLBACK(git_push_update_reference_cb)(const char *refname, cons
 /**
  * Callback to resolve URLs before connecting to remote
  *
+ * @param url_resolved The resolved URL. The URL is freed by the caller.
+ *                     If GIT_PASSTHROUGH is returned `url_resolved` should not be allocated.
  * @param url The URL to resolve
- * @param url_resolved The resolved URL
  * @param direction GIT_DIRECTION_FETCH or GIT_DIRECTION_PUSH
  * @param payload The callback payload
- * @return 0 on success, otherwise an error
+ * @return 0 on success, GIT_PASSTHROUGH or an error
  */
-typedef int GIT_CALLBACK(git_resolve_url_cb)(const char *url, char **url_resolved, int direction, void *payload);
+typedef int GIT_CALLBACK(git_resolve_url_cb)(char **url_resolved, const char *url, int direction, void *payload);
 
 /**
  * The callback settings structure
--- a/src/remote.c
+++ b/src/remote.c
@@ -686,7 +686,7 @@ static int git_remote__urlresolve(char **resolved_url, const char *url, int dire
 	*resolved_url = git__strdup(url);
 	GIT_ERROR_CHECK_ALLOC(*resolved_url);
 
-	return GIT_OK;
+	return 0;
 }
 
 int git_remote__urlfordirection(char **url_out, struct git_remote *remote, int direction, const git_remote_callbacks *callbacks)
--- a/src/remote.c
+++ b/src/remote.c
@@ -670,31 +670,39 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
 	return set_url(repo, remote, CONFIG_PUSHURL_FMT, url);
 }
 
-static int git_remote__urlresolve(const char *url, char **resolved_url, int direction, const git_remote_callbacks *callbacks)
+static int git_remote__urlresolve(char **resolved_url, const char *url, int direction, const git_remote_callbacks *callbacks)
 {
+	int status;
+
+	if (!url)
+		return -1;
+
 	if (callbacks && callbacks->resolve_url) {
-		return callbacks->resolve_url(url, resolved_url, direction, callbacks->payload);
+		status = callbacks->resolve_url(resolved_url, url, direction, callbacks->payload);
+		if (status != GIT_PASSTHROUGH)
+			return status;
 	}
 
 	*resolved_url = url ? git__strdup(url) : NULL;
 
+	GIT_ERROR_CHECK_ALLOC(*resolved_url);
+
 	return GIT_OK;
 }
 
-int git_remote__urlfordirection(struct git_remote *remote, int direction, const git_remote_callbacks *callbacks, char **url_out)
+int git_remote__urlfordirection(char **url_out, struct git_remote *remote, int direction, const git_remote_callbacks *callbacks)
 {
 	assert(remote);
 	assert(direction == GIT_DIRECTION_FETCH || direction == GIT_DIRECTION_PUSH);
 
 	if (direction == GIT_DIRECTION_FETCH) {
-		return git_remote__urlresolve(remote->url, url_out, direction, callbacks);
+		return git_remote__urlresolve(url_out, remote->url, direction, callbacks);
 	}
 
 	if (direction == GIT_DIRECTION_PUSH) {
-		return git_remote__urlresolve(remote->pushurl ? remote->pushurl : remote->url, url_out, direction, callbacks);
+		return git_remote__urlresolve(url_out, remote->pushurl ? remote->pushurl : remote->url, direction, callbacks);
 	}
 
-	/* unreachable */
 	return GIT_EINVALID;
 }
 
@@ -739,15 +747,11 @@ int git_remote__connect(git_remote *remote, git_direction direction, const git_r
 
 	t = remote->transport;
 
-	if ((error = git_remote__urlfordirection(remote, direction, callbacks, &url)) != 0)
-		goto on_error;
-
-	if (url == NULL) {
+	if ((error = git_remote__urlfordirection(&url, remote, direction, callbacks))) {
 		git_error_set(GIT_ERROR_INVALID,
 			"Malformed remote '%s' - missing %s URL",
 			remote->name ? remote->name : "(anonymous)",
 			direction == GIT_DIRECTION_FETCH ? "fetch" : "push");
-		error = -1;
 		goto on_error;
 	}
 
@@ -779,8 +783,7 @@ on_error:
 	if (t)
 		t->free(t);
 
-	if (url)
-		free(url);
+	free(url);
 
 	if (t == remote->transport)
 		remote->transport = NULL;
--- a/src/remote.c
+++ b/src/remote.c
@@ -674,6 +674,8 @@ static int git_remote__urlresolve(git_buf *resolved_url, const char *url, int di
 {
 	int status;
 
+	git_buf_clear(resolved_url);
+
 	if (!url)
 		return -1;
 
--- a/src/remote.c
+++ b/src/remote.c
@@ -670,21 +670,32 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
 	return set_url(repo, remote, CONFIG_PUSHURL_FMT, url);
 }
 
-const char* git_remote__urlfordirection(git_remote *remote, int direction)
+static int git_remote__urlresolve(const char *url, char **resolved_url, int direction, const git_remote_callbacks *callbacks)
+{
+	if (callbacks && callbacks->resolve_url) {
+		return callbacks->resolve_url(url, resolved_url, direction, callbacks->payload);
+	}
+
+	*resolved_url = url ? git__strdup(url) : NULL;
+
+	return GIT_OK;
+}
+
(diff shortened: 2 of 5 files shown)


include/git2/remote.h
```diff
@@ -474,6 +474,17 @@ typedef int GIT_CALLBACK(git_push_negotiation)(const git_push_update **updates,
...
+/**
+ * Callback to resolve URLs before connecting to remote
+ *
+ * @param url The URL to resolve
+ * @param url_resolved The resolved URL
+ * @param direction GIT_DIRECTION_FETCH or GIT_DIRECTION_PUSH
+ * @param payload The callback payload
+ * @return 0 on success, otherwise an error
+ */
+typedef int GIT_CALLBACK(git_resolve_url_cb)(const char *url, char **url_resolved, int direction, void *payload);
```
pks-t: The output-parameter `url_resolved` should be the first argument to this function

include/git2/remote.h
```diff
@@ -474,6 +474,17 @@ typedef int GIT_CALLBACK(git_push_negotiation)(const git_push_update **updates,
  */
 typedef int GIT_CALLBACK(git_push_update_reference_cb)(const char *refname, const char *status, void *data);
 
+/**
+ * Callback to resolve URLs before connecting to remote
+ *
+ * @param url The URL to resolve
+ * @param url_resolved The resolved URL
```
pks-t: Could you please clarify that the resolved URL is expected to belong to the caller afterwards?

src/remote.c
```diff
@@ -670,21 +670,32 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
 	return set_url(repo, remote, CONFIG_PUSHURL_FMT, url);
 }
 
-const char* git_remote__urlfordirection(git_remote *remote, int direction)
+static int git_remote__urlresolve(const char *url, char **resolved_url, int direction, const git_remote_callbacks *callbacks)
 {
-	assert(remote);
+	if (callbacks && callbacks->resolve_url) {
+		return callbacks->resolve_url(url, resolved_url, direction, callbacks->payload);
+	}
```
pks-t: No need for those braces. Edit: I see there's precedent though in the other functions. So you're free to do as you wish ;)

src/remote.c
```diff
@@ -670,21 +670,32 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
 	return set_url(repo, remote, CONFIG_PUSHURL_FMT, url);
 }
 
-const char* git_remote__urlfordirection(git_remote *remote, int direction)
+static int git_remote__urlresolve(const char *url, char **resolved_url, int direction, const git_remote_callbacks *callbacks)
```
pks-t: Same here. The output-parameters belong first

src/remote.c
```diff
@@ -670,21 +670,32 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
...
+	if (callbacks && callbacks->resolve_url) {
+		return callbacks->resolve_url(url, resolved_url, direction, callbacks->payload);
+	}
+
+	*resolved_url = url ? git__strdup(url) : NULL;
+
+	return GIT_OK;
+}
 
+int git_remote__urlfordirection(struct git_remote *remote, int direction, const git_remote_callbacks *callbacks, char **url_out)
```
pks-t: Output parameter should be first.

src/remote.c
```diff
@@ -670,21 +670,32 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
...
+		return git_remote__urlresolve(remote->url, url_out, direction, callbacks);
 	}
 
 	if (direction == GIT_DIRECTION_PUSH) {
-		return remote->pushurl ? remote->pushurl : remote->url;
+		return git_remote__urlresolve(remote->pushurl ? remote->pushurl : remote->url, url_out, direction, callbacks);
 	}
 
-	return NULL;
+	/* unreachable */
```
pks-t: Not exactly unreachable :P In release-code, the above assert would not exist and thus callers may reach this statement

src/remote.c
```diff
@@ -670,21 +670,32 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
 	return set_url(repo, remote, CONFIG_PUSHURL_FMT, url);
 }
 
-const char* git_remote__urlfordirection(git_remote *remote, int direction)
+static int git_remote__urlresolve(const char *url, char **resolved_url, int direction, const git_remote_callbacks *callbacks)
 {
-	assert(remote);
+	if (callbacks && callbacks->resolve_url) {
+		return callbacks->resolve_url(url, resolved_url, direction, callbacks->payload);
```
pks-t: We should handle the case where `resolve_url` returns `GIT_EPASSTHROUGH` or a positive value (whichever makes more sense).

```
if (!callbacks || !callbacks->resolve_url ||
    (error = callbacks->resolve_url(url,...) == GIT_EPASSTHROUGH) {
        *resolved_url = url ? git__strdup(url) : NULL;
        GITERR_ALLOC_CHECK(resolved_url);
}
return error;
```

src/remote.c
```diff
@@ -728,25 +739,28 @@ int git_remote__connect(git_remote *remote, git_direction direction, const git_r
 
 	t = remote->transport;
 
-	url = git_remote__urlfordirection(remote, direction);
+	if ((error = git_remote__urlfordirection(remote, direction, callbacks, &url)) != 0)
```
pks-t: `< 0` should be sufficient. We don't expect the function to return positive.

src/remote.c
```diff
@@ -757,10 +771,16 @@ int git_remote__connect(git_remote *remote, git_direction direction, const git_r
...
+
 	return 0;
 
 on_error:
-	t->free(t);
+	if (t)
+		t->free(t);
+
+	if (url)
+		free(url);
```
pks-t: No need for the `if`

tests/network/remote/remotes.c
```diff
@@ -28,30 +28,78 @@ void test_network_remote_remotes__cleanup(void)
...
+	cl_git_pass(git_remote__urlfordirection(_remote, GIT_DIRECTION_FETCH, &callbacks, &url));
+	cl_assert_equal_s(url, "fetchresolve");
+
+	free(url);
+
+	cl_git_pass(git_remote__urlfordirection(_remote, GIT_DIRECTION_PUSH, &callbacks, &url));
+	cl_assert_equal_s(url, "pushresolve");
+
+	free(url);
+}
```
pks-t: Another test would be nice that verifies that if the callback returns e.g. `GIT_EPASSTHROUGH`, that we'll get the unresolved URL.

include/git2/remote.h
mikeando: Also probably worth explicitly calling out the case where we return GIT_EPASSTHROUGH here as a case where we don't change anything if you're adding that change too.

src/remote.c
eaigner: Mhm. I wonder where my misconception comes from that `free(NULL)` would be illegal... I'm obviously traumatized by some other language.

include/git2/remote.h
eaigner: I don't white understand the case for `GIT_EPASSTHROUGH`. If the callback is not set, the `url` is automatically copied to `resolved_url`. If the caller is set, the caller has to ensure `resolved_url` is set. Do you suggest that the callback can return `GIT_EPASSTHROUGH`, and the caller automatically `strdup`s it?

src/remote.c
eaigner: So just remove the comment and return an error?

pks-t: Yeah, guess that's fine.

tests/network/remote/remotes.c
```diff
@@ -28,30 +28,110 @@ void test_network_remote_remotes__cleanup(void)
...
+
+static int test_network_remote_remotes__urlresolve_passthrough_callback(char **url_resolved, const char *url, int direction, void *payload)
+{
+	#pragma unused(url_resolved)
+	#pragma unused(url)
+	#pragma unused(direction)
+	#pragma unused(payload)
```
pks-t: Please use `GIT_UNUSED` instead of `#pragma`, which is not part of the C standard

src/remote.c
```diff
@@ -670,21 +670,39 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
...
+	if (callbacks && callbacks->resolve_url) {
+		status = callbacks->resolve_url(resolved_url, url, direction, callbacks->payload);
+		if (status != GIT_PASSTHROUGH)
+			return status;
+	}
 
+	*resolved_url = git__strdup(url);
+	GIT_ERROR_CHECK_ALLOC(*resolved_url);
+
+	return GIT_OK;
```
pks-t: Unusual to return `GIT_OK` instead of `0`, but I don't care :P

include/git2/remote.h
pks-t: Sorry, didn't see your comment. But yes, that's exactly it.

eaigner:
somehow the build system does not automatically build at the moment

pks-t:
Thanks for alerting, I didn't notice

pks-t:
It looks like CI is building, but there's no status checks in our PRs anymore. @ethomson, got any clue why that is?

ethomson:
> It looks like CI is building, but there's no status checks in our PRs anymore. @ethomson, got any clue why that is?

Let me find out.

pks-t:
Trying another rebuild, as I saw other PRs do have checks again.

ethomson:
I don't 100% understand the checks architecture between GitHub and Azure Pipelines, but I _think_ that there's only one-shot to get this hooked up.  I think that GitHub notifies Azure Pipelines at PR creation and Azure Pipelines then sets itself up as a check.  I think that one of those two deliveries failed, and so now this PR will forever not have checks from Azure Pipelines.  Unfortunately that means we'll have to click through to the build manually.

I'll raise an issue to see if we can investigate.

eaigner:
Well https://dev.azure.com/libgit2/libgit2/_build/results?buildId=1847 says all checks have passed. Shouldn't that suffice?

ethomson:
> Shouldn't that suffice?

Absolutely.

@tiennou what are you thinking here?

tiennou:
Weeell, on one hand I have some concern about adding a "backend" (this word ^^)-specific function that's more or less a workaround, eg. if I were to add libssh to the party*, I'd had to handle this customization point on top of its `.ssh/config` support (does we do ours before, after ?). Hence my technical answer would be to NAK, invoking layering violations, and the need for upstreaming.

On the other hand, I understand where this is coming from, the libssh2 project has limited bandwidth at the moment, and it really is a big chunk of actual missing functionality. Because of all this I'm reluctant to shot it down…

* which I'm debating doing, just so I can finally stop worrying about baking libssh2-specific things in our 1.0 interface 😉.

ethomson:
Indeed the impetus for this functionality came from the libssh2 limitation, however this does feel generally useful.  It's basically programmatic `insteadOf` without having to deal with the configuration.

tests/network/remote/remotes.c
```diff
@@ -28,30 +28,78 @@ void test_network_remote_remotes__cleanup(void)
...
+static int test_network_remote_remotes__urlresolve_callback(const char *url, char **url_resolved, int direction, void *payload)
+{
+	const char *url_out;
+
+	cl_assert(git__strcmp(url, "git://github.com/libgit2/libgit2") == 0);
+	cl_assert(git__strcmp(payload, "payload") == 0);
+
+	url_out = (direction == GIT_DIRECTION_PUSH) ? "pushresolve" : "fetchresolve";
+
+	*url_resolved = git__strdup(url_out);
```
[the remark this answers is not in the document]
eaigner: I don't know how the custom allocators should be correctly used, I just try to match surrounding code, and since everywhint is using the `git__` style, I used that. So no `git__` allocators in test code?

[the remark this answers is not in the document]
tiennou: Well, that's a pretty specific case here: usually you'd be right, but outside users won't be able to access our `git__strdup`/underlying allocator, and instead will pass their own allocated pointer. This might not be fatal, but the Win32 build can use a different allocator for the library (or the user code might have its own), and it will complain.

Now, this pointer has to be freed *at some point*, but you can't before you've passed it to the library, and we don't give back control afterward to free it. Maybe there's a way to do that by passing a `git_buf` instead of a `void *`, but it needs some more thought because that free *has* to happen. And I might also be completely wrong about all of it, BTW 😉.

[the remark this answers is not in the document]
eaigner: I have no idea what the issue is and what you expect me to do here. In any case, it is a test case. What does it matter?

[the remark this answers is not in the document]
eaigner: Also, don't people pay M$ to fix their toy os? Why does Win32 require some custom allocator implementation instead of the POSIX ones? But probably off topic...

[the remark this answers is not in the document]
eaigner: Also, If users allocating stuff seems to be a problem, it seems to me there is a bigger underlying issue here.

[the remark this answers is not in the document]
eaigner: @tiennou I've replaced the `**char` with a buffer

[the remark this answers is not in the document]
ethomson: Toy OS or not 😒, there's nothing to "fix".  You have to use a `free` function that your allocator specifies.  If you're using `malloc`, then you need to call `free`.  If you _stop_ using `malloc`, and you pick some _other_ allocator, like `jemalloc` or hoard or `tcmalloc`, then you have to use _it's_ free function.

If you take some piece of memory that was given to you by `tcmalloc`, and hand it to libgit2 and libgit2 naively calls `free` on it, then you will crash.

This is possible even within libgit2 itself:  libgit2 has a pluggable allocator, so that if you want it to use `jemalloc`, you can.  So `git__malloc` might not be `malloc`.  So you need to make sure to use `git__free` instead of `free`.

Now, why does this make a difference for Windows?  Because there's a debugging allocator that's frequently used since it adds guard pages and the like, and makes debugging memory issues easier in your application.  But you can't mix and match: if some application is built with the debugging allocator, and allocates some memory, and libgit2 is _not_ built with the debugging allocator, and tries to free that memory, then the application will crash.

But this is not unique to Windows.  This is just a particular instance of the larger problem, on all allocators.  Whoever creates the memory also needs to free the memory.

As a result, libgit2 never calls `free` on anything that the user gave us.  And we never return something and expect a user to `free` it.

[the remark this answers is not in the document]
tiennou: > Whoever creates the memory also needs to free the memory.

Because those allocators might be different, thanks for the clarification @ethomson.

In short, one of the hacky solutions (which I'm not actually recommending but which is somewhat sane) to fix this issue is to use a `static char *`, free it at the beginning of the function, alloc it if it's needed, all because libgit2 doesn't know how the free should be done. You lose reentrancy by doing that, but it works.

What need to happen with that memory must be closely checked against what kind of memory-borrowing we provide, and I'm not sure we have support for passing around a `git_buf` with a custom deallocator… I'll try to review our options tonight, if that's okay.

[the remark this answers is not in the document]
jacquesg: You could use ```alloca``` (```_alloca``` on Windows) if the scope is limited to the function.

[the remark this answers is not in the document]
eaigner: @ethomson sorry for blowing up like this... feel a little guilty. It's just that this PR already takes so long, and then after many changes it seems as it would be discarded in the last minute... lost my cool for a bit.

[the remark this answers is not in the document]
ethomson: No need to apologize.  😀 

I'm sorry that this is taking so long, and I understand your frustration.  I regret that the core contributors have so little time to devote to the project right now.  So it will feel like we are silent, then swoop in only to disappear again for a long stretch.  I know from experience that this is both disheartening and frustrating as a contributor.  So, my apologies for that.  🙏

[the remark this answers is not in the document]
ethomson: I'm happy with this `git_buf` approach.  Callers can set it if they wish using `git_buf_set`, at which point we do a copy.

ethomson:
I'm happy with this `git_buf` approach.  Callers can set it if they wish using `git_buf_set`, at which point we do a copy, so we control both the allocator and the free function.

I'm heads-down on the day job though so I'm very scattered, and I'd like to hold off just to make sure that I haven't missed anything before we 🚢 this.

ethomson:
Sorry, to clarify, hold off _for another core contributor to 👍 this_ in case I'm missing something.

tests/network/remote/remotes.c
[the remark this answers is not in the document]
tiennou: Sorry for the hard feelings @eaigner. I was feeling pretty good about your PR, until I noticed the allocator problem, I wasn't that available yesterday, and it went downhill from there.

Right now I've squashed, polished, and refiled your PR as #5062. I don't really care which goes it, it's about the same code, I've just made some slight code moves for cleanliness + the struct upgrader (Note that I seem to have fixup-ed the wrong commit though, so it not quite how I wanted it to look: the upgrader changes are supposed to be on their own).

> Callers can set it if they wish using `git_buf_set`, at which point we do a copy.

~Isn't that private ? Right now the issue is that our private API is visible in the test suite, which is not true in general.~ My word, it is 😭.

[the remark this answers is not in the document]
ethomson: > Isn't that private ?

Right, it's not and historically for the same reason as here, to allow callers to give us data.  I admit it's a bit wonky but likely the least bad option afforded to us.  (Definitely better than the odb backend allocator, sigh.)

> I don't really care which goes...

I do; I want to merge this PR.  @eaigner has been exceedingly patient and made a lot of iterations on this PR and I want to see this one get merged.

(Putting myself in his shoes, if I'd done a lot of work and the result was a different PR getting merged then I'd be upset.)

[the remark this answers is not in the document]
eaigner: i don't mind. I just want to stop worrying about this PR, and start worrying about the next PR #5057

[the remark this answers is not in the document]
tiennou: > (Putting myself in his shoes, if I'd done a lot of work and the result was a different PR getting merged then I'd be upset.)

I do really want to thank you for that work — even though it looks like I rewrote half of it, it wouldn't have been written it in the first place if it had only been up to me 😁.

On the way forward, I don't mind taking care of the last mile it has to go through to land, but I'd prefer to do so from #5062.  The other solution is that I repurpose #5062 to be just the struct upgrading, then you can make use of that here to fix the bc-break ?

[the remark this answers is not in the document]
eaigner: Go ahead with #5062

PR #5057, opened 2019-04-21
rebase: orig_head and onto accessors

eaigner:
The rebase struct stores fields with information about the current
rebase process, which were not accessible via a public interface.

Accessors for getting the `orig_head` and `onto` branch
names and object ids have been added.

### DIFF
--- a/src/rebase.c
+++ b/src/rebase.c
@@ -1327,6 +1327,22 @@ int git_rebase_finish(
 	return error;
 }
 
+const char *git_rebase_orig_head_name(git_rebase *rebase) {
+	return rebase->orig_head_name;
+}
+
+const git_oid *git_rebase_orig_head_id(git_rebase *rebase) {
+	return &rebase->orig_head_id;
+}
+
+const char *git_rebase_onto_name(git_rebase *rebase) {
+	return rebase->onto_name;
+}
+
+const git_oid *git_rebase_onto_id(git_rebase *rebase) {
+	return &rebase->onto_id;
+}
+
 size_t git_rebase_operation_entrycount(git_rebase *rebase)
 {
 	assert(rebase);
--- a/include/git2/rebase.h
+++ b/include/git2/rebase.h
@@ -200,6 +200,34 @@ GIT_EXTERN(int) git_rebase_open(
 	const git_rebase_options *opts);
 
 /**
+ * Gets the original `HEAD` ref name for merge rebases.
+ *
+ * @return The original `HEAD` ref name
+ */
+GIT_EXTERN(const char *) git_rebase_orig_head_name(git_rebase *rebase);
+
+/**
+ * Gets the original `HEAD` id for merge rebases.
+ *
+ * @return The original `HEAD` id
+ */
+GIT_EXTERN(const git_oid *) git_rebase_orig_head_id(git_rebase *rebase);
+
+/**
+ * Gets the `onto` ref name for merge rebases.
+ *
+ * @return The `onto` ref name
+ */
+GIT_EXTERN(const char *) git_rebase_onto_name(git_rebase *rebase);
+
+/**
+ * Gets the `onto` id for merge rebases.
+ *
+ * @return The `onto` id
+ */
+GIT_EXTERN(const git_oid *) git_rebase_onto_id(git_rebase *rebase);
+
+/**
  * Gets the count of rebase operations that are to be applied.
  *
  * @param rebase The in-progress rebase
--- a/tests/rebase/merge.c
+++ b/tests/rebase/merge.c
@@ -44,6 +44,10 @@ void test_rebase_merge__next(void)
 	git_status_list *status_list;
 	const git_status_entry *status_entry;
 	git_oid pick_id, file1_id;
+	git_oid master_id, beef_id;
+
+	git_oid_fromstr(&master_id, "efad0b11c47cb2f0220cbd6f5b0f93bb99064b00");
+	git_oid_fromstr(&beef_id, "b146bd7608eac53d9bf9e1a6963543588b555c64");
 
 	cl_git_pass(git_reference_lookup(&branch_ref, repo, "refs/heads/beef"));
 	cl_git_pass(git_reference_lookup(&upstream_ref, repo, "refs/heads/master"));
@@ -52,6 +56,12 @@ void test_rebase_merge__next(void)
 	cl_git_pass(git_annotated_commit_from_ref(&upstream_head, repo, upstream_ref));
 
 	cl_git_pass(git_rebase_init(&rebase, repo, branch_head, upstream_head, NULL, NULL));
+
+	cl_assert_equal_s("refs/heads/beef", git_rebase_orig_head_name(rebase));
+	cl_assert_equal_oid(&beef_id, git_rebase_orig_head_id(rebase));
+
+	cl_assert_equal_s("master", git_rebase_onto_name(rebase));
+	cl_assert_equal_oid(&master_id, git_rebase_onto_id(rebase));
 
 	cl_git_pass(git_rebase_next(&rebase_operation, rebase));
 


src/rebase.c
```diff
@@ -1327,6 +1327,34 @@ int git_rebase_finish(
 	return error;
 }
 
+const git_oid *git_rebase_orig_head_id(git_rebase *rebase) {
+	if (rebase->type == GIT_REBASE_TYPE_MERGE) {
+		return &rebase->orig_head_id;
+	}
+	return NULL;
+}
```
ethomson: Nit: we don't do braces around single lines.  eg:

```
if (rebase->type == GIT_REBASE_TYPE_MERGE)
    return &rebase->orig_head_id;

return NULL;
```

But tbqh, I would just:

```
return (rebase->type == GIT_REBASE_TYPE_MERGE) ? &rebase->orig_head_id : NULL;
```

and for the strings you can just return them directly, no?  (They're either null or they're not, no need to check the rebase type).

But that's less important

eaigner: I had it that way before but I was not sure if that would remain the same for other rebase types, since other rebase types can have different file structures/objects inside the `.git/rebase-merge` dir (i'm currently looking into implementing interactive rebase).

But I guess for now thats true, and it probably wont make a difference if I return it directly.

tiennou:
Since you're poking around interactive rebase, [this](https://github.com/libgit2/libgit2/compare/master...tiennou:wip/rebase-interactive) is my own poke around the rebase code. Started last year, quickly rebased, may contain hazardous material, YMMV.

src/rebase.c
```diff
@@ -1327,6 +1327,22 @@ int git_rebase_finish(
...
 
+const git_oid *git_rebase_orig_head_id(git_rebase *rebase) {
+	return (rebase->type == GIT_REBASE_TYPE_MERGE) ? &rebase->orig_head_id : NULL;
+}
+
+const char *git_rebase_orig_head_name(git_rebase *rebase) {
+	return (rebase->type == GIT_REBASE_TYPE_MERGE) ? rebase->orig_head_name : NULL;
+}
+
+const git_oid *git_rebase_onto_id(git_rebase *rebase) {
```
tiennou: That's minor, but I don't like how those two names parse: "onto" makes me expect that something would happen, but those are accessors…

What about `git_rebase_ref_(orig_head|onto)(_id)`. Alternatively, maybe the implementation can be changed to copy the annotated commits instead of the current oid/name split, which would make the final interface a 2-function `git_rebase_annotated_(orig_head|onto)` ?

eaigner: Personally I would let the caller decide if he wants to create a new object for the annotated commit. I'd go with `git_rebase_(orig_head|onto)_ref` and `git_rebase_(orig_head|onto)_id`.

eaigner: I for one am mainly interested in the name/ref alone.

eaigner: Oh I see... you mean storing `git_annotated_commit *branch` as `orig_head` and `git_annotated_commit *onto` as `onto` instead of separate values in `rebase_init_merge`?

eaigner: Ok, now returning `annotated_commit`s. Also has the benefit of returning fully qualified ref names.

eaigner: But i dont know... simplest would be to just omit annotated commits and oids and just return the ref names. If someone wants the ref, they could always dwim it themselves.

src/rebase.c
```diff
@@ -1327,6 +1327,33 @@ int git_rebase_finish(
...
+static int git_rebase_get_annotated_name(git_annotated_commit **out, const char *name, git_rebase *rebase) {
+	int error;
+	git_reference *ref;
+
+	if (rebase->type != GIT_REBASE_TYPE_MERGE) {
+		git_error_set(GIT_ERROR_REBASE, "invalid rebase type");
+		return -1;
+	}
+
+	if ((error = git_reference_dwim(&ref, rebase->repo, name)))
```
tiennou: Hmm, I don't like that an external change could cause a failure there… But we'd be in a lot of trouble already if that was the case.

src/rebase.c
```diff
@@ -1327,6 +1327,33 @@ int git_rebase_finish(
...
+
+	error = git_annotated_commit_from_ref(out, rebase->repo, ref);
+
+	git_reference_free(ref);
+
+	return error;
+}
+
+int git_rebase_annotated_orig_head(git_annotated_commit **out, git_rebase *rebase) {
+	return git_rebase_get_annotated_name(out, rebase->orig_head_name, rebase);
```
tiennou: "Generally", I'd expect that kind of accessor to be of the form `assert(s); return s->member;`, so it has stronger guarantees (ie. you can always get back what you put in). Maybe that could be refactored, but as far as I'm concerned what you're doing here is fine.

eaigner: Well, like I said above, we don't really seem to need the OID at all, so maybe its just best to return the `orig_head` and `onto` names as `const char *` and let the caller resolve/dwim the name himself.

ethomson: There's a bit of a problem with that:  we aren't guaranteed to have the ref names in the future.  Right now we only support two types of rebase (merge-style, and an in-memory format).  The in-memory rebase is something we invented, and since it doesn't write to the filesystem it has no compatibility issues with git, so we can guarantee whatever we want there.

`rebase --merge` emulation has this information, it stores the `onto_name` in a file.  But, other types of rebase _don't_.  You can look at `rebase -i` and see that it has `onto` (an OID) but not the name.  (Because _of course_ it does.)  So when we go to emulate interactive rebase, we'll be missing that information.

So I think we should either expose `onto` and `onto_name` as an oid and a `const char *` (respectively)...  annotated commits are another option, but they're _really_ heavy when all you want is a name and an oid (they actually read the ODB for the commit ID, new up a commit, etc).

Same with `orig_head` for similarity, I think.  We could probably expose the "orig head name" as such to avoid people having to know that it's the "orig head" (an oid) vs the "head name" (a string).

I would propose:

```
git_oid *git_rebase_onto(git_rebase *rebase);
const char *git_rebase_onto_name(git_rebase *rebase);
git_oid *git_rebase_orig_head(git_rebase *rebase);
const char *git_rebase_orig_head_name(git_rebase *rebase);
```

You could also have `onto_id` and `orig_head_id` if you thought that made things more expressive?  I don't care whether the API mirrors the filenames or not.

I would expose _both_ though.  You mention that the caller could resolve them, but there's a pathological case where a user might start a rebase and then edit a ref involved in that rebase in the middle of it.  We should return the literal rebase data.  If a user wanted the `git_oid` for the `orig_head` then if they're expected to get that by resolving the branch name, it could be wrong.  Let's give them the literal data to avoid any surprises or confusion.

eaigner: @ethomson thats how it was in the original commit https://github.com/libgit2/libgit2/commit/44dca762ea3cf47ea92c87780e193d0fff13421a

eaigner:
I reverted it to the original approach, since annotated commits seem overkill to me and I think its better to reflect the actual files on disk without having some voodoo magic inbetween. I left the string accessors with `name` instead of `ref` in their names, since its a name (`char *`) and not an actual `git_reference` object.

So the interface looks like this again:

```
const char *git_rebase_orig_head_name(git_rebase *rebase);
const git_oid *git_rebase_orig_head_id(git_rebase *rebase);
const char *git_rebase_onto_name(git_rebase *rebase);
const git_oid *git_rebase_onto_id(git_rebase *rebase);
```

src/rebase.c
tiennou: > they actually read the ODB for the commit ID, new up a commit, etc

My point was that this cost is payed by the caller, which passes us those annotated commits. 

From an implementer's perspective, I like annotated commits better than a ref/oid pair, especially since I already looked them up at some point (and then there's the context that an annotated commit carries: things like is that a post-fetch merge will be used to generate a merge-commit message). If I had a separate "rebase-operation" class, it wouldn't be able to give me back the object identity it had: I would have to keep track of the initial values passed in my class before the call to `git_rebase`.

I'm fine with the proposal as it stands, and sorry for the nitpicking.

ethomson:
LGTM.  Thanks @eaigner!

eaigner:
🎉😊

PR #5060, opened 2019-04-26
Loosen restriction on wildcard "*" refspecs

### LINKED ISSUE #5058: Parsing of refspec in remote fetch configuration different from git
Based on an issue reported in https://github.com/dotnet/sourcelink/issues/256.

### Reproduction steps

Add a configuration entry like so:

```
[remote "origin"]
        url = https://github.com/somerepo.git
        fetch = +refs/heads/package-*:refs/remotes/origin/package-*
```

and read remotes information using libgit2 APIs.

The issue is that the fetch refspec does not end with `/*`. libgit2 is able to read the configuration if the refspec does end with `/*`.

```
fetch = +refs/heads/package/*:refs/remotes/origin/package/*
``` 

### Expected behavior

Succeeds.

### Actual behavior

```
'+refs/heads/package-*:refs/remotes/origin/package-*' is not a valid refspec.
```

libgit2-a904fc6

Windows.

pks-t:
Upstream has changed the restrictions on wildcard "*" refspecs. Previously, a wildcard in a refspec was only allowed if it was the only character in a component. This restriction has been lifted to allow wildcards to be nested between other characters in that component with cd377f45c9 (refs: loosen restriction on wildcard "*" refspecs, 2015-07-22).

More details in the commits.

Fixes #5058

### DIFF
--- a/src/refs.c
+++ b/src/refs.c
@@ -897,14 +897,13 @@ static int is_valid_ref_char(char ch)
 	case '\\':
 	case '?':
 	case '[':
-	case '*':
 		return 0;
 	default:
 		return 1;
 	}
 }
 
-static int ensure_segment_validity(const char *name)
+static int ensure_segment_validity(const char *name, char may_contain_glob)
 {
 	const char *current = name;
 	char prev = '\0';
@@ -926,6 +925,12 @@ static int ensure_segment_validity(const char *name)
 
 		if (prev == '@' && *current == '{')
 			return -1; /* Refname contains "@{" */
+
+		if (*current == '*') {
+			if (!may_contain_glob)
+				return -1;
+			may_contain_glob = 0;
+		}
 
 		prev = *current;
 	}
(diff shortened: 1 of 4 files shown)


src/refs.c
```diff
@@ -927,6 +926,12 @@ static int ensure_segment_validity(const char *name)
 		if (prev == '@' && *current == '{')
 			return -1; /* Refname contains "@{" */
 
+		if (*current == '*') {
+			if (!may_contain_glob)
+				return -1;
```
ethomson: I realize that you didn't introduce this problem, but reviewing your change shows that we might have one with error reporting.  We return `-1` out of this function w/o setting an error message.  When called by `git_reference__normalize_name`, it will set an error message iff the error code is `GIT_EINVALIDSPEC`.  So it looks to me like we'll return an error without setting a message.  I think we want to return `GIT_EINVALIDSPEC` from this function on any error?

pks-t: Nope, in fact we're good, even though it's non-obvious. First note that we do not set any error code via `ensure_segment_validity` at all, but only the segment length. But we do initialize `error = GIT_EINVALIDSPEC` in the calling function and make sure that it doesn't get set to a different value throughout the function.

ethomson: Cool.  Thanks for the explanation.

PR #5061, opened 2019-04-26
git_revwalk_push_range: do not crash if range is missing

hvoigt:
If someone passes just one ref (i.e. "master") and misses passing the
range we should be nice and return an error code instead of crashing.

I am not sure if this is being overly protective, but IMO a library should not crash just because the content in a string is wrong.

### DIFF
--- a/src/revwalk.c
+++ b/src/revwalk.c
@@ -194,6 +194,12 @@ int git_revwalk_push_range(git_revwalk *walk, const char *range)
 	if ((error = git_revparse(&revspec, walk->repo, range)))
 		return error;
 
+	if (!revspec.to) {
+		git_error_set(GIT_ERROR_INVALID, "invalid revspec. Only range supported.");
+		error = GIT_EINVALIDSPEC;
+		goto out;
+	}
+
 	if (revspec.flags & GIT_REVPARSE_MERGE_BASE) {
 		/* TODO: support "<commit>...<commit>" */
 		git_error_set(GIT_ERROR_INVALID, "symmetric differences not implemented in revwalk");
--- a/src/revwalk.c
+++ b/src/revwalk.c
@@ -203,7 +203,8 @@ int git_revwalk_push_range(git_revwalk *walk, const char *range)
 	if (revspec.flags & GIT_REVPARSE_MERGE_BASE) {
 		/* TODO: support "<commit>...<commit>" */
 		git_error_set(GIT_ERROR_INVALID, "symmetric differences not implemented in revwalk");
-		return GIT_EINVALIDSPEC;
+		error = GIT_EINVALIDSPEC;
+		goto out;
 	}
 
 	if ((error = push_commit(walk, git_object_id(revspec.from), 1, false)))
--- a/tests/revwalk/basic.c
+++ b/tests/revwalk/basic.c
@@ -400,6 +400,15 @@ void test_revwalk_basic__push_range(void)
 	cl_git_pass(test_walk_only(_walk, commit_sorting_segment, 2));
 }
 
+void test_revwalk_basic__push_range_merge_base(void)
+{
+	revwalk_basic_setup_walk(NULL);
+
+	git_revwalk_reset(_walk);
+	git_revwalk_sorting(_walk, 0);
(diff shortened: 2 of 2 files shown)


ethomson:
Hmm.  Should we also be checking for `revspec.from`?  Could somebody pass a range of `...<to>`?

If so, we would likely also crash on the push commit?

src/revwalk.c
```diff
@@ -203,6 +203,11 @@ int git_revwalk_push_range(git_revwalk *walk, const char *range)
 	if ((error = push_commit(walk, git_object_id(revspec.from), 1, false)))
 		goto out;
 
+	if (!revspec.to) {
+		git_error_set(GIT_ERROR_INVALID, "no revspec to found needs to be in format FROM..TO");
+		return GIT_EINVALIDSPEC;
+	}
```
pks-t: I agree that this fix is needed, and @ethomson is probably right in that there ought to be the same check for `revspec.from`. I'd move up this check though to be located directly after the `git_revparse` call: first do all the error checking, and only afterwards should we do anything. As right now, we may end up in an inconsistent state where we've pushed `revspec.from` but then afterwards failed due to `revspec.to` not being set.

hvoigt:
> Hmm. Should we also be checking for `revspec.from`? Could somebody pass a range of `...<to>`?
> 
> If so, we would likely also crash on the push commit?

@ethomson I was not aware that `...<to>` is a valid revision range. For `..<to>`cli git does not complain, but I also do not get any revisions selected.

Not sure how to test for an empty `revspec.from` though, since `...<to>` is returned early and an empty from is always filled.

src/revwalk.c
eaigner: I got a mild stroke reading that error sentence. Could use some punctuation ;) or something like `invalid revspec`

pks-t:
> @ethomson I was not aware that ...<to> is a valid revision range. For ..<to>cli git does not complain, but I also do not get any revisions selected.

Leaving out any side of the symmetric difference notation will have it be replaced by "HEAD". So "...<to>" is equivalent to "HEAD...<to>". Recipe to test:

```
$ git init repo
$ cd repo
$ git commit --allow-empty -mm1
$ git commit --allow-empty -mm2
$ git commit --allow-empty -mm3
$ git checkout HEAD~2
$ git rev-list ...master
20bb0b6 (master) m3
c22e6e3 m2
```

hvoigt:
> Leaving out any side of the symmetric difference notation will have it be replaced by "HEAD". So "...<to>" is equivalent to "HEAD...<to>". Recipe to test:
> 
> ```
> $ git init repo
> $ cd repo
> $ git commit --allow-empty -mm1
> $ git commit --allow-empty -mm2
> $ git commit --allow-empty -mm3
> $ git checkout HEAD~2
> $ git rev-list ...master
> 20bb0b6 (master) m3
> c22e6e3 m2
> ```

Ok if it is defined that way, we do not need to check `revspec.from`. It will always be filled. Either with HEAD or a concrete commit.

I checked the code of `git_revparse()` and it tells me the same. `revspec.from`will always be filled:

1. If one side of `..` or `...` is empty we will have HEAD there.
2. If we pass in a single revision only `revspec.from` will be filled.

Will move the check up and change the error message.

@eaigner maybe next time, instead of complaining about your mild stroke, you could add a concrete suggestion, with a fixed error message, that you prefer. That would be a more constructive critic and remove the mild offence that I received, thanks!

eaigner:
@hvoigt that was meant as a joke, not an insolt (lol). also I suggested a message.

hvoigt:
This should be ready for inclusion.

ethomson:
Thanks @hvoigt.  I merged this manually.

hvoigt:
Thanks @ethomson I just discovered that there is a resolve button, which I should have probably pressed. Would that allow the merge on github? Sorry, I have not done that many reviews through github yet.

ethomson:
No worries, @hvoigt, yours was fine.  I did it so that I could add another commit when I did the push, and I mentioned it explicitly so that there was additional visibility into what I did.

Thanks again!

PR #5062, opened 2019-04-29
Remote URL last-chance resolution

tiennou:
This is #5056, in 3 commits:

- add a helper to upgrade options structures if needed
- update the git_remote_callback users to the autoupgrader
- add a callback to `git_remote_callbacks` to change the URL before the connection is made.

### DIFF
--- a/include/git2/remote.h
+++ b/include/git2/remote.h
@@ -475,6 +475,20 @@ typedef int GIT_CALLBACK(git_push_negotiation)(const git_push_update **updates, 
 typedef int GIT_CALLBACK(git_push_update_reference_cb)(const char *refname, const char *status, void *data);
 
 /**
+ * Callback to resolve URLs before connecting to remote
+ *
+ * If you return GIT_PASSTHROUGH, you don't need to write anything to
+ * url_resolved.
+ *
+ * @param url_resolved The buffer to write the resolved URL to
+ * @param url The URL to resolve
+ * @param direction GIT_DIRECTION_FETCH or GIT_DIRECTION_PUSH
+ * @param payload Payload provided by the caller
+ * @return 0 on success, GIT_PASSTHROUGH or an error
+ */
+typedef int GIT_CALLBACK(git_url_resolve_cb)(git_buf *url_resolved, const char *url, int direction, void *payload);
+
+/**
  * The callback settings structure
  *
  * Set the callbacks to be called by the remote when informing the user
@@ -562,6 +576,12 @@ struct git_remote_callbacks {
 	 * as the last parameter.
 	 */
 	void *payload;
+
+	/**
+	 * Resolve URL before connecting to remote.
+	 * The returned URL will be used to connect to the remote instead.
+	 */
+	git_url_resolve_cb resolve_url;
 };
 
 #define GIT_REMOTE_CALLBACKS_VERSION 1
--- a/src/remote.c
+++ b/src/remote.c
@@ -670,21 +670,44 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
 	return set_url(repo, remote, CONFIG_PUSHURL_FMT, url);
 }
 
-const char* git_remote__urlfordirection(git_remote *remote, int direction)
+static int resolve_url(git_buf *resolved_url, const char *url, int direction, const git_remote_callbacks *callbacks)
 {
-	assert(remote);
+	int status;
 
+	if (callbacks && callbacks->resolve_url) {
+		git_buf_clear(resolved_url);
+		status = callbacks->resolve_url(resolved_url, url, direction, callbacks->payload);
+		if (status != GIT_PASSTHROUGH) {
+			git_error_set_after_callback_function(status, "git_resolve_url_cb");
+			git_buf_sanitize(resolved_url);
+			return status;
+		}
+	}
+
+	return git_buf_sets(resolved_url, url);
+}
+
+int git_remote__urlfordirection(git_buf *url_out, struct git_remote *remote, int direction, const git_remote_callbacks *callbacks)
+{
+	const char *url = NULL;
+
+	assert(remote);
 	assert(direction == GIT_DIRECTION_FETCH || direction == GIT_DIRECTION_PUSH);
 
 	if (direction == GIT_DIRECTION_FETCH) {
-		return remote->url;
+		url = remote->url;
+	} else if (direction == GIT_DIRECTION_PUSH) {
+		url = remote->pushurl ? remote->pushurl : remote->url;
 	}
 
-	if (direction == GIT_DIRECTION_PUSH) {
-		return remote->pushurl ? remote->pushurl : remote->url;
+	if (!url) {
+		git_error_set(GIT_ERROR_INVALID,
+			"malformed remote '%s' - missing %s URL",
+			remote->name ? remote->name : "(anonymous)",
+			direction == GIT_DIRECTION_FETCH ? "fetch" : "push");
+		return GIT_EINVALID;
 	}
-
-	return NULL;
+	return resolve_url(url_out, url, direction, callbacks);
 }
 
 int set_transport_callbacks(git_transport *t, const git_remote_callbacks *cbs)
@@ -707,7 +730,7 @@ static int set_transport_custom_headers(git_transport *t, const git_strarray *cu
 int git_remote__connect(git_remote *remote, git_direction direction, const git_remote_callbacks *callbacks, const git_remote_connection_opts *conn)
 {
 	git_transport *t;
-	const char *url;
+	git_buf url = GIT_BUF_INIT;
 	int flags = GIT_TRANSPORTFLAGS_NONE;
 	int error;
 	void *payload = NULL;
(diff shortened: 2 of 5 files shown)


tiennou:
Rebased, with the unneeded ➰ allocator removed. Most of the code from #5056 has thus been restored.

If that's even possible, it would be nice to have this one in quickly: any new contribution that touches one of our structs is going to have to do something like this (and some are pesky). The sooner this happens, the easier it gets for contributors. Maybe. That's how I felt when I noticed the change in #5056.

I've tried to take into account some of the ideas from the other backward-compatibility PR (git_refdb_backend?), but this one is simpler, thus I don't think it's sufficient I'll try to revisit the other one, though, in case it's not). It's missing documentation which I'd like to add before this goes in, but a first review round seemed like a good idea.

ethomson:
I don't understand why you're incrementing the struct version here?  You're adding a field, that's not an API breaking change?

include/git2/remote.h
```diff
@@ -474,6 +474,24 @@ typedef int GIT_CALLBACK(git_push_negotiation)(const git_push_update **updates,
  */
 typedef int GIT_CALLBACK(git_push_update_reference_cb)(const char *refname, const char *status, void *data);
 
+/**
+ * Callback to resolve URLs before connecting to remote
+ *
+ * You need to be careful not to overflow url_resolved. If it's too small,
+ * returning a positive value equal to the size required will cause the
+ * resolution to be tried again with a buffer of the requested size.
```
ethomson: Is this comment accurate?  I don't see this behavior.

src/remote.c
```diff
@@ -670,21 +670,46 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
...
+		url = remote->pushurl ? remote->pushurl : remote->url;
+	} else {
+		assert(direction == GIT_DIRECTION_FETCH || direction == GIT_DIRECTION_PUSH);
 	}
 
-	if (direction == GIT_DIRECTION_PUSH) {
-		return remote->pushurl ? remote->pushurl : remote->url;
+	if (!url) {
+		git_error_set(GIT_ERROR_INVALID,
+					  "Malformed remote '%s' - missing %s URL",
```
ethomson: Error message should start with lower case.

include/git2/remote.h
tiennou: Oops, a leftover from the old way, before I re-read buffer.h. Please disregard.

tiennou:
It's binary-breaking because if I get a *old* v1 struct without the member and try to access it as a *new*, we'll read some stack data. Hence the bump. I don't think there's a way around, is it?

ethomson:
Yes, this breaks ABI but not API.  We've never tried to be _ABI_-consistent.

It's a noble goal that we should pursue but let's drop those two commits for now and get this merged so we can 🚢.  We do need a strategy for coping with API breaking struct changes, which it seems like this aligns to as well.  So let's split those two commits out into a new PR.  I suspect that there's going to be a few subtle nuances to struct compatibility that we should spend some time to 🤔 about.

tiennou:
Rebased, I've removed the updater and will refile as needed.

src/remote.c
```diff
@@ -670,21 +670,46 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
...
 {
-	assert(remote);
+	int status;
 
-	assert(direction == GIT_DIRECTION_FETCH || direction == GIT_DIRECTION_PUSH);
+	if (callbacks && callbacks->resolve_url) {
+		git_buf_clear(resolved_url);
+		status = callbacks->resolve_url(resolved_url, url, direction, callbacks->payload);
+		if (status != GIT_PASSTHROUGH) {
+			git_buf_sanitize(resolved_url);
```
pks-t: Is this needed? Implementations of this callback only have `git_buf_set` available, which will always NUL-terminate the buffer. If the callback implementation does nothing at all, then `git_buf_clear` will already have the buffer NUL-terminated.

src/remote.c
```diff
@@ -670,21 +670,46 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
...
+			return status;
+		}
+	}
+
+	git_buf_sets(resolved_url, url);
```
pks-t: We should check the return value here. Otherwise we might end up with silently bogus contents in `resolved_url` in case we're OOM.

src/remote.c
```diff
@@ -670,21 +670,46 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
...
+
+	assert(remote);
 
 	if (direction == GIT_DIRECTION_FETCH) {
-		return remote->url;
+		url = remote->url;
+	} else if (direction == GIT_DIRECTION_PUSH) {
+		url = remote->pushurl ? remote->pushurl : remote->url;
+	} else {
+		assert(direction == GIT_DIRECTION_FETCH || direction == GIT_DIRECTION_PUSH);
```
pks-t: Style nit: the assert should probably be moved to the top so that preconditions to that function are easily spottable.

src/remote.c
tiennou: My initial round at this was accessing the ptr directly. I don't have strong opinions, but I think it makes sense as a "boundary" measure: any git_buf we allow userspace to see must be sanitized, just in case (my initial implementation forgot to increment the buffer size, for example).

src/remote.c
tiennou: I was evading Xcode's warnings (for the one time it warns about new things, it's a complete crazy one 😭). The analyzer saw the access to an uninitialized `url` if that assert is hoisted above, because (I think) the `if/else` makes it reconsider its analysis that it's only 2 values.

Frankly, it's the first time I'm seeing this warning, and it's… strange. And gone. Oh Xcode.

src/remote.c
pks-t: Oh, right. I thought that `git_buf` was completely opaque to users of libgit2, but in fact it isn't. So it's fine as-is, I guess

src/remote.c
pks-t: It cannot be uninitialized as you're initializing it to `NULL` above. Huh. Maybe you've added the `NULL` initializer at a later point?

src/remote.c
tiennou: Heh, I had the exact inverse feeling 😮. A public struct with no function at all. Turns out I was wrong, and I've been listening to a synthetized choir of tibetan monks reading the header contents out loud for 2 days now. It's quite melodic 😉.

src/remote.c
tiennou: That ought to be it (this has just been fixed-up and rebased).

src/remote.c
pks-t: It's definitely true that `git_buf` is somewhat strangely exposed. What I find most interesting is that we do not even provide any way to correctly initialize a `git_buf` except for saying `git_buf buf = {0}`.

src/remote.c
pks-t: This still remains, but you've just moved the old block with exact same woring and casing

include/git2/remote.h
```diff
@@ -562,6 +576,12 @@ struct git_remote_callbacks {
 	 * as the last parameter.
 	 */
 	void *payload;
+
+	/**
+	 * Resolve URL before connecting to remote.
+	 * The returned URL will be used to connect to the remote instead.
+	 */
+	git_resolve_url_cb resolve_url;
```
ethomson: I'd like this to be `git_url_resolve_cb`, since that sounds more like something that acts on (resolves) an object (url) instead of something that urls a resolve.

I'm otherwise :+1: on this, so I'm happy to make this change and merge it though to get this unblocked if you're busy.

src/remote.c
tiennou: Yeah, I'd say because exposing the `__const` OOM buf would be weird ? Anyways, I think this is okay as is, maybe something to clarify in the develop guide…

include/git2/remote.h
tiennou: I think it sounds weird in any order, so I renamed. A few weeks later, I have the feeling it could be provided as some kind of "proxying" internal transport, but maybe that's too much (that would make this use case go through the already existing custom transport callback just above this one).

src/remote.c
```diff
@@ -670,21 +670,47 @@ int git_remote_set_pushurl(git_repository *repo, const char *remote, const char*
...
+	if (callbacks && callbacks->resolve_url) {
+		git_buf_clear(resolved_url);
+		status = callbacks->resolve_url(resolved_url, url, direction, callbacks->payload);
+		if (status != GIT_PASSTHROUGH) {
+			git_error_set_after_callback_function(status, "git_resolve_url_cb");
+			git_buf_sanitize(resolved_url);
+			return status;
+		} else {
+			/* reset passthrough */
+			status = 0;
```
pks-t: This is never used

tests/network/remote/remotes.c
```diff
@@ -28,28 +28,97 @@ void test_network_remote_remotes__cleanup(void)
...
+
+	git_buf_dispose(&url);
+}
+
+static int urlresolve_passthrough_callback(git_buf *url_resolved, const char *url, int direction, void *payload)
+{
+	GIT_UNUSED(url_resolved);
+	GIT_UNUSED(url);
+	GIT_UNUSED(direction);
+	GIT_UNUSED(payload);
```
pks-t: Sometimes I wish we'd be allowed to use variadic macros. Would make such unused blocks at least a tad less awful

eaigner:
ping

ethomson:
Yep, this looks good to me.  @tiennou can you remove that unused assignment to `status` and we can 🚢

src/remote.c
tiennou: Fixed!

eaigner: Original author was lost :(

tiennou: Restored, sorry about that (GitX hard-rewrites amends, and I keep not fixing that bug).

tests/network/remote/remotes.c
tiennou: I've recently learnt that there's an `__attribute__((unused))` that does the same thing, also available as `__unused` (for nicer declarations). These can go on parameters, and GCC and Clang looks to be fine with it. Just my 2¢.

src/remote.c
eaigner: Just finishing up on branch graphs, identity management is next in my client (beta.gitfox.app). If you are on macOS, maybe check it out ;)

ethomson:
Thanks @eaigner and @tiennou!

tests/network/remote/remotes.c
pks-t: Yeah, I've used them in my own projects, too. I'm not sure whether I like those more, though.

PR #5063, opened 2019-05-02
cmake: correctly detect if system provides `regcomp`

pks-t:
Same as #5007 but with master branch as target.


pks-t:
Thanks, @ethomson!

PR #5064, opened 2019-05-02
Release v0.28.2

pks-t:
Bugfixes intended for v0.28.2. Please let me know if you'd like to propose additional PRs.

Note that this already includes #5063, even though it's not merged yet. We should first get that one merged before pulling v0.28.2.


pks-t:
#5063 has landed and I've added #5023 to th emix. I consider this ready now -- if noone objects then I propose to release v0.28.2 on May 6th.

danielgindi:
I would like #5065 to make it. It's a simple one, which would spare lots of scripts being executed just to fix symlinks... Makes git with symlinks seamless on win32.

pks-t:
I've included #5065 and its fixup #5077 now

PR #5065, opened 2019-05-05
Support symlinks for directories in win32

danielgindi:
PR #4713 had one thing missing, which is correctly setting the `SYMBOLIC_LINK_FLAG_DIRECTORY` for directories, as Windows weirdly handles symlinks differently than unix-like systems.

This issue is discussed with @ethomson in #4713.

### DIFF
--- a/src/win32/posix_w32.c
+++ b/src/win32/posix_w32.c
@@ -402,8 +402,12 @@ int p_symlink(const char *target, const char *path)
 		git__utf8_to_16(target_w, MAX_PATH, target) < 0)
 		return -1;
 
-	if (!CreateSymbolicLinkW(path_w, target_w,
-	    SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE))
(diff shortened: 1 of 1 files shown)


src/win32/posix_w32.c
```diff
@@ -402,8 +402,12 @@ int p_symlink(const char *target, const char *path)
 		git__utf8_to_16(target_w, MAX_PATH, target) < 0)
 		return -1;
 
-	if (!CreateSymbolicLinkW(path_w, target_w,
-	    SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE))
+	DWORD dwFlags = SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE;
```
pks-t: We need to declare variables at the beginning of its scope due to C89-compatibility.

danielgindi: Hmm
It seems like all of your CI machines compile it fine! (Which is reasonable because all modern compilers support even C++17)
But I'll make the change if that's a requirement.

ethomson:
Thanks for the fix @danielgindi!

ethomson:
And yes @pks-t I'm in favor of backporting this.

PR #5066, opened 2019-05-07
Update README.md

LevonAr:
Just added a link that helped me build this. I'm a super noob so I had to find out more info on my own by scouring google because there wasn't enough on the instructions for someone like me who has very little contextual understanding of what is going on.


ethomson:
Fixed via f6b3f72634821a9fc899783964edf86f50087cad

PR #5073, opened 2019-05-16
Configuration parsing: validate section headers with quotes

### LINKED ISSUE #5072: Configuration parsing: Incorrect parsing of section header
### Reproduction steps

The parser skips any character between whitespace and `"` in between the section name and the subsection name.

Consider configuration file 
```
[x z "y"] a = 1
```

### Expected behavior

Error.

```
D:\repository>git config --list
fatal: bad config line 1 in file .git/config
```

### Actual behavior

libgit2 reads the configuration without reporting any error.

a904fc6d12ffc20aa1a238a78000921494222e89

Windows

ethomson:
When parsing configuration section headers, we previously stopped at the first space, and skipped to the first quotation mark.  This allows us to support valid section headers like `[foo "bar"]` or `[foo        "bar"]`.

However, by skipping from the end of the section name to the first quotation mark, we ignore any invalid characters there.  For example, `[foo skipped "bar"]` which is not legal.

Validate that there is only whitespace between the end of the section name and the first quotation mark.  This also adds additional tests that were useful for my understanding and should help prevent regression.

Fixes #5072

### DIFF
--- a/src/config_parse.c
+++ b/src/config_parse.c
@@ -75,7 +75,7 @@ static int parse_section_header_ext(git_config_parser *reader, const char *line,
 
 	first_quote = strchr(line, '"');
 	if (first_quote == NULL) {
-		set_parse_error(reader, 0, "Missing quotation marks in section header");
+		set_parse_error(reader, 0, "missing quotation marks in section header");
 		goto end_error;
 	}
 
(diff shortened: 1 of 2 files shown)


src/config_parse.c
```diff
@@ -192,7 +191,7 @@ static int parse_section_header(git_config_parser *reader, char **section_out)
 	do {
 		if (git__isspace(c)){
 			name[name_length] = '\0';
-			result = parse_section_header_ext(reader, line, name, section_out);
+			result = parse_section_header_ext(reader, line, pos, name, section_out);
```
pks-t: I really don't like this method name. It's not an extended version of the normal `parse_section_header`, but instead it parses the subsection-part of section headers (e.g. the quoted part). So I think it should just be called `parse_subsection_header` or something like that.

ethomson: Agreed.  Fixed.

PR #5074, opened 2019-05-19
Ignore: only treat one leading slash as a root identifier

### LINKED ISSUE #5070: .gitignore parsing: The parser skips multiple leading `/`
### Reproduction steps

The parser skips multiple leading `/`, but it should only skip one

Consider `.gitignore` file with the following content:
```
//x.txt
```

### Expected behavior

`git_ignore_path_is_ignored("x.txt")` returns false.
Compare with output of `git check-ignore "x.txt"`.

### Actual behavior

`git_ignore_path_is_ignored("x.txt")` returns true. 

a904fc6d12ffc20aa1a238a78000921494222e89

Windows

ethomson:
Per #5070, we should not treat multiple leading slashes as identical to a single leading slash.  One leading slash indicates the directory root, but _only the first_ should indicate that.

Fixes #5070

### DIFF
--- a/src/attr_file.c
+++ b/src/attr_file.c
@@ -604,7 +604,8 @@ int git_attr_fnmatch__parse(
 		if (*scan == '/') {
 			spec->flags = spec->flags | GIT_ATTR_FNMATCH_FULLPATH;
 			slash_count++;
-			if (pattern == scan)
+
+			if (slash_count == 1 && pattern == scan)
 				pattern++;
 		}
 		/* remember if we see an unescaped wildcard in pattern */
--- a/tests/status/ignore.c
+++ b/tests/status/ignore.c
@@ -413,6 +413,30 @@ void test_status_ignore__leading_slash_ignores(void)
 	cl_assert_equal_i(0, counts.wrong_sorted_path);
 }
 
+void test_status_ignore__multiple_leading_slash(void)
+{
+	static const char *test_files[] = {
+		"empty_standard_repo/a.test",
+		"empty_standard_repo/b.test",
+		"empty_standard_repo/c.test",
+		"empty_standard_repo/d.test",
+		NULL
+	};
+
+	make_test_data("empty_standard_repo", test_files);
+	cl_git_mkfile(
+		"empty_standard_repo/.gitignore",
+		"a.test\n"
+		"/b.test\n"
+		"//c.test\n"
+		"///d.test\n");
+
+	assert_is_ignored("a.test");
+	assert_is_ignored("b.test");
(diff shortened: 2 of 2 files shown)


pks-t:
This is weird and contrary to how most other tools handle this. Are we sure this is expected behaviour in git.git?

ethomson:
> This is weird and contrary to how most other tools handle this. Are we sure this is expected behaviour in git.git?

Shockingly, yes.

ethomson:
```
hypnotoad:Temp/TestRepos/028% git status --short
?? a.txt
hypnotoad:Temp/TestRepos/028% echo 'a.txt' > .gitignore
hypnotoad:Temp/TestRepos/028% git status --short
?? .gitignore
hypnotoad:Temp/TestRepos/028% echo '/a.txt' > .gitignore
hypnotoad:Temp/TestRepos/028% git status --short
?? .gitignore
hypnotoad:Temp/TestRepos/028% echo '//a.txt' > .gitignore
hypnotoad:Temp/TestRepos/028% git status --short
?? .gitignore
?? a.txt
```

pks-t:
Well, sometimes there's a mismatch between how git.git works and how it's intended to work :P Do you know if it is documented in any way? If not, I think we shouldn't blindly assume this to be intended and instead ask for clarification on the mailing list. I'd definitely volunteer instead of pushing this additional work on you.

ethomson:
>  I'd definitely volunteer instead of pushing this additional work on you.

I'm happy if you want to chase this down.  :D  But even if this wasn't intended, I think they're probably stuck with this behavior now.

pks-t:
> But even if this wasn't intended, I think they're probably stuck with this behavior now.

I don't know. It's obviously broken, if you ask me, as such a pattern cannot ever match anything given that files cannot contain slashes.

ethomson:
@pks-t did you end up asking on the mailing list?

pks-t:
On Wed, Jun 05, 2019 at 10:05:04AM -0700, Edward Thomson wrote:
> @pks-t did you end up asking on the mailing list?

Not yet, sorry about that. I finally have two days of hacking
time available today and tomorrow, so I'll dig into git.git
now and will try to come up with a patch that "fixes" old
behaviour to get the discussion going.

pks-t:
By the way, I naturally found some different edge cases which we fail to handle correctly while reading gitignore(5) :D I'll dig into those

ethomson:
Ah, yes, thanks, I should have quoted the docs which do shed light on the fact that this insanity is...  well, at least it's documented, if not expected.

PR #5078, opened 2019-05-20
Remove warnings

ethomson:
We've still got some warnings when compiling on Windows amd64.  I want them eliminated.


deps/http-parser/http_parser.c
```diff
@@ -1992,6 +1992,9 @@ http_parse_host(const char * buf, struct http_parser_url *u, int found_at) {
   const char *p;
   size_t buflen = u->field_data[UF_HOST].off + u->field_data[UF_HOST].len;
 
+  if (buflen > UINT16_MAX)
+    return 1;
```
pks-t: It's unfortunate that we're diverging from upstream http-parser. I think we should apply this for now, but try to upstream those changes and then re-sync our copy.

src/iterator.c
```diff
@@ -1303,7 +1303,12 @@ static int filesystem_iterator_entry_init(
 		sizeof(filesystem_iterator_entry), path_len);
 	GIT_ERROR_CHECK_ALLOC_ADD(&entry_size, entry_size, 2);
 
-	entry = git_pool_malloc(&frame->entry_pool, entry_size);
+	if (entry_size > UINT32_MAX) {
+		git_error_set(GIT_ERROR_REPOSITORY, "file path too long");
+		return -1;
+	}
+
+	entry = git_pool_malloc(&frame->entry_pool, (uint32_t)entry_size);
```
pks-t: I'd argue that `git_pool_malloc` should be using `size_t`, but as it's of theoretical value only I'm happy with this, too.

src/merge.c
```diff
@@ -862,12 +862,18 @@ static int merge_conflict_invoke_driver(
 		(error = git_odb_write(&oid, odb, buf.ptr, buf.size, GIT_OBJECT_BLOB)) < 0)
 		goto done;
 
+	if (buf.size > UINT32_MAX) {
+		git_error_set(GIT_ERROR_MERGE, "merged file contents too large for index");
+		error = -1;
+		goto done;
+	}
```
pks-t: So this means we're not able to solve merge conflicts on 4GB files? Unlikely, but it could very well be that somebody has written a merge conflict driver for binary files that is able to resolve such conflicts. On the other hand, this is not a new restriction and at least we're being explicit now.

src/trailer.c
```diff
@@ -42,23 +42,22 @@ static const char *next_line(const char *str)
 }
 
 /*
- * Return the position of the start of the last line. If len is 0, return -1.
+ * Return the position of the start of the last line. If len is 0, return 0.
```
pks-t: We cannot discern text with zero length and text with a single line only, now, can we? Both would return `0`. I think it'd be cleaner to just use an out parameter here

ethomson:
> Oooh, thanks for working on this! I think I've never before seen a warning-free build on MSVC. Didn't know that this was possible

We're warning free on MSVC x86 builds today; this just fixes up amd64.  One of the tricky bits here is that in the win32 ecosystem is that `int` and `long` are always 32 bit (LLP64).  So if one sloppily assumes that they're in an LP64 world where `long` is 64 bits, then that's obviously very wrong.

src/merge.c
ethomson: Right, I'd forgotten how the index handles files > 4GB.  In fact, we should be truncating the file size to a `uint32_t` in the cache.  I updated this to remove this (incorrect) check.

pks-t:
Mhh. It's unfortunate that VSTS deletes the latest build on still existing pull requests :/ Makes it impossible to see what's caused the error without a rebuild

pks-t:
Huh. Not even rebuilding works. Bummer

tiennou:
(Build manually requeued: https://dev.azure.com/libgit2/libgit2/_build/results?buildId=2082)

pks-t:
Cool, thanks @tiennou!

ethomson:
> Mhh. It's unfortunate that VSTS deletes the latest build on still existing pull requests :/ Makes it impossible to see what's caused the error without a rebuild

Yeah, I need to fix this so that we retain builds >30 days.

src/trailer.c
ethomson: Agreed; fixed.

ethomson:
So we're now warning-free on all our core platforms (with the exception of xdiff) and I've enabled `/WX` (the equivalent of `-Werror`) on MSVC and `-Werror` on mingw.

For xdiff, there are warnings on Windows, largely due to LP vs LLP differences.  Fixing them here would be a lot of code churn, so instead, I'd like to work with core git itself to ensure that we have a shared implementation that's warning-free on all platforms.  As a result, I've disabled `/WX` (ie, just allowed the warnings without promoting them to errors) for the xdiff files.  This will ensure that the rest of libgit2 stays warning free while we work with git to update xdiff.

ethomson:
I'd love to get this merged reasonably soon, so that we don't accrue more warnings as time goes on.  😀  But I'll leave this open for a bit longer to make sure that everyone has some time to get 👀 on it.

CMakeLists.txt
```diff
@@ -235,10 +235,19 @@ ELSE ()
...
+	# MinGW uses gcc, which expects POSIX formatting for printf, but
+	# uses the Windows C library, which uses its own format specifiers.
+	# Disable format specifier warnings.
+	IF(MINGW)
+		DISABLE_WARNINGS(format)
+		DISABLE_WARNINGS(format-security)
+	ELSE()
+		ENABLE_WARNINGS(format)
+		ENABLE_WARNINGS(format-security)
+		DISABLE_WARNINGS(documentation-deprecated-sync)
```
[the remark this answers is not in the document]
ethomson: mingw fails with `cc1.exe: warning: unrecognized command line option '-Wno-documentation-deprecated-sync'`.  This is despite cmake believing that it's supported:

```
-- Performing Test IS_WNO_DOCUMENTATION_DEPRECATED_SYNC_SUPPORTED
-- Performing Test IS_WNO_DOCUMENTATION_DEPRECATED_SYNC_SUPPORTED - Success
```

(sigh.)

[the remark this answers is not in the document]
pks-t: Yeah, this one seems to be very peculiar. I've noticed several times that it's popping up in CMake logs when there's compiler errors. Maybe we should just drop that flag altogether? Does any modern platform support it?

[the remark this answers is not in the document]
pks-t: Also:

```
 ~ $ gcc -Wdocumentation-deprecated-sync
gcc: error: unrecognized command line option '-Wdocumentation-deprecated-sync'
gcc: fatal error: no input files
compilation terminated.
 ~ $ gcc -Wno-documentation-deprecated-sync
gcc: fatal error: no input files
compilation terminated.
```

It recognizes that it doesn't know it in the positive case, but it doesn't notice in the negative one. Maybe it's a bug in GCC?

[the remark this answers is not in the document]
ethomson: googling _suggests_ that clang supports it...?  Let's see what happens without it.

pks-t:
By the way, you also need to rebase ;)

CMakeLists.txt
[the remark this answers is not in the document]
ethomson: Yep, clang is definitely unhappy.  We certainly could add the `GIT_DEPRECATED` macro back to add that attribute in, but that still seemed rather heavy-handed today.

It looks like gcc is perfectly happy with _or_ without that option, though.  We should just add that flag for clang...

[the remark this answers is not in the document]
pks-t: *sigh* Yeah, I agree that it'd be too heavy-handed. With so many deprecations going on, users would probably get overwhelmed with a huge amount of warnings. Not much of a fan of adding compiler-specific flags either, but if that's the only way forward then let's do that?

src/merge.c
pks-t: Just to be sure, I can't remember how this looked before. Is this already the updated version? Cannot see any "outdated" marker.

CMakeLists.txt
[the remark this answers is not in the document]
ethomson: Done.

src/merge.c
ethomson: Right!  Sorry, I must not have pushed up a fixed.  (It's now pushed up.)

Issue #5079, opened 2019-05-21
Two copies of `zlib` are linked when using local build of zlib 


dbrnz:
### Reproduction steps
```
git clone https://github.com/libgit2/libgit2.git
cd libgit2
mkdir build
cd build
cmake   -DBUILD_CLAR=OFF -DUSE_HTTPS=OpenSSL -DUSE_SSH=OFF -DWINHTTP=OFF \
        -DOPENSSL_ROOT_DIR=/Users/dave/build/OpenCOR/src/plugins/thirdParty/OpenSSL/ext \
        -DOPENSSL_SSL_LIBRARY=/Users/dave/build/OpenCOR/src/plugins/thirdParty/OpenSSL/ext/lib/libssl.1.0.0.dylib \
        -DOPENSSL_CRYPTO_LIBRARY=/Users/dave/build/OpenCOR/src/plugins/thirdParty/OpenSSL/ext/lib/libcrypto.1.0.0.dylib \
        -DZLIB_INCLUDE_DIR=/Users/dave/build/OpenCOR/src/plugins/thirdParty/zlib/ext/include \
        -DZLIB_LIBRARY=/Users/dave/build/OpenCOR/src/plugins/thirdParty/zlib/ext/lib/libz.1.dylib \
        -G Ninja ..
ninja
$ otool -L libgit2.dylib
```
### Expected behavior
```
libgit2.dylib:
	@rpath/libgit2.28.dylib (compatibility version 28.0.0, current version 0.28.0)
	@rpath/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	@rpath/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	@rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.250.1)
```
### Actual behavior
```
libgit2.dylib:
	@rpath/libgit2.28.dylib (compatibility version 28.0.0, current version 0.28.0)
	@rpath/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	@rpath/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	@rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
[...]

PR #5080, opened 2019-05-21
We've already added `ZLIB_LIBRARIES` to `LIBGIT2_LIBS` so don't also add the `z` library

dbrnz:
This resolves issue #5079.


src/CMakeLists.txt
```diff
@@ -327,12 +327,6 @@ IF(NOT USE_BUNDLED_ZLIB)
 	IF(ZLIB_FOUND)
 		LIST(APPEND LIBGIT2_SYSTEM_INCLUDES ${ZLIB_INCLUDE_DIRS})
 		LIST(APPEND LIBGIT2_LIBS ${ZLIB_LIBRARIES})
-		IF(APPLE OR CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
-			LIST(APPEND LIBGIT2_LIBS "z")
-			LIST(APPEND LIBGIT2_PC_LIBS "-lz")
```
pks-t: I agree that the first line is definitely bogus, we don't want to append twice to `LIBGIT2_LIBS`. But appending to `LIBGIT2_PC_LIBS` is in fact required. Doing so will not cause us to link twice, but the `PC_LIBS` variable will cause us to append this lib to the generated pkgconfig file.

dbrnz: Agree -- updates committed.

src/CMakeLists.txt
```diff
@@ -327,12 +327,11 @@ IF(NOT USE_BUNDLED_ZLIB)
...
-		ELSE()
-			LIST(APPEND LIBGIT2_PC_REQUIRES "zlib")
-		ENDIF()
+        	IF(APPLE OR CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
+            		LIST(APPEND LIBGIT2_PC_LIBS "-lz")
+        	ELSE()
+            		LIST(APPEND LIBGIT2_PC_REQUIRES "zlib")
+        	ENDIF()
```
pks-t: Sorry to be a nag, but there's some leading spaces now in front of the tabs. If you convert that to all-tabs then we're ready to go!

dbrnz: Tabs versus spaces can be a pain! Hopefully all is now good.

PR #5083, opened 2019-05-22
cmake: disable fallthrough warnings for PCRE

ethomson:
Our PCRE dependency has uncommented fallthroughs in switch statements.
Turn off warnings for those in the PCRE code.

(These failed in the nightly.)


ethomson:
This passed in a manual nightly run, merging.

PR #5084, opened 2019-05-23
repository: fix garbage return value

eaigner:
error was never initialized and a garbage value returned on success.


pks-t:
Yup. Unlikely to happen as the user would need to set both `GIT_REPOSITORY_FOREACH_HEAD_SKIP_REPO` and `GIT_REPOSITORY_FOREACH_HEAD_SKIP_WORKTREES`, but a bug nevertheless. Thanks for catching and fixing!

PR #5085, opened 2019-05-24
SECURITY.md: split out security-relevant bits from readme

pks-t:
GitHub has recently introduced a new set of tools that aims to
ease the process around vulnerability reports and security fixes.
Part of those tools is a new security tab for projects that will
display contents from a new SECURITY.md file.

Move relevant parts from README.md to this new file to make use
of this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,9 +103,7 @@ We ask that you not open a GitHub Issue for help, only for bug reports.
 
 **Reporting Security Issues**
 
-In case you think to have found a security issue with libgit2, please do not
-open a public issue.  Instead, you can report the issue to the private mailing
-list [security@libgit2.com](mailto:security@libgit2.com).
+Please have a look at SECURITY.md.
 
 What It Can Do
 ==============
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,14 @@
+# Security Policy
+
+## Supported Versions
+
+This project will always provide security fixes for the latest two released
+versions. E.g. if the latest version is v0.28.x, then we will provide security
+fixes for both v0.28.x and v0.27.y, but no later versions.
+
+## Reporting a Vulnerability
+
+In case you think to have found a security issue with libgit2, please do not
+open a public issue.  Instead, you can report the issue to the private mailing
+list [security@libgit2.com](mailto:security@libgit2.com). We will acknowledge
+receipt of your message in at most three days and try to clarify further steps.


tiennou:
Maybe we could setup a #security channel on Slack and advertise that as well ? Dunno, I remember being unsure about the one I had to report a while ago, and it was nice to have a private chat with @carlosmn about whether it really was security-related.

ethomson:
> Maybe we could setup a #security channel on Slack and advertise that as well 

Maybe - but I’m torn.  The goal of the mailing list is that it’s 1:many instead of wide open like slack would be.  We wouldn’t want people to report potential security problems in a public place.  😕

ethomson:
(Having said that, yes, @tiennou, you should be part of that security reporting.)

pks-t:
> (Having said that, yes, @tiennou, you should be part of that security reporting.)

Agreed.

> Maybe we could setup a #security channel on Slack and advertise that as well

I'd prefer not to have that. There should be exactly one channel for communicating security issues, not multiple ones that might be missed. And as everybody nowadays has an email address, but not everyone may want to sign up to Slack to report such issues, in my opinion it's clear that we should stick with mails only.

That being said, I'd always encourage everybody to just report to security@libgit2.org in case where he's unsure. It's not like that'd kick off any kind of machinery, but we'll just get notified and can still assess whether we want to treat the issue as security-related or not.

tiennou:
My suggestion was to have a maintainer-only place, and Slack seemed like it'd fit the bill, and give us a more realtime channel for those issues/questions (I like realtime better). Though I also like the paper-trail an email leaves. Your call, as always 😉.

pks-t:
> maintainer-only place

Ah, sorry, misunderstood you. I guess it's no secret that I'm no fan of Slack, mostly because I cannot really use it on any device ;) But I guess the new GitHub security stuff already got us covered: as soon as we've got a security flaw, we'd create a security advisory which is essentially a private issue which we can use to discuss. If you think that's sufficient, then I'd vote to use that. If you on the other hand think that we'd need a more real-time channel to help you feel more comfortable when it comes to handling such issues, then I'd not be opposed to that, either.

pks-t:
Merging this for now anyway, as I feel like the discussion here is mostly orthogonal to the PR. If we had a private #security channel, then we'd probably not mention that in SECURITY.md anyway :)

tiennou:
Looking at the new Security tab:

> Privately discuss, fix, and publish information about security vulnerabilities in your repository's code.

Privately ? Would you look at that 😆. Let's keep the orthogonality orthogonal then ^^.

PR #5086, opened 2019-05-24
Restore NetBSD support

jacquesg:
NetBSD 7 added support for ```posix_fallocate```.

See:
https://www.netbsd.org/changes/changes-7.0.html
http://www.cpantesters.org/cpan/report/ca7e64c8-7d3a-11e9-9ccb-b1ae95f66eb9


ethomson:
I'd like us to have some CI for some other platforms (like FreeBSD and NetBSD - I envision a situation where we ssh into some build hosts) but until then, I'm going to have to take your word for this.  😁

PR #5087, opened 2019-05-24
cmake: detect availability of posix_fallocate(3)

pks-t:
Instead of having fragile checks for platform definitions to
guess whether posix_fallocate(3) is available or not, let's just
use CMake's `CHECK_FUNCTION_EXISTS` to detect that. This is both
more accurate and simpler.

### DIFF
--- a/src/posix.c
+++ b/src/posix.c
@@ -157,14 +157,30 @@ int p_rename(const char *from, const char *to)
 
 int p_fallocate(int fd, off_t offset, off_t len)
 {
-#if defined (__APPLE__) || (defined (__NetBSD__) && __NetBSD_Version__ < 700000000)
+	int error = 0, cur;
+	char buf[BUFSIZ];
+	off_t pos = 0;
+
+#if defined(GIT_USE_FALLOCATE_POSIX)
+	if ((error = posix_fallocate(fd, offset, len)) < 0)
+		if (errno == ENOTSUP)
+			goto fallback;
+#elif defined(GIT_USE_FALLOCATE_SOLARIS)
+	struct flock64 fl;
+
+	fl.l_whence = SEEK_SET;
+	fl.l_start = offset;
+	fl.l_len = len;
+
+	if ((error = fcntl(fd, F_ALLOCSP64, &fl)) < 0)
+		goto fallback;
+#elif defined(GIT_USE_FALLOCATE_BSD)
 	fstore_t prealloc;
 	struct stat st;
 	size_t newsize;
-	int error;
 
 	if ((error = p_fstat(fd, &st)) < 0)
-		return error;
+		return -1;
 
 	if (git__add_sizet_overflow(&newsize, offset, len)) {
 		errno = EINVAL;
@@ -182,15 +198,45 @@ int p_fallocate(int fd, off_t offset, off_t len)
 
 	/*
 	 * fcntl will often error when the file already exists; ignore
-	 * this error since ftruncate will also resize the file (although
+	 * this error since our fallback will also resize the file (although
 	 * likely slower).
 	 */
-	fcntl(fd, F_PREALLOCATE, &prealloc);
-
-	return ftruncate(fd, (offset + len));
+	if ((error = fcntl(fd, F_PREALLOCATE, &prealloc)) < 0)
+		goto fallback;
+#elif defined(GIT_USE_FALLOCATE)
+	do {
+		error = fallocate(fd, 0, offset, len);
+	} while (error < 0 && errno == EINTR);
+	if (error < 0 && errno == ENOTSUP)
+		goto fallback;
 #else
-	return posix_fallocate(fd, offset, len);
+	goto fallback;
 #endif
+
+fallback:
+	/* save offset, seek, write zeroes, restore */
+	if ((error = p_lseek(fd, 0, SEEK_CUR)) < 0)
+		return -1;
+
+	cur = error;
+
+	memset(&buf, '\0', sizeof(buf));
+
+	if ((error = p_lseek(fd, offset, SEEK_SET)) < 0)
+		goto cleanup;
+
+	while (pos < len) {
+		size_t wlen = MIN(len - pos, (off_t)sizeof(buf));
+		if ((error = p_write(fd, buf, wlen)) < 0)
+			goto cleanup;
+		pos += wlen;
+	}
+
+cleanup:
+	if ((error = p_lseek(fd, cur, SEEK_SET)) < 0)
+		return error;
+
+	return error;
 }
 
 #endif /* GIT_WIN32 */
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -53,6 +53,27 @@ IF (HAVE_FUTIMENS)
 	SET(GIT_USE_FUTIMENS 1)
 ENDIF ()
 
+CHECK_SYMBOL_EXISTS(posix_fallocate fcntl.h HAVE_POSIX_FALLOCATE)
+CHECK_SYMBOL_EXISTS(fallocate fcntl.h HAVE_FALLOCATE)
+CHECK_C_SOURCE_COMPILES(
+	"#include <fcntl.h>
+	int main(void) { return F_PREALLOCATE; }"
+	HAVE_F_PREALLOCATE)
+CHECK_C_SOURCE_COMPILES(
+	"#include <fcntl.h>
+	int main(void) { return F_ALLOCSP64; }"
+	HAVE_F_ALLOCSP64)
+IF(HAVE_F_PREALLOCATE)
+	# APPLE OR FREEBSD OR NETBSD
+	SET(GIT_USE_FALLOCATE_BSD 1)
+ELSEIF(HAVE_F_ALLOCSP64)
+	SET(GIT_USE_FALLOCATE_SOLARIS 1)
+ELSEIF(HAVE_FALLOCATE)
+	SET(GIT_USE_FALLOCATE 1)
+ELSEIF (HAVE_POSIX_FALLOCATE)
+	SET(GIT_USE_FALLOCATE_POSIX 1)
+ENDIF ()
+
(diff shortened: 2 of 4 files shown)


pks-t:
On Fri, May 24, 2019 at 07:01:26AM -0700, Etienne Samson wrote:
> I do hope there's no such problem as the `qsort/_r/_s`
> conundrum, but I don't recall inconsistent signatures)

Well, I'd definitely hope that if a function already has a
`posix_` prefix, that implementers will in fact implement what
POSIX specifies. But you never know ;)

ethomson:
> Well, I'd definitely hope that if a function already has a
`posix_` prefix, that implementers will in fact implement what
POSIX specifies. But you never know ;)

Indeed.  The GCC builds are complaining about something, though I don't know what that is.

pks-t:
On Sat, May 25, 2019 at 01:29:16AM -0700, Edward Thomson wrote:
> Indeed.  The GCC builds are complaining about something, though I don't know what that is.

Huh. For whatever reason, those systems do not detect
`posix_fallocate`, even though they should definitely have it
available:

    2019-05-24T11:54:32.6862203Z -- Looking for posix_fallocate
    2019-05-24T11:54:32.7131435Z -- Looking for posix_fallocate - not found

Will need to investigate why that is.

tiennou:
I've pushed a fixup [here](https://github.com/libgit2/libgit2/compare/master...tiennou:pks/cmake-posix-fallocate). I've ended up doing a check for the posix_fallocate symbol, and if that fails, it'll check if we're on some kind of BSD.

This test is "looser" that the original NetBSD preprocessor checks provided by @jacquesg, but if I understood correctly, NetBSD gained `posix_fallocate` in v7, so doing the generic check first should handle all BSD (modulo Apple-specific modifications).

jacquesg:
FreeBSD has had ```posix_fallocate``` for many years, wouldn't worry about it.

jacquesg:
I would also check for it on Solaris, as I believe it is only available from 11 onwards. Version 10 is still in wide-spread use.

tiennou:
Thanks for the insight. I haven't been able to find a reference about what was supported on Solaris then. If it's the fcntl stuff as the BSDs need it's okay, but if it's not then (I'd really like to not have to implement a "stupid" write zeroes fallback)…

jacquesg:
Here is some guidance on how to implement it for Solaris: https://fossies.org/linux/dpkg/lib/dpkg/fdio.c

jacquesg:
Ok, so it doesn't seem like any version of NetBSD or OpenBSD has ```F_PREALLOCATE```. So in summary:

* FreeBSD: Has ```posix_fallocate``` since version 7 (added about 11 years ago)
* OpenBSD: Doesn't have ```posix_fallocate``` or ```F_PREALLOCATE```
* NetBSD: Has ```posix_fallocate``` from version 7, doesn't have ```F_PREALLOCATE```
* macOS: Has ```F_PREALLOCATE```
* Solaris: Has ```F_ALLOCSP64```.

krytarowski:
This code is still not fully correct.

On NetBSD posix_fallocate(2) will report EOPNOTSUPP for frequently used filesystems (we need kernel support for each filesystem) and must have a fallback for the ftruncate(2) function.

krytarowski:
Example: https://github.com/llvm-mirror/llvm/blob/90a7bd7e82703a62ec7b54a817ca6f4fde080e8d/lib/Support/Unix/Path.inc#L505

tiennou:
Okay, I've updated my fixup commit to CMake checks all of those, choose the one available, and fallback to `ftruncate` on `ENOTSUP`/nothing being available.

ethomson:
Gah, the build expired.  I'll have to figure out how to remove the build expirations.  In the meantime, I queued a new build against this branch manually, and it's [still failing](https://dev.azure.com/libgit2/libgit2/_build/results?buildId=2047), unfortunately.

tiennou:
@pks-t Do you mind if I take over this one ? I'm responsible for the half-working platform support, have a tentative fix handy which I forgot to point out here (there should be a [lone commit](https://github.com/libgit2/libgit2/compare/master...tiennou:pks/cmake-posix-fallocate) I've added to my fork). As you wish, but since it impacts platforms not quite under CI, it feel best to fix it "fast" (maybe) ?

pks-t:
On Wed, Jun 12, 2019 at 04:33:28AM -0700, Etienne Samson wrote:
> @pks-t Do you mind if I take over this one ? I'm responsible
> for the half-working platform support, have a tentative fix
> handy which I forgot to point out here (there should be a [lone
> commit](https://github.com/libgit2/libgit2/compare/master...tiennou:pks/cmake-posix-fallocate)
> I've added to my fork). As you wish, but since it impacts
> platforms not quite under CI, it feel best to fix it "fast"
> (maybe) ?

Nope, I'm fine with that. Sorry for not investing more time here,
and thanks for taking over!

pks-t:
Updated with @tiennou's changes

tiennou:
Okay, so the Linux test failed because I have inconsistent semantics, I believe because of the `ftruncate` fallback I've added without thinking.

My expectation of whatever POSIX describes as `fallocate`, is that a call to `p_fallocate(fd, 0, 1024);` cannot be assumed to mean that I want the file backed by fd to be truncated at 1024, only that I expect that part of the file to exist (else, what would happen with multiple calls, in case I wanted to zero parts of a file for some subsequent `mmap`s). So I've reimplemented the fallback as a dumb write loop (@pks-t if you can pull that, that'd be 🙏). Does anyone sees something wrong with that understanding ?

ethomson:
> Okay, so the Linux test failed because I have inconsistent semantics, I believe because of the ftruncate fallback I've added without thinking.

Hmm, what version of Linux are we building on that's missing `posix_fallocate` and is thus falling into the `ftruncate` fallback?

ethomson:
In any case, yes, I agree that we need to avoid _shrinking_ a file with `ftruncate`, but it should be much faster to just do arithmetic to determine if we would be shrinking or not, and only calling `ftruncate` opportunistically (or doing nothing).  Is there something wrong with the code that was there from bd132046b04875f928e52d16363fb73f8e85dded#diff-af200ec3b10b961c568b86610775effe ?

tiennou:
> Hmm, what version of Linux are we building on that's missing posix_fallocate and is thus falling into the ftruncate fallback?

Dunno, but I'd wager it's also filesystem-specific, and we're in crazy vm shenanigans territory, so that could have been the fallback.

> In any case, yes, I agree that we need to avoid shrinking a file with ftruncate, but it should be much faster to just do arithmetic to determine if we would be shrinking or not, and only calling ftruncate opportunistically (or doing nothing). Is there something wrong with the code that was there from bd13204#diff-af200ec3b10b961c568b86610775effe ?

Not per se, I could have just added a check to make sure we'd not cause the file to shrink. Would that work better ? It's just that I'm expecting I'll have to juggle offsets, lengths and seek positions to convert "fallocate hole" into "ftruncate complete file but don't make it smaller", and the dumb loop seemed safer…

tiennou:
@pks-t branch updated again (Xcode's renamer doesn't look inside #ifdefs). Do you have an opinion on the dump-write-loop vs. ftruncate-unless-it-would-shrink approach ?

src/posix.c
```diff
@@ -182,15 +198,45 @@ int p_fallocate(int fd, off_t offset, off_t len)
...
+		error = fallocate(fd, 0, offset, len);
+	} while (error < 0 && errno == EINTR);
+	if (error < 0 && errno == ENOTSUP)
+		goto fallback;
 #else
-	return posix_fallocate(fd, offset, len);
+	goto fallback;
 #endif
+
+fallback:
```
pks-t: I don't quite get it. Aren't we now _always_ executing the fallback, regardless of whether the platform-specific implementation succeeded or not?

src/posix.c
```diff
@@ -182,15 +198,45 @@ int p_fallocate(int fd, off_t offset, off_t len)
...
 	 */
-	fcntl(fd, F_PREALLOCATE, &prealloc);
-
-	return ftruncate(fd, (offset + len));
+	if ((error = fcntl(fd, F_PREALLOCATE, &prealloc)) < 0)
+		goto fallback;
+#elif defined(GIT_USE_FALLOCATE)
+	do {
+		error = fallocate(fd, 0, offset, len);
+	} while (error < 0 && errno == EINTR);
```
pks-t: Can `fallocate` set `EAGAIN`?

src/posix.c
```diff
@@ -182,15 +198,45 @@ int p_fallocate(int fd, off_t offset, off_t len)
...
+	if (error < 0 && errno == ENOTSUP)
+		goto fallback;
 #else
-	return posix_fallocate(fd, offset, len);
+	goto fallback;
```
pks-t: I'd just remove the `else` block, but that's a matter of taste

src/posix.c
tiennou: Heh, true, I seem to have forgot a few `returns` when whatever platform-specific version succeeded. IOW, only failures of the (potential) platform fallocate would cause the fallback to run.

src/posix.c
tiennou: It's protecting against an "unused label" warning on macOS case, actually.

src/posix.c
tiennou: That's copy-pasted from https://github.com/libgit2/libgit2/pull/5087#issuecomment-495979590, actually. And isn't  `EAGAIN` usually async network stuff ? I'm scared now…

pks-t:
> So if there's any chance to drop that madness, I'm all for it now

Yeah, seeing how this escalated quickly I'm all for dropping `p_fallocate` completely and using some other sorcery for the bigfile test.

ethomson:
> Yeah, seeing how this escalated quickly I'm all for dropping p_fallocate completely and using some other sorcery for the bigfile test.

+1

pks-t:
Closing, this has been fixed via #5114 by removing `p_fallocate` altogether.

PR #5088, opened 2019-05-24
cache: fix cache eviction using deallocated key

### LINKED ISSUE #5081: cache: crash on evict
I have some sporadic crashes when using a `revwalk` object. Here is the log:

```
malloc: *** error for object ...: pointer being freed was not allocated

#6. Crashed: NSOperationQueue 0x60000039d3c0 (QOS: USER_INITIATED)
0  libsystem_kernel.dylib         0x7fff5d32a2c6 __pthread_kill + 10
1  libsystem_pthread.dylib        0x7fff5d3dfbf1 pthread_kill + 284
2  libsystem_c.dylib              0x7fff5d2946a6 abort + 127
3  libsystem_malloc.dylib         0x7fff5d3a2aef malloc_vreport + 545
4  libsystem_malloc.dylib         0x7fff5d3a28b0 malloc_report + 151
5  Git                            0x10cce7e0e git_odb_object__free + 30
6  Git                            0x10ccae68e git_commit_list_parse + 814
7  Git                            0x10cd1a614 add_parents_to_list + 100
8  Git                            0x10cd1a10d git_revwalk_next + 333
9  Git                            0x10cd2e6d4 git_smart__negotiate_fetch + 500
10 Git                            0x10ccc51b1 git_fetch_negotiate + 609
11 Git                            0x10cd0fb9a git_remote_download + 1194
12 Git                            0x10cd0fd1a git_remote_fetch + 202
13 Git                            0x10cb8bc4b globalinit_33_46240B7271BAB13E345CE2E50BE21379_func5 + 51947
14 Gitfox                         0x10c95bc53 AsyncOpFetch.execute(txn:) (AsyncOpFetch.swift:69)
```

Upon looking more closely at `git_commit_list_parse`, I think it could be that

    git_odb_object *obj;

is not properly set to `NULL` on initialization.

pks-t:
When evicting cache entries, we first retrieve the object that is
to be evicted, delete the object and then finally delete the key
from the cache. In case where the cache eviction caused us to
free the cached object, though, its key will point to invalid
memory now when trying to remove it from the cache map. On my
system, this causes us to not properly remove the key from the
map, as its memory has been overwritten already and thus the key
lookup it will fail and we cannot delete it.

Fix this by only decrementing the refcount of the evictee after
we have removed it from our cache map. Add a test that caused a
segfault previous to that change.

Fixes #5081

### DIFF
--- a/src/cache.c
+++ b/src/cache.c
@@ -134,9 +134,8 @@ static void cache_evict_entries(git_cache *cache)
 
 		evict_count--;
 		evicted_memory += evict->size;
-		git_cached_obj_decref(evict);
-
 		git_oidmap_delete(cache->map, key);
+		git_cached_obj_decref(evict);
 	}
 
 	cache->used_memory -= evicted_memory;
--- a/tests/blame/simple.c
+++ b/tests/blame/simple.c
@@ -203,6 +203,14 @@ void test_blame_simple__trivial_libgit2(void)
 	check_blame_hunk_index(g_repo, g_blame, 49, 60, 1, 0, "d12299fe", "src/git.h");
 }
 
+/* This was leading to segfaults on some systems during cache eviction. */
+void test_blame_simple__trivial_libgit2_under_cache_pressure(void)
+{
+	ssize_t old_max_storage = git_cache__max_storage;
+	git_cache__max_storage = 1024 * 1024;
+	test_blame_simple__trivial_libgit2();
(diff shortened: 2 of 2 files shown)


pks-t:
Verified by @eaigner to fix the issue in #5081

ethomson:
Nice, thanks the helpful report @eaigner and the fix @pks-t.

Issue #4473, opened 2018-01-11
Request: Please consider publishing to conan.io


Phrohdoh:
Hello, libgit2 team / community / maintainers!

_Note: I am not affiliated with conan.io or JFrog in any way._

I am not a C developer by any means (I'm quite positive I would lose all of my limbs in a matter of minutes with C) which goes to show that I am also quite clueless regarding dependency management but from what I understand C (and C++) mostly consist of copying headers and implementation files of dependencies in-tree.

Given the above it is quite difficult, in my limited experience, to stay up to date (or even track which version of a library you are using).

It is my belief that more projects and developers (myself included of course) could benefit from the work put into libgit2 if it were easier to depend on in a trackable way.

This is where my request for you all (and if you decide to and I can help please let me know how!) to consider publishing libgit2 releases to [conan.io](https://conan.io/index.html) which markets itself as "Conan, the C / C++ Package Manager for Developers."

[`conan`](http://docs.conan.io/en/latest/getting_started.html) use is driven by the [`conanfile.txt`](http://docs.conan.io/en/latest/reference/conanfile_txt.html) which allows downstream projects to declare dependencies which the conan tool will resolve and build as specified in the manifest.

If the overhead of management and setup of publishing to conan.io outweighs the benefits then please note so and close this.

Thank you for your time!

**Edit**: Corrected a link.

ethomson:
Hi @phrohdoh - thanks for the suggestion, we'll take a look at [conan](http://conan.io/)!

impsnldavid:
I've had a go at an initial conan package recipe. 

https://github.com/impsnldavid/conan-libgit2
https://bintray.com/impsnldavid/public-conan/libgit2%3Aimpsnldavid

Having problems with libssh on Windows, but builds without this seem OK. Would be grateful for testers on other OS's.

memsharded:
There is also the original recipe from @p-brz: https://github.com/p-brz/libgit2-conan.git that was in the old conan.io server (now in read-only, to be deprecated conan-transit repository in bintray)
This recipe worked well prior to conan 1.0, but now might require updating it. 

The recipe from @impsnldavid looks very good, with little work it could be included in conan-center (though it first needs to have some of its dependencies that is still not in conan-center, but in the bincrafters repo, like libssh2)

danimtb:
[...]

PR #5091, opened 2019-06-02
Add Conan recipe

ovidiub13:
Hi, this library is used on a project I work on (Fritzing) and we're planning to start using Conan. We need this to be available as a Conan package so I've started work on writing a Conan recipe for libgit2.

I'm marking this as WIP, until I test it enough and develop the CI system to automate creating the packages for all platforms and compilers.

I'm making this PR so that the Conan and LibGit2 communities can review my proposal to ensure it's something that both communities are happy with.

This is in relation to https://github.com/libgit2/libgit2/issues/4473 and https://github.com/conan-io/wishlist/issues/225.


CMakeLists.txt
```diff
@@ -13,6 +13,10 @@
 
 PROJECT(libgit2 C)
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8.11)
+
+include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+conan_basic_setup()
```
pks-t: I don't think we want to include this in our CMakeLists.txt, as it would require everybody to build libgit2 with Conan.

CMakeLists.txt
```diff
@@ -43,28 +47,28 @@ INCLUDE(EnableWarnings)
...
+OPTION(USE_SSH					"Link with libssh2 to enable SSH support"		 			ON)
+OPTION(USE_HTTPS				"Enable HTTPS support. Can be set to a specific backend"	ON)
+OPTION(USE_GSSAPI				"Link with libgssapi for SPNEGO auth"						OFF)
+OPTION(USE_STANDALONE_FUZZERS	"Enable standalone fuzzers (compatible with gcc)"			OFF)
+OPTION(VALGRIND					"Configure build for valgrind"								OFF)
+OPTION(USE_EXT_HTTP_PARSER		"Use system HTTP_Parser if available"						ON)
+OPTION(DEBUG_POOL				"Enable debug pool allocator"								OFF)
+OPTION(ENABLE_WERROR			"Enable compilation with -Werror"							OFF)
+OPTION(USE_BUNDLED_ZLIB			"Use the bundled version of zlib"							OFF)
+OPTION(DEPRECATE_HARD			"Do not include deprecated functions in the library"		OFF)
```
pks-t: Please avoid such major reformatting, as it makes it quite hard to see what was actually changed

conanfile.py
```diff
@@ -0,0 +1,46 @@
+from conans import ConanFile, CMake
+
+
+class Libgit2Conan(ConanFile):
+    name = "libgit2"
+    version = "0.28.0"
```
pks-t: If we were to check in packaging recipes, then I'd really prefer to not include any changing bits like version numbers. There's quite a lot of different packaging systems, and having to update bits like these in all of them becomes a major burden on us, as we needn't forget to do so on every release.

Instead, I'd propose to parse the version number from our central source-of-truth version number, which is located in "include/git2/version.h".

conanfile.py
```diff
@@ -0,0 +1,46 @@
...
+    topics = ("<Put some tag here>", "<here>", "<and here>")
+
+    settings = "os", "compiler", "build_type", "arch"
+    options = {"shared": [True, False]}
+    default_options = "shared=False"
+
+    generators = "cmake"
+    # exports_sources = "CMakeLists.txt", "src/*", "include/*", "cmake/*", "deps/*", \
+    #     "tests/*", "libgit2.pc.in", "libgit2_clar.supp", "fuzzers/*"
+    exports_sources = "*"
```
pks-t: I'd expect that only sources in "src/*", "include/*" and "deps/*" are required

conanfile.py
```diff
@@ -0,0 +1,46 @@
...
+    no_copy_source = False
+
+    scm = {
+        "type": "git",
+        "subfolder": "hello",
```
pks-t: This looks like a leftover :)

conanfile.py
```diff
@@ -0,0 +1,46 @@
...
+    def package(self):
+        self.copy("*.h", dst="include", src="src")
+        self.copy("*.lib", dst="lib", keep_path=False)
+        self.copy("*.dll", dst="bin", keep_path=False)
+        self.copy("*.dylib*", dst="lib", keep_path=False)
+        self.copy("*.so", dst="lib", keep_path=False)
+        self.copy("*.a", dst="lib", keep_path=False)
+
+    def package_info(self):
+        self.cpp_info.libs = ["libgit2"]
```
pks-t: This would need "includedirs", too, right?

test_package/example.cpp
```diff
@@ -0,0 +1,6 @@
+#include <iostream>
+#include "hello.h"
```
pks-t: I doubt that this can actually be built via conan. There's no "hello.h", and furthermore I'd definitely say that such a test package should be written in plain C89, not C++

ovidiub13:
Hi @pks-t. Thank you for your reply.

Regarding supporting Conan, it basically only needs one file, the `conanfile.py`. The rest (the `test_package` directory) are just for testing the package to confirm it's built properly.

I've evaluated several package managers for C++ during my last job, and found only Conan to fulfill all the requirements for a package manager in the sense that other languages have. But this is just my opinion, and not necessarily a fact.

When I started writing this recipe, I was a bit unsure weather it should go under the libgit2 community maintenance or the Conan community's. I've decided to just write it first, and try with the libgit2 community. If this is not desired, I have no problem in moving the recipe under the Conan community's maintenance.

> The packaging instructions should be built in a way to keep down their maintenance cost.
I completely agree with this. As you can see this PR is marked as WIP just because of that. I wanted to make sure that I get the package working and after that to clean it up and make it low maintenance.

> I'd really prefer to not clutter our root-directory with files for such tools. 
I'd argue that the root of the project is exactly the right place for the Conan recipe (`conanfile.py`). But I have no problem moving the Conan files in a subdirectory.

Let's revisit this once it's ready for review.

Regarding it's readiness, I do apologize for the delay in finishing this, my free time is fairly limited. Any feedback is greatly appreciated.

CMakeLists.txt
ovidiub13: I agree. I plan to add this via an option. This is still in MVP phase.

CMakeLists.txt
ovidiub13: VS Code shows these lines as not whitespace aligned, while GH shows them aligned. I don't know why. I'll revert.

conanfile.py
ovidiub13: I agree. I'm still making the MVP, I'll update this after MVP is done.

conanfile.py
ovidiub13: I thought so too, but I've hit some issues with the testing tool `libgit2_clar` not being compiled, but then I remembered I forgot a Conan default that puts all binaries in a `bin` directory.
I need to discuss this with someone who knows libgit2 better than me.

conanfile.py
ovidiub13: oups...

conanfile.py
ovidiub13: yes

test_package/example.cpp
ovidiub13: I've updated this, but haven't pushed yet, since it's not a working build, and I don't want to have 1k commits, each with a minor fix.

pks-t:
> When I started writing this recipe, I was a bit unsure weather it should go under the libgit2 community maintenance or the Conan community's. I've decided to just write it first, and try with the libgit2 community. If this is not desired, I have no problem in moving the recipe under the Conan community's maintenance.

Fair enough. Let's see how it turns out and then we can still decide. At least no effort will be wasted in the end.

> Regarding it's readiness, I do apologize for the delay in finishing this, my free time is fairly limited. Any feedback is greatly appreciated.

No worries, take your time :) We all wish we'd have more time available to work on our pet projects, but reality often looks different

CMakeLists.txt
Croydon: Tip (for further recipes): Use `tools.replace_in_file()` within the Conan recipe or use an optional Conan CMake wrapper so that Conan stays optional

Croydon:
Conan has a new central service called Conan Center index, which builds packages for various configurations and also allows the community to more easily maintain recipes together

I recommend to close this pull request and start contributing here: https://github.com/conan-io/conan-center-index/pull/286

Thank you for contributing to the Conan ecosystem 😄

PR #5092, opened 2019-06-03
checkout: Fix removing untracked files by path in subdirectories

### LINKED ISSUE #5089: git_checkout_head FORCE | REMOVE_UNTRACKED doesn't remove untracked files in a subdirectory
### Reproduction steps
1. Save:

**main.c**:
```c
#include <stdio.h>
#include <git2.h>

int main(int argc, char *argv[]) {
	git_libgit2_init();	

	if (argc != 3) {
		fprintf(stderr, "Usage: ...");
		return 2;
	}

	git_repository *repo = NULL;
	int error = git_repository_open(&repo, argv[1]);
	if (error < 0) {
		const git_error *e = giterr_last();
		fprintf(stderr, "Error %d/%d: %s\n", error, e->klass, e->message);
		return 1;
	}

	git_checkout_options opts = GIT_CHECKOUT_OPTIONS_INIT;
	opts.checkout_strategy = GIT_CHECKOUT_FORCE | GIT_CHECKOUT_REMOVE_UNTRACKED;
	char *paths[] = {argv[2]};
	git_strarray paths_arr = {paths, 1};
	opts.paths = paths_arr;

	error = git_checkout_head(repo, &opts);
	if (error < 0) {
		const git_error *e = giterr_last();
		fprintf(stderr, "Error %d/%d: %s\n", error, e->klass, e->message);
		return 1;
	}
}
```

**run.sh**:
```sh
#!/bin/sh
set -e

gcc -g main.c -lgit2 -o discard-test

echo Preparing
rm -rf test-repo
git init test-repo
cd test-repo
mkdir code
touch code/run code/foo
git add code/run
git commit -m "Add code/run"

echo Before
git status

cd ..
./discard-test test-repo code/foo
cd test-repo

echo After
git status
```

2. Run `run.sh`. Tweak the gcc line as needed to get the example to compile.

(The example leaks memory, it's just an example, real code should free stuff of course)

### Expected behavior
"code/foo" is removed.

### Actual behavior
"code/foo" remains untouched.

ac070afecc33e9f263353455debd07876b0e91bf

Xubuntu 18.04.2 x86_64
macOS 10.14.5

segevfiner:
The checkout code didn't iterate into a subdir if it didn't match the pathspec, but since the pathspec might match files in the subdir we should recurse into it (In contrast to gitignore handling).

I'm not sure if there is a better way to handle this without fully recursing, nor that there aren't other places in the checkout code that require a similar change to fully recurse. Also the `git_iterator_advance_into` docs talks about some edge case that causes it to return `GIT_ENOTFOUND` which I'm not sure if needs special handling here or not.

Fixes #5089

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -373,8 +373,13 @@ static int checkout_action_wd_only(
 	if (!git_pathspec__match(
 			pathspec, wd->path,
 			(data->strategy & GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH) != 0,
-			git_iterator_ignore_case(workdir), NULL, NULL))
-		return git_iterator_advance(wditem, workdir);
+			git_iterator_ignore_case(workdir), NULL, NULL)) {
+		if (wd->mode != GIT_FILEMODE_TREE) {
+			return git_iterator_advance(wditem, workdir);
+		} else {
+			return git_iterator_advance_into(wditem, workdir);
+		}
+	}
 
 	/* check if item is tracked in the index but not in the checkout diff */
 	if (data->index != NULL) {
--- a/tests/checkout/head.c
+++ b/tests/checkout/head.c
@@ -109,6 +109,32 @@ void test_checkout_head__do_not_remove_untracked_file_in_subdir(void)
 	cl_assert(git_path_isfile("testrepo/tracked/subdir/untracked"));
 }
 
+void test_checkout_head__do_remove_untracked_paths(void)
+{
+	git_checkout_options opts = GIT_CHECKOUT_OPTIONS_INIT;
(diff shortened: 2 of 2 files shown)


ethomson:
Thanks for the PR @segevfiner - I'll have to dig into this a bit.  Indeed this is a tricky one so I want to make sure that we don't regress on perf in the standard case.  I'll try to take a look this week.  Thanks!

extrawurst:
@ethomson I am running into issue #5089 aswell. Any reason not to merge this?

ethomson:
Thanks @segevfiner - I swapped the order of the logic in the conditional for consistency and merged this manually.

PR #5095, opened 2019-06-06
ignore: handle escaped trailing whitespace

pks-t:
The gitignore's pattern format specifies that "Trailing spaces
are ignored unless they are quoted with backslash ("\")". We do
not honor this currently and will treat a pattern "foo\ " as if
it was "foo\" only and a pattern "foo\ \ " as "foo\ \".

The current gitignore format does offer some very special
surprises, though: given a pattern "foo \ ", upstream git will in
fact strip all spaces and treat it as "foo", only.

Fix our code to handle those special cases and add tests to avoid
regressions.

### DIFF
--- a/src/attr_file.c
+++ b/src/attr_file.c
@@ -560,13 +560,15 @@ void git_attr_path__free(git_attr_path *info)
  */
 
 /*
- * Determine the length of trailing spaces.
+ * Determine the length of trailing spaces. Escaped spaces do not count as
+ * trailing whitespace.
  */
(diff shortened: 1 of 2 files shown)


pks-t:
@ethomson: I first thought that the things you discovered were weird. But this "foo \ " being treated as "foo" tops it all

pks-t:
I take that back. Seems like I just failed to use git-check-ignore correctly. Embarassing :/ Will amend

PR #5098, opened 2019-06-07
Data-driven tests

pks-t:
tests: allow for simple data-driven tests

Right now, we're not able to use data-driven tests at all. E.g.
given a set of tests which we'd like to repeat with different
test data, one has to hand-write any outer loop that iterates
over the test data and then call each of the test functions. Next
to being bothersome, this also has the downside that error
reporting is quite lacking, as one never knows which test data
actually produced failures.

So let's implement the ability to re-run complete test modules
with changing test data. To retain backwards compatibility and
enable trivial addition of new runs with changed test data, we
simply extend clar's generate.py. Instead of scanning for a
single `_initialize` function, each test module may now implement
multiple `_initialize_foo` functions. The generated test harness
will then run all test functions for each of the provided
initializer functions, making it possible to provide different
test data inside of each of the initializer functions. Example:

```
void test_git_example__initialize_with_nonbare_repo(void)
{
        g_repo = cl_git_sandbox_init("testrepo");
}

void test_git_example__initialize_with_bare_repo(void)
{
        g_repo = cl_git_sandbox_init("testrepo.git");
}

void test_git_example__cleanup(void)
{
        cl_git_sandbox_cleanup();
}

void test_git_example__test1(void)
{
        test1();
}

void test_git_example__test2(void)
{
        test2();
}
```

Executing this test module will cause the following output:

```
$ ./libgit2_clar -sgit::example
git::example (with nonbare repo)..
git::example (with bare repo)..
```

---

The goal was basically to have the least-intrusive change that gives us the most bang for the buck. No changes were required for any of our existing test modules, but we can now easily set up multiple test runs of a complete module by just specifying multiple initializers.

### DIFF
--- a/tests/generate.py
+++ b/tests/generate.py
@@ -24,8 +24,8 @@ class Module(object):
         def render(self):
             out = "\n".join("extern %s;" % cb['declaration'] for cb in self.module.callbacks) + "\n"
 
-            if self.module.initialize:
-                out += "extern %s;\n" % self.module.initialize['declaration']
+            for initializer in self.module.initializers:
+                out += "extern %s;\n" % initializer['declaration']
 
             if self.module.cleanup:
                 out += "extern %s;\n" % self.module.cleanup['declaration']
@@ -41,7 +41,19 @@ class Module(object):
 
     class InfoTemplate(Template):
         def render(self):
-            return Template(
+            templates = []
+
+            initializers = self.module.initializers
+            if len(initializers) == 0:
+                initializers = [ None ]
+
+            for initializer in initializers:
+                name = self.module.clean_name()
+                if initializer and initializer['short_name'].startswith('initialize_'):
+                    variant = initializer['short_name'][len('initialize_'):]
+                    name += " (%s)" % variant.replace('_', ' ')
+
+                template = Template(
             r"""
     {
         "${clean_name}",
@@ -49,14 +61,17 @@ class Module(object):
     ${cleanup},
         ${cb_ptr}, ${cb_count}, ${enabled}
     }"""
-            ).substitute(
-                clean_name = self.module.clean_name(),
-                initialize = self._render_callback(self.module.initialize),
-                cleanup = self._render_callback(self.module.cleanup),
-                cb_ptr = "_clar_cb_%s" % self.module.name,
-                cb_count = len(self.module.callbacks),
-                enabled = int(self.module.enabled)
-            )
+                ).substitute(
+                    clean_name = name,
+                    initialize = self._render_callback(initializer),
+                    cleanup = self._render_callback(self.module.cleanup),
+                    cb_ptr = "_clar_cb_%s" % self.module.name,
+                    cb_count = len(self.module.callbacks),
+                    enabled = int(self.module.enabled)
+                )
+                templates.append(template)
+
+            return ','.join(templates)
 
     def __init__(self, name):
         self.name = name
@@ -86,7 +101,7 @@ class Module(object):
         regex = re.compile(TEST_FUNC_REGEX % self.name, re.MULTILINE)
 
         self.callbacks = []
-        self.initialize = None
+        self.initializers = []
         self.cleanup = None
 
         for (declaration, symbol, short_name) in regex.findall(contents):
@@ -96,8 +111,8 @@ class Module(object):
                 "symbol" : symbol
             }
 
-            if short_name == 'initialize':
-                self.initialize = data
+            if short_name.startswith('initialize'):
+                self.initializers.append(data)
             elif short_name == 'cleanup':
                 self.cleanup = data
             else:
--- a/tests/object/cache.c
+++ b/tests/object/cache.c
@@ -2,10 +2,35 @@
 #include "repository.h"
 
 static git_repository *g_repo;
+static size_t cache_limit;
+static int object_type;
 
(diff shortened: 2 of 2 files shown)


ethomson:
Well, isn't this clever.  I'm just driving by at the moment, so I'll need to put a little more 👀 on the implementation but I love the idea.

Hey @vmg, we've been making a lot of changes (okay, not _a lot_, but also not zero) to clar lately.  Do you want us to try to upstream these back to you, or...?

vmg:
Nah, clearly the version here in libgit2 is the "definitive" version of Clar. I may just delete my repo and point people here. You're way ahead. :sweat_smile:

ethomson:
> Nah, clearly the version here in libgit2 is the "definitive" version of Clar. I may just delete my repo and point people here. You're way ahead. 😅

I like that it's broken out into a separate repo though (I use it in other projects as well).  🤔

vmg:
Could we place it in a separate folder here in libgit2 so it's easier to copy&paste into other projects?

ethomson:
If you're anxious to get rid of https://github.com/vmg/clar then we can either fork that over to the @libgit2 account (or to one of our personal accounts) and make that the new canonical place.  Don't know if anybody else has a strong preference.

I lean away from @libgit2 a little bit, because I don't want anybody to think that the Team As A Whole supports it.

tiennou:
Same here, I'd really like `clar` to keep a "canonical" home, outside of any other project, whether it's as part of the org or where it still lives (no strong opinion). I've tried once to try to resync with upstream via a subtree merge, but it didn't end well (IIRC, there were conflicting/incomplete fixes for a bunch of the same issues in both "ours" and "theirs", plus the XML report generation).

It might be possible to import it as a git submodule instead ? Not sure how we feel about that though, but it might be better located as `tests/deps{,/clar}` or something, so tracking "upstream" is easier.

I've started to use it as the test harness in libssh2, and I'm already carrying some custom stuff there (more vararg-y macros, so tests can build more complex messages). So I'm fine to be added to some @libgit2/clar team, if we feel one is needed.

pks-t:
@tiennou: happy to convert more in case you know of some where it would make sense to do so! Merge analysis in #5017  is an obvious next one, but that's not merged yet.

> If you're anxious to get rid of https://github.com/vmg/clar then we can either fork that over to the @libgit2 account (or to one of our personal accounts) and make that the new canonical place. Don't know if anybody else has a strong preference.

I don't. I feel like I'm usually the one driving for new features in clar for the libgit2 project, so I might as well fork it into my own personal namespace. But in general, I always prefer the more "official" way of using an org for that. "clar" is taken already, but we may opt for something like "clar-tests". I'd also prefer the second approach in order to not become the sole maintainer, but instead to have a team of maintainers like in libgit2.

> Could we place it in a separate folder here in libgit2 so it's easier to copy&paste into other projects?

I agree that we should probably refactor this. Currently, we're quite inconsistent with regards to third-party dependencies, where we have some in "deps/" and some mixed into our own sources. If we agree to do that, then I might clean it up in the near-term future.

tiennou:
> happy to convert more in case you know of some where it would make sense to do so!

I was thinking about the worktree/bare/non-bare. Arguably, I'm looking very hard at most of `repo`, because running that using a bunch of normal, bare, shallow, in-memory, mostly-memory, repositories would make the most sense, but it's harder (though maybe taking a look at "how" would benefit the proposed implementation).

pks-t:
On Tue, Jun 11, 2019 at 12:52:00AM -0700, Etienne Samson wrote:
> > happy to convert more in case you know of some where it would make sense to do so!
> 
> I was thinking about the worktree/bare/non-bare. Arguably, I'm
> looking very hard at most of `repo`, because running that using
> a bunch of normal, bare, shallow, in-memory, mostly-memory,
> repositories would make the most sense, but it's harder (though
> maybe taking a look at "how" would benefit the proposed
> implementation).

I agree, but I think we should keep it out of this PR. As you
say, it's probably quite involved and will require some
refactoring/moving around, while the test case presented here is
a rather simple conversion that serves as a nice demonstration of
how to use the new functionality.

ethomson:
> I agree, but I think we should keep it out of this PR.

👍 

This is neat, I like it a lot.  Thanks @pks-t.

pks-t:
Thanks for merging, @ethomson! We still didn't agree on how to fare with clar, though. Should we maintain it as a separate project "clar-{framework,tests,testing}/clar" (proposals welcome!) and then re-synchronize our own copy with such an upstream project?

tiennou:
I've just filed https://github.com/vmg/clar/pull/81, which carries our current changes.

pks-t:
Cool, thanks for your effort! If @vmg wants to keep maintaing clar then that's probably the best way ahead. To me it sounded like he didn't, though, but it might be that I interpreted too much into his words. :)

ethomson:
Sounds like @vmg would like to create an org for this, so I've created https://github.com/clar-test/clar and invited all of you.  @vmg do you want to do that magic to make that one the root of the network?  I forget how that works.

vmg:
@ethomson I think I need admin access to the `clar` repo in the new org before I can do that!

ethomson:
Fixed!

ethomson:
@vmg I've also deleted https://github.com/clar-test/clar, so you're welcome to do that transfer if you'd like.

PR #5099, opened 2019-06-07
tests: checkout: fix symlink.git being created outside of sandbox

pks-t:
The function `populate_symlink_workdir` creates a new
"symlink.git" repository with a relative path "../symlink.git".
As the current working directory is the sandbox, the new
repository will be created just outside of the sandbox.

Fix this by using `clar_sandbox_path`.


ethomson:
Nice find.  This is indeed bothersome.

PR #5101, opened 2019-06-08
Rename options initialization functions

ethomson:
Rename opt init functions to `options_init`

In libgit2 nomenclature, when we need to verb a direct object, we name a function `git_directobject_verb`.  Thus, if we need to init an options structure named `git_foo_options`, then the name of the function that does that should be `git_foo_options_init`.

The previous names of `git_foo_init_options` is close - it _sounds_ as if it's initializing the options of a `foo`, but in fact `git_foo_options` is its own noun that should be respected.

Deprecate the old names; they'll now call directly to the new ones.

Additionally, add an initializer function for `git_apply_options`.

### DIFF
--- a/tests/diff/blob.c
+++ b/tests/diff/blob.c
@@ -39,7 +39,7 @@ void test_diff_blob__initialize(void)
 
 	g_repo = cl_git_sandbox_init("attr");
 
-	cl_git_pass(git_diff_init_options(&opts, GIT_DIFF_OPTIONS_VERSION));
+	cl_git_pass(git_diff_options_init(&opts, GIT_DIFF_OPTIONS_VERSION));
 	opts.context_lines = 1;
 
 	memset(&expected, 0, sizeof(expected));
--- a/examples/describe.c
+++ b/examples/describe.c
@@ -63,7 +63,7 @@ static void do_describe_single(git_repository *repo, describe_options *opts, con
 	git_object *commit;
 	git_describe_result *describe_result;
 	git_buf buf = { 0 };
-	
+
 	if (rev) {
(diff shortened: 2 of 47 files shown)


tests/diff/blob.c
```diff
@@ -520,19 +520,19 @@ void test_diff_blob__can_compare_a_binary_blob_and_a_text_blob(void)
  * +++ b/a0f7217
  * @@ -1,6 +1,6 @@
  *  Here is some stuff at the start
- * 
+ *
```
tiennou: Spurious whitespace change ? The build was red, but I haven't yet looked at the build log.

ethomson:
> The build was red

Thanks, I forgot to build locally with hard deprecation enabled.

ethomson:
Hey @pks-t, are you also 👍 on renaming these?  I think that they're more "correct" this way, but naming is pretty subjective.

pks-t:
I'm :+1: Those names have bothered me several times, so the change is very welcome. You forgot to convert one instance in "src/proxy.c", otherwise this looks good to me!

ethomson:
> I'm 👍 Those names have bothered me several times, so the change is very welcome. You forgot to convert one instance in "src/proxy.c", otherwise this looks good to me!

I'm not seeing it, I only see `git_proxy_options_init` as the new impl (which `git_proxy_init_options` delegates to).

pks-t:
`git_proxy_options_dup` still makes use of `git_proxy_init_options`.

ethomson:
> `git_proxy_options_dup` still makes use of `git_proxy_init_options`.

Aha - yes!  I missed that twice.  Thanks, and fixed.

PR #5108, opened 2019-06-11
Handle URLs with a colon after host but no port

### LINKED ISSUE #5100: Handling of port component in URLs (with schema) differs from git
### Reproduction steps

build libgit2 with examples (`lg2`), then

    mkdir test
    cd test
    git init
    git remote add origin https://github.com:/libgit2/libgit2.git
    lg2 fetch origin

### Expected behavior

The repository is fetched 

### Actual behavior

    Bad news:
     malformed URL 'https://github.com:/libgit2/libgit2.git'

e50d138e898dd034161663873f75716086266f86

Gentoo linux

ethomson:
Core git copes with URLs that have a colon after the port, but no actual numeric value.  eg `http://example.com:/foo.git` or `http://example.com:`.  That's horrible, but RFC 3986 says:

> URI producers and normalizers should omit the port component and its
> ":" delimiter if port is empty or if its value would be the same as
> that of the scheme's default.

Which indicates that they may and therefore we must accept it.

Test that we can handle URLs with a colon but no following port number, fix our http-parser implementation, and require use of our http-parser which corrects this issue.

Fixes #5100.


ethomson:
I suspect this will be contentious.  😀

ethomson:
I updated this to prefer out http-parser but allow users to specify the system http-parser if they want, even sans a version check, because this bug really _isn't_ that serious.  Basically, it allows them to opt-in to very slightly broken URL parsing if they want.

I updated the CMake option for this so that users with an existing option wouldn't _accidentally_ opt-in to it, which also allowed us to align closer to the zlib option.

CMakeLists.txt
```diff
@@ -60,10 +60,10 @@ OPTION(USE_HTTPS			"Enable HTTPS support. Can be set to a specific backend" ON)
 OPTION(USE_GSSAPI			"Link with libgssapi for SPNEGO auth"			OFF)
 OPTION(USE_STANDALONE_FUZZERS		"Enable standalone fuzzers (compatible with gcc)"	OFF)
 OPTION(VALGRIND				"Configure build for valgrind"				OFF)
-OPTION(USE_EXT_HTTP_PARSER		"Use system HTTP_Parser if available"			 ON)
 OPTION(DEBUG_POOL			"Enable debug pool allocator"				OFF)
 OPTION(ENABLE_WERROR			"Enable compilation with -Werror"			OFF)
 OPTION(USE_BUNDLED_ZLIB    		"Use the bundled version of zlib"			OFF)
+OPTION(USE_BUNDLED_HTTP_PARSER  	"Use the bundled HTTP Parser"				 ON)
```
pks-t: Ah, didn't know we already had an option. If we're about to change, then we should use a generic `USE_HTTP_PARSER=(bundled|system)`, if you ask me. The reasoning is that this would be much easier to extend. For example at a future point, we'd be able to add an option `USE_HTTP_PARSER=/path/to/http_parser`.

docs/changelog.md
```diff
@@ -1,3 +1,18 @@
...
+---------
+
+### Changes or improvements
+
+* The CMake option to use a system http-parser library, instead of the
+  bundled dependency, has changed.  This is due to a deficiency in
+  http-parser that we have fixed in our implementation.  The bundled
+  library is now the default, but if you wish to force the use of the
+  system http-parser implementation despite incompatibilities, you can
+  specify `-DUSE_BUNDLED_HTTP_PARSER=OFF` to CMake.
```
pks-t: We should treat changes to CMake options as breaking changes and thus highlight this in a separate section

ethomson: Generally we have a section titled "Breaking API Changes" at the bottom.  CMake changes don't feel like they fit there, so I created a new section "Breaking CMake configuration changes" and slotted this in there.  I also moved the "breaking" stuff to the top.  We should be done breaking APIs, but if not, it should be the first thing people see in the release notes, IMO.

pks-t:
On Thu, Jun 13, 2019 at 11:58:11AM -0700, Edward Thomson wrote:
> > @@ -1,3 +1,18 @@
> +v0.28 + 1
> +---------
> +
> +### Changes or improvements
> +
> +* The CMake option to use a system http-parser library, instead of the
> +  bundled dependency, has changed.  This is due to a deficiency in
> +  http-parser that we have fixed in our implementation.  The bundled
> +  library is now the default, but if you wish to force the use of the
> +  system http-parser implementation despite incompatibilities, you can
> +  specify `-DUSE_BUNDLED_HTTP_PARSER=OFF` to CMake.
> 
> Generally we have a section titled "Breaking API Changes" at
> the bottom.  CMake changes don't feel like they fit there, so I
> created a new section "Breaking CMake configuration changes"
> and slotted this in there.  I also moved the "breaking" stuff
> to the top.  We should be done breaking APIs, but if not, it
> should be the first thing people see in the release notes, IMO.

Fully agreed, thanks!

ethomson:
Upstream: https://github.com/nodejs/http-parser/pull/483

PR #5109, opened 2019-06-13
tests: merge::analysis: use test variants to avoid duplicated test suites

pks-t:
Since commit 394951ad4 (tests: allow for simple data-driven
tests, 2019-06-07), we have the ability to run a given test suite
with multiple variants. Use this new feature to deduplicate the
test suites for merge::{trees,workdir}::analysis into a single
test suite.


ethomson:
NIce.  Thanks!

Issue #5093, opened 2019-06-03
Consecutive asterisks are not collapsed to regular asterisks when they can't be interpreted as directories


nathansobo:
In Atom, we're using libgit2 to determine whether files are ignored. We recently upgraded, and got a user complaint about a file that was previously shown as ignored being shown as not ignored even though Git continued to ignore it (https://github.com/atom/tree-view/issues/1317).

It turns out that the user was incorrectly using the `**` pattern in their file, with a `**.log` pattern to ignore all log files. They could have just used `*.log` instead, but apparently Git automatically collapses `**` to `*` in situations where `**` does not make sense.

In the documentation provided by `man gitignore`, the following is stated about double asterisks in gitignore patterns:

> Two consecutive asterisks (`**`) in patterns matched against full pathname may have special meaning:
>
> * A leading `**` followed by a slash means match in all directories. For example, `**/foo` matches file or directory `foo` anywhere, the same as pattern `foo`. `**/foo/bar` matches file or directory `bar` anywhere that is directly under directory `foo`.
> * A trailing `/**` matches everything inside. For example, `abc/**` matches all files inside directory `abc`, relative to the location of the .gitignore file, with infinite depth.
> * A slash followed by two consecutive asterisks then a slash matches zero or more directories. For example, `a/**/b` matches `a/b`, `a/x/b`, `a/x/y/b` and so on.
> * **Other consecutive asterisks are considered regular asterisks and will match according to the previous rules.**

Note the final bullet, which I have bolded.

I'm not sure how much we care about libgit2 matching the exact behavior of Git, and this admittedly a pretty strange edge case. We can tell the user to fix thier ignore file, but I thought it was worth reporting anyhow.

### Reproduction steps

This test can be added to `tests/attr/ignore.c`

```c
void test_attr_ignore__collapse_spurious_stars(void)
{
	cl_git_rewritefile(
		"attr/.gitignore",
		"**.log\n"
	);

	assert_is_ignored(true, "a.log");
	assert_is_ignored(true, "dir1/a.log");
}
```

### Expected behavior

The assertions should pass, indicating that `a.log` and `dir1/a.log` are ignored.

### Actual behavior

The assertions fail.

### Version of libgit2 (release number or SHA1)

ac070afecc33e9f263353455debd07876b0e91bf

### Operating system(s) tested

macOS 10.14.4 (18E226)

pks-t:
[...]

Issue #5068, opened 2019-05-15
Configuration file: Incorrect matching of conditional includes


tmat:
### Reproduction steps

`libgit2` uses `FNM_PATHNAME|FNM_LEADING_DIR` `fnmatch` flags to match the conditional include pattern, but it seems that git does not do so.

1) libgit2 matches a directory prefix of git directory while git does not.
2) libgit2 does not match wildcards `?`, `*` and `[]` in the pattern to `/` in the path, while git does.

Consider repository with git directory `D:\repository\.git` with `config` file that contains 

```
[includeIf "gitdir:D:/repository"] 
  path = cfg1
[includeIf "gitdir:D:/repository?.git"] 
  path = cfg2
```

and configuration files `D:\repository\.git\cfg1` and `D:\repository\.git\cfg2` with content:

```
[a]
  b = 1
```

and

```
[a]
  c = 1
```

respectively.

git does not include `cfg` file:

```
> git config --list
...
includeif.gitdir:D:/repository.path=cfg1
includeif.gitdir:D:/repository?.git.path=cfg2
a.c=1
```

### Version of libgit2 (release number or SHA1)

a904fc6d12ffc20aa1a238a78000921494222e89

### Operating system(s) tested

Windows

pks-t:
Yeah, our "gitdir:" includes have a heavy mismatch in semantics compared to git.git. I accidentally stumbled upon this issue yesterday when taking a look at why our gitignore parser behaves differently: turns out git.git has migrated to use wildmatch instead of fnmatch in 2014, which caused quite a lot of incompatibilities.

So I've migrated our code to use wildmatch, too, in #5110. This PR fixes several issues in our config parser as well as in our conditional include handling.

I'm a bit confused, though. The first includeIf was obviously a fault on our side, and that issue got fixed via #5110. But the second condition "D:/repository?.git" shouldn't match at all, but from your `git config --list` I can see that "a.c=1" is in fact included. libgit2 does the right thing with #5110, and furthermore I'm unable to reproduce this on my (Linux) system with git v2.22.0:

```
$ git config --list
...
includeif.gitdir:/tmp/test.path=cfg1
includeif.gitdir:/tmp/test?.git.path=cfg2
```

Which version of git were you using?

tmat:
Using git version 2.17.1.windows.2

pks-t:
[...]

PR #5110, opened 2019-06-13
Replace fnmatch with wildmatch

pks-t:
Upstream git.git has replaced fnmatch with wildmatch back in commit 70a8fc999d (stop using fnmatch (either native or compat), 2014-02-15). wildmatch has different semantics than fnmatch has and thus may show slightly different behaviour than fnmatch. In fact, by replacing fnmatch with wildmatch, we fix compatibility with git.git in at least two places:

- gitignores and gitattributes now correctly handle patterns like "**.foo", which previously wouldn't match a file "x.foo"
- conditional includes now work (more?) consistently with how it works upstream

There's probably more compatibility improvements as fnmatch/wildmatch are used at various locations throught the tree, but only those two struck me while working on the replacement.

Anyway. Getting this merged will have to wait until @ethomson has determined whether we're even allowed to use "wildmatch.c" from git.git, depending on its license history.


pks-t:
Fixes #5093

pks-t:
Rebased to fix conflicts with #5097

pks-t:
Fixes #5068

ethomson:
Would you mind adding a paragraph to `COPYING`:

```
The bundled wildmatch code is licensed under the BSD license:

Copyright Rich Salz.
All rights reserved.

Redistribution and use in any form are permitted provided that the
following restrictions are are met:

1.  Source distributions must retain this entire copyright notice
    and comment.
2.  Binary distributions must include the acknowledgement ``This
    product includes software developed by Rich Salz'' in the
    documentation or other materials provided with the
    distribution.  This must not be represented as an endorsement
   or promotion without specific prior written permission.
3.  The origin of this software must not be misrepresented, either
    by explicit claim or by omission.  Credits must appear in the
    source and documentation.
4.  Altered versions must be plainly marked as such in the source
    and documentation and must not be misrepresented as being the
    original software.

THIS SOFTWARE IS PROVIDED ``AS IS'' AND WITHOUT ANY EXPRESS OR IMPLIED
WARRANTIES, INCLUDING, WITHOUT LIMITATION, THE IMPLIED WARRANTIES OF
MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
```

pks-t:
> Would you mind adding a paragraph to COPYING

Done!

ethomson:
Nice one, thanks for this @pks-t!

PR #5111, opened 2019-06-13
Documentation fixes

tiennou:
I've made some runs of my custom docurium, and here are the results.

I'm not sure how commit 1 is possible but you never know ^^. Note that I'm adding documentation for things which are just out of one-pass-declaration-parsing, so either we're fine with them as a first approximation, or I'll need help to clarify (a.k.a. help is welcome).

### DIFF
--- a/include/git2/repository.h
+++ b/include/git2/repository.h
@@ -640,6 +640,18 @@ GIT_EXTERN(int) git_repository_message_remove(git_repository *repo);
  */
 GIT_EXTERN(int) git_repository_state_cleanup(git_repository *repo);
 
+/**
+ * Callback used to iterate over each FETCH_HEAD entry
+ *
+ * @see git_repository_fetchhead_foreach
+ *
+ * @param ref_name The reference name
+ * @param remote_url The remote URL
+ * @param oid The reference target OID
+ * @param is_merge Was the reference the result of a merge
+ * @param payload Payload passed to git_repository_fetchhead_foreach
+ * @return non-zero to terminate the iteration
+ */
 typedef int GIT_CALLBACK(git_repository_fetchhead_foreach_cb)(const char *ref_name,
 	const char *remote_url,
 	const git_oid *oid,
@@ -662,6 +674,15 @@ GIT_EXTERN(int) git_repository_fetchhead_foreach(
 	git_repository_fetchhead_foreach_cb callback,
 	void *payload);
 
+/**
+ * Callback used to iterate over each MERGE_HEAD entry
+ *
+ * @see git_repository_mergehead_foreach
+ *
+ * @param oid The merge OID
+ * @param payload Payload passed to git_repository_mergehead_foreach
+ * @return non-zero to terminate the iteration
+ */
 typedef int GIT_CALLBACK(git_repository_mergehead_foreach_cb)(const git_oid *oid,
 	void *payload);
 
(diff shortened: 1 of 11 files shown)


tiennou:
Not sure what to make of it, but there are a lots of holes caused by things I've "fixed" in docurium.

For example, our [filter](https://libgit2.org/libgit2/#HEAD/group/filter) groups are [different](http://tiennou.github.io/libgit2/#HEAD/group/filter). I have [hashsig](http://tiennou.github.io/libgit2/#HEAD/group/hashsig), which doesn't show up at all now. I think that's the docparser-cleanups branch I have (passing the tempdir as the include path), I'll try to confirm. Not trying to be dense, I'm having a hard time remembering what's fixing what where how 😜.

On the other hand, [git_odb_backend](https://libgit2.org/libgit2/#HEAD/type/git_odb_backend) (whose "ours" version should look better after this PR gets in), which I ought to be still mangling because the typedef in "types.h" gets parsed after the struct "sys/odb_backend.h", overwriting the comments).

pks-t:
> I'm not sure how commit 1 is possible but you never know ^^

You mean why there's no compile errors due to unknown prototypes/functions? Most users include "git2.h" directly which transitively includes all of our external headers, including "indexer.h". But docurium wouldn't do that, hence why it will probably fail to generate proper documentation here.

include/git2/repository.h
```diff
@@ -640,6 +640,18 @@ GIT_EXTERN(int) git_repository_message_remove(git_repository *repo);
  */
 GIT_EXTERN(int) git_repository_state_cleanup(git_repository *repo);
 
+/**
+ * Callback used to iterate over each FETCH_HEAD entries
```
pks-t: Typo: "entry"

include/git2/sys/filter.h
```diff
@@ -197,6 +197,9 @@ typedef int GIT_CALLBACK(git_filter_apply_fn)(
 	const git_buf *from,
 	const git_filter_source *src);
 
+/**
+ * Callback used for filter streaming
+ */
```
pks-t: I guess there's no parameter descriptions here as you weren't able to quite make sense of it? Fair enough if that's the case, I'm not able to make much sense of it either without diving into code.

tiennou: Exactly. I'm not sure what's the correct way forward though: this will silence the docurium warning, since it's now documented. But the "cleverer" checks are made via `-Wdocumentation`, and by design it ignores missing docblocks/directives (hence no missing `@param` arg).

include/git2/repository.h
tiennou: Isn't `FETCH_HEAD` a file whose contents a list of "one or more entries" ? I find it grammatically confusing that you can singularize over "one or more entries"… I can rephrase as "Callback used to iterate over FETCH_HEAD contents`, but that's still plural 🤣.

pks-t: It is. But you're not iterating over each entries, but over each entry :)

include/git2/sys/filter.h
tiennou: I've dropped that comment, for now. Let's wait for some pressure to investigate to show up 😉.

tiennou:
Rebased, should be ready to 🚢.

ethomson:
Looks good to me.  Thanks, @tiennou!

PR #5112, opened 2019-06-14
deps: ntlmclient: disable implicit fallthrough warnings

pks-t:
The ntlmclient dependency has quite a lot of places with implicit
fallthroughs. As at least modern GCC has enabled warnings on
implicit fallthroughs by default, the developer is greeted with a
wall of warnings when compiling that dependency.

Disable implicit fallthrough warnings for ntlmclient to fix this
issue.

### DIFF
--- a/deps/ntlmclient/CMakeLists.txt
+++ b/deps/ntlmclient/CMakeLists.txt
@@ -2,6 +2,8 @@ FILE(GLOB SRC_NTLMCLIENT "ntlm.c" "unicode_builtin.c" "util.c")
 
 ADD_DEFINITIONS(-DNTLM_STATIC=1)
 
+DISABLE_WARNINGS(implicit-fallthrough)
+
 IF (HTTPS_BACKEND STREQUAL "SecureTransport")
 	ADD_DEFINITIONS(-DCRYPT_COMMONCRYPTO)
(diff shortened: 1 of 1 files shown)


tiennou:
LGTM (nightlies-only bionic builds failed because of that). Though I see
```
/src/deps/ntlmclient/unicode_builtin.c: At top level:
cc1: error: unrecognized command line option '-Wno-documentation-deprecated-sync' [-Werror]
cc1: all warnings being treated as errors
```
Hopefully this is just a confused GCC ?

pks-t:
> Hopefully this is just a confused GCC ?

Yeah, I don't really know where that comes from, but we had that issue since several months already. I've noticed it multiple times, but I've got no idea when exactly it pops up and I didn't think it important enough to really care

ethomson:
I'm happy enough to fix this in ntmlclient itself, if you'd prefer.  My goal in building ntlmclient was for it to basically _be_ part of libgit2 (in terms of codestyle, etc) but available for others as well.  🤷‍♂

pks-t:
@ethomson: how about merging this as a short-term fix with the mid-term goal being to fix it in ntlmclient?

ethomson:
Welp, I tried going through `unicode_builtin` to add `/* fallthrough */` comments, but gcc appears confused about the switch-within-a-switch, so...  I think that `no-implicit-fallthrough` is probably the best approach.  🙇

Issue #3910, opened 2016-08-25
git_stash_save really slow on large repositories


mikesart:
Calling git_stash_save with libigt2 v0.24.1 on the linux repository takes 8 seconds to stash a single untracked file on my machine. Using the git command takes less than 2 seconds. I am running on the current version of Debian Stretch: "4.6.0-1-amd64 #1 SMP Debian 4.6.4-1 (2016-07-18) x86_64 GNU/Linux"

I've attached a simple cpp sample, and put it in a gist as well.

[stashtest.zip](https://github.com/libgit2/libgit2/files/436167/stashtest.zip)
https://gist.github.com/mikesart/a99585317a8f57f0f57b9e5bd17d37c0

It appears the time is going into git_diff_tree_to_workdir and git_diff_index_to_workdir. I will update this bug report with the results on Windows tomorrow.

To reproduce:

mikesart@mikesart-cos:~/dev/linux/linux-src (master)$ git remote -v
origin  git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git (fetch)
origin  git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git (push)

mikesart@mikesart-cos:~/dev/linux/linux-src (master)$ git rev-parse HEAD
4935e04ef431c7980ba6e8145ad4c697abccf6b9

mikesart@mikesart-cos:~/dev/linux/linux-src (master)$ echo "blah" > blah.txt

mikesart@mikesart-cos:~/dev/linux/linux-src (master)$ git status --short
?? blah.txt

mikesart@mikesart-cos:~/dev/linux/linux-src (master)$ time git stash -u
Saved working directory and index state WIP on master: 4935e04 Merge branch 'for-linus-4.8-rc4' of git://git.kernel.org/pub/scm/linux/kernel/git/rw/uml
HEAD is now at 4935e04 Merge branch 'for-linus-4.8-rc4' of git://git.kernel.org/pub/scm/linux/kernel/git/rw/uml

**real    0m1.864s**
user    0m1.336s
sys     0m0.796s

mikesart@mikesart-cos:~/dev/linux/linux-src (master)$ stashtest .
Opening repository '.'
Calling git_repository_open_ext...
  0.47ms
Calling git_signature_default...
  0.09ms
Calling git_stash_save...
  **7935.53ms**

mikesart:
[...]

PR #5113, opened 2019-06-14
stash: avoid recomputing tree when committing worktree

pks-t:
When creating a new stash, we need to create there separate
commits storing differences stored in the index, untracked
changes as well as differences in the working directory. The
first two will only be done conditionally if the equivalent
options "git stash --keep-index --include-untracked" are being
passed to `git_stash_save`, but even when only creating a stash
of worktree changes we're much slower than git.git. Using our new
stash example:

        $ time git stash
        Saved working directory and index state WIP on (no branch): 2f7d9d47575e Linux 5.1.7

        real    0m0.528s
        user    0m0.309s
        sys     0m0.381s

        $ time lg2 stash

        real    0m27.165s
        user    0m13.645s
        sys     0m6.403s

As can be seen, libgit2 is more than 50x slower than git.git!

When creating the stash commit that includes all worktree
changes, we create a completely new index to prepare for the new
commit and populate it with the entries contained in the index'
tree. Here comes the catch: by populating the index with a tree's
contents, we do not have any stat caches in the index. This means
that we have to re-validate every single file from the worktree
and see whether it has changed.

The issue can be fixed by populating the new index with the
repo's existing index instead of with the tree. This retains all
stat cache information, and thus we really only need to check
files that have changed stat information. This is semantically
equivalent to what we previously did: previously, we used the
tree of the commit computed from the index. Now we're just using
the index directly.

And, in fact, the cache is doing wonders:

        time lg2 stash

        real    0m1.836s
        user    0m1.166s
        sys     0m0.663s

We're now performing 15x faster than before and are only 3x
slower than git.git now.

---

Note that this also contains a new stash example. I'd been too lazy to come up with a more complex setup, so I decided to just implement "lg2 stash" and use that to test performance.

### DIFF
--- a/examples/stash.c
+++ b/examples/stash.c
@@ -1,0 +1,157 @@
+/*
+ * libgit2 "stash" example - shows how to use the stash API
+ *
+ * Written by the libgit2 contributors
+ *
+ * To the extent possible under law, the author(s) have dedicated all copyright
+ * and related and neighboring rights to this software to the public domain
+ * worldwide. This software is distributed without any warranty.
+ *
+ * You should have received a copy of the CC0 Public Domain Dedication along
+ * with this software. If not, see
+ * <http://creativecommons.org/publicdomain/zero/1.0/>.
+ */
+
+#include <stdarg.h>
+
+#include "common.h"
+
+enum subcmd {
+	SUBCMD_APPLY,
+	SUBCMD_LIST,
+	SUBCMD_POP,
+	SUBCMD_PUSH
+};
+
+struct opts {
+	enum subcmd cmd;
+	int argc;
+	char **argv;
+};
+
+static void usage(const char *fmt, ...)
+{
+	va_list ap;
+
+	fputs("usage: git stash list\n", stderr);
+	fputs("   or: git stash ( pop | apply )\n", stderr);
+	fputs("   or: git stash [push]\n", stderr);
+	fputs("\n", stderr);
+
+	va_start(ap, fmt);
+	vfprintf(stderr, fmt, ap);
+	va_end(ap);
+
+	exit(1);
+}
+
+static void parse_subcommand(struct opts *opts, int argc, char *argv[])
+{
+	char *arg = (argc < 2) ? "push" : argv[1];
+	enum subcmd cmd;
+
+	if (!strcmp(arg, "apply")) {
+		cmd = SUBCMD_APPLY;
+	} else if (!strcmp(arg, "list")) {
+		cmd = SUBCMD_LIST;
+	} else if (!strcmp(arg, "pop")) {
+		cmd = SUBCMD_POP;
+	} else if (!strcmp(arg, "push")) {
+		cmd = SUBCMD_PUSH;
+	} else {
+		usage("invalid command %s", arg);
+		return;
+	}
+
+	opts->cmd = cmd;
+	opts->argc = (argc < 2) ? argc - 1 : argc - 2;
+	opts->argv = argv;
+}
+
+static int cmd_apply(git_repository *repo, struct opts *opts)
+{
+	if (opts->argc)
+		usage("apply does not accept any parameters");
+
+	check_lg2(git_stash_apply(repo, 0, NULL),
+		  "Unable to apply stash", NULL);
+
+	return 0;
+}
+
+static int list_stash_cb(size_t index, const char *message,
+			 const git_oid *stash_id, void *payload)
+{
+	UNUSED(stash_id);
+	UNUSED(payload);
+	printf("stash@{%"PRIuZ"}: %s\n", index, message);
+	return 0;
+}
+
+static int cmd_list(git_repository *repo, struct opts *opts)
+{
+	if (opts->argc)
+		usage("list does not accept any parameters");
+
+	check_lg2(git_stash_foreach(repo, list_stash_cb, NULL),
+		  "Unable to list stashes", NULL);
+
+	return 0;
+}
+
+static int cmd_push(git_repository *repo, struct opts *opts)
+{
+	git_signature *signature;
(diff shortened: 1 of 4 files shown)


pks-t:
Fixes #3910

src/stash.c
```diff
@@ -398,28 +398,23 @@ static int commit_worktree(
...
 
-	if ((error = git_commit_tree(&i_tree, i_commit)) < 0)
-		goto cleanup;
-
-	if ((error = git_index_new(&i_index)) < 0 ||
-		(error = git_repository__cvar(&ignorecase, repo, GIT_CVAR_IGNORECASE)) < 0)
-		goto cleanup;
+	if ((error = git_repository_index(&r_index, repo) < 0) ||
+	    (error = git_index_new(&i_index)) < 0 ||
+	    (error = git_index__fill(i_index, &r_index->entries) < 0) ||
```
ethomson: Hmm, I haven't done a deep analysis here, but is there not a behavior difference?  Previously, we'd create an index from the commit tree, now we're using the existing index.  So if there are staged changes, those will now be included in `i_index` while they were not before.

(Apologies if there's already a guard here to avoid that).

I think that we have a stat-preserving way to make these match if that's needed.  I'm reasonably sure that `git_index_read_tree` will actually preserve the stat cache.  So adding that back after filling `i_index` with stat information should keep `i_index` identical to `i_tree` with the correct stat information (for matching entries).

But maybe I'm off base here and this function only gets called when `i_index` == HEAD, in which case, there's no reason to worry about that.  Indeed, I would expect unit tests to catch this case, but I wanted to mention it while it was top of mind for me.

I can do a more detailed 👀 soon.

pks-t:
On Fri, Jun 14, 2019 at 06:07:36AM -0700, Edward Thomson wrote:
>  	parents[0] = b_commit;
>  	parents[1] = i_commit;
>  	parents[2] = u_commit;
>  
> -	if ((error = git_commit_tree(&i_tree, i_commit)) < 0)
> -		goto cleanup;
> -
> -	if ((error = git_index_new(&i_index)) < 0 ||
> -		(error = git_repository__cvar(&ignorecase, repo, GIT_CVAR_IGNORECASE)) < 0)
> -		goto cleanup;
> +	if ((error = git_repository_index(&r_index, repo) < 0) ||
> +	    (error = git_index_new(&i_index)) < 0 ||
> +	    (error = git_index__fill(i_index, &r_index->entries) < 0) ||
> 
> Hmm, I haven't done a deep analysis here, but is there not a
> behavior difference?  Previously, we'd create an index from the
> commit tree, now we're using the existing index.  So if there
> are staged changes, those will now be included in `i_index`
> while they were not before.
> 
> (Apologies if there's already a guard here to avoid that).
> 
> I think that we have a stat-preserving way to make these match
> if that's needed.  I'm reasonably sure that
> `git_index_read_tree` will actually preserve the stat cache.
> So adding that back after filling `i_index` with stat
> information should keep `i_index` identical to `i_tree` with
> the correct stat information (for matching entries).
> 
> But maybe I'm off base here and this function only gets called
> when `i_index` == HEAD, in which case, there's no reason to
> worry about that.  Indeed, I would expect unit tests to catch
> this case, but I wanted to mention it while it was top of mind
> for me.
> 
> I can do a more detailed 👀 soon.

I _think_ we're fine here, even though I initially wondered about
the exact same thing. Note: the `i_commit` passed in to
`commit_worktree` is actually created from the current repo's
index. Thus by using `i_tree` from `i_index`, we were essentially
already using the contents from the index. Now we just avoid
going via the tree, but directly copy the index including all its
stats caches.

pks-t:
Ping :)

examples/stash.c
```diff
@@ -0,0 +1,157 @@
...
+
+	check_lg2(git_stash_pop(repo, 0, NULL),
+		  "Unable to pop stash", NULL);
+
+	printf("Dropped refs/stash@{0}\n");
+
+	return 0;
+}
+
+int lg2_stash(git_repository *repo, int argc, char *argv[])
```
tiennou: Nothing standing out here codewise: I just want to point out that Docurium seemingly chokes on examples with no docblocks (add.c has that problem). But it could be something else — I haven't looked at the example processing system *at all*.

pks-t: "add.c" does have docblocks, though, so I'd expect it to be something else. Does this break Docurium in a way that needs to be fixed before we merge this or can we just proceed? I'm obviously asking out of sheer lazyness 🙄

tiennou: No worries, it's just a reminder (mostly to me), it's a Docurium bug in any case, and — on the offchance it is docblock-related — it doesn't cause anything else than a blank example page.

IIRC most examples have a "main" docblock at the top that describes "usage", that'd be fine by me. Or this can be merged, and I'll check when I get near Docurium again.

src/stash.c
implausible: Any chance you're free to double check this soon :smile: @ethomson.

pks-t:
Too many branches, deleted the wrong one by accident 🙄

pks-t:
Rebased to fix conflicts

ethomson:
OK, I've read through our tests and they seem very thorough.  And I've run this through manual testing, with a variety of changes in the index and working directory, and this matches the behavior of git in all of them.

Thanks for doing this @pks-t !

gjcampbell:
This fix was awesome. Can --untracked-changes be optimized as well?

pks-t:
> This fix was awesome. Can --untracked-changes be optimized as well?
They probably could be optimized, but that'd require a proper test case that shows that we perform much worse than e.g. the official git implementation. Please feel free to create an issue with a reproducer, this would make it much more likely for anybody to tackle the issue.

PR #5114, opened 2019-06-14
Removal of `p_fallocate`

pks-t:
The `p_fallocate` platform is currently in use in our tests,
only, but it proved to be quite burdensome to get it implemented
in a cross-platform way. The only "real" user is the test
object::tree::read::largefile, where it's used to allocate a
large file in the filesystem only to commit it to the repo and
read its object back again. We can simplify this quite a bit by
just using an in-memory buffer of 4GB. Sure, this cannot be used
on platforms with low resources. But creating 4GB files is not
any better, and we already skip the test if the environment
variable "GITTEST_INVASIVE_FS_SIZE" is not set. So we're arguably
not worse off than before.

Supersedes #5087

### DIFF
--- a/tests/object/tree/read.c
+++ b/tests/object/tree/read.c
@@ -79,46 +79,36 @@ void test_object_tree_read__two(void)
 
 void test_object_tree_read__largefile(void)
 {
-	git_reference *ref;
+	const git_tree_entry *entry;
+	git_index_entry ie;
 	git_commit *commit;
+	git_object *object;
+	git_index *index;
 	git_tree *tree;
 	git_oid oid;
-	const git_tree_entry *entry;
-	git_object *object;
-	git_buf file = GIT_BUF_INIT;
-	int fd;
-	git_index *idx;
+	char *buf;
 
 	if (!cl_is_env_set("GITTEST_INVASIVE_FS_SIZE"))
 		cl_skip();
 
-	cl_git_pass(git_reference_lookup(&ref, g_repo, "refs/heads/master"));
-	cl_git_pass(git_repository_index(&idx, g_repo));
+	cl_assert(buf = git__calloc(1, BIGFILE_SIZE));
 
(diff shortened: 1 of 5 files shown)


tests/object/tree/read.c
```diff
@@ -79,46 +79,34 @@ void test_object_tree_read__two(void)
 
 void test_object_tree_read__largefile(void)
 {
-	git_reference *ref;
+	const git_tree_entry *entry;
+	git_index_entry ie = { 0 };
```
tiennou: CI [NAKed](https://dev.azure.com/libgit2/libgit2/_build/results?buildId=2095&view=logs&jobId=7351e2b3-4d1a-54ac-0ff4-44bd85f4818d&taskId=49622b32-88bd-5294-a7a5-08280f44966c&lineStart=602&lineEnd=603&colStart=1&colEnd=1) this.

ethomson:
> I 🙇 to these changes. Thanks for taking care of it @pks-t !

Same, thanks @tiennou and @pks-t to make sure that we're building carefully!

pks-t:
On Fri, Jun 14, 2019 at 05:51:27AM -0700, Etienne Samson wrote:
> CI [NAKed](https://dev.azure.com/libgit2/libgit2/_build/results?buildId=2095&view=logs&jobId=7351e2b3-4d1a-54ac-0ff4-44bd85f4818d&taskId=49622b32-88bd-5294-a7a5-08280f44966c&lineStart=602&lineEnd=603&colStart=1&colEnd=1) this.

I hit this issue so many times already. I always wonder why it is
that CI catches those things while my own system doesn't. I mean
we're both using GCC -- are they setting up extra CFLAGS?

Anyway, should be fixed now, thanks for catching this.

ethomson:
> I hit this issue so many times already. I always wonder why it is that CI catches those things while my own system doesn't. I mean we're both using GCC -- are they setting up extra CFLAGS?

Huh.  Maybe?  More likely same CFLAGS but different compiler versions?

PR #5116, opened 2019-06-15
Commit creation API streamlining

tiennou:
This is a rebase of #3075, with the attempt at a more general low-level API I've envisioned for #4913 ([comment](https://github.com/libgit2/libgit2/pull/4913#discussion_r247658573)).

It's slightly different in that I (should) have taken care of not changing anything with old API, and only implement new things on top on newly added functions.

And then I'm adding low-level methods to manipulate a `git_commit_desc`, which serves to keep track of "the parts of a commit", via the following way.
- prepare a buffer given a `git_commit_desc`
- write a commit object to a repository, given a `git_commit_desc`, and some options.

This is still very much in flux, as I'm still trying to see how much things I can delegate to the low-level helpers (stuff like `fromstate` becomes an option when object-writing). Ideally in the end it should be possible to reimplement all the new "helpers" from #3075 using those two, but I wanted the "cleaner API" first. Thus, any design insights would be welcome.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -22,6 +22,23 @@ v0.28 + 1
 * libgit2 can now correctly cope with URLs where the host contains a colon
   but a port is not specified.  (eg `http://example.com:/repo.git`).
 
+### API additions
+
+* `git_commit_create_on_ex()` takes on the role which
+  `git_commit_create()` had when provided a reference name to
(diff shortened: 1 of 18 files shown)


ethomson:
We've made some nice changes in the commit API space; certainly I'm willing to do more, but this has become very outdated. 🙏

PR #5117, opened 2019-06-15
Change API instances of `fromnoun` to `from_noun` (with an underscore)

ethomson:
We were inconsistent in our API naming, sometimes using `git_object_fromnoun` and sometimes using `git_object_from_noun`.

The majority of functions are named `from_something` (with an underscore) instead of `fromsomething`.  Update the functions for consistency with the rest of the library.

### DIFF
--- a/examples/blame.c
+++ b/examples/blame.c
@@ -72,7 +72,7 @@ int lg2_blame(git_repository *repo, int argc, char *argv[])
 	 * Get the raw data inside the blob for output. We use the
 	 * `commitish:path/to/file.txt` format to find it.
 	 */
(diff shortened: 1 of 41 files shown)


ethomson:
Note that I didn't touch the `git_oid` functions.  I think that these are relatively simple changes that shouldn't be very contentious.  I think that the `git_oid` functions may require more thought and discussion.

pks-t:
> Note that I didn't touch the git_oid functions.

At least not the `git_oid_from<foo>` ones, you _did_ touch the `git_oid_is_zero` one. :)

PR #5118, opened 2019-06-15
object: rename git_object__size to git_object_size

ethomson:
We don't use double-underscores in the public API.

### DIFF
--- a/include/git2/deprecated.h
+++ b/include/git2/deprecated.h
@@ -235,6 +235,8 @@ GIT_EXTERN(void) giterr_set_oom(void);
 #define GIT_OBJ_OFS_DELTA GIT_OBJECT_OFS_DELTA
 #define GIT_OBJ_REF_DELTA GIT_OBJECT_REF_DELTA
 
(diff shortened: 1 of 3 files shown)


carlosmn:
I guess I'm a bit late to this, but should this even be public API? Could we rename it to make it clearer that it's not the size of the object itself but of our structure to hold the data?

I went in today expecting `git_object_size` to return the size of the cached object i.e. the uncompressed size of the underlying object.

ethomson:
> I guess I'm a bit late to this, but should this even be public API?

🤷‍♂ 

Yeah, looking at it, I don't know why it's public.

ethomson:
Now would be a good time to fix this since we haven't had a release that does anything here.  If you wanted to roll the rename back and just deprecate `git_object__size` entirely, then that would be fine with me.

carlosmn:
I *can* be useful if you're really trying to account for memory usage from objects accurately (since this is on top of the actual inflated object). But even then, you don't necessarily know how many objects are cached anyway.

So yeah, I think removing it is probably the best option at this point.

PR #5119, opened 2019-06-15
attr: rename constants and macros for consistency

ethomson:
Our enumeration values are not generally suffixed with `T`.  Further,
our enumeration names are generally more descriptive.


pks-t:
Got conflicts with "deprecated.h" because of #5117, though

ethomson:
Fixed up them conflicts, :shipit:

PR #5123, opened 2019-06-16
Move `git_off_t` to `git_object_size_t`

ethomson:
`git_off_t` is problematic: first, it's used to represent the size of
a blob, but that's represented in git as an `unsigned long`, meaning
that we've lost a bit (on 64-bit machines).  Second, `git_off_t` is a
poor name, suggesting that it's an offset, not a size (which makes
sense, given that it's a signed type).

Migrate `git_off_t` to `git_object_size_t`, an unsigned 64-bit type.

### DIFF
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -258,12 +258,12 @@ typedef enum {
  * abbreviated to something reasonable, like 7 characters.
  */
 typedef struct {
-	git_oid     id;
-	const char *path;
-	git_off_t   size;
-	uint32_t    flags;
-	uint16_t    mode;
-	uint16_t    id_abbrev;
+	git_oid            id;
+	const char        *path;
+	git_object_size_t  size;
+	uint32_t           flags;
+	uint16_t           mode;
+	uint16_t           id_abbrev;
 } git_diff_file;
 
 /**
--- a/src/diff_file.c
+++ b/src/diff_file.c
@@ -330,8 +330,10 @@ static int diff_file_content_load_workdir_file(
 	if (fd < 0)
 		return fd;
 
-	if (!fc->file->size &&
-		!(fc->file->size = git_futils_filesize(fd)))
+	if (!fc->file->size)
+	    error = git_futils_filesize(&fc->file->size, fd);
+
+	if (error < 0 || !fc->file->size)
 		goto cleanup;
 
 	if ((diff_opts->flags & GIT_DIFF_SHOW_BINARY) == 0 &&
--- a/src/diff_file.c
+++ b/src/diff_file.c
@@ -62,7 +62,7 @@ static int diff_file_content_init_common(
 	git_diff_driver_update_options(&fc->opts_flags, fc->driver);
 
 	/* make sure file is conceivable mmap-able */
-	if ((git_off_t)((size_t)fc->file->size) != fc->file->size)
+	if ((size_t)fc->file->size != fc->file->size)
 		fc->file->flags |= GIT_DIFF_FLAG_BINARY;
 	/* check if user is forcing text diff the file */
 	else if (fc->opts_flags & GIT_DIFF_FORCE_TEXT) {
--- a/src/futils.c
+++ b/src/futils.c
@@ -112,7 +112,7 @@ int git_futils_truncate(const char *path, int mode)
 	return 0;
 }
 
-git_off_t git_futils_filesize(git_file fd)
+int git_futils_filesize(uint64_t *out, git_file fd)
 {
 	struct stat sb;
 
@@ -121,7 +121,13 @@ git_off_t git_futils_filesize(git_file fd)
 		return -1;
 	}
 
-	return sb.st_size;
+	if (sb.st_size < 0) {
+		git_error_set(GIT_ERROR_INVALID, "invalid file size");
+		return -1;
+	}
+
+	*out = sb.st_size;
+	return 0;
 }
 
 mode_t git_futils_canonical_mode(mode_t raw_mode)
@@ -309,16 +315,14 @@ int git_futils_mmap_ro(git_map *out, git_file fd, git_off_t begin, size_t len)
 int git_futils_mmap_ro_file(git_map *out, const char *path)
 {
 	git_file fd = git_futils_open_ro(path);
-	git_off_t len;
+	uint64_t len;
 	int result;
 
 	if (fd < 0)
 		return fd;
 
-	if ((len = git_futils_filesize(fd)) < 0) {
-		result = -1;
+	if ((result = git_futils_filesize(&len, fd)) < 0)
 		goto out;
-	}
 
 	if (!git__is_sizet(len)) {
 		git_error_set(GIT_ERROR_OS, "file `%s` too large to mmap", path);
--- a/src/futils.c
+++ b/src/futils.c
@@ -1110,7 +1110,7 @@ int git_futils_filestamp_check(
 #if defined(GIT_USE_NSEC)
 		stamp->mtime.tv_nsec == st.st_mtime_nsec &&
 #endif
-		stamp->size  == (git_off_t)st.st_size   &&
+		stamp->size  == (uint64_t)st.st_size   &&
 		stamp->ino   == (unsigned int)st.st_ino)
 		return 0;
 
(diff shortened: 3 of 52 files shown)


ethomson:
Aha, we do actually treat `git_off_t` as an offset in the packfile handling (and as a length, of course).  I'll have to step through here and 👀 how we're dealing with this and tighten it up.

ethomson:
Rebased to fix conflicts.

ethomson:
Rebased.

ethomson:
@pks-t any thoughts here?  I'd love to think about a release 🔜 and include this.

pks-t:
On Fri, Aug 16, 2019 at 02:41:53AM -0700, Edward Thomson wrote:
> @pks-t any thoughts here?  I'd love to think about a release 🔜
> and include this.

I hope to be able to find enought time to pick up some loose ends
in libgit2 this Friday. And I fully agree that we should start
thinking about a release soon, it's been quite some time.

include/git2/diff.h
```diff
@@ -603,7 +603,7 @@ typedef struct {
 	int    new_lineno;   /**< Line number in new file or -1 for deleted line */
 	int    num_lines;    /**< Number of newline characters in content */
 	size_t content_len;  /**< Number of bytes of data */
-	git_off_t content_offset; /**< Offset in the original file to the content */
+	git_object_size_t content_offset; /**< Offset in the original file to the content */
```
pks-t: Does it really make sense to use `git_object_size_t` here? Diffs aren't necessary referring to any objects at all but, so it seems a bit dirty to me to mix these up.

src/diff_file.c
```diff
@@ -46,7 +46,7 @@ static int diff_file_content_init_common(
 {
 	fc->opts_flags = opts ? opts->flags : GIT_DIFF_NORMAL;
 
-	if (opts && opts->max_size >= 0)
+	if (opts && (int64_t)opts->max_size >= 0)
```
pks-t: Mh. This comparison doesn't make sense anymore, as `max_size` is now unsigned. If it becomes negative due to casting then I'd consider this a bug.

include/git2/diff.h
```diff
@@ -414,10 +414,10 @@ typedef struct {
 
 	/**
 	 * A size (in bytes) above which a blob will be marked as binary
-	 * automatically; pass a negative value to disable.
+	 * automatically; pass `GIT_OBJECT_SIZE_MAX` to disable.
 	 * Defaults to 512MB.
 	 */
-	git_off_t   max_size;
+	git_object_size_t max_size;
```
pks-t: Hm, this changes our API, doesn't it?

src/diff_xdiff.c
```diff
@@ -160,7 +160,7 @@ static int git_xdiff_cb(void *priv, mmbuffer_t *bufs, int len)
 		else if (line.origin == GIT_DIFF_LINE_DELETION)
 			line.content_offset = bufs[1].ptr - info->xd_old_data.ptr;
 		else
-			line.content_offset = -1;
+			line.content_offset = GIT_OBJECT_SIZE_MAX;
```
pks-t: This changes semantics too, as we pass these to the callers

src/futils.c
```diff
@@ -112,7 +112,7 @@ int git_futils_truncate(const char *path, int mode)
 	return 0;
 }
 
-git_off_t git_futils_filesize(git_file fd)
+int git_futils_filesize(git_object_size_t *out, git_file fd)
```
pks-t: Not sure how I feel about these. `off_t` directly maps to the mmap functions of the OS, using `git_object_size_t` feels weird to me

include/git2/diff.h
ethomson: Hmm, perhaps not.  The rare good use of `git_off_t`?

src/diff_file.c
ethomson: `max_size` is still signed for backcompat.

ethomson: Yeah, ignore my last comment.  You're right.

include/git2/diff.h
ethomson: In practice, if callers ever provided a negative value, it was caught in `diff_file_content_init_common`, and ignored.  
 https://github.com/libgit2/libgit2/pull/5123#discussion-diff-317085663R49 was doing the same.  So as-is we are API compatible, but if we remove that odd-looking check, then we are in fact changing the API.

src/diff_xdiff.c
ethomson: Sorry, can you give me more information about what you mean?

src/futils.c
ethomson: Well, `mmap` does take an `off_t` but only for the offset into the file.  That doesn't correspond to the length of the file or memory-mapped segment.

I definitely don't think that we should use a signed type for file length, it simply can't be negative.  We wouldn't want to use `size_t` here, since on a 32-bit system, we want to be able to get the size of a file larger than 4 GB.

We could call this the more direct `uint64_t` here.

src/diff_xdiff.c
pks-t: We're populating the `git_diff_line` struct here, which we directly pass to the user via the `data_cb` callback. So previously callers would've seen `content_offset = -1`, now he sees `content_offset = UINT64_MAX` if a line was unchanged. I could imagine people having checks for that

src/futils.c
pks-t: Hum. Some platforms handle negative offsets in `mmap` just fine and treat it as valid. It's questionable though whether we're going to ever need that, especially considering that we'd have to fight with cross-platform behaviour here. So :+1: for `uint64_t`.

src/diff_file.c
ethomson: Fixed.

src/futils.c
ethomson: Updated for `uint64_t`

src/diff_xdiff.c
ethomson: Moved this back to a signed int to support this.

include/git2/diff.h
ethomson: Moved this back to a signed int for sanity.

include/git2/diff.h
ethomson: Fixed.

src/futils.h
```diff
@@ -330,7 +330,7 @@ extern int git_futils_fake_symlink(const char *new, const char *old);
  */
 typedef struct {
 	struct timespec mtime;
-	git_off_t  size;
+	git_object_size_t size;
```
pks-t: Some file APIs now use `uint64_t`, some use `git_object_size_t`. Shouldn't this one here use `uint64_t`, as well?

src/posix.c
```diff
@@ -235,7 +235,7 @@ int git__mmap_alignment(size_t *alignment)
 }
 
 
-int p_mmap(git_map *out, size_t len, int prot, int flags, int fd, git_off_t offset)
+int p_mmap(git_map *out, size_t len, int prot, int flags, int fd, off64_t offset)
```
pks-t: Quick question, no need to change: why did you opt to convert to `off64_t` instead of just typedeffing `git_off_t` to `off64_t`?

src/win32/map.c
```diff
@@ -99,7 +99,7 @@ int p_mmap(git_map *out, size_t len, int prot, int flags, int fd, git_off_t offs
 		return -1;
 	}
 
-	assert(sizeof(git_off_t) == 8);
+	assert(sizeof(off64_t) == 8);
```
pks-t: Does this make sense? I mean it's very unlikely that `off64_t` is ever going to be something else than 8 bytes

src/futils.h
ethomson: Yes, this probably makes sense.

src/win32/map.c
ethomson: 😬 Yes, that's very true.  Removed entirely.

src/posix.c
ethomson: I thought that it was clearer for internal uses to use something Very Obviously 64 bit, and something that is a Standard Type.  `git_off_t` feels like the sort of thing that is perhaps useful for our external API, but less useful for our internal.

Your comment did make me rethink a few things, so I actually kept `git_off_t` in the external type places that it was used.  I think this is a pretty reasonable compromise.

src/diff_file.c
```diff
@@ -330,8 +330,10 @@ static int diff_file_content_load_workdir_file(
 	if (fd < 0)
 		return fd;
 
-	if (!fc->file->size &&
-		!(fc->file->size = git_futils_filesize(fd)))
+	if (!fc->file->size)
+	    error = git_futils_filesize(&fc->file->size, fd);
+
+	if (error < 0 || !fc->file->size)
```
pks-t: This does look a bit weird. Seems like we do not allow loading empty files from disk here. Your change only makes code a lot clearer here, so I'm fine with this as is.

PR #5124, opened 2019-06-17
cmake: default NTLM client to off if no HTTPS support

pks-t:
If building libgit2 with `-DUSE_HTTPS=NO`, then CMake will
generate an error complaining that there's no usable HTTPS
backend for NTLM. In fact, it doesn't make sense to support NTLM
when we don't support HTTPS. So let's should just have
NTLM default to OFF when HTTPS is disabled to make life easier
and to fix our OSSFuzz builds failing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,11 @@ OPTION(DEPRECATE_HARD			"Do not include deprecated functions in the library"	OFF
    SET(REGEX_BACKEND			"" CACHE STRING "Regular expression implementation. One of regcomp_l, pcre2, pcre, regcomp, or builtin.")
 
 IF (UNIX)
-	OPTION(USE_NTLMCLIENT		"Enable NTLM support on Unix."				ON )
+	IF (NOT USE_HTTPS)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -66,7 +66,11 @@ OPTION(DEPRECATE_HARD			"Do not include deprecated functions in the library"	OFF
    SET(REGEX_BACKEND			"" CACHE STRING "Regular expression implementation. One of regcomp_l, pcre2, pcre, regcomp, or builtin.")
 
 IF (UNIX)
-	OPTION(USE_NTLMCLIENT		"Enable NTLM support on Unix."				ON )
+	IF (NOT USE_HTTPS)
+	    OPTION(USE_NTLMCLIENT		"Enable NTLM support on Unix."				OFF )
+	ELSE()
+	    OPTION(USE_NTLMCLIENT		"Enable NTLM support on Unix."				ON )
+	ENDIF()
```
pks-t: It's quite verbose, but as far as I know there's no other way to do it due to `USE_HTTPS` accepting not only ON/OFF, but specific backends, too.

ethomson:
Makes sense.

PR #5125, opened 2019-06-18
Compare buffers in diff example

albfan:
Adding compare buffers feature in diff example

This tries to mimic `git diff --no-index`

```
echo hello > file1
echo bye > file2
git diff --no-index file1 file2
diff --git 1/file1 2/file2
index 5c1b14949..b023018ca 100644
--- 1/file1
+++ 2/file2
@@ -1 +1 @@
-hello
+bye
```

Right now this patch does:

```
$ ./diff --files --text --color
buf: diff --git a/file1 b/file2
index ce01362..b023018 100644
--- a/file1
+++ b/file2
@@ -1 +1 @@
-hello
+bye
diff --git a/file1 b/file2
index ce01362..b023018 100644
--- a/file1
+++ b/file2
@@ -1 +1 @@
-hello
+bye
```

Actual problems:

-  Read real files
- Do not detect buffers as binary files
- Use `--no-index` as parameter
- Check two files are passed and they exists

### DIFF
--- a/examples/common.c
+++ b/examples/common.c
@@ -14,6 +14,14 @@
 
 #include <assert.h>
 #include <stdio.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#ifdef _WIN32
+# include <io.h>
+#else
+# include <fcntl.h>
+# include <unistd.h>
+#endif
 #include <string.h>
 #include <errno.h>
 
@@ -391,3 +399,37 @@ out:
 	free(pubkey);
 	return error;
 }
+
+char *read_file(const char *path)
+{
+	ssize_t total = 0;
+	char *buf = NULL;
+	struct stat st;
+	int fd = -1;
+
+	if ((fd = open(path, O_RDONLY)) < 0 || fstat(fd, &st) < 0)
+		goto out;
+
+	if ((buf = malloc(st.st_size + 1)) == NULL)
+		goto out;
+
+	while (total < st.st_size) {
+		ssize_t bytes = read(fd, buf + total, st.st_size - total);
+		if (bytes <= 0) {
+			if (errno == EAGAIN || errno == EINTR)
+				 continue;
+			free(buf);
+			buf = NULL;
+			goto out;
+		}
+		total += bytes;
+	}
+
+	buf[total] = '\0';
+
+out:
+	if (fd >= 0)
+		close(fd);
+	return buf;
+}
+
--- a/examples/common.h
+++ b/examples/common.h
@@ -64,6 +64,14 @@ extern int lg2_tag(git_repository *repo, int argc, char **argv);
 extern void check_lg2(int error, const char *message, const char *extra);
 
 /**
+ * Read a file into a buffer
+ *
+ * @param path The path to the file that shall be read
+ * @return NUL-terminated buffer if the file was successfully read, NULL-pointer otherwise
+ */
+extern char *read_file(const char *path);
+
+/**
  * Exit the program, printing error to stderr
  */
 extern void fatal(const char *message, const char *extra);
@@ -89,7 +97,7 @@ struct args_info {
 /**
  * Check current `args` entry against `opt` string.  If it matches
  * exactly, take the next arg as a string; if it matches as a prefix with
- * an equal sign, take the remainder as a string; if value not supplied, 
+ * an equal sign, take the remainder as a string; if value not supplied,
  * default value `def` will be given. otherwise return 0.
  */
 extern int optional_str_arg(
--- a/examples/common.h
+++ b/examples/common.h
@@ -12,10 +12,25 @@
  * <http://creativecommons.org/publicdomain/zero/1.0/>.
  */
 
+#include <sys/types.h>
+#include <sys/stat.h>
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
 #include <git2.h>
+#include <fcntl.h>
+
+#ifdef _WIN32
+# include <io.h>
+# include <Windows.h>
+# define open _open
+# define read _read
+# define close _close
+# define ssize_t int
+# define sleep(a) Sleep(a * 1000)
+#else
+# include <unistd.h>
+#endif
 
 #ifndef PRIuZ
 /* Define the printf format specifer to use for size_t output */
--- a/examples/diff.c
+++ b/examples/diff.c
@@ -52,6 +52,7 @@ struct opts {
 	git_diff_options diffopts;
 	git_diff_find_options findopts;
 	int color;
+	int no_index;
 	int cache;
 	int output;
 	git_diff_format_t format;
@@ -66,14 +67,16 @@ static void parse_opts(struct opts *o, int argc, char *argv[]);
 static int color_printer(
 	const git_diff_delta*, const git_diff_hunk*, const git_diff_line*, void*);
 static void diff_print_stats(git_diff *diff, struct opts *o);
+static void compute_diff_no_index(git_diff **diff, struct opts *o);
 
 int lg2_diff(git_repository *repo, int argc, char *argv[])
 {
 	git_tree *t1 = NULL, *t2 = NULL;
 	git_diff *diff;
+
 	struct opts o = {
 		GIT_DIFF_OPTIONS_INIT, GIT_DIFF_FIND_OPTIONS_INIT,
-		-1, 0, 0, GIT_DIFF_FORMAT_PATCH, NULL, NULL, "."
+		-1, -1, 0, 0, GIT_DIFF_FORMAT_PATCH, NULL, NULL, "."
 	};
 
 	parse_opts(&o, argc, argv);
@@ -86,49 +89,54 @@ int lg2_diff(git_repository *repo, int argc, char *argv[])
 	 *  * &lt;sha1&gt;
 	 *  * --cached
 	 *  * --nocache (don't use index data in diff at all)
+	 *  * --no-index &lt;file1&gt; &lt;file2&gt;
 	 *  * nothing
 	 *
 	 * Currently ranged arguments like &lt;sha1&gt;..&lt;sha2&gt; and &lt;sha1&gt;...&lt;sha2&gt;
 	 * are not supported in this example
 	 */
 
-	if (o.treeish1)
-		treeish_to_tree(&t1, repo, o.treeish1);
-	if (o.treeish2)
-		treeish_to_tree(&t2, repo, o.treeish2);
+	if (o.no_index >= 0) {
+		compute_diff_no_index(&diff, &o);
+	} else {
+		if (o.treeish1)
+			treeish_to_tree(&t1, repo, o.treeish1);
+		if (o.treeish2)
+			treeish_to_tree(&t2, repo, o.treeish2);
 
-	if (t1 && t2)
-		check_lg2(
-			git_diff_tree_to_tree(&diff, repo, t1, t2, &o.diffopts),
-			"diff trees", NULL);
-	else if (o.cache != CACHE_NORMAL) {
-		if (!t1)
-			treeish_to_tree(&t1, repo, "HEAD");
-
-		if (o.cache == CACHE_NONE)
+		if (t1 && t2)
 			check_lg2(
-				git_diff_tree_to_workdir(&diff, repo, t1, &o.diffopts),
+				git_diff_tree_to_tree(&diff, repo, t1, t2, &o.diffopts),
+				"diff trees", NULL);
+		else if (o.cache != CACHE_NORMAL) {
+			if (!t1)
+				treeish_to_tree(&t1, repo, "HEAD");
+
+			if (o.cache == CACHE_NONE)
+				check_lg2(
+					git_diff_tree_to_workdir(&diff, repo, t1, &o.diffopts),
+					"diff tree to working directory", NULL);
+			else
+				check_lg2(
+					git_diff_tree_to_index(&diff, repo, t1, NULL, &o.diffopts),
+					"diff tree to index", NULL);
+		}
+		else if (t1)
+			check_lg2(
+				git_diff_tree_to_workdir_with_index(&diff, repo, t1, &o.diffopts),
 				"diff tree to working directory", NULL);
 		else
 			check_lg2(
-				git_diff_tree_to_index(&diff, repo, t1, NULL, &o.diffopts),
-				"diff tree to index", NULL);
+				git_diff_index_to_workdir(&diff, repo, NULL, &o.diffopts),
+				"diff index to working directory", NULL);
+
+		/** Apply rename and copy detection if requested. */
+
+		if ((o.findopts.flags & GIT_DIFF_FIND_ALL) != 0)
+			check_lg2(
+				git_diff_find_similar(diff, &o.findopts),
+				"finding renames and copies", NULL);
 	}
-	else if (t1)
-		check_lg2(
-			git_diff_tree_to_workdir_with_index(&diff, repo, t1, &o.diffopts),
-			"diff tree to working directory", NULL);
-	else
-		check_lg2(
-			git_diff_index_to_workdir(&diff, repo, NULL, &o.diffopts),
-			"diff index to working directory", NULL);
-
-	/** Apply rename and copy detection if requested. */
-
-	if ((o.findopts.flags & GIT_DIFF_FIND_ALL) != 0)
-		check_lg2(
-			git_diff_find_similar(diff, &o.findopts),
-			"finding renames and copies", NULL);
 
 	/** Generate simple output using libgit2 display helper. */
 
@@ -156,6 +164,38 @@ int lg2_diff(git_repository *repo, int argc, char *argv[])
 	git_tree_free(t2);
 
 	return 0;
+}
+
+static void compute_diff_no_index(git_diff **diff, struct opts *o) {
+	git_patch *patch = NULL;
+	char *file1_str = NULL;
+	char *file2_str = NULL;
+	git_buf buf = {0};
+
+	if (!o->treeish1 || !o->treeish2) {
+		usage("two files should be provided as arguments", NULL);
+	}
+	file1_str = read_file(o->treeish1);
+	if (file1_str == NULL) {
+		usage("file cannot be read", o->treeish1);
+	}
+	file2_str = read_file(o->treeish2);
+	if (file2_str == NULL) {
+		usage("file cannot be read", o->treeish2);
+	}
+	check_lg2(
+		git_patch_from_buffers(&patch, file1_str, strlen(file1_str), o->treeish1, file2_str, strlen(file2_str), o->treeish2, &o->diffopts),
+		"patch buffers", NULL);
+	check_lg2(
+		git_patch_to_buf(&buf, patch),
+		"patch to buf", NULL);
+	check_lg2(
+		git_diff_from_buffer(diff, buf.ptr, buf.size),
+		"diff from patch", NULL);
+	git_patch_free(patch);
+	git_buf_dispose(&buf);
+	free(file1_str);
+	free(file2_str);
 }
 
 static void usage(const char *message, const char *arg)
@@ -223,9 +263,10 @@ static void parse_opts(struct opts *o, int argc, char *argv[])
 			o->output |= OUTPUT_DIFF;
 			o->format = GIT_DIFF_FORMAT_PATCH;
 		}
-		else if (!strcmp(a, "--cached"))
+		else if (!strcmp(a, "--cached")) {
 			o->cache = CACHE_ONLY;
-		else if (!strcmp(a, "--nocache"))
+			if (o->no_index >= 0) usage("--cached and --no-index are incompatible", NULL);
+		} else if (!strcmp(a, "--nocache"))
 			o->cache = CACHE_NONE;
 		else if (!strcmp(a, "--name-only") || !strcmp(a, "--format=name"))
 			o->format = GIT_DIFF_FORMAT_NAME_ONLY;
@@ -238,7 +279,10 @@ static void parse_opts(struct opts *o, int argc, char *argv[])
 			o->format = GIT_DIFF_FORMAT_RAW;
 			o->diffopts.id_abbrev = 40;
 		}
-		else if (!strcmp(a, "--color"))
+		else if (!strcmp(a, "--no-index")) {
+			o->no_index = 0;
+			if (o->cache == CACHE_ONLY) usage("--cached and --no-index are incompatible", NULL);
+		} else if (!strcmp(a, "--color"))
 			o->color = 0;
 		else if (!strcmp(a, "--no-color"))
 			o->color = -1;
--- a/examples/for-each-ref.c
+++ b/examples/for-each-ref.c
@@ -1,5 +1,4 @@
 #include <git2.h>
-#include <stdio.h>
 #include "common.h"
 
 static int show_ref(git_reference *ref, void *data)
--- a/examples/general.c
+++ b/examples/general.c
@@ -36,6 +36,8 @@
  * [pg]: https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain
  */
 
+#include "common.h"
+
 /**
  * ### Includes
  *
@@ -43,9 +45,7 @@
  * that you need.  It should be the only thing you need to include in order
  * to compile properly and get all the libgit2 API.
  */
-#include <git2.h>
-#include <stdio.h>
-#include <string.h>
+#include "git2.h"
 
 static void oid_parsing(git_oid *out);
 static void object_database(git_repository *repo, git_oid *oid);
--- a/examples/index-pack.c
+++ b/examples/index-pack.c
@@ -1,21 +1,5 @@
 #include "common.h"
 
-#include <sys/types.h>
-#include <sys/stat.h>
-#include <fcntl.h>
-#ifdef _WIN32
-# include <io.h>
-# include <Windows.h>
-
-# define open _open
-# define read _read
-# define close _close
-
-#define ssize_t int
-#else
-# include <unistd.h>
-#endif
-
 /*
  * This could be run in the main loop whilst the application waits for
  * the indexing to finish in a worker thread
--- a/examples/status.c
+++ b/examples/status.c
@@ -13,12 +13,6 @@
  */
 
 #include "common.h"
-#ifdef _WIN32
-# include <Windows.h>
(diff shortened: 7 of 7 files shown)


albfan:
Not sure if there is any better code to load an entire file into a char* (or void*):

https://codereview.stackexchange.com/q/163424

neithernut:
[mmap](http://man7.org/linux/man-pages/man2/mmap.2.html) (or rather the [portability-interface](https://github.com/libgit2/libgit2/blob/e6cdd17c846648aa8e1e025fa1988475886a551e/src/map.h) provided by libgit2)?

ethomson:
> mmap (or rather the portability-interface provided by libgit2)?

Let's not add an example that uses that, it's private API and not available to consumers.

I think that the best example would be simplest, and avoiding `mmap` is definitely the simple route.

albfan:
@ethomson we plan to use in libgit2-glib to be able to compare random buffers: See https://gitlab.gnome.org/albfan/diferencia/issues/3

But for sure we can close this PR.

albfan:
Read files is implemented right now:

```
$ echo hello > file1
$ echo bye > file2
$ ./lg2 diff --no-index file1 file2
diff --git a/file1 b/file2
index ce01362..b023018 100644
--- a/file1
+++ b/file2
@@ -1 +1 @@
-hello
+bye
```

Let me know any further change to do. We plan to use this on https://gitlab.gnome.org/albfan/diferencia/issues/3 and integrate in libgit2-glib so anyone can diff files without being in a repo

examples/diff.c
```diff
@@ -86,49 +151,80 @@ int lg2_diff(git_repository *repo, int argc, char *argv[])
...
-			"diff trees", NULL);
-	else if (o.cache != CACHE_NORMAL) {
-		if (!t1)
-			treeish_to_tree(&t1, repo, "HEAD");
-
-		if (o.cache == CACHE_NONE)
+			git_patch_from_buffers(&patch, file1_str, strlen(file1_str), o.treeish1, file2_str, strlen(file2_str), o.treeish2, &o.diffopts),
+			"patch buffers", NULL);
+		check_lg2(
+		   git_patch_to_buf(&buf, patch),
```
tiennou: Though not your fault, I kinda dislike that style…

This is a styleguide question for maintainers, do we forgo `check_lg2` if that's too heavy-handed/confusing, or switch to one of our `if ((err = func()) < 0 || …)` patterns instead, if it's mostly related to one task (here loading a diff from a buffer) ?

examples/diff.c
```diff
@@ -67,13 +75,70 @@ static int color_printer(
...
+
+    return result;
+}
+
 int lg2_diff(git_repository *repo, int argc, char *argv[])
 {
 	git_tree *t1 = NULL, *t2 = NULL;
 	git_diff *diff;
+	git_patch *patch = NULL;
+	char* file1_str = NULL;
```
tiennou: Nitpick: `char *ptr`.

examples/diff.c
```diff
@@ -67,13 +75,70 @@ static int color_printer(
...
+/*
+ * Read a file path
+ *
+ * path: a file path (char*)
+ *
+ * returns char* with the contents of the file (must be freed) or NULL if
+ * file does not exist or memory allocation failed
+ */
+char* read_file(const char* path) {
+    FILE* file;
```
tiennou: Indent. Also, `char *`.

examples/diff.c
```diff
@@ -67,13 +75,70 @@ static int color_printer(
...
 
+/*
+ * Read a file path
+ *
+ * path: a file path (char*)
+ *
+ * returns char* with the contents of the file (must be freed) or NULL if
+ * file does not exist or memory allocation failed
+ */
+char* read_file(const char* path) {
```
tiennou: I'd be more comfortable having that in "the example's common library". Reading files from a `char*` seems generally useful to example code.

tiennou:
> Let's not add an example that uses that, it's private API and not available to consumers.

You do realize some of us are itching to run git's test suite against our "examples", right? Right 😉? Just to say that I'd like to revisit that. Not being able to access libgit2's stdlib makes writing examples painful when you're left longing for `futils`…

examples/diff.c
albfan: Will do ASAP. I'm impressed by your quick review, Yay!

examples/diff.c
```diff
@@ -67,13 +75,70 @@ static int color_printer(
...
+        return NULL;
+	 }
+
+    result = malloc(sizeof(char) * READ_SIZE + 1);
+
+    if (result == NULL) {
+        return NULL;
+	 }
+
+    while ((ch = getc(file)) != EOF) {
```
neithernut: Lecture mode: I see this a lot: people reading an entire file using `getc()`. I assume some devs are not aware that this is usually incredibly slow since it translates to a system call, just to read a single byte. Hence, one of the answers in [the link you provided](https://codereview.stackexchange.com/questions/163424/read-file-into-char) mentions using `fread()` instead.

For a mere example, using `getc()` is acceptable. But if you ever do something like this for production code, you should make avoiding reading single bytes a habit. Btw: if you're on a *nix, you can observe the difference in performance using [`dd`](http://man7.org/linux/man-pages/man1/dd.1.html) with different `bs`-values.

examples/diff.c
```diff
@@ -67,13 +75,70 @@ static int color_printer(
...
+        result[pos++] = ch;
+
+        if(pos >= capacity) {
+            capacity += READ_SIZE;
+
+            result = realloc(result, sizeof(char) * capacity + 1);
+
+            if(result == NULL) {
+                return NULL;
```
neithernut: You leak `file` here.

examples/diff.c
```diff
@@ -67,13 +75,70 @@ static int color_printer(
...
+				}
+        }
+    }
+
+    fclose(file);
+
+    result = realloc(result, sizeof(char) * pos);
```
neithernut: This extra reallocation might not be worth it. You gain a maximum of `READ_SIZE-1` bytes, which is just a quarter of a kb with your default value. Unless `READ_SIZE` is much greater than the size of a memory page or you allocate some _small_ objects afterwards, you will not gain any memory back. It doesn't matter for a short-lived process anyway.

examples/diff.c
```diff
@@ -67,13 +75,70 @@ static int color_printer(
...
+    file = fopen(path, "r");
+
+    if (!file) {
+        return NULL;
+	 }
+
+    result = malloc(sizeof(char) * READ_SIZE + 1);
+
+    if (result == NULL) {
+        return NULL;
```
neithernut: You leak `file` here.

examples/diff.c
albfan: Just this review remains, not sure how to fix this: 

Is it something like this?
```
check_lg2(
    !git_patch_from_buffers(&patch, file1_str, strlen(file1_str), o.treeish1,
        file2_str, strlen(file2_str), o.treeish2, &o.diffopts) &&
    !git_patch_to_buf(&buf, patch) &&
    !git_diff_from_buffer(&diff, buf.ptr, buf.size),
    "diff buffers", NULL);
```

tiennou: @pks-t Any opinions ? I'm not a fan of having `check_lg2` in example code in the first place (it makes the error handling brittle), and now it impacts readability (somewhat).

I don't think its worth blocking that PR on that though, that's just food for thought for the next time I look at example code.

examples/common.c
```diff
@@ -391,3 +395,61 @@ int cred_acquire_cb(git_cred **out,
...
+char *read_file(const char *path) {
+	FILE *file;
+	char *result;
+	size_t pos = 0;
+	size_t capacity = READ_SIZE;
+	size_t blocksRead= 0;
+	size_t totalRead = 0;
+
+	file = fopen(path, "r");
+
```
tiennou: Style: 2-line `if`s can lose their braces.

examples/common.c
```diff
@@ -391,3 +395,61 @@ int cred_acquire_cb(git_cred **out,
...
+	if (!file) {
+		return NULL;
+	}
+
+	result = malloc(sizeof(char) * capacity + 1);
+
```
tiennou: Ditto.

examples/common.c
```diff
@@ -391,3 +395,61 @@ int cred_acquire_cb(git_cred **out,
...
+	if (result == NULL) {
+		return NULL;
+	}
+
+	while(1) {
+		size_t read = fread(result + (blocksRead * capacity) + totalRead, 1, capacity - totalRead, file);
```
tiennou: `read` as a local might confuse some compilers. `readLen` ?

examples/common.c
```diff
@@ -391,3 +395,61 @@ int cred_acquire_cb(git_cred **out,
 	free(pubkey);
 	return error;
 }
+
+char *read_file(const char *path) {
+	FILE *file;
+	char *result;
+	size_t pos = 0;
+	size_t capacity = READ_SIZE;
```
tiennou: `BUFSIZ`? That's supposed to be a somewhat optimal value from `<stdio.h>`.

examples/diff.c
```diff
@@ -25,6 +26,8 @@
  * how to do renaming detection and patch formatting.
  */
 
+char git_buf__initbuf[1];
```
tiennou: That's supposed to be a private implementation detail of `git_buf`, and there should be no need to define it — it's used as the default static value that git_bufs are initialized to point to. I believe things will break in strange ways if it gets duplicated.

examples/diff.c
```diff
@@ -181,12 +221,12 @@ static int color_printer(
 
 	if (*last_color >= 0) {
 		switch (line->origin) {
-		case GIT_DIFF_LINE_ADDITION:  color = 3; break;
```
tiennou: Some spurious whitespace churn here.

examples/common.c
```diff
@@ -391,3 +395,61 @@ int cred_acquire_cb(git_cred **out,
...
+		return NULL;
+	}
+
+	result = malloc(sizeof(char) * capacity + 1);
+
+	if (result == NULL) {
+		return NULL;
+	}
+
+	while(1) {
```
tiennou: I think you could simplify that to :

```c
size_t pos = 0, blockReads = 1;
while (pos += fread(results + pos, BUFSIZE, 1, file) && !ferror(file) && !feof(file)) {
    blockReads += 1;
    result = realloc(result, sizeof(result) * ((++blocksRead)) + 1);
    …
}
```

You're also reading `READ_LEN` objects of length 1, instead of 1 object of `READ_LEN`. I'm not sure it really matters for performance, but it's still… strange 😉.

examples/diff.c
```diff
@@ -86,49 +94,81 @@ int lg2_diff(git_repository *repo, int argc, char *argv[])
...
+			usage("file cannot be read", o.treeish1);
+		}
+		file2_str = read_file(o.treeish2);
+		if (file2_str == NULL) {
+			usage("file cannot be read", o.treeish2);
+		}
+	}
 
-	if (t1 && t2)
+	if (file1_str && file2_str) {
```
tiennou: Nitpick: I'm thinking that adding

```c
static int build_diff_from_buffers(git_diff **out, const char *b1, size_t b1len, const char *b2 size_t b2len, struct opts *diffopts);
static int build_diff_from_trees(git_diff **out, git_repository *repo, git_tree *t1, git_tree *t2, struct opts *diffopts);
```
would improve readability. That "main" function is getting pretty long, and the error path gets twisty.

examples/common.c
albfan: Yes, that's my personal choice, but if without braces are all around the code so I will fix that

examples/diff.c
albfan:
```
/usr/bin/ld: CMakeFiles/lg2.dir/diff.c.o: en la función `lg2_diff':
/home/alberto/projects/c/libgit2/examples/diff.c:116: referencia a `git_buf__initbuf' sin definir
collect2: error: ld devolvió el estado de salida 1
make[2]: *** [examples/CMakeFiles/lg2.dir/build.make:436: examples/lg2] Error 1
make[1]: *** [CMakeFiles/Makefile2:359: examples/CMakeFiles/lg2.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

I didn't find a way to solve this, maybe make diff depends on buffer.c in CMakeList.txt?

```
src/buffer.c
16:char git_buf__initbuf[1];
```

examples/common.c
albfan: there will be a problem with ferror if that doesn't happen at first (like read half of the file, no?)

I agreed your code is more readable but not able to replace as suggested

examples/diff.c
albfan: This will imply to remove the check_lg2 and check exit status for `build_diff_from_trees` so i'm not sure how to do that in a proper way

examples/diff.c
ethomson: Don't set `buf = git_buf__initbuf`, either set `buf = {0}` or don't initialize it and call `git_buf_init(&buf)`.

albfan: ok, I was using `GIT_BUF_INIT` macro

examples/common.c
tiennou: My code is missing a follow-up `if (ferror(file)) /* cleanup */;` as well, so there's that. I don't think it can happen: `pos` will have advanced anyways, and the `realloc` only needs to happen if we read a full block.

albfan:
We are on air! https://gitlab.gnome.org/GNOME/libgit2-glib/commit/d5f41735489ac31f389de66edaba7a2c55d67024

thanks!. If you want to avoid creating a patch to extract a diff to do this, I would be glad to help.

examples/diff.c
```diff
@@ -13,6 +13,7 @@
  */
 
 #include "common.h"
+#include "buffer.h"
```
tiennou: This causes CI to fail:
```
In file included from /src/src/common.h:182:0,
                 from /src/src/buffer.h:10,
                 from /src/examples/diff.c:16:
/src/src/util.h:23:0: error: "ARRAY_SIZE" redefined [-Werror]
 #define ARRAY_SIZE(x) (sizeof(x)/sizeof(x[0]))
 ^
In file included from /src/examples/diff.c:15:0:
/src/examples/common.h:34:0: note: this is the location of the previous definition
 #define ARRAY_SIZE(x) (sizeof(x)/sizeof(*x))
 ^
```

I don't think it's needed, now that the `git_buf__initbuf` is gone.

pks-t:
On Tue, Jun 18, 2019 at 01:12:32PM -0700, Etienne Samson wrote:
> > Let's not add an example that uses that, it's private API and
> > not available to consumers.
> 
> You do realize some of us are itching to run git's test suite
> against our "examples", right? Right 😉? Just to say that I'd
> like to revisit that. Not being able to access libgit2's stdlib
> makes writing examples painful when you're left longing for
> `futils`…

I agree with @ethomson, though. While it'd be nice to be able to
run git.git's tests against libgit2, the most important goal of
examples is to demonstrate how to use libgit2's API in another
project. As such, we just cannot use our own internal APIs, as
otherwise it may become much harder for others to simply lift out
the code from our examples into their own codebase. I also want
to avoid that we start using internal APIs by accident.

pks-t:
On Wed, Jun 19, 2019 at 01:09:10AM -0700, Etienne Samson wrote:
> tiennou commented on this pull request.
> 
> 
> 
> >  		check_lg2(
> -			git_diff_tree_to_tree(&diff, repo, t1, t2, &o.diffopts),
> -			"diff trees", NULL);
> -	else if (o.cache != CACHE_NORMAL) {
> -		if (!t1)
> -			treeish_to_tree(&t1, repo, "HEAD");
> -
> -		if (o.cache == CACHE_NONE)
> +			git_patch_from_buffers(&patch, file1_str, strlen(file1_str), o.treeish1, file2_str, strlen(file2_str), o.treeish2, &o.diffopts),
> +			"patch buffers", NULL);
> +		check_lg2(
> +		   git_patch_to_buf(&buf, patch),
> 
> @pks-t Any opinions ? I'm not a fan of having `check_lg2` in example code in the first place (it makes the error handling brittle), and now it impacts readability (somewhat).

Will try to have a look tomorrow. No promise, though.

albfan:
Just to mention, we are facing a problem when differing files with no newline at end. Seems patch is not parseable.

Is there any config or workaround? Compare buffers without need to convert to a patch would be nice. Right now we just always add newline at end if it not exists.

Let me know if I can rework any part to get this merged

pks-t:
On Thu, Jun 27, 2019 at 04:07:01AM -0700, Alberto Fanjul wrote:
> Just to mention, we are facing a problem when differing files
> with no newline at end. Seems patch is not parseable.
> 
> Is there any config or workaround? Compare buffers without need
> to convert to a patch would be nice. Right now we just always
> add newline at end if it not exists.

This sounds like a bug to me. Could you please create a bug for
this so that we can handle it separately?

> Let me know if I can rework any part to get this merged

I'll try to give feedback tomorrow :)

albfan:
@pks-t check #5153

examples/common.c
```diff
@@ -391,3 +391,58 @@ int cred_acquire_cb(git_cred **out,
 	free(pubkey);
 	return error;
 }
+
+char *read_file(const char *path) {
+	FILE *file;
```
pks-t: May I propose this alternate implementation:

```
char *read_file(const char *path)
{
    ssize_t total = 0;
    char *buf = NULL;
    struct stat st;
    int fd = -1;

    if ((fd = open(path, O_RDONLY)) < 0 || fstat(fd, &st) < 0)
        goto out;

    if ((buf = malloc(st.st_size + 1)) == NULL)
        goto out;

    while (total < st.st_size) {
        ssize_t bytes = read(fd, buf + total, st.st_size - total);
        if (bytes <= 0 && errno != EAGAIN && errno != EINTR) {
            free(buf);
            buf = NULL;
            goto out;
        }
        total += bytes;
    }

    buf[total] = '\0';

out:
    if (fd >= 0)
        close(fd);
    return buf;
}
```

examples/common.h
```diff
@@ -63,6 +63,16 @@ extern int lg2_tag(git_repository *repo, int argc, char **argv);
  */
 extern void check_lg2(int error, const char *message, const char *extra);
 
+/**
+ * Read a file path
+ *
+ * path: a file path (char *)
+ *
+ * returns char * with the contents of the file (must be freed) or NULL if
+ * file does not exist or memory allocation failed
```
pks-t: We use a custom documentation format that's similar to Javadoc. So this should probably be something like:

```
/**
 * Read a file into a buffer
 * @param path: path to the file that shall be read
 * @return NUL-terminated buffer if the file was successfully read, NULL-pointer otherwise
 */
```

examples/diff.c
```diff
@@ -86,49 +91,81 @@ int lg2_diff(git_repository *repo, int argc, char *argv[])
...
 	 *
 	 * Currently ranged arguments like &lt;sha1&gt;..&lt;sha2&gt; and &lt;sha1&gt;...&lt;sha2&gt;
 	 * are not supported in this example
 	 */
 
-	if (o.treeish1)
-		treeish_to_tree(&t1, repo, o.treeish1);
-	if (o.treeish2)
-		treeish_to_tree(&t2, repo, o.treeish2);
+	if (o.no_index >= 0) {
```
pks-t: This function is getting rather complex. I'd be very grateful if you'd split it up into multiple functions that compute a diff. E.g. something like:

```
if (o.so_index >= 0)
    compute_diff_no_index(&patch);
else if (o.cache == CACHE_NONE)
    compute_diff_no_cache(&patch);
else
    ...
```

Otherwise it's really hard to review this, even though it's not your fault :(

examples/diff.c
```diff
@@ -238,6 +275,8 @@ static void parse_opts(struct opts *o, int argc, char *argv[])
 			o->format = GIT_DIFF_FORMAT_RAW;
 			o->diffopts.id_abbrev = 40;
 		}
+		else if (!strcmp(a, "--no-index"))
+			o->no_index = 0;
```
pks-t: We should catch incompatibilities between options. Like for example "--cached" doesn't make any sense with "--no-index", right?

examples/common.c
neithernut: Would that work on Windows?

pks-t: Fair question, but yes it does, I've just tested with MSVC2017. It requires some different header files than on Unix systems, though, most importantly "io.h"

albfan: Done

examples/common.h
albfan: Done

examples/diff.c
albfan: Done

examples/diff.c
albfan: Resolved for new no-index, not sure about rest of the code

examples/common.c
albfan: This code breaks on linux:

```
$ gdb --args ./examples/lg2 diff --no-index --cached examples/file1 examples/file2
GNU gdb (GDB) 8.3
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./examples/lg2...
(gdb) run
Starting program: /home/alberto/projects/c/libgit2/examples/lg2 diff --no-index --cached examples/file1 examples/file2
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7ebadaf in diff_print_info_init_fromdiff (pi=0x7fffffffd460, out=0x7fffffffd440, diff=0x2, format=GIT_DIFF_FORMAT_PATCH, cb=0x55555555e4de <color_printer>, 
    payload=0x7fffffffd5a0) at /home/alberto/projects/c/libgit2/src/diff_print.c:74
74		git_repository *repo = diff ? diff->repo : NULL;
(gdb) bt
#0  0x00007ffff7ebadaf in diff_print_info_init_fromdiff (pi=0x7fffffffd460, out=0x7fffffffd440, diff=0x2, format=GIT_DIFF_FORMAT_PATCH, 
    cb=0x55555555e4de <color_printer>, payload=0x7fffffffd5a0) at /home/alberto/projects/c/libgit2/src/diff_print.c:74
#1  0x00007ffff7ebc764 in git_diff_print (diff=0x2, format=GIT_DIFF_FORMAT_PATCH, print_cb=0x55555555e4de <color_printer>, payload=0x7fffffffd5a0)
    at /home/alberto/projects/c/libgit2/src/diff_print.c:690
#2  0x000055555555e1bb in lg2_diff (repo=0x555555662490, argc=5, argv=0x7fffffffd7a0) at /home/alberto/projects/c/libgit2/examples/diff.c:153
#3  0x0000555555561081 in run_command (fn=0x55555555de34 <lg2_diff>, repo=0x555555662490, args=...) at /home/alberto/projects/c/libgit2/examples/lg2.c:42
#4  0x000055555556136f in main (argc=6, argv=0x7fffffffd798) at /home/alberto/projects/c/libgit2/examples/lg2.c:110
(gdb) 
```

Didn't dig into it. Not sure what is the problem

examples/common.c
```diff
@@ -14,6 +14,9 @@
 
 #include <assert.h>
 #include <stdio.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
```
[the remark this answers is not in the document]
pks-t: Oh, and you're missing "unistd.h" for `close` and `read` :)

examples/common.c
```diff
@@ -391,3 +394,37 @@ int cred_acquire_cb(git_cred **out,
...
+        total += bytes;
+    }
+
+    buf[total] = '\0';
+
+out:
+    if (fd >= 0)
+        close(fd);
+    return buf;
+}
```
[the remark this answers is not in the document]
albfan: Yes, fixed

examples/common.c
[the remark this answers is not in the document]
albfan: Yes, I expected something like that for windows.

examples/common.h
```diff
@@ -63,6 +63,14 @@ extern int lg2_tag(git_repository *repo, int argc, char **argv);
  */
 extern void check_lg2(int error, const char *message, const char *extra);
 
+/**
+ * Read a file into a buffer
+ *
+ * @param path: path to the file that shall be read
```
neithernut: The build currently fails because the `:` is interpreted as part of the parameter name (`clang` tries to match the documentation with the signature). You should omit the `:`.

albfan: Opps, fixed!

pks-t:
I've amended a commit to this PR to fix Win32 builds so that you don't have to handle this :)

albfan:
@pks-t  Nice to see you in action! haha

examples/diff.c
```diff
@@ -158,6 +166,38 @@ int lg2_diff(git_repository *repo, int argc, char *argv[])
...
+	file1_str = read_file(o->treeish1);
+	if (file1_str == NULL) {
+		usage("file cannot be read", o->treeish1);
+	}
+	file2_str = read_file(o->treeish2);
+	if (file2_str == NULL) {
+		usage("file cannot be read", o->treeish2);
+	}
+	check_lg2(
+		git_patch_from_buffers(&patch, file1_str, strlen(file1_str), o->treeish1, file2_str, strlen(file2_str), o->treeish2, &o->diffopts),
```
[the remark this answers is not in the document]
albfan: I would do this in a different PR

albfan:
Still finding github PR hard to follow

> Merging can be performed automatically once the requested changes are addressed.

Can't find anything else to do. Is there some "I approve this PR" check to press or something?

neithernut:
One of the maintainers has to approve the changes (and then merge them). This may take a while for this project due to them being understaffed/limited in bandwidth. So there is nothing else to do except for waiting patiently.

albfan:
@neithernut I was sure about that. All the process was really awesome and supportive. We will try to provide feedback on #5161, and use as patch meanwhile it is being reviewed.

pks-t:
#5158 has been merged, so I've merged this now. Thanks a lot for your work!

PR #5128, opened 2019-06-19
More documentation

tiennou:
I've added a "missing field documentation" warning to Docurium, and this is a first pass at it.

I've noticed that Docurium seems to dislike the `GIT_CALLBACK` macro when it's used in a struct definition. Typedefs of those seems to be fine though.

### DIFF
--- a/include/git2/checkout.h
+++ b/include/git2/checkout.h
@@ -261,7 +261,7 @@ typedef void GIT_CALLBACK(git_checkout_perfdata_cb)(
  *
  */
 typedef struct git_checkout_options {
-	unsigned int version;
+	unsigned int version; /**< The version */
 
 	unsigned int checkout_strategy; /**< default will be a safe checkout */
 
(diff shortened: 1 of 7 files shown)


include/git2/checkout.h
```diff
@@ -271,29 +271,46 @@ typedef struct git_checkout_options {
...
 	git_checkout_progress_cb progress_cb;
+
+	/** Payload passed to progress_cb */
 	void *progress_payload;
 
-	/** When not zeroed out, array of fnmatch patterns specifying which
-	 *  paths should be taken into account, otherwise all files.  Use
-	 *  GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH to treat as simple list.
+	/**
+	 * A list of fnmatch patterns or paths.
```
pks-t: It's not an `fnmatch` pattern anymore, but a `wildmatch` pattern. They are slightly different

tiennou: Fixed, and ready to go!

pks-t:
And thanks again for your awesome work!

PR #5131, opened 2019-06-21
fileops: fix creation of directory in filesystem root

### LINKED ISSUE #5130: Cannot init repo in root directory with absolute path
### Reproduction steps
Call `git_repository_init_ext` with path `/foo` on a posix-like system (not windows) for a path that does not exist.

```c
git_repository *repo;
git_repository_init_options initopts = GIT_REPOSITORY_INIT_OPTIONS_INIT;
git_libgit2_init();
initopts.flags |= GIT_REPOSITORY_INIT_MKDIR | GIT_REPOSITORY_INIT_MKPATH;
int error = git_repository_init_ext(&repo, "/foo", &initopts);
if (error) {
    const git_error *err = git_error_last();
    fprintf(stderr, "[%d] %s\n", err->klass, err->message);
}
```

### Expected behavior
Creates a new repo `/foo/.git`

### Actual behavior
Fails with `[2] failed to make directory '/foo/.git': No such file or directory`

55a1535dd51c62f4e4ed764412200bbd49f3a1ab

Linux or macOS.

### Notes
This is caused by the change in #5050.

pks-t:
In commit 45f24e787 (git_repository_init: stop traversing at
windows root, 2019-04-12), we have fixed `git_futils_mkdir` to
correctly handle the case where we create a directory in
Windows-style filesystem roots like "C:\repo".

The problem here is an off-by-one: previously, to that commit,
we've been checking wether the parent directory's length is equal
to the root directory's length incremented by one. When we call
the function with "/example", then the parent directory's length
("/") is 1, but the root directory offset is 0 as the path is
directly rooted without a drive prefix. This resulted in `1 == 0 +
1`, which was true. With the change, we've stopped incrementing
the root directory length, and thus now compare `1 <= 0`, which
is false.

The previous way of doing it was kind of finicky any non-obvious,
which is also why the error was introduced. So instead of just
re-adding the increment, let's explicitly add a condition that
aborts finding the parent if the current parent path is "/".

---

Fixes #5130

### DIFF
--- a/src/fileops.c
+++ b/src/fileops.c
@@ -495,7 +495,9 @@ int git_futils_mkdir(
 		 * equal to length of the root path).  The path may be less than the
 		 * root path length on Windows, where `C:` == `C:/`.
 		 */
-		if ((len == 1 && parent_path.ptr[0] == '.') || len <= root_len) {
+		if ((len == 1 && parent_path.ptr[0] == '.') ||
+		    (len == 1 && parent_path.ptr[0] == '/') ||
+		    len <= root_len) {
 			relative = make_path.ptr;
 			break;
 		}
--- a/tests/repo/init.c
+++ b/tests/repo/init.c
@@ -878,14 +878,55 @@ void test_repo_init__at_filesystem_root(void)
 	git_repository_free(repo);
 }
 
-void test_repo_init__nonexistent_paths(void)
+void test_repo_init__nonexisting_directory(void)
 {
+	git_repository_init_options opts = GIT_REPOSITORY_INIT_OPTIONS_INIT;
 	git_repository *repo;
 
+	/*
+	 * If creating a repo with non-existing parent directories, then libgit2
+	 * will by default create the complete directory hierarchy if using
+	 * `git_repository_init`. Thus, let's use the extended version and not
+	 * set the `GIT_REPOSITORY_INIT_MKPATH` flag.
+	 */
+	cl_git_fail(git_repository_init_ext(&repo, "nonexisting/path", &opts));
+}
+
+void test_repo_init__nonexisting_root(void)
+{
 #ifdef GIT_WIN32
+	git_repository *repo;
+
+	/*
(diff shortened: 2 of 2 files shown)


pks-t:
Funny. CI now fails on Linux because we now start _succeeding_ creation of repos in a "/non/existing/path", where the first path component doesn't exist either. This is obviously the correct thing to do, because `git_repository_init` defaults to `GIT_REPOSITORY_INIT_MKPATH` to create leading directories. And Azure seems to be kind of lenient with regards to access rights because we can create "/non"

pks-t:
@ethomson: are we executing tests as root? Otherwise we wouldn't be able to create the "/usr/non/" directory...

ethomson:
> @ethomson: are we executing tests as root? Otherwise we wouldn't be able to create the "/usr/non/" directory...

We are.

pks-t:
On Mon, Jun 24, 2019 at 09:40:59AM -0700, Edward Thomson wrote:
> > @ethomson: are we executing tests as root? Otherwise we wouldn't be able to create the "/usr/non/" directory...
> 
> We are.

Saw that already. I'll put up a PR in a few minutes to fix that.

ehuss:
I'm curious what the status of this is.  Is there anything I can do to help?

pks-t:
> I'm curious what the status of this is. Is there anything I can do to help?

Probably not too much, but in fact you've just caused me to rethink my approach to how to fix the tests for now. So thanks for asking :)

This was previously blocked on the fact that we were testing in Docker images as root user. But as my changes in libgit2/docker-build#1 weren't yet merged, I've decided to go an alternative route. Thinking about it now the previous test was arguably wrong anyway, so I think we're better off with the new tests now.

The only thing left now is a review to get this merged. It doesn't rely on the Docker PR anymore, even though I still think that we should avoid building and testing as root user.

pks-t:
As always, thanks a lot for your review @tiennou!

Issue #5126, opened 2019-06-19
git_repository_hashfile re-reads config unnecessarily


ethomson:
Some data in #5115 suggests that we're re-reading the configuration unnecessarily in `git_repository_hashfile`.

Given this code:

```
// Usage:  echo foo > a && git add a && git commit -m "Initial commit" && test.exe a
#include <git2.h>
#include <stdio.h>

int main(int argc, char *argv[]) {
	git_libgit2_init();
	git_repository *repo = NULL;
	int error = 0;
	char const *repo_path = ".git";
	if ((error = git_repository_open(&repo, repo_path)) == 0) {
		fprintf(stderr, "git_repository_open(\"%s\") succeeded\n", repo_path);
		fprintf(stderr, "Press ENTER to continue..."); getchar();
		for (int i = 1; i < argc; ++i) {
			git_oid oid;
			if (git_repository_hashfile(&oid, repo, argv[i], GIT_OBJECT_BLOB, NULL) == 0) {
				fprintf(stderr, "git_repository_hashfile(%s) succeeded\n", argv[i]);
			} else {
				fprintf(stderr, "git_repository_hashfile(%s) failed\n", argv[i]);
			}
		}
		fprintf(stderr, "Finished; press ENTER to quit..."); getchar();
		git_repository_free(repo);
	} else {
		fprintf(stderr, "git_repository_open(\"%s\") returned: %d\n", repo_path, error);
	}
	return error;
}
```

procmon shows multiple reads of the config and the index.  https://github.com/libgit2/libgit2/issues/5115#issuecomment-502407992

We can likely reduce many of these.

mehrdadn:
Similarly with `gitattributes`/`.git/info/attributes`/`.gitattributes`.

mehrdadn:
Just a note, I wouldn't just focus on the config files—there seem to be even worse offenders:

- `gitattributes`/`.git/info/attributes`/`.gitattributes` are accessed on *every single* call to `git_repository_hashfile`, not just on the first call. And not just 1 time, but *4 times*. They should only be accessed once, and only in the beginning, not on every call to `git_repository_hashfile`.

- The file whose path passed to `git_repository_hashfile` is also accessed *5 times*: twice for `git_path_isdir`, once for `git_path_find_dir`, and once for reading. These can be reduced to 1, or 2 calls at most.

ethomson:
> They should only be accessed once, and only in the beginning, not on every call to git_repository_hashfile.

[...]

PR #5132, opened 2019-06-21
config_file: implement stat cache to avoid repeated rehashing

pks-t:
To decide whether a config file has changed, we always hash its
complete contents. This is unnecessarily expensive, as
well-behaved filesystems will always update stat information for
files which have changed. So before computing the hash, we should
first check whether the stat info has actually changed for either
the configuration file or any of its includes. This avoids having
to re-read the configuration file and its includes every time
when we check whether it's been modified.

Tracing the for-each-ref example previous to this commit, one can
see that we repeatedly re-open both the repo configuration as
well as the global configuration:

        $ strace lg2 for-each-ref |& grep config
        access("/home/pks/.gitconfig", F_OK)    = -1 ENOENT (No such file or directory)
        access("/home/pks/.config/git/config", F_OK) = 0
        access("/etc/gitconfig", F_OK)          = -1 ENOENT (No such file or directory)
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        access("/tmp/repo/.git/config", F_OK)   = 0
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        open("/tmp/repo/.git/config", O_RDONLY|O_CLOEXEC) = 3
        stat("/home/pks/.gitconfig", 0x7ffd15c05290) = -1 ENOENT (No such file or directory)
        access("/home/pks/.gitconfig", F_OK)    = -1 ENOENT (No such file or directory)
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        access("/home/pks/.config/git/config", F_OK) = 0
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        open("/home/pks/.config/git/config", O_RDONLY|O_CLOEXEC) = 3
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        open("/tmp/repo/.git/config", O_RDONLY|O_CLOEXEC) = 3
        stat("/home/pks/.gitconfig", 0x7ffd15c051f0) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        open("/home/pks/.config/git/config", O_RDONLY|O_CLOEXEC) = 3
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        open("/tmp/repo/.git/config", O_RDONLY|O_CLOEXEC) = 3
        stat("/home/pks/.gitconfig", 0x7ffd15c05090) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        open("/home/pks/.config/git/config", O_RDONLY|O_CLOEXEC) = 3
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        open("/tmp/repo/.git/config", O_RDONLY|O_CLOEXEC) = 3
        stat("/home/pks/.gitconfig", 0x7ffd15c05090) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        open("/home/pks/.config/git/config", O_RDONLY|O_CLOEXEC) = 3
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        open("/tmp/repo/.git/config", O_RDONLY|O_CLOEXEC) = 3
        stat("/home/pks/.gitconfig", 0x7ffd15c05090) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        open("/home/pks/.config/git/config", O_RDONLY|O_CLOEXEC) = 3

With the change, we only do stats for those files and open them a
single time, only:

        $ strace lg2 for-each-ref |& grep config
        access("/home/pks/.gitconfig", F_OK)    = -1 ENOENT (No such file or directory)
        access("/home/pks/.config/git/config", F_OK) = 0
        access("/etc/gitconfig", F_OK)          = -1 ENOENT (No such file or directory)
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        access("/tmp/repo/.git/config", F_OK)   = 0
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        open("/tmp/repo/.git/config", O_RDONLY|O_CLOEXEC) = 3
        stat("/home/pks/.gitconfig", 0x7ffe70540d20) = -1 ENOENT (No such file or directory)
        access("/home/pks/.gitconfig", F_OK)    = -1 ENOENT (No such file or directory)
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        access("/home/pks/.config/git/config", F_OK) = 0
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        open("/home/pks/.config/git/config", O_RDONLY|O_CLOEXEC) = 3
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        stat("/home/pks/.gitconfig", 0x7ffe70540ca0) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.gitconfig", 0x7ffe70540c80) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        stat("/home/pks/.gitconfig", 0x7ffe70540b40) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.gitconfig", 0x7ffe70540b20) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        stat("/home/pks/.gitconfig", 0x7ffe70540b40) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.gitconfig", 0x7ffe70540b20) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0
        stat("/tmp/repo/.git/config", {st_mode=S_IFREG|0644, st_size=92, ...}) = 0
        stat("/home/pks/.gitconfig", 0x7ffe70540b40) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.gitconfig", 0x7ffe70540b20) = -1 ENOENT (No such file or directory)
        stat("/home/pks/.config/git/config", {st_mode=S_IFREG|0644, st_size=1154, ...}) = 0

The following benchmark has been performed with and without the
stat cache in a best-of-ten run:

```

int lg2_repro(git_repository *repo, int argc, char **argv)
{
        git_config *cfg;
        int32_t dummy;
        int i;

        UNUSED(argc);
        UNUSED(argv);

        check_lg2(git_repository_config(&cfg, repo),
                        "Could not obtain config", NULL);

        for (i = 1; i < 100000; ++i)
                git_config_get_int32(&dummy, cfg, "foo.bar");

        git_config_free(cfg);
        return 0;
}
```

Without stat cache:

        $ time lg2 repro
        real    0m1.528s
        user    0m0.568s
        sys     0m0.944s

With stat cache:

        $ time lg2 repro
        real    0m0.526s
        user    0m0.268s
        sys     0m0.258s

This benchmark shows a nearly three-fold performance improvement.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -284,8 +284,6 @@ static int config_set(git_config_backend *cfg, const char *name, const char *val
 	if ((error = config_write(b, name, key, NULL, esc_value)) < 0)
 		goto out;
 
-	error = config_refresh(cfg);
-
 out:
 	git_config_entries_free(entries);
 	git__free(esc_value);
@@ -352,8 +350,6 @@ static int config_set_multivar(
 	if ((result = config_write(b, name, key, &preg, value)) < 0)
 		goto out;
 
-	result = config_refresh(cfg);
-
 out:
 	git__free(key);
 	p_regfree(&preg);
@@ -383,9 +379,6 @@ static int config_delete(git_config_backend *cfg, const char *name)
 	}
 
 	if ((error = config_write(b, name, entry->name, NULL, NULL)) < 0)
-		goto out;
-
-	if ((error =  config_refresh(cfg)) < 0)
 		goto out;
 
 out:
@@ -424,9 +417,6 @@ static int config_delete_multivar(git_config_backend *cfg, const char *name, con
 	}
 
 	if ((result = config_write(b, name, key, &preg, NULL)) < 0)
-		goto out;
-
-	if ((result = config_refresh(cfg)) < 0)
 		goto out;
 
 out:
@@ -1208,7 +1198,12 @@ static int config_write(diskfile_backend *cfg, const char *orig_key, const char 
 		git_buf_attach(&cfg->locked_content, git_buf_detach(&buf), len);
 	} else {
 		git_filebuf_write(&file, git_buf_cstr(&buf), git_buf_len(&buf));
-		result = git_filebuf_commit(&file);
+
+		if ((result = git_filebuf_commit(&file)) < 0)
+			goto done;
+
+		if ((result = config_refresh(&cfg->header.parent)) < 0)
+			goto done;
 	}
 
 done:
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -200,9 +200,27 @@ out:
 	return error;
 }
 
+static int config_refresh_from_buffer(git_config_backend *cfg, const char *buf, size_t buflen)
+{
+	diskfile_backend *b = GIT_CONTAINER_OF(cfg, diskfile_backend, header.parent);
+	git_config_entries *entries = NULL;
+	int error;
+
+	if ((error = git_config_entries_new(&entries)) < 0 ||
+	    (error = config_read_buffer(entries, b->header.repo, &b->file,
+					b->header.level, 0, buf, buflen)) < 0 ||
+	    (error = config_set_entries(cfg, entries)) < 0)
+		goto out;
+
+	entries = NULL;
+out:
+	git_config_entries_free(entries);
+	return error;
+}
+
 static int config_refresh(git_config_backend *cfg)
 {
-	diskfile_backend *b = (diskfile_backend *)cfg;
+	diskfile_backend *b = GIT_CONTAINER_OF(cfg, diskfile_backend, header.parent);
 	git_config_entries *entries = NULL;
 	int error, modified;
 
@@ -1230,7 +1248,7 @@ static int config_write(diskfile_backend *cfg, const char *orig_key, const char 
 		if ((result = git_filebuf_commit(&file)) < 0)
 			goto done;
 
-		if ((result = config_refresh(&cfg->header.parent)) < 0)
+		if ((result = config_refresh_from_buffer(&cfg->header.parent, buf.ptr, buf.size)) < 0)
 			goto done;
 	}
 
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -170,16 +170,41 @@ out:
 	return error;
 }
 
-static int config_refresh(git_config_backend *cfg)
+static int config_set_entries(git_config_backend *cfg, git_config_entries *entries)
 {
 	diskfile_backend *b = (diskfile_backend *)cfg;
-	git_config_entries *entries = NULL, *tmp;
+	git_config_entries *old = NULL;
 	git_config_file *include;
-	int error, modified;
-	uint32_t i;
+	int error;
+	size_t i;
 
 	if (b->header.parent.readonly)
 		return config_error_readonly();
+
+	git_array_foreach(b->file.includes, i, include)
+		config_file_clear(include);
+	git_array_clear(b->file.includes);
+
+	if ((error = git_mutex_lock(&b->header.values_mutex)) < 0) {
+		git_error_set(GIT_ERROR_OS, "failed to lock config backend");
+		goto out;
+	}
+
+	old = b->header.entries;
+	b->header.entries = entries;
+
+	git_mutex_unlock(&b->header.values_mutex);
+
+out:
+	git_config_entries_free(old);
+	return error;
+}
+
+static int config_refresh(git_config_backend *cfg)
+{
+	diskfile_backend *b = (diskfile_backend *)cfg;
+	git_config_entries *entries = NULL;
+	int error, modified;
 
 	error = config_is_modified(&modified, &b->file);
 	if (error < 0 && error != GIT_ENOTFOUND)
@@ -188,29 +213,12 @@ static int config_refresh(git_config_backend *cfg)
 	if (!modified)
 		return 0;
 
-	if ((error = git_config_entries_new(&entries)) < 0)
+	if ((error = git_config_entries_new(&entries)) < 0 ||
+	    (error = config_read(entries, b->header.repo, &b->file, b->header.level, 0)) < 0 ||
+	    (error = config_set_entries(cfg, entries)) < 0)
 		goto out;
 
-	/* Reparse the current configuration */
-	git_array_foreach(b->file.includes, i, include) {
-		config_file_clear(include);
-	}
-	git_array_clear(b->file.includes);
-
-	if ((error = config_read(entries, b->header.repo, &b->file, b->header.level, 0)) < 0)
-		goto out;
-
-	if ((error = git_mutex_lock(&b->header.values_mutex)) < 0) {
-		git_error_set(GIT_ERROR_OS, "failed to lock config backend");
-		goto out;
-	}
-
-	tmp = b->header.entries;
-	b->header.entries = entries;
-	entries = tmp;
-
-	git_mutex_unlock(&b->header.values_mutex);
-
+	entries = NULL;
 out:
 	git_config_entries_free(entries);
 
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -62,6 +62,7 @@ typedef struct {
 } diskfile_parse_state;
 
 static int config_read(git_config_entries *entries, const git_repository *repo, git_config_file *file, git_config_level_t level, int depth);
+static int config_read_buffer(git_config_entries *entries, const git_repository *repo, git_config_file *file, git_config_level_t level, int depth, const char *buf, size_t buflen);
 static int config_write(diskfile_backend *cfg, const char *orig_key, const char *key, const p_regex_t *preg, const char *value);
 static char *escape_value(const char *ptr);
 
@@ -845,6 +846,46 @@ static int read_on_variable(
 	return result;
 }
 
+static int config_read_buffer(
+	git_config_entries *entries,
+	const git_repository *repo,
+	git_config_file *file,
+	git_config_level_t level,
+	int depth,
+	const char *buf,
+	size_t buflen)
+{
+	diskfile_parse_state parse_data;
+	git_config_parser reader;
+	int error;
+
+	if (depth >= MAX_INCLUDE_DEPTH) {
+		git_error_set(GIT_ERROR_CONFIG, "maximum config include depth reached");
+		return -1;
+	}
+
+	/* Initialize the reading position */
+	reader.file = file;
+	git_parse_ctx_init(&reader.ctx, buf, buflen);
+
+	/* If the file is empty, there's nothing for us to do */
+	if (!reader.ctx.content || *reader.ctx.content == '\0') {
+		error = 0;
+		goto out;
+	}
+
+	parse_data.repo = repo;
+	parse_data.file_path = file->path;
+	parse_data.entries = entries;
+	parse_data.level = level;
+	parse_data.depth = depth;
+
+	error = git_config_parse(&reader, NULL, read_on_variable, NULL, NULL, &parse_data);
+
+out:
+	return error;
+}
+
 static int config_read(
 	git_config_entries *entries,
 	const git_repository *repo,
@@ -852,16 +893,9 @@ static int config_read(
 	git_config_level_t level,
 	int depth)
 {
-	diskfile_parse_state parse_data;
-	git_config_parser reader;
 	git_buf contents = GIT_BUF_INIT;
 	struct stat st;
 	int error;
-
-	if (depth >= MAX_INCLUDE_DEPTH) {
-		git_error_set(GIT_ERROR_CONFIG, "maximum config include depth reached");
-		return -1;
-	}
 
 	if (p_stat(file->path, &st) < 0) {
 		error = git_path_set_error(errno, file->path, "stat");
@@ -871,27 +905,13 @@ static int config_read(
 	if ((error = git_futils_readbuffer(&contents, file->path)) < 0)
 		goto out;
 
-	git_parse_ctx_init(&reader.ctx, contents.ptr, contents.size);
-
 	git_futils_filestamp_set_from_stat(&file->stamp, &st);
 	if ((error = git_hash_buf(&file->checksum, contents.ptr, contents.size)) < 0)
 		goto out;
 
-	/* Initialize the reading position */
-	reader.file = file;
-	git_parse_ctx_init(&reader.ctx, contents.ptr, contents.size);
-
-	/* If the file is empty, there's nothing for us to do */
-	if (!reader.ctx.content || *reader.ctx.content == '\0')
+	if ((error = config_read_buffer(entries, repo, file, level, depth,
+					contents.ptr, contents.size)) < 0)
 		goto out;
-
-	parse_data.repo = repo;
-	parse_data.file_path = file->path;
-	parse_data.entries = entries;
-	parse_data.level = level;
-	parse_data.depth = depth;
-
-	error = git_config_parse(&reader, NULL, read_on_variable, NULL, NULL, &parse_data);
 
 out:
 	git_buf_dispose(&contents);
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -44,7 +44,7 @@ typedef struct {
 	git_filebuf locked_buf;
 	git_buf locked_content;
 
-	struct config_file file;
+	git_config_file file;
 } diskfile_backend;
 
 typedef struct {
@@ -95,9 +95,9 @@ static git_config_entries *diskfile_entries_take(diskfile_header *h)
 	return entries;
 }
 
-static void config_file_clear(struct config_file *file)
+static void config_file_clear(git_config_file *file)
 {
-	struct config_file *include;
+	git_config_file *include;
 	uint32_t i;
 
 	if (file == NULL)
@@ -133,7 +133,7 @@ static int config_open(git_config_backend *cfg, git_config_level_t level, const 
 	return res;
 }
 
-static int config_is_modified(int *modified, struct config_file *file)
+static int config_is_modified(int *modified, git_config_file *file)
 {
 	git_config_file *include;
 	git_buf buf = GIT_BUF_INIT;
@@ -659,7 +659,7 @@ static char *escape_value(const char *ptr)
 static int parse_include(git_config_parser *reader,
 	diskfile_parse_state *parse_data, const char *file)
 {
-	struct config_file *include;
+	git_config_file *include;
 	git_buf path = GIT_BUF_INIT;
 	char *dir;
 	int result;
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -143,6 +143,9 @@ static int config_is_modified(int *modified, git_config_file *file)
 
 	*modified = 0;
 
+	if (!git_futils_filestamp_check(&file->stamp, file->path))
+		goto check_includes;
+
 	if ((error = git_futils_readbuffer(&buf, file->path)) < 0)
 		goto out;
 
@@ -154,6 +157,7 @@ static int config_is_modified(int *modified, git_config_file *file)
 		goto out;
 	}
 
+check_includes:
 	git_array_foreach(file->includes, i, include) {
 		if ((error = config_is_modified(modified, include)) < 0 || *modified)
 			goto out;
@@ -861,6 +865,7 @@ static int config_read(
 	diskfile_parse_state parse_data;
 	git_config_parser reader;
 	git_buf contents = GIT_BUF_INIT;
+	struct stat st;
 	int error;
 
 	if (depth >= MAX_INCLUDE_DEPTH) {
@@ -868,11 +873,17 @@ static int config_read(
 		return -1;
 	}
 
+	if (p_stat(file->path, &st) < 0) {
+		error = git_path_set_error(errno, file->path, "stat");
+		goto out;
+	}
+
 	if ((error = git_futils_readbuffer(&contents, file->path)) < 0)
 		goto out;
 
 	git_parse_ctx_init(&reader.ctx, contents.ptr, contents.size);
 
+	git_futils_filestamp_set_from_stat(&file->stamp, &st);
 	if ((error = git_hash_buf(&file->checksum, contents.ptr, contents.size)) < 0)
 		goto out;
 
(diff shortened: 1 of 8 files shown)


pks-t:
Fixes #5126? At least parts of it, I guess.

pks-t:
So this in fact breaks Windows-based systems when reading and writing files at nearly the same time due to Windows usually not using sub-second file stats and thus we're racy. To be honest, I don't think this matters and I'd still like to get this merged. If you have two processes reading and writing files at nearly the same time, then I find it totally reasonable to say "Yup, this is racy without any external synchronization mechanism."

mehrdadn:
@pks-t: It seems to avoid the reading of the files (which is great!), but their opening or most other calls---those are still extremely slow. The change was only a handful of FS calls.

An easy (although conservative) way to avoid this in many situations on Windows is to do the following:  the very first time the `.git` directory is accessed, you call `handle = FindFirstChangeNotification(".git/", TRUE, flags)` with all possible filter flags set. What you then do is, every time you _would_ have re-opened a file or tried to stat it, you now instead just call `WaitForSingleObject(handle, 0)` to see if the wait is satisfied. If the wait is satisfied (i.e. doesn't time out), then this means something has changed, and you can do all the stat/reloading checks then. But if it isn't, then you avoid hitting the file system entirely and assume the cache is good.

The above can be in a separate PR; doesn't have to be this one. Just wanted to explain how to do it, since it's quite easy. (You don't even need to implement it on non-Windows systems if you don't want, since just stat'ing like usual on those is probably already probably fast enough.)

mehrdadn:
There's also a harder way to accomplish something similar, which is to keep the config files actually open, then use them directly (e.g. `fstat`) on subsequent calls. That's the "proper" way to access files in general. I expect (though I'm not sure) that it wouldn't be as fast as the above approach since it would still hit the file system drivers, but it would still be much faster than what's currently being done, since the file doesn't need to be reopened every time. However, it can lock the config files and prevents them from being replaced or modified. That may be either more or less user-friendly depending on your point of view. Personally I think the `FindFirstChangeNotification` approach is the way to go, but they're both improvements and in fact not even mutually exclusive, so both of them could be done independently if desired (and they're both performance improvements over the current implementation).

pks-t:
This is the general and non-Windows-specific approach. Yours just wouldn't work in a cross-platform way, at least as far as I know. If there _is_ a cross-platform way then I'd certainly be happy to hear about it

> However, it can lock the config files and prevent them from being replaced or modified.

Yeah, that'd be a huge problem. Many IDEs or editors just keep repos open all the time while the editor is alive, and we just cannot lock any files like that.

mehrdadn:
So what if it's not a platform specific approach though? Just `#ifdef _WIN32`.

pks-t:
> So what if it's a platform specific approach though? Just #ifdef _WIN32.

Yeah, maybe in another increment. I'm not a Windows guy and this here already improves runtime performance by a lot on all platforms.

mehrdadn:
Oh yeah, like I said, it doesn't need to be on this PR or anything. I just mentioned it as a TODO so we don't forget. I might be able to help with the Windows parts too later.

pks-t:
> It seems to avoid the reading of the files (which is great!), but their opening or most other calls---those are still extremely slow. The change was only a handful of FS calls.

By the way: this isn't quite true. Yes, it only removes a handful of calls, but in fact it removes the most hurting ones. Reads are always a lot less efficient compared to stats, but most importantly we now completly avoid re-hashing the file contents to see whether they're still the same. Previously, we would've _always_ hashed the contents.

pks-t:
At a later point we might decide to roll this into our filebuffers. It seems like a good match to have stuff like that as part of them, as there's multiple locations where we use this sort of caching. Furthermore, that'd allow us to more easily implement platform-specific heuristics similar to the one proposed by @mehrdadn

mehrdadn:
> Reads are always a lot less efficient compared to stats,

@pks-t: No, it's *not* wrong. I've been trying to tell this to all of you on both threads, but it seems you still don't realize the sheer magnitude of the numbers. So let me put it in terms of actual numbers for. 

Here's what it looks like on a real benchmark:

- Windows:

	30.15 us/call access()
	34.55 us/call stat()
	30.27 us/call open()/close()
	35.46 us/call open()/fstat()/close()
	36.38 us/call open()/read(16k)/close()

- Linux:

	0.81 us/call access()
	0.52 us/call stat()
	1.28 us/call open()/close()
	1.72 us/call open()/fstat()/close()
	2.47 us/call open()/read(16k)/close()

Here's what you should be observing:

- On Windows, `stat` is **slower** than just opening/closing the file
- On Windows, `stat` is ***over 8x slower* than reading 16k of data**
- On Linux, `stat` is **less than 5x faster** than reading 16k of data.**
- On Linux, `stat` are merely 2.5x` faster than `open`/`close`.
- On Linux, `stat` is **66x FASTER** than on Windows.
- On Windows, the overhead of accessing a file by path is **25 TIMES more than on Linux**, and **dominates the I/O time**.
- On both platforms, **read() has approximately the same performance**.

So you can see `stat` is **pretty comparable with `read` (Linux), if not significantly slower (Windows)**, until you start getting into the **100+ KB** range. And I'm sure you realize that the average file on git is smaller than that, so this is the uncommon case.

Now maybe try benchmarking this yourself your Mac and letting me know what you see:

	#include <stddef.h>
	#include <stdlib.h>
	#include <stdio.h>
	#ifdef _WIN32
	#include <io.h>
	#else
	#include <unistd.h>
	#define O_BINARY 0
	#endif
	#include <errno.h>
	#include <fcntl.h>
	#include <time.h>
	#include <sys/stat.h>

	int fileperftest()
	{
		char const *const name = "testfile123";
		size_t const count =
	#ifdef _WIN32
			1 << 14
	#else
			1 << 20
	#endif
			;
		for (int op = -2; op <= 2; ++op)
		{
			for (int trial = 0; trial < 1; ++trial)
			{
				unsigned int bufsize = 1 << 14;
				void *const buf = malloc(bufsize);
				{
					int const fd = open(name, O_BINARY | O_CREAT | O_WRONLY | O_TRUNC, S_IREAD | S_IWRITE);
					if (fd != -1)
					{
						if (write(fd, buf, bufsize) == -1) { return errno; }
						close(fd);
					}
					else { return errno; }
				}
				clock_t const tbegin = clock();
				for (size_t i = 0; i != count; ++i)
				{
					if (op == -2)
					{
						if (access(name, 0) == -1)
						{ return errno; }
					}
					else if (op == -1)
					{
						struct stat stat_;
						if (stat(name, &stat_) == -1)
						{ return errno; }
					}
					else
					{
						int const fd = open(name, O_BINARY | O_RDONLY);
						if (fd != -1)
						{
							if (op == 1) { struct stat stat_; fstat(fd, &stat_); }
							else if (op == 2) { read(fd, buf, bufsize); }
							close(fd);
						}
						else { return errno; }
					}
				}
				unlink(name);
				free(buf);
				fprintf(stdout, "%.2f us/call %s\n",
					(clock() - tbegin) * 1000000.0 / (CLOCKS_PER_SEC * count),
					op == -2 ? "access()" : op == -1 ? "stat()" : op == 0 ? "open()/close()" : op == 1 ? "open()/fstat()/close()" : "open()/read()/close()");
				fflush(stdout);
			}
		}
		return 0;
	}

	int main() { return fileperftest(); }

ethomson:
> Now maybe try benchmarking this yourself your Mac and letting me know what you see

I don’t understand why you make any assumption about what technologies any of us use.

mehrdadn:
> > Now maybe try benchmarking this yourself your Mac and letting me know what you see
> 
> I don’t understand why you make any assumption about what technologies any of us use.

@ethomson well because he said "I'm not a Windows guy". As to why Mac and not Linux, though, I thought he was the one who said he's the MacOS developer but turns out it was tiennou who said that on the other thread, so I misremembered that bit.

But I'm not "making assumptions", I'm just going by what you yourselves have told me!

ethomson:
> So this in fact breaks Windows-based systems when reading and writing files at nearly the same time due to Windows usually not using sub-second file stats and thus we're racy. 

Indeed it does seem to.  We could add a 100 ns sleep on win32 to that config test, which *should* ensure that we actually update the filetime.

mehrdadn:
> > So this in fact breaks Windows-based systems when reading and writing files at nearly the same time due to Windows usually not using sub-second file stats and thus we're racy.
> 
> Indeed it does seem to. We could add a 100 ns sleep on win32 to that config test, which _should_ ensure that we actually update the filetime.

Sleeping for 100ns on Windows is not as easy as it sounds. You'll have to spin loop and remeasure.

Also, I think he meant `stat`'s precision is 1 second, not 100 nanoseconds. For the latter you'll want to use something like `FindFirstFile` or `GetFileTime`.

ethomson:
> Also, I think he meant stat's precision is 1 second, not 100 nanoseconds. For the latter you'll want to use something like FindFirstFile or GetFileTime.

Our `stat` emulation uses a `FILETIME`.

mehrdadn:
Oh I see, cool.

pks-t:
On Fri, Jun 21, 2019 at 02:35:47PM -0700, mehrdadn wrote:
> > Reads are always a lot less efficient compared to stats,
> 
> @pks-t: No, it's *not* wrong. I've been trying to tell this to
> all of you on both threads, but it seems you still don't
> realize the sheer magnitude of the numbers. So let me put it in
> terms of actual numbers for. 

"Not quite true" and "wrong" are different things. Furthermore I
claimed that the most important part is that we stop re-hashing
contents.

pks-t:
@ethomson: rebased. I've introduced a 1ms sleep, which should probably fix the test

pks-t:
Changed to 10ms due to it still being flaky

ethomson:
> "Not quite true" and "wrong" are different things. Furthermore I
claimed that the most important part is that we stop re-hashing
contents.

I'd like to hit pause here for a second.  ⏸

Things over here in the libgit2 community have seemed a bit contentious recently, and I'd like to nip this in the bud and make sure that we all get on the same page.

Obviously, it's very difficult to convey things like tone online.  So things that are meant as jokes may not come across that way, and readers may miss the intent and instead hear attacks.  Things that are meant as **emphasis** may be interpreted as rudeness.

So a quick reminder about our [code of conduct](https://github.com/libgit2/libgit2/blob/master/docs/code_of_conduct.md), which reminds us that this aims to be a positive environment, and that we should all be:

* Using welcoming and inclusive language
* Being respectful of differing viewpoints and experiences
* Gracefully accepting constructive criticism
* Focusing on what is best for the community
* Showing empathy towards other community members

And a reminder that our goals are to build a Git repository management library that is first correct and second performant, and that we need to work together to do that.  This is an all-volunteer project, and we're all here because we _want_ to be.  We don't need to argue about degrees of incorrectness, and we don't need comments that can be perceived as rude.  We need collaboration and inclusiveness.

Thanks in advance!

pks-t:
@ethomson: are you good with these changes and opening up the potential for races?

ethomson:
I think so.  For the filesystem-based config system, it's hard for me to imagine that 

But we could add some tuning to the FS-based config system to disable stat in favor of a more heavy handed approach (to minimize that NTFS attribute update lag, or if they're trying to use something worse even than NTFS) or to disable reloading entirely.

But I think that this is good for the defaults.  👍

pks-t:
> But we could add some tuning to the FS-based config system to disable stat in favor of a more heavy handed approach (to minimize that NTFS attribute update lag, or if they're trying to use something worse even than NTFS) or to disable reloading entirely.

Well, the more heavy handed approach is the one we had before, right? :) We could use a global option to switch behaviour here, of course. Did you mean to imply that?

mehrdadn:
> But we could add some tuning to the FS-based config system to disable stat in favor of a more heavy handed approach (to minimize that NTFS attribute update lag, or if they're trying to use something worse even than NTFS) or to disable reloading entirely.

I got confused and didn't really follow what's happening here, but just a note: as far as I know (although I'm not 100% sure about this) there shouldn't be a "lag" per se in NTFS file time updating. Rather, the situation is like this:

> When writing to a file, the last write time is not fully updated until all handles that are used for writing are closed.

So to update the file time I'd expect the only thing to do is to just close all file descriptors used for writing. I _think_ you shouldn't need to change how you actually read the file time (whether it's via `stat` or `FindFirstFile` or something else) and I _think_ there shouldn't be a lag in between. If you're seeing a lag that isn't accounted for in this way, please do send me a repro so I can see if there's a better way to resolve it...

mehrdadn:
And here's a demo that illustrates this:

	#!/usr/bin/env python3

	import os

	def stat_via_dir(filename):
		basename = os.path.basename(filename)
		with os.scandir(os.path.dirname(filename) or '.') as it:
			for entry in it:
				if entry.name == basename:
					return entry.stat()

	def main(program, filename='testfile'):
		print(os.stat(filename).st_mtime)
		with open(filename, 'r+b') as f:
			print(stat_via_dir(filename).st_mtime)
			f.write(b'\n')
			print(os.stat(filename).st_mtime)
			print(stat_via_dir(filename).st_mtime)
		print(os.stat(filename).st_mtime)
		print(stat_via_dir(filename).st_mtime)

	if __name__ == '__main__':
		import sys
		raise SystemExit(main(*sys.argv))

ethomson:
The problem isn't really a lag of updating the filesystem, it's that NTFS only has 100ns resolution for timestamps, so you may not be able to identify a quick update based on timestamp alone.

@pks-t - we could do like the index and instead of _just_ examining timestamp for cache eviction, we could hold on to more bits of data (file size, etc).  This might, if nothing else, help us create tests that aren't racy.

> We could use a global option to switch behaviour here, of course. Did you mean to imply that?

Yeah, I meant to imply that, but I also think that we should get this merged independent of that possibility in the future.

mehrdadn:
I'm confused, if 100ns is the problem then why do you need to sleep for 10ms?

pks-t:
On Tue, Jun 25, 2019 at 10:20:39AM -0700, Edward Thomson wrote:
> @pks-t - we could do like the index and instead of _just_
> examining timestamp for cache eviction, we could hold on to
> more bits of data (file size, etc).  This might, if nothing
> else, help us create tests that aren't racy.

We do that already: `git_futils_filestamp` uses mtime, size and
inode number. The reason why that doesn't matter is that
config::stress:quick_write always re-writes the same key with a
value of the same length.

I could rewrite the test to alwaysappend new config entries, but
do we want to actually check that?

ethomson:
> We do that already: `git_futils_filestamp` uses mtime, size and inode number. The reason why that doesn't matter is that config::stress:quick_write always re-writes the same key with a value of the same length.

:+1:

> I'm confused, if 100ns is the problem then why do you need to sleep for 10ms?

Not sure offhand, it would indeed be worth understanding why a 1ms sleep wasn't sufficient.

pks-t:
On Tue, Jun 25, 2019 at 01:35:01PM -0700, Edward Thomson wrote:
> > I'm confused, if 100ns is the problem then why do you need to
> > sleep for 10ms?
> 
> Not sure offhand, it would indeed be worth understanding why a
> 1ms sleep wasn't sufficient.

Indeed, that confused me, too. But I think I understand it now:
this is about how we update the configuration file when modifying
it.

When we modify the configuration, then we are in fact quite lazy
about how we recompute the updated configuration. Instead of
modifying the in-memory representation of the configuration and
writing out the file, we write out the file, discard our
in-memory representation and then re-parse it. And in fact, we do
this via `config_refresh`, which will now check for mtime, too.
As there is no sleep in between (nor should there be!), we fail
to update our in-memory representation and thus keep on using an
out-of-date one.

I'll come up with a fix.

pks-t:
Updated. I've optimized how we handle refreshes after calling `config_write` so that we do not have to re-read the on-disk file but instead handle this from the in-memory buffer which we had already available anyway. This avoids any timing issues due to my stat cache when writing config files and should thus fix any spurious errors introduced by it. I've thus lowered the sleep to 1ms again, which should now be sufficient

mehrdadn:
Awesome :) is the 1ms sleep due to the 100ns (and the fact that it's hard to go below 1ms)? Or do you need it to be actually 1ms long? Because if you only need the 100ns resolution I could write you a faster sleep...

pks-t:
> Awesome :) is the 1ms sleep due to the 100ns (and the fact that it's hard to go below 1ms)? Or do you need it to be actually 1ms long? Because if you only need the 100ns resolution I could write you a faster sleep...

The first reason was that implementing `cl_msleep` is trivial on both Win32 and Unix by mapping to `Sleep(x)` and `usleep(1000*x)`, respectively. So like this we didn't need much code, especially considering that it's of limited use for our tests, only. And delaying tests by 10 ms in total is probably not really going to be noticeable. Especially not on Windows, which is slow as hell anyway when it comes to test suite execution :(

The second reason was that we also need to consider systems built with "-DUSE_NSEC=NO", where we might not even have that level of granularity for `fstat`. In that case, sleeping for less than a millisecond would possibly cause the stat cache to trip again.

So thanks for your offer, but I don't really think it's necessary for now :)

src/config_file.c
```diff
@@ -284,8 +284,6 @@ static int config_set(git_config_backend *cfg, const char *name, const char *val
 	if ((error = config_write(b, name, key, NULL, esc_value)) < 0)
 		goto out;
 
-	error = config_refresh(cfg);
```
tiennou: Commit message says "cocked" instead of "cached".

src/config_file.c
```diff
@@ -170,47 +170,55 @@ static int config_is_modified(int *modified, git_config_file *file)
...
 	git_mutex_unlock(&b->header.values_mutex);
 
+out:
+	git_config_entries_free(old);
+	return error;
+}
+
+static int config_refresh(git_config_backend *cfg)
+{
+	diskfile_backend *b = (diskfile_backend *)cfg;
```
tiennou: Wouldn't that warrant one of those `GIT_CONTAINER_OF` things, just in case ?

src/config_file.c
```diff
@@ -851,34 +872,27 @@ static int read_on_variable(
...
 	git_config_level_t level,
-	int depth)
+	int depth,
+	const char *buf,
+	size_t buflen)
 {
 	diskfile_parse_state parse_data;
 	git_config_parser reader;
-	git_buf contents = GIT_BUF_INIT;
 	int error;
```
tiennou: This needs to be initialized, as there's an error path uncovering it.
```
/Users/vsts/agent/2.153.2/work/1/s/src/config_file.c:898:6: error: variable 'error' is used uninitialized whenever 'if' condition is true [-Werror,-Wsometimes-uninitialized]
        if (!reader.ctx.content || *reader.ctx.content == '\0')
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/vsts/agent/2.153.2/work/1/s/src/config_file.c:910:9: note: uninitialized use occurs here
        return error;
               ^~~~~
```

pks-t: Yup, fixed!

src/config_file.c
pks-t: Oops, fixed!

src/config_file.c
pks-t: Definitely. Embarassing that I myself didn't think about it :P

pks-t:
Thanks for your review, @tiennou! All comments should be addressed now

src/config_file.c
```diff
@@ -1197,7 +1244,12 @@ static int config_write(diskfile_backend *cfg, const char *orig_key, const char
 		git_buf_attach(&cfg->locked_content, git_buf_detach(&buf), len);
 	} else {
 		git_filebuf_write(&file, git_buf_cstr(&buf), git_buf_len(&buf));
-		result = git_filebuf_commit(&file);
+
+		if ((result = git_filebuf_commit(&file)) < 0)
+			goto done;
+
+		if ((result = config_refresh_from_buffer(&cfg->header.parent, buf.ptr, buf.size)) < 0)
```
tiennou: Build is still red, because Windows fails and/or leaks. I've looked around for the leak and pinpointed one possible occurrence here, but since there's also timing issues at hand, maybe not ?

`buf` was "attached" in the other code path, but not here?

pks-t: Nah, that was just a quick test of mine :P Reverted it again to the previous state, which passes

pks-t:
Thanks for your reviews, @tiennou!

PR #5134, opened 2019-06-21
Config parser separation

pks-t:
This is the missing piece to really separate the configuration parser and the configuration file backend. Goal is to make the code easier to understand as well as to enable easier refactoring. But I guess the result speaks for itself.

Note: this PR includes the two commits from #5132. I wanted to avoid having to fix conflicts, as I expect that #5132 will land quite fast.

### DIFF
--- a/src/config_parse.c
+++ b/src/config_parse.c
@@ -482,7 +482,7 @@ int git_config_parse(
 	git_config_parser_variable_cb on_variable,
 	git_config_parser_comment_cb on_comment,
 	git_config_parser_eof_cb on_eof,
-	void *data)
+	void *payload)
 {
 	git_parse_ctx *ctx;
 	char *current_section = NULL, *var_name = NULL, *var_value = NULL;
@@ -522,7 +522,7 @@ int git_config_parse(
 			git_parse_advance_chars(ctx, result);
 
 			if (on_section)
-				result = on_section(parser, current_section, line_start, line_len, data);
+				result = on_section(parser, current_section, line_start, line_len, payload);
 			/*
 			 * After we've parsed the section header we may not be
 			 * done with the line. If there's still data in there,
@@ -542,13 +542,13 @@ int git_config_parse(
 		case ';':
 		case '#':
 			if (on_comment) {
-				result = on_comment(parser, line_start, line_len, data);
+				result = on_comment(parser, line_start, line_len, payload);
 			}
 			break;
 
 		default: /* assume variable declaration */
 			if ((result = parse_variable(parser, &var_name, &var_value)) == 0 && on_variable) {
-				result = on_variable(parser, current_section, var_name, var_value, line_start, line_len, data);
+				result = on_variable(parser, current_section, var_name, var_value, line_start, line_len, payload);
 				git__free(var_name);
 				git__free(var_value);
 			}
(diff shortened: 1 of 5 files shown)


pks-t:
Disentangled this from #5132

pks-t:
Any comments? I'd like to fast track this as it will cause conflicts with the other config updates I have in flight

src/config_parse.c
```diff
@@ -477,7 +477,7 @@ int git_config_parse(
 	git_config_parser_variable_cb on_variable,
 	git_config_parser_comment_cb on_comment,
 	git_config_parser_eof_cb on_eof,
-	void *data)
+	void *payload)
```
tiennou: 👍

src/config_parse.h
```diff
@@ -22,7 +22,7 @@ typedef struct config_file {
 } git_config_file;
 
 typedef struct {
-	git_config_file *file;
+	const char *configname;
```
tiennou: `file_path` ? I think `configname` dilutes the aspect that it's just a filesystem "token" or something…

pks-t:
> LGTM, minor nitpicks only, though I'm somewhat interested about the includes moving down one level. Isn't that something that somehow "in-memory" would want ?

Do you mean to say that there should be "config_file.h" and "config_mem.h" headers?

src/config_parse.h
pks-t: I explicitly wanted to avoid this. Backends may not use files at all, most importantly the memory backend doesn't care for filenames. I thus wanted to have this be generic

tiennou: Our "outer layer" doesn't really know about anything other than files, and will generally needs a "label" that will usually looks like a file path. It's just about labelling that data, be it file-backed or in-memory. `file_identifier`, `config_label`, or something else ?

Really, it's because `configname` sounds like a "configuration preset name", not like "the user-visible name of the configuring thing".

Feel free to keep it as is though, just wanted to see if some bikeshedding would make it read nicer  😉.

tiennou:
My "includes moving down one level" was about [this)(https://github.com/libgit2/libgit2/pull/5134/files#diff-4a559d4edd82554487be350b248f2668R32): I seem to understand that there would be no way now (if there had ever been) to build a complete in-memory configuration comprised of multiple includes, because that include mechanism was just moved down to the file-backend only part, and/or was never extended to refer to non-file-backed includes ?

pks-t:
> My "includes moving down one level" was about this: I seem to understand that there would be no way now (if there had ever been) to build a complete in-memory configuration comprised of multiple includes, because that include mechanism was just moved down to the file-backend only part, and/or was never extended to refer to non-file-backed includes ?

Oh, that's what you mean by includes! Thought you referred to C header includes.

Yeah, the memory backend is completely unable to handle this. The problem is that in order to compute includes, you need to compute paths relative to the parsed file's location. As the memory-stuff doesn't have any associated path per se, I simply made them not honor includes at all. That being said, I'm not opposed at all to that at a later point.

src/config_parse.h
pks-t: Fair enough. Would you be happy with `config_identifier`?

tiennou:
Thanks for the clarification 👍. I had a hunch it wasn't really supported.

src/config_parse.h
tiennou: `config_identifier` it is then. Thanks!

pks-t: Thinking about it some more. Maybe we _should_ just leave this at `file` or `file_path`. At least it seems like an obvious route to go at a later point when implementing in-memory includes.

pks-t:
Rebased to fix conflicts with #5132, renamed `configname` to `path`

PR #5136, opened 2019-06-23
largefile tests: only write 2GB on 32-bit platforms

ethomson:
Don't try to feed 4 GB of data to APIs that only take a `size_t` on 32-bit platforms.


ethomson:
I'm going to go ahead and merge this to get our nightlies 💚 .

pks-t:
Seems reasonable to me, thanks!

PR #5140, opened 2019-06-23
Re-run flaky tests

ethomson:
Re-run the flaky online integration tests.  Since these hit actual network endpoints and those may be occasionally down, we want to re-run them up to 5 times.  This prevents us from failing the builds just because the badssl endpoint is down, for example.

### DIFF
--- a/ci/test.ps1
+++ b/ci/test.ps1
@@ -29,8 +29,32 @@ function run_test {
 	$TestCommand = (ctest -N -V -R "^$TestName$") -join "`n" -replace "(?ms).*\n^[0-9]*: Test command: ","" -replace "\n.*",""
 	$TestCommand += " -r${BuildDir}\results_${TestName}.xml"
 
-	Invoke-Expression $TestCommand
-	if ($LastExitCode -ne 0) { $global:Success = $false }
+	if ($Env:GITTEST_FLAKY_RETRY -gt 0) {
+		$AttemptsRemain = $Env:GITTEST_FLAKY_RETRY
+	} else {
+		$AttemptsRemain = 1
+	}
+
+	$Failed = 0
+	while ($AttemptsRemain -ne 0) {
+		if ($Failed -eq 1) {
+			Write-Host ""
+			Write-Host "Re-running flaky $TestName tests..."
+			Write-Host ""
+		}
+
+		Invoke-Expression $TestCommand
+		if ($LastExitCode -eq 0) {
+			$Failed = 0
+			break
+		} else {
+			$Failed = 1
+		}
+
+		$AttemptsRemain = $AttemptsRemain - 1
+	}
+
+	if ($Failed -eq 1) { $global:Success = $false }
 }
 
 Write-Host "##############################################################################"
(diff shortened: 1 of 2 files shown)


ci/test.ps1
```diff
@@ -29,8 +29,32 @@ function run_test {
 	$TestCommand = (ctest -N -V -R "^$TestName$") -join "`n" -replace "(?ms).*\n^[0-9]*: Test command: ","" -replace "\n.*",""
 	$TestCommand += " -r${BuildDir}\results_${TestName}.xml"
 
-	Invoke-Expression $TestCommand
-	if ($LastExitCode -ne 0) { $global:Success = $false }
```
pks-t: One thing I wondered about recently: do we really need the Powershell file at all? I saw that Azure does support Bash on Windows via Git for Windows. If that's the case, then we should drop this file completely so that we don't always have to keep both in sync. I'd naturally volunteer to do that.

ethomson: That's true, it does.  I have no love for PowerShell...  I'm happy to do that, too, but I've got a few other things going on before I can look at it.

ethomson:
> I'm generally 👍 on this, feels like it is long overdue. I've been wondering about whether this is the right layer to fix this, though. Wouldn't it make more sense to have this in the test code itself? Like that, developers would benefit from it, too, and we'd be able to mark single tests as flaky instead of having to re-do all tests of a particular test suite.

Right.  This was my first thought as well, actually.  But then I realized it would require more thought than I wanted to give it.  😀 

I think there are two issues here:

1. We need a way to decorate tests as flaky.  This could be as simple as `/* flaky */` at the end of a declaration...  in fact, that's probably the smart way to do this as far as clar's parsing goes.

(But then I started yakshaving, because I wish clar's parser was smart enough to deal with commented out tests.  So then I started thinking more about using llvm or something, and then things got _really_ out of control.)

We could also mark a suite as flaky on the command-line, which might actually be smarter still.  eg, `-fonline`.

2. All tests would need to get tightened up (or we'd fail them for memory leaks).  This is not impossible by any means, but it is an added effort to go through and identify them.  But if we do the whole-suite marking on the command line then we'd need to do a bunch en masse.

🤷‍♂ 

So I decided to knock this out quickly so that I'd be less angry.  😉

ci/test.sh
```diff
@@ -60,7 +60,32 @@ run_test() {
...
+		if [ "$FAILED" -eq 0 ]; then
+			break
+		fi
+
+		ATTEMPTS_REMAIN="$(($ATTEMPTS_REMAIN-1))"
+	done
+
+	if [ "$FAILED" -ne 0 ]; then
+		SUCCESS=0
+	fi
```
pks-t: Previously we were executing `failure`, don't we have to do that now, too?

pks-t:
Fair enough ;) So let's get this merged quickly -- we can still improve in the future.

ci/test.sh
ethomson: Ach, that ended up dead.  I removed it, but I also improved the error message a little, inspired by that function.

PR #5141, opened 2019-06-24
azure: drop powershell

pks-t:
Quick and dirty PR to drop all our Powershell scripts and use Bash across all platforms. I bet it won't succeed yet, but I've been a bit lazy to fire up a Windows host :P


ethomson:
Oh, jeez.  I think that the MSVC builds are failing because they're running inside Git for Windows (aka msys aka mingw) and so it thinks that you want to build them _with_ mingw.

And I think that the mingw builds are failing because they don't understand Windows-style paths (eg `C:\foo`) and you need to convert them to `/c/foo`.  There's an app for this that ships with cygwin but ughhhh.

pks-t:
On Mon, Jun 24, 2019 at 11:42:02AM -0700, Edward Thomson wrote:
> Oh, jeez.  I think that the MSVC builds are failing because
> they're running inside Git for Windows (aka msys aka mingw) and
> so it thinks that you want to build them _with_ mingw.
> 
> And I think that the mingw builds are failing because they
> don't understand Windows-style paths (eg `C:\foo`) and you need
> to convert them to `/c/foo`.  There's an app for this that
> ships with cygwin but ughhhh.

Yeah, I expected to have some problems with CMake as I've noticed
in the past that it can be quite weird to use it in Cygwin, too,
when not building with Makefiles. Should've marked this PR as
WIP.

ethomson:
> Should've marked this PR as WIP.

You were clear in setting expectations.  I just wanted to 👀 to see how bad it was.  😀 

I'm heads down on a bunch of things right now but I'm happy to help look at these when I have some time.  (ETA for that unknown.)

pks-t:
Cherry-picked fixes from #5143 as this PR here implicitly enables Werror.

pks-t:
Jeez, this is quickly ballooning in size due to the ENABLE_WERROR stuff. @ethomson, please feel free to either merge this PR in favour of yours or to cherry-pick any fixes.

pks-t:
As this PR balloons up in size, builds are getting redder and redder. I just don't get how compilation could've passed with Werror in some runs on Windows, but then out of nowhere there's new warnings popping up. I mean... why didn't the compiler notice these the first time it ran?

Maybe I should split this up. Most commits are just mind-numbing handholding for the compiler so that it stops crying and thus trivial to review, though. Don't mistake me: this is indeed a very good thing, but it's really a pain to do.

pks-t:
Linux and macOS are green again. Windows is failing due to memory leaks in our fuzzers. MinGW is failing as it cannot find a bash shell.

pks-t:
Rebased on latest master now that the Werror and fuzzer portability PRs have landed

pks-t:
I really ain't got no clue on how to fix the MinGW builds. If putting Git for Windows into PATH, then CMake complains about sh.exe being found which is not supported. If removing GfW from PATH, then Azure complains about not being able to locate the bash executable. @ethomson, you've got any tip?

ethomson:
Sure, I can take a look tomorrow.

pks-t:
@ethomson: finally got this solved. The fix was to not set PATH at all for our jobs, but instead to only adjust PATH for the CMake invocations.

pks-t:
Rebased to fix conflicts

pks-t:
Thanks a lot for your review, @ethomson! :)

PR #5143, opened 2019-06-25
ci: build with ENABLE_WERROR on Windows

ethomson:
Build with -Werror's equivalent (/WX) on MSVC.

I neglected to do this in the previous warnings PR, oops.  I think that there are also some lingering issues in things like examples that I didn't build locally with `/WX` and missed since I didn't add this in the CI setup.


pks-t:
Hope you don't mind me pushing this forward! :)

ethomson:
ON THE CONTRARY.  🎉🎉🎉🎉🎉

PR #5149, opened 2019-06-28
repository: fix left-over file for symlink check

### LINKED ISSUE #5147: Create repository has symlink left on Windows
### Reproduction steps
1. Windows 10 1809, enable developer mode so that symlink can be created without UAC elevation.
2. Install git for Windows 2.22 and enable symlinks during setup. Therefore `core.symlinks true`
3. Create a directory such as `C:\123`, this will be used as repository root.
4. Run following code
```C
git_repository *repo = NULL;
git_repository_init_options options = GIT_REPOSITORY_INIT_OPTIONS_INIT;
options.flags = GIT_REPOSITORY_INIT_MKPATH | GIT_REPOSITORY_INIT_EXTERNAL_TEMPLATE
git_repository_init_ext(&repo, "C:\\123", &options);
```

`.git` directory is created and initialized.
However, a temporary `.git2_abcdef` symlink is created in the working directory but not automatically deleted.

### Expected behavior
symlink automatically deleted by libgit2.

### Actual behavior
symlink not deleted.

v0.28.0-232-ge50d138

Windows 10 Pro 1809

### Cause
In repository.c `are_symlinks_supported()`,
1. Create a temp file `.git2_abcdef` where `abcdef` is random
2. Delete that temp file
3. Create a symlink `.git2_abcdef` that points to `testing`. However the file `testing` does not exist.
If target does not exist, `CreateSymbolicLinkW()` will create the symlink as directory type, even do not specify `SYMBOLIC_LINK_FLAG_DIRECTORY`
4. Check if `.git2_abcdef` is a symlink
5. Delete `.git2_abcdef` using `DeleteFileW()`. If symlink type becomes directory, `DeleteFileW()` does not work. Here is the problem why the symlink is left!!!

### Solutions
A. To test symlink, point to a file that exists, instead of that `testing`.
Create 2 temporary files, one is symlink that points to the other file. Then delete these 2 files.
B. Use `RemoveDirectoryW()` as fallback

pks-t:
To determine whether the host we're running on supports symlinks, we
try to create a temporary symlink in the repo's working directory. If
creation succeeded, then symlinks are supported. This works just fine on
non-Windows hosts, but on Windows hosts it does leave behind the test
file without deleting it correctly.

The reason for this is that on Windows, one has to delete symlinks to
files with `DeleteFile`, but symlinks to directories with
`RemoveDirectory`. As we are creating the symlink with a non-existing
target of "testing", the symlink will be a directory symlink, but we
still use `p_unlink` which will map to `DeleteFile` and thus deletion
fails. Note: in the case where the target repo has a file "testing"
inside of its working directory, then we'd actually succeed to unlink
the file.

The issue can be fixed by creating another file and link to that one.
One might think of easier alternatives, e.g. linking to ".git" and then
just using `p_rmdir` on Windows. But this would fall flat for e.g.
submodules, which have a ".git" file, and for cases where the workdir
and gitdir live in different places.

Furthermore it's debatable whether `p_unlink` should emulate unlink(3P)
more closely. But this would require us to check what kind of file we
want to unlink, which could potentially have a non-negligible impact on
performance.

---

Fixes #5147

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -1421,7 +1421,8 @@ static bool are_symlinks_supported(const char *wd_path)
 	git_buf xdg_buf = GIT_BUF_INIT;
 	git_buf system_buf = GIT_BUF_INIT;
 	git_buf programdata_buf = GIT_BUF_INIT;
-	git_buf path = GIT_BUF_INIT;
+	git_buf filepath = GIT_BUF_INIT;
+	git_buf linkpath = GIT_BUF_INIT;
 	int fd;
 	struct stat st;
 	int symlinks = 0;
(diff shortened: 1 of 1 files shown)


pks-t:
Closing in favour of #5151

Issue #5127, opened 2019-06-19
Bug in win32_find_git_in_path()


mehrdadn:
[The following lines in `win32_find_git_in_path`](/libgit2/libgit2/blob/635693d3bc55770ec7a6640ba3f2f0ee434a6042/src/win32/findfile.c#L95) are buggy:

			/* replace "bin\\" or "cmd\\" with subdir */
			wcscpy(&root.path[root.len - 4], subdir);

No check is performed to ensure that the path in fact ends with `"bin\\"` or `"cmd\\"`, and indeed my `git` is not located in such a path. This results in chopping off a directory name in the middle and produces a completely invalid path.

It may be more reasonable to do nothing if the path suffix isn't recognized... or alternatively, to chop off the last path component rather than the last 4 characters.

csware:
Please provide your path structure, especially where to find `etc/gitconfig` and `usr/share/` relative to your `git.exe`.

mehrdadn:
I have my own custom directory layout; these files don't even exist on my builds. That's why I didn't provide any. But I wasn't asking for explicit support of a different path structure. I was just pointing out there shouldn't be a buffer overrun in the program.

csware:
Ok, then generally the whole detection code should be skipped as it is just used to find these directories...

mehrdadn:
Yeah I don't know the best solution -- just know it shouldn't be a buffer overflow!

csware:
There is no buffer overflow or underflow happening.

mehrdadn:
Buffer chopping, whatever you want to call it -- you get what I mean.

csware:
The code also seems to require an update for newer Git for Windows versions. I'll have a look...

ethomson:
I believe this was fixed via #5150

PR #5150, opened 2019-06-28

csware:
1. Don't mangle paths for non Git for Windows installations (cf. #5127)
2. Git for Windows also has a Win64 package
2. Support Git for Windows >= 2.0, the Git root directory is not in `PROGRAMFILES\\Git` anymore, but under another folder `mingw64` or `mingw32`.
3. Prevent possible buffer overflow

### DIFF
--- a/src/win32/findfile.c
+++ b/src/win32/findfile.c
@@ -91,9 +91,17 @@ static int win32_find_git_in_path(git_buf *buf, const wchar_t *gitexe, const wch
 			continue;
 		wcscpy(&root.path[root.len], gitexe);
 
-		if (_waccess(root.path, F_OK) == 0 && root.len > 5 && (root.len - 4 + wcslen(subdir) < MAX_PATH)) {
-			/* replace "bin\\" or "cmd\\" with subdir */
-			wcscpy(&root.path[root.len - 4], subdir);
+		if (!_waccess(root.path, F_OK)) {
+			/* replace "bin\\" or "cmd\\" of a Git for Windows installation with subdir OR append path */
+			if (root.len > 5 && wcscmp(root.path - 4, L"cmd\\") || wcscmp(root.path - 4, L"bin\\")) {
+				if (root.len - 4 + wcslen(subdir) >= MAX_PATH)
+					continue;
+				wcscpy(&root.path[root.len - 4], subdir);
+			} else {
+				if (root.len + wcslen(subdir) >= MAX_PATH)
+					continue;
+				wcscat(root.path, subdir);
+			}
 
 			win32_path_to_8(buf, root.path);
 			return 0;
--- a/src/win32/findfile.c
+++ b/src/win32/findfile.c
@@ -68,6 +68,42 @@ static wchar_t* win32_walkpath(wchar_t *path, wchar_t *buf, size_t buflen)
 	return (path != base) ? path : NULL;
 }
 
+static int win32_find_git_for_windows_architecture_root(_findfile_path *root_path, const wchar_t *subdir)
+{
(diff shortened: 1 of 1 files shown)


src/win32/findfile.c
```diff
@@ -91,7 +91,7 @@ static int win32_find_git_in_path(git_buf *buf, const wchar_t *gitexe, const wch
 			continue;
 		wcscpy(&root.path[root.len], gitexe);
 
-		if (_waccess(root.path, F_OK) == 0 && root.len > 5) {
+		if (_waccess(root.path, F_OK) == 0 && root.len > 5 && (root.len - 4 + wcslen(subdir) < MAX_PATH)) {
```
pks-t: Yup. We'd have avoided this issue if we simply used a `git_buf` instead, but this fix looks good to me

src/win32/findfile.c
```diff
@@ -189,6 +226,12 @@ int git_win32__find_system_dirs(git_buf *out, const wchar_t *subdir)
 			&buf, HKEY_CURRENT_USER, REG_MSYSGIT_INSTALL_LOCAL, subdir) && buf.size)
 		git_buf_join(out, GIT_PATH_LIST_SEPARATOR, out->ptr, buf.ptr);
 
+#ifdef WIN64
+	if (!win32_find_git_in_registry(
+			&buf, HKEY_LOCAL_MACHINE, REG_MSYSGIT_INSTALL_LOCAL, subdir) && buf.size)
+		git_buf_join(out, GIT_PATH_LIST_SEPARATOR, out->ptr, buf.ptr);
```
pks-t: Don't these directories exist on x86, too? At least you check them in `win32_find_git_in_path` unconditionally, too.

src/win32/findfile.c
```diff
@@ -68,6 +68,42 @@ static wchar_t* win32_walkpath(wchar_t *path, wchar_t *buf, size_t buflen)
...
+static int win32_find_git_for_windows_architecture_root(_findfile_path *root_path, const wchar_t *subdir)
+{
+	/* Git for Windows >= 2 comes with a special architecture root (mingw64 and mingw32)
+	 * under which "etc" and the "share" folders are located, check which we need (none is also ok) */
+
+	static const wchar_t *architecture_roots[3] = {
+		L"mingw64\\",
+		L"mingw32\\",
+		NULL,
+	};
```
pks-t: Using `ARRAY_SIZE` instead of a sentinel value would've been a bit easier to read, but that might be my personal taste, only.

src/win32/findfile.c
csware: There's a suble difference... See the definition of `REG_MSYSGIT_INSTALL`.

csware: Maybe I should make this a bit more clear...

src/win32/findfile.c
csware: I tried to stick to the coding style in that file.

src/win32/findfile.c
csware: Please not, this is wchar_t not char.

src/win32/findfile.c
csware: On x64 systems the following `win32_find_git_in_registry` call uses the `Wow6432Node` redirect (i.e., the x86 version of Git for Windows....

pks-t:
@ethomson: do you have the time to review this soon-ish?

csware:
@pks-t @ethomson Any news on this?

ethomson:
This was (unintentionally) closed as part of renaming the master branch to main.

Reopened as #5773

PR #5151, opened 2019-06-29
w32: fix unlinking of directory symlinks

### LINKED ISSUE #5147: Create repository has symlink left on Windows

pks-t:
This implements unlinking for directory symlinks on Win32. I've been annoyed by our misnamed "fileops.h" while at it and just renamed it to "futils.h", as it should be named. I hope this doesn't generate too many conflicts, otherwise I'll boot it out of this PR.

Supersedes #5149, fixes #5147

### DIFF
--- a/fuzzers/standalone_driver.c
+++ b/fuzzers/standalone_driver.c
@@ -8,7 +8,7 @@
 #include <stdio.h>
 
 #include "git2.h"
-#include "fileops.h"
+#include "futils.h"
 #include "path.h"
 
 extern int LLVMFuzzerTestOneInput(const unsigned char *data, size_t size);
--- a/src/apply.c
+++ b/src/apply.c
@@ -18,7 +18,7 @@
 #include "git2/repository.h"
 #include "array.h"
 #include "patch.h"
-#include "fileops.h"
+#include "futils.h"
 #include "delta.h"
 #include "zstream.h"
 #include "reader.h"
--- a/src/attr_file.h
+++ b/src/attr_file.h
@@ -14,7 +14,7 @@
 #include "vector.h"
 #include "pool.h"
 #include "buffer.h"
-#include "fileops.h"
+#include "futils.h"
 
 #define GIT_ATTR_FILE			".gitattributes"
 #define GIT_ATTR_FILE_INREPO	"attributes"
--- a/src/blob.h
+++ b/src/blob.h
@@ -12,7 +12,7 @@
 #include "git2/blob.h"
 #include "repository.h"
 #include "odb.h"
-#include "fileops.h"
+#include "futils.h"
 
 struct git_blob {
 	git_object object;
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -1893,11 +1893,18 @@ static int checkout_create_the_new(
 				return error;
 		}
 
-		if (actions[i] & CHECKOUT_ACTION__UPDATE_BLOB) {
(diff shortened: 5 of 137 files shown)


pks-t:
I've force pushed fixes to more errors around symlink handling on Win32:

- When creating symlinks with a relative target, we didn't correctly determine the type of the target file as we checked for the target relative to the current directory instead of relative to the symlink directory. Such symlinks were then usually created as directory symlinks.
- When performing checkouts, we have to postpone handling of symlinks to the end. Otherwise, we may create symlinks to blobs which haven't yet been checked out, which would cause us to create the symlinks as directory symlinks instead of as file symlinks.

ethomson:
Should we just skip the attribute check entirely, and try to delete the file, and if that fails, try to delete the directory?

pks-t:
On Sat, Jul 20, 2019 at 03:19:36AM -0700, Edward Thomson wrote:
> Should we just skip the attribute check entirely, and try to
> delete the file, and if that fails, try to delete the
> directory?

I think we shouldn't. unlink(3P) doesn't ever unlink directories.
So if somebody called `unlink` on a non-symlink directory, then
we should fail. If we didn't perform the attribute check, then
we'd erroneously remove that directory.

pks-t:
Rebased to fix conflicts

ethomson:
> I think we shouldn't. unlink(3P) doesn't ever unlink directories.  So if somebody called `unlink` on a non-symlink directory, then we should fail. If we didn't perform the attribute check, then we'd erroneously remove that directory.

Aha, yes, that's very much true.  I was thinking that this was a unlink-symlink path, but obviously it's an unlink-whatever-was-passed-to-it function.

ethomson:
Thanks!

PR #5152, opened 2019-06-29
Fix Regression: attr: Correctly load system attr file (on Windows)

csware:
Regression introduced in commit 5452e49fce21f726bec19519da7f012e3f19e736 on PR #4967.

I don't know how to provide a test for this.

### DIFF
--- a/src/attr.c
+++ b/src/attr.c
@@ -333,6 +333,7 @@ static int attr_setup(git_repository *repo, git_attr_session *attr_session)
 				       NULL, git_repository_attr_cache(repo)->cfg_attr_file)) < 0)
 		goto out;
 
+	git_buf_clear(&path); /* git_repository_item_path expects an empty buffer, because it uses git_buf_set */
 	if ((error = git_repository_item_path(&path, repo, GIT_REPOSITORY_ITEM_INFO)) < 0 ||
 	    (error = preload_attr_file(repo, attr_session, GIT_ATTR_FILE__FROM_FILE,
 				       path.ptr, GIT_ATTR_FILE_INREPO)) < 0) {
--- a/src/attr_file.c
+++ b/src/attr_file.c
@@ -919,6 +919,7 @@ int git_attr_session__init(git_attr_session *session, git_repository *repo)
 {
 	assert(repo);
 
+	memset(session, 0, sizeof(*session));
 	session->key = git_atomic_inc(&repo->attr_session_key);
 
 	return 0;
--- a/tests/attr/repo.c
+++ b/tests/attr/repo.c
@@ -2,6 +2,7 @@
 #include "fileops.h"
 #include "git2/attr.h"
 #include "attr.h"
+#include "sysdir.h"
 
 #include "attr_expect.h"
 #include "git2/sys/repository.h"
@@ -17,6 +18,7 @@ void test_attr_repo__cleanup(void)
 {
 	cl_git_sandbox_cleanup();
 	g_repo = NULL;
+	cl_sandbox_set_search_path_defaults();
 }
 
 static struct attr_expected get_one_test_cases[] = {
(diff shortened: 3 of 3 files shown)


csware:
An alternative form would be:

	if (((error = system_attr_file(&path, attr_session)) < 0 && error != GIT_ENOTFOUND) ||
		 (error == 0 && (error = preload_attr_file(repo, attr_session, GIT_ATTR_FILE__FROM_FILE,
							   NULL, path.ptr)) != GIT_ENOTFOUND))
		goto out;

but I doubt that another version is better readable...

The problem is that `preload_attr_file` should only be called if `system_attr_file` returns 0 and that `goto out` should also be called when `preload_attr_file` returns 0 or when `error < 0 && error != GIT_ENOTFOUND` ,

csware:
@pks-t @ethomson Could you please review (and merge)?

src/attr.c
```diff
@@ -322,12 +322,12 @@ static int attr_setup(git_repository *repo, git_attr_session *attr_session)
...
-				       NULL, path.ptr)) < 0) {
-		if (error != GIT_ENOTFOUND)
-			goto out;
-	}
+	error = system_attr_file(&path, attr_session);
+	if (error == 0)
+		error = preload_attr_file(repo, attr_session, GIT_ATTR_FILE__FROM_FILE,
+					  NULL, path.ptr);
+	if (error != GIT_ENOTFOUND)
+		goto out;
```
pks-t: Yup, this indeed did change behaviour. Previous to my commit 5452e49fce21f726bec19519da7f012e3f19e736, we would've returned if ``preload_attr_file` returned any value except `GIT_ENOTFOUND`. Now we only return if there was an error, but will not return if we were successful. 

But was previous behaviour really the correct thing to do? The intention of `attr_setup` is to pre-cache all attribute files that are relevant to the current repository. So we want to load the global attributes file, the repo attributes file as well as attributes files from the working directory and index. Previously, we would've returned as soon as we have successfully preloaded the system attributes file without preloading any of the others.

You say that the fix here is that `preload_attr_file` should only be called if `error == 0`. But this is indeed not true: `system_attr_file` may only return either a negative value or 0, there is no codepath that returns positive. So with my refactorings, `preload_attr_file` is _still_ called if and only if `error == 0`. So the real change you do here is the return on success.

So I'm not yet convinced that the old behaviour is the correct thing to do. Could you please provide a reproducer with the changing behaviour you have encountered? Just want to make up my own mind whether your proposed fix is the correct thing to do or if you were previously relying on libgit2's faulty behaviour.

csware: I already posted a case where this failed.

csware: Considering to change a behavior is fine, but it should not be called "Refactoring", but it should also not break things.

csware: Btw in TortoiseGit I also have tests which rely on a working .gitattributes files (e.g., with a filter) and this is working correctly with libgit2 even with the early exit here.

csware: In my case w/o the early exit `attr_setup` fails here: https://github.com/libgit2/libgit2/blob/master/src/attr.c#L340 and returns -1

csware: I just found the reason: `git_buf_sets` fails in `git_repository_item_path` if the buffer is not empty. Comming up with a new PR if my tests run w/o errors now.

csware: @pks-t Commit pushed.

csware: Btw. I just checked the coude again and the early exit was fine, because the "other" files are loaded *again* after the call to `attr_setup` on line https://github.com/libgit2/libgit2/blob/master/src/attr.c#L515.

pks-t: > I already posted a case where this failed: #4967 (comment)

Oops, sorry, didn't see that one.

> Considering to change a behavior is fine, but it should not be called "Refactoring" and it should also not break things.

I fully agree with you. The change in behaviour definitely wasn't intended, even though it _seems_ to me like the right thing to do in hindsight. But as said, I don't yet know if I'm correct about that.

> I just found the reason: git_buf_sets fails in git_repository_item_path if the buffer is not empty. Comming up with a new PR if my tests run w/o errors now.

Ah, interesting.

> Btw. I just checked the coude again and the early exit was fine, because the "other" files are loaded again after the call to attr_setup on line https://github.com/libgit2/libgit2/blob/master/src/attr.c#L515.

That's... weird. The current code could be improved by quite a lot, if you ask me. Either we do the loading in `attr_setup`, or if we load the files anyway after `attr_setup` then we should just inline the one file we were _not_ loading in `attr_setup`, which is the global one. After all, `attr_setup` is used in this one location, only.

I'll take a look at your changes and try to dig into that code a bit more.

src/attr.c
```diff
@@ -333,6 +333,7 @@ static int attr_setup(git_repository *repo, git_attr_session *attr_session)
 				       NULL, git_repository_attr_cache(repo)->cfg_attr_file)) < 0)
 		goto out;
 
+	git_buf_clear(&path); /* git_repository_item_path expects an empty buffer, becasue it uses git_buf_set */
```
pks-t: Yup, this fix definitely looks correct to me. Previously, we would've usually never encountered this code path due to the wrong early exit.

Anyway. I'll have to investigate why we have so much duplicated code in `attr_setup` and `collect_attr_files`.

csware: First of all I'd like to have this regression fixed.

pks-t: Cannot argue with that. Does your proposed change fix the issues you're seeing? If so, then we may just merge this quickly and be done with it. I'm currently writing a testcase that I'd like to include here, shouldn't take me too long.

Anyway, why we have two passes: we need to preload gitattributes files as they may contain macros that are used in other files. So the first pass is only to make macros available, the second pass is to re-parse files to make use of macros. But in fact, we do this the wrong way. gitattributes(5) specifies that macros may only be parsed from either global files, the repo's file in ".git/info/attributes" or the top-level working directory only. It explicitly says that macros defined in subdirectories shall _not_ be honored, but we fail to respect this. I'll create a PR to fix this.

pks-t:
I've added two tests to this PR. attr::repo::sysdir_with_session would've failed without your fix.

pks-t:
Will merge if you give me green light @csware

csware:
Gogogo! ;)

pks-t:
Thanks a lot, @csware!

Issue #2663, opened 2014-10-26
Cannot use IPv6 address for remote operations with WIN_HTTP


csware:
Try clone `http://[1:2::3]/git/repo.git`, you'll get invalid URL error.

The reason for this is, that `t->connection_data.host` contains the RAW-IP-address as a string (`1:2::3` in this example). However, WinHttpConnect requires the raw IPv6-addresses within square brackets: `[1:2::3]`.

A possible solution could be to use `WinHttpCrackUrl` for parsing the URL... If we agree on a possible solution, I can create a PR for this.

csware:
Fixed by fe41e5823986aff01054b7db4b869b2896ea4c30 resp. PR #5741.

PR #5155, opened 2019-07-02
Fix parsing of IPv6 addresses on Windows PCs

wolfgang-paul:
Hello,
We want to use libgit2 on Windows 10 based PCs for managing repositories on PCs and embedded devices. The embedded devices are accessible only via IPv6 addresses, as they only have unique IPv6 addresses but not unique IPv4 ones. Unfortunately, currently it's not possible to use libgit2 on Windows based PCs together with IPv6 addresses.
The purpose of the changes is to keep the opening and closing brackets of IPv6 addresses as part of the URL on Windows based PCs. In order to keep libgit2's behaviour on non Windows PCs as it is, the changes are only valid if WIN32 is defined.
Additional tests check libgits2's URL parser together with IPv4 and IPv6 addresses. All network::urlparse tests run successfully both if WIN32 is defined and if not. I also added IPv4 and IPv6 tests based on the specific tests inside the "maint/v0.28" branch (libgit2 v0.28.2) in my fork together with the changes inside http_parser.c and they also work fine.

### DIFF
--- a/deps/http-parser/http_parser.c
+++ b/deps/http-parser/http_parser.c
@@ -2015,7 +2015,13 @@ http_parse_host(const char * buf, struct http_parser_url *u, int found_at) {
         break;
 
       case s_http_host_v6:
+#ifdef WIN32
+      case s_http_host_v6_start:
+      case s_http_host_v6_end:
+        if ((s != s_http_host_v6) && (s != s_http_host_v6_start) && (s != s_http_host_v6_end)) {
+#else
         if (s != s_http_host_v6) {
+#endif
           u->field_data[UF_HOST].off = (uint16_t)(p - buf);
         }
         u->field_data[UF_HOST].len++;
--- a/tests/network/urlparse.c
+++ b/tests/network/urlparse.c
@@ -673,6 +673,6 @@ void test_network_urlparse__user_pass_port_ipv6(void)
 
 void test_network_urlparse__fails_ipv6(void)
 {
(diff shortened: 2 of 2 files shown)


ethomson:
Hi!  Thanks for the pull request (and especially for the tests), and sorry that we're deficient with IPv6 on Windows.

Can you help me understand this a little bit better?  My quick read makes me think that the Windows networking stack expects `[` and `]` bounded addresses.  eg, the goal is to preserve them in the `[xxx]`...

Another option here would be to re-insert the `[` and `]` when using the Windows networking stack.  This might be preferable for a few reasons:  first, http-parser is an external dependency, and deviating from upstream makes snapping back to their changes more difficult.  Second, defining a canonical format (whether that's with or without `[` / `]`) might be nice, since libgit2 end-users can plug in their own transport.  It would be strange if somebody was trying to use their own HTTP stack and we gave them `[` / `]` bounded addresses on Windows and not on Unix.  I think that using a single format for consistency would be more usable for them.

csware:
Might be related to #2663

wolfgang-paul:
Hello @ethomson,
Thank you for your quick response. We totally agree with you to keep the parser(s) used by libgit2 operating system independent and therefore not to insert specific characters OS dependent inside http-parser.
We will think about what we can do next to solve our problem with IPv6 addresses.
Wolfgang

ethomson:
Closed with #5741

PR #5158, opened 2019-07-05
patch_parse: do not depend on parsed buffer's lifetime

pks-t:
When parsing a patch from a buffer, we let the patch lines point into
the original buffer. While this is efficient use of resources, this also
ties the lifetime of the parsed patch to the parsed buffer. As this
behaviour is not documented anywhere in our API it is very surprising to
its users.

Untie the lifetime by duplicating the lines into the parsed patch. Add a
test that verifies that lifetimes are indeed independent of each other.

---

Noticed while testing #5125

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -588,8 +588,8 @@ static int parse_hunk_body(
 
 		memset(line, 0x0, sizeof(git_diff_line));
 
-		line->content = ctx->parse_ctx.line + prefix;
 		line->content_len = ctx->parse_ctx.line_len - prefix;
+		line->content = git__strndup(ctx->parse_ctx.line + prefix, line->content_len);
 		line->content_offset = ctx->parse_ctx.content_len - ctx->parse_ctx.remain_len;
 		line->origin = origin;
 		line->num_lines = 1;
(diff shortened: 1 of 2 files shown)


ethomson:
Hmm, so a very quick skim indicates that `git_patch_from_buffer` is only exposed for tests so zooming out a little bit, does this break every dependency that patches have on other objects?  Or are we now copying things out of (say) `git_blob`s but still dependent on them for other data?

ethomson:
> Hmm, so a very quick skim ...

Hmm, after a slightly less quick skim, I think that my very quick skim was too quick for me to understand the lifecycle of the buffer and usage.  I'm going to have to 👀 this over the weekend to grok it in fullness.

pks-t:
You can definitely trigger this error by using our public APIs, only. See the "git diff --no-index" example in #5125, which does trigger this error by free'ing the file contents before printing the patch.

eaigner:
I think I've just encountered this myself after parsing a patch

BEFORE

```
diff --git a/test/diff2.txt b/test/diff2.txt
index 212a1d6..0dff799 100644
--- a/test/diff2.txt
+++ b/test/diff2.txt
@@ -1,3 +1,2 @@
 This is just
-a text file to
 changes
\ No newline at end of file
```

AFTER (contains garbage at 1st diff line)

```
diff --git a/test/diff2.txt b/test/diff2.txt
index 212a1d6..0dff799 100644
--- a/test/diff2.txt
+++ b/test/diff2.txt
@@ -1,3 +1,2 @@
 n_ATis just
-a text file to
 changes
```

pks-t:
Yeah, definitely looks like the same issue. It's really easy to trigger, unfortunately

Issue #5153, opened 2019-06-29
diff from buf from patch fails if any buffer has no newline at e


albfan:
### Reproduction steps

Using the example in #5125 

```
printf 'Hello' > file1
echo Hello > file2
./examples/lg2 diff --no-index file1 file2
diff from patch [-1] - invalid patch hunk at line 7
```
while git works correctly

```
git diff --no-index file1 file2
diff --git 1/file1 2/file2
index 5ab2f8a43..e965047ad 100644
--- 1/file1
+++ 2/file2
@@ -1 +1 @@
-Hello
\ No newline at end of file
+Hello
```

The problem is on https://github.com/libgit2/libgit2/blob/master/src/patch_parse.c#L582 and how to handle lines starting with `\`

### Expected behavior

Output:

```
diff --git 1/file1 2/file2
index 5ab2f8a43..e965047ad 100644
--- 1/file1
+++ 2/file2
@@ -1 +1 @@
-Hello
\ No newline at end of file
+Hello
```

### Actual behavior

```
diff from patch [-1] - invalid patch hunk at line 7
```

### Version of libgit2 (release number or SHA1)

master

### Operating system(s) tested

linux (arch linux)

Related with #5125

albfan:
@pks-t We can dive into this if you want too

```
\ No newline at end of file
```

Are like context lines

albfan:
@neithernut Anyone to ping to fix this? I can implement myself if that would be accepted

pks-t:
@albfan: a fix is already available in #5159

ethomson:
Fixed via #5159

PR #5159, opened 2019-07-05
patch_parse: handle missing newline indicator in old file

### LINKED ISSUE #5157: diff: git_diff_from_buffer fails to parse diff with comment lines
Consider this diff produced by `git` after adding a newline at the end

```
diff --git a/test/diff2.txt b/test/diff2.txt
index 212a1d6d8..b27c6ecb3 100644
--- a/test/diff2.txt
+++ b/test/diff2.txt
@@ -1,4 +1,4 @@
 This is just
 a text file to
 test git diff file
-changes
\ No newline at end of file
+changes
```

feeding this to `git_diff_from_buffer` will result in an error

     invalid patch hunk at line 10

Looking at `patch_parse.c:parse_hunk_body` it seems that comment lines like

    \ No newline at end of file

are completely ignored and the default error is thrown

pks-t:
When either the old or new file contents have no newline at the end of
the file, then git-diff(1) will print out a "\ No newline at end of
file" indicator. While we do correctly handle this in the case where the
new file has this indcator, we fail to parse patches where the old file
is missing a newline at EOF.

Fix this bug by handling and missing newline indicators in the old file.
Add tests to verify that we can parse such files.

---

I've tried to modify our patch printing logic to also emit the "No newline at end of file" markers, but it turned out to be non-trivial. Many of the diff/patch structs are part of our public API and thus cannot be easily changed, so I didn't quite know where to put the missing newline indicators. So I've stopped short of actually verifying that we round-trip such patches correctly when parsing and printing them, as we do not do this correctly right now.

Fixes #5157

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -206,16 +206,20 @@ static int apply_hunk(
 			goto done;
 		}
 
-		if (line->origin == GIT_DIFF_LINE_CONTEXT ||
-			line->origin == GIT_DIFF_LINE_DELETION) {
-			if ((error = git_vector_insert(&preimage.lines, line)) < 0)
-				goto done;
-		}
-
-		if (line->origin == GIT_DIFF_LINE_CONTEXT ||
-			line->origin == GIT_DIFF_LINE_ADDITION) {
-			if ((error = git_vector_insert(&postimage.lines, line)) < 0)
-				goto done;
+		switch (line->origin) {
+			case GIT_DIFF_LINE_CONTEXT:
+				if ((error = git_vector_insert(&preimage.lines, line)) < 0 ||
+				    (error = git_vector_insert(&postimage.lines, line)) < 0)
+					goto done;
+				break;
+			case GIT_DIFF_LINE_DELETION:
+				if ((error = git_vector_insert(&preimage.lines, line)) < 0)
+					goto done;
+				break;
+			case GIT_DIFF_LINE_ADDITION:
+				if ((error = git_vector_insert(&postimage.lines, line)) < 0)
+					goto done;
+				break;
 		}
 	}
 
--- a/src/apply.c
+++ b/src/apply.c
@@ -199,7 +199,7 @@ static int apply_hunk(
 
 	for (i = 0; i < hunk->line_count; i++) {
 		size_t linenum = hunk->line_start + i;
-		git_diff_line *line = git_array_get(patch->lines, linenum);
+		git_diff_line *line = git_array_get(patch->lines, linenum), *prev;
 
 		if (!line) {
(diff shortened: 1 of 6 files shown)


pks-t:
Fixes #5153

albfan:
@pks-t Nice! didn't see that. we can add the patch until it is fixed, really great! looking for this to be merged and the roadmap for releasing.

eaigner:
I don't know if this is related, but I tested your PR and it seems to work. But when I try to apply this patch, I cannot.

I applied the following patch in `git` without problems to the index

```
diff --git a/test/diff2.txt b/test/diff2.txt
index 212a1d6d8..b27c6ecb3 100644
--- a/test/diff2.txt
+++ b/test/diff2.txt
@@ -1,4 +1,4 @@
 This is just
 a text file to
 test git diff file
-changes
\ No newline at end of file
+changes
```

with

    era$ git apply --cached -- p1.patch

However, when I try to apply it via `git_apply`, it gives the following error

    hunk at line 1 did not apply

eaigner:
I have noticed that none of the patch parse tests actually verify if a patch was correctly parsed.

For instance parsing the following snippet with `git_diff_from_buffer`

```
diff --git a/test/diff2.txt b/test/diff2.txt
index 212a1d6..0dff799 100644
--- a/test/diff2.txt
+++ b/test/diff2.txt
@@ -1,3 +1,2 @@
 This is just
-a text file to
 changes
\ No newline at end of file
```

and then printing the resulting diff using `git_diff_to_buf` results in

```
diff --git a/test/diff2.txt b/test/diff2.txt
index 212a1d6..0dff799 100644
--- a/test/diff2.txt
+++ b/test/diff2.txt
@@ -1,3 +1,2 @@
 This is just
-a text file to
 changes
```

Notice that the `\ No newline at end of file` is missing afterward.
The patch could not be applied again in this state.

albfan:
That can be either we don't print out the diff about newline ir we are just ignoring it. I will do some checks too provide feedback

eaigner:
But `\ No newline at end of file` cannot be ignored, or the patch will not be correct.

eaigner:
I fixed this in #5161, please check!

pks-t:
> I have noticed that none of the patch parse tests actually verify if a patch was correctly parsed.

I know, as noticed in my commit message :) Anyway, I'm closing this PR in favour of yours, so thanks a lot for fixing this!

pks-t:
This supersedes #5161 again.

ethomson:
Fixes #5177

ethomson:
Nice fix, thanks @eaigner and @pks-t for the code, and @albfan, @ddevault and @mkaito for the helpful reports.

ddevault:
Thanks!

PR #5161, opened 2019-07-06
patch_parse: properly parse EOFNL comment lines

eaigner:
The patch output of #5159 did not produce a valid EOF comment

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -199,11 +199,22 @@ static int apply_hunk(
 
 	for (i = 0; i < hunk->line_count; i++) {
 		size_t linenum = hunk->line_start + i;
+		git_diff_line *next;
 		git_diff_line *line = git_array_get(patch->lines, linenum);
 
 		if (!line) {
 			error = apply_err("preimage does not contain line %"PRIuZ, linenum);
 			goto done;
+		}
+
+		/* lookahead for eofnl */
+		if (i < hunk->line_count - 1) {
+			next = git_array_get(patch->lines, linenum + 1);
+			if (next->origin == GIT_DIFF_LINE_CONTEXT_EOFNL ||
+					next->origin == GIT_DIFF_LINE_DEL_EOFNL ||
+					next->origin == GIT_DIFF_LINE_ADD_EOFNL) {
+				line->content_len -= 1;
+			}
 		}
 
 		if (line->origin == GIT_DIFF_LINE_CONTEXT ||
--- a/src/diff.c
+++ b/src/diff.c
@@ -474,17 +474,20 @@ static int line_cb(
 	GIT_UNUSED(hunk);
 
 	switch (line->origin) {
-	    case GIT_DIFF_LINE_ADDITION:
-		git_buf_putc(&buf, '+');
-		break;
-	    case GIT_DIFF_LINE_DELETION:
-		git_buf_putc(&buf, '-');
-		break;
-	    case GIT_DIFF_LINE_CONTEXT:
-		break;
-	    default:
-		git_error_set(GIT_ERROR_PATCH, "invalid line origin for patch");
-		return -1;
+		case GIT_DIFF_LINE_ADDITION:
+			git_buf_putc(&buf, '+');
+			break;
+		case GIT_DIFF_LINE_DELETION:
+			git_buf_putc(&buf, '-');
+			break;
+		case GIT_DIFF_LINE_CONTEXT:
+		case GIT_DIFF_LINE_CONTEXT_EOFNL:
+		case GIT_DIFF_LINE_ADD_EOFNL:
+		case GIT_DIFF_LINE_DEL_EOFNL:
+			break;
+		default:
+			git_error_set(GIT_ERROR_PATCH, "invalid line origin for patch");
+			return -1;
 	}
 
 	git_buf_put(&buf, line->content, line->content_len);
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -578,6 +578,16 @@ static int parse_hunk_body(
 			old_lineno = -1;
 			break;
 
+		case '\\':
+			/*
+			 * If there are no oldlines left, then this is probably
+			 * the "\ No newline at end of file" marker. Do not
+			 * verify its format, as it may be localized.
+			 */
+			if (!oldlines)
+				continue;
+			/* fall through */
+
 		default:
 			error = git_parse_err("invalid patch hunk at line %"PRIuZ, ctx->parse_ctx.line_num);
 			goto done;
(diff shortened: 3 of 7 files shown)


eaigner:
@pks-t I based this on your PR and validated the patch output. Would appreciate a quick look if I missed something.

src/patch_parse.c
```diff
@@ -578,6 +587,13 @@ static int parse_hunk_body(
 			old_lineno = -1;
 			break;
 
+		case '\\':
+			prefix = 0;
```
eaigner: @pks-t I'm not quite sure what the prefix `int` is supposed to do, and I kind of find it odd that it is subtracted from `parse_ctx.line` which is a `char *`. What is it?

eaigner:
`apply::fromdiff::patching_correctly_truncates_source` still failing, but i'm not quite sure what the goal of this test is

src/patch_parse.c
eaigner: Nevermind... simple pointer arithmetic 🙄

albfan:
```
* 329eedb84 (HEAD -> apply) ensure valid patch output
* c8406a3b5 (pks/patch-parse-old-missing-nl) patch_parse: handle missing newline indicator in old file
* 960d2a070 (origin/wip/albfan/diff_buffers, wip/albfan/diff_buffers) examples: consolidate includes into "common.h"
* 3be09b6c0 Compare buffers in diff example
*   398412ccd (upstream/master, upstream/HEAD, pks-t/master, origin/master, master) Merge pull request #5143 from libgit2/ethomson/warnings
```

I just rebase all the patches to test with examples, and will review (just if that helps). I called it https://github.com/albfan/libgit2/pull/new/fix-patch-to-diff-no-newline

albfan:
@pks-t without a deeper lot at this, `lg2 diff --no-index` works as it should with files without newline at end.

Still not sure how this is translated to a hunk, so I will test with our vala integration project: https://gitlab.gnome.org/albfan/diferencia

src/apply.c
```diff
@@ -199,13 +199,24 @@ static int apply_hunk(
...
 			error = apply_err("preimage does not contain line %"PRIuZ, linenum);
 			goto done;
 		}
 
+		/* lookahead for eofnl */
+		if (i < hunk->line_count - 1) {
+			next = git_array_get(patch->lines, linenum + 1);
+			if (next->origin == GIT_DIFF_LINE_CONTEXT_EOFNL ||
+					next->origin == GIT_DIFF_LINE_DEL_EOFNL ||
+					next->origin == GIT_DIFF_LINE_ADD_EOFNL) {
```
pks-t: Nit: indentation is off here. If conditionals span multiple lines, then the conditions should be indented by four spaces relative to the `if`, only:

```
if (next->origin == GIT_DIFF_LINE_CONTEXT_EOFNL ||
    next->origin == GIT_DIFF_LINE_DEL_EOFNL ||
    next->origin == GIT_DIFF_LINE_ADD_EOFNL) {
        ...
}
```

tests/patch/parse.c
```diff
@@ -27,6 +27,18 @@ static void ensure_patch_validity(git_patch *patch)
 	cl_assert_equal_i(0, delta->new_file.size);
 }
 
+static void ensure_identical_patch_inout(const char *content) {
+	git_buf buf = GIT_BUF_INIT;
+	git_patch *patch;
+
+	cl_git_pass(git_patch_from_buffer(&patch, content, strlen(content), NULL));
+	cl_git_pass(git_patch_to_buf(&buf, patch));
+	cl_assert_equal_strn(git_buf_cstr(&buf), content, strlen(content));
```
pks-t: This should be a `cl_assert_equal_str`, otherwise we wouldn't notice if the generated patch had unexpected trailing content

tests/patch/parse.c
```diff
@@ -27,6 +27,18 @@ static void ensure_patch_validity(git_patch *patch)
 	cl_assert_equal_i(0, delta->new_file.size);
 }
 
+static void ensure_identical_patch_inout(const char *content) {
```
pks-t: Style nit: Opening braces belong on their own line

tests/diff/patchid.c
```diff
@@ -22,7 +22,7 @@ void test_diff_patchid__simple_commit(void)
 
 void test_diff_patchid__filename_with_spaces(void)
 {
-	verify_patch_id(PATCH_APPEND_NO_NL, "f0ba05413beaef743b630e796153839462ee477a");
+	verify_patch_id(PATCH_APPEND_NO_NL, "cc0d67fc516f3378baa8a72b248da6e26c5dbf4a");
```
pks-t: This is incorrect, though. git-patch-id(1) prints "f0ba05413beaef743b630e796153839462ee477a", so we cannot simply alter the expected ID here. So this indicates that we have to adjust how we calculate the patch ID when there's no trailing newline. Just let me know whether you want to amend that or whether I should take over again :)

src/diff.c
```diff
@@ -474,17 +474,20 @@ static int line_cb(
...
+			git_buf_putc(&buf, '-');
+			break;
+		case GIT_DIFF_LINE_CONTEXT:
+		case GIT_DIFF_LINE_CONTEXT_EOFNL:
+		case GIT_DIFF_LINE_ADD_EOFNL:
+		case GIT_DIFF_LINE_DEL_EOFNL:
+			break;
+		default:
+			git_error_set(GIT_ERROR_PATCH, "invalid line origin for patch");
+			return -1;
```
pks-t: It would be nice to avoid re-indenting the whole block so that it's easier to spot what exactly has changed

src/apply.c
eaigner: But the project is set to indent with TAB

tests/diff/patchid.c
eaigner: yes, please take over. probably better someone fixes this that knows what the issue is ^^

src/apply.c
pks-t: Yeah. We had some misunderstandings around that in the past, but our rule is to follow git.git and thus linux.git's style. They use 8-space tabs in general, but indent multiline conditionals with four spaces. I've never used editorconfig, but if it is possible to specify this rule then we should certainly fix it.

pks-t:
Oops, sorry for closing your PR, I confused this one with #5159. Anyway, I've added your fix to #5159 now and fixed two issues with patch IDs and another one with how we strip newlines when applying the patch.

I can reopen this one if you feel like the combined fix in #5159 isn't sufficient yet, but I didn't want to force-push over your PR as I did quite some changes to your patch.

eaigner:
It's fine! Thanks for filling in the missing stuff. /closed

PR #5163, opened 2019-07-12
Ignore VS2017 specific files and folders

csware:
When using VS2017 there are some files in the libgit2 folder which should be ignored.


pks-t:
Thanks a lot, @csware!

csware:
Generally, why does Microsoft first create a .vs direcotry and then stores files outside of it?!

PR #5168, opened 2019-07-16
clar: correctly account for "data" suites when counting

tiennou:
Failing to do that makes clar miss the last of the suites, as all duplicated "data" would have not been accounted for.

### DIFF
--- a/tests/generate.py
+++ b/tests/generate.py
@@ -210,7 +210,7 @@ class TestSuite(object):
                     module.modified = True
 
     def suite_count(self):
-        return len(self.modules)
+        return sum(len(module.initializers) for module in self.modules.values())
 
     def callback_count(self):
         return sum(len(module.callbacks) for module in self.modules.values())


tiennou:
(Filed upstream: https://github.com/vmg/clar/pull/82)

tiennou:
No problem. I got quite confused by the "no such testsuite z::zz" messages in `libssh2` 😉.

I think the filtering would also gain to be improved — right now the filter syntax with "data labels" looks like this: `-ssystem::suite (label)::testcase`, and you can't refer to all labelled testcases of a group at once (IOW `-ssystem::suite` is not recognized. Also, I'm not fond of spaces + parenthesis appearing in those specifiers.

Would changing the syntax to something like `-ssystem::suite[-label]::testcase` would be acceptable ? That would allow filtering on any specific suite's "data label" (`-ssystem::suite[-label]`), all of them (`-ssystem::suite`), or all "data labels" (`-s-label`). That last one is accidental, but could be made to work.

(Just my 2c, my current pet clar peeve is skip reasons so it might get implemented first 😉).

pks-t:
Yeah, I just stumbled over the exact same issue. Didn't realize at first that we in fact stopped executing those tests altogether!

I'm certainly open to changing how the syntax looks like, and filtering by data label would certainly be useful.

pks-t:
But wait, I'm already using your PR but it's still wrong. Are we skipping over modules which have no initializers at all now?

pks-t:
In fact it works correctly without this PR :/

tests/generate.py
```diff
@@ -210,7 +210,7 @@ def disable(self, excluded):
                     module.modified = True
 
     def suite_count(self):
-        return len(self.modules)
+        return sum(len(module.initializers) for module in self.modules.values())
```
tiennou: > Are we skipping over modules which have no initializers at all now?

Sounds like it does, yes. `min(☝️, 1)` then ?

tiennou: I have no such test suites in `libssh2`, which explains why I missed it 😞.

pks-t: Yeah, I tried the same locally, but in fact this didn't fix the issue for me. Didn't dig deeper yet, though

pks-t: Oops. Because we both had the same issue of using `min` instead of `max`. See #5175 for a fix

PR #5173, opened 2019-07-18
ignore: fix determining whether a shorter pattern negates another

### LINKED ISSUE #5166: status: git_status_list_new errors on does_negate_rule
Steps:

- clone `git@github.com:nodejs/node.git`
- run `lg2 status` in the repository

Expected:

- No error

Actual:

- libgit errors out with `Could not get status [-1] - error matching pattern`

`git status` works just fine.

Apparently `libgit2` is not comfortable with the `node`s repo `.gitignore` file.

The error stems from `ignore.c:does_negate_rule:145`

The offending rule seems to be `!**/node_modules/**`

pks-t:
When computing whether we need to store a negative pattern, we iterate
through all previously known patterns and check whether the negative
pattern undoes any of the previous ones. In doing so we call `wildmatch`
and check it's return for any negative error values. If there was a
negative return, we will abort and bubble up that error to the caller.

In fact, this check for negative values stems from the time where we
still used `fnmatch` instead of `wildmatch`. For `fnmatch`, negative
values indicate a "real" error, while for `wildmatch` a negative value
may be returned if the matching was prematurely aborted. A premature
abort may for example also happen if the pattern matches a prefix of the
haystack if the pattern is shorter. Returning an error in that case is
the wrong thing to do.

Fix the code to compare for equality with `WM_MATCH`, only. Negative
values returned by `wildmatch` are perfectly fine and thus should be
ignored. Add a test that verifies we do not see the error.

---

I've skimmed through the tree to see whether there's any other places where we fail to use `wildmatch` correctly. The only place that struck me was when matching pathspecs, but I failed to come up with any breaking pattern. I think this one is fine, as `WM_ABORT_*` is only returned if the text is a prefix of the pattern. And as we're swallowing all negative values up the callchain we should be good.

Fixes #5166

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -141,13 +141,8 @@ static int does_negate_rule(int *out, git_vector *rules, git_attr_fnmatch *match
 		if (git_buf_oom(&buf))
 			goto out;
 
-		if ((error = wildmatch(git_buf_cstr(&buf), path, wildmatch_flags)) < 0) {
-			git_error_set(GIT_ERROR_INVALID, "error matching pattern");
-			goto out;
-		}
-
 		/* if we found a match, we want to keep this rule */
-		if (error != WM_NOMATCH) {
+		if ((wildmatch(git_buf_cstr(&buf), path, wildmatch_flags)) == WM_MATCH) {
 			*out = 1;
 			error = 0;
 			goto out;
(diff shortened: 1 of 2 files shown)


andreacanton:
Hi everyone, somebody knows when this fix will be released?
Thank you

pks-t:
We do not yet have any schedule for the next releases. Given that it has been nearly three months since doing the last bugfix release I'd be up to prepare one towards the end of this month.

pks-t:
With that being said, though, this only fixes an issue that has been introduced with the conversion to `wildmatch`, and in fact there is no release with the `wildmatch` code yet. So there is no need to backport this PR anyway

PR #5176, opened 2019-07-19
repository: do not initialize HEAD if it's provided by templates

### LINKED ISSUE #5148: git_repository_init_ext does not respect an existing head when not given a specific initial_head to create
### Version and downstream/background info

For completeness, this is using libgit2 through cargo via git2-rs.

```
cargo 1.35.0 (6f3e9c367 2019-04-04)
release: 1.35.0
commit-hash: 6f3e9c367abb497c64f360c3839dab5e74928d5c
commit-date: 2019-04-04
```

which specifies `git2` `^0.8.0`, which binds to some libgit2 somehow, I don't actually have a version number or release. However I've actually figured out this behaviour by looking at the current source, specifically at or around commit c4c1500a8bc99066cbb4340717d91020c02a75ae.

### Reproduction steps

Original repro is:
1. Configure gitconfig with an `init.templatedir`
2. In the template, have a `HEAD` file which contains: `ref: refs/heads/main`
3. Create a new cargo project with `cargo new foo`.

Pared down, this is actually doing a `git_repository_init_ext()` with the `git_repository_init_options *opts` set to:

```c
GIT_REPOSITORY_INIT_MKDIR | GIT_REPOSITORY_INIT_MKPATH | GIT_REPOSITORY_INIT_EXTERNAL_TEMPLATE
```

### Expected behavior

That the git repo is initialised with the default HEAD / branch set to `main`.

### Actual behavior

The git repo is initialised with the default HEAD / branch set to `master`.

### Analysis

I traced the process and found that my template's `HEAD` is correctly copied in place, and then a little later the `HEAD.lock` is opened, `ref: refs/heads/master` is written in, and the lock is swapped back as is usual. Here's the openat call:

```strace
[00007fbcc606b199] openat(AT_FDCWD, "/home/code/rust/foo2/.git/HEAD.lock", O_WRONLY|O_CREAT|O_EXCL, 0666) = 3
 > /usr/lib/libpthread-2.29.so(__open64+0x59) [0x13199]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(p_open+0xc5) [0x652a25]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(git_futils_creat_locked+0x14) [0x65cca4]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(git_filebuf_open_withsize+0x533) [0x65ea93]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(git_repository_create_head+0x68) [0x64a2a8]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(git_repository_init_ext+0x5fc) [0x64ad3c]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(git2::repo::Repository::init+0xaf) [0x3a62df]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(cargo::util::vcs::GitRepo::init+0x19) [0x292979]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(cargo::ops::cargo_new::mk+0x984) [0x1dc874]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(cargo::ops::cargo_new::new+0x559) [0x1d95e9]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(cargo::commands::new::exec+0xc0) [0x126130]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(cargo::cli::main+0x1a20) [0xe23d0]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(cargo::main+0xe0) [0x10fa90]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(std::rt::lang_start::{{closure}}+0x3) [0x106083]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(std::panicking::try::do_call+0x13) [0x909cb3]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(__rust_maybe_catch_panic+0x1a) [0x9156aa]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(std::rt::lang_start_internal+0x35d) [0x90a87d]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(main+0x22) [0x112132]
 > /usr/lib/libc-2.29.so(__libc_start_main+0xf3) [0x26ee3]
 > /home/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo(_start+0x29) [0xd3029]
```

In source, I'm looking at [the `git_repository_init_ext` routine](https://github.com/libgit2/libgit2/blob/c4c1500a8bc99066cbb4340717d91020c02a75ae/src/repository.c#L2092-L2097), specifically the three steps to creating a repo:

```c
if (!(error = repo_init_structure(
		repo_path.ptr, wd, opts)) &&
	!(error = repo_init_config(
		repo_path.ptr, wd, opts->flags, opts->mode)))
	error = git_repository_create_head(
		repo_path.ptr, opts->initial_head);
```

(`repo_init_structure` correctly copies the template `HEAD` in place.)

and [the `git_repository_create_head` routine](https://github.com/libgit2/libgit2/blob/c4c1500a8bc99066cbb4340717d91020c02a75ae/src/repository.c#L1376), specifically as to how, when the `opts->initial_head` is not given, it writes the `HEAD` without checking what's in it first.

So that's my bug report: `git_repository_create_head` does not respect an existing head when not given a specific head to create.

Alternatively: `git_repository_init_ext` unconditionally calls `git_repository_create_head` when not given `opts->initial_head`, even though I believe it should check for an existing head first as the template may override that bit.

pks-t:
When using templates to initialize a git repository, then git-init(1)
will copy over all contents of the template directory. These will be
preferred over the default ones created by git-init(1). While we mostly
do the same, there is the exception of "HEAD". While we do copy over the
template's HEAD file, afterwards we'll immediately re-initialize its
contents with either the default "ref: refs/origin/master" or the init
option's `initial_head` field.

Let's fix the inconsistency with upstream git-init(1) by not overwriting
the template HEAD, but only if the user hasn't set `opts.initial_head`.
If the `initial_head` field has been supplied, we should use that
indifferent from whether the template contained a HEAD file or not. Add
tests to verify we correctly use the template directory's HEAD file and
that `initial_head` overrides the template.

---

This PR also includes some refactorings to update code style in several places which I was touching.

Fixes #5148


src/repository.c
```diff
@@ -2085,7 +2086,18 @@ int git_repository_init_ext(
 		/* TODO: reinitialize the templates */
 	} else {
 		if ((error = repo_init_structure(repo_path.ptr, wd, opts)) < 0 ||
-		    (error = repo_init_config(repo_path.ptr, wd, opts->flags, opts->mode)) < 0 ||
+		    (error = repo_init_config(repo_path.ptr, wd, opts->flags, opts->mode)) < 0)
+			goto out;
+
+		if ((error = git_buf_joinpath(&head_path, repo_path.ptr, GIT_HEAD_FILE)) < 0)
```
tiennou: Style: I'd merge that `if` with the one above.

(I'm also still worried that every time a `GIT_HEAD_FILE` occurrence is added, it usually means custom refdb won't know about it).

src/repository.c
```diff
@@ -2056,53 +2056,48 @@ int git_repository_init_ext(
 	const char *given_repo,
 	git_repository_init_options *opts)
 {
-	int error;
-	git_buf repo_path = GIT_BUF_INIT, wd_path = GIT_BUF_INIT,
-		common_path = GIT_BUF_INIT;
+	git_buf repo_path = GIT_BUF_INIT, wd_path = GIT_BUF_INIT, common_path = GIT_BUF_INIT;
```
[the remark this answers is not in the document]
pks-t: Fixed

src/repository.c
pks-t: > Style: I'd merge that if with the one above.

Fixed

> (I'm also still worried that every time a GIT_HEAD_FILE occurrence is added, it usually means custom refdb won't know about it).

Yup, I thought about that, too. In this case it's fine though, as we're creating a completely new repository with `git_repository_init_ext`, which doesn't support custom backends anyway. So we need to have `GIT_HEAD_FILE` created here. _But_, I think in the long term we should fix `git_repository_create_head`. This one is a generic version and should thus use the refdb.

pks-t:
Thanks for your review, I've amended fixes

ethomson:
Thanks for tackling this, @pks-t.  Really appreciate all the cleanups with the fix.

PR #5179, opened 2019-07-20
patch_parse: fix segfault due to line containing static contents

pks-t:
With commit dedf70ad2 (patch_parse: do not depend on parsed buffer's
lifetime, 2019-07-05), all lines of the patch are allocated with
`strdup` to make lifetime of the parsed patch independent of the buffer
that is currently being parsed. In patch b08932824 (patch_parse: ensure
valid patch output with EOFNL, 2019-07-11), we introduced another
code location where we add lines to the parsed patch. But as that one
was implemented via a separate pull request, it wasn't converted to use
`strdup`, as well. As a consequence, we generate a segfault when trying
to deallocate the potentially static buffer that's now in some of the
lines.

Use `git__strdup` to fix the issue.

---

This bug has been introduced due to #5158 and #5159 having bad interactions with each other.

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -653,7 +653,7 @@ static int parse_hunk_body(
 
 		memset(line, 0x0, sizeof(git_diff_line));
 
-		line->content = ctx->parse_ctx.line;
+		line->content = git__strdup(ctx->parse_ctx.line);
 		line->content_len = ctx->parse_ctx.line_len;
 		line->content_offset = ctx->parse_ctx.content_len - ctx->parse_ctx.remain_len;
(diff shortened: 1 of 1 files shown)


ethomson:
Oops, nice catch.

pks-t:
Yeah. Unfortunately this breaks all current test runs, so I'd like to get this in as quick as possible

ethomson:
> Yeah. Unfortunately this breaks all current test runs, so I'd like to get this in as quick as possible

Indeed.  I'm going to go ahead and merge it.

pks-t:
Thanks a lot for the fast lane! :)

PR #5180, opened 2019-07-20
fuzzer: use futils instead of fileops

ethomson:
Updated after the renamification in #5151


pks-t:
Thanks for cleaning up after me!

PR #5181, opened 2019-07-21
config_file: refresh when creating an iterator

### LINKED ISSUE #5167: git_remote_list does not detect config changes on disk
### Reproduction steps
1. `git_repository_open` a repository twice (`A` and `B`).
2. `git_remote_create` on repo `A`.
3. `git_remote_list` on repo `B`.

### Expected behavior
The created remote should be listed in both repositories.

### Actual behavior
Only repo `A` sees the new repository.
Repo `B` will not list the new repository.

I have a [branch on my fork](https://github.com/Mr-Wallet/libgit2/tree/issue/multi-thread-remote-list) which includes a [test](https://github.com/Mr-Wallet/libgit2/commit/7ee3eed88fab61017b71455197cdf9c0a514884a) reproducing the issue.

Repo `B` _can_ be induced to bust its config cache with `git_repository_config_snapshot`, but this is burdensome if I'm not interested in the config, because I need to immediately free it.

This issue is a concern when multi-threading, where object shouldn't be shared between threads. After I finish creating a remote on one thread, I would expect to be able to read the remote list on another thread at any point, and get accurate results.

I'm not very familiar with LibGit2, so maybe there's already a cache-busting mechanism, but I was unable to find any mention in the API.

Assuming this isn't just a matter of me doing it wrong, some possible solutions:
- Always bust the cache on `git_remote_list`.
- Always `stat` the config(s) and bust the cache if they've been modified since the last time the configs were parsed.
- Provide an option to `git_remote_list` for busting the cache.
- Directly expose/document a function that busts the cache in the same way that `git_repository_config_snapshot` does, but without taking an `out` parameter which I need to manually free.

f92d495d44bff2bdb6fd99524a93986ea268c9e8 (`master` as of `2019-07-15T22:00:00Z`)

The hyperlinked test was run on Windows 10 x64, but our downstream project ([nodegit](https://github.com/nodegit/nodegit)) showed this behavior on Linux and MacOS as well.

pks-t:
When creating a new iterator for a config file backend, then we should
always make sure that we're up to date by calling `config_refresh`.
Otherwise, we might not notice when another process has modified the
configuration file and thus will represent outdated values.

Add two tests to config::stress that verify that we get up-to-date
values when reading configuration entries via `git_config_iterator`.

---

I think this wasn't caused by #5132 but was a pre-existing issue already. There simply was no code path that would've called `config_refresh` when creating a new iterator.

Fixes #5167

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -270,13 +270,12 @@ static int config_iterator_new(
 	struct git_config_backend *backend)
 {
 	diskfile_header *bh = GIT_CONTAINER_OF(backend, diskfile_header, parent);
-	git_config_entries *entries;
+	git_config_entries *entries = NULL;
 	int error;
 
-	if ((error = git_config_entries_dup(&entries, bh->entries)) < 0)
-		return error;
-
-	if ((error = git_config_entries_iterator_new(iter, entries)) < 0)
+	if ((error = config_refresh(backend)) < 0 ||
+	    (error = git_config_entries_dup(&entries, bh->entries)) < 0 ||
+	    (error = git_config_entries_iterator_new(iter, entries)) < 0)
 		goto out;
 
 out:
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -231,8 +231,10 @@ static int config_refresh(git_config_backend *cfg)
 	git_config_entries *entries = NULL;
 	int error, modified;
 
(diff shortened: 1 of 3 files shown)


ethomson:
Thanks much for the helpful report and the very nice tests @Mr-Wallet!

PR #5184, opened 2019-07-24
Fix example checkout to forbid rather than require --

novalis:
Make the example program for checkout follow git syntax, where "--" indicates a file.  This was likely just a strcmp return value confusion.

### DIFF
--- a/examples/checkout.c
+++ b/examples/checkout.c
@@ -195,7 +195,7 @@ int lg2_checkout(git_repository *repo, int argc, char **argv)
 		fprintf(stderr, "unhandled\n");
 		err = -1;
 		goto cleanup;
-	} else if (strcmp("--", args.argv[args.pos])) {
+	} else if (!strcmp("--", args.argv[args.pos])) {
 		/**
 		 * Try to checkout the given path
 		 */


examples/checkout.c
```diff
@@ -195,7 +195,7 @@ int lg2_checkout(git_repository *repo, int argc, char **argv)
 		fprintf(stderr, "unhandled\n");
 		err = -1;
 		goto cleanup;
-	} else if (strcmp("--", args.argv[args.pos])) {
+	} else if (!strcmp("--", args.argv[args.pos])) {
```
pks-t: Yeah, this definitely looks like an error to me. If you ask me, though, this block could be improved even further: `else if (args.pos == args.argc - 1) { resolve_refish... } else { fprintf(stderr, "Only switching between branches is supported"); }`. We only support checking out a ref anyway, so we could just assert that we received a single parameter only and, if so, unconditionally pass that to `resolve_refish`. Because as the code stands right now, we just ignore every parameter after the first one.

novalis: Sure -- there's a ton of possible improvements here.  I just happened to notice that while trying to investigate #5185 (which turned out to be only for merges).  So I'm unlikely to make further changes right now.  Feel free to merge or ignore.

pks-t: Fair enough!

PR #5186, opened 2019-07-24
config: separate file and snapshot backends

pks-t:
This completely tears apart the shared code between config file (`diskfile_backend`) and config snapshots (`diskfile_readonly_backend`). The previous code was hard to understand with the `diskfile_header` structure, error prone and didn't have proper separation of concerns. With this, we now have a completely backend-agnostic snapshotting backend that can create a read-only snapshot of any other config backend.

This PR doesn't yet go the extra step of making use of that extra functionality, but only does refactorings that should keep all semantics in place. But hopefully with less bugs and more enlightenment when one reads the code.

/cc @csware

### DIFF
--- a/src/config_entries.c
+++ b/src/config_entries.c
@@ -67,6 +67,36 @@ int git_config_entries_new(git_config_entries **out)
 	return error;
 }
 
+int git_config_entries_dup_entry(git_config_entries *entries, const git_config_entry *entry)
+{
+	git_config_entry *duplicated;
+	int error;
+
+	duplicated = git__calloc(1, sizeof(git_config_entry));
+	GIT_ERROR_CHECK_ALLOC(duplicated);
+
+	duplicated->name = git__strdup(entry->name);
+	GIT_ERROR_CHECK_ALLOC(duplicated->name);
+
+	if (entry->value) {
+		duplicated->value = git__strdup(entry->value);
+		GIT_ERROR_CHECK_ALLOC(duplicated->value);
+	}
+	duplicated->level = entry->level;
+	duplicated->include_depth = entry->include_depth;
+
+	if ((error = git_config_entries_append(entries, duplicated)) < 0)
+		goto out;
+
+out:
+	if (error && duplicated) {
+		git__free((char *) duplicated->name);
+		git__free((char *) duplicated->value);
+		git__free(duplicated);
+	}
+	return error;
+}
+
 int git_config_entries_dup(git_config_entries **out, git_config_entries *entries)
 {
 	git_config_entries *result = NULL;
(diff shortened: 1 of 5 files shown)


src/config_file.c
```diff
@@ -284,6 +284,25 @@ static int config_iterator_new(
 	return error;
 }
 
+static int config_iterator_new_readonly(
```
tiennou: commitmsg: `Unforutanely`.

src/config_file.c
```diff
@@ -121,21 +117,21 @@ static void config_file_clear(diskfile *file)
...
+	if ((res = git_config_entries_new(&b->entries)) < 0)
 		return res;
 
 	if (!git_path_exists(b->file.path))
 		return 0;
 
-	if (res < 0 || (res = config_read(b->header.entries, repo, &b->file, level, 0)) < 0) {
-		git_config_entries_free(b->header.entries);
-		b->header.entries = NULL;
+	if (res < 0 || (res = config_read(b->entries, repo, &b->file, level, 0)) < 0) {
```
tiennou: I don't think `res` can be negative here, as l126 would have handled it.

src/config_entries.c
```diff
@@ -67,6 +67,36 @@ int git_config_entries_new(git_config_entries **out)
 	return error;
 }
 
+int git_config_entries_dup_entry(git_config_entries *entries, const git_config_entry *entry)
+{
+	git_config_entry *dup;
```
tiennou: I think that *could* induce cringing from special-snowflake compilers ? A point for the C90-compatible discussion maybe ?

src/config_entries.c
```diff
@@ -67,6 +67,36 @@ int git_config_entries_new(git_config_entries **out)
...
+	int error;
+
+	dup = git__calloc(1, sizeof(git_config_entry));
+	GIT_ERROR_CHECK_ALLOC(dup);
```
tiennou: Good catch. I'm surprised Coverity didn't complain, and noticed that it doesn't seem to be receiving updates ("May", it says)…

pks-t: Maybe this is related to our move to Azure Pipelines?

src/config_entries.c
pks-t: Good catch, I've renamed it to `entry` instead

src/config_file.c
pks-t: True. I've tried to refrain from updating existing code though to more clearly highlight what changed across the moves.

src/config_file.c
pks-t: Fixed.

src/config_entries.c
```diff
@@ -67,6 +67,36 @@ int git_config_entries_new(git_config_entries **out)
 	return error;
 }
 
+int git_config_entries_dup_entry(git_config_entries *entries, const git_config_entry *entry)
```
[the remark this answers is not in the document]
pks-t: Fixed.

pks-t:
Fixed all your comments except the one issue that was pre-existing, thanks a lof for your review! I've also fixed a memory leak that I've introduced

src/config_entries.c
ethomson: 🙁 It was definitely working at one point, then coverity went offline.  I'll take a look when I have a moment.

PR #5189, opened 2019-07-27
Optionally read `.gitattributes` from HEAD

ethomson:
Currently, we honor `.gitattributes`, behaving like git - reading from the working directory and / or the index.  We do not read any attributes out of `HEAD`.

However, some consumers may want to examine a bare repository, and read the file contents _as if_ the filters were being applied.  Provide an update to `git_blob_filter` that take a new option to indicate that filters should be read from `HEAD`.

Deprecate `git_blob_filtered_content` in favor of `git_blob_filter`.


include/git2/deprecated.h
```diff
@@ -89,6 +89,11 @@ GIT_EXTERN(int) git_blob_create_fromstream_commit(
 	git_writestream *stream);
 GIT_EXTERN(int) git_blob_create_frombuffer(
 	git_oid *id, git_repository *repo, const void *buffer, size_t len);
+GIT_EXTERN(int) git_blob_filtered_content(
```
pks-t: Mh. A short pointer would be nice for users: `/* Deprecated in favor of @git_blob_filter */`

tests/filter/systemattrs.c
```diff
@@ -0,0 +1,54 @@
...
+
+void test_filter_systemattrs__initialize(void)
+{
+	g_repo = cl_git_sandbox_init("crlf");
+	cl_must_pass(p_unlink("crlf/.gitattributes"));
+
+    cl_git_pass(git_libgit2_opts(
+        GIT_OPT_GET_SEARCH_PATH, GIT_CONFIG_LEVEL_SYSTEM, &system_attr_path));
+	cl_git_pass(git_buf_joinpath(&system_attr_path,
+	    system_attr_path.ptr, "gitattributes"));
```
pks-t: Indentation

tests/filter/systemattrs.c
```diff
@@ -52,3 +52,30 @@ void test_filter_systemattrs__reads_system_attributes(void)
 	git_buf_dispose(&buf);
```
pks-t: Typo in commit message: "optionall"

include/git2/blob.h
```diff
@@ -102,6 +102,12 @@ GIT_EXTERN(git_off_t) git_blob_rawsize(const git_blob *blob);
 typedef enum {
 	/** When set, filters will not be applied to binary files. */
 	GIT_BLOB_FILTER_CHECK_FOR_BINARY = (1 << 0),
+
+	/**
+	 * When set, filters will no load configuration from the
```
pks-t: "not"

src/blob.c
```diff
@@ -422,6 +422,9 @@ int git_blob_filter(
 	    git_blob_is_binary(blob))
 		return 0;
 
+	if ((opts.flags & GIT_BLOB_FILTER_NO_SYSTEM_ATTRIBUTES) != 0)
+		flags |= GIT_FILTER_NO_SYSTEM_ATTRIBUTES;
+
 	if (!(error = git_filter_list_load(
 	    &fl, git_blob_owner(blob), blob, path,
```
pks-t: Indentation here is off

include/git2/attr.h
```diff
@@ -120,12 +120,20 @@ GIT_EXTERN(git_attr_value_t) git_attr_value(const char *attr);
 
 /**
  * Check attribute flags: Using the system attributes file.
+ * Check attribute flags: controlling extended attribute behavior.
```
pks-t: Was the intent here to erase the preceding line?

src/attr_file.c
```diff
@@ -157,6 +158,22 @@ int git_attr_file__load(
...
+			goto cleanup;
+
+		/*
+		 * Do not assume that data straight from the ODB is NULL-terminated;
+		 * copy the contents of a file to a buffer to work on.
+		 */
+		blobsize = git_blob_rawsize(blob);
+
+		GIT_ERROR_CHECK_BLOBSIZE(blobsize);
+		git_buf_put(&content, git_blob_rawcontent(blob), (size_t)blobsize);
```
pks-t: There's no error check for `content` afterwards, we directly detach its contents, so we have to check for OOM here

src/attr.c
```diff
@@ -460,7 +471,7 @@ static int push_attr_file(
 static int push_one_attr(void *ref, const char *path)
 {
 	attr_walk_up_info *info = (attr_walk_up_info *)ref;
-	git_attr_file_source src[2];
+	git_attr_file_source src[GIT_ATTR_FILE_NUM_SOURCES];
```
pks-t: This is weird. We had three sources previously (in memory, from file, from index). Was this simply wrong and could've led to out-of-bounds access?

include/git2/blob.h
```diff
@@ -108,6 +108,12 @@ typedef enum {
 	 * system-wide `gitattributes` in `/etc` (or system equivalent).
```
pks-t: There's a word missing in the commit message here: "When `GIT_BLOB_FILTER_ATTTRIBUTES_FROM_HEAD` to `git_blob_filter`". Probably "passing"?

tests/filter/bare.c
```diff
@@ -0,0 +1,114 @@
...
+static git_repository *g_repo = NULL;
+static git_blob_filter_options filter_opts;
+
+void test_filter_bare__initialize(void)
+{
+	cl_fixture_sandbox("crlf.git");
+	cl_git_pass(git_repository_open(&g_repo, "crlf.git"));
+
+	filter_opts.flags |= GIT_BLOB_FILTER_NO_SYSTEM_ATTRIBUTES;
+	filter_opts.flags |= GIT_BLOB_FILTER_ATTTRIBUTES_FROM_HEAD;
```
pks-t: If we were correctly checking the version of the struct then we'd produce an error, right? So first, we need to initialize `filter_opts` correctly so that its version is set to "1". And second, we need to include the options version check

tests/filter/bare.c
```diff
@@ -0,0 +1,114 @@
...
+	cl_assert_equal_i(0, git_blob_rawsize(blob));
+	cl_assert_equal_s("", git_blob_rawcontent(blob));
+
+	memset(&buf, 0, sizeof(git_buf));
+	cl_git_pass(git_blob_filter(&buf, blob, "file.bin", &filter_opts));
+	cl_assert_equal_sz(0, buf.size);
+	cl_assert_equal_s("", buf.ptr);
+	git_buf_dispose(&buf);
+
+	memset(&buf, 0, sizeof(git_buf));
```
pks-t: Shouldn't `git_buf_dispose` leave the buffer in a ready-to-use state already?

tests/filter/bare.c
```diff
@@ -0,0 +1,114 @@
...
+	git_blob *blob;
+	git_buf buf;
+
+	cl_git_pass(git_revparse_single(
+		(git_object **)&blob, g_repo, "e69de29")); /* zero-byte */
+
+	cl_assert_equal_i(0, git_blob_rawsize(blob));
+	cl_assert_equal_s("", git_blob_rawcontent(blob));
+
+	memset(&buf, 0, sizeof(git_buf));
```
pks-t: I sometimes hate that we do not expose any mechanism to properly initialize `git_buf`. Should we maybe start exposing `GIT_BUF_INIT` and `git_buf_init`?

tests/resources/crlf.git/COMMIT_EDITMSG
```diff
@@ -1 +1 @@
-Add .gitattributes
+subdir with no translation
```
pks-t: You might consider dropping some of the files in "crlf.git", like "description", "COMMIT_EDITMSG", "info/exclude".

include/git2/attr.h
ethomson: Fixed.

include/git2/deprecated.h
ethomson: Fixed.

tests/filter/bare.c
ethomson: Fixed.

tests/filter/systemattrs.c
ethomson: Fixed.

src/attr.c
ethomson: In this codepath we call `attr_decide_sources` which will only include the source `GIT_ATTR_FILE__FROM_FILE` and/or `GIT_ATTR_FILE__FROM_INDEX`, and may also include `GIT_ATTR_FILE__FROM_HEAD`.  This will over-allocate by at least one, but that seems acceptable for the readability and maintainability trade-off.

tests/resources/crlf.git/COMMIT_EDITMSG
ethomson: Fixed.

tests/filter/bare.c
ethomson: Maybe `GIT_BUF_INIT`, but not `git_buf_init` (at least not without modification, since `git_buf_init` does an allocation.)

tests/filter/bare.c
ethomson: Yep, switched over to that.  Thanks!

include/git2/blob.h
ethomson: Fixed.

src/attr_file.c
ethomson: Fixed.

PR #5191, opened 2019-07-28
config: check if we are running in a sandboxed environment

### LINKED ISSUE #5190: config: fetching global config does not work in a sandboxed environment
If a Mac app is sandboxed, the `$HOME` environment variable points to the container, instead of the actual user home directory, thus preventing the default configuration to load.

On a sandboxed environment `$HOME` points to: `/Users/<usr>/Library/Containers/<app-bundle-id>/Data`

The correct way to fetch the home directory for a sandboxed environment according to Apple would be

```
#include <sys/types.h>
#include <pwd.h>

struct passwd *pw;
pw = getpwuid(getuid());
pw->pw_dir // correct $HOME
```

eaigner:
On macOS the $HOME environment variable returns the path to the sandbox
container instead of the actual user $HOME for sandboxed apps.
To get the correct path, we have to get it from the password file entry.

Closes #5190

### DIFF
--- a/src/sysdir.c
+++ b/src/sysdir.c
@@ -82,15 +82,25 @@ static int git_sysdir_guess_global_dirs(git_buf *out)
 #else
 	int error;
 	uid_t uid, euid;
+	const char *sandbox_id;
 
 	uid = getuid();
 	euid = geteuid();
 
+	/**
+	 * If APP_SANDBOX_CONTAINER_ID is set, we are running in a
+	 * sandboxed environment on macOS.
+	 */
+	sandbox_id = getenv("APP_SANDBOX_CONTAINER_ID");
+
 	/*
 	 * In case we are running setuid, use the configuration
 	 * of the effective user.
+	 *
+	 * If we are running in a sandboxed environment on macOS,
+	 * we have to get the HOME dir from the password entry file.
 	 */
-	if (uid == euid)
+	if (!sandbox_id && uid == euid)
 	    error = git__getenv(out, "HOME");
 	else
 	    error = get_passwd_home(out, euid);


src/sysdir.c
```diff
@@ -82,15 +82,22 @@ static int git_sysdir_guess_global_dirs(git_buf *out)
...
+	const char *sandbox_id;
 
 	uid = getuid();
 	euid = geteuid();
 
+	/*
+	 * If APP_SANDBOX_CONTAINER_ID is set, we are running in a
+	 * sandboxed environment on macOS.
+	 */
+	sandbox_id = getenv("APP_SANDBOX_CONTAINER_ID");
```
pks-t: This needs to be `p_getenv`

eaigner: Can't find `p_getenv` anywhere p_getenv: `Implicit declaration of function 'p_getenv'`. Is this defined somewhere externally?

eaigner: `p_getenv` is also not used anywhere else in the project, except a non active code branch in `remote.c` which is not compiled.

neithernut: I *think* he meant `git__getenv()`, which is used to query the `HOME` environment variable.

eaigner: Yes, but `git__getenv` is just a wrapper around `getenv` that returns a buffer, which we don't need. So what would be the point of that?

tiennou: `p_` is the POSIX-compat prefix. This could be a suggestion to implement `p_getenv` on macOS with a fallback for `$HOME` (which could be considered — otherwise there's a big chance the next `getenv("HOME")` added will suffer the same issue). But that felt heavy-handed, and I'm missing the point made as well…

ethomson: Don't implement a `p_getenv` - we used to have one, that's why we added `git__getenv` because there's no POSIXy way to do this on Windows.

I think this is :+1: as-is, since it's non-Windows only.

pks-t: Yeah, sorry about that. A quick grep for `getenv` returned `p_getenv` to me, which is why I assumed that we had the POSIX wrappers in place. One reason why I hate ifdef'fed blocks, I'll create a PR to just remove that block.

pks-t:
Thank you, @eaigner!

PR #5192, opened 2019-07-29
object: deprecate git_object__size for removal

carlosmn:
In #5118 we remove the double-underscore to make it a normally-named public
function. However, this is not an interesting function outside of the library
and it takes up a name for something that could be more useful.

Remove the single-underscore version as we have not done any releases with it.


carlosmn:
If it turns out someone really needs it we can bring it back as `git_object_sizeof` or something like that. I'm glad to have this name free now for a a getter for the size of the *git* object.

Issue #5154, opened 2019-06-30
DllMain included in static library on Windows


kevinlul:
### Reproduction steps
Linking against both `libgit2` and `irrlicht` statically in a project. 

### Expected behavior
The project compiles and links successfully.

### Actual behavior
Linking fails with a duplicate definition of `DllMain`.

### Proposed fix
In https://github.com/libgit2/libgit2/blob/master/src/global.c, wrap `BOOL WINAPI DllMain` in a preprocessor flag checking if we are building as a static library. The same bug also exists in other libraries that compile this into their binaries even if not building a DLL.

### Version of libgit2 (release number or SHA1)
0.28.2

### Operating system(s) tested
Windows 10

ethomson:
Hmm, interesting.  Indeed we should not be defining a `DllMain` unless we're actually building a dll.

However, we'll need to clean up our TLS data a bit differently: https://stackoverflow.com/a/3245082/729881

thiloschulz:
Hi, I had the same error. If this still bothers you, see https://github.com/libgit2/libgit2/pull/5194. It requires the application writer to call git_libgit2_dllmain() from their own DllMain() though.

kevinlul:
Since the issue only occurs when two libraries contain `DllMain` even when static linking, I resolved my specific issue by building `irrlicht` from source and patching it so that `DllMain` was not compiled when linking statically. A similar solution could be used here.

thiloschulz:
If an application has a DllMain() that handles the thread destruction case then it is usually for a good reason. Having a static library does not make issues like thread local storage go away. So just #undeffing it may result in memory leaks and other bad behaviour.

thiloschulz:
Hmmm.. continuing that thought: DllMain() is never called for an .EXE - so libgit2 will leak memory in either case if you link it statically into your problem.
This means the method from https://stackoverflow.com/a/3245082/729881 is the only way to do it properly in this case.

thiloschulz:
As became evident from the discussion in https://github.com/libgit2/libgit2/issues/5207, the solution from https://stackoverflow.com/a/3245082/729881 is still not complete.

A thread may persist even if git_libgit2_shutdown() has called TlsFree() on the TLS slot. In this case, the "destructor thread" will call TlsGetValue() on a freed slot (memory leak), or even worse: on a slot  allocated for another purpose in the meantime and try to free that one (heap corruption).

[...]

PR #5194, opened 2019-07-30
Fix DllMain() symbol clash on static libgit2 builds on Windows

thiloschulz:
When linking your own user application DLL with its own DllMain() function against a static libgit2 build, the DllMain() Symbol from libgit2 clashes. For the static build case, I have added a git_libgit2_dllmain() that must be called from the user application's DllMain() function.


thiloschulz:
Mhm. Build error probably due to rebase. I will take a look at it later today.

thiloschulz:
Alright. Build errors have been fixed. The new git_libgit2_dllmain() function however is not included in the documentation yet due to the #ifdeffing I presume. You probably will know better than me how to make it visible to doxygen without making the git_libgit2_dllmain() prototype visible in all build configurations.

src/global.c
```diff
@@ -387,3 +370,48 @@ git_global_st *git__global_state(void)
 }
 
 #endif /* GIT_THREADS */
+
+#ifdef GIT_WIN32 
+BOOL git__dllmain(HINSTANCE hInstDll, DWORD fdwReason, LPVOID lpvReserved)
```
pks-t: All `git__dllmain` implementations should be static

CMakeLists.txt
```diff
@@ -118,6 +118,10 @@ IF (DEPRECATE_HARD)
 	ADD_DEFINITIONS(-DGIT_DEPRECATE_HARD)
 ENDIF()
 
+IF (NOT BUILD_SHARED_LIBS)
+	ADD_DEFINITIONS(-DGIT_STATIC)
+ENDIF()
```
pks-t: I think I'd prefer if this was present in "src/features.h.in" to make the definition available in "global.h".

include/git2/global.h
```diff
@@ -38,6 +38,28 @@ GIT_EXTERN(int) git_libgit2_init(void);
  */
 GIT_EXTERN(int) git_libgit2_shutdown(void);
 
+#if defined(GIT_WIN32) && defined(_INC_WINDOWS)
```
pks-t: If it was declared in "features.h.in", then this here could become `#if defined(GIT_WIN32) && defined(GIT_STATIC)`.

src/global.c
```diff
@@ -387,3 +370,48 @@ git_global_st *git__global_state(void)
...
+}
+#else
+BOOL git_libgit2_dllmain(HINSTANCE hInstDll, DWORD fdwReason, LPVOID lpvReserved)
+{
+	GIT_UNUSED(hInstDll);
+	GIT_UNUSED(fdwReason);
+	GIT_UNUSED(lpvReserved);
+
+	return TRUE;
+}
```
pks-t: And subsequently this function wouldn't need a definition `#ifndef GIT_STATIC`.

src/transports/auth_ntlm.c
```diff
@@ -5,8 +5,9 @@
  * a Linking Exception. For full terms see the included COPYING file.
  */
 
-#include "git2.h"
 #include "common.h"
+
+#include "git2.h"
```
pks-t: In fact, "common.h" should be dropped altogether. Instead, "auth_ntlm.h" should come first and "auth_ntlm.h" should have an include on "common.h" first

include/git2/global.h
thiloschulz: The test for _INC_WINDOWS is absolutely necessary because we don't know at which point the user application has included Windows.h or whether the user application wants to include Windows.h at all.
Otherwise we must include Windows.h from git2.h so that the windows data types for the DLL handler prototype are available.

thiloschulz: This is probably not what you want.

src/global.c
thiloschulz: Thank you for your review!
You're absolutely right of course.

src/global.c
thiloschulz: This requirement is on purpose. If you do it my way, a user application that uses git_libgit2_dllmain() and supports linking against a static _or_ shared build of libgit2 does not need to handle those cases separately.

include/git2/global.h
pks-t: Fair enough :)

src/global.c
pks-t: Agreed

include/git2/global.h
thiloschulz: You may have a point though. I am not quite sure for which use case this thread handling via DllMain() is meant. Does the HTTPS or any other Transport mode create threads internally in libgit2, or is it for when a user application uses libgit2 in ist own multiple threads?

If it is the former, we really should let the compiler bail out so the user knows he must do something (call git_libgit2_dllmain() for static links) or it breaks libgit2.

thiloschulz:
I am closing this pull request because it will cover the case when you link a static libgit2 into your own DLL, but not if you link it into an .EXE. See discussion https://github.com/libgit2/libgit2/issues/5154

PR #5193, opened 2019-07-29
transports: Avoid walking over non-commit objects

biodranik:
The updated version of stalled #4383. The issue is still there for many years and needs to be fixed.

References may point to objects that are not commits, such as tags to
blobs. In that case, the reference must not be walked over further.

Fixes #3595.

PR #5195, opened 2019-07-31
smart: use push_glob instead of manual filtering

### LINKED ISSUE #3595: libgit2 assumes everything not under refs/tags are commits

tiennou:
The code worked under the assumption that anything under `refs/tags` are tag objects, and all the rest would be peelable to a commit. As it is completely valid to have tags to blobs under a non `refs/tags` ref, this would cause failures when trying to peel a tag to a commit (#3595).

Fix the broken filtering by switching to `git_revwalk_push_glob`, which already handles this case.

Fixes #3595, supersedes #4383, #5193.


src/transports/smart_protocol.c
```diff
@@ -358,7 +314,10 @@ int git_smart__negotiate_fetch(git_transport *transport, git_repository *repo, c
 	if ((error = git_pkt_buffer_wants(wants, count, &t->caps, &data)) < 0)
 		return error;
 
-	if ((error = fetch_setup_walk(&walk, repo)) < 0)
+	if ((error = git_revwalk_new(&walk, repo)) < 0)
+		return error;
```
ethomson: `goto on_error` to clean up `data`?

ethomson: If you felt like it, making `data` the first param to `git_pkt_buffer_wants` might help in understanding that it's an out param...

tiennou: Fixed.

This is in the vicinity of #5105, hence I didn't want more extensive changes here. If that's fine with you I'll tackle that out there.

tests/refs/foreachglob.c
```diff
@@ -48,8 +48,8 @@ static void assert_retrieval(const char *glob, int expected_count)
 
 void test_refs_foreachglob__retrieve_all_refs(void)
 {
-	/* 12 heads (including one packed head) + 1 note + 2 remotes + 7 tags */
-	assert_retrieval("*", 22);
+	/* 12 heads (including one packed head) + 1 note + 2 remotes + 7 tags + 1 blob */
+	assert_retrieval("*", 23);
```
pks-t: Please excuse my ignorance, but I really don't get why there's any changes to either refs::foreachglob or refs::iterator. As far as I can see, both these test suites act on a repository that has been opened via `git_repository_open`. No cloning took place and thus no negotiation via smart protocol. But the code changes you do is only in the smart protocol, so I'm wondering why values in either of those test suites should have changed at all?

ethomson: Good question - I agree.  I think it was a cherry-pick from that other PR, and that introduced a new ref?  I think that this doesn't fit here anymore.

tiennou: Note that I got a green build locally, and was surprised to have it come back red on CI (looking into it). AFAICS this test only adds one of those weird tags-to-non-commits, thus ensuring we correctly ignore it in whatever place we do not (😉) ; hence I'd vote to keep it. I can try adding a "clone this repo" in another testcase to exercise the actual fix to the smart transport.

pks-t: But there is no file added to any test repo, so the reference counts couldn't have changed.

ethomson: Right.  Since `git` looks for the presence of file named `HEAD` to determine whether a directory is a repository or not, our test resources always appear as actual repositories.  So you've probably still got a dirty test directory (with those refs) but you don't know it until you actually look in there.

This has bothered me for some time.

Maybe we should treat `HEAD`s like `.git` files/folders and rename them during test resource copying, to prevent this problem.

tiennou: That was it, sorry about that. GitX sometimes eats `add` entries when building amend indexes, and it slipped out as an untracked file. This should be good now.

pks-t: Nothing to be sorry about! :)

src/transports/smart_protocol.c
```diff
@@ -270,50 +270,6 @@ static int store_common(transport_smart *t)
...
-	git_reference *ref = NULL;
-	int error;
-
-	if ((error = git_reference_list(&refs, repo)) < 0)
-		return error;
-
-	if ((error = git_revwalk_new(&walk, repo)) < 0)
-		return error;
-
-	git_revwalk_sorting(walk, GIT_SORT_TIME);
```
pks-t: By the way, one more question: previously, we've been sorting by time, now we aren't. While obvious that this will greatly improve performance as we now don't have to go through all commits, it's not immediately obvious to me why this is fine and didn't find anything in the commit message.

tiennou: True, I took @carlosmn's comment about blanket-replacing at face-value 😉. Checking around, [pack-protocol](https://github.com/git/git/blob/master/Documentation/technical/pack-protocol.txt#L268) seems to imply there's no expected order so this seems fine — though a second opinion would be welcome.

biodranik:
Tested this change, it works fine. LGTM.

src/transports/smart_protocol.c
pks-t: Diving into the code a bit more. Previously, we've sorted all commits by date, which would cause us to compare the newest 256 commits present in our own repos with the HAVEs of the remote. Now with `push_glob`, we're pusing the commits in an unspecified order that is completely independent of the date. If one has a repo with more than 256 refs, then it may happen that we now compare 256 _old_ commits with the HAVEs. Heuristically, it's obvious that the first variant which adds commits by date may be a lot more efficient in the common case.

This analysis neglects that previously we had to load all commits of the repo to properly sort them by time, which was quite inefficient, too. I think we can do better than that, though: if `push_glob` were to use `git_commit_list_insert_by_date`, then we'd avoid inserting commits in a suboptimal order but we still wouldn't suck in all the repo's commit objects.

So I'd propose to modify `push_glob` to accept another parameter that determines how to sort pushed commits. `push_glob` will pass than on to `push_commit`, which will either use `git_commit_list_insert` or `git_commit_list_insert_by_date`.

tiennou: Thanks for the in-depth analysis @pks-t. I've amended the PR by API-fying a bit more (DRYing all the functions), so that the revwalk could be tweaked accordingly. Not sure if it's worth making this API public instead of the current one — no idea how useful it would be to "outsiders" — so it's a little rough around the edges (a single `flags` might have been sufficient).

tiennou: (though re-reading I'm now unsure if the insertion was to happen from an "initially sorted revwalk", as previously, or if it's okay with just the insertion being done by commit date. Right now the latter is done).

src/revwalk.h
```diff
@@ -50,4 +50,22 @@ struct git_revwalk {
...
+typedef struct {
+	int uninteresting;
+	int from_glob;
+} git_revwalk__schedule_options;
+
+#define GIT_REVWALK__SCHEDULE_OPTIONS_INIT { 0 }
+
+int git_revwalk__schedule_commit(git_revwalk *walk,
+								 const git_oid *oid,
+								 const git_revwalk__schedule_options *opts);
```
pks-t: Indentation of these functions looks a bit funny on GitHub

src/revwalk.c
```diff
@@ -83,7 +83,9 @@ int git_revwalk__schedule_commit(git_revwalk *walk, const git_oid *oid, const gi
 
 	commit->uninteresting = opts->uninteresting;
 	list = walk->user_input;
-	if (git_commit_list_insert(commit, &list) == NULL) {
+	if ((opts->insert_by_date &&
+		 git_commit_list_insert_by_date(commit, &list) == NULL) ||
```
pks-t: Sorry to be a nag, but there's another leading space here. While at it, you might also just convert it to use our currently accepted coding style of indenting subsequent if lines by four spaces, only

src/revwalk.c
```diff
@@ -38,7 +38,7 @@ git_commit_list_node *git_revwalk__commit_lookup(
 	return commit;
 }
 
-static int push_commit(git_revwalk *walk, const git_oid *oid, int uninteresting, int from_glob)
+int git_revwalk__schedule_commit(git_revwalk *walk, const git_oid *oid, const git_revwalk__schedule_options *opts)
```
pks-t: Hum. I'm definitely happy to expose these functions internally. But even though I'd prefer `schedule` in hindsight, we are already set in stone because of `git_revwalk_push` is exposed to users of libgit2 already. And mixing `schedule` and `push` if both in fact do the same thing is going to confuse people :/ So unfortunately, I'd prefer to stick with `git_revwalk__push_*`.

src/revwalk.h
tiennou: Fixed.

src/revwalk.c
tiennou: No problem, I forgot the obligatory "correct Xcode indenter" step 😉.

src/revwalk.c
tiennou: Agreed, it doesn't feel that useful to warrant a change, so let's wait for a "bigger impact change" (say, parallel revwalking) to revisit the public API/deprecate things. I've reverted to the current "push" naming.

src/revwalk.c
```diff
@@ -38,7 +38,7 @@ git_commit_list_node *git_revwalk__commit_lookup(
 	return commit;
 }
 
-static int push_commit(git_revwalk *walk, const git_oid *oid, int uninteresting, int from_glob)
+int git_revwalk__push_commit(git_revwalk *walk, const git_oid *oid, const git_revwalk__push_options *opts)
```
pks-t: We should probably fall back to default options if `opts == NULL`?

tiennou: I wanted the argument to be mandatory, mostly because the "mode" kinda is (OID vs glob, else it's ambiguous what should happen w.r.t to peeling). But I forgot to assert that though 😉.

pks-t: Fair enough, it's internal anyway so all callers can be checked trivially. Let's just go ahead

pks-t:
Thanks a lot, @tiennou!

PR #5196, opened 2019-08-01
config: implement "onbranch" conditional

pks-t:
With Git v2.23.0, the conditional include mechanism gained another new
conditional "onbranch". As the name says, it will cause a file to be
included if the "onbranch" pattern matches the currently checked out
branch.

Implement this new condition and add a bunch of tests.


csware:
Should we wait for Git 2.23 to be released before merging?

pks-t:
> Should we wait for Git 2.23 to be released before merging?

Well, the master branch may hold pre-release code and thus I'd say no, especially considering that git.git has already pulled release candidates containing these new conditions. Furthermore, it's not like there's any change in behaviour for pre-existing repos without explicitly making use of this feature.

But I'm biased here due to being the author, so maybe @ethomson should weigh in here.

pks-t:
On Thu, Aug 08, 2019 at 01:09:06AM -0700, Patrick Steinhardt wrote:
> > Should we wait for Git 2.23 to be released before merging?
> 
> Well, the master branch may hold pre-release code and thus I'd
> say no, especially considering that git.git has already pulled
> release candidates containing these new conditions.
> Furthermore, it's not like there's any change in behaviour for
> pre-existing repos without explicitly making use of this
> feature.
> 
> But I'm biased here due to being the author, so maybe @ethomson
> should weigh in here.

Doesn't matter now anymore as v2.23 has been released by now, so
the feature is officially out there :)

PR #5198, opened 2019-08-02
azure: build Docker images as part of the pipeline

pks-t:
The Docker images used for our continuous integration builds currently
live in the libgit2/libgit2-docker repository. To make any changes in
them, one has to make a PR there, get it reviewed, re-build the images
and publish them to Docker Hub. This process is slow and tedious, making
it harder than necessary to perform any updates to our Docker-based
build pipeline.

To fix this, we include all Dockerfiles used by Azure from the mentioned
repository and inline them into our own repo. Instead of having to
manually push them to the CI, it will now build the required containers
on each pull request, allowing much greater flexibility.

---

This is more of a test whether this is feasible at all with regards  to build times. It'd be rad if it was possible as that'd reduce another maintenance burden.


pks-t:
Sure enough, upgrading to Ubuntu Xenial uncovered multiple bugs in Valgrind, which will simply fail all the time due to a bug in Vex: https://bugs.launchpad.net/ubuntu/+source/valgrind/+bug/1574437 and a bug with libgcrypt. Other than that this works as expected now

pks-t:
After a whole lot of tries this finally builds. It was more problematic than expected, but this is mostly caused by upgrading our CI from Trusty (which has been EOL'd in April 2019) to Xenial. With Xenial came the usual set of bugs in dependencies we're already used to with Trusty, but I wanted to solve these without requiring Bintray. In fact, this development process already demonstrates why I'm apt to replace the current setup which uses an external Docker registry: It's so much easier to test changes to how the images are set up, to iterate and maintain.

So here's the catch: this PR adds about 1:50m to build the Docker images. BUT: It also shaves off nearly two minutes of build time by converting to Ninja. So we are actually a tiny bit faster than before. While we could decrease build time by two minutes if we were to continue using the registry and switch to Ninja regardless, I think that the benefits of building ad-hoc images greatly overweigh saving two minutes of build time. Especially considering that the slow path is still Windows or rather MinGW.

pks-t:
Oh, by the way: this PR possibly also fixes Coverity. Coverity builds have been referencing libgit2/trusty-openssl images, and I don't think we generate these anymore. So I've switched it to use the new infrastructure as well, which could possibly fix it.

ethomson:
Neat, nice to see the reduction in time from using Ninja.  Curious, though:  what if instead of regenerating the docker image every time as part of libgit2's CI, we most that CI bit out into its own stage that only runs on changes to the Dockerfiles.  IOW, we have a CI process for the docker images that builds and deploys to dockerhub.  That way we only regenerate them when they've changed and we keep that 2 minutes of savings that you shaved off by using Ninja?

That's sort of why I had dockerfiles in their own repo that we might do that someday.  (Though, to be honest, that's not necessary, we could use path scopes to avoid rebuilding things, but I think it simplifies things.)

pks-t:
I thought about that too, but rather as a nice optimization after this is merged (if we agree on it, that is). I'd really prefer to keep the Dockerfiles in here, though. Having a second repository adds just too much hassle, if you ask me:

- Repos need need to kept in sync
- More management overhead due to PRs in multiple repos. As a result, PRs in the other repo do not get reviewed as "quickly" as we do here (yeah, take "quick" with multiple grains of salt).
- No atomic changes possible
- Harder to discover and understand
- One cannot test changes to the CI in sync with changes to libgit2 without having access to both repos
- In the future, there might be maintenance branches that require different CI setups, which would require additional logic

I'd love to have a cache here, though. I think I saw that one can enable Docker registries with Azure, and I'd imagine that we could just push images in there. If possible, that would be a very nice improvement. We needn't care about tags in there, though, all builds need to be cached based on Dockerfile contents only, otherwise we'd have problems if e.g. a PR makes changes to a Dockerfile.

pks-t:
Rebased. I'd still be in favor of merging this :) Another usecase is the PCRE/PCRE2 conversion, where I'd love to simply install libpcre2 to have it tested as part of our CI

ethomson:
So, I would say that broadly speaking I'm :+1: on this.  In a perfect world, we wouldn't re-build the containers each time and would cache them.  But that's not something that we need to block this on.

pks-t:
:tada: :tada: :tada: Thanks a lot, @ethomson! :)

Issue #5133, opened 2019-06-21
packbuilder performance/connection timeout


cschlack:
I'm not able to push some big repositories to GitHub using libgit2 -> it's failing with a network error. Is there a way to accelerate the packbuilder and/or to prevent timeouts on the network?

### Reproduction steps

code see https://gist.github.com/cschlack/bbc5ff9afa436ab9db3a3f5add5bdaae
test-repository (one branch, one commit, 24 files, 384MB):
https://github.com/cschlack/beautiful-noise

### Expected behavior

```
> git remote add origin git@github.com:cschlack/newrepo.git
> git push --set-upstream origin master
Enumerating objects: 26, done.
Counting objects: 100% (26/26), done.
Delta compression using up to 4 threads
Compressing objects: 100% (26/26), done.
Writing objects:  0% (0/26), [...]
```
Transmission starts on my machine after 3,5minutes.

### Actual behavior

Output:
```
> remote_push beautiful_noise.git git@github.com/cschlack/newrepo
packbuilder stage 0:  1/0  - 3s
packbuilder stage 1:  0/26 - 3s
[...]
packbuilder stage 1: 24/26 - 412s
packbuilder stage 1: 25/26 - 434s
packbuilder stage 1: 26/26 - 434s
writing objects: 0% (0/26), 12 bytes - 434s
[...]
```

It takes > 7 minutes before transmitting starts.

With an even bigger repository:

```
remote_push even_bigger_repo.git git@github.com/cschlack/newrepo
[...]
packbuilder stage 1: 131/132 - 1079s
packbuilder stage 1: 132/132 - 1107s
SSH could not write data: Failure while draining incoming flow (23)
```

The connection times out!
After > 15 minutes libgit2 exits with a network error.

### Version of libgit2 (release number or SHA1)

libgit2 v0.28.2

### Operating system(s) tested

linux, MacOS

stmllr:
confirmed with Rugged:
Using a clone from https://github.com/cschlack/beautiful-noise, when pushing it to an empty GitHub repository it ends up with SSL error.

``` ruby
require 'rugged'

path = 'beautiful-noise.git'
url = 'https://github.com/stmllr/big-repo.git'
credentials = Rugged::Credentials::UserPassword.new(username: 'USER_TOKEN', password: '')
git = Rugged::Repository.bare(path)
remote = git.remotes.create_anonymous(url)
remote.push(git.refs.map(&:name), credentials: credentials)

Traceback (most recent call last):
[...]

PR #5201, opened 2019-08-09
smart: implement keepalive

pks-t:
This is a first proposal on how to implement keepalive for the smart backend as documented in git-config(1), "receive.keepAlive". It was required to change both stream as well as transport backends and extend them with a `poll` function to support timeouts, but in fact it was done in an API-backwards-compatible way: if no `poll` is implemented, we simply assume that there's always data to be read and thus cannot support keepalive for such backends.

Note that this is rather to be seen as an RFC for now. The `poll` function would probably need a `p_poll` wrapper, which I haven't yet done. To make this clearer, I've added `#include <poll.h>` in the middle of the source code.

I think the most important questions I have is: do we want to use `poll`? Or would it be preferred to have a `read_ext` function that accepts a timeout parameter, as well?

Inspired by #5133.

### DIFF
--- a/include/git2/sys/stream.h
+++ b/include/git2/sys/stream.h
@@ -36,6 +36,7 @@ typedef struct git_stream {
 	int GIT_CALLBACK(set_proxy)(struct git_stream *, const git_proxy_options *proxy_opts);
 	ssize_t GIT_CALLBACK(read)(struct git_stream *, void *, size_t);
 	ssize_t GIT_CALLBACK(write)(struct git_stream *, const char *, size_t, int);
+	int GIT_CALLBACK(poll)(struct git_stream *, int);
 	int GIT_CALLBACK(close)(struct git_stream *);
 	void GIT_CALLBACK(free)(struct git_stream *);
 } git_stream;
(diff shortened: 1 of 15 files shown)


ethomson:
> I think the most important questions I have is: do we want to use poll? Or would it be preferred to have a read_ext function that accepts a timeout parameter, as well?

That's a good question.  Our networking code is largely blocking and has no timeouts.  I think that we should just include a `poll` now as-is and that if we want to add timeouts that we should look holistically at where we're at rather than adding it in just this one place.

pks-t:
On Mon, Aug 12, 2019 at 03:34:02AM -0700, Edward Thomson wrote:
> > I think the most important questions I have is: do we want to
> > use poll? Or would it be preferred to have a read_ext
> > function that accepts a timeout parameter, as well?
> 
> That's a good question.  Our networking code is largely
> blocking and has no timeouts.  I think that we should just
> include a `poll` now as-is and that if we want to add timeouts
> that we should look holistically at where we're at rather than
> adding it in just this one place.

Fair enough, I'll polish up this PR then. Most importantly, we
should let the `poll` callback functions accept some flags like
`POLL_IN` and `POLL_OUT`, otherwise we're painting us into a
corner again.

PR #5202, opened 2019-08-13
Security updates from 0.28.3

ethomson:
Security updates ported from the 0.28.3 branch:

* A carefully constructed commit object with a very large number
of parents may lead to potential out-of-bounds writes or
potential denial of service.

* The ProgramData configuration file is always read for compatibility
with Git for Windows and Portable Git installations. The ProgramData
location is not necessarily writable only by administrators, so we
now ensure that the configuration file is owned by the administrator
or the current user.


ddevault:
Has this been assigned a CVE?

ethomson:
CVE-2019-1211

ddevault:
Thanks!

PR #5204, opened 2019-08-14
clone: Test whitespace in url over ssh

ianhattendorf:
See #5187.


pks-t:
From #5187:

>> Would it be possible to get SSH working on the spaces in the name repo?

> Hmm, in a perfect world, we would do this indeed. But that would mean setting up an SSH public key in an actual user account in Azure DevOps. Even if that user was set to read-only for repositories, I'm a little bit loathe to set that up without thinking through it more deeply.

So the issue is that we don't have an SSH key set up in Azure and thus we cannot test this properly, did I get that right? If so, couldn't we just use the local SSH daemon we're spinning up in CI and create a repo with spaces ourselves?

PR #5206, opened 2019-08-17
CMake pkg-config modulification

### LINKED ISSUE #5203: PCRE dependency information is not stored in libgit2.pc
### Reproduction steps

Configure and compile libgit2 such that it decides to link dynamically against libpcre.

### Expected behavior

This fact should be stored in `libgit2.pc` in case we want to statically include the library.

### Actual behavior

It is not there

Current master (08cfa43d0e1a9214) presumably since the PCRE functionality was introduced.

Debian bullseye, Ubuntu Xenial.

tiennou:
This is the CMake module I wrote for mbedTLS, which tries to streamline writing `pkg-config` files, in an attempt to fix #5203.

This is incomplete (I think), but I'll need guidance, because macOS usually doesn't really care about *.pc files anyways.

Here's a [gist](https://gist.github.com/tiennou/a943861bc8d028178a7f57827f7a72e3/revisions) of the current changes introduced, at least on macOS. I do think the `Libs.private` looks "saner", but the `Requires.private` isn't there yet.


cmake/Modules/PkgBuildConfig.cmake
```diff
@@ -0,0 +1,101 @@
...
+        "Description: ${PKGCONFIG_DESCRIPTION}\n"
+        "Version: ${PKGCONFIG_VERSION}\n"
+    )
+
+    if(NOT DEFINED PKGCONFIG_LIBS_SELF)
+        set(PKGCONFIG_LIBS_SELF "${PKGCONFIG_FILE}")
+    endif()
+
+    if(NOT DEFINED PKGCONFIG_LIBS)
+        set(PKGCONFIG_LIBS "-l${PKGCONFIG_LIBS_SELF}")
```
pks-t: This seems a bit weird to me. So you're setting the PKGCONFIG_LIBS variable to the filename of the pkgconfig file?

tiennou: It's part of a multi-step fallback (which is obviously undocumented :wink:). The goal is to have sane defaults (ie. the project, dylib & pkg-config identifier are all equal), but allow customization. So `PKGCONFIG_LIBS` is the underlying, substitued-in value, but we need to preserve what's in it (eg. mbedtls has a dependency on pthreads, which isn't covered by pkg-config), we have somewhat non-matching names — we're known as `libgit2` to pkg-config, but our `LIBS_SELF` name for linking is `-lgit2`.

In short, it's the reverse, the filename defaults to the pkg-config `NAME`, unless `FILENAME` is passed. The `-l$self` flag defaults to `SELF_LIBS`, `FILENAME` or `NAME`, in that order.

I *could* have considered that `NAME` must be a known target (which would allow to `GET_TARGET_PROPERTY`), but I didn't. It's just a dump "collect, customize, and write" wrapper.

pks-t: Fair enough, thanks for explaining! :)

tiennou:
Okay, I've added some quoting in the hope that would be sufficient 😉. I've updated the gist with an OpenSSL, non-PCRE run, as a syntax check.

pks-t:
Tried it on my own system and everything looked good to me. Thanks a lot, @tiennou!

PR #5208, opened 2019-08-20
apply: git_apply_to_tree fails to apply patches that add new files

mkostyuk:
git_apply_to_tree() cannot be used to apply patches with new files.

An attempt to apply such a patch fails because git_apply_to_tree() tries to
remove a non-existing file from an old index.

The solution is to modify git_apply_to_tree() to git_index_remove() when the
patch states that the modified files is removed.

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -649,9 +649,12 @@ int git_apply_to_tree(
 	for (i = 0; i < git_diff_num_deltas(diff); i++) {
 		delta = git_diff_get_delta(diff, i);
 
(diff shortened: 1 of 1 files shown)


ethomson:
This definitely makes sense and looks correct.  We should have a test that exercises this, though.

mkostyuk:
Hi Edward. Could you please check if the test case is okay? It looks okay for me, but maybe I missed something obvious. Thank you.

mkostyuk:
Thank you.

PR #5209, opened 2019-08-20
apply: Fix a patch corruption related to EOFNL handling

mkostyuk:
Use of apply's API can lead to an improper patch application and a corruption
of the modified file.

The issue is caused by mishandling of the end of file changes if there are
several hunks to apply. The new line character is added to a line from a wrong
hunk.

The solution is to modify apply_hunk() to add the newline character at the end
of a line from a right hunk.

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -219,7 +219,7 @@ static int apply_hunk(
 			case GIT_DIFF_LINE_CONTEXT_EOFNL:
 			case GIT_DIFF_LINE_DEL_EOFNL:
 			case GIT_DIFF_LINE_ADD_EOFNL:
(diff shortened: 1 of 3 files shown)


pks-t:
Thanks a lot for your fix!

Could you please provide test data that has resulted in corruption previous to this change? Additional kudos if it comes as a unit test to verify that this actually fixes the issue and that we won't regress in the future. It's perfectly fine though if you provide the reproduction steps, only, then I'd help you out and implement a unit test for you :)

mkostyuk:
Hi Patrick, could you please check the unit test? Thank you.

ethomson:
Thanks @mkostyuk!

PR #5210, opened 2019-08-20
ignore: correct handling of nested rules overriding wild card unignore

### LINKED ISSUE #4963: Negate patterns in .gitignore

buddyspike:
problem:
filesystem_iterator loads .gitignore files in top-down order.
subsequently, ignore module evaluates them in the order they are loaded.
this creates a problem if we have unignored a rule (using a wild card)
in a sub dir and ignored it again in a level further below (see the test
included in this patch).

solution:
process ignores in reverse order.

closes #4963

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -444,7 +444,7 @@ static bool ignore_lookup_in_rules(
 int git_ignore__lookup(
 	int *out, git_ignores *ignores, const char *pathname, git_dir_flag dir_flag)
 {
-	unsigned int i;
+	size_t i;
 	git_attr_file *file;
 	git_attr_path path;
 
@@ -458,8 +458,11 @@ int git_ignore__lookup(
 	if (ignore_lookup_in_rules(out, ignores->ign_internal, &path))
 		goto cleanup;
 
-	/* next process files in the path */
-	git_vector_foreach(&ignores->ign_path, i, file) {
+	/* next process files in the path.
+	 * this process has to process ignores in reverse order
+	 * to ensure correct prioritization of rules
+	 */
+	git_vector_rforeach(&ignores->ign_path, i, file) {
 		if (ignore_lookup_in_rules(out, file, &path))
 			goto cleanup;
 	}
--- a/tests/ignore/status.c
+++ b/tests/ignore/status.c
@@ -1290,3 +1290,47 @@ void test_ignore_status__leading_spaces_are_significant(void)
 	assert_is_ignored(" # not a comment");
 	assert_is_ignored("d.test");
 }
+
+void test_ignore_status__override_nested_wildcard_unignore(void)
+{
+	git_repository *repo = cl_git_sandbox_init("empty_standard_repo");
+	git_status_list *statuslist;
+	git_status_options opts = GIT_STATUS_OPTIONS_INIT;
+	const git_status_entry *status;
+
+	cl_git_pass(git_futils_mkdir_r("empty_standard_repo/dir", 0777));
+	cl_git_pass(git_futils_mkdir_r("empty_standard_repo/dir/subdir", 0777));
+	cl_git_mkfile("empty_standard_repo/.gitignore", "a.test\n");
+	cl_git_mkfile("empty_standard_repo/dir/.gitignore", "!*.test\n");
+	cl_git_mkfile("empty_standard_repo/dir/subdir/.gitignore", "a.test\n");
+	cl_git_mkfile("empty_standard_repo/dir/a.test", "pong");
+	cl_git_mkfile("empty_standard_repo/dir/subdir/a.test", "pong");
+
+	opts.show = GIT_STATUS_SHOW_INDEX_AND_WORKDIR;
+	opts.flags = GIT_STATUS_OPT_INCLUDE_UNTRACKED | GIT_STATUS_OPT_RECURSE_UNTRACKED_DIRS;
+
+	cl_git_pass(git_status_list_new(&statuslist, repo, &opts));
+	cl_assert_equal_sz(4, git_status_list_entrycount(statuslist));
+
+	status = git_status_byindex(statuslist, 0);
+	cl_assert(status != NULL);
+	cl_assert_equal_s(".gitignore", status->index_to_workdir->old_file.path);
+	cl_assert_equal_i(GIT_STATUS_WT_NEW, status->status);
+
+	status = git_status_byindex(statuslist, 1);
+	cl_assert(status != NULL);
+	cl_assert_equal_s("dir/.gitignore", status->index_to_workdir->old_file.path);
+	cl_assert_equal_i(GIT_STATUS_WT_NEW, status->status);
+
+	status = git_status_byindex(statuslist, 2);
+	cl_assert(status != NULL);
+	cl_assert_equal_s("dir/a.test", status->index_to_workdir->old_file.path);
+	cl_assert_equal_i(GIT_STATUS_WT_NEW, status->status);
+
+	status = git_status_byindex(statuslist, 3);
+	cl_assert(status != NULL);
+	cl_assert_equal_s("dir/subdir/.gitignore", status->index_to_workdir->old_file.path);
+	cl_assert_equal_i(GIT_STATUS_WT_NEW, status->status);
+
+	git_status_list_free(statuslist);
+}


src/ignore.h
```diff
@@ -51,6 +52,10 @@ enum {
 	GIT_IGNORE_TRUE = 1,
 };
 
+enum {
+	GIT_IGNORE_TOP_DOWN_DISCOVERY = (1u << 0),
+};
```
pks-t: I'm wondering whether this new flag is needed at all. We are unconditionally setting it for the filesystem iterator and thus we'll always use it for discovering ignores now, right? If so, then I'd say to not introduce the flag but just convert to use `git_vector_rforeach`

src/ignore.c
```diff
@@ -458,10 +458,25 @@ int git_ignore__lookup(
...
+	 * this process has to consider the order that .ignore
+	 * files were discovered to ensure correct prioritization of rules
+	 */
+	if (ignores->flags & GIT_IGNORE_TOP_DOWN_DISCOVERY)
+	{
+		git_vector_rforeach(&ignores->ign_path, i, file) {
+			if (ignore_lookup_in_rules(out, file, &path)) {
+				goto cleanup;
+			}
+		}
```
pks-t: I think this makes sense to me. There cannot be an ignore rule that is invalidated by its parent rules, but the reverse can be true as demonstrated by your tests. The downside is that we need to always go all the way to the top for files that were ignored in the top-level directory, where we could previously immediately abort in the first iteration.

Naturally, correctness outweighs performance though.

tests/status/worktree.c
```diff
@@ -1315,3 +1315,47 @@ void test_status_worktree__at_head_parent(void)
 	git_tree_free(parent_tree);
 	git_status_list_free(statuslist);
 }
+
+void test_status_worktree__override_nested_wildcard_unignore(void)
```
pks-t: You could've saved yourselve quite some pain by using "tests/ignore/status.c" or "tests/ignore/path.c", but somehow nobody discovers that test suite :P It's fine if you don't want to go the extra mile and convert it.

src/ignore.h
ethomson: Right, I was wondering same.  Would it ever be correct to evaluate them bottom up instead of top down?  Seems like no.

tests/status/worktree.c
buddyspike: Sorry, I saw these tests and the utility methods available. I will move them 😉

src/ignore.h
buddyspike: Thanks for the review and your comments. Using `git_vector_rforeach` was my first choice too. Then I found out that `git_ignore_path_is_ignored` api walks the tree up to discover ignores (https://github.com/libgit2/libgit2/blob/master/src/ignore.c#L327). Please let me know if my understanding is incorrect.

src/ignore.c
buddyspike: We could try to work out the starting `ignore` by taking a minimum of current depth or length of `ignores`. What do you think?

tests/status/worktree.c
buddyspike: @pks-t I moved the test to `tests/ignore/status.c`. However, I could not use utilities like `assert_is_ignored` because it does not go through `filesystem_iterator`. I think this is the reason why I left the tests in `worktree.c` initially.

src/ignore.h
ethomson: Is that correct, though?  I feel like you've discovered that we're walking up and shouldn't be.  It seems like that's true here, too?

buddyspike: I think it is fine to walk up in `git_ignore__for_path`. Otherwise, it would not discover `.gitignore` in a parent directory right? 😉

buddyspike: Thanks, @ethomson.

From what I can gather, `git_ignore__for_path` is a lower-level function used in a couple of places.  Given a path, it should return `ignores` applicable to that path. For example, when the input path is `a/b/main.c`, it should discover `.gitignore` files in both `a` and `b`.

Since the starting point is `a/b/main.c`, it makes sense to walk up the tree looking for ignores. I think this is the optimal way to do this in this scenario. Because if we start discovering ignore files from `a` here, we would have to put additional checks to traversing down unwanted paths (e.g. imagine other directories such as `a/c`).

One place `git_ignore__for_path` is used is a public API called `git_ignore_path_is_ignored` (https://libgit2.org/libgit2/#HEAD/group/ignore/git_ignore_path_is_ignored). We should not break this functionality.

`filesystem_iterator` on the other hand starts things off from the root (it actually uses `git_ignore__for_path` to initialize the root level, but that largely irrelevant I guess). As it travels down the tree, it loads up `.gitignore` files on its way. It also uses the ignores discovered so far to decide whether to descend or not.

So given the two patterns in use, IMHO, it's valid to have two ways to discover ignore files.

Hopefully, this clears it up. Thanks a lot again.

buddyspike: >  I think this is the optimal way to do this in this scenario. Because if we start discovering ignore files from a here, we would have to put additional checks to traversing down unwanted paths (e.g. imagine other directories such as a/c).

Ignore this part 😄 
I think we can change this function to do a top-down lookup.

tests/status/worktree.c
pks-t: Fair enough!

src/ignore.h
pks-t: Sorry, I still don't quite get it and I've long given up hope of completely understanding our gitignore logic. At some point in time I feel like we should just rewrite it altogether, I can't imagine git.git's to be this... convoluted.

Anyway. I've just did a quick test to replace `git_vector_foreach` with `git_vector_rforeach` unconditionally and didn't get a single test failure. So either it's fine to _always_ iterate in reverse order indicating we don't need the flag. Or alternatively we're missing a test case for the szenario you're describing.

I hate to pile up more work on your plate, but could you please try to create a testcase for the szenario where your new flag is relevant? I guess that'd help both @ethomson and myself greatly.

Thanks a lot!

buddyspike: @pks-t lolz!

It's funny that it didn't fail for you. When I did it, there were a few breaking tests. Let me try that again and perhaps I can give up on completing the tests for `git_path_walk_down` function I just finished writing 😉 .

buddyspike: @pks-t you are right! Tests are passing now. Something must have changed in master since attempted this approach. Here's a quick test to show that applying `git_vector_rforeach` is not sufficient.

```
void test_ignore_status__unignore_in_subdir(void)
{
	static const char *test_files[] = {
		"empty_standard_repo/subdir/a.test",
		"empty_standard_repo/a.test",
		NULL
	};

	make_test_data("empty_standard_repo", test_files);
	cl_git_mkfile(
		"empty_standard_repo/.gitignore",
		" a.test\n");

	cl_git_mkfile(
		"empty_standard_repo/subdir/.gitignore",
		"a.test\n"
		"!a.test\n");

	assert_is_ignored("a.test");
	refute_is_ignored("subdir/a.test");
}

```

Could you please give it whirl and let me know. In the meantime, I will continue with my change to this PR.

Thanks a lot again for your support!

buddyspike: @pks-t sorry my bad, above tests works. There was an extra space in a rule 😊 
I will make the adjustments to this PR.

buddyspike: @pks-t & @ethomson I've made the requested change now. I tried running tests with this patch in a few tags but all of them work. Not sure why tests failed when I first tried it. Sorry about the lengthy dialog. Thanks again!

buddyspike:
@pks-t & @ethomson,
I double-checked the code involved in two scenarios and they are independent. In fact, it has been like that for a long time. I must have observed failing tests due to a local change. Sorry about the confusion and let me know if there's anything else I can do here to help.

ethomson:
Seems correct to me.  Thanks for the work here @buddyspike!  The ignore functionality is deeply tricky so I'm glad you were able to fix it up for us.

PR #5212, opened 2019-08-23
Use an HTTP scheme that supports the given credentials

### LINKED ISSUE #5178: Assertion failure when remote proxy supports both SPNEGO/Negotiate and NTLM
# Reproduction steps
`./libgit2_clar -sonline::clone::proxy` with `GITTEST_REMOTE_PROXY{HOST,USER,PASS}` set pointing to a proxy that supports both SPNEGO and NTLM (in my case, squid 4).

### Expected behavior
Tests pass.

### Actual behavior
```
% ./libgit2_clar -vv -sonline::clone::proxy                                                                        ✭
Loaded 374 suites:

libgit2_clar: ../src/transports/auth_negotiate.c:90: negotiate_next_token: Assertion `buf && ctx && ctx->configured && cred && cred->credtype == GIT_CREDTYPE_DEFAULT' failed.
[1]    15811 abort (core dumped)  ./libgit2_clar -vv -sonline::clone::proxy
```

master (a33c0de2edace210ccb99f2dc886971b8fa68382)

Linux & Mac

It looks like in `http.c`, `init_auth` relies on `scheme_for_challenge` to init the server `auth_context`, which doesn't take into account whether `server->cred->credtype` is compatible with that scheme. I attempted to fix it below, but am still failing a couple tests and I'm sure I'm violating some preconditions.

```
diff --git a/src/transports/http.c b/src/transports/http.c
index d727851b7..55030b06d 100644
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -129,17 +129,19 @@ typedef struct {
        size_t *bytes_read;
 } parser_context;

-static git_http_auth_scheme *scheme_for_challenge(const char *challenge)
+static git_http_auth_scheme *scheme_for_challenge(const char *challenge, git_cred *cred)
 {
        git_http_auth_scheme *scheme = NULL;
        size_t i;

        for (i = 0; i < ARRAY_SIZE(auth_schemes); i++) {
                const char *scheme_name = auth_schemes[i].name;
+               const git_credtype_t scheme_types = auth_schemes[i].credtypes;
                size_t scheme_len;

                scheme_len = strlen(scheme_name);
-               if (strncasecmp(challenge, scheme_name, scheme_len) == 0 &&
+               if ((!cred || (cred->credtype & scheme_types)) &&
+                   strncasecmp(challenge, scheme_name, scheme_len) == 0 &&
                    (challenge[scheme_len] == '\0' || challenge[scheme_len] == ' ')) {
                        scheme = &auth_schemes[i];
                        break;
@@ -256,7 +258,7 @@ static int set_authentication_types(http_server *server)
        size_t i;

        git_vector_foreach(&server->auth_challenges, i, challenge) {
-               if ((scheme = scheme_for_challenge(challenge)) != NULL) {
+               if ((scheme = scheme_for_challenge(challenge, NULL)) != NULL) {
                        server->authtypes |= scheme->type;
                        server->credtypes |= scheme->credtypes;
                }
@@ -432,7 +434,7 @@ static int init_auth(http_server *server)
        size_t i;

        git_vector_foreach(&server->auth_challenges, i, c) {
-               s = scheme_for_challenge(c);
+               s = scheme_for_challenge(c, server->cred);

                if (s && !!(s->credtypes & server->credtypes)) {
                        scheme = s;
```

ethomson:
Don't try to select the "best" scheme that the server offers us without taking into account the credentials that the user has selected.  Otherwise, the "best" scheme may not support the given credentials.

eg, many servers use both Negotiate and NTLM as a fallback.  Negotiate _only_ works with default credentials, while our NTLM implementation _only_ works with username/password credentials.  As a result, our caller has given us a particular kind of credentials and we must match those credentials to the schemes offered.

Instead of using the "best" overall scheme, use the "best" scheme (ordered first) that our credentials support.

Fixes #5178

### DIFF
--- a/src/transports/auth.c
+++ b/src/transports/auth.c
@@ -70,6 +70,6 @@ int git_http_auth_dummy(
 	GIT_UNUSED(url);
 
 	*out = NULL;
-	return 0;
+	return GIT_PASSTHROUGH;
 }
 
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -430,6 +430,7 @@ static int init_auth(http_server *server)
 	git_http_auth_scheme *s, *scheme = NULL;
 	char *c, *challenge = NULL;
 	size_t i;
+	int error;
 
 	git_vector_foreach(&server->auth_challenges, i, c) {
 		s = scheme_for_challenge(c);
@@ -446,12 +447,14 @@ static int init_auth(http_server *server)
 		return -1;
 	}
 
-	if (scheme->init_context(&server->auth_context, &server->url) < 0)
-		return -1;
+	if ((error = scheme->init_context(&server->auth_context, &server->url)) == GIT_PASSTHROUGH)
+		return 0;
+	else if (error < 0)
+		return error;
 
 	if (server->auth_context->set_challenge &&
-		server->auth_context->set_challenge(server->auth_context, challenge) < 0)
-		return -1;
+		(error = server->auth_context->set_challenge(server->auth_context, challenge)) < 0)
+		return error;
 
 	return 0;
 }
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -129,17 +129,22 @@ typedef struct {
 	size_t *bytes_read;
 } parser_context;
 
-static git_http_auth_scheme *scheme_for_challenge(const char *challenge)
+static git_http_auth_scheme *scheme_for_challenge(
+	const char *challenge,
+	git_cred *cred)
 {
 	git_http_auth_scheme *scheme = NULL;
 	size_t i;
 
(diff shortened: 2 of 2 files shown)


ethomson:
Thanks @ianhattendorf for the patch, I made sure to list you as the author on the commit.

ethomson:
Also, @ianhattendorf, can you double-check this?  You mentioned in #5178 that you were seeing some test failures.  If you're still seeing them with this branch, can you give me more details?

ethomson:
I'm going to go ahead and merge this since it's strictly an improvement.  I'd be curious to follow up on your proxy problems, but at the moment, I can't recreate them.

PR #5213, opened 2019-08-24
Fix include casing for case-sensitive filesystems.

boardwalk:
This recently added include broke compilation on mingw32 with case-sensitive filesystems. Lowercasing here fixes compilation for me and mingw32 standardizes on lowercase include paths, so this seems like the right thing to do.


ethomson:
Looks good to me.  Thanks @boardwalk!

PR #5214, opened 2019-08-27
Memory allocation fixes for diff generator

pks-t:
Some more fixes inspired by recent crashes that have been reported


pks-t:
Thanks Ed!

PR #5216, opened 2019-08-29
ntlm: fix failure to find openssl headers

hackhaslam:
Building the NTLM module with OpenSSL chosen as the HTTPS backend results in file-not-found errors for OpenSSL headers. I gather that this was added in #5052. I'm not sure if this is the most appropriate fix, but it works for me.


ethomson:
Huh!  That's surprising - I'm curious if there's something different about your setup than our CI...  do you know if there's anything obviously different?

hackhaslam:
I suppose that the difference is that those instances are relying on the system openssl dev package which installs headers in `/usr/include`? I'm running cmake with a flag to get it to find my own openssl that I built elsewhere. I'm just guessing.

ethomson:
> I suppose that the difference is that those instances are relying on the system openssl dev package which installs headers in /usr/include? I'm running cmake with a flag to get it to find my own openssl that I built elsewhere. I'm just guessing.

Nice, thanks for the info.  I wonder if I'll need to do a similar thing for mbedTLS or something.  I'll play around a bit.

pks-t:
> Nice, thanks for the info. I wonder if I'll need to do a similar thing for mbedTLS or something. I'll play around a bit.

Yes, definitely, this will fail whenever the SSL implementation headers are in a location that is not part of the standard include search path. So the fix looks good to me and should be applied, probably with additional fixes for the other libraries. Happy to merge this single fix as a standalone, though.

PR #5224, opened 2019-09-10
open:fix memory leak when passing NULL to git_repository_open_ext

lrm29:
There is an example on the website for checking if a directory looks like a repository:

https://libgit2.org/docs/guides/101-samples/#repositories_openable

This example leaks memory. This pull request includes a test point which leaks memory and is caught by valgrind, plus a fix in git_repository_open_ext to dispose of buffers on the early return.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -800,7 +800,7 @@ int git_repository_open_ext(
 	unsigned is_worktree;
 	git_buf gitdir = GIT_BUF_INIT, workdir = GIT_BUF_INIT,
 		gitlink = GIT_BUF_INIT, commondir = GIT_BUF_INIT;
-	git_repository *repo;
+	git_repository *repo = NULL;
 	git_config *config = NULL;
 
 	if (flags & GIT_REPOSITORY_OPEN_FROM_ENV)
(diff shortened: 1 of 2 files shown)


src/repository.c
```diff
@@ -812,8 +812,13 @@ int git_repository_open_ext(
 	error = find_repo(
 		&gitdir, &workdir, &gitlink, &commondir, start_path, flags, ceiling_dirs);
 
-	if (error < 0 || !repo_ptr)
+	if (error < 0 || !repo_ptr) {
```
ethomson: Instead of adding duplicate `free`s here, I would encourage you to jump to the cleanup label so that we can consolidate cleanup in a single place.  That's the general pattern we try to use, it makes it less likely that we would end up freeing in one return path but not the other if we end up adding new allocations (for instance).

ethomson:
Nice catch.  A minor request for a stylistic tweak, if you don't mind.

src/repository.c
lrm29: That makes sense, I've updated the review. I initialised 'repo' to NULL in order to handle the cleanup case where error < 0 and repo_ptr is NULL.

ethomson:
Looks great!  Thanks for finding this and the helpful pull request.  _Love_ that you added a test for it!  😃

PR #5226, opened 2019-09-13
regexp: implement a new regular expression API

### LINKED ISSUE #5217: With PCRE2 selected we link against libc's functions causing crashes
# Reproduction steps

```
$ cmake . -DREGEX_BACKEND=pcre2
$ make
$ ./libgit2_clar -scheckout::index::honor_coresymlinks_default_true
```

### Expected behavior

Successful test run

### Actual behavior

Crash due to `malloc(): invalid size (unsorted)`

```
% nm libgit2_clar | grep regcomp
                 U regcomp@@GLIBC_2.2.5
```

5fc27aac6f47aaf06f454805f77c942c5a08393b

Debian testing

### Analysis

I have checked that we do run against the right headers and the final compilation step does try to link against pcre2. However, the `pcre2posix.h` header defines the basic names e.g. `regcomp` rather than `pcre2_regcomp` with a corresponding `#define` like pcre does.

In the pcre case, it is clear we are linking against pcre functions but it looks like in the pcre2 case, the linker is happy to find these functions in libc.

Often you can determine which version gets linked via the order of link arguments to the compiler, but I don't know if we can really put anything before libc.

An alternative would be to use the specific headers and `#define` the prefix into our method instead of relying on the header.

pks-t:
We currently support a set of different regular expression backends with
PCRE, PCRE2, regcomp(3P) and regcomp_l(3). The current implementation of
this is done via a simple POSIX wrapper that either directly uses
supplied functions or that is a very small wrapper.

To support PCRE and PCRE2, we use their provided <pcreposix.h> and
<pcre2posix.h> wrappers. These wrappers are implemented in such a way
that the accompanying libraries pcre-posix and pcre2-posix provide the
same symbols as the libc ones, namely regcomp(3P) et al. This works out
on some systems just fine, most importantly on glibc-based ones, where
the regular expression functions are implemented as weak aliases and
thus get overridden by linking in the pcre{,2}-posix library. On other
systems we depend on the linking order of libc and pcre library, and as
libc always comes first we will end up with the functions of the libc
implementation. As a result, we may use the structures `regex_t` and
`regmatch_t` declared by <pcre{,2}posix.h>, but use functions defined by
the libc, leading to segfaults.

The issue is not easily solvable. Somed distributions like Debian have
resolved this by patching PCRE and PCRE2 to carry custom prefixes to all
the POSIX function wrappers. But this is not supported by upstream and
thus inherently unportable between distributions. We could instead try
to modify linking order, but this starts becoming fragile and will not
work e.g. when libgit2 is loaded via dlopen(3P) or similar ways. In the
end, this means that we simply cannot use the POSIX wrappers provided by
the PCRE libraries at all.

Thus, this commit introduces a new regular expression API. The new API
is on a tad higher level than the previous POSIX abstraction layer, as
it tries to abstract away any non-portable flags like e.g. REG_EXTENDED,
which has no equivalents in all of our supported backends. As there are
no users of POSIX regular expressions that do _not_ reguest REG_EXTENDED
this is fine to be abstracted away, though. Due to the API being
higher-level than before, it should generally be a tad easier to use
than the previous one.

Note: ideally, the new API would've been called `git_regex_foobar` with
a file "regex.h" and "regex.c". Unfortunately, this is currently
impossible to implement due to naming clashes between the then-existing
"regex.h" and <regex.h> provided by the libc. As we add the source
directory of libgit2 to the header search path, an include of <regex.h>
would always find our own "regex.h". Thus, we have to take the bitter
pill of adding one more character to all the functions to disambiguate
the includes.

To improve guarantees around cross-backend compatibility, this commit
also brings along an improved regular expression test suite
core::regexp.

---

Supersedes #5218 
Fixes #5217


ethomson:
This looks like a good improvement to me.  Do you need help with the macOS and Windows failures?

pks-t:
On Fri, Sep 20, 2019 at 03:03:56AM -0700, Edward Thomson wrote:
> This looks like a good improvement to me.  Do you need help
> with the macOS and Windows failures?

I didn't yet have time to dig deeper. I'll have a look tomorrow
and let you know if I need any help, thanks :)

pks-t:
And another green PR :)

ethomson:
Thanks for doing this work @pks-t, what a nice improvement.

PR #5229, opened 2019-09-15
Update chat resources in README.md

ddevault:


pks-t:
@carlosmn: I haven't seen you in #libgit2 in a while. Would it be possible to either give me channel ops or to transfer ownership?

ethomson:
I'm not there often but I'd like `+o` as well for redundancy if nothing else.

pks-t:
> I'm not there often but I'd like +o as well for redundancy if nothing else.

Agreed, it's always best to have a high bus factor.

ddevault:
Thanks!

Issue #5199, opened 2019-08-03
diff: crash in git_diff_tree_to_index


eaigner:
This issue might be related to #5164
```
Incident Identifier: 5f2bcf63-db47-423b-bec2-2de5c7f44952
CrashReporter Key:   3C7B61AB-580D-4C3B-BC5F-E3BF898659E1
Hardware Model:      MacBookPro14,3
Process:         Gitfox [94801]
Path:            /Applications/Gitfox.app/Contents/MacOS/Gitfox
Identifier:      com.bytieful.Gitfox
Version:         2103 (2103)
Code Type:       x86_64
Parent Process:  launchd [1]

Date/Time:       undefined
Launch Time:     undefined
OS Version:      Mac OS X 10.14.6 (18G84)
Report Version:  104

Exception Type:  SIGABRT
Exception Codes: #0 at 0x7fff7e21e2c6
Crashed Thread:  0

Thread 0 Crashed:
0   libsystem_kernel.dylib               0x00007fff7e21e2c6 __pthread_kill + 10
1   libsystem_c.dylib                    0x00007fff7e1886a6 abort + 126
2   libsystem_malloc.dylib               0x00007fff7e297077 malloc_vreport + 544
3   libsystem_malloc.dylib               0x00007fff7e296e38 malloc_report + 150
4   Git                                  0x000000010bed8151 tree_iterator_free + 160
5   Git                                  0x000000010bed8446 git_iterator_free + 21
6   Git                                  0x000000010bebe05c git_diff_tree_to_index + 379
7   Git                                  0x000000010bebe609 git_diff_tree_to_workdir_with_index + 120
```

eaigner:
I have expanded the `DIFF_FROM_ITERATORS` snippet of doom for better readability here.

```
int git_diff_tree_to_index2(
                            git_diff **out,
                            git_repository *repo,
                            git_tree *old_tree,
                            git_index *index,
                            const git_diff_options *opts)
{
    git_diff *diff = ((void*)0);
    git_iterator_flag_t iflag = GIT_ITERATOR_DONT_IGNORE_CASE |
    GIT_ITERATOR_INCLUDE_CONFLICTS;
    _Bool index_ignore_case = 0;
    int error = 0;

    (__builtin_expect(!(out && repo), 0) ? __assert_rtn(__func__, "/Users/era/git/libgit2/src/diff_generate.c", 1346, "out && repo") : (void)0);

    *out = ((void*)0);

[...]

PR #5232, opened 2019-09-19
buffer: fix writes into out-of-memory buffers

pks-t:
This fixes two issues with `git_buf`:

1. An infinite loop when allocating a huge amount of bytes close to `SIZE_MAX`. While obviously a bogus scenario, I still think the fix is worthwhile.
2. We failed to correctly ensure sizes of OOM'd buffers, which resulted in us writing into the `git_buf__oom` buffer.

Found while investigating #5199 once again. This is probably not the cause for the observed crashes, though, as the stack traces point somewhere else.

### DIFF
--- a/tests/core/buffer.c
+++ b/tests/core/buffer.c
@@ -231,9 +231,9 @@ check_buf_append(
 	git_buf_puts(&tgt, data_b);
 	cl_assert(git_buf_oom(&tgt) == 0);
 	cl_assert_equal_s(expected_data, git_buf_cstr(&tgt));
-	cl_assert(tgt.size == expected_size);
+	cl_assert_equal_i(tgt.size, expected_size);
 	if (expected_asize > 0)
-		cl_assert(tgt.asize == expected_asize);
+		cl_assert_equal_i(tgt.asize, expected_asize);
 
 	git_buf_dispose(&tgt);
 }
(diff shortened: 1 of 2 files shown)


ethomson:
Apparently you can't rebuild until the build finishes completely.  I cancelled & requeued manually.

pks-t:
Thanks! I tried to read the macOS log that was failing, but it really is completely unreadable due to spam caused by leak checks. :/

pks-t:
Oops, there was a memory leak in core::buffer::avoid_printing_into_oom_buffer. `git_buf_dispose` isn't going to do anything sensible if the buffer's pointer was swapped out for `git_buf__oom`.

tests/core/buffer.c
```diff
@@ -1204,3 +1204,37 @@ void test_core_buffer__dont_grow_borrowed(void)
 
 	cl_git_fail_with(GIT_EINVALID, git_buf_grow(&buf, 1024));
 }
+
+void test_core_buffer__dont_hit_infinite_loop_when_resizing(void)
+{
+	git_buf buf = GIT_BUF_INIT;
+
+	cl_git_pass(git_buf_puts(&buf, "foobar"));
+	cl_git_fail(git_buf_try_grow(&buf, SIZE_MAX, true));
```
ethomson: It looks like macOS isn't behaving the way you want.  I think that you're relying on allocation semantics here and that you may need to substitute an intentionally failing allocator here that asserts that it ends up with `SIZE_MAX` and fails appropriately.

pks-t: Yeah, I expected this to happen but hoped it wouldn't, so I just gave it a try. I've fixed the tests to not rely on allocation semantics

pks-t:
Finally green with another memory leak fix in the OOM case.

PR #5234, opened 2019-09-19
tests: online variants

pks-t:
My initial goal has been to clean up our online tests a bit, because I had quite some problems to grasp everything that's going on with environment variables etc. While splitting up the tests I realized that it's quite easy to use our new clar variants for online::clone::http, so I've implemented them.

Due to this we now have additional tests with GitLab (I manage the libgit2 group in GitLab and thus added the required repos). I'd love to test against Azure more thoroughly as well, but I'll have to defer to @ethomson for this. Could you maybe push our libgit2/testgitrepository and libgit2/testemptyproject repos to Azure so that we can add these as variants?


ethomson:
Yes!  They're already there, in fact:

https://dev.azure.com/libgit2/test/_git/empty and
https://dev.azure.com/libgit2/test/_git/test

pks-t:
Oh, perfect, thanks for letting me know, in that case I'll make use of them. Will have to find out why there is a new memory leak in libssh2 now that even exists if the SSH tests aren't executed at all. Will do after I've polished up the other PRs.

pks-t:
There was one real memory leak on macOS which I've fixed. The other leaks already exist in master, no idea why I uncovered them just now :/

pks-t:
Well, they do exist on master now, too. Meh. Will fix...

https://dev.azure.com/libgit2/libgit2/_build/results?buildId=2664

pks-t:
Rebased on top of #5240. Added "test" and "empty" repos to the Azure variant

PR #5237, opened 2019-09-21
cred: add missing private header in GSSAPI block

tiennou:
Should have been part of 8bf0f7eb26c65b2b937b1f40a384b9b269b0b76d, which I missed by having relied too much on Xcode's refactorer (#blameitontheIDE).


pks-t:
Thanks a lot for the fix, @tiennou!

PR #5238, opened 2019-09-21
macOS GSS Support

### LINKED ISSUE #5230: Can't compile with GSSAPI enabled on macOS
I am trying to compile libgit2 with GSSAPI enabled for macOS, but I can't seem to get it working. It looks like macOS provides a kerberos framework that was [first available in macOS 10.0](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/OSX_Technology_Overview/SystemFrameworks/SystemFrameworks.html), which is what I'd expect libgit2 to use when trying to find GSSAPI on mac. However, FindGSSAPI.cmake doesn't appear to be looking for that framework, and the default UNIX case doesn't work on my mac. 

Is there a way to get this working using the existing cmake build that I'm not aware of?

## Reproduction steps
cmake -DUSE_GSSAPI=ON ..

### Expected behavior
To find GSSAPI and enable SPNEGO authentication support.

### Actual behavior
Doesn't find GSSAPI.

0.28.2

macOS 10.14.6

tiennou:
This is the needed changes to get the Apple GSS.framework used as a GSS provider on macOS (and only it, actually). This is completely untested, but as I managed to get deprecation warnings on #5237 it went downhill quickly.

Should fix #5230.


src/CMakeLists.txt
```diff
@@ -236,12 +236,20 @@ ADD_FEATURE_INFO(ntlmclient GIT_NTLM "NTLM authentication support for Unix")
 # Optional external dependency: libgssapi
 IF (USE_GSSAPI)
 	FIND_PACKAGE(GSSAPI)
+	IF (APPLE)
+		INCLUDE(FindGSSFramework)
+	ENDIF()
```
pks-t: I think that if there's multiple possible backends, we should expose this as another CMake option so that one may explicitly choose the backend.

src/transports/auth_negotiate.c
```diff
@@ -14,8 +14,12 @@
 #include "auth.h"
 #include "git2/sys/cred.h"
 
+#ifdef __APPLE__
+#import <GSS/GSS.h>
```
pks-t: Is the API really completely the same? That's pleasant

src/CMakeLists.txt
```diff
@@ -236,12 +236,20 @@ ADD_FEATURE_INFO(ntlmclient GIT_NTLM "NTLM authentication support for Unix")
...
 ENDIF()
 IF (GSSAPI_FOUND)
 	SET(GIT_GSSAPI 1)
 	LIST(APPEND LIBGIT2_LIBS ${GSSAPI_LIBRARIES})
+	ADD_FEATURE_INFO(SPNEGO GIT_GSSAPI "SPNEGO authentication support")
+ENDIF()
+IF (GSSFRAMEWORK_FOUND)
+	SET(GIT_GSSAPI 1)
+	LIST(APPEND LIBGIT2_LIBS ${GSSFRAMEWORK_LIBRARIES})
+	ADD_FEATURE_INFO(SPNEGO GIT_GSSAPI "SPNEGO authentication support (fwk)")
```
ethomson: What is fwk?  Framework?  I think more verbosity here is better, eg `SPNEGO authentication support (macOS GSS.Framework)`

src/CMakeLists.txt
ethomson: That might make sense.  I can imagine gssapi getting installed via homebrew and you want to select one or the other.

src/transports/auth_negotiate.c
tiennou: It's… strange. The `/usr` headers have their symbols marked as deprecated in favor of GSS, so this makes sure we don't select the `/usr` ones. And there's a more CoreFoundation-style API available as well, which is different. But using the old symbols while linking to GSS is fine 🤷‍♂.

src/CMakeLists.txt
tiennou: True, I wrote that kinda not expecting such quick success. I'll reword.

src/CMakeLists.txt
tiennou: Isn't it always about backends these days ? 😆

I'll add the selection code, but I mostly filed that so that it could be "runtime checked" before going in, since there's interest. Not just yet though 😉.

pks-t: Yeah, it is. I wonder whether we could write CMake modules to handle these with less boilerplate and more consistently. Definitely not required to be part of this PR, though. Maybe I'll some day find some time to tackle this.

ethomson:
Trying this, I get:

```
-- Could NOT find GSSAPI (missing: GSSAPI_LIBRARIES)
CMake Error at src/CMakeLists.txt:240 (INCLUDE):
  INCLUDE could not find load file:

    FindGSSFramework
```

tiennou:
Rebased, I've added the missing files, copyedited the backend selection, tested both methods locally, and added GSS to the macOS bots.

pks-t:
CI still complains about not being able to find the SelectGSSAPI module :/

src/CMakeLists.txt
```diff
@@ -233,15 +233,9 @@ IF (USE_NTLMCLIENT)
...
-	FIND_PACKAGE(GSSAPI)
-ENDIF()
-IF (GSSAPI_FOUND)
-	SET(GIT_GSSAPI 1)
-	LIST(APPEND LIBGIT2_LIBS ${GSSAPI_LIBRARIES})
-ENDIF()
-ADD_FEATURE_INFO(SPNEGO GIT_GSSAPI "SPNEGO authentication support")
+# Optional external dependency: GSSAPI
+
+INCLUDE(SelectGSSApi)
```
pks-t: Ah, that's why. You've named the file SelectGSSAPI, not SelectGSSApi. That bites us on case-sensitive filesystems.

ethomson: Let's stick with GSSAPI (all caps) which is the way it refers to itself and is documented.

tiennou: Cool, it's the casing I chose ^^.

src/transports/auth_negotiate.c
pks-t: So if the user explicitly chose "-DUSE_GSSAPI=gssapi", then we'd still choose the Apple one. I think we need to put backend information into "features.h.in" and use that instead of `__APPLE__`.

tiennou:
Rebased. Note that it seems Xcode 11 dislikes something in how we setup on macOS — I can't clean-build master without a small fix since I upgraded, because I'm seeing this error

```
: No such file or directory: '/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/System/Library/Frameworks/CoreFoundation.framework'
```

even on `master`. I think the previous `ld` might have been more lenient by accepting an SDK path like this instead of the "standard" `-framework CoreFoundation`. I'll investigate and file a separate PR for that though.

pks-t:
Thanks a lot, @tiennou!

PR #5240, opened 2019-09-21
valgrind: suppress memory leaks in libssh2_session_handshake

pks-t:
On Ubuntu, the combination of libgcrypt and libssh2 is quite old and
known to contain memory leaks. We thus have several functions listed in
our suppressions file that are known to leak. Due to a recent update of
libssh2 or libgcrypt, there now is a new memory leak caused by
libssh2_session_handshake that causes the CI to fail.

Add a new suppression to fix the issue.

### DIFF
--- a/script/valgrind.supp
+++ b/script/valgrind.supp
@@ -56,6 +56,15 @@
 }
 
 {
+	ignore-libssh2-init
+	Memcheck:Leak
+	...
+	fun:gcry_control
+	fun:libssh2_init
+	...
+}
+
+{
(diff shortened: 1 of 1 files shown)


pks-t:
Another point in case to provide a libssh backend :roll_eyes:

pks-t:
I'm fast forwarding this to unbreak CI

ethomson:
> I'm fast forwarding this to unbreak CI

Thanks!  👍

PR #5242, opened 2019-09-25
Don't use enum for flags

csware:
Using an `enum` causes trouble when used with C++ as bitwise operations are not possible w/o casting (e.g., `opts.flags &= ~GIT_BLOB_FILTER_CHECK_FOR_BINARY;` is invalid as there is no `&=` operator for `enum`).

I'm unsure whether `uint32_t` is a good way, because then I get "signed/unsigned" warnings in the case above - no warnigns would be generated with a plain `int` (cf. the following is needed with `uint32_t`: `opts.flags &= ~static_cast<uint32_t>(GIT_BLOB_FILTER_CHECK_FOR_BINARY);`).

### DIFF
--- a/include/git2/blob.h
+++ b/include/git2/blob.h
@@ -122,8 +122,8 @@ typedef enum {
 typedef struct {
 	int version;
 
-	/** Flags to control the filtering process */
-	git_blob_filter_flag_t flags;
+	/** Flags to control the filtering process, see `git_blob_filter_flag_t` above */
+	uint32_t flags;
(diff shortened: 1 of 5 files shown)


pks-t:
Fully agreed. I already highlighted such issues in other PRs in the past, using enums for bitfields is the wrong thing to do.
We already use unsigned types to represent such bitfields, and I think we should just continue doing so. So `uint32_t` looks perfectly fine to me.

PR #5243, opened 2019-09-26
Memory optimizations for config entries

pks-t:
This provides two optimizations for our config entries code.

The first optimization is rather straight forward and an obvious win, most importantly because it decreases code complexity by not having to track two separate lists of config entries anymore.

The second commit is a micro-optimization for how we store multivar entries. As all entry names are the same for a multivar, we can just deduplicate these and thus greatly decrease memory consumption (e.g. the attached test case goes down from 10GB RAM to a few hundred kilobytes, only). In fact, we have a deeper problem here because it is actually trivial for a comparatively tiny configuration file to blow up in size during parsing, and we cannot fix that without introducing a major interface break. So I'm split whether that optimization is worth it or not, but at least the code complexity doesn't really increase by a lot.

More details are in the commit messages.


src/config_entries.c
```diff
@@ -11,6 +11,7 @@ typedef struct config_entry_list {
 	struct config_entry_list *next;
 	struct config_entry_list *last;
 	git_config_entry *entry;
+	bool first;
```
tiennou: Sorry, I'm having a hard time understanding your good-looking commit message, so I'll try to rephrase it :

```
Whenever adding a configuration entry to the config entries structure,
we allocate two list heads.

- The first list head is added to the global list of config entries, to be
able to iterate over configuration entries while preserving their order.

- The second list head is added to the map of entries. When iterating,
an already existing entry would be added to this head, while a new
one would be directly added to the map.
```

I'm confused by "its" in your version, so I might be misreading, and the difference is meaningful ^^.

```
We can completely get rid of this secondary list by just adding a
`first` field to the first configuration entry list.
```

"`first` field to the list structure itself." I think reads clearer, "first configuration entry list" sounds like "first list head" above.

I'd split the "solution" and "optimization" ¶'s as well. Thanks for the read 😉.

src/config_entries.c
```diff
@@ -148,71 +126,42 @@ void git_config_entries_free(git_config_entries *entries)
...
-
-	if ((existing = git_strmap_get(entries->map, entry->name)) == NULL) {
-		/*
-		 * We only ever inspect `last` from the first config
-		 * entry in a multivar. In case where this new entry is
-		 * the first one in the entry map, it will also be the
-		 * last one at the time of adding it, which is
-		 * why we set `last` here to itself. Otherwise we
-		 * do not have to set `last` and leave it set to
-		 * `NULL`.
```
tiennou: "Thank you for the read" wasn't enough 🤣.

src/config_entries.c
```diff
@@ -108,7 +108,8 @@ static void config_entries_free(git_config_entries *entries)
 	list = entries->list;
 	while (list != NULL) {
 		next = list->next;
-		git__free((char *) list->entry->name);
```
tiennou: IMHO the usage of "key" in the commitmsg makes it less clear that name and key are the same thing.

```
Multivars are configuration entries that have many values for the same name;
we can thus micro-optimize this case by just retaining the name of the first
configuration entry and freeing all the others, letting them point to string of
the first entry.
```

pks-t:
Thanks @tiennou, I've amended the commit messages to make them (hopefully) easier to understand :)

PR #5244, opened 2019-09-26
Provide flags to limit config file lookup when open a repository

immortalbugs:
By supporting specifiying config file locations, we minimize filesystem
calls to check if config file exists. This optimization helps when
repositories reside on network storage.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -803,8 +803,13 @@ int git_repository_open_ext(
 	git_repository *repo = NULL;
 	git_config *config = NULL;
 
-	if (flags & GIT_REPOSITORY_OPEN_FROM_ENV)
-		return _git_repository_open_ext_from_env(repo_ptr, start_path);
+	if (flags & GIT_REPOSITORY_OPEN_FROM_ENV) {
+		error = _git_repository_open_ext_from_env(repo_ptr, start_path);
+		if (!error && repo_ptr) {
+			(*repo_ptr)->open_flags |= flags;
+		}
+		return error;
+	}
 
 	if (repo_ptr)
 		*repo_ptr = NULL;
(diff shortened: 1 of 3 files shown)


src/repository.c
```diff
@@ -1016,14 +1022,25 @@ int git_repository_config__weakptr(git_config **out, git_repository *repo)
 		git_buf programdata_buf = GIT_BUF_INIT;
 		git_config *config;
 
-		git_config_find_global(&global_buf);
-		git_config_find_xdg(&xdg_buf);
-		git_config_find_system(&system_buf);
-		git_config_find_programdata(&programdata_buf);
+		if (!(repo->open_flags & (uint32_t)GIT_REPOSITORY_OPEN_NO_GLOBAL_CONFIG)) {
```
tiennou: `OPEN_SKIP` ? I don't know how consistent we are, but positivity's always a win 😉.

hackworks: @tiennou Thank you for the detailed comments

From an API point of view and extensibility, I completely agree this PR handles a very specific requirement by increasing complexity (adding flags exposed in the API). Having specific APIs to support various approaches to construct a repository would be much better instead of overloading a single API with extra flags (leads to confusion and makes it inflexible).

It is more like having explicit constructors in C++ as against having a single constructor with a multitude of arguments supporting different requirements.

Let me take a closer look at your work in #4856 and I am more inclined to closing this PR in favor of your approach.

tiennou: Just to be clear, I don't think what I'm envisioning is going to happen anyways on such a short notice, so to speak — it's just that I added support for the `GIT_CONFIG` envvar (as specified by `git-config`). In other words, maybe it is possible to just point `GIT_CONFIG` to your file, instead of changing the C API, and the stacking behavior could be controlled. I'm not even sure it's possible, but I'd really prefer any modification to go on top of #4856, as you could alternatively make it possible to just build a `git_config *` and put that in `git_repository_open_options`. IIRC. YMMV, this is WIP for a reason 😉.

PR #5245, opened 2019-09-27
Add vcpkg installation instructions

grdowns:
libgit2 is available as a port in [vcpkg](https://github.com/Microsoft/vcpkg), a C++ library manager that simplifies installation for libgit2 and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build libgit2, ready to be included in their projects. 

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/libgit2/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Table of Contents
 * [Initialization](#initialization)
 * [Threading](#threading)
 * [Conventions](#conventions)
+* [Building libgit2 - Using vcpkg](#building-libgit2---using-vcpkg)
 * [Building libgit2 - Using CMake](#building-libgit2---using-cmake)
     * [Building](#building)
     * [Installation](#installation)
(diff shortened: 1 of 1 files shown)


ethomson:
Thanks @grdowns!  I must admit that I'm not very familiar with vcpkg, can you help me understand a bit better?

Is vcpkg expected to be used by people who want to build it as a dependency or to work on libgit2 itself?  I feel like it's probably the former, and I wonder if we should disambiguate this in the instructions?

I wonder if we should break this out into a "use libgit2" section that includes instructions for installing with vcpkg, homebrew, package managers, etc?  I hesitate a bit here since if you were running an LTS version of an OS their package may be very not-up-to-date and we don't provide our own (eg) Ubuntu repositories with the latest.

grdowns:
Thanks for the feedback, @ethomson! I'd be happy to fill you in on vcpkg wrt this PR.

You're correct that it's the former. As per the portfile linked above, we run the existing CMake build integration and place the resulting binaries and pkgconfig files in a standardized directory layout. Vcpkg ports (libraries *ported* into vcpkg) automatically register with MSBuild, can be included via a CMake toolchain file, or exported as standalone libraries. This makes library consumption pretty easy as a downstream for use in their own projects.

As for disambiguating whether this is for downstream consumers or libgit2 contributors, I think the fact that it's a question in the first place warrants some slight change in wording to save others some time and keep the barrier to entry low.

As for breaking vcpkg into a "use libgit2" or a "package manager" section, that's completely reasonable! Many repos take this same approach, as it makes it clear that the packages aren't necessarily maintained by the code authors. It can help to direct some of the "why is package X not working on platform Y?" questions away from the repo maintainers and it keeps the README's a bit more hygienic.

Lastly I think it's worth noting that vcpkg has dependencies on git, curl, unzip, tar, and g++ >= 6. Cmake and Ninja are downloaded during vcpkg's bootstrap phase. Some LTS OS's won't have access to g++ >= 6 in their standard repositories. If you think this is worth noting, we can include that information in this PR.

I'll make an update to this PR to address these points. Thanks again!

ethomson:
This was (unintentionally) closed as part of renaming the `master` branch to `main`.

I tried to simplify things a bit in #5772

Issue #5225, opened 2019-09-11
libssh2 deprecated in Redhat 7 removed in 8.


toddr:
I just thought I'd give you a heads up. libssh2-devel is no longer available in Redhat 8. Per Redhat: 

https://bugzilla.redhat.com/show_bug.cgi?id=1660344

> Long story short, libssh2 has been removed from RHEL-8 intentionally.
> 
> libssh is now the supported SSH library in RHEL-8.  It has incompatible ABI/API with libssh2, so libssh2-based application needs to be migrated to libssh unless they support both the libraries (as, for example upstream curl does).

I don't know what other distros are doing but you might consider if switching to libssh might be desirable in the near term.

ethomson:
Thanks for the details.  I'd like to think that we have abstracted the transport layer sufficiently that we could support libssh2 or libssh (I would hate to pull the rug out from under people who use and even package libssh2 themselves).

In any case, I think that this makes the libssh support more urgent.

cuviper:
I intend to work on `libssh` support here, and I'd be happy to collaborate. My own interest comes indirectly from Rust/Cargo's use of `libgit2`, and how we're building that toolchain in RHEL 8.

cuviper:
(I don't mind working independently either, just don't want to duplicate effort.)

pks-t:
@tiennou: well, if you've started working on it already then I'm happy to defer this to you. I guess collaborating is rather hard without having an immediate communications channel, and I'd probably have to start from scratch first to get the whole picture (that's not because of how you have started the work, but simply to help myself understand everything)

tiennou:
@cuviper Sorry about that, I forgot to `git add` it… I've rebased on top of #5223, and there's only one API-point left to convert — I'm [here](https://github.com/libgit2/libgit2/compare/master...tiennou:feature/libssh-backend#diff-6d6978d8ac6c25b7de74fce87bea7d18R407), and I'm looking at the mismatch with the `LIBSSH2` typedefs, actually. And then test, test, test 😉. The main "logic" to the port hides the differences inside `ssh_backend.h`, using a mess of `#define`s, to keep the code in `ssh.c` agnostic and clear.

@pks-t I have hem, stuff, scheduled for the next couple of weeks. I just bit the bullet about writing the backend this week, out of frustration, so it's kinda strange to have it accelerate like this… In short, I'm kinda deferring to @cuviper (thanks!) for a few weeks.

cuviper:
[...]

Issue #5219, opened 2019-09-02
ssh: check if private key is encrypted with passphrase


eaigner:
I'm crossfiling this [libssh2 issue](https://github.com/libssh2/libssh2/issues/403) since I don't know in what repo the issue belongs.

Basically when trying to connect with an encrypted private key, the request just times out with a generic error instead of failing properly. I don't know if it is possible to check for that with the currently surfaced api of libssh2, but it seems not.

The problem is in most cases the passphrase is not provided up-front like the API requires it now. I think the better flow (if possible) would be:

- Try to auth with given private key
- If it fails, invoke other callback to request the passphrase

pks-t:
I could imagine just introducing a new cred type, e.g. "GIT_CREDTYPE_SSH_KEY_PASSPHRASE". If we later determine that the passed key is in fact encrypted, then we'd execute the credentials callback with `allowed_types` set to the new cred type. I guess that should trivially solve your usecase and also make use of the existing credentials infrastructure. What's needed though is any way to actually notice that the key is encrypted, and that'd need to be supported by libssh2. Judging from your issue there, it doesn't seem like this support exists yet, right?

eaigner:
I've already implemented a check for encrypted keys on my part in Swift, more details on my question here https://stackoverflow.com/questions/57655324/check-if-ssh-private-key-is-encrypted

The other check for linux style keys (PEM header) is simply to check for the `DEK-Info` header

```
import Foundation

private let opensshMagic = "openssh-key-v1"

public class SSHPrivateKey {

    public struct OpenSSHKey {
        let cipherName: String
        let kdfName: String
        let kdfOptions: Data
        let numKeys: Int

        var isEncrypted: Bool {
            return cipherName != "none"
        }
    }

    public let data: Data

    public init(data: Data) {
        self.data = data
    }

    public func openSSHKey() -> OpenSSHKey? {
        // #define AUTH_MAGIC      "openssh-key-v1"
        //
        // byte[]  AUTH_MAGIC
        // string  ciphername
        // string  kdfname
        // string  kdfoptions
        // int     number of keys N
        // string  publickey1
        // string  publickey2
        // ...
        // string  publickeyN
        // string  encrypted, padded list of private keys

        guard let magic = opensshMagic.data(using: .utf8) else {
            return nil
        }

        if data.prefix(magic.count) != magic {
            return nil
        }

        var offset = magic.count + 1

        guard let cipherName = data.consumeString(offset: &offset),
[...]

Issue #4452, opened 2017-12-26
SSH: Pass hostkey to certificate check callback


aw32:
Feature request:
There is the certificate check callback for remote connections.
For the SSH certificate check `git_cert_hostkey` can only contain the MD5 and/or SHA1 hash of the hostkey as passed from libssh2. Since MD5 and SHA1 are not considered trustable anymore, it would be nice to receive the actual hostkey as returned by libssh2.
Would you consider this reasonable enough to accept a pull request?

ethomson:
I'm not very familiar with libssh2, but if it's still another format that libssh2 uses, then this seems reasonable.  @carlosmn?

carlosmn:
We provide these checksums as it's what libssh2 lets us have. It looks like they also provide the raw hostkey, but we do not currently have a mechanism for providing multiple forms of certificate.

carlosmn:
It looks like this was added in #5307 and #5704

Issue #4289, opened 2017-06-30
You shouldn't need a public key to ssh auth


MrStonedOne:
### Reproduction steps
call git_cred_ssh_key_new() with just a private key
### Expected behavior
it works, the private key already has the public key encoded in it.
### Actual behavior
it fails
### Version of libgit2 (release number or SHA1)
c26ce78 (n/a, head given)
### Operating system(s) tested
windows 7

downstream bug report: tgstation/tgstation-server#96

pks-t:
To authenticate a session with an SSH key, we execute `libssh2_userauth_publickey_fromfile` with the username, public key, private key and key passphrase provided by the credentials callback. This function in turn passes these values on to `libssh2_userauth_publickey_fromfile_ex`. As you can see in its documentation [1], it is valid to pass in a publickey of `NULL`. But unfortunately this is only supported when libssh2 itself is linked against OpenSSL, as OpenSSL is required to do the factorization to compute the public key.

Seeing that libssh2 does not provide a method to derive the public key from the private key, I would refrain from implementing this ourselves. To properly do so, we should probably link against OpenSSL, and in that case one could argue that you should build libssh2 against OpenSSL, as well.

So the proper solution would be to try and link libssh2 against OpenSSL. Seeing you are on Windows, you're probably using the bundled version of libssh2, which we currently include into our own build instructions without the support for OpenSSL at all. Incidentally, I've created #4303, which instead builds libssh2 as a subdirectory, making use of its own CMake build instructions. As soon as it is merged, it might then be possible to link the embedded libssh2 against OpenSSL when it is configured correctly.

[1]: https://www.libssh2.org/libssh2_userauth_publickey_fromfile_ex.html

PR #5253, opened 2019-10-02
[WIP] libssh backend

tiennou:
This adds support for libssh as a possible backend for the SSH transport. The code is untested — the test suite runs fine locally, but I want to see how CI reacts to it as I don't expect everything SSH to be covered by it (`keyboard-interactive` 🙄).


src/transports/ssh_backend.h
```diff
@@ -0,0 +1,477 @@
...
+					session, c->username, c->prompt_callback);
+#elif defined(GIT_LIBSSH)
+
+				rc = ssh_userauth_kbdint(session, c->username, NULL);
+				if (rc == SSH_AUTH_INFO) {
+					const char *name = ssh_userauth_kbdint_getname(session);
+					const char *instr = ssh_userauth_kbdint_getinstruction(session);
+					int num_prompts = ssh_userauth_kbdint_getnprompts(session);
+					size_t i;
+					ssh_userauth_kbdint_getprompt(session, i, NULL);
```
tiennou: This is the API endpoint I'm having issues with — the public callback with the libssh2 types — with the libssh API.

 I'm not sure how to best proceed forward from here :
- break the API and make a more generic callback, with no exposed types from the backends (the new `git_ssh_` types).
- provide a libssh-specific version of it, passing only the session. This is contrary to the guideline that the API should not depend on preprocessing, and would cause Docurium to miss it (as an example).
- make the SSH session accessible, so we don't have to wrap so closely libssh2, and deprecate anything that's not really needed.

Any suggestions ?

src/transports/ssh_backend.h
```diff
@@ -0,0 +1,477 @@
...
+				break;
+			}
+#ifdef GIT_SSH_MEMORY_CREDENTIALS
+			case GIT_CREDTYPE_SSH_MEMORY: {
+				git_cred_ssh_key *c = (git_cred_ssh_key *)cred;
+
+				assert(c->username);
+				assert(c->privatekey);
+
+				rc = libssh2_userauth_publickey_frommemory(
```
tiennou: This is TBD.

tiennou:
I forgot to tag the crux of the issue in the PR's cover, but please see #5225 for context.

To clarify my perspective on #5307, ideally `git_cert` could have a `git_cert_ssh_cred` opaque object, or something like it, which would allow the user's code to just use the capabilities of their chosen backend, to access key material, auth-methods, or SSH channels directly.

This is not currently handled here, because I hit one case where the interface we already provide cannot be ported to libssh (the custom signing mechanism), so here's a summary of "currently known issues caused by libssh2": apparently it's gone from Debian (#5225), some requested features are missing (#5258, #5219, #4452), or depend on the libssh2's backend and key types supported (#4289, OpenSSL/libgcrypt only), and IIRC there's another issue…

CMakeLists.txt
```diff
@@ -52,7 +52,7 @@ OPTION(TAGS				"Generate tags"						OFF)
 OPTION(PROFILE				"Generate profiling information"			OFF)
 OPTION(ENABLE_TRACE			"Enables tracing support"				OFF)
 OPTION(LIBGIT2_FILENAME			"Name of the produced binary"				OFF)
-OPTION(USE_SSH				"Link with libssh2 to enable SSH support"		 ON)
+OPTION(USE_SSH				"Enable SSH support. Can be set to a specific backend (libssh2)" ON)
```
cuviper: I think this needs to be `SET` to properly cache the backend value, like: 

```suggestion
   SET(USE_SSH				"ON" CACHE STRING "Enable SSH support. Can be set to a specific backend (libssh2)")
```

src/transports/ssh.h
```diff
@@ -8,7 +8,73 @@
...
+
+#define git_ssh_session_new() libssh2_session_init()
+#define git_ssh_session_disconnect(s, msg) \
+	libssh2_session_disconnect(s->session, msg);
+
+struct git_ssh_channel {
+	LIBSSH2_CHANNEL *channel;
+};
+
+#define git_ssh_channel_close(c) libssh2_channel_close(c->channel)
```
cuviper:
```suggestion
#define git_ssh_channel_close(c) libssh2_channel_close(c)
```

src/transports/ssh.h
```diff
@@ -8,7 +8,73 @@
...
+#define git_ssh_channel_free(c) libssh2_channel_free(c->channel)
```
cuviper:
```suggestion
#define git_ssh_channel_free(c) libssh2_channel_free(c)
```

src/transports/ssh_libssh2.c
```diff
@@ -0,0 +1,376 @@
...
+				 * libssh2_session_abstract(), which will replace the existing
+				 * pointer as is done below. This is safe for now (at time of writing),
+				 * but may not be valid in future.
+				 */
+				*abstract = &payload;
+
+				rc = libssh2_userauth_keyboard_interactive(
+					s->session, c->username, git_ssh__kbdint_cb);
+
+				*abstract = old_abstract;
```
cuviper: I see you added `old_abstract` in commit 98c9f6a6, but its declaration got lost in the following commit 749f3440.

extrawurst:
@tiennou is this still worked on?

tiennou:
@extrawurst I've rebased it, updated CI, and the only showstopper now is that the `libssh` backend seem to require some system-level configuration files to exist.

@cuviper Thanks for the review, much appreciated ❤️. Issues (apart from the runtime config ones, and unforeseen bugs 😉) should be fixed.

carlosmn:
> This is the API endpoint I'm having issues with — the public callback with the libssh2 types — with the libssh API.

There's a fundamental conflict between the goals of the library and the realities of how different all these libraries and network protocols are as we both want to

1. have the same interface for performing Git operations over an arbitrary network protocol; and
2. allow the caller to react in whichever is the best way in their environment (language, user interaction)

The former goal means things like auth can get awkward as most HTTP auth doesn't look like SSH auth (and some HTTP auth doesn't look like a lot of other HTTP auth) but the latter means that we want to avoid processing the data too much when the end-user application (or whatever ends up calling the library) needs to interact with the library implementing the network protocol. This is made even more important by the fact that we want to allow callers of the library to plug in their own network layer and then they can devise an arbitrarily complex authentication method as long as both agree.

The wish to have the caller in control but also make it easy to perform the most common operations leads to weird things in the API like libssh2- and curl-specific which try to give the caller all the info they might need short of handing over a pointer to whatever context/handle each library has. At the same time it's rather disappointing how the information is presented by some libraries (it seems libcurl does the data processing we're trying not to do here), but that's a different discussion.

As far as your question about the callback types having specific libssh2 structures, I remember doing this due to the methods needed for the more complex interactions like keyboard-interactive, seem to be tied enough to libssh2 that did not seem realistic that we could provide some generic method for this. And from what I can tell, libssh has a rather different mechanism for doing this. Maybe it's not very sensible for this library to support this.

The same goes for certs where we only got two hashes from libssh2. It looks like the issue described in #5258 and #4452  has been fixed now, though it's not super helpful now that libssh2 isn't in some distros, I suppose. But we do have a place to put it so the caller can decide for themselves how they want to hash the cert if they need to. libssh expects you to ask for which hash you're interested in, which makes sense for it, but I don't know how much it makes for us to expose that detail.

For what it's worth, I have come to think that the design with the callbacks for host and user authentication were a mistake, but changing it to work with returning control to the user by returning is not a small task.

Ultimately, some of this probably isn't solvable as part of this library. We can try to keep it working with the features we have (and maybe not bother with keyboard-interactive for libssh?) and with libssh we can even fill in the info as to whether "we" think the ssh server is something you should trust. But we're still just not going to provide functionality for managing your `known_hosts` file. If an app does want to provide that, they're going to have to get dirty with libssh themselves or perhaps provide their own transport that talks to their app via some other means.

All this to say that attaching auth logic into libgit2 is always going to leave us with something somewhat annoying and the real solution is probably to provide some simple options in the library and leave an escape hatch for something like ssh where the amount of complexity is high and also all of this should be very much outside of the things we should have to care about.

Maybe at a later date we can have even built-in transports look like things you plug in so you can keep your own pointer to something if you need more than bog standard behaviour. But that's for when many other pressing problems have been solved.

src/transports/ssh_libssh.c
```diff
@@ -0,0 +1,321 @@
...
+				break;
+			}
+#ifdef GIT_SSH_MEMORY_CREDENTIALS
+			case GIT_CREDTYPE_SSH_MEMORY: {
+				git_cred_ssh_key *c = (git_cred_ssh_key *)cred;
+
+				assert(c->username);
+				assert(c->privatekey);
+
+				rc = libssh2_userauth_publickey_frommemory(
```
carlosmn, line 291: There are a series of `ssh_pki_import_privkey_` functions provided by libssh which let you specify an in-memory key in different formats. The one that would let us keep the same for both looks like it would be `ssh_pki_import_privkey_base64` as it oddly doesn't seem to allow you just pass the raw data, so we'd have to first convert it into base64.

carlosmn: Or it looks like they might mean base64 as in the format you would find on-disk

tiennou (commented):
A few notes, sadly it's been a while so the context-switch is kinda cold on my side but that's what you get when you drop off the face of the earth.

Just to clarify, the libssh backend as it is should now handle "our" `keyboard-interactive` interface via a trampoline that just collects all challenges and converts them. It's just that actually setting up a test for that requires a harness/configuration which I haven't really looked at, so it's untested.

The thing that's (IIRC) completely missing from libssh is a way to support our custom hashing. Conversely, libssh seems to refer to GSSAPI mic authentication, which is absent from libssh2 (and gave me a good chuckle as a long-time Uplink player — I wasn't aware that voice authenticated SSH was a thing 😅).

I do agree that, if I was designing this from a blank slate, I'd go for a much more simple "try once with what the user gave us (pass, file/memory pkey, whatever, even a blanket this-asked-for-auth `void *` handle), or even and fail immediately with no retries", but API-wise that's not possible.

Anyway, I'll try to revisit that soonish, thanks for the review @carlosmn !

src/transports/ssh_libssh.c
```diff
@@ -0,0 +1,321 @@
...
+}
+
+int git_ssh_session_server_hostkey(git_ssh_session *s, git_cert_hostkey *cert)
+{
+	ssh_key pkey;
+	unsigned char hash[32];
+	unsigned char *hash_ptr = (unsigned char *)&hash;
+	size_t hash_len;
+
+	if (ssh_get_server_publickey(s->session, &pkey) == SSH_OK) {
```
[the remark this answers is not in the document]
tiennou: Fixed. Note that I'm putting the base64-encoded raw data, because that's what libssh provides.

src/transports/ssh_libssh2.c
```diff
@@ -0,0 +1,437 @@
...
+		return -1;
+	}
+
+	return 0;
+}
+
+int git_ssh_session_server_is_known(git_ssh_session *s, int *valid)
+{
+	GIT_UNUSED(s);
+	*valid = 0;
```
tiennou, line 190: This ensures that libssh2's behavior doesn't change (default to "can't say it's valid", a.k.a `0`), but now libssh checks for known_hosts validity, while libssh2 doesn't (and it looks messy to handle appropriately). I'm not even sure if it makes sense to update the known_host file afterwards…

jkaiser-ioki:
Just to add my 2 cents. GitKraken (https://www.gitkraken.com/) is using libgit2, but due to the underlying usage of libssh2, it is currently not possible to use the client when talking to git remotes on ssh-hardened servers (e.g. only ed25519 host keys, or restricted allowed-cipher-sets, ...). Seeing this go through would probably improve things significantly on that front :-)

ethomson:
We have openssh support now — I really like libssh (and not so much libssh2) but it feels like openssh is the thing to use for "better openssh compatibility"

Issue #5256, opened 2019-10-07
File locking broken on POSIX operating systems


henkesn:
### Reproduction steps
1. Create a transaction and lock a ref
2. Create another transaction and lock the same ref

### Expected behavior
=> git_transaction_lock_ref should return an error code != 0

### Actual behavior
- On windows an error is returned
- on Linux (and Windows with WSL) 0  is returned
- the <ref>.lock file is still created

### Version of libgit2 (release number or SHA1)
f04a58b

### Operating system(s) tested
Windows (bare and WSL), Linux

Perhaps find_locked could prevent such cases.

henkesn:
Today, I investigated a litte further and debugged into this issue comparing Windows and Linux. I found out that the Windows file locking behavior keeps the lock file locked (accidentally).

The problem is IMHO that `refdb_fs_backend__lock` (which is the implementation of the `refdb->lock()`) uses [`loose_lock`](https://github.com/libgit2/libgit2/blob/master/src/refdb_fs.c#L847), which itself sets [`GIT_FILEBUF_FORCE`](https://github.com/libgit2/libgit2/blob/master/src/refdb_fs.c#L807) as file flag and leads to the lock file to be [unlinked](https://github.com/libgit2/libgit2/blob/master/src/filebuf.c#L48).

Actually, I don't really get the meaning of `loose_lock` and there is nothing like a non-enforcing `lock`.
I tried to remove the `GIT_FILEBUF_FORCE` flag, but then several other tests fail.
Having this flag for all locks, the transactional refsdb looks broken to me.

henkesn:
Ok, I think I got it. With 817c28201e2a95 (quite some time ago) there were made io performance improvements. The flag `GIT_FILEBUF_FORCE` was introduced. The semantic is not that an existing lock should be delete but that `lock_file` should assert that the path exists (`mkdir -p`). With this commit the `unlink` call also found its way into the code but seems just wrong. IMHO the flag name should be changed.
This looks a major issue to me, as it breaks locks for all non-directory-root .locks on Linux and under some circumstances also under Windows.

henkesn:
I removed the `unlink` call and renamed the flag, see PR #5257.

PR #5257, opened 2019-10-07
Fix file locking on POSIX OS

henkesn:
See #5256

Should fail on Linux but be ok on Windows.

### DIFF
--- a/src/filebuf.c
+++ b/src/filebuf.c
@@ -44,18 +44,14 @@ static int verify_last_error(git_filebuf *file)
 static int lock_file(git_filebuf *file, int flags, mode_t mode)
 {
 	if (git_path_exists(file->path_lock) == true) {
-		if (flags & GIT_FILEBUF_FORCE)
-			p_unlink(file->path_lock);
-		else {
-			git_error_clear(); /* actual OS error code just confuses */
-			git_error_set(GIT_ERROR_OS,
-				"failed to lock file '%s' for writing", file->path_lock);
-			return GIT_ELOCKED;
-		}
+		git_error_clear(); /* actual OS error code just confuses */
+		git_error_set(GIT_ERROR_OS,
+			"failed to lock file '%s' for writing", file->path_lock);
+		return GIT_ELOCKED;
 	}
 
 	/* create path to the file buffer is required */
-	if (flags & GIT_FILEBUF_FORCE) {
+	if (flags & GIT_FILEBUF_CREATE_LEADING_DIRS) {
 		/* XXX: Should dirmode here be configurable? Or is 0777 always fine? */
 		file->fd = git_futils_creat_locked_withpath(file->path_lock, 0777, mode);
 	} else {
--- a/src/filebuf.h
+++ b/src/filebuf.h
@@ -19,7 +19,7 @@
 
 #define GIT_FILEBUF_HASH_CONTENTS		(1 << 0)
 #define GIT_FILEBUF_APPEND				(1 << 2)
-#define GIT_FILEBUF_FORCE				(1 << 3)
+#define GIT_FILEBUF_CREATE_LEADING_DIRS	(1 << 3)
 #define GIT_FILEBUF_TEMPORARY			(1 << 4)
 #define GIT_FILEBUF_DO_NOT_BUFFER		(1 << 5)
 #define GIT_FILEBUF_FSYNC				(1 << 6)
--- a/tests/refs/transactions.c
+++ b/tests/refs/transactions.c
@@ -108,3 +108,24 @@ void test_refs_transactions__unlocked_set(void)
 	cl_git_fail_with(GIT_ENOTFOUND, git_transaction_set_target(g_tx, "refs/heads/foo", &id, NULL, NULL));
 	cl_git_pass(git_transaction_commit(g_tx));
 }
+
+void test_refs_transactions__error_on_locking_locked_ref(void)
+{
+	git_oid id;
+	git_transaction *g_tx_with_lock;
+	git_repository *g_repo_with_locking_tx;
+	const char *g_repo_path = git_repository_path(g_repo);
+	
+	/* prepare a separate transaction in another instance of testrepo and lock master */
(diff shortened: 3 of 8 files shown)


henkesn:
Ready to merge.

tests/refs/transactions.c
```diff
@@ -108,3 +108,24 @@ void test_refs_transactions__unlocked_set(void)
...
 }
+
+void test_refs_transactions__error_on_locking_locked_ref(void)
+{
+	git_oid id;
+	git_transaction *g_tx_with_lock;
+	git_repository *g_repo_with_locking_tx;
+	const char *g_repo_path = git_repository_path(g_repo);
+	
+	// prepare a separate transaction in another instance of testrepo and lock master
```
pks-t: We're trying to be as C90 compatible as possible, thus you should use `/* */` style comments

tests/refs/transactions.c
```diff
@@ -108,3 +108,24 @@ void test_refs_transactions__unlocked_set(void)
...
+	git_repository_open(&g_repo_with_locking_tx, g_repo_path);
+    cl_git_pass(git_transaction_new(&g_tx_with_lock, g_repo_with_locking_tx));
```
pks-t: This line uses spaces ;)

tests/refs/transactions.c
```diff
@@ -108,3 +108,24 @@ void test_refs_transactions__unlocked_set(void)
...
+
+void test_refs_transactions__error_on_locking_locked_ref(void)
+{
+	git_oid id;
+	git_transaction *g_tx_with_lock;
+	git_repository *g_repo_with_locking_tx;
+	const char *g_repo_path = git_repository_path(g_repo);
+	
+	// prepare a separate transaction in another instance of testrepo and lock master
+	git_repository_open(&g_repo_with_locking_tx, g_repo_path);
```
pks-t: Should be wrapped in `cl_git_pass`

tests/refs/transactions.c
```diff
@@ -108,3 +108,24 @@ void test_refs_transactions__unlocked_set(void)
...
+    cl_git_pass(git_transaction_new(&g_tx_with_lock, g_repo_with_locking_tx));
+	cl_git_pass(git_transaction_lock_ref(g_tx_with_lock, "refs/heads/master"));
+
+	// lock reference for set_target
+	git_oid_fromstr(&id, "a65fedf39aefe402d3bb6e24df4d4f5fe4547750");
```
pks-t: `cl_git_pass`

src/filebuf.h
```diff
@@ -19,7 +19,7 @@
 
 #define GIT_FILEBUF_HASH_CONTENTS		(1 << 0)
 #define GIT_FILEBUF_APPEND				(1 << 2)
-#define GIT_FILEBUF_FORCE				(1 << 3)
+#define GIT_FILEBUF_CREATE_MISS_DIR		(1 << 3)
```
pks-t: This is not only a rename of the flag, but in fact you completely change semantics. Previously, this flag would've just removed pre-existing locks, while the new flag will just create leading directories. I take it that the `GIT_FILEBUF_FORCE` flag is not really needed anymore after introducing `GIT_FILEBUF_CREATE_MISS_DIR`, in which case it's fine to remove it. But when one goes back in history at a later point in time, it's really confusing when one reads about "renaming" a flag when what you're really doing is creating a new one and removing the old now-unused one in the same step.

I'd also vote for a flag name "GIT_FILEBUF_CREATE_LEADING_DIRS"

src/filebuf.c
```diff
@@ -44,18 +44,14 @@ static int verify_last_error(git_filebuf *file)
...
-		}
+		git_error_clear(); /* actual OS error code just confuses */
+		git_error_set(GIT_ERROR_OS,
+			"failed to lock file '%s' for writing", file->path_lock);
+		return GIT_ELOCKED;
 	}
 
 	/* create path to the file buffer is required */
-	if (flags & GIT_FILEBUF_FORCE) {
+	if (flags & GIT_FILEBUF_CREATE_MISS_DIR) {
```
pks-t: Yeah, makes a lot of sense to have a separate flag for this

src/filebuf.c
```diff
@@ -44,18 +44,14 @@ static int verify_last_error(git_filebuf *file)
...
-		else {
-			git_error_clear(); /* actual OS error code just confuses */
-			git_error_set(GIT_ERROR_OS,
-				"failed to lock file '%s' for writing", file->path_lock);
-			return GIT_ELOCKED;
-		}
+		git_error_clear(); /* actual OS error code just confuses */
+		git_error_set(GIT_ERROR_OS,
+			"failed to lock file '%s' for writing", file->path_lock);
+		return GIT_ELOCKED;
```
pks-t: Judging from the callsites I tend to agree that none of them really wanted to force-override pre-existing locks.

src/filebuf.h
henkesn: I agree for your flag name. But I don't really think that this is a new flag. I think the intention and usage of the flag was exactly to what your suggested name describes. The unlink was a bug due to the bad naming. Actually I don't really understand how it could get introduced, as it was even the inventor of the flag who also added the unlink. What do you propose as a resolution? I think a good explanation commit message should be ok.

henkesn:
@ethomson, that's question which also came to my mind and I tried to comment out the 
```c
if (flags & GIT_FILEBUF_CREATE_LEADING_DIRS)
```
 part. I think at least all new files in new leading directories, e. g. `reference_create` during `clone` or `fetch` need the leading directories to be created. Though it's arguable if that has to take place as part of `lock_file`, respecting that directory creation has to be handled somewhere else for non-locking stuff.

ethomson:
>  I think at least all new files in new leading directories, e. g. `reference_create` during `clone` or `fetch` need the leading directories to be created. Though it's arguable if that has to take place as part of `lock_file`, respecting that directory creation has to be handled somewhere else for non-locking stuff.

Yeah.  I agree that for this change that we should just update the semantics and revisit if we want on the recreating directories.

Thanks for catching this and creating a fix.  I'll let @pks-t take one more review.

PR #5264, opened 2019-10-11
refs: unlock unmodified refs on transaction commit

henkesn:
Refs which are locked in a transaction without an altered target,
still should to be unlocked on `git_transaction_commit`.
`git_transaction_free` also unlocks refs but the moment of calling of `git_transaction_free`
cannot be controlled in all situations.
Some binding libs call `git_transaction_free` on garbage collection or not at all if the
application exits before and don't provide public access to `git_transaction_free`.
It is better to release locks as soon as possible.

### DIFF
--- a/src/transaction.c
+++ b/src/transaction.c
@@ -334,7 +334,13 @@ int git_transaction_commit(git_transaction *tx)
 				return error;
 		}
 
-		if (node->ref_type != GIT_REFERENCE_INVALID) {
+		if (node->ref_type == GIT_REFERENCE_INVALID) {
+			/* ref was locked but not modified */
+			if ((error = git_refdb_unlock(tx->db, node->payload, false, false, NULL, NULL, NULL)) < 0) {
+				return error;
+			}
+			node->committed = true;
+		} else {
 			if ((error = update_target(tx->db, node)) < 0)
 				return error;
 		}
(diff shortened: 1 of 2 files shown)


henkesn:
Btw, would good to have this released together with #5257 because the suddenly working locks could lead to running into this issue to more frequently.

src/transaction.c
```diff
@@ -334,7 +334,11 @@ int git_transaction_commit(git_transaction *tx)
 				return error;
 		}
 
-		if (node->ref_type != GIT_REFERENCE_INVALID) {
+		if (node->ref_type == GIT_REFERENCE_INVALID) {
+			/* ref was locked but not modified */
+			git_refdb_unlock(tx->db, node->payload, false, false, NULL, NULL, NULL);
```
pks-t: I think we want to check the return code here, don't we?

henkesn: @pks-t, thank you for the review. What kind of error handling do you propose? Just returning an error immediately on failed unlock would break committing the rest of the transaction, which is imho more important to avoid. Should we collect errors and then return it/them after the transaction? What about multiple errors? I'm a bit torn.

pks-t: In fact, I'd say to just return an error immediately. In essence, it really isn't that much different to the case where the reftype is valid:

1. We call  `update_target`
2. `update_target` returns an error immediately if `git_refdb_unlock` fails
3. `git_transaction_commit` aborts if `update_target` fails

So if you ask me, we should treat errors just the same in this new code and directly return the error code.

henkesn: Alright!

pks-t: Thanks! I guess this looks good to me now, but I'll wait for others to chime in. Please feel free to ping me again if you haven't received another answer in a few days

pks-t:
Thanks again for your fix!

PR #5265, opened 2019-10-13
cmake: correct the link stanza for CoreFoundation

tiennou:
LIBRARIES is the (absolute?) path to the library.
LDFLAGS is the full linker stanza to correctly link with this lib.

By passing LIBRARIES as LIBGIT_LIBS, the linker ends up with the
absolute path for the SDK'ed version of CoreFoundation (which doesn't
exist), instead of the familiar `-framework CoreFoundation`.

I have no idea why it stopped working, but AFAIU it shouldn't have, since Xcode 10 wasn't complaining 🤷‍♂.


ethomson:
Interesting.  I'm also surprised that this changed, but the whole `-framework` flag is a little odd and unique to macOS.

But you mention an Xcode different -- I'm surprised that this started with an Xcode change and not a CMake change.  `LIBGIT2_LIBS` is just a semicolon-separated list of paths to libraries.  That list doesn't go to the linker directly, does it?  Wouldn't CMake split that apart and turn that into `-L<path_to_library>` and `-l<library_name>` flags for the linker?

Did it forget how to deal with `.framework`s?

Is the new Xcode not compatible with older CMakes?

tiennou:
`-framework` is not standard, but it works mostly the same as `-l/-L`, as far as the `-L` part is usually inside SDKs (which are *really* not standard ^^).

To be clear, I wasn't suspecting CMake at all, but it's clear that something isn't interpreting the "old way" the same now. To me, we might have been misusing CMake's autosplitter by passing it the absolute path to the framework bundle, which the old Xcode kindly fixed up. The new version doesn't do that anymore (I'd say for enforcement reasons — if I were to build for iOS, my GSS framework should not be the one from the macOS SDK).

Sadly I don't have Xcode 10 anymore, Homebrew might have updated CMake, so it's going to be pretty weird to investigate what is happening. I'll try to grab some build logs though…

ethomson:
> Sadly I don't have Xcode 10 anymore, Homebrew might have updated CMake, so it's going to be pretty weird to investigate what is happening. I'll try to grab some build logs though…

I guess it really doesn't _matter_ why, I was really just curious.  I wouldn't waste any more time on it.  😁

Issue #5262, opened 2019-10-11
Last part of stash message with a lot of \n passed to next read oid function.


lolgear:
### Reproduction steps
1. Create repository.
2. Add stashes with messages.
3. Add '\n' characters to stash messages.
4. Catch error "invalid characters in OID".

### Investigation
While you are reading message to 'first' occurrence of '\n', you will skip last part of message that can contain '\n'.
```
<valid_ref1> ... message
end_of_message
<valid_ref2>
```

or actual

```
76050078fc2923fdffb1b45e7231b4edfe4542b7 3cb51e2ace7b2aa4ebaa8985cf492117736b70f3 Author Name <author@email.opensource> 1566897223 +0300	On master: abc.
abc
3cb51e2ace7b2aa4ebaa8985cf492117736b70f3 e7454a3710b9ae4fbbdbe7fafa0699990355d88c Author Name <author@email.opensource> 1566897270 +0300	On master: abc
```

You will output "On master: abc." as a message. It is correct for sha-760.
But..
Last part of it (abc again, but with \n bracers around, \nabc\n) will exist and will be passed to next read oid function.

### Expected behavior
Skip last part of a message with a lot of '\n' and output only first part before first '\n'. (as `git stash list` do)

### Actual behavior
Do not skip last part of a message with a lot of '\n' and get invalid characters in oid error.

### Version of libgit2 (release number or SHA1)
Current master

### Operating system(s) tested
macOS 10.15

### Example
Example git repository in attachment.
[GitUpInvalidCharacters.zip](https://github.com/libgit2/libgit2/files/3716638/GitUpInvalidCharacters.zip)

ethomson:
Fixed via #5275

PR #5266, opened 2019-10-13
Reflog/Parse skip newlines at the end of last message.

lolgear:
- [x] #5262 

It is done by seek forward to a character that following after invalid character.

```c
if (get_oid_old < 0) {
  // get invalid character position
  int position = get_position();
  seek_forward(position + 1);
}
```


pks-t:
Thanks a lot for your pull request! I've taken a deeper look at the code to try to find out the root cause. The issue is that we do not sanitize entries that we're adding to the reflog ".git/logs/refs/stash". As a result, if the user specifies multi-line reflog messages, the reflog will end up with these exact newlines and thus break the format.

So two things that need to be done:

1. libgit2 needs to sanitize reflog messages in all places properly
2. We should just skip over non-parseable lines in the reflog. Git itself does so, we should do the same.

I'll open a PR in a few minutes that tackles both these issues. I hope you don't mind me effectively taking over the issue, but I think that the deeper underlying issues need to be fixed. Anyway, thanks a lot for bringing this to our attention and proposing a fix!

pks-t:
Please see #5275

ethomson:
Closed via #5275.  Thanks for reporting this issue and the collaboration to fix it.

lolgear:
Fixed in #5275

PR #5269, opened 2019-10-15
fuzzers: add a new fuzzer for patch parsing

durin42:
I was looking at this code anyway because the sr.ht people nerdsniped
me, and it gave me that "I should fuzz this" feeling. So have a fuzzer!

### DIFF
--- a/fuzzers/corpora/patch_parse/patch_fuzzer-patch.diff
+++ b/fuzzers/corpora/patch_parse/patch_fuzzer-patch.diff
@@ -1,0 +1,45 @@
+diff --git a/fuzzers/patch_fuzzer.c b/fuzzers/patch_fuzzer.c
+new file mode 100644
+index 000000000..76186b6fb
+--- /dev/null
++++ b/fuzzers/patch_fuzzer.c
+@@ -0,0 +1,39 @@
++/*
++ * libgit2 patch fuzzer target.
++ *
++ * Copyright (C) the libgit2 contributors. All rights reserved.
++ *
++ * This file is part of libgit2, distributed under the GNU GPL v2 with
++ * a Linking Exception. For full terms see the included COPYING file.
++ */
++
++#include "git2.h"
++#include "patch.h"
++#include "patch_parse.h"
++
++#define UNUSED(x) (void)(x)
++
++int LLVMFuzzerInitialize(int *argc, char ***argv)
++{
++        UNUSED(argc);
++        UNUSED(argv);
++
++        if (git_libgit2_init() < 0)
++                abort();
++
++        return 0;
++}
++
++int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
++{
++        if (size < 1) {
++                return 0;
++        }
++        git_patch* patch;
++        git_patch_options opts = {(uint32_t)data[0]};
++        int status = git_patch_from_buffer(&patch, (const char*)data+1, size-1, &opts);
++        if (status == 0 && patch) {
++                git_patch_free(patch);
++        }
++        return 0;
++}
--- a/fuzzers/patch_parse_fuzzer.c
+++ b/fuzzers/patch_parse_fuzzer.c
@@ -1,0 +1,38 @@
+/*
+ * libgit2 patch parser fuzzer target.
+ *
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#include "git2.h"
+#include "patch.h"
+#include "patch_parse.h"
+
+#define UNUSED(x) (void)(x)
+
+int LLVMFuzzerInitialize(int *argc, char ***argv)
+{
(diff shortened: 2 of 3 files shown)


fuzzers/corpora/patch_parse/patch_fuzzer-patch.diff
```diff
@@ -0,0 +1,45 @@
...
++                return 0;
++        }
++        git_patch* patch;
++        git_patch_options opts = {(uint32_t)data[0]};
++        int status = git_patch_from_buffer(&patch, (const char*)data+1, size-1, &opts);
++        if (status == 0 && patch) {
++                git_patch_free(patch);
++        }
++        return 0;
++}
```
pks-t: Heh, clever :P

fuzzers/patch_parse_fuzzer.c
```diff
@@ -0,0 +1,39 @@
...
+		abort();
+
+	return 0;
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+	if (size) {
+		git_patch *patch;
+		git_patch_options opts = {(uint32_t)data[0]};
```
pks-t: `git_patch_options opts = GIT_PATCH_OPTIONS_INIT`

fuzzers/patch_parse_fuzzer.c
```diff
@@ -0,0 +1,39 @@
...
+		git_patch_from_buffer(&patch, (const char *)data + 1, size - 1,
+		                      &opts);
```
pks-t: If this function returns an error, libgit2 doesn't guarantee that `patch` has been initialized to `NULL`. Thus you either need to check for the return value of this function or otherwise initialize `patch = NULL`

fuzzers/patch_parse_fuzzer.c
```diff
@@ -0,0 +1,39 @@
...
+		if (patch) {
+			git_patch_free(patch);
+		}
```
pks-t: No need for the conditional. `git_patch_free` handles `NULL` parameters just fine.

fuzzers/corpora/patch_parse/patch_fuzzer-patch.diff
durin42: I mean, I needed a sample diff, I had one right there, etc. :P

fuzzers/patch_parse_fuzzer.c
durin42: Done

fuzzers/patch_parse_fuzzer.c
durin42: Nice, done.

durin42:
Also, in case you find it useful, I formatted the new c code with this clang-format config:

```
BasedOnStyle: LLVM
IndentWidth: 8
UseTab: ForIndentation
BreakBeforeBraces: Linux
AllowShortIfStatementsOnASingleLine: false
IndentCaseLabels: false
AllowShortBlocksOnASingleLine: false
AllowShortFunctionsOnASingleLine: false
IncludeCategories:
  - Regex:           '^<'
    Priority:        1
  - Regex:           '^"'
    Priority:        2
```

It seems to roughly match what you're already doing (it's cribbed from Mercurial's config, which appears to be similar, I suspect they both originated with the Linux kernel.)

fuzzers/patch_parse_fuzzer.c
```diff
@@ -0,0 +1,38 @@
...
+
+	return 0;
+}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+	if (size) {
+		git_patch *patch = NULL;
+		git_patch_options opts = GIT_PATCH_OPTIONS_INIT;
+		opts.prefix_len = (uint32_t)data[0];
```
pks-t: Sorry, I somehow didn't see this in my first round, but I'm wondering about the first byte of `data` here. So you're skipping the first byte and use it as `prefix_len`, right? But judging from the fuzzing corpora, the first byte is simply part of the diff, and thus it is _very_ likely that the initial line is a non-valid diff header already. So shouldn't we just leave the `prefix_len` alone and just use `data` as input for `git_patch_from_buffer` directly?

E.g. just:

```
git_patch *patch = NULL;
git_patch_from_buffer(&patch, data, size, NULL);
git_patch_free(patch);
```

durin42:
The included corpora files should both have a leading 0x01. GitHub doesn't
seem to show that...

pks-t:
> The included corpora files should both have a leading 0x01. GitHub doesn't
seem to show that...

Ah, fair. Not too sure about that though, to be honest. I somehow guess that the fuzzers will generate a lot less valid patches and thus impact speed, especially so because it's unlikely that the fuzzer will generate correct prefixes depending on this prefix byte. For simplicity's sake I'd say to not have that and just keep the default options. I assume there will be enough things to be fixed already.

durin42:
From experience fuzzing Mercurial's parsers, the fuzzing system is absurdly
good at figuring out prefix bytes like this. But if you feel strongly I can
rip it out. What's your preference?

PR #5272, opened 2019-10-17
Various examples shape-ups

tiennou:
Here are some commits to the example code that shape it up a little bit, the kind of things that tends to be lost on various branches when debugging. Note that this hasn't been *really* tested — I don't expect any obviously broken things, but as I don't remember the actual goal of 3f68eff, for example, I'm not confident 😜.

As a side note, I've rebased the [example/pull WIP](https://github.com/tiennou/libgit2/compare/examples/pull) on top of that, as I've managed to pinpoint another lost-but-interesting commit with the tracking-remote creation helper started but incomplete.


tiennou:
Rebased, I took some time to correct all review comments tackled, and made a few more drastic changes — the "assert.h" cleanup is now global, since it's in both libgit2's and lg2's "common.h".

examples/add.c
```diff
@@ -133,33 +116,45 @@ void print_usage(void)
...
-			i++;
+		}
+		else if (!strcmp(args->argv[args->pos], "--")) {
+			args->pos++;
 			break;
-		} else {
-			fprintf(stderr, "Unsupported option %s.\n", argv[i]);
+		}
+		else {
+			fprintf(stderr, "Unsupported option %s.\n", curr);
```
pks-t: `else {` belongs on the same line as the closing brace

examples/args.c
```diff
@@ -1,3 +1,5 @@
+#include <assert.h>
```
pks-t: You're reintroducing an <assert.h> include here.

examples/lg2.c
```diff
@@ -10,7 +10,7 @@ struct {
 	git_command_fn fn;
 	char requires_repo;
 } commands[] = {
-	{ "add",          lg2_add,          1 },
+	{ "add",          lg2_add,    		1 },
```
pks-t: The conversion here looks wrong

examples/rev-list.c
```diff
@@ -15,23 +15,32 @@
 
 #include "common.h"
 
-static int revwalk_parseopts(git_repository *repo, git_revwalk *walk, int nopts, char **opts);
+#include <assert.h>
```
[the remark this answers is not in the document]
pks-t: This assert still exists :)

PR #5273, opened 2019-10-17
patch_parse: handle patches without extended headers

### LINKED ISSUE #5267: libgit2's patch parser could ignore missing `index [:hexdigit:]+..[:hexdigit:]+ lines
### Reproduction Steps

1. Generate a git-format diff with Mercurial, eg https://lists.sr.ht/~sircmpwn/sr.ht-dev/%3Caea421ff5fe46339493b.1570722314%40devahi%3E
2. Try to parse the diff with `git_patch_from_buffer()` or similar

### Expected Behavior
The patch parses fine.

### Actual Behavior
You get a parse error, of the form `invalid patch header at line NNN`.

### Investigation
I messed around with diff contents until I got past this particular parse error. If I add the line `index abcd..abcd` after the `diff --git ...` line and before the `--- a/...` line, the parse succeeds. That *appears* to be the only issue in patch-header parsing to allow hg-originated patches to parse just fine. I didn't get past that point in my investigation before other things required my attention, but I was convinced to at least file this bug.

### Version of libgit2
`ef5a3851f`

Reproduces on both Linux and macOS

Example:
The patch at https://lists.sr.ht/~sircmpwn/sr.ht-dev/%3Caea421ff5fe46339493b.1570722314%40devahi%3E, or any output from `hg export --git SHA1` should get you started.

dlax:
Extended header lines (especially the "index <hash>..<hash> <mode>") are
not required by "git apply" so it import patches. So we allow the
from-file/to-file lines (--- a/file\n+++ b/file) to directly follow the
git diff header.

This fixes #5267.

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -389,6 +389,7 @@ static const parse_header_transition transitions[] = {
 	{ "index "              , STATE_DIFF,       STATE_INDEX,      parse_header_git_index },
 	{ "index "              , STATE_END,        STATE_INDEX,      parse_header_git_index },
 
+	{ "--- "                , STATE_DIFF,       STATE_PATH,       parse_header_git_oldpath },
 	{ "--- "                , STATE_INDEX,      STATE_PATH,       parse_header_git_oldpath },
 	{ "+++ "                , STATE_PATH,       STATE_END,        parse_header_git_newpath },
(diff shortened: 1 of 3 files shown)


ddevault:
Thanks!

PR #5276, opened 2019-10-19
patch_parse: fixes for fuzzing errors

pks-t:
These are the first results of #5269.


pks-t:
I'm going ahead and merge this. The fixes should be rather obvious, and these bugs currently hinder progress of the fuzzer.

PR #5278, opened 2019-10-22
A collection of puzzling scripts

tiennou:
This is my private collection of macOS-oriented scripts. There's *many* things in here, some which are only loosely related to directed acyclic graphs, like abstract syntax tree mangling, virtual machine herding, buildsystem spelunking, etc. You know the drill :wink:.

I don't intend for this to go in like this, but it's somehow nicer if I file all this as a PR, and keep that commit always updated in the meantime. Maybe even close it, I just wanted to have some place to point questions at 😆.


ethomson:
I'm fascinated that you build with OpenSSL.  Is that just for testing, or is there some advantage to it over the SSL framework in macOS?

lolgear:
@tiennou 
I don't see `brew install` in your scripts... I would like to have a VM with any Linux distribution with desktop environment + IDE, I guess, or some good editor as `SublimeText`.
It would be nice to run tests on this VM and to kill it if something bad has happened. Far more better than develop in host OS.

tiennou:
So, normally running `./build.sh` with the correct Homebrew dependencies installed *should* result in a usable Xcode project.

If you want a Linux VM, the Vagrantfile has a bunch of mostly generic "start this image and run this script" (see `MACHINES` for a list of known "boxes"). The script is global, usually fails, and then it's a simple `vagrant ssh` and you should have an almost setup `~/libgit2` as the currently checked out branch.

This is completely unrelated to CI, which already supports running all tests using Azure and Dockerfiles, and is quite older than that that I'm only using locally.

ethomson:
Closing as stale

Issue #5281, opened 2019-10-23
config: crash when creating config snapshot


eaigner:
~~~
Hardware Model:      MacBookPro13,3
Code Type:       x86_64
Parent Process:  launchd [1]

Date/Time:       undefined
Launch Time:     undefined
OS Version:      Mac OS X 10.15.1 (19B77a)
Report Version:  104

Exception Type:  SIGSEGV
Exception Codes: SEGV_MAPERR at 0x4f2295f27810
Crashed Thread:  11

Thread 11 Crashed:
0   libsystem_platform.dylib             0x00007fff66136732 0x7fff66135000 + 5938
1   libsystem_c.dylib                    0x00007fff65fe6fdf 0x7fff65f90000 + 356319
2   Git                                  0x000000010e65b6ab stdalloc__strdup + 10
3   Git                                  0x000000010e6735fc git_config_entries_dup_entry + 91
4   Git                                  0x000000010e67387c git_config_entries_dup + 107
5   Git                                  0x000000010e674294 config_iterator_new + 51
6   Git                                  0x000000010e67682b config_readonly_open + 58
7   Git                                  0x000000010e67006f git_config_snapshot + 270
8   Git                                  0x000000010e6d6132 git_repository_config_snapshot + 33
9   Git                                  0x000000010e6d0824 git_remote_lookup + 259
10  Git                                  0x000000010e662a72 git_branch_upstream_name + 417
11  Git                                  0x000000010e662e26 git_branch_upstream + 85
~~~

eaigner:
Nah, sorry. Not my machine :(

pks-t:
Yeah, thought so. I'm currently investigating whether it's somehow possible to end up with a configuration entry without key, as that's my most likely lead right now.

pks-t:
Do you by any chance have concurrence access to the repository/config from multiple threads at the same time or is there a single thread that performs all git-related functionality?

pks-t:
Mh, okay. By now I assume that there might be a lifecycle issue somewhere due to improper use of refcounting of the config entries struct, mostly because I didn't find any way to allocate an entry config entry name at all.

pks-t:
I think I've found a very likely candidate, as there is a rather severe race condition when creating iterators. When one thread e.g. refreshes config values while another one wants to create an iterator, then we might have free'd the config entries while duplicating them because we didn't properly increase the refcount. This bug would be very likely to show the exact same trace you've posted.

Please see #5282 for a fix.

PR #5282, opened 2019-10-24
config_file: fix race when creating an iterator

pks-t:
When creating a configuration file iterator, then we first refresh the
backend and then afterwards duplicate all refreshed configuration
entries into the iterator in order to avoid seeing any concurrent
modifications of the entries while iterating. The duplication of entries
is not guarded, though, as we do not increase the refcount of the
entries that we duplicate right now. This opens us up for a race, as
another thread may concurrently refresh the repository configuration and
thus swap out the current set of entries. As we didn't increase the
refcount, this may lead to the entries being free'd while we iterate
over them in the first thread.

Fix the issue by properly handling the lifecycle of the backend's
entries via `config_file_entries_take` and `git_config_entries_free`,
respectively.

---

I think this may be the likely cause of #5281, as it fixes a severe race condition in exactly the code path that #5281 shows.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -250,17 +250,19 @@ static int config_file_iterator(
 	struct git_config_backend *backend)
 {
 	config_file_backend *b = GIT_CONTAINER_OF(backend, config_file_backend, parent);
-	git_config_entries *entries = NULL;
+	git_config_entries *dupped = NULL, *entries = NULL;
 	int error;
 
 	if ((error = config_file_refresh(backend)) < 0 ||
-	    (error = git_config_entries_dup(&entries, b->entries)) < 0 ||
-	    (error = git_config_entries_iterator_new(iter, entries)) < 0)
+	    (error = config_file_entries_take(&entries, b)) < 0 ||
+	    (error = git_config_entries_dup(&dupped, entries)) < 0 ||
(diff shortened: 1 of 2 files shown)


eaigner:
Does clar provide a facility to test async behavior?

pks-t:
On Sat, Oct 26, 2019 at 01:04:24AM -0700, Erik Aigner wrote:
> Does clar provide a facility to test async behavior?

Well, we do have some tests in "tests/threads", but that's about
it. Things is that "docs/threading.md" explicitly says:

    Unless otherwise specified, libgit2 objects cannot be safely
    accessed by multiple threads simultaneously.

Thus, things cannot be used from multiple threads at the same
time, at least officially. The truth is probably somewhat mixed,
as we tried to get some subsystems to behave just fine with
locking and refcounting. I'm always happy to fix any race
conditions as well, but one should always keep in mind our
official stance.

eaigner:
Yes, I know. Still, its hard to avoid race conditions if you are not using libgit in a commandline fashion, at which point one might just as well parse git output ;)

pks-t:
Rebased to fix conflicts with #5293

pks-t:
Thanks @eaigner and @tiennou for your reviews!

PR #5283, opened 2019-10-24
examples: checkout: implement guess heuristic for remote branches

### LINKED ISSUE #5277: add a simple example to show how to switch to an branch
git_clone seems to allow to clone directly on a branch.
But How to switch on branch after ?
stackoverflow does not have an working example.

Thanks to provide an simple example directly in your code
equivalent to:
git clone http:// ...
```
git branch -a
*  master
  remotes/origin/branch_dev
  remotes/origin/master
```
git checkout branch_dev

```
git branch -a
* branch_dev
  master
  remotes/origin/branch_dev
  remotes/origin/master
```

pks-t:
This pull request implements the guess heuristic used by upstream's git-checkout(1) and git-switch(1), respectively. It will automatically create local branches from existing remote branches if any ref exists that matches. Honestly, I think that this is kind of complex as part of our examples, and I feel like it might be sensible to provide e.g. a flag in `git_checkout` that enables the guessing heuristic.

Fixes #5277, I guess?

### DIFF
--- a/examples/checkout.c
+++ b/examples/checkout.c
@@ -112,9 +112,10 @@ static void print_perf_data(const git_checkout_perfdata *perfdata, void *payload
  * This is the main "checkout <branch>" function, responsible for performing
  * a branch-based checkout.
  */
-static int perform_checkout_ref(git_repository *repo, git_annotated_commit *target, checkout_options *opts)
+static int perform_checkout_ref(git_repository *repo, git_annotated_commit *target, const char *target_ref, checkout_options *opts)
 {
 	git_checkout_options checkout_opts = GIT_CHECKOUT_OPTIONS_INIT;
+	git_reference *ref = NULL, *branch = NULL;
 	git_commit *target_commit = NULL;
 	int err;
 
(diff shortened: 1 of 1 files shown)


ethomson:
> I feel like it might be sensible to provide e.g. a flag in git_checkout that enables the guessing heuristic.

Instead of a flag on checkout, I think a new API might be useful for callers to get the annotated commit from a "refish"?  (Is there a better name than that?)

tiennou:
I think I'd stash this kind of function under `branch`, eg. `git_branch_create_tracking` (which I have some dead code for somewhere), and then let the checkout/switch part use that ? Because AFAIU there's no need to touch the working copy to create a tracking branch (says the guy who just read about locally-tracking branches 🙄) ?

> annotated commit from a "refish"

I'm tempted to say that annotated commit actually are libgit2's "refish" — at least that's how the API felt when I needed to use merges/fetches/etc. Maybe I'm missing something though ?

pks-t:
I'm torn between both promises. I honestly never quite got what annotated commits are about, but I guess describing them as "refish" which remembers where the actual commit was resolved from is probably kinda close. I can't quite imagine how branch creation would look like if we were to use annotated commits for this case, so on first sight `git_branch_create_tracking` sounds more obvious to me. Might just be me though for not having seen the approach @ethomson has in mind.

Anyway, can we merge this? Any future additions to the API can lift code out of the examples as required.

tiennou:
I didn't as well, until I wrote [this example](https://github.com/libgit2/libgit2/blob/fb439c975a2de33f5b0c317f3fdea49dc94b27dc/examples/checkout.c#L114), and was enlightened about "how to convert refishs" in GitX. If that helps, this one is also a likely candidate for code-lifting.

Issue #4617, opened 2018-04-10
CMake: Transient dependency failures


tiennou:
Since The Great CMake Refactoring, it seems there are a few issues still when using the Xcode generator.

Symptoms are, when building a dependent "target" (IOW `libgit2_clar`, or one of the examples), you'll get linking failures (lots), because it seems the `git2` -> `git2_internal` dependency is ignored. Note that this can cause weird issues, since you could end up changing things in `git2_internal` *without* triggering a rebuild/relink of dependent targets (hilarious debugger sessions usually ensue).

You can workaround this by making sure dependencies are up to date yourself. Thanks, *build system* 😜.

tiennou:
I somewhat suspect the CMake generator to be at fault here (since AFAICS the normal one is not affected), but I haven't really checked…

pks-t:
That's awkward. Which version of CMake are you using?

tiennou:
Right now 3.11.0, but I've been coping with that for a while. For clarity, I'll define names :

`A` - "linking failure when forgetting git2_internal"
`B` - "weird debugging"
`C` - I'll add "changing testsuites doesn't invalidate things", so you can end up with missing make targets when switching branches, or linking failures when renaming testcases.

I'll try to come up with reproduction steps, C is the easiest, (I got it yesterday evening, workaround is to kill all clar related caches and re-run CMake, AFAICS it will not fix itself from Xcode) and the one I've seen only once is B (working on the CRLF/UTF-8 BOM stuff IIRC).

(And a CMake re-run means Xcode will revert to its 4-spaces default  indentation setting 😉)

PR #5284, opened 2019-10-25
CMake modernization

pks-t:
This morning, I set out to do some small cleanups for our CMake code. Naturally, it turned into a full-blown refactoring doing many things I didn't intend in the beginning. But if you ask me, the results are hugely beneficial and make our build code much much easier to read, self contained and correct.

So what changed? I've mostly modified all of our code to make heavy use of library targets. So instead of globally setting up required build options, link options are tied to each of these targets and will get applied when using `TARGET_LINK_LIBRARIES`, only.

Unfortunately, I realized later on that my code currently depends on features introduced in CMake v3.12, only. Most importantly, `TARGET_LINK_LIBRARIES` didn't work with object libraries before that, breaking one of my main assumptions. Thus CI builds will probably fail, too. I'm quite sure that I'm actually able to work around that issue on older versions, though, but I first wanted to get some comments.

As usual, I highly recommend to go through commit-wise. I think otherwise, you'll quickly get lost due to the huge diff.


src/CMakeLists.txt
```diff
@@ -54,6 +54,20 @@ IF (MSVC_IDE)
 	SET_SOURCE_FILES_PROPERTIES(win32/precompiled.c COMPILE_FLAGS "/Ycprecompiled.h")
 ENDIF ()
 
+# Collect sourcefiles
+FILE(GLOB LIBGIT2_SOURCES
```
tiennou: I want to point out that `file(GLOB …)` warns that :

>  If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.

This is somewhat unexpected, and sounds eerily like #4617, as the testsuite is also globbed. Maybe it would be worth to just flatten the list here ?

tiennou: (the commit message is missing a word at the end of the 3rd line)

CMakeLists.txt
```diff
@@ -14,8 +14,8 @@
 PROJECT(libgit2 C)
 CMAKE_MINIMUM_REQUIRED(VERSION 3.5.1)
 
-# Add find modules to the path
-LIST(APPEND CMAKE_MODULE_PATH ${libgit2_SOURCE_DIR}/cmake)
+# Prepend our own modules to the path
```
tiennou: I suspect something went wrong with the commit message here.

```
As we need to override some modules with our own version,
make sure we insert our custom location to CMake's module
path first.
```

src/CMakeLists.txt
```diff
@@ -132,25 +132,8 @@ IF (WIN32 AND EMBED_SSH_PATH)
...
-		LIST(APPEND LIBGIT2_PC_LIBS "-lwinhttp")
-	ENDIF ()
-
-	LIST(APPEND LIBGIT2_LIBS "rpcrt4" "crypt32" "ole32")
-	LIST(APPEND LIBGIT2_PC_LIBS "-lrpcrt4" "-lcrypt32" "-lole32")
-ENDIF()
-
-Include(SelectHTTPSBackend)
+LINK_DEPENDENCY(git2internal PUBLIC WinHTTP)
+LINK_DEPENDENCY(git2internal PUBLIC HttpsBackend)
```
tiennou: I may have missed the commit where `LINK_DEPENDENCY` was introduced…

cmake/LinkDependency.cmake
```diff
@@ -0,0 +1,4 @@
+FUNCTION(LINK_DEPENDENCY TARGET SCOPE DEPENDENCY)
```
tiennou: Called it 😉.

CMakeLists.txt
```diff
@@ -100,6 +100,11 @@ IF (DEPRECATE_HARD)
 	ADD_DEFINITIONS(-DGIT_DEPRECATE_HARD)
 ENDIF()
 
+# Installation paths
```
tiennou: Does it matter if you move it higher up ? I kinda like that the whole option setup is around the top of the file, as a documentation source. Maybe gathering all *known usable knobs* would ease some issues ? That's completely cosmetic though, but given that CMake code has a tendency to ~melt~ make you doubt from where it's coming from. Here, I was kinda hoping those variables were global CMake variables we're defining for clarity, now I don't know 😉.

src/CMakeLists.txt
```diff
@@ -15,13 +15,20 @@ SET(LIBGIT2_PC_LIBS "")
 ### Setup of library targets
 ###
 
-ADD_LIBRARY(git2internal OBJECT)
+# This is a workaround for old CMake versions which didn't yet
+# allow libraries without any source files.
```
tiennou: I'm somewhat surprised you need more dummies, for platforms others than macOS — that workaround is part of #4439, and it's unknown if recent versions of the Xcode generator work without it.

src/CMakeLists.txt
```diff
@@ -167,24 +167,34 @@ ADD_FEATURE_INFO(threadsafe THREADSAFE "threadsafe support")
 ### Detect and link against dependencies
 ###
 
+FUNCTION(OBJECT_LIBRARY_LINK TARGET)
+	IF(${CMAKE_VERSION} VERSION_LESS 3.13)
+		FOREACH(ARG IN LISTS ARGN)
+			SET_PROPERTY(TARGET ${TARGET} APPEND PROPERTY INTERFACE_LINK_LIBRARIES ${ARG})
+		ENDFOREACH()
+	ELSE()
+		TARGET_LINK_LIBRARIES(${TARGET} ${SCOPE} ${ARGN})
```
tiennou: I don't understand where `${SCOPE}` is coming from ?

src/CMakeLists.txt
```diff
@@ -167,24 +167,34 @@ ADD_FEATURE_INFO(threadsafe THREADSAFE "threadsafe support")
 ### Detect and link against dependencies
 ###
 
+FUNCTION(OBJECT_LIBRARY_LINK TARGET)
```
tiennou: I don't like the `OBJECT_LIBRARY_LINK` name, as we already have too much `TARGET` `LINK` and `LIBRARY` around (after just having read about the new `TARGET_LINK_OBJECTS` 🤣). I'd recommend sticking closer to the known `TARGET_LINK_LIBRARIES`, so `LG2_TARGET_LINK_LIBRARIES` ?

pks-t:
Closing as I've chosen a more iterative approach

PR #5289, opened 2019-10-30
commit: verify objects exist in git_commit_with_signature

### LINKED ISSUE #5288: git_commit_create_with_signature does not check the tree exists
### Reproduction steps

Create a signed commit via `git_commit_create_with_signature` where a parent or the tree does not exist.

### Expected behavior

(I guess more desired behaviour) the function call fails because the data fails to validate

### Actual behavior

No verification is done and now I have a corrupt repository.

All that have this functionality AFAICT, I'm looking at ef5a3851fdece852569ffebf3537883223744a7a

Observed on Debian, but the code affects everything.

#### Notes

We do perform a check in `git_commit_create_buffer` which is all well and good if what you're doing is shelling out to `gpg` or something on the same host. When your keys are in a different system and need everything to happen over network calls in a distributed and replicated system, you might find that the host you ran it on has the object but it's missing for some.

One option is to add checking to the system we have calling this function, but at another level, the library has no idea where the buffer we pass into `git_commit_create_with_signature` comes from and it could also just be something I created with an awk script.

carlosmn:
There can be a significant difference between the system where we created the                                                                              
buffer (if at all) and when the caller provides us with the contents of a                                                                                  
commit.

Verify that the commit we are being asked to create references objects which do
exist in the target repository.

This is not terribly efficient since we allocate a commit, but AFAICT it's what we have available. Obviously our test suite fails with these changes because we were creating commits that referenced non-existent trees and commits.

Fixes #5288

### DIFF
--- a/src/commit.c
+++ b/src/commit.c
@@ -878,6 +878,14 @@ static void format_header_field(git_buf *out, const char *field, const char *con
 	git_buf_putc(out, '\n');
 }
 
+static const git_oid *commit_parent_from_commit(size_t n, void *payload)
+{
+	const git_commit *commit = (const git_commit *) payload;
+
+	return git_array_get(commit->parent_ids, n);
+
+}
+
 int git_commit_create_with_signature(
 	git_oid *out,
 	git_repository *repo,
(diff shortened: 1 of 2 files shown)


src/commit.c
```diff
@@ -890,12 +898,26 @@ int git_commit_create_with_signature(
...
 
-	/* We start by identifying the end of the commit header */
+	/* The first step is to verify that all the tree and parents exist */
+	parsed = git__calloc(1, sizeof(git_commit));
+	GIT_ERROR_CHECK_ALLOC(parsed);
+	if ((error = commit_parse(parsed, commit_content, strlen(commit_content), 0)) < 0)
+		goto cleanup;
+
+	if ((error = validate_tree_and_parents(&parents, repo, &parsed->tree_id, commit_parent_from_commit, parsed, NULL, true)) < 0)
+		goto cleanup;
```
mhagger: Is it certain that `parents` doesn't have to be cleared if `validate_tree_and_parents` returns an error?

ethomson: Looks like it to me - the error case of `validate_tree_and_parents` fn will clear the parents array before returning, if it had ever allocated anything into it.

ethomson:
LGTM.  Thanks, @carlosmn, for finding and fixing strange ordering bugs!

ethomson:
It looks like the existing `verify` function doesn't honor `GIT_OPT_ENABLE_STRICT_OBJECT_CREATION`?  Or am I missing something?

This limitation doesn't hold this up, since it's an important fix.  But I wanted to ask.

carlosmn:
> It looks like the existing `verify` function doesn't honor `GIT_OPT_ENABLE_STRICT_OBJECT_CREATION`? Or am I missing something?

Do you mean `validate_tree_and_parents`? The `GIT_OPT_ENABLE_STRICT_OBJECT_CREATION` option, aka `git_object__strict_input_validation` is checked within `git_object__is_valid` so even though we're passing `true` for `validate` it's still down to the setting whether we actually perform the check.

ethomson:
> Do you mean validate_tree_and_parents? The GIT_OPT_ENABLE_STRICT_OBJECT_CREATION option, aka git_object__strict_input_validation is checked within git_object__is_valid so even though we're passing true for validate it's still down to the setting whether we actually perform the check.

Yes, somehow I forgot how variables work and boolean logic.  Thanks!

PR #5285, opened 2019-10-26
Follow 308 redirect in WinHTTP transport

pcpthm:
CC #4675. The issue is closed by #4848 but WinHTTP was forgotten.
The issue caused failures of GitLab repository fetching on Windows.

PR #5275, opened 2019-10-18
reflogs: fix behaviour around reflogs with newlines

pks-t:
This pull request has been inspired by #5266 and aims to fix two things:

1. We fail to parse corrupted reflogs, which git is reading just fine.
2. We corrupt the reflog if somebody calls `git_stash_save` with a commit message containing newlines, as we print the newlines verbosely into the stash's reflog.

This pull request fixes both issues. It makins our parser more lenient with regards to broken lines in the reflog. Last, it will sanitize reflog messages to not contain any newlines anymore. While we previously rejected such messages in one location, we didn't at another place in libgit2 (refdb_fs). And as git just replaces newlines with spaces, I've modified both places to accept reflog messages with newlines, but replace them with spaces now.

PR #5291, opened 2019-11-01
Release 0.99

ethomson:
I want to release an 0.99 in advance of a 1.0.

This is basically a 1.0 release candidate *but* we have historically not seen a lot of people installing "release candidates" in production and using them seriously.  We want Serious Work on this release candidate, so let's call it an 0.99, and a real release.

After that, I propose that we take a branch for 1.0 and keep working in master on post-1.0 work.

What else are we waiting on for 0.99:
- [x] #5123 
- [x] #5285 

**Cut**:
~~#5286 - this is an annoying bug if you hit it _but_ it's very rare, so I'm tempted to just kick it out to post-1.0~~


lolgear:
@ethomson 
Hi!

- [ ] Annoying issue #5266 is fixed in PR #5275. 

Could you also include this PR in this release?

ethomson:
@lolgear Done.  Thanks!

pks-t:
Sounds good to me. We're long overdue for a new release anyway, and there has been quite a lot of work since the last one. I'll make sure to have another go at #5123 this week to get it merged. Not sure whether I can provide valuable input for #5285 or 5286, but I'll see.

I think we should aim to get #5282 merged, too, and maybe even #5243. Otherwise I'm fine.

tiennou:
Here's my "nomination" list I would have personally liked:

- #4620, to let users see the hook support, and maybe have some more supported ones soon ?
- #5105, shallow-fetch support preparation, a small one that takes on too much for its own post-1.0 good.
- #4316, for the incomplete reflogs and data destruction bug.

lolgear:
@ethomson 
Hi in 2020 :)
Is there a plan to make a release soon?

ethomson:
I've updated this with more recent changes and authors (as of `cc4f4cbea48ac00a5edec1b3570ac3d2ef10fe77`).

pks-t:
> Is there a plan to make a release soon?

The final piece we're waiting on is #5286. It's probably going to get merged soonish, after which we'll wait one or two weeks to let it settle before we release v0.99.

ethomson:
@pks-t am i okay to merge this or would it break your release script with the version updates?

pks-t:
> @pks-t am i okay to merge this or would it break your release script with the version updates?

You mean with the autors amended? That's fine, the script currently only searches until the next version section and includes everything between current and previous header. Would mean that these authors are part of the release notes, but I consider it a good thing to give credit where credit is due.

Anyway, one thing I notice is that my scripting is currently too strict. It verifies that for a version X, that the release section's previous release is X-1. So e.g. for v0.27.1 it would be v0.27.0 or v0.27, for v0.28 it would be v0.27.x. With the jump to v0.99 we're breaking this assumption.

Furthermore, the current assumption is that the soversion matches our minor version, which obviously also breaks with v0.99. Guess I should just remove that check altogether and only verify MAJOR/MINOR/REVISION. Anyway, I guess I'll fine tune the release script anyway as soon as we're ready to go, and a second pair of eyes would be a good thing, too.

pks-t:
> Anyway, I guess I'll fine tune the release script anyway as soon as we're ready to go, and a second pair of eyes would be a good thing, too.

Ah, didn't see you've done a review already.

ethomson:
> You mean with the autors amended? 

I was worried about the version updates that I'd done in `include/git2/version.h`.

pks-t:
On Sun, Feb 02, 2020 at 04:43:20AM -0800, Edward Thomson wrote:
> > You mean with the autors amended? 
> 
> I was worried about the version updates that I'd done in `include/git2/version.h`. 

You can always run the script with "--dryrun". It will tell if
there's anything it doesn't like, but won't do any uploading
stuff. Anyway, I'll update the script to loosen the version check
for our changelog and use the MAJOR.MINOR soversion.

include/git2/version.h
```diff
@@ -7,12 +7,12 @@
...
-#define LIBGIT2_VERSION "0.28.0"
+#define LIBGIT2_VERSION "0.99.0"
 #define LIBGIT2_VER_MAJOR 0
-#define LIBGIT2_VER_MINOR 28
+#define LIBGIT2_VER_MINOR 99
 #define LIBGIT2_VER_REVISION 0
 #define LIBGIT2_VER_PATCH 0
 
-#define LIBGIT2_SOVERSION 28
+#define LIBGIT2_SOVERSION "0.99"
```
pks-t: Hum. Any reason why we don't go all the way and say "0.99.0", matching LIBGIT2_VERSION?

ethomson: I don't think that we want our soversion to include teeny version numbers - IMO, we shouldn't be making ABI incompatible changes in teeny releases.  Are there other considerations that we should be thinking about?

With this change, we also get a nice set up of our library links on the filesystem:

```
 1852 -rwxr-xr-x   1 ethomson  staff  1895388 Feb  3 10:51 libgit2.0.99.0.dylib*
    0 lrwxr-xr-x   1 ethomson  staff       20 Feb  3 10:51 libgit2.0.99.dylib@ -> libgit2.0.99.0.dylib
    0 lrwxr-xr-x   1 ethomson  staff       18 Feb  3 10:51 libgit2.dylib@ -> libgit2.0.99.dylib
```

pks-t:
On Mon, Feb 03, 2020 at 02:53:02AM -0800, Edward Thomson wrote:
> 
> 
> 
> >  #define LIBGIT2_VER_REVISION 0
>  #define LIBGIT2_VER_PATCH 0
>  
> -#define LIBGIT2_SOVERSION 28
> +#define LIBGIT2_SOVERSION "0.99"
> 
> I don't think that we want our soversion to include teeny version numbers - IMO, we shouldn't be making ABI incompatible changes in teeny releases.  Are there other considerations that we should be thinking about?
> 
> With this change, we also get a nice set up of our library links on the filesystem:
> 
> ```
>  1852 -rwxr-xr-x   1 ethomson  staff  1895388 Feb  3 10:51 libgit2.0.99.0.dylib*
>     0 lrwxr-xr-x   1 ethomson  staff       20 Feb  3 10:51 libgit2.0.99.dylib@ -> libgit2.0.99.0.dylib
>     0 lrwxr-xr-x   1 ethomson  staff       18 Feb  3 10:51 libgit2.dylib@ -> libgit2.0.99.dylib

Fair enough. Quoting CMake's documentation [1]:

    For shared libraries and executables on Windows the VERSION
    attribute is parsed to extract a <major>.<minor> version
    number. These numbers are used as the image version of the
    binary.

So CMake already expects "major.minor", which is another argument
in favor of leaving out "revision". I've updated the release
script to reflect that.

[1]: https://cmake.org/cmake/help/v3.15/prop_tgt/SOVERSION.html

docs/changelog.md
```diff
@@ -1,5 +1,283 @@
-v0.28 + 1
+v0.99
 ---------
```
pks-t: The script expects a newline between "----" and the body and complains if it doesn't find it. Initially, I've also had a check that verifies that section header and version are of equal lengths, but I've dropped that

pks-t:
I've made one amendment to fix up the missing newline after the version header in our changelog. While at it, I've squashed commits so that we only have one for our changelog and one for bumping the version.

Issue #5292, opened 2019-11-01
Creating a snapshop of a snapshot caused null pointer dereference


csware:
### Reproduction steps

1. Create a config snapshot.
2. Create a snapshot of the snapshot (e.g., set a snapshot config using `git_repository_set_config` and then call `git_repository_set_head(repo, "refs/heads/other")` or `git_repository_hashfile`).

### Expected behavior

Creating a snapshot of a snapshot works.

Why is snapshotting a `config_mem` not possible? I though snapshotting can always be used to have an in memory read-only representation of any config which can be used everywhere (e.g., to avoid disk access).

### Actual behavior

Null-pointer dereference in `git_repository_config_snapshot`, because `snapshot` is not set for the snapshot backend.

### Version of libgit2 (release number or SHA1)

current master (tested on 257dd59d031f169a55afb55d0c34527fac71c11e)

pks-t:
In #5293 you said it would partly fix the error reported here. So I wonder whether there is there a second part that needs fixing?

csware:
`config_mem` cannot be snapshotted at all right now.

pks-t:
See #5299

ethomson:
Fixed via #5299

PR #5293, opened 2019-11-01
Fix crash if snapshotting a config_snapshot

csware:
Partly fixes issue #5292.


csware:
Another approach (maybe with better performance) could be to just increase the reference counter and return the very same instance for `config_snapshots`.

ethomson:
Interesting!  It's snapshots all the way down!

turtle

I don't see why we wouldn't allow a snapshot to be created from another snapshot.  This seems like a reasonable fix.  I agree that it's not as performant as it _could_ be but I also think that this is an uncommon enough use case (you're obviously the first person to try it 😀) that I'm happy enough to worry about the performance if/when it becomes an issue that you encounter.  That is to say, if all users are happy enough with the perf, then I'm happy enough with the perf.  😄 

I'd love @pks-t's thoughts.

PR #5296, opened 2019-11-05
"Generic" repository layout

tiennou:
This adds a low-level "layout" structure/subsystem, which inherits the current "filesystem resolution process" (a.k.a `git_repository_item_path()`), and most of the "names" `git` uses. This allows the rest of the code to ask about "where do I find the data for $item", and be less knowledgable about any filesystem involvement ([example](https://github.com/libgit2/libgit2/compare/master...tiennou:feature/repository-layout?expand=1#diff-2818d3d12d3a4a63c94f2bd6d54d8afaR31)). The longterm goal is clearly to deprecate `git_repository_new` and provide more easily-customizable options at a lower level, but that's out of scope for now — I just want the repository-opening step to be less confusing, and I want more control over what's considered valid and/or overridable 😉.

This is actually a partial rebase of #4856, with a focus on refactoring/cleanups instead of feature-filling — mostly because IMHO there are low-level details that have completely conflicting semantics, and I can't manage to make sense of them when trying to preserve our own "rules" w.r.t to "opening from the environment". I don't know how much of it is in scope for #5291, hence I'm only drafting this 😉. Note that I don't intend to replace #4856 yet, since it's somehow closer to completion than this, but I'll either rebase later or polish up, fix and 🚢 if there's a need for it.


tiennou:
And, without further ado, here's my notes/questions about this undertaking :

- [This](https://github.com/libgit2/libgit2/pull/4856#pullrequestreview-301181223) has some insights on the needs for a refactor.

The TL;DR is that, between the short-circuited codepaths of `bare`, `env`, `gitlink`, and the `bare`, `normal`, `custom refdb/odb/config`, `new`, and `worktree`/`submodule` repositories types, it's confusing.

- This is still 🔴. It's likely more crazy filesystem testcases are needed to polish up the behavior — IIRC there's already an issue about a path not being absolute in some case.

- I stared a hard time at the additions to the private `items` array.

I don't think "items" like `HEAD`, `FETCH_HEAD`, and friends belong in here : IMHO those should be part of the refdb layer, because it would benefit from "an automated write this ref with this data when you commit this transaction" support, which refdb already has. But right now there's no such support for "freeform storage" — our refdb is only interested in OIDs, symrefs, and reflogging, and all other uses usually go through the filesystem, bypassing any custom refdb backend black magic (there are accesses to `.git/HEAD` still).

Some others are less clear — I'm aware of "info/grafts" but it's along other files in "info", and the "*_MSG" ones are also kinda temporary ones… But there's still a transaction feel to them 🙄.

Thus, I've chosen to add `git_repository_item_t` entries, which are public. I'd like to know if that makes sense for all of them 😉.

tiennou:
Annnnd I've de-drafted it 🤣.

PR #5298, opened 2019-11-06
patch_parse: fix segfault when header path contains whitespace only

pks-t:
When parsing header paths from a patch, we reject any patches with empty
paths as malformed patches. We perform the check whether a path is empty
before sanitizing it, though, which may lead to a path becoming empty
after the check, e.g. if we have trimmed whitespace. This may lead to a
segfault later when any part of our patching logic actually references
such a path, which may then be a `NULL` pointer.

Fix the issue by performing the check after sanitizing. Add tests to
catch the issue as they would have produced a segfault previosuly.

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -69,27 +69,24 @@ static int parse_header_path_buf(git_buf *path, git_patch_parse_ctx *ctx, size_t
 {
 	int error;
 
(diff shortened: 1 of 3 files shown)


ethomson:
Nice one.  Thanks @pks-t!

PR #5299, opened 2019-11-06
config_mem: implement support for snapshots

pks-t:
Similar as in commit dadbb33b6 (Fix crash if snapshotting a
config_snapshot, 2019-11-01), let's implement snapshots for in-memory
configuration entries. As this deletes more code than it adds, it
doesn't make any sense to not allow for this and allows users to treat
config backends mostly the same.


ethomson:
Good call.  Thanks @pks-t!

PR #5300, opened 2019-11-06
branch: clarify documentation around branches

tiennou:
This comes after looking around at #5283, just a few documentation and commentation done.

It seems there's no "create the tracking branch for that branch using that remote" function, which is the low-level operation that's missing from people workflows — right now you can use `git_branch_set_upstream` to do the "verifications & configuration updating", but you have to have created a reference deemed worthwhile of that code to succeed, which is 😒.

### DIFF
--- a/include/git2/branch.h
+++ b/include/git2/branch.h
@@ -126,6 +126,12 @@ GIT_EXTERN(void) git_branch_iterator_free(git_branch_iterator *iter);
  * The new branch name will be checked for validity.
  * See `git_tag_create()` for rules about valid names.
  *
+ * Note that if the move succeeds, the old reference object will not
+ + be valid anymore, and should be freed immediately by the user using
+ + `git_reference_free()`.
+ *
+ * @param out New reference object for the updated name.
+ *
  * @param branch Current underlying reference of the branch.
  *
  * @param new_branch_name Target name of the branch once the move
--- a/include/git2/branch.h
+++ b/include/git2/branch.h
@@ -75,9 +75,9 @@ GIT_EXTERN(int) git_branch_create_from_annotated(
 /**
  * Delete an existing branch reference.
  *
- * If the branch is successfully deleted, the passed reference
- * object will be invalidated. The reference must be freed manually
- * by the user.
+ * Note that if the deletion succeeds, the reference object will not
+ * be valid anymore, and should be freed immediately by the user using
+ * `git_reference_free()`.
  *
  * @param branch A valid reference representing a branch
  * @return 0 on success, or an error code.
@@ -145,17 +145,14 @@ GIT_EXTERN(int) git_branch_move(
  * Lookup a branch by its name in a repository.
  *
  * The generated reference must be freed by the user.
- *
  * The branch name will be checked for validity.
- * See `git_tag_create()` for rules about valid names.
+ *
+ * @see git_tag_create for rules about valid names.
  *
  * @param out pointer to the looked-up branch reference
- *
  * @param repo the repository to look up the branch
- *
  * @param branch_name Name of the branch to be looked-up;
  * this name is validated for consistency.
- *
  * @param branch_type Type of the considered branch. This should
  * be valued with either GIT_BRANCH_LOCAL or GIT_BRANCH_REMOTE.
  *
@@ -169,65 +166,74 @@ GIT_EXTERN(int) git_branch_lookup(
 	git_branch_t branch_type);
 
 /**
- * Return the name of the given local or remote branch.
+ * Get the branch name
  *
- * The name of the branch matches the definition of the name
- * for git_branch_lookup. That is, if the returned name is given
- * to git_branch_lookup() then the reference is returned that
- * was given to this function.
+ * Given a reference object, this will check that it really is a branch (ie.
+ * it lives under "refs/heads/" or "refs/remotes/"), and return the branch part
+ * of it.
  *
- * @param out where the pointer of branch name is stored;
- * this is valid as long as the ref is not freed.
- * @param ref the reference ideally pointing to a branch
+ * @param out Pointer to the abbreviated reference name.
+ *        Owned by ref, do not free.
  *
- * @return 0 on success; otherwise an error code (e.g., if the
- *  ref is no local or remote branch).
+ * @param ref A reference object, ideally pointing to a branch
+ *
+ * @return 0 on success; GIT_EINVALID if the reference isn't either a local or
+ *         remote branch, otherwise an error code.
  */
 GIT_EXTERN(int) git_branch_name(
 		const char **out,
 		const git_reference *ref);
 
 /**
- * Return the reference supporting the remote tracking branch,
- * given a local branch reference.
+ * Get the upstream of a branch
  *
- * @param out Pointer where to store the retrieved
- * reference.
+ * Given a reference, this will return a new reference object corresponding
+ * to its remote tracking branch. The reference must be a local branch.
  *
+ * @see git_branch_upstream_name for details on the resolution.
+ *
+ * @param out Pointer where to store the retrieved reference.
  * @param branch Current underlying reference of the branch.
  *
  * @return 0 on success; GIT_ENOTFOUND when no remote tracking
- * reference exists, otherwise an error code.
+ *         reference exists, otherwise an error code.
  */
 GIT_EXTERN(int) git_branch_upstream(
 	git_reference **out,
 	const git_reference *branch);
 
 /**
- * Set the upstream configuration for a given local branch
+ * Set a branch's upstream branch
+ *
+ * This will update the configuration to set the branch named `name` as the upstream of `branch`.
+ * Pass a NULL name to unset the upstream information.
+ *
+ * @note the actual tracking reference must have been already created for the
+ * operation to succeed.
  *
  * @param branch the branch to configure
+ * @param branch_name remote-tracking or local branch to set as upstream.
  *
- * @param upstream_name remote-tracking or local branch to set as
- * upstream. Pass NULL to unset.
- *
- * @return 0 or an error code
+ * @return 0 on success; GIT_ENOTFOUND if there's no branch named `branch_name`
+ *         or an error code
  */
-GIT_EXTERN(int) git_branch_set_upstream(git_reference *branch, const char *upstream_name);
+GIT_EXTERN(int) git_branch_set_upstream(
+	git_reference *branch,
+	const char *branch_name);
 
 /**
- * Return the name of the reference supporting the remote tracking branch,
- * given the name of a local branch reference.
+ * Get the upstream name of a branch
  *
- * @param out Pointer to the user-allocated git_buf which will be
- * filled with the name of the reference.
+ * Given a local branch, this will return its remote-tracking branch information,
+ * as a full reference name, ie. "feature/nice" would become
+ * "refs/remote/origin/feature/nice", depending on that branch's configuration.
  *
- * @param repo the repository where the branches live
- *
+ * @param out the buffer into which the name will be written.
+ * @param repo the repository where the branches live.
  * @param refname reference name of the local branch.
  *
- * @return 0, GIT_ENOTFOUND when no remote tracking reference exists,
- *     otherwise an error code.
+ * @return 0 on success, GIT_ENOTFOUND when no remote tracking reference exists,
+ *         or an error code.
  */
 GIT_EXTERN(int) git_branch_upstream_name(
 	git_buf *out,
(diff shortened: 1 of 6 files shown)


include/git2/branch.h
```diff
@@ -169,116 +166,127 @@ GIT_EXTERN(int) git_branch_lookup(
...
- * Set the upstream configuration for a given local branch
+ * Set a branch's upstream branch
  *
- * @param branch the branch to configure
+ * This will update the configuration to set the branch named `name` as the upstream of `branch`.
+ * Pass a NULL name to unset the upstream information.
  *
- * @param upstream_name remote-tracking or local branch to set as
- * upstream. Pass NULL to unset.
+ * @note the actual tracking reference must have been already created for the
```
tiennou: I was tempted to change `git_branch_set_upstream` to take a `git_reference` instead of a `const char *`, to make it more usable — it makes it clear that the reference *must* exist beforehand. But that's API-breaking, so…

include/git2/branch.h
```diff
@@ -169,116 +166,127 @@ GIT_EXTERN(int) git_branch_lookup(
...
 	const git_reference *branch);
 
 /**
- * Return the name of remote that the remote tracking branch belongs to.
+ * Extract the remote name of a branch
  *
- * @param out Pointer to the user-allocated git_buf which will be filled with the name of the remote.
+ * This will returns the name of the remote a remote-tracking branch belongs to.
+ * IOW, it extracts "test" from "refs/remotes/test/master", checks all remotes
+ * for a compatible refspec, and returns the first match.
```
tiennou: This one is the one I was most surprised by — I didn't expect a round-trip through the list of remotes to happen here…

I'm tempted to provide a lower-level version under that name and rename this one to be less "accessor"-y, so that you can get the name quickly-ish. Opinions ?

src/branch.c
```diff
@@ -597,97 +597,99 @@ static int unset_upstream(git_config *config, const char *shortname)
 	return -1;
 }
 
-int git_branch_set_upstream(git_reference *branch, const char *upstream_name)
+int git_branch_set_upstream(git_reference *branch, const char *branch_name)
 {
-	git_buf key = GIT_BUF_INIT, value = GIT_BUF_INIT;
+	git_buf key = GIT_BUF_INIT, remote_name = GIT_BUF_INIT, merge_refspec = GIT_BUF_INIT;
```
tiennou: Only NFC in this hunk — hopefully, I made use of Xcode's Refactor 😉.

include/git2/branch.h
```diff
@@ -169,116 +166,127 @@ GIT_EXTERN(int) git_branch_lookup(
...
- * was given to this function.
+ * Given a reference object, this will check that it really is a branch (ie.
+ * it lives under "refs/heads/" or "refs/remotes/"), and return the branch part
+ * of it.
  *
- * @param out where the pointer of branch name is stored;
- * this is valid as long as the ref is not freed.
- * @param ref the reference ideally pointing to a branch
+ * @param out Pointer to the underlying, truncated, reference name.
+ *        Owned by ref, do not free.
```
pks-t: Truncated sounds a bit weird here. I'm not a native speaker, but doesn't truncation indicate that we're chopping something off of the end?

include/git2/branch.h
pks-t: I agree, the interface is weird. We could create `git_branch_set_upstream_ext` as usual, but nobody's been asking for it.

include/git2/branch.h
```diff
@@ -169,116 +166,127 @@ GIT_EXTERN(int) git_branch_lookup(
...
  */
 GIT_EXTERN(int) git_branch_is_checked_out(
 	const git_reference *branch);
 
 /**
- * Return the name of remote that the remote tracking branch belongs to.
+ * Extract the remote name of a branch
  *
- * @param out Pointer to the user-allocated git_buf which will be filled with the name of the remote.
+ * This will returns the name of the remote a remote-tracking branch belongs to.
```
pks-t: Typo: "will returns"

include/git2/branch.h
```diff
@@ -169,116 +166,127 @@ GIT_EXTERN(int) git_branch_lookup(
...
+ * IOW, it extracts "test" from "refs/remotes/test/master", checks all remotes
```
pks-t: I'd refrain from using abbreviations like "IOW" to help people who aren't native speakers

include/git2/branch.h
pks-t: At first this does sound weird to me, too. But you could have a remote that isn't named anything like what the refspecs will do. E.g. the remote "foobar" could definitely fetch into "barfoo" if the refspec is configured like that.

So we definitely need this roundtrip for correctness, but I wonder about extracting "test" first. Shouldn't we use the reference as-is and directly check it against refspecs, returning the first remote with a matching one?

pks-t: And by the way, you document the function to return `GIT_EAMBIGUOUS` in there are several matching refspecs, which sounds reasonable to me. But here in the description you say that we're returning the first matching remote, which directly contradicts that.

src/branch.c
```diff
@@ -597,97 +597,99 @@ static int unset_upstream(git_config *config, const char *shortname)
 	return -1;
 }
 
-int git_branch_set_upstream(git_reference *branch, const char *upstream_name)
+int git_branch_set_upstream(git_reference *branch, const char *branch_name)
```
pks-t: To not loose the upstream-y nature, what about `upstream_branch`? The other renames are definitely helpful, though.

src/branch.c
```diff
@@ -597,97 +597,99 @@ static int unset_upstream(git_config *config, const char *shortname)
...
 	}
 
-	/*
-	 * If it's local, the remote is "." and the branch name is
-	 * simply the refname. Otherwise we need to figure out what
-	 * the remote-tracking branch's name on the remote is and use
-	 * that.
+	/* Now we need to "guess" that branch's remote name. A "local tracking
+	 * branch", since it tracks a branch in the current repository, get to use
+	 * ".". Otherwise, git_branch_remote_name knows what to do.
```
pks-t: To be honest, I find the old comment easier to parse.

include/git2/branch.h
tiennou: True, though I'm not as well, and I'm realizing I lack an "anti-synonym" for truncation 😆. This is actually the ref name, offset by its "namespace" length, so `Pointer to the underlying reference name, with some offset added.` or something ?

include/git2/branch.h
tiennou: I'm not sure about the roundtrips. IIRC reading the code left me with the feeling that a branch has a "belongs to" relation to its remote (hence the `"." means "the current repository's branches"` comment), and that some of those roundtrips aren't actually correct. For example, you can load a branch using its underlying reference name, but a subsequent operation with it will fail because this one will notice an ambiguity, or the configuration is somehow "broken" — the "branch.*" configuration key has been renamed, or its refspecs changed; I've been guilty of doing both, and I know GitX trusts the refdb for its remote name by reading `git for-each-ref` instead of doing this, hence all the suspicion 😉.

src/branch.c
tiennou: Not sure. My reasoning was that "upstream" is only meaningful when it's actually set, and that you call this function with a branch handy. To me, "upstream" muddies the waters by not being really what's needed here, ie, you could take a `git_remote *` object, and it'd make more sense.

Maybe that's what I'm misunderstanding — so, just to make sure I'm not, here's my "glossary":
- a "branch" is a reference in some "well-known" refdb location (either `refs/heads/` for a local/`"."` branch, or `refs/remotes/`).
- its "branch name" is the part under `refs/heads/` for a local branch, or `refs/remotes/*/` for a remote one.
- a "tracking branch" has a config section, that specifies its "upstream configuration", lives in `refs/heads/` (not sure if that's always true), and should (ideally) resolve to another reference under those configuration rules.
- "upstream" is the "logical" remote, known via the "rev-refspec" resolution.
- "upstream name" is the rev-refspec branch name (the ref's shortname, a.k.a. the refname, offset by "refs/remote/$rev-refspec/").
- "upstream branch" is the reference with that full rev-refspec name, a.k.a. the branch's tracking reference/branch.

I think what's confusing is that a branch is both "a reference name" at lookup, but "a name from a config key" later on, and the API reflects this confusion. IMHO, this could be fixed by just making `git_branch_lookup` more consistent — say `git_branch_lookup(git_reference **out, git_repository *repo, const char *branch_name, git_remote *remote)`. Hence, you pass a `NULL` remote for a "local, maybe tracking" branch, and remote ones can be checked for validity using the refspecs.

src/branch.c
tiennou: True. I'll rephrase as `If it's a local-tracking branch, its remote is "." (as "the local repository"), and the branch name is simply the refname`, and keep the rest as-is (I was confused only by "local" as "existing in the current repo" — that's true of remote branches as well 😉).

include/git2/branch.h
pks-t: What about "Pointer to the abbreviated reference name"? That would match gitglossary(7):

    ref
        A name that begins with refs/ (e.g.  refs/heads/master) that points
        to an object name or another ref (the latter is called a symbolic
        ref). For convenience, a ref can sometimes be abbreviated when used
        as an argument to a Git command; see gitrevisions(7) for details.
        Refs are stored in the repository.

src/branch.c
pks-t: Except for upstream name I think you've got everything right. For "upstream name" itself, I don't think that it actually has a well defined meaning.

The thing that bothers me is that `branch` and `branch_name` are so similar to each other, giving some false signals at first. The thing is we _are_ about to set the upstream in this function, right? That's why I thought that `upstream_branch` would be fine. I'm definitely open to better proposals, though.

tiennou: > Except for upstream name I think you've got everything right. For "upstream name" itself, I don't think that it actually has a well defined meaning.

I'm clearly unsure as well, but it seems to describe what's provided, so 👍 ? "Upstream branch name" would be a more clear version, but it's a mouthful…

> The thing that bothers me is that branch and branch_name are so similar to each other, giving some false signals at first.

Agreed. I take `branch_name` as things like "master", or "origin/master". `branch` is still a reference under a "well-known" location. To me, `upstream_branch` refers to the end goal, and because it needs to be understood to be provided correctly (and given that branches are so transparent already), it's more confusing, as I think it's the first parameter that takes a "branch name" without actually being named `branch_name`. Thanks for the feedback though, I have some good documentation fodder now for naming things correctly 😉.

include/git2/branch.h
```diff
@@ -169,116 +171,132 @@ GIT_EXTERN(int) git_branch_lookup(
...
- * @param out Pointer to the user-allocated git_buf which will be
- * filled with the name of the reference.
+ * Get the upstream name of a branch
  *
- * @param repo the repository where the branches live
+ * Given a local branch, this will return its remote-tracking branch information,
+ * as a full reference name, ie. "feature/nice" would become
+ * "refs/remote/origin/feature/nice", depending on that branch's configuration.
  *
+ * @param out The buffer into which the name will be written.
```
pks-t: "The" needs to be lower case for consistency

include/git2/branch.h
```diff
@@ -169,116 +171,132 @@ GIT_EXTERN(int) git_branch_lookup(
...
- * Determine if the current local branch is pointed at by HEAD.
+ * Determine if HEAD points to the given branch
  *
- * @param branch Current underlying reference of the branch.
+ * @param branch A reference to a local branch.
  *
- * @return 1 if HEAD points at the branch, 0 if it isn't,
- * error code otherwise.
+ * @return 1 if HEAD points at the branch, 0 if it isn't, or a negative value
+ * as an error code.
```
pks-t: The second line is missing indentation

include/git2/branch.h
```diff
@@ -169,116 +171,132 @@ GIT_EXTERN(int) git_branch_lookup(
...
- * Return the name of remote that the remote tracking branch belongs to.
+ * Find the remote name of a remote-tracking branch
  *
- * @param out Pointer to the user-allocated git_buf which will be filled with the name of the remote.
+ * This will return the name of the first remote with a matching
+ * remote-tracking branch.
+ * In other words, given the "ref/remotes/test/master" reference,
+ * it'll extract "test" and check all configured branches for one
+ * with a matching refspec, or an error if it's ambiguous and more
+ * than one remote matched.
```
pks-t: I think this still isn't quite right. Proposal:

"This will return the name of the remote whose fetch refspec is matching the given branch. E.g. given a branch "refs/remotes/test/master", it will extract the "test" part. If refspecs from multiple remotes match, the function will return GIT_EAMBGIUOUS."

include/git2/branch.h
tiennou: Do we ? This isn't really enforced in any way, and actually isn't [really](https://libgit2.org/libgit2/#HEAD/group/blame/git_blame_options_init) [consistent](https://libgit2.org/libgit2/#HEAD/group/branch/git_branch_upstream_remote) — I realize it's likely because a lot of those discrepancies are my work (ha! 🤣). My vote usually goes for "make it look good", so I try to make "full-fledged sentences" when contextually appropriate…

include/git2/branch.h
```diff
@@ -126,6 +126,11 @@ GIT_EXTERN(void) git_branch_iterator_free(git_branch_iterator *iter);
  * The new branch name will be checked for validity.
  * See `git_tag_create()` for rules about valid names.
  *
+ * Note that if the move succeeds, the "old" reference object
+ * is a dead pointer. You should *not* free it.
```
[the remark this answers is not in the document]
tiennou: I misread some code, so this is actually wrong — I keep forgetting that the refdb doesn't trust anything, so it always reload reference before doing anything. I added a commit to "fix" `git_reference__set_name` and make it more clear that it's dangerous.

tiennou:
Rebased, I've added some tests to make sure it works correctly on the one case I've seen failing. Should be green, as hopefully the one test I've added has been working sanely here 😜.

include/git2/branch.h
pks-t: Well, I was only referring to consistency in this single doc comment as all the other params were starting with a lower-case letter :)

src/refs.c
```diff
@@ -91,18 +91,23 @@ git_reference *git_reference__alloc(
 	return ref;
 }
 
-git_reference *git_reference__set_name(
-	git_reference *ref, const char *name)
+git_reference *git_reference__realloc(
+	git_reference **ptr_to_ref, const char *name)
```
pks-t: Wouldn't it make more sense to have it return an error code and instead let it update the pointer in-place? Feel free to leave it as-is though.

src/refs.c
```diff
@@ -91,18 +91,23 @@ git_reference *git_reference__alloc(
...
+
+	namelen = strlen(name);
+
 	if (!GIT_ADD_SIZET_OVERFLOW(&reflen, sizeof(git_reference), namelen) &&
 		!GIT_ADD_SIZET_OVERFLOW(&reflen, reflen, 1) &&
-		(rewrite = git__realloc(ref, reflen)) != NULL)
+		(rewrite = git__realloc(*ptr_to_ref, reflen)) != NULL)
 		memcpy(rewrite->name, name, namelen + 1);
 
+	ptr_to_ref = NULL;
```
pks-t: `*ptr_to_ref = NULL;`. That's why tests are failing right now

src/refs.c
tiennou: Not sure. I wanted to make the call "usable, but in a safe way". If it's implementation-defined whether realloc will realloc to a smaller size, and I need to know if the old ref is now invalid, having the pointer be NULL on return signals that case, and the higher-level code can adapt.

Arguably, most of that code comes in preparation of some more work 😉, but I want to keep it public API/documentation specific there. For example, I think having a "higher" low watermark on refname lengths could help with the invalidation (something like 32 could skip quite a few reallocs).

src/refs.c
tiennou: Heh 😉. Fixed, and rebased.

src/refs.c
pks-t: Yeah, just leave it as-is then. Your change is an improvement compared to the current code and doesn't introduce anything new that one could complain about, so it can only be a win for us.

pks-t:
Thanks a lot, @tiennou!

PR #5302, opened 2019-11-09
fileops: correct error return on p_lstat failures when mkdir

tiennou:
IIRC I got a strange return once from lstat, which translated in a weird
error class/message being reported. As a safety measure, enforce a -1 return in
that case.


ethomson:
LGTM.  I don't think we would ever want to stop here and not report an error back to the caller.  Either we ignore it (like in the `ENOENT` case) or we error.  The current behavior definitely seems like a bug.

PR #5303, opened 2019-11-10
patch_parse: use paths from "---"/"+++" lines for binary patches

pks-t:
For some patches, it is not possible to derive the old and new file
paths from the patch header's first line, most importantly when they
contain spaces. In such a case, we derive both paths from the "---" and
"+++" lines, which allow for non-ambiguous parsing. We fail to use these
paths when parsing binary patches without data, though, as we always
expect the header paths to be filled in.

Fix this by using the "---"/"+++" paths by default and only fall back to
header paths if they aren't set. If neither of those paths are set, we
just return an error. Add two tests to verify this behaviour, one of
which would have previously caused a segfault.

---

Sigh. Still not quite fixed, the fuzzer managed to find another variation that causes a segfault. Hopefully this here is the last one.

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -878,12 +878,18 @@ static int parse_patch_binary_nodata(
 	git_patch_parsed *patch,
 	git_patch_parse_ctx *ctx)
 {
+	const char *old = patch->old_path ? patch->old_path : patch->header_old_path;
+	const char *new = patch->new_path ? patch->new_path : patch->header_new_path;
+
+	if (!old || !new)
+		return git_parse_err("corrupt binary data without paths at line %"PRIuZ, ctx->parse_ctx.line_num);
+
 	if (git_parse_advance_expected_str(&ctx->parse_ctx, "Binary files ") < 0 ||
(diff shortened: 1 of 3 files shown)


ethomson:
This looks good to me, so I'm going to go ahead and merge it.  If:

> Sigh. Still not quite fixed, the fuzzer managed to find another variation that causes a segfault. Hopefully this here is the last one.

indicates that you weren't quite done then let's just follow up with a second PR.

pks-t:
On Sat, Nov 16, 2019 at 03:22:02PM -0800, Edward Thomson wrote:
> This looks good to me, so I'm going to go ahead and merge it.
> If:
> 
> > Sigh. Still not quite fixed, the fuzzer managed to find
> > another variation that causes a segfault. Hopefully this here
> > is the last one.
> 
> indicates that you weren't quite done then let's just follow up
> with a second PR.

Uh, no, sorry for not making myself clear. I've already touched
this exact function two times before but didn't manage to catch
all edge-cases in other PRs. Anyway, fingers crossed I got it
right this time :/

PR #5305, opened 2019-11-16
Adds support for multiple SSH auth mechanisms being used sequentially

kas-luthor:
Needed to make eg. 2FA work

Fixes the following problem:
- An SSH server has multiple authentication mechanisms configured (`pubkey` and `keyboard-interactive`, both are required).
1. When connecting to the server it responds sends only `pubkey` as an available mechanism
2. libgit correctly invokes `cred_acquire_cb` with `GIT_CREDTYPE_SSH_KEY` set and `GIT_CREDTYPE_SSH_INTERACTIVE` unset.
3. The `pubkey` authentication is completed successfully, but the server responds with a failure code since not all mechanisms have successfully completed.
4. The connection's available mechanisms are updated to only include `keyboard-interactive` on the server side.
5. libgit incorrectly invokes `cred_acquire_cb` with `GIT_CREDTYPE_SSH_KEY` set and `GIT_CREDTYPE_SSH_INTERACTIVE` unset.
- This repeats until the process is terminated by the server or `cred_acquire_cb`

### DIFF
--- a/src/transports/ssh.c
+++ b/src/transports/ssh.c
@@ -518,6 +518,7 @@ static int _git_ssh_setup_conn(
 {
 	git_net_url urldata = GIT_NET_URL_INIT;
 	int auth_methods, error = 0;
+	int sub_error;
 	size_t i;
 	ssh_stream *s;
 	git_cred *cred = NULL;
@@ -638,6 +639,15 @@ post_extract:
 		}
 
 		error = _git_ssh_authenticate_session(session, cred);
+
+		if (error == GIT_EAUTH) {
+			/* refresh auth methods */
+			sub_error = list_auth_methods(&auth_methods, session, urldata.username);
+			if (sub_error < 0) {
+				error = sub_error;
+				goto done;
+			}
+		}
 	}
 
 	if (error < 0)
--- a/src/transports/ssh.c
+++ b/src/transports/ssh.c
@@ -518,7 +518,6 @@ static int _git_ssh_setup_conn(
 {
 	git_net_url urldata = GIT_NET_URL_INIT;
 	int auth_methods, error = 0;
-	int sub_error;
 	size_t i;
 	ssh_stream *s;
 	git_cred *cred = NULL;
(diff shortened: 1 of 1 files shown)


pks-t:
Thanks for your PR! I'm not that knowledgeable around our SSH backend, but maybe @tiennou can chime in here?

src/transports/ssh.c
```diff
@@ -638,6 +639,15 @@ static int _git_ssh_setup_conn(
 		}
 
 		error = _git_ssh_authenticate_session(session, cred);
+
+		if (error == GIT_EAUTH) {
```
tiennou: You can clobber `error` here, so no need to declare `sub_error`.

```c
/* refresh auth methods */
if (error == GIT_EAUTH &&
    error = list_auth_methods(…))
        goto done;
```

kas-luthor: I disagree, with that change `error` will no longer be `GIT_EAUTH` after the call to `list_auth_methods()` and cancel the loop.

tiennou: Actually… there's nothing in here that actually prevents a server from swamping you into a loop by repeatedly switching auth methods and continuously failing to authenticate, so maybe that's something that needs to happen… Not really related, but you've just mentioned a loop, and I happened to completely miss its condition 😉.

So; I'm okay with your initial version then. If you happen to add a auth counter or something, that would be lovely. I'm not sure how it would *really work* though…

pks-t: > So; I'm okay with your initial version then. If you happen to add a auth counter or something, that would be lovely. I'm not sure how it would really work though…

Does the server even need to switch auth methods? He could just refuse the authentication repeatedly, which would cause us to loop infinitely here until one manages to send the correct credentials. Or am I missing something?

kas-luthor: @pks-t Authentication can be cancelled by the callback by returning an error code

pks-t:
Thanks a lot for your work, @kas-luthor.

> So; I'm okay with your initial version then. If you happen to add a auth counter or something, that would be lovely. I'm not sure how it would really work though…

I guess we can just wait for somebody to complain and add it later, if required at all.

PR #5306, opened 2019-11-19
diff: complete support for git patchid

herrerog:
Hi,
I noticed that git_diff_patchid wasn't returning a correct hash in some cases.
So I fixed those and added some tests for it.
Please let me know what you think.
Thanks!

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -426,41 +426,7 @@ static void strip_spaces(git_buf *buf)
 	git_buf_truncate(buf, len);
 }
 
-static int file_cb(
-	const git_diff_delta *delta,
-	float progress,
-	void *payload)
-{
-	struct patch_id_args *args = (struct patch_id_args *) payload;
-	git_buf buf = GIT_BUF_INIT;
-	int error;
-
-	GIT_UNUSED(progress);
-
-	if (!args->first_file &&
-	    (error = flush_hunk(&args->result, &args->ctx)) < 0)
-		goto out;
-	args->first_file = 0;
-
-	if ((error = git_buf_printf(&buf,
-				    "diff--gita/%sb/%s---a/%s+++b/%s",
-				    delta->old_file.path,
-				    delta->new_file.path,
-				    delta->old_file.path,
-				    delta->new_file.path)) < 0)
-		goto out;
-
-	strip_spaces(&buf);
-
-	if ((error = git_hash_update(&args->ctx, buf.ptr, buf.size)) < 0)
-		goto out;
-
-out:
-	git_buf_dispose(&buf);
-	return error;
-}
-
-static int patchid_line_cb(
+int git_diff_patchid_print_callback__to_buf(
 	const git_diff_delta *delta,
 	const git_diff_hunk *hunk,
 	const git_diff_line *line,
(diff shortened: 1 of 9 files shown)


ethomson:
Overall, this looks very nice to me.  Thanks for the PR!

/cc @pks-t to look at some of the factoring, especially since patchid was his work originally and he has more familiarity.

Note that the build failures appear to be due to #5308.

src/diff.c
```diff
@@ -426,81 +426,39 @@ static void strip_spaces(git_buf *buf)
...
-		return -1;
-	}
-
-	git_buf_put(&buf, line->content, line->content_len);
 	strip_spaces(&buf);
 
-	if ((error = git_hash_update(&args->ctx, buf.ptr, buf.size)) < 0)
+	if (line->origin == GIT_DIFF_LINE_FILE_HDR && !args->first_file)
+		if ((error = (flush_hunk(&args->result, &args->ctx))) < 0)
+			goto out;
```
pks-t: Stylistic nit: the nested if isn't required here and we usually use `&&` chaining in these cases. It's a matter of taste though, so feel free to ignore this comment.
Please remove the braces around `flush_hunk`, though.

src/diff.c
```diff
@@ -526,7 +484,10 @@ int git_diff_patchid(git_oid *out, git_diff *diff, git_diff_patchid_options *opt
 	if ((error = git_hash_ctx_init(&args.ctx)) < 0)
 		goto out;
 
-	if ((error = git_diff_foreach(diff, file_cb, NULL, NULL, patchid_line_cb, &args)) < 0)
+	if ((error = (git_diff_print(diff,
+				     GIT_DIFF_FORMAT_PATCH_ID,
+				     git_diff_patchid_print_callback__to_buf,
+				     &args))) < 0)
```
pks-t: I like your approach, it's a lot simpler than what I was doing before.

src/diff.c
```diff
@@ -426,81 +426,39 @@ static void strip_spaces(git_buf *buf)
...
-
-	git_buf_put(&buf, line->content, line->content_len);
 	strip_spaces(&buf);
 
-	if ((error = git_hash_update(&args->ctx, buf.ptr, buf.size)) < 0)
+	if (line->origin == GIT_DIFF_LINE_FILE_HDR && !args->first_file)
+		if ((error = (flush_hunk(&args->result, &args->ctx))) < 0)
+			goto out;
+
+	if ((error = (git_hash_update(&args->ctx, buf.ptr, buf.size))) < 0)
```
pks-t: Well, at least you're consistent :P Please follow libgit2's style and consistently remove braces around function invocations.

src/diff_print.c
```diff
@@ -571,7 +571,7 @@ static int diff_print_patch_file(
 		(pi->flags & GIT_DIFF_FORCE_BINARY);
 	bool show_binary = !!(pi->flags & GIT_DIFF_SHOW_BINARY);
 	int id_strlen = pi->id_strlen;
-	bool print_index = true;
+	bool print_index = (pi->format != GIT_DIFF_FORMAT_PATCH_ID);
```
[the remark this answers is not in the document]
herrerog: Thanks!

src/diff.c
```diff
@@ -426,81 +426,39 @@ static void strip_spaces(git_buf *buf)
...
+	const git_diff_line *line,
 	void *payload)
 {
 	struct patch_id_args *args = (struct patch_id_args *) payload;
 	git_buf buf = GIT_BUF_INIT;
-	int error;
+	int error = 0;
 
-	GIT_UNUSED(progress);
+	git_buf_sanitize(&buf);
```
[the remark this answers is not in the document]
herrerog: Right, I fixed that.

src/diff.c
herrerog: Ah good catch for the `&&`. This looks better.
I will remove all braces around functions calls.

src/diff.c
herrerog: Lol yeah, not sure why I was thinking this was the way to go for libgit2.

src/diff.c
herrerog: Thanks

herrerog:
Thanks for your nice feedback, happy to help :)
I updated the branch.

pks-t:
Thanks a lot for this well-crafted pull request, @herrerog!

PR #5307, opened 2019-11-20
ssh: include sha256 host key hash when supported

palmin:
libssh2 1.9.0 supports SHA-256 host key fingerprints and by checking if LIBSSH2_HOSTKEY_HASH_SHA256 is defined this can be included in git_cert_hostkey when libssh2 supports this without breaking builds when older versions of libssh2 does not.

### DIFF
--- a/include/git2/cert.h
+++ b/include/git2/cert.h
@@ -78,6 +78,8 @@ typedef enum {
 	GIT_CERT_SSH_MD5 = (1 << 0),
 	/** SHA-1 is available */
 	GIT_CERT_SSH_SHA1 = (1 << 1),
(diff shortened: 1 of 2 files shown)


pks-t:
> but there's a plan to provide a replacement backend

Or at least a parallel SSH backend. Not sure if we really want to ditch libssh2 in the near-term future, even if we have libssh as a new (and probably much better) alternative.

I'm :+1: on this PR, as I imagine it to be useful even across different SSH backends. So as @tiennou approved this either and didn't mention any specific incompatibilities with the upcoming libssh backend I'm going to merge this.

Thanks a lot for your nice change!

PR #5308, opened 2019-11-23
CI Build Updates

ethomson:
Our CI builds are currently flaky.  Some of these issues are related to the upgrade to xenial for our Linux build and test runs.  Some of them snuck in because that move destabilized our CI and we didn't catch problems.

Here are some issues:

1. valgrind is reporting errors that it should be suppressing.  This is because our valgrind suppression is suppressing a number of errors in symbols in `libssh2.so`, but we're currently statically linking against `libssh2`.  Thus the suppression file can't match those symbols and we see a number of valgrind errors.
2. valgrind errors _aren't failing the builds_.  The version of valgrind that we were using from the PPA is broken and `--exit-errorcode` is ignored.  😳
3. OpenSSL has several new issues that need to be suppressed.  It _appears_ that this is an interaction between the versions of OpenSSL, gcc and valgrind that we have on the build/test image reporting a false positive for uninitialized memory.  As best I can tell, valgrind misses some subtle memory initializations with some gcc optimizations.  Amusingly, these are only noticed when talking to Azure Repos.  Presumably the Windows TLS stack or the ciphers in use are sufficiently different for valgrind to notice these.  (But only _sometimes_, sigh.)

Fixes:

1. Build a shared library during the libssh2 compilation step so that we can target suppressions against symbols in `libssh2.so`.
2. Build the latest version of valgrind instead of using the PPA, so that `--exit-errorcode` works again.
3. Quiet valgrind down.
   i.  Extend the suppressions for valgrind to include uninitialized usage stemming from OpenSSL.
   ii. Mark memory returned from OpenSSL as initialized explicitly.

In addition, there are a few build process enhancements:

1. Provide a default CMake generator in the build script.  It explicitly passes `-G`, so it needs a default to supply if none was given in the environment.
2. Show the distribution information again.
3. Break the dockerfile up into stages.  This allows it to cache stages and re-use unchanged ones, to prevent (eg) long apt installs that don't actually change the dependencies.  Now `docker build` will pick up again from the first modified stage, re-using the unchanged stages from the cache.
4. Don't delete the apt cache.  This aides in local debugging, so that you can `apt-get install ...` tools in the container.  This bloats the image somewhat, but since we don't store it in a registry, this is of limited pain.
5. Cache docker layers in Azure Pipelines.  This prevents us from needing to rebuild the entire image on every build.


ethomson:
It gets better: the version of valgrind on the Ubuntu Xenial images has a broken `--error-exitcode`.

```
root@40e03a14632b:/tmp# valgrind -q --leak-check=full --show-reachable=yes --error-exitcode=2 --suppressions=/src/script/valgrind.supp  ./foo
==7019== 1,024 bytes in 1 blocks are definitely lost in loss record 1 of 1
==7019==    at 0x4C2DB4F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==7019==    by 0x40059E: main (in /tmp/foo)
==7019==
root@40e03a14632b:/tmp# echo $?
0
```

Sigh.

ethomson:
Merging this so that CI is reliable again.

azure-pipelines/docker.yml
```diff
@@ -4,9 +4,19 @@ steps:
...
     displayName: 'Register Docker QEMU'
 
+- task: cache@2
+  displayName: Cache Docker layers
+  inputs:
+    key: docker
+    path: /tmp/dockercache
+- script: |
+    if [ -f /tmp/dockercache/${{parameters.docker.image}}.tar ]; then docker load < /tmp/dockercache/${{parameters.docker.image}}.tar; fi
+  displayName: 'Load Docker cache'
```
pks-t: :tada: :tada: :tada: Great change

PR #5310, opened 2019-11-28
Add compat typdef for git_attr_t

lberk:
Some libraries haven't updated to git_attr_value_t and break.  Adding
the comapt typedef as suggested.

@pks-t please let me know if there's anything else needed.  I took a brief look but didn't see an obvious place where to add tests.  If required please let me know.

### DIFF
--- a/include/git2/deprecated.h
+++ b/include/git2/deprecated.h
@@ -68,6 +68,21 @@ GIT_BEGIN_DECL
 
 /**@}*/
 
+/** @name Deprecated Attribute type
+ *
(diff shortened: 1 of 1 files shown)


ethomson:
Oops.  Makes sense.  Thanks for the fix.

Instead of starting a new attribute section, can you put this in the attribute section right above?

lberk:
@ethomson done!

ethomson:
Perfect.  Thanks!  And apologies for the oversight.

PR #5312, opened 2019-11-28
patch_parse: fix out-of-bounds reads caused by integer underflow

pks-t:
The patch format for binary files is a simple Base85 encoding with a
length byte as prefix that encodes the current line's length. For each
line, we thus check whether the line's actual length matches its
expected length in order to not faultily apply a truncated patch. This
also acts as a check to verify that we're not reading outside of the
line's string:

	if (encoded_len > ctx->parse_ctx.line_len - 1) {
		error = git_parse_err(...);
		goto done;
	}

There is the possibility for an integer underflow, though. Given a line
with a single prefix byte, only, `line_len` will be zero when reaching
this check. As a result, subtracting one from that will result in an
integer underflow, causing us to assume that there's a wealth of bytes
available later on. Naturally, this may result in an out-of-bounds read.

Fix the issue by checking both `encoded_len` and `line_len` for a
non-zero value. The binary format doesn't make use of zero-length lines
anyway, so we need to know that there are both encoded bytes and
remaining characters available at all.

This patch also adds a test that works based on the last error message.
Checking error messages is usually too tightly coupled, but in fact
parsing the patch failed even before the change. Thus the only
possibility is to use e.g. Valgrind, but that'd result in us not
catching issues when run without Valgrind. As a result, using the error
message is considered a viable tradeoff as we know that we didn't start
decoding Base85 in the first place.

### DIFF
--- a/src/patch_parse.c
+++ b/src/patch_parse.c
@@ -808,7 +808,7 @@ static int parse_patch_binary_side(
 
 		encoded_len = ((decoded_len / 4) + !!(decoded_len % 4)) * 5;
 
-		if (encoded_len > ctx->parse_ctx.line_len - 1) {
+		if (!encoded_len || !ctx->parse_ctx.line_len || encoded_len > ctx->parse_ctx.line_len - 1) {
 			error = git_parse_err("truncated binary data at line %"PRIuZ, ctx->parse_ctx.line_num);
 			goto done;
 		}
--- a/tests/patch/parse.c
+++ b/tests/patch/parse.c
@@ -184,6 +184,14 @@ void test_patch_parse__binary_file_path_without_body_paths(void)
 					  strlen(PATCH_BINARY_FILE_PATH_WITHOUT_BODY_PATHS), NULL));
 }
 
+void test_patch_parse__binary_file_with_truncated_delta(void)
+{
+	git_patch *patch;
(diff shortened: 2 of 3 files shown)


src/patch_parse.c
```diff
@@ -808,7 +808,7 @@ static int parse_patch_binary_side(
 
 		encoded_len = ((decoded_len / 4) + !!(decoded_len % 4)) * 5;
 
-		if (encoded_len > ctx->parse_ctx.line_len - 1) {
```
tiennou: I'd extract that as `static int parse_binary_patch_length(git_patch_parse_ctx *ctx, size_t decoded_len)`, just to make sure the overflow is handled only once, every time 😉. Not sure how generic that is, but maybe that can be made public at the parser layer (this idiom comes from libssh2, which looks at binary data *all the time*) ?

pks-t: Right now there's only a single location where we calculate the binary header length, and putting this into a more generic layer feels a bit premature to me. Binary patches use a special prefix-length-encoding that is currently in use for binary patches only, and I don't expect this to be used anywhere else anytime soon.

So I think we should definitely pull it out into a separate function as soon as there is a second user. But right now I don't quite think it to be worthwhile. Are you good with this?

tiennou: I'm fine with the current implementation, sorry for being overly pedantic 😉.

pks-t: No need to be sorry at all, I'm always happy to get your input!

PR #5314, opened 2019-11-29
global: convert to fiber-local storage to fix exit races

### LINKED ISSUE #5207: Heap is corrupted during shutdown and TLS cleanup when loaded as DLL in a .NET process.
### Reproduction steps
1. Create .NET console application with P/Invoke bindings to the following native LibGit2 library functions (i.e., **_not_** using LibGit2Sharp; the native DLL is loaded by the CLR when the first P/Invoke binding is called):
    - `git_libgit2_init`
    - `git_repository_open`
    - `git_repository_free`
    - `git_libgit2_shutdown` 
1. Create a long running .NET thread (T2) that spins.
1. On the initial thread (T1), initialise LibGit2 calling `git_libgit2_init`.
1. On T1, open a new LG2 repository handle calling `git_repository_open`.
1. In quick succession/simultaneously:
    - Terminate the long running .NET thread T2.
    - From T1, close the LG2 repo calling `git_repository_free`.
    Shutdown the library calling `git_libgit2_shutdown`.

#### Further notes
It appears that the termination of the long running .NET thread (T2) causes `DllMain` to be called by Windows with the `DLL_THREAD_DETATCH` reason, leading to TLS cleanup.
At the same time calling `git_libgit2_shutdown` from T1 invokes the same TLS deallocating functions, possibly exposing a concurrency bug.

### Expected behavior
No app crash, the library is shutdown and TLS is freed.

### Actual behavior
The native LibGit2 library corrupts the heap attempting to do a double-free on some thread-local storage (TLS) pointer. This only occurs sometimes due to timing and ordering of instructions between threads.

```
Unhandled exception at <address> (ntdll.dll) in App.exe: 0xC0000374: A heap has been corrupted (parameters: 0x00007FFAAB5E27F0). occurred

ntdll.dll!RtlReportFatalFailure(_EXCEPTION_RECORD * ExceptionRecord) Line 150C
ntdll.dll!RtlReportCriticalFailure(long StatusCode, void * FailureInfo, unsigned long BreakIfDbgPresent) Line 218C
ntdll.dll!RtlpHeapHandleError(long ErrorLevel) Line 344C
ntdll.dll!RtlpHpHeapHandleError(_HEAP_FAILURE_TYPE FailureType, unsigned __int64 HeapAddress, unsigned __int64 Address) Line 670C
ntdll.dll!RtlpLogHeapFailure(_HEAP_FAILURE_TYPE FailureType, void * HeapAddress, void * Address, void * Param1, void * Param2, void * Param3) Line 158C
ntdll.dll!RtlFreeHeap(void * HeapHandle, unsigned long Flags, void * BaseAddress) Line 352C
git2.dll!_free_base(void * block) Line 107C++
[Inline Frame] git2.dll!git__free(void *) Line 246C
[Inline Frame] git2.dll!git__free_tls_data() Line 226C
git2.dll!DllMain(HINSTANCE__ * hInstDll, uns
```

1.0.165 (acquired from NuGet package LibGit2Sharp.NativeBinaries)

Windows 10 (and .NET Core 2.1)

pks-t:
On Windows platforms, we automatically clean up the thread-local storage
upon detaching a thread via `DllMain()`. The thing is that this happens
for every thread of applications that link against the libgit2 DLL, even
those that don't have anything to do with libgit2 itself. As a result,
we cannot assume that these unsuspecting threads make use of our
`git_libgit2_init()` and `git_libgit2_shutdow()` reference counting,
which may lead to racy situations:

    Thread 1                    Thread 2

    git_libgit2_shutdown()
                                DllMain(DETACH_THREAD)
                                git__free_tls_data()
    git_atomic_dec() == 0
    git__free_tls_data()
    TlsFree(_tls_index)
                                TlsGetValue(_tls_index)

Due to the second thread never having executed `git_libgit2_init()`, the
first thread will clean up TLS data and as a result also free the
`_tls_index` variable. When detaching the second thread, we
unconditionally access the now-free'd `_tls_index` variable, which is
obviously not going to work out well.

Fix the issue by completely removing the `DllMain()` function. Its sole
purpose was to call `git__free_tls_data()` on detaching threads. But as
we require that every thread making use of libgit2 needs to call
`git_libgit2_init()` and `git_libgit2_shutdow()` anyway, we can just
move cleanup of thread-local storage to `git_libgit2_shutdown()`. This
nicely avoids any issues of non-libgit2-related threads trying to clean
up their TLS while still making sure that every libgit2-using thread
will do so. Furthermore, it will avoid free'ing of `_tls_index` while
any thread still uses libgit2 due to the reference-counting semantics of
the global init/shutdown procedures.

---

Fixes #5207, #5154.

### DIFF
--- a/src/global.c
+++ b/src/global.c
@@ -141,14 +141,21 @@ static void shutdown_common(void)
  */
 #if defined(GIT_THREADS) && defined(GIT_WIN32)
 
-static DWORD _tls_index;
+static DWORD _fls_index;
 static volatile LONG _mutex = 0;
+
+static void WINAPI fls_free(void *st)
+{
+	git__global_state_cleanup(st);
+	git__free(st);
+}
 
 static int synchronized_threads_init(void)
 {
 	int error;
 
-	_tls_index = TlsAlloc();
+	if ((_fls_index = FlsAlloc(fls_free)) == FLS_OUT_OF_INDEXES)
+		return -1;
 
 	git_threads_init();
 
@@ -190,9 +197,7 @@ int git_libgit2_shutdown(void)
 	if ((ret = git_atomic_dec(&git__n_inits)) == 0) {
 		shutdown_common();
 
-		git__free_tls_data();
-
-		TlsFree(_tls_index);
+		FlsFree(_fls_index);
 		git_mutex_free(&git__mwindow_mutex);
 
 #if defined(GIT_MSVC_CRTDBG)
@@ -213,7 +218,7 @@ git_global_st *git__global_state(void)
 
 	assert(git_atomic_get(&git__n_inits) > 0);
 
-	if ((ptr = TlsGetValue(_tls_index)) != NULL)
+	if ((ptr = FlsGetValue(_fls_index)) != NULL)
 		return ptr;
 
 	ptr = git__calloc(1, sizeof(git_global_st));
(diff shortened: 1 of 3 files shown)


src/global.c
```diff
@@ -201,6 +198,17 @@ int git_libgit2_shutdown(void)
 #endif
 	}
 
+	/*
+	 * Always clean up thread-specific global data before
+	 * free'ing the TLS index. We obviously need to do this
+	 * before calling `TlsFree()`, but in fact after
+	 * shutting down all subsystems.
+	 */
+	git__free_tls_data();
```
pks-t: Mh, in fact this isn't quite right either, as we'll now always free TLS data when calling `git_libgit2_shutdown()`. But if the same thread has called `git_libgit2_init()` multiple times, then we definitely don't want that. We could fix the issue by adding another reference count to the global state, but I dunno...

pks-t:
The best solution is probably to switch from TLS to FLS, as we can specify an FlsCallback similar to `pthread_key_create`. I'll take a stab at that.

pks-t:
Hum, the results look a lot better than expected. Windows' threading code got a lot cleaner due to the conversion and should now be race-free in the above described failure scenario.

Yay for FLS.

ethomson:
Look at you with your fancy Windows fibers.  I didn't know that this was a thing that you could do.

ethomson:
There are lots of weird idiosyncrasies to using fibers.  But we're not really _using fibers_, just _fiber-local storage_, and Raymond Chen has basically [blessed this pattern](https://devblogs.microsoft.com/oldnewthing/20191011-00/?p=102989), so I'm :+1: on this.

Clever solution.  Thanks for digging in to this.

PR #5315, opened 2019-11-29
Fix copy&paste in git_cherrypick_commit docstring

kastiglione:
Updates the docstring of `git_cherrypick_commit` to fix some incorrect mentions of reverting


ethomson:
Oops!  Good catch.  Thanks @kastiglione

PR #5317, opened 2019-12-03
MSVC: Fix warning C4133 on x64: "function": Incompatible types - from "unsigned long *" to "size_t *"

csware:


src/regexp.c
```diff
@@ -78,7 +78,7 @@ int git_regexp_search(const git_regexp *r, const char *string, size_t nmatches,
 int git_regexp_compile(git_regexp *r, const char *pattern, int flags)
 {
 	unsigned char errmsg[1024];
-	unsigned long erroff;
+	size_t erroff;
```
ethomson: Good catch.  This should maybe be `PCRE2_SIZE` per the PCRE2 docs themselves.  Will that always be `size_t`?

csware: It's hard-coded in pcre2.h, but I think `PCRE2_SIZE` is better...

ethomson:
Awesome.  Thanks for the fix!

PR #5319, opened 2019-12-04
worktree: mimic 'git worktree add' behavior.

herrerog:
When adding a worktree using `git worktree add <path>`, if a reference
named `refs/heads/$(basename <path>)` already exist, it is checkouted in
the worktree.
Mimic this behavior in libgit2.

### DIFF
--- a/src/worktree.c
+++ b/src/worktree.c
@@ -278,7 +278,8 @@ int git_worktree_add(git_worktree **out, git_repository *repo,
 	const char *name, const char *worktree,
 	const git_worktree_add_options *opts)
 {
-	git_buf gitdir = GIT_BUF_INIT, wddir = GIT_BUF_INIT, buf = GIT_BUF_INIT;
+	git_buf gitdir = GIT_BUF_INIT, wddir = GIT_BUF_INIT;
+	git_buf buf = GIT_BUF_INIT, ref_buf = GIT_BUF_INIT;
 	git_reference *ref = NULL, *head = NULL;
 	git_commit *commit = NULL;
 	git_repository *wt = NULL;
@@ -365,12 +366,24 @@ int git_worktree_add(git_worktree **out, git_repository *repo,
 		if ((err = git_reference_dup(&ref, wtopts.ref)) < 0)
 			goto out;
 	} else {
-		if ((err = git_repository_head(&head, repo)) < 0)
+		if ((err = git_buf_printf(&ref_buf, "refs/heads/%s", name)) < 0)
 			goto out;
-		if ((err = git_commit_lookup(&commit, repo, &head->target.oid)) < 0)
-			goto out;
-		if ((err = git_branch_create(&ref, repo, name, commit, false)) < 0)
-			goto out;
+		if (!git_reference_lookup(&ref, repo, ref_buf.ptr)) {
+			if (git_branch_is_checked_out(ref)) {
+				git_error_set(GIT_ERROR_WORKTREE,
+					      "reference %s is already checked out",
+					      ref_buf.ptr);
+				err = -1;
+				goto out;
+			}
+		} else {
+			if ((err = git_repository_head(&head, repo)) < 0)
+				goto out;
+			if ((err = git_commit_lookup(&commit, repo, &head->target.oid)) < 0)
+				goto out;
+			if ((err = git_branch_create(&ref, repo, name, commit, false)) < 0)
+				goto out;
+		}
 	}
 
 	/* Set worktree's HEAD */
@@ -392,6 +405,7 @@ out:
 	git_buf_dispose(&gitdir);
 	git_buf_dispose(&wddir);
 	git_buf_dispose(&buf);
+	git_buf_dispose(&ref_buf);
 	git_reference_free(ref);
 	git_reference_free(head);
 	git_commit_free(commit);
--- a/src/worktree.c
+++ b/src/worktree.c
@@ -278,8 +278,7 @@ int git_worktree_add(git_worktree **out, git_repository *repo,
 	const char *name, const char *worktree,
 	const git_worktree_add_options *opts)
 {
-	git_buf gitdir = GIT_BUF_INIT, wddir = GIT_BUF_INIT;
-	git_buf buf = GIT_BUF_INIT, ref_buf = GIT_BUF_INIT;
+	git_buf gitdir = GIT_BUF_INIT, wddir = GIT_BUF_INIT, buf = GIT_BUF_INIT;
 	git_reference *ref = NULL, *head = NULL;
 	git_commit *commit = NULL;
 	git_repository *wt = NULL;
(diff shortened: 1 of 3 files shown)


src/worktree.c
```diff
@@ -365,12 +365,25 @@ int git_worktree_add(git_worktree **out, git_repository *repo,
 		if ((err = git_reference_dup(&ref, wtopts.ref)) < 0)
 			goto out;
 	} else {
-		if ((err = git_repository_head(&head, repo)) < 0)
-			goto out;
-		if ((err = git_commit_lookup(&commit, repo, &head->target.oid)) < 0)
-			goto out;
-		if ((err = git_branch_create(&ref, repo, name, commit, false)) < 0)
+		git_buf ref_buf = GIT_BUF_INIT;
```
tiennou: This must be `git_buf_dispose`d of — CI reports a leak.

herrerog: > Hmm, isn't this supposed to be passed as `git_worktree_add_options`'s [ref field](https://libgit2.org/libgit2/#HEAD/type/git_worktree_add_options) ? As far as I can see this changes the behavior, as the old code would have _never_ reused a reference, instead returning an error. Now you can _never_ now if this actually happened, and you can end-up reusing by accident another unrelated reference.

`ref` field looks a bit different to me, it's when you want to checkout `ref` in the worktree.
libgit2 will checkout this `ref` only if it already exists and is not already checked out.

When user doesn't pass `ref` field, current version of `git_worktree_add` creates a new branch with the `name` of the worktree in the main working tree and checkout HEAD in the new worktree.

The only thing this commit is changing is when `ref` field is not passed as an option and branch `refs/heads/<name>` already exists in the main working tree.
It would have error out previously. But with this commit, it will now checkout the commit pointed by `refs/heads/<name>` in the new worktree without any error.

It's the same behavior when you do the following:
`git worktree add ../libgit2-wt`
Git create the new worktree in ../libgit2-wt and create a branch `libgit2-wt` in the main repositoriy.
`rm -Rf ../libgit2-wt`
`git worktree prune`
Here we deleted the `libgit2-wt` and pruned it but Git kept the `libgit2-wt` branch in the main repository.

`git worktree add ../libgit2-wt`
Adding the worktree again works whereas `libgit2-wt` already exists.

This currently doesn't work with libgit2 and that's what I'm trying to fix.

Please let me know if I'm missing something.
Thanks for the review.

herrerog:
> Hmm, isn't this supposed to be passed as `git_worktree_add_options`'s [ref field](https://libgit2.org/libgit2/#HEAD/type/git_worktree_add_options) ? As far as I can see this changes the behavior, as the old code would have _never_ reused a reference, instead returning an error. Now you can _never_ now if this actually happened, and you can end-up reusing by accident another unrelated reference.

`ref` field looks a bit different to me, it's when you want to checkout `ref` in the worktree.
libgit2 will checkout this `ref` only if it already exists and is not already checked out.

When user doesn't pass `ref` field, current version of `git_worktree_add` creates a new branch with the `name` of the worktree in the main working tree and checkout HEAD in the new worktree.

The only thing this commit is changing is when `ref` field is not passed as an option and branch `refs/heads/<name>` already exists in the main working tree.
It would have error out previously. But with this commit, it will now checkout the commit pointed by `refs/heads/<name>` in the new worktree without any error.

It's the same behavior when you do the following:
`git worktree add ../libgit2-wt`
Git create the new worktree in ../libgit2-wt and create a branch `libgit2-wt` in the main repositoriy.
`rm -Rf ../libgit2-wt`
`git worktree prune`
Here we deleted the `libgit2-wt` and pruned it but Git kept the `libgit2-wt` branch in the main repository.

`git worktree add ../libgit2-wt`
Adding the worktree again works whereas `libgit2-wt` already exists.

This currently doesn't work with libgit2 and that's what I'm trying to fix.

Please let me know if I'm missing something.
Thanks for the review.

src/worktree.c
herrerog: I fixed that, thanks.

herrerog:
I have added a test `test_worktree_worktree__add_remove_add` which add a worktree, delete it and add it back. This may help understanding what I'm trying to achieve.

src/worktree.c
```diff
@@ -365,12 +366,24 @@ int git_worktree_add(git_worktree **out, git_repository *repo,
...
+			}
+		} else {
+			if ((err = git_repository_head(&head, repo)) < 0)
+				goto out;
+			if ((err = git_commit_lookup(&commit, repo, &head->target.oid)) < 0)
+				goto out;
+			if ((err = git_branch_create(&ref, repo, name, commit, false)) < 0)
+				goto out;
+		}
 	}
```
pks-t: Hm. I feel like you've got a point with the `git_branch_is_checked_out` call, we're not doing that e.g. if `wtopts.ref` is set. Let me propose the following flow:

```
if (wtopts.ref) {
    git_reference_dup(&ref, wtopts.ref);
} else if (git_branch_lookup(&ref, name) == 0) {
    /* Use the ref */
} else {
    git_branche_create(&ref);
}

if (git_branch_is_checked_out(ref)) {
    error out;
}
```

I think we ought to give the user a option to turn off re-use of branches, though. So maybe we should add a flag to `git_worktree_add_options`?

herrerog: I think we are already checking if `wopts.refs` is checkouted at the beginning of the `git_worktree_add` function with below code:
```
if (wtopts.ref) {
		if (!git_reference_is_branch(wtopts.ref)) {
			git_error_set(GIT_ERROR_WORKTREE, "reference is not a branch");
			err = -1;
			goto out;
		}

		if (git_branch_is_checked_out(wtopts.ref)) {
			git_error_set(GIT_ERROR_WORKTREE, "reference is already checked out");
			err = -1;
			goto out;
		}
}
```
So that should be fine.
Should we add a `force` option as in `git worktree` command ?

pks-t: Ah, fair enough. Didn't read into the code and only saw this single instance of `if (wtopts.ref)` in the diff. I'd prefer to have that check whether it's checked out in a single location as proposed by me to make it easier to extend, bu that's definitely not something you'll have to fix.

I'd like to not call this `force`, as this is a frequently overloaded name and doesn't really tell what's being forced. Something like "allow_reusing_refs" or "allow_existing_branch" would be better? I'm open for better proposals though!

herrerog: After thinking a bit more about it, the `force` flag would make sense only if the branch passed as `wtopts.ref` or inferred from `name` of the worktree is already checked out.
This pull request currently aborts if the branch is already checked out.
If the branch is not checked out, it will simply reuse it. So I guess that's fine from an user point of view.
Let me know what you think.

pks-t: > This pull request currently aborts if the branch is already checked out.

Let's keep it this way for now. Can still add it later if requested.

pks-t (dismissed):
Sorry for taking so long to review this and thanks for your changes!

pks-t:
Rebased branch to re-trigger CI.

pks-t:
I've added another commit that hides this new feature behind an option to address @tiennou's remark.

herrerog:
> I've added another commit that hides this new feature behind an option to address @tiennou's remark.

Thanks, this looks good to me.

ethomson:
I manually merged this and fixed up the conflicts. Thanks!

PR #5320, opened 2019-12-04
Minor doc improvements

josharian:


docs/changelog.md
```diff
@@ -27,8 +27,10 @@ v0.28 + 1
 
 ### Changes or improvements
 
-* libgit2 can now correctly cope with URLs where the host contains a colon
-  but a port is not specified.  (eg `http://example.com:/repo.git`).
+* libgit2 now correctly handles more URLs, such as
+  `http://example.com:/repo.git` (colon but no port),
+  `http://example.com` (no path),
+  and `http://example.com:8080/` (path is /, nonstandard port).
```
pks-t: Didn't we handle both URLs with no path and / path with nonstandard port just fine before? I'm not too sure about this, but I thought the only change we introduced was that we are now able to parse empty ports with that release.

include/git2/stash.h
```diff
@@ -126,7 +126,7 @@ typedef int GIT_CALLBACK(git_stash_apply_progress_cb)(
 typedef struct git_stash_apply_options {
 	unsigned int version;
 
-	/** See `git_stash_apply_flags_t`, above. */
+	/** See `git_stash_apply_flags`, above. */
```
pks-t: Indeed :+1:

docs/changelog.md
josharian: Parsing empty ports may have been the motivating case, but neither of those other cases worked for me. Please see the commit message, where I documented what the previous behavior was. Should I provide reproduction instructions as well?

For the record, I bisected the change in behavior to c6ab183. It turns out they still aren't actually fully working yet: #5321. But I hope that that will get fixed before the next release.

pks-t: Ah, perfect! I didn't expect nice commit messages, but here they are. So no, this is perfectly fine in that case and makes sense to me.

PR #5326, opened 2019-12-07
Fixes error message when pushing an already existing annotated tag

kas-luthor:


ethomson:
Hi @kas-luthor, thanks for the pull request!  Unfortunately, I don't have a lot of context here.  Can you provide some more details about what this change does and why?

Including the following would be great:

Reproduction steps

Expected behavior

Actual behavior

Ultimately, we'd also like to have a test that exhibits this behavior so that we can ensure that we don't regress.  Once we understand why this change is being made, we can work on ensuring that we can test it.  Thanks!

kas-luthor:
Nevermind, I'm unable to reproduce the error after a clean build. I guess something went wrong while switching branches?

Reproduction steps:
1. Initialize two new repositories
2. Set one as the origin of the other
3. Create a commit and push it
4. Create an annotated tag and push it
5. (Force-)Create a new annotated tag with the same name and push it

Expected behavior: pushing fails with `GIT_ENONFASTFORWARD`
Actual behavior: pushing fails with `GIT_ERROR_INVALID` and message `object is no commit object`

PR #5327, opened 2019-12-09
Introduce GIT_ASSERT macros

ethomson:
Provide macros to replace usages of `assert`.  A true `assert` is punishing as a library.  Instead we should do our best to not crash.

`GIT_ASSERT_ARG(x)` will now assert that the given argument complies to some format and sets an error message and returns `-1` if it does not.

`GIT_ASSERT(x)` is for internal usage, and available as an internal consistency check.  It will set an error message and return `-1` in the event of failure.

### DIFF
--- a/src/diff_generate.c
+++ b/src/diff_generate.c
@@ -1206,8 +1206,9 @@ int git_diff__from_iterators(
 
 	/* make iterators have matching icase behavior */
 	if (DIFF_FLAG_IS_SET(diff, GIT_DIFF_IGNORE_CASE)) {
-		git_iterator_set_ignore_case(old_iter, true);
-		git_iterator_set_ignore_case(new_iter, true);
+		if ((error = git_iterator_set_ignore_case(old_iter, true)) < 0 ||
+		    (error = git_iterator_set_ignore_case(new_iter, true)) < 0)
+			goto cleanup;
 	}
 
 	/* finish initialization */
--- a/src/filter.c
+++ b/src/filter.c
@@ -266,7 +266,8 @@ int git_filter_register(
 {
 	int error;
 
-	assert(name && filter);
+	GIT_ASSERT_ARG(name);
+	GIT_ASSERT_ARG(filter);
 
 	if (git_rwlock_wrlock(&filter_registry.lock) < 0) {
 		git_error_set(GIT_ERROR_OS, "failed to lock filter registry");
@@ -293,7 +294,7 @@ int git_filter_unregister(const char *name)
 	git_filter_def *fdef;
 	int error = 0;
 
-	assert(name);
+	GIT_ASSERT_ARG(name);
 
 	/* cannot unregister default filters */
 	if (!strcmp(GIT_FILTER_CRLF, name) || !strcmp(GIT_FILTER_IDENT, name)) {
@@ -618,7 +619,7 @@ int git_filter_list_contains(
 {
 	size_t i;
 
-	assert(name);
+	GIT_ASSERT_ARG(name);
 
 	if (!fl)
 		return 0;
@@ -639,7 +640,8 @@ int git_filter_list_push(
 	git_filter_def *fdef = NULL;
 	git_filter_entry *fe;
 
-	assert(fl && filter);
+	GIT_ASSERT_ARG(fl);
+	GIT_ASSERT_ARG(filter);
 
 	if (git_rwlock_rdlock(&filter_registry.lock) < 0) {
 		git_error_set(GIT_ERROR_OS, "failed to lock filter registry");
@@ -684,9 +686,8 @@ static int buf_stream_write(
 	git_writestream *s, const char *buffer, size_t len)
 {
 	struct buf_stream *buf_stream = (struct buf_stream *)s;
-	assert(buf_stream);
-
-	assert(buf_stream->complete == 0);
+	GIT_ASSERT_ARG(buf_stream);
+	GIT_ASSERT(buf_stream->complete == 0);
 
 	return git_buf_put(buf_stream->target, buffer, len);
 }
@@ -694,9 +695,9 @@ static int buf_stream_write(
 static int buf_stream_close(git_writestream *s)
 {
 	struct buf_stream *buf_stream = (struct buf_stream *)s;
-	assert(buf_stream);
+	GIT_ASSERT_ARG(buf_stream);
 
-	assert(buf_stream->complete == 0);
+	GIT_ASSERT(buf_stream->complete == 0);
 	buf_stream->complete = 1;
 
 	return 0;
@@ -740,7 +741,7 @@ int git_filter_list_apply_to_data(
 		&writer.parent)) < 0)
 			return error;
 
-	assert(writer.complete);
+	GIT_ASSERT(writer.complete);
 	return error;
 }
 
@@ -759,7 +760,7 @@ int git_filter_list_apply_to_file(
 		filters, repo, path, &writer.parent)) < 0)
 			return error;
 
-	assert(writer.complete);
+	GIT_ASSERT(writer.complete);
 	return error;
 }
 
@@ -790,7 +791,7 @@ int git_filter_list_apply_to_blob(
 		filters, blob, &writer.parent)) < 0)
 			return error;
 
-	assert(writer.complete);
+	GIT_ASSERT(writer.complete);
 	return error;
 }
 
@@ -809,7 +810,7 @@ static int proxy_stream_write(
 	git_writestream *s, const char *buffer, size_t len)
 {
 	struct proxy_stream *proxy_stream = (struct proxy_stream *)s;
-	assert(proxy_stream);
+	GIT_ASSERT_ARG(proxy_stream);
 
 	return git_buf_put(&proxy_stream->input, buffer, len);
 }
@@ -821,7 +822,7 @@ static int proxy_stream_close(git_writestream *s)
 	git_error_state error_state = {0};
 	int error;
 
-	assert(proxy_stream);
+	GIT_ASSERT_ARG(proxy_stream);
 
 	error = proxy_stream->filter->apply(
 		proxy_stream->filter,
@@ -856,11 +857,12 @@ static int proxy_stream_close(git_writestream *s)
 static void proxy_stream_free(git_writestream *s)
 {
 	struct proxy_stream *proxy_stream = (struct proxy_stream *)s;
-	assert(proxy_stream);
 
-	git_buf_dispose(&proxy_stream->input);
-	git_buf_dispose(&proxy_stream->temp_buf);
-	git__free(proxy_stream);
+	if (proxy_stream) {
+		git_buf_dispose(&proxy_stream->input);
+		git_buf_dispose(&proxy_stream->temp_buf);
+		git__free(proxy_stream);
+	}
 }
 
 static int proxy_stream_init(
@@ -914,7 +916,7 @@ static int stream_list_init(
 		git_filter_entry *fe = git_array_get(filters->filters, filter_idx);
 		git_writestream *filter_stream;
 
-		assert(fe->filter->stream || fe->filter->apply);
+		GIT_ASSERT(fe->filter->stream || fe->filter->apply);
 
 		/* If necessary, create a stream that proxies the traditional
 		 * application.
--- a/src/futils.c
+++ b/src/futils.c
@@ -185,7 +185,8 @@ int git_futils_readbuffer_updated(
 	git_buf buf = GIT_BUF_INIT;
 	git_oid checksum_new;
 
-	assert(out && path && *path);
+	GIT_ASSERT_ARG(out);
+	GIT_ASSERT_ARG(path && *path);
 
 	if (updated != NULL)
 		*updated = 0;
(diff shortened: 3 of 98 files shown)


tests/core/assert.c
```diff
@@ -0,0 +1,37 @@
...
+{
+	cl_git_fail(dummy_fn(NULL));
+	cl_assert(git_error_last());
+	cl_assert_equal_i(GIT_ERROR_INVALID, git_error_last()->klass);
+	cl_assert_equal_s("invalid argument: 'myarg'", git_error_last()->message);
+
+	cl_git_fail(dummy_fn(hello_world));
+	cl_assert(git_error_last());
+	cl_assert_equal_i(GIT_ERROR_INVALID, git_error_last()->klass);
+	cl_assert_equal_s("invalid argument: 'myarg != hello_world'", git_error_last()->message);
```
pks-t: I'd love to have a third test here to verify that `dummy_fn("foo")` will in fact not assert.

tests/core/assert.c
```diff
@@ -0,0 +1,37 @@
...
+}
+
+void test_core_assert__internal(void)
+{
+	cl_git_fail(bad_math());
+	cl_assert(git_error_last());
+	cl_assert_equal_i(GIT_ERROR_INTERNAL, git_error_last()->klass);
+	cl_assert_equal_s("unrecoverable internal error: '1 + 1 == 3'", git_error_last()->message);
+}
+
```
pks-t: There's a trailing newline here :P

tiennou:
I'm in favor of a fix for those "explicit asserts" we carry around, so many thanks for the proposal. 

Here are my thoughts on the subject : I like that `<assert.h>` gives us builds that are as verbose as possible, and fail as precisely as possible, and from the top of my head, I can think of a few usecases we have for asserts internally :
  - NULL-argument-checks, of which there are two kinds, in the closest function they're used in, because preciseness, and ones at the "main function entrypoint". There are many of those.
  - precondition checks, like the accessors in `repository.c`, or "known" enum values. Those are much less common, at least in the parts of the source code I've seen.

To me the former aid as a development/debugging tool, as you can quickly pinpoint unexpected API usage problems — IOW it's trust but verify, as "userland" doesn't have to worry until they have to debug 😉. The latter not so much, as it's usually not possible to fix at runtime, and should be treated as an error instead.

I think that if we want to keep "only one kind of assert" route, I'd go for something behind a CMake option — let's say `ASSERT_HARD` — and select between a `HARD` mode that uses `<assert.h>` + `__builtin_unreachable`, and a normal one that fall through to `git_error` ?

Another suggestion I'd make would be to have the second macro be called `GIT_ASSERT_CHECK` or something softer ? Ideally, if it's about never asserting, maybe we should go for `GIT_ERROR_CHECK_ARGS()`/`GIT_ERROR_CHECK()` ?

ethomson:
I pondered a `GIT_ASSERT` that did a real `assert` _or_ setting an error and jumping out, but I wanted to hear somebody ask before I added the additional complexity.  It sounds like you want that, @tiennou.

Do we need another CMake option, or could we get away with this being a release/debug build difference?

pks-t:
> Do we need another CMake option, or could we get away with this being a release/debug build difference?

Just to make it explicit: debug is assert(3P), release returns an error? If so, then I'm :+1: on that and agree with your sentiment of waiting for complaints first.

ethomson:
> Just to make it explicit: debug is assert(3P), release returns an error? If so, then I'm 👍 on that and agree with your sentiment of waiting for complaints first.

One thing to point out is that whenever we have a debug vs release switch, these are the sorts of things that are tricky to test and can cause end-user errors that we miss.  So we should be vigilant about ensuring that the semantics here are what we want.

ethomson:
I've updated this with two changes:

1. In debug builds, we fall back to `assert(3)`.  This can be overridden by setting `GIT_ASSERT_HARD=0` to use our new functions in debug builds, or `GIT_ASSERT_HARD=1` to use `assert(3)` in release builds.

2. I've added assertion macros that will work for functions that do not return `int`.  The tricky bit here is that there are functions that don't return a value, because there's no recoverable failure case for end-users, but that _do_ have `assert`s in them.

Why is issue 2 a problem?  Well, we could get into a failure case where we stop the function due to an assertion, but the user has no way of knowing that there was a problem, and will blindly call us back with the same object, which may now be in a totally invalid state.  We should validate that any time we assert and _don't_ return an error that we validate the state of the world on subsequent calls and assert in an error-producing way (rather than just crash).

Note that this is not a _new_ problem, it existed any time somebody built without assertions, but I think that it's a problem that exists in the code base.

Pausing here for comments before I do any conversion of `assert` to `GIT_ASSERT`.

pks-t:
> Note that this is not a new problem, it existed any time somebody built without assertions, but I think that it's a problem that exists in the code base.

I agree that this is a very valid concern. I've always disliked functions that do not have a return value in our code base. If they have an assert as part of their body it means they _can_ fail, and as a consequence they should be able to signal errors to their callers and thus return an error code.

As you rightfully point out, this issue is made even more grave right now because we may not even catch situations that'd usually raise an assert in release builds. I could even imagine it to result in silent data corruption or at least an inconsistent state in release builds that would've resulted in a crash in debug builds. And a crash is definitely preferable to doing weird stuff.

So if you haven't guessed yet, my preference would be to not need an assert macro for functions without return value, but to instead convert them to report errors. I don't know how many instances of such functions there are, but I assume converting them all would be a major task and not really feasible?

ethomson:
> So if you haven't guessed yet, my preference would be to not need an assert macro for functions without return value, but to instead convert them to report errors. I don't know how many instances of such functions there are, but I assume converting them all would be a major task and not really feasible?

Interesting point.  I was mostly concerned about public functions.  So, let's put internal functions aside, since we can do whatever we want with them.

There are a total of **84** `GIT_EXTERN` marked functions in our public API.

A lot of them are things like `git_foo_free`.  Which...  can't assert.  Either you gave it `NULL`, and that's no problem!  Or you gave it non-`NULL` but we wouldn't have any way of determining whether the pointer was valid or not.  So we just have to feed it to the configured `free` function and we'll either crash or not.  🤷‍♂ 

If I remove the `free` and `dispose` functions, then that gets us down to 25.  Things like `git_error_clear` probably can't fail.  But things like `git_oid_fmt`, where we legitimately might want to assert on a `NULL` out param (or input for that matter).  And even things that are `GIT_EXTERN(void) git_foo(void)` could want to assert based on internal state.

So ...  here's a question: let's ignore the amount of work involved.  Is changing a function that returns `void` to returning `int` a _breaking API change_?  That was why I was 😟 but when you replied without mentioning it, I actually thought about it a little bit more.  I would argue: **no**, it's really **not** a breaking change.

Nobody's code is going to stop compiling because we've given it an `int` return type -- okay, not unless they have warnings enabled such that they get warned on ignoring return type, and I think that "newly emitting a warning" is safely within the bounds of API compatibility.

I think that's a great path forward, personally.

pks-t:
> So ... here's a question: let's ignore the amount of work involved. Is changing a function that returns void to returning int a breaking API change? That was why I was  but when you replied without mentioning it, I actually thought about it a little bit more. I would argue: no, it's really not a breaking change.

I agree with you. And 25 functions doesn't sound too unreasonable to get fixed without spending too much effort on it.

ethomson:
I rebased this on master and dropped the semantics to support `void` functions.  Now the `GIT_ASSERT` macros must return _something_ in the case of an assertion, which - I think - makes sense.

Unless somebody's violently opposed to this, I'll update the codebase to use these macros over the weekend.

pks-t:
> I rebased this on master and dropped the semantics to support void functions. Now the GIT_ASSERT macros must return something in the case of an assertion, which - I think - makes sense.

I think so, too. If there's any reason for an `assert`, the function already has a failing condition and thus should return an error code. End of the story, if you ask me.

src/assert_safe.h
```diff
@@ -0,0 +1,55 @@
...
+
+/*
+ * In a debug build, we'll assert(3) for aide in debugging.  In release builds,
+ * we will provide macros that will set an error message that indicate a failure
+ * and return.  Note that memory leaks can occur in a release-mode assertion
+ * failure -- it is impractical to provide safe clean up routines in these very
+ * extreme failures, but care should be taken to not leak very large objects.
+ */
+
+#if (defined(_DEBUG) || defined(GIT_ASSERT_HARD)) && GIT_ASSERT_HARD != 0
```
pks-t, line 18: Shouldn't this rather be `#if defined(_DEBUG) || (defined(GIT_ASSERT_HARD) && GIT_ASSERT_HARD)`?

pks-t: And I think we should also include <assert.h>, right?

src/assert_safe.h
```diff
@@ -0,0 +1,55 @@
...
+/** Internal consistency check to return the `fail` param on failure. */
+# define GIT_ASSERT_EXT(expr, fail) \
+	GIT_ASSERT__EXT(expr, GIT_ERROR_INTERNAL, "unrecoverable internal error", fail)
+
+/**
+ * Assert that a consumer-provided argument is valid, setting an
+ * actionable error message and returning the `fail` param if not.
+ */
+# define GIT_ASSERT_ARG_EXT(expr, fail) \
+	GIT_ASSERT__EXT(expr, GIT_ERROR_INVALID, "invalid argument", fail)
```
pks-t, line 44: I'm not really much of a fan of the `_EXT` suffix, as it doesn't really convey any meaning without having a look at the function's documentation. So how about the much longer but more expressive `GIT_ASSERT_WITH_RETVAL` and `GIT_ASSERT_ARG_WITH_RETVAL`? They are definitely more tedious to write, but I'd expect most of the code to only require `GIT_ASSERT` and `GIT_ASSERT_ARG` anyway.

ethomson:
I've been going through and converting our uses of `assert` to the new macros.  This is mostly mechanical, but there are a few notes:

- I've done it file-at-a-time so that I can pause and continue the conversion, minimize conflicts and so that reviewers can pick up where they left off
- The public facing functions should use the new assertions
- I didn't add any new assertions, to reduce scope and risk.  This seems like a reasonable follow-up.
- There are a few internal functions where we weren't returning anything, and I've updated many of them to return things
- There are a few internal functions where we weren't returning anything but propagating error checking would be a significant change.  In those cases, I've removed usage of `assert`.  These are internal functions that don't take user input.

I'm nearly done with the files in `src/` (but not subdirectories), sans a few trickier files that I'll come back to.  It might make sense to start reviewing this though?  It would be nice to get moving forward rather quickly since it will touch nearly every file, and I fear the conflicts.

src/assert_safe.h
ethomson: I don't think so - `GIT_ASSERT_HARD` is meant to override `_DEBUG`.  So `GIT_ASSERT_HARD=0` will go to our error set and return -1 functions even when `_DEBUG` is set.  `GIT_ASSERT_HARD=1` will go to the system assert, even when `_DEBUG` is unset.

pks-t:
On Sun, Apr 05, 2020 at 02:55:49PM -0700, Edward Thomson wrote:
> I'm nearly done with the files in `src/` (but not subdirectories),
> sans a few trickier files that I'll come back to.  It might make sense
> to start reviewing this though?  It would be nice to get moving
> forward rather quickly since it will touch nearly every file, and I
> fear the conflicts.

Agreed, I'll prioritize this PR.

src/assert_safe.h
pks-t: Ok. The only case that makes me wonder is if `GIT_ASSERT_HARD` isn't defined at all, as I'm not quite sure how above expression would be evaluated. So can this actually happen?

src/assert_safe.h
```diff
@@ -21,28 +21,35 @@
...
+/**
+ * Assert that a consumer-provided argument is valid, setting an
+ * actionable error message and returning the `fail` param if not.
+ */
+# define GIT_ASSERT_ARG_WITH_RETVAL(expr, fail) \
+	GIT_ASSERT__WITH_RETVAL(expr, GIT_ERROR_INVALID, "invalid argument", fail)
 
-/* Internal consistency check to stop the function. */
-# define GIT_ASSERT(expr) do { \
+# define GIT_ASSERT__WITH_RETVAL(expr, code, msg, fail) do { \
```
pks-t, line 49: Not sure if I'm misreading, but it looks to me like `GIT_ASSERT_WITH_RETVAL` has different parameters in the `GIT_ASSERT_HARD` case compared to the `!GIT_ASSERT_HARD` case.

Edit: misread it. There's two underscores here. Easy to miss that

src/apply.c
```diff
@@ -5,8 +5,6 @@
  * a Linking Exception. For full terms see the included COPYING file.
  */
 
-#include "apply.h"
-
```
pks-t, line 9: We have the rule that a given implementation should always include its own header first. 
This is mainly to make sure that we've got the include chain correct: apply.c -> apply.h -> common.h. With the changed include, we now have our public headers first and are thus missing "common.h" as first include.

This is another thing I should probably have in our coding style document.

ethomson:
> This is a lot of changes, thanks for taking care of it! I hope you don't mind me asking, but could we maybe split this up into smaller chunks?

Seems reasonable.  #5509 is the first piece of this.  I'll do conversions of `void` functions to return `int`s next once that's 👍

PR #5336, opened 2019-12-13
cred: change enum to git_credential_t and GIT_CREDENTIAL_*

ethomson:
We have standardized on a trailing `_t` for enum types, instead of using "type" in the name.  So `git_credtype_t` is more appropriately named `git_credential_t`.

Change the name, keeping the previous name and values as deprecated.


ethomson:
We've gone on a tear and removed abbreviations.  Maybe `git_credential_t` and moving `cred` -> `credential` would be more helpful?  It's more code churn but it feels more consistent.

pks-t:
Hm, I'm honestly not too sure about this. I mean we _are_ discerning credential types here, right? There's definitely some inconsistencies in our tree:

- git_merge_diff_type_t
- git_rebase_type_t
- git_object_t
- git_reference_t

Well, seeing that we've already moved from `git_otype_t` to `git_object_t` and that object and reference types are probably one of the most commonly used types, let's go ahead and remove the "type" stuff. I'd love to do the same for `git_merge_diff_type_t` and `git_rebase_type_t` then, though. Do you want to tackle this or shall I do it?

And yes, if we're already dropping "type" then I'd vote for going with "git_credential_t".

ethomson:
> Do you want to tackle this or shall I do it?

I'll knock this out over the weekend.

ethomson:
> I'd love to do the same for git_merge_diff_type_t and git_rebase_type_t then, though. Do you want to tackle this or shall I do it?

I did this over in #5364

ethomson:
> And yes, if we're already dropping "type" then I'd vote for going with "git_credential_t".

OK, I've refactored `git_cred` to be `git_credential`, `git_credtype_t` to be `git_credential_t` and `GIT_CREDTYPE` to be `GIT_CREDENTIAL`.

I also renamed the `cred.[ch]` files to be `credential.[ch]`.  This was...  a little odd - especially for `cred_helpers.h`, and `sys/cred.h` which we expect callers to include directly (instead of as part of `git2.h`).

include/git2/cred_helpers.h
```diff
@@ -7,47 +7,9 @@
...
-		git_cred **cred,
-		const char *url,
-		const char *user_from_url,
-		unsigned int allowed_types,
-		void *payload);
-
+/* These declarations have moved. */
+#ifndef GIT_DEPRECATE_HARD
+# include "git2/credential_helpers.h"
+#endif
```
pks-t: Wouldn't we want to do the same header-forwarding for "git2/cred.h"?

src/streams/openssl.c.bak
```diff
@@ -0,0 +1,795 @@
+/*
```
pks-t: This file doesn't look like it belongs here :)

include/git2/cred_helpers.h
ethomson: We could, but our guidance has always been for people to include `git2.h`, which will bring in all the things except `sys/` and - somewhat oddly - `cred_helpers.h`.  So people should not have been including `git2/cred.h` directly.

This feels a little weird to me though.  I almost feel like we should just include these functions as part of the overall `git2.h` experience.

src/streams/openssl.c.bak
ethomson: Oops.  Thanks - fixed!

src/transports/credential.c
```diff
@@ -0,0 +1,476 @@
...
+{
+	return git_credential_get_username(cred);
+}
+
+int git_cred_userpass_plaintext_new(
+	git_credential **out,
+	const char *username,
+	const char *password)
+{
+	return git_credential_userpass_plaintext_new(out,username, password);
```
pks-t: Missing space

PR #5338, opened 2019-12-13
patch_parse: fix undefined behaviour due to arithmetic on NULL pointers

pks-t:
Doing arithmetic with NULL pointers is undefined behaviour in the C
standard. We do so regardless when parsing patches, as we happily add a
potential prefix length to prefixed paths. While this works out just
fine as the prefix length is always equal to zero in these cases, thus
resulting in another NULL pointer, it still is undefined behaviour and
was pointed out to us by OSSfuzz.

Fix the issue by checking whether paths are NULL, avoiding the
arithmetic if they are.


ethomson:
Nice catch, thanks @pks-t

PR #5339, opened 2019-12-13
netops: handle intact query parameters in service_suffix removal

### LINKED ISSUE #5321: /info/refs http request malformed when base remote path is /
### Reproduction steps

Construct an http git remote whose URL has path `/`, such as `http://localhost:8080/`. Do a push.

### Expected behavior

libgit2 sends an HTTP request with URL `/info/refs?service=git-receive-pack`. (This is what command line git does.)

### Actual behavior

libgit2 sends an HTTP request with URL `/info/refs?service=git-receive-pack/info/refs?service=git-receive-pack`.

7acb764eeeea22cbfaa4dd625bc8516966ed96f3 (master at time of posting)

Remote URLs of this form also fail with v28, but for different reasons. (c6ab183e9c960b74471a7d106a4deb0c8b28a8ec, which postdates v28, fixed other bugs with such URLs.)

### OS

macOS

josharian:
Some servers leave the query parameters intact in the
Location header when responding with a redirect.
The service_suffix removal check as written assumed
that the server removed them.
Handle both cases.

Along with PR #5325, this fixes #5321.

Please review skeptically; my C is extremely rusty.

### DIFF
--- a/src/netops.c
+++ b/src/netops.c
@@ -171,25 +171,46 @@ int gitno_connection_data_handle_redirect(
 
 	/* Remove the service suffix if it was given to us */
 	if (service_suffix) {
+		/*
+		 * Some servers strip the query parameters from the Location header
+		 * when sending a redirect. Others leave it in place.
+		 * Check for both, starting with the stripped case first,
+		 * since it appears to be more common.
+		 */
 		const char *service_query = strchr(service_suffix, '?');
+		size_t full_suffix_len = strlen(service_suffix);
 		size_t suffix_len = service_query ?
-			(size_t)(service_query - service_suffix) : strlen(service_suffix);
+			(size_t)(service_query - service_suffix) : full_suffix_len;
 		size_t path_len = strlen(url->path);
+		ssize_t truncate = -1;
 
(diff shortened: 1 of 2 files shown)


src/netops.c
```diff
@@ -171,25 +171,41 @@ int gitno_connection_data_handle_redirect(
...
-				/* Ensure we leave a minimum of '/' as the path */
-				if (suffix_offset == 0)
-					suffix_offset++;
+				truncate = suffix_offset;
+			}
+		}
 
-				url->path[suffix_offset] = '\0';
+		/* Check for retained query parameters. */
+		if ((truncate == -1) && (git__suffixcmp(url->path, service_suffix) == 0)) {
```
pks-t: We usually have no braces around comparisons

josharian: Done

josharian:
> I honestly find it very hard to reason about this, but this is no fault of yours.

That's kind, but I'll take the fault. I've added more documentation in the code.

> We do have tests in "tests/network/redirect.c" that are supposed to exercise this function. 

Thanks for the pointer. I hadn't found those, only the tests against live servers. I've added two tests. The first already passed; the second previously failed.

Issue #5324, opened 2019-12-06
`pack::packbuilder` tests failing on v0.28.3


krallin:
### Reproduction steps

I took a fresh clone of v0.28.3, and I'm building it and running the packbuilder tests using:

```
make -DUSE_HTTPS=OFF -DUSE_NTLMCLIENT=OFF .. && cmake --build . && ./libgit2_clar -spack::packbuilder
```

### Expected behavior

The tests pass.

### Actual behavior

Unfortunately, this fails on my system, with the following errors:

```

  1) Failure:
pack::packbuilder::create_pack [/home/torozco/local/foo/libgit2/tests/pack/packbuilder.c:139]
  String mismatch: hex != "5d410bdf97cf896f9007681b92868471d636954b"
  '5f688cedc3a5132f15fc3352a35bc5dcb6a536a4' != '5d410bdf97cf896f9007681b92868471d636954b' (at byte 1)

  2) Failure:
pack::packbuilder::get_hash [/home/torozco/local/foo/libgit2/tests/pack/packbuilder.c:151]
  String mismatch: hex != "7f5fa362c664d68ba7221259be1cbd187434b2f0"
  'f120f6bbad709d0e3c021ca8f9d7c26c3882497b' != '7f5fa362c664d68ba7221259be1cbd187434b2f0' (at byte 0)

  3) Failure:
pack::packbuilder::permissions_standard [/home/torozco/local/foo/libgit2/tests/pack/packbuilder.c:175]
  Function call failed: (stat("pack-7f5fa362c664d68ba7221259be1cbd187434b2f0.idx", &statbuf))
  error -1 - <no message>

  4) Failure:
pack::packbuilder::permissions_readonly [/home/torozco/local/foo/libgit2/tests/pack/packbuilder.c:175]
  Function call failed: (stat("pack-7f5fa362c664d68ba7221259be1cbd187434b2f0.idx", &statbuf))
  error -1 - <no message>

  5) Failure:
pack::packbuilder::permissions_readwrite [/home/torozco/local/foo/libgit2/tests/pack/packbuilder.c:175]
  Function call failed: (stat("pack-7f5fa362c664d68ba7221259be1cbd187434b2f0.idx", &statbuf))
  error -1 - <no message>
```

It looks like those are all symptoms of the same underlying problem (i.e. the hash generated by packbuilder is different).

On another machine, I'm able to get the tests to pass. Here is is the configure output for the machine I'm working on:

```
[...]

PR #5340, opened 2019-12-14
Refactor packfile code to use zstream abstraction

pks-t:
While investigating #5324, I noticed that the zlib-related code in packfile looks mighty complicated and could use some love. I thus refactored it to use our own `git_zstream` abstraction, which is a much nicer interface.

### DIFF
--- a/src/pack.c
+++ b/src/pack.c
@@ -13,9 +13,6 @@
 #include "odb.h"
 #include "oid.h"
 #include "sha1_lookup.h"
-#include "zstream.h"
-
-#include <zlib.h>
 
 /* Option to bypass checking existence of '.keep' files */
 bool git_disable_pack_keep_file_checks = false;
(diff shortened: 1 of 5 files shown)


pks-t:
The code tended to get stuck randomly depending on how fast bytes came in through the network layer. The issue is taht `git_zstream_get_output` reads until all output bytes have been filled _or_ the stream is at end. If the amount of input bytes cannot fill the output buffer _and_ will not cause the current object to be finished, then we get stuck in an endless loop around `git_zstream_get_output_chunk`.

I tried fixing `git_zstream_get_output`, but it's near impossible to get right as other callers then fail. By converting to `git_zstream_get_output_chunk`, though, we are able to get the streaming code correct, where it will now return GIT_EBUFS as expected in case we didn't get enough input bytes. That fixes the endless loop.

Thus: never use `git_zstream_get_output` in a streaming scenario, but only ever `git_zstream_get_output_chunk`.

pks-t:
Yup, green now. @ethomson, want to have another look? As said, the only change was to replace `git_zstream_get_output` with `git_zstream_get_output_chunk`.

ethomson:
Thanks for the ping @pks-t  -  I restarted the build and then forgot about it.

PR #5345, opened 2019-12-22
emscripten support

petersalomonsen:
This takes a different approach than #4400 as there are hardly any changes to existing code. A separate transport is added for emscripten, and the built-in examples are used for calling from js (added an extra example for commit). Instead of adding a workaround for emscripten mmap, I've submitted a PR for handling offset https://github.com/emscripten-core/emscripten/pull/10095, and https://github.com/emscripten-core/emscripten/pull/10526.

Would be great with some feedback on this approach compared to my previous pull-request (#4400).


petersalomonsen:
added example for `push` and also a local git http server for testing.

pks-t:
I'm still not sure about whether we want to support Emscripten in libgit2 directly, which is probably why I've deferred reviewing this for so long. Sorry about that! I'd imagine that we should treat Emscripten the same as we do treat bindings: provide the required bits in order to make it work together so that you can live as a downstream user of libgit2.

As far as I see, the only real change required on our side is swapping out the HTTP transport. I'm honestly not too keen to include the Emscripten transport directly, but we already provide `git_transport_register` with which you can register an externally hosted Emscripten transport. The only thing we'd have to do is make the HTTP transport build-time configurable, which I'd be fine with.

Does the above make sense to you? Do you feel like this is a feasible way to go?

Anyway, I'd definitely like to pull in your changes to our examples. Would you like to spin up a separate PR for them?

petersalomonsen:
Yes that sounds like a good idea. Good to get this clarified. 

If HTTP transport can be configurable at build time, I can set up a separate project for emscripten and reference libgit2 from there.

And yes, I'll look into making a separate PR for the examples.

Thanks for review and response :-)

ethomson:
Thanks @petersalomonsen for your patience.  And thanks @pks-t for this wonderful suggestion.  I think that this is a great path forward.

I'm really thrilled to see that there is so little code churn.  emscripten seems to have become incredibly mature!  The mmap support in particular is great to see.

petersalomonsen:
moved to separate project here:

https://github.com/petersalomonsen/wasm-git

pks-t:
On Sun, Feb 23, 2020 at 02:19:23AM -0800, Peter Salomonsen wrote:
> moved to separate project here:
> 
> https://github.com/petersalomonsen/wasm-git

Cool! Please feel free to create a PR for libgit2 that links to your
project in our README's language bindings to make it more discoverable.

PR #5347, opened 2019-12-31
Support `core.longpaths` on Windows

ianhattendorf:
Initial test run at supporting paths > `260` on Windows (#3053). Works as far as I can tell (clone, commit, checkout, etc.) but definitely needs more testing.

TODO:
- [x] ~~Do we need to respect `core.longpaths`, or can we simply always support long paths? I'm assuming we need to respect it (https://xkcd.com/1172/), but AFAIK this exists because CLI Git for Windows doesn't support long paths across all commands/internal scripts, and is seen as experimental.~~
- [x] ~~Potentially keep static buffers at `260`, dynamically allocate past that. Is an increase in path static buffer size from `532`/`778` bytes to `8204`/`12286` bytes too much for Win32? Static is nice here since there is no allocation performance penalty regardless of whether short or long paths are used. Do we support variants of Windows with small stack sizes (e.g. Windows CE)?~~
- [x] Tests tests tests tests tests
- [x] Test cleanup fails (`DeleteFileW(buffer)`)


ethomson:
Awesome!

> Do we need to respect core.longpaths, or can we simply always support long paths?

I think we should honor it.  There are just too many tools on windows that don’t understand long paths that this should be opt in so that people who shoot themselves in the foot had to explicitly load the gun.

ianhattendorf:
A few updates:
- I added a new devops job to test Windows 64 bit Visual Studio with longpaths enabled. I didn't want to modify any existing Windows jobs since longpaths off is still the default, however wanted to verify nothing else breaks when they're enabled. Let me know if this is too much additional CI time or if there's a better option.
- Instead of supporting `core.longpaths`, we respect the option `GIT_OPT_SET_WINDOWS_LONGPATHS`. This allows some configurability, without threading `core.longpaths` through every function call.

I'm sure this still needs more tests for the longpaths case, however I believe it's ready to go functionally.

pde-cds:
Thank you for solving this long existing problem @ianhattendorf 
Is the review of this PR planned in the near future?

japj:
I also ran into a long path issue during a clone of a repository.
what is currently blocking this from getting into master?

gigi81:
I also run into this issue today. any chance this could be merged?

pde-cds:
@ethomson @pks-t 
Is the review of this MR currently planned? 
Or is anything other blocking this issue from getting merged?

rupreck:
Really would like to see this merged.  Lack of support for this is becoming increasingly a problem with various project package components (including node_modules) and testing frameworks with deep descriptive paths unable to be consumed.

ethomson:
I would too!  Could you do a code review?  I'd love to see timings on non-long paths to ensure we haven't regressed.  Maybe creating a test framework that can help us understand that would be a good first step for a new contributor.

ethomson:
@ianhattendorf sorry about the long delay.  I'm trying to go through our backlog of issues.  I really like this approach, I think that it makes sense.  It's not what I had in mind, and this is better.

I have a few comments and requests - but since it's been so long, I wanted to check in and see if this was still something that you wanted to pursue.  If you're interested in iterating on it, then I'll post feedback.  If you've moved on - and there's no problem if you have - it's been nearly a year, and that's my fault, and I apologize - but if you don't have any appetite to pick up the review, then I can just move this forward.  There's nothing major, I don't think, this looks quite nice as-is.

ianhattendorf:
@ethomson no problem, I'm sure you're busy. Glad you had the chance to look through it. I'm still up for working through everything when you are 👍

cjmanca:
Is this still being worked on? I'm also interested in it being incorporated

ianhattendorf:
@ethomson rebased onto latest master. I'm sure you're busy with the holidays coming up, but when you have time it would be great if you could leave any feedback you have.

navono:
any progress?

sarath-kumar-harikrishnan:
@ianhattendorf @ethomson any updates/timelines that we can expect this feature gets rolled out to users

Rossbro2:
Just ran into an issue today that seems it could be solved by this PR.  Any ETA?  Much appreciated.

pde-cds:
@ethomson @ianhattendorf What is currently preventing this PR from getting merged?
Considering all the people commenting here i still think this needs to be included in libgit2 asap.

ethomson:
This depends on #5823.

I'm deleting "me too!" or "why isn't this merged yet!" comments.  Please don't post those.  (If you want this feature as-is, include this PR in your own build of libgit2.)

ethomson:
I opened #5857 with this, and a few minor changes.  Of note:

1. Our constants are `GIT_something`, so `WIN_GIT_PATH_MAX` should be `GIT_WIN_PATH_MAX` or something similar.
2. `clar/fs.h` shouldn't depend on libgit2 directly, if possible - it should stay standalone so that we can pull in [upstream changes](https://github.com/clar-test/clar) whenever we need to.
3. Since we've added support for enforcing `core.longpaths` in #5823, we can remove the `GIT_OPT_SET_WINDOWS_LONGPATHS` option and enforce `core.longpaths` on a repository-by-repository basis.

ethomson:
Thanks for this @ianhattendorf!  I'm going to bring it in over at #5857!  🎉

PR #5350, opened 2020-01-08
futils_mktmp: don't use umask.

jmgao:
Previously, we were using `umask(mask = umask(0))` to fetch the current
umask in order to apply it to the desired mode, but this is broken in
the presence of threads. There is no portable way to directly fetch
umask without mutating it.

Instead, create the file with 0777 permissions, and fstat it to reverse
engineer what umask was when the file was created, and then apply that
as a mask on the mode.

Fixes: jmgao/pore#46

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -22,6 +22,12 @@ char git_buf__oom[1];
 	    ((d) > (b)->asize && git_buf_grow((b), (d)) < 0))\
 		return -1;
 
+int git_buffer_global_init(void)
+{
+	/* Seed rand for git_buf_put_rand. */
+	srand((unsigned int)time(NULL));
+	return 0;
+}
 
 int git_buf_init(git_buf *buf, size_t initial_size)
 {
@@ -239,6 +245,20 @@ int git_buf_puts(git_buf *buf, const char *string)
 {
 	assert(string);
 	return git_buf_put(buf, string, strlen(string));
+}
+
+static const char rand_chars[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+
+int git_buf_put_rand(git_buf *buf, size_t chars)
+{
+	size_t i;
+
+	for (i = 0; i < chars; ++i) {
+		if (git_buf_putc(buf, rand_chars[rand() % strlen(rand_chars)]) == -1)
+			return -1;
+	}
+
+	return 0;
 }
 
 static const char base64_encode[] =
--- a/src/futils.c
+++ b/src/futils.c
@@ -7,6 +7,7 @@
 
 #include "futils.h"
 
+#include "buffer.h"
 #include "global.h"
 #include "strmap.h"
 #include <ctype.h>
@@ -21,32 +22,30 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
 		GIT_MKDIR_PATH | GIT_MKDIR_SKIP_LAST | GIT_MKDIR_VERIFY_DIR);
 }
 
+/* Reimplementation of mkstemp, except with a umask-modified mode, instead of 0600. */
 int git_futils_mktmp(git_buf *path_out, const char *filename, mode_t mode)
 {
+	/* glibc tries 238328 times, OpenBSD tries INT_MAX times. */
+	unsigned int tries = INT_MAX;
 	int fd;
-	mode_t mask;
+	const int open_flags = O_RDWR | O_CREAT | O_EXCL | O_BINARY | O_CLOEXEC;
 
-	p_umask(mask = p_umask(0));
+	while (tries--) {
+		git_buf_sets(path_out, filename);
+		git_buf_puts(path_out, "_git2_");
+		git_buf_put_rand(path_out, 8);
 
-	git_buf_sets(path_out, filename);
-	git_buf_puts(path_out, "_git2_XXXXXX");
+		if (git_buf_oom(path_out))
+			return -1;
 
-	if (git_buf_oom(path_out))
-		return -1;
+		if ((fd = p_open(path_out->ptr, open_flags, mode)) == -1)
+			continue;
 
-	if ((fd = p_mkstemp(path_out->ptr)) < 0) {
-		git_error_set(GIT_ERROR_OS,
-			"failed to create temporary file '%s'", path_out->ptr);
-		return -1;
+		return fd;
 	}
 
-	if (p_chmod(path_out->ptr, (mode & ~mask))) {
-		git_error_set(GIT_ERROR_OS,
-			"failed to set permissions on file '%s'", path_out->ptr);
-		return -1;
-	}
-
-	return fd;
+	git_error_set(GIT_ERROR_OS, "failed to create temporary file");
+	return -1;
 }
 
 int git_futils_creat_withpath(const char *path, const mode_t dirmode, const mode_t mode)
--- a/src/buffer.h
+++ b/src/buffer.h
@@ -20,6 +20,8 @@
 extern char git_buf__initbuf[];
 extern char git_buf__oom[];
 
+int git_buffer_global_init(void);
+
 /* Use to initialize buffer structure when git_buf is on stack */
 #define GIT_BUF_INIT { git_buf__initbuf, 0, 0 }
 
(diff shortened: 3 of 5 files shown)


ethomson:
Reading this quickly on my phone, so apologies if I misread - but it looks to me like this opens the new file world-writable, which would give us a different race condition.

Instead, why don't we add a new function -- `p_getumask` -- which  uses `getumask(3)` where it's available (eg, Linux) and reimplements it where it's not.  This reimplementation can do the create a file and check its mode mechanism that you've implemented here, but it then deletes that file.  This removes the world-writable file that may become a configuration file race.

jmgao:
I considered something like that, but it doesn't seem obvious that there's a good place to try to create the file to probe for umask, since there might be, for example, a noexec mount that screws with your probed umask. Now that I think about it, we can just do this twice to avoid the race: once to get the umask and throw away, and once with the proper permissions.

(getumask is [apparently vaporware](http://man7.org/linux/man-pages/man3/getumask.3.html): grepping through the kernel confirms that there isn't a getumask syscall.)

jmgao:
Wait, we don't need to chmod at all, we can just open with that mode directly. chmod was only needed before because mkstemp creates the files with 0600 permissions.

neithernut:
> getumask is apparently vaporware: grepping through the kernel confirms that there isn't a getumask syscall.

Well, `getumask(3)` is a libc function. And a GNU extension. According to the man-page it's equivalent to:

    mode_t getumask(void) {
        mode_t mask = umask(0);
        umask(mask);
        return mask;
    }

So there is probably really no benefit in using it if it is available.

Pedantic nit: the term "vaporware" refers to software that was only announced but never implemented. This function _was_ implemented, so it doesn't fit the definition. Still, it's pretty useless if you consider portability.

jmgao:
It's declared in the glibc headers, but there's only an implementation on Hurd. If you try to call it on Linux, you'll [fail to link](https://gcc.godbolt.org/z/SgWc5-).

src/futils.c
```diff
@@ -21,32 +21,41 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
 		GIT_MKDIR_PATH | GIT_MKDIR_SKIP_LAST | GIT_MKDIR_VERIFY_DIR);
 }
 
+static void git_buf_put_rand(git_buf *buf, size_t chars)
```
ethomson: This probably belongs in `buffer.c`.

src/futils.c
```diff
@@ -21,32 +21,41 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
...
+{
+	size_t i;
+	const char* alphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+
+	for (i = 0; i < chars; ++i) {
```
ethomson: MInor formatting nit: we'd prefer no braces around a single line.

src/futils.c
```diff
@@ -21,32 +21,41 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
...
+		git_buf_putc(buf, alphabet[rand() % strlen(alphabet)]);
+	}
+}
+
+// Reimplementation of mkstemp, except with a umask-modified mode, instead of 0600.
```
ethomson: Minor nit: we use `/*` and `*/` comments, we don't use `//`.

src/futils.c
```diff
@@ -21,32 +21,41 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
...
+	unsigned int tries = INT_MAX;
 	int fd;
-	mode_t mask;
-
-	p_umask(mask = p_umask(0));
+	const int open_flags = O_RDWR | O_CREAT | O_EXCL | O_BINARY | O_CLOEXEC;
 
-	git_buf_sets(path_out, filename);
-	git_buf_puts(path_out, "_git2_XXXXXX");
+	while (--tries > 0) {
```
ethomson: Minor request: I think I'd prefer `tries--` here so that we could have just a single try by specifying `tries = 1`.

src/futils.c
```diff
@@ -21,32 +21,41 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
...
-		return -1;
+		if (git_buf_oom(path_out))
+			return -1;
 
-	if ((fd = p_mkstemp(path_out->ptr)) < 0) {
-		git_error_set(GIT_ERROR_OS,
-			"failed to create temporary file '%s'", path_out->ptr);
-		return -1;
-	}
+		if ((fd = p_open(path_out->ptr, open_flags, mode)) == -1) {
```
ethomson: MInor formatting nit: we'd prefer no braces around a single line.

src/futils.c
jmgao: I was hesitant to put it in buffer.c because the "randomness" is really bad, but there doesn't seem to be any existing abstraction for randomness. It's probably mostly fine in this case, because we're opening the file with the precise permissions we want, so the worst that happens is that if multiple processes try to create temp files, a few of them will have to retry a few times, but I didn't want to expose this function to users that would expect actual randomness.

jmgao:
Sorry, I've been meaning to dig into the Windows failures, but I haven't managed to scrape up the time. The error seems like a leak that presumably should happen on all platforms, but after staring at the code a bit, I don't see where it is.

Maybe the test is returning early and failing to clean up the buffer because one of the assertions is failing on windows? I didn't see anything that indicated that in the logging, but I might have missed it.

src/futils.c
ethomson: It’s still best to put it in with the rest of the buffer code from an organizational standpoint.  That’s all internal to the library and I don’t think anyone will go assuming that it’s crypto-worthy.

Having said that, we may want to introduce a random number generator for the library that produces better randomness but `rand` is so meh that it might be nice.

ethomson: It’s still best to put it in with the rest of the buffer code from an organizational standpoint.  That’s all internal to the library and I don’t think anyone will go assuming that it’s crypto-worthy.

Having said that, we may want to introduce a random number generator for the library that produces better randomness but `rand` is so meh that it might be nice.

ethomson:
> Maybe the test is returning early and failing to clean up the buffer because one of the assertions is failing on windows?

That’s correct - our tests don’t tear everything down when they fail, so we’ll detect leaks there.  Apologies that this isn’t more obvious; our CI grew rather organically, and leak detection was added on.

We may be able to disable printing the leaks if clar exits with a nonzero exit code.  If nothing else, we could print a big banner at the bottom of the output on a failure.

ethomson:
I do have a concern about `rand` without `srand`.  From [a man page](http://man7.org/linux/man-pages/man3/rand.3.html):

> If no seed value is provided, the rand() function is automatically seeded with a value of 1.

Which would mean that our temporary files are deterministic.  I can imagine a denial of service attack based on this - or perhaps more likely, we fail to create new files because we neglected to clean up the old ones accidentally.

src/futils.c
jmgao: Fixed(ish: by adding another line to the block)

src/futils.c
jmgao: Done.

src/futils.c
jmgao: Done.

src/futils.c
jmgao: Done.

src/futils.c
jmgao: Done.

jmgao:
Disabled the test on windows, and cleaned up all of the nits: hopefully this revision passes CI.

src/buffer.c
```diff
@@ -22,6 +22,12 @@ char git_buf__oom[1];
 	    ((d) > (b)->asize && git_buf_grow((b), (d)) < 0))\
 		return -1;
 
+int git_buffer_global_init(void)
+{
+	/* Seed rand for git_buf_put_rand. */
+	srand((unsigned int)time(NULL));
```
pks-t: We really shouldn't be using `rand()` and `srand()` at all. Seeding via `time(NULL)` only gives you a negligible amount of bits, as you can typically deduce at what time a certain command is running. I don't know about Windows, but on Unix-like systems most nowadays support calls like `getrandom()` that are much saner to use than the combination of `rand()` and `srand()`.

src/futils.c
```diff
@@ -21,32 +22,30 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
...
+	unsigned int tries = INT_MAX;
 	int fd;
-	mode_t mask;
+	const int open_flags = O_RDWR | O_CREAT | O_EXCL | O_BINARY | O_CLOEXEC;
 
-	p_umask(mask = p_umask(0));
+	while (tries--) {
+		git_buf_sets(path_out, filename);
+		git_buf_puts(path_out, "_git2_");
+		git_buf_put_rand(path_out, 8);
```
pks-t: I'd argue that manually generating the name doesn't make a lot of sense. Instead, we should probably be using something like tempnam(3P) that properly handles this for us, and that would also remove the need for `git_buf_put_rand` and the random-seeding hassle.

ethomson:
Closed via #6178

PR #5351, opened 2020-01-09
index: replace map macros with inline functions

pks-t:
Traditionally, our maps were mostly implemented via macros that had
weird call semantics. This shows in our index code, where we have macros
that insert into an index map case-sensitively or insensitively, as they
still return error codes via an error parameter. This is unwieldy and,
most importantly, not necessary anymore, due to the introduction of our
high-level map API and removal of macros.

Replace them with inlined functions to make code easier to read.

### DIFF
--- a/src/index.c
+++ b/src/index.c
@@ -27,29 +27,6 @@
 #include "git2/config.h"
 #include "git2/sys/index.h"
 
-#define INSERT_IN_MAP_EX(idx, map, e, err) do {				\
-		if ((idx)->ignore_case)					\
-			(err) = git_idxmap_icase_set((git_idxmap_icase *) (map), (e), (e)); \
-		else							\
-			(err) = git_idxmap_set((map), (e), (e));	\
-	} while (0)
-
-#define INSERT_IN_MAP(idx, e, err) INSERT_IN_MAP_EX(idx, (idx)->entries_map, e, err)
-
-#define LOOKUP_IN_MAP(v, idx, k) do {					\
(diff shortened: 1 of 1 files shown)


pks-t:
I've amended another commit that goes into the same direction, but fixes a real bug where we'd always try to resize index maps twice on case-insensitive systems by accident.

src/index.c
```diff
@@ -148,6 +125,36 @@ static int write_index(git_oid *checksum, git_index *index, git_filebuf *file);
 static void index_entry_free(git_index_entry *entry);
 static void index_entry_reuc_free(git_index_reuc_entry *reuc);
 
+GIT_INLINE(int) index_map_set(git_index *idx, git_idxmap *map, git_index_entry *e)
+{
+	if (!map)
+		map = idx->entries_map;
```
ethomson: I found it a little odd that it took an `idx` and a `map` and did this switch.  Maybe splitting this into two functions?

`int index_map_set(git_idxmap *map, git_index_entry *e, bool ignore_case)`

and

`int index_entry_set(git_index *idx, git_index_entry *e)` which calls into that?  Names provisional, of course.

I have the same feelings about `index_map_resize`...

pks-t: Yeah, I know, I was pondering the same. I guess I'll just go with `index_map_set`, not even introducing a second `index_entry_set`. The additional boilerplate isn't even worth it to skip a single parameter, if you ask me. Will update in a few minutes

PR #5355, opened 2020-01-10
win32: fix relative symlinks pointing into dirs

### LINKED ISSUE #5252: Windows: directory relative symlinks do not work
When cloning a repository with symbolic links, the links will end up with forward slashes.  These files cannot be read using typical APIs like `CreateFileW` (without `FILE_FLAG_OPEN_REPARSE_POINT`) which result in `ERROR_INVALID_NAME`.  This assumes you have `core.symlinks` to true and have the necessary developer mode enabled.

### Reproduction steps
1. Create a repository with a file symbolic link between directories. For example:
   - `mkdir somedir`
   - `echo contents > somedir\foo`
   - `mklink foo somedir\foo`
   - Init a repo, and add these two files
2. Have a libgit2 project which will clone the sample repo.
3. Notice that the resulting files have forward slashes, like this:
   `10/01/2019  06:43 PM    <SYMLINK>      foo [somedir/foo]`
   Also notice that these files cannot be opened using Windows apis (like in Notepad).  Beware that msys tools (like `cat`) seem to be able to access them. I would guess that maybe they use `FILE_FLAG_OPEN_REPARSE_POINT`, read the link content, and translate the path manually.

### Expected behavior

Should be able to open symlinked files on windows.

### Actual behavior

Files fail to open.

08cfa43d0e1a9214a2f1239593686078e75e5636

Windows 10

pks-t:
On Windows platforms, we need some logic to emulate symlink(3P) defined
by POSIX. As unprivileged symlinks on Windows are a rather new feature,
our current implementation is comparatively new and still has some
rough edges in special cases.

One such case is relative symlinks. While relative symlinks to files in
the same directory work as expected, libgit2 currently fails to create
reltaive symlinks pointing into other directories. This is due to the
fact that we forgot to translate the Unix-style target path to
Windows-style. Most importantly, we are currently not converting
directory separators from "/" to "\".

Fix the issue by calling `git_win32_path_canonicalize` on the target.
Add a test that verifies our ability to create such relative links
across directories.

Fixes #5252.

### DIFF
--- a/src/win32/posix_w32.c
+++ b/src/win32/posix_w32.c
@@ -439,8 +439,16 @@ int p_symlink(const char *target, const char *path)
 	git_win32_path target_w, path_w;
 	DWORD dwFlags;
 
+	/*
+	 * Convert both target and path to Windows-style paths. Note that we do
+	 * not want to use `git_win32_path_from_utf8` for converting the target,
+	 * as that function will automatically pre-pend the current working
+	 * directory in case the path is not absolute. As Git will instead use
+	 * relative symlinks, this is not someting we want.
+	 */
 	if (git_win32_path_from_utf8(path_w, path) < 0 ||
-	    git__utf8_to_16(target_w, MAX_PATH, target) < 0)
+	    git__utf8_to_16(target_w, MAX_PATH, target) < 0 ||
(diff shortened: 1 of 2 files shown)


ethomson:
Good one, thanks @pks-t

PR #5359, opened 2020-01-15
Make type mismatch errors consistent

tniessen:
The same error message appears in other places with the word `"the"` included. API consumers in higher-level languages such as JavaScript (via nodegit) might consume these strings, so they should be consistent.


ethomson:
Good eye.  Thanks for the fix.

PR #5360, opened 2020-01-15
refs: refuse to delete HEAD

### LINKED ISSUE #5357: reject attempts to delete HEAD
I just wrote some buggy code in which I accidentally deleted HEAD, rather than its symbolic target. However, HEAD is required to exist.

Given that deleting HEAD is not a legitimate git operation (it turns a git dir into a non-git dit), perhaps libgit2 could prevent this by returning an error when `git_reference_delete` is called on HEAD. I'm not sure what error code/class would be the right one, though.

Sound ok?

josharian:
This requires adding a new symbolic ref to the testrepo fixture.
Some of the existing tests attempt to delete HEAD, expecting a different failure. Introduce and use a non-HEAD symbolic ref instead.
Adjust a few other tests as needed.

Fixes #5357

### DIFF
--- a/src/refs.c
+++ b/src/refs.c
@@ -145,6 +145,11 @@ int git_reference_delete(git_reference *ref)
 	const git_oid *old_id = NULL;
 	const char *old_target = NULL;
 
+	if (!strcmp(ref->name, "HEAD")) {
+		git_error_set(GIT_ERROR_REFERENCE, "cannot delete HEAD");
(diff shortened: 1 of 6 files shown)


ethomson:
LGTM.  Going to hold this open for a bit longer for additional feedback.

pks-t:
Yup, makes sense to me. Thanks a lot, @josharian!

PR #5362, opened 2020-01-17
path: preserve SSH prefixes when resolving relative paths

### LINKED ISSUE #5346: git_submodule_resolve_url get wrong result when repo use ssh url
### Reproduction steps
Open a repo with ssh url like `git@github.com:libgit2/libgit2.git`
use git_submodule_resolve_url with relative url like `../../libuv/libuv.git`
### Expected behavior
get result `git@github.com:libuv/libuv.git`
### Actual behavior
get wrong result `libuv/libuv.git`
commit id 6777db8e8343014d57c41b335e0c60af44d4e9f2
macos 10.15.2

pks-t:
The function `git_path_resolve_relative` is not only used for paths, but
also for URLs. As a result, it needs to detect certain URL prefixes like
for example schemata ("http://") which are not to be removed during any
directory traversals.

We already handle the schemata part correctly, but we are missing the
special case of SSH. In contrast to most other protocols, SSH uses a
colon instead of a slash after the host part, like for example in
"git@example.com:foo". As a result, `git_path_resolve_relative` would've
treated the complete "git@example.com:foo" part as a single directory
part and thus remove it during upwards-traversal.

This problem is relatively easy to hit when having submodules with a
relative URL. As `git_submodule_resolve_url` calls out to
`git_path_resolve_relative`, it mis-treated SSH URLs and thus didn't
compute them correctly.

Fix the issue by special-casing the SSH protocol. Detection makes use
either of the "ssh://" prefix or with a heuristic that tries to detect
whether there's a colon before a slash. Note that especially the second
heuristic is error prone, but it is in fact what git uses right now.

Add some tests to verify behaviour.

---

Fixes #5346

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -784,6 +784,26 @@ int git_path_find_dir(git_buf *dir, const char *path, const char *base)
 	return error;
 }
 
+static int url_prefix(const char *path)
+{
+	const char *p = path, *colon, *slash;
+
+	/* Skip over host part in SSH protocols */
+	if (!git__prefixcmp(path, "ssh://") && (colon = strchr(p + strlen("ssh://"), ':')))
+		return (colon + 1) - path;
+
+	/* Skip over URL schema, if any */
+	for (; *p && git__isalpha(*p); ++p);
+	if (p[0] == ':' && p[1] == '/' && p[2] == '/')
+		return (p + 3) - path;
+
+	/* Same as the SSH case above, but without explicit ssh:// schema. */
+	if ((colon = strchr(p, ':')) && ((slash = strchr(p, '/')) && slash > colon))
+		return (colon + 1) - path;
+
+	return 0;
+}
+
 int git_path_resolve_relative(git_buf *path, size_t ceiling)
 {
 	char *base, *to, *from, *next;
@@ -799,11 +819,8 @@ int git_path_resolve_relative(git_buf *path, size_t ceiling)
 		ceiling = git_path_root(path->ptr) + 1;
 
 	/* recognize URL prefixes that should not be backed over */
-	if (ceiling == 0) {
-		for (next = path->ptr; *next && git__isalpha(*next); ++next);
-		if (next[0] == ':' && next[1] == '/' && next[2] == '/')
-			ceiling = (next + 3) - path->ptr;
-	}
+	if (ceiling == 0)
+		ceiling = url_prefix(path->ptr);
 
 	base = to = from = path->ptr + ceiling;
 
--- a/tests/submodule/resolve.c
+++ b/tests/submodule/resolve.c
@@ -1,0 +1,115 @@
+#include "clar_libgit2.h"
+
+#include "path.h"
+
+static git_repository *g_repo;
+static git_buf g_buf;
+
+void test_submodule_resolve__initialize(void)
+{
+	cl_git_pass(git_buf_init(&g_buf, 0));
+	g_repo = cl_git_sandbox_init("testrepo");
+}
+
+void test_submodule_resolve__cleanup(void)
+{
+	git_buf_dispose(&g_buf);
+	cl_git_sandbox_cleanup();
+}
+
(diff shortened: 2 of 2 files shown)


src/path.c
```diff
@@ -784,6 +784,26 @@ int git_path_find_dir(git_buf *dir, const char *path, const char *base)
 	return error;
 }
 
+static int url_prefix(const char *path)
```
ethomson: I think that this either needs to return `size_t`.

ethomson:
Hmm.  I worry that we don't fully understand the URL structure here.  For example, the following test fails:

```
void test_submodule_resolve__dont_backup_over_root(void)
{
    assert_resolves("https://example.com/", "../child", "https://example.com/child");
}
```

Should it?  Is there a case where we would have a configuration like this?  It seems poorly formed, but in a way that is plausible that an end-user would set up and that git supports.

I also worry that we don't understand DOS paths.  The following tests fail:

```
void test_submodule_resolve__dos_paths_with_absolute(void)
{
    assert_resolves("c:/foo/bar", "/rooted/path", "c:/rooted/path");
}

void test_submodule_resolve__dont_backup_over_dos_root(void)
{
    assert_resolves("c:/", "../sibling", "c:/sibling");
}

void test_submodule_resolve__dont_backup_over_dos_root_without_slash(void)
{
    assert_resolves("c:", "../sibling", "c:/sibling");
}
```

The first one seems like an actual problem.  The latter two fall into the same category of weird things as the http one, above.

I'm a bit surprised that we can't just make use of path and URL canonicalization functions here.

pks-t:
> Hmm. I worry that we don't fully understand the URL structure here. For example, the following test fails:
> I'm a bit surprised that we can't just make use of path and URL canonicalization functions here.

The problem is basically that SSH URLs are not necessarily distinguishable from filesystems paths. Take for example "user@host:repo". It's a perfectly valid filename, but in fact it would usually be an SSH URL. So all you can do is use heuristics to guess whether its a path or URL. But how to do that? You cannot check for "@", as "host:repo" would also be a valid SSH path. So you need to do something with the colon, but again it's hard to properly distinguish those.

I'll see how to fix the test cases you've got. But in the end I agree that this is a mess.

pks-t:
Well.

```
$ git submodule--helper resolve-relative-url-test '(null)' https://example.com/foobar ../../x
https://x
$ git submodule--helper resolve-relative-url-test '(null)' 'c:/foo/bar' '/rooted/path'
c:/foo/bar//rooted/path
$ git submodule--helper resolve-relative-url-test '(null)' c: '../sibling'
c:sibling
$ git submodule--helper resolve-relative-url-test '(null)' c:/ '../sibling'
c:sibling
```

Haha. Well, our logic sucks, but so does upstream's. I'll see if I can get us to suck in the same way. But at least one of your listed testcases does resolve correctly (http://example.com/foobar + "../../x")

pks-t:
Huh. I feel like using our path logic is simply wrong, as it's not about paths in the end but about URLs. `git_net_url` isn't usable either, as it doesn't support SSH-style URLs. So maybe we'll have to invent some custom logic for this? Especially considering that git has such weird rules I don't feel like polluting other subsystems with this weirdness by putting it into centrally available functions. Opinions?

pks-t:
Some more fun examples:

```
void test_submodule_resolve__ssh_removes_prefix(void)
{
	assert_resolves("git@git.example.com:foo/bar", "../x", "git@git.example.com:foo/x");
	assert_resolves("git@git.example.com:foo/bar", "../../x", "git@git.example.com:x");
	assert_resolves("git@git.example.com:foo/bar", "../../../x", ".:x");
}

void test_submodule_resolve__absolute_removes_prefix(void)
{
	assert_resolves("/foo", "x", "/foo/x");
	assert_resolves("/foo", "../x", "/x");
	assert_resolves("/foo", "../../x", "x");
}
```

PR #5365, opened 2020-01-18
Return int from non-free functions

ethomson:
Return `int` from all the functions that aren't `dispose` or `free` functions.  This allows us to gracefully fail on bad inputs (instead of an `assert` or crashing with a segmentation fault).

Only functions that cannot fail will remain void.  This includes:

* `*_dispose` and `*_free`: which should gracefully take `NULL` pointers, ignore them, and continue.
* `giterr_clear`: which cannot fail (since it just updates pointers in global state)
* `giterr_set_oom`: which cannot fail (since it just updates pointers in global state)

### DIFF
--- a/include/git2/attr.h
+++ b/include/git2/attr.h
@@ -238,8 +238,11 @@ GIT_EXTERN(int) git_attr_foreach(
  * disk no longer match the cached contents of memory.  This will cause
  * the attributes files to be reloaded the next time that an attribute
  * access function is called.
+ *
+ * @param repo The repository containing the gitattributes cache
+ * @return 0 on success, or an error code
  */
-GIT_EXTERN(void) git_attr_cache_flush(
+GIT_EXTERN(int) git_attr_cache_flush(
 	git_repository *repo);
 
 /**
--- a/src/errors.c
+++ b/src/errors.c
@@ -95,19 +95,25 @@ void git_error_vset(int error_class, const char *fmt, va_list ap)
 		set_error_from_buffer(error_class);
 }
 
-void git_error_set_str(int error_class, const char *string)
+int git_error_set_str(int error_class, const char *string)
 {
 	git_buf *buf = &GIT_GLOBAL->error_buf;
 
 	assert(string);
 
-	if (!string)
-		return;
+	if (!string) {
+		git_error_set(GIT_ERROR_INVALID, "unspecified caller error");
+		return -1;
+	}
 
 	git_buf_clear(buf);
 	git_buf_puts(buf, string);
-	if (!git_buf_oom(buf))
-		set_error_from_buffer(error_class);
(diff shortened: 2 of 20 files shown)


src/errors.c
```diff
@@ -95,19 +95,23 @@ void git_error_vset(int error_class, const char *fmt, va_list ap)
...
-	if (!string)
-		return;
+	if (!string) {
+		git_error_set(GIT_ERROR_INVALID, "unspecified caller error");
+		return -1;
+	}
 
 	git_buf_clear(buf);
 	git_buf_puts(buf, string);
 	if (!git_buf_oom(buf))
```
siprbaum: Should it return an error then `git_buf_oom` is true?

include/git2/attr.h
```diff
@@ -238,8 +238,11 @@ GIT_EXTERN(int) git_attr_foreach(
  * disk no longer match the cached contents of memory.  This will cause
  * the attributes files to be reloaded the next time that an attribute
  * access function is called.
+ *
+ * @param repo The repository containing the gitattributes cache
+ * @return 0 on success, non-zero callback return value, or error code
```
siprbaum: Copy & Paste error from the git_attr_foreach above.
It should read
```suggestion
 * @return 0 on success or an error code on failure
```

ethomson: Oops!  Thanks.  :)

src/errors.c
ethomson: Yeah, I think that you're right.  I was sort of thinking that callers wouldn't really need to know when there was an OOM situation, as they were trampolining out of this anyway and there's very little that they can do when we've run out of memory.  But I don't really see a reason why we shouldn't.  Updated to return `-1`.

src/index.c
```diff
@@ -539,13 +539,19 @@ int git_index_clear(git_index *index)
...
+	if (error)
+		goto done;
+
 	index_free_deleted(index);
 
-	git_index_reuc_clear(index);
-	git_index_name_clear(index);
+	if ((error = git_index_name_clear(index)) < 0 ||
+	    (error = git_index_reuc_clear(index)) < 0)
+	    goto done;
```
pks-t: Nit: indentation

src/index.c
```diff
@@ -3277,8 +3287,9 @@ static int git_index_read_iterator(
 		}
 	}
 
-	git_index_name_clear(index);
-	git_index_reuc_clear(index);
+	if ((error = git_index_name_clear(index)) < 0 ||
+	    (error = git_index_reuc_clear(index)) < 0)
+	    goto done;
```
pks-t: Here as well

ethomson:
Fixed!  Thanks for the review @pks-t

Issue #5343, opened 2019-12-19
"cred.h" is missing in releas tarball


maxmeyer:
### Reproduction steps

1. Download release tarball https://github.com/libgit2/libgit2/archive/v0.28.4.tar.gz
2. Browse archive and look for "cred.h"

### Expected behavior

"cred.h" is part of tarbal, as this is required to build "git2go".

### Actual behavior

"cred.h" is missing in release tarball

### Version of libgit2 (release number or SHA1)

0.28.4

### Operating system(s) tested

Arch Linux (https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/libgit2)

ethomson:
Weeeeeeeeird.  These are created automatically -- @carlosmn do you know how these are generated or know who does?

maxmeyer:
Any progress on this?

pks-t:
Well, "cred.h" was renamed to "credentials.h" now. So maybe the bug is fixed by this :P

Anyway, I feel like we should start assembling our own archives. Maybe we should create a script that takes the new version as argument and then:

- verifies that we've incremented the version correctly in all the required places
- verifies release notes from our CHANGELOG.md
- creates the archives
- creates a GitHub release using the archive and release notes extracted from CHANGELOG.md

This would implicitly document our release process, ensure consistency and makes sure that nothing is forgotten. I'll have a look at the GitHub API and try to come up with something that can then also be used for our upcoming v0.99 release.

ethomson:
> Anyway, I feel like we should start assembling our own archives. 

I'm not opposed to this but I'm disappointed that there's no way to remove the automatically generated ones.  It feels like if we start assembling our own then it won't be obvious to people which one to use.

I do happen to know someone who works at GitHub, though, so I can inquire about possible solutions.

pks-t:
> I'm not opposed to this but I'm disappointed that there's no way to remove the automatically generated ones. It feels like if we start assembling our own then it won't be obvious to people which one to use.

Definitely true. I've created #5372, which introduces such a new release script, but I've noted the same confusion that results from having multiple archives. We could probably counteract that by choosing proper labels. "Download this here, it's the real thing"

ethomson:
Closing this as stale.

PR #5372, opened 2020-01-30
Release script

pks-t:
The current release process is not documented in any way. As a result,
it's not obvious how releases should be done at all, like e.g. which
locations need adjusting.

To fix this, let's introduce a new script that shall from now on be used
to do all releases. As input it gets the tree that shall be released,
the repository in which to do the release, credentials to
authenticate against GitHub and the new version. E.g. executing the
following will create a new release v0.32:

    $ ./script/release.py 0.32.0 --user pks-t --password ****

While the password may currently be your usual GitLab password, it's
recommended to use a personal access token intead.

The script will then perform the following steps:

    1. Verify that "include/git2/version.h" matches the new version.

    2. Verify that "docs/changelog.md" has a section for that new
       version.

    3. Extract the changelog entries for the current release from
       "docs/changelog.md".

    4. Generate two archives in "tar.gz" and "zip" format via "git
       archive" from the tree passed by the user. If no tree was passed,
       we will use "HEAD".

    5. Create the GitHub release using the extracted changelog entries
       as well as tag and name information derived from the version
       passed by the used.

    6. Upload both code archives to that release.

This should cover all steps required for a new release and thus ensures
that nothing is missing that shouldn't be.

---

Note that it seems to be impossible to turn off GitHub's auto-generated archives, so I find it questionable whether it makes sense to upload our own code archives. We can control them better though, which would allow us to fix e.g. #5343. On the other hand, users might be confused when there's seemingly multiple downloads for the same thing.

We could probably counteract that by giving proper names. Right now they're the same as the default generated ones. See https://github.com/pks-t/test/releases for some examples of releases done via that script.

Anyway, even if we end up not generating our own tarballs I feel like having this script is a good step.

### DIFF
--- a/script/release.py
+++ b/script/release.py
@@ -1,0 +1,171 @@
+#!/usr/bin/env python
+
+from collections import namedtuple
+
+import argparse
+import base64
+import copy
+import json
+import subprocess
+import sys
+import urllib.parse
+import urllib.request
+import urllib.error
+
+class Error(Exception):
+    pass
+
+class Version(object):
+    def __init__(self, version):
+        versions = version.split(sep='.')
+        if len(versions) < 2 or len(versions) > 3:
+            raise Error("Invalid version string '{}'".format(version))
+        self.major = int(versions[0])
+        self.minor = int(versions[1])
+        self.revision = int(versions[2]) if len(versions) == 3 else 0
+
+    def __str__(self):
+        return '{}.{}.{}'.format(self.major, self.minor, self.revision)
+
+    def __eq__(self, other):
+        return self.major == other.major and self.minor == other.minor and self.revision == other.revision
+
+def verify_version(version):
+    expected = {
+        'VERSION':      [ '"{}"'.format(version), None ],
+        'VER_MAJOR':    [ str(version.major), None ],
+        'VER_MINOR':    [ str(version.minor), None ],
+        'VER_REVISION': [ str(version.revision), None ],
+        'VER_PATCH':    [ '0', None ],
+        'SOVERSION':    [ '"{}.{}"'.format(version.major, version.minor), None ],
+    }
+
+    with open('include/git2/version.h') as f:
+        lines = f.readlines()
+
+    for key in expected.keys():
+        define = '#define LIBGIT2_{} '.format(key)
+        for line in lines:
+            if line.startswith(define):
+                expected[key][1] = line[len(define):].strip()
+                break
+        else:
+            raise Error("version.h: missing define for '{}'".format(key))
+
+    for k, v in expected.items():
+        if v[0] != v[1]:
+            raise Error("version.h: define '{}' does not match (got '{}', expected '{}')".format(k, v[0], v[1]))
+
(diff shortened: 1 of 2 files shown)


ethomson:
This looks good to me overall.  I wonder if we might want to change the name of the release that we upload, away from "Source code" to disambiguate it from the GitHub-generated ones...

Maybe "Release source: libgit2-x.y.z.tar.gz"?  Or "Official release source code: ..." or something like that?

script/release.py
```diff
@@ -0,0 +1,175 @@
...
+        return '{}.{}'.format(self.major, self.minor)
+
+def verify_version(version):
+    expected = {
+        'VERSION':      [ '"{}"'.format(version), None ],
+        'VER_MAJOR':    [ str(version.major), None ],
+        'VER_MINOR':    [ str(version.minor), None ],
+        'VER_REVISION': [ str(version.revision), None ],
+        'VER_PATCH':    [ '0', None ],
+        'SOVERSION':    [ str(version.minor), None ],
```
ethomson: What should our strategy for soversion here be in a post-1.0 world?  We have a couple of choices - we could just keep it monotonically increasing and it doesn't have to have anything to do with the actual version numbers.  But I don't actually love that.

We could track it against the major/minor version numbers, which is consistent with what we've been doing - eg 0.29 -> 29, 1.0 -> 100, 1.1 -> 101.  This is the strategy that I'd advocate for.

pks-t:
> Maybe "Release source: libgit2-x.y.z.tar.gz"? Or "Official release source code: ..." or something like that?

Everything's better than what I currently have, and I think "Release source: libgit2-x.y.z.tar.gz" is good. Wouldn't it need to be "Release sources", though (plural)?

> We could track it against the major/minor version numbers, which is consistent with what we've been doing - eg 0.29 -> 29, 1.0 -> 100, 1.1 -> 101. This is the strategy that I'd advocate for.

I see other projects use something like "libgiblib.so.1.0.6", "libgio-2.0.so.0.6000.7", "libgirara-gtk3.so.3.1" etc. So we could change our pattern to "libgit2.so.$MAJOR.$MINOR" with a SOVERSION of "$MAJOR.$MINOR"?

pks-t:
> I see other projects use something like "libgiblib.so.1.0.6", "libgio-2.0.so.0.6000.7", "libgirara-gtk3.so.3.1" etc. So we could change our pattern to "libgit2.so.$MAJOR.$MINOR" with a SOVERSION of "$MAJOR.$MINOR"?

Or go all the way to "libgit2.so.$MAJOR.$MINOR.$REVISION", where systems provide the typical symlinks from "libgit2.so.$MAJOR" -> "ligit2.so.$MAJOR.$MINOR" -> "libgit2.so.$MAJOR.$MINOR.$REVISION" so applications can choose how specific to be.

ethomson:
Yeah, that's a good idea.  I updated the soversion to major.minor in https://github.com/libgit2/libgit2/pull/5291/commits/2827c8a38917d9fe95d22d301c7118ae3113a665

PR #5374, opened 2020-01-31
tests: diff: verify that we are able to diff with empty subtrees

### LINKED ISSUE #3933: diff fails on empty directory
When a tree object contains a subtree that is empty, libgit2-diff does not work any more correctly.
In the example gist at https://gist.github.com/mrh1997/a7f23599231fef236ec85bf48868165d an erornous repo is created (with a single commit with one file ('datafile') and an empty directory ('EmptyDir')).
when diffing this file with git you git 'datafile', when doing the diff with pygit2/libgit2 you get nothing
=> It should show an error OR run the same behaviour that git...

Addmittedly this is a non-usual use-case as empty trees should not be created,
libgit2 provides the tools to do so and you get no feedback if your software uses these tools in an invalid manner (and creates empty dirs).

In my case I needed two days to track down the problem due to the missing error code....

pks-t:
While it is not allowed for a tree to have an empty tree as child (e.g.
an empty directory), libgit2's tree builder makes it easy to create such
trees. As a result, some applications may inadvertently end up with such
an invalid tree, and we should try our best and handle them.

One such case is when diffing two trees, where one of both trees has
such an empty subtree. It was reported that this will cause our diff
code to fail. While I wasn't able to reproduce this error, let's still
add a test that verifies we continue to handle them correctly.

---

Fixes #3933 

The test is a near 1:1 translation of the Gist provided in that issue and I've added some more checks to verify that our behaviour is correct. I'd thus think that the issue has been fixed since.

### DIFF
--- a/tests/diff/tree.c
+++ b/tests/diff/tree.c
@@ -524,3 +524,52 @@ void test_diff_tree__diff_configs(void)
 	cl_assert_equal_i(7, expect.line_adds);
 	cl_assert_equal_i(15, expect.line_dels);
 }
+
+void test_diff_tree__diff_tree_with_empty_dir_entry_succeeds(void)
+{
+	const char *content = "This is a blob\n";
+	const git_diff_delta *delta;
+	git_oid empty_tree, invalid_tree, blob;
+	git_buf patch = GIT_BUF_INIT;
+	git_treebuilder *builder;
+
+	g_repo = cl_git_sandbox_init("empty_standard_repo");
+
+	cl_git_pass(git_blob_create_from_buffer(&blob, g_repo, content, strlen(content)));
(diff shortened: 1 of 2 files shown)


ethomson:
Seems reasonable.  Since we can't uncover a bug here and this is just additional test coverage (which I'm grateful for) let's land this after 0.99.

pks-t:
> Seems reasonable. Since we can't uncover a bug here and this is just additional test coverage (which I'm grateful for) let's land this after 0.99.

Want to do another review or shall I just go ahead and merge?

ethomson:
> Want to do another review or shall I just go ahead and merge?

:shipit:

PR #5375, opened 2020-01-31
azure-pipelines: properly expand negotiate passwords

pks-t:
To allow testing against a Kerberos instance, we have added variables
for the Kerberos password to allow authentication against LIBGIT2.ORG in
commit e5fb5fe5a (ci: perform SPNEGO tests, 2019-10-20). To set up the
password, we assign

    "GITTEST_NEGOTIATE_PASSWORD=$(GITTEST_NEGOTIATE_PASSWORD)"

in the environmentVariables section which is then passed through to a
template. As the template does build-time expansion of the environment
variables, it will expand the above line verbosely, and due to the
envVar section not doing any further expansion the password variable
will end up with the value "$(GITTEST_NEGOTIATE_PASSWORD)" in the
container's environment.

Fix this fixed by doing expansion of GITTEST_NEGOTIATE_PASSWORD at
build-time, as well.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,7 +20,7 @@ jobs:
        CC=gcc
        CMAKE_GENERATOR=Ninja
        CMAKE_OPTIONS=-DUSE_HTTPS=OpenSSL -DREGEX_BACKEND=builtin -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DVALGRIND=on -DUSE_GSSAPI=ON
-       GITTEST_NEGOTIATE_PASSWORD=$(GITTEST_NEGOTIATE_PASSWORD)
+       GITTEST_NEGOTIATE_PASSWORD=${{ variables.GITTEST_NEGOTIATE_PASSWORD }}
 
 - job: linux_amd64_xenial_gcc_mbedtls
   displayName: 'Linux (amd64; Xenial; GCC; mbedTLS)'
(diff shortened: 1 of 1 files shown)


ethomson:
The reason that this is failing is because you're opening a pull request from a fork.  Azure Pipelines doesn't provide secrets to forks, to prevent untrusted users from exfiltrating secrets.  (The idea being that people with write access to the repository are trusted - which in our case is a pretty reasonable assumption).

I think that we should just disable the negotiate tests for now, since everybody's PRs will fail unless they're opening them from the main repository.  Long term we should test whether the build is coming from a fork or the main repository, and only run the negotiate tests on the latter.  But we can follow up with that in a different PR.

pks-t:
> I think that we should just disable the negotiate tests for now, since everybody's PRs will fail unless they're opening them from the main repository. Long term we should test whether the build is coming from a fork or the main repository, and only run the negotiate tests on the latter. But we can follow up with that in a different PR.

Well, yeah. The root cause though is that variable expansion in azure-pipelines.yml doesn't work as we think it does. Because of that, GITTEST_NEGOTIATE_PASSWORD always equals "$(GITTEST_NEGOTIATE_PASSWORD)" (yup, not expanded), which my PR should fix now. If we use the "correct" variable expansion, then we automatically disable the SPNEGO tests as the variable is empty, which we check for.

pks-t:
Green now. I've updated title and description.

ethomson:
Now that I'm on my computer, I see what you're saying.

PR #5378, opened 2020-02-02
Honor GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH for all checkout types 

ethomson:
During checkout with GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH, we were previously applying the pathspec filter for the baseline iterator during checkout, as well as the target **tree**.  (We were not applying the pathspecs when the target was not a tree.)

This was an oversight; in fact, we should apply the pathspec filter to _all_ checkout targets, not just trees.

Add a helper function to set the iterator pathspecs from the given checkout pathspecs, and call it everywhere.

Includes the test provided by https://github.com/libgit2/libgit2/commit/f37c77b4186714b0a973cab1c0a9dcf8efd19fae with some minor modifications.

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -2544,6 +2544,17 @@ cleanup:
 #define CHECKOUT_INDEX_DONT_WRITE_MASK \
 	(GIT_CHECKOUT_DONT_UPDATE_INDEX | GIT_CHECKOUT_DONT_WRITE_INDEX)
 
+GIT_INLINE(void) setup_pathspecs(
+	git_iterator_options *iter_opts,
+	const git_checkout_options *checkout_opts)
+{
+	if (checkout_opts &&
+		(checkout_opts->checkout_strategy & GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH)) {
+		iter_opts->pathlist.count = checkout_opts->paths.count;
+		iter_opts->pathlist.strings = checkout_opts->paths.strings;
+	}
+}
+
 int git_checkout_iterator(
 	git_iterator *target,
 	git_index *index,
@@ -2586,6 +2597,8 @@ int git_checkout_iterator(
 	workdir_opts.start = data.pfx;
 	workdir_opts.end = data.pfx;
 
+	setup_pathspecs(&workdir_opts, opts);
+
 	if ((error = git_iterator_reset_range(target, data.pfx, data.pfx)) < 0 ||
 		(error = git_iterator_for_workdir_ext(
 			&workdir, data.repo, data.opts.target_directory, index, NULL,
@@ -2596,10 +2609,8 @@ int git_checkout_iterator(
 		GIT_ITERATOR_IGNORE_CASE : GIT_ITERATOR_DONT_IGNORE_CASE;
 	baseline_opts.start = data.pfx;
 	baseline_opts.end = data.pfx;
-	if (opts && (opts->checkout_strategy & GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH)) {
-		baseline_opts.pathlist.count = opts->paths.count;
-		baseline_opts.pathlist.strings = opts->paths.strings;
-	}
+
+	setup_pathspecs(&baseline_opts, opts);
 
 	if (data.opts.baseline_index) {
 		if ((error = git_iterator_for_index(
@@ -2689,6 +2700,7 @@ int git_checkout_index(
 	git_index *index,
 	const git_checkout_options *opts)
 {
+	git_iterator_options iter_opts = GIT_ITERATOR_OPTIONS_INIT;
 	int error, owned = 0;
 	git_iterator *index_i;
 
@@ -2716,7 +2728,9 @@ int git_checkout_index(
 		return error;
 	GIT_REFCOUNT_INC(index);
 
-	if (!(error = git_iterator_for_index(&index_i, repo, index, NULL)))
+	setup_pathspecs(&iter_opts, opts);
+
+	if (!(error = git_iterator_for_index(&index_i, repo, index, &iter_opts)))
 		error = git_checkout_iterator(index_i, index, opts);
 
 	if (owned)
@@ -2773,10 +2787,7 @@ int git_checkout_tree(
 	if ((error = git_repository_index(&index, repo)) < 0)
 		return error;
 
-	if (opts && (opts->checkout_strategy & GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH)) {
-		iter_opts.pathlist.count = opts->paths.count;
-		iter_opts.pathlist.strings = opts->paths.strings;
-	}
+	setup_pathspecs(&iter_opts, opts);
 
 	if (!(error = git_iterator_for_tree(&tree_i, tree, &iter_opts)))
 		error = git_checkout_iterator(tree_i, index, opts);
--- a/tests/checkout/index.c
+++ b/tests/checkout/index.c
@@ -89,6 +89,65 @@ void test_checkout_index__can_remove_untracked_files(void)
 	cl_assert_equal_i(false, git_path_isdir("./testrepo/dir"));
 }
 
+void test_checkout_index__can_disable_pathspec_match(void)
+{
+	static git_index *index;
+	git_oid commit_id;
+	git_checkout_options g_opts = GIT_CHECKOUT_OPTIONS_INIT;
+	git_object *g_object;
+
+	char *files_to_checkout[] = { "test11.txt", "test12.txt"};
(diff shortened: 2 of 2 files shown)


src/checkout.c
```diff
@@ -2544,6 +2544,17 @@ static int checkout_data_init(
 #define CHECKOUT_INDEX_DONT_WRITE_MASK \
 	(GIT_CHECKOUT_DONT_UPDATE_INDEX | GIT_CHECKOUT_DONT_WRITE_INDEX)
 
+GIT_INLINE(void) setup_pathspecs(
+	git_iterator_options *iter_opts,
+	const git_checkout_options *checkout_opts)
+{
+	if (checkout_opts &&
+		(checkout_opts->checkout_strategy & GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH)) {
```
pks-t: Nit: indentation. Continuation lines start with the opening brace :)

pks-t:
Does this ever work? :(

tests/checkout/index.c
```diff
@@ -89,6 +89,65 @@ void test_checkout_index__can_remove_untracked_files(void)
...
+		GIT_CHECKOUT_FORCE |
+		GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH;
+	g_opts.paths.strings = files_to_checkout;
+	g_opts.paths.count = files_to_checkout_size;
+	cl_git_pass(git_checkout_index(g_repo, NULL, &g_opts));
+
+	/* The only files that have been reverted to their original content
+	   should be test10.txt and test11.txt */
+	check_file_contents("testrepo/test9.txt", "modified\n");
+	check_file_contents("testrepo/test10.txt", "original\n");
```
pks-t: macOS fails here:

    1:   1) Failure:
    1: checkout::index::can_disable_pathspec_match [/Users/runner/runners/2.164.7/work/1/s/tests/checkout/index.c:146]
    1:   testrepo/test10.txt
    1:   file content mismatch at byte 0

ethomson:
> Does this ever work? :(

No, I'm not sure what happened but something changed along the way.  😡

pks-t:
> No, I'm not sure what happened but something changed along the way.

Like CI systems always magically tend to over time.

tests/checkout/index.c
```diff
@@ -89,6 +89,65 @@ void test_checkout_index__can_remove_untracked_files(void)
...
+	git_object_free(g_object);
+	g_object = NULL;
+
+	cl_git_pass(git_repository_index(&index, g_repo));
+
+	/* We create 4 files and commit them */
+	cl_git_mkfile("testrepo/test9.txt", "original\n");
+	cl_git_mkfile("testrepo/test10.txt", "original\n");
+	cl_git_mkfile("testrepo/test11.txt", "original\n");
+	cl_git_mkfile("testrepo/test12.txt", "original\n");
```
lucasderraugh: Is it perhaps just missing the `./testrepo...`. It seems like all the other tests follow that format except for this one.

tests/checkout/index.c
ethomson: That's really weird; it works fine locally.  🤔

src/checkout.c
ethomson: Show me what you think our continuation standard looks like.  😛 

Are you saying:

```
if (checkout_opts &&
   (checkout_opts->...
```

pks-t:
On Mon, Feb 10, 2020 at 10:01:19AM -0800, Edward Thomson wrote:
> > @@ -2544,6 +2544,17 @@ static int checkout_data_init(
>  #define CHECKOUT_INDEX_DONT_WRITE_MASK \
>  	(GIT_CHECKOUT_DONT_UPDATE_INDEX | GIT_CHECKOUT_DONT_WRITE_INDEX)
>  
> +GIT_INLINE(void) setup_pathspecs(
> +	git_iterator_options *iter_opts,
> +	const git_checkout_options *checkout_opts)
> +{
> +	if (checkout_opts &&
> +		(checkout_opts->checkout_strategy & GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH)) {
> 
> Show me what you think our continuation standard looks like.  😛 
> 
> Are you saying:
> 
> ```
> if (checkout_opts &&
>    (checkout_opts->...
> ```

I feel like we have that discussion every few months :P Indentation
should be like:

```
if (foo &&
    bar)
	code;
```

or

```
if ((foo && bar &&
     baz && qux) ||
    test)
	code;
```

Anway. We need a style guide. I'll see if I find some time to dig around
in linux.git and git.git and create one for us.

ethomson:
> I feel like we have that discussion every few months :P Indentation should be like:

OK, this was my understanding, too.  But I'm reasonably sure that you've given me feedback about indentation where I followed that pattern - so if this happens again then we'll need to follow up to understand what's going on.

pks-t:
On Tue, Feb 11, 2020 at 02:02:25AM -0800, Edward Thomson wrote:
> > I feel like we have that discussion every few months :P Indentation should be like:
> 
> OK, this was my understanding, too.  But I'm reasonably sure that
> you've given me feedback about indentation where I followed that
> pattern - so if this happens again then we'll need to follow up to
> understand what's going on.

Huh, fair enough. Good to know we're on the same page, though.

ethomson:
> Huh, fair enough. Good to know we're on the same page, though.

Possible that I was mistaken and my editor was doing something funny behind my back.  I've been trying to use VS Code and the experience has overall been good, but not perfect.

lucasderraugh:
I know it seems that we can't figure out why this test fails on a particular environment, but if there is anything I can do / look into to get this moved along I'd love to help.

ethomson:
Can you reproduce the test failures?

lucasderraugh:
It passes the entire suite when I run from my MacBook Pro running Catalina 10.15.3 with APFS. Could we try running the tests again. I don't remember what configuration was failing before and the history is gone from CI.

ethomson:
I rebased this on master, which kicked off a new CI run.  However, it's now passing.  I suspect that this is because we recently had to move from macOS-10.13 to macOS-10.15 due to [Azure Pipelines deprecating the old OS](https://devblogs.microsoft.com/devops/removing-older-images-in-azure-pipelines-hosted-pools/).

So.  That gives us a place to start looking.  Not sure if the filesystems were different on macOS 10.13 or not.  I don't remember if those images were being built with APFS or HFS+.

lucasderraugh:
Hmm, I'd have to find a 10.13 image to go back to HFS+. I can try and figure out how to test in that environment again.

lucasderraugh:
So I'm trying it out on a 10.12 machine with the old file system and looking at the repository it indeed does not revert the individual files. I'm not sure what would cause that to behave differently, but checking the state right after triggering `cl_git_pass(git_checkout_index(g_repo, NULL, &g_opts));` still shows every file as "modified".

tests/checkout/index.c
```diff
@@ -89,6 +89,65 @@ void test_checkout_index__can_remove_untracked_files(void)
...
+	cl_git_pass(git_index_write(index));
+
+	cl_repo_commit_from_index(&commit_id, g_repo, NULL, 0, "commit our test files");
+
+	/* We modify the content of all 4 of our files */
+	cl_git_rewritefile("testrepo/test9.txt", "modified\n");
+	cl_git_rewritefile("testrepo/test10.txt", "modified\n");
+	cl_git_rewritefile("testrepo/test11.txt", "modified\n");
+	cl_git_rewritefile("testrepo/test12.txt", "modified\n");
+
```
lucasderraugh: So testing this out on 10.12, this test always fails. However, inserting the following right after this line fixes the test:
```
cl_git_pass(git_index_write(index));
```
Inspecting the repository right after the checkout does confirm that the 2 files get reverted back to the "original" text. I'm not entirely sure what that means or why it works fine under newer operating systems. @ethomson anything else I should test on this or are we ok with merging this in as it is? From what I can tell this fault in the older OS isn't related to the `GIT_CHECKOUT_DISABLE_PATHSPEC_MATCH` flag and the same issue occurs if you simply try to force checkout the entire index without the flag or files specified.

ethomson:
> @ethomson anything else I should test on this or are we ok with merging this in as it is?

No, we're code-frozen for the 1.0 release this week.  But also, I definitely want to understand why this isn't behaving the way that we expect.  It sounds like you've analyzed it to the point where it seems unrelated to this change, but we still need to understand what's wrong here.  Probably some assumptions in the tests themselves that we should be able to correct relatively easily.

I can take a look post-1.0.

Thanks for the analysis!

ethomson:
Rebased on top of master.

PR #5381, opened 2020-02-04
azure: tests: re-run flaky proxy tests

pks-t:
The proxy tests regularly fail in our CI environment. Unfortunately,
this is expected due to the network layer. Thus, let's re-try the proxy
tests up to five times in case they fail.


pks-t:
Added another fix. Reruns wouldn't count as successful due to us not resetting FAILED. Hopefully this gets our CI stable again, it's been failing a lot recently. /cc @ethomson

PR #5384, opened 2020-02-04
winhttp: support optional client cert

ianhattendorf:
Some reverse proxies/applications can be configured to require client certificates (mutual TLS authentication). It's also possible that the server requests a client certificate, but doesn't require it (allowing anonymous access). For example, an HAProxy frontend can be configured as so: `  bind *:8443 ssl crt /etc/ssl/certs/server.pem ca-file /etc/ssl/certs/ca.crt verify required`. `verify required` indicates that the server requires a valid client certificate to continue. Changing `verify required` to `verify optional` causes the server to request a client certificate, but allow anonymous access as well. In either case, `WinHttpSendRequest` will return `ERROR_WINHTTP_CLIENT_AUTH_CERT_NEEDED`. If this is returned, the request can be retried after setting the option `WINHTTP_OPTION_CLIENT_CERT_CONTEXT` to either `WINHTTP_NO_CLIENT_CERT_CONTEXT` (in which case we are requesting anonymous access) or by attaching a valid client certificate context.

Originally reported in: #4204
Related PR: #4787

### DIFF
--- a/src/transports/winhttp.c
+++ b/src/transports/winhttp.c
@@ -864,31 +864,28 @@ static int do_send_request(winhttp_stream *s, size_t len, bool chunked)
 
 static int send_request(winhttp_stream *s, size_t len, bool chunked)
 {
-	int request_failed = 1, cert_valid, client_cert_requested, error, attempts = 0;
+	int request_failed = 1, error, attempts = 0;
 	DWORD ignore_flags, send_request_error;
 
 	git_error_clear();
 
 	while (request_failed && attempts++ < 3) {
+		int cert_valid = 1;
+		int client_cert_requested = 0;
 		request_failed = 0;
-		cert_valid = 1;
-		client_cert_requested = 0;
 		if ((error = do_send_request(s, len, chunked)) < 0) {
 			send_request_error = GetLastError();
 			request_failed = 1;
 			switch (send_request_error) {
-				case ERROR_WINHTTP_SECURE_FAILURE: {
+				case ERROR_WINHTTP_SECURE_FAILURE:
 					cert_valid = 0;
 					break;
-				}
-				case ERROR_WINHTTP_CLIENT_AUTH_CERT_NEEDED: {
+				case ERROR_WINHTTP_CLIENT_AUTH_CERT_NEEDED:
 					client_cert_requested = 1;
 					break;
-				}
-				default: {
+				default:
 					git_error_set(GIT_ERROR_OS, "failed to send request");
 					return -1;
-				}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


src/transports/winhttp.c
```diff
@@ -864,42 +864,68 @@ static int do_send_request(winhttp_stream *s, size_t len, bool chunked)
...
+		if (client_cert_requested) {
+			/*
+			 * Client certificates are not supported, explicitly tell the server that
+			 * (it's possible a client certificate was requested but is not required)
+			 */
+			if (!WinHttpSetOption(s->request, WINHTTP_OPTION_CLIENT_CERT_CONTEXT, WINHTTP_NO_CLIENT_CERT_CONTEXT, 0)) {
+				git_error_set(GIT_ERROR_OS, "failed to set client cert context");
+				return -1;
+			}
+		}
```
pks-t: So we don't start supporting client certs, but simply start telling servers we ain't got one. I guess this is a good fix until we might land client certificates some time in the future and matches what Microsoft documents for this API.

pks-t:
Forgot to say: thanks for your effort!

src/transports/winhttp.c
ianhattendorf: Yep, to support client certs we would need to provide a callback for users to specify which client cert they would like to use. We can do this by querying `WINHTTP_OPTION_CLIENT_CERT_ISSUER_LIST` to get a list of valid client certs for the provided CA, providing this list in the callback, and letting the user select one. We could also provide a default callback that just picks the 1st valid cert. I'm able to successfully authenticate using this method, I would just need some free time to implement the callback and polish/test it so I'd like to handle that in a follow up.

ianhattendorf:
>
>
> Forgot to say: thanks for your effort!

Thanks for getting to this so quickly!

pks-t:
On Thu, Feb 06, 2020 at 08:25:02AM -0800, Ian Hattendorf wrote:
> > Forgot to say: thanks for your effort!
> 
> Thanks for getting to this so quickly!

Note that we're currently in code freeze until getting v1.0 out, so only
bugfixes get in. Will merge as soon as that's out.

ianhattendorf:
@pks-t just wanted to ping you to make sure this doesn't get forgotten since it looks like v1.0 is out now.

src/transports/winhttp.c
```diff
@@ -864,42 +864,65 @@ static int do_send_request(winhttp_stream *s, size_t len, bool chunked)
...
+				default:
+					git_error_set(GIT_ERROR_OS, "failed to send request");
+					return -1;
+			}
 		}
 
-		request_failed = 1;
-		cert_valid = 0;
-	}
+		if (!request_failed || send_request_error == ERROR_WINHTTP_SECURE_FAILURE) {
```
pks-t, line 892: I think this could be `if (!request_failed || !cert_valid)`, which would make it code clearer to me.

ethomson:
Thanks @ianhattendorf - sorry for the delay here, I'm trying to make my way through the backlog.  LGTM.

Issue #5385, opened 2020-02-06
recent regression adding http headers to push/fetch


josharian:
### Reproduction steps

I use a bearer token in an Authorization header to authorize with my server.

Using git2go:

```go
opts := new(git.PushOptions)
opts.Headers = []string{"Authorization: Bearer " + bearer}
err := remote.Push(refspecs, opts)
```

This worked until recently. At current master (559751714b23f63ead726e9a773be8b07e3551ae), there is no Authorization header present when the request reaches the server.

I have bisected to one of these three commits:

* 7372573b5f1113b8522e2588fac1c529ddcedb0a
* bf55facf15e5b6897ff7305162a77d92089f6c82
* b9c5b15a7958ab4ecb83504a6d858efd18610297

The first two of these commits don't compile: `src/transports/http.c:89:2: error: unknown type name 'git_http_authtype_t'`. The last one seems pretty plausible.

cc @ethomson

josharian:
Just to make sure, I threaded `GIT_OPT_ENABLE_HTTP_EXPECT_CONTINUE` through to git2go and enabled it; it did not fix the problem.

pks-t:
Thanks for your report! I bet it's commit b9c5b15a7958ab4ecb83504a6d858efd18610297, which refactors basically everything related to HTTP.

pks-t:
Could you please verify that #5387 fixes your issue?

josharian:
#5387 fixes it, thanks!

PR #5386, opened 2020-02-06
mwindow: set limit on number of open files

### LINKED ISSUE #2758: GC mwindows on too many open files
We currently attempt to free unused map windows for packfiles when we go over a threshold. We however do not account for the number of open packfiles.

There's been at least two distinct reports of people running into issues when they have repositories which do network operations in small batches. These accumulate a lot of tiny packfiles and we end up having them all open at once, which can reach the system's limit for open file descriptors.

We should have a configurable limit on how many open file descriptors we want to have for packfiles and do a soft-GC similar to what we do for memory (address space).

josharian:
Doing many network operations can lead to lots of small packfiles.
The existing mwindow limit was only the total size of mmap'd files,
not on their number, so having lots of small packfiles could exhaust
the allowed number of open files, particularly on macOS, where
the default ulimit is very low.

Add a limit to the number of files as well as their total size,
and set the default limit low enough that even macOS users should not
hit it during normal use.

Fixes #2758

### DIFF
--- a/src/mwindow.c
+++ b/src/mwindow.c
@@ -22,8 +22,12 @@
 #define DEFAULT_MAPPED_LIMIT \
 	((1024 * 1024) * (sizeof(void*) >= 8 ? 8192ULL : 256UL))
 
+/* default ulimit -n on macOS is just 256 */
+#define DEFAULT_OPEN_LIMIT 128
+
 size_t git_mwindow__window_size = DEFAULT_WINDOW_SIZE;
 size_t git_mwindow__mapped_limit = DEFAULT_MAPPED_LIMIT;
+unsigned int git_mwindow__open_limit = DEFAULT_OPEN_LIMIT;
 
 /* Whenever you want to read or modify this, grab git__mwindow_mutex */
 static git_mwindow_ctl mem_ctl;
@@ -268,7 +272,8 @@ static git_mwindow *new_window(
 
 	ctl->mapped += (size_t)len;
 
-	while (git_mwindow__mapped_limit < ctl->mapped &&
+	while ((git_mwindow__mapped_limit < ctl->mapped ||
+	       git_mwindow__open_limit < ctl->open_windows) &&
 			git_mwindow_close_lru(mwf) == 0) /* nop */;
 
 	/*
--- a/include/git2/common.h
+++ b/include/git2/common.h
@@ -205,7 +205,9 @@ typedef enum {
 	GIT_OPT_GET_PACK_MAX_OBJECTS,
 	GIT_OPT_SET_PACK_MAX_OBJECTS,
 	GIT_OPT_DISABLE_PACK_KEEP_FILE_CHECKS,
-	GIT_OPT_ENABLE_HTTP_EXPECT_CONTINUE
+	GIT_OPT_ENABLE_HTTP_EXPECT_CONTINUE,
+	GIT_OPT_GET_MWINDOW_OPEN_LIMIT,
+	GIT_OPT_SET_MWINDOW_OPEN_LIMIT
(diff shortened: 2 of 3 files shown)


josharian:
I'm not sure how to write a test for this.

Doing some manual testing (applied to master a little ways back, due to https://github.com/libgit2/libgit2/issues/5385), this helps with the situation in #2758, but not quite in the way I expected/hoped.

I still see O(#packfiles) open files using `lsof` with this patch applied. The difference this patch makes is that when the number of packfiles shrinks, so does the number of open files.

I don't understand enough about what's going on in the rest of the codebase that causes this combination of effects. I'd prefer that it strictly limited the number of open files. Hints and suggestions welcome!

For my purposes, this patch is at least enough to ameliorate the worst of #2758, because I can at least keep things under control by occasionally running `git gc`.

lhchavez:
IIRC one way to reproduce this issue is to create >$OPEN_LIMIT packfiles where each packfile has a single commit object that points to a commit in another packfile. that way you never run into degenerate cases like needing a superlong chain of packfiles with deltas smeared among them (IIRC there is yet another codepath that limits the number of delta chains).

vanilla git and IIUC libgit2 (with this change applied) should both be able to handle that case correctly.

src/mwindow.c
```diff
@@ -268,7 +272,8 @@ static git_mwindow *new_window(
 
 	ctl->mapped += (size_t)len;
 
-	while (git_mwindow__mapped_limit < ctl->mapped &&
+	while ((git_mwindow__mapped_limit < ctl->mapped ||
+	       git_mwindow__open_limit < ctl->open_windows) &&
```
pks-t: Hum. I think this doesn't exactly work the way we need it to work.Thing is we track mwindows via a list ordered by recency of last usage. If we just call `git_mwindow_close_lru` until using less than `open_limit` files, then we will most likely close mwindows of other files, which doesn't help us at all in that case. Instead, we'd need to have some kind of heuristic that checks which file has the least frequently used mwindows and close all mwindows of that file only. Rinse and repeat until we are below `open_limit`.

Basically, instead of iterating over `git_mwindow`s, we need to iterate over `git_mwindow_file`s and determine which one of them to close.

josharian: I see, I think.

From a quick look, that surgery might require more time than I really wanted to dedicate right now.

Do you think there is anyone who already knows this code that might be able to work on a proper fix? (If not, I may have to, since this issue is critical for me.)

lhchavez: i may be able to help, since it's also affecting us. let me see if i can wrangle a few hours this weekend.

pks-t:
On Thu, Feb 06, 2020 at 11:04:15AM -0800, lhchavez wrote:
> lhchavez commented on this pull request.
> > @@ -268,7 +272,8 @@ static git_mwindow *new_window(
>  
>  	ctl->mapped += (size_t)len;
>  
> -	while (git_mwindow__mapped_limit < ctl->mapped &&
> +	while ((git_mwindow__mapped_limit < ctl->mapped ||
> +	       git_mwindow__open_limit < ctl->open_windows) &&
> 
> i may be able to help, since it's also affecting us. let me see if i can wrangle a few hours this weekend.

Let me know if you don't find the time, then I could take over.
Shouldn't take too long, at least if I have a proper testcase.

src/mwindow.c
lhchavez: I could wrangle a few hours! #5396

josharian: Hooray! Thanks!! I've been under water fighting another bug, but I'll patch in your PR and give it a whirl soon.

ethomson:
Fixed via #5396

PR #5387, opened 2020-02-06
transports: http: fix custom headers not being applied

### LINKED ISSUE #5385: recent regression adding http headers to push/fetch

pks-t:
In commit b9c5b15a7 (http: use the new httpclient, 2019-12-22), the HTTP
code got refactored to extract a generic HTTP client that operates
independently of the Git protocol. Part of refactoring was the creation
of a new `git_http_request` struct that encapsulates the generation of
requests. Our Git-specific HTTP transport was converted to use that in
`generate_request`, but during the process we forgot to set up custom
headers for the `git_http_request` and as a result we do not send out
these headers anymore.

Fix the issue by correctly setting up the request's custom headers.

---

Fixes #5385.

### DIFF
--- a/src/transports/http.c
+++ b/src/transports/http.c
@@ -346,6 +346,7 @@ static int generate_request(
 	request->credentials = transport->server.cred;
 	request->proxy = use_proxy ? &transport->proxy.url : NULL;
 	request->proxy_credentials = transport->proxy.cred;
+	request->custom_headers = &transport->owner->custom_headers;
 
 	if (stream->service->method == GIT_HTTP_METHOD_POST) {
 		request->chunked = stream->service->chunked;
--- a/tests/online/clone.c
+++ b/tests/online/clone.c
@@ -392,6 +392,21 @@ void test_online_clone__credentials(void)
 	cl_fixture_cleanup("./foo");
 }
 
+void test_online_clone__credentials_via_custom_headers(void)
+{
+	const char *creds = "libgit3:libgit3";
+	git_buf auth = GIT_BUF_INIT;
+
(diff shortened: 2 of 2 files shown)


ethomson:
Oops, what an embarassing oversight.  Thanks @pks-t, this looks good to me.

Maybe we should have a network test that adds an `Authorization` header so that we verify that we don't break this in the future.  (We can follow up with that later though to get this merged quickly.)

pks-t:
> Maybe we should have a network test that adds an Authorization header so that we verify that we don't break this in the future. (We can follow up with that later though to get this merged quickly.)

I'd definitely love to have a fix. Let me see whether I can add one in online::clone using Basic auth.

pks-t:
Amended a test. The test works for me if setting up my own creds and fails without the fix.

pks-t:
Modified the test to not require any environment variables being set, as that caused CI to skip it.

PR #5388, opened 2020-02-06
Handle repository format v1

bk2204:
Git has supported repository format version 1 for some time.  This format is just like version 0, but it supports extensions. Implementations must reject extensions that they don't support.

Add support for this format version and reject any extensions but extensions.noop, which is the only extension we currently support.

This is a prerequisite to adding SHA-256 support.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -62,7 +62,8 @@ static const struct {
 	{ GIT_REPOSITORY_ITEM_COMMONDIR, GIT_REPOSITORY_ITEM_GITDIR, "worktrees", true }
 };
 
-static int check_repositoryformatversion(git_config *config);
+static int check_repositoryformatversion(int *version, git_config *config);
+static int check_extensions(git_config *config, int version);
 
 #define GIT_COMMONDIR_FILE "commondir"
 #define GIT_GITDIR_FILE "gitdir"
@@ -72,6 +73,7 @@ static int check_repositoryformatversion(git_config *config);
 #define GIT_BRANCH_MASTER "master"
 
 #define GIT_REPO_VERSION 0
+#define GIT_REPO_MAX_VERSION 1
 
 git_buf git_repository__reserved_names_win32[] = {
 	{ DOT_GIT, 0, CONST_STRLEN(DOT_GIT) },
@@ -804,6 +806,7 @@ int git_repository_open_ext(
 		gitlink = GIT_BUF_INIT, commondir = GIT_BUF_INIT;
 	git_repository *repo = NULL;
 	git_config *config = NULL;
+	int version = 0;
 
 	if (flags & GIT_REPOSITORY_OPEN_FROM_ENV)
 		return _git_repository_open_ext_from_env(repo_ptr, start_path);
@@ -845,7 +848,10 @@ int git_repository_open_ext(
 	if (error < 0 && error != GIT_ENOTFOUND)
 		goto cleanup;
 
-	if (config && (error = check_repositoryformatversion(config)) < 0)
+	if (config && (error = check_repositoryformatversion(&version, config)) < 0)
+		goto cleanup;
+
+	if ((error = check_extensions(config, version) < 0))
 		goto cleanup;
 
 	if ((flags & GIT_REPOSITORY_OPEN_BARE) != 0)
@@ -1341,11 +1347,11 @@ bool git_repository__reserved_names(
 }
 #endif
 
-static int check_repositoryformatversion(git_config *config)
+static int check_repositoryformatversion(int *version, git_config *config)
 {
-	int version, error;
+	int error;
 
-	error = git_config_get_int32(&version, config, "core.repositoryformatversion");
+	error = git_config_get_int32(version, config, "core.repositoryformatversion");
 	/* git ignores this if the config variable isn't there */
 	if (error == GIT_ENOTFOUND)
 		return 0;
@@ -1353,14 +1359,33 @@ static int check_repositoryformatversion(git_config *config)
 	if (error < 0)
 		return -1;
 
-	if (GIT_REPO_VERSION < version) {
+	if (GIT_REPO_MAX_VERSION < *version) {
 		git_error_set(GIT_ERROR_REPOSITORY,
-			"unsupported repository version %d. Only versions up to %d are supported.",
-			version, GIT_REPO_VERSION);
+			"unsupported repository version %d; only versions up to %d are supported",
+			*version, GIT_REPO_MAX_VERSION);
 		return -1;
 	}
 
 	return 0;
+}
+
+static int check_valid_extension(const git_config_entry *entry, void *payload)
+{
+	GIT_UNUSED(payload);
+
+	if (!strcmp(entry->name, "extensions.noop"))
+		return 0;
+
+	git_error_set(GIT_ERROR_REPOSITORY, "unsupported extension name %s", entry->name);
+	return -1;
+}
+
+static int check_extensions(git_config *config, int version)
+{
+	if (version < 1)
+		return 0;
+
+	return git_config_foreach_match(config, "^extensions\\.", check_valid_extension, NULL);
 }
 
 int git_repository_create_head(const char *git_dir, const char *ref_name)
@@ -1574,11 +1599,15 @@ static int repo_init_config(
 	git_config *config = NULL;
 	bool is_bare = ((flags & GIT_REPOSITORY_INIT_BARE) != 0);
 	bool is_reinit = ((flags & GIT_REPOSITORY_INIT__IS_REINIT) != 0);
+	int version = 0;
 
 	if ((error = repo_local_config(&config, &cfg_path, NULL, repo_dir)) < 0)
 		goto cleanup;
 
-	if (is_reinit && (error = check_repositoryformatversion(config)) < 0)
+	if (is_reinit && (error = check_repositoryformatversion(&version, config)) < 0)
+		goto cleanup;
+
+	if ((error = check_extensions(config, version) < 0))
 		goto cleanup;
 
 #define SET_REPO_CONFIG(TYPE, NAME, VAL) do { \
--- a/tests/repo/open.c
+++ b/tests/repo/open.c
@@ -35,7 +35,53 @@ void test_repo_open__format_version_1(void)
 
 	git_config_free(config);
 	git_repository_free(repo);
+
+	git_repository_open(&repo, "empty_bare.git");
+	cl_assert(git_repository_path(repo) != NULL);
+	cl_assert(git__suffixcmp(git_repository_path(repo), "/") == 0);
+	git_repository_free(repo);
+}
+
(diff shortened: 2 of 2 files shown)


ethomson:
> Implementations must reject extensions that they don't support.

We need to have a plan for this, since it's hard to really know what extensions we support.  Extensions are a mixed bag of plumbing and application layer configuration.

Concretely: libgit2 can support `worktreeConfig` just fine, but `preciousObjects` is something that an application author may wish to support.  We need to allow somebody to build a repacker on top of libgit2 that can honor it while obeying the "if a version-1 repository specifies any extensions.* keys that the running git has not implemented, the operation MUST NOT proceed" notion.

I presume that this means that applications will need a mechanism by which they can register that they understand a particular extension, and then we can just claim understanding on their behalf when opening a config.

One gotcha here is that git [has specified identical language](https://git-scm.com/docs/gitrepository-layout) in the _unknown extension in the configuration_ case as the _unknown _value_ to an extension configuration_ case:

> If a version-1 repository specifies any `extensions.*` keys that the running git has not implemented, the operation MUST NOT proceed. Similarly, if the value of any known key is not understood by the implementation, the operation MUST NOT proceed.

One is to assume that "the operation MUST NOT proceed" in the first sentence must refer to _literally any_ git operation, because you have no way of knowing whether the extension applies to the operation that you were asked to do.  That is to say, if an implementation doesn't understand the `preciousObjects` extension, then it can't _do anything at all_ and all operations must fail.

However, if one is to apply that same thinking to the next sentence, then does `git clone` need to validate that the value of `preciousObjects` is "understood by the implementation"?  This would mean that we would need implementations to give us some list of values that are legitimate, which is terrible.

🤔

src/repository.c
```diff
@@ -62,7 +62,8 @@ static const struct {
 	{ GIT_REPOSITORY_ITEM_COMMONDIR, GIT_REPOSITORY_ITEM_GITDIR, "worktrees", true }
 };
 
-static int check_repositoryformatversion(git_config *config);
+static int check_repositoryformatversion(git_config *config, int *version);
```
ethomson: Style nit: we specify "out" type params first.  eg, this prototype should be:

```c
static int check_repositoryformatversion(int *version, git_config *config);
```

src/repository.c
```diff
@@ -1341,28 +1347,47 @@ bool git_repository__reserved_names(
...
 	if (error == GIT_ENOTFOUND)
 		return 0;
 
 	if (error < 0)
 		return -1;
 
-	if (GIT_REPO_VERSION < version) {
+	if (GIT_REPO_MAX_VERSION < *version) {
 		git_error_set(GIT_ERROR_REPOSITORY,
 			"unsupported repository version %d. Only versions up to %d are supported.",
```
ethomson: I know you didn't create this problem, but while you're here, would you mind fixing this?

Let's not split sentences, since it can make application code that tries to display error messages weird.  They're often trying to incorporate our messages into their own and they're probably doing that by adding a period to the end themselves.  eg:

```
unsupported repository version %d; only versions up to %d are supported
```

tests/repo/open.c
```diff
@@ -35,7 +35,53 @@ void test_repo_open__format_version_1(void)
 
 	git_config_free(config);
 	git_repository_free(repo);
+
+	git_repository_open(&repo, "empty_bare.git");
+	cl_assert(git_repository_path(repo) != NULL);
+	cl_assert(git__suffixcmp(git_repository_path(repo), "/") == 0);
+	git_repository_free(repo);
```
ethomson: I don't understand this addition, can you explain it in a comment on the commit?

tests/repo/open.c
```diff
@@ -35,7 +35,53 @@ void test_repo_open__format_version_1(void)
...
+	cl_git_pass(git_config_set_int32(config, "core.repositoryformatversion", 1));
+	cl_git_pass(git_config_set_int32(config, "extensions.noop", 1));
+
+	git_config_free(config);
+	git_repository_free(repo);
+
+	git_repository_open(&repo, "empty_bare.git");
+	cl_assert(git_repository_path(repo) != NULL);
+	cl_assert(git__suffixcmp(git_repository_path(repo), "/") == 0);
+	git_repository_free(repo);
```
ethomson: Similarly, this feels misplaced.  Why's it in this test?

ethomson:
For more information, see the comments in https://github.com/libgit2/libgit2/pull/3253

ethomson:
Also - to clarify my comments above - I'm not looking for you to solve this problem _here_, I'm just talking it over.  Obviously there are a great many extensions that we can implement ourselves without worrying about application support above us and certainly `noop` is one of them.  ;)

tiennou:
There's also some points [here](https://github.com/libgit2/libgit2/pull/5105#issuecomment-520415973), about one way to implement those "extensions points", at the protocol level. I planned to share that with the repository layer, but never went through with it… It's been a while, as well 😉.

tests/repo/open.c
bk2204: This is just the way the tests above check for success.  If you're happy with just checking for success opening the repository, I'm happy to just do that.

src/repository.c
bk2204: Sure, I can fix that.

bk2204:
> > Implementations must reject extensions that they don't support.
> 
> We need to have a plan for this, since it's hard to really know what extensions we support. Extensions are a mixed bag of plumbing and application layer configuration.
> 
> Concretely: libgit2 can support `worktreeConfig` just fine, but `preciousObjects` is something that an application author may wish to support. We need to allow somebody to build a repacker on top of libgit2 that can honor it while obeying the "if a version-1 repository specifies any extensions.* keys that the running git has not implemented, the operation MUST NOT proceed" notion.
> 
> I presume that this means that applications will need a mechanism by which they can register that they understand a particular extension, and then we can just claim understanding on their behalf when opening a config.

I think there are several options that we can support out of the box just natively.  You've mentioned `worktreeConfig` as one; `noop` is a trivial example of another.  We _may_ want to allow people to indicate they support certain, specified other ones, such as `preciousObjects`.  However, that should be a whitelist of options we know people can support.  If someone lied to us and said they supported `extensions.objectFormat`, we should not allow them to do that, because doing so can and will corrupt the repository.

All of that, however, is orthogonal to this series, which implements only the barest, most essential v1 repository format, and nothing more.  My goal is to get SHA-256 support working, so at the moment, I'm not planning to offer folks a way to register extensions.  That would be good and valuable work for someone else to pick up, though.

> One gotcha here is that git [has specified identical language](https://git-scm.com/docs/gitrepository-layout) in the _unknown extension in the configuration_ case as the _unknown _value_ to an extension configuration_ case:
> 
> > If a version-1 repository specifies any `extensions.*` keys that the running git has not implemented, the operation MUST NOT proceed. Similarly, if the value of any known key is not understood by the implementation, the operation MUST NOT proceed.
> 
> One is to assume that "the operation MUST NOT proceed" in the first sentence must refer to _literally any_ git operation, because you have no way of knowing whether the extension applies to the operation that you were asked to do. That is to say, if an implementation doesn't understand the `preciousObjects` extension, then it can't _do anything at all_ and all operations must fail.

Yes, that's my understanding.  If the config is set and not understood, no operation by that implementation may proceed on the repository.
 
> However, if one is to apply that same thinking to the next sentence, then does `git clone` need to validate that the value of `preciousObjects` is "understood by the implementation"? This would mean that we would need implementations to give us some list of values that are legitimate, which is terrible.

Yes, `git clone` needs to hard fail if the config is set in the global or system configuration and it's not understood.  That's the way Git works, and it's important for data integrity if people are relying on their Git server not destroying objects (e.g., for compliance or backup reasons).

I think we can have a list of extensions which we are willing to support, whether internally or with help from the program calling libgit2.  If we don't know about it, then users can't specify that they handle it.  That's the safe and robust approach.

ethomson:
> I think we can have a list of extensions which we are willing to support, whether internally or with help from the program calling libgit2. If we don't know about it, then users can't specify that they handle it. That's the safe and robust approach.

I agree.  My concern is that the interface for users to specify this is severely disappointing, since they have to specify all the values that they support as well.

If the extension model in git was to define the supported value range and keep it immutable, it would be significantly easier for us.  That's not a thing we can solve here, of course, but I'll bring this up on the git mailing list.

tests/repo/open.c
ethomson: Huh!  OK, there's probably a reason for this, but I've forgotten what it is.  😀 Let's keep it.

tiennou:
My own "design plan" had been to get #4856 merged first, mostly because the current repository open path is somewhat convoluted (there's a few discrepancies because of custom/in-memory repositories, worktrees and envvars collusion 😉). This would enable passing the supported list of options could be passed as part of repository opening, as a `git_strarray`, or maybe some "validation" callback ; which could also benefit the custom/in-memory users…

How to check for options would then be either handled internally as "the default list", a.k.a the only extensions libgit2 has support for/knows about, or delegated to the user's code.

src/repository.c
bk2204: Fixed.

bk2204:
I think I'd prefer to merge this series first, since #4856 appears to not have seen any changes since October and also isn't passing tests.  This series should, with my latest changes, be ready to merge as it is.

I realize it doesn't have a lot of additional support for user extensibility, but it is a framework to build upon, and it is also completely API-compatible with older versions.

ethomson:
> I realize it doesn't have a lot of additional support for user extensibility, but it is a framework to build upon, and it is also completely API-compatible with older versions.

Seems reasonable.  I think that there are open questions around how we support user-extensibility. 
 In the meantime, there's none here so this seems like we should :+1: .  Thanks, @bk2204.

We're basically locked down for the imminent 0.99, we'll :shipit: this asap once we open up the tree again.

ethomson:
@pks-t as well for review

pks-t:
On Wed, Apr 01, 2020 at 02:24:05PM -0700, Edward Thomson wrote:
> @pks-t as well for review

I'll review it soonish, thanks for the reminder. This is a requirement
for the reftable backend, so I'm interested in it anyway.

Automobileservices4:
#5388

pks-t:
This PR looks fine as is to me, thanks a lot for working on it @bk2204!

> Seems reasonable. I think that there are open questions around how we support user-extensibility.

I guess we need a mechanism similar to the transport registration, where an application may provide a callback it's being passed to all extensions that a repository requires. But, this is going to open a can of worms I've already been stumbling over when implementing reftables. Want to handle an extension that says the reference database is not the fs-backend one anymore? Well, good luck doing that without implementing this support in libgit2 directly.

We'll need to grow more extensible in some places, refdbs is only one of them. And I'm not sure we can grow generic enough to allow applications to handle arbitrary extensions in the long run. Which doesn't mean we shouldn't try our best.

Anyway, this PR is a good step in the right direction to support extensions at all and doesn't force us down any specific route, so I'm happy to merge it. Thanks again!

Automobileservices4:
#5388

PR #5391, opened 2020-02-07
Coverity fixes

pks-t:
Coverity's alive again, spitting out a bunch of errors introduced since July 2019. This PR fixes a bunch of them.

### DIFF
--- a/src/push.c
+++ b/src/push.c
@@ -349,8 +349,9 @@ static int queue_objects(git_push *push)
 		if (git_oid_is_zero(&head->oid))
 			continue;
 
-		/* TODO */
-		git_revwalk_hide(rw, &head->oid);
+		if ((error = git_revwalk_hide(rw, &head->oid)) < 0 &&
+		    error != GIT_ENOTFOUND && error != GIT_EINVALIDSPEC && error != GIT_EPEEL)
+			goto on_error;
 	}
 
 	error = git_packbuilder_insert_walk(push->pb, rw);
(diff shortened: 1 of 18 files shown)


src/push.c
```diff
@@ -349,8 +349,8 @@ static int queue_objects(git_push *push)
 		if (git_oid_is_zero(&head->oid))
 			continue;
 
-		/* TODO */
-		git_revwalk_hide(rw, &head->oid);
+		if ((error = git_revwalk_hide(rw, &head->oid)) < 0 && error != GIT_ENOTFOUND)
```
ethomson: I think that we need to filter out other non-errors here.  (object is not a commit)  But otherwise, I 👍 this PR once CI is green.

pks-t: Easier said than done. Quoting `git_revwalk__push_commit`, which is invoked by `git_revwalk_hide`:

```
error = git_object_peel(&obj, oobj, GIT_OBJECT_COMMIT);
	git_object_free(oobj);

	if (error == GIT_ENOTFOUND || error == GIT_EINVALIDSPEC || error == GIT_EPEEL) {
		/* If this comes from e.g. push_glob("tags"), ignore this */
		if (opts->from_glob)
			return 0;

		git_error_set(GIT_ERROR_INVALID, "object is not a committish");
		return -1;
	}
```

pks-t: I'll just update the return statement to return `error` verbosely.

PR #5395, opened 2020-02-07
Use error code GIT_EAUTH for authentication failures

### LINKED ISSUE #5389: use error code GIT_EAUTH for HTTP auth failures
b9c5b15 is backwards incompatible. (Related: #5385.) Prior to that commit, HTTP auth failures had an error class of `GIT_ERROR_NET`. That commit changes them to `GIT_ERROR_HTTP`. I will adjust my code accordingly, but it should probably be called out in the release notes.

That said, it'd be nicer to have a more precise error code for auth failures. For lack of a good alternative, I currently check the error class and then do a string match for the word "authentication", because the error code is a generic `GIT_ERROR`. There is already an error code for authentication failures, `GIT_EAUTH`, but it appears that only the ssh network operations return it.

I'd like to suggest that HTTP use it too. Now is a particularly good time to switch, since people may have to update their error handling in light of b9c5b15 anyway.

cc @ethomson @pks-t

josharian:
See the individual commit messages for more detail.

Fixes #5389


src/transports/winhttp.c
```diff
@@ -972,7 +972,7 @@ static int winhttp_stream_read(
 	/* Enforce a reasonable cap on the number of replays */
 	if (replay_count++ >= GIT_HTTP_REPLAY_MAX) {
 		git_error_set(GIT_ERROR_HTTP, "too many redirects or authentication replays");
-		return -1;
+		return GIT_ERROR; /* not GIT_EAUTH because the exact cause is not clear */
```
ethomson: 🤔  This is an interesting one.  I wonder if we should examine all the attempts and return `GIT_EAUTH` if they were all authentication failures and `-1` otherwise.  (Similarly with the other loops.)

What do you think?

src/transports/httpclient.c
```diff
@@ -585,8 +585,8 @@ static int apply_credentials(
 		if (auth->connection_affinity)
 			free_auth_context(server);
 	} else if (!token.size) {
-		git_error_set(GIT_ERROR_HTTP, "failed to respond to authentication challange");
-		error = -1;
+		git_error_set(GIT_ERROR_HTTP, "failed to respond to authentication challenge");
```
ethomson: Oops, good catch.  Thanks!

src/transports/winhttp.c
josharian: My instinct would be to return `GIT_EAUTH` if any were authentication failures, since the other redirects are likely to be leading up to or a consequence of the auth failure. That is, it seems to me that auth failure should be sticky. Or maybe our divergent instincts mean that we should leave it as is. Either way, I'd prefer to do that as a follow-up.

src/transports/httpclient.c
```diff
@@ -598,6 +598,7 @@ static int apply_credentials(
 	} else if (!token.size) {
 		git_error_set(GIT_ERROR_HTTP, "failed to respond to authentication challenge");
 		error = GIT_EAUTH;
+		error = -1;
```
josharian, line 601: This doesn't look right. I think we want `error = GIT_EAUTH`, but we definitely don't need both assignments. :)

ethomson: Yep, agreed.  Fixed in main.

PR #5396, opened 2020-02-08
mwindow: set limit on number of open files

lhchavez:
There are some cases in which repositories accrue a large number of
packfiles. The existing mwindow limit applies only to the total size of
mmap'd files, not on their number. This leads to a situation in which
having lots of small packfiles could exhaust the allowed number of open
files, particularly on macOS, where the default ulimit is very low
(256).

This change adds a new configuration parameter
(GIT_OPT_SET_MWINDOW_FILE_LIMIT) that sets the maximum number of open
packfiles, with a default of 128. This is low enough so that even macOS
users should not hit it during normal use.

Based on PR #5386, originally written by @josharian.

Fixes: #2758


josharian:
I tried this out in my application. It definitely helps.

It doesn't seem to limit the number of open files to anywhere particularly close to the set value.

I set a limit of 32 and ran for a while, doing a bunch of little network operations, and found that lsof stabilized at around 64 open `.pack` files and around 45 open `.idx` files.

I changed the limit to 16, and found that it stabilized at around 32 open `.pack` files and 45 open `.idx` files.

At a limit of 8 (!), it stabillized at around 16 open `.pack` files and 45 open `.idx` files.

So it seems like maybe there's a factor-of-two bug in the `.pack` file handling, and something entirely different or missing in the `.idx` file handling.

The mere fact of stability and no errors is already a huge improvement, though; thank you!

josharian:
Update: as I continue to run, the number of open `.idx` files continues to grow, albeit slowly. Currently at 68 and growing. (File limit set to 32.)

lhchavez:
yeah, there is no limit on the number of `.idx` files opened, since this only checks `.pack`. still i'm surprised about that 2x discrepancy O_o. let me take another look

josharian:
Oh, I bet lsof reports two lines for each open .pack file, one for the regular open and one for the mmap.

josharian:
Is there a good way to set a limit on the number of `.idx` files open?

lhchavez:
> Oh, I bet lsof reports two lines for each open .pack file, one for the regular open and one for the mmap.

TIL (haven't used `lsof` too much myself). normally i just count the number of entries in `/proc/$PID/fd` but that might be a linux-ism.

> I wonder whether they count against the "open files" limit? They have the same fd...

they don't! that limit is really "number of open file descriptors", so if the same fd appears twice, it's only counted once by the kernel.

> Is there a good way to set a limit on the number of `.idx` files open?

i'll try something later today. if it doesn't pan out, i'll leave a few TODOs to document and fix later since this change by itself is still making some amount of progress.

lhchavez:
oh wait, i just realized that all the `.idx` files are also mmapped! no paths in `pack_index_check` exit that function without closing `fd`:

https://github.com/libgit2/libgit2/blob/ee3307a183e39d602b25fa94831c6fc09e7c1b61/src/pack.c#L209-L230

confirmed that this works by running

```shell
libgit2$ git apply<<EOF
diff --git a/tests/pack/filelimit.c b/tests/pack/filelimit.c
index 884a2668f..76a15a2b0 100644
--- a/tests/pack/filelimit.c
+++ b/tests/pack/filelimit.c
@@ -30,6 +30,10 @@ void test_pack_filelimit__open_repo_with_1025_packfiles(void)
                ++i;
        cl_assert_equal_i(1025, i);

+       fprintf(stderr, "wait...");
+       fflush(stderr);
+       fgetc(stdin);
+
        git_revwalk_free(walk);
        git_repository_free(repo);
 }
EOF
libgit2$ cmake --build .
libgit2$ ulimit -n140  # cutting it close
libgit2$ ./libgit2_clar -spack::filelimit &
libgit2$ sleep 1
libgit2$ lsof -p `pidof libgit2_clar` | grep REG | wc -l
1294
libgit2$ lsof -p `pidof libgit2_clar` | grep REG | grep -v '\bmem\b' | wc -l
129
libgit2$ ls -l /proc/`pidof libgit2_clar`/fd | wc -l  # 5 extra FDs = stdio + random/urandom devices
134
```

so it seems that all the extra entries in lsof are expected and this patch WAI :fireworks:.

josharian:
I’m on my phone, so maybe I’m missing something, but mmap’d files still keep the fd open. (I was double-counting the .pack files by including mmap reference and regular reference, but the mmap-only .idx handles still count.)

josharian:
> mmap’d files still keep the fd open

I should google before typing. Apparently this is not so. TIL! :)

josharian:
Naive question: do we need the .pack files open, or could we mmap-and-close them too?

lhchavez:
> Naive question: do we need the .pack files open, or could we mmap-and-close them too?

IIUC the mwindow is kept open since the entirety of the file is not mmapped, possibly to avoid running out of memory ([especially address space in 32-bit architectures](https://www.sublimetext.com/blog/articles/use-mmap-with-care)) since the `.idx` files are extremely small compared to the `.pack` files.

josharian:
One possible future optimization: If the pack file is less than `git_mwindow__window_size`, close it immediately after mmap, since we know that we are going to mmap the entirety of the file anyway.

josharian:
Ping @pks-t and @ethomson. It'd be lovely to get this merged; I've been running with this for a while and it has been stable and has solved the issues I was having.

jyn514:
Any updates? This is causing trouble for us in docs.rs, we're slowly leaking file descriptors. In the following image, the spike to 1.8k fds correspondings to deploying a PR using libgit2.

https://cdn.discordapp.com/attachments/606508111628468225/724214122694180904/2020-06-21--12-48-12.png

The spike is unfortunate but we can live with, but the number of file descriptors goes up over time and will eventually crash the server.

include/git2/common.h
```diff
@@ -227,8 +229,18 @@ typedef enum {
...
+ *
+ *	* opts(GIT_OPT_GET_MWINDOW_FILE_LIMIT, size_t *):
+ *
+ *		> Get the maximum number of files that will be mapped at any time by the
+ *		> library
+ *
+ *	* opts(GIT_OPT_SET_MWINDOW_FILE_LIMIT, size_t):
+ *
+ *		> Set the maximum number of files that can be mapped at any time
+ *		> by the library
```
pks-t, line 243: I wonder whether we want to be more conservative here and default to `0`, which would essentially mean "No mwindow file limit at all" like we previously had it. 128 does sound quite low to me and I can imagine it to cause quite some performance regressions when handling many repos at once.

src/mwindow.c
```diff
@@ -242,9 +240,38 @@ static int git_mwindow_close_lru(git_mwindow_file *mwf)
...
+/*
+ * Close the file that contains the least recently used window. Called under
+ * lock from new_window.
+ */
+static int git_mwindow_close_lru_file(void)
+{
+	git_mwindow_ctl *ctl = &mem_ctl;
+	git_mwindow_file *lru_f = NULL, *cur;
+	size_t i;
+	git_mwindow *lru_w = NULL, *lru_l = NULL;
```
pks-t, line 252: I know you're just adapting to existing code style, but could we please give these more descriptive names?

src/mwindow.c
```diff
@@ -242,9 +240,38 @@ static int git_mwindow_close_lru(git_mwindow_file *mwf)
 	return 0;
 }
 
+/*
+ * Close the file that contains the least recently used window. Called under
+ * lock from new_window.
```
pks-t, line 245: Does this strategy make sense? It could be that this file does have the LRU window, but it could also at the same time have the MRU window as well. So we'd be pulling out the carpet under our feet if we were to close that.

A different strategy would be to scan each file for the MRU window and then chosing the LRU window across these.

Sorry if it's what you're doing here already, I find the mwindow code extremely hard to read.

tests/pack/filelimit.c
```diff
@@ -0,0 +1,36 @@
...
+	git_oid id;
+	int i;
+
+	/*
+	 * This repository contains 1025 packfiles, each with one commit, one tree,
+	 * and two blobs. The first blob (README.md) has the same content in all
+	 * commits, but the second one (file.txt) has a different content in each
+	 * commit.
+	 */
+	cl_git_pass(git_repository_open(&repo, cl_fixture("1025.git")));
```
pks-t, line 17: Instead of adding a new test repository for this, how about we do something similar to the following:

```
void test_pack_filelimit__initialize_tiny(void)
{
        cl_git_pass(git_libgit2_opts(GIT_OPT_SET_MWINDOW_FILES, 1));
}

void test_pack_filelimit__initialize_medium(void)
{
        cl_git_pass(git_libgit2_opts(GIT_OPT_SET_MWINDOW_FILES, 100));
}

void test_pack_filelimit__initialize_huge(void)
{
        cl_git_pass(git_libgit2_opts(GIT_OPT_SET_MWINDOW_FILES, 1024));
}

void test_pack_filelimit__cleanup(void)
{
       cl_git_pass(git_libgit2_opts(GIT_OPT_SET_MWINDOW_FILES, MWINDOW_FILES_DEFAULT));
}

void test_pack_filelimit__revwalk(void)
{
        /* your test with a preexisting repo */
}
```

src/mwindow.c
lhchavez: nope, i did the naïve LRU.

and now that i take another look, the current implementation could close a file that has an in-use window! fix to that incoming, and while i'm at it, i'll do the LR-MRU algorithm you suggested.

tests/pack/filelimit.c
lhchavez: the problem with that approach is that if there were to be a regression, since none of the test repositories have enough packfiles to reproduce the issue (the test repos have a maximum of 4 packfiles), the tests would still pass :(

and changing the open file limits to trigger the issue would have other problems because we would need to keep it so low that accidentally changing the order in which files are opened or closing a file a bit late would cause this test to spuriously fail, making the test unnecessarily fragile.

i haven't found many other ways to test this, but here are a few alternatives:

* create a repository with only ~100 packfiles and change the open file limit in the CI (but that would lead to the tests behaving differently on developers' machines)
* same as above (create a repository with ~100 packfiles), but call `setrlimit` programmatically during the initialization of the test.
* generate the packfiles at runtime (although due to https://github.com/libgit2/libgit2/issues/5399 this behavior is currently sort of slow).

wdyt?

include/git2/common.h
lhchavez: Done.

src/mwindow.c
lhchavez: Of course!

tests/pack/filelimit.c
lhchavez: Okay, I improved this a bit and added part of your suggestion (although had to drop the 1024 file limit since that would cause Linux machines with default limits to fail the tests) and I'm now asserting that the number of open files is the expected one, but we still need a test repo with a few packfiles: none of the pre-existing repos managed to open more than one packfile at a time.

But with this we can prune the testrepo to a pretty low number of packfiles (say, 16). Would that make it more palatable?

pks-t: > the problem with that approach is that if there were to be a regression, since none of the test repositories have enough packfiles to reproduce the issue (the test repos have a maximum of 4 packfiles), the tests would still pass :(

True, 4 maximum isn't that many. It could be detected in case we set limit to 1, though.

> generate the packfiles at runtime (although due to #5399 this behavior is currently sort of slow).

Did you try using the `git_packbuilder` interface to do this? If we settle for a "sane" limit like generating a 100 packfiles I'd imagine it to not cause much of a slowdown.

> But with this we can prune the testrepo to a pretty low number of packfiles (say, 16). Would that make it more palatable?

Yeah, 16 would definitely be fine with me, if using the packbuilder interface doesn't work.

lhchavez: > > the problem with that approach is that if there were to be a regression, since none of the test repositories have enough packfiles to reproduce the issue (the test repos have a maximum of 4 packfiles), the tests would still pass :(
> 
> True, 4 maximum isn't that many. It could be detected in case we set limit to 1, though.

It couldn't, since at most all the pre-existing repositories use one packfile per ref.

> 
> > generate the packfiles at runtime (although due to #5399 this behavior is currently sort of slow).
> 
> Did you try using the `git_packbuilder` interface to do this? If we settle for a "sane" limit like generating a 100 packfiles I'd imagine it to not cause much of a slowdown.

Yes, even with `git_packbuilder`, adding packfiles is potentially quadratic since every time a new packfile appears, the cache is flushed and in the worst case every packfile needs to be read before finding the right one.

Now that I'm explicitly checking how many files are still open, just 16 commits does the job, so the quadratic argument above has been mitigated.

> 
> > But with this we can prune the testrepo to a pretty low number of packfiles (say, 16). Would that make it more palatable?
> 
> Yeah, 16 would definitely be fine with me, if using the packbuilder interface doesn't work.

src/mwindow.c
```diff
@@ -174,82 +178,142 @@ int git_mwindow_contains(git_mwindow *win, off64_t offset)
...
-	git__free(lru_w);
+	git__free(lru_window);
 	ctl->open_windows--;
 
 	return 0;
 }
 
+/*
+ * Close the file that does not have any open windows AND contains the
+ * least-recently-used most-recently-used window.
```
pks-t, line 277: This comment is a bit confusing :P Proposal: `Close the file whose most-recentlty-used window is the least-recently used one across all currently open files`?

tests/pack/filelimit.c
```diff
@@ -0,0 +1,138 @@
+#include "clar_libgit2.h"
+#include "mwindow.h"
+#include "global.h"
+
+#include <git2.h>
+#include "git2/sys/commit.h"
+#include "git2/sys/mempack.h"
+
+static size_t _expected_open_mwindow_files = 0;
+static size_t _original_mwindow_file_limit = 0;
```
pks-t, line 10: Leading underscores are reserved by the C/libc implementations and thus aren't allowed to be used. Sorry to be nitpicky here, I bet you'd not even have to look far to found counterexamples in the current code already where we're doing exactly what you're doing here.

tests/pack/filelimit.c
```diff
@@ -0,0 +1,138 @@
...
+	cl_fixture_cleanup(path.ptr);
+	git_buf_dispose(&path);
+}
+
+/*
+ * Create a packfile with one commit, one tree, and two blobs. The first blob
+ * (README.md) has the same content in all commits, but the second one
+ * (file.txt) has a different content in each commit.
+ */
+void create_packfile_commit(
```
pks-t, line 50: :tada: Thanks a lot for doing this! Did you see any major slowdowns introduced by this way of creating the packfiles? Mostly just curious as you indicated that might be the case

tests/pack/filelimit.c
```diff
@@ -0,0 +1,138 @@
...
+	const size_t commit_count = 16;
+	unsigned int open_windows;
+
+	/*
+	 * Create a repository and populate it with 16 commits, each in its own
+	 * packfile.
+	 */
+	cl_git_pass(git_buf_joinpath(&path, clar_sandbox_path(), "repo.git"));
+	cl_git_pass(git_repository_init(&repo, path.ptr, true));
+	for (i = 1; i <= commit_count; ++i) {
```
pks-t, line 108: Micro-nit: we usually start iterating from `0`, not `1`. You can then just pass `i+1` to `create_packfile_commit`.

tests/pack/filelimit.c
```diff
@@ -0,0 +1,138 @@
...
+		parent_id = &id;
+	}
+
+	cl_git_pass(git_revwalk_new(&walk, repo));
+	cl_git_pass(git_revwalk_sorting(walk, GIT_SORT_TOPOLOGICAL));
+	cl_git_pass(git_revwalk_push_ref(walk, "refs/heads/master"));
+
+	/*
+	 * Walking the repository requires eventually opening each of the packfiles.
+	 */
```
pks-t, line 119: Nit: single-line comments don't need the newlines, so you can just say `/* Walking ... the packfiles. */` in a single line.

tests/pack/filelimit.c
lhchavez: with only 15 files it's imperceptible. since it's quadratic, it starts becoming annoying on the order of ~100 and painful beyond 1k.

tests/pack/filelimit.c
lhchavez: i (and my editor) forget that this repository does not have a strict 80-col line length.

src/mwindow.c
lhchavez: added the bit to mention that the file should not have any open windows to your proposal.

ethomson:
LGTM too, thanks @lhchavez !

src/mwindow.c
```diff
@@ -174,82 +178,143 @@ int git_mwindow_contains(git_mwindow *win, off64_t offset)
...
+	git_mwindow *lru_window = NULL;
+	size_t i;
+
+	git_vector_foreach(&ctl->windowfiles, i, current_file) {
+		git_mwindow *mru_window = NULL;
+		if (!git_mwindow_scan_recently_used(
+				current_file, &mru_window, NULL, true, GIT_MWINDOW__MRU)) {
+			continue;
+		}
+		if (!lru_window || lru_window->last_used > mru_window->last_used)
```
tiennou, line 295: Just as a heads up, Coverity seems to have flagged this line (CID 1437618) as `Execution cannot reach the expression lru_window->last_used > mru_window->last_used inside this statement`, since `lru_window` is initialized to `NULL` and nothing in the loop updates it.

lhchavez: neat! GG Coverity :D opened https://github.com/libgit2/libgit2/pull/5783 to address it.

thanks for the heads-up! and just out of curiosity, is there a way to get Coverity reports running on GH so that we can all benefit from that code analysis goodness?

tiennou: Sure, the project page is here : https://scan.coverity.com/projects/libgit2. Make yourself an account over there and I'll add you up, which should allow you to get to the reported issues. Sadly, I don't think there's a better GH integration.

lhchavez: all right, that should be enough. Thanks again!

PR #5397, opened 2020-02-10
CIFuzz dog food

Leo-Neat:
OSS-Fuzz was wondering if libgit2 would dogfood CIFuzz,
a new feature being rolled out in OSS-Fuzz. It allows for pull
requests to be fuzzed in CI using fuzz targets enabled in OSS-Fuzz.
It will be active in dry-run mode meaning that it will never fail.


Leo-Neat:
@ethomson  When you get a chance can you have a look at this dog food? Let me know if you have any questions.

PR #5398, opened 2020-02-11
openssl: fix Valgrind issues in nightly builds

### LINKED ISSUE #5394: httpclient: reads on uninitialized memory
I've been trying to reproduce valgrind errors listed in our [nightly jobs](https://libgit2.visualstudio.com/libgit2/_build/results?buildId=3305&view=logs&j=06c8e146-9583-5f95-f4c5-26a0fe76850a&t=36635f40-af6f-5b1a-9736-6853f5d84f9c) but was unable up to now. I also don't quite get where the uninitialized memory should come from, but yeah. @ethomson, maybe you've got a clue?

pks-t:
Our nightly builds are reporting quite a lot of issues due to use of uninitialized bytes, as reported in #5394. The root cause is inconsistencies between our normal and our nightly pipelines, where our nightly jobs missed "-DVALGRIND". We thus didn't use the sanitizing macro provided by Valgrind for these bytes, causing the warnings.

This PR fixes two issues: first, it consolidates -DVALGRIND into -DUSE_LEAKS_CHECKER=valgrind to avoid above errors going forward. Fixing this breaks our Bionic builds, though, as they use OpenSSL 1.1, which made `struct SSL` opaque. The second commit thus switches us over to using our own allocator that initializes all bytes and sets it up for use by OpenSSL, removing the old Valgrind macros. Added benefit is that this works generically across leaks checkers, even though it's not yet wired up for the others.

Closes #5394.


ethomson:
> We thus didn't use the sanitizing macro provided by Valgrind for these bytes, causing the warnings.

Aha.  😭  Well, that makes sense as least.

pks-t:
Btw, failure is nightlies is currently due to ARM32. They're the last thing on my CI spring cleanup list, after that nightlies should be up to task again.

ethomson:
Regrettably, the error in the ARM builds is not actually in our code, it's a problem in QEMU when running Go images.  Which docker is.  One can, as best as I can tell, no longer run docker multiarch images that are ARM32 or ARM64.

I propose that we remove those from the nightly matrix for now.  I like testing nightlies on ARM, so we might want to consider putting real hardware here.  I had considered this before, and it's probably time to reconsider.

Let's get this in and then release 0.99?

pks-t:
> Regrettably, the error in the ARM builds is not actually in our code, it's a problem in QEMU when running Go images. Which docker is. One can, as best as I can tell, no longer run docker multiarch images that are ARM32 or ARM64.

There's some good news here: it's not the Qemu/Docker/Go combination in general, but this has been broken by me when introducing gosu for unprivileged builds, which is a Go-based binary. I've replaced it with sudo in #5406, fixing our ARM builds. There's one test failure due to insufficient RAM on ARM32 which I fix in that PR, too, as well as some smallish cleanups to our CI infrastructure.

> I propose that we remove those from the nightly matrix for now. I like testing nightlies on ARM, so we might want to consider putting real hardware here. I had considered this before, and it's probably time to reconsider.

So luckily, this is not required. We should just merge #5406 to fix them properly.

> Let's get this in and then release 0.99?

Yup, agreed. Let's merge this here and #5406 and then go for v0.99. Care to give the other PR a review so that we can fast-track it? I'll then handle the release on Wednesday or Thursday, depending on how fast we are.

ethomson:
> There's some good news here: it's not the Qemu/Docker/Go combination in general, but this has been broken by me when introducing gosu for unprivileged builds, which is a Go-based binary. I've replaced it with sudo in #5406, fixing our ARM builds. There's one test failure due to insufficient RAM on ARM32 which I fix in that PR, too, as well as some smallish cleanups to our CI infrastructure.

That's great.  I was thinking that the go errors were _docker_, not _gosu_ -- docker is written in go, IIRC, and I was definitely hitting this locally, even when (trying to) override the `entrypoint` for the container.  Possible that my dockerfu is just weak and I was doing this wrong and still hitting `gosu` somehow.  Thanks for taking care of this.

Issue #5399, opened 2020-02-13
support git 2.21's multi-pack-index


lhchavez:
Starting from version 2.21, git supports [`multi-pack-index`](https://github.blog/2019-02-24-highlights-from-git-2-21/#multi-pack-indexes), which allows for O(log _n_) scan of _n_ objects among _m_ packfiles (instead of O(_m_ log _n/m_)) without needing to do a garbage collection to compact all the objects into a single `.pack` file. This is desirable in cases where there are lots of [Delta islands](https://github.blog/2019-02-24-highlights-from-git-2-21/#delta-islands) or it is otherwise undesirable to merge large numbers of packfiles.

This should be relatively straightforward to implement. [`odb_pack.c`](https://github.com/libgit2/libgit2/blob/master/src/odb_pack.c) would need to be modified so that the `multi-pack-index`'s index is consulted first. Only if an OID is not found there, it should fall back to the current behavior of iterating over all the `.pack` files, although [`packfile_load__cb()`](https://github.com/libgit2/libgit2/blob/ee3307a183e39d602b25fa94831c6fc09e7c1b61/src/odb_pack.c#L197-L230) should be modified to ignore any files that are indexed in `multi-pack-index` (to avoid degenerating back into O(_m_ log _n/m_) complexity).

This can be split in four chunks for easier reviewing:

* [x] Support for parsing `multi-pack-index` files, and a fuzzer because parsing is hard. ~~#5401~~ merged
* [x] Support for reading existing `multi-pack-index` files. ~~#5403~~ merged
* [x] Support for creating a new `multi-pack-index` file from a set of `.pack` files. #5404
* [x] Support for creating / updating a `multi-pack-index` file from an open repository. #5405

Relevant documentation:

* https://github.com/git/git/blob/master/Documentation/technical/multi-pack-index.txt
* https://github.com/git/git/blob/master/Documentation/technical/pack-format.txt

lhchavez:
Part 1 should be out for review within ~a week.

pks-t:
Quick note so you don't feel ignored: I'll start reviewing your code as soon as we've released v1.0, which should be soonish. Looking forward to having multipack index support!

sluongng:
what is the status for this issue?

I see active commits but I'm not quite sure regarding the progress.

@lhchavez do you need any help? should we split the issue/PRs into even smaller parts?

lhchavez:
> what is the status for this issue?

[...]

PR #5401, opened 2020-02-15
midx: Introduce a parser for multi-pack-index files

lhchavez:
This change is the first in a series to add support for git's
multi-pack-index. This should speed up large repositories significantly.

Part of: #5399


src/multipack.h
```diff
@@ -0,0 +1,77 @@
...
+
+/*
+ * An entry within the Object Offset table.
+ */
+struct git_multipack_index_object_offset {
+	/* The index of the packfile name within the Packfile Names table. */
+	uint32_t pack_int_id;
+	/* The offset in bytes within the packfile for the object. */
+	uint32_t offset;
+} __attribute__((packed));
```
lhchavez: argh, forgot this was a gcc thing. i'll remove it.

ethomson:
Instead of strictly reverting the `sha1_lookup` change, let's leave it in `pack.c`, that way we don't have two additional files in the tree for a single function.  We can provide it internally for the library and name it `git_pack__lookup_sha1` or something.

lhchavez:
> Instead of strictly reverting the `sha1_lookup` change, let's leave it in `pack.c`, that way we don't have two additional files in the tree for a single function. We can provide it internally for the library and name it `git_pack__lookup_sha1` or something.

good point! done! (also took the opportunity to add a comment to the function and rename the arguments).

ethomson:
Sorry about the delay here.  Let's move forward with this.  Would you mind rebasing this on master so that we can get an up-to-date CI pass?

lhchavez:
> Sorry about the delay here. Let's move forward with this. Would you mind rebasing this on master so that we can get an up-to-date CI pass?

Done!

ethomson:
One thought that I have is that maybe this should be called `git_midx` instead of `git_multipack_index`.  a) `midx` is significantly shorter 😀  but b) the multipack index is the atomic unit here, if I understand correctly.  `git_multipack_index` makes it sound like there's an index on an atomic unit called a `git_multipack`.  But there is not.

Since this is an internal-only type, I don't necessarily think that it's a hard and fast requirement to make this change, but I think that the update would be an improvement.  What do you think @lhchavez?

lhchavez:
> One thought that I have is that maybe this should be called `git_midx` instead of `git_multipack_index`. a) `midx` is significantly shorter  but b) the multipack index is the atomic unit here, if I understand correctly. `git_multipack_index` makes it sound like there's an index on an atomic unit called a `git_multipack`. But there is not.
> 
> Since this is an internal-only type, I don't necessarily think that it's a hard and fast requirement to make this change, but I think that the update would be an improvement. What do you think @lhchavez?

sounds reasonable. Done!

ethomson:
Cooooool.  I'm excited to see this.  I'm going to chip away at your other PRs, I apologize for making you wait so long in the review queue - I haven't had much spare time in the last few months.

ethomson (approved):

PR #5403, opened 2020-02-17
midx: Support multi-pack-index files in odb_pack.c

lhchavez:
This change adds support for reading multi-pack-index files from the
packfile odb backend. This also makes git_pack_file objects open their
backing failes lazily in more scenarios, since the multi-pack-index can
avoid having to open them in some cases (yay!).

This change also refreshes the documentation found in src/odb_pack.c to
match the updated code.

Part of: #5399

### DIFF
--- a/src/midx.c
+++ b/src/midx.c
@@ -114,7 +114,7 @@ static int midx_parse_oid_lookup(
 		return midx_error("missing OID Lookup chunk");
 	if (chunk_oid_lookup->length == 0)
 		return midx_error("empty OID Lookup chunk");
-	if (chunk_oid_lookup->length != idx->num_objects * 20)
+	if (chunk_oid_lookup->length != idx->num_objects * GIT_OID_RAWSZ)
 		return midx_error("OID Lookup chunk has wrong length");
 
 	idx->oid_lookup = oid = (git_oid *)(data + chunk_oid_lookup->offset);
@@ -181,7 +181,7 @@ int git_midx_parse(
 
 	assert(idx);
 
-	if (size < sizeof(struct git_midx_header) + 20)
+	if (size < sizeof(struct git_midx_header) + GIT_OID_RAWSZ)
 		return midx_error("multi-pack index is too short");
 
 	hdr = ((struct git_midx_header *)data);
@@ -201,7 +201,7 @@ int git_midx_parse(
 	last_chunk_offset =
 			sizeof(struct git_midx_header) +
 			(1 + hdr->chunks) * 12;
-	trailer_offset = size - 20;
+	trailer_offset = size - GIT_OID_RAWSZ;
 	if (trailer_offset < last_chunk_offset)
 		return midx_error("wrong index size");
 	git_oid_cpy(&idx->checksum, (git_oid *)(data + trailer_offset));
@@ -353,7 +353,7 @@ bool git_midx_needs_refresh(
 		return true;
 	}
 
-	if (p_lseek(fd, -20, SEEK_END) < 0) {
+	if (p_lseek(fd, -GIT_OID_RAWSZ, SEEK_END) < 0) {
 		p_close(fd);
 		return true;
 	}
@@ -361,7 +361,7 @@ bool git_midx_needs_refresh(
 	bytes_read = p_read(fd, &idx_checksum, GIT_OID_RAWSZ);
 	p_close(fd);
 
-	if (bytes_read < 0)
+	if (bytes_read != GIT_OID_RAWSZ)
 		return true;
 
 	return git_oid_cmp(&idx_checksum, &idx->checksum) == 0;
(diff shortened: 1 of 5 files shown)


lhchavez:
This contains the same commit as #5401 for easier reviewing, but don't know of any way of telling GitHub to use a PR as a diffbase for another PR.

src/midx.c
```diff
@@ -325,6 +327,46 @@ int git_midx_open(
...
+		p_close(fd);
+		return true;
+	}
+
+	if (p_lseek(fd, -20, SEEK_END) < 0) {
+		p_close(fd);
+		return true;
+	}
+
+	bytes_read = p_read(fd, &idx_checksum, GIT_OID_RAWSZ);
```
ethomson, line 361: Are there any cases where `lseek` could back up `-20` from the EOF but then we couldn't read those 20 bytes?  In that case `idx_checksum` might not be fully read and we might be comparing the expected checksum against the uninitialized memory which might give a false positive.

I don't think that this is really possible, I'm mostly just thinking out loud.

One thing that did jump out at me here is that we're using `git_oid`s and `GIT_OID_RAWSZ` but `lseek`ing `-20`.  Maybe we should `lseek` (`0 - GIT_OID_RAWSZ`).  I don't think that we'll ever change that but with the forthcoming sha256 work, I can imagine a world where oids and sizes are a little more complicated, so having that explicitly called out here might be useful?  Then again, there are likely a million places in our codebase where we assume `20`, so...  maybe not worth it until we stumble across it.  Again, mostly just thinking out loud.

src/midx.c
```diff
@@ -399,6 +441,23 @@ int git_midx_entry_find(
...
+		git_midx_file *idx,
+		git_odb_foreach_cb cb,
+		void *data)
+{
+	size_t i;
+	int error;
+
+	assert(idx);
+
+	for (i = 0; i < idx->num_objects; ++i)
```
ethomson, line 454: Stylistically, I think that we prefer braces around branching statements, treating an `if` as an implicit multi-line statement.  (Not sure if I'm describing this well.)  eg:

```
for (i = 0; i < idx->num_objects; ++i) {
    if ((error = cb(&idx->oid_lookup[i], data)) != 0)
        return git_error_set_after_callback(error);
}
```

I say "I think" because there are a handful of stylistic grey areas, and I think that this is one of them.  Maybe this should be better expressed as "it's my opinion that ...".

No need to change this - again, grey area, perfectly reasonable - I just thought that I'd mention it.

ethomson:
This looks good to me, thanks for breaking this up into easily reviewable pieces @lhchavez.  I'd like to do a little more reading to catch myself up on midx, it's been a while since I've thought about it.  But I think that this is ready to go, I just want to make sure that I understand everything that's coming a little bit better.  😀

src/midx.c
lhchavez: Let's change it, I needed to push a new change anyways.

src/midx.c
lhchavez: > Are there any cases where `lseek` could back up `-20` from the EOF but then we couldn't read those 20 bytes? In that case `idx_checksum` might not be fully read and we might be comparing the expected checksum against the uninitialized memory which might give a false positive.

Hmm without the file being modified? I don't think so, unless there's a freak coincidence where the last 20 bytes being in an unreadable sector or a corrupted block or something similarly implausible. the good thing is that the memory is initialized to 0 in L337! but it may be a good idea to change the condition in L364 to be `!= GIT_OID_RAWSZ`.

probably the "correct" thing to do would be to introduce support for [`pread(2)`](https://man7.org/linux/man-pages/man2/pread.2.html) (#5583 comes to mind) to avoid this seek-then-read shenanigans.

> I don't think that this is really possible, I'm mostly just thinking out loud.
> 
> One thing that did jump out at me here is that we're using `git_oid`s and `GIT_OID_RAWSZ` but `lseek`ing `-20`. Maybe we should `lseek` (`0 - GIT_OID_RAWSZ`). I don't think that we'll ever change that but with the forthcoming sha256 work, I can imagine a world where oids and sizes are a little more complicated, so having that explicitly called out here might be useful? Then again, there are likely a million places in our codebase where we assume `20`, so... maybe not worth it until we stumble across it. Again, mostly just thinking out loud.

ah that's a good idea!

lhchavez:
> This looks good to me, thanks for breaking this up into easily reviewable pieces @lhchavez. I'd like to do a little more reading to catch myself up on midx, it's been a while since I've thought about it. But I think that this is ready to go, I just want to make sure that I understand everything that's coming a little bit better.

lmk if there's anything i can do to help with that learning process.

ethomson:
Thanks for this - I merged it manually to squash your fixups in, and to handle the conflicts that I introduced with the assert handling in #5327

PR #5404, opened 2020-02-18
midx: Add a way to write multi-pack-index files

lhchavez:
This change adds the git_midx_writer_* functions to allow to
write and create `multi-pack-index` files from `.idx`/`.pack` files.
    
Part of: #5399


ethomson:
I'm surprised that we write the midx to a `git_buf` and then write the `git_buf` to disk.  I expected us to not use an intermediate `git_buf`.  How big is a midx file going to be?

lhchavez:
> I'm surprised that we write the midx to a `git_buf` and then write the `git_buf` to disk. I expected us to not use an intermediate `git_buf`. How big is a midx file going to be?

as large as all the .pack files in a repo combined (my copy of linux' is 1G).

IIRC I did that for simplicity, but i can refactor things so that `git_midx_writer_dump` and `git_midx_writer_commit` both call the same function that creates the midx and call a `write_cb(void *buf, size_t size, void *payload)` to avoid that extra in-memory copy.

ethomson:
> IIRC I did that for simplicity, but i can refactor things so that git_midx_writer_dump and git_midx_writer_commit both call the same function that creates the midx and call a write_cb(const char *buf, size_t size, void *payload) to avoid that extra in-memory copy.

I think that would be nice if you could - I think that midx is going to be used in large repositories, so avoiding that malloc of a contiguous hunk of memory to store the data would be a win.

lhchavez:
> > IIRC I did that for simplicity, but i can refactor things so that git_midx_writer_dump and git_midx_writer_commit both call the same function that creates the midx and call a write_cb(const char *buf, size_t size, void *payload) to avoid that extra in-memory copy.
> 
> I think that would be nice if you could - I think that midx is going to be used in large repositories, so avoiding that malloc of a contiguous hunk of memory to store the data would be a win.

refactored and rebased!

src/midx.c
```diff
@@ -475,3 +481,393 @@ void git_midx_free(git_midx_file *idx)
...
+		return error;
+	cb_data = &hash_cb_data;
+	write_cb = midx_write_hash;
+
+	git_vector_sort(&w->packs);
+	git_vector_foreach (&w->packs, i, p) {
+		git_buf relative_index = GIT_BUF_INIT;
+		struct object_entry_cb_state state = {
+				.pack_index = (uint32_t)i,
+				.object_entries_array = &object_entries_array,
```
ethomson, line 682: Also here.

PR #5406, opened 2020-02-18
azure: fix ARM32 builds by replacing gosu(1)

pks-t:
Our nightly builds are currently failing due to our ARM-based jobs.
These jobs crash immediately when entering the Docker container with a
exception thrown by Go's language runtime. As we're able to successfully
builds the Docker images in previous steps, it's unlikely to be a bug in
Docker itself. Instead, this exception is thrown by gosu(1), which is a
Go-based utility to drop privileges and run by our entrypoint.

Fix the issue by dropping gosu(1) in favor of su(1).


pks-t:
Both our regular and nightly pipelines now succeed, as can be seen by the two checks passing, including ARM

ethomson:
LGTM :shipit:

PR #5408, opened 2020-02-19
README: update our build matrix to reflect current releases

pks-t:
As noted in docs/release.md, we only provide security updates for the
latest two releases. Let's thus drop the build status of both v0.27 and
v0.26 branches, adding the new v0.99 branch instead.


ethomson:
LGTM.  To clarify, I'm really thinking of this as a 1.0 prerelease, so I'm expecting that we'll remove 0.99 and replace it with 1.0 when we do release 1.0.  (Not keeping 0.99 and 1.0 around.)

Does that align with your thinking as well?

PR #5412, opened 2020-02-19
Fix #5410: fix installing libgit2.pc in wrong location

### LINKED ISSUE #5410: 0.99.0: libgit2.pc is not installed in wrong location
As it is in the subject.

It would be good to add install that file in cmake files :)

kloczek:
Remove using custom PKG_BUILD_PREFIu, PKG_BUILD_LIBDIR and
PKG_BUILD_INCLUDEDIR variables.
Use cmake CMAKE_INSTALL_PREFIX, LIB_INSTALL_DIR, INCLUDE_INSTALL_DIR instead.
This patch fixes install libgit2.pc file in correct location and simpifies
cmake module.

When typical rpm package is build is passed typical set of cmakle switches. with base paths:
```
$ rpm -E %cmake | grep -- -D
        -DBUILD_SHARED_LIBS=ON \
        -DCMAKE_AR=${AR} \
        -DCMAKE_C_FLAGS_RELEASE="-DNDEBUG" \
        -DCMAKE_CXX_FLAGS_RELEASE="-DNDEBUG" \
        -DCMAKE_Fortran_FLAGS_RELEASE="-DNDEBUG" \
        -DCMAKE_INSTALL_PREFIX=/usr \
        -DCMAKE_NM=${NM} \
        -DCMAKE_RANLIB=${RANLIB} \
        -DCMAKE_VERBOSE_MAKEFILE=ON \
        -DINCLUDE_INSTALL_DIR=/usr/include \
        -DLIB_INSTALL_DIR=/usr/lib64 \
        -DLIB_SUFFIX=64 \
        -DSHARE_INSTALL_PREFIX=/usr/share \
        -DSYSCONF_INSTALL_DIR=/etc
```
This patch switches to use those paths instead custom one.

### DIFF
--- a/cmake/Modules/PkgBuildConfig.cmake
+++ b/cmake/Modules/PkgBuildConfig.cmake
@@ -1,10 +1,5 @@
 # pkg-config file generation
 #
-# Uses the following globals:
-# - PKG_BUILD_PREFIX: the build location (aka prefix). Defaults to CMAKE_INSTALL_PREFIX
-# - PKG_BUILD_LIBDIR: the libdir location. Defaults to ${prefix}/lib.
-# - PKG_BUILD_INCLUDEDIR: the includedir location. Defaults to ${prefix}/include.
-#
 
 function(pkg_build_config)
     set(options)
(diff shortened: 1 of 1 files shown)


pks-t:
Indeed, thanks for reporting and fixing the issue. Your changes do in fact fix the issue. @tiennou, want to have a look?

ethomson:
I think that this makes sense.  I'm not in love with changing this right before 1.0 though, in case somebody's relying on the status quo for their packaging / build process. 

We okay with merging this after 1.0?

pks-t:
@ethomson: I'd argue this is a regression we've introduced with v0.99, so it should be fixed for v1.0.

ethomson:
> @ethomson: I'd argue this is a regression we've introduced with v0.99, so it should be fixed for v1.0.

In that case, :shipit:

pks-t:
Thanks for your report and the fix, @kloczek!

kloczek:
Thx guys for quick merge :)

PR #5413, opened 2020-02-20
Fix typo on GIT_USE_NEC

csware:


ethomson:
Makes sense.  I propose that since we're locked down for 1.0 that we get _two approvals_ before getting merged into master.  @pks-t ?

pks-t:
> Makes sense. I propose that since we're locked down for 1.0 that we get two approvals before getting merged into master. @pks-t ?

I ain't got much of an opinion on that, but it can't hurt either. So let's do that.

Anyway, the fix looks obviously corerct to me. Thanks, @csware!

PR #5422, opened 2020-02-24
CMake booleans

pks-t:
I've noticed that Gentoo carries a fix in their libgit2-v0.99 ebuild to modify how they pass some CMake build instructions. This is due to us doing a strict "IF (VAR STREQUAL ON)" in some places, which is much too strict and neglects the other boolean values that CMake knows like "on" (case insensitive), "true", "yes", "1".

This PR fixes the four locations I have spotted, which is GSS, HTTPS, SHA1 and tracing. The first two commits are required to fix the regression, the third commit is merely a cleanup. @ethomson, let me know if I should defer the last commit to after v1.0.


ethomson:
Let’s move the streamlining into post-1.0

pks-t:
I've moved the streamlining commit into #5440.

ethomson:
Much magic.

PR #5425, opened 2020-02-26
pack: Improve error handling for get_delta_base()

lhchavez:
This change moves the responsibility of setting the error upon failures
of get_delta_base() to get_delta_base() instead of its callers. That
way, the caller chan always check if the return value is negative and
mark the whole operation as an error instead of using garbage values,
which can lead to crashes if the `.pack` files are malformed.

### DIFF
--- a/src/pack.c
+++ b/src/pack.c
@@ -490,6 +490,12 @@ int git_packfile_resolve_header(
 
 		base_offset = get_delta_base(p, &w_curs, &curpos, type, offset);
 		git_mwindow_close(&w_curs);
+
+		if (base_offset < 0) {
+			error = (int)base_offset;
+			return error;
+		}
+
 		if ((error = git_packfile_stream_open(&stream, p, curpos)) < 0)
 			return error;
 		error = git_delta_read_header_fromstream(&base_size, size_p, &stream);
@@ -508,8 +514,14 @@ int git_packfile_resolve_header(
 			return error;
 		if (type != GIT_OBJECT_OFS_DELTA && type != GIT_OBJECT_REF_DELTA)
 			break;
+
 		base_offset = get_delta_base(p, &w_curs, &curpos, type, base_offset);
 		git_mwindow_close(&w_curs);
+
+		if (base_offset < 0) {
+			error = (int)base_offset;
+			return error;
+		}
 	}
 	*type_p = type;
 
@@ -586,10 +598,6 @@ static int pack_dependency_chain(git_dependency_chain *chain_out,
 		base_offset = get_delta_base(p, &w_curs, &curpos, type, obj_offset);
 		git_mwindow_close(&w_curs);
 
-		if (base_offset == 0) {
-			error = packfile_error("delta offset is zero");
-			goto on_error;
-		}
 		if (base_offset < 0) { /* must actually be an error code */
 			error = (int)base_offset;
 			goto on_error;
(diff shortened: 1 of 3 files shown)


lhchavez:
note that the build issues are caused by mbedtls not being able to be downloaded, and are unrelated to this change: https://dev.azure.com/libgit2/libgit2/_build/results?buildId=3441&view=logs&j=bfcc9059-3690-5047-368c-49dfe7d9b8d3&t=8167bc0c-7141-585d-94fc-b6d47d2e4c77&l=1170

pks-t:
Seems like certificate for https://tls.mbed.org has expired or is somehow invalid now. Let's wait a bit and if the problem persists we could just add "-k" to curl for now.

@ethomson: we started caching Docker images, but we still build them on every run, right? Is this expected or are we somehow invalidating caches?

Anyway, back the proposed changes, they do look like a welcome improvement to me. Does this fix a bug you've experienced? Otherwise I'd like to defer this until after v1.0 due to the current code freeze.

lhchavez:
yes, this fixes a crash that can happen if the .pack files are somehow mangled. with this change, it nicely errors out instead of crashing.

the good thing is that the only garbage values that can come out of this are numbers in the range [-10, 0], so i couldn't think of ways in which this could escalate into arbitrary code execution or attacker-controlled reads.

lhchavez:
tests now pass \o/

ethomson:
> @ethomson: we started caching Docker images, but we still build them on every run, right? Is this expected or are we somehow invalidating caches?

We cache some of the layers; our containers are too big to cache them all.  So we get some speedup here but not as much as if we could cache the entire thing.

src/pack.c
```diff
@@ -490,6 +490,12 @@ int git_packfile_resolve_header(
 
 		base_offset = get_delta_base(p, &w_curs, &curpos, type, offset);
 		git_mwindow_close(&w_curs);
+
+		if (base_offset < 0) {
+			error = (int)base_offset;
+			return error;
+		}
```
pks-t, line 497: These are a bit awkward and don't really fit into our greater error handling patterns. Would you mind updating `get_delta_base` to return the base offset via an out parameter and have the direct return value indicate errors, only? Let me know if that exceeds the time you want to invest here

lhchavez: I would not mind! I also found these a bit awkward and I was aiming for the change with the smallest possible delta, so I thought that a refactor in the same change was not ideal.

But if that's a possibility, I would also prefer to do the cleanup now.

Done!

pks-t:
> I would not mind! I also found these a bit awkward and I was aiming
> for the change with the smallest possible delta, so I thought that a
> refactor in the same change was not ideal.

Now that we've got v1.0.0 released we can be less strict about that
again.

> But if that's a possibility, I would also prefer to do the cleanup
> now.
> 
> Done!

Cool, thanks a lot. I'll review your changes soonish.

Automobileservices4:
#5425

src/pack.c
```diff
@@ -583,17 +591,11 @@ static int pack_dependency_chain(git_dependency_chain *chain_out,
...
 			break;
 
-		base_offset = get_delta_base(p, &w_curs, &curpos, type, obj_offset);
+		error = get_delta_base(&base_offset, p, &w_curs, &curpos, type, obj_offset);
 		git_mwindow_close(&w_curs);
 
-		if (base_offset == 0) {
-			error = packfile_error("delta offset is zero");
-			goto on_error;
-		}
```
pks-t, line 592: You're removing this error case here, but as far as I can see it's not been added to `get_delta_base`.

lhchavez: Done. Not sure if it's needed in both places that I added. I would assume `git_oidmap_get` would not store an invalid entry, so maybe I am being paranoid.

pks-t: Better be defensive while programming than running into errors because of taking too many things for granted :+1:

Automobileservices4:
#5425

PR #5417, opened 2020-02-21
deps: ntlmclient: fix missing htonll symbols on FreeBSD and SunOS

pks-t:
The ntlmclient dependency defines htonll on Linux-based systems, only.
As a result, non-Linux systems will run into compiler and/or linker
errors due to undefined symbols.

Fix this issue for FreeBSD, OpenBSD and SunOS/OpenSolaris by including
the proper headers and defining the symbol accordingly.

---

We should probably merge this upstream first, but given that upstream means @ethomson I've decided to create the PR here first to let people in #5414 verify the fix more easily.

Fixes #5414

PR #5426, opened 2020-02-26
deps: ntlmclient: fix htonll on big endian FreeBSD

pks-t:
In commit 3828ea67b (deps: ntlmclient: fix missing htonll symbols on
FreeBSD and SunOS, 2020-02-21), we've fixed compilation on BSDs due to
missing `htonll` wrappers. While we are now using `htobe64` for both
Linux and OpenBSD, we decided to use `bswap64` on FreeBSD. While correct
on little endian systems, where we will swap from little- to big-endian,
we will also do the swap on big endian systems. As a result, we do not
use network byte order on such systems.

Fix the issue by using htobe64, as well.

---

@mfechner: could you please confirm that this still works as expected?
@ethomson: this is the follow-up to #5417

### DIFF
--- a/deps/ntlmclient/compat.h
+++ b/deps/ntlmclient/compat.h
@@ -32,7 +32,7 @@
 #elif defined(__FreeBSD__)
 /* See man page bwaps64(9) */
 # include <sys/endian.h>
-# define htonll bswap64
+# define htonll htobe64
 #elif defined(sun) || defined(__sun)
 /* See man page byteorder(3SOCKET) */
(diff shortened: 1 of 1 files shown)


ethomson:
I validated this in my FreeBSD environment, so I'm going to go ahead and merge it.  @mfechner please let us know if this doesn't work.

mfechner:
It compiles fine. But I cannot test if the libs works fine.

pks-t:
Cool, thanks for confirming, @mfechner!

PR #5429, opened 2020-02-28
win32: fix path canonicalization for paths starting with "../"

### LINKED ISSUE #5424: Windows: symbolic links with .. are not created correctly
When cloning a directory with symbolic links that contain `..` on Windows, the `..` segments are silently discarded, resulting in a broken symlink.

Note that this was recently changed in #5355.  cc @pks-t

### Reproduction steps
1. Create a repository with a file symbolic link between directories. For example: 
   - `echo contents > foo`
   - `mkdir somedir`
   - `mklink somedir\foo ..\foo`
   - Init a repo, and add these two files
2. Have a libgit2 project which will clone the sample repo.
3. Notice that the resulting symlink is "foo -> foo"

### Expected behavior

`..` should be preserved in the symlink.

### Actual behavior

`..` is removed from symlink.

172239021f7ba04fe7327647b213799853a9eb89

Windows 10

pks-t:
The path canonicalization function for Windows strips leading "../".
This behaviour has been introduced with commit cceae9a25 (win32: use
NT-prefixed "\\?\" paths, 2014-12-01), saying that

> we must not rely on the system to translate "." and ".." in paths

The reasoning is that we mostly try to use NT-prefixed "\\?\" paths,
which does turn off path canonicalization. What's weird though is that
if we simply strip leading "..", then there is no way that we can end up
with the correct path, as we do not adjust any of the following path
components accordingly.

One place where this bites us is with the path canonicalization
for symlink targets introduced with commit 7d55bee6d (win32: fix
relative symlinks pointing into dirs, 2020-01-10). If creating a symlink
with target "../target", then the canonicalization function will strip
the leading "../" and create a symlink with target "target".

Given that stripping this prefix cannot be correct in general, let's
remove the stripping code and instead leave such prefixes intact. If
understanding the code correctly, then the change _should_ fix cases as
with the symlink while not breaking anything else.

---

@ethomson: this is a regression in v0.99, so we should fix it for v1.0. I'm reeeeeeally unsure about the fix, though, so please review it with extra scrutiny. You might also know more about the original code as you're the author ;)

Fixes #5424

### DIFF
--- a/src/win32/path_w32.c
+++ b/src/win32/path_w32.c
@@ -110,11 +110,10 @@ int git_win32_path_canonicalize(git_win32_path path)
 
 		else if (len == 2 && from[0] == L'.' && from[1] == L'.') {
 			if (to == base) {
-				/* no more path segments to strip, eat the "../" */
+				/* Skip over leading "../" */
 				if (*next == L'\\')
 					len++;
-
-				base = to;
+				to += len;
 			} else {
 				/* back up a path segment */
 				while (to > base && to[-1] == L'\\') to--;
--- a/tests/core/posix.c
+++ b/tests/core/posix.c
@@ -190,6 +190,26 @@ void test_core_posix__symlink_resolves_to_correct_type(void)
 	git_buf_dispose(&contents);
 }
 
+void test_core_posix__relative_symlink(void)
+{
+       git_buf contents = GIT_BUF_INIT;
+
+       if (!git_path_supports_symlinks(clar_sandbox_path()))
+               clar__skip();
+
+       cl_must_pass(git_futils_mkdir("dir", 0777, 0));
+       cl_git_mkfile("file", "contents");
+       cl_git_pass(p_symlink("../file", "dir/link"));
+       cl_git_pass(git_futils_readbuffer(&contents, "dir/link"));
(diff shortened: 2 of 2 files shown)


src/win32/path_w32.c
```diff
@@ -110,11 +110,10 @@ int git_win32_path_canonicalize(git_win32_path path)
 
 		else if (len == 2 && from[0] == L'.' && from[1] == L'.') {
 			if (to == base) {
-				/* no more path segments to strip, eat the "../" */
+				/* Skip over leading "../" */
```
ethomson: This isn't a _leading_ `../`, though, at least not always.

`base` is the results of `path_skip_prefix`, which will skip over things that we never touch - like drive prefixes and UNC share prefixes.

So if we tried to leave these in place, then we'd leave them when they exist immediately behind the drive prefixes.

For example, `C:\..\foo\bar` _should_ be canonicalized as `C:\foo\bar` (just like you would canonicalize `/../foo/bar` as `/foo/bar`).  That's the goal of this bit.

ethomson: I think that we want to #5435

ethomson:
Fixed via #5435

pks-t:
Thanks, @ethomson!

PR #5433, opened 2020-03-01
azure-pipelines: download GlobalSign's certificate manually

ethomson:
[tls.mbed.org](tls.mbed.org) is not providing their [full certificate chain](https://www.ssllabs.com/ssltest/analyze.html?d=tls.mbed.org).  Download their intermediate cert and use it (and the normal cacert list) during the mbedTLS download.


ethomson:
Even though this is a dirty hack, I'm merging this to unblock us.  Looking forward to reverting this 🔜.

PR #5435, opened 2020-03-02
win32: don't canonicalize relative paths

ethomson:
Our path canonicalization (beginning in  cceae9a) was meant to deal with absolute paths, converting things like `C:\Foo\Bar` into `\\?\C:\Foo\Bar` so that we could avoid some of the dangers of NTFS (like eliding trailing dots in filenames).  But by using those paths, we need to do canonicalization ourselves, as giving `\\?\C:\Foo\..\Bar` to the filesystem APIs will return an error instead of resolving the path.

These canonicalization function were not meant to deal with relative paths.  In particular, this means that leading ".." segments at the beginning of paths were removed, to translate `C:\..\..\..\` into the correct `\\?\C:\`.  (This is despite having some tests around relative path components, they were not well thought out since they were not intended to be used.)

Update the documentation on the function to point out that it's not to be used for relative paths, and remove the tests that suggest that it should be used that way.

Add a new function that will keep relative paths relative, while converting them to UTF-16 and translating `/` to `\\`.

Use this new relative path handling function when creating Windows symlinks.

### DIFF
--- a/src/win32/path_w32.h
+++ b/src/win32/path_w32.h
@@ -11,13 +11,25 @@
 #include "vector.h"
 
 /**
- * Create a Win32 path (in UCS-2 format) from a UTF-8 string.
+ * Create a Win32 path (in UCS-2 format) from a UTF-8 string.  If the given
+ * path is relative, then it will be turned into an absolute path by having
+ * the current working directory prepended.
  *
  * @param dest The buffer to receive the wide string.
  * @param src The UTF-8 string to convert.
  * @return The length of the wide string, in characters (not counting the NULL terminator), or < 0 for failure
  */
 extern int git_win32_path_from_utf8(git_win32_path dest, const char *src);
+
+/**
+ * Create a Win32 path (in UCS-2 format) from a UTF-8 string.  If the given
+ * path is relative, then it will not be turned into an absolute path.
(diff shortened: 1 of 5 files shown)


ethomson:
Note that it feels a little weird to jam relative path handling into this.  I considered splitting it out into two different functions, especially since this function is a bit hairy already.  But in the end I just modified this one instead of trying to duplicate hairy code.

src/win32/path_w32.h
```diff
@@ -11,21 +11,36 @@
...
 
+/**
+ * Create a Win32 path (in UCS-2 format) from a UTF-8 string.  If the given
+ * path is relative, then it will not be turned into an absolute path.
+ *
+ * @param dest The buffer to receive the wide string.
+ * @param src The UTF-8 string to convert.
+ * @return The length of the wide string, in characters (not counting the NULL terminator), or < 0 for failure
+ */
+extern int git_win32_path_relative_from_utf8(git_win32_path dest, const char* src);
```
pks-t: Sorry, but minor nit: the `*` is with the name, not with the type ;)

ethomson: oops!  fixed, thanks!

PR #5437, opened 2020-03-02
ci: provide globalsign certs for bionic

ethomson:
tls.mbed.org has neglected to send their full certificate chain.  Add their intermediate cert manually.  🙄


ethomson:
Going ahead here, too, to get our nightlies 💚

Issue #5438, opened 2020-03-03
libgit2 when built with pcre2 regex engine has broken pkg-config dependency


ignatenkobrain:
I have updated libgit2 to 0.99 in Fedora and built it with pcre2 regex engine, but then we get broken dependencies....

```
DEBUG util.py:596:    - nothing provides pkgconfig(libpcre2) needed by libgit2-devel-0.99.0-1.fc33.x86_64
```

ignatenkobrain:
The pcre2 in Fedora provides following pkg-config files:

```
pkgconfig(libpcre2-16) = 10.34
pkgconfig(libpcre2-32) = 10.34
pkgconfig(libpcre2-8) = 10.34
pkgconfig(libpcre2-posix) = 10.34
```

PR #5439, opened 2020-03-03
Set proper pkg-config dependency for pcre2

ignatenkobrain:
Fixes: https://github.com/libgit2/libgit2/issues/5438


ethomson:
Thanks @ignatenkobrain for the fix and @pks-t for reviewing it.

PR #5440, opened 2020-03-05
CMake: backend selection streamlining

pks-t:
This is basically #5422 but with the streamlining commit on top that we've decided to defer to after v1.0.


pks-t:
Rebased to drop upstreamed commits.

pks-t:
Rebased to fix conflicts.

Issue #5443, opened 2020-03-07
Segfault when calling git_blame_buffer


xldenis:
Hi! 

I managed to trigger a segfault when running git blame buffer. 

It seems like if a blame buffer is run on a file with a single hunk then the `line_cb` is called before the `hunk_cb` and tries to dereference a null `current_hunk` pointer. 

If I add more lines in a second commit then the problem goes away.

I can't follow the the actual process of blaming a buffer, I get lost with all the callbacks etc.. However, I made a minimal testcase that causes this segfault and added it to a fork of the repo.
 
### Reproduction steps

Using my branch at github.com/xldenis/libgit2

Build and then run the test:

 `cmake -sblame::buffer::segfault`

### Expected behavior

I expect it to return a blame result with two hunks

### Actual behavior

segfault :(

### Version of libgit2 (release number or SHA1)

cb17630b75dbdc502c21da8e73641fa932bc8c6b

### Operating system(s) tested

OS X

pks-t:
Thanks a lot for your report! Could you please verify whether #5445 fixes your problem?

xldenis:
I'll check it out right now.

xldenis:
My test passes with your commit.

PR #5445, opened 2020-03-08
Fix segfault when calling git_blame_buffer()

lhchavez:
This change makes sure that the hunk is not null before trying to
dereference it. This avoids segfaults, especially when blaming against a
modified buffer (i.e. the index).

Fixes: #5443


pks-t:
I've rebased your branch on top of current master to give CI another try.

tests/blame/buffer.c
```diff
@@ -17,6 +17,28 @@ void test_blame_buffer__cleanup(void)
 	git_repository_free(g_repo);
 }
 
+void test_blame_buffer__index(void)
+{
+	const git_blame_hunk *hunk;
+	const char *buffer = "Hello\nWorld!";
+
+	git_blame_free(g_fileblame);
+	g_fileblame = NULL;
```
pks-t: These two lines shouldn't be needed, we already clean up `g_fileblame` in `test_blame_buffer__cleanup`.

lhchavez: ASan complains if I remove them. this is because this is executed after `test_blame_buffer__initialize()`, so if I remove them, I would be leaking some stuff from the previously opened `g_fileblame`.

let me add a comment to avoid confusion.

lhchavez: Comment added!

PR #5447, opened 2020-03-09
odb: Add git_odb_options

lhchavez:
This change introduces the git_odb_options struct and a bunch of
`*_ext()` functions that consumes it. This does nothing for now, but
paves the way to allow callers to choose whether or not to enable the
multi-pack-index feature, similar to what vanilla git does.

Part of: #5399


ethomson:
Interesting.  I had assumed that a multi-pack index enabled repository would have a new odb backend type, and that we'd have a loose store, pack store (where there's a 1:1 pack/index mapping) and a multi-pack index store (where there's a many:1 pack/index mapping).

lhchavez:
it was a lot easier to implement by making the packfile backend also know about midx: there are scenarios where  you want to have a midx and some stray packfiles because you only want to regenerate the midx file once a week (it's cheaper than a full gc, but it still costs some time). by splitting that into its own backend, a communication between both should be set up OR in cases where objects are not found in the odb at all, traversal through every single packfile would happen twice, which makes it even worse D:

PR #5451, opened 2020-03-10
azure: fix errors due to curl and removal of old VM images

pks-t:
We currently pass the "--silent" flag to most invocations of curl, but
in fact this does not only suppress the progress meter, but also any
errors. So let's also pass "--show-error", too.


pks-t:
Btw, this PR here was triggered by recent Linux failures. I didn't know what's going on because "--silent" actually swallowed all errors. Funny enough, now Linux builds work but all the others generate errors:

> ##[error]This job was temporarily blocked because it uses a Microsoft hosted agent (MacOS-10.13) that will be permanently removed on March 23rd, 2020.  See https://aka.ms/blocked-hosted-agent for more information.
> ,##[error]The remote provider was unable to process the request.

This shows for all our macOS and Windows jobs. Seems like we'll have to update our images.

pks-t:
Yeez. The link does mention the supposed replacements, but as far as I can see there is no macOS available yet. Either I'm just blind or this forced migration to something that doesn't exist is... suboptimal. @ethomson, you got any clue?

pks-t:
Never mind, been blind. The first link on that shortlink was pointing to GitHub Actions page, which said the macOS images are not available. The other one linked to Microsofts blog post, which does list macOS images.

pks-t:
Also: our documentation generation fails now because either Docurium or rugged have upgraded their minimum CMake requirements to v3.5 :(

pks-t:
And another thing. The Secure Channel API we're currently using has been deprecated. Due to code freeze I propose to only disable deprecation warnings for now and tackle this problem at a later point.

pks-t:
Another fix is required in ntlmclient because of deprecated CommonCrypto functions.

pks-t:
I only wanted to fix how we invoke curl... :sob:

pks-t:
Finally green. We should get this merged fast to unblock our CI.

azure-pipelines.yml
```diff
@@ -89,31 +89,34 @@ jobs:
...
-  pool: Hosted
+  pool:
+    vmImage: 'windows-2019'
   steps:
   - template: azure-pipelines/bash.yml
     parameters:
       environmentVariables:
-        CMAKE_GENERATOR: Visual Studio 12 2013
-        CMAKE_OPTIONS: -DMSVC_CRTDBG=ON -DDEPRECATE_HARD=ON -DUSE_SHA1=HTTPS
+        CMAKE_GENERATOR: Visual Studio 16 2019
```
ethomson: should we move to `vs2017-win2016` instead?  i sort of like running on the oldest version of windows available.

azure-pipelines/coverity.yml
```diff
@@ -5,7 +5,7 @@ jobs:
 - job: coverity
   displayName: 'Coverity'
   pool:
-    vmImage: 'Ubuntu 16.04'
+    vmImage: 'ubunt-18.04'
```
ethomson: typo, should be `ubuntu`

pks-t:
On Tue, Mar 10, 2020 at 03:40:35PM -0700, Edward Thomson wrote:
> 
> 
> 
> >    steps:
>    - template: azure-pipelines/bash.yml
>      parameters:
>        environmentVariables:
> -        CMAKE_GENERATOR: Visual Studio 12 2013
> -        CMAKE_OPTIONS: -DMSVC_CRTDBG=ON -DDEPRECATE_HARD=ON -DUSE_SHA1=HTTPS
> +        CMAKE_GENERATOR: Visual Studio 16 2019
> 
> should we move to `vs2017-win2016` instead?  i sort of like running on
> the oldest version of windows available.

I did that at first, but then thought that running on the newest
platform available to be a good thing. My reasoning is mostly that newer
versions of MSVS will introduce newer analytics into the compiler
toolchain that let us discover issues more quickly.

I can see why we'd want to opt for the oldest version available, though,
in case we accidentally start using APIs that aren't there yet in older
versions. So I'll change it.

pks-t:
I noticed now that I was able to fix everything that wasn't the original error, that the original error showed up again. @ethomson, what do you think about just passing "--insecure" to curl for the time being? It completely disables any certificate checks, but I'm not sure I care for the time being to unblock CI.

pks-t:
I went ahead and added the "--insecure" flag. Let's see if that fixes it -- I'm definitely open to another solution, though

ethomson:
I just checked the mbedTLS certificate, and it looks like they're delivering the full chain.  I pushed a change to remove `--insecure` to make sure that it works.  🤞

ethomson:
OK, CI looks 👍 with this change.  :shipit: ?

pks-t:
On Tue, Mar 17, 2020 at 03:30:27AM -0700, Edward Thomson wrote:
> OK, CI looks 👍 with this change.  :shipit: ?

:+1:

Issue #5453, opened 2020-03-12
pkg-config file contains invalid relative paths


mgorny:
### Reproduction steps

Build without passing `INCLUDE_INSTALL_DIR` or `LIB_INSTALL_DIR` to CMake.

### Expected behavior

pkg-config file containing correct `-L` and `-I` arguments.

### Actual behavior

pkg-config file has:

```
prefix="/usr/local"
libdir="lib"
includedir="include"

Name: libgit2
Description: The git library, take 2
Version: 0.99.0
Libs: -L${libdir} -lgit2
Libs.private: -lrt -lpthread -lssh2
Requires.private: openssl libpcre zlib
Cflags: -I${includedir} 
```

Note that this means it produces `-Llib` and `-Iinclude` instead of the combined paths.

### Version of libgit2 (release number or SHA1)

be36db28f64de8ed5ca9870708a56b6668c6c482.

### Operating system(s) tested

Gentoo Linux

mgorny:
My suggestion would be to use standard CMake `GNUInstallDirs` module instead of reinventing the wheel. It automatically provides both relative paths and absolute paths, the latter suitable for using in generated `.pc` files.

mgorny:
Original bug report: https://bugs.gentoo.org/712270

pks-t:
Thanks a lot for your bug report! Could you please verify that #5455 fixes the problem?

pks-t:
Closed via #5455. Please feel free to reopen in case you feel like the error wasn't addressed properly. And thanks again for your report!

PR #5455, opened 2020-03-13
cmake: use install directories provided via GNUInstallDirs

pks-t:
We currently hand-code logic to configure where to install our artifacts
via the `LIB_INSTALL_DIR`, `INCLUDE_INSTALL_DIR` and `BIN_INSTALL_DIR`
variables. This is reinventing the wheel, as CMake already provide a way
to do that via `CMAKE_INSTALL_<DIR>` paths, e.g. `CMAKE_INSTALL_LIB`.
This requires users of libgit2 to know about the discrepancy and will
require special hacks for any build systems that handle these variables
in an automated way. One such example is Gentoo Linux, which sets up
these paths in both the cmake and cmake-utils eclass.

So let's stop doing that: the GNUInstallDirs module handles it in a
better way for us, especially so as the actual values are dependent on
CMAKE_INSTALL_PREFIX. This commit removes our own set of variables and
instead refers users to use the standard ones.

As a second benefit, this commit also fixes our pkgconfig generation to
use the GNUInstallDirs module. We had a bug there where we ignored the
CMAKE_INSTALL_PREFIX when configuring the libdir and includedir keys, so
if libdir was set to "lib64", then libdir would be an invalid path. With
GNUInstallDirs, we can now use `CMAKE_INSTALL_FULL_LIBDIR`, which
handles the prefix for us.

---

This fixes another regression for our pkgconfig files introduces with v0.99. The changes are a bit more on the heavy side, but it is definitely the right thing to do, if you ask me. The old way was just broken and reinventing the wheel for no good use at all, so thanks a lot to @mgorny for pointing out the `GNUInstallDirs` module.


pks-t:
Fixes #5453

cmake/Modules/PkgBuildConfig.cmake
```diff
@@ -1,6 +1,8 @@
 # pkg-config file generation
 #
 
+INCLUDE(GNUInstallDirs)
```
mgorny: I suppose this doesn't hurt but if you swapped the includes in `src/`, this inherit shouldn't be necessary. I think the general idea is to include `GNUInstallDirs` in top CMakeLists since it's a typical project-global module.

cmake/Modules/PkgBuildConfig.cmake
```diff
@@ -27,8 +29,8 @@ function(pkg_build_config)
     # Write .pc "header"
     file(WRITE "${PKGCONFIG_FILE}"
         "prefix=\"${CMAKE_INSTALL_PREFIX}\"\n"
```
mgorny: I don't think `prefix` will ever be used.

pks-t:
On Fri, Mar 13, 2020 at 09:50:35PM -0700, Michał Górny wrote:
> mgorny approved this pull request.
> 
> Thank you. I've left a few minor suggestions inline. Other than that,
> the patch seems to work fine, get correct libdir without extra options
> and install correct pkg-config file.
> 
> > @@ -1,6 +1,8 @@
>  # pkg-config file generation
>  #
>  
> +INCLUDE(GNUInstallDirs)
> 
> I suppose this doesn't hurt but if you swapped the includes in `src/`,
> this inherit shouldn't be necessary. I think the general idea is to
> include `GNUInstallDirs` in top CMakeLists since it's a typical
> project-global module.

Yeah, I did ponder that, too, so let me amend the commit.

> > @@ -27,8 +29,8 @@ function(pkg_build_config)
>      # Write .pc "header"
>      file(WRITE "${PKGCONFIG_FILE}"
>          "prefix=\"${CMAKE_INSTALL_PREFIX}\"\n"
> 
> I don't think `prefix` will ever be used.

Probably not. I guess it's only there in pkgconfig files that use it as
a prefix to the others, e.g. `libdir=${prefix}/lib`. As we don't set it
up that way it's kind of useless. I still want to keep it though, if not
only for the sake of "We've always had it".

Anyway, thanks for your feedback!

ethomson:
Hmm, it looks like this moves our required cmake version forward by quite a bit.

> CMake 3.5.1 or higher is required.  You are running version 3.0.2

pks-t:
On Thu, Mar 19, 2020 at 04:03:02AM -0700, Edward Thomson wrote:
> Hmm, it looks like this moves our required cmake version forward by quite a bit.
> 
> > CMake 3.5.1 or higher is required.  You are running version 3.0.2

That's mighty odd. GNUInstallDirs is definitely available in v3.0.2:

https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html#module:GNUInstallDirs

Does CMake say anything specific about which instruction causes it to
fail? Could it be that you've generated the build dir with a newer CMake
first and then used an older one later?

ethomson:
Hmm.  I don't know what's going on precisely.  I was skimming the output from the Generate Documentation job in our CI.  It's failing to build libgit2 in the context of rugged.  :/

ethomson:
I kicked the build and it seems to be resolved.

pks-t:
On Sat, Mar 21, 2020 at 09:49:42AM -0700, Edward Thomson wrote:
> I kicked the build and it seems to be resolved.

Cool, thanks a lot for checking again!

Issue #5449, opened 2020-03-09
git_refdb_backend_fs does not set refdb backend version


ddevault:
## Reproduction steps

```c
git_refdb_backend backend = {0};
git_repository *repo = /* ... */;

git_refdb_backend_fs(&backend, repo);
assert(backend.version != 0); // FAIL
```

### Expected behavior

Should populate the version field.

Aside: it would be nice if it took a path instead of fetching it out of a `git_repository`, which might not have a path. This forces us to create a git_repository with the default backends, then throw them out later, if we want to manually create a refdb_backend_fs.

### Actual behavior

Version number is left unpopulated.

### Version of libgit2 (release number or SHA1)

e23b8b446522f17eb3329b62201aae7674f06e13

### Operating system(s) tested

Linux

pks-t:
Indeed, thanks for your report! I've fixed that in #5456, could you please verify it fixes your issue?

> Aside: it would be nice if it took a path instead of fetching it out of a git_repository, which might not have a path. This forces us to create a git_repository with the default backends, then throw them out later, if we want to manually create a refdb_backend_fs.

That'd make a lot of sense to me, yeah. I already got quite a lot on my backlog for post v1.0, but this should probably be quite easy to implement. So I'd be happy if somebody picked it up, but feel free to give me a ping after v1.0 is out.

ddevault:
I can confirm that #5456 fixes the immediate problem, but I ran into some more. I'll investigate next week and open additional tickets as necessary, but this particular issue is resolved when #5456 is merged. Thanks!

ddevault:
>I'll investigate next week and open additional tickets as necessary

Investigation concluded, it was my fault - not libgit2.

pks-t:
Fixed via #5456. Thanks again for your report!

PR #5456, opened 2020-03-13
refdb_fs: initialize backend version

pks-t:
While the `git_refdb_backend()` struct has a version, we do not
initialize it correctly when calling `git_refdb_backend_fs()`. Fix this
by adding the call to `git_refdb_init_backend()`.


pks-t:
Fixes #5449

PR #5462, opened 2020-03-25
WIP: Reftables support

pks-t:
So this is a _very_ rough reftables refdb implementation for libgit2. It's main intent is to start a discussion, see what prerequisites we have in libgit2 to properly implement refdbs other than refdb_fs and whether I'm using the reftable API correctly. Thoughts:

- While we theoretically support reference databases other than the filesystem-based one, our support for that is... lacking. We have dozens of places where we assume $GITDIR/HEAD exists, we write references like MERGE_HEAD, ORIG_HEAD manually without consulting the refdb, lack support to create new repos with a different refdb as we create the filesystem structures like $GITDIR/refs and $GITDIR/HEAD directly in repository.c without consulting the refdb at all. We'll have to carefully vet our code to remove all accesses to anything refish outside of the refdb and instead call into the refdb layer.

- It seems to me that reftables currently don't support a few operations that we expect from any given refdb implementation. Most importantly I couldn't figure out how to delete and rename a reflog and how to implement explicit locking (e.g. what the file backend does when creating .git/refs/heads/master.lock).

- I found it quite hard to get into reftables as the interface exposes about quite a lot of the low-level details. This is probably my own fault as I usually tend to dive heads-deep into anything new without looking left and right first. But one resource that would've helped quite a lot is something like high-level examples that represent common git operations. I feel like I've got it right by now, but I'd be happy to get feedback on this to see how wrong I am. 

- Are there any testvectors around? Like a repository that one may embed into libgit2 itself that uses reftables to verify we behave as expected. I couldn't get a hold of any and as it's currently non-trivial to use reftables anywhere I wasn't able to come up with tests yet. As a result, my current code will probably fail very quickly.

- At some places I feel like there's high potential for namespacing issues. E.g. `struct iterator` is highly generic and may trivially clash with any preexisting types. It would be nice to use namespace prefixes for all exported symbols.

- I think it would help if the C files were layed out in a more "standard" library way. Most importantly, separate out the "reftable.h" header file from the rest into its own folder so that one can add it to the include path without having to care about accidentally overriding "tree.h" that's part of libgit2 itself.

- I didn't find any way to override memory allocation functions. I think we've discussed that at the Contributor's summit already.

Anyway, thanks a lot for working on reftables @hanwen, this is really exciting! I'd be happy to get feedback on my current implementation, which basically represents how I think it should work. Chances are high I've got it completely wrong, though ;)


deps/reftable/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+disable_warnings(sign-compare)
+disable_warnings(declaration-after-statement)
+disable_warnings(unused-parameter)
```
pks-t: I don't care too much about sign-comparison and unused parameter warnings, even though it'd be nice to not have to disable them. I feel like the declaration-after-statement warnings got a point, though, to make it C90 compatible like libgit2 aims to be.

src/refdb_reftable.c
```diff
@@ -0,0 +1,513 @@
...
+static int reftable_exists(
+	int *exists,
+	git_refdb_backend *_backend,
+	const char *ref_name)
+{
+	reftable_backend *backend = GIT_CONTAINER_OF(_backend, reftable_backend, parent);
+	struct ref_record ref = { 0 };
+	int error;
+
+	if ((error = stack_read_ref(backend->stack, ref_name, &ref)) < 0)
```
pks-t: I'd probably have to call `stack_reload` everytime we call into the refdb as we don't know whether any concurrent process has modified the reftable and as libgit2 may be a long-running process. Is this going to be a performance hog or do you expect it to work just fine?

src/refdb_reftable.c
```diff
@@ -0,0 +1,513 @@
...
+	it = git__calloc(1, sizeof(*it));
+	GIT_ERROR_CHECK_ALLOC(it);
+	it->parent.next = reftable_iterator_next;
+	it->parent.next_name = reftable_iterator_next_name;
+	it->parent.free = reftable_iterator_free;
+	it->table = table;
+	it->glob = git__strdup(glob);
+	GIT_ERROR_CHECK_ALLOC(it->glob);
+
+	if ((error = merged_table_seek_ref(table, &it->iter, "")) < 0)
```
pks-t: It was a bit hard to discover how to actually get the first entry of a ref table. Documenting this in the different `seek_ref` and `seek_log` calls would probably help.

src/refdb_reftable.c
```diff
@@ -0,0 +1,514 @@
...
+	if (error > 0)
+		return 0;
+
+	ref_record_clear(&record);
+	return 1;
+}
+
+static int reftable_reflog__ensure_log(git_refdb_backend *_backend, const char *name)
+{
+	/* TODO: doesn't make any sense in the context of reftable? */
```
pks-t: I don't think this makes any sense in the context of reftable, but I'd say it is our issue and not reftable's.

src/refdb_reftable.c
```diff
@@ -0,0 +1,514 @@
...
+		if (entry == NULL) {
+			error = -1;
+			goto out;
+		}
+
+		if ((error = reftable_log_fill(&logs[i], entry->committer,
+					       &entry->oid_old, &entry->oid_cur,
+					       NULL, entry->msg)) < 0)
+			goto out;
+	}
```
pks-t: I'll probably need to deduplicate pre-existing reflog entries. With filesystem-based reflog, we're rewriting the complete reflog every time, I think, but with reftables I'll need to figure out which entries are new.

src/refdb_reftable.c
```diff
@@ -0,0 +1,514 @@
...
+out:
+	git__free(logs);
+	return error;
+}
+
+static int reftable_reflog__rename(git_refdb_backend *_backend, const char *old_name, const char *new_name)
+{
+	GIT_UNUSED(_backend);
+	GIT_UNUSED(old_name);
+	GIT_UNUSED(new_name);
```
pks-t: Is there a better way to rename a reflog than to go through all reflog entries and adjust their name?

src/refdb_reftable.c
```diff
@@ -0,0 +1,514 @@
...
+	return 0;
+}
+
+static int reftable_reflog__delete(git_refdb_backend *_backend, const char *name)
+{
+	/* Log deletion not yet supported */
```
pks-t: I think this doesn't work yet, at least `log_record_is_deletion` always returns `false`.

src/refdb_reftable.c
hanwen: It will read the reftable list, which is a small file, so it is relatively cheap. 

However, the reftable instance you hold will give you a consistent snapshot of the refdb at some point in time, so it's up to you to decide whether you really need this. If there is a concurrent process, you'd normally have some sort of synchronization with that process, no?

src/refdb_reftable.c
hanwen: good point, I'll get to that.

src/refdb_reftable.c
hanwen: ack.

src/refdb_reftable.c
hanwen: The spec says


"Contrary to traditional reflog (which is a file), renames are encoded as a combination of ref deletion and ref creation."

but yes, you could delete and create entries too (but you need to compact the whole thing to get rid of traces of the data)

src/refdb_reftable.c
hanwen: I don't understand the question.

deps/reftable/CMakeLists.txt
hanwen: the main code is c90 clean, but I guess you're compiling tests too?

src/refdb_reftable.c
```diff
@@ -0,0 +1,514 @@
...
+		out->new_hash = (unsigned char *) new_id->id;
+	out->ref_name = (char *) reference;
+	out->message = (char *) message;
+	return 0;
+}
+
+static int reftable_writer(struct writer *wr, void *_arg)
+{
+	reftable_writer_arg *arg = (reftable_writer_arg *) _arg;
+	writer_set_limits(wr, arg->index, arg->index + arg->nrefs + arg->nlogs);
```
hanwen: this should just be 

 writer_set_limits(wr, arg->index, arg->index)

the name update_index is misleading. It's better to think of it as "logical timestamp", and all the entries (both log & ref) should be written at the same update_index timestamp. Later, after compaction, this will makes it clear that these entries were written together.

src/refdb_reftable.c
```diff
@@ -0,0 +1,514 @@
...
+		return 0;
+	}
+
+	return GIT_ITEROVER;
+}
+
+static void reftable_iterator_free(git_reference_iterator *_it)
+{
+	reftable_iterator *it = GIT_CONTAINER_OF(_it, reftable_iterator, parent);
+	merged_table_free(it->table);
```
hanwen: I think you don't need to keep a reference to this.

src/refdb_reftable.c
```diff
@@ -0,0 +1,514 @@
...
+
+static int reftable_iterator_next(git_reference **out, git_reference_iterator *_it)
+{
+	reftable_iterator *it = GIT_CONTAINER_OF(_it, reftable_iterator, parent);
+	struct ref_record ref;
+	int error;
+
+	while (1) {
+		if ((error = iterator_next_ref(it->iter, &ref)) < 0)
+			return error;
```
hanwen: shouldn't you translate error spaces somewhere?

src/refdb_reftable.c
pks-t: This is mostly about `git_reflog` semantics in libgit2. When you pass it to the filesystem-based reflog, you need to have the complete reflog history in order to re-write it in full. So with `reftable` I'll need to see which entry is the last one that we've already written to disk and only start writing after that one.

src/refdb_reftable.c
pks-t: Understood, thanks!

src/refdb.c
```diff
@@ -35,27 +35,40 @@ int git_refdb_new(git_refdb **out, git_repository *repo)
...
+		goto out;
+
+	if ((error = git_buf_joinpath(&refs, git_repository_path(repo), "refs")) < 0)
+		goto out;
+
+	/*
+	 * Use filesystem-based backend in case ".git/refs" is a directory,
+	 * otherwise use the reftable backend.
+	 */
+	if (git_path_isfile(refs.ptr)) {
```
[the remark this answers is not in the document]
pks-t: Will do, thanks!

src/refdb_reftable.c
hanwen: ok. In reftable, the log entries are stored newest-first, so a simple seek to the refname will tell you the last entry written.

src/refdb_reftable.c
pks-t: So having the iterator for that merged table alone is sufficient to keep it alive?

src/refdb_reftable.c
pks-t: Indeed, I've got that on my todo list.

pks-t:
> Why do you want explicit locking? The API is written as if you'd be
> doing transactions, i.e. you try to make the change, and if you get a
> LOCK_ERROR, you'll have to try again.

Because we're kind of in a special place with libgit2. With filesystem based refs, you can explicitly lock references to keep other applications from updating them by themselves. As libgit2 may be part of a long-running application (e.g. an IDE, but it could be anything) where the author of that application may have the need to lock references for a longer time, we support a transactional API that keeps other Git applications from updating them. While it's true that you would get an error anyway with the reftable API, this kind of opportunistic locking wouldn't help in such a usecase.

Thanks for your answers, I'll make sure to update the PR soonish with your feedback!

deps/reftable/CMakeLists.txt
pks-t: Initially I did, but I've removed them as soon as I noticed. So maybe it's not necessary anymore to disable those warnings, let me check again.

src/refdb_reftable.c
pks-t: > If there is a concurrent process, you'd normally have some sort of synchronization with that process, no?

Unfortunately, no. As explained below, libgit2 may be part of a long running application like e.g. an IDE where you don't know what other tools the user may be running. So we have to consistently refresh all files that could potentially change, like references or the config file.

deps/reftable/CMakeLists.txt
pks-t: I've checked again, it's "dump.c". Didn't look like an executable that's not part of the library, which is why I've compiled it in. Maybe it would be worthwhile to split out tests as well as standalone executables into their own subdirectory to allow just adding "*.c" to your project.

hanwen:
On Wed, Mar 25, 2020 at 3:09 PM Patrick Steinhardt <notifications@github.com>
wrote

>
>    -
>
>    At some places I feel like there's high potential for namespacing
>    issues. E.g. struct iterator is highly generic and may trivially clash
>    with any preexisting types. It would be nice to use namespace prefixes for
>    all exported symbols.
>
>
I fixed this.


>
>    -
>
>    -
>
>    I think it would help if the C files were layed out in a more
>    "standard" library way. Most importantly, separate out the "reftable.h"
>    header file from the rest into its own folder so that one can add it to the
>    include path without having to care about accidentally overriding "tree.h"
>    that's part of libgit2 itself.
>
>
I moved reftable.h in a different directory.

>
>    -
>
>    I didn't find any way to override memory allocation functions. I think
>    we've discussed that at the Contributor's summit already.
>
>
I added this. reftable_set_malloc().


-- 
Han-Wen Nienhuys - hanwenn@gmail.com - http://www.xs4all.nl/~hanwen

deps/reftable/system.h
```diff
@@ -0,0 +1,53 @@
...
+#include <inttypes.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/time.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <zlib.h>
```
pks-t: This currently doesn't compile on Windows due to <sys/time.h> not being found. I think it should work just fine if changing this to `<time.h>`. Instead, we could probably have the equivalent of `REFTABLE_IN_GITCORE`, but I think I'd rather prefer the header to be agnostic of libgit2.

pks-t:
Thanks a lot for addressing my concerns, @hanwen! I've updated the PR to include all of them.

deps/reftable/system.h
hanwen: On linux, this would yield: 

c/stack.c:226:12: error: implicit declaration of function 'gettimeofday' [-Werror=implicit-function-declaration]
  226 |  int err = gettimeofday(&deadline, NULL);
      |            ^~~~~~~~~~~~

pks-t:
On Fri, Mar 27, 2020 at 07:05:28AM -0700, Han-Wen Nienhuys wrote:
> @hanwen commented on this pull request.
> 
> On linux, this would yield: 
> 
> c/stack.c:226:12: error: implicit declaration of function 'gettimeofday' [-Werror=implicit-function-declaration]
>   226 |  int err = gettimeofday(&deadline, NULL);
>       |            ^~~~~~~~~~~~

Too bad. So the reftable would need to check for some Windows-specific
defines and adjust accordingly.

By the way, getting back to a different topic: `HEAD` and its various
other incarnations `*_HEAD`. Are those expected to be managed by
reftable, too? It would make sense for some of them, but on the other
hand we have some that are references, but have a special format. The
most common one that comes to my mind is `FETCH_HEAD`, but there are
others, too.

hanwen:
For the ref-like objects, this is what JGit does
 https://eclipse.googlesource.com/jgit/jgit/+/459fc28cf6a22992b7d168cfd52286cd76ca68bc/org.eclipse.jgit/src/org/eclipse/jgit/lib/RefDatabase.java#511

(https://eclipse.googlesource.com/jgit/jgit/+/459fc28cf6a22992b7d168cfd52286cd76ca68bc/org.eclipse.jgit/src/org/eclipse/jgit/internal/storage/file/RefDirectory.java#113)

In JGit, HEAD is expected to be managed by RefDatabase. Unfortunately, nobody has been able to articulate to me what the expectations/abstractions for this in git-core should do. Can you do transactional updates with HEAD and another branch together in libgit2?

pks-t:
On Sat, Mar 28, 2020 at 08:18:02AM -0700, Han-Wen Nienhuys wrote:
> In JGit, HEAD is expected to be managed by RefDatabase. Unfortunately,
> nobody has been able to articulate to me what the
> expectations/abstractions for this in git-core should do. Can you do
> transactional updates with HEAD and another branch together in
> libgit2?

Yes, this can be done via our transaction API. It does require the
reference backend to support locking though, which is not currently
possible with the reftable refdb.

I don't know if the issue is that reftables do not support locking or
whether our own assumptions were too centered on file refs and not
really on the transactional part. Anyway, this is nothing we can easily
change as it's part of our public interface.

hanwen:
Hey Patrick, 

can you check out 

https://godoc.org/github.com/google/reftable#Addition

and 

https://godoc.org/github.com/google/reftable#Stack.NewAddition

would that work as an API for you?

pks-t:
Just to make sure I understand it, this is how I'd imagine it to look like in C:

```C
/* Create a new transactional addition for the stack */
reftable_stack_new_addition(&addition, stack);
/* Write a set of ref/log records via the usual writer callback */
reftable_addition_add(addition, reftable_backend_write_records, &args1);
/* Writing multiple times is okay for a single addition */
reftable_addition_add(addition, reftable_baickend_write_records, &args2);
/* Now commit changes to persist it to disk and update the stack */
reftable_addition_commit(addition);
```

If I understand it correctly, then `Close()` will _not_ commit any data but just release the Addition, right? So it's kind of a rollback/abort function.

If my above assumptions are correct then the API should work for us, thanks a lot!

deps/reftable/include/reftable.h
```diff
@@ -0,0 +1,502 @@
...
+ ****************************************************************/
+
+/* a stack is a stack of reftables, which can be mutated by pushing a table to
+ * the top of the stack */
+struct reftable_stack;
+
+/* open a new reftable stack. The tables will be stored in 'dir', while the list
+   of tables is in 'list_file'. Typically, this should be .git/reftables and
+   .git/refs respectively.
+*/
```
pks-t: This comment is outdated, right? I think it's `.git/reftables` and `.git/reftables/tables.list` now. In which case it feels like the second parameter can be removed or at least made optional as the tables file is contained in `dir` anyway.

deps/reftable/include/reftable.h
```diff
@@ -0,0 +1,502 @@
...
+	/* boolean: do not generate a SHA1 => ref index. */
+	int skip_index_objects;
+
+	/* how often to write complete keys in each block. */
+	int restart_interval;
+
+	/* 4-byte identifier ("sha1", "s256") of the hash.
+	 * Defaults to SHA1 if unset
+	 */
+	uint32_t hash_id;
```
pks-t: I think this doesn't quite work as advertised. I've set up `reftable_write_options opts ={0}`, where I'd expect it to just default to "sha1". But `reftable_new_merged_table()` fails when comparing `r->hash_id != hash_id`.

pks-t: It would also be helpful to have some macros available so that one may set up the expected values without having to define them locally.

pks-t:
I've added a test vector and was able to get a first test to pass with the reftable backend, which reads a reference and then renames it. More tests and bug fixes will follow.

I've fixed the refdb detection to now consult "extensions.refStorage" as mandated by the documentation.

I've noticed that another requirement is to get repository format version 1 implemented. There's a temporary commit in here that just removes the error check.

deps/reftable/include/reftable.h
hanwen: I fixed this.

deps/reftable/include/reftable.h
hanwen: I can't repro your complaint, see

 https://github.com/google/reftable/commit/8a04e7f2831233bdedaa6a39111200ce4052eff8

can you provide a repro testcase?

hanwen:
https://github.com/google/reftable/commit/7aabfe0f39ce40f47a9a0f63e9c8b10f753e5afe for the transaction api in C.

hanwen:
BTW, - I would be grateful for any feedback about reftable on windows. I don't have windows, and I'd like to keep it that way.

pks-t:
Rebased on current master, which now includes support for repository format v1 and extensions.

> BTW, - I would be grateful for any feedback about reftable on windows. I don't have windows, and I'd like to keep it that way.

I may dig up my Windows VM as soon as I've got the backend working on my own machine. Until then there's currently only the CI that provides any kind of feedback on Windows.

deps/reftable/include/reftable.h
hanwen: ping?

pks-t: Sorry, forgot to check in for comments. The following fails if I do not set up `options.hash_id`:

```c
const char *dir, *table;
struct reftable_write_options options = {0};
struct reftable_ref_record ref = {0};
struct reftable_stack *stack;

dir = "/path/to/testrepo-reftable/.git/reftable";
table = "/path/to/testrepo-reftable/.git/reftable/tables.list";

/* This line is required, otherwise `reftable_new_stack` fails. */
options.hash_id = 0x73686131;

cl_must_pass(reftable_new_stack(&stack, dir, table, options));
cl_must_pass(reftable_stack_reload(stack));
cl_must_pass(reftable_stack_read_ref(stack, "refs/heads/master", &ref));
```

pks-t: Note that I'm currently not using an up to date version of the reftable library as I first want to try to get what I have completely functioning (which it mostly is already, but there's still some issues with writing for preexisting branches).

hanwen: which version are you using?

btw, if you are writing existing branches, make sure you sure you set an update_index that is beyond the update_index of the existing branch.

hanwen: thanks, fixed in https://github.com/google/reftable/commit/f781daf143482095be9bba632de61bc0281b8f36

hanwen:
@pks-t  - does libgit2 already have code to detect file / directory conflicts? Looks like I may need to write and/or refactor that to check transactions, and wondering if that makes sense to put into the reftable library.

pks-t:
On Mon, Apr 27, 2020 at 11:19:38AM -0700, Han-Wen Nienhuys wrote:
> @pks-t  - does libgit2 already have code to detect file / directory
> conflicts? Looks like I may need to write and/or refactor that to
> check transactions, and wondering if that makes sense to put into the
> reftable library.

No, not yet, but it was on my todo list. I wouldn't complain if it was
moved into reftable, though.

Didn't have much time lately due to the world being even crazier than it
used to be, but I hope to tackle the missing pieces soonish again.

pks-t:
Rebased on latest master, updated the reftable library to 66d38fe (C: move block_iter_next comment to header, 2020-04-28), fixed a set of bugs in the reflog implementation and improved error handling in most places. All in all the PR is close to being finished, but I've stumbled over one weird bug where compressing of the reftable stack seems to fail after writing symbolic references. No idea yet whether that's a fault of mine or in the reftable library, but I'll investigate.

hanwen:
ok.  I'm working on refname verification. I finished the Go code https://github.com/google/reftable/commit/2abf6f32a0f17e720d40e821ea224ac3e7618760 , and will try to get the C code done today.

hanwen:
refname validation will be done automatically if you upgrade to the latest version:

https://github.com/google/reftable/commit/c8f1f44b463ca1ddd96efebe1b009b00873bf628

pks-t:
Quick status update: I've updated the reftable code to the latest revision, made some more fixes and added more tests. By now, about 80% of our refs:: tests pass with the reftable code, the only thing missing right now is the reflog. As soon as that's implemented I'd consider this PR as ready to be reviewed.

pks-t:
@hanwen I've found one bug (or at least unexpected behaviour) with the reftable implementation: if I call `reftable_backend_write_records` with no reference records but `n` log records, where each of the log records has the same name, then it will return an error. I've traced this down to `writer_add_record`:

```C
static int writer_add_record(struct reftable_writer *w, struct record rec)
{
	int result = -1;
	struct slice key = { 0 };
	int err = 0;
	record_key(rec, &key);
	if (slice_compare(w->last_key, key) >= 0) {
		goto exit;
	}

...
```

So if the last-written entry has the exact same key as the entry we want to write an error is returned. The check looks quite explicit and thus looks like this is expected behaviour. But if a given application wants to write multiple log records at once for a given reference, then this check doesn't make a lot of sense to me.

hanwen:
> @hanwen I've found one bug (or at least unexpected behaviour) with the reftable implementation: if I call `reftable_backend_write_records` with no reference records but `n` log records, where each of the log records has the same name, then it will return an error. I've traced this down to `writer_add_record`:
> 
> ```c
> static int writer_add_record(struct reftable_writer *w, struct record rec)
> {
> 	int result = -1;
> 	struct slice key = { 0 };
> 	int err = 0;
> 	record_key(rec, &key);
> 	if (slice_compare(w->last_key, key) >= 0) {
> 		goto exit;
> 	}
> 
> ...
> ```
> 
> So if the last-written entry has the exact same key as the entry we want to write an error is returned. The check looks quite explicit and thus looks like this is expected behaviour. But if a given application wants to write multiple log records at once for a given reference, then this check doesn't make a lot of sense to me.

I think you're trying to write records at the same update_index. The update_index is the logical timestamp, and you have to specify them (decreasing order) to disambiguate reflog entries. For reflog entries, the update_index appended to the key.

pks-t:
> I think you're trying to write records at the same update_index. The update_index is the logical timestamp, and you have to specify them (decreasing order) to disambiguate reflog entries. For reflog entries, the update_index appended to the key.

@hanwen Huh, I thought I remembered you said it's okay to use the same update index for all entries written via `reftable_stack_add`. The callback I have for `reftable_stack_add` thus sets them all to the same:

```C
static int reftable_backend_writer(struct reftable_writer *wr, void *_arg)
{
	reftable_backend_writer_args *arg = (reftable_backend_writer_args *) _arg;
	int i;

	for (i = 0; i < arg->nrefs; i++)
		arg->refs[i].update_index = arg->index;
	for (i = 0; i < arg->nlogs; i++)
		arg->logs[i].update_index = arg->index;

	reftable_writer_set_limits(wr, arg->index, arg->index);
	if (reftable_writer_add_refs(wr, arg->refs, arg->nrefs) < 0 ||
	    reftable_writer_add_logs(wr, arg->logs, arg->nlogs) < 0)
		return -1;

	return 0;
}
```

Should I increase the index for each reference and log record?

hanwen:
ref records are keyed by ref_name. log records are keyed by (ref_name + update_index).

When you write a normal update of a ref, you should write the ref record and log record at the same update_index.

When you write a complete reflog (which happens if you are converting from a different storage format, for example), you have to write different reflog entries for the same ref at different update_indices.  See https://gerrit.googlesource.com/jgit/+/5072af65248ee76b76d91f3c367565c0a674de9d/org.eclipse.jgit/src/org/eclipse/jgit/internal/storage/file/FileReftableDatabase.java#537 for how JGit does this.

pks-t:
> ref records are keyed by ref_name. log records are keyed by (ref_name + update_index).

Aha, thanks a lot for clearing up this misconception of mine! I'll update my code as soon as I find the time to do it, then.

pks-t:
Update: Most of our reflog-related tests now pass, with the only exception being deletion of reflog entries. I think that's the last part required to move this out of RFC phase.

ethomson:
Hey @pks-t - I know that your efforts of reftable have landed in https://github.com/git/git itself - which is very exciting.

I rebased this onto our current main to get a sense for interoperability. I thought that it would be worth updating with the most recent reftable code. So some questions:

Is git itself the canonical upstream for reftable? I assume so.

But if so, would you be interested in patches to make reftable (in the git tree) not actually tied to git? It looks like there are some efforts to disentangle reftable from git, but there are also things like a dependency on `st_mult` which makes its extraction difficult for many reasons.

I wouldn't want to spend a lot of time trying to figure out how to get the newest reftable work if I'm looking in the wrong place and the git version is something adapted from an upstream that we could share in common.

hanwen:
> Is git itself the canonical upstream for reftable? I assume so.

yes. Patrick did a large number of fixes on top of my code.

> It looks like there are some efforts to disentangle reftable from git,

AFAIK it's rather the reverse: it started out as nominally separate from git, but then over time more git-core dependencies crept in.

ethomson:
Thanks @hanwen - makes sense. I knew that you had tried to separate things out a bit (thank you!); it looks like some of that has persisted but some new philosophies have snuck in as well.

pks-t:
> It looks like there are some efforts to disentangle reftable from git,

@ethomson I have in fact a patch series pending that disentangles the reftable library from the Git codebase again. I have plans to send this to the mailing list next week.

ethomson:
@pks-t acked - I will hold off on doing anything until you push forward the disentanglement PR and then we can figure out how to proceed here

pks-t:
Makes sense. Let's also connect next week on Git Merge and chat about
this.

leventov:
> @ethomson I have in fact a patch series pending that disentangles the reftable library from the Git codebase again. I have plans to send this to the mailing list next week.

@pks-t have you published those somewhere?

ethomson:
Yep, it's on the mailing list as well and being enqueued for release. Once it lands in git, we'll pick it up as well.

pks-t:
I've updated the pull request a bit:

- It's now based on the latest `main` branch.
- The bundled reftable library has been updated to the latest version of Git.
- The way we handle reflogs has been adapted to work better.
- Various other small fixes.

What is still missing:

- There still is a bug with reflogs somewhere, which causes 16 tests to fail.
- There is no conflict detection yet for D/F-style conflicts, which causes another three test failures.
- The integration bits in "reftable/system.c" need to be made actually portable.

pks-gitlab:
> There is no conflict detection yet for D/F-style conflicts, which causes another three test failures.

This has been addressed now, there is proper conflict detection in the reftable backend now. Still remaining is:

- "reftable/system.{c,h}" portability fixes.
- Reflogs are still somewhat broken.

Overall, 13 tests are currently failing, all due to reflogs.

pks-t:
In the current state all tests now pass on Linux. There's a couple patches I've sent upstream to fix issues with the reftable library in Git, but those are rather minor. Still needed is an adjustment of "deps/reftable/system.c" to work with non-Linux systems as well as CI changes to actually run tests with reftables enabled. I should be able to wrap that up next week, I guess.

pks-t:
Closing in favor of https://github.com/libgit2/libgit2/pull/7117, as I'm working on this with my GitLab account right now.

PR #5465, opened 2020-03-26
credentials: provide backcompat for opaque structs

### LINKED ISSUE #5415: 0.99.0: kind of ABI/API changes issue
Looks like ABI/API changes in 0.99.0 are quitre subsptantial and by this libgit2-glib can ot be build 0.99.0. SONAME of the library changed as well (which is good in this case).

However pkgconfig file still is libgit2.pc. I think that probably it would be good to change it to libgit2-1.pc to not confuse pkgconfig detection.

ethomson:
The credential structures are now opaque and defined in `sys/credential.h`.  However, we should continue to provide them for backward compatibility, unless `GIT_DEPRECATED_HARD` is set.

Fixes #5415

### DIFF
--- a/include/git2/deprecated.h
+++ b/include/git2/deprecated.h
@@ -41,6 +41,13 @@
  */
 #ifndef GIT_DEPRECATE_HARD
 
+/*
+ * The credential structures are now opaque by default, and their
+ * definition has moved into the `sys/credential.h` header; include
+ * them here for backward compatibility.
+ */
+#include "sys/credential.h"
+
 /**
  * @file git2/deprecated.h
  * @brief libgit2 deprecated functions and values


include/git2/deprecated.h
```diff
@@ -41,6 +41,13 @@
  */
 #ifndef GIT_DEPRECATE_HARD
 
+/*
+ * The credential structures are now opaque by default, and their
+ * definition has moved into the `sys/credential.h` header; include
+ * them here for backward compatibility.
+ */
+#include "sys/credential.h"
```
pks-t: I thought of adding these headers to "include/git2/credential.h", but this is definitely the better location to have it.

PR #5466, opened 2020-03-26
patch: correctly handle mode changes for renames

### LINKED ISSUE #5460: git_diff_from_buffer fails to parse a patch when it involves a rename and file mode change
### Reproduction steps

```diff
diff --git a/RENAMED.md b/README.md
old mode 100644
new mode 100755
similarity index 100%
rename from RENAMED.md
rename to README.md

```

using the above as input to `git_diff_from_buffer`, it fails with the message:

```
failed - invalid patch header at line 4
```

This seems to occur only when there is a rename and file mode change involved in the diff, since it operates fine if I do not have a rename involved.

---

I've attached a small reproduction of the behaviour. [reproduction.zip](https://github.com/libgit2/libgit2/files/4370818/reproduction.zip)

```bash
$ swift run
```

### Expected behaviour

diff is parsed without an error being omitted.

### Actual behaviour

```
failed - invalid patch header at line 4
```

0.28.4

macOS 10.15

pks-t:
When generating a patch for a renamed file whose mode bits have changed
in addition to the rename, then we currently fail to parse the generated
patch. Furthermore, when generating a diff we output mode bits after the
similarity metric, which is different to how upstream git handles it.

Fix both issues by adding another state transition that allows
similarity indices after mode changes and by printing mode changes
before the similarity index.

---

Fixes #5460

### DIFF
--- a/src/diff_print.c
+++ b/src/diff_print.c
@@ -359,9 +359,9 @@ int diff_delta_format_similarity_header(
 		abort();
 
 	if ((error = git_buf_puts(&old_path, delta->old_file.path)) < 0 ||
-		(error = git_buf_puts(&new_path, delta->new_file.path)) < 0 ||
-		(error = git_buf_quote(&old_path)) < 0 ||
-		(error = git_buf_quote(&new_path)) < 0)
+	    (error = git_buf_puts(&new_path, delta->new_file.path)) < 0 ||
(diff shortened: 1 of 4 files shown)


pks-t:
@ethomson: do we want to include this in v1.0?

PR #5467, opened 2020-03-26
Release v0.28.5

pks-t:
This is in preparation of release v0.28.5. It includes all PRs tagged with backport-v0.28.5, except for:

- Azure PRs. I'll pick them as required to get CI running.
- #5373. I'd have loved to pick #5373, but it's non-trivial to backport.
- the release script PR

Backporting: `$ ./script/backport.sh 4996 5140 5158 5159 5179 5191 5208 5209 5210 5214 5224 5232 5248 5257 5264 5273 5276 5285 5295 5298 5302 5303 5312 5314 5333 5337 5360 5413 5445 5456`. There are three merge conflicts, all three of them are trivial to resolve.


pks-t:
I've un-tagged all Azure pull requests in favor of adding one manual commit to upgrade our Windows machines to vs2017-win2016, as the old hosted pool is deprecated and is scheduled for removal on March 31st.

pks-t:
This took a lot longer than anticipated due to some CI changes I had to backport. Cherrypicking those commits wasn't viable as there were just too many changes, so I instead crafted minimal commits that get our CI working until after March 31st, where current macOS and Windows machines will get deprecated.

I've also cherrypicked the release scripts and adjusted its soname format to reflect what we had in the v0.28 release branch and added a changelog.

All in all, this should be ready now for a release together with v1.0.

ethomson:
Agree that #5373 would be helpful.  Since the `gitno_connection_data` -> `git_net_url` change was my mess, I took a look at cherry-picking it.  Seemed not too bad.  See c643e70a74b33bf94630a3ea69325ea91968285a if you're interested.

pks-t:
On Fri, Mar 27, 2020 at 02:53:07AM -0700, Edward Thomson wrote:
> Agree that #5373 would be helpful.  Since the `gitno_connection_data`
> -> `git_net_url` change was my mess, I took a look at cherry-picking
> it.  Seemed not too bad.  See c643e70a74b33bf94630a3ea69325ea91968285a
> if you're interested.

No mess at all, it's been a great refactoring to streamline it with our
codebase and heavily needed, if you ask me.

Anyway, thanks for doing the cherry-pick for me. I had the exact same
code, except that I somehow wasn't able to find
`gitno_connection_data_fmt`. No idea how I missed it.

Do you have any other PRs you want to propose for this bugfix release?

pks-t:
On Fri, Mar 27, 2020 at 05:24:19AM -0700, Patrick Steinhardt wrote:
> Anyway, thanks for doing the cherry-pick for me. I had the exact same
> code, except that I somehow wasn't able to find
> `gitno_connection_data_fmt`. No idea how I missed it.

Which is because it didn't exist yet. Unfortunately it segfaults on my
system in checkout::index::honor_coresymlinks_default_true when passing
connection data to `gitno_connection_data_fmt` that has no host.

ethomson:
Thanks - it looks like in that case, we should just just return the path as-is.  Oy, lots of differences between v0.28 and v0.99.  Does 7ac289b95c1580c24fda5c2a386cf2a05be96e5c seem more reasonable?

pks-t:
On Fri, Mar 27, 2020 at 05:37:39AM -0700, Edward Thomson wrote:
> Thanks - it looks like in that case, we should just just return the path as-is.

I'd think so, too.

> Oy, lots of differences between v0.28 and v0.99.

I know, right? v0.99 has been taking a lot longer than I initially
realized and has a huge swath of changes in general.

> Does 7ac289b95c1580c24fda5c2a386cf2a05be96e5c seem more reasonable?

It does, thanks a lot! It does make me worry a bit that we backport such
a patch with quite some changes, especially given there was a bug
in the backport. But I guess I'd still like to take it in, given that we
really don't want to be leaking credentials.

Issue #4960, opened 2019-01-25
1.0 planning / next few releases


ethomson:
Hey, it's just shy of _five years ago_ when we did our [first planning for 1.0](https://github.com/libgit2/libgit2/issues/2136).  😆😢😭

But seriously, here's my thinking for a 1.0 schedule:  let's be quality-focused over the next month or so.  I'm not _opposed_ to merging feature work, but I want to prioritize bug fixing and API stability.

* **0.28**:  January 2019  
Our next release.  We've done some (responsible!) refactoring of some public API for consistency.  But we're not quite done yet, and it's been _forever_ since we've had a release.  So I want to push out any further refactorings into the (yes) next release.  I want to button up the last few changes that need to land in _this_ release and get 0.28 out the door.  Many consumers are asking for a proper stable release and we're already behind.

* **0.99**:  Late February/Early March 2019
I'd like 0.99 to be a real prerelease of 1.0.  I'd like to see us finalize any (still responsible) refactoring of the public API.  I want to see us warning free on _all platforms_ not just the Unix platforms.  Fundamentally, I want this to really be a staging of the actual release, and we really expect this to be the *last* changes before 1.0.  We should expect the public APIs to be frozen, and only take minor bugfixes for 1.0.  I would expect that we'd branch off master for a freeze so that we can pick up post-1.0 development.  🤞 

* **1.0**: Late March 2019
Ideally contains no changes from 0.99 except to bump the version number and SONAME.  🎉

(I'm not strictly opposed to sneaking an 0.29 in here if we feel we need it but I want to focus on stablility, and get us to 1.0 reasonably quickly but responsibly.  I would rather iterate on an 0.99 and release an 0.99.1, 0.99.2 if that's actually necessary.  This is a subtle difference, I concede, but I think there's a psychological difference.)

Here's some things that I would like to see resolved and shipped in 0.99:

- [x] `git_off_t` is...  surprising.  Why is it signed?  Why is it named after an "offset" when it's a size?  Maybe these are fine, but they're not obvious, so let's reconsider it and fix it or document why it is the way it is.
- [x] What is going on with `include/git2/inttypes.h`?  We're publishing an inline `imaxdiv` function.  😢 
[...]

PR #5471, opened 2020-03-28
Release v1.0

pks-t:
This is in preparation for v1.0, including the update to our changelog and the version bump. I don't think we settled for any name yet, so I just chose a temporary one ("Aprilscherz" -> April Fools' joke). @carlosmn, you always manage to come up with weird German words, so if you want to give this a go I'd be happy to change it ;)

Right now, it looks like we should be able to release both v1.0 and v0.28.5 right on schedule on April 1st.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,3 +1,54 @@
+v1.0
+----
+
+This is release v1.0 "Luftschloss", which is the first stabe release of
+libgit2. The API will stay compatible across all releases of the same major
+version. This release includes bugfixes only and supersedes v0.99, which will
+stop being maintained. Both v0.27 and v0.28 stay supported in accordance with
+our release policy.
+
+### Changes or improvements
+
+- CMake was converted to make use of the GNUInstallDirs module for both our
+  pkgconfig and install targets in favor of our custom build options
+  `BIN_INSTALL_DIR`, `LIB_INSTALL_DIR` and `INCLUDE_INSTALL_DIR`. Instead, you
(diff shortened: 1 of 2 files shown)


ethomson:
In https://github.com/libgit2/libgit2/issues/4960, I suggested:

> By the way, I have a name in mind for our 1.0 release. "Luftschloss (air castle)". The fluent German speakers can correct me, but my understanding is:
> 
> > This refers to an impossible or unrealistic dream -- like building a castle in the air -- and is used to describe that which, sadly, just cannot be achieved.
>
> This seems appropriate for something that we sat down in a meeting room to plan ... 5 years ago?

pks-t:
On Sat, Mar 28, 2020 at 06:15:26AM -0700, Edward Thomson wrote:
> In https://github.com/libgit2/libgit2/issues/4960, I suggested:

Oh, sorry, I forgot about that one! Let me amend it.

ethomson:
Let's make sure that we release v0.28.5 _before_ v1.0, since the ordering on the releases page is by date (ordered newest first).  We want to make sure that v1.0 shows up on top.  😀

pks-t:
On Mon, Mar 30, 2020 at 03:17:25AM -0700, Edward Thomson wrote:
> Let's make sure that we release v0.28.5 _before_ v1.0, since the
> ordering on the releases page is by date (ordered newest first).  We
> want to make sure that v1.0 shows up on top.  😀 

Exactly, thanks for the reminder and for your review!

PR #5473, opened 2020-03-30
fetchhead: strip credentials from remote URL

ethomson:
If fetching from an anonymous remote via its URL, then the URL gets
written into the FETCH_HEAD reference. This is mainly done to give
valuable context to some commands, like for example git-merge(1), which
will put the URL into the generated MERGE_MSG. As a result, what gets
written into FETCH_HEAD may become public in some cases. This is
especially important considering that URLs may contain credentials, e.g.
when cloning 'https://foo:bar@example.com/repo' we persist the complete
URL into FETCH_HEAD and put it without any kind of sanitization into the
MERGE_MSG. This is obviously bad, as your login data has now just leaked
as soon as you do git-push(1).

When writing the URL into FETCH_HEAD, upstream git does strip
credentials first. Let's do the same by trying to parse the remote URL
as a "real" URL, removing any credentials and then re-formatting the
URL. In case this fails, e.g. when it's a file path or not a valid URL,
we just fall back to using the URL as-is without any sanitization. Add
tests to verify our behaviour.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -43,6 +43,11 @@ This is a bugfix release with the following changes:
 
 - Fixes for several memory leaks.
 
+- When fetching from an anonymous remote using a URL with authentication
+  information provided in the URL (eg `https://foo:bar@example.com/repo`),
+  we would erroneously include the literal URL in the FETCH_HEAD file.
+  We now remove that to match git's behavior.
+
 v0.28.4
 --------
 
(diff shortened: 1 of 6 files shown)


ethomson:
Let's sneak this in before v0.28.5.  😀

pks-t:
Oops, did I somehow drop your backport again? I thought I actually cherry-picked it.

pks-t:
@ethomson: looks good to me, thanks for the backport! We should add one more entry to the changelog, though.

ethomson:
Added a changelog entry.

pks-t:
Thanks, @ethomson!

PR #5474, opened 2020-03-30
gitignore: clean up patterns from old times

pks-t:
The gitignore file currently has a lot of patterns for files that we
shouldn't write anymore since we have migrated to CMake, as everybody is
expected to do out-of-source builds anyway. Let's remove them.


.gitignore
```diff
@@ -1,37 +1,8 @@
...
-*.o
-*.a
-*.exe
-*.gcda
-*.gcno
-*.gcov
-.lock-wafbuild
-.waf*
 build/
 build-amiga/
```
ethomson: We should be able to get rid of `build-amiga`, too, I would think.

pks-t:
On Mon, Mar 30, 2020 at 03:45:40AM -0700, Edward Thomson wrote:
> 
> 
> 
> > @@ -1,37 +1,8 @@
> -/tests/clar.suite
> -/tests/clar.suite.rule
> -/tests/.clarcache
> -/apidocs
> -/trash-*.exe
> -/libgit2.pc
> -/config.mak
> -*.o
> -*.a
> -*.exe
> -*.gcda
> -*.gcno
> -*.gcov
> -.lock-wafbuild
> -.waf*
>  build/
>  build-amiga/
> 
> We should be able to get rid of `build-amiga`, too, I would think.

I thought Amiga was vital to our project :( I've amended this PR ;)

ethomson:
> I thought Amiga was vital to our project :( I've amended this PR ;)

Obviously it is, but they should be using the `build` directory like everybody else.  😝

Issue #3751, opened 2016-04-25
Function to parse author and committer information from environment


joshtriplett:
`git_signature_default` parses the `user.name` and `user.email` values from `git_config`, but does not allow for other sources of that information.  Per `man git-commit-tree`:

```
       author and
       committer information is taken from the following environment
       variables, if set:

           GIT_AUTHOR_NAME
           GIT_AUTHOR_EMAIL
           GIT_AUTHOR_DATE
           GIT_COMMITTER_NAME
           GIT_COMMITTER_EMAIL
           GIT_COMMITTER_DATE

       (nb "<", ">" and "\n"s are stripped)

       In case (some of) these environment variables are not set, the
       information is taken from the configuration items user.name and
       user.email, or, if not present, the environment variable EMAIL
```

While the existing behavior of `git_signature_default` should likely not change to take environment variables into account, I'd suggest adding a `git_signature_author_env` and `git_signature_committer_env` that do.  Those would be the appropriate defaults to pass when creating a commit, for any program mimicking the git command-line tools.

venkatesh-coder:
Is this issue still not resolved, I'd like to work on this issue seems achievable to me. Let me know.
Thanks for creating a great library for us. Thank you.

venkatesh-coder:
Thanks @ethomson, I will try my best. I will ask questions if I need more clarification.

venkatesh-coder:
as @joshtriplett mentioned, I created functions `git_signature_author_env` and `git_signature_committer_env` that gets environment variables mentioned in man page for *git-commit* and stores them as `git_signature` struct. I just want to clarify some things
   * there are functions like `git_commit_create`, `git_commit__create_internal` and `git_commit__create_buffer_internal` (first two functions are wrappers for `git_commit__create_buffer_internal`) that sets author and committer details. which function best fits to use `git_signature_author_env` and `git_signatue_committer_env` to change `author` and `committer`.
   * I think `git_commit_create` is right function because others part of the code might use `git_commit__create_buffer_internal` or `git_commit__create_internal` for other purposes the `git_commit__create_buffer_internal` function being the lowest level function which writes to commit `buf`.
   * can I use the `git_signature_author_env` and `git_signature_committer_env` to change the parameters `author` and `committer` that being passed to `git_commit__create_internal` in `git_commit_create`
   My assumptions might be wrong please help me.
Thank you.

PR #5479, opened 2020-04-01
Add create-signature-from-env functions, per #3751

paulroub:
Use GIT_AUTHOR_NAME and GIT_AUTHOR_EMAIL if available (or
GIT_COMMITTER_NAME and GIT_COMMITTER_EMAIL)

Fall back to EMAIL if GIT_AUTHOR_EMAIL/GIT_COMMITTER_EMAIL is not available

Use the default user info (user.name, user.email) as last resorts.

Return GIT_ENOTFOUND if we don't have enough info for a signature

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,3 +1,13 @@
+v1.?
+----
+
+### Changes or improvements
+
+- Add two new functions, `git_signature_committer_env()` and
+  `git_signature_author_env()`, which create action signatures based on the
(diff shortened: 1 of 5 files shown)


ethomson:
Hey @paulroub - regrets, I think that you opened this during a time where I was busy with other work. I sort of lost sight of this, and I think that this has been resolved with #6706. So I'm closing this.

Hope you're well, and apologies for ghosting you. 🙏

PR #5481, opened 2020-04-03
CMake cleanups

pks-t:
While I've outlined my greater vision of where our CMake setup should move in #5284, I've realized it's much too big to review it properly and thus decided to take a more iterative approach to modernizing our CMake build instructions. This is the first set of commits towards that goal and an initial spring cleaning. All changes should probably be uncontroversional and make sense on their own.


src/CMakeLists.txt
```diff
@@ -317,7 +318,9 @@ ELSE()
 	MESSAGE(FATAL_ERROR "Unsupported architecture (CMAKE_SIZEOF_VOID_P is unset)")
 ENDIF()
 
-CONFIGURE_FILE(features.h.in git2/sys/features.h)
+# Generate feature headers
+CONFIGURE_FILE(features.h.in ${libgit2_BINARY_DIR}/include/git2/sys/features.h)
```
ethomson, line 322: `features.h` is actually not installed at present, this is something that's only built and used locally.  Let's keep it private API without a compelling reason to promote it.

src/CMakeLists.txt
```diff
@@ -317,7 +318,9 @@ ELSE()
...
+CONFIGURE_FILE(version.h.in ${libgit2_BINARY_DIR}/include/git2/version.h)
```
ethomson, line 323: Generating this file will mean that people _must_ use CMake to build us.  At the moment, at least TortoiseGit uses a build step for us that doesn't involve CMake.

I think that they include us as a submodule.  Then they build us by defining the necessary things (eg `-DGIT_TRACE=1`) directly.  This, and the `git2.rc` changes will definitely break their ability to do this.

What's the motivation for defining the version in cmake instead of `version.h`?

jacquesg: The perl bindings also don't use CMake to build the library.

src/CMakeLists.txt
pks-t: Fair enough, makes sense :+1:

src/CMakeLists.txt
pks-t: @ethomson The motivation is so that we can use it for the `PROJECT()` setup function. I'm not sure whether you're allowed to do things like extracting the version previous to doing the `PROJECT()` call, but we should definitely put the version in there. I'll see whether it's possible to do so with our current `version.h` regex.

pks-t:
I've removed the version thing for now and will move it into a separate PR so that we should be left with the non-controversial bits, only.

src/CMakeLists.txt
YueLinHo: /cc @csware

csware: Forcing users to use CMake isn't a good idea.

pks-t:
Rebased to fix conflicts.

PR #5482, opened 2020-04-04
docs: add documentation for our coding style

pks-t:
For years, we've repeatedly had confusion about what our actual coding
style is not only for newcomers, but also across the core contributors.
This can mostly be attributed to the fact that we do not have any coding
conventions written down. This is now a thing of the past with the
introduction of a new document that gives an initial overview of our
style and most important best practices for both our C codebase as well
as for CMake.

While the proposed coding style for our C codebase should be rather
uncontroversial, the coding style for CMake might be. This can be
attributed to multiple facts. First, the CMake code base doesn't really
have any uniform coding style and is quite outdated in a lot of places.
Second, the proposed coding style actually breaks with our existing one:
we currently use all-uppercase function names and variables, but the
documented coding style says we use all-lowercase function names but
all-uppercase variables.

It's common practice in CMake to write variables in all upper-case, and
in fact all variables made available by CMake are exactly that. As
variables are case-sensitive in CMake, we cannot and shouldn't break
with this. In contrast, function calls are case insensitive, and modern
CMake always uses all-lowercase ones. I argue we should do the same to
get in line with other codebases and to reduce the likelihood of
repetitive strain injuries.

So especially for CMake, the proposed coding style says something we
don't have yet. I'm fine with that, as the document explicitly says that
it's what we want to have and not what we have right now.

---

This is long overdue, so I finally sat down today and just did it. The document should serve as a base we can point to, but doesn't aim to be complete right now. If I've missed anything important then I'm happy to amend it right now. For everything else I want to defer it to a later point in time and treat it as a living document that we amend every time we notice code that's conflicting with our perceived coding style, but where the actual style isn't documented.

### DIFF
--- a/docs/coding-style.md
+++ b/docs/coding-style.md
@@ -1,0 +1,364 @@
+# libgit2 Coding Style
+
+This documentation describes the preferred coding style for the libgit2 project.
+While not all parts of our code base conform to this coding style, the outlined
+rules are what we aim for.
+
+Note that in no case do we accept changes that convert huge parts of the code
+base to use our coding style. Instead, it is encouraged to modernize small parts
+of code you're going to modify anyway for a given change you want to introduce.
+A good rule to follow is the Boy Scout Rule: "Leave the campground cleaner than
+you found it."
+
+## C Coding Style
+
+The following sections define the coding style for all code files and headers.
+
+### Indentation and Alignment
+
+Code is indented by tabs, where a tab is 8 spaces. Each opening scope increases
+the indentation level.
+
+```c
+int foobar(int void)
+{
+	if (condition)
+		doit();
+	/* Body */
+}
+```
+
+Switch statements have their `case`s aligned with the `switch` keyword. Case
+bodies are indented by an additional level. Case bodies should not open their
+own scope to declare variables.
+
+```c
+switch (c) {
+case 'a':
+case 'b':
+	return 0;
+default:
+	return -1;
+}
+```
+
+Multi-line conditions should be aligned with the opening brace of the current
+statement:
+
+```c
+if (one_very_long_condition(c) &&
+    another_very_long_condition(c))
+	doit();
+```
+
+### Spaces
+
+There must be no space between the function and its arguments, arguments must be
+separated by a space:
+
+```c
+int doit(int first_arg, int second_arg);
+doit(1, 2);
+```
+
+For any binary or ternary operators, the arguments and separator must be
+separated by a space:
+
+```c
+1 + 2;
+x ? x : NULL;
+```
+
+Unary operators do not have a space between them and the argument they refer to:
+
+```c
+*c
+&c
+```
+
+The `sizeof` operator always must not have a space and must use braces around
+the type:
+
+```
+sizeof(int)
+```
+
+There must be a space after the keywords `if`, `switch`, `case`, `do` and
+`while`.
+
+### Braces
+
+Functions must have their opening brace on the following line:
+
+```c
+void foobar(void)
+{
+	doit();
+}
+```
+
+For conditions, braces should be placed on the same line as the condition:
+
+```c
+if (condition(c)) {
+	doit();
+	dothat();
+}
+
+while (true) {
+	doit();
+}
+```
+
+In case a condition's body has a single line, only, it's allowed to omit braces,
+except if any of its `else if` or `else` branches has more than one line:
+
+```c
+if (condition(c))
+	doit();
+
+if (condition(c))
+	doit();
+else if (other_condition(c))
+	doit();
+
+/* This example must use braces as the `else if` requires them. */
+if (condition(c)) {
+	doit();
+} else if (other_condition(c)) {
+	doit();
+	dothat();
+} else {
+	abort();
+}
+```
+
+### Comments
+
+Comments must use C-style `/* */` comments. C++-style `// `comments are not
+allowed in our codebase. This is a strict requirement as libgit2 tries to be
+compliant with the ISO C90 standard, which only allows C-style comments.
+
+Single-line comments may have their opening and closing tag on the same line:
+
+```c
+/* This is a short comment. */
+```
+
+For multi-line comments, the opening and closing tag should be empty:
+
+```c
+/*
+ * This is a rather long and potentially really unwiedly but informative
+ * multiline comment that helps quite a lot.
+ */
+```
+
+Public functions must have documentation that explain their usage, internal
+functions should have a comment. We use Docurium to generate documentation
+derived from these comments, which uses syntax similar to Doxygen. The first
+line should be a short summary of what the function does. More in-depth
+explanation should be separated from that first line by an empty line.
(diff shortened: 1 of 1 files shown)


docs/coding-style.md
```diff
@@ -0,0 +1,325 @@
...
+
+There must be no space between the function and its arguments, arguments must be
+separated by a space:
+
+```c
+int doit(int first_arg, int second_arg);
+doit(1, 2);
+```
+
+For any operators except for `*` and `&`, the arguments and separator must be
```
ethomson, line 64: We should probably indicate the _unary_ `*` and `&`.  It took me a moment to unpack what this was saying.

But upon reflection, I think that we never use spaces with unary operators and always use spaces with binary.

docs/coding-style.md
```diff
@@ -0,0 +1,325 @@
...
+	doit();
+}
+```
+
+In case a condition's body has a single line, only, it's allowed to omit braces:
+
+```c
+if (condition(c))
+	doit();
+```
```
ethomson, line 106: We should indicate how `else` and `else if` fit in.  We should also indicate that if any of those have a multi-line body then that promotes all of them to have braces.

docs/coding-style.md
```diff
@@ -0,0 +1,325 @@
...
+All structures declared in the libgit2 project must have a `typedef`, we do not
+use `struct type` variables. Type names follow the same schema as functions.
+
+### Error Handling
+
+The libgit2 project mostly uses error codes to indicate errors. Error codes are
+always of type `int`, where `0` indicates success and a negative error code
+indicates an error case. In some cases, positive error codes may be used to
+indicate special cases. Returned values that are not an error code should be
+returned via an out value:
```
ethomson, line 193: We should indicate that out params come first.

docs/coding-style.md
```diff
@@ -0,0 +1,325 @@
...
+To avoid repetitive and fragile error handling in case a function has resources
+that need to be free'd, we use `goto out`s:
+
+```c
+int doit(char **out, int arg)
+{
+	int error = 0;
+	char *c;
+
+	c = malloc(strlen("Got an argument") + 1);
```
ethomson, line 214: I wonder if we should have an assorted tips section.  Seeing this made me remember that C++ programmers often like to cast malloc and we might want to add a reminder that we don't do that.

docs/coding-style.md
```diff
@@ -0,0 +1,325 @@
...
+```
+
+### Naming
+
+While CMake is completely case-insensitive when it comes to function calls, we
+want to agree on a common coding style for this. To reduce the danger of
+repetitive strain injuries, all function calls should be lower-case (NB: this is
+not currently the case yet, but introduced as a new coding style by this
+document). Similarly, declarations of functions should be lower-case, too.
+
```
ethomson, line 304: woof, I don't know how I feel about this.  I don't _love_ that our CMake config is mostly ALL UPPERCASE, it does feel a little more "right" to me to have it that way instead of a mix of ALL UPPER for variables and all_lower for functions.

ethomson:
Thanks for putting this together!  I had a few thoughts while reading the C bits that we might want to expand upon.  I also want to argue about the CMake decision.  😀

docs/coding-style.md
pks-t: Yup, agreed.

docs/coding-style.md
pks-t: Fair, will add it in

docs/coding-style.md
pks-t: Indeed!

docs/coding-style.md
pks-t: It might certainly make sense. Are you fine with leaving this out for this PR and add it in at a later point?

docs/coding-style.md
pks-t: Haha, I knew this is going to be a tough sell ;) Thing is, CMake _always_ uses UPPERVARS and lowerfuncs. Take a look at their documentation, their exposed variables, "An Introduction to Modern CMake" (https://cliutils.gitlab.io/modern-cmake/). I feel like we shouldn't try to be special here and just do what's common practice.

pks-t: And by the way, I only mean this repetitive strain remark half-jokingly: I always got a quite noticeable pain in my little fingers after typing CMake code for more than half an hour :(

pks-t:
Updated to address your comments, thanks @ethomson!

docs/coding-style.md
ethomson: Works for me.

docs/coding-style.md
ethomson: From their documentation, it looks like they also lower-case things like `if`, `else`...  should we adopt that, too?

pks-t:
On Sun, Apr 05, 2020 at 09:44:03AM -0700, Edward Thomson wrote:
> 
> 
> 
> > +```cmake
> +IF(CONDITION)
> +	DOIT()
> +ELSE()
> +	DOTHAT()
> +ENDIF()
> +```
> +
> +### Naming
> +
> +While CMake is completely case-insensitive when it comes to function calls, we
> +want to agree on a common coding style for this. To reduce the danger of
> +repetitive strain injuries, all function calls should be lower-case (NB: this is
> +not currently the case yet, but introduced as a new coding style by this
> +document). Similarly, declarations of functions should be lower-case, too.
> +
> 
> From their documentation, it looks like they also lower-case things like `if`, `else`...  should we adopt that, too?

Yes, that was my intention here. The only thing we should continue to
uppercase in my opinion is variables, as they're case-sensitive and
CMake uses uppercase variables themselves. As you're asking, I should
probably make that explicit.

pks-t:
On Sun, Apr 05, 2020 at 10:34:56AM -0700, Patrick Steinhardt wrote:
> On Sun, Apr 05, 2020 at 09:44:03AM -0700, Edward Thomson wrote:
> > @ethomson commented on this pull request.
> > 
> > 
> > 
> > > +```cmake
> > +IF(CONDITION)
> > +	DOIT()
> > +ELSE()
> > +	DOTHAT()
> > +ENDIF()
> > +```
> > +
> > +### Naming
> > +
> > +While CMake is completely case-insensitive when it comes to function calls, we
> > +want to agree on a common coding style for this. To reduce the danger of
> > +repetitive strain injuries, all function calls should be lower-case (NB: this is
> > +not currently the case yet, but introduced as a new coding style by this
> > +document). Similarly, declarations of functions should be lower-case, too.
> > +
> > 
> > From their documentation, it looks like they also lower-case things like `if`, `else`...  should we adopt that, too?
> 
> Yes, that was my intention here. The only thing we should continue to
> uppercase in my opinion is variables, as they're case-sensitive and
> CMake uses uppercase variables themselves. As you're asking, I should
> probably make that explicit.

I should, especially as I'm doing it wrong in the above example already.
Habits.

ethomson:
LGTM, thanks @pks-t

pks-t:
Thanks a lot for your review, @ethomson!

PR #5483, opened 2020-04-04
Fix typo causing removal of symbol 'git_worktree_prune_init_options'

xSetech:
Commit 0b5ba0d replaced this function with an "option_init"
equivallent, but misspelled the replacement function. As a result, this
symbol has been missing from libgit2.so ever since.

### DIFF
--- a/src/worktree.c
+++ b/src/worktree.c
@@ -506,7 +506,7 @@ int git_worktree_prune_options_init(
 	return 0;
 }
 
-int git_worktree_pruneinit_options(git_worktree_prune_options *opts,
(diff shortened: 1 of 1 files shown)


xSetech:
Here's the referenced commit:
https://github.com/libgit2/libgit2/commit/0b5ba0d744e69da5dc8c08d167c83dd87ed83af2

xSetech:
I found this using software I think was built using the epel/RHEL/CentOS 7 version of libgit2. I had a newer copy in my LD_LIBRARY_PATH and the software failed with an error about missing this symbol.

PR #5487, opened 2020-04-05
deps: ntlmclient: use htobe64 on NetBSD too

niacat:
unbreaks the build. should probably use this by default with the operating systems that don't have it being the exception.


ethomson:
LGTM.  Thanks @niacat

PR #5491, opened 2020-04-17
odb_backend: Implement support for the  manual reordering of odb backends

tonydelanuez:
libgit2 has historically assumed that most objects for long-running
operations will be packed, therefore GIT_LOOSE_PRIORITY is set to
1 by default, and GIT_PACKED_PRIORITY is set at 2. 

When a client allows libgit2 to set the default backends, it would be helpful to
be able to rearrange the default backends without having to completely
set up a new odb from scratch. This commit adds a method to set
the priority for a backend in an odb and relies on the sorting in
git_odb_add_backend to rearrange the backends.

A use case for this would be to: create a new odb for a repository, let libgit2 set the default backends, then manually set the packed_backend priority to 1, loose_backend to 2, and add a custom backend with priority 3 to trigger the sort and have the custom backend be the first backend accessed.

### DIFF
--- a/include/git2/common.h
+++ b/include/git2/common.h
@@ -205,7 +205,9 @@ typedef enum {
 	GIT_OPT_GET_PACK_MAX_OBJECTS,
 	GIT_OPT_SET_PACK_MAX_OBJECTS,
 	GIT_OPT_DISABLE_PACK_KEEP_FILE_CHECKS,
-	GIT_OPT_ENABLE_HTTP_EXPECT_CONTINUE
+	GIT_OPT_ENABLE_HTTP_EXPECT_CONTINUE,
+	GIT_OPT_SET_ODB_PACKED_PRIORITY,
+	GIT_OPT_SET_ODB_LOOSE_PRIORITY
 } git_libgit2_opt_t;
 
 /**
@@ -403,6 +405,14 @@ typedef enum {
  *		> When connecting to a server using NTLM or Negotiate
  *		> authentication, use expect/continue when POSTing data.
  *		> This option is not available on Windows.
+ *
+ *   opts(GIT_OPT_SET_ODB_PACKED_PRIORITY, int priority)
+ *      > Override the default priority of the packed ODB backend which
+ *      > is added when default backends are assigned to a repository
+ *
+ *   opts(GIT_OPT_SET_ODB_LOOSE_PRIORITY, int priority)
+ *      > Override the default priority of the loose ODB backend which
+ *      > is added when default backends are assigned to a repository
  *
  * @param option Option key
  * @param ... value to set the option
--- a/src/odb.c
+++ b/src/odb.c
@@ -23,14 +23,14 @@
 
 #define GIT_ALTERNATES_FILE "info/alternates"
 
+#define GIT_ALTERNATES_MAX_DEPTH 5
+
 /*
  * We work under the assumption that most objects for long-running
  * operations will be packed
  */
-#define GIT_LOOSE_PRIORITY 1
-#define GIT_PACKED_PRIORITY 2
-
-#define GIT_ALTERNATES_MAX_DEPTH 5
+int git_odb_loose_priority = 1;
+int git_odb_packed_priority = 2;
 
 bool git_odb__strict_hash_verification = true;
 
@@ -572,12 +572,12 @@ int git_odb__add_default_backends(
 
 	/* add the loose object backend */
 	if (git_odb_backend_loose(&loose, objects_dir, -1, db->do_fsync, 0, 0) < 0 ||
-		add_backend_internal(db, loose, GIT_LOOSE_PRIORITY, as_alternates, inode) < 0)
+		add_backend_internal(db, loose, git_odb_loose_priority, as_alternates, inode) < 0)
 		return -1;
 
 	/* add the packed file backend */
 	if (git_odb_backend_pack(&packed, objects_dir) < 0 ||
-		add_backend_internal(db, packed, GIT_PACKED_PRIORITY, as_alternates, inode) < 0)
+		add_backend_internal(db, packed, git_odb_packed_priority, as_alternates, inode) < 0)
 		return -1;
 
 	return load_alternates(db, objects_dir, alternate_depth);
--- a/src/settings.c
+++ b/src/settings.c
@@ -61,6 +61,8 @@ extern size_t git_mwindow__window_size;
 extern size_t git_mwindow__mapped_limit;
 extern size_t git_indexer__max_objects;
 extern bool git_disable_pack_keep_file_checks;
+extern int git_odb_packed_priority;
+extern int git_odb_loose_priority;
 
 static int config_level_to_sysdir(int config_level)
 {
@@ -289,6 +291,14 @@ int git_libgit2_opts(int key, ...)
 
 	case GIT_OPT_ENABLE_HTTP_EXPECT_CONTINUE:
 		git_http__expect_continue = (va_arg(ap, int) != 0);
+		break;
+
+	case GIT_OPT_SET_ODB_PACKED_PRIORITY:
+		git_odb_packed_priority = va_arg(ap, int);
+		break;
+
+	case GIT_OPT_SET_ODB_LOOSE_PRIORITY:
+		git_odb_loose_priority = va_arg(ap, int);
 		break;
 
 	default:
--- a/tests/odb/sorting.c
+++ b/tests/odb/sorting.c
@@ -68,3 +68,26 @@ void test_odb_sorting__alternate_backends_sorting(void)
 
 	check_backend_sorting(_odb);
 }
+
+void test_odb_sorting__override_default_backend_priority(void)
+{
+	git_odb *new_odb;
+	git_odb_backend *loose, *packed, *backend;
+	cl_git_pass(git_libgit2_opts(GIT_OPT_SET_ODB_LOOSE_PRIORITY, 5));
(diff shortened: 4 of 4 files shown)


src/odb.c
```diff
@@ -1521,6 +1521,25 @@ void git_odb_backend_data_free(git_odb_backend *backend, void *data)
...
+		return -1;
+	}
+
+	for (i = 0; i < db->backends.length; ++i){
+		if (i == index) {
+			backend_internal *internal = git_vector_get(&db->backends, i);
+			internal->priority = priority;
+		}
+	}
+	return 0;
```
hackworks, line 1540: Sorting the vector after modifying priority will make this function independent. In the unit test, it is getting re-sorted since you are adding a new backend after invoking this function.

src/odb.c
```diff
@@ -1521,6 +1521,25 @@ void git_odb_backend_data_free(git_odb_backend *backend, void *data)
...
+{
+	size_t i;
+	assert(db);
+
+	if (index > db->backends.length - 1) {
+		git_error_set(GIT_ERROR_INVALID, "invalid odb backend index");
+		return -1;
+	}
+
+	for (i = 0; i < db->backends.length; ++i){
```
ethomson, line 1534: Why do you have this loop?  You should be able to `get_vector_get` using the specified `index`.

src/odb.c
```diff
@@ -1521,6 +1521,25 @@ void git_odb_backend_data_free(git_odb_backend *backend, void *data)
 	git__free(data);
 }
 
+int git_odb_set_backend_priority(git_odb *db, size_t index, int priority)
+{
+	size_t i;
+	assert(db);
+
+	if (index > db->backends.length - 1) {
```
ethomson, line 1529: When `db->backends.length == 0`, this subtraction will make it `SIZE_MAX` and out-of-bounds `index` values will be let through.

ethomson:
Thanks for the PR!  I left a few comments inline.

src/odb.c
hackworks: Once we sort the vector at the end of this function, we can assume the backends will always be sorted based on priority. This will allow us to get the entry at a given index.

src/odb.c
tonydelanuez: @ethomson good point! I addressed this in a new commit by asserting the db->backends.length.

src/odb.c
tonydelanuez: @hackworks Great call. I've gone and updated this in a new commit to manually set the vector sorted flag to false and triggered a sort manually. Thank you 👍

src/odb.c
tonydelanuez: Fixed this in a new commit with a call to `git_vector_get`.

src/odb.c
tonydelanuez: 🤦 My previous fix won't work for release builds. Updated with more logic based on fetching the object from the vector. Thanks again!

tonydelanuez:
@hackworks I thought about that as well - it took some mind-mapping to come up with a correct test for this change for that reason. 

This function is most useful when you've got the default backends with their default priorities (as defined in the macros) and maybe 1-2 more custom backends. Since you know the priority of the backends, and set the priority of the customs, you know the initial vector ordering. 

After rearranging, it's a bit like [playing the shell game](https://en.wikipedia.org/wiki/Shell_game). 
I'm open to ideas here, and totally agree with your point.

tonydelanuez:
I've changed this up a bit to make the interaction cleaner as a result of the above conversation. @hackworks and I agreed that allowing for the overriding of the default backends' priority values serves the same purpose as the original code in this PR but with a much better API. 

A client can now set `git_libgit2_opts(GIT_OPT_SET_ODB_LOOSE_PRIORITY, <VALUE>);` in order to change the default priority of the loose backend or  `git_libgit2_opts(GIT_OPT_SET_ODB_PACKED_PRIORITY, <VALUE>);` to change the default priority of the packed backend.

ethomson:
Interesting pivot!  I hadn't considered this way to solve this problem.  I don't have any immediate objections.  @pks-t care to weigh in?

include/git2/common.h
```diff
@@ -205,7 +205,9 @@ typedef enum {
 	GIT_OPT_GET_PACK_MAX_OBJECTS,
 	GIT_OPT_SET_PACK_MAX_OBJECTS,
 	GIT_OPT_DISABLE_PACK_KEEP_FILE_CHECKS,
-	GIT_OPT_ENABLE_HTTP_EXPECT_CONTINUE
+	GIT_OPT_ENABLE_HTTP_EXPECT_CONTINUE,
+	GIT_OPT_SET_ODB_PACKED_PRIORITY,
+	GIT_OPT_SET_ODB_LOOSE_PRIORITY
```
pks-t, line 210: I'm sorry to be late to the game and propose a different angle, but I don't think having this as a global option is a sensible thing to do. Talking about this in a matter of priorities feels really indirect to me and not that great from a usability point of perspective. That's not your fault but ours, but I'd like to provide a better interface, especially as we will have to do the same for reference databases at some point in time.

Our existing implementation doesn't really allow for some special usecases.

- Somebody who never wants any loose objects at all may want to use a packed ODB implementation only, or vice versa.
- It's hard to couple custom backends with given ones.
- We do not easily allow for creation of repos with a custom set of ODB/refbd implementations.

So in the future, I have the intent to extend ODB/refdbs with some kind of constructor that allows it to say "Here's a new repo that wasn't initialized yet. Do your thing and set up data structures you expect the gitdir to have." That will require some refactoring, as we've got that logic all over the place right now for refdbs, and I expect it's the same for the ODB. But most importantly, this cannot be solved via global priorities. While it would work for our builtin ones, it wouldn't for external ones.

So what do I expect it to look like in the future? I'm not quite sure yet, to be honest. But I think that the safest way to approach this right now is to add a field to `git_repository_open_options` that allows users to pass in an array of object/reference database constructors. A similar thing could be done for `git_repository_new`: pass in a set of constructors that, given a gitdir path, set up their data structures and return a new refdb/odb. Priorities would then be derived from the index of that array.

What do you think?

hackworks: Do you thinking having repository open taken in a vector of backends meet the requirement? The contents and order in the vector will define the backends as requested by the caller. It allows adding user implemented object backends too.

Since we want to support user implemented backends too, it might be best to avoid `git_repository_open_options`. IIUC, any option should be defined in libgit2 and this will prevent treating user implemented object backends as first class citizens.

In short, I am proposing a new function to take in a vector of backends and register it as object backend in the same order as defined in the vector. If there is an alternates file, register the same backends as alternates.

Looking forward to hearing your feedback.

hackworks: Taking in an array/vector of backends would work. However, we could lose the ability to lazy instantiate the backends like it happens today for bare repositories.

If a user just wants to control the order of default backend and make it behave like the git binary, we will be forced to instantiate the loose and pack backends before opening the repository.

pks-t:
On Mon, May 04, 2020 at 05:31:42PM -0700, Dhruva Krishnamurthy wrote:
> @hackworks commented on this pull request.
> 
> 
> 
> > @@ -205,7 +205,9 @@ typedef enum {
>  	GIT_OPT_GET_PACK_MAX_OBJECTS,
>  	GIT_OPT_SET_PACK_MAX_OBJECTS,
>  	GIT_OPT_DISABLE_PACK_KEEP_FILE_CHECKS,
> -	GIT_OPT_ENABLE_HTTP_EXPECT_CONTINUE
> +	GIT_OPT_ENABLE_HTTP_EXPECT_CONTINUE,
> +	GIT_OPT_SET_ODB_PACKED_PRIORITY,
> +	GIT_OPT_SET_ODB_LOOSE_PRIORITY
> 
> Do you thinking having repository open taken in a vector of backends
> meet the requirement? The contents and order in the vector will define
> the backends as requested by the caller. It allows adding user
> implemented object backends too.

I think that should work, yes.

> Since we want to support user implemented backends too, it might be
> best to avoid `git_repository_open_options`. IIUC, any option should
> be defined in libgit2 and this will prevent treating user implemented
> object backends as first class citizens.

I'd argue exactly the other way round. By using global options, you
force every single repository to be opened with the same vector of
backends, which may not be wanted and is thus inflexible design. By
adding a new field to `git_repository_open_options`, developers can pick
per repository what they actually want.

I don't quite get why you think it's not possible to treat
user-implemented backends if using a vector here. It should work just
fine -- users pass in vectors of anything that implements
`git_odb_backend` and we don't care what subtype it is.

> In short, I am proposing a new function to take in a vector of
> backends and register it as object backend in the same order as
> defined in the vector. If there is an alternates file, register the
> same backends as alternates.

I'd still argue it should be part of `git_repository_open_opts`. There's
one thing we're missing right now, though, and that's probably where the
above statement ("We cannot treat user backends the same") stems from:
some kind of constructor. Instead of passing in object databases
themselves, we should instead allow the user to pass in a vector of
functions that, given a gitdir, should inspect the gitdir, see if it is
something the backend can handle, and return an `git_odb` in case it
can.

> Taking in an array/vector of backends would work. However, we could
> lose the ability to lazy instantiate the backends like it happens
> today for bare repositories.

With the above design of using constructors, we are still lazily
initializing the backend.

I hope my proposal makes sense to you?

ethomson:
Sorry to let this sit @tonydelanuez - I think this is a fine approach.  :shipit:

ethomson:
I rebase this to fix up the conflicts and manually merged it.  Thanks @tonydelanuez!

PR #5494, opened 2020-04-18
Fix binary diff showing /dev/null

### LINKED ISSUE #5493: Binary diff showing /dev/null
### Reproduction steps

1. Use the "create_repo.sh" file as a BASH script to construct a minimal repo for testing purposes.
2. Then compile the "main.c" client to link with libgit2 and demonstrate the behaviour on the minimal repo.

[files.zip](https://github.com/libgit2/libgit2/files/4496495/files.zip)

### Expected behavior

<pre>
diff --git a/ascfile b/ascfile
index 46e25a7..249c1f3 100644
--- a/ascfile
+++ b/ascfile
@@ -1 +1 @@
-Hello World 1
+Hello World 2!
diff --git a/binfile b/binfile
index ff08ca1..<strong>f43e1ec</strong> 100755
<strong>Binary files a/binfile and b/binfile differ</strong>
</pre>

### Actual behavior

<pre>
diff --git a/ascfile b/ascfile
index 46e25a7..249c1f3 100644
--- a/ascfile
+++ b/ascfile
@@ -1 +1 @@
Hello World 1
Hello World 2!
diff --git a/binfile b/binfile
index ff08ca1..<strong>0000000</strong> 100755
<strong>Binary files a/binfile and /dev/null differ</strong>
</pre>

1.0 (commit 918a7d19553a9a4181bf1561b8b828a5f82e60e0)

Debian 64-bit, RISC OS 5.27

kevinjswinton:
Fixes #5493 where a changed binary file's content in the working tree isn't displayed correctly, instead showing an OID of zero, and with its path being reported incorrectly as "/dev/null".

Initial focus was on the `maybe_modified` function in `diff_generate.c`: the binary file is stat'd, but because `modified_uncertain` is set to 0, `maybe_modified` does not (further down) calculate the relevant OID. Setting `modified_uncertain` to 1 rectifies the issue, but the test suite suggests an increased load on OID calculations which I suspect is best avoided where possible.

Instead, `patch_generate.c` can be adjusted to fix the problem by ensuring that for binary files both old **and** new files are loaded. This PR supplies that fix.

Note: with this fix in place, two tests then failed. On examination, both tests were seen to be deliberately broken *because* of this outstanding issue, so this PR also supplies the relevant corrections to the broken tests.

### DIFF
--- a/src/patch_generate.c
+++ b/src/patch_generate.c
@@ -209,9 +209,7 @@ static int patch_generated_load(git_patch_generated *patch, git_patch_generated_
 
 	if ((error = git_diff_file_content__load(
 			&patch->ofile, &patch->base.diff_opts)) < 0 ||
-		should_skip_binary(patch, patch->ofile.file))
-		goto cleanup;
-	if ((error = git_diff_file_content__load(
+	    (error = git_diff_file_content__load(
 			&patch->nfile, &patch->base.diff_opts)) < 0 ||
 		should_skip_binary(patch, patch->nfile.file))
 		goto cleanup;
--- a/tests/diff/format_email.c
+++ b/tests/diff/format_email.c
@@ -487,7 +487,7 @@ void test_diff_format_email__binary(void)
 	"Subject: [PATCH] Modified binary file\n" \
 	"\n" \
 	"---\n" \
-	" binary.bin | Bin 3 -> 0 bytes\n" \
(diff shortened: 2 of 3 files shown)


ptrait0695:
thankyou for solving this #kevinjswinton. I couldn't find a solution on stackoverflow

ethomson:
LGTM.  Thanks for the fix!

ethomson (approved):

PR #5495, opened 2020-04-21
libxdiff upgrade in sync with git 2.26.2

j143:
Upgrading libxdiff to the latest version of the core git repository.
Along with the changes made from the core git repo,

Some changes made to `xpatience.c` and `xmerge.c` are also considered.

I will clean the history in a while.

### DIFF
--- a/src/diff_xdiff.c
+++ b/src/diff_xdiff.c
@@ -259,5 +259,5 @@ void git_xdiff_init(git_xdiff_output *xo, const git_diff_options *opts)
 	if (flags & GIT_DIFF_MINIMAL)
 		xo->params.flags |= XDF_NEED_MINIMAL;
 
-	xo->callback.outf = git_xdiff_cb;
+	xo->callback.out_hunk = git_xdiff_cb;
 }
--- a/src/diff_xdiff.c
+++ b/src/diff_xdiff.c
@@ -259,5 +259,5 @@ void git_xdiff_init(git_xdiff_output *xo, const git_diff_options *opts)
 	if (flags & GIT_DIFF_MINIMAL)
 		xo->params.flags |= XDF_NEED_MINIMAL;
 
-	xo->callback.out_hunk = git_xdiff_cb;
+	xo->callback.out_line = git_xdiff_cb;
 }
--- a/src/patch_generate.c
+++ b/src/patch_generate.c
@@ -753,14 +753,14 @@ git_diff_driver *git_patch_generated_driver(git_patch_generated *patch)
 }
 
 void git_patch_generated_old_data(
-	char **ptr, size_t *len, git_patch_generated *patch)
+	char **ptr, long int *len, git_patch_generated *patch)
 {
 	*ptr = patch->ofile.map.data;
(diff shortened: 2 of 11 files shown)


ethomson:
We should share the xdiff with them, _however_ this reverts a number of changes that we made to decrease the likelihood of integer overflows.  What we really need to do is unify with git on a single, shared piece of source for xdiff.  Our utility functions are not identical (`xmalloc` or whatever vs `git__malloc`) but they should be close enough that we can stub them out in a common way for the shared bit so that each project need only provide the translation layer between the shared abstraction and their own utility functions.

This is something that we're also going to want for the shared reftable code.  It seems like `xdiff` is a good place to get that sorted out before we attack reftable, since we're already both relying upon it.

@dscho showed some interest in working on this from the git side -- the integer hardening, at least -- so I'm cc'ing him, though I suspect that this is not the highest priority for him.

Would this be a thing that you'd be interested in attacking, @j143?

dscho:
I am certainly interested! These days, though, I am *so* out of time, I will gladly let other people handle this.

BTW AFAIR Mercurial also has an xdiff fork. And I was under the impression that upstream libxdiff stalled developing?

ethomson:
> BTW AFAIR Mercurial also has an xdiff fork. And I was under the impression that upstream libxdiff stalled developing?

Yes to both.  We discussed briefly at the summit and believe that hg uses git's as canonical.

j143:
Submodule created from [git/git](https://github.com/git/git) repo is at [j143/xdiff](https://github.com/j143/xdiff)

1. https://github.com/git/git/pull/765
2. https://github.com/libgit2/libgit2/pull/5498

dscho:
Seeing as Mercurial does follow git.git, but for obvious reasons does not use submodules, I wonder whether we can do without them here, too.

It's not only that submodules are _so_ awkward to use, it's also that they have proven to have serious security problems, by design. The _vast_ majority of Git's vulnerabilities require submodules in the practical attack vectors.

The problem is so obvious that there are already movements forming to deprecate submodules.

Besides, libgit2 itself does not support submodules out of the box, does it? So why require one in libgit2's own source code?

It might be better to extract xdiff's history via `git filter-branch` and then start to maintain it like gitk, via subtree merges.

But all of this would be better discussed on the Git mailing list. The decision will have to be made there, after all.

They won't see this here discussion, which is a bit sad.

ethomson:
> Seeing as Mercurial does follow git.git, but for obvious reasons does not use submodules, I wonder whether we can do without them here, too.

Yeah, I don't want submodules in the libgit2 code base.  The hard part isn't getting the code out of git, it's the changing of git's code to be usable in the context of libgit2.  We need a mechanism by which the _same literal file_ is shared between the two projects.  How we share that file isn't what needs to be solved, we can copy it manually.

> Besides, libgit2 itself does not support submodules out of the box, does it?

We do support submodules, yes.

ethomson:
I'd be happy to to get closer to git's implementation, but since there's been no updates here based on the given feedback, I'm closing this.

PR #5498, opened 2020-04-22
Xdiff as submodule

j143:
convert xdiff to submodule, present base is from `git/git`. Need to apply the `libgit2` commits on a case by case basis.


j143:
Submodule option is not preferred. 😅 

Subtree option would be good.

PR #5500, opened 2020-04-25
MSVC: Enable Control Flow Guard (CFG)

phkelley:
This feature requires Visual Studio 2015 (`MSVC_VERSION` = 1900) or later. As the minimum required CMake version is currently less than 3.7, `GREATER_EQUAL` is not available to us and we must invert the result of the `LESS` operator.


ethomson:
Thanks @phkelley !

PR #5503, opened 2020-05-07
Modified `config_file_open()` to work in macOS sandboxing.

wilshipley:
- Modified `config_file_open()` so it returns GIT_ENOTFOUND if the config file is not *readable*, which happens on global config files under macOS sandboxing (note that for some reason `access(..., F_OK)` DOES work with sandboxing, but it is lying). Without this read check sandboxed applications on macOS can not open any repository, because `config_file_read()` will return GIT_ERROR when it cannot read the global /Users/username/.gitconfig file, and the upper layers will just completely abort on GIT_ERROR when attempting to load the global config file, so no repositories can be opened.

### DIFF
--- a/include/git2/sys/features.h
+++ b/include/git2/sys/features.h
@@ -1,0 +1,44 @@
+#ifndef INCLUDE_features_h__
+#define INCLUDE_features_h__
+
+//#define GIT_DEBUG_POOL 1
+//#define GIT_TRACE 1
+#define GIT_THREADS 1
+//#define GIT_MSVC_CRTDBG 1
+
+#define GIT_ARCH_64 1
+//#define GIT_ARCH_32 1
+
+#define GIT_USE_ICONV 1
+//#define GIT_USE_NSEC 1
+//#define GIT_USE_STAT_MTIM 1
+//#define GIT_USE_STAT_MTIMESPEC 1
+#define GIT_USE_STAT_MTIME_NSEC 1
+//#define GIT_USE_FUTIMENS 1
+
+#define GIT_REGEX_REGCOMP_L
+//#define GIT_REGEX_REGCOMP
+//#define GIT_REGEX_PCRE
+//#define GIT_REGEX_PCRE2
+//#define GIT_REGEX_BUILTIN 1
+
+//#define GIT_SSH 1
+//#define GIT_SSH_MEMORY_CREDENTIALS 1
+
+//#define GIT_NTLM 1
+//#define GIT_GSSAPI 1
+//#define GIT_GSSFRAMEWORK 1
+
+//#define GIT_WINHTTP 1
+#define GIT_HTTPS 1
+//#define GIT_OPENSSL 1
+#define GIT_SECURE_TRANSPORT 1
+//#define GIT_MBEDTLS 1
+
+//#define GIT_SHA1_COLLISIONDETECT 1
+//#define GIT_SHA1_WIN32 1
+#define GIT_SHA1_COMMON_CRYPTO 1
+//#define GIT_SHA1_OPENSSL 1
+//#define GIT_SHA1_MBEDTLS 1
+
+#endif
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -108,10 +108,12 @@ static int config_file_open(git_config_backend *cfg, git_config_level_t level, c
 	if ((res = git_config_entries_new(&b->entries)) < 0)
 		return res;
 
-	if (!git_path_exists(b->file.path))
+	if (!git_path_exists(b->file.path)) {
 		return 0;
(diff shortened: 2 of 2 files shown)


ethomson:
What does `git` itself do here?

wilshipley:
I don’t know — I’ve never tried to run `git` from inside a sandboxed app. Since it’s a command-line developer tool it’s not sandboxed by default. But libgit2 simply can't function without this if it’s linked into a sandboxed app right now. It seems like the architecture was built to support failing gracefully, since it does a check for `F_OK` first and just returns if `F_OK` returns false. The problem is that the sandbox always returns positive for `F_OK`, even if you cannot access the file in reality, so then later when libgit2 attempts to actually read from the file it hits errors and bails on opening the repository all together (which I would argue is a bit conservative — the architecture is otherwise resilient against missing a `.gitconfig` file).

wilshipley:
A more comprehensive solution to this wold be to modify the caller of `config_file_read()` so it does NOT abort opening the repository when a `.gitconfig` file isn’t read (has an I/O error or is mis-formatted, say), since they are optional.

I went for the smallest change I could make to get this working, but it’s not necessarily the best for libgit2.

wilshipley:
I added some braces in hopes of pleasing the AMD / Windows compiler. I’ve got to admit I’m not super familiar with C++ warnings.

include/git2/sys/features.h
```diff
@@ -0,0 +1,44 @@
+#ifndef INCLUDE_features_h__
```
pks-t, line 1: You definitely don't want to commit this file ;)

src/config_file.c
```diff
@@ -108,8 +108,12 @@ static int config_file_open(git_config_backend *cfg, git_config_level_t level, c
 	if ((res = git_config_entries_new(&b->entries)) < 0)
 		return res;
 
-	if (!git_path_exists(b->file.path))
+	if (!git_path_exists(b->file.path)) {
 		return 0;
+    }
+    if (p_access(b->file.path, R_OK) < 0) {
+        res = GIT_ENOTFOUND; // without this read check sandboxed applications on macOS can't open any repository, because `config_file_read()` below will return GIT_ERROR when it can't read the global /Users/username/.gitconfig file, and the upper layers will just completely abort on GIT_ERROR when loading a config file (which seems a bit extreme).
+    }
```
pks-t, line 116: Some style tips:

- we typically avoid braces around single-line statements
- we use tabs for indentation
- we only use C-style comments (`/* */`)

Simulating this behaviour with git is in fact quite trivial:

```bash
$ git config --list
user.name = Patrick Steinhardt
...
$ chmod u-r ~/.config/git/config
$ git config --list
```

So git continues to work just fine in case it can't read a given file and won't even raise a warning. So I'd argue that instead of setting `res = GIT_ENOTFOUND`, we should just `return 0;` directly similar to how we treat a non-existing path.

pks-t: Oh, and to avoid wasting your time, you should probably format the comment like following:

```C
if (p_access(b->file.path, R_OK) != 0) {
	/*
	 * Git will silently ignore any configuration files that are not readable,
	 * and so should we. This is also required for sandboxed applications on
	 * macOS, where the global git config file will not be readable by libgit2.
	 */
	return 0;
}
```

ethomson:
> But libgit2 simply can't function without this if it’s linked into a sandboxed app right now.

Yeah - sorry, for more context, I was wondering if this was something that we should make optional.  If git itself errors on unreadable configuration files (generally) then we probably want to make this opt-in.  But git is also resilient in the face of unreadable configuration files and silently ignores them:

```
zoidberg:~% chmod 0000 .gitconfig
zoidberg:~% cat .gitconfig
cat: .gitconfig: Permission denied
zoidberg:~% git config --list --show-origin
zoidberg:~% chmod 0644 .gitconfig
zoidberg:~% cat .gitconfig
...configuration data goes here...
zoidberg:~% git config --list --show-origin
file:/Users/ethomson/.gitconfig ...configuration data goes here...
```

So I'm very much 👍 on this change.

> A more comprehensive solution to this wold be to modify the caller of config_file_read() so it does NOT abort opening the repository when a .gitconfig file isn’t read (has an I/O error or is mis-formatted, say), since they are optional.

I think that we want to keep erroring on poorly formatted configuration files for compatibility with git.  If that's something that you're interested in then I do think that we should put that behind an option (but it sounds like keeping things sandbox-friendly is sufficient?)

I agree with @pks-t's stylistic requests, but otherwise this looks good to me.

Thanks @wilshipley!

ethomson:
@wilshipley I hope you don't mind, I wanted to expedite getting this 🚢 so I opened #5527.

PR #5509, opened 2020-05-10
Introduce GIT_ASSERT macros

ethomson:
This PR includes _only_ the `GIT_ASSERT` macros from #5327, not any of the conversions to use them.  This is the first in a multi-step breakdown of #5327 for simplicity.


src/assert_safe.h
```diff
@@ -0,0 +1,58 @@
...
+
+# define GIT_ASSERT(expr) assert(expr)
+# define GIT_ASSERT_ARG(expr) assert(expr)
+
+# define GIT_ASSERT_WITH_RETVAL(expr, fail) assert(expr)
+# define GIT_ASSERT_ARG_WITH_RETVAL(expr, fail) assert(expr)
+#else
+
+/** Internal consistency check to stop the function. */
+# define GIT_ASSERT(expr) GIT_ASSERT_WITH_RETVAL(expr, -1);
```
pks-t, line 30: This and `GIT_ASSERT_ARG` both have trailing semicolons which shouldn't be there.

ethomson: Oops, correct.  Thanks!

PR #5511, opened 2020-05-11
Update package.json

suhaibmujahid:
Change the outdated version in the `package.json` file to the current version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "libgit2",
-  "version": "0.27.0",
+  "version": "1.0.0",
   "repo": "https://github.com/libgit2/libgit2",
   "description": " A cross-platform, linkable library implementation of Git that you can use in your application.",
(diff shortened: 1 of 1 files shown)


pks-t:
Thanks a lot for updating this! Honestly, my first reaction is to just remove the file altogether as seemingly not a lot of people care. If they did, we would've gotten complaints earlier. But you doing the update may indicate that some actually do, so I wonder whether you actually use clib or if this is just something you stumbled over.

If we decide we want to keep it because there are some real users, would you mind also updating `scripts/release.py` to verify its version matches the new release's version? Like this, we won't forget to update in the future.

suhaibmujahid:
Thank you @pks-t. Actually I don't use `clib` yet but I was trying it out and I found out that the version is outdated.

I am ready to update `scrpts/release.py` if you decide to keep the file.

ethomson:
Yeah, let's keep this for now.  @suhaibmujahid if you want to update the python release script, that would be great!

PR #5512, opened 2020-05-11
README.md: Add instructions for building in MinGW environment

A-Ovchinnikov-mx:
Add instructions for building libgit2 in MinGW environment


ethomson:
I think that @A-Ovchinnikov-mx is the expert here.  Certainly these seem reasonable to me.

Thanks @A-Ovchinnikov-mx!

A-Ovchinnikov-mx:
I'm not an expert tbh, but this appears to be rather simple to fix 😄

PR #5515, opened 2020-05-12
tests: checkout: fix flaky test due to mtime race

### LINKED ISSUE #5514: Flaky testcase checkout::index::can_disable_pathspec_match
Starting with #5378, the newly introduced testcase checkout::index::can_disable_pathspec_match is flaky on my machine:

```bash
 ~/Development/libgit2/build:master $ ./libgit2_clar -scheckout::index::can_disable_pathspec_match
Loaded 386 suites:
.

 ~/Development/libgit2/build:master $ ./libgit2_clar -scheckout::index::can_disable_pathspec_match
Loaded 386 suites:
F

  1) Failure:
checkout::index::can_disable_pathspec_match [../tests/checkout/index.c:146]
  testrepo/test10.txt
  file content mismatch at byte 0
```

Bisecting the flaky failure points to commit 24bd12c4f91f841b70bd65252aa25b221bf63eb5, which is rather boring as that commit simply introduced the test to our tree.

Ain't got the time to debug the cause right now, so I thought I'd create an issue for now to document it.

pks-t:
The test checkout::index::can_disable_pathspec_match creates a set of
files, rewrites them and then checks them out again with a pathspec with
the expectancy that a subset of files matching the pathspec is restored
to their original contents. One implementation detail of the test is
that both original content and modified content have the same number of
bytes. As a result, we'll only regard these files as changed in case
their mtime doesn't match what's in the index. And, if we do not think a
file was modified, we'll not restore its contents, resulting in a racy
test setup.

Fix the issue by using differently sized contents for original and
modified content. Like this, the stat cache will always be different and
not depend on the mtime anyomre, fixing the raciness.

Fixes #5514

### DIFF
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -217,9 +217,10 @@ static bool checkout_is_workdir_modified(
 	ie = git_index_get_bypath(data->index, wditem->path, 0);
 
 	if (ie != NULL &&
-		git_index_time_eq(&wditem->mtime, &ie->mtime) &&
-		wditem->file_size == ie->file_size &&
-		!is_filemode_changed(wditem->mode, ie->mode, data->respect_filemode)) {
+	    !git_index_entry_newer_than_index(ie, data->index) &&
+	    git_index_time_eq(&wditem->mtime, &ie->mtime) &&
+	    wditem->file_size == ie->file_size &&
+	    !is_filemode_changed(wditem->mode, ie->mode, data->respect_filemode)) {
 
 		/* The workdir is modified iff the index entry is modified */
 		return !is_workdir_base_or_new(&ie->id, baseitem, newitem) ||
--- a/tests/checkout/index.c
+++ b/tests/checkout/index.c
@@ -91,24 +91,18 @@ void test_checkout_index__can_remove_untracked_files(void)
 
 void test_checkout_index__can_disable_pathspec_match(void)
 {
-	static git_index *index;
-	git_oid commit_id;
-	git_checkout_options g_opts = GIT_CHECKOUT_OPTIONS_INIT;
-	git_object *g_object;
-
 	char *files_to_checkout[] = { "test10.txt", "test11.txt"};
-	size_t files_to_checkout_size = 2;
+	git_checkout_options opts = GIT_CHECKOUT_OPTIONS_INIT;
+	git_object *objects;
+	git_index *index;
 
 	/* reset to beginning of history (i.e. just a README file) */
-	g_opts.checkout_strategy =
-		GIT_CHECKOUT_FORCE | GIT_CHECKOUT_REMOVE_UNTRACKED;
+	opts.checkout_strategy = GIT_CHECKOUT_FORCE | GIT_CHECKOUT_REMOVE_UNTRACKED;
 
-	cl_git_pass(git_revparse_single(&g_object, g_repo, "8496071c1b46c854b31185ea97743be6a8774479"));
-	cl_git_pass(git_checkout_tree(g_repo, g_object, &g_opts));
-	cl_git_pass(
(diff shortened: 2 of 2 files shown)


ethomson:
Yoinks.  Why isn't the flaky git index handling fixing this issue?

pks-t:
On Wed, May 13, 2020 at 12:56:01AM -0700, Edward Thomson wrote:
> Yoinks.  Why isn't the racy git index handling fixing this issue?

Sorry, I'm at a loss. Which racy git index handling do you actually
mean? I mean it's expected that in some cases, we may mis-detect files
as unmodified in case their time, size and mode matches what we have
cached, so without looking at its contents we can't do much about it,
right?

ethomson:
We should have implemented the [racy git mitigations](https://git-scm.com/docs/racy-git/en)

pks-t:
On Thu, May 14, 2020 at 01:58:14AM -0700, Edward Thomson wrote:
> We should have implemented the [racy git mitigations](https://git-scm.com/docs/racy-git/en)

Interesting, today I learned. Didn't know that the cache effectively
gets ignored in case mtime of index and a given file are the same, but
it makes a lot of sense.

Honestly, I'm surprised the issue happens anyway, as my system
should support nanosecond resolution. And at this fine-grained
resolution, it's quite unlikely to be fast enough to create a real race
here. But I remember we had similar issues in the past where we fixed
racy tests by changing test file's content lengths to work around them.
So maybe this is a real bug in libgit2, where we improperly evaluate
mtime?

pks-t:
On Sat, May 16, 2020 at 03:03:33AM -0700, Patrick Steinhardt wrote:
> On Thu, May 14, 2020 at 01:58:14AM -0700, Edward Thomson wrote:
> Honestly, I'm surprised the issue happens anyway, as my system
> should support nanosecond resolution. And at this fine-grained
> resolution, it's quite unlikely to be fast enough to create a real race
> here. But I remember we had similar issues in the past where we fixed
> racy tests by changing test file's content lengths to work around them.
> So maybe this is a real bug in libgit2, where we improperly evaluate
> mtime?

But indeed, it does somehow happen:

```
$ stat /tmp/libgit2_tests_*/testrepo/{test10.txt,.git/index}
  File: /tmp/libgit2_tests_Johnpn/testrepo/test10.txt
  Size: 9               Blocks: 8          IO Block: 4096   regular file
Device: 13h/19d Inode: 532770      Links: 1
Access: (0644/-rw-r--r--)  Uid: ( 1000/     pks)   Gid: ( 1000/     pks)
Access: 2020-05-16 13:25:51.161782349 +0200
Modify: 2020-05-16 13:25:51.161782349 +0200
Change: 2020-05-16 13:25:51.161782349 +0200
 Birth: -
  File: /tmp/libgit2_tests_Johnpn/testrepo/.git/index
  Size: 217             Blocks: 8          IO Block: 4096   regular file
Device: 13h/19d Inode: 532781      Links: 1
Access: (0644/-rw-r--r--)  Uid: ( 1000/     pks)   Gid: ( 1000/     pks)
Access: 2020-05-16 13:25:51.161782349 +0200
Modify: 2020-05-16 13:25:51.161782349 +0200
Change: 2020-05-16 13:25:51.161782349 +0200
 Birth: -
$ git -C /tmp/libgit2_tests_*/testrepo/ status
 M test10.txt
```

Unlikely, but I guess it's some filesystem semantics or the timer not
being fine-grained enough to record any difference. But yeah, seems like
we fail to do the racy index update correctly in the checkout case.

pks-t:
Okay, there indeed was an issue with how we used the index when trying to determine whether a file was modified during checkout. Thanks a lot @ethomson for calling this out, this is definitely a much better fix than just pampering over it in the test :+1:

PR #5523, opened 2020-05-15
cmake: Sort source files for reproducible builds

### LINKED ISSUE #5521: Build not reproducible
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) front, I came across that libgit2 doesn't build reproducibly.

Here are the CI logs: https://salsa.debian.org/debian/libgit2/-/jobs/661147

It'd be nice if we can have a fix for this? :)

pks-t:
We currently use `FILE(GLOB ...)` in most places to find source and
header files. This is problematic in that the order of files returned
depends on the operating system's directory iteration order and may thus
not be deterministic. As a result, we link object files in unspecified
order, which may cause the linker to emit different code across runs.

Fix this issue by sorting all code used as input to the libgit2 library
to improve the reliability of reproducible builds.

Fixes #5521


utkarsh2102:
Additionally, I also wanted you to know that in Debian, we remove the embedded `deps/` but use the packaged ones instead. I hope this has nothing to do with it or does it?

pks-t:
Huh, that's unfortunate, especially so as I can't reproduce locally
anymore: each run produces the exact same binary on my machine. I
wouldn't think that removing "deps/" should break reproducibility, but
they're kind of fragile, so don't take my word for it.

You probably have some custom patches, right? My recommendation would be
to go from the most vanilla configuration of libgit2, check
reproducability, and then iteratively apply the next batch of
modifcications. Rinse and repeat, until you find the culprit. Likely,
it's already happening in the vanilla configuration.

Maybe hard to do, but if you have a minimal script that shows the
un-reproducible build then that might be very helpful to us to reproduce
the issue.

pks-t:
By the way, this PR should still be merged regardless of whether it
fixes the original issue, as it definitely fixes _some_ unreproducible
builds already. I'll create a separate PR as soon as we find out what
the issue is that Debian is running into.

PR #5526, opened 2020-05-23
git_pool_init: allow the function to fail

ethomson:
`git_pool_init` should be allowed to fail.  At the moment, `git_pool_init` will `assert` that the given arguments are correct.  We would like to use `GIT_ASSERT` instead.

This change is in preparation for removing uses of `assert` and replacing them with our `GIT_ASSERT`, however that requires a function that returns a value.

### DIFF
--- a/src/pack-objects.c
+++ b/src/pack-objects.c
@@ -141,13 +141,10 @@ int git_packbuilder_new(git_packbuilder **out, git_repository *repo)
 	pb = git__calloc(1, sizeof(*pb));
 	GIT_ERROR_CHECK_ALLOC(pb);
 
-	if (git_oidmap_new(&pb->object_ix) < 0)
+	if (git_oidmap_new(&pb->object_ix) < 0 ||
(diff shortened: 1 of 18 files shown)


src/pack-objects.c
```diff
@@ -141,13 +141,10 @@ int git_packbuilder_new(git_packbuilder **out, git_repository *repo)
...
-		goto on_error;
-
-	if (git_oidmap_new(&pb->walk_objects) < 0)
-		goto on_error;
-
-	git_pool_init(&pb->object_pool, sizeof(struct walk_object));
+	if (git_oidmap_new(&pb->object_ix) < 0 ||
+	    git_oidmap_new(&pb->walk_objects) < 0 ||
+	    git_pool_init(&pb->object_pool, sizeof(struct walk_object)) < 0)
+	    goto on_error;
```
pks-t: Nit: indentation of the `goto` is wrong.

PR #5527, opened 2020-05-23
Handle unreadable configuration files

ethomson:
A quick cleanup of #5503, with an added test case.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -111,6 +111,15 @@ static int config_file_open(git_config_backend *cfg, git_config_level_t level, c
 	if (!git_path_exists(b->file.path))
 		return 0;
 
+	/*
+	 * git silently ignores configuration files that are not
+	 * readable.  We emulate that behavior.  This is particularly
+	 * important for sandboxed applications on macOS where the
+	 * git configuration files may not be readable.
+	 */
+	if (p_access(b->file.path, R_OK) < 0)
+		return GIT_ENOTFOUND;
+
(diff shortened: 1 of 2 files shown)


wilshipley:
Yay, thanks for moving forward with this, I got distracted with other burning fires. I’ll update my branch as soon as you push this so we stay in sync.

pks-t:
I've rebased the PR to trigger CI, which seemingly didn't run.

src/config_file.c
```diff
@@ -111,6 +111,15 @@ static int config_file_open(git_config_backend *cfg, git_config_level_t level, c
...
 		return 0;
 
+	/*
+	 * git silently ignores configuration files that are not
+	 * readable.  We emulate that behavior.  This is particularly
+	 * important for sandboxed applications on macOS where the
+	 * git configuration files may not be readable.
+	 */
+	if (p_access(b->file.path, R_OK) < 0)
+		return GIT_ENOTFOUND;
```
pks-t, line 121: I was wondering whether we should return `0` as we do above if the path doesn't exist. But opening repos still works just fine if returning `GIT_ENOTFOUND` here.

pks-t:
Thanks a lot to both of you, @ethomson and @wilshipley!

ethomson:
Thanks @pks-t, and thanks again to you, too @wilshipley - I've got plenty of burning fires of my own so I know the feeling.  🙃

PR #5528, opened 2020-05-23
clar: use internal functions instead of /bin/cp and /bin/rm

ethomson:
clar has historically shelled out to `/bin/cp` to copy test fixtures into a sandbox and `/bin/rm` to clean them up.  This has two deficiencies:

1. `/bin/cp` is slower than simply opening the source and destination and copying them in a read/write loop.  On my Mac, the `/bin/cp` based approach takes ~2:40 for a full test pass.  Using a read/write loop to copy the files ourselves takes ~1:50.  Similarly, `/bin/rm` is slower than doing this internally.  Moving to an internal traversal shaves another 20 seconds off.

   These numbers are less impressive on Linux, but using `sendfile` there makes a notable improvement.

2. It's noisy.  Since the leak detector follows fork/exec, we'll end up running the leak detector on `/bin/cp` and `/bin/rm`.  This would be fine, except that the leak detector spams the console on startup and shutdown, so it adds a _lot_ of additional information to the test runs that is useless.  By not forking and using this internal system, we see much less output.

Since this repo actually has tests and fixtures, I'm opening this PR here and will port it to https://github.com/clar-test/clar once approved.

### DIFF
--- a/tests/clar/fs.h
+++ b/tests/clar/fs.h
@@ -455,19 +455,46 @@ fs_copy(const char *source, const char *_dest)
 }
 
 static void
-fs_rm(const char *source)
+fs_rmdir_helper(const char *path)
 {
-	char *argv[4];
+	DIR *dir;
+	struct dirent *d;
 
-	argv[0] = "/bin/rm";
-	argv[1] = "-Rf";
-	argv[2] = (char *)source;
-	argv[3] = NULL;
+	cl_assert_(dir = opendir(path), "Could not open dir");
+	while ((d = (errno = 0, readdir(dir))) != NULL) {
+		char *child;
 
-	cl_must_pass_(
-		shell_out(argv),
-		"Failed to cleanup the sandbox"
-	);
+		if (!strcmp(d->d_name, ".") || !strcmp(d->d_name, ".."))
+			continue;
+
+		child = joinpath(path, d->d_name, -1);
+		fs_rm(child);
+		free(child);
+	}
+
+	cl_assert_(errno == 0, "Failed to iterate source dir");
+	closedir(dir);
+
+	cl_must_pass_(rmdir(path), "Could not remove directory");
+}
+
+static void
+fs_rm(const char *path)
+{
+	struct stat st;
+
+	if (lstat(path, &st)) {
+		if (errno == ENOENT)
+			return;
+
+		cl_fail("Cannot copy; cannot stat destination");
+	}
+
+	if (S_ISDIR(st.st_mode)) {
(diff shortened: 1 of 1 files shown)


tests/clar/fs.h
```diff
@@ -254,74 +262,212 @@ cl_fs_cleanup(void)
...
+		((void)(source_len)); /* unused */
+
+		while ((ret = read(in, buf, sizeof(buf))) > 0) {
+			size_t len = (size_t)ret;
+
+			while (len && (ret = write(out, buf, len)) > 0) {
+				cl_assert(ret <= (ssize_t)len);
+				len -= ret;
+			}
+			cl_assert(ret >= 0);
```
pks-t, line 385: Nit: this can be `cl_assert(ret == 0)`

tests/clar/fs.h
```diff
@@ -254,74 +262,212 @@ cl_fs_cleanup(void)
...
+		}
+		cl_assert(ret == 0);
+	}
+#endif
+
+	close(in);
+	close(out);
+
```
pks-t, line 393: Nit: trailing newline

tests/clar/fs.h
```diff
@@ -254,74 +262,212 @@ cl_fs_cleanup(void)
...
+	const char *dest;
+	struct stat source_st, dest_st;
+
+	cl_must_pass_(lstat(source, &source_st), "Failed to stat copy source");
+
+	if (lstat(_dest, &dest_st) == 0) {
+		const char *base;
+		int base_len;
+
+		/* Target exists; append basename */
```
pks-t, line 409: I think this is the case where we copy a file into a pre-existing directory, right? In that case, we should also assert `S_ISDIR(dest_st.mode)` so that one gets a proper error message if it isn't.

pks-t:
> It's noisy. Since the leak detector follows fork/exec, we'll end up running the leak detector on /bin/cp and /bin/rm. This would be fine, except that the leak detector spams the console on startup and shutdown, so it adds a lot of additional information to the test runs that is useless. By not forking and using this internal system, we see much less output.

Ah, interesting. I always wondered where all those messages on macOS came from, but that makes a lot of sense to me!

tests/clar/fs.h
ethomson: Writing the last chunk of the buffer will return a positive `ret` (indicating the number of bytes written).  That will be subtracted from `len`, which will now be `0`.  In this case, `ret` will be positive but we should stop the loop successfully.

PR #5531, opened 2020-05-23
mempack: Use threads when building the pack

joshtriplett:
The mempack ODB backend creates a packbuilder internally to write out a
pack; call git_packbuilder_set_threads on that packbuilder, to use
threads for packing if available.


ethomson:
Thanks!

PR #5532, opened 2020-05-23
git_packbuilder_write: Allow setting path to NULL to use the default path

joshtriplett:
If given a NULL path, write to the object path of the repository.

Includes tests.

### DIFF
--- a/src/pack-objects.c
+++ b/src/pack-objects.c
@@ -1385,6 +1385,7 @@ int git_packbuilder_write(
 	void *progress_cb_payload)
 {
 	int error = -1;
+	git_buf object_path = GIT_BUF_INIT;
 	git_indexer_options opts = GIT_INDEXER_OPTIONS_INIT;
 	git_indexer *indexer = NULL;
 	git_indexer_progress stats;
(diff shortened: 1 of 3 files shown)


src/pack-objects.c
```diff
@@ -1384,37 +1384,48 @@ int git_packbuilder_write(
...
 	git_indexer_progress stats;
 	struct pack_write_context ctx;
 	int t;
 
 	PREPARE_PACK;
 
+	if (path == NULL) {
+		if ((error = git_repository_item_path(&object_path, pb->repo, GIT_REPOSITORY_ITEM_OBJECTS)) < 0)
+			goto cleanup;
+		if ((error = git_buf_joinpath(&object_path, git_buf_cstr(&object_path), "pack")) < 0)
```
ethomson, line 1400: The pack dir should probably be a `GIT_REPOSITORY_ITEM` of its own.  I opened https://github.com/libgit2/libgit2/issues/5539 to track.

ethomson:
Thanks @joshtriplett!

PR #5535, opened 2020-05-29
strarray refactoring

ethomson:
We standardized on `dispose` as the verb for freeing the contents of a structure, but not the structure itself.  Thus, `git_strarray_free` should be `git_strarray_dispose`.  While we're here, we shouldn't offer the `git_strarray_copy` function publicly; there's no value in it for users.

### DIFF
--- a/include/git2/deprecated.h
+++ b/include/git2/deprecated.h
@@ -524,6 +524,30 @@ typedef int GIT_CALLBACK(git_headlist_cb)(git_remote_head *rhead, void *payload)
 
 /**@}*/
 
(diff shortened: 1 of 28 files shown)


include/git2/deprecated.h
```diff
@@ -524,6 +524,42 @@ typedef int GIT_CALLBACK(git_headlist_cb)(git_remote_head *rhead, void *payload)
 
 /**@}*/
 
+/** @name Deprecated String Buffer Functions
```
pks-t, line 527: This should probably read "Deprecated String Array Functions"

PR #5536, opened 2020-06-01
httpclient: support googlesource

### LINKED ISSUE #5525: Failure to clone repository when `git` CLI succeeds
### Reproduction steps

Build and run this program:

```c
#include <assert.h>
#include <git2/clone.h>
#include <git2/global.h>
#include <git2/errors.h>
#include <stdio.h>

int main() {
  assert(git_libgit2_init() >= 0);
  git_repository *repo;
  const char *url = "https://chromium.googlesource.com/chromium/llvm-project/llvm/lib/Fuzzer";
  const char *local = "local-dir";
  int rc = git_clone(&repo, url, local, NULL);
  if (rc == GIT_OK) {
    printf("ok!\n");
    return 0;
  }
  const git_error *err = git_error_last();
  printf("rc %d\n", rc);
  printf("error %s\n", err->message);
  printf("klass %d\n", err->klass);
}
```

compared to:

```
$ git clone https://chromium.googlesource.com/chromium/llvm-project/llvm/lib/Fuzzer
Cloning into 'Fuzzer'...
remote: Total 4444 (delta 3448), reused 4444 (delta 3448)
Receiving objects: 100% (4444/4444), 669.15 KiB | 3.05 MiB/s, done.
Resolving deltas: 100% (3448/3448), done.
$ echo $?
0
```

### Expected behavior

The same result of cloning for libgit2 vs the `git` CLI

### Actual behavior

It looks like libgit2 fails with an inflate error, but the `git` CLI succeeds

b7b872f51074ef8231ae8b94dd93eada127c9365 (current master)

Linux (Ubuntu 18.04)

---

This was originally reported at https://github.com/rust-lang/cargo/issues/8258 and I'm forwarding here because I'm not sure how best to next debug this. Is there a way to extract out the stream of data to inspect whether it's actually an invalid zlib stream? Is this perhaps a case where the git CLI isn't checking an error code?

Some debugging showed that the inflation call [failed here](https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/inflate.c#L1212-L1213) which seems like it's a maybe-optional check (that seems like a checksum?) which the `git` CLI may be ignoring.

I'm mostly curious if libgit2 folks can help debug this to see if it's an issue with this library, the server, or the `git` CLI

ethomson:
After the [great smart http refactoring of 2020](https://github.com/libgit2/libgit2/pull/5286), we [no longer support googlesource](https://github.com/libgit2/libgit2/issues/5525).  It turns out that google code - by default - sends very large data packets (65520) which is very neat to the upper limit of a packet size for a git data packet over the smart protocols.  This illustrates some problems in our handling of very large data packets.

In particular, we read a block of data from the client and try to put it in the output buffer.  When the output buffer (for data packets, this is a buffer of 65536) is nearly full, we do not take that into account and do a read of a block, despite the fact that we cannot fit it into the caller's output buffer.  We then discard what cannot be sent back.

While debugging this, I found two other problems as well.  This PR fixes the root cause and these other two issues.

* First, this adds an integration test that clones (a small repository) from googlesource that illustrates our deficiencies.
* Next, we ensure that `git_http_client_read_body` will always return `0` at the end of an http response.  If a buffer size that is being read into is very near the number of bytes sent by the remote server, then we may be in a position where we have read all the content in the http response, and the only bytes remaining are part of the http metadata, like the zero-length chunk signifier of the end of the stream. In that case, we do not actually return _content bytes_, but should signify that we have finished reading the stream.  Identify this by the `on_message_complete` callback and return `0` to the caller.
* Next, we only read _at most_ what the client has requested.  `git_http_client_read_body` takes a buffer size that should be respected when reading from the stream.  Without this, we would read data from the server, but not return it to the client, which would mean that it was lost.
* Finally, we should clear the interim read buffer when starting a new request.  The read buffer is used when we read data from the remote server that contains both headers _and_ body content.  The header data will be returned to the caller (in `read_response`) but the body data should be saved for a subsequent call to `read_body`.  If such call never comes, we should clear the saved data, it should not be returned to callers for future requests.

Fixes #5525

### DIFF
--- a/src/transports/httpclient.c
+++ b/src/transports/httpclient.c
@@ -1419,15 +1419,20 @@ int git_http_client_read_body(
 	client->parser.data = &parser_context;
 
 	/*
-	 * Clients expect to get a non-zero amount of data from us.
-	 * With a sufficiently small buffer, one might only read a chunk
-	 * length.  Loop until we actually have data to return.
+	 * Clients expect to get a non-zero amount of data from us,
+	 * so we either block until we have data to return, until we
+	 * hit EOF or there's an error.  Do this in a loop, since we
+	 * may end up reading only some stream metadata (like chunk
+	 * information).
 	 */
 	while (!parser_context.output_written) {
 		error = client_read_and_parse(client);
 
 		if (error <= 0)
 			goto done;
+
+		if (client->state == DONE)
+			break;
 	}
 
 	assert(parser_context.output_written <= INT_MAX);
--- a/src/transports/httpclient.c
+++ b/src/transports/httpclient.c
@@ -1195,7 +1195,7 @@ static void complete_response_body(git_http_client *client)
 	/* If we're not keeping alive, don't bother. */
 	if (!client->keepalive) {
 		client->connected = 0;
-		return;
+		goto done;
 	}
 
 	parser_context.client = client;
@@ -1209,6 +1209,9 @@ static void complete_response_body(git_http_client *client)
 		git_error_clear();
 		client->connected = 0;
 	}
+
+done:
+	git_buf_clear(&client->read_buf);
 }
 
 int git_http_client_send_request(
--- a/src/transports/httpclient.c
+++ b/src/transports/httpclient.c
@@ -1038,6 +1038,7 @@ on_error:
 
 GIT_INLINE(int) client_read(git_http_client *client)
 {
+	http_parser_context *parser_context = client->parser.data;
 	git_stream *stream;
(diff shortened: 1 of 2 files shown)


alexcrichton:
Oh wow nice find! This looks to fix the Rust side of things that I was testing, thanks for this!

paolobarbolini:
This seems to fix the issue I tested. Thanks!

PR #5541, opened 2020-06-04
clar: add tap output option

ethomson:
This adds a new command-line option to clar, `-t`, which will produce output in the popular TAP format, which can be used by multiple different frontends to display test output.

This retroactively names the existing clar output format as "clap" (the "clar protocol"), and adds support for tap (the [test anywhere protocol](https://testanything.org/)).

This adds support for [TAP version 13](https://testanything.org/tap-version-13-specification.html).

It includes the optional YAML block to provide more information about test failures.  Since this is not yet standardized, we have copied the output format of the popular [node tap](https://node-tap.org/tap-protocol).

### DIFF
--- a/tests/checkout/crlf.c
+++ b/tests/checkout/crlf.c
@@ -108,7 +108,7 @@ done:
 		git_buf details = GIT_BUF_INIT;
 		git_buf_printf(&details, "filename=%s, system=%s, autocrlf=%s, attrs={%s}",
 			git_path_basename(actual_path->ptr), systype, cd->autocrlf, cd->attrs);
-		clar__fail(__FILE__, __LINE__,
+		clar__fail(__FILE__, __func__, __LINE__,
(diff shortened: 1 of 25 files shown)


ethomson:
The last commit adds information about the function that failed.  I think that this is a useful addition _but_ it is a breaking change for clar consumers that use the `clar__` (double-underscored) functions.

I feel like these are internal and so this is probably okay (they're used in older bits of our tests that probably should be doing something less internalsy).  But I think that it's an open question how we would really roll this out in clar itself.  Probably doing a release ahead of it makes sense.

ethomson:
> I don't particularly mind about the breaking-change of adding the function that erred, but we might want to reach out to any other parties that are interested in clar. I think there were a few select people, right? Anyway, the changes all look sane to me, thanks for doing them!

Yeah, I'll try to ping people and / or make enough noise in a PR in the [clar repo](https://github.com/clar-test/clar) that people get the message.

PR #5546, opened 2020-06-05
Refactor "global" state

ethomson:
The notion of "global" state has become somewhat confused over time; `global.c` handles the libraries initialization and shutdown functions, state that's global to the library _and_ thread-local data.  This is an attempt to clean this up, breaking each of these notions out into its own separate section instead of keeping them comingled and simplifying the initialization and shutdown of the library by using our existing patterns in places that we had not done that.

* `git_libgit2_init` sets up global state for a number of subsystems (it sets up the allocator functions, does hash initialization, etc).  _Most_ of these subsystems have their init functions set up in an `init_callbacks` array, and their init functions register shutdown callbacks.  But not all; some have initialization scattered around the `git_libgit2_init` machinery.  Move them all into the same init/shutdown pattern:

  * thread initialization
  * win32 debug stack
  * thread-local data (error state, etc)
  * user-configurable settings (user agent string, etc)
  * mwindow mutex

* The "global state" (`GIT_GLOBAL`) is really _thread-local_ state and contains things like the current thread's error data.  Rename it accordingly, so that it's easier to reason about.

* Moves the current thread's return data on Windows (for `pthread_exit` emulation) out of the shared TLS data structure, and managing it with the rest of the threading initialization/teardown routines.

* Adds a "runtime" layer to handle the init/shutdown functions, so that we `git_runtime_shutdown_register` to register a shutdown handler with the runtime (instead of `git__shutdown_register` - the `git__` name feels like an antipattern)


src/global.c
```diff
@@ -136,42 +103,58 @@ static void shutdown_common(void)
...
 
-	if ((_fls_index = FlsAlloc(fls_free)) == FLS_OUT_OF_INDEXES)
-		return -1;
+#elif defined(GIT_THREADS) && defined(_POSIX_THREADS)
 
-	error = init_common();
+static pthread_mutex_t init_mutex = PTHREAD_MUTEX_INITIALIZER;
 
-	return error;
+GIT_INLINE(int) mutex_lock(void)
```
pks-t: We now have essentially two different implementations of mutexes. Is the "thread-utils.h" implementation dependent on data we haven't initialized yet?

src/tlsdata.c
```diff
@@ -0,0 +1,163 @@
...
+{
+	git_tlsdata *tlsdata;
+
+	if ((tlsdata = FlsGetValue(fls_index)) != NULL)
+		return tlsdata;
+
+	if ((tlsdata = git__calloc(1, sizeof(git_tlsdata))) == NULL)
+		return NULL;
+
+	git_buf_init(&tlsdata->error_buf, 0);
```
pks-t, line 73: `git_buf_init` may return an error, even though it won't if used with `initial_size == 0`. I'd still argue we should check for errors.

src/tlsdata.c
```diff
@@ -0,0 +1,163 @@
...
+
+#elif defined(GIT_THREADS)
+# error unknown threading model
+#else
+
+static git_tlsdata tlsdata;
+
+static void git_tlsdata_global_shutdown(void)
+{
+	tlsdata_dispose(&tlsdata);
```
pks-t, line 134: Does this compile? `tlsdata_dispose` is defined after its use and I can't see a previous declaration.

src/tlsdata.h
```diff
@@ -14,12 +14,6 @@ typedef struct {
 	git_error error_t;
 	git_buf error_buf;
 	char oid_fmt[GIT_OID_HEXSZ+1];
-
-	/* On Windows, this is the current child thread that was started by
-	 * `git_thread_create`.  This is used to set the thread's exit code
-	 * when terminated by `git_thread_exit`.  It is unused on POSIX.
-	 */
-	git_thread *current_thread;
```
pks-t: Very nice to get rid of this.

src/global.c
```diff
@@ -27,153 +27,32 @@
...
+		git_merge_driver_global_init,
+		git_transport_ssh_global_init,
+		git_stream_registry_global_init,
+		git_openssl_stream_global_init,
+		git_mbedtls_stream_global_init,
+		git_mwindow_global_init,
+		git_settings_global_init
+	};
+
+	return git_runtime_init(init_fns, ARRAY_SIZE(init_fns));
```
pks-t, line 52: Hum. You're probably keeping "global.c" and "runtime.c" separate in order to not leak all the different headers into "global.c"? I'd have just moved all of "global.c" into "runtime.c", mostly because I'm not much of a fan of the name "global" and think that "runtime" is a much better name. Given that we now only have `git_libgit2_init()` and `git_libgit2_shutdown()` left, we could also rename this file to "libgit2.c" instead.

src/tlsdata.c
ethomson: It's declared at the top of the file.

src/tlsdata.c
ethomson: Fixed.

src/global.c
ethomson: This was poorly named and a leftover from prior refactorings.  On Windows, we use a spinlock here - we don't use `git_mutex` (which maps to a `CriticalSection`).  This is efficient and has a static initializer (since it operates on a `LONG`).

On Unix, however, we have to use a pthread mutex, since that gives us a static initializer.  (We can't call `pthread_something_init` since the goal of this function is to manage our initialization lifetime.)

I added comments and disambiguated the names to help show what's going on here and why.

src/global.c
ethomson: I moved this to libgit2.c, and moved all the `git_libgit2_` functions into that file for our understanding.

src/tlsdata.c
pks-t: Ah, somehow I didn't see that. :+1:

src/runtime.c
```diff
@@ -0,0 +1,147 @@
...
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#include "common.h"
+#include "runtime.h"
+
+static git_runtime_shutdown_fn shutdown_callback[32];
```
pks-t, line 11: It's a wee bit unfortunate we now have a static limit on shutdowns again. But given that overall this PR is quite an improvement, I'm fine with this tradeoff.

src/transports/ssh.c
```diff
@@ -934,7 +934,7 @@ int git_transport_ssh_global_init(void)
 		return -1;
 	}
 
-	git__on_shutdown(shutdown_ssh);
+	git_runtime_shutdown_register(shutdown_ssh);
```
pks-t, line 937: We're ignoring the function's return code here

src/sysdir.c
```diff
@@ -189,7 +189,7 @@ int git_sysdir_global_init(void)
 	for (i = 0; !error && i < ARRAY_SIZE(git_sysdir__dirs); i++)
 		error = git_sysdir__dirs[i].guess(&git_sysdir__dirs[i].buf);
 
-	git__on_shutdown(git_sysdir_global_shutdown);
+	git_runtime_shutdown_register(git_sysdir_global_shutdown);
```
pks-t, line 192: We're ignoring the function's return code here.

src/win32/thread.c
```diff
@@ -60,7 +61,7 @@ int git_threads_global_init(void)
 	if ((fls_index = FlsAlloc(NULL)) == FLS_OUT_OF_INDEXES)
 		return -1;
 
-	git__on_shutdown(git_threads_global_shutdown);
+	git_runtime_shutdown_register(git_threads_global_shutdown);
```
pks-t, line 64: And here, too.

src/libgit2.c
```diff
@@ -48,3 +76,280 @@ int git_libgit2_shutdown(void)
 {
 	return git_runtime_shutdown();
 }
+
+/* Library settings */
```
pks-t, line 80: Moving stuff like version or feature information definitely makes a ton of sense to me. But I'm not sure I'm a huge fan of moving settings over, too. I kind of get why you're doing it, but if you ask me we should instead work on clearly separating settings and make them more standalone. Like clearly identifying options with `git_setting_user_agent`, `git_setting_ssl_ciphers` etc.

WDYT?

src/libgit2.h
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+#ifndef INCLUDE_libgit2_h__
+#define INCLUDE_libgit2_h__
+
+extern int git_libgit2_settings_global_init(void);
```
pks-t, line 10: This declaration shouldn't be needed at all, right? We're making direct use of it in `libgit2.c`.

src/alloc.c
```diff
@@ -21,8 +26,23 @@ static int setup_default_allocator(void)
...
+}
+#endif
+
 int git_allocator_global_init(void)
 {
+#if defined(GIT_MSVC_CRTDBG)
+	git_win32__crtdbg_stacktrace_init();
+	git_win32__stack_init();
+
+	git__on_shutdown(allocator_global_shutdown);
```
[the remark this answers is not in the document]
ethomson: I started down that road, but `win32/*` is only included in win32 builds, so we'd need an `ifdef` around it in the list of initializers.  I thought this was preferable (but perhaps only slightly.)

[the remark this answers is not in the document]
ethomson: Yeah, thinking about this more, I think I do prefer one ifdef over the several that I added...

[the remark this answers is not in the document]
ethomson: Actually, I just refactored the "msvc crtdbg" stuff entirely, giving it a new name and a consolidated initialization point.

pks-t:
@ethomson It's becoming quite hard to review due to the PRs growing scope. Would it be a huge burden to ask you to split up this PR into multiple ones to get those parts in more quickly? Fine with a "no" if you think the effort for this is too high.

src/sysdir.c
ethomson: Fixed.

src/libgit2.h
ethomson: Fixed.

src/win32/thread.c
ethomson: Fixed.

src/transports/ssh.c
ethomson: Fixed.

ethomson:
> @ethomson It's becoming quite hard to review due to the PRs growing scope. Would it be a huge burden to ask you to split up this PR into multiple ones to get those parts in more quickly? Fine with a "no" if you think the effort for this is too high.

Sure. I split out the renaming of crtdbg into #5580.

src/libgit2.c
ethomson: Agree with you here, and I agree that this is continuing to grow in scope.  Let's tackle this in a followup PR?

pks-t:
On Sun, Jul 12, 2020 at 01:22:45PM -0700, Edward Thomson wrote:
> 
> 
> 
> > @@ -48,3 +76,280 @@ int git_libgit2_shutdown(void)
>  {
>  	return git_runtime_shutdown();
>  }
> +
> +/* Library settings */
> 
> Agree with you here, and I agree that this is continuing to grow in scope.  Let's tackle this in a followup PR?

Fair enough. Is the PR currently in a state where it makes sense to do
another review?

ethomson:
> Fair enough. Is the PR currently in a state where it makes sense to do
> another review?
> 

Yes I think so

ethomson:
I renamed the thread-local data that we carry around as part of our "global" state to "threadstate".  Upon seeing #5602, it's clear that we should have a generic implementation of an internal TLS API so that (eg) checkout can use it and that "tlsdata" is extremely ambiguous in this regard.

ethomson:
Since this is largely an internal refactoring, and I think that I addressed all of the feedback, I'm going to merge this to unblock #5602.

PR #5547, opened 2020-06-05
CMake modernization pt2

pks-t:
This is the next set of CMake modernizations. I've moved us to use the `project(VERSION ...)` keyword, make more use of the git2internal target and switched the HTTPS and hash backend detection to use interface libraries. The last part is probably the most interesting and I consider it to be kind of a test balloon to see whether it works as expected or blow up in our face, but if it works out the plan is to move most of our dependencies to use interface library targets. The benefit is that our main build instructions should become much more readable.


pks-t:
Removed the two commits that convert HTTPS and hash backends to use interface targets, as they broke on some platforms. I'll come back to that at a later point for the sake of having a smaller PR that gets in more quickly.

CMakeLists.txt
```diff
@@ -13,7 +13,7 @@
 
 CMAKE_MINIMUM_REQUIRED(VERSION 3.5.1)
 
-project(libgit2 C)
+project(libgit2 VERSION "1.0.0" LANGUAGES C)
```
ethomson, line 16: 1.0.1 now.  Sorry it took so long to review.

ethomson:
LGTM.  Thanks for tackling this.

CMakeLists.txt
pks-t: Indeed. Rebased the MR and fixed the version, thanks!

Issue #5304, opened 2019-11-15
The submodule.<name>.ignore config is not respected


sschuberth:
Note, this is coming from https://github.com/gitahead/gitahead/issues/289; as such I'm not using `libgit2` myself as a library.

### Reproduction steps

The [gitahead](https://github.com/gitahead/gitahead) client, which uses `libgit2`, lists submodules as "-dirty" even if they are configured to ignore untracked files via [submodule.&lt;name&gt;.ignore](https://git-scm.com/docs/git-config#Documentation/git-config.txt-submoduleltnamegtignore) being set to `ignore = untracked`.

### Expected behavior

I'd expect untracked files being ignored in the status for such submodules, just like "git status" / "git diff" on the command line do.

### Actual behavior

submodules with untracked files as listed as dirty despite the configuration to ignore untracked files.

### Version of libgit2 (release number or SHA1)

[libgit2 @ 16a2faf](https://github.com/stinb/libgit2/tree/16a2fafa341cbdfece21cec6a15ecd0f7bc43a5f)

### Operating system(s) tested

Ubuntu Linux 18.04 64-bit

pks-t:
Yeah, true, we're completely missing support for this. I'll add the "up for grabs" label for now as I have some other things to tackle first before I got the time to fix the issue here.

Thanks for your report!

PR #5549, opened 2020-06-07
submodule: Read superproject configs for `ignore'

clickyotomy:
Introduces an additional step to check configuration values for `ignore` from the parent repository values when reading configuration for a submodule in `submodule_read_config`.

According to gitmodules(5):
> If this option is also present in the submodules entry in .git/config of the superproject, the setting there  will override the one found in .gitmodules.

Ref: https://kernel.org/pub/software/scm/git/docs/gitmodules.html.

Fixes: #5304.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -75,7 +75,7 @@ with v0.28.0.
 
 * Several function signatures have been changed to return an `int` to
   indicate error conditions.  We encourage you to check them for errors
-  in the standard way. 
+  in the standard way.
 
   * `git_attr_cache_flush`
   * `git_error_set_str`
@@ -194,7 +194,7 @@ with v0.28.0.
 
 * Several attribute macros have been renamed: `GIT_ATTR_TRUE` is now
   `GIT_ATTR_IS_TRUE`, `GIT_ATTR_FALSE` is now `GIT_ATTR_IS_FALSE`,
-  `GIT_ATTR_UNSPECIFIED` is now `GIT_ATTR_IS_UNSPECIFIED`.  The 
+  `GIT_ATTR_UNSPECIFIED` is now `GIT_ATTR_IS_UNSPECIFIED`.  The
   attribute enum `git_attr_t` is now `git_attr_value_t` and its
   values have been renamed: `GIT_ATTR_UNSPECIFIED_T` is now
   `GIT_ATTR_VALUE_UNSPECIFIED`, `GIT_ATTR_TRUE_T` is now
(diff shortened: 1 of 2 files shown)


docs/changelog.md
```diff
@@ -1380,7 +1380,7 @@ v0.23
       `git_submodule_set_fetch_recurse_submodules()`.
 
 * `git_submodule_status()` no longer takes a submodule instance but a
-  repsitory, a submodule name and an ignore setting.
+  repository, a submodule name and an ignore setting.
```
ethomson, line 1383: I regret that we made spelling errors _but_ I don't love changing the `changelog` after the fact.  It feels like once we've released the release that we should keep the corresponding changelog static (warts and all).

src/submodule.c
```diff
@@ -1941,7 +1943,17 @@ static int submodule_read_config(git_submodule *sm, git_config *cfg)
 		goto cleanup;
 	}
 
-	if ((error = get_value(&value, cfg, &key, sm->name, "ignore")) == 0) {
+	/* For `ignore', first check the config of the parent repository. */
+	if (git_repository_config_snapshot(&repo_snap_cfg, sm->repo) == 0) {
+		repo_cfg_cleanup = true;
```
ethomson, line 1948: No need to set a boolean here - instead, we can just set `repo_snap_cfg` to `NULL` in its initialization and then `free` it without a test.

src/submodule.c
```diff
@@ -1941,7 +1943,17 @@ static int submodule_read_config(git_submodule *sm, git_config *cfg)
...
+		if (get_value(&value, repo_snap_cfg, &key, sm->name, "ignore") == 0) {
+			override_ignore = true;
+		}
+	}
+
+	ign_cfg = override_ignore ? repo_snap_cfg : cfg;
+
+	if ((error = get_value(&value, ign_cfg, &key, sm->name, "ignore")) == 0) {
```
ethomson, line 1956: It seems like we're looking up the value from `repo_snap_cfg`, then deciding whether we should look up from `repo_snap_cfg` or `cfg` - and then looking it up again?

I think the more natural pattern here would be looking up the value from `cfg` and then trying to look it up from the parent.  If that succeeds, we'll use it.  If not, continue using the one from the `cfg`.

src/submodule.c
clickyotomy: Makes sense. Will initialize `repo_snap_cfg` to `NULL`. Thanks!

src/submodule.c
clickyotomy: I have made a small change with the latest update while preserving the
lookup order. This way, we can use the value that is set in `gitmodules`
(like you suggested) and also return the right errors (if any) when
fetching values either from the parent or the submodule.

If you would still like the `gitmodules` lookup to happen first, I will
make that change in the next patch.

Thanks.

PR #5550, opened 2020-06-07
Introduce CI with GitHub Actions

ethomson:
I've been a bit frustrated with the way we use docker containers in our build.  Previously, we would have to manually pre-build the docker containers any time they changed.  This was - obviously - disappointing because any workflow that involves the world "manual" is incredibly disappointing - but it had the advantage of not needing to build the containers as part of the CI process.  We moved to a process where we build the containers and then run them every time.  This mitigated the manual processes but it increased the build time for our Linux builds rather substantially.

We tried to mitigate this by caching some of the layers of the image, but there's a limited amount of cache space available.  But we _could_ save them instead in a package service colocated with our CI.  (Numbers after the break.)

* This moves our Linux build containers into GitHub Packages; we will identify the most recent commit that updated the docker descriptions, and then look for a docker image in libgit2's GitHub Packages registry for a container with the tag corresponding to that description.  If there is not one, we will build the container and then push it to GitHub Packages.

* We no longer need to manage authentication with our own credentials or PAT tokens.  GitHub Actions provides a GITHUB_TOKEN that can publish artifacts, packages and commits to our repository within a workflow run.

* We will use a matrix to build our various CI steps.  This allows us to keep configuration in a single place without multiple YAML files.

---

On my most recent run, with a cache hit for the containers, the impact is:

* Linux: 15.5 min -> 8.5 min
* macOS: no change
* Windows Visual Studio: 13 min -> 10.5 min
* Windows mingw: 19.5 min -> 16.5 min

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,202 @@
+# Continuous integration and pull request validation builds for the
+# master and maintenance branches.
+name: CI Build
+
+on:
+  push:
+    branches: [ master, maint/* ]
+  pull_request:
+    branches: [ master, maint/* ]
+
+env:
+  docker-registry: docker.pkg.github.com
+  docker-config-path: azure-pipelines/docker
+
+jobs:
+  # Build the docker container images that we will use for our Linux
+  # builds.  This will identify the last commit to the repository that
+  # updated the docker images, and try to download the image tagged with
+  # that sha.  If it does not exist, we'll do a docker build and push
+  # the image up to GitHub Packages for the actual CI/CD runs.  We tag
+  # with both the sha and "latest" so that the subsequent runs need not
+  # know the sha.  Only do this on CI builds (when the event is a "push")
+  # because PR builds from forks lack permission to write packages.
+  build_containers:
+    name: Create docker image
+    strategy:
+      matrix:
+        container:
+        - xenial
+        - bionic
+        - docurium
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+      if: github.event_name == 'push'
+    - name: Download existing container
+      run: azure-pipelines/getcontainer.sh ${{ env.docker-config-path }}/${{ matrix.container }}
+      env:
+        DOCKER_REGISTRY: ${{ env.docker-registry }}
+        GITHUB_TOKEN: ${{ secrets.github_token }}
+      if: github.event_name == 'push'
+    - name: Build and publish image
+      run: |
+        docker build -t ${{ env.docker-registry-container-sha }} --build-arg BASE=${{ matrix.container.base }} -f ${{ matrix.container }} .
+        docker push ${{ env.docker-registry-container-sha }}
+      working-directory: ${{ env.docker-config-path }}
+      if: github.event_name == 'push' && env.docker-container-exists != 'true'
+
+  # Run our CI/CD builds.  We build a matrix with the various build targets
+  # and their details.  Then we build either in a docker container (Linux)
+  # or on the actual hosts (macOS, Windows).
+  build:
+    name: Build
+    needs: [build_containers]
+    strategy:
+      matrix:
+        platform:
+        - # Xenial, GCC, OpenSSL
+          image: xenial
+          env:
+            CC: gcc
+            CMAKE_GENERATOR: Ninja
+            CMAKE_OPTIONS: -DUSE_HTTPS=OpenSSL -DREGEX_BACKEND=builtin -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON
+          os: ubuntu-latest
+        - # Xenial, GCC, mbedTLS
+          image: xenial
+          env:
+            CC: gcc
+            CMAKE_GENERATOR: Ninja
+            CMAKE_OPTIONS: -DUSE_HTTPS=mbedTLS -DUSE_SHA1=HTTPS -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON
+          os: ubuntu-latest
+        - # Xenial, Clang, OpenSSL
+          image: xenial
+          env:
+            CC: clang
+            CMAKE_GENERATOR: Ninja
+            CMAKE_OPTIONS: -DUSE_HTTPS=OpenSSL -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON
+          os: ubuntu-latest
+        - # Xenial, Clang, mbedTLS
+          image: xenial
+          env:
+            CC: clang
+            CMAKE_OPTIONS: -DUSE_HTTPS=mbedTLS -DUSE_SHA1=HTTPS -DREGEX_BACKEND=pcre -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON
+            CMAKE_GENERATOR: Ninja
+          os: ubuntu-latest
+        - # macOS
+          os: macos-10.15
+          env:
+            CC: clang
+            CMAKE_OPTIONS: -DREGEX_BACKEND=regcomp_l -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=leaks -DUSE_GSSAPI=ON
+            CMAKE_GENERATOR: Ninja
+            PKG_CONFIG_PATH: /usr/local/opt/openssl/lib/pkgconfig
+            SKIP_SSH_TESTS: true
+            SKIP_NEGOTIATE_TESTS: true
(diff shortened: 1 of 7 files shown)


ethomson:
I suspect that this first build will fail since I'm trying to push the docker containers indiscriminately, and pull requests don't have permission to write to our package registry (for - hopefully - obvious reasons).  I got this working on my fork, but I'll have to consider a bit more how to cope with PRs.

ethomson:
Ah, no, of course it works, the permissions on the package registry allow me to write to it.  This is working as designed. :)  But I would like to have a nice failure case for people opening a PR from a fork, where there's no permission to write to the package registry.

.github/workflows/main.yml
```diff
@@ -0,0 +1,209 @@
...
+  docker-config-path: azure-pipelines/docker
+
+jobs:
+  # Build the docker container images that we will use for our Linux
+  # builds.  This will identify the last commit to the repository that
+  # updated the docker images, and try to download the image tagged with
+  # that sha.  If it does not exist, we'll do a docker build and push
+  # the image up to GitHub Packages for the actual CI/CD runs.  We tag
+  # with both the sha and "latest" so that the subsequent runs need not
+  # know the sha.
```
pks-t, line 22: That's awesome! Does this also work for PRs from forks that change Dockerfiles? Seems like it does based on below `github.repository` reference. :tada:

.github/workflows/main.yml
```diff
@@ -0,0 +1,209 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+    - name: Calculate image name
+      run: |
+        echo "::set-env name=docker-container::${{ github.repository }}/${{ matrix.container.name }}"
+        echo "::set-env name=docker-sha::$(git log -1 --pretty=format:"%h" -- ${{ env.docker-config-path }})"
```
pks-t, line 40: Is this to catch the case where e.g. the image changes and uses a different version of Docker or its config?

.github/workflows/main.yml
```diff
@@ -0,0 +1,209 @@
...
+    - name: Set up build environment
+      run: azure-pipelines/${{ matrix.platform.setup }}
+      shell: bash
+      if: matrix.platform.setup != ''
+    - name: Build and test
+      run: |
+        if [ -n "${{ matrix.platform.image }}" ]; then
+          docker_container="${{ github.repository }}/${{ matrix.platform.image }}:latest"
+          docker login https://${{ env.docker-registry }} -u ${{ github.actor }} -p ${{ github.token }}
+          docker pull ${{ env.docker-registry }}/${docker_container}
```
pks-t, line 170: Is this explicit pull required? I thought `docker run` can be directly passed the image we want to run.

.github/workflows/main.yml
ethomson: I pushed up a change so that for CI builds (merges into master) that we will fetch the container if it exists for the current hash of the Dockerfiles, and it will build a new one and push it up if it doesn't.  For PR builds, we will just always build a new docker container andr run it for the build.

But this is too far in the other direction - what we really want is to use the container from GitHub Packages for PRs that don't touch the Dockerfiles, and build their own when they do.  I'll fix this up next...

ethomson:
OK - now we try to fetch the docker container that was built for the last update to the Dockerfile.  (Done by looking at the git history for the given Dockerfile; using the abbreviated commit hash as the container's tag.)  If that is found, we'll use it for the build.  If it's not found _and_ we are running a CI build then we'll build the container and push it up to the package registry so that subsequent builds can use it.

But if we're running in a PR validation build and we can't find an existing container (because the PR is updating the Dockerfile) then we'll just build the container locally and run our build inside of that.  We don't try to push up to the package registry because we couldn't, do to the permissions (forks can't write to the upstream package registry).  Once merged, though, the CI build will do the proper caching.

.github/workflows/main.yml
```diff
@@ -0,0 +1,202 @@
...
+            CC: gcc
+            CMAKE_OPTIONS: -DUSE_HTTPS=OpenSSL -DREGEX_BACKEND=builtin -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON
+            CMAKE_GENERATOR: Ninja
+          os: ubuntu-latest
+        - # Xenial, GCC, mbedTLS
+          image: xenial
+          env:
+            CC: gcc
+            CMAKE_GENERATOR: Ninja
+            CMAKE_OPTIONS: -DUSE_HTTPS=mbedTLS -DUSE_SHA1=HTTPS -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON
```
pks-t, line 73: Nit: the order of options is different from all the other platforms here

.github/workflows/main.yml
```diff
@@ -0,0 +1,202 @@
...
+            CMAKE_OPTIONS: -DUSE_HTTPS=OpenSSL -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON
+            CMAKE_GENERATOR: Ninja
+          os: ubuntu-latest
+        - # Xenial, Clang, mbedTLS
+          image: xenial
+          env:
+            CC: clang
+            CMAKE_OPTIONS: -DUSE_HTTPS=mbedTLS -DUSE_SHA1=HTTPS -DREGEX_BACKEND=pcre -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON
+            CMAKE_GENERATOR: Ninja
+          os: ubuntu-18.04
```
pks-t, line 88: If this is the docker host, then we should be fine using "ubuntu-latest" here, too?

.github/workflows/main.yml
```diff
@@ -0,0 +1,202 @@
...
+            SKIP_SSH_TESTS: true
+            SKIP_NEGOTIATE_TESTS: true
+      fail-fast: false
+    env: ${{ matrix.platform.env }}
+    runs-on: ${{ matrix.platform.os }}
+    steps:
+    - name: Check out repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
pks-t, line 144: Do we actually care for the fetch depth? I seem to recall we had a test that dug into libgit2's own history, but not sure if we fixed that. If we did, then removing the fetch depth might speed up the clone by a few seconds?

.github/workflows/main.yml
```diff
@@ -0,0 +1,202 @@
...
+    - name: Set up build environment
+      run: azure-pipelines/setup-${{ matrix.platform.setup-script }}.sh
+      shell: bash
+      if: matrix.platform.setup-script != ''
+    - name: Download container
```
pks-t, line 149: `getcontainer.sh` may also build the container if it doesn't exist, right? In that case we should rename this step to make clear it's not only about downloading

.github/workflows/main.yml
ethomson: We do care - it's so that we can identify the last commit that touched the Dockerfiles.

.github/workflows/main.yml
ethomson: This is intentional - the UI shows the values in the matrix in the order they're specified, so this job shows up as Linux (gcc, Ninja, -DUSE_HTTPS=mbedTLS).  I wanted to front-load the important pieces of data.

.github/workflows/main.yml
ethomson: No, it only does a `pull` and sets the environment variables as appropriate.

.github/workflows/main.yml
ethomson: Indeed - thanks - fixed!

.github/workflows/main.yml
pks-t: @ethomson Sorry for not making myself clear, but I referred to the order `CC/CMAKE_GENERATOR/CMAKE_OPTIONS`.

.github/workflows/main.yml
pks-t: Oh, sure! I forgot we obviously need that info both during Docker build and libgit2 build.

.github/workflows/main.yml
pks-t: Ok, cool!

.github/workflows/main.yml
ethomson: I see - should be fixed.

ethomson:
Going to go ahead and merge this to unblock the x86 Windows builds.  I didn't remove Azure Pipelines but I think that might be the logical Next Step.

pks-t:
On Tue, Jun 16, 2020 at 08:22:13AM -0700, Edward Thomson wrote:
> Going to go ahead and merge this to unblock the x86 Windows builds.  I
> didn't remove Azure Pipelines but I think that might be the logical
> Next Step.

Before we do that we still need to convert our Coverity builds first.
Also, wouldn't we loose coverage on some platforms we're currently
testing in our nightlies?

PR #5551, opened 2020-06-08
Missing declarations

pks-t:
Over time, we have accumulated quite a lot of functions with missing
prototypes, missing `static` keywords or which were completely unused.
It's easy to miss these mistakes, but luckily GCC and Clang both have
the `-Wmissing-declarations` warning. Enabling this will cause them to
emit warnings for every not-static function that doesn't have a previous
declaration. This is a very sane thing to enable, and with the preceding
commits all these new warnings have been fixed.

So let's always enable this warning so we won't introduce new instances
of them.

---

Note that I've also renamed some now-static functions to match our coding style:

> All public functions must have a `git` prefix as well as a prefix indicating
their respective subsystem. E.g. a function that opens a repository should be
called `git_repository_open()`. Functions that are not public but declared in
an internal header file for use by other subsystems should follow the same
naming pattern. File-local static functions must not have a `git` prefix, but
should have a prefix indicating their respective subsystem.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -21,6 +21,8 @@ SET(LIBGIT2_INCLUDES
 SET(LIBGIT2_SYSTEM_INCLUDES "")
 SET(LIBGIT2_LIBS "")
 
+enable_warnings(missing-declarations)
+
 # Enable tracing
 IF(ENABLE_TRACE)
 	SET(GIT_TRACE 1)
--- a/src/blame.c
+++ b/src/blame.c
@@ -538,7 +538,9 @@ int git_blame_options_init(git_blame_options *opts, unsigned int version)
 	return 0;
 }
 
+#ifndef GIT_DEPRECATE_HARD
 int git_blame_init_options(git_blame_options *opts, unsigned int version)
 {
 	return git_blame_options_init(opts, version);
 }
+#endif
--- a/src/blob.c
+++ b/src/blob.c
@@ -445,6 +445,7 @@ int git_blob_filter(
 
 /* Deprecated functions */
 
+#ifndef GIT_DEPRECATE_HARD
 int git_blob_create_frombuffer(
 	git_oid *id, git_repository *repo, const void *buffer, size_t len)
 {
@@ -491,3 +492,4 @@ int git_blob_filtered_content(
 
 	return git_blob_filter(out, blob, path, &opts);
 }
+#endif
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -133,10 +133,12 @@ void git_buf_dispose(git_buf *buf)
 	git_buf_init(buf, 0);
 }
 
+#ifndef GIT_DEPRECATE_HARD
 void git_buf_free(git_buf *buf)
 {
 	git_buf_dispose(buf);
 }
+#endif
 
 void git_buf_sanitize(git_buf *buf)
 {
--- a/src/cache.c
+++ b/src/cache.c
@@ -41,27 +41,6 @@ int git_cache_set_max_object_size(git_object_t type, size_t size)
 	return 0;
 }
 
-void git_cache_dump_stats(git_cache *cache)
-{
-	git_cached_obj *object;
-
-	if (git_cache_size(cache) == 0)
-		return;
-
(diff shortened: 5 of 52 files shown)


pks-t:
Ah, I forgot about deprecated functions. I see multiple possible solutions, even though I'm not sure which one I prefer:

- Wrap definitions of deprecated functions into `#ifndef GIT_DEPRECATE_HARD` so they don't get compiled into the final library. This would break the ABI for any existing applications.
- Disable `-Wmissing-declarations` in case we compile with `-DDEPRECATE_HARD=ON`. Not much of a fan of this.
- Come up with some macro `GIT_DEPRECATED` that performs some magic so the compiler stops complaining. This'd also have the neat effect that every deprecated function is explicitly marked in the source code, which it currently isn't. This seems the best solution to me, but I ain't got an idea how to implement it.

/cc @ethmoson

pks-t:
On Mon, Jun 08, 2020 at 04:25:37AM -0700, Patrick Steinhardt wrote:
> Ah, I forgot about deprecated functions. I see multiple possible
> solutions, even though I'm not sure which one I prefer:
> 
> - Wrap definitions of deprecated functions into `#ifndef
> GIT_DEPRECATE_HARD` so they don't get compiled into the final library.
> This would break the ABI for any existing applications. - Disable
> `-Wmissing-declarations` in case we compile with
> `-DDEPRECATE_HARD=ON`. Not much of a fan of this. - Come up with some
> macro `GIT_DEPRECATED` that performs some magic so the compiler stops
> complaining. This'd also have the neat effect that every deprecated
> function is explicitly marked in the source code, which it currently
> isn't. This seems the best solution to me, but I ain't got an idea how
> to implement it.

Actually, I'm questioning whether deprecations currently work as
expected. In case you compile a library with `-DDEPRECATE_HARD=ON`,
there won't be any prototype visible in the compilation units for the
deprecated functions, as "deprecated.h" is guarded by a preprocessor
macro. Which effectively means that we also won't ever see a
`GIT_EXTERN` declaration for the functions prototype.

As a result, these functions will not have correct visibility in the
resulting library and you can't use deprecated functions in a library
compiled with hard deprecation on some platforms. So while do we even
compile them into the library in the first place? Which makes me think
we should just remove the deprecated functions altogether with
preprocessor macros in case we use hard deprecation or have code that
implements them include the "deprecated.h" header with `#define
GIT_DEPRECATE_HARD` to make prototypes visible.

ethomson:
> As a result, these functions will not have correct visibility in the
resulting library and you can't use deprecated functions in a library
compiled with hard deprecation on some platforms.

Right, that's correct and expected.  In fact, what our CMake description says for this is:

> Do not include deprecated functions in the library

Users are not expected to build the library with this enabled - users should build the library normally and set `GIT_DEPRECATE_HARD` before including `git2.h` if they want.  This is something that we added to the CMakeLists.txt just so that we could ensure that we never called anything deprecated within the library source itself or in tests.

I think it's therefore safe to wrap the function bodies themselves in `ifndef GIT_DEPRECATE_HARD`.

pks-t:
On Mon, Jun 08, 2020 at 10:55:40AM -0700, Edward Thomson wrote:
> > As a result, these functions will not have correct visibility in the
> resulting library and you can't use deprecated functions in a library
> compiled with hard deprecation on some platforms.
> 
> Right, that's correct and expected.  In fact, what our CMake description says for this is:
> 
> > Do not include deprecated functions in the library
> 
> Users are not expected to build the library with this enabled - users should build the library normally and set `GIT_DEPRECATE_HARD` before including `git2.h` if they want.  This is something that we added to the CMakeLists.txt just so that we could ensure that we never called anything deprecated within the library source itself or in tests.

Aha, that makes sense! I thought it wouldn't make sense for anybody who
is a distributor of libgit2 to enable -DDEPRECATE_HARD, but only for
people who directly embed libgit2 into their project. And for these,
not compiling deprecated functions is definitely safe and will even safe
them some bytes. So thanks for the clarification!

> I think it's therefore safe to wrap the function bodies themselves in `ifndef GIT_DEPRECATE_HARD`.

Okay, cool. Let me do the legwork :+1:

pks-t:
> Okay, cool. Let me do the legwork :+1:

Done :)

pks-t:
Fixed one issue on MinGW, so CI's green now. :tada:

PR #5566, opened 2020-06-26
Access symrefs via refdb

pks-t:
This is another set of commits that fixes a subset of our codebase that's accessing symrefs directly via the filesystem, circumventing the refdb. This is a layering violation and also breaks our own abstractions, making reference backends essentially unfunctional in some situations.

### DIFF
--- a/src/cherrypick.c
+++ b/src/cherrypick.c
@@ -7,11 +7,12 @@
 
 #include "common.h"
 
-#include "repository.h"
 #include "filebuf.h"
-#include "merge.h"
-#include "vector.h"
 #include "index.h"
+#include "merge.h"
+#include "refdb.h"
+#include "repository.h"
+#include "vector.h"
 
 #include "git2/types.h"
 #include "git2/merge.h"
@@ -99,9 +100,24 @@ static int cherrypick_normalize_opts(
 
 static int cherrypick_state_cleanup(git_repository *repo)
 {
-	const char *state_files[] = { GIT_CHERRYPICK_HEAD_FILE, GIT_MERGE_MSG_FILE };
+	const char *state_files[] = { GIT_MERGE_MSG_FILE };
+	git_refdb *refdb = NULL;
+	int error;
 
-	return git_repository__cleanup_files(repo, state_files, ARRAY_SIZE(state_files));
+	assert(repo);
+
+	if ((error = git_repository_refdb(&refdb, repo)) < 0 ||
+	    (error = git_refdb_delete(refdb, GIT_CHERRYPICK_HEAD_FILE, NULL, NULL)) < 0)
+		if (error != GIT_ENOTFOUND)
+			goto out;
+
+
+	if ((error = git_repository_cleanup_files(repo, state_files, ARRAY_SIZE(state_files))) < 0)
+		goto out;
+
+out:
+	git_refdb_free(refdb);
+	return error;
 }
 
 static int cherrypick_seterr(git_commit *commit, const char *fmt)
--- a/src/merge.c
+++ b/src/merge.c
@@ -13,10 +13,10 @@
 #include "revwalk.h"
 #include "commit_list.h"
 #include "path.h"
+#include "refdb.h"
 #include "refs.h"
 #include "object.h"
 #include "iterator.h"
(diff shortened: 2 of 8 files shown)


ethomson:
I took only a quick glance at this, but I'm pretty torn about this.  It seems like we're moving things like `MERGE_HEAD` into the refdb abstraction?  It's not clear that that's really the right thing to me.  Are those _really_ references?  My thinking is no, they may look like direct references, but are actually repository metadata.

In fact, `MERGE_HEAD` is _not_ a reference.  It's a newline separated list of _all_ the OIDs that are being merged.  In the case of an octopus merge, this is multiple lines of OIDs, which is not a legal ref.

```bash
$  git merge --no-commit asdf bsdf
Fast-forwarding to: asdf
Trying simple merge with bsdf
Automatic merge went well; stopped before committing as requested
$ cat .git/MERGE_HEAD
d0505b82383d9daaff46fa82454edff02ef4031e
935db3f28cc0dcb60fa25a95208ae7864cd972e5
```

Moreover, this sets up a really painful distinction between (say) `REVERT_HEAD` which will only have a single OID and could live in the refdb, and `MERGE_HEAD` or `FETCH_HEAD` which are not references and could not.  And it shards the data about in-progress operations into the refdb for things like cherry-pick, where the `CHERRYPICK_HEAD` is in one place (the repository directory) but the sequencer data is in another.

Putting all that aside for a moment: this feels like a place where we need to align with git.  Would git put the MERGE_HEAD in the reftable?  If so, we'll need to figure these problems out.  If not, I feel like we should not be innovating here.

pks-t:
On Tue, Jul 14, 2020 at 01:28:29AM -0700, Edward Thomson wrote:
> I took only a quick glance at this, but I'm pretty torn about this.  It seems like we're moving things like `MERGE_HEAD` into the refdb abstraction?  It's not clear that that's really the right thing to me.  Are those _really_ references?  My thinking is no, they may look like direct references, but are actually repository metadata.
> 
> In fact, `MERGE_HEAD` is _not_ a reference.  It's a newline separated list of _all_ the OIDs that are being merged.  In the case of an octopus merge, this is multiple lines of OIDs, which is not a legal ref.
> 
> ```bash
> $  git merge --no-commit asdf bsdf
> Fast-forwarding to: asdf
> Trying simple merge with bsdf
> Automatic merge went well; stopped before committing as requested
> $ cat .git/MERGE_HEAD
> d0505b82383d9daaff46fa82454edff02ef4031e
> 935db3f28cc0dcb60fa25a95208ae7864cd972e5
> ```
> 
> Moreover, this sets up a really painful distinction between (say) `REVERT_HEAD` which will only have a single OID and could live in the refdb, and `MERGE_HEAD` or `FETCH_HEAD` which are not references and could not.  And it shards the data about in-progress operations into the refdb for things like cherry-pick, where the `CHERRYPICK_HEAD` is in one place (the repository directory) but the sequencer data is in another.
> 
> Putting all that aside for a moment: this feels like a place where we need to align with git.  Would git put the MERGE_HEAD in the reftable?  If so, we'll need to figure these problems out.  If not, I feel like we should not be innovating here.

Yup, it _is_ very complicated and we start treating some files more
special than others, but I'm following what upstream does (except for
MERGE_HEAD, which you're right about).

See https://public-inbox.org/git/pull.539.v19.git.1593457018.gitgitgadget@gmail.com/,
more specifically patches 6-8.

pks-t:
By the way, this once again highlights why the reftable backend is such
a good thing. Using the filesystem for storing stuff is just such an
awkward way to go about this, and those
refs-but-not-really-except-for-most-of-the-time are just plain awful.
How is one going to tell whether a file is a ref or a pseudo-ref?
Maybe it just looks like one but in fact isn't like MERGE_HEAD? Having
called that one `*_HEAD` adds insult to injury, as it looks even more
like a normal symbolic reference like that.

I really look forward to having a sane reference backend one day in core
git :)

ethomson:
> See https://public-inbox.org/git/pull.539.v19.git.1593457018.gitgitgadget@gmail.com/,
more specifically patches 6-8.

OK - thanks for the link - I'll bring this discussion there.  :D

PR #5567, opened 2020-06-28
Make the tests pass cleanly with MemorySanitizer

lhchavez:
This change:

* Initializes a few variables that were being read before being
  initialized.
* Includes https://github.com/madler/zlib/pull/393. As such,
  it only works reliably with `-DUSE_BUNDLED_ZLIB=ON`.


deps/ntlmclient/ntlm.c
```diff
@@ -47,11 +47,9 @@ ntlm_client *ntlm_client_init(ntlm_client_flags flags)
 {
 	ntlm_client *ntlm = NULL;
 
-	if ((ntlm = malloc(sizeof(ntlm_client))) == NULL)
+	if ((ntlm = calloc(1, sizeof(ntlm_client))) == NULL)
```
pks-t, line 50: It would probably make sense to split out all changes to ntlmclient into a separate commit so that @ethomson can easily cherry-pick them to upstream.

src/object.c
```diff
@@ -86,7 +86,9 @@ int git_object__from_raw(
 	GIT_ERROR_CHECK_ALLOC(object);
 	object->cached.flags = GIT_CACHE_STORE_PARSED;
 	object->cached.type = type;
-	git_odb_hash(&object->cached.oid, data, size, type);
+	if ((error = git_odb_hash(&object->cached.oid, data, size, type)) < 0) {
+		return error;
+	}
```
pks-t, line 91: No need for braces around single-line bodies :)

src/odb_loose.c
```diff
@@ -273,7 +273,7 @@ static int read_loose_packlike(git_rawobj *out, git_buf *obj)
 static int read_loose_standard(git_rawobj *out, git_buf *obj)
 {
 	git_zstream zstream = GIT_ZSTREAM_INIT;
-	unsigned char head[MAX_HEADER_LEN], *body = NULL;
+	unsigned char head[MAX_HEADER_LEN] = {0}, *body = NULL;
```
pks-t, line 276: Does zlib access these bytes before writing into the provided buffers? Or is it that zlib may write short, leaving the remainder uninitialized?

deps/ntlmclient/ntlm.c
lhchavez: Now live at #5571

src/object.c
lhchavez: Argh, amending code is hard.

src/odb_loose.c
lhchavez: Ah this was a workaround that I did before cherry-picking https://github.com/madler/zlib/pull/393, but it's not needed now that the root cause has been fixed. Removed.

ethomson:
Thanks @lhchavez!

Neustradamus:
@lhchavez: Please look upstream.

PR #5569, opened 2020-06-28
Add CI support for Memory and UndefinedBehavior Sanitizers

lhchavez:
This change adds two new build targets: MSan and UBSan. This is because
even though OSS-Fuzz is great and adds a lot of coverage, it only does
that for the fuzz targets, so the rest of the codebase is not
necessarily run with the Sanitizers ever :( So this change makes sure
that MSan/UBSan warnings don't make it into the codebase.

As part of this change, the Ubuntu focal container is introduced. It
builds mbedTLS and libssh2 as debug libraries into /usr/local and as
MSan-enabled libraries into /usr/local/msan. This latter part is needed
because MSan requires the binary and all its dependent libraries to be
built with MSan support so that memory allocations and deallocations are
tracked correctly to avoid false positives.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,6 +28,7 @@ jobs:
         container:
         - xenial
         - bionic
+        - focal
         - docurium
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


lhchavez:
This change depends on #5568 and #5567. They are embedded in this PR for testing the whole thing end-to-end and ease of reviewing.

ethomson:
Thanks for tackling this.

I was a little nervous about bringing this in to the main CI (thinking that maybe it should be kept nightly).  The two new builds only add 6 and 9 minutes, so that's still faster than our Windows builds.  _But_ I'm a bit worried about parallelism, we'll eventually start starving ourselves of available runners in the CI system.  But let's merge this here and 🤞 - we can fix this in the future if we need to.

How bad was it to add a new platform and docker image to this system?  It's grown up a bit and has moving parts (build the image, put the image in the container registry, then use that image in subsequent steps, _or maybe not_)?

lhchavez:
> Thanks for tackling this.
> 
> I was a little nervous about bringing this in to the main CI (thinking that maybe it should be kept nightly). The two new builds only add 6 and 9 minutes, so that's still faster than our Windows builds. _But_ I'm a bit worried about parallelism, we'll eventually start starving ourselves of available runners in the CI system. But let's merge this here and  - we can fix this in the future if we need to.
> 
> How bad was it to add a new platform and docker image to this system? It's grown up a bit and has moving parts (build the image, put the image in the container registry, then use that image in subsequent steps, _or maybe not_)?

not _too_ bad. copy+paste got me 90% of the way, the rest was just making tweaks here and there to make everything line up.

this was done in the middle of the effort to make the rest of the codebase Sanitizer-safe, so there was some instant gratification which made it perceptually better.

PR #5570, opened 2020-06-30
refdb: a set of preliminary refactorings for the reftable backend

pks-t:
I've decided to pull out some of the preliminary work for the reftable backend in order to keep #5462 focussed on the already-intense refdb logic. This PR is what I ended up with. I did several refactorings to pull out functions that are commonly needed for refdb backend implementations and moved the resolving logic from the refs code into the refdb code, mostly in order to deduplicate the three different symref resolvers we had. This also fixes a bug where we could spin indefinitely resolving symrefs in case there's a symref loop in our repo.

### DIFF
--- a/src/refdb.c
+++ b/src/refdb.c
@@ -326,7 +326,7 @@ int git_refdb_should_write_reflog(int *out, git_refdb *db, const git_reference *
 
 int git_refdb_should_write_head_reflog(int *out, git_refdb *db, const git_reference *ref)
 {
-	git_reference *head = NULL, *peeled = NULL;
+	git_reference *head = NULL, *resolved = NULL;
 	const char *name;
 	int error;
 
@@ -344,22 +344,15 @@ int git_refdb_should_write_head_reflog(int *out, git_refdb *db, const git_refere
 		goto out;
 
 	/* Go down the symref chain until we find the branch */
-	while (git_reference_type(head) == GIT_REFERENCE_SYMBOLIC) {
-		if ((error = git_refdb_lookup(&peeled, db, git_reference_symbolic_target(head))) < 0)
-			break;
-
-		git_reference_free(head);
-		head = peeled;
-		peeled = NULL;
-	}
-
-	if (error < 0) {
+	if ((error = git_refdb_resolve(&resolved, db, git_reference_symbolic_target(head), -1)) < 0) {
 		if (error != GIT_ENOTFOUND)
 			goto out;
 		error = 0;
 		name = git_reference_symbolic_target(head);
+	} else if (git_reference_type(resolved) == GIT_REFERENCE_SYMBOLIC) {
+		name = git_reference_symbolic_target(resolved);
 	} else {
-		name = git_reference_name(head);
+		name = git_reference_name(resolved);
 	}
 
 	if (strcmp(name, ref->name))
@@ -368,7 +361,7 @@ int git_refdb_should_write_head_reflog(int *out, git_refdb *db, const git_refere
 	*out = 1;
 
 out:
-	git_reference_free(peeled);
+	git_reference_free(resolved);
 	git_reference_free(head);
 	return error;
 }
--- a/src/refdb.c
+++ b/src/refdb.c
@@ -17,6 +17,9 @@
 #include "reflog.h"
 #include "posix.h"
 
+#define DEFAULT_NESTING_LEVEL	5
+#define MAX_NESTING_LEVEL		10
+
 int git_refdb_new(git_refdb **out, git_repository *repo)
 {
 	git_refdb *db;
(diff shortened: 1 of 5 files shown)


src/refdb.c
```diff
@@ -231,6 +286,83 @@ int git_refdb_reflog_read(git_reflog **out, git_refdb *db,  const char *name)
...
+
+	case GIT_LOGALLREFUPDATES_ALWAYS:
+		*out = 1;
+		break;
+	}
+
+	return 0;
+}
+
+int git_refdb_should_write_head_reflog(int *out, git_refdb *db, const git_reference *ref)
```
pks-t, line 326: Note: I've decided to keep both functions private for now so I can iterate in case we need to change the signature. In the ideal case, the refdb would already handle this transparently to the backend. But given that the `write` callback is supposed to handle reflog updates itself, this burden got pushed onto the implementer of the backend. So at some point, we may want to expose both functions to help third-party refdb backend implementations be consistent, but I'm not sure yet whether we've got sufficient information available here.

ethomson: Agreed - this makes sense for now.

src/refs.c
```diff
@@ -214,52 +214,17 @@ int git_reference_lookup_resolved(
...
+	int error = 0;
 
 	assert(ref_out && repo && name);
 
-	*ref_out = NULL;
-
-	if (max_nesting > MAX_NESTING_LEVEL)
-		max_nesting = MAX_NESTING_LEVEL;
-	else if (max_nesting < 0)
-		max_nesting = DEFAULT_NESTING_LEVEL;
```
ethomson, line 230: It looks like `DEFAULT_NESTING_LEVEL` is now unused and can get cleaned up.

`MAX_NESTING_LEVEL`, on the other hand, remains used in `get_terminal`.  🤔  Seems disappointing to have it defined twice.  Can we make it (or maybe both?) a global but internal-to-the-library constant?

ethomson: Oops, I should start doing proper reviews where the comments are batched until the end, because this question was obviously answered for me in the next commit.  🙃

src/refdb.h
```diff
@@ -30,6 +30,27 @@ int git_refdb_lookup(
 	git_refdb *refdb,
 	const char *ref_name);
 
+/**
+ * Resolve the reference by following symbolic references.
+ *
+ * Given a reference name, this function will follow any symbolic references up
+ * to `max_nesting` deep and return the resovled direct reference. If any of
```
ethomson, line 37: typo: `resolved`.

src/refs.c
```diff
@@ -221,10 +221,17 @@ int git_reference_lookup_resolved(
 	assert(ref_out && repo && name);
 
 	if ((error = reference_normalize_for_repo(normalized, repo, name, true)) < 0 ||
-	    (error = git_repository_refdb__weakptr(&refdb, repo)) < 0 ||
-	    (error = git_refdb_resolve(ref_out, refdb, normalized, max_nesting)) < 0)
+	    (error = git_repository_refdb__weakptr(&refdb, repo)) < 0)
 		return error;
 
+	if ((error = git_refdb_resolve(ref_out, refdb, normalized, max_nesting)) < 0) {
```
ethomson, line 224: Do we ever use this pattern elsewhere where we both return an error _and_ return meaningful data (especially data that needs to be freed?).  This feels super ambiguous to me.

I think that we should return non-negative in this case - even `0` vs `1` would be an improvement, IMHO.

A `bool *exists` in the signature could be helpful, but it doesn't really feel very ergonomic.

I don't see the removal of the ad-hoc function that you mentioned, maybe we should bundle this commit with its removal so that it's clearer how/why this is changing?

ethomson: Sorry, I was reviewing out of order.  Now I understand better.

My complaint holds - this pattern feels really unwieldy to me ...  are there ways that you think we might be able to improve it?

ethomson:
Thanks, as usual, for the helpful to review series.  Only a few comments, otherwise this is great.

src/refdb.c
```diff
@@ -134,6 +137,58 @@ int git_refdb_lookup(git_reference **out, git_refdb *db, const char *ref_name)
...
+		goto out;
+
+	for (nesting = 0; nesting < max_nesting; nesting++) {
+		git_reference *resolved;
+
+		if (ref->type == GIT_REFERENCE_DIRECT)
+			break;
+
+		if ((error = git_refdb_lookup(&resolved, db, git_reference_symbolic_target(ref))) < 0) {
+			/* If we found a symoblic reference with a nonexistent target, return it. */
```
ethomson, line 166: typo: symbolic

ethomson:
Nit: 96fb8039acce8ecc927294bdaef47bf949558555 has a typo in the commit message (`git_refdb_resovle` -> `git_refdb_resolve`)

72a89c70b02f0b605c6036cb3e98fda1dfb6d68c has a typo in the commit message (`symoblic` -> `symbolic`)

src/refs.c
pks-t: > My complaint holds - this pattern feels really unwieldy to me ... are there ways that you think we might be able to improve it?

@ethomson Yeah, I totally get your point and expected such a comment to come up. Mostly because I hate the interface myself and find it really hard to use correctly, too. How about we just document it may also return the last symbolic reference pointing to a nonexistent branch and pushing the burden onto the caller to check whether the returned reference is a symbolic reference or a direct reference?

pks-t: Done. Honestly, the resulting code does look a lot cleaner on call-sites now. Thanks for raising this point!

pks-t:
> Nit: [96fb803](https://github.com/libgit2/libgit2/commit/96fb8039acce8ecc927294bdaef47bf949558555) has a typo in the commit message (`git_refdb_resovle` -> `git_refdb_resolve`)
> 
> [72a89c7](https://github.com/libgit2/libgit2/commit/72a89c70b02f0b605c6036cb3e98fda1dfb6d68c) has a typo in the commit message (`symoblic` -> `symbolic`)

Fixed both :+1:

pks-t:
Thanks, @ethomson!

PR #5571, opened 2020-06-30
Make NTLMClient Memory and UndefinedBehavior Sanitizer-clean

lhchavez:
This change makes the code pass the libgit2 tests cleanly when
MSan/UBSan are enabled. Notably:

* Changes malloc/memset combos into calloc for easier auditing.
* Makes `write_buf` return early if the buffer length is empty to avoid
  arithmetic with NULL pointers (which UBSan does not like).
* Initializes a few arrays that were sometimes being read before being
  written to.


ethomson:
Thanks!

PR #5574, opened 2020-07-08
p_chmod: Android compatibility

### LINKED ISSUE #5565: chmod failing shouldn't be treated as error when happens on a fat* partition
### Reproduction steps
Run any clone operation targeting a folder in fat* partition in a posix environment
### Expected behavior
The operation is completed successfully
### Actual behavior
The operation falis with the error ``failed to set permissions on file 'file': Operation not permitted``
libgit2 1.0.0 built with android ndk21 targeting android 16 via vcpkg
Android 5.0 lollipop
Android 4.4 kitkat
### Additional information
This issue is caused by the fact that the fat filesystem doesn't support unix-style file permissions, so it will return error  every time the function is called

kevinlul:
Fix #5565

Pre-Android 5 did not implement a virtual filesystem atop FAT partitions for Unix permissions, which causes chmod to fail. However, Unix permissions have no effect on Android anyway as file permissions are not actually managed this way, so treating it as a no-op across all Android is safe.

### DIFF
--- a/src/unix/posix.h
+++ b/src/unix/posix.h
@@ -62,7 +62,15 @@ GIT_INLINE(int) p_fsync(int fd)
 #define p_snprintf snprintf
 #define p_mkstemp(p) mkstemp(p)
 #define p_chdir(p) chdir(p)
+#ifndef __ANDROID__
 #define p_chmod(p,m) chmod(p, m)
+#else
+/* Pre-Android 5 did not implement a virtual filesystem atop FAT partitions for
+   Unix permissions, which causes chmod to fail. However, Unix permissions have
+   no effect on Android anyway as file permissions are not actually managed this
+   way, so treating it as a no-op across all Android is safe. */
+#define p_chmod(p,m) 0
+#endif
 #define p_rmdir(p) rmdir(p)
 #define p_access(p,m) access(p,m)
 #define p_ftruncate(fd, sz) ftruncate(fd, sz)


src/unix/posix.h
```diff
@@ -62,7 +62,15 @@ GIT_INLINE(int) p_fsync(int fd)
 #define p_snprintf snprintf
 #define p_mkstemp(p) mkstemp(p)
 #define p_chdir(p) chdir(p)
+#ifndef __ANDROID__
 #define p_chmod(p,m) chmod(p, m)
+#else
+/* Pre-Android 5 did not implement a virtual filesystem atop FAT partitions for
```
lhchavez, line 68: Would it make sense to also gate this depending on the value of `__ANDROID_API__`? That way on the Android versions where this does matter uses the real `chmod(2)`?

kevinlul: The problem occurs at runtime so I guess it could work to not use the no-op if a sufficiently new Android platform is the build target. Doing `statfs` on every `chmod` call seems excessive when the `chmod` doesn't ultimately have an effect anyway.

kevinlul:
What steps are needed to move forward on this?

ethomson:
Sorry for the delay @kevinlul - I made a few formatting changes for consistency and merged this manually.  I appreciate you fixing this!

ethomson:
Fixed with e8ab3db99e13bcb330e91e0ca0265b60483a5e47

PR #5578, opened 2020-07-11
Friendlier getting started in the lack of git_libgit2_init

ethomson:
We require the library to be initialized with git_libgit2_init before it is functional.  However, if a user tries to uses the library without doing so - as they might when getting started with the library for the first time - we will likely crash.

This PR attempts to set up some guard rails to help a user out in this case.

Now, instead of having _no_ allocator by default, we'll have an allocator that always fails, and never tries to set an error message (since the thread-local state is set up by git_libgit2_init).  We've modified the error retrieval function to (try to) ensure that the library has been initialized before getting the thread-local error message.

(Unfortunately, we cannot determine if the thread local storage has actually been configured, this does require initialization by git_libgit2_init.  But a naive attempt should be good enough for most cases.)

Of course, this pre-supposes that the first program a user builds will actually check the error message.  I'm not certain that this is generally true.  I actually often write a quick little program and forget to call git_libgit2_init (but do error checking).  But I don't think that I'm a very typical user of the library.

It might be valuable to `fprintf(stderr, "hey! call git_libgit2_init!\n")`...?  But I thought that I'd start with this PR here and iterate.

Depends on #5546

### DIFF
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -8,10 +8,22 @@
 #include "alloc.h"
 #include "runtime.h"
 
+#include "allocators/failalloc.h"
 #include "allocators/stdalloc.h"
 #include "allocators/win32_leakcheck.h"
 
-git_allocator git__allocator;
+/* Fail any allocation until git_libgit2_init is called. */
+git_allocator git__allocator = {
+	git_failalloc_malloc,
+	git_failalloc_calloc,
+	git_failalloc_strdup,
+	git_failalloc_strndup,
+	git_failalloc_substrdup,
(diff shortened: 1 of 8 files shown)


ethomson:
Given:

```
#include <stdio.h>
#include <git2.h>

int main(int argc, char **argv)
{
	git_repository *foo;

	printf("%d\n", git_repository_open(&foo, "."));
	printf("%s\n", git_error_last()->message);
}
```

Before: (crash)

After:

```
-1
libgit2 has not been initialized; you must call git_libgit2_init
```

PR #5580, opened 2020-07-12
msvc crtdbg -> win32 leakcheck

ethomson:
"msvc crtdbg" is a mouthfull that is not particularly indicative of what it does.  Let's rename it to "win32 leakcheck".


ethomson:
Depends on #5546

PR #5563, opened 2020-06-17
Access HEAD via the refdb backends

pks-t:
There's quite a few places where we're accessing HEAD directly via the filesystem, which is completely broken in case repos use a different refdb backend than the filesystem-based one. This pull request addresses a subset of these which is updating HEAD in the context of worktrees.

PR #5581, opened 2020-07-13
Respect `init.defaultBranch` setting

ethomson:
The git project is moving towards [more inclusive language](https://github.com/git/git/commit/11cbda2add5d3eb7c415f1f6dd8186181a7f9874).  In particular, they're making the name of the "default" branch (currently "master") configurable with the `init.defaultBranch` setting.

We should follow suit for compatibility.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -70,7 +70,7 @@ static int check_extensions(git_config *config, int version);
 
 #define GIT_FILE_CONTENT_PREFIX "gitdir:"
 
-#define GIT_BRANCH_MASTER "master"
+#define GIT_BRANCH_DEFAULT "master"
 
 #define GIT_REPO_VERSION 0
 #define GIT_REPO_MAX_VERSION 1
@@ -1408,9 +1408,6 @@ int git_repository_create_head(const char *git_dir, const char *ref_name)
 	    (error = git_filebuf_open(&ref, ref_path.ptr, 0, GIT_REFS_FILE_MODE)) < 0)
 		goto out;
 
-	if (!ref_name)
-		ref_name = GIT_BRANCH_MASTER;
-
 	if (git__prefixcmp(ref_name, GIT_REFS_DIR) == 0)
 		fmt = "ref: %s\n";
 	else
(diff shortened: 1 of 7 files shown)


src/repository.c
```diff
@@ -2061,6 +2058,43 @@ static int repo_init_directories(
...
+	int error;
+
+	if ((error = git_buf_joinpath(&head_path, repo_dir, GIT_HEAD_FILE)) < 0)
+		goto out;
+
+	/*
+	 * A template may have set a HEAD; use that unless it's been
+	 * overridden by the caller's given initial head setting.
+	 */
+	if (git_path_exists(head_path.ptr) && !given)
```
pks-t, line 2075: This kind of crosses my current effort to stop using direct file access whenever writing symbolic references. We really shouldn't be doing that at all and always use the refdb. I realize we're not quite there yet when initializing new repositories as there may not yet be a refdb, but we could maybe do this as a preliminary refactoring.

ethomson: I just moved this, it's not net new.  I would expect the conflict would be reasonably easy to resolve?  It looks like #5563 is ready to merge so, in fact, it seems like _my_ conflict to resolve?

pks-t:
On Mon, Jul 13, 2020 at 01:33:02PM -0700, Edward Thomson wrote:
> 
> 
> 
> > @@ -2061,6 +2058,43 @@ static int repo_init_directories(
>  	return error;
>  }
>  
> +static int repo_init_head(const char *repo_dir, const char *given)
> +{
> +	git_config *cfg = NULL;
> +	git_buf head_path = GIT_BUF_INIT, cfg_branch = GIT_BUF_INIT;
> +	const char *initial_head = NULL;
> +	int error;
> +
> +	if ((error = git_buf_joinpath(&head_path, repo_dir, GIT_HEAD_FILE)) < 0)
> +		goto out;
> +
> +	/*
> +	 * A template may have set a HEAD; use that unless it's been
> +	 * overridden by the caller's given initial head setting.
> +	 */
> +	if (git_path_exists(head_path.ptr) && !given)
> 
> I just moved this, it's not net new.  I would expect the conflict
> would be reasonably easy to resolve?  It looks like #5563 is ready to
> merge so, in fact, it seems like _my_ conflict to resolve?

Uh. Yes. Didn't remember I actually already did the conversion. Feel
free to ignore in this case :)

PR #5582, opened 2020-07-13
config_entries: Avoid excessive map operations

pks-t:
When appending config entries, we currently always first get the
currently existing map entry and then afterwards update the map to
contain the current config value. In the common scenario where keys
aren't being overridden, this is the best we can do. But in case a key
gets set multiple times, then we'll also perform these two map
operations. In extreme cases, hashing the map keys will thus start to
dominate performance.

Let's optimize the pattern by using a separately allocated map entry.
Currently, we always put the current list entry into the map and update
it to get any overridden multivar. As these list entries are also used
to iterate config entries, we cannot update them in-place in the map and
are thus forced to always set the map to contain the new entry. But with
a separately allocated map entry, we can now create one once per config
key and insert it into the map. Whenever appending a new config value
with the same key, we can now just update the map entry in-place instead
of having to replace the map entry completely.

This reduces calls to the hashing function by half and trades the
improved runtime for one more allocation per unique config key. Given
that the refactoring arguably improves code readability by splitting
concerns of the `config_entry_list` type and not having to track it in
two different structures, this alone would already be reason enough to
take the trade.

Given a pathological case of a gitconfig with 100.000 repeated keys and
a section of length 10.000 characters, this reduces runtime by half from
approximately 14 seconds to 7 seconds as expected.

### DIFF
--- a/src/config_entries.c
+++ b/src/config_entries.c
@@ -11,8 +11,12 @@ typedef struct config_entry_list {
 	struct config_entry_list *next;
 	struct config_entry_list *last;
 	git_config_entry *entry;
-	bool first;
 } config_entry_list;
+
+typedef struct {
+	git_config_entry *entry;
+	bool multivar;
+} config_entry_map_head;
 
 typedef struct config_entries_iterator {
 	git_config_iterator parent;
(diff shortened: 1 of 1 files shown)


pks-t:
Triggered by a reported timeout when parsing config files by oss-fuzz. The case really is quite pathological, but it was fun to think about this issue. And honestly, resulting code feels cleaner to me and is easier to understand.

ethomson:
Nice, thanks @pks-t!

src/config_entries.c
```diff
@@ -127,40 +133,42 @@ void git_config_entries_free(git_config_entries *entries)
...
+		 * with a lot of same keys. As for multivars the entry's
+		 * key will be the same for all entries, we can just free
+		 * all except the first entry's name and just re-use it.
+		 */
 		git__free((char *) entry->name);
-		entry->name = existing->entry->name;
+		entry->name = map_head->entry->name;
 	} else {
-		head->first = 1;
+		map_head = git__calloc(1, sizeof(*map_head));
```
siprbaum, line 150: Is this safe when the allocation fails?
E.g. I see a check to GIT_ERROR_CHECK_ALLOC a few lines below after the other git_calloc call,
but not here.

Issue #4019, opened 2016-12-02
Degrade gracefully if mmap is not allowed on this file system


Keno:
libgit2 returns an error when it cannot mmap a file. Certain file systems, e.g. network file systems, disallow mmap (or at least MAP_SHARED). It would be nice if libgit2 degraded gracefully.

### Reproduction steps
Try to use a libgit2-based project on a network file system. In my case, this was DataWarp/GPFS, but I suspect the same is true for others as well. Example strace log:
```
open("/global/u1/k/kfischer/.julia/v0.6/tmpD5GGuN/METADATA/.git/objects/pack/pack_git2_t122RZ", O_RDWR|O_CREAT|O_EXCL, 0600) = 22
[snip]
mmap(NULL, 11, PROT_READ|PROT_WRITE, MAP_SHARED, 22, 0) = -1 ENOSYS (Function not implemented)
```

### Expected behavior
If the mmap fails, fall back to some other method (e.g. flock with explicit reads/writes).

### Actual behavior
Surfaces the error to the application.

### Version of libgit2 (release number or SHA1)
4cf1ec7cff28da8838a2f0a9fb330e312ea3f963

### Operating system(s) tested
Cray Linux

amm1994:
2

PR #5583, opened 2020-07-15
Build with NO_MMAP

0xdky:
This patch fixes building libgit2 without mmap (NO_MMAP)

Since 'mmap' does not update the position in file descriptor. Hence, we have to capture and restore the position of file descriptor in 'mmap' emulation.

### DIFF
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -603,6 +603,23 @@ static void hash_partially(git_indexer *idx, const uint8_t *data, size_t size)
 
 static int write_at(git_indexer *idx, const void *data, off64_t offset, size_t size)
 {
+#ifdef NO_MMAP
+	size_t remaining_size = size;
+	const char *ptr = (const char *)data;
+
+	/* Handle data size larger that ssize_t */
+	while (remaining_size > 0) {
+		ssize_t nb;
+		HANDLE_EINTR(nb, p_pwrite(idx->pack->mwf.fd, (void *)ptr,
+					  remaining_size, offset));
+		if (nb <= 0)
+			return -1;
+
+		ptr += nb;
+		offset += nb;
+		remaining_size -= nb;
+	}
+#else
 	git_file fd = idx->pack->mwf.fd;
 	size_t mmap_alignment;
 	size_t page_offset;
@@ -627,6 +644,7 @@ static int write_at(git_indexer *idx, const void *data, off64_t offset, size_t s
 	map_data = (unsigned char *)map.data;
 	memcpy(map_data + page_offset, data, size);
 	p_munmap(&map);
+#endif
 
 	return 0;
 }
--- a/src/posix.c
+++ b/src/posix.c
@@ -240,14 +240,15 @@ int p_mmap(git_map *out, size_t len, int prot, int flags, int fd, off64_t offset
 {
 	GIT_MMAP_VALIDATE(out, len, prot, flags);
 
-	out->data = NULL;
-	out->len = 0;
-
-	if ((prot & GIT_PROT_WRITE) && ((flags & GIT_MAP_TYPE) == GIT_MAP_SHARED)) {
-		git_error_set(GIT_ERROR_OS, "trying to map shared-writeable");
+	/* writes cannot be emulated without handling pagefaults since write happens by
+	 * writing to mapped memory */
+	if (prot & GIT_PROT_WRITE) {
+		git_error_set(GIT_ERROR_OS, "trying to map %s-writeable",
+				((flags & GIT_MAP_TYPE) == GIT_MAP_SHARED) ? "shared": "private");
 		return -1;
 	}
 
+	out->len = 0;
 	out->data = git__malloc(len);
 	GIT_ERROR_CHECK_ALLOC(out->data);
 
--- a/src/posix.c
+++ b/src/posix.c
@@ -238,6 +238,9 @@ int git__mmap_alignment(size_t *alignment)
 
 int p_mmap(git_map *out, size_t len, int prot, int flags, int fd, off64_t offset)
 {
+	const char *ptr;
+	size_t remaining_len;
+
 	GIT_MMAP_VALIDATE(out, len, prot, flags);
 
 	/* writes cannot be emulated without handling pagefaults since write happens by
@@ -248,15 +251,30 @@ int p_mmap(git_map *out, size_t len, int prot, int flags, int fd, off64_t offset
 		return -1;
 	}
 
+	if (!git__is_ssizet(len)) {
+		errno = EINVAL;
+		return -1;
+	}
+
 	out->len = 0;
 	out->data = git__malloc(len);
 	GIT_ERROR_CHECK_ALLOC(out->data);
 
-	if (!git__is_ssizet(len) ||
-		(p_lseek(fd, offset, SEEK_SET) < 0) ||
-		(p_read(fd, out->data, len) != (ssize_t)len)) {
-		git_error_set(GIT_ERROR_OS, "mmap emulation failed");
-		return -1;
+	remaining_len = len;
+	ptr = (const char *)out->data;
+	while (remaining_len > 0) {
+		ssize_t nb;
+		HANDLE_EINTR(nb, p_pread(fd, (void *)ptr, remaining_len, offset));
+		if (nb <= 0) {
+			git_error_set(GIT_ERROR_OS, "mmap emulation failed");
+			git__free(out->data);
+			out->data = NULL;
+			return -1;
+		}
+
+		ptr += nb;
+		offset += nb;
+		remaining_len -= nb;
 	}
 
 	out->len = len;
@@ -267,6 +285,10 @@ int p_munmap(git_map *map)
 {
 	GIT_ASSERT_ARG(map);
 	git__free(map->data);
+
+	/* Initializing will help debug use-after-free */
+	map->len = 0;
+	map->data = NULL;
 
 	return 0;
 }
--- a/src/posix.h
+++ b/src/posix.h
@@ -89,6 +89,12 @@
 #define EAFNOSUPPORT (INT_MAX-1)
 #endif
 
+/* Compiler independent macro to handle signal interrpted system calls */
+#define HANDLE_EINTR(result, x) do {					\
+		result = (x);						\
+	} while (result == -1 && errno == EINTR);
+
+
 /* Provide a 64-bit size for offsets. */
 
 #if defined(_MSC_VER)
@@ -118,6 +124,9 @@ typedef int git_file;
 
 extern ssize_t p_read(git_file fd, void *buf, size_t cnt);
 extern int p_write(git_file fd, const void *buf, size_t cnt);
+
+extern ssize_t p_pread(int fd, void *data, size_t size, off64_t offset);
+extern ssize_t p_pwrite(int fd, const void *data, size_t size, off64_t offset);
 
 #define p_close(fd) close(fd)
 #define p_umask(m) umask(m)
--- a/src/win32/posix_w32.c
+++ b/src/win32/posix_w32.c
@@ -981,3 +981,73 @@ int p_inet_pton(int af, const char *src, void *dst)
 	errno = EINVAL;
 	return -1;
 }
+
+ssize_t p_pread(int fd, void *data, size_t size, off64_t offset)
+{
+	HANDLE fh;
+	DWORD rsize = 0;
+	OVERLAPPED ov = {0};
+	LARGE_INTEGER pos = {0};
+	off64_t final_offset = 0;
+
+	/* Fail if the final offset would have overflowed to match POSIX semantics. */
+	if (!git__is_ssizet(size) || git__add_int64_overflow(&final_offset, offset, (int64_t)size)) {
+		errno = EINVAL;
+		return -1;	
+	}
+
+	/*
+	 * Truncate large writes to the maximum allowable size: the caller
+	 * needs to always call this in a loop anyways.
+	 */
+	if (size > INT32_MAX) {
+		size = INT32_MAX;
+	}
+
+	pos.QuadPart = offset;
+	ov.Offset = pos.LowPart;
+	ov.OffsetHigh = pos.HighPart;
+	fh = (HANDLE)_get_osfhandle(fd);
+
+	if (ReadFile(fh, data, (DWORD)size, &rsize, &ov)) {
+		return (ssize_t)rsize;
+	}
+
+	set_errno();
+	return -1;
+}
+
+ssize_t p_pwrite(int fd, const void *data, size_t size, off64_t offset)
+{
+	HANDLE fh;
+	DWORD wsize = 0;
+	OVERLAPPED ov = {0};
+	LARGE_INTEGER pos = {0};
+	off64_t final_offset = 0;
+
+	/* Fail if the final offset would have overflowed to match POSIX semantics. */
+	if (!git__is_ssizet(size) || git__add_int64_overflow(&final_offset, offset, (int64_t)size)) {
+		errno = EINVAL;
+		return -1;	
+	}
+
+	/*
+	 * Truncate large writes to the maximum allowable size: the caller
+	 * needs to always call this in a loop anyways.
+	 */
+	if (size > INT32_MAX) {
+		size = INT32_MAX;
+	}
+
+	pos.QuadPart = offset;
+	ov.Offset = pos.LowPart;
+	ov.OffsetHigh = pos.HighPart;
+	fh = (HANDLE)_get_osfhandle(fd);
+
+	if (WriteFile(fh, data, (DWORD)size, &wsize, &ov)) {
+		return (ssize_t)wsize;
+	}
+
+	set_errno();
+	return -1;
+}
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -96,6 +96,17 @@ jobs:
             ASAN_SYMBOLIZER_PATH: /usr/bin/llvm-symbolizer-10
             TSAN_OPTIONS: suppressions=/home/libgit2/source/script/thread-sanitizer.supp second_deadlock_stack=1
           os: ubuntu-latest
+        - # Focal, Clang 10, mmap emulation (NO_MMAP)
+          container:
+            name: focal
+          env:
+            CC: clang-10
+            CFLAGS: -DNO_MMAP
+            CMAKE_OPTIONS: -DCMAKE_PREFIX_PATH=/usr/local
+            CMAKE_GENERATOR: Ninja
+            SKIP_SSH_TESTS: true
+            SKIP_NEGOTIATE_TESTS: true
+          os: ubuntu-latest
         - # macOS
           os: macos-10.15
           env:
@@ -112,6 +123,15 @@ jobs:
             ARCH: amd64
             CMAKE_GENERATOR: Visual Studio 16 2019
             CMAKE_OPTIONS: -A x64 -DWIN32_LEAKCHECK=ON -DDEPRECATE_HARD=ON
+            SKIP_SSH_TESTS: true
+            SKIP_NEGOTIATE_TESTS: true
+        - # Windows amd64 Visual Studio (NO_MMAP)
+          os: windows-2019
+          env:
+            ARCH: amd64
+            CMAKE_GENERATOR: Visual Studio 16 2019
+            CFLAGS: -DNO_MMAP
+            CMAKE_OPTIONS: -A x64 -DDEPRECATE_HARD=ON
             SKIP_SSH_TESTS: true
             SKIP_NEGOTIATE_TESTS: true
         - # Windows x86 Visual Studio
--- a/src/unix/posix.h
+++ b/src/unix/posix.h
@@ -101,4 +101,7 @@ GIT_INLINE(int) p_futimes(int f, const struct p_timeval t[2])
 # define p_futimes futimes
 #endif
 
+#define p_pread(f, d, s, o) pread(f, d, s, o)
+#define p_pwrite(f, d, s, o) pwrite(f, d, s, o)
+
 #endif


ethomson:
Cool, thanks for tackling this.  :)

Stylistically, I would slightly prefer it if we didn't interleave big blocks of `ifdef` right before the return.  It feels a little hard to reason about what the mmap emulation in `write_at` is doing.  Let's make them self-contained blocks so that the flow of the function is:

```
static int write_at(git_indexer *idx, const void *data, off64_t offset, size_t size)
{
#ifdef NO_MMAP
    ...mmap emulation goes here...
    return;
#else
    ...mmap usage goes here...
    return;
#endif
}
```

We could also just have `write_at` as an indirection between mmap-using and mmap-not-using functions, but what I don't love is the dangling completion in the `NO_MMAP` case, you have to scroll through a bunch of `ifdef`d out code to see what happens next in the function.

hackworks:
@ethomson I have incorporated the suggested changes. Could I please get some reviews on this pull request?

src/posix.c
```diff
@@ -216,9 +217,44 @@ int p_write(git_file fd, const void *buf, size_t cnt)
...
+	while (nbyte) {
+		ssize_t r;
+		r = pread(fd, b, nbyte, offset);
+		if (r < 0) {
+			if (errno == EINTR || GIT_ISBLOCKED(errno))
+				continue;
+			return -1;
+		}
+		if (!r) {
+			errno = EPIPE;
```
lhchavez, line 244: `EPIPE` is not in the list of errors that can be returned by either [`read(2)`](https://man7.org/linux/man-pages/man2/read.2.html#ERRORS) or [`lseek(2)`](https://man7.org/linux/man-pages/man2/lseek.2.html#ERRORS) (which [`pread(2)`](https://man7.org/linux/man-pages/man2/pread.2.html#ERRORS) claims to be the legal errors).

instead, maybe break here like in the `p_read()` implementation. the caller would still need to check whether the return value was a short read, for consistency with `p_read()`, but that's already done.

src/posix.c
```diff
@@ -266,7 +303,78 @@ int p_munmap(git_map *map)
...
+{
+#ifdef GIT_WIN32
+	/* FIXME: Implement using Win32 overlapped IO */
+
+	ssize_t ret;
+	int nfd = dup(fd);
+	
+	(void) ((ret = p_lseek(nfd, offset, SEEK_SET) < 0 ? -1 : 0) ||
+		((ret = p_write(nfd, buf, nbyte)) < 0) ||
+		(idx->do_fsync && (ret = fsync(nfd) < 0 : -1 : ret)));
```
lhchavez, line 323: where is `idx` declared?

src/posix.c
```diff
@@ -266,7 +303,78 @@ int p_munmap(git_map *map)
...
 
+ssize_t p_pwrite(git_file fd, const void *buf, size_t nbyte, git_off_t offset)
+{
+#ifdef GIT_WIN32
+	/* FIXME: Implement using Win32 overlapped IO */
+
+	ssize_t ret;
+	int nfd = dup(fd);
+	
+	(void) ((ret = p_lseek(nfd, offset, SEEK_SET) < 0 ? -1 : 0) ||
```
lhchavez, line 321: this style doesn't seem to appear anywhere else in the codebase. would it be possible to use the more common

```c
if ((error = ...) < 0)
  goto cleanup;
if ((error = ...) < 0)
  goto cleanup;
...
cleanup:
  close(nfd);
  return error;
```

src/posix.c
hackworks: Coding derp - copy/paste!

src/posix.c
```diff
@@ -216,9 +217,42 @@ int p_write(git_file fd, const void *buf, size_t cnt)
...
+{
+#ifdef GIT_WIN32
+	/* FIXME: Implement using Win32 overlapped IO */
 
-#include "map.h"
+	ssize_t ret;
+	int nfd = dup(fd);
+	
+	(void) ((ret = p_lseek(nfd, offset, SEEK_SET) < 0 ? -1 : 0) ||
+		((ret = p_read(nfd, buf, nbyte)) < 0));
```
lhchavez, line 229: same as with `p_pwrite`.

lhchavez (dismissed):
lgtm (although i am not a reviewer with write access, so my vote is just informative)

0xdky:
@ethomson Could I please get some help to get this merged OR if you have any concerns/review comments, I will be happy to  work on them.

src/indexer.c
```diff
@@ -604,30 +604,8 @@ static void hash_partially(git_indexer *idx, const uint8_t *data, size_t size)
...
-
-	if ((error = p_mmap(&map, page_offset + size, GIT_PROT_WRITE, GIT_MAP_SHARED, fd, page_start)) < 0)
-		return error;
-
-	map_data = (unsigned char *)map.data;
-	memcpy(map_data + page_offset, data, size);
-	p_munmap(&map);
-
+	if (p_pwrite(idx->pack->mwf.fd, data, size, offset) != (ssize_t)size)
+		return -1;
```
ethomson, line 608: I like that you've noticed that there's a posix call that matches our intent here - our posix indirection layer is really meant to use the actual posix functions when they're available, or implement a suitable fallback when they're not.  Can we do that here?

ie, can we just use `pwrite` when it's available?  I think that it's possible that there's no system that both has `pwrite` and lacks `mmap`, but the reason that I ask isn't because I think that the posix function is necessarily super useful for us.  What it _does_ do is act as a baseline in tests to ensure that _our_ `pwrite` implementation acts the same.  This is important because we expect `p_some_posix_fn` to act like `some_posix_fn`.  If it _doesn't_ then we're going to make errors when we call it because this is really meant to be a compatibility layer (at least as much as is possible).

(If our `p_pwrite` function really _isn't_ like `pwrite` then we should just keep calling it `write_at` or something.)

Anyway, this is clever, I wouldn't have thought of emulating `pwrite`.

src/posix.c
```diff
@@ -216,9 +217,48 @@ int p_write(git_file fd, const void *buf, size_t cnt)
 	return 0;
 }
 
-#ifdef NO_MMAP
+ssize_t p_pread(git_file fd, void *buf, size_t nbyte, git_off_t offset)
```
ethomson, line 220: If we have indirection between the `unix` and `win32` compatibility layers, they should go into a file in `unix/` and `win32/` directories (eg `win32/posix_w32.c`) so that we can avoid the ifdefs.

src/posix.c
```diff
@@ -216,9 +217,48 @@ int p_write(git_file fd, const void *buf, size_t cnt)
...
+	const git_off_t cpos = p_lseek(fd, 0, SEEK_CUR);
+
+	if ((error = p_lseek(fd, offset, SEEK_SET) < 0 ? -1 : 0))
+		goto cleanup;
+
+	if ((error = p_read(fd, buf, nbyte)) < 0)
+		goto cleanup;
+
+cleanup:
+	if (-1 == p_lseek(fd, cpos, SEEK_SET))
```
ethomson, line 235: `if (foo == -1)` please, not the backward form.  😄

src/posix.c
```diff
@@ -216,9 +217,48 @@ int p_write(git_file fd, const void *buf, size_t cnt)
...
+ssize_t p_pread(git_file fd, void *buf, size_t nbyte, git_off_t offset)
+{
+#ifdef GIT_WIN32
+	/* TODO: Implement using Win32 overlapped IO to avoid updating fd */
 
-#include "map.h"
+	ssize_t error;
+	const git_off_t cpos = p_lseek(fd, 0, SEEK_CUR);
+
+	if ((error = p_lseek(fd, offset, SEEK_SET) < 0 ? -1 : 0))
```
ethomson, line 228: I had to read this a few times - I think that we could simplify its understandability, at the expense of an extra line or two.  

```
if (p_lseek(fd, offset, SEEK_SET) < 0) {
    error = -1;
    goto cleanup;
}
```

But since we don't really want to propagate the error value, we want to return `-1` on failure, then it might make sense to have an `on_error` label that just returns `-1`.

src/posix.c
```diff
@@ -216,9 +217,48 @@ int p_write(git_file fd, const void *buf, size_t cnt)
...
+		goto cleanup;
+
+	if ((error = p_read(fd, buf, nbyte)) < 0)
```
ethomson, line 231: We probably want out function to return `-1` on any failure, not propagate `p_read`'s failure value

src/indexer.c
```diff
@@ -604,30 +604,8 @@ static void hash_partially(git_indexer *idx, const uint8_t *data, size_t size)
...
-	page_start = offset - page_offset;
-
-	if ((error = p_mmap(&map, page_offset + size, GIT_PROT_WRITE, GIT_MAP_SHARED, fd, page_start)) < 0)
-		return error;
-
-	map_data = (unsigned char *)map.data;
-	memcpy(map_data + page_offset, data, size);
-	p_munmap(&map);
-
+	if (p_pwrite(idx->pack->mwf.fd, data, size, offset) != (ssize_t)size)
```
ethomson, line 607: Also, I don't like casting a `size_t` to an `ssize_t`.  I realize that it's Probably Fine on Most Systems.  But there aren't significant guarantees about these types -- IIRC, `ssize_t` is guaranteed to support a range of `-1` to the largest chunk of data that can be written (or something equally hand waving like that).  If we take a `size_t` and call a thing that reads/writes `ssize_t`s, we should probably do it in a loop, writing as many as `SSIZE_MAX` bytes _or_ `size` at a time (whichever is smaller) until we've written all of the `size` or hit an error.

src/posix.c
```diff
@@ -266,7 +307,84 @@ int p_munmap(git_map *map)
...
+		}
+		if (!r)
+			break;
+
+		nbyte -= r;
+		offset += r;
+		b += r;
+	}
+
+	/* FIXME: p_write should be returning bytes written and not 0 */
```
ethomson, line 357: Not sure that I understand this comment, is it outdated?

src/indexer.c
0xdky: I have implemented p_pwrite using pwrite or p_mmap in src/posix.c since mmap based IO was used prior to this change. If there are systems where mmap based IO is more performant, we will not cause a regression.

I will replace the call of p_pwrite in write_at using a loop to avoid the typecasting of ssize_t to size_t.

Please let me know if I am missing something.

src/posix.c
0xdky: `p_write` in `src/posix.c` returns 0 on success or -1 on error. However, POSIX `write` returns the actual bytes written. Hence, from a consistency point of view, it would be desirable to make that change (not in the scope/related of this PR)

src/posix.c
0xdky: Since this is win32 specific code, I moved it to `win32/posix_w32.c` and used `mmap` based IO for emulating `p_pread` and `p_pwrite`

src/indexer.c
0xdky: Done

0xdky:
@ethomson Thank you for the review and suggestions to refactor the code.

I have incorporated review comments by having unix/win32 specific POSIX emulation. On Unix, I use `pread` + `pwrite` and on win32, I use memory mapped IO since that will avoid having to reset the file descriptor position if I use `read` + `write`.

Since `mmap` itself is emulated on win32, there might be a more direct way to implement `pread` and `pwrite` using overlapped file IO.

0xdky:
@ethomson Build setup is failing for some environments and the results say that it has been failing since last couple of builds but with no additional logs. Could you please help me with this?

ethomson:
> @ethomson Build setup is failing for some environments and the results say that it has been failing since last couple of builds. Could you please help me with this?

I'm deprecating the Azure Pipelines builds, moving over to GitHub Actions fully.  Please ignore the Azure Pipelines failures; apologies for the noise.

ethomson:
If I'm reading this right, you've removed the mmap for Unix altogether?  Was this intentional?

The goal of the `p_` functions is to emulate posix like functionality.  They shouldn't wrap the posix functions and then add more.  `p_pwrite` should be able to be `#define`d directly to `pwrite` in most cases.

The looping logic should be in `write_at`

0xdky:
> If I'm reading this right, you've removed the mmap for Unix altogether? Was this intentional?

I just feel this really reflects my initial intent of just fixing this to support `mmap` emulation. I feel it is best to close this PR and start from fresh. Let me see if I can salvage this PR.

0xdky:
> If I'm reading this right, you've removed the mmap for Unix altogether? Was this intentional?
> 
> The goal of the `p_` functions is to emulate posix like functionality. They shouldn't wrap the posix functions and then add more. `p_pwrite` should be able to be `#define`d directly to `pwrite` in most cases.
> 
> The looping logic should be in `write_at`

0xdky:
> If I'm reading this right, you've removed the mmap for Unix altogether? Was this intentional?

I just feel this really does not reflect my initial intent of just fixing this to support mmap emulation. I feel it is best to close this PR and start from fresh. Let me see if I can salvage this PR.

> The goal of the `p_` functions is to emulate posix like functionality. They shouldn't wrap the posix functions and then add more. `p_pwrite` should be able to be `#define`d directly to `pwrite` in most cases.
> 
> The looping logic should be in `write_at`

src/posix.c has `p_read` and `p_write` that has looping to handle EINTR. I followed those as guidelines. Since we are actually making underlying `pread` and `pwrite` handle EINTR and it is not pertaining to any logic in `write_at`, I feel the EINTR based retry should still be in their `p_...` implementations.

0xdky:
> If I'm reading this right, you've removed the mmap for Unix altogether? Was this intentional?

Fixed it by adding conditional code in `write_at` to use `pwrite` if `NO_MMAP` is defined. Since `write_at` is not a POSIX call, I have retained conditional code in `src/indexer.c` and not moved it elsewhere.

0xdky:
@ethomson Could you please take a look at this updated PR? I would like to start using this functionality to build other workflows over NFS. This is a blocker when using git+libgit2 over NFS.

src/win32/posix_w32.c
```diff
@@ -981,3 +981,55 @@ int p_inet_pton(int af, const char *src, void *dst)
...
+ssize_t p_pread(git_file fd, void *data, size_t size, git_off_t offset)
+{
+	HANDLE fh;
+	DWORD rsize = 0;
+	OVERLAPPED ov = {0};
+	LARGE_INTEGER pos = {0};
+
+	/* TODO: Loop over to read in smaller chunks to support large reads */
+	if (!git__is_uint32(size)) {
+		errno = E2BIG;
```
lhchavez, line 994: maybe

```suggestion
		errno = EINVAL;
```

? `E2BIG` is used for argument lists, like in [`execve(2)`](https://man7.org/linux/man-pages/man2/execve.2.html):

>        E2BIG  The total number of bytes in the environment (envp) and
>              argument list (argv) is too large.

src/indexer.c
```diff
@@ -603,6 +603,20 @@ static void hash_partially(git_indexer *idx, const uint8_t *data, size_t size)
...
 static int write_at(git_indexer *idx, const void *data, off64_t offset, size_t size)
 {
+#ifdef NO_MMAP
+	size_t remaining_size = size;
+
+	/* Handle data size larger that ssize_t */
+	while (remaining_size > 0) {
+		ssize_t nb = p_pwrite(idx->pack->mwf.fd, data,
+				min(remaining_size, SSIZE_MAX),
+				offset + (size - remaining_size));
```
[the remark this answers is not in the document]
0xdky: Thank you! That makes it robust.

src/posix.h
```diff
@@ -119,6 +119,9 @@ typedef int git_file;
 extern ssize_t p_read(git_file fd, void *buf, size_t cnt);
 extern int p_write(git_file fd, const void *buf, size_t cnt);
 
+extern ssize_t p_pread(git_file fd, void *data, size_t size, git_off_t offset);
+extern ssize_t p_pwrite(git_file fd, const void *data, size_t size, git_off_t offset);
```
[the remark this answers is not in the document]
0xdky: I was so used to explicitly handling EINTR, never thought GNU libc has a macro for doing that. Thank you for the above macro and the code does look cleaner now.

src/indexer.c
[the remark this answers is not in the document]
0xdky: `if (remaining_size > (size_t)(SSIZE_MAX - offset)) {
 		errno = EOVERFLOW;
 		return -1;
}`

I replaced it with the above block at the begin of function so that we catch the error early and can fail without making any changes via call to `p_pwrite`

src/posix.c
```diff
@@ -240,21 +240,23 @@ int p_mmap(git_map *out, size_t len, int prot, int flags, int fd, off64_t offset
...
 	}
 
+	out->len = 0;
 	out->data = git__malloc(len);
 	GIT_ERROR_CHECK_ALLOC(out->data);
 
 	if (!git__is_ssizet(len) ||
-		(p_lseek(fd, offset, SEEK_SET) < 0) ||
-		(p_read(fd, out->data, len) != (ssize_t)len)) {
+		(p_pread(fd, out->data, len, offset) != (ssize_t)len)) {
```
lhchavez, line 256: this may need to be rewritten to call it in a loop since short reads are now possible.

src/win32/posix_w32.c
```diff
@@ -981,3 +981,55 @@ int p_inet_pton(int af, const char *src, void *dst)
...
+	HANDLE fh;
+	DWORD wsize = 0;
+	OVERLAPPED ov = {0};
+	LARGE_INTEGER pos = {0};
+
+	/* TODO: Loop over to write in smaller chunks to support large reads */
+	if (!git__is_uint32(size)) {
+		errno = EINVAL;
+		return -1;
+	}
```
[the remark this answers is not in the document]
0xdky: Good call to let the caller drive looping.

However, I am not sure if is a good idea to mimic POSIX limits in Win32 with the following code:
```
if (!git__is_ssizet(size) || git__add_int64_overflow(&final_offset, offset, (int64_t)size)) {
	errno = EINVAL;
	return -1;	
}
```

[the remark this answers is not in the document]
lhchavez: given that `p_pwrite`/`p_pread` is trying to provide a POSIX-compatible implementation of `pwrite`/`pread`, it should in theory behave with the same semantics and impose similar restrictions. otherwise, the caller needs to know what semantics to expect of it, which kind of defeats the purpose of having compatibility wrappers.

(or did I misinterpret your comment?)

src/win32/posix_w32.c
```diff
@@ -981,3 +981,62 @@ int p_inet_pton(int af, const char *src, void *dst)
...
+}
+
+ssize_t p_pwrite(int fd, const void *data, size_t size, off64_t offset)
+{
+	HANDLE fh;
+	DWORD wsize = 0;
+	OVERLAPPED ov = {0};
+	LARGE_INTEGER pos = {0};
+
+	/*
```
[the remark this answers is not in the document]
0xdky: I incorrectly offloaded the responsibility to check for potential overflow to the caller. I agree, the underlying API implementation should try to enforce error handling as early as possible without spring a surprise when the request is retried for completion in a loop.

src/win32/posix_w32.c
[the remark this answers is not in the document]
0xdky: I was wrong in trying to let Win32 behave differently. I agree having similar behavior is better that surprises due to differences in native implementations.

lhchavez (approved):
:tada:

lhchavez:
hmm seems like I still can't merge PRs. but this is ready to go from my side!

thanks for this PR, @0xdky!

0xdky:
> hmm seems like I still can't merge PRs. but this is ready to go from my side!
> 
> thanks for this PR, @0xdky!

Thank you very much @lhchavez for reviewing & insightful feedback. Looking forward to @ethomson help with review+merging please.

ethomson:
This is a _really_ crisp and clean PR.  Thanks so much for the improvements you've made to it, it really fits in nicely.  I also appreciate you going the extra mile with the Windows implementation.

I took a pass through the mmap emulation code and it looks good to me.  I didn't test it exhaustively, but this seems like it's in a very solid state for merging.

I have one or two stylistic notes, but they're more informative and not really blockers, and I want to get this in since you've been waiting on me for a while and I want to land this PR.

Thanks again for your patience and for iterating on a great PR.

src/indexer.c
```diff
@@ -603,6 +603,23 @@ static void hash_partially(git_indexer *idx, const uint8_t *data, size_t size)
...
+		ssize_t nb;
+		HANDLE_EINTR(nb, p_pwrite(idx->pack->mwf.fd, (void *)ptr,
+					  remaining_size, offset));
+		if (nb <= 0)
+			return -1;
+
+		ptr += nb;
+		offset += nb;
+		remaining_size -= nb;
+	}
```
ethomson, line 621: I think that I would have place the return here so that the entire function was encapsulated either in the `#if` or the `#else` blocks.  Not major, but having to jump down to see what was happening next was jarring.  It would have been nice to have each branch fully within their own space so that I didn't have to jump around.

src/posix.h
```diff
@@ -89,6 +89,12 @@
 #define EAFNOSUPPORT (INT_MAX-1)
 #endif
 
+/* Compiler independent macro to handle signal interrpted system calls */
+#define HANDLE_EINTR(result, x) do {					\
```
ethomson, line 93: We tend to avoid macros to do things like this...  again, my preference is generally to avoid having to jump around too much within the code base to understand what a piece of code is doing.  I don't love macros in general, and in particular, I don't like macros that loop (or do flow control).

Again, there are definitely exceptions in our codebase (there's some `vector_foreach` type of macro that is in pretty regular use).  But I think that something that's used just twice is a place where we can safely repeat ourselves without too much downside.  It think that the clarity outweighs the lack of DRYness here.

This is mostly personal opinion, and again, I don't think that it's so important that we shouldn't merge this.  The benefits from this PR far outweigh any macro-related gripes that I have.  😁   Just trying to convey some stylistic thoughts.

ethomson (commented):
Love this overall.  Two quick comments that might be useful for future PRs.  But overall, it looks like you really understand the libgit2 "style".  Great work.

src/indexer.c
0xdky: I agree. Having self-contained blocks makes it easier to read & you know for certain there is no dependency on anything outside the block.

0xdky: I agree. Having self-contained blocks makes it easier to read & you know for certain there is no dependency on anything outside the block.

haampie:
Hi @0xdky / @ethomson, thanks for this, this is a sledgehammer approach for me to work around https://github.com/libgit2/libgit2/issues/4019.

Can we have an equivalent CMake option too? `-DLIBGIT2_NO_MMAP=ON` or so

Issue #5591, opened 2020-08-01
Memory error in `git_mwindow_open` when running against multiple large repositories


andrewhickman:
### Reproduction steps

The Rust program [`here`](https://github.com/andrewhickman/mgit-crash-repro) reliably reproduces the issue on my system when run against a directory with 32 copies of a large repository (I used the [rust repository](https://github.com/rust-lang/rust.git)).

The program spawns a thread for each repository which opens the repository and calls `git_status_list_new`. 

Reducing the number of threads or the size of the repositories seems to prevent the issue occurring.

```
fn main() {
    let join_handles: Vec<thread::JoinHandle<()>> =
        std::fs::read_dir(std::env::args_os().nth(1).expect("expected argument"))
            .unwrap()
            .map(|entry| std::thread::spawn(|| get_status(entry.unwrap().path())))
            .collect();

    for handle in join_handles {
        handle.join().unwrap();
    }
}

fn get_status(path: PathBuf) {
    let repo = git2::Repository::open(path).unwrap();
    let _ = repo.statuses(None).unwrap();
}
```

### Expected behavior

Success (or possibly an resouce exhaustion error?)

### Actual behavior

The program crashes with `(exit code: 0xc0000374, STATUS_HEAP_CORRUPTION)`. The stack trace of the failing thread:
```
mgit-crash-repro.exe!git_mwindow_close_lru_window() Line 269 (c:\Users\andre\.cargo\registry\src\github.com-1ecc6299db9ec823\libgit2-sys-0.12.9+1.0.1\libgit2\src\mwindow.c:269)
mgit-crash-repro.exe!new_window(int fd, __int64 size, __int64 offset) Line 336 (c:\Users\andre\.cargo\registry\src\github.com-1ecc6299db9ec823\libgit2-sys-0.12.9+1.0.1\libgit2\src\mwindow.c:336)
mgit-crash-repro.exe!git_mwindow_open(git_mwindow_file * mwf, git_mwindow * * cursor, __int64 offset, unsigned __int64 extra, unsigned int * left) Line 407 (c:\Users\andre\.cargo\registry\src\github.com-1ecc6299db9ec823\libgit2-sys-0.12.9+1.0.1\libgit2\src\mwindow.c:407)
[...]

PR #5593, opened 2020-08-02
Make the pack and mwindow implementations data-race-free

lhchavez:
This change fixes a packfile heap corruption that can happen when interacting
with multiple packfiles concurrently across multiple threads. This is
exacerbated by setting a lower mwindow open file limit.

This change:

* Renames most of the internal methods in pack.c to clearly indicate
  that they expect to be called with a certain lock held, making
  reasoning about the state of locks a bit easier.
* Splits the `git_pack_file` lock in two: the one in `git_pack_file`
  only protects the `index_map`. The protection to `git_mwindow_file` is
  now in that struct.
* Explicitly checks for freshness of the `git_pack_file` in
  `git_packfile_unpack_header`: this allows the mwindow implementation
  to close files whenever there is enough cache pressure, and
  `git_packfile_unpack_header` will reopen the packfile if needed.
* After a call to `p_munmap()`, the `data` and `len` fields are poisoned
  with `NULL` to make use-after-frees more evident and crash rather than
  being open to the possibility of heap corruption.
* Adds a test case to prevent this from regressing in the future.

Fixes: #5591

### DIFF
--- a/src/pack.c
+++ b/src/pack.c
@@ -12,6 +12,7 @@
 #include "mwindow.h"
 #include "odb.h"
 #include "oid.h"
+#include "oidarray.h"
 
 /* Option to bypass checking existence of '.keep' files */
 bool git_disable_pack_keep_file_checks = false;
@@ -195,7 +196,8 @@ static void pack_index_free(struct git_pack_file *p)
 	}
 }
 
-static int pack_index_check(const char *path, struct git_pack_file *p)
+/* Run with the packfile lock held */
+static int pack_index_check_locked(const char *path, struct git_pack_file *p)
 {
 	struct git_pack_idx_header *hdr;
 	uint32_t version, nr, i, *index;
@@ -301,7 +303,8 @@ static int pack_index_check(const char *path, struct git_pack_file *p)
 	return 0;
 }
 
-static int pack_index_open(struct git_pack_file *p)
+/* Run with the packfile lock held */
+static int pack_index_open_locked(struct git_pack_file *p)
 {
 	int error = 0;
 	size_t name_len;
@@ -324,17 +327,10 @@ static int pack_index_open(struct git_pack_file *p)
 		return -1;
 	}
 
-	if ((error = git_mutex_lock(&p->lock)) < 0) {
-		git_buf_dispose(&idx_name);
-		return error;
-	}
-
 	if (p->index_version == -1)
-		error = pack_index_check(idx_name.ptr, p);
+		error = pack_index_check_locked(idx_name.ptr, p);
 
 	git_buf_dispose(&idx_name);
-
-	git_mutex_unlock(&p->lock);
 
 	return error;
 }
@@ -1015,12 +1011,13 @@ static int packfile_open(struct git_pack_file *p)
 	git_oid sha1;
 	unsigned char *idx_sha1;
 
-	if (p->index_version == -1 && pack_index_open(p) < 0)
-		return git_odb__error_notfound("failed to open packfile", NULL, 0);
-
-	/* if mwf opened by another thread, return now */
 	if (git_mutex_lock(&p->lock) < 0)
 		return packfile_error("failed to get lock for open");
+
+	if (pack_index_open_locked(p) < 0) {
+		git_mutex_unlock(&p->lock);
+		return git_odb__error_notfound("failed to open packfile", NULL, 0);
+	}
 
 	if (p->mwf.fd >= 0) {
 		git_mutex_unlock(&p->lock);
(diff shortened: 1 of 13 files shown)


src/pack.c
```diff
@@ -965,62 +993,75 @@ int get_delta_base(
...
 
 	/* TODO: open with noatime */
 	p->mwf.fd = git_futils_open_ro(p->pack_name);
 	if (p->mwf.fd < 0)
 		goto cleanup;
 
-	if (p_fstat(p->mwf.fd, &st) < 0 ||
-		git_mwindow_file_register(&p->mwf) < 0)
+	if (p_fstat(p->mwf.fd, &st) < 0)
+		goto cleanup;
```
ethomson, line 1062: I realize that you didn't introduce this problem, but while you're here, maybe we could:

```suggestion
	if (p_fstat(p->mwf.fd, &st) < 0) {
		git_error_set(GIT_ERROR_OS, "could not stat packfile");
		goto cleanup;
	}
```

src/pack.c
```diff
@@ -1164,28 +1211,42 @@ int git_packfile_alloc(struct git_pack_file **pack_out, const char *path)
...
 {
-	const unsigned char *index = p->index_map.data;
-	const unsigned char *end = index + p->index_map.len;
+	const unsigned char *index, *end;
+	int error;
+	uint32_t off32;
+	off64_t off;
+
+	if ((error = git_mutex_lock(&p->lock)) < 0)
+		return error;
```
ethomson, line 1222: `git_mutex_*` is one of the outlier set of functions that are prefixed `git_` but don't set error messages when they fail.  They rely on the caller to know that they need to do that and use `GIT_ERROR_OS`.  This stinks and we should fix this.  It should be `p_mutex_lock` or something.  In the meantime, would you mind:

```suggestion
	if ((error = git_mutex_lock(&p->lock)) < 0) {
		git_error_set(GIT_ERROR_OS, "failed to lock packfile reader");
		return error;
	}
```

src/pack.c
```diff
@@ -1197,19 +1258,20 @@ int git_pack_foreach_entry(
...
 	uint32_t i;
 	int error = 0;
 
-	if (index == NULL) {
-		if ((error = pack_index_open(p)) < 0)
-			return error;
-
-		assert(p->index_map.data);
+	if ((error = git_mutex_lock(&p->lock)) < 0)
+		return error;
```
ethomson, line 1266:
```suggestion
	if ((error = git_mutex_lock(&p->lock)) < 0) {
		git_error_set(GIT_ERROR_OS, "failed to lock packfile reader");
		return error;
	}
```

src/pack.c
```diff
@@ -1285,19 +1352,21 @@ static int pack_entry_find_offset(
...
 	off64_t offset;
 	const unsigned char *current = 0;
+	int error;
 
 	*offset_out = 0;
 
-	if (p->index_version == -1) {
-		int error;
+	if ((error = git_mutex_lock(&p->lock)) < 0)
+		return error;
```
ethomson, line 1360:
```suggestion
	if ((error = git_mutex_lock(&p->lock)) < 0) {
		git_error_set(GIT_ERROR_OS, "failed to lock packfile reader");
		return error;
	}
```

src/pack.c
```diff
@@ -1395,11 +1464,26 @@ int git_pack_entry_find(
 	if (error < 0)
 		return error;
 
+	/* if mwf opened by another thread, return now */
+	error = git_mutex_lock(&p->lock);
+	if (error < 0)
+		return error;
```
ethomson, line 1470:
```suggestion
	if (error < 0) {
		git_error_set(GIT_ERROR_OS, "failed to lock packfile reader");	
		return error;
	}
```

src/pack.c
```diff
@@ -1395,11 +1464,26 @@ int git_pack_entry_find(
...
+	error = git_mutex_lock(&p->mwf.lock);
+	if (error < 0) {
+		git_mutex_unlock(&p->lock);
+		return error;
```
ethomson, line 1474:
```suggestion
		git_error_set(GIT_ERROR_OS, "failed to lock packfile reader");	
		return error;
```

src/pack.c
```diff
@@ -965,62 +993,75 @@ int get_delta_base(
  *
  ***********************************************************/
 
-void git_packfile_close(struct git_pack_file *p, bool unlink_packfile)
+/* Run with the mwindow lock held */
+void git_packfile_close_locked(struct git_pack_file *p, bool unlink_packfile)
 {
+	bool locked = true;
+	if (git_mutex_lock(&p->lock) < 0) {
+		git_error_set(GIT_ERROR_OS, "failed to lock pack file");
```
ethomson, line 1001: let's be consistent with "packfile"
```suggestion
		git_error_set(GIT_ERROR_OS, "failed to lock packfile");
```

ethomson (commented):
This is a great change overall.  Just a few minor nits that I noticed.

src/pack.c
lhchavez: huh, i have no idea where i got this from.

src/pack.c
lhchavez: noted! will do that in the future.

lhchavez:
> This is a great change overall. Just a few minor nits that I noticed.

Added and rebased.

lhchavez:
Found that there had been some changes in master that make this change difficult to land as it was. in particular, it now has a strong dependency on the packfile lock refactoring done in #5595, so I now made it depend on that change to make things easier.

note that there is a very scary comment in `script/thread-sanitizer.supp`. as far as i could tell, there's no (current!) way in which a deadlock could arise although that proof is kinda hand-wavy.

i tried several things to disentangle the eviction from the lock, but could not find anything that worked :(

ethomson:
Thanks for these correctness fixes, these are amazing.

Issue #5592, opened 2020-08-02
Make libgit2 ThreadSanitizer-safe


lhchavez:
### Reproduction steps

Build libgit2 with ThreadSanitizer and run the tests. The `object` suite is a good representative example:

```shell
$ cmake \
  -DUSE_BUNDLED_ZLIB=ON \
  -DCMAKE_C_FLAGS="-fsanitize=thread -fno-optimize-sibling-calls -fno-omit-frame-pointer" \
  -DCMAKE_GENERATOR=Ninja \
  ..
$ ninja libgit2_clar && ./libgit2_clar -sobject::cache
Loaded 390 suites: 
Started (test status codes: OK='.' FAILURE='F' SKIPPED='S')

object::cache (cache no blobs).==================
WARNING: ThreadSanitizer: data race (pid=1201398)
  Read of size 4 at 0x7b4400000888 by thread T2:
    #0 git_repository__configmap_lookup ../src/config_cache.c:114 (libgit2_clar+0xbb370)
    #1 git_odb__set_caps ../src/odb.c:675 (libgit2_clar+0x138862)
    #2 git_repository_odb__weakptr ../src/repository.c:1114 (libgit2_clar+0x193317)
    #3 git_object_lookup_prefix ../src/object.c:184 (libgit2_clar+0x13483e)
    #4 git_object_lookup ../src/object.c:253 (libgit2_clar+0x134ad3)
    #5 cache_parsed ../tests/object/cache.c:140 (libgit2_clar+0x39f74a)
    #6 <null> <null> (libtsan.so.0+0x2d1af)

  Previous write of size 4 at 0x7b4400000888 by thread T1:
    #0 git_repository__configmap_lookup ../src/config_cache.c:124 (libgit2_clar+0xbb430)
    #1 git_odb__set_caps ../src/odb.c:675 (libgit2_clar+0x138862)
    #2 git_repository_odb__weakptr ../src/repository.c:1114 (libgit2_clar+0x193317)
    #3 git_repository_odb ../src/repository.c:1135 (libgit2_clar+0x19348e)
    #4 cache_raw ../tests/object/cache.c:162 (libgit2_clar+0x39fac5)
    #5 <null> <null> (libtsan.so.0+0x2d1af)

  Location is heap block of size 280 at 0x7b4400000780 allocated by main thread:
    #0 calloc <null> (libtsan.so.0+0x305ca)
[...]

PR #5594, opened 2020-08-02
Improve the support of atomics

lhchavez:
This change:

* Starts using GCC's and clang's `__atomic_*` intrinsics instead of the
  `__sync_*` ones, since the former supercede the latter (and can be
  safely replaced by their equivalent `__atomic_*` version with the
  sequentially consistent model).
* Makes `git_atomic64`'s value `volatile`. Otherwise, this will make
  ThreadSanitizer complain.
* Adds ways to load the values from atomics. As it turns out,
  unsynchronized read are okay only in some architectures, but if we
  want to be correct (and make ThreadSanitizer happy), those loads
  should also be performed with the atomic builtins.
* Fixes two ThreadSanitizer warnings, as a proof-of-concept that this
  works:
  - Avoid directly accessing `git_refcount`'s `owner` directly, and
    instead makes all callers go through the `GIT_REFCOUNT_*()` macros,
    which also use the atomic utilities.
  - Makes `pool_system_page_size()` race-free.

Part of: #5592

### DIFF
--- a/src/global.c
+++ b/src/global.c
@@ -12,6 +12,7 @@
 #include "sysdir.h"
 #include "filter.h"
 #include "merge_driver.h"
+#include "pool.h"
 #include "streams/registry.h"
 #include "streams/mbedtls.h"
 #include "streams/openssl.h"
@@ -38,7 +39,8 @@ static git_global_init_fn git__init_callbacks[] = {
 	git_stream_registry_global_init,
 	git_openssl_stream_global_init,
 	git_mbedtls_stream_global_init,
-	git_mwindow_global_init
+	git_mwindow_global_init,
+	git_pool_global_init
 };
 
 static git_global_shutdown_fn git__shutdown_callbacks[ARRAY_SIZE(git__init_callbacks)];
--- a/src/pool.c
+++ b/src/pool.c
@@ -21,23 +21,19 @@ struct git_pool_page {
 
 static void *pool_alloc_page(git_pool *pool, size_t size);
 
-static size_t pool_system_page_size(void)
+#ifndef GIT_DEBUG_POOL
+
+static size_t system_page_size = 0;
+
+int git_pool_global_init(void)
 {
-	static git_atomic_ssize cached_size = {0};
-	size_t page_size = 0;
-
-	if ((page_size = (size_t)git_atomic_ssize_get(&cached_size)) == 0) {
-		if (git__page_size(&page_size) < 0)
-			page_size = 4096;
-		/* allow space for malloc overhead */
-		page_size -= (2 * sizeof(void *)) + sizeof(git_pool_page);
-		git_atomic_ssize_set(&cached_size, (int64_t)page_size);
-	}
-
-	return page_size;
+	if (git__page_size(&system_page_size) < 0)
+		system_page_size = 4096;
+	/* allow space for malloc overhead */
+	system_page_size -= (2 * sizeof(void *)) + sizeof(git_pool_page);
+	return 0;
 }
 
-#ifndef GIT_DEBUG_POOL
 int git_pool_init(git_pool *pool, size_t item_size)
 {
 	assert(pool);
@@ -45,7 +41,7 @@ int git_pool_init(git_pool *pool, size_t item_size)
 
 	memset(pool, 0, sizeof(git_pool));
 	pool->item_size = item_size;
-	pool->page_size = pool_system_page_size();
+	pool->page_size = system_page_size;
 
 	return 0;
 }
@@ -114,6 +110,11 @@ bool git_pool__ptr_in_pool(git_pool *pool, void *ptr)
 }
 
 #else
+
+int git_pool_global_init(void)
+{
+	return 0;
+}
 
 static int git_pool__ptr_cmp(const void * a, const void * b)
 {
--- a/src/pool.c
+++ b/src/pool.c
@@ -23,17 +23,18 @@ static void *pool_alloc_page(git_pool *pool, size_t size);
 
 static size_t pool_system_page_size(void)
 {
-	static size_t size = 0;
+	static git_atomic_ssize cached_size = {0};
+	size_t page_size = 0;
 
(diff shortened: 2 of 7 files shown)


ethomson:
I've only had a cursory 👀 so far, but I like what you're doing here.  I want to look a bit more deeply, and I'd love @pks-t 's eyes as well.

src/pool.c
```diff
@@ -23,17 +23,18 @@ static void *pool_alloc_page(git_pool *pool, size_t size);
...
-	if (!size) {
-		size_t page_size;
+	if ((page_size = (size_t)git_atomic_ssize_get(&cached_size)) == 0) {
 		if (git__page_size(&page_size) < 0)
 			page_size = 4096;
 		/* allow space for malloc overhead */
-		size = (page_size - (2 * sizeof(void *)) - sizeof(git_pool_page));
+		page_size -= (2 * sizeof(void *)) + sizeof(git_pool_page);
+		git_atomic_ssize_set(&cached_size, (int64_t)page_size);
 	}
```
pks-t, line 35: As I said in the other MR already: I think we should just use a global initializer instead of penalizing each pool page allocation with atomics, which may be a heavyweight operation on some platforms.

src/thread-utils.h
```diff
@@ -7,12 +7,31 @@
...
+#elif defined(__GNUC__)
+
 # if (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 1))
 #  error Atomic primitives do not exist on this version of gcc; configure libgit2 with -DTHREADSAFE=OFF
 # endif
+# if (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 7))
+#  define GIT_BUILTIN_ATOMIC
+# else
+#  define GIT_BUILTIN_SYNC
+# endif
```
pks-t, line 29: Do we provide any guarantees about our minimum required compiler versions?

src/thread-utils.h
```diff
@@ -7,12 +7,31 @@
...
+#  error Atomic primitives do not exist on this version of clang; configure libgit2 with -DTHREADSAFE=OFF
+# else
+#  define GIT_BUILTIN_ATOMIC
+# endif
+
+#elif defined(__GNUC__)
+
 # if (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 1))
 #  error Atomic primitives do not exist on this version of gcc; configure libgit2 with -DTHREADSAFE=OFF
 # endif
```
pks-t, line 24: `#elif`?

pks-t:
Thanks for doing this, @lchavez! The only thing I'm on the edge about is the pool page size. I think we should just move it into a global initializer and be done about it to not penalize pool usage with atomics. Other than that, things look good to me :+1:

src/thread-utils.h
lhchavez: Derp. Yes.

src/pool.c
lhchavez: done, and even went a step further: since this is only used in the debug alloc pool, only initializing the system page size in that case. "don't pay for what you don't use" and stuff.

src/thread-utils.h
lhchavez: i'm kind of trusting the release docs in this case: https://www.gnu.org/software/gcc/gcc-4.7/changes.html

> Support for atomic operations specifying the C++11/C11 memory model has been added. These new `__atomic` routines replace the existing `__sync` built-in routines.

src/global.c
```diff
@@ -38,7 +39,8 @@ static git_global_init_fn git__init_callbacks[] = {
 	git_stream_registry_global_init,
 	git_openssl_stream_global_init,
 	git_mbedtls_stream_global_init,
-	git_mwindow_global_init
+	git_mwindow_global_init,
+	git_pool_system_page_size_global_init
```
ethomson, line 43:
```suggestion
	git_pool_global_init
```

src/pool.c
```diff
@@ -21,31 +21,27 @@ struct git_pool_page {
...
-		if (git__page_size(&page_size) < 0)
-			page_size = 4096;
-		/* allow space for malloc overhead */
-		page_size -= (2 * sizeof(void *)) + sizeof(git_pool_page);
-		git_atomic_ssize_set(&cached_size, (int64_t)page_size);
-	}
+static size_t system_page_size = 0;
 
-	return page_size;
+int git_pool_system_page_size_global_init(void)
```
ethomson, line 28: Let's make this the init function for the whole of the pool subsystem.  If we ever need to do more, it can be here.  This lets us be a little more general with the name of the init function, and a little briefer as well.

```suggestion
int git_pool_global_init(void)
```

src/pool.c
```diff
@@ -115,6 +111,11 @@ bool git_pool__ptr_in_pool(git_pool *pool, void *ptr)
 
 #else
 
+int git_pool_system_page_size_global_init(void)
```
ethomson, line 114:
```suggestion
int git_pool_global_init(void)
```

src/pool.h
```diff
@@ -135,4 +135,12 @@ extern uint32_t git_pool__open_pages(git_pool *pool);
 #endif
 extern bool git_pool__ptr_in_pool(git_pool *pool, void *ptr);
 
+/**
+ * This function is being called by our global setup routines to
+ * initialize the system pool size.
+ *
+ * @return 0 on success, <0 on failure
+ */
+extern int git_pool_system_page_size_global_init(void);
```
ethomson, line 144:
```suggestion
extern int git_pool_global_init(void);
```

ethomson (commented):
Looks great, a few minor tweaks requested.

lhchavez:
> Looks great, a few minor tweaks requested.

Done and rebased while I was at it.

ethomson:
Thanks @lhchavez!

ethomson (approved):

PR #5595, opened 2020-08-03
Make the odb race-free

lhchavez:
This change adds all the necessary locking to the odb to avoid races in
the backends.
    
Part of: #5592


Alovchin91:
Hi @lhchavez @ethomson ,

This PR breaks libgit2 functionality on Windows arm64: https://github.com/rust-lang/git2-rs/issues/645

I can't say I understand very well what's going on here but I'll do my best to at least find the exact place where it crashes.

If you need to use my device to debug, I'm all in.

Issue #5590, opened 2020-07-31
Repository fetch failure, only on i686


alexcrichton:
### Reproduction steps

First, create a local `main.c` with:

```c
#include <assert.h>
#include <git2/remote.h>
#include <git2/repository.h>
#include <git2/errors.h>
#include <git2/global.h>
#include <stdio.h>

int main() {
  assert(git_libgit2_init() == 1);

  git_repository *repo;
  assert(git_repository_init(&repo, "index-clone", 0) == 0);

  git_remote *remote;
  assert(git_remote_create_anonymous(&remote, repo, "index/.git") == 0);

  git_fetch_options fetch_opts = GIT_FETCH_OPTIONS_INIT;

  char *strings[] = {"HEAD:refs/remotes/origin/head"};
  git_strarray refspecs;
  refspecs.strings = strings;
  refspecs.count = 1;

  int err = git_remote_fetch(remote, &refspecs, &fetch_opts, NULL);
  printf("err = %d\n", err);
  if (err == 0)
    return 0;
  const git_error *last = git_error_last();
  if (last != NULL) {
    printf("message: %s\n", last->message);
    printf("klass: %d\n", last->klass);
  }

  return 0;
}
````

Next, fetch the `index.bz2` file from [this repository](https://github.com/ehuss/zlib-repro). That's actually a tar file so it can be extracted locally to a folder called `index`.

Next the program is executed.

### Expected behavior

Successful clone on both x86_64 and i686

### Actual behavior

On x86_64:

(after building libgit2 with `cmake -G Ninja .. -DCMAKE_BUILD_TYPE=Release`)

```sh
$ gcc main.c -l git2
$ ./a.out
err = 0
```

On i686:

(after building libgit2 with `cmake -G Ninja .. -DCMAKE_C_FLAGS=-m32 -DOPENSSL_SSL_LIBRARY=/usr/lib/i386-linux-gnu/libssl.so -DOPENSSL_CRYPTO_LIBRARY=/usr/lib/i386-linux-gnu/libcrypto.so -DCMAKE_BUILD_TYPE=Release`)

```sh
$ gcc main.c -l git2 -m32
$ ./a.out
err = -1
message: error inflating zlib stream
klass: 5
```

### Version of libgit2 (release number or SHA1)

a83fd5107879d18b31ff8173ea062136256321be (current master)

### Operating system(s) tested

Linux, Ubuntu 16.04

---

[...]

PR #5599, opened 2020-08-05
zstream: handle Z_BUF_ERROR appropriately in get_output_chunk

ethomson:
Our processing loop in git_zstream_get_output_chunk does not handle
`Z_BUF_ERROR` appropriately at the end of a compressed window.

From the zlib manual, inflate will return:

> Z_BUF_ERROR if no progress was possible or if there was not enough
> room in the output buffer when Z_FINISH is used. Note that Z_BUF_ERROR
> is not fatal, and inflate() can be called again with more input and
> more output space to continue decompressing.

In our loop, we were waiting until we got the expected size, then
ensuring that we were at `Z_STREAM_END`.  We are not guaranteed to be,
since zlib may be in the `Z_BUF_ERROR` state where it has consumed a
full window's worth of data, but it doesn't know that it's really at the
end of the stream.  There _could_ be more compressed data, but it
doesn't _know_ that there's not until we make a subsequent call.

We can change the loop to look for the end of stream instead of our
expected size.  This allows us to call inflate one last time when we are
at the end of a window (and in the `Z_BUF_ERROR` state), allowing it to
recognize the end of the stream, and move from the `Z_BUF_ERROR` state
to the `Z_STREAM_END` state.

If we do this, we need another exit condition: when `bytes == 0`, then
no progress could be made and we should stop trying to inflate.  This
will be an error case, caught by the size and/or end-of-stream test.

### DIFF
--- a/src/pack.c
+++ b/src/pack.c
@@ -844,7 +844,10 @@ static int packfile_unpack_compressed(
 		unsigned int window_len;
 		unsigned char *in;
 
-		in = pack_window_open(p, mwindow, *position, &window_len);
+		if ((in = pack_window_open(p, mwindow, *position, &window_len)) == NULL) {
+			error = -1;
+			goto out;
+		}
 
 		if ((error = git_zstream_set_input(&zstream, in, window_len)) < 0 ||
 		    (error = git_zstream_get_output_chunk(data + total, &bytes, &zstream)) < 0) {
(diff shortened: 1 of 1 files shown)


ethomson:
Fixes #5590

pks-t (approved):
I really don't like the zlib API at all. It's so hard to get right and has so many weird edge cases. Anyway, thanks for fixing this, LGTM!

PR #5602, opened 2020-08-06
Build generic internal library interface for thread local storage

implausible:
As the library starts using more threads internally to accelerate certain operations, it becomes more necessary to easily usable TLS handles. My threaded checkout PR I opened awhile ago would be able to address the abundance of allocations I introduced around the `target_path` and `tmp` buffers in `checkout_data` by converting those parts of the `checkout_data` structure to TLS. That way, the performance benefits of not over allocating in single threaded checkout does not suffer at all.

This PR replaces the way we setup the global state on TLS. I was hoping that it might be fairly simple to bring in this change set with #5546 instead of using the proposed tlsdata from that PR.

Thoughts? @ethomson @pks-t

### DIFF
--- a/src/global.c
+++ b/src/global.c
@@ -29,7 +29,6 @@ git_mutex git__mwindow_mutex;
 typedef int (*git_global_init_fn)(void);
 
 static git_global_init_fn git__init_callbacks[] = {
-	git_allocator_global_init,
 	git_hash_global_init,
 	git_sysdir_global_init,
 	git_filter_global_init,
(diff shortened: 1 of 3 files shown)


ethomson:
Sorry about the delay here.  I had a chance to look at this over the weekend.  I think that this generally makes sense.  I'm trying to figure out how to juggle #5546 and this together.  Should be reasonably straightforward.

ethomson:
This is great @implausible, thanks!  While I was reviewing it, I was curious about whether we could do this in a way that avoided allocations, and so I wanted to try it.  I opened up #5720 with that result, so I'll merge there and close this.

I think that this abstraction is a nice improvement, and I have several thoughts for how we can use it.  Thanks again.

implausible:
Ah, that's awesome. I can't remember what it was, but I had run into an issue trying to do it without an allocation. I much prefer your tweaks.

PR #5603, opened 2020-08-09
fetch: allow commits to be fetched by OID

alyssais:
Git can fetch commits by OID directly if the full OID is provided, and the server allows it.

I have this basic functionality implemented here, but I have a number of problems. Mostly they all come down to libgit2 assuming that the ref(s) it’s fetching will always be present in an ls-remote response, which obviously is not the case when fetching by commit hash.

- Constructing the `git_remote_head` by hand feels a bit bad, since basically nowhere else does it. Additionally, I’m not sure I understand the memory model. Is pointing to the refspec src as the name okay? If not, how would I create a string that would be properly freed when no longer in use.

- How would I get FETCH_HEAD to be updated properly? The current implementation does a second ls-remote rather than looking at `remote->refs`, so it looks like it’ll be difficult to add support for OIDs.

- What happens (or should happen) if a branch name collides with an OID?

I figured there was no way I’d be able to work out the answers to these questions on my own, so by opening this PR I’m hoping for some help from somebody who understands git/libgit2 better than I do.

### DIFF
--- a/src/fetch.c
+++ b/src/fetch.c
@@ -51,6 +51,37 @@ static int maybe_want(git_remote *remote, git_remote_head *head, git_odb *odb, g
 	return git_vector_insert(&remote->refs, head);
 }
 
+static int want_oids(git_remote *remote, git_odb *odb)
+{
+	const char *src;
+	git_oid oid;
+	git_refspec *spec;
+	git_remote_head *head;
+	size_t i;
+
+	git_vector_foreach(&remote->active_refspecs, i, spec) {
+		src = git_refspec_src(spec);
+
+		if (git_oid_fromstrp(&oid, src) < 0) {
+			git_error_clear();
+			continue;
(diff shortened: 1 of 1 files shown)


ethomson:
> How would I get FETCH_HEAD to be updated properly? The current implementation does a second ls-remote rather than looking at remote->refs, so it looks like it’ll be difficult to add support for OIDs.

I think that we should be able to update `FETCH_HEAD` correctly with the information that we have, and shouldn't need to do a second `ls-remote`.  We should be doing that _independent_ of needing to support object IDs, I think...

> What happens (or should happen) if a branch name collides with an OID?

Interesting question.  I'm not sure offhand, I always look at what git does in this case to understand what we should do.

ethomson:
This was a little more complex - it was fixed in #6203.

PR #5609, opened 2020-08-17
examples-checkout: discard changes for path

petersalomonsen:
add support for `checkout -- <file>...` to discard changes in working directory


examples/checkout.c
```diff
@@ -250,6 +250,8 @@ int lg2_checkout(git_repository *repo, int argc, char **argv)
 	checkout_options opts;
 	git_repository_state_t state;
 	git_annotated_commit *checkout_target = NULL;
+	git_strarray paths;
```
pks-t, line 253: You can move this declaration into the if-statement to limit its scope.

examples/checkout.c
```diff
@@ -265,12 +267,18 @@ int lg2_checkout(git_repository *repo, int argc, char **argv)
...
-		err = 1;
-		goto cleanup;
+		git_checkout_options copts = GIT_CHECKOUT_OPTIONS_INIT;
+		copts.checkout_strategy = GIT_CHECKOUT_FORCE;
+
+		paths.count = args.argc - args.pos;
+		paths.strings = &args.argv[args.pos];
+		copts.paths = paths;
+
+		err = git_checkout_head(repo, &copts);
```
pks-t, line 280: We should print an error message if the function failed.

pks-t (changes requested):
Two minor comments, but otherwise this looks good to me, thanks a lot!

pks-t:
Oh, one more thing. If there's no paths after the path separator, then we shouldn't do anything but print an error and exit. Otherwise we might discard data as there are no paths

examples/checkout.c
petersalomonsen: Added printing error if `err != 0`. Not sure how to provoke an error in order to test it though.

petersalomonsen:
> Oh, one more thing. If there's no paths after the path separator, then we shouldn't do anything but print an error and exit. Otherwise we might discard data as there are no paths

added check on `paths.count == 0`:

```
if (paths.count == 0) {
	fprintf(stderr, "error: no paths specified\n");
        return -1;
}
```

petersalomonsen:
btw: should there be any tests for the examples?

examples/checkout.c
```diff
@@ -265,12 +266,29 @@ int lg2_checkout(git_repository *repo, int argc, char **argv)
...
+		 * Try to checkout the given path(s)
 		 */
 
-		fprintf(stderr, "unhandled path-based checkout\n");
-		err = 1;
-		goto cleanup;
+		git_checkout_options copts = GIT_CHECKOUT_OPTIONS_INIT;
+		git_strarray paths;
+
+		copts.checkout_strategy = GIT_CHECKOUT_FORCE;
```
ethomson, line 275: Sorry that I'm getting to this quite late.  Why are we upgrading this to a force checkout when paths are specified?  These should be orthogonal.

petersalomonsen: The purpose here is to be able to do the same as `git checkout -- path/to/file.txt` which will cause changes in that file to be reverted.

If I don't upgrade to force, then changes in the given path are not reverted.

Not how sure how to do this without force.

ethomson: then that's a bug somewhere - either the example is misusing the api elsewhere or `git_checkout` isn't working right in this case.  what is the state of the files in the working directory and index before you run this?  are there staged changes?  does the working directory match the index or head or neither?

petersalomonsen: Below is the code from my test case. As you can see it creates a clean repository, writes a file that is added and committed. Then make some modifications and calling the checkout method to discard the changes.

```
        FS.mkdir('test');
        FS.chdir('test');
        lg.callMain(['init', '.']);

        FS.writeFile('test.txt', 'abcdef');
        lg.callMain(['add', 'test.txt']);
        lg.callMain(['commit', '-m', 'test commit']);
        assert.equal(FS.readFile('test.txt', {encoding: 'utf8'}), 'abcdef');
        FS.writeFile('test.txt', 'abcdefg');
        lg.callMain(['checkout', '--', 'test.txt']);
        lg.callMain(['status']);
        assert.equal(FS.readFile('test.txt', {encoding: 'utf8'}), 'abcdef',
                'expecting file content to be reverted');
```

from here: https://github.com/petersalomonsen/wasm-git/blob/master/test/checkout.spec.js#L21

ethomson: Your test should work as-is, without setting the force option.  Just like it works in the git command-line client.  (`['checkout', '--', 'test.txt']` shouldn't imply `--force`.)  This is a bug that we should fix, whether it's in the example or `git_checkout_head`.

petersalomonsen: all right. I'll see if I can find some time to investigate it :-) would be good to sort it out.

ethomson: Just to be clear - I wasn't suggesting that you needed to be the one to do this work.  :)  Just that we should figure out what's going on here.  The examples should show new users the best practices for using the APIs and we don't necessarily want people defaulting to force when it's not (or - at least - shouldn't be) necessary.

petersalomonsen: sure, I didn't get the impression that you suggested I should do it :-) but then it's annoying too, so can't help not trying to figure out what it is :-)

I found something here:
https://github.com/libgit2/libgit2/blob/master/src/checkout.c#L499

changing from:
`*action = CHECKOUT_ACTION_IF(FORCE, UPDATE_BLOB, NONE);`

to

`*action = CHECKOUT_ACTION__UPDATE_BLOB;`

fixes so that I don't have to use `FORCE`.

However a test should be made for this to be sure that this check does not serve other purposes.

petersalomonsen: ok I found a test that relies on this (checking out without given paths should not overwrite modified files by default), so I wrote a bugfix so that this test pass, and also added a new test for checking out with given paths.

does it look ok? @ethomson @pks-t 

if so, would you like me to add an issue, and reference this in this PR (and the commit comment)?

petersalomonsen:
rebased on latest, and also including feature for creating new branch (checkout -b) and setting upstream from https://github.com/petersalomonsen/wasm-git/pull/36

PR #5612, opened 2020-08-21
deps: ntlmclient: fix htonll for Haiku

mmuman:
Use B_HOST_TO_BENDIAN_INT64 for that.

### DIFF
--- a/deps/ntlmclient/compat.h
+++ b/deps/ntlmclient/compat.h
@@ -21,7 +21,9 @@
 # include <stdbool.h>
 #endif
 
-#ifdef __linux__
+#if defined(_WIN32) || defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


pks-t (approved):
Thanks, your changes look good to me. We'll still need to fix macOS 10.9, but that definitely doesn't need to be part of this PR. @ethomson, want to have another look?

pks-t:
Maybe we should just use the generic version I've posted at #5614. There's just too many platforms and it's not like this code is performance critical in any way, so providing a generic platform feels a lot saner to me.

ethomson:
I'm going to merge this as-is while we sort out how to provide a more generic version in #5614

ethomson:
Thanks @mmuman!

mmuman:
:-)

PR #5617, opened 2020-08-27
Fix crash in git_describe_commit when opts are NULL.

docbacardi:
The argument "opts" can be NULL, which selects default options. Do not access
"opts" directly but only the normalized copy.


ethomson:
Good catch.  Thanks @docbacardi!

ethomson (approved):

PR #5618, opened 2020-08-28
Introduce clang-tidy

lhchavez:
This change adds -DENABLE_CLANG_TIDY, which makes the compilation invoke
clang-tidy in parallel. This should help catch a lot more bugs. It also
enables it for the UBSan CI job to avoid slowing the whole CI down.

As proof of concept, only the clang-analyzer-core.NullDereference and a
few other core checks were enabled to avoid changing a ton of things at
once. This is _not_ enabled in tests, since there are a lot more things
to fix there.

A potential memory leak in `annotated_commit_init()` was found and fixed
with this.


ethomson:
> A potential memory leak in annotated_commit_init() was found and fixed
with this.

Our approach to out-of-memory situations on small allocations is to not bother trying to clean up other (small) allocations.  The thinking being that if you can't allocate 5k then you're well and truly hosed and freeing 2k is unlikely to help the situation any.

Dealing with large amounts of memory are a bit different, we shouldn't hold on to them if an allocation fails (because that may be _why_ the allocation failed), nor should we assume that we're in a terrible state of one of them fails (because several GB simply may not be available to us).

If I'm reading this correctly, this looks like it falls into the category of allocation-failure-caused-leaks that we previously decided that we'd just be okay with living with.

I'm not particularly bullish on keeping this policy - it existed before I got here and I don't particularly _love_ it.  In particular, it makes analysis more difficult.  But I think that we should decide if a) we _want_ to change the policy, and if yes, then b) how to adapt to the new policy (incrementally or go fix all the problems).

lhchavez:
> > A potential memory leak in annotated_commit_init() was found and fixed
> > with this.
> 
> Our approach to out-of-memory situations on small allocations is to not bother trying to clean up other (small) allocations. The thinking being that if you can't allocate 5k then you're well and truly hosed and freeing 2k is unlikely to help the situation any.
> 
> Dealing with large amounts of memory are a bit different, we shouldn't hold on to them if an allocation fails (because that may be _why_ the allocation failed), nor should we assume that we're in a terrible state of one of them fails (because several GB simply may not be available to us).
> 
> If I'm reading this correctly, this looks like it falls into the category of allocation-failure-caused-leaks that we previously decided that we'd just be okay with living with.
> 
> I'm not particularly bullish on keeping this policy - it existed before I got here and I don't particularly _love_ it. In particular, it makes analysis more difficult. But I think that we should decide if a) we _want_ to change the policy, and if yes, then b) how to adapt to the new policy (incrementally or go fix all the problems).

makes sense. i'm approaching this from the "makes analysis more difficult" angle. my end goal is to eventually have better guarantees about memory leaks, UB, and thread-safety, and this is just one step towards that.

PR #5620, opened 2020-08-29
patch_parse: handle absence of "index" header for new/deleted cases

dlax:
This follows up on 11de594f85479e4804b07dc4f7b33cfe9212bea0 which added support for parsing patches without extended headers (the "index <hash>..<hash> <mode>" line); issue #5267.

We now allow transition from "file mode" state to "path" state directly if there is no "index", which will happen for patches adding or deleting files as demonstrated in added test case.


ethomson:
LGTM.  Thanks @dlax!

ethomson:
Oops, wrong button!  😨  Reopened and merged.  😀

ethomson (approved):

PR #5621, opened 2020-08-31
Return false instead of segfaulting when checking for default port

kim:
`default_port_for_scheme` returns NULL if the scheme is not one of the
builtin ones. This may cause a segmentation fault if a custom transport
URL happens to contain a port number, and this code path is triggered
(e.g. by setting git_fetch_options->update_fetchhead to 1).


ethomson:
LGTM.  Thanks!

ethomson (approved):

Issue #5573, opened 2020-07-07
Incorrect Use of HTTP Upgrade Header


jriddell:
libgit incorrectly fails if the server it is talking to sends an HTTP upgrade header.

Code causing failure is here
https://github.com/libgit2/libgit2/blame/96a5f38f51bc53895000787542f92d0a3352c026/src/transports/httpclient.c#L1124

This is a misunderstanding of the upgrade header which is a suggestion to upgrade from http/1.1 to http/2 not a requirement.

RFC7230 section 6.7, an Upgrade header in a normal response merely informs the client that the server supports upgrading to other protocols, and the client can ask for such an upgrade in a later request. The server requiring an upgrade is via the 426 Upgrade Required response code, not the mere presence of the Upgrade response header.

csware:
This issue can be closed.

ethomson:
Thanks for the report @jriddell and for the fix @csware

PR #5624, opened 2020-09-08
Don't fail if a HTTP server announces he supports a protocol upgrade

csware:
cf. RFC7230 section 6.7, an Upgrade header in a normal response merely informs the client that the server supports upgrading to other protocols, and the client can ask for such an upgrade in a later request. The server requiring an upgrade is via the 426 Upgrade Required response code, not the mere presence of the Upgrade response header.

(closes issue #5573)

Signed-off-by: Sven Strickroth <email@cs-ware.de>

### DIFF
--- a/src/transports/httpclient.c
+++ b/src/transports/httpclient.c
@@ -1120,11 +1120,6 @@ GIT_INLINE(int) client_read_and_parse(git_http_client *client)
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


ethomson:
LGTM.  Thanks, @csware!

ethomson (approved):

Issue #5506, opened 2020-05-09
validating a branch name?


bradwood:
Hi,

I'm looking for the api equivalent of this cli call
```shell
git check-ref-format --branch <branch_name>
```
I'm aware that I can call `git_reference_is_valid_name()` but this excerpt from `git-check-ref-format(1)` concerns me with this approach:

> With the `--branch` option, the command takes a name and checks if it can be used as a valid branch name (e.g. when creating a new branch). But be cautious when using the previous checkout syntax that may refer to a detached HEAD state. The rule `git check-ref-format --branch $name` implements may be stricter than what `git check-ref-format refs/heads/$name` says (e.g. a dash may appear at the beginning of a ref component, but it is explicitly forbidden at the beginning of a branch name). 

I also found this post downstream on this topic. https://github.com/libgit2/libgit2sharp/issues/267

Any suggestions on this, or plans to implement something? I can knock together a regex if needed, but somehow it feels as this functionality should be in the ref or branch API.

ethomson:
Interesting question.  I think that if branch naming validity is not decidable using `git_reference_is_valid_name` then we should offer `git_branch_is_valid_name` and `git_tag_is_valid_name` as well.

bradwood:
It is definitely different. The libgit2 implementation seems to be the same as that in the cli, but sans the `--branch` switch.

So, a valid ref name to `git_reference_is_valid_name` would be something like `refs/heads/-badbranch` even though `-badbranch` is not a valid branch name.

Also, it requires at least one `/` which a branch name clearly doesn't.

bradwood:
@ethomson - any movement on this one? Cheers

ethomson:
Sorry @bradwood, I'm afraid not.  A pull request would definitely be welcome if this was something that you wanted to implement yourself.  🙂

bradwood:
I would, but my c sucks! I'm actually waiting for the downstream rust
wrapper... I will work around it. Thanks.

csware:
Here a link which additional checks TortoiseGit performs: https://github.com/TortoiseGit/TortoiseGit/blob/5170b17c9c12f0e956a1cf123a56ed11f7437312/src/Git/Git.cpp#L246

ethomson:
We have these functions now! Thanks @csware

PR #5625, opened 2020-09-08
Add git_branch_is_name_valid and git_tag_is_name_valid

csware:
Add separate function to validate branch name.

Cross-ref: issue #5506

### DIFF
--- a/include/git2/tag.h
+++ b/include/git2/tag.h
@@ -365,6 +365,16 @@ GIT_EXTERN(int) git_tag_peel(
  */
 GIT_EXTERN(int) git_tag_dup(git_tag **out, git_tag *source);
 
+/**
+ * Determine whether a tag name is valid
+ *
+ * @param name A tag name to test
+ *
+ * @return 1 if the tag name is acceptable; 0 if it isn't
+ */
+GIT_EXTERN(int) git_tag_is_name_valid(
+	const char* name);
+
 /** @} */
 GIT_END_DECL
 #endif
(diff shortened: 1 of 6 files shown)


include/git2/tag.h
```diff
@@ -365,6 +365,16 @@ GIT_EXTERN(int) git_tag_peel(
  */
 GIT_EXTERN(int) git_tag_dup(git_tag **out, git_tag *source);
 
+/**
+ * Determine whether a tag name is valid
+ *
+ * @param name A tag name to test
+ *
+ * @return 1 if tag name not valid, 0 if it is valid, an error code otherwise.
```
ethomson, line 373: `is_name_valid` is a boolean question.  We should return something truthy if it's true, and false if it's false.  

`if (!git_branch_is_name_valid()) { name is valid }` doesn't make sense.  Further, this matches `git_reference_is_name_valid`.

csware: Updated.

ethomson:
Thanks @csware - your pull request highlighted a problem with the existing `is_name_valid` functions, that they cannot communicate a failure to the user, and it's conflated with an invalid name.  I opened #5659 to deprecate the `is_name_valid` functions and introduce `name_is_valid` functions.

I've included these changes over there, modified to be able to communicate failure to the user, use the new naming conventions, and follow our code formatting and style rules.

ethomson:
Included in #5659 - thanks again.

PR #5626, opened 2020-09-09
boolean config parsing fails in some cases with mapped values

csware:
fixes issue #5555

### DIFF
--- a/tests/config/read.c
+++ b/tests/config/read.c
@@ -955,10 +955,12 @@ void test_config_read__get_mapped(void)
 	cl_git_mkfile("./testconfig", "[header]\n  key1 = 1\n  key2 = true\n  key3\n  key4 = always\n  key5 = false\n  key6 = 0\n  key7 = never\n");
 	cl_git_pass(git_config_open_ondisk(&cfg, "./testconfig"));
 
-	// check aprsing bool and string
+	// check parsing bool and string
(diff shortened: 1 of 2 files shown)


tests/config/read.c
```diff
@@ -928,3 +928,64 @@ void test_config_read__nosection(void)
...
+};
+
+void test_config_read__get_mapped(void)
+{
+	git_config *cfg;
+	int val;
+	int known_good;
+
+	cl_set_cleanup(&clean_test_config, NULL);
+	cl_git_mkfile("./testconfig", "[header]\n  key1 = 1\n  key2 = true\n  key3\n  key4 = always\n  key5 = false\n  key6 = 0\n  key7 = never\n");
```
lhchavez, line 955: do you mind also adding `on` and `off`? those two values don't seem to be covered by any other test cases.

(optional: for clarity, would it make sense to split the individual lines in the file? that can be achieved similarly to

https://github.com/libgit2/libgit2/blob/1e987525b3bd86fc29e7ff67d973861efd22faf7/tests/config/include.c#L61-L64

it took me a while to find the key that did not have a value ^^;; )

lhchavez (approved):
(not a contributor with write access, so my vote is only informational)

ethomson (approved):

Issue #5627, opened 2020-09-09
Compilation fails on FreeBSD


interkosmos:
### Reproduction steps
On FreeBSD, run:
```
$ git clone https://github.com/libgit2/libgit2
$ cd libgit2/
$ mkdir build && cd build/
$ cmake ..
$ cmake --build .
[ 27%] Built target git2internal
[ 28%] Built target ntlmclient
[ 28%] Built target http-parser
[ 28%] Built target git2
[ 28%] Building C object tests/CMakeFiles/libgit2_clar.dir/clar.c.o
In file included from /tmp/libgit2/tests/clar.c:786:
/tmp/libgit2/tests/clar/fs.h:277:11: fatal error: 'copyfile.h' file not found
# include <copyfile.h>
          ^~~~~~~~~~~~
1 error generated.
gmake[2]: *** [tests/CMakeFiles/libgit2_clar.dir/build.make:582: tests/CMakeFiles/libgit2_clar.dir/clar.c.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:289: tests/CMakeFiles/libgit2_clar.dir/all] Error 2
gmake: *** [Makefile:160: all] Error 2
```       

### Expected behavior
Compiles on FreeBSD.

### Actual behavior
Compilation fails with error stating that `copyfile.h` is missing.

### Version of libgit2 (release number or SHA1)
1e987525b3bd86fc29e7ff67d973861efd22faf7

### Operating system(s) tested
FreeBSD 12.1 (amd64)

pks-t:
This has been fixed via #5628, right? I'm thus closing the issue, thanks a lot for your report and fix!

interkosmos:
Yes.

PR #5628, opened 2020-09-09
Fixed includes for FreeBSD

interkosmos:
Please review this request that fixes the includes in `tests/clar/fs.h` for FreeBSD. See issue #5627.

### DIFF
--- a/tests/clar/fs.h
+++ b/tests/clar/fs.h
@@ -273,8 +273,14 @@ cl_fs_cleanup(void)
 # include <sys/sendfile.h>
 #endif
 
-#if defined(__APPLE__) || defined(__FreeBSD__)
+#if defined(__APPLE__)
 # include <copyfile.h>
+#endif
+
+#if defined(__FreeBSD__)
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/uio.h>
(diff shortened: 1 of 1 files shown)


ethomson:
Oops!  Sorry about that, this was my bad.

I think that you'll also want to remove the `__FreeBSD__` test in `fs_copyfile_helper`:

```
#if USE_FCOPYFILE && defined(__APPLE__)
```

tests/clar/fs.h
```diff
@@ -273,10 +273,16 @@ cl_fs_cleanup(void)
...
-#if defined(__APPLE__) || defined(__FreeBSD__)
+#if defined(__APPLE__)
 # include <copyfile.h>
 #endif
 
+#if defined(__FreeBSD__)
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/uio.h>
+#endif
```
ethomson, line 284: I don't think that this is necessary.  `sys/types.h` is already included (above) and none of these changes appear to be necessary on my FreeBSD 12 system.

ethomson:
LGTM.  Thanks for fixing this!

ethomson (approved):

Issue #5459, opened 2020-03-23
config parsing can break on multiline with multiple quotes


ehuss:
### Reproduction steps
Create a `.gitconfig` with the following content and attempt to load it with libgit2:

```
[alias]
    foo = !ls  "x" \
        ls  "# comment2" \
        $HOME
```

### Expected behavior
libgit2 should be able to parse this into:
`!ls  x         ls  # comment2         $HOME`

### Actual behavior
Fails with `invalid configuration key`

### Version of libgit2 (release number or SHA1)
be36db28f64de8ed5ca9870708a56b6668c6c482

### Notes
I *think* the crux of the issue is [this line](https://github.com/libgit2/libgit2/blob/9a1024465cf3650d513d3e997435683fcf518e2b/src/config_parse.c#L354).  AIUI, when it is parsing the second line `in_quotes = 2` because `parse_variable` passes the quote count.  `!!in_quotes` is a value of 1 which means it thinks it is still within quotes.  A change roughly like this seems like a start (although the `in_quotes` could probably be renamed to `quote_count` consistently throughout instead of trying to use `%2` in various places):

```diff
diff --git a/src/config_parse.c b/src/config_parse.c
index 48ad1164f..8c0efbbbc 100644
--- a/src/config_parse.c
+++ b/src/config_parse.c
@@ -351,7 +351,7 @@ static int parse_multiline_variable(git_config_parser *reader, git_buf *value, i
                }

                /* If it was just a comment, pretend it didn't exist */
-               quote_count = strip_comments(line, !!in_quotes);
+               quote_count = strip_comments(line, in_quotes);
                if (line[0] == '\0')
                        goto next;
```

Reported by a user which has this elaborate alias: https://github.com/laughinghan/dotfiles/blob/cd846f8069fb10fb25df072da50ec6d9f25f98e7/.gitconfig#L34

csware:
This is still an issue with 1e987525b3bd86fc29e7ff67d973861efd22faf7 it's even worse: teh config file with a new line only consisting of "$HOME" cannot be parsed at all.

I compiled two new test cases for this issue: https://github.com/csware/libgit2/tree/config-multiline-parse

ethomson:
Fixed via #5629 - thanks @csware!

PR #5629, opened 2020-09-09
Fix config file parsing with multi line values containing quoted parts

csware:
Cross-ref: issue #5459

### DIFF
--- a/src/config_parse.c
+++ b/src/config_parse.c
@@ -451,7 +451,7 @@ static int parse_variable(git_config_parser *reader, char **var_name, char **var
 			git_buf_attach(&multi_value, value, 0);
 			value = NULL;
 
-			if (parse_multiline_variable(reader, &multi_value, quote_count) < 0 ||
+			if (parse_multiline_variable(reader, &multi_value, quote_count % 2) < 0 ||
 			    git_buf_oom(&multi_value)) {
 				error = -1;
 				git_buf_dispose(&multi_value);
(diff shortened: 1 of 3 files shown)


src/config_parse.c
```diff
@@ -351,7 +351,7 @@ static int parse_multiline_variable(git_config_parser *reader, git_buf *value, i
 		}
 
 		/* If it was just a comment, pretend it didn't exist */
-		quote_count = strip_comments(line, !!in_quotes);
+		quote_count = strip_comments(line, in_quotes % 2);
```
pks-t, line 354: Is this change even required now? With the below change for how we call `parse_multiline_variable`, were now only ever passing either `0` or `1` to this function, aren't we?

tests/resources/config/config12
```diff
@@ -11,3 +11,10 @@ and other stuff !@#"
 multiline comment # with ;\n special chars \
 and other stuff !@#"  #^^^
       back = "this is \ba phrase"
+      dollar = some $sign
+      multiquotes = !ls  "x" \
+                    ls  "# comment2" \
+                    $HOME
+      multiquotes2 = !ls  "x" \
+                     ls  "\"# comment2" \
+                     $HOME\"
```
pks-t, line 20: Oof. I sometimes really wonder how people come up with such config files :)

csware: Yes, it seems to work without this change.

src/config_parse.c
csware: Yes, it seems to work without this change.

tests/resources/config/config12
csware: Don't ask ;)

ethomson (approved):

PR #5632, opened 2020-09-10
Fix typo: Make ifndef macroname the same as the define name

csware:
Make ifndef macroname the same as the define name


pks-t (approved):
Looks good to me. Thanks, @csware!

PR #5636, opened 2020-09-18
Fix release build warnings

pks-t:
This fixes some warnings generated by GCC when building with `-DCMAKE_BUILD_TYPE=Release`.


ethomson:
I'm surprised those only show up in release builds!

ethomson (approved):

pks-t:
My guess is that the compiler only realises these because it's inlining functions more aggressively. But yeah, compilers are mostly a set of black magic anyway :)

Issue #5622, opened 2020-08-31
need update to libpcre 8.43 


squarti:
libgit2 consumes libpcre 8.42, which has the following security advisories:

https://vuln.whitesourcesoftware.com/vulnerability/CVE-2019-20838

Our vulnerability scanner, which is flagging us as being insecure because of these, suggests moving up to at least v8.43 of this dependency to mitigate these issues.

See https://github.com/libgit2/libgit2/blob/master/deps/pcre/pcre.h#L45

csware:
Wouldn't this also be the option to upgrade to pcre2 10.x?

ethomson:
Fixed via #5637 

@csware why do you want pcre2?

csware:
pcre is kinda deprecated.

PR #5637, opened 2020-09-18
deps: pcre: upgrade to v8.44

pks-t:
With CVE-2019-20838, a vulnerability was found in PCRE v8.42 and lower
causing a buffer overread with crafted queries. This commit thus
upgrades the vendored copy of libpcre from v8.42 to v8.44.

Fixes https://github.com/libgit2/libgit2/issues/5622


pks-t:
This PR is once again a record of our CMake-crimes :( Builds are failing because we declare all warnings globally, which really isn't something we shouild be doing.

ethomson (commented):
We made a number of changes to pcre to exclude unnecessary things and ensure that we could use it as a direct dependency instead of trying to make it work in its literal form.  It's disappointing to change dependencies - since there could be more complex integrations of new versions.  But since pcre (the original pcre, the version that _we_ use) is in maintenance mode, I think that the simple solution here is to apply patches from 8.42 to 8.44 instead of trying to boil the ocean and get it in literally.

ethomson:
See #5649

ethomson:
Fixed via #5649

PR #5638, opened 2020-09-18
Target-based CMake pt. 1

pks-t:
This converts some smallish subset of our CMake build instructions to use target-based properties instead of the global ones. This should probably unblock https://github.com/libgit2/libgit2/pull/5637


ethomson:
We've made several changes to our cmake configuration; I think that we've improved here, but happy to make further changes.

PR #5644, opened 2020-09-29
Feature: Fix file lock

### LINKED ISSUE #5642: [Bug] Merge Failed if there is a `.lock` file exists
### Reproduction steps
I use the `git_merge` to merge other branch into current branch. But I always get a error: `failed to lock file 'Podfile.lock' for writing:`. 
I try to use the `git merge` command to merge the branch, the git will merge success with conflict: The `Podfile` have conflict.
I check the `git_filebuf_open_withsize` function in `libgit2`, I get:
```

		/* create the locking path by appending ".lock" to the original */
		GIT_ERROR_CHECK_ALLOC_ADD(&alloc_len, path_len, GIT_FILELOCK_EXTLENGTH);
		file->path_lock = git__malloc(alloc_len);
		GIT_ERROR_CHECK_ALLOC(file->path_lock);

		memcpy(file->path_lock, file->path_original, path_len);
		memcpy(file->path_lock + path_len, GIT_FILELOCK_EXTENSION, GIT_FILELOCK_EXTLENGTH);
```
The `file->path_lock` will append the `.lock` to the original. And the original path is `Podfile`, so the `path_lock` is `Podfile.lock`. But the `Podfile.lock` exists!!! 

Is it a bug?

### Expected behavior

Merge success with a conflict file.

### Actual behavior

Get a error: `failed to lock file 'Podfile.lock' for writing:`
v1.0.1
macOS 10.15

Whirlwind:
Fix #5642


ethomson:
I don't think that this is going to be sufficient - it just moves the problem to have a different extension.  This might be a little less likely to occur, but if somebody accidentally checks in one of these files - as people are want to do - then we've really only moved the goalposts.

We need to do a test and increment so that if `.lock` exists, we move to `.lock~0` or something.  There's prior art here in creating files during D/F conflicts in the checkout code.  Let's abstract that out and reuse it.

ethomson:
Also, most of our lock files _should_ be using `.lock`.  This is necessary for compatibility with git.  It's a _lockfile_ after all, and its presence indicates that concurrent operations can't happen.  Changing its name means that now we only have that signal with ourselves, which breaks the ecosystem.

Locking references (for example) needs to continue to be done with `.lock`, in this system.  Checkout should use a different function for writing files.

ethomson:
Closing as stale

PR #5645, opened 2020-09-30
Export CMake target

maharmstone:
Hi,

This patch makes it so libgit2 exports the files needed for cmake to recognize it as a package - i.e., from other projects you can do:

```
find_package(git2 REQUIRED)
target_link_libraries(project git2)
```

Tested with Gentoo and on Windows.

This is necessary in particular when compiling a cmake project using MSVC and vcpkg - there doesn't seem to be any easy way of getting it to find git2.lib otherwise.

Thanks


CMakeLists.txt
```diff
@@ -299,3 +300,12 @@ ENDIF()
...
 FEATURE_SUMMARY(WHAT DISABLED_FEATURES DESCRIPTION "Disabled features:")
+
+configure_package_config_file(
+    "git2Config.cmake.in" "${CMAKE_CURRENT_BINARY_DIR}/git2Config.cmake"
+    INSTALL_DESTINATION "lib/cmake/git2"
+)
+
+install(EXPORT git2-targets DESTINATION lib/cmake/git2)
+
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/git2Config.cmake" DESTINATION "lib/cmake/git2")
```
ethomson, line 311: No camel casing here, please.

Is there any prior art you can point me to here about nomenclature?  I feel like this should be `libgit2`, not `git2`.

ethomson:
Neat, this seems reasonable.  A few minor nits.

maharmstone:
Thanks. I agree - I've force-pushed a new version with the changes made.

I had to rename the library target from git2 to libgit2, and then give it a blank prefix so the file wouldn't get called "liblibgit2.so".

maharmstone:
And again - I needed to change examples/CMakeLists.txt as well.

ethomson:
@pks-t has been doing a lot of work on cmake so having his set of eyes would also be helpful. :)

maharmstone:
I've rebased this against the current master. Would somebody be able to review it before the next release, please?

maharmstone:
Good question. This is what I ended up doing instead in my CMakeLists.txt:

```
find_package(PkgConfig REQUIRED)
pkg_check_modules(LIBGIT2 REQUIRED libgit2)

target_include_directories(target PUBLIC ${LIBGIT2_INCLUDE_DIRS})
target_link_libraries(target ${LIBGIT2_LINK_LIBRARIES})
```

PR #5652, opened 2020-10-06
Sample implementation of `git log --follow <pathspec>`

artagnon:
This is a work-in-progress, and any hints would be appreciated.

Thanks.

### DIFF
--- a/examples/log.c
+++ b/examples/log.c
@@ -421,8 +421,6 @@ static int parse_options(
 			else
 				/** Try failed revision parse as filename. */
 				break;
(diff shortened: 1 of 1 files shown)


artagnon:
Okay, I think I have a better handle on things now: `git_diff_tree_to_tree` needs to be taught to be smarter about initializing the `start` and `end` of the `tree_iterators`; I'm not sure how we can avoid going through all the tree entries. And then, the comparison in `git_diff__from_iterators` needs to be updated to run diffs against non-matching tree entries.

PR #5654, opened 2020-10-06
Change bare free to allocator free (fixes #5653)

### LINKED ISSUE #5653: Fix unbalanced custom allocator malloc/free in path.c
The free here is of a pointer that was malloc'd with git__malloc:
https://github.com/libgit2/libgit2/blob/master/src/path.c#L2048

### Reproduction steps
(code inspection)

Use a custom allocator.
Exercise the git_path_validate_system_file_ownership codepath at the above line with a non-NULL info pointer.

### Expected behavior
All malloc/free are paired.

### Actual behavior
With a custom allocator, the free can result in a crash or other bad behavior (this will be easier to spot with instrumentation like Address Sanitizer).

1.0.1

macOS 10.15.4/Windows 10

Pull request coming shortly. The fix here is quite trivial.

dan-tull:
The info pointer was allocated with git__malloc, so needs to be free'd with git__free.
This bug can lurk pretty easily since if there's no custom allocator this is fine.

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -2045,7 +2045,7 @@ int git_path_validate_system_file_ownership(const char *path)
 		git_error_set(GIT_ERROR_INVALID, "programdata configuration file owner is not valid");
 		ret = GIT_ERROR;
 	}
-	free(info);
+	git__free(info);
 
 cleanup:
 	if (descriptor)


dan-tull:
I suppose I should have made a separate branch for this, but honestly I only forked/cloned it to make this specific bugfix, so I don't have any intention of pushing other commits to that branch until this resolves.

ethomson:
Good find.  Thanks for the fix.

PR #5656, opened 2020-10-07
Fix error return for invalid extensions.

ehuss:
`git_repository_open_ext` and `repo_init_config` were returning 1 instead of -1 when there is an unsupported extension in the config.


ethomson:
Oof.  Good catch, thanks for the fix.

ethomson (approved):

PR #5659, opened 2020-10-11
Deprecate `is_valid_name` functions; replace with `name_is_valid` functions

ethomson:
`git_reference_is_name_valid` and `git_remote_is_name_valid` can fail due to issues like out-of-memory, but cannot communicate this failure to the user.  Replace them with functions that can communicate failure as usual (with an integer return value) and communicate validity with an out param.

The new functions are `name_is_valid`, which better communicates that you're checking the validity of a given name, not checking the name's validity of a given reference/remote.

Include #5625, but with the new format.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,3 +1,18 @@
+vNext
+-----
+
+### Changes or improvements
+
+* Branch and tag name validation functions have been introduced:
+  `git_branch_name_is_valid` will check if a branch name is valid,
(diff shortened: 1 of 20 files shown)


csware:
This PR seems to include lots of unrelated changes.

ethomson:
> This PR seems to include lots of unrelated changes.

Indeed, I rebased this on #5658 instead of master.  Thanks for the heads up.

PR #5661, opened 2020-10-12
Make the Windows leak detection more robust

lhchavez:
This change:

* Increases MY_ROW_LIMIT to 2M, since it has been failing in #5595's
  tests since it's _super_ close to the limit.
* Calls `git_repository_free()` on a `git_repository` that was being
  leaked only in Windows.
* Marks the global `git_repository` on `tests/repo/init.c` as `NULL`
  after being freed to make any accidental access more noisy.
* Uses `cl_assert_equal_i()` in `test_trace_windows_stacktrace__leaks`
  to make the test failures more actionable.
* Renames the globals in `tests/repo/init.c` so that they don't start
  with an underscore.

### DIFF
--- a/src/win32/w32_crtdbg_stacktrace.c
+++ b/src/win32/w32_crtdbg_stacktrace.c
@@ -57,7 +57,7 @@ static CRITICAL_SECTION g_crtdbg_stacktrace_cs;
  * it and try again.
  */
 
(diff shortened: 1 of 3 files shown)


ethomson:
Great stuff.  Thanks!

ethomson (approved):

Issue #5663, opened 2020-10-13
Building without threads gives linking error for: git___load


jeroen:
I run into a bug introduced by https://github.com/libgit2/libgit2/pull/5594 when building `-DTHREADSAFE=OFF`. 

```
[ 47%] Building C object src/CMakeFiles/git2internal.dir/index.c.obj
In file included from C:/msys64/home/mingw-packages/mingw-w64-libgit2/src/libgit2-1.1.0/src/common.h:81,
                 from C:/msys64/home/mingw-packages/mingw-w64-libgit2/src/libgit2-1.1.0/src/index.h:10,
                 from C:/msys64/home/mingw-packages/mingw-w64-libgit2/src/libgit2-1.1.0/src/index.c:8:
C:/msys64/home/mingw-packages/mingw-w64-libgit2/src/libgit2-1.1.0/src/index.c: In function 'git_index_set_caps':
C:/msys64/home/mingw-packages/mingw-w64-libgit2/src/libgit2-1.1.0/src/thread-utils.h:344:32: warning: implicit declaration of function 'git___load'; did you mean 'git__load'? [-Wimplicit-function-declaration]
 #define git__load(ptr) (void *)git___load((void * volatile *)&ptr)
                                ^~~~~~~~~~
C:/msys64/home/mingw-packages/mingw-w64-libgit2/src/libgit2-1.1.0/src/thread-utils.h:344:32: note: in definition of macro 'git__load'
 #define git__load(ptr) (void *)git___load((void * volatile *)&ptr)
                                ^~~~~~~~~~
C:/msys64/home/mingw-packages/mingw-w64-libgit2/src/libgit2-1.1.0/src/index.c:50:46: note: in expansion of macro 'GIT_REFCOUNT_OWNER'
 #define INDEX_OWNER(idx) ((git_repository *)(GIT_REFCOUNT_OWNER(idx)))
                                              ^~~~~~~~~~~~~~~~~~
C:/msys64/home/mingw-packages/mingw-w64-libgit2/src/libgit2-1.1.0/src/index.c:574:26: note: in expansion of macro 'INDEX_OWNER'
   git_repository *repo = INDEX_OWNER(index);
                          ^~~~~~~~~~~
```

Eventually the build succeeds, however the `libgit2.a` library gives linking errors looking for `git___load`. I think the problem is that `git___load` is defined within a `#ifdef GIT_THREADS` block, but not the macro that uses it.

[...]

PR #5664, opened 2020-10-13
Define `git___load` when building with `-DTHREADSAFE=OFF`

lhchavez:
This should allow folks that build in non-thread-safe environments to
still be able to build the library.

Fixes: #5663

### DIFF
--- a/src/thread-utils.h
+++ b/src/thread-utils.h
@@ -311,6 +311,11 @@ GIT_INLINE(volatile void *) git___swap(
 	return old;
 }
 
+GIT_INLINE(volatile void *) git___load(void * volatile *ptr)
+{
+	return *ptr;
+}
+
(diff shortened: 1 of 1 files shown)


jeroen:
Maybe we should also add a CI build that builds without threads?

ethomson:
> Maybe we should also add a CI build that builds without threads?

If being able to build in a non-threaded environment has value, then yeah, we should probably have a CI step that runs this.  Maybe only in our nightlies because it's so rare to break something here and our matrix is already huge.

ethomson:
Thanks @lhchavez!

On a tangential note, I agree that having CI for non-threaded systems would be helpful.  In addition, this file is getting a little complicated, we might want to think about how to reduce the cyclomatic complexity.  😓

ethomson (approved):

PR #5686, opened 2020-10-24
Update describe.c

siddhantrao23:
Fixes issue #5672 

Uses the return value of the `git_oid_tostr()` to set error in `git_error_set()` which was previously not done.


ethomson:
I don't think that the problem as described is that the return value isn't being used, it's that if the function were to fail, then we're ignoring that failure.  This doesn't seem to change that.

PR #5690, opened 2020-10-31
Fix the `-DTHREADSAFE=OFF` build

lhchavez:
This change avoids using the `(void)0` construct for some of the mutex
`#define`s, since that makes the "return type" of those "functions" to
be `void` instead of `int`.

### DIFF
--- a/src/thread-utils.h
+++ b/src/thread-utils.h
@@ -235,37 +235,38 @@ GIT_INLINE(int64_t) git_atomic64_get(git_atomic64 *a)
 
 #else
 
-GIT_INLINE(int) git_threads_global_init(void) { return 0; }
+GIT_INLINE(int) git___noop(void) { return 0; }
+
+#define git_threads_global_init git___noop
 
 #define git_thread unsigned int
-#define git_thread_create(thread, start_routine, arg) 0
-#define git_thread_join(id, status) (void)0
+#define git_thread_create(thread, start_routine, arg) git___noop()
+#define git_thread_join(id, status) git___noop()
 
 /* Pthreads Mutex */
 #define git_mutex unsigned int
-GIT_INLINE(int) git_mutex_init(git_mutex *mutex) \
-	{ GIT_UNUSED(mutex); return 0; }
-GIT_INLINE(int) git_mutex_lock(git_mutex *mutex) \
-	{ GIT_UNUSED(mutex); return 0; }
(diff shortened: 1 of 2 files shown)


lhchavez:
oops!

ethomson:
It should be returning an `int`?  Is there a platform where it's not?

our win32 function [returns an `int`](https://github.com/libgit2/libgit2/blob/20450cbe7d161abbb3eb52d96867aeaa6dd99e12/src/win32/thread.h#L50)

posix is just a [define](https://github.com/libgit2/libgit2/blob/20450cbe7d161abbb3eb52d96867aeaa6dd99e12/src/unix/pthread.h#L28) for pthread which [returns an `int`](https://linux.die.net/man/3/pthread_mutex_unlock)

and the no-thread version should [be `0`](https://github.com/libgit2/libgit2/blob/20450cbe7d161abbb3eb52d96867aeaa6dd99e12/src/thread-utils.h#L250)

lhchavez:
> It should be returning an `int`? Is there a platform where it's not?
> 
> [...]
> 
> and the no-thread version should [be `0`](https://github.com/libgit2/libgit2/blob/20450cbe7d161abbb3eb52d96867aeaa6dd99e12/src/thread-utils.h#L250)

the no-thread version! that's a `void`, not an `int`. i think the solution is to avoid using the `(void)0` construct and instead use a real inline function that returns 0.

ethomson:
> the no-thread version! that's a void, not an int. i think the solution is to avoid using the (void)0 construct and instead use a real inline function that returns 0.

ah so it is.  i focused on the `0` and didn't think about the cast in front of it.  i think an inline function that returns `0` makes sense.  certainly it's easier to reason about than what's there.  😢

src/thread-utils.h
```diff
@@ -235,37 +235,38 @@ GIT_INLINE(int64_t) git_atomic64_get(git_atomic64 *a)
 
 #else
 
-GIT_INLINE(int) git_threads_global_init(void) { return 0; }
+GIT_INLINE(int) git___noop(void) { return 0; }
```
lhchavez, line 238: couldn't think of a better name. also, lmk if we need to do something clever to explicitly mark the arguments as unused (right now it's a non-issue on my machine).

ethomson:
Thanks for this!

ethomson (approved):

PR #5692, opened 2020-11-06
Add missing worktree_dir check and test case

### LINKED ISSUE #5280: git_worktree_validate return true if the worktree dir not exists.
I add a worktree to my repo. After remove the worktree path from my finder, the `git_worktree_validate` still return 0.

I check the code in the `worktree.c`:

https://github.com/libgit2/libgit2/blob/b246bed5ab83035d8aef95f1b7ff10dd746db7cb/src/worktree.c#L235-L262

It seems that it does not check the `worktree_path` exists. It check the `gitdir_path`、`commondir_path` and `parent_path`, but these paths are the origin git path, it exists.

rbmclean:
Fixes #5280

### DIFF
--- a/src/worktree.c
+++ b/src/worktree.c
@@ -262,7 +262,7 @@ int git_worktree_validate(const git_worktree *wt)
 	
 	if (!git_path_exists(wt->worktree_path)) {
 		git_error_set(GIT_ERROR_WORKTREE,
-			"worktree directory ('%s') does not exist ",
+			"worktree directory '%s' does not exist",
 			wt->worktree_path);
 		return GIT_ERROR;
 	}
--- a/src/worktree.c
+++ b/src/worktree.c
@@ -259,7 +259,14 @@ int git_worktree_validate(const git_worktree *wt)
 			wt->commondir_path);
 		return GIT_ERROR;
 	}
-
+	
+	if (!git_path_exists(wt->worktree_path)) {
+		git_error_set(GIT_ERROR_WORKTREE,
+			"worktree directory ('%s') does not exist ",
+			wt->worktree_path);
+		return GIT_ERROR;
+	}
+	
 	return 0;
 }
 
--- a/tests/worktree/worktree.c
+++ b/tests/worktree/worktree.c
@@ -616,11 +616,8 @@ void test_worktree_worktree__validate_invalid_worktreedir(void)
 	git_worktree *wt;
 
 	cl_git_pass(git_worktree_lookup(&wt, fixture.repo, "testrepo-worktree"));
-	git__free(wt->worktree_path);
-	wt->worktree_path = "/path/to/invalid/worktreedir";
-
+	p_rename("testrepo-worktree", "testrepo-worktree-tmp");
 	cl_git_fail(git_worktree_validate(wt));
 
-	wt->worktree_path = NULL;
 	git_worktree_free(wt);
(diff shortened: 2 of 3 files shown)


tests/worktree/worktree.c
```diff
@@ -610,3 +610,17 @@ void test_worktree_worktree__foreach_worktree_lists_all_worktrees(void)
...
 	cl_git_pass(git_repository_foreach_worktree(fixture.repo, foreach_worktree_cb, &counter));
 }
+
+void test_worktree_worktree__validate_invalid_worktreedir(void)
+{
+	git_worktree *wt;
+
+	cl_git_pass(git_worktree_lookup(&wt, fixture.repo, "testrepo-worktree"));
+	git__free(wt->worktree_path);
+	wt->worktree_path = "/path/to/invalid/worktreedir";
```
ethomson, line 620: This feels a little bit like too much magic to me.  😀  Is there a way that we can invalidate the work tree?  If I’m reading this correctly, we can just rm the work tree itself which would serve the same purpose but test the intent more directly and without having to change the innards of the work tree object.

ethomson:
Thanks for the fix and the test!  Wonder if we can tweak the test to be a little more direct but otherwise 👍

rbmclean:
My understanding of clar isn't great, but I didn't want to delete the worktree as I suspected it would cause issues. Now it's renamed and seems to still work as intended. I ran the test suite twice in a row on my machine to make sure it was renamed back correctly.

rbmclean:
Is there anything else that needs to be done for this?

tests/worktree/worktree.c
```diff
@@ -610,3 +610,14 @@ void test_worktree_worktree__foreach_worktree_lists_all_worktrees(void)
...
 	cl_git_pass(git_repository_foreach_worktree(fixture.repo, foreach_worktree_cb, &counter));
 }
+
+void test_worktree_worktree__validate_invalid_worktreedir(void)
+{
+	git_worktree *wt;
+
+	cl_git_pass(git_worktree_lookup(&wt, fixture.repo, "testrepo-worktree"));
+	p_rename("testrepo-worktree", "testrepo-worktree-tmp");
+	cl_git_fail(git_worktree_validate(wt));
```
ethomson, line 620:
```suggestion
	cl_git_fail(git_worktree_validate(wt));
	p_rename("testrepo-worktree-tmp", "testrepo-worktree");
```

src/worktree.c
```diff
@@ -259,7 +259,14 @@ int git_worktree_validate(const git_worktree *wt)
 			wt->commondir_path);
 		return GIT_ERROR;
 	}
-
+	
+	if (!git_path_exists(wt->worktree_path)) {
+		git_error_set(GIT_ERROR_WORKTREE,
+			"worktree directory ('%s') does not exist ",
```
ethomson, line 265:
```suggestion
			"worktree directory '%s' does not exist",
```

ethomson (commented):
This looks good to me - I would move the cleanup rename into the test itself, and omit the trailing space in the error message, but otherwise this looks good.

ethomson:
Since the changes that I'm suggesting are super minor and stylistic, I'm going to go ahead and commit the suggestions that I made.

Thanks again for the PR!

ethomson (approved):

PR #5704, opened 2020-11-23
Also add the raw hostkey to `git_cert_hostkey`

lhchavez:
`git_cert_x509` has the raw encoded certificate. Let's do the same for
the SSH certificate for symmetry.


ethomson:
LGTM.  Thanks!

PR #5705, opened 2020-11-23
[WIP] Use vectored file IO for mmap emulation

0xdky:
`p_pread` and `p_pwrite` can optionally use vectored IO by using underlying `preadv` and `pwritev`.
`mmap` emulation uses `p_pread` and writing pack files use `p_pwrite` (from `write_at`).

Reading objects and writing pack files will not benefit from vectored IO with `mmap` emulation and should help with reduced latency when IO bandwidth is available.

### DIFF
--- a/src/win32/posix_w32.c
+++ b/src/win32/posix_w32.c
@@ -981,3 +981,73 @@ int p_inet_pton(int af, const char *src, void *dst)
 	errno = EINVAL;
 	return -1;
 }
+
+ssize_t p_pread(git_file fd, void *data, size_t size, git_off_t offset)
+{
+	size_t mmap_alignment;
+	size_t page_offset;
+	git_off_t page_start;
+	unsigned char *map_data;
+	git_map map;
+	int error;
+
+	assert(data && size);
+
+	if (!git__is_ssizet(size)) {
+		errno = EINVAL;
+		SetLastError(ERROR_INVALID_PARAMETER);
+		return -1;
+	}
+
+	if ((error = git__mmap_alignment(&mmap_alignment)) < 0)
+		return error;
+
+	/* the offset needs to be at the mmap boundary for the platform */
+	page_offset = offset % mmap_alignment;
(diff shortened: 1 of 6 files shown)


0xdky:
@ethomson The change to add vectored IO I was mentioning. This PR is built on: https://github.com/libgit2/libgit2/pull/5583

src/win32/posix_w32.c
```diff
@@ -981,3 +981,73 @@ int p_inet_pton(int af, const char *src, void *dst)
...
+	}
+
+	if ((error = git__mmap_alignment(&mmap_alignment)) < 0)
+		return error;
+
+	/* the offset needs to be at the mmap boundary for the platform */
+	page_offset = offset % mmap_alignment;
+	page_start = offset - page_offset;
+
+	if ((error = p_mmap(&map, page_offset + size, GIT_PROT_WRITE,
```
ethomson, line 1044: If you try to build with `NO_MMAP` on windows, we'll call `p_pwrite` which calls `mmap`, which seems to violate the user's request and expectations.  Let's not implement `p_pwrite` and `p_pread` on Windows, let's just disable the ability to build with `NO_MMAP` on Windows so that these don't get called and then not provide a `p_pwrite` or `p_pread` implementation in win32 at all.  If somebody wants to fix the `NO_MMAP` situation, they can provide a Windows-specific implementation of `p_pread`/`p_pwrite`.

0xdky: I realized preadv/pwritev does not implement scatter/gather IO. I prefer to rework this PR.

Since this PR is built on top of: https://github.com/libgit2/libgit2/pull/5583, I request @ethomson to please review that instead. I have implemented pread/pwrite for win32 using Win32 overlapped file IO.

0xdky:
Prefer to close this PR and work on io_uring based approach instead on top of https://github.com/libgit2/libgit2/pull/5583

PR #5710, opened 2020-11-24
Make `git_index` hold `conflicts`/`git_merge_file_result`

someniceone:
Make `git_index` hold `conflicts`/`git_merge_file_result`, so we can get `conflicts` directly from `git_index` to preview them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,9 @@ build/
 tags
 CMakeSettings.json
 .vs
+
+.idea/
+.vscode/
+.history/
+Testing/
+cmake-build-debug/
--- a/examples/merge.c
+++ b/examples/merge.c
@@ -182,15 +182,17 @@ static void output_conflicts(git_index *index)
 	const git_index_entry *ancestor;
 	const git_index_entry *our;
 	const git_index_entry *their;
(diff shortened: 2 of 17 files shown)


ethomson:
I'm not sure that I entirely understand your goal here.  `git_index` should be a map to the index data structure on-disk, which doesn't contain this information.  Nor do I think that people should be required to compute file merge results just to get information out of the index.  If you could clarify your intent a little bit, it might help me understand a path forward.

someniceone:
> I'm not sure that I entirely understand your goal here. `git_index` should be a map to the index data structure on-disk, which doesn't contain this information. Nor do I think that people should be required to compute file merge results just to get information out of the index. If you could clarify your intent a little bit, it might help me understand a path forward.

Thank you for comment, in my case I need to preview the `merge conflicts` before real merging, at the same time I found that `git_merge_commits()` produces the conflicts as by-product but dropped later, so I was considering how to pass it out, `git_index` looks a good place to hold the `conflicts`, that's why I did that. 

But you point out `index` should be just a `map`, so it should not hold big data, so is it better that just pass out `conflicts` from function `git_merge_commits()` ? 

Thank you!

PR #5723, opened 2020-12-09
Fix allowed creation of file that already exists by `git_apply_to_tree`

### LINKED ISSUE #5717: `git_apply_to_tree` produces an incorrect result instead of failing
### Reproduction steps

Apply a patch that creates a file to a tree that already contains a file with the same name.

1. Have an existing tree with a file called `file` with content `content-1`

2. Use `git_apply_to_tree` to apply the following patch to it:

```
From 82f0310e7e359109c71f89e7bc6477dad8abcd80 Mon Sep 17 00:00:00 2001
From: Test Author <test.author@example.com>
Date: Thu, 3 Dec 2020 13:36:41 +0100
Subject: [PATCH] conflicting file add

---
 file | 1 +
 1 file changed, 1 insertion(+)
 create mode 100644 file

diff --git a/file b/file
new file mode 100644
index 0000000..df947bd
--- /dev/null
+++ b/file
@@ -0,0 +1 @@
+content-2
\ No newline at end of file
-- 
2.29.1

```

### Expected behavior

Applying the patch should fail. 

Starting state:
```shell
❯ git ls-tree head
100644 blob e5291437448e70f16b5a3fa828834d043b2f30db	file

❯ git cat-file -p e5291437448e70f16b5a3fa828834d043b2f30db 
content-1
```

`git apply`:
```shell
> pbpaste | git apply 
error: file: already exists in working directory
```

`git am`:
```shell
❯ pbpaste | git am
Applying: conflicting file add
error: file: already exists in index
Patch failed at 0001 conflicting file add
hint: Use 'git am --show-current-patch=diff' to see the failed patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".
```

`git am -3`:
```shell
❯ pbpaste | git am -3
Applying: conflicting file add
Using index info to reconstruct a base tree...
Falling back to patching base and 3-way merge...
CONFLICT (add/add): Merge conflict in file
Auto-merging file
error: Failed to merge in the changes.
Patch failed at 0001 conflicting file add
hint: Use 'git am --show-current-patch=diff' to see the failed patch
When you have resolved this problem, run "git am --continue".
If you prefer to skip this patch, run "git am --skip" instead.
To restore the original branch and stop patching, run "git am --abort".

❯ cat file
<<<<<<< HEAD
content-1
=======
content-2
>>>>>>> conflicting file add
```

### Actual behavior

Result is a tree with `file` containing `content-2content-1`.

```shell
❯ git ls-tree head
100644 blob 8305b7a6ccc9bb7dcff4492bbd7fb405777e4c59	file

❯ git cat-file -p 8305b7a6ccc9bb7dcff4492bbd7fb405777e4c59
content-2content-1
```

v30.0.18

macOS 10.15, Debian 10.6

rbmclean:
Fixes #5717

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -562,7 +562,6 @@ static int apply_one(
 		post_entry.path = filename;
 		post_entry.mode = mode;
 		git_oid_cpy(&post_entry.id, &post_id);
-
 		if ((error = git_index_add(postimage, &post_entry)) < 0)
 			goto done;
 	}
@@ -650,7 +649,12 @@ int git_apply_to_tree(
 	 */
 	for (i = 0; i < git_diff_num_deltas(diff); i++) {
 		delta = git_diff_get_delta(diff, i);
-
+		if (delta->status == GIT_DELTA_ADDED &&
+			git_index_find(NULL, postimage, delta->new_file.path) != GIT_ENOTFOUND) {
+				git_error_clear();
+				error = GIT_EEXISTS;
+				goto done;
+		}
 		if (delta->status == GIT_DELTA_DELETED ||
 			delta->status == GIT_DELTA_RENAMED) {
 			if ((error = git_index_remove(postimage,
--- a/tests/apply/apply_helpers.h
+++ b/tests/apply/apply_helpers.h
@@ -68,6 +68,16 @@
 	"+This is a new file!\n" \
 	"+Added by a patch.\n"
 
+#define DIFF_DUPLICATE_FILE \
+	"diff --git a/veal.txt b/veal.txt\n" \
+	"new file mode 100644\n" \
+	"index 0000000..6370543\n" \
+	"--- /dev/null\n" \
+	"+++ b/veal.txt\n" \
+	"@@ -0,0 +1,2 @@\n" \
+	"+This is a duplicate file!\n" \
+	"+It should not be added.\n"
+
 #define DIFF_EXECUTABLE_FILE \
 	"diff --git a/beef.txt b/beef.txt\n" \
 	"old mode 100644\n" \
--- a/tests/apply/tree.c
+++ b/tests/apply/tree.c
@@ -87,6 +87,10 @@ void test_apply_tree__adds_file(void)
 	cl_git_pass(git_apply_to_tree(&index, repo, a_tree, diff, NULL));
 	merge_test_index(index, expected, 7);
 
+	cl_git_pass(git_diff_from_buffer(&diff,
+		DIFF_DUPLICATE_FILE, strlen(DIFF_DUPLICATE_FILE)));
+	cl_git_fail(git_apply_to_tree(&index, repo, a_tree, diff, NULL));
+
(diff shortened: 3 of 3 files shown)


src/apply.c
```diff
@@ -650,7 +649,12 @@ int git_apply_to_tree(
 	 */
 	for (i = 0; i < git_diff_num_deltas(diff); i++) {
 		delta = git_diff_get_delta(diff, i);
-
+		if (delta->status == GIT_DELTA_ADDED &&
```
ethomson, line 652: We probably need to handle the rename case as well, when the `new_file.path` exists in the postimage already.

src/apply.c
```diff
@@ -650,7 +649,12 @@ int git_apply_to_tree(
...
+			git_index_find(NULL, postimage, delta->new_file.path) != GIT_ENOTFOUND) {
+				git_error_clear();
+				error = GIT_EEXISTS;
```
ethomson, line 655: I think that we should check the return code more distinctly - we should only return `GIT_EEXISTS` when `git_index_find` returns 0, otherwise we should propagate its return information.

src/apply.c
```diff
@@ -562,7 +562,6 @@ static int apply_one(
 		post_entry.path = filename;
 		post_entry.mode = mode;
 		git_oid_cpy(&post_entry.id, &post_id);
-
```
ethomson, line 565: Spurious whitespace change here.

ethomson:
Thanks @rbmclean for tackling this!  Just a few minor comments.

rbmclean:
Alright, your suggestions have been added, though I can also add a test case for the `GIT_DELTA_RENAMED` behavior from this patch if you want.

SamiHiltunen:
@rbmclean I've tested this with the code where I identified this and the error is now correctly returned. Thanks a lot for fixing this! 🎉

@ethomson do you think you'd have the time to review this soon again? Looking very much forward to this landing.

SamiHiltunen:
@ethomson wanted to carefully ask about this again. Is this good to go or can I help in anyway with this? Very much looking forward to this landing still. Thanks again!

arxanas:
Hi, is there any update on this? I'm also affected by #5717. Is there anything I can do to help?

It looks like @ethomson is on parental leave. @lhchavez would you be able to merge this, or suggest someone who can?

src/apply.c
arxanas: After testing out this commit, I think this logic is incorrect. This does detect if the diff had a duplicate file, but it also returns `GIT_EEXISTS` for valid index+diff applications.

The postimage is initialized as a copy of the preimage. So if the preimage already contains file `foo`, and the diff also attempts to add file `foo`, it will return `GIT_EEXISTS` here. But we should actually produce an index with conflicts in that case, rather than returning an error.

Issue #5725, opened 2020-12-10
Big performance regression when scanning for workdir changes in repo with submodules


jeroen:
We are using a [metarepo](https://github.com/r-universe/ropensci) that contain submodules for many other projects. Recently we started noticing bad performance after updating to a new distro with a newer version of libgit2.

The regression happens in calls that scan for workdir changes, such as `git_status_list_new()` and `git_index_add_all`.

Below is a small example that illustrates the problem. What we want is the same as `git status` in the command line, to see if any of the submodules has been updated. But even if none of the submodules are actually initialized, this takes 2 sec. The git command line takes less than 0.1 sec for the same thing.

```c
#include <git2.h>
#include <stdio.h>

int main(){
	git_libgit2_init();
	git_repository *repo = NULL;
	if(git_repository_open(&repo, "ropensci")){
		fprintf(stderr, "Initial clone...");
		git_clone(&repo, "https://github.com/r-universe/ropensci", "ropensci", NULL);
		fprintf(stderr, "OK\n");
	}
	git_status_options opts = GIT_STATUS_OPTIONS_INIT;
	opts.flags = GIT_STATUS_OPT_INCLUDE_UNTRACKED |
		GIT_STATUS_OPT_RENAMES_HEAD_TO_INDEX |
		GIT_STATUS_OPT_SORT_CASE_SENSITIVELY;
	git_status_list *list = NULL;
	fprintf(stderr, "Running git_status_list_new...");
	git_status_list_new(&list, repo, &opts);
	fprintf(stderr, "done!\n");
	size_t len = git_status_list_entrycount(list);
	fprintf(stderr, "Found %d changes!\n", (int) len);
	return 0;
}
```

ethomson:
Do you know the prior version that was less slow?  Do you have an idea of what that perf looked like?

jeroen:
Looking into my past GHA logs now... performance for this call has actually always been bad, but it got 2x slower somewhere in between libgit2 0.28.4 and 1.0.0.

I have no idea what it can possibly be doing that is taking so long, if none of these submodules are even initialized?

lhchavez:
if you call `git_repository_submodule_cache_all(repo);` (needs a `#include <git2/sys/repository.h>`) just before `git_status_list_new()`, it no longer needs to parse the `.gitmodules` file for every single submodule, making it as fast (or faster?) as shelling out.

jeroen:
[...]

PR #5727, opened 2020-12-11
Cache the parsed submodule config when diffing

lhchavez:
This change makes that anything that calls `git_diff__from_iterators`
(any of the `git_diff_xxx` functions) only need to parse the
`.gitmodules` file once. This can be avoided by calling
`git_repository_submodule_cache_all(...)`, but we can do that safely for
the user with no change in semantics.

Fixes: #5725

### DIFF
--- a/src/diff_generate.c
+++ b/src/diff_generate.c
@@ -680,6 +680,8 @@ typedef struct {
 	git_iterator *new_iter;
 	const git_index_entry *oitem;
 	const git_index_entry *nitem;
+	git_strmap *submodule_cache;
+	bool submodule_cache_initialized;
 } diff_in_progress;
 
 #define MODE_BITS_MASK 0000777
@@ -694,6 +696,7 @@ static int maybe_modified_submodule(
 	git_submodule *sub;
 	unsigned int sm_status = 0;
 	git_submodule_ignore_t ign = diff->base.opts.ignore_submodules;
+	git_strmap *submodule_cache = NULL;
 
 	*status = GIT_DELTA_UNMODIFIED;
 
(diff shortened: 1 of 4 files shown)


src/diff_generate.c
```diff
@@ -694,15 +696,37 @@ static int maybe_modified_submodule(
...
 		ign == GIT_SUBMODULE_IGNORE_ALL)
 		return 0;
 
-	if ((error = git_submodule_lookup(
-			&sub, diff->base.repo, info->nitem->path)) < 0) {
+	if (diff->base.repo->submodule_cache != NULL) {
+		submodule_cache = diff->base.repo->submodule_cache;
+	} else {
+		if (!info->submodule_cache_initialized) {
+			info->submodule_cache_initialized = true;
```
ethomson, line 711: Can we maybe pull this into a helper function?  `git_submodule_cache_init(git_strmap **out)`?  It would actually dovetail nicely with a `git_submodule_cache_free` that `git_diff__from_iterators` could call.

If you really wanted to be fancy, we could create a `git_submodule_cache` struct that has a refcount so that you could create a new one or borrow the repository's and we could drop the `submodule_cache_initialized` out of the diff struct.  But this is probably more work for limited gain.

ethomson:
Minor nit requesting some refactoring, but overall, this is clearly the right thing to be doing.  Thanks for digging in and fixing it.

src/diff_generate.c
lhchavez: sounds good!

i ended up not adding the refcount (for now) since there are some cases where the upfront cache generation fails (see https://github.com/libgit2/libgit2/runs/1535037963?check_suite_focus=true#step:7:1289 for an example), but looking up the submodules individually succeeds. so i still want to have that explicit `submodule_cache_initialized` flag to allow for this case.

ethomson:
👍

jeroen:
Thank you for digging into this!

PR #5728, opened 2020-12-11
blame: add blame progress callback

KOLANICH:


include/git2/blame.h
```diff
@@ -51,6 +51,19 @@ typedef enum {
...
+/**
+ * Blame progress callback.
+ *
+ * Called for each blame suspect.
+ *
+ * - 'suspect' is the id of the suspect commit.
+ *
+ * Returning a non-zero value will abort the blame.
+ */
+typedef int (*git_blame_progress_cb)(
```
Uncommon, line 63: "Progress" seems like a misnomer to me. Since it only provides the commit ID, that doesn't tell you how far through the process it is. It's really a mechanism for aborting, but I'm not quite clear on the use case.

KOLANICH: The commits are not mine, they are by @hackhaslam. So, if you are ready to merge them, then merge, if you are not and wanna discuss something, discuss with him, since he already uses this in `gitahead` and making an incompatible non-fast-forwardable change will only cause more mess. I just try to upstream them in order to make using libgit2 package from a distro instead of a custom fork a bit closer.

hackhaslam: The patches in the GitAhead fork of libgit2 are *not* ready to be upstreamed. I've already proposed patches for the changes that I believed to be acceptable. I understand your desire to build GitAhead from a vanilla libgit2, but that's not a realistic goal at this point.

PR #5729, opened 2020-12-11
hash: changed git_repository_hashfile to hash symlinks the same as when they are added to the index

KOLANICH:
hash: changed git_repository_hashfile to hash symlinks the same as when they are added to the index

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -2675,6 +2675,7 @@ int git_repository_hashfile(
 	git_file fd = -1;
 	uint64_t len;
 	git_buf full_path = GIT_BUF_INIT;
+	struct stat st;
 
 	 /* as_path can be NULL */
 	GIT_ASSERT_ARG(out);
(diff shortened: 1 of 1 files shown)


src/repository.c
```diff
@@ -2691,6 +2692,11 @@ int git_repository_hashfile(
 	if (error < 0)
 		return error;
 
+	if (!p_lstat(full_path.ptr, &st) && S_ISLNK(st.st_mode)) {
```
ethomson, line 2695: We're already doing a `stat` to get the file size (below).  Can we not just have a single stat for both purposes?

ethomson:
This needs tests.  I see that you rebased all your PRs?  Should I be providing feedback on them or not?  A prior message suggested that you weren't planning on responding to feedback since you didn't feel that these were "your" changes.

KOLANICH:
>A prior message suggested that you weren't planning on responding to feedback since you didn't feel that these were "your" changes.

Currently I am not going to change the code in a way that can cause rebase conflicts to gitahead author, and especially am not going to change the behavior. It is his patches, he has written them and he depends on them. I cannot telepathically know all the context he has in his head.

About writing tests - I have no plans currently about it.

ethomson:
Then I'm going to close these pull requests.  There's little use to having them open if there's no hope of them getting merged.

PR #5730, opened 2020-12-11
remote: add callbacks for connect/disconnect

KOLANICH:


ethomson:
This is one of seven pull requests you just opened without a description, or useful comments in the commit messages.  Also without tests.

These seem like they may be useful changes, but I'd like to understand why you're proposing these changes - both in the pull request message and the commit message - and to see test coverage for these changes.

KOLANICH:
TBH, it is not mine changes, it is changes by the author of gitahead (a promising GUI git client) @hackhashlam . I just tried to replace inlined libgit2 by the lib from the distro, and found it is impossible, since they have own patches. So I selected the ones looking upstreamable and sent PRs here.

hackhaslam:
The patches in the GitAhead fork of libgit2 are not ready to be upstreamed. I've already proposed patches for the changes that I believed to be acceptable. I understand your desire to build GitAhead from a vanilla libgit2, but that's not a realistic goal at this point.

hackhaslam:
@KOLANICH If you want to take them one at a time and start a discussion, I'm certainly open to join in the discussion and give the reasons for why I added these patches in the first place. Dumping them all like this is not likely to be productive.

KOLANICH:
@hackhaslam, this commit seems to have very little to discuss, since the code is very straightforward. Though I see 1 thing that may need to be discussed: we call different callbacks for connecting and disconnecting. I propose to merge them into 1, called with an additional argument - a pointer to a C string, which can have some values from a discrrete set. So, the callback can identify the event using `switch` on a pointer value (without using string functions), or can show a progress message to a user in a generic way.

hackhaslam:
@KOLANICH I think the more salient questions that need to be discussed are, what is this? What is it for? What are the use cases? Why do we think it's necessary?

In this particular case, I added the connect/disconnect callbacks because I needed to remember the `remote` to cancel the transfer later on. This is most important in submodule update, which looks up the remotes internally. Without this, there isn't a way to cancel the transfers initiated by submodule update.

However, as I said, and to @ethomson's point, I don't think these patches are ready to be submitted upstream. They need to be fleshed out. They need tests and documentation changes.

KOLANICH:
>In this particular case, I added the connect/disconnect callbacks because I needed to remember the remote to cancel the transfer later on.

Thanks for the info. I haven't dug deep into gitahead, so misassumed their use. I wonder if using callbacks solely for tracking state is a good idea at all. Can the state be tracked in libgit itself?

KOLANICH:
@Murmele

PR #5732, opened 2020-12-11
checkout: notify for deleted files

KOLANICH:


Murmele:
what is the current status of it?

KOLANICH:
IDK. This patch was a part of the fork of libgit used in GitAhead git GUI. Now GitAhead is dead and its alive fork is GittyUp. It is likely it also contains this patch. But having own forks of libs within apps is a very unhealthy situation. It must not be done this way. So I tried to upstream the patches in the hope to make the moment when the tools use the same lib the whole system uses closer.

Murmele:
I am one of the maintainer of Gittyup, I try to get rid of all of those patches and try to update libgit version to the most current one

KOLANICH:
@Murmele, I guess I should close this PR because I have neither time to work on it nor experience with libgit2. So probably the most optimal solution for you is to create an own one.

Murmele:
@KOLANICH yes I already started with it

PR #5734, opened 2020-12-11
patch: added flag to disable mmap

KOLANICH:


ethomson:
I’m not sure that I see the utility here - I feel like building without mmap support would be the logical solution to this problem; why finer grained mechanisms to disable mmap and why only here?  Going to close this but happy to revisit if there’s a real useful reason for this.

PR #5735, opened 2020-12-13
Make git__strntol64() ~70%* faster

lhchavez:
This change uses compiler intrinsics to detect overflows instead of
using divisions to detect potential overflow. This makes the code faster
and makes it easier to read as a bonus side-effect!

Some of the things this quickens:

* Config parsing.
* Tree parsing.
* Smart protocol negotiation.

\* Measured by running `libgit2_clar` with `-fno-optimize-sibling-calls
-fno-omit-frame-pointer` under `perf(1)`:

```shell
$ perf diff --symbols=git__strntol64 --compute=ratio \
  --percentage=absolute baseline.data perf.data
# Event 'cycles'
#
# Baseline           Ratio  Shared Object
# ........  ..............  .............
#
     0.25%        0.321836  libgit2_clar
```

### DIFF
--- a/src/integer.h
+++ b/src/integer.h
@@ -77,6 +77,11 @@ GIT_INLINE(int) git__is_int(long long p)
 # define git__sub_int_overflow(out, one, two) \
     __builtin_ssub_overflow(one, two, out)
 
+# define git__add_int64_overflow(out, one, two) \
+    __builtin_add_overflow(one, two, out)
+# define git__multiply_int64_overflow(out, one, two) \
+    __builtin_mul_overflow(one, two, out)
+
 /* Use Microsoft's safe integer handling functions where available */
 #elif defined(_MSC_VER)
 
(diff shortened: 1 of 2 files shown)


ethomson:
Cool.  Thanks!

ethomson:
Regrettably, some versions of clang/compiler-rt/libgcc/<insert toolchain interaction here> lack `__builtin_mul_overflow`.

https://bugs.llvm.org/show_bug.cgi?id=16404 is the upstream bug - I think - but there's lots of finger waggling it seems.

https://github.com/android/ndk/issues/294 and https://github.com/android/ndk/issues/506 are helpful references as well.

Here's our test failures in nightly: https://github.com/libgit2/libgit2/runs/1574212900?check_suite_focus=true

Any chance that you would be interested in taking a look at this?

lhchavez:
hahaha, as it turns out, i've encountered this in the past u___u https://github.com/aosp-mirror/platform_system_core/commit/fae195b1e45ffcf2ba14bd92faf89f06c5cc4be7

w.c.s. i can add more `#ifdef`s to use the fallback in clang+i686.

PR #5741, opened 2020-12-17
Handle ipv6 addresses

ethomson:
Supersedes #5155


ethomson:
Regrettable that we don't have IPv6 on the build agents, but I've done some local testing on Unix and Windows.  Very possible that there remain issues, but I'm going to merge as-is.

PR #5742, opened 2020-12-18
Avoid using `__builtin_mul_overflow` with the clang+32-bit combo

lhchavez:
This causes clang to produce an undefined reference to `__mulodi4`. This
could be fixed by statically linking some compiler-rt libraries to
provide this symbol, but let's first stop the bleeding since doing the
correct long-term fix requires some non-trivial CMake knowledge which I
lack.

### DIFF
--- a/src/integer.h
+++ b/src/integer.h
@@ -79,8 +79,12 @@ GIT_INLINE(int) git__is_int(long long p)
 
 # define git__add_int64_overflow(out, one, two) \
     __builtin_add_overflow(one, two, out)
-# define git__multiply_int64_overflow(out, one, two) \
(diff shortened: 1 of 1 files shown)


src/integer.h
```diff
@@ -156,6 +160,10 @@ GIT_INLINE(bool) git__add_int64_overflow(int64_t *out, int64_t one, int64_t two)
 	return false;
 }
 
+#endif
+
+/* If we could not provide an intrinsic implementation for this, provide a (slow) fallback. */
+#if !defined(git__multiply_int64_overflow)
```
ethomson, line 166: Clever.  I like it.

ethomson:
Thanks @lhchavez!

PR #5743, opened 2020-12-19
Third attempt to fix the 32-bit version of `git__multiply_int64_overf…

lhchavez:
…low`

This change should now fix the issue for realsies. It should still be faster
than the original implementation since it only does one integer division
instead of two.

`./libgit2_clar -score::strtol` passes on a 32-bit Docker.


ethomson:
Thanks!

PR #5744, opened 2020-12-19
Add tests for `git__multiply_int64_overflow`

lhchavez:
As it turns out, the implementation of `git__multiply_int64_overflow` is
full of edge cases and tricky arithmetic. That means that it should have
unit tests, which only run when the fallback implementation is used to avoid
requiring all builds to support 128-bit arithmetic.

As a result, a bug in `git__strntol64` was found (and fixed!) in
clang+32-bit.


ethomson:
Holy cow.  Thank you for digging in on this.

PR #5747, opened 2020-12-20
Add tests for atomic-related functions

lhchavez:
There were some subtle semantic differences between the various
implementations of atomic functions. Now they have tests and are better
documented to avoid this from happening again in the future.


ethomson:
This looks like it's adding more than just tests - not sure if those snuck in during the rebase?  Either way it might be nice to break those out into their own commit.

lhchavez:
> This looks like it's adding more than just tests - not sure if those snuck in during the rebase? Either way it might be nice to break those out into their own commit.

oh, haha, yeah. i originally started writing just tests and then found about the actual failures and fixed them. let me rename the commit to better describe the intent of this PR.

PR #5749, opened 2020-12-21
Fix the `-DENABLE_WERROR=ON` build for gcc 10.2

lhchavez:
This change makes it possible to build with newer versions of gcc
without warnings. There were two warnings issued:

* gcc 8 added
  [`-Wstringop-truncation`](https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8/),
  which warns if a call to `strncpy(3)` is prone to accidentally
  truncating the destination string, since `strncpy(3)` does NOT add a
  terminating `NULL` if the destination buffer is not large enough to
  hold the input.

  This change uses the pattern suggested in
  https://us-cert.cisa.gov/bsi/articles/knowledge/coding-practices/strncpy-and-strncat
  to fix the locations flagged by gcc.
* There was a potentially uninitialized access of `dest` in `fs_copy`.

### DIFF
--- a/tests/clar/fs.h
+++ b/tests/clar/fs.h
@@ -396,7 +396,7 @@ static void
 fs_copy(const char *source, const char *_dest)
 {
 	char *dbuf = NULL;
-	const char *dest;
+	const char *dest = NULL;
 	struct stat source_st, dest_st;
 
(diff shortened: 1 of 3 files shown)


ethomson:
Cool, thanks for this.  I also brought them over to [clar's upstream](https://github.com/clar-test/clar/pull/90)

PR #5750, opened 2020-12-22
Add support for additional SSH hostkey types.

arroz:
The raw hostkey in the certificate is incredibly useful to subsequently compare and add it to a known_hosts file. However, the current code only support RSA and DSS typed keys. LibSSH2 supports a few more, including ED25519 which is the default hostkey type for many systems (macOS, for example, and I believe Raspbian as well).

This patch adds support for ECDSA_256, ECDSA_384, ECDSA_521 and ED25519 hostkey types on top of the already existing ones.

### DIFF
--- a/src/transports/ssh.c
+++ b/src/transports/ssh.c
@@ -580,6 +580,8 @@ post_extract:
 				case LIBSSH2_HOSTKEY_TYPE_DSS:
 					cert.raw_type = GIT_CERT_SSH_RAW_TYPE_DSS;
 					break;
+					
+#ifdef LIBSSH2_HOSTKEY_TYPE_ECDSA_256
 				case LIBSSH2_HOSTKEY_TYPE_ECDSA_256:
 					cert.raw_type = GIT_CERT_SSH_RAW_TYPE_KEY_ECDSA_256;
 					break;
@@ -589,9 +591,13 @@ post_extract:
 				case LIBSSH2_KNOWNHOST_KEY_ECDSA_521:
 					cert.raw_type = GIT_CERT_SSH_RAW_TYPE_KEY_ECDSA_521;
 					break;
+#endif
+					
+#ifdef LIBSSH2_HOSTKEY_TYPE_ED25519
 				case LIBSSH2_HOSTKEY_TYPE_ED25519:
 					cert.raw_type = GIT_CERT_SSH_RAW_TYPE_KEY_ED25519;
 					break;
+#endif
 				default:
 					cert.raw_type = GIT_CERT_SSH_RAW_TYPE_UNKNOWN;
 			}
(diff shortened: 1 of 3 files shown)


arroz:
Oops… what is the libssh2 version the build servers and running with?

arroz:
Ah I see it's 1.8.2, not 1.9.0 that already supports the new key types. What's the best approach here? Wrapping the code in a few `#ifdef`s, or upgrading the minimum version of libssh2 required?

lhchavez:
> Ah I see it's 1.8.2, not 1.9.0 that already supports the new key types. What's the best approach here? Wrapping the code in a few `#ifdef`s, or upgrading the minimum version of libssh2 required?

probably both. add the `#ifdef`s to the code and tweak the version in the [focal container](https://github.com/libgit2/libgit2/blob/f4b473fa842a84147d8036a27bac5c026c44dad4/ci/docker/focal#L51) so that it has _some_ coverage, while leaving the version of the xenial container untouched.

FWIW ubuntu is still on [1.8.0](https://packages.ubuntu.com/groovy/libssh2-1) :(

arroz:
Sounds good, I'll fix the patch tonight or tomorrow morning. Thanks for the quick reply!

arroz:
CI is happy! 😊 Let me know if any other improvements are needed.

src/transports/ssh.c
```diff
@@ -580,6 +580,24 @@ static int _git_ssh_setup_conn(
...
 					break;
+					
+#ifdef LIBSSH2_HOSTKEY_TYPE_ECDSA_256
+				case LIBSSH2_HOSTKEY_TYPE_ECDSA_256:
+					cert.raw_type = GIT_CERT_SSH_RAW_TYPE_KEY_ECDSA_256;
+					break;
+				case LIBSSH2_HOSTKEY_TYPE_ECDSA_384:
+					cert.raw_type = GIT_CERT_SSH_RAW_TYPE_KEY_ECDSA_384;
+					break;
+				case LIBSSH2_KNOWNHOST_KEY_ECDSA_521:
```
lhchavez, line 591: is it possible for only a subset of these to be defined? (say, only 256 but not 384 or 521). or were all of these constants introduced in the same commit?

lhchavez (approved):
:+1: from my side too, with an optional comment.

src/transports/ssh.c
arroz: Yeah, 256, 384 and 521 were all introduced at the same time in 62b825c8.
ED25519 was added a bit later with 030922925.

ethomson:
Thanks @arroz - and I'll second those congratulations on your first PR.  Very cool!

Thanks @lhchavez for the review.

ethomson:
On an unrelated note:

Screen Shot 2020-12-23 at 12 18 15

I gave you write privileges @lhchavez, which was frankly, long overdue.  Thanks for all the help lately.

PR #5759, opened 2021-01-04
Add documentation for git_blob_filter_options.version

### LINKED ISSUE #5756: Lack of documentation for "version" member of struct "git_blob_filter_options"
### Expected behavior
Detailed usage documentation describing how to value this property when passing options to `git_blob_filter`.

### Actual behavior
No usage documentation at all.

HEAD (fe41e5823986aff01054b7db4b869b2896ea4c30)

---

Definition:
https://github.com/libgit2/libgit2/blob/fe41e5823986aff01054b7db4b869b2896ea4c30/include/git2/blob.h#L119-L127

---

I have no idea what to put for `version` when passing options to `git_blob_filter`. As a result, I still use `git_blob_filtered_content` despite documentation saying it's deprecated.

JoshuaS3:
"The struct version; set to `GIT_BLOB_FILTER_OPTIONS_VERSION`."

Resolves #5756.

### DIFF
--- a/include/git2/blob.h
+++ b/include/git2/blob.h
@@ -120,6 +120,7 @@ typedef enum {
  * The options used when applying filter options to a file.
  */
 typedef struct {
+	/** The struct version; set to `GIT_BLOB_FILTER_OPTIONS_VERSION`. */
 	int version;
 
 	/** Flags to control the filtering process, see `git_blob_filter_flag_t` above */
--- a/include/git2/blob.h
+++ b/include/git2/blob.h
@@ -118,9 +118,11 @@ typedef enum {
 
 /**
  * The options used when applying filter options to a file.
+ *
+ * Initialize with `GIT_BLOB_FILTER_OPTIONS_VERSION`.
+ *
  */
 typedef struct {
-	/** The struct version; set to `GIT_BLOB_FILTER_OPTIONS_VERSION`. */
 	int version;
 
 	/** Flags to control the filtering process, see `git_blob_filter_flag_t` above */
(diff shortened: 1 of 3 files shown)


include/git2/blob.h
```diff
@@ -120,6 +120,7 @@ typedef enum {
  * The options used when applying filter options to a file.
```
lhchavez, line 120: Can this have the blurb that's found in other structs like https://github.com/libgit2/libgit2/blob/fe41e5823986aff01054b7db4b869b2896ea4c30/include/git2/clone.h#L96-L104 to use the better-supported initialization patterns?

JoshuaS3: Updated documentation in new commit.

> `git_blob_filter_options_init` still needed; should be added in another
commit/PR (it's out of the scope of this PR, #5759), update this
documentation again.

include/git2/blob.h
```diff
@@ -118,6 +118,9 @@ typedef enum {
 
 /**
  * The options used when applying filter options to a file.
+ *
+ * Initialize with `GIT_BLOB_FILTER_OPTIONS_VERSION`.
```
lhchavez, line 122:
```suggestion
 * Initialize with `GIT_BLOB_FILTER_OPTIONS_INIT`.
```

lhchavez (approved):
one last thing and :+1:.

include/git2/blob.h
ethomson: I agree with @lhchavez, we should fix this while we're here.  `git_blob_filter_options_init` is in scope.  I pushed up a commit that added it.  Would you mind revising the doc block to refer to it?

ethomson:
🎉  Thanks for this great PR @JoshuaS3 !

Issue #5757, opened 2021-01-04
support git 2.19's commit-graph


lhchavez:
Starting from version 2.19, git supports [`commit-graph`](https://github.blog/2018-09-10-highlights-from-git-2-19/#user-content-commit-graphs), which allows for some significant improvements in the performance of [several graph-based operations](https://devblogs.microsoft.com/devops/supercharging-the-git-commit-graph-iii-generations/). Notably computing the merge base between two (or more) commits and determining whether one commit is ahead/behind of another.

This should be relatively straightforward to implement. Similar to #5399, this can be split into several chunks for ease of reviewing:

- [ ] Support for parsing existing `commit-graph` files and chains, and a fuzzer because parsing is hard.
- [ ] Support for reading the commit metadata and generation numbers from existing `commit-graph` files and chains. (in other words, this will ignore any path-based bloom filters).
- [ ] Modify `paint_down_to_common()` to take advantage of the `commit-graph` metadata, if available.
- [ ] Modify the functions in `src/graph.c` to take advantage of the `commit-graph` metadata, if available.
- [ ] Support for creating a `commit-graph` file from an open repository.
- [ ] Support for creating `commit-graph` chains, so that the cost of updating a `commit-graph` is reduced to only the commits that have changed since the last file / chain was written.

Relevant documentation:

- https://github.com/git/git/blob/master/Documentation/technical/commit-graph.txt
- https://github.com/git/git/blob/master/Documentation/technical/commit-graph-format.txt

PR #5763, opened 2021-01-07
commit-graph: Support lookups of entries in a commit-graph

lhchavez:
This change introduces `git_commit_graph_entry_find()` and
`git_commit_graph_entry_parent()`. These two functions allow a much
faster lookup of commits by ID, since the ODB does not need to be
consulted, the commit object does not need to be inflated, and the
contents of the commit object do not need to be parsed.

Part of: #5757


ethomson:
Woo, commit graphs, very exciting!  🎉

PR #5764, opened 2021-01-07
commit-graph: Introduce `git_commit_graph_needs_refresh()`

lhchavez:
This change introduces a function that allows the caller to know whether
the `commit-graph` file has not been modified since it was parsed.

Part of: #5757


ethomson:
🎉

PR #5765, opened 2021-01-07
commit-graph: Use the commit-graph in revwalks

lhchavez:
This change makes revwalks a bit faster by using the `commit-graph` file
(if present). This is thanks to the `commit-graph` allow much faster
parsing of the commit information by requiring near-zero I/O (aside from
reading a few dozen bytes off of a `mmap(2)`-ed file) for each commit,
instead of having to read the ODB, inflate the commit, and parse it.

Part of: #5757

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -465,6 +465,13 @@ int git_odb_new(git_odb **out)
 		git__free(db);
 		return -1;
 	}
+	if (git_buf_init(&db->objects_dir, 0) < 0) {
+		git_vector_free(&db->backends);
+		git_cache_dispose(&db->own_cache);
+		git_mutex_free(&db->lock);
+		git__free(db);
+		return -1;
+	}
 
 	*out = db;
 	GIT_REFCOUNT_INC(db);
@@ -612,6 +619,17 @@ int git_odb__add_default_backends(
 	git_mutex_unlock(&db->lock);
 #endif
 
+	if (git_mutex_lock(&db->lock) < 0) {
+		git_error_set(GIT_ERROR_ODB, "failed to acquire the odb lock");
+		return -1;
+	}
+	if (git_buf_len(&db->objects_dir) == 0 && git_buf_sets(&db->objects_dir, objects_dir) < 0) {
+		git_mutex_unlock(&db->lock);
+		git_odb_free(db);
+		return -1;
+	}
+	git_mutex_unlock(&db->lock);
+
 	/* add the loose object backend */
 	if (git_odb_backend_loose(&loose, objects_dir, -1, db->do_fsync, 0, 0) < 0 ||
 		add_backend_internal(db, loose, GIT_LOOSE_PRIORITY, as_alternates, inode) < 0)
@@ -742,6 +760,8 @@ static void odb_free(git_odb *db)
 	if (locked)
 		git_mutex_unlock(&db->lock);
 
+	git_buf_dispose(&db->objects_dir);
+	git_commit_graph_free(db->cgraph);
 	git_vector_free(&db->backends);
 	git_cache_dispose(&db->own_cache);
 	git_mutex_free(&db->lock);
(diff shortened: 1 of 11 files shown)


src/odb.h
```diff
@@ -127,6 +131,13 @@ int git_odb__read_header_or_object(
 	git_odb_object **out, size_t *len_p, git_object_t *type_p,
 	git_odb *db, const git_oid *id);
 
+/*
+ * Attempt to get the ODB's commit graph. This object is still owned by the
+ * ODB. If the repository does not contain a commit graph, it will return zero
+ * and `*out` will be set to NULL.
+ */
+int git_odb__get_commit_graph(git_commit_graph_file **out, git_odb *odb);
```
ethomson, line 139: 🤔  I think that returning ENOTFOUND would be a little more idiomatic in this case.

src/odb.c
```diff
@@ -612,6 +619,17 @@ int git_odb__add_default_backends(
 	git_mutex_unlock(&db->lock);
 #endif
 
+	if (git_mutex_lock(&db->lock) < 0) {
+		git_error_set(GIT_ERROR_ODB, "failed to acquire the odb lock");
+		return -1;
+	}
+	if (git_buf_len(&db->objects_dir) == 0 && git_buf_sets(&db->objects_dir, objects_dir) < 0) {
```
ethomson, line 626: I don't love this interaction between the ODB and the commit graph...  It feels a little magical that when you're using the default backends, we store the object directory and a commit graph springs into being, but that's the only opportunity to use a commit graph.   If I were adding my own backends, I would also want to be able to add a commit graph.  This is notable for users like LibGit2Sharp which -- and I don't actually think that this was a good idea, but here we are -- do the moral equivalent of adding the default backends themselves.  🙃 

Naively - and this is coming from a place where you know a lot more about this than I do, so these may be untenable suggestions, but - I would sort of expect:

- the odb can have multiple commit graphs
- users can add a commit graph (`git_odb_add_commit_graph`) much like they can add a backend.
- `git_odb__add_default_backends` adds the commit graph for `objects_dir`

Does multiple commit graphs make sense here?  Obviously it will complicate some of the logic, but unreasonably so?

lhchavez: i can't think of a compelling use case to have multiple commit graphs, since a commit graph is a singleton entity of the repository. furthermore, it would complicate things conceptually, because the information from the graphs _could_ be inconsistent among them. say, if one of the commit-graph files goes against the recommendation of not creating it from a shallow repository and the other one does not (arguably, this is the user shooting themselves in the foot). this is opposed to multiple backends, since conceptually one backend cannot contradict another (in theory at least. they can all "lie" and return a different object if they want to). and even repositories with multiple alternates use a single commit graph! https://git-scm.com/docs/commit-graph#_chains_across_multiple_object_directories

if we ignore the generation number contradiction problem outlined above, from the implementation POV, handling multiple commit graphs is not _unreasonably_ complicated: change the singleton commit graph object to a vector of them. at query time, iterate through all of them and arbitrarily choose one entry (out of the commit graphs that contain that one commit).

so with all that in mind, wdyt about adding a restriction of having at most one commit graph? i'm behind explicitly setting one into the odb (i didn't know about that LibGit2Sharp restriction!), so i would be following your expectations with one tweak: `git_odb_set_commit_graph` instead of `git_odb_add_commit_graph`.

lhchavez: i just realized something else, hahaha. in order to avoid paying for what you don't use, it might be good to introduce a layer of indirection: if folks manually add a commit graph, that commit graph will be eagerly loaded AND error checked (to get a better feedback loop). but if the commit graph is implicitly added by `git_odb__add_default_backends`, it won't be loaded upfront, but until the first time it's needed (kinda how it is right now).

this won't change the interface from what the expectations are, it's just going to be a little bit more code.

ethomson: > so with all that in mind, wdyt about adding a restriction of having at most one commit graph? i'm behind explicitly setting one into the odb (i didn't know about that LibGit2Sharp restriction!), so i would be following your expectations with one tweak: git_odb_set_commit_graph instead of git_odb_add_commit_graph.

:+1:

I was getting a little architecture astronauty, perhaps.  😁  Especially since it's all very theoretical that anybody would want a non-default commit-graph at the moment.  I think that your recommendation makes sense!

ethomson:
🎉  Thanks @lhchavez!

PR #5767, opened 2021-01-07
graph: Create `git_graph_reachable_from_any()`

lhchavez:
This change introduces a new API function
`git_graph_reachable_from_any()`, that answers the question whether a
commit is reachable from any of the provided commits through following
parent edges.

This function can take advantage of optimizations provided by the
existence of a `commit-graph` file, since it makes it faster to know
whether, given two commits X and Y, X cannot possibly be an reachable
from Y.

Part of: #5757

### DIFF
--- a/include/git2/graph.h
+++ b/include/git2/graph.h
@@ -68,8 +68,8 @@ GIT_EXTERN(int) git_graph_descendant_of(
 GIT_EXTERN(int) git_graph_reachable_from_any(
 	git_repository *repo,
 	const git_oid *commit,
-	size_t length,
-	const git_oid descendant_array[]);
+	const git_oid descendant_array[],
+	size_t length);
 
 /** @} */
 GIT_END_DECL
--- a/include/git2/graph.h
+++ b/include/git2/graph.h
@@ -43,8 +43,9 @@ GIT_EXTERN(int) git_graph_ahead_behind(size_t *ahead, size_t *behind, git_reposi
  * Note that a commit is not considered a descendant of itself, in contrast
  * to `git merge-base --is-ancestor`.
  *
- * @param commit a previously loaded commit.
- * @param ancestor a potential ancestor commit.
+ * @param repo the repository where the commits exist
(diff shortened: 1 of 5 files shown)


include/git2/graph.h
```diff
@@ -53,6 +54,23 @@ GIT_EXTERN(int) git_graph_descendant_of(
...
+ * @param length the number of commits in the provided `descendant_array`
+ * @param descendant_array oids of the commits
+ * @return 1 if the given commit is an ancestor of any of the given potential
+ * descendants, 0 if not, error code otherwise.
+ */
+GIT_EXTERN(int) git_graph_reachable_from_any(
+	git_repository *repo,
+	const git_oid *commit,
+	size_t length,
+	const git_oid descendant_array[]);
```
ethomson, line 72:
```suggestion
	const git_oid descendant_array[],
	size_t length);
```

src/graph.c
```diff
@@ -176,19 +176,74 @@ int git_graph_ahead_behind(size_t *ahead, size_t *behind, git_repository *repo,
...
-	/* No merge-base found, it's not a descendant */
-	if (error == GIT_ENOTFOUND)
+	return git_graph_reachable_from_any(repo, ancestor, 1, commit);
+}
+
+int git_graph_reachable_from_any(
+		git_repository *repo,
+		const git_oid *commit_id,
+		size_t length,
+		const git_oid descendant_array[])
```
ethomson, line 189:
```suggestion
		const git_oid descendant_array[],
		size_t length)		
```

ethomson (dismissed):
We are not perfectly consistent but we usually do `array, len` instead of `len, array`.  (in fact, it's `git_merge_base` that is the outlier, 😢).  But I'd like to keep new functionality using `array, len`

include/git2/graph.h
lhchavez: Done!

ethomson (approved):
Thanks!

PR #5768, opened 2021-01-07
midx: Fix a bug in `git_midx_needs_refresh()`

lhchavez:
The very last check in the freshness check for the midx was wrong ><
This was also because this function was not tested.


ethomson:
👍

PR #5770, opened 2021-01-07
Cope with empty default branch

### LINKED ISSUE #5761: assertion/crash with empty defaultBranch
### Reproduction steps

In `~/.gitconfig` set the following:

```
[init]
defaultBranch=""
```

And run the following which initializes a repository and fetches into it with a specific refspec mapping:

```c
#include <assert.h>
#include <git2/remote.h>
#include <git2/repository.h>
#include <git2/errors.h>
#include <git2/global.h>
#include <stdio.h>

int main() {
  assert(git_libgit2_init() == 1);

  git_repository *repo;
  assert(git_repository_init(&repo, "repo", 0) == 0);

  git_remote *remote;
  assert(git_remote_create_anonymous(&remote, repo, "https://github.com/octocat/Hello-World") == 0);

  git_fetch_options fetch_opts = GIT_FETCH_OPTIONS_INIT;

  char *strings[] = {"refs/heads/*:refs/remotes/origin/*"};
  git_strarray refspecs;
  refspecs.strings = strings;
  refspecs.count = 1;

  int err = git_remote_fetch(remote, &refspecs, &fetch_opts, NULL);
  printf("err = %d\n", err);
  if (err == 0)
    return 0;
  const git_error *last = git_error_last();
  if (last != NULL) {
    printf("message: %s\n", last->message);
    printf("klass: %d\n", last->klass);
  }

  return 0;
}
```

(Note: I don't know why someone would end up with an empty defaultBranch, but I've had two independent user reports of ending up with that config.)

### Expected behavior

Should successfully fetch the remote, or return some error message about the weird empty defaultBranch.

### Actual behavior

Assertion error or seg fault.

The top few frames of the stack are:

```
  * frame #0: 0x0000000107dcd7d5 libgit2.1.1.0.dylib`git__prefixcmp(str="", prefix="efs/heads/") at util.c:247:7
    frame #1: 0x0000000107d8835c libgit2.1.1.0.dylib`git_reference__is_branch(ref_name=0x0000000000000000) at refs.c:1207:9
    frame #2: 0x0000000107d9194e libgit2.1.1.0.dylib`ref_to_update(update=0x00007ffee7f682b8, remote_name=0x00007ffee7f682d0, remote=0x00007fc5bb40af90, spec=0x00007fc5bb614a70, ref_name=0x0000000000000000) at remote.c:1097:8
    frame #3: 0x0000000107d9170e libgit2.1.1.0.dylib`remote_head_for_ref(out=0x00007ffee7f68390, remote=0x00007fc5bb40af90, spec=0x00007fc5bb614a70, update_heads=0x00007ffee7f683f8, ref=0x00007fc5bb611b00) at remote.c:1144:15
    frame #4: 0x0000000107d91358 libgit2.1.1.0.dylib`git_remote_write_fetchhead(remote=0x00007fc5bb40af90, spec=0x00007fc5bb614a70, update_heads=0x00007ffee7f683f8) at remote.c:1182:13
```

It looks like [`remote_head_for_ref`](https://github.com/libgit2/libgit2/blob/f0d7922c9bafff38e12978625f467aafebe75145/src/remote.c#L1136-L1142) does not check for a NULL return value for the `ref_name`.  I'm not sure if there is some greater underlying issue, though.

f0d7922c9bafff38e12978625f467aafebe75145

macOS, Linux

ethomson:
Previously, when `init.defaultbranch` was present in the config but set to an empty string, we treated it as if the empty string was the default branch.  This meant we set `HEAD` to `refs/heads/`.  Instead, when `init.defaultbranch` is empty, we should ignore it.

Also, fetch should not try to create `FETCH_HEAD` entries that include an invalid branch.

Fixes #5761

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -1141,6 +1141,16 @@ static int remote_head_for_ref(git_remote_head **out, git_remote *remote, git_re
 		ref_name = git_reference_name(resolved_ref);
 	}
 
+	/*
+	 * The ref name may be unresolvable - perhaps it's pointing to
+	 * something invalid.  In this case, there is no remote head for
+	 * this ref.
+	 */
+	if (!ref_name) {
+		error = 0;
+		goto cleanup;
+	}
+
 	if ((error = ref_to_update(&update, &remote_name, remote, spec, ref_name)) < 0)
 		goto cleanup;
 
--- a/src/repository.c
+++ b/src/repository.c
@@ -2092,7 +2092,8 @@ static int repo_init_head(const char *repo_dir, const char *given)
 	if (given) {
 		initial_head = given;
 	} else if ((error = git_config_open_default(&cfg)) >= 0 &&
-	           (error = git_config_get_string_buf(&cfg_branch, cfg, "init.defaultbranch")) >= 0) {
+	           (error = git_config_get_string_buf(&cfg_branch, cfg, "init.defaultbranch")) >= 0 &&
+	           *cfg_branch.ptr) {
 		initial_head = cfg_branch.ptr;
 	}
 
--- a/tests/fetchhead/nonetwork.c
+++ b/tests/fetchhead/nonetwork.c
@@ -319,6 +319,16 @@ static int assert_master_for_merge(const char *ref, const char *url, const git_o
 	return 0;
 }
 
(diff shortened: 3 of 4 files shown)


lhchavez (approved):
(assuming the CI is happy)

PR #5773, opened 2021-01-07

ethomson:
Resurrecting #5150


csware:
So, any news with this PR?

csware:
@ethomson What do we do with this PR? Is there is anything needed, I cannot change anything.

csware:
All people relying on the autotection won't get the right config file ATM.

Btw. Git dropped the applkication config layer, should we drop it on libgit2 as well (or at least the auto-detection for it)?

ethomson:
Fixed in #6151 - thanks again @csware

Issue #5564, opened 2020-06-18
Support for partial clones


ThomasSowders:
Feature request: Add support for cloning and using repositories with the [partial clone](https://git-scm.com/docs/partial-clone/en) feature.

olsen232:
#5776 has been closed since it is superceded by #5993 (and follow-ups)

olsen232:
Basic plan - 
1. New error code EMISSING for when objects are referenced by the ODB, but missing from the ODB. #5993 
2. Detect when packfiles are promisor packfiles - #5996 
3. New error code EPROMISED - return it when an EMISSING occurs in a promisor-packfile (not yet done)

olsen232:
@ethomson just looking at recent commits, it seems like you are the best at making things happen in libgit2 - can you comment on if this is a reasonable plan, and maybe even find me a reviewer with the necessary permissions?

ethomson:
> @ethomson just looking at recent commits, it seems like you are the best at making things happen in libgit2 - can you comment on if this is a reasonable plan, and maybe even find me a reviewer with the necessary permissions?

👋  Hey @olsen232, sorry for the delay in getting back to you.  This seems reasonable to me.  I need to do a bit more research on partial clones but overall, I think that this seems like the right direction.

erikmchut:
@olsen232 are you still working on some of the prerequisites for partial clones?

olsen232:
Hi @erikmchut and @ethomson 

am I still working on it? Not really - I have a fork of libgit2 that seems to do what I want it to for now. I think Edward got a bit put off by the ambiguity of the problem, or git's implementation, or of my approach - the way it works in Git means that asking "is object 30f94c... missing but promised?" is unanswerable in a reasonable amount of time in the general case, but asking "is object 30f94c at path foo/bar/baz missing but promised" is answerable. This kinda messes up the existing error codes paradigm a bit and nothing ever got merged to libgit2.

The PR that got stuck is here: https://github.com/libgit2/libgit2/pull/5993

PR #5776, opened 2021-01-08
RFC: separate object lookup error for promisor packs

rcoup:
There's currently no support in libgit2 for partial clones (#5564). This attempts to start support them by distinguishing object-lookup faults between `ENOTFOUND` (the requested oid is bad or the repo is corrupt) from a new `EMISSING` error (indicating the oid is referenced from a promisor pack, and therefore _should_ be available remotely).

1. Detect a promisor packfile by looking for a `.packfile` sidecar and track it in an attribute of `git_pack_file`
2. When an object is referenced from a promisor packfile (eg. tree_entry → blob), raise EMISSING instead of ENOTFOUND if the object isn't present anywhere in the ODB.

To be clear, there's no support for _fetching_ missing objects from a promisor remote.

The code doesn't actually do (2) properly when there are multiple ODB backends in place, but hopefully it conveys the basic idea.

Conceptually, EMISSING could also be used for things like shallow clones where parent commits aren't available locally, etc.

### DIFF
--- a/src/pack.c
+++ b/src/pack.c
@@ -5,6 +5,8 @@
  * a Linking Exception. For full terms see the included COPYING file.
  */
 
+#include <string.h>
+
 #include "pack.h"
 
 #include "delta.h"
@@ -1171,6 +1173,8 @@ int git_packfile_alloc(struct git_pack_file **pack_out, const char *path)
 	struct stat st;
 	struct git_pack_file *p;
 	size_t path_len = path ? strlen(path) : 0, alloc_len;
+	size_t root_len;
+	char promisor_path[GIT_PATH_MAX];
 
 	*pack_out = NULL;
 
@@ -1190,8 +1194,7 @@ int git_packfile_alloc(struct git_pack_file **pack_out, const char *path)
 	 * the index looks sane.
 	 */
 	if (git__suffixcmp(path, ".idx") == 0) {
-		size_t root_len = path_len - strlen(".idx");
-
+		root_len = path_len - strlen(".idx");
 		if (!git_disable_pack_keep_file_checks) {
 			memcpy(p->pack_name + root_len, ".keep", sizeof(".keep"));
 			if (git_path_exists(p->pack_name) == true)
@@ -1199,6 +1202,22 @@ int git_packfile_alloc(struct git_pack_file **pack_out, const char *path)
 		}
 
 		memcpy(p->pack_name + root_len, ".pack", sizeof(".pack"));
+	} else {
+		root_len = path_len - strlen(".pack");
+	}
+
+	/*
+	 * Check for a .promisor file.
+	 */
+	memcpy(promisor_path, p->pack_name, root_len);
+	if (root_len + strlen(".promisor") > GIT_PATH_MAX) {
+		git_error_set(GIT_ERROR_OS, "failed to check packfile promisor");
+		git__free(p);
+		return -1;
+	}
+	memcpy(promisor_path + root_len, ".promisor", sizeof(".promisor"));
+	if (!p_stat(promisor_path, &st) && S_ISREG(st.st_mode)) {
+		p->pack_promisor = 1;
 	}
 
 	if (p_stat(p->pack_name, &st) < 0 || !S_ISREG(st.st_mode)) {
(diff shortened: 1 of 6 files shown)


olsen232:
Hi github maintainers, I'm willing to do some work to help improve libgit2's partial clone support. What can I do to help this along?

src/pack.c
```diff
@@ -1190,15 +1194,30 @@ int git_packfile_alloc(struct git_pack_file **pack_out, const char *path)
...
+		root_len = path_len - strlen(".idx");
 		if (!git_disable_pack_keep_file_checks) {
 			memcpy(p->pack_name + root_len, ".keep", sizeof(".keep"));
 			if (git_path_exists(p->pack_name) == true)
 				p->pack_keep = 1;
 		}
 
 		memcpy(p->pack_name + root_len, ".pack", sizeof(".pack"));
+	} else {
+		root_len = path_len - strlen(".pack");
```
ethomson, line 1206: In [line 1181](https://github.com/libgit2/libgit2/pull/5776/files#diff-226580d1635752a6f3386d44293260fbd4a33f2f405ba0b91fbd35c3d96315f9R1181) we ensure that `path_len` is at least `strlen(".idx")`.  If it were `strlen(".idx")`, here we would set `root_len` to `(size_t)-1`.

I think that it's unlikely that there's a call path that would call us with just `.idx` as the path but some sort of guard seems like it should be in order.

ethomson (commented):
Sorry about the delay, @rcoup; I haven't had a lot of bandwidth in 2020.  I think that the direction that this is going is largely correct.

> The code doesn't actually do (2) properly when there are multiple ODB backends in place, but hopefully it conveys the basic idea.

Is the problem here that we're short circuiting the lookup on the first promisor pack, and a subsequent odb backend may have the object?  (In other words, we erroneously report `EMISSING` even though we have the object?)  Or is there a deeper underlying problem that I'm not seeing yet?

> Conceptually, EMISSING could also be used for things like shallow clones where parent commits aren't available locally, etc.

I wonder if that's what we would want to do.  I agree that they're similar (object exists on the remote and doesn't exist locally) but I wonder if they should be distinct error codes so that a caller can distinguish them more easily.  🤔

rcoup:
> Is the problem here that we're short circuiting the lookup on the first promisor pack, and a subsequent odb backend may have the object? (In other words, we erroneously report EMISSING even though we have the object?) 

Yes. Presumably there might be some other complexities wrt multiple ODB backends.

>>Conceptually, EMISSING could also be used for things like shallow clones where parent commits aren't available locally, etc.
> 
> I wonder if that's what we would want to do. I agree that they're similar (object exists on the remote and doesn't exist locally) but I wonder if they should be distinct error codes so that a caller can distinguish them more easily. 🤔

Yes, maybe `EPROMISED` & `EGRAFT` (or `ESHALLOW_GRAFT`?) to make it clearer?

olsen232:
FYI I picked up this PR and started to address the review comments - however I ran into a major obstacle - `git_object_lookup` doesn't actually have the necessary context to know if an object is non-existent, missing-without-good-reason (ie corrupt), or promised. In order to say that an object is promised, we should ensure the following:

1. The object is not in the ODB.
2. The object is referenced by an object that *is* in the ODB.
3. The object that *is* in the ODB is in a packfile, and, that packfile is a promisor packfile.
Since `git_object_lookup` doesn't have any context except the OID that we're looking for, we can't make the distinction at this layer without changing the API.

This PR doesn't currently do those things and lacks the context to do so without changing the libgit2 API. I think this PR promotes ENOTFOUND to EPROMISED basically every time if there is a promisor packfile involved, which is too eager (unless we define the error code as meaning "this object might be promised", as opposed to "is promised"). I think this PR is probably best simply closed - there are two other PRs that I will try to write instead:

- Separate error code eg EMISSING for when an object lookup by path fails due to missing object (not due to an invalid path). For instance, if I run `git_tree_entry_bypath(&entry, root, "ab/cd/ef") and "ab/cd/ef" is not a valid path - perhaps ab has no child cd - then I get ENOTFOUND, as I should. But if I run `git_tree_entry_bypath(&entry, root, "ab/cd/ef")` and "ab/cd/ef" is valid, as far as can be tested, but an object is missing, I should get this new error instead. (Eg if ab does have a child cd, but that child is nowhere to be found.

- Separate error code eg EPROMISED for when an object is EMISSING (as in the last PR) *AND* the packfile where we would expect to find the object is a promisor packfile. I would salvage some of this PR to check if the promisor file exists.

Limitations - I can only add these codes to the path API eg `git_tree_entry_bypath`, `git_tree_entry_byname`, `git_object_lookup_bypath`. git itself also has this limitation eg if you run `git cat-file $SOME_OBJECT_ID` it doesn't have enough context to know if the object is non-existent or promised, so it has to search all promisor backends for it.

rcoup:
Closed in favour of #5993 and follow-ups

PR #5787, opened 2021-01-31
Update .gitignore

### LINKED ISSUE #5779: .gitignore file contains "tags" which ignores some folders in the test directory
The "tags" pattern in the .gitignore file;

https://github.com/libgit2/libgit2/blob/487f2a8287c4d1d41d39e05ac53d0e50c679d7e9/.gitignore#L5

means that (for example), the contents of these directories would be ignored:

tests/resources/empty_bare.git/refs/tags
tests/resources/testrepo.git/refs/tags
tests/resources/push_src/.gitted/refs/tags
tests/resources/empty_standard_repo/.gitted/refs/tags
tests/resources/describe/.gitted/refs/tags
tests/resources/testrepo/.gitted/refs/tags

Could the pattern be updated to not ignore these folders? I'm not sure what the pattern was supposed to refer to.

lrm29:
Resolves #5779: Only exclude tags files/folders and build folders at the repository root.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
-build/
+/build/
 .DS_Store
 *~
 .*.swp
-tags
+/tags
 CMakeSettings.json
 .vs


lhchavez (commented):
do we also want to exclude `build`? where is that one needed?

lrm29:
I can revert the change to exclude only **build** directories at the repository root. I made the assumption that it was to exclude the **build** directory typically created as part of a cmake build. There are no directories named build in libgit2's source at the moment.

ethomson:
Yes, it's common practice (and our docs encourage) you to use `build` for your cmake directory.  I think that we should keep it in our `.gitignore`.

ethomson:
Thanks for the fix!

PR #5791, opened 2021-02-07
clone: set refs/remotes/origin/HEAD to default branch when branch is specified, attempt 2

### LINKED ISSUE #5751: refs/remotes/origin/HEAD is not created when cloning a branch
### Reproduction steps

1. Set `checkout_branch` in `git_clone_options`
2. Call `git_clone` passing these options

### Expected behavior

`refs/remotes/origin/HEAD` file is created pointing to the remote default branch.

### Actual behavior

`refs/remotes/origin/HEAD` file is not being created at all.

v1.1.0
also f4b473fa842a84147d8036a27bac5c026c44dad4

Windows 10

A-Ovchinnikov-mx:
Another attempt to fix #5751 :)

I have first reverted the previous attempt and re-introduced changes that seem to work correctly. Please let me know if you don't want the revert though.

### DIFF
--- a/src/clone.c
+++ b/src/clone.c
@@ -259,30 +259,38 @@ cleanup:
 
 static int update_head_to_branch(
 		git_repository *repo,
-		const char *remote_name,
+		git_remote *remote,
 		const char *branch,
 		const char *reflog_message)
 {
 	int retcode;
 	git_buf remote_branch_name = GIT_BUF_INIT;
 	git_reference* remote_ref = NULL;
+	git_buf default_branch = GIT_BUF_INIT;
 
-	GIT_ASSERT_ARG(remote_name);
+	GIT_ASSERT_ARG(remote);
 	GIT_ASSERT_ARG(branch);
 
 	if ((retcode = git_buf_printf(&remote_branch_name, GIT_REFS_REMOTES_DIR "%s/%s",
-		remote_name, branch)) < 0 )
+		git_remote_name(remote), branch)) < 0 )
 		goto cleanup;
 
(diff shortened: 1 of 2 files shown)


A-Ovchinnikov-mx:
@ethomson Hi, could you please have a look at this and help me find a way to resolve the issue with the test? Many thanks! :)

ethomson:
Right.  I guess the tricky part here is whether we should require that the remote's default branch fits in the refspec configuration or not.  It seems reasonable that you should be able to clone a single branch and that branch not be the remote's default.

So I would say that we should:

When validating the remote default branch (`git_remote__matching_refspec` in `src/clone.c`), let's change that `refspec == NULL` test.  Keep erroring _if_ there was no checkout branch specified in the clone opts (ie, we cannot determine the default branch).  But remove the error _if_ there _was_ a checkout branch specified.

We should only set the remote `HEAD` if we were able to determine it.

We should update the `online::remotes::single_branch` test to ensure that we didn't create a remote tracking `HEAD`.

Does this seem reasonable?

A-Ovchinnikov-mx:
Wouldn't it be simpler to just add

```cpp
if (!git_remote__matching_refspec(remote, git_buf_cstr(default_branch)))
    goto cleanup;
```

after the attempt to get remote branch? 🤔 This way we won't be creating HEAD ref is it doesn't fit the refspec config.

I'm gonna check whether Git command line client creates it when you do this.

A-Ovchinnikov-mx:
Hi @ethomson I think I've managed to solve the issue with tests. Could you please give this Pull Request a review? 🙂

A-Ovchinnikov-mx:
Hi @ethomson Would you have some time to look into this? 🙂

PR #5800, opened 2021-02-16
refdb_fs: Check git_sortedcache wlock/rlock errors

mamapanda:
To prevent a race condition, these lock calls should probably be checked for errors.

### DIFF
--- a/src/refdb_fs.c
+++ b/src/refdb_fs.c
@@ -254,7 +254,8 @@ static int loose_lookup_to_packfile(refdb_fs_backend *backend, const char *name)
 	if ((error = loose_parse_oid(&oid, name, &ref_file)) < 0)
 		goto done;
 
-	git_sortedcache_wlock(backend->refcache);
+	if ((error = git_sortedcache_wlock(backend->refcache)) < 0)
+		goto done;
 
 	if (!(error = git_sortedcache_upsert(
 			(void **)&ref, backend->refcache, name))) {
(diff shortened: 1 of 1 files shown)


src/refdb_fs.c
```diff
@@ -254,7 +254,9 @@ static int loose_lookup_to_packfile(refdb_fs_backend *backend, const char *name)
 	if ((error = loose_parse_oid(&oid, name, &ref_file)) < 0)
 		goto done;
 
-	git_sortedcache_wlock(backend->refcache);
+	if ((error = git_sortedcache_wlock(backend->refcache)) < 0) {
+		goto done;
+	}
```
lhchavez, line 259:
```suggestion
	if ((error = git_sortedcache_wlock(backend->refcache)) < 0)
		goto done;
```

we can lose the braces for consistency with the other blocks above. same in L763 for consistency with this.

lhchavez (commented):
nice, thanks for this change!

i'll try adding support for gcc's [` __attribute__((warn_unused_result))`](https://gcc.gnu.org/onlinedocs/gcc-4.1.1/gcc/Function-Attributes.html) to avoid regressions and have the compiler warn us when cases like this happen.

src/refdb_fs.c
mamapanda: Thanks, dunno how I missed that

lhchavez (approved):
\o/

ethomson:
Thanks!

PR #5801, opened 2021-02-16
index: Check git_vector_dup error in write_entries

mamapanda:
If I'm understanding this right, `git_vector_sort` will segfault when allocating `case_sorted.contents` fails


lhchavez (dismissed):
another instance where a compiler warning would have saved us. thanks for these patches! :D

optional, for extra safety: initialize `entries = NULL` and `case_sorted =  GIT_VECTOR_INIT`. then we can change L2890-L2891 with an unconditional

```c
done:
	git_vector_free(&case_sorted);
```

and L2871 with `goto done;`.

ethomson:
Great eye - thanks for the fix!

ethomson (approved):

PR #5802, opened 2021-02-17
Introduce GIT_WARN_UNUSED_RESULT

lhchavez:
This change adds the GIT_WARN_UNUSED_RESULT annotation, which makes the
compiler warn when a return result is not used. This avoids bugs.

### DIFF
--- a/src/sortedcache.h
+++ b/src/sortedcache.h
@@ -58,7 +58,7 @@ typedef struct {
  *        may be NULL.  The cache makes it easy to load this and check
  *        if it has been modified since the last load and/or write.
  */
-int git_sortedcache_new(
+GIT_WARN_UNUSED_RESULT int git_sortedcache_new(
 	git_sortedcache **out,
 	size_t item_path_offset, /* use offsetof(struct, path-field) macro */
 	git_sortedcache_free_item_fn free_item,
@@ -71,7 +71,7 @@ int git_sortedcache_new(
  * - `copy_item` can be NULL to just use memcpy
  * - if `lock`, grabs read lock on `src` during copy and releases after
  */
-int git_sortedcache_copy(
+GIT_WARN_UNUSED_RESULT int git_sortedcache_copy(
 	git_sortedcache **out,
 	git_sortedcache *src,
 	bool lock,
@@ -88,7 +88,7 @@ void git_sortedcache_free(git_sortedcache *sc);
 void git_sortedcache_incref(git_sortedcache *sc);
 
 /* Get the pathname associated with this cache at creation time */
-const char *git_sortedcache_path(git_sortedcache *sc);
+GIT_WARN_UNUSED_RESULT const char *git_sortedcache_path(git_sortedcache *sc);
 
 /*
  * CACHE WRITE FUNCTIONS
@@ -100,7 +100,7 @@ const char *git_sortedcache_path(git_sortedcache *sc);
  */
 
 /* Lock sortedcache for write */
-int git_sortedcache_wlock(git_sortedcache *sc);
+GIT_WARN_UNUSED_RESULT int git_sortedcache_wlock(git_sortedcache *sc);
 
 /* Unlock sorted cache when done with write */
 void git_sortedcache_wunlock(git_sortedcache *sc);
@@ -120,7 +120,8 @@ void git_sortedcache_wunlock(git_sortedcache *sc);
  *
  * @return 0 if up-to-date, 1 if out-of-date, <0 on error
  */
-int git_sortedcache_lockandload(git_sortedcache *sc, git_buf *buf);
+GIT_WARN_UNUSED_RESULT int git_sortedcache_lockandload(
+	git_sortedcache *sc, git_buf *buf);
 
 /* Refresh file timestamp after write completes
  * You should already be holding the write lock when you call this.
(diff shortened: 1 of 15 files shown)


ethomson:
Naive question - is there an attribute for the inverse of this - where we accept that it's okay that the return value is unused?

I think that for _most_ of our codebase, we should enforce checking return codes.  I can only think of a few outliers (like the `buf` functions, where we accept that an OOM situation will eventually get checked, but we might chain several functions together before we do that check).

I have no idea how bad `-Wunused-result` would _actually_ be, and it looks like [it has some pain points of its own](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425), I'm just curious.

lhchavez:
> Naive question - is there an attribute for the inverse of this - where we accept that it's okay that the return value is unused?

not as a standard attribute :( the closest thing would be a custom clang-tidy check. i personally wouldn't be opposed to investigating how to get one working, but it's probably better to start with an allow-list of functions we know we want to keep using their return values always plus possibly another macro to mark the ignoring of a return value as acceptable (i.e. this change), and once the whole codebase is clean, we can switch it around.

> 
> I think that for _most_ of our codebase, we should enforce checking return codes. I can only think of a few outliers (like the `buf` functions, where we accept that an OOM situation will eventually get checked, but we might chain several functions together before we do that check).
> 
> I have no idea how bad `-Wunused-result` would _actually_ be, and it looks like [it has some pain points of its own](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425), I'm just curious.

we're already using `-Wunused-result`! that's what powers the warnings for functions annotated with `GIT_WARN_UNUSED`: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html says:

> `-Wno-unused-result`
> Do not warn if a caller of a function marked with attribute `warn_unused_result` (see Function Attributes) does not use its return value. The default is `-Wunused-result`.

lhchavez:
> > Naive question - is there an attribute for the inverse of this - where we accept that it's okay that the return value is unused?
> 
> not as a standard attribute :( the closest thing would be a custom clang-tidy check. i personally wouldn't be opposed to investigating how to get one working, but it's probably better to start with an allow-list of functions we know we want to keep using their return values always plus possibly another macro to mark the ignoring of a return value as acceptable (i.e. this change), and once the whole codebase is clean, we can switch it around.
> 
> > I think that for _most_ of our codebase, we should enforce checking return codes. I can only think of a few outliers (like the `buf` functions, where we accept that an OOM situation will eventually get checked, but we might chain several functions together before we do that check).
> > I have no idea how bad `-Wunused-result` would _actually_ be, and it looks like [it has some pain points of its own](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425), I'm just curious.

btw tweaking `GIT_UNUSED()` to https://github.com/libgit2/libgit2/pull/5802/commits/4ebb573f58c4896b5416dc16e7bb303f641db770#diff-07a0ff24f834c533d980cd176412c1cf4777562a3a10fe49ae049e06adff4487R46 eliminates most of the pain points. it uses a gcc-ism, but then again the `GIT_WARN_UNUSED` macro also does.

> 
> we're already using `-Wunused-result`! that's what powers the warnings for functions annotated with `GIT_WARN_UNUSED`: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html says:
> 
> > `-Wno-unused-result`
> > Do not warn if a caller of a function marked with attribute `warn_unused_result` (see Function Attributes) does not use its return value. The default is `-Wunused-result`.

ethomson:
> we're already using -Wunused-result! that's what powers the warnings for functions annotated with GIT_WARN_UNUSED: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html says:

Ohhhh, I didn't fully understand the interaction here.  This makes sense.  Thanks for the explanation.

src/sortedcache.h
```diff
@@ -155,26 +157,29 @@ int git_sortedcache_remove(git_sortedcache *sc, size_t pos);
...
 /* Lock sortedcache for read */
-int git_sortedcache_rlock(git_sortedcache *sc);
+GIT_WARN_UNUSED_RESULT int git_sortedcache_rlock(git_sortedcache *sc);
 
 /* Unlock sorted cache when done with read */
 void git_sortedcache_runlock(git_sortedcache *sc);
 
 /* Lookup item by key - returns NULL if not found */
-void *git_sortedcache_lookup(const git_sortedcache *sc, const char *key);
+GIT_WARN_UNUSED_RESULT void *git_sortedcache_lookup(
```
ethomson, line 166: I like `GIT_WARN_UNUSED_RESULT` for things that return errors that really must be paid attention to, for example things that allocate.  I'm less bullish on things like this.  One on hand, I think that I understand the motivation: _why are you calling this function_ if you don't care about the result?  It's side effect free.  So its _only purpose in life_ is to compute a result for you.  Calling it and ignoring the result is just a waste of cycles.

🤔

I feel like we should draw up some simple guidelines about where and when we use `GIT_WARN_UNUSED_RESULT`...  It feels like you're proposing a place where we land on it _almost always_.  Is that roughly accurate?

lhchavez: i could also be okay with restricting its use to only places that allocate or grab a mutex (that case is also pretty scary).

if/when another clang-tidy solution exists where everything is checked except for a few allow-listed functions, things'll be less verbose.

ethomson:
Sorry for the long delay.  One last question: does this need to be in the public api?  I’d rather keep this defined in some header in `src/`

lhchavez:
> Sorry for the long delay. One last question: does this need to be in the public api? I’d rather keep this defined in some header in `src/`

it doesn't need to, let me move it!

lhchavez:
> > Sorry for the long delay. One last question: does this need to be in the public api? I’d rather keep this defined in some header in `src/`
> 
> it doesn't need to, let me move it!

welp, not sure what happened, but now the MSVC builds are not picking the macro up :/

ethomson:
This is really nice.  Thanks.

Issue #5799, opened 2021-02-16
Extra want in git_remote_download


FintanH:
### Reproduction steps

I haven't been able to reduce the steps to something concrete, but I'm hoping that you might have some insight for me where I could work towards narrowing down the case to something reproducible :)

We use the Rust bindings of `libgit2` in [radicle-link](https://github.com/radicle-dev/radicle-link/) and it's the backbone of our protocol. The scenario is that we have a node that will download references from my machine when I publish the references and announce them to the network. From a fresh slate, everything works fine for about 4 sets of references and then the issue occurs.

Running with `GIT_TRACE_PACKET` we see the following order of messages:
```
 INFO  librad::git::p2p::server          > upload pack 
18:20:02.180060 common-main.c:48                  version 2.29.2
18:20:02.180087 common-main.c:49                  start git upload-pack --strict --timeout=5 --stateless-rpc .
18:20:02.180128 git.c:445                         cmd_name upload-pack (upload-pack)
18:20:02.201361 pkt-line.c:80           packet:  upload-pack< want eb2101795d25bd680646940dc47f7933d24adde0 multi_ack_detailed side-band-64k include-tag thin-pack ofs-delta 
18:20:02.201518 pkt-line.c:80           packet:  upload-pack< 0000
18:20:02.201537 pkt-line.c:80           packet:  upload-pack< have ca7e13861f7a1d071e5ec695b941c2d010799dc8
18:20:02.201632 pkt-line.c:80           packet:  upload-pack> ACK ca7e13861f7a1d071e5ec695b941c2d010799dc8 common
18:20:02.201652 pkt-line.c:80           packet:  upload-pack< want eb2101795d25bd680646940dc47f7933d24adde0 multi_ack_detailed side-band-64k include-tag thin-pack ofs-delta 
18:20:02.201662 usage.c:64                        error git upload-pack: expected SHA1 list, got 'want eb2101795d25bd680646940dc47f7933d24adde0 multi_ack_detailed side-band-64k include-tag thin-pack ofs-delta '
fatal: git upload-pack: expected SHA1 list, got 'want eb2101795d25bd680646940dc47f7933d24adde0 multi_ack_detailed side-band-64k include-tag thin-pack ofs-delta '
18:20:02.201680 usage.c:68                        exit elapsed:0.022073 code:128
18:20:02.201688 trace2/tr2_tgt_normal.c:123       atexit elapsed:0.022086 code:128
[...]

PR #5804, opened 2021-02-20
Don't send want-list more than once in a packfile negotiation

lhchavez:
This change makes libgit2 be a bit more compliant with the packfile
negotiation spec[1] by avoiding sending the want-list more than once per
negotiation. This also is in line with how git's `find_common()`[2]
works, in which the
`want`s/`shallow`s/`deepen`s/`deepen-since`s/`deepen-not`s are sent once
before the second phase of the packfile negotiation starts, and that
phase only sends `have`s.

Fixes: #5799

1: https://git-scm.com/docs/pack-protocol#_packfile_negotiation
2: https://github.com/git/git/blob/2283e0e9af55689215afa39c03beb2315ce18e83/fetch-pack.c#L267

### DIFF
--- a/src/transports/smart_protocol.c
+++ b/src/transports/smart_protocol.c
@@ -389,9 +389,6 @@ int git_smart__negotiate_fetch(git_transport *transport, git_repository *repo, c
 			git_pkt_ack *pkt;
 			unsigned int j;
 
-			if ((error = git_pkt_buffer_wants(wants, count, &t->caps, &data)) < 0)
-				goto on_error;
-
 			git_vector_foreach(&t->common, j, pkt) {
 				if ((error = git_pkt_buffer_have(&pkt->oid, &data)) < 0)
 					goto on_error;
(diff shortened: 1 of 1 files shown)


src/transports/smart_protocol.c
```diff
@@ -389,9 +389,6 @@ int git_smart__negotiate_fetch(git_transport *transport, git_repository *repo, c
 			git_pkt_ack *pkt;
 			unsigned int j;
 
-			if ((error = git_pkt_buffer_wants(wants, count, &t->caps, &data)) < 0)
-				goto on_error;
-
 			git_vector_foreach(&t->common, j, pkt) {
 				if ((error = git_pkt_buffer_have(&pkt->oid, &data)) < 0)
```
lhchavez, line 393: i'm still unsure whether this block and the one in L415 still need to be here...

ethomson:
I don't remember all the rules and expectations that git has here...  @carlosmn may have some more insight into this than I do.  🤔

carlosmn:
In RPC mode (i.e. HTTP or v2 IIUC) each request exists on its own and maybe routed to an arbitrary node. Thus we need to recap the previous rounds if we perform multiple negotiation rounds. This includes the haves and wants as well as any commits which the server specified were common in the previous round. I haven't looked deeply into the reported issue, but you should be careful that you don't break the HTTP negotation here.

Issue #5808, opened 2021-02-27
git_status_list_new can fail for files greater than 4GB


J0Nes90:
When working with bigger files (>4GB) libgit2 fails with "failed to read descriptor: invalid argument" when calling git_status_list_new.
The issue has two aspects:

First, modified_uncertain calls git_diff__oid_for_entry which calls git_odb__hashfd_filtered with entry.file_size from the git index - which is truncated to 32 bits. So for any file exceeding that 32 bit filesize limit this call fails when computing the hash for the file 
https://github.com/libgit2/libgit2/blob/0fb58396835f789b3c662184003ff34cbf98d248/src/diff_generate.c#L649

Second, when working with filters (git LFS in this case), the call to git_odb__hashfd_filtered reads the entire file into memory which fails as well for any size exceeding ssize_t https://github.com/libgit2/libgit2/blob/0fb58396835f789b3c662184003ff34cbf98d248/src/odb.c#L260

In my pull request I propose a solution for this by:

First, do not take the file size from the index but read it instead:
https://github.com/J0Nes90/libgit2/commit/c2d3929d6bcfc1705fe9011cba77d32bacf42e11

Second, use streaming filters to not load the entire file in memory:
https://github.com/J0Nes90/libgit2/commit/19233869f47d40a3ae6fecb5a440416d9e741749

Any feedback is highly appreciated! I also tested a repository with such big files in GitKraken and I could see the very same issue and error message as well. So this might be very interesting for @ianhattendorf

personalizedrefrigerator:
I would like to link this issue to #4230!

PR #5809, opened 2021-02-27
Fix/odb hash fd stream

J0Nes90:
This PR is a proposed fix for issue: https://github.com/libgit2/libgit2/issues/5808

Feedback is highly appreciated :)

### DIFF
--- a/include/git2/filter.h
+++ b/include/git2/filter.h
@@ -11,6 +11,7 @@
 #include "types.h"
 #include "oid.h"
 #include "buffer.h"
+#include "posix.h"
 
 /**
  * @file git2/filter.h
@@ -160,6 +161,18 @@ GIT_EXTERN(int) git_filter_list_apply_to_file(
 	const char *path);
 
 /**
+ * Apply a filter list to the contents of a git_file on disk
+ *
+ * @param out buffer into which to store the filtered file
+ * @param filters the list of filters to apply
+ * @param fd the file to filter
+ */
+GIT_EXTERN(int) git_filter_list_apply_to_git_file(
+	git_buf *out,
+	git_filter_list *filters,
+	git_file fd);
+
+/**
  * Apply a filter list to the contents of a blob
  *
  * @param out buffer into which to store the filtered file
@@ -196,6 +209,18 @@ GIT_EXTERN(int) git_filter_list_stream_file(
 	git_filter_list *filters,
 	git_repository *repo,
 	const char *path,
+	git_writestream *target);
+
+/**
+ * Apply a filter list to a git_file as a stream
+ *
+ * @param filters the list of filters to apply
+ * @param fd the file to filter
+ * @param target the stream into which the data will be written
+ */
+GIT_EXTERN(int) git_filter_list_stream_git_file(
+	git_filter_list *filters,
+	git_file fd,
 	git_writestream *target);
 
 /**
--- a/src/diff_generate.c
+++ b/src/diff_generate.c
@@ -645,8 +645,11 @@ int git_diff__oid_for_entry(
 		if (fd < 0)
 			error = fd;
 		else {
+			uint64_t file_size;
+			if ((error = git_futils_filesize(&file_size, fd)) < 0)
+				file_size = (uint64_t)entry.file_size;
 			error = git_odb__hashfd_filtered(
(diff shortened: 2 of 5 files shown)


J0Nes90:
Mh, those error seem to come from my posix.h include in filter.h. I did that to get access to git_filter. I will have a look on Monday.

src/diff_generate.c
```diff
@@ -645,8 +645,11 @@ int git_diff__oid_for_entry(
 		if (fd < 0)
 			error = fd;
 		else {
+			uint64_t file_size;
+			if ((error = git_futils_filesize(&file_size, fd)) < 0)
+				file_size = (uint64_t)entry.file_size;
 			error = git_odb__hashfd_filtered(
-				out, fd, (size_t)entry.file_size, GIT_OBJECT_BLOB, fl);
+				out, fd, (size_t)file_size, GIT_OBJECT_BLOB, fl);
```
ethomson, line 652: I think that I would take the opportunity to invert the `oid_for_entry` / `oid_for_file` interaction.

Right now, if we call `oid_for_file` with the actual known size of the file (as a `git_object_size_t`), then we discard the high bits to stuff that into a `git_index_entry` and call `oid_for_entry`.  You're proposing doing another `stat` to look up the filesize.

Instead, let's make `oid_for_entry` call `oid_for_file` - any existing callers of `oid_for_file` can work unchanged (and accurately, since they don't cast away the high bits of the size).  If we really need to look up the file size on disk when we have a `git_index_entry` (since it is only 32 bits by definition) then we can do it only for them.

However, looking a bit more closely, I don't actually know why `git_odb__hashfd_filtered` _takes_ a size.  It doesn't need it.  In fact, I'm a little skeptical that there aren't some bugs here when we _do_ give it a size.  I haven't done an audit but you can imagine a scenario where you pass the size to this function, and it reads exactly _up to that size_ and then stops and returns the hash of the data _up to that point_ even though there's still more bytes in that file descriptor.

This is not a big deal if we always stat and read - it's a race condition, sure, but that's rather undefined.  My bigger concern would be if we take a `git_index_entry` out of the index, and assume that we should read `file_size` bytes.  Maybe we don't do that and my concerns are unfounded, but if we want the OID for a file on-disk, I'm thinking that there's no reason not to just _read the whole file on disk_ and not take a size parameter at all.

include/git2/filter.h
```diff
@@ -159,6 +160,18 @@ GIT_EXTERN(int) git_filter_list_apply_to_file(
...
+ * Apply a filter list to the contents of a git_file on disk
+ *
+ * @param out buffer into which to store the filtered file
+ * @param filters the list of filters to apply
+ * @param fd the file to filter
+ */
+GIT_EXTERN(int) git_filter_list_apply_to_git_file(
+	git_buf *out,
+	git_filter_list *filters,
+	git_file fd);
```
ethomson, line 173: I don't want to expose an API where people give us file descriptors.  `git_file` should be an internal abstraction.

I think that adding this helper method is reasonable, but it looks like we can safely keep it an internal function, does that seem correct?

ethomson:
Cool, thanks for tackling this.  There are undoubtedly still a few more places where we assume that we can pull a full blob into memory like this that are going to hurt you.  (And some of them are just sort of "the way git is designed" like CR/LF filtering.  😢 )  Sorry about that.

But this is a really positive step forward.  Two minor requests but I think this is very much in the right direction to solve these issues.  Thanks!

include/git2/filter.h
J0Nes90: This makes sense, I've moved the 2 functions to the private API

src/diff_generate.c
J0Nes90: I like the idea of inverting the `oid_for_entry` / `oid_for_file` interaction. I implemented that idea in the updated pull request. This causes more calls to `stat` when computing the `oid_for_entry`. One could optimise this by first checking if the `git_index_entry` might be exceeding 32 bits, what do you think?

I am not sure if I understand 100% of your reasoning and what other implications my changes might have regarding `git_odb__hashfd_filtered`. Frankly speaking I am not feeling confident refactoring that part of the code.

J0Nes90:
You are welcome! Our app makes heavy usage of git LFS filters and during our beta phase we stumbled upon this issue. Beside that, libgit2 seems to handle large files fairly well. I was able to smudge and clean huge files on mac and windows

PR #5814, opened 2021-03-01
winhttp: skip certificate check if unable to send request

ianhattendorf:
In some circumstances (e.g. when proxies are involved), winhttp will fail to reach the WINHTTP_CALLBACK_STATUS_SENDING_REQUEST phase. If this occurs, we'll error with ERROR_WINHTTP_INCORRECT_HANDLE_STATE when attempting to query the server certificate context (see https://docs.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpsendrequest#remarks).

To avoid this, verify that WinHttpSendRequest has reached the WINHTTP_CALLBACK_STATUS_SENDING_REQUEST phase before checking the certificate. Since we're using WinHTTP in synchronous mode, we know for sure that once WinHttpSendRequest returns we've either sent it successfully or not.

NOTE: WINHTTP_CALLBACK_STATUS_SENDING_REQUEST appears to be deprecated with no direct replacement. WINHTTP_CALLBACK_STATUS_SENDREQUEST_COMPLETE is only available in async mode, and there doesn't appear to be a method of querying this flag outside of the status callback.

### DIFF
--- a/src/transports/winhttp.c
+++ b/src/transports/winhttp.c
@@ -111,7 +111,8 @@ typedef struct {
 	DWORD post_body_len;
 	unsigned sent_request : 1,
 		received_response : 1,
-		chunked : 1;
+		chunked : 1,
+		status_sending_request_reached: 1;
 } winhttp_stream;
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


ianhattendorf:
(Confirming test fails on CI before pushing fix)

ianhattendorf:
image

```
    Start 6: proxy

6: Test command: D:\a\libgit2\libgit2\build\Debug\libgit2_clar.exe "-v" "-sonline::clone::proxy"
6: Test timeout computed to be: 10000000
6: Loaded 394 suites: 
6: Started (test status codes: OK='.' FAILURE='F' SKIPPED='S')
6: 
6: online::clone...F..
6: 
6:   1) Failure:
6: online::clone::proxy_credentials_in_url_https [D:\a\libgit2\libgit2\tests\online\clone.c:888]
6:   Function call failed: (git_clone(&g_repo, "https://github.com/libgit2/TestGitRepository", "./foo", &g_options))
6:   error -1 - failed to get server certificate: The handle is in the wrong state for the requested operation
```

ethomson:
lgtm!  Thanks @ianhattendorf!  🎉

PR #5815, opened 2021-03-03
tree: deprecate `git_treebuilder_write_with_buffer`

ethomson:
The function `git_treebuilder_write_with_buffer` is unnecessary; we
always write with a buffer, so we might as well keep it around in the
treebuilder struct instead of recreating it.  Move the buffer as the
trebuilder as a cache, meaning the `write_with_buffer` function is
unnecessary.


lhchavez (approved):
UBSan proving its worth!

ethomson:
> UBSan proving its worth!

So good.  If we've got to write C then I'm happy to have all these guardrails.  😁   Thanks for all your work on the sanitization.

PR #5817, opened 2021-03-03
Check git_signature_dup failure

mamapanda:
I figured these calls should probably be checked, since `git_signature_dup` can have an allocation failure.

There's also one in `hunk_from_entry`, but I'm not sure if that call has to be checked because `blame_internal` doesn't check `hunk_from_entry` for any errors (should it?).

### DIFF
--- a/src/blame.c
+++ b/src/blame.c
@@ -76,6 +76,14 @@ static git_blame_hunk* new_hunk(
 	return hunk;
 }
 
+static void free_hunk(git_blame_hunk *hunk)
+{
+	git__free((void*)hunk->orig_path);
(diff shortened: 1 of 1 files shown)


lhchavez (approved):
> I figured these calls should probably be checked, since `git_signature_dup` can have an allocation failure.

yes they should!

> There's also one in `hunk_from_entry`, but I'm not sure if that call has to be checked because `blame_internal` doesn't check `hunk_from_entry` for any errors (should it?).

yeah, anything that has an allocation failure should be checked. but this change is good as-is. thanks for making these!

ethomson:
Thanks @mamapanda!

PR #5823, opened 2021-03-20
Working directory path validation

ethomson:
Add `git_path_validate_ondisk` which validates... well, on-disk paths.  These paths could be working directory paths or things inside the `.git` directory.

The goal is to limit Windows path lengths to `MAX_PATH` in general.  For working directory paths when `core.longpaths` is set, we will allow arbitrary length paths.

We do _not_ consult `core.longpaths` for things outside the working directory and will restrict any paths inside `.git` to be less than `MAX_PATH`.  Note that this is _not_ in keeping with Git for Windows, which does no validation and will fail silently when trying to construct a directory where a critical component of the `.git` directory would be longer than `MAX_PATH`.  🤷 

This converts several path writers to use this, but not all.  In particular, submodules and worktrees will require some more careful thought.  In addition, I'll perform a more extensive audit before converting this from a draft to a real PR.  🙃


ethomson:
Thinking through this, I think that a rough validation plan is:

1. Anything that reads/writes in the working directory (eg, checkout, staging, patch application, etc.)  There are several of these - and there are some weird cases around worktrees and submodules that need to be dealt with - but this is largely pretty straightforward since we have a repository and can look up its configuration.

2. Anything that deals with references.  We should validate that loose references _or_ reflogs will not end up getting written with > MAX_PATH characters.  There's no prior art here in Git for Windows that we are required to emulate:  although Git for Windows did not really define whether `core.longpaths` applies to references or just things in the working directory, in practice it seems to ignore the setting and just blindly try writing very long reference names and ignoring the failure.  🤷 

3. Everything else inside the `.git` folder: we can limit the gitpath to `MAX_PATH - length_of_longest_filename_inside`.  (I think that this is the length of a loose object.)  This means that we don't need to validate _every since call to `CreateFile`, we can just validate the length of the git directory when we open a repository.  😅

ethomson:
Going to go ahead and merge this to land #5347

PR #5824, opened 2021-03-20
fix check for ignoring of negate rules

palmin:
A gitignore rule like "*.txt" should be negated by "!dir/test.txt" even though one is inside a directory since *.txt isn't restricted to any directory.

This is fixed by making wildcard match treat * like ** by excluding the WM_PATHNAME flag when the rule isn't for a full path.

This problem can be replicated with a repository with the following .gitignore file:

     *.txt
     !/dir/test.txt

by creating a `dir/test.txt` file.

Commandline Git shows `dir/test.txt` as untracked while libgit2 without this PR shows it as ignored.

### DIFF
--- a/src/ignore.c
+++ b/src/ignore.c
@@ -101,7 +101,7 @@ static int does_negate_pattern(git_attr_fnmatch *rule, git_attr_fnmatch *neg)
  */
 static int does_negate_rule(int *out, git_vector *rules, git_attr_fnmatch *match)
 {
-	int error = 0, wildmatch_flags;
+	int error = 0, wildmatch_flags, effective_flags;
 	size_t i;
 	git_attr_fnmatch *rule;
 	char *path;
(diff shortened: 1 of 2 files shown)


lhchavez (commented):
can there be a new test case for this in [tests/ignore/path.c](https://github.com/libgit2/libgit2/blob/main/tests/ignore/path.c)?

bonus points if there are other tests cases added to make sure that other edge cases like ignores with wildcards that match intermediate directories behave as they do in git (i couldn't find any that didn't, fwiw).

palmin:
@lhchavez There is now a test case.

src/ignore.c
```diff
@@ -141,8 +141,17 @@ static int does_negate_rule(int *out, git_vector *rules, git_attr_fnmatch *match
 		if (git_buf_oom(&buf))
 			goto out;
 
+		/* if rule isn't for full path we match without PATHNAME flag
+		 * as lines like *.txt should match something like dir/test.txt
+		 * requiring * to also match /
+		*/
```
ethomson, line 147:
```suggestion
		/*
		 * if rule isn't for full path we match without PATHNAME flag
		 * as lines like *.txt should match something like dir/test.txt
		 * requiring * to also match /
		 */
```

src/ignore.c
```diff
@@ -141,8 +141,17 @@ static int does_negate_rule(int *out, git_vector *rules, git_attr_fnmatch *match
...
+		effective_flags = wildmatch_flags;
+		if (!(rule->flags & GIT_ATTR_FNMATCH_FULLPATH)) {
+			effective_flags &= ~WM_PATHNAME;
+		}
```
ethomson, line 151:
```suggestion
		if (!(rule->flags & GIT_ATTR_FNMATCH_FULLPATH))
			effective_flags &= ~WM_PATHNAME;
```

ethomson, line 151:
```suggestion
		if (!(rule->flags & GIT_ATTR_FNMATCH_FULLPATH))
			effective_flags &= ~WM_PATHNAME;
```

ethomson:
LGTM!  Thanks!  🎉

PR #5832, opened 2021-04-01
Default to GIT_BRANCH_DEFAULT if init.defaultBranch is empty string

ianhattendorf:
`git_repository_is_empty` is returning `false` if `init.defaultBranch` is set to `""`. It probably shouldn't be an empty string, but work around this like in the referenced issue/PR.

See:
- https://github.com/libgit2/libgit2/issues/5761
- https://github.com/libgit2/libgit2/pull/5770


ethomson:
Thanks!

PR #5839, opened 2021-04-13
Fix bug in git_diff_find_similar.

### LINKED ISSUE #5811: Reproducible "unrecoverable internal error: 'delta_is_split(tgt)'" from git_diff_find_similar
### Reproduction steps

1. Make a temp work dir for reproducing this bug: `mkdir foo && cd foo`
2. Clone the reduced test repository at https://github.com/staktrace/libgit2-bug-find-similar: `git clone https://github.com/staktrace/libgit2-bug-find-similar` (it's "reduced" in the sense that only the two relevant commits are retained, but each probably contains more files than is needed to reproduce the problem)
3. Copy the two files in the gist at https://gist.github.com/staktrace/c592da01ca8d49830c3ef4311dd39f4d into the `foo` folder as well. This contains a minimal program that uses the libgit2 library and reproduces the problem
4. Run `make libgit2 && make testcase`. (Note: if you already have libgit2 installed on your system you can skip the `make libgit2` step and modify the paths in the `make testcase` command accordingly)

### Expected behavior

The test program runs without any errors

### Actual behavior

The `git_diff_find_similar` call at the end of the testcase triggers an `unrecoverable internal error: 'delta_is_split(tgt)'` which seems like it's hitting a bug inside libgit2.

Current master branch, SHA `1bc8dec4c65ee7c7a005c2773b381272a74fc254`

Linux (Ubuntu 20.04.1 LTS)

staktrace:
When a split src gets turned into a rename, it should
also lose the IS_RENAME_TARGET flag, so that it doesn't
get processed in a subsequent iteration as a rename target.
Doing so can cause an assertion failure because it no
longer has the SPLIT flag.

Fixes #5811.


staktrace:
Looking for feedback on this patch. I spent a long time trying to wrap my head around this find_similar code and I'm not really confident I understood it properly. But assuming my understanding is correct, this seems like the right fix. If so I can write up tests and so on, but I'd rather get the proper fix first. Thanks!

ethomson:
> I spent a long time trying to wrap my head around this find_similar code and I'm not really confident I understood it properly.

Indeed!  I didn't write this code _and_ I haven't played with rename detection / rewrite splitting in a loooong time, so I'll need to get up to speed, too.  Any chance that you have a minimum reproducible example that we could use for testing?

staktrace:
There's a fairly reduced example in #5811. I'm working on reducing it further to include as a testcase for this patch but I have other time commitments so progress is a bit slow. Hopefully I'll get it done this coming weekend (or sooner).

staktrace:
@ethomson Ok, I further reduced the testcase and added it to the clar suite in the second commit in this PR. Please take a look when you get a chance!

staktrace:
Please let me know if there's anything else I can do to help move this along!

staktrace:
Rebased to main and dropped [wip] status. I have used this patch on a bunch of repos and it seems to be working well as far as I can tell.

staktrace:
@arrbee as it seems you have touched/written most of this file, could you do a review?

arrbee:
> @arrbee as it seems you have touched/written most of this file, could you do a review?

I have not touched libgit2 code in about 6 years, so I feel like it would be inappropriate for me to approve this PR. It would be stepping on the toes of the active maintainers. That being said, I've read through the reported issue and traced through the code, at this does seem like a good fix to me. Once the file in question has been labeled as a successfully matched rename, it seems correct to remove it from the list so as the iteration continues it is not reconsidered and does not cause the assertion to fail.

This isn't fresh enough in my mind to say if there is a different underlying issue where the assertion is actually incorrect and a file should be allowed to be the result of a rewrite of 2 sources (e.g. a hypothetical where two almost identical files are merged into a single resulting file that is labeled as the rewritten result of both), but I don't think that would be expected behavior in the first place. Therefore I expect this fix is good.

This is me trying to dredge up 6 year old recollection of the code, so I apologize that I'm not super confident in giving feedback here. I just felt I should comment in case it has any value to the active maintainers. :wave: 😍

staktrace:
Thank you, that's all I was looking for - just some extra confidence that the patch seems reasonable. @ethomson I realize you're busy but if there's anything else I can do to get this merged please let me know. If not I'll probably have to end up maintaining a fork which is quite inconvenient just for a two line fix :/

ethomson:
Thanks for the fix and the very helpful test.  This is great!

And thanks @arrbee, and welcome back.  Any time you want to take back over as maintainer, we'd be happy to have you.  😁

ethomson (approved):

staktrace:
Great, thank you both!

PR #5841, opened 2021-04-14
Checkout dry-run

J0Nes90:
With `GIT_CHECKOUT_NONE` you do get notified about conflicts but that's pretty much it. 
With this simple addition one can set `GIT_CHECKOUT_DRY_RUN` which will notify following the normal checkout routines but breaks just before the working directory gets touched. 
This becomes really handy when you want to do a proper checkout dry run to see what will be changed

### DIFF
--- a/include/git2/checkout.h
+++ b/include/git2/checkout.h
@@ -177,6 +177,9 @@ typedef enum {
 	/** Normally checkout writes the index upon completion; this prevents that. */
 	GIT_CHECKOUT_DONT_WRITE_INDEX = (1u << 23),
 
+	/** Stop checkout after the notifications happend but before the working directory is touched. */
+	GIT_CHECKOUT_DRY_RUN = (1u << 24),
+	
 	/**
 	 * THE FOLLOWING OPTIONS ARE NOT YET IMPLEMENTED
 	 */
--- a/include/git2/checkout.h
+++ b/include/git2/checkout.h
@@ -177,7 +177,10 @@ typedef enum {
 	/** Normally checkout writes the index upon completion; this prevents that. */
 	GIT_CHECKOUT_DONT_WRITE_INDEX = (1u << 23),
 
-	/** Stop checkout after the notifications happend but before the working directory is touched. */
+	/**
+	 * Show what would be done by a checkout.  Stop after sending
+	 * notifications; don't update the working directory or index.
+	 */
 	GIT_CHECKOUT_DRY_RUN = (1u << 24),
 	
 	/**
--- a/src/checkout.c
+++ b/src/checkout.c
@@ -2606,6 +2606,9 @@ int git_checkout_iterator(
 	if ((error = checkout_get_actions(&actions, &counts, &data, workdir)) != 0)
 		goto cleanup;
 
+	if (data.strategy & GIT_CHECKOUT_DRY_RUN)
+		goto cleanup;
+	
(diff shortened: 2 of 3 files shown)


ethomson:
🤔  I feel like this is what `GIT_CHECKOUT_NONE` was _intended_ to do.  🙁

Instead of introducing a new checkout type, what if we made `GIT_CHECKOUT_NONE` behave as the mechanism that you're proposing?

J0Nes90:
If that is indeed the intend of GIT_CHECKOUT_NONE, sure. I just thought changing the current behavior might (will) break running code

ethomson:
> If that is indeed the intend of GIT_CHECKOUT_NONE, sure. I just thought changing the current behavior might (will) break running code

Yeah, I appreciate that.  There's definitely some chance that people are doing a `CHECKOUT_NONE` and not expecting callbacks to fire.  But it feels like a reasonable change to me, I'd love to see the impact that it has on our tests (if any) as a gauge of the impact.  (I suspect it will be minimal.)

J0Nes90:
Actually, the advantage of my method is that you can do a dry-run for both, `GIT_CHECKOUT_SAFE` and `GIT_CHECKOUT_FORCE`.
It would be strange to mix those with `GIT_CHECKOUT_NONE`

ethomson:
Aha, yes, this makes sense.  I'd like to think through the semantics of how things that call checkout (eg patch application or merge) would work here - I'd need to double check but I think that there are some places that suggest that a merge dry run sets the checkout strategy to NONE.  But this seems very reasonable to me.

ethomson:
Sorry for the delay here.  I like this approach.  Can you clarify the documentation on what `GIT_CHECKOUT_NONE` does?  It still says that it does a "dry run" which now has a new definition.  I'd love it if we could define precisely what happens with `GIT_CHECKOUT_NONE` in the same way that you've documented precisely what happens with `DRY_RUN`.

I think that in a perfect world, we would reshuffle things so that `GIT_CHECKOUT_SAFE = 0`, `GIT_CHECKOUT_NONE` and `GIT_CHECKOUT_DRY_RUN` are aliases for another.  That way you can dry-run a safe or a force checkout, and there's a good default of `SAFE`.  We would probably give it another name like `DEFAULT` and keep `SAFE` an alias for compatibility.

Maybe this is a useful exercise for 2.0.  🤔

include/git2/checkout.h
```diff
@@ -177,6 +177,9 @@ typedef enum {
 	/** Normally checkout writes the index upon completion; this prevents that. */
 	GIT_CHECKOUT_DONT_WRITE_INDEX = (1u << 23),
 
+	/** Stop checkout after the notifications happend but before the working directory is touched. */
```
ethomson, line 180:
```suggestion
	/**
	 * Show what would be done by a checkout.  Stop after sending
	 * notifications; don't update the working directory or index.
	 */
```

ethomson (approved):

ethomson:
Thanks for this!

WizzKidd777 (commented):
var sriToolbox = require("sri-toolbox");

var jquerySourceCode = file("jquery-1.10.2.min.js");

var integrity = sriToolbox.generate({
    algorithms: ["sha256"]
}, jquerySourceCode);
//=> "sha256-C6CB9UYIS9UJeqinPHWTHVqh/E1uhG5Twh+Y5qFQmYg="

PR #5853, opened 2021-05-01
diff:add option to ignore blank line changes

yuuri:
I found out that we have defined the parameter  (`XDF_IGNORE_BLANK_LINES`) is in `src/xdiff/xdiff.h`.

I added support for this parameter(`GIT_DIFF_IGNORE_BLANK_LINES`) in `include/git2/diff.h` and `src/diff_xdiff.c`
This is my first PR, if there is anything wrong, please point it out and I will be happy to improve it.😊

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /tags
 CMakeSettings.json
 .vs
+.idea
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -168,6 +168,10 @@ typedef enum {
 	 *  can apply given diff information to binary files.
 	 */
 	GIT_DIFF_SHOW_BINARY = (1u << 30),
+
+	/** Ignore blank lines */
+	GIT_DIFF_IGNORE_BLANK_LINES = (1u << 31),
+
 } git_diff_option_t;
 
 /**
--- a/src/diff_xdiff.c
+++ b/src/diff_xdiff.c
@@ -259,5 +259,8 @@ void git_xdiff_init(git_xdiff_output *xo, const git_diff_options *opts)
 	if (flags & GIT_DIFF_MINIMAL)
 		xo->params.flags |= XDF_NEED_MINIMAL;
 
+	if (flags & GIT_DIFF_IGNORE_BLANK_LINES)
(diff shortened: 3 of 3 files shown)


ethomson:
Hi @yuuri!  Looks good!  One thing that would be really helpful would be a test that exercises the behavior.  Would you be able to add one?

Our test framework is bespoke so I’m happy to help - but if you can provide the inputs and the “what would git do” output with this flag, that would be a great start!

yuuri:
Wow,@ethomson.Thanks for your comment! I am now reading the test code,so when understand it,I will add test function at once.
Now,please allow me to show you what I want to say.

As we known，when we want to show diff compare two commit，we can use `git diff commit_sha1 commit_sha2`,
it will show us two commit difference.And I notice git diff can add a parameter to ignore blank lines. Command look likes this 
`git diff --ignore-blank-lines `

1. origin diff show 
```
$ git diff
diff --git a/main.py b/main.py
index b847ef7..30ceabc 100644
--- a/main.py
+++ b/main.py
@@ -15,6 +15,8 @@ class Question:
         response = requests.get(api_url)
         return response.json()

+
+
     def parser_json(self):
         info = self.request_url()
         question_info = info.get('stat_status_pairs', None)
@@ -25,6 +27,7 @@ class Question:
             info_list.append(info_dict)

         df = DataFrame(info_list)
+
         df.to_csv('question.csv')
         print(df)
```

2.when i add --ignore-blank-lines

```
$ git diff --ignore-blank-lines
diff --git a/main.py b/main.py
index b847ef7..42c5793 100644
--- a/main.py
+++ b/main.py
@@ -15,6 +15,10 @@ class Question:
         response = requests.get(api_url)
         return response.json()

+
+    def test_func(self):
+        pass
+
     def parser_json(self):
         info = self.request_url()
         question_info = info.get('stat_status_pairs', None)
```

I want make libgit2 to support feature like this.

ethomson:
Hi @yuuri - thanks again for the patch.  I added https://github.com/libgit2/libgit2/commit/6b1f6e00bf7c91b3b5230e20ecb57a7d9792cee7 as a unit test on top!

yuuri:
@ethomson,Thank you very much! I will learn how to write test func by your code 😄

PR #5858, opened 2021-05-06
test: clean up memory leaks

ethomson:


lhchavez (approved):
i was about to be surprised that i had missed this one, but then i saw the big `#ifdef`

ethomson:
> i was about to be surprised that i had missed this one, but then i saw the big #ifdef

Unfortunately, the windows leakchecker doesn't fail the test run.  I'm not sure if there's an easy solution for that, I'll need to investigate how it works.  😣

PR #5861, opened 2021-05-12
[Branch] Branch upstream with format

lolgear:
It is a convenient accessor for merge branch name.

- [x] static `git_branch_upstream_with_format` has been added.
- [x] public `git_branch_upstream_merge` has been added.

### DIFF
--- a/src/branch.c
+++ b/src/branch.c
@@ -492,11 +492,13 @@ static int git_branch_upstream_with_format(git_buf *buf, git_repository *repo, c
 	return error;
 }
 
-int git_branch_upstream_remote(git_buf *buf, git_repository *repo, const char *refname) {
+int git_branch_upstream_remote(git_buf *buf, git_repository *repo, const char *refname)
+{
 	git_branch_upstream_with_format(buf, repo, refname, "branch.%s.remote");
 }
 
-int git_branch_upstream_merge(git_buf *buf, git_repository *repo, const char *refname) {
+int git_branch_upstream_merge(git_buf *buf, git_repository *repo, const char *refname)
+{
 	git_branch_upstream_with_format(buf, repo, refname, "branch.%s.merge");
 }
 
--- a/src/branch.c
+++ b/src/branch.c
@@ -492,6 +492,14 @@ static int git_branch_upstream_with_format(git_buf *buf, git_repository *repo, c
 	return error;
 }
 
+int git_branch_upstream_remote(git_buf *buf, git_repository *repo, const char *refname) {
+	git_branch_upstream_with_format(buf, repo, refname, "branch.%s.remote");
+}
+
+int git_branch_upstream_merge(git_buf *buf, git_repository *repo, const char *refname) {
+	git_branch_upstream_with_format(buf, repo, refname, "branch.%s.merge");
+}
+
 int git_branch_remote_name(git_buf *buf, git_repository *repo, const char *refname)
 {
 	git_strarray remote_list = {0};
--- a/src/branch.c
+++ b/src/branch.c
@@ -468,35 +468,10 @@ cleanup:
 	return error;
 }
 
-typedef enum {
-  GIT_BRANCH_UPSTREAM_FORMAT_REMOTE = 1,
-  GIT_BRANCH_UPSTREAM_FORMAT_MERGE = 2
-} git_branch_upstream_format_type_t;
-
-static const char* git_branch_upstream_format_string_for_id(git_branch_upstream_format_type_t id)
-{
-  switch (id) {
-  case GIT_BRANCH_UPSTREAM_FORMAT_REMOTE: return "branch.%s.remote";
-  case GIT_BRANCH_UPSTREAM_FORMAT_MERGE: return "branch.%s.merge";
-  default: return ""; // OK?
-  };
-}
-
-static const char* git_branch_upstream_format_name_for_id(git_branch_upstream_format_type_t id)
-{
-  switch (id) {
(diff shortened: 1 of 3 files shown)


src/branch.c
```diff
@@ -468,7 +468,7 @@ int git_branch_upstream_name(
 	return error;
 }
 
-int git_branch_upstream_remote(git_buf *buf, git_repository *repo, const char *refname)
+static int git_branch_upstream_with_format(git_buf *buf, git_repository *repo, const char *refname, const char *format)
```
ethomson, line 471: I like what you've done here.  But if we're making this more generic, we'll want to take another argument that indicates what we're fetching, and use it in the error message in this function (line 487).

src/branch.c
```diff
@@ -492,6 +492,14 @@ int git_branch_upstream_remote(git_buf *buf, git_repository *repo, const char *r
 	return error;
 }
 
+int git_branch_upstream_remote(git_buf *buf, git_repository *repo, const char *refname) {
```
ethomson, line 495: For functions, we would put the brace on a new line.

```suggestion
int git_branch_upstream_remote(git_buf *buf, git_repository *repo, const char *refname)
{
```

src/branch.c
```diff
@@ -492,6 +492,14 @@ int git_branch_upstream_remote(git_buf *buf, git_repository *repo, const char *r
...
+	git_branch_upstream_with_format(buf, repo, refname, "branch.%s.remote");
+}
+
+int git_branch_upstream_merge(git_buf *buf, git_repository *repo, const char *refname) {
```
ethomson, line 499:
```suggestion
int git_branch_upstream_merge(git_buf *buf, git_repository *repo, const char *refname)
{
```

ethomson (changes requested):
This looks good overall, just a few minor issues that I'd like to see addressed.  Also, could you add a test?

src/branch.c
lolgear: Could you suggest this argument?

I think that it is better to add enum for formats

```c
typedef enum {
  GIT_BRANCH_UPSTREAM_FORMAT_REMOTE = 1,
  GIT_BRANCH_UPSTREAM_FORMAT_MERGE = 2
} git_branch_upstream_format;

static const char* git_branch_upstream_format_for_id(git_branch_upstream_format id) {
  switch (id) {
  case GIT_BRANCH_UPSTREAM_FORMAT_REMOTE: return "branch.%s.remote";
  case GIT_BRANCH_UPSTREAM_FORMAT_MERGE: return "branch.%s.merge";
  default: return ""; // OK?
  };
}
```

And determine error message by this enum argument.

lolgear:
I guess this PR requires backward function.

```c
git_branch_name_from_upstream_format(git_buf *buf, git_repository *repo, const char *refname, const char **formats);
```

This method allows you to pass different formats and check if they exist in config.
First successful lookup will return local name of the branch.

What do you think?

src/branch.c
```diff
@@ -480,18 +505,28 @@ int git_branch_upstream_remote(git_buf *buf, git_repository *repo, const char *r
...
 }
 
+int git_branch_upstream_remote(git_buf *buf, git_repository *repo, const char *refname)
+{
+	return git_branch_upstream_with_format(buf, repo, refname, GIT_BRANCH_UPSTREAM_FORMAT_REMOTE);
+}
+
+int git_branch_upstream_merge(git_buf *buf, git_repository *repo, const char *refname)
+{
+	return git_branch_upstream_with_format(buf, repo, refname, GIT_BRANCH_UPSTREAM_FORMAT_MERGE);
```
ethomson, line 527: Instead of adding indirection with the enum and the `switch` and added functions, I'd just pass it in.  Much simpler to just have a function like:

```
static int git_branch_upstream_with_format(git_buf *buf, git_repository *repo, const char *refname, const char *config, const char *name)
```

and then have:

```
int git_branch_upstream_merge(git_buf *buf, git_repository *repo, const char *refname)
{
	return git_branch_upstream_with_format(buf, repo, refname, "branch.%s.merge", "merge");
}
```

No need to worry about match arms in the switch that aren't exercised, etc.

lolgear: Ok, changed.

lolgear:
@ethomson 
Could you check latest changes?

lolgear:
Could anybody check this PR?

lolgear:
@ethomson 
Could you check latest changes?
I added a test.

ethomson:
Thanks @lolgear!  I pushed one minor fix to reformat the test using tabs for indentation.  Appreciate your work here, thanks for the PR and for your patience!

lolgear:
No worries!
I hope you can release new version soon. ( Well, "soon" is one-two-three months... )

tests/refs/branches/upstream.c
```diff
@@ -71,6 +71,29 @@ void test_refs_branches_upstream__upstream_remote(void)
...
+
+	assert_config_entry_value(repository, "branch.test.remote", "test");
+	assert_config_entry_value(repository, "branch.test.merge", "refs/heads/master");
+
+	git_reference_free(branch);
+
+	/* check merge branch */
+    cl_git_pass(git_branch_upstream_merge(&buf, repository, "refs/heads/test"));
+    cl_assert_equal_s("refs/heads/master", buf.ptr);
+    git_buf_dispose(&buf);
```
lhchavez, line 94:
```suggestion
    git_buf_dispose(&buf);
    
    cl_git_sandbox_cleanup();
```

this is causing some leaks in master D:

PR #5886, opened 2021-05-14
[Transports]: SSH libssh2 error file handling has been added.

lolgear:
- [x] libssh2 error file handling has been added.

### DIFF
--- a/src/transports/ssh.c
+++ b/src/transports/ssh.c
@@ -362,10 +362,13 @@ static int _git_ssh_authenticate_session(
 		case GIT_CREDENTIAL_SSH_KEY: {
 			git_credential_ssh_key *c = (git_credential_ssh_key *)cred;
 
-			if (c->privatekey)
+			if (c->privatekey) {
 				rc = libssh2_userauth_publickey_fromfile(
 					session, c->username, c->publickey,
 					c->privatekey, c->passphrase);
+				if (rc == LIBSSH2_ERROR_FILE)
+					rc == LIBSSH2_ERROR_AUTHENTICATION_FAILED;
+			}
 			else
 				rc = ssh_agent_auth(session, c);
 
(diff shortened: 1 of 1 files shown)


src/transports/ssh.c
```diff
@@ -362,10 +362,13 @@ static int _git_ssh_authenticate_session(
 		case GIT_CREDENTIAL_SSH_KEY: {
 			git_credential_ssh_key *c = (git_credential_ssh_key *)cred;
 
-			if (c->privatekey)
+			if (c->privatekey) {
 				rc = libssh2_userauth_publickey_fromfile(
 					session, c->username, c->publickey,
 					c->privatekey, c->passphrase);
+				if (rc == LIBSSH2_ERROR_FILE)
+					rc == LIBSSH2_ERROR_AUTHENTICATION_FAILED;
```
tiennou, line 370: That's a comparison, not an assignment. Additionally, I don't understand what this fix is about, as a `LIBSSH2_ERROR_FILE` error would have been caught by the code at line 435, and wrapped into one of our own.

lolgear: Hmmm... Yes, you are right. 
Should be closed, I guess.
However, it returns `GIT_EAUTH` in this PR.

ethomson:
I don't understand libssh2 well enough to know when it would return `LIBSSH2_ERROR_FILE`, and whether it should be an auth failure or a proper failure, or what's happening now.  An issue describing the problems or - better - a test showing the problems would be a great first step.  In the meantime, I'm going to close this for now.

tiennou:
`LIBSSH2_ERROR_FILE` is usually returned when the key file is invalid (unreadable, not usable as per negotiation — think RSA-with-server-expecting-DSA mismatch, or PEM type not understood). As such, no authentication has taken place yet, so I don't think it makes sense to report it as an authentication error.

ethomson:
> `LIBSSH2_ERROR_FILE` is usually returned when the key file is invalid (unreadable, not usable as per negotiation — think RSA-with-server-expecting-DSA mismatch, or PEM type not understood). As such, no authentication has taken place yet, so I don't think it makes sense to report it as an authentication error.

That makes sense to me.  Thanks for the clarification.

Issue #5889, opened 2021-05-16
[Submodule] Missing retain function


lolgear:
I guess that I'm wrong and you could tell an API function for this, but...
If I would like to iterate over submodules and convert them to my objects, I need a retain function.

```c
int git_submodule_foreach(
	git_repository *repo,
	git_submodule_cb callback,
	void *payload) {/*...*/}
```

```c
void git_submodule_retain(git_submodule *sm)
{
	assert(sm);
	GIT_REFCOUNT_INC(sm);
}
```

### Reproduction steps

1. Iterate over submodules.
2. Callback will return an instance of submodule.
3. After iteration it will release instance of submodule.

### Expected behavior
Instance isn't freed. ( I have a function that retains this instance ).

### Actual behavior
Instance is freed.

### Version of libgit2 (release number or SHA1)
3532c5f49a7158096666764b099f872e779625c2

### Operating system(s) tested
macOS 11.3.1

ethomson:
I can see the value here. To match with prior art, this function should be `git_submodule_dup`. (It would simply refcount and return itself behind the scenes, not actually make a copy.)

lolgear:
@ethomson Could you check mentioned PR?

PR #5890, opened 2021-05-16
[Submodule] Git submodule dup

lolgear:
- [x] `git_submodule_dup` has been added.
- [x] #5889

### DIFF
--- a/include/git2/submodule.h
+++ b/include/git2/submodule.h
@@ -224,11 +224,13 @@ GIT_EXTERN(int) git_submodule_lookup(
 	const char *name);
 
 /**
- * Dup (retain) a submodule
+ * Create an in-memory copy of a submodule. The copy must be explicitly
+ * free'd or it will leak.
  *
- * @param submodule Submodule object
+ * @param out Pointer to store the copy of the submodule
+ * @param source Original tag to copy
  */
-GIT_EXTERN(git_submodule *) git_submodule_dup(git_submodule *submodule);
+GIT_EXTERN(int) git_submodule_dup(git_submodule **out, git_submodule *source);
 
 /**
  * Release a submodule
--- a/include/git2/submodule.h
+++ b/include/git2/submodule.h
@@ -224,6 +224,13 @@ GIT_EXTERN(int) git_submodule_lookup(
 	const char *name);
 
 /**
+ * Dup (retain) a submodule
+ *
+ * @param submodule Submodule object
+ */
+GIT_EXTERN(git_submodule *) git_submodule_dup(git_submodule *submodule);
+
+/**
  * Release a submodule
  *
  * @param submodule Submodule object
--- a/include/git2/submodule.h
+++ b/include/git2/submodule.h
@@ -227,7 +227,7 @@ GIT_EXTERN(int) git_submodule_lookup(
  * Create an in-memory copy of a submodule. The copy must be explicitly
  * free'd or it will leak.
  *
- * @param out Pointer to store the copy of the submodule. Cannot be NULL.
+ * @param out Pointer to store the copy of the submodule.
  * @param source Original submodule to copy.
  */
 GIT_EXTERN(int) git_submodule_dup(git_submodule **out, git_submodule *source);
--- a/include/git2/submodule.h
+++ b/include/git2/submodule.h
@@ -227,8 +227,8 @@ GIT_EXTERN(int) git_submodule_lookup(
  * Create an in-memory copy of a submodule. The copy must be explicitly
  * free'd or it will leak.
  *
- * @param out Pointer to store the copy of the submodule
- * @param source Original tag to copy
+ * @param out Pointer to store the copy of the submodule. Cannot be NULL.
+ * @param source Original submodule to copy.
  */
 GIT_EXTERN(int) git_submodule_dup(git_submodule **out, git_submodule *source);
 
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -1857,7 +1857,7 @@ static void submodule_release(git_submodule *sm)
 git_submodule* git_submodule_dup(git_submodule *sm)
 {
 	if (!sm)
-		return;
+		return NULL;
 	GIT_REFCOUNT_INC(sm);
 	return sm;
 }
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -1854,12 +1854,9 @@ static void submodule_release(git_submodule *sm)
 	git__free(sm);
 }
 
-git_submodule* git_submodule_dup(git_submodule *sm)
+int git_submodule_dup(git_submodule **out, git_submodule *source)
 {
-	if (!sm)
-		return NULL;
-	GIT_REFCOUNT_INC(sm);
-	return sm;
+	return git_object_dup((git_object **)out, (git_object *)source);
 }
 
 void git_submodule_free(git_submodule *sm)
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -1854,6 +1854,14 @@ static void submodule_release(git_submodule *sm)
 	git__free(sm);
 }
 
+git_submodule* git_submodule_dup(git_submodule *sm)
+{
+	if (!sm)
+		return;
+	GIT_REFCOUNT_INC(sm);
+	return sm;
+}
+
 void git_submodule_free(git_submodule *sm)
 {
 	if (!sm)
--- a/src/submodule.c
+++ b/src/submodule.c
@@ -1857,9 +1857,9 @@ static void submodule_release(git_submodule *sm)
 int git_submodule_dup(git_submodule **out, git_submodule *source)
 {
 	GIT_ASSERT_ARG(out);
+	GIT_ASSERT_ARG(source);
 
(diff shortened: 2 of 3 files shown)


src/submodule.c
```diff
@@ -1854,6 +1854,14 @@ static void submodule_release(git_submodule *sm)
 	git__free(sm);
 }
 
+git_submodule* git_submodule_dup(git_submodule *sm)
+{
+	if (!sm)
+		return;
```
neithernut, line 1860: I guess you meant to return either `sm` or, explicitly, `NULL`?

neithernut (changes requested):
Disclaimer: not a maintainer.

src/submodule.c
lolgear: Sure, which one is more appropriate (in current codebase)?

neithernut: I think that would be `NULL`.

ethomson: I would emulate the existing `dup` functions like `git_commit_dup` and return an `int` and use an out parameter.  This way we can change the implementation later in case we add some failure case.

lolgear: @ethomson 

Should I just use `git_object_dup` function?

```c
int git_submodule_dup(git_submodule **out, git_submodule *source) {
  return git_object_dup((git_object **)out, (git_object *)source);
}
```

src/submodule.c
```diff
@@ -1854,6 +1854,11 @@ static void submodule_release(git_submodule *sm)
 	git__free(sm);
 }
 
+int git_submodule_dup(git_submodule **out, git_submodule *source)
+{
+	return git_object_dup((git_object **)out, (git_object *)source);
```
tiennou, line 1859: That won't work — `git_submodule` is not a proper subtype/C "subclass" of `git_object`, so this will actually clobber memory in the underlying structure when incrementing the refcount.

IIRC your previous version (using `GIT_REFCOUNT_INC`) was correct but the function's prototype wasn't.

lolgear: I can't figure out how to adjust `*out` parameter with previous implementation.
Something like this?
```swift
int git_submodule_dup(git_submodule **out, git_submodule *source)
{
  if (!out) { return -1; }
  if (!source) { return -1; }
  *out = source;
  GIT_REFCOUNT_INC(*out);
  return 0;
}
```

lolgear: By the way, project has different styles in spaces/tabs and nullability check.
Is there any (current) guide to look at?

tiennou: There's docs/coding-style.md, which is supposed to be current and seems to be. I'd maybe add something about the assertion macros (which are new-ish) and when to use them — ideally for "impossible" or "API misuse" conditions.

src/submodule.c
```diff
@@ -1854,6 +1854,17 @@ static void submodule_release(git_submodule *sm)
 	git__free(sm);
 }
 
+int git_submodule_dup(git_submodule **out, git_submodule *source)
+{
+	GIT_ASSERT_ARG(out);
+
+	if (source != NULL)
```
tiennou, line 1861: Yeah… `if (source)` would have been fine here. Sorry, another project has "all conditions must be explicit", so…

lolgear: There are a lot of occurrences of `!= NULL` in this file. Should I change it to implicit check?

lolgear:
Could anybody check this PR?

include/git2/submodule.h
```diff
@@ -223,6 +223,15 @@ GIT_EXTERN(int) git_submodule_lookup(
 	git_repository *repo,
 	const char *name);
 
+/**
+ * Create an in-memory copy of a submodule. The copy must be explicitly
+ * free'd or it will leak.
+ *
+ * @param out Pointer to store the copy of the submodule
```
tiennou, line 230:
```suggestion
 * @param out Pointer to store the copy of the submodule. Cannot be NULL.
```

include/git2/submodule.h
```diff
@@ -223,6 +223,15 @@ GIT_EXTERN(int) git_submodule_lookup(
...
+ * @param source Original tag to copy
```
tiennou, line 231:
```suggestion
 * @param source Original submodule to copy.
```

tiennou (commented):
Just a few minor docblock tweaks (I skimmed the header initially). A quick test that exercises the new function and check that both submodules work and can be freed separately would be ❤️, just to be sure nothing regresses.

lolgear:
@tiennou 
Could you show me an example of this test?
This function is so common that I can't figure out whether I should add test in separate file or add it in on of the existed files.

tiennou:
Sure, I'd put it as `test_submodule_lookup__can_be_dupped`, which would `git_submodule_lookup()`, dup that, check that one of the "basic" properties work (name/owner/url, IIRC any one of them, as it's just about accessing the underlying storage), then free the first and assert whatever you checked at the beginning is still what we expect. In case anything goes wrong with memory in the future, the test would cause a segfault.

lolgear:
@tiennou 
Could you check latest changes?
I guess that it is done, but I can't run tests. 
A lot of errors at "configure" phase.

src/submodule.c
```diff
@@ -1854,6 +1854,17 @@ static void submodule_release(git_submodule *sm)
...
+		GIT_REFCOUNT_INC(source);
+
```
ethomson, line 1863: Is there a reason `source` is optional?  I would make it required.

```suggestion
	GIT_ASSERT_ARG(source);

	GIT_REFCOUNT_INC(source);

```

include/git2/submodule.h
```diff
@@ -223,6 +223,15 @@ GIT_EXTERN(int) git_submodule_lookup(
 	git_repository *repo,
 	const char *name);
 
+/**
+ * Create an in-memory copy of a submodule. The copy must be explicitly
+ * free'd or it will leak.
+ *
+ * @param out Pointer to store the copy of the submodule. Cannot be NULL.
```
ethomson, line 230: Our docs generally tell you when a param is optional, instead of when a param is required.

```suggestion
 * @param out Pointer to store the copy of the submodule.
```

tests/submodule/lookup.c
```diff
@@ -42,6 +42,46 @@ void test_submodule_lookup__simple_lookup(void)
 	assert_submodule_exists(g_repo, "sm_added_and_uncommited/");
 }
 
+void test_submodule_lookup__can_be_dupped(void)
+{
+	git_submodule *sm;
+	git_submodule *sm_duplicate;
+	const char *oid = "480095882d281ed676fe5b863569520e54a7d5c0";
+
+	// Check original
```
ethomson, line 51: We don't use `//` comments in this project...

```suggestion
	/* Check original */
```

tests/submodule/lookup.c
```diff
@@ -42,6 +42,46 @@ void test_submodule_lookup__simple_lookup(void)
...
+	cl_assert(git__suffixcmp(git_submodule_url(sm), "/submod2_target") == 0);
+
+	cl_assert(git_oid_streq(git_submodule_index_id(sm), oid) == 0);
+	cl_assert(git_oid_streq(git_submodule_head_id(sm), oid) == 0);
+	cl_assert(git_oid_streq(git_submodule_wd_id(sm), oid) == 0);
+
+	cl_assert(git_submodule_ignore(sm) == GIT_SUBMODULE_IGNORE_NONE);
+	cl_assert(git_submodule_update_strategy(sm) == GIT_SUBMODULE_UPDATE_CHECKOUT);
+
+	// Duplicate and free original
```
ethomson, line 65:
```suggestion
	/* Duplicate and free original */
```

tests/submodule/lookup.c
```diff
@@ -42,6 +42,46 @@ void test_submodule_lookup__simple_lookup(void)
...
+	cl_assert(git_submodule_dup(&sm_duplicate, sm) == 0);
+	git_submodule_free(sm);
+
+	// Check duplicate
```
ethomson, line 69:
```suggestion
	/* Check duplicate */
```

ethomson (changes requested):
Overall looks great, a few minor suggestions.

lolgear:
@ethomson Fixed.

tests/submodule/lookup.c
lolgear: @ethomson 
Fixed

tests/submodule/lookup.c
lolgear: @ethomson 
Fixed

tests/submodule/lookup.c
lolgear: @ethomson 
Fixed

include/git2/submodule.h
lolgear: @ethomson 
Fixed

src/submodule.c
lolgear: @ethomson 
Fixed

ethomson:
Thanks @lolgear!

ethomson:
Oops, I didn't hit merge - my bad!  Thanks again @lolgear!

PR #5893, opened 2021-05-19
Ensure that completely ignored diff is empty

ethomson:
When all contents of a diff are ignored, the resulting buffer should be empty but is instead malformed (includes a header but no hunks).

(This adds a failing test only at this point.)

### DIFF
--- a/src/libgit2/diff_print.c
+++ b/src/libgit2/diff_print.c
@@ -29,6 +29,7 @@ typedef struct {
 	const char *new_prefix;
 	uint32_t flags;
 	int id_strlen;
+	unsigned int sent_file_header;
 	git_oid_t oid_type;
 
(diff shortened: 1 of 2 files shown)


carlosmn:
It looks like this broke the tests at https://github.com/libgit2/rugged/blob/57257cbb7ea4b40debf5ec244e27ebffeaf40b16/test/diff_test.rb#L921 where we're just asking for the patch header. It looks like waiting to generate the patch header until we try to generate the content doesn't work if we never even wanted to generate any of the content but do want the header.

ethomson:
@carlosmn bleh. I suppose that instead of delaying sending the header until we send the first piece of content, we need to go find the first piece of content, and if there is some, enqueue it and deliver the header.

PR #5897, opened 2021-05-27
docs: fix incorrect comment marker

tiennou:
The broken ending marker would cause the rest of the file to be parsed as part of the comment.


lhchavez (approved):

ethomson:
Good catch!

Issue #5856, opened 2021-05-05
Invalid check for availability of futimens function for macOS platform


lolgear:
### Reproduction steps
Open src/unix/posix.h

### Expected behavior
Everything works fine.

### Actual behavior
Xcode warning: `futimens` is only available for macOS 10.13 or newer.
I build it for a project with minimum deployment target 10.10.

### Version of libgit2 (release number or SHA1)
Latest master.

### Operating system(s) tested
macOS 10.16
Xcode 12.5

ethomson:
What version of cmake?  It [claims to honor the deployment target version](https://gitlab.kitware.com/cmake/cmake/-/issues/17101) for `futimens`

lolgear:
I use Cmake 3.20.2 (Homebrew installed)
I run 

```sh
#!/bin/sh

export MACOSX_DEPLOYMENT_TARGET=10.10

rm -rf "xcode"
mkdir "xcode"
cd "xcode"
cmake -G "Xcode" ..
open "libgit2.xcodeproj"
```

Open Xcode 12.5 and build git2 target.
Xcode complains about `futimens`. 
Maybe project settings are wrong in Xcode.

lolgear:
Could you confirm that my test by Xcode project is relevant?

tiennou:
I can confirm that something is fishy, here's the [output](https://travis-ci.org/github/libgit2/objective-git/jobs/771143695#L1510) from a run of libgit2/objective-git), that builds with `SDKROOT\=/Applications/Xcode-12.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.0.sdk` and `MACOSX_DEPLOYMENT_TARGET\=10.10`.

lolgear:
I posted a question about this issue at [CMake](https://discourse.cmake.org/t/how-to-properly-configure-cmake-script-to-run-with-osx-deployment-target-10-10/3395) forums.

And at one moment I figured out that I do something wrong.
I compared generated `.c` file and a file that is proposed in a CMake [futimens](https://gitlab.kitware.com/cmake/cmake/-/issues/17101) issue.

They are different. Generated file doesn't include headers to search for.
So, it defines a function futimens without any context and any search.

I suggested to use `check_symbol_exists` with a provided header `sys/stat.h`.

```cmake
enable_warnings(error)
check_symbol_exists(futimens sys/stat.h HAVE_FUTIMENS)
[...]

PR #5898, opened 2021-05-28
[CMake] Check futimens function via symbol.

lolgear:
Addressed issue

- [x] #5856 

It seems that `check_function_exists` doesn't respect `MACOSX_DEPLOYMENT_TARGET`.
Example file with `futimens` test also doesn't contain any header and compiles with or without `futimens` function at `sys/stat.h`.

Script
```sh
export MACOSX_DEPLOYMENT_TARGET=10.10

rm -rf "xcode"
mkdir "xcode"
cd "xcode"
cmake --debug-trycompile -G "Xcode" .. 
```

( and add `message(FATAL_ERROR)` right after check function `futimens`. )

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -33,10 +33,13 @@ IF(ENABLE_TRACE)
 ENDIF()
 ADD_FEATURE_INFO(tracing GIT_TRACE "tracing support")
 
-CHECK_FUNCTION_EXISTS(futimens HAVE_FUTIMENS)
+enable_warnings(error)
+CHECK_SYMBOL_EXISTS(futimens sys/stat.h HAVE_FUTIMENS)
+# CHECK_FUNCTION_EXISTS(futimens HAVE_FUTIMENS)
 IF (HAVE_FUTIMENS)
 	SET(GIT_USE_FUTIMENS 1)
 ENDIF ()
+DISABLE_WARNINGS(error)
 
 CHECK_PROTOTYPE_DEFINITION(qsort_r
 	"void qsort_r(void *base, size_t nmemb, size_t size, void *thunk, int (*compar)(void *, const void *, const void *))"


lolgear:
Could anybody check this PR?

src/CMakeLists.txt
```diff
@@ -33,10 +33,13 @@ IF(ENABLE_TRACE)
 ENDIF()
 ADD_FEATURE_INFO(tracing GIT_TRACE "tracing support")
 
-CHECK_FUNCTION_EXISTS(futimens HAVE_FUTIMENS)
+enable_warnings(error)
```
ethomson, line 36: Is the `ENABLE_WARNINGS` / `DISABLE_WARNINGS` functional here?  Does `CHECK_SYMBOL_EXISTS` produce some warning in some cases that you're relying on failing the build?  Or is this leftover from debugging?  This isn't clear to me.

src/CMakeLists.txt
```diff
@@ -33,10 +33,13 @@ IF(ENABLE_TRACE)
...
+CHECK_SYMBOL_EXISTS(futimens sys/stat.h HAVE_FUTIMENS)
+# CHECK_FUNCTION_EXISTS(futimens HAVE_FUTIMENS)
```
ethomson, line 38: Let's remove this entirely, not comment it out.

lolgear: Could we rely on path "sys/stat.h" on all platforms? I guess it is better to add if (Apple) guard here.

src/CMakeLists.txt
lolgear: "Check symbol exists" functions generates only a warning and "futimens" function treated as found.
"enable_warnings(error)" forces cmake to treat futimens properly ( not found ).

Test script
```sh
#!/bin/sh

export MACOSX_DEPLOYMENT_TARGET=10.10

rm -rf "xcode"
mkdir "xcode"
cd "xcode"
cmake -G "Xcode" ..
```

ethomson: I see, that makes sense.

lolgear:
Could you also tell which style should be used?
I can't figure out which functions are uppercased and which are lowercased. ( they are equal, right? )

src/CMakeLists.txt
ethomson: Yes, I agree - I broke this out into a separate file in #5962 - can you please double-check that this still works for you?  It looks like it works for me... 🤔

lolgear: It works as expected.

PR #5900, opened 2021-05-30
Tolerate readlink size less than st_size

dtolnay:
This creates compatibility with systems that overestimate the symlink path's size, such as Android (https://issuetracker.google.com/189629152).

The GNU coreutils `stat` tool is similarly tolerant of `readlink` returning a path that is smaller than reported by `st_size`:

- https://github.com/coreutils/coreutils/blob/370c294018316f005bb259f51024497333d24ee0/src/stat.c#L1482
- https://github.com/coreutils/gnulib/blob/c45faf7f48bcbe57df3bbf8df81a016c5f15df55/lib/areadlink-with-size.c#L97-L99

Without this change, I observed that [Cargo](https://github.com/rust-lang/cargo) (which is built on libgit2) would tend to fail many commands with "failed to read symlink data for '...'" on Android in repositories containing symlinks.

### DIFF
--- a/src/odb.c
+++ b/src/odb.c
@@ -297,14 +297,15 @@ int git_odb__hashlink(git_oid *out, const char *path)
 		GIT_ERROR_CHECK_ALLOC(link_data);
 
 		read_len = p_readlink(path, link_data, size);
-		link_data[size] = '\0';
-		if (read_len != size) {
+		if (read_len == -1) {
 			git_error_set(GIT_ERROR_OS, "failed to read symlink data for '%s'", path);
 			git__free(link_data);
 			return -1;
 		}
+		GIT_ASSERT(read_len <= size);
+		link_data[read_len] = '\0';
 
-		result = git_odb_hash(out, link_data, size, GIT_OBJECT_BLOB);
+		result = git_odb_hash(out, link_data, read_len, GIT_OBJECT_BLOB);
 		git__free(link_data);
 	} else {
 		int fd = git_futils_open_ro(path);


src/odb.c
```diff
@@ -297,14 +297,9 @@ int git_odb__hashlink(git_oid *out, const char *path)
 		GIT_ERROR_CHECK_ALLOC(link_data);
 
 		read_len = p_readlink(path, link_data, size);
-		link_data[size] = '\0';
-		if (read_len != size) {
-			git_error_set(GIT_ERROR_OS, "failed to read symlink data for '%s'", path);
-			git__free(link_data);
-			return -1;
-		}
```
lhchavez, line 305: would it be possible to maintain this branch (with the `!= size` turned into `== -1` maybe)? that way the `link_data` does not leak in case of an error.

src/odb.c
```diff
@@ -297,14 +297,9 @@ int git_odb__hashlink(git_oid *out, const char *path)
 		GIT_ERROR_CHECK_ALLOC(link_data);
 
 		read_len = p_readlink(path, link_data, size);
-		link_data[size] = '\0';
```
lhchavez, line 300: is there a way to preserve this in some fashion? maybe `link_data[read_len] = '\0';` after the success check.

src/odb.c
dtolnay: Yep, good call. Fixed.

src/odb.c
dtolnay: We can, but it seems redundant &mdash; the buffer is being passed as `const void *` + `size_t` to git_odb_hash, which hashes that many bytes. It wouldn't be looking for the null terminator.

Is there a debugging rationale for having the null terminator anyway? Like maybe so that printing `link_data` in a debugger prints the right amount of data?

dtolnay: Added `link_data[read_len] = '\0';`.

lhchavez (approved):
lgtm! (although i can't merge changes)

src/odb.c
ethomson: > Is there a debugging rationale for having the null terminator anyway? Like maybe so that printing link_data in a debugger prints the right amount of data?

I'm lacking the historical context.  It's possible that we originally called a different function here that expected nul terminated string, or `git_odb_hash` has changed, etc.  I agree that this is unnecessary, but I also agree that having nul terminated strings in a debugger is helpful.  Given all of that, this seems like a nice fix.  If you wanted to attack the nul termination problem separately, I wouldn't argue, but I also don't think that it's a big problem either.  🤷‍♂️

PR #5908, opened 2021-06-03
Fix memory leak in git_smart__connect

punkymaniac:
The call to git_proxy_options_dup will replace the url pointer of the
proxy.
But if the url pointer is already set, the old address will be lost
forever and will never be free.

### DIFF
--- a/src/transports/smart.c
+++ b/src/transports/smart.c
@@ -226,8 +226,7 @@ static int git_smart__connect(
 	t->url = git__strdup(url);
 	GIT_ERROR_CHECK_ALLOC(t->url);
 
-	if (t->proxy.url)
-		git_proxy_options_clear(&t->proxy);
+	git_proxy_options_clear(&t->proxy);
 
 	if (git_proxy_options_dup(&t->proxy, proxy) < 0)
 		return -1;
--- a/src/transports/smart.c
+++ b/src/transports/smart.c
@@ -226,6 +226,9 @@ static int git_smart__connect(
 	t->url = git__strdup(url);
 	GIT_ERROR_CHECK_ALLOC(t->url);
 
+	if (t->proxy.url)
+		git_proxy_options_clear(&t->proxy);
+
 	if (git_proxy_options_dup(&t->proxy, proxy) < 0)
 		return -1;
 
--- a/tests/online/fetch.c
+++ b/tests/online/fetch.c
@@ -3,9 +3,19 @@
 static git_repository *_repo;
 static int counter;
 
+static char *_remote_proxy_scheme = NULL;
+static char *_remote_proxy_host = NULL;
+static char *_remote_proxy_user = NULL;
+static char *_remote_proxy_pass = NULL;
+
 void test_online_fetch__initialize(void)
 {
 	cl_git_pass(git_repository_init(&_repo, "./fetch", 0));
+
+    _remote_proxy_scheme = cl_getenv("GITTEST_REMOTE_PROXY_SCHEME");
+    _remote_proxy_host = cl_getenv("GITTEST_REMOTE_PROXY_HOST");
+    _remote_proxy_user = cl_getenv("GITTEST_REMOTE_PROXY_USER");
+    _remote_proxy_pass = cl_getenv("GITTEST_REMOTE_PROXY_PASS");
 }
 
 void test_online_fetch__cleanup(void)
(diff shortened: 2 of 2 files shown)


src/transports/smart.c
```diff
@@ -226,6 +226,9 @@ static int git_smart__connect(
 	t->url = git__strdup(url);
 	GIT_ERROR_CHECK_ALLOC(t->url);
 
+	if (t->proxy.url)
```
lhchavez, line 229: optional: the next line can be called unconditionally: `git_proxy_options_clear` takes care to reset the pointer to `NULL`, so it can be called multiple times without problems.

lhchavez (approved):
looks reasonable!

just out of curiosity, how does this happen? calling `t->connect(...)` more than once with the same `git_transport` object?

punkymaniac:
This is happen when I call `git_remote_fetch` after a `git_remote_connect` with a proxy configuration.
It doesn't happen without the proxy configuration.

src/transports/smart.c
punkymaniac: It's correct, I fix it.

lhchavez:
gotcha. do you mind adding a test?

```diff
diff --git a/tests/online/fetch.c b/tests/online/fetch.c
index f939a16b8..8029903a9 100644
--- a/tests/online/fetch.c
+++ b/tests/online/fetch.c
@@ -3,9 +3,19 @@
 static git_repository *_repo;
 static int counter;
 
+static char *_remote_proxy_scheme = NULL;
+static char *_remote_proxy_host = NULL;
+static char *_remote_proxy_user = NULL;
+static char *_remote_proxy_pass = NULL;
+
 void test_online_fetch__initialize(void)
 {
 	cl_git_pass(git_repository_init(&_repo, "./fetch", 0));
+
+	_remote_proxy_scheme = cl_getenv("GITTEST_REMOTE_PROXY_SCHEME");
+	_remote_proxy_host = cl_getenv("GITTEST_REMOTE_PROXY_HOST");
+	_remote_proxy_user = cl_getenv("GITTEST_REMOTE_PROXY_USER");
+	_remote_proxy_pass = cl_getenv("GITTEST_REMOTE_PROXY_PASS");
 }
 
 void test_online_fetch__cleanup(void)
@@ -14,6 +24,11 @@ void test_online_fetch__cleanup(void)
 	_repo = NULL;
 
 	cl_fixture_cleanup("./fetch");
+
+	git__free(_remote_proxy_scheme);
+	git__free(_remote_proxy_host);
+	git__free(_remote_proxy_user);
+	git__free(_remote_proxy_pass);
 }
 
 static int update_tips(const char *refname, const git_oid *a, const git_oid *b, void *data)
@@ -207,3 +222,29 @@ void test_online_fetch__twice(void)
 
 	git_remote_free(remote);
 }
+
+void test_online_fetch__proxy(void)
+{
+	git_remote *remote;
+	git_buf url = GIT_BUF_INIT;
+	git_fetch_options fetch_opts;
+
+	if (!_remote_proxy_host || !_remote_proxy_user || !_remote_proxy_pass)
+		cl_skip();
+
+	cl_git_pass(git_buf_printf(&url, "%s://%s:%s@%s/",
+		_remote_proxy_scheme ? _remote_proxy_scheme : "http",
+		_remote_proxy_user, _remote_proxy_pass, _remote_proxy_host));
+	printf("%s\n", url.ptr);
+
+	cl_git_pass(git_fetch_options_init(&fetch_opts, GIT_FETCH_OPTIONS_VERSION));
+	fetch_opts.proxy_opts.type = GIT_PROXY_SPECIFIED;
+	fetch_opts.proxy_opts.url = url.ptr;
+
+	cl_git_pass(git_remote_create(&remote, _repo, "test", "https://github.com/libgit2/TestGitRepository.git"));
+	cl_git_pass(git_remote_connect(remote, GIT_DIRECTION_FETCH, NULL, &fetch_opts.proxy_opts, NULL));
+	cl_git_pass(git_remote_fetch(remote, NULL, &fetch_opts, NULL));
+
+	git_remote_free(remote);
+	git_buf_dispose(&url);
+}
```

punkymaniac:
Of course you can.

ethomson:
> Of course you can.

I think that @lhchavez is politely requesting that you add the test to your PR.  I think that's a great idea, having this test in here would be great.  Thanks for working on this @punkymaniac!

punkymaniac:
Sorry, I misunderstand.
I will add the test.

lhchavez (approved):

ethomson:
Thanks @punkymaniac for the fix and @lhchavez for the review and test!

PR #5912, opened 2021-06-16
Syntax fixes

boretrk:
Fix two syntactical issues
1) Unnamed structs/unions aren't allowed
2) git_array_x(y)->z macro expanded to (NULL)->z on non-optimizing compilers

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -86,7 +86,7 @@ jobs:
           env:
             CC: gcc
             CMAKE_GENERATOR: Ninja
-            CMAKE_OPTIONS: -DUSE_HTTPS=OpenSSL -DREGEX_BACKEND=builtin -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON -DDEBUG_STRICT_ALLOC=ON
+            CMAKE_OPTIONS: -DUSE_HTTPS=OpenSSL -DREGEX_BACKEND=builtin -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON -DDEBUG_STRICT_ALLOC=ON -DDEBUG_ARRAY_CASTING=ON
           os: ubuntu-latest
         - # Xenial, GCC, mbedTLS
           container:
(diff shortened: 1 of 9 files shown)


lhchavez:
small request: can the "non-optimizing compilers" be named in the commit descriptions? that way it's going to be easier to know how to test these cases in the future.

given that it's going to be hard to maintain compatibility with such compilers (and they're not currently present in CI), I *think* it would be preferrable to make `git_array_last` and friends inline functions. that way they're going to be harder to (accidentally) misuse.

To avoid backsliding on the unnamed structs, we _could_ pass the [`-Wc11-extensions`](https://clang.llvm.org/docs/DiagnosticsReference.html#wc11-extensions) flag to a recent version of clang (maybe [here](https://github.com/libgit2/libgit2/blob/0d0150d8eb5fe5654af4172fe2f923f5d1317eb7/.github/workflows/main.yml#L120)).

boretrk:
I had a look at the possibility of replacing the array macros with inline functions.
They would either have to be type-specific versions for the places they are used or that they have to return void *.

Returning void * would trade a compile time error that only VBCC complains about (as far as we know) with the protection from miscasted structures so I think that option is way worse than a possibility that future changes might reintroduce this problem.

There are 21 different type arrays that are used in 100 different places so type specific inlines isn't really convenient either.
If we go that route we can prevent the specific misuse by having the macro use the same structure as many other functions where the first argument is where to store the value.
#define git_array_get(d, a, i) do {*(d) = ((i) < (a).size) ? &(a).ptr[(i)] : NULL;} while(0)
or
#define git_array_get(d, a, i) (((i) < (a).size) ? ((*(d) = &(a).ptr[(i)]), 0) : ((*(d) = NULL), -1))
to also return -1 on error.

boretrk:
Two other alternatives:

I could add a wrapper around the array macros
```
#ifdef GIT_TEST_ARRAY
#define git_array_wrapper(x) ((void *)(x))
#else
#define git_array_wrapper(x) (x)
#endif

#define git_array_get(a, i) git_array_wrapper(((i) < (a).size) ? &(a).ptr[(i)] : NULL)
```
Then you could test that the array isn't used in a way that isn't compatible with the datatype of NULL, but it requires compilation with GIT_TEST_ARRAY defined.
You still get the type checking when it isn't defined.

The fourth alternative would be to add a NULL-value to the array itself.
```
#define git_array_t(type) struct { type *ptr, *default; size_t size, asize; }

#define GIT_ARRAY_INIT { NULL, NULL, 0, 0 }

#define git_array_get(a, i) (((i) < (a).size) ? &(a).ptr[(i)] : (a).default)
```
This will return a NULL-value that has the same type as the array items and can be dereferenced in the same manner.
A possible benefit is that you can assign a.default element to arrays that you get when you try to access beyond the end of it.

@lhchavez Any particular preference?

boretrk:
OK, added optional array testing based on what @ethomson did in #5951
Could probably do with a review now.

boretrk:
OK, this is really three different issues. I'm going to split it into separate PR so that they can be reviewed and accepted on individual basis.

PR #5913, opened 2021-06-16
Make `FIND_PACKAGE(PythonInterp)` prefer `python3`

lhchavez:
This change makes it possible to find a python (version 3) interpreter
on systems where python2 and python3 coexist (where python2 is found as
`/usr/bin/python`).


ethomson:
Naive question: I thought clar worked in python2 still.  No?

lhchavez:
> Naive question: I thought clar worked in python2 still. No?

oh! i didn't see the `from __future__ import print_function` that's typically a good signal for the file being Python2, so i thought it didn't, but there's a `from __future__ import with_statement`

made this change just prefer python3 (instead of restrict to only being python3).

ethomson:
Yeah -- we tried to keep clar python 2 compatible, although I'm not sure how useful that actually is anymore.  I feel like that ship has probably sailed _but_ I also know that some people like ye olde systems.  🤷 

Thanks for the fix!

PR #5915, opened 2021-06-16
Consider files executable only if the user can execute them

novalis:
This is what git.git does, so we should follow suit.


novalis:
@carlosmn Can you review this when you get a chance?  Thanks.

ethomson:
Thanks for the fix!

PR #5917, opened 2021-06-18
docs: fix some missing includes that cause Docurium to error out

tiennou:
This adds a few missing includes so that Docurium doesn't skip some documentation when parsing.


ethomson:
Thanks @tiennou!

PR #5922, opened 2021-06-20
remote: handle error case

ethomson:
If `git_reference_resolve` fails with an actual error (not `GIT_ENOTFOUND`) it should not be ignored.


ethomson:
Indeed; the error handling here is odd and wide, but effective enough and not the problem I was seeing elsewhere.  Sorry for the spam!

PR #5924, opened 2021-06-22
Add failing test: push tag on blob

mckenfra:
If you tag a blob and use libgit2 to push that tag, it fails with error message: ```object is not a committish```

See https://github.com/libgit2/libgit2/issues/3595

This pull request adds a test that tries to do the above and fails.


ethomson:
Fixed via #6898

ethomson:
Thanks for the report!

PR #5926, opened 2021-06-23
config: fix included configs not refreshed more than once

Batchyx:
If an included config is refreshed twice, the second update is not taken
into account.

This is because the list of included files is cleared after re-reading
the new configuration, instead of being cleared before.

Fix it and add a test case to check for this bug.

This was probably introduced by commit a0dc3027f118119d3336db08251c646291f2c098

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -164,22 +164,26 @@ out:
 	return error;
 }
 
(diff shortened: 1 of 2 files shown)


ethomson:
Thanks for the fix and the very helpful test!  🎉

PR #5929, opened 2021-06-26
Define WINHTTP_NO_CLIENT_CERT_CONTEXT if needed

jacquesg:
Older versions of Mingw may not have this defined.


ethomson:
Thanks!

Issue #5556, opened 2020-06-12
Unable to clone on AWS-EFS (NFS) failed to truncate pack file Permission denied


Georgegriff:
## System information
  - OS/version/architecture: Kubernetes (Docker image on ubi8-minimal as non-root user) writing to shared mount that is AWS-EFS (which i believe uses nfs)

![image](https://user-images.githubusercontent.com/9056958/84484514-c6e01b80-ac92-11ea-83cf-2dfc3c4a17e8.png)

May be related too: 

https://github.com/JuliaLang/Pkg.jl/issues/1247
https://github.com/libgit2/git2go/issues/194

We have confirmed outside of the library in restricted we can:
- Write to shared mount
- Read from shared mount
- Delete from shared mount

More detail:

```bash
[2020-06-12T10:04:44.858] [ERROR] remotes - Error testing remote connection [Error: failed to truncate pack file '/mnt/assets/git/tmp/41091e79-81df-48f2-a5a0-d6a901b1
2e82/objects/pack/pack_git2_wXV3By': Permission denied] {
  errno: -1,                                             
  errorFunction: 'Remote.fetch'
} 

```

```bash
bash-4.4$ ls -la /mnt/assets/git/tmp/5fc54c2b-b118-4e97-a062-300d946a43ce/
total 36
drwxr-sr-x. 6 1000680000 2002 6144 Jun 12 10:06 .
drwxr-sr-x. 3 1000680000 2002 6144 Jun 12 10:06 ..
-rw-r--r--. 1 1000680000 2002   23 Jun 12 10:06 HEAD
-rw-r--r--. 1 1000680000 2002  182 Jun 12 10:06 config
-rw-r--r--. 1 1000680000 2002   73 Jun 12 10:06 description
drwxr-sr-x. 2 1000680000 2002 6144 Jun 12 10:06 hooks
drwxr-sr-x. 2 1000680000 2002 6144 Jun 12 10:06 info
drwxr-sr-x. 4 1000680000 2002 6144 Jun 12 10:06 objects
drwxr-sr-x. 4 1000680000 2002 6144 Jun 12 10:06 refs
bash-4.4$ ls -la /mnt/assets/git/tmp/5fc54c2b-b118-4e97-a062-300d946a43ce/objects/
total 16
drwxr-sr-x. 4 1000680000 2002 6144 Jun 12 10:06 .
drwxr-sr-x. 6 1000680000 2002 6144 Jun 12 10:06 ..
[...]

PR #5935, opened 2021-07-06
Fix pack file permission issue on NFS

pfitzseb:
This fixes https://github.com/libgit2/libgit2/issues/5556 by simply applying the patch in https://github.com/libgit2/libgit2/issues/5556#issuecomment-701404709.

I have absolutely no idea whether this is the right fix for the issue; if it is I'd be happy to figure out how to add a test and whatnot ;)

### DIFF
--- a/src/futils.c
+++ b/src/futils.c
@@ -24,10 +24,10 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
 
 int git_futils_mktmp(git_buf *path_out, const char *filename, mode_t mode)
 {
-	int fd;
-	mode_t mask;
-
-	p_umask(mask = p_umask(0));
+	int fd, count;
+	char *template;
+	uint64_t random;
+	const char *alphabet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_-";
 
 	git_buf_sets(path_out, filename);
 	git_buf_puts(path_out, "_git2_XXXXXX");
@@ -35,19 +35,33 @@ int git_futils_mktmp(git_buf *path_out, const char *filename, mode_t mode)
 	if (git_buf_oom(path_out))
 		return -1;
 
-	if ((fd = p_mkstemp(path_out->ptr)) < 0) {
-		git_error_set(GIT_ERROR_OS,
+	// This avoids a call to `mkstemp(3)`, because that way there is no need to
+	// call `chmod(2)`, which causes issues under some platforms (e.g. with NFS).
+	template = path_out->ptr + strlen(path_out->ptr) - 6;
+	random = (uint64_t)(git__timer() * 1.0E9);
+	for (count = 0; count < TMP_MAX; count++) {
+		uint64_t template_value = random;
+		int i;
+		// MMIX PRNG.
+		random = random * 0x5851F42D4C957F2Dull + 0x14057B7EF767814Full;
+
+		for (i = 0; i < 6; i++) {
+			template[i] = alphabet[template_value & 0x3f];
+			template_value >>= 6;
+		}
+
+		fd = p_open(path_out->ptr, O_RDWR | O_CREAT | O_EXCL, mode);
+		if (fd < 0) {
+			if (errno == EEXIST)
+					continue;
+			break;
+		}
+
+		return fd;
+	}
+	git_error_set(GIT_ERROR_OS,
 			"failed to create temporary file '%s'", path_out->ptr);
-		return -1;
-	}
-
-	if (p_chmod(path_out->ptr, (mode & ~mask))) {
-		git_error_set(GIT_ERROR_OS,
-			"failed to set permissions on file '%s'", path_out->ptr);
-		return -1;
-	}
-
-	return fd;
+	return -1;
 }
 
 int git_futils_creat_withpath(const char *path, const mode_t dirmode, const mode_t mode)
--- a/src/futils.c
+++ b/src/futils.c
@@ -35,14 +35,16 @@ int git_futils_mktmp(git_buf *path_out, const char *filename, mode_t mode)
 	if (git_buf_oom(path_out))
 		return -1;
 
-	// This avoids a call to `mkstemp(3)`, because that way there is no need to
-	// call `chmod(2)`, which causes issues under some platforms (e.g. with NFS).
+	/*
+	 * This avoids a call to `mkstemp(3)`, because that way there is no need to
+	 * call `chmod(2)`, which causes issues under some platforms (e.g. with NFS).
+	 */
 	template = path_out->ptr + strlen(path_out->ptr) - 6;
 	random = (uint64_t)(git__timer() * 1.0E9);
 	for (count = 0; count < TMP_MAX; count++) {
 		uint64_t template_value = random;
 		int i;
-		// MMIX PRNG.
+		/* MMIX PRNG. */
 		random = random * 0x5851F42D4C957F2Dull + 0x14057B7EF767814Full;
 
 		for (i = 0; i < 6; i++) {


lhchavez:
can you give this patch a quick try? https://github.com/libgit2/libgit2/issues/5556#issuecomment-875237533 if this works, we won't need to even involve `chmod(2)` and still keep the packfiles read-only in all platforms that don't have this issue.

pfitzseb:
Yep, that works fine. You wanna take it from here or should I update this PR?

lhchavez:
> Yep, that works fine. You wanna take it from here or should I update this PR?

your choice!

pfitzseb:
Ok, let me know what else needs doing here :) Not sure how to go about writing a test for this though...

src/futils.c
```diff
@@ -24,30 +24,44 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
...
 	if (git_buf_oom(path_out))
 		return -1;
 
-	if ((fd = p_mkstemp(path_out->ptr)) < 0) {
-		git_error_set(GIT_ERROR_OS,
-			"failed to create temporary file '%s'", path_out->ptr);
-		return -1;
-	}
+	// This avoids a call to `mkstemp(3)`, because that way there is no need to
+	// call `chmod(2)`, which causes issues under some platforms (e.g. with NFS).
```
lhchavez, line 39:
```suggestion
	/*
	 * This avoids a call to `mkstemp(3)`, because that way there is no need to
	 * call `chmod(2)`, which causes issues under some platforms (e.g. with NFS).
	 */
```

oh crud, this is totally my fault. I forgot that we use block comments in this project. 🙇

src/futils.c
```diff
@@ -24,30 +24,44 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
...
+	template = path_out->ptr + strlen(path_out->ptr) - 6;
+	random = (uint64_t)(git__timer() * 1.0E9);
+	for (count = 0; count < TMP_MAX; count++) {
+		uint64_t template_value = random;
+		int i;
+		// MMIX PRNG.
```
lhchavez, line 45:
```suggestion
		/* MMIX PRNG. */
```

lhchavez (commented):
> Ok, let me know what else needs doing here :) Not sure how to go about writing a test for this though...

I myself don't know how to test this, haha. The current set of tests pass for all platforms, which means that the patched code is still POSIX-compliant, which _probably_ means that the change is safe-ish.

We can't quite write a unit test for this (to prevent regressions) since it requires the underlying filesystem to be NFS. *maybe* having a nightly test for this is the right move (create a new Docker container based on the [focal](https://github.com/libgit2/libgit2/blob/main/ci/docker/focal) one where both the checkout and `/tmp` are localhost NFS-mounted directories), but that's out of the scope of this PR.

If you're interested in helping with the NFS nightly test, we can help in the Slack.

src/futils.c
pfitzseb: Well, I could've taken the time to look at the codebase myself instead of blindly copying your patch :P

boretrk:
I think it would be better to define a GIT_TMP_MAX with a specific value rather than loop on TMP_MAX here.

According to [linux.die.net](https://linux.die.net/man/3/mktemp) some systems have a TMP_MAX as low as 26 since they mainly use the pid for the "random" name and that is the reason to why mktemp() shouldn't be used.

[OpenGroup](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdio.h.html) says that "On XSI-conformant systems, the value of {TMP_MAX} is at least 10000." but otherwise it can be as low as 25.

boretrk:
> The current set of tests pass for all platforms, which means that the patched code is still POSIX-compliant

The current tests doesn't test that the library itself is POSIX-compliant, they test that the library works on some POSIX-comliant systems.
To test against that you'd have to test on a strict environment that treats anything that is implementation specific as an error or you'd have to add input validation to the POSIX macros in unix.h and compile specifically for testing. (Maybe only if _DEBUG is defined.)

This particular case is problematic since [OpenGroup](https://pubs.opengroup.org/onlinepubs/9699919799/functions/chmod.html) says:
> Any file descriptors currently open by any process on the file could possibly become invalid if the mode of the file is changed to a value which would deny access to that process. One situation where this could occur is on a stateless file system. This behavior will not occur in a conforming environment.

So part of the problem is that there are filesystems that aren't POSIX compliant and we want to support them.

I think older versions stated that the state of open filedescriptors are undefined after chmod.

It could be that fchmod() on the opened filedescriptor works too, it would be a slightly smaller change.

pfitzseb:
> Yep, that works fine. You wanna take it from here or should I update this PR?

Ok, turns out I didn't test the current implementation properly. I *does not* work, while creating the pack file with `644` prems does (https://github.com/libgit2/libgit2/issues/5556#issuecomment-701404709).

I also tried
```diff
diff --git a/src/futils.c b/src/futils.c
index a44820875..2934a3037 100644
--- a/src/futils.c
+++ b/src/futils.c
@@ -41,7 +41,7 @@ int git_futils_mktmp(git_buf *path_out, const char *filename, mode_t mode)
 		return -1;
 	}
 
-	if (p_chmod(path_out->ptr, (mode & ~mask))) {
+	if (p_fchmod(fd, (mode & ~mask))) {
 		git_error_set(GIT_ERROR_OS,
 			"failed to set permissions on file '%s'", path_out->ptr);
 		return -1;
diff --git a/src/unix/posix.h b/src/unix/posix.h
index 7b3325e78..dd5424a22 100644
--- a/src/unix/posix.h
+++ b/src/unix/posix.h
@@ -75,8 +75,10 @@ GIT_INLINE(int) p_fsync(int fd)
  */
 #ifdef __ANDROID__
 # define p_chmod(p,m) 0
+# define p_fchmod(p,m) 0
 #else
 # define p_chmod(p,m) chmod(p, m)
+# define p_fchmod(p,m) fchmod(p, m)
 #endif
 
 /* see win32/posix.h for explanation about why this exists */
```
, but that doesn't do the trick either.

boretrk:
Is this limited to packfiles or do you get the same problems when adding loose objects?
I think both call git_futils_mktmp() with a write protected filemode.

Maybe a solution would be to move the chmod from git_futils_mktmp and call it after the files are closed in git_filebuf_commit() and git_indexer_commit()?

Is it the chmod itself that fails or is it the writing after? If the chmod fails we could call it directly after git_futils_mktmp() anyway and depending on the result flag in the filebuf/indexer that the file should be write protected after closing instead.

pfitzseb:
Yeah, I can confirm that that patch fixes the issue.

boretrk:
It is mainly aimed at not doing chmod on open file descriptors, something that the patch from @lhchavez also takes care of.
That it solves the truncate issue is a result of postponing the chmod until the file is closed.

Not being able to truncate a write protected file while still being able to write to it with the open file descriptor seems more to be a bug in the NFS driver and would probably be better to fix on that end.

On the side of libgit I think the best solution would have been a configuration variable like core.symlinks that says if the filesystem supports file modification on write protected files so that the file can be opened with p_open as a write protected file where it is supported or be write protected after closing otherwise.
Unfortunately git doesn't have a config flag like that and it would be a bit odd to introduce it just on the libgit end.

Out of curiosity, how does regular git handle NFS?

ethomson:
> Out of curiosity, how does regular git handle NFS?

This is a great question.  Although loose objects and packfiles are `0444` by convention, is this a necessary thing?  Should we just best-effort try to set it `0444` and fallback if that fails?

boretrk:
> Should we just best-effort try to set it 0444 and fallback if that fails?

From what I could get from the problem description it would probably not be a good solution for NFS.
It seems like chmod works just fine and writing to the file afterwards too. It is only the truncate on a write protected file that fails.

A best effort attempt to lock the file would however solve an issue I have on AmigaOS3.x, but so would the patch from @lhchavez 
The benefit of that patch is that the file is write protected from the start without any gap between creation and protection.

The NFS issue could probably be solved if the indexer didn't write more data than necessary and didn't need the truncate to begin with, but I don't know enough about how the paging works to know if that is easy to implement.

pfitzseb:
So what's the way forward here? I'm afraid I'm too unfamiliar with all of this to make any sort of decision on which is the right fix.

boretrk:
> So what's the way forward here? I'm afraid I'm too unfamiliar with all of this to make any sort of decision on which is the right fix.

It sounds to me like you have ran into a [bug in the NFS server](https://marc.info/?l=openbsd-bugs&m=121663519107996)
If that is the case then I think everyone would benefit more from the bug being fixed on that end rather than libgit handling it.
Maybe you can try the sequence in that link to verify this?
```
cd /localdir
touch file
chmod a-w file
cp file /nfsdir
```

A possible workaround could be to check if the ftruncate fails and in that case temporarily unlock the file for the ftruncate and try again, maybe with a disclaimer that it is a workaround to handle NFS.

boretrk:
This was probably fixed by #6039, Did that one fix the problem for @pfitzseb too?

Would it still be of interest to apply https://github.com/libgit2/libgit2/issues/5556#issuecomment-875237533 ?
It makes sure that the file is protected from the start.
It could also help out on systems and/or filesystems that doesn't support changing mode on open files, for example AmigaOS.

ethomson:
Looks like this was fixed in #6039 - let me know if not.

PR #5936, opened 2021-07-06
git__timer: Limit ITimer usage to AmigaOS4

boretrk:
The ITimer interface is specific to AmigaOS4 and doesn't exist on 3.x and lower.


ethomson:
I don't have an Amiga, so I'll have to take your word for it @boretrk!  😁  Thanks for the fix!

PR #5938, opened 2021-07-07
Fix wrong time_t used in function

NattyNarwhal:
This function doesn't interoperate with any system functions that use the system `time_t`, but rather only works with the `git_time_t` type in libgit2, which could be a different width than the system one. Fixes a compile warning.


NattyNarwhal:
(I didn't touch any of the other ones because they all seemed to just work with system functions that return or expect `time_t`.)

lhchavez (approved):
just out of curiosity, what compiler / environment was warning about this?

NattyNarwhal:
....Visual C++ 6. I've found out what's actually not C89 in libgit2, at least :grimacing:

ethomson:
Thanks!  🎉

PR #5940, opened 2021-07-07
Patch documentation

punkymaniac:


include/git2/blame.h
```diff
@@ -108,41 +108,59 @@ GIT_EXTERN(int) git_blame_options_init(
...
 	size_t orig_start_line_number;
+
+	/**
+	 * The author of `orig_commit_id`. If `GIT_BLAME_USE_MAILMAP` has been
+	 * specified, it will contain the canonical real name and email address.
+	 */
 	git_signature *orig_signature;
 
+	/**
+	 * The 1 iff the hunk has been tracked to a boundary commit (the root, or
```
ethomson, line 161: Careful here - we try to keep an 80 column width (which means wrapping at around 76 characters) - especially for prose (not code) and we tend to assume a tab is 8 spaces.  This is getting a bit wide.

ethomson:
Thanks again for your work here - this is extremely valuable.  I took the liberty of adding documentation for `GIT_STATUS_OPT_INCLUDE_UNREADABLE` since I noticed that there wasn't any.

PR #5941, opened 2021-07-07
stdintification: use int64_t and INT64_C instead of long long

### LINKED ISSUE #5939: long long is not standards-compliant C89
There's some usage of "long long" in libgit2, which aims for C89. Thankfully, the usage is limited:

```
$ git grep "long long"
deps/pcre/CMakeLists.txt:CHECK_TYPE_SIZE("long long"             LONG_LONG)
deps/pcre/CMakeLists.txt:CHECK_TYPE_SIZE("unsigned long long"    UNSIGNED_LONG_LONG)
deps/zlib/zconf.h:#  define z_longlong long long
src/hash/sha1/generic.h:        unsigned long long size;
src/integer.h:GIT_INLINE(int) git__is_int(long long p)
src/integer.h:  return p == (long long)r;
src/khash.h:typedef unsigned long long khint64_t;
src/win32/w32_util.h:   long long winTime = ((long long)ft->dwHighDateTime << 32) + ft->dwLowDateTime;
src/win32/w32_util.h:   long long ticks = (tv.tv_sec * 10000000LL) +
```

The reason it's a problem is because it's actually a [GCC extension](http://gcc.gnu.org/onlinedocs/gcc-4.3.4/gcc/Long-Long.html) before C99. In my situation, this is was because I was compiling with a compiler that *didn't* support them. Said compiler is VC++6, which instead uses `_int64`. (Yes, I know it's VC++6. I have a branch for legacy compatibility (it already works on XP!); cleaning up what I can to make support for older Windows clean. The changes are surprisingly not that invasive already, I'll just need to `#ifdef` it and such.)

Another interesting thing is that the specifiers to force an integer literal to "long long" is different. `long long` uses `(U)LL`, but `_int64` uses `(u)i64`. This probably requires a macro to bridge, which can be made part of `stdint`. (In fact, isn't there `(U)INT64_C`?) The non-test usages of it are:

```
$ git grep -E "[0-9][Uu]?[Ll][Ll]"
deps/ntlmclient/ntlm.c: (!!((a) & 0x01ll) ^ !!((a) & 0x02ll) ^ \
deps/ntlmclient/ntlm.c:  !!((a) & 0x04ll) ^ !!((a) & 0x08ll) ^ \
deps/ntlmclient/ntlm.c:  !!((a) & 0x10ll) ^ !!((a) & 0x20ll) ^ \
deps/ntlmclient/ntlm.c:  !!((a) & 0x40ll) ^ !!((a) & 0x80ll))
src/commit_graph.c:     e->commit_time |= (e->generation & 0x3ull) << 32ull;
src/diff_xdiff.h:#define GIT_XDIFF_MAX_SIZE (1024LL * 1024 * 1023)
src/hashsig.c:#define HASHSIG_HASH_START        0x012345678ABCDEF0LL
src/mwindow.c:  ((1024 * 1024) * (sizeof(void*) >= 8 ? 8192ULL : 256UL))
src/util.h:#define MSB(x, bits) ((x) & (~0ULL << (bitsizeof(x) - (bits))))
src/win32/w32_util.h:   winTime -= 116444736000000000LL; /* Windows to Unix Epoch conversion */
src/win32/w32_util.h:   long long ticks = (tv.tv_sec * 10000000LL) +
src/win32/w32_util.h:           (tv.tv_usec * 10LL) + 116444736000000000LL;
```

One workaround is to just convert them all to `stdint.h`'s `(u)int64_t`, which can centralize the handling of the special types. Of course, *that's* still C99, but it can be shimmed with libgit2's substitute header for compilers that don't support C99. (Though there are problems including that header, at least with my VC++6 setup...)

NattyNarwhal:
While both are C99, the type and macro constant can be used in C89 with shims (i.e the one included). `long long` can't be (and it's a GCC extension), so using the stdint types will be preferable. In addition, we also need to use the macro for constants, because those can differ pre-C99. (As an example, MSVC uses `_int64` instead of `long long`, and `i64` as the constant suffix instead of `ll`.)

Not touched were things in `deps/`; the NTLM library isn't used on Windows and zlib seems to handle MSVC fine already. I assume the changes can be replicated upstreams and synced anyways.

Tested with gcc 11 on Fedora 34 x64. (As mentioned in the previous comment, I intend to use this with MSVC, but I'm still cleaning up my branch. :grimacing: )

Fixes #5939


src/mwindow.c
```diff
@@ -20,7 +20,7 @@
 		: 32 * 1024 * 1024)
 
 #define DEFAULT_MAPPED_LIMIT \
-	((1024 * 1024) * (sizeof(void*) >= 8 ? 8192ULL : 256UL))
+	((1024 * 1024) * (sizeof(void*) >= 8 ? UINT64_C(8192) : 256UL))
```
NattyNarwhal, line 23: Not sure what's up with that UL, which *is* valid, but may not be what was intended.

src/khash.h
```diff
@@ -137,11 +137,7 @@ typedef unsigned int khint32_t;
 typedef unsigned long khint32_t;
 #endif
 
-#if ULONG_MAX == ULLONG_MAX
-typedef unsigned long khint64_t;
-#else
-typedef unsigned long long khint64_t;
-#endif
+typedef int64_t khint64_t;
```
NattyNarwhal, line 140: I hope this simplification is OK.

src/mwindow.c
lhchavez: yeah, that `UL` looks very suspicious. maybe wrap it with `UINT64_C(...)` to avoid the implicit conversion in the conditional.

lhchavez (approved):
i'm in favor of this change \o/

since this change doesn't touch `deps/` it might not be feasible at the time, but maybe it's worth adding `-Wlong-long` to one of the gcc CI builds later _if_ it helps prevent backsliding (i.e. if it complains in all/some of the places that this PR touched). if that doesn't work, maybe adding https://clang.llvm.org/extra/clang-tidy/checks/google-runtime-int.html and https://github.com/libgit2/libgit2/pull/5618 will do the trick.

src/khash.h
ethomson: We've already deviated from `khash.h` and I think that it's basically unmaintained, so I'm not too worried about deviating from upstream.  Therefore, I'm definitely okay with dropping the indirection around whether `ULONG_MAX == ULLONG_MAX` - that definitely makes sense.  But does this need to be `uint64_t` instead of `int64_t`?  I haven't even glanced at the arithmetic in the file, so it may be obvious that it is safe to move it to signed, but that's what jumps out at me from a cursory glance.

ethomson:
I'm also :+1: on this, generally speaking.  But I also worry about the long-term drift.  Tackling these sort of issues one time is a frustrating experience (and thanks for the endeavor, @NattyNarwhal).  But having to revisit them every few months because we aren't catching them in CI is an added frustration that we should avoid.

lhchavez:
> I'm also  on this, generally speaking. But I also worry about the long-term drift. Tackling these sort of issues one time is a frustrating experience (and thanks for the endeavor, @NattyNarwhal). But having to revisit them every few months because we aren't catching them in CI is an added frustration that we should avoid.

the flags or `clang-tidy` should help with this!

src/khash.h
NattyNarwhal: Oops, you're right! I missed that.

tests/graph/commit_graph.c
```diff
@@ -66,7 +66,7 @@ void test_graph_commit_graph__parse_octopus_merge(void)
 	cl_git_pass(git_oid_fromstr(&id, "348f16ffaeb73f319a75cec5b16a0a47d2d5e27c"));
 	cl_assert_equal_oid(&e.tree_oid, &id);
 	cl_assert_equal_i(e.generation, 7);
-	cl_assert_equal_i(e.commit_time, 1447083009ull);
+	cl_assert_equal_i(e.commit_time, INT64_C(1447083009));
```
ethomson, line 69:
```suggestion
	cl_assert_equal_i(e.commit_time, UINT64_C(1447083009));
```

ethomson (commented):
I noticed one more sign change.  Otherwise, this is great, thanks!

PR #5943, opened 2021-07-11
Fix worktree iteration when repository has no common directory

### LINKED ISSUE #5851: "Assertion failed path && path_out" in src/path.c while trying to rename reference on custom refdb repo without commonpath set
Hello

### Reproduction steps
1. Try to use a custom backend implementing refdb without filesystem e.g.

```c
git_odb_backend *odb_backend = NULL;
git_refdb_backend *refdb_backend = NULL;
git_odb *odb = NULL;
git_refdb *refdb = NULL;
git_repository *repository = NULL;
git_init_custom_backends(&odb_backend, &refdb_backend, path); // magic
git_odb_new(&odb);
git_odb_add_backend(odb, odb_backend, 0);
git_repository_wrap_odb(&repository, odb);
git_refdb_new(&refdb, repository);
git_refdb_set_backend(refdb, refdb_backend);
git_repository_set_refdb(repository, refdb);
```

2. Implement `rename` method of `struct git_refdb_backend` for refdb_backend as required in documentation.
3. Try to rename a reference (deleting and recreating a reference pointing to the same oid works fine).

### Expected behavior
Reference gets renamed and all is fine.

### Actual behavior
Reference gets renamed (repository state is valid on subsequent launch), then 
assertion error happens on 
git_reference_rename -> git_repository_foreach_worktree -> git_repository_open(&worktree_repo, repo->commondir) -> git_repository_open_ext(repo_out, path, GIT_REPOSITORY_OPEN_NO_SEARCH, NULL) -> ... -> git_path_prettify()
execution path because `commondir` member is unset on custom repository.

cabfa3b30cfae24ec6d7d75c5cf052e052cbf131

the offending call seems to be added on https://github.com/libgit2/libgit2/commit/7216b048f5bd3cdad6416295d128f0173cb74533
despite the comment about supporting non-filesystem-based refdbs.

Windows 10

Potentially similar issue seems to be fixed at https://github.com/libgit2/libgit2/pull/4967, but it seems that `git_repository_set_path` function was never added and there is no valid way to set it?
Is there any way to circumvent this issue?

Any help appreciated.

kcsaul:
This is intended to fix #5851, which highlights an assertion within git_path_prettify that occurs during git_repository_foreach_worktree when it attempts to open the repository because repo->commondir is null.  This occurs in contexts where the repository has been created explicitly in conjunction with a custom odb/refdb backend.

git_repository_foreach_worktree has been changed to check when commondir is null, and in that case apply the callback operation to the repository supplied only, as it would appear there are no linked worktrees to iterate.

### DIFF
--- a/src/repository.c
+++ b/src/repository.c
@@ -2339,6 +2339,12 @@ int git_repository_foreach_worktree(git_repository *repo,
 	int error;
 	size_t i;
 
+	/* apply operation to repository supplied when commondir is empty, implying there's
+	 * no linked worktrees to iterate, which can occur when using custom odb/refdb
+	 */
+	if (!repo->commondir)
+		return cb(repo, payload);
+
 	if ((error = git_repository_open(&worktree_repo, repo->commondir)) < 0 ||
 	    (error = cb(worktree_repo, payload) != 0))
 		goto out;


kcsaul:
Based on a [prior comment](https://github.com/libgit2/libgit2/pull/4967#issuecomment-505868336), I'm under the impression there's insufficient unit tests for custom backend implementations which may have exposed this issue.  It would seem there's still an outstanding question as to how such unit tests might be implemented (unless some have been implemented since that comment was made?).  With that in mind, I'm assuming this fix could be merged without any additional unit tests being written.

ethomson:
> Based on a prior comment, I'm under the impression there's insufficient unit tests for custom backend implementations which may have exposed this issue. It would seem there's still an outstanding question as to how such unit tests might be implemented (unless some have been implemented since that comment was made?). With that in mind, I'm assuming this fix could be merged without any additional unit tests being written.

I agree that we don't necessarily need tests.  But this is the second or third such issue uncovered, which is disappointing.  I want to take a little look and see if we can't do something simple with tests here.  There's at least one relatively major consumer of libgit2 (aside from you) that's using custom db's entirely and would be facing such problems.

kcsaul:
@ethomson - I wonder if such problems could be avoided in future by moving functionality which interacts with each working directory behind a worktree backend interface.  This would enable libgit2 consumers using custom db's entirely to provide a custom worktree iterator and implement hooks for functionality that needs to interact with the working directory (i.e. checkout, diff).

Appreciate this suggestion will be perceived as a very big change in approach, but if feasible, it'd mitigate risk of such problems arising again, remove the need for defining any tests (for which you'd have to anticipate how a custom db may work), and it'd allow us to reuse worktree related functions in scenarios where the 'working directory' has been virtualized.

ethomson (approved):

PR #5945, opened 2021-07-15
git__timer: Allow compilation on systems without CLOCK_MONOTONIC

boretrk:
Makes usage of CLOCK_MONOTONIC conditional.

Functions that uses git__timer checks for negative time difference to handle clock synchronization when a non-monotonic timer is used.

According to [opengroup.org](https://pubs.opengroup.org/onlinepubs/9699919799/functions/gettimeofday.html) gettimeofday() have unspecified behaviour if tzp is set to anything but NULL.


ethomson:
LGTM.  Thanks!

PR #5948, opened 2021-07-17
Variadic arguments aren't in C89

### LINKED ISSUE #5947: Varadic macros are a C99 feature
`src/trace.h` uses varadic macros, which are a C99 feature:

```
$ gcc -std=c89 -Wall -pedantic test.c 
test.c:3:15: warning: anonymous variadic macros were introduced in C99 [-Wvariadic-macros]
    3 | #define MACRO(...) __VA_ARGS__
      |               ^~~
```

This should probably be changed to a wrapper function for non-C99 compilers.

NattyNarwhal:
This refactors this to a wrapper function, as well as changing the existing API have a va_list and ... version.

Fixes #5947

### DIFF
--- a/src/trace.h
+++ b/src/trace.h
@@ -21,31 +21,53 @@ struct git_trace_data {
 
 extern struct git_trace_data git_trace__data;
 
-GIT_INLINE(void) git_trace__write_fmt(
+GIT_INLINE(void) git_trace__vwrite_fmt(
 	git_trace_level_t level,
-	const char *fmt, ...)
+	const char *fmt, va_list ap)
 {
 	git_trace_cb callback = git_trace__data.callback;
 	git_buf message = GIT_BUF_INIT;
-	va_list ap;
 
-	va_start(ap, fmt);
 	git_buf_vprintf(&message, fmt, ap);
-	va_end(ap);
 
 	callback(level, git_buf_cstr(&message));
 
 	git_buf_dispose(&message);
 }
 
+GIT_INLINE(void) git_trace__write_fmt(
+	git_trace_level_t level,
+	const char *fmt, ...)
+{
+	va_list ap;
+
+	va_start(ap, fmt);
+	git_trace__vwrite_fmt(level, fmt, ap);
+	va_end(ap);
+}
+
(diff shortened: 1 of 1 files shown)


src/trace.h
```diff
@@ -21,31 +21,53 @@ struct git_trace_data {
...
+	va_list ap;
+
+	va_start(ap, fmt);
+	git_trace__vwrite_fmt(level, fmt, ap);
+	va_end(ap);
+}
+
 #define git_trace_level()		(git_trace__data.level)
+/* Varadic macros are a C99 feature */
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
```
ethomson, line 51: I'd say we should just drop this indirection.  At this point, `git_trace` isn't used within our own codebase, it only exists as a thing that allows you - a libgit2 contributor - to dig in to things within your code.

For background - originally, I wanted to make sure that this code was always inlined (since `inline` is just a hint) - to make sure that we would add as little overhead as possible when tracing was off.  The assumption being that we would ship a lot of tracing code within libgit2 itself.  We haven't - and I don't think that we have a lot of plans to, at least not yet.

Let's just make it a `GIT_INLINE` function always instead of trying to make it a macro if that's supported and an inline function if it's not.  This won't hurt anything today (anybody putting trace instrumentation into libgit2 is doing it because they have tracing on.  And we can always revisit if we find it to need optimization - that data will guide the optimization we do, I think.

NattyNarwhal: It actually *is* - `src/transports/httpclient.c`. If it was just user code only, I wouldn't have caught it :grimacing:

ethomson: Oh ha!  I used that when I was refactoring httpclient and forgot to pull it out.  Or I made a conscious decision to keep it.  I actually don't recall.  😁 

Either way, I think simpler is better and let's just use an inline function.  I went overboard with the macro magic in the first place and now we have an opportunity to right that wrong.  😁

ethomson:
I manually merged this - I think that the `write` function was unnecessary with the `vwrite` so I collapsed them into a single function.  Thanks for the fix!

Issue #5920, opened 2021-06-20
iter_load_loose_paths generates paths with empty segments


boretrk:
`iter_load_loose_paths()` appends a '/' on `backend->commonpath` in all cases.
`setup_namespace()` generates a `commonpath` without trailing '/' only if `repo->namespace` is set.

### Reproduction steps
Print paths from p_open/p_creat/p_stat/p_lstat or apply the following diff:
```
diff --git a/src/posix.c b/src/posix.c
index bf764ae6b..15430be0c 100644
--- a/src/posix.c
+++ b/src/posix.c
@@ -109,6 +109,9 @@ int p_open(const char *path, volatile int flags, ...)
 {
 	mode_t mode = 0;
 
+	if(strstr(path, "//")!=NULL)
+		fprintf(stderr, "Empty path segment: %s\n", path);
+
 	if (flags & O_CREAT) {
 		va_list arg_list;
 
@@ -122,6 +125,8 @@ int p_open(const char *path, volatile int flags, ...)
 
 int p_creat(const char *path, mode_t mode)
 {
+	if(strstr(path, "//")!=NULL)
+		fprintf(stderr, "Empty path segment: %s\n", path);
 	return open(path, O_WRONLY | O_CREAT | O_TRUNC | O_BINARY | O_CLOEXEC, mode);
 }
 
diff --git a/src/unix/posix.h b/src/unix/posix.h
index 7b3325e78..295ace0d9 100644
--- a/src/unix/posix.h
+++ b/src/unix/posix.h
@@ -22,8 +22,20 @@ typedef int GIT_SOCKET;
 
 #define p_lseek(f,n,w) lseek(f, n, w)
 #define p_fstat(f,b) fstat(f, b)
-#define p_lstat(p,b) lstat(p,b)
-#define p_stat(p,b) stat(p, b)
+
+GIT_INLINE(int) p_lstat(const char *path, struct stat *buf)
+{
+	if(strstr(path, "//")!=NULL)
+		fprintf(stderr, "Empty path segment: %s\n", path);
+	return lstat(path, buf);
+}
+
+GIT_INLINE(int) p_stat(const char *path, struct stat *buf)
+{
+	if(strstr(path, "//")!=NULL)
+		fprintf(stderr, "Empty path segment: %s\n", path);
+	return stat(path, buf);
+}
 
 #if defined(GIT_USE_STAT_MTIMESPEC)
 # define st_atime_nsec st_atimespec.tv_nsec
[...]

PR #5950, opened 2021-07-18
open: input validation for empty segments in path

boretrk:
As mentioned in issue #5920 there are a couple of places that tries to open paths containing empty path segments.
It would be nice if this could be caught in testing.
This PR uses the linker to overload the open() function and tests path for empty segment.
The idea is that we could add input validation to more POSIX calls to catch usage that is implementation defined or have unspecified behavior.

Maybe there is a more generic way to test for this?
We could check for _DEBUG and add the test to the p_open() function and let it fail if the path looks malformed.
Unfortunately that wouldn't have any impact on test cases where p_open() is expected to fail for other reasons but maybe that is a smaller problem.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -86,7 +86,7 @@ jobs:
           env:
             CC: gcc
             CMAKE_GENERATOR: Ninja
-            CMAKE_OPTIONS: -DUSE_HTTPS=OpenSSL -DREGEX_BACKEND=builtin -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON -DDEBUG_STRICT_ALLOC=ON
+            CMAKE_OPTIONS: -DUSE_HTTPS=OpenSSL -DREGEX_BACKEND=builtin -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=valgrind -DUSE_GSSAPI=ON -DDEBUG_STRICT_ALLOC=ON -DDEBUG_STRICT_OPEN=ON
           os: ubuntu-latest
         - # Xenial, GCC, mbedTLS
           container:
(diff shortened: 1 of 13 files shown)


ethomson:
> We could check for _DEBUG and add the test to the p_open() function and let it fail if the path looks malformed.

Yeah, I think that's probably a nice approach.  Maybe we shouldn't reuse `DEBUG`, and instead add a new `GIT_DEBUG_STRICT_OPEN`...

What do you think about...

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 655697979..04972d0fd 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@ OPTION(USE_GSSAPI			"Link with libgssapi for SPNEGO auth"			OFF)
 OPTION(USE_STANDALONE_FUZZERS		"Enable standalone fuzzers (compatible with gcc)"	OFF)
 OPTION(USE_LEAK_CHECKER			"Run tests with leak checker"				OFF)
 OPTION(DEBUG_POOL			"Enable debug pool allocator"				OFF)
+OPTION(DEBUG_STRICT_OPEN		"Enable strict validation on open(2)"			OFF)
 OPTION(ENABLE_WERROR			"Enable compilation with -Werror"			OFF)
 OPTION(USE_BUNDLED_ZLIB    		"Use the bundled version of zlib. Can be set to one of Bundled(ON)/Chromium. The Chromium option requires a x86_64 processor with SSE4.2 and CLMUL"			OFF)
    SET(USE_HTTP_PARSER			"" CACHE STRING "Specifies the HTTP Parser implementation; either system or builtin.")
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 4fde16d8a..1812bb740 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -6,6 +6,11 @@ IF(DEBUG_POOL)
 ENDIF()
 ADD_FEATURE_INFO(debugpool GIT_DEBUG_POOL "debug pool allocator")
 
+IF(DEBUG_STRICT_OPEN)
+	SET(GIT_DEBUG_STRICT_OPEN 1)
+ENDIF()
+ADD_FEATURE_INFO(debugopen GIT_DEBUG_STRICT_OPEN "debug strict open validation")
+
 INCLUDE(PkgBuildConfig)
 INCLUDE(SanitizeBool)
```

and using that as the invariant for `popen`?

ethomson:
(See what I did over in #5951 in case that illustrates better what I mean.)

ethomson:
Neat!  Thanks!

PR #5951, opened 2021-07-19
Optional stricter allocation checking (for `malloc(0)` cases)

ethomson:
Add `GIT_DEBUG_STRICT_ALLOC` to help identify problematic callers of allocation code that pass a `0` size to the allocators and then expect a non-`NULL` return.

When given a 0-size allocation, `malloc` _may_ return either a `NULL` _or_ a pointer that is not writeable.  Most systems return a non-`NULL` pointer; AIX is an outlier.  We should be able to cope with this AIXy behavior, so this adds an option to emulate it.


ethomson:
Fixes #5714

src/allocators/stdalloc.c
```diff
@@ -9,34 +9,56 @@
...
 {
-	void *ptr = calloc(nelem, elsize);
+	void *ptr;
 
 	GIT_UNUSED(file);
 	GIT_UNUSED(line);
 
-	if (!ptr) git_error_set_oom();
+#ifdef GIT_DEBUG_STRICT_ALLOC
+	if (!elsize)
```
lhchavez, line 38: the manpages claim that it can return `NULL` when either arg is zero:

```suggestion
	if (!elsize || !nelem)
```

ethomson: Aha, thanks, I meant to investigate that more closely.  😁

PR #5954, opened 2021-07-28
Fix typo in general.c

Crayon2000:


ethomson:
Good catch!  Thanks!  🎉

PR #5957, opened 2021-07-30
Fix one memory leak in master

lhchavez:
There was one test that wasn't correctly disposing of the repository.


ethomson:
Thanks for the fix!

PR #5962, opened 2021-07-30
cmake: extended futimens checking on macOS

ethomson:


lolgear:
Yes, it works as expected

lolgear:
#5856

PR #5968, opened 2021-08-01
Split the CI tests' steps

lhchavez:
Previously, if a CI run failed, one would need to go hunt for the step
that failed among the very long error logs.

This test now runs each test step in its own little `step`, so that the
error for a particular test is always displayed at the very end and is
easy to find.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -232,33 +232,189 @@ jobs:
       run: docker build -t ${{ env.docker-registry-container-sha }} -f ${{ env.dockerfile }} .
       working-directory: ${{ env.docker-config-path }}
       if: matrix.platform.container.name != '' && env.docker-container-exists != 'true'
-    - name: Build and test
+    - name: Build
       run: |
-        export GITTEST_NEGOTIATE_PASSWORD="${{ secrets.GITTEST_NEGOTIATE_PASSWORD }}"
-
+        mkdir build && chmod 0777 build
         if [ -n "${{ matrix.platform.container.name }}" ]; then
           docker run \
               --rm \
               --user libgit2:libgit2 \
               -v "$(pwd)/source:/home/libgit2/source" \
+              -v "$(pwd)/build:/home/libgit2/build" \
               -w /home/libgit2 \
               -e ASAN_SYMBOLIZER_PATH \
               -e CC \
               -e CFLAGS \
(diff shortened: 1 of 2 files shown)


ethomson:
> Previously, if a CI run failed, one would need to go hunt for the step that failed among the very long error logs.

Yes, this has been annoying me, too.  The fuzzers are super verbose and I feel like I have to keep scrolling and hunting through them.

We [went out of our way to design for this](https://github.com/libgit2/libgit2/blob/main/ci/test.sh#L54) - so that we wouldn't stop on the first test failure - but now I don't remember why.

I think that the easier fix would be to just stop when a test fails instead of breaking these out into multiple executions?

lhchavez:
> > Previously, if a CI run failed, one would need to go hunt for the step that failed among the very long error logs.
> 
> Yes, this has been annoying me, too. The fuzzers are super verbose and I feel like I have to keep scrolling and hunting through them.
> 
> We [went out of our way to design for this](https://github.com/libgit2/libgit2/blob/main/ci/test.sh#L54) - so that we wouldn't stop on the first test failure - but now I don't remember why.
> 
> I think that the easier fix would be to just stop when a test fails instead of breaking these out into multiple executions?

i have found that splitting them can also help gauge what's slow overall. and in the future even make things parallel :O 

one thing that can be done is to keep the tests going even in the face of failures.

ethomson:
Yeah, I see where you're going with this.  I added a5d6a5768f5537ef0a408cf51ce8c13742412041 as an interim step.  I'd like to clean up the YAML if we're going to do this.  I feel like it's already unwieldly and this isn't helping.  Some of the improvements in GitHub Actions should be able to help us out here.

lhchavez:
> Yeah, I see where you're going with this. I added [a5d6a57](https://github.com/libgit2/libgit2/commit/a5d6a5768f5537ef0a408cf51ce8c13742412041) as an interim step. I'd like to clean up the YAML if we're going to do this. I feel like it's already unwieldly and this isn't helping. Some of the improvements in GitHub Actions should be able to help us out here.

ooh are there improvements on Actions in the way? i couldn't find anything in the current docs to make this less verbose :(

ethomson:
I think that we've meaningfully improved our CI/CD experience and this is obsolete. LMK if I misunderstand.

Issue #5969, opened 2021-08-03
[CMake] ntlmclient. Headers for common crypto are excluded in Xcode target.


lolgear:
Too many words for simple 

https://github.com/libgit2/libgit2/blob/e65229ee972c113413eeca77853213352129bd47/deps/ntlmclient/CMakeLists.txt#L8

You may see that there is no clue of included headers at this common crypto condition branch.
Even more, there is no clue whether unicode is handled via `_builtin` or `_iconv`.

I don't understand what is going on (which headers are included or not), because target in Xcode doesn't contain any header but it works, because `HEADERS_SEARCH_PATH` flag is pointing to `deps/ntlmclient`.

ethomson:
In particular, they're getting defined in https://github.com/libgit2/libgit2/blob/main/cmake/SelectHTTPSBackend.cmake - the `ntlmclient` project is a bit odd.  I wrote it, basically for libgit2's use, but I wanted to make it able to standalone as well.  That's why it's in `deps/` but also just uses our configuration.

lolgear:
But I still don't understand which headers it uses.
I see a lot of includes in `ntlm.c` ( in Xcode project navigator ).
However, headers are missing.
I'm a bit confused why it doesn't contain headers in target. ( They are visible, of course, via HEADERS_SEARCH_PATH flag ), but they are not listed as headers in target.

For example, `http-parser` contains both `http_parser.c` and `http_parser.h`.
I expect the same for ntlmclient. ( All mandatory headers are included ).

Also I don't see a flag which switch `unicode_builtin` and `unicode_iconv` implementations in ntlmclient.

lolgear:
`ntlm.c` includes

```c
#include "ntlm.h"
#include "unicode.h"
#include "utf8.h"
#include "crypt.h"
#include "compat.h"
#include "util.h"
```

so these headers should be added to a Xcode target.

This change is a try to understand which headers are required by ntlmclient since it contains different implementations of crypto-backends.
Also this change will not affect ( should not affect ) a build and only a cosmetic change ( add proper headers to a target in CMakeLists.txt )

ethomson:
[...]

PR #5973, opened 2021-08-07
[CMake] ntlmclient. missing headers have been added.

lolgear:
This PR adds missing headers.

- [x] ntlmclient headers files have been added.
- [x] hash/sha1.h umbrella header has been added.
- [x] Addressed issue #5969
- [x] `deps/ntlmclient/crypt.h` missing `ntlmclient.h` include. `crypt.h` has functions definitions which uses `ntlm_client` symbol.


ethomson:
I pulled the ntlmclient dep changes in as part of #5974.

lolgear:
@ethomson 
Could you include "ntlmclient.h" into crypt.h?
struct ntlm_client is defined in ntlmclient and used in crypt.h. ( Yes, I know, it is pedantic ).

lolgear:
To be more pedantic, ntlmclient.h is an umbrella header for ntlmclient. So, it must contain all framework modules that are considered public...

PR #5978, opened 2021-08-08
amiga: use ';' as path list separator on AmigaOS

boretrk:
Like on Windows ':' is used for volume names in absolute paths.


ethomson:
Neat! I'm learning about Amiga.  😁

PR #5979, opened 2021-08-08
[Distribution] Swift package manager

lolgear:
This Draft PR contains Swift package manager distribution. ( kind of )

This repository has external dependencies ( libssh, libssl ) which should be properly handled by Cmake SPM generator ( if it would exist ).

However, it works and compiles a project ( with some changes ).

`.binaryTargets` should be removed or should be added by someone who would like to use this Package.swift ( yeah, you need to download a repository and you need to download a binary target to the same folder ).

I think some parts of this PR can be merged.
Do you need `Package.swift` for this library?
Well, it depends.

Addressed discussions/issues

- [x] #5965
- [x] #5971
- [x] #5975
- [x] #5977


lolgear:
Also SPM is still broken.
It can't find correct headers at all.
It can't exclude headers and headers search paths.

All these "required" changes should be fixed by SPM at one day.
I believe they will fix it.

Please, treat this PR as an exercise only. ( It compiles, yes, but I didn't check if it works on real project ).

lolgear:
@ethomson 
Could you check these two PRs?

#5979 
#5973 

I guess SPM PR can be merged as a sketch for people who would like to use this library via SPM ( as submodule but with SPM ).
Maybe it requires a different location ( templates/ directory )?
Then build.sh/install.sh script may copy this file into project root.

Second PR adds headers that are missing in targets.

Uncommon:
I would love to see this progress. I think it could be very useful. @lolgear is there anything more you think needs to be done for this to be fully ready? @ethomson (or some other official person) do you have any thoughts on this?

ethomson:
> I would love to see this progress. I think it could be very useful.

I don't get it yet.  😁 

Who would use this and when?  Swift developers who want to compile libgit2 and FFI against it?

Having some instructions for what this is and how to use it seems like it would be useful.  Maybe this is just a few sentences in the `README` if this is obvious enough for the people who care about it how to use it.  I don't know.

Here's one problem: I'm not going to maintain this.  This is a C project, not a Swift project, and few people use Xcode at all.  (And are happy to use cmake when we do.)  So putting this in a `contrib` directory seems like a better place to put it.  Is that even possible?

lolgear:
### Overview
This PR contains only a draft, but it works for different project. ( I hope it will be merged ).
Why is it useful? Because people don't want to use XcodeProject files, even if they are generated by cmake.
Cmake has its own limitations, because it doesn't use Xcode build settings and put everything in other linker or other C flags.
It also can't generate relative paths as far as I remember.
One big issue with it that it can't generate .xcframeworks ( new binary format from Apple ).
In these circumstances you have to either:
- keep project with fixed file paths ( should be relative so everyone could build your project ).
- generate libgit2 project in new setup ( new developer download project or you want to download clean version from remote ).

SPM solve many issues. 
- You don't need to generate SPM file each time you add new features in libgit2. 
- It gives you support for xcfraneworks for free. 
- It is small and easy-to-use by human and by difftool. ( XcodeProject is the worst file format for git ever ).
- It is flexible ( you can adjust settings in compile time, because it is Source file ).

So, SPM is a virtue...

But it is also very rigid about headers and sources.

It uses -fmodules flag and current libgit2 doesn't compile.
Also it generates warnings if folders layout is not ok.
It can't ignore headers and ( that is why ) ntlmclient is treated as misconfigured. ( ntlmclient.h doesn't contain other headers in the same directory and it is also has a name as target directory and is treated as umbrella header ).

I still think that SPM can be a good alternative to CMake. ( for C libraries in Apple ecosystem right now ).

### Current state

This PR is encountering some issues. The blocking issue is -fmodules which builds Darwin module with all headers. Here we have collisions of `struct entry`, it is defined in system header and in libgit2.

Other issues are solved by my previous PRs ( as far as I remember ).

It would be nice if @ethomson can apply patches to `struct entry` ( by adding unique prefixes ).

### Purpose and final thoughts

Future is... future is unknown.
I'm very upset about libssh2. It doesn't provide wrappers for CommonCrypto and Security frameworks. So, it is impossible to use libgit2 without libssh2 without libopenssl.

It is also impossible to keep this file in a root directory, because it requires binaries ( libssh2, libssl ).

Current SPM doesn't support ( am I missing something, because Xcode was released several days ago ) scripts out-of-the-box ( but it is possible to do some scripting in it ).

### Steps

Without patches for `struct entry` it is impossible ( very good word, isn't it? ) to use this library as SPM package.

1. Add patches for `struct entry` ( by adding prefixes )
2. Prepare a script which copies this file to root directory
3. Solve binary dependencies. ( via script ). It is possible to download dependencies to `binaries` directory.

mattt:
@lolgear For your use case, could you instead link to libgit2 in Swift Package Manager using a `systemLibrary` target? That approach has worked well for me in my Swift wrapper: https://github.com/swiftdocorg/git.

mbernson:
> For your use case, could you instead link to libgit2 in Swift Package Manager using a systemLibrary target? That approach has worked well for me in my Swift wrapper: https://github.com/swiftdocorg/git

@mattt I tried this approach, but it didn't seem to work for my Mac app that has git functionality. For it to be a portable `.app` bundle, it needs to contain all of its dependencies within. Unless I'm missing something here. :)

For now I've forked libgit2 and added @lolgear's Package.swift file with some modifications.

Perhaps in the future the [SwiftGit2 project](https://github.com/SwiftGit2/SwiftGit2) can fill this need by switching their library over to SPM, and maintaining a libgit2 fork.

ethomson:
I regret that this is one of those things that this is outside the scope of libgit2 itself. I’d encourage you to think of libgit2 as a dependency and maybe vendor it with a submodule like libgit2sharp does (for example). This moves the business logic of swift into a place where somebody who is a swift expert can maintain it. (And not in this repository, where the maintainers do not know swift and shouldn’t be responsible for it, for the sake of the swift users.)

PR #5983, opened 2021-08-12
clone: don't test remote url against filesystem

### LINKED ISSUE #5982: clone doesn't work if path exists locally
When cloning a repository libgit2 tests the given url against the filesystem to determine if a local clone should be done.
If the url also resolves to a path that exists locally then the clone fails.

### Reproduction steps
```
mkdir -p "git:/github.com/libgit2/TestGitRepository"
lg2 clone git://github.com/libgit2/TestGitRepository TestGitRepository
```

### Expected behavior
```
git clone git://github.com/libgit2/TestGitRepository TestGitRepository
Cloning into 'TestGitRepository'...
remote: Enumerating objects: 70, done.
remote: Total 70 (delta 0), reused 0 (delta 0), pack-reused 70
Receiving objects: 100% (70/70), 5.93 KiB | 2.96 MiB/s, done.
Resolving deltas: 100% (3/3), done.'
```

### Actual behavior
```
lg2 clone git://github.com/libgit2/TestGitRepository TestGitRepository

ERROR 6: could not find repository from 'git:/github.com/libgit2/TestGitRepository'
Bad news:
 could not find repository from 'git:/github.com/libgit2/TestGitRepository'

```

631ae5a67f17fd12c4c2b2406bec96fe61c473ff

Linux Mint 20.2

boretrk:
resolves #5982 

I removed the directory check in create_and_configure_origin
If the path given is to a bundle it will not be a directory but should probably have an absolute path anyway?

No test added, I'm not sure about the best approach to test this.
I suspect windows will act up on url's containing ':'

### DIFF
--- a/src/clone.c
+++ b/src/clone.c
@@ -317,6 +317,17 @@ static int default_remote_create(
 }
 
 /*
+ * Helper function to determine if a string actually is an URL and not a local
+ * path.
+ * Filesystems might misinterpret an URL as a local file so we should avoid
+ * accessing the filesystem with those if possible.
+ */
+GIT_INLINE(bool) url_has_scheme(const char *url)
+{
+	return strstr(url, "://") != NULL;
+}
+
+/*
  * submodules?
  */
 
@@ -332,8 +343,9 @@ static int create_and_configure_origin(
 	git_remote_create_cb remote_create = options->remote_cb;
 	void *payload = options->remote_cb_payload;
 
-	/* If the path exists and is a dir, the url should be the absolute path */
-	if (git_fs_path_root(url) < 0 && git_fs_path_exists(url) && git_fs_path_isdir(url)) {
+	/* If the path is local and exists it should be the absolute path. */
+	if (!url_has_scheme(url) && git_fs_path_root(url) < 0 &&
+	    git_fs_path_exists(url)) {
 		if (p_realpath(url, buf) == NULL)
 			return -1;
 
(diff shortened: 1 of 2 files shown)


boretrk:
The test from #5950 could be expanded to also include stat, but git_path_exists and git_path_isdir doesn't specifically check errno and have no way of returning invalid path rather than non-existing path so it doesn't seem like a good option here.

ethomson:
Yeah, Windows is going to be the tricky bit here for sure.  Let me do some testing over there.

boretrk:
An alternative to checking the url for "://" could be to split up `transport_find_fn()` so that it can be asked if it is a known url, local path or file:// style url, but I don't know if we really gain anything from that.

boretrk:
I figured that the way to test this is to add a "http:/"-folder in the current directory for the online cloning tests.
I'm very uncertain on the proper way to do this in the clar environment so if @ethomson could take a look at it it would be great.

We can't really test this on Windows systems, but then again, it will never occur on those either.

src/clone.c
```diff
@@ -332,8 +343,9 @@ static int create_and_configure_origin(
 	git_remote_create_cb remote_create = options->remote_cb;
 	void *payload = options->remote_cb_payload;
 
-	/* If the path exists and is a dir, the url should be the absolute path */
-	if (git_fs_path_root(url) < 0 && git_fs_path_exists(url) && git_fs_path_isdir(url)) {
+	/* If the path is local and exists it should be the absolute path */
+	if (!url_has_scheme(url) && git_fs_path_root(url) < 0 &&
+	    git_fs_path_exists(url)) {
```
boretrk, line 348: I intentionally dropped `git_fs_path_isdir(url)` here since cloning from bundles could be a thing in the future.

boretrk:
Restructured and added comments to make the flow clearer.
I think it should be good to go, if it is OK to use `git_futils_mkdir_r` in the test initialization that is.

PR #5990, opened 2021-08-15
Also look for a library with a 'd' suffix, for Windows

dfaure-kdab:


ethomson:
Is the `d` for debug?

dfaure-kdab:
Yes, libraries in debug mode have a 'd' suffix on Windows.
Passing both names as done here is very very common in cmake FindFoo files.

dfaure-kdab:
That's an excellent point. Looking for a debug library when building in release would be the famous unwanted mix-n-match on Windows. For some reason all the CMake code I know just looks for both names though...

Ah, I found how proper find modules in cmake itself do it.
They call find_library twice, once for a _RELEASE var (with release names) and once for a _DEBUG var (with debug names) and then call select_library_configurations() from SelectLibraryConfigurations.cmake to sort it out.

dfaure-kdab:
Ping? I think this is ready to go in now.

PR #5993, opened 2021-08-16
New GIT_EMISSING error

olsen232:
There are lots of different reasons a git object might not be found, and so far all are reported as GIT_ENOTFOUND.
This change gives one of these possible reasons a distinct error code - the case where we found a tree-entry describing some object, but we can't find the object. The error code is tentatively named GIT_EMISSING.

This shouldn't happen "normally" - "normally" the ODB is guaranteed not to contain such dangling pointers - but can happen for two reasons:
- the ODB is corrupt
- the repository was created by performing a partial clone, and still has missing objects that can be fetched from a promisor remote.

The second reason is probably the more interesting (most libgit2 clients probably don't really care about learning more about if or how the repo is corrupt, but they might want to better support partial clones) - and this reason could be usefully separated out into another code, GIT_EPROMISED. (Unfortunately we can't guarantee that the promisor remote will definitely have the missing object, but, it's pretty likely if the object is referenced but missing from a packfile that is marked as a promisor packfile).

Possible objection:
These error codes are special cases of GIT_ENOTFOUND, but unfortunately there's currently no way to add a subtype to an error. (GIT_EPROMISED would also ideally be a subtype of ENOTFOUND, perhaps even a sub-subtype of EMISSING). Clients who are currently checking for ENOTFOUND may have to update their code to also check for this new error code.

Related to old PR - https://github.com/libgit2/libgit2/pull/5776 - but that one has a faulty algorithm, so I just started over.

### DIFF
--- a/include/git2/errors.h
+++ b/include/git2/errors.h
@@ -58,11 +58,26 @@ typedef enum {
 	GIT_EMISMATCH       = -33,	/**< Hashsum mismatch in object */
 	GIT_EINDEXDIRTY     = -34,	/**< Unsaved changes in the index would be overwritten */
 	GIT_EAPPLYFAIL      = -35,	/**< Patch application failed */
-	GIT_EMISSING        = -36,  /**< Encountered tree-entry not backed by object */
 } git_error_code;
+
+
+/**
+ * Extra information for certain error codes.
+ * This extra information might not always be available - for example, more information
+ * is available when lookung up an object by path than when looking it up by OID.
+ */
+typedef enum {
+	GIT_NO_SUBCODE      = 0,
+
+	/** Subcodes for GIT_ENOTFOUND */
+	GIT_ENOSUCHPATH     = -3001, /**< Object at requested path DOES NOT exist */
+	GIT_EOBJECTMISSING  = -3002, /**< Object at requested path SHOULD exist, but is not in the ODB. */
+} git_error_subcode;
 
 /**
  * Structure to store extra details of the last error that occurred.
+ * The error code is not included since it is returned directly by the erroring function,
+ * but the error subcode and error class are included.
  *
  * This is kept on a per-thread basis if GIT_THREADS was defined when the
  * library was build, otherwise one is kept globally for the library
@@ -70,6 +85,7 @@ typedef enum {
 typedef struct {
 	char *message;
 	int klass;
+	int subcode;
 } git_error;
 
 /** Error classes */
--- a/src/commit.c
+++ b/src/commit.c
@@ -622,7 +622,7 @@ int git_commit_tree(git_tree **tree_out, const git_commit *commit)
 	GIT_ASSERT_ARG(commit);
 	error = git_tree_lookup(tree_out, commit->object.repo, &commit->tree_id);
 	return (error == GIT_ENOTFOUND) ?
-		git_odb__error_missing(&commit->tree_id) : error;
+		git_odb__error_notfound_missing(&commit->tree_id) : error;
 }
 
 const git_oid *git_commit_parent_id(
--- a/src/errors.c
+++ b/src/errors.c
@@ -26,18 +26,19 @@ static git_error g_git_uninitialized_error = {
 	GIT_ERROR_INVALID
 };
 
-static void set_error_from_buffer(int error_class)
+static void set_error_from_buffer(int error_class, int error_subcode)
 {
 	git_error *error = &GIT_THREADSTATE->error_t;
 	git_buf *buf = &GIT_THREADSTATE->error_buf;
 
 	error->message = buf->ptr;
 	error->klass = error_class;
+	error->subcode = error_subcode;
 
 	GIT_THREADSTATE->last_error = error;
 }
 
-static void set_error(int error_class, char *string)
+static void set_error(int error_class, int error_subcode, char *string)
 {
 	git_buf *buf = &GIT_THREADSTATE->error_buf;
 
@@ -47,7 +48,7 @@ static void set_error(int error_class, char *string)
 		git__free(string);
 	}
 
-	set_error_from_buffer(error_class);
+	set_error_from_buffer(error_class, error_subcode);
 }
 
 void git_error_set_oom(void)
@@ -98,7 +99,7 @@ void git_error_vset(int error_class, const char *fmt, va_list ap)
 	}
 
 	if (!git_buf_oom(buf))
-		set_error_from_buffer(error_class);
+		set_error_from_buffer(error_class, GIT_NO_SUBCODE);
 }
 
 int git_error_set_str(int error_class, const char *string)
@@ -113,14 +114,14 @@ int git_error_set_str(int error_class, const char *string)
 	if (git_buf_oom(buf))
 		return -1;
 
-	set_error_from_buffer(error_class);
+	set_error_from_buffer(error_class, GIT_NO_SUBCODE);
 	return 0;
 }
 
 void git_error_clear(void)
 {
 	if (GIT_THREADSTATE->last_error != NULL) {
-		set_error(0, NULL);
+		set_error(0, 0, NULL);
 		GIT_THREADSTATE->last_error = NULL;
 	}
 
@@ -153,12 +154,13 @@ int git_error_state_capture(git_error_state *state, int error_code)
 	state->oom = (error == &g_git_oom_error);
 
 	if (error) {
-		state->error_msg.klass = error->klass;
+		state->error.klass = error->klass;
+		state->error.subcode = error->subcode;
 
 		if (state->oom)
-			state->error_msg.message = g_git_oom_error.message;
+			state->error.message = g_git_oom_error.message;
 		else
-			state->error_msg.message = git_buf_detach(error_buf);
+			state->error.message = git_buf_detach(error_buf);
 	}
 
 	git_error_clear();
@@ -171,11 +173,11 @@ int git_error_state_restore(git_error_state *state)
 
 	git_error_clear();
 
-	if (state && state->error_msg.message) {
+	if (state && state->error.message) {
 		if (state->oom)
 			git_error_set_oom();
 		else
-			set_error(state->error_msg.klass, state->error_msg.message);
+			set_error(state->error.klass, state->error.subcode, state->error.message);
 
 		ret = state->error_code;
 		memset(state, 0, sizeof(git_error_state));
@@ -190,7 +192,7 @@ void git_error_state_free(git_error_state *state)
 		return;
 
 	if (!state->oom)
-		git__free(state->error_msg.message);
+		git__free(state->error.message);
 
 	memset(state, 0, sizeof(git_error_state));
 }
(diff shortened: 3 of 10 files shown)


rcoup:
> Clients who are currently checking for ENOTFOUND may have to update their code to also check for this new error code.

Libgit2 doesn't support working with partial clone repositories at the moment, so I'm not sure this is an _actual_ compatibility concern.

rcoup:
I think there might be some more cases that could result in GIT_EMISSING:

1. `git_commit_tree()` returns the root tree for a commit. Get a tree-less clone via `--filter=tree:0`
2. `git_reference_peel()` can return trees, commits, or blobs 
3. `git_tag_peel()` similarly
4. `git_object_peel()` similarly
4. `git_treewalk()` should probably handle EMISSING too? Currently it will abort the walk. To not break the existing API, the `mode` parameter could become a set of bitwise flags encompassing the existing `git_treewalk_mode` (eg: `MODE_PRE=0`, `MODE_POST=1`, `MISSING_OK=2`, ...) and if the flag is set the walk skips over missing entries. Or maybe this should become part of the EPROMISED work instead, and missing should always abort walks?

neithernut:
> These error codes are special cases of GIT_ENOTFOUND, but unfortunately there's currently no way to add a subtype to an error. (GIT_EPROMISED would also ideally be a subtype of ENOTFOUND, perhaps even a sub-subtype of EMISSING). Clients who are currently checking for ENOTFOUND may have to update their code to also check for this new error code.

Maybe it would make sense to introduce a macro or inlined function for checking whether an error falls under the "not found"-umbrella.

olsen232:
Addressing comments - @rcoup's suggested cases:
1. `git_commit_tree()` - done
2. `git_reference_peel()` - done (via `git_object_peel()`)
3. `git_tag_peel()` - done (via `git_object_peel()`)
(Note that case 2 and 3 only return GIT_EMISSING if they try to peel a commit into a tree and the tree wasn't there - if the ref or tag has a missing target, that is still reported as GIT_EPEEL. I'm not sure if missing targets for refs or tags is indicative of corruption or partial clones the way that a tree-entry or commit that is unbacked by an object is, so I don't know if it makes sense to change the error code - it might not be any more informative than GIT_EPEEL is right now).
4. `git_object_peel()` - done (via `git_commit_tree()`)

5. Not done - it's a good idea, but might wait on the official reviewer to tell me which option they prefer.

@neithernut re: error checking function - not done. It's a good idea, but might wait on the official reviewer to tell me which option they prefer.

olsen232:
I had another idea - making a separate "error-subtype" int or bitfield or whatever in the thread-local, same as is used to store the last error message. That way all these errors could stay as "GIT_ENOTFOUND" errors, but if you also queried the error subtype you would find out if they were missing / promised.

Other frameworks built on top of libgit2 which have exceptions instead of error codes would continue to represent these as "ObjectNotFoundException" or whatever, but those frameworks that chose to could add the the extra data as a field to the exception object - ie, most calling code would still simply catch ObjectNotFoundException, but some clients would then also check the caught exception to find out if the object that wasn't found was missing and/or promised.

Or, if these error codes are too tricky, I guess the error messages themselves convey this information in much the same way - you can get it if you really need it.

neithernut:
> Or, if these error codes are too tricky, I guess the error messages themselves convey this information in much the same way - you can get it if you really need it.

Error messages are usually directed at users. I doubt any binding author would willingly depend on a specific message (format).

Imo error codes are the way to go, a thread-local error-state if it can't be done without breaking existing users. But I'm not a maintainer, so...

ethomson:
I'm a little curious about the approach: instead of thinking about `GIT_ENOTFOUND` as a "supertype" of not found errors and then introducing `GIT_EMISSING` and `GIT_EPROMISED`, should we keep `GIT_ENOTFOUND` to mean "not found and not promised" and then add only `GIT_EPROMISED`?

I feel like we'll get to a place where the `GIT_ENOTFOUND` error code is just unused.

This also lets clients keep their existing behavior around not found items (except those that are promised, of course).

I wonder how breaking this is.  (And this may be why you were thinking through the bitfield?)  I would want to do a little more thinking about whether we think that there's a lot of users who are doing something with `GIT_ENOTFOUND` errors that we would break.  We _could_ make this an opt-in change in minor versions, and then introduce it "completely" in the next major.  But I'm leaning towards thinking that's not necessary.

(I'd love to hear arguments one way or the other. 😁 )

olsen232:
Trying to design the behaviour - sorry it got a bit long.

I will start by saying that these different types of errors have to be sort of best-effort or context-sensitive, unless we seriously redesign the libgit2 API. Looking up an object by its path involves knowing at least its immediate parent, whereas looking up an object by its ID does not involve knowing its immediate parent - nor is there a straight forward way of finding its parent(s) except for searching the entire ODB linearly. So if a client looks an object up by ID, and it's not found, it's always going to be GIT_ENOTFOUND since we don't know anything more than that about whether it "should" be there. As far as I know, finding a parent that references the requested object is the only way of making the distinction.

This has two implications for me - the first is that it means that the basic GIT_ENOTFOUND can never go away (there are also other reasons why it can't - it currently mostly means "the object you have asked for clearly doesn't exist at all" which won't go away either, no matter how we divide up PROMISED and MISSING). The second is that it means we should probably make the distinction between these types of errors subtle - making too big a deal of it would indicate that these error codes reliably describe totally distinct situations, which they do not.

I'm more and more inclined to return all these possible errors as GIT_ENOTFOUND, and then to have extra information you can query separately, which is filled in when it is available. With this approach, we can have as much or as little extra info as we want - we don't have to decide which new situations are important enough to break existing clients behaviour in the next major version. I've listed lots of possibilities below for how we could surface a little or a lot of information (I got a bit carried away).

1. Here is the most basic distinction, which you just suggested - 2 states, PROMISED=true/false:
- PROMISED if the object was looked up by path, and at some point the path went dead due to a TreeEntry not backed by an Object*, and that TreeEntry was found in a promisor packfile. (*Or a similar situation where a commit is not backed by a tree)

2. Here is the sort of distinction I was aiming for so far - 3 states, split PROMISED=false into MISSING and NO_EXTRA_INFO.
- MISSING if the object was looked up by path, and at some point the path went dead due to a TreeEntry not backed by an Object. 
- PROMISED if MISSING and that TreeEntry was found in a promisor packfile.
- Otherwise NO_EXTRA_INFO

3. Another possibility to add even more info - 4 states, split NO_EXTRA_INFO into NO_EXTRA_INFO and NONEXISTENT.
NO_EXTRA_INFO vs NONEXISTENT vs MISSING vs PROMISED
- NO_EXTRA_INFO if the object was looked up by OID and we didn't find it and don't know anything more about it.
- NONEXISTENT if we looked it up by path and found an object at that path definitely doesn't exist - there's no TreeEntry with the right name.
- MISSING and PROMISED as before (which still only work when looked up by path)

4. Another possibility to add even more info: - split PROMISED into OBJECT-IS-PROMISED and PARENT-IS-PROMISED:
- OBJECT-IS-PROMISED if the requested object's parent is present in the ODB and references the object, but the object itself is not. In this case we can be sure the object is promised. 
- PARENT-IS-PROMISED if its grandparent is present in the ODB and references its parent, but the parent is missing*. Without the immediate parent we don't actually know if the requested object even exists or not, but it should probably still be handled more like PROMISED than like NOTFOUND (ie, the client should now try to fetch it if it wants it).

Finally, to answer your first question - it feels to me as if information conveyed by GIT_EMISSING should still be surfaced if we have it (particularly if we are surfacing this info without changing existing error codes, so we can be as nuanced as we like). It's true that most clients will gain more value out of knowing if an object is promised or not, but in order to figure out if an object is promised, the information we gather might also tell us that the object is in that strange in-between state - not promised, not-non-existent, but also not present in the repo right now, for reasons unknown. Probably something has gone wrong. Most clients aren't gonna be able to do much about it but it since we went to the effort to find it out, and (when present) it indicates quite a different situation to "the object you have asked for simply doesn't exist", we may as well surface it. But, my feelings here aren't strong - basically any of the possibilities above seem okay for me.

ethomson:
> So if a client looks an object up by ID, and it's not found, it's always going to be GIT_ENOTFOUND since we don't know anything more than that about whether it "should" be there.

I don't understand this bit yet - wouldn't any object that is promised be in the promiser pack?  Are you saying that the promiser pack is only a subset of the promised object IDs?  How does querying by path vs object ID change this calculation?

I feel like you're saying that if my partial clone is of depth 1 of a repository, I will get a single tree (the repository root) and a promiser pack with only the object ids referenced by that tree?  And I would need to page in multiple trees as I recurse down to `/one/two/three/four`?

I didn't realize that, I thought the promisor pack was basically _all_ the object IDs in the remote - it would do the same walk as before it would just produce a set of complete objects and a set of promised objects.

olsen232:
Yeah sorry I wrote a little bit about this in https://github.com/libgit2/libgit2/pull/5776 but I should have been more clear about it in this PR which is actually implementing it.

See https://git-scm.com/docs/partial-clone/2.19.0#_handling_missing_objects - notably "This means that there is no need for the client to explicitly maintain an expensive-to-modify list of missing objects."

A promisor packfile is a packfile which contains objects which reference other objects, and those other objects are just missing which means "promised" (as far as I understand it). There's no special placeholders for the missing objects (except for the references themselves) and no lists of missing IDs.

Yes so if you have only a single commit-tree in your ODB, without its children, and your plan is to walk the entire commit-tree, then a wise client would refetch the commit up front, this time with the entire depth, before starting the walk. (An unwise client would just start walking and fetch every time it found a missing object, which would fetch every object one by one - and of course, libgit2 simply fails at the moment if you tell it to walk a tree that has missing objects). Or if you just want a single object at `one/two/three/four`, a wise client would request this particular object directly from the server (and an unwise client would make separate fetch requests for `one`, then `one/two`, then `one/two/three` then `one/two/three/four`).

I don't actually know much about how fetching missing objects work - it's outside the scope of this change. git will fetch missing objects for you if you do `git ls-tree SOME-PATH` or `git cat-file -p SOME-ID` which is probably going to be sufficient for my needs, but would probably be good to add libgit2 support for fetching missing objects at some stage.

olsen232:
Hi @ethomson - sorry if the above got a bit philosophical. Is there anything I can do to help this along, or is there anything in particular where you feel we don't know how to proceed?

I'm still thinking we can make this work with an extra error "subcode" that attentive-to-detail clients can query to find out what sort of "not-found" error they've encountered. Then divide the "not-found" errors into about 3 or 4 states. Most clients won't do anything with that data, but it's not a breaking change, and it's there for those that need it.

olsen232:
I took the liberty of adding another commit that conveys the extra information we have so far as an error "subcode" - I can revert it if it doesn't make sense. So far the extra subcodes are "OBJECTMISSING" - as in, we have a tree-entry, but we don't have the object that it refers to - and "NOSUCHPATH" - the user supplied a path but we know there is no tree-entry with that name. In a follow-up PR I would add the PROMISED one that would be basically an OBJECTMISSING except it happened in a promisor-packfile. What do you think @ethomson?

TAYLORKIRA:
hello sorry for the inconvenience but if you allow me I would like accompanied by something to what you have just said

PR #5994, opened 2021-08-17
Fix LIBGIT2_FILENAME not being passed to the resource compiler

jairbubbles:
`LIBGIT2_FILENAME` was not being passed to the resource compiler anymore since https://github.com/libgit2/libgit2/pull/5547

It is used when compiling native binaries for libgit2sharp and was detected in https://github.com/libgit2/libgit2sharp.nativebinaries/pull/114.


ethomson (approved):

ethomson:
Thanks for the fix!

PR #5995, opened 2021-08-17
ci: update centos builds

ethomson:


ethomson:
Our CI is tricky because the main build creates the containers that the nightly build uses - but only once it's merged into `main`.  (Until then, a PR build of the main CI can build containers that the remainder of that PR build will use, but then they're discarded so that somebody can't poison our actual main branch or nightly CI runs.)

This is usually a good situation, except when you want to change a container that you only run nightly, then you get into the fun routine of pushing endless yaml updates.

I'm going to get around this by pulling the centos 7 and 8 builds into the main build temporarily in this PR, so that we can go through the build the container / use the built container process instead of doing this churn in main.

ethomson:
_Aha_.  We're passing `-e PKG_CONFIG_PATH` to Docker.  This will pass through an empty environment variable `PKG_CONFIG_PATH=`, which clears the CentOS 7/8 environment variable `PKG_CONFIG_PATH=/usr/local/lib/pkgconfig`.

PR #5997, opened 2021-08-18
sha1dc: remove conditional for <sys/types.h>

boretrk:
src/hash/sha1/sha1dc/sha1.c is testing if \_\_unix\_\_ is defined before including <sys/types.h>
Since it is included unconditionally elsewhere it isn't necessary to include it conditionally and it prevents systems that doesn't define \_\_unix\_\_ from specifying endianess in the include files.


ethomson:
Seems ok to me!

PR #6005, opened 2021-08-25
C11 warnings

boretrk:
GCC and Clang defaults to allowing some C11 specific extensions even if C99 is specified
This PR enables warnings when compiling with Clang.

To enable warnings for GCC -Wc99-c11-compat has to be added, but this reveals a similar issue in win32/reparse.h. That code seems to be pulled in from the Windows 2000 SDK. Is it safe to fix this?

This is the same code as in #5912, but since it was 3 different issues baked into one and the array related test needed some more thought I decided to split it up in separate PR.


ethomson:
Appreciate you splitting it up - makes it easy to review.  👍

PR #6006, opened 2021-08-25
GCC C11 warnings

boretrk:
#6005 only added warnings when compiling with Clang.
This adds the same test for GCC.

I opted against using the DUMMYUNIONNAME from https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/ns-ntifs-_reparse_data_buffer?redirectedfrom=MSDN because I felt that it would clutter up the code.
Any arguments for sticking with the original naming?


ethomson:
> I opted against using the DUMMYUNIONNAME

💯

PR #6011, opened 2021-08-26
filter: filter drivers stop taking git_buf as user input

ethomson:
Deprecate the custom filter's `apply` function; this requires users to take the entire file, filter it, and then return the entire filtered contents.  This is appropriate for small files, but not large ones.  Further, it uses our `git_buf` for input and output; I would like to remove occurrences of `git_buf` being used as input to the library.  (Only as output to users that they can manage the lifecycle of.)  This will simplify our API surface and the `git_buf` usage significantly.

Note that I'm not _particularly_ bullish about the removal of the `apply` function.  It _is_ simpler in many ways; one option is to deprecate it and replace it with a function that works with some other buffer type (eg, `git_filter_buf`) that users can write to.  I would prefer targeted buffers in the `sys` headers to solve these problems instead of making `git_buf` a little bit public, a little bit private and overall very confusing.

Really I think that exceedingly few people have implemented a custom filter.  @palmin - I know you have - I'd love your input on this.

### DIFF
--- a/include/git2/sys/filter.h
+++ b/include/git2/sys/filter.h
@@ -167,16 +167,16 @@ typedef void GIT_CALLBACK(git_filter_shutdown_fn)(git_filter *self);
  *
  * The `payload` will be a pointer to a reference payload for the filter.
  * This will start as NULL, but `check` can assign to this pointer for
- * later use by the `apply` callback.  Note that the value should be heap
- * allocated (not stack), so that it doesn't go away before the `apply`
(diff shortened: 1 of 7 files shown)


palmin:
> Really I think that exceedingly few people have implemented a custom filter. @palmin - I know you have - I'd love your input on this.

@ethomson Thank you for taking my use-case into account.

The LFS filter in Working Copy works in a streaming manner. There are draconian memory constraints on iOS giving me little choice. I will not miss the `apply` function.

PR #6013, opened 2021-08-28
CMake. hash sha1 header has been added.

lolgear:
- [x] hash/sha1.h header has been added.

Previous PR #5973


lolgear:
@ethomson 
Could you check this PR before shipping new release?

cmake/SelectHashes.cmake
```diff
@@ -56,6 +56,8 @@ ELSE()
 	MESSAGE(FATAL_ERROR "Asked for unknown SHA1 backend: ${USE_SHA1}")
 ENDIF()
 
+FILE(GLOB SRC_SHA1_UMBRELLA "hash/sha1.h")
+list(APPEND SRC_SHA1 ${SRC_SHA1_UMBRELLA})
```
ethomson, line 60: I don't think that we need a variable here - I think that this should just be:

```suggestion
FILE(GLOB SRC_SHA1_UMBRELLA "hash/sha1.h")
list(APPEND SRC_SHA1 "hash/sha1.h")
```

ethomson:
Thanks!

PR #6014, opened 2021-08-29
Support for legacy Windows

NattyNarwhal:
(Upstreaming a branch I've had kicking for a month or two. I have not rebased yet; caveat emptor)

I've been [making a source control provider](https://visualgit.io/) compatible with [the APIs Microsoft has for it.](https://docs.microsoft.com/en-us/visualstudio/extensibility/source-control-plug-ins?view=vs-2019)

I've built up a patchset and this is a cleaned up version that should cleanly support the changes when possible. I can verify the gcc/amd64 Linux build isn't regressed by this. The changes to Vista+ should also be minimal, because everything is tidy ifdefs. (You should be able to test most of this by setting _WIN32_WINNT, which this PR makes changeable at the CMake level. Such a method could also be used to raise Windows versions supported if needed.) Many of the ifdefs handle the older SDK versions as well.

This should support Windows 2000 (primarily tested on XP) and Visual C++ 6 (with processor pack for align and 2003/02 Platform SDK); NT 4 should be trivial to support with 1. changing the version API checks 2. removing all flags for UTF-8 conversions [since those were added in 2000 SP4/XP](https://sporks.space/2021/08/27/utf-8-conversion-issues-on-legacy-windows/), but I have not added these just yet. Technically 9x is supportable, but it'd require Unicows to wrap wide APIs and I don't know if it supports \\?\.

Related, old VC++ does make for a good DeathStation 9000 in terms of no C89 or GNU extensions. I wonder if it might be possible to hack up CI with it considering some versions of it were freely available :thinking:

### DIFF
--- a/src/errors.h
+++ b/src/errors.h
@@ -37,8 +37,18 @@ GIT_INLINE(int) git_error_set_after_callback_function(
 }
 
 #ifdef GIT_WIN32
+#if defined(_MSC_VER) && _MSC_VER >= 1300
 #define git_error_set_after_callback(code) \
 	git_error_set_after_callback_function((code), __FUNCTION__)
+#else
+/* VC++6 doesn't support __FUNCTION__, fallback to __FILE__:__LINE__, with
+ * macro to handle the fact _LINE__ isn't a string
+ */
+#define __AS_STRING__(x) #x
+#define __LINESTRING__ __AS_STRING__(__LINE__)
+#define git_error_set_after_callback(code) \
+	git_error_set_after_callback_function((code), __FILE__ ":" __LINESTRING__)
+#endif
 #else
 #define git_error_set_after_callback(code) \
 	git_error_set_after_callback_function((code), __func__)
--- a/src/win32/posix_w32.c
+++ b/src/win32/posix_w32.c
@@ -666,6 +666,7 @@ static int getfinalpath_w(
 	git_win32_path dest,
 	const wchar_t *path)
 {
+#if _WIN32_WINNT >= 0x0600
 	HANDLE hFile;
 	DWORD dwChars;
 
@@ -687,6 +688,14 @@ static int getfinalpath_w(
 
 	/* The path may be delivered to us with a namespace prefix; remove */
 	return (int)git_win32_path_remove_namespace(dest, dwChars);
+#else
+	/* GetFinalPathNameByHandle is Vista only, though I think a reimpl is
+	 * possible. We could revert 307712613b77e8290a2c5d08ef3ae81c1e3139f3
+	 * which made it dynamically loaded, but it doesn't change the
+	 * semantics anyways; it'd just return -1. So let's do that for now.
+	 */
+	return -1;
+#endif
 }
 
 static int follow_and_lstat_link(git_win32_path path, struct stat* buf)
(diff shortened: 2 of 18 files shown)


src/errors.h
```diff
@@ -37,9 +37,19 @@ GIT_INLINE(int) git_error_set_after_callback_function(
...
 #else
+/* VC++6 doesn't support __FUNCTION__, fallback to __FILE__:__LINE__, with
+ * macro to handle the fact _LINE__ isn't a string
+ */
+#define __AS_STRING__(x) #x
+#define __LINESTRING__ __AS_STRING__(__LINE__)
+#define git_error_set_after_callback(code) \
+	git_error_set_after_callback_function((code), __FILE__ ":" __LINESTRING__)
+#endif
+#else
```
boretrk, line 52: Since \_\_func\_\_ doesn't exist in C89 I'd personally prefer to have this one used for more cases and only use \_\_func\_\_ if defined(\_\_STDC_VERSION\_\_) && (\_\_STDC_VERSION\_\_ >= 199901L)

Considering that the tests uses \_\_func\_\_ a lot it could also be worth considering making a `#define __func__ __FUNCTION__` in a include file somewhere for the windows case, and maybe even have your fallback as an alternative for C89 compliance.

src/win32/posix_w32.c
```diff
@@ -457,6 +459,10 @@ int p_symlink(const char *target, const char *path)
 		return -1;
 
 	return 0;
+#else
+	/* Symlinks are unavailable before Vista. */
+	return git_futils_fake_symlink(target, path);
```
ethomson, line 464: Let's not teach the posix emulation layer about git's notion of fake symlinks.  Can we move this up a level and have whatever "can we do symlinks on this platform" check we have return `false`?

ethomson (commented):
I must admit to being less than excited about the prospect of supporting ancient versions of Windows.  But if we have a maintainer for these ancient versions then I'm happier.  I don't want to keep this support up myself.

How would CI work do you think?  Do we just need an on MSVC on a new platform?  Or do we really need to build on some old operating system with an old compiler?  Coaxing the GitHub actions runner to run on win2k sounds tricky and is definitely a way that I don't want to spend my weekend.

NattyNarwhal:
I totally understand if you don't want to merge this; I myself thought it might not have been worth it (both originally doing, then submitting here), but considered it anyways since I tried to keep the support clean enough I could maintain it. Being C89 kinda invites the possibility though.

I'll see what I can do about cleaning this up further. I wasn't too sure on what was actually C89 or just a GNU extension in a trenchcoat or not.

Regarding CI: I suspect you can just run under WINE an old MSVC toolchain (at least 2003 on were freely available) with an older Platform SDK. Failing that, you could just probably use the existing Windows build infrastructure with `_WINNT_WIN32` set downlevel; it works *much* better than`_XOPEN_SOURCE` and friends.

src/win32/posix_w32.c
NattyNarwhal: I believe this is actually a revert-commit-turned-into-ifdef from before libgit2 was taught about real Win32 symlinks. It's understandable, though I'll have to think of the best way to handle this, since it'll probably be useful for porting to other environments as well.

PR #6015, opened 2021-08-29
tests: change comments to c89 style

boretrk:
A couple of c99 comments have snuck their way into tests/config/read.c and tests/remote/fetch.c


ethomson:
😢

PR #6016, opened 2021-08-29
Introduce `create_commit_cb`, deprecate `signing_cb`

ethomson:
Deprecate the `signing_cb` in `git_rebase`.  Instead, provide users the opportunity to extend commit creation in arbitrary ways, by providing a `create_commit_cb`.  If users wish to perform signing, they can do so in that callback.


ethomson:
😰  Good catch.

PR #6023, opened 2021-08-30
Add a .clang-format with our style

ethomson:
I started with the https://github.com/git/git `.clang-format`, with a few tweaks.  This still isn't quite right, but it's a little bit closer.  This feels like a thing that we can iterate on here.


lhchavez (approved):
any `.clang-format` file is great! :D

Issue #6028, opened 2021-09-03
Crash in threadstate_dispose (pointer being freed was not allocated)


arroz:
In some situations, `threadstate->error_t.message` will be pointing to `git_buf__initbuf`. When the `git__free` function runs, it causes the following crash: `malloc: *** error for object 0xwhatever: pointer being freed was not allocated`

### Reproduction steps

I have a test that reproduces this but cannot commit it, since it needs a HTTPS server with a real repository on it and a self-signed certificate, so badssl.com wont do. Here it is, for reference:

```
static int cert_accept_invalid_cert(git_cert *cert, int valid, const char* host, void *payload)
{
    GIT_UNUSED(cert); GIT_UNUSED(host); GIT_UNUSED(payload);
    
    cl_assert(!valid);
    
    return 0;
}

void *threadedAcceptInvalidCert(void *arg) {
    git_remote *remote;
    
    git_remote_callbacks callbacks = GIT_REMOTE_CALLBACKS_INIT;
    callbacks.certificate_check = cert_accept_invalid_cert;
    
    GIT_UNUSED(arg);
    
    cl_git_pass(git_remote_create_detached(&remote, "https://bad-cert-server/git/someRepo.git"));
    cl_git_pass(git_remote_connect(remote, GIT_DIRECTION_FETCH, &callbacks, NULL, NULL));
    cl_git_pass(git_remote_disconnect(remote));

    git_thread_exit(0);
}

void test_online_badssl__accept_invalid_cert_in_thread(void) {
    git_thread thread;
    int ret;
    
    if (!g_has_ssl)
        cl_skip();
    
    ret = git_thread_create(&thread, threadedAcceptInvalidCert, NULL);
    cl_assert(ret == 0);
    
    git_thread_join(&thread, NULL);
}
```

### Expected behavior

Should not crash. :)

### Actual behavior

Crashes with the following error:

```
libgit2_clar(66020,0x700005918000) malloc: *** error for object 0x1005453d8: pointer being freed was not allocated
libgit2_clar(66020,0x700005918000) malloc: *** set a breakpoint in malloc_error_break to debug
```

### Version of libgit2 (release number or SHA1)

v1.2.0 (but it was happening before as well)

### Operating system(s) tested

macOS

ethomson:
Thanks for the report and the fix!

PR #6029, opened 2021-09-03
#6028: Check if `threadstate->error_t.message` is not `git_buf__initbuf` before freeing.

arroz:
This follows the same principle as `buffer.c` where the same check is done before freeing the buffer. It fixes the crash described in #6028.

### DIFF
--- a/src/threadstate.c
+++ b/src/threadstate.c
@@ -36,7 +36,8 @@ static void threadstate_dispose(git_threadstate *threadstate)
 	if (!threadstate)
 		return;
 
-	git__free(threadstate->error_t.message);
+    if (threadstate->error_t.message != git_buf__initbuf)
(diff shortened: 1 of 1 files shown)


src/threadstate.c
```diff
@@ -36,7 +36,8 @@ static void threadstate_dispose(git_threadstate *threadstate)
 	if (!threadstate)
 		return;
 
-	git__free(threadstate->error_t.message);
+    if (threadstate->error_t.message != git_buf__initbuf)
+        git__free(threadstate->error_t.message);
```
ethomson, line 40: LGTM - just a quick whitespace fix:

```suggestion
	if (threadstate->error_t.message != git_buf__initbuf)
		git__free(threadstate->error_t.message);
```

ethomson:
Thanks for tracking this down!

PR #6031, opened 2021-09-04
Support custom git extensions

ethomson:
Allow users to specify additional repository extensions that they want to support.  For example, callers can specify that they support `preciousObjects` and then may open repositories that support `extensions.preciousObjects`.

Similarly, callers may opt out of supporting extensions that the library itself supports.

### DIFF
--- a/src/libgit2.c
+++ b/src/libgit2.c
@@ -52,6 +52,7 @@ static void libgit2_settings_global_shutdown(void)
 {
 	git__free(git__user_agent);
 	git__free(git__ssl_ciphers);
+	git_repository__free_extensions();
 }
 
 static int git_libgit2_settings_global_init(void)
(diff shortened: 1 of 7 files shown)


jairbubbles:
Thx @ethomson, I created a draft PR to test it in libgit2sharp: https://github.com/libgit2/libgit2sharp/pull/1908

I can confirm it works as expected as I was able to open repositories with "partialclone" extension!  🤘🏼

While writing the tests I was not sure about the behavior when setting the extensions two times in a row:

``` csharp
 GlobalSettings.SetExtensions(new[] { "!noop" });
 Assert.Empty(GlobalSettings.GetExtensions());

 GlobalSettings.SetExtensions(new[] { "partialclone", "newext" });
 Assert.Equal(new[] { "noop", "partialclone", "newext" }, GlobalSettings.GetExtensions()); // 👈🏼 I expected "noop" to remain disabled

```

ethomson:
Right - sorry, I expect you to call it one time with what you support.  (Subsequent calls will reset the supported extensions.)

I think that this is not super important in practice since most people will only call it once at startup and forget about it.  But it made testing easier in the library.

If you have strong opinions that I did this wrong then I'm definitely open to reconsidering.

jairbubbles:
> If you have strong opinions that I did this wrong then I'm definitely open to reconsidering.

@ethomson Not at all, I just mentioned it in case this behavior was not expected.

src/libgit2.c
```diff
@@ -367,6 +368,28 @@ int git_libgit2_opts(int key, ...)
 		git_odb__loose_priority = va_arg(ap, int);
 		break;
 
+	case GIT_OPT_SET_EXTENSIONS:
```
jairbubbles, line 371: (code indent looks incorrect)

jairbubbles:
🥳

jairbubbles:
@ethomson Do you plan to release a version that includes this in a near future?

ethomson:
@jairbubbles Yes.  I would like to do a v1.3.0 this weekend.

jairbubbles:
Sounds like a nice plan. I'll wait for 1.3.0 and update it in libGit2Sharp.

PR #6032, opened 2021-09-05
remote: Mark `git_remote_name_is_valid` as `GIT_EXTERN`

lhchavez:
This change makes `git_remote_name_is_valid` be part of the public
interface of the library. This is needed for other language bindings to
be able to find this symbol (like in git2go, when linking against
libgit2 dynamically).


ethomson:
Ooooops.  Sorry.  😥

PR #6035, opened 2021-09-05
filter: use a `git_oid` in filter options, not a pointer

ethomson:
Using a `git_oid *` in filter options was a mistake; it is a deviation
from our typical pattern, and callers in some languages that GC may need
very special treatment in order to pass both an options structure and a
pointer outside of it.


ethomson:
Closed in favor of a nicer, gentler #6067

PR #6037, opened 2021-09-06
checkout: make safe checkout the default

ethomson:
Make `GIT_CHECKOUT_SAFE` the default.  `NONE` is never what the user wants _by default_; people expect checkout to, well, check things out. Instead, it should be an opt-in "dry run" mode.

But, in fact, it should be a dry run and fire notifications.  So we keep `DRY_RUN` and remove the `NONE` mode entirely.

This removes some odd code in internal callers of `checkout` that takes a `git_checkout_options` and updates the mode to `SAFE`.  This is now unnecessary since everything has better defaults.


ethomson:
I made some updates to remove the API breaking parts of these changes; this should be ABI breaking only.

Issue #2338, opened 2014-05-09
Win32: can't clone to a remote (mounted) directory


kitbellew:
Steps to reproduce:
1. mount SMB network share (call it `z:`); in my case, I am using `\\psf\Home` under Parallels, which is how it accesses the home directory on the host MacOSX.
2. go to a subdirectory of `z:` (call it `z:\git2test`) and attempt to clone; the simplest is to run any of the clone CLAR tests (for instance, `-sclone::nonetwork::local`) with `CLAR_TMP` set to `z:\git2test` (must be full path).
3. fail with "failed to inflate packfile" or variation thereof.

My attempts to debug the problem suggest the issue lies in using `mmap()` together with `write()`. MSDN (http://msdn.microsoft.com/en-us/library/windows/desktop/aa366537(v=vs.85).aspx) specifies that:

"Although CreateFileMapping works with remote files, it does not keep them coherent. For example, if two computers both map a file as writable, and both change the same page, each computer only sees its own writes to the page. When the data gets updated on the disk, it is not merged.

A mapped file and a file that is accessed by using the input and output (I/O) functions (ReadFile and WriteFile) are not necessarily coherent."

Therefore, one solution is not to use CreateFileMapping in cases where data is being written simultaneously.

PR #6039, opened 2021-09-07
indexer: Avoid one `mmap(2)`/`munmap(2)` pair per `git_indexer_append` call

lhchavez:
This change makes `append_to_pack` completely rely on `p_pwrite` to do
all its I/O instead of splitting it between `p_pwrite` and a
`mmap(2)`/`munmap(2)`+`memcpy(3)`. This saves a good chunk of user CPU
time and avoids making two syscalls per round, but doesn't really cut
down a lot of wall time (~1% on cloning the
[git](https://github.com/git/git.git) repository).

Part of: #6038
Fixes: #5556

### DIFF
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -1307,8 +1307,11 @@ int git_indexer_commit(git_indexer *idx, git_indexer_progress *stats)
 
 #if !defined(NO_MMAP) && defined(GIT_WIN32)
 	/*
-	 * Truncate file to undo rounding up to next page_size in append_to_pack only
-	 * when mmap was used, to prevent failures in non-Windows remote filesystems.
+	 * Some non-Windows remote filesystems fail when truncating files if the
+	 * file permissions change after opening the file (done by p_mkstemp).
+	 *
+	 * Truncation is only needed when mmap is used to undo rounding up to next
+	 * page_size in append_to_pack.
 	 */
 	if (p_ftruncate(idx->pack->mwf.fd, idx->pack->mwf.size) < 0) {
 		git_error_set(GIT_ERROR_OS, "failed to truncate pack file '%s'", idx->pack->pack_name);
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -603,7 +603,6 @@ static void hash_partially(git_indexer *idx, const uint8_t *data, size_t size)
 
 static int write_at(git_indexer *idx, const void *data, off64_t offset, size_t size)
 {
-#ifdef NO_MMAP
 	size_t remaining_size = size;
 	const char *ptr = (const char *)data;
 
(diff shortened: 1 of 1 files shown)


boretrk:
It seems to me that this will make it possible to remove the `p_ftruncate` in `git_indexer_commit`
That should solve #5556 and close #5935
Does the preceding `git_mwindow_free_all` also have something to do with this?
```diff
diff --git a/src/indexer.c b/src/indexer.c
index d546888cc..eb56ccc92 100644
--- a/src/indexer.c
+++ b/src/indexer.c
@@ -1279,12 +1279,6 @@ int git_indexer_commit(git_indexer *idx, git_indexer_progress *stats)
        if (git_mwindow_free_all(&idx->pack->mwf) < 0)
                goto on_error;

-       /* Truncate file to undo rounding up to next page_size in append_to_pack */
-       if (p_ftruncate(idx->pack->mwf.fd, idx->pack->mwf.size) < 0) {
-               git_error_set(GIT_ERROR_OS, "failed to truncate pack file '%s'", idx->pack->pack_name);
-               return -1;
-       }
-
        if (idx->do_fsync && p_fsync(idx->pack->mwf.fd) < 0) {
                git_error_set(GIT_ERROR_OS, "failed to fsync packfile");
                goto on_error;
```
With that being the only use of ftruncate it can be removed from the posix layer and the tests too, but I don't see any harm in keeping it around in case it is needed in the future.

lhchavez:
> It seems to me that this will make it possible to remove the `p_ftruncate` in `git_indexer_commit`
> That should solve #5556 and close #5935
> Does the preceding `git_mwindow_free_all` also have something to do with this?
> 
> ```diff
> diff --git a/src/indexer.c b/src/indexer.c
> index d546888cc..eb56ccc92 100644
> --- a/src/indexer.c
> +++ b/src/indexer.c
> @@ -1279,12 +1279,6 @@ int git_indexer_commit(git_indexer *idx, git_indexer_progress *stats)
>         if (git_mwindow_free_all(&idx->pack->mwf) < 0)
>                 goto on_error;
> 
> -       /* Truncate file to undo rounding up to next page_size in append_to_pack */
> -       if (p_ftruncate(idx->pack->mwf.fd, idx->pack->mwf.size) < 0) {
> -               git_error_set(GIT_ERROR_OS, "failed to truncate pack file '%s'", idx->pack->pack_name);
> -               return -1;
> -       }
> -
>         if (idx->do_fsync && p_fsync(idx->pack->mwf.fd) < 0) {
>                 git_error_set(GIT_ERROR_OS, "failed to fsync packfile");
>                 goto on_error;
> ```
> 
> With that being the only use of ftruncate it can be removed from the posix layer and the tests too, but I don't see any harm in keeping it around in case it is needed in the future.

neat!

ethomson:
This seems reasonable to me.  @carlosmn any objections?

carlosmn:
This looks fine except this probably re-opens https://github.com/libgit2/libgit2/issues/2338 which is why we're so mmap-heavy. The issue here is that Windows may keep different views to a networked file for the mmap- and open-accessed versions of the file, so writing via `write()` is not reflected on the data `mmap()` is able to read.

IIUC this is a Windows quirk so you could disable the mmap stuff on non-Windows systems to keep it working there.

lhchavez:
> This looks fine except this probably re-opens #2338 which is why we're so mmap-heavy. The issue here is that Windows may keep different views to a networked file for the mmap- and open-accessed versions of the file, so writing via `write()` is not reflected on the data `mmap()` is able to read.
> 
> IIUC this is a Windows quirk so you could disable the mmap stuff on non-Windows systems to keep it working there.

turns out, supporting multiple platforms is hard! thanks for the pointer, i added a comment to explain the oddness of the code for future spelunkers.

i also tested this locally (Linux) with NFS and SMB and they both work! \o/

lhchavez:
And it also works in AWS-EFS per the other thread \o/².

ethomson:
Great fix, thanks again @lhchavez!

Issue #6040, opened 2021-09-08
GNUC typeof keyword isn't ISO C friendly


duncanthomson:
### Description

I'm trying to build libgit2 using clang using `-std=c99` and/or `-std=c11` and libstdc++. This setup defines the `__GNUC__` preprocessor token.

I'm seeing build errors like the following:
```
libgit2/src/path.h:673:2: warning: implicit declaration of function 'typeof' is invalid in C99 [-Wimplicit-function-declaration]
        GIT_UNUSED(path);                                                                                               
        ^                                       
libgit2/src/cc-compat.h:49:3: note: expanded from macro 'GIT_UNUSED'
                typeof(x) _unused __attribute__((unused));                     \
                ^                                

ibgit2/src/path.h:673:2: error: expected ';' after expression
libgit2/src/cc-compat.h:49:13: note: expanded from macro 'GIT_UNUSED'                                                   
                typeof(x) _unused __attribute__((unused));                     \                                        
                          ^                     
```

### Reproduction steps

These can be reproduced/replicated by attempting to compile a single .c file from the command line within the `libgit2` repo directory:
```
clang -D_GNU_SOURCE -DLIBGIT2_NO_FEATURES_H -Iinclude -Isrc -std=c99 -o ~/output/checkout.c.o -c src/checkout.c
```

### Expected behavior

libgit2 claims to be [_Written in a well-supported subset of C99_](https://libgit2.org/). Given that, I'd expect it to be C99 compliant and to build cleanly under clang with `-std=c99` and GNU C extensions.

### Actual behavior

The build errors can be circumvented by dropping the ISO C standard specification, e.g. removing `-std=c99`, but that doesn't seem right. [This gcc page](https://gcc.gnu.org/onlinedocs/gcc/Typeof.html) states:

> _If you are writing a header file that must work when included in ISO C programs, write `__typeof__` instead of `typeof`._

Consequently, could libgit2 consider switching to using `__typeof__` consistently, in place of `typeof`? At present, use of `typeof` is confined to a very small number of `#if defined(__GNUC__)` blocks. I'll raise a PR imminently.

### Version of libgit2 (release number or SHA1)

HEAD of master as of 8th Sept 2021: f1b89a201e9329e6df48f8d6cf320781570c936a

### Operating system(s) tested

Linux/ubuntu 20.04 LTS

duncanthomson:
Thank you for considering this issue and accepting/merging a fix. I'm closing this as "done".

PR #6041, opened 2021-09-08
Use __typeof__ GNUC keyword for ISO C compatibility

duncanthomson:
For issue: https://github.com/libgit2/libgit2/issues/6040

This PR converges on the use of `__typeof__` instead of `typeof` for ISO C compatibility with this GCC extension.

All uses of `typeof` or `__typeof__` are presently confined to `#if defined(__GNUC__) / #endif` blocks, which should limit the impact of this change.

Details:
- Remove the `GIT_TYPEOF` macro, since this is not used in the libgit2 project.
- Replace all occurrences of `typeof(x)` with `__typeof__(x)`.

### DIFF
--- a/src/cc-compat.h
+++ b/src/cc-compat.h
@@ -29,12 +29,6 @@
 #	endif
 #endif
 
-#ifdef __GNUC__
-#	define GIT_TYPEOF(x) (__typeof__(x))
-#else
-#	define GIT_TYPEOF(x)
-#endif
-
 #if defined(__GNUC__)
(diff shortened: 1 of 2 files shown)


lhchavez:
does clang with `-std=c99` build cleanly after this change? if so, maybe it _may_ be worth (in a separate PR) adding `-std=c99` to the xenial builders in `.github/workflows/main.yml`. that way we won't backslide ^^;;

ethomson:
Nice catch, thanks for the fix!

PR #6043, opened 2021-09-09
Fix config parsing for multiline with multiple quoted comment chars

### LINKED ISSUE #6019: config parsing can break on multiline with multiple quotes with multiple lines
### Reproduction steps

Create a `.gitconfig` with the following content and attempt to load it with libgit2:

```
[alias]
	myalias1 = log --pretty='case foo in\
*) sign=-S ";;" \
*) sign=%h ";;" \
esac \
'
```

### Expected behavior

libgit2 should be able to parse this into:
`log --pretty='case foo in*) sign=-S ;; *) sign=%h ;; esac '`

### Actual behavior

Fails with `invalid configuration key`

f4b473fa842a84147d8036a27bac5c026c44dad4

any

### Notes

This is another instance of #5459.  The issue seems to be in the same place, where `parse_multiline_variable` is using `!!in_quotes` which results in using the wrong value for whether or not the line is in quotes.

basile-henry:
Fixes #6019

The issue this PR fixes happened when parsing a git config with the following:

- A multiline variable
- Multiple quoted comment characters (on separate lines)

 NB: comment characters (`';'` or `'#'`) need to be quoted in order for them to be part of a string, otherwise they indicate the start of a comment. Comments are surprisingly permitted in the middle of a multiline variable.

I have added a small test that showcases the issue and then fixed it in a subsequent commit.
The fix is simply to keep the count of quotes in the multiline as an int instead of making it a bool before passing it to the `strip_comments` function. This seems to work just fine and all the tests now pass. I do wonder what was intended by the boolean "casting" :thinking:


ethomson:
> I do wonder what was intended by the boolean "casting" 🤔

I do, too!

Thanks, @basile-henry!

Issue #5670, opened 2020-10-14
soname versioning


jeroen:
If the ABI is stable perhaps there is no need to keep bumping the soname? I couldn't find anything about the soname policy in the readme file. A good example is [libcurl soname policy](https://github.com/curl/curl/blob/master/docs/libcurl/ABI.md#soname-bumps) (which hasn't changed their soname since 2006).

It would be really nice if the soname was stable, so that distributions don't have to rebuild all reverse dependencies when they update libgit2.

ethomson:
> If the ABI is stable perhaps there is no need to keep bumping the soname?

The ABI isn't stable.  This is an effort that we'll tackle in 2.0.  If we do that, we'll stabilize the soname.

carterbox:
Is it time to reconsider how the soname is chosen for this project? I read through some of [this discussion about API stability](https://github.com/libgit2/libgit2/discussions/5882?sort=old), and I got the feeling that the developers of libgit2 may not be aware that the ABI and API of a library may be versioned separately.

> The ABI isn't stable.

For packagers, I would say it's more important that the soname accurately reflects ABI breaks than the ABI be stable.

On unix platforms, the linker uses the soname to decide whether two libraries are ABI compatible, but also multiple soversions of a library can coexist on a single machine. So it's much easier for packagers if the soname actually tracks with ABI-breaks because then downstream packages which are built against different API versions of libgit2 can coexist in the same environment because they can each link to their own ABI-compatible libgit.

I guess what I am proposing is that libgit2 transition from using a single version to separately versioning the API and ABI where the major version of the soname changes whenever there is an ABI breaking change (which may not conincide with a breaking API change).

For example, based on my cursory reading of the release notes, the ABI/API versions could have been as such:

API / release | ABI / soname | notes
-- | -- | -- |
1.0.0 | 1.0.0 |
1.0.1 | 1.0.1 |
1.1.0 | 1.1.0 |
1.1.1 | 1.1.1 |
1.2.0 | 1.2.0 |
1.3.0 | 1.3.0 |
1.4.0 | 2.0.0 | ABI breaks, but API is backwards compatible |
1.4.1 | 2.0.2 |
1.4.2 | 2.0.2 |
1.5.0 | 2.1.0 |
1.6.1 | 2.2.1 |
1.7.0 | 3.0.0 | ABI breaks, but API is backwards compatible |
1.8.0 | 4.0.0 | ABI breaks, but API is backwards compatible |

[...]

PR #6051, opened 2021-09-13
Stabilize soname

stac47:
After the release 1.x has been released, we can perhaps expect an ABI
stabilization so that we don't need to recompile the libraries which
depends upon libgit2.

Running abidiff between 1.1.0 and 1.2.0 leads to no binary incompatible
changes.

This is already mentioned in #5670.

### DIFF
--- a/include/git2/version.h
+++ b/include/git2/version.h
@@ -13,6 +13,6 @@
 #define LIBGIT2_VER_REVISION 0
 #define LIBGIT2_VER_PATCH 0
 
-#define LIBGIT2_SOVERSION "1.2"
+#define LIBGIT2_SOVERSION "1"
 
 #endif
--- a/script/release.py
+++ b/script/release.py
@@ -37,7 +37,7 @@ def verify_version(version):
         'VER_MINOR':    [ str(version.minor), None ],
         'VER_REVISION': [ str(version.revision), None ],
         'VER_PATCH':    [ '0', None ],
-        'SOVERSION':    [ '"{}.{}"'.format(version.major, version.minor), None ],
+        'SOVERSION':    [ '"{}"'.format(version.major), None ],
     }
 
     # Parse CMakeLists
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -393,7 +393,7 @@ IDE_SPLIT_SOURCES(git2)
 
 if(SONAME)
 	set_target_properties(git2 PROPERTIES VERSION ${libgit2_VERSION})
-	set_target_properties(git2 PROPERTIES SOVERSION "${libgit2_VERSION_MAJOR}.${libgit2_VERSION_MINOR}")
+	set_target_properties(git2 PROPERTIES SOVERSION "${libgit2_VERSION_MAJOR}")
 	if(LIBGIT2_FILENAME)
 		target_compile_definitions(git2 PRIVATE LIBGIT2_FILENAME=\"${LIBGIT2_FILENAME}\")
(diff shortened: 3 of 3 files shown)


ethomson:
I think that there's two issues here.  Yes, I expect we'll do better about ABI stability beginning in 2.0.

But second, I'm surprised that `abidiff` isn't showing any differences.  We may not have changed function signatures, but we've _definitely_ changed the structures of some of the public `struct`s in ABI-changing ways.

include/git2/version.h
```diff
@@ -13,6 +13,6 @@
 #define LIBGIT2_VER_REVISION 0
 #define LIBGIT2_VER_PATCH 0
 
-#define LIBGIT2_SOVERSION "1.2"
+#define LIBGIT2_SOVERSION "1"
```
ethomson, line 16: Why would we change the soversion from `1.2` to `1`?  Even if we stop updating the soversion on new releases, we're not going to go backward are we?

stac47:
Hello,

Ah I was not aware that ABI stability was planned for the 2.x series. I will wait then.

Regarding abidiff, I did not say there were no changes, but the only reported as non-backward incompatibility is the soname.
```
ELF SONAME changed
Functions changes summary: 0 Removed, 8 Changed (697 filtered out), 32 Added functions
Variables changes summary: 0 Removed, 0 Changed, 0 Added variable

SONAME changed from 'libgit2.so.1.1' to 'libgit2.so.1.2'

32 Added functions:

  [A] 'function int git_attr_foreach_ext(git_repository*, git_attr_options*, const char*, int (const char*, const char*, void*)*, void*)'    {git_attr_foreach_ext}
  [A] 'function int git_attr_get_ext(const char**, git_repository*, git_attr_options*, const char*, const char*)'    {git_attr_get_ext}
  [A] 'function int git_attr_get_many_ext(const char**, git_repository*, git_attr_options*, const char*, size_t, const char**)'    {git_attr_get_many_ext}
  [A] 'function int git_blob_filter_options_init(git_blob_filter_options*, unsigned int)'    {git_blob_filter_options_init}
[...]
8 functions with some indirect sub-type change:

  [C] 'function void git_annotated_commit_free(git_annotated_commit*)' at annotated_commit.c:218:1 has some indirect sub-type changes:
    parameter 1 of type 'git_annotated_commit*' has sub-type changes:
      in pointed to type 'typedef git_annotated_commit' at types.h:198:1:
        underlying type 'struct git_annotated_commit' at annotated_commit.h:27:1 changed:
          type size hasn't changed
          1 data member changes (2 filtered):
            type of 'git_commit* git_annotated_commit::commit' changed:
              in pointed to type 'typedef git_commit' at types.h:136:1:
                underlying type 'struct git_commit' at commit.h:19:1 changed:
                  type size hasn't changed
                  1 data member change:
                    type of 'git_object git_commit::object' changed:
                      underlying type 'struct git_object' at object.h:19:1 changed:
[...]
```
Regards,
Laurent

include/git2/version.h
stac47: Yes, your are clearly true!

ethomson:
I'm not familiar with `abidiff`, but I'm thinking that https://github.com/libgit2/libgit2/commit/d3bdf33b58f16939a4fd43ab541dcd2ee535b6a3 is an ABI breaking change of that struct.

stac47:
Let's close this one since ABI stabilization is planned on major 2.

ethomson:
Sorry about v1.3 @stac47 - I broke the ABI a little bit there, too, but I want you to know that I felt bad doing it.  😢 

I'm thinking about v2.0 now and ABI stabilization is important to me for that.

PR #6052, opened 2021-09-14
Add test to ensure empty proxy env behaves like unset env

sathieu:


sathieu:
Small addition to #6026.

tests/remote/httpproxy.c
```diff
@@ -136,4 +136,8 @@ void test_remote_httpproxy__env(void)
 	/* configuration overrides environment variables */
 	cl_setenv("NO_PROXY", "github.none");
 	assert_config_match("http.https://github.com.proxy", "http://localhost:11/");
+
+	/* empty env behaves like unset env */
+	cl_setenv("HTTPS_PROXY", "");
+	assert_proxy_is(NULL);
```
ethomson, line 142: I dug in to the test failures because I thought that the proxy detection code was returning `""` as the proxy or something, it was not.

This check needs to happen above the configuration change (or should otherwise undo the configuration change at line 138).  The configuration overrides the environment variable, so this test would never be true.

I moved this up above and merged it.  Thanks for the fix!  👍

sathieu:
Thanks @ethomson!

PR #6053, opened 2021-09-14
Discover libssh2 without pkg-config

stac47:
Hello,

I need to build the library without relying upon pkg-config. What I propose is to fallback on a traditional library discovery of libssh2 when the pkg-config is not available.

I saw that many other dependencies are discovered with simple `FIND_PACKAGE` so I propose the same for libssh2.
Regards,


src/CMakeLists.txt
```diff
@@ -232,6 +232,13 @@ ENDIF()
 # Optional external dependency: libssh2
 IF (USE_SSH)
 	FIND_PKGLIBRARIES(LIBSSH2 libssh2)
+    IF (NOT LIBSSH2_FOUND)
+        FIND_PACKAGE(LibSSH2)
+        SET(LIBSSH2_INCLUDE_DIRS ${LIBSSH2_INCLUDE_DIR})
+        GET_FILENAME_COMPONENT(LIBSSH2_LIBRARY_DIRS "${LIBSSH2_LIBRARY}" DIRECTORY)
+        SET(LIBSSH2_LIBRARIES ${LIBSSH2_LIBRARY})
+        SET(LIBSSH2_LDFLAGS "-lssh2")
+    ENDIF()
```
ethomson, line 241:
```suggestion
	IF (NOT LIBSSH2_FOUND)
		FIND_PACKAGE(LibSSH2)
		SET(LIBSSH2_INCLUDE_DIRS ${LIBSSH2_INCLUDE_DIR})
		GET_FILENAME_COMPONENT(LIBSSH2_LIBRARY_DIRS "${LIBSSH2_LIBRARY}" DIRECTORY)
		SET(LIBSSH2_LIBRARIES ${LIBSSH2_LIBRARY})
		SET(LIBSSH2_LDFLAGS "-lssh2")
	ENDIF()
```

ethomson:
Thanks, this seems reasonable to me!

PR #6055, opened 2021-09-14
Longpath filter bug

### LINKED ISSUE #6054: git_index_add_bypath fails with "path too long" when longpaths is true and filters are set
### Reproduction steps

Add a test similar to test_win32_longpath__status_and_add but set "core.autocrlf" to true:

```
void test_win32_longpath__status_and_add_with_filter(void)
{
#ifdef GIT_WIN32
	git_repository *repo = cl_git_sandbox_init("testrepo");
	git_index *index;
	git_buf out = GIT_BUF_INIT;
	unsigned int status_flags;

	cl_repo_set_bool(repo, "core.longpaths", true);
        cl_repo_set_bool(repo, "core.autocrlf", true);
	cl_git_pass(git_repository_workdir_path(&out, repo, LONG_FILENAME));

	cl_git_rewritefile(out.ptr, "This is a long path.\r\n");

	cl_git_pass(git_status_file(&status_flags, repo, LONG_FILENAME));
	cl_assert_equal_i(GIT_STATUS_WT_NEW, status_flags);

	cl_git_pass(git_repository_index(&index, repo));
	cl_git_pass(git_index_add_bypath(index, LONG_FILENAME));

	cl_git_pass(git_status_file(&status_flags, repo, LONG_FILENAME));
	cl_assert_equal_i(GIT_STATUS_INDEX_NEW, status_flags);

	git_index_free(index);
	git_buf_dispose(&out);
#endif
}
```

### Expected behavior

git_index_add_bypath passes.

### Actual behavior

git_index_add_bypath fails. The reason is this function is called with a null git_repository:

bool should_validate_longpaths(git_repository *repo);

The function git_blob__create_from_paths calls write_file_filtered but doesn't pass the git_repository down to git_filter_list_apply_to_file.

A simple fix would be to pass the git_repository to write_file_filtered. I'll create a pull request with this idea.

v1.2.0.

Windows.

lrm29:
Fixes #6054.

An alternative might be to handle the null git_repository case in the longpath check ((bool) should_validate_longpaths(git_repository *repo)). Could also put an assert on "repo" argument in should_validate_longpaths.

### DIFF
--- a/tests/win32/longpath.c
+++ b/tests/win32/longpath.c
@@ -64,15 +64,11 @@ void test_win32_longpath__workdir_path_validated(void)
 #endif
 }
 
-void test_win32_longpath__status_and_add(void)
-{
-#ifdef GIT_WIN32
-	git_repository *repo = cl_git_sandbox_init("testrepo");
+static void assert_longpath_status_and_add(git_repository* repo) {
 	git_index *index;
 	git_buf out = GIT_BUF_INIT;
 	unsigned int status_flags;
 
-	cl_repo_set_bool(repo, "core.longpaths", true);
 	cl_git_pass(git_repository_workdir_path(&out, repo, LONG_FILENAME));
 
 	cl_git_rewritefile(out.ptr, "This is a long path.\r\n");
@@ -88,6 +84,16 @@ void test_win32_longpath__status_and_add(void)
 
 	git_index_free(index);
 	git_buf_dispose(&out);
+}
+
+void test_win32_longpath__status_and_add(void)
+{
+#ifdef GIT_WIN32
+	git_repository *repo = cl_git_sandbox_init("testrepo");
+
+	cl_repo_set_bool(repo, "core.longpaths", true);
+
+	assert_longpath_status_and_add(repo);
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


tests/win32/longpath.c
```diff
@@ -64,15 +64,11 @@ void test_win32_longpath__workdir_path_validated(void)
 #endif
 }
 
-void test_win32_longpath__status_and_add(void)
-{
-#ifdef GIT_WIN32
-	git_repository *repo = cl_git_sandbox_init("testrepo");
+static void assert_longpath_status_and_add(git_repository* repo) {
 	git_index *index;
```
ethomson, line 68:
```suggestion
#ifdef GIT_WIN32
static void assert_longpath_status_and_add(git_repository* repo) {
	git_index *index;
```

tests/win32/longpath.c
```diff
@@ -88,5 +84,27 @@ void test_win32_longpath__status_and_add(void)
 
 	git_index_free(index);
 	git_buf_dispose(&out);
+}
```
ethomson, line 87:
```suggestion
}
#endif
```

ethomson (changes requested):

ethomson:
Looks good to me - we'll need to `ifdef` around that new function to prevent unused warnings, but otherwise 👍

ethomson:
Thanks for the fix.  I changed the unit test a bit in 005c740b15bf2b77e5ef1112aed887751fc2c827 to check the contents in the repo to ensure that filtering was applied.

PR #6058, opened 2021-09-16
Allow proxy options when connecting with a detached remote.

### LINKED ISSUE #6059: Allow proxy options when connecting with a detached remote.
### Reproduction steps

Create a detached remote (so remote->repo is null) using git_remote_create_detached.
Calling git_remote_connect then fails, but only when git_proxy_options are passed in. This is because git_remote__http_proxy expects the remote to have a repo attached.

v1.2.0

Windows and Linux

lrm29:
Fixes  #6059.

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -884,7 +884,7 @@ static void url_config_trim(git_net_url *url)
 
 static int http_proxy_config(char **out, git_remote *remote, git_net_url *url)
 {
-	git_config *cfg;
+	git_config *cfg = NULL;
 	git_buf buf = GIT_BUF_INIT;
 	git_net_url lookup_url = GIT_NET_URL_INIT;
 	int error;
--- a/src/remote.c
+++ b/src/remote.c
@@ -888,10 +888,20 @@ static int http_proxy_config(char **out, git_remote *remote, git_net_url *url)
 	git_buf buf = GIT_BUF_INIT;
 	git_net_url lookup_url = GIT_NET_URL_INIT;
 	int error;
+	int cleanup_config = 0;
 
-	if ((error = git_net_url_dup(&lookup_url, url)) < 0 ||
-	    (error = git_repository_config__weakptr(&cfg, remote->repo)) < 0)
+	if ((error = git_net_url_dup(&lookup_url, url)) < 0)
 		goto done;
+
+	if (remote->repo) {
+		if ((error = git_repository_config__weakptr(&cfg, remote->repo)) < 0)
+			goto done;
+	} else {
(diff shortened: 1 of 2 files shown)


src/remote.c
```diff
@@ -888,11 +888,21 @@ static int http_proxy_config(char **out, git_remote *remote, git_net_url *url)
 	git_buf buf = GIT_BUF_INIT;
 	git_net_url lookup_url = GIT_NET_URL_INIT;
 	int error;
+	int cleanup_config = 0;
```
ethomson, line 891:
```suggestion
```

src/remote.c
```diff
@@ -888,11 +888,21 @@ static int http_proxy_config(char **out, git_remote *remote, git_net_url *url)
...
 
-	if ((error = git_net_url_dup(&lookup_url, url)) < 0 ||
-	    (error = git_repository_config__weakptr(&cfg, remote->repo)) < 0)
+	if ((error = git_net_url_dup(&lookup_url, url)) < 0)
 		goto done;
 
+	if (remote->repo) {
+		if ((error = git_repository_config__weakptr(&cfg, remote->repo)) < 0)
```
ethomson, line 897:
```suggestion
		if ((error = git_repository_config(&cfg, remote->repo)) < 0)
```

src/remote.c
```diff
@@ -888,11 +888,21 @@ static int http_proxy_config(char **out, git_remote *remote, git_net_url *url)
...
+			goto done;
+	} else {
+		if ((error = git_config_open_default(&cfg)) < 0)
+			goto done;
+
+		cleanup_config = 1;
```
ethomson, line 903:
```suggestion
```

src/remote.c
```diff
@@ -922,6 +932,9 @@ static int http_proxy_config(char **out, git_remote *remote, git_net_url *url)
 	error = lookup_config(out, cfg, "http.proxy");
 
 done:
+	if (cleanup_config)
+		git_config_free(cfg);
+
```
ethomson, line 937:
```suggestion
	git_config_free(cfg);
```

ethomson (changes requested):
I think that we can avoid keeping this `cleanup_config` variable by getting a refcounted pointer for the repository's config, instead of a weak pointer.  Then we'll just free it always.  There's no real overhead to getting a regular config and `git_config_free`ing it, the only real advantage to weak pointers is that you don't need to worry about freeing them.

ethomson:
Thanks for this!  I might recommend a minor cleanup.

ethomson:
Looks good to me!  Thanks for the fix!

PR #6060, opened 2021-09-16
merge: Check file mode when resolving renames

ccstolley:
When determining if ours or theirs changed, we check the oids but not
their respective file modes. This can lead to merges introducing incorrect
file mode changes (eg., in a revert). A simple linear example might be:

commit A - introduces file `foo` with chmod 0755
commit B - updates some unrelated file
commit C - renames `foo` to `bar` and chmod 0644

If B is reverted, `bar` will unexpectedly acquire mode 0755. This is
reproduced in the `revert/rename.c` test below.


ethomson:
Thanks for the fix!

Issue #5949, opened 2021-07-18
[Feature Request] support `--no-checkout` for `git_worktree_add` 


Whirlwind:
`git worktree add [Path] --no-checkout` will allow user skip checkout the files.

In the libgit2, it will force checkout the files, and there is not option to skip it.

I hope the `git_worktree_add_options` add the `checkout` option.

ethomson:
This seems like a good idea.  Naively, I suspect that this is an easy fix, so I've added that label.  This seems like a good first issue for a [new contributor](https://up-for-grabs.net/#/filters?names=508)!

Jagroop-Singh:
Hey there, I'll grab this one

Whirlwind:
I try to add the feature for my project:
```
	if (wtopts.checkout == 1) {
		/* Checkout worktree's HEAD */
		git_checkout_options coopts = GIT_CHECKOUT_OPTIONS_INIT;
		coopts.checkout_strategy = GIT_CHECKOUT_FORCE;
		if ((err = git_checkout_head(wt, &coopts)) < 0)
			goto out;
	}
```
But, I found that everything is deleted in the stage, not the unstaged. I don't know how to keep them in the unstaged ( not in the index).

alexandre1985:
how is the progress @Jagroop-Singh ?
I also want to grab a newbie coding issue....

Jagroop-Singh:
@alexandre1985 You can go ahead, I've looked at the code of git/git and no dice on where the code is. I've tried recursive listing with cat, and grepped for the relevant keywords, and following the code has led me in circles.

Jagroop-Singh:
I couldn't find where the logic for --no-checkout exists

alexandre1985:
@Jagroop-Singh I am also stuck there.... 🤔

ethomson:
Looking at this: https://github.com/libgit2/libgit2/blob/e65229ee972c113413eeca77853213352129bd47/src/worktree.c#L406

The way that I would approach this problem is to extend `git_worktree_add_options` to include `git_checkout_options`.

You can look at other functions (like merge) to see this.

alexandre1985:
what is GIT_INIT_STRUCTURE_FROM_TEMPLATE() ?

punkymaniac:
In short, This macro initialise a struct from it's template.

[...]

PR #6062, opened 2021-09-20
Allow user checkout options on git_worktree_add

punkymaniac:
Extend the `git_worktree_add_options` to include `git_checkout_options`.

github issue #5949

### DIFF
--- a/include/git2/worktree.h
+++ b/include/git2/worktree.h
@@ -86,10 +86,15 @@ typedef struct git_worktree_add_options {
 
 	int lock; /**< lock newly created worktree */
 	git_reference *ref; /**< reference to use for the new worktree HEAD */
+
+	/**
+	 * Options for the checkout.
+	 */
+	git_checkout_options checkout_opts;
 } git_worktree_add_options;
 
 #define GIT_WORKTREE_ADD_OPTIONS_VERSION 1
-#define GIT_WORKTREE_ADD_OPTIONS_INIT {GIT_WORKTREE_ADD_OPTIONS_VERSION,0,NULL}
+#define GIT_WORKTREE_ADD_OPTIONS_INIT {GIT_WORKTREE_ADD_OPTIONS_VERSION,0,NULL,GIT_CHECKOUT_OPTIONS_INIT}
 
 /**
  * Initialize git_worktree_add_options structure
--- a/src/worktree.c
+++ b/src/worktree.c
@@ -407,8 +407,6 @@ int git_worktree_add(git_worktree **out, git_repository *repo,
 	/* Checkout worktree's HEAD */
 	if (opts != NULL)
 		memcpy(&coopts, &wtopts.checkout_opts, sizeof(coopts));
-	else
-		coopts.checkout_strategy = GIT_CHECKOUT_FORCE;
 	if ((err = git_checkout_head(wt, &coopts)) < 0)
 		goto out;
 
--- a/src/worktree.c
+++ b/src/worktree.c
@@ -405,7 +405,10 @@ int git_worktree_add(git_worktree **out, git_repository *repo,
 		goto out;
 
 	/* Checkout worktree's HEAD */
-	coopts.checkout_strategy = GIT_CHECKOUT_FORCE;
+	if (opts != NULL)
+		memcpy(&coopts, &wtopts.checkout_opts, sizeof(coopts));
+	else
+		coopts.checkout_strategy = GIT_CHECKOUT_FORCE;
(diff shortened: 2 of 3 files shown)


include/git2/worktree.h
```diff
@@ -86,6 +86,11 @@ typedef struct git_worktree_add_options {
 
 	int lock; /**< lock newly created worktree */
 	git_reference *ref; /**< reference to use for the new worktree HEAD */
+
+	/**
+	 * Options for the checkout.
+	 */
+	git_checkout_options *coopts;
```
ethomson, line 93: These should be a struct within the struct, not a pointer to another options struct.  See for example `git_submodule_update_options`.

```suggestion
	git_checkout_options checkout_opts;
```

That means adding default checkout options to `GIT_WORKTREE_ADD_OPTIONS_INIT`.

Curious why `FORCE` is the default?  I know that you didn't introduce that code but I'm curious now that we're revisiting it.  What happens if this is `SAFE` instead?

punkymaniac: Without pointer I have trouble to check if the checkout options was given by the user or not.

But maybe we could remove the default `GIT_CHECKOUT_FORCE` and only use the checkout options as implemented in `git_submodule_updates_options`.

ethomson: You shouldn't need to check if a user gave us checkout options.  User should either provide a `NULL` worktree options - and we'll use defaults for those options (and those include defaults for checkout options).  Or they gave us worktree options that include checkout options, and we'll use them.

Users are required to provide the options that they expect.  The `GIT_*_OPTIONS_INIT` macros and the `git_*_options_init` functions are responsible for giving them good defaults.  They can also pass `NULL` to get these options.

We don't need to give users the option to pass a half-empty options structure by passing in a pointer to another options structure.  This makes FFI languages extremely difficult because they'd need to create an opts struct, pin it, and then put a pointer to that in the other opts structure.

We have a lot of patterns where opts structure include other opts structures; we should not deviate from the existing pattern.

punkymaniac:
It's seem that the default behaviour `FORCE` will be here from the beginning of `git_worktree_add` see dea7488e93.

punkymaniac:
Sorry for the delay to write the test.

Check fail because `git_buf` was now replaced by `git_str` for internal usage (see f0e693b ).
I can fix it, but in the libgit2 workflow, do we prefer rebase or merge ?

include/git2/worktree.h
```diff
@@ -86,10 +86,15 @@ typedef struct git_worktree_add_options {
 
 	int lock; /**< lock newly created worktree */
 	git_reference *ref; /**< reference to use for the new worktree HEAD */
+
+	/**
+	 * Options for the checkout.
+	 */
+	git_checkout_options checkout_opts;
```
ethomson, line 93: We're not super consistent about `opts` vs `options` here, but I'd prefer `options` when a struct is embedded within another struct.

```suggestion
	git_checkout_options checkout_options;
```

src/worktree.c
```diff
@@ -405,7 +405,8 @@ int git_worktree_add(git_worktree **out, git_repository *repo,
 		goto out;
 
 	/* Checkout worktree's HEAD */
-	coopts.checkout_strategy = GIT_CHECKOUT_FORCE;
+	if (opts != NULL)
+		memcpy(&coopts, &wtopts.checkout_opts, sizeof(coopts));
```
ethomson, line 409: We can guarantee that `wtopts` is validly initialized, even when `opts` is NULL.  I think that we should just:

```suggestion
	memcpy(&coopts, &wtopts.checkout_opts, sizeof(coopts));
```

tests/worktree/worktree.c
```diff
@@ -292,6 +292,28 @@ void test_worktree_worktree__add_with_explicit_branch(void)
...
 
+void test_worktree_worktree__add_no_checkout(void)
+{
+	git_worktree *wt;
+	git_repository *wtrepo;
+	git_index *index;
+	git_str path = GIT_STR_INIT;
+	git_worktree_add_options opts = GIT_WORKTREE_ADD_OPTIONS_INIT;
+
+	opts.checkout_opts.checkout_strategy = GIT_CHECKOUT_NONE;
```
ethomson, line 303:
```suggestion
	opts.checkout_options.checkout_strategy = GIT_CHECKOUT_NONE;
```

ethomson (commented):
This is a great PR, thanks for the fix.  I have just a few stylistic requests.

ethomson:
I went ahead and manually merged this.  Thanks again!

PR #6064, opened 2021-09-20
ci: pull libssh2 from www.libssh2.org

ethomson:
libssh2.org and www.libssh2.org were previously identical; now this is a
redirect.

### DIFF
--- a/ci/docker/centos7
+++ b/ci/docker/centos7
@@ -18,7 +18,7 @@ RUN yum install -y \
 
 FROM yum AS libssh2
 RUN cd /tmp && \
-    curl https://libssh2.org/download/libssh2-1.8.0.tar.gz | tar -xz && \
+    curl --location --silent --show-error https://www.libssh2.org/download/libssh2-1.8.0.tar.gz | tar -xz && \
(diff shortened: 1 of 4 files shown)


ci/docker/centos7
```diff
@@ -18,7 +18,7 @@ RUN yum install -y \
 
 FROM yum AS libssh2
 RUN cd /tmp && \
-    curl https://libssh2.org/download/libssh2-1.8.0.tar.gz | tar -xz && \
+    curl https://www.libssh2.org/download/libssh2-1.8.0.tar.gz | tar -xz && \
```
lhchavez, line 21:
```suggestion
    curl --location https://www.libssh2.org/download/libssh2-1.8.0.tar.gz | tar -xz && \
```

might still be good to throw in the flag to let curl know to follow redirects in case the libssh2 folks want to move stuff around.

lhchavez (approved):

ci/docker/centos7
ethomson: 👍  Looks like the other Dockerfiles already did this, so I just unified their arguments across the board.

PR #6072, opened 2021-09-24
api : expose `git_config__update_entry` à la `git_config_update_entry`

257:


ethomson:
Not saying "no", but this is a very internal feeling API.  It's quick & dirty, and not even really used.  Why is this useful to you?  What APIs are you using _instead_ of this?

257:
> Why is this useful to you? 

UPDATE, needless to say, is a must for any db.

I'm [ab]using init files for storing some human generated description of media files for podcast feeds. code is not out yet but i could dump it in a repo if need be. please just let me know.

in my case, in particular, url for enclosures tags of items (raw 2.0) could be set relative by users and I end up pretending the xmlbase to it. sometimes updating the init accordingly... I'm probably going to ditch updating the init at the end though :)

> What APIs are you using _instead_ of this?

pretty much everything in `git_config__update_entry` :)
I have a verbatim, copy-paste, 0-mode if I remember correctly, of  `git_config__update_entry` in my code. i use that.

PR #6083, opened 2021-10-02
tests: rename test runner to `libgit2_tests`, build option to `BUILD_TESTS`.

ethomson:
Nobody knows what CLAR is.  The test building option should be `BUILD_TESTS` and the test runner to `libgit2_tests`


ethomson:
We're breaking everybody's habits today with the `git_buf` change.  Today is just a day of churn!

PR #6086, opened 2021-10-02
Pedantic header test

boretrk:
Users of libgit2 might want to be able to use stricter compiler flags than what is used to compile the actual library.
This PR adds a separate target that only includes "git2.h" and does nothing else.
That way the headers can be tested with compiler flags that are a bit too much for the actual code.
For now it enables -pedantic and turns off gnu extensions.

There are also a whole bunch of trailing commas removed from enums.


ethomson:
I think that this makes sense.  I'd love to see if we can bury it inside `tests/` though instead of adding a new top-level directory that's really a test that users don't care about and won't interact with directly.

boretrk:
There are a couple of more enums with trailing commas in the src/ directory.
I didn't remove them since they aren't part of what a user would include.
If it is of interest I could clean them up in this PR too. (But regressions won't be caught.)

boretrk:
Rebased to merge cleanly with main again, should be good to go.

tests/CMakeLists.txt
```diff
@@ -79,3 +79,15 @@ add_clar_test(ssh                 -v -sonline::push -sonline::clone::ssh_cert -s
...
 add_clar_test(auth_clone_and_push -v -sonline::clone::push -sonline::push)
+
+# Header file validation project
+add_executable(headertest headertest.c)
+set_target_properties(headertest PROPERTIES C_STANDARD 90)
+set_target_properties(headertest PROPERTIES C_EXTENSIONS OFF)
+target_include_directories(headertest PRIVATE ${LIBGIT2_INCLUDES})
+
+if (MSVC)
+    target_compile_options(headertest PUBLIC /W4 /WX)
```
ethomson, line 90:
```suggestion
	target_compile_options(headertest PUBLIC /W4 /WX)
```

tests/CMakeLists.txt
```diff
@@ -79,3 +79,15 @@ add_clar_test(ssh                 -v -sonline::push -sonline::clone::ssh_cert -s
...
+else()
+    target_compile_options(headertest PUBLIC -Wall -Wextra -pedantic -Werror)
```
ethomson, line 92:
```suggestion
	target_compile_options(headertest PUBLIC -Wall -Wextra -pedantic -Werror)
```

tests/CMakeLists.txt
```diff
@@ -79,3 +79,15 @@ add_clar_test(ssh                 -v -sonline::push -sonline::clone::ssh_cert -s
 add_clar_test(proxy               -v -sonline::clone::proxy)
 add_clar_test(auth_clone          -v -sonline::clone::cred)
 add_clar_test(auth_clone_and_push -v -sonline::clone::push -sonline::push)
+
+# Header file validation project
```
ethomson, line 83:
```suggestion
#
# Header file validation project: ensure that we do not publish any sloppy
# definitions in our headers and that a consumer can include <git2.dll>
# even when they have aggressive C90 warnings enabled.
#

```

ethomson (approved):

ethomson:
Sorry for the delay.  This is really great.  I appreciate you moving this out into a `tests` directory.  I'm going to pull in my suggestions manually - I thought I could do it in the GitHub UI but cannot.

PR #6093, opened 2021-10-16
Solve compilation warning reported by MSVC related to possible loss of data

rafael-santiago:
Hi!

I was trying to build the library on ``MSVC`` (2019) and it was claiming about possible loss of data. This commit just adds the necessary casting to silent off this kind of annoying warning.

I chose the casting because I suppose that the use of ``int`` or ``long`` in replacement of ``size_t`` seems to be a project decision, so I kept it up.

I felt tempted to change ``int xdl_num_out()`` to ``size_t xdl_num_out()`` (since it returns the buffer size taking its delta from current position to its head), again, due to the massive use of ``long`` instead of ``size_t``, it seems to be a choice  so I just made the things clearer for the compiler not complain about.

### DIFF
--- a/src/xdiff/xdiffi.c
+++ b/src/xdiff/xdiffi.c
@@ -364,7 +364,7 @@ int xdl_do_diff(mmfile_t *mf1, mmfile_t *mf2, xpparam_t const *xpp,
 	kvdf += xe->xdf2.nreff + 1;
 	kvdb += xe->xdf2.nreff + 1;
 
-	xenv.mxcost = xdl_bogosqrt(ndiags);
+	xenv.mxcost = xdl_bogosqrt((long)ndiags);
 	if (xenv.mxcost < XDL_MAX_COST_MIN)
 		xenv.mxcost = XDL_MAX_COST_MIN;
 	xenv.snake_cnt = XDL_SNAKE_CNT;
--- a/src/xdiff/xutils.c
+++ b/src/xdiff/xutils.c
@@ -62,14 +62,14 @@ int xdl_emit_diffrec(char const *rec, long size, char const *pre, long psize,
 
 void *xdl_mmfile_first(mmfile_t *mmf, long *size)
 {
-	*size = mmf->size;
+	*size = (long)mmf->size;
 	return mmf->ptr;
 }
 
 
 long xdl_mmfile_size(mmfile_t *mmf)
 {
-	return mmf->size;
+	return (long)mmf->size;
 }
 
 
@@ -339,7 +339,7 @@ int xdl_num_out(char *out, long val) {
 		*str++ = '0';
 	*str = '\0';
 
-	return str - out;
+	return (int)(str - out);
 }
 
 int xdl_emit_hunk_hdr(long s1, long c1, long s2, long c2,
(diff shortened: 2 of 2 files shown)


ethomson:
> I chose the casting because I suppose that the use of `int` or `long` in replacement of `size_t` seems to be a project decision, so I kept it up.

`xdiff` is a dependency that we picked up from git and was originally http://www.xmailserver.org/xdiff-lib.html but has been modified by git and further modified by us.  I don't think that there were ever any thoughtful decisions made about integer type usage, to be honest.

The compiler is definitely _correct_ here, these should be warnings.

But our `WERROR` builds remove the warnings as errors from all of the `xdiff` code.  So these warnings shouldn't be preventing you from building, just annoying.  I don't love removing correct warnings, even if they are annoying.

What I'd really like to do is to pull in the most recent xdiff from git and see what's going on there.  What I'd really _really_ like to do is to have a single unified xdiff that we both use, but the last conversation I had about this with any git contributors was typically unenthusiastic.

rafael-santiago:
> xdiff is a dependency that we picked up from git and was originally http://www.xmailserver.org/xdiff-lib.html but has been modified by git and further modified by us. I don't think that there were ever any thoughtful decisions made about integer type usage, to be honest.
>The compiler is definitely correct here, these should be warnings.

Yeah, I agree with you and with the compiler, I do not understand why after so many standards people still prefer using ``int``, ``long`` for expressing memory size, offsets or anything that at end will be directly linked to cpu wordsize. Just a thing is legacy it does not mean it is a theorem. Old platforms that do not support size_t should be patched to support it by using some explicit typedefs based on them possibilities, instead of shifting back 🦖 ☄️  all new ones to use ``long`` or ``int``.

>But our WERROR builds remove the warnings as errors from all of the xdiff code. So these warnings shouldn't be preventing you from building, just annoying. I don't love removing correct warnings, even if they are annoying.

For sure, but such an elementar programming mistake being warned and not fixed along several releases could sound "sloppish", since the use of ``long`` or ``int`` is a kind of "all in a day's work" due to be a 3rd party stuff. Personaly I find that disable a warning is a last resort and almost all cases the compiler is right. In cases of data losing warning, if the programmer is aware about the implications and issues (due to it being released since then, I find people here are, for sure) the casting is the right way of co-exist with this 3rd party coding issue in a more concious way, you know, the programmer is saying:

>>"yes it could happen but for our specific cases we are aware about it and until now we can live with this data loss posibility"

Otherwise it is like "no one seen, nobody cares". You know, a way of programatically saying that the issue reported was checked taking into consideration the agreed design. All developers are aware about.

>What I'd really like to do is to pull in the most recent xdiff from git and see what's going on there. What I'd really really like to do is to have a single unified xdiff that we both use

Bad news! The latest version is still using ``long`` and ``int``. More over, ``int xdl_emit_hunk_hdr``, at first glance, seems different from the implementation here. 

> but the last conversation I had about this with any git contributors was typically unenthusiastic.

I understand you! Resistance on going ahead is such a dichotomic thing in the whooooooole "dynamic" IT world 🦖 ☄️ 💥  It is normal!

PR #6098, opened 2021-10-26
Add CIFuzz

AdamKorcz:
Adds CIFuzz for libgit2 which will run its fuzzers for 600 seconds when a pull request is made. The 600 seconds can be modified.

The documentation of CIFuzz can be found here: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

Libgit2's fuzzers can be found here: https://github.com/libgit2/libgit2/tree/main/fuzzers

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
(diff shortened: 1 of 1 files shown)


ethomson:
Hey @AdamKorcz - I'm pleased to see more fuzzing - more testing (and fuzzing).  I didn't review this since the build didn't pass.  If this is something that you want to continue to work on, that's great.  I'd like to see the build + fuzz job succeed (or at least fail from the fuzzer in a way that demonstrates a libgit2 problem), and I'd also like to understand what makes this different than the fuzzing that we already do in CI.  Looking forward to it.  Thanks!

ethomson:
Since I don't fully understand the rationale here, I'm going to close this. Happy to reconsider if I'm able to understand it better in the future. 🙏

PR #6101, opened 2021-10-28
remotes: fix insteadOf/pushInsteadOf handling

mkhl:
a repo has a url, and a named repo may also have a pushurl.
each of these can be matched by insteadOf/pushInsteadOf configuration.

this means that want to test:
- an anonymous repo (a url)
- a named repo with a url
- a named repo with a url and pushurl

and for each of these matching configuration:

- only insteadOf
- only pushInsteadOf
- both insteadOf and pushInsteadOf

this changeset updates the tests to test all of these combinations,
and updates the implementation so the tests pass.

all expectations have been checked against the git command line client,
the results of which cna be checked with this script (run from the build directory):

```sh
#!/bin/bash
export GIT_DIR=../tests/resources/testrepo2/.gitted
for remote in insteadof-{,push}url-{fetch,push,both}; do
	echo $remote
	git remote get-url $remote
	git remote get-url --push $remote
done
```

in addition, remotes without an explicit pushurl and without a pushInsteadOf configuration return NULL for their pushurl
to preserve the current behavior.

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -30,7 +30,7 @@
 
 static int dwim_refspecs(git_vector *out, git_vector *refspecs, git_vector *refs);
 static int lookup_remote_prune_config(git_remote *remote, git_config *config, const char *name);
-char *apply_insteadof(git_config *config, const char *url, int direction);
+char *apply_insteadof(git_config *config, const char *url, int direction, int *matched);
 
 static int add_refspec_to(git_vector *vector, const char *string, bool is_fetch)
 {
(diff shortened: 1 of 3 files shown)


src/remote.c
```diff
@@ -30,7 +30,7 @@
 
 static int dwim_refspecs(git_vector *out, git_vector *refspecs, git_vector *refs);
 static int lookup_remote_prune_config(git_remote *remote, git_config *config, const char *name);
-char *apply_insteadof(git_config *config, const char *url, int direction);
+char *apply_insteadof(git_config *config, const char *url, int direction, int *matched);
```
ethomson, line 33: Generally we have out params as the first parameters, not the last.  And I'd use a boolean type here:

```suggestion
char *apply_insteadof(bool *matched, git_config *config, const char *url, int direction);
```

mkhl: good point, thanks for the hint!
i've pushed an update and will look into the failing tests next.

ethomson: Thanks!  I'm really happy to see this functionality.  Thanks for tackling it.

mkhl: oh oops, i forgot to update the local variables. with that the tests pass on my end.

mkhl: > Thanks! I'm really happy to see this functionality. Thanks for tackling it.

that's very kind of you to say, thank you!
this project had been incredibly easy to get started working on :relaxed:

ethomson:
Thanks again for this PR, I'm happy to have this fixed.  🙏

Issue #6065, opened 2021-09-20
`git_commit_summary` creates too long output for some commits


ArekPiekarz:
### Description
`git_commit_summary` produces too long paragraphs for some commits, like for this one in the Linux repository - https://github.com/torvalds/linux/commit/3218357c94af92478ef39163163a81e654385320. It creates 925 characters long text (almost the whole commit message) instead of 46 characters long (just the first line).

### Reproduction steps
Create a project with following files:

main.cpp:
```c++
#include "deps/libgit2/include/git2.h"
#include <iostream>

int main()
{
    git_libgit2_init();
    const std::string repoPath = "path_to_linux_repository";  // INSERT YOUR OWN

    git_repository* repo = nullptr;
    if (const auto status = git_repository_open(&repo, repoPath.c_str()); status != 0)
    {
        throw std::runtime_error("Failed to open repository at " + repoPath + ", status: " + std::to_string(status));
    }

    git_oid id{};
    if (const auto status = git_oid_fromstr(&id, "3218357c94af92478ef39163163a81e654385320"); status != 0)
    {
        throw std::runtime_error("Failed to create OID, status: " + std::to_string(status));
    }
    git_commit* commit = nullptr;
    if (const auto status = git_commit_lookup(&commit, repo, &id); status != 0)
    {
        throw std::runtime_error("Failed to lookup commit, status: " + std::to_string(status));
    }
    const std::string summary = git_commit_summary(commit);
    std::cout << "git_commit_summary returned (size: " + std::to_string(summary.size()) + "): " << summary << "\n";
    return 0;
}
```
CMakeLists.txt:
```cmake
project(libgit2-commit-summary-test)
cmake_minimum_required(VERSION 3.0)

add_subdirectory(deps/libgit2)
add_executable(${PROJECT_NAME} main.cpp)
target_link_libraries(${PROJECT_NAME} git2)
```
Also download libgit2 as the dependency:
```
mkdir deps && cd deps
git clone https://github.com/libgit2/libgit2.git
cd ..
```
Build and run the project:
```
mkdir build && cd build
cmake ..
cmake --build .
./libgit2-commit-summary-test
```

### Expected behavior
The summary should be: `[MIPS] More uaccess.h fixes with gcc >= 4.0.1.`

### Actual behavior
The summary is:
[...]

PR #6103, opened 2021-11-01
Corrects git commit summary as described in #6065

PyllrNL:
The issue #6065 describes a commit message by Linus that contains some white space characters between the two newlines.

This Pull Request adds a test case simulating the commit message by Linus and adds a fix to the git_commit_summary()
function.

### DIFF
--- a/src/commit.c
+++ b/src/commit.c
@@ -563,6 +563,10 @@ const char *git_commit_summary(git_commit *commit)
 					space = msg;
 					space_contains_newline = false;
 				}
+
+                if(space_contains_newline && next_character == '\n')
+                    break;
+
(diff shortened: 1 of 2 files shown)


boretrk:
Maybe switch indentation from spaces to tabs like the surrounding code?

ethomson:
Hi @PyllrNL - sorry that it took me a while to get around to reviewing this.  In the meantime #6125 also tackled this problem.  😢  I know that digging into a file parser and making a fix is a challenge, and it's disappointing that I let this sit so long that we had two PRs that fixed the same issue.  This is a good lesson to me to be quicker about feedback and code reviews.

PR #6107, opened 2021-11-08
Support checking for object existence without refresh

joshtriplett:
Support checking for object existence without refresh

Looking up a non-existent object currently always invokes `git_odb_refresh`. If looking up a large batch of objects, many of which may legitimately not exist, this will repeatedly refresh the ODB to no avail.

Add a `git_odb_exists_ext` that accepts flags controlling the ODB lookup, and add a flag to suppress the refresh. This allows the user to control if and when they refresh (for instance, refreshing once before starting the batch).

Add tests for ODB refresh.

Update documentation for the ODB backend refresh logic: Commit b1a6c316a6070fac4ab1ec5792979838f7145c39 moved auto-refresh into the pack backend, and added a comment accordingly. Commit 43820f204ea32503b4083e3b6b83f30a0a0031c9 moved auto-refresh back *out* of backends into the ODB layer, but didn't update the comment.

### DIFF
--- a/include/git2/odb.h
+++ b/include/git2/odb.h
@@ -22,6 +22,17 @@
  */
 GIT_BEGIN_DECL
 
+/** Flags controlling the behavior of ODB lookup operations */
+typedef enum {
+	/**
+	 * Don't call `git_odb_refresh` if the lookup fails. Useful when doing
+	 * a batch of lookup operations for objects that may legitimately not
+	 * exist. When using this flag, you may wish to manually call
+	 * `git_odb_refresh` before processing a batch of objects.
+	 */
+	GIT_ODB_LOOKUP_NO_REFRESH = (1 << 0),
+} git_odb_lookup_flags_t;
+
 /**
  * Function type for callbacks from git_odb_foreach.
  */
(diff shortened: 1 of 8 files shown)


ethomson:
Seems like a reasonable addition to me.  I've only taken a quick glance - I'd like to take a few minutes to 👀 the code a bit deeper, but thanks for adding this.  👍

include/git2/odb.h
```diff
@@ -22,6 +22,17 @@
...
 
+/** Flags controlling the behavior of ODB lookup operations */
+typedef enum {
+	/**
+	 * Don't call `git_odb_refresh` if the lookup fails. Useful when doing
+	 * a batch of lookup operations for objects that may legitimately not
+	 * exist. When using this flag, you may wish to manually call
+	 * `git_odb_refresh` before processing a batch of objects.
+	 */
+	GIT_ODB_LOOKUP_NO_REFRESH = (1 << 0),
```
boretrk, line 33:
```suggestion
	GIT_ODB_LOOKUP_NO_REFRESH = (1 << 0)
```
For C90 compliance, otherwise #6086 will fail.

ethomson:
Thanks!

PR #6109, opened 2021-11-10
Document that `git_odb` is thread-safe

joshtriplett:
Commit 4ae41f9c639d246d34dac89c3f1d9451c9cfa8d3 made `git_odb`
race-free, and added internal locking. Update `docs/threading.md`
accordingly, so that APIs built atop libgit2 (e.g. language bindings)
can count on this.

(This appears to be true as far as I can tell, at least.)


ethomson:
👍  Thanks!

PR #6113, opened 2021-11-11
Add missing-declarations warning globally

ethomson:
At present, we enable `missing-declarations` warnings in the source tree but not in fuzzers, test, or examples.  Move this warning to the top-level so that it covers all our source.

This corrects a number of places where we were _not_ declaring our functions and either adds an `extern` declaration in a header or moves them to a `static` declaration.

Older versions of gcc (prior to 6.0), like what we use on xenial, are unhappy with our test function declarations.  Generate a header file to declare them and include it when necessary.  Do not do this over-optimistically, since it means that any changes to our test functions produces a complete recompilation due to the new dependency. Xenial users will need to suffer.


tests/email/create.c.bak
```diff
@@ -0,0 +1,386 @@
+#include "clar.h"
```
lhchavez, line 1: was this intentional?

lhchavez (commented):
hooray for compiler-correctness!

PR #6118, opened 2021-11-16
cmake: upgrade to >= 3.15.0, remove unnecessary object library

ethomson:
We create [an object library](https://github.com/libgit2/libgit2/blob/main/src/CMakeLists.txt#L1) so that we can [use the `$<TARGET_OBJECTS>` generator](https://github.com/libgit2/libgit2/blob/main/src/CMakeLists.txt#L226) to simplify carrying around the list of objects.  But this is an unnecessary level of indirection on top of creating the library itself, and definitely annoying for IDE users to reason about.

cmake 3.15.0 allows the `TARGET_OBJECTS` generator to work with any library, not just object libraries.  Upgrade to this version.

cmake 3.15.0 is now 2.5 years old, so it seems reasonable to use as a minimum requirement.  The disappointment here is that Ubuntu Xenial does not support it, but given that _it_ is now aged out of LTS, it feels like libgit2 need not compile with the tooling out-of-the-box.  We should now look to Focal as the default for the "best experience" out of the box.

This doesn't remove the ability to _run_ or _build_ on Xenial, just makes the developer experience slightly less great.

(But I'm willing to listen to arguments.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # CMake build script for the libgit2 project
 # See `README.md` for build instructions.
 
-cmake_minimum_required(VERSION 3.5.1)
+cmake_minimum_required(VERSION 3.15.0)
 
 project(libgit2 VERSION "1.3.0" LANGUAGES C)
 
--- a/ci/docker/xenial
+++ b/ci/docker/xenial
@@ -5,7 +5,6 @@ RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
         bzip2 \
         clang \
-        cmake \
         curl \
         gcc \
         git \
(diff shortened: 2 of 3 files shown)


ethomson:
@boretrk and @NattyNarwhal, I'm looking mostly at you here when I start making newer things required.  😁

boretrk:
I'm completely fine with it. I don't hope to ever be able to run any version of CMake natively.
I prefer if the CMakeLists are as humanly readable as possible.

NattyNarwhal:
I'm using CMake 3.5 because it supports some older generators (VC++6, cough), but I guess I could fall back to raw NMake. I'm building lg2 on XP for easier compat with toolchain (VC++6 compiler, suitably old Platform SDK) stuff, so that can be 3.13. In theory, I might be easier to build on newer Windows w/ old toolchain, but I don't know how reliable it could be.

I should probably pivot towards writing my own project/NMake files instead of placing burden on CMake requirements, but I don't know what the stance of the project has on those. Personally, I will say it feels slightly awkward to be writing C89 but expecting a fairly new toolchain.

ethomson:
> I'm using CMake 3.5 because it supports some older generators (VC++6, cough)

🤔  Okay.  This is good to know.  Let me noodle on this a bit and see what I can come up with.

ethomson:
Let's not.  This is an improvement, but it's not such an improvement that we need to start breaking workflows.

@NattyNarwhal I want to do this some day, but we don't need to rush it.  If you can start thinking about what a future with a newer cmake might look like, that would be :+1:.  But no hurry; let's check in about this in a few months.

PR #6125, opened 2021-11-25
git_commit_summary: ignore lines with spaces

stforek:
Fixes libgit2/libgit2#6065


boretrk:
Either this one or #6103 would work. FWIW I thought this one was easier to follow.

ethomson:
Thanks for the fix @stforek !

PR #6133, opened 2021-12-06
CLI (redux)

ethomson:
A command-line interface for libgit2.  The goal is for it to be git-compatible.

1. The libgit2 developers can more easily dogfood libgit2 to find bugs and performance issues.

2. There is growing usage of libgit2's examples as a client; libgit2's examples should be exactly that - simple code samples that illustrate libgit2's usage.  This satisfies that desire for a client more directly.

3. By producing a client ourselves, we can better understand the needs of client creators, possibly producing a shared "middleware" for commonly-used pieces of client functionality like interacting with external tools.

4. Since git is the reference implementation, we may be able to benefit from git's unit tests, running their test suite against our CLI to ensure correct behavior.

We've done a lot of the preliminary refactoring work already, so this is a _relatively_ small set of changes (at least compared to #5507).

1. The source tree is rearranged: the "pure" utility code is moved into the `util` folder and the remainder of the libgit2 code is moved into the `libgit2` folder.  This makes it easier to use the utility code from other tools (like a CLI).

2. A new executable has been added.  At present, all the commands are linked into a single executable, which statically links libgit2, but the design is intended to be flexible enough to produce a binary-per-command or to dynamically link libgit2 if this proves useful.

3. Two command-line tools were introduced, `cat-file` and `hash-object`.  I chose these since they're relatively simple - one of the few examples of git commands that _more or less_ do a single thing - and having the ability to inspect the ODB is going to be extremely useful as we start looking at SHA256.


ChronicallySerious:
I am not a libgit2 developer but I use libgit2 as a tool in both my hobby and work, so just wanted to get some understanding of this change. 

If I understand correctly, this is going to be a CLI which allows interfacing to libgit2 functions directly? So is it like a replacement for git, but for power users? (Power user meaning users that would use it like Git but with more explicit plumbing commands)

Or is it a completely CI helper sort of thing which makes CI and testing easier?

And is there a possibility that the CLI ends up being a faster, more advanced replacement for Git?

Thanks! And thank you for maintaining this library!

NattyNarwhal:
I get the impression it's basically turning the examples (which are kinda simple reimplementations of Git commands or similar) into a more fully featured porcelain of its own. That'd be useful for CI, making sure git has independent reimplementations, etc.

I think this is a good idea for other libgit2 users; without the samples, I'd have been lost in using some functionality. Some stuff is also hairy enough (i.e. rebases) I'd like to see a sample to make sure I feel confident in my approach, so having a realistic implementation could be great there.

boretrk:
Overall I think this is a great idea. Being able to use some of gits unit tests would hopefully avoid a lot of issues people encounter.

I agree with @pks-t in #5507 that the util library seems a bit out of place, but maybe I misunderstand the purpose of it.
Is the idea that util mainly is a collection of functions that is shared between the cli and libgit2 or is the intention that all users should use functions in util too?

If it is the former it sort of removes the dogfooding aspect of this change.
IMO the ability to use gits unit tests still justifies the change.
https://github.com/libgit2/libgit2/blob/main/docs/projects.md indicates that an earlier idea was to go this way with the example code, but that leads to poor examples since most code will be unrelated to the library.
Having a separate cli and making the example code less functional and more explanatory would be better I think.

If it is the latter case, why is it better to have it in a separate area rather than with the rest of the library?
More specifically, how would a developer know if a function belongs in one or the other?

boretrk:
An alternative would be to keep the cli in a separate project. If it is "just another library user" then the cli and the adapted git tests can be kept separately.
The downside to that approach is that changes that breaks the cli tests won't be caught until they find their way into the cli project.

ethomson:
> Is the idea that util mainly is a collection of functions that is shared between the cli and libgit2

That's right.  Users should not be using the things in `util`, and we won't export them as public API.  It's things like our string manipulation functions.  But it's not shared state or only intended to be used by _just_ the cli and libgit2.

You can imagine that our CLI - really _any_ C application - needs some utility library to do things like at a level above the standard C library, and give us cross-platform mechanisms for doing so.  Working with strings, reading files, slightly more complex data structures like stacks and such.  We _could_ do this with something like [glib](https://docs.gtk.org/glib/).  And, in fact, if we were to reboot libgit2 then using glib would probably be smart.  But I don't really want to add a new dependency for the CLI when we've already got our own utility functions to do these things, and in a way that the developers are comfortable with.

In fact, I think about the `util` folder in very much the same way I think about `glib`.  It's grown out of libgit2 to be its own standalone utility library (just like glib grew out of gimp).  And so it should be able to be its own standalone library that anybody could use.  I intend to pull a copy of that folder out, call it `gu` (for "git utilities") and `s/git_/gu_/g` and make it available as a standalone project.  I want this utility library in a different project that I'm working on 

So I'm looking at `util` as a standalone dependency that both libgit2 and the CLI depend on.  We could go "all in" on this plan and move util into `deps`, and give it its own name and unique function signatures (eg `gu_str` instead of `git_str`).  But ...  I don't know, it feels a little ... heavy handed?  Presumptuous?  

My goal was to take an incremental approach towards getting there.  But maybe what we have here is too noncommittal and vague.

> If it is the former it sort of removes the dogfooding aspect of this change.

I hope not!  The CLI is intended to only ever interact with the public functions of libgit2.  None of the utility functions are actually necessary to interact with libgit2, and we shouldn't ever share any data between the cli and libgit2 _through_ the utility library.

(This is actually more aspirational than fact today.  The error handling code _does_ share state because I haven't unravelled that piece yet.  I'm a bit torn on how to handle it - should the utility code to have its own separate error handling, or should it have none and users bring their own, or some optional pluggable model...  I just haven't found a model that I love yet, but we certainly could make a quick and dirty change to make it happen.)

Does this clarify the goals here or are you still concerned?

boretrk:
No, it clarifies the idea behind it pretty well.
I with that goal I don't really think it is a problem to move the functionality to deps at all but I also don't see it as something that needs to be hurried.

I'm not exactly sure how to phrase my concern regarding the dogfooding part.
I guess I think that it could be easy to be content with a library interface that is only easy to use as long as you have the util library, but as long as the library is available externally and/or one keeps that in mind then that might not be a problem.

carlosmn:
Sorry it took me so long to go through this, I had mapped out a couple of thoughts in my head and then I guess never typed it out, oops.

So overall, yeah it makes sense to have a cli even if it's just to show some pattern and/or make sure the API makes some sense. My only comment there is that, if the main focus of this repository is the library, I would have expected `src/` to directly contain the library code and the cli to be elsewhere (a subdir, or another dir in the top-level) but that's rather minor.

I would also be cautious with making the util library too general, lest you end up also trying to maintain another glib.

PR #6136, opened 2021-12-08
Draft: add blame progress callback

Murmele:
Add blame progress callback.

Originally created by the GitAhead creator. I would like to get it mainlined so upgrading to newer versions get easier

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,7 @@ This is release v1.3.0, "Zugunruhe".  This release includes only minor new featu
 ## New Features
 * Support custom git extensions by @ethomson in https://github.com/libgit2/libgit2/pull/6031
 * Introduce `git_email_create`; deprecate `git_diff_format_email` by @ethomson in https://github.com/libgit2/libgit2/pull/6061
+* Add blame progress callback
 
 ## Deprecated APIs
 * `git_oidarray_free` is deprecated; callers should use `git_oidarray_dispose`
--- a/include/git2/blame.h
+++ b/include/git2/blame.h
@@ -77,6 +77,19 @@ typedef enum {
 } git_blame_flag_t;
 
 /**
+ * Blame progress callback.
+ *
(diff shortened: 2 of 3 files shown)


boretrk:
Duplicate of #5728

extrawurst:
it is sad to see these contributions go unnoticed for so long that duplicates of Pas even pop up :(

boretrk:
In #5728 the author claims that the PRs that have been duplicated here aren't ready to be upstreamed.
GPL allows for it to be merged upstream against the authors wishes but wouldn't it be a good idea to figured out why they weren't ready back then and if anything has changed in that regard first?

extrawurst:
Let’s ping @hackhaslam then :)

hackhaslam:
My comment about not being ready to be upstreamed was a general statement about all of the remaining patches in the GitAhead fork of libgit2. I felt like they needed more work and I wasn't ready to defend them.

This specific change is probably fine as is for what it does. It could be fleshed out a bit to provide actual progress. The use case is to abort long running blame operations. For example, if you start a blame on a file, then navigate away to some other file, it's useful to abort the current blame operation first.

boretrk:
Well, that resolves that part at least.

As for why it hasn't been merged I'm not a maintainer so I can only speculate.
Looking at the number of issues and pull requests I suspect most of the problem is that not enough people help with reviewing others pull requests.

Another could be that this PR and the previous duplicates were part of a bunch of requests that were originally added without description, but that is fixed now.

What haven't been addressed is the concern about the naming being misleading in previous PR.
Can the callback reasonably be used for progress report and if not, can the callback be extended to make that possible? Otherwise a more accurate name might be better.

It could also be worth checking out https://github.com/libgit2/libgit2/blob/main/docs/contributing.md#pull-requests
Tests and API documentation seems to be missing.

Again, I'm not a maintainer so there might be other reasons for not merging this PR.

PR #6137, opened 2021-12-09
add context lines accessor

Murmele:
Add function to access context line settings


boretrk:
Duplicate of #5733

boretrk:
This could be a nice function to have, but it seems a bit specialized.
This is more a personal choice but I think it feels a bit more natural to have something like a `const git_diff_options * git_patch_diff_opts(const git_patch *patch);` function, much like `git_patch_get_delta()` to get information about how the patch was generated.
On the other hand, maybe that exposes too much?

PR #6138, opened 2021-12-09
refs: Speed up packed lookups.

ccstolley:
Currently ref lookups require loading the entire packed-refs file into
a hashmap in memory. For repos with large numbers of refs this can be
painfully slow.

This patch replaces the existing lookup code and instead mmap()'s the
packed-refs file and performs a binary search to locate the ref entry.
Git uses a similar approach.

The old hash table codepath is still used for unsorted packed-refs files.
A number of the test fixture repos were updated with the `sorted` trait so
that this new path is properly exercised.

This patch also fixes a minor bug where the "peeled" trait is never
parsed correctly.

### DIFF
--- a/src/refdb_fs.c
+++ b/src/refdb_fs.c
@@ -65,9 +65,13 @@ typedef struct refdb_fs_backend {
 	git_iterator_flag_t iterator_flags;
 	uint32_t direach_flags;
 	int fsync;
+	git_map packed_refs_map;
+	git_mutex prlock; /* protect packed_refs_map */
+	bool sorted;
 } refdb_fs_backend;
 
 static int refdb_reflog_fs__delete(git_refdb_backend *_backend, const char *name);
+static const char * packed_set_peeling_mode(const char *data, size_t data_sz, refdb_fs_backend *backend);
 
 GIT_INLINE(int) loose_path(
 	git_str *out,
@@ -137,27 +141,9 @@ static int packed_reload(refdb_fs_backend *backend)
 	scan = (char *)packedrefs.ptr;
 	eof  = scan + packedrefs.size;
 
-	backend->peeling_mode = PEELING_NONE;
-
-	if (*scan == '#') {
-		static const char *traits_header = "# pack-refs with: ";
-
-		if (git__prefixcmp(scan, traits_header) == 0) {
-			scan += strlen(traits_header);
-			eol = strchr(scan, '\n');
-
-			if (!eol)
-				goto parse_failed;
-			*eol = '\0';
-
-			if (strstr(scan, " fully-peeled ") != NULL) {
-				backend->peeling_mode = PEELING_FULL;
-			} else if (strstr(scan, " peeled ") != NULL) {
-				backend->peeling_mode = PEELING_STANDARD;
-			}
-
-			scan = eol + 1;
-		}
+	scan = (char *)packed_set_peeling_mode(scan, packedrefs.size, backend);
+	if (!scan) {
+		goto parse_failed;
 	}
 
 	while (scan < eof && *scan == '#') {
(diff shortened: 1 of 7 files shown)


src/refdb_fs.c
```diff
@@ -137,27 +141,9 @@ static int packed_reload(refdb_fs_backend *backend)
 	scan = (char *)packedrefs.ptr;
 	eof  = scan + packedrefs.size;
 
-	backend->peeling_mode = PEELING_NONE;
-
-	if (*scan == '#') {
-		static const char *traits_header = "# pack-refs with: ";
```
ccstolley, line 143: Note the trailing space on this constant. The header usually looks like 
```
# pack-refs with: peeled fully-peeled sorted 
```
Later, when we strstr() for " peeled ", we'll never find it, because that trailing space snips it off.

ethomson:
Thanks @ccstolley - I gave this a quick 👀 but I'd like to give it a more detailed examination since there's lots of pointer arithmetic and I don't trust myself to review it as carefully as I'd like in a quick pass.  This is exciting to have implemented!

cstolley:
Also fwiw we've been testing this patch on github.com against `git show-ref` output and have not observed any differences.

ccstolley:
Also fwiw we've been testing this patch on github.com against `git show-ref` output and have not observed any differences.

src/refdb_fs.c
```diff
@@ -65,9 +65,13 @@ typedef struct refdb_fs_backend {
 	git_iterator_flag_t iterator_flags;
 	uint32_t direach_flags;
 	int fsync;
+	git_map packed_refs_map;
+	git_mutex prlock; /* protect packed_refs_map */
+	bool sorted;
 } refdb_fs_backend;
 
 static int refdb_reflog_fs__delete(git_refdb_backend *_backend, const char *name);
+static const char * packed_set_peeling_mode(const char *data, size_t data_sz, refdb_fs_backend *backend);
```
ethomson, line 74:
```suggestion
static char *packed_set_peeling_mode(char *data, size_t data_sz, refdb_fs_backend *backend);
```

src/refdb_fs.c
```diff
@@ -466,10 +452,176 @@ static int ref_error_notfound(const char *name)
 	return GIT_ENOTFOUND;
 }
 
-static int packed_lookup(
-	git_reference **out,
-	refdb_fs_backend *backend,
-	const char *ref_name)
+static const char *packed_set_peeling_mode(
+        const char *data,
```
ethomson, line 456:
```suggestion
static char *packed_set_peeling_mode(
        char *data,
```

src/refdb_fs.c
```diff
@@ -137,27 +141,9 @@ static int packed_reload(refdb_fs_backend *backend)
 	scan = (char *)packedrefs.ptr;
```
ethomson, line 141:
```suggestion
	scan = packedrefs.ptr;
```

src/refdb_fs.c
```diff
@@ -137,27 +141,9 @@ static int packed_reload(refdb_fs_backend *backend)
...
-			} else if (strstr(scan, " peeled ") != NULL) {
-				backend->peeling_mode = PEELING_STANDARD;
-			}
-
-			scan = eol + 1;
-		}
+	scan = (char *)packed_set_peeling_mode(scan, packedrefs.size, backend);
+	if (!scan) {
+		goto parse_failed;
 	}
```
ethomson, line 147:
```suggestion
	if (!scan)
		goto parse_failed;
```

src/refdb_fs.c
```diff
@@ -466,10 +452,176 @@ static int ref_error_notfound(const char *name)
...
-	git_reference **out,
-	refdb_fs_backend *backend,
-	const char *ref_name)
+static const char *packed_set_peeling_mode(
+        const char *data,
+        size_t data_sz,
+        refdb_fs_backend *backend)
+{
+	static const char *traits_header = "# pack-refs with:";
+	const char *eol;
```
ethomson, line 461:
```suggestion
	char *eol;
```

src/refdb_fs.c
```diff
@@ -466,10 +452,176 @@ static int ref_error_notfound(const char *name)
...
+
+/*
+ * Find beginning of packed-ref record pointed to by p.
+ *   buf - a lower-bound pointer to some memory buffer
+ *   p - an upper-bound pointer to the same memory buffer
+ */
+static const char *start_of_record(const char *buf, const char *p)
+{
+	const char *nl = p;
+	while (1) {
```
ethomson, line 559:
```suggestion
	while (true) {
```

src/refdb_fs.c
```diff
@@ -494,6 +646,86 @@ static int packed_lookup(
...
+
+	if ((error = packed_map_check(backend)) < 0)
+		return error;
+
+	if (!backend->sorted) {
+		return packed_unsorted_lookup(out, backend, ref_name);
+	}
+
+	left = backend->packed_refs_map.data;
+	right = data_end = ((const char *)backend->packed_refs_map.data) +
```
ethomson, line 665: This cast should be unnecessary?  🤔 
```suggestion
	right = data_end = backend->packed_refs_map.data +
```

src/refdb_fs.c
```diff
@@ -466,10 +452,176 @@ static int ref_error_notfound(const char *name)
...
+	end = memchr(rec, '\n', data_end);
+	if (end) {
+		data_end = end - rec;
+	}
+
+	cmp_val = memcmp(rec, ref_name, min(ref_len, data_end));
+
+	if (cmp_val == 0 && data_end != ref_len) {
+		return (data_end > ref_len) ? 1 : -1;
+	}
```
ethomson, line 617:
```suggestion
	if (cmp_val == 0 && data_end != ref_len)
		return (data_end > ref_len) ? 1 : -1;
```

src/refdb_fs.c
```diff
@@ -466,10 +452,176 @@ static int ref_error_notfound(const char *name)
...
+	if (data_end < GIT_OID_HEXSZ + 3) {
+		/* an incomplete (corrupt) record is treated as less than ref_name */
+		return -1;
+	}
+	data_end -= GIT_OID_HEXSZ + 1;
+
+	end = memchr(rec, '\n', data_end);
+	if (end) {
+		data_end = end - rec;
+	}
```
ethomson, line 611:
```suggestion
	if (end)
		data_end = end - rec;
```

src/refdb_fs.c
```diff
@@ -466,10 +452,176 @@ static int ref_error_notfound(const char *name)
...
+ *   p - a lower-bound pointer to the same memory buffer
+ */
+static const char *end_of_record(const char *p, const char *end)
+{
+	while (1) {
+		size_t sz = end - p;
+		p = memchr(p, '\n', sz);
+		if (!p) {
+			return end;
+		}
```
ethomson, line 584:
```suggestion
		if (!p)
			return end;
```

src/refdb_fs.c
```diff
@@ -466,10 +452,176 @@ static int ref_error_notfound(const char *name)
...
+	}
+
+	if (backend->packed_refs_map.data) {
+		git_mutex_unlock(&backend->prlock);
+		return error;
+	}
+
+	fd = git_futils_open_ro(backend->refcache->path);
+	if (fd < 0) {
+		git_mutex_unlock(&backend->prlock);
```
ethomson, line 515: I might have a goto here...  this is just about the size of a function where it has enough return points that reading through to make sure that we've unlocked the mutex on all of them is not _hard_, exactly, but a little meh.  Having a `goto done` or the like helps (me, at least) reason a little bit better about that.  Though this is rather stylistic, I concede.

src/refdb_fs.c
```diff
@@ -466,10 +452,176 @@ static int ref_error_notfound(const char *name)
...
+
+static int packed_map_check(refdb_fs_backend *backend)
+{
+	int error = 0;
+	git_file fd = -1;
+	struct stat st;
+
+	if ((error = git_mutex_lock(&backend->prlock)) < 0) {
+		return error;
+	}
```
ethomson, line 506:
```suggestion
	if ((error = git_mutex_lock(&backend->prlock)) < 0)
		return error;
```

src/refdb_fs.c
```diff
@@ -466,10 +452,176 @@ static int ref_error_notfound(const char *name)
...
+			backend->peeling_mode = PEELING_FULL;
+		} else if (git__memmem(data, eol - data, peeled, strlen(peeled))) {
+			backend->peeling_mode = PEELING_STANDARD;
+		}
+
+		if (git__memmem(data, eol - data, sorted, strlen(sorted))) {
+			backend->sorted = true;
+		} else {
+			backend->sorted = false;
+		}
```
ethomson, line 491:
```suggestion
		backend->sorted = git__memmem(data, eol - data, sorted, strlen(sorted));
```

src/refdb_fs.c
```diff
@@ -466,10 +452,176 @@ static int ref_error_notfound(const char *name)
...
+		eol = memchr(data, '\n', data_sz);
+
+		if (!eol)
+			return NULL;
+
+		if (git__memmem(data, eol - data, fully_peeled, strlen(fully_peeled))) {
+			backend->peeling_mode = PEELING_FULL;
+		} else if (git__memmem(data, eol - data, peeled, strlen(peeled))) {
+			backend->peeling_mode = PEELING_STANDARD;
+		}
```
ethomson, line 485:
```suggestion
		if (git__memmem(data, eol - data, fully_peeled, strlen(fully_peeled)))
			backend->peeling_mode = PEELING_FULL;
		else if (git__memmem(data, eol - data, peeled, strlen(peeled)))
			backend->peeling_mode = PEELING_STANDARD;
```

src/refdb_fs.c
```diff
@@ -466,10 +452,176 @@ static int ref_error_notfound(const char *name)
...
+        size_t data_sz,
+        refdb_fs_backend *backend)
+{
+	static const char *traits_header = "# pack-refs with:";
+	const char *eol;
+	backend->peeling_mode = PEELING_NONE;
+
+	if (data_sz == 0 || *data != '#') {
+		return data;
+	}
```
ethomson, line 466:
```suggestion
	if (data_sz == 0 || *data != '#')
		return data;
```

src/refdb_fs.c
```diff
@@ -494,6 +646,86 @@ static int packed_lookup(
...
+{
+	int error = 0;
+	const char *left, *right, *data_end;
+
+	if ((error = packed_map_check(backend)) < 0)
+		return error;
+
+	if (!backend->sorted) {
+		return packed_unsorted_lookup(out, backend, ref_name);
+	}
```
ethomson, line 662:
```suggestion
	if (!backend->sorted)
		return packed_unsorted_lookup(out, backend, ref_name);
```

ethomson (commented):
Many apologies on the delay in reviewing this.

I have a couple of stylistic nits.  I don't love casting away `const`ness when possible, so I'd suggest that `packed_set_peeling_mode` just deals with mutable `char *` since that's a bit more honest.  This is really the only thing that I would really prefer to see changed.  Everything else is quite minor and a "since I'm making suggestions, I might as well add a few more" type of suggestion.

There's also a cast "up" to `const char *` which I think should be unnecessary.  Otherwise, there's just a few places where a single line conditional consequence has braces around it, which is generally something that we try to avoid.  It looks like those might have existed already and come from refactoring.  Not that important but since I'm suggesting some changes anyway, I thought that I'd suggest those, too.

src/refdb_fs.c
ethomson: Actually ,looking at this again, this feels a little unnecessary?  `git__prefixncmp` should catch both of these cases, and quickly enough that a separate test shouldn't be necessary.  🤔

ethomson:
Sorry, one question that I have as I've read through this is about the persistent mmap that happens.  It looks like there's no "cache busting" mechanism if there's a new packed-refs file on disk, is that correct?  That could only happen on non-Windows, I think, because I would expect Windows to lock the packed-refs file for the duration of the fs backend's existence, which also seems not great.

🤔 

How much worse is it to do the `mmap` on every `lookup`?  Since this is the simpler solution (than maintaining a mutex), I assume that you tried this first and it was not sufficiently speedy...?  I wonder if we should have a persistent mmap mode (only available on non-Windows?) and a mmap-per-lookup?  🤔

src/refdb_fs.c
ccstolley: I agree, I think we can remove this test entirely. Thanks!

ccstolley:
@ethomson In cdf05a1 I decided to avoid `mmap()` completely on Windows, and just copy it into a buffer in memory instead. This still gives us a nice speed-up over the old code, avoids file locking issues and keeps the majority of the codepaths the same on both platforms. Let me know what you think.

In ac07a40 I added a filestamp check for `packed-refs`, to address your concerns about the file changing underneath us.

NellWhite (approved):

NellWhite (approved):

ethomson:
Thanks @ccstolley for this work - this is a great PR with a definite improvement.  Really appreciate it.  💯

PR #6139, opened 2021-12-10
added flag to disable mmap

Murmele:
This was originally created for the GitAhead GUI. Maybe it is done for performance reasons.

### DIFF
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -162,6 +162,9 @@ typedef enum {
 	 */
 	GIT_DIFF_SHOW_UNMODIFIED = (1u << 26),
 
+	/* Disable mmap in patches generated from this diff.*/
+	GIT_DIFF_DISABLE_MMAP = (1u << 27),
+
 	/** Use the "patience diff" algorithm */
 	GIT_DIFF_PATIENCE = (1u << 28),
 	/** Take extra time to find minimal diff */
--- a/src/diff_file.c
+++ b/src/diff_file.c
@@ -348,7 +348,7 @@ static int diff_file_content_load_workdir_file(
 		goto cleanup;
 
 	/* if there are no filters, try to mmap the file */
-	if (fl == NULL) {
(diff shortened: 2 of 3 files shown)


ethomson:
Maybe it is?  I'd love to know more about the rationale for this.

Murmele:
@hackhaslam I hope I don't disturb you but maybe you can say more about the reason why disabling mmap might be useful

hackhaslam:
If I remember correctly, mapping the file keeps the file handle/descriptor open until it's unmapped. When I stored the status diff, it would keep all of those files open and prevent other processes from modifying them. I think that this was mostly (or exclusively?) a problem on Windows. The alternative probably would have been to copy the diff into some other data structure and free it right away. It's been quite a while, so I could be mistaken about any of this.

Murmele:
@hackhaslam Thank you for the response! At least we have something and I can trying out to find more about it. Maybe it is not needed anymore.

inkydragon:
> This looks fine except this probably re-opens #2338 which is why we're so mmap-heavy. The issue here is that Windows may keep different views to a networked file for the mmap- and open-accessed versions of the file, so writing via `write()` is not reflected on the data `mmap()` is able to read.
> 
> IIUC this is a Windows quirk so **you could disable the mmap stuff on non-Windows systems** to keep it working there.
>
> _Originally posted by `@carlosmn` in https://github.com/libgit2/libgit2/issues/6039#issuecomment-915006722_

Maybe this option should be marked for non-windows use.
And warn people that this option may cause some problems if used on Windows systems.

boretrk:
To me it feels a bit strange to have it as an option to diff.
If the case is that mmap works on some systems but not others then shouldn't it be set on compile-time rather than being a runtime flag?
Or is there some benefit to being able to disable mmap for one call and not another?

ethomson:
> Or is there some benefit to being able to disable mmap for one call and not another?

I'll second this question - this feels like a place where we'd want to disable mmap en masse, not just for diff.  🤔  Let me know if I'm thinking about this wrong.

PR #6141, opened 2021-12-10
CMake modules not found when using libgit2 as cmake submodule

Murmele:
Use Project variable, otherwise when using libgit2 as submodule in another project, the paths are incorrect

When using libgit2 as cmake submodule in another project the following variables are set incorrectly:
- CMAKE_SOURCE_DIR
- CMAKE_BINARY_DIR
They are set to the source dir of the main project not of libtgit2


ethomson:
Hmm.  Looks like there's a few build failures - our Xenial images have an older cmake, possibly a compatibility problem?

Murmele:
I will check out what's the problem

Qix-:
It's because you missed a number of usages of `CMAKE_SOURCE|BINARY_DIR`. #6146 supercedes this PR and has all of the variables fixed correctly.

Qix-:
@ethomson this can be closed since it was fixed in #6146.

ethomson:
Thanks @Murmele and @Qix- - apologies for the inconvenience.

PR #6144, opened 2021-12-13
SHA256: add a SHA256 implementation backend

ethomson:
This adds SHA256 backend support for all the supported current backends.  This is a necessary precondition for git sha256 support, but does not actually add that.


src/hash/rfc6234/sha.h
```diff
@@ -0,0 +1,355 @@
...
+   conditions are met:
+
+   - Redistributions of source code must retain the above
+     copyright notice, this list of conditions and
+     the following disclaimer.
+
+   - Redistributions in binary form must reproduce the above
+     copyright notice, this list of conditions and the following
+     disclaimer in the documentation and/or other materials provided
+     with the distribution.
```
boretrk, line 18: The libgit2 linking exception does not require this so using this would require a license change?

https://github.com/amosnier/sha-2 have a more permissive license, but I haven't checked to see if the code is usable.

CMakeLists.txt
```diff
@@ -25,7 +25,8 @@ option(USE_NSEC                "Support nanosecond precision file mtimes and cti
 # Backend selection
 option(USE_SSH                 "Link with libssh2 to enable SSH support"               OFF)
 option(USE_HTTPS               "Enable HTTPS support. Can be set to a specific backend" ON)
-option(USE_SHA1                "Enable SHA1. Can be set to CollisionDetection(ON)/HTTPS/Generic" ON)
+option(USE_SHA1                "Enable SHA1. Can be set to CollisionDetection(ON)/HTTPS" ON)
+option(USE_SHA256              "Enable SHA256. Can be set to HTTPS/Builtin" ON)
```
boretrk, line 29: TBH I don't see a good use case for having different settings for SHA1 and SHA256.
The way I see it is that if you want it to be standalone you are going to use builtin on both, if you are using openssl you are going to do so on both and so on.
Possibly it could be of interest to turn SHA256 support off for size/performance/portability reasons, but unless that is something that will be supported then I think it is better to just have one option.
That would make it possible to share conditionals in the cmakefile and put the code in the same ifdefs.

boretrk: Is there any particular reason to why these options are booleans?

ethomson: libgit2 _should_ be built with SHA1DC, so the recommended solution is to use SHA1DC ("builtin") and whatever SHA256 comes with your HTTPS stack.

When I suggested that we get rid of the ability to actually use any of the other SHA1 implementations, I got pushback.  I forget by whom.  But there are definitely people using them.  🙁 

But I think that it was useful in a sense - we kept the other implementations around so that the code didn't bitrot completely until we could attack SHA256, because we'd want it back.  (The Windows implementation is particularly egregious and we'd certainly be inclined to simplify the whole hash subsystem if we were to remove it.)

We haven't taken quite the strong stance on this that [Linus recommended](https://lore.kernel.org/git/CA+55aFwSe9BF8e0hLk9pp3FVD5LaVY5GRdsV3fbNtgzekJadyA@mail.gmail.com/):

> Btw, the one thing I *would* suggest is that the git community just also says that the current hash is not SHA1, but SHA1DC.
> 
> Support for "plain" SHA1 should be removed entirely. If we add a lot of new infrastructure to support a new more secure hash, we should not have the old fallback for the known-weak one. Just make SHA1DC the only one git can be built with.

I'm not opposed to ripping the bandaid off and mandating SHA1DC as our hash, though.

Ultimately, I agree that there are a lot of knobs here.  Right now I hope that we have good defaults so that you don't ever have to turn them.

I hoped to simplify our cmake setup so that you would never accidentally get into a situation where you're trying to use Windows for HTTPS, OpenSSL for SHA1, mbedTLS for SHA256 and builtin crypto for NTLM.

I think that the simplest option would be a single knob for "choose your crypto stack".  I'm happy to be more opinionated here, but what I don't want is to have a "choose your crypto stack" option and then a "let's disable SHA256" option.  I feel like that would get us to a place where we have the same number of knobs but fewer choices.

Happy to hear your thoughts here.

As for the "why they're booleans", they're not (really).  "ON" and "OFF" are legal, where "ON" sets it to our default.  You can also specify a particular backend, or "HTTPS".  Maybe these should properly be `set(... CACHE STRING ...)` - this is one of those cmake things where it seems to work well enough, and this is the way that we've done it in the past.

src/hash/rfc6234/sha.h
ethomson: I don't think so.  I think that our code is still GPL + Linking Exception but this component is license compatible with us. There's a lot of prior art here that we've included already.  (See the `COPYING` file.)

We do need to include this license in the `COPYING` file so that end-users can ~~bury it~~ include it in their documentation.

CMakeLists.txt
boretrk: Oh, I only wondered about the boolean value because I once wanted to try different options and ccmake made things impractical.
As for the "turning off SHA256" remark I hope that didn't sound like a feature request. I was mostly looking for situations where different settings could be desirable. Having such an option would probably be annoying to maintain.

Given the recommendation from Linus I can only assume people want different implementations because they are more optimized, but it sounds like using them should come with a big warning that they aren't git compatible.
As far as I am concerned dropping everything but SHA1DC for the SHA1 hashes seems like the "correct" thing to do.

So my vote goes to always using SHA1DC and only having an option to select what SHA256 implementation to use.
With dynamically loaded OpenSSL I can see that one could want the builtin SHA256 since all local operations could be done without loading OpenSSL.

ethomson: > and ccmake made things impractical.

Doh.  I'm not super familiar with ccmake but yeah, this feels like the risk in using the wrong type.  😢  I think that we should probably be more thoughtful about this.

boretrk: In this particular case it was just me not knowing that that particular setting wasn't supposed to be used.
Changing type to INTERNAL will prevent it from showing up in ccmake but it will still be settable with -D, but it also implies FORCE so it might not be a good alternative...
I think the current way to handle sha1 is fine, or rather changing it doesn't necessarily need to be done in this PR.
I do however think that updating licensing information in docs/contributing.md and COPYING should be in the commit that pulls in the SHA256-algorithm.

PR #6146, opened 2021-12-23
cmake: use project-specific root variable instead of CMAKE_SOURCE_DIR

Qix-:
This effectively reverts 84083dcc8bd41332ccac9d7b537f3e254d79011c, which broke all users of libgit2 that use it as a CMake subdirectory (via `add_subdirectory()`), and uses `PROJECT_SOURCE_DIR`/`PROJECT_BINARY_DIR` instead. This is because `CMAKE_SOURCE_DIR` refers to the root-most CMake directory, which in the case of `add_subdirectory()` is a parent project to libgit2 and thus the paths don't make any sense to the configuration files. 

Corollary, `CMAKE_SOURCE_DIR` only makes sense if the CMake project is always the root project - which can rarely be guaranteed.

In all honesty, CMake should deprecate and eventually remove `CMAKE_SOURCE_DIR` and `CMAKE_BINARY_DIR`. It's been the source of headaches and confusion for years, they're rarely useful over `CMAKE_CURRENT_(SOURCE|BINARY)_DIR`, and they cause a lot of confusing configuration and source code layouts to boot.

Any time they are used, they break `add_subdirectory()` almost 100% of the time, cause confusing error messages, and hide subtle bugs.


Qix-:
FWIW I asked about this 5 years ago and they're pretty adamant that CMake isn't confusing here.

https://gitlab.kitware.com/cmake/cmake/-/issues/16255

So it isn't libgit2's fault. CMake is just insanely confusing.

ethomson:
Apologies for breaking you here, and thanks for the additional information.

Qix-:
No problem at all :) Thanks for the quick turnaround! I love this library, thanks for all the work put into it. <3

PR #6150, opened 2021-12-24
Remove PSGit from Language Bindings section of README

cestrand:
PSGit is not using libgit2 library. It is stub of a library and is not developed further.


ethomson:
Looks like you’re right - thanks!

PR #6151, opened 2021-12-25

csware:
Resurrecting #5150
Updated #5773

In-between the initial PR and now there have been several changes to Git for Windows >= 2:
* up to 2.24 an architecture specific subfolder was necessary for the etc and usr folders
* starting from 2.24 the etc folder is located directly in the installation folder but the usr folder is still in an architecture specific subfolder

### DIFF
--- a/src/win32/findfile.c
+++ b/src/win32/findfile.c
@@ -70,6 +70,46 @@ static wchar_t *win32_walkpath(wchar_t *path, wchar_t *buf, size_t buflen)
 	return (path != base) ? path : NULL;
 }
 
+static int win32_find_git_for_windows_architecture_root(_findfile_path *root_path, const wchar_t *subdir)
+{
+	/* Git for Windows >= 2 comes with a special architecture root (mingw64 and mingw32)
+	 * under which the "share" folder is located, check which we need (none is also ok) */
+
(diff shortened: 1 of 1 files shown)


src/win32/findfile.c
```diff
@@ -183,6 +232,12 @@ int git_win32__find_system_dirs(git_str *out, const wchar_t *subdir)
 			&buf, HKEY_CURRENT_USER, REG_MSYSGIT_INSTALL_LOCAL, subdir) && buf.size)
 		git_str_join(out, GIT_PATH_LIST_SEPARATOR, out->ptr, buf.ptr);
 
+#ifdef _WIN64
```
ethomson, line 235: Is this defined by mingw also, or just MSVC?

csware: What would be the alternative for mingw?

csware: Could we write `#if defined(_WIM64) || defined(__MINGW64__)`?

ethomson: I think that we can just do `GIT_ARCH_64` here - this file will only ever be included for Windows.

csware: Fixed.

ethomson:
Thanks!  I merged this manually.

PR #6152, opened 2021-12-26
Don't guess/search PROGRAMDATA\Git config any more

csware:
Since Git for Windows >= 2.24 this is not used any more.

I'm unsure whether we should proceed to drop even more PROGRAMDATA related code.


ethomson:
Thanks! Getting this ready for v2.0.

PR #6153, opened 2021-12-29
Disabling setting `CMAKE_FIND_LIBRARY_SUFFIXES` on Apple platforms.

arroz:
On macOS, since Big Sur, the libraries were moved to a cache. The SDK comes
with stubs in the SDK (`/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/lib/`
or whatever SDK version one has installed) where most have the `.tbd` suffix
(although some still are `.a`). Forcing `CMAKE_FIND_LIBRARY_SUFFIXES` on Apple
platforms broke building, unless one has copies of the libraries installed
elsewhere (like Brew), as many libraries (like `iconv` or `pcre`) are not
found.

This fix disables setting the `CMAKE_FIND_LIBRARY_SUFFIXES` to `.a` if
the platform is `APPLE` when building static libs.


arroz:
Note: this wasn't caught by the CI since it builds the dynamic lib, not the static one, on macOS, I think. Otherwise, I think it would have failed to build the tests since they require linking with `libiconv`.

arroz:
I'm wondering if this check should be there at all. Even if the user wants a static version of libgit2, that doesn't necessarily mean they want to include the static versions of the necessary system libs (zlib, iconv, etc), and that may bring license issues with it.

In my use case, the reason I'm building static versions of OpenSSL/libgit2/libssh2 is just to avoid nightmarish problems linking my program against those libs, and not to make an executable that doesn't depend on anything else. I'm happy offloading to the built-in libs in the OS as much as I can. :)

ethomson:
> I'm wondering if this check should be there at all. Even if the user wants a static version of libgit2, that doesn't necessarily mean they want to include the static versions of the necessary system libs (zlib, iconv, etc), and that may bring license issues with it.

🤔  I agree with you.  I think that there's two use cases here - one, I want to build libgit2 as a static library that also statically includes its dependencies, and two, I want to build libgit2 as a static library but want to use the system dependencies.  It seems reasonable to have two different options, but I think that it probably should be two _different_ options to avoid breaking anybody who's relying on the existing behavior.

arroz:
Yeah, I agree. I wonder if there should be a specific option for including the static dependencies, or if that can be controlled with arguments passed to the cmake command already.

There's the additional issue with macOS where restricting based on the file name extension wont work, due to the shared cache + stubs that Apple introduced in Big Sur. In that situation, I don't even know if the static versions of the libraries are available anywhere, but if they are, that is likely controlled by arguments passed to the linker. The CI server works around that by using the libraries from Brew, but if those are not used, currently the macOS build cannot find iconv. It still builds without it (although I assume it has problems with HFS names, and maybe APFS as well?), but if I'm not mistaken, the test suite fails to build because of that.

I'm not entirely sure about the best way to fix this, as I don't know CMAKE at all (this patch is the only CMAKE line I wrote in my life 😄). Let me know if you prefer to handle this.

arroz:
Update PR introducing `LINK_WITH_STATIC_LIBRARIES`. This way, disabling `BUILD_SHARED_LIBS` will build static libs dynamically linked against the system libraries, and on top of that, enabling `LINK_WITH_STATIC_LIBRARIES` forces linking only with static libs.

Regarding macOS: I checked on Catalina (before Apple introduced the big lib cache in Big Sur) and even Catalina didn't come with the `.a` versions of the libraries. So linking statically against system libraries is not possible anyway, unless they are installed through Brew, like the CI server is doing. With this patch, disabling `BUILD_SHARED_LIBS` allows building static libs, but wont break linking with `iconv` bundled with the OS and SDK.

arroz:
Hi @ethomson, just checking if you think this is good enough to merge or if further work is needed. Thank you!

ethomson:
Thanks for the ping @arroz - this looks good to me!

arroz:
Thank you! 😊

Issue #6154, opened 2021-12-30
`git_status_list_new` fails to merge h2i and i2wc changes when used in case insensitive repos/file systems


arroz:
This is a bit of an edge case, but I happened to hit it:

### Reproduction steps
On a repository configured with `ignorecase = true`, do the following:
1) Create two files, for example `file.txt` and `Wow.txt` (make sure the second one starts with an upper case letter, and that when sorted in a case insensitive way, the file with lower case name comes before the one with upper case).
2) Stage and commit them.
3) Delete the `Wow.txt` file, and `git add` that change.
4) Rename `file.txt` to `file-renamed.txt`, and `git add` that change.
5) Edit `file-renamed.txt` and change the file.
6) Run code similar to the following:

```
git_status_list *statuslist;
git_status_options opts = GIT_STATUS_OPTIONS_INIT;
opts.flags |= GIT_STATUS_OPT_RENAMES_INDEX_TO_WORKDIR;
opts.flags |= GIT_STATUS_OPT_RENAMES_HEAD_TO_INDEX;
git_status_list_new(&statuslist, repo, &opts);
```

The list will contain 3 entries, instead of the expected 2, with `file-renamed.txt` repeated (one entry for the HEAD->index change, and another for the index->workdir change).

### Expected behavior
There should be only two entries in the status list. The entry for `file-renamed.txt` should have deltas for both HEAD->index and index->workdir).

### Cause
This is caused by `apply_splits_and_deletes` failing to initialize the `onto` vector with the correct comparison function (it should obtain it from the `diff->deltas` vector, instead of using the default one). I'll submit a PR with fix + test soon.

ethomson:
Thanks for the fix!

PR #6155, opened 2021-12-30
#6154 git_status_list_new case insensitive fix

arroz:
Fixes issue #6154.


arroz:
I'm closing this and re-submitting, with a better implementation of the test, to avoid all the mess in between and make history clean.

Issue #6134, opened 2021-12-06
devel/libgit2 1.3.0 fails to build, error: static declaration of 'HMAC_CTX_cleanup' follows non-static declaration


Nuadh:
libgit2-1.3.0 fails to build on FreeBSD 12.2-RELEASE-p11, due to me using libressl instead of openssl, with the following error:

/wrkdirs/usr/ports/devel/libgit2/work/libgit2-1.3.0/deps/ntlmclient/crypt_openssl.c:49:20: error: static declaration of 'HMAC_CTX_cleanup' follows non-static declaration
static inline void HMAC_CTX_cleanup(HMAC_CTX *ctx)
                   ^
/usr/local/include/openssl/hmac.h:90:6: note: previous declaration is here
void HMAC_CTX_cleanup(HMAC_CTX *ctx);

This issue was introduced with commit 5158b0b70ade89268b22b7c388802b5f5b6debce.

HMAC_CTX_cleanup was deprecated for public use and removed in a patch in 2015 (with version 1.1.0, I think) in openssl. For some reason, libressl still have this around. Attached is a (very minor) patch I used to "fix" this.

[patch-deps_ntlmclient_crypt__openssl.c.txt](https://github.com/libgit2/libgit2/files/7662305/patch-deps_ntlmclient_crypt__openssl.c.txt)
.

Nuadh:
Actually working patch..

[patch-deps_ntlmclient_crypt__openssl.c.txt](https://github.com/libgit2/libgit2/files/7698092/patch-deps_ntlmclient_crypt__openssl.c.txt)

belotn:
Hello,

Same here, `/usr/ports/devel/libgit2/work/libgit2-1.3.0/deps/ntlmclient/crypt_openssl.c:49:20: error: static declaration of 'HMAC_CTX_cleanup' follows non-static declaration
static inline void HMAC_CTX_cleanup(HMAC_CTX *ctx)`

ethomson:
Thanks, this should be fixed by #6157.  Let me know if not!

PR #6157, opened 2021-12-31
ntmlclient: fix linking with libressl

boretrk:
Made a PR from the patch provided by @Nuadh in #6134

libressl still declares HMAC_CTX_cleanup so we can't create the dummy function in that case.

### DIFF
--- a/deps/ntlmclient/crypt_openssl.c
+++ b/deps/ntlmclient/crypt_openssl.c
@@ -44,7 +44,7 @@ static inline void HMAC_CTX_free(HMAC_CTX *ctx)
 
 #endif
 
-#if OPENSSL_VERSION_NUMBER >= 0x10100000L || defined(CRYPT_OPENSSL_DYNAMIC)
+#if (OPENSSL_VERSION_NUMBER >= 0x10100000L && !LIBRESSL_VERSION_NUMBER) || defined(CRYPT_OPENSSL_DYNAMIC)
 
 static inline void HMAC_CTX_cleanup(HMAC_CTX *ctx)
 {


ethomson:
Thanks!

botovq:
`HMAC_CTX_cleanup` will be [removed in LibreSSL 3.5][1], so this will break again.

The ifdef will be come a bit unwieldy:

```
#if ((OPENSSL_VERSION_NUMBER >= 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)) || LIBRESSL_VERSION_NUMBER >= 0x3050000f) || defined(CRYPT_OPENSSL_DYNAMIC)
```

Perhaps it would make more sense to test from cmake whether the OpenSSL lib provides `HMAC_CTX_new` /`HMAC_CTX_cleanup` or not?

[1]: https://github.com/openbsd/src/commit/d56fe77f5d8577f74e9d3de24ffbd0fded961ec0

Nuadh:
> `HMAC_CTX_cleanup` will be [removed in LibreSSL 3.5](https://github.com/openbsd/src/commit/d56fe77f5d8577f74e9d3de24ffbd0fded961ec0), so this will break again.
> 
> The ifdef will become a bit unwieldy:
> 
> ```
> #if ((OPENSSL_VERSION_NUMBER >= 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)) || LIBRESSL_VERSION_NUMBER >= 0x3050000f) || defined(CRYPT_OPENSSL_DYNAMIC)
> ```
> 
> Perhaps it would make more sense to test from cmake whether the OpenSSL lib provides `HMAC_CTX_new` /`HMAC_CTX_cleanup` or not?

Indeed that would be a better alternative:)

PR #6158, opened 2022-01-01
Add `rename_threshold` to `git_status_options`.

arroz:
Adds the `rename_threshold` option to `git_status_options`, that works in a similar way to `git_diff_find_options` and core git's `-M` option.


ethomson:
LGTM.  Thanks!

PR #6159, opened 2022-01-01
#6154 git_status_list_new case insensitive fix

arroz:
Fixes issue #6154.

(Sorry for the re-submission, improved test implementation so no new test repo had to be added).


ethomson:
I merged this manually to resolve the conflicts.  Thanks for the fix!  One thing to keep in mind is that we don't use `//` style comments in libgit2 - I cleaned those up manually.  (I should probably add a linter rule here.  🤔)

PR #6160, opened 2022-01-02
mktmp: use fchmod for changing mode

boretrk:
Using fchmod in cases where we have a filehandle allows PosixLib to work around the issue with being unable to change mode on open files on AmigaOS.
It also seems to me like it would be the better option for other systems too.
Unfortunately it seems like Windows doesn't have a direct equivalent?

### DIFF
--- a/src/futils.c
+++ b/src/futils.c
@@ -26,7 +26,7 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
 
 int git_futils_mktmp(git_str *path_out, const char *filename, mode_t mode)
 {
-	int fd;
+	int fd, r;
 	mode_t mask;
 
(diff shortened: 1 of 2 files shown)


ethomson:
I think that I'm happy with #5350 if we stop rolling our own temp names.  It makes me a shade uneasy.

boretrk:
I think there is a reluctance to use tempnam() based on the following text from [The Open Group Base Specifications Issue 7](https://pubs.opengroup.org/onlinepubs/9699919799/functions/tempnam.html)
> The tempnam() function is marked obsolescent.
An alternative could be using tmpnam() for the random part of the string since we build the prefix anyway. It is also marked obsolescent but since it is part of C99 it is unlikely to not be supported.
I'll look into it.

PR #6164, opened 2022-01-05
Fix typos

rex4539:


COPYING
```diff
@@ -1032,7 +1032,7 @@ by Eric Young (eay@cryptsoft.com).
 The implementation was written so as to conform with Netscapes SSL.
 
 This library is free for commercial and non-commercial use as long as
-the following conditions are aheared to.  The following conditions
+the following conditions are adhered to.  The following conditions
```
ethomson, line 1035: I don't feel good changing somebody else's copyright statement, even if it's to correct typos.  Has the upstream updated their copyright statement to fix the typos (or did we copy and paste this wrong?)  If their copyright statement contains a typo, even in the latest version, then that's just an error in their copyright statement that we should propagate.  It's a legal document that they produced that we should not alter.

rex4539: Understood. Reverted.

ethomson:
Thanks for fixing the typos!

PR #5833, opened 2021-04-03
Features/sparse checkout

J0Nes90:
### Sparse Checkout

This is a proposal for a very basic but functional sparse checkout implementation in libgit2. As this is a major feature of git I can imagine that I am missing certain edge cases in git and libgit2 especially. It lacks certain features (see below) but I hope this might be a good start to improve and build upon. 

### What does this do?
* Status / Diff takes `GIT_INDEX_ENTRY_SKIP_WORKTREE` into account
* Skips sparse files on checkout
* `sparse-checkout` file is treated as an attr_file, hence following gitignore rules
* Sparse checkout can be enabled with `core.sparseCheckout`
* Expose an interface function `git_sparse_check_path` to query whether or not a file is excluded from checkout
* Set `GIT_INDEX_ENTRY_SKIP_WORKTREE` flag on `git_index_read_tree` based on the sparse-checkout ruleset.

### What does it **not** do?
* Smart trickery to determine whether or not a full tree is sparse (would be a very welcome optimization)
* Support `core.sparseCheckoutCone`
* No support for `git sparse-checkout list`
* No direct support for` git sparse-checkout reapply` (but this can be achieved through checkout)
* No direct support for `git sparse-checkout add / set` (so the sparse-checkout file has to be filled manually)
* I guess no support for work trees? (I did not test this)
* No sparse checkout option on clone (but this can be achieved with `GIT_CHECKOUT_NONE` and `git_index_read_tree`)

PR #6169, opened 2022-01-12
Sparse Checkout

YuKitsune:
This PR aims to implement basic sparse checkout functionality into LibGit2 building on @jochenhz's [original PR](https://github.com/libgit2/libgit2/pull/5833).

# What does it do?
- Status / Diff takes GIT_INDEX_ENTRY_SKIP_WORKTREE into account
- `sparse-checkout` file is treated as an `attr_file`, hence following `gitignore` rules
- Skips sparse files on checkout
- Sparse checkout can be enabled with `core.sparseCheckout`
-  support `init`, `set`, `add`, `reapply` and `disable` commands all supported
- Expose an interface function `git_sparse_check_path` to query whether or not a file is excluded from checkout
- Set `GIT_INDEX_ENTRY_SKIP_WORKTREE` flag on `git_index_read_tree` based on the sparse checkout ruleset.

# What does it not do?
- Smart trickery to determine whether or not a full tree is sparse (would be a very welcome optimization)
- Support `core.sparseCheckoutCone`
- No sparse checkout option on clone (but this can be achieved with `GIT_CHECKOUT_NONE` and `git_index_read_tree`)

# Note
- I've got some tests around worktrees, they seem to work
- Given how complex sparse checkout can be, I'm almost certain I haven't gotten this implementation 100% correct, though feel free to prove me wrong 👍

### DIFF
--- a/tests/libgit2/refs/revparse.c
+++ b/tests/libgit2/refs/revparse.c
@@ -399,7 +399,7 @@ void test_refs_revparse__date(void)
 	 * a65fedf HEAD@{1335806603 -0900}: commit:
 	 * be3563a HEAD@{1335806563 -0700}: clone: from /Users/ben/src/libgit2/tests/resour
 	 */
-	test_object("HEAD@{10 years ago}", "be3563ae3f795b2b4353bcce3a527ad0a4f7f644");
+	test_object("HEAD@{11 years ago}", "be3563ae3f795b2b4353bcce3a527ad0a4f7f644");
 
 	test_object("HEAD@{1 second}", "a65fedf39aefe402d3bb6e24df4d4f5fe4547750");
 	test_object("HEAD@{1 second ago}", "a65fedf39aefe402d3bb6e24df4d4f5fe4547750");


src/sparse.c
```diff
@@ -0,0 +1,706 @@
...
+	git_str infopath = GIT_STR_INIT;
+    const char *filename = GIT_SPARSE_CHECKOUT_FILE;
+	git_attr_file_source source = { GIT_ATTR_FILE_SOURCE_FILE, git_str_cstr(&infopath), filename, NULL };
+	git_str filepath = GIT_STR_INIT;
+
+	if ((error = git_str_joinpath(&infopath, repo->gitdir, "info")) < 0) {
+		if (error != GIT_ENOTFOUND)
+			goto done;
+		error = 0;
+	}
```
YuKitsune, line 123: Would like to have used something like `git_repository__item_path` here, but it would always get the info path for the root repository, not the worktrees. Keen to know if there's a nicer way to do this.

include/git2/diff.h
```diff
@@ -444,6 +444,11 @@ typedef struct {
 	 * Defaults to "b".
 	 */
 	const char *new_prefix;
+
+	/** Skip files in the diff that are excluded by the `sparse-checkout` file.
+	 * Set to 1 to skip sparse files, 0 otherwise
+	 */
+	int skip_sparse_files;
```
boretrk, line 451: This one might require some discussion on how flags are handled.
There is a flags field, but it is almost full. Only  `(1u << 27)` is available.

Most of the flags are in enums, I think the reason for this is that it provides some type safety while still being passable as arguments to functions.
In this particular case the flags are never used as a parameter, the user passes the git_diff_options struct instead.
This means that it would be possible to specify the flags in bitfields as `unsigned int skip_sparse_files:1;`
Unfortunately we can't split out the existing enum to a regular bitfield without  being incompatible but IMO that would be a good thing to do for flags passed in a larger struct like this.

src/sparse.h
```diff
@@ -0,0 +1,39 @@
...
+	git_attr_file *sparse;
+	int ignore_case;
+} git_sparse;
+
+enum {
+	GIT_SPARSE_UNCHECKED = -2,
+	GIT_SPARSE_NOTFOUND = -1,
+	GIT_SPARSE_NOCHECKOUT = 0,
+	GIT_SPARSE_CHECKOUT = 1,
+};
```
boretrk, line 27: Could we give this enum a typedef so that the proper type can be used instead of int?

src/util.c
```diff
@@ -303,6 +303,22 @@ char *git__strtok(char **end, const char *sep)
...
 }
 
+char *git__strtok_(char **end, const char *sep, ...)
+{
+	char* res;
+	va_list seps;
+
+	va_start(seps, sep);
+
+	res = git__strtok(end, va_arg(seps, const char*));
```
boretrk, line 313: This can't be right.
`git__strtok_` is called with `buf = git__strtok_(&scan, "\r\n", "\n", "\0");`
`va_start` says that variable arguments starts after `sep`, so `seps` points to `"\n"`
Without any form of loop `git__strtok` is only called for `"\n"` and not for `"\r\n"` or `"\0"`

.gitignore
```diff
@@ -6,3 +6,4 @@
 CMakeSettings.json
 .vs
 .idea
+cmake-build-debug
```
boretrk, line 9: Does this really belong here?

src/checkout.c
```diff
@@ -390,8 +390,13 @@ static int checkout_action_wd_only(
 
 		if (wd->mode != GIT_FILEMODE_TREE) {
 			if (!error) { /* found by git_index__find_pos call */
-				notify = GIT_CHECKOUT_NOTIFY_DIRTY;
-				remove = ((data->strategy & GIT_CHECKOUT_FORCE) != 0);
+				const git_index_entry *e = git_index_get_byindex(data->index, pos);
+				if (e == NULL ||
+						(e->flags_extended & GIT_INDEX_ENTRY_SKIP_WORKTREE) == 0 ||
+						(data->strategy & GIT_CHECKOUT_REMOVE_SPARSE_FILES) != 0) {
```
boretrk, line 396: Do you think it would be possible to add some comment on what is going on here?

include/git2/diff.h
boretrk: Pinging in @ethomson since he probably have given some more thought on what to do when we run out of flags.

src/sparse.c
```diff
@@ -0,0 +1,709 @@
...
+    GIT_ASSERT_ARG(patterns);
+    GIT_ASSERT_ARG(sparse);
+
+    if ((error = git_futils_readbuffer(&data, sparse->sparse->entry->fullpath)) < 0)
+        return error;
+
+    scan = (char *)git_str_cstr(&data);
+    while (!error && *scan) {
+
+		buf = git__strtok_(&scan, "\r\n", "\n", "\0");
```
boretrk, line 260: This isn't how strtok works. strtok matches any character in the string.
`git__strtok(&scan, "\r\n")` is what you want here.
If either `'\r'` or `'\n'` is encountered it would split on them and empty segments are skipped over.
So "\r\n" would first match on '\r' and then skip over the '\n'.
It would also take "\n\n\r" or any combination of them.

Bit over all it seems a bit strange to have this splitting and parsing here.
https://www.git-scm.com/docs/git-sparse-checkout says:
> By default, the sparse-checkout file uses the same syntax as .gitignore files.

Shouldn't we aim to have the same function for parsing them?

I haven't really thought it through but I have this feeling that it could be possible to merge the spare checkout list and the ignore list for some checks but maybe there is something that prevents that.

src/util.c
YuKitsune: Thanks, I think I misunderstood how `git__strtok` worked when I was writing this.
I've since removed it.

src/sparse.c
YuKitsune: I've replaced `git__strtok_` with `git__strtok`, I couldn't find any "git ignore list"-like functions around, so I quickly threw this together. I'm open to some ideas on how we can make this list function a bit nicer.

I've also consolidated the ignore parsing functions, so that `parse_sparse_file` now uses `parse_ignore_file` instead. I'd like to keep the `parse_sparse_file` function around since cone mode should change the parsing behaviour a bit.

src/sparse.h
YuKitsune: I've called it `git_sparse_status` for now, though I'm open to some better name suggestions.

YuKitsune:
Bump.
Is there anything I can do to help progress this further?

VinnyOG:
@YuKitsune Is this PR in a state that it's working for you? I would like to patch it into a project that I'm working on and wondering if there are any rough edges I should know about.

Thanks for your work in continuing on this!

YuKitsune:
@VinnyOG We haven't gotten around to making use of this. Our use-case requires a new build of LibGit2Sharp, which proved to be too much of a pain.

extrawurst:
I wish this would be getting attention

YuKitsune:
I've mucked up the rebase and the PR auto-closed... 🤦

Issue #6170, opened 2022-01-12
libssh not found on head (but works with 1.3.0)


axkibe:
I cloned git head and tried to build it, out of some reason it just cannot find libssh2 (but libssh2-1-dev is installed). However it does work as expected out of the box when downloading libgit2-1.3.0.tar.gz so it must have been recently broken 

### Reproduction steps

mkdir build
cmake ..

### Expected behavior

cmake  find SSH, SSH transport support

### Actual behavior

-- Disabled features:
 ...
 * SSH, SSH transport support

### Version of libgit2 (release number or SHA1)

e2e3f3e4cde7871f86eabacc6a639439b70a8fc7

### Operating system(s) tested

debian bullseye

axkibe:
Okay it's just disabled now (I guess there was some reason for that)
It was only confusing since it reports it as not found instead of disabled as per (default) choice.
PR to make this more intuitive.

ethomson:
Thanks for the update - going to close this since you have a PR to improve the messaging.

axkibe:
It ain't merged tough.

PR #6171, opened 2022-01-12
report LIBSSH2 as not enabled instead of not found if it isn't enabled.

axkibe:
"fixes" #6170 report USE_SSH2 as disabled instead of not found.


cmake/SelectSSH.cmake
```diff
@@ -12,30 +12,32 @@ if(USE_SSH)
...
 
-	check_library_exists("${LIBSSH2_LIBRARIES}" libssh2_userauth_publickey_frommemory "${LIBSSH2_LIBRARY_DIRS}" HAVE_LIBSSH2_MEMORY_CREDENTIALS)
-	if(HAVE_LIBSSH2_MEMORY_CREDENTIALS)
-		set(GIT_SSH_MEMORY_CREDENTIALS 1)
+		check_library_exists("${LIBSSH2_LIBRARIES}" libssh2_userauth_publickey_frommemory "${LIBSSH2_LIBRARY_DIRS}" HAVE_LIBSSH2_MEMORY_CREDENTIALS)
+		if(HAVE_LIBSSH2_MEMORY_CREDENTIALS)
+			set(GIT_SSH_MEMORY_CREDENTIALS 1)
+		endif()
+	else()
+		message(STATUS "LIBSSH2 not found. Set CMAKE_PREFIX_PATH if it is installed outside of the default search path.")
```
boretrk, line 27: This message seems redundant with the FATAL_ERROR one above. (Won't both be output on the same condition?)

axkibe: One message is about USE_SSH not enabled in cmake, the other is about the package not being found.

I got very confused, because cmake told me it couldn't find libssh2 but it just wasn't enabled as config option (which is now off by default for some reason)

PS: I'm sorry there is now another commit to the PR I wasn't aware it would add to already existing PRs. Fixing it.

axkibe: Oh I see what you mean. I didn't introduce this redundancy, it's already in main.
See: https://github.com/libgit2/libgit2/blob/main/cmake/SelectSSH.cmake
 
What this commit adds is a different message in case of USE_SSH is OFF (which is default now). Otherwise this prints libssh2 not found, just because USE_SSH is off.

PR #6178, opened 2022-01-14
futils_mktmp: don't use umask

boretrk:
Modification of  #5350 with ~~tempnam()~~ mktemp() since tempnam() isn't supported on the macos and windows builds.

For some reason the test doesn't fail when used on its own. Maybe something have changed since then?

~~[The Open Group Base Specifications Issue 7](https://pubs.opengroup.org/onlinepubs/9699919799/functions/tempnam.html) marks tempnam() as obsolescent and states that "The tempnam() function may be removed in a future version."
It also limits the prefix to 5 bytes so appending "\_git2\_" to the prefix is a bit pointless.~~

### DIFF
--- a/src/futils.c
+++ b/src/futils.c
@@ -26,30 +26,32 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
 
 int git_futils_mktmp(git_str *path_out, const char *filename, mode_t mode)
 {
+	const int open_flags = O_RDWR | O_CREAT | O_EXCL | O_BINARY | O_CLOEXEC;
+	/* TMP_MAX is unrelated to mktemp but should provide a reasonable amount */
+	unsigned int tries = TMP_MAX;
 	int fd;
-	mode_t mask;
 
-	p_umask(mask = p_umask(0));
+	while (tries--) {
+		git_str_sets(path_out, filename);
+		git_str_puts(path_out, "_git2_XXXXXX");
 
-	git_str_sets(path_out, filename);
-	git_str_puts(path_out, "_git2_XXXXXX");
+		if (git_str_oom(path_out))
+			return -1;
 
-	if (git_str_oom(path_out))
-		return -1;
+		/* Using mktemp is safe when we open with O_CREAT | O_EXCL */
+		p_mktemp(path_out->ptr);
+		/* mktemp sets template to empty string on failure */
+		if (path_out->ptr[0] == '\0')
+			break;
 
-	if ((fd = p_mkstemp(path_out->ptr)) < 0) {
-		git_error_set(GIT_ERROR_OS,
-			"failed to create temporary file '%s'", path_out->ptr);
-		return -1;
+		if ((fd = p_open(path_out->ptr, open_flags, mode)) >= 0)
+			return fd;
 	}
 
-	if (p_chmod(path_out->ptr, (mode & ~mask))) {
-		git_error_set(GIT_ERROR_OS,
-			"failed to set permissions on file '%s'", path_out->ptr);
-		return -1;
-	}
-
-	return fd;
+	git_error_set(GIT_ERROR_OS,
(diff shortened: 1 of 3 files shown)


src/futils.c
```diff
@@ -26,27 +26,25 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
 
 int git_futils_mktmp(git_str *path_out, const char *filename, mode_t mode)
 {
-	int fd;
-	mode_t mask;
-
-	p_umask(mask = p_umask(0));
+	const int open_flags = O_RDWR | O_CREAT | O_EXCL | O_BINARY | O_CLOEXEC;
+	int fd = -1;
+	unsigned int tries = TMP_MAX;
```
boretrk, line 31: TMP_MAX isn't really related to mktemp, but there also isn't any suggested alternative

src/futils.c
```diff
@@ -26,27 +26,25 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
...
 	if (git_str_oom(path_out))
 		return -1;
 
-	if ((fd = p_mkstemp(path_out->ptr)) < 0) {
-		git_error_set(GIT_ERROR_OS,
-			"failed to create temporary file '%s'", path_out->ptr);
-		return -1;
+	while (fd == -1 && tries--) {
+		if (mktemp(path_out->ptr) == NULL)
+			break;
```
boretrk, line 41: Using mktemp could be problematic on some systems.
From [www.man7.org](https://www.man7.org/linux/man-pages/man3/mktemp.3.html):
> Some implementations follow 4.3BSD and replace XXXXXX by the current process ID and a single letter, so that at most 26 different names can be returned.

src/futils.c
```diff
@@ -26,30 +26,28 @@ int git_futils_mkpath2file(const char *file_path, const mode_t mode)
...
 {
+	const int open_flags = O_RDWR | O_CREAT | O_EXCL | O_BINARY | O_CLOEXEC;
+	unsigned int tries = TMP_MAX;
 	int fd;
-	mode_t mask;
 
-	p_umask(mask = p_umask(0));
+	while (tries--) {
+		git_str_sets(path_out, filename);
+		git_str_puts(path_out, "_git2_XXXXXX");
```
boretrk, line 35: We could help mktemp() out a bit by adding a couple of random number so the template string before passing it on.
```diff
-               git_str_sets(path_out, filename);
-               git_str_puts(path_out, "_git2_XXXXXX");
+               git_str_clear(path_out);
+               git_str_printf(path_out, "%s_git2_%04XXXXXXX", filename, rand());
```
It could be useful for systems that base the mktemp string on the PID.
As for possible DoS attacks I think that a malicious person with write access to the repository can do a bit more damage than just that.

src/posix.h
```diff
@@ -130,6 +131,7 @@ extern ssize_t p_pwrite(int fd, const void *data, size_t size, off64_t offset);
 
 #define p_close(fd) close(fd)
 #define p_umask(m) umask(m)
+#define p_mktemp(p) mktemp(p)
```
boretrk, line 134: Windows have different implementations for `p_mkstemp` depending on `_MSC_VER`, is that needed for `p_mktemp` too?

src/futils.c
ethomson: I wouldn't want to use `rand` - we can't be guaranteed that the caller ever invoked `srand` in a meaningful way and we wouldn't want to do it ourselves because they may be relying on an `srand` seed for some sort of reproducibility.

I appreciate your reminder that this is limited to the repository:

> a malicious person with write access to the repository can do a bit more damage than just that.

I'd been thinking about this function as a generic temporary file creation utility; since we don't use this to write into `/tmp`, I don't think that we need to be too careful about it.  I added a note on the header so that we remember its usefulness and merged it manually.

Thanks again for this!

PR #6179, opened 2022-01-17
c99: change single bit flags to unsigned

boretrk:
Just a minor fix, technically a set signed flag can return -1.


ethomson:
Nice one - thanks!

PR #6184, opened 2022-01-19
diff_driver: split global_drivers array into separate elements

boretrk:
C99 doesn't allow structures with flexible length array to be placed in an array.


ethomson:
👍

PR #6185, opened 2022-01-22
cmake: disable some gnu extensions

boretrk:
Setting `C_EXTENSIONS` to `OFF` turns off some extensions and will catch the C++ comments. Probably some more things, but nothing that we currently use.

Since it is on the git2internal target it won't apply do deps/
I guess anything in deps is currently compiled with whatever `CMAKE_C_STANDARD` and `CMAKE_C_EXTENSIONS` defaults to.

A side point here is that we should probably set `C_STANDARD` to `99` when building for release.
Otherwise `__STDC_VERSION__` wont be set to `199901` and we need to rely on rely on "hacky" `__GNUC__` checks to get flexible arrays and inline to work.
C90 is still nice for syntax checking.

### DIFF
--- a/src/transports/smart_protocol.c
+++ b/src/transports/smart_protocol.c
@@ -521,7 +521,6 @@ int git_smart__download_pack(
 	int error = 0;
 	struct network_packetsize_payload npp = {0};
 
-	// TODO
 	git_indexer_progress_cb progress_cb = t->connect_opts.callbacks.transfer_progress;
 	void *progress_payload = t->connect_opts.callbacks.payload;
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_library(git2internal OBJECT)
 set_target_properties(git2internal PROPERTIES C_STANDARD 90)
+set_target_properties(git2internal PROPERTIES C_EXTENSIONS OFF)
 
 
 if(DEPRECATE_HARD)


src/transports/smart_protocol.c
```diff
@@ -521,7 +521,7 @@ int git_smart__download_pack(
 	int error = 0;
 	struct network_packetsize_payload npp = {0};
 
-	// TODO
+	/* TODO */
```
boretrk, line 524: Hey @ethomson, this one feels a bit out of place. Todo what?

ethomson: Haha, sorry, my bad.  I usually mark things with a `// TODO` because that ridiculous comment style should jump out at me in a PR.  :trollface:  Usually I make sure that I address everything and make sure that all the TODOs are pulled out when I clean up my commit series before I send a PR.  Obviously I missed this one this time, oops.

The comment needs to go and the code surrounding it should be happy and not have anything that needs attention.

ethomson, line 524:
```suggestion
```

ethomson:
@boretrk you've got this marked as a draft still - do you want this in v1.4.0?

boretrk:
> @boretrk you've got this marked as a draft still - do you want this in v1.4.0?

Ah, right. I wanted to figure out what to do with the "release versions doesn't really get the nice C99 features like flexible arrays and inlined functions unless we specify C_STANDARD to 99" but that should probably go into another PR anyway.
Right now we rely on the compiler being GCC and adding that stuff anyway, but I'm thinking that C90 only should be used for syntax validation.

This one should be good to go, but isn't really necessary either since it only helps with tracking down those "//"-comments (And possibly some other stuff that we don't use)

ethomson:
Cool, I'm going to ship this and we can follow up with more!

PR #6187, opened 2022-01-23
COPYING: remove regex copyright, add PCRE copyright

ethomson:
We now bundle PCRE for our regular expression implementation; include
that copyright notice and remove the now-unused regular expression
copyright.


boretrk (approved):
LGTM

Issue #6189, opened 2022-01-27
merge.conflictStyle zdiff3 yields error


taladar:
### Reproduction steps

Use git 2.35 and set the new conflictstyle zdiff3

```
git config --global merge.conflictStyle zdiff3
```

### Expected behavior

Operations still work even if the zdiff3 conflict style might be substituted for some default value. Ideally it would be supported of course.

### Actual behavior

This happens when calling e.g. cargo-deny which uses git2-rs which uses libgit2

```
2022-01-26 10:53:03 [ERROR] failed to fetch advisory database: git operation failed: unknown style 'zdiff3' given for 'merge.conflictstyle'; class=Checkout (20)
```

### Version of libgit2 (release number or SHA1)

not quite sure but I assume the most recent release

I first reported this here on git2-rs

https://github.com/rust-lang/git2-rs/issues/795

### Operating system(s) tested

Gentoo Linux

ethomson:
This is certainly a worthwhile feature - but also, it feels like cargo-deny might want to take more control over the way in which it gets/writes/etc advisory databases.  It feels like this shouldn't be impacted by a user's git config settings.

taladar:
I am not the author, just a user.

> It feels like this shouldn't be impacted by a user's git config settings.

I agree. I assume it is read as part of a pull operation but I haven't looked into it in detail. For something like that (pure consumption of upstream git repository of data) what would probably be more appropriate would be a fetch + reset --hard

ethomson:
Fixed via #6195

PR #6195, opened 2022-01-30
merge: support zdiff3 conflict styles

ethomson:
Add support for zdiff3 using the xdiff from git v2.35.0.

Update to the xdiff used in git v2.35.0, with updates to our build configuration to ignore the sort of warnings that we normally care about (signed/unsigned mismatch, unused, etc.)

Any git-specific abstraction bits are now redefined for our use in `git-xdiff.h`.  It is a (wildly optimistic) hope that we can use that indirection layer to standardize on a shared xdiff implementation.


ethomson:
Fixes #6189

PR #6200, opened 2022-02-03
C90: add inline macro to xdiff and mbedtls

boretrk:
To be able to use unmodified xdiff code together with git-xdiff.h we need to make sure that `inline` exists with C90 (Or switch to C99)
This PR does the former.

deps/ntmlclient is left alone for now but it has the same problem.

We could consider doing something similar with the `GIT_INLINE(type)` macro but that would be a lot of changes for no particular benefit.

### DIFF
--- a/src/streams/mbedtls.c
+++ b/src/streams/mbedtls.c
@@ -23,12 +23,14 @@
 #endif
 
 /* Work around C90-conformance issues */
-#if defined(_MSC_VER)
-# define inline __inline
-#elif defined(__GNUC__)
-# define inline __inline__
-#else
-# define inline
+#if !defined(__STDC_VERSION__) || (__STDC_VERSION__ < 199901L)
+# if defined(_MSC_VER)
+#  define inline __inline
+# elif defined(__GNUC__)
+#  define inline __inline__
+# else
+#  define inline
+# endif
 #endif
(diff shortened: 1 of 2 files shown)


ethomson:
Haven't looked at this deeply yet but naively my thought is that we should never use the bare `inline` keyword, we should always use `GIT_INLINE`.  This definitely feels like a bug in `mbedtls.c` if it uses `inline` instead.

xdiff is a little trickier and obviously can't be `GIT_INLINE`.  I'm curious though if there any downsides to redefining `inline`?  🤔

boretrk:
>  This definitely feels like a bug in `mbedtls.c` if it uses `inline` instead.

Ah, as far as I can tell it doesn't actually use `inline` for itself.
It just defines `inline` for the `<mbedtls/*.h>` includes and undefines it directly after.

We could make it a bit more stable by placing it before the other includes, that way the `inline` trickery won't undefine any previous `inline` define, but then again, if that were to happen we would get an error on the redefine so keeping it where it is might give us an early warning if something odd is going on.

ethomson:
> It just defines inline for the <mbedtls/*.h> includes and undefines it directly after.

Oh, jeez.  Thanks, that makes sense.

Issue #5442, opened 2020-03-06
Is there a config for worktree?


Whirlwind:
```
❯ git config -h
usage: git config [<options>]

Config file location
    --global              use global config file
    --system              use system config file
    --local               use repository config file
    --worktree            use per-worktree config file
    -f, --file <file>     use given config file
    --blob <blob-id>      read config from given blob object
```
The git command line have a config file for the worktree mode, and have a `extensions.worktreeConfig` to enable it.

I found that the `git_repository_config` could not read the config file.

Now, I must get the `.git/worktrees/xxx/config.worktree` path, and add the config file into the `git_config`.

pks-t:
Unfortunately we do not have support for that yet. I'll hope to be able to tackle that after v1.0 is out.

Griffon26:
We are using google's git-repo and it also uses the worktreeconfig extension. Having support for that in libgit2 would hopefully allow us to use TortoiseGit with git-repo.

csware:
cf. issue #6044

PR #6202, opened 2022-02-06
Support the extension for worktree-specific config

### LINKED ISSUE #6044: [Feature] Support `extensions.worktreeconfig` for `repositoryformatversion=1`
```
$ git clone --filter=blob:none git@xxx:xx/xx.git
$ cd xx
$ git worktree add ../worktree
$ cd ../worktree
$ git config extensions.worktreeConfig true
```

I have a origin repository with `--filter=blob:none`, and a worktree repository. Then I enable the worktree config in the worktree.

In the libgit2, it could not support the worktree  config with `filter`:

`unsupported extension name extensions.worktreeconfig`

I check the libgit2:
```
static int check_valid_extension(const git_config_entry *entry, void *payload)
{
	GIT_UNUSED(payload);

	if (!strcmp(entry->name, "extensions.noop"))
		return 0;

	git_error_set(GIT_ERROR_REPOSITORY, "unsupported extension name %s", entry->name);
	return -1;
}

static int check_extensions(git_config *config, int version)
{
	if (version < 1)
		return 0;

	return git_config_foreach_match(config, "^extensions\\.", check_valid_extension, NULL);
}
```

And I check the version: `repositoryformatversion = 1`, So all extension configs will failed.

vermiculus:
Fix #6044.

Tasks:

- [x] Define a new level for worktree config
- [x] Read worktree config when loading repo config
- [x] Put `extensions.worktreeconfig` on the allow-list (don't bomb out a la #6044)
- [x] Unit tests

Open questions:

- How does this interact (if at all) with git.git:`enum config_scope`?
- Are there any places in the code that are assuming these values don't change?
- What other areas of the code need to be updated to adjust?
- ~[Does this break ABI?](https://github.com/libgit2/libgit2/blob/main/docs/api-stability.md)~ Yes, but evidently there's no way to avoid this.
- ~(meta) What subsystem even is this?  (for the commit subject prefix)~


src/repository.c
```diff
@@ -56,6 +56,7 @@ static const struct {
 	{ GIT_REPOSITORY_ITEM_COMMONDIR, GIT_REPOSITORY_ITEM_GITDIR, "packed-refs", false },
 	{ GIT_REPOSITORY_ITEM_COMMONDIR, GIT_REPOSITORY_ITEM_GITDIR, "remotes", true },
 	{ GIT_REPOSITORY_ITEM_COMMONDIR, GIT_REPOSITORY_ITEM_GITDIR, "config", false },
+	{ GIT_REPOSITORY_ITEM_COMMONDIR, GIT_REPOSITORY_ITEM_GITDIR, "config.worktree", false },
```
vermiculus, line 59: I am unsure about this definition (and don't yet know how to functionally test my changes outside of the unit test I introduced in the prior commit).

vermiculus:
I've realized while writing up a corresponding PR to rust-lang/git2-rs that inserting a value here could have a detrimental impact on compatibility with existing bindings.

Does this need to wait for a version break of some sort? Or do I need to come up with an alternate implementation of `git_config_get_*` to be able to use some sort of explicit 'this level overrides this other level' definition?

If we want to keep the existing int-comparison based overrides, we should probably take the opportunity to insert some wiggle room by making these values multiples of ten instead.

vermiculus:
Pinging on this. What's the next step here? Does this patch break ABI? Is there anything we can actually do about that?

ethomson:
Thanks for the ping @vermiculus - I'm happy to review this, I should be able to take some time today.

You're correct that it _does_ break ABI, and _no_, I don't think that there's anything that we can do about this.  If this is the correct ordering, then we'll give people a heads up in the next release but I'm afraid that there's little else that we can do here.

vermiculus:
Thanks for taking the time 😃

If there's no way to avoid breaking ABI, will it be just as much work for consumers if we change *all* the enum values to allow for future inserts?

```c
GIT_CONFIG_LEVEL_PROGRAMDATA = 10,
GIT_CONFIG_LEVEL_SYSTEM = 20,
// etc.
```

I can't see why or how we'd ever have another value (even at a conceptual level), but I imagine the same was said before `git-worktree` came along.

vermiculus:
Coming back around to this. What's the next step here?

Expecting I'll need to rebase to resolve the new conflicts, that I don't expect those conflicts to be terribly difficult to resolve.

ethomson:
Yep, sorry for the conflicts.  😢  If you rebase, I think that it's ready to go.

vermiculus:
No worries! Conflicts are a sign of active development 😉 

Rebased; looks like the conflicts were resolved automatically.

This is ready to roll from my end assuming you're good with https://github.com/libgit2/libgit2/pull/6202#discussion_r800196633.

vermiculus:
Pinging on this again 😄

bierbaum:
While testing this, I found the following bug: https://github.com/libgit2/libgit2/pull/6246

After that patch is merged (or integrated here) libgit2 is able to open repos with `extensions.worktreeconfig` set.

vermiculus:
Thanks for testing and finding that issue, @bierbaum 😄 I'll happily rebase once your PR is merged.

Otherwise I could rebase on it now and both can merge at the same time -- but it does feel like your fix can stand alone.

ethomson:
Sorry for the delay, @vermiculus. And thanks, again.

ethomson:
Hmm. Not sure what’s going on with the tests. It’s possible that this was a problem caused by #6252 but I’ll have to take a closer look tomorrow.

ethomson:
🤔 Doesn't look like it's related to #6252; these test failures look legitimate. I'm afraid that I will need to refamiliarize myself with the config code to offer any suggestions for how to proceed.

vermiculus:
a5a6eeef86c490d73d5e82bba29fef7cd1eb5f2c includes some logging output for investigation. It appears the search space is null for these two calls:

```
log: new call to git_config_entries_get_unique
log: needle: branch.local-track.remote
log: haystack:
log: end
log: 'branch.local-track.remote' NOT FOUND

log: new call to git_config_entries_get_unique
log: needle: branch.track-local.remote
log: haystack:
log:   branch.local-track.remote => .
log: end
log: 'branch.track-local.remote' NOT FOUND
```

vermiculus:
I'm investigating `config::rename::can_rename` as a test that's failing -- I think all the other failures are coming from the `git_config_delete_entry` in `rename_config_entries_cb`.

I think I've learned that at `git_config_delete_entry`, the `git_config` struct has all the expected 'entries' (in quotes because I'm using `git_config_foreach` to print everything and I'm not 100% sure what it does).

Once we get into `config_file_delete`, though, `backend->entries` does not have all the entries that were printed before. It's worth noting that `backend->level:=6` here -- 6 is the new 'worktree' value this patch introduced, whereas I'm willing to bet the value that needs to be deleted is on level 5.

Continuing investigation on that assumption.

vermiculus:
I believe the above patch is in the right direction (based on ~three hours' learning about config backends...) though some tests are still failing. I'll try to come back to this in a few hours.

src/libgit2/config.c
```diff
@@ -606,8 +607,19 @@ static int get_backend_for_use(git_config_backend **out,
...
 		if (!backend->backend->readonly) {
-			*out = backend->backend;
-			return 0;
+			switch (use) {
+			case BACKEND_USE_SET:
+				*out = backend->backend;
+				return 0;
+			case BACKEND_USE_DELETE:
+				/* return the first backend that contains the given config-name */
+				backend->backend->get(backend->backend, name, &entry);
```
vermiculus, line 616: Do I need to lock/unlock the backend here? Is there a risk of data getting written out from underneath me?

csware:
cross-ref: issue #6044 issue #5442

src/libgit2/config.c
```diff
@@ -606,8 +607,19 @@ static int get_backend_for_use(git_config_backend **out,
...
+				if (entry != NULL) {
+					*out = backend->backend;
+					return 0;
+				}
```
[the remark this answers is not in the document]
vermiculus: If fixing the pipeline is that simple, I'll give you a beer emoji. I'll try that right now.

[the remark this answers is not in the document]
vermiculus: I'm still going to try it (what's the worst that can happen?) but I don't think this will resolve the issue given that we weren't freeing `entry` before:

```diff
@@ -606,8 +607,19 @@ static int get_backend_for_use(git_config_backend **out,
 
 	git_vector_foreach(&cfg->backends, i, backend) {
 		if (!backend->backend->readonly) {
-			*out = backend->backend;
-			return 0;
+			switch (use) {
+			case BACKEND_USE_SET:
+				*out = backend->backend;
+				return 0;
+			case BACKEND_USE_DELETE:
+				/* return the first backend that contains the given config-name */
+				backend->backend->get(backend->backend, name, &entry);
+				if (entry != NULL) {
+					*out = backend->backend;
+					return 0;
+				}
+				break;
+			}
 		}
 	}
```

vermiculus:
Rebased and reworded the last commit. Turns out just the subject of 'return only backends that contain the key' was *not* helpful in figuring out why I wrote the patch a *year* ago. I'm pretty sure I was able to decipher the reason this was needed, though, based on https://github.com/libgit2/libgit2/pull/6202#issuecomment-1094037943.

vermiculus:
@ethomson looks like my approval to run workflows has lapsed; any chance you can renew it? Thanks!

vermiculus:
@ethomson Just saw that https://github.com/libgit2/libgit2/labels/v1.8 was added. Cool! As it happens, I'm also looking at the possibility of spending some $DAY_JOB hours on this in the near future. Any chance I could get permission to run the tests again?

vermiculus:
@ethomson We're hoping to look at this again next week; any chance I can get approval to run the workflow tests?

vermiculus:
@ethomson I believe this is ready to go – or at least ready for review. The only failures I see in my CI runs are of the form:

```
4: online::customcertFF
4: 
4:   1) Failure:
4: online::customcert::file [/home/libgit2/source/tests/libgit2/online/customcert.c:68]
4:   Function call failed: (git_clone(&g_repo, "https://test.libgit2.org:1443/anonymous/test.git", "./cloned", ((void *)0)))
4:   error -1 - failed to resolve address for test.libgit2.org: Temporary failure in name resolution
4: 
4:   2) Failure:
4: online::customcert::path [/home/libgit2/source/tests/libgit2/online/customcert.c:76]
4:   Function call failed: (git_clone(&g_repo, "https://test.libgit2.org:2443/anonymous/test.git", "./cloned", ((void *)0)))
4:   error -1 - failed to resolve address for test.libgit2.org: Temporary failure in name resolution
```

and this may even be expected in my somewhat-hacked-together setup (force-pushing to `vermiculus:main` to get CI to run).

Let me know if there's any way I can help expedite review.

Thanks!

vermiculus:
@zivarah brought up during internal code review that this PR could end at daeef1552b4fe93454ded5cb8acda84992d89fa0 and a separate PR be opened for ..daeef1552b4fe93454ded5cb8acda84992d89fa0. I'm happy to do so, but I thought I'd ask what your preference would be as a maintainer. Let me know.

ethomson:
> @zivarah brought up during internal code review that this PR could end at https://github.com/libgit2/libgit2/commit/f4b11136eb348e1473d4922a61ff4a1331f79a93 and a separate PR be opened for ..https://github.com/libgit2/libgit2/commit/daeef1552b4fe93454ded5cb8acda84992d89fa0. I'm happy to do so, but I thought I'd ask what your preference would be as a maintainer. Let me know.

I'm happy with two things intermingled in a single PR, no need to split it out. I'm sorry that I haven't had time to 👀 this yet, but I'm hopeful to do it this weekend. 🙏

csware:
fixes #6650

src/repository.c
vermiculus: Resolving this after the introduction of other unit tests and verifying behavior with the debugger.

src/libgit2/config.c
```diff
@@ -593,36 +593,67 @@ static int get_backend_for_use(git_config_backend **out,
...
+		   single config level with git_config_open_level and
+		   provide that as the config. In this case, there
+		   will only be one backend in the config. */
+		if (len > 1 && backend->level == GIT_CONFIG_LEVEL_WORKTREE)
+			continue;
+
+		/* If we're trying to delete a piece of config, make
+		   sure the backend we return actually defines it in
+		   the first place. */
+		if (use == BACKEND_USE_DELETE) {
```
ethomson, line 628: I need to go through and refresh my understanding of our configuration logic. This change _makes sense_ if the perspective of libgit2 is _when we say delete a configuration key, then just delete it wherever it is_. (And I think that's the semantics that our current config data sort of implies.)

This is different than `git config`, which _takes_ a config entry to operate on.

But I think that this change actually suggests a deeper problem with our API. IIUC, the goal of this change is to go find the config that we should delete a key in. What should we do if there's a multivar split over two configuration files? Instead `get_backend_for_use`, should we iterate over all backends?

(Not asking you to make this change, asking for your opinion on the API.)

src/libgit2/config.c
```diff
@@ -593,36 +593,67 @@ static int get_backend_for_use(git_config_backend **out,
...
 	}
 
 	git_vector_foreach(&cfg->backends, i, backend) {
-		if (!backend->backend->readonly) {
-			*out = backend->backend;
-			return 0;
+		if (backend->backend->readonly)
+			continue;
+
+		/* git-config doesn't update worktree-level config
```
ethomson, line 616: This comment is interesting — isn't this true of all `git config` entry levels? IOW, doesn't `git config` always operate on the local configuration unless otherwise specified?

I actually think that this exposes an oddity in our whole ordering. Our configuration levels appear to be trying to support a scenario where we have a system of ordering so that higher priority backends can override the data in lower priority backends. (eg, if I have `foo.bar` in system, and `foo.bar` in local, then the `foo.bar` in local is used.)

But we also want a system that's relatively flexible so that library users can plug in their own backends. (In other words, we _don't_ want to just always default to the local backend, which is what we _would_ do in the absence of any pluggability.)

The wrinkle here is that you're introducing a new backend that should take priority over reads but should _not_ take priority over writes. I think that we should consider doing two things:

1. For the sort of generic, unopinionated configuration system, we should have two priority ordered lists — one for read priority, and one for write priority
2. When a repository is created by default, it should set the read priority in the order that we've specified here. The write priority should only have a single option, which is the local configuration. (By default, calling set without a specific entry level should probably just always go there.)

This lets users have control over the ordering, but sets a default that aligns with git.

Does this make sense? Am I entirely off base? Am I overthinking things?

src/libgit2/config.c
vermiculus: > What should we do if there's a multivar split over two configuration files? 

Certainly this case would surprise folks; I would not expect to get different results with repeated calls:

```c
git_config_delete_entry(cfg, "foo.bar"); // deletes --local
git_config_delete_entry(cfg, "foo.bar"); // deletes --global
```

That just feels 'wrong'. It would feel much more appropriate to say

```c
git_config_delete_entry_at_level(cfg, GIT_CONFIG_LEVEL_LOCAL, "foo.bar"); // deletes --local
git_config_delete_entry_at_level(cfg, GIT_CONFIG_LEVEL_GLOBAL, "foo.bar"); // deletes --global
// with companion functions
git_config_set_string_at_level(cfg, GIT_CONFIG_LEVEL_LOCAL, "foo.bar", "baz");
// etc.
```

It was surprising to us to learn that the current config API did not work this way. In a vacuum, I would consider `git_config_set_string_at_level` to be the correct abstraction for a C library: explicit and unambiguous even if a little verbose.

I feel comfortable enough with the config code at this point to tackle such a thing in a separate PR if that's a direction you'd like to go. (In fact, I've rather been enjoying this codebase – it's a nice change of pace for me 😄.) It probably deserves a more robust design process, though.

It would also make sense to me to incorporate the worktree-config support *only* into that new API if that helps resolve concerns with this PR, but that will come with different considerations for doing due diligence to ensure consumer applications continue to 'work reasonably correctly'. (Although maybe that's not as high of a concern in this context as long as we make the appropriate deprecation warnings; I've not worked with this particular kind of project in a FOSS setting before.)

src/libgit2/config.c
vermiculus: I don't think you're over-thinking things. I agree that the different behavior for reading/writing is not something that's nicely explicit in the current API.

This feels very related to https://github.com/libgit2/libgit2/pull/6202#discussion_r1501780259. I will have to noodle on these points more (I've got some household chores to do in the meantime), but I suspect my other reply might be a start toward a resolution here.

src/libgit2/config.c
ethomson: I'm not positive but my thinking here is that maybe we should keep things mostly as they are but, yes, avoid surprise:

I think that the "simple" `git_config` functions should probably mirror the git command line itself. Reads return the "effective" configuration (applying the config files in order of priority and giving a result.) But we change writes/deleted to always refer to the local.

If somebody wants a specific level then they can open that file and operate on it directly. I don't think we necessarily need APIs to read and write at a specific level directly, I think that people can probably go get the config file that they're interested in.

But this rather presupposes that git's model is right here — and I'm increasingly convinced that it actually is.

The rather tricky part is the *user* configuration(s). We may need to be more thoughtful than "always write to local".

vermiculus: I was reviewing this comment once more before writing up what this would look like

> I think that the "simple" git_config functions should probably mirror the git command line itself. Reads return the "effective" configuration (applying the config files in order of priority and giving a result.) But we change writes/deleted to always refer to the local.

I think that's a good instinct as it would follow some principle of least surprise.

What I'm not sure of is what is meant by this:

> The rather tricky part is the _user_ configuration(s). We may need to be more thoughtful than "always write to local".

By 'user configurations', do you mean things like `--global` and `--system` config?

Here's what I'm going to propose:

> For reads: keep the current implementation and use the most specific level available to us.
> 
> For writes:
> 1. If there is exactly one config backend, use it.
> 2. If `--local` is available, use it.
> 3. Signal an error (as `git` does).

The alternative to (3) (which might be the 'more thoughtful' approach, as best as I can imagine one) would be to set the most specific level of config available to us if `--local` is not available, but I rather think that might be too smart for this 'plumbing' function. I do think the need of 'set this at the system/global level' would be met with `git_config_open_level` followed by `git_config_set_string` following the logic proposed above. (Some wrapper 'porcelain' could be devised as above, but it does seem like that would be low-value API.)

vermiculus: I have pushed c82bef33a090217e731944e91385e1f3095c1fee which takes the approach outlined above. Before the final version, I would like to add a test that `git_config_open_level(&out, cfg, GIT_CONFIG_LEVEL_LOCAL)` only ever returns a config with one backend (even in degenerate cases) since I'm now relying on that behavior in the base case of `get_backend_for_use`.

Note that tests introduced in #4217 are non-trivially altered (@pks-t), but I believe the test is still effective in demonstrating that the `readonly` flag is respected.

There is another problem `test_remote_httpproxy__env` that I've not tracked down yet; some details are in the commit though.

ethomson: I spent a little time staring at this over the weekend and there's much more complexity here than I expected. You may have identified some of these problems as well. In particular, backends are shared across multiple `git_config` instances. (The repository has a configuration with multiple backends with an ordering. When you create a snapshot, you get a new configuration with read only backends. And you can `open_level`, you get a new configuration object with only one of those backends.)

The other "gotcha" is that we support users of the library to create their own custom configuration, `GIT_CONFIG_LEVEL_APP` (or higher) which is higher-priority than the system/global/etc configuration. And we have no idea what they're doing. So trying to be clever about where writes go needs to take into account the `APP` level which may or may not exist and may or may not be writeable.

The current flow today is "write to the highest level backend that isn't readonly", which _may_ be `APP` and if it's not it _will_ be `LOCAL`. 

So, because of this shared state, you can't simply make the worktree backend "not writable" for the repository configuration, since it will also be not writable when you `open_level` it, which you _do_ want to write to. 🙃 

And trying to hardcode any logic about defaulting to the local configuration is tricky, since there might be `APP`-levels that you need to contend with (and that are being written to today and we would like to preserve backcompat).

My initial thought was that we could re-use the `readonly` bit (or possibly make a new "writes need to be explicit" bit) and make the worktree config readonly when it's created via `git_repository_open`, but that doesn't work, because subsequent `open_level` will happily just give you the existing worktree config which is... readonly.

Instead, I'm coming back to the notion that configuration needs separate read and write ordering. Providing a `set_writeorder` API allows callers (in particular, `git_repository`) to set that they want `LOCAL` to be the only write target during configuration creation. Users can subsequently add an `APP` level, which will be be the highest-level writer (but they can change that themselves with `set_writeorder`).

Ultimately, `set_writeorder` only applies to the things that are in your config object. So if you want to override the default write order of the thing(s) you just added, you will need to call it again after adding them. (You can't reconfigure your write order.)

This required us to pull apart the refcounted backend object from the list of backends that are in a config so that they can be separate. Alas. In any case, I pushed up a branch [`ethomson/worktree-config`](https://github.com/libgit2/libgit2/tree/ethomson/worktree-config) ([diff](https://github.com/libgit2/libgit2/compare/main...ethomson/worktree-config)). I'd love your feedback.

I also found and fixed some interesting bugs, notably that you can't `git_transaction_free` an uncommitted transaction today. And that if you start a transaction and then mess around with backends, you're going to have a bad time.

ethomson: I think that we were typing our comments at roughly the same time. 😅 

Right, so, my concern here is definitely around the `CONFIG_LEVEL_APP` no longer being a write target. I think that's critical to backcompat... Our CLI uses it to map `-c` command-line arguments into a configuration backend (though this is read-only). I know that at least one hosting provider uses `CONFIG_LEVEL_APP`, and ... I have no idea if they're using it to write config data or only read?

I'm loathe to break it, though. 😬

ethomson:
Fixed via #6756

Issue #5905, opened 2021-06-02
Broken pipe error while pushing over HTTPS


pluehne:
When pushing to repositories using the HTTPS protocol, we sporadically observed an issue where `git_remote_push` fails with a `SIGPIPE`. When using the Rust wrapper `git2-rs`, the error message reads as follows:

>code: -1
>klass: 2
>message: libgit2 error (2): SSL error: syscall failure: Broken pipe

While debugging this issue, we noticed that it can be reliably reproduced when pushing the `master` branch of https://github.com/algolia/docsearch-configs to an empty repository over HTTPS.

We tried finding the root of the issue, but we were unable to understand what exactly is causing the broken pipe error so far. To us, it looks like a timeout occurs directly before or while sending data to an SSL socket such that the connection is terminated prematurely, resulting in an error the next time data is sent. The reason for that is unclear to us. We also saw that the push operation consistently fails trying to push a specific object (number 17, size: 116732 bytes).

It would be nice if someone more knowledgeable with the networking implementation could take a look :slightly_smiling_face:.

### Reproduction steps

Make sure that you have the updated `push` example with the changes from #5904 on your machine. (Our patch adds support for authentication to the `push` example, which is helpful for testing—most Git servers don’t provide unauthenticated write access after all. We considered this a simpler way to reproduce the issue than building and debugging a separate code example.)

1. Create a new repository you have write access to on a server such as GitHub.com.

2. Clone https://github.com/algolia/docsearch-configs.

   ```shell
   $ cd /tmp
   $ git clone https://github.com/algolia/docsearch-configs
   $ cd docsearch-config
   ```

3. Set the `origin` remote to the repository you created in step 1. Make sure to use an HTTPS URL.

   ```shell
   $ git remote set-url origin https://github.com/<account>/<repository>
   ```

4. Run the `push` example from your working copy of `libgit2`. This will push the previously obtained `master` branch to the `origin` remote, which was updated to point to the target repository in the previous step.

   ```shell
   $ cd <path to libgit2 build directory>
   $ ./examples/lg2 --git-dir /tmp/docsearch-config push
   ```

   Provide your credentials as needed. On GitHub.com, it is advisable to use a [personal access token](https://docs.github.com/en/github/authenticating-to-github/keeping-your-account-and-data-secure/creating-a-personal-access-token) for that purpose (`token` can be used as a user name in that case).

### Expected behavior

[...]

PR #6205, opened 2022-02-07
push: Prepare pack before sending pack header.

ccstolley:
For large pushes, preparing the pack can take a while. Currently we send the pack header first, followed by preparing the pack and then finally sending the pack. Unfortunately github.com will terminate a git-receive-pack command over http if it is idle for more than 10 seconds. This is easily exceeded for a large push, and so the push is rejected with a Broken Pipe error.

This patch moves the pack preparation ahead of sending the pack header, so that the timeout is avoided.

`prepare_pack()` can be called multiple times but will only do the work once, so the original `PREPARE_PACK` call inside `git_packbuilder_foreach()` remains.

Resolves https://github.com/libgit2/libgit2/issues/5905 .

### DIFF
--- a/src/pack-objects.h
+++ b/src/pack-objects.h
@@ -97,5 +97,7 @@ struct git_packbuilder {
 };
 
 int git_packbuilder__write_buf(git_str *buf, git_packbuilder *pb);
+int prepare_pack(git_packbuilder *pb);
+
 
 #endif
--- a/src/pack-objects.h
+++ b/src/pack-objects.h
@@ -97,7 +97,7 @@ struct git_packbuilder {
 };
 
 int git_packbuilder__write_buf(git_str *buf, git_packbuilder *pb);
-int prepare_pack(git_packbuilder *pb);
+int git_packbuilder__prepare(git_packbuilder *pb);
 
 
 #endif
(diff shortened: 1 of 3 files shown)


ccstolley:
> it shows that there is a deficiency in the public API of the packbuilder

Are there specific use cases you had in mind for such an API? I'm reluctant to introduce a new API without a concrete justification for it. Perhaps I'm thinking too narrowly, but to me preparing the pack is a lower level detail that should be abstracted away by the library. And this bug illustrates that getting that timing right can be subtle.

src/pack-objects.h
```diff
@@ -97,5 +97,7 @@ struct git_packbuilder {
 };
 
 int git_packbuilder__write_buf(git_str *buf, git_packbuilder *pb);
+int prepare_pack(git_packbuilder *pb);
```
ethomson, line 100: I don't think that we need to expose this publicly (yet) - I agree that exposing APIs that aren't "needed" gives us a maintenance burden moving forward.  But let's give it a nicer name even if it's an internal API.

```suggestion
int git_packbuilder__prepare(git_packbuilder *pb);
```

ethomson:
LGTM.  I don't think that we need to make it public yet, necessarily - but I do think that we should expect that we might want to in the future if we have more data.  In the meantime, let's tighten up the internal name a bit and 🚢

ethomson:
Thanks @ccstolley - looking good!  😎

PR #6207, opened 2022-02-08
mktmp: improve our temp file creation

ethomson:
**Hey, why are our CI builds taking four hours?**

I was hasty about merging #6178 and ignored @boretrk's [sage advice](https://github.com/libgit2/libgit2/pull/6178#discussion_r785324851) about adding randomness to `mktemp`.  It turns out that mingw64's `mktemp` is... just terrible.  It's a single alphabetic character (`a`-`z`) followed by the process id.  Which meant that in our test where we [add a bunch of data to the repository](https://github.com/libgit2/libgit2/blob/37763d38fbf5db932be8d6a0070255839e7719ca/tests/threads/diff.c#L170) in different threads caused temp file races against each other.  At some point when 26 threads had temp files, we would end up in a [big loop](https://github.com/libgit2/libgit2/blob/4ac5972811a711f9b26e34fd6b6d6df7f5407953/src/futils.c#L31) retrying new filenames, but having exhausted them, we would just sit in that loop for a while until one of the threads that had a temp file would eventually finish.

(Or we deadlocked, and never ended up writing anything at all, and the [lack of error checking](https://github.com/libgit2/libgit2/blob/37763d38fbf5db932be8d6a0070255839e7719ca/tests/threads/diff.c#L170) simply means that the failures were ignored.)

**Okay, so what are we going to do about it?**

This PR gets rid of our reliance on `mktemp` and `mkstemp`, a) because they might suck, and b) it's not [what we want anyway](https://github.com/jmgao/pore/issues/46).  We're not writing temp files into writable directories (`/tmp`) and need something that gives us a mode `0600` file.  Instead, we're writing temp files into repositories and need something that respects `umask`.

Regrettably, this is not _entirely_ trivial.  We do not want to use `rand`, because we cannot guarantee that the process calling us has ever called `srand`.  And we cannot _call_ `srand` ourselves, because the process calling us _may have called srand_ and depends on a particular sequence number.  (Test harnesses and fuzzers will often `srand` with a random seed, and report that seed so that if there's a failure, it's reproducible by using the same seed in subsequent runs).

So this PR:

1. Introduces the [xoroshiro256**](https://prng.di.unimi.it) pseudo-random number generator.  xoroshiro256** is fast and simple, and has the uncommon attribute of being easy to reason about code (64 bit types are `uint64_t`) instead of very academic (the code in my paper is a `long int`, what do you mean portable?).  This is `git_rand_next`, and I've run the output against PractRand to ensure that there were no copy-pasta errors.
2. Introduces `git_rand_seed` which will seed the PRNG.  We do this by pulling 64 bits out of the system's entropy pool (`CryptGenRandom` on Windows, or `getentropy` on *BSD and Linux).  If we're on a non-Windows system and `getentropy` is not available (or fails), we'll fall back to trying to read from `/dev/urandom`.  If none of those work, we'll use the system time (hopefully with a resolution better than a second) and xor with some system values like load average, uptime, process ID, etc.  This strategy was [largely borrowed from libressl](https://github.com/libressl-portable/openbsd/blob/4e9048830a68da79247f30aba182b1599da139b9/src/lib/libcrypto/crypto/getentropy_linux.c#L323), but not as aggressive, since we're putting a temp file in a directory that we control, not sending your credit card numbers across the internet.
3. Update our `mktemp` function to use our new PRNG and use it consistently.

Notes:
1. I'm not a cryptanalyst, but I feel confident about xoroshiro256** based on its adoption by Java.
2. Even if I didn't, we don't need to worry too much about attacks here, since we're only ever writing within the repository and our goal is race/collision-avoidance.  That said, we should probably not use this in `/tmp` or world-writable directories.
3. We put a mutex around our state boxes for thread safety.  It's probably fine for our needs.  We could, I guess, try to pull it into a thread-local, but then we'd need to re-seed each thread independently _and_ try to ensure that they were separate seeds.  This is easy in the `getentropy()` case, hard in the `time(NULL) ^ getpid()` case.  It's probably fine for our needs.

/cc @boretrk


src/rand.c
```diff
@@ -0,0 +1,231 @@
...
+	*seed |= ((uint64_t)tv.tv_usec << 40);
+	*seed |= ((uint64_t)tv.tv_sec);
+
+	*seed ^= ((uint64_t)getpid() << 32);
+	*seed ^= ((uint64_t)getpgid(0));
+	*seed ^= ((uint64_t)getppid() << 32);
+	*seed ^= ((uint64_t)getsid(0));
+	*seed ^= ((uint64_t)getuid() << 32);
+	*seed ^= ((uint64_t)getgid());
+
```
boretrk, line 124: I think there is a missing call to getloadavg(loadavg,3) here, unfortunately it isn't a POSIX function so we might need to check if it exists.

src/rand.c
```diff
@@ -0,0 +1,231 @@
...
+	*seed ^= ((uint64_t)loadavg[1]);
+	*seed ^= ((uint64_t)loadavg[2]);
+
+	/* Mix in the addresses of some functions and variables */
+	*seed ^= (((uint64_t)((uintptr_t)getseed)) << SEED_SHIFT);
+	*seed ^= (((uint64_t)((uintptr_t)seed)));
+	*seed ^= (((uint64_t)((uintptr_t)printf)) << SEED_SHIFT);
+	*seed ^= (((uint64_t)((uintptr_t)errno)));
+
+	*seed ^= ((uint64_t)git__timer());
```
boretrk, line 135: Given that this one returns a double I'm thinking that we could possibly get some more bits out of it either by multiplying with some constant to get some of the sub-second information into the seed or write it to a `union {double, uint64_t}` to get access to all the bits.

ethomson: 🤔  Isn't `double` 64-bit on all data models?

boretrk: The idea with using a union would be to write the double and read it as an integer to get all 64 bits into the seed.

Technically double can be other sizes if `__STDC_IEC_559__` isn't defined but I haven't seen this outside of some 8-bit processors where double is equivalent to float. I don't think that is something we need to take into consideration.

src/rand.c
```diff
@@ -0,0 +1,231 @@
...
+
+	*seed ^= ((uint64_t)loadavg[0]);
+	*seed ^= ((uint64_t)loadavg[1]);
+	*seed ^= ((uint64_t)loadavg[2]);
+
+	/* Mix in the addresses of some functions and variables */
+	*seed ^= (((uint64_t)((uintptr_t)getseed)) << SEED_SHIFT);
+	*seed ^= (((uint64_t)((uintptr_t)seed)));
+	*seed ^= (((uint64_t)((uintptr_t)printf)) << SEED_SHIFT);
+	*seed ^= (((uint64_t)((uintptr_t)errno)));
```
boretrk, line 133: I'm thinking that `getseed` and `printf` are the same for all threads here and probably doesn't add much. (And because of some microcontrollers with separate address spaces for functions and data C99 doesn't allow casting function pointers to data pointers.)

Are `seed` and `errno` at unique addresses for each thread?
Otherwise `&tv` should give a pointer to a stack variable.

src/rand.c
ethomson: Oh, yes, I see what you're saying.  Apologies; I never work with floats and so I was just thinking about it as a big hunk of bytes.  Indeed a simple cast here will drop a lot of the data we wanted in this first place.

src/rand.c
```diff
@@ -0,0 +1,231 @@
...
+	*seed ^= ((uint64_t)getpid() << 32);
+	*seed ^= ((uint64_t)getpgid(0));
+	*seed ^= ((uint64_t)getppid() << 32);
+	*seed ^= ((uint64_t)getsid(0));
+	*seed ^= ((uint64_t)getuid() << 32);
+	*seed ^= ((uint64_t)getgid());
+
+	*seed ^= ((uint64_t)loadavg[0]);
+	*seed ^= ((uint64_t)loadavg[1]);
+	*seed ^= ((uint64_t)loadavg[2]);
```
boretrk, line 127: Quick test gave me 0.18, 0.06 and 0.01 for the avg values on an unloaded system so just converting all loadavg values to nearest integer might not give a lot of randomness.
```
*seed ^= *(uint64_t *)&loadavg[0];
*seed ^= *(uint64_t *)&loadavg[1];
*seed ^= *(uint64_t *)&loadavg[2];
```
could give a bit more randomness, but with xor the values will probably mostly cancel each other out.
I think just mixing in loadavg[0] and skipping the others will give a just as good value as all three of them together.
Adding them before xoring them in might be a better alternative.

src/rand.c
```diff
@@ -0,0 +1,231 @@
...
+	*seed = 0;
+	*seed |= ((uint64_t)tv.tv_usec << 40);
+	*seed |= ((uint64_t)tv.tv_sec);
+
+	*seed ^= ((uint64_t)getpid() << 32);
+	*seed ^= ((uint64_t)getpgid(0));
+	*seed ^= ((uint64_t)getppid() << 32);
+	*seed ^= ((uint64_t)getsid(0));
+	*seed ^= ((uint64_t)getuid() << 32);
+	*seed ^= ((uint64_t)getgid());
```
boretrk, line 123: I'm thinking that xor have a tendency to cancel identical values out.
Maybe switching `^=` to `+=` or mixing them up could help a bit?

src/rand.c
```diff
@@ -0,0 +1,231 @@
...
+	*seed ^= ((uint64_t)GetCurrentProcessId() << 32);
+	*seed ^= ((uint64_t)GetCurrentThreadId());
+
+	GetSystemTimes(&idletime, &kerneltime, &usertime);
+	*seed |= ((uint64_t)idletime.dwLowDateTime);
+	*seed |= ((uint64_t)idletime.dwHighDateTime << 32);
+	*seed |= ((uint64_t)kerneltime.dwLowDateTime << 32);
+	*seed |= ((uint64_t)kerneltime.dwHighDateTime);
+	*seed |= ((uint64_t)usertime.dwLowDateTime);
+	*seed |= ((uint64_t)usertime.dwHighDateTime << 32);
```
boretrk, line 66: Is this supposed to be `|=` for these 6?
It "feels" like they were supposed to be `^=`

src/rand.c
ethomson: I wasn't aiming for uniqueness across threads - we'll seed once for the library at `git_libgit2_init` time for all threads - I was aiming for uniqueness across process invocations to avoid a replay attack.  But for `pthread`, that will be a noop on anything that doesn't do ASLR, which I actually assumed was more common than it is - my Mac doesn't do ASLR, at least not by default.  🤔

And that should have been `&errno`, not `errno`.  Meh.

In any case, I think that I got a bit belt-and-suspenders here; certainly I can remove the function pointers here for broader compatibility, especially since it's not the macOS and Linux cases that are going to fall into this function.

src/rand.c
ethomson: Meh, I actually meant to compose the high and low bits into a single 64 bit number and then xor _that_.

src/rand.c
boretrk: Sorry, I got a bit of a tunnel vision here. The cases that doesn't have either `getentropy` or `/dev/urandom` probably lacks most other security functions.

boretrk:
LGTM

An alternative to the shift/xor deal for the seed generation could have been to just feed the values into one of the SHA algorithms and let it deal with shuffling the bits around, but that is probably overkill.

src/rand.c
ethomson: I do, too, at this point.  This feels like a lot of code for "temp file in a place that we own", but there are always surprises.

ethomson:
Shipping it to unblock our CI

PR #6208, opened 2022-02-08
diff_file: fix crash if size of diffed file changes in workdir

jorio:
diff_file_content_load_workdir_file() maps a file from the workdir into memory. It uses git_diff_file.size to determine the size of the memory mapping.

If this value goes stale, the mmaped area would be sized incorrectly. This could occur if an external program changes the contents of the file after libgit2 had cached its size.

This used to segfault if the file becomes smaller (mmaped area too large). This patch causes diff_file_content_load_workdir_file to fail without crashing if it detects that the file size has changed.


ethomson:
Ouch, this is a problem, indeed!  Thanks for the PR, I'll 👀 it soon.

src/diff_file.c
```diff
@@ -328,16 +328,25 @@ static int diff_file_content_load_workdir_file(
...
-	if (error < 0 || !fc->file->size)
+	if (error < 0 || !new_file_size)
 		goto cleanup;
 
+	/* if file size doesn't match cached value, abort */
+	if (fc->file->size && fc->file->size != new_file_size)
+	{
+		error = -1;
+		goto cleanup;
+	}
```
ethomson, line 346:
```suggestion
	if (fc->file->size && fc->file->size != new_file_size) {
		git_error_set(GIT_ERROR_FILESYSTEM, "file changed before we could read it");
		error = -1;
		goto cleanup;
	}
```

src/diff_file.c
```diff
@@ -328,16 +328,25 @@ static int diff_file_content_load_workdir_file(
...
 
 	if (fd < 0)
 		return fd;
 
-	if (!fc->file->size)
-	    error = git_futils_filesize(&fc->file->size, fd);
+	error = git_futils_filesize(&new_file_size, fd);
 
-	if (error < 0 || !fc->file->size)
+	if (error < 0 || !new_file_size)
```
ethomson, line 338: I don't think that I understand this test...  it seems like we could have truncated the file to zero bytes and do not treat that as an error.  Should we not always treat a non-zero `fc->file->size` and a differing `fc->file->size` to `new_file_size` as an error?  I do think it makes sense to skip `mmap`ing the zero byte case, but I think that we need to do this below the test beneath it...

```suggestion
	if (error < 0 )
```

src/diff_file.c
jorio: I agree we should skip this test. I don't think the original test made sense here either (testing fc->file->size if git_futils_filesize is successful).
Plus, removing this test ensures that we correctly handle a file becoming empty. (I'll add another test case for this soon)

src/diff_file.c
jorio: I agree we should skip this test. I don't think the original test made sense here either (testing fc->file->size if git_futils_filesize is successful).
Plus, removing this test ensures that we correctly handle a file becoming empty. (I'll add another test case for this soon)

jorio:
Thank you for your review!

I just realized that a minor issue remains - `diff_file_content_load_workdir_file` assumes that the size of the file is unknown if `fc->file->size == 0`, but this is ambiguous if the file is actually empty.

I'd expect diff_file_content_load_workdir_file to fail in the test case below (which I didn't add to the PR) when an empty file becomes non-empty.

We call mmap with the correct size in this case, so it's not a huge deal, but I still believe this should fail for consistency. Perhaps I'll look into this in a future PR.

```C
void test_diff_externalmodifications__empty_file_becomes_non_empty(void)
{
	git_index *index;
	git_diff *diff;
	git_patch* patch;
	git_str path = GIT_STR_INIT;

	cl_git_pass(git_str_joinpath(&path, git_repository_workdir(g_repo), "README"));

	/* Empty out the file */
	cl_git_mkfile(path.ptr, "");

	/* Get a diff */
	cl_git_pass(git_repository_index(&index, g_repo));
	cl_git_pass(git_diff_index_to_workdir(&diff, g_repo, index, NULL));
	cl_assert_equal_i(1, git_diff_num_deltas(diff));
	cl_assert_equal_i(0, git_diff_get_delta(diff, 0)->new_file.size);

	/* Simulate file modification after we've gotten the diff */
	cl_git_mkfile(path.ptr, "hello");

	/* Attempt to get a patch <--------- This passes, but it should arguably fail */
	cl_git_fail(git_patch_from_diff(&patch, diff, 0));

	git_index_free(index);
	git_diff_free(diff);
	git_str_dispose(&path);
}
```

ethomson:
I pushed up a change so that diff will flag when it knows the object size; if it does we now enforce that is doesn’t change.  This tried to make a 0 size file unambiguous.

I made it a positive (the size is known) both to match the existing ID flag and because I think - but I’m not certain - that the only places that size gets set are during diff generation from iterators.  If I missed a place then we’ll merely have a size but not assert it rather than the other way around.

Thanks much for this PR!  It’s a definite improvement and I really appreciate you taking the time to create a unit test that reproduces.  Huge thanks.

ethomson:
(Sorry for taking over your PR, by the way, I was hoping to get v1.4.0 this weekend if time permits and wanted to ensure that this was included.)

PR #6209, opened 2022-02-09
Two additions to the MemPack ODB backend

olsen232:
- group_by_commit flag - make the group-by-commit behaviour that the MemPack
backend already exhibits able to be turned on or off (when off, MemPack just
dumps everything you write to it, regardless of its object type).

- dump_to_pack_dir - dump the MemPack to disk in the default location, as opposed
to dumping it to a buffer in memory that still needs to be written somewhere.

With these two changes, it is easier for a libgit2 client to add a MemPack backend,
make any arbitrary writes to the ODB, and then dump those writes directly to a packfile,
skipping the loose objects step entirely. The use-case for needing packfiles as opposed
to loose objects is not common, but that is what MemPack does already, and this makes
MemPack easier to use and more full-featured.

Backwards compatible - leaves existing functionality unchanged.


ethomson:
Hi @olsen232!  This seems reasonable, but I'd like to take a deeper look soon 🔜 👀, as soon as we push v1.4.0 out.  Exciting stuff!

PR #6212, opened 2022-02-13
Fix crashes in example programs on Windows (sprintf_s not compatible with snprintf)

apnadkarni:
When building with Visual C++, the example programs crash in some instances because they expect `snprintf` to allow NULL buffers to be passed with a length argument of 0 to retrieve the required length of output buffer (as in C99 spec). However, `common.h` defines `snprintf` as `sprintf_s` for MS C and MingW and the latter does not accept NULL buffers raising an assert.

Fix is simply to define `snprintf` as `_snprintf` for instead of `sprintf_s`.

### DIFF
--- a/examples/common.h
+++ b/examples/common.h
@@ -45,7 +45,7 @@
 #endif
 
 #ifdef _MSC_VER
-#define snprintf sprintf_s
+#define snprintf _snprintf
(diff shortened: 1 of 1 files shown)


ethomson:
Makes sense.  Thanks!

PR #6219, opened 2022-02-14
Free parent and ref in lg2_commit before returning.

apnadkarni:


ethomson:
Looks good - thanks @apnadkarni!

PR #6222, opened 2022-02-18
Cli benchmark

ethomson:


ethomson:
Opened for testing (oops).

PR #6226, opened 2022-02-21
remote: do store the update_tips callback error value

carlosmn:
We use `git_error_set_after_callback_function` to determine whether
`update_tips` returned an error but do not store its return value making us
think it always returns 0.

Fix it by adding the common patter of storing it inside the `if` when calling it.

---

I have verified this fixes the rugged tests when updating to v1.4.1 and am somewhat surprised that the libgit2 tests did not catch this. It looks like this broke in the refactoring in b523776785e542792bd22f185ed0a678ab07c872 which is part of #6203 as the previous code would hard-code a `goto` to an error exit rather than use any of these fancy functions.

### DIFF
--- a/src/remote.c
+++ b/src/remote.c
@@ -1852,7 +1852,7 @@ static int update_one_tip(
 	}
 
 	if (callbacks && callbacks->update_tips != NULL &&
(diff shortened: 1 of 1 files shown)


ethomson:
Oops!  Thanks @carlosmn

PR #6227, opened 2022-02-21
odb: initialize `object` before use

ethomson:
Newer gcc is complaining about `object` being potentially not
initialized; initialize it.


ethomson:
It looks like a newer gcc picks up more than just this one - closing this until we can take a more holistic pass.

PR #6233, opened 2022-02-25
[cmake] allow C_STANDARD to be set from commandline

257:
without this, cross-compiling for Android with `ExternalProject_Add()`,
where targets are not available to the *SuperProject* for setting
propeties on them, build fails. this will allow said *SueprPorject* to
set the C_STANDARD vi CMAKE_ARGS.

### DIFF
--- a/src/libgit2/CMakeLists.txt
+++ b/src/libgit2/CMakeLists.txt
@@ -2,7 +2,12 @@
 # git library functionality.
 
 add_library(libgit2 OBJECT)
-set_target_properties(libgit2 PROPERTIES C_STANDARD 90)
+
+set(CSTD 90)
+if(C_STANDARD)
+	set(CSTD ${C_STANDARD})
+endif()
+set_target_properties(libgit2 PROPERTIES C_STANDARD ${CSTD})
 set_target_properties(libgit2 PROPERTIES C_EXTENSIONS OFF)
 
 include(PkgBuildConfig)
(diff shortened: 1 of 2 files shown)


257:
RFC really, not sure if it's the best way to do it. feel free to comment/edit etc.

boretrk:
Some thoughts that aren't specific to this PR but has to do with `C_STANDARD` in general:

1. I don't think the `C_STANDARD` set on the target is used for the added libraries in `deps/` (At least ntmlclient doesn't generate errors on `//`-style comments and `inline`)

3. If `C_STANDARD` isn't set for a specific target it defaults to `CMAKE_C_STANDARD` so maybe it would be better to only set that one once in the outer `CMakeLists.txt` and remove `C_STANDARD` from individual targets, but that would probably generate a whole bunch of warnings from `deps/` and make it necessary to relax the warning flags a bit or fix those warnings.

4. Since some C99 features like `inline` and flexible arrays are used I think it would be a good idea to default to C99 and compile with C90 in some integration test.
If we do that a bunch of `#ifdef` for different compilers could probably be removed since we could just check for C99 and let C90 systems be functional but somewhat inefficient.
To do something like that we would probably need to change how warnings are applied a bit since we might want to use different warnings on C90 and C99 (Maybe slap on `-pedantic` when compiling with C99)

ethomson:
Fixed via #6911

PR #6239, opened 2022-02-28
cmake: export libraries needed to compile against libgit2

ethomson:
If users are using us as a direct dependency (via `add_subdirectory`)
then they may want to know the libraries to link to; tell them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,6 +131,14 @@ if(BUILD_FUZZERS)
 endif()
 
 
+# Export for people who use us as a dependency
+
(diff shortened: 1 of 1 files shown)


tohosys:
unsubscribe

tohosys:
2022년 2월 28일 (월) 오후 1:21, Edward Thomson ***@***.***>님이 작성:

andrm:
Is there documentation on how to use this? We use libgit2 via git submodule in an out of tree bin dir and it worked well until 1.4.x, stopped working in 1.5.0 for linking problems.

andrm:
Found it finally: target name changed from "git2" to "libgit2package". Use the latter in your parent target_link_libraries.

PR #6240, opened 2022-02-28
Swap `GIT_DIFF_LINE_(ADD|DEL)_EOFNL` to match other Diffs

xphoniex:
Generating a diff using `git_diff_tree_to_tree` and `git_diff_from_buffer` return different result for the same patch of:
```
$ git diff 17ec0f3 394a717
diff --git a/.env b/.env
index f9fb22f..ed7b51e 100644
--- a/.env
+++ b/.env
@@ -1 +1 @@
-hello=12345
+hello=123456
\ No newline at end of file
```
Diff:
```
from repo:::
- hello=12345
+ hello=123456<                // Notice missing \n and `<` origin, meaning deletion
\ No newline at end of file
.
from buff:::
- hello=12345
+ hello=123456
> \ No newline at end of file
```

We didn't add the LF, we deleted it!

### DIFF
--- a/src/libgit2/patch_parse.c
+++ b/src/libgit2/patch_parse.c
@@ -558,9 +558,9 @@ fail:
 
 static int eof_for_origin(int origin) {
 	if (origin == GIT_DIFF_LINE_ADDITION)
-		return GIT_DIFF_LINE_ADD_EOFNL;
-	if (origin == GIT_DIFF_LINE_DELETION)
 		return GIT_DIFF_LINE_DEL_EOFNL;
+	if (origin == GIT_DIFF_LINE_DELETION)
+		return GIT_DIFF_LINE_ADD_EOFNL;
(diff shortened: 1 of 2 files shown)


xphoniex:
I'm looking at failed CIs and can't really tell if there's something wrong with the code when it showing leaked memory or something wrong with bitbucket that's causing this. Can you please take a look @ethomson ?

Test code is pretty straighforward, and the only thing I'm not freeing is [`git_diff_line *line;`](https://github.com/libgit2/libgit2/pull/6240/files#diff-b59442b2dcca409f1296551417d04b3da0e9e4ae073ad6bd0e8cf726f975ee13R295) which I couldn't find a `_free` function for e.g. `git_diff_line_free`.

xphoniex:
Polite reminder @ethomson .

PR #6244, opened 2022-03-13
Fix crash when regenerating a patch with unquoted spaces in filename

jorio:
This PR fixes a null pointer deref in the following scenario:

1. Write a patch that creates or deletes a file (with `new file mode ...` or `delete file mode ...`), **and** has unquoted filenames with spaces in the 1st line of the header.
For example, take PATCH_ORIGINAL_NEW_FILE_WITH_SPACE in patch_common.h:
    ```patch
    diff --git a/sp ace.txt b/sp ace.txt
    new file mode 100644
    index 000000000..789819226
    --- /dev/null
    +++ b/sp ace.txt
    @@ -0,0 +1 @@
    +a
    ```
2. Get a `git_diff` from the patch with `git_diff_from_buffer`
3. Regenerate a patch with `git_diff_to_buf`

I'm not sure the proposed fix is optimal, since it doesn't reproduce the same `diff --git` header line. But, it at least prevents the crash, and vanilla `git apply` accepts the resulting patch.

### DIFF
--- a/src/libgit2/diff_print.c
+++ b/src/libgit2/diff_print.c
@@ -316,10 +316,9 @@ static int diff_print_oid_range(
 static int diff_delta_format_path(
 	git_str *out, const char *prefix, const char *filename)
 {
-	if (!filename)
-	{
+	if (!filename) {
 		/* don't prefix "/dev/null" */
-		return git_str_puts(out, "/dev/null");
+ 		return git_str_puts(out, "/dev/null");
 	}
 
 	if (git_str_joinpath(out, prefix, filename) < 0)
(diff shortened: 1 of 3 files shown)


src/libgit2/diff_print.c
```diff
@@ -316,6 +316,12 @@ static int diff_print_oid_range(
 static int diff_delta_format_path(
 	git_str *out, const char *prefix, const char *filename)
 {
+	if (!filename)
+	{
+		/* don't prefix "/dev/null" */
+		return git_str_puts(out, "/dev/null");
+	}
```
ethomson, line 323:
```suggestion
	if (!filename) {
		/* don't prefix "/dev/null" */
 		return git_str_puts(out, "/dev/null");
	}
```

ethomson:
Thanks for the fix @jorio - I'm adding a minor formatting changes to match the libgit2 style a little bit better. I really appreciate the fix and the very nice tests.

PR #6246, opened 2022-03-16
Fix a string concatenation bug when validating extensions

bierbaum:
As builtin extensions are evaluated in the latter half of `check_valid_extension`, a string `cfg` is concatenated with the static string 'extension.' and the value from `builtin_extension`, before being compared with the configured value. This string is not being cleared while iterating through the names of the extensions. Because there is currently only one extension (`noop`), the bug was never noticeable.

This patch corrects the behavior by clearing the string on each iteration, as is done in for user extensions in the preceding block.


bierbaum:
I noticed this while testing the following patch: https://github.com/vermiculus/libgit2/tree/sallred/support-worktree-config

ethomson:
Oops, good catch. Thanks!

PR #6248, opened 2022-03-20
util: add fallback for non-standard getloadavg()

nielx:
The getloadavg() function is not part of the POSIX specification. While it is
implemented on *BSD and Linux, it is not available on the Haiku platform.
This causes linking errors when using libgit2. This commit adds a check if
the function is available while building for the target platform, and if it
is not, it will fall back to drand48() which is part of the POSIX standard
and should be available on all platforms.

While the drand48() function does not provide the same guarantee of
randomness as using system data, it does allow libgit2 continue to work on
more platforms.

This change also reintroduces the check for getentropy(), which seems to have
fallen through the cracks in the refactoring to the structure done in ef4ab29
and c3b7ace.

Finally, when building on Haiku, it now links to libgnu.so, which provides
the qsort_r symbol.

### DIFF
--- a/src/util/rand.c
+++ b/src/util/rand.c
@@ -106,7 +106,13 @@ GIT_INLINE(int) getseed(uint64_t *seed)
 		return -1;
 	}
 
+# if defined(GIT_RAND_GETLOADAVG)
 	getloadavg(loadavg, 3);
+# else
+	loadavg[0] = drand48();
+	loadavg[1] = drand48();
+	loadavg[2] = drand48();
+# endif
 
 	*seed = 0;
 	*seed |= ((uint64_t)tv.tv_usec << 40);
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -68,6 +68,12 @@ check_prototype_definition(qsort_r
 
 check_function_exists(qsort_s GIT_QSORT_S)
 
+# util/rand: getentropy and getloadavg
+
+check_function_exists(getentropy GIT_RAND_GETENTROPY)
+
+check_function_exists(getloadavg GIT_RAND_GETLOADAVG)
+
 # determine architecture of the machine
 
 if(CMAKE_SIZEOF_VOID_P EQUAL 8)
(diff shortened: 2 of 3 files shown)


src/util/rand.c
```diff
@@ -106,7 +106,13 @@ GIT_INLINE(int) getseed(uint64_t *seed)
 		return -1;
 	}
 
+# if defined(GIT_RAND_GETLOADAVG)
 	getloadavg(loadavg, 3);
+# else
+	loadavg[0] = drand48();
+	loadavg[1] = drand48();
+	loadavg[2] = drand48();
+# endif
```
ethomson, line 115: Thanks for this, it definitely makes sense to not use `getloadavg` where that function isn't available. I sort of wonder if we shouldn't change this to always try to use `drand48` (since we missed out on other random data from `/dev/urandom`) and then only mix in `getloadavg` data where available.

More concretely, I'm thinking initializing `loadavg`: `double loadavg[3] = { 0, 0, 0 };` and then:

```suggestion
# if defined(GIT_RAND_GETLOADAVG)
	getloadavg(loadavg, 3);
# endif
```

And then using `*rand48`... but should we be using `lrand48` instead of `drand48`? Are all the `*rand48` family of functions generally available or is `drand48` "special"? I must admit that I'm not as familiar with them.

nielx: Both `lrand48()` and `drand48()` are standard POSIX, though some man pages seem to suggest that `rand()` (which is standard C library) or `random()` (which is POSIX) are preferred. To be honest, I chose `drand48()` because it was the easiest no-cast solution, though they could be built differently as well.

As for the point of whether initializing the values to 0 would be alright, and then only add a value if `getloadavg()` is available, I do not think I can give much guidance as I do not know the code  base that well. My hunch is that the fundamental question is what these random numbers are used for. If it is for something highly important for security, like generating keys for cryptography, then I would imagine that adding an alternative randomness would be important, though with the caveat of pointing out that `drand48()` is a pseudo-random number generator, and I do not know if that would also classify as insecure in this particular use.

If the randomness is not used for cryptography, then I think initializing the values to `0.0` by default and only loading them up with something if `getloadavg()` is available is the cleanest solution.

nielx: Would you like me to rewrite the PR in any way?

Begasus:
Can confirm that with this patch cargo-c can be build which in turn enables us to build rav1e (on Haiku)

src/util/rand.c
ethomson: Sorry @nielx - I missed your question. I was hoping that we would just avoid mixing the randomness and the notion of a load average like in my suggestion.

Looking at `drand48`, it needs to be seeded, so it seems not optimal for us since this function is a seed for our PRNG.

I'm going ahead and merging this manually. Thanks for opening the PR!

ethomson:
Merged manually. Thanks again!

Issue #6254, opened 2022-03-28
Build fails on Windows (both mingw and VC++) with option -DEMBED_SSH_PATH


apnadkarni:
I have alternatives for fixes but need advice before making a pull request.

### Reproduction steps

VC++

```
cmake  .. -A x64 -DBUILD_EXAMPLES=ON -DEMBED_SSH_PATH="D:/src/AAThirdparty/C,C++/libssh2-1.10.0" -DCMAKE_INSTALL_PREFIX="D:/temp/lg2" -DDEPRECATE_HARD=ON  -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON
cmake --build .
```

MingW
```
cmake .. -G "MinGW Makefiles" -DBUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=Release -DUSE_BUNDLED_ZLIB=ON -DEMBED_SSH_PATH=/d/src/AAThirdparty/C,C++/libssh2-1.10.0 -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON  -DDEPRECATE_HARD=ON
cmake --build .
```
### Expected behavior

Build completes successfully as it did for 1.3.0.

### Actual behavior

Build fails. See below for diagnosis and details.

### Version of libgit2 (release number or SHA1)
1.4.x, main (1.5.x)

### Operating system(s) tested
Windows, MingW and VC++

### Details

There are multiple issues with the EMBED_SSH_PATH option. The first is that the libssh is built twice - once directly via [src/CMakeLists.txt](https://github.com/libgit2/libgit2/blob/182d0d1ee933de46bf0b5a6ec269bafa77aba9a2/src/CMakeLists.txt#L85) and then again via the include of [SelectSSH](https://github.com/libgit2/libgit2/blob/182d0d1ee933de46bf0b5a6ec269bafa77aba9a2/src/CMakeLists.txt#L99).

Clearly one of these has to be removed. Both alternatives lead to further issues as below.

#### Alternative 1 - keep the `include SelectSSH`

[...]

PR #6256, opened 2022-03-29
Fix for build failure with -DEMBED_SSH_PATH.

apnadkarni:
https://github.com/libgit2/libgit2/issues/6254
This fix is Alternative 2 in above link (do not use `SelectSSH.cmake` on Windows if `EMBED_SSH_PATH` is enabled. Use the fragment in `CMakeLists.txt` instead)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,7 +1,13 @@
 add_library(git2internal OBJECT)
 set_target_properties(git2internal PROPERTIES C_STANDARD 90)
-set_target_properties(git2internal PROPERTIES C_EXTENSIONS OFF)
 
+# Embedded libssh require compiler extensions (gnu90, not c90) because it
+# uses the "inline" keyword
+if(WIN32 AND EMBED_SSH_PATH)
+	set_target_properties(git2internal PROPERTIES C_EXTENSIONS ON)
+else()
+	set_target_properties(git2internal PROPERTIES C_EXTENSIONS OFF)
+endif()
 
 if(DEPRECATE_HARD)
         add_definitions(-DGIT_DEPRECATE_HARD)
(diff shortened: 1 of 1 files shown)


ethomson (commented):
Hmm. I'm surprised that we're including the libssh sources as part of the git2internal target. I would have expected we build them separately. I wonder if instead of turning extensions off for all of the git2internal target if we can turn them off just for libssh2?

I don't like this weird pattern where we include libssh2 as if it's our own. But it apparently has some use. Limiting the blast radius on this change would be a nice compromise if possible.

src/CMakeLists.txt
```diff
@@ -88,6 +94,8 @@ if(WIN32 AND EMBED_SSH_PATH)
 	target_sources(git2internal PRIVATE ${SRC_SSH})
 
 	list(APPEND LIBGIT2_SYSTEM_INCLUDES "${EMBED_SSH_PATH}/include")
+        list(APPEND LIBGIT2_SYSTEM_LIBS "crypt32" "bcrypt")
+        list(APPEND LIBGIT2_PC_LIBS "-lcrypt32" "-lbcrypt")
 	file(WRITE "${EMBED_SSH_PATH}/src/libssh2_config.h" "#define HAVE_WINCNG\n#define LIBSSH2_WINCNG\n#include \"../win32/libssh2_config.h\"")
```
ethomson, line 99: In fact, instead of trying to create a libssh2 config, why don't we just treat it as a normal dep and run `cmake` on it, exactly like we do for our other dependencies?

apnadkarni:
> I wonder if instead of turning extensions off for all of the git2internal target if we can turn them off just for libssh2?

I'm afraid I have no idea how :-) Not knowing cmake I'm confused as to how it decides what flags to pick up for dependencies.

> I don't like this weird pattern where we include libssh2 as if it's our own. But it apparently has some use. Limiting the blast radius on this change would be a nice compromise if possible.

I agree. I'm guessing it is easier to include sources directly on Windows than build that library and statically link and that is why it is implemented that way (note 1.3.0 also built the same way). Note this is also true of `zlib` if the `USE_BUNDLED_ZLIB` option is on.

> In fact, instead of trying to create a libssh2 config, why don't we just treat it as a normal dep and run cmake on it, exactly like we do for our other dependencies?

Well, as far as I can tell, other dependencies also seem to do this afaict (when the bundling options are enabled). For example, see the relevant fragments for [zlib](https://github.com/libgit2/libgit2/blob/db4461d556ddb7915349d094ba74ef01d7a2d701/cmake/SelectZlib.cmake#L29) and [http-parser](https://github.com/libgit2/libgit2/blob/db4461d556ddb7915349d094ba74ef01d7a2d701/cmake/SelectHTTPParser.cmake#L14)

The difference is that those have no problem with the change of the C standard from gnu90 to c90.

I think I need to actually learn cmake (really no clue how it operates with respect to paths and dependencies) before doing anything further. Unfortunately, time spent on OSS being limited as for everyone else, and not being a priority for my primary project, that might not happen for couple of weeks.

In the meanwhile, you can reject this PR or keep it pending as per your preference.

apnadkarni:
I'm closing this as I am using the workaround I describe in #6254. Note builds with EMBED_SSH_PATH remains broken.

PR #6260, opened 2022-04-05
midx: Fix an undefined behavior (left-shift signed overflow)

lhchavez:
There was a missing check to ensure that the `off64_t` (which is a
signed type) didn't overflow when parsing it from the midx file. This
shouldn't have huge repercusions since the parsed value is immediately
validated afterwards, but then again, there is no such thing as "benign"
undefined behavior.

This change makes all the bitwise arithmetic happen with unsigned types
and is only casted to `off64_t` until the very end.

Thanks to Taotao Gu for finding and reporting this!


ethomson:
Thanks for the fix @lhchavez!

PR #6276, opened 2022-04-14
remote: Delete a now-inexistent API declaration

lhchavez:
6fc6eeb66c40310086c8f059cae41de69ad4c6da replaced the remote options
with `git_remote_connect_options`. The function definitions were
removed, but one function declaration remained, causing linker errors if
one tried to use it.

This change removes the declaration of
`git_transport_smart_proxy_option` to better reflect reality.


ethomson:
Oops, good catch. Thanks!

PR #6277, opened 2022-04-16
Fix missing include

cschlack:
`#include <git2/sys/transport.h>` leads to the following error message:

```
compiling test.c
In file included from test.c:2:
/opt/homebrew/Cellar/libgit2/1.4.2_1/include/git2/sys/transport.h:38:9: error: unknown type name 'git_remote_connect_options'
                const git_remote_connect_options *connect_opts);
                      ^
```

`git_remote_connect_options` is defined in `git2/remote.h`

Including `"git2/remote.h"` in `include/git2/sys/transport.h` fixes the issue

### DIFF
--- a/include/git2/sys/transport.h
+++ b/include/git2/sys/transport.h
@@ -9,6 +9,7 @@
 #define INCLUDE_sys_git_transport_h
 
 #include "git2/net.h"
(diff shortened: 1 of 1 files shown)


ethomson:
Good catch, thanks!

PR #6281, opened 2022-04-20
Push options

bansheerubber:
My classmates and I, with help from someone on the GitKraken team, implemented a way to upload push options.

A refresher on what push options are: https://git-scm.com/docs/git-push#Documentation/git-push.txt--oltoptiongt

Justification: Push options are a popular way to specify additional parameters for Git hosting services. GitLab uses them to specify CI variables in their CI pipeline (https://docs.gitlab.com/ee/user/project/push_options.html), for instance. With the advent of Git GUIs using libgit2 to implement their Git support, we feel it important to reduce the functionality gap between the Git CLI and libgit2.

---

Intended use:
```c
git_remote *remote = ...;
git_strarray refspecs = ...;
git_push_options options;
const char *args[2] = {
    "Lorem Ipsum es simplemente el texto",
    "That still only counts as one!"
};
options.arguments = (git_strarray){
    args,
    2
};

git_remote_push(remote, &refspecs, &options);
```
Should be noted that the way we designed this potentially breaks ABI compatibility. Open to suggestions on how to re-implement if this is not desired.

Let me know if there are any questions on the implementation. Tests ended up being a little weird because there didn't seem to be a nice way to implement them in the normal `ctest` suite, so we implemented our tests as CI tests. The changes we made aren't super complicated, so I'm hoping the code should be more-or-less self-explanatory but I of course could be wrong about that.

### DIFF
--- a/ci/hooks/pre-receive
+++ b/ci/hooks/pre-receive
@@ -1,0 +1,2 @@
+#!/bin/sh
+printf "$GIT_PUSH_OPTION_0$GIT_PUSH_OPTION_1$GIT_PUSH_OPTION_2" > %file%
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -84,6 +84,14 @@ if [ -z "$SKIP_GITDAEMON_TESTS" ]; then
 	echo "Starting git daemon..."
 	GITDAEMON_DIR=`mktemp -d ${TMPDIR}/gitdaemon.XXXXXXXX`
 	git init --bare "${GITDAEMON_DIR}/test.git" >/dev/null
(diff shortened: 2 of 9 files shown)


ethomson:
Cool! Thanks for the contribution @bansheerubber! I'll review it 🔜 . 🎉

bansheerubber:
> Cool! Thanks for the contribution @bansheerubber! I'll review it soon . tada

Thank you! Forgot to tag them in the post, but a lot of help was received from my classmates @pireads, @lotdeef, and @PSI497.

ethomson:
Fixed via #6439 - thanks much for your contributions.

Issue #6241, opened 2022-03-07
cannot find -llibgit2package when build examples


lucasoshiro:
Hi!

In the current `main` branch (commit 2a0d0bd19b5d13e2ab7f3780e094404828cbb9a7), I was trying to build the examples and then I got this error:

~~~
/usr/bin/ld: cannot find -llibgit2package: No such file or directory
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/lg2.dir/build.make:545: lg2] Error 1
make[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/lg2.dir/all] Error 2
make: *** [Makefile:91: all] Error 2
~~~

After a bisect, I found that this was probably due to the change in 91ba089663f5efc3bd4ba14a5099372cf5ce57a6, in PR #6133. In the commit before the merge of that PR, it doesn't happen. I don't have knowledge in CMake, so I don't know how to solve this.

### Reproduction steps

Building examples, from the project root:

~~~bash
mkdir examples/build
cd examples/build
cmake ..
cmake --build .
~~~

### Expected behavior

Build the examples without errors, as in commit eca9e1cfec2be5a55401290de9eb944bcb54c867

### Actual behavior

The `ld` error `cannot find -llibgit2package: No such file or directory`.

### Version of libgit2 (release number or SHA1)
In Manjaro, 1.4.1-1. In Debian, 1.1.0+dfsg.1-4.

### Operating system(s) tested

- Manjaro 21.2.4
  - cmake 3.22.2
  - gcc 11.2.0
  - GNU Make 4.3
  - GNU ld 2.38

- Debian 11 Bullseye (in a Docker container)
  - cmake 3.18
  - gcc 10.2.1
  - GNU make 4.3
  - GNU ld 2.35.2

ethomson:
Ah, thanks.  I forgot that we had a separate `CMakeFiles.txt` in the examples folder.  😢

ethomson:
I grepped through our docs looking for where we might suggest that the examples were buildable outside of libgit2. I don't see anything. The suggestion is to `cmake -DBUILD_EXAMPLES=ON`.

If there's a place that I've missed, let's remove that and keep the examples as part of the libgit2 build. So, closing this. 🙏

PR #6282, opened 2022-04-20
Fix examples/CMakeLists.txt

thehans:
This fixes issue #6241.  

`examples/CMakeListst.txt` has been made independent of the rest of libgit2's build.  
This is done by making use of pkg-config, since a `libgit2.pc` file is provided during installation.

Source files have been listed explicitly, as `file(GLOB ...)` is [discouraged for setting sources](https://cmake.org/cmake/help/latest/command/file.html?highlight=file%20glob#glob).
> Note We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. ...

Some build instructions have also been added to the examples readme.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /build/
+/examples/build/
 .DS_Store
 *~
 .*.swp
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@
 /tags
 CMakeSettings.json
 .vs
+.vscode
 .idea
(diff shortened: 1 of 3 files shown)


ethomson:
I grepped through our docs looking for where we might suggest that the examples were buildable outside of libgit2. I don't see anything. The suggestion is to `cmake -DBUILD_EXAMPLES=ON`.

If there's a place that I've missed, let's remove that and keep the examples as part of the libgit2 build. So, closing this. 🙏

PR #6285, opened 2022-04-28
blame: add blame progress callback

Murmele:


Murmele:
Is it possible to find out of how many suspects the blame exists to show also a progress?

Murmele:
@ethomson I think this is finished

Murmele:
@ethomson

PR #6288, opened 2022-04-29
A couple of simplications around mwindow

carlosmn:
I was looking at this code and found a couple of places where we've way too verbose for what we're doing and figured we can make this easier to read.

The first change makes us include the hash/extra size in the window "contains" check as we only ever use them (and would want to use them) in pairs.

The second change eliminates some rather verbose comparisons in favour of maths. The values are already `1` and `-1` so we can multiply instead of chaining several comparisons. This is likely to be gentler on the branch predictor due to fewer conditional jumps but that's unlikely to be measurable.


ethomson:
👍 Thanks @carlosmn

PR #6290, opened 2022-05-01
commit: add support to take ENV variables for commit info

venlinz:
Fixes issue: https://github.com/libgit2/libgit2/issues/3751

This will add support for git_commit__create_internal() for getting AUTHOR and COMMITTER information from the ENVIRONMENT.
variables considered as follows,
           GIT_AUTHOR_NAME
           GIT_AUTHOR_EMAIL // for AUTHOR information
and
           GIT_COMMITTER_NAME
           GIT_COMMITTER_EMAIL // for  information

if one of the env variable in a set(AUTHOR env pair)
and other is not set or set to empty string(eg.: GIT_AUTHOR_NAME is set and GIT_AUTHOR_EMAIL is set to empty string or not setted at all) this patch will
defaults to global values for the empty string value or unsetted ENV variable and takes the properly set ENV variable with non-empty string and stores as a `git_signature` structure in `git_commit__create_internal`, but in git CLI the commit will fail instead of defaulting to
global defaults for empty string value for a ENV variable such as GIT_AUTHOR_EMAIL.

on the other hand if one of a variable in pair is not set(eg. GIT_AUTHOR_NAME is set and GIT_AUTHOR_EMAIL is not set) and other one is set to valid value ENV git CLI sets global defaults values for unsetted ENV variable.
I'm using git version 2.36.0.

### DIFF
--- a/src/libgit2/commit.c
+++ b/src/libgit2/commit.c
@@ -139,6 +139,8 @@ static int git_commit__create_internal(
 	git_str buf = GIT_STR_INIT;
 	const git_oid *current_id = NULL;
 	git_array_oid_t parents = GIT_ARRAY_INIT;
+	git_signature *env_author = NULL;
+	git_signature *env_committer = NULL;
 
 	if (update_ref) {
 		error = git_reference_lookup_resolved(&ref, repo, update_ref, 10);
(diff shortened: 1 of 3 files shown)


venkatesh-coder:
I am working on adding tests to check the functionality of this patch , if you think the patch does the correct job please let me know, if it's not solves the please let me know where I went wrong. I eagerly looking forward for reply thank you.

ethomson:
Fixed via #6706 ; thanks for the patch, sorry that this went a while without a review and that caused a duplicated effort. 🙏

PR #6291, opened 2022-05-03
midx: do not verify the checksum on load

carlosmn:
This is something we only want to do during explicit verification rather than on
every load.

Verifying does not seem like a big deal when we're running with test workloads
but once your `multi-pack-index` reaches gigabytes, we spend more time hashing
this than doing any work.

This is basically what we do for single-pack index files as well. We trust that he checksum is fine while running regularly.

cc @lhchavez as author of our implementation

### DIFF
--- a/src/libgit2/midx.c
+++ b/src/libgit2/midx.c
@@ -179,7 +179,6 @@ int git_midx_parse(
 	uint32_t i;
 	off64_t last_chunk_offset, chunk_offset, trailer_offset;
 	size_t checksum_size;
-	unsigned char checksum[GIT_HASH_SHA1_SIZE];
 	int error;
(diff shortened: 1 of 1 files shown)


carlosmn:
On a 2GB midx, looking up the commit for HEAD takes 12s vs 0.3s with this change (this is with bundle/ruby timing on top)

carlosmn:
Note that such checks are appropriate when using the midx to write out a new one, so as not to propagate corrupted data.

lhchavez (approved):
welp, that's certainly way too slow. let's go with this and if it's causing problems later we can figure out where to re-add this in a place that makes a bit more sense.

ethomson:
Just checking: I assume this is how git itself behaves?

Issue #6294, opened 2022-05-07
patch: unquoted paths containing spaces are not parsed correctly


cavaquinho:
When a patch contains an unquoted pathname with spaces, there are multiple situations where that path ends up being parsed incorrectly, potentially resulting in a crash.

In a parsed patch for:
1.  a newly created text file: the old path is `NULL` if the path has spaces, but it is the correct non-null path if spaces are not present.
2. a renamed binary file: paths with spaces are parsed only up to their first space character, i.e., they are parsed incorrectly.
3. a non-renamed, modified binary file: paths are not parsed at all. They are `NULL`.

When a path is `NULL` in a patch/diff:

- The patch/diff is unprintable. `git_diff_to_buf()` crashes.
- The patch/diff is un-applicable. Calling `git_apply()` with a diff created from such a patch fails with an "invalid 'filename'" assertion, which is triggered by an assert in a function that expects the pathname to never be NULL.

I have some tests and some changes that fix this. I’ll submit them in a push request.

### Reproduction steps
I’ll be adding a few test cases in a PR that demonstrate the incorrect behavior.

### Expected behavior
It should always be possible to parse the correct file paths from a diff header, even if they have spaces and are unquoted.

### Actual behavior
Crash due to `NULL` pointer dereference, or the incorrect pathnames are parsed from the patch text as stated above.

### Version of libgit2 (release number or SHA1)
main @ bdab22384cc61d315005a65456a9f9563bb27c8f

### Operating system(s) tested
macOS 12.3.1 (21E258)

PR #6295, opened 2022-05-07
Cavaquinho/patch unquoted path w space fix 6294

cavaquinho:
Fixes issue #6294: unquoted paths containing spaces are not parsed correctly.

The comments in the code explain some of problems in the existing code and the solutions that were adopted.

### DIFF
--- a/src/libgit2/patch_parse.c
+++ b/src/libgit2/patch_parse.c
@@ -89,18 +89,31 @@ static int parse_header_path_buf(git_str *path, git_patch_parse_ctx *ctx, size_t
 	return 0;
 }
 
-static int parse_header_path(char **out, git_patch_parse_ctx *ctx)
+static int parse_header_path_up_to_len(
+	char **out,
+	git_patch_parse_ctx *ctx,
+	size_t header_path_len)
 {
 	git_str path = GIT_STR_INIT;
 	int error;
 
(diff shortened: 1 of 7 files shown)


arroz:
@ethomson Can this one also be reviewed and merged? Thank you. :)

src/libgit2/patch_parse.c
```diff
@@ -338,23 +351,276 @@ static int parse_header_dissimilarity(
...
+				old_path_minus_prefix_len = old_path_len - old_prefix_len;
+				new_path_minus_prefix_len = new_path_len - new_prefix_len;
+
+				if (old_path_minus_prefix_len == new_path_minus_prefix_len &&
+					strncmp(old_path_base + old_prefix_len,
+							new_path_base + new_prefix_len,
+							old_path_minus_prefix_len) == 0)
+				{
+					*out_old_path = strndup(old_path_base, old_path_len);
+					*out_new_path = strndup(new_path_base, new_path_len);
```
zawata, line 431:
```suggestion
					*out_old_path = git__strndup(old_path_base, old_path_len);
					*out_new_path = git__strndup(new_path_base, new_path_len);
```

I get an undefined symbol here on msvc. I think the error is dependent on the version(I'm on 2017) but at any rate, libgit2 uses it's own versions of these symbols

Issue #6296, opened 2022-05-07
diff: zeroed out OIDs in reversed index to workdir diff


cavaquinho:
The deltas in a reversed `git_diff_index_to_workdir()` may contain zeroed out OIDs where there should be a non-zero one. The same non-reversed diff contains all the correct OIDs in the same equivalent non-reversed deltas.

### Reproduction steps
I’ll be adding a test case in a PR that demonstrate the incorrect behavior.

### Expected behavior
The OIDs in the deltas of a reversed `git_diff_index_to_workdir()` should be the same as the OIDs in the equivalent non-reversed diff (just reversed).

### Actual behavior
The OIDs in some deltas are incorrectly zeroed out.

### Version of libgit2 (release number or SHA1)
main @ bdab22384cc61d315005a65456a9f9563bb27c8f

### Operating system(s) tested
macOS 12.3.1 (21E258)

PR #6297, opened 2022-05-07
Cavaquinho/diff index to workdir zeroed oids 6296

cavaquinho:
Fixes issue #6296: zeroed out OIDs in reversed index to workdir diff


arroz:
@ethomson Can this one be reviewed and merged? Thanks.

PR #6298, opened 2022-05-09
add callback before a remote connects and add a callback before a remote will be disconnected

Murmele:
Problem Description:
When updating submodules, the remote of the submodule is not visible outside of the submodule (submodule is a private type and therefore not intended to be used outside). So in that case it is not possible to abort an update operation by aborting the remote connection.

Benefit:
If during the submodule update (fetch, ...) a callback will be called, outside of the submodule the remote is available. This remote can be stored and if the application wants to abort the operation it just stops the operation on that remote.

from #5730 
In this particular case, I added the connect/disconnect callbacks because I needed to remember the remote to cancel the transfer later on. This is most important in submodule update, which looks up the remotes internally. Without this, there isn't a way to cancel the transfers initiated by submodule update.

### DIFF
--- a/tests/libgit2/network/remote/remotes.c
+++ b/tests/libgit2/network/remote/remotes.c
@@ -4,6 +4,7 @@
 #include "remote.h"
 
 static git_remote *_remote;
+static git_remote *_https_remote;
 static git_repository *_repo;
 static const git_refspec *_refspec;
 
@@ -12,6 +13,7 @@ void test_network_remote_remotes__initialize(void)
 	_repo = cl_git_sandbox_init("testrepo.git");
 
 	cl_git_pass(git_remote_lookup(&_remote, _repo, "test"));
+	cl_git_pass(git_remote_lookup(&_https_remote, _repo, "https"));
 
 	_refspec = git_remote_get_refspec(_remote, 0);
 	cl_assert(_refspec != NULL);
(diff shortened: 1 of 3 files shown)


ethomson:
> Why git_remote_ready_cb remote_ready; is not defined as GIT_CALLBACK()?

It is - https://github.com/libgit2/libgit2/blob/258df9c18d4876ad59a2cfa9d78a313f867d633e/include/git2/remote.h#L538 ? 🤔

Murmele:
> > Why git_remote_ready_cb remote_ready; is not defined as GIT_CALLBACK()?
> 
> It is -
> 
> https://github.com/libgit2/libgit2/blob/258df9c18d4876ad59a2cfa9d78a313f867d633e/include/git2/remote.h#L538
> ? 🤔

Ah I haven't seen this thanks. Is this the way it should be added to the struct? So I will change it for my callbacks too

Murmele:
Current error in the tests: "remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
"

Murmele:
@ethomson

AHSauge:
What is needed here to progress this pull request? This is one of the few things remaining before Gittyup can be built against a plain upstream libgit2.

Looking critically at the pull request myself, I imagine the following needs to be resolved

- [ ] Tests and builds need to pass, which currently isn't the case
- [ ] I don't believe the changes to `git_remote_disconnect` is backwards compatible, thus this needs to be added as a new function to avoid breaking the API. Alternatively, I imagine that `git_remote_callbacks` could be stored in the `git_remote` object avoid needing this change. From what I can tell, this would be similar to `git_push` for example.

Anything else?

I'm happy to provide an updated pull request for this if needed.

PR #6301, opened 2022-05-10
checkout: notify for deleted files

Murmele:
Notify caller that the file was deleted and will be added to the working directory

In GitAhead and Gittyup it is used to show the user a notification that the file is checked out.


Murmele:
TODO: how the second condition of the if in tree.c line 713 and 701 gets triggered?

ethomson:
I don't think that I understand. Can you add a unit test to help elucidate this change?

Murmele:
Hi @ethomson sorry I am not yet finish with the pullrequest. I will work on it to improve

PR #6302, opened 2022-05-13
New checkout option: disabled_filters

julianmesa-gitkraken:
Add the possibility of disabling specific filters instead of disabling all of them.

For example we can disable LFS filter during checkout, so we can execute `git lfs checkout` after checkout to improve performance in LFS repos.

Special thanks to @AlexaXs for helping me

### DIFF
--- a/include/git2/checkout.h
+++ b/include/git2/checkout.h
@@ -342,6 +342,8 @@ typedef struct git_checkout_options {
 
 	/** Payload passed to perfdata_cb */
 	void *perfdata_payload;
+
+	git_strarray disabled_filters; /**< don’t apply these specific filters */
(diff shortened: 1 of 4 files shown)


julianmesa-gitkraken:
@ethomson Any reason why this PR is closed?

ethomson:
Hi @julianmesa-gitkraken - here, too, I don't know. Even in the timeline view it says it was closed after your query. I'm stumped. 🤷‍♂️

ethomson:
I understand that Git Kraken has stopped using libgit2 - which is disappointing, but understandable. I'm going to close this PR as a result. We should improve filter execution perf instead of trying to avoid executing filters. 🙏

PR #6303, opened 2022-05-13
filter: Fix Segfault

zawata:
Registering a git_filter without setting the `stream` callback causes the git_filter_list_apply_* functions(which eventually call filter.c:setup_stream) to call buffered_legacy_stream_new which sets the legacy_write_fn callback instead of the write_fn callback. but buffered_stream_close will only ever call write_fn callback...thus the segfault.

This appears to have been introduced with the git_buf/git_str split.
Hopefully I've taken the right approach here? I ran this against the nodegit test suite(which hasn't implemented filter streaming yet) and it fixed the crashes I was experiencing.

### DIFF
--- a/src/libgit2/filter.c
+++ b/src/libgit2/filter.c
@@ -902,6 +902,16 @@ static int buffered_stream_close(git_writestream *s)
 
 	GIT_ASSERT_ARG(buffered_stream);
 
+#ifndef GIT_DEPRECATE_HARD
+	if(buffered_stream->write_fn == NULL) {
+		error = buffered_stream->legacy_write_fn(
+			buffered_stream->filter,
+			buffered_stream->payload,
+			(git_buf *)buffered_stream->output,
+			(git_buf *)&buffered_stream->input,
+			buffered_stream->source);
+	} else
+#endif
 	error = buffered_stream->write_fn(
 		buffered_stream->filter,
 		buffered_stream->payload,
--- a/src/libgit2/filter.c
+++ b/src/libgit2/filter.c
@@ -904,13 +904,15 @@ static int buffered_stream_close(git_writestream *s)
 
 #ifndef GIT_DEPRECATE_HARD
 	if (buffered_stream->write_fn == NULL) {
-		 git_buf legacy_output = GIT_BUF_INIT,
-		         legacy_input = GIT_BUF_INIT;
+		git_buf legacy_output = GIT_BUF_INIT,
+		        legacy_input = GIT_BUF_INIT;
 
-		 legacy_output.ptr = buffered_stream->output->ptr;
-		 legacy_output.size = buffered_stream->output->size;
-		 legacy_input.ptr = buffered_stream->input.ptr;
-		 legacy_input.size = buffered_stream->input.size;
+		legacy_output.ptr = buffered_stream->output->ptr;
+		legacy_output.size = buffered_stream->output->size;
+		legacy_output.reserved = buffered_stream->output->asize;
+		legacy_input.ptr = buffered_stream->input.ptr;
+		legacy_input.size = buffered_stream->input.size;
+		legacy_input.reserved = buffered_stream->input.asize;
 
 		error = buffered_stream->legacy_write_fn(
 			buffered_stream->filter,
--- a/src/libgit2/filter.c
+++ b/src/libgit2/filter.c
@@ -903,12 +903,20 @@ static int buffered_stream_close(git_writestream *s)
 	GIT_ASSERT_ARG(buffered_stream);
 
 #ifndef GIT_DEPRECATE_HARD
-	if(buffered_stream->write_fn == NULL) {
+	if (buffered_stream->write_fn == NULL) {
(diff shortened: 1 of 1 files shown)


zawata:
I'm currently just casting the input and output `git_str *` to `git_buf *` which I'm sure isn't best practice, but git_but_tostr was wiping the git_buf. I'm not sure of a better way to do it.

src/libgit2/filter.c
```diff
@@ -902,6 +902,16 @@ static int buffered_stream_close(git_writestream *s)
...
 
+#ifndef GIT_DEPRECATE_HARD
+	if(buffered_stream->write_fn == NULL) {
+		error = buffered_stream->legacy_write_fn(
+			buffered_stream->filter,
+			buffered_stream->payload,
+			(git_buf *)buffered_stream->output,
+			(git_buf *)&buffered_stream->input,
+			buffered_stream->source);
+	} else
```
ethomson, line 913:
```suggestion
	if (buffered_stream->write_fn == NULL)
		error = buffered_stream->legacy_write_fn(
			buffered_stream->filter,
			buffered_stream->payload,
			(git_buf *)buffered_stream->output,
			(git_buf *)&buffered_stream->input,
			buffered_stream->source);
	else
```

ethomson:
Instead of casting a `git_str` to a `git_buf` - which is safe today but may change in the future - how does an explicit decomposition sound? Can you check my suggestion @zawata ?

src/libgit2/filter.c
ethomson: Yes, you're right, I think that we do. Otherwise we'll end up leaking memory?

ethomson:
I think that we might also need to swap back after invocation? Can you double-check me on this with your repro cases?

```diff
commit fdc2f61adda5805232d120369f3d2e10e72cb14b
Author: Edward Thomson <ethomson@edwardthomson.com>
Date:   Mon Jun 20 11:51:08 2022 -0400

    filter: restore `git_str` values for legacy write
    
    We need to swap the `git_buf` and `git_str` values both before and after
    legacy filter invocation to ensure that we free the memory allocated
    correctly.

diff --git a/src/libgit2/filter.c b/src/libgit2/filter.c
index 788ed9cb5..80a3cae67 100644
--- a/src/libgit2/filter.c
+++ b/src/libgit2/filter.c
@@ -893,6 +893,17 @@ static int buffered_stream_write(
 	return git_str_put(&buffered_stream->input, buffer, len);
 }
 
+#ifndef GIT_DEPRECATE_HARD
+# define BUF_TO_STRUCT(b, s) \
+	(b)->ptr = (s)->ptr; \
+	(b)->size = (s)->size;  \
+	(b)->reserved = (s)->asize;
+# define STRUCT_TO_BUF(s, b) \
+	(s)->ptr = (b)->ptr; \
+	(s)->size = (b)->size; \
+	(s)->asize = (b)->reserved;
+#endif
+
 static int buffered_stream_close(git_writestream *s)
 {
 	struct buffered_stream *buffered_stream = (struct buffered_stream *)s;
@@ -907,12 +918,8 @@ static int buffered_stream_close(git_writestream *s)
 		git_buf legacy_output = GIT_BUF_INIT,
 		        legacy_input = GIT_BUF_INIT;
 
-		legacy_output.ptr = buffered_stream->output->ptr;
-		legacy_output.size = buffered_stream->output->size;
-		legacy_output.reserved = buffered_stream->output->asize;
-		legacy_input.ptr = buffered_stream->input.ptr;
-		legacy_input.size = buffered_stream->input.size;
-		legacy_input.reserved = buffered_stream->input.asize;
+		BUF_TO_STRUCT(&legacy_output, buffered_stream->output);
+		BUF_TO_STRUCT(&legacy_input, &buffered_stream->input);
 
 		error = buffered_stream->legacy_write_fn(
 			buffered_stream->filter,
@@ -920,6 +927,9 @@ static int buffered_stream_close(git_writestream *s)
 			&legacy_output,
 			&legacy_input,
 			buffered_stream->source);
+
+		STRUCT_TO_BUF(buffered_stream->output, &legacy_output);
+		STRUCT_TO_BUF(&buffered_stream->input, &legacy_input);
 	} else
 #endif
 	error = buffered_stream->write_fn(
```

ethomson:
@zawata LMK when this branch is 👍 and I'll merge it. We'll ship a new 1.4 once this lands. 🥳

zawata:
sorry I've been fighting my build system all morning :sweat_smile: 

you will need to copy back the buf contents. I didn't test it after your first suggestion and my inital PR worked because I was just casting so the git_str's got updated anyways.

I applied your patch and barring any ci failures(which I don't anticipate) then this should be good to go i think

ethomson:
> you will need to copy back the buf contents. I didn't test it after your first suggestion and my inital PR worked because I was just casting so the git_str got updated anyways.

Yep, that was my oversight. 😓 

> I think that CI failure is unrelated?

Definitely. Google Source has been... frustrating on many levels lately.

Thanks for this PR @zawata !

PR #6304, opened 2022-05-17
Fix degraded performance using GIT_USE_NSEC on repos cloned with GIT_USE_NSEC disabled

julianmesa-gitkraken:
Fix degrade performance using GIT_USE_NSEC on repos cloned with GIT_USE_NSEC disabled.
We just mimic the GIT_USE_NSEC disabled behavior when nanoseconds are 0.
There is a edge case when nanoseconds is 0 in a repo cloned with GIT_USE_NSEC enabled, never found it but anyway that file will be handle fine just the sha1 will be generated to check if has changed.

### DIFF
--- a/src/libgit2/index.h
+++ b/src/libgit2/index.h
@@ -84,6 +84,8 @@ GIT_INLINE(bool) git_index_time_eq(const git_index_time *one, const git_index_ti
 		return false;
 
 #ifdef GIT_USE_NSEC
+	if (one->nanoseconds == 0 || two->nanoseconds == 0)
+		return true;
 	if (one->nanoseconds != two->nanoseconds)
 		return false;
 #endif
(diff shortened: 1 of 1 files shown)


ethomson:
Hmm.  Not sure that I understand what the problem is exactly yet. What is not supporting nanoseconds in the index in 2022? Are you mixing and matching git clients that understand nanoseconds with ones that do not? 🤔

ianhattendorf:
NodeGit for the longest time didn't have `GIT_USE_NSEC` enabled. It was enabled back in 2019 but reverted due to this issue affecting performance for users who had cloned repos with `GIT_USE_NSEC` disabled. We're hoping that being able to work in "mixed mode" will allow us to transition users over to nanosecond precision.

ethomson:
Gotcha. I wonder what git itself does here? It may have a more graceful upgrade strategy. 🤔

julianmesa-gitkraken:
@ethomson Any reason why this PR is closed?

ethomson:
I simply can't find any similar behavior in git that this emulates.

My understanding of the problem here is:
* update an index in a git client that is built with nanoseconds
* update an index in a git client that is _not_

During a movement from a git client that _doesn't_ understand nanoseconds to one that _does_, it will thrash the index, adding the nanoseconds. This is regrettable, but it _should_ be a one-time operation.

I understand the `== 0` test during racy detection, but I'm concerned about the larger implications. It's been 18 months, and I keep coming back to this PR, staring at it, and feeling that it's risky.

My understanding is that gitkraken has decided to embrace git instead of libgit2. So it seems like you have a different set of problems, and this is not a priority for you any longer.

Unless you continue to have interest here, and we can understand how this matches git compatibility or otherwise de-risk this, I'm going to close this.

PR #6305, opened 2022-05-17
refs: fix missing error message

zawata:
Super minor but it appears the packed ref lookup rewrite(#6138) didn't set the "not found" error message correctly and it broke a test in nodegit :sweat_smile:

I haven't checked for other missing errors in this file but when I get some time I can look it over just in case

### DIFF
--- a/src/libgit2/refdb_fs.c
+++ b/src/libgit2/refdb_fs.c
@@ -740,7 +740,7 @@ static int packed_lookup(
 			return 0;
 		}
 	}
-	return GIT_ENOTFOUND;
+	return ref_error_notfound(ref_name);
 
 parse_failed:
 	git_error_set(GIT_ERROR_REFERENCE, "corrupted packed references file");


ethomson:
If NodeGit is relying on this, then probably some other people are as well. So we can fix this. But I'd like libgit2 to stop unnecessarily setting error messages for things. It's very noisy in the debugger, and it's unnecessary formatting, allocation, etc. We should set an error message when we return `-1`. `GIT_ENOTFOUND`, for example, doesn't need an error message because the error return value tells you what the problem was.

(This also means that we need to be more thoughtful about propagating error codes. If merge - for example - encounters a `GIT_ENOTFOUND` that stops it from doing a merge, then that should be promoted to a true `-1` error.)

Anyway, this is all sort of future-thinking. My point is that maybe NodeGit shouldn't be testing error strings that libgit2 returns. It feels brittle. 🤔

PR #6309, opened 2022-05-20
midx: fix large offset table check.

ccstolley:
It's insufficient to only check if the offset high order bit is set, as some object offsets may have their high order bit set but still not use the large offsets table.

We must also check to see if object_large_offsets are in use.

This can cause libgit2 to falsely claim objects do not exist when they actually do, because it is unable to find the object in the index.

cc: @lhchavez as the original author of MIDX support.


ethomson:
Thanks for the fix!

PR #6321, opened 2022-06-14
repo: allow administrator to own the configuration

### LINKED ISSUE #6279: Ownership check on windows seems too strict
### Reproduction steps

When a process is run with "Run as Administrator" or some similar non-UAC mode, then libgit2 is unable to open repositories it creates. This is because the files are owned by Administrator, but the current user is still the local user. The check [here](https://github.com/libgit2/libgit2/blob/63970244fb2d49794e7b5d268a2defe4299fd3ad/src/util/fs_path.c#L1882-L1896) will then fail.

I think this ownership validation is a little too strict simply for *opening* a repository. The git-for-windows CLI will also allow opening repositories owned by Administrator if the user account is also a member of Administrators (see [is_path_owned_by_current_sid](https://github.com/git-for-windows/git/blob/56cb39bb4a8b6afb583e61198aa4108c182631b4/compat/mingw.c#L3524-L3532)).

This comes up on GitHub Actions when running something with the default Powershell.  It seems to be run without UAC and will create files owned by `BUILTIN\Administrators` but the user SID is `runneradmin`.

Steps:
1. Run a process as Administrator
2. Run a process that uses libgit2 to initialize a new repository, and then open the repository.

### Expected behavior

Allow to open a repository owned by Administrators group when the current user is in the Administrators group.

### Actual behavior

Fails with "is not owned by current user" error.

63970244fb2d49794e7b5d268a2defe4299fd3ad

Windows 10 Version 2004
Windows 10.0.20348 (Windows Server 2022 Datacenter)

ethomson:
Update our ownership checks that were introduced in libgit2 v1.4.3
(to combat CVE 2022-24765). These were not compatible with git's; git
itself allows administrators to own the path. Our checks now match
this behavior.

Fixes #6279

### DIFF
--- a/src/libgit2/repository.c
+++ b/src/libgit2/repository.c
@@ -512,7 +512,7 @@ static int validate_ownership(const char *repo_path)
 	bool is_safe;
 	int error;
 
-	if ((error = git_fs_path_owner_is_current_user(&is_safe, repo_path)) < 0) {
+	if ((error = git_fs_path_owner_is_system_or_current_user(&is_safe, repo_path)) < 0) {
 		if (error == GIT_ENOTFOUND)
 			error = 0;
 
--- a/tests/libgit2/repo/open.c
+++ b/tests/libgit2/repo/open.c
@@ -484,9 +484,10 @@ void test_repo_open__validates_dir_ownership(void)
 	cl_git_pass(git_repository_open(&repo, "empty_standard_repo"));
 	git_repository_free(repo);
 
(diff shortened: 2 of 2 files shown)


fxcoudert:
Will this be backported to the 1.4 branch please?

ethomson:
Yes. This will be in the next 1.4 release. 🔜

ethomson:
Yes. This will be in the next 1.4 release. 🔜

Issue #4725, opened 2018-07-16
Stashing specific files


Uncommon:
I have sometimes wished for the ability to stash specific files, and I just had a user request for this as well.

There is the `GIT_STASH_KEEP_INDEX` option, which is one way to achieve this. Even better, though, would be the ability to pass either a list of files or a file filter function. I think the latter might be better since it's more powerful.

I'd probably be able to tackle this myself, so any feedback on appropriate API design and such would be great.

tiennou:
If I were to add something like that, I'd do something like this :

```c
typedef struct git_stash_save_opts {
   int version;
   const git_signature *stasher; /*< can be null, default to repo's signature */
   const char *msg; /*< can be null */
   int flags;
   git_strarray pathspec;
}

/*  standard libgit2 option boilerplate */

GIT_EXTERN(int) git_stash_save_with_opts(git_oid *out, git_repository *repo, const git_stash_save_opts *opts);
```

Note that, from a cursory glance, diff seems to be the underlying mechanism used for choosing what to stash, and it seems to handle [pathspecs](https://github.com/libgit2/libgit2/blob/master/include/git2/diff.h#L415). Though there's a bunch of callback just below, I don't know if they permit that kind of filtering, ie. they seem to be "FYI" callbacks, so maybe you'll need to do pathspecs anyways…

A possible alternative might be to allow stashing indexes directly (if that makes sense ?). That would make users go though the index API (which already handles files however you want), and be somewhat more "generic" (ie. that might enable bare-repo stashing 😉). That's from the top of my head, though, so it might not be viable…

extrawurst:
This would be super useful! @tiennou / @Uncommon are you still planning to do this?

Uncommon:
I'm still interested in having this at some point, although I have no timeline for it.

extrawurst:
ideally we can just create a stash from a bunch of patches, to allow stashing the whole index, a whole file, paths or whatever combination.

JWess:
GitKraken is built on libgit2, and it has a [partial stash feature request](https://feedback.gitkraken.com/suggestions/194826/support-partial-stash) with 223 upvotes that is being blocked. As a GitKraken user, I would LOVE to see this resolved.

PibePlayer:
I really need this for my workflow. Please, someone :)

PR #6330, opened 2022-06-21
stash: partial stash specific files

gitkraken-jacobw:
There's this concept called partial stashing in Git where users are able to stash desired files without having to stash and also reset the entire workdir. I based my implementation on tiennou's suggestion from #4725. This introduces options to `git_stash_save` where developers are able to supply the paths that they want to have stashed from the workdir. I could have gone further and have exposed a `git_index*` option  as tiennou suggested, but I chose not to at this moment. Let me know if I should expose a `git_index*` option so that users can supply an arbitrary index to be stashed.

I also introduced `GIT_STASH_KEEP_ALL`, which prevents resetting the index and workdir once a stash has been completed.

### DIFF
--- a/src/libgit2/stash.c
+++ b/src/libgit2/stash.c
@@ -196,7 +196,7 @@ static int stash_to_index(
 static int stash_update_index_from_paths(
 	git_repository *repo,
 	git_index *index,
-	git_strarray *paths)
+	const git_strarray *paths)
 {
 	unsigned int status_flags;
 	size_t i, error = 0;
@@ -420,7 +420,7 @@ static int build_stash_commit_from_tree(
 	git_commit *i_commit,
 	git_commit *b_commit,
 	git_commit *u_commit,
-	git_tree *tree)
+	const git_tree *tree)
 {
 	const git_commit *parents[] = {	NULL, NULL, NULL };
 
@@ -613,7 +613,7 @@ static int has_changes_cb(const char *path, unsigned int status, void *payload) 
 static int ensure_there_are_changes_to_stash_paths(
 	git_repository *repo,
 	uint32_t flags,
-	git_strarray *paths)
+	const git_strarray *paths)
 {
 	int error;
 	git_status_options opts = GIT_STATUS_OPTIONS_INIT;
@@ -674,7 +674,7 @@ int git_stash_save(
 }
 
 int git_stash_save_with_opts(
-	git_oid *out, git_repository *repo, git_stash_save_options *opts)
+	git_oid *out, git_repository *repo, const git_stash_save_options *opts)
 {
 	git_index *index = NULL, *paths_index = NULL;
 	git_commit *b_commit = NULL, *i_commit = NULL, *u_commit = NULL;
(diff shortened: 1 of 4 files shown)


extrawurst:
this would be so nice to have! thanks @gitkraken-jacobw for providing this!

ethomson:
Sorry about the long delay here - I'm trying to get through some of the backlog of pull requests.

In general, this is great. Following up on your question:

> @ethomson Quick question: the Git CLI introduced a [new paradigm for stashing](https://git-scm.com/docs/git-stash#Documentation/git-stash.txt-push-p--patch-S--staged-k--no-keep-index-u--include-untracked-a--all-q--quiet-m--messageltmessagegt--pathspec-from-fileltfilegt--pathspec-file-nul--ltpathspecgt82308203) a while back via the git stash push command. It looks like libgit2 is still using the old and now-deprecated git stash save paradigm. Are there any plans on changing libgit2's save paradigm over to the push paradigm, especially since git stash push supports a pathspec sort of like how this PR does?

I'm not super familiar with the new stash push. What are the other differences are there between `push` and `save` besides pathspecs?

Concretely, should `git_stash_save_with_opts` instead be called `git_stash_push`? 🤔

extrawurst:
I Hope it’s just a naming thing. Being able to support partial stashing would be huuuge for all apps being based on libgit2 😍

ethomson (dismissed):
This PR looks very good -- there are a few formatting issues that I'll take care of. Appreciate the new functionality and the useful tests. 🙏

src/libgit2/stash.c
```diff
@@ -193,6 +193,30 @@ static int stash_to_index(
 	return git_index_add(index, &entry);
 }
 
+static int stash_update_index_from_paths(
+	git_repository *repo,
+	git_index *index,
+	const git_strarray *paths)
+{
+	unsigned int status_flags;
+	size_t i, error = 0;
```
ethomson, line 202: `error` should be an `int` here.

src/libgit2/stash.c
```diff
@@ -520,6 +599,50 @@ static int ensure_there_are_changes_to_stash(git_repository *repo, uint32_t flag
...
+	git_repository *repo,
+	uint32_t flags,
+	const git_strarray *paths)
+{
+	int error;
+	git_status_options opts = GIT_STATUS_OPTIONS_INIT;
+
+	opts.show  = GIT_STATUS_SHOW_INDEX_AND_WORKDIR;
+	opts.flags = GIT_STATUS_OPT_EXCLUDE_SUBMODULES
+		| GIT_STATUS_OPT_INCLUDE_UNMODIFIED
```
ethomson, line 623: operators should be at the end of the line, not the start of the next line.

src/libgit2/stash.c
```diff
@@ -592,7 +760,13 @@ int git_stash_save(
 	git_commit_free(i_commit);
 	git_commit_free(b_commit);
 	git_commit_free(u_commit);
-	git_index_free(index);
+	git_tree_free(tree);
+	
+	if (has_paths) {
+		git_reference_free(head);
+		git_index_free(index);
```
ethomson, line 767: This looks like it's being loaded regardless of `has_paths`

PR #6333, opened 2022-06-27
cmake: drop posix dependency from pcre* detection

jpalus:
since f585b12 neither PCRE nor PCRE2 backends rely on POSIX regex, so
reflect it in library detection logic

### DIFF
--- a/cmake/FindPCRE.cmake
+++ b/cmake/FindPCRE.cmake
@@ -16,19 +16,18 @@
 # PCRE_FOUND	- True if pcre found.
 
 # Look for the header file.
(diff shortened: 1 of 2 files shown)


ethomson:
This seems correct to me. But do we need to correct the `pcre.h` / `pcreposix.h` detection in `FindPCRE.cmake` as well? 🤔

jpalus:
Yes PCRE detection needs adjustment as well: both include file and linked libraries. Let me handle it too.

jpalus:
Updated PR with PCRE changes.

PR #6340, opened 2022-07-04
ntlmclient: LibreSSL 3.5 removed HMAC_CTX_cleanup

vishwin:
https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.5.0-relnotes.txt

Modify guard to declare dummy function.


ethomson:
Thanks for the fix!

PR #6345, opened 2022-07-07
Fixes #6344: git_branch_move now renames the reflog instead of deleting.

### LINKED ISSUE #6344: git_branch_move always deletes the reflog instead of renaming it
### Reproduction steps
1. For a branch that has an existing reflog, call `git_branch_move` to rename the branch.
2. Notice the reflog was deleted and a new one created, with only a single line recording the rename.

### Expected behavior
The reflog should be renamed and the rename record appended.

### Actual behavior
The reflog is deleted and re-created.

Main (currently `1a94d97e5f5668d25d677471a81dba384267a428` at the time of this writing).

macOS 12.4

arroz:


arroz:
Please don't merge yet, may need a tweak in error recovery.

arroz:
Ok, ready to merge is CI and reviewers are happy. :)

src/libgit2/refdb_fs.c
```diff
@@ -1769,26 +1769,35 @@ static int refdb_fs_backend__rename(
 		(error = refdb_fs_backend__lookup(&old, _backend, old_name)) < 0)
 		return error;
 
+	/* Try to rename the refog; it's ok if the old doesn't exist */
+	/* Must be done before calling refdb_fs_backend__delete, otherwise */
```
ethomson, line 1773: Instead of trying to do all the recovery during failures - I wonder if we refactor this a bit to stop calling `refdb_fs_backend__delete`. It almost looks like this was factored in such a way that we could do that. Would the following work:

1. lock the old
2. lock the new
3. create the new
4. rename the reflog
5. delete the old

does this make sense? is this the right ordering if something goes wrong? 🤔

arroz: Give me a few days to look into this. I didn't like much how it ended up, especially because if an error happens during recovering, it will mess up the error class and message that are stored in the thread storage.

I don't know if there's a way to always guarantee that everything ends up in a consistent state, since we have no transactions here. On the bright side, it's likely many of these operations usually either all fail (because the repo has no write permission, for example) or all succeed.

Regarding the refactoring: yeah, it could probably be done like that. When I was writing the patch, I hesitated between the changes in the PR or moving the reflog deletion to `git_refdb_delete`. I'm not entirely sure which are the assumptions behind this code, specifically if the file-system implementation of references (`refdb_fs.c`) should know it has to handle reflog when it does other things like deleting/moving references, or if it should only do what it is told to (read/write reference files), and the knowledge that reflog files also need to have operations performed on them as part of deletion/moving/etc should be on the layer above in `refdb.c` (where `git_refdb_delete` lives).

I am tempted to pick the latter, because a reflog would exist no matter what, regardless of the backing storage of the reference, but I need to read some more code to know if this makes sense. That, however, would require moving the reflog management to `refdb.c` for all functions, to keep things consistent. There are some reflog functions there already (`git_refdb_reflog_read` and `git_refdb_should_write_reflog`) which is also an indication that managing the reflog when moving/deleting branches might be the work of `refdb.c` and not `refdb_fs.c`, since it means `refdb.c` is aware of the existence of reflogs, even if it delegates to the implementation how to actually read or write them.

csware:
cross-ref: issue #6344

arroz:
Sorry for the delay, "a few days" ended up being more than I expected.

I couldn't make the order of the operations exactly like the one suggested by @ethomson since that order causes `test_refs_branches_move__can_move_a_local_branch_to_a_partially_colliding_namespace` to fail, as it would be trying to create/lock a file on a path that already exists, and currently occupied by a directory. So I shuffled things a bit.

Per the previous comment, I don't think there's a way to always guarantee everything is rolled back (especially because the rollback itself could fail), so I went with a best effort. Let me know if this is OK to be merged.

arroz:
Oops hold on.

tiennou:
Just as it maybe-useful to you, #4316 was an attempt at tackling (most? I don't recall 😞) the reflog-losing issues, so there might be something to salvage from it (the testcases, mostly). But it's dated and likely to require a rewrite.

ethomson:
Thanks for the fix!

PR #6350, opened 2022-07-12
zlib: update bundled zlib to v1.2.12

ethomson:
Update zlib to v1.2.12, as prior versions had memory corruption bugs. It is not known that there is a security vulnerability in libgit2 based on these bugs, but we are updating to be cautious.


chcg:
@ethomson In the meantime https://github.com/madler/zlib/releases/tag/v1.2.13 was released with another CVE finding.

PR #6356, opened 2022-07-14
clar: call ftruncate from util_tests instead of libgit2_tests

boretrk:
`ctest -V` doesn't work properly since it tries to test ftruncate from libgit2_tests while it has been moved to util_tests


boretrk:
Ah, util has its own tests, so this one is all wrong. Closing it for now.

PR #6357, opened 2022-07-15
clar: remove ftrunacte from libgit2 tests

boretrk:
The ftruncate test have been moved to util and shouldn't/can't be done from the libgit2 tests anymore


ethomson:
Thanks!

PR #6358, opened 2022-07-17
util/rand: Fix unused variable warning when building without threads

tomasskare:
When building with USE_THREADS=OFF (GIT_THREADS not defined), the
state_lock variable was still defined, but never used, causing a
warning.

### DIFF
--- a/src/util/rand.c
+++ b/src/util/rand.c
@@ -15,7 +15,9 @@ See <http://creativecommons.org/publicdomain/zero/1.0/>. */
 #endif
 
 static uint64_t state[4];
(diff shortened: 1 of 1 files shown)


ethomson:
Yikes - we definitely aim to be warning free. I'm disappointed that our threads=off nightly build missed this -- possibly just because it's an older compiler. This one is a little subtle, it's the `#define` to `git__noop` that drops the argument.

I'd like to avoid that `#ifdef` if possible -- I opened #6432 with an alternate approach.

ethomson:
(Fixed via #6432)

PR #6359, opened 2022-07-17
fix log example

albfan:
- Parse correctly options
- Use correctly repo loaded in general runner.
- Add `--oneline` option


albfan:
is this fix worth. I use log example a lot to test features, and I think others would benefit from an example to traverse a git repo

ethomson (approved):

ethomson:
Thanks!

PR #6361, opened 2022-07-19
clone: don't mix up "http://url" with "http:/url" when figuring out if we should do a local clone

boretrk:
New attempt at solving #5982, this time using `git_net_str_is_url` from util

Essentially it tries to sort out if we should clone locally without asking the filesystem until it is absolutely necessary since linux can be a bit unreliable on paths containing `://` and treat them as `:/`


ethomson:
Thanks, this seems like a reasonable change. Let's take it.

PR #6362, opened 2022-07-20
push: revparse refspec source, so you can push things that are not refs

sven-of-cord:
I want to push a commit by OID to a remote branch. Currently, push parses the refspecs such that the source must be the name of a ref (it uses git_reference_name_to_id to resolve it). This commit changes it so push uses git_revparse_single to resolve the source of the refspec. This allows for OIDs or other revs (e.g. `HEAD~2`) to be pushed.

### DIFF
--- a/src/libgit2/push.c
+++ b/src/libgit2/push.c
@@ -385,11 +385,18 @@ static int calculate_work(git_push *push)
 	git_vector_foreach(&push->specs, i, spec) {
 		if (spec->refspec.src && spec->refspec.src[0]!= '\0') {
 			/* This is a create or update.  Local ref must exist. */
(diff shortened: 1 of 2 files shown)


ethomson:
Nice. I was going to say that this enabled a lot of other patterns _besides_ OID strings but looking at the test, it seems that these are patterns that we should have supported in the first place! 😅 

This looks good to me, but @carlosmn is the fetch and push expert, I'd love him to 👍 this if he has a moment.

sven-of-cord:
Anything I can do to help this get merged? Would be useful to have.

ethomson:
Sorry about the delay -- thanks @sven-of-cord !

PR #6363, opened 2022-07-22
winhttp: support long custom headers

kcsaul:
Support for long custom headers is needed where they're used to supply access tokens for authentication purposes.

In the case of Azure DevOps, the system access token created for a pipeline job is over 1000 characters.  When attempting to clone a repository using libgit2 with a large token like this being passed along via the git_fetch_options::custom_headers, it fails with the error message: 'failed to convert custom header to wide characters: The data area passed to a system call is too small'.

The cause of the error is within winhttp.c, where it reuses the ct buffer, the size of which is currently 100 characters.


kcsaul:
Updated winhttp to use existing function that allocates appropriately sized buffer while converting custom headers to wide characters (instead of expanding/continuing to use fixed length buffer).

ethomson:
Awesome, thanks @kcsaul !

PR #6373, opened 2022-08-08
fix compile on Windows with -DWIN32_LEAN_AND_MEAN

christoph-cullmann:
Ensure the needed wincrypt.h is included.

Without this, rand.c will not compile if -DWIN32_LEAN_AND_MEAN is set in the compile flags.

### DIFF
--- a/src/util/rand.c
+++ b/src/util/rand.c
@@ -14,6 +14,10 @@ See <http://creativecommons.org/publicdomain/zero/1.0/>. */
 # include <sys/random.h>
 #endif
 
+#if defined(GIT_WIN32)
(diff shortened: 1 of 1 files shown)


src/util/rand.c
```diff
@@ -14,6 +14,10 @@ See <http://creativecommons.org/publicdomain/zero/1.0/>. */
 # include <sys/random.h>
 #endif
 
+#if defined(GIT_WIN32)
+#include <wincrypt.h>
```
ethomson, line 18:
```suggestion
# include <wincrypt.h>
```

ethomson:
Thanks for catching this!

PR #6376, opened 2022-08-09
 Define correct off64_t for AIX

bzEq:
`off64_t` is `long long` on AIX, see `/usr/include/sys/types.h` on AIX.


ethomson:
Does AIX not define `int64_t`? I would have expected the final else would have sufficed here, especially since we use `int64_t` in a lot of other places. 🤔

ethomson:
Thanks for the fix!

PR #6377, opened 2022-08-10
Fixes #6365 : Uppercase windows.h include fails build in case-sensitive OS

### LINKED ISSUE #6365: Uppercase windows.h include fails build in case-sensitive OS
https://github.com/libgit2/libgit2/blob/22f382539d78f69aa91721ad62bbd7229750a043/src/cli/opt.c#L26

When cross-compiling v1.5.0 with MinGW-w64 on Linux, I'm running into an error:

```
src/libgit2-1.5.0/src/cli/opt.c:26:11: fatal error: Windows.h: No such file or directory
   26 | # include <Windows.h>
      |           ^~~~~~~~~~~
compilation terminated.
```

It should be all-lowercase `windows.h`.

Vinz2008:


fxcoudert:
Would it make sense to backport this patch to the 1.5 branch?

PR #6378, opened 2022-08-12
experimental.h: allow building without CMake-generated experimental header

csware:
Not all libgit2 users use cmake.

Issue introduced in: #6191

Similar fix for features.h: #4346

Maybe a more generic flag should be introduced.

### DIFF
--- a/include/git2.h
+++ b/include/git2.h
@@ -28,7 +28,9 @@
 #include "git2/diff.h"
 #include "git2/email.h"
 #include "git2/errors.h"
-#include "git2/experimental.h"
+#ifndef LIBGIT2_NO_EXPERIMENTAL_H
+# include "git2/experimental.h"
+#endif
 #include "git2/filter.h"
 #include "git2/global.h"
 #include "git2/graph.h"
--- a/include/git2/oid.h
+++ b/include/git2/oid.h
@@ -9,7 +9,9 @@
 
 #include "common.h"
 #include "types.h"
-#include "experimental.h"
+#ifndef LIBGIT2_NO_EXPERIMENTAL_H
+# include "experimental.h"
+#endif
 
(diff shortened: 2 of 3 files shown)


ethomson:
Definitely we need a solution for this. I think that there's two use cases here:

1. A piece of software (like TortoiseGit) brings libgit2 in as a dependency into their source tree and uses something other than cmake.
2. libgit2 is getting installed onto the system (via dpkg / rpm / chocolatey / etc)

What we don't want to do is install headers that let people `#define` their way into a crash without understanding it. If I were to casually read that header file in `git2.h`, I would conclude that I can `#define LIBGIT2_NO_EXPERIMENTAL_H` in my code to turn off experimental functionality. But these are not runtime configurable, if I was built with experimental sha256 support then that's _on_. There's no turning it off at this stage, and if you `#define LIBGIT2_NO_EXPERIMENTAL_H` in your code and build it then you're going to crash.

🤔 

I wonder if we can ship `experimental.h` in the source tree, so that users in category #1 don't need to worry about its existence, and then cmake will still generate `experimental.h` and move that up in the include path so that it's seen first. This feels a bit clunky - if I'm getting started with libgit2, then I might get confused about why `experimental.h` is ignored. But I'd rather move the pain to libgit2 developers instead of libgit2 users.

I wonder if we can further improve this by getting rid of `experimental.h.in` and making the in-tree `experimental.h` the input to cmake generation. 🤔

csware:
Maybe the library needs to be named differently when experimental features are enabled to avoid unintended replacements of the binaries.

ethomson:
It is.

ethomson:
@csware I implemented what I described above ☝️ - can you try this and let me know if it works for your setup?

PR #6382, opened 2022-08-14
Fix memory leak

csware:
`data` is allocated using `pcre2_match_data_create` but not freed in all code paths


134130 (approved):
LGTM

ethomson:
Good catch, thanks!

Issue #5492, opened 2020-04-17
git_reference_foreach does not honor per worktree refs


csware:
### Reproduction steps

1. Clone a repository and create a separate worktree for it (using `git worktree`).
2. Start bisect on the worktree
3. Retireve all refs for the worktree using `git_reference_foreach`

### Expected behavior

`git_reference_foreach` returns the `bisec/good` and `bisec/bad` refs.

### Actual behavior

`git_reference_foreach` does not return the `bisec/good` and `bisec/bad` refs.
Those refs are only returned in the main repo when bisecting is started there.

The refs `bisect/good` and `bisec/bad` are stored in `.git/worktree/NAME/refs/bisect/...`.

### Version of libgit2 (release number or SHA1)

918a7d19553a9a4181bf1561b8b828a5f82e60e0, still in 1.5.0

### Operating system(s) tested

Win10 1909, x64

csware:
cf. https://github.com/gitster/git/blob/efe3874640e2e58209ddbb3b072f48f6b7094f34/refs.c#L721 and https://github.com/gitster/git/blob/efe3874640e2e58209ddbb3b072f48f6b7094f34/refs/files-backend.c#L218

csware:
@pks-t @ethomson Could you please have a look at this?

csware:
@pks-t @ethomson Any news on this?

JohnAZoidberg:
Seems to continue being an issue: https://gitlab.com/tortoisegit/tortoisegit/-/issues/3897

ethomson:
Fixed via #6387 - thanks again!

PR #6387, opened 2022-08-18
Make refdb_fs (hopefully) fully aware of per worktree refs

csware:
This makes refdb_fs fully aware of per worktree refs (fixes issue #5492).

The code has some duplication but does the trick.

### DIFF
--- a/src/libgit2/refdb_fs.c
+++ b/src/libgit2/refdb_fs.c
@@ -403,7 +403,9 @@ static const char *loose_parse_symbolic(git_str *file_content)
 static bool is_per_worktree_ref(const char *ref_name)
 {
 	return git__prefixcmp(ref_name, "refs/") != 0 ||
-	    git__prefixcmp(ref_name, "refs/bisect/") == 0;
+	       git__prefixcmp(ref_name, "refs/bisect/") == 0 ||
+	       git__prefixcmp(ref_name, "refs/worktree/") == 0 ||
+	       git__prefixcmp(ref_name, "refs/rewritten/") == 0;
 }
 
 static int loose_lookup(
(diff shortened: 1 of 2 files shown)


csware:
@ethomson Any change to et a review?

src/libgit2/refdb_fs.c
```diff
@@ -863,6 +867,49 @@ static int iter_load_loose_paths(refdb_fs_backend *backend, refdb_fs_iter *iter)
 			error = git_vector_insert(&iter->loose, ref_dup);
 	}
 
+	if (!error && git_repository_is_worktree(backend->repo) == 1) {
+		git_iterator_free(fsit);
```
ethomson, line 871: I think from a layering perspective, the filesystem refdb is not the right place for this — that would mean that _every_ reference db would have to handle the per-corktree references themselves. I think that we should hoist the worktree logic into `refs.c` instead. In theory, the reference database should be able to operate without any knowledge of a repository (indeed, should be able to operate without a repository instance at all).

Reference databases should just handle the storage of references, and the `git_reference` family of functions should handle the business logic, including understanding whether there's a worktree or not.

Let me know if this makes sense — I can take a stab at an implementation on the references if you'd like.

csware: Good point.

However, here, both parts are within the filesystem.

ethomson: Not sure that I understand? Are you saying that since worktrees are in the filesystem and the default git-compatible ref database is also on the filesystem, that this is a reasonable separation of concerns?

ethomson: Er, no, you're saying that the worktree reference storage living in the commondir is actually an implementation detail of the git-compatible ref database implementation itself. 🤔 

That's a bit disappointing, but I think that you may be correct given our current database architecture.

ethomson:
Thanks for the fix! I hope to come back and maybe re-think the abstraction around worktree refs databases, so that each refdb backend doesn't need to think about worktrees. In the meantime, this is great. 🎉

Issue #6366, opened 2022-07-27
git_worktree_lookup returns GIT_ERROR instead of GIT_ENOTFOUND for a non-existing worktree


arroz:
### Reproduction steps
Call `git_worktree_lookup` with a name that does not match the internal name of any existing worktree.

### Expected behavior
An error code of `GIT_ENOTFOUND` should be returned.

### Actual behavior
An error code of `GIT_ERROR` is returned instead.

### Version of libgit2 (release number or SHA1)
Current main (22f382539d78f69aa91721ad62bbd7229750a043 at time of writing).

### Operating system(s) tested
macOS but should happen on any OS.

arroz:
Likely cause: this line should return `GIT_ENOTFOUND` instead of -1 (the code for `GIT_ERROR`): https://github.com/libgit2/libgit2/blob/main/src/libgit2/worktree.c#L136

csware:
Fixed using https://github.com/libgit2/libgit2/pull/6395.

PR #6395, opened 2022-08-31
#6366: When a worktree is missing, return GIT_ENOTFOUND.

arroz:
Rationale: I preferred to not touch the function I mention in #6366 since that check mostly assumes the directory exists and is checking its contents (for corrupted worktrees, etc). So I added an additional check specific to the lookup function. If the directory does not exist, it means the worktree does not exist as well.


ethomson:
Thanks for the fix!

Issue #6182, opened 2022-01-18
Environment Variables for config paths are not considered


michha:
### Reproduction steps
- set `GIT_CONFIG_GLOBAL` to `C:\work\.gitconfig`

### Expected behavior
- git_config_find_global should respect the environment variables

### Operating system(s) tested
- Windows

### What else to say
- I use the [git-tfs](https://github.com/git-tfs/git-tfs/) tool that is using the [LibGit2Sharp](https://github.com/libgit2/libgit2sharp) Library, but I am confident, that the underlying problem is in this repo. Feel free to correct me 😄 
- https://git-scm.com/docs/git-config#_environment - Documentation about the environment variables
- Probably `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_NOSYSTEM` should also be considered in future versions

michha:
Any comment on this issue?

ethomson:
It needs a motivated hacker to send a pull request. I'm happy to talk about the structure of the changes and the best path forward with someone who wants to tackle this. Do you want to work on this?

clnoll:
Hi @ethomson, I've opened a [PR](https://github.com/libgit2/libgit2/pull/6399) that should fix this but would like feedback on the approach when you've got time.

PR #6399, opened 2022-09-07
Work In Progress: load config from environment variables if present

clnoll:
This is a first attempt at having libgit2 use environment variables for config paths, to address https://github.com/libgit2/libgit2/issues/6182.

This PR modifies the original code to read the config at the path specified by the `GIT_CONFIG_GLOBAL` environment variable, and to avoid reading the `$HOME/.gitconfig` and `$XDG_CONFIG_HOME/git/config` files if `GIT_CONFIG_GLOBAL` is set, as specified in the [git-scm](https://git-scm.com/docs/git#Documentation/git.txt-codeGITCONFIGGLOBALcode) documentation.

In addition to unit tests, I've created a minimal example to show that this code change indeed picks up the value from the environment variable: https://github.com/clnoll/libgit2-config-env-var. It can be run by:
```
git clone https://github.com/clnoll/libgit2-config-env-var.git
cd libgit2-config-env-var
cargo build --release
GIT_CONFIG_GLOBAL=.gitconfig_test_env_var ./target/release/libgit2-config-env-var  # prints "In custom config section" to stdout
./target/release/libgit2-config-env-var  # prints "Error: config value 'myconfigsection.custom' was not found" to stdout
```
Output: 
image

If this implementation looks good I can also add handling for `GIT_CONFIG_SYSTEM` and `GIT_CONFIG_NOSYSTEM` following a similar pattern.

### DIFF
--- a/src/libgit2/config.c
+++ b/src/libgit2/config.c
@@ -1145,6 +1145,24 @@ int git_config__find_xdg(git_str *path)
 	return git_sysdir_find_xdg_file(path, GIT_CONFIG_FILENAME_XDG);
 }
 
+int git_config__find_global_and_xdg(
+	git_str *global_buf,
+	git_str *xdg_buf) {
+	int error;
+	error = git__getenv(global_buf, "GIT_CONFIG_GLOBAL");
+
+	/* Only read the $HOME/.gitconfig or $XDG_CONFIG_HOME/git/config files if
+	GIT_CONFIG_GLOBAL is not set, per
+	https://git-scm.com/docs/git#Documentation/git.txt-codeGITCONFIGGLOBALcode */
+	if (error == GIT_ENOTFOUND) {
+		git_error_clear();
+		git_config__find_global(global_buf);
+		git_config__find_xdg(xdg_buf);
+		return 0;
+	}
+	return error;
+}
+
 int git_config_find_system(git_buf *path)
 {
 	GIT_BUF_WRAP_PRIVATE(path, git_sysdir_find_system_file, GIT_CONFIG_FILENAME_SYSTEM);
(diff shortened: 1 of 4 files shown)


src/libgit2/config.c
```diff
@@ -1145,6 +1145,24 @@ int git_config__find_xdg(git_str *path)
...
+int git_config__find_global_and_xdg(
+	git_str *global_buf,
+	git_str *xdg_buf) {
+	int error;
+	error = git__getenv(global_buf, "GIT_CONFIG_GLOBAL");
+
+	/* Only read the $HOME/.gitconfig or $XDG_CONFIG_HOME/git/config files if
+	GIT_CONFIG_GLOBAL is not set, per
+	https://git-scm.com/docs/git#Documentation/git.txt-codeGITCONFIGGLOBALcode */
+	if (error == GIT_ENOTFOUND) {
```
clnoll, line 1157: TODO: do we need different handling for other error cases?

dandavison:
Hi @ethomson, I'm the author of [Delta](https://github.com/dandavison/delta), which uses libgit2 via the Rust bindings, and I believe that this PR would be useful for delta at least, e.g. https://github.com/dandavison/delta/issues/1301 https://github.com/dandavison/delta/issues/1113. If you find time, I'd love to know what your thoughts are on this PR / config env var support in libgit2.

ethomson:
Hi @dandavison - thanks for the PR - I pulled in support for environment variables that specify configuration paths in in during some other refactoring in #6544

ethomson:
Sorry to duplicate the effort 😞 - I needed to update the index loading when environment variables were set, and I remembered #6182 was a problem when I was doing it, but I had forgotten that there was a PR open to add support. Thanks so much - again - for the PR.

PR #6405, opened 2022-09-19
Support non-cmake builds with an in-tree `experimental.h`

ethomson:
Provide an empty `include/git2/experimental.h` file for users who do not build with cmake. When building with cmake, we will generate the `experimental.h`, prefer it during builds, and install it with `cmake --target install`.

A different implementation to solve #6378


ethomson:
Gah, no. Of course this doesn't work because `#include "experimental.h"` will prefer the one in the current directory. 🤔

csware:
I think this will do the trick.

ethomson:
There we go, we supplant the empty `experimental.h` with compile definitions. Rather gross, but should be effective. 😬

ethomson:
Cool, thanks @csware!

PR #6408, opened 2022-10-01
Ignore missing 'safe.directory' config during ownership checks

kcsaul:
Fixes an issue identified where a not found error is returned from the repo ownership validation when 'safe.directory' isn't configured, rather than the more meaningful ownership error (GIT_EOWNER).


ethomson:
Thanks @kcsaul (and @ehuss -- sorry for the delay that caused you to open a similar PR.)

PR #6416, opened 2022-10-20
http: Update httpclient options when reusing an existing connection.

slackner:
Httpclient internally stores a copy of the `certificate_check` callback and payload. When connecting via HTTPS, and if the server sends back `"Connection: close"` after the first request, the following request would attempt to re-use the httpclient and call the (now outdated) callback. In particular for pygit2 this is a problem, since callbacks / payloads are only valid for the duration of a libgit2 call, leading to a `ffi.from_handle()` error and crashing the Python interpreter.

The error looks like this:
> Fatal Python error: b_from_handle: ffi.from_handle() detected that the address passed points to garbage. If it is really the result of ffi.new_handle(), then the Python object has already been garbage collected

Possibly related to https://github.com/libgit2/pygit2/issues/1001.


ethomson:
Makes sense - thanks for the PR!

PR #6419, opened 2022-11-02
chunk-format: create an API from common code

derrickstolee:
The commit-graph and multi-pack-index file formats use common substructures. This is so much the case that commit_graph.c and midx.c have some common code that can be extracted into its own API.

Start by pulling out the write_offset() and write_chunk_header() methods. Later changes will update the API with the reading half of the API.

There are similar changes in the Git project in git/git:chunk-format.[c|h] that could be used as an example for future abstractions, such as populating a struct with the full list of chunks, then letting the chunk API write each of the chunks using callbacks to the specifics within each chunk.

This is a super-simple refactoring. There are more complicated things we could do if we want to mimic [the full chunk-format API in `git/git`](https://github.com/git/git/compare/5a3b130cad0d5c770f766e3af6d32b41766374c0...c4ff24bbb354377a6a7937744fbbef2898243fc7), which has been helpful when quickly prototyping new file formats. I wanted to keep my first contribution small so I could learn about contributing to this community.

### DIFF
--- a/src/libgit2/chunk_format.c
+++ b/src/libgit2/chunk_format.c
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) the libgit2 contributors. All rights reserved.
+ *
+ * This file is part of libgit2, distributed under the GNU GPL v2 with
+ * a Linking Exception. For full terms see the included COPYING file.
+ */
+
+#include "chunk_format.h"
+
+int write_offset(off64_t offset, chunk_format_write_cb write_cb, void *cb_data)
+{
+	int error;
+	uint32_t word;
+
+	word = htonl((uint32_t)((offset >> 32) & 0xffffffffu));
+	error = write_cb((const char *)&word, sizeof(word), cb_data);
+	if (error < 0)
+		return error;
+	word = htonl((uint32_t)((offset >> 0) & 0xffffffffu));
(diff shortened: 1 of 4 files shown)


ethomson:
Hey @derrickstolee -- thanks for the PR, and apologies for taking so long -- I've been a bit underwater this fall. 😢 

Just a couple of stylistic notes:

1. Every function that is `extern` should be prefixed with `git_`, even if it's not publicly exported. (Helper functions that only have file scope should not be `git_` prefixed.)
2. Function names should include the "area" or "object" followed by the action. They're underscore-separated with increasing specificity. In other words, `git_object_write` or `git_object_read` is preferred over `git_write_object` and `git_read_object`. Perhaps I spent too long writing Java, but I think about this as the `write` and `read` methods on a `git_object` struct. Concretely, a function that writes a chunk header is probably best named `git_chunk_header_write` or `git_chunk_write_header`. (Definitely the former if you have a `git_chunk_header` struct.)
3. Similarly, filenames should match up nicely with the area or object name. In other words, `chunk.c` should contain `git_chunk_...` functions.

derrickstolee:
I haven't gotten around to this, and it might be less important than I thought. Closing for now and I might revisit if the need arises.

PR #6420, opened 2022-11-03
commit-graph: only verify csum on git_commit_graph_open().

derrickstolee:
It is expensive to compute the sha1 of the entire commit-graph file each time we open it. Git only does this if it is re-writing the file.

This patch will only verify the checksum when calling the external API git_commit_graph_open(), which explicitly says it opens and verifies the commit graph in the documentation.

For internal library calls, we call git_commit_graph_get_file(), which mmaps the commit-graph file in read-only mode. Therefore it is safe to skip the validation check there.

Tests were added to check that the validation works in the happy path, and prevents us from opening the file when validation fails.

(Note from Derrick Stolee: This patch was applied internally at GitHub after we recognized the performance impact it had during an upgrade of libgit2. The original author left the company before we remembered to send it upstream.)

cc/ @ccstolley 
cc/ @carlosmn

### DIFF
--- a/src/libgit2/commit_graph.c
+++ b/src/libgit2/commit_graph.c
@@ -201,7 +201,6 @@ int git_commit_graph_file_parse(
 	struct git_commit_graph_chunk *last_chunk;
 	uint32_t i;
 	off64_t last_chunk_offset, chunk_offset, trailer_offset;
-	unsigned char checksum[GIT_HASH_SHA1_SIZE];
 	size_t checksum_size;
 	int error;
 	struct git_commit_graph_chunk chunk_oid_fanout = {0}, chunk_oid_lookup = {0},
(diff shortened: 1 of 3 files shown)


lhchavez (dismissed):
oh yikes, should have paid more attention to that case. thanks for the fix!

ccstolley:
Thanks for getting this upstream @derrickstolee :heart:

derrickstolee:
Oops. Working on the build failures.

src/libgit2/commit_graph.c
```diff
@@ -234,11 +233,6 @@ int git_commit_graph_file_parse(
 		return commit_graph_error("wrong commit-graph size");
 	memcpy(file->checksum, (data + trailer_offset), checksum_size);
 
-	if (git_hash_buf(checksum, data, (size_t)trailer_offset, GIT_HASH_ALGORITHM_SHA1) < 0)
```
lhchavez, line 237: huh, turns out that removing this code causes UBSan to fail: https://github.com/libgit2/libgit2/actions/runs/3387366257/jobs/5628155915#step:7:1685

which means that we also need to make the  arithmetic below a bit more carefully so that all operations happen on `uint_64`s and cast to `off64_t` only at the very end.

derrickstolee: Thanks for the assistance! I'll be offline until Monday, but will tackle this first thing.

lhchavez (approved):
don't have merging privileges, but this looks good. thanks!

ethomson:
Sorry for the delay, I've been busy at home and heads down at work. Thanks @derrickstolee ! :shipit:

PR #6421, opened 2022-11-03
Fix leak in git_tag_create_from_buffer

julianmesa-gitkraken:
If the tag already exists and we are not forcing overwrite we need to free ref_name buffer before return the "tag already exists" error.


lhchavez (approved):
don't have merging privileges, but looks good!

PR #6427, opened 2022-11-12
Set GIT_EOWNER when safe.directory is not configured

ehuss:
When there is an ownership violation, and `safe.directory` is not configured, libgit2 will return a somewhat confusing GIT_ENOTFOUND message with `config value 'safe.directory' was not found`. I'm not sure if this was intentional, but I think it is helpful to instead return GIT_EOWNER to indicate the actual underlying error.

### DIFF
--- a/src/libgit2/repository.c
+++ b/src/libgit2/repository.c
@@ -580,9 +580,13 @@ static int validate_ownership(git_repository *repo)
 			break;
 	}
 
-	if (is_safe ||
-	    (error = validate_ownership_config(&is_safe, validation_paths[0])) < 0)
+	if (is_safe)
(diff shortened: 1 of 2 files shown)


kcsaul:
@ehuss Looks like a fix for the same issue I've submitted #6408 for, which is still pending review.

ehuss:
Oh my, thanks for the heads up! I searched the issue tracker, but I often forget to also search open PRs.  I'll go ahead and close.

ethomson:
Sorry, that's my bad. I've been both busy at home and heads down on work for the past few weeks. I'll 👀 this 🔜. Many apologies for the delay.

PR #6435, opened 2022-11-25
transport: fix capabilities calculation

russell:
This looks like a typo to me, from what i can see these lines were added at the same time and because of how capabilities are calculated, it's likely that this code will work in situations where these capabilities were the last ones.


ethomson:
Yikes. Thanks for the fix.

PR #6436, opened 2022-11-27
Initial bazel support

zaucy:
This PR adds [bazel](https://bazel.build/) with bzlmod support which enables bazel users to consume libgit2 as a [bazel module](https://bazel.build/build/bzlmod). I've also submitting these changes to the bazel central registry (BCR): https://github.com/bazelbuild/bazel-central-registry/pull/298

It would be great to have this merged in, but it's also not _required_ in order for it to be submitted to the BCR. It is _recommended_, however.

NOTE: These BUILD files don't have every build option available in libgit2 without bazel.


.bazelrc
```diff
@@ -0,0 +1 @@
+common --experimental_enable_bzlmod
```
meteorcloudy, line 1: If you are using Bazel 6.0 or later, please just use `--enable_bzlmod` since it's no longer experimental.

meteorcloudy (approved):

.bazelrc
alexeagle: there should be a `.bazelversion` file added to this repo as well, or you'll never be able to reason about "if you are using Bazel 6.0"

ethomson:
Sorry this has languished. I know nothing about bezel and can provide no feedback.

It seems like this leads people a bit into a disappointment — for example, it _appears to_ compile in the simple SHA256 support instead of using (eg) OpenSSL. That's the absolute worst-case scenario fallback code. Using it is going to be a _huge_ perf hit.

And I'm never going to have any idea how / when I break it.

I'm not opposed to letting bazel users consume libgit2 easier - in fact, that sounds great - but it needs a maintainer and I'm too ignorant for it to be me.

ethomson:
I'd be happy to merge Bazel support if someone is going to maintain it — but I don't want Bazel support that leads people into a bad situation. I'm closing this but I'd be happy to reconsider if there's an option that (eg) adds SHA256 support but uses a proper SHA256 implementation.

zaucy:
understandable. I had gone off this trail to add official support a while ago. I might pick it up again with a more complete setup as you rightfully criticized. If anyone picks this up let me know and I'd be happy to collaborate. :)

PR #6437, opened 2022-11-28
Rename bundled xdiff symbols to avoid conflicts

georgthegreat:
At the time both mercurial and libgit2 projects bundle xdiff.
When they are linked statically, a duplicated symbol error is raised by the linker.

Rename these symbols to avoid conflicts.

### DIFF
--- a/src/libgit2/blame_git.c
+++ b/src/libgit2/blame_git.c
@@ -385,7 +385,7 @@ static int diff_hunks(mmfile_t file_a, mmfile_t file_b, void *cb_data, git_blame
 		return -1;
 	}
 
-	return xdl_diff(&file_a, &file_b, &xpp, &xecfg, &ecb);
+	return git_xdl_diff(&file_a, &file_b, &xpp, &xecfg, &ecb);
 }
 
 static void fill_origin_blob(git_blame__origin *o, mmfile_t *file)
--- a/src/libgit2/diff_xdiff.c
+++ b/src/libgit2/diff_xdiff.c
@@ -223,7 +223,7 @@ static int git_xdiff(git_patch_generated_output *output, git_patch_generated *pa
 	    git_patch_generated_new_data(&info.xd_new_data.ptr, &info.xd_new_data.size, patch) < 0)
 		return -1;
 
-	xdl_diff(&info.xd_old_data, &info.xd_new_data,
+	git_xdl_diff(&info.xd_old_data, &info.xd_new_data,
 		&xo->params, &xo->config, &xo->callback);
 
 	git_diff_find_context_clear(&findctxt);
--- a/src/libgit2/merge_file.c
+++ b/src/libgit2/merge_file.c
@@ -77,7 +77,7 @@ static int merge_file__xdiff(
 	mmbuffer_t mmbuffer;
 	git_merge_file_options options = GIT_MERGE_FILE_OPTIONS_INIT;
 	const char *path;
-	int xdl_result;
+	int git_xdl_result;
 	int error = 0;
 
 	memset(out, 0x0, sizeof(git_merge_file_result));
@@ -141,7 +141,7 @@ static int merge_file__xdiff(
 
 	xmparam.marker_size = options.marker_size;
 
-	if ((xdl_result = xdl_merge(&ancestor_mmfile, &our_mmfile,
+	if ((git_xdl_result = git_xdl_merge(&ancestor_mmfile, &our_mmfile,
 		&their_mmfile, &xmparam, &mmbuffer)) < 0) {
 		git_error_set(GIT_ERROR_MERGE, "failed to merge files");
 		error = -1;
@@ -158,7 +158,7 @@ static int merge_file__xdiff(
 		goto done;
 	}
 
-	out->automergeable = (xdl_result == 0);
+	out->automergeable = (git_xdl_result == 0);
 	out->ptr = (const char *)mmbuffer.ptr;
 	out->len = mmbuffer.size;
 	out->mode = git_merge_file__best_mode(
--- a/src/libgit2/xdiff/git-xdiff.h
+++ b/src/libgit2/xdiff/git-xdiff.h
@@ -27,18 +27,18 @@
 # endif
 #endif
 
-#define xdl_malloc(x) git__malloc(x)
-#define xdl_free(ptr) git__free(ptr)
-#define xdl_realloc(ptr, x) git__realloc(ptr, x)
+#define git_xdl_malloc(x) git__malloc(x)
+#define git_xdl_free(ptr) git__free(ptr)
+#define git_xdl_realloc(ptr, x) git__realloc(ptr, x)
 
 #define XDL_BUG(msg) GIT_ASSERT(msg)
 
-#define xdl_regex_t git_regexp
-#define xdl_regmatch_t git_regmatch
+#define git_xdl_regex_t git_regexp
+#define git_xdl_regmatch_t git_regmatch
 
-GIT_INLINE(int) xdl_regexec_buf(
-	const xdl_regex_t *preg, const char *buf, size_t size,
-	size_t nmatch, xdl_regmatch_t pmatch[], int eflags)
+GIT_INLINE(int) git_xdl_regexec_buf(
+	const git_xdl_regex_t *preg, const char *buf, size_t size,
+	size_t nmatch, git_xdl_regmatch_t pmatch[], int eflags)
 {
 	GIT_UNUSED(preg);
 	GIT_UNUSED(buf);
--- a/src/libgit2/xdiff/xdiff.h
+++ b/src/libgit2/xdiff/xdiff.h
@@ -84,7 +84,7 @@ typedef struct s_xpparam {
 	unsigned long flags;
 
 	/* -I<regex> */
-	xdl_regex_t **ignore_regex;
+	git_xdl_regex_t **ignore_regex;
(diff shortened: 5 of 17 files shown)


ethomson:
I understand the desire here... _but_ this prevents us from being able to easily update the xdiff code based on changes in git itself. There's not a _lot_ going on that bit of the code but, eg [here](https://github.com/libgit2/libgit2/commit/1458fb56efad36a5d231793b63b466d5a07a5488) we updated to be able to support zealous diff3. We also made an effort to make it _easier_ to update in the future by adding an indirection layer for libgit2 / git to share (though I should pick this back up and get it "across the line" in git itself).

georgthegreat:
Maybe there is support with external / system xdiff library?
If there is, I could unbundle / devendor it from both libgit2 and mercurial.

Is there any kind of official upstream for xdiff?

ethomson:
> Is there any kind of official upstream for xdiff?

Right, this is sort of the problem -- http://www.xmailserver.org/xdiff.html is the original source, but the version in git has had many things removed because git didn't use them, and of course now has had several things changed or added.

At this point, both git and libgit2 look to git's version as our upstream. I don't know what hg is using.

If it's using git's, then we could certainly share a common external implementation. I think that the way that it would work best is to just pull the code out of git itself and package it up as a shared library. libgit2 would happily have an option for using an external xdiff or using the builtin (much like we do with zlib).

One caveat here is that since git views this as "their code" and internal code, I wouldn't expect that they would not have a strong view of API compatibility between versions. I think that's probably okay for these purposes but I thought it worth a caution.

I just had a quick glance at hg's source and it seems to be a subset of xdiff as well. Whether it was a subset of the original xdiff or git's, I cannot tell at a quick glance. But if its original provenance was git, then things seems like a possible path forward.

boretrk:
For your particular problem you might be able to use [objcopy](https://www.tutorialspoint.com/unix_commands/objcopy.htm) with --redefine-syms to rename the xdiff-related symbols in one of the libraries.

> --redefine-sym old=new 	Change the name of a symbol old, to new. This can be useful when one is trying link two things together for which you have no source, and there are name collisions.
> --redefine-syms=filename 	Apply --redefine-sym to each symbol pair "old new" listed in the file filename. filename is simply a flat file, with one symbol pair per line. Line comments may be introduced by the hash character. This option may be given more than once.

georgthegreat:
Indeed , there is a bunch of cruthy fixes available.
At the time we compile mercurial with the following CFLAGS:
```
    -Dxdl_bogosqrt=_mercurial_xdl_bogosqrt
    -Dxdl_build_script=_mercurial_xdl_build_script
    -Dxdl_cha_alloc=_mercurial_xdl_cha_alloc
    -Dxdl_cha_free=_mercurial_xdl_cha_free
    -Dxdl_cha_init=_mercurial_xdl_cha_init
    -Dxdl_change_compact=_mercurial_xdl_change_compact
    -Dxdl_diff=_mercurial_xdl_diff
    -Dxdl_do_diff=_mercurial_xdl_do_diff
    -Dxdl_free_env=_mercurial_xdl_free_env
    -Dxdl_free_script=_mercurial_xdl_free_script
    -Dxdl_get_hunk=_mercurial_xdl_get_hunk
    -Dxdl_guess_lines=_mercurial_xdl_guess_lines
    -Dxdl_hash_record=_mercurial_xdl_hash_record
    -Dxdl_hashbits=_mercurial_xdl_hashbits
    -Dxdl_mmfile_first=_mercurial_xdl_mmfile_first
    -Dxdl_mmfile_size=_mercurial_xdl_mmfile_size
    -Dxdl_prepare_env=_mercurial_xdl_prepare_env
    -Dxdl_recmatch=_mercurial_xdl_recmatch
    -Dxdl_recs_cmp=_mercurial_xdl_recs_cmp
```

What I want is to find a more scalable and steady solution.

ethomson:
@georgthegreat if we pulled this out into `deps/xdiff` -- where it arguably should be anyway -- and provided an option to use the "system xdiff" like we do for http-parser, would that be helpful? or would somebody need to create the xdiff redistributable package?

georgthegreat:
> or would somebody need to create the xdiff redistributable package?

We use NixOs / nixpkgs to build our dependencies (this is not 100% accurate, but I suggest to omit some details to simpify stuff). Unfortunately I see no trace of xdiff neither in nixpkgs, not in Ubuntu apt repository.

If there would be "official" xdiff repository hosted by libgit2 (i. e. `github.com/libgit2/xdiff`), we could pull it out as a separate project and then devendor it from both libgit2 and mercurial (I have not talked to mercurial guys yet, nor did I make a thorough comparison of xdiff patches applied by these projects).

Will such approach work for libgit2?

boretrk:
> If there would be "official" xdiff repository hosted by libgit2 (i. e. `github.com/libgit2/xdiff`), we could pull it out as a separate project and then devendor it from both libgit2 and mercurial.

Would this even simplify anything?
If I understood correctly you wanted a solution that doesn't require extra work when the xdiff interface changes, but if xdiff is moved to be a common module between libgit2 and mercurial you instead get a situation where both projects have to use the same version of xdiff or you can't even update one of them.

ethomson:
I think it solves for the problem where you link hg and libgit2 and both of them have xdiff symbols. Sharing a single xdiff solves this (though there are other solutions).

I think that having a single xdiff that git, libgit2, and hg share would be nice.

georgthegreat:
@ethomson, it looks like mercurial project uses mailing list mercurial-devel@mercurial-scm.org for communications.
I have started a conversation there, but I do not know if they have a public web-interface for this list.

ethomson:
Thank you @georgthegreat - you rock. I'm hopeful that I don't need to subscribe 😅 but let me know if I would be helpful and I will join.

ethomson:
Hi @georgthegreat - curious what the response was from the Mercurial team?

georgthegreat:
Sorry, but I got no response from the team.

ethomson:
@georgthegreat I moved xdiff into `deps` in #6482. I'm not going to close this issue since I don't _think_ that my PR solved your problem directly.

georgthegreat:
Moving it will partially solve my initial problem.
Being able to devendor this dependency is helpful.

It would also be nice to move xdiff into another (independencty buildable) repo.
It is possible?

ethomson:
> It would also be nice to move xdiff into another (independencty buildable) repo.

I created this as https://github.com/libgit2/xdiff

georgthegreat:
> I created this as https://github.com/libgit2/xdiff

Thanks.
We will take a look and even upstream some of our patches.

In a due time.

georgthegreat:
Apparently, something does not work or at least does not work properly.

The build result does not contain and libraries (i. e. neither libxdiff.so not libxdiff.a is generated):
```
(dflt) thegeorg@jakku:~/contrib/xdiff@main$ ls CMakeFiles/xdiff.dir
xdiffi.c.o  xemit.c.o  xhistogram.c.o  xmerge.c.o  xpatience.c.o  xprepare.c.o  xutils.c.o
```

@ethomson could you, please, take a look?

georgthegreat:
Sorry for delaying the testing :-(

georgthegreat:
UPD: this make libxdiff buildable for us:
https://github.com/libgit2/xdiff/pull/1/files

georgthegreat:
Even worse.

xdiff can not be compiled on Windows without src/util/regexp.h:

```
In file included from $(SOURCE_ROOT)/contrib/libs/libgit2/src/libgit2/checkout.c:29:
In file included from $(SOURCE_ROOT)/contrib/libs/libgit2/src/libgit2/diff_xdiff.h:13:
In file included from $(SOURCE_ROOT)/contrib/restricted/xdiff/xdiff.h:30:
$(SOURCE_ROOT)/contrib/restricted/xdiff/git-xdiff.h(19,10): fatal error: 'regex.h' file not found
#include <regex.h>
         ^~~~~~~~~
1 error generated.
```

An attempt to build it with musl fails either:
```
In file included from $(SOURCE_ROOT)/contrib/libs/libgit2/src/libgit2/merge_file.c:14:
In file included from $(SOURCE_ROOT)/contrib/libs/libgit2/src/libgit2/diff_xdiff.h:13:
In file included from $(SOURCE_ROOT)/contrib/restricted/xdiff/xdiff.h:30:
$(SOURCE_ROOT)/contrib/restricted/xdiff/git-xdiff.h:60:56: error: use of undeclared identifier 'REG_STARTEND'
    return regexec(preg, buf, nmatch, pmatch, eflags | REG_STARTEND);
                                                       ^
1 error generated.
```

georgthegreat:
At the time, xdiff is neither unbundlable, nor independently buildable as deps/xdiff extensively includes headers from src/util.

Also, deps/xdiff bears LGPL license, hence linking exception is not fully applicable to libgit2.

ethomson:
> At the time, xdiff is neither unbundlable, nor independently buildable as deps/xdiff extensively includes headers from src/util.

It doesn't require anything in `src/util`, per se, it requires a compatible regex implementation. PRs are welcome to change this.

> Also, deps/xdiff bears LGPL license, hence linking exception is not fully applicable to libgit2.

I'm not sure that I fully understand what you're saying here.

georgthegreat:
> I'm not sure that I fully understand what you're saying here.

I believe that GPL-2.0 WITH GCC-exception-2.0 can be used in Apache-2.0 projects, while LGPL code can be used only with dynamic linkage, if I interpret [this text](https://www.apache.org/licenses/GPL-compatibility.html).

Issue #5335, opened 2019-12-12
Support push options


sathieu:
Git supports push options since v2.10.

They can be set on CLI, (`-o`) or using the `push.pushOption` config.

Please add support for this in libgit2 and pygit2.

Refs:
- Git's client-side push options: https://github.com/git/git/commit/f6a4e61fbb647928d615a0befaec163a5d2bf4af
- Gitlab's server-side push options: https://docs.gitlab.com/ee/user/project/push_options.html#push-options

pablonardone:
Was there any progress for this issue?

Issue #6400, opened 2022-09-08
heads up: security issues in git with respect to embedded bare repos


calestyo:
Hey.

This is merely a heads up about a security issue that was found a while ago in git itself and which was discussed in detail at [[0](https://github.com/justinsteven/advisories/blob/main/2022_git_buried_bare_repos_and_fsmonitor_various_abuses.md)] as well as [[1](https://lwn.net/ml/git/kl6lsfqpygsj.fsf@chooglen-macbookpro.roam.corp.google.com/)] with a good overview and simple demo of an exploit at [[2](https://lwn.net/Articles/892755/)] and with a "fix" (i.e. rather a safeguard, that unfortunately even seems to default to the unsafe option) pending for release at [[3](https://github.com/git/git/pull/1261)].

In short, when cloning an untrusted repository and inspecting it (simple git commands or even `cd`ing into them, with things like `git-prompt` enabled, may be enough) allow apparently arbitrary code execution. Code which is controlled by a possible attacker.

Could you please check whether libgit2 is also affected by this issue and if so, provide some fix?

Similarly, it must be checked whether the work tree of a git repo would contain a `.git` dir, which would allow for the same attack with non-bare repos.
Git apparently already forbids this since long.

Thanks for checking,
Chris.

boretrk:
I don't think libgit2 even supports fsmonitor, at least I couldn't find any references to it.

calestyo:
AFAIU, it's no just fsmonitor, but in principle whenever the git implementation would parse/use the embedded bare repos's git-config or e.g. it's hooks.

extrawurst:
Libgit2 does not support hooks

calestyo:
Hmm, well but AFAICS from a just very quick glance on its sources, it **does** parse git’s `config`.
Not sure whether it also does so for embedded bare repos, though.

Even if none of this would be supported by libgit2 right now, it would probably still make sense to somehow take care that no future version of it, which may add support for further features, becomes exploitable. I mean in one year from now, no one will likely remember this subtle but dangerous issue.


And in principle git is also vulnerable to having a rogue `.git`, even if it cannot checked out as part of a repo (not sure whether libgit2 prevents even that).
Consider a tarball which is downloaded from somewhere and contains a `.git`. It's not too hard to imagine that some tool using libgit2 is called within that.
[...]

PR #6439, opened 2022-12-01
Implement push options on push

russell:
This PR implements passing push options on push if the server supports the capability.  Push options are sent after the initial handshake https://github.com/git/git/blob/master/Documentation/gitprotocol-pack.txt#L563

I implemented this because SourceHut uses push-options to skip CI, but after looking at https://github.com/libgit2/libgit2/issues/5335 I see that GitLab has the same feature.  Gitea implemented a small amount of functionality using push-options, but it's not as critical for it.

Unfortunately I wasn't aware of PR https://github.com/libgit2/libgit2/pull/6281 but I did find it before I implemented any tests, so I have rebased and improved the tests suite from that PR.

This PR adds a field to the push_options struct called push_options, this might seem a bit confusing, but push-options are the name of the options in the Git CLI.  Though the name could be shortened to options if that was clearer.

The test suite can only be run on focal or newer Ubuntu's. This is because older versions of the git CLI doesn't support receiving push options (or doesn't advertise the capability). When testing libgit2 against focal, I found that SSH doesn't work, it looks like an incompatibility between libssh2 and openssh, with the logs up to debug3 I was able to spot some differences between the ssh client an libssh2, but i really don't understand ssh well enough to know why the connection is rejected.  I tried upgrading libssh2 to 1.10.0, but that didn't help.

fixes https://github.com/libgit2/libgit2/issues/5335


russell:
Also, push options for local push can't be supported until hooks are implemented for local repos,  i didn't look into this too much.  It should be added separately i think to prevent this PR growing too much

boretrk:
> push options for local push can't be supported until hooks are implemented for local repos

If you look into hooks it could be a good idea to check the security issue mentioned in #6400 to see if it applies.

ethomson:
Thank you so much for pulling this together. This is great work by you and the other contributors. I apologize that it took so long to address this PR, but I love the simplifications that you made.

I merged `main` into this, since the PR had accumulated some conflicts during the time that I spent not reviewing it. I also made some minor tweaks — I changed the name to `remote_push_options` which... is not great, but since we had `push_options` already, I felt like it helped disambiguate. I also changed the CI test behavior a little bit so that we could keep mutable data contained within the sandbox.

Thanks again - I think that this is a nice feature and I appreciate you and everybody else who went above and beyond to add integration tests. 🙏

PR #6443, opened 2023-01-04
Do not register fake repository as owner of ODB

oreiche:
Accessing ODB is guaranteed to be thread-safe. Registering a fake repository (created via git_repository_wrap_odb()) may not register itself as owner of the ODB in order to maintain that guarantee. Otherwise, accessing objects from ODB will try to obtain the cache from the owning repository (via odb_cache()) and produce a race if this ODB is concurrently used to create fake repositories in other threads. Consequently, operations on fake repositories will interact with the ODB's cache instead of the repository's cache.

### DIFF
--- a/include/git2/sys/repository.h
+++ b/include/git2/sys/repository.h
@@ -97,9 +97,10 @@ GIT_EXTERN(int) git_repository_set_config(git_repository *repo, git_config *conf
  *
  * @param repo A repository object
  * @param odb An ODB object
+ * @param set_owner Register repository as owner of ODB
  * @return 0 on success, or an error code
  */
-GIT_EXTERN(int) git_repository_set_odb(git_repository *repo, git_odb *odb);
+GIT_EXTERN(int) git_repository_set_odb(git_repository *repo, git_odb *odb, bool set_owner);
 
 /**
  * Set the Reference Database Backend for this repository
--- a/src/libgit2/repository.c
+++ b/src/libgit2/repository.c
@@ -89,10 +89,12 @@ git_str git_repository__reserved_names_posix[] = {
 };
 size_t git_repository__reserved_names_posix_len = 1;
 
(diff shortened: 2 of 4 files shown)


ethomson:
Sorry for the delay. I've come back to this PR a few times to try to understand what's going on and what you're really trying to solve. More information there would be helpful, because ultimately, it seems like you have an object database and are trying to create multiple fake repositories on top of it. The immediate problem to this approach is that `wrap_odb` sets the odb owner to the new fake repository, which means that trying to call it multiple times means that the odb will always look to the most recently created repository's cache.

_If_ we wanted to try to change this, we should do so without breaking the API of `set_odb`. There's no need for that, we should just change `wrap_odb` to handle the updates itself. For example:

```diff
diff --git a/src/libgit2/repository.c b/src/libgit2/repository.c
index 97f776c4a349..9a6083c399cf 100644
--- a/src/libgit2/repository.c
+++ b/src/libgit2/repository.c
@@ -1184,9 +1184,17 @@ int git_repository__wrap_odb(
 
 	repo->oid_type = oid_type;
 
-	git_repository_set_odb(repo, odb);
-	*out = repo;
+	/*
+	 * Do not GIT_REFCOUNT_OWN this odb; doing so would
+	 * cause the odb to use this repository's object cache
+	 * instead of its own cache. Using the odb cache allows
+	 * multiple fake repositories to be created via `wrap_odb`
+	 * and they will not try to override the odb cache.
+	 */
+	GIT_REFCOUNT_INC(odb);
+	GIT_ASSERT(git_atomic_swap(repo->_odb, odb) == NULL);
 
+	*out = repo;
 	return 0;
 }

```

However, this is only _part_ of the problem. The objects in the cache _themselves_ also have an owner, and the odb cache shouldn't have objects from different repositories inside of it.

So - I'm curious why you can't new up multiple object databases and wrap _those_? Are they in-memory object databases?

PR #6444, opened 2023-01-05
Use `git_clone__submodule` to avoid file checks in workdir

### LINKED ISSUE #5830: git_submodule_update should call git_clone__submodule instead of git_clone to omit the path check
### Reproduction steps

1) Create a repository that has a submodule called "mysubmodule".
2) Change directory to a folder which has a subfolder called "mysubmodule".
3) Clone the main repository and then call git_submodule_update on "mysubmodule".

### Expected behavior

The submodule is initialised and cloned.

### Actual behavior

Unable to clone the submodule with error:
Error code = -4, Error class = 3. "mysubmodule" exists and is not an empty directory.

The path check in git__clone checks the current directory, rather than the repository working directory. If omitting the path check is undesirable, perhaps there's another solution.

1.0.0

Windows

abizjak:
`git_clone` checks for existence of (non-empty) directories that would clash with what is about to be cloned.

This is problematic when cloning submodules since they make sense in the context of a parent module, so they should not use the current working dir.

Since in `git_submodule_update` we clone the submodule only when it is not yet initialized we do not need to perform directory checks.

Fixes #5830

### DIFF
--- a/src/libgit2/submodule.c
+++ b/src/libgit2/submodule.c
@@ -1393,7 +1393,7 @@ int git_submodule_update(git_submodule *sm, int init, git_submodule_update_optio
 		 */
 		clone_options.checkout_opts.checkout_strategy = GIT_CHECKOUT_NONE;
 
-		if ((error = git_clone(&sub_repo, submodule_url, sm->path, &clone_options)) < 0 ||
+		if ((error = git_clone__submodule(&sub_repo, submodule_url, sm->path, &clone_options)) < 0 ||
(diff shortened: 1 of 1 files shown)


lrm29:
Thanks for looking at this it had slipped my mind. I had worked around this by changing directory before cloning... I think it would be nice to lock down with a test case in tests/libgit2/submodule/update.c

ethomson:
Thanks for the PR and the review!

PR #6452, opened 2023-01-20
push: use resolved oid as the source

ethomson:
211c97195e2ebcf68e27782715eb756823ad5a91 attempts to use the parsed OID but inverted the arguments to `git_oid_cpy`.


carlosmn:
I wonder why the test suite didn't seem to catch this. I was able to reproduce this both here and in rugged by using a `file://` URL for push tests.

PR #6455, opened 2023-01-23
Support the notion of a home directory separately from global configuration directory

### LINKED ISSUE #6453: known_hosts fail to load on Windows
Fetching a repo over SSH fails on Windows because it fails to find the home directory. This can happen when launching from Windows Powershell, Cmd, or other environments that do not set HOME.

### Reproduction steps

Fetch an SSH repo from github.com using Windows Powershell or Cmd.

### Expected behavior

libgit2 should be able to find the user's home directory.  There are multiple ways to do that, though I'm not sure of the benefits of different approaches. Checking for USERPROFILE would probably be the easiest. The Powershell port of OpenSSH uses `ProfileImagePath` in the registry, falling back to `GetWindowsDirectoryW` if that fails.  (Another option is `SHGetFolderPathW`, though I'm not sure I would recommend that.)

Also, I think a failure to load or process the known_hosts should not be a hard error if the user has also defined their own `certificate_check_cb` callback. That is, the code [here](https://github.com/libgit2/libgit2/blob/036fe1afba37d9b20509671d3048c6088f018134/src/libgit2/transports/ssh.c#L544-L558) shouldn't be a hard error.

### Actual behavior

The error is:

```
invalid or unknown remote ssh hostkey
```

1.5.1

Windows 10 Pro

ethomson:
Today, we muddle the notion of the global git configuration (which lives in the home directory) and the home directory. This means that -- for example -- when our configuration parsing code tries to include something that is tilde-prefixed, for example `~/.git/other_config`, then we look up the _global configuration path_ to resolve this.

We also have some code that tries to look up the home directory by looking at the `HOME` environment variable because the global configuration would be the wrong place to look for this, and may be reset by our tests, the binding's tests, or the application itself.

This PR breaks the notion of the home directory from the global configuration path. At the moment, they both have the same default lookup mechanisms, but can be configured separately. This allows test frameworks to configure them independently. This is especially useful for things like ssh configuration, which we may want to set up at the beginning of a CI run (based on the ssh server information) and then use within the test execution.

Since this brings the existing Windows home directory lookup logic (which was used for the global configuration path) _to the home directory_ logic. As a result, this fixes #6453.

I tried to structure this in a way that we can cherry-pick only the bits related directly to #6453 to the v1.4 and v1.5 maintenance branches, so that (either) they will simply use the global config path for ssh keys or use a home dir that is not runtime-configurable. (I can see arguments for doing either, but I haven't yet formed a strong opinion. I would like to ensure this actually resolves #6453 before dealing with the backporting.)


ethomson:
I backported this into v1.4 and v1.5. 😅

PR #6456, opened 2023-01-23
SHA256: more SHA256 support

ethomson:
Opening a draft PR with continued work on SHA256 support.


ethomson:
Going to go ahead and 🚢 this since it's behind a feature flag.

PR #6460, opened 2023-01-25
ci: always create test summaries, even on failure

ethomson:
When the dependent jobs fail -- possibly due to test failures -- we should still produce the job summary that shows those test failures.


ethomson:
The tests fail, which is useful, because it proves that this PR is successful:

Screen Shot 2023-01-25 at 11 30 29 AM

Issue #6430, opened 2022-11-20
Add a callback for the safe.directory error handling


csware:
Have a repository with a directory not owned by the current user (or an admin).

The git_repository_open returns GIT_EOWNER and provides a string containing an error message (cf. https://github.com/libgit2/libgit2/commit/af9e00321f06383ccf61f4ca9616716bec0eb2da).

For GUI applications using libgit2 there is no way to get the reported path in an easy way, e.g. for providing an option for adding it to the allow list (is it the gitdir, the worktree, or something else). Git for Windows also performs some normalizations (such as resolving symlinks/mount points/junktions) - this is also necessary for libgit2 (cf. #6379) an d would make application developers reimplement the very same logic.

I suppose a callback is needed to which the path is reported so that it can be used easily for further handling such as allowing.

### Version of libgit2 (release number or SHA1)

1.5.0

ethomson:
Yeah, this is a pain point. Today you'd need to screen-scrape the error message which is obviously disappointing.

In general, this feels like another case of the fact that we need a warnings subsystem that callers can subscribe to. In this case, you should be able to get a structured piece of data that indicates that there was a safe directory error, get the directory in question, and be able to return something that tells libgit2 to continue its handling of the problem (in this case, returning an error) or overriding the problem.

(Maybe even retrying?)

Since these problems can happen at the repository level, this would presumably need to happen globally. Some callback added at the `git_libgits_opts` level.

Let me sketch out an idea and we can see if it works.

ethomson:
@csware see https://github.com/libgit2/libgit2/pull/6485 🤔

csware:
In gerneral, looks good.

Is there a need that the callback can ask to continue or ignore? I acknowledge, that this makes it very flexible, however, also could allow to implement a different logic. Do we need this freedom?

ethomson:
> Maybe a specific struct might be better that using varargs if multiple parameters are needed?! (also in the sense of extensability)

Yeah, that's what was proposed here -- https://github.com/libgit2/libgit2/pull/3722/files -- which I also like. I was hoping that there was an _even simpler_ path forward, rather than creating a bunch of structs. But... I do concede that my PR is simple from _libgit2's point-of-view_ and perhaps not _an end-user's_, while #3722 might be simpler for an end-user. I'll need to play around a bit with both. 😓

csware:
Btw. I just notices that the callback would also need to pass the right path for UNC paths (i.e., prefixed with `prefix)/`)...

PR #6485, opened 2023-02-15
Simple warnings framework

ethomson:
Here's a potential simple warnings framework.

I think that the names 'continue' and 'ignore' are probably rather misleading. But bikeshedding on the naming aside, I think that this will help #6430

### DIFF
--- a/include/git2.h
+++ b/include/git2.h
@@ -69,6 +69,7 @@
 #include "git2/tree.h"
 #include "git2/types.h"
 #include "git2/version.h"
+#include "git2/warning.h"
 #include "git2/worktree.h"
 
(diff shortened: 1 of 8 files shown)


ethomson:
This is the fourth PR to add warnings. 😅 This one is - I think - the simplest, at least for us. We can just define and invoke warnings without a lot of ceremony. This does mean that the user basically has a `switch` statement over their supported warning types.

I think that's okay? With the possible exception of having to deal with a `varargs` style callback. I would _expect_ that any reasonable FFI would need to deal with this. But I'd like to validate this a little bit more with (eg) libgit2sharp.

Issue #6491, opened 2023-02-17
git_repository_open_bare isn't setting oid_type


arroz:
When opening a repository with `git_repository_open_bare`, the `oid_type` isn't set. This causes all the code that relies on it (like loading packed refs) to fail, since `GIT_OID_SHA1` has value of 1, but the memory contents will be 0 if not set to any value.

arroz:
@ethomson I guess the best way to fix this is extract the code that loads config, checks version, loads and sets the oid_type from `git_repository_open_ext` into a separate function and calling it from both loading functions. Let me know if you want a patch (it's broken in main right now) or if you can handle that as part of the ongoing SHA256 effort (thank you for that, btw!).

ethomson:
Oops! I'd welcome a patch. I won't be able to look at this until tomorrow at the earliest.

arroz:
Here it is: https://github.com/libgit2/libgit2/pull/6492

PR #6492, opened 2023-02-17
#6491: Sets oid_type on repos open with git_repository_open_bare

arroz:


ethomson:
Thanks, @arroz !

As I gaze upon the various ways that you can open a repository -- `git_repository_open`, `git_repository_open_ext`, `git_repository_open_from_env`, and `git_repository_open_bare`, I am saddened. It seems super complicated. But even something seemingly simple like making `git_repository_open_bare` work with `open_ext` is nontrivial. 😓

(A problem for another day.)

Thanks again.

arroz:
Yeah, there's a lot of steps, and it's not entirely obvious how to build a common structure. The function I added is not great, it has that feeling of not really fitting, having side effects, etc.

Reading both `open_ext` and `open_bare` line by line, it seems `open_ext` can be as efficient as `open_bare` using the correct flags (ignoring the micro-optimizations of skipping "if" statements, just thinking on the really slow stuff like FS access). For example, there is already `GIT_REPOSITORY_OPEN_NO_SEARCH` to avoid searching, and `GIT_REPOSITORY_OPEN_BARE` to force bare. The advantage of skipping the config is lost with SHA256, since it always needs to be loaded no matter what. There are additional optimizations that probably can be done, like skipping the `is_worktree` check if `GIT_REPOSITORY_OPEN_BARE` is set. So maybe the implementation or `open_bare` can be replaced with a call to `open_ext` setting those flags, without any relevant loss of performance.

Still, this should probably be done in collaboration with someone from the company that is "hosting git repositories and need to access them efficiently" (per the documentation) just to be sure nothing breaks. :)

Issue #6467, opened 2023-02-06
Clarification of license for site using web assembly + static linking?


liquidaty:
Hi,

Thank you for this awesome project. We are big fans and would like to use it more, but have a license question. 

In particular, we'd like to build a website that loads web assembly into its page (compiled from closed source) and makes use of libgit2. Seems straightforward enough.

Ideally, we'd simply develop a proprietary binary that dynamically links to libgit2, and distribute them both together without the requirement of disclosing proprietary source code: a clear-cut case that would appear to squarely fall within the bounds of permissibility as described in [COPYING](https://github.com/libgit2/libgit2/blob/main/COPYING). And to make it easier, others have [already compiled libgit2 to wasm](https://www.npmjs.com/package/wasm-git/v/0.0.4) so we know that's feasible.

Alas, there is no simply way to load dynamic libraries, this due to issues such as [this 4kb limit one](https://github.com/emscripten-core/emscripten/issues/11753). Maybe there is some undocumented workaround-- many have been suggested-- but by far the easiest solution would be to statically link.

So the question is, would that be permissible? I suppose the first question to answer that is: if I host a website that loads web assembly into a user's browser, is that the same as "distribution" (i.e. analogous to distributing a .exe file that is run locally on a PC)? Our understanding is that the whole purpose of the license is to allow distribution so long as the shared library can be swapped out. However, in the context of user accessing the software through a browser, the user cannot make such a swap (since the user cannot control over the execution environment), so the use case would seem to be more analogous to a user who SSHs into a server that contains a single version of a proprietary combined work, which the user can execute (e.g. "/opt/bin/combinedwork") but not modify and is not authorized to copy, which is a use case that would seem to squarely fit in the "not a distribution" category.

Just to clarify, I'm not asking this question for all licenses similar to this library's (though it very well might be applicable). Rather, I'm just asking what the intent is, by the original authors who wrote the code and chose the license, for this particular project. That said, I would still be interested to hear any feedback anyone may have.

neithernut:
Disclaimer: I'm not a maintainer of this project.

Libgit2 is licensed under GPLv2, not LGPLv2. However, `COPYING` does include a linking exception, which is of interest in your case. If I read it correctly it's rather permissive and allows you to link libgit2 _statically_ into a binary and distribute it. (Last time I checked LGPLv2 didn't allow this, but I could be mistaken).

[...]

PR #6494, opened 2023-02-20
README: clarify the linking exception

ethomson:
We say that you can link libgit2 "unmodified"... "without having to release its source code". Clarify that you can modify libgit2 - but you must release _its_ source code back - and you can link libgit2 without having to release _your software's_ source code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,11 @@ and on Git hosting providers like [GitHub](https://github.com/),
 We perform the merge every time you click "merge pull request".
 
 `libgit2` is licensed under a **very permissive license** (GPLv2 with a special
-Linking Exception). The GPLv2 requires you to make changes you make _to libgit2
-itself_ available publicly, but the linking exception means that you can link
-libgit2 with your software without it becoming "infected" by the libgit2
-license. Additionally, the example code has been released to the public
-domain (see the [separate license](examples/COPYING) for more information).
+Linking Exception). This means that you can link against the library with any
+kind of software without making that software fall under the GPL.
+Changes to libgit2 would still be covered under its GPL license.
+Additionally, the example code has been released to the public domain (see the
+[separate license](examples/COPYING) for more information).
 
 Table of Contents
 =================
--- a/README.md
+++ b/README.md
@@ -25,10 +25,11 @@ and on Git hosting providers like [GitHub](https://github.com/),
 We perform the merge every time you click "merge pull request".
 
 `libgit2` is licensed under a **very permissive license** (GPLv2 with a special
-Linking Exception).  This basically means that you can link it (unmodified)
-with any kind of software without having to release its source code.
-Additionally, the example code has been released to the public domain (see the
-[separate license](examples/COPYING) for more information).
+Linking Exception). The GPLv2 requires you to make changes you make _to libgit2
+itself_ available publicly, but the linking exception means that you can link
+libgit2 with your software without it becoming "infected" by the libgit2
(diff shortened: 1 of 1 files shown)


ethomson:
Fixes #6467

README.md
```diff
@@ -25,10 +25,11 @@ and on Git hosting providers like [GitHub](https://github.com/),
 We perform the merge every time you click "merge pull request".
 
 `libgit2` is licensed under a **very permissive license** (GPLv2 with a special
-Linking Exception).  This basically means that you can link it (unmodified)
-with any kind of software without having to release its source code.
-Additionally, the example code has been released to the public domain (see the
-[separate license](examples/COPYING) for more information).
+Linking Exception). The GPLv2 requires you to make changes you make _to libgit2
+itself_ available publicly, but the linking exception means that you can link
```
carlosmn, line 29: The GPL does not require publishing modifications publicly. it requires you to provide source code to those to whom you have provided the compiled version.

carlosmn:
I gather the issue here is the question of whether doing a modification to libgit2 and then linking against that would still allow for the application using said library to remain proprietary (or under some other license). The modification in the linked issue seems to be conversion into webassembly, which to me seems too mechanical to become derivative work (and instead would... just be the same thing, I guess). But I am not a lawyer.

But as a senior software developer, I also get to say "it depends" a lot. The GPL works by placing restrictions on the distribution of "derived works" which is a thing in Copyright law. Ultimately a judge would be one to decide what happens. But the linking exception indicates that the authors do not consider using the library (and thus linking to it) to make a program so intrinsically bound to its functionality that it would cause the calling program to become a derived work. But whether something is a derived work doesn't necessarily hinge on how it's linked, but whether the work is, you know derived.

I'm going to stop myself before I go too deep into analysing what modification and distribution means over the last 30 years, and get back into what seemingly has caused the confusion that prompted this change.

The `(unmodified)` text goes all the way back so it wasn't in response to a particular earlier question (that I can tell). It reads to me to be trying to indicate that you can add libgit2 to your package manger and use it in your app and that's all good. Static or dynamic linking should not make a difference here (if nothing else this is evidenced by bindings from the same authors linking statically and having a different license).

But modifying libgit2 and then linking against that... that's probably fine, but it would depend on what the modifications are. What would it meant to make changes to expose internal mechanisms or utilities and then link against that? Unfortunately you'd have to say "it depends" because we don't know what those hypothetical modifications are. But in general I do agree that the word is more confusing than it should be.

I might reword it to something like

>`libgit2` is licensed under a **very permissive license** (GPLv2 with a special
Linking Exception).  This means that you can link against the library with any
kind of software without making that software fall under the GPL.
Changes to libgit2 would still be covered under its GPL license.
Additionally, the example code has been released to the public domain (see the
[separate license](examples/COPYING) for more information).

README.md
ethomson: True.

ethomson:
> I might reword it to something like
> 
> > libgit2 is licensed under a very permissive license (GPLv2 with a special
Linking Exception). This means that you can link against the library with any
kind of software without making that software fall under the GPL.
Changes to libgit2 would still be covered under its GPL license.
Additionally, the example code has been released to the public domain (see the
[separate license](https://github.com/libgit2/libgit2/pull/examples/COPYING) for more information).

✅  done

liquidaty:
Thank you for making these changes. This is very helpful and greatly appreciated! Am hopeful I will have no more comments to pass on...

PR #6495, opened 2023-02-21
Update README.md

BirdboyBolu:


extrawurst:
its probably good to provide some context of why these changes should be considered

BirdboyBolu:
Hello @extrawurst, I changed a few passive words to action words. The changes were little and I am still looking through more. I wanted to give a description at the end. Thank you

extrawurst:
> I wanted to give a description at the end.

@BirdboyBolu its best practice if its work in progress to mark the PR as *draft* so no one ends up doing unnecessary feedback

carlosmn:
This is more than stylistic changes. There are deeper changes to meaning here.

ethomson:
Indeed. Is it ideal for powering things? Sure, I guess, but that's fluffy and subjective. That it *does* power things is objective and worth celebrating.

If you're interested in iterating on this then I will leave this PR open, but these changes are not an improvement over what's there, as they stand.

PR #6500, opened 2023-02-23
Correct the definition of "empty" in the docs for `git_repository_is_empty`

timrogers:
This improves the documentation for `git_repository_is_empty` which currently does not accurately describe libgit2's definition of "empty".

It says that HEAD must point to the "unborn master branch", when in fact, this is not the case if the repo's `init.defaultBranch` configuration is set. If it is set, it will check that HEAD points there. Only if it is not set does it fall back to `master`.


ethomson:
Thanks @timrogers ! Appreciate you noticing the issue here and correcting it.

PR #6503, opened 2023-02-24
Pass hostkey & port to host verify callback

fxcoudert:
This is a patch that has been applied for 2 years in Julia. Quoting from @StefanKarpinski in his original pull request:

> It seems that no one actually verifies SSH host identity with libgit2 because the callback doesn't give enough information do so correctly:
> 
> * It doesn't give the actual host key fingerprint, but rather three
>   different hashes thereof. This means we cannot distinguish a known
>   hosts entry that has a different type (`ssh-rsa`, `ssh-dsa`, etc.)
>   from an entry with a matching type and a fingerprint mismatch: the
>   former should be treated as an unknown host whereas the latter is a
>   host key mismatch; they cannot be distinguished without this patch.
> * If the user connects on a non-default port (i.e. not 22), this is not
>   passed to the callback in any way. Since there can be different known
>   host entries for different ports and they should be treated as
>   distinct, this also means the current API cannot be used to verify
>   hosts serving SSH on non-standard ports. This patch passes the port.
> 
> I will try to upstream some version of this patch to libgit2.

I have adapted the patch slightly. I do not know libgit2 internals enough to know whether `snprintf` and `alloca` are the best way to handle the memory allocation/string printing in that place.

- Original PR at Julia: https://github.com/JuliaLang/julia/pull/39324
- Example of bug created by the current situation: https://github.com/JuliaLang/julia/issues/38777 “Host verification failure for packages in SSH remotes at a non-standard port”

### DIFF
--- a/src/libgit2/transports/ssh.c
+++ b/src/libgit2/transports/ssh.c
@@ -651,6 +651,8 @@ static int check_against_known_hosts(
 	return ret;
 }
 
+#define SSH_DEFAULT_PORT 22
+
 /*
  * Perform the check for the session's certificate against known hosts if
  * possible and then ask the user if they have a callback.
@@ -748,9 +750,16 @@ static int check_certificate(
 	if (check_cb != NULL) {
 		git_cert_hostkey *cert_ptr = &cert;
 		git_error_state previous_error = {0};
+		const char *host_ptr = host;
+		git_str host_and_port = GIT_STR_INIT;
+
+		if (port != SSH_DEFAULT_PORT) {
+			git_str_printf(&host_and_port, "%s:%d", host, port);
+			host_ptr = host_and_port.ptr;
+		}
 
 		git_error_state_capture(&previous_error, error);
-		error = check_cb((git_cert *) cert_ptr, cert_valid, host, check_cb_payload);
+		error = check_cb((git_cert *) cert_ptr, cert_valid, host_ptr, check_cb_payload);
 		if (error == GIT_PASSTHROUGH) {
 			error = git_error_state_restore(&previous_error);
 		} else if (error < 0 && !git_error_last()) {
@@ -758,12 +767,11 @@ static int check_certificate(
 		}
 
 		git_error_state_free(&previous_error);
+		git_str_dispose(&host_and_port);
 	}
 
 	return error;
 }
-
-#define SSH_DEFAULT_PORT "22"
 
 static int _git_ssh_setup_conn(
 	ssh_subtransport *t,
--- a/src/libgit2/transports/ssh.c
+++ b/src/libgit2/transports/ssh.c
@@ -788,15 +788,8 @@ static int _git_ssh_setup_conn(
 	s->session = NULL;
 	s->channel = NULL;
 
-	if (git_net_str_is_url(url))
-		error = git_net_url_parse(&s->url, url);
-	else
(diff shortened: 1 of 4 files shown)


fxcoudert:
- Make `SSH_DEFAULT_PORT` an integer instead of a string (there are no current users of that macro in the file)
- Avoid `alloca()` which seems to be problematic on Windows builders

src/libgit2/transports/ssh.c
```diff
@@ -748,23 +750,32 @@ static int check_certificate(
...
+		char *host_and_port;
+
+		if (port == SSH_DEFAULT_PORT) {
+			host_and_port = strdup(host);
+		} else {
+			/* Maximal port number is 65535, which has 5 digits. */
+			size_t n = strlen(host) + 5 + 2;
+			host_and_port = malloc(n);
+			snprintf(host_and_port, n, "%s:%d", host, port);
+		}
```
ethomson, line 762: We should use our `git_str` class here, and we should avoid the unnecessary malloc. Something more like:

```
const char *host_ptr = host;
git_str host_and_port = GIT_STR_INIT;

if (port != SSH_DEFAULT_PORT) {
    git_str_printf(&host_and_port, "%s:%d", host, port);
    host_ptr = host_and_port.ptr;
}

// use host_ptr

git_str_free(&host_and_port);

File: src/libgit2/transports/ssh.c [hunk 2]
```diff
@@ -748,23 +750,29 @@ static int check_certificate(
...
-		error = check_cb((git_cert *) cert_ptr, cert_valid, host, check_cb_payload);
+		error = check_cb((git_cert *) cert_ptr, cert_valid, host_ptr, check_cb_payload);
 		if (error == GIT_PASSTHROUGH) {
 			error = git_error_state_restore(&previous_error);
 		} else if (error < 0 && !git_error_last()) {
 			git_error_set(GIT_ERROR_NET, "unknown remote host key");
 		}
 
 		git_error_state_free(&previous_error);
+		git_str_free(&host_and_port);
```
ethomson, line 770:
```suggestion
		git_str_dispose(&host_and_port);
```

ethomson: Sorry, I got the function name wrong. 😓

src/libgit2/transports/ssh.c
```diff
@@ -748,23 +750,29 @@ static int check_certificate(
 	if (check_cb != NULL) {
 		git_cert_hostkey *cert_ptr = &cert;
 		git_error_state previous_error = {0};
+		const char *host_ptr = host;
+		git_str host_and_port = GIT_STR_INIT;
+
+		if (port != SSH_DEFAULT_PORT) {
```
boretrk, line 756: _git_ssh_setup_conn sets port to -1 to indicate default but here it is compared to SSH_DEFAULT_PORT that is defined as 22.

ethomson: It does. But `git_net_url_parse` and friends should enforce that the port is a number, and set them to the default if the port number is omitted. So the `port = -1` should never be reached. As a result, I changed the `strtol` failure to an actual error. Good catch, thanks!

ethomson:
Thanks @fxcoudert ! I'm sorry to hear that Julia has been carrying patches around for a few years. 😢

Issue #6049, opened 2021-09-11
`git_repository_is_empty` returns wrong value for `main` branch


ArekPiekarz:
### Description
Below is a program that initializes two repositories, one with a master branch, second with a main branch.
For each it checks if the repository is empty, but gets different results for each branch.

### Reproduction steps
Create a project with following files:

main.cpp:
```c++
#include "deps/libgit2/include/git2.h"
#include <filesystem>
#include <iostream>

namespace fs = std::filesystem;

void test(const std::string& repoDir, const std::string& gitInitCommand)
{
    std::cout << "Testing repoDir: '" << repoDir << "', command: '" << gitInitCommand << "'\n";
    const auto repoPath = fs::temp_directory_path() / repoDir;
    if (fs::exists(repoPath))
    {
        fs::remove_all(repoPath);
    }

    fs::create_directory(repoPath);
    const auto command = "cd " + repoPath.string() + " && " + gitInitCommand;
    if (const auto result = std::system(command.c_str()); result != 0)
    {
        throw std::runtime_error("Failed to execute command: " + command);
    }

    git_repository* repo = nullptr;
    if (const auto result = git_repository_open(&repo, repoPath.c_str()); result != 0)
    {
        throw std::runtime_error("Failed to open repository at " + repoPath.string() + ", result: " + std::to_string(result));
    }

    const auto isEmpty = git_repository_is_empty(repo);
    std::cout << "git_repository_is_empty returned: " << std::to_string(isEmpty) << "\n\n";
}

int main()
{
    git_libgit2_init();
    test("empty_repo_master_branch", "git init");
    test("empty_repo_main_branch",   "git init --initial-branch main");
    return 0;
}
```

CMakeLists.txt:
```cmake
project(libgit2-empty-repo-test)
cmake_minimum_required(VERSION 3.0)

add_subdirectory(deps/libgit2)
add_executable(${PROJECT_NAME} main.cpp)
target_link_libraries(${PROJECT_NAME} git2)
```
Also download libgit2 as the dependency:
```
mkdir deps && cd deps
git clone https://github.com/libgit2/libgit2.git
cd ..
```
Build and run the project:
```
mkdir build && cd build
cmake ..
cmake --build .
./libgit2-empty-repo-test
```

### Expected behavior
`git_repository_is_empty` should return `1` for both branches.

### Actual behavior
`git_repository_is_empty` returns `1` for the master branch, but `0` for the main branch.
Full output:
```
[...]

PR #6508, opened 2023-02-26
BREAKING CHANGE: Change the behaviour of `git_repository_is_empty` to better align with expectations in a world where `master` isn't always the "initial branch"

timrogers:
This pull request proposes a change to the definition of an "empty" repository used in `git_repository_is_empty`.

At present, an empty is repository is one which:

1. has just been initialised
2. has no references apart from `HEAD`
3. has a `HEAD` pointing to the unborn `master` branch or, if specified, the unborn branch specified in the `init.defaultBranch` configuration value

This changes the definition of empty, removing requirement 3.

Why? Since `git_repository_is_empty` was first implemented, there has been significant cultural change in the Git community towards using an "initial branch" other than `master`.

Because of this change, the current implementation of `git_repository_is_empty` considers repos to be *not empty* which an ordinary observer would consider to be empty.

Specifically, this happens where a repo is initialized with `git init --initial-branch`. An ordinary observer would consider such a repo to be empty, but currently, libgit2 considers it to be non-empty if the `initial-branch` is not `master`. This is because, whilst `git_repository_is_empty` checks the `init.defaultBranch` to try to figure out the initial branch, `git init --initial-branch` doesn't set this value, as you can prove like this:

```bash
git config init.defaultBranch
# nothing is printed to STDOUT
mkdir test-repo
cd test-repo
git init --initial-branch main
git config init.defaultBranch
# nothing is printed to STDOUT, despite the `initial-branch` option
```

I'm not necessarily expecting this PR to be merged as is or ever - but I want to start a conversation about this, given that the current implementation is surprising and likely confusing.

See https://github.com/libgit2/libgit2/issues/6049.


timrogers:
The tests aren't working yet, but happy to get them working if this is a proposal worth moving forward with.

ethomson:
Thanks @timrogers. I haven't read the code yet but agree that `is_empty`... doesn't make much sense as-is.

ethomson:
I'd be curious for @carlosmn's take since I suspect he's using it production (and I am not).

timrogers:
> Thanks @timrogers. I haven't read the code yet but agree that `is_empty`... doesn't make much sense as-is.

Thanks! I'm conscious that breaking changes are a big deal, but this behaviour has caused some confusion for our team, and I'm sure we're not alone. I think we should do something to make this better - but I'd love the maintainers' thoughts on the right path.

timrogers:
@carlosmn 👋🏻 Just a reminder that it would be great if you could take a look at this!

PR #6521, opened 2023-03-02
Ensure that `git_index_add_all` handles ignored directories

### LINKED ISSUE #6517: .gitignore breaks git_index_add_all in certain cases
### Reproduction steps

Add `.gitignore` containing the following specification

```
.idea
```

Create a directory with ignore name.
Create a file inside it.

Construct a `git_repository` from the repo and invoke 
```
git_index_add_all(
        index,
        /* pathspec = */ nullptr,
        GIT_INDEX_ADD_FORCE,
        /* callback = */ nullptr,
        /* payload = */ nullptr
)
```
### Actual behavior

Receive a git_error with `code=-1`, `message=invalid path: '.idea/`.
 
### Expected behavior

Necessary files are added to the index just like original git does.
Why .gitignore affects this routine despite the explicit request to be ignored?

1.5.0

linux-x86_64

ethomson:
When all files beneath a directory are ignored, our diff machinery will return the directory itself. Obviously we do not operate on directories in the index, so `git_index_add_all` should ignore this case.

Fixes #6517

### DIFF
--- a/src/libgit2/index.c
+++ b/src/libgit2/index.c
@@ -3509,7 +3509,8 @@ static int index_apply_to_wd_diff(git_index *index, int action, const git_strarr
 			GIT_DIFF_RECURSE_UNTRACKED_DIRS;
 
 		if (flags == GIT_INDEX_ADD_FORCE)
-			opts.flags |= GIT_DIFF_INCLUDE_IGNORED;
+			opts.flags |= GIT_DIFF_INCLUDE_IGNORED |
+			              GIT_DIFF_RECURSE_IGNORED_DIRS;
 	}
 
 	if ((error = git_diff_index_to_workdir(&diff, repo, index, &opts)) < 0)
--- a/tests/libgit2/index/addall.c
+++ b/tests/libgit2/index/addall.c
@@ -441,6 +441,29 @@ void test_index_addall__callback_filtering(void)
 	git_index_free(index);
 }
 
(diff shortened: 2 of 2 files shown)


src/libgit2/index.c
```diff
@@ -3453,6 +3453,13 @@ static int apply_each_file(const git_diff_delta *delta, float progress, void *pa
 		    &match, NULL))
 		return 0;
 
+	/*
+	 * diff returns the directory when all files are ignored; we never
```
georgthegreat, line 3457: Is this behavior of diff intentional?
Filtering out trees in function named apply_each_file looks ... weird.

georgthegreat:
Unfortunately this does not work as intended.
GIT_INDEX_ADD_FORCE should add ignored files disregarding .gitignore contents.

At the time `/folder/asdf` is not added to the index (this is what I am seeing in my local executions with this patch applied).

ethomson:
> GIT_INDEX_ADD_FORCE should add ignored files disregarding .gitignore contents.

Ah, yes, thanks. I had it flipped around in my head.

In any case, the problem here is that diff coalesces a folder that has contents that are all ignored into a single diff entry. We need to expand that out to add files to the index. This update should do that.

georgthegreat:
@ethomson the latest version works for us.
Thanks a lot for the quick fix!

PR #6523, opened 2023-03-04
Fix potential null dereference

ZhipengXue97:
**Description**
Fix the null dereference warning detected by static analyse tool infer@facebook

**Warning Type**
Null Dereference

**omponent Name**
libgit2/tests/libgit2/merge/merge_helpers.c (line:148)

**Additional Information**
pointer `reuc` last assigned on line 148 could be null and is dereferenced at line 150.

### DIFF
--- a/tests/libgit2/merge/merge_helpers.c
+++ b/tests/libgit2/merge/merge_helpers.c
@@ -145,7 +145,7 @@ void merge__dump_reuc(git_index *index)
 
 	printf ("\nREUC:\n");
 	for (i = 0; i < git_index_reuc_entrycount(index); i++) {
-		reuc = git_index_reuc_get_byindex(index, i);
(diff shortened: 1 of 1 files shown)


ethomson:
Some feedback on this change:

* You can't put `NULL`s into the REUC (`index_insert_reuc` does a `NULL` check), so the only way `git_index_reuc_get_byindex` can return `NULL` is if the index is out of range
* We do a check of the size of the REUC right before calling this, and there's nothing writing to the REUC to mutate it, including other threads
* This is a debugging function used to dump data when running the tests
* The function is never called

I mention this because your profile says that you're a PhD candidate, and this may be part of your research.

ZhipengXue97:
@ethomson Thanks very much for your reply!
I am researching how to improve the accuracy of static analysis tools, and your response has been very helpful.

PR #6527, opened 2023-03-08
sysdir: Do not declare win32 functions on non-win32 platforms

Batchyx:
These declaration poses problems on some embedded or retro Linux systems that deliberately disable support for wchar_t from their libc.

```
<overlong path>/libgit2-1.6.2/src/libgit2/sysdir.h:138:53: error: unknown type name ‘wchar_t’
  138 | extern int git_win32__set_registry_system_dir(const wchar_t *mock_sysdir);
      |                                                     ^~~~~~~
```


ethomson:
Thanks for the fix!

PR #6530, opened 2023-03-14
pack: use 64 bits for the number of objects

carlosmn:
Keeping it as a 32-bit value means the min size calculation overflows or gets truncated which can lead to issues with large packfiles.

--

It's a bit disappointing that the compiler doesn't seem to care that we've been cutting off this value when we have a lot of objects.

### DIFF
--- a/src/libgit2/pack.c
+++ b/src/libgit2/pack.c
@@ -200,7 +200,8 @@ static void pack_index_free(struct git_pack_file *p)
 static int pack_index_check_locked(const char *path, struct git_pack_file *p)
 {
 	struct git_pack_idx_header *hdr;
-	uint32_t version, nr, i, *index;
+	uint32_t version, i, *index;
+	uint64_t nr = 0;
 	void *idx_map;
 	size_t idx_size;
 	struct stat st;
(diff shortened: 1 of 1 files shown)


ethomson:
Do we need to change `git_pack_file.num_objects` as well?

carlosmn:
`git_pack_file.num_objects` would not have quite the same issue as we're not overflowing 32 bits when counting objects, but when multiplying the object count by the size of the entries. That can overflow 32-bits and you could get the same fix by casting `(uint64_t) nr` in the `min_size` calculation instead. Maybe doing it like this is more confusing than casting? I thought this would be a bit more elegant, but maybe not?

The number of objects would still be limited to `UINT32_MAX` as the first-level fanout table only has space for 4 bytes in each entry. So to actually answer you question, the number of objects in the pack is fine (and arguably more correct) to be 32 bits.

boretrk:
I prefer casting it in the calculation. It makes it clear to me that 64bit is needed for the calculation.

ethomson:
Aha - thanks, I was reading on my phone and really only noticed the build error. I think that casting might make more sense and would likely resolve this compiler warning: https://github.com/libgit2/libgit2/actions/runs/4417160914/jobs/7759039903#step:7:274

ethomson:
Thanks @carlosmn !

PR #6535, opened 2023-03-21
Introduce timeouts on sockets

ethomson:
Make socket I/O non-blocking and add optional timeouts.

Users may now set `GIT_OPT_SET_SERVER_CONNECT_TIMEOUT` to set a shorter connection timeout. (The connect timeout cannot be longer than the operating system default.) Users may also now configure the socket read and write timeouts with `GIT_OPT_SET_SERVER_TIMEOUT`.

By default, connects still timeout based on the operating system defaults (typically 75 seconds) and socket read and writes block.

Add a test against our custom testing git server that ensures that we can timeout reads against a slow server.


ethomson:
Note: this is stacked on #6533.

mascguy:
Folks, I'm the maintainer of this lib for the MacPorts project. And one quick observation: The use of macOS constant `errSSLNetworkTimeout` in file `src/libgit2/streams/stransport.c`, now causes build failures for macOS 10.8 thru 10.12. (As that constant is only available in 10.13+.)

I'm happy to write up a formal issue. But just wanted to give you folks a proactive heads-up, to start the discussion on what the best approach to fixing this is.

Thoughts...?

mascguy:
> Folks, I'm the maintainer of this lib for the MacPorts project. And one quick observation: The use of macOS constant `errSSLNetworkTimeout` in file `src/libgit2/streams/stransport.c`, now causes build failures for macOS 10.8 thru 10.12. (As that constant is only available in 10.13+.)
> 
> I'm happy to write up a formal issue. But just wanted to give you folks a proactive heads-up, to start the discussion on what the best approach to fixing this is.
> 
> Thoughts...?

Oh, and we're tracking this issue via ticket:

[67774 - libgit2-devel @1.7.0: builds fail for 10.8 thru 10.12: use of undeclared identifier 'errSSLNetworkTimeout'](https://trac.macports.org/ticket/67774)

mascguy:
> > Folks, I'm the maintainer of this lib for the MacPorts project. And one quick observation: The use of macOS constant `errSSLNetworkTimeout` in file `src/libgit2/streams/stransport.c`, now causes build failures for macOS 10.8 thru 10.12. (As that constant is only available in 10.13+.)
> > I'm happy to write up a formal issue. But just wanted to give you folks a proactive heads-up, to start the discussion on what the best approach to fixing this is.
> > Thoughts...?
> 
> Oh, and we're tracking this issue via ticket:
> 
> [67774 - libgit2-devel @1.7.0: builds fail for 10.8 thru 10.12: use of undeclared identifier 'errSSLNetworkTimeout'](https://trac.macports.org/ticket/67774)

The simplest and least fragile option for us, would be to simply disable use of SecurityFramework for 10.12 and earlier. That's presently being done automatically for 10.7 and earlier, via checks in CMake file `cmake/SelectHTTPSBackend.cmake`. Though it doesn't look (?) like there's a formal option to easily force that. (Would be easy enough to patch said file to add one, but we'd prefer to collaborate with you folks on that.)

ethomson:
> The simplest and least fragile option for us, would be to simply disable use of SecurityFramework for 10.12 and earlier.

I think it would be disappointing to use OpenSSL on macOS when there's a perfectly good SSL library that gets software updates with the system. I realize that MacPorts and Homebrew likely will do software updates, but somebody compiling libgit2 from source on 10.11 likely _won't_.

I think that we can likely just hardcode the value (like we do for ioErr)

mascguy:
> > The simplest and least fragile option for us, would be to simply disable use of SecurityFramework for 10.12 and earlier.
> 
> I think it would be disappointing to use OpenSSL on macOS when there's a perfectly good SSL library that gets software updates with the system. I realize that MacPorts and Homebrew likely will do software updates, but somebody compiling libgit2 from source on 10.11 likely _won't_.
> 
> I think that we can likely just hardcode the value (like we do for ioErr)

Sounds good, thank you for the quick response!

Is this something we should submit a PR for, or would you folks prefer to make the change?

ethomson:
> Is this something we should submit a PR for, or would you folks prefer to make the change?

You're very much welcome to. I won't have time to look at it until this weekend or early next week.

mascguy:
> > Is this something we should submit a PR for, or would you folks prefer to make the change?
> 
> You're very much welcome to. I won't have time to look at it until this weekend or early next week.

Done, thanks again for your help and quick response Edward!

[PR 6610 - stransport: macOS: replace errSSLNetworkTimeout, with hard-coded value](https://github.com/libgit2/libgit2/pull/6610)

Issue #6553, opened 2023-05-03
Tree lookup fails with "fake" repositories since 1.6.1


oreiche:
### Reproduction steps
The following C code passes with libgit2 <=1.5.2 and fails with >= 1.6.1

~~~c
#include <git2.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

int main(int argc, const char *argv[]) {
  char template[] = "/tmp/tmpdir.XXXXXX";
  char *tmpdir = mkdtemp(template);

  git_libgit2_init();

  git_oid blob_id;
  git_oid tree_id;
  git_odb *odb = NULL;
  { // create test repo and obtain odb
    git_repository *repo = NULL;
    if (git_repository_init(&repo, tmpdir, /*bare does not matter*/0) != 0) {
      fprintf(stderr, "Could not initialize repository: %s\n",
              git_error_last()->message);
      return 1;
    }
    if (git_repository_odb(&odb, repo) != 0) {
      fprintf(stderr, "Could not open ODB: %s\n", git_error_last()->message);
      return 1;
    }
    if (git_odb_write(&blob_id, odb, "foo", 3, GIT_OBJECT_BLOB) != 0) {
       fprintf(stderr, "Could not write blob: %s\n",
               git_error_last()->message);
       return 1;
    }
    git_treebuilder *tb = NULL;
    if (git_treebuilder_new(&tb, repo, NULL) != 0) {
      fprintf(stderr, "Could not create tree builder: %s\n",
              git_error_last()->message);
      return 1;
    }
    if (git_treebuilder_insert(NULL, tb, "foo", &blob_id, GIT_FILEMODE_BLOB) !=
        0) {
      fprintf(stderr, "Could not insert blob foo: %s\n",
              git_error_last()->message);
      return 1;
    }
    if (git_treebuilder_write(&tree_id, tb) != 0) {
      fprintf(stderr, "Could not write tree: %s\n", git_error_last()->message);
      return 1;
    }
    git_treebuilder_free(tb);
    git_repository_free(repo);
  }

  git_repository *repo = NULL;
  //if (git_repository_open(&repo, tmpdir) != 0) { // "real" repo works
  if (git_repository_wrap_odb(&repo, odb) != 0) {  // "fake" repo fails
    fprintf(stderr, "Could not create \"fake\" repository: %s\n",
            git_error_last()->message);
    return 1;
  }

  git_blob *blob = NULL;
[...]

PR #6554, opened 2023-05-03
Fix missing oid type for "fake" repositories

oreiche:
... otherwise git_tree__parse_raw() will fail to obtain the correct oid size, which causes the entire parse to fail.

Fixes issue https://github.com/libgit2/libgit2/issues/6553.

### DIFF
--- a/src/libgit2/repository.c
+++ b/src/libgit2/repository.c
@@ -1152,6 +1152,8 @@ int git_repository_wrap_odb(git_repository **repo_out, git_odb *odb)
 	repo = repository_alloc();
 	GIT_ERROR_CHECK_ALLOC(repo);
 
+	repo->oid_type = GIT_OID_DEFAULT;
+
 	git_repository_set_odb(repo, odb);
 	*repo_out = repo;
 


src/libgit2/repository.c
```diff
@@ -1147,15 +1147,24 @@ int git_repository_open_from_worktree(git_repository **repo_out, git_worktree *w
...
+	int error;
+	git_repository *repo = NULL;
+	git_config *config = NULL;
 
 	repo = repository_alloc();
 	GIT_ERROR_CHECK_ALLOC(repo);
 
 	git_repository_set_odb(repo, odb);
+
+	if ((error = obtain_config_and_set_oid_type(&config, repo)) < 0)
```
ethomson, line 1159: Interesting. I was going to argue that this was not necessary because a fake repository would never have a config of its own. But then I poked around to validate this thinking. This made me realize that a) git limits the repository format version config and / or per-repo extensions to the repo config and doesn't look at system / global. and b) we do not do this.

That's something that we should fix independently of this problem. But it means that we can simplify this to not bother looking at the config. So I think that the simpler solution is just to `repo->oid_type = GIT_OID_DEFAULT;`...

But I think that in the long term, though, this function needs to take an oid type. I think that I might do:

```c
#ifdef GIT_EXPERIMENTAL_SHA256
int git_repository_wrap_odb(git_repository **repo_out, git_odb *odb, git_oid_t oid_type)
#else
int git_repository_wrap_odb(git_repository **repo_out, git_odb *odb)
#endif
{
#ifndef GIT_EXPERIMENTAL_SHA256
	git_oid_t oid_type = GIT_OID_DEFAULT;
#endif

	repo = repository_alloc();
	GIT_ERROR_CHECK_ALLOC(repo);

	repo->oid_type = oid_type;

	git_repository_set_odb(repo, odb);

	*repo_out = repo;
	return 0;
}
```

ethomson:
Good catch, thanks for the fix. I made a simplifying suggestion -- and a more complex suggestion as well. 😁

src/libgit2/repository.c
oreiche: Alright, I updated the branch to implement the simple solution :)

ethomson:
Thanks again for the fix!

PR #6555, opened 2023-05-03
util: detect all possible qsort_r and qsort_s variants

DimitryAndric:
As reported in https://bugs.freebsd.org/271234, recent versions of FreeBSD have adjusted the prototype for `qsort_r()` to match the POSIX interface. This causes libgit2's CMake configuration check to fail to detect `qsort_r()`, making it fall back to `qsort_s()`, which in libgit2 also has an incompatible interface. With recent versions of clang this results in a "incompatible function pointer types" compile error.

Summarizing, there are four variations of 'qsort-with-context':
* old style BSD `qsort_r()`, used in FreeBSD 13 and earlier, where the comparison function has the context parameter first
* GNU or POSIX `qsort_r()`, also used in FreeBSD 14 and later, where the comparison function has the context parameter last
* C11 `qsort_s()`, where the comparison function has the context parameter last
* Microsoft `qsort_s()`, where the comparison function has the context parameter first

Add explicit detections for all these variants, so they get detected as (in the same order as above):

* `GIT_QSORT_R_BSD`
* `GIT_QSORT_R_GNU`
* `GIT_QSORT_S_C11`
* `GIT_QSORT_S_MSC`

An additional complication is that on FreeBSD 14 and later, `<stdlib.h>` uses the C11 `_Generic()` macro mechanism to automatically select the correct `qsort_r()` prototype, depending on the caller's comparison function argument. This breaks CMake's `check_prototype_definition()` functionality, since it tries to redefine the function, and the _Generic macro is expanded inline causing a compile error.

Work around that problem by putting the function names in parentheses, to prevent the preprocessor from using a macro to replace the function name.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -58,15 +58,29 @@ add_feature_info(futimens GIT_USE_FUTIMENS "futimens support")
 
 # qsort
 
+# old-style FreeBSD qsort_r() has the 'context' parameter as the first argument
+# of the comparison function:
 check_prototype_definition(qsort_r
-	"void qsort_r(void *base, size_t nmemb, size_t size, void *thunk, int (*compar)(void *, const void *, const void *))"
+	"void (qsort_r)(void *base, size_t nmemb, size_t size, void *context, int (*compar)(void *, const void *, const void *))"
 	"" "stdlib.h" GIT_QSORT_R_BSD)
 
+# GNU or POSIX qsort_r() has the 'context' parameter as the last argument of the
+# comparison function:
 check_prototype_definition(qsort_r
-	"void qsort_r(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *arg)"
+	"void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context)"
 	"" "stdlib.h" GIT_QSORT_R_GNU)
 
-check_function_exists(qsort_s GIT_QSORT_S)
+# C11 qsort_s() has the 'context' parameter as the last argument of the
+# comparison function, and returns an error status:
+check_prototype_definition(qsort_s
+	"errno_t (qsort_s)(void *base, rsize_t nmemb, rsize_t size, int (*compar)(const void *, const void *, void *), void *context)"
+	"0" "stdlib.h" GIT_QSORT_S_C11)
+
+# MSC qsort_s() has the 'context' parameter as the first argument of the
+# comparison function, and as the last argument of qsort_s():
+check_prototype_definition(qsort_s
+	"void (qsort_s)(void *base, size_t num, size_t width, int (*compare )(void *, const void *, const void *), void *context)"
+	"" "stdlib.h" GIT_QSORT_S_MSC)
 
 # random / entropy data
 
--- a/src/util/util.c
+++ b/src/util/util.c
@@ -18,7 +18,7 @@
 # endif
 # include <windows.h>
 
-# ifdef GIT_QSORT_S
+# ifdef GIT_QSORT_S_MSC
 #  include <search.h>
 # endif
 #endif
@@ -673,7 +673,7 @@ size_t git__unescape(char *str)
 	return (pos - str);
 }
 
-#if defined(GIT_QSORT_S) || defined(GIT_QSORT_R_BSD)
+#if defined(GIT_QSORT_S_MSC) || defined(GIT_QSORT_R_BSD)
 typedef struct {
 	git__sort_r_cmp cmp;
 	void *payload;
@@ -690,7 +690,8 @@ static int GIT_LIBGIT2_CALL git__qsort_r_glue_cmp(
 
 #if !defined(GIT_QSORT_R_BSD) && \
 	!defined(GIT_QSORT_R_GNU) && \
-	!defined(GIT_QSORT_S)
+	!defined(GIT_QSORT_S_C11) && \
+	!defined(GIT_QSORT_S_MSC)
 static void swap(uint8_t *a, uint8_t *b, size_t elsize)
 {
 	char tmp[256];
@@ -721,12 +722,14 @@ static void insertsort(
 void git__qsort_r(
 	void *els, size_t nel, size_t elsize, git__sort_r_cmp cmp, void *payload)
 {
-#if defined(GIT_QSORT_R_BSD)
+#if defined(GIT_QSORT_R_GNU)
+	qsort_r(els, nel, elsize, cmp, payload);
+#elif defined(GIT_QSORT_S_C11)
+	qsort_s(els, nel, elsize, cmp, payload);
+#elif defined(GIT_QSORT_R_BSD)
 	git__qsort_r_glue glue = { cmp, payload };
 	qsort_r(els, nel, elsize, &glue, git__qsort_r_glue_cmp);
-#elif defined(GIT_QSORT_R_GNU)
-	qsort_r(els, nel, elsize, cmp, payload);
-#elif defined(GIT_QSORT_S)
+#elif defined(GIT_QSORT_S_MSC)
 	git__qsort_r_glue glue = { cmp, payload };
 	qsort_s(els, nel, elsize, git__qsort_r_glue_cmp, &glue);
 #else
--- a/src/util/git2_features.h.in
+++ b/src/util/git2_features.h.in
@@ -26,7 +26,8 @@
 
 #cmakedefine GIT_QSORT_R_BSD
 #cmakedefine GIT_QSORT_R_GNU
-#cmakedefine GIT_QSORT_S
+#cmakedefine GIT_QSORT_S_C11
+#cmakedefine GIT_QSORT_S_MSC
 
 #cmakedefine GIT_SSH 1
 #cmakedefine GIT_SSH_MEMORY_CREDENTIALS 1


src/CMakeLists.txt
```diff
@@ -58,15 +58,29 @@ add_feature_info(futimens GIT_USE_FUTIMENS "futimens support")
...
 # qsort
 
+# old-style BSD qsort_r() has the 'thunk' parameter as the first argument of the
+# comparison function:
 check_prototype_definition(qsort_r
-	"void qsort_r(void *base, size_t nmemb, size_t size, void *thunk, int (*compar)(void *, const void *, const void *))"
+	"void (qsort_r)(void *base, size_t nmemb, size_t size, void *thunk, int (*compar)(void *, const void *, const void *))"
 	"" "stdlib.h" GIT_QSORT_R_BSD)
 
+# new-style BSD and GNU qsort_r() has the 'thunk' parameter as the last
```
emaste, line 67: This is maybe "new style FreeBSD qsort_r" -- I don't know that other BSDs have picked it up?

emaste:
Do both the `GIT_QSORT_R_BSD` and `GIT_QSORT_R_GNU` tests succeed on FreeBSD 14 (via _Generic)?

src/util/util.c
```diff
@@ -726,7 +727,9 @@ void git__qsort_r(
 	qsort_r(els, nel, elsize, &glue, git__qsort_r_glue_cmp);
 #elif defined(GIT_QSORT_R_GNU)
```
delphij, line 728: The code uses glue when dealing with traditional BSD qsort_r.  When both are present, it seems that GIT_QSORT_R_GNU should be preferred.

src/CMakeLists.txt
delphij: maybe use POSIX qsort_r?  (reference: https://www.austingroupbugs.net/view.php?id=900 )

delphij (commented):
I think the detection should prefer the POSIX/GNU qsort_r() variant when both are available, because the support of traditional BSD style qsort_r() is done with glues.

src/CMakeLists.txt
DimitryAndric: Neither NetBSD nor OpenBSD even have `qsort_r()` (and they don't have `qsort_s()` either), so technically it is more correct to use the term 'FreeBSD', or indeed just 'POSIX' here.

I guess the "old-style BSD" comment should then really be "old-style FreeBSD", but I'm inclined to leave the `GIT_QSORT_R_BSD` as it has always been.

src/util/util.c
DimitryAndric: They won't both be present, either the old variant or the new variant will be detected at configure time, but not both. Still, putting the most preferred variants at the top is less ambiguous, I guess.

delphij (approved):
Looks good to me.

ethomson:
Thanks for the PR! I'm not seeing a system sort get selected on our Linux or macOS CI builds, which I would have expected. Is my expectation wrong or do we need to adjust the cmake magic a bit?

DimitryAndric:
> Thanks for the PR! I'm not seeing a system sort get selected on our Linux or macOS CI builds, which I would have expected. Is my expectation wrong or do we need to adjust the cmake magic a bit?

Hmm that's an interesting one, I see for example with the Ubuntu Xenial build (https://github.com/libgit2/libgit2/actions/runs/4918385982/jobs/8796807358?pr=6555):

```
 -- Checking prototype qsort_r for GIT_QSORT_R_BSD - False
-- Checking prototype qsort_r for GIT_QSORT_R_GNU - False
-- Checking prototype qsort_s for GIT_QSORT_S_C11 - False
-- Checking prototype qsort_s for GIT_QSORT_S_MSC - False
```

But when I spin up a local Xenial container, and run the build in there, it works as expected (it only finds the GNU/POSIX variant):

```
-- Checking prototype qsort_r for GIT_QSORT_R_BSD - False
-- Checking prototype qsort_r for GIT_QSORT_R_GNU - True
-- Checking prototype qsort_s for GIT_QSORT_S_C11 - False
-- Checking prototype qsort_s for GIT_QSORT_S_MSC - False
```

Similar for the macOS build (https://github.com/libgit2/libgit2/actions/runs/4918385982/jobs/8796807907?pr=6555) which shows:

```
 -- Checking prototype qsort_r for GIT_QSORT_R_BSD
-- Checking prototype qsort_r for GIT_QSORT_R_BSD - False
-- Checking prototype qsort_r for GIT_QSORT_R_GNU
-- Checking prototype qsort_r for GIT_QSORT_R_GNU - False
-- Checking prototype qsort_s for GIT_QSORT_S_C11
-- Checking prototype qsort_s for GIT_QSORT_S_C11 - False
-- Checking prototype qsort_s for GIT_QSORT_S_MSC
-- Checking prototype qsort_s for GIT_QSORT_S_MSC - False
```

whereas on my own Mac (M1 with macOS 13.3.1 and Xcode 14.3) it correctly detects only the old-style BSD variant:

```
-- Checking prototype qsort_r for GIT_QSORT_R_BSD
-- Checking prototype qsort_r for GIT_QSORT_R_BSD - True
-- Checking prototype qsort_r for GIT_QSORT_R_GNU
-- Checking prototype qsort_r for GIT_QSORT_R_GNU - False
-- Checking prototype qsort_s for GIT_QSORT_S_C11
-- Checking prototype qsort_s for GIT_QSORT_S_C11 - False
-- Checking prototype qsort_s for GIT_QSORT_S_MSC
-- Checking prototype qsort_s for GIT_QSORT_S_MSC - False
```

Can anybody get at the `build/CMakeFiles/CMakeError.log` files in the failing CI containers? Whatever goes wrong should be visible in there.

DimitryAndric:
Ah, I could pull the container image used for building, and reproduce the problem in there, but it's due to `-DENABLE_WERROR=ON` being used in these CI jobs:

```
FAILED: /usr/bin/cc    -D_GNU_SOURCE  -Werror -Werror -Wall -Wextra -fvisibility=hidden -fPIC -Wno-documentation-deprecated-sync -Wno-missing-field-initializers -Wmissing-declarations -Wstrict-aliasing -Wstrict-prototypes -Wdeclaration-after-statement -Wshift-count-overflow -Wunused-function -Wint-conversion -Wc99-c11-compat -Wformat -Wformat-security -o CMakeFiles/cmTC_da6d8.dir/CheckPrototypeDefinition.c.o   -c CheckPrototypeDefinition.c
CheckPrototypeDefinition.c: In function 'qsort_r':
CheckPrototypeDefinition.c:14:22: error: unused parameter 'base' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                      ^
CheckPrototypeDefinition.c:14:35: error: unused parameter 'nmemb' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                   ^
CheckPrototypeDefinition.c:14:49: error: unused parameter 'size' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                                 ^
CheckPrototypeDefinition.c:14:61: error: unused parameter 'compar' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                                             ^
CheckPrototypeDefinition.c:14:112: error: unused parameter 'context' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                                                                                                ^
CheckPrototypeDefinition.c: At top level:
cc1: error: unrecognized command line option '-Wno-documentation-deprecated-sync' [-Werror]
cc1: all warnings being treated as errors
ninja: build stopped: subcommand failed.
```

The test function added by CMake always has an empty body, so the incoming parameters are indeed purposefully unused. The warning should be suppressed, but it's tricky to do this in a platform-independent way, i..e `__attribute__((unused))` is specific to gcc and clang, and not supported by MSVC.

Note also the error on `-Wno-documentation-deprecated-sync`, but I have no idea where that comes from, not from the changes in this pull request, in any case.

DimitryAndric:
Ah, I could pull the container image used for building, and reproduce the problem in there, but it's due to `-DENABLE_WERROR=ON` being used in these CI jobs:

```
FAILED: /usr/bin/cc    -D_GNU_SOURCE  -Werror -Werror -Wall -Wextra -fvisibility=hidden -fPIC -Wno-documentation-deprecated-sync -Wno-missing-field-initializers -Wmissing-declarations -Wstrict-aliasing -Wstrict-prototypes -Wdeclaration-after-statement -Wshift-count-overflow -Wunused-function -Wint-conversion -Wc99-c11-compat -Wformat -Wformat-security -o CMakeFiles/cmTC_da6d8.dir/CheckPrototypeDefinition.c.o   -c CheckPrototypeDefinition.c
CheckPrototypeDefinition.c: In function 'qsort_r':
CheckPrototypeDefinition.c:14:22: error: unused parameter 'base' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                      ^
CheckPrototypeDefinition.c:14:35: error: unused parameter 'nmemb' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                   ^
CheckPrototypeDefinition.c:14:49: error: unused parameter 'size' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                                 ^
CheckPrototypeDefinition.c:14:61: error: unused parameter 'compar' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                                             ^
CheckPrototypeDefinition.c:14:112: error: unused parameter 'context' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                                                                                                ^
CheckPrototypeDefinition.c: At top level:
cc1: error: unrecognized command line option '-Wno-documentation-deprecated-sync' [-Werror]
cc1: all warnings being treated as errors
ninja: build stopped: subcommand failed.
```

The test function added by CMake always has an empty body, so the incoming parameters are indeed purposefully unused. The warning should be suppressed, but it's tricky to do this in a platform-independent way, i..e `__attribute__((unused))` is specific to gcc and clang, and not supported by MSVC.

Note also the error on `-Wno-documentation-deprecated-sync`, but I have no idea where that comes from, not from the changes in this pull request, in any case.

DimitryAndric:
Ah, I could pull the container image used for building, and reproduce the problem in there, but it's due to `-DENABLE_WERROR=ON` being used in these CI jobs:

```
FAILED: /usr/bin/cc    -D_GNU_SOURCE  -Werror -Werror -Wall -Wextra -fvisibility=hidden -fPIC -Wno-documentation-deprecated-sync -Wno-missing-field-initializers -Wmissing-declarations -Wstrict-aliasing -Wstrict-prototypes -Wdeclaration-after-statement -Wshift-count-overflow -Wunused-function -Wint-conversion -Wc99-c11-compat -Wformat -Wformat-security -o CMakeFiles/cmTC_da6d8.dir/CheckPrototypeDefinition.c.o   -c CheckPrototypeDefinition.c
CheckPrototypeDefinition.c: In function 'qsort_r':
CheckPrototypeDefinition.c:14:22: error: unused parameter 'base' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                      ^
CheckPrototypeDefinition.c:14:35: error: unused parameter 'nmemb' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                   ^
CheckPrototypeDefinition.c:14:49: error: unused parameter 'size' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                                 ^
CheckPrototypeDefinition.c:14:61: error: unused parameter 'compar' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                                             ^
CheckPrototypeDefinition.c:14:112: error: unused parameter 'context' [-Werror=unused-parameter]
 void (qsort_r)(void *base, size_t nmemb, size_t size, int (*compar)(const void *, const void *, void *), void *context) {
                                                                                                                ^
CheckPrototypeDefinition.c: At top level:
cc1: error: unrecognized command line option '-Wno-documentation-deprecated-sync' [-Werror]
cc1: all warnings being treated as errors
ninja: build stopped: subcommand failed.
```

The test function added by CMake always has an empty body, so the incoming parameters are indeed purposefully unused. The warning should be suppressed, but it's tricky to do this in a platform-independent way, i..e `__attribute__((unused))` is specific to gcc and clang, and not supported by MSVC.

Note also the error on `-Wno-documentation-deprecated-sync`, but I have no idea where that comes from, not from the changes in this pull request, in any case.

ethomson:
> Ah, I could pull the container image used for building, and reproduce the problem in there, but it's due to -DENABLE_WERROR=ON being used in these CI jobs:

_Aha_. Thanks for digging deep on this. That makes sense.

I don't see an easy way around this. 🙃 

Thanks again for the fix.

DimitryAndric:
Actually I have a fix lined up for this, but I'll put it in another pull request :)

mfechner:
@ethomson do you plan to back-port this also to 1.5 and 1.6 branch? Should I create a new Merge Request for this?
It is needed e.g. for gitaly on FreeBSD as gitaly currently depends on git2go which depends on libgit2 1.5.

mfechner:
I added the required patches for FreeBSD for version 1.5 now here:
https://github.com/mfechner/libgit2/tree/maint/v1.5

This version is also used for gitaly to make to compilable on FreeBSD 14 which is included here:
https://cgit.freebsd.org/ports/commit/?id=64cfdbdb78b0806923a161f5a14d967d75b9fe43

PR #6559, opened 2023-05-10
remote: add update_refs with git_refspec

ethomson:
`git` shows the _remote_ and _local_ ref information during a fetch. For example:

```
 * [new branch]          ethomson/nonblocking   -> origin/ethomson/nonblocking
```

This is not information that users of libgit2 can currently surface since our `update_tips` callback includes only src/dest OIDs, and the _local_ ref name. Nor can one reliably reverse engineer the remote branch name, since a remote configuration could map two remote specs to a local spec. For example:

```
        fetch = +refs/pull/*/head:refs/remotes/asdf/*
        fetch = +refs/heads/*:refs/remotes/asdf/*
```

In this case, given `refs/remotes/asdf/foo` in the `update_tips`, a caller would have no idea whether the remote was `refs/pull/foo/head` or `refs/heads/foo`.

Provide the matching remote `git_refspec` so that callers can `rtransform` themselves into the remote.

### DIFF
--- a/examples/fetch.c
+++ b/examples/fetch.c
@@ -13,20 +13,24 @@ static int progress_cb(const char *str, int len, void *data)
  * updated. The message we output depends on whether it's a new one or
  * an update.
  */
-static int update_cb(const char *refname, const git_oid *a, const git_oid *b, void *data)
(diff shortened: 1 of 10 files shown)


fxcoudert:
This ABI-breaking change is not listed in the 1.9.0 release notes. Could it be added?

PR #6569, opened 2023-05-25
Ensure well-defined value range for args to <ctype.h> functions.

he32:
Found while building on NetBSD's pkgsrc, which defaults to building with `-Wchar-subscripts`.  To quote the NetBSD ctype(3) man page:

```
     The argument of these functions is of type int, but only a very
     restricted subset of values are actually valid.  The argument must either
     be the value of the macro EOF (which has a negative value), or must be a
     non-negative value within the range representable as unsigned char.
     Passing invalid values leads to undefined behavior.
```

Hence, nearly half of the `signed char` values result in undefined behaviour, which cannot be good.

With these changes applied, libgit2 builds without warning on pkgsrc.

### DIFF
--- a/deps/ntlmclient/ntlm.c
+++ b/deps/ntlmclient/ntlm.c
@@ -988,9 +988,9 @@ static inline bool generate_lm_hash(
 	keystr2_len = (password_len > 7) ? MIN(14, password_len) - 7 : 0;
 
 	for (i = 0; i < keystr1_len; i++)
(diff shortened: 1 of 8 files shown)


ethomson:
Fixed via #6770

ethomson:
Thanks for the PR! We iterated a bit in #6679 and I pushed an update. LMK if this doesn't resolve the issue for netbsd.

PR #6572, opened 2023-05-31
Git blame buffer gives the wrong result in many cases where there are…

thosey:
… multiple local edits.

I added 7 new failing tests to tests/libgit2/blame/buffer.c. 

blame.c has been updated to get these tests to pass. 

Fixes include:
    shift_hunks_by now no longer shifts hunks before the start line. 
    Adjusting the wedge line in the case where a new hunk deletes and adds multiple lines.
    In buffer_line_cb, hunks are now shifted after the current diff line for additions.
    Fixing the logic for removing a line in buffer_line_cb to work with multi-line deletions.


Murmele:
In Gittyup we have a similar patch like your one. It would be nice if this one gets merged, so we don't need our patch anymore

Issue #6576, opened 2023-06-10
git_diff_index_to_workdir fails to load untracked content when GIT_DIFF_REVERSE and GIT_DIFF_SHOW_UNTRACKED_CONTENT are set.


arroz:
### Reproduction steps
Assume the following code on a workdir with an untracked file with at least one line of content:

```
git_diff_options opts = GIT_DIFF_OPTIONS_INIT;
git_diff *diff = NULL;
opts.flags |= GIT_DIFF_INCLUDE_UNTRACKED | GIT_DIFF_SHOW_UNTRACKED_CONTENT | GIT_DIFF_REVERSE;
git_diff_index_to_workdir(&diff, repo, NULL, &opts)
```

This diff will not properly load the content when inspected. Specifically, the following call will never get run `diff_hunk_cb` or `diff_line_cb`:

```
git_diff_foreach(diff, diff_file_cb, diff_binary_cb, diff_hunk_cb, diff_line_cb, &exp)
```

### Expected behavior

The call above should call `diff_hunk_cb` once and `diff_line_cb` for each content line.

### Actual behavior

The calls never happen because, internally, the content is never loaded.

### Version of libgit2 (release number or SHA1)

`2f20fe8869d7a1df7c9b7a9e2939c1a20533c6dc`

### Operating system(s) tested

macOS

PR #6577, opened 2023-06-10
#6576 git_diff_index_to_workdir reverse now loads untracked content

arroz:


arroz:
The CI failure doesn't seem related to this patch.

PR #6578, opened 2023-06-10
Get the latest shallow clone support

pysco68:
Because we need it 💛


tests/headertest/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 # even when they have aggressive C90 warnings enabled.
 
 add_executable(headertest headertest.c)
-set_target_properties(headertest PROPERTIES C_STANDARD 90)
+set_target_properties(headertest PROPERTIES C_STANDARD 11)
```
boretrk, line 6: The reason headertest is done with `C_STANDARD 90` is to make sure that projects using C90 can link against libgit2 even if the library itself requires a different standard to compile.

For compliance with [the POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/) it is probably not a good idea to exceed ISO/IEC 9899:1999

pysco68: Hey @boretrk I'm super sorry, this wasn't supposed to be a  PR against upstream, but rather for our internal fork. 

We just "fixed up" the build system so it would work in our setup. 

We will probably want to upstream some of the things we did at some point but except [3730623](https://github.com/libgit2/libgit2/pull/6578/commits/373062301b9201b31e21fa7386e7d63ada93f3e0) (which is dealing with Symlinks that change the target type on Windows [e.g. a symlink that points to a folder and in a following revision to a file or vice versa] because Windows doesn't handle this transparently [basically a symlink is either a folder-symlink or a file-symlink but doesn't mutate, you have to recreate it]) this is really just a dirty state to have it working in our build (boringSSL vs openSSL, a non-system provided ZLIB and libSSH2 etc).

Again, sorry for the noise,

If you don't mind I'll just close this up,

Best,
- Yannic

PR #6581, opened 2023-06-19
Use #!/bin/bash for script with bash-specific commands

roehling:
The test/Resources/push.sh calls pushd/popd, which is not a POSIX shell feature but a Bash extension.

### DIFF
--- a/tests/resources/push.sh
+++ b/tests/resources/push.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #creates push_src repo for libgit2 push tests.
 set -eu
 
--- a/tests/resources/push.sh
+++ b/tests/resources/push.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #creates push_src repo for libgit2 push tests.
 set -eu
 


tests/resources/push.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
ethomson, line 1: Makes sense - but should we:

```suggestion
#!/usr/bin/env bash
```

roehling: I don't think that's necessary. The `#!/usr/bin/env foo` idiom was invented for things like Python scripts, which might be run from a virtual environment, so you cannot hard-code the interpreter path. Bash has no such use-case and `/bin/bash` is not going to go away any time soon, so why pay the cost for an additional `exec("/usr/bin/env")` syscall?

ethomson: This isn't in a tight loop or a place where any syscall is expensive. But the performance question needs to take a backseat to correctness and usability -- where is `bash` on a Solaris system?

roehling: It's in `/bin/bash` AFAIK, but everything before Solaris 11 has a stone-age bash which is not even POSIX compliant.

neithernut: Friends of mine in the NixOS-Community would likely point out that `bash` would need to be resolved by `/usr/bin/env` there too. IIRC their `/bin` only contains `sh` (and another link to/copy of `env`) but that may well have changed in the mean-time.

ethomson: Right. The other option, of course, is that we get rid of the bashisms entirely which is also good.

ethomson:
Thanks for the fix!

PR #6583, opened 2023-06-22
midx: allow unknown chunk ids in multi-pack index files

carlosmn:
These chunks work like extensions where it's fine not to know what one means. We can skip over it and keep processing the file instead of erroring out.

---

I think there's probably a different bug here as well where erroring out of the midx means we don't really keep looking, as the bug I found here is that having a midx with REIDX, aka reverse index meant that we couldn't find an object, but IIUC we should have been able to fall back on to the per-pack index there.


ethomson:
Thanks!

PR #6584, opened 2023-06-22
pack: cast the number of objects to size_t

carlosmn:
Similar to a previous change where we had to change the casting when loading the index file, we also need to make sure we don't restrict the numbers to 32-bit when looking up objects in packfiles.

This was done about three years ago in git itself, but we never got he update in this library.

---

This is the potential bug I mention in #6583 where failing to parse the multi-pack index we then failed to parse the regular index as well. This is what git did in https://github.com/git/git/commit/f86f769550ef7fb5162a9cd4be5e2be7981d063b


ethomson:
Makes sense -- thanks!

Issue #16, opened 2010-11-25
Build Error on x64


joefeser:
I am not an expert in c++, but I figured I would ask to see if there is something that I am doing wrong.

The configure worked fine, it is the build-shared that is failing.

C:\rootdir\libgit2>python waf build-shared
Waf: Entering directory `C:\rootdir\libgit2\build\shared'
[ 1/23] c: src\blob.c -> build\shared\src\blob.c.0.o
[ 2/23] c: src\delta-apply.c -> build\shared\src\delta-apply.c.0.o
blob.c
delta-apply.c
[ 3/23] c: src\commit.c -> build\shared\src\commit.c.0.o
commit.c
[ 4/23] c: src\hash.c -> build\shared\src\hash.c.0.o
hash.c
..\..\src\commit.c(291) : warning C4101: 'short_message' : unreferenced local variable
[ 5/23] c: src\hashtable.c -> build\shared\src\hashtable.c.0.o
[ 6/23] c: src\errors.c -> build\shared\src\errors.c.0.o
errors.c
hashtable.c
[ 7/23] c: src\index.c -> build\shared\src\index.c.0.o
[ 8/23] c: src\filelock.c -> build\shared\src\filelock.c.0.o
index.c
filelock.c
..\..\src\index.c(277) : warning C4244: '=' : conversion from '__time64_t' to 'unsigned int', possible loss of data
..\..\src\index.c(278) : warning C4244: '=' : conversion from '__time64_t' to 'unsigned int', possible loss of data
..\..\src\index.c(286) : warning C4244: '=' : conversion from '__int64' to 'unsigned int', possible loss of data
[ 9/23] c: src\fileops.c -> build\shared\src\fileops.c.0.o
fileops.c
[10/23] c: src\oid.c -> build\shared\src\oid.c.0.o
oid.c
[11/23] c: src\person.c -> build\shared\src\person.c.0.o
person.c
[12/23] c: src\odb.c -> build\shared\src\odb.c.0.o
odb.c
C:\rootdir\libgit2\src\git/zlib.h(5) : fatal error C1083: Cannot open include file: 'zlib.h': No such file or directory
Waf: Leaving directory`C:\rootdir\libgit2\build\shared'
Build failed
 -> task failed (exit status 2):
        {task 43901896L: c odb.c -> odb.c.0.o}
[...]

PR #6585, opened 2023-06-22
midx: do not try to look at every object in the index

carlosmn:
Similar to previous issues around doing way too much verification at loading time, checking whether the object index chunk is monotonic is better left for git-fsck instead of every single time we want to look at something in the repository.

As midx files grow, this starts taking more and more times. As an example, I went looking for this because it's taking about 1.5s to do a single object lookup in a repository that's ended up with a 7G multi-pack-index file.

----

For comparison without this loop we're down to 40ms. There's another loop for the fanout that I've left since `perf` didn't show pretty much any of `git2 cat-file` and it's not that terrible to have a little bit of checking to see that we didn't load complete nonsense.

### DIFF
--- a/src/libgit2/midx.c
+++ b/src/libgit2/midx.c
@@ -114,8 +114,6 @@ static int midx_parse_oid_lookup(
 		const unsigned char *data,
 		struct git_midx_chunk *chunk_oid_lookup)
 {
-	uint32_t i;
-	unsigned char *oid, *prev_oid, zero_oid[GIT_OID_MAX_SIZE] = {0};
 	size_t oid_size = git_oid_size(idx->oid_type);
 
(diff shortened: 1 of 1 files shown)


carlosmn:
This is in the same vein as #16

ethomson:
I wonder if midx sanity checking should be opt-in instead of removed entirely? 🤔

ethomson:
It looks like we don't have obvious plumbing to add a flag to do validation during parsing. Seems like a nice-to-have feature in the future (so that one could implement a `git fsck` like functionality) but there's probably A Lot To Do across the library to make that happen.

carlosmn:
Indeed this is a generation or fsck feature, rather than something that would make a lot of sense when you're hitting the open function for this data.

PR #6597, opened 2023-07-17
proxy: Return an error for invalid proxy URLs instead of crashing.

lrm29:
I see the following behaviour in 1.7.0 but not in 1.5.0.

If the user's Git config specifies the proxy like this then libgit2 now crashes:

```
[http]
    proxy = myhost:1234
```

The scheme (and port) are required but many users leave it off (curl defaults to http I read in the man pages).

In this pull request I propose checking the validity of the url after parsing it. Something similar is done in winhttp.c, so I replaced that with a call to git_net_url_valid but not sure if that's the right thing to do.

### DIFF
--- a/src/libgit2/transports/http.c
+++ b/src/libgit2/transports/http.c
@@ -334,16 +334,13 @@ static int lookup_proxy(
 		return 0;
 	}
 
-	if (!proxy)
-		goto done;
-
-
-	if ((error = git_net_url_parse(&transport->proxy.url, proxy) < 0))
+	if (!proxy ||
+	    (error = git_net_url_parse(&transport->proxy.url, proxy)) < 0)
 		goto done;
 
 	if (!git_net_url_valid(&transport->proxy.url)) {
-		git_error_set(GIT_ERROR_HTTP, "invalid proxy url: %s", proxy);
-		error = GIT_EINVALIDSPEC;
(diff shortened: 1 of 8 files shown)


ethomson:
Yikes. Thanks for this - we should definitely _not_ crash.

However, that's the common format that proxies are usually specified in, in my experience. We shouldn't be so rigid as to require it to be a URL -- especially in the environment variable and configuration options.

I added a few commits on top of this to relax parsing rules -- would you mind taking a look? https://github.com/libgit2/libgit2/tree/ethomson/proxy

ethomson:
> Did you intentionally not call git_net_url_parse_http in winhttp.c?

Doh. It was not intentional, and I pushed an update, but I think that you beat me to it. Added in https://github.com/libgit2/libgit2/commit/38b60fd062c245b2efbb5079f1a2c868c915207d but I must have seen that after you pulled. 🙁 

> I don't know if this generally works with SOCKS proxies at all (https://github.com/git/git/commit/6d7afe07f29df75f831a46fb0f657fa37e561779)? If so would a test with a scheme that's not empty or "http" be worth adding to your parser tests?

It does not - I haven't seen a SOCKS proxy in the wild in a _very_ long time, nor has anybody asked (yet).

ethomson:
Cool. I'm going to merge this into main and tag this to backport into the v1.7 branch. If you have any feedback from having it in your codebase, that would be 💯

PR #6631, opened 2023-09-10
add dl to LIBGIT2_SYSTEM_LIBS

christopherfujino:
Fixes https://github.com/libgit2/libgit2/issues/6632

Arguably the real issue here is that the `FindOpenSSL.cmake` module should have included this in the `OPENSSL_CRYPTO_LIBRARIES` variable: https://github.com/Kitware/CMake/blob/master/Modules/FindOpenSSL.cmake#L544


cmake/SelectHTTPSBackend.cmake
```diff
@@ -54,7 +54,7 @@ if(USE_HTTPS)
 
 		set(GIT_OPENSSL 1)
 		list(APPEND LIBGIT2_SYSTEM_INCLUDES ${OPENSSL_INCLUDE_DIR})
-		list(APPEND LIBGIT2_SYSTEM_LIBS ${OPENSSL_LIBRARIES})
+		list(APPEND LIBGIT2_SYSTEM_LIBS ${OPENSSL_LIBRARIES} ${CMAKE_DL_LIBS})
```
christopherfujino, line 57: Since this doesn't seem to be an issue when dynamically linked, maybe this should be guarded by a check for `LINK_WITH_STATIC_LIBRARIES`?

cmake/SelectHTTPSBackend.cmake
```diff
@@ -55,6 +55,9 @@ if(USE_HTTPS)
 		set(GIT_OPENSSL 1)
 		list(APPEND LIBGIT2_SYSTEM_INCLUDES ${OPENSSL_INCLUDE_DIR})
 		list(APPEND LIBGIT2_SYSTEM_LIBS ${OPENSSL_LIBRARIES})
+		if(LINK_WITH_STATIC_LIBRARIES STREQUAL ON)
```
ethomson, line 58:
```suggestion
		# Static OpenSSL (lib crypto.a) requires libdl, include it explicitly
		if(LINK_WITH_STATIC_LIBRARIES STREQUAL ON)
```

ethomson:
Thanks for the fix!

PR #6636, opened 2023-09-19
Reject potential option injections over ssh

carlosmn:
Like git (and basically everyone else, that was a fun CVE) did a while ago, we need to reject urls that might inject options into ssh which could cause it to execute arbitrary commands instead of what we mean for it to execute.

This is on the yet-unreleased exec-based transport so a public PR should be fine.

As git does, we also reject suspicious paths which may be trying to inject options into the command we ask the remote to execute. This is out of an abundance of caution.

I'm opening this up as a draft as we should have at least a little bit of testing there.

I put the function in with the `net` ones because I couldn't think of where to put it. git puts it with some path functions but our organisation is different.

### DIFF
--- a/src/libgit2/transports/ssh_exec.c
+++ b/src/libgit2/transports/ssh_exec.c
@@ -132,25 +132,21 @@ static int get_ssh_cmdline(
 	const char *default_ssh_cmd = "ssh";
 	int error;
 
-	/* Safety check: like git, we forbid paths that look like an option as
-	 * that could lead to injection to ssh that can make us do unexpected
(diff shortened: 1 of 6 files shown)


ethomson:
Oops I fat fingered the mobile app and marked this for ready for review. 😅

carlosmn:
OK I think we're good to go now.

ethomson:
Thanks @carlosmn -- sorry for the delay here. I was thinking that `process` might make more sense? I just went ahead and pushed up a change since it's been so long. 😓

PR #6638, opened 2023-09-23
blame example: Fix support for line range in CLI

wetneb:
The -L option announced by the CLI was ignored so far.

By the way, I have been using this example binary to investigate performance differences between `git blame` and libgit2's own, and I notice that libgit2's blame is significantly slower in a lot of cases. Is this expected? Happy to open an issue about this if it is not.


ethomson:
Thanks for the fix! I'd like to tackle blame improvements around perf in 2024!

PR #6642, opened 2023-10-07
git_diff_find_similar doesn't always remove unmodified deltas

yori:
Unmodified deltas are not always removed by git_diff_find_similar() when GIT_DIFF_FIND_REMOVE_UNMODIFIED is set. It only works correctly when a rename or copy is found in the diff. Somewhat understandably, this is the only case covered by the unit tests.

Non-blob deltas also avoid removal because of the way they are excluded as rename candidates. I've added a test for this using the "submodules" fixture. Maybe there's a simpler way to test for non-blob deltas?

### DIFF
--- a/src/libgit2/diff_tform.c
+++ b/src/libgit2/diff_tform.c
@@ -718,8 +718,15 @@ static bool is_rename_source(
 	git_diff_delta *delta = GIT_VECTOR_GET(&diff->deltas, delta_idx);
 
 	/* skip things that aren't blobs */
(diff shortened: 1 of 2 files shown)


ethomson:
Thanks for this, @yori - and apologies for the delay, it took me a few read-throughs to remember the diff code and how it was all working together. This looks good to me. I think that we should pull the test out of the target handling, though, and into its own function.

(Since this took me so long to review, I just went ahead and made that change on top of your PR, I hope you don't mind. 🙏 )

Thanks again!

PR #6654, opened 2023-11-09
Fix handling of paths with white spaces on diff parsing

Nfsaavedra:
## Context

While I was using [pygit2](https://github.com/libgit2/pygit2) I got a _segmentation fault_ when trying to use [`pygit2.Repository.apply`](https://www.pygit2.org/repository.html#pygit2.Repository.apply) with a diff obtained by using [`parse_diff`](https://www.pygit2.org/diff.html#pygit2.Diff.parse_diff). The diff contained a removal of a file with spaces on its path and the path was not quoted: 

`diff --git a/.run/Run IDE with Plugin.run.xml b/.run/Run IDE with Plugin.run.xml`

What I found was that when I transformed the diff obtained from [`parse_diff`](https://www.pygit2.org/diff.html#pygit2.Diff.parse_diff) back to a string, the previous line was changed to:

`diff --git a/.run/Run IDE with Plugin.run.xml /dev/null`

For some reason that I still don't fully understand, but that makes some sense to me, the `/dev/null` created a _segmentation fault_ on the apply. I looked into [pygit2](https://github.com/libgit2/pygit2) and found out that the problem probably was on `libgit2`.

## Changes

I was able to debug and find out that the function `header_path_len` was the culprit. Since previously this function stopped on a white space if the path is not quoted, then the header I gave as example would be incorrectly parsed. Given this, I changed the function so it only stops when it reaches ` b/` or ` "b/`. This still has a problem if a folder called ` b` (white space followed by b) or ` "b` is used but it seems pretty unlikely.

## Tests

I created a test for delete mode since a test for create mode already existed. However I found it weird that the tests expected that on create mode the result for the old file path was NULL. The reason I found it weird is that for "normal" paths, i.e. paths without spaces, the result was the path of the file instead of NULL. For this reason, I also changed the test for create mode to go accordingly with the behavior of "normal" paths.

### DIFF
--- a/src/libgit2/patch_parse.c
+++ b/src/libgit2/patch_parse.c
@@ -45,14 +45,27 @@ static int git_parse_err(const char *fmt, ...)
 	return -1;
 }
 
-static size_t header_path_len(git_patch_parse_ctx *ctx)
+
+static size_t header_path_len(git_patch_parse_ctx *ctx, bool last_path)
 {
 	bool inquote = 0;
 	bool quoted = git_parse_ctx_contains_s(&ctx->parse_ctx, "\"");
 	size_t len;
+	size_t max_len = last_path ? ctx->parse_ctx.line_len :
+	                       	     ctx->parse_ctx.line_len - 3;
 
-	for (len = quoted; len < ctx->parse_ctx.line_len; len++) {
-		if (!quoted && git__isspace(ctx->parse_ctx.line[len]))
+	for (len = quoted; len < max_len; len++) {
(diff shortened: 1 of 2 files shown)


arroz:
We have a patch waiting for what I think is the same problem: https://github.com/libgit2/libgit2/pull/6295 I had missed the review there, just committed the suggested code.

Nfsaavedra:
> We have a patch waiting for what I think is the same problem: #6295 I had missed the review there, just committed the suggested code.

Oops forgot to check that! I'll close this PR

PR #6656, opened 2023-11-13
Remove check for empty email or name

proller:
Some commits in public repos have no name or email and we cant re-create them with this check

	$ git clone https://github.com/ClickHouse/ClickHouse.git
	$ cd ClickHouse
	$ git cat-file -p 8b0d48f4789b8f54a19c25cecd9205e2ecca88e8
	$ git cat-file -p 54e7115ead8661bad33f8be879548270b4bdb400


ethomson:
We can't trivially remove the tests, since they enforce the correct behavior of data being created, as well as being parsed. (You can see that from the failing tests in CI.)

We need to be a bit more thoughtful about how we parse data, I agree, but it's going to be a big dig.

Issue #6655, opened 2023-11-11
1.7.1: self test failures on Cygwin


fd00:
### Reproduction steps
```
$ wget https://github.com/libgit2/libgit2/archive/refs/tags/v1.7.1.tar.gz
$ tar xf v1.7.1.tar.gz
$ cd libgit2-1.7.1
$ cmake -G Ninja -DBUILD_CLI:BOOL=OFF -DBUILD_EXAMPLES:BOOL=ON -DREGEX_BACKEND:STRING=pcre2 -DUSE_SSH:BOOL=ON .
$ ninja
$ ninja test
```

https://github.com/cygwin/scallywag/actions/runs/6831370933/job/18580898875

### Expected behavior

All test pass

### Actual behavior

```
Test project /cygdrive/d/a/scallywag/playground/libgit2-1.7.1-1.x86_64/build
      Start  1: offline
 1/12 Test  #1: offline ..........................***Failed  453.91 sec
      Start  2: invasive
 2/12 Test  #2: invasive .........................***Failed    1.52 sec
      Start  3: online
 3/12 Test  #3: online ...........................   Passed   39.53 sec
      Start  4: online_customcert
 4/12 Test  #4: online_customcert ................   Passed    2.41 sec
      Start  5: gitdaemon
 5/12 Test  #5: gitdaemon ........................   Passed    6.19 sec
      Start  6: gitdaemon_namespace
 6/12 Test  #6: gitdaemon_namespace ..............   Passed    0.11 sec
      Start  7: gitdaemon_sha256
 7/12 Test  #7: gitdaemon_sha256 .................   Passed    0.11 sec
      Start  8: ssh
 8/12 Test  #8: ssh ..............................   Passed    7.01 sec
      Start  9: proxy
 9/12 Test  #9: proxy ............................   Passed    5.20 sec
      Start 10: auth_clone
10/12 Test #10: auth_clone .......................   Passed    1.76 sec
      Start 11: auth_clone_and_push
11/12 Test #11: auth_clone_and_push ..............   Passed    6.11 sec
      Start 12: util
12/12 Test #12: util .............................***Failed    1.05 sec

75% tests passed, 3 tests failed out of 12

Total Test time (real) = 524.93 sec

The following tests FAILED:
          1 - offline (Failed)
          2 - invasive (Failed)
         12 - util (Failed)
Errors while running CTest
[...]

PR #6659, opened 2023-11-17
util: suppress some uninitialized variable warnings

boretrk:
A couple of warnings were uncovered in #6655
This should take care of the easy ones.

The larger problem is the ctype warnings where we have a bit of a mix between using the proper `isalnum()` and rolling our own with `git__isalnum()`
The posix versions take an integer in to be able to handle EOF from `fgetc()` so when a char is passed it has to be casted to unsigned so that 128-255 isn't extended to negative integers.
The `git__isalnum()` function takes care of it by simply not handling characters outside of the ascii set and will behave incorrectly on systems using ISO-8859-1 or other charsets using more than 7 bits.

I'm not entirely sure why we have the `git__isalnum()` functions. I assume it is to support systems where `ctype.h` isn't available?
In that case I think the current implementation should be used only when ctype doesn't exists.
There is also the thing where casting characters to unsigned everywhere is a bit tedious, so maybe we should change the functions to take an unsigned char in?

### DIFF
--- a/src/util/fs_path.c
+++ b/src/util/fs_path.c
@@ -1938,12 +1938,13 @@ static int sudo_uid_lookup(uid_t *out)
 {
 	git_str uid_str = GIT_STR_INIT;
 	int64_t uid;
-	int error;
+	int error = -1;
 
-	if ((error = git__getenv(&uid_str, "SUDO_UID")) == 0 &&
-	    (error = git__strntol64(&uid, uid_str.ptr, uid_str.size, NULL, 10)) == 0 &&
-	    uid == (int64_t)((uid_t)uid)) {
+	if (git__getenv(&uid_str, "SUDO_UID") == 0 &&
+		git__strntol64(&uid, uid_str.ptr, uid_str.size, NULL, 10) == 0 &&
+		uid == (int64_t)((uid_t)uid)) {
 		*out = (uid_t)uid;
+		error = 0;
 	}
 
 	git_str_dispose(&uid_str);
--- a/src/util/futils.h
+++ b/src/util/futils.h
@@ -25,7 +25,7 @@ extern int git_futils_readbuffer(git_str *obj, const char *path);
 extern int git_futils_readbuffer_updated(
 	git_str *obj,
 	const char *path,
-	unsigned char checksum[GIT_HASH_SHA1_SIZE],
+	unsigned char checksum[GIT_HASH_SHA256_SIZE],
 	int *updated);
 extern int git_futils_readbuffer_fd_full(git_str *obj, git_file fd);
 extern int git_futils_readbuffer_fd(git_str *obj, git_file fd, size_t len);
--- a/src/util/regexp.c
+++ b/src/util/regexp.c
@@ -125,7 +125,7 @@ int git_regexp_search(const git_regexp *r, const char *string, size_t nmatches, 
 
 	if ((data = pcre2_match_data_create(nmatches, NULL)) == NULL) {
 		git_error_set_oom();
-		goto out;
+		return -1;
 	}
 
 	if ((error = pcre2_match(*r, (const unsigned char *) string, strlen(string),
--- a/src/util/errors.c
+++ b/src/util/errors.c
@@ -376,6 +376,9 @@ void git_error_system_set(int code)
 /* Deprecated error values and functions */
 
 #ifndef GIT_DEPRECATE_HARD
+
+#include "git2/deprecated.h"
+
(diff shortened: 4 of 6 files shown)


boretrk:
Also replaced index() with strchr(). index() isn't POSIX and won't compile on some systems.
I also got some compiler warnings in errors.c because the deprecated stuff weren't pre-declared.

src/util/futils.h
```diff
@@ -25,7 +25,7 @@ extern int git_futils_readbuffer(git_str *obj, const char *path);
 extern int git_futils_readbuffer_updated(
 	git_str *obj,
 	const char *path,
-	unsigned char checksum[GIT_HASH_SHA1_SIZE],
+	unsigned char checksum[GIT_HASH_SHA256_SIZE],
```
ethomson, line 28:
```suggestion
	unsigned char checksum[GIT_HASH_MAX_SIZE],
```

src/util/fs_path.c
```diff
@@ -1956,7 +1956,8 @@ int git_fs_path_owner_is(
 	git_fs_path_owner_t owner_type)
 {
 	struct stat st;
-	uid_t euid, sudo_uid;
+	/* Initialize sudo_uid to invalid uid_t to suppress warning */
+	uid_t euid, sudo_uid = (uid_t)-1;
```
ethomson, line 1960: I think that the compiler is right here. In `sudo_uid_lookup`, we can return `0` but not set `*out`.

```suggestion
	uid_t euid, sudo_uid;
```

I think that we should change `sudo_uid_lookup` instead, something like:

```c
static int sudo_uid_lookup(uid_t *out)
{
	git_str uid_str = GIT_STR_INIT;
	int64_t uid;
	int error = -1;

	if (git__getenv(&uid_str, "SUDO_UID") == 0 &&
	    git__strntol64(&uid, uid_str.ptr, uid_str.size, NULL, 10) == 0 &&
	    uid == (int64_t)((uid_t)uid)) {
		*out = (uid_t)uid;
		error = 0;
	}

	git_str_dispose(&uid_str);
	return error;
}
```

src/util/regexp.c
```diff
@@ -125,7 +125,7 @@ int git_regexp_search(const git_regexp *r, const char *string, size_t nmatches,
 
 	if ((data = pcre2_match_data_create(nmatches, NULL)) == NULL) {
 		git_error_set_oom();
-		goto out;
+		return GIT_ERROR;
```
ethomson, line 128: For consistency, we usually just use `-1`, even though we have `GIT_ERROR`. 

```suggestion
		return -1;
```

src/util/futils.h
boretrk: Currently `git_futils_readbuffer_updated()` is very strict in its usage of SHA256.
It calculates a new hash using `GIT_HASH_ALGORITHM_SHA256` and updates the passed checksum buffer.
Changing it to `GIT_HASH_MAX_SIZE` should be "safe" since its always larger, but calling it with any checksum other than SHA256 will fail.
Maybe we should pass in a `git_hash_ctx *` instead and make `git_futils_readbuffer_updated()` sensitive to the type of hash used?

From what I can tell it is only used by `git_grafts_refresh()` that uses SHA256 to track changes in a private struct unrelated to what the repository uses.

ethomson: Oops, yes, you're absolutely right. Thanks for the clarification.

boretrk:
OK, I think this one is good to go now.
The character classification stuff will probably have to be its own thing.

The `git__isalpha()` stuff seems to have been added in 0f49200c9a to solve issues with the config parsing so it should probably be left alone?
Over all it seems like we need a set of character type functions that behaves as if locale is set to `"C"` but is this really always the case?
So when is it correct to use locale sensitive character type classification and when shall we use the git__xxxx() versions?

For reference: When locale is set to "C" we get ISO/IEC 9899:1999 5.2.1 Character sets:
> 3) Both the basic source and basic execution character sets shall have the following members: the 26 uppercase letters of the Latin alphabet
A B C D E F G H I J K L M N O P Q R S T U V W X Y Z
the 26 lowercase letters of the Latin alphabet
a b c d e f g h i j k l m n o p q r s t u v w x y z
the 10 decimal digits
0 1 2 3 4 5 6 7 8 9
the following 29 graphic characters
! " # % & ' ( ) * + , - . / : ; < = > ? [ \ ] ^ _ { | } ~
the space character, and control characters representing horizontal tab, vertical tab, and form feed.

ethomson (approved):

PR #6660, opened 2023-11-21
Make xdiff unbundlable [DO NOT MERGE]

georgthegreat:
This is an MVP to demonstrate how xdiff can work without libgit2 and yet remain pretty portable.
It

* removes `git__` typedefs for malloc functions
* reimplements certain attributes
* removed regexp functionality as libgit2 does not use it

Continues #6437

### DIFF
--- a/deps/xdiff/git-xdiff.h
+++ b/deps/xdiff/git-xdiff.h
@@ -14,7 +14,9 @@
 #ifndef INCLUDE_git_xdiff_h__
 #define INCLUDE_git_xdiff_h__
 
-#include "regexp.h"
+#include <limits.h>
+#include <stdint.h>
+#include <stddef.h>
 
 /* Work around C90-conformance issues */
 #if !defined(__STDC_VERSION__) || (__STDC_VERSION__ < 199901L)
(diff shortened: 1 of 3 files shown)


ethomson:
The goal of https://github.com/libgit2/xdiff is that _it_ is the thing that you consume. For example, it already has the indirection around malloc.

This allows libgit2 to keep using its pluggable malloc everywhere, including xdiff (this is a _must_ for us) but other consumers can use whatever malloc they want.

If you have portability changes, suggest them in that repo, please.

georgthegreat:
I do not quite understand the _pluggable malloc_ idea, btw.

Allocators in C / C++ world provide ABI, not API, hence they are pluggable out of the box.
One just provides `extern "C" malloc()` implementation and links with the library.
This protocol is implemented by every known allocator including jemalloc, tcmalloc, mimalloc and more.

Why libgit2 uses custom protocol?

georgthegreat:
This is the regexp.h part only:
https://github.com/libgit2/libgit2/pull/6670

ethomson:
libgit2 runs in a wide variety of places - for example, rugged, the ruby bindings, use `xmalloc` and `xfree`, which are special ruby FFI functionality to allocate memory that can be tracked in the GC system. https://github.com/libgit2/rugged/blob/0b10975821f3ffbdd83b3d7176040d7cbab0d900/ext/rugged/rugged_allocator.c#L11

(as opposed to normal `malloc` and `free` which are the system allocators).

georgthegreat:
Well, you do not patch zlib with these custom allocations, do you?

Once we have placed xdiff of the same level as zlib, it should not use this approach.

ethomson:
Closing this — libgit2 isn't the place for xdiff changes

PR #6662, opened 2023-12-06
fetch: enable deepening/shortening shallow clones

kempniu:
A shallow repository can currently only be completely unshallowed, which is caused by mark_local() only marking locally-existing objects as wanted if the fetch depth is set to `INT_MAX` (`GIT_FETCH_DEPTH_UNSHALLOW`). This prevents deepening the history of a shallow clone to an arbitrary number of commits, which may be preferable over full unshallowing for large repositories.

Enable deepening and shortening shallow clones by marking locally-existing objects as wanted whenever the fetch depth is set to any non-default value (either `GIT_FETCH_DEPTH_UNSHALLOW` or an arbitrary positive integer).

### DIFF
--- a/tests/libgit2/online/shallow.c
+++ b/tests/libgit2/online/shallow.c
@@ -164,3 +164,102 @@ void test_online_shallow__unshallow(void)
 	git_revwalk_free(walk);
 	git_repository_free(repo);
 }
+
+void test_online_shallow__deepen_six(void)
+{
+	git_str path = GIT_STR_INIT;
+	git_repository *repo;
+	git_revwalk *walk;
+	git_clone_options clone_opts = GIT_CLONE_OPTIONS_INIT;
+	git_fetch_options fetch_opts = GIT_FETCH_OPTIONS_INIT;
+	git_remote *origin = NULL;
+	git_oid oid;
+	git_oid *roots;
+	size_t roots_len;
+	size_t num_commits = 0;
+	int error = 0;
+
+	clone_opts.fetch_opts.depth = 5;
+	clone_opts.remote_cb = remote_single_branch;
+
+	git_str_joinpath(&path, clar_sandbox_path(), "deepen_6");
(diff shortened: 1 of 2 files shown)


tests/libgit2/online/shallow.c
```diff
@@ -164,3 +164,102 @@ void test_online_shallow__unshallow(void)
...
+	cl_assert_equal_i(error, GIT_ITEROVER);
+
+	git__free(roots);
+	git_remote_free(origin);
+	git_str_dispose(&path);
+	git_revwalk_free(walk);
+	git_repository_free(repo);
+}
+
+void test_online_shallow__shorten_four(void)
```
ethomson, line 218:
```suggestion
void test_online_shallow__shorten_five(void)
```

It looks like this actually deepens five not four?

ethomson:
Thanks for the fix! Sorry for the delay, I had to read the surrounding comment a few times to remember how this all works. 😅 

One minor question about the name of the test - otherwise looks great.

ethomson (approved):

tests/libgit2/online/shallow.c
kempniu: Both `test_online_shallow__deepen_six()` and `test_online_shallow__shorten_four()` start by cloning the test repository with a depth of 5 (`clone_opts.fetch_opts.depth = 5;`).

`test_online_shallow__deepen_six()` then fetches from the repository again, but with a depth of 6 (`fetch_opts.depth = 6;`), to demonstrate that libgit2 is able to make a shallow history *deeper*.

`test_online_shallow__shorten_four()`, which you called out in your comment, performs the post-cloning fetch with a depth of 4 (`fetch_opts.depth = 4;`), to demonstrate that libgit2 is also able to make a shallow history *shorter*.

Does this help?

ethomson: Oops. 😓  Yes, sorry, I misread that.

ethomson:
Thanks @kempniu !

Issue #6649, opened 2023-10-23
Performance issue with repository owner check


csware:
The owner check https://github.com/libgit2/libgit2/blob/45fd9ed7ae1a9b74b957ef4f337bc3c8b3df01b5/src/libgit2/repository.c#L545 adds a significant performance issue, because the path normalization `git_fs_path_prettify_dir` https://github.com/libgit2/libgit2/blob/45fd9ed7ae1a9b74b957ef4f337bc3c8b3df01b5/src/libgit2/repository.c#L586 is called for every `safe.directory` value in .gitconfig.

### Reproduction steps

Have many `safe.directory` entries in .gitconfig, potentially several pointing to unavailable UNC-paths such as `%(prefix)///192.168.123.123/Foo/bar/repo1` or NFS shares.

libgit2 will contact each of those for path validation.

### Version of libgit2 (release number or SHA1)

All since https://github.com/libgit2/libgit2/commit/be3a78cc933169bab12dccdfcbb173cb48261103

### Operating system(s) tested

Windows 10

csware:
cross-ref: https://tortoisegit.org/issue/4073

csware:
Fixed using 25e2b9d8c217e92eb70cde85bb4007a3dec0428d and finally using https://github.com/libgit2/libgit2/pull/6739.

PR #6668, opened 2023-12-11
Do not normalize safe.directory paths

csware:
Vanilla Git does not do it either (cf. https://github.com/gitster/git/blob/master/setup.c#L1150) and calling git_fs_path_prettify_dir can cause performance issues (cf. issue #6649).

Still needed: expansion of "~/"... (but not this PR)

### DIFF
--- a/src/libgit2/repository.c
+++ b/src/libgit2/repository.c
@@ -685,9 +685,26 @@ static int validate_ownership(git_repository *repo)
 		goto done;
 
 	if (!is_safe) {
-		git_error_set(GIT_ERROR_CONFIG,
-			"repository path '%s' is not owned by current user",
-			path);
+		git_str nice_path = GIT_STR_INIT;
+#ifdef GIT_WIN32
+		/* see comment above in validate_ownership_cb */
+		if (!strncasecmp(path, "//", strlen("//")))
+			git_str_puts(&nice_path, "%(prefix)/");
+#endif
+		git_str_puts(&nice_path, path);
+		if (!git_str_oom(&nice_path)) {
+			if (git_str_len(&nice_path) > 1 && nice_path.ptr[git_str_len(&nice_path) - 1] == '/')
+				git_str_shorten(&nice_path, 1);
+			git_error_set(
+			        GIT_ERROR_CONFIG,
+			        "repository path '%s' is not owned by current user.\n\nTo add an exception use the path '%s'.",
+			        path, nice_path.ptr);
+		} else
+			git_error_set(
+			        GIT_ERROR_CONFIG,
+			        "repository path '%s' is not owned by current user.",
+			        path);
+		git_str_dispose(&nice_path);
 		error = GIT_EOWNER;
 	}
 
--- a/src/libgit2/repository.c
+++ b/src/libgit2/repository.c
@@ -537,8 +537,7 @@ static int read_gitfile(git_str *path_out, const char *file_path)
 }
 
 typedef struct {
-	const char *repo_path;
-	git_str tmp;
+	git_str repo_path;
 	bool *is_safe;
 } validate_ownership_data;
 
@@ -582,9 +581,7 @@ static int validate_ownership_cb(const git_config_entry *entry, void *payload)
 		         strncmp(test_path, "//wsl.localhost/", strlen("//wsl.localhost/")) != 0)
 			test_path++;
 #endif
-
-		if (git_fs_path_prettify_dir(&data->tmp, test_path, NULL) == 0 &&
-		    strcmp(data->tmp.ptr, data->repo_path) == 0)
+		if (strcmp(test_path, data->repo_path.ptr) == 0)
 			*data->is_safe = true;
 	}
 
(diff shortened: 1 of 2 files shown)


src/libgit2/repository.c
```diff
@@ -582,10 +582,14 @@ static int validate_ownership_cb(const git_config_entry *entry, void *payload)
 		         strncmp(test_path, "//wsl.localhost/", strlen("//wsl.localhost/")) != 0)
 			test_path++;
 #endif
-
-		if (git_fs_path_prettify_dir(&data->tmp, test_path, NULL) == 0 &&
-		    strcmp(data->tmp.ptr, data->repo_path) == 0)
+		git_str_sets(&tmp, data->repo_path);
```
ethomson, line 585: We were passing the `git_str` here to avoid re-allocing a `git_str` on every call. We can just re-use the one we have. No need to undo that, I don't think.

src/libgit2/repository.c
```diff
@@ -582,10 +582,14 @@ static int validate_ownership_cb(const git_config_entry *entry, void *payload)
 		         strncmp(test_path, "//wsl.localhost/", strlen("//wsl.localhost/")) != 0)
 			test_path++;
 #endif
-
-		if (git_fs_path_prettify_dir(&data->tmp, test_path, NULL) == 0 &&
```
ethomson, line 586: I think that instead of `prettify`, we need only call `path_to_dir`.

src/libgit2/repository.c
ethomson: Also, you're looking at `data->repo_path`, not `test_path`, which we've mutated to handle WSL paths. We should keep that.

ethomson:
Agree that we don't need to hit the filesystem on every excluded path. I think that the simplest fix is:

```diff
diff --git a/src/libgit2/repository.c b/src/libgit2/repository.c
index 05ece6efc..9fa616fb0 100644
--- a/src/libgit2/repository.c
+++ b/src/libgit2/repository.c
@@ -583,8 +583,11 @@ static int validate_ownership_cb(const git_config_entry *entry, void *payload)
                        test_path++;
 #endif

-               if (git_fs_path_prettify_dir(&data->tmp, test_path, NULL) == 0 &&
-                   strcmp(data->tmp.ptr, data->repo_path) == 0)
+               if (git_str_sets(&data->tmp, test_path) < 0 ||
+                   git_fs_path_to_dir(&data->tmp) < 0)
+                       return -1;
+
+               if (strcmp(data->tmp.ptr, data->repo_path) == 0)
                        *data->is_safe = true;
        }

```

This keeps the `git_str` that is used between invocations of the callback to avoid re-allocs, but switches from `prettify_path` to `path_to_dir`, which should avoid touching the filesystem and merely append a `/` since that's our (sort of dubious) pattern for directories.

This does not add the trailing `/` check on inputs. I don't know how valuable that is - I think that specifying a folder with a trailing slash is perfectly legitimate and that git should allow it. But if you're bullish on exact compatibility here, I think that's a reasonable opinion as well.

ethomson:
Thanks @csware for digging in here to understand the perf problem - it's slow for me to develop / test on Windows - and apologies for causing it in the first place.

src/libgit2/repository.c
csware: I don't think that using `path_to_dir` is the right way. This would accept `safe.directory` entries ending with a slash which vanilla Git does not accept.

ethomson: Is that actually a problem, though? We also accept `//wsl.localhost/...` prefixed paths which Git for Windows doesn't accept.

csware: From my point of view, I would keep it as consistent as possible.

csware: Let's do one fix at the time. It's ok for me if we take your suggested patch.

csware:
@ethomson I'd like to talk to you about libgit2, some open PRs and TortoiseGit. Would be nice if we could meet next year.

src/libgit2/repository.c
ethomson: > From my point of view, I would keep it as consistent as possible.

OK. Let's do that. This feels like a (minor) bug in git that it doesn't allow trailing slashes, but I've often said that libgit2's goal is to be "bug for bug compatible".

csware: I updated the branch and hope that I did not forget anything.

csware: @ethomson 
I updated the branch and hope that I did not forget anything.

csware:
@ethomson 
Is there an open point here?

ethomson:
Manually merged with a tweak - thanks again!

csware:
This is also relevant for 1.7, right?!

csware:
Is there a reason why you ,merged both commits and then reverted the second one which improves the error message?

csware:
@ethomson
Is there a reason why you ,merged both commits and then reverted the second one which improves the error message?

src/libgit2/repository.c
```diff
@@ -582,9 +581,7 @@ static int validate_ownership_cb(const git_config_entry *entry, void *payload)
 		         strncmp(test_path, "//wsl.localhost/", strlen("//wsl.localhost/")) != 0)
```
csware, line 581: Remove this `else if` part if `%(prefix)/` is not required.

PR #6670, opened 2023-12-13
Make xdiff independent from regexp.h

georgthegreat:
I will backport these to libgit2/xdiff later on.

### DIFF
--- a/deps/xdiff/git-xdiff.h
+++ b/deps/xdiff/git-xdiff.h
@@ -14,7 +14,11 @@
 #ifndef INCLUDE_git_xdiff_h__
 #define INCLUDE_git_xdiff_h__
 
-#include "regexp.h"
+#include "git2_util.h"
+
+#include <limits.h>
+#include <stdint.h>
+#include <stddef.h>
 
 /* Work around C90-conformance issues */
 #if !defined(__STDC_VERSION__) || (__STDC_VERSION__ < 199901L)
@@ -35,22 +39,5 @@
 #define xdl_realloc(ptr, x) git__realloc(ptr, x)
 
 #define XDL_BUG(msg) GIT_ASSERT(!msg)
-
-#define xdl_regex_t git_regexp
-#define xdl_regmatch_t git_regmatch
-
-GIT_INLINE(int) xdl_regexec_buf(
-	const xdl_regex_t *preg, const char *buf, size_t size,
-	size_t nmatch, xdl_regmatch_t pmatch[], int eflags)
-{
-	GIT_UNUSED(preg);
-	GIT_UNUSED(buf);
-	GIT_UNUSED(size);
-	GIT_UNUSED(nmatch);
-	GIT_UNUSED(pmatch);
-	GIT_UNUSED(eflags);
-	GIT_ASSERT("not implemented");
-	return -1;
-}
 
 #endif
--- a/deps/xdiff/xdiff.h
+++ b/deps/xdiff/xdiff.h
@@ -83,10 +83,6 @@ typedef struct s_mmbuffer {
 typedef struct s_xpparam {
 	unsigned long flags;
 
-	/* -I<regex> */
-	xdl_regex_t **ignore_regex;
-	size_t ignore_regex_nr;
-
 	/* See Documentation/diff-options.txt. */
 	char **anchors;
 	size_t anchors_nr;
--- a/deps/xdiff/xdiffi.c
+++ b/deps/xdiff/xdiffi.c
@@ -1011,46 +1011,6 @@ static void xdl_mark_ignorable_lines(xdchange_t *xscr, xdfenv_t *xe, long flags)
 	}
 }
 
-static int record_matches_regex(xrecord_t *rec, xpparam_t const *xpp) {
-	xdl_regmatch_t regmatch;
-	int i;
-
-	for (i = 0; i < xpp->ignore_regex_nr; i++)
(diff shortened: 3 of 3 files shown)


ethomson:
My understanding is that your goal is produce an xdiff library, and then link libgit2 to it, and maybe something else as well.

In that case, what you need to do is create your own version of https://github.com/libgit2/xdiff, and create your own version of `git-xdiff.h`. That's what we've done in libgit2, and that's what we hope that https://github.com/git/git will do as well.

If you don't support regex, that's fine, then your `git-xdiff.h` should look something like:

```c
#define xdl_regex_t void *
 #define xdl_regmatch_t void *

 GIT_INLINE(int) xdl_regexec_buf(
 	const xdl_regex_t *preg, const char *buf, size_t size,
 	size_t nmatch, xdl_regmatch_t pmatch[], int eflags)
 {
 	return -1;
 }
```

But don't remove regex support from xdiff. git uses it (and we hope this is a shared dependency), and we might in the future.

georgthegreat:
I can have any version of xdiff and libgit2 indeed, I just _do not want to_ become a maintainer of yet another fork with evergrowing difference from its upstream (I assume libgit2/xdiff to be current upstream as there is no better option here).

libgit2 does not use regexp functionality from xdiff and regexp.h is not portable, so this can be considered as a portability fix.

ethomson:
I don't think I understand... I mean, you also don't have our allocators do you?

georgthegreat:
We do not have _that_ custom allocators, we just have jemalloc / tcmalloc / mimalloc.

But this PR is not related to allocators at all, it just removes unused and unportable functionality.

ethomson:
I simply don't understand what it is that you're trying to do.

I created https://github.com/libgit2/xdiff for you, with no dependencies on libgit2. The xdiff _in libgit2_ has dependencies _on libgit2_. Now you're in libgit2, trying to make xdiff remove its dependencies _on libgit2_.

Our xdiff _uses our code_ because _of course it does_. git's xdiff _uses their code_ because _of course it does_.

What's the point of https://github.com/libgit2/xdiff, except _to be a place that has xdiff that doesn't use libgit2's code_. 🥴 

Please help me understand what it is that you want? I thought you wanted to be able to build an xdiff that was shared between your build of libgit2 and something else? Why are you making any changes to the xdiff in this source tree if that's the case?

georgthegreat:
I suggest the following:

1. (publicly) Make libgit2/xdiff repo independent from libgit2 itself
2. (publicly) Make difference between libgit2/libgit2/deps/xdiff as minimal from libgit2/xdiff as possible
3. (internally?) Add a small patch over libgit2/xdiff to make it fully independent from libgit2/xdiff

Link both libgit2 and mercurial to xdiff from (3).

ethomson:
Thanks - just thinking through this. Let's start on point 1 - is it not independent today?

georgthegreat:
It is independent yet it is non portable.

At the time it includes regex.h which is not present on Windows.

ethomson:
> It is independent yet it is non portable.
>
> At the time it includes regex.h which is not present on Windows.

OK. Let's do a lighter lift, though.

Right now xdiff's upstream is _git_, not libgit2. I want to be able to take xdiff out of git, when they make a change, and bring it to libgit2 with few changes. We went through and updated all the math to avoid integer overflows, and merging that when they made changes was near impossible. I don't want to be in a place where we're making big changes to the logic in this repo.

Ideally we'll get to a place where we're _literally_ sharing this code base and have a single point of indirection. I can pick that thread back up with the git mailing list.

In the meantime, why not something like this?

```diff
diff --git a/git-xdiff.h b/git-xdiff.h
index d812d0d..e87e60e 100644
--- a/git-xdiff.h
+++ b/git-xdiff.h
@@ -16,7 +16,6 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
-#include <regex.h>
 
 /* Work around C90-conformance issues */
 #if !defined(__STDC_VERSION__) || (__STDC_VERSION__ < 199901L)
@@ -47,8 +46,23 @@
 
 #define XDL_BUG(msg) do { fprintf(stderr, "fatal: %s\n", msg); exit(128); } while(0)
 
-#define xdl_regex_t regex_t
-#define xdl_regmatch_t regmatch_t
+#ifdef XDL_NO_REGEX
+
+# define xdl_regex_t void *
+# define xdl_regmatch_t void *
+
+inline int xdl_regex_buf(
+	const xdl_regex_t *preg, const char *buf, size_t size,
+	size_t nmatch, xdl_regmatch_t pmatch[], int eflags)
+{
+    return 15; /* REG_ASSERT */
+}
+
+#else
+# include <regex.h>
+
+# define xdl_regex_t regex_t
+# define xdl_regmatch_t regmatch_t
 
 inline int xdl_regexec_buf(
 	const xdl_regex_t *preg, const char *buf, size_t size,
@@ -60,4 +74,6 @@ inline int xdl_regexec_buf(
     return regexec(preg, buf, nmatch, pmatch, eflags | REG_STARTEND);
 }
 
+#endif /* XDL_NO_REGEX */
+
 #endif
```

(In https://github.com/libgit2/xdiff)

georgthegreat:
Ok, I put regex usage under (defined-by-default) flag:
https://github.com/libgit2/xdiff/pull/4

ethomson:
Closing this — libgit2 isn't the place for changes to xdiff.

PR #6671, opened 2023-12-14
Support setting oid type for in-memory repositories

kcsaul:
This change adds support for setting oid type when creating an in-memory repository using git_repository_new (in conjunction with using custom backends).

Needed to prevent "ambiguous lookup - OID prefix is too short" error during git_object_lookup_prefix.

### DIFF
--- a/include/git2/sys/repository.h
+++ b/include/git2/sys/repository.h
@@ -9,6 +9,7 @@
 
 #include "git2/common.h"
 #include "git2/types.h"
+#include "git2/oid.h"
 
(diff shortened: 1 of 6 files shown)


ethomson:
Great catch. Thanks for the fix.

I wonder if we should take a `git_repository_new_options` here? The name is a bit regrettable, but matches our style. This could help future-proof that API in case we want to add more things later.

ethomson:
Makes sense. Thanks again @kcsaul

PR #6683, opened 2023-12-14
Fix 'path cannot exist in repository' during diff for in-memory repository

kcsaul:
Fixes obscure bug caused by change shown below in 54a22f536557b9a25752c6b59829dff4cbee99e0.

image

For an in-memory repository without an index setup, nor any directory on-disk, git_repository_index_weakptr can now fail with a not found error returned by repository_index_path, as a consequence of this, attr_setup returns the not found error in this scenario, which then results in an error when attempting to lookup the driver for comparing a file during a diff.

Elsewhere, the not found errors are ignored within attr_setup directly, or indirectly by callers of that, but error is escaping via the diff driver lookup, hence fix added to ignore it directly within attr_setup.

### DIFF
--- a/src/libgit2/attr.c
+++ b/src/libgit2/attr.c
@@ -424,8 +424,12 @@ static int attr_setup(
 			goto out;
 
 	if ((error = git_repository_index__weakptr(&idx, repo)) < 0 ||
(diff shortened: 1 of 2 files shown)


ethomson:
Looks good - thanks!

PR #6685, opened 2023-12-15
Add Cygwin to the CI

ethomson:


fd00:
This action may be easy to implement Github Actions
- https://github.com/cygwin/cygwin-install-action

ethomson:
Yeah I saw that (and used it as an inspiration for our setup script). Thanks!

I want to keep everything in a setup script (if possible) so that it's easy to bootstrap a build on a local dev box but if there's a compelling reason to use a setup action instead I'm not opposed.

PR #6691, opened 2023-12-17
tests: remove test for strcasecmp

boretrk:
strcasecmp is a posix function, testing it doesn't make sense.
Functions that needs unsigned compare should use git__strcasecmp()

If people are using signed chars then they will probably run into some issue where sorted strings end up in a different order but this is a problem with the functions calling strcasecmp instead of git__strcasecmp, not strcasecmp itself.

### DIFF
--- a/src/util/unix/posix.h
+++ b/src/util/unix/posix.h
@@ -54,8 +54,6 @@ GIT_INLINE(int) p_fsync(int fd)
 #define p_send(s,b,l,f) send(s,b,l,f)
 #define p_inet_pton(a, b, c) inet_pton(a, b, c)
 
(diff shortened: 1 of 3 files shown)


ethomson:
This is one of those "what were we thinking!?" sort of moments. 🤔  And usually I can remember why we did something that seems silly five years down the line, but this one, I'm stumped.

ethomson:
I'm _guessing_ that this was just a "let's show that our strcasecmp works like the standard" but... that's obviously not the case in a world where strcasecmp is driven by locale (and we're not explicitly setting it to `C` or something.)

So... anyway, this all seems good to me. 🙏

PR #6701, opened 2023-12-21
documentation with troubleshooting steps updated

Olusoji01:
"Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
-- Could NOT find GSSAPI (missing: GSSAPI_LIBRARIES GSSAPI_INCLUDE_DIR) 
-- Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY OPENSSL_INCLUDE_DIR)
logo".

The error messages have been investigated and documentation with troubleshooting steps have been updated

PR #6703, opened 2023-12-21
Update troubleshooting.md

Suprithvarma1:


ethomson:
This looks like a dup of #6701 - if you two are working together, please pair up in that PR. 🙏

PR #4409, opened 2017-11-12
issue #3751 (Function to parse author and committer information from environment)

cpoerschke:
work-in-progress for issue #3751 (Function to parse author and committer information from environment)

```
cd build
cmake --build . && ./libgit2_clar -v -srepo::init
```

PR #6706, opened 2023-12-21
Use environment variables when creating signatures

u-quark:
When creating an action signature (e.g. for a commit author and committer) read the following environment variables that can override the configuration options:

 * `GIT_AUTHOR_NAME` is the human-readable name in the "author" field.
 * `GIT_AUTHOR_EMAIL` is the email for the "author" field.
 * `GIT_AUTHOR_DATE` is the timestamp used for the "author" field.
 * `GIT_COMMITTER_NAME` sets the human name for the "committer" field.
 * `GIT_COMMITTER_EMAIL` is the email address for the "committer" field.
 * `GIT_COMMITTER_DATE` is used for the timestamp in the "committer" field.
 * `EMAIL` is the fallback email address in case the user.email configuration value isn't set. If this isn't set, Git falls back to the system user and host names.

This is taken from the git documentation chapter "10.8 Environment Variables":

https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

This PR adds support for reading these environment variables by adding two new functions `git_signature_default_author` and `git_signature_default_committer` and deprecates the `git_signature_default` function.

Fixes: https://github.com/libgit2/libgit2/issues/3751

Prior work:
 * https://github.com/libgit2/libgit2/pull/4409
 * https://github.com/libgit2/libgit2/pull/5479
 * https://github.com/libgit2/libgit2/pull/6290

### DIFF
--- a/examples/commit.c
+++ b/examples/commit.c
@@ -39,7 +39,7 @@ int lg2_commit(git_repository *repo, int argc, char **argv)
 	git_index *index;
 	git_object *parent = NULL;
 	git_reference *ref = NULL;
-	git_signature *signature;
+	git_signature *author_signature, *committer_signature;
 
 	/* Validate args */
 	if (argc < 3 || strcmp(opt, "-m") != 0) {
@@ -63,21 +63,25 @@ int lg2_commit(git_repository *repo, int argc, char **argv)
 
 	check_lg2(git_tree_lookup(&tree, repo, &tree_oid), "Error looking up tree", NULL);
 
-	check_lg2(git_signature_default(&signature, repo), "Error creating signature", NULL);
+	check_lg2(git_signature_default_author(&author_signature, repo),
+			"Error creating author signature", NULL);
+	check_lg2(git_signature_default_committer(&committer_signature, repo),
+			"Error creating committer signature", NULL);
 
 	check_lg2(git_commit_create_v(
 		&commit_oid,
 		repo,
 		"HEAD",
-		signature,
-		signature,
+		author_signature,
+		committer_signature,
 		NULL,
 		comment,
 		tree,
 		parent ? 1 : 0, parent), "Error creating commit", NULL);
 
 	git_index_free(index);
-	git_signature_free(signature);
+	git_signature_free(author_signature);
+	git_signature_free(committer_signature);
 	git_tree_free(tree);
 	git_object_free(parent);
 	git_reference_free(ref);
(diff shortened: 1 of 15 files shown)


ethomson:
Thank you for the PR - I agree that we need this functionality. I took a quick first pass on the changes and overall, this is an excellent PR. I will take a closer look 🔜 — thanks again.

u-quark:
Hi, thanks for looking into this!

I am also keen on getting this functionality merged! Please let me know if you want a different approach of some specific implementation on this PR, I don't mind changing it.

I am not sure I recreated 100% the CI build environment but I managed to compile with clang with similar features enabled on cmake so hopefully this time CI will pass.

Cheers.

ethomson (changes requested):

ethomson:
A few minor tweaks requested - thanks for putting this together, I think that this is a really useful improvement. 🙏

u-quark:
Thanks of the review @ethomson!

csware:
The usage of environment variables should always be opt-in in libgit2

ethomson:
Sorry, I've been slow to look at this. One of the things that I've found is that if you want to emulate `git commit` (for example), then you really want to a) look at the environment variables, but b) fall back to the _same time_ for both author and committer if you're using the current timestamp. I think that this needs to be a single API to get both instead of individual APIs.

Per @csware 's comment, I also think that switching all the behavior of the things that produce commits internally to look at the environment variables is a Big Change that would be unexpected in a small release. I would suggest that we keep the current behavior (wherein it does not honor the `GIT_AUTHOR_` and `GIT_COMMITTER_` environment variables for now. We can potentially make this opt-in (you can imagine a rebase option that says that it should honor the environment variable). We could also change the behavior of libgit2 to start paying attention to environment variables by default in a future release, with an explicit opt-out. But for now, I think that it needs to remain opt-in.

I'm going to push up a commit to this based on this to adjust...

u-quark:
Thanks for looking at this @ethomson, I've also been busy and didn't chase this.

I understand your points and they make sense! Both changes LGTM. I will have a closer look later this week and we should be able to merge this.

ethomson:
Thanks - sorry to bring the chaotic energy and just push a commit on top of your branch. I appreciate you opening this PR and apologize again for the delay.

ethomson:
Thanks again for the PR!

u-quark:
I just had some time to look into this. This is great! Works for me like a charm!

Thanks for merging this and sorry again for the late response.

PR #6712, opened 2024-01-10
Add git_apply_to_buf method

georgthegreat:


src/libgit2/apply.c
```diff
@@ -385,6 +385,27 @@ static int apply_binary(
 	return error;
 }
 
+int git_apply_to_buf(
+	git_buf *contents_out,
+	char **filename_out,
+	unsigned int *mode_out,
+	const char *source,
+	size_t source_len,
+	git_diff *diff,
```
ethomson, line 394: You can't apply a diff and get a single file out. A diff contains multiple files. Assuming the `0` index doesn't make sense. Maybe this makes sense if the `idx` is a parameter to this function?

ethomson:
Closing this - let me know if you want to discuss a different api

Issue #6074, opened 2021-09-25
http-parser library is unmaintained


dschepler:
The GitHub page for the http-parser library indicates that it is now unmaintained, and suggests llhttp as a replacement to migrate to.  I don't know how much of an impact that would have on libgit2 moving forward; but I thought it might be a good idea for there to be an issue to mark the fact and any thoughts of libgit2 maintainers on implications.

ethomson:
Hi @dschepler - there was some discussion of this over in https://github.com/libgit2/libgit2/discussions/5828.  I'm happy to move to llhttp, but we'd need to port that change in https://github.com/nodejs/http-parser/pull/483 over if it's not there.  This hasn't been a high priority for me, but I'd welcome a PR that moves us over to llhttp if it was something that you were interested in.

sgallagher:
I'm not a subject matter expert for libgit2 OR llhttp, but I've got the beginnings of a patch for this in #6713 
I'd appreciate some help driving it to completion.

glaubitz:
llhttp is part of NodeJS which means that any distribution that packages libgit2 and wants to use the system http parser will no longer be able to build libgit2 on platforms that don't have support for NodeJS.

I don't think this change was wise.

glaubitz:
> llhttp is run by the Node.js project, but it has its own separate releases. Fedora for example ships it as an independent system library. It's not built as part of Node.

Is there a separate upstream repository that does not contain any Javascript code?
 
> Note also that http-parser was also run by the Node.js project and delivered as a separate release in the same way.

Yes, but it was a standalone C package without any additional JavaScript code.

Debian decided to make llhttp part of the [node-undici](https://tracker.debian.org/pkg/node-undici) package which in turn has a build dependency on NodeJS which is not available on all architectures.

So, Debian will have to split llhttp from the node-undici package, although it's unfortunate that upstream didn't split the C from the JavaScript code.

ethomson:
> llhttp is part of NodeJS which means that any distribution that packages libgit2 and wants to use the system http parser will no longer be able to build libgit2 on platforms that don't have support for NodeJS.

I don't understand. Why can't those distributions that don't want to use llhttp just keep using the http-parser that they've been using?

sgallagher:
@glaubitz Just so we're clear, libgit2 didn't *drop* support for http-parser, it just replaced the default with llhttp.

[...]

PR #6713, opened 2024-01-10
Enable llhttp for HTTP parsing

sgallagher:
Fixes: https://github.com/libgit2/libgit2/issues/6074


src/libgit2/transports/httpclient.c
```diff
@@ -7,7 +7,36 @@
 
 #include "common.h"
 #include "git2.h"
-#include "http_parser.h"
+
+#define USE_LLHTTP 1
```
yselkowitz, line 11: Leftover from devel testing?  This is also probably the cause of the CI failures, although CI seems to use the bundled dependencies so it may not really test how this builds with llhttp.

yselkowitz (changes requested):

sergio-correia:
`src/util/net.c` also has an `http_parser.h` include that should likely be removed.

ci/docker/fedora
```diff
@@ -0,0 +1,61 @@
...
+	openssl-devel \
+	openssh-server \
+	git-daemon \
+	java-1.8.0-openjdk-headless \
+	sudo \
+	python3 \
+	krb5-workstation \
+	krb5-libs \
+    krb5-devel \
+    pcre2-devl \
```
sergio-correia, line 24: typo: `pcre2-devel`

sergio-correia:
@ethomson: Hi, would you mind running the CI for this PR, please, so we see what is failing? Thanks in avance.

src/libgit2/transports/httpclient.c
sgallagher: Yeah, that was a glaring mistake.

sgallagher:
Thanks to a lot of help from @sergio-correia, I think this is in proper shape for review. @ethomson if you could kick off the pipeline check, I'd appreciate it.

sgallagher:
I can't reproduce that valgrind error locally. Given where it's located in the code, I suspect it may have been a bug in glibc getaddrinf() that has been fixed in the meantime. Could you retry the check?

ethomson:
Hiya - many apologies for the delay. Thank you so much for all this — one question I have is whether we really need to support _both_ llhttp and the old Node http_parser? It would simplify the code quite a bit to only support the newfangled llhttp, which we could bundle as a dependency (removing the old Node http_parser).

If there's a reason to keep the old support, I'm willing. Just curious.

sgallagher:
> Hiya - many apologies for the delay. Thank you so much for all this — one question I have is whether we really need to support _both_ llhttp and the old Node http_parser? It would simplify the code quite a bit to only support the newfangled llhttp, which we could bundle as a dependency (removing the old Node http_parser).
> 
> If there's a reason to keep the old support, I'm willing. Just curious.

I mostly left it alone because I wasn't prepared to try to entirely extract http-parser since you were carrying a bundled copy. If you're willing to make the transition directly, I'm fully on-board. Upstream http-parser has been dead for four years and I have a sneaking suspicion that there are probably a number of unaddressed security vulnerabilities there that have gone unreported.

ethomson:
Thanks for tackling this PR - it's been on my wish-list for a while to move to llhttp for a plurality of reasons.

After thinking about it a bit more, I realized that other distros should be able to continue to use their existing http-parsers. So I kept the option to use http-parser _or_ llhttp. But I added llhttp as the bundled option, and removed the http-parser that we were bundling. I also brought some of our abstractions a bit more in line with libgit2's typical code style. This lets the http client talk to a generic http parser without having to know what kind of http parser it is, and encapsulates the http parser switching logic in http parser code.

Thanks again! Excited to see this.

sgallagher:
> Thanks for tackling this PR - it's been on my wish-list for a while to move to llhttp for a plurality of reasons.

Happy to help :)

> After thinking about it a bit more, I realized that other distros should be able to continue to use their existing http-parsers. So I kept the option to use http-parser _or_ llhttp. But I added llhttp as the bundled option, and removed the http-parser that we were bundling. I also brought some of our abstractions a bit more in line with libgit2's typical code style. This lets the http client talk to a generic http parser without having to know what kind of http parser it is, and encapsulates the http parser switching logic in http parser code.

Thanks for get it over the finish line. Obviously it's my first contribution to libgit2, so I appreciate the help.

> Thanks again! Excited to see this.

Us too! This is (I think) the last major package in Fedora that's pulling in http-parser, so getting rid of that dependency here is a big deal for us.

ethomson:
> Thanks for get it over the finish line. Obviously it's my first contribution to libgit2, so I appreciate the help.

Appreciate your kind words - I tend to just jump in and meddle in PRs which is not necessarily the most constructive (or polite) way to work.

Can you triple check me and make sure that I didn't break anything for your use case? I'll hit the shiny green button if it's all good.

sgallagher:
Everything looks good to me, though I am curious why you moved the CI run for Fedora to Nightly rather than as part of the merge request workflow. Is it just too slow?

ethomson:
Ah right. Yeah, I want to keep the CI runs minimal enough to just cover the various things that we support, but still be fast. And then nightlies can be more exhaustive.

There's already more than I would like in the CIs, they're awfully slow. I'm happy to rethink what we have where.

PR #6717, opened 2024-01-15
merge: fix incorrect rename detection for empty files.

herrerog:
When merging trees containing multiple empty files, make sure a rename
is not detected between each empty files.

For example
Ancestor has files:
 * foo.c with content
 * bar.c empty

Ours has:
 * foo.c with content
 * boo.c empty

Theirs has:
 * foo.c with other content
 * bar.c with content

`merge_trees()` will incorrectly apply the content of theirs's `bar.c` in
ours's  `boo.c` thinking `bar.c` has been renamed to `boo.c`.
This happens because both are empty and their sha1 are the same. Thus
`merge_diff_mark_similarity_exact()` incorrectly mark it as renamed.


src/libgit2/merge.c
```diff
@@ -1220,11 +1220,18 @@ static int merge_diff_mark_similarity_exact(
 
 	/* Build a map of object ids to conflicts */
 	git_vector_foreach(&diff_list->conflicts, i, conflict_src) {
-		/* Items can be the source of a rename iff they have an item in the
+		/* Items can be the source of a rename if they have an item in the
```
ethomson, line 1223: This is not a typo. Are we changing the logic?

herrerog: No I didn't change the logic, I really thought it was a typo and wanted to fix it while at here.
Should I remove the hunk?

ethomson: Yes please. It's an abbreviation for "if and only if".

herrerog: Ah I didn't know, that's fixed. Thanks.

ethomson:
👋  Thanks for the fix - sorry it's taken so long to come back to it. I did make one change, which was just to move the test _out_ of the `merge::trivial` class and into `merge::renames`. The rationale is that this covers rename functionality and is _not_ related to the trivial resolution steps that git itself defines.

Issue #6640, opened 2023-09-29
Add troubleshooting steps in docs


naty55:
Hi I installed and build the library on Ubuntu Linux and had some issues, I want to add their solutions to documentation 
```
Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
-- Could NOT find GSSAPI (missing: GSSAPI_LIBRARIES GSSAPI_INCLUDE_DIR) 
-- Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY OPENSSL_INCLUDE_DIR)
```

bondress:
This error is raised because the [pkg-config](https://en.wikipedia.org/wiki/Pkg-config) utility is not available on your system.

Using PkgConfig with CMake is not a truly cross-platform solution, as Windows does not come with the pkg-config utility installed. (The PCL developers should instead use find_package() in their CMake. Perhaps, this is worth opening up a bug report on their Github.) On Linux, this is an easy fix; you can install pkg-config like this:

`sudo apt-get install pkg-config`

**Source:** https://stackoverflow.com/questions/33380020/errorcould-not-find-pkgconfig-missing-pkg-config-executable


Summary, for the impatient
To fix this error "-- Could NOT find GSSAPI (missing: GSSAPI_LIBRARIES GSSAPI_INCLUDE_DIR)", run these commands: 

```
sudo ln -s /usr/bin/krb5-config.mit /usr/bin/krb5-config
sudo ln -s /usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2 /usr/lib/libgssapi_krb5.so
sudo apt-get install python-pip libkrb5-dev
sudo pip install gssapi
```
**Source**: https://stackoverflow.com/questions/30896343/how-to-install-gssapi-python-module

To fix this error "-- Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY OPENSSL_INCLUDE_DIR)", run this command:

`sudo apt install libssl-dev`

After you install it, you can now run your compilation which caused the error.

**Source**: https://www.debugpoint.com/could-not-find-openssl/

Suprithvarma1:
[...]

PR #6719, opened 2024-01-17
Troubleshooting Steps

ChudeObiamaka:
This error is raised because the pkg-config is not available in your system
Steps to resolve the problem
Install PkgConfig using the package manager
sudo apt-get update
sudo apt-get install pkg-config
PkgConfig tells the compiler where to find dependent libraries. Installing it with apt-get usually resolves the issue.

If GSSAPI is missing, Install the GSSAPI development package
sudo apt-get update
sudo apt-get install libkrb5-dev
If the issue persists, ensure that the GSSAPI libraries and include files are in standard locations. You can check their locations using
dpkg -L libgssapi-dev
Make sure that the paths are included in the compiler and linker search paths

If OpenSSL is missing, Install the OpenSSL development package
sudo apt-get update
sudo apt-get install libssl-dev
Installing libssl-dev provides the required OpenSSL libraries and headers.
You may also need to set the OpenSSL root directory path as a CMake variable. For example
cmake -DOPENSSL_ROOT_DIR=/usr/lib/ssl ..
Ensure to include any other details about the error messages in the documentation along with these troubleshooting steps. Let me know if you need any clarification or have additional questions!

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,13 +1,21 @@
-Troubleshooting libgit2 Problems
-================================
+This error is raised because the pkg-config is not available in your system
+Steps to resolve the problem
+Install PkgConfig using the package manager
+sudo apt-get update
+sudo apt-get install pkg-config
+PkgConfig tells the compiler where to find dependent libraries. Installing it with apt-get usually resolves the issue.
 
-CMake Failures
---------------
+If GSSAPI is missing, Install the GSSAPI development package
+sudo apt-get update
+sudo apt-get install libkrb5-dev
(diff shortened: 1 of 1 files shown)


ChudeObiamaka:
Please review my PR @naty55

naty55 (changes requested):
Hi @ChudeObiamaka 
I guess I don't have the permissions to approve and merge PRs, Anyway it looks great to me but I'd add some styling for the `commands/code` blocks (md format lets you do it quite easily)

naty55:
fixes #6640

ethomson:
Hi - if you're interested in pushing this forward, please don't remove things like headers, etc, or formatting.

Issue #6708, opened 2023-12-23
Allow set timeout for SSH connection.


SuperBo:
Following documentation, there is no way to set timeout for ssh connection.

`GIT_OPT_SET_SERVER_CONNECT_TIMEOUT: Sets the timeout (in milliseconds) to attempt connections to > a remote server. This is supported only for HTTP(S) connections > and is not supported by SSH. Set to 0 to use the system default. > Note that this may not be able to be configured longer than the > system default, typically 75 seconds.`

### Expected behavior

This timeout should also apply SSH connection. Libssh2 has an API for setting ssh connection timeout.
https://libssh2.org/libssh2_session_set_timeout.html

### Actual behavior

git_remote_fetch hang forever if can't connect to the repository.

### Version of libgit2 (release number or SHA1)
v1.7.1
### Operating system(s) tested
Mac OS

PR #6721, opened 2024-01-19
set SSH timeout

vafada:
fixes https://github.com/libgit2/libgit2/issues/6708 ?

### DIFF
--- a/src/libgit2/transports/ssh_libssh2.c
+++ b/src/libgit2/transports/ssh_libssh2.c
@@ -23,6 +23,8 @@
 
 #define OWNING_SUBTRANSPORT(s) ((ssh_subtransport *)(s)->parent.subtransport)
 
+extern int git_socket_stream__connect_timeout;
+
 static const char cmd_uploadpack[] = "git-upload-pack";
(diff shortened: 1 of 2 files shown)


src/libgit2/transports/ssh_libssh2.c
```diff
@@ -539,6 +541,10 @@ static int _git_ssh_session_create(
 		return -1;
 	}
 
+	if (git_socket_stream__connect_timeout > 0) {
+		libssh2_session_set_timeout(s, git_socket_stream__connect_timeout);
```
ethomson, line 545: We should inherit the `git_socket_stream__connect_timeout` from the socket stream. So we already have connection timeouts.

I think that for this, we want to use `GIT_OPT_SET_SERVER_TIMEOUT`. So I think that this should be `git_socket_stream__timeout` instead...

vafada: updated PR

ethomson:
Looks great — I added one change to the docs that is related but independent. (They stated that the connect timeout was not supported for SSH, in fact, it is.)

Thanks for the PR! 🎉

ethomson (approved):

PR #6723, opened 2024-01-22
config: properly delete or rename section containing multivars

### LINKED ISSUE #6722: Cannot delete or rename a branch whose config section contains multivar entries
If a branch configuration contains duplicated variables (multivar), the branch can neither be removed nor renamed using `libgit2`. This has been noticed using `git branchless` which uses `git2-rs` which uses `libgit2`, in combination with `git publish`. `git publish` stores several `gitpublishto` and `gitpublishcc` entries when a patch series is sent via email, with the various recipients. When a branch is merged upstream, `git branchless sync --pull` will attempt to remove it and fail because of multiple entries.

### Reproduction steps

Create a repository and set multivar config entries for this branch, for example:

```ini
[branch.br2]
gitpublishto = example1@example.com
gitpublishto = example2@example.com
```

Then delete this branch using `libgit2`.

All this can be done with a test in `tests/libgit2/branches/delete.c`:

```c
void test_refs_branches_delete__can_delete_a_local_branch_with_multivar(void)
{
	git_reference *branch;
	git_config *cfg;

	cl_git_pass(git_repository_config(&cfg, repo));
	cl_git_pass(git_config_set_multivar(
	        cfg, "branch.br2.gitpublishto", "^$", "example1@example.com"));
	cl_git_pass(git_config_set_multivar(
	        cfg, "branch.br2.gitpublishto", "^$", "example2@example.com"));
	cl_git_pass(git_branch_lookup(&branch, repo, "br2", GIT_BRANCH_LOCAL));
	cl_git_pass(git_branch_delete(branch));
	git_reference_free(branch);
}
```

### Expected behavior

The branch can be deleted.

### Actual behavior

The branch cannot be deleted because of an error: "entry is not unique due to being a multivar". This is due to the config section deleting (or renaming).

25e2b9d8c217e92eb70cde85bb4007a3dec0428d

NixOS (Linux)

### Resolution

A PR follows.

samueltardieu:
Renaming a config section or deleting its content removes each entry after copying it in its new place if needed. However, since each entry may be part of a multivar, deletion must only remove the exact entry that has just been copied.

Fix #6722

### DIFF
--- a/src/libgit2/config.c
+++ b/src/libgit2/config.c
@@ -1509,35 +1509,31 @@ static int rename_config_entries_cb(
 	int error = 0;
 	struct rename_data *data = (struct rename_data *)payload;
 	size_t base_len = git_str_len(data->name);
-	git_str raw_value = GIT_STR_INIT, value = GIT_STR_INIT;
+	git_str value = GIT_STR_INIT;
 
-	if (base_len > 0 &&
-		!(error = git_str_puts(data->name, entry->name + data->old_len)))
-	{
-		error = git_config_set_multivar(
-		        data->config, git_str_cstr(data->name), "^$",
-		        entry->value);
+	if (base_len > 0) {
+		if ((error = git_str_puts(data->name,
+			entry->name + data->old_len)) < 0 ||
+		    (error = git_config_set_multivar(
+			data->config, git_str_cstr(data->name), "^$",
+			entry->value)) < 0)
+			goto cleanup;
 	}
 
-	if (!error) {
-		if ((error = git_str_puts_escape_regex(
-		             &raw_value, entry->value)))
-			goto cleanup;
-		git_str_grow(&value, git_str_len(&raw_value) + 2);
-		git_str_putc(&value, '^');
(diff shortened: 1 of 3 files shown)


ethomson:
Thanks @samueltardieu - this looks great on first glance, but I'd like to spend a few more minutes to give this a deeper review. Great first PR, I appreciate this.

samueltardieu:
Hi @ethomson. Anything I need to do on my side?

ethomson:
Hi @samueltardieu - sorry about the delay. This is a great fix. I took one more look and realized that we should be able to avoid the copy when constructing the escaped value. I think that we don't need to create the intermediate `value` str with the escaped regex — the `git_str_escape_regex` function should _append_ to an existing buffer. So it should be safe to just putc the `^`, then escape the regex into the value string.

I pushed up a commit to your branch to do this - would you mind 👀 just to make sure that I didn't make any logical errors?

Thanks again! 🙏

samueltardieu:
It is much better indeed.

ethomson:
Thanks for the fix!

PR #6729, opened 2024-02-06
Correct index add directory/file conflict detection

ethomson:
`has_dir_name` is used to check for directory/file collisions, and attempts to determine whether the index contains a file with a directory name that is a proper subset of the new index entry that we're trying to add.

To determine directory name, the function would walk the path string backwards to identify a `/`, stopping at the end of the string. However, the function assumed that the strings did not start with a `/`. If the paths contain only a single `/` at the beginning of the string, then the function would continue the loop, erroneously, when they should have stopped at the first character.

Correct the order of the tests to terminate properly.

Thanks to the Amazon AWS Security team for finding and responsibly disclosing this.

### DIFF
--- a/src/libgit2/index.c
+++ b/src/libgit2/index.c
@@ -1185,10 +1185,13 @@ static int has_dir_name(git_index *index,
 		size_t len, pos;
 
 		for (;;) {
-			if (*--slash == '/')
-				break;
+			slash--;
+
 			if (slash <= entry->path)
 				return 0;
+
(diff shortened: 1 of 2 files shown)


ethomson:
Note: this was included in [v1.6.5](https://github.com/libgit2/libgit2/releases/v1.6.5) and [v1.7.2](https://github.com/libgit2/libgit2/releases/v1.7.2).

PR #6730, opened 2024-02-06
Correct `git_revparse_single` and add revparse fuzzing

ethomson:
1. Introduce address sanitization during builds. Surprisingly, although we enabled various sanitizers (thread, memory, undefinedbehavior) we did _not_ have address sanitizer running. Add that.

2. Address sanitizer correctly detected a failure to correctly parse capabilities in a first pkt. In our test corpus we have a message with no capabilities (the length of the message is the ref advertisement plus a trailing NUL). We walk off the end of that string trying to read capabilities. Now we honor the `len` before checking for caps.

3. Update revparse to correctly stop parsing `spec` when we find a trailing `@`. Thanks to the Amazon AWS Security team for finding and responsibly disclosing this.


ethomson:
Note: this was included in [v1.6.5](https://github.com/libgit2/libgit2/releases/v1.6.5) and [v1.7.2](https://github.com/libgit2/libgit2/releases/v1.7.2).

PR #6736, opened 2024-02-12
GIT_RAND_GETENTROPY: do not include sys/random.h

semarie:
CMakeLists.txt is looking for `unistd.h` (where the `getentropy` function is present for at least Linux and OpenBSD).

It isn't necessary to include `sys/random.h` (which is Linux only).

It unbreak the build on OpenBSD.

### DIFF
--- a/src/util/rand.c
+++ b/src/util/rand.c
@@ -10,10 +10,6 @@ See <http://creativecommons.org/publicdomain/zero/1.0/>. */
 #include "rand.h"
 #include "runtime.h"
 
-#if defined(GIT_RAND_GETENTROPY)
-# include <sys/random.h>
-#endif
-
 #if defined(GIT_WIN32)
 # include <wincrypt.h>
 #endif


semarie:
For reference, the [CMakeLists.txt#L87](https://github.com/libgit2/libgit2/blob/8535fdb9cbad8fcd15ee4022ed29c4138547e22d/src/CMakeLists.txt#L87) detection.

From [getentropy(3)](https://www.man7.org/linux/man-pages/man3/getentropy.3.html)  man page, it should be at least glibc 2.25.

semarie:
The test failing seems unrelated at first look: the job compiled fine (so the include removal was accepted by the compiler), and it failed during `online` test (it failed several times in row).

the test seems to failed due to `Syscall param socketcall.sendto(msg) points to uninitialised byte(s)`

ethomson:
> The test failing seems unrelated at first look: the job compiled fine (so the include removal was accepted by the compiler), and it failed during online test (it failed several times in row).

Indeed - it looks like we need to update our valgrind suppression settings — OpenSSL has some cases where it (intentionally) doesn't initialize memory and valgrind treats that as tainted, and thus our data is treated as tainted as a result as well.

ethomson (approved):
Thanks for the fix!

PR #6738, opened 2024-02-16
Support index.skipHash true config

### LINKED ISSUE #6531: git 2.40.0 index.skipHash incompatible with libgit2
The recently introduced `index.skipHash` config setting seems to create index files that are incompatible with libgit2. This option is enabled with `feature.manyFiles` which some users have enabled to improve performance.

See: https://github.com/git/git/commit/ffd923868574a004ec20a80dbc3b813da9a93b08

### Reproduction steps
1. With git 2.40.0, run `git clone -c index.skipHash=true https://github.com/libgit2/libgit2.git`
2. Run `libgit2/build/examples/lg2 status`

### Expected behavior
Successfully opens the repo.

### Actual behavior
Fails with the error:

```
Could not get status [-1] - invalid data in index - calculated checksum does not match expected
```

8f8e805e3f37e49563c33a0c67fcf7f0d2e9101c or 1.6.2

Any

parnic:
When this is set, the checksum at the end of the index is zero.

Fixes #6531


src/libgit2/index.c
```diff
@@ -2849,8 +2850,9 @@ static int parse_index(git_index *index, const char *buffer, size_t buffer_size)
 	/*
 	 * SHA-1 or SHA-256 (depending on the repository's object format)
 	 * over the content of the index file before this checksum.
+	 * Note: checksum may be 0 if index.skipHash is set to true.
```
ethomson, line 2853: My reading of this suggested that `index.skipHash` was supported during read; I'd suggest we change the messaging a bit to when we _encounter_ a null hash.

```suggestion
	 * Note: checksum may be 0 if the index was written by a client
	 * where index.skipHash was set to true.
```

src/libgit2/index.c
```diff
@@ -2849,8 +2850,9 @@ static int parse_index(git_index *index, const char *buffer, size_t buffer_size)
...
 	 */
-	if (memcmp(checksum, buffer, checksum_size) != 0) {
+	if (memcmp(zero_checksum, buffer, checksum_size) != 0 && memcmp(checksum, buffer, checksum_size) != 0) {
```
ethomson, line 2855:
```suggestion
	if (memcmp(zero_checksum, buffer, checksum_size) != 0 &&
	    memcmp(checksum, buffer, checksum_size) != 0) {
```

tests/libgit2/status/worktree.c
```diff
@@ -1360,3 +1360,12 @@ void test_status_worktree__at_head_parent(void)
 	git_tree_free(parent_tree);
 	git_status_list_free(statuslist);
 }
+
+void test_status_worktree__skip_hash(void)
+{
+	git_repository *repo = cl_git_sandbox_init("status_skipHash");
```
ethomson, line 1366: I'd like this to not be bicapitalized; it doesn't fit with our other unit tests.

```suggestion
	git_repository *repo = cl_git_sandbox_init("status_skiphash");
```

ethomson (commented):
Thanks for the fix - I would request a few minor formatting suggestions. 🙏

parnic:
> Thanks for the fix - I would request a few minor formatting suggestions. 🙏

Sure thing - all done.

ethomson:
I think that we need to `git_index_free` in the test - I pushed up a quick change to your branch to make sure that's the problem.

Issue #6737, opened 2024-02-14
Repository ownership validation does not skip `%(prefix)/` on Windows


csware:
### Reproduction steps

Use Windows, open a repository on a network share.
Whitelist that directory with a prefix of `%(prefix)/`

PR #6668 contained a full fix, but the changes applied to main are incomplete.

### Expected behavior

The safe.directory exception is accepted

### Actual behavior

The safe.directory exception doesn't match

### Version of libgit2 (release number or SHA1)

8535fdb9cbad8fcd15ee4022ed29c4138547e22d

### Operating system(s) tested

Windows 10

csware:
Furthermore, I'm unsue whether the prefix-stuff is still needed for network shares, as vanilla Git now (Git 2.43) also supports `//192.168.0.1/share/repo` as safe.directory.

csware:
Also, as an improvement, the error message (https://github.com/libgit2/libgit2/blob/8535fdb9cbad8fcd15ee4022ed29c4138547e22d/src/libgit2/repository.c#L708-L710) should not contain a trailing slash to make it easier for users to know what to whitelist.

csware:
PS: A fix could also be released for 1.7.3

ethomson:
Thanks for the report — I think that I wrapped these issues up in #6739.

PR #6739, opened 2024-02-18
More `safe.directory` improvements

ethomson:
The gift that keeps on giving — it will rival `core.autocrlf` for user disappointment.

A number of changes here:

1. We ensure that we cope correctly with `%(prefix)` in front of paths.
2. We stop requiring that UNC paths start with `%(prefix)`
3. Report the directory path without a trailing slash
4. Ensure that root paths are correctly handled
5. Ensure that the repository path is canonicalized on Windows — we were already doing this on POSIX, and this aligns with Git itself
6. To cope with this, we need to start canonicalizing the clar sandbox paths so that the repository path is deterministic and we can continue joining paths simply


ethomson:
The gift that keeps on giving — it will rival `core.autocrlf` for user disappointment.

A number of changes here:

1. We ensure that we cope correctly with `%(prefix)` in front of paths.
2. We stop requiring that UNC paths start with `%(prefix)`
3. Report the directory path without a trailing slash
4. Ensure that root paths are correctly handled
5. Ensure that the repository path is canonicalized on Windows — we were already doing this on POSIX, and this aligns with Git itself
6. To cope with this, we need to start canonicalizing the clar sandbox paths so that the repository path is deterministic and we can continue joining paths simply

ethomson:
Fixes #6737

csware:
As another improvement, the error message should provide the `%(prefix)/` as vanilla Git does.

A fix would also be nice to have for 1.7.3

ethomson:
> As another improvement, the error message should provide the %(prefix)/ as vanilla Git does.

Right - sorry - I went back to your PR and realized that I never actually finished typing up the comments. We need to avoid multi-line error messages since some people use them in odd places. I don't have a great idea for how to plumb this through (the directory that's problematic vs the configuration command that you would run). This _feels_ like a good use for the "warnings API" that we keep talking about and never shipping, so let me give it some thought.

ethomson:
I'm willing to back port to v1.7, but I'm also preparing to cut v1.8 if that makes more sense for you.

csware:
I'm testing the main branch right now...

csware:
> > As another improvement, the error message should provide the %(prefix)/ as vanilla Git does.
> 
> Right - sorry - I went back to your PR and realized that I never actually finished typing up the comments. We need to avoid multi-line error messages since some people use them in odd places. I don't have a great idea for how to plumb this through (the directory that's problematic vs the configuration command that you would run). This _feels_ like a good use for the "warnings API" that we keep talking about and never shipping, so let me give it some thought.

I'm not talking about presenting the git command to use (which may not make any sense if Git is not installed). Just printing: this is the path you need to whitelist.

PR #6747, opened 2024-02-24
Fix broken links

### LINKED ISSUE #6570: URLs in example code are broken
The [`libgit2` example code](/libgit2/libgit2/tree/main/examples) has many URLs that do not point to anything anymore. 

For example, `README.md` contains this URL: http://libgit2.github.com/libgit2/ex/HEAD/general.html
The URL redirects to https://libgit2.org/

Here is another example, taken from [`general.c`](/libgit2/libgit2/blob/main/examples/general.c#L737-L744):

```c
/**
 * ### Config Files
 *
 * The [config API][config] allows you to list and update config values
 * in any of the accessible config file locations (system, global, local).
 *
 * [config]: http://libgit2.github.com/libgit2/#HEAD/group/config
 */
```

The URL (http://libgit2.github.com/libgit2/#HEAD/group/config) redirects to https://libgit2.org/#HEAD/group/config, which just shows the `libgit2.org` front page. 

1) Where might the information that the author intended be found?
2) The URLs should be corrected

csware:
fixes  #6570

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ Getting Help
 **Getting Help**
 
 If you have questions about the library, please be sure to check out the
-[API documentation](http://libgit2.github.io/libgit2/).  If you still have
+[API documentation](https://libgit2.org/libgit2/).  If you still have
 questions, reach out to us on Slack or post a question on 
 [StackOverflow](http://stackoverflow.com/questions/tagged/libgit2) (with the `libgit2` tag).
 
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ Getting Help
 **Getting Help**
 
 If you have questions about the library, please be sure to check out the
-[API documentation](http://libgit2.github.com/libgit2/).  If you still have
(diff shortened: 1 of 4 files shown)


csware:
The tests are flaky.

ethomson:
Why did libgit2.github.com break without any warning? I know that we were special cased or something to keep that domain but that's very annoying. @carlosmn any way you can find out what happened?

ethomson:
It looks like all these links work again?

csware:
No.

ethomson:
Oh, I see, it's not preserving the path in the redirect. That's probably my fault from when I moved off Google Domains.

ethomson:
Nope, it's GitHub, not including the path in our redirects. 🙃 

Can we update the links to `libgit2.org` instead of `github.io`?

csware:
Done.

ethomson:
Thanks!

PR #6749, opened 2024-02-24
Fix broken regexp that matches submodule names containing ".path"

### LINKED ISSUE #6580: git_submodule_foreach() fails for certain submodules names
For certain repos `git_submodule_foreach` seems to not work without any reason or error.

I can't narrow down what causes this; it happens for `.gitmodules`, for example certain names with a dot icw/ `shallow = true`. Below an example.

### Reproduction steps

 - git clone https://github.com/jeroen/libgit2bug
 - call any program with `git_submodule_foreach` to list the submodules. 

For example this program:

```c
#include <string.h>
#include <stdio.h>
#include <git2.h>

static int submodule_count(git_submodule *submod, const char *name, void *payload){
  int *count = payload;
  printf("Found submodule %s\n", git_submodule_name(submod));
  *count = (*count) + 1;
  return 0;
}

int main(int argc, char *argv[]) {
  git_repository *repo = NULL;
  git_libgit2_init();
  int err = git_repository_open(&repo, ".");
  if(err){
    const git_error *info = giterr_last();
    printf("Failure in git_repository_open: %s\n", info->message);
    return 1;
  }
  int n = 0;
  git_submodule_foreach(repo, submodule_count, &n);
  printf("This function has %d submodules\n", n);
  return 0;
}

```

Alternatively you can use the libgit2 example app: `libgit2/examples/a.out status --list-submodules` on that same repo.

### Expected behavior

We iterate over the submodules.

### Actual behavior

No submodules are listed.

1.6.4

MacOS, Linux.

csware:
fixes #6580

### DIFF
--- a/src/libgit2/submodule.c
+++ b/src/libgit2/submodule.c
@@ -196,7 +196,7 @@ static void free_submodule_names(git_strmap *names)
  */
 static int load_submodule_names(git_strmap **out, git_repository *repo, git_config *cfg)
 {
-	const char *key = "submodule\\..*\\.path";
+	const char *key = "^submodule\\..*\\.path$";
 	git_config_iterator *iter = NULL;
 	git_config_entry *entry;
 	git_str buf = GIT_STR_INIT;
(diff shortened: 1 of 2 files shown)


ethomson:
Nice catch. Thanks for the fix!

jeroen:
Thanks!

PR #6756, opened 2024-03-10
Separate config reader and writer backend priorities (for worktree configs)

ethomson:
For #6202, we add additional customization on read/write priority, so that we can have independent read and write ordering. This allows us to send writes to the local configuration backend _even if_ there is a higher-priority read backend (like a worktree configuration). Repositories will configure the read/write order specifically, and users can add a `APP` level configuration which takes priority (over both). Users can further tweak priority themselves.


ethomson:
It looks like we support reading the `worktreeconfig` extension — but it looks like we're always supporting the worktree config _even if_ the worktreeconfig extension isn't specified. We should validate that.

ethomson:
cc @vermiculus - curious if you have any feedback here.

ethomson:
Going to merge this, I think it’s a nice little api change. Thanks again @vermiculus

zivarah:
> cc @vermiculus - curious if you have any feedback here.

@ethomson apologies for the lack of response, both myself and @vermiculus got a little busy and this slipped through. Thanks for running with this!

PR #6759, opened 2024-03-13
Allow libgit2 to be compiled with mbedtls3.

adamharrison:
OK; turns out there were relatively minimal changes required to have this happen.

Some background and reason for the changes:

1. `mbedtls/config.h` has been replaced by `mbedtls/build_info.h` mbedtls3. In either event, this file does not need to be included directly, as it is included by default in all lower other mbedtls headers.
2. The internals of the ssl_conf have been flagged as private. In theory, these shouldn't have been used directly in the first place, I think. However, given that we are using them, I decided to not change the system, and simply attach the appropriate prefix.

I think that's all we need. It compiles for me. Any thoughts?


adamharrison:
It looks like this will actually require an update to ntlmclient; as it stands, it depends on `md4`, which is removed from `mbedtls3`. We can probably include that algorithm statically in the worst case, copied from the mbedtls2.

ethomson:
Thanks for opening this! I agree that freeing things in the `private` data structure is not the right thing to do. Looking at it a bit closer, we don't even need to heap allocate that data. I moved it all to static data.

I also updated ntlmclient to cope with the lack of MD4 in mbedTLS.

I pushed up those changes on top of your PR branch. Thanks again!

ethomson:
Thanks @adamharrison!

Issue #6728, opened 2024-02-06
Tests fail when built against zlib-ng-compat


sgallagher:
### Reproduction steps
On Fedora 40 (or any other system with zlib-ng-compat-devel available), build libgit2 with -DUSE_BUNDLED_ZLIB=OFF

### Expected behavior

The test suite should succeed.

### Actual behavior

```
  1) Failure:
pack::packbuilder::create_pack [/home/sgallagh/localworkspace/libgit2/tests/libgit2/pack/packbuilder.c:143]
  String mismatch: hex != "5d410bdf97cf896f9007681b92868471d636954b"
  '5f688cedc3a5132f15fc3352a35bc5dcb6a536a4' != '5d410bdf97cf896f9007681b92868471d636954b' (at byte 1)

  2) Failure:
pack::packbuilder::get_name [/home/sgallagh/localworkspace/libgit2/tests/libgit2/pack/packbuilder.c:151]
  String mismatch: "7f5fa362c664d68ba7221259be1cbd187434b2f0" != git_packbuilder_name(_packbuilder)
  '7f5fa362c664d68ba7221259be1cbd187434b2f0' != 'f120f6bbad709d0e3c021ca8f9d7c26c3882497b' (at byte 0)

  3) Failure:
pack::packbuilder::write_default_path [/home/sgallagh/localworkspace/libgit2/tests/libgit2/pack/packbuilder.c:159]
  Expression is not true: git_fs_path_exists("objects/pack/pack-7f5fa362c664d68ba7221259be1cbd187434b2f0.idx")

  4) Failure:
pack::packbuilder::permissions_standard [/home/sgallagh/localworkspace/libgit2/tests/libgit2/pack/packbuilder.c:184]
  Function call failed: (stat("pack-7f5fa362c664d68ba7221259be1cbd187434b2f0.idx", &statbuf))
  error -1 - no error

  5) Failure:
pack::packbuilder::permissions_readonly [/home/sgallagh/localworkspace/libgit2/tests/libgit2/pack/packbuilder.c:184]
  Function call failed: (stat("pack-7f5fa362c664d68ba7221259be1cbd187434b2f0.idx", &statbuf))
  error -1 - no error

  6) Failure:
pack::packbuilder::permissions_readwrite [/home/sgallagh/localworkspace/libgit2/tests/libgit2/pack/packbuilder.c:184]
[...]

PR #6762, opened 2024-03-14
packbuilder: adjust nondeterministic tests

ethomson:
The packbuilder tests should be deterministic. This comment suggests that they are:

```
/*
 * By default, packfiles are created with only one thread.
 * Therefore we can predict the object ordering and make sure
 * we create exactly the same pack as git.git does when *not*
 * reusing existing deltas (as libgit2).
 *
 * $ cd tests/resources/testrepo.git
 * $ git rev-list --objects HEAD | \
 *  git pack-objects -q --no-reuse-delta --threads=1 pack
 * $ sha1sum pack-7f5fa362c664d68ba7221259be1cbd187434b2f0.pack
 * 5d410bdf97cf896f9007681b92868471d636954b
 *
 */
```

but it is disappointingly incorrect. As is evidenced by the fact that -- at least on _my_ machine -- that command does not actually produce that output any longer.

Variations in things like the which compression library is actually used, and its defaults, mean that we cannot accurately predict or enforce the bytes in the packfile from one system to another.

Adjust the tests so that they do not believe that they should enforce the bytes in the packfile. This allows broader compatibility with zlib-compatible compression libraries, or other compression levels.

### DIFF
--- a/tests/libgit2/pack/packbuilder.c
+++ b/tests/libgit2/pack/packbuilder.c
@@ -98,9 +98,6 @@ void test_pack_packbuilder__create_pack(void)
 {
 	git_indexer_progress stats;
 	git_str buf = GIT_STR_INIT, path = GIT_STR_INIT;
-	git_hash_ctx ctx;
-	unsigned char hash[GIT_HASH_SHA1_SIZE];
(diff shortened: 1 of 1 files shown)


ethomson:
Fixes #5324 
Fixes #6728

Issue #6777, opened 2024-03-21
1.8.0: Calling git_commit_create_from_stage without author and committer does not use the default committer


florianpircher:
### Reproduction steps

Call `git_commit_create_from_stage` with either

- `given_opts` of `NULL` or
- `given_opts` with both `author` and `committer` of `NULL`.

### Expected behavior

The default author and committer are used, as per

https://github.com/libgit2/libgit2/blob/d74d491481831ddcd23575d376e56d2197e95910/include/git2/commit.h#L409-L413

### Actual behavior

The default committer is not used, leaving the internal `committer` variable of `git_commit_create_from_stage` uninitialized, leading to `EXC_BAD_ACCESS` or similar issues.

### Version of libgit2 (release number or SHA1)

d74d491481831ddcd23575d376e56d2197e95910 (1.8.0)

### Operating system(s) tested

macOS 14.2.1 (23C71)

### Debugging

The variables are set up here:

https://github.com/libgit2/libgit2/blob/d74d491481831ddcd23575d376e56d2197e95910/src/libgit2/commit.c#L1089-L1097

and checked here:

https://github.com/libgit2/libgit2/blob/d74d491481831ddcd23575d376e56d2197e95910/src/libgit2/commit.c#L1110-L1120

However, if `(author = opts.author) == NULL`, the condition short-circuits and `committer = opts.committer` is never run, leaving `committer` uninitialized. The code

https://github.com/libgit2/libgit2/blob/d74d491481831ddcd23575d376e56d2197e95910/src/libgit2/commit.c#L1118-L1119

will thus most likely not be executed, meaning the default committer is never set and instead the uninitialized `committer` is used in the following code.

***

Thanks for the addition of `git_commit_create_from_stage`, this is a handy shortcut!

Uncommon:
I just now ran into this and was about to file an issue. Good thing I checked first :)

ethomson:
Oops - thanks for the report.

PR #6781, opened 2024-03-24
commit: Fix git_commit_create_from_stage without author and committer

florianpircher:
This fixes the issue described in #6777. It’s a minor code change, thus I did not touch any other files.


ethomson:
Thanks!

PR #6798, opened 2024-04-21
add deprecated attribute to avoid warning and makes it easier to upgr…

schriftgestalt:
…ade old code


ethomson:
Probably a good idea but let’s use `GIT_DEPRECATED`?

schriftgestalt:
I was looking at this in Xcode and got a lot warnings that the attributes are missing. But it seems that not all compilers are happy with those.

Issue #6801, opened 2024-04-28
Regression, calls to !git_error_last() not updated after commit 3618a2a


jdavid:
This is a regression in libgit2 v1.8.0
Error introduced in commit 3618a2aa45893a88fbb2d1e0eb97d530c7dc4f4d
It shows app with Windows (winhttp).
Found in pygit2's test suite, see https://ci.appveyor.com/project/jdavid/pygit2/builds/49513776
It currently blocks the release of pygit2, see https://github.com/libgit2/pygit2/issues/1285

Commit 3618a2aa45893a88fbb2d1e0eb97d530c7dc4f4d introduces this change:

```
@@ -184,6 +189,9 @@ const git_error *git_error_last(void)
        if ((threadstate = git_threadstate_get()) == NULL)
                return &tlsdata_error;
 
+       if (!threadstate->last_error)
+               return &no_error;
+
        return threadstate->last_error;
 }
```

But winhttp.c is not updated:

```
static int certificate_check(winhttp_stream *s, int valid)
{
    [...]

    if (error < 0 && !git_error_last())    <----
        git_error_set(GIT_ERROR_HTTP, "user cancelled certificate check");
```

Here since commit 3618a2aa45893a88fbb2d1e0eb97d530c7dc4f4d `!git_error_last()` always evaluates to false.

To find out where the issue is I made my fork of libgit2 and reverted the change introduced in commit 3618a2aa45893a88fbb2d1e0eb97d530c7dc4f4d. The result is that the tests pass, see https://ci.appveyor.com/project/jdavid/pygit2/builds/49705963

Of course I think the proper fix would be to update `winhttp.c`

A `git grep` shows that this same error happens in other places in libgit2:

```
$ git --no-pager grep "\!git_error_last()"
src/libgit2/transports/ssh_libssh2.c:		if (!git_error_last())
src/libgit2/transports/winhttp.c:		if (!git_error_last())
src/libgit2/transports/winhttp.c:	if (error < 0 && !git_error_last())
src/libgit2/transports/winhttp.c:				if (!git_error_last())
src/util/fs_path.c:			if (!git_error_last())
```

### Reproduction steps

Error found in pygit2 CI (AppVeyor), see https://ci.appveyor.com/project/jdavid/pygit2/builds/49513776/job/v4dw1jrduwk69k20

### Expected behavior

The test should pass.

### Actual behavior

The test fails.

### Version of libgit2 (release number or SHA1)

[...]

PR #6802, opened 2024-04-29
transport: provide a useful error message during cancellation

ethomson:
Since 3618a2a, `git_error_last` does not return NULL when there was no error. Adapt to this when testing for a user cancelling a callback but not setting an error message, testing for `klass` of `GIT_ERROR_NONE`.


ethomson:
Fixes #6801

PR #6808, opened 2024-05-08
examples: fix memory leak in for-each-ref.c

qaqland:


qaqland:
this change should has nothing to do with actions' check🤔 why some of them failed

ethomson:
Thanks!

PR #6809, opened 2024-05-08
Revparse: Correctly accept ref with '@' at the end

csware:
libgit2 doesn't accept refs with an '@' at the end in `git_revparse_single`.

cross-ref: https://gitlab.com/tortoisegit/tortoisegit/-/issues/4133


ethomson:
Thanks!

ethomson (approved):

PR #6817, opened 2024-05-15
libgit2 v1.8.1

ethomson:
Update changelog and version numbers to release v1.8.1


Oc199 (commented):
I can't belive why you can do that thing

PR #6819, opened 2024-05-20
Fix constness issue introduced in #6716

georgthegreat:
These functions should accept const pointer to const objects.


ethomson:
Fixed via #6829

PR #6828, opened 2024-06-11
Add assert

Murmele:
Description: In an older version of libgit2 in git_object_lookup_prefix was a check that repo is valid, but now there is repo->oid_type in the git_object_lookup and must be checked as well


ethomson:
Thanks!

PR #6833, opened 2024-06-15
add possibility to use local build libssh2

Murmele:
For building windows applications it is quite comfortable to build libssh2 directly with cmake and including into the package.


ethomson:
I would like some documentation about how you would make this work. If I wanted to build an app with both libgit2 and libssh2, I don't know how I should set up my CMake configuration to achieve that.

Can you describe how someone else would succeed with this option? I regret that I lock the necessary background to understand it yet.

Murmele:
The usecase for it is building windows is much easier when the project can be fetched directly from cmake.

In Gittyup we currently use it like this:

```cmake
if(USE_SYSTEM_LIBSSH2)
  if(PKG_CONFIG_FOUND)
	pkg_check_modules(LIBSSH2 REQUIRED libssh2)
  endif()

else()
  set(BUILD_EXAMPLES
      OFF
      CACHE BOOL "" FORCE)
  set(BUILD_TESTING
      OFF
      CACHE BOOL "" FORCE)

  add_subdirectory(libssh2)

  set(LIBSSH2_FOUND
      TRUE
      PARENT_SCOPE)
  set(LIBSSH2_LIBRARIES
      libssh2
      PARENT_SCOPE)
  set(LIBSSH2_INCLUDE_DIRS
      ${CMAKE_CURRENT_SOURCE_DIR}/libssh2/include
      PARENT_SCOPE)
endif()
```

Now all LIBSSH2 variables are set and libgit2 can use it directly.

PR #6834, opened 2024-06-18
Limit .gitattributes and .gitignore files to 100 MiB

csware:
Git introduced this 100 MiB limit in commits https://github.com/gitster/git/commit/3c50032ff528 (attr: ignore overly large gitattributes files, 2022-12-01) and https://github.com/gitster/git/commit/e7c3d1ddba0b (dir.c: reduce max pattern file size to 100MB, 2024-06-05).


ethomson:
Makes sense, thanks!

PR #6836, opened 2024-06-22
odb: conditional git_hash_ctx_cleanup in git_odb_stream

gensmusic:
As discussed in https://github.com/libgit2/libgit2/discussions/6835

When `git_odb_stream` is a read stream, `hash_ctx` is not used. Therefore, check if `hash_ctx` can be freed during the release. This allows implementers of custom ODB backends to not worry about the creation of `hash_ctx` for now.


ethomson:
Thanks!

PR #6837, opened 2024-06-22
Fix docs for git_odb_stream_read return value.

ehuss:
I believe the original docs for the return value of `git_odb_stream_read` might be incorrect. It currently says that it returns 0 on success. However, that means there would be no way to know how many bytes were actually saved to the buffer. `loose_backend__readstream_read` appears to return the total number of bytes read. I don't know if there are any other backends that implement readstream (I don't see any).


ethomson:
Good catch, thanks!

PR #6840, opened 2024-07-03
Add cmake package targets

bmarques1995:
These changes will let the user call the command `find_package(libgit2)` when the library is installed, this can be use as a 
guarantee that the package is properly installed.


bmarques1995:
> Any progress on this PR? This's really helpful to someone who wants to get libgit2 easily used in a CMake based project.

The task is ready, but need to be accepted

src/cli/CMakeLists.txt
```diff
@@ -53,4 +53,4 @@ if(MSVC_IDE)
 	set_source_files_properties(win32/precompiled.c COMPILE_FLAGS "/Ycprecompiled.h")
 endif()
 
-install(TARGETS git2_cli RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+install(TARGETS git2_cli EXPORT ${LIBGIT2_TARGETS_EXPORT_NAME} RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```
ethomson, line 56: Curious why we'd export the CLI targets? How would somebody make use of this and why would they want it? 🤔

ethomson:
Thanks for your patience. Every time I look at cmake, my eyes glaze over, and I need to go lie down. Just one question to start with about the CLI.

src/cli/CMakeLists.txt
bmarques1995: You are correct, for the git cli, the export is unnecessary.

ethomson:
Thanks for the fix! This will land in v1.9.

PR #6842, opened 2024-07-03
Headers cleanup

anatol:
A small cleanup for libgit2 headers

### DIFF
--- a/include/git2/email.h
+++ b/include/git2/email.h
@@ -86,30 +86,6 @@ typedef struct {
 
 /**
  * Create a diff for a commit in mbox format for sending via email.
- *
- * @param out buffer to store the e-mail patch in
- * @param diff the changes to include in the email
- * @param patch_idx the patch index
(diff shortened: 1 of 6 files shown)


ethomson:
Oops, good catch on `email_from_diff`. It was actually supposed to be in the `sys/email.h` header. Would you mind zapping it from `email.h` instead? 🙏

anatol:
> Oops, good catch on `email_from_diff`. It was actually supposed to be in the `sys/email.h` header. Would you mind zapping it from `email.h` instead? 🙏

Sure, done. It turns there are some users that expect the function to be at `git2/mail.h`, so I fixed them as well.

ethomson:
Awesome, thanks! 🙏

anatol:
@ethomson could you please approve the github workflow to see if my last patch fixes the compilation issue?

ethomson:
Thanks!

PR #6846, opened 2024-07-07
Fix shallow root maintenance during fetch

kcsaul:
Fixes issue where the shallow roots array size wasn't being copied as part of preparing to perform a fetch.

This resulted in the shallow roots file being removed unexpectedly at the end of the fetch, even though there's still one or more shallow root commits as they weren't included/ancestors of the branches/commits fetched, which subsequently led to invalid/missing object errors.

### DIFF
--- a/src/libgit2/transports/smart_protocol.c
+++ b/src/libgit2/transports/smart_protocol.c
@@ -394,6 +394,7 @@ static int setup_shallow_roots(
 
 		memcpy(out->ptr, wants->shallow_roots,
 		       sizeof(git_oid) * wants->shallow_roots_len);
+		out->size = wants->shallow_roots_len;
 	}
 
 	return 0;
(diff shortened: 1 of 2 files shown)


ethomson:
Hmm, sounds right, but it looks like you've uncovered a memory leak. 🤔

kcsaul:
@ethomson - Could you please have another look at this?

I believe the memory leak you'd spotted was a knock-on impact of an existing test failing due to the fix.  I've updated the online::shallow::shorten_four test to reflect what git does when fetching with depth 5, and then 4 afterwards.  It doesn't actually remove the original set of shallow roots, presumably because they're not traversed on the server during the subsequent fetch.  It's not until git gc is run the unreachable shallow roots are removed.

Now that I've realized where to add a new test - I've added one to prove existing shallow roots will be preserved during fetch for an unrelated branch, and that they'll be removed as expected during subsequent fetches.

Outstanding/failing test appears to be one of the flaky network tests.

ethomson:
> I believe the memory leak you'd spotted was a knock-on impact of an existing test failing due to the fix. I've updated the online::shallow::shorten_four test to reflect what git does when fetching with depth 5, and then 4 afterwards. It doesn't actually remove the original set of shallow roots, presumably because they're not traversed on the server during the subsequent fetch. It's not until git gc is run the unreachable shallow roots are removed.

Ah, that makes sense. Perfect, thanks so much. 🙏

PR #6847, opened 2024-07-10
Add unsigned specifier to origin.

apetrai:
Reduces chance of undefined behavior between different machines.

### DIFF
--- a/include/git2/diff.h
+++ b/include/git2/diff.h
@@ -629,7 +629,7 @@ typedef enum {
  * just a span of bytes inside the larger file.
  */
 typedef struct {
-	char   origin;       /**< A git_diff_line_t value */
+	unsigned char   origin;       /**< A git_diff_line_t value */
 	int    old_lineno;   /**< Line number in old file or -1 for added line */
 	int    new_lineno;   /**< Line number in new file or -1 for deleted line */
 	int    num_lines;    /**< Number of newline characters in content */


ethomson:
Can you help me understand the motivation here a bit better? What's the UB?

apetrai:
I apologise for not making myself clear @ethomson 

 - The char type can be either signed or unsigned depending on the compiler and platform, which is implementation-defined.
- Negative Values with Signed char: If char is signed, it can hold negative values (typically ranging from -128 to 127). When dealing with character values outside the ASCII range (0-127), signed char can produce unexpected negative values.
- Library Functions and Undefined Behavior: Standard library functions like isalpha() and isdigit() expect an int argument that is either EOF or representable as an unsigned char. Passing a negative char value to these functions without casting can result in undefined behavior.

- This (little) change ensure consistent and predictable behavior across different compilers and platforms, avoiding potential bugs caused by the ambiguous signedness of char.

apetrai:
Additionally, since as the comment suggests this is a  `git_diff_lint_t` value which I assume is the difference of lines of a given file, the `unsigned` char is guaranteed across all machines to display difference values more than 127. Up to 255.

I am quite new to this project and to the documentation, I apologise if this change won't help, but I believe that just by adding an `unsigned` specifier, it could reduce the chance of UB on different machines.

PR #6848, opened 2024-07-10
Add more robust reporting to SecureTransport errors on macos

vcfxb:
This pull request ensures errors from `SecureTransport` on macos are handled properly when `CFStringGetCStringPtr` returns `null`. This occurs when the string encoding conversion is not cheap/efficient. 

I am making this pull request after encountering a `SecureTransport error: (null); class=Net (12)` while working on https://github.com/mitre/hipcheck. With these changes, I now get an error message from macos instead of `(null)`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -75,4 +75,5 @@ Tim Clem
 Tim Harder
 Torsten Bögershausen
 Trent Mick
+Venus Xeon-Blonde
 Vicent Marti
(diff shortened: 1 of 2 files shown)


ethomson:
Sorry for the delay here - I was looking back through old PRs and realized that I never actually published my comments. 😓  Many apologies.

I went ahead and manually merged this, with a few stylistic updates. Thanks for the contribution!

PR #6859, opened 2024-08-16
Fix contradictory phrase in SECURITY.md

Kyle-Ye:


ethomson:
Thanks! That was definitely misleading.

PR #6862, opened 2024-08-17
Add SecCopyErrorMessageString for iOS and update README for iOS

### LINKED ISSUE #6861: Add SecCopyErrorMessageString for iOS
SecCopyErrorMessageString is introduced to iOS since iOS 11.3 (29 March 2018)

https://developer.apple.com/documentation/security/1394686-seccopyerrormessagestring?language=objc

https://github.com/libgit2/libgit2/blob/4d38f13619430f75a5c6890ccf8ea30c77db4738/src/libgit2/streams/stransport.c#L29-L38

This is a easy to fix issue. I'd like to file a PR to fix it.

Kyle-Ye:
- **Add iOS build instruction in README**
- **Fix iOS build issue**
- **Add SecCopyErrorMessageString API when building for iOS**

Close #6861

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -73,19 +73,6 @@ jobs:
             PKG_CONFIG_PATH: /usr/local/opt/openssl/lib/pkgconfig
             SKIP_SSH_TESTS: true
             SKIP_NEGOTIATE_TESTS: true
-        - name: "iOS"
-          id: ios
-          os: macos-12
-          setup-script: ios
-          env:
-            CC: clang
-            CMAKE_OPTIONS: -DREGEX_BACKEND=regcomp -DDEPRECATE_HARD=ON -DUSE_LEAK_CHECKER=leaks -DUSE_GSSAPI=ON -DCMAKE_TOOLCHAIN_FILE=../ios.toolchain.cmake -DCMAKE_SYSTEM_NAME=iOS -DPLATFORM=OS64
-            CMAKE_GENERATOR: Ninja
-            PKG_CONFIG_PATH: /usr/local/opt/openssl/lib/pkgconfig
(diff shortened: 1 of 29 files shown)


Kyle-Ye:
Some notes for the reviewer:
1. Do we want to add iOS CI check here?
2. There is a flaky Windows test case failing. Maybe a rerun will temporary pass it.
[Experimental Features / Build: Windows (SHA256, amd64, Visual Studio) (pull_request)](https://github.com/libgit2/libgit2/actions/runs/10432211307/job/28892689247?pr=6862)
3. The iOS CI is failing due to a permission issue
```
ln: /usr/local/lib: Permission denied
```
I'll align the host os to macos-12 as the same version of the existing macOS CI to try to fix it.

Uncommon:
Are these changes actually specific to iOS, or do they apply to macOS as well?

Kyle-Ye:
> Are these changes actually specific to iOS, or do they apply to macOS as well?

specific to iOS

Kyle-Ye:
The CI is failing due to libgit2_tests issue.

It will be `build/libgit2_tests.app/libgit2_tests` instead of `build/libgit2_tests` for iOS.

And we can fix it by modifying `tests/libgit2/CMakeLists.txt`. But it will then be killed since it is an iOS app and we are running it directly on macOS.

I suggest for now we just skip the tests on iOS and add them back later.

ethomson:
> I suggest for now we just skip the tests on iOS and add them back later.

👍 

Thanks; I want to figure out how this integrates with #6848 since they're touching roughly the same bits of code. Will get these both merged this weekend. Thanks for this effort!

ethomson:
Thanks! I moved the iOS build into the nightlies, and I reverted the clar changes so that we could think a bit more how (or if) we would address iOS on clar.

PR #6863, opened 2024-08-20
diff: do not autoexpand unmodified trees while calculating tree to tree diff

yerseg:
During the process of calculating the tree-to-tree diff, all trees are expanded and leaf entries are compared. This can be a performance hit because the algorithm complexity becomes O(all files in the repository) instead of O(modified files between the compared trees).

I have made an attempt to improve the performance of this process by introducing a check. The check applies only to trees that have the same name. If the trees are in an unmodified state compared to each other, they are skipped. Other trees are expanded as usual.

This modification applies only to the tree-to-tree diff and doesn't have influence on other kinds of diff.

### DIFF
--- a/src/libgit2/diff_generate.c
+++ b/src/libgit2/diff_generate.c
@@ -802,7 +802,8 @@ static int maybe_modified_submodule(
 
 static int maybe_modified(
 	git_diff_generated *diff,
-	diff_in_progress *info)
+	diff_in_progress *info,
+	bool skip_pathspec_match)
 {
 	git_oid noid;
 	git_delta_t status = GIT_DELTA_MODIFIED;
@@ -817,8 +818,10 @@ static int maybe_modified(
 
 	git_oid_clear(&noid, diff->base.opts.oid_type);
 
-	if (!diff_pathspec_match(&matched_pathspec, diff, oitem))
-		return 0;
+	if (!skip_pathspec_match) {
+		if (!diff_pathspec_match(&matched_pathspec, diff, oitem))
+			return 0;
+	}
 
 	/* on platforms with no symlinks, preserve mode of existing symlinks */
 	if (S_ISLNK(omode) && S_ISREG(nmode) && new_is_workdir &&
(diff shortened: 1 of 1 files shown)


yerseg:
@ethomson Hi! Can you approve workflows again, please?

yerseg:
@ethomson sorry to disturb you but what do you think about these changes? Do they seem valid or not?

ethomson:
Hi @yerseg - no disturbance, thanks for the ping. I'll try to 👀 this soon. 🙏

yerseg:
@ethomson what’s up? will you have free time to review this?

OrHayat:
hi @ethomson 
i analyized a script that uses pygit2 and it was too slow
after checking with claude he found the same gap as this pr
and i checked and found that this pr exists so i built pygit2 with that branch and ran benchamrk
it improved performance by alot.
to make results reproducible i  reran it on this git and this is the results.

The branch is based on v1.8.2+27, so I cherry-picked its 7 commits onto v1.9.7 —
applies clean, same +101/-15 in `diff_generate.c`. Built v1.9.7 and v1.9.7+PR with
identical flags (Release, `BUILD_SHARED_LIBS=ON`, `USE_SSH=OFF`, system zlib, SHA1
CollisionDetection) and ran both through pygit2 1.20.1 built from source against each.
macOS 15, arm64.

Test repo is libgit2.git itself at `main`: 16450 commits, 15606 files, 76 MiB pack.
Each of the newest 3000 single-parent commits diffed against its parent.

| | v1.9.7 | + this PR | |
|---|---|---|---|
| `diff(main_tree, main_tree)` x1000 | 3.54s | 0.01s | ~300x |
| 3000 tree diffs, deltas only | 7.90s | 0.19s | 42x |
| 3000 diffs + `git_diff_patchid` | 10.11s | 2.13s | 4.7x |

git on the same 3000 commits: `git log --raw` 0.15s, `git log -p | git patch-id --stable`
1.08s. So the tree walk now lands near git, where before it was ~50x off. What is left
in the patchid row is xdiff and hashing, which this PR does not touch.

Both builds produce identical output: same delta count per commit (21394 total), same
`git_diff_patchid` for all 3000, same path lists on the commits I checked.

Two caveats: this is measured through pygit2 rather than a C harness (same overhead on
both sides, so the ratios hold), and I benchmarked without running the test suite.

PR #6874, opened 2024-09-02
Make packbuilder interruptible using progress callback

roberth:
Specifically, forward errors from `packbuilder->progress_cb`

This allows the callback to gracefully terminate long-running operations when the application is interrupted.

Interruption could be `^C` in the terminal, but this could be any other condition or event, as this is up to the callback function to implement.

My specific motivation for this is to improve the responsiveness of the Nix package manager to interrupts in the terminal, in the context of https://github.com/NixOS/nix/pull/11330, which proposes to write packfiles instead of loose objects for performance.


ethomson:
Helpful fix, thanks!

PR #6875, opened 2024-09-02
Add `git_mempack_write_thin_pack`

roberth:
Unlike existing functions, this produces a _thin_ packfile by making use of the fact that only new objects appear in the mempack Object Database.

A thin packfile only contains certain objects, but not its whole closure of references. This makes it suitable for efficiently writing sets of new objects to a local repository, by avoiding many small I/O operations.

This relies on `write_pack` (e.g. `git_packbuilder_write_buf`) to implement the "recency order" optimization step.

Basic measurements comparing against the writing of individual objects show a speedup during when writing large amounts of content on machines with comparatively slow I/O operations, and little to no change on machines with fast I/O operations.

I have used this in https://github.com/NixOS/nix/pull/11330, in conjunction with #6874

### DIFF
--- a/src/libgit2/odb_mempack.c
+++ b/src/libgit2/odb_mempack.c
@@ -132,6 +132,29 @@ cleanup:
 	return err;
 }
 
+int git_mempack_write_thin_pack(git_odb_backend *backend, git_packbuilder *pb)
+{
+	struct memory_packer_db *db = (struct memory_packer_db *)backend;
+	const git_oid *oid;
+	size_t iter = 0;
+	int err;
+
+	while (true) {
+		err = git_oidmap_iterate(NULL, db->objects, &iter, &oid);
+
+		if (err == GIT_ITEROVER)
+			break;
+		else if (err != 0)
+			return err;
+
+		err = git_packbuilder_insert(pb, oid, NULL);
+		if (err != 0)
+			return err;
+	}
+
+	return 0;
+}
+
 int git_mempack_dump(
 	git_buf *pack,
 	git_repository *repo,
--- a/include/git2/sys/mempack.h
+++ b/include/git2/sys/mempack.h
@@ -45,6 +45,25 @@ GIT_BEGIN_DECL
 GIT_EXTERN(int) git_mempack_new(git_odb_backend **out);
 
 /**
+ * Write a thin packfile with the objects in the memory store.
+ *
+ * A thin packfile is a packfile that does not contain its transitive closure of
+ * references. This is useful for efficiently distributing additions to a
+ * repository over the network, but also finds use in the efficient bulk
+ * addition of objects to a repository, locally.
+ *
+ * This operation performs the (shallow) insert operations into the
+ * `git_packbuilder`, but does not write the packfile to disk;
+ * see `git_packbuilder_write_buf`.
+ *
+ * It also does not reset the in-memory object database; see `git_mempack_reset`.
+ *
+ * @param backend The mempack backend
+ * @param pb The packbuilder to use to write the packfile
+ */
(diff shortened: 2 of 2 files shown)


ethomson:
Thanks! Really useful addition; I'll review this shortly. 🙏

Mic92:
@ethomson friendly ping.

src/libgit2/odb_mempack.c
```diff
@@ -132,6 +132,35 @@ static int git_mempack__dump(
...
+	    wrt heuristics: ask your favorite LLM to translate https://git-scm.com/docs/pack-heuristics/en
+		to actual normal reference documentation. */
+	while (true) {
+		err = git_oidmap_iterate(NULL, db->objects, &iter, &oid);
+		if (err == GIT_ITEROVER) {
+			err = 0;
+			break;
+		}
+		if (err != 0)
+			return err;
```
ethomson, line 154: It looks like we don't actually do anything with `err` outside of this loop, so we don't need to set it before breaking?

```suggestion

		if (err == GIT_ITEROVER)
			break;
		else if (err != 0)
			return err;
```

src/libgit2/odb_mempack.c
```diff
@@ -132,6 +132,35 @@ static int git_mempack__dump(
 	return err;
 }
 
+int git_mempack_write_thin_pack(git_odb_backend *backend, git_packbuilder *pb)
+{
+	struct memory_packer_db *db = (struct memory_packer_db *)backend;
+	const git_oid *oid;
+	size_t iter = 0;
+	int err = -1;
```
ethomson, line 140: It looks like we don't depend on the initial condition? A minor nit, but

```suggestion
	int err;
```

ethomson (commented):
Thanks for this. I made two very trivial minor suggestions in keeping with the library's styling. Otherwise, a great addition. 🙏

src/libgit2/odb_mempack.c
```diff
@@ -132,6 +132,34 @@ static int git_mempack__dump(
...
+{
+	struct memory_packer_db *db = (struct memory_packer_db *)backend;
+	const git_oid *oid;
+	size_t iter = 0;
+	int err;
+
+	/* TODO: Implement the recency heuristics.
+	    For this it probably makes sense to only write what's referenced
+		through commits, an option I've carved out for you in the docs.
+	    wrt heuristics: ask your favorite LLM to translate https://git-scm.com/docs/pack-heuristics/en
```
Mic92, line 145:
```suggestion
	    wrt heuristics: ask your favorite LLM to translate https://git-scm.com/docs/pack-heuristics
```

roberth: Hmm wait, I thought I removed this comment because the recency heuristics are implemented at a later stage.

ethomson: Awesome; thanks. Once CI finishes, I'll merge.

roberth: Done, and I've squashed the edits

Mic92:
CI is green!

Mic92:
Question, will this patch be part of the v1.8.2 or later?

roberth:
Thank you @ethomson  for reviewing and accepting!

Ericson2314:
> I'm going to cut a release 🔜

Any ETA on this?

Mic92:
1.8.4 was released a while ago but this patch is no included. Are the patch releases made of a different branch?

Ericson2314:
@Mic92 Ah I see it is in 1.9.0: https://github.com/libgit2/libgit2/blob/main/docs/changelog.md#v190

PR #6876, opened 2024-09-06
push: report a push status even the push has failed

yerseg:
In case when `unpack_ok` is false push status can be non-empty and contains important information about unpacking failure cause. For example, Azure DevOps git server behave like that. 
There is an example of error message is provided by Azure:
"The blob object e4f9f86ce8fcaa697809bcb6efbd91eddc50b04a was rejected: VS403658: The object is 1074200997 bytes, which is greater than the 10485760 byte limit for objects in this repository." 
If we go to cleanup immediately it will be no information about the cause for client.


ethomson:
Thanks for the fix; this is a helpful change!

PR #6877, opened 2024-09-06
ssl: ability to add raw X509 certs to the cert store

yerseg:
There are custom cert stores on macOS and Windows (Apple Keychain, Windows Credentials Manager). In our product we enumerate certs in the stores and add them to SSL cert store. It seems like useful feature at all.


yerseg:
Thank you for review! 
1. I've enhanced the doc. Seems like it is enough information about a cert lifetime.
2. I've not fully understood your point :( What did you mean? This feature only allows you to add raw certificates. The responsibility for obtaining the certificates, however, lies with the client code and not the library.

ethomson:
Thanks; I think that's a helpful improvement!

> I've not fully understood your point :( What did you mean? This feature only allows you to add raw certificates. The responsibility for obtaining the certificates, however, lies with the client code and not the library.

In your initial comment, you mentioned that "there are custom cert stores on macOS and Windows"; but I don't understand what that comment has to do with this PR.

In your application, are you using those custom cert stores on macOS and Windows directly, and then libgit2 makes use of them just because that's the architecture of those libraries? But there's no real way to do this on OpenSSL, which is why you're creating this PR?

Or are there custom stores on those systems, but this PR doesn't address those?

This is just my ignorance about how these custom certificate systems work on all those platforms, I'm hoping to get some more of this knowledge about how you're working.

Thanks!

yerseg:
Sorry for the misunderstanding. I mentioned certificate stores to provide context for the feature. Let me explain further:  

On macOS, there's the Apple Keychain, and on Windows, the Windows Credential Manager. Both offer APIs to list certificates in the store and retrieve them in X509 format, which our application does.  

Once we retrieve the certificates, we need to add them to the SSL backend in libgit2. My changes make this possible.

ethomson:
Oh! You're building with OpenSSL on all platforms and you're pulling carts out of the various platform specific subsystems and then feeding them into OpenSSL. Is that roughly accurate?

I had assumed that you were building with OpenSSL on Linux (and using the win32 and Mac TLS stacks on windows and Mac, respectively).

That all makes sense; like I said, just curious about your use case and how / or if we should think about this setting for Mac and windows.

ethomson:
(No need to apologize btw, I was making assumptions.)

yerseg:
Yes, you’re right. We’re building with openssl on all platforms (win, mac, linux) because of our app specific. We don’t like the OS specific ssl stacks for less transparency.
Also when building libgit2 we use patched cmake-scripts with customized openssl target because of static linking.
I don’t know is it possible to provide full support of certificates managing with openssl on mac and win. It will lead usage of the additional platform APIs and so I don’t expect that kind of support in the libgit.

ethomson:
Question: does this compile with `cmake -DUSE_HTTPS=OpenSSL-Dynamic`? I think that we might need to load `X509_STORE_add_cert` symbol there.

include/git2/common.h
```diff
@@ -336,6 +337,16 @@ typedef enum {
...
+ * 
+ *  * opts(GIT_OPT_ADD_SSL_X509_CERT, const X509 *cert)
+ *     
+ *      > Add a raw X509 certificate into the SSL certs store.
+ *      > The added certificate is not persisted and will be used
+ *      > only during application lifetime. Also there is no ability
+ *      > to remove a certificate from the store during app lifetime
+ *      > after it was added. 
+ * 		> 
+ * 		> - `cert` is the raw X509 cert will be added to cert store.
```
ethomson, line 349:
```suggestion
 * 
 *  * opts(GIT_OPT_ADD_SSL_X509_CERT, const X509 *cert)
 *     
 *  		 > Add a raw X509 certificate into the SSL certs store.
 *  		 > The added certificate is not persisted and will be used
 *  		 > only during application lifetime. Also there is no ability
 *  		 > to remove a certificate from the store during app lifetime
 *  		 > after it was added. 
 * 		> 
 * 		> - `cert` is the raw X509 cert will be added to cert store.
```

yerseg:
> Question: does this compile with cmake -DUSE_HTTPS=OpenSSL-Dynamic? I think that we might need to load X509_STORE_add_cert symbol there.

Fixed in commit below.

yerseg:
I have also added a test that is similar to others in the 'online::customcert' test suite. However, I need assistance setting up an additional test server hosted at `https://test.libgit2.org:3443/` and generating a test cert. This cert must replace the template `self-signed.pem.raw` I've already placed in the repo.

ethomson:
I needed the cert and the key, so I created a new one for test.libgit2.org and updated the PR. I also noticed that the custom cert tests would fail unless you cleared the context between runs, so I added a (currently) internal function to do that.

This passes the CI runs - please take a look and make sure that I didn't break anything?

yerseg:
Your changes looks fine. Thanks for help!

ethomson:
Thanks @yerseg !

Issue #6275, opened 2022-04-14
Difference in behaviour between git ls-remote and libgit2 equivalent


apnadkarni:
I'm not sure if this is a bug or a by-design difference from the official git. The official git implementation includes a peeled entry (with ^{} suffix) for refs to blobs. The libgit2 implementation (git_remote_ls -> local_ls) does not due to this [check ](https://github.com/libgit2/libgit2/blob/63970244fb2d49794e7b5d268a2defe4299fd3ad/src/libgit2/transports/local.c#L112) which only attempts to peel if the ref is under the `refs/tags` directory. In the example below, the ref is under `refs/blobs` directory.

Only bringing it to your notice in case this is a bug and not an intended difference.

### Reproduction steps

Console session:

```
D:\Temp>git clone \src\libgit2\tests\resources\testrepo.git lg2-test
Cloning into 'lg2-test'...
done.

D:\Temp>cd lg2-test

D:\Temp\lg2-test>git ls-remote origin
a65fedf39aefe402d3bb6e24df4d4f5fe4547750        HEAD
521d87c1ec3aef9824daf6d96cc0ae3710766d91        refs/blobs/annotated_tag_to_blob
1385f264afb75a56a5bec74243be9b367ba4ca08        refs/blobs/annotated_tag_to_blob^{}
a4a7dce85cf63874e984719f4fdd239f5145052f        refs/heads/br2
a4a7dce85cf63874e984719f4fdd239f5145052f        refs/heads/cannot-fetch
e90810b8df3e80c413d903f631643c716887138d        refs/heads/chomped
258f0e2a959a364e40ed6603d5d44fbb24765b10        refs/heads/haacked
a65fedf39aefe402d3bb6e24df4d4f5fe4547750        refs/heads/master
a65fedf39aefe402d3bb6e24df4d4f5fe4547750        refs/heads/not-good
41bc8c69075bbdb46c5c6f0566cc8cc5b46e8bd9        refs/heads/packed
4a202b346bb0fb0db7eff3cffeb3c70babbd2045        refs/heads/packed-test
763d71aadf09a7951596c9746c024e7eece7c7af        refs/heads/subtrees
e90810b8df3e80c413d903f631643c716887138d        refs/heads/test
9fd738e8f7967c078dceed8190330fc8648ee56a        refs/heads/track-local
e90810b8df3e80c413d903f631643c716887138d        refs/heads/trailing
[...]

PR #6881, opened 2024-09-12
transport: do not filter tags based on ref dir in local

rindeal:
Let `git_object_type()` do the filtering instead.

Fixes: https://github.com/libgit2/libgit2/issues/6275

### DIFF
--- a/src/libgit2/transports/local.c
+++ b/src/libgit2/transports/local.c
@@ -108,10 +108,6 @@ static int add_ref(transport_local *t, const char *name)
 		return error;
 	}
 
-	/* If it's not a tag, we don't need to try to peel it */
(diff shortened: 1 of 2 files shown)


ethomson:
Hi @rindeal! This seems like it makes sense! The tests are failing, though - are the tests wrong (and should have been reporting these refs in the first place?) or is the implementation not quite right?

ethomson:
> As detailed in the linked issue, the function wasn't peeling refs/blobs/annotated_tag_to_blob, resulting in the output being one line shorter. I've updated the tests so all should be well now.

Ah, thanks, I think that I may have noticed that in my initial review and forgotten when I came back to this. In any case, very cool, thanks for the fix.

PR #6882, opened 2024-09-14
Add LIBGIT2_VER_CHECK Macro for Version Comparison

HamedMasafi:
This merge introduces a LIBGIT2_VERSION_CHECK macro in git2/version.h, similar to Qt's QT_VERSION_CHECK. The macro allows developers to compare the current version of libgit2 at compile time, making it easier to ensure compatibility with specific versions of the library.

Key Changes:
- Added LIBGIT2_VER_CHECK macro for version comparison.
- Defined LIBGIT2_VER_NUMBER as a single integer representing the current version.
- Updated comments to explain the usage of the new macro.

This enhancement will improve the ability to manage version-dependent code more efficiently. Please review the changes and provide feedback!


tawhidnazari57 (approved):

ethomson:
I think that this is generally useful. I'd rather see our version number be human-parseable in a way that reflects the version number. For example:

```
#define LIBGIT2_VERSION_NUMBER ( \
    (LIBGIT2_VER_MAJOR * 1000000) +
    (LIBGIT2_VER_MINOR * 10000) +
    (LIBGIT2_VER_REVISION * 100) +
     LIBGIT2_VER_PATCH )
```

so that v1.8.2 is `1080200`.

This also lets callers just say 

`#if LIBGIT2_VERSION_NUMBER >= 1060300`

which seems pretty readable to me. But I think that it's worth offering up a macro in case that's _not_ very readable. But I think that I would put it as something that evaluates to truthy given a version number. For example:

```
#define LIBGIT2_VERSION_CHECK(a,b,c) (LIBGIT2_VERSION_NUMBER >= ((a)*1000000)+((b)*10000)+((c)*100))
```

Then you'd use it as

`#if LIBGIT2_VERSION_CHECK(1,6,3)`...

ethomson:
Thanks! I updated based on my suggestions.

PR #6884, opened 2024-09-20
Leak in truncate_racily_clean in index.c

lstoppa:
In truncate_racily_clean, when git_diff_index_to_workdir fails, we leak the local variable git_vector paths.


ethomson:
Good catch; thanks!

Issue #6873, opened 2024-09-02
Error: not able to open repository “unsupported extension name extensions.preciousobjects”.


ao2:
libgit2 does not seem to handle AOSP manifest repositories.

I experienced the issue both with [gitg](https://gitlab.gnome.org/GNOME/gitg) which used `libgit2`, and with the `lg2` example command.

### Reproduction steps

```
$ cd /tmp
$ mkdir AOSP
$ cd AOSP
$ repo init -u https://android.googlesource.com/platform/manifest -b main
Downloading Repo source from https://gerrit.googlesource.com/git-repo
repo has been initialized in /tmp/AOSP
$ cd .repo/manifests
$ /tmp/examples/build/lg2 status
Unable to open repository '%s' '.' [-1] - unsupported extension name extensions.preciousobjects
$ gitg 
Error: not able to open repository “unsupported extension name extensions.preciousobjects”.
```

### Expected behavior

`libgit2` is able to open the repo and list all the branches and commits.

### Actual behavior

`libgit2` fails to open the repository with the following error:

```
Unable to open repository '%s' '.' [-1] - unsupported extension name extensions.preciousobjects
```

### Version of libgit2 (release number or SHA1)

1.7.2

### Operating system(s) tested

Debian Unstable / Linux

ethomson:
In this case, this feels like a useful thing to report to gitg, if you haven't already.

The `preciousObjects` extension is a bit of an oddity. It has nothing to do with client support; instead, it indicates that the client should not do any deletion of objects. We do not implicitly include this extension, since we don't know what the client itself is going to do.

If gitg understands this extension -- either because they don't do any object deletion (ever), or because they understand and obey `preciousObjects` then they should [indicate to libgit2 that they understand the extension](https://github.com/libgit2/libgit2/blob/782e29c906f6e44b120843356f286b6a97d89f88/include/git2/common.h#L474).

Probably our example should, too, so that people see how to use it. (Do note that lg2 is _example code_, though, it's not for production.)

ao2:
Thanks @ethomson I will report to `gitg` too.

Does a similar reasoning also apply for extensions like `extensions.partialclone`? Is that also supposed to be handled by the client?

I ask because this is also something that libgit2 complains about in AOSP project repositories:

```
$ cd /tmp/AOSP
$ repo sync -c -j8 --optimize --no-tags
$ cd build/core/
$ gitg 
Error: not able to open repository “unsupported extension name extensions.partialclone”.
```

[...]

PR #6886, opened 2024-09-24
repo: support the preciousObjects extension

ethomson:
libgit2 implicitly supports precious objects, since there's no gc command, nor even any option in our object database functionality to delete an object from the odb.

In the future, when we support deleting objects, or a gc functionality, we will need to honor the preciousObjects extension and reject those APIs when it is set. In the meantime, since users cannot use libgit2 (directly) to delete an object, we can simply add this to our allowlist.


ethomson:
Fixes #6873 6873

PR #6887, opened 2024-09-24
diff: always print the file header on GIT_DIFF_FORMAT_PATCH_HEADER 

carlosmn:
In `diff_print_patch_file` we try to avoid printing the file if we're not sure
that there will be some content. This works fine if we have any later functions
that might get called as part of printing. But when given the format
`GIT_DIFF_FORMAT_PATCH_HEADER`, this is the only function to get called.

Force us to print the diff header for the file in such a situation. This may
still print the header where e.g. `GIT_DIFF_FORMAT_PATCH` wouldn't, but it's
better than this option not working at all.

This fixes a regression introduced in #5893 that's making the rugged tests fail to pass. This sort of regresses that fix by sending out the header more than we'd like, but this is a fix I could make quickly.

An alternative might be to still call the hunk function but have it return early if the format is just asking for the header (that might require teaching a few of these functions though as we could still instead land in the binary function).

### DIFF
--- a/src/libgit2/diff_print.c
+++ b/src/libgit2/diff_print.c
@@ -580,8 +580,12 @@ static int diff_print_patch_file_binary(
 	return error;
 }
 
-GIT_INLINE(int) should_force_header(const git_diff_delta *delta)
+GIT_INLINE(int) should_force_header(const diff_print_info *pi, const git_diff_delta *delta)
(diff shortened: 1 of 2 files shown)


ethomson:
Closed in favor of #6888

